PR #33341, opened 2021-06-08
Drop support for 1.9 proxies

howardjohn:
We do not support Istiod 1.11 and 1.9 proxy, so clean this up a bit

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/accesslog.go
+++ b/pilot/pkg/networking/core/v1alpha3/accesslog.go
@@ -28,25 +28,16 @@ import (
 	structpb "github.com/golang/protobuf/ptypes/struct"
 
 	meshconfig "istio.io/api/mesh/v1alpha1"
-	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pilot/pkg/networking/util"
 	"istio.io/istio/pkg/util/protomarshal"
 	"istio.io/pkg/log"
 )
 
 const (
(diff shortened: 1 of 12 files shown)


hzxuzhonghu:
We donot support every release version in production, we support every other version currently. Like 1.9->1.11, with this it is hard to do graceful upgrade.

howardjohn:
You have Istiod 1.11 connecting to 1.9 proxies? Or 1.9 proxies connecting
to 1.11 proxies?

I don't know how the first one would ever work. We break it in almost every
version, such as the xds v3 migration?

hzxuzhonghu:
We want this: Istiod 1.11 connecting to 1.9 proxies
We did not do graceful upgrade before, as community does not support it.  If we keeps the compatibility, surely users can do upgrade according to their business.  For example, users are reluctant to upgrade their DB unless they must do.

howardjohn:
I would recommend you follow this approach: https://istio.io/latest/blog/2021/revision-tags/

Then there is no cross-version traffic but you can still skip versions. Otherwise it's not safe nor is it something we have ever even attempted to support

hzxuzhonghu:
IC, looks super good feature. Thanks！

hzxuzhonghu (approved):

lei-tang (approved):

lei-tang:
Quoted from the PR description "We do not support Istiod 1.11 and 1.9 proxy", which version(s) of proxies are supported by istiod 1.11?

howardjohn:
1.11 and 1.10

Issue #33298, opened 2021-06-07
Gateway-API's Gateway does not work when it points to the "second" k8s service


nak3:
**Bug description**

- K8S Gateway does not work when it points to the "second" k8s service. (Pleases see the reproduce steps below for what is the second service.)

[X] Networking

**Expected behavior**

- K8S Gateway works with the "second " k8s service.

**Steps to reproduce the bug**

#### 1. install gateway-api, istio and sleep client pod

```
kubectl apply -k 'github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.3.0'
istioctl install -y
kubectl apply -f samples/sleep/sleep.yaml
```

#### 2. Deploy a second with `istio: ingressgateway` selector

```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Service
metadata:
  name: knative-local-gateway
  namespace: istio-system
spec:
  type: ClusterIP
  selector:
    istio: ingressgateway
  ports:
    - name: http2
      port: 80
      targetPort: 8081
EOF
```

Now, the service and endpoint like this:

```
$ kubectl get svc -n istio-system  -o wide
NAME                    TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)                                      AGE   SELECTOR
istio-ingressgateway    NodePort    10.96.150.167   <none>        15021:31634/TCP,80:30081/TCP,443:30267/TCP   13m   app=istio-ingressgateway,istio=ingressgateway
istiod                  ClusterIP   10.96.201.250   <none>        15010/TCP,15012/TCP,443/TCP,15014/TCP        14m   app=istiod,istio=pilot
knative-local-gateway   ClusterIP   10.96.145.203   <none>        80/TCP                                       12m   istio=ingressgateway

$ kubectl get ep -n istio-system 
NAME                    ENDPOINTS                                                        AGE
istio-ingressgateway    10.244.1.4:15021,10.244.1.4:8080,10.244.1.4:8443                 11m
istiod                  10.244.1.5:15012,10.244.1.5:15010,10.244.1.5:15017 + 1 more...   12m
knative-local-gateway   10.244.1.4:8081                                                  10m
```

(I updated istio-ingressgateway svc to NodePort, but it should not be matter.)

#### 3. Create Gateway to use `knative-local-gateway.istio-system`

```
cat <<EOF | kubectl apply -f -
apiVersion: networking.x-k8s.io/v1alpha1
kind: GatewayClass
metadata:
  name: istio
spec:
[...]

PR #33348, opened 2021-06-09
Fix using new Gateway api with legacy gw annotation

howardjohn:
This ensures the annotation doesn't mess with the new gateway mode,
since it doesn't make sense there.

Fixes https://github.com/istio/istio/issues/33298#issuecomment-857465111

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/model/gateway.go
+++ b/pilot/pkg/model/gateway.go
@@ -275,8 +275,10 @@ func resolvePorts(number uint32, instances []*ServiceInstance, legacyGatewaySele
 	}
 	ports := map[uint32]struct{}{}
 	for _, w := range instances {
-		if _, directPortTranslation := w.Service.Attributes.Labels[DisableGatewayPortTranslationLabel]; directPortTranslation {
+		if _, directPortTranslation := w.Service.Attributes.Labels[DisableGatewayPortTranslationLabel]; directPortTranslation && legacyGatewaySelector {
 			// Skip this Service, they opted out of port translation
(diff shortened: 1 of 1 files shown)


pilot/pkg/model/gateway.go
```diff
@@ -275,8 +275,10 @@ func resolvePorts(number uint32, instances []*ServiceInstance, legacyGatewaySele
 	}
 	ports := map[uint32]struct{}{}
 	for _, w := range instances {
-		if _, directPortTranslation := w.Service.Attributes.Labels[DisableGatewayPortTranslationLabel]; directPortTranslation {
+		if _, directPortTranslation := w.Service.Attributes.Labels[DisableGatewayPortTranslationLabel]; directPortTranslation && legacyGatewaySelector {
```
stevenctl, line 278: is `legacyGatewaySelector` ever true? It seems like `gatewayWithInstances.legacyGatewaySelector` has no assignments

howardjohn: I think its just your IDE missing the un-named assignment , I had the same problem.. https://github.com/istio/istio/blob/00f1548764e77b94276f49adc9539a6fad02c5ab/pilot/pkg/model/push_context.go#L1752

stevenctl (approved):

Issue #33351, opened 2021-06-09
"istioctl admin" subcommand for control-plane operations


esnible:
We have an "istioctl admin" subcommand, but currently it only holds the command to set control plane logging levels.

Move all commands that administer Istio itself to that subcommand.

Move all commands that require access to the control plane to that subcommand.

The design for this work being tracked in [this RFC](https://docs.google.com/document/d/1NPkq0vN83vCZM01S8JVqoCzmf8E4JfT_Vwr9m0Jke68/edit#).

The direction of this was approved as part of UX's 2021 roadmap and 1.11 roadmaps.
Moving `istioctl proxy-status` there, and limiting the functionality in the original, was approved for 1.10 but slipped.

esnible:
I am closing this due to comments on the closed PR.

PR #33352, opened 2021-06-09
Move control plane subcommands to "istioctl admin"

esnible:
Resolves https://github.com/istio/istio/issues/33351

### DIFF
--- a/istioctl/cmd/admin.go
+++ b/istioctl/cmd/admin.go
@@ -20,6 +20,24 @@ import (
 	"github.com/spf13/cobra"
 )
 
+func adminDashboard() *cobra.Command {
+	return &cobra.Command{
+		Use:     "dashboard",
+		Aliases: []string{"dash", "d"},
+		Short:   "Access to Istio web UIs",
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 0 {
+				return fmt.Errorf("unknown dashboard %q", args[0])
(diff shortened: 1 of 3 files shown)


esnible:
test all

howardjohn (commented):
This is highly personal preference, so I wanted to give my 2c. I am not a big fan of this change. I think it adds unnecessary toil (need to change commands), especially in areas we already added toil (manifest apply -> manifest install -> install -> admin install. Other applications just had `foo install` the whole time and beat us on day 0 experience).

kubectl also has varying degrees of privilege require to run commands, without this distinction. For example, `kubectl cordon` or `kubectl certificate approve`. Discoverability of these is solved by groupings of commands _in documentation_ (see kubectl help), not in the command itself, as well as good error handling for cases when users without permissions try to run the command.

Looking at the commands, its also not clear what "admin" is. Why can I not generate a manifest as a user? That requires no privileges. `install` can also be used to deploy gateways to a namespace which is not an admin operation nor requires admin privileges. If we describe "admin" as uses istio-system privs, then there are also quite a few other commands that seem like they should be `admin` - bug-report, verify-install, analyze, proxy-status... its really almost all of them.

I would recommend we follow kubectl's approach, and instead of grouping all commands in `Available Commands`, we split it into `Admin Commands` and `User Commands` (or whatever we choose to call it). However, I am not a UX expert

PR #33355, opened 2021-06-09
WIP: Adding Grpc metrics and debug

costinm:
gRPC provides a number of useful debug tools - admin allows checking the internal channels, zpages provide 
traces and metrics for the specific workload ( including the agent ). Also when aggregated the metrics will
follow the typical pattern/naming expected for gRPC - including messages sent in streaming connections, etc.

We are using OC already, so not a new dependency.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/bootstrap/monitoring.go
+++ b/pilot/pkg/bootstrap/monitoring.go
@@ -23,6 +23,7 @@ import (
 	ocprom "contrib.go.opencensus.io/exporter/prometheus"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.opencensus.io/stats/view"
+	"go.opencensus.io/zpages"
 
 	"istio.io/pkg/log"
 	"istio.io/pkg/monitoring"
@@ -63,6 +64,8 @@ func addMonitor(mux *http.ServeMux) error {
 			log.Errorf("Unable to write version string: %v", err)
 		}
 	})
+
+	zpages.Handle(mux, "/debug")
 
 	return nil
 }
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -34,6 +34,7 @@ import (
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/h2c"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/admin"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/reflection"
 	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -760,6 +761,8 @@ func (s *Server) initSecureDiscoveryService(args *PilotArgs) error {
 	s.secureGrpcServer = grpc.NewServer(opts...)
 	s.XDSServer.Register(s.secureGrpcServer)
 	reflection.Register(s.secureGrpcServer)
+
+	admin.Register(s.secureGrpcServer)
 
 	s.addStartFunc(func(stop <-chan struct{}) error {
 		go func() {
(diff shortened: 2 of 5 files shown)


pilot/pkg/bootstrap/server.go
```diff
@@ -761,6 +762,8 @@ func (s *Server) initSecureDiscoveryService(args *PilotArgs) error {
 	s.XDSServer.Register(s.secureGrpcServer)
 	reflection.Register(s.secureGrpcServer)
 
+	admin.Register(s.secureGrpcServer)
```
howardjohn, line 765: I am only seeing metrics for IstioCertificateService for some reason, not sure if that is expected?

howardjohn: And on agent not seeing any either. Does it only support unary?

its possibly a good thing to not expose on the agent.. we have concerns about metrics cardinality at agent

pkg/istio-agent/xds_proxy.go
```diff
@@ -62,6 +63,8 @@ import (
 	"istio.io/istio/security/pkg/nodeagent/caclient"
 	"istio.io/istio/security/pkg/pki/util"
 	"istio.io/pkg/log"
+
+	"go.opencensus.io/zpages"
```
howardjohn, line 67: any idea how expensive this is? Safe to always enable?

pilot/pkg/bootstrap/server.go
howardjohn: Oh I think it lazy loads the metrics, and it only registers after stream close. So when once I had disconnected once they did show up.

This adds 256 new metrics to the agent. I think its way too much, and not *that* useful - envoy itself is only adding 286 and it already causes complaints.

pkg/istio-agent/xds_proxy.go
costinm: Just some extra handlers - using the existing data.

pilot/pkg/bootstrap/server.go
costinm: I'll add an env variable to control. 

Like Istio, gRPC has some 'standard' metrics and debug tools - and it seems reasonable to support them, but agree we may not want them in all cases.

For proxyless - there is no envoy, and we are mostly blind ( metrics-wise ).

pilot/pkg/bootstrap/monitoring.go
```diff
@@ -64,6 +65,8 @@ func addMonitor(mux *http.ServeMux) error {
 		}
 	})
 
+	zpages.Handle(mux, "/debug")
```
adiprerepa, line 68: should this be wrapped with an `if features.GRPCDebug`?

hzxuzhonghu: What can zpages show?

pkg/istio-agent/xds_proxy.go
```diff
@@ -615,6 +622,7 @@ func (p *XdsProxy) buildUpstreamClientDialOpts(sa *Agent) ([]grpc.DialOption, er
 	// Make sure the dial is blocking as we dont want any other operation to resume until the
 	// connection to upstream has been made.
 	dialOptions := []grpc.DialOption{
+		grpc.WithStatsHandler(new(ocgrpc.ClientHandler)),
```
hzxuzhonghu, line 625: need to make it optional, i remember grpc tracing was disabled dual to performance

PR #33356, opened 2021-06-09
Allow gateway-api to function without Kubernetes RBAC

howardjohn:
This implements most of
https://docs.google.com/document/d/1gDhj058q2PKKI9SWpxE23ExEXdOkLiueT8JA_oEJ8Mo/edit#heading=h.bmkc9oooxqbj

Another PR will try to figure out the multicluster change
I tried to capture context in the comments in the code

### DIFF
--- a/pilot/pkg/xds/sds.go
+++ b/pilot/pkg/xds/sds.go
@@ -93,28 +93,64 @@ func needsUpdate(proxy *model.Proxy, updates model.XdsUpdates) bool {
 	return false
 }
 
-// Currently only same namespace is allowed. In the future this will be expanded.
-func (s *SecretGen) proxyAuthorizedForSecret(proxy *model.Proxy, sr SecretResource) error {
-	if proxy.ConfigNamespace != sr.Namespace {
-		return fmt.Errorf("SDS is currently only supporting accessing secret within the same namespace. Secret namespace %q does not match proxy namespace %q",
-			sr.Namespace, proxy.ConfigNamespace)
+func parseResources(names []string, proxy *model.Proxy) []SecretResource {
+	res := make([]SecretResource, 0, len(names))
+	for _, resource := range names {
+		sr, err := parseResourceName(resource, proxy.VerifiedIdentity.Namespace, string(proxy.Metadata.ClusterID))
+		if err != nil {
+			pilotSDSCertificateErrors.Increment()
+			log.Warnf("error parsing resource name: %v", err)
+			continue
+		}
+		if proxy.VerifiedIdentity.Namespace != sr.Namespace {
+			pilotSDSCertificateErrors.Increment()
+			log.Warnf("requested secret %v not accessible for proxy %v: SDS is currently only supporting accessing secret within the same namespace. "+
+				"Secret namespace %q does not match proxy namespace %q",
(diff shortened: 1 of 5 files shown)


howardjohn:
Put a hold to add better tests

pilot/pkg/xds/sds.go
```diff
@@ -102,50 +103,101 @@ func (s *SecretGen) proxyAuthorizedForSecret(proxy *model.Proxy, sr SecretResour
...
+	if err != nil {
+		log.Warnf("proxy %v is from an unknown cluster, cannot retrieve certificates: %v", proxy.ID, err)
+		return nil, model.DefaultXdsLogDetails, nil
+	}
+
+	resources := parseResources(w.ResourceNames, proxy)
+	authMode := getAuthorizationMode(resources, proxy)
+	if authMode != OnlyLocalCertificateReferences {
+		if err := secrets.Authorize(proxy.VerifiedIdentity.ServiceAccount, proxy.VerifiedIdentity.Namespace); err != nil {
+			log.Warnf("proxy %v is not authorized to receive secrets: %v", proxy.ID, err)
```
stevenctl, line 185: do we want to increment the metric here?

pilot/pkg/xds/sds.go
```diff
@@ -102,50 +103,101 @@ func (s *SecretGen) proxyAuthorizedForSecret(proxy *model.Proxy, sr SecretResour
...
 	}
 	var updatedSecrets map[model.ConfigKey]struct{}
 	if !req.Full {
 		updatedSecrets = model.ConfigsOfKind(req.ConfigsUpdated, gvk.Secret)
 	}
+
+	// TODO: For the new gateway-api, we should always search the config namespace and stop reading across all clusters
+	secrets, err := s.secrets.ForCluster(proxy.Metadata.ClusterID)
+	if err != nil {
+		log.Warnf("proxy %v is from an unknown cluster, cannot retrieve certificates: %v", proxy.ID, err)
```
stevenctl, line 177: ditto: metric

pilot/pkg/xds/sds.go
```diff
@@ -102,50 +103,101 @@ func (s *SecretGen) proxyAuthorizedForSecret(proxy *model.Proxy, sr SecretResour
...
 
+	for _, sr := range resources {
 		if updatedSecrets != nil {
 			if !containsAny(updatedSecrets, relatedConfigs(model.ConfigKey{Kind: gvk.Secret, Name: sr.Name, Namespace: sr.Namespace})) {
 				// This is an incremental update, filter out secrets that are not updated.
 				continue
 			}
 		}
 
-		if err := s.proxyAuthorizedForSecret(proxy, sr); err != nil {
```
stevenctl, line 144: Shouldn't we still do this for the `CrossNamespaceCertificateReferences` case?

howardjohn: We should - and we still do (we have tests as well). This however is dead code, so I removed the function. The new check is where we have `requested secret %v not accessible for proxy %v: SDS is currently only supporting accessing secret within the same namespace. `.

Would be good to have a sanity check that actually makes sense - coming back to this 2 weeks later I think it is but this is critical code

stevenctl (approved):

costinm (approved):
Seems safe...

Issue #21517, opened 2020-02-26
Request for istioctl subcommand to get Envoy version in use on pod


yanniszark:
**Describe the feature request**

Get the envoy version used by the Istio proxy.
Currently, the [latest version of Istio Docs](https://istio.io/docs/ops/diagnostic-tools/proxy-cmd/#what-envoy-version-is-istio-using) instruct the user to do:

```console
$ kubectl exec -it PODNAME -c istio-proxy -n NAMESPACE pilot-agent request GET server_info
```

However, that returns the commit from envoy/proxy.
We want to get the version of the Envoy proxy running, that is a commit on envoyproxy/envoy or istio/envoy.

**Use Case**

Given an issue in envoyproxy/envoy and a PR that fixes it, we want a straightforward way of checking if the PR's commits are inside the Istio proxy that is running.

**Describe alternatives you've considered**

Currently, to do that, we have to go to istio/proxy, open the WORKSPACE file and get the ENVOY_SHA used, as well as Envoy's repo.

There is an old issue: https://github.com/istio/istio/issues/6572
and PR: https://github.com/istio/istio/pull/11524
which are supposed to add this information to labels on the Docker images.
However, it seems that this isn't the case:

```console
$ docker inspect -f '{{json .Config.Labels }}' istio/proxyv2:1.1.6
null

$ docker inspect -f '{{json .Config.Labels }}' istio/proxyv2:1.3.1
{"envoy-vcs-ref":"","istio-api-vcs-ref":""}

$ docker inspect -f '{{json .Config.Labels }}' istio/proxyv2:1.4.5
null
```

[X] Docs
[X] User Experience
[X] Developer Infrastructure

This is mainly @dimara's work, I'm just opening the issue.
/cc @dimara @elevran

esnible:
How about `istioctl proxy-config  bootstrap <pod> | jq .bootstrap.node.hiddenEnvoyDeprecatedBuildVersion` as a work-around?

The output would be

```
"524cfb2be117f343fd25c609bdd622eca7c2b4f5/1.13.1-dev/Clean/RELEASE/BoringSSL"
```

`jq` is a common command line tool for parsing JSON, see https://stedolan.github.io/jq/

esnible:
@yanniszark does this resolve your issue?

yanniszark:
@esnible thanks for your answer.
I will try that out asap.
In any case, I believe we should document this in the docs, because this is a valid use case.
I would propose amending the current section to reflect that.
[...]

PR #33359, opened 2021-06-09
Add version summary option to istioctl proxy-config bootstrap

ryantking:
https://github.com/istio/istio/issues/21517

This PR adds a `short` output type for the `istioctl proxy-config bootstrap` command that shows Istio and Envoy information in an easily human-readable format.

Example:

```
> istioctl proxy-config bootstrap -o short details-v1-79f774bdb9-6hxw6
Istio Version:       1.10.0
Istio Proxy Version: 436f365a8007cd8a13a9f1321e7cce94bcc8883e
Envoy Version:       1.18.3/Clean/RELEASE/BoringSSL
```

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


ryantking:
@googlebot I signed it!

linsun:
LGTM, thank you @ryantking.  I assume this cmd could also work?

```
istioctl proxy-config bootstrap -o short deploy/details
```

@esnible could you PTAL?

linsun:
would be good to create a release note for this in the [release notes dir](https://github.com/istio/istio/tree/master/releasenotes).

ryantking:
@linsun Yes, `istioctl pc bootstrap -o deploy/details-v1` will work all the same. Also added a release note, thanks for looking at it.

releasenotes/notes/33359.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: feature
+area: istioctl
+issue:
+  - i1517
```
shamsher31, line 5:
```suggestion
  - 21517
```

releasenotes/notes/33359.yaml
```diff
@@ -0,0 +1,8 @@
...
+releaseNotes:
+  - |
+    **Added** istioctl proxy-config bootstrap now hoas a short output option (`-o short`) that shows an Istio and Envoy version summary.
```
shamsher31, line 8:
```suggestion
    **Added** `istioctl proxy-config bootstrap` now has a short output option (`-o short`) that shows an Istio and Envoy version summary.
```

istioctl/cmd/proxyconfig.go
```diff
@@ -872,7 +875,7 @@ func bootstrapConfigCmd() *cobra.Command {
 		},
 	}
 
-	bootstrapConfigCmd.Flags().StringVarP(&outputFormat, "output", "o", jsonOutput, "Output format: one of json|yaml")
+	bootstrapConfigCmd.Flags().StringVarP(&outputFormat, "output", "o", jsonOutput, "Output format: one of json|yaml|short")
```
shamsher31, line 878: Can you please also add `Example` [here](https://github.com/istio/istio/blob/d9ab7bb10dbacdc057b49321bf5b88b2128620e8/istioctl/cmd/proxyconfig.go#L834) on how to use this command. This will be helpful for docs purposes.

istioctl/pkg/writer/envoy/configdump/configdump.go
```diff
@@ -133,3 +136,88 @@ func (c *ConfigWriter) PrintFullSummary(cf ClusterFilter, lf ListenerFilter, rf
...
+	return tw.Flush()
+}
+
+func (c *ConfigWriter) getIstioVersionInfo(bootstrapDump *envoy_admin_v3.BootstrapConfigDump) (version, sha string) {
+	const (
+		istioVersionKey  = "ISTIO_VERSION"
+		istioProxyShaKey = "ISTIO_PROXY_SHA"
+	)
+
+	md := bootstrapDump.GetBootstrap().GetNode().GetMetadata().GetFields()
```
shamsher31, line 177: This is neat. 
image

istioctl/pkg/writer/envoy/configdump/testdata/configdump.json
```diff
@@ -0,0 +1,11131 @@
+{
```
howardjohn, line 1: could we not check in 11k lines for this test? I think we only need a very small subset

istioctl/cmd/proxyconfig.go
```diff
@@ -860,19 +863,19 @@ func bootstrapConfigCmd() *cobra.Command {
...
-				// TODO: cobra default value of global flag is passed instead of local flag
-				// i.e short is passed instead of json as declared below
-				// return fmt.Errorf("output format %q not supported", outputFormat)
+				return fmt.Errorf("output format %q not supported", outputFormat)
 			}
 		},
 	}
 
-	bootstrapConfigCmd.Flags().StringVarP(&outputFormat, "output", "o", jsonOutput, "Output format: one of json|yaml")
+	bootstrapConfigCmd.Flags().StringVarP(&outputFormat, "output", "o", jsonOutput, "Output format: one of json|yaml|short")
```
ryantking, line 878: @howardjohn @shamsher31 

The default value here is being overridden by a different command's flag setup since the commands all share a global and use `-o short` by default, which is changing the default behavior to be the new version summary versus JSON causing the istioctl pilot tests to fail.

Do we want `-o short` to be default behavior across the board for the entire command tree, if so, I'll update the tests to explicitly add a `-o json` flag. Otherwise, I'll just shadow the `outputFormat` variable in the function so it respects the json default.

linsun: my 2 cents - probably not good to change the default behavior on this, which I missed earlier while reviewing it.   There likely users building automations on this today and could break them if we change `-o short` to be the default.

ryantking: I tend to agree, the only real issue I see is that output is global to the proxy-config command so we'd be overwriting the default value based on the subcommand. Probably fine, just something worth noting.

ryantking:
@howardjohn Can I get another set of eyes now that CI is passing and the other contents have been addressed?

howardjohn (approved):

linsun:
@shamsher31 can you approve this if you are good with the latest?

linsun (approved):

shamsher31 (approved):
@ryantking  Thanks for working on this.

PR #33361, opened 2021-06-10
Fix cache invalidation issues in EDS

howardjohn:
This was ran 4x in a row with assertions enabled without failures (https://prow.istio.io/pr-history/?org=istio&repo=istio&pr=33361)

Comments inline on why changes are needed

### DIFF
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -187,7 +187,12 @@ func (s *DiscoveryServer) getOrCreateEndpointShard(serviceName, namespace string
 		ServiceAccounts: sets.Set{},
 	}
 	s.EndpointShardsByService[serviceName][namespace] = ep
-
+	// Clear the cache here to avoid race in cache writes (see edsCacheUpdate for details).
+	s.Cache.Clear(map[model.ConfigKey]struct{}{{
+		Kind:      gvk.ServiceEntry,
+		Name:      serviceName,
+		Namespace: namespace,
+	}: {}})
 	return ep, true
 }
 
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -900,6 +900,9 @@ func (ps *PushContext) getExportedDestinationRuleFromNamespace(owningNamespace s
 // IsClusterLocal indicates whether the endpoints for the service should only be accessible to clients
 // within the cluster.
 func (ps *PushContext) IsClusterLocal(service *Service) bool {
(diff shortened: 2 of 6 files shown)


howardjohn:
// TODO: make a Networks subset of all the network stuff, assign it its own version, and use that as a key

stevenctl (approved):

pilot/pkg/xds/eds.go
```diff
@@ -187,7 +187,12 @@ func (s *DiscoveryServer) getOrCreateEndpointShard(serviceName, namespace string
 		ServiceAccounts: sets.Set{},
 	}
 	s.EndpointShardsByService[serviceName][namespace] = ep
-
+	// Clear the cache here to avoid race in cache writes (see edsCacheUpdate for details).
+	s.Cache.Clear(map[model.ConfigKey]struct{}{{
```
hzxuzhonghu, line 191: Still confused at the comment `Essentially, we transition from v0 -> v1 ->
	// v0 -> invalidate -> v1.` 

Isn't this run order by order, and eventually the eds will be cached with v1?

hzxuzhonghu: The reason why i put it here is after this pr, there are lots of places we are clearing the cache, looks not very readable

howardjohn:
```
| Endpoint State | Thread 1            | Thread 2            |
|----------------|---------------------|---------------------|
| v1             |                     |                     |
|                | Start generating v1 |                     |
| v2             |                     |                     |
|                |                     | Start generating v2 |
|                |                     | Write v2            |
|                | Write v1            |                     |
```

^ Is the problem. If we do this all future reads will be of v1, instead of v2. This is a bad bug.

What the token/cache clearing solves is that when when `endpoint state -> v2` occurs, we will invalidate the token from `start generating v1`. Therefor, when the `write v1` occurs, it is rejected. Basically if any state changes between `start` and `write`, the write is ignored.

hzxuzhonghu: got it for this case, but we can not assume that `Start generating v1` happens before v2, so this is what caused my confusion.

howardjohn: If endpoint state is v2, then how do we start generating v1? I consider "endpoint state" to be EndpointShardsByService, and "start generating" when we fetch the EndpointShardsByService. So it should not be possible to "start v1" after "endpoint state is v2"? Not sure if that makes sense, let me know.. this is a bit tricky

PR #33363, opened 2021-06-10
Merge gateway topology test into common suite

howardjohn:
We can now efficiently update mesh config, no need for a new suite

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/integration/pilot/common/routing.go
+++ b/tests/integration/pilot/common/routing.go
@@ -970,7 +970,7 @@ spec:
 	return cases
 }
 
-func XFFGatewayCase(apps *EchoDeployments) []TrafficTestCase {
+func XFFGatewayCase(apps *EchoDeployments, gateway string) []TrafficTestCase {
 	cases := []TrafficTestCase{}
 
 	destinationSets := []echo.Instances{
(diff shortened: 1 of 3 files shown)


ericvn (approved):

jacob-delgado (approved):

tests/integration/pilot/common/routing.go
```diff
@@ -987,12 +987,12 @@ func XFFGatewayCase(apps *EchoDeployments) []TrafficTestCase {
 			name:   d[0].Config().Service,
 			config: httpGateway("*") + httpVirtualService("gateway", fqdn, d[0].Config().PortByName("http").ServicePort),
 			skip:   false,
-			call:   apps.Ingress.CallWithRetryOrFail,
+			call:   apps.Naked[0].CallWithRetryOrFail,
```
stevenctl, line 990: Is calling from Naked truly the same thing as calling from outside the cluster here? K8s won't do anything when it sees the `gateway` address?

stevenctl (approved):

tests/integration/pilot/common/routing.go
howardjohn: Its not literally the same thing. In the context of this test I think its similar enough though, we are testing envoy side behavior that doesn't really care where the request came from anyways. Technically the feature really does care where it comes from, but we are already faking it with explicit XFF header so its not dependent on environment. I figured it was better to reduce LoadBalancer type when not needed

PR #33367, opened 2021-06-10
istioctl proxy-config command support pod-name auto-completion

silenceshell:

When auto-completion is enabled, istioctl could auto complete most command, this is convenient.
However, for `istioctl proxyconfig` command, it wants users to fill in pod-name, but this command could not auto complete pod names in current namespace, so users have to use kubectl to get pods name first, then copy and paste to fill in istioctl command.

`istioctl proxy-config <clusters|listeners|routes|endpoints|bootstrap|log|secret> <pod-name[.namespace]>`

This is a little burdensome. This PR will add func to custom `istioctl proxy-config * ` to auto fill in pods name in current namespace, and will helps a lot.

```bash
$ istioctl proxy-config route <tab><tab>
debian-77dc9c5f4f-nscvz                      nginx-74f5bd857c-7dvkp                       spring-cloud-consul-school-7759b8ffcc-f22q2
$ kubectl get pods
NAME                                          READY   STATUS    RESTARTS   AGE
debian-77dc9c5f4f-nscvz                       2/2     Running   0          2d22h
nginx-74f5bd857c-7dvkp                        2/2     Running   0          2d22h
spring-cloud-consul-school-7759b8ffcc-f22q2   2/2     Running   0          21d
```
---

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[X] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/istioctl/cmd/proxyconfig.go
+++ b/istioctl/cmd/proxyconfig.go
@@ -20,13 +20,13 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
 	"regexp"
 	"strings"
 
 	"github.com/ghodss/yaml"
 	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"istio.io/istio/istioctl/pkg/util/handlers"
 	"istio.io/istio/istioctl/pkg/writer/envoy/clusters"
--- a/istioctl/cmd/proxyconfig.go
+++ b/istioctl/cmd/proxyconfig.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
 	"regexp"
 	"strings"
@@ -340,6 +341,7 @@ func clusterConfigCmd() *cobra.Command {
 				return fmt.Errorf("output format %q not supported", outputFormat)
 			}
 		},
+		ValidArgsFunction: validArgsFunction,
 	}
 
 	clusterConfigCmd.PersistentFlags().StringVarP(&outputFormat, "output", "o", summaryOutput, "Output format: one of json|yaml|short")
@@ -351,6 +353,18 @@ func clusterConfigCmd() *cobra.Command {
 		"Envoy config dump JSON file")
 
 	return clusterConfigCmd
+}
+
+func validArgsFunction(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	if len(args) != 0 {
+		return nil, cobra.ShellCompDirectiveNoFileComp
+	}
+
+	podsName, err := getPodsNameInDefaultNamespace(toComplete)
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveNoFileComp
+	}
+	return podsName, cobra.ShellCompDirectiveNoFileComp
 }
 
 func allConfigCmd() *cobra.Command {
@@ -447,6 +461,7 @@ func allConfigCmd() *cobra.Command {
 			}
 			return nil
 		},
+		ValidArgsFunction: validArgsFunction,
 	}
 
 	allConfigCmd.PersistentFlags().StringVarP(&outputFormat, "output", "o", summaryOutput, "Output format: one of json|yaml|short")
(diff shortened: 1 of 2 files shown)


silenceshell:
hi @shamsher31 
Thank you for your kind reply! Would you like to help to review this PR?

shamsher31 (commented):
@silenceshell thanks for working on this. I really like this feature. 
But like I said this is more like a feature and not an issue. I would like to see this implemented for all the istioctl commands and not just for the `proxy-config`. Can this be more generic to work with other istioctl commands where the pod name is needed?

silenceshell:
We could add those commands to `__istioctl_custom_func`. 
For example, to add support for `istioctl experimental describe pod`, we need add `istioctl_experimental_describe_pod` to cases.

```
__istioctl_custom_func() {
    case ${last_command} in
        istioctl_proxy-config_* | istioctl_experimental_describe_pod)
            __istioctl_get_resource_pod
            return
```

silenceshell:
@shamsher31 Maybe we could add support for other command in new PR? 
Or I could gather commands that need this feature, and update in this PR. I am not sure which way is better :)

marckhouzam:
> I would like to see this implemented for all the istioctl commands and not just for the `proxy-config`.

Be aware that starting with Cobra 1.0, it is possible to code such custom completions in Go instead of bash.
Doing this has the added benefit of being portable to all shells supported by Cobra (bash, zsh, fish, powershell).
It may not be as trivial as calling `kubectl` however, unless it is called from Go code.

Here is the doc if you are interested: https://github.com/spf13/cobra/blob/v1.1.3/shell_completions.md#customizing-completions

silenceshell:
> > I would like to see this implemented for all the istioctl commands and not just for the `proxy-config`.
> 
> Be aware that starting with Cobra 1.0, it is possible to code such custom completions in Go instead of bash.
> Doing this has the added benefit of being portable to all shells supported by Cobra (bash, zsh, fish, powershell).
> It may not be as trivial as calling `kubectl` however, unless it is called from Go code.
> 
> Here is the doc if you are interested: https://github.com/spf13/cobra/blob/v1.1.3/shell_completions.md#customizing-completions

Thanks. I will try to use this method.

silenceshell:
@marckhouzam @shamsher31 
Now implement this feature in Go instead of bash, with better code reading experience.

istioctl/cmd/proxyconfig.go
```diff
@@ -993,3 +1014,25 @@ func getPodNameBySelector(labelSelector string) ([]string, string, error) {
...
 }
+
+func getPodsNameInDefaultNamespace(toComplete string) ([]string, error) {
+	kubeClient, err := kubeClient(kubeconfig, configContext)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx := context.Background()
+	podList, err := kubeClient.CoreV1().Pods(defaultNamespace).List(ctx, metav1.ListOptions{})
```
shamsher31, line 1025: The `getPodName` is [already available](https://github.com/istio/istio/blob/a9d6b12c44557cc1893a54a202dbcabe7a5e7c0a/istioctl/cmd/proxyconfig.go#L965-L978) and you can directly use it here. 

The `proxy-config` can also look into other namespaces and not just the default namespace. That func already handles custom namespace logic.

shamsher31 (commented):
Overall LGTM, other than one comment.

istioctl/cmd/proxyconfig.go
silenceshell: yeah.. This feature is a candy for users to auto completion pods name in current namespace, that's because the command use `<pod-name[.namespace]>` format to specify a pod, that is difficult for completion as the pod-name is before namespace name.
`istioctl proxy-config clusters <pod-name[.namespace]>`

So it is a little heavy to use the `getPodName` func as we only want to get pods in current namespace.

shamsher31: Ok, can you try installing Istio in a custom namespace other than `istio-system` and then run `proxy-config` to try auto-completion for pods?

silenceshell: Thanks for your advice, I will reinstall istio in other namespace in my env later.
However I am not sure what this is to verify for, as the `defaultNamespace` is current-context's namespace specified in kubeconfig, maybe `default`, `istio-demo`, etc,  it has no relationship with `istio-system`.

silenceshell: @shamsher31 
I installed istio in namespace `istio-x`, and proxy-config auto-completion still works fine.

shamsher31 (approved):
Thanks, LGTM.

PR #33369, opened 2021-06-10
make sure daemonset has at least one pod ready

### LINKED ISSUE #32887: istioctl install does not wait for install cni to become ready
istioctl should wait for install-cni container to become ready at installation time.

holooooo:
Fixes #32887 

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


bianpengyuan (commented):
Thanks a lot! 

Since you are here, can you also remove maxUnavailable logic (line 282 - 289)? MaxUnavailable is to control rollout pace, and installer should only consider Daemonset installation done when all pods are ready.

holooooo:
so many failed test 😳 what should I do

bianpengyuan:
Build fails:

```
+ make dockerx.pushx
mkdir -p /home/prow/go/src/istio.io/istio/out/linux_amd64/logs
mkdir -p /home/prow/go/src/istio.io/istio/out/linux_amd64/release
GOOS=linux GOARCH=amd64 LDFLAGS='-extldflags -static -s -w' common/scripts/gobuild.sh /home/prow/go/src/istio.io/istio/out/linux_amd64/ ./istioctl/cmd/istioctl ./pilot/cmd/pilot-discovery ./pkg/test/echo/cmd/client ./pkg/test/echo/cmd/server ./operator/cmd/operator ./cni/cmd/istio-cni ./cni/cmd/istio-cni-repair ./cni/cmd/istio-cni-taint ./cni/cmd/install-cni ./tools/istio-iptables ./tools/bug-report
# istio.io/istio/operator/pkg/helmreconciler
operator/pkg/helmreconciler/wait.go:31:2: imported and not used: "k8s.io/apimachinery/pkg/util/intstr"
real	0m5.838s
user	0m30.728s
sys	0m6.404s
```

bianpengyuan (approved):

operator/pkg/helmreconciler/wait.go
```diff
@@ -279,18 +278,13 @@ func daemonsetsReady(daemonsets []*appsv1.DaemonSet) (bool, []string) {
...
-				maxUnavailable = int(ds.Status.DesiredNumberScheduled)
-			}
-			expectedReady := int(ds.Status.DesiredNumberScheduled) - maxUnavailable
-			if !(int(ds.Status.NumberReady) >= expectedReady) {
+			if ds.Status.NumberReady != ds.Status.DesiredNumberScheduled {
 				scope.Infof("DaemonSet is not ready: %s/%s. %d out of %d expected pods are ready",
-					ds.Namespace, ds.Name, ds.Status.NumberReady, expectedReady)
+					ds.Namespace, ds.Name, ds.Status.NumberReady, ds.Status.UpdatedNumberScheduled)
+				notReady = append(notReady, "DaemonSet/"+ds.Namespace+"/"+ds.Name)
+			} else if ds.Status.DesiredNumberScheduled == 0 {
```
bianpengyuan, line 285: Should this be `ds.Status.NumberReady`?

operator/pkg/helmreconciler/wait.go
```diff
@@ -279,18 +278,13 @@ func daemonsetsReady(daemonsets []*appsv1.DaemonSet) (bool, []string) {
...
-				ds.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable,
-				int(ds.Status.DesiredNumberScheduled),
-				true)
-			if err != nil {
-				// set max unavailable to the number of desired replicas if the value is invalid
-				maxUnavailable = int(ds.Status.DesiredNumberScheduled)
-			}
-			expectedReady := int(ds.Status.DesiredNumberScheduled) - maxUnavailable
-			if !(int(ds.Status.NumberReady) >= expectedReady) {
+			if ds.Status.NumberReady != ds.Status.DesiredNumberScheduled {
```
bianpengyuan, line 281: Can you add a bit comment here about why we have these two cases?

operator/pkg/helmreconciler/wait.go
holooooo: Here is `else if`,  so it's equivalent to `ds.Status.NumberReady`.

I think DesiredNumberScheduled is better than NumberReady in explain the initialization status

operator/pkg/helmreconciler/wait.go
holooooo: sure

morvencao (commented):
If the original issue only happens in 1 node case, I don't think it's a good idea to update the current ready check logic. Adding check for 1 node case should be sufficient.
If you insist, please add testing for install/upgrade cases.

holooooo:
First time apply daemonset, desired num is 0，0-max unavailable less the 0，and also less then ready num.


It means if you get daemon set obj&nbsp; by post response,or get too fast.Old logic will return ready immediately.



 
---Original---
From: "Morven ***@***.***&gt;
Date: 2021/6/11 19:06:32
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [istio/istio] make sure daemonset has at least one pod ready (#33369)




 
@morvencao commented on this pull request.
 
If the original issue only happen in 1 node case, I don't think it's a good idea to update the current ready check logic.
 If you insist, please add testing for install/upgrade cases.

bianpengyuan:
> If the original issue only happens in 1 node case, I don't think it's a good idea to update the current ready check logic. Adding check for 1 node case should be sufficient.
> If you insist, please add testing for install/upgrade cases.

It not only happens on one node. The case I ran into is that at installation time, there is a brief period where (NumberDesired = -1, NumberReady = 0). Current logic will consider this as ready. It looks like @holooooo also sees the case where (NumberDesired = 0, NumberReady = 0).

operator/pkg/helmreconciler/wait.go
bianpengyuan: Can we reverse the condition here?

```
if desired <= 0 {
  log initializing.
} else if ds.Status.NumberReady != ds.Status.DesiredNumberScheduled {
  log some pod is not ready.
}
```

I saw cases where (desired = -1, ready = 0), so this logic make more sense to me.

holooooo:
> > If the original issue only happens in 1 node case, I don't think it's a good idea to update the current ready check logic. Adding check for 1 node case should be sufficient.
> > If you insist, please add testing for install/upgrade cases.
> 
> It not only happens on one node. The case I ran into is that at installation time, there is a brief period where (NumberDesired = -1, NumberReady = 0). Current logic will consider this as ready. It looks like @holooooo also sees the case where (NumberDesired = 0, NumberReady = 0).

I delete expectedReady field, so there will never have -1 case.(controller manager count every available node as NumberDesired, so it also never be negative)

bianpengyuan:
> > > If the original issue only happens in 1 node case, I don't think it's a good idea to update the current ready check logic. Adding check for 1 node case should be sufficient.
> > > If you insist, please add testing for install/upgrade cases.
> > 
> > 
> > It not only happens on one node. The case I ran into is that at installation time, there is a brief period where (NumberDesired = -1, NumberReady = 0). Current logic will consider this as ready. It looks like @holooooo also sees the case where (NumberDesired = 0, NumberReady = 0).
> 
> I delete expectedReady field, so there will never have -1 case.(controller manager count every available node as NumberDesired, so it also never be negative)

Ah I see. Thanks! Could you please take a look at this comment https://github.com/istio/istio/pull/33369#discussion_r650092053?

bianpengyuan (approved):
Thanks!

morvencao (approved):

PR #33374, opened 2021-06-10
Implement Wasm API

dgn:
This is the implementation of the API introduced in https://github.com/istio/api/pull/1940.

What works:
- reading of WasmPlugin CRD, pushing updates through ECDS
- deploying filters hosted on HTTP(S) servers or present locally in the proxy container
- fetching OCI containers (as soon as https://github.com/istio/istio/pull/34430 is merged)
- phase+priority

What's missing:
- image pull secrets
- signature verification

these two will be added later

@mandarjog @douglas-reid @bianpengyuan @jwendell

### DIFF
--- a/pilot/pkg/model/extensions.go
+++ b/pilot/pkg/model/extensions.go
@@ -22,14 +22,14 @@ import (
 	envoy_extensions_filters_http_wasm_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/wasm/v3"
 	envoy_extensions_wasm_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/wasm/v3"
 	"github.com/gogo/protobuf/types"
-	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes/any"
-	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/anypb"
 	"google.golang.org/protobuf/types/known/durationpb"
 
 	extensions "istio.io/api/extensions/v1alpha1"
 	"istio.io/istio/pkg/config"
 	"istio.io/istio/pkg/util/gogo"
+	"istio.io/istio/pkg/util/gogoprotomarshal"
 )
 
 const (
@@ -42,7 +42,7 @@ type WasmPluginWrapper struct {
 	Name      string
 	Namespace string
 
-	ExtensionConfiguration *envoy_extensions_filters_http_wasm_v3.Wasm
+	ExtensionConfiguration *envoy_config_core_v3.TypedExtensionConfig
 }
 
 func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
@@ -54,13 +54,13 @@ func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
 
 	cfg := &any.Any{}
 	if wasmPlugin.PluginConfig != nil && len(wasmPlugin.PluginConfig.Fields) > 0 {
-		cfgJSON, err := protojson.Marshal(proto.MessageV2(wasmPlugin.PluginConfig))
+		cfgJSON, err := gogoprotomarshal.ToJSON(wasmPlugin.PluginConfig)
 		if err != nil {
 			log.Warnf("wasmplugin %v/%v discarded due to json marshaling error: %s", plugin.Namespace, plugin.Name, err)
 			return nil
 		}
 		cfg = gogo.MessageToAny(&types.StringValue{
-			Value: string(cfgJSON),
+			Value: cfgJSON,
 		})
 	}
 	var datasource *envoy_config_core_v3.AsyncDataSource
@@ -108,22 +108,31 @@ func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
 			},
 		}
 	}
-	return &WasmPluginWrapper{
-		Name:       plugin.Name,
-		Namespace:  plugin.Namespace,
-		WasmPlugin: *wasmPlugin,
-		ExtensionConfiguration: &envoy_extensions_filters_http_wasm_v3.Wasm{
-			Config: &envoy_extensions_wasm_v3.PluginConfig{
-				Name:          plugin.Name,
-				RootId:        wasmPlugin.PluginName,
-				Configuration: cfg,
-				Vm: &envoy_extensions_wasm_v3.PluginConfig_VmConfig{
-					VmConfig: &envoy_extensions_wasm_v3.VmConfig{
-						Runtime: defaultRuntime,
-						Code:    datasource,
-					},
+	typedConfig, err := anypb.New(&envoy_extensions_filters_http_wasm_v3.Wasm{
+		Config: &envoy_extensions_wasm_v3.PluginConfig{
+			Name:          plugin.Name,
+			RootId:        wasmPlugin.PluginName,
+			Configuration: cfg,
+			Vm: &envoy_extensions_wasm_v3.PluginConfig_VmConfig{
+				VmConfig: &envoy_extensions_wasm_v3.VmConfig{
+					Runtime: defaultRuntime,
+					Code:    datasource,
 				},
 			},
 		},
+	})
+	if err != nil {
+		log.Warnf("wasmplugin %s/%s failed to marshal to TypedExtensionConfig: %s", plugin.Namespace, plugin.Name, err)
+		return nil
+	}
+	ec := &envoy_config_core_v3.TypedExtensionConfig{
+		Name:        plugin.Name,
+		TypedConfig: typedConfig,
+	}
+	return &WasmPluginWrapper{
+		Name:                   plugin.Name,
+		Namespace:              plugin.Namespace,
+		WasmPlugin:             *wasmPlugin,
+		ExtensionConfiguration: ec,
 	}
 }
--- a/pilot/pkg/model/extensions.go
+++ b/pilot/pkg/model/extensions.go
@@ -15,6 +15,7 @@ package model
 
 import (
 	"net/url"
+	"strings"
 	"time"
 
 	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
@@ -57,6 +58,7 @@ func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
 		var sha256 string
 		u, err := url.Parse(wasmPlugin.Url)
 		if err != nil {
+			log.Warnf("wasmplugin %v/%v discarded due to failure to parse URL: %s", plugin.Namespace, plugin.Name, err)
 			return nil
 		}
 		if wasmPlugin.XSha256 == nil {
@@ -75,8 +77,7 @@ func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
 				Specifier: &envoy_config_core_v3.AsyncDataSource_Local{
 					Local: &envoy_config_core_v3.DataSource{
 						Specifier: &envoy_config_core_v3.DataSource_Filename{
-							// remove 'file://' prefix
-							Filename: wasmPlugin.Url[7:],
+							Filename: strings.TrimPrefix(wasmPlugin.Url, "file://"),
 						},
 					},
 				},
--- a/pilot/pkg/model/extensions.go
+++ b/pilot/pkg/model/extensions.go
@@ -46,77 +46,84 @@ type WasmPluginWrapper struct {
 }
 
 func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
-	if wasmPlugin, ok := plugin.Spec.(*extensions.WasmPlugin); ok {
-		cfg := &any.Any{}
-		if wasmPlugin.PluginConfig != nil && len(wasmPlugin.PluginConfig.Fields) > 0 {
-			cfgJSON, _ := protojson.Marshal(proto.MessageV2(wasmPlugin.PluginConfig))
-			cfg = gogo.MessageToAny(&types.StringValue{
-				Value: string(cfgJSON),
-			})
-		}
-		var datasource *envoy_config_core_v3.AsyncDataSource
-		var sha256 string
-		u, err := url.Parse(wasmPlugin.Url)
+	var ok bool
+	var wasmPlugin *extensions.WasmPlugin
+	if wasmPlugin, ok = plugin.Spec.(*extensions.WasmPlugin); !ok {
+		return nil
+	}
+
+	cfg := &any.Any{}
+	if wasmPlugin.PluginConfig != nil && len(wasmPlugin.PluginConfig.Fields) > 0 {
+		cfgJSON, err := protojson.Marshal(proto.MessageV2(wasmPlugin.PluginConfig))
 		if err != nil {
-			log.Warnf("wasmplugin %v/%v discarded due to failure to parse URL: %s", plugin.Namespace, plugin.Name, err)
+			log.Warnf("wasmplugin %v/%v discarded due to json marshaling error: %s", plugin.Namespace, plugin.Name, err)
 			return nil
 		}
-		if wasmPlugin.XSha256 == nil {
-			// field is required, so we're setting a string for unmarshaling to not fail
-			// on the agent side. this will never reach envoy
-			sha256 = "nil"
-		} else {
-			sha256 = wasmPlugin.GetSha256()
-		}
-		// when no scheme is given, default to oci://
-		if u.Scheme == "" {
-			u.Scheme = "oci"
-		}
-		if u.Scheme == "file" {
-			datasource = &envoy_config_core_v3.AsyncDataSource{
-				Specifier: &envoy_config_core_v3.AsyncDataSource_Local{
-					Local: &envoy_config_core_v3.DataSource{
-						Specifier: &envoy_config_core_v3.DataSource_Filename{
-							Filename: strings.TrimPrefix(wasmPlugin.Url, "file://"),
-						},
-					},
-				},
-			}
-		} else {
-			datasource = &envoy_config_core_v3.AsyncDataSource{
-				Specifier: &envoy_config_core_v3.AsyncDataSource_Remote{
-					Remote: &envoy_config_core_v3.RemoteDataSource{
-						HttpUri: &envoy_config_core_v3.HttpUri{
-							Uri:     u.String(),
-							Timeout: durationpb.New(30 * time.Second),
-							HttpUpstreamType: &envoy_config_core_v3.HttpUri_Cluster{
-								// this will be fetched by the agent anyway, so no need for a cluster
-								Cluster: "_",
-							},
-						},
-						Sha256: sha256,
-					},
-				},
-			}
-		}
-		return &WasmPluginWrapper{
-			Name:       plugin.Name,
-			Namespace:  plugin.Namespace,
-			WasmPlugin: *wasmPlugin,
-			ExtensionConfiguration: &envoy_extensions_filters_http_wasm_v3.Wasm{
-				Config: &envoy_extensions_wasm_v3.PluginConfig{
-					Name:          plugin.Name,
-					RootId:        wasmPlugin.PluginName,
-					Configuration: cfg,
-					Vm: &envoy_extensions_wasm_v3.PluginConfig_VmConfig{
-						VmConfig: &envoy_extensions_wasm_v3.VmConfig{
-							Runtime: defaultRuntime,
-							Code:    datasource,
-						},
+		cfg = gogo.MessageToAny(&types.StringValue{
+			Value: string(cfgJSON),
+		})
+	}
+	var datasource *envoy_config_core_v3.AsyncDataSource
+	var sha256 string
+	u, err := url.Parse(wasmPlugin.Url)
+	if err != nil {
+		log.Warnf("wasmplugin %v/%v discarded due to failure to parse URL: %s", plugin.Namespace, plugin.Name, err)
+		return nil
+	}
+	if wasmPlugin.XSha256 == nil {
+		// field is required, so we're setting a string for unmarshaling to not fail
+		// on the agent side. this will never reach envoy
+		sha256 = "nil"
+	} else {
+		sha256 = wasmPlugin.GetSha256()
+	}
+	// when no scheme is given, default to oci://
+	if u.Scheme == "" {
+		u.Scheme = "oci"
+	}
+	if u.Scheme == "file" {
+		datasource = &envoy_config_core_v3.AsyncDataSource{
+			Specifier: &envoy_config_core_v3.AsyncDataSource_Local{
+				Local: &envoy_config_core_v3.DataSource{
+					Specifier: &envoy_config_core_v3.DataSource_Filename{
+						Filename: strings.TrimPrefix(wasmPlugin.Url, "file://"),
 					},
 				},
 			},
 		}
+	} else {
+		datasource = &envoy_config_core_v3.AsyncDataSource{
+			Specifier: &envoy_config_core_v3.AsyncDataSource_Remote{
+				Remote: &envoy_config_core_v3.RemoteDataSource{
+					HttpUri: &envoy_config_core_v3.HttpUri{
+						Uri:     u.String(),
+						Timeout: durationpb.New(30 * time.Second),
+						HttpUpstreamType: &envoy_config_core_v3.HttpUri_Cluster{
+							// this will be fetched by the agent anyway, so no need for a cluster
+							Cluster: "_",
+						},
+					},
+					Sha256: sha256,
+				},
+			},
+		}
 	}
-	return nil
+	return &WasmPluginWrapper{
+		Name:       plugin.Name,
+		Namespace:  plugin.Namespace,
+		WasmPlugin: *wasmPlugin,
+		ExtensionConfiguration: &envoy_extensions_filters_http_wasm_v3.Wasm{
+			Config: &envoy_extensions_wasm_v3.PluginConfig{
+				Name:          plugin.Name,
+				RootId:        wasmPlugin.PluginName,
+				Configuration: cfg,
+				Vm: &envoy_extensions_wasm_v3.PluginConfig_VmConfig{
+					VmConfig: &envoy_extensions_wasm_v3.VmConfig{
+						Runtime: defaultRuntime,
+						Code:    datasource,
+					},
+				},
+			},
+		},
+	}
 }
--- a/pilot/pkg/model/extensions.go
+++ b/pilot/pkg/model/extensions.go
@@ -62,15 +62,15 @@ func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
 			return nil
 		}
 		if wasmPlugin.XSha256 == nil {
-			if u.Scheme == "http" || u.Scheme == "https" {
-				// SHA256 is required for .wasm deployments fetched from HTTP/HTTPS URLs
-				return nil
-			}
 			// field is required, so we're setting a string for unmarshaling to not fail
 			// on the agent side. this will never reach envoy
 			sha256 = "nil"
 		} else {
 			sha256 = wasmPlugin.GetSha256()
+		}
+		// when no scheme is given, default to oci://
+		if u.Scheme == "" {
+			u.Scheme = "oci"
 		}
 		if u.Scheme == "file" {
 			datasource = &envoy_config_core_v3.AsyncDataSource{
@@ -87,7 +87,7 @@ func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
 				Specifier: &envoy_config_core_v3.AsyncDataSource_Remote{
 					Remote: &envoy_config_core_v3.RemoteDataSource{
 						HttpUri: &envoy_config_core_v3.HttpUri{
-							Uri:     wasmPlugin.Url,
+							Uri:     u.String(),
 							Timeout: durationpb.New(30 * time.Second),
 							HttpUpstreamType: &envoy_config_core_v3.HttpUri_Cluster{
 								// this will be fetched by the agent anyway, so no need for a cluster
--- a/pilot/pkg/model/extensions.go
+++ b/pilot/pkg/model/extensions.go
@@ -1,0 +1,121 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package model
+
+import (
+	"net/url"
+	"time"
+
+	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	envoy_extensions_filters_http_wasm_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/wasm/v3"
+	envoy_extensions_wasm_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/wasm/v3"
+	"github.com/gogo/protobuf/types"
+	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/ptypes/any"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/durationpb"
+
+	extensions "istio.io/api/extensions/v1alpha1"
+	"istio.io/istio/pkg/config"
+	"istio.io/istio/pkg/util/gogo"
+)
+
+const (
+	defaultRuntime = "envoy.wasm.runtime.v8"
+)
+
+type WasmPluginWrapper struct {
+	extensions.WasmPlugin
+
+	Name      string
+	Namespace string
+
+	ExtensionConfiguration *envoy_extensions_filters_http_wasm_v3.Wasm
+}
+
+func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
+	if wasmPlugin, ok := plugin.Spec.(*extensions.WasmPlugin); ok {
+		cfg := &any.Any{}
+		if wasmPlugin.PluginConfig != nil && len(wasmPlugin.PluginConfig.Fields) > 0 {
+			cfgJSON, _ := protojson.Marshal(proto.MessageV2(wasmPlugin.PluginConfig))
+			cfg = gogo.MessageToAny(&types.StringValue{
+				Value: string(cfgJSON),
+			})
+		}
+		var datasource *envoy_config_core_v3.AsyncDataSource
+		var sha256 string
+		u, err := url.Parse(wasmPlugin.Url)
+		if err != nil {
+			return nil
+		}
+		if wasmPlugin.XSha256 == nil {
+			if u.Scheme == "http" || u.Scheme == "https" {
+				// SHA256 is required for .wasm deployments fetched from HTTP/HTTPS URLs
+				return nil
+			}
+			// field is required, so we're setting a string for unmarshaling to not fail
+			// on the agent side. this will never reach envoy
+			sha256 = "nil"
+		} else {
+			sha256 = wasmPlugin.GetSha256()
+		}
+		if u.Scheme == "file" {
+			datasource = &envoy_config_core_v3.AsyncDataSource{
+				Specifier: &envoy_config_core_v3.AsyncDataSource_Local{
+					Local: &envoy_config_core_v3.DataSource{
+						Specifier: &envoy_config_core_v3.DataSource_Filename{
+							// remove 'file://' prefix
+							Filename: wasmPlugin.Url[7:],
+						},
+					},
+				},
+			}
+		} else {
+			datasource = &envoy_config_core_v3.AsyncDataSource{
+				Specifier: &envoy_config_core_v3.AsyncDataSource_Remote{
+					Remote: &envoy_config_core_v3.RemoteDataSource{
+						HttpUri: &envoy_config_core_v3.HttpUri{
+							Uri:     wasmPlugin.Url,
+							Timeout: durationpb.New(30 * time.Second),
+							HttpUpstreamType: &envoy_config_core_v3.HttpUri_Cluster{
+								// this will be fetched by the agent anyway, so no need for a cluster
+								Cluster: "_",
+							},
+						},
+						Sha256: sha256,
+					},
+				},
+			}
+		}
+		return &WasmPluginWrapper{
+			Name:       plugin.Name,
+			Namespace:  plugin.Namespace,
+			WasmPlugin: *wasmPlugin,
+			ExtensionConfiguration: &envoy_extensions_filters_http_wasm_v3.Wasm{
+				Config: &envoy_extensions_wasm_v3.PluginConfig{
+					Name:          plugin.Name,
+					RootId:        wasmPlugin.PluginName,
+					Configuration: cfg,
+					Vm: &envoy_extensions_wasm_v3.PluginConfig_VmConfig{
+						VmConfig: &envoy_extensions_wasm_v3.VmConfig{
+							Runtime: defaultRuntime,
+							Code:    datasource,
+						},
+					},
+				},
+			},
+		}
+	}
+	return nil
+}
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -1720,9 +1720,6 @@ func (ps *PushContext) initWasmPlugins(env *Environment) error {
 	sortConfigByCreationTime(wasmplugins)
 	ps.wasmPluginsByNamespace = map[string][]*WasmPluginWrapper{}
 	for _, plugin := range wasmplugins {
-		if ps.wasmPluginsByNamespace[plugin.Namespace] == nil {
-			ps.wasmPluginsByNamespace[plugin.Namespace] = []*WasmPluginWrapper{}
-		}
 		if pluginWrapper := convertToWasmPluginWrapper(&plugin); pluginWrapper != nil {
 			ps.wasmPluginsByNamespace[plugin.Namespace] = append(ps.wasmPluginsByNamespace[plugin.Namespace], pluginWrapper)
 		}
@@ -1736,6 +1733,10 @@ func (ps *PushContext) WasmPlugins(proxy *Proxy) map[extensions.PluginPhase][]*W
 	if proxy == nil {
 		return nil
 	}
+	var workloadLabels labels.Collection
+	if proxy.Metadata != nil && len(proxy.Metadata.Labels) > 0 {
+		workloadLabels = labels.Collection{proxy.Metadata.Labels}
+	}
 	matchedPlugins := make(map[extensions.PluginPhase][]*WasmPluginWrapper)
 	// First get all the extension configs from the config root namespace
 	// and then add the ones from proxy's own namespace
@@ -1743,10 +1744,6 @@ func (ps *PushContext) WasmPlugins(proxy *Proxy) map[extensions.PluginPhase][]*W
 		// if there is no workload selector, the config applies to all workloads
 		// if there is a workload selector, check for matching workload labels
 		for _, plugin := range ps.wasmPluginsByNamespace[ps.Mesh.RootNamespace] {
-			var workloadLabels labels.Collection
-			if proxy.Metadata != nil && len(proxy.Metadata.Labels) > 0 {
-				workloadLabels = labels.Collection{proxy.Metadata.Labels}
-			}
 			if plugin.Selector == nil || workloadLabels.IsSupersetOf(plugin.Selector.MatchLabels) {
 				matchedPlugins[plugin.Phase] = append(matchedPlugins[plugin.Phase], plugin)
 			}
@@ -1756,10 +1753,6 @@ func (ps *PushContext) WasmPlugins(proxy *Proxy) map[extensions.PluginPhase][]*W
 	// To prevent duplicate extensions in case root namespace equals proxy's namespace
 	if proxy.ConfigNamespace != ps.Mesh.RootNamespace {
 		for _, plugin := range ps.wasmPluginsByNamespace[proxy.ConfigNamespace] {
-			var workloadLabels labels.Collection
-			if proxy.Metadata != nil && len(proxy.Metadata.Labels) > 0 {
-				workloadLabels = labels.Collection{proxy.Metadata.Labels}
-			}
 			if plugin.Selector == nil || workloadLabels.IsSupersetOf(plugin.Selector.MatchLabels) {
 				matchedPlugins[plugin.Phase] = append(matchedPlugins[plugin.Phase], plugin)
 			}
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -17,6 +17,7 @@ package model
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"sort"
 	"strconv"
 	"strings"
@@ -25,6 +26,7 @@ import (
 
 	"go.uber.org/atomic"
 
+	extensions "istio.io/api/extensions/v1alpha1"
 	meshconfig "istio.io/api/mesh/v1alpha1"
 	networking "istio.io/api/networking/v1alpha3"
 	"istio.io/istio/pilot/pkg/features"
@@ -203,6 +205,9 @@ type PushContext struct {
 
 	// envoy filters for each namespace including global config namespace
 	envoyFiltersByNamespace map[string][]*EnvoyFilterWrapper
+
+	// wasm plugins for each namespace including global config namespace
+	wasmPluginsByNamespace map[string][]*WasmPluginWrapper
 
 	// AuthnPolicies contains Authn policies by namespace.
 	AuthnPolicies *AuthenticationPolicies `json:"-"`
@@ -1117,6 +1122,10 @@ func (ps *PushContext) createNewContext(env *Environment) error {
 		return err
 	}
 
+	if err := ps.initWasmPlugins(env); err != nil {
+		return err
+	}
+
 	if err := ps.initEnvoyFilters(env); err != nil {
 		return err
 	}
@@ -1137,7 +1146,8 @@ func (ps *PushContext) updateContext(
 	oldPushContext *PushContext,
 	pushReq *PushRequest) error {
 	var servicesChanged, virtualServicesChanged, destinationRulesChanged, gatewayChanged,
-		authnChanged, authzChanged, envoyFiltersChanged, sidecarsChanged, telemetryChanged, gatewayAPIChanged bool
+		authnChanged, authzChanged, envoyFiltersChanged, sidecarsChanged, telemetryChanged, gatewayAPIChanged,
+		wasmPluginsChanged bool
 
 	for conf := range pushReq.ConfigsUpdated {
 		switch conf.Kind {
@@ -1151,6 +1161,8 @@ func (ps *PushContext) updateContext(
 			gatewayChanged = true
 		case gvk.Sidecar:
 			sidecarsChanged = true
+		case gvk.WasmPlugin:
+			wasmPluginsChanged = true
 		case gvk.EnvoyFilter:
 			envoyFiltersChanged = true
 		case gvk.AuthorizationPolicy:
@@ -1225,6 +1237,14 @@ func (ps *PushContext) updateContext(
 		}
 	} else {
 		ps.Telemetry = oldPushContext.Telemetry
+	}
+
+	if wasmPluginsChanged {
+		if err := ps.initWasmPlugins(env); err != nil {
+			return err
+		}
+	} else {
+		ps.wasmPluginsByNamespace = oldPushContext.wasmPluginsByNamespace
 	}
 
 	if envoyFiltersChanged {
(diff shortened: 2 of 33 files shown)


pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,104 @@
...
+
+	Name      string
+	Namespace string
+
+	ExtensionConfiguration *envoy_extensions_filters_http_wasm_v3.Wasm
+}
+
+func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
+	if wasmPlugin, ok := plugin.Spec.(*extensions.WasmPlugin); ok {
+		cfg := &any.Any{}
```
howardjohn, line 49: This feels a bit awkward to me, although it may be correct. We take a Struct, covert it to a json string, then create a StringValue then wrap it in Any?

In the API discussion we agreed on JSON but I expected this to be:  We take a Struct, then wrap it in Any. Does that not end up the same/equivalent inside the WASM VM?

Just not sure the difference between the two on the VM side. If the struct ends up in the wrong format then I guess this is the way to go

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,104 @@
...
+		if wasmPlugin.PluginConfig != nil && len(wasmPlugin.PluginConfig.Fields) > 0 {
+			cfgJSON, _ := protojson.Marshal(proto.MessageV2(wasmPlugin.PluginConfig))
+			cfg = gogo.MessageToAny(&types.StringValue{
+				Value: string(cfgJSON),
+			})
+		}
+		var datasource *envoy_config_core_v3.AsyncDataSource
+		if strings.HasPrefix(wasmPlugin.Url, "http") {
```
howardjohn, line 57: is https allowed?

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,104 @@
...
+			datasource = &envoy_config_core_v3.AsyncDataSource{
+				Specifier: &envoy_config_core_v3.AsyncDataSource_Remote{
+					Remote: &envoy_config_core_v3.RemoteDataSource{
+						HttpUri: &envoy_config_core_v3.HttpUri{
+							Uri:     wasmPlugin.Url,
+							Timeout: durationpb.New(10 * time.Second),
```
howardjohn, line 63: Isn't this sort of low for a large pull?

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,104 @@
...
+							HttpUpstreamType: &envoy_config_core_v3.HttpUri_Cluster{
+								// this will be fetched by the agent anyway, so no need for a cluster
+								Cluster: "idontexist",
```
howardjohn, line 66: any negative impact of a fake cluster here? Feels a bit odd.. In theory we may not use the agent in the future

howardjohn: But probably not a high priority

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,193 @@
...
+		// copy extensions map
+		extensions := make(map[networking.EnvoyFilter_Patch_FilterClass][]*model.WasmPluginWrapper)
+		for phase, list := range pluginsMap {
+			extensions[phase] = []*model.WasmPluginWrapper{}
+			extensions[phase] = append(extensions[phase], list...)
+		}
+		hcm := &hcm_filter.HttpConnectionManager{}
+		hcmIndex := -1
+		for i, f := range fc.Filters {
+			if f.Name == wellknown.HTTPConnectionManager && f.GetTypedConfig() != nil {
```
howardjohn, line 92: This is fine for a POC but for the real implementation we need to hook deeper into the LDS generation to avoid double marshalling

howardjohn (commented):
Nice!

pilot/pkg/model/extensions.go
dgn: Yeah tbh I was hoping it would have been easier. I spent hours just on this part. The docs say if you put in a Struct it outputs JSON, but doing that didn't work, at least with the extension I tested with.. This was the only solution I got working. There probably is a simpler/better way though

pilot/pkg/model/extensions.go
dgn: I think technically the agent supports it, that's why I used `HasPrefix` - didn't have a chance to test it yet though. For a real solution this will need to actually parse the url properly and look at the scheme, as we want to support docker:// as well (and probably even default to that if there's no scheme)

pilot/pkg/model/extensions.go
dgn: Yeah this is just here because the field needs to be present, if not the agent will refuse to unmarshal it. Empty string works too I guess

pilot/pkg/model/extensions.go
dgn: increased to 30s

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
dgn: I moved it to right after plugin gen now

pkg/config/validation/validation.go
```diff
@@ -3458,3 +3459,19 @@ func validateTelemetryProviders(providers []*telemetry.ProviderRef) error {
...
+	func(cfg config.Config) (Warning, error) {
+		spec, ok := cfg.Spec.(*extensions.WasmPlugin)
+		if !ok {
+			return nil, fmt.Errorf("cannot cast to service entry")
+		}
+
+		errs := Validation{}
+
+		errs = appendValidation(errs,
+			validateWorkloadSelector(spec.Selector),
```
howardjohn, line 3474: Is this the only validation we need? Its much easier to remove validation than to add it later

pkg/wasm/cache.go
```diff
@@ -136,7 +136,8 @@ func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration
 		// Get sha256 checksum and check if it is the same as provided one.
 		sha := sha256.Sum256(b)
 		dChecksum = hex.EncodeToString(sha[:])
-		if checksum != "" && dChecksum != checksum {
+		// checksum sent by istiod can be "nil" if not set by user - magic value used to avoid unmarshaling errors
+		if checksum != "" && checksum != "nil" && dChecksum != checksum {
```
mathetake, line 140: OCI also validates the checksum if it is not empty: https://github.com/istio/istio/blob/master/pkg/wasm/imagefetcher.go#L83

so maybe better we handle this before entering this switch statement.

dgn: thanks, good point. I'm moving it to the conversion code

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,114 @@
...
+			})
+		}
+		var datasource *envoy_config_core_v3.AsyncDataSource
+		var sha256 string
+		if wasmPlugin.XSha256 == nil {
+			if strings.HasPrefix(wasmPlugin.Url, "http") {
+				// SHA256 is required for .wasm deployments fetched from HTTP/HTTPS URLs
+				return nil
+			}
+			sha256 = "nil"
```
shamsher31, line 63: `sha256` should get value before return? or better to just assign default value as "nil"? 
```suggestion
		if wasmPlugin.XSha256 == nil {
			if strings.HasPrefix(wasmPlugin.Url, "http") {
				// SHA256 is required for .wasm deployments fetched from HTTP/HTTPS URLs
				sha256 = "nil"
				return nil
			}
```

releasenotes/notes/wasmplugin.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: feature
+area: telemetry
+issue: []
+releaseNotes:
+  - |
+    **Added** Support for Istio WasmPlugin API
```
shamsher31, line 7: nit:
```suggestion
    **Added** support for Istio WasmPlugin API.
```

pilot/pkg/model/extensions.go
dgn: no, for http[s] URLs not having a sha256 value is invalid. for OCI it's optional

pkg/config/validation/validation.go
dgn: I added some more validation

pkg/config/validation/validation.go
```diff
@@ -3458,3 +3460,57 @@ func validateTelemetryProviders(providers []*telemetry.ProviderRef) error {
...
+		)
+		return errs.Unwrap()
+	})
+
+func validateWasmPluginURL(pluginURL string) error {
+	if pluginURL == "" {
+		return fmt.Errorf("url field needs to be set")
+	}
+	validSchemes := map[string]bool{
+		"": true, "file": true, "http": true, "https": true, "oci": true,
```
mathetake, line 3486: I imagine if "file" is given, we would hit https://github.com/istio/istio/blob/master/pkg/wasm/cache.go#L159?

dgn: no- because if the user sets a `file://` URL, we'll generate a `local` datasource for the HTTP filter, which the wasm conversion code will ignore (it only converts `remote` datasources)

dgn: see https://github.com/istio/istio/blob/74d1c06903ded69df9f090b1e3fe88fd8686804d/pilot/pkg/model/extensions.go#L73-L83

mathetake: oh I see! Thanks!

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,121 @@
...
+		if wasmPlugin.PluginConfig != nil && len(wasmPlugin.PluginConfig.Fields) > 0 {
+			cfgJSON, _ := protojson.Marshal(proto.MessageV2(wasmPlugin.PluginConfig))
+			cfg = gogo.MessageToAny(&types.StringValue{
+				Value: string(cfgJSON),
+			})
+		}
+		var datasource *envoy_config_core_v3.AsyncDataSource
+		var sha256 string
+		u, err := url.Parse(wasmPlugin.Url)
+		if err != nil {
```
bianpengyuan, line 59: Should have a error log here. There is no log either after returning to the caller.

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,121 @@
...
+			return nil
+		}
+		if wasmPlugin.XSha256 == nil {
+			if u.Scheme == "http" || u.Scheme == "https" {
+				// SHA256 is required for .wasm deployments fetched from HTTP/HTTPS URLs
```
bianpengyuan, line 64: Though this is not true with xds agent. We intercept this in wasm cache, even it does not have sha256, so I think we should be fine to not special case it here.

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,121 @@
...
+		} else {
+			sha256 = wasmPlugin.GetSha256()
+		}
+		if u.Scheme == "file" {
+			datasource = &envoy_config_core_v3.AsyncDataSource{
+				Specifier: &envoy_config_core_v3.AsyncDataSource_Local{
+					Local: &envoy_config_core_v3.DataSource{
+						Specifier: &envoy_config_core_v3.DataSource_Filename{
+							// remove 'file://' prefix
+							Filename: wasmPlugin.Url[7:],
```
bianpengyuan, line 79: nit: can we strip the prefix explicitly here?

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+		}
+		return &WasmPluginWrapper{
+			Name:       plugin.Name,
+			Namespace:  plugin.Namespace,
+			WasmPlugin: *wasmPlugin,
+			ExtensionConfiguration: &envoy_extensions_filters_http_wasm_v3.Wasm{
+				Config: &envoy_extensions_wasm_v3.PluginConfig{
+					Name:          plugin.Name,
+					RootId:        wasmPlugin.PluginName,
```
bianpengyuan, line 108: Why do we set root id here? This will need to be compiled into the Wasm code. We should leave it unset.

pilot/pkg/model/push_context.go
```diff
@@ -1594,6 +1614,88 @@ func (ps *PushContext) initTelemetry(env *Environment) (err error) {
...
+			if prio := slice[i].Priority; prio != nil {
+				iPriority = prio.Value
+			}
+			jPriority := int64(math.MinInt64)
+			if prio := slice[j].Priority; prio != nil {
+				jPriority = prio.Value
+			}
+			// if priority is the same, in order to still have a
+			// deterministic ordering, we sort based on name + url
+			if iPriority == jPriority {
```
bianpengyuan, line 1686: Don't we already sort deterministically based on creation time?

bianpengyuan:
We probably need an integration test before checking this in. We already have an integration test for EnvoyFilter based Wasm module distribution. We can convert that one to use WasmPlugin API.

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,149 @@
...
+			newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_AUTHZ)
+			newHTTPFilters = append(newHTTPFilters, httpFilter)
+		default:
+			newHTTPFilters = append(newHTTPFilters, httpFilter)
+		}
+	}
+	// append all remaining extensions at the end (router is not yet in the chain so this is correct)
+	newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_AUTHN)
+	newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_AUTHZ)
+	// TODO: stats are currently injected using EnvoyFilter, we should migrate them to
```
bianpengyuan, line 90: This TODO is not right. We are implementing those filters as first class as part of efforts for telemetry api.

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,149 @@
...
+
+func injectExtensions(filterChain []*hcm_filter.HttpFilter, exts map[extensions.PluginPhase][]*model.WasmPluginWrapper) []*hcm_filter.HttpFilter {
+	// copy map as we'll manipulate it in the loop
+	extMap := make(map[extensions.PluginPhase][]*model.WasmPluginWrapper)
+	for phase, list := range exts {
+		extMap[phase] = []*model.WasmPluginWrapper{}
+		extMap[phase] = append(extMap[phase], list...)
+	}
+	newHTTPFilters := make([]*hcm_filter.HttpFilter, 0)
+	for _, httpFilter := range filterChain {
```
bianpengyuan, line 71: This part of logic feels a bit hard for me to follow.. Can we go with something simple like:

```go
// construct a first class extation map based on name.
firstClassExts := map[string]*hcm_filter.HttpFilter
for _, f := range filterChain {
  firstClassExts[f.Name] = f
}

// Construct a new ext list based on phase.
newHTTPFilters := make([]*hcm_filter.HttpFilter, 0)
newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_AUTHN)
if f, ok := firstClassExts[securitymodel.EnvoyJwtFilterName]; ok {
  newHTTPFilters = append(newHTTPFilters, f)
}
if f, ok := firstClassExts[securitymodel.AuthnFilterName]; ok {
  newHTTPFilters = append(newHTTPFilters, f)
}
newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_AUTHZ)
if f, ok := firstClassExts[firstClassExts[securitymodel.RBACFilterName]; ok {
  newHTTPFilters = append(newHTTPFilters, f)
}
newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_STATS)
```

Not sure if anything would make this logic not work, but this feels easier to follow and has less duplications.

releasenotes/notes/wasmplugin.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: feature
+area: telemetry
```
bianpengyuan, line 3: Should be `extensibility`.

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
dgn: I didn't know that. Can you point me to relevant changes? Are they not injected using EnvoyFilter anymore?

bianpengyuan: https://github.com/istio/istio/pull/33583

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
bianpengyuan: I think we know all the first class plugins? They are a limited set and can be enumerated here.

pilot/pkg/model/extensions.go
mathetake: iiuc, root id should also be set here, not only in a Wasm binary. this is only used by C++ and AssemblyScript SD, and rootId here is be used to select one of plugin instances associated with rootID inside these SDK. https://github.com/istio/api/blob/master/extensions/v1alpha1/wasm.proto#L229-L232

pilot/pkg/model/push_context.go
dgn: oh, yes that's absolutely right. thanks

pilot/pkg/model/extensions.go
bianpengyuan: I suppose most user do not have multiple roots in one module, so we should just leave it empty here? We don't expose root id in Istio Wasm API, because we don't think people are using this.

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
dgn: that's right- but what I mean is that this is incompatible with pilot's plugin system. There could be any number of filters in the chain that we don't know about depending on the plugins being used

pilot/pkg/model/extensions.go
dgn: should we really be more loose than envoy here and allow using wasm modules without the sha?

bianpengyuan: Yeah it is to make testing easy. We just need to be very explicit that SHA is highly recommended to avoid repeated fetches.

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
bianpengyuan: Not sure I follow. What is the incompatibility? Can you provide an example?

releasenotes/notes/wasmplugin.yaml
dgn: fixed, thanks!

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,150 @@
...
+		switch httpFilter.Name {
+		case securitymodel.EnvoyJwtFilterName:
+			newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_AUTHN)
+			newHTTPFilters = append(newHTTPFilters, httpFilter)
+		case securitymodel.AuthnFilterName:
+			newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_AUTHN)
+			newHTTPFilters = append(newHTTPFilters, httpFilter)
+		case authzmodel.RBACHTTPFilterName:
```
yangminzhu, line 79: I think this won't work with the CUSTOM and AUDIT action in the authorization policy.

In authorization policy, we have a special CUSTOM action that will add an extra RBAC filter (shadow mode, not used for enforcing but only for matching the request) in the beginning of the filter chain. So the order in the filter chain is:

1. `RBAC filter` shadow mode to match requests for CUSTOM action
2. `ext-authz filter` enforcement of CUSTOM action, triggered based on match results
3. `JWT filter`
4. `istio-authn filter`
5. `RBAC filter` enforcement mode for ALLOW and CUSTOM action

Today, the CUSTOM action does not allow matching on attributes requiring authentication (e.g. mTLS or JWT) because the RBAC shadow mode filter is added before the JWT/authn filter. 

In the future, we will allow users to configure the order of the CUSTOM action. Some users just need it in the beginning of the filter chain (current behavior, useful for use cases like token exchange), other users may choose to put it after authn filter.

The same issue also applies to the AUDIT action where we also reuses the RBAC filter in shadow mode for matching requests instead of doing authorization.

We do not have this WasmPlugin API when we first added the CUSTOM action, it's an internal implementation about which Envoy filter should be used for what purposes, in other words, we do not mandate that jwt filter can only used for authn and RBAC filter can only be used for authz.

The above example of CUSTOM action essentially means authZ may not always happen after the authN. It could be a 2-level authN/Z system where the 1st level is an external authN/Z check with token exchange and the 2nd level is a local authN/Z check on the exchanged token.

In the short term, maybe the API should clarify the definition of the `phase` or add to support for `LOCAL_AUTHN`, `LOCAL_AUTHZ` and `EXTERNAL_AUTH`?

In the long term, I feel this should really align and merge with the ExtensionProvider in the mesh config, maybe something like:

```protobuf
message ExtensionPlugin {
  WorkloadSelector selector = 1;
  PluginPhase phase = 2;
  google.protobuf.Int64Value priority = 3;

  oneof extension_provider {
    WasmPlugin wasm = 4;
    ExternalAuthz ext_authz = 5;
    ZipkinTracing zipkin = 6;
    Stackdriver stackdriver = 7;
    // More extensions to be added.
  }
  // Alternative of using oneof field: repeated ExtensionProvider extension_providers = 4;
}

```

cc @mandarjog @louiscryan @smawson @howardjohn

bianpengyuan: > In the short term, maybe the API should clarify the definition of the phase or add to support for LOCAL_AUTHN, LOCAL_AUTHZ and EXTERNAL_AUTH?

Feels that it might be confusing for Wasm API user to have such detailed phase definition. I would expect most ext authz user won't overlap with Wasm user. We should definitely update the implementation for CUSTOM. For simplicity, maybe we can treat AUTHZ and AUTHN in Wasm API as local ones, and do not consider CUSTOM in Wasm plugin ordering for now (i.e. keep the current behavior to put them at the front of the filter chain), and gather feedback if EXT_AUTH phase is really necessary for Wasm API.

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,150 @@
...
+			continue
+		}
+		mutable.FilterChains[fcIndex].HTTP = injectExtensions(fc.HTTP, extensionsMap)
+	}
+}
+
+func injectExtensions(filterChain []*hcm_filter.HttpFilter, exts map[extensions.PluginPhase][]*model.WasmPluginWrapper) []*hcm_filter.HttpFilter {
+	// copy map as we'll manipulate it in the loop
+	extMap := make(map[extensions.PluginPhase][]*model.WasmPluginWrapper)
+	for phase, list := range exts {
```
bianpengyuan, line 66: Should we define priorities for built in filters, so that Wasm filters can be injected after the built in ones, not always before them?

mathetake:
@dgn @bianpengyuan Can I ask what's the blocker here?

bianpengyuan:
I don't think there is any blocker according to last Extension & Telemetry work group meeting. @dgn we should really try to get this in this week. 1.12 cut will happen at 10/14. If we delay more, Wasm API will likely slide for another release...

pkg/config/validation/validation.go
```diff
@@ -3477,3 +3479,57 @@ func validateTelemetryProviders(providers []*telemetry.ProviderRef) error {
...
+		}
+		return nil
+	}
+	if len(plugin.GetSha256()) != 64 {
+		return fmt.Errorf("sha256 field must be 64 characters long")
+	}
+	validCharacters := "0123456789abcdef"
+	for _, char := range plugin.GetSha256() {
+		if !strings.ContainsRune(validCharacters, char) {
+			return fmt.Errorf("sha256 field must be alphanumeric: %s", plugin.GetSha256())
```
mathetake, line 3531: alphanumeric sometimes includes capital letters (A-Z) so maybe we should be more explicit in this comment: like "must match [a-z0-9] pattern"

mathetake (commented):
overall looks good to me, thank you!

pilot/pkg/model/extensions.go
dgn: alright, removing this check

pkg/config/validation/validation_test.go
```diff
@@ -6676,3 +6677,71 @@ func TestValidateTelemetry(t *testing.T) {
...
+			&extensions.WasmPlugin{
+				Url: "http://test.com/test",
+				XSha256: &extensions.WasmPlugin_Sha256{
+					Sha256: "test",
+				},
+			},
+			"sha256 field must be 64 characters long", "",
+		},
+		{
+			"valid oci",
```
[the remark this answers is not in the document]
dgn: no scheme will be interpreted as `oci://`, see https://github.com/istio/istio/blob/742c86b736a9b50a048e3be473cffa0e6d755619/pilot/pkg/model/extensions.go#L71-L74

pkg/config/validation/validation_test.go
```diff
@@ -6676,3 +6677,71 @@ func TestValidateTelemetry(t *testing.T) {
...
+			&extensions.WasmPlugin{
+				Url: "test.com/test",
```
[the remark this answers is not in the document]
dgn: see above

pkg/config/validation/validation.go
```diff
@@ -3477,3 +3479,57 @@ func validateTelemetryProviders(providers []*telemetry.ProviderRef) error {
...
+			return fmt.Errorf("sha256 field must be set when fetching .wasm module from HTTP(s) URL")
+		}
+		return nil
+	}
+	if len(plugin.GetSha256()) != 64 {
+		return fmt.Errorf("sha256 field must be 64 characters long")
+	}
+	validCharacters := "0123456789abcdef"
+	for _, char := range plugin.GetSha256() {
+		if !strings.ContainsRune(validCharacters, char) {
```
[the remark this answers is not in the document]
dgn: I'm now matching against `[a-z0-9]{64}` using a regexp, as that is also what the error states. it'll be slightly slower than looking at the runes themselves, but more readable and the performance should not be in any way problematic (on my machine, 1000 checks took ~80ms, and 1000 WasmPlugins is a lot)

pkg/config/validation/validation.go
dgn: error message has been adjusted

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
dgn: injecting after a builtin filter can also be done my using the next phase, with a high priority. do you think that's enough?

pkg/config/validation/validation.go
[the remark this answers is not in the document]
dgn: makes sense

dgn:
This is ready for reviews now. @mandarjog @douglas-reid @bianpengyuan @jwendell @mathetake PTAL

pkg/config/validation/validation.go
```diff
@@ -3477,3 +3479,53 @@ func validateTelemetryProviders(providers []*telemetry.ProviderRef) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to parse url: %s", err)
+	}
+	if _, found := validSchemes[u.Scheme]; !found {
+		return fmt.Errorf("url contains unsupported scheme: %s", u.Scheme)
+	}
+	return nil
+}
+
+func validateWasmPluginSHA(plugin *extensions.WasmPlugin) error {
```
mandarjog, line 3518: What happens when sha verification fails?
We are not enforcing it here because the validator does not do call outs.
How does the user know? 
@bianpengyuan  cc

mandarjog (commented):
Thanks @dgn 
over all looks good.

pilot/pkg/model/push_context.go
```diff
@@ -1690,6 +1710,81 @@ func (ps *PushContext) initTelemetry(env *Environment) (err error) {
...
+	// First get all the extension configs from the config root namespace
+	// and then add the ones from proxy's own namespace
+	if ps.Mesh.RootNamespace != "" {
+		// if there is no workload selector, the config applies to all workloads
+		// if there is a workload selector, check for matching workload labels
+		for _, plugin := range ps.wasmPluginsByNamespace[ps.Mesh.RootNamespace] {
+			var workloadLabels labels.Collection
+			if proxy.Metadata != nil && len(proxy.Metadata.Labels) > 0 {
+				workloadLabels = labels.Collection{proxy.Metadata.Labels}
+			}
```
mathetake, line 1749: I think we don't need to repeat this inside the loop - looks independent of `plugin` loop var.

pilot/pkg/model/push_context.go
```diff
@@ -1690,6 +1710,81 @@ func (ps *PushContext) initTelemetry(env *Environment) (err error) {
...
+		}
+	}
+
+	// To prevent duplicate extensions in case root namespace equals proxy's namespace
+	if proxy.ConfigNamespace != ps.Mesh.RootNamespace {
+		for _, plugin := range ps.wasmPluginsByNamespace[proxy.ConfigNamespace] {
+			var workloadLabels labels.Collection
+			if proxy.Metadata != nil && len(proxy.Metadata.Labels) > 0 {
+				workloadLabels = labels.Collection{proxy.Metadata.Labels}
+			}
```
mathetake, line 1762: same here

pkg/config/validation/validation.go
```diff
@@ -3477,3 +3479,53 @@ func validateTelemetryProviders(providers []*telemetry.ProviderRef) error {
 	}
 	return nil
 }
+
+// ValidateWasmPlugin validates a WasmPlugin.
+var ValidateWasmPlugin = registerValidateFunc("ValidateWasmPlugin",
+	func(cfg config.Config) (Warning, error) {
```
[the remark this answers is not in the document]
dgn: nope- it's freeform, and already typed to `*types.Struct` - any such struct should be convertable to json

pilot/pkg/model/push_context.go
dgn: agreed. this was copied from EnvoyFilter, didn't refactor it. fixed this now

pilot/pkg/model/push_context.go
dgn: fixed, see above

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,174 @@
...
+		hasName[n] = true
+	}
+	for _, list := range wasmPlugins {
+		for _, p := range list {
+			typedConfig, _ := anypb.New(p.ExtensionConfiguration)
+			ec := &envoy_config_core_v3.TypedExtensionConfig{
+				Name:        p.Name,
+				TypedConfig: typedConfig,
+			}
+			if _, ok := hasName[ec.GetName()]; ok {
```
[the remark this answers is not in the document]
dgn: done

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,174 @@
...
+		ConfigType: &hcm_filter.HttpFilter_ConfigDiscovery{
+			ConfigDiscovery: &envoy_config_core_v3.ExtensionConfigSource{
+				ConfigSource: defaultConfigSource,
+				TypeUrls:     []string{"type.googleapis.com/" + wasmFilterType},
+			},
+		},
+	}
+}
+
+// InsertedExtensionConfigurations returns extension configurations added via EnvoyFilter.
```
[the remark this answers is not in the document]
dgn: I coped this from EnvoyFilter and forgot to fix the comment. done

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,174 @@
...
+	newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_STATS)
+	newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_UNSPECIFIED_PHASE)
+	return newHTTPFilters
+}
+
+func popAppend(list []*hcm_filter.HttpFilter,
+	filterMap map[extensions.PluginPhase][]*model.WasmPluginWrapper,
+	phase extensions.PluginPhase) []*hcm_filter.HttpFilter {
+	for _, ext := range filterMap[phase] {
+		if filter := toEnvoyHTTPFilter(ext); filter != nil {
```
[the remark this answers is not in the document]
dgn: fixed

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,174 @@
...
+	// The one assumption we make is about the ordering of the builtin filters. This is used to
+	// position WasmPlugins relatively to the builtin filters according to their phase: when
+	// we see the Stats filter, we know that all WasmPlugins with phases AUTHN, AUTHZ and STATS
+	// must be injected before it. This method allows us to inject WasmPlugins in the right spots
+	// while retaining any filters that were unknown at the time of writing this algorithm,
+	// in linear time. The assumed ordering of builtin filters is:
+	//
+	// 1. Istio JWT, 2. Istio AuthN, 3. RBAC, 4. Stats, 5. Metadata Exchange
+	//
+	// The only exception being ext-auch, where two additional filters are at the beginning:
```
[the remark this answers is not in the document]
dgn: fixed

pilot/pkg/model/push_context.go
```diff
@@ -1690,6 +1710,81 @@ func (ps *PushContext) initTelemetry(env *Environment) (err error) {
...
+func (ps *PushContext) initWasmPlugins(env *Environment) error {
+	wasmplugins, err := env.List(gvk.WasmPlugin, NamespaceAll)
+	if err != nil {
+		return err
+	}
+
+	sortConfigByCreationTime(wasmplugins)
+	ps.wasmPluginsByNamespace = map[string][]*WasmPluginWrapper{}
+	for _, plugin := range wasmplugins {
+		if ps.wasmPluginsByNamespace[plugin.Namespace] == nil {
```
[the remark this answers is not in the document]
dgn: fixed

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,122 @@
...
+	Namespace string
+
+	ExtensionConfiguration *envoy_extensions_filters_http_wasm_v3.Wasm
+}
+
+func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
+	if wasmPlugin, ok := plugin.Spec.(*extensions.WasmPlugin); ok {
+		cfg := &any.Any{}
+		if wasmPlugin.PluginConfig != nil && len(wasmPlugin.PluginConfig.Fields) > 0 {
+			cfgJSON, _ := protojson.Marshal(proto.MessageV2(wasmPlugin.PluginConfig))
```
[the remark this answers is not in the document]
dgn: added

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,122 @@
...
+	extensions.WasmPlugin
+
+	Name      string
+	Namespace string
+
+	ExtensionConfiguration *envoy_extensions_filters_http_wasm_v3.Wasm
+}
+
+func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
+	if wasmPlugin, ok := plugin.Spec.(*extensions.WasmPlugin); ok {
```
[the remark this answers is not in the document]
dgn: fixed

dgn:
> @dgn: The following tests **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> integ-security-k8s-tests_istio 	[449600e](https://github.com/istio/istio/commit/449600ee0fa688c4bbe9dee733504481ec549961) 	[link](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/33374/integ-security-k8s-tests_istio/1442488399943438336) 	true 	`/test integ-security-k8s-tests_istio`
> integ-security-multicluster-tests_istio 	[449600e](https://github.com/istio/istio/commit/449600ee0fa688c4bbe9dee733504481ec549961) 	[link](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/33374/integ-security-multicluster-tests_istio/1442488403986747392) 	true 	`/test integ-security-multicluster-tests_istio`

These test failures are unrelated to this PR

pilot/pkg/model/push_context.go
```diff
@@ -1690,6 +1710,81 @@ func (ps *PushContext) initTelemetry(env *Environment) (err error) {
...
+			}
+			if plugin.Selector == nil || workloadLabels.IsSupersetOf(plugin.Selector.MatchLabels) {
+				matchedPlugins[plugin.Phase] = append(matchedPlugins[plugin.Phase], plugin)
+			}
+		}
+	}
+
+	// To prevent duplicate extensions in case root namespace equals proxy's namespace
+	if proxy.ConfigNamespace != ps.Mesh.RootNamespace {
+		for _, plugin := range ps.wasmPluginsByNamespace[proxy.ConfigNamespace] {
```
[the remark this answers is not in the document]
dgn: I'll push this out; I copied this method from EnvoyFilter - I think a refactor should touch both (and look for more potential duplicates)

pilot/pkg/model/push_context_test.go
```diff
@@ -427,6 +428,163 @@ func TestEnvoyFilterOrder(t *testing.T) {
...
+				ConfigNamespace: "testns-1",
+				Metadata: &NodeMetadata{
+					Labels: map[string]string{
+						"istio": "ingressgateway",
+					},
+				},
+			},
+			expectedExtensions: map[extensions.PluginPhase][]*WasmPluginWrapper{
+				extensions.PluginPhase_AUTHN: {
+					convertToWasmPluginWrapper(&wasmPlugins[2]),
```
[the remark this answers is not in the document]
dgn: thanks, makes a lot of sense. I made it a map.

pilot/pkg/model/push_context_test.go
```diff
@@ -427,6 +428,163 @@ func TestEnvoyFilterOrder(t *testing.T) {
...
+			},
+		},
+	}
+
+	testCases := []struct {
+		name               string
+		node               *Proxy
+		expectedExtensions map[extensions.PluginPhase][]*WasmPluginWrapper
+	}{
+		{
```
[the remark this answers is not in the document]
dgn: I just pushed some more tests into `pilot/pkg/networking/core/v1alpha3/extension/wasmplugin_test.go`, with that we have ~90% unit test coverage, plus the integration test

dgn:
Thank you for the thorough reviews @mandarjog and @mathetake. I addressed all your comments; there seems to be some issue with a security integration test that is unrelated to this PR. I don't have time to look at it now, but might have tomorrow

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,129 @@
...
+		})
+	}
+	var datasource *envoy_config_core_v3.AsyncDataSource
+	var sha256 string
+	u, err := url.Parse(wasmPlugin.Url)
+	if err != nil {
+		log.Warnf("wasmplugin %v/%v discarded due to failure to parse URL: %s", plugin.Namespace, plugin.Name, err)
+		return nil
+	}
+	if wasmPlugin.XSha256 == nil {
```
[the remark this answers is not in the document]
howardjohn: I  suspect it will fix the unmarshalling issues and `"nil"` workaround as well

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -621,6 +622,9 @@ func (configgen *ConfigGeneratorImpl) buildInboundFilterchains(in *plugin.InputP
 			}
 		}
 	}
+	if in.Node.Type == model.SidecarProxy {
+		extension.AddWasmPluginsToMutableObjects(mutable, in.Push.WasmPlugins(in.Node))
```
howardjohn, line 626: can you add some comment explaining why only sidecar?

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
bianpengyuan: This is just a random idea coming from the ext authz filter ordering comment, so that we can get complete control over ordering of built-in & Wasm filters if we have predefined priority of built in filters. We discussed this a bit at E&T WG meeting and decided this can save to later iteration.

pkg/config/validation/validation.go
bianpengyuan: User will need to look at the metrics (https://istio.io/latest/docs/ops/configuration/extensibility/wasm-module-distribution/#monitor-wasm-module-distribution see `istio_agent_wasm_remote_fetch_count`), or agent/istiod logs.

mathetake: yeah the code is here:
http: https://github.com/istio/istio/blob/master/pkg/wasm/cache.go#L140
oci: https://github.com/istio/istio/blob/master/pkg/wasm/cache.go#L151

pilot/pkg/model/extensions.go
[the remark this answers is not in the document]
dgn: agreed; this has caused me nothing but pain :slightly_smiling_face: I'll do this after this PR has merged though, because I don't want to delay it even further by waiting on an additional API PR to merge

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,181 @@
...
+	hasName := make(map[string]bool)
+	for _, n := range names {
+		hasName[n] = true
+	}
+	for _, list := range wasmPlugins {
+		for _, p := range list {
+			if _, ok := hasName[p.Name]; !ok {
+				continue
+			}
+			typedConfig, err := anypb.New(p.ExtensionConfiguration)
```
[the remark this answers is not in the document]
mandarjog: Use case is an extension deployed to all sidecars, so moving it to push context makes sense.

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	cfg := &any.Any{}
+	if wasmPlugin.PluginConfig != nil && len(wasmPlugin.PluginConfig.Fields) > 0 {
+		cfgJSON, err := protojson.Marshal(proto.MessageV2(wasmPlugin.PluginConfig))
+		if err != nil {
+			log.Warnf("wasmplugin %v/%v discarded due to json marshaling error: %s", plugin.Namespace, plugin.Name, err)
+			return nil
+		}
+		cfg = gogo.MessageToAny(&types.StringValue{
```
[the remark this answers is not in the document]
dgn: makes sense, but it introduces a cyclic dependecy on `pilot/pkg/model`, which I'll have to work around

pilot/pkg/model/extensions.go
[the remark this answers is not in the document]
howardjohn: Ok, lets make sure we do it before we ship this or it will be a breaking change I think

pkg/config/validation/validation.go
```diff
@@ -3477,3 +3479,53 @@ func validateTelemetryProviders(providers []*telemetry.ProviderRef) error {
...
+func validateWasmPluginSHA(plugin *extensions.WasmPlugin) error {
+	if plugin.XSha256 == nil {
+		return nil
+	}
+	if len(plugin.GetSha256()) != 64 {
+		return fmt.Errorf("sha256 field must be 64 characters long")
+	}
+	for _, r := range plugin.GetSha256() {
+		if !('a' <= r && r <= 'z' || '0' <= r && r <= '9') {
+			return fmt.Errorf("sha256 field must match [a-z0-9]{64} pattern")
```
mathetake, line 3527: sorry actually this should be in hex :D
```suggestion
		if !('a' <= r && r <= 'f') {
			return fmt.Errorf("sha256 field must match [a-f]{64} pattern")
```

mathetake (approved):
LGTM from me, thank you for working on this!

pkg/config/validation/validation.go
dgn: that's absolutely right :man_facepalming:

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,181 @@
...
+			if err != nil {
+				scope.Warnf("wasmplugin %s/%s failed to marshal to TypedExtensionConfig: %s", p.Namespace, p.Name, err)
+				continue
+			}
+			ec := &envoy_config_core_v3.TypedExtensionConfig{
+				Name:        p.Name,
+				TypedConfig: typedConfig,
+			}
+			result = append(result, proto.Clone(ec).(*envoy_config_core_v3.TypedExtensionConfig))
```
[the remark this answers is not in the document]
dgn: keeping the `Clone()`, as I now moved the marshaling to the conversion (triggered in PushContext)

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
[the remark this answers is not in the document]
dgn: makes sense, moved it to the conversion which is triggered in push_context.go

pilot/pkg/networking/core/v1alpha3/listener_builder.go
dgn: added comment

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,181 @@
...
+	statsFilterName = "istio.stats"
+)
+
+var (
+	defaultConfigSource = &envoy_config_core_v3.ConfigSource{
+		ConfigSourceSpecifier: &envoy_config_core_v3.ConfigSource_Ads{
+			Ads: &envoy_config_core_v3.AggregatedConfigSource{},
+		},
+		ResourceApiVersion:  envoy_config_core_v3.ApiVersion_V3,
+		InitialFetchTimeout: &durationpb.Duration{Seconds: 0},
```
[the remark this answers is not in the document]
dgn: added comment

pilot/pkg/model/extensions.go
```diff
@@ -0,0 +1,129 @@
...
+func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
+	var ok bool
+	var wasmPlugin *extensions.WasmPlugin
+	if wasmPlugin, ok = plugin.Spec.(*extensions.WasmPlugin); !ok {
+		return nil
+	}
+
+	cfg := &any.Any{}
+	if wasmPlugin.PluginConfig != nil && len(wasmPlugin.PluginConfig.Fields) > 0 {
+		cfgJSON, err := protojson.Marshal(proto.MessageV2(wasmPlugin.PluginConfig))
```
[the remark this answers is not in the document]
dgn: thanks. using gogoprotomarshal.ToJSON now

pilot/pkg/model/extensions.go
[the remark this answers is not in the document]
dgn: if you don't mind I'll keep this as is and submit a follow-up PR; I have the changes ready, but it requires a refactor (moving the MessageTo* funcs into its own package) that needs a lot changes all over the code base (all references of networkutil.MessageTo* must be replaced with new pkg name)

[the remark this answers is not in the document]
howardjohn: :+1:

pilot/pkg/networking/core/v1alpha3/listener_builder.go
mandarjog: @dgn i don’t see the comment. Did you push ?

dgn: you'll have to the `Files changed` tab to see it, not sure why it doesn't show up here. see https://github.com/istio/istio/blob/ee35d0176ba4cfdc23099315a67e43755114d779/pilot/pkg/networking/core/v1alpha3/listener_builder.go#L625-L628

mandarjog: Thanks @dgn. We need to make sure we have this behaviour specified in the API doc.

mandarjog (approved):
Thanks @dgn

bianpengyuan (commented):
lgtm other than some nits and extension config resource name. Thanks!

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
yangminzhu: okay if we decide to not support ext-authz for now, please make sure to document this properly so that we do not surprise the users.

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -0,0 +1,169 @@
...
+		switch httpFilter.Name {
+		case securitymodel.EnvoyJwtFilterName:
+			newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_AUTHN)
+			newHTTPFilters = append(newHTTPFilters, httpFilter)
+		case securitymodel.AuthnFilterName:
+			newHTTPFilters = popAppend(newHTTPFilters, extMap, extensions.PluginPhase_AUTHN)
+			newHTTPFilters = append(newHTTPFilters, httpFilter)
+		case authzmodel.RBACHTTPFilterName:
+			// when using ext-auth, the first filter will be an RBAC filter. we ignore that one
+			if index > 0 {
```
yangminzhu, line 102: What if the user only has authz (non ext-authz)? In this case the first filter will be the RBAC filter for the ALLOW/DENY action (non ext-authz). 

I think it's not reliable to assume it's an RBAC filter for ext-authz by just checking the index of the filter, this might not be true for some configurations and hard to maintain in the long term, it would be better to use some explicit fields to pass this information.

dgn: you're right; just looking at the index won't work. This will need some more thought. I'll remove the ext-authz support for now, it'll have to come later

pkg/config/validation/validation.go
```diff
@@ -3477,3 +3479,53 @@ func validateTelemetryProviders(providers []*telemetry.ProviderRef) error {
 	}
 	return nil
 }
+
+// ValidateWasmPlugin validates a WasmPlugin.
+var ValidateWasmPlugin = registerValidateFunc("ValidateWasmPlugin",
+	func(cfg config.Config) (Warning, error) {
+		spec, ok := cfg.Spec.(*extensions.WasmPlugin)
+		if !ok {
+			return nil, fmt.Errorf("cannot cast to service entry")
```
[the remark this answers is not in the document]
dgn: fixed

pkg/wasm/convert.go
```diff
@@ -108,29 +108,36 @@ func convert(resource *any.Any, cache Cache) (newExtensionConfig *any.Any, sendN
...
-	if wasmStruct.TypeUrl != wasmHTTPFilterType {
-		wasmLog.Debugf("typed extension config %+v does not contain wasm http filter", wasmStruct)
-		return
-	}
-
 	wasmHTTPFilterConfig := &wasm.Wasm{}
-	if err := conversion.StructToMessage(wasmStruct.Value, wasmHTTPFilterConfig); err != nil {
-		wasmLog.Debugf("failed to convert extension config struct %+v to Wasm HTTP filter", wasmStruct)
+	// Wasm filter can be configured using typed struct and Wasm filter type
+	wasmLog.Infof("original extension config resource %+v", ec)
```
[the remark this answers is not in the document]
dgn: fixed

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
yangminzhu: we can skip the support of ext-authz but should not break when ext-authz is used. The current code actually does not work when ext-authz is used (it would take the very first RBAC filter as AUTHZ position)

a quick way is probably to search the filter chain backwards (from end to beginning):
1. If the current filter is neither RBAC nor ext-authz filter, continue the search
2. If the current filter is ext-authz filter, this means there is no authorization policy of action ALLOW/DENY/AUDIT, stop the loop and treat it as no authz found
3. If the current filter is RBAC filter, you can treat it as the authz policy of action ALLOW/DENY/AUDIT and inject the WASM filter accordingly

This will not support ext-authz filter but also won't break when ext-authz is being used.

howardjohn (approved):
A few nits that will be addressed later. Approving now with hold label for extensions team to give final sign off

mandarjog:
Added `do-not-merge` to give @bianpengyuan and @yangminzhu time to look at the PR again.
Thanks.

bianpengyuan:
I don't think Yangmin and my comments have been fully addressed yet. @dgn are you still working on them?

dgn:
> I don't think Yangmin and my comments have been fully addressed yet. @dgn are you still working on them?

the only thing on my list that is missing is a positive integration test, which I'd like to postpone, otherwise we're not going to make the feature freeze deadline. Is there anything else we'll need to take care of before merging this? If so, I'm afraid this will have to be postponed again

mandarjog:
@bianpengyuan can we get this in with more integration tests to follow immediately? The promise is that we will get more in before ship date.

@dgn if the ext-authz issue can be addressed post facto without api/ behavior breakage we should get it this in now.

bianpengyuan (approved):

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -152,6 +153,7 @@ func (configgen *ConfigGeneratorImpl) buildGatewayListeners(builder *ListenerBui
 					log.Warn("generateListenerAndFilterChains: failed to build listener for gateway: ", err.Error())
 				}
 			}
+			extension.AddWasmPluginsToMutableObjects(&mutable.MutableObjects, builder.push.WasmPlugins(builder.node))
```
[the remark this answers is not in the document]
halfrost: @mandarjog I want to the follow-up of this enhancement, could you tell me in which pr has related enhancements' design?

PR #33375, opened 2021-06-10
Fix multi secret update did not handle cluster delete

hzxuzhonghu:
Fixed a bug, when secret update, should take into consideration the clusters that are being deleted in the secret. 
For example:

Original secret:

```
data:
  cluster1: "kubeconfig1"
  cluster2: "kubeconfig2"
```
Updated secret:

```
data:
  cluster1: "kubeconfig1"
```

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/kube/secretcontroller/secretcontroller.go
+++ b/pkg/kube/secretcontroller/secretcontroller.go
@@ -66,10 +66,9 @@ type removeClientCallback func(clusterID cluster.ID) error
 
 // Controller is the controller implementation for Secret resources
 type Controller struct {
-	kubeclientset kubernetes.Interface
-	namespace     string
-	queue         workqueue.RateLimitingInterface
-	informer      cache.SharedIndexInformer
+	namespace string
+	queue     workqueue.RateLimitingInterface
+	informer  cache.SharedIndexInformer
 
 	cs *ClusterStore
 
@@ -85,7 +84,7 @@ type Controller struct {
 
 // Cluster defines cluster struct
 type Cluster struct {
-	secretName    string
+	clusterID     string
 	kubeConfigSha [sha256.Size]byte
 
 	// Client for accessing the cluster.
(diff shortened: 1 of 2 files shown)


hzxuzhonghu:
@istio/wg-networking-maintainers This is ready to go.

stevenctl (commented):
Seems ok, confused how "update" is handled now. Also the new structure makes it harder to check for duplicate clusters across multiple secrets, can we add a way to handle that before merging?

hzxuzhonghu:
Do you mean it uses `addSecret` to handle update? I donot change that in this pr, if you feels that, i can do.

stevenctl:
It seems your PR makes it so if we have an update, we do the following:

```
delete(cluster)
update(cluster)
```

`update` is implemented as:


https://github.com/istio/istio/blob/9a4c8ddd2bcfc4c100472c5954c38a7791b69a23/pilot/pkg/serviceregistry/kube/controller/multicluster.go#L295-L300

So if I understand correctly, this means we'll call `delete` twice, and I'm assuming the second call will fail.

hzxuzhonghu:
@stevenctl PTAL

```
	for clusterID, kubeConfig := range s.Data {
		action, callback := "Adding", c.addCallback
		if prev := c.cs.Get(secretKey, clusterID); prev != nil {
			action, callback = "Updating", c.updateCallback
			// clusterID must be unique even across multiple secrets
			// TODO： warning
			kubeConfigSha := sha256.Sum256(kubeConfig)
			if bytes.Equal(kubeConfigSha[:], prev.kubeConfigSha[:]) {
				log.Infof("%s cluster_id=%v from secret=%v: (kubeconfig are identical)", clusterID, secretKey)
				continue
			}
		}
```
The cluster in the secret will not be deleted above, because we only delete non existing cluster at the first step.

hzxuzhonghu:
/test unit-tests_istio


ping @stevenctl

stevenctl (approved):
I see, thanks for the fix!

PR #33378, opened 2021-06-10
fix dns resolve for service entry with multi addresses

silenceshell:
dns resolve for service entry with multi addresses only get the last address.
For example, create a service entry as below.

```yaml
kind: ServiceEntry
metadata:
  name: all-tcp-external
  namespace: istio-demo
spec:
  addresses:
  - 192.168.0.204/32
  - 192.168.0.231/32
  hosts:
  - www.example.com
  location: MESH_EXTERNAL
  ports:
  - name: http
    number: 80
```

And try to dig `www.example.com` in a pod with sidecar injected, will only get the last address(192.168.0.231 in this example).

```bash
[root@debian-77dc9c5f4f-nscvz /]# dig www.example.com

; <<>> DiG 9.11.4-P2-RedHat-9.11.4-26.P2.el7_9.5 <<>> www.example.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 54996
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0
;; WARNING: recursion requested but not available

;; QUESTION SECTION:
;www.example.com.		IN	A

;; ANSWER SECTION:
www.example.com.	30	IN	A	192.168.0.231

;; Query time: 0 msec
;; SERVER: 10.96.0.10#53(10.96.0.10)
;; WHEN: Thu Jun 10 14:06:59 UTC 2021
;; MSG SIZE  rcvd: 64
```

we know pilot-discovery converts service entry with multi addresses to multi services.
`pilot/pkg/serviceregistry/serviceentry/conversion.go:buildServices`

```go
func buildServices(hostAddresses []*HostAddress, namespace string, ports model.PortList, location networking.ServiceEntry_Location,
	resolution model.Resolution, exportTo map[visibility.Instance]bool, selectors map[string]string, saccounts []string,
	ctime time.Time, labels map[string]string) []*model.Service {
	out := make([]*model.Service, 0, len(hostAddresses))
	for _, ha := range hostAddresses {
		out = append(out, &model.Service{
			CreationTime: ctime,
			MeshExternal: location == networking.ServiceEntry_MESH_EXTERNAL,
			Hostname:     host.Name(ha.host),
			Address:      ha.address,
...
		})
	}
	return out
}
```

So in `BuildNameTable` when responding xds messages to sidecar, it stores the `nameInfo` to `out.Table[string(svc.Hostname)]` without checking whether `svc.Hostname` is exist in `out.Table` or not.

Thus will cause the `out.Table` only contains the last address of service entry, and name resolve on sidecar will only get the last address.
`pilot/pkg/networking/core/v1alpha3/name_table.go:BuildNameTable`

```go
func (configgen *ConfigGeneratorImpl) BuildNameTable(node *model.Proxy, push *model.PushContext) *nds.NameTable {
...
	out := &nds.NameTable{
		Table: map[string]*nds.NameTable_NameInfo{},
	}
	for _, svc := range push.Services(node) {
...
		nameInfo := &nds.NameTable_NameInfo{
			Ips:      addressList,
			Registry: svc.Attributes.ServiceRegistry,
		}
		if svc.Attributes.ServiceRegistry == string(serviceregistry.Kubernetes) {
			// The agent will take care of resolving a, a.ns, a.ns.svc, etc.
			// No need to provide a DNS entry for each variant.
			nameInfo.Namespace = svc.Attributes.Namespace
			nameInfo.Shortname = svc.Attributes.Name
		}
		out.Table[string(svc.Hostname)] = nameInfo
```

This PR will append the Ips of prev `nameInfo` in `out.Table` with the same `svc.Hostname` to the current `nameInfo`, and dig will get all IP addresses for host of service entry.

```bash
[root@debian-77dc9c5f4f-nscvz /]# dig www.example.com

; <<>> DiG 9.11.4-P2-RedHat-9.11.4-26.P2.el7_9.5 <<>> www.example.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 58814
;; flags: qr aa rd; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 0
;; WARNING: recursion requested but not available

;; QUESTION SECTION:
;www.example.com.		IN	A

;; ANSWER SECTION:
www.example.com.	30	IN	A	192.168.0.231
www.example.com.	30	IN	A	192.168.0.204

;; Query time: 0 msec
;; SERVER: 10.96.0.10#53(10.96.0.10)
;; WHEN: Thu Jun 10 14:33:48 UTC 2021
;; MSG SIZE  rcvd: 95
```

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/name_table.go
+++ b/pilot/pkg/networking/core/v1alpha3/name_table.go
@@ -119,6 +119,10 @@ func (configgen *ConfigGeneratorImpl) BuildNameTable(node *model.Proxy, push *mo
 			nameInfo.Namespace = svc.Attributes.Namespace
 			nameInfo.Shortname = svc.Attributes.Name
 		}
+
+		if ni, ok := out.Table[string(svc.Hostname)]; ok {
+			nameInfo.Ips = append(nameInfo.Ips, ni.Ips...)
+		}
 		out.Table[string(svc.Hostname)] = nameInfo
 	}
 	return out


pilot/pkg/networking/core/v1alpha3/name_table.go
```diff
@@ -119,6 +119,10 @@ func (configgen *ConfigGeneratorImpl) BuildNameTable(node *model.Proxy, push *mo
 			nameInfo.Namespace = svc.Attributes.Namespace
 			nameInfo.Shortname = svc.Attributes.Name
 		}
+
+		if ni, ok := out.Table[string(svc.Hostname)]; ok {
```
howardjohn, line 123: I am not sure this is safe. We don't have a way to tell if there are two services for the same hostname because they are in one ServiceEntry, or if they are two service entries in different namespaces, etc.

silenceshell: Thanks for you quick repose!
Is it correct to add two service entries with the same hosts? I have just created two service entries on 1.10(in different namespace), and istio do not give warns, but I found the hosts is resolved to the address of the service entry which is in one namespace(perhaps the namespace whose has an alphabetical order).
I have also created two service entries with same hosts in the same namespace, the behavior is similar.

howardjohn: The intent is that `push.Services(node)` will return the correct service based on the Sidecar scoping. If there is no scope selecting only a single one then one is chosen (oldest)

hzxuzhonghu: I think this is better than before, ServiceEntry has many known issues. Users can influence each other unintendedly by mistake. For example, set different addresses for the same hostname in different ses.
Just one point should deduplicate ips.

hzxuzhonghu:
One further step, when changed the protocol to HTTP, the rds it generated is not complete, missing `192.168.0.204` in virtualHosts.Domains

PR #33383, opened 2021-06-10
No need to apply istio-root-ca-cert configmap to special kubernetes system namespaces.

SpecialYang:
We no need to apply `istio-root-ca-cert` configmap to special kubernetes system namespaces, because istio never use these namespaces, such as kube-system, kube-public.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
@@ -16,6 +16,7 @@ package controller
 
 import (
 	"fmt"
+	"istio.io/istio/pkg/kube/inject"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


howardjohn (commented):
This seems like a good idea but would be good to get some feedback from others as well

hzxuzhonghu (commented):
We have clusterlocal namespaces, maybe add them all to it.

SpecialYang:
> We have clusterlocal namespaces, maybe add them all to it.

Now, these special kubernetes system namespaces are included in sidecar inject module which will do not inject sidecar for pods in these namespaces. So these namespaces no need istio-root-ca-cert. Do you mean we should add other namespaces in ignored namespaces?

hzxuzhonghu:
I mean in `pilot/pkg/model/cluster_local.go` there is a `defaultClusterLocalNamespaces` var

SpecialYang:
> I mean in `pilot/pkg/model/cluster_local.go` there is a `defaultClusterLocalNamespaces` var

Currently, the namespaces in `defaultClusterLocalNamespaces ` only have `kube-system` and I did not find any place to modify this var. So I think we only need to consider `IngoredNamespaces` var in inject module which also have `kube-system` namespace.

hzxuzhonghu (approved):

Issue #33385, opened 2021-06-10
regression: custom TLS settings in envoyfilter no longer merge


howardjohn:
An envoyfilter like:

```
      value:
         transport_socket:
           name: my-custom-config
           typed_config:
             ...
```

No longer has an effect. Broken by https://github.com/istio/istio/commit/1461bd714e10aad760a2a7ad1c817004e3297be2

PR #33386, opened 2021-06-10
Fix regression when using custom transport_socket envoyfilter

howardjohn:
Fixes https://github.com/istio/istio/issues/33385

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/envoyfilter/cluster_patch.go
+++ b/pilot/pkg/networking/core/v1alpha3/envoyfilter/cluster_patch.go
@@ -65,7 +65,9 @@ func ApplyClusterMerge(pctx networking.EnvoyFilter_PatchContext, efw *model.Envo
 }
 
 // Test if the patch contains a config for TransportSocket
-func mergeTransportSocketCluster(c *cluster.Cluster, cp *model.EnvoyFilterConfigPatchWrapper) (bool, error) {
+// Returns a boolean indicating if the merge was handled by this function; if false, it should still be called
+// outside of this function.
+func mergeTransportSocketCluster(c *cluster.Cluster, cp *model.EnvoyFilterConfigPatchWrapper) (merged bool, err error) {
 	cpValueCast, okCpCast := (cp.Value).(*cluster.Cluster)
 	if !okCpCast {
 		return false, fmt.Errorf("cast of cp.Value failed: %v", okCpCast)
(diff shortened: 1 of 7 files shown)


pilot/pkg/networking/core/v1alpha3/envoyfilter/cluster_patch.go
```diff
@@ -82,9 +82,16 @@ func mergeTransportSocketCluster(c *cluster.Cluster, cp *model.EnvoyFilterConfig
 				break
 			}
 		}
+		if tsmPatch == nil && len(c.GetTransportSocketMatches()) > 0 {
+			c.GetTransportSocketMatches()[0].TransportSocket = cpValueCast.TransportSocket
```
howardjohn, line 86: not sure this is right.. if they have a transport socket match and the name doesn't match what should we actually do?

ramaraochavali: Yeah. I think this is not correct. This might incorrectly replace another context. May be we should log a warning and skip?

ramaraochavali:
Looks good. Apart from that comment.

pilot/pkg/networking/core/v1alpha3/envoyfilter/cluster_patch.go
howardjohn: My concern was that this means you can effectively not use custom transport socket in istio auto mtls. Maybe that is ok. The user that complained about the issue was only using for listener which is un-ambiguous. I will remove that for now, it didn't work prior to the PR that broke things anyways, so its a new feature essentially

pilot/pkg/networking/core/v1alpha3/envoyfilter/listener_patch.go
```diff
@@ -192,7 +192,7 @@ func patchFilterChain(patchContext networking.EnvoyFilter_PatchContext,
 }
 
 // Test if the patch contains a config for TransportSocket
-func mergeTransportSocketListener(fc *xdslistener.FilterChain, lp *model.EnvoyFilterConfigPatchWrapper) (bool, error) {
+func mergeTransportSocketListener(fc *xdslistener.FilterChain, lp *model.EnvoyFilterConfigPatchWrapper) (alreadyMerged bool, err error) {
```
mandarjog, line 195: Additiona comment.
“ and return true if patch was applied”.

pilot/pkg/networking/core/v1alpha3/envoyfilter/listener_patch.go
```diff
@@ -178,12 +178,12 @@ func patchFilterChain(patchContext networking.EnvoyFilter_PatchContext,
 			// nothing more to do in other patches as we removed this filter chain
 			return
 		} else if lp.Operation == networking.EnvoyFilter_Patch_MERGE {
-			ret, err := mergeTransportSocketListener(fc, lp)
+			alreadyMerged, err := mergeTransportSocketListener(fc, lp)
```
mandarjog, line 181: Nit: “merged”

pilot/pkg/networking/core/v1alpha3/envoyfilter/cluster_patch.go
```diff
@@ -82,9 +82,17 @@ func mergeTransportSocketCluster(c *cluster.Cluster, cp *model.EnvoyFilterConfig
 				break
 			}
 		}
+		if tsmPatch == nil && len(c.GetTransportSocketMatches()) > 0 {
+			// If we merged we would get both a transport_socket and transport_socket_matches which is not valid
+			// Drop the filter
+			return true, nil
```
mandarjog, line 88: What does true represent in this function ?
I think it means “I did all that was needed, but patch may not have been applied”
If so it is good define that in the function header.

mandarjog (approved):
Thnx.

ramaraochavali (approved):

PR #32081, opened 2021-04-09
do not recompute sidecar and gateway if not changed

ramaraochavali:
Do not recompute sidecar and merge gateways if they have not changed while pushing to proxy
[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ X] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[X] Does not have any changes that may affect Istio users.

PR #33391, opened 2021-06-11
cleanup set workload labels

ramaraochavali:
If proxy bootstrap does not provide labels in metadata(happens for noninjected proxies) and by that time pod connects to istiod if the pod is not available in istiod pod cache, we never set work load labels later -and any sidecar that is applied to it with workload selector does not work

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -819,6 +819,19 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
 	node.ServiceInstances = instances
 }
 
+// SetWorkloadLabels will set the node.Metadata.Labels only when it is nil.
+func (node *Proxy) SetWorkloadLabels(env *Environment) {
+	// First get the workload labels from node meta
+	if len(node.Metadata.Labels) > 0 {
+		return
+	}
+	// Fallback to calling GetProxyWorkloadLabels
+	l := env.GetProxyWorkloadLabels(node)
+	if len(l) > 0 {
+		node.Metadata.Labels = l[0]
+	}
+}
+
 // DiscoverIPVersions discovers the IP Versions supported by Proxy based on its IP addresses.
 func (node *Proxy) DiscoverIPVersions() {
 	for i := 0; i < len(node.IPAddresses); i++ {
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -819,20 +819,6 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
 	node.ServiceInstances = instances
 }
 
-// SetWorkloadLabels will set the node.Metadata.Labels only when it is nil.
-func (node *Proxy) SetWorkloadLabels(env *Environment) {
-	// First get the workload labels from node meta
-	if len(node.Metadata.Labels) > 0 {
-		return
-	}
-
-	// Fallback to calling GetProxyWorkloadLabels
-	l := env.GetProxyWorkloadLabels(node)
-	if len(l) > 0 {
-		node.Metadata.Labels = l[0]
-	}
-}
-
 // DiscoverIPVersions discovers the IP Versions supported by Proxy based on its IP addresses.
 func (node *Proxy) DiscoverIPVersions() {
 	for i := 0; i < len(node.IPAddresses); i++ {
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -554,6 +554,9 @@ func (s *DiscoveryServer) initProxyMetadata(node *core.Node) (*model.Proxy, erro
 	if err != nil {
 		return nil, err
 	}
+	if len(meta.Labels) == 0 {
+		log.Warnf("proxy %s does not have labels in metadata, some functionality might not work properly", node.Id)
+	}
 	proxy, err := model.ParseServiceNodeWithMetadata(node.Id, meta)
 	if err != nil {
 		return nil, err
@@ -573,7 +576,6 @@ func (s *DiscoveryServer) initializeProxy(node *core.Node, con *Connection) erro
 	if err := s.WorkloadEntryController.RegisterWorkload(proxy, con.Connect); err != nil {
 		return err
 	}
-
 	s.computeProxyState(proxy, nil)
 
 	// Get the locality from the proxy's service instances.
@@ -621,7 +623,6 @@ func (s *DiscoveryServer) updateProxy(proxy *model.Proxy, request *model.PushReq
 
 func (s *DiscoveryServer) computeProxyState(proxy *model.Proxy, request *model.PushRequest) {
 	proxy.SetServiceInstances(s.Env.ServiceDiscovery)
-	proxy.SetWorkloadLabels(s.Env)
 	// Precompute the sidecar scope and merged gateways associated with this proxy.
 	// Saves compute cycles in networking code. Though this might be redundant sometimes, we still
 	// have to compute this because as part of a config change, a new Sidecar could become
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -554,9 +554,6 @@ func (s *DiscoveryServer) initProxyMetadata(node *core.Node) (*model.Proxy, erro
 	if err != nil {
 		return nil, err
 	}
-	if len(meta.Labels) == 0 {
-		log.Warnf("proxy %s does not have labels in metadata, some functionality might not work properly", node.Id)
-	}
 	proxy, err := model.ParseServiceNodeWithMetadata(node.Id, meta)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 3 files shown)


kyessenov:
Is it possible that we ship some xDS config to proxy before pod cache is ready? That may expose the workload for whatever brief time.

howardjohn:
> Is it possible that we ship some xDS config to proxy before pod cache is ready? That may expose the workload for whatever brief time.

+1, this is not secure. The pod MUST send labels over node metadata. We have proven RBAC bypass is possible in real world clusters when this is not sent back in ~1.4 which is why we added the labels over node meta

kyessenov:
I don't necessary agree that the pod needs to send labels but we should hold response until we get pod info. This is no different than network policy enforcement (controller wait for sync before enforcement).

howardjohn:
@kyessenov that is either a very hard or impossible problem to solve, depending on what we are willing to break. XDS clients need not be pods, and they may come from arbitrary things, including VMs, client tools (istioctl), etc.

kyessenov:
I don't really care about superficial uses of xDS for debugging or client tooling. Those could be easily differentiated from the rest. Why is it impossible in other cases? Istiod already knows about all endpoints and their labels for load balancing and telemetry. This is just a reverse lookup.

howardjohn:
VMs don't have a pod either. There are also some more bespoke topologies where clients may connect from other clusters but Istiod is not reading pods in those clusters.

kyessenov:
VMs have workload entry. The interesting thing is istiod treating some workloads as servers but not clients (maybe one cluster writes, another reads). Is that very common? The basic single logical mesh should have all the metadata in one istiod.

howardjohn:
VMs *can* (and should) have a workload entry. There is no requirement that they MUST have one

kyessenov:
Right, I'm not asking to require label-less proxies. But if proxy has no metadata, and we know that istiod should have it, why don't we permit metadata population in istiod? The user can decide whether their topology fits and whether they have all workload entries. The critical functional piece is holding the response until metadata is synced.

howardjohn:
Because its complex and and I haven't seen a compelling reason to not include the metadata, so I am not sure its actually needed. In the original description "non-injected proxies" where mentioned. They can get the meta from downward API

ramaraochavali:
> 
Because its complex and and I haven't seen a compelling reason to not include the metadata, so I am not sure its actually needed. In the original description "non-injected proxies" where mentioned. They can get the meta from downward API

I agree. That is how we are using now.

BTW, It used to be like that earlier i.e. we were setting workload labels on every push. I changed it in this PR https://github.com/istio/istio/pull/32081 to not set.

hzxuzhonghu:
For pods, we mount the labels with downwardAPI, so it could not happen. We could not change labels after pod started. For vm, it should keep consistent. 
```
  - downwardAPI:
      defaultMode: 420
      items:
      - fieldRef:
          apiVersion: v1
          fieldPath: metadata.labels
        path: labels
      - fieldRef:
          apiVersion: v1
          fieldPath: metadata.annotations
        path: annotations
      - path: cpu-limit
        resourceFieldRef:
          containerName: istio-proxy
          divisor: 1m
          resource: limits.cpu
      - path: cpu-request
        resourceFieldRef:
          containerName: istio-proxy
          divisor: 1m
          resource: requests.cpu
    name: istio-podinfo

```

ramaraochavali:
So we do not want to get labels from pod? 

Should we log a warning if node metadata does not have labels?

ramaraochavali:
If this is what we want - I will fix the unit tests. Some UT seems to depend on this call.

pilot/pkg/xds/ads.go
```diff
@@ -554,6 +554,9 @@ func (s *DiscoveryServer) initProxyMetadata(node *core.Node) (*model.Proxy, erro
 	if err != nil {
 		return nil, err
 	}
+	if len(meta.Labels) == 0 {
```
howardjohn, line 557: No labels is fine valid - it is different to have empty labels and unset labels

ramaraochavali: >  it is different to have empty labels and unset labels

How do we distinguish? If metadata does not have labels - doesn't it still cause some functionality like sidecar not function correctly?

howardjohn: Sidecar functions perfectly fine - you cannot select the pod as it has no labels. Just like Service, etc.

I am worried about someone getting spammed with logs when they have perfectly valid pods

ramaraochavali: I was referring to Sidecar created with Workloadselectors - May be we should warn in sidecar code when it finds sidecar with workload selector in namespace but proxy does not have labels.

I will remove it from here

howardjohn: That is also valid, you can have some pods with labels and some pods without labels in the same namespace?

ramaraochavali: Probably. But since the pods in the namespace are hopefully fewer - it does not spam that much in the sidecar logic. 

Are the other changes are OK? If yes. I can fix the tests.

ramaraochavali:
@howardjohn @hzxuzhonghu this is ready for review

pilot/pkg/model/context.go
```diff
@@ -819,20 +819,6 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
...
 
-// SetWorkloadLabels will set the node.Metadata.Labels only when it is nil.
-func (node *Proxy) SetWorkloadLabels(env *Environment) {
-	// First get the workload labels from node meta
-	if len(node.Metadata.Labels) > 0 {
-		return
-	}
-
-	// Fallback to calling GetProxyWorkloadLabels
-	l := env.GetProxyWorkloadLabels(node)
```
howardjohn, line 830: so do we never call GetProxyWorkloadLabels anymore?

ramaraochavali: yes. We always expect and use labels from Metadata only.

howardjohn: If the function is never used then seems like we should clean it up entirely ?

ramaraochavali: Yes. I was thinking of that. Since it is in service discovery interface - involves lot more changes. Will do it in a follow-up

hzxuzhonghu (changes requested):
I donot think for vm it is right. 

1. does vm application can connect to pilot without labels ?

2. WLG allows setting labels, which i understrand is used to build wle.

Both the above labels will not be honored, right?

ramaraochavali:
Yes. VMs can have endpoints labels https://github.com/istio/istio/blob/master/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go#L808 - So are you saying we should keep this functionality as is ? If yes, I think we should continue to call SetWorkloadlabels when proxy connects. But https://github.com/istio/istio/blob/f401b494bfa956990cabe61a2c501fd0ccb0d3af/pilot/pkg/model/context.go#L823 does not look at endpoint labels if metadata has labels - Is that what we want or do we want to combine both?

hzxuzhonghu:
I vote for `endpoint label override metadata labels `, 

The label can be changed, even k8s pod (not sure how often), user can change the label on the flight., the metadata labels was introduced to not query pod(because pod cache may not be synced when proxy connects ).

ramaraochavali:
> The label can be changed, even k8s pod (not sure how often), user can change the label on the flight., the metadata labels was introduced to not query pod(because pod cache may not be synced when proxy connects ).

With out pod restart?

If yes, +1 for overriding endpoint labels over metadata if it exists. 

@howardjohn  are you fine with that?

howardjohn:
I think we had some issue in the past with dynamic label changes not
triggering a push properly? So we may need to handle that.

I think it is good to support dynamic labels since kubernetes does , but if
we do we need to make sure it works fully I think? so it may take a bit
more effort?

kyessenov:
The fact that labels can change at runtime is a good reason not to use injected labels and always rely on pod cache to get them. There is a subtle race condition since bootstrap is also using labels in the template (so our telemetry is screwed when pod labels change).

ramaraochavali:
@howardjohn Does not endpoint/pod update when label changes? If yes it may be easier to handle

hzxuzhonghu:
No

ramaraochavali:
@howardjohn @hzxuzhonghu I reverted it to a state where it used to be i.e. setWorkloadlabels on init proxy and on every push. PTAL.

We can discuss more about dynamic labels and improve it.

ramaraochavali:
@hzxuzhonghu can you PTAL when you get chance?

ramaraochavali:
@hzxuzhonghu gentle ping. Any concerns with this?

hzxuzhonghu (approved):
Thanks，good for VM bootstrap without labels

PR #33113, opened 2021-05-26
check if virtual service destinations has matching ports

ramaraochavali:
https://github.com/istio/istio/pull/20408 we added support for adding additional virtual host destinations. But it does not check if the matching destination has the matching ports when egress listener has ports. This PR fixes. Please see the test case added for the problem.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[X] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #33393, opened 2021-06-11
check if virtual service destinations has matching ports (take 2)

ramaraochavali:
Reimplement https://github.com/istio/istio/pull/33113 . Basically changed it to pick unique services and moved the matching logic outside of loop.
[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


pilot/pkg/model/sidecar.go
```diff
@@ -348,7 +351,15 @@ func ConvertToSidecarScope(ps *PushContext, sidecarConfig *config.Config, config
 				// Default to this hostname in our config namespace
 				if s, ok := ps.ServiceIndex.HostnameAndNamespace[host.Name(h)][configNamespace]; ok {
 					// This won't overwrite hostnames that have already been found eg because they were requested in hosts
-					addService(s)
+					var vss *Service
```
stevenctl, line 354: nit: seems more readable to have

```
vss := s
```

and I'm not sure if the nil check is necessary? Can we have a nil service in the index?

stevenctl (approved):

pilot/pkg/model/sidecar.go
ramaraochavali: nil check is needed for other case i.e. matchPort is enabled and service does not have port - we just return nil. The same reason I have n't assigned vss to s.

pilot/pkg/model/sidecar.go
```diff
@@ -371,7 +382,15 @@ func ConvertToSidecarScope(ps *PushContext, sidecarConfig *config.Config, config
 						sort.Strings(ns)
 						// Pick first namespace alphabetically
 						// This won't overwrite hostnames that have already been found eg because they were requested in hosts
-						addService(byNamespace[ns[0]])
+						var vss *Service
+						if matchPort {
+							vss = serviceMatchingListenerPort(byNamespace[ns[0]], listener)
```
l8huang, line 387: @ramaraochavali could you please kindly let me know if any design doc related to this change? Thanks.

we are using an old istio version, virtual service's destination is egress gateway service, which uses a port number different from sidecar's egress listener. This change cause it doesn't work -- serviceMatchingListenerPort() return nil.

ramaraochavali: Unfortunately there is no design doc except for the description here https://github.com/istio/istio/pull/33113#issue-902198310.

l8huang: @ramaraochavali thanks for your reply. This breaks our egress gateway use case, so I trying to understand if any reason behind this change.

ramaraochavali: The reason is the sidecar egress listener port is intended to restrict services that are explicitly listed in the sidecar definition. Later it has been enhanced to also bring in services that match the virtual service destinations - when we did that we did not consider port. So it was bringing destinations that does not have that matching port which are useless and is bringing lot more clusters. So that change was made.
I think there are two ways to solve this
- Either you explicitly list that service in your sidecar
- Ignore this check for external services
Not sure if there are other options @hzxuzhonghu any ideas?

hzxuzhonghu: >Ignore this check for external services

I donot agree with this way, it makes external service being treated specially.

Not sure should we add some system namespace here, like [kube-system, istio-system], maybe not needed to

howardjohn: Explicitly list that service

l8huang: > Either you explicitly list that service in your sidecar

@ramaraochavali do you mean add the wanted service in Sidecar's `egress` listener? that would cause extra outbound listener being generated, right?

Per my understanding, it's reasonable that route destination could be different from listener's port, so instead of checking destination service's port against listener port by
```
vss = serviceMatchingListenerPort(byNamespace[ns[0]], listener)
```
it should do
```
vss = serviceMatchingVirtualServicePorts(byNamespace[ns[0]], ports)
```

the `if matchPort` condition is not required here. Please kindly correct me if I misunderstand.

ramaraochavali: matchPort block is used to look at **service** if egress listener has ports. As done [here](https://github.com/istio/istio/blob/8f7d578ce031dba8394104a90e5cfe31a9c417d8/pilot/pkg/model/sidecar.go#L360) for same namespace - We are doing the same here. Our goal is to always filter services based on port.

l8huang: Your goal is clear, but when inferring service from VirtualService destination, **the service should match route destination port, instead of listener's port**. Below if the latest code on master branch, IMO the `if matchPort` is not required:


```
					if len(ns) > 0 {
						sort.Strings(ns)
						// Pick first namespace alphabetically
						// This won't overwrite hostnames that have already been found eg because they were requested in hosts
						var vss *Service
						if matchPort {
							vss = serviceMatchingListenerPort(byNamespace[ns[0]], listener)
						} else {
							vss = serviceMatchingVirtualServicePorts(byNamespace[ns[0]], ports)
						}
						if vss != nil {
							addService(vss)
						}
					}

```

PR #33399, opened 2021-06-11
Add configmap write access for remote cluster

frankbu:
Istiod needs to be able to create the ca root cert configmap on a pure (non-config) remote cluster.

[ ] Docs
[x] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/manifests/charts/base/templates/clusterrole.yaml
+++ b/manifests/charts/base/templates/clusterrole.yaml
@@ -159,5 +159,10 @@ rules:
   - apiGroups: [""]
     resources: ["configmaps"]
     verbs: ["create", "get", "list", "watch", "update"]
+
+  # Needed to support patching of isitiod revisions
+  - apiGroups: ["admissionregistration.k8s.io"]
+    resources: ["mutatingwebhookconfigurations"]
+    verbs: ["get", "list", "watch", "update", "patch"]
 {{- end}}
 ---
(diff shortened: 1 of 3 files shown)


manifests/charts/base/templates/clusterrole.yaml
```diff
@@ -154,15 +154,10 @@ rules:
...
     resources: ["subjectaccessreviews"]
     verbs: ["create"]
-{{- if or .Values.global.externalIstiod }}
+{{- if .Values.global.externalIstiod }}
+  # Needed to install CA root cert on remote clusters
   - apiGroups: [""]
     resources: ["configmaps"]
     verbs: ["create", "get", "list", "watch", "update"]
-  - apiGroups: ["admissionregistration.k8s.io"]
-    resources: ["mutatingwebhookconfigurations"]
```
howardjohn, line 162: why remove this?

frankbu, line 162: We do install the mutatingwebhookconfiguration on a (non config cluster) remote cluster, but we don't support revisions there, so these aren't needed. I'll add it back if I'm mistaken (or in the future, if/when we need it).

cc @howardjohn Is this right?

PR #33400, opened 2021-06-11
[tf] Consolidate --istio.test.revision and --istio.test.revisions flags

Monkeyanator:
Removes some confusion with having two revision-related flags. Now, if passed `--istio.test.revision`, the test framework internally adds a single revision to its `Revisions` field. If a test needs to access an arbitrary revision it can get it with `ctx.Settings().Revisions.,Default()`.

Also separates out the compatibility testing into a separate test flag `--istio.test.compatibility`. This lets you make test framework aware of multiple revisions without transparently deploying echoes.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


pkg/test/framework/resource/flags.go
```diff
@@ -46,19 +46,36 @@ func SettingsFromCommandLine(testID string) (*Settings, error) {
...
+			" -istio.test.deprecation_failure must not be used at the same time")
 	}
 
-	if s.Revision != "" && s.Revisions != nil {
-		return nil,
-			fmt.Errorf("cannot use --istio.test.revision and --istio.test.revisions at the same time," +
+	if s.Revision != "" {
+		if s.Revisions != nil {
+			return fmt.Errorf("cannot use --istio.test.revision and --istio.test.revisions at the same time," +
 				" --istio.test.revisions will take precedence and --istio.test.revision will be ignored")
```
stevenctl, line 66: This message is a bit confusing. It seems like if one flag takes precedence it should be a warning rather than an error.

pkg/test/framework/resource/flags.go
```diff
@@ -46,19 +46,36 @@ func SettingsFromCommandLine(testID string) (*Settings, error) {
...
+		}
+		// use Revision as the sole revision in RevVerMap
+		s.Revisions = RevVerMap{
+			s.Revision: "",
```
stevenctl, line 70: Assume the version is filled in with the latest somewhere else? If not is it ok to use this empty value in templates?

pkg/test/framework/resource/flags.go
```diff
@@ -46,19 +46,36 @@ func SettingsFromCommandLine(testID string) (*Settings, error) {
...
+		}
 	}
 
-	return s, nil
+	if s.Compatibility && s.Revisions == nil {
```
stevenctl, line 74: Seems like `Compatibility` is off by default. Will that break jobs that rely on the existing behavior?

pkg/test/framework/resource/flags.go
Monkeyanator: Passing both flags should result in an error IMO since there is no clear precedence order (probably a misconfig if both are set).

pkg/test/framework/resource/flags.go
Monkeyanator: Yeah so we've been using `""` as a sentinel meaning "latest version", we should probably have an explicit constant instead.

pkg/test/framework/resource/flags.go
Monkeyanator: Great point... OSS has no jobs using it but I'll ask around in T&R tomorrow to see if others do.

Monkeyanator: From  T&R seems like we're good.

stevenctl (approved):

Issue #33398, opened 2021-06-11
`istioctl ps` results in auth errors in Istiod


howardjohn:
This is just running plain `istioctl ps`. I then get this error in Istiod:

```
2021-06-11T17:58:27.598851Z     error   Failed to authenticate /debug/syncz [Authenticator ClientCertAuthenticator: no client certificate is presented Authenticator KubeJWTAuthenticator: target JWT extraction error: no HTTP authorization header exists]
```

Running on master. The command does work, but has thr error

PR #33403, opened 2021-06-11
Fix istioctl auth errors

howardjohn:
Fixes https://github.com/istio/istio/issues/33398

Before, the logic was: try ProxyGet, then fallback to Exec because
ProxyGet doesn't work with firewall rules.

New logic: just port-forward, which is what we use for envoy commands
already.

This is preferred since it makes the request come from localhost (thus
bypassing the istiod authz) and works with firewall so we don't need the
exec fallback.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


ericvn (approved):

shamsher31 (approved):

howardjohn:
@istio/wg-networking-maintainers ptal

ramaraochavali (approved):

Issue #33405, opened 2021-06-11
1.10.1 issue with mixing TLS PASSTHROUGH and SIMPLE modes


maciekm:
**Bug description**
given a mix of gateways with TLS PASSTHROUGH and SIMPLE modes: 
```yaml
apiVersion: networking.istio.io/v1beta1
kind: Gateway
metadata:
  name: gateway-private
  namespace: istio-system
spec:
  selector:
    app: istio-ingressgateway
  servers:
  - hosts:
    - '*.example.com'
    port:
      name: http
      number: 80
      protocol: HTTP
    tls:
      httpsRedirect: true
  - hosts:
    - '*.example.com'
    port:
      name: https
      number: 443
      protocol: HTTPS
    tls:
      credentialName: wildcard-private-cert
      mode: SIMPLE
---
apiVersion: networking.istio.io/v1beta1
kind: Gateway
  name: gateway-foo
  namespace: istio-system
spec:
  selector:
    app: istio-ingressgateway
  servers:
  - hosts:
    - foo.example.com
    port:
      name: http
      number: 80
      protocol: HTTP
    tls:
      httpsRedirect: true
  - hosts:
    - foo.example.com
    port:
      name: tls-foo
      number: 443
      protocol: TLS
    tls:
      mode: PASSTHROUGH
```
in insto **1.10.1** `istiod` complains:
```
2021-06-11T21:31:51.074820Z	warn	ads	ADS:LDS: ACK ERROR router~10.52.1.100~istio-ingressgateway-internal-6bc6b9cf87-q96jl.istio-system~istio-system.svc.cluster.local-171 Internal:Error adding/updating listener(s) 0.0.0.0_8443: duplicate listener 0.0.0.0_8443 found
```
and `istio-ingressgateway` complains:
```
2021-06-11T21:55:03.277665Z	warn	Envoy proxy is NOT ready: config not received from Pilot (is Pilot running?): cds updates: 2 successful, 0 rejected; lds updates: 0 successful, 2 rejected
```
proxy-config is:
```
$ istioctl proxy-config listeners istio-ingressgateway-xyz.istio-system
ADDRESS PORT  MATCH                  DESTINATION
0.0.0.0 8080  ALL                    Route: http.80
0.0.0.0 8443  SNI: foo.example.com   Cluster: outbound|443||foo-server.foo.svc.cluster.local
0.0.0.0 15021 ALL                    Inline Route: /healthz/ready*
[...]

PR #33411, opened 2021-06-13
fix gateway merge on same port

ramaraochavali:
https://github.com/istio/istio/issues/33405

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

[] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

### DIFF
--- a/pilot/pkg/model/gateway.go
+++ b/pilot/pkg/model/gateway.go
@@ -231,10 +231,11 @@ func MergeGateways(gateways []gatewayWithInstances) *MergedGateway {
 							log.Warnf("TLS server without TLS options %s %s", gatewayName, s.String())
 							continue
 						}
-						if mergedServers[serverPort] == nil {
-							mergedServers[serverPort] = &MergedServers{Servers: []*networking.Server{s}}
-						} else {
-							mergedServers[serverPort].Servers = append(mergedServers[serverPort].Servers, s)
+						for port := range mergedServers {
+							if port.Number == serverPort.Number {
+								mergedServers[port].Servers = append(mergedServers[port].Servers, s)
+								break
+							}
 						}
 					}
 				} else {
--- a/pilot/pkg/model/gateway.go
+++ b/pilot/pkg/model/gateway.go
@@ -231,11 +231,10 @@ func MergeGateways(gateways []gatewayWithInstances) *MergedGateway {
 							log.Warnf("TLS server without TLS options %s %s", gatewayName, s.String())
 							continue
 						}
-						for port := range mergedServers {
-							if port.Number == serverPort.Number {
-								mergedServers[port].Servers = append(mergedServers[port].Servers, s)
-								break
-							}
+						if mergedServers[serverPort] == nil {
+							mergedServers[serverPort] = &MergedServers{Servers: []*networking.Server{s}}
+						} else {
+							mergedServers[serverPort].Servers = append(mergedServers[serverPort].Servers, s)
 						}
 					}
 				} else {
--- a/pilot/pkg/model/gateway_test.go
+++ b/pilot/pkg/model/gateway_test.go
@@ -121,7 +121,7 @@ func TestMergeGateways(t *testing.T) {
 		{
 			"simple-passthrough",
 			[]config.Config{gwSimple, gwPassthrough},
-			1,
+			2,
 			2,
 			map[string]int{"https.443.https.foo-simple.not-default-2": 1},
 			2,
--- a/pilot/pkg/model/gateway_test.go
+++ b/pilot/pkg/model/gateway_test.go
@@ -22,6 +22,7 @@ import (
 	"istio.io/istio/pkg/config"
 )
 
+// nolint lll
 func TestMergeGateways(t *testing.T) {
 	gwHTTPFoo := makeConfig("foo1", "not-default", "foo.bar.com", "name1", "http", 7, "ingressgateway", "", networking.ServerTLSSettings_SIMPLE)
 	gwHTTPbar := makeConfig("bar1", "not-default", "bar.foo.com", "bname1", "http", 7, "ingressgateway", "", networking.ServerTLSSettings_SIMPLE)
--- a/pilot/pkg/model/gateway_test.go
+++ b/pilot/pkg/model/gateway_test.go
@@ -23,14 +23,17 @@ import (
 )
 
 func TestMergeGateways(t *testing.T) {
-	gwHTTPFoo := makeConfig("foo1", "not-default", "foo.bar.com", "name1", "http", 7, "ingressgateway", "")
-	gwHTTPbar := makeConfig("bar1", "not-default", "bar.foo.com", "bname1", "http", 7, "ingressgateway", "")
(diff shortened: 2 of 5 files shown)


pilot/pkg/model/gateway_test.go
```diff
@@ -109,9 +113,17 @@ func TestMergeGateways(t *testing.T) {
 			"http-http2-server-config",
 			[]config.Config{gwHTTPWildcard, gwHTTP2Wildcard},
 			1,
-			2,
+			1,
```
ramaraochavali, line 116: This is not related to this change. This was never being validated because test case issue.

releasenotes/notes/33405.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+issue:
+  - 33405
+releaseNotes:
+  - |
+    **Fixed** a bug where multiple gateways on same port with SIMPLE and PASSTHROUGH modes is not working correctly.
```
shamsher31, line 8:
```suggestion
    **Fixed** a bug where multiple gateways on the same port with `SIMPLE` and `PASSTHROUGH` modes are not working correctly.

```

pilot/pkg/model/gateway.go
```diff
@@ -231,10 +231,11 @@ func MergeGateways(gateways []gatewayWithInstances) *MergedGateway {
 							log.Warnf("TLS server without TLS options %s %s", gatewayName, s.String())
 							continue
 						}
-						if mergedServers[serverPort] == nil {
-							mergedServers[serverPort] = &MergedServers{Servers: []*networking.Server{s}}
-						} else {
-							mergedServers[serverPort].Servers = append(mergedServers[serverPort].Servers, s)
+						for port := range mergedServers {
+							if port.Number == serverPort.Number {
+								mergedServers[port].Servers = append(mergedServers[port].Servers, s)
```
howardjohn, line 236: I am a bit concerned about this. In buildGatewayListeners, we use the key.Protocol field, which will now be wrong here, as with .Bind. Also, its n^2 and folks have O(10k) gateways so it may be a scalability issue

ramaraochavali:
@howardjohn changed the gateway listener build logic to handle this case. PTAL

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -59,7 +64,7 @@ func (configgen *ConfigGeneratorImpl) buildGatewayListeners(builder *ListenerBui
 
 	actualWildcard, _ := getActualWildcardAndLocalHost(builder.node)
 	errs := istiomultierror.New()
-	listeners := make([]*listener.Listener, 0)
+	mutableopts := make(map[string]mutableListenerOpts)
```
howardjohn, line 67: can you indicate what the key is?

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -124,54 +129,75 @@ func (configgen *ConfigGeneratorImpl) buildGatewayListeners(builder *ListenerBui
...
-						server, mergedGateway.GatewayNameForServer[server])
-					filterChainOpts = append(filterChainOpts, tcpChainOpts...)
-					for i := 0; i < len(tcpChainOpts); i++ {
-						filterChains = append(filterChains, istionetworking.FilterChain{ListenerProtocol: istionetworking.ListenerProtocolTCP})
+					// This is the case of TCP or PASSTHROUGH.
+					// Check if we have already built a listener for this - this happens when there are multiple gateways
+					// one with SIMPLE TLS and another with PASSTHROUGH. In that case we should append the filter chains
+					// to the existing listener.
+					mops, exists := mutableopts[lname]
+					if exists {
```
howardjohn, line 165: This seems a bit odd that only in this spot we are messing with other listeners, and everywhere else we just create a big list of filter chains. I  am a bit concerned this doesn't actually handle all cases. For example, what if we have HTTP and GRPC ports? don't we have the same issue? And why do we have the code here, but not for HTTPS - can't the same happen in that case?

ramaraochavali: Yes. Agree that it is bit odd that is handled in this case. I was n't able to wrap my head around all possibilities. But thinking about it a bit more, we can make this generic. I have changed. I think we will have to add more tests in future.

howardjohn: I think te allowed setup are:
* N HTTP servers (megred into a single filtre chain)
* N TLS/HTTPS servers and 0 or 1 TCP server
* 1 TCP server
Is that accurate?

ramaraochavali: Yes. It seems correct.

ramaraochavali:
@howardjohn added more test cases to cover scenarios. Seems to work fine. Let me know if you would like any more cases to be added.

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -114,64 +120,69 @@ func (configgen *ConfigGeneratorImpl) buildGatewayListeners(builder *ListenerBui
...
+			mutable = &MutableListener{
+				MutableObjects: istionetworking.MutableObjects{
+					// Note: buildListener creates filter chains but does not populate the filters in the chain; that's what
+					// this is for.
+					FilterChains: newFilterChains,
+				},
+			}
+			mutableopts[lname] = mutableListenerOpts{mutable: mutable, opts: opts}
+		} else {
+			mopts.opts.filterChainOpts = append(mopts.opts.filterChainOpts, opts.filterChainOpts...)
```
howardjohn, line 183: There is a small bug here. It definitely exists in current master, not sure if it did in 1.9. The iteration of MergedServers is randomized order due to golang maps, so we may get different filter chains. This doesn't matter to envoy logic, but *might* trigger incorrect behavior in some places like envoy diffing, istioctl diffing, istiod caching (not a concern now or in the near future, but in theory...)

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -184,18 +195,21 @@ func (configgen *ConfigGeneratorImpl) buildGatewayListeners(builder *ListenerBui
...
 		}
-
+	}
+	listeners := make([]*listener.Listener, 0)
+	for _, ml := range mutableopts {
+		ml.mutable.Listener = buildListener(*ml.opts, core.TrafficDirection_OUTBOUND)
 		// Filters are serialized one time into an opaque struct once we have the complete list.
-		if err := mutable.build(opts); err != nil {
-			errs = multierror.Append(errs, fmt.Errorf("gateway omitting listener %q due to: %v", mutable.Listener.Name, err.Error()))
+		if err := ml.mutable.build(*ml.opts); err != nil {
```
howardjohn, line 203: nit: can we just make `ml.opts` not a pointer?

howardjohn: If not no problem

howardjohn (approved):
Looks good, I like the new changes :+1:

PR #33420, opened 2021-06-14
Add a lot more leak checks and fix some leaks

howardjohn:
First, move all leak check main to dedicate files. This makes it much easier to see which ones have leak check enabled, and also avoids accidentally deleting the check when its in a random spot in some file. Can also search for all packages without leak checker enable with: `find pilot -type d '!' -exec test -e "{}/leak_test.go" \; -exec bash -c 'echo  "$(ls {} | grep test.go | wc -l) {}"' \; | sort -n`

Next, add the leak check to a lot of important packages and fix a lot of leaks. All of these are about just not stopping properly/

### DIFF
--- a/pilot/pkg/bootstrap/server_test.go
+++ b/pilot/pkg/bootstrap/server_test.go
@@ -380,7 +380,7 @@ func TestIstiodCipherSuites(t *testing.T) {
 		},
 	}
 
-	for _, c := range cases {
+	for _, c := range cases[:1] {
 		t.Run(c.name, func(t *testing.T) {
 			configDir, err := ioutil.TempDir("", "TestIstiodCipherSuites")
 			if err != nil {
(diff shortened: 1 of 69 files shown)


pilot/pkg/bootstrap/server_test.go
```diff
@@ -380,7 +380,7 @@ func TestIstiodCipherSuites(t *testing.T) {
 		},
 	}
 
-	for _, c := range cases {
+	for _, c := range cases[:1] {
```
stevenctl, line 383: not intended to merge?

howardjohn: thanks!

pilot/pkg/serviceregistry/kube/controller/controller_test.go
```diff
@@ -162,6 +162,7 @@ func TestServices(t *testing.T) {
 				t.Fatalf("GetService(%q) => %s, should not exist", missing, svc.Hostname)
 			}
 		})
+		return
```
stevenctl, line 165: same here?

douglas-reid (approved):
The changes to the wg-policies-and-telemetry-maintainers owned files LGTM.

lei-tang (approved):

stevenctl (approved):

PR #33426, opened 2021-06-14
autoregistration: allow bootstrap node to override locality

S-Chan:
During testing of VM autoregistration, we discovered that it is not
possible to override the WorkloadGroup template's locality from the
bootstrap node locality. The core issue is that istiod initializes the
proxy struct from the bootstrap node before passing it to the
autoregistration controller; however, the node's locality is not set on
the proxy at that time. This is fixed by passing the bootstrap node
along to autoregistration code in addition to the proxy.

Change-Id: Id9b1d2f0cc5d2cf9416670c26b0a490d302ea039
Reviewed-on: https://gerrit.musta.ch/c/public/istio/+/1391
Reviewed-by: Weibo He <weibo.he@airbnb.com>
Reviewed-by: Jungho Ahn <jungho.ahn@airbnb.com>

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

@hzxuzhonghu @howardjohn @stevenctl

### DIFF
--- a/pilot/pkg/controller/workloadentry/workloadentry_controller.go
+++ b/pilot/pkg/controller/workloadentry/workloadentry_controller.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff"
+	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	"github.com/gogo/protobuf/types"
 	"golang.org/x/time/rate"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -252,7 +253,7 @@ func setConnectMeta(c *config.Config, controller string, conTime time.Time) {
 	delete(c.Annotations, DisconnectedAtAnnotation)
 }
 
-func (c *Controller) RegisterWorkload(proxy *model.Proxy, conTime time.Time) error {
+func (c *Controller) RegisterWorkload(proxy *model.Proxy, node *core.Node, conTime time.Time) error {
 	if !features.WorkloadEntryAutoRegistration || c == nil {
 		return nil
 	}
@@ -266,14 +267,14 @@ func (c *Controller) RegisterWorkload(proxy *model.Proxy, conTime time.Time) err
 	c.adsConnections[makeProxyKey(proxy)]++
 	c.mutex.Unlock()
 
-	if err := c.registerWorkload(entryName, proxy, conTime); err != nil {
+	if err := c.registerWorkload(entryName, proxy, node, conTime); err != nil {
 		log.Errorf(err)
 		return err
 	}
 	return nil
 }
 
-func (c *Controller) registerWorkload(entryName string, proxy *model.Proxy, conTime time.Time) error {
+func (c *Controller) registerWorkload(entryName string, proxy *model.Proxy, node *core.Node, conTime time.Time) error {
 	wle := c.store.Get(gvk.WorkloadEntry, entryName, proxy.Metadata.Namespace)
 	if wle != nil {
 		lastConTime, _ := time.Parse(timeFormat, wle.Annotations[ConnectedAtAnnotation])
@@ -301,7 +302,7 @@ func (c *Controller) registerWorkload(entryName string, proxy *model.Proxy, conT
 		return fmt.Errorf("auto-registration WorkloadEntry of %v failed: cannot find WorkloadGroup %s/%s",
 			proxy.ID, proxy.Metadata.Namespace, proxy.Metadata.AutoRegisterGroup)
 	}
-	entry := workloadEntryFromGroup(entryName, proxy, groupCfg)
+	entry := workloadEntryFromGroup(entryName, proxy, node, groupCfg)
 	setConnectMeta(entry, c.instanceID, conTime)
 	_, err := c.store.Create(*entry)
 	if err != nil {
@@ -598,7 +599,7 @@ func mergeLabels(labels ...map[string]string) map[string]string {
 
 var workloadGroupIsController = true
 
-func workloadEntryFromGroup(name string, proxy *model.Proxy, groupCfg *config.Config) *config.Config {
+func workloadEntryFromGroup(name string, proxy *model.Proxy, node *core.Node, groupCfg *config.Config) *config.Config {
 	group := groupCfg.Spec.(*v1alpha3.WorkloadGroup)
 	entry := group.Template.DeepCopy()
 	entry.Address = proxy.IPAddresses[0]
@@ -619,8 +620,8 @@ func workloadEntryFromGroup(name string, proxy *model.Proxy, groupCfg *config.Co
 	if proxy.Metadata.Network != "" {
 		entry.Network = proxy.Metadata.Network
 	}
-	if proxy.Locality != nil {
-		entry.Locality = util.LocalityToString(proxy.Locality)
+	if node.Locality != nil {
+		entry.Locality = util.LocalityToString(node.Locality)
 	}
 	if proxy.Metadata.ProxyConfig != nil && proxy.Metadata.ProxyConfig.ReadinessProbe != nil {
 		annotations[status.WorkloadEntryHealthCheckAnnotation] = "true"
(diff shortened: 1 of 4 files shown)


pilot/pkg/xds/ads.go
```diff
@@ -568,7 +568,7 @@ func (s *DiscoveryServer) initializeProxy(node *core.Node, con *Connection) erro
 	proxy := con.proxy
 	// this should be done before we look for service instances, but after we load metadata
 	// TODO fix check in kubecontroller treat echo VMs like there isn't a pod
-	if err := s.WorkloadEntryController.RegisterWorkload(proxy, con.Connect); err != nil {
+	if err := s.WorkloadEntryController.RegisterWorkload(proxy, node, con.Connect); err != nil {
```
howardjohn, line 571: so seems like we are passing `node` here since we haven't yet computed proxy.Locality (L580 and on). Would it work to just re-arrange the order?

S-Chan: IIUC no, because L580 gets proxy.Locality from the service instances, which is derived from the autoregistered WorkloadEntry. Hence the comment ```// this should be done before we look for service instances, but after we load metadata```

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -619,8 +620,8 @@ func workloadEntryFromGroup(name string, proxy *model.Proxy, groupCfg *config.Co
 	if proxy.Metadata.Network != "" {
 		entry.Network = proxy.Metadata.Network
 	}
-	if proxy.Locality != nil {
-		entry.Locality = util.LocalityToString(proxy.Locality)
+	if node.Locality != nil {
+		entry.Locality = util.LocalityToString(node.Locality)
```
hzxuzhonghu, line 624: This is right in case in pilot-agent, it is set correctly. I am not sure it does.

S-Chan: You mean the node's locality may not always be set? In that case, the locality of the WorkloadGroup template (if any) would be used, right?

pilot/pkg/xds/ads.go
stevenctl: Makes sense. Wondering about this comment:

https://github.com/istio/istio/blob/245872d64a87675a40c542efb8374a3126900a3a/pilot/pkg/xds/ads.go#L580-L581

1. Why do we expect this? Isn't the point that different services may have different locality? 
2. If this expectation is true, why does it break in the case for this PR.

pilot/pkg/controller/workloadentry/workloadentry_controller.go
howardjohn: The workload group in all other cases, configures the proxy directly I think? And then is passed to istiod via XDS.

So this seems different than that pattern, by dynamically reading it from the WG. Should we instead make the agent get the locality from the WG in the istioctl workgroup command?

pilot/pkg/xds/ads.go
S-Chan: Thanks for looking @stevenctl.

Re 1:

```
// ServiceInstance represents an individual instance of a specific version
// of a service. It binds a network endpoint (ip:port), the service
// description (which is oblivious to various versions) and a set of labels
// that describe the service version associated with this instance.
//
// Since a ServiceInstance has a single IstioEndpoint, which has a single port,
// multiple ServiceInstances are required to represent a workload that listens
// on multiple ports.
```
Source: https://github.com/airbnb/istio/blob/master/pilot/pkg/model/service.go#L210-L217

IIUC, a proxy is associated with one or more ServiceInstances, but proxy should only have a single IP, so all the ServiceInstances associated would also have the same IP.

Re 2:

My understanding is that in the case of Autoregistration, ServiceInstance is derived from 1) the auto-registered WorkloadEntry, and 2) the Service/ServiceEntry selecting the WorkloadEntry. So before autoregistration finishes creating the WorkloadEntry, there are no associated ServiceInstances with the proxy, and thus no locality to set.

pilot/pkg/controller/workloadentry/workloadentry_controller.go
S-Chan: We're not currently using the istioctl workload commands. Our use case is 1 WorkloadGroup that acts as a template for WorkloadEntries spanning multiple localities, so we need to override it dynamically. I'd argue that it is currently already possible to override fields from WorkloadGroup template dynamically, e.g. labels, annotations, network and readinessprobe.

pilot/pkg/xds/ads.go
S-Chan: Actually, I think it's possible for a proxy's ServiceInstances to have different IPs, but it seems reasonable that a single running proxy should not be associated with multiple localities, since locality is meant (IMO) to model physical proximity rather than logical topology.

pilot/pkg/controller/workloadentry/workloadentry_controller.go
hzxuzhonghu: Is it possible for users to have no WLG?

What we want is an end-to-end solution, this is not complete. But maybe add a followup?

pilot/pkg/xds/ads.go
stevenctl: This shouldn't matter, since the ServiceInstance won't be there until the WorkloadEntry is created.

pilot/pkg/controller/workloadentry/workloadentry_controller.go
S-Chan: Hmm the code paths changed in this PR are just for autoregistration, so a WorkloadGroup is expected: https://github.com/istio/istio/blob/67362755b7ca63ff1442d627a733627ed5c75df4/pilot/pkg/controller/workloadentry/workloadentry_controller.go#L260-L265

Happy to discuss any potential followups that should be added.

pilot/pkg/xds/ads.go
S-Chan: Yep. So I think the expectation of "all instances [of a proxy] to have ... the same locality" still holds after this PR. Let me know if I misunderstood your questions.

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -252,7 +253,7 @@ func setConnectMeta(c *config.Config, controller string, conTime time.Time) {
 	delete(c.Annotations, DisconnectedAtAnnotation)
 }
 
-func (c *Controller) RegisterWorkload(proxy *model.Proxy, conTime time.Time) error {
+func (c *Controller) RegisterWorkload(proxy *model.Proxy, node *core.Node, conTime time.Time) error {
```
hzxuzhonghu, line 256: node can be get from Proxy

hzxuzhonghu:
@S-Chan Can you update, so we can make it in 1.11

S-Chan:
@hzxuzhonghu thanks for the tip - updated.

hzxuzhonghu (approved):
LGTM

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -619,8 +619,8 @@ func workloadEntryFromGroup(name string, proxy *model.Proxy, groupCfg *config.Co
 	if proxy.Metadata.Network != "" {
 		entry.Network = proxy.Metadata.Network
 	}
-	if proxy.Locality != nil {
-		entry.Locality = util.LocalityToString(proxy.Locality)
+	if proxy.XdsNode.Locality != nil {
+		entry.Locality = util.LocalityToString(proxy.XdsNode.Locality)
```
hzxuzhonghu, line 623: better add a comment, that proxy.Locality is unset when register workloadEntry

S-Chan: Done.

Monkeyanator (approved):

PR #33338, opened 2021-06-08
Add caching of CDS responses

howardjohn:
I have placed a hold to allow all 3 "assignees" to review (and anyone interested) before merging, as this is a fairly complex change

Before:

After:

Cache hit rate is 95%+:

A single cluster is ~900B typically. With 60k max entries, this results in at most 60mb overhead.

Istiod is using ~6GB memory for this case so that is only 1% extra memory usage, well worth the CPU gains

PR #33431, opened 2021-06-15
Refactor cluster patch

hzxuzhonghu:

Now this is the general way to do cluster patch, like listener patch.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


howardjohn (changes requested):
We used to do things this way and it was intentionally changed to support https://github.com/istio/istio/pull/33338. I would prefer we don't revert or we will not be able to cache CDS

hzxuzhonghu:
They should not conflict, the cache key used by cds is not changed.

howardjohn:
We need to apply the patches to individual clusters, not all clusters in
aggregate. The cache works on a per-cluster basis not a collection of
clusters

PR #33435, opened 2021-06-15
cache blackhole and passthrough clusters

hzxuzhonghu:
These two clusters only depend on meshconfig.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


pilot/pkg/networking/core/v1alpha3/configgen.go
```diff
@@ -24,16 +24,24 @@ type ConfigGeneratorImpl struct {
...
 }
 
 // Called when mesh config is changed.
-func (configgen *ConfigGeneratorImpl) MeshConfigChanged(_ *meshconfig.MeshConfig) {
+func (configgen *ConfigGeneratorImpl) MeshConfigChanged(mesh *meshconfig.MeshConfig) {
 	accessLogBuilder.reset()
+
+	// Reset blackhole and passthrough clusters
+	cb := NewClusterBuilder(nil, &model.PushContext{Mesh: mesh})
+	configgen.CDSCache.SetBlackHoleCluster(cb.buildBlackHoleCluster())
```
howardjohn, line 45: I think this can race with BuildClusters. Not a -race detected race, but we can accidentally revert the changes.

howardjohn (commented):
Slightly worried about the complexity of this for only caching 2 clusters. Its not that bad now but will get even more complex once we fix the race? Maybe we can do it with https://github.com/istio/istio/pull/33338?

Or at least a subset of that PR, changing the BuildClusters to return discovery.Resource so we can skip marshaling in the configgen?

pilot/pkg/networking/core/v1alpha3/configgen.go
hzxuzhonghu: There is a lock in CDSCache.

hzxuzhonghu:
I will test  #33338 to have a look at what's the performance improve

pilot/pkg/networking/core/v1alpha3/configgen.go
howardjohn: Yes its not a data race but you compute the result THEN acquire the lock. Without computing the resource inside the lock you may write out of order. See the "token" stuff in XdsCache it was done to solve similar issue

howardjohn: the concern is: compute v1, compute V2, write V2, write v1

ramaraochavali:
I think better to do benchmark to see if the complexity is worth. I have tried in the past, but it did not give significant benefits so left it.  Is n't this automatically handled by cluster caching?

hzxuzhonghu:
Yes,  if we have all cds cache, then it is not needed.

PR #33437, opened 2021-06-15
Make the productpage app compatible with IPv6 on Linux

djannot:
[ ] Docs
[ ] Installation
[ X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

[X ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

### DIFF
--- a/samples/bookinfo/src/productpage/productpage.py
+++ b/samples/bookinfo/src/productpage/productpage.py
@@ -425,6 +425,8 @@ if __name__ == '__main__':
 
     p = int(sys.argv[1])
     logging.info("start at port %s" % (p))
-    # Python does not work on an IPv6 only host
-    # https://bugs.python.org/issue10414
-    app.run(host='0.0.0.0', port=p, debug=True, threaded=True)
+    # Make it compatible with IPv6 if Linux
+    if sys.platform == "linux":
+        app.run(host='::', port=p, debug=True, threaded=True)
+    else:
+        app.run(host='0.0.0.0', port=p, debug=True, threaded=True)


samples/bookinfo/src/productpage/productpage.py
```diff
@@ -425,6 +425,8 @@ def flush(self):
 
     p = int(sys.argv[1])
     logging.info("start at port %s" % (p))
-    # Python does not work on an IPv6 only host
-    # https://bugs.python.org/issue10414
-    app.run(host='0.0.0.0', port=p, debug=True, threaded=True)
+    # Make it compatible with IPv6 if Linux
+    if sys.platform == "linux":
```
linsun, line 429: would this break ipv4?  if not, should we do it on all platforms?

djannot: The previous way was only working on IPv4 on any platform.
Now, it works with IPv4 and IPv6 on linux. And IPv4 on any other platform.

craigbox:
Sorry for letting this one slip through.  Is there anywhere else that needs to be changed (that's the only service you hit from the outside?)

djannot:
@craigbox I've just signed the CLA

craigbox:
The commit author was "denisjannot@macbook-pro-de-denis.home".  You'll have to do something like this: https://gist.github.com/trey/9588090

djannot:
@craigbox done

craigbox:
Did you push it?  (You may need to force)

If that doesn't work I can manually override the CLA as you've signed it.

djannot:
Sorry, it should be good now

craigbox:
manually verified CLA

PR #33450, opened 2021-06-15
Improve delegating hostnames in k8s gateway

howardjohn:
* Fix bug with conflicting port names in gateway
* Always insert metadata.name label, to support old k8s versions
* Fix empty hostname in route to properly mean "all hostnames"
* Add test that handles all 3 of the above

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -155,6 +155,24 @@ func isRouteMatch(cfg config.Config, gateway config.Meta,
...
+const NamespaceNameLabel = "kubernetes.io/metadata.name"
+
+func toNamespaceSet(name string, labels map[string]string) klabels.Labels {
+	// If namespace label is not set, implicitly insert it to support older Kubernetes versions
+	if labels[NamespaceNameLabel] == name {
+		// Already set, avoid copies
+		return klabels.Set(labels)
+	}
+	// First we need a copy to not modify the underlying object
+	ret := make(map[string]string, len(labels))
```
stevenctl, line 168: very nit: `len + 1`?

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -1023,7 +1045,7 @@ func buildListener(obj config.Config, l k8s.Listener, listenerIndex int) (*istio
 			// TODO currently we 1:1 support protocols in the API. If this changes we may
 			// need more logic here.
 			Protocol: string(l.Protocol),
-			Name:     fmt.Sprintf("%v-%v-gateway-%s-%s", strings.ToLower(string(l.Protocol)), l.Port, obj.Name, obj.Namespace),
+			Name:     fmt.Sprintf("%d-gateway-%s-%s", listenerIndex, obj.Name, obj.Namespace),
```
stevenctl, line 1048: Seems a little less descriptive – the `host` -> `hostname` seems to give this mapping a more human-readable way. Any reason not to include both?

stevenctl (approved):

stevenctl:
lgtm – just an aesthetic comment

pilot/pkg/config/kube/gateway/conversion.go
howardjohn: This should be entirely internal, so I am not sure the name matters too much. I think hostname may not be distinct and may have some characters like `*` that could cause issues?

stevenctl: I figured the purpose was debugging – in which case the index is enough.

PR #33456, opened 2021-06-15
bootstrap: add patching and an integration test

### LINKED ISSUE #26830: Bootstrap patch API
There has been frequent request to be able to patch boostrap in a generic way, ala `EnvoyFilterAPI for boostrap`

Use cases
1. Add new static clusters. (some clusters are `required` to be static)
2. Add/Update/modify stats configuration.
3. Add or update `runtime` configuration.

Existing solutions.

## sidecar.istio.io/bootstrapOverride ##

use `env=ISTIO_BOOTSTRAP_OVERRIDE` or `sidecar.istio.io/bootstrapOverride`  to point to custom_bootstrap.json
   The contents of `custom_bootstrap.json` are then merged into bootstrap proto.
    
    This solution cannot be used to change the defaults in an upgradeable way.
    It is possible to change the `istio-sidecar-injector` config map, but it does not survive upgrades.

## Overmount /var/lib/istio/envoy/envoy_bootstrap_tmpl.json ##
You can change the full bootstrap template, but even this is not upgrade safe.

Ideas
1. Sidecar injector watches EnvoyFilter#bootstrap or a new CR and sends the patches as `ENV` at the time of injection.
2. Istiod does boostrap generation where it can apply and validate bootstrap and send.

@kyessenov

kyessenov:
Signed-off-by: Kuat Yessenov <kuat@google.com>

Fixes #26830


pilot/pkg/xds/bootstrapds.go
```diff
@@ -54,9 +56,23 @@ func (e *BootstrapGenerator) Generate(proxy *model.Proxy, _ *model.PushContext,
...
 	}, model.DefaultXdsLogDetails, nil
 }
+
+func (e *BootstrapGenerator) applyPatches(bs *bootstrapv3.Bootstrap, proxy *model.Proxy, push *model.PushContext) *bootstrapv3.Bootstrap {
+	patches := push.EnvoyFilters(proxy)
+	if patches == nil {
+		return bs
+	}
+	for _, patch := range patches.Patches[networking.EnvoyFilter_BOOTSTRAP] {
+		if patch.Operation == networking.EnvoyFilter_Patch_MERGE {
```
mandarjog, line 73: Error when other operations are used, or is it already covered in validation ?

tests/integration/telemetry/stats/prometheus/customizemetrics/customize_metrics_test.go
```diff
@@ -212,6 +232,26 @@ func setupEnvoyFilter(ctx resource.Context) error {
...
 	}
 
+	// enable custom tag in the stats
+	bootstrapPatch := `
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: bootstrap-tag
+spec:
+  configPatches:
```
mandarjog, line 242: Can we also add a test for `stats_matcher` ? We do have specific annotations but would it be better to have BDS cover everything ?

mandarjog (approved):
lgtm sans comments.

pilot/pkg/xds/bootstrapds.go
```diff
@@ -54,9 +56,23 @@ func (e *BootstrapGenerator) Generate(proxy *model.Proxy, _ *model.PushContext,
...
 	}
+	bs = e.applyPatches(bs, proxy, push)
 	return model.Resources{
 		&discovery.Resource{
 			Resource: util.MessageToAny(bs),
 		},
 	}, model.DefaultXdsLogDetails, nil
 }
+
+func (e *BootstrapGenerator) applyPatches(bs *bootstrapv3.Bootstrap, proxy *model.Proxy, push *model.PushContext) *bootstrapv3.Bootstrap {
```
howardjohn, line 67: I think we need safety guards here like https://github.com/istio/istio/blob/980b8f6bbeb205c7d471cd34b16b64ef6beb17cd/pilot/pkg/networking/core/v1alpha3/envoyfilter/rc_patch.go#L36-L39

Otherwise arbitrary inputs can crash Istio

tests/integration/telemetry/stats/prometheus/customizemetrics/customize_metrics_test.go
kyessenov: We could use EnvoyFilter for stats inclusions. I just didn't want to create a brand-new integration test since it requires a new echo deployment. Do you want to just check envoy stats for presence of some extra tag with the right bootstrap patch?

pilot/pkg/xds/bootstrapds.go
kyessenov: Validation is not checking for matching operation/applyTo. Error is not an option since we have to return something. Is logging OK here?

pilot/pkg/model/push_context.go
```diff
@@ -1627,10 +1627,12 @@ func (ps *PushContext) initEnvoyFilters(env *Environment) error {
 
 // EnvoyFilters return the merged EnvoyFilterWrapper of a proxy
 func (ps *PushContext) EnvoyFilters(proxy *Proxy) *EnvoyFilterWrapper {
+	fmt.Printf("MATCHING ENVOY FILTERS: %#v\n", proxy)
```
howardjohn, line 1630: cleanup

kyessenov: Yeah, sorry. I'm debugging why envoy filters are not passing through pilot.

pilot/pkg/xds/bootstrapds.go
kyessenov: Done.

howardjohn (approved):

pilot/pkg/xds/bootstrapds.go
mandarjog: Can we add it to the validating webhook? a separate PR is fine. 
Logging is ok here.

mandarjog: Do we have any precedent for incrementing metric for validation errors in pilot like this?

kyessenov: Validation is more of a hassle than it's worth IMHO. Merge is the only operation that makes sense with bootstrap, so it's a fairly obvious mistake not to use Merge.

tests/integration/telemetry/stats/prometheus/customizemetrics/customize_metrics_test.go
mandarjog: Yes that should suffice. Someone is going to use it that way.

pilot/pkg/xds/bootstrapds.go
kyessenov: Added a metric increment for malformed bootstrap patch.

tests/integration/telemetry/stats/prometheus/customizemetrics/customize_metrics_test.go
kyessenov: Let's do that in a follow up. This customize metric test is rather hard to set up right (and it took awhile to debug it), so let's not complicate it further.

mandarjog: Lgtm.

Issue #32539, opened 2021-04-29
Service Entries do not Respect subjectAltNames


Kmoneal:
**Bug description**

SubjectAltNames in Service Entry does not behave as expected. When providing SubjectAltNames, the certificate received from the specified host does not check that any SubjectAltNames matches the SAN in the certificate. As a result, certificates are only checked against the host name they are received from.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

When SubjectAltNames is used in Service Entries, certificates received by the specified host must contain at least one of the SANs listed or the connection will be refused.

**Steps to reproduce the bug**

Apply the yaml below and use curl or a similar command to `en.wikipedia.org/wiki/Main_Page`. The connection should be terminated but is not due to `subjectAltNames: - www.google.com` not being used in the [Service Entry](https://istio.io/latest/docs/reference/config/networking/service-entry/).
```
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
  name: wikipedia
spec:
  hosts:
  - "en.wikipedia.org"
  ports:
  - number: 80
    name: http
    protocol: HTTP
    targetPort: 443
  resolution: DNS
  subjectAltNames:
  - "www.google.com"
---
apiVersion: networking.istio.io/v1beta1
kind: DestinationRule
metadata:
  name: wikipedia
spec:
  host: en.wikipedia.org
  trafficPolicy:
    tls:
      mode: SIMPLE
```

**Version** (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version --short` if you used Helm)

*istio version*
```
client version: 1.9.1
control plane version: 1.9.1
data plane version: 1.9.1
```
*kubectl*
```
Client Version: v1.20.4-dirty
Server Version: v1.19.6-eks-49a6c0
```

**How was Istio installed?**

istioctl

**Environment where the bug was observed (cloud vendor, OS, etc)**

Clound environment (EKS)

Additionally, please consider running `istioctl bug-report` and attach the generated cluster-state tarball to this issue.
Refer [cluster state archive](http://istio.io/help/bugs/#generating-a-cluster-state-archive) for more details.

Kmoneal:
[proxy-config.txt](https://github.com/istio/istio/files/6400397/proxy-config.txt)

[...]

PR #33472, opened 2021-06-16
Env Var allowing proxies to utilize OS CA Cert

Kmoneal:
* Introduce VERIFY_CERTIFICATE_AT_CLIENT as a bool env var
* Select an existing CA cert bundle from a set of known file locations
* If bool set and DR does not set CaCertificates, replace the proxies' CA cert with an OS CA cert found in the file system

Address istio/istio #32539

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


Kmoneal:
Described in https://docs.google.com/document/d/100Jvwjr9KkURiF2HtiOtRlAON8TqgskUyLfT17_xoRM/edit#heading=h.xw1gqgyqs5b this is the initial Env Var portion to be expanded on.

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -470,6 +471,24 @@ func selectTrafficPolicyComponents(policy *networking.TrafficPolicy) (
...
+		"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
+		"/etc/pki/tls/cacert.pem",                           // OpenELEC
+		"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
+		"/etc/ssl/cert.pem",                                 // Alpine Linux
+		"/usr/local/etc/ssl/cert.pem",                       // FreeBSD
+
+	}
+	if features.VerifyCertAtClient && tls.CaCertificates == "" {
+		for _, cert := range certFiles {
+			if _, err := os.Stat(cert); err == nil {
```
howardjohn, line 486: This needs to be done in the proxy, not the control plane. The control plane can run on a completely distinct environment

howardjohn (changes requested):

pilot/pkg/features/pilot.go
```diff
@@ -477,6 +477,12 @@ var (
 	// New behavior (true): we create listener 0.0.0.0_8080 and route http.8080. This has no conflicts; routes are 1:1 with listener.
 	UseTargetPortForGatewayRoutes = env.RegisterBoolVar("PILOT_USE_TARGET_PORT_FOR_GATEWAY_ROUTES", true,
 		"If true, routes will use the target port of the gateway service in the route name, not the service port.").Get()
+
+	VerifyCertAtClient = env.RegisterBoolVar(
```
jacob-delgado, line 481: This needs to get passed down. It should be stored in a struct (haven't looked at that code in a while).

pilot/pkg/model/authentication.go
```diff
@@ -312,7 +313,24 @@ func SdsCertificateConfigFromResourceName(resource string) (SdsCertificateConfig
 	} else if strings.HasPrefix(resource, "file-root:") {
 		filesString := strings.TrimPrefix(resource, "file-root:")
 		split := strings.Split(filesString, ResourceSeparator)
-		if len(split) != 1 {
+		if filesString == "system" && len(split) != 1 {
```
jacob-delgado, line 316: Does `SdsCertificateConfigFromResourceName` have a unit test? We need some level of testing before it can be merged.

pilot/pkg/model/authentication.go
```diff
@@ -312,7 +313,24 @@ func SdsCertificateConfigFromResourceName(resource string) (SdsCertificateConfig
...
+			var certFiles = []string{
```
howardjohn, line 317: I feel like this needs to be under some agent/ folder. I was pretty surprised to find SdsCertificateConfigFromResourceName was only ever called from the agent currently, which makes me think its fairly fragile and may lead to an import from Istiod, which would be incorrect.

pilot/pkg/model/authentication.go
```diff
@@ -312,7 +313,24 @@ func SdsCertificateConfigFromResourceName(resource string) (SdsCertificateConfig
...
+				"/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora/RHEL 6
+				"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
+				"/etc/pki/tls/cacert.pem",                           // OpenELEC
+				"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
+				"/etc/ssl/cert.pem",                                 // Alpine Linux
+				"/usr/local/etc/ssl/cert.pem",                       // FreeBSD
+			}
+
+			for _, cert := range certFiles {
+				if _, err := os.Stat(cert); err == nil {
```
howardjohn, line 328: small perf nit: should we wrap this in a `once` like go does? Its fairly slow to check the filesystem on every call here

pilot/pkg/model/authentication.go
jacob-delgado: This is an exported method that doesn't have any dependencies within the package. I'd consider moving it to a file within a folder used by the agent. If someone else needs this functionality at a later time, it could be moved to a shared folder.

pkg/istio-agent/agent_test.go
```diff
@@ -158,6 +158,39 @@ func TestAgent(t *testing.T) {
 			return a
 		}).Check(t, cfg.GetRootResourceName(), cfg.GetResourceName())
 	})
+	t.Run("File system certs", func(t *testing.T) {
+		// User sets FileMountedCerts. They also need to set ISTIO_META_TLS_CLIENT* to specify the
+		// file paths. CA communication is disabled. mTLS is always used for authentication with
+		// Istiod, never JWT.
```
jacob-delgado, line 164: Change the comments to reflect this test

pkg/istio-agent/agent_test.go
```diff
@@ -535,6 +569,30 @@ func copyCerts(t *testing.T, dir string) {
 	}
 }
 
+func copyCertsWithOSRootCA(t *testing.T, dir string) {
+	caRootPath := security.GetOSRootPath()
+	if caRootPath == "" {
+		t.Fatal("OS CA Root Cert could not be found.")
+	}
+	caRootCertSlices := strings.Split(caRootPath, "/")
+	caRootCert := caRootCertSlices[len(caRootCertSlices)-1]
```
jacob-delgado, line 578: `caRootCert := filepath.Base(caRootPath)`

pkg/istio-agent/agent_test.go
```diff
@@ -535,6 +569,30 @@ func copyCerts(t *testing.T, dir string) {
...
+	if caRootCert == "" {
+		t.Fatal("OS CA Root Cert couldn't be found.")
+	}
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		t.Fatal(err)
```
jacob-delgado, line 583: In cases when it errors out with `err`, you should use `t.Fatalf("why this is bad", err)`

pkg/istio-agent/agent_test.go
```diff
@@ -535,6 +569,30 @@ func copyCerts(t *testing.T, dir string) {
...
+	}
```
jacob-delgado, line 584: I don't think you have to call `os.MkdirAll` as `dir = makedirtmp()` was previously called creating this directory

pkg/security/security.go
```diff
@@ -344,3 +348,23 @@ func ExtractRequestToken(req *http.Request) (string, error) {
 
 	return "", fmt.Errorf("no bearer token exists in HTTP authorization header")
 }
+
+func GetOSRootPath() string {
+	// Get and store the OS CA certificate path
```
jacob-delgado, line 353: point to where this exists in go, https://golang.org/src/crypto/x509/root_linux.go

pkg/security/security.go
```diff
@@ -344,3 +348,23 @@ func ExtractRequestToken(req *http.Request) (string, error) {
...
+		"/etc/ssl/cert.pem",                                 // Alpine Linux
+		"/usr/local/etc/ssl/cert.pem",                       // FreeBSD
+	}
+
+	for _, cert := range certFiles {
+		if _, err := os.Stat(cert); err == nil {
+			return cert
+		}
+	}
+	return ""
```
jacob-delgado, line 369: Either throw an error or put a logging statement as a WARN

releasenotes/notes/33472.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: feature
+area: security
+issue:
+  - 33472
+releaseNotes:
+  - |
+    **Added** VerifyCertAtClient Environmental Variable to pilot. Setting VerifyCertAtClient to `true` will verify server certificates using the OS CA certificate.
```
jacob-delgado, line 8:
```suggestion
    **Added** `VERTIFY_CERT_AT_CLIENT` environmental variable to pilot. Setting `VERTIFY_CERT_AT_CLIENT` to `true` will verify server certificates using the OS CA certificates when not using a `DestinationRule` `caCertificates` field.
```

pkg/security/security.go
```diff
@@ -344,3 +348,23 @@ func ExtractRequestToken(req *http.Request) (string, error) {
 
 	return "", fmt.Errorf("no bearer token exists in HTTP authorization header")
 }
+
+func GetOSRootPath() string {
```
jacob-delgado, line 352: GetOSRootPath -> GetOSCARootFilePath

pkg/istio-agent/agent_test.go
```diff
@@ -158,6 +158,39 @@ func TestAgent(t *testing.T) {
...
+	t.Run("File system certs", func(t *testing.T) {
+		// User sets FileMountedCerts. They also need to set ISTIO_META_TLS_CLIENT* to specify the
+		// file paths. CA communication is disabled. mTLS is always used for authentication with
+		// Istiod, never JWT.
+		dir := mktemp()
+		copyCertsWithOSRootCA(t, dir)
+
+		osRootPath := security.GetOSRootPath()
+		caRootCertSlices := strings.Split(osRootPath, "/")
+		caRootCert := caRootCertSlices[len(caRootCertSlices)-1]
```
jacob-delgado, line 170: `caRootCert := filepath.Base(caRootPath)`

pkg/istio-agent/agent_test.go
```diff
@@ -158,6 +158,39 @@ func TestAgent(t *testing.T) {
 			return a
 		}).Check(t, cfg.GetRootResourceName(), cfg.GetResourceName())
 	})
+	t.Run("File system certs", func(t *testing.T) {
```
jacob-delgado, line 161: Let's be more descriptive in what this is testing

pilot/pkg/networking/core/v1alpha3/cluster_test.go
```diff
@@ -2346,3 +2346,57 @@ func TestTelemetryMetadata(t *testing.T) {
 		})
 	}
 }
+
+func TestSystemCACert(t *testing.T) {
+	features.VerifyCertAtClient = true
```
jacob-delgado, line 2351:
```
defer func() {
 features.VerifyCertAtClient = false
}
```

pilot/pkg/networking/core/v1alpha3/cluster_test.go
```diff
@@ -2346,3 +2346,57 @@ func TestTelemetryMetadata(t *testing.T) {
...
+			},
+		},
+		Tls: &networking.ClientTLSSettings{
+			CaCertificates: "",
+		},
+	}
+	selectTrafficPolicyComponents(policy)
+	features.VerifyCertAtClient = false
+	if policy.Tls.CaCertificates != "file-root:system" {
+		t.Error("Could not set OS CA Cert: ")
```
jacob-delgado, line 2366:
```suggestion
		t.Error("Could not set OS CA Cert")
```

pilot/pkg/networking/core/v1alpha3/cluster_test.go
```diff
@@ -2346,3 +2346,57 @@ func TestTelemetryMetadata(t *testing.T) {
 		})
 	}
 }
+
+func TestSystemCACert(t *testing.T) {
```
jacob-delgado, line 2350: Let's combine them and use subtests, see https://blog.golang.org/subtests

security/pkg/nodeagent/sds/sdsservice.go
```diff
@@ -200,12 +204,17 @@ func (s *sdsservice) Close() {
...
-func toEnvoySecret(s *security.SecretItem) *tls.Secret {
+func toEnvoySecret(s *security.SecretItem, caRootPath string) *tls.Secret {
 	secret := &tls.Secret{
 		Name: s.ResourceName,
 	}
-
-	cfg, ok := model.SdsCertificateConfigFromResourceName(s.ResourceName)
+	cfg := nodeagentutil.SdsCertificateConfig{}
+	ok := false
+	if s.ResourceName == "file-root:system" {
```
jacob-delgado, line 213: Let's make this a constant that is referenced across. Magic strings can be hard to debug.

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -460,7 +463,8 @@ func (sc *SecretManagerClient) readFileWithTimeout(path string) ([]byte, error)
 	}
 }
 
-func (sc *SecretManagerClient) generateFileSecret(resourceName string) (bool, *security.SecretItem, error) {
+func (sc *SecretManagerClient) generateFileSecret(resourceName string, caRootPath string) (bool, *security.SecretItem, error) {
```
jacob-delgado, line 466: Should we pass in caRootPath or set it in the previous function and use the getter here?

security/pkg/nodeagent/cache/secretcache_test.go
```diff
@@ -603,3 +604,30 @@ func TestProxyConfigAnchors(t *testing.T) {
...
+	fakeCACli, err := mock.NewMockCAClient(time.Hour, false)
+	if err != nil {
+		t.Fatalf("Error creating Mock CA client: %v", err)
+	}
+	opt := &security.Options{}
+
+	fakePlugin := mock.NewMockTokenExchangeServer(nil)
+	opt.TokenExchanger = fakePlugin
+
+	sc := createCache(t, fakeCACli, func(resourceName string) {}, security.Options{})
```
jacob-delgado, line 620: Let's also use table driven tests for this too

jacob-delgado, line 620: nit: move `certPath := security.GetOSRootPath()` to here for locality reasons

security/pkg/nodeagent/util/util.go
```diff
@@ -105,3 +106,62 @@ func OutputKeyCertToDir(dir string, privateKey, certChain, rootCert []byte) erro
...
+
+// IsRootCertificate returns true if this config represents a root certificate config.
+func (s SdsCertificateConfig) IsRootCertificate() bool {
+	return s.CaCertificatePath != ""
+}
+
+// IsKeyCertificate returns true if this config represents key certificate config.
+func (s SdsCertificateConfig) IsKeyCertificate() bool {
+	return s.CertificatePath != "" && s.PrivateKeyPath != ""
+}
```
jacob-delgado, line 144: Look for a common area between pilot and the agent. It may not exist. We should keep to DRY.

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2778,3 +2780,358 @@ func TestBuildAutoMtlsSettings(t *testing.T) {
...
+						CaCertificates: constants.DefaultRootCert,
+						Mode:           networking.ClientTLSSettings_SIMPLE,
+					},
+				},
+			},
+			expectedSubsetClusters: []*cluster.Cluster{},
+		},
+	}
+
+	for caseNumber, tt := range cases {
```
jacob-delgado, line 3064: for _, tt := range cases

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2778,3 +2780,358 @@ func TestBuildAutoMtlsSettings(t *testing.T) {
...
+					Tls: &networking.ClientTLSSettings{
+						CaCertificates: "",
+						Mode:           networking.ClientTLSSettings_SIMPLE,
+					},
+				},
+			},
+			expectedSubsetClusters: []*cluster.Cluster{},
+		},
+		{
+			name:        "destination rule with CaCertificates",
```
jacob-delgado, line 3039: use name instead of caseNumber

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2778,3 +2780,358 @@ func TestBuildAutoMtlsSettings(t *testing.T) {
...
+
+	cases := []struct {
+		name                   string
+		cluster                *cluster.Cluster
+		clusterMode            ClusterMode
+		service                *model.Service
+		port                   *model.Port
+		networkView            map[network.ID]bool
+		destRule               *networking.DestinationRule
+		expectedSubsetClusters []*cluster.Cluster
```
jacob-delgado, line 3012: `expectedCaCertificateName string`

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2778,3 +2780,358 @@ func TestBuildAutoMtlsSettings(t *testing.T) {
...
+			if err != nil {
+				t.Errorf("Could not parse destination rule: %v", err)
+			}
+			dr := &networking.DestinationRule{}
+			err = json.Unmarshal(byteArray, &dr)
+			if err != nil {
+				t.Errorf("Could not unmarshal destination rule: %v", err)
+			}
+			ca := dr.TrafficPolicy.Tls.CaCertificates
+			if caseNumber == 0 && ca != "file-root:system" {
```
jacob-delgado, line 3130:
```
if ca != tt.expectedCaCertificateName {
  t.Errorf("%v: got unexpected caCertitifcates field. Expected (%v), received (%v)", tt.name, tt.expectedCaCertificateName, ca)
```

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2778,3 +2780,358 @@ func TestBuildAutoMtlsSettings(t *testing.T) {
...
+			cg := NewConfigGenTest(t, TestOptions{
+				ConfigPointers: []*config.Config{cfg},
+				Services:       []*model.Service{tt.service},
+			})
+			cg.MemRegistry.WantGetProxyServiceInstances = instances
+			cb := NewClusterBuilder(cg.SetupProxy(nil), &model.PushRequest{Push: cg.PushContext()}, nil)
+
+			ec := NewMutableCluster(tt.cluster)
+			destRule := cb.req.Push.DestinationRule(cb.proxy, tt.service)
+			subsetClusters := cb.applyDestinationRule(ec, tt.clusterMode, tt.service, tt.port, tt.networkView, destRule)
```
jacob-delgado, line 3102: nit: // ACT

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2778,3 +2780,358 @@ func TestBuildAutoMtlsSettings(t *testing.T) {
...
+							UseClientProtocol: true,
+						},
+					},
+					Tls: &networking.ClientTLSSettings{
+						CaCertificates: constants.DefaultRootCert,
+						Mode:           networking.ClientTLSSettings_SIMPLE,
+					},
+				},
+			},
+			expectedSubsetClusters: []*cluster.Cluster{},
```
jacob-delgado, line 3060: I would recommend removing this if not needed

security/pkg/nodeagent/sds/sdsservice.go
Kmoneal: Constant is in `pkg/security/security.go` with other constants

releasenotes/notes/33472.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: feature
+area: security
+issue:
+  - 33472
+releaseNotes:
+  - |
+    **Added** `VERTIFY_CERT_AT_CLIENT` environmental variable to pilot. Setting `VERTIFY_CERT_AT_CLIENT` to `true` will verify server certificates using the OS CA certificates when not using a `DestinationRule` `caCertificates` field [Issue #33472](github.com/istio/istio/issues/33472).
```
incfly, line 8: do we still use pilot? or should be istiod?

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -735,3 +744,16 @@ func (sc *SecretManagerClient) UpdateConfigTrustBundle(trustBundle []byte) error
 func (sc *SecretManagerClient) mergeConfigTrustBundle(rootCert []byte) []byte {
 	return pkiutil.AppendCertByte(sc.getConfigTrustBundle(), rootCert)
 }
+
+func (sc *SecretManagerClient) GetCARootPath() string {
```
incfly, line 748: Doc string for public method.

incfly: same for others.

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2715,3 +2718,149 @@ func TestBuildAutoMtlsSettings(t *testing.T) {
 		})
 	}
 }
+
+func TestApplyDestinationRuleOSCACert(t *testing.T) {
+	features.VerifyCertAtClient = true
```
incfly, line 2723: to be complete the test case should test the feature disable behavior as well, test case can have this as part of the dimension.

```golang
{
  enableVerifyCertAtClient: true/false,
  drContainsCaCertificate: true/false
}
```

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -497,7 +500,13 @@ func (sc *SecretManagerClient) generateFileSecret(resourceName string) (bool, *s
 		// Check if the resource name refers to a file mounted certificate.
 		// Currently used in destination rules and server certs (via metadata).
 		// Based on the resource name, we need to read the secret from a file encoded in the resource name.
-		cfg, ok := model.SdsCertificateConfigFromResourceName(resourceName)
+		cfg := security.SdsCertificateConfig{}
+		ok := false
+		if resourceName == security.FileRootSystemCACert {
```
incfly, line 505: It seems to be all these special casing for the `caRootPath` do not follow the existing SDS case switching.

Instead of adding a new parameter to the function, (and 90% callers will now have to pass an empty string), can we model the caroot path as a special `resourceName` instead?


add @shankgan who is more familiar in this code.

pilot/cmd/pilot-agent/options/security.go
```diff
@@ -52,6 +52,7 @@ func NewSecurityOptions(proxyConfig *meshconfig.ProxyConfig, stsPort int, tokenM
 		SecretRotationGracePeriodRatio: secretRotationGracePeriodRatioEnv,
 		STSPort:                        stsPort,
 		CertSigner:                     certSigner.Get(),
+		CARootPath:                     security.GetOSRootFilePath(),
```
incfly, line 55: also it's confused to me that

following the code path, ca root path is only set here in the options, at the startup time, right? and then later on, this field is passed to GenerateSecret() in sdsservice.go without ever changing.

why we have to parameterize it in this case? would it work if we provide an `sync.Once` initialization and later on always use that value? My understanding is that in the startup, we can do init to detect what ca cert is available and keep using it from that point on.

pilot/pkg/model/authentication.go
incfly: +1. And the value seems not neeed to passed around, and parameteize and special casing it. It's fair to assume the proxy image's filesystem stay the same during the pod running time and keep use that value.

releasenotes/notes/33472.yaml
Kmoneal: This should be istiod

pilot/cmd/pilot-agent/options/security.go
Kmoneal: Getting the OS CA file path was moved to init but with that I don't think `sync.Once` was needed but I would be happy to add it if you find it necessary. This removes the added parameter in `GenerateSecret()`.

releasenotes/notes/33472.yaml
Kmoneal: Fixed to istiod

security/pkg/nodeagent/cache/secretcache.go
Kmoneal: Doc added to public function and variable. GetCARootPath and SetCARootPath no longer needed with change to get the OS CA file path on init

security/pkg/nodeagent/cache/secretcache.go
Kmoneal: Made the resourceName check into it's own case and removed the parameter added to the function.

jacob-delgado (approved):
Kenan demoed this to me and we walked through the various test cases proposed in the RFC using badssl.com and its various subdomains (self-signed.badssl.com, untrusted-root.badssl.com, expired.badssl.com).

releasenotes/notes/33472.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: feature
+area: security
+issue:
+  - 33472
+releaseNotes:
+  - |
+    **Added** `VERTIFY_CERT_AT_CLIENT` environmental variable to istiod. Setting `VERTIFY_CERT_AT_CLIENT` to `true` will verify server certificates using the OS CA certificates when not using a `DestinationRule` `caCertificates` field [Issue #33472](github.com/istio/istio/issues/33472).
```
howardjohn, line 8: can they also set `caCertificates: system` explicitly?

If so (which I think they should?) we should add a warning to the webhook when we detect an explicit path to use `system` instead. Definitely should be a followup though, not in this PR. Lets get this in first

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -493,11 +496,20 @@ func (sc *SecretManagerClient) generateFileSecret(resourceName string) (bool, *s
 			// Adding cert is sufficient here as key can't change without changing the cert.
 			sc.addFileWatcher(cf.CertificatePath, resourceName)
 		}
+	case resourceName == security.FileRootSystemCACert:
+		sdsFromFile = true
+		if sc.caRootPath != "" {
+			if sitem, err = sc.generateRootCertFromExistingFile(sc.caRootPath, resourceName, false); err == nil {
+				sc.addFileWatcher(sc.caRootPath, resourceName)
```
howardjohn, line 503: note to others: addFileWatcher dedupes so we will have at most 1 watch :+1:

howardjohn (approved):

releasenotes/notes/33472.yaml
Kmoneal: They would be able to use `system` and it will treat it the same.

Issue #33475, opened 2021-06-16
TestDefaultInPlaceUpgradeFromPreviousMinorRelease test failures


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-122_istio_postsubmit/1405197196374052864
https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-122_istio_postsubmit/1405183852791468032

cc @jacob-delgado

howardjohn:
Oh I looked at the test results wrong, this was broken on v1.22.0-alpha.0 -> alpha.3 upgrade

PR #33479, opened 2021-06-16
Skip tests broken in 1.22

howardjohn:
* Ingress old API version was removed, bump it up one versions we can
* legacy-unknown signer is removed. This disables 4 test suites. THESE
FEATURES ARE BROKEN - it is not just a testing issue.
* Upgrade tests are broken on 1.22, so disabled. This is not a big deal
right now, but when we update presubmit to 1.22 it likely will be.

I have tagged everyone that has features they worked on broken.

Fixes https://github.com/istio/istio/issues/33475

### DIFF
--- a/tests/integration/pilot/ingress_test.go
+++ b/tests/integration/pilot/ingress_test.go
@@ -221,16 +221,16 @@ func TestIngress(t *testing.T) {
 				apiVersion = "v1"
 			}
 
-			ingressClassConfig := `
+			ingressClassConfig := fmt.Sprintf(`
 apiVersion: networking.k8s.io/%s
 kind: IngressClass
 metadata:
   name: istio-test
 spec:
-  controller: istio.io/ingress-controller`
+  controller: istio.io/ingress-controller`, apiVersion)
 
 			ingressConfigTemplate := `
-apiVersion: networking.k8s.io/v1
+apiVersion: networking.k8s.io/v1beta
 kind: Ingress
 metadata:
   name: %s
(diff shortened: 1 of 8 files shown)


tests/integration/pilot/ingress_test.go
```diff
@@ -216,16 +216,21 @@ func TestIngress(t *testing.T) {
...
 			})
 
+			apiVersion := "v1beta1"
+			if t.Clusters().Default().MinKubeVersion(19) {
+				apiVersion = "v1"
+			}
+
 			ingressClassConfig := `
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/%s
```
lei-tang, line 225: Is "%s" a typo in the literal string "networking.k8s.io/%s"?

howardjohn: no, we sprintf it later. I guess we could probably just do it here though

tests/integration/pilot/multi_version_revision_test.go
```diff
@@ -199,4 +199,8 @@ func skipIfK8sVersionUnsupported(t framework.TestContext) {
 	if !t.Clusters().Default().MinKubeVersion(16) {
 		t.Skipf("k8s version not supported for %s (<%s)", t.Name(), "1.16")
 	}
+	// Kubernetes 1.22 drops support for a number of legacy resources, so we cannot install the old versions
+	if !t.Clusters().Default().MaxKubeVersion(21) {
```
lei-tang, line 203: !t.Clusters().Default().MaxKubeVersion(21) checks the minor k8s version is at most 21 but does not check the major k8s version. If the k8s version is 2.1 (major-version=2, minor-version=1), what is the behavior of !t.Clusters().Default().MaxKubeVersion(21)?

howardjohn: Kubernetes will never have a 2.0 according to their project. If they do, we can change it.

lei-tang (approved):

Monkeyanator (approved):

stevenctl (approved):

PR #33514, opened 2021-06-18
Refine logic on FindRootCAForXX

Ruoyu-y:
* Remove duplicated code on finding root CA in istio-agent
* Make the well-known cert location a const

Signed-off-by: Ruoyu.Ying <ruoyu.ying@intel.com>

Found some duplicated code in istio-agent. So made some enhancement on that.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[x] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/istio-agent/agent.go
+++ b/pkg/istio-agent/agent.go
@@ -82,6 +82,9 @@ const (
 	// this may be replaced with ./etc/certs, if a root-cert.pem is found, to
 	// handle secrets mounted from non-citadel CAs.
 	CitadelCACertPath = "./var/run/secrets/istio"
+
+	// defaultCACertPath is the well-known location of ssl CA cert path on ubuntu
+	defaultCACertPath = "/etc/ssl/certs/ca-certificates.crt"
 )
 
 const (
@@ -221,7 +224,7 @@ func (a *Agent) generateNodeMetadata() (*model.Node, error) {
 		// Envoy only supports load from file. If we want to use system certs, use best guess
 		// To be more correct this could lookup all the "well known" paths but this is extremely \
 		// unlikely to run on a non-debian based machine, and if it is it can be explicitly configured
-		provCert = "/etc/ssl/certs/ca-certificates.crt"
+		provCert = defaultCACertPath
 	}
 	var pilotSAN []string
 	if a.proxyConfig.ControlPlaneAuthPolicy == mesh.AuthenticationPolicy_MUTUAL_TLS {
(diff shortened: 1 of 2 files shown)


Ruoyu-y:
@myidpt @howardjohn I am new to the Istio community and want to help refine the code. Can you guys help with the review? Thanks!

pkg/istio-agent/agent.go
```diff
@@ -520,38 +493,66 @@ func fileExists(path string) bool {
...
+	// Using K8S - this is likely incorrect, may work by accident.(https://github.com/istio/istio/issues/22161)
+	if a.secOpts.PilotCertProvider == constants.CertProviderKubernetes {
+		return k8sCAPath, true, nil
+	}
+
+	// Using defaultRootCertFilePath while using the custom root certificate mounted in a well known location
+	if a.secOpts.PilotCertProvider == constants.CertProviderCustom {
+		return security.DefaultRootCertFilePath, true, nil
+	}
+
```
shankgan, line 534: Isn' there supposed to be a case for certProvidernone?

pkg/istio-agent/agent.go
```diff
@@ -470,45 +473,15 @@ func (a *Agent) Close() {
 // In addition it deals with the case the XDS server is on port 443, expected with a proper cert.
 // /etc/ssl/certs/ca-certificates.crt
 func (a *Agent) FindRootCAForXDS() (string, error) {
-	var rootCAPath string
-
-	if a.cfg.XDSRootCerts == security.SystemRootCerts {
-		// Special case input for root cert configuration to use system root certificates
-		return "", nil
-	} else if a.cfg.XDSRootCerts != "" {
```
shankgan, line 478: can this option be safely removed?

pkg/istio-agent/agent.go
```diff
@@ -520,38 +493,66 @@ func fileExists(path string) bool {
 
 // Find the root CA to use when connecting to the CA (Istiod or external).
 func (a *Agent) FindRootCAForCA() (string, error) {
-	var rootCAPath string
-
-	if a.cfg.CARootCerts == security.SystemRootCerts {
-		return "", nil
-	} else if a.cfg.CARootCerts != "" {
```
shankgan, line 527: Is this option no longer being used?

Ruoyu-y: I am not sure if this one is still in use. If we can confirm that this option is no longer being used, we can remove this part.

Ruoyu-y:
@shankgan BTW,  i would also like to confirm if the case using k8s as cert provider is still valid.

Ruoyu-y:
@shankgan Can you help with the answer? Thanks!

shankgan:
> @shankgan BTW, i would also like to confirm if the case using k8s as cert provider is still valid.
 Could be removed some time eventually, but for now, this is still in use

Ruoyu-y:
> > @shankgan BTW, i would also like to confirm if the case using k8s as cert provider is still valid.
> > Could be removed some time eventually, but for now, this is still in use

Thanks. Then i will keep this option and remove the one to use SystemRootCerts.

Ruoyu-y:
@shankgan Any more comments on this?

PR #33515, opened 2021-06-18
short path check for SubsetOf

silenceshell:
Add short path check for `SubsetOf`, mainly for cases when `len(that) < len(i)`, to avoid meaningless map iteration.

[ ] Docs
[ ] Installation
[ ] Networking
[X] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/config/labels/collection.go
+++ b/pkg/config/labels/collection.go
@@ -18,7 +18,7 @@ package labels
 // collection of labels
 type Collection []Instance
 
-// HasSubsetOf returns true if the input labels are a super set of one labels in a
+// HasSubsetOf returns true if the input labels are a superset of one labels in a
 // collection or if the tag collection is empty
 func (c Collection) HasSubsetOf(that Instance) bool {
 	if len(c) == 0 {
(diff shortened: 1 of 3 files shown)


howardjohn (commented):
Do we have any benchmarks?

silenceshell:
Here is an example: https://goplay.space/#zu4jL7qFz9z
I create two maps whose size is 100,000 and 100,001, and the 100,000 elements of the two maps are the same.
So the original SubsetOf will iterate 100,000 times(cost 4ms), while the new SubsetOfNew has just to check length(cost 156ns), as the bigger map will never be subset of the smaller map.

```
result false, original SubsetOf cost time 4.506658ms
result false, new SubsetOf cost time 156ns
2021-06-19 00:16:19.866003 +0800 CST m=+0.097447895
```

FYI: goplay.space treate time.Now as a fixed value(2009-11-10 23:00:00 +0000 UTC m=+0.000000001), so we should run the code in local env.

silenceshell:
@howardjohn Could you help to review this PR? Many thanks!

hzxuzhonghu (approved):

PR #33522, opened 2021-06-18
Improve merging of proxy config proxyMetadata

howardjohn:
This is useful for pod annotation overrides and for
shared_user_mesh_config. Basically if the top level sets a proxy
metadata, which may be for critical fields like CA_ADDRESS, anyone below
it will override them completely which will often lead to breakages.

This allows a deeper merge

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/releasenotes/notes/pc-merge-metadata.yaml
+++ b/releasenotes/notes/pc-merge-metadata.yaml
@@ -1,0 +1,6 @@
+apiVersion: release-notes/v2
+kind: feature
+area: installation
+releaseNotes:
+- |
+  **Improved** the `meshConfig.defaultConfig.proxyMetadata` field to do a deep megre when overriden, rather than replacing all values.
--- a/releasenotes/notes/pc-merge-metadata.yaml
+++ b/releasenotes/notes/pc-merge-metadata.yaml
@@ -3,4 +3,4 @@ kind: feature
 area: installation
 releaseNotes:
 - |
-  **Improved** the `meshConfig.defaultConfig.proxyMetadata` field to do a deep megre when overriden, rather than replacing all values.
+  **Improved** the `meshConfig.defaultConfig.proxyMetadata` field to do a deep merge when overriden, rather than replacing all values.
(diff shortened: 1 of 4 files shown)


costinm (approved):

releasenotes/notes/pc-merge-metadata.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: release-notes/v2
+kind: feature
+area: installation
+releaseNotes:
+- |
+  **Improved** the `meshConfig.defaultConfig.proxyMetadata` field to do a deep megre when overriden, rather than replacing all values.
```
Monkeyanator, line 6:
```suggestion
  **Improved** the `meshConfig.defaultConfig.proxyMetadata` field to do a deep merge when overriden, rather than replacing all values.
```

Monkeyanator (approved):

jacob-delgado (approved):

ericvn (approved):

nrjpoddar (commented):
Can this lead to upgrade breakages due to change in behavior?

mandarjog:
In theory yes it can. Previously we used replace semantics which means if the users knows what they are doing they can do everything including removal of prior keys.

Merge semantics are more natural, but if have provided proxy config with the knowledge of replace semantics then it will break you.

we need to identify specific keys as examples to see if this is academic or a real breakage.

nrjpoddar:
@howardjohn can you look at my and @mandarjog comments? I want to make sure we cover any upgrade issues that may arise due to this change.

howardjohn:
@nrjpoddar I feel about the same as Mandar. Like any behavioral change, included pure bug fixes, this *could* impact a user. However, I find it unlikely. The most common case of mesh-wide proxyMetadata is to configure control plane access (CA address, and the root cert configs - required for external Istiod). The problem that arises is then that essentially for external istiod, you cannot set the proxy metadata on a per-pod basis (since you will break control plane access).

I have never seen any case of someone intentionally unsetting a variable. In the case that they are, this is something still support, they just need to do it explicitly rather than implicitly.

Issue #28791, opened 2020-11-11
istio-agent non-daemon mode (generate only)


costinm:
We need to be able to run istio-agent - or a smaller standalone binary - to do the init and generate the 
certificate files, but not start envoy or keep running.

This is needed for apps like Prometheus, which can use the mTLS certificate directly.

It is also needed for gRPC 'proxyless'. In this mode, we would also generate the bootstrap file for
gRPC which is similar but smaller than Envoy bootstrap.

**Describe the feature request**

**Describe alternatives you've considered**

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Additional context**

markdroth:
As we discussed, for gRPC, I think we'll actually need daemon mode, because we'll need istio-agent to continue to update the cert files as the certs rotate.  So the main thing we need is for istio-agent to generate the gRPC bootstrap file, and it might make sense to split that into a separate issue.

In the future, we can collaborate on ways to eliminate the dependency on istio-agent.  This will probably involve having a CertificateProvider plugin that can speak the Istio CA API (https://github.com/istio/api/blob/master/security/v1alpha1/ca.proto) and adding the ability to set STS call creds for the communication with the xDS server.  At that point, we can move the code to generate the gRPC bootstrap file out of istio-agent and into a standalone bootstrap generator, which users can run on their nodes before starting their application.

howardjohn:
Related: https://github.com/istio/istio/issues/28712

costinm:
Agreed - I think we'll want both 'daemon' and 'exec/cron' mode ( similar with gcloud command ).

costinm:
Updated requirement: 
- should be able to run as an XDS proxy ( performing istio-specific auth ), but without starting envoy
- instead of starting envoy, should start an app, using the reminder of the command line ( i.e. if an app
is started with "/usr/local/bin/myApp --foo --bar", we can use "istio-agent /usr/local/bin/myApp --foo --bar"

This will be used until gRPC ( or other 'proxyless' frameworks ) support all required Istio-specific auth
methods and CA protocol. The apps will still have no proxy in the data path.

stevenctl:
I think the exact CLI needs some adjustments. We should have separate flags to say:
 1. Don't run envoy
[...]

PR #33528, opened 2021-06-18
Add CLI option to pilot-agent for exec mode

ruigulala:
Add support to start a custom command/app in pilot-agent.
The custom app can run with/without envoy proxy.
To be consistent, rely on the previously added environment
variable to control whether to enable envoy proxy.

Issue: https://github.com/istio/istio/issues/28791

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[X] Test and Release
[X] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/cmd/pilot-agent/main.go
+++ b/pilot/cmd/pilot-agent/main.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/exec"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/cobra/doc"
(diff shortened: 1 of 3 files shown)


pilot/cmd/pilot-agent/main.go
```diff
@@ -160,6 +161,27 @@ var (
 	}
 )
 
+func newExecCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "exec -- COMMAND [args...]",
```
howardjohn, line 166: Not sure I understand why a user would run this vs just running `command` directly?

costinm: The command must run after the agent is ready. If you start agent in background - you have no way to know when it's ready, and the app will likely start. Also likely you need shell - which in containerless is not possible.

pilot/cmd/pilot-agent/main.go
```diff
@@ -160,6 +161,27 @@ var (
...
+		Short: "Execute a command in pilot-agent",
+		FParseErrWhitelist: cobra.FParseErrWhitelist{
+			// Allow unknown flags for backward-compatibility.
+			UnknownFlags: true,
+		},
+		PersistentPreRunE: configureLogging,
+		RunE: func(c *cobra.Command, args []string) error {
+			cmd := exec.Command(args[0], args[1:]...)
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
```
costinm, line 176: I think what is missing here is the part where the agent initializes and starts, before the actual exec...

pilot/cmd/pilot-agent/main.go
hzxuzhonghu: Just read through the context, have the same confusion with john,  

>Also likely you need shell - which in containerless is not possible.

Why is it not possible? I understand in vm it is more likely.

pilot/cmd/pilot-agent/main.go
ruigulala: Done.

pilot/cmd/pilot-agent/main.go
ruigulala: Added the part to initialize agent based on Costin's comment.

costinm: Shell: in distroless there is no shell AFAIK. 

VMs have a shell - but there is also the use case of docker ( for example in google CloudRun or other docker exec envs ), and the image might be distroless-based.

Also there is the case for proxyless gRPC ( we support it as experimental ) - where we  want to start the app after the init is done, so the files are in place.

costinm (approved):

costinm (changes requested):
On further thought ( and trying to use this ): can you remove the extra command, and instead 
do the exec for the user app in the normal proxy command ? 

I believe users can pass "-- extra args" - and this will be available after parsing the args.

I will follow up with a change to handle the case pilot-agent runs as root, to make sure envoy is started as 1337 and the app exec can also change the UID. 

The main target is running in Docker ( like CloudRun ), possibly without iptables.

ruigulala:
Talked with @costinm , we've decided to drop this for now and revisit in the future.

```
I think we can drop it for now - we found all kind of other gaps. We are doing the same for the cloud-run-mesh ( CloudRun support for Istio )
```

Issue #33222, opened 2021-06-02
DNS Proxying metrics


lmarszal:
**Describe the feature request**
I'd like to see following metrics from agent's DNS proxy:
- total number of DNS requests (in order to calculate requests per second)
- number of requests forwarded upstream
- request duration

We might consider additional dimensions - like response code (NOERROR, NXDOMAIN) and request type (A, AAA, MX, ANY etc).

**Describe alternatives you've considered**
Alternative is to implement this telemetry directly in every app.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[x] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Additional context**
We used to have some problems with DNS in our cluster in the past. DNS Proxying greatly reduces our coredns load. What we want to achieve is to have some additional visibility into how many DNS calls application is making (on per-app basis) and how long does it take. Adding these as agent metrics is a great opportunity do implement this once for every service in our mesh.

ramaraochavali:
Makes sense. I am actually planning to add some of these metrics.

PR #33532, opened 2021-06-19
dns: add dns proxy stats

ramaraochavali:
Fixes https://github.com/istio/istio/issues/33222
[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/dns/monitoring.go
+++ b/pilot/pkg/dns/monitoring.go
@@ -27,12 +27,12 @@ func init() {
 
 var (
 	requests = monitoring.NewSum(
-		"dns_total_requests",
+		"dns_requests_total",
 		"Total number of DNS requests.",
 	)
 
 	upstreamRequests = monitoring.NewSum(
-		"dns_total_upstream_requests",
+		"dns_upstream_requests_total",
 		"Total number of DNS requests forwarded to upstream.",
 	)
 
--- a/pilot/pkg/dns/monitoring.go
+++ b/pilot/pkg/dns/monitoring.go
@@ -18,13 +18,6 @@ import (
 	"istio.io/pkg/monitoring"
 )
 
-func init() {
-	monitoring.MustRegister(requests)
-	monitoring.MustRegister(upstreamRequests)
-	monitoring.MustRegister(failures)
-	monitoring.MustRegister(requestDuration)
-}
-
 var (
 	requests = monitoring.NewSum(
 		"dns_requests_total",
@@ -47,3 +40,10 @@ var (
 		[]float64{1, 3, 5, 10, 25, 50},
 	)
 )
+
+func registerStats() {
+	monitoring.MustRegister(requests)
+	monitoring.MustRegister(upstreamRequests)
+	monitoring.MustRegister(failures)
+	monitoring.MustRegister(requestDuration)
+}
--- a/pilot/pkg/dns/monitoring.go
+++ b/pilot/pkg/dns/monitoring.go
@@ -1,0 +1,49 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dns
+
+import (
+	"istio.io/pkg/monitoring"
+)
+
+func init() {
+	monitoring.MustRegister(requests)
+	monitoring.MustRegister(upstreamRequests)
+	monitoring.MustRegister(failures)
+	monitoring.MustRegister(requestDuration)
+}
+
+var (
+	requests = monitoring.NewSum(
+		"dns_total_requests",
+		"Total number of DNS requests.",
+	)
+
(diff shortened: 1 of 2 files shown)


pilot/pkg/dns/monitoring.go
```diff
@@ -0,0 +1,49 @@
...
+func init() {
+	monitoring.MustRegister(requests)
+	monitoring.MustRegister(upstreamRequests)
+	monitoring.MustRegister(failures)
+	monitoring.MustRegister(requestDuration)
+}
+
+var (
+	requests = monitoring.NewSum(
+		"dns_total_requests",
```
howardjohn, line 30: I think prometheus standard (and openmetrics requirement) is that it ends with _total. So `dns_requests_total`.

pilot/pkg/dns/monitoring.go
```diff
@@ -0,0 +1,49 @@
...
+		"Total number of DNS requests.",
+	)
+
+	upstreamRequests = monitoring.NewSum(
+		"dns_total_upstream_requests",
```
howardjohn, line 35: @douglas-reid should this be a tag?

pilot/pkg/dns/monitoring.go
```diff
@@ -0,0 +1,49 @@
...
+		"Total number of DNS requests forwarded to upstream.",
+	)
+
+	failures = monitoring.NewSum(
+		"dns_total_upstream_failures",
+		"Total number of DNS requests forwarded to upstream.",
+	)
+
+	requestDuration = monitoring.NewDistribution(
+		"dns_upstream_request_time_ms",
```
howardjohn, line 45: cardinality may be an issue here @nrjpoddar @douglas-reid any guidance?

howardjohn (commented):
I think the idea here is good, but we should have careful review from telemetry folks @istio/wg-policies-and-telemetry-maintainers

ramaraochavali:
@mandarjog @douglas-reid can you PTAL?

ramaraochavali:
@douglas-reid can you PTAL?

pilot/pkg/dns/monitoring.go
nrjpoddar: Are these metrics going to be enabled by default when DNS proxy will be enabled or is it going to be opt in?

ramaraochavali: It is not opt-in currently. Enabled when dns proxy is on.

ramaraochavali:
@howardjohn @nrjpoddar @douglas-reid gentle ping. Can you PTAL when you get chance?

pilot/pkg/dns/monitoring.go
nrjpoddar: So if a user has enabled DNS_PROXY, then on upgrades they will see an increase in their metric storage after this change. Should we guard it and keep default off?

ramaraochavali: yes. But since DNS proxy is still experimental and the amount of metrics we are adding here are not too many - I am guessing it should be fine. However, I am not against adding a flag defaulted to false, if you think it is better

nrjpoddar: I will let others weigh in too but an opt-in option might be better

mandarjog: adding 8 bucket distribution is like adding another istio standard distribution metric sans other labels. We can probably reduce it to 6. So I don’t think cardinality is a big issue. Since dns itself is experimental we can leave this in and create an option based on feedback.

we many need to systematically add metrics opt in for everything for the internal metric generation, but does not have to be here on an adhoc basis.

ramaraochavali: Thanks I have reduced it to 6 buckets. @howardjohn can you PTAL?

nrjpoddar: SGTM.

nrjpoddar (approved):

pilot/pkg/dns/monitoring.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"istio.io/pkg/monitoring"
+)
+
+func init() {
+	monitoring.MustRegister(requests)
+	monitoring.MustRegister(upstreamRequests)
+	monitoring.MustRegister(failures)
+	monitoring.MustRegister(requestDuration)
```
hzxuzhonghu, line 25: Just one point, why register these metrics when even dns proxy is unused.

hzxuzhonghu: Hha, i am not a fan of doing everything in `init`

ramaraochavali: Fixed

hzxuzhonghu (commented):
lgtm

pilot/pkg/dns/monitoring.go
howardjohn: Should we use seconds instead of ms? See https://prometheus.io/docs/practices/naming/#base-units

pilot/pkg/dns/monitoring.go
```diff
@@ -0,0 +1,49 @@
...
+		"Total number of DNS requests.",
+	)
+
+	upstreamRequests = monitoring.NewSum(
+		"dns_upstream_requests_total",
+		"Total number of DNS requests forwarded to upstream.",
+	)
+
+	failures = monitoring.NewSum(
+		"dns_total_upstream_failures",
```
howardjohn, line 33: dns_total_upstream_failures -> dns_upstream_failures_total to follow https://prometheus.io/docs/practices/naming/#metric-names (and consistency with upstreamRequests)

Issue #33115, opened 2021-05-26
1.10 regression: idle_timeout leads to NACK


catman002:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ *] Upgrade

Upgrade to version 1.10.0, publish istio service, **istio-proxy** cannot be started, error prompt:

[root@node0 maixiaolan]# **kubectl describe pod basemq-57f9b7b96c-xvr7t**  
```
...
...
  Normal   Created    <invalid>                       kubelet            Created container istio-proxy
  Normal   Started    <invalid>                       kubelet            Started container istio-proxy
  Warning  Unhealthy  <invalid> (x2 over <invalid>)   kubelet            Readiness probe failed: Get "http://10.244.166.181:15021/healthz/ready": dial tcp 10.244.166.181:15021: connect: connection refused
  Warning  Unhealthy  <invalid> (x13 over <invalid>)  kubelet            Readiness probe failed: HTTP probe failed with statuscode: 503
```
[root@node0 maixiaolan]# **kubectl logs basemq-c6bd6f877-n2lbs -c istio-proxy -f**
```
2021-05-26T12:28:57.793386Z	info	FLAG: --concurrency="2"
2021-05-26T12:28:57.793486Z	info	FLAG: --domain="maixiaolan-test.svc.cluster.local"
2021-05-26T12:28:57.793504Z	info	FLAG: --help="false"
2021-05-26T12:28:57.793515Z	info	FLAG: --log_as_json="false"
2021-05-26T12:28:57.793525Z	info	FLAG: --log_caller=""
2021-05-26T12:28:57.793536Z	info	FLAG: --log_output_level="default:info"
2021-05-26T12:28:57.793545Z	info	FLAG: --log_rotate=""
2021-05-26T12:28:57.793555Z	info	FLAG: --log_rotate_max_age="30"
2021-05-26T12:28:57.793566Z	info	FLAG: --log_rotate_max_backups="1000"
2021-05-26T12:28:57.793577Z	info	FLAG: --log_rotate_max_size="104857600"
2021-05-26T12:28:57.793587Z	info	FLAG: --log_stacktrace_level="default:none"
2021-05-26T12:28:57.793607Z	info	FLAG: --log_target="[stdout]"
[...]

PR #33533, opened 2021-06-19
add integration test for idle timeout in DR

ramaraochavali:
[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/integration/pilot/common/routing.go
+++ b/tests/integration/pilot/common/routing.go
@@ -650,6 +650,29 @@ func useClientProtocolCases(apps *EchoDeployments) []TrafficTestCase {
 	return cases
 }
 
+// destinationRuleCases contains tests some specific DestinationRule tests.
+func destinationRuleCases(apps *EchoDeployments) []TrafficTestCase {
+	var cases []TrafficTestCase
+	client := apps.PodA
+	destination := apps.PodC[0]
+	cases = append(cases,
+		TrafficTestCase{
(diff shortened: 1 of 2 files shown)


ramaraochavali:
@howardjohn PTAL. Added integration test for DR specifying idle timeout only https://github.com/istio/istio/issues/33115 case.

tests/integration/pilot/common/routing.go
```diff
@@ -1747,6 +1770,23 @@ spec:
...
+metadata:
+  name: %s
+spec:
+  host: %s
+  trafficPolicy:
+    tls:
+      mode: DISABLE
+    connectionPool:
+      http:
+        idleTimeout: 100s
```
howardjohn, line 1785: are we actually testing the idle timeout? It seems we just make sure this doesn't break things entirely - which I guess is fine but maybe we can comment that to avoid making the test seem like its doing more than it is?

I don't have any good ideas to fully test the field

howardjohn (approved):

tests/integration/pilot/common/routing.go
ramaraochavali: No. I was not thinking of validating idle timeout behaviour.  It is meant to validate config is generated correctly. Added comment.

Issue #33418, opened 2021-06-14
`format-go` misses updatesd


howardjohn:
See https://github.com/istio/istio/pull/33417

kailun-qin:
Related golangci-lint changes: https://github.com/golangci/golangci-lint/pull/2069.

howardjohn:
I am pretty sure golangci-lint update fixed it

PR #33541, opened 2021-06-21
Add integfuzz tag to lint and format configs

kailun-qin:
Partially addresses https://github.com/istio/istio/issues/33418

Signed-off-by: Kailun Qin <kailun.qin@intel.com>

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/common/config/.golangci-format.yml
+++ b/common/config/.golangci-format.yml
@@ -13,6 +13,7 @@ run:
   deadline: 20m
   build-tags:
   - integ
+  - integfuzz
   # which dirs to skip: they won't be analyzed;
   # can use regexp here: generated.*, regexp is applied on full path;
   # default value is empty list, but next dirs are always skipped independently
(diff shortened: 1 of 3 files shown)


common/config/.golangci-format.yml
```diff
@@ -13,6 +13,7 @@ run:
   deadline: 20m
   build-tags:
   - integ
+  - integfuzz
```
ericvn, line 16: Changes in the common directory come from istio/common-files changes, else these would be overridden with the next common-files update.

common/config/.golangci.yml
```diff
@@ -13,6 +13,7 @@ run:
   deadline: 20m
   build-tags:
   - integ
+  - integfuzz
```
ericvn, line 16: Similar to above being a common-files change.

ericvn: I'm not sure if this change being merged into common-files will cause failures in the other repos as it automatically applied in the repos. We would need the main_test.go file merged prior to the common-files update.

howardjohn (approved):

PR #33542, opened 2021-06-21
do not create istio CA if PILOT_CERT_PROVIDER is kubernetes and workload certs are signed from external CA

irisdingbj:
```
pilot:
    k8s:
      env:
      - name: EXTERNAL_CA
        value: ISTIOD_RA_KUBERNETES_API
      - name: K8S_SIGNER
        value: kubernetes.io/legacy-unknown
      - name: PILOT_CERT_PROVIDER
        value: kubernetes
```
Take above config as an example, we do not need to enable ca function in istiod at all since we will delegate all certs generation(including istiod certs) to the external ca and istiod is just working as an RA.
 
Currently we will still `createIstioCA` and generate unneeded certs.  See https://github.com/istio/istio/blob/master/pilot/pkg/bootstrap/server.go#L1050 for details.

With this change, we can get below benefits: 

1. save effort for creating an needless CA server and certs --those codes are dead under this case. 
2.  Support using external signer to sign istiod certs as well : Curretly we do not support specify signer for istiod certs even if `PILOT_CERT_PROVIDER` is `kubernetes` as we will use the defaut one `kubernetes.io/legacy-unknown`.


pilot/pkg/bootstrap/certcontroller.go
```diff
@@ -129,10 +130,23 @@ func (s *Server) initDNSCerts(hostname, customHost, namespace string) error {
 
 	var certChain, keyPEM, caBundle []byte
 	var err error
-	if features.PilotCertProvider == constants.CertProviderKubernetes {
+	if s.RA != nil {
```
shankgan, line 133: Given we already have a configuration "PilotCertProvider" to designate who signs Istiod DNS certs, don't think its a good idea to derive this config from anywhere else, including s.RA. Perhaps we could say that if the signerName is configured, we will use that signer to sign DNS certificates if PilotCertProvider=Kubernetes

pilot/pkg/bootstrap/server.go
```diff
@@ -1046,15 +1040,14 @@ func (s *Server) maybeCreateCA(caOpts *caOptions) error {
...
-
-		// TODO: Issue #27606 If External CA is configured, use that to sign DNS Certs as well. IstioCA need not be initialized
-		if s.CA, err = s.createIstioCA(corev1, caOpts); err != nil {
-			return fmt.Errorf("failed to create CA: %v", err)
-		}
 		if caOpts.ExternalCAType != "" {
 			if s.RA, err = s.createIstioRA(s.kubeClient, caOpts); err != nil {
 				return fmt.Errorf("failed to create RA: %v", err)
 			}
+		} else {
```
shankgan, line 1047: don't think this is a good idea. I think if an ExternalCA is configured, we should still use it just to sign workloads certificates and not Istiod's certificate unless the PilotCertProvider is explicitly set to Kubernetes. Note that the External CA's root needs to be distributed to workloads if they are to talk to Istiod that is signed by an external CA

pilot/pkg/bootstrap/certcontroller.go
irisdingbj: uf `PilotCertProvider=Kubernetes && singerName is configured ` we will not create CA now.

pilot/pkg/bootstrap/server.go
irisdingbj: Thanks!  changed to If PilotCertProvider=Kubernetes && singerName is configured we will not create CA now.

pilot/pkg/bootstrap/certcontroller.go
```diff
@@ -129,10 +130,22 @@ func (s *Server) initDNSCerts(hostname, customHost, namespace string) error {
 
 	var certChain, keyPEM, caBundle []byte
 	var err error
-	if features.PilotCertProvider == constants.CertProviderKubernetes {
+	if features.PilotCertProvider == constants.CertProviderKubernetes && signerName != "" && s.RA != nil {
```
howardjohn, line 133: Can you add comments? Its not 100% clear why we have this complex set of logic to determine this now - so in a year it will probably be confusing

pilot/pkg/bootstrap/certcontroller.go
```diff
@@ -129,10 +130,22 @@ func (s *Server) initDNSCerts(hostname, customHost, namespace string) error {
...
+			strings.Join(names, ","), hostnamePrefix+".csr.secret", namespace, caCertFile, signerName)
+		if err != nil {
+			return fmt.Errorf("failed genrating ker cert by k8s: %v", err)
+		}
+		caBundle = s.RA.GetCAKeyCertBundle().GetRootCertPem()
+	} else if features.PilotCertProvider == constants.CertProviderKubernetes {
 		log.Infof("Generating K8S-signed cert for %v", names)
 		certChain, keyPEM, _, err = chiron.GenKeyCertK8sCA(s.kubeClient.CertificatesV1beta1().CertificateSigningRequests(),
-			strings.Join(names, ","), hostnamePrefix+".csr.secret", namespace, defaultCACertPath)
+			strings.Join(names, ","), hostnamePrefix+".csr.secret", namespace, defaultCACertPath, "")
```
howardjohn, line 148: Why no signerName here? Isn't that already a supported configuration?

pilot/pkg/bootstrap/certcontroller.go
```diff
@@ -129,10 +130,22 @@ func (s *Server) initDNSCerts(hostname, customHost, namespace string) error {
...
+	if features.PilotCertProvider == constants.CertProviderKubernetes && signerName != "" && s.RA != nil {
+		log.Infof("Generating K8S-signed cert for %v using signer %v", names, signerName)
+		caCertFile := path.Join(ra.DefaultExtCACertDir, constants.CACertNamespaceConfigMapDataName)
+		if _, err := os.Stat(caCertFile); err != nil {
+			caCertFile = defaultCACertPath
+		}
+		certChain, keyPEM, _, err = chiron.GenKeyCertK8sCA(s.kubeClient.CertificatesV1beta1().CertificateSigningRequests(),
+			strings.Join(names, ","), hostnamePrefix+".csr.secret", namespace, caCertFile, signerName)
+		if err != nil {
+			return fmt.Errorf("failed genrating ker cert by k8s: %v", err)
```
howardjohn, line 142: Some typos on this line. Also we should avoid "k8s" in user facing messages

pilot/pkg/bootstrap/certcontroller.go
```diff
@@ -129,10 +130,22 @@ func (s *Server) initDNSCerts(hostname, customHost, namespace string) error {
 
 	var certChain, keyPEM, caBundle []byte
 	var err error
-	if features.PilotCertProvider == constants.CertProviderKubernetes {
+	if features.PilotCertProvider == constants.CertProviderKubernetes && signerName != "" && s.RA != nil {
+		log.Infof("Generating K8S-signed cert for %v using signer %v", names, signerName)
+		caCertFile := path.Join(ra.DefaultExtCACertDir, constants.CACertNamespaceConfigMapDataName)
```
howardjohn, line 135: Why do we use a different ca cert file here?

tests/integration/security/external_ca_istiodcerts/main_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+func TestMain(m *testing.M) {
+	// Integration test for testing interoperability with external CA's that are integrated with K8s CSR API
+	// and sign istiod certs as well
+	// Refer to https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/
+	framework.NewSuite(m).
+		Label(label.CustomSetup).
+		RequireMinVersion(18).
+		// https://github.com/istio/istio/issues/22161. 1.22 drops support for legacy-unknown signer
+		RequireMaxVersion(21).
```
howardjohn, line 80: Seems not great to add a new test that we already know is broken on Kubernetes latest version..

howardjohn (commented):
Sorry for all of the comments, it is because I am not sure what problem we are trying to solve. Can you expand on the PR description and comments in the code?

pilot/pkg/bootstrap/certcontroller.go
irisdingbj:
```
  pilot:
    k8s:
      env:
      - name: EXTERNAL_CA
        value: ISTIOD_RA_KUBERNETES_API
      - name: K8S_SIGNER
        value: kubernetes.io/legacy-unknown
      - name: PILOT_CERT_PROVIDER
        value: kubernetes
```
Take above config as an example,  we do not need to enable ca function in istiod at all since we will delegate all certs generation to the external ca and istiod is just working as an RA.

howardjohn:
What benefits do we get?

irisdingbj:
we can skip creating a no useful CA and initializing all the no-needed certs. This also opens a door to let user define different signer to for istiod certs.

tests/integration/security/external_ca_istiodcerts/main_test.go
irisdingbj: In the future after we figure out how to support different k8s signers we can enable these cases back.

pilot/pkg/bootstrap/certcontroller.go
irisdingbj: If we are using RA, we should use the root cert from RA.

pilot/pkg/bootstrap/certcontroller.go
irisdingbj: This is to keep our current behaviour, when we define `PILOT_CERT_PROVIDER` as `kubernetes` , Kubernetes will use `kubernetes.io/legacy-unknown` as the signer by default.

irisdingbj:
@howardjohn Please see the updated description for this PR. Thanks!

hzxuzhonghu (commented):
generally lgtm

irisdingbj:
@shankgan A ping on this, Thanks!

pilot/pkg/bootstrap/certcontroller.go
shankgan: This check isn't entirely accurate. It is still acceptable to use the Kubernetes cert provider even if s.RA is nil. Also, this creates a lot of backward compatibility issues by FORCING a signer name to be present, since K8S_SIGNER is still an experimental config whereas using K8s to sign certs is already GA. I would suggest removing the newly added checks

shankgan: In general, IMO, s.RA (and its corresponding config which is EXTERNAL_CA) should only be used to determine the workload cert signer, not the control-plane cert signer. @irisdingbj - WDYT?

pilot/pkg/bootstrap/certcontroller.go
shankgan: To clarify: If we use ANY external K8s signer, this means that the CA cert file will be different. 
Note: We still need a good way to propagate the external ca to all the workloads so that they can connect with Istiod.

shankgan:
@irisdingbj - I think this is already being done, but to double confirm - From the perspective of config grouping, can we keep these two separate - EXTERNAL_CA (env variable indicating that the RA code path is triggered for data-plane mTLS workload certificates), and PILOT_CERT_PROVIDER (which indicates who is signing the control-plane certificates). These two code paths can then share the K8S_SIGNER config to figure out how to route the CSR requests.

pilot/pkg/bootstrap/server.go
```diff
@@ -1046,10 +1045,11 @@ func (s *Server) maybeCreateCA(caOpts *caOptions) error {
 			}
 		}
 		// May return nil, if the CA is missing required configs - This is not an error.
-
-		// TODO: Issue #27606 If External CA is configured, use that to sign DNS Certs as well. IstioCA need not be initialized
-		if s.CA, err = s.createIstioCA(corev1, caOpts); err != nil {
-			return fmt.Errorf("failed to create CA: %v", err)
+		noCA := features.PilotCertProvider == constants.CertProviderKubernetes && caOpts.ExternalCASigner != ""
```
shankgan, line 1048: I would suggest having this logic when registering/running the CA grpc service. The logic is getting a little too complex and we should avoid  having this in too many places. Could we have this in a function?

tests/integration/security/external_ca_istiodcerts/main_test.go
shankgan: Not sure what the purpose of both the tests are.

 There is already a test that uses chiron (legacy-unknown signer) to sign DNS certs in tests/integration/security/ chiron. 

There is also a test that does e2e reachability check in tests/integration/security/ chiron/external_ca that does mTLS reachability checks when the workload certs are signed by K8s default signer

irisdingbj: Yes. The case you mentioned is to cover 1 scenario for each: one is to cover DNS cert and the other is to cover workload certs, So both of them have CA server enabled.   This new test  is to cover the `AND` scenario:  both istiod DNS cert and workload certs are signed from legacy-unknown signer and CA server is disabled.

pilot/pkg/bootstrap/certcontroller.go
irisdingbj: Agree. We can extend this in the future to allow user define signer for istiod certs as well via introducing a new Env Var:
`PILOT_CERT_SIGNER`

pilot/pkg/bootstrap/certcontroller.go
irisdingbj: fixed

pilot/pkg/bootstrap/server.go
irisdingbj: sure, move this into disableCAServer()

irisdingbj:
> @irisdingbj - I think this is already being done, but to double confirm - From the perspective of config grouping, can we keep these two separate - EXTERNAL_CA (env variable indicating that the RA code path is triggered for data-plane mTLS workload certificates), and PILOT_CERT_PROVIDER (which indicates who is signing the control-plane certificates). These two code paths can then share the K8S_SIGNER config to figure out how to route the CSR requests.



> @irisdingbj - I think this is already being done, but to double confirm - From the perspective of config grouping, can we keep these two separate - EXTERNAL_CA (env variable indicating that the RA code path is triggered for data-plane mTLS workload certificates), and PILOT_CERT_PROVIDER (which indicates who is signing the control-plane certificates). These two code paths can then share the K8S_SIGNER config to figure out how to route the CSR requests.

Yep, this makes  the code logic more clear. I have also removed the integration test.

irisdingbj:
@howardjohn @shankgan @hzxuzhonghu PTAL, Thanks!

hzxuzhonghu (approved):
lg

pilot/pkg/bootstrap/server.go
```diff
@@ -960,7 +959,7 @@ func (s *Server) initIstiodCerts(args *PilotArgs, host string) error {
 
 // createPeerCertVerifier creates a SPIFFE certificate verifier with the current istiod configuration.
 func (s *Server) createPeerCertVerifier(tlsOptions TLSOptions) (*spiffe.PeerCertVerifier, error) {
-	if tlsOptions.CaCertFile == "" && s.CA == nil && features.SpiffeBundleEndpoints == "" {
+	if tlsOptions.CaCertFile == "" && s.CA == nil && features.SpiffeBundleEndpoints == "" && !s.DisableCAServer() {
```
shankgan, line 962: Why do we need s.CA == nil & s.DisableCAServer()? My understanding is that one of those conditions is sufficient, since s.CA == nil if s.DisableCAServer() is true?

pilot/pkg/bootstrap/server.go
```diff
@@ -1192,3 +1197,10 @@ func (s *Server) initWorkloadTrustBundle(args *PilotArgs) error {
 	log.Infof("done initializing workload trustBundle")
 	return nil
 }
+
+// DisableCAServer returns whether CA functionality is disabled in istiod.
+// It return true only if istiod certs is signed by Kubernetes and
+// workload certs are signed by external CA
+func (s *Server) DisableCAServer() bool {
```
shankgan, line 1204: nit: rename as checkDisableCa()? or isDisableCa(). Not sure if this function needs to be exported outside the bootstrap package.

irisdingbj: changed to `isDisableCa()`

pilot/pkg/bootstrap/server.go
irisdingbj: we are using `!s.isDisableCa()` so s.CA == nil is needed.

irisdingbj:
@shankgan PTAL, Thanks!

shankgan (approved):

PR #33550, opened 2021-06-21
Support OpenMetrics format for metrics merging

### LINKED ISSUE #33474: Metrics merging does not work with openmetrics format metrics
(**NOTE**: This is used to report product bugs:
  To report a security vulnerability, please visit <https://istio.io/about/security-vulnerabilities>
  To ask questions about how to use Istio, please visit <https://discuss.istio.io>)

**Bug description**

When utilizing the metrics merging functionality with an application that supports openmetrics type metrics, e.g. Java-based apps using JMX exporter, prometheus fails to parse the merged metrics it receives from the istio agent.

The end result is a scrape error similar to the following (this example is for JMX exporter's metrics):
`level=debug ts=2021-06-15T02:19:45.732Z caller=scrape.go:1149 component="scrape manager" scrape_pool=kubernetes-pods target=http://10.116.2.7:15020/stats/prometheus msg="Append failed" err="invalid metric type \"info\""`

The root cause seems to be that the [istio agent will pass through the `Accept` header for scrape requests](https://github.com/istio/istio/blob/332164ff8cfbb692636d34bb36cac17fab8de0a5/pilot/cmd/pilot-agent/status/server.go#L483), which for a prometheus scrape request will be `application/openmetrics-text; version=0.0.1,text/plain;version=0.0.4;q=0.5,*/*;q=0.1`

This causes problems for metrics exporters that do indeed support the openmetrics exposition format, e.g. JMX exporter, as based on prometheus' `Accept` header, [it will write its metrics in the openmetrics format](https://github.com/prometheus/client_java/blob/b178c5e1879e8d3d3c6e50ede8968340ba9f2488/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java#L56).

So, the scrape from the agent to the app will receive openmetrics format, which will then be merged with the envoy metrics and istio agent metrics, which are of the standard text-based prometheus format. The response from the agent to the initial scrape is then sent as type `text/plain`. The metrics are then attempted to be parsed according to the text format, which results in the above error.

[ ] Docs
[ ] Installation
[ ] Networking
[X] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

The metrics merging should be explicit that only the text based format metrics are supported (via `Accept` to applications), so the resulting merged metrics are all of valid text type.

1. Use default sidecar injection on an e.g. tomcat app with JMX exporter
2. Correctly annotate workload with prom annotations (i.e. standard metrics merging)
3. Check prometheus targets or debug log to see error

I built a simple tomcat application that uses JMX exporter and deployed it with out of the box istio.
I can share it here if desired.

```
% istioctl version --remote
client version: 1.8.6
control plane version: 1.8.6
data plane version: 1.8.6 (3 proxies)

% k version --short
Client Version: v1.20.4
Server Version: v1.19.9-gke.1900
```

lgadban:
Fixes #33474 by enhancing the metrics merging functionality to support the OpenMetrics format in addition to the standard text-based metrics format.

This is accomplished by passing through the `Accept` header on the agent's scrape request to the application metrics endpoint and respecting the metrics format returned via the `Content-Type`. If there are no app metrics, the merging will default to `FmtText`.

Additionally, some post-processing of the envoy metrics must be done, namely removing empty lines, as they are not compatible as-is with FmtOpenMetrics. Also, the application metrics must be rendered last to correctly follow the FmtOpenMetrics spec.

[ ] Docs
[ ] Installation
[ ] Networking
[X] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/cmd/pilot-agent/status/server.go
+++ b/pilot/cmd/pilot-agent/status/server.go
@@ -412,7 +412,7 @@ func (s *Server) handleStats(w http.ResponseWriter, r *http.Request) {
 			log.Errorf("failed scraping application metrics: %v", err)
 			metrics.AppScrapeErrors.Increment()
 		}
-		format = negotaiteMetricsFormat(contentType)
+		format = negotiateMetricsFormat(contentType)
 	}
 
 	if agent, err = scrapeAgentMetrics(); err != nil {
@@ -423,13 +423,13 @@ func (s *Server) handleStats(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", string(format))
 
 	// Write out the metrics
-	if _, err := w.Write(envoy); err != nil {
-		log.Errorf("failed to write envoy metrics: %v", err)
-		metrics.EnvoyScrapeErrors.Increment()
-	}
 	if _, err := w.Write(agent); err != nil {
 		log.Errorf("failed to write agent metrics: %v", err)
 		metrics.AgentScrapeErrors.Increment()
+	}
+	if _, err := w.Write(envoy); err != nil {
+		log.Errorf("failed to write envoy metrics: %v", err)
+		metrics.EnvoyScrapeErrors.Increment()
 	}
 	// App metrics must go last because if they are FmtOpenMetrics,
 	// they will have a trailing "# EOF" which terminates the full exposition
@@ -439,7 +439,7 @@ func (s *Server) handleStats(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func negotaiteMetricsFormat(contentType string) expfmt.Format {
+func negotiateMetricsFormat(contentType string) expfmt.Format {
 	mediaType, _, err := mime.ParseMediaType(contentType)
 	if err == nil && mediaType == expfmt.OpenMetricsType {
 		return expfmt.FmtOpenMetrics
--- a/pilot/cmd/pilot-agent/status/server.go
+++ b/pilot/cmd/pilot-agent/status/server.go
@@ -448,7 +448,7 @@ func negotiateMetricsFormat(contentType string) expfmt.Format {
 }
 
 func processMetrics(metrics []byte) []byte {
-	return []byte(strings.ReplaceAll(string(metrics), "\n\n", "\n"))
+	return bytes.ReplaceAll(metrics, []byte("\n\n"), []byte("\n"))
 }
 
 func scrapeAgentMetrics() ([]byte, error) {
--- a/pilot/cmd/pilot-agent/status/server.go
+++ b/pilot/cmd/pilot-agent/status/server.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"mime"
 	"net"
 	"net/http"
 	"net/http/pprof"
@@ -387,6 +388,7 @@ type PrometheusScrapeConfiguration struct {
 // the application metrics and merge them together.
 // The merge here is a simple string concatenation. This works for almost all cases, assuming the application
 // is not exposing the same metrics as Envoy.
+// This merging works for both FmtText and FmtOpenMetrics and will use the format of the application metrics
 // Note that we do not return any errors here. If we do, we will drop metrics. For example, the app may be having issues,
 // but we still want Envoy metrics. Instead, errors are tracked in the failed scrape metrics/logs.
 func (s *Server) handleStats(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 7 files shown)


pilot/cmd/pilot-agent/status/server_test.go
```diff
@@ -312,6 +312,114 @@ my_metric{app="bar"} 0
...
+					t.Fatalf("write failed: %v", err)
+				}
+			}))
+			defer envoy.Close()
+			app := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				format := expfmt.NegotiateIncludingOpenMetrics(r.Header)
+				var negotiatedMetrics string
+				if format == expfmt.FmtOpenMetrics {
+					negotiatedMetrics = appOpenMetrics
+				} else if format == expfmt.FmtText {
```
howardjohn, line 381: Are metrics in FmtText not compatible with FmtOpenMetrics? My (very loose) understanding was FmtOpenMetrics is just adding some additional legal syntax, so its more of a superset of FmtText. Is that not accurate?

lgadban: I also am not an expert, but there seems to be differences in the actual metric types in `FmtOpenMetrics` that are not present in `FmtText`. Notably, `FmtOpenMetrics` has support for [the `info` type](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info) which isn't a valid type in `FmtText`.

howardjohn: Right, what I was thinking is:

* Prometheus sets Accept=FmtText. All apps/envoy/agent must return FmtText, and everything works
* Prometheus sets Accept=FmtOpenMetrics. The app returns the new FmtOpenMetrics, but envoy returns FmtText since that is all it supports. Because FmtText is a subset of FmtOpenMetrics, prometheus interprets it all correctly

We should never have a case where prometheus sets Accept=FmtText and gets back FmtOpenMetrics(?)


Is the issue when we respond we don't properly tell Prometheus which format we actually responded with (FmtOpenMetrics)? And we should add some response header?

lgadban: Got it, I see what you are getting at. 

In theory we should be able to:

- pass through the `Accept` from Prom to the application container, i.e. app will correctly negotiate `Content-Type` from Prom's `Accept`
- based on `Content-Type` from app container negotiation, set response header for merged metrics to this `Content-Type` (either FmtText or FmtOpenMetrics). Since agent and envoy metrics will be FmtText, they should be compatible with either format
- write out app container metrics last to support [the "# EOF" termination sequence required by FmtOpenMetrics](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#overall-structure)

A quick test with this approach exposes another issue; it seems Prometheus' FmtOpenMetrics parser does not like blank lines in the metrics, such as those in the envoy metrics, e.g.:
```
# TYPE envoy_cluster_assignment_stale counter
envoy_cluster_assignment_stale{cluster_name="xds-grpc"} 0

# TYPE envoy_cluster_assignment_timeout_received counter
envoy_cluster_assignment_timeout_received{cluster_name="xds-grpc"} 0

# TYPE envoy_cluster_bind_errors counter
envoy_cluster_bind_errors{cluster_name="xds-grpc"} 0
```

The end result is a parse error: `"INVALID" "\n" is not a valid start token`.

I also further tested this out by adding a blank line in Prom's parse testing: https://github.com/prometheus/prometheus/blob/main/pkg/textparse/openmetricsparse_test.go#L32-L33, which fails with the same parse error:
```
--- FAIL: TestOpenMetricsParse (0.00s)
    /Users/law/codebase/prometheus/pkg/textparse/openmetricsparse_test.go:225: 
        	Error Trace:	openmetricsparse_test.go:225
        	Error:      	Received unexpected error:
        	            	"INVALID" "\n" is not a valid start token
        	            	github.com/prometheus/prometheus/pkg/textparse.(*OpenMetricsParser).Next
        	            		/Users/law/codebase/prometheus/pkg/textparse/openmetricsparse.go:355
        	            	github.com/prometheus/prometheus/pkg/textparse.TestOpenMetricsParse
        	            		/Users/law/codebase/prometheus/pkg/textparse/openmetricsparse_test.go:221
        	            	testing.tRunner
        	            		/usr/local/Cellar/go/1.16/libexec/src/testing/testing.go:1194
        	            	runtime.goexit
        	            		/usr/local/Cellar/go/1.16/libexec/src/runtime/asm_amd64.s:1371
        	Test:       	TestOpenMetricsParse
FAIL
FAIL	github.com/prometheus/prometheus/pkg/textparse	0.126s
FAIL
```

Lastly, a quick hacked together build with the envoy metrics removed was scraped successfully from Prometheus.
In that case, I had istio agent metrics written as FmtText then app metrics written as FmtOpenMetrics and a response `Content-Type` of FmtOpenMetrics.

So unfortunately I don't think the existing envoy metrics are compatible with FmtOpenMetrics.

howardjohn: > So unfortunately I don't think the existing envoy metrics are compatible with FmtOpenMetrics.

Hmm.. I wonder if this is something we can change in Envoy?

The reason I am concerned with the PR as-is is it means applications cannot use OpenMetrics

@douglas-reid @ramaraochavali thoughts?

douglas-reid: Sure enough, the [spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#abnf) doesn't include a provision for empty lines.

Instead of forcing only one format, can we make the merging do the conversion?  Surely we can parse and translate in-agent, right?

lgadban: > Hmm.. I wonder if this is something we can change in Envoy?

I believe so, I'll open an issue for this.

>Instead of forcing only one format, can we make the merging do the conversion?

This seems viable, I'll test it out

lgadban:
The latest changes will support both FmtText and FmtOpenMetrics with the text-based merging and works around envoy's metrics not being compatible by post-processing them to remove the newlines. More sophisticated parsing and processing could be done, but the current state works without adding _too_ much additional complexity.

To test it properly I used Prometheus' OpenMetrics parser, please let me know if there are concerns with pulling it in as a dependency.

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -387,42 +388,67 @@ type PrometheusScrapeConfiguration struct {
...
+func negotiateMetricsFormat(contentType string) expfmt.Format {
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err == nil && mediaType == expfmt.OpenMetricsType {
+		return expfmt.FmtOpenMetrics
 	}
+	return expfmt.FmtText
+}
+
+func processMetrics(metrics []byte) []byte {
+	return []byte(strings.ReplaceAll(string(metrics), "\n\n", "\n"))
```
howardjohn, line 451: very nit: I think you can do bytes.ReplaceAll as well to avoid double conversion

howardjohn (commented):
This looks great!

cc @ramaraochavali @douglas-reid would be great to have a second look

pilot/cmd/pilot-agent/status/server.go
lgadban: great call! updated

lgadban:
Thanks for the review @howardjohn!

@ramaraochavali @douglas-reid would either of you be able to take a look?

linsun:
LGTM!  thank you @lgadban!

@douglas-reid @ramaraochavali PTAL, if no major issues, I would like to get this approved as we have users waiting on this fix!

ramaraochavali (approved):
LGTM

nrjpoddar (commented):
Can this cause any regression for an existing user? Previously we were always sending content-type FmtText, now it seems we will send it based on application response type. Is my understanding correct?

douglas-reid (approved):

lgadban:
> Can this cause any regression for an existing user? Previously we were always sending content-type FmtText, now it seems we will send it based on application response type. Is my understanding correct?

You are correct that we will now respond with the Content-Type according to the application metrics.
The previous behavior didn't explicitly set a Content-Type so it defaulted to `text/plain; charset=utf-8`.

In the case where the application metrics *only* support FmtText, and given that envoy and agent metrics are FmtText, this worked fine.

The problem was that Prometheus' scrape request will allow for FmtOpenMetrics, and that is being passed through in the application scrape request.
So in the case where the application metrics understand and support FmtOpenMetrics, the previous behavior would result in a broken scrape because the application will respond with FmtOpenMetrics metrics that will then be interleaved with incompatible metrics as well as not having the correct Content-Type set.

All of that to say, I don't see a scenario where users can be hit by a regression, as the behavior is changing only for the case where application metrics are FmtOpenMetrics, which is currently broken through metrics merging.

> Also this change would benefit from a integration test if we exercise merging in them.

This is good idea although I don't see any integration tests for metrics merging. Think it's worth opening an issue for?

linsun:
@lgadban correct me if I am wrong, I believe we will continue send `FmtText` as the default scenario unless we detect from scaping that `FmtOpenMetrics` should be used instead.  So this should not impact existing users.

Regarding integration test, it would be nice to look into that as a follow up - do require setting up proper backend system.

linsun (approved):

lgadban:
> @lgadban correct me if I am wrong, I believe we will continue send `FmtText` as the default scenario unless we detect from scaping that `FmtOpenMetrics` should be used instead. So this should not impact existing users.

@linsun yes, exactly

ryantking (approved):

PR #33556, opened 2021-06-21
improve Equals func performance 

silenceshell:

1. Add length check for maps with different length, to get result quickly.
2. As length has been checked, and the two maps have the same size, so we only need to check `i.SubsetOf(that)`.

I also add a small example to show the benchmark: https://play.golang.org/p/ZBC5Q9venNY
Run the example in local env, and will get the results:

```
--------- case 1 for length compare ------------
result false, original Equals cost time 233.469785ms
result false, new Equals cost time 162ns
--------- case 2 for one SubsetOf ------------
result true, new Equals 1 cost time 478.357971ms
result true, new Equals 2 cost time 232.908134ms
result true, new Equals 3 cost time 525.778057ms
2021-06-22 00:10:38.98496 +0800 CST m=+3.524822916
```

Case 1 shows two maps with different length, after adding length check, we only need 162ns for `EqualsNew1`, other than 233.469785ms for the original `Equals`.
Case 2 shows two maps with same length, same key & value, we could see `EqualsNew2` with one `SubsetOf` checking, costs almost **half time** of `EqualsNew1`. 
And case 2 also shows that `reflect.DeepEqual` is not a good implement for this situation(maybe the `reflect` is the cause).

---

[ ] Docs
[ ] Installation
[ ] Networking
[X] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/config/labels/instance.go
+++ b/pkg/config/labels/instance.go
@@ -80,7 +80,10 @@ func (i Instance) Equals(that Instance) bool {
 	if that == nil {
 		return i == nil
 	}
-	return i.SubsetOf(that) && that.SubsetOf(i)
+	if len(i) != len(that) {
+		return false
+	}
+	return i.SubsetOf(that)
 }
 
 // Validate ensures tag is well-formed
--- a/pilot/pkg/model/config_test.go
+++ b/pilot/pkg/model/config_test.go
@@ -179,6 +179,11 @@ func TestLabelsEquals(t *testing.T) {
 	}{
 		{
 			a: nil,
+			b: nil,
+			want: true,
+		},
+		{
+			a: nil,
 			b: labels.Instance{"a": "b"},
 		},
 		{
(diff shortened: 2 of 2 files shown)


howardjohn:
Thanks for the PR! One note - https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go golang has pretty good native support for benchmarks that may help in the future to test your changes

pkg/config/labels/instance.go
```diff
@@ -80,7 +80,10 @@ func (i Instance) Equals(that Instance) bool {
 	if that == nil {
 		return i == nil
 	}
-	return i.SubsetOf(that) && that.SubsetOf(i)
+	if len(i) != len(that) {
```
howardjohn, line 83: could you expand TestLabelsEquals a bit to ensure this doesn't result in any behavioral changes? Right now the tests look pretty light, and I am a bit worried about some edge case

howardjohn (approved):
thanks!

silenceshell:
> Thanks for the PR! One note - https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go golang has pretty good native support for benchmarks that may help in the future to test your changes

Thanks for your advice!

PR #33559, opened 2021-06-21
add new wfuzz fuzzer and tomcat backend for security fuzz testing

yangminzhu:
This PR adds a new fuzzer `wfuzz` and backend `tomcat` in the security fuzzing test, also refactored the code a little bit.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


tests/integration/security/fuzz/fuzz_test.go
```diff
@@ -45,44 +49,138 @@ spec:
...
 
+var (
+	// Known unsupported path parameter ("/bla;foo") normalization for Tomcat.
+	dotdotPwnIgnoreTomcat = []string{
+		"/../private/secret.html;index.html <- VULNERABLE!",
+		"/../private/secret.html;index.htm <- VULNERABLE!",
+		"/..%5Cprivate%5Csecret.html;index.html <- VULNERABLE!",
+		"/..%5Cprivate%5Csecret.html;index.htm <- VULNERABLE!",
+	}
+	wfuzzIgnoreTomcat = []string{
```
howardjohn, line 60: can you explain this?

tests/integration/security/fuzz/fuzz_test.go
```diff
@@ -45,44 +49,138 @@ spec:
 `
 )
 
+var (
+	// Known unsupported path parameter ("/bla;foo") normalization for Tomcat.
+	dotdotPwnIgnoreTomcat = []string{
+		"/../private/secret.html;index.html <- VULNERABLE!",
```
howardjohn, line 55: is it smart enough to ignore permutations of similar bypass?

tests/integration/security/fuzz/fuzz_test.go
yangminzhu: This is actually the same as the `dotdotPwnIgnoreTomcat` for the path parameter (e.g. "bla;foo")

howardjohn: I don't understand how we just have this list? Aren't their infinite paths that will normalize to the same URL that would have to be here?

tests/integration/security/fuzz/fuzz_test.go
```diff
@@ -45,44 +49,138 @@ spec:
...
+		"/../private/secret.html;index.htm <- VULNERABLE!",
+		"/..%5Cprivate%5Csecret.html;index.html <- VULNERABLE!",
+		"/..%5Cprivate%5Csecret.html;index.htm <- VULNERABLE!",
```
howardjohn, line 58: same question as below - the `index.html` part doesn't matter, right? It could be `index.foobar` and still bypass. So aren't the infinite possible bypasses?

tests/integration/security/fuzz/fuzz_test.go
yangminzhu: the fuzzer we used only generates finite number of paths in the test, that's why we can just ignore these when testing with the tomcat server.

tests/integration/security/fuzz/fuzz_test.go
yangminzhu: I think not, I didn't find such options in it.

tests/integration/security/fuzz/fuzz_test.go
yangminzhu: yes, there are infinite possible bypasses but the tool will only test a few number of them.

yangminzhu:
@howardjohn Mind to take a look again? thanks.

howardjohn (approved):

Issue #33469, opened 2021-06-16
Multi cluster DNS breaks due to they way iptables and conntrack operates


astrikos:
(**NOTE**: This is used to report product bugs:
  To report a security vulnerability, please visit <https://istio.io/about/security-vulnerabilities>
  To ask questions about how to use Istio, please visit <https://discuss.istio.io>)

**Bug description**
We use multi cluster istio and the federation feature using istio's very own DNS.
In order for istio to capture all DNS traffic from app containers there is among others one iptable rule like
`REDIRECT udp – anywhere 169.254.20.10 udp dpt:domain redir ports 15053`

From time to time though we experience DNS resolution errors for endpoints that are from the federated clusters and they supposed to be answered from istio's DNS feature.
Our understanding is that in theory every single DNS UDP packet coming from application inside a container will be captured and redirected to istio port 15053 and then istio might make an additional query to local resolver (53 port) if needed.

Firing up couple of tcpdumps for port 53 and port 15053 we noticed that whenever we saw resolution errors we could only see a packet going directly to the local resolver (169.254.20.10) and nothing on 15053 port of istio. That explained the resolution errors since local resolver doesn't know anything about endpoints from different clusters.

Next we enabled verbose logging in istio for the DNS parts (`dns:debug`).
The logging stream was really big but we managed to verify that when we see resolution failures we don't see any requests in the logs of istio, which in theory prints all requests and answers it served.

What we tried after that is to add LOG to iptables (nat table in the beginning) of istio-proxy using nsenter from the instances to see if we see any packets that don't match the REDIRECT rule and just go to the local resolver instead of the istio.
Surprisingly for the cases that we get resolution errors we couldn't see any logged packet, but for the cases where we got a legit response we were able to see the packets logged in /var/log/kern.log

In order to make sure packets hit iptables we added a LOG rule in the raw table of iptables, that everything should be available. In there we were able to see the logged packets, so packet that resolution was failing did reach iptables.

As can be seen from this picure https://stuffphilwrites.com/wp-content/uploads/2014/09/FW-IDS-iptables-Flowchart-v2019-04-30-1.png before a packet traverses the nat table, the conntrack table will be consulted. So if we have an entry in conntrack with same tuple (protocol, srcIP, srcPort, dstIP, dstPort) the packet will follow the route that the packet that created this entry took.

[...]

PR #33572, opened 2021-06-22
Address UDP DNS issues due to conntrack

astrikos:
The problem this PR tries to solve is described in https://github.com/istio/istio/issues/33469

In cases where an application does a lot of DNS UDP requests, we have cases where a UDP requests picks a source port that is already defined inside the conntrack tables and its packet is routed directly to the DNS resolvers.
That makes the new packet to be also routed directly to DNS resolvers without traversing the NAT iptables and hence the REDIRECT rules to send it to the local Istio DNS port.
This way we get DNS resolution errors.

The fix that implements this commit is to use separate conntrack zones for different flows. One flow will be the communication with the application, and the other will be the communication with DNS resolver.
More specific zone 2 will be UDP DNS packets that are coming from the application and return to it. Zone 1 will be UDP DNS packets that are coming from istio and going to DNS resolvers and vice versa.
Here is a picture to help you understand how the packets are classified in each conntrack zone
conntrack

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tools/istio-iptables/pkg/cmd/run.go
+++ b/tools/istio-iptables/pkg/cmd/run.go
@@ -676,6 +676,131 @@ func HandleDNSUDP(
 			}
 		}
 	}
+	// Split UDP DNS traffic to separate conntrack zones
+	addConntrackZoneDNSUDP(ops, iptables, ext, cmd, proxyUID, proxyGID, dnsServersV4, captureAllDNS)
+}
+
+// addConntrackZoneDNSUDP is a helper function to add iptables rules to split DNS traffic
+// in two separate conntrack zones to avoid issues with UDP conntrack race conditions.
+// Traffic that goes from istio to DNS servers and vice versa are zone 1 and traffic from
+// DNS client to istio and vice versa goes to zone 2
+func addConntrackZoneDNSUDP(
+	ops Ops, iptables *builder.IptablesBuilderImpl, ext dep.Dependencies,
+	cmd, proxyUID, proxyGID string, dnsServersV4 []string, captureAllDNS bool) {
+	const paramIdxRaw = 4
+	var raw []string
+	opsStr := opsToString[ops]
+	table := constants.RAW
+	chainOUTPUT := constants.OUTPUT
+	chainPREROUTING := constants.PREROUTING
+
+	// TODO: add ip6 as well
+	for _, uid := range split(proxyUID) {
+		// Packets with dst port 53 from istio to zone 1. These are Istio calls to upstream resolvers
+		raw = []string{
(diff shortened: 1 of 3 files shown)


astrikos:
@googlebot I signed it!

howardjohn (commented):
/ok-to-test

astrikos:
hm couple of tests for `integ-telemetry-k8s-tests_istio` in https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/33572/integ-telemetry-k8s-tests_istio/1407647351278931968 fail. Not sure why is that :thinking: 
Seeing the logs don't ring a bell where the iptables could affect this. cc @howardjohn do you have any pointers where to look to make the tests happy?

astrikos:
>hm couple of tests for integ-telemetry-k8s-tests_istio in https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/33572/integ-telemetry-k8s-tests_istio/1407647351278931968 fail. Not sure why is that thinking
Seeing the logs don't ring a bell where the iptables could affect this. cc @howardjohn do you have any pointers where to look to make the tests happy?

scratch that, I rerun it and it's happy now. Do I need to do something myself for the `release-notes_istio` test or someone else is doing that?

Let me know if there is any feedback about this, or I should have to do something more before it's ready for review

howardjohn (commented):
This looks great to me. I would love to get more eyes on this since its pretty complex. @lambdai @ramaraochavali @hzxuzhonghu ?

ramaraochavali:
i do not have much knowledge on iptables. So leave it for others.

howardjohn (approved):

astrikos:
@howardjohn thanks for pushing this forward.
Do you know in which release this will be included? is there a chance to be in the next one? 1.10.x?
If you need to include also a release note, I can prepare something.

howardjohn:
We could probably backport to 1.10 (branch: release-1.10) with a feature flag that is off by default. I would be a bit worried about if this could cause subtle breakages in a patch release otherwise, but off by default is safe.

astrikos:
>We could probably backport to 1.10 (branch: release-1.10) with a feature flag that is off by default

Yeah we would like that, if you don't have time I could create a PR if you give me some pointers how to add the flag (if there is any examples how to pass it down to the dns file would be great)

>I would be a bit worried about if this could cause subtle breakages in a patch release otherwise, but off by default is safe.

Agree, a feature flag would do for us, we could test is as well in our infra, since we are affected already by it.

Issue #33569, opened 2021-06-22
High istiod/discovery memory usage since upgrade to 1.10


Stono:
**Bug description**
We recently upgraded from `1.9.5` to `1.10.1`, and we are finding the memory usage of istiod much higher and quite unstable, as can be seen from the graph here:

![Screenshot 2021-06-22 at 11 05 13](https://user-images.githubusercontent.com/1486729/122906143-bf455f00-d349-11eb-818f-7b756a1fecfe.png)

This is in mb, so we're seeing 35%ish swings on the average.  The average has gone from `400mb` to `1.3gb`, and we're peaking as high as `1.7gb`.

Zooming in to the lower time frames, and a single pod we see:
![Screenshot 2021-06-22 at 11 09 09](https://user-images.githubusercontent.com/1486729/122906681-514d6780-d34a-11eb-9a19-170b0d958232.png)

07:00 is roughly the time of day our cluster starts taking deployments from developers, so i'd say this memory usage is related to config pushes, but the times don't really marry with the istiod logs: 

[logs.log.tar.gz](https://github.com/istio/istio/files/6693294/logs.log.tar.gz)

[ ] Docs
[ ] Installation
[ ] Networking
[x] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**
Memory to be stable, and memory increases in istiod to be acknowledged as expected in release notes if they are indeed expected.

**Steps to reproduce the bug**
Unsure, obviously we have a big deployment with a lot of configuration, and welcome anyone to live debug with me.

**Version** (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version --short` if you used Helm)
1.10.1

**How was Istio installed?**

**Environment where the bug was observed (cloud vendor, OS, etc)**

Additionally, please consider running `istioctl bug-report` and attach the generated cluster-state tarball to this issue.
Refer [cluster state archive](http://istio.io/help/bugs/#generating-a-cluster-state-archive) for more details.

howardjohn:
If you could get a memory profile it would be helpful: https://github.com/istio/istio/wiki/Analyzing-Istio-Performance#memory.

howardjohn:
The only change that stands out to me is go 1.16 upgrade which has changes to how memory is managed/reported: https://golang.org/doc/go1.16#runtime

Stono:
[...]

PR #33576, opened 2021-06-22
Add logs for full push reason, when feasible

howardjohn:
This ensures for single-config updates, we get a log of what triggered
the push. In some cases, there may be 100s of triggers, so logging all
of them is a bit excessive - but if there is just one its reasonable,
and this is the most common case.

Example:
```
ads     Push debounce stable[1] 90 for 74 configs: 100.540837ms since last change, 192.854347ms since last push, full=true
ads     Push debounce stable[2] 1 for config VirtualService/route.default: 100.215996ms since last change, 100.215807ms since last push, full=true
ads     Push debounce stable[3] 2 for config ServiceEntry/shell.default.svc.cluster.local.default: 100.99523ms since last change, 101.540648ms since last push, full=true
```

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/xds/discovery.go
+++ b/pilot/pkg/xds/discovery.go
@@ -363,6 +363,15 @@ func (s *DiscoveryServer) handleUpdates(stopCh <-chan struct{}) {
 	debounce(s.pushChannel, stopCh, s.debounceOptions, s.Push, s.CommittedUpdates)
 }
 
+// extractFirstConfigKey grabs the "first" element in the map. This is random, so this function
+// should only be used when we have pre-validated that the map has only a single item
+func extractFirstConfigKey(c map[model.ConfigKey]struct{}) model.ConfigKey {
+	for x := range c {
+		return x
+	}
+	return model.ConfigKey{}
+}
+
 // The debounce helper function is implemented to enable mocking
 func debounce(ch chan *model.PushRequest, stopCh <-chan struct{}, opts debounceOptions, pushFn func(req *model.PushRequest), updateSent *atomic.Int64) {
 	var timeChan <-chan time.Time
@@ -391,9 +400,17 @@ func debounce(ch chan *model.PushRequest, stopCh <-chan struct{}, opts debounceO
 		if eventDelay >= opts.debounceMax || quietTime >= opts.debounceAfter {
 			if req != nil {
 				pushCounter++
-				log.Infof("Push debounce stable[%d] %d: %v since last change, %v since last push, full=%v",
-					pushCounter, debouncedEvents,
-					quietTime, eventDelay, req.Full)
+				configsUpdated := len(req.ConfigsUpdated)
+				if configsUpdated == 1 {
+					update := extractFirstConfigKey(req.ConfigsUpdated)
+					log.Infof("Push debounce stable[%d] %d for config %v: %v since last change, %v since last push, full=%v",
+						pushCounter, debouncedEvents, update.String(),
+						quietTime, eventDelay, req.Full)
+				} else {
+					log.Infof("Push debounce stable[%d] %d for %d configs: %v since last change, %v since last push, full=%v",
+						pushCounter, debouncedEvents, configsUpdated,
+						quietTime, eventDelay, req.Full)
+				}
 
 				free = false
 				go push(req, debouncedEvents)
--- a/pilot/pkg/model/config.go
+++ b/pilot/pkg/model/config.go
@@ -17,6 +17,7 @@ package model
 import (
 	"crypto/md5"
 	"encoding/binary"
+	"fmt"
 	"net"
 	"sort"
 	"strings"
@@ -41,6 +42,10 @@ type ConfigKey struct {
 	Kind      config.GroupVersionKind
 	Name      string
 	Namespace string
+}
+
+func (key ConfigKey) String() string {
+	return fmt.Sprintf("%s/%s.%s", key.Kind.Kind, key.Name, key.Namespace)
 }
 
 func (key ConfigKey) HashCode() uint64 {


hzxuzhonghu (changes requested):
Looks very hacky for me, not common usage.

howardjohn:
This was requested by @Stono. I agree with him that it's an extremely useful log. When you are seeing full pushes in the cluster, today there is no way to see why they are occuring.

What harm is adding a small additional info to the log?

hzxuzhonghu:
Why 1 update should be treated differently?

howardjohn:
Well we can't log 4000 configs in 1 line for sure, that is too spammy. The
vast majority are single config updates, so this lets us log useful info in
majority care without adding spam

hzxuzhonghu:
I donot think it is that useful with only part of the events printed. In case a exception occur, we could still not diagnose what's going wrong

howardjohn:
Why is it not useful to have the logs for 95% of cases just because we
don't have them for 5% of them?

It's not like this is some extreme complexity, it's a couple lines of log

hzxuzhonghu:
For 95%, the logs should be spammy, why not change it to debug and make it work for all

pilot/pkg/xds/discovery.go
```diff
@@ -391,9 +400,17 @@ func debounce(ch chan *model.PushRequest, stopCh <-chan struct{}, opts debounceO
 		if eventDelay >= opts.debounceMax || quietTime >= opts.debounceAfter {
 			if req != nil {
 				pushCounter++
-				log.Infof("Push debounce stable[%d] %d: %v since last change, %v since last push, full=%v",
-					pushCounter, debouncedEvents,
-					quietTime, eventDelay, req.Full)
+				configsUpdated := len(req.ConfigsUpdated)
+				if configsUpdated == 1 {
```
hzxuzhonghu, line 404: @howardjohn This kind of code looks really urgly and unreadable. And could not tell how useful it is.

howardjohn:
If we make it debug log it's probably also going to turn in tons of other
logs. I am not sure why it's so bad to have an extra 20 characters or so
here. I have ran this a bit and found it quite useful already so I am not
sure what the issue is. It's low cost, and according to one of Istio's top
users would be helpful for debugging

hzxuzhonghu:
As you say if it includes 95 % of all the update events, the cost will be huge, And I think we don't have priority orders for users. Every user should be equally treated.

howardjohn:
> As you say if it includes 95 % of all the update events, the cost will be huge, And I think we don't have priority orders for users. Every user should be equally treated.

Ok, then I will go talk to our users and find enough users that will find this to be useful. Would that be sufficient?

howardjohn:
and I can provide benchmarks to prove this doesn't have any performance cost, as you mentioned the cost will be huge which I do not believe is accurate

pilot/pkg/xds/discovery.go
howardjohn: Can you give constructive feedback on his to improve it instead of calling it ugly? Ugly is subjective and to me it's fine, so I am not sure how to improve. Thanks!

pilot/pkg/xds/discovery.go
```diff
@@ -363,6 +363,15 @@ func (s *DiscoveryServer) handleUpdates(stopCh <-chan struct{}) {
 	debounce(s.pushChannel, stopCh, s.debounceOptions, s.Push, s.CommittedUpdates)
 }
 
+// extractFirstConfigKey grabs the "first" element in the map. This is random, so this function
```
adiprerepa, line 366: Map ordering is random like you said, but here is an alternative that might be more explicit (up to you):
```go
keys := reflect.ValueOf(c).MapKeys()
return keys[rand.Intn(len(keys))].Interface()
```

adiprerepa (commented):
Lgtm

adiprerepa:
> For 95%, the logs should be spammy, why not change it to debug and make it work for all

Am I misunderstanding in thinking that this change only adds a few characters or so if and only if there is one config updated? Do we expect the output of logs to be reliable/uniform? If not, I don't see a downside to the change.

stevenctl (approved):
really useful – i find myself adding a similar log for debugging frequently

Stono:
The hostility here is strong.  Relax!

For context, I asked for this because of the high memory usage in https://github.com/istio/istio/issues/33569 when a full push occurs, which subsequently highlighted to me that we were getting full pushes when I believe we shouldn't https://github.com/istio/istio/issues/33600, and as an end user it wan't easy for me to see why this was happening and I needed John to build a custom image with more logging enabled.  As an end user, a log without context, is just noise.

As full pushes are one of the more expensive operations for the control plane and also historically have been the cause of pain (503s due to proxy reloads, etc), being able to understand why they're happening empowers the operator to take action to minimise them, and in the case of an issue, raise it with more context.

adiprerepa (approved):

hzxuzhonghu:
Is it useful that we change a conflict log level from debug to info and add a flag to control only log it once? I think we have many such cases.

hzxuzhonghu:
Could we do this? If so the code will be a mess...

howardjohn:
> Is it useful that we change a conflict log level from debug to info and add a flag to control only log it once? I think we have many such cases.

I am not sure what you mean, could you clarify?

hzxuzhonghu:
I mean something like https://github.com/istio/istio/blob/3aafcff55379d1ce51a62b700a46d63c637daef6/pilot/pkg/networking/core/v1alpha3/listener.go#L329 is really useful. but I donot think we can add a bool flag to control we log only once but not every time we push.


I could not see much diff here and there.  Even we know 95% of all the configs updates, then we miss 5%, how can you say the 5% is unimportant, rather mostly I think the 5 % are under high load, and could be more important.

howardjohn:
That log is totally different, it's triggered for every push * every proxy * every port. The log here is only for every push, which due to denouncing is at most 10/s.

As far as the 5% of cases - I agree maybe it would be useful to expose those, but I don't have any good suggestions there. I don't think we need to block an improvement for most cases just because it doesn't help in all cases

Issue #33534, opened 2021-06-19
Websocket over http2 does not work on version 10.1


skk2142:
[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Bug description**
Envoy does not forward websocket upgrade request to the application. It instead just returns an 400 http error. Non-websocket requests work fine. I have added envoy debug logs.
This is how the request looks like:
```
gateway-dark-5b7488dd7b-zjwrl istio-proxy ':method', 'GET'
gateway-dark-5b7488dd7b-zjwrl istio-proxy ':authority', 'gully.to'
gateway-dark-5b7488dd7b-zjwrl istio-proxy ':scheme', 'https'
gateway-dark-5b7488dd7b-zjwrl istio-proxy ':path', '/graphql'
gateway-dark-5b7488dd7b-zjwrl istio-proxy 'pragma', 'no-cache'
gateway-dark-5b7488dd7b-zjwrl istio-proxy 'cache-control', 'no-cache'
gateway-dark-5b7488dd7b-zjwrl istio-proxy 'user-agent', 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.106 Safari/537.36'
gateway-dark-5b7488dd7b-zjwrl istio-proxy 'origin', 'https://gully.to'
gateway-dark-5b7488dd7b-zjwrl istio-proxy 'sec-websocket-version', '13'
gateway-dark-5b7488dd7b-zjwrl istio-proxy 'accept-encoding', 'gzip, deflate, br'
gateway-dark-5b7488dd7b-zjwrl istio-proxy 'accept-language', 'en-US,en;q=0.9,fr;q=0.8,la;q=0.7'
gateway-dark-5b7488dd7b-zjwrl istio-proxy 'x-prop-gateway-dark', 'true'
gateway-dark-5b7488dd7b-zjwrl istio-proxy 'sec-websocket-extensions', 'permessage-deflate; client_max_window_bits'
gateway-dark-5b7488dd7b-zjwrl istio-proxy 'sec-websocket-protocol', 'graphql-transport-ws'
gateway-dark-5b7488dd7b-zjwrl istio-proxy 'x-forwarded-for', '74.72.131.115'
[...]

PR #33581, opened 2021-06-22
Fix websockets when using h2 upgrade

Monkeyanator:
[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/listener.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener.go
@@ -283,6 +283,9 @@ func (configgen *ConfigGeneratorImpl) buildSidecarInboundHTTPListenerOptsForPort
 			},
 			ServerName:          EnvoyServerName,
 			DelayedCloseTimeout: features.DelayedCloseTimeout,
+			Http2ProtocolOptions: &core.Http2ProtocolOptions{
+				AllowConnect: true,
+			},
 		},
 	}
 	// See https://github.com/grpc/grpc-web/tree/master/net/grpc/gateway/examples/helloworld#configure-the-proxy
--- a/pilot/pkg/networking/core/v1alpha3/listener.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener.go
@@ -281,16 +281,14 @@ func (configgen *ConfigGeneratorImpl) buildSidecarInboundHTTPListenerOptsForPort
 				Uri:     true,
 				Dns:     true,
 			},
-			ServerName:          EnvoyServerName,
-			DelayedCloseTimeout: features.DelayedCloseTimeout,
-			Http2ProtocolOptions: &core.Http2ProtocolOptions{
-				AllowConnect: true,
-			},
+			ServerName:           EnvoyServerName,
+			DelayedCloseTimeout:  features.DelayedCloseTimeout,
+			Http2ProtocolOptions: http2ProtocolOptions(),
 		},
 	}
 	// See https://github.com/grpc/grpc-web/tree/master/net/grpc/gateway/examples/helloworld#configure-the-proxy
 	if pluginParams.ServiceInstance.ServicePort.Protocol.IsHTTP2() {
-		httpOpts.connectionManager.Http2ProtocolOptions = &core.Http2ProtocolOptions{}
+		httpOpts.connectionManager.Http2ProtocolOptions = http2ProtocolOptions()
 		if pluginParams.ServiceInstance.ServicePort.Protocol == protocol.GRPCWeb {
 			httpOpts.addGRPCWebFilter = true
 		}
--- a/tests/integration/pilot/common/routing.go
+++ b/tests/integration/pilot/common/routing.go
@@ -1976,7 +1976,7 @@ spec:
 
 	for _, client := range clients {
 		cases = append(cases, TrafficTestCase{
-			name:   fmt.Sprintf("websocket-h2-upgrade"),
+			name:   "websocket-h2-upgrade",
 			config: h2UpgradeDR(destination.Config().Service),
 			call:   client.CallWithRetryOrFail,
 			opts: echo.CallOptions{
--- a/tests/integration/pilot/common/routing.go
+++ b/tests/integration/pilot/common/routing.go
@@ -1955,3 +1955,38 @@ func serverFirstTestCases(apps *EchoDeployments) []TrafficTestCase {
 
 	return cases
 }
+
+func websocketTestCases(apps *EchoDeployments) []TrafficTestCase {
+	cases := make([]TrafficTestCase, 0)
+	clients := apps.PodA
+	destination := apps.PodC[0]
+	h2UpgradeDR := func(dst string) string {
+		return fmt.Sprintf(`apiVersion: networking.istio.io/v1beta1
+kind: DestinationRule
+metadata:
+  name: %s
+spec:
+  host: %s
+  trafficPolicy:
+    connectionPool:
+      http:
+        h2UpgradePolicy: UPGRADE
+	`, dst, dst)
+	}
+
+	for _, client := range clients {
+		cases = append(cases, TrafficTestCase{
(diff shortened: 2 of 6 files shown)


pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -283,6 +283,9 @@ func (configgen *ConfigGeneratorImpl) buildSidecarInboundHTTPListenerOptsForPort
 			},
 			ServerName:          EnvoyServerName,
 			DelayedCloseTimeout: features.DelayedCloseTimeout,
+			Http2ProtocolOptions: &core.Http2ProtocolOptions{
```
Monkeyanator, line 286: can we just set `Http2ProtocolOptions` on all HTTP listeners like this?

howardjohn, line 286: do we need it on outbound as well?

howardjohn: and gateway

tests/integration/pilot/common/routing.go
```diff
@@ -1955,3 +1955,38 @@ func serverFirstTestCases(apps *EchoDeployments) []TrafficTestCase {
...
+	for _, client := range clients {
+		cases = append(cases, TrafficTestCase{
+			name:   fmt.Sprintf("websocket-h2-upgrade"),
+			config: h2UpgradeDR(destination.Config().Service),
+			call:   client.CallWithRetryOrFail,
+			opts: echo.CallOptions{
+				Target:   destination,
+				PortName: "http",
+				Scheme:   scheme.WebSocket,
+				Count:    50,
```
Monkeyanator, line 1986: For some reason this config takes a while to apply and with less requests than this the test passes 100% of the time

pilot/pkg/networking/core/v1alpha3/listener.go
howardjohn: @zhiminxiang this will make https://github.com/knative/serving/issues/7933#issuecomment-786139169 not required I think

tests/integration/pilot/common/routing.go
```diff
@@ -1955,3 +1955,38 @@ func serverFirstTestCases(apps *EchoDeployments) []TrafficTestCase {
...
+		return fmt.Sprintf(`apiVersion: networking.istio.io/v1beta1
+kind: DestinationRule
+metadata:
+  name: %s
+spec:
+  host: %s
+  trafficPolicy:
+    connectionPool:
+      http:
+        h2UpgradePolicy: UPGRADE
```
howardjohn, line 1973: I think we want to test more than just upgrade, also client sending WS over h2 directly?

pilot/pkg/networking/core/v1alpha3/listener.go
howardjohn: I think this will end up conflicting here: https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/v1alpha3/listener.go#L290

As for whether it can be set all of the time - I _think_ so, since the protocol selection is determined by `codec_type`, so I would imagine it doesn't matter for non-h2 traffic (despite the line I linked seeming to imply it is doing something). @lambdai ?

tests/integration/pilot/common/routing.go
Monkeyanator: I don't think `gorilla/websockets` (what echo client uses) supports websockets over H2? [https://github.com/gorilla/websocket/issues/417](https://github.com/gorilla/websocket/issues/417)

Monkeyanator: For context this was more to directly test the broken case in [https://github.com/istio/istio/issues/33534](https://github.com/istio/istio/issues/33534)

pilot/pkg/networking/core/v1alpha3/listener.go
Monkeyanator: From [these docs](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/upgrades#websocket-over-http-2-or-http-3-hops):

> This is accomplished via Extended CONNECT (RFC8441) support, turned on by setting allow_connect true at the second layer Envoy.

I took this to mean that the inbound listeners need this set but not the outbound although I'm not entirely clear on what "second layer" referred to there.

And yeah I think this should be added to gateways

hzxuzhonghu:
Pretty good, seems there should be a release note

howardjohn (commented):
Discussing with @kyessenov and @lambdai offline, there are some concerns about this. From the envoy docs:

> This mode of CONNECT support can create major security holes if not configured correctly, as the upstream will be forwarded unsanitized headers if they are in the body payload. Please use with caution

Basically we may have to worry about bypassing http policies

Monkeyanator:
I am honestly struggling with the Envoy docs on this one (probably I need to read `CONNECT` RFC and come back), but from: 

> Envoy CONNECT support is off by default (Envoy will send an internally generated 403 in response to CONNECT requests). CONNECT support can be enabled via the upgrade options described above, setting the upgrade value to the special keyword “CONNECT”.

Since we only set

```go
// Allow websocket upgrades
websocketUpgrade := &hcm.HttpConnectionManager_UpgradeConfig{UpgradeType: "websocket"}
connectionManager.UpgradeConfigs = []*hcm.HttpConnectionManager_UpgradeConfig{websocketUpgrade}
```

wouldn't this just allow websocket upgrade request proxying over H2 (here's [an example config](https://github.com/envoyproxy/envoy/blob/e795b658f65b0e522b50fae5f1d3f90ea653ab7a/configs/terminate_http2_connect.yaml#L37) enabling the dangerous CONNECT proxying described, not the same as what we do here)? Confusing since if that's the case, I'm not sure what `AllowConnect` does on its own...

thesuperzapper:
@howardjohn @Monkeyanator I think this needs another look.

The warning from the Envoy docs that was raised by @howardjohn was actually under the [following section](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/upgrades#connect-support), not the section for [enabling `allow_connect`](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/upgrades#websocket-over-http-2-or-http-3-hops), and I am pretty sure it does not apply here.

We need to give a way to set `allow_connect` to `true` for a specific sidecar `HttpConnectionManager` (probably using the VirtualService or DestinationRule), so we can avoid people having to make lots of EnvoyFilters like this:

```yaml
kind: EnvoyFilter
metadata:
  name: {{ MY_FILTER }}
  namespace: {{ MY_NAMESPACE }}
spec:
  workloadSelector:
    labels:
      aaa: {{ MY_LABEL }}}
  configPatches:
    - applyTo: NETWORK_FILTER
      match:
        context: SIDECAR_INBOUND
        proxy:
          proxyVersion: '^1\.*.*'
        listener:
          filterChain:
            filter:
              name: envoy.filters.network.http_connection_manager
      patch:
        operation: MERGE
        value:
          typed_config:
            ## https://www.envoyproxy.io/docs/envoy/v1.23.1/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager
            "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager

            ## https://www.envoyproxy.io/docs/envoy/v1.23.1/api-v3/config/core/v3/protocol.proto#envoy-v3-api-msg-config-core-v3-http2protocoloptions
            http2_protocol_options:
              allow_connect: true
```

For reference, this is such a long-standing issue with Istio that Alibaba has a help article on it:
- https://www.alibabacloud.com/blog/alibaba-cloud-service-mesh-faq-1-how-to-use-the-websocket-over-http2-protocol_600045

PR #13582, opened 2019-04-24
fix: Allow use of hostname in meshNetworks gateways

mikebryant:
Part of #13581 - might well fix it?

PR #33582, opened 2021-06-22
Refactoring network gateway storage.

nmittler:
The `PushContext` previously was manually merging `MeshNetworks` with the network gateways provided by the service registries.

Added a `Gateways` structure as an abstraction around the collection of gateways with methods for all current use cases.

Replaced `PushContext` methods with basic access to the `Gateways` object.

Also removed locking around access to `Gateways`. I don't think it's necessary since its already a snapshot in time of the gateways from `MeshNetworks` and the registries.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/model/network.go
+++ b/pilot/pkg/model/network.go
@@ -1,0 +1,144 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package model
+
+import (
+	"net"
+)
+
+const (
+	noCluster = ""
+)
+
+// NetworkID is the unique identifier for a network.
+type NetworkID string
+
+func (id NetworkID) Equals(other NetworkID) bool {
+	return SameOrEmpty(string(id), string(other))
+}
+
+// ClusterID is the unique identifier for a k8s cluster.
+type ClusterID string
+
+func (id ClusterID) Equals(other ClusterID) bool {
+	return SameOrEmpty(string(id), string(other))
+}
+
+// Gateway is the gateway of a network
+type Gateway struct {
+	// Network is the ID of the network where this Gateway resides.
+	Network NetworkID
+	// Cluster is the ID of the k8s cluster where this Gateway resides.
+	Cluster ClusterID
+	// gateway ip address
+	Addr string
+	// gateway port
+	Port uint32
+}
+
+// Gateways provides information about network Gateway resources throughout the mesh.
+type Gateways interface {
+	// IsMultiNetworkEnabled returns true if one or more Gateways are configured.
+	IsMultiNetworkEnabled() bool
+
+	// All returns all configured Gateway resources in the entire mesh.
+	All() []*Gateway
+
+	// ForNetwork returns all Gateway resources configured for the given network.
+	ForNetwork(network NetworkID) []*Gateway
+
+	// ForNetworkAndCluster returns all Gateway resources configured for the given network and cluster.
+	ForNetworkAndCluster(network NetworkID, cluster ClusterID) []*Gateway
+}
+
+// newGateways creates a new Gateways from the Environment by merging
+// together the MeshNetworks and ServiceRegistry-specific gateways.
+func newGateways(env *Environment) Gateways {
+	// Generate the a snapshot of the state of gateways by merging the contents of
+	// MeshNetworks and the ServiceRegistries.
+	gatewayMap := make(map[NetworkID]map[ClusterID][]*Gateway)
+
+	addGateway := func(gateway *Gateway) {
+		// Get (or create) an entry for the network.
+		gatewaysByCluster := gatewayMap[gateway.Network]
+		if gatewaysByCluster == nil {
+			gatewaysByCluster = make(map[ClusterID][]*Gateway)
+			gatewayMap[gateway.Network] = gatewaysByCluster
+		}
+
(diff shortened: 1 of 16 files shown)


pilot/pkg/model/network.go
```diff
@@ -0,0 +1,144 @@
...
+type gatewaysImpl struct {
+	gateways map[NetworkID]map[ClusterID][]*Gateway
+}
+
+func (gws *gatewaysImpl) IsMultiNetworkEnabled() bool {
+	return len(gws.gateways) > 0
+}
+
+func (gws *gatewaysImpl) All() []*Gateway {
+	out := make([]*Gateway, 0)
```
stevenctl, line 127: nit: estimate capacity with `len(gws.gateways * len(gws.gateways[firstKey])`?

pilot/pkg/model/network.go
```diff
@@ -0,0 +1,144 @@
...
+	// Cluster is the ID of the k8s cluster where this Gateway resides.
+	Cluster ClusterID
+	// gateway ip address
+	Addr string
+	// gateway port
+	Port uint32
+}
+
+// Gateways provides information about network Gateway resources throughout the mesh.
+type Gateways interface {
```
howardjohn, line 52: Why do we need an interface? There will only ever be one implementation I assume?

pilot/pkg/model/network.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// newGateways creates a new Gateways from the Environment by merging
+// together the MeshNetworks and ServiceRegistry-specific gateways.
+func newGateways(env *Environment) Gateways {
+	// Generate the a snapshot of the state of gateways by merging the contents of
+	// MeshNetworks and the ServiceRegistries.
+	gatewayMap := make(map[NetworkID]map[ClusterID][]*Gateway)
+
+	addGateway := func(gateway *Gateway) {
```
stevenctl, line 73: nit: why not make this a method on `gatewaysImpl` – the only thing we're accessing in the parent scope is `gatewayMap` which is essentially `gatewaysImpl.gateways`

pilot/pkg/model/push_context.go
```diff
@@ -1881,46 +1871,11 @@ func instancesEmpty(m map[int][]*ServiceInstance) bool {
...
-		// - the computed map from meshNetworks (triggered by reloadNetworkLookup, the ported logic from getGatewayAddresses)
-		ps.networkGateways[network] = append(ps.networkGateways[network], gateways...)
-	}
-}
-
-func (ps *PushContext) NetworkGateways() map[string][]*Gateway {
-	ps.networksMu.RLock()
-	defer ps.networksMu.RUnlock()
-	return ps.networkGateways
+	ps.gateways = newGateways(env)
```
howardjohn, line 1874: can we call it networkGateways or something? "gateway" is already overloaded (2x with the k8s gateway)

pilot/pkg/model/network.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+	}
+
+	// First, load gateways from the static MeshNetworks config.
+	meshNetworks := env.Networks()
+	if meshNetworks != nil {
+		for network, networkConf := range meshNetworks.Networks {
+			gws := networkConf.Gateways
+			for _, gw := range gws {
+				if gwIP := net.ParseIP(gw.GetAddress()); gwIP != nil {
```
stevenctl, line 94: Can we add an `else { warn }` – some users get really confused when hostnames don't work from mesh networks.

At one point it was supported: https://github.com/istio/istio/pull/13582

but now it's not...

pilot/pkg/model/service.go
```diff
@@ -417,6 +422,14 @@ type IstioEndpoint struct {
 	DiscoverabilityPolicy EndpointDiscoverabilityPolicy `json:"-"`
 }
 
+// GetLoadBalancingWeight returns the LbWeight for this endpoint, or DefaultLoadBalancingWeight if not set.
+func (ep *IstioEndpoint) GetLoadBalancingWeight() uint32 {
+	if ep.LbWeight > 0 {
+		return ep.LbWeight
+	}
+	return DefaultLoadBalancingWeight
```
stevenctl, line 430: Have you seen a case where we have to fallback to this? If so that's an issue elsewhere and we might want a Debug log here.

pilot/pkg/serviceregistry/aggregate/controller.go
```diff
@@ -181,16 +181,10 @@ func mergeService(dst, src *model.Service, srcCluster string) {
 }
 
 // NetworkGateways merges the service-based cross-network gateways from each registry.
-func (c *Controller) NetworkGateways() map[string][]*model.Gateway {
-	gws := map[string][]*model.Gateway{}
+func (c *Controller) NetworkGateways() []*model.Gateway {
+	gws := make([]*model.Gateway, 0)
```
stevenctl, line 185: very nit: it's more common to `var gws []*model.Gateway` and use `make` when we can set `len/cap`

pilot/pkg/xds/ep_filters.go
```diff
@@ -134,6 +136,16 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocLbEndpointsAn
 	return filtered
 }
 
+func (b *EndpointBuilder) gatewaysByNetwork() map[model.NetworkID][]*model.Gateway {
```
stevenctl, line 139: There's a lot of converting back and forth between slice and map. Can we expose the underlying map in push.Gateways, maybe comment a warning not to modify it.

pilot/pkg/model/network.go
nmittler: This seemed more appropriate, since `Gateways` is effectively immutable. Adding an `addGateway` method would suggest that it isn't.

pilot/pkg/model/service.go
nmittler: oh this was supposed to be reverted.

pilot/pkg/xds/ep_filters.go
nmittler: My plan was to eliminate that in the follow-on PR that will modify this load balancing algorithm.

stevenctl (approved):

PR #31226, opened 2021-03-04
Remove cluster label stats plugin customization

bianpengyuan:
[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any changes that may affect Istio users.

PR #33583, opened 2021-06-22
Implement metrics portion of telemetry API

howardjohn:
This implements the `metrics` portion of the API. 

Current state:
* Pilot fully generates equivalent stats and mx filters as EnvoyFilters
* None of https://docs.google.com/document/d/1FMvwyoooGeiACf9b7R0A6iJyB6ZuiP7hv5-XDmXtHLo/edit#heading=h.xw1gqgyqs5b is implemented. The changes here are effectively off by default, as everything is skipped when no Telemetry CR is found. For a user to use this, they should first remove all EnvoyFilters, then create Telemetry CR. This is intended to be improved to allow more graceful migrations (see design doc) but out of scope of the initial PR.
* Only prometheus is supported
* Overrides are not supported at all. Only enable/disable of metrics entirely.
* Per-workload configuration is supported. One interesting change here, now that we allow this, is that we need inbound metadata exchange on ALL pods, if any pod has telemetry enabled. We may need to change this to always enable MX unconditionally, since we may not know all configs in the mesh for multicluster; for now we would need to wait until we have a more graceful migration plan in order to do that.
* Basic integration tests added

### DIFF
--- a/pilot/pkg/model/telemetry.go
+++ b/pilot/pkg/model/telemetry.go
@@ -15,7 +15,11 @@
 package model
 
 import (
+	"strings"
+
+	meshconfig "istio.io/api/mesh/v1alpha1"
 	tpb "istio.io/api/telemetry/v1alpha1"
+	"istio.io/istio/pilot/pkg/util/sets"
 	"istio.io/istio/pkg/config/labels"
 	"istio.io/istio/pkg/config/schema/collections"
 	istiolog "istio.io/pkg/log"
@@ -62,6 +66,11 @@ func GetTelemetries(env *Environment) (*Telemetries, error) {
 	}
 
 	return telemetries, nil
+}
+
+// AnyTelemetryExists determines if there are an Telemetries present in the entire mesh.
+func (t *Telemetries) AnyTelemetryExists() bool {
+	return len(t.NamespaceToTelemetries) > 0
 }
 
 func (t *Telemetries) EffectiveTelemetry(proxy *Proxy) *tpb.Telemetry {
@@ -116,8 +125,31 @@ func shallowMerge(parent, child *tpb.Telemetry) *tpb.Telemetry {
 	}
 	merged := parent.DeepCopy()
 	shallowMergeTracing(merged, child)
+	mergeMetrics(merged, child)
 	shallowMergeAccessLogs(merged, child)
 	return merged
+}
+
+func mergeMetrics(parent *tpb.Telemetry, child *tpb.Telemetry) {
+	if len(parent.GetMetrics()) == 0 {
+		parent.Metrics = child.Metrics
+		return
+	}
+	if len(child.GetMetrics()) == 0 {
+		return
+	}
+
+	// TODO proper merge of matches
+	childMetrics := child.Metrics[0]
+	mergedMetrics := parent.Metrics[0]
+	if len(childMetrics.Providers) != 0 {
+		mergedMetrics.Providers = childMetrics.Providers
+	}
+
+	// TODO deep merge
+	if childMetrics.Overrides != nil {
+		mergedMetrics.Overrides = childMetrics.Overrides
+	}
 }
 
 func shallowMergeTracing(parent, child *tpb.Telemetry) {
@@ -168,4 +200,45 @@ func shallowMergeAccessLogs(parent *tpb.Telemetry, child *tpb.Telemetry) {
 	if childLogging.GetDisabled() != nil {
 		mergedLogging.Disabled = childLogging.Disabled
 	}
+}
+
+// PrometheusEnabled determines if any of the extension providers are for prometheus
+func PrometheusEnabled(providers []*meshconfig.MeshConfig_ExtensionProvider) bool {
+	for _, prov := range providers {
+		if _, ok := prov.Provider.(*meshconfig.MeshConfig_ExtensionProvider_Prometheus); ok {
+			return true
+		}
+	}
+	return false
+}
+
+// MetricsEnabled determines if there is any `metrics` configuration for the provided mesh configuration and Telemetry.
+// Note that this can return `true` even when `spec` is `nil`, in the case there is a default provider.
+func MetricsEnabled(spec *tpb.Telemetry, mesh *meshconfig.MeshConfig) bool {
+	return len(MetricsProviders(spec, mesh)) > 0
+}
+
+// MetricsProviders returns a list of all metrics extension providers used by the spec
+// (`metrics.providers`) and mesh configuration (`defaultProviders.metrics`).
+func MetricsProviders(spec *tpb.Telemetry, mesh *meshconfig.MeshConfig) []*meshconfig.MeshConfig_ExtensionProvider {
+	providers := sets.NewSet(mesh.GetDefaultProviders().GetMetrics()...)
+	if len(spec.GetMetrics()) > 0 {
+		// handle multiple
+		if prov := spec.Metrics[0].GetProviders(); len(prov) > 0 {
+			providers.Insert(strings.ToLower(prov[0].Name))
+		}
+	}
+	if len(providers) == 0 {
+		return nil
+	}
+
+	ret := []*meshconfig.MeshConfig_ExtensionProvider{}
+
+	for _, p := range mesh.ExtensionProviders {
+		name := strings.ToLower(p.Name)
+		if providers.Contains(name) {
+			ret = append(ret, p)
+		}
+	}
+	return ret
 }
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -41,6 +41,7 @@ import (
 	authn_model "istio.io/istio/pilot/pkg/security/model"
 	"istio.io/istio/pilot/pkg/serviceregistry/provider"
 	"istio.io/istio/pilot/pkg/util/sets"
+	xdsfilters "istio.io/istio/pilot/pkg/xds/filters"
 	v3 "istio.io/istio/pilot/pkg/xds/v3"
 	"istio.io/istio/pkg/config"
 	"istio.io/istio/pkg/config/labels"
@@ -196,6 +197,11 @@ func (cb *ClusterBuilder) applyDestinationRule(mc *MutableCluster, clusterMode C
 	// discovery type.
 	maybeApplyEdsConfig(mc.cluster)
 
+	// TODO do we need it in buildSubsetCluster?
+	if opts.proxy.Type == model.Router || opts.direction == model.TrafficDirectionOutbound {
+		cb.applyMetadataExchange(opts.mutable.cluster)
+	}
+
 	if destRule != nil {
 		mc.cluster.Metadata = util.AddConfigInfoMetadata(mc.cluster.Metadata, destRule.Meta)
 	}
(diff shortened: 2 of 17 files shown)


pilot/pkg/xds/filters/filters.go
```diff
@@ -154,6 +160,20 @@ var (
 		})},
 	}
 
+	TCPClusterMx = &cluster.Filter{
+		Name: MxFilterName,
+		TypedConfig: util.MessageToAny(&udpa.TypedStruct{
```
kyessenov, line 165: I think the filter config should be compiled-in now that's it's first-class.

pilot/pkg/xds/filters/filters.go
```diff
@@ -185,51 +205,139 @@ var (
...
+	}
+}
+
+func statsConfigForClass(class networking.ListenerClass) statsConfig {
+	switch class {
+	case networking.ListenerClassSidecarInbound:
+		return statsConfig{
+			Metrics: []*metricsConfig{{
+				Dimensions: map[string]string{
+					"destination_cluster": "node.metadata['CLUSTER_ID']",
```
kyessenov, line 286: Add a bug to make them first class.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -209,6 +215,13 @@ func (cb *ClusterBuilder) applyDestinationRule(mc *MutableCluster, clusterMode C
 	return subsetClusters
 }
 
+func (cb *ClusterBuilder) applyMetadataExchange(c *cluster.Cluster) {
+	spec := cb.push.Telemetry.EffectiveTelemetry(cb.proxy)
```
bianpengyuan, line 219: EnvoyFilter based solution applies TCP MX exchange filter regardless if the proxy has metrics enabled or not. I think the upstream peer should be able to get source metadata regardless of whether client has metric enabled or not? Otherwise upstream stats will have missing labels.

pilot/pkg/xds/filters/filters.go
bianpengyuan: I thikn they are already first classed https://github.com/istio/proxy/pull/3229. It was a mistake that they were not removed from inbound in this PR: https://github.com/istio/istio/pull/31226/files

pilot/pkg/networking/util/util.go
```diff
@@ -503,23 +499,21 @@ func BuildLbEndpointMetadata(networkID network.ID, tlsMode, workloadname, namesp
 	// server does not have sidecar injected, and request fails to reach server and thus metadata exchange does not happen.
 	// Due to performance concern, telemetry metadata is compressed into a semicolon separted string:
 	// workload-name;namespace;canonical-service-name;canonical-service-revision;cluster-id.
-	if features.EndpointTelemetryLabel {
```
bianpengyuan, line 506: Remove this feature flag completely?

pkg/config/mesh/mesh_test.go
```diff
@@ -164,7 +164,7 @@ defaultConfig:
 	if len(got.DefaultProviders.GetMetrics()) != 0 {
 		t.Errorf("default providers deep merge failed, got %v", got.DefaultProviders.GetMetrics())
 	}
-	if len(got.ExtensionProviders) != 2 {
+	if len(got.ExtensionProviders) <= 1 {
```
bianpengyuan, line 167: nit: would be great if we explicitly check the providers here.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -579,6 +594,7 @@ func (cb *ClusterBuilder) buildBlackHoleCluster() *cluster.Cluster {
 		ConnectTimeout:       gogo.DurationToProtoDuration(cb.push.Mesh.ConnectTimeout),
 		LbPolicy:             cluster.Cluster_ROUND_ROBIN,
 	}
+	cb.applyMetadataExchange(c)
```
bianpengyuan, line 597: Do we need MX for blackhole cluster? I know we probably add it via EnvoyFilter before, but does not seem to be necessary?

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -1292,9 +1282,17 @@ func buildHTTPConnectionManager(listenerOpts buildListenerOpts, httpOpts *httpLi
 
 	routerFilterCtx := configureTracing(listenerOpts, connectionManager)
 
+	spec := listenerOpts.push.Telemetry.EffectiveTelemetry(listenerOpts.proxy)
+	telemetry := model.TelemetryEnabled(spec, listenerOpts.push.Mesh)
 	filters := make([]*hcm.HttpFilter, len(httpFilters))
 	copy(filters, httpFilters)
 
+	// We add metadata exchange on inbound if any telemetry exists, as a client may expect metadata exchange.
+	// For outbound, we only need to set it if our own proxy has telemetry enabled.
```
bianpengyuan, line 1291: Same question, upstream will have missing labels if downstream disable telemetry.

pilot/pkg/xds/filters/filters.go
bianpengyuan: I thikn they are already first classed https://github.com/istio/proxy/pull/3229. It was a mistake that they were not removed from inbound in this PR: https://github.com/istio/istio/pull/31226/files

bianpengyuan:
Curious, I saw the recent PR went in to add CDS cache. I dont see any explicit logic here add telemetry bit to cache key calculation, so will this change work along with the cache?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
howardjohn: Good point. There were a lot of questionable filters in EnvoyFilter logic, I just matched them currently since it made testing a lot easier (diff of configdump). Open to chaning it though; this seems like a case where it clearly doesn't matter

pilot/pkg/networking/util/util.go
howardjohn: The problem is, previously we have all or nothing telemetry - either its enabled mesh wide, or not. Now, it can be enabled per workload. As a result, we don't know for which endpoints we need this label and for which we do not. We need it if the other side doesn't have MX filters, but we have no way to know that. Some bad options:
* use `tlsMode` label as a proxy for "has sidecar" and expect they will add MX. This won't work since its overloading `tlsMode` which may not always equal "has sidecar"
* Add a new label for "has sidecar" - avoided for complexity
* Check if any Telemetry exists, only add if so - doesn't actually work for multicluster, since we cannot know the full config of all pods in other clusters

So this will result in no change for any cluster using telemetry (API or Envoyfilter), but a moderate EDS perf cost when telemetry is completely disabled. This extra cost is not ideal, but I think better than the alternatives.

We could keep the flag and not set it in `manifests/` automatically I suppose, but it doesn't seem like something we want to keep around long term?

WDYT

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
howardjohn: I see, we can switch to AnyTelemetryExists then?

bianpengyuan: Yeah I think so.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
bianpengyuan: With EnvoyFilter, we lack of fine granular control on what to insert and what not. Since we are implementing it first class, we should do it right instead of matching EnvoyFilter.

pilot/pkg/networking/util/util.go
bianpengyuan: >  This extra cost is not ideal, but I think better than the alternatives.

Yeah I agree, we probably should just remove this flag. I think most large deployment should have telemetry enabled, and swallowed the performance impact. Always adding these metadata makes sense to me.

howardjohn:
> Curious, I saw the recent PR went in to add CDS cache. I dont see any explicit logic here add telemetry bit to cache key calculation, so will this change work along with the cache?

Yes. The cache already has proxy.Type and direction (part of cluster name) as keys. The only conditional thing we have on telemetry is IsAnyTelemetryEnabled, which is a pushcontext level setting, and pushcontext version is a part of the key.

Great question though - prior to addressing some other comments I think it was broken

pilot/pkg/model/telemetry.go
```diff
@@ -116,10 +124,33 @@ func shallowMerge(parent, child *tpb.Telemetry) *tpb.Telemetry {
 	}
 	merged := parent.DeepCopy()
 	shallowMergeTracing(merged, child)
+	shallowMergeMetrics(merged, child)
 	shallowMergeAccessLogs(merged, child)
 	return merged
 }
 
+func shallowMergeMetrics(parent *tpb.Telemetry, child *tpb.Telemetry) {
```
douglas-reid, line 132: nit: should we call this `mergeMetrics` in anticipation of the more nuanced merging that will happen?

pilot/pkg/model/telemetry.go
```diff
@@ -169,3 +200,39 @@ func shallowMergeAccessLogs(parent *tpb.Telemetry, child *tpb.Telemetry) {
...
+func PrometheusEnabled(providers []*meshconfig.MeshConfig_ExtensionProvider) bool {
+	for _, prov := range providers {
+		if _, ok := prov.Provider.(*meshconfig.MeshConfig_ExtensionProvider_Prometheus); ok {
+			return true
+		}
+	}
+	return false
+}
+
+func TelemetryProviders(spec *tpb.Telemetry, mesh *meshconfig.MeshConfig) []*meshconfig.MeshConfig_ExtensionProvider {
```
douglas-reid, line 217: nit: here and above, should we add comments on the exported methods? I ask here, because I feel like this is more of a `MetricsProviders` at the moment (is that right?).

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -209,6 +215,12 @@ func (cb *ClusterBuilder) applyDestinationRule(mc *MutableCluster, clusterMode C
 	return subsetClusters
 }
 
+func (cb *ClusterBuilder) applyMetadataExchange(c *cluster.Cluster) {
+	if cb.push.Telemetry.AnyTelemetryExists() {
```
douglas-reid, line 219: where does it makes sense to document this coupling?  should we put this in the API docs?

pilot/pkg/networking/core/v1alpha3/networkfilter.go
```diff
@@ -38,8 +40,43 @@ import (
 // redisOpTimeout is the default operation timeout for the Redis proxy filter.
 var redisOpTimeout = 5 * time.Second
 
+// ListenerClass defines the class of the listener
+type telemetryMode int
+
+const (
+	statsAndMetadataExchange telemetryMode = iota
```
douglas-reid, line 47: we will want MX for logs and traces potentially too, right?  for instance, i may want to log client workload name or similar from the server. does that impact this `telemetryMode` calculation?

douglas-reid (commented):
Thanks for tackling. Just had a few questions to clarify my own understanding.

pilot/pkg/networking/core/v1alpha3/networkfilter.go
howardjohn: I did not know we use MX for those? If so, it seems like the current EnvoyFilter implementation is broken since MX is only applied when telemetry.v2.enabled=true which is not related to logging or traces?

pilot/pkg/model/telemetry.go
howardjohn: ack, will do

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
howardjohn: I think its likely a short term "feature flag" enablement of sorts. Long term we will probably have a more graceful migration from envoyfilter to Telemetry API.

So probably wherever we document Telemetry as experimental, we mention how this works and that you need to disable EnvoyFitler to enable Telemetry

pilot/pkg/networking/core/v1alpha3/networkfilter.go
bianpengyuan: We need it for stackdriver log.

douglas-reid (approved):
I don't see anything that looks problematic, especially since this is just meant to establish a foothold for an experimental feature. Let's get this in, and iterate.

PR #33594, opened 2021-06-23
do not register delta xds env var

ramaraochavali:
Since Delta Xds is not fully implemented, registering this var adding to documentation and people wants to try it and see if it helps their config delivery. Either we should make it clear in the description that it is not ready for use or register it as hidden var so that docs are not generated. This PR moves it to hidden var.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -15,7 +15,6 @@
 package features
 
 import (
-	"strconv"
 	"time"
 
 	"github.com/golang/protobuf/ptypes/duration"
@@ -446,9 +445,9 @@ var (
 			"These checks are both expensive and panic on failure. As a result, this should be used only for testing.",
 	).Get()
 
-	DeltaXds = RegisterHiddenBoolVar("ISTIO_DELTA_XDS", false,
+	DeltaXds = env.RegisterBoolVar("ISTIO_DELTA_XDS", false,
 		"If enabled, pilot will only send the delta configs as opposed to the state of the world on a "+
-			"Resource Request").Get()
+			"Resource Request. This feature uses the delta xds api, but does not currently send the actual deltas.").Get()
 
 	EnableLegacyAutoPassthrough = env.RegisterBoolVar(
 		"PILOT_ENABLE_LEGACY_AUTO_PASSTHROUGH",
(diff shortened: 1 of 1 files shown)


pilot/pkg/features/pilot.go
```diff
@@ -445,7 +446,7 @@ var (
 			"These checks are both expensive and panic on failure. As a result, this should be used only for testing.",
 	).Get()
 
-	DeltaXds = env.RegisterBoolVar("ISTIO_DELTA_XDS", false,
+	DeltaXds = RegisterHiddenBoolVar("ISTIO_DELTA_XDS", false,
```
howardjohn, line 449: Aren't all env vars experimental? So I am not sure why we want to hide just this one. Delta XDS does actually work (just not efficiently, but correctly).

ramaraochavali: It is confusing to have it enable and it sends full config. I agree it is experimental but even in experimental people are expecting "delta" part works - That is the reason I was thinking about this. Otherwise if we do not want to hide, we need to expand the doc that right now "delta" part is not functional fully.

adiprerepa: How about we make it clear in the env var description that it does not fully work instead of registering it hidden? People should feel free to use expensive delta xds if they feel the need to

ramaraochavali: I am fine either way. Just want to make this clear that enabling delta right now does not do a "real" delta.

pilot/pkg/features/pilot.go
```diff
@@ -445,7 +446,7 @@ var (
...
 		"If enabled, pilot will only send the delta configs as opposed to the state of the world on a "+
 			"Resource Request").Get()
```
adiprerepa, line 451:
```suggestion
			"Resource Request. This feature uses the delta xds api, but does not currently send the actual deltas.").Get()
```

adiprerepa (changes requested):

ramaraochavali:
@adiprerepa Updated. PTAL

adiprerepa (approved):
lgtm -- thanks!

PR #33597, opened 2021-06-23
Updated URLs in Grafana dashboard

craigbox:
Updates URLs in the default Grafana mesh dashboard.

(Will this get pushed through to Grafana.com on each release?)

[X] Policies and Telemetry


howardjohn (approved):
> (Will this get pushed through to Grafana.com on each release?)

yes

eduardobaitello:
Just taking advantage of this MR to ask a question:

Does it make sense for all metrics on this dashboard to use  `reporter="source"` as filter?

I mean, thinking that the intention is to generate metrics of requests **received by the workloads**, wouldn't it be correct to filter by `reporter="destination"`?

PR #33609, opened 2021-06-24
Record push context init time

hzxuzhonghu:
It is helpful to do load test than debug log

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/xds/discovery.go
+++ b/pilot/pkg/xds/discovery.go
@@ -329,7 +329,9 @@ func (s *DiscoveryServer) Push(req *model.PushRequest) {
 	if err != nil {
 		return
 	}
-	pushContextInitTime.Record(time.Since(t0).Seconds())
+	initContextTime := time.Since(t0)
+	log.Debugf("InitContext %v for push took %s", versionLocal, initContextTime)
+	pushContextInitTime.Record(initContextTime.Seconds())
 
 	versionMutex.Lock()
 	version = versionLocal
--- a/pilot/pkg/xds/discovery.go
+++ b/pilot/pkg/xds/discovery.go
@@ -329,9 +329,7 @@ func (s *DiscoveryServer) Push(req *model.PushRequest) {
 	if err != nil {
 		return
 	}
-
-	initContextTime := time.Since(t0)
-	log.Debugf("InitContext %v for push took %s", versionLocal, initContextTime)
+	pushContextInitTime.Record(time.Since(t0).Seconds())
 
 	versionMutex.Lock()
 	version = versionLocal
--- a/pilot/pkg/xds/monitoring.go
+++ b/pilot/pkg/xds/monitoring.go
@@ -119,9 +119,9 @@ var (
 	rdsSendErrPushes = pushes.With(typeTag.Value("rds_senderr"))
 
 	pushContextInitTime = monitoring.NewDistribution(
-		"pilot_pushcontext_init_time",
+		"pilot_pushcontext_init_seconds",
 		"Total time in seconds Pilot takes to init pushContext.",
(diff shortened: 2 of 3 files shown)


pilot/pkg/xds/monitoring.go
```diff
@@ -118,6 +118,12 @@ var (
 	ldsSendErrPushes = pushes.With(typeTag.Value("lds_senderr"))
 	rdsSendErrPushes = pushes.With(typeTag.Value("rds_senderr"))
 
+	pushContextInitTime = monitoring.NewDistribution(
+		"pilot_pushcontext_init_time",
```
howardjohn, line 122: Based on https://prometheus.io/docs/practices/naming/#metric-names, I think this should be `pilot_pushcontext_init_seconds` perhaps?

cc @douglas-reid

pilot/pkg/xds/monitoring.go
```diff
@@ -118,6 +118,12 @@ var (
...
+		"Total time in seconds Pilot takes to init pushContext.",
+		[]float64{.01, .1, 1, 3, 5, 10, 20, 30},
```
howardjohn, line 124: `3, 5, 10, 20, 30` buckets seem pretty large, maybe we should have some more smaller buckets and less large ones?

pilot/pkg/xds/discovery.go
```diff
@@ -329,9 +329,7 @@ func (s *DiscoveryServer) Push(req *model.PushRequest) {
 	if err != nil {
 		return
 	}
-
-	initContextTime := time.Since(t0)
-	log.Debugf("InitContext %v for push took %s", versionLocal, initContextTime)
```
howardjohn, line 334: Can we keep the log? Metrics do not fully replace logs as they are aggregated

howardjohn (commented):
:+1: overall

pilot/pkg/xds/monitoring.go
douglas-reid: yes, let's use the units.

btw, @howardjohn are we going to use `pilot` forever?  or should new metrics get `istiod` prefixes?  wdyt?

howardjohn: Its a good question that I don't have a good answer to. It seems nice to have things be consistent, and I could see the argument that its the "pilot component within Istiod"... if we did it again surely it would be `istiod_` (or maybe even `istio_`), but might not be worth changing

hzxuzhonghu: We'd better not change istiod_xxx, those metrics are to end users, we should not change over and over.

pilot/pkg/xds/monitoring.go
hzxuzhonghu: SG

pilot/pkg/xds/discovery.go
hzxuzhonghu: Sure

pilot/pkg/xds/monitoring.go
douglas-reid: Yes, I agree it is too late to change existing metrics (at least, without a long transition). To me, the question is *new* metrics. I wonder if "pilot_" means anything to users at this point (particularly new users). I almost buy the component-within-istiod bit, but it feels like then it should be `istiod_pilot_....` (generally the single word prefix is the name of the application).

hzxuzhonghu: I donot think it should be so complex, just a name consistent with previous ones.

howardjohn (approved):
lgtm

PR #33610, opened 2021-06-24
Added comments to DiscoveryServer mutex

hzxuzhonghu:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/xds/discovery.go
+++ b/pilot/pkg/xds/discovery.go
@@ -106,12 +106,11 @@ type DiscoveryServer struct {
 	// incremental updates. This is keyed by service and namespace
 	EndpointShardsByService map[string]map[string]*EndpointShards
 
-	// pushChannel is the buffer used for debouncing,
-	// after debouncing the pushRequest will be send to pushQueue
+	// pushChannel is the buffer used for debouncing.
+	// after debouncing the pushRequest will be sent to pushQueue
 	pushChannel chan *model.PushRequest
 
-	// mutex used for Environment.PushContext
-	// TODO: is this necessary?
+	// mutex used for protecting Environment.PushContext
 	updateMutex sync.RWMutex
 
 	// pushQueue is the buffer that used after debounce and before the real xds push.
--- a/pilot/pkg/xds/discovery.go
+++ b/pilot/pkg/xds/discovery.go
@@ -90,9 +90,6 @@ type DiscoveryServer struct {
 	ProxyNeedsPush func(proxy *model.Proxy, req *model.PushRequest) bool
 
 	concurrentPushLimit chan struct{}
-	// mutex protecting global structs updated or read by ADS service, including ConfigsUpdated and
-	// shards.
-	mutex sync.RWMutex
 
 	// InboundUpdates describes the number of configuration updates the discovery server has received
 	InboundUpdates *atomic.Int64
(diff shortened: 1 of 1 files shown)


pilot/pkg/xds/discovery.go
```diff
@@ -103,13 +100,18 @@ type DiscoveryServer struct {
...
 	// incremental updates. This is keyed by service and namespace
 	EndpointShardsByService map[string]map[string]*EndpointShards
 
+	// pushChannel is the buffer used for debouncing,
+	// after debouncing the pushRequest will be send to pushQueue
 	pushChannel chan *model.PushRequest
 
-	// mutex used for config update scheduling (former cache update mutex)
+	// mutex used for Environment.PushContext
+	// TODO: is this necessary?
```
hzxuzhonghu, line 114: cc @ramaraochavali

pilot/pkg/xds/discovery.go
```diff
@@ -103,13 +100,18 @@ type DiscoveryServer struct {
...
 	CommittedUpdates *atomic.Int64
 
+	// mutex used for protecting shards.
+	mutex sync.RWMutex
 	// EndpointShards for a service. This is a global (per-server) list, built from
 	// incremental updates. This is keyed by service and namespace
 	EndpointShardsByService map[string]map[string]*EndpointShards
 
+	// pushChannel is the buffer used for debouncing,
+	// after debouncing the pushRequest will be send to pushQueue
```
ramaraochavali, line 110: nit: will be sent

pilot/pkg/xds/discovery.go
howardjohn: Yes it is. At the very least, it is required for caching to properly invalidate (https://github.com/istio/istio/blob/2c583b8567d361753d6be56d62cc63089407c406/pilot/pkg/xds/discovery.go#L523). But also probably required to prevent standard data races.

Note we have failing tests without the line linked

pilot/pkg/xds/discovery.go
```diff
@@ -103,13 +100,18 @@ type DiscoveryServer struct {
 	// the push context, which means that the next push to a proxy will receive this configuration.
 	CommittedUpdates *atomic.Int64
 
+	// mutex used for protecting shards.
+	mutex sync.RWMutex
 	// EndpointShards for a service. This is a global (per-server) list, built from
 	// incremental updates. This is keyed by service and namespace
 	EndpointShardsByService map[string]map[string]*EndpointShards
 
+	// pushChannel is the buffer used for debouncing,
```
adiprerepa, line 109:
```suggestion
	// pushChannel is the buffer used for debouncing.
```

pilot/pkg/xds/discovery.go
adiprerepa:
```suggestion
	// after debouncing, the pushRequest will be sent to pushQueue
```

pilot/pkg/xds/discovery.go
adiprerepa: there is also rarely a downside to having a mutex, just to be on the safe side.

hzxuzhonghu: PushContext is different, before each epoch of push, it is reinitiated. So it is worth revisit, so i left it there

hzxuzhonghu: @howardjohn  why need to protect dropCacheForRequest?  underlying cache has its own lock

howardjohn (approved):

pilot/pkg/xds/discovery.go
hzxuzhonghu: I believe you are right but seems this is very likely error-prone if someone later touched these part without so much knowledge.

PR #33611, opened 2021-06-24
Prevent stale cache

hzxuzhonghu:
For the CacheToken added, it is to prevent stale cache.

But there are still cases that it can be stale, e.g. A thread generates eds for a proxy runs super slow, but other threads for the other proxies run super fast.
```
v1 --> proxy2 eds pushed--> v2 --> (proxy1 eds pushed for v1, cache stale endpoint) -> proxy2 eds pushed -> proxy 1 eds push for v2
```

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/model/xds_cache.go
+++ b/pilot/pkg/model/xds_cache.go
@@ -22,7 +22,6 @@ import (
 	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	"github.com/google/go-cmp/cmp"
 	"github.com/hashicorp/golang-lru/simplelru"
-	"go.uber.org/atomic"
 	"google.golang.org/protobuf/testing/protocmp"
 
 	"istio.io/istio/pilot/pkg/features"
@@ -122,23 +121,14 @@ type CacheToken uint64
 // XdsCache interface defines a store for caching XDS responses.
 // All operations are thread safe.
 type XdsCache interface {
-	// Add adds the given XdsCacheEntry with the value to the cache. A token, returned from Get, must
-	// be included or writes will be (silently) dropped. Additionally, if the cache has been
-	// invalided between when a token is fetched from Get and when Add is called, the write will be
-	// dropped. This ensures stale data does not overwrite fresh data when dealing with concurrent
+	// Add adds the given XdsCacheEntry with the value for the given pushContext to the cache.
+	// If the cache has been updated to a newer push context, the write will be dropped silently.
+	// This ensures stale data does not overwrite fresh data when dealing with concurrent
 	// writers.
-	Add(entry XdsCacheEntry, token CacheToken, value *discovery.Resource)
+	Add(entry XdsCacheEntry, pushRequest *PushRequest, value *discovery.Resource)
 	// Get retrieves the cached value if it exists. The boolean indicates
 	// whether the entry exists in the cache.
-	//
-	// A CacheToken is additionally included in the response. This must be used for subsequent writes
-	// to this key. This ensures that if the cache is invalidated between our read and write, we do
-	// not persist stale data.
-	//
-	// Standard usage:
-	// if obj, token, f := cache.Get(key); f { ...do something... }
-	// else { computed := expensive(); cache.Add(key, token, computed); }
-	Get(entry XdsCacheEntry) (*discovery.Resource, CacheToken, bool)
+	Get(entry XdsCacheEntry) (*discovery.Resource, bool)
 	// Clear removes the cache entries that are dependent on the configs passed.
 	Clear(map[ConfigKey]struct{})
 	// ClearAll clears the entire cache.
@@ -156,7 +146,6 @@ func NewXdsCache() XdsCache {
 		store:            newLru(),
 		configIndex:      map[ConfigKey]sets.Set{},
 		typesIndex:       map[config.GroupVersionKind]sets.Set{},
-		nextToken:        atomic.NewUint64(0),
 	}
 }
 
@@ -167,16 +156,15 @@ func NewLenientXdsCache() XdsCache {
 		store:            newLru(),
 		configIndex:      map[ConfigKey]sets.Set{},
 		typesIndex:       map[config.GroupVersionKind]sets.Set{},
-		nextToken:        atomic.NewUint64(0),
 	}
 }
 
 type lruCache struct {
 	enableAssertions bool
 	store            simplelru.LRUCache
-	// nextToken stores the next token to use. The content here doesn't matter, we just need a cheap
-	// unique identifier.
-	nextToken   *atomic.Uint64
+	// token stores the latest token of the store, used to prevent stale data overwrite.
+	// It is refreshed when Clear or ClearAll are called
+	token       CacheToken
 	mu          sync.RWMutex
 	configIndex map[ConfigKey]sets.Set
 	typesIndex  map[config.GroupVersionKind]sets.Set
@@ -222,36 +210,39 @@ func (l *lruCache) assertUnchanged(key string, existing *discovery.Resource, rep
 	}
 }
 
-func (l *lruCache) Add(entry XdsCacheEntry, token CacheToken, value *discovery.Resource) {
-	if !entry.Cacheable() {
+func (l *lruCache) Add(entry XdsCacheEntry, pushReq *PushRequest, value *discovery.Resource) {
+	if !entry.Cacheable() || pushReq == nil {
 		return
 	}
+	// Start unset
+	if pushReq.Start.Equal(time.Time{}) {
+		panic("token cannot be empty!")
+	}
+	// TODO: It will not overflow until year 2262
+	token := CacheToken(pushReq.Start.UnixNano())
 	l.mu.Lock()
 	defer l.mu.Unlock()
 	k := entry.Key()
 	cur, f := l.store.Get(k)
-	toWrite := cacheValue{value: value}
 	if f {
-		if token != cur.(cacheValue).token {
+		// This is the stale or same resource
+		if token <= cur.(cacheValue).token || token <= l.token {
 			// entry may be stale, we need to drop it. This can happen when the cache is invalidated
 			// after we call Get.
 			return
 		}
-		// Otherwise, make sure we write the current token again. We don't change the key on writes; the
-		// same token will be used for a value until its invalidated
-		toWrite.token = cur.(cacheValue).token
-	} else {
-		// This is our first time seeing this; this means it was invalidated recently and this is our
-		// first write, or we forgot to call Get before.
+		if l.enableAssertions {
+			l.assertUnchanged(k, cur.(cacheValue).value, value)
+		}
+	}
+
+	if token < l.token {
 		return
 	}
-	if l.enableAssertions {
-		if toWrite.token == 0 {
-			panic("token cannot be empty. was Get() called before Add()?")
-		}
-		l.assertUnchanged(k, cur.(cacheValue).value, value)
-	}
+
+	toWrite := cacheValue{value: value, token: token}
 	l.store.Add(k, toWrite)
+	l.token = token
 	indexConfig(l.configIndex, k, entry)
 	indexType(l.typesIndex, k, entry)
 	size(l.store.Len())
@@ -262,9 +253,9 @@ type cacheValue struct {
 	token CacheToken
 }
 
-func (l *lruCache) Get(entry XdsCacheEntry) (*discovery.Resource, CacheToken, bool) {
+func (l *lruCache) Get(entry XdsCacheEntry) (*discovery.Resource, bool) {
 	if !entry.Cacheable() {
-		return nil, 0, false
+		return nil, false
 	}
 	l.mu.Lock()
 	defer l.mu.Unlock()
@@ -272,28 +263,21 @@ func (l *lruCache) Get(entry XdsCacheEntry) (*discovery.Resource, CacheToken, bo
 	val, ok := l.store.Get(k)
 	if !ok {
 		miss()
-		// If the entry is not found at all, this is our first read of it. We will generate and store
-		// a new token. Subsequent writes must include it.
-		tok := CacheToken(l.nextToken.Inc())
-		l.store.Add(k, cacheValue{token: tok})
-		indexConfig(l.configIndex, k, entry)
-		indexType(l.typesIndex, k, entry)
-		return nil, tok, false
+		return nil, false
 	}
 	cv := val.(cacheValue)
 	if cv.value == nil {
 		miss()
-		// We have generated a token previously, so return that, but this is still a cache miss as
-		// no value is stored.
-		return nil, cv.token, false
+		return nil, false
 	}
 	hit()
-	return cv.value, cv.token, true
+	return cv.value, true
 }
 
 func (l *lruCache) Clear(configs map[ConfigKey]struct{}) {
 	l.mu.Lock()
 	defer l.mu.Unlock()
+	l.token = CacheToken(time.Now().UnixNano())
 	for ckey := range configs {
 		referenced := l.configIndex[ckey]
 		delete(l.configIndex, ckey)
@@ -312,6 +296,7 @@ func (l *lruCache) Clear(configs map[ConfigKey]struct{}) {
 func (l *lruCache) ClearAll() {
 	l.mu.Lock()
 	defer l.mu.Unlock()
+	l.token = CacheToken(time.Now().UnixNano())
 	l.store.Purge()
 	l.configIndex = map[ConfigKey]sets.Set{}
 	l.typesIndex = map[config.GroupVersionKind]sets.Set{}
@@ -350,10 +335,10 @@ type DisabledCache struct{}
 
 var _ XdsCache = &DisabledCache{}
 
-func (d DisabledCache) Add(key XdsCacheEntry, token CacheToken, value *discovery.Resource) {}
+func (d DisabledCache) Add(key XdsCacheEntry, pushReq *PushRequest, value *discovery.Resource) {}
 
-func (d DisabledCache) Get(XdsCacheEntry) (*discovery.Resource, CacheToken, bool) {
-	return nil, 0, false
+func (d DisabledCache) Get(XdsCacheEntry) (*discovery.Resource, bool) {
+	return nil, false
 }
 
 func (d DisabledCache) Clear(configsUpdated map[ConfigKey]struct{}) {}
--- a/pilot/pkg/model/xds_cache.go
+++ b/pilot/pkg/model/xds_cache.go
@@ -211,14 +211,10 @@ func (l *lruCache) assertUnchanged(key string, existing *discovery.Resource, rep
 }
 
 func (l *lruCache) Add(entry XdsCacheEntry, pushReq *PushRequest, value *discovery.Resource) {
-	if !entry.Cacheable() || pushReq == nil {
+	if !entry.Cacheable() || pushReq == nil || pushReq.Start.Equal(time.Time{}) {
 		return
 	}
-	// Start unset
-	if pushReq.Start.Equal(time.Time{}) {
-		panic("token cannot be empty!")
-	}
-	// TODO: It will not overflow until year 2262
+	// It will not overflow until year 2262
 	token := CacheToken(pushReq.Start.UnixNano())
 	l.mu.Lock()
 	defer l.mu.Unlock()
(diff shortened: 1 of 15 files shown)


hzxuzhonghu:
Still need to fix tests

pilot/pkg/xds/eds.go
```diff
@@ -418,7 +419,8 @@ func (eds *EdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
 				Resource: util.MessageToAny(l),
 			}
 			resources = append(resources, resource)
-			eds.Server.Cache.Add(builder, token, resource)
+			token := parsePushVersion(push.PushVersion)
```
howardjohn, line 422: push version does not contain the full state of the world. Endpoints can change without it changing

howardjohn (changes requested):
Can you give some more info about what this fixes? From your description I am a bit confused, as the case you showed is exactly what the token part is meant to address. We have, IMO, extremely good tests here with the EnableUnsafeAssertions setting. With that enabled, we found 10+ extremely subtle races that were fixed by the token, and finally have passing all tests with that enabled. If there is a bug that is not caught by that, it would be good to have a concrete example so we can improve it.

I suspect with this change, EnableUnsafeAssertions=true will start to fail tests.

hzxuzhonghu:
I can write a test to show you the cache is not correct.

pilot/pkg/xds/eds.go
hzxuzhonghu: you reminded me, that's a little trouble

hzxuzhonghu:
@howardjohn THis is the time sequence I want to show, maybe my description above is not clear. https://github.com/istio/istio/compare/master...hzxuzhonghu:xds-cache?expand=1 Hope this can be more understandable

pilot/pkg/xds/eds.go
hzxuzhonghu: We can use the PushRequest.Start as the token, it is always increasing

howardjohn:
L249 should not happen. it cannot get v1 after v2 push
context/endpointshard has been created as we always grab the latest. Any
time some one has gotten v1 bit v2 is available their token should
invalidate

hzxuzhonghu:
EDS is special, how about cds, lds, rds?

pilot/pkg/model/xds_cache.go
```diff
@@ -167,16 +157,14 @@ func NewLenientXdsCache() XdsCache {
...
 	}
 }
 
 type lruCache struct {
 	enableAssertions bool
 	store            simplelru.LRUCache
-	// nextToken stores the next token to use. The content here doesn't matter, we just need a cheap
-	// unique identifier.
-	nextToken   *atomic.Uint64
+	// token stores the latest token of the store, used to prevent stale data overwrite.
```
hzxuzhonghu, line 166: With this v1 can not overwrite v2

howardjohn:
For other types I was thinking of putting the push version in the key: https://github.com/istio/istio/pull/33338/files#diff-e3b667b806175fc66cb679217641e1638639507c364df2432a60c1d4e87a9739R138? It seems to work ok

hzxuzhonghu:
What about his general way, it doesn't care whether it is eds or cds, what's more no `Get` is required before `Add`. It is more of the cache side fix.

pilot/pkg/model/xds_cache.go
```diff
@@ -230,28 +218,27 @@ func (l *lruCache) Add(entry XdsCacheEntry, token CacheToken, value *discovery.R
...
-	if l.enableAssertions {
-		if toWrite.token == 0 {
-			panic("token cannot be empty. was Get() called before Add()?")
+		if l.enableAssertions {
+			l.assertUnchanged(k, cur.(cacheValue).value, value)
 		}
-		l.assertUnchanged(k, cur.(cacheValue).value, value)
 	}
+
+	if token == 0 || token < l.token {
```
howardjohn, line 233: nit: should we add the 
```
		if toWrite.token == 0 {
			panic("token cannot be empty. was Get() called before Add()?")
		}
```

to ensure we do not ever forget the token?

pilot/pkg/model/xds_cache.go
```diff
@@ -230,28 +218,27 @@ func (l *lruCache) Add(entry XdsCacheEntry, token CacheToken, value *discovery.R
...
+		return
+	}
+
+	// token cannot be empty.
+	toWrite := cacheValue{value: value, token: token}
+
 	l.store.Add(k, toWrite)
+	l.token = token
```
howardjohn, line 241: I would have thought this is a data race.. how come this doesn't fail tests?

pilot/pkg/model/xds_cache.go
```diff
@@ -262,31 +249,27 @@ type cacheValue struct {
...
-		// a new token. Subsequent writes must include it.
-		tok := CacheToken(l.nextToken.Inc())
-		l.store.Add(k, cacheValue{token: tok})
-		return nil, tok, false
+		return nil, false
 	}
 	cv := val.(cacheValue)
 	if cv.value == nil {
 		miss()
 		// We have generated a token previously, so return that, but this is still a cache miss as
```
howardjohn, line 267: nit: comment is stale now I think

pilot/pkg/xds/eds.go
```diff
@@ -418,7 +418,7 @@ func (eds *EdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
 				Resource: util.MessageToAny(l),
 			}
 			resources = append(resources, resource)
-			eds.Server.Cache.Add(builder, token, resource)
+			eds.Server.Cache.Add(builder, model.CacheToken(req.Start.Nanosecond()), resource)
```
howardjohn, line 421: nit: maybe Add should take a PushRequest object so each user doesn't need to know it must use req.Start.Nanosecond?

howardjohn: Also we will need to do some refactoring in CDS to pass push request through for https://github.com/istio/istio/pull/33338, and in the future lds/rds. Shouldn't be a problem, just a note

pilot/pkg/model/xds_cache.go
```diff
@@ -130,15 +129,7 @@ type XdsCache interface {
 	Add(entry XdsCacheEntry, token CacheToken, value *discovery.Resource)
 	// Get retrieves the cached value if it exists. The boolean indicates
 	// whether the entry exists in the cache.
```
howardjohn, line 131: This looks like it has the opportunity to be a big improvement - right now the cache is definitely very complex.

One concern I have - is the push request start time sufficient? For example, the EndpointShards may change during the middle of a push? For example:

```
|Push Request| Endpoint State | Thread 1            | Thread 2            |
|------------|----------------|---------------------|---------------------|
| Push 1     | v1             |                     |                     |
|            | v1             |                     |                     |
|            |                | Start generating v1 |                     |
|            | v2             |                     |                     |
|            |                |                     | Start generating v2 |
|            |                |                     | Write v2, token=1   |
|            |                | Write v1, token=1   |                     |
```

I think in this case we end up with endpoint state `v1` since the write isn't invalidated. Both writes use the same token but have different state

pilot/pkg/model/xds_cache.go
hzxuzhonghu: With this patch, `Get` is not needed before `Add`

howardjohn: We can change the text to just say "token cannot be empty" right? its still invalid to have no token?

pilot/pkg/model/xds_cache.go
hzxuzhonghu: l.token is protected, and has this interception 
`	if token == 0 || token < l.token {
		return
	}`

pilot/pkg/model/xds_cache.go
hzxuzhonghu: will remove

pilot/pkg/model/xds_cache.go
howardjohn: forgot the whole thing is locked, thanks

pilot/pkg/xds/eds.go
hzxuzhonghu: OK, in case others miss the intention here.

pilot/pkg/model/xds_cache.go
hzxuzhonghu: In this case the v2 push.StartTime > v1 push.StartTime, so the `Write V1` will be stopped by 

```
	if token == 0 || token < l.token {
		return
	}

```

hzxuzhonghu: l.token = v2 push.StartTime, 
token =   v1 push.StartTime

howardjohn: In my example they are in the same push. The endpoint state (endpoint shards) can change during the middle of a push

pilot/pkg/model/xds_cache.go
hzxuzhonghu: Yes, token must be provided, otherwise equals to 0

pilot/pkg/model/xds_cache.go
hzxuzhonghu: v1 and v2 event enqueued to pushChannel and dequeued from it is in order, right? 

And then they will be set StartTime (v1<v2)and  enqueued to pushQueue, even later v2 is processed (let's say generate eds)before v1, it will end up caching v2.

hzxuzhonghu: > The endpoint state (endpoint shards) can change during the middle of a push

It does not influence the correctness, for every ep shard change, it will trigger a new round of eds push at least. The xds cache will then be renewed too.

howardjohn: I think we may be mixing up "Push version" (ie a PushRequest) which has the start time, and the "endpoint version". We do not literally have an "endpoint version" (although we could easily add one: see https://github.com/istio/istio/pull/33620/files#diff-c6215f1e2a80c0d2afe120f4616972dfb339bcb51e8815cc5e171750bc11ae18R154) in the code, but there is a logical version that we can consider to be incremented each time edsUpdate occurs.

In the example where I think things may not work above, there is only every a single push.

Long sequence of events:
* Service Z scales from 0 -> 1. We consider this "Endpoint version v1". This event triggers a pushChannel update
* Push debounce completes (from "endpoint version v1" event), we start a push with StartTime=1
* For proxy-1, we start eds generate. When we fetch the endpoint shards, we get the endpoint version v1
* Service Z scales from 1 -> 2. We consider this "Endpoint version v2". This event triggers a pushChannel update
* For proxy-2, we start eds generate. Note: this is still from the push with StartTime=1
* proxy-2 completes EDS generation. It used "endpoint version v2" to compute this, so we have 2 endpoints. We write this to the cache. _the write succeeds since the token is not less than the current token (both are 1 from push with StartTime=1)_
* proxy-1 completes EDS generation. It used "endpoint version v1" to compute this, so we have 1 endpoint. We write this to the cache. The write succeeds since it was the first write, so no token issue. Token=1 (from StartTime). **BUG: we went backwards**
* Push debounce completes (from "endpoint version v2" event), we start a push with StartTime=2. This clears the cache.

pilot/pkg/model/xds_cache.go
hzxuzhonghu: done

pilot/pkg/model/xds_cache.go
hzxuzhonghu: >proxy-1 completes EDS generation. It used "endpoint version v1" to compute this, so we have 1 endpoint. We write this to the cache. The write succeeds since it was the first write, so no token issue. Token=1 (from StartTime).

You are right, let me think about it

hzxuzhonghu: It seems tricky, I think https://github.com/istio/istio/blob/4477ee5d75e64fcc0b7e890484278fa535ba4907/pilot/pkg/xds/eds.go#L152-L159 and here https://github.com/istio/istio/blame/4477ee5d75e64fcc0b7e890484278fa535ba4907/pilot/pkg/xds/eds.go#L191 are both for the eds cache.

How about adding a flag in xDSCache to control whether to allow `Add` with the same token the cache has? When clear the cache, we mark the flag = true, and then set a data like eds with the same token the cache already has, prevent it.  only allow cache with a newer token.

howardjohn: Hmm.. I thought the goal of this PR was to avoid all of these places where we have to clear the cache under locks because its extremely complex (which I agree with). If we do it this way, then we still have the same issue so I am not sure its much simpler?

hzxuzhonghu: With it we can remove clear cache in different places, and avoid Get before Add.

And likely the above case https://github.com/istio/istio/pull/33611#discussion_r660251243 can be resolved

hzxuzhonghu: Only do this once in `dropCacheForRequest` is enough.

pilot/pkg/model/xds_cache.go
```diff
@@ -167,19 +156,21 @@ func NewLenientXdsCache() XdsCache {
...
-	nextToken   *atomic.Uint64
-	mu          sync.RWMutex
-	configIndex map[ConfigKey]sets.Set
-	typesIndex  map[config.GroupVersionKind]sets.Set
+	// token stores the latest token of the store, used to prevent stale data overwrite.
+	token CacheToken
+	// It is set when Clear or ClearAll are called, indicates `Add` a cache with same token above is disallowed.
+	// It is reset when after a successful `Add` call, in this case the token will be incremented.
+	// So it can prevent old cache recorded combined with token.
+	disallowCacheWithSameToken bool
```
hzxuzhonghu, line 170: This is to guarantee after a new pushReq, we donot allow cache outdated data from an old version push.

pilot/pkg/model/xds_cache.go
hzxuzhonghu: An alternative way is to set a newer token when calling clear cache, and directly reject cache with elder token.

hzxuzhonghu: Changed to this

hzxuzhonghu:
@howardjohn I think all the above comments have been addressed. Any other concern?

hzxuzhonghu:
ping @howardjohn in case you forget it

pilot/pkg/model/xds_cache.go
```diff
@@ -222,36 +210,39 @@ func (l *lruCache) assertUnchanged(key string, existing *discovery.Resource, rep
...
 
-func (l *lruCache) Add(entry XdsCacheEntry, token CacheToken, value *discovery.Resource) {
-	if !entry.Cacheable() {
+func (l *lruCache) Add(entry XdsCacheEntry, pushReq *PushRequest, value *discovery.Resource) {
+	if !entry.Cacheable() || pushReq == nil {
 		return
 	}
+	// Start unset
+	if pushReq.Start.Equal(time.Time{}) {
+		panic("token cannot be empty!")
```
howardjohn, line 219: nit: Not sure we want to panic here, since cache is optional. Would not want to make all istiod panic due to a bug

pilot/pkg/model/xds_cache.go
```diff
@@ -222,36 +210,39 @@ func (l *lruCache) assertUnchanged(key string, existing *discovery.Resource, rep
...
-	if l.enableAssertions {
-		if toWrite.token == 0 {
-			panic("token cannot be empty. was Get() called before Add()?")
+		if l.enableAssertions {
+			l.assertUnchanged(k, cur.(cacheValue).value, value)
 		}
-		l.assertUnchanged(k, cur.(cacheValue).value, value)
 	}
+
+	if token < l.token {
```
howardjohn, line 239: nit: above we have `<=` and here we have `<`. Is that intended?

pilot/pkg/model/xds_cache.go
```diff
@@ -222,36 +210,39 @@ func (l *lruCache) assertUnchanged(key string, existing *discovery.Resource, rep
...
-	} else {
-		// This is our first time seeing this; this means it was invalidated recently and this is our
-		// first write, or we forgot to call Get before.
-		return
-	}
-	if l.enableAssertions {
-		if toWrite.token == 0 {
-			panic("token cannot be empty. was Get() called before Add()?")
+		if l.enableAssertions {
+			l.assertUnchanged(k, cur.(cacheValue).value, value)
```
howardjohn, line 235: I think this no longer does anything since L229 should always cause it to return

pilot/pkg/xds/delta.go
```diff
@@ -310,6 +310,7 @@ func (s *DiscoveryServer) processDeltaRequest(req *discovery.DeltaDiscoveryReque
 	}
 
 	push := s.globalPushContext()
+	request.Start = time.Now()
```
howardjohn, line 313: Rewriting Start here is completely changing the meaning of Start, which will impact metrics, etc. The field already indicates `This represents the time of adding to the PushQueue.` which is no longer accurate

pilot/pkg/model/xds_cache.go
```diff
@@ -222,36 +210,39 @@ func (l *lruCache) assertUnchanged(key string, existing *discovery.Resource, rep
...
+func (l *lruCache) Add(entry XdsCacheEntry, pushReq *PushRequest, value *discovery.Resource) {
+	if !entry.Cacheable() || pushReq == nil {
 		return
 	}
+	// Start unset
+	if pushReq.Start.Equal(time.Time{}) {
+		panic("token cannot be empty!")
+	}
+	// TODO: It will not overflow until year 2262
+	token := CacheToken(pushReq.Start.UnixNano())
```
howardjohn, line 222: In the old token model, tokens were generated+read under a consistent lock. This guaranteed consistency in them.

In the new model, the token is generated by calling time.Now() across many goroutines. Does go guarantee that time.Now is monotonically increasing *across goroutines*?

If not, we may run into issues

howardjohn (commented):
Overall looks like a good simplification, a few small comments

pilot/pkg/model/xds_cache.go
howardjohn: After fixing this to assert properly (I think - maybe my change was bad) by changing `<=` to `<`, it failed a lot

pilot/pkg/model/xds_cache.go
hzxuzhonghu: OK, i can just return here. In normal case, this can not happen, just for tests

pilot/pkg/xds/delta.go
hzxuzhonghu: good catch

pilot/pkg/model/xds_cache.go
hzxuzhonghu: It is not monotonically increasing, the StartTime is set in a debouncing, single thread. In 
ProxyUpdate, it can also be set, i donot think it matters even if duplicate time used.

pilot/pkg/model/xds_cache.go
hzxuzhonghu: Yes, current the implementation is to update the token when we call the `Clear`, see comments here https://github.com/istio/istio/pull/33611/files#r661912888

pilot/pkg/model/xds_cache.go
howardjohn: Then we need some other way to trigger the assertUnchanged. Right now it will never be triggered, so we have no test coverage of the change here?

hzxuzhonghu: looking into this

hzxuzhonghu:
@howardjohn Figured out the panic caused by the last two commits (introduced before), and I have separate fixes, which should be cherrypicked to 1.11

hzxuzhonghu:
@howardjohn Fixed

pilot/pkg/model/xds_cache.go
hzxuzhonghu: This is caused by the bugs from other pr, which are fixed now

howardjohn (commented):
Sorry didn' get a chance to look today, will review first thing tomorrow

pilot/pkg/xds/ads.go
```diff
@@ -231,7 +232,6 @@ func (s *DiscoveryServer) processRequest(req *discovery.DiscoveryRequest, con *C
 		}
 	}
 
-	push := s.globalPushContext()
 	request.Reason = append(request.Reason, model.ProxyRequest)
 	return s.pushXds(con, push, versionInfo(), con.Watched(req.TypeUrl), request)
```
howardjohn, line 236: nit: if request always has the push, we can remove `push` argument from pushXds?

howardjohn (approved):
/hold

to run assert test a few times

hzxuzhonghu:
This command is cool

PR #33612, opened 2021-06-24
Fixed incorrect No data in Grafana dashboard panels

msvticket:
Many panels show No data if Prometheus has a scrape interval configured to be larger than 15s. Using
the `$__rate_interval` introduced in Grafana 7.2 fixes just that.

[ ] Docs
[ ] Installation
[ ] Networking
[X] Policies and Telemetry
[ ] Security
[ ] Test and Release
[X] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/releasenotes/notes/grafana-rate-interval.yaml
+++ b/releasenotes/notes/grafana-rate-interval.yaml
@@ -30,7 +30,7 @@ issue: []
 # release notes.
 releaseNotes:
 - |
-  **Fixed** that many panels in the Grafana dashboards showed **No data** if Prometheus had a scrape
+  **Fixed** an issue where many panels in the Grafana dashboards showed **No data** if Prometheus had a scrape
   interval configured to be larger than `15s`.
 
 # upgradeNotes is a markdown listing of any changes that will affect the upgrade
--- a/releasenotes/notes/grafana-rate-interval.yaml
+++ b/releasenotes/notes/grafana-rate-interval.yaml
@@ -31,7 +31,7 @@ issue: []
 releaseNotes:
 - |
   **Fixed** that many panels in the Grafana dashboards showed **No data** if Prometheus had a scrape
-  interval configured to be larger than 15s.
+  interval configured to be larger than `15s`.
 
 # upgradeNotes is a markdown listing of any changes that will affect the upgrade
 # process. This will appear in the release notes.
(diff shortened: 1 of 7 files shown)


msvticket:
@googlebot I signed it!

releasenotes/notes/grafana-rate-interval.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+# issue is a list of GitHub issues resolved in this note.
+# If issue is not in the current repo, specify its full URL instead.
+issue: []
+
+# releaseNotes is a markdown listing of any user facing changes. This will appear in the
+# release notes.
+releaseNotes:
+- |
+  **Fixed** that many panels in the Grafana dashboards showed **No data** if Prometheus had a scrape
```
shamsher31, line 33:
```suggestion
  **Fixed** an issue where many panels in the Grafana dashboards showed **No data** if Prometheus had a scrape
```

releasenotes/notes/grafana-rate-interval.yaml
```diff
@@ -0,0 +1,46 @@
...
+  interval configured to be larger than 15s.
```
shamsher31, line 34:
```suggestion
  interval configured to be larger than `15s`.
```

howardjohn:
One thing we should look into is how many users are on 7.1 or older. I have no idea currently - could be 1% or 99%

shamsher31:
@googlebot I consent

douglas-reid:
@howardjohn anything more to review here (i'm just digging through my backlog).

howardjohn:
I think we need to understand impact of Grafana version requirement

> One thing we should look into is how many users are on 7.1 or older. I have no idea currently - could be 1% or 99%

douglas-reid:
> I think we need to understand impact of Grafana version requirement
> 
> > One thing we should look into is how many users are on 7.1 or older. I have no idea currently - could be 1% or 99%

Do we have any updated insights here?  

@msvticket are there ways to achieve the same goal with older primitives?

msvticket:
> > I think we need to understand impact of Grafana version requirement
> > > One thing we should look into is how many users are on 7.1 or older. I have no idea currently - could be 1% or 99%
> 
> Do we have any updated insights here?
> 
> @msvticket are there ways to achieve the same goal with older primitives?

No.

My rationale for it not being that much of a problem to require 7.2 is that if you install Grafana using istio you will as far as I understand currently (master) get version 7.5.5. But it would be good to add some more information also in the other main distribution route for these Dashboards: https://grafana.com/grafana/dashboards

I'm thinking of something along the lines of:

> Check revisions to get the dashboard corresponding to your version of istio. Not that from version 1.11.? of istio the Dashboards require Grafana 7.2 or newer.
> See https://istio.io/latest/docs/ops/integrations/grafana/ for more details on the integration of istio with Grafana.

PR #33628, opened 2021-06-24
Simplify kube controller network gateway logic

nmittler:
Now that `NetworkGateways` just returns a list of gateways, we can simplify the internal data structures.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/network.go
+++ b/pilot/pkg/serviceregistry/kube/controller/network.go
@@ -16,7 +16,10 @@ package controller
 
 import (
 	"net"
+	"reflect"
+	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/yl2chen/cidranger"
 
@@ -31,7 +34,7 @@ type namedRangerEntry struct {
 	network net.IPNet
 }
 
-// returns the IPNet for the network
+// Network returns the IPNet for the network
 func (n namedRangerEntry) Network() net.IPNet {
 	return n.network
 }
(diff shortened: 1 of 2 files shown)


pilot/pkg/serviceregistry/kube/controller/network.go
```diff
@@ -278,3 +275,44 @@ func (c *Controller) getNodePortGatewayServices() []*model.Service {
...
+
+func (s gatewaySet) toArray() []*model.NetworkGateway {
+	gws := make([]*model.NetworkGateway, 0, len(s))
+	for gw := range s {
+		gw := gw
+		gws = append(gws, &gw)
+	}
+
+	// Sort the array so that it's stable.
+	sort.SliceStable(gws, func(i, j int) bool {
```
stevenctl, line 304: Why do we need them sorted? It sounds like a familiar requirement somewhere but I forgot..

And if we do want them sorted, would it make more sense to sort in the place where we aggregate them?

stevenctl (approved):

pilot/pkg/serviceregistry/kube/controller/network.go
nmittler: We probably don't actually need to sort. I just generally prefer to add stability to the ordering of the returned list.

WRT how we sort. I'm not sure it matters much ... there might be a slight performance improvement by sorting as we add, but there honestly won't be that many gateways to make a difference in practice. This code is simpler.

PR #33620, opened 2021-06-24
WIP: do not merge: more assertion tests

howardjohn:

PR #33629, opened 2021-06-24
 Add cluster to SDS cache key

howardjohn:


adiprerepa (commented):
why are we doing this? I don't have enough context

howardjohn:
Seeing some issues in https://github.com/istio/istio/pull/33620. Basically if the same secret name with different contents is in different cluster we may cache the wrong one

pilot/pkg/xds/sds.go
```diff
@@ -42,10 +42,11 @@ type SecretResource struct {
 	Name         string
 	Namespace    string
 	ResourceName string
+	Cluster      string
 }
 
 func (sr SecretResource) Key() string {
-	return "sds://" + sr.Type + "/" + sr.Name + "/" + sr.Namespace
+	return "sds://" + sr.Type + "/" + sr.Name + "/" + sr.Namespace + "/" + sr.Cluster
```
adiprerepa, line 49: it doesn't matter, but doesn't it make more sense for the ordering to be:

```
sds://<Cluster>/<Namespace>/<Type>/<Name>
```

adiprerepa (approved):
lgtm

Issue #33635, opened 2021-06-25
pilot-agent in vm use an incorrect serviceCluster


kebe7jun:
**Bug description**

VM's pilot-agent with an incorrect serviceCluster flag.

See:

https://github.com/istio/istio/blob/f37b1cea5ac8cf01b65eabdd4e25fa4f49375bb0/pilot/cmd/pilot-agent/main.go#L174

pilot-agent with a default serviceCluster flag, which makes serviceCluster in env is not active.

**Affected product area (please put an X in all that apply)**

[X] Installation

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[X] Virtual Machine
[ ] Multi Control Plane

PR #33636, opened 2021-06-25
Add serviceCluster to pilot-agent flag

kebe7jun:
Fix https://github.com/istio/istio/issues/33635

[X] Installation


howardjohn:
@kyessenov IIRC you made some changes here recently?

adiprerepa (commented):
lgtm but yielding

howardjohn:
This is reverting https://github.com/istio/istio/pull/32826 more or less I think, so adding @kyessenov as approver

kyessenov:
/approve
It's not reverting, the referenced PR is focused on k8s while this is for VMs. There is no injection template for VMs so it's fine to do whatever.

howardjohn:
Where is sidecar getting service cluster? If its proxy config, then that is also available for VMs. It feels a bit off to use flag for VM and not for sidecar, especially since the flag is marked as deprecated

PR #33644, opened 2021-06-25
Add IST0139 analysis for install workflow

### LINKED ISSUE #33537: Run IST0139 before install
```
istioctl x revision tag set default --revision=foo
istioctl install
```

Will result in conflicting webhooks. We run IST0139 to detect this if you do the other order, but we should detect it in `install` as well to prevent this

akshitgrover:
FilterRenderedComponentManifest method allows to filter particular k8s object from rendered istio component manifest

Signed-off-by: Akshit Grover <akshit.grover2016@gmail.com>

Fixes #33537 

This is WIP PR, First part of getting the correct data to analyse using IST0139 ie. MutatingWebhookConfiguration is implemented.
The second part of analysing the same using webhook analyser is yet to be implemented.

[ ] Docs
[ X ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ X ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/operator/pkg/helmreconciler/apply.go
+++ b/operator/pkg/helmreconciler/apply.go
@@ -17,15 +17,6 @@ package helmreconciler
 import (
 	"context"
 	"fmt"
-	"istio.io/istio/galley/pkg/config/analysis"
-	"istio.io/istio/galley/pkg/config/analysis/analyzers/webhook"
-	"istio.io/istio/galley/pkg/config/analysis/diag"
-	"istio.io/istio/galley/pkg/config/analysis/local"
-	cfgKube "istio.io/istio/galley/pkg/config/source/kube"
-	"istio.io/istio/istioctl/pkg/util/formatting"
-	"istio.io/istio/operator/pkg/apis/istio/v1alpha1"
-	"istio.io/istio/pkg/config/resource"
-	istioConfigSchema "istio.io/istio/pkg/config/schema"
 	"strings"
 	"time"
 
@@ -101,23 +92,8 @@ func (h *HelmReconciler) ApplyManifest(manifest name.Manifest, serverSideApply b
 	// Objects are applied in groups: namespaces, CRDs, everything else, with wait for ready in between.
 	nsObjs := object.KindObjects(changedObjects, name.NamespaceStr)
 	crdObjs := object.KindObjects(changedObjects, name.CRDStr)
-	mutatingWebhookObjs := object.KindObjects(changedObjects, name.MutatingWebhookConfigurationStr)
-	otherObjs := object.ObjectsNotInLists(changedObjects, nsObjs, crdObjs, mutatingWebhookObjs)
-	for _, objList := range []object.K8sObjects{nsObjs, crdObjs, otherObjs, mutatingWebhookObjs} {
-		// Analyze mutating webhook objects
-		if len(objList) > 0 {
-			if objList[0].Kind == name.MutatingWebhookConfigurationStr {
-				err := h.analyzeWebhooks(objList)
-				if err != nil {
-					if !h.opts.Force {
-						errs = util.AppendErr(errs, err)
-						continue
-					}
-					scope.Error("invalid webhook configs; continuing because of --force")
-				}
-			}
-		}
-
+	otherObjs := object.ObjectsNotInLists(changedObjects, nsObjs, crdObjs)
+	for _, objList := range []object.K8sObjects{nsObjs, crdObjs, otherObjs} {
 		// For a given group of objects, apply in sorted order of priority with no wait in between.
 		objList.Sort(object.DefaultObjectOrder())
 		for _, obj := range objList {
@@ -251,63 +227,6 @@ func (h *HelmReconciler) serverSideApply(obj *unstructured.Unstructured) error {
 	opts := []client.PatchOption{client.ForceOwnership, client.FieldOwner(fieldOwnerOperator)}
 	if err := h.client.Patch(context.TODO(), obj, client.Apply, opts...); err != nil {
 		return fmt.Errorf("failed to update resource with server-side apply for obj %v: %v", objectStr, err)
-	}
-	return nil
-}
-
-func (h *HelmReconciler) analyzeWebhooks(whs object.K8sObjects) error {
-	// Get YAML string from K8sObject
-	var whYAMLs []string
-	whNameMap := make(map[string]bool)
-	for _, whObj := range whs {
-		yaml, err := whObj.YAML()
-		if err != nil {
-			return err
-		}
-		whYAMLs = append(whYAMLs, string(yaml))
-		whNameMap[whObj.Name] = true
-	}
-	if len(whYAMLs) == 0 {
-		return nil
-	}
-
-	sa := local.NewSourceAnalyzer(istioConfigSchema.MustGet(), analysis.Combine("webhook", &webhook.Analyzer{}),
-		resource.Namespace(h.iop.Spec.GetNamespace()), resource.Namespace(v1alpha1.Namespace(h.iop.Spec)), nil, true, 30*time.Second)
-	var localWebhookYAMLReaders []local.ReaderSource
-	for _, wh := range whYAMLs {
-		whReaderSource := local.ReaderSource{
-			Name:   "",
-			Reader: strings.NewReader(wh),
-		}
-		localWebhookYAMLReaders = append(localWebhookYAMLReaders, whReaderSource)
-	}
-	err := sa.AddReaderKubeSource(localWebhookYAMLReaders)
-	if err != nil {
-		return err
-	}
-
-	k := cfgKube.NewInterfaces(h.restConfig)
-	sa.AddRunningKubeSource(k)
-
-	// Analyze webhooks
-	res, err := sa.Analyze(make(chan struct{}))
-	if err != nil {
-		fmt.Println(err)
-		return err
-	}
-	relevantMessages := diag.Messages{}
-	for _, msg := range res.Messages.FilterOutLowerThan(diag.Error) {
-		if _, ok := whNameMap[msg.Resource.Metadata.FullName.Name.String()]; ok {
-			relevantMessages = append(relevantMessages, msg)
-		}
-	}
-	if len(relevantMessages) > 0 {
-		o, err := formatting.Print(relevantMessages, formatting.LogFormat, false)
-		if err != nil {
-			return err
-		}
-		// nolint
-		return fmt.Errorf("creating default tag would conflict:\n%v", o)
 	}
 	return nil
 }
--- a/operator/pkg/helmreconciler/apply.go
+++ b/operator/pkg/helmreconciler/apply.go
@@ -23,6 +23,7 @@ import (
 	"istio.io/istio/galley/pkg/config/analysis/local"
 	cfgKube "istio.io/istio/galley/pkg/config/source/kube"
 	"istio.io/istio/istioctl/pkg/util/formatting"
+	"istio.io/istio/operator/pkg/apis/istio/v1alpha1"
 	"istio.io/istio/pkg/config/resource"
 	istioConfigSchema "istio.io/istio/pkg/config/schema"
 	"strings"
@@ -270,20 +271,8 @@ func (h *HelmReconciler) analyzeWebhooks(whs object.K8sObjects) error {
 		return nil
 	}
 
-	globalI, ok := h.iop.Spec.Values["global"]
-	if !ok {
-		return fmt.Errorf("failed to get global values")
-	}
-	global, ok := globalI.(map[string]interface{})
-	if !ok {
-		return fmt.Errorf("failed to get global values")
-	}
-	istioNamespace, ok := global["istioNamespace"].(string)
-	if !ok {
-		return fmt.Errorf("failed to get istioNamespace")
-	}
 	sa := local.NewSourceAnalyzer(istioConfigSchema.MustGet(), analysis.Combine("webhook", &webhook.Analyzer{}),
-		resource.Namespace(h.iop.Spec.GetNamespace()), resource.Namespace(istioNamespace), nil, true, 30*time.Second)
+		resource.Namespace(h.iop.Spec.GetNamespace()), resource.Namespace(v1alpha1.Namespace(h.iop.Spec)), nil, true, 30*time.Second)
 	var localWebhookYAMLReaders []local.ReaderSource
 	for _, wh := range whYAMLs {
 		whReaderSource := local.ReaderSource{
--- a/operator/pkg/helmreconciler/apply.go
+++ b/operator/pkg/helmreconciler/apply.go
@@ -17,6 +17,14 @@ package helmreconciler
 import (
 	"context"
 	"fmt"
+	"istio.io/istio/galley/pkg/config/analysis"
+	"istio.io/istio/galley/pkg/config/analysis/analyzers/webhook"
+	"istio.io/istio/galley/pkg/config/analysis/diag"
+	"istio.io/istio/galley/pkg/config/analysis/local"
+	cfgKube "istio.io/istio/galley/pkg/config/source/kube"
+	"istio.io/istio/istioctl/pkg/util/formatting"
+	"istio.io/istio/pkg/config/resource"
+	istioConfigSchema "istio.io/istio/pkg/config/schema"
 	"strings"
 	"time"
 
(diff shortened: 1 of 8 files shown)


akshitgrover:
@howardjohn 
Can you take a look at the implementation of the method to get MutatingWebhookConfiguration?
I'll call the webhook analyser after [FilterRenderedComponentManifest](https://github.com/istio/istio/pull/33644/files#diff-525ae81754807c29a4ae13370ad8e74c7948b57128c60967939feeeea4c8184fR159) is called in the [Reconcile](https://github.com/istio/istio/pull/33644/files#diff-525ae81754807c29a4ae13370ad8e74c7948b57128c60967939feeeea4c8184fR149) method of HelmReconciler.

akshitgrover:
Sorry! I might have bombarded your inbox with emails, I was actually trying the commands by prow bot as suggested by the istio-testing bot as well. It didn't work so deleted those comments

akshitgrover:
@howardjohn 
Apparently, Previous approach of getting the webhookconfig from the manifestMap was a redundant effort as the same thing was being done in the [ApplyManifest](https://github.com/istio/istio/pull/33644/files#diff-0b48478b9467d599862392852a780c736b172f60790c6203d4762e5da862804dR50) method on per k8sObject basis. So, The cleaner way was to use the same thing instead of adding a new method for the same.
Now, The only thing left is to use the same for analysis, For which I have added the implementation and injected the same in the [ApplyManifest](https://github.com/istio/istio/pull/33644/files#diff-0b48478b9467d599862392852a780c736b172f60790c6203d4762e5da862804dR50) method.

The only which is left is to skip this analysis if --skip-confirmation is set to true, Not sure if we need this. If yes, It'd be trickier to implement as the args are not propagated to the reconciler method.

PTAL at the implementation

howardjohn:
> The only which is left is to skip this analysis if --skip-confirmation is set to true, Not sure if we need this. If yes, It'd be trickier to implement as the args are not propagated to the reconciler method.

I think we still want to warn, possibly even failing unless `--force` is set?

operator/pkg/helmreconciler/apply.go
```diff
@@ -230,3 +251,72 @@ func (h *HelmReconciler) serverSideApply(obj *unstructured.Unstructured) error {
...
+		whYAMLs = append(whYAMLs, string(yaml))
+		whNameMap[whObj.Name] = true
+	}
+	if len(whYAMLs) == 0 {
+		return nil
+	}
+
+	globalI, ok := h.iop.Spec.Values["global"]
+	if !ok {
+		return fmt.Errorf("")
```
howardjohn, line 273:
```suggestion
		return fmt.Errorf("failed to get global value")
```

Should have some useful error aside from `""`. There are a few other locations as well

operator/pkg/helmreconciler/apply.go
```diff
@@ -230,3 +251,72 @@ func (h *HelmReconciler) serverSideApply(obj *unstructured.Unstructured) error {
...
+	}
+	istioNamespace, ok := global["istioNamespace"].(string)
+	if !ok {
+		return fmt.Errorf("")
+	}
+	sa := local.NewSourceAnalyzer(istioConfigSchema.MustGet(), analysis.Combine("webhook", &webhook.Analyzer{}),
+		resource.Namespace(h.iop.Spec.GetNamespace()), resource.Namespace(istioNamespace), nil, true, 30*time.Second)
+	var localWebhookYAMLReaders []local.ReaderSource
```
howardjohn, line 285: We have a lot of code about extracting the webhook yamls. Would it be simpler to just pass all the YAMLs to the analyzer code? I believe it will just ignore resources it doesn't care about (non-webhooks)

operator/pkg/helmreconciler/apply.go
```diff
@@ -92,8 +100,21 @@ func (h *HelmReconciler) ApplyManifest(manifest name.Manifest, serverSideApply b
...
 	crdObjs := object.KindObjects(changedObjects, name.CRDStr)
-	otherObjs := object.ObjectsNotInLists(changedObjects, nsObjs, crdObjs)
-	for _, objList := range []object.K8sObjects{nsObjs, crdObjs, otherObjs} {
+	mutatingWebhookObjs := object.KindObjects(changedObjects, name.MutatingWebhookConfigurationStr)
+	otherObjs := object.ObjectsNotInLists(changedObjects, nsObjs, crdObjs, mutatingWebhookObjs)
+	for _, objList := range []object.K8sObjects{nsObjs, crdObjs, otherObjs, mutatingWebhookObjs} {
+		// Analyze mutating webhook objects
+		if len(objList) > 0 {
+			if objList[0].Kind == name.MutatingWebhookConfigurationStr {
+				err := h.analyzeWebhooks(objList)
```
howardjohn, line 109: One issue with this is the order:
```
$ /istioctl/cmd/istioctl install -y
✔ Istio core installed
- Processing resources for Istiod.                                                                                                                                                    2021-06-28T17:27:09.163949Z     error   installer       creating default tag would conflict:
Error [IST0139] (MutatingWebhookConfiguration istio-sidecar-injector ) Webhook overlaps with others: [istio-revision-tag-default/namespace.sidecar-injector.istio.io]. This may cause injection to occur twice.
Error [IST0139] (MutatingWebhookConfiguration istio-sidecar-injector ) Webhook overlaps with others: [istio-revision-tag-default/object.sidecar-injector.istio.io]. This may cause injection to occur twice.
Error [IST0139] (MutatingWebhookConfiguration istio-sidecar-injector ) Webhook overlaps with others: [istio-revision-tag-default/rev.namespace.sidecar-injector.istio.io]. This may cause injection to occur twice.
Error [IST0139] (MutatingWebhookConfiguration istio-sidecar-injector ) Webhook overlaps with others: [istio-revision-tag-default/rev.object.sidecar-injector.istio.io]. This may cause injection to occur twice.
✘ Istiod encountered an error: creating default tag would conflict:
Error [IST0139] (MutatingWebhookConfiguration istio-sidecar-injector ) Webhook overlaps with others: [istio-revision-tag-default/namespace.sidecar-injector.istio.io]. This may cause injection to occur twice.
Error [IST0139] (MutatingWebhookConfiguration istio-sidecar-injector ) Webhook overlaps with others: [istio-revision-tag-default/object.sidecar-injector.istio.io]. This may cause injection to occur twice.
Error [IST0139] (MutatingWebhookConfiguration istio-sidecar-injector ) Webhook overlaps with others: [istio-revision-tag-default/rev.namespace.sidecar-injector.istio.io]. This may cause injection to occur twice.
Error [IST0139] (MutatingWebhookConfiguration istio-sidecar-injector ) Webhook overlaps with others: [istio-revision-tag-default/rev.object.sidecar-injector.istio.io]. This may cause injection to occur twice.
✔ Ingress gateways installed
- Pruning removed resources                                                                                                                                                           Error: failed to install manifests: errors occurred during operation
```

So the install still happens. Ideally we do it upfront, and fail without writing anything to the cluster

howardjohn (commented):
Looks like a great start!

operator/pkg/helmreconciler/apply.go
```diff
@@ -230,3 +251,72 @@ func (h *HelmReconciler) serverSideApply(obj *unstructured.Unstructured) error {
...
+
+	globalI, ok := h.iop.Spec.Values["global"]
+	if !ok {
+		return fmt.Errorf("")
+	}
+	global, ok := globalI.(map[string]interface{})
+	if !ok {
+		return fmt.Errorf("")
+	}
+	istioNamespace, ok := global["istioNamespace"].(string)
```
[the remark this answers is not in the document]
akshitgrover: I was actually looking for the same thing.
I couldn't find, I'll keep looking by the time can you suggest who can I reach out for the same?

operator/pkg/helmreconciler/apply.go
akshitgrover: Yes definitely, Since it's a draft PR I didn't add anything yet

operator/pkg/helmreconciler/apply.go
akshitgrover: This actually make a lot of sense
I had the same concern that a lot of duplication of this logic is there in places (content.go and tag.go) 
Will work on this as well

operator/pkg/helmreconciler/apply.go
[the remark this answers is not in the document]
howardjohn: @morvencao or @richardwxn may be more familiar

operator/pkg/helmreconciler/apply.go
akshitgrover: I actually wanted some insights on this, Initially I had implemented what you have mentioned.
The issue with that is, Analyzer also looks for a valid istiod service running in the cluster. If validate webhooks upfront it will always fail and happy installation path never goes through.

akshitgrover:
> > The only which is left is to skip this analysis if --skip-confirmation is set to true, Not sure if we need this. If yes, It'd be trickier to implement as the args are not propagated to the reconciler method.
> 
> I think we still want to warn, possibly even failing unless `--force` is set?

Can you guide me a bit on this
I am not sure I understand the flow here completely

operator/pkg/helmreconciler/apply.go
akshitgrover: Since, The installation path is just like kubectl apply it doesn't add unnecessary churn
If user fixes the issue with the webhook and run the install command, Only the webhook object will be created and all else remains the same.
So, This was like a clean and stable solution I could think of like just skipping the webhook creation on error and letting everything else go through.

akshitgrover: Since, The installation path is just like kubectl apply it doesn't add unnecessary churn
If user fixes the issue with the webhook and run the install command, Only the webhook object will be created and all else remains the same.
So, This was like a clean and stable solution I could think of like just skipping the webhook creation on error and letting everything else go through.

operator/pkg/helmreconciler/apply.go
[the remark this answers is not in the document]
akshitgrover: Is there any better way to get values from global map of IstioOperatorSpec?
cc @morvencao @richardwxn

operator/pkg/helmreconciler/apply.go
akshitgrover: @howardjohn 
Should we think on a solution for this?
Or partial install works?

operator/pkg/helmreconciler/apply.go
akshitgrover: I looked at the repetitive implementations, It doesn't look like extracting webhooks is taking is the clutter, Just that procedure to get metadata like namespaces and converting into YAML strings is the issue. This is the result of having different ways to fetch data at different places in the code.
Ideally, We should wrap this up in a single function in utils somewhere and that might help us do away with some of this
But, It's not extracting webhooks.

operator/pkg/helmreconciler/apply.go
howardjohn: I would prefer to fail (or prompt?) rather than partial install. Generally it seems favorable to have atomic operations, so everything succeeds or nothing does.

cc @Monkeyanator if you have opinions here

Monkeyanator: I agree that ideally we should fail rather than continue with partial install. Istio is loud on install and error messages can be easy to ignore IMO

akshitgrover: I have updated the PR
Now, we remove anything installed in case there are some errors and --force is not set

operator/pkg/helmreconciler/apply.go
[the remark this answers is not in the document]
akshitgrover: This how it's done in most places
https://github.com/istio/istio/blob/master/operator/pkg/helmreconciler/reconciler.go#L555

operator/pkg/helmreconciler/apply.go
akshitgrover: Added error messages

operator/pkg/helmreconciler/apply.go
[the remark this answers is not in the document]
morvencao: @akshitgrover If you're looking for how to get the namespace of istio control plane, we do have a method for that, see: https://github.com/istio/istio/blob/70e42c66694e1d264c304d94f3ca0ed535d80877/operator/pkg/apis/istio/v1alpha1/common.go#L30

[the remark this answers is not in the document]
akshitgrover: Updated the PR with this
Thank you

operator/pkg/helmreconciler/apply.go
howardjohn: I would much prefer we do an upfront check. Adding things then removing them may lead to unexpected behavior (example: I already have istio install, and I rerun `istioctl install`. Now it will remove my original installation).

akshitgrover: @howardjohn 
Issue with upfront validation is that it will always fail because webhook analyzer requires Istiod/Pilot component corresponding to the webhook being analysed installed.
So the upfront validation will always fail.

To acknowledge your concern, If you rerun istioctl install will not remove original installation as there will be no changedObjects to be deployed
But in case istioctl install --revision=somevalue is given and it results in some conflicts in this case we might run into some errors.
I think what we should do is instead of removing allObjects we should just remove the changedObjects that should be fine, WDYT?

akshitgrover: Or we can skip the part in the webhook analyser where it checks for the presence of Istiod service for the `istioctl install` workflow

howardjohn: if we pass all the yaml we are about to apply (including the service) would that work?

if not, I would prefer to skip that step.

Thanks for iterating on this?

akshitgrover: @howardjohn 
I have updated the workflow to do the analysis upfront before creating anything on the cluster
If force is not set installation will exit with an error without creating/deleting any object from the cluster else normal flow will continue if force is set

howardjohn (approved):
Looks great!

akshitgrover:
Tests are failing because of https://github.com/istio/istio/pull/33644/files#diff-525ae81754807c29a4ae13370ad8e74c7948b57128c60967939feeeea4c8184fR581 statement

We'll have to filter and just the valid resources, Will raise a PR with the fix

akshitgrover:
@howardjohn 
I'll raise a PR with the fix to pass test runs
What about the release-notes tests are there going to be any release notes for this if not can we add release-notes-none label so that particular test can pass?

shamsher31 (commented):
Please add release notes. Refer [this](https://github.com/istio/istio/blob/master/releasenotes/notes/18487.yaml) or any [other notes](https://github.com/istio/istio/tree/master/releasenotes/notes) for reference.

akshitgrover:
@shamsher31 
I've added the release notes

Thank you

operator/pkg/helmreconciler/reconciler.go
```diff
@@ -540,6 +555,59 @@ func CreateNamespace(cs kubernetes.Interface, namespace string, network string)
...
+		return err
+	}
+
+	k := cfgKube.NewInterfaces(h.restConfig)
+	sa.AddRunningKubeSource(k)
+
+	// Analyze webhooks
+	res, err := sa.Analyze(make(chan struct{}))
+	if err != nil {
+		fmt.Println(err)
```
shamsher31, line 596: Please remove this line as the error is already returned to the caller.

operator/pkg/helmreconciler/reconciler.go
```diff
@@ -540,6 +555,59 @@ func CreateNamespace(cs kubernetes.Interface, namespace string, network string)
...
+		return err
+	}
+	relevantMessages := res.Messages.FilterOutBasedOnResources(parsedK8sObjects)
+	if len(relevantMessages) > 0 {
+		o, err := formatting.Print(relevantMessages, formatting.LogFormat, false)
+		if err != nil {
+			return err
+		}
+		// nolint
+		return fmt.Errorf("creating default tag would conflict:\n%v", o)
```
shamsher31, line 606: Maybe you can just use `h.opts.Log` here. Why do you need `// nolint`?

releasenotes/notes/33537.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: release-notes/v2
+kind: feature
+area: istioctl
+issue:
+  - 33537
+
+releaseNotes:
+  - |
+    **Added** `istioctl install` will now do IST0139 analysis on webhooks.
```
shamsher31, line 9:
```suggestion
    **Added** `istioctl install` will now do `IST0139` analysis on webhooks.
```

shamsher31 (commented):
Overall LGTM. just few minor comments.

operator/pkg/helmreconciler/reconciler.go
akshitgrover: This makes sense, I forgot to remove it
This was there for initial implementation

operator/pkg/helmreconciler/reconciler.go
akshitgrover: Remove nolint directive
We need to return an error, h.opts.Log is for logging I guess it doesn't return the error

releasenotes/notes/33537.yaml
akshitgrover: Done 👍🏽

shamsher31 (approved):

PR #33650, opened 2021-06-26
ext-authz: add e2e tests for related header options in authz extension provider

yangminzhu:
This PR extends the sample ext-authz server and add new e2e tests to cover the new `headersToUpstreamOnAllow`, `headersToDownstreamOnDeny` and `includeAdditionalHeadersInCheck` options in the extension provider.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/integration/security/main_test.go
+++ b/tests/integration/security/main_test.go
@@ -75,7 +75,12 @@ meshConfig:
       service: %q
       port: 8000
       pathPrefix: "/check"
-      includeHeadersInCheck: ["x-ext-authz"]
+      headersToUpstreamOnAllow: ["x-ext-authz-*"]
+      headersToDownstreamOnDeny: ["x-ext-authz-*"]
+      includeRequestHeadersInCheck: ["x-ext-authz"]
+      includeAdditionalHeadersInCheck:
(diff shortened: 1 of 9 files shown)


tests/integration/security/main_test.go
```diff
@@ -75,7 +75,12 @@ meshConfig:
       service: %q
       port: 8000
       pathPrefix: "/check"
-      includeHeadersInCheck: ["x-ext-authz"]
+      headersToUpstreamOnAllow: ["x-ext-authz-*"]
```
howardjohn, line 78: nit: we may want to start using istio.PatchMeshConfig now. This was a bad idea before since mesh config changes took ~2 mins to propagate, now they should be instance

howardjohn (approved):

ericvn (approved):

tests/integration/security/main_test.go
yangminzhu: thanks, will update later.

PR #33653, opened 2021-06-26
fuzzing: Add mesh fuzzer

AdamKorcz:
This PR adds a fuzzer for the mesh parsing routines.

Once this fuzzer is merged, I will make the necessary changes on the OSS-fuzz side to run this fuzzer continuously.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/fuzz/mesh_fuzzer.go
+++ b/tests/fuzz/mesh_fuzzer.go
@@ -1,0 +1,30 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// nolint: golint
+package fuzz
+
+import (
+	"istio.io/istio/pkg/config/mesh"
+)
+
+func FuzzParseMeshNetworks(data []byte) int {
+	_, _ = mesh.ParseMeshNetworks(string(data))
+	return 1
+}
+
+func FuzzValidateMeshConfig(data []byte) int {
+	_, _ = mesh.ApplyMeshConfigDefaults(string(data))
+	return 1
+}


howardjohn:
> Once this fuzzer is merged, I will make the necessary changes on the OSS-fuzz side to run this fuzzer continuously.

Is it possible to have all of the logic in istio/istio repo? Then oss-fuzz just calls ./run-fuzz.sh or whatever and never changes?

This seems more sustainable and easier to test

AdamKorcz:
> Is it possible to have all of the logic in istio/istio repo? Then oss-fuzz just calls ./run-fuzz.sh or whatever and never changes?

Absolutely. It is tried and tested by many other integrated projects.
I will set it up in a separate PR once the two fuzzers are integrated so we can include them as well in the same PR.

AdamKorcz:
I see the release note test is failing. Sorry for that, it is my bad, as I did not check that part in the PR message.

I have just updated the initial PR message.

tests/fuzz/mesh_fuzzer.go
```diff
@@ -0,0 +1,34 @@
...
+	mn, err := mesh.ParseMeshNetworks(string(data))
+	if err != nil {
+		return 0
+	}
+	mesh.ResolveHostsInNetworksConfig(mn)
+	return 1
+}
+
+func FuzzValidateMeshConfig(data []byte) int {
+	_, _ = mesh.ApplyMeshConfigDefaults(string(data))
```
howardjohn, line 32: should we return 0 on error like in FuzzParseMeshNetworks?

tests/fuzz/mesh_fuzzer.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"istio.io/istio/pkg/config/mesh"
+)
+
+func FuzzParseMeshNetworks(data []byte) int {
+	mn, err := mesh.ParseMeshNetworks(string(data))
+	if err != nil {
+		return 0
+	}
+	mesh.ResolveHostsInNetworksConfig(mn)
```
howardjohn, line 27: This is doing a real network DNS lookup. Is that legal/a good idea in a fuzzer?

howardjohn (commented):
lgtm overall

tests/fuzz/mesh_fuzzer.go
AdamKorcz: I suggest we don't. The return values don't make any noticeable difference in practice.

tests/fuzz/mesh_fuzzer.go
AdamKorcz: Thanks for the heads up. We don't want to do that. I have removed that part.

howardjohn (approved):

PR #33658, opened 2021-06-28
Add an integration test for CNI race condition repair.

bianpengyuan:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/integration/pilot/cni/cni_race_test.go
+++ b/tests/integration/pilot/cni/cni_race_test.go
@@ -1,0 +1,155 @@
+// +build integ
+//  Copyright Istio Authors
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package cni
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"path/filepath"
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	istioKube "istio.io/istio/pkg/kube"
+	"istio.io/istio/pkg/test/env"
+	"istio.io/istio/pkg/test/framework"
+	"istio.io/istio/pkg/test/framework/components/cluster"
+	"istio.io/istio/pkg/test/framework/components/istio"
+	"istio.io/istio/pkg/test/framework/components/istioctl"
+	"istio.io/istio/pkg/test/framework/label"
+	"istio.io/istio/pkg/test/framework/resource"
+	"istio.io/istio/pkg/test/scopes"
+	"istio.io/istio/pkg/test/shell"
+	"istio.io/istio/pkg/test/util/retry"
+	"istio.io/istio/tests/integration/security/util"
+	"istio.io/istio/tools/istio-iptables/pkg/constants"
+)
+
+var (
+	ist  istio.Instance
+	apps = &util.EchoDeployments{}
+)
+
+func TestMain(m *testing.M) {
+	s := framework.NewSuite(m).Label(label.CustomSetup).Setup(istio.Setup(&ist, nil))
+	if !ist.Settings().EnableCNI {
+		s.Skip("CNI race condition mitigation is only tested when CNI is enabled.")
+	}
+	s.Setup(func(t resource.Context) error {
+		return util.SetupApps(t, ist, apps, false)
+	}).Run()
+}
+
+func TestCNIRaceRepair(t *testing.T) {
+	framework.NewTest(t).
+		Run(func(t framework.TestContext) {
+			cluster := t.Clusters().Default()
+			// To begin with, delete CNI Daemonset to simulate a CNI race condition
+			if err := deleteCNIDaemonset(cluster); err != nil {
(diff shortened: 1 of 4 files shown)


tests/integration/pilot/cni/cni_race_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+func TestCNIRaceRepair(t *testing.T) {
+	framework.NewTest(t).
+		Features("traffic.cni.race-condition-repair").
+		Run(func(t framework.TestContext) {
+			if !ist.Settings().EnableCNI {
+				t.Skip("CNI race condition mitigation is only tested when CNI is enabled.")
+			}
+			cluster := t.Clusters().Default()
+			// To begin with, delete CNI Daemonset to simulate a CNI race condition.
```
howardjohn, line 66: Why is this enough? Once its installed once isn't that sufficient to get iptables? Or is it cleaning up on shutdown

tests/integration/pilot/cni/cni_race_test.go
```diff
@@ -0,0 +1,158 @@
...
+			deleteCNIDaemonset(t, cluster)
+
+			// Rollout restart instances in namespace 1, and wait for a broken instance.
+			rolloutCmd := fmt.Sprintf("kubectl rollout restart deployment -n %s", apps.Namespace1.Name())
```
howardjohn, line 70: There is a .Restart() on echo.Instance that may be preferred.

tests/integration/pilot/cni/cni_race_test.go
```diff
@@ -0,0 +1,158 @@
...
+		Features("traffic.cni.race-condition-repair").
+		Run(func(t framework.TestContext) {
+			if !ist.Settings().EnableCNI {
+				t.Skip("CNI race condition mitigation is only tested when CNI is enabled.")
+			}
+			cluster := t.Clusters().Default()
+			// To begin with, delete CNI Daemonset to simulate a CNI race condition.
+			deleteCNIDaemonset(t, cluster)
+
+			// Rollout restart instances in namespace 1, and wait for a broken instance.
```
howardjohn, line 69: Do we need all? Or just a single pod?

tests/integration/pilot/cni/cni_race_test.go
```diff
@@ -0,0 +1,158 @@
...
+			return errors.New("CNI Daemonset pod still exists after deletion")
+		}
+		return nil
+	}, retry.Delay(1*time.Second), retry.Timeout(80*time.Second))
+}
+
+func deployCNIDaemonset(ctx framework.TestContext, c cluster.Cluster) error {
+	args := []string{
+		"install", "-f", filepath.Join(env.IstioSrc,
+			"tests/integration/pilot/cni/testdata/cni.yaml"),
```
howardjohn, line 108: can we save the old yaml and re-apply it? There may be customizations done outside the test framework?

tests/integration/pilot/cni/cni_race_test.go
bianpengyuan: Yeah, CNI network plugin gets cleaned up on removal.

tests/integration/pilot/cni/cni_race_test.go
bianpengyuan: Tricky thing is that workload will never become ready until CNI gets reinstalled. However `Restart` waits for workload to be ready, and I don't want to add a SkipWaiting to `Restart` just for this test. Let me know if you think it is fine to add skipWaiting option.

tests/integration/pilot/cni/cni_race_test.go
bianpengyuan: I think just a single pod is enough, which indicates there are pods stuck in race condition. Later we verify that no pod is broken, which proves that repair has kicked in.

tests/integration/pilot/cni/cni_race_test.go
bianpengyuan: Done.

tests/integration/pilot/cni/cni_race_test.go
```diff
@@ -0,0 +1,174 @@
...
+	"istio.io/istio/tests/integration/security/util"
+	"istio.io/istio/tools/istio-iptables/pkg/constants"
+)
+
+var (
+	ist  istio.Instance
+	apps = &util.EchoDeployments{}
+)
+
+func TestMain(m *testing.M) {
```
howardjohn, line 46: why do we need a whole new test suite for this? It doesn't seem like its installing anything custom

tests/integration/pilot/cni/cni_race_test.go
```diff
@@ -0,0 +1,174 @@
...
+	framework.NewSuite(m).
+		Label(label.CustomSetup).
+		Setup(istio.Setup(&ist, nil)).
+		Setup(func(t resource.Context) error {
+			return util.SetupApps(t, ist, apps, false)
```
howardjohn, line 51: Why do we need so many? Don't we only need 1 to test this?

tests/integration/pilot/cni/testdata/cni.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: install.istio.io/v1alpha1
```
howardjohn, line 1: I think this is dead code

howardjohn (approved):
LGTM, left some comments that may speed things up since this tests adds 90s to our test runtime

tests/integration/pilot/cni/cni_race_test.go
bianpengyuan: Move into the common suite.

tests/integration/pilot/cni/testdata/cni.yaml
bianpengyuan: Removed. thanks!

PR #33662, opened 2021-06-28
support signer from proxyConfig in istio-agent side

irisdingbj:
istio-agent part to support 
https://docs.google.com/document/d/1rxTSLuepcE_OJHj4C23RMJcgyIZzlBI-GqO8Y37SPko/edit#

Testing will be added when istiod part is ready.

### DIFF
--- a/pkg/istio-agent/agent.go
+++ b/pkg/istio-agent/agent.go
@@ -88,6 +88,7 @@ const (
 	MetadataClientCertKey   = "ISTIO_META_TLS_CLIENT_KEY"
 	MetadataClientCertChain = "ISTIO_META_TLS_CLIENT_CERT_CHAIN"
 	MetadataClientRootCert  = "ISTIO_META_TLS_CLIENT_ROOT_CERT"
+	MetadataCertSigner      = "ISTIO_META_CERT_SIGNER"
 )
 
 // Agent contains the configuration of the agent, based on the injected
--- a/security/pkg/nodeagent/caclient/providers/citadel/client.go
+++ b/security/pkg/nodeagent/caclient/providers/citadel/client.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/gogo/protobuf/types"
 	"go.uber.org/atomic"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
@@ -81,9 +82,17 @@ func (c *CitadelClient) Close() {
 
 // CSR Sign calls Citadel to sign a CSR.
 func (c *CitadelClient) CSRSign(csrPEM []byte, certValidTTLInSec int64) ([]string, error) {
+	crMetaStruct := &types.Struct{
+		Fields: map[string]*types.Value{
+			security.CertSigner: {
+				Kind: &types.Value_StringValue{StringValue: c.opts.CertSigner},
+			},
+		},
+	}
 	req := &pb.IstioCertificateRequest{
 		Csr:              string(csrPEM),
 		ValidityDuration: certValidTTLInSec,
+		Metadata:         crMetaStruct,
 	}
 	if err := c.reconnectIfNeeded(); err != nil {
 		return nil, err
--- a/security/pkg/pki/ra/k8s_ra.go
+++ b/security/pkg/pki/ra/k8s_ra.go
@@ -47,14 +47,23 @@ func NewKubernetesRA(raOpts *IstioRAOptions) (*KubernetesRA, error) {
 	return istioRA, nil
 }
 
-func (r *KubernetesRA) kubernetesSign(csrPEM []byte, caCertFile string) ([]byte, error) {
+func (r *KubernetesRA) kubernetesSign(csrPEM []byte, caCertFile string, certSigner string) ([]byte, error) {
+	certSignerDomain := r.raOpts.CertSignerDomain
+	if certSignerDomain != "" && certSigner != "" {
+		certSigner = certSignerDomain + "/" + certSigner
+	} else {
+		certSigner = r.raOpts.CaSigner
+	}
+	if certSignerDomain == "" && certSigner != "" {
+		return nil, raerror.NewError(raerror.CertGenError, fmt.Errorf("certSignerDomain is requiered for signer %s", certSigner))
+	}
 	usages := []cert.KeyUsage{
 		cert.UsageDigitalSignature,
 		cert.UsageKeyEncipherment,
 		cert.UsageServerAuth,
 		cert.UsageClientAuth,
 	}
-	certChain, _, err := chiron.SignCSRK8s(r.csrInterface, csrPEM, r.raOpts.CaSigner,
+	certChain, _, err := chiron.SignCSRK8s(r.csrInterface, csrPEM, certSigner,
 		nil, usages, "", caCertFile, true, false)
 	if err != nil {
 		return nil, raerror.NewError(raerror.CertGenError, err)
@@ -68,7 +77,8 @@ func (r *KubernetesRA) Sign(csrPEM []byte, certOpts ca.CertOpts) ([]byte, error)
 	if err != nil {
 		return nil, err
 	}
-	return r.kubernetesSign(csrPEM, r.raOpts.CaCertFile)
+	certSigner := certOpts.CertSigner
+	return r.kubernetesSign(csrPEM, r.raOpts.CaCertFile, certSigner)
 }
 
 // SignWithCertChain is similar to Sign but returns the leaf cert and the entire cert chain.
--- a/security/pkg/pki/ra/k8s_ra.go
+++ b/security/pkg/pki/ra/k8s_ra.go
@@ -49,13 +49,13 @@ func NewKubernetesRA(raOpts *IstioRAOptions) (*KubernetesRA, error) {
 
 func (r *KubernetesRA) kubernetesSign(csrPEM []byte, caCertFile string, certSigner string) ([]byte, error) {
 	certSignerDomain := r.raOpts.CertSignerDomain
+	if certSignerDomain == "" && certSigner != "" {
+		return nil, raerror.NewError(raerror.CertGenError, fmt.Errorf("certSignerDomain is requiered for signer %s", certSigner))
+	}
 	if certSignerDomain != "" && certSigner != "" {
 		certSigner = certSignerDomain + "/" + certSigner
 	} else {
 		certSigner = r.raOpts.CaSigner
-	}
-	if certSignerDomain == "" && certSigner != "" {
-		return nil, raerror.NewError(raerror.CertGenError, fmt.Errorf("certSignerDomain is requiered for signer %s", certSigner))
 	}
 	usages := []cert.KeyUsage{
 		cert.UsageDigitalSignature,
--- a/security/pkg/server/ca/server.go
+++ b/security/pkg/server/ca/server.go
@@ -78,12 +78,15 @@ func (s *Server) CreateCertificate(ctx context.Context, request *pb.IstioCertifi
 	}
 
 	// TODO: Call authorizer.
-
+	crMetadata := request.Metadata.GetFields()
+	certSigner := crMetadata[security.CertSigner].GetStringValue()
+	log.Infof("cert signer %s", certSigner)
 	_, _, certChainBytes, rootCertBytes := s.ca.GetCAKeyCertBundle().GetAll()
 	certOpts := ca.CertOpts{
 		SubjectIDs: caller.Identities,
 		TTL:        time.Duration(request.ValidityDuration) * time.Second,
 		ForCA:      false,
+		CertSigner: certSigner,
 	}
 	cert, signErr := s.ca.Sign([]byte(request.Csr), certOpts)
 	if signErr != nil {
--- a/security/pkg/server/ca/server.go
+++ b/security/pkg/server/ca/server.go
@@ -80,7 +80,7 @@ func (s *Server) CreateCertificate(ctx context.Context, request *pb.IstioCertifi
 	// TODO: Call authorizer.
 	crMetadata := request.Metadata.GetFields()
 	certSigner := crMetadata[security.CertSigner].GetStringValue()
-	log.Infof("cert signer %s", certSigner)
+	log.Debugf("cert signer from workload %s", certSigner)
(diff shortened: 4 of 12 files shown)


hzxuzhonghu:
what is the use case

pkg/istio-agent/agent.go
```diff
@@ -88,6 +88,9 @@ const (
 	MetadataClientCertKey   = "ISTIO_META_TLS_CLIENT_KEY"
 	MetadataClientCertChain = "ISTIO_META_TLS_CLIENT_CERT_CHAIN"
 	MetadataClientRootCert  = "ISTIO_META_TLS_CLIENT_ROOT_CERT"
+	CertSignerLabel         = "cert.istio.io/signer"
```
howardjohn, line 91: can you add comments? especially what the format is?

Also this should be in istio/api, under the labels/ package.

pkg/istio-agent/agent.go
```diff
@@ -88,6 +88,9 @@ const (
...
+	CertSignerNSLabel       = "namespace:"
```
howardjohn, line 92: what does the namespace do?

security/pkg/nodeagent/caclient/providers/citadel/client.go
```diff
@@ -85,7 +85,9 @@ func (c *CitadelClient) CSRSign(csrPEM []byte, certValidTTLInSec int64) ([]strin
 	if err := c.reconnectIfNeeded(); err != nil {
 		return nil, err
 	}
-	ctx := metadata.NewOutgoingContext(context.Background(), metadata.Pairs("ClusterID", c.opts.ClusterID))
+	metakv := []string{"ClusterID", c.opts.ClusterID}
```
howardjohn, line 88: Currently this PR has no tests. It would be good to have tests of this feature.

howardjohn (commented):
Can you add a link to the approved design doc in the description? Thanks!

pkg/istio-agent/agent.go
irisdingbj: Thanks for this pointer! Should we mark this as `EXPERIMENTAL` ?

security/pkg/nodeagent/caclient/providers/citadel/client.go
irisdingbj: yes, plan to submit the istio-agent part first. And then will the part for isitod and will add tests for it together.

pkg/istio-agent/agent.go
irisdingbj: This is the prefix:  If   signer is defined via pod label, eg: `cert.isito.io/signer=prod`, the generated signer info will be `pod:prod` , If   namespace name is used as  signer info, eg: `staging`,  the generated signer info will be `namespace:staging` , then we can see which one has higher priority when comparing with the one defined via env  `K8S_SIGNER `

irisdingbj:
@howardjohn moved cert-signer into proxyConfig (proxymetadata), PTAL, Thanks!

pkg/istio-agent/agent.go
```diff
@@ -658,7 +659,7 @@ func (a *Agent) newSecretManager() (*cache.SecretManagerClient, error) {
 			log.Infof("Using CA %s cert with certs: %s", a.secOpts.CAEndpoint, caCertFile)
 		}
 	}
-
+	a.secOpts.CertSigner = a.proxyConfig.ProxyMetadata[MetadataCertSigner]
```
howardjohn, line 662: ProxyMetadata becomes an environment variable. We should just read from the environment in pilot/cmd/pilot-agent/options/options.go

security/pkg/nodeagent/caclient/providers/citadel/client.go
```diff
@@ -85,7 +85,11 @@ func (c *CitadelClient) CSRSign(csrPEM []byte, certValidTTLInSec int64) ([]strin
 	if err := c.reconnectIfNeeded(); err != nil {
 		return nil, err
 	}
-	ctx := metadata.NewOutgoingContext(context.Background(), metadata.Pairs("ClusterID", c.opts.ClusterID))
+	metaKV := []string{"ClusterID", c.opts.ClusterID}
+	if c.opts.CertSigner != "" {
+		metaKV = append(metaKV, security.CertSigner, c.opts.CertSigner)
```
howardjohn, line 90: Should we set it in IstioCertificateRequest.Metadata? I am not sure we want to overload `metadata` so much (which is really HTTP headers) to encode information about the request. The current usage of cluster is a bit of a hack

security/pkg/pki/ra/k8s_ra.go
```diff
@@ -73,6 +73,12 @@ func (r *KubernetesRA) Sign(csrPEM []byte, certOpts ca.CertOpts) ([]byte, error)
 		return nil, err
 	}
 	csrName := chiron.GenCsrName()
+	signerFromRA := r.raOpts.CaSigner
+	certSigner := certOpts.CertSigner
+	if certSigner == "" {
+		certSigner = signerFromRA
+	}
+	r.raOpts.CertSigner = certSigner
```
howardjohn, line 81: This seems like its modifying the KubernetesRA options, which is intended to be static I believe. This means that one CSR will impact all future CSRs. I think that we need a local variable, not modifying our full state

security/pkg/pki/ra/k8s_ra.go
```diff
@@ -73,6 +73,12 @@ func (r *KubernetesRA) Sign(csrPEM []byte, certOpts ca.CertOpts) ([]byte, error)
 		return nil, err
 	}
 	csrName := chiron.GenCsrName()
+	signerFromRA := r.raOpts.CaSigner
+	certSigner := certOpts.CertSigner
```
howardjohn, line 77: Don't we need to check it is within CertSignerDomain?

security/pkg/nodeagent/caclient/providers/citadel/client.go
irisdingbj: Thanks for the suggestion! added certSigner into IstioCertificateRequest.Metadata now

security/pkg/pki/ra/k8s_ra.go
irisdingbj: Added

pkg/istio-agent/agent.go
irisdingbj: Thanks! fixed

security/pkg/pki/ra/k8s_ra.go
irisdingbj: Thanks! fixed

security/pkg/pki/ra/k8s_ra.go
irisdingbj: fixed, Thanks!

security/pkg/pki/ra/k8s_ra.go
irisdingbj: fixed, Thanks!

security/pkg/nodeagent/caclient/providers/citadel/client.go
irisdingbj: moved to  IstioCertificateRequest.Metadata

pkg/istio-agent/agent.go
irisdingbj: Thanks! fixed

security/pkg/nodeagent/caclient/providers/citadel/client.go
```diff
@@ -85,7 +85,9 @@ func (c *CitadelClient) CSRSign(csrPEM []byte, certValidTTLInSec int64) ([]strin
 	if err := c.reconnectIfNeeded(); err != nil {
 		return nil, err
 	}
-	ctx := metadata.NewOutgoingContext(context.Background(), metadata.Pairs("ClusterID", c.opts.ClusterID))
+	metakv := []string{"ClusterID", c.opts.ClusterID}
+	metakv = append(metakv, security.CertSigner, c.opts.CertSigner)
```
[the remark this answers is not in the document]
irisdingbj: moved to IstioCertificateRequest.Metadata and put certSigner here

security/pkg/nodeagent/caclient/providers/citadel/client.go
```diff
@@ -78,14 +82,27 @@ func (c *CitadelClient) Close() {
 
 // CSR Sign calls Citadel to sign a CSR.
 func (c *CitadelClient) CSRSign(csrPEM []byte, certValidTTLInSec int64) ([]string, error) {
+	crMetadata := map[string]interface{}{
+		security.CertSigner: c.opts.CertSigner,
+	}
+	crMetaStruct, err := mapToStruct(crMetadata)
```
howardjohn, line 88: very nit: you can just construct the Struct directly? to avoid some complexity and error handling

security/pkg/nodeagent/caclient/providers/citadel/client.go
howardjohn: It is still set here as well, can you clean this up?

howardjohn (commented):
Looks good overall! Could use a few tests though.

@shankgan can you give a final look?

security/pkg/nodeagent/caclient/providers/citadel/client.go
```diff
@@ -201,3 +218,19 @@ func (c *CitadelClient) reconnectIfNeeded() error {
 	citadelClientLog.Errorf("recreated connection")
 	return nil
 }
+
+func mapToStruct(metadata map[string]interface{}) (*types.Struct, error) {
```
shankgan, line 222: would it make sense to add a test for this function? Also, is this functionality not available in any of the current Istio packages?

security/pkg/server/ca/server.go
```diff
@@ -78,12 +78,14 @@ func (s *Server) CreateCertificate(ctx context.Context, request *pb.IstioCertifi
 	}
 
 	// TODO: Call authorizer.
-
+	crMetadata := request.Metadata.GetFields()
+	certSigner := crMetadata[security.CertSigner].GetStringValue()
```
shankgan, line 82: Is there a chance that this could be an empty string? Is that case handled?

shankgan:
Overall LGTM.

Any plans to add the e2e tests? Are there any changes pending on the Istiod side? Or are most of the changes in the ClusterRole and ClusterRoleBindings required for Istiod to create/sign CSR requests using the custom signer?

irisdingbj:
> Overall LGTM.
> 
> Any plans to add the e2e tests? Are there any changes pending on the Istiod side? Or are most of the changes in the ClusterRole and ClusterRoleBindings required for Istiod to create/sign CSR requests using the custom signer?

Thanks @shankgan @howardjohn for the review!

This PR is only for istio-agent part. We need change in istiod part as well besides the ClusterRole and ClusterRoleBindings like: make sure the cert-chain it got includes full chain and verify it using his own cert-chain. 

An integration test will be added as well after https://github.com/istio/istio/issues/33906 is ready.  @zhlsunshine is actively working on this now.

security/pkg/nodeagent/caclient/providers/citadel/client.go
irisdingbj: sure

security/pkg/server/ca/server.go
irisdingbj: Yes, we handled this case:  will use the value from `r.raOpts.CaSigner` if it is an empty string like below
```
	if certSigner == "" {
		certSigner = r.raOpts.CertSignerDomain + "/" + r.raOpts.CaSigner
	} else {
		certSigner = r.raOpts.CertSignerDomain + "/" + certSigner
	}
```

security/pkg/nodeagent/caclient/providers/citadel/client.go
irisdingbj: cleaned up

security/pkg/nodeagent/caclient/providers/citadel/client.go
irisdingbj: Thanks for the great suggestion!

security/pkg/nodeagent/caclient/providers/citadel/client.go
irisdingbj: remvoed this function now --consturct the struct directly according to the suggestion from John

irisdingbj:
@howardjohn @shankgan PTAL, Thanks!

security/pkg/pki/ra/k8s_ra.go
```diff
@@ -47,9 +47,18 @@ func NewKubernetesRA(raOpts *IstioRAOptions) (*KubernetesRA, error) {
...
+func (r *KubernetesRA) kubernetesSign(csrPEM []byte, csrName string, caCertFile string, certSigner string) ([]byte, error) {
+	certSignerDomain := r.raOpts.CertSignerDomain
+	if r.raOpts.CertSignerDomain != "" {
+		certSignerDomain += "/"
+	}
+	if certSigner == "" {
+		certSigner = certSignerDomain + r.raOpts.CaSigner
+	} else {
+		certSigner = certSignerDomain + certSigner
+	}
```
howardjohn, line 59: Hmm. I am thinking this may make more sense like:

```suggestion
certSignerDomain := r.raOpts.CertSignerDomain
if certSignerDomain != "" && certSigner != "" {
	certSigner = certSignerDomain+ "/" + certSigner
} else {
	certSigner = r.raOpts.CaSigner
}
```

Basically:
* use r.raOpts.CaSigner directly, not adding domain. This is owned by the control plane admin, so they can configure it however they want
* If CertSignerDomain is not set, do not allow CSR to overwrite the certSigner. This ensures that the control plane has opted into allowing CSR overwrite; otherwise this opens a security hole that anyone can request any signer, which is dangerous if Istiod happens to run with higher RBAC privileges than strictly required.

shankgan: +1 this..Control Plane needs to explicitly buy into the agent selecting its own signer. I would log an error if an agent  wants to use a signer and the control-plane doesn't have a CertSignerDomain set.

irisdingbj: fixed!  Also logged error when CertSignerDoamin is empty but certSigner is defined.

irisdingbj:
@howardjohn @shankgan PTAL, Thanks!

howardjohn (approved):
Thanks for all the iteration on this!

PR #33673, opened 2021-06-28
Fix a few data races

howardjohn:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/operator/pkg/util/progress/progress.go
+++ b/operator/pkg/util/progress/progress.go
@@ -63,7 +63,7 @@ func (p *Log) createStatus(maxWidth int) string {
 	wait := make([]string, 0, len(p.components))
 	for c, l := range p.components {
 		comps = append(comps, name.UserFacingComponentName(name.ComponentName(c)))
-		wait = append(wait, l.waiting...)
+		wait = append(wait, l.waitingResources()...)
 	}
 	sort.Strings(comps)
 	sort.Strings(wait)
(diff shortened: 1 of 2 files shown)


pkg/kube/portforwarder.go
```diff
@@ -115,7 +113,7 @@ func newPortForwarder(restConfig *rest.Config, podName, ns, localAddress string,
 		[]string{fmt.Sprintf("%d:%d", localPort, podPort)},
 		stopCh,
 		readyCh,
-		output,
+		io.Discard,
```
howardjohn, line 116: the `output` writer apparently needs to be threadsafe. We don't actually read the output anyways though, so just us io.Discard is easiest

operator/pkg/util/progress/progress.go
```diff
@@ -201,6 +204,8 @@ func (p *ManifestLog) ReportFinished() {
 	if p == nil {
 		return
 	}
+	p.mu.Lock()
+	defer p.mu.Unlock()
```
hzxuzhonghu, line 208: Seems this lock should not protect report, because the `report` function is already protected by this lock, not sure how the test passed, i am not very familiar with this code

howardjohn: i agree, let me fix it

hzxuzhonghu (approved):
lgtm

howardjohn:
@istio/wg-environments-maintainers ptal

nmittler (approved):

PR #33679, opened 2021-06-28
Fuzzing: Add pilot model fuzzer

AdamKorcz:
This PR adds a fuzzer that creates an `Environment` with a fuzzed `ServiceDiscovery`. That env is then passed to `pc.InitContext(env, nil, nil)`.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/fuzz/pilot_model_fuzzer.go
+++ b/tests/fuzz/pilot_model_fuzzer.go
@@ -1,0 +1,152 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// nolint: golint
+package fuzz
+
+import (
+	fuzz "github.com/AdaLogics/go-fuzz-headers"
+	"istio.io/istio/pilot/pkg/model"
+	"istio.io/istio/pkg/config/host"
+	"istio.io/istio/pkg/config/labels"
+	"istio.io/istio/pkg/config/mesh"
+)
+
+func NewSI(f *fuzz.ConsumeFuzzer) (*model.ServiceInstance, error) {
+	si := &model.ServiceInstance{}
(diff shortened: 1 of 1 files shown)


tests/fuzz/pilot_model_fuzzer.go
```diff
@@ -0,0 +1,152 @@
...
+	pc := model.NewPushContext()
+	_ = pc.InitContext(env, nil, nil)
+
+	return 1
+}
+
+// The following code is taken as-is
+// from istio/pilot/pkg/model/push_context_test.go:
+
+var _ model.ServiceDiscovery = &localServiceDiscovery{}
```
howardjohn, line 115: Can we use memory.NewServiceDiscovery() instead of implementing a new instance?

tests/fuzz/pilot_model_fuzzer.go
```diff
@@ -0,0 +1,152 @@
...
+	env.IstioConfigStore = model.MakeIstioStore(store)
+	env.ServiceDiscovery = &localServiceDiscovery{
+		services:         services,
+		serviceInstances: serviceInstances,
+	}
+
+	env.Watcher = mesh.NewFixedWatcher(m)
+	env.Init()
+	pc := model.NewPushContext()
+	_ = pc.InitContext(env, nil, nil)
```
howardjohn, line 107: This is great! one concern I have - how often do we actually get to this point in the code?

It may be a good idea to add a small README.md to tests/fuzz/ on how to run these locally, then I can answer that myself :slightly_smiling_face:

howardjohn (commented):
:+1: this is great

tests/fuzz/pilot_model_fuzzer.go
AdamKorcz: Yes, the fuzzer has been updated. Does it look better?

tests/fuzz/pilot_model_fuzzer.go
AdamKorcz: > how often do we actually get to this point in the code?

I improved the fuzzer a bit, and it hits `InitContext` often after only a few minutes of running. 

> It may be a good idea to add a small README.md to tests/fuzz/

I will be happy to do that later this week. I am thinking of adding a minimal Dockerfile as well.

Marking this comment as resolved, but feel free to reopen if need be.

howardjohn: Nice! I think I figured out how to get it running as well (was blocked by a bug in go-fuzz previously)

howardjohn (approved):

PR #33682, opened 2021-06-29
Fix incorrect indexing in xds cache

howardjohn:
When we generate a token, we do not add it to the indexes. This means
its not properly invalidated. This leads to the following sequence of
events:

* Generate token 1, while endpoint state is v1
* Endpoint update to v2. Invalidate cache // BUG - token 1 is not removed since its not in the index
* Write token 1 at v1
* Fetch token 1, while endpoint state is v2
* Write token 1 at v2 // This triggers the assertion, as the cache should not mutate

With the fix:

* Generate token 1, while endpoint state is v1
* Endpoint update to v2. Invalidate cache
* Write token 1 at v1 // SKIPPED - invalidated
* Fetch token 1, while endpoint state is v2
* Write token 1 at v2 // This triggers the assertion, as the cache should not mutate

### DIFF
--- a/pilot/pkg/model/xds_cache.go
+++ b/pilot/pkg/model/xds_cache.go
@@ -252,8 +252,8 @@ func (l *lruCache) Add(entry XdsCacheEntry, token CacheToken, value *discovery.R
 		l.assertUnchanged(k, cur.(cacheValue).value, value)
 	}
 	l.store.Add(k, toWrite)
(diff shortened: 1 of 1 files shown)


hzxuzhonghu:
It is a little hard to comprehend,  it looks https://github.com/istio/istio/pull/33611 can cover this case

nmittler (approved):

stevenctl (approved):

PR #33689, opened 2021-06-29
Defer should not be used in a loop

Haleygo:
`defer fw.Close()` shouldn't be used in a loop, in that case, fw.Close() will not happen at the end of each iteration, but only after the for statement (after all iterations)

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


Haleygo:
all the defer will be executed together after all iterations down.

PR #33703, opened 2021-06-29
Minor fix to xds tests to support subtests

nmittler:
Currently, most methods for XDS tests rely on a `testing.T` variable that is stored within a struct. This typically references the parent test and if it fails during a subtest, it will bring down all sibling tests with it.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


howardjohn (approved):
This is a bit annoying.. wish there was a better way to solve this, but I don't know of any. One thing we could do is add a .WithT() function to return a modified AdsTest, but it would probably be incorrectly used a lot

PR #33707, opened 2021-06-29
Adding debug/exportz for MCS

nmittler:
Also doing some other misc cleanup.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/serviceexportcache.go
+++ b/pilot/pkg/serviceregistry/kube/controller/serviceexportcache.go
@@ -17,6 +17,7 @@ package controller
 import (
 	"fmt"
 
+	klabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
 	mcsCore "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
@@ -31,6 +32,7 @@ import (
 type serviceExportCache interface {
 	isExported(name types.NamespacedName) bool
 	HasSynced() bool
+	ExportedServices() []string
 }
 
 // newServiceExportCache creates a new serviceExportCache that observes the given cluster.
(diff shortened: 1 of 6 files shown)


pilot/pkg/model/service.go
```diff
@@ -541,6 +541,16 @@ type ServiceDiscovery interface {
 
 	// NetworkManager returns a list of network gateways that can be used to access endpoints residing in this registry..
 	NetworkGateways() []*NetworkGateway
+
+	// ExportedServices returns a list of services exported by cluster.
+	ExportedServices() []ExportedService
```
howardjohn, line 546: It seems a bit heavy to add an interface to ServiceDiscovery, which is (and will be forever?) only implemented by a single controller and only for debug interface. Maybe we can just directly access the registry if it exists or something?

pilot/pkg/serviceregistry/kube/controller/serviceexportcache.go
```diff
@@ -121,3 +143,14 @@ func (c disabledServiceExportCache) isExported(types.NamespacedName) bool {
 func (c disabledServiceExportCache) HasSynced() bool {
 	return true
 }
+
+func (c disabledServiceExportCache) ExportedServices() []model.ExportedService {
+	// MCS is disabled - show everything as exported.
+	return []model.ExportedService{
+		{
+			Cluster:   "*",
```
howardjohn, line 151: We should document the `*` semantic somewhere?

Or maybe just return `nil` here, which would be different than `[]` which means no service disabled. That may be a bit subtle though..

nmittler: Yeah I like `nil` better ... done.

pilot/pkg/model/service.go
nmittler: I don't disagree that it's a little heavy handed. However, it is just following the pattern established with gateways. I suppose it's possible that other types of clusters could provide some sort of MCS-like mechanism.

In general, I think I prefer this to having some sort of break-glass mechanism that directly accesses kube controllers. IDK ... WDYT?

@stevenctl any opinion?

stevenctl: I sort-of lean towards John's side here. I think if it's for debug purposes we can just do some casts from `ServiceDiscovery -> Aggregate` then call `GetRegistries` and cast to k8s controllers etc etc. 

If we were using this for actual generation logic I would probably lean the other way.

stevenctl: Or even implementing it in aggregate cluster and just leaving  it off the interface

nmittler:
@howardjohn @stevenctl PTAL

stevenctl (approved):

PR #33709, opened 2021-06-29
Fuzzing: add endpoint slice fuzzer

AdamKorcz:
This PR adds a fuzzer that creates and deletes endpointslices randomly.
The test-helper `NewFakeDiscoveryServer` leaks memory, and the fuzzer runs out after 20-30 seconds. 

If it is not interesting to fix the memory leak and improve the `NewFakeDiscoveryServer` to run continuously, then I suggest this fuzzer is not merged. 

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/fuzz/service_registry_fuzzer.go
+++ b/tests/fuzz/service_registry_fuzzer.go
@@ -1,0 +1,99 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
I am not sure how effective this would be. Just adding an endpointslice and exiting probably will not actually do anything at all. There is a long propogation cycle of the event which will almost certainly take longer than it takes to exit the function. That same cycle also means that even if we did somehow wait for the event to be processed, the fuzzer may be very slow.

perhaps it would be more effective to directly call https://github.com/istio/istio/blob/77bc53bcd4d5a60f9d82c45d641a1a5cff7b25a5/pilot/pkg/serviceregistry/kube/controller/controller.go#L678 or similar?

AdamKorcz:
Based on https://github.com/istio/istio/pull/33709#pullrequestreview-701139857 I suggest we close this one and find a better way to fuzz the kube controller. If anyone disagrees, feel free to reopen.

PR #33711, opened 2021-06-29
wip: experiment re-using cluster generator for gRPC

stevenctl:


pilot/pkg/model/context.go
```diff
@@ -339,6 +339,25 @@ type WatchedResource struct {
...
+	filter map[string]struct{}
+}
+
+// IncludesResource allows filtering based on the ResourceNames of a WatchedResource.
+func (w *WatchedResource) IncludesResource(name string) bool {
+	if len(w.ResourceNames) == 0 {
+		return true
+	}
+	if w.filter == nil {
+		// generate the filter only once, assume ResourceNames is always set first and not-changed
```
howardjohn, line 353: This seems like an assumption that will not be true in many cases. Is this a grpc-only thing that it doesn't change?

pilot/pkg/networking/core/configgen.go
```diff
@@ -35,7 +35,7 @@ type ConfigGenerator interface {
 	BuildListeners(node *model.Proxy, push *model.PushContext) []*listener.Listener
 
 	// BuildClusters returns the list of clusters for the given proxy. This is the CDS output
-	BuildClusters(node *model.Proxy, push *model.PushContext) []*cluster.Cluster
+	BuildClusters(node *model.Proxy, push *model.PushContext, w *model.WatchedResource) []*cluster.Cluster
```
howardjohn, line 38: Passing some set of configs to generate makes sense, but `model.WatchedResource` may not be the best thing for delta XDS in the future. Maybe just a set of resources?

We can always change it for delta when needed though

pilot/pkg/networking/grpcgen/grpcgen.go
```diff
@@ -54,15 +54,37 @@ import (
...
+	return subsetClusterKey("", hostname, port)
+}
+
+func subsetClusterKey(subset, hostname string, port int) string {
+	return model.BuildSubsetKey(model.TrafficDirectionOutbound, subset, host.Name(hostname), port)
+}
 
 func (g *GrpcConfigGenerator) Generate(proxy *model.Proxy, push *model.PushContext,
 	w *model.WatchedResource, updates *model.PushRequest) (model.Resources, model.XdsLogDetails, error) {
+	log.Infof("begininning gRPC  %s, gen for %s, names %v", shortnames[w.TypeUrl], proxy.ID, w.ResourceNames)
```
howardjohn, line 77:
```suggestion
	log.Debugf("beginning gRPC %s, gen for %s, names %v", v3.GetShortType(w.TypeUrl), proxy.ID, w.ResourceNames)
```

pilot/pkg/networking/grpcgen/grpcgen.go
```diff
@@ -54,15 +54,37 @@ import (
...
+
 	switch w.TypeUrl {
 	case v3.ListenerType:
 		return g.BuildListeners(proxy, push, w.ResourceNames), model.DefaultXdsLogDetails, nil
 	case v3.ClusterType:
-		return g.BuildClusters(proxy, push, w.ResourceNames), model.DefaultXdsLogDetails, nil
+		if g.Cds != nil {
```
howardjohn, line 83: do we intend to have both paths long term?

pilot/pkg/xds/v3/model.go
```diff
@@ -36,7 +36,7 @@ const (
 	ProxyConfigType = apiTypePrefix + "istio.mesh.v1alpha1.ProxyConfig"
 	// DebugType requests debug info from istio, a secured implementation for istio debug interface.
 	DebugType     = "istio.io/debug"
-	BootstrapType = apiTypePrefix + "envoy.config.bootstrap.v3.Bootstrap"
+	BootstrapType = apiTypePrefix + "envoy.config.bootstrap.v3.BootstrapPath"
```
howardjohn, line 39: looks like a bad refactor?

pilot/pkg/model/context.go
```diff
@@ -857,6 +876,11 @@ func (node *Proxy) SupportsIPv6() bool {
 	return node.ipv6Support
 }
 
+// GRPCGen returns true if the "proxy" expects gRPC-compatible xDS responses.
+func (node *Proxy) GRPCGen() bool {
+	return node.Metadata.Generator == "grpc"
```
howardjohn, line 881: Does grpc gracefully handle the stuff we put in cluster? the mtls (especially auto mtls), custom filters for telemetry, etc?

pilot/pkg/model/context.go
stevenctl: Yes, but even that could change. I'll just add a util to generate a `Set` and query `empty or contains`

pilot/pkg/networking/grpcgen/grpcgen.go
stevenctl: no, this is for experimentation

stevenctl: Although it could be useful if someone only needs basic service discovery and has found a bug in the re-use path. Probably shouldn't be using in a non-experimental environment anyway.

pilot/pkg/model/context.go
stevenctl: No. That's the biggest concern here – the goal is to attempt to use as many istio configs with gRPC as via integration tests to find the breakages. 

At what point does it become "too many" gRPC only branches in the generator?

pilot/pkg/networking/core/configgen.go
adiprerepa: agree -- `model.WatchedResource` is exclusively for `DiscoveryRequest` and `DiscoveryResponse`

adiprerepa: agree -- model.WatchedResource is exclusively for DiscoveryRequest and DiscoveryResponse

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -69,7 +69,7 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
 // For outbound: Cluster for each service/subset hostname or cidr with SNI set to service hostname
 // Cluster type based on resolution
 // For inbound (sidecar only): Cluster for each inbound endpoint port and for each service port
-func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *model.PushContext) []*cluster.Cluster {
+func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *model.PushContext, w *model.WatchedResource) []*cluster.Cluster {
```
hzxuzhonghu, line 72: I believe it is better to not pollute the BuildClusters implement, for grpc gen, can you call buildOutboundClusters if possible.

pilot/pkg/model/context.go
```diff
@@ -339,6 +339,25 @@ type WatchedResource struct {
 	// Note that Envoy may send multiple requests for the same type, for
 	// example to update the set of watched resources or to ACK/NACK.
 	LastRequest *discovery.DiscoveryRequest
+
+	// filter is the internal map allowing filters based on ResourceNames
+	filter map[string]struct{}
```
hzxuzhonghu, line 344: This field is not refreshed?

pilot/pkg/networking/core/v1alpha3/cluster.go
stevenctl: I think I agree as I test this further.

Issue #33712, opened 2021-06-29
Combine CNI install and repair container


bianpengyuan:
Separation of install & repair makes monitoring and debugging unnecessarily complicated. We can combine these two containers into one.

bianpengyuan:
This is done.

PR #33713, opened 2021-06-29
Combine CNI repair and install containers.

bianpengyuan:
Current package organization in cni/ folder is a bit awkward after this change, and there will be a follow up PR to make it align with this combination. Also there are several other clean ups such as log scope and config package consolidation which will be done in the followups.

https://github.com/istio/istio/issues/33712

### DIFF
--- a/cni/pkg/repair/options.go
+++ b/cni/pkg/repair/options.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Istio Authors
+// Copyright Istio Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


cni/pkg/repair/options.go
```diff
@@ -73,6 +58,7 @@ func parseFlags() (filters *repair.Filters, options *ControllerOptions) {
 	pflag.String("field-selectors", "", "A set of field selectors in label=value format that will be added to the pod list filters")
 
 	// Repair Options
+	pflag.Bool("enabled", true, "Whether enable race condition repair or not.")
```
howardjohn, line 61: is `enabled` scoped in some way to indicate its for repair?

howardjohn (approved):

cni/pkg/repair/options.go
bianpengyuan: Yeah it has `viper.SetEnvPrefix("REPAIR")` call underneath, but actually I am not sure the implication of setting that since install is also using viper. I updated to remove that and attach `repair` prefix to all flags directly.

cni/pkg/repair/options.go
```diff
@@ -0,0 +1,149 @@
...
+	"istio.io/pkg/log"
+)
+
+type controllerOptions struct {
+	RepairOptions *Options `json:"repair_options"`
+	RunAsDaemon   bool     `json:"run_as_daemon"`
+	Enabled       bool     `json:"enabled"`
+}
+
+func init() {
```
mandarjog, line 34: Is this all net new code including options?

bianpengyuan: They were copied from the original repair main.go.

PR #33715, opened 2021-06-30
Fix split-horizon LB for MCS

### LINKED ISSUE #33526: MCS issues with multi-network configurations
**Bug description**

This bug was discovered while working to [add MCS support to Istio](https://github.com/istio/istio/pull/32863).

The k8s MCS spec allows exporting endpoints for a service on a cluster-by-cluster basis, so you can be in a state where some endpoints for the service are discoverable mesh-wide and others are only discoverable within the same cluster.

In multi-network configurations (i.e. where endpoints are accessed via gateways) if a service's endpoints are only exported in a subset of clusters, non-exported endpoints may be incorrectly accessible from outside of its cluster.

Consider the following configuration with endpoints for `Service A` in each cluster and a `ServiceExport` only in `Cluster 2`:

- Network 1:
  - Cluster 1:
    - eastwestgateway
    - svc-a-ep1
  - Cluster 2:
    - eastwestgateway
    - svc-a-ep2
    - **ServiceExport: svc-a**
- Network 2:
  - Cluster 3:
    - eastwestgateway
    - svc-a-ep3

When sending traffic from `Cluster 3`, traffic should be load balanced between `svc-a-ep2` (since it was exported) and `svc-a-ep3` (the in-cluster endpoint). However, it's actually able to hit all of the endpoints.

Here's my understanding (guess?) as to what's going on...

The way that we look up gateways is by *network*, not *cluster*. The client in `Cluster 3` only sees `svc-a-ep2` and `svc-a-ep3`, but when it tries to reach `svc-a-ep2` in the other network, it tries to reach it by loadbalancing across all of the gateways for that network. In effect, the fact that we're going through a gateway bypasses the export policy of the endpoint.

It should be noted that this is **not** a problem for the existing `Cluster Local` logic, which is applied at the service (rather than endpoint) level. So all endpoints for the service across the mesh are either local to the cluster or mesh-wide, and we never have this in-between state which triggers this bug. 

[ ] Docs
[ ] Installation
[x] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

**How was Istio installed?**

nmittler:
Previous logic was broken for the case when only some clusters export service endpoints (via MCS ServiceExport). In this case, it's possible that we can accidentally hit unexported endpoints that share the same network as exported endpoints.

Refactoring the load balancing logic to prefer gateways with the same cluster as the endpoint. Going forward, we should recommend installing east-west gateway in all clusters to support the MCS use case.

Fixes #33526

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/networking/util/util.go
+++ b/pilot/pkg/networking/util/util.go
@@ -490,14 +490,6 @@ func BuildLbEndpointMetadata(networkID network.ID, tlsMode, workloadname, namesp
 		FilterMetadata: map[string]*pstruct.Struct{},
 	}
 
-	if networkID != "" {
-		addIstioEndpointLabel(metadata, "network", &pstruct.Value{Kind: &pstruct.Value_StringValue{StringValue: networkID.String()}})
-	}
-	// TODO(nmittler): Is there any concern with adding an extra label?
-	if clusterID != "" {
-		addIstioEndpointLabel(metadata, "cluster", &pstruct.Value{Kind: &pstruct.Value_StringValue{StringValue: clusterID.String()}})
-	}
-
 	if tlsMode != "" && tlsMode != model.DisabledTLSModeLabel {
 		metadata.FilterMetadata[EnvoyTransportSocketMetadataKey] = &pstruct.Struct{
 			Fields: map[string]*pstruct.Value{
--- a/pilot/pkg/networking/util/util.go
+++ b/pilot/pkg/networking/util/util.go
@@ -480,9 +480,9 @@ func MergeAnyWithAny(dst *any.Any, src *any.Any) (*any.Any, error) {
 }
 
 // BuildLbEndpointMetadata adds metadata values to a lb endpoint
-func BuildLbEndpointMetadata(network network.ID, tlsMode, workloadname, namespace string,
+func BuildLbEndpointMetadata(networkID network.ID, tlsMode, workloadname, namespace string,
 	clusterID cluster.ID, labels labels.Instance) *core.Metadata {
-	if network == "" && (tlsMode == "" || tlsMode == model.DisabledTLSModeLabel) && !features.EndpointTelemetryLabel {
+	if networkID == "" && (tlsMode == "" || tlsMode == model.DisabledTLSModeLabel) && !features.EndpointTelemetryLabel {
 		return nil
 	}
 
@@ -490,8 +490,12 @@ func BuildLbEndpointMetadata(network network.ID, tlsMode, workloadname, namespac
 		FilterMetadata: map[string]*pstruct.Struct{},
 	}
 
-	if network != "" {
-		addIstioEndpointLabel(metadata, "network", &pstruct.Value{Kind: &pstruct.Value_StringValue{StringValue: network.String()}})
+	if networkID != "" {
+		addIstioEndpointLabel(metadata, "network", &pstruct.Value{Kind: &pstruct.Value_StringValue{StringValue: networkID.String()}})
+	}
+	// TODO(nmittler): Is there any concern with adding an extra label?
+	if clusterID != "" {
+		addIstioEndpointLabel(metadata, "cluster", &pstruct.Value{Kind: &pstruct.Value_StringValue{StringValue: clusterID.String()}})
 	}
 
 	if tlsMode != "" && tlsMode != model.DisabledTLSModeLabel {
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -315,11 +315,9 @@ func (s *DiscoveryServer) generateEndpoints(b EndpointBuilder) *endpoint.Cluster
 		return buildEmptyClusterLoadAssignment(b.clusterName)
 	}
 
-	// If networks are set (by default they aren't) apply the Split Horizon
-	// EDS filter on the endpoints
-	if b.push.NetworkManager().IsMultiNetworkEnabled() {
-		llbOpts = b.EndpointsByNetworkFilter(llbOpts)
-	}
+	// Apply the Split Horizon EDS filter, if applicable.
+	llbOpts = b.EndpointsByNetworkFilter(llbOpts)
+
 	if model.IsDNSSrvSubsetKey(b.clusterName) {
 		// For the SNI-DNAT clusters, we are using AUTO_PASSTHROUGH gateway. AUTO_PASSTHROUGH is intended
 		// to passthrough mTLS requests. However, at the gateway we do not actually have any way to tell if the
--- a/pilot/pkg/xds/eds_sh_test.go
+++ b/pilot/pkg/xds/eds_sh_test.go
@@ -92,20 +92,20 @@ func TestSplitHorizonEds(t *testing.T) {
 			want: expectedResults{
 				weights: map[string]uint32{
 					// 1 local endpoint
-					"10.1.0.1": 4,
+					"10.1.0.1": 2,
 
 					// 2 endopints on network 2, go through single gateway.
-					"159.122.219.2": 8,
+					"159.122.219.2": 4,
 
 					// 3 endpoints on network 3, weights split across 2 gateways.
-					"159.122.219.3": 6,
-					"179.114.119.3": 6,
+					"159.122.219.3": 3,
+					"179.114.119.3": 3,
 
 					// no gateway defined for network 4 - treat as directly reachable.
-					"10.4.0.1": 4,
-					"10.4.0.2": 4,
-					"10.4.0.3": 4,
-					"10.4.0.4": 4,
+					"10.4.0.1": 2,
+					"10.4.0.2": 2,
+					"10.4.0.3": 2,
+					"10.4.0.4": 2,
 				},
 			},
 		},
@@ -117,21 +117,21 @@ func TestSplitHorizonEds(t *testing.T) {
 			want: expectedResults{
 				weights: map[string]uint32{
 					// 2 local endpoints
-					"10.2.0.1": 4,
-					"10.2.0.2": 4,
+					"10.2.0.1": 2,
+					"10.2.0.2": 2,
 
 					// 1 endpoint on network 1, accessed via gateway.
-					"159.122.219.1": 4,
+					"159.122.219.1": 2,
 
 					// 3 endpoints on network 3, weights split across 2 gateways.
-					"159.122.219.3": 6,
-					"179.114.119.3": 6,
+					"159.122.219.3": 3,
+					"179.114.119.3": 3,
 
 					// no gateway defined for network 4 - treat as directly reachable.
-					"10.4.0.1": 4,
-					"10.4.0.2": 4,
-					"10.4.0.3": 4,
-					"10.4.0.4": 4,
+					"10.4.0.1": 2,
+					"10.4.0.2": 2,
+					"10.4.0.3": 2,
+					"10.4.0.4": 2,
 				},
 			},
 		},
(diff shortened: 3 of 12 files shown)


pilot/pkg/networking/util/util.go
```diff
@@ -480,18 +480,22 @@ func MergeAnyWithAny(dst *any.Any, src *any.Any) (*any.Any, error) {
...
 	metadata := &core.Metadata{
 		FilterMetadata: map[string]*pstruct.Struct{},
 	}
 
-	if network != "" {
-		addIstioEndpointLabel(metadata, "network", &pstruct.Value{Kind: &pstruct.Value_StringValue{StringValue: network.String()}})
+	if networkID != "" {
+		addIstioEndpointLabel(metadata, "network", &pstruct.Value{Kind: &pstruct.Value_StringValue{StringValue: networkID.String()}})
+	}
+	// TODO(nmittler): Is there any concern with adding an extra label?
```
nmittler, line 496: @howardjohn any concerns here?

howardjohn, line 496: Yes. I don't have the numbers off hand but I think even adding one label leads to a big impact on EDS cost (since there are so many endpoints). Why do we need it ?

Actually confused why we have `network` as well.. We should check the benchmark results

pilot/pkg/xds/ep_filters.go
```diff
@@ -61,73 +60,91 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocLbEndpointsAn
...
+		// Create a map to keep track of the gateways used and their aggregate weights.
+		gatewayWeights := make(map[model.NetworkGateway]uint32)
+
+		// Process all of the endpoints.
 		for i, lbEp := range ep.llbEndpoints.LbEndpoints {
+			// Copy the endpoint in order to expand the load balancing weight. This will allow
+			// us to more easily spread traffic to the endpoint across multiple network gateways, increasing
+			// reliability of the endpoint. When multiplying, be careful to avoid overflow - clipping the
+			// result at the maximum value for uint32.
+			weight := b.scaleEndpointLBWeight(lbEp, endpointLBWeightScaleFactor)
```
stevenctl, line 72: Two comments:
* For the most part the changes here and the problem you're solving could be dealt with simply by filtering the gateways before calculating multiples/performing filtering _but_  it may be easier to understand (took me a while to wrap my head around the original logic)
* I'm not confident a constant scale factor is a good idea. We _should_ be able to calculate one that guarantees divisibility by len(gateways):
   * `factorial(max(possibleGatewayLens))`

nmittler: > For the most part the changes here and the problem you're solving could be dealt with simply by filtering the gateways before calculating multiples/performing filtering but it may be easier to understand (took me a while to wrap my head around the original logic)

Agreed, refactoring the logic wasn't necessary, but I think it is an improvement for understandability and future maintenance as a result.

> I'm not confident a constant scale factor is a good idea. We should be able to calculate one that guarantees divisibility by len(gateways)

Do you have a specific case in mind as to why it might not be a good idea? This is a reasonably hot data path, so I'd like to avoid a lot of extra traversals if possible. Acknowledge that 4 is a bit of a magic number, but it's also reasonably high WRT spreading traffic across gateways on the same network or network+cluster.

pilot/pkg/networking/util/util.go
howardjohn: Only one way to find out: https://github.com/istio/istio/pull/33717

nmittler: @stevenctl ^^

I suspect the reason for `network` was similar to why I added `cluster` ... some information from `IstioEndpoint` is lost post transformation to the Envoy structs.

pilot/pkg/xds/ep_filters.go
```diff
@@ -138,6 +155,24 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocLbEndpointsAn
 	return filtered
 }
 
+func (b *EndpointBuilder) selectNetworkGateways(nw network.ID, c cluster.ID) []*model.NetworkGateway {
```
stevenctl, line 158: This is a big improvement – can you add a comment explaining the benefits:
* Potentially reduce hops
* MCS case

pilot/pkg/xds/ep_filters.go
stevenctl: The second we hit 5 gateways we can have issues where we don't maintain the original balance (if only slightly). It's the sort of subtle problem that would be hard to diagnose in the future. Systems or vendors that prefer several gateways for reliability could be quick to cause issues. 

The factorial logic isn't very different from our original multiplier calculation, if anything it's slightly quicker and would produce a smaller common factor. 

I don't feel _too_ strongly, but I don't think the cost is high. We can check with benchmarks.

pilot/pkg/networking/util/util.go
stevenctl: We only need it for this in-control-plane logic. If we can fetch it after the fact (find the original istio endpoints) or if we can strip it after this is used that would be good. 

For the "stripping after use" case, we probably want to only add it conditionally, based on MultiNetworkEnabled

pilot/pkg/xds/eds.go
```diff
@@ -315,11 +315,9 @@ func (s *DiscoveryServer) generateEndpoints(b EndpointBuilder) *endpoint.Cluster
 		return buildEmptyClusterLoadAssignment(b.clusterName)
 	}
 
-	// If networks are set (by default they aren't) apply the Split Horizon
-	// EDS filter on the endpoints
-	if b.push.NetworkManager().IsMultiNetworkEnabled() {
```
hzxuzhonghu, line 320: nit: I prefer this way

nmittler: I'd rather have this sort of logic bundled into the network filter and always apply it. It makes the call site simpler and pushes all the networking-related decisions down into the filter.

pilot/pkg/networking/util/util.go
nmittler: I've removed `network` and `cluster` from LBEndpoint metadata.

howardjohn:
I would like to review this closely but may not find time today, can you bug me if I don't get to it and am blocking?

nmittler:
@howardjohn @stevenctl I think I've addressed all of your comments. I ended up fishing the cluster/network through by passing the original `IstioEndpoint` to the endpoint builder. PTAL

stevenctl:
At a glance this looks good, but I am curious how `kb/msg` is going down so much. I haven't reviewed the latest changes, are we dropping some metadata? 

```
name                                          old time/op        new time/op        delta
EndpointGeneration/1/100-8                          1.78ms ± 4%        1.45ms ± 2%  -18.52%  (p=0.008 n=5+5)
EndpointGeneration/10/10-8                          1.22ms ± 2%        1.01ms ± 1%  -17.24%  (p=0.008 n=5+5)
EndpointGeneration/100/10-8                         12.1ms ± 1%        10.1ms ± 1%  -16.22%  (p=0.008 n=5+5)
EndpointGeneration/1000/1-8                         12.5ms ± 2%        10.3ms ± 2%  -17.59%  (p=0.008 n=5+5)
name                                          old alloc/op       new alloc/op       delta
EndpointGeneration/1/100-8                           444kB ± 0%         374kB ± 0%  -15.87%  (p=0.008 n=5+5)
EndpointGeneration/10/10-8                           272kB ± 0%         245kB ± 0%   -9.94%  (p=0.008 n=5+5)
EndpointGeneration/100/10-8                         2.58MB ± 0%        2.33MB ± 0%   -9.47%  (p=0.016 n=4+5)
EndpointGeneration/1000/1-8                         2.57MB ± 0%        2.34MB ± 0%   -9.15%  (p=0.008 n=5+5)
name                                          old allocs/op      new allocs/op      delta
EndpointGeneration/1/100-8                           8.41k ± 0%         6.91k ± 0%  -17.84%  (p=0.008 n=5+5)
EndpointGeneration/10/10-8                           5.08k ± 0%         4.34k ± 0%  -14.59%  (p=0.008 n=5+5)
EndpointGeneration/100/10-8                          46.8k ± 0%         39.8k ± 0%  -15.07%  (p=0.016 n=4+5)
EndpointGeneration/1000/1-8                          46.3k ± 0%         39.3k ± 0%  -15.22%  (p=0.008 n=5+5)
name                                          old kb/msg         new kb/msg         delta
EndpointGeneration/1/100-8                            11.5 ± 0%           9.1 ± 0%  -20.89%  (p=0.008 n=5+5)
EndpointGeneration/10/10-8                            8.53 ± 0%          6.13 ± 0%  -28.14%  (p=0.008 n=5+5)
EndpointGeneration/100/10-8                           83.2 ± 0%          59.2 ± 0%  -28.84%  (p=0.008 n=5+5)
EndpointGeneration/1000/1-8                           83.6 ± 0%          59.6 ± 0%  -28.71%  (p=0.008 n=5+5)
name                                          old resources/msg  new resources/msg  delta
EndpointGeneration/1/100-8                             100 ± 0%           100 ± 0%     ~     (all equal)
EndpointGeneration/10/10-8                            10.0 ± 0%          10.0 ± 0%     ~     (all equal)
EndpointGeneration/100/10-8                           10.0 ± 0%          10.0 ± 0%     ~     (all equal)
EndpointGeneration/1000/1-8                           1.00 ± 0%          1.00 ± 0%     ~     (all equal)
```

howardjohn (commented):
@stevenctl I believe this incorporates https://github.com/istio/istio/pull/33717

nmittler:
Yeah that's right

stevenctl (approved):

pilot/pkg/xds/eds_sh_test.go
```diff
@@ -296,3 +337,12 @@ func addNetwork(server *xds.FakeDiscoveryServer, id network.ID, network *meshcon
 	meshNetworks.Networks = c
 	server.Env().SetNetworks(&meshNetworks)
 }
+
+func getLbEndpointAddrs(eps []*endpoint.LbEndpoint) []string {
+	addrs := make([]string, 0)
+	for _, lbEp := range eps {
+		addrs = append(addrs, lbEp.GetEndpoint().Address.GetSocketAddress().Address)
+	}
+	sort.Strings(addrs)
```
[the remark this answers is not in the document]
nmittler: Discussed offline. Decided that this sort should remain, so that the return values are deterministic. The eds test has a sort as well, which was the main concern here, but it's only used for formatting for the error strings - not validation.

pilot/pkg/xds/eds_sh_test.go
```diff
@@ -271,13 +309,16 @@ func initRegistry(server *xds.FakeDiscoveryServer, clusterNum int, gatewaysIP []
 	})
 	istioEndpoints := make([]*model.IstioEndpoint, numOfEndpoints)
 	for i := 0; i < numOfEndpoints; i++ {
+		addr := fmt.Sprintf("10.%d.0.%d", networkNum, i+1)
+		fmt.Printf("NM: adding address %v\n", addr)
```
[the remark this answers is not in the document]
nmittler: Fixed by https://github.com/istio/istio/pull/33776

pilot/pkg/xds/ep_filters.go
```diff
@@ -61,73 +61,101 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocLbEndpointsAn
...
+			// TODO BTS may allow us to work around this
+			if b.mtlsChecker.isMtlsDisabled(lbEp) {
+				continue
+			}
+
+			// Apply the weight for this endpoint to the network gateways.
+			remainingWeight := weight
+			for remainingWeight > 0 {
+				// Spread the remaining weight across the gateways.
+				weightPerGateway := remainingWeight / uint32(len(gateways))
```
hzxuzhonghu, line 107: How about 3/2? How can it reflect the real proportion
It is confusing how the weight is calculated.

pilot/pkg/xds/ep_filters.go
```diff
@@ -61,73 +61,101 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocLbEndpointsAn
...
+			weight := b.scaleEndpointLBWeight(lbEp, scaleFactor)
+			lbEp := proto.Clone(lbEp).(*endpoint.LbEndpoint)
+			lbEp.LoadBalancingWeight = &wrappers.UInt32Value{
+				Value: weight,
+			}
+
+			istioEndpoint := ep.istioEndpoints[i]
+			epNetwork := istioEndpoint.Network
+			epCluster := istioEndpoint.Locality.ClusterID
+			gateways := b.selectNetworkGateways(epNetwork, epCluster)
```
hzxuzhonghu, line 81: Are only cluster local endpoints visible to the cluster gateway? How did you make it?

nmittler: The gateway (like all proxies) will only see endpoints for the service that are cluster local or have been exported.

pilot/pkg/xds/ep_filters.go
nmittler: All we're doing here is spreading the weight for the endpoint across all gateways. We could simply round robin across each gateway, subtracting 1 from the endpoint's weight each time (in essence, that's what this code is doing). The division here is just a way to quickly determine the portion of the endpoint weight that each gateway should receive. This calculation should never be `0`, since we scale each endpoint weight by the maximum gateways per network. But even if it were, we adjust it to `1` below.

PR #27449, opened 2020-09-21
Fix issue with short-circuiting match all routes

howardjohn:
Fixes https://github.com/istio/istio/issues/26263 by reverting
https://github.com/istio/istio/pull/19471

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any changes that may affect Istio users.

PR #33720, opened 2021-06-30
Do not add other route when catchall route is built

hzxuzhonghu:
When there are multi HTTPRoute.Match set, if one is catch all, then ignore other match conditions, so no need to build envoy route for them.

VirtualHost.Routes will be matched in order.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -271,29 +271,30 @@ func BuildHTTPRoutesForVirtualService(
...
-			for _, match := range http.Match {
-				if r := translateRoute(push, node, http, match, listenPort, virtualService, serviceRegistry, gatewayNames); r != nil {
+			return out, nil
+		}
+
+		tmp := []*route.Route{}
+		for _, match := range http.Match {
+			if r := translateRoute(push, node, http, match, listenPort, virtualService, serviceRegistry, gatewayNames); r != nil {
+				tmp = append(tmp, r)
+				// This is a catch all path. Routes are matched in order, so we will never go beyond this match
```
howardjohn, line 287: I vaguely recall we used to do this and reverted it for some reason.. let me see if I can dig up the history

howardjohn: https://github.com/istio/istio/pull/27449 was what I was thinking of

hzxuzhonghu: HTTPRoute.Match
	// Match conditions to be satisfied for the rule to be
	// activated. All conditions inside a single match block have AND
	// semantics, while the list of match blocks have OR semantics. The rule
	// is matched if any one of the match blocks succeed.

This is the api doc, it says it matches when anyone condition matches. If there is a match all, all other routes are redundant

hzxuzhonghu: @howardjohn After a close look at the issue, it is different, with this pr it does not remove the `provider-global-default` catch all route. There is ut for this. And i am testing manually in my env

hzxuzhonghu: Verified it does not remove the catchall `provider-global-default`
```
[
    {
        "name": "8000",
        "virtualHosts": [
            {
                "name": "allow_any",
                "domains": [
                    "*"
                ],
                "routes": [
                    {
                        "name": "allow_any",
                        "match": {
                            "prefix": "/"
                        },
                        "route": {
                            "cluster": "PassthroughCluster",
                            "timeout": "0s",
                            "maxGrpcTimeout": "0s"
                        }
                    }
                ],
                "includeRequestAttemptCount": true
            },
            {
                "name": "httpbin.default.svc.cluster.local:8000",
                "domains": [
                    "httpbin.default.svc.cluster.local",
                    "httpbin.default.svc.cluster.local:8000",
                    "httpbin",
                    "httpbin:8000",
                    "httpbin.default.svc",
                    "httpbin.default.svc:8000",
                    "httpbin.default",
                    "httpbin.default:8000",
                    "10.96.196.41",
                    "10.96.196.41:8000"
                ],
                "routes": [
                    {
                        "name": "GET-/",
                        "match": {
                            "prefix": "/",
                            "caseSensitive": true,
                            "headers": [
                                {
                                    "name": ":method",
                                    "exactMatch": "GET"
                                }
                            ]
                        },
                        "route": {
                            "cluster": "outbound|8000||httpbin.default.svc.cluster.local",
                            "timeout": "4s",
                            "retryPolicy": {
                                "retryOn": "gateway-error,connect-failure,retriable-status-codes",
                                "numRetries": 3,
                                "perTryTimeout": "1s",
                                "retryHostPredicate": [
                                    {
                                        "name": "envoy.retry_host_predicates.previous_hosts"
                                    }
                                ],
                                "hostSelectionRetryMaxAttempts": "5"
                            },
                            "maxGrpcTimeout": "4s"
                        },
                        "metadata": {
                            "filterMetadata": {
                                "istio": {
                                    "config": "/apis/networking.istio.io/v1alpha3/namespaces/default/virtual-service/bar"
                                }
                            }
                        },
                        "decorator": {
                            "operation": "httpbin.default.svc.cluster.local:8000/*"
                        },
                        "requestHeadersToAdd": [
                            {
                                "header": {
                                    "key": "x-envoy-retriable-status-codes",
                                    "value": "408,423"
                                },
                                "append": true
                            }
                        ]
                    },
                    {
                        "name": "provider-global-default",
                        "match": {
                            "prefix": "/"
                        },
                        "route": {
                            "cluster": "outbound|8000||httpbin.default.svc.cluster.local",
                            "timeout": "0s",
                            "retryPolicy": {
                                "retryOn": "connect-failure,refused-stream,unavailable,cancelled,retriable-status-codes",
                                "numRetries": 2,
                                "retryHostPredicate": [
                                    {
                                        "name": "envoy.retry_host_predicates.previous_hosts"
                                    }
                                ],
                                "hostSelectionRetryMaxAttempts": "5",
                                "retriableStatusCodes": [
                                    503
                                ]
                            },
                            "maxGrpcTimeout": "0s"
                        },
                        "metadata": {
                            "filterMetadata": {
                                "istio": {
                                    "config": "/apis/networking.istio.io/v1alpha3/namespaces/default/virtual-service/bar"
                                }
                            }
                        },
                        "decorator": {
                            "operation": "httpbin.default.svc.cluster.local:8000/*"
                        }
                    }
                ],
                "includeRequestAttemptCount": true
            }
        ],
        "validateClusters": false
    }
]

```

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -271,29 +271,30 @@ func BuildHTTPRoutesForVirtualService(
...
-			if r := translateRoute(push, node, http, nil, listenPort, virtualService, serviceRegistry, gatewayNames); r != nil {
+			r := translateRoute(push, node, http, nil, listenPort, virtualService, serviceRegistry, gatewayNames)
+			if r != nil {
 				out = append(out, r)
 			}
-			catchall = true
-		} else {
-			for _, match := range http.Match {
-				if r := translateRoute(push, node, http, match, listenPort, virtualService, serviceRegistry, gatewayNames); r != nil {
+			return out, nil
```
howardjohn, line 280: the old code would return an error if translateRoute returns nil. is the behavior intentionally changed here? (L300 of old code)

pilot/pkg/networking/core/v1alpha3/route/route.go
howardjohn: Thanks for looking into it!

howardjohn (changes requested):
I think this is very close to being a safe optimization. However, my concern is the impact on tracing.

Consider a route like:
```
  - match:
    - name: foo
      uri:
        prefix: /foo
    - uri:
        prefix: /
```

This is actually NOT exactly equivalent to 
```
  - match:
    - uri:
        prefix: /
```

Because we won't get a segment like
```
    - decorator:
        operation: echo.default.svc.cluster.local:80/foo*
```

in the envoy configuration.

One thing I noticed - we do already have a webhook warning for this case. So if a user does accidentally set this, they will be alerted and can change the VS themselves to optimize it?

WDYT?

pilot/pkg/networking/core/v1alpha3/route/route.go
hzxuzhonghu: It is to short circuit it, here it always return a catchall route?

hzxuzhonghu:
Yes. the above case equals to: 
```
- match:
    - uri:
        prefix: /
```

We can improve the validation

howardjohn:
> Yes. the above case equals to:
> 
> ```
> - match:
>     - uri:
>         prefix: /
> ```
> 
> We can improve the validation

That was my point. These are **not** equal. They generate different traces, because the traces include the URL match

hzxuzhonghu:
Yes, the trace could be different, but does that make much sense? In real case, if user allows traffic matching /a or /b or anything, routed to the same destination,  they don't care about what span name it is. There is not stats inferred from this.
 User mostly care about original and destination service, and related attributes from traces.

howardjohn:
I am not sure it's safe to assume that users don't care about it, I could easily see folks depending on this. It would be good to get some more feedback from @istio/wg-policies-and-telemetry-maintainers .

The biggest complaints about Istio we are seeing is subtle breaking changes like this so I am a hesitant to make then where possible

mandarjog:
Do we allow a way to indicate metadata, decorator or typed-filter-config ?

we should not apply the optimization of any of these things are present. For example a filter may use the some route metadata to make a different decision even though the destination is the same.

hzxuzhonghu:
@mandarjog From envoy doc, decorator only influences the trace span name. 

```
// The operation name associated with the request matched to this route. If tracing is
	// enabled, this information will be used as the span name reported for this request.
```

It has nothing todo with metadata. Below is how the span name is generated, I think we are too concerned about such small changes.  Honestly, there is no change that breaks compatibility can match mixer deprecation.

```
func getRouteOperation(in *route.Route, vsName string, port int) string {
	path := "/*"
	m := in.GetMatch()
	ps := m.GetPathSpecifier()
	if ps != nil {
		switch ps.(type) {
		case *route.RouteMatch_Prefix:
			path = m.GetPrefix() + "*"
		case *route.RouteMatch_Path:
			path = m.GetPath()
		case *route.RouteMatch_SafeRegex:
			path = m.GetSafeRegex().GetRegex()
		}
	}
```

PR #33723, opened 2021-06-30
use allowed audiences instead of default audience

### LINKED ISSUE #33722: sidecar authenticate error if use `first-party-jwt` policy
**Bug description**
When I use first-party-jwt as JWT_POLICY, sidecar can not access istiod.

istiod logs:

```logs
2021-06-30T04:47:53.344303Z	warn	serverca	Authentication failed for 10.244.1.53:37692: Authenticator ClientCertAuthenticator at index 0 got error: no verified chain is found. Authenticator KubeJWTAuthenticator at index 1 got error: failed to validate the JWT from cluster "cluster1": the service account authentication returns an error: [invalid bearer token, token audiences ["https://kubernetes.default.svc.cluster.local"] is invalid for the target audiences ["istio-ca"]].
2021-06-30T04:47:53.494182Z	warn	serverca	Authentication failed for 10.244.1.53:37692: Authenticator ClientCertAuthenticator at index 0 got error: no verified chain is found. Authenticator KubeJWTAuthenticator at index 1 got error: failed to validate the JWT from cluster "cluster1": the service account authentication returns an error: [invalid bearer token, token audiences ["https://kubernetes.default.svc.cluster.local"] is invalid for the target audiences ["istio-ca"]].
```

[X] Security

**Expected behavior**
sidecar connect istiod successfully.

kind create cluster
install istio with `--set values.global.jwtPolicy=first-party-jwt`

latest.

kebe7jun:
Fix: https://github.com/istio/istio/issues/33722

### DIFF
--- a/security/pkg/server/ca/authenticate/kubeauth/kube_jwt.go
+++ b/security/pkg/server/ca/authenticate/kubeauth/kube_jwt.go
@@ -74,7 +74,10 @@ func (a *KubeJWTAuthenticator) AuthenticatorType() string {
 	return KubeJWTAuthenticatorType
 }
 
-const DefaultKubernetesAudience = "kubernetes.default.svc"
+var defaultAllowedKubernetesAudiences = map[string]bool{
+	"kubernetes.default.svc":               true,
+	"kubernetes.default.svc.cluster.local": true,
+}
 
 func (a *KubeJWTAuthenticator) AuthenticateRequest(req *http.Request) (*security.Caller, error) {
 	targetJWT, err := security.ExtractRequestToken(req)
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
Do we need to do this? first-party-jwt has been deprecated for 2 years. If users are already customizing the default audience they should probably just not use first-party-jwt? Is there a case where that is not acceptable?

kebe7jun:
> Do we need to do this? first-party-jwt has been deprecated for 2 years. If users are already customizing the default audience they should probably just not use first-party-jwt? Is there a case where that is not acceptable?

Although we don't use it much now, it does cause an unusable bug.
I think we should provide users with this choice.

security/pkg/server/ca/authenticate/kubeauth/kube_jwt.go
```diff
@@ -74,8 +74,6 @@ func (a *KubeJWTAuthenticator) AuthenticatorType() string {
 	return KubeJWTAuthenticatorType
 }
 
-const DefaultKubernetesAudience = "kubernetes.default.svc"
```
howardjohn, line 77: should we just allow `kubernetes.default.svc` OR `kubernetes.default.svc.cluster.local`? 

I think that we should work out of the box with common environments (like https://github.com/istio/istio/issues/33744), and for bespoke cases they should use third-party-jwt. WDYT?

kebe7jun: Ok, fixed.

security/pkg/server/ca/authenticate/kubeauth/kube_jwt.go
```diff
@@ -112,18 +127,18 @@ func (a *KubeJWTAuthenticator) authenticate(targetJWT string, clusterID cluster.
...
 				// For backwards compatibility, if first-party-jwt is used and they don't require 3p, allow it but warn
 				// This is intended to support first-party-jwt on Kubernetes 1.21+, where BoundServiceAccountTokenVolume
-				// became default and started setting an audience to DefaultKubernetesAudience.
+				// became default and started setting an audience to one of defaultAllowedKubernetesAudiences.
 				// Users should disable first-party-jwt, but we don't want to break them on upgrade
 				log.Warnf("Insecure first-party-jwt option used to validate token; use third-party-jwt")
 				aud = nil
 			} else {
-				log.Warnf("Received token with aud %q, but expected %q. BoundServiceAccountTokenVolume, "+
+				log.Warnf("Received token with aud %q, but expected on of %q. BoundServiceAccountTokenVolume, "+
```
howardjohn, line 139:
```suggestion
				log.Warnf("Received token with aud %q, but expected one of %q. BoundServiceAccountTokenVolume, "+
```

security/pkg/server/ca/authenticate/kubeauth/kube_jwt.go
```diff
@@ -112,18 +127,18 @@ func (a *KubeJWTAuthenticator) authenticate(targetJWT string, clusterID cluster.
 	// tolerate the unbound tokens.
 	if !util.IsK8SUnbound(targetJWT) || security.Require3PToken.Get() {
 		aud = security.TokenAudiences
-		if tokenAud, _ := util.ExtractJwtAud(targetJWT); len(tokenAud) == 1 && tokenAud[0] == DefaultKubernetesAudience {
+		if tokenAud, _ := util.ExtractJwtAud(targetJWT); isAllowedAudiences(tokenAud) {
```
howardjohn, line 130: can we keep the len check? Then isAllowedAudiences can be simpler, like:

```go
if _, f := DefaultKubernetesAudience[tokenAud[0]; f {
 ....
}
```

And change DefaultKubernetesAudience to a map instead of a slice, then we can easily lookup.

howardjohn (commented):
lgtm! just minor nits then I will approve. Thanks!

security/pkg/server/ca/authenticate/kubeauth/kube_jwt.go
kebe7jun: Thanks! Updated.

howardjohn (approved):

PR #33724, opened 2021-06-30
Build: refactor to support build multi-arch images

morlay:
Signed-off-by: Morlay <morlay.null@gmail.com>

---

out dir rename from

```
out/linux_amd64/
    pilot-agent
```

to 
```
out/linux/
   pilot-agent-amd64
```

all binaries used in docker with arch suffix.

Then we could be use build multi-arch images (now need to wait istio-proxy for arm64 provided)
```
TARGET_ARCH=amd64 make build-linux
TARGET_ARCH=arm64 make build-linux
make dockerx.push DOCKER_ARCHITECTURES=linux/amd64,linux/arm64 
```

All `${TARGETARCH:-amd64}` in Dockerfiles is for `docker build` if not used `docker buildx`

----

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


howardjohn (commented):
This is great! Thanks!

It does not seem great to change the linux_amd64 folder name or binary names, as it may break a lot of users that are expecting those. It also seems a bit unexpected to have the arch name in the binary, typically those seem to be without suffix.

Would it work to change things like:
* keep out/linux_amd64 and out/linux_arm64 directories
* binary names remain unchanged
* When we copy to `dockerx_build`, we set up directory like:
```
. Dockerfile.pilot
  amd64/
  | pilot-discovery
  arm64/
  | pilot-discovery
```

and in the dockerfiles I guess we need to modify everything from `COPY client` to `COPY ${TARGETARCH}/client`? Or even better, some sort of change dir in the docker context itself, although I don't think that is possible.

The reason I prefer this is it is less disruptive to developers, the only thing that changes is the docker folder format which I don't think anyone really cares about.

WDYT?

cc @ericvn @brian-avery

morlay:
i have try this in local first. 

but i found it is not easy to change the Makefile 

copy target like `docker.app` 

example，`${ISTIO_OUT_LINUX}/client`

we couldn't copy two arch binaries at same time.

we may not added amd64 and arm64 dir here both，because multi arch builds should be optional.

morlay:
@howardjohn 
I created another PR follow your suggestion https://github.com/istio/istio/pull/33763
but have to create a scripts `docker-copy.sh` to process multi-arch binaries copy without change Makefile too much

PR #33727, opened 2021-06-30
Delegation virtual service allows regex match

### LINKED ISSUE #29845: URL regex match NOT support for in delegated VS, makes ingress's url mapping NOT work.
**Bug description**
"url match does not support regex match for delegating" error in case of regex url in delegated VS. 

Thanks to  Virtual Service Chainning #22118, help decouple the routing rules of each service with each other and avoide big Virtualservice.  In product seniro, usually  root VS specifys the mappings of path and delegated VS.  But it turned out that  regex decscribed  ingress path is NOT permitted. It is  common product case  in  kubernets ingress and  istio ingress gateway to use regex url .

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

**How was Istio installed?**

ChenglongGao327:
It is very difficult to check whether two regular expressions are in conflict, so the checking of conflict is not considered for the moment.
Regardless of the HTTPRoute type, the same validation logic is used to validate the networking.HTTPRoute object

[x] Docs
[x] Installation
[ ] Networking
[x] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Fixes #29845

### DIFF
--- a/pilot/pkg/model/virtualservice.go
+++ b/pilot/pkg/model/virtualservice.go
@@ -429,8 +429,13 @@ func stringMatchConflict(root, leaf *networking.StringMatch) bool {
 	if root == nil || leaf == nil {
 		return false
 	}
-	// regex match is not allowed
-	if root.GetRegex() != "" || leaf.GetRegex() != "" {
+
+	// TODO: check regex conflict
+	// if one has regex ,the others can not be set
+	if root.GetRegex() != "" && (leaf.GetRegex() != "" || leaf.GetPrefix() != "" || leaf.GetExact() != "") {
+		return true
+	}
+	if leaf.GetRegex() != "" && (root.GetPrefix() != "" || root.GetExact() != "") {
 		return true
 	}
 	// root is exact match
@@ -459,7 +464,7 @@ func stringMatchConflict(root, leaf *networking.StringMatch) bool {
 		}
 	}
 
-	return true
+	return false
 }
 
 func isRootVs(vs *networking.VirtualService) bool {
--- a/pkg/config/validation/virtualservice.go
+++ b/pkg/config/validation/virtualservice.go
@@ -53,7 +53,7 @@ func validateHTTPRoute(http *networking.HTTPRoute, delegate bool) (errs Validati
 	errs = WrapError(validateHTTPRouteConflict(http, routeType))
 
 	// check http route match requests
-	errs = appendValidation(errs, validateHTTPRouteMatchRequest(http, routeType))
+	errs = appendValidation(errs, validateHTTPRouteMatchRequest(http))
 
 	// header manipulation
 	for name, val := range http.Headers.GetRequest().GetAdd() {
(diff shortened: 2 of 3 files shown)


pilot/pkg/model/virtualservice.go
```diff
@@ -429,9 +429,10 @@ func stringMatchConflict(root, leaf *networking.StringMatch) bool {
 	if root == nil || leaf == nil {
 		return false
 	}
-	// regex match is not allowed
+	// TODO: check regex conflict
 	if root.GetRegex() != "" || leaf.GetRegex() != "" {
-		return true
+		log.Warnf("It is a risk of conflict that either root(%s) or delegate(%s) have regex set, "+
```
howardjohn, line 434: This is not an appropriate place to log, it is not visible to user applying the virtual service and will be very spammy since this is computed many many times

pilot/pkg/model/virtualservice.go
```diff
@@ -459,7 +460,7 @@ func stringMatchConflict(root, leaf *networking.StringMatch) bool {
 		}
 	}
 
-	return true
+	return false
```
howardjohn, line 463: Why is this changed?

howardjohn (changes requested):
Thanks for the PR!

One thing I wanted to clarify - Does this mean the root and leaf can have completely conflict regex and we just use the root one?

If so, why don't we just only allow regex in the root when leaf is `""`?

It would be useful to have some YAML examples of how this is intended to be used.

hzxuzhonghu:
I think it is to leave the responsibility of judging conflicts to users when they enjoy the flexibility.  

>If so, why don't we just only allow regex in the root when leaf is ""?

I agree with this point, this can be done in merging not in validation stage.

pilot/pkg/model/virtualservice.go
ChenglongGao327: > Why is this changed?

Matchtype is one of exact / prefix / regex. If matchtype is regex, remind users to make sure that there is no conflict between root and leaf，so change to false

pilot/pkg/model/virtualservice.go
ChenglongGao327: > This is not an appropriate place to log, it is not visible to user applying the virtual service and will be very spammy since this is computed many many times

i will change the log level to debug and warn the user when the mergeHTTPRoutes function  executes

ChenglongGao327:
> Thanks for the PR!
> 
> One thing I wanted to clarify - Does this mean the root and leaf can have completely conflict regex and we just use the root one?
> 
> If so, why don't we just only allow regex in the root when leaf is `""`?
> 
> It would be useful to have some YAML examples of how this is intended to be used.

ChenglongGao327:
> If so, why don't we just only allow regex in the root when leaf is `""`?

Check regex when merging，if one has regex , the others can not be set

hzxuzhonghu (commented):
lgtm, will leave to @howardjohn

pilot/pkg/model/virtualservice.go
howardjohn: I believe this changes the behavior when non-regex is used. For example, `prefix: ""` would be `true` before but now `false`?

I would expect the regex section to change only, not the fallback value.

howardjohn (commented):
Can we add some real tests using this? Right now I am not sure if this will work or not since there is no test coverage beyond the validation

fedyiv:
Hi @ramaraochavali , @ChenglongGao327 do you have any plans to proceed with this feature?

fedyiv:
Thanks a lot @ramaraochavali! Could you please advise if it is possible to predict in what istio release this feature might be included? I checked commits included into 1.14.1 and it was not among them

ericvn:
The 1.14 branch cut was roughly April 21st. Since the mentioned PR merged into the main branch the beginning of May it would have been after 1.14 was cut. The commit will be in 1.15.

Issue #33491, opened 2021-06-17
TCP liveness and readiness probes


modulo11:
**Describe the feature request**

The support for TCP probes in Istio is currently broken, but there appears to be a need (#15178 #28737) and it would be beneficial for us as well. We created a [PoC](https://github.com/istio/istio/compare/master...sap-contributions:translate-port-probes) for that which we'd like feedback for. It works the following way:

- pilot-agent offers a HTTP endpoint e.g. under /check/{port}
- If a request comes in, it checks if that port is opened in the running container
- If it's open, return 200 - if not 404

Already present TCP probes would then be rewritten to HTTP based probes querying the above endpoint. As it's running as a sidecar, the check can go against localhost.

If you're interested in such a solution, we'd be willing to do the work getting it production ready.

**Describe alternatives you've considered**
- None

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

PR #28737, opened 2020-11-10
Add health check using TCP

shamsher31:
Based on https://github.com/istio/istio.io/pull/8246#issuecomment-704190021

Follow up for https://github.com/istio/istio.io/pull/8246

```
$ kubectl get pods
NAME                             READY   STATUS    RESTARTS   AGE
liveness-http-86c7cc8f45-s4h7m   1/1     Running   0          1m
liveness-tcp-59f6d56f7-69f7k     1/1     Running   0          2m
```
Using the HTTP probe
```
$ kubectl describe pod liveness-http-86c7cc8f45-s4h7m
...
Liveness:       http-get http://:8001/foo delay=5s timeout=1s period=5s #success=1 #failure=3
```

Using TCP probe
```
$ kubectl describe pod liveness-tcp-59f6d56f7-69f7k
...
Liveness:       tcp-socket :8001 delay=5s timeout=1s period=5s #success=1 #failure=3
```

Issue #15178, opened 2019-06-27
istio should support tcp liveness and readiness probes  of applications


kuaikuai:
the part of  the yaml file as following:
```
    spec:
      containers:
      - image: hello-node:v1
        imagePullPolicy: IfNotPresent
        name: hello-node
        ports:
        - containerPort: 8080
        resources: {}
        readinessProbe:
          tcpSocket:
            port: 8080
          initialDelaySeconds: 5
          periodSeconds: 10
        livenessProbe:
          tcpSocket:
            port: 8080
          initialDelaySeconds: 15
          periodSeconds: 20
```
I enter into the  envoy pod,  and use an iptables rule to disable 8080 ports so that analog apps are not available. 
`iptables -A INPUT -p tcp -d 127.0.0.1 --dport 8080 -j DROP`
k8s doesn't detect the app are no longer serviceable.

hzxuzhonghu:
For tcp probe, kubelet will try to establish a connection with podip:port by default.

With sidecar injected, the kubelet can still establish a connection with envoy , so the probe still thinks it is healthy. Tcp connection  with app will be established until there is real data transport.

kuaikuai:
@hzxuzhonghu,
I have two ideas：
1. change proxy_init, and a iptables rule, let kubelet connect the app directly.
or
2. change istioctl kube-inject & proxy_agent & pod spec, use proxy_agent detect the app.
    add annotations like: sidecar.istio.io/rewriteAppHTTPProbers: "true"

hzxuzhonghu:
I think 2 is feasible.  
If you would openup a pr i would be glad to review, though personally i think k8s tcp probe does not make much sense.

bappr:
@hzxuzhonghu  
> though personally i think k8s tcp probe does not make much sense.

Can u explain more about that plz?
I am fully using tcp probes cause my apps are not customisable. And it's a nice generic way to test the readiness and liveness of your servers.
Therefore, I am getting hit by this issue.

```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: nginx
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
      annotations:
        sidecar.istio.io/inject: "true"
    spec:
      containers:
      - image: nginx
        livenessProbe:
          tcpSocket:
            port: 8080
        name: main
        ports:
        - containerPort: 8080
          name: http
        readinessProbe:
[...]

PR #33734, opened 2021-06-30
Translate port probes into http probes

c0d1ngm0nk3y:
TCP probes do currently not work with istio sidecar, but there appears to be a need (#15178 #28737).

We have created this draft to collect feedback for a possible solution. It works the following way:

* pilot-agent offers a HTTP endpoint e.g. under /check/{port}
* If a request comes in, it checks if that port is opened in the running container
* If it's open, return 200 - if not 404
* Already present TCP probes would then be rewritten to HTTP based probes querying the above endpoint. As it's running as a sidecar, the check can go against localhost.

see #33491 

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pilot/cmd/pilot-agent/status/server.go
+++ b/pilot/cmd/pilot-agent/status/server.go
@@ -658,8 +658,9 @@ func (s *Server) handleAppProbeTCPSocket(w http.ResponseWriter, prober *Prober) 
 	port := prober.TCPSocket.Port.IntValue()
 	timeout := time.Duration(prober.TimeoutSeconds) * time.Second
 
-	// FIXME(phil9909): Currently, the IP table rules still seem to redirect traffic from inside the pod into the envoy when connection using the external IP of the pod
-	// Therefore we hardcode it to localhost for now. See https://istio.io/latest/blog/2021/upcoming-networking-changes/
+	// FIXME(phil9909): Currently, the IP table rules still seem to redirect traffic from inside the pod into the envoy
+	// when connection using the external IP of the pod. Therefore we hardcode it to localhost for now.
+	// See https://istio.io/latest/blog/2021/upcoming-networking-changes/
 	// addr := fmt.Sprintf("%s:%d", s.appProbersDestination, port)
 	addr := fmt.Sprintf("localhost:%d", port)
 	conn, err := net.DialTimeout("tcp", addr, timeout)
--- a/pilot/cmd/pilot-agent/status/server.go
+++ b/pilot/cmd/pilot-agent/status/server.go
@@ -213,39 +213,51 @@ func NewServer(config Options) (*Server, error) {
 	s.appProbeClient = make(map[string]*http.Client, len(s.appKubeProbers))
 	// Validate the map key matching the regex pattern.
 	for path, prober := range s.appKubeProbers {
-		if prober.HTTPGet == nil {
-			// client only needed for http probes
-			continue
+		err := validateAppKubeProber(path, prober)
+		if err != nil {
+			return nil, err
 		}
-		if !appProberPattern.Match([]byte(path)) {
-			return nil, fmt.Errorf(`invalid key, must be in form of regex pattern %v`, appProberPattern)
-		}
-		if prober.HTTPGet == nil {
-			return nil, fmt.Errorf(`invalid prober type, must be of type httpGet`)
-		}
-		if prober.HTTPGet.Port.Type != intstr.Int {
-			return nil, fmt.Errorf("invalid prober config for %v, the port must be int type", path)
-		}
-		localAddr := UpstreamLocalAddressIPv4
-		if config.IPv6 {
-			localAddr = UpstreamLocalAddressIPv6
-		}
-		d := &net.Dialer{
-			LocalAddr: localAddr,
-		}
-		// Construct a http client and cache it in order to reuse the connection.
-		s.appProbeClient[path] = &http.Client{
-			Timeout: time.Duration(prober.TimeoutSeconds) * time.Second,
-			// We skip the verification since kubelet skips the verification for HTTPS prober as well
-			// https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes
-			Transport: &http.Transport{
-				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
-				DialContext:     d.DialContext,
-			},
+		if prober.HTTPGet != nil {
+			localAddr := UpstreamLocalAddressIPv4
+			if config.IPv6 {
+				localAddr = UpstreamLocalAddressIPv6
+			}
+			d := &net.Dialer{
+				LocalAddr: localAddr,
+			}
+			// Construct a http client and cache it in order to reuse the connection.
+			s.appProbeClient[path] = &http.Client{
+				Timeout: time.Duration(prober.TimeoutSeconds) * time.Second,
+				// We skip the verification since kubelet skips the verification for HTTPS prober as well
+				// https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes
+				Transport: &http.Transport{
+					TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+					DialContext:     d.DialContext,
+				},
+			}
 		}
 	}
 
 	return s, nil
+}
+
+func validateAppKubeProber(path string, prober *Prober) error {
+	if !appProberPattern.Match([]byte(path)) {
+		return fmt.Errorf(`invalid key, must be in form of regex pattern %v`, appProberPattern)
+	}
+	if prober.HTTPGet == nil && prober.TCPSocket == nil {
+		return fmt.Errorf(`invalid prober type, must be of type httpGet or tcpSocket`)
+	}
+	if prober.HTTPGet != nil && prober.TCPSocket != nil {
+		return fmt.Errorf(`invalid prober, type must be either httpGet or tcpSocket`)
+	}
+	if prober.HTTPGet != nil && prober.HTTPGet.Port.Type != intstr.Int {
+		return fmt.Errorf("invalid prober config for %v, the port must be int type", path)
+	}
+	if prober.TCPSocket != nil && prober.TCPSocket.Port.Type != intstr.Int {
+		return fmt.Errorf("invalid prober config for %v, the port must be int type", path)
+	}
+	return nil
 }
 
 // FormatProberURL returns a set of HTTP URLs that pilot agent will serve to take over Kubernetes
--- a/pilot/cmd/pilot-agent/status/server.go
+++ b/pilot/cmd/pilot-agent/status/server.go
@@ -589,7 +589,7 @@ func (s *Server) handleAppProbe(w http.ResponseWriter, req *http.Request) {
 		s.handleAppProbeHTTPGet(w, req, prober, path)
 	}
 	if prober.TCPSocket != nil {
-		s.handleAppProbeTCPSocket(w, req, prober, path)
+		s.handleAppProbeTCPSocket(w, prober)
 	}
 }
 
@@ -654,7 +654,7 @@ func (s *Server) handleAppProbeHTTPGet(w http.ResponseWriter, req *http.Request,
 	w.WriteHeader(response.StatusCode)
 }
 
-func (s *Server) handleAppProbeTCPSocket(w http.ResponseWriter, _ *http.Request, prober *Prober, path string) {
+func (s *Server) handleAppProbeTCPSocket(w http.ResponseWriter, prober *Prober) {
 	port := prober.TCPSocket.Port.IntValue()
 	timeout := time.Duration(prober.TimeoutSeconds) * time.Second
 
--- a/pilot/cmd/pilot-agent/status/server.go
+++ b/pilot/cmd/pilot-agent/status/server.go
@@ -170,7 +170,7 @@ func NewServer(config Options) (*Server, error) {
 	s := &Server{
 		statusPort:            config.StatusPort,
 		ready:                 probes,
-		appProbersDestination: config.PodIP,
+		appProbersDestination: wrapIPv6(config.PodIP),
 		envoyStatsPort:        config.EnvoyPrometheusPort,
 		fetchDNS:              config.FetchDNS,
 	}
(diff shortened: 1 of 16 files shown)


c0d1ngm0nk3y:
@googlebot I consent.

howardjohn:
The high level concept seems like it could be a good idea. In the general case, it seems like an improvement for sure. My only concern is if there is some reason the users were using TCP probes instead of HTTP probes, where this would not work... I cannot really think of any. "Performance" I guess, but that doesn't seem like a real constraint here.

WDYT @ramaraochavali @incfly 

In terms of implementation, we would want to follow https://github.com/istio/istio/blob/d5fd44e6f29f62969d6ed2e893a917a7ab429adc/pilot/cmd/pilot-agent/status/server.go#L530 and https://github.com/istio/istio/blob/cf657ff502e296ab9638248456e3ad7af2715e2e/pkg/kube/inject/app_probe.go#L63

ramaraochavali:
Yeah. Idea in general makes sense to me as well.

c0d1ngm0nk3y:
Unfortunately, this part of the code is not unit tested at all. Please correct me if I am wrong. All private functions in the call stack.

I was thinking... How about refactoring `webhook.go:applyRewrite()` into an object (e.g. `PodRewriter`)?

* The code should be unit testable
* Additional functionality in that area would be easier to test
* It seems to me that some of that existing code might be worth unit testing
* I think the tests could be quite handy from a documentation point of view (i.e. this is how pods get rewritten by the webhook)

@howardjohn WDYT? Is it worth to put it into this draft to see how this would look like? Or have I missed something?

c0d1ngm0nk3y:
The code is a bit specific checking `HTTP probes` (e.g. `ShouldRewriteAppHTTPProbers`). Should we make it clearer that also `TCP probes` are affected? But there is this `annotation` `sidecar.istio.io/rewriteAppHTTPProbers`, which should only affect `HTTP probes`.

howardjohn:
TestInjection has pretty extensive unit test coverage of the injection, it does it at a higher level (input yaml -> injected yaml). IIRC it has tests for probe rewrite. I think we can extend that with TCP probe?

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -261,6 +261,7 @@ func (s *Server) Run(ctx context.Context) {
 	mux.HandleFunc(`/stats/prometheus`, s.handleStats)
 	mux.HandleFunc(quitPath, s.handleQuit)
 	mux.HandleFunc("/app-health/", s.handleAppProbe)
```
howardjohn, line 263: Do we need/want a new path? Or just add a field to `Prober` (from KubeAppProbers) to pass the TCP settings?

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -527,6 +528,23 @@ func (s *Server) handleQuit(w http.ResponseWriter, r *http.Request) {
...
 
+func (s *Server) handleTcpProbe(w http.ResponseWriter, req *http.Request) {
+	portStr := strings.TrimPrefix(req.URL.Path, "/tcp-probe/")
+
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		log.Errorf("could not parse port number when handling %s", req.URL.Path)
+	}
+
+	conn, err := net.Dial("tcp", fmt.Sprintf("localhost:%d", port))
```
howardjohn, line 539: We need to use `s.appProbersDestination`, not localhost. https://istio.io/latest/blog/2021/upcoming-networking-changes/ for a long explanation about why that is

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -527,6 +528,23 @@ func (s *Server) handleQuit(w http.ResponseWriter, r *http.Request) {
 	notifyExit()
 }
 
+func (s *Server) handleTcpProbe(w http.ResponseWriter, req *http.Request) {
+	portStr := strings.TrimPrefix(req.URL.Path, "/tcp-probe/")
+
+	port, err := strconv.Atoi(portStr)
```
howardjohn, line 534: This is not secure, as it allows connecting to arbitrary ports. We should follow the model in handleAppProbe that has a strict allow-list of legal requests

pilot/cmd/pilot-agent/status/server.go
c0d1ngm0nk3y: Thanks

c0d1ngm0nk3y:
> TestInjection has pretty extensive unit test coverage of the injection, it does it at a higher level (input yaml -> injected yaml). IIRC it has tests for probe rewrite. I think we can extend that with TCP probe?

Sounds good. We will have a look.

incfly:
Great idea and LGTM. We also have the integration test for probes, FYI, https://github.com/istio/istio/blob/4f1587eb64f9f6303f92dd4fa4771553fa0b4c01/tests/integration/security/mtls_healthcheck_test.go#L38

pilot/cmd/pilot-agent/status/server.go
incfly: +1 this still belongs to the same app probe. tcp-probe be same level with /debug can caused confusion to think it's istio built-in. or at least `app-tcp-probe`

pkg/kube/inject/app_probe.go
```diff
@@ -77,6 +86,18 @@ func convertAppProber(probe *corev1.Probe, newURL string, statusPort int) *corev
 	return p
 }
 
+// convertAppProberTCPSocket returns an overwritten `Probe` (TcpSocket) for pilot agent to take over.
+func convertAppProberTCPSocket(probe *corev1.Probe, newURL string, statusPort int) *corev1.Probe {
+	p := probe.DeepCopy()
+	// the sidecar intercepts all tcp connections, so we change it to a special HTTP probe
+	p.HTTPGet = &corev1.HTTPGetAction{}
+	p.HTTPGet.Port = intstr.FromInt(statusPort)
+	// the path contains the original port of the application
```
incfly, line 95: same comments as john. we should not rely on client indicating the port, instead the original tcp port value can be marshaled to json env var and pilot agent only allows that port.

phil9909:
In the current state of the PR the TCP probe translation can be enabled and disabled via the [`sidecar.istio.io/rewriteAppHTTPProbers`](https://github.com/istio/api/blob/b04257312323f5a80b192b99cb08c8b20dd5cc39/annotation/annotations.yaml#L83-L90) annotation. This is sub-optimal as it has "HTTP" in it's name. We see (at least) two alternatives here:
1. deprecate `sidecar.istio.io/rewriteAppHTTPProbers` and replace it with a `sidecar.istio.io/rewriteAppProbers` annotation, which will configure both HTTP and TCP rewrite. During the deprecation period `rewriteAppHTTPProbers` would probably also configure the TCP behavior.
1. add a separate `sidecar.istio.io/rewriteAppTCPProbers` annotation. 

@c0d1ngm0nk3y and I are in favor of option 1, as we don't think there is a big use-case for having two separate settings.
@incfly @howardjohn WDYT?

howardjohn:
I would just have rewriteAppHTTPProbers apply to both, and not rename/deprecate. The annotation is extremely niche - I would almost go as far as to say we should remove it. 

if you read it like "rewrite app [probers, of any type] to HTTP probers [pointing to istio-agent]" it makes sense :slightly_smiling_face:

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -601,6 +615,19 @@ func (s *Server) handleAppProbe(w http.ResponseWriter, req *http.Request) {
 	w.WriteHeader(response.StatusCode)
 }
 
+func (s *Server) handleAppProbeTCPSocket(w http.ResponseWriter, req *http.Request, prober *Prober, path string) {
+	port := prober.TCPSocket.Port.IntValue()
+	timeout := time.Duration(prober.TimeoutSeconds) * time.Second
+
+	conn, err := net.DialTimeout("tcp", fmt.Sprintf("%s:%d", s.appProbersDestination, port), timeout)
+	if err != nil {
+		w.WriteHeader(404)
```
howardjohn, line 624:
```suggestion
		w.WriteHeader(http.StatusInternalServerError)
```

Seems like it may be more appropriate? 404 seems like it would suggest we hit some URL that wasn't configured, rather than the health check failed

howardjohn (commented):
/test all

phil9909:
@googlebot I consent.

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -211,37 +212,53 @@ func NewServer(config Options) (*Server, error) {
...
 
+func validateAppKubeProber(path string, prober *Prober) error {
+	if !appProberPattern.Match([]byte(path)) {
+		return fmt.Errorf(`invalid key, must be in form of regex pattern %v`, appProberPattern)
+	}
+	if prober.HTTPGet == nil && prober.TCPSocket == nil {
+		return fmt.Errorf(`invalid prober type, must be of type httpGet or tcpSocket`)
+	}
+	if prober.HTTPGet != nil && prober.TCPSocket != nil {
+		return fmt.Errorf(`invalid prober, type must be either httpGet or tcpSocket`)
```
howardjohn, line 251: dupe of L247

howardjohn (commented):
This looks great!

pilot/cmd/pilot-agent/status/server.go
c0d1ngm0nk3y: You mean L250 and L247? One checks if none is set, the other one if both are set.

howardjohn: Ah, ok thanks

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -211,37 +212,53 @@ func NewServer(config Options) (*Server, error) {
...
+			}
 		}
 	}
 
 	return s, nil
 }
 
+func validateAppKubeProber(path string, prober *Prober) error {
+	if !appProberPattern.Match([]byte(path)) {
+		return fmt.Errorf(`invalid key, must be in form of regex pattern %v`, appProberPattern)
```
incfly, line 245: key? Or do you mean "path"?

c0d1ngm0nk3y: We just moved this existing check, but I guess you are right. `path` sounds correct.

howardjohn:
@c0d1ngm0nk3y is this ready to go? If so, you can move it out of "draft" mode

c0d1ngm0nk3y:
> @c0d1ngm0nk3y is this ready to go? If so, you can move it out of "draft" mode

The integration test is still failing for some reason. Once we figured that out, we will remove the "draft" from this PR.

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -601,6 +627,24 @@ func (s *Server) handleAppProbe(w http.ResponseWriter, req *http.Request) {
...
+func (s *Server) handleAppProbeTCPSocket(w http.ResponseWriter, prober *Prober) {
+	port := prober.TCPSocket.Port.IntValue()
+	timeout := time.Duration(prober.TimeoutSeconds) * time.Second
+
+	// FIXME(phil9909): Currently, the IP table rules still seem to redirect traffic from inside the pod into the envoy
+	// when connection using the external IP of the pod. Therefore we hardcode it to localhost for now.
+	// See https://istio.io/latest/blog/2021/upcoming-networking-changes/
+	// addr := fmt.Sprintf("%s:%d", s.appProbersDestination, port)
+	addr := fmt.Sprintf("localhost:%d", port)
+	conn, err := net.DialTimeout("tcp", addr, timeout)
```
phil9909, line 639: FYI @howardjohn

howardjohn: When you hit this issue, were you using a custom net.Dialer with source address set? see https://github.com/istio/istio/blob/2d99cd2817b2a939c85806020bd0acc29b261d5e/pilot/cmd/pilot-agent/status/server.go#L228

If not - you should use that and it should work. That is why HTTP works; the iptables rules should apply the same for TCP and HTTP traffic

phil9909: Thanks, I updated the code accordingly. It's working now.

I additionally had to fix the unit tests after the change, as `::6` is not a real loopback address.
This revealed that there are no unit tests for the IPv6 based HTTP probes. I can add some in a follow up PR.

releasenotes/notes/33734.yaml
```diff
@@ -0,0 +1,19 @@
...
+releaseNotes:
+- |
+  '**Improved** TCP probes now working as expected: When using TCP probes with older versions of istio the check was always successful, even if the application didn't open the port.'
+
+upgradeNotes:
+  - title: TCP probes now working as expected
+    content: |
+        When using TCP probes with older versions of istio the check was always successful, even if the application didn't open the port.
+        This may cause problems when upgrading: If you had a missconfiguration in a TCP probe (e.g wrong port) you maybe haven't noticed.
+        After the upgrade a missconfigured TCP probe will fail and therefore might cause downtimes.
```
howardjohn, line 16: Can we add a feature flag to opt-out of this in https://github.com/istio/istio/blob/dbf6a9e1c123bf6abbab45683b1c6c1f7660916c/pilot/pkg/features/pilot.go#L28-L27? We can make it default on, but this allows users that may be broken by this to have more time to fix their applications.

howardjohn (commented):
LGTM!

@incfly ptal

releasenotes/notes/33734.yaml
c0d1ngm0nk3y: > Can we add a feature flag to opt-out of this?

Is it worth it? It was very strange if it worked before. IMHO, I think you even could call it `bug` that it worked before even if the port is not available. On the other hand, if we put it into the releasenotes like this, it might be good to have this option.

c0d1ngm0nk3y: @howardjohn The feature flag should be documented somewhere, right? Can you maybe point us to an example?

howardjohn: Its automatically documented when you use RegisterEnvVar through some automation we have.

I totally agree with what you said, but at the same time we get users with massive deployments where even bug fixes break them because they relied on the bugs... a bit annoying, but a feature flag is pretty easy to add an makes everyone happy.

howardjohn (approved):
Fantastic PR :+1:  thanks for all your work here

pantelis-karamolegkos:
We are trying to use tcp-based health and ready probes.

Istio seems to be listening to **all** ports.

This PR [implies](https://github.com/istio/istio/pull/33734#issue-933761559) that this functionality is not available.

> TCP probes do currently not work with istio sidecar, but there appears to be a need (#15178 #28737).

Doesn't this contradict [this](https://istio.io/latest/docs/tasks/security/authorization/authz-tcp/) page however?

PR #33739, opened 2021-06-30
Add all current fuzzers to fuzz regression test

howardjohn:
This ensures they pass at least a few simple inputs, and as we hit bugs
we can add regression test cases
@AdamKorcz


AdamKorcz:
Great! Will add fuzzers to the test as I upload more.

ericvn (approved):

PR #33731, opened 2021-06-30
Fuzzing: Run FuzzInitContext on OSS-fuzz

AdamKorcz:
Enables `FuzzInitContext` to run continuously by OSS-fuzz.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[x] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[x] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #33741, opened 2021-06-30
Fuzzing: Add analyzer fuzzer

AdamKorcz:
Adds a fuzzer for the analyzer. The fuzzer creates fuzzed config files and passes them to a `SourceAnalyzer` with which `Analysis` is run.

If https://github.com/istio/istio/pull/33739 is merged first, then this PR should hold until this fuzzer is added to `tests/fuzz/regression_test.go`.

Will also need a rebase if https://github.com/istio/istio/pull/33731 is merged first.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/fuzz/analyzer_fuzzer.go
+++ b/tests/fuzz/analyzer_fuzzer.go
@@ -1,0 +1,174 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// nolint: golint
+package fuzz
+
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"time"
+
+	fuzz "github.com/AdaLogics/go-fuzz-headers"
+
+	"istio.io/istio/galley/pkg/config/analysis"
+	"istio.io/istio/galley/pkg/config/analysis/analyzers"
+	"istio.io/istio/galley/pkg/config/analysis/local"
+	"istio.io/istio/galley/pkg/config/scope"
+	"istio.io/istio/pkg/config/schema"
+	"istio.io/istio/pkg/config/schema/collection"
+	"istio.io/pkg/log"
+)
+
+var availableAnalyzers = analyzers.All()
+
+// createRandomConfigFile creates a single fuzzed config file
+func createRandomConfigFile(f *fuzz.ConsumeFuzzer) (string, error) {
+	data, err := f.GetBytes()
+	if err != nil {
+		return "nobytes", err
+	}
+	tmpfile, err := ioutil.TempFile("", "example")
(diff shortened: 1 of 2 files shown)


tests/fuzz/analyzer_fuzzer.go
```diff
@@ -0,0 +1,174 @@
...
+		configFile, err := createRandomConfigFile(f)
+		if err != nil {
+			return files, err
+		}
+		defer os.Remove(configFile)
+		of, err := os.Open(configFile)
+		if err != nil {
+			return files, err
+		}
+		files = append(files, local.ReaderSource{Name: configFile, Reader: of})
```
howardjohn, line 77: Should we use strings.NewReader instead of a real file?

AdamKorcz: Yes, that should have a positive impact on performance. Updated fuzzer with `bytes.NewReader`

tests/fuzz/analyzer_fuzzer.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	configFiles, err := createRandomConfigFiles(f)
+	if err != nil {
+		return 0
+	}
+
+	sa := local.NewSourceAnalyzer(schema.MustGet(), analysis.Combine("testCase", analyzer), "", "istio-system", cr, true, 10*time.Second)
+	if addMeshConfig {
+		err = sa.AddFileKubeMeshConfig(meshConfigFile)
```
howardjohn, line 149: I would be fine with adding a function to directly set from bytes or `MeshConfig` type if file is too slow

tests/fuzz/oss_fuzz_build.sh
```diff
@@ -19,3 +19,8 @@ compile_go_fuzzer istio.io/istio/tests/fuzz FuzzParseAndBuildSchema fuzz_parse_a
 compile_go_fuzzer istio.io/istio/tests/fuzz FuzzConfigValidation fuzz_config_validation
 compile_go_fuzzer istio.io/istio/tests/fuzz FuzzParseMeshNetworks fuzz_parse_mesh_networks
 compile_go_fuzzer istio.io/istio/tests/fuzz FuzzValidateMeshConfig fuzz_validate_mesh_config
+compile_go_fuzzer istio.io/istio/tests/fuzz FuzzAnalyzer fuzz_analyzer
+
+
+# Create seed corpora:
+zip "${OUT}"/fuzz_analyzer_seed_corpus.zip "${SRC}"/istio/galley/pkg/config/analysis/analyzers/testdata/*.yaml
```
howardjohn, line 26: nice! should it be `${SRC}"/galley/...` though? I would have thought SRC=gopath/src/istio.io/istio, but maybe not?

howardjohn (approved):

tests/fuzz/oss_fuzz_build.sh
AdamKorcz: istio's path in the OSS-fuzz environment is $SRC/istio: https://github.com/google/oss-fuzz/blob/master/projects/istio/Dockerfile#L20

tests/fuzz/analyzer_fuzzer.go
AdamKorcz: This fuzzer is not particularly fast, but the bottleneck seems to be `NewSourceAnalyzer`.

Setting from bytes would give a boost to execution speed, but it would still be below 100 execs/s.

PR #33742, opened 2021-06-30
include multi-network eds logic in benchmark

stevenctl:
We don't currently measure this – we need to set gateways/networks to find out the impact of #33715


adiprerepa (approved):
lgtm

nmittler (approved):

howardjohn (approved):

Issue #33672, opened 2021-06-28
Endpointslice races lead to bad state


howardjohn:
CI failure example: https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-119_istio_postsubmit/1409422406954520576

Logs with some custom lines added:
```
2021-06-28T17:56:18.868986Z     info    kube    Handle event delete for service consistent-hash in namespace echo
2021-06-28T17:56:18.900750Z     info    kube    Handle event add for service consistent-hash in namespace echo
2021-06-28T17:56:18.900906Z     error   kube    howardjohn: set es cache consistent-hash.echo.svc.cluster.local/consistent-hash-nr49f -> 3
2021-06-28T17:56:18.900978Z     info    ads     Full push, new service echo/consistent-hash.echo.svc.cluster.local
2021-06-28T17:56:18.901028Z     info    ads     Full push, service accounts changed, consistent-hash.echo.svc.cluster.local
2021-06-28T17:56:18.902185Z     error   kube    howardjohn: endpoint consistent-hash: add
2021-06-28T17:56:18.902463Z     error   kube    howardjohn: set es cache consistent-hash.echo.svc.cluster.local/consistent-hash-jprc6 -> 3
2021-06-28T17:56:18.906168Z     error   kube    howardjohn: endpoint consistent-hash: delete
2021-06-28T17:56:18.906358Z     error   kube    howardjohn: delete es cache consistent-hash.echo.svc.cluster.local/consistent-hash-nr49f
2021-06-28T17:56:18.906415Z     info    ads     Incremental push, service consistent-hash.echo.svc.cluster.local has no endpoints
```

Basically what is happening in the test is we created and delete a service a lot, with the same name. We process the endpointslice updates incorrectly and end up clearing the cache. Events are `update nr49f; update jprc6 (at this point we have duplicates!! code should dedupe); delete nr49f (bug: we delete all the endpoints, even though jprc6 still has them`

Possible related to https://github.com/istio/istio/issues/27946
Blocker for https://github.com/istio/istio/issues/27472

cc @jacob-delgado

adiprerepa:
Previous dedup PR: https://github.com/istio/istio/pull/27603/files

adiprerepa:
dedup logic we should follow (`kube-proxy`) https://github.com/kubernetes/kubernetes/blob/aa0017ad138ac88db5c4c4531749d18d4ed31879/pkg/proxy/endpointslicecache.go

[...]

PR #33750, opened 2021-06-30
fix endpointslice deletion bug

adiprerepa:
fixes https://github.com/istio/istio/issues/33672

this fix makes sure that the cache is not wiped when a single delete is called.

cc @howardjohn

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/endpoints.go
+++ b/pilot/pkg/serviceregistry/kube/controller/endpoints.go
@@ -189,7 +189,7 @@ func (e *endpointsController) forgetEndpoint(endpoint interface{}) []*model.Isti
 			e.c.pods.endpointDeleted(key, ea.IP)
 		}
 	}
-	return make([]*model.IstioEndpoint, 0)
+	return nil
 }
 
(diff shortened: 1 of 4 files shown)


pilot/pkg/serviceregistry/kube/controller/endpoints.go
```diff
@@ -181,14 +181,15 @@ func (e *endpointsController) onEvent(curr interface{}, event model.Event) error
...
-func (e *endpointsController) forgetEndpoint(endpoint interface{}) {
+func (e *endpointsController) forgetEndpoint(endpoint interface{}) []*model.IstioEndpoint {
 	ep := endpoint.(*v1.Endpoints)
 	key := kube.KeyFunc(ep.Name, ep.Namespace)
 	for _, ss := range ep.Subsets {
 		for _, ea := range ss.Addresses {
 			e.c.pods.endpointDeleted(key, ea.IP)
 		}
 	}
+	return make([]*model.IstioEndpoint, 0)
```
howardjohn, line 192: very nit: `return nil` will be equivalent here but slightly more efficient

howardjohn (approved):

PR #28800, opened 2020-11-11
Refactor injection internals

howardjohn:
This implements the first two parts of https://docs.google.com/document/d/1Rmp9B3DDypgMCau-YAqidx_r3qvKLZj6jUX-t22zj84/. Please see "Refactor injector internals" and "Make sidecar injection idempotent" of that doc for some additional details

Fixes https://github.com/istio/istio/issues/25931

This deeply modifies the logic we use for injection. Rather than hand crafting a json patch, which is tedious and not flexible, we just do a merge of the template with the original pod, then directly modify the `corev1.Pod`. This has extremely minimal user impact; all existing templates should continue to work, but some additional options may now be available.

This can be seen by the fact the InjectTest has no changes and still passes. It should be noted the Webhook test DOES have a lot of changes - this is because the jsonpatch library we use is generating the patches, and it does so differently than we used to. However, the end result to the user is the same.

This change is quite large. Sorry. I think the new code is far simpler and more maintainable though.

PR #33752, opened 2021-06-30
istio: re-invoke sidecar injector IfNeeded

weibohe:
#28800 made the sidecar injector webhook idempotent. Internally we have another mutating webhook that updates the istio rev label to control the rollout of new istio control plane versions. Since sidecar injector webhook is idempotent, changing the reinvocationPolicy to IfNeeded so the sidecar injector can be invoked if related labels are updated.

[ ] Docs
[x] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

[ ] Does not have any user-facing changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

@howardjohn

### DIFF
--- a/releasenotes/notes/reinvoke-injector-ifneeded.yaml
+++ b/releasenotes/notes/reinvoke-injector-ifneeded.yaml
@@ -1,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: feature
+area: installation
+issue:
+releaseNotes:
+  - |
+    **Updated** the sidecar injector webhook reinvocationPolicy to IfNeeded.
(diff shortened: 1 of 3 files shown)


howardjohn (commented):
As discussed offline, I *think* this is fine. I believe @ramaraochavali looked into this, so maybe can provide some insight

releasenotes/notes/reinvoke-injector-ifneeded.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: feature
+area: installation
+issue:
+releaseNotes:
+  - |
+    **Updated** the sidecar injector webhook reinvocationPolicy to IfNeeded.
```
shamsher31, line 7:
```suggestion
    **Added** `reinvocationPolicy` for the sidecar injector webhook.
```

ramaraochavali:
@weibohe We found this is not fully helping if there is another webhook `reinvocationPolicy` set to `IfNeeded`. I think k8s would not try after two attempts or some thing  like that was the reason- I could not find docs now.

PR #33756, opened 2021-06-30
implement Istio APIs in xDS for gRPC

stevenctl:
- [X] Code organization touchups

* CDS
- [X] Apply ISTIO_MUTUAL tls (by default, or by dr)
- [x] Apply compatible loadbalancing policies
- [x] Generate subset clusters (need work in routes gen to be useful)
- [ ] Warn the user somehow to indicate ignored parts of DR config (gRPC might be somewhat forgiving if we specify it, but when their implementation comes along it may NACK what we send)

LDS

* Nothing much should change, change the names/route names maybe

RDS:

- [X] Re-use VirtualHost gen to get VirtualService features

Testing:
- [X] Run echo app in unit tests against a FakeDiscoveryServer
  - [ ] at least 2 or 3 VS features tested
  - [ ] at least 1 DR 
- [ ] figure out a way to have many "local" addresses that works on OSX


pilot/pkg/networking/grpcgen/cds.go
```diff
@@ -0,0 +1,216 @@
...
+		push:           push,
+		hostname:       hostname,
+		portNum:        portNum,
+		defaultCluster: defaultCluster,
+		svc:            svc,
+		port:           port,
+	}, nil
+}
+
+func (b *clusterBuilder) build() []*clusterv3.Cluster {
```
stevenctl, line 106: Most of the logic is similar, I think we can just share the code if we expose:
* BuildSubsetCluster
* BuildDefaultCluster
* BuildClusterOpts and members
* Add something to allow us to override how tlscontext is built

mandarjog:
@stevenctl is there a design doc associated with this change?

stevenctl:
@mandarjog https://github.com/istio/istio/pull/33711 was an attempt at re-use. Listeners, and part of clusters, seem to be somewhat strict for now on the gRPC side. 

Eventually we will be able to re-use without too much branching.

Consider this a prototype, still figuring out what works and how their client behaves.

Design doc incoming.

pilot/pkg/networking/grpcgen/cds.go
adiprerepa: i think we should do this before merge

pilot/pkg/networking/grpcgen/cds.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	// try to resolve the service and port
+	var port *model.Port
+	svc := push.ServiceForHostname(node, hostname)
+	if svc == nil {
+		log.Warnf("cds gen for %s: did not find service for cluster %s", node.ID, defaultClusterName)
+	} else {
+		var ok bool
+		port, ok = svc.Ports.GetByPort(portNum)
```
adiprerepa, line 107: imo this is cleaner
```suggestion
		port, ok := svc.Ports.GetByPort(portNum)
```

pilot/pkg/networking/grpcgen/cds.go
```diff
@@ -0,0 +1,270 @@
...
+
+// applyPolicy mutates the give cluster (if not-nil) so that the given merged traffic policy applies.
+func (b *clusterBuilder) applyPolicy(c *clusterv3.Cluster, trafficPolicy *networking.TrafficPolicy) {
+	// cluster can be nil if it wasn't requested
+	if c == nil || trafficPolicy == nil {
+		return
+	}
+	b.applyTLS(c, trafficPolicy)
+	b.applyLoadBalancing(c, trafficPolicy)
+	// TODO status or log when unsupported features are included
```
adiprerepa, line 206: what do you mean by unsupported features?

pilot/pkg/networking/grpcgen/cds.go
```diff
@@ -0,0 +1,270 @@
...
+
+func (b *clusterBuilder) applyLoadBalancing(c *clusterv3.Cluster, policy *networking.TrafficPolicy) {
+	switch policy.LoadBalancer.GetSimple() {
+	case networking.LoadBalancerSettings_ROUND_ROBIN:
+	// ok
+	default:
+		log.Warnf("cannot apply LbPolicy %s to %s", policy.LoadBalancer.GetSimple(), b.node.ID)
+	}
+
+	// TODO https://github.com/grpc/proposal/blob/master/A42-xds-ring-hash-lb-policy.md
```
adiprerepa, line 217: so once ring hash load balancing gets implemented in go, we will allow it to be applied?

pilot/pkg/networking/grpcgen/cds.go
```diff
@@ -0,0 +1,270 @@
...
+
+}
+
+// TransportSocket proto message has a `name` field which is expected to be set to exactly this value by the
+// management server (see grpc/xds/internal/client/xds.go securityConfigFromCluster).
+const transportSocketName = "envoy.transport_sockets.tls"
+
+// buildTLSContext creates a TLS context that assumes 'default' name, and credentials/tls/certprovider/pemfile
+// (see grpc/xds/internal/client/xds.go securityConfigFromCluster).
+func buildTLSContext(sans []string) *tlsv3.UpstreamTlsContext {
```
adiprerepa, line 250: seems generic enough that maybe name should be passed in instead of set to a default value

pilot/pkg/networking/grpcgen/grpcecho_test.go
```diff
@@ -0,0 +1,190 @@
...
+//      namespace: {server.namespace or "default"}
+//      labels: {"app": "grpc", "version": "{server.version}"}
+//    spec:
+//      address: {grpcEchoHost}
+//      ports:
+//        grpc: {generated portnum}
+func newConfigGenTest(t *testing.T, discoveryOpts xds.FakeOptions, servers ...echoCfg) *configGenTest {
+	cgt := &configGenTest{T: t}
+	for i, s := range servers {
+		// TODO this breaks without extra ifonfig aliases on OSX, and probably elsewhere
```
adiprerepa, line 53: why does it break?

pilot/pkg/networking/grpcgen/grpcecho_test.go
```diff
@@ -0,0 +1,190 @@
...
+	if err := expectAlmost(distribution["v1"], 20); err != nil {
+		t.Fatal(err)
+	}
+	if err := expectAlmost(distribution["v2"], 80); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func expectAlmost(got, want int) error {
+	if math.Abs(float64(want-got)) > 10 {
```
adiprerepa, line 186: is this margin of error standard or arbitrary? Its probably fine but im curious

pilot/pkg/networking/grpcgen/grpcecho_test.go
```diff
@@ -0,0 +1,190 @@
...
+		return fmt.Errorf("expected ~%d but got %d", want, got)
```
adiprerepa, line 187:
```suggestion
		return fmt.Errorf("expected within 10 of %d but got %d", want, got)
```

pilot/pkg/networking/grpcgen/grpcgen_test.go
```diff
@@ -96,14 +102,28 @@ func TestGRPC(t *testing.T) {
 		},
 	})
 
+	store.Create(config.Config{
```
adiprerepa, line 105: confused -- why was this not required before but is now?

pilot/pkg/networking/grpcgen/lds.go
```diff
@@ -0,0 +1,99 @@
...
+
+	for _, el := range node.SidecarScope.EgressListeners {
+		for _, sv := range el.Services() {
+			shost := string(sv.Hostname)
+			if len(filter) > 0 {
+				// DiscReq has a filter - only return services that match
+				if !filter[shost] {
+					continue
+				}
+			}
```
adiprerepa, line 56:
```suggestion
		        // DiscReq has a filter - only return services that match
			if !filter[shost] {
				continue
			}
```
i dont think the outer if is needed

pilot/pkg/networking/grpcgen/lds.go
```diff
@@ -0,0 +1,99 @@
...
+			for _, p := range sv.Ports {
+				hp := net.JoinHostPort(shost, strconv.Itoa(p.Port))
+				ll := &envoy_config_listener_v3.Listener{
+					Name: hp,
+				}
+
+				ll.Address = &envoy_config_core_v3.Address{
```
adiprerepa, line 63: probably not in the scope of this PR, but do we not have our own constructors for envoy resources? I feel like it would save a lot of time

adiprerepa (commented):
really awesome
I think delta will take somewhat of a similar approach architecturally

pilot/pkg/networking/grpcgen/cds.go
stevenctl: The `InstanceName`? This is tightly coupled for now with the CertProvider instance that we set up here:

https://github.com/istio/istio/blob/876407b38ce3fa8221d44de66f16a15d38fba799/pkg/istio-agent/grpcxds/grpc_bootstrap.go#L138


I can add some comments to make this clear.

pilot/pkg/networking/grpcgen/cds.go
stevenctl: gRPC does support it now, I just have a TODO to actually apply it in our CDS

pilot/pkg/networking/grpcgen/cds.go
stevenctl: gRPC will NACK certain xDS configs we throw at it – this is why we have a separate generator. 

For example, clusters can't really set any lb policy except for round robin: 

https://github.com/grpc/grpc-go/blob/30dfb4b933a50fd366d7ed36ed4f71dbba2d382e/xds/internal/xdsclient/xds.go#L632

pilot/pkg/networking/grpcgen/cds.go
stevenctl: `port` needs to be in the outer scope, so I could do 

```
svcPort, ok := ...
...
port = svcPort
```

if that feels cleaner?

pilot/pkg/networking/grpcgen/lds.go
stevenctl: The semantics are different – if the filter has a len of 0, that means we don't want to do filtering at all, so we _shouldn't_ skip.

pilot/pkg/networking/grpcgen/lds.go
stevenctl: They may be shaped many ways, I think creating a "constructor" would have to be some builder-pattern style thing like `ListenerBuilder().WithAddress(AddressBuilder().WithSocketAddress(addr, port) `etc etc. Would be cool, but probably not needed.

pilot/pkg/networking/grpcgen/grpcgen_test.go
stevenctl: This will make sure we build TLS context in CDS, now the test covers whether gRPC NACKs that config.

pilot/pkg/networking/grpcgen/cds.go
adiprerepa: nevermind, what I was trying to do isnt possible. Your way is good

pilot/pkg/networking/grpcgen/cds.go
adiprerepa: yeah maybe thats a good idea

pilot/pkg/networking/grpcgen/lds.go
adiprerepa: java builder ptsd

but yes that might be better

pilot/pkg/networking/grpcgen/lds.go
adiprerepa: ahh i see. thanks for clarifying

costinm (approved):
Seems to be safe for the existing code, and gRPC has coverage. We can continue to iterate.

stevenctl:
Follow-ups:

- Doc/blogpost describing how to try out the gRPC support
- Design doc describing what's been done, and the long-term vision for gRPC support

pilot/pkg/networking/grpcgen/cds.go
```diff
@@ -0,0 +1,270 @@
...
+			Name:     c.Name,
+			Resource: util.MessageToAny(c),
+		})
+	}
+	return resp
+}
+
+// newClusterFilter builds a filtering map to determine which clusters need to be built.
+// gRPC will usually request each subset individually, regardless of if a previous response included it.
+func newClusterFilter(names []string) map[string]sets.Set {
```
howardjohn, line 59: nit: Can you comment that this is basically a map of "resource name without subset" -> "resource names with subsets"?

pilot/pkg/networking/grpcgen/cds.go
```diff
@@ -0,0 +1,270 @@
...
+			b.applyPolicy(c, trafficPolicy)
+			subsetClusters = append(subsetClusters, c)
+		}
+	}
+
+	return
+}
+
+// applyPolicy mutates the give cluster (if not-nil) so that the given merged traffic policy applies.
+func (b *clusterBuilder) applyPolicy(c *cluster.Cluster, trafficPolicy *networking.TrafficPolicy) {
```
hzxuzhonghu, line 200: nit: applyTrafficPolicy

pilot/pkg/networking/grpcgen/cds.go
```diff
@@ -0,0 +1,270 @@
...
+	"istio.io/istio/pilot/pkg/model"
+	"istio.io/istio/pilot/pkg/networking/core/v1alpha3"
+	"istio.io/istio/pilot/pkg/networking/util"
+	"istio.io/istio/pilot/pkg/util/sets"
+	"istio.io/istio/pkg/config/host"
+)
+
+// BuildClusters handles a gRPC CDS request, used with the 'ApiListener' style of requests.
+// The main difference is that the request includes Resources to filter.
+func (g *GrpcConfigGenerator) BuildClusters(node *model.Proxy, push *model.PushContext, names []string) model.Resources {
```
hzxuzhonghu, line 35: Does grpc request for cds with names?

pilot/pkg/networking/grpcgen/cds.go
```diff
@@ -0,0 +1,270 @@
...
+		resp = append(resp, &discovery.Resource{
+			Name:     c.Name,
+			Resource: util.MessageToAny(c),
+		})
+	}
+	return resp
+}
+
+// newClusterFilter builds a filtering map to determine which clusters need to be built.
+// gRPC will usually request each subset individually, regardless of if a previous response included it.
```
hzxuzhonghu, line 58: I could not understand how the filter work

pilot/pkg/networking/grpcgen/cds.go
stevenctl: Yes. It will ask for exact names, usually a request per route (ex, a route that shifts between 3 clusters will only ask for those 3 clusters at a time, other things will be requested separately)

pilot/pkg/networking/grpcgen/cds.go
stevenctl: I updated the comment, hope that helps?

pilot/pkg/networking/grpcgen/lds.go
```diff
@@ -0,0 +1,99 @@
...
+	"istio.io/istio/pilot/pkg/networking/util"
+)
+
+// BuildListeners handles a LDS request, returning listeners of ApiListener type.
+// The request may include a list of resource names, using the full_hostname[:port] format to select only
+// specific services.
+func (g *GrpcConfigGenerator) BuildListeners(node *model.Proxy, push *model.PushContext, names []string) model.Resources {
+	resp := model.Resources{}
+
+	filter := map[string]bool{}
```
howardjohn, line 37: what is the key of this? Also may prefer `map[string]struct{}{}` since `map[string]bool` can be ambiguous (a field is not set vs set to false - what does it mean?)

pilot/pkg/networking/grpcgen/lds.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, name := range names {
+		if strings.Contains(name, ":") {
+			n, _, err := net.SplitHostPort(name)
+			if err == nil {
+				name = n
+			}
+		}
+		filter[name] = true
```
howardjohn, line 45: Why is port not part of the key? Seems like we return ports they didn't ask for?

pilot/pkg/networking/grpcgen/lds.go
```diff
@@ -0,0 +1,99 @@
...
+		for _, sv := range el.Services() {
+			shost := string(sv.Hostname)
+			if len(filter) > 0 {
+				// DiscReq has a filter - only return services that match
+				if !filter[shost] {
+					continue
+				}
+			}
+			for _, p := range sv.Ports {
+				hp := net.JoinHostPort(shost, strconv.Itoa(p.Port))
```
howardjohn, line 58: so if they request `foo.com`, we return a listener for `foo.com:80`? It seems like that would not work?

pilot/pkg/networking/grpcgen/lds.go
```diff
@@ -0,0 +1,99 @@
...
+				ll := &listener.Listener{
```
howardjohn, line 59: very nit, but it seems more readable to construct everything together rather than a bunch of `ll.xyz = ...`

pilot/pkg/networking/grpcgen/grpcecho_test.go
howardjohn: is it plausible to skip if it is? Would probably not be ideal to break tests for people

howardjohn (commented):
LGTM overall

pilot/pkg/networking/grpcgen/lds.go
stevenctl: Yeah this can be cleaned up – lds is 99% copypaste from the original code into a new file

howardjohn (approved):

PR #33757, opened 2021-06-30
Fix endpointslice duplicate IPs on service update

howardjohn:
We should be reading from the cache which is more efficient and handles
deduplication properly, rather than recomputing.

The PR is basically a few lines of a code and a lot of refactoring to
get the test to work. The problem was the XdsUpdater isn't actually set
in FakeXdsServer; in the TestSameIPEndpointSlicing test which this is
based on the Wait() always times out (so its not actually waiting)
because we use a xdsUpdater that isn't actually wired up to the
controller.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/serviceregistry/serviceregistry_test.go
+++ b/pilot/pkg/serviceregistry/serviceregistry_test.go
@@ -43,87 +43,27 @@ import (
 	"istio.io/istio/pilot/pkg/xds"
 	"istio.io/istio/pilot/test/util"
 	"istio.io/istio/pilot/test/xdstest"
-	"istio.io/istio/pkg/cluster"
 	"istio.io/istio/pkg/config"
 	"istio.io/istio/pkg/config/host"
 	"istio.io/istio/pkg/config/mesh"
 	"istio.io/istio/pkg/config/schema/collections"
 	"istio.io/istio/pkg/config/schema/gvk"
 	kubeclient "istio.io/istio/pkg/kube"
-	"istio.io/istio/pkg/test"
 	"istio.io/istio/pkg/test/util/retry"
 )
-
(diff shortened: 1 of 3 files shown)


pilot/pkg/serviceregistry/serviceregistry_test.go
```diff
@@ -961,9 +901,66 @@ func TestEndpointsDeduping(t *testing.T) {
...
+			}, {
+				Name: "http-other",
+				Port: 90,
+			}},
+			Selector:  labels,
+			ClusterIP: "9.9.9.9",
+		},
+	})
+	xdsUpdater.WaitOrFail(t, "svcupdate")
+	expectEndpoints(t, s, "outbound|80||service.namespace.svc.cluster.local", []string{"1.2.3.4:80"})
```
adiprerepa, line 957: How does this verify that there are no duplicates? We expect the same thing on line 934 right?

adiprerepa (commented):
basically lgtm

pilot/pkg/serviceregistry/serviceregistry_test.go
howardjohn: if you revert the few lines of the code change I made in endpointslice.go, this would actually fail (with the IP duplicated).

So we are ensuring that even after the service changes, we still only have one IP

adiprerepa: makes sense

adiprerepa (approved):
:tada:

Issue #33760, opened 2021-07-01
BUG: VirtualService match no service will be honored to route 80


hzxuzhonghu:
Create a vs as below:

```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: test
spec:
  hosts:
  - abc    # match no service
  http:
  - name: "test"
    route:
    - destination:
        host: xxx    # no service
```

The route 80 will be like

```
[
    {
        "name": "80",
        "virtualHosts": [
            {
                "name": "abc.default.svc.cluster.local:80",
                "domains": [
                    "abc.default.svc.cluster.local",
                    "abc.default.svc.cluster.local:80"
                ],
                "routes": [
                    {
                        "name": "test",
                        "match": {
                            "prefix": "/wpcatalog",
                            "caseSensitive": true
                        },
                        "route": {
                            "cluster": "outbound|80||xxx.default.svc.cluster.local",
                            "timeout": "0s",
                            "retryPolicy": {
                                "retryOn": "connect-failure,refused-stream,unavailable,cancelled,retriable-status-codes",
                                "numRetries": 2,
                                "retryHostPredicate": [
                                    {
                                        "name": "envoy.retry_host_predicates.previous_hosts"
                                    }
                                ],
                                "hostSelectionRetryMaxAttempts": "5",
                                "retriableStatusCodes": [
                                    503
                                ]
                            },
                            "maxGrpcTimeout": "0s"
                        },
                        "metadata": {
                            "filterMetadata": {
                                "istio": {
                                    "config": "/apis/networking.istio.io/v1alpha3/namespaces/default/virtual-service/test"
                                }
                            }
                        },
                        "decorator": {
                            "operation": "xxx.default.svc.cluster.local:80/wpcatalog*"
                        }
                    }
                ],
                "includeRequestAttemptCount": true
            },

```

zhlsunshine:
@hzxuzhonghu What should be the expected behavior?  Such as verify the destination host. then output the error message and ignore this vs if there is no matched service?

hzxuzhonghu:
I have a fix https://github.com/istio/istio/pull/33764
 IMO if vs.hosts matches no service, it should not be respected at all. But seems the integration test depends on this.

PR #33764, opened 2021-07-01
Added selecting virtualservice

hzxuzhonghu:

Only handle virtualServices related to this route, instead of all

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/httproute.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute.go
@@ -182,6 +182,7 @@ func (configgen *ConfigGeneratorImpl) buildSidecarOutboundHTTPRouteConfig(node *
 }
 
 // TODO: merge with IstioEgressListenerWrapper.selectVirtualServices
+// selectVirtualServices selects the virtual services by matching given services' host names.
 func selectVirtualServices(virtualServices []config.Config, servicesByName map[host.Name]*model.Service) []config.Config {
 	out := make([]config.Config, 0)
 	for _, c := range virtualServices {
--- a/pilot/pkg/networking/core/v1alpha3/httproute.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute.go
@@ -181,6 +181,46 @@ func (configgen *ConfigGeneratorImpl) buildSidecarOutboundHTTPRouteConfig(node *
 	return out
 }
 
+// TODO: merge with IstioEgressListenerWrapper.selectVirtualServices
+func selectVirtualServices(virtualServices []config.Config, servicesByName map[host.Name]*model.Service) []config.Config {
+	out := make([]config.Config, 0)
+	for _, c := range virtualServices {
+		rule := c.Spec.(*networking.VirtualService)
+		var match bool
+
+		// Selection algorithm:
+		// virtualservices have a list of hosts in the API spec
+		// if any host in the list matches one service hostname, select the virtual service
+		// and break out of the loop.
+		for _, h := range rule.Hosts {
+			// TODO: This is a bug. VirtualServices can have many hosts
+			// while the user might be importing only a single host
+			// We need to generate a new VirtualService with just the matched host
+			if servicesByName[host.Name(h)] != nil {
+				match = true
+				break
+			}
+
+			for svcHost := range servicesByName {
+				if host.Name(h).Matches(svcHost) {
+					match = true
+					break
+				}
+			}
+
+			if match {
+				break
+			}
+		}
+
+		if match {
+			out = append(out, c)
+		}
+	}
+
+	return out
+}
+
 func BuildSidecarOutboundVirtualHosts(node *model.Proxy, push *model.PushContext,
 	routeName string, listenerPort int) []*route.VirtualHost {
 	var virtualServices []config.Config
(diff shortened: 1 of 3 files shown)


hzxuzhonghu:
fix #33760

pilot/pkg/networking/core/v1alpha3/httproute_test.go
```diff
@@ -957,9 +996,6 @@ func TestSidecarOutboundHTTPRouteConfig(t *testing.T) {
 				"test-private.com:80": {
 					"test-private.com": true, "test-private.com:80": true, "9.9.9.9": true, "9.9.9.9:80": true,
 				},
-				"test-private-3.com:80": {
```
howardjohn, line 960: can you add some explanation? This seems like a major breaking change

howardjohn: Oh nevermind, I see the linked issue now

howardjohn (commented):
I am not sure what problems this is solving. What is the problem with adding the route?

hzxuzhonghu:
There are two caveats:
1. for each route, unknown vses are used to build http route
2.  only for route 80, the virtualhost includes the unknown hosts. (It is not intended, but there are some tests depend on this behavior)

hzxuzhonghu:
IMO, we can leave the hacky 80 only for listenerPort=80 to not break compatibility if someone depends on this behavior, and reduce the cpu cost

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -199,15 +199,11 @@ func buildSidecarVirtualHostsForVirtualService(
...
-	// going to send a separate RDS request
-	// Note that we need to build non-default HTTP routes only for the virtual services.
-	// The services in the serviceRegistry will always have a default route (/)
 	if len(serviceByPort) == 0 {
-		// This is a gross HACK. Fix me. Its a much bigger surgery though, due to the way
-		// the current code is written.
-		serviceByPort[80] = nil
+		// no services or no service port for the listener port
+		return nil
 	}
```
hzxuzhonghu, line 205: This is the legacy hack code

ramaraochavali:
This change makes sense to me

howardjohn:
>  (It is not intended, but there are some tests depend on this behavior)

Are these unit tests? Or the security integration tests that are failing?

If its integration tests I would be worried this would be impacting users. For unit tests maybe not, since unit tests sometimes test impossible configurations

hzxuzhonghu:
The the security integration tests depends on this.


Before this patch, even no SE defined for `www.google.com`, the eventual output is: 
1.  one VirtualHost generated for `google.com` on route 80. 
2.  there are also VirtualHost built for each route as an intermediate procedure, but not added at all. 

If you have concern of breaking some compatibility, we can leave 1, and probihit  2. 

In this way it is safer, there is no user-sensitive behavior.

```
kind: VirtualService
metadata:
  name: test
spec:
  hosts:
  - www.google.com
  http:
  - name: "test"
    route:
    - destination:
        host: www.google.com
```

hzxuzhonghu:
@ramaraochavali @howardjohn Reverted the hack behavior for 80

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -230,6 +270,11 @@ func (configgen *ConfigGeneratorImpl) buildSidecarOutboundVirtualHosts(node *mod
 		}
 	}
 
+	// This is hack to keep consistent with previous behavior.
+	if listenerPort != 80 {
+		// only select virtualServices that matches a service
+		virtualServices = selectVirtualServices(virtualServices, servicesByName)
```
howardjohn, line 276: I do not think we have any test coverage of this change. If you comment it out all tests still pass. 

This also makes it hard to review the rest as I don't understand the full scope of the behavior changed by the PR, so I am not sure if this is safe or not. We would want to make sure these routes we are adding are truly never used before removing them, as we have had a lot of regressions in this config in the past

hzxuzhonghu: Yup, it passes without this selection. I explained above https://github.com/istio/istio/pull/33764#issuecomment-873505981
>there are also VirtualHost built for each route as an intermediate procedure, but not added at all.

I am thinking about adding a test for this, but does not come up with a good way

hzxuzhonghu: addes test

hzxuzhonghu:
@howardjohn @ramaraochavali ptal, will depend on this to do rds cache, otherwise, there will be too many vses

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -181,6 +181,46 @@ func (configgen *ConfigGeneratorImpl) buildSidecarOutboundHTTPRouteConfig(node *
 	return out
 }
 
+// TODO: merge with IstioEgressListenerWrapper.selectVirtualServices
+func selectVirtualServices(virtualServices []config.Config, servicesByName map[host.Name]*model.Service) []config.Config {
```
ramaraochavali, line 185: nit: Can you add a comment on what this does?

ramaraochavali (commented):
LGTM.Will keep it open for @howardjohn review. I will approve on Monday.

pilot/pkg/networking/core/v1alpha3/httproute.go
hzxuzhonghu: yes, make sense

hzxuzhonghu:
ping @howardjohn

ramaraochavali (approved):
@howardjohn I am approving this as it LGTM. If you have further comments, @hzxuzhonghu can do a followup.

pilot/pkg/networking/core/v1alpha3/httproute.go
howardjohn: Removing this still passes all tests I think

howardjohn: Oh but we have tests for the function itself, i see

Issue #33765, opened 2021-07-01
multiple egress hosts in Sidecar select same service


kebe7jun:
**Bug description**
apply the following sidecar config:

```yaml
apiVersion: networking.istio.io/v1beta1
kind: Sidecar
metadata:
  name: default
  namespace: istio-system
spec:
  egress:
  - hosts:
    - '*/*'
    - istio-system/*
```

the istiod show the following logs:
```console
2021-07-01T09:36:20.501701Z	info	ads	EDS: PUSH for node:s5-5796576b6f-vgphd.default resources:44 size:6.7kB empty:0 cached:44/44
2021-07-01T09:36:20.503460Z	info	ads	LDS: PUSH for node:s5-5796576b6f-vgphd.default resources:22 size:65.5kB
2021-07-01T09:36:20.504226Z	info	ads	RDS: PUSH for node:s5-5796576b6f-vgphd.default resources:10 size:20.1kB
2021-07-01T09:36:25.362251Z	info	ads	Push Status: {
    "pilot_conflict_outbound_listener_tcp_over_current_tcp": {
        "10.96.178.84:15443": {
            "proxy": "s5-5796576b6f-vgphd.default",
            "message": "Listener=10.96.178.84:15443 AcceptedTCP=istio-ingressgateway.istio-system.svc.cluster.local RejectedTCP=istio-ingressgateway.istio-system.svc.cluster.local TCPServices=1"
        },
        "10.96.178.84:443": {
            "proxy": "s5-5796576b6f-vgphd.default",
            "message": "Listener=10.96.178.84:443 AcceptedTCP=istio-ingressgateway.istio-system.svc.cluster.local RejectedTCP=istio-ingressgateway.istio-system.svc.cluster.local TCPServices=1"
        },
        "10.96.179.59:15012": {
            "proxy": "s5-5796576b6f-vgphd.default",
            "message": "Listener=10.96.179.59:15012 AcceptedTCP=istiod.istio-system.svc.cluster.local RejectedTCP=istiod.istio-system.svc.cluster.local TCPServices=1"
        },
        "10.96.179.59:443": {
            "proxy": "s5-5796576b6f-vgphd.default",
            "message": "Listener=10.96.179.59:443 AcceptedTCP=istiod.istio-system.svc.cluster.local RejectedTCP=istiod.istio-system.svc.cluster.local TCPServices=1"
        },
        "10.96.188.94:15012": {
            "proxy": "s5-5796576b6f-vgphd.default",
[...]

PR #33766, opened 2021-07-01
fix multiple egress hosts select same service error

kebe7jun:
Fix: https://github.com/istio/istio/issues/33765


hzxuzhonghu:
Looks good, wondering if this conflict can be validated

kebe7jun:
why the tests always Expected...

howardjohn (approved):

hzxuzhonghu:
Yeah, i mean if we can prevent this in galley. Which will be once for all.

kebe7jun:
> Yeah, i mean if we can prevent this in galley. Which will be once for all.

sounds good, I will create another PR to do that.

PR #33773, opened 2021-07-01
Extend `istioctl tag set default` to control validation

Monkeyanator:
Implements [https://docs.google.com/document/d/1P1mNlrJOs-KHVqcKTrFLrvCUiLgyZ-Vlf7ezaZNvD6g/edit?usp=sharing](https://docs.google.com/document/d/1P1mNlrJOs-KHVqcKTrFLrvCUiLgyZ-Vlf7ezaZNvD6g/edit?usp=sharing)

Makes it so that running `istioctl tag set default --revision` moves which revision handles validation.

- [x] Disable validation by default on revisioned webhooks
- [x] Ensure that each revision patches validating webhooks with matching istio.io/rev
- [x] Add template for default validating webhook configuration
- [x] Create logic in istioctl tag to create and generate the default validator
- [ ] Istioctl tag delete default should remove the default validator
- [x] Make it possible to point default tag at pre-1.11 revisions
- [ ] Write integration test for istioctl tag set validation changes
- [x] Add logic to installer to make an Istio installation default if there is no existing default
- [x] Provide a Helm-centric way to switch default revision
- [x] Purge deprecated VWH when the default is set

Requires the following documentation changes:

1. For Helm install, when installing base, use the `--defaultRevision=<rev>` option.
2. Document the process of installing a new revision, evaluating it, and switching the default revision with `istioctl tag set default --revision <rev>`.
3. Document default revision semantics including (1) that it handles injection for `istio-injection=enabled` and (2) it handles resource validation for the entire cluster.


istioctl/pkg/tag/templates/validatingwebhook.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: admissionregistration.k8s.io/v1
```
howardjohn, line 1: Just curious why we have a copy under istioctl for this, but mutating webhook is using the chart's version? Is this unique from the chart version?

istioctl/pkg/tag/templates/validatingwebhook.yaml
```diff
@@ -0,0 +1,47 @@
...
+        namespace: {{ or .IstioNamespace "istio-system" }}
+        path: "/validate"
+      {{- end }}
+      caBundle: {{ or .CABundle "" }}
+    rules:
+      - operations:
+          - CREATE
+          - UPDATE
+        apiGroups:
+          - security.istio.io
```
howardjohn, line 34: This list is not complete, and i see its not in manifests/ either. So that is a bug in current master branch I think

istioctl/pkg/tag/templates/validatingwebhook.yaml
Monkeyanator: Mutating webhooks can move here too since we decided being able to set tags through IstioOperator is not a priority.

istioctl/pkg/tag/templates/validatingwebhook.yaml
Monkeyanator: Yup I noticed that was missing telem API group.

istioctl/pkg/tag/util.go
```diff
@@ -0,0 +1,171 @@
...
+		return false, err
+	}
+	return len(tagWhs) > 0 || len(whs) > 0, nil
+}
+
+// DeactivateIstioInjectionWebhook deactivates the istio-injection webhook from the given MutatingWebhookConfiguration if exists.
+// used rather than just deleting the webhook since we want to keep it around after changing the default so user can later
+// switch back to it. This is a hack but it is meant to cover a corner case where a user wants to migrate from a non-revisioned
+// old version and then later decides to switch back to the old revision again.
+func DeactivateIstioInjectionWebhook(ctx context.Context, client kubernetes.Interface) error {
```
Monkeyanator, line 142: This tries to make it possible to switch back to a non-revisioned install after making a revisioned install the default, a bit gross

istioctl/pkg/tag/templates/validatingwebhook.yaml
Monkeyanator: Will handle in future PR.

istioctl/pkg/tag/templates/validatingwebhook.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
```
costinm, line 3: Would it be hard to move this to /manifests/charts/default/templates/ - in preparation for adding a Chart.yaml ? And also when executing the template, set the variables to match helm ( {{ .Values.tag }}, etc.

If it's too hard we can leave it for later - or just duplicate it. I have a chart with equivalent functionality in costinm/istiod/manifests/.. - I was planning to just leave it as a sample.

Alternatively, you could use the template from 'base', or add few extra {{ if }} to 
the base so it renders tags ( and helm install base --set tag=foo --set rev=bar applies just the webhooks, the rest is skipped ). This will keep things in sync.
( istioctl has code to render the chart, so may be even less code )

istioctl/pkg/tag/util.go
costinm: Right, but I can't think of a better option. For 1.11+ we should just not install this webhook, even for the 'revision=""' - so it will not be a long-term problem, once 1.10 is out of support we can remove this hack.

Make sure you document the behavior - and the fact that if user make changes to the 1.10 'unrevisioned', like enable access logs - it will revert your hack. 

It may be a case of just documenting a manual rollback procedure for 1.10 - and making sure in future we are fine.

costinm (commented):
In general it seems fine and better than what is currently in master. 
But I suspect we may need to iterate a bit more.

nrjpoddar (commented):
I want to make sure the [design doc](https://docs.google.com/document/d/1P1mNlrJOs-KHVqcKTrFLrvCUiLgyZ-Vlf7ezaZNvD6g/edit?ts=60e5ef35) is approved and the flow works for Helm along with istioctl before we merge this PR in.

Monkeyanator:
> I want to make sure the [design doc](https://docs.google.com/document/d/1P1mNlrJOs-KHVqcKTrFLrvCUiLgyZ-Vlf7ezaZNvD6g/edit?ts=60e5ef35) is approved and the flow works for Helm along with istioctl before we merge this PR in.

Added a section for how this can work with Helm @nrjpoddar

istioctl/pkg/tag/util.go
```diff
@@ -0,0 +1,168 @@
...
+		// no revision with default, no action required.
+		return nil
+	}
+	if len(whs) > 1 {
+		return fmt.Errorf("expected a single webhook for default revision")
+	}
+	webhook := whs[0]
+	for i := range webhook.Webhooks {
+		wh := webhook.Webhooks[i]
+		// this is an abomination, but if this isn't a per-revision webhook, we want to make it ineffectual
```
howardjohn, line 156: Is it possible to just keep the MutatingWebhookConfiguration with empty list of Webhooks?

istioctl/pkg/tag/util.go
```diff
@@ -0,0 +1,168 @@
...
+
+	nsNames := make([]string, len(namespaces.Items))
+	for i, ns := range namespaces.Items {
+		nsNames[i] = ns.Name
+	}
+	return nsNames, nil
+}
+
+// GetWebhookName extracts tag name from webhook object.
+func GetWebhookName(wh admit_v1.MutatingWebhookConfiguration) (string, error) {
```
howardjohn, line 91: nit: GetWebhookTagName may be more accurate

operator/cmd/mesh/install.go
```diff
@@ -185,6 +186,32 @@ func runApplyCmd(cmd *cobra.Command, rootArgs *rootArgs, iArgs *installArgs, log
...
 	}
 
+	// Make this revision the default if none exists
+	exists, err := revtag.DefaultRevisionExists(context.Background(), kubeClient)
+	if err != nil {
+		return err
+	}
+	rev := iop.Spec.Revision
+	if !exists || rev == "" {
+		cmd.Print("Making this installation the default for injection and validation.")
```
howardjohn, line 196:
```suggestion
		cmd.Print("Making this installation the default for injection and validation.\n")
```

howardjohn: or Println if they have it

istioctl/cmd/tag.go
```diff
@@ -276,63 +255,36 @@ revision tag before removing using the "istioctl x revision tag list" command.
...
 	if resName == "" {
-		resName = fmt.Sprintf("%s-%s", "istio-revision-tag", tag)
+		resName = fmt.Sprintf("%s-%s", "istio-revision-tag", tagName)
 	}
 	if err := analyzeWebhook(resName, tagWhYAML, kubeClient.RESTConfig()); err != nil {
 		// if we have a conflict, we will fail. If --skip-confirmation is set, we will continue with a
 		// warning; when actually applying we will also confirm to ensure the user does not see the
 		// warning *after* it has applied
-		if skipConfirmation {
+		if !skipConfirmation {
```
howardjohn, line 281: Not sure I understand why this was inverted?

manifests/charts/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
```diff
@@ -50,36 +50,5 @@ webhooks:
           {{- else }}
           - "{{ .Values.revision }}"
           {{- end }}
-  # Webhook handling default validation
```
howardjohn, line 53: should we remove the istiod-istio-system one as well now?

operator/cmd/mesh/install.go
```diff
@@ -185,6 +186,32 @@ func runApplyCmd(cmd *cobra.Command, rootArgs *rootArgs, iArgs *installArgs, log
 		return fmt.Errorf("failed to install manifests: %v", err)
 	}
 
+	// Make this revision the default if none exists
```
howardjohn, line 189: I don't know how much of this is broken already, but `istioctl x unintsall` leads around all the webhooks (including validation) which ends up breaking

howardjohn (commented):
overall looks good!

istioctl/pkg/tag/util.go
Monkeyanator: Yeah the reason I did it this way is we need to keep the injection URL around _somewhere_ so we can create new tags from the webhook. Right now the webhooks are the only place it's persisted.

istioctl/cmd/tag.go
Monkeyanator: When we automatically set the default revision in the installer, we need it to happen without user confirmation (like when we do `install -y` in test framework). This makes it so that we only do the duplicate injection analysis and confirmation if `skip-confirmation` is disabled.

manifests/charts/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
Monkeyanator: I still don't think we can because of the resource pruning removing the old, active validator?

operator/cmd/mesh/install.go
Monkeyanator: Hm yeah it seems like the webhooks get removed when doing `uninstall --purge` but not `uninstall --revision` (even though the `istio.io/rev` label is there)... I'll try and fix as part of this PR.

istioctl/pkg/tag/util.go
howardjohn: Ah, lame. Thats annoying. I don't know how often users are actually doing this but i was pretty confusing looking at `kubectl get mutatingwebhooks` and seeing a bunch, when the yaml shows disabled. In theory we could stick it in an annotation but I am not sure that is worth the complexity, since in some cases its not just a simple URL (port, cabundle, etc).

I guess the comment is accurate :slightly_smiling_face:

manifests/charts/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
howardjohn: Aren't we replacing it with a new active validator though? I guess not if we are not the default?

howardjohn: I think fixing the old validator is a must-do for 1.11

istioctl/pkg/tag/util.go
Monkeyanator: I have no doubt that someone, somewhere will eventually come to rely on `"istio.io/deactivated": "never-match"` for injection and we'll have to keep it around :smile:

istioctl/pkg/tag/templates/validatingwebhook.yaml
Monkeyanator: Makes sense, moved the template to `manifests/charts/default` and the command now renders from there. There's also a `defaultRevision` option now on the base chart.

manifests/charts/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
Monkeyanator: Just tested and it seems like removing the webhook doesn't actually purge it from old installations... should be clear to remove.

howardjohn:
Should we explicitly prune it if/when we set a new default? That way users
won't have a broken webhook laying around?

manifests/charts/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
howardjohn: Should we explicitly purge it when we set the default? That way we don't get a stale webhook, which will either give noisy errors in Istiod at best, or block config at worst?

Monkeyanator:
> Should we explicitly prune it if/when we set a new default?

@howardjohn the reason it's kept around here is so that users can switch back to it if things go wrong after the initial switch. Maybe we can just do away with the whole `DeactivateWebhook` hack and delete the webhook and document the caveat?

manifests/charts/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
Monkeyanator: Made it so that (1) validating webhooks get removed on `istioctl uninstall --revision` and (2) we explicitly purge the old one when setting default.

pkg/webhooks/validation/controller/controller.go
```diff
@@ -121,6 +120,7 @@ const (
 type reconcileRequest struct {
 	event       eventType
 	description string
+	webhookName string // if empty, ALL webhooks should be updated in response
```
Monkeyanator, line 123: had to make validation controller patch all webhooks with matching `istio.io/rev` since for Helm base install we need `istiod` to patch in the cert

operator/cmd/mesh/install.go
Monkeyanator: Fixed.

Monkeyanator:
@costinm @howardjohn believe I've addressed most of the feedback, PTAL when you have the chance.

howardjohn:
It seems like `istiod-default-validator` is not patched by any istiod revision. Works for most cases, but if cabundle changes it will explode

pkg/webhooks/validation/controller/controller.go
```diff
@@ -279,37 +282,76 @@ func (c *Controller) processNextWorkItem() (cont bool) {
...
 
-var retryRequest = &reconcileRequest{
-	event:       retryEvent,
-	description: "retry reconcile request",
+// updateAll updates all webhooks matching the controller's revision, generally in response
+// to a CA bundle update. updateAll reports an error only when there's an issue listing webhooks,
+// not when it has an issue updating a single webhook so that we can retry separately for the
+// two cases.
+func (c *Controller) updateAll() error {
+	whs, err := c.client.AdmissionregistrationV1().ValidatingWebhookConfigurations().List(context.Background(), metav1.ListOptions{})
```
howardjohn, line 323: I would prefer `c.webhookInformer.GetStore().List()`. Its cached already, so no call to api-server like here, and already has filter logic for `-l istio.io/rev=<rev>`

howardjohn: Or if not, we should probably at least filter?

howardjohn (approved):

pkg/webhooks/validation/controller/controller.go
Monkeyanator: Makes sense, fixed

jacob-delgado:
What are the steps to migrate users from the older chart to the new ones? This is starting to look pretty good. The helm changes look fine, but some clarity on steps would be useful.

Monkeyanator:
> What are the steps to migrate users from the older chart to the new ones?

The only action required is that Helm users will have to add `--defaultRevision default` (or whatever revision) when installing the new base charts. This will (1) remove the old validator and (2) create the new default validator `istio-default-validator`.

pkg/webhooks/validation/controller/controller.go
```diff
@@ -170,37 +171,13 @@ func makeHandler(queue workqueue.Interface, gvk schema.GroupVersionKind) *cache.
...
 			}
-			req := &reconcileRequest{
+			req := reconcileRequest{
 				event:       updateEvent,
+				webhookName: currObj.GetName(),
 				description: fmt.Sprintf("update event (%v, Kind=%v) %v", gvk.GroupVersion(), gvk.Kind, key),
 			}
 			queue.Add(req)
 		},
-		DeleteFunc: func(curr interface{}) {
```
Monkeyanator, line 179: Deleted this since I don't believe we need to reconcile on deletion and it caused race conditions when switching revision tag on default validator

jacob-delgado (approved):
Helm changes lgtm and doc is approved

esnible (commented):
I like the direction.

Would it be reasonable to also modify `revision list` to add columns for both mutating and validating webhooks?

Would it be reasonable to modify `revision describe <tag>` to add section for validating webhook?

Monkeyanator:
> Would it be reasonable to also modify revision list to add columns for both mutating and validating webhooks?

@esnible I think we should avoid a adding column for validating webhook to `istioctl tag list` since we only expect this to ever happen with the `default` tag. Might also leak too much implementation detail about how tags are implemented in the command output. 

IMO we just need to make sure that the meaning of the `default` tag is well documented (both in the command help text and actual docs).

Monkeyanator:
@istio/wg-docs-maintainers @istio/wg-user-experience-maintainers PTAL when you have the chance, thanks!

ericvn (approved):

therealmitchconnors (approved):

ericvn:
As noted, this PR needs some istio.io changes and moving the istio.io tests to after this commit has caused some failures. See  https://github.com/istio/istio.io/pull/10208

PR #33774, opened 2021-07-01
allow WorkloadEntry port mapping to use regular port name

stevenctl:
In the current state, a workload entry with

```
ports:
  grpc: 17070
```

and a service with:

```
ports:
- name: grpc
   port: 7070
```

does _not_ result in the port mapping resulting with 17070 in eds

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -844,7 +844,7 @@ func (c *Controller) serviceInstancesFromWorkloadInstances(svc *model.Service, r
 	}
 
 	// Now get the target Port for this service port
-	targetPort, targetPortName, explicitTragetPort := findServiceTargetPort(servicePort, k8sService)
+	targetPort, targetPortName, explicitTargetPort := findServiceTargetPort(servicePort, k8sService)
 	if targetPort == 0 {
 		targetPort = reqSvcPort
 	}
@@ -868,7 +868,7 @@ func (c *Controller) serviceInstancesFromWorkloadInstances(svc *model.Service, r
 				matchedPort := wi.PortMap[targetPortName]
 				if matchedPort != 0 {
 					istioEndpoint.EndpointPort = matchedPort
-				} else if explicitTragetPort {
+				} else if explicitTargetPort {
 					// No match found, and we expect the name explicitly in the service, skip this endpoint
 					continue
 				}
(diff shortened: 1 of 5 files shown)


howardjohn (commented):
Seems to make sense, can we add a test case to https://github.com/istio/istio/blob/1274725f3d6d95c6ba638113ea503a6d04bda666/pilot/pkg/serviceregistry/serviceregistry_test.go#L567 though?

pilot/pkg/serviceregistry/kube/controller/util.go
```diff
@@ -104,8 +104,10 @@ func findServiceTargetPort(servicePort *model.Port, k8sService *v1.Service) (int
 		if p.Name == servicePort.Name || p.Port == int32(servicePort.Port) {
 			if p.TargetPort.Type == intstr.Int && p.TargetPort.IntVal > 0 {
 				targetPort = int(p.TargetPort.IntVal)
-			} else {
+			} else if p.TargetPort.StrVal != "" {
 				targetPortName = p.TargetPort.StrVal
+			} else {
```
hzxuzhonghu, line 109: I guess this could never happen, let me verify

hzxuzhonghu: Yes, this branch can never enter, because k8s will by default set the targetPort = port if user does not set it.

stevenctl: I see – thanks for clarifying. Is the behavior in the description correct though? I was really confused when I tried to do it myself.

hzxuzhonghu: Yes, it is expected

pilot/pkg/serviceregistry/serviceregistry_test.go
```diff
@@ -1108,6 +1149,15 @@ func makePod(t *testing.T, c kubernetes.Interface, pod *v1.Pod) {
 
 func makeService(t *testing.T, c kubernetes.Interface, svc *v1.Service) {
 	t.Helper()
+	// avoid mutating input
+	svc = svc.DeepCopy()
+	// simulate actual k8s behavior
+	for i, port := range svc.Spec.Ports {
```
stevenctl, line 1155: Thanks for the pointer @hzxuzhonghu!

stevenctl:
@hzxuzhonghu @howardjohn ping

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -844,7 +844,7 @@ func (c *Controller) serviceInstancesFromWorkloadInstances(svc *model.Service, r
 	}
 
 	// Now get the target Port for this service port
-	targetPort, targetPortName := findServiceTargetPort(servicePort, k8sService)
+	targetPort, targetPortName, explicitTragetPort := findServiceTargetPort(servicePort, k8sService)
```
howardjohn, line 847: explicitTragetPort -> explicitTargetPort

howardjohn (approved):

PR #33778, opened 2021-07-01
[1.10] Fix endpointslice duplicate IPs on service update (#33757)

howardjohn:
We should be reading from the cache which is more efficient and handles
deduplication properly, rather than recomputing.

The PR is basically a few lines of a code and a lot of refactoring to
get the test to work. The problem was the XdsUpdater isn't actually set
in FakeXdsServer; in the TestSameIPEndpointSlicing test which this is
based on the Wait() always times out (so its not actually waiting)
because we use a xdsUpdater that isn't actually wired up to the
controller.

(cherry picked from commit 1274725f3d6d95c6ba638113ea503a6d04bda666)

Fixes https://github.com/istio/istio/issues/33761

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/endpointslice.go
+++ b/pilot/pkg/serviceregistry/kube/controller/endpointslice.go
@@ -196,12 +196,7 @@ func (esc *endpointSliceController) buildIstioEndpointsWithService(name, namespa
 		return nil
 	}
 
-	endpoints := make([]*model.IstioEndpoint, 0)
-	for _, es := range slices {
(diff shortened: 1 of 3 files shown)


ramaraochavali (approved):
Code changes LGTM. Have not reviewed the tests thoroughly though.

Monkeyanator (approved):

PR #33782, opened 2021-07-02
Make gateway status writing optional and add leader election

howardjohn:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/bootstrap/configcontroller.go
+++ b/pilot/pkg/bootstrap/configcontroller.go
@@ -158,8 +158,22 @@ func (s *Server) initK8SConfigStore(args *PilotArgs) error {
 	}
 	s.ConfigStores = append(s.ConfigStores, configController)
 	if features.EnableServiceApis {
-		s.environment.GatewayAPIController = gateway.NewController(s.kubeClient, configController, args.RegistryOptions.KubeOptions)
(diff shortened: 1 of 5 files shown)


pilot/pkg/model/context.go
```diff
@@ -1044,4 +1044,5 @@ func (node *Proxy) IsVM() bool {
 type GatewayController interface {
 	ConfigStoreCache
 	Recompute(GatewayContext) error
+	SetStatusWrite(enabled bool)
```
ramaraochavali, line 1047: It seems SetStatusWrite is a too narrow scoped function to be in the GatewayController interface. Can we add a generic callback functiond like "OnNewLeader"  - we can have a boolean indicating if this controller is the leader so that we can do these types of things?

ramaraochavali: BTW, Just saw this  https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L165 - So we may not need full interface here though

howardjohn: Did a small refactor to avoid needing anything here

ramaraochavali (approved):

PR #33786, opened 2021-07-02
Add sidecar egress hosts conflict warnings

kebe7jun:
For https://github.com/istio/istio/pull/33766#issuecomment-872667419

### DIFF
--- a/pkg/config/validation/validation_test.go
+++ b/pkg/config/validation/validation_test.go
@@ -5123,57 +5123,58 @@ func TestValidateSidecar(t *testing.T) {
 		name  string
 		in    *networking.Sidecar
 		valid bool
+		warn  bool
 	}{
-		{"empty ingress and egress", &networking.Sidecar{}, false},
+		{"empty ingress and egress", &networking.Sidecar{}, false, false},
 		{"default", &networking.Sidecar{
 			Egress: []*networking.IstioEgressListener{
 				{
 					Hosts: []string{"*/*"},
 				},
 			},
-		}, true},
+		}, true, false},
 		{"import local namespace with wildcard", &networking.Sidecar{
 			Egress: []*networking.IstioEgressListener{
 				{
 					Hosts: []string{"./*"},
 				},
 			},
-		}, true},
+		}, true, false},
 		{"import local namespace with fqdn", &networking.Sidecar{
 			Egress: []*networking.IstioEgressListener{
 				{
 					Hosts: []string{"./foo.com"},
 				},
 			},
-		}, true},
+		}, true, false},
 		{"import nothing", &networking.Sidecar{
 			Egress: []*networking.IstioEgressListener{
 				{
 					Hosts: []string{"~/*"},
 				},
 			},
-		}, true},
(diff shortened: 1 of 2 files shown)


ramaraochavali:
LGTM. Will leave it for others to review

howardjohn (changes requested):
Making previously valid configurations invalid is a breaking API change. At most we can warn.

kebe7jun:
> Making previously valid configurations invalid is a breaking API change. At most we can warn.

@howardjohn  Is this an indication that we only need to print the logs?

hzxuzhonghu:
Add a warning, and add a TODO maybe in 1.12 we can make it this way.

FYI: `WrapWarning`

kebe7jun:
> Add a warning, and add a TODO maybe in 1.12 we can make it this way.
> 
> FYI: `WrapWarning`

Thanks for your suggestions, I did it.

howardjohn:
> Add a warning, and add a TODO maybe in 1.12 we can make it this way.

We can never make it this way without it being a breaking API change. To do this it would require a new ApiVersion

pkg/config/validation/validation_test.go
```diff
@@ -5580,10 +5580,53 @@ func TestValidateSidecar(t *testing.T) {
...
+			Egress: []*networking.IstioEgressListener{
+				{
+					Hosts: []string{
+						"test/*",
+						"test/a.com",
+					},
+				},
+			},
+		}, false},
+		{"skip(1.11): sidecar egress duplicated with wildcarded same namespace .", &networking.Sidecar{
```
howardjohn, line 5613: what does "skip(1.11)" mean?

howardjohn: Oh I see. We should test the warnings as well, using checkValidationMessage like some other tests

hzxuzhonghu:
@howardjohn https://kubernetes.io/docs/reference/using-api/deprecation-policy/

howardjohn:
Yes, I would think this violates "Rule #1: API elements may only be removed
by incrementing the version of the API group"

Removing an element (either key or legal value) and making it invalid are
effectively the same?

kebe7jun:
@howardjohn  I see the same action is allowed in Kubernetes: https://github.com/kubernetes/kubernetes/pull/103245.

howardjohn:
Looks like that may be for an alpha feature flag protected field?

kebe7jun:
> Looks like that may be for an alpha feature flag protected field?

Yes, but this one does not affect the API and resources, it is only print some warnings in pilot, I can not do this also?

kebe7jun:
> Looks like that may be for an alpha feature flag protected field?

Yes, but this one does not affect the API and resources, it is only print some warnings in pilot, can we not do this either?

hzxuzhonghu:
LGTM, it only add some warning in validation, does not break any old configs

howardjohn:
Sorry I either misread it or didn't look at the update - adding a warning is good, just not error

hzxuzhonghu (approved):

PR #33788, opened 2021-07-02
Add updateInterval args

devincd:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/cmd/pilot-discovery/app/cmd.go
+++ b/pilot/cmd/pilot-discovery/app/cmd.go
@@ -138,7 +138,7 @@ func addFlags(c *cobra.Command) {
 	c.PersistentFlags().DurationVar(&serverArgs.ShutdownDuration, "shutdownDuration", 10*time.Second,
 		"Duration the discovery server needs to terminate gracefully")
 	c.PersistentFlags().DurationVar(&serverArgs.StatusReporterOptions.UpdateInterval, "updateInterval", 500*time.Millisecond,
-		"Interval duration the reporter updates status")
+		"The interval that status reporter updates status")
 
 	// RegistryOptions Controller options
 	c.PersistentFlags().StringVar(&serverArgs.RegistryOptions.FileDir, "configDir", "",
--- a/pilot/cmd/pilot-discovery/app/cmd.go
+++ b/pilot/cmd/pilot-discovery/app/cmd.go
@@ -137,6 +137,8 @@ func addFlags(c *cobra.Command) {
 		"comma separated list of networking plugins to enable")
 	c.PersistentFlags().DurationVar(&serverArgs.ShutdownDuration, "shutdownDuration", 10*time.Second,
 		"Duration the discovery server needs to terminate gracefully")
+	c.PersistentFlags().DurationVar(&serverArgs.StatusReporterOptions.UpdateInterval, "updateInterval", 500*time.Millisecond,
+		"Interval duration the reporter updates status")
 
 	// RegistryOptions Controller options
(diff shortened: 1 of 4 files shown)


hzxuzhonghu (approved):
lgtm

releasenotes/notes/add-updateinterval-args.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+releaseNotes:
+  - |
+    **Added** a flag setting `updateInterval` in pilot and its default value is 500ms.
```
hzxuzhonghu, line 6: It needs to be more clear about what it means

pilot/cmd/pilot-discovery/app/cmd.go
```diff
@@ -137,6 +137,8 @@ func addFlags(c *cobra.Command) {
 		"comma separated list of networking plugins to enable")
 	c.PersistentFlags().DurationVar(&serverArgs.ShutdownDuration, "shutdownDuration", 10*time.Second,
 		"Duration the discovery server needs to terminate gracefully")
+	c.PersistentFlags().DurationVar(&serverArgs.StatusReporterOptions.UpdateInterval, "updateInterval", 500*time.Millisecond,
+		"Interval duration the reporter updates status")
```
hzxuzhonghu, line 141: `The interval that status reporter updates status`

ramaraochavali:
Do we want to do this via command line arg or env var? Right now status is enabled via Env var and I guess at some point it might become part of meshconfig?
cc: @howardjohn

pilot/cmd/pilot-discovery/app/cmd.go
```diff
@@ -137,6 +137,8 @@ func addFlags(c *cobra.Command) {
 		"comma separated list of networking plugins to enable")
 	c.PersistentFlags().DurationVar(&serverArgs.ShutdownDuration, "shutdownDuration", 10*time.Second,
 		"Duration the discovery server needs to terminate gracefully")
+	c.PersistentFlags().DurationVar(&serverArgs.StatusReporterOptions.UpdateInterval, "updateInterval", 500*time.Millisecond,
+		"The interval that status reporter updates status")
```
brian-avery, line 141: How does this sound?

> Interval to update the XDS distribution status.

releasenotes/notes/add-updateinterval-args.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+releaseNotes:
+  - |
+    **Added** a flag setting `updateInterval` that is the interval that status reporter updates status and its default value is 500ms.
```
brian-avery, line 6: Based on the above comment: 

```suggestion
    **Added** a flag, `updateInterval`, to configure the interval to update the XDS distribution status at. This defaults to 500ms.
```

devincd:
> We have been trying to move away from using command line arguments which present a number of problems. Generally the progression is environment variable (experimental) -> mesh config (when not experimental).
> 
> Could you describe the use case for configuring this?

I just want to let `UpdateInterval` (the interval that status reporter updates status) to be configurable, instead of the default 500ms. So using command line arguments, environment variable and mesh config are both ok. I will submit another PR with environment variable.

pilot/cmd/pilot-discovery/app/cmd.go
devincd: It sounds good. Thinks

PR #33792, opened 2021-07-02
address dns metrics comments

ramaraochavali:
[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/dns/monitoring.go
+++ b/pilot/pkg/dns/monitoring.go
@@ -35,9 +35,9 @@ var (
 	)
 
 	requestDuration = monitoring.NewDistribution(
-		"dns_upstream_request_time",
+		"dns_upstream_request_duration_seconds",
 		"Total time in seconds Istio takes to get DNS response from upstream.",
-		[]float64{.01, .1, 0.5, 1, 3, 5},
+		[]float64{.005, .001, 0.01, 0.1, 1, 5},
 	)
 )
 
--- a/pilot/pkg/dns/monitoring.go
+++ b/pilot/pkg/dns/monitoring.go
@@ -30,14 +30,14 @@ var (
 	)
 
 	failures = monitoring.NewSum(
-		"dns_total_upstream_failures",
+		"dns_upstream_failures_total",
 		"Total number of DNS requests forwarded to upstream.",
 	)
 
 	requestDuration = monitoring.NewDistribution(
-		"dns_upstream_request_time_ms",
-		"Total time in milliseconds Istio takes to get DNS response from upstream.",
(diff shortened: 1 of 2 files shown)


ramaraochavali:
@howardjohn PTAL

pilot/pkg/dns/monitoring.go
```diff
@@ -30,14 +30,14 @@ var (
...
-		"dns_total_upstream_failures",
+		"dns_upstream_failures_total",
 		"Total number of DNS requests forwarded to upstream.",
 	)
 
 	requestDuration = monitoring.NewDistribution(
-		"dns_upstream_request_time_ms",
-		"Total time in milliseconds Istio takes to get DNS response from upstream.",
-		[]float64{1, 3, 5, 10, 25, 50},
+		"dns_upstream_request_time",
```
howardjohn, line 38:
```suggestion
		"dns_upstream_request_duration_seconds",
```

pilot/pkg/dns/monitoring.go
```diff
@@ -30,14 +30,14 @@ var (
...
+		"Total time in seconds Istio takes to get DNS response from upstream.",
+		[]float64{.01, .1, 0.5, 1, 3, 5},
```
howardjohn, line 40: The buckets seem pretty skewed to extreme latency, while before we had max 50ms. I think we could probably have something like:
`[]float64{.005, .001, 0.01, 0.1, 1, 5},`

WDYT?

pilot/pkg/dns/monitoring.go
```diff
@@ -30,14 +30,14 @@ var (
 	)
 
 	failures = monitoring.NewSum(
-		"dns_total_upstream_failures",
+		"dns_upstream_failures_total",
 		"Total number of DNS requests forwarded to upstream.",
 	)
 
 	requestDuration = monitoring.NewDistribution(
```
howardjohn, line 37: We need to change the .Record call to use seconds now I think

pilot/pkg/dns/monitoring.go
ramaraochavali: Makes sense will change

pilot/pkg/dns/monitoring.go
ramaraochavali: correct. Missed that. Thanks for catching it.

ramaraochavali:
@howardjohn fixed.

howardjohn (approved):
for future viewers - this is backwards compatible because we never shipped the original names

Issue #33737, opened 2021-06-30
retryRemoteLocalities uses xds v2 API, broken on recent versions


howardjohn:
```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: ratings-route
spec:
  hosts:
  - echo.default.svc.cluster.local
  http:
  - route:
    - destination:
        host: echo.default.svc.cluster.local
    retries:
      attempts: 3
      perTryTimeout: 2s
      retryOn: gateway-error,connect-failure,refused-stream
      retryRemoteLocalities: true
```

```
2021-06-30T15:23:13.908540Z     warning envoy config    gRPC config for type.googleapis.com/envoy.config.route.v3.RouteConfiguration rejected: The v2 xDS major version is deprecated and disabled by default. Support for v2 will be removed from Envoy at the start of Q1 2021. You may make use of v2 in Q4 2020 by following the advice in https://www.envoyproxy.io/docs/envoy/latest/faq/api/transition. (envoy.config.retry.previous_priorities.PreviousPrioritiesConfig)
```

The fix is easy but this indicates a clear testing gap

pavanponugoti:
We have encountered this issue during Istio upgrade from 1.9.5 to 1.10.2. After removing retryRemoteLocalities from virtual services, ILBs started working.
Also, this link doesn't seem to exist: https://www.envoyproxy.io/docs/envoy/latest/faq/api/transition 
`Internal:The v2 xDS major version is deprecated and disabled by default. Support for v2 will be removed from Envoy at the start of Q1 2021. You may make use of v2 in Q4 2020 by following the advice in https://www.envoyproxy.io/docs/envoy/latest/faq/api/transition. (envoy.config.retry.previous_priorities.PreviousPrioritiesConfig)`

ramaraochavali:
Closing this as it is fixed

PR #33793, opened 2021-07-02
fix remote retry localities

ramaraochavali:
https://github.com/istio/istio/issues/33737

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

[] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/route/retry/retry.go
+++ b/pilot/pkg/networking/core/v1alpha3/route/retry/retry.go
@@ -19,8 +19,9 @@ import (
 	"strconv"
 	"strings"
 
-	previouspriorities "github.com/envoyproxy/go-control-plane/envoy/config/retry/previous_priorities"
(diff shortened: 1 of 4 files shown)


ramaraochavali:
@howardjohn @hzxuzhonghu Will add an integration test. Is there any test that can be modified to test this behaviour? I could not find any retry related to tests? Atleast it would be good to validate that this config is not rejected by Envoy if validating the remote locality retry is difficult (which might be the case I guess)

howardjohn (approved):

howardjohn:
The closest test is TestLocality. Its not easy to test retries I think.. not sure of any grat way

PR #33803, opened 2021-07-02
Improve mesh config merge logic for Telemetry

howardjohn:
This adds improved semantics for the new providers fields:
* Extensions providers will replace existing providers with the same
name. If there are none, they will append.
* Default providers is a deeper merge (like proxy config), so you can configure the default
provider for tracing without impacting metrics (for example).

Unit tests are provided that will show the above as well

Both of these should have ~no impact today. However, in the future we
will have built in defaults for providers, at which point these will be
critical to avoid any additions completely wiping out the defaults.

This needs to be rebase on https://github.com/istio/istio/pull/33801 before merge


douglas-reid (approved):
Looks reasonable to me. Probably needs a look from someone with a deeper understanding of this codebase, however.

costinm (approved):

Issue #33799, opened 2021-07-02
Add observability into connected clusters


howardjohn:
Currently, it can be hard to tell which clusters are connected when using multicluster. It would be great to have a way to see this. This could be debug endpoint and/or istioctl endpoint

Strawman UX:

```
$ istioctl clusters list
NAME            SECRET                 STATUS
cluster-123     cluster-123-secret      healthy
```

cc @stevenctl @esnible @nmittler

stevenctl:
Any proposal for how we define `STATUS`? I can see SYNCING, SYNCED (actually just initial sync) and SYNC TIMEOUT but if our connection breaks afterward it gets tricky. 

We emit a metric now though `controller_sync_errors_total` (probably need to go back and add `pilot_`) that can be helpful. The [way we collect this](https://github.com/kubernetes/client-go/blob/6711973af1c55d544f08252297ebf6dab2031975/tools/cache/shared_informer.go#L169-L182) doesn't provide an equiv for finding out when it becomes "healthy" again though.

howardjohn:
I didn't have any great ideas, just having two columns seemed a bit minimal so I threw in a third :slightly_smiling_face:. We can definitely leave it out if we dont have a a good way to measure it

PR #33804, opened 2021-07-02
add clusterz multicluster debug endpoint

stevenctl:
related to https://github.com/istio/istio/issues/33799, not sure if the "sync status" is the best – timeout can be confusing and technically can switch from timeout -> synced.

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/multicluster.go
+++ b/pilot/pkg/serviceregistry/kube/controller/multicluster.go
@@ -359,10 +359,11 @@ func (m *Multicluster) GetRemoteKubeClient(clusterID cluster.ID) kubernetes.Inte
 	return nil
 }
 
-func (m *Multicluster) InitSecretController(stop <-chan struct{}) {
+func (m *Multicluster) InitSecretController(stop <-chan struct{}) *secretcontroller.Controller {
 	m.secretController = secretcontroller.StartSecretController(
 		m.client, m.AddMemberCluster, m.UpdateMemberCluster, m.DeleteMemberCluster,
 		m.secretNamespace, m.syncInterval, stop)
+	return m.secretController
 }
 
 func (m *Multicluster) HasSynced() bool {
--- a/pilot/pkg/xds/debug.go
+++ b/pilot/pkg/xds/debug.go
@@ -198,6 +198,7 @@ func (s *DiscoveryServer) AddDebugHandlers(mux, internalMux *http.ServeMux, enab
 
 	s.addDebugHandler(mux, internalMux, "/debug/inject", "Active inject template", s.InjectTemplateHandler(webhook))
 	s.addDebugHandler(mux, internalMux, "/debug/mesh", "Active mesh config", s.MeshHandler)
+	s.addDebugHandler(mux, internalMux, "/debug/clusterz", "List remote clusters where istiod reads endpoints", s.clusterz)
 	s.addDebugHandler(mux, internalMux, "/debug/networkz", "List cross-network gateways", s.networkz)
 	s.addDebugHandler(mux, internalMux, "/debug/exportz", "List endpoints that been exported via MCS", s.exportz)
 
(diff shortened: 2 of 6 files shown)


pilot/pkg/xds/discovery.go
```diff
@@ -15,6 +15,7 @@
 package xds
 
 import (
+	"istio.io/istio/pilot/pkg/serviceregistry/kube/controller"
```
nmittler, line 18: nit: move this import

pkg/kube/secretcontroller/secretcontroller.go
```diff
@@ -296,6 +318,13 @@ func (c *Controller) HasSynced() bool {
 	return synced
 }
 
+func (c *Controller) Clusters() map[string]map[cluster.ID]*Cluster {
+	if c == nil {
```
nmittler, line 322: why would `c` be `nil`?

pilot/pkg/xds/discovery.go
```diff
@@ -143,7 +144,11 @@ type DiscoveryServer struct {
 	Cache model.XdsCache
 
 	// JwtKeyResolver holds a reference to the JWT key resolver instance.
-	JwtKeyResolver *model.JwksResolver
+	JwtKeyResolver         *model.JwksResolver
```
nmittler, line 147: nit: spacing probably doesn't pass lint

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -369,3 +369,11 @@ func (m *Multicluster) InitSecretController(stop <-chan struct{}) {
 func (m *Multicluster) HasSynced() bool {
 	return m.secretController.HasSynced()
 }
+
+// TODO remove this and talk directly to top-level secretcontroller
```
nmittler, line 373: nit: add your name to the todo: `TODO(stevenctl): Remove ...`

pilot/pkg/util/informermetric/informerutil.go
```diff
@@ -28,7 +28,7 @@ var (
 	clusterLabel = monitoring.MustCreateLabel("cluster")
 
 	errorMetric = monitoring.NewSum(
-		"controller_sync_errors_total",
+		"pilot_controller_sync_errors_total",
```
howardjohn, line 31: did we already ship this? It may be backwards incompatible to change it now

pilot/pkg/xds/debug.go
```diff
@@ -864,6 +864,27 @@ func (s *DiscoveryServer) exportz(w http.ResponseWriter, _ *http.Request) {
 	writeJSON(w, jsonMap)
 }
 
+func (s *DiscoveryServer) clusterz(w http.ResponseWriter, req *http.Request) {
+	var out []map[string]interface{}
```
howardjohn, line 868: very nit: I would create a small struct for this

howardjohn (commented):
Nice!

pilot/pkg/bootstrap/server.go
```diff
@@ -117,7 +117,7 @@ type Server struct {
 	kubeClient kubelib.Client
 
 	multicluster      *kubecontroller.Multicluster
-	secretsController *kubesecrets.Multicluster
+	secretsController *kubesecrets.Multiclust[er
```
ramaraochavali, line 120: nit:  unnecessary [

ramaraochavali:
Nice Improvement!!!

pkg/kube/secretcontroller/secretcontroller.go
stevenctl: It should be impossible, but I think you can hit debug endpoints before this stuff gets fully initialized.

howardjohn (approved):

nmittler (approved):

stevenctl:
https://github.com/istio/istio/pull/33804/commits/c86312f6cc0e5a12bd62f21efdd93161fd7389bb is needed to avoid the `xds` package pulling in a bunch of stuff it doesn't need and inflating istio-agent binary size.

hzxuzhonghu (approved):

howardjohn:
The unit test failure means you must be importing a k8s library in the agent now

stevenctl:
Yeah.. likely need to add some simpler map/struct or function ref for DiscoveryServer to hold and Secret controller to update.. feels hacky but I can't think of a cleaner or more OO way

nmittler:
Yeah I had to do something similar when I added exportz

lei-tang (approved):

PR #33805, opened 2021-07-02
Decrease flush interval of access logs

howardjohn:
This is a UX improvement, by making access logs more responsive. There
is a tradeoff in responsiveness in performance; we could go even lower
but I feel 1s strikes the best balance (10s is the default).

The max buffer size is 64k. Each log is ~256 bytes. This gives us 263 logs per flush if the internal doesn't triger it
If we have 1 QPS: we would write 6/minute before and 60/minute after
If we have 10 QPS: we would write 22/minute before and 60/minute after
If we have 100 QPS: we would write 227/minute before and after
If we have 1000 QPS: we would write 2270/minute before and after

So this has no impact on users with >250 QPS.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/envoy/proxy.go
+++ b/pkg/envoy/proxy.go
@@ -124,6 +124,13 @@ func (e *envoy) args(fname string, epoch int, bootstrapConfig string) []string {
 		"--parent-shutdown-time-s", fmt.Sprint(int(convertDuration(e.ParentShutdownDuration) / time.Second)),
 		"--local-address-ip-version", proxyLocalAddressType,
 		"--bootstrap-version", "3",
+		// Reduce default flush interval from 10s to 1s. The access log buffer size is 64k and each log is ~256 bytes
+		// This means access logs will be written once we have ~250 requests, or ever 1s, which ever comes first.
+		// Reducing this to 1s optimizes for UX while retaining performance.
+		// At low QPS access logs are unlikely a bottleneck, and these users will now see logs after 1s rather than 10s.
+		// At high QPS (>250 QPS) we will log the same amount as we will log due to exceeding buffer size, rather
+		// than the flush interval.
+		"--file-flush-interval-msec", "1000",
 		"--disable-hot-restart", // We don't use it, so disable it to simplify Envoy's logic
 	}
 	if e.ProxyConfig.LogAsJSON {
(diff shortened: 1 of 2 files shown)


pkg/envoy/proxy.go
```diff
@@ -124,6 +124,13 @@ func (e *envoy) args(fname string, epoch int, bootstrapConfig string) []string {
 		"--parent-shutdown-time-s", fmt.Sprint(int(convertDuration(e.ParentShutdownDuration) / time.Second)),
 		"--local-address-ip-version", proxyLocalAddressType,
 		"--bootstrap-version", "3",
+		// Reduce default flush interval from 10s to 1s. The access log buffer size is 64k and each log is ~256 bytes
+		// This means access logs will be written once we have ~250 requests, or ever 1s, which ever comes first.
+		// Reducing this to 1s optimizes for UX while retaining performance.
+		// At low QPS access logs are unlikely a bottleneck, and these users will now see logs after 1s rather than 10s.
+		// At high QPS (>250 QPS) we will log the same amount as we will log due to exceeding buffer size, rather
+		// than the flush interval.
+		"--file-flush-interval-msec", "1000",
```
hzxuzhonghu, line 133: Will this influence grpc acls? I feel we should also leave this option to users

hzxuzhonghu: Found, the grpc flush interval default is 1s. This option does not influence grpc al.

hzxuzhonghu (approved):

pkg/envoy/proxy.go
```diff
@@ -124,6 +124,13 @@ func (e *envoy) args(fname string, epoch int, bootstrapConfig string) []string {
 		"--parent-shutdown-time-s", fmt.Sprint(int(convertDuration(e.ParentShutdownDuration) / time.Second)),
 		"--local-address-ip-version", proxyLocalAddressType,
 		"--bootstrap-version", "3",
+		// Reduce default flush interval from 10s to 1s. The access log buffer size is 64k and each log is ~256 bytes
+		// This means access logs will be written once we have ~250 requests, or ever 1s, which ever comes first.
```
ramaraochavali, line 128: nit: /s/ever/every

ramaraochavali:
lgtm other than minor comment

Issue #32901, opened 2021-05-17
Combine istio-iptables binary into istio-cni


bianpengyuan:
Currently istio-cni and istio-iptables are two separated binaries. This unnecessarily increases size of istio-cni container, and make test hard. Instead of exec istio-iptables from istio-cni, the iptables logic can be built into istio-cni directly.

bianpengyuan:
Two items left to close this:
 1) Instead of call iptables via cobra command and set param via viper, extract out iptables command logic to a top level function and CNI could call directly with configurations struct as input.
 2) Remove os.Exit from iptables, which makes sure iptables log can be always reported to install-cni container.

PR #33806, opened 2021-07-02
Build istio-iptables into istio-cni.

bianpengyuan:
For #32901. Also add a file output path param to iptables, which will be used for CNI dry run test in later PRs.

### DIFF
--- a/cni/pkg/plugin/iptables.go
+++ b/cni/pkg/plugin/iptables.go
@@ -17,14 +17,16 @@
 package plugin
 
 import (
-	"fmt"
-	"os/exec"
-	"strings"
+	"github.com/spf13/viper"
 
-	"istio.io/pkg/log"
+	"istio.io/istio/tools/istio-iptables/pkg/cmd"
+	"istio.io/istio/tools/istio-iptables/pkg/constants"
+	"istio.io/pkg/env"
 )
 
-var nsSetupProg = "istio-iptables"
+var (
+	dryRunFilePath = env.RegisterStringVar("DRY_RUN_FILE_PATH", "", "")
+)
 
 type iptables struct{}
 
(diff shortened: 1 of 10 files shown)


cni/pkg/plugin/iptables.go
```diff
@@ -35,32 +38,28 @@ func newIPTables() InterceptRuleMgr {
...
-		"-m", rdrct.redirectMode,
-		"-i", rdrct.includeIPCidrs,
-		"-b", rdrct.includePorts,
-		"-d", rdrct.excludeInboundPorts,
-		"-o", rdrct.excludeOutboundPorts,
-		"-x", rdrct.excludeIPCidrs,
-		"-k", rdrct.kubevirtInterfaces,
-	}
+	viper.Set(constants.CNIMode, true)
+	viper.Set(constants.NetworkNamespace, netns)
```
howardjohn, line 42: is this safe to mutate? Seems like we are mutating global variables from within normal function. Does CNI call this multiple times? Or even worse, in parallel?

tools/istio-iptables/pkg/validation/validator.go
```diff
@@ -23,6 +23,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/prometheus/common/log"
```
howardjohn, line 26: wrong log package :slightly_smiling_face:

tools/istio-iptables/pkg/dependencies/implementation.go
```diff
@@ -60,12 +61,24 @@ var XTablesCmds = sets.NewSet(
...
+	log.Infof("%s %s\n", cmd, strings.Join(args, " "))
+	if r.CNIMode {
+		originalCmd := cmd
+		cmd = constants.NSENTER
+		args = append([]string{fmt.Sprintf("--net=%v", r.NetworkNamespace), "--", originalCmd}, args...)
+	}
 	externalCommand := exec.Command(cmd, args...)
-	externalCommand.Stdout = os.Stdout
+	if r.CNIMode {
+		externalCommand.Stdout = os.Stderr
```
howardjohn, line 78: Its not clear to me why its different, could use a comment

howardjohn (commented):
I know it is marked WIP but overall LGTM :+1:

tools/istio-iptables/pkg/validation/validator.go
bianpengyuan: Thanks! This is unrelated changes to this PR. Removed.

cni/pkg/plugin/iptables.go
bianpengyuan: Yeah, for every new pod, kubelet will exec a new CNI network plugin process. This function will only be executed once per process.

tools/istio-iptables/pkg/dependencies/implementation.go
bianpengyuan: Sorry it is debug left over. I was trying to figure out why CNI network configuration fails during development.

howardjohn (approved):

Issue #31936, opened 2021-04-06
Istio CNI: Retry if iptables programming fails [xtables locked]


mandarjog:
Istio CNI programs iptables rules inside pod network namespace.
Based how many pods are starting and how many other tools are reprogramming iptables there may be contention on the xtables lock.

When this happens we see 
```
nsenter failed, out: ...     Another app is currently holding the xtables lock. Perhaps you want to use the -w option
```
note: `iptables-restore` does not support `-w` options suggested by the error message. `iptables` itself supports it.

We need to ensure that we put `nsenter iptables-restore` in a backoff retry loop to ensure that we do not fail on retryable error.
`xtables lock` in the error message is a very good indicator of retryability.

mandarjog:
@therealmitchconnors for Beta graduation.

howardjohn:
see https://github.com/istio/istio/pull/30394 as well

mandarjog:
@bianpengyuan since you are fixing a bunch of other cni issues can you take this too?
The approach in the attached pr does not work since there is no “-w” option in iptables-restore so we just need to put it in a back off loop. There is a simple back off available in the k8s library.

mandarjog:
https://pkg.go.dev/k8s.io/client-go/util/retry#OnError

PR #33810, opened 2021-07-03
Add retry to iptables command for xtables lock contention.

bianpengyuan:
fix https://github.com/istio/istio/issues/31936. To verify this, I tested the change with this commit https://github.com/bianpengyuan/istio/commit/11a99106126237555a7e9484b25393eb94c0efce, which proves retry logic is indeed triggered.


tools/istio-iptables/pkg/dependencies/implementation.go
```diff
@@ -82,15 +83,34 @@ func (r *RealDependencies) execute(cmd string, ignoreErrors bool, args ...string
...
 	externalCommand := exec.Command(cmd, args...)
 	stdout := &bytes.Buffer{}
 	stderr := &bytes.Buffer{}
 	externalCommand.Stdout = stdout
 	externalCommand.Stderr = stderr
 
-	err := externalCommand.Run()
+	for i := 0; i < 10; i++ {
+		stderr.Reset()
+		err = externalCommand.Run()
```
howardjohn, line 96: are commands re-runnable like this? I would have thought we had to reconstruct it?

It may (or may not, not a big issue) be slightly simpler to do that anyways to avoid the resets on the buffers

tools/istio-iptables/pkg/dependencies/implementation.go
```diff
@@ -82,15 +83,34 @@ func (r *RealDependencies) execute(cmd string, ignoreErrors bool, args ...string
...
+		}
+
+		if !isXTablesLockError(stderr, exitCode) {
+			// Command succeeded, or failed not because of xtables lock.
+			break
+		}
+
+		// If command failed because xtables was locked, try the command again.
+		// Note we retry invoking iptables command explicitly instead of using the `-w` option of iptables,
+		// because as of iptables 1.6.x (version shipped with bionic), iptables-restore does not support `-w`.
```
howardjohn, line 110: We are on 1.8 now in master branch, so for docker images we would potentially use it. But I guess that does not work for VMs and CNI which use host iptables.

howardjohn (commented):
Overall looks good, just a few small nits

tools/istio-iptables/pkg/dependencies/implementation.go
```diff
@@ -82,15 +83,34 @@ func (r *RealDependencies) execute(cmd string, ignoreErrors bool, args ...string
...
 	log.Infof("Running command: %s %s", cmd, strings.Join(args, " "))
 	externalCommand := exec.Command(cmd, args...)
 	stdout := &bytes.Buffer{}
 	stderr := &bytes.Buffer{}
 	externalCommand.Stdout = stdout
 	externalCommand.Stderr = stderr
 
-	err := externalCommand.Run()
+	for i := 0; i < 10; i++ {
+		stderr.Reset()
```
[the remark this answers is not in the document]
howardjohn: I mean should we reset both

howardjohn (approved):

howardjohn:
LGTM but put a hold to allow @mandarjog or @luksa to take a look if they are interested

tools/istio-iptables/pkg/dependencies/implementation.go
bianpengyuan: Good point! Done.

tools/istio-iptables/pkg/dependencies/implementation.go
bianpengyuan: Ok sg. I can do that in a follow up.

tools/istio-iptables/pkg/dependencies/implementation.go
```diff
@@ -82,15 +83,34 @@ func (r *RealDependencies) execute(cmd string, ignoreErrors bool, args ...string
...
+
+		if !isXTablesLockError(stderr, exitCode) {
+			// Command succeeded, or failed not because of xtables lock.
+			break
+		}
+
+		// If command failed because xtables was locked, try the command again.
+		// Note we retry invoking iptables command explicitly instead of using the `-w` option of iptables,
+		// because as of iptables 1.6.x (version shipped with bionic), iptables-restore does not support `-w`.
+		time.Sleep(100 * time.Millisecond)
```
mandarjog, line 111: Add jitter?

tools/istio-iptables/pkg/dependencies/implementation_test.go
```diff
@@ -0,0 +1,62 @@
...
+func TestIsXTablesLockError(t *testing.T) {
+	xtableError := []struct {
+		name    string
+		errMsg  string
+		errCode XTablesExittype
+		want    bool
+	}{
+		{
+			name:    "no error",
+			errMsg:  "Another app is currently holding the xtables lock.",
```
mandarjog, line 31: The difference between this and the last test is error code.
However the error message is the same, that does not look right.

tools/istio-iptables/pkg/dependencies/implementation.go
bianpengyuan: Added retry backoff. It uses 0.5 as randomization factor.

tools/istio-iptables/pkg/dependencies/implementation_test.go
bianpengyuan: Yeah this won't happen in reality. Updated.

tools/istio-iptables/pkg/dependencies/implementation.go
```diff
@@ -90,20 +93,42 @@ func (r *RealDependencies) execute(cmd string, ignoreErrors bool, args ...string
...
 	log.Infof("Running command: %s %s", cmd, strings.Join(args, " "))
-	externalCommand := exec.Command(cmd, args...)
-	stdout := &bytes.Buffer{}
-	stderr := &bytes.Buffer{}
-	externalCommand.Stdout = stdout
-	externalCommand.Stderr = stderr
 
-	err := externalCommand.Run()
+	var stdout, stderr *bytes.Buffer
+	b := backoff.NewExponentialBackOff()
```
mandarjog, line 105: What is the max interval ?
If you are using backoff lib, you can just use the retry function in it which is imo cleaner.

tools/istio-iptables/pkg/dependencies/implementation.go
```diff
@@ -90,20 +93,43 @@ func (r *RealDependencies) execute(cmd string, ignoreErrors bool, args ...string
...
+
+		if !isXTablesLockError(stderr, exitCode) {
+			// Command succeeded, or failed not because of xtables lock.
+			break
+		}
+
+		// If command failed because xtables was locked, try the command again.
+		// Note we retry invoking iptables command explicitly instead of using the `-w` option of iptables,
+		// because as of iptables 1.6.x (version shipped with bionic), iptables-restore does not support `-w`.
+		time.Sleep(b.NextBackOff())
```
mandarjog, line 130: on the last re-try,  this code will sleep and then exit the loop.

I think `backoff.Retry` or ``backoff.RetryWithTimeout` will do the right thing.

bianpengyuan: Updated to use backoff.Retry, with max elapse time as 10s and max interval 2s.

tools/istio-iptables/pkg/dependencies/implementation.go
```diff
@@ -90,20 +94,44 @@ func (r *RealDependencies) execute(cmd string, ignoreErrors bool, args ...string
...
-	externalCommand.Stderr = stderr
 
-	err := externalCommand.Run()
+	var stdout, stderr *bytes.Buffer
+
+	b := backoff.NewExponentialBackOff()
+	b.InitialInterval = 100 * time.Millisecond
+	b.MaxInterval = 2 * time.Second
+	b.MaxElapsedTime = 10 * time.Second
+	_ = backoff.Retry(func() error {
```
mandarjog, line 111: `err := ` or `err =`, then we don't need to have names return var.

Retry should return an error if max retry timeout has elapsed, or if we get a permanent error.

mandarjog (approved):
apart from a nit, lgtm

tools/istio-iptables/pkg/dependencies/implementation.go
bianpengyuan: Done.

PR #32163, opened 2021-04-14
wasm pkg: add OCI packaged Wasm binary fetcher.

mathetake:
Added the Proxy-Wasm OCI image fetcher in wasm pkg. The OCI image is only specified [here](https://docs.google.com/document/d/1Uf-3RPFRyNl6ExYUeZERlB9Y8kyxZFaB6NbDZyguXwI/edit).

the actual use case of this fetcher will come with the implementation of Wasm Extension API(https://github.com/istio/api/pull/1940).

Signed-off-by: Takeshi Yoneda <takeshi@tetrate.io>

PR #29991, opened 2021-01-12
Add a Wasm pkg with a http fetcher and a local file cache.

bianpengyuan:
ref: https://github.com/istio/istio/issues/29989

PR #33812, opened 2021-07-03
pkg/wasm: add Docker/OCI image fetcher.

mathetake:
This PR adds the fetcher implementation of [Wasm Image Specification](https://github.com/solo-io/wasm/tree/master/spec) which Wasm-SIG has worked on, and this is intended to be used in the implementation of the new [Wasm Plugin API](https://github.com/istio/api/pull/1940).

This is the sequel to httpfetcher (https://github.com/istio/istio/pull/29991), and the rework of the previous PR https://github.com/istio/istio/pull/32163.

Signed-off-by: Takeshi Yoneda <takeshi@tetrate.io>

cc @bianpengyuan @dgn @mandarjog


pkg/wasm/imagefetcher.go
```diff
@@ -0,0 +1,238 @@
...
+	} else {
+		fetchOpt = remote.WithAuth(&authn.Basic{Username: opt.Username})
+	}
+	return &ImageFetcher{
+		fetchOpt: fetchOpt,
+	}
+}
+
+// Fetch is the entrypoint after variant detection the image's manifest annotations.
+func (o *ImageFetcher) Fetch(url string) ([]byte, error) {
```
mathetake, line 64: @dgn this will be the entry point for your API implementation, so please put your comments here!

pkg/wasm/imagefetcher.go
```diff
@@ -0,0 +1,238 @@
...
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+)
+
+const (
+	wasmPluginFileName = "plugin.wasm"
```
howardjohn, line 31: nit: if there is an spec/doc indicating this magic file name, would be good to link here

mathetake: yeah right, thanks.

mathetake:
OK test passes besides I have no idea about what I can do for this, 

> ERROR: Some modules have unrecognized licenses:
  github.com/opencontainers/runc: path '/home/prow/go/pkg/mod/github.com/opencontainers/runc@v0.1.1/Godeps/_workspace/src/github.com/golang/protobuf/LICENSE'

howardjohn:
> OK test passes besides I have no idea about what I can do for this,
> 
> > ERROR: Some modules have unrecognized licenses:
> > github.com/opencontainers/runc: path '/home/prow/go/pkg/mod/github.com/opencontainers/runc@v0.1.1/Godeps/_workspace/src/github.com/golang/protobuf/LICENSE'

Basically we are adding a new dependency on runc implicitly. Somehow our tool fails to detect https://github.com/opencontainers/runc/blob/v0.1.1/Godeps/_workspace/src/github.com/golang/protobuf/LICENSE is a BSD license.

Interestingly, through a chain of old dependencies, we end up importing runc for 4 years ago...

In the short term, we can add an exception in https://github.com/istio/common-files/blob/master/files/common/config/license-lint.yml

howardjohn:
https://github.com/helm/helm/pull/9675/files may help but I don't think it will

mathetake:
Thanks, I added the exception in license-lint.yaml for now

howardjohn:
The file is synced from istio/common-files, can you send a PR to add it
there as well? otherwise it will get reverted when the next sync occurs.
Thanks!

mathetake:
just created the PR https://github.com/istio/common-files/pull/445, Thanks!

pkg/wasm/imagefetcher.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch image: %v", err)
+	}
+
+	manifest, err := img.Manifest()
+	if err != nil {
+		return nil, fmt.Errorf("could not retrieve manifest: %v", err)
+	}
+
+	if manifest.MediaType == types.DockerManifestSchema2 {
```
bianpengyuan, line 81: Can we add some info log when trying to parse each variant of image? So that if there is any error, we can map the error message returned with the image variant.

bianpengyuan (approved):

pkg/wasm/imagefetcher.go
mathetake: all the returned error contains the information which variant is tried to parse and how it failed (L87 and L105), so maybe it is redundant for emitting log vs emitting error message at callsite (API implementation). 

FWIW, parsing as OCI images is tricky because we cannot know the variant in advance and instead have to do try-catch style parsing so that's the reason why I use the multi error in L105. https://github.com/istio/istio/blob/8d854173484b6e39d3cd9cfdc5df8b8aeaaa05d5/pkg/wasm/imagefetcher_test.go#L244-L246

In any case, I think it would be simpler to allow callers to emit errors or not in one place after calling `Fetch`, given that errors contain per-variant errors(L105) or docker-variant error (L87). 
Also if this doesn't meet the requirement from API impl side, it is easy to revisit this later:) 

sg?

bianpengyuan: Ah yeah, make sense thanks!

mandarjog (approved):
lgtm

howardjohn (approved):
Approving new dependency - other files are already approved by E&T team

mathetake:
@mandarjog good to go?

dgn (approved):

mathetake:
😕

bianpengyuan:
@brian-avery any idea about how to fix release notes check failure?

Issue #33146, opened 2021-05-27
Sidecar in VM fails to start due to "failed to warm certificate: failed to generate workload certificate: create certificate"


liorokman:
**Bug description**

   I installed the Istio sidecar into a virtual machine, as specified [in the documentation](https://istio.io/latest/docs/setup/install/virtual-machine/). 

   The virtual machine connected successfully for 1 day, and on the next day, the envoy log on the VM started filling up with the following errors:

```
2021-05-27T10:28:07.342344Z     warn    ca      ca request failed, starting attempt 1 in 91.589072ms
2021-05-27T10:28:07.434806Z     warn    ca      ca request failed, starting attempt 2 in 203.792343ms
2021-05-27T10:28:07.639557Z     warn    ca      ca request failed, starting attempt 3 in 364.729652ms
2021-05-27T10:28:08.005300Z     warn    ca      ca request failed, starting attempt 4 in 830.723933ms
2021-05-27T10:28:08.836493Z     warn    sds     failed to warm certificate: failed to generate workload certificate: create certificate: rpc error: code = Unavailable desc = connection closed
```

The VM was up and running the entire time, except that I periodically (for various tests) restarted the istio service on the VM using `systemctl`.

I don't know if this is relevant, but during this time period, the Kubernetes cluster running Istio was hibernated, and Istio was only restarted after the certificate in `/etc/certs/cert-bundle.pem` had already expired.

At this point restarting the istio service using `systemctl` on the VM did not fix the problem.

I was able to get the environment working only by manually deleting (as root) the `/etc/certs/key.pem` and `/etc/certs/cert-chain.pem` files. Once these files were deleted, the next time I restarted the istio service using `systemctl` everything connected correctly.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ X] Security
[ ] Test and Release
[ X] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

The envoy process on the VM should successfully connect to the pilot-agent and retrieve certificates. If there is an existing `cert-chain.pem`  that is not valid, it should be ignored and the sidecar should behave as if this file did not exist.

**Steps to reproduce the bug**

Connect a virtual machine sidecar so that the virtual machine sidecar retrieves a certificate into `/etc/certs/cert-chain.pem`, and make Istio unavailable at the point in time when the sidecar certificate needs to be rotated, so that the sidecar certificate expires without being rotated.

[...]

PR #33813, opened 2021-07-03
fix cert warm failure with tests

williamaronli:
ref: https://github.com/istio/istio/issues/33146

test result:

2021-07-11T20:50:49.594881Z     info    JWT policy is third-party-jwt
2021-07-11T20:50:49.597638Z     info    CA Endpoint istiod.istio-system.svc:15012, provider Citadel
2021-07-11T20:50:49.597669Z     info    Using CA istiod.istio-system.svc:15012 cert with certs: /etc/certs/root-cert.pem
2021-07-11T20:50:49.597674Z     info    sssssssssssssssssssssssss
2021-07-11T20:50:49.597780Z     info    citadelclient   Citadel client using custom root cert: istiod.istio-system.svc:15012
2021-07-11T20:50:49.613607Z     info    Opening status port 15020
2021-07-11T20:50:49.649332Z     info    ads     All caches have been synced up in 62.466585ms, marking server ready
2021-07-11T20:50:49.649807Z     info    sds     SDS server for workload certificates started, listening on "etc/istio/proxy/SDS"
2021-07-11T20:50:49.661775Z     info    sds     Start SDS grpc server
2021-07-11T20:50:49.661899Z     debug   dns     initialized DNS search=[c.williamliistiotest.internal. google.internal.] servers=[169.254.169.254:53]
2021-07-11T20:50:49.662161Z     info    xdsproxy        Initializing with upstream address "istiod.istio-system.svc:15012" and cluster "Kubernetes"
2021-07-11T20:50:49.662462Z     info    Pilot SAN: [istiod.istio-system.svc]
2021-07-11T20:50:49.662874Z     info    dns     Starting local udp DNS server at localhost:15053
2021-07-11T20:50:49.662918Z     info    dns     Starting local tcp DNS server at localhost:15053
2021-07-11T20:50:49.662984Z     info    starting Http service at 127.0.0.1:15004
2021-07-11T20:50:49.704073Z     info    ads     ADS: new connection for node:vmwilliam.foo-1
**2021-07-11T20:50:49.723052Z     warn    citadelclient   cert expired, using token instead**
2021-07-11T20:50:49.845128Z     info    citadelclient   Citadel client using custom root cert: istiod.istio-system.svc:15012

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[X] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


security/pkg/nodeagent/caclient/providers/citadel/client.go
```diff
@@ -127,13 +130,24 @@ func (c *CitadelClient) getTLSDialOption() (grpc.DialOption, error) {
...
 					return &certificate, nil
 				}
+				var isExpired bool
+				isExpired, err = c.isCertExpired(filepath.Join(c.opts.ProvCert, "cert-chain.pem"))
+				if err != nil {
+					citadelClientLog.Warnf("cannot parse the cert chain, using token instead: %v", err)
+					return &tls.Certificate{}, nil
+				}
+				if isExpired {
+					citadelClientLog.Warnf("cert expired, using token instead")
```
howardjohn, line 148: I think that Istiod should allow mTLS cert *or* token. Its not the clients responsibility to ensure it sends only valid credentials, Istiod should be modified to properly use OR for auth, instead of short-circuiting when the TLS cert is invalid

security/pkg/nodeagent/caclient/providers/citadel/testdata/notexpired-cert.pem
```diff
@@ -0,0 +1,34 @@
+-----BEGIN CERTIFICATE-----
```
howardjohn, line 1: This will expire in 1 month, causing all unit tests to fail. Please make it 10+ years, and add a comment indicating the command used to generate it

howardjohn (changes requested):

security/pkg/nodeagent/caclient/providers/citadel/client.go
williamaronli: So the logic here just add that. if the certificate is expire we will check the token instead

security/pkg/nodeagent/caclient/providers/citadel/testdata/notexpired-cert.pem
williamaronli: ok I will change it

security/pkg/nodeagent/caclient/providers/citadel/client.go
howardjohn: I mean on the server side in Istiod. This is just the client side

williamaronli: server side logic is much difficult to change. As it will only check one of the token or cert. And in our original logic, we will only attach the token when the cert is empty in the CSR

howardjohn: I think its more correct behavior, it should not be much more complex, we just need to modify the TLs config on the server (https://github.com/istio/istio/blob/cd4e73cd8fdca373d89ae1b8175a1b0c2d4d0658/pilot/pkg/bootstrap/server.go#L749) instead of on the client.

Authentication already checks for one of the auth methods working (not all): https://github.com/istio-testing/istio/blob/6fa684aaef46264a68d4ef377b2bbb433bc87971/pilot/pkg/xds/authentication.go#L56. The problem is we short-circuit at the TLS level

liminw: I think Istiod is already validating cert and reject it if cert is expired. This PR is to fix client side behavior. With valid token and expired cert, token should be used instead. John, are you suggesting that client should always send all credentials and let Istiod decide which one to use?

howardjohn: Yes

liminw: This sounds good to me. Thanks.

williamaronli: I have just checked the code howard mention
Current the failure for this is our tls mode choose the 
`VerifyClientCertIfGiven` Type
https://github.com/williamaronli/istio/blob/cd4e73cd8fdca373d89ae1b8175a1b0c2d4d0658/pilot/pkg/bootstrap/server.go#L749

And tls failed when they receive an expired cert according to:

```
// VerifyClientCertIfGiven indicates that a client certificate should be requested
	// during the handshake, but does not require that the client sends a
	// certificate. If the client does send a certificate it is required to be
	// valid.
	VerifyClientCertIfGiven
```
So when client side send a expired cert to it the tls will fail.


So there may be two solutions:
1. I change it to other options like :
```
// RequireAnyClientCert indicates that a client certificate should be requested
	// during the handshake, and that at least one certificate is required to be
	// sent by the client, but that certificate is not required to be valid.
	RequireAnyClientCert
```
So if changes this tls will not verify the cert, our authentication function will verify it, as right now the csr will contain both token and cert
https://github.com/istio-testing/istio/blob/6fa684aaef46264a68d4ef377b2bbb433bc87971/pilot/pkg/xds/authentication.go#L56

however this may change the security level on server side

2. Another solution is just like this PR:
We check the cert's validation in client side and if it is not valid we don't include that cert in the CSR

@howardjohn is my understanding correct?

howardjohn (approved):

PR #33816, opened 2021-07-03
Fuzzing: Add second config validation fuzzer

AdamKorcz:
Adds a second fuzzer for config validation. The fuzzer targets the same API but its approach to creating a fuzzed config is a bit different in that it utilizes Istio APIs to generate a `Spec` from json.

I suggest we run both continuously and compare their performance.

________________________________________

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


tests/fuzz/config_validation_fuzzer.go
```diff
@@ -69,9 +69,43 @@ func FuzzConfigValidation(data []byte) int {
 	}
 
 	iobj = crdclient.TranslateObject(object, gvk, "cluster.local")
-	_, err = r.Resource().ValidateConfig(*iobj)
+	_, _ = r.Resource().ValidateConfig(*iobj)
+	return 1
+}
+
+func FuzzConfigValidation2(data []byte) int {
```
howardjohn, line 76: Would be good to have a comment explaining why we have two. The PR description would be great

howardjohn (approved):

tests/fuzz/config_validation_fuzzer.go
AdamKorcz: Sure, let me just add that before it gets merged.

PR #33818, opened 2021-07-04
fix multiple clusters issue with external istiod setup

ramaraochavali:
There is a bug because of which we are not able to add more than 1 remote cluster in External Istiod mode. This PR fixes it and adds tests for it.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/multicluster.go
+++ b/pilot/pkg/serviceregistry/kube/controller/multicluster.go
@@ -256,8 +256,7 @@ func (m *Multicluster) AddMemberCluster(clusterID cluster.ID, rc *secretcontroll
 		}
 		// Patch validation webhook cert
 		if m.caBundleWatcher != nil {
-			controller.NewValidatingWebhookController(client, m.revision,
-				m.secretNamespace, m.caBundleWatcher).Run(clusterStopCh)
+			controller.NewValidatingWebhookController(client, m.revision, m.secretNamespace, m.caBundleWatcher).Run(clusterStopCh)
 		}
 	}
 
--- a/pilot/pkg/serviceregistry/kube/controller/multicluster.go
+++ b/pilot/pkg/serviceregistry/kube/controller/multicluster.go
@@ -256,7 +256,7 @@ func (m *Multicluster) AddMemberCluster(clusterID cluster.ID, rc *secretcontroll
 		}
 		// Patch validation webhook cert
 		if m.caBundleWatcher != nil {
-			controller.NewValidatingWebhookController(client, m.revision, m.secretNamespace, m.caBundleWatcher).Run(clusterStopCh)
+			go controller.NewValidatingWebhookController(client, m.revision, m.secretNamespace, m.caBundleWatcher).Run(clusterStopCh)
 		}
 	}
 
--- a/pkg/kube/secretcontroller/secretcontroller.go
+++ b/pkg/kube/secretcontroller/secretcontroller.go
@@ -366,10 +366,10 @@ func (c *Controller) processItem(se secretEvent) error {
 		return fmt.Errorf("error fetching object %s error: %v", se.key, err)
 	}
 	if exists {
-		log.Debugf("secret %s exists in informer, processing it", se.key)
+		log.Debugf("secret %s exists in informer cache, processing it", se.key)
 		c.addSecret(se.key, obj.(*corev1.Secret))
 	} else {
-		log.Debugf("secret %s does not exiss in informer, deleting it", se.key)
+		log.Debugf("secret %s does not exist in informer cache, deleting it", se.key)
 		c.deleteSecret(se.key)
 	}
 
--- a/pkg/kube/secretcontroller/secretcontroller.go
+++ b/pkg/kube/secretcontroller/secretcontroller.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 
 	"istio.io/istio/pilot/pkg/features"
+	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pkg/cluster"
 	"istio.io/istio/pkg/kube"
 	"istio.io/pkg/log"
@@ -98,6 +99,12 @@ type Cluster struct {
 	SyncTimeout *atomic.Bool
 }
 
+// secretEvent captures the event details of secret changes.
+type secretEvent struct {
+	event model.Event
+	key   string
+}
+
 // Run starts the cluster's informers and waits for caches to sync. Once caches are synced, we mark the cluster synced.
 // This should be called after each of the handlers have registered informers, and should be run in a goroutine.
 func (r *Cluster) Run() {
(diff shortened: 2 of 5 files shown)


pkg/kube/secretcontroller/secretcontroller.go
```diff
@@ -319,42 +319,52 @@ func (c *Controller) runWorker() {
...
-		c.addSecret(key, obj.(*corev1.Secret))
+	log.Debugf("processing secret event %s for secret %s", se.event.String(), se.key)
+	if se.event == model.EventAdd || se.event == model.EventUpdate {
+		obj, exists, err := c.informer.GetIndexer().GetByKey(se.key)
+		if err != nil {
+			return fmt.Errorf("error fetching object %s error: %v", se.key, err)
+		}
+		// For add and update events, we expect secret in informer cache.
+		// we should requeue the event for later processing.
+		if !exists {
```
ramaraochavali, line 360: Not sure if this is needed but added it to test this behaviour suspecting this might be the reason when we were testing in our environment.

pkg/webhooks/validation/controller/controller.go
```diff
@@ -241,14 +241,16 @@ func newController(
 }
 
 func (c *Controller) Run(stop <-chan struct{}) {
-	defer c.queue.ShutDown()
+	go func() {
```
ramaraochavali, line 244: This is the actual fix. ValidationController is blocking and waiting for "stop". So the callback is never finished and `queue.done()` is not called and hence queue does not process the second secret.

ramaraochavali:
The actual fix is in ValidationController - But added bunch of log lines and refactored secret controller while I was testing this in our environment - which I think would be a general improvement.

pkg/kube/secretcontroller/secretcontroller.go
hzxuzhonghu: The item can not exist if it is deleted later, before this worker run, otherwise it must exist.

Add back it not right.

hzxuzhonghu: So i doubt the event is not necessary if no other intention

pkg/webhooks/validation/controller/controller.go
hzxuzhonghu: I think it is safer to modify https://github.com/istio/istio/blob/9b16ea834c7301808aaa1723e2237254ce791b7e/pilot/pkg/serviceregistry/kube/controller/multicluster.go#L259
run it in a separate goroutine

hzxuzhonghu: >and hence queue does not process the second secret.


I am confused. The queue here is to handle ValidatingWebhookConfiguration

ramaraochavali: Correct but it does not come out. It blocks the call back i.e. `AddMemberCluster`. It will never return.

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -256,8 +256,7 @@ func (m *Multicluster) AddMemberCluster(clusterID cluster.ID, rc *secretcontroll
 		}
 		// Patch validation webhook cert
 		if m.caBundleWatcher != nil {
-			controller.NewValidatingWebhookController(client, m.revision,
-				m.secretNamespace, m.caBundleWatcher).Run(clusterStopCh)
+			controller.NewValidatingWebhookController(client, m.revision, m.secretNamespace, m.caBundleWatcher).Run(clusterStopCh)
```
hzxuzhonghu, line 259: Change it to `go controller.NewValidatingWebhookController(client, m.revision, m.secretNamespace, m.caBundleWatcher).Run(clusterStopCh)` And no need to change others, which seems not very right

BTW: donot we have integration tests?

ramaraochavali: We do not have integration tests for this. I changed to Run ValidatingWebhookController in a separate go routine.

I removed requeuing the event.,

Any other changes you mean? I prefer to leave the event in the queue because we can figure out the event type easily without doing bunch of checks

hzxuzhonghu: Make sense now

hzxuzhonghu: Unbelievable this code path is not run in multi cluster tests

pkg/kube/secretcontroller/secretcontroller.go
```diff
@@ -319,42 +330,51 @@ func (c *Controller) runWorker() {
...
-
-	obj, exists, err := c.informer.GetIndexer().GetByKey(key)
-	if err != nil {
-		return fmt.Errorf("error fetching object %s error: %v", key, err)
-	}
-
-	if exists {
-		c.addSecret(key, obj.(*corev1.Secret))
+	log.Infof("processing secret event %s for secret %s", se.event.String(), se.key)
+	if se.event == model.EventAdd || se.event == model.EventUpdate {
```
hzxuzhonghu, line 364: We donot need to distinguish the event here,  for add and update, the logic is same

For delete( no object found) just call deleteSecret.

pkg/kube/secretcontroller/secretcontroller.go
```diff
@@ -319,42 +330,51 @@ func (c *Controller) runWorker() {
...
+		if err != nil {
+			return fmt.Errorf("error fetching object %s error: %v", se.key, err)
+		}
+		// For add and update events, we expect secret in informer cache.
+		// If it is not there it might be the case of it getting deleted and we are processing add/update event now.
+		if !exists {
+			log.Debugf("secret %s does not exist in informer cache, might have been deleted?", se.key)
+		}
+		log.Debugf("secret key %s exists in informer, adding/updating it", se.key)
+		c.addSecret(se.key, obj.(*corev1.Secret))
```
hzxuzhonghu, line 375: If obj = nil, when run into this function it will panic

pkg/kube/secretcontroller/secretcontroller.go
ramaraochavali: Changed

pkg/kube/secretcontroller/secretcontroller.go
ramaraochavali: fixed

pilot/pkg/serviceregistry/kube/controller/multicluster.go
stevenctl: We don't cover "istiod-less" remotes in tests currently. The "remote cluster" type we have documented still runs istiod and runs it's own webhook cert patcher.

stevenctl (approved):
Thanks for adding a test!

hzxuzhonghu (approved):

pilot/pkg/serviceregistry/kube/controller/multicluster.go
ramaraochavali: is it easy to add that test? it would be good to have one.

PR #33821, opened 2021-07-04
add retry conditions to existing virtual service tests

ramaraochavali:
Adds retry conditions to existing virtual service to validate that config is not rejected by Envoy.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/integration/pilot/common/routing.go
+++ b/tests/integration/pilot/common/routing.go
@@ -136,11 +136,11 @@ spec:
   - route:
     - destination:
         host: {{ .dstSvc }}
-	retries:
+    retries:
       attempts: 3
       perTryTimeout: 2s
-	  retryOn: gateway-error,connect-failure,refused-stream
-	  retryRemoteLocalities: true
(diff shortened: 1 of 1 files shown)


tests/integration/pilot/common/routing.go
```diff
@@ -120,6 +120,8 @@ func virtualServiceCases(skipVM bool) []TrafficTestCase {
 	})
 	var cases []TrafficTestCase
 	cases = append(cases,
+		// Retry conditions have been added to just validate that config is correct.
```
howardjohn, line 123: should we just add another case? This makes the "Added header" one more complicate

ramaraochavali: Yes. Thanks.  as usual your idea makes lot of sense :-). Moved.

howardjohn (approved):

PR #33825, opened 2021-07-05
Clean up istio-iptables log.

bianpengyuan:
Follow up PR of https://github.com/istio/istio/pull/33808. After switching iptables log to istio log pkg, several log lines do not make much sense. Change including:
* Combine iptables env and variables into a single multiple lines log entry
* Log iptables command stdout/stderr output with istio log pkg.
* Hardening the xtables error message parsing to avoid possible out-of-range crashing.

Example log output after this change: https://gist.githubusercontent.com/bianpengyuan/42e21d8fe68f2ecc675d4f2dbcb8415e/raw/f89c18766e52564fd6e88be872ef32c80f62f195/istio-init.log

### DIFF
--- a/tools/istio-iptables/pkg/dependencies/implementation.go
+++ b/tools/istio-iptables/pkg/dependencies/implementation.go
@@ -63,38 +63,47 @@ var XTablesCmds = sets.NewSet(
 // RealDependencies implementation of interface Dependencies, which is used in production
 type RealDependencies struct{}
 
-func (r *RealDependencies) execute(cmd string, redirectStdout bool, args ...string) error {
-	log.Infof("%s %s", cmd, strings.Join(args, " "))
+func (r *RealDependencies) execute(cmd string, ignoreErrors bool, args ...string) error {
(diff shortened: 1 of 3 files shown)


ramaraochavali (approved):

tools/istio-iptables/pkg/dependencies/implementation.go
```diff
@@ -63,38 +63,47 @@ var XTablesCmds = sets.NewSet(
...
-	externalCommand.Stdout = os.Stdout
-	// TODO Check naming and redirection logic
-	if !redirectStdout {
-		externalCommand.Stderr = os.Stderr
+	stdout := &bytes.Buffer{}
+	stderr := &bytes.Buffer{}
+	externalCommand.Stdout = stdout
+	externalCommand.Stderr = stderr
+
+	if len(stdout.String()) != 0 {
```
sha-rath, line 74: len of stdout would always be 0 since Run() is executed after this statement.

sha-rath (commented):
Execute Run() before printing the cmd output

PR #33833, opened 2021-07-05
Fix External should also be ProviderID

kebe7jun:
External should also be ProviderID type

### DIFF
--- a/pilot/pkg/networking/apigen/apigen.go
+++ b/pilot/pkg/networking/apigen/apigen.go
@@ -142,7 +142,7 @@ func (g *APIGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *
 		svcs := push.Services(proxy)
 		for _, s := range svcs {
 			// Ignore services that are result of conversion from ServiceEntry.
-			if s.Attributes.ServiceRegistry == serviceregistry.External {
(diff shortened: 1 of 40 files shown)


pilot/pkg/networking/apigen/apigen.go
```diff
@@ -142,7 +142,7 @@ func (g *APIGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *
 		svcs := push.Services(proxy)
 		for _, s := range svcs {
 			// Ignore services that are result of conversion from ServiceEntry.
-			if s.Attributes.ServiceRegistry == serviceregistry.External {
+			if s.Attributes.ServiceRegistry == string(serviceregistry.External) {
```
devincd, line 145: I think let `s.Attributes.ServiceRegistry` to be `ProviderID` type is more better?

hzxuzhonghu: good suggestion

kebe7jun: thanks, fixed.

hzxuzhonghu (approved):

kebe7jun:
@shamsher31 can you please take a look here?

shamsher31 (approved):
Thanks

PR #33836, opened 2021-07-05
Remove duplicate code

ChenglongGao327:
if user enable debug log, reuse query map in request

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/security/pkg/stsservice/tokenmanager/google/tokenexchangeplugin.go
+++ b/security/pkg/stsservice/tokenmanager/google/tokenexchangeplugin.go
@@ -228,22 +228,13 @@ func (p *Plugin) constructFederatedTokenRequest(parameters security.StsRequestPa
 	}
 	req.Header.Set("Content-Type", contentType)
 	if pluginLog.DebugEnabled() {
-		dQuery := map[string]string{
-			"audience":           aud,
-			"grantType":          parameters.GrantType,
-			"requestedTokenType": tokenType,
-			"subjectTokenType":   parameters.SubjectTokenType,
-			"subjectToken":       "redacted",
-			"scope":              reqScope,
-		}
-		dJSONQuery, _ := json.Marshal(dQuery)
+		query["subjectToken"] = "redacted"
+		dJSONQuery, _ := json.Marshal(query)
 		dReq, _ := http.NewRequest("POST", federatedTokenEndpoint, bytes.NewBuffer(dJSONQuery))
 		dReq.Header.Set("Content-Type", contentType)
 
-		if pluginLog.DebugEnabled() {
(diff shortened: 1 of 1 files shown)


shankgan (dismissed):

howardjohn (commented):
I actually don't think we should do this. The risk is that a future refactor will move things around, and subjectToken will now be mutated. This would probably not be detected or fail tests since it only happens at debug enabled, which is not default.

Mutability introduces bugs

shankgan:
> I actually don't think we should do this. The risk is that a future refactor will move things around, and subjectToken will now be mutated. This would probably not be detected or fail tests since it only happens at debug enabled, which is not default.
> 
> Mutability introduces bugs

Want to understand this: As far as I see, there are no changes to behavior introduced by the change. Are you saying that this code could be refactored better? Could that be done in a later PR perhaps? Any future refactor will have to change the code within the debug block with or without this change

shankgan:
> > I actually don't think we should do this. The risk is that a future refactor will move things around, and subjectToken will now be mutated. This would probably not be detected or fail tests since it only happens at debug enabled, which is not default.
> > Mutability introduces bugs
> 
> Want to understand this: As far as I see, there are no changes to behavior introduced by the change. Are you saying that this code could be refactored better? Could that be done in a later PR perhaps? Any future refactor will have to change the code within the debug block, with or without this change

sorry, my bad..I misread this..Agree with your earlier statement

security/pkg/stsservice/tokenmanager/google/tokenexchangeplugin.go
```diff
@@ -228,22 +228,13 @@ func (p *Plugin) constructFederatedTokenRequest(parameters security.StsRequestPa
...
-		dQuery := map[string]string{
-			"audience":           aud,
-			"grantType":          parameters.GrantType,
-			"requestedTokenType": tokenType,
-			"subjectTokenType":   parameters.SubjectTokenType,
-			"subjectToken":       "redacted",
-			"scope":              reqScope,
-		}
-		dJSONQuery, _ := json.Marshal(dQuery)
+		query["subjectToken"] = "redacted"
```
shankgan, line 231: nit: would be safer to copy the query map to another map and set the redacted field in that map

ChenglongGao327:
> I actually don't think we should do this. The risk is that a future refactor will move things around, and subjectToken will now be mutated. This would probably not be detected or fail tests since it only happens at debug enabled, which is not default.
> 
> Mutability introduces bugs

yeah，agree  
I just consider that these codes are duplicated

PR #33842, opened 2021-07-05
enhance dns proxy metrics

lmarszal:
Enhance DNS proxy metrics - originally filed as https://github.com/istio/istio/issues/33222, initially fixed in https://github.com/istio/istio/pull/33532

My goal is to get dns proxy metrics as close to [coredns metrics](https://coredns.io/plugins/metrics/), as reasonable. I remember we used to struggle with coredns metrics - i.e. we used to see lots of NXDOMAIN errors and were unable to tell which app is causing them. This PR would be significant improvement for me - bringing similar observability capabilities we have for http request to DNS requests.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/dns/dns.go
+++ b/pilot/pkg/dns/dns.go
@@ -176,12 +176,15 @@ func (h *LocalDNSServer) ServeDNS(proxy *dnsProxy, w dns.ResponseWriter, req *dn
 	}
 	log.Debugf("request %v", req)
 
+	defer func() {
+		requestDuration.Record(float64(time.Since(start).Milliseconds()))
+		responses.With(rcodeToLabelValue(response.Rcode)).Increment()
+	}()
+
 	if len(req.Question) == 0 {
 		response = new(dns.Msg)
 		response.SetReply(req)
 		response.Rcode = dns.RcodeServerFailure
-		requestDuration.Record(float64(time.Since(start).Milliseconds()))
-		responses.With(rcodeToLabelValue(response.Rcode)).Increment()
 		_ = w.WriteMsg(response)
 		return
 	}
@@ -193,8 +196,6 @@ func (h *LocalDNSServer) ServeDNS(proxy *dnsProxy, w dns.ResponseWriter, req *dn
 		response = new(dns.Msg)
 		response.SetReply(req)
 		response.Rcode = dns.RcodeServerFailure
-		requestDuration.Record(float64(time.Since(start).Milliseconds()))
-		responses.With(rcodeToLabelValue(response.Rcode)).Increment()
 		_ = w.WriteMsg(response)
 		return
 	}
@@ -233,8 +234,6 @@ func (h *LocalDNSServer) ServeDNS(proxy *dnsProxy, w dns.ResponseWriter, req *dn
 	// but we don't compress on the outbound, we will run into issues. For example, if the compressed
 	// size is 450 bytes but uncompressed 1000 bytes now we are outside of the non-eDNS UDP size limits
 	response.Truncate(size(proxy.protocol, req))
-	requestDuration.Record(float64(time.Since(start).Milliseconds()))
-	responses.With(rcodeToLabelValue(response.Rcode)).Increment()
 	_ = w.WriteMsg(response)
 }
 
--- a/pilot/pkg/dns/dns.go
+++ b/pilot/pkg/dns/dns.go
@@ -167,6 +167,7 @@ func (h *LocalDNSServer) UpdateLookupTable(nt *nds.NameTable) {
 // ServerDNS is the implementation of DNS interface
 func (h *LocalDNSServer) ServeDNS(proxy *dnsProxy, w dns.ResponseWriter, req *dns.Msg) {
 	requests.Increment()
+	start := time.Now()
 	var response *dns.Msg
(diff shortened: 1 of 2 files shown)


lmarszal:
CC @ramaraochavali

pilot/pkg/dns/dns.go
```diff
@@ -218,17 +223,18 @@ func (h *LocalDNSServer) ServeDNS(proxy *dnsProxy, w dns.ResponseWriter, req *dn
...
 		response = h.queryUpstream(proxy.upstreamClient, req, log)
-		requestDuration.Record(float64(time.Since(start).Milliseconds()))
+		upstreamRequestDuration.Record(float64(time.Since(start).Milliseconds()))
 		log.Debugf("upstream response for hostname %q : %v", hostname, response)
 	}
 	// Compress the response - we don't know if the incoming response was compressed or not. If it was,
 	// but we don't compress on the outbound, we will run into issues. For example, if the compressed
 	// size is 450 bytes but uncompressed 1000 bytes now we are outside of the non-eDNS UDP size limits
 	response.Truncate(size(proxy.protocol, req))
+	requestDuration.Record(float64(time.Since(start).Milliseconds()))
```
bianpengyuan, line 236: Should we use defer to record requestDuration and responses?

pilot/pkg/dns/monitoring.go
```diff
@@ -34,16 +36,39 @@ var (
...
 	monitoring.MustRegister(failures)
-	monitoring.MustRegister(requestDuration)
+	monitoring.MustRegister(upstreamRequestDuration)
+}
+
+func rcodeToLabelValue(rcode int) monitoring.LabelValue {
+	name, found := dns.RcodeToString[rcode]
+	if found {
+		return rcodeLabel.Value(name)
+	} else {
```
bianpengyuan, line 71: nit: no need to have this else block.

pilot/pkg/dns/monitoring.go
```diff
@@ -34,16 +36,39 @@ var (
...
+		"Total number of DNS responses.",
+		monitoring.WithLabels(rcodeLabel),
+	)
 )
 
 func registerStats() {
 	monitoring.MustRegister(requests)
 	monitoring.MustRegister(upstreamRequests)
 	monitoring.MustRegister(failures)
-	monitoring.MustRegister(requestDuration)
```
bianpengyuan, line 48: requestDuration metric is not registered.

pilot/pkg/dns/monitoring.go
lmarszal: done

pilot/pkg/dns/dns.go
lmarszal: done

pilot/pkg/dns/monitoring.go
lmarszal: totally forgot about these - done

ramaraochavali:
@lmarszal Thanks for this enhancement. I have made few more changes https://github.com/istio/istio/pull/33792, let us get that in and then you can rebase on top of it?

lmarszal:
> @lmarszal Thanks for this enhancement. I have made few more changes #33792, let us get that in and then you can rebase on top of it?

Sure, no rush :)

howardjohn:
Haven't had a chance to do a full review but initial thoughts - out main concern with the initial PR was the added cardinality of another histogram. Unlikely coredns where the metric is served out of one pod, this is served out of ~every pod so if can be expensive. That's probably fine if irs useful, but it's not cheap.

In real environment is this suowing useful data? I somewhat expect cache hits to all be ~1ms

lmarszal:
@howardjohn That's valid point. I'm adding two additional metrics. Let's discuss them separately:

`dns_responses_total` - sums up to `dns_requests_total`. Additional dimension (and what's useful to me) is rcode. This would allow us to detect when dns proxy starts to return errors or when app starts to make queries resulting with NXDOMAIN - which could be a symptom of some problem.

With your comment in mind, we might want to drop one of the metrics in favour of the other, but I'd like to keep additional dimension.

`dns_request_time_ms` - I agree that local requests will usually resolve in less than 1ms. And (if I understand the NDS code correctly) this metric will be flat for clusters with `REGISTRY_ONLY` outbound traffic policy for most of the time. Although for clusters with `ALLOW_ANY` outbound traffic policy it might be the case that significant portion of dns requests are forwarded upstream (I'm thinking mostly of PHP apps that can't pool database or http connections ; when such dependencies are not part of the mesh thus their name are unknown to NDS). In such cases it might come in handy to have a histogram of all dns queries.

And as previously - we can have a single histogram, but differentiate between local and upstream resolution by adding a label (e.g. `dns_request_time_ms{resolved="locally|upstream"}`) or two separate histograms (as proposed).

howardjohn:
@istio/wg-policies-and-telemetry-maintainers would appreciate some advice here

bianpengyuan:
Do we care much about the latency of DNS queries hitting the local cache? I would think probably no... I would imagine this metric will be a histogram which has majority points concentrating in the lowest bucket + duplicated information recorded by upstream latency metric.

PR #33852, opened 2021-07-06
fix mutatingwebhook patch problem

irisdingbj:
when rootCABundle rotates, mutatingwebhook fails to pick up the latest one.  This PR utilize the keycertBundleWatcher to fix the problem.

### DIFF
--- a/pkg/webhooks/webhookpatch.go
+++ b/pkg/webhooks/webhookpatch.go
@@ -39,7 +39,6 @@ import (
 var (
 	errWrongRevision     = errors.New("webhook does not belong to target revision")
 	errNoWebhookWithName = errors.New("webhook configuration did not contain webhook with target name")
-	whc                  *cache.ListWatch
 )
 
 // WebhookCertPatcher listens for webhooks on specified revision and patches their CA bundles
@@ -55,17 +54,11 @@ type WebhookCertPatcher struct {
 	// File path to the x509 certificate bundle used by the webhook server
 	// and patched into the webhook config.
 	CABundleWatcher *keycertbundle.Watcher
+	whcLw           *cache.ListWatch
 }
 
 // Run runs the WebhookCertPatcher
 func (w *WebhookCertPatcher) Run(stopChan <-chan struct{}) {
-	whc = cache.NewFilteredListWatchFromClient(
-		w.client.AdmissionregistrationV1().RESTClient(),
-		"mutatingwebhookconfigurations",
-		"",
-		func(options *metav1.ListOptions) {
-			options.LabelSelector = fmt.Sprintf("%s=%s", label.IoIstioRev.Name, w.revision)
-		})
 	go w.queue.Run(stopChan)
 	go w.runWebhookController(stopChan)
 	go w.startCaBundleWatcher(stopChan)
@@ -75,18 +68,26 @@ func (w *WebhookCertPatcher) Run(stopChan <-chan struct{}) {
 func NewWebhookCertPatcher(
 	client kubernetes.Interface,
 	revision, webhookName string, caBundleWatcher *keycertbundle.Watcher) (*WebhookCertPatcher, error) {
+	whcLw := cache.NewFilteredListWatchFromClient(
+		client.AdmissionregistrationV1().RESTClient(),
+		"mutatingwebhookconfigurations",
+		"",
+		func(options *metav1.ListOptions) {
+			options.LabelSelector = fmt.Sprintf("%s=%s", label.IoIstioRev.Name, revision)
+		})
 	return &WebhookCertPatcher{
 		client:          client,
 		revision:        revision,
 		webhookName:     webhookName,
 		CABundleWatcher: caBundleWatcher,
 		queue:           queue.NewQueue(time.Second * 2),
+		whcLw:           whcLw,
 	}, nil
 }
 
 func (w *WebhookCertPatcher) runWebhookController(stopChan <-chan struct{}) {
 	_, c := cache.NewInformer(
-		whc,
+		w.whcLw,
 		&v1.MutatingWebhookConfiguration{},
 		0,
 		cache.ResourceEventHandlerFuncs{
@@ -190,7 +191,7 @@ func (w *WebhookCertPatcher) startCaBundleWatcher(stop <-chan struct{}) {
 	for {
 		select {
 		case <-watchCh:
-			lists, err := whc.List(options)
+			lists, err := w.whcLw.List(options)
 			if err != nil {
 				log.Errorf("failed to get mutatingWebhookConfigurations %s", err)
 				break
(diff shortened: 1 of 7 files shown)


pkg/webhooks/webhookpatch.go
```diff
@@ -167,3 +178,27 @@ func (w *WebhookCertPatcher) patchMutatingWebhookConfig(
 	_, err = client.Update(context.TODO(), config, metav1.UpdateOptions{})
 	return err
 }
+
+// startCaBundleWatcher listens for updates to the CA bundle and patches the webhooks.
+func (w *WebhookCertPatcher) startCaBundleWatcher(stop <-chan struct{}) {
+	watchCh := w.CABundleWatcher.AddWatcher()
+	options := metav1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", label.IoIstioRev.Name, w.revision)}
+	whcList, _ := w.client.AdmissionregistrationV1().MutatingWebhookConfigurations().List(context.TODO(), options)
```
hzxuzhonghu, line 186: maybe this should be moved to the loop

irisdingbj: Fixed, PTAL, Thanks! @hzxuzhonghu

hzxuzhonghu (approved):

pkg/webhooks/webhookpatch.go
```diff
@@ -167,3 +177,31 @@ func (w *WebhookCertPatcher) patchMutatingWebhookConfig(
...
+// startCaBundleWatcher listens for updates to the CA bundle and patches the webhooks.
+func (w *WebhookCertPatcher) startCaBundleWatcher(stop <-chan struct{}) {
+	watchCh := w.CABundleWatcher.AddWatcher()
+	options := metav1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", label.IoIstioRev.Name, w.revision)}
+	for {
+		select {
+		case <-watchCh:
+			whcList, err := w.client.AdmissionregistrationV1().MutatingWebhookConfigurations().List(context.TODO(), options)
+			if err != nil {
+				log.Debugf("failed to get mutatingWebhookConfigurations %s", err)
```
hzxuzhonghu, line 190: Errorf

irisdingbj: fixed

pkg/webhooks/webhookpatch.go
```diff
@@ -101,8 +106,8 @@ func (w *WebhookCertPatcher) runWebhookController(stopChan <-chan struct{}) {
 
 func (w *WebhookCertPatcher) updateWebhookHandler(oldConfig, newConfig *v1.MutatingWebhookConfiguration) {
 	if oldConfig.ResourceVersion != newConfig.ResourceVersion {
-		for i, wh := range newConfig.Webhooks {
-			if strings.HasSuffix(wh.Name, w.webhookName) && !bytes.Equal(newConfig.Webhooks[i].ClientConfig.CABundle, w.caCertPem) {
+		for _, wh := range newConfig.Webhooks {
+			if strings.HasSuffix(wh.Name, w.webhookName) {
```
howardjohn, line 110: This change seems suspicious, now we patch even when no changes needed?

irisdingbj: This get triggered  when mutatingwebhookconfiguration added or updated.  For mutatingwebhookconfiguration,  we do not update it after creating.   Since we handle the caBundle update seperately we can remove this useless check.

pkg/webhooks/webhookpatch.go
```diff
@@ -167,3 +177,31 @@ func (w *WebhookCertPatcher) patchMutatingWebhookConfig(
...
 }
+
+// startCaBundleWatcher listens for updates to the CA bundle and patches the webhooks.
+func (w *WebhookCertPatcher) startCaBundleWatcher(stop <-chan struct{}) {
+	watchCh := w.CABundleWatcher.AddWatcher()
+	options := metav1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", label.IoIstioRev.Name, w.revision)}
+	for {
+		select {
+		case <-watchCh:
+			whcList, err := w.client.AdmissionregistrationV1().MutatingWebhookConfigurations().List(context.TODO(), options)
```
howardjohn, line 188: runWebhookController is already creating a watch for MWHs matching this query. We should use that to avoid extra calls to API server?

pkg/webhooks/webhookpatch.go
howardjohn: I still don't think its useless.. its good to skip updates when they are not needed, updates are fairly expensive.

Even if we are not triggering updates, they can happen for all sorts of reasons. Why not add a simple 1 line check to avoid a full api server call?

howardjohn (commented):
Minor comment but looks good otherwise!

pkg/webhooks/webhookpatch.go
irisdingbj: add the check back for update case

pkg/webhooks/webhookpatch.go
irisdingbj: Thanks, reuse now.

pkg/webhooks/webhookpatch.go
```diff
@@ -30,13 +30,16 @@ import (
...
+	"istio.io/istio/pilot/pkg/keycertbundle"
 	"istio.io/istio/pkg/queue"
+	"istio.io/istio/pkg/webhooks/util"
 	"istio.io/pkg/log"
 )
 
 var (
 	errWrongRevision     = errors.New("webhook does not belong to target revision")
 	errNoWebhookWithName = errors.New("webhook configuration did not contain webhook with target name")
+	whc                  *cache.ListWatch
```
howardjohn, line 42: We cannot use global variables, can you keep it scoped in functions/WebhookCertPatcher?

pkg/webhooks/webhookpatch.go
```diff
@@ -113,8 +123,8 @@ func (w *WebhookCertPatcher) updateWebhookHandler(oldConfig, newConfig *v1.Mutat
 }
 
 func (w *WebhookCertPatcher) addWebhookHandler(config *v1.MutatingWebhookConfiguration) {
-	for i, wh := range config.Webhooks {
-		if strings.HasSuffix(wh.Name, w.webhookName) && !bytes.Equal(config.Webhooks[i].ClientConfig.CABundle, w.caCertPem) {
+	for _, wh := range config.Webhooks {
+		if strings.HasSuffix(wh.Name, w.webhookName) {
```
howardjohn, line 127: Copying my comment which somehow got hidden:

I still don't think its useless.. its good to skip updates when they are not needed, updates are fairly expensive.

Even if we are not triggering updates, they can happen for all sorts of reasons. Why not add a simple 1 line check to avoid a full api server call?

pkg/webhooks/webhookpatch.go
irisdingbj: sure, fixed

howardjohn (approved):

PR #33855, opened 2021-07-06
BootstrapNodeMetadata unmarshalJSON fails

piotrkpc:
when PROXY_VIA_AGENT is set due to missing
`,string` tag in struct

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ x ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -477,7 +477,7 @@ type BootstrapNodeMetadata struct {
 	Owner string `json:"OWNER,omitempty"`
 
 	// ProxyViaAgent specifies whether xDS streams are proxied through the agent.
-	ProxyViaAgent bool `json:"PROXY_VIA_AGENT,omitempty"`
+	ProxyViaAgent bool `json:"PROXY_VIA_AGENT,string,omitempty"`
 
 	// PilotSAN is the list of subject alternate names for the xDS server.
 	PilotSubjectAltName []string `json:"PILOT_SAN,omitempty"`
--- a/pilot/pkg/model/context_test.go
+++ b/pilot/pkg/model/context_test.go
@@ -622,3 +622,16 @@ func TestGlobalUnicastIP(t *testing.T) {
 		})
 	}
 }
+func TestBootstrapNodeMetadata(t *testing.T) {
+	t.Run("UnmarshalJSON", func(t *testing.T) {
+		var bnm model.BootstrapNodeMetadata
+		sampleJSON := `{"PROXY_VIA_AGENT":"true"}`
+		err := bnm.UnmarshalJSON([]byte(sampleJSON))
+		if err != nil {
+			t.Errorf("%v", err)
+		}
+		if bnm.ProxyViaAgent != true {
+			t.Errorf("got %v, want %v", bnm.ProxyViaAgent, true)
+		}
+	})
+}


bianpengyuan:
When is PROXY_VIA_AGENT set to string and causing the unmarshal error?

piotrkpc:
hi @bianpengyuan we are getting this error on `istio-ingressgateway` startup
```
2021-07-07T11:11:55.657154Z	warning	envoy config	StreamAggregatedResources gRPC config stream closed: 2, failed to unmarshal node metadata ({"CLUSTER_ID":"Kubernetes","INSTANCE_IPS":"100.96.6.13","ISTIO_PROXY_SHA":"istio-proxy:5952f5fb5f85f495ac6d0925539dcd445cd618a0","ISTIO_VERSION":"1.10.2","LABELS":{"app":"istio-ingressgateway","chart":"gateways","heritage":"Tiller","install.operator.istio.io/owning-resource":"unknown","istio":"ingressgateway","istio.io/rev":"default","operator.istio.io/component":"IngressGateways","pod-template-hash":"59798cf48","release":"istio","service.istio.io/canonical-name":"istio-ingressgateway","service.istio.io/canonical-revision":"latest","sidecar.istio.io/inject":"false"},"MESH_ID":"cluster.local","NAME":"istio-ingressgateway-59798cf48-h58cj","NAMESPACE":"istio-system","OWNER":"kubernetes://apis/apps/v1/namespaces/istio-system/deployments/istio-ingressgateway","PILOT_CERT_PROVIDER":"istiod","PILOT_SAN":["istiod.istio-system.svc"],"PLATFORM_METADATA":{"azure_Name":"shoot--kyma-stage--c-67001a4","azure_kubernetes.io-cluster-shoot--kyma-stage--c-67001a4":"1","azure_kubernetes.io-role-node":"1","azure_location":"westeurope","azure_name":"shoot--kyma-stage--c-67001a4-cpu-worker-0-z1-775bd-vrq7s","azure_node.kubernetes.io_role":"node","azure_vmId":"caffae5b-845c-4e97-b442-8762eff5c25a","azure_worker.garden.sapcloud.io_group":"cpu-worker-0","azure_worker.gardener.cloud_pool":"cpu-worker-0","azure_worker.gardener.cloud_system-components":"true"},"PROV_CERT":"var/run/secrets/istio/root-cert.pem","PROXY_CONFIG":{"binaryPath":"/usr/local/bin/envoy","concurrency":"0","configPath":"./etc/istio/proxy","controlPlaneAuthPolicy":"MUTUAL_TLS","discoveryAddress":"istiod.istio-system.svc:15012","drainDuration":"45s","parentShutdownDuration":"60s","proxyAdminPort":"15000","proxyMetadata":{},"serviceCluster":"istio-ingressgateway","statNameLength":"189","statusPort":"15020","terminationDrainDuration":"5s","tracing":{"zipkin":{"address":"zipkin.kyma-system:9411"}}},"PROXY_VIA_AGENT":"true","ROUTER_MODE":"sni-dnat","SERVICE_ACCOUNT":"istio-ingressgateway-service-account","UNPRIVILEGED_POD":"true","WORKLOAD_NAME":"istio-ingressgateway"}): json: cannot unmarshal string into Go struct field BootstrapNodeMetadata2.PROXY_VIA_AGENT of type bool
```
we are not running pilot-agent and the ingressgateway stuck in this state:
kopec@piorts-nb ~ » k get pods 
NAME                                                READY   STATUS    RESTARTS   AGE
application-connector-certs-sync-76965d9c75-wqkq4   1/1     Running   0          45h
istio-ingressgateway-59798cf48-2cgf9                1/1     Running   0          46h
istio-ingressgateway-59798cf48-h58cj                0/1     Running   0          49m
istio-ingressgateway-59798cf48-rwftw                0/1     Running   0          49m
istio-ingressgateway-59798cf48-zfp2x                0/1     Running   0          49m
istiod-5b8f656446-b5wv4                             1/1     Running   0          46h
istiod-5b8f656446-pq68c                             1/1     Running   0          49m

bianpengyuan:
Does not happen to me... Do you happen to set `PROXY_VIA_AGENT` in env variable or customized bootstrap override?

pilot/pkg/model/context.go
```diff
@@ -477,7 +477,7 @@ type BootstrapNodeMetadata struct {
 	Owner string `json:"OWNER,omitempty"`
 
 	// ProxyViaAgent specifies whether xDS streams are proxied through the agent.
-	ProxyViaAgent bool `json:"PROXY_VIA_AGENT,omitempty"`
+	ProxyViaAgent bool `json:"PROXY_VIA_AGENT,string,omitempty"`
```
howardjohn, line 480: TIL about `string` option...

In other places we use StringBool. We should probably be consistent across the two. Likely this method is better

Tomasz-Smelcerz-SAP:
@bianpengyuan I am assisting @piotrkpc on this issue. I have found what follows:
- Both istio-system/ingressgateways and Pod sidecars in user namespaces are affected by the problem: unable to parse string "true" into Boolean
- For ingressgateways I was able to provide `PROXY_VIA_AGENT` env var, because there's a Deployment. And while setting it to `"true"` yields the same error, setting it to '"false"` surprisingly works, the Pods start normally and seem to work. I don't know all the consequences of switching this value though.
- For injected sidecars I was able to "fix" the problem by providing custom bootstrap override. Sample below

Custom bootstrap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: istio-custom-bootstrap-config
  namespace: tomek
data:
  custom_bootstrap.json: |
    {
      "node": {
        "metadata": {
           "PROXY_VIA_AGENT": true
        }
      }
    }
```

The custom bootstrap fix would require to patch every single Pod in the cluster though. Not really acceptable long-term solution. Is there a way to inject it globally?
Where should we look to find the root cause of this strange behaviour (which configuration sources)? 
We provisioned hundreds of clusters and this is the only one with this problem (so far), so we think there must be something in the configuration, or cluster environment, that is causing this errors.

Tomasz-Smelcerz-SAP:
The same happens when I disable automatic sidecar injection and inject manually with `istioctl inject`. I did that to have full control over sidecar envs. But I don't see anything suspicious there, actually the same set of envs work just fine on another cluster. So this has to be a config external to the sidecar, perhaps something in istiod.

bianpengyuan:
`PROXY_VIA_AGENT` should already be default and I don't think you need to configure it at all, no?

piotrkpc:
@bianpengyuan yes, the thing is more complicated - please see this issue for details: https://github.com/istio/istio/issues/33989

Issue #33857, opened 2021-07-06
istioctl not showing default profile name during install


murphye:
```
istioctl install
This will install the Istio 1.10.1  profile with ["Istio core" "Istiod" "Ingress gateways"] components into the cluster. Proceed? (y/N)
```

Between `1.10.1  profile` there are 2 spaces. Here it should say `1.10.1 default profile`.

ericvn:
Noted this also installing the 1.11.0-beta.0. Labeled the PR to be cherry-picked.

PR #33864, opened 2021-07-07
fix istioctl install display error

hanxiaop:
This PR is to fix issue #33857. It will show `This will install the Istio 1.10.1 default profile with` instead of `This will install the Istio 1.10.1  profile with` if no arg is provided.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/operator/cmd/mesh/install.go
+++ b/operator/cmd/mesh/install.go
@@ -347,7 +347,7 @@ func getProfileNSAndEnabledComponents(iop *v1alpha12.IstioOperator) (string, str
 	}
 
 	if configuredNamespace := v1alpha12.Namespace(iop.Spec); configuredNamespace != "" {
-		return iop.Spec.Profile, configuredNamespace, enabledComponents, nil
+		return renderWithDefault(iop.Spec.Profile, "default"), configuredNamespace, enabledComponents, nil
 	}
 	return renderWithDefault(iop.Spec.Profile, "default"), name.IstioDefaultNamespace, enabledComponents, nil
 }
--- a/operator/cmd/mesh/install.go
+++ b/operator/cmd/mesh/install.go
@@ -347,14 +347,7 @@ func getProfileNSAndEnabledComponents(iop *v1alpha12.IstioOperator) (string, str
 	}
 
 	if configuredNamespace := v1alpha12.Namespace(iop.Spec); configuredNamespace != "" {
-		return renderWithDefault(iop.Spec.Profile, "default"), configuredNamespace, enabledComponents, nil
+		return iop.Spec.Profile, configuredNamespace, enabledComponents, nil
 	}
-	return renderWithDefault(iop.Spec.Profile, "default"), name.IstioDefaultNamespace, enabledComponents, nil
-}
-
-func renderWithDefault(s, def string) string {
-	if s != "" {
-		return s
-	}
-	return def
+	return iop.Spec.Profile, name.IstioDefaultNamespace, enabledComponents, nil
 }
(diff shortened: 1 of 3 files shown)


releasenotes/notes/33864.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: istioctl
+issue:
+  - 33857
+releaseNotes:
+  - |
+    **Fixed** a message format bug for `istioctl install` command with the default file setting.
```
shamsher31, line 8:
```suggestion
    **Fixed** an issue where the `default` profile name was missing in the `istioctl install` confirmation prompt message.
```

shamsher31 (commented):
LGTM 
```
$ istioctl install
This will install the Istio 1.11.0 default profile with ["Istio core" "Istiod" "Ingress gateways"] components into the cluster. Proceed? (y/N)
```

```
$ istioctl install --set profile=demo
This will install the Istio 1.11.0 demo profile with ["Istio core" "Istiod" "Ingress gateways" "Egress gateways"] components into the cluster. Proceed? (y/N)
```

```
$ istioctl install --set profile=minimal
This will install the Istio 1.11.0 minimal profile with ["Istio core" "Istiod"] components into the cluster. Proceed? (y/N)
```

shamsher31 (approved):

releasenotes/notes/33864.yaml
monsteredc: @shamsher31 Thanks for the suggestion!

operator/cmd/mesh/install.go
```diff
@@ -347,7 +347,14 @@ func getProfileNSAndEnabledComponents(iop *v1alpha12.IstioOperator) (string, str
 	}
 
 	if configuredNamespace := v1alpha12.Namespace(iop.Spec); configuredNamespace != "" {
-		return iop.Spec.Profile, configuredNamespace, enabledComponents, nil
+		return renderWithDefault(iop.Spec.Profile, "default"), configuredNamespace, enabledComponents, nil
```
howardjohn, line 350: Should we just be setting the iop.Spec.Profile value somewhere? Not sure if we are already providing defaults directly to the spec in other parts of the code or not

monsteredc: @howardjohn It seems like when flags are setting like "profile=demo", `overlaySetFlagValues` will set the path like {spec.profile: "demo"}. However, if none profile flag is provided, the profile field will be empty. I've revised and PTAL.

operator/pkg/manifest/shared.go
```diff
@@ -537,6 +539,20 @@ func overlaySetFlagValues(iopYAML string, setFlags []string) (string, error) {
 	return string(out), nil
 }
 
+// addDefaultProfileFlagIfNotSet adds the `profile=default` value to set flags if no profile value is set.
+func addDefaultProfileFlagIfNotSet(setFlags []string) []string {
```
howardjohn, line 543: I don't think this is the right layer to apply this fix. What about something more like
```diff
--- a/operator/pkg/manifest/shared.go
+++ b/operator/pkg/manifest/shared.go
@@ -212,6 +212,9 @@ func GenIOPFromProfile(profileOrPath, fileOverlayYAML string, setFlags []string,
        if ns := GetValueForSetFlag(setFlags, "values.global.istioNamespace"); ns != "" {
                finalIOP.Namespace = ns
        }
+       if finalIOP.Spec.Profile == "" {
+               finalIOP.Spec.Profile = name.DefaultProfileName
+       }
        return util.ToYAMLWithJSONPB(finalIOP), finalIOP, nil
 }
```

howardjohn (changes requested):

operator/pkg/manifest/shared.go
monsteredc: @howardjohn Thanks! Looks much better.

howardjohn (approved):
Thanks!

PR #33869, opened 2021-07-07
pkg/kube: fix client.PodExecCommands repeated error message

wzshiming:
fix `client.PodExecCommands` repeated error message if has stderr

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -573,9 +573,10 @@ func (c *client) PodExecCommands(podName, podNamespace, container string, comman
 			if len(stderr) > 0 {
 				err = fmt.Errorf("error exec'ing into %s/%s %s container: %v\n%s",
 					podName, podNamespace, container, err, stderr)
+			} else {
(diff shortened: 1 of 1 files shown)


pkg/kube/client.go
```diff
@@ -573,9 +573,10 @@ func (c *client) PodExecCommands(podName, podNamespace, container string, comman
 			if len(stderr) > 0 {
 				err = fmt.Errorf("error exec'ing into %s/%s %s container: %v\n%s",
 					podName, podNamespace, container, err, stderr)
+			} else {
+				err = fmt.Errorf("error exec'ing into %s/%s %s container: %v",
+					podName, podNamespace, container, err)
```
bianpengyuan, line 578: Not because this PR, but it feels wrong to me to log as `podname/podNamespace`. Since you are here, could you swap these two log args?

howardjohn (approved):

PR #33874, opened 2021-07-07
Does renaming support ipv4/6 field names

ChenglongGao327:
rename field name ipv4Support to ipv4  and ipv6Support to ipv6

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ramaraochavali:
Any specific reason for this change? In dual IP proxies it indictates whether the proxy supports ipv4 or ipv6 - In that sense `ipv4Support` may be more meaningful

hzxuzhonghu:
yes, the rename looks meaningless. I am thinking we should leave one field, currently, they are opposite, looks redundant

howardjohn:
They are not opposites, you can support one or both (although neither is
probably not possible).For dual stack

hzxuzhonghu:
@howardjohn At least for now, the values are opposite.  For dual stack, i am not sure how it is supported.

If we make it one day in future.
`IPMode`  [ `IPV4`, `IPV6`, `DUAL_STACK`]looks more reasonable.

PR #33878, opened 2021-07-07
Add auto detect autoInject policy in `istioctl x injector list`

kebe7jun:
If disable sidecar autoinject with `--set values.global.proxy.autoInject=disabled`, the `istioctl x injector list` shows incorrect disabled pods count.

### DIFF
--- a/istioctl/cmd/injector-list.go
+++ b/istioctl/cmd/injector-list.go
@@ -28,6 +28,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	api_pkg_labels "k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/yaml"
 
 	"istio.io/api/annotation"
 	"istio.io/api/label"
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
This seems fine but we should deprecate the autoInject option. I don't think there is any good reason to use it anymore.

kebe7jun:
> This seems fine but we should deprecate the autoInject option. I don't think there is any good reason to use it anymore.

In some cases, there is a relatively large namespace, but because of additional things to consider, such as the impact of Sidecar on application performance or resources, we have to do automatic injection for only some of these Pods. In fact, many of our customers use it this way.
Even if we were to deprecate this feature, which we don't seem to have planned, the injector command is really flawed.

howardjohn:
The setting was made long ago, before Kubernetes supported objectSelector.
You should be able to do everything you want in the webhook directly, which
is strictly better than this setting (injector is never even called). See
https://preliminary.istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy
.

I am fine with this PR, since the setting does exist - but nobody should be
using this I think.

PR #32649, opened 2021-05-05
Istiod CA: Refactor CA interface and use correct API of said interface when signing certificates

shankgan:
Part of fix for https://github.com/istio/istio/issues/32403

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #33898, opened 2021-07-08
Google Certificate Authority Service plugin

shankgan:
Add to the Istio Agent a plugin client to get workload certificates signed by [Google Certificate Authority Service](https://cloud.google.com/certificate-authority-service)

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[X] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/security/pkg/nodeagent/caclient/providers/google-cas/client.go
+++ b/security/pkg/nodeagent/caclient/providers/google-cas/client.go
@@ -22,47 +22,34 @@ import (
 	privateca "cloud.google.com/go/security/privateca/apiv1"
 	"google.golang.org/api/option"
 	privatecapb "google.golang.org/genproto/googleapis/cloud/security/privateca/v1"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/test/bufconn"
 	"google.golang.org/protobuf/types/known/durationpb"
 	"k8s.io/apimachinery/pkg/util/rand"
 
 	"istio.io/istio/pkg/security"
-	"istio.io/istio/security/pkg/nodeagent/caclient"
-	"istio.io/istio/security/pkg/nodeagent/caclient/providers/google-cas/mock"
 	"istio.io/pkg/log"
 )
 
-var googleCAClientLog = log.RegisterScope("googlecas", "Google CAS client debugging", 0)
+var googleCASClientLog = log.RegisterScope("googlecas", "Google CAS client debugging", 0)
 
 // GoogleCASClient: Agent side plugin for Google CAS
 type GoogleCASClient struct {
 	caSigner string
 	caClient *privateca.CertificateAuthorityClient
-	provider *caclient.TokenProvider
 }
 
 // NewGoogleCASClient create a CA client for Google CAS.
-func NewGoogleCASClient(capool string, provider *caclient.TokenProvider, lis *bufconn.Listener) (security.Client, error) {
-	var err error
-	caClient := &GoogleCASClient{caSigner: capool, provider: provider}
+func NewGoogleCASClient(capool string, options ...option.ClientOption) (security.Client, error) {
+	caClient := &GoogleCASClient{caSigner: capool}
 	ctx := context.Background()
+	var err error
 
-	if lis == nil {
-		caClient.caClient, err = privateca.NewCertificateAuthorityClient(ctx,
-			option.WithGRPCDialOption(grpc.WithPerRPCCredentials(provider)))
-	} else {
-		// For Unit Testing
-		caClient.caClient, err = privateca.NewCertificateAuthorityClient(ctx,
-			option.WithoutAuthentication(),
-			option.WithGRPCDialOption(grpc.WithContextDialer(mock.ContextDialerCreate(lis))),
-			option.WithGRPCDialOption(grpc.WithInsecure()))
-	}
+	caClient.caClient, err = privateca.NewCertificateAuthorityClient(ctx, options...)
+
 	if err != nil {
-		googleCAClientLog.Errorf("unable to initialize google cas caclient: %v", err)
+		googleCASClientLog.Errorf("unable to initialize google cas caclient: %v", err)
 		return nil, err
 	}
-	googleCAClientLog.Debugf("Intitialized Google CAS plugin with endpoint: %v", capool)
+	googleCASClientLog.Debugf("Intitialized Google CAS plugin with endpoint: %v", capool)
 	return caClient, nil
 }
 
@@ -108,6 +95,8 @@ func (r *GoogleCASClient) createCertReq(name string, csrPEM []byte, lifetime tim
 // CSR Sign calls Google CAS to sign a CSR.
 func (r *GoogleCASClient) CSRSign(csrPEM []byte, certValidTTLInSec int64) ([]string, error) {
 	certChain := []string{}
+
+	rand.Seed(time.Now().UnixNano())
 	name := fmt.Sprintf("csr-workload-%s", rand.String(8))
 	creq := r.createCertReq(name, csrPEM, time.Duration(certValidTTLInSec)*time.Second)
 
@@ -115,7 +104,7 @@ func (r *GoogleCASClient) CSRSign(csrPEM []byte, certValidTTLInSec int64) ([]str
 
 	cresp, err := r.caClient.CreateCertificate(ctx, creq)
 	if err != nil {
-		googleCAClientLog.Errorf("unable to create certificate: %v", err)
+		googleCASClientLog.Errorf("unable to create certificate: %v", err)
 		return []string{}, err
 	}
 	certChain = append(certChain, cresp.GetPemCertificate())
@@ -136,7 +125,7 @@ func (r *GoogleCASClient) GetRootCertBundle() ([]string, error) {
 	}
 	resp, err := r.caClient.FetchCaCerts(ctx, req)
 	if err != nil {
-		googleCAClientLog.Errorf("error when getting root-certs from CAS pool: %v", err)
+		googleCASClientLog.Errorf("error when getting root-certs from CAS pool: %v", err)
 		return trustbundle, err
 	}
 	for _, certChain := range resp.CaCerts {
--- a/security/pkg/nodeagent/caclient/providers/google-cas/client.go
+++ b/security/pkg/nodeagent/caclient/providers/google-cas/client.go
@@ -1,0 +1,150 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package caclient
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	privateca "cloud.google.com/go/security/privateca/apiv1"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+	privatecapb "google.golang.org/genproto/googleapis/cloud/security/privateca/v1"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/durationpb"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	"istio.io/istio/pkg/security"
+	"istio.io/istio/security/pkg/nodeagent/caclient"
+	"istio.io/pkg/log"
+)
+
+var googleCAClientLog = log.RegisterScope("googlecas", "Google CAS client debugging", 0)
+
+type GoogleCASClient struct {
+	caSigner string
+	caClient *privateca.CertificateAuthorityClient
+	provider *caclient.TokenProvider
+}
+
+// NewGoogleCASClient create a CA client for Google CA.
+func NewGoogleCASClient(endpoint string, provider *caclient.TokenProvider) (security.Client, error) {
+	var err error
+	caClient := &GoogleCASClient{caSigner: endpoint, provider: provider}
+	ctx := context.Background()
+
+	// grpc.WithPerRPCCredentials(provider),
+	// security.CARetryInterceptor()
+	caClient.caClient, err = privateca.NewCertificateAuthorityClient(ctx,
+		option.WithGRPCDialOption(grpc.WithPerRPCCredentials(provider)))
+	if err != nil {
+		googleCAClientLog.Errorf("unable to initialize google cas caclient: %v", err)
+		return nil, err
+	}
+	// TODO: Initialize the certChain from the CA
+	return caClient, nil
+}
+
+func (r *GoogleCASClient) createCertReq(name string, csrPEM []byte, lifetime time.Duration) *privatecapb.CreateCertificateRequest {
+	subjectConfig := &privatecapb.CertificateConfig_SubjectConfig{
+		// Empty Subject
+		Subject: &privatecapb.Subject{},
+	}
+	keyUsage := &privatecapb.KeyUsage{
+		BaseKeyUsage: &privatecapb.KeyUsage_KeyUsageOptions{
+			DigitalSignature: true,
+			KeyEncipherment:  true,
+		},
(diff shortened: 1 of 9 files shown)


JimmyCYJ:
cc @JimmyCYJ

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,157 @@
...
+		Format: privatecapb.PublicKey_PEM,
+		Key:    csrPEM,
+	}
+
+	creq := &privatecapb.CreateCertificateRequest{
+		Parent:        r.caSigner,
+		CertificateId: name,
+		Certificate: &privatecapb.Certificate{
+			Lifetime: durationpb.New(lifetime),
+			CertificateConfig: &privatecapb.Certificate_Config{
```
JimmyCYJ, line 94: According to https://pkg.go.dev/google.golang.org/genproto/googleapis/cloud/security/privateca/v1#Certificate, this is a oneoff (*Certificate_PemCsr or *Certificate_PemCsr). Could you add a comment explaining the reason of choosing Certificate_Config?

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,157 @@
...
+		googleCAClientLog.Errorf("unable to initialize google cas caclient: %v", err)
+		return nil, err
+	}
+	return caClient, nil
+}
+
+func (r *GoogleCASClient) createCertReq(name string, csrPEM []byte, lifetime time.Duration) *privatecapb.CreateCertificateRequest {
+	subjectConfig := &privatecapb.CertificateConfig_SubjectConfig{
+		// Empty Subject
+		Subject: &privatecapb.Subject{},
```
JimmyCYJ, line 69: Why are we using an empty struct. 
By looking at https://pkg.go.dev/google.golang.org/genproto/googleapis/cloud/security/privateca/v1#Subject, shall we pass `CommonName`, `Organization` and others into the struct from config/environment variables?

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,157 @@
...
+func (r *GoogleCASClient) CSRSign(csrPEM []byte, certValidTTLInSec int64) ([]string, error) {
+	certChain := []string{}
+	name := fmt.Sprintf("csr-workload-%s", rand.String(8))
+	creq := r.createCertReq(name, csrPEM, time.Duration(certValidTTLInSec)*time.Second)
+
+	ctx := context.Background()
+
+	cresp, err := r.caClient.CreateCertificate(ctx, creq)
+	if err != nil {
+		googleCAClientLog.Errorf("unable to create certificate: %v", err)
```
JimmyCYJ, line 119: Sometimes a CSR sign failure is caused by client side configuration. Can we add more information for debugging purposes, e.g. private CA endpoint information?

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,157 @@
...
+		return []string{}, err
+	}
+	certChain = append(certChain, cresp.GetPemCertificate())
+	certChain = append(certChain, cresp.GetPemCertificateChain()...)
+	return certChain, nil
+}
+
+func (r *GoogleCASClient) GetRootCertBundle() ([]string, error) {
```
JimmyCYJ, line 127: Add comment?

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+
+	for rootCert := range rootCertMap {
+		trustbundle = append(trustbundle, rootCert)
+	}
+	return trustbundle, nil
+}
+
+func (r *GoogleCASClient) Close() {
```
JimmyCYJ, line 156: Is this intended to be a no-op?

security/pkg/nodeagent/caclient/providers/google/client.go
```diff
@@ -135,6 +135,10 @@ func (cl *googleCAClient) getTLSDialOption() (grpc.DialOption, error) {
 	return grpc.WithTransportCredentials(creds), nil
 }
 
+func (cl *googleCAClient) GetRootCertBundle() ([]string, error) {
```
JimmyCYJ, line 138: If we don't expect this method to be called on googleCAClient/CitadelClient, maybe we can add a log before return. That would make it easier for troubleshooting.

JimmyCYJ (commented):
Thanks for implementing this. Left some comments to help me better understand this flow.

howardjohn (commented):
Looks pretty good!

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,157 @@
...
+func NewGoogleCASClient(capool string, provider *caclient.TokenProvider, lis *bufconn.Listener) (security.Client, error) {
+	var err error
+	caClient := &GoogleCASClient{caSigner: capool, provider: provider}
+	ctx := context.Background()
+
+	if lis == nil {
+		caClient.caClient, err = privateca.NewCertificateAuthorityClient(ctx,
+			option.WithGRPCDialOption(grpc.WithPerRPCCredentials(provider)))
+	} else {
+		caClient.caClient, err = privateca.NewCertificateAuthorityClient(ctx,
```
pmansour, line 54: is this expected to ever happen in production? if not, is it possible to avoid adding this code path to the production code and instead use dependency injection to do this in tests?

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,157 @@
...
+		},
+		ExtendedKeyUsage: &privatecapb.KeyUsage_ExtendedKeyUsageOptions{
+			ServerAuth: true,
+			ClientAuth: true,
+		},
+	}
+
+	// Derive Public Key from the csrPEM
+	publicKey := &privatecapb.PublicKey{
+		// TODO: Need to get key-type from security options
```
pmansour, line 84: don't need to specify key type anymore, private CA automatically determines that from the csrPEM.

security/pkg/nodeagent/caclient/providers/google-cas/client.go
pmansour: this code uses Certificate_Config, and sends the CSR as a purely key-delivery mechanism. Comment explaining this makes sense though.

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	return caClient, nil
+}
+
+func (r *GoogleCASClient) createCertReq(name string, csrPEM []byte, lifetime time.Duration) *privatecapb.CreateCertificateRequest {
+	subjectConfig := &privatecapb.CertificateConfig_SubjectConfig{
+		// Empty Subject
+		Subject: &privatecapb.Subject{},
+	}
+	keyUsage := &privatecapb.KeyUsage{
```
pmansour, line 71: nit: can this be inlined below for clarity?

(and same for all other sections of the cert request -- having them inline would make it easier to figure out what gets included, since each variable is only used once)

security/pkg/nodeagent/caclient/providers/google-cas/client.go
pmansour: you don't need to specify any part of the subject config here, since this code uses [identity reflection](https://cloud.google.com/certificate-authority-service/docs/using-identity-reflection), which means the CA sets the certificate identity based on the caller's credential.

perhaps it would be less confusing to make the entire SubjectConfig empty (rather than explicitly populate an empty Subject{})?

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,157 @@
...
+	creq := &privatecapb.CreateCertificateRequest{
+		Parent:        r.caSigner,
+		CertificateId: name,
+		Certificate: &privatecapb.Certificate{
+			Lifetime: durationpb.New(lifetime),
+			CertificateConfig: &privatecapb.Certificate_Config{
+				Config: &privatecapb.CertificateConfig{
+					SubjectConfig: subjectConfig,
+					X509Config: &privatecapb.X509Parameters{
+						KeyUsage: keyUsage,
```
pmansour, line 98: can you also explicitly add CaOptions with `isCa=false`?

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,157 @@
...
+	var rootCertMap map[string]struct{} = make(map[string]struct{})
+	var trustbundle []string = []string{}
+	var err error
+
+	ctx := context.Background()
+
+	req := &privatecapb.FetchCaCertsRequest{
+		CaPool: r.caSigner,
+	}
+	resp, err := r.caClient.FetchCaCerts(ctx, req)
```
pmansour, line 137: will this be called from individual workload sidecars? IIRC, individual workloads won't have permission to call this API, but only the control-plane identity (e.g. istiod)

pmansour (commented):
thanks!

security/pkg/nodeagent/caclient/providers/google-cas/client.go
shankgan: Yes, the intent is for this code to be directly invoked from all the individual workload sidecars. This would require some sort of Google workload identity to be enabled on the cluster + all workloads in the pool would need to have the IAM permissions to retrieve certificates/ca-certificates from a particular pool

security/pkg/nodeagent/caclient/providers/google-cas/client.go
shankgan: Will need to research this. This code is only intended for testing the plugin.

security/pkg/nodeagent/caclient/providers/google/client.go
shankgan: We do expect it to be invoked. This is in a different PR https://github.com/istio/istio/pull/32649

security/pkg/nodeagent/caclient/providers/google-cas/client.go
shankgan: I have added some additional debugs when initializing the plugin that prints the ca-pool information. IMO, all info needed for debugging will be present in "err"

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,157 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	privateca "cloud.google.com/go/security/privateca/apiv1"
+	"google.golang.org/api/option"
+	privatecapb "google.golang.org/genproto/googleapis/cloud/security/privateca/v1"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/test/bufconn"
```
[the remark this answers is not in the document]
shankgan: This code path is only ever used to test the plugin library as part of unit-tests. Do we still need another library?

security/pkg/nodeagent/caclient/providers/google-cas/client.go
shankgan: Nope. Fixed this.

security/pkg/nodeagent/caclient/providers/google-cas/client.go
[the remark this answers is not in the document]
myidpt: Can we get rid of this dependency? Please look at my following comment.

security/pkg/nodeagent/caclient/providers/google-cas/client.go
myidpt: Let's get rid of this block - you can inject the option into this function, so that the construction of option only exists in the unit test.
NewGoogleCASClient(capool string, provider *caclient.TokenProvider, option *option.ClientOption)...

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		googleCAClientLog.Errorf("unable to initialize google cas caclient: %v", err)
+		return nil, err
+	}
+	googleCAClientLog.Debugf("Intitialized Google CAS plugin with endpoint: %v", capool)
+	return caClient, nil
+}
+
+func (r *GoogleCASClient) createCertReq(name string, csrPEM []byte, lifetime time.Duration) *privatecapb.CreateCertificateRequest {
+	var isCA bool = false
```
myidpt, line 70: Since isCA is always false, why not get rid of this variable?

security/pkg/nodeagent/caclient/providers/google-cas/client.go
myidpt: googleCAClientLog -> googleCASClientLog

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,158 @@
...
+	var err error
+
+	ctx := context.Background()
+
+	req := &privatecapb.FetchCaCertsRequest{
+		CaPool: r.caSigner,
+	}
+	resp, err := r.caClient.FetchCaCerts(ctx, req)
+	if err != nil {
+		googleCAClientLog.Errorf("error when getting root-certs from CAS pool: %v", err)
```
myidpt, line 139: googleCAClientLog -> googleCASClientLog

security/pkg/nodeagent/caclient/providers/google-cas/client.go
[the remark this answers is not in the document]
shankgan: fixed

security/pkg/nodeagent/caclient/providers/google-cas/client.go
shankgan: fixed

security/pkg/nodeagent/caclient/providers/google-cas/client.go
```diff
@@ -0,0 +1,145 @@
...
+			SubjectMode: privatecapb.SubjectRequestMode_REFLECTED_SPIFFE,
+		},
+	}
+	return creq
+}
+
+// CSR Sign calls Google CAS to sign a CSR.
+func (r *GoogleCASClient) CSRSign(csrPEM []byte, certValidTTLInSec int64) ([]string, error) {
+	certChain := []string{}
+	name := fmt.Sprintf("csr-workload-%s", rand.String(8))
```
myidpt, line 98: Do you need to call Seed first?
https://pkg.go.dev/k8s.io/apimachinery/pkg/util/rand#Seed

security/pkg/nodeagent/caclient/providers/google-cas/client.go
shankgan: the golang protobuf generated code accepts a bool pointer. Haven't found a way to pass this value in without defining a var first

security/pkg/nodeagent/caclient/providers/google-cas/client.go
shankgan: done

myidpt (approved):

myidpt (approved):

security/pkg/nodeagent/caclient/providers/google-cas/client_test.go
```diff
@@ -0,0 +1,110 @@
...
+			service:            mock.CASService{CertPEM: fakeCert, CertChainPEM: fakeCertChain, CaCertBundle: fakeCaBundle},
+			expectedCert:       fakeCombinedCert,
+			expectedCertBundle: fakeExpectedRootCaBundle,
+			expectedErr:        status.Error(codes.InvalidArgument, "malformed ca path"),
+		},
+	}
+
+	for id, tc := range testCases {
+		// create a local grpc server
+		s, lis, err := mock.CreateServer(&tc.service)
```
howardjohn, line 70: This test is leaking goroutines. Can you add `leak_test.go` to the package (copy from another package)?

howardjohn (approved):

security/pkg/nodeagent/caclient/providers/google-cas/client_test.go
shankgan: It appears this is common to all the caclient plugins..will fix this in a separate PR

PR #33900, opened 2021-07-08
do some cleanup in kube pkg

gy95:
modify some comment to comply with programming specifications, and simplify some codes

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ X] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -127,7 +127,7 @@ type Client interface {
 	// IstioInformer returns an informer for the istio client
 	IstioInformer() istioinformer.SharedInformerFactory
 
-	// GatewayApiInformer returns an informer for the gateway-api client
+	// GatewayAPIInformer returns an informer for the gateway-api client
 	GatewayAPIInformer() gatewayapiinformer.SharedInformerFactory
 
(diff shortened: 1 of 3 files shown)


pkg/kube/client.go
```diff
@@ -1027,7 +1026,7 @@ func removeEmptyFiles(files []string) []string {
 
 func isEmptyFile(f string) bool {
 	fileInfo, err := os.Stat(f)
-	if err != nil {
+	if err != nil && !os.IsExist(err) {
```
hzxuzhonghu, line 1029: I think for other err, the fileInfo will be nil
And caling fileInfo.Size() will panic

gy95: oh, yes, i thought if we need to judge the file exist, there are two conditons: one is err is nil, the other is err is isExist Error, but    the source code is only when err is not nil will return non-nil fileInfo

hzxuzhonghu (approved):
/ok-to-test

Issue #8658, opened 2018-09-12
Remove wildcard * restriction on the gateways


vadimeisenbergibm:
The following check does not allow wildcard host * 
https://github.com/istio/istio/blob/188baf0c93308b741087ca7556d4a77967777b69/pilot/pkg/model/validation.go#L1316
```
else if appliesToMesh && host == "*" {
			errs = appendErrors(errs, fmt.Errorf("wildcard host * is not allowed for virtual services bound to the mesh gateway"))
			allHostsValid = false
		}
```

Why do we have this restriction? Why `*.com` is OK and `*` is not? The use case is directing all the SNI traffic thru an egress gateway, see https://github.com/istio/istio.github.io/pull/1984/commits/095c4c1436049f7af93552fc86282783c2a4593b

linsun:
@frankbu do you know why we don't support wildcard * on the gateways?  Or is this only for the egress gateway @vadimeisenbergibm ?   host: "*" is listed as an example here https://github.com/istio/api/blob/master/networking/v1alpha3/gateway.proto#L74

vadimeisenbergibm:
@rshriram Do you know why `*` hosts are not allowed?

```
else if appliesToMesh && host == "*" {
   errs = appendErrors(errs, fmt.Errorf("wildcard host * is not allowed for virtual services bound to the mesh gateway"))
```

jdamour98:
Hello? No answer? I have 2 pods running behind my ingress gateway, and I want to route to a pod based purely on the uri prefix. Any reason this isn't allowed? @rshriram @linsun

anyyyy:
Hello.What is the reason for restricting mesh gateways from using wildcards "*"? If I apply VirtualService to the created gateway now, then the policy rules in VirtualService will not take effect between services in the grid. . Is my understanding wrong, or if there are other solutions.

hzxuzhonghu:
Wildcard host in vs is very risky. 

https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/route_matching#route-matching

ramaraochavali:
I did not understand why it is only risky for mesh gateway not for others?

hzxuzhonghu:
I did not think about the difference, should take a look at the original context

sreejesh-radhakrishnan-db:
any update please?  will host = "*" be allowed? or atleast host ="me.*"  ?

PR #33904, opened 2021-07-08
allow wildcard hosts for mesh gateway

ramaraochavali:
Fixes https://github.com/istio/istio/issues/8658. I could not find a reason why this should be restricted for mesh gateway.  We need this when the same virtual service needs to be applied for "mesh" and ingress gateway.  If there is a concern for allowing this for "mesh", I can change this to allow if there one gateway other than "mesh" (but I could not think of a case for it). We have not documented this restriction in any of our APIs.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


howardjohn (commented):
What happens with multiple `*` VirtualServices?

ramaraochavali:
> What happens with multiple * VirtualServices?

You mean a single VS having multiple '*' hosts? We will have to stop it from having I think - Is that what you are asking?

howardjohn:
No, multiple with `*`. Won't they conflict?

ramaraochavali:
> No, multiple with `*`. Won't they conflict?

I think they wont because such a config is rejected by Envoy saying "there can not be more than one wildcard domain" - So no one would have tried it.

But if that is concern, we can validate that when we build domains and add only one - WDYT?

ramaraochavali:
But I think that is different but related problem to this though - since we allow that already allow that for non mesh gateways now.

howardjohn:
So right now envoy will NACK? That seems like something we should be
catching..

ramaraochavali:
> So right now envoy will NACK? That seems like something we should be
catching..

We are deduping domains so it must filtered. But I will have to add a test to confirm. If we are good with relaxing this - I can add that test and if it sends duplicate `*` can fix it.

ramaraochavali:
@howardjohn @hzxuzhonghu added test to validate the behaviour when duplicate wild card hosts are specified across virtual services. Found a bug when virtual host specifies a wild card, we are building domains based on matching service instead of host in VS.  PTAL.

pilot/pkg/networking/core/v1alpha3/httproute_test.go
```diff
@@ -142,6 +142,95 @@ func TestGenerateVirtualHostDomains(t *testing.T) {
...
+	if routeCfg == nil {
+		t.Fatalf("got nil route for %s", routeName)
+	}
+	expectedHosts := map[string][]string{
+		"allow_any":  {"*"},
+		"*.com:7442": {"*.com", "*.com:7442"},
+	}
+	expectedDestination := map[string]string{
+		"allow_any":  "PassthroughCluster",
+		"*.com:7442": "outbound|7442||test-service-one.com",
```
howardjohn, line 216: So this is picked by oldest config?

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -138,8 +148,12 @@ func separateVSHostsAndServices(virtualService config.Config,
 		for svcHost, svc := range serviceRegistry {
 			// *.foo.global matches *.global
 			if svcHost.Matches(hostname) {
+				// Change the host name of service to the host name in virtual service as we have to build domains based
+				// on virtual service host names.
+				svc.Hostname = hostname
```
howardjohn, line 153: is mutation ok here? Is this not some struct that is shared beyond this function?

hzxuzhonghu: looks not ok, it is a pointer and shared by all proxies

ramaraochavali: It is safe to mutate. Because we build new service objects here https://github.com/istio/istio/blob/dbf6a9e1c123bf6abbab45683b1c6c1f7660916c/pilot/pkg/networking/core/v1alpha3/httproute.go#L220

pilot/pkg/networking/core/v1alpha3/httproute_test.go
ramaraochavali: Yes. That is correct.

pilot/pkg/networking/core/v1alpha3/route/route.go
hzxuzhonghu: The other case listenerPort=0, it is not reconstructed

ramaraochavali: I have copied there also

howardjohn: What if we have `*.foo.global` and `*.global`? Both will match the `svc`, and then try to mutate the Hostname?

hzxuzhonghu:
I think we need to weigh the benefits and risks.  What is the real use case, for setting `*` in mesh gateway? If `*` is set, all the HTTP traffic will be matched by this policy. There is no once for all thing, different services need different routing policies. Not to mention the conflict possibilities increased by allowing wildcard hosts.

ramaraochavali:
> I think we need to weigh the benefits and risks. What is the real use case, for setting * in mesh gateway? If * is set, all the HTTP traffic will be matched by this policy. There is no once for all thing, different services need different routing policies. Not to mention the conflict possibilities increased by allowing wildcard hosts.

But that is not new with `*` though. The same can happen with `*.com`, `*.net`   and more over this is a valid config from gateways. This makes route reuse in virtual services difficult.  I understand your point but exposing `*` in mesh but was wondering why is it OK in gateway but not on mesh. 

Only case where it makes sense for gateway I can think of is if a gateway exposes only one app (let us call it main app and main app routes to all other services) and it is OK to route every traffic from that gateway to the app - so `*` is a valid option as the route is picked by SNI. Is that what you are thinking as well? If gateway exposes multiple services - would not we hit the same problem?

pilot/pkg/networking/core/v1alpha3/route/route.go
ramaraochavali: This is no different from what we already do today. We pick first matching service. Here we are picking first matching host in VS. So should be fine.

ramaraochavali:
Put a hold so that we can discuss fully

hzxuzhonghu:
I agree with you gateway may face the same problem. maybe could ask for @rshriram for more details.

pilot/pkg/networking/core/v1alpha3/route/route.go
ramaraochavali: Handled this and added in https://github.com/istio/istio/pull/34137 as this PR looks like might need more discussion.

ramaraochavali:
Closing this because I think it does not make sense to support this at mesh. I even think we should support *.com etc at mesh (possibly at gateway also). We can add validations later.

PR #33905, opened 2021-07-08
Add more validation for MeshConfig

ChenglongGao327:
add validation for
ProxyHttpPort/IngressControllerMode/AccessLogEncoding/OutboundTrafficPolicy/DefaultServiceExportTo/DefaultVirtualServiceExportTo/DefaultDestinationRuleExportTo/H2UpgradePolicy/PathNormalization

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -1474,6 +1474,44 @@ func ValidateMeshConfig(mesh *meshconfig.MeshConfig) (errs error) {
 		errs = multierror.Append(errs, multierror.Prefix(err, "invalid proxy listen port:"))
 	}
 
+	if mesh.ProxyHttpPort != 0 {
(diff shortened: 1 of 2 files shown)


pkg/config/validation/validation.go
```diff
@@ -1507,6 +1545,67 @@ func ValidateMeshConfig(mesh *meshconfig.MeshConfig) (errs error) {
 	return
 }
 
+func validateIngressControllerMode(mode meshconfig.MeshConfig_IngressControllerMode) error {
```
ramaraochavali, line 1548: Is n't this an enum? How will it have any other value?

ramaraochavali (changes requested):
It seems most of them are enums and why do we need additional validations for them?

PR #33907, opened 2021-07-08
Use set.Contains to check grpc msg

shamsher31:
Follow up for https://github.com/istio/istio/pull/33893


ramaraochavali (changes requested):
We want to check partial messages only - not exact message. So the existing code looks correct.

shamsher31:
Sorry, I miss understood.

PR #33908, opened 2021-07-08
modify some codes of pkg/kube

gy95:
1. fix some comment errors, comply with programming specifications
2. fix some log errors

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/kube/inject/inject.go
+++ b/pkg/kube/inject/inject.go
@@ -147,7 +147,6 @@ const (
 )
 
 // UnmarshalConfig unmarshals the provided YAML configuration, while normalizing the resulting configuration
-// nolint: staticcheck
 func UnmarshalConfig(yml []byte) (Config, error) {
 	var injectConfig Config
 	if err := yaml.Unmarshal(yml, &injectConfig); err != nil {
@@ -465,10 +464,10 @@ func injectionStatus(pod *corev1.Pod) *SidecarInjectionStatus {
 
 	// default case when injected pod has explicit status
 	var iStatus SidecarInjectionStatus
-	if err := json.Unmarshal(statusBytes, &iStatus); err == nil {
-		return &iStatus
+	if err := json.Unmarshal(statusBytes, &iStatus); err != nil {
+		return nil
 	}
-	return nil
+	return &iStatus
 }
 
 func parseTemplate(tmplStr string, funcMap map[string]interface{}, data SidecarTemplateData) (bytes.Buffer, error) {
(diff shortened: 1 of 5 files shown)


pkg/kube/inject/inject.go
```diff
@@ -465,10 +464,10 @@ func injectionStatus(pod *corev1.Pod) *SidecarInjectionStatus {
 
 	// default case when injected pod has explicit status
 	var iStatus SidecarInjectionStatus
-	if err := json.Unmarshal(statusBytes, &iStatus); err == nil {
-		return &iStatus
+	if err := json.Unmarshal(statusBytes, &iStatus); err != nil {
```
ramaraochavali, line 467: What is the intent of this change?

pkg/kube/secretcontroller/secretcontroller.go
```diff
@@ -141,7 +141,7 @@ func (c *ClusterStore) Get(secretKey string, clusterID cluster.ID) *Cluster {
 	return c.remoteClusters[secretKey][clusterID]
 }
 
-// Get existing clusters registered for the given secret
+// GetExistingClustersFor get existing clusters registered for the given secret
```
ramaraochavali, line 144: GetExistingClustersFor returns

pkg/kube/inject/inject.go
gy95: IMO, the go programming specification is first process error circumstance , then process normal circumstance, so do the change

pkg/kube/secretcontroller/secretcontroller.go
gy95: ok

pkg/kube/inject/validate.go
```diff
@@ -163,7 +163,7 @@ func splitPorts(portsString string) []string {
 func parsePort(portStr string) (int, error) {
 	port, err := strconv.ParseUint(strings.TrimSpace(portStr), 10, 16)
 	if err != nil {
-		return 0, fmt.Errorf("failed parsing port '%d': %v", port, err)
+		return 0, fmt.Errorf("failed parsing port '%s': %v", portStr, err)
```
howardjohn, line 166: What problem is this solving?

gy95: > What problem is this solving?

if err is not nil, port will be always `0` , so we should return the origin `string` format, but not parsed `int` format

hzxuzhonghu (approved):
/ok-to-test

PR #33917, opened 2021-07-08
Update helm integration tests to use recommended approach when using …

jacob-delgado:
…helm upgrades

Helm does not upgrade or remove CRDs on upgrade. Run kubectl apply -f on the base charts CRDs folder to make sure the latest version of Istio is upgraded properly.

Update: The charts checked in were from 1.9.0 and not 1.10.0. This was a bug and even though the proper images were being downloaded, the charts themselves were from an older release.

[ ] Docs
[X] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[X] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


tests/integration/helm/upgrade/util.go
```diff
@@ -115,6 +125,14 @@ func deleteIstio(cs cluster.Cluster, h *helm.Helm) error {
 		return fmt.Errorf("wating for istio namespace to be deleted: %v", err)
 	}
 
+	// CRDs are not removed as part of the helm uninstall process
+	execCmd := fmt.Sprintf("kubectl delete -n %d -f %d", helmtest.IstioNamespace,
```
howardjohn, line 129: I don't think we should recommend this or do it in tests, this will make pretty large breaking changes to users clusters. This is a bad in docs (wipe out etcd essentially), and also annoying or even dangerous in tests (when running locally especially)

jacob-delgado:
See https://github.com/istio/istio.io/pull/10074

tests/integration/helm/upgrade/util.go
jacob-delgado: You are correct. People testing this in their clusters could also get bad behavior (e.g. OpenShift) if run locally. It's not a good practice and I've removed this line. The docs don't have this at all either. They do say to make a backup, but not to do a delete on any CRDs.

howardjohn (approved):

PR #33920, opened 2021-07-08
Support custom DNS alt hosts

nmittler:
This is necessary to support the Kubernetes MCS `clusterset.local` host. This is only preliminary work. Future PRs will provide full support for the alias.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


pkg/dns/client/dns_test.go
```diff
@@ -419,7 +444,14 @@ func initDNS(t test.Failer) *LocalDNSServer {
 	}
 	testAgentDNS.resolvConfServers = []string{srv}
 	testAgentDNS.StartDNS()
-	testAgentDNS.searchNamespaces = []string{"ns1.svc.cluster.local", "svc.cluster.local", "cluster.local"}
+	testAgentDNS.searchNamespaces = []string{
```
howardjohn, line 447: searchNamespaces comes from k8s configuring /etc/resolv.conf. In real world, how is this going to work?

nmittler: I didn't actually need this change.

nmittler:
@howardjohn PTAL

pkg/dns/proto/nds.proto
```diff
@@ -20,14 +20,24 @@ package istio.networking.nds.v1;
...
+
+        // The k8s service name. Only applies when registry=`Kubernetes`
         string shortname = 3;
+
+        // The k8s namespace for the service. Only applies when registry=`Kubernetes`
         string namespace = 4;
+
+        // List of alternate hosts to map to the IPs.
+        // Only applies when registry=`Kubernetes`
+        repeated string alt_hosts = 5;
```
howardjohn, line 37: Thinking about compatibility:
* If we have new istiod, we will send the config and agent will silently ignore. No problem
* If we have new proxy (not supported technically), Istiod will not send alt_hosts. No problem
So LGTM

howardjohn (approved):

pkg/dns/proto/nds.proto
nmittler: yeah exactly

PR #33921, opened 2021-07-08
envoy: set so_reuseport

kyessenov:
Signed-off-by: Kuat Yessenov <kuat@google.com>

In preparation for new defaults, set SO_REUSEPORT on virtual listener sockets. This enables kernel load balancing for TCP accepts instead of in-process implementation. The high risk area is hot-restart which is not used in istio.

Ref: https://github.com/envoyproxy/envoy/pull/17259

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/listener_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener_builder.go
@@ -384,6 +384,7 @@ func (lb *ListenerBuilder) buildVirtualOutboundListener(configgen *ConfigGenerat
 		UseOriginalDst:   proto.BoolTrue,
 		FilterChains:     filterChains,
 		TrafficDirection: core.TrafficDirection_OUTBOUND,
+		ReusePort:        true,
 	}
 	accessLogBuilder.setListenerAccessLog(lb.push, lb.node, ipTablesListener)
 	lb.virtualOutboundListener = ipTablesListener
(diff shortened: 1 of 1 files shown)


mandarjog:
@kyessenov is this to start testing istio with this change before the change lands upstream?

I think we will need to provide a way to use old behaviour if there are any gremlins later.

mandarjog (approved):
Lgtm
To get a head start on testing.
Do we want this change in 1.11? I don’t think so.

kyessenov:
I think it's a safe change on Linux. This is only affecting sidecars with a few workers and isolated network namespaces. It's less restrictive so it's unlikely to break someone. The only issue might be kernel interaction but it's been long enough that widely popular kernels are up to date.

kyessenov:
Putting hold so more people provide feedback.

lambdai (approved):
Can you update the description? 
s/set SO_REUSEPORT on virtual sockets/set SO_REUSEPORT on virtual listeners/

Explicitly set to true will shortcut the branch to use the Envoy runtime flag in the ref envoy PR. But we should build some confidence of the reuseport behavior prior to istio release

howardjohn (commented):
Is there any reason we only do it on virtual listener? I think the envoy PR applies to all listeners(?)

kyessenov:
/retest

@howardjohn What other listeners do you have in mind? Ingress might be interesting but given that it's usually high concurrency, there will be significant behavior changes. Sidecars are usually not using all cores. Health endpoints should not be shared I think between multiple envoys since that defeats their purpose.

howardjohn:
I was thinking of gateway specifically. My understanding is once the envoy
PR merged those will start using reuseport?

So it seems like that's either a good thing and we can explicitly enable it
not to get some early testing, or it's not a good thing and we should
explicitly disable it?

I am fine with not doing it in this PR, I was mostly just curious if it was
intentional not set

kyessenov:
@howardjohn There are more unknowns unknowns with gateways so we need to keep it separate. We should probably do it, but I'm not as confident that it won't break someone as this change.

kyessenov:
multi-cluster tests are very flakey.
/retest

kyessenov:
6 times and counting
/retest

howardjohn:
Given https://github.com/envoyproxy/envoy/pull/17259 merged, do we still need this?

Issue #32437, opened 2021-04-23
Improve CNI supportability


bianpengyuan:
Currently CNI is not very easy to debug. This issue various supportability improvements that will be needed for CNI beta graduation.

* Add label / prefix to CNI binary (istio-cni, istio-iptable) log, so that they can be easily filtered out from other kubelet logs.
* Collect better signals from istio-cni and istio-iptable plugin. The only CNI related metric we collected right now is pod repaired. We need to do some research on whether it is possible to collect metrics from cni binaries. Some ideas such as cni binaries write to a host file, istio-install periodically read and parse the file, and generates z pages or stats based on that. A default dashboard should also be provided. This might borrow the Calico's [logging solution](https://github.com/projectcalico/cni-plugin/blob/7495c0279c34faac315b82c1838bca638e23dbbe/internal/pkg/utils/utils.go#L710).
* istioctl integration, such as command to collect logs, get pod repaired, and bug report tools, etc.

bianpengyuan:
cc @mandarjog @therealmitchconnors @howardjohn @stewartbutler @ostromart  Please add anything that you think will be valuable for CNI supportability.

bianpengyuan:
This is done.

believening:
teeToUDSServer only returns a zapcore.Core, but check the err

https://github.com/istio/pkg/blob/fbcf820ec6e9beb2de2928e83d70d103c920b77f/log/config.go#L354-L364
```	
	if options.teeToUDSServer {
		// build uds core.
		core = teeToUDSServer(core, options.udsSocketAddress, options.udsServerPath)
		if err != nil {
			return err
		}
		captureCore = teeToUDSServer(captureCore, options.udsSocketAddress, options.udsServerPath)
		if err != nil {
			return err
		}
	}
```

PR #33926, opened 2021-07-09
Add UDS log to CNI plugin.

bianpengyuan:
For #32437

Make CNI plugin also sends log to CNI install container. This PR also contains a fix to make log level really work for CNI network plugin, and a clean up of plugin log messages to make it really usable. Also make CNI integration test sends log to CNI daemonset.

At info level, CNI network plugin logs looks like the follow in the cni container for a pod: https://gist.githubusercontent.com/bianpengyuan/a7cabdc1651df1dd3709fd386176e0e0/raw/7f1760bb7f01f9b7347a451bf6cf0989ed6ccb53/cni-info.log. At debug level, CNI network plugin logs looks like: https://gist.githubusercontent.com/bianpengyuan/1e3a4b905935e92b35c05cd386d76d57/raw/842fc2c68cab38895319c58c2c72c019262fa9ff/cni-debug.log


tests/integration/iop-cni-defaults.yaml
```diff
@@ -0,0 +1,25 @@
...
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+spec:
+  components:
+    cni:
+      namespace: kube-system
+      enabled: true
+      k8s:
+        env:
+          - name: LOG_UDS_ADDRESS
```
howardjohn, line 11: Why not the default? Is this just for testing until its stable?

releasenotes/notes/cni-uds-log.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+issue:
+  - 32437
+
+releaseNotes:
+  - |
+    **Added** CNI network plugin log collection at CNI Daemonset via UDS.
```
howardjohn, line 9:
```suggestion
    **Improved** CNI network plugin to send logs to the CNI daemonset. This allows viewing CNI logs using `kubectl logs`, instead of looking at kubelet logs.
```

Tried to make it more user-facing

howardjohn (commented):
Looks good!

howardjohn:
Played around with this locally, we end up with a lot of logs like:
```
2021-07-09T15:26:37.662344Z     info    Running command: iptables-restore --noflush /tmp/iptables-rules-1625844397662294469.txt604761520
2021-07-09T15:26:37.667146Z     info    Writing following contents to rules file: /tmp/ip6tables-rules-1625844397667116264.txt133047631

2021-07-09T15:26:37.667215Z     info    Running command: ip6tables-restore --noflush /tmp/ip6tables-rules-1625844397667116264.txt133047631
2021-07-09T15:26:37.669162Z     info    Running command: iptables-save
```

It would be ideal if we had context (pod name) on all of these

cni/pkg/log/uds.go
```diff
@@ -0,0 +1,128 @@
...
+	var body []byte
+	if req.Body != nil {
+		data, err := ioutil.ReadAll(req.Body)
+		if err != nil {
+			log.Errorf("Failed to read log report from cni plugin: %v", err)
+			return
+		}
+		body = data
+	}
+	l.processLogBody(body)
```
mandarjog, line 91: What is the purpose of calling processLogBody on a zero bytes body?
If there isn’t you can have an early exit and don’t need body and data as separate vars.

cni/pkg/plugin/iptables.go
```diff
@@ -57,10 +57,9 @@ func (ipt *iptables) Program(netns string, rdrct *Redirect) error {
 	log.Infof("nsenter args: %s", strings.Join(nsenterArgs, " "))
 	out, err := exec.Command("nsenter", nsenterArgs...).CombinedOutput()
 	if err != nil {
-		log.WithLabels("err", err, "out", out).Errorf("nsenter failed ")
-		log.Infof("nsenter out: %s", out)
+		log.Errorf("nsenter failed %v, with output %v", err, out)
```
mandarjog, line 60: Looks like you will have to plumb log context all the way thru from cni, otherwise pod name and other params will not be available.

cni/cmd/istio-cni/main.go
```diff
@@ -29,13 +29,13 @@ import (
...
-	loggingOptions.OutputPaths = []string{"stderr"}
-	loggingOptions.JSONEncoding = true
-	if err := log.Configure(loggingOptions); err != nil {
+	if err := log.Configure(plugin.GetLoggingOptions("")); err != nil {
 		os.Exit(1)
 	}
 	// TODO: implement plugin version
 	skel.PluginMain(plugin.CmdAdd, plugin.CmdCheck, plugin.CmdDelete, version.All,
 		fmt.Sprintf("CNI plugin istio-cni %v", istioversion.Info.Version))
+	if err := log.Sync(); err != nil {
```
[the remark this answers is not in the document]
bianpengyuan: No the http client which log.Sync uses has default 100ms timeout. I added a comment to clarify.

cni/pkg/log/uds.go
```diff
@@ -0,0 +1,128 @@
...
+			log.Debug("CNI log server terminated")
+		}
+	}()
+
+	return nil
+}
+
+func (l *UDSLogger) handleLog(w http.ResponseWriter, req *http.Request) {
+	var body []byte
+	if req.Body != nil {
```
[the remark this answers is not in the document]
bianpengyuan: thanks! done.

tests/integration/iop-cni-defaults.yaml
bianpengyuan: Yeah, more testing, and I also thought this will need a value option, which I want to avoid at least for now. But look again, we could just hard code the log volume. If user want to tune the path, they can edit the overlay and env variable as wanted. I've updated it to be enabled by default.

cni/pkg/log/uds.go
bianpengyuan: Done.

bianpengyuan:
@howardjohn @mandarjog re pod identifier for iptables log, I've added two separator log lines to denote start and end of iptables output for a pod. https://gist.githubusercontent.com/bianpengyuan/28de8f4a7c6d943b9d63ee9431a456c8/raw/faceb7d46257cfe5f90aa7fa32fa3c78291df458/cni-log-new.log is what it looks like now.

CNI/Iptables logs from different pods won't interleave because
1) CNI/iptables logs for a pod is reported in a single request
2) the logger locks the log printing for each request.

If we find out logs from CNI/IPTables is too large and needs to be divided into separated requests, we can revisit and think about ways to add identifier to each log lines.

releasenotes/notes/cni-uds-log.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+issue:
+  - 32437
+
+releaseNotes:
+  - |
+    **Improved** CNI network plugin to send logs to the CNI daemonset. This allows viewing CNI logs using `kubectl logs`, instead of looking at kubelet logs.
```
mandarjog, line 9: This is surely an improvement.
How will users that used to look at kubelet logs, know to look here?
It is a `change`, not just improvement.

howardjohn: Should we log in both places?

bianpengyuan: Istio CNI doc will be revamped, and this will be clearly documented. I think there should be a release notes about CNI beta graduation, as part of headlines for 1.11 release, and this will be mentioned there as well, not just change notes.

bianpengyuan: We log in both places. The UDS code paths sends a copy to CNI daemonset.

bianpengyuan: > We log in both places

The unit test covers verification for this: https://github.com/istio/istio/pull/33926/files#diff-370d69585b4a3d56f5b5e0c3df20124efea862278838276bc5666474a59b4ee6R55

bianpengyuan:
@mandarjog @howardjohn gentle ping for review, thanks!

howardjohn (approved):

PR #33865, opened 2021-07-07
Fix ServiceInstance selection with labels

hzxuzhonghu:
This is a regression recently introduced.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

Issue #23821, opened 2020-05-13
TLS origination for egress traffic gets 503


JimmyCYJ:
**Bug description**
I am following the user guide and testing this step
https://preliminary.istio.io/docs/tasks/traffic-management/egress/egress-tls-origination/#tls-origination-for-egress-traffic

$kubectl exec -it $SOURCE_POD -c sleep -- curl -sL -o /dev/null -D - http://edition.cnn.com/politics
HTTP/1.1 503 Service Unavailable
date: Wed, 13 May 2020 18:32:28 GMT
server: envoy
content-length: 0

curl to https succeeds
$kubectl exec -it $SOURCE_POD -c sleep -- curl -sL -o /dev/null -D - https://edition.cnn.com/politics
HTTP/2 200 
content-type: text/html; charset=utf-8
x-servedbyhost: ::ffff:127.0.0.1

**Expected behavior**
curl command should return 200

**Steps to reproduce the bug**

**Version (include the output of `istioctl version --remote` and `kubectl version` and `helm version` if you used Helm)**
$istioctl version
client version: 1.6.0-beta.1
control plane version: 1.6.0-beta.1
data plane version: 1.6.0-beta.1 (3 proxies)

**How was Istio installed?**
$istioctl manifest apply

**Environment where bug was observed (cloud vendor, OS, etc)**
GKE

nrjpoddar:
I think we need to address this and shouldn’t release with this regression. @jacob-delgado @brian-avery @costinm feels like a release blocker. Don’t have time today to test/triage it.

jacob-delgado:
I confirmed that this does happen. I've attached all of my envoy configs to help make it easier.
[envoy-config.tar.gz](https://github.com/istio/istio/files/4632036/envoy-config.tar.gz)

lambdai:
The listener and http conn manager looks good. HCM has the route and referring to cluster `outbound|443|tls-origination|edition.cnn.com` as expected.

lambdai:
That cluster is originated from the virtual service. It is missing in envoy config.

lambdai:
My latest finding is that pilot attempted to generate the cluster but somehow pilot cannot fill the hostname for that strict DNS cluster. 
The strict dns cluster with no endpoint is considered a invalid cluster config. Pilot won't push to envoy.

At my side these pilot logs are dumped:
```
        "outbound|443|tls-origination|edition.cnn.com": {
            "proxy": "sleep-6bdb595bcb-nbv4l.default",
[...]

PR #33929, opened 2021-07-09
add warning if locality endpoints are missing for DNS clusters

ramaraochavali:
[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -128,10 +128,9 @@ func (cb *ClusterBuilder) buildSubsetCluster(opts buildClusterOpts, destRule *co
 		clusterType = cluster.Cluster_ORIGINAL_DST
 	}
 	if !(isPassthrough || clusterType == cluster.Cluster_EDS) {
-		if len(subset.Labels) != 0 {
-			lbEndpoints = cb.buildLocalityLbEndpoints(proxyNetworkView, service, opts.port.Port, []labels.Instance{subset.Labels})
-		} else {
-			lbEndpoints = cb.buildLocalityLbEndpoints(proxyNetworkView, service, opts.port.Port, nil)
+		lbEndpoints = cb.buildLocalityLbEndpoints(proxyNetworkView, service, opts.port.Port, []labels.Instance{subset.Labels})
+		if len(lbEndpoints) == 0 {
+			log.Warnf("locality endpoints missing for cluster %s", subsetClusterName)
 		}
 	}
 
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -130,7 +130,7 @@ func (cb *ClusterBuilder) buildSubsetCluster(opts buildClusterOpts, destRule *co
 	if !(isPassthrough || clusterType == cluster.Cluster_EDS) {
 		lbEndpoints = cb.buildLocalityLbEndpoints(proxyNetworkView, service, opts.port.Port, []labels.Instance{subset.Labels})
 		if len(lbEndpoints) == 0 {
-			log.Warnf("locality endpoints missing for cluster %s", subsetClusterName)
+			log.Debugf("locality endpoints missing for cluster %s", subsetClusterName)
 		}
 	}
 


pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -128,10 +128,9 @@ func (cb *ClusterBuilder) buildSubsetCluster(opts buildClusterOpts, destRule *co
 		clusterType = cluster.Cluster_ORIGINAL_DST
 	}
 	if !(isPassthrough || clusterType == cluster.Cluster_EDS) {
-		if len(subset.Labels) != 0 {
-			lbEndpoints = cb.buildLocalityLbEndpoints(proxyNetworkView, service, opts.port.Port, []labels.Instance{subset.Labels})
-		} else {
-			lbEndpoints = cb.buildLocalityLbEndpoints(proxyNetworkView, service, opts.port.Port, nil)
+		lbEndpoints = cb.buildLocalityLbEndpoints(proxyNetworkView, service, opts.port.Port, []labels.Instance{subset.Labels})
+		if len(lbEndpoints) == 0 {
+			log.Warnf("locality endpoints missing for cluster %s", subsetClusterName)
```
howardjohn, line 133: When does this happen and what should a user do if they see this warning?

ramaraochavali: This happens if subset clusters with DNS resolution does not have members - happened for us in Service Entry (primarily because of this regression #33865  - it took quite some time to figure out that is the problem.

howardjohn: Hmm, my only concern is that its not actually a misconfig/bug to have zero, so there may be quite a bit of spam?

ramaraochavali: >  I am imaging you see 503 UH, then can look at the cluster from config dump and see no endpoints and figure out something wrong.

Yes. We are seeing 503s and we are getting "no_cluster" incremented on the route.  This cluster is not available in config_dump - I suspect because Envoy may be rejecting it as it is a STRICT_DNS cluster with no endpoints - I could not find that in logs. I could not find why the cluster was not created. That was the motivation. I agree it is not a misconfiguration but a bug in Istio. If Envoy NACKs are available this is is not needed though.

howardjohn:
If it's NACK we should already have logs in both istiod and agent, we well
as magic?

ramaraochavali:
yes .NACK logs are available in Istiod but I could not find it. Let me check again if I can reproduce it.

ramaraochavali:
I could not find logs indicating NACK or any thing. But changed this to debug so that it may be useful in debug.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -128,10 +128,9 @@ func (cb *ClusterBuilder) buildSubsetCluster(opts buildClusterOpts, destRule *co
 		clusterType = cluster.Cluster_ORIGINAL_DST
 	}
 	if !(isPassthrough || clusterType == cluster.Cluster_EDS) {
-		if len(subset.Labels) != 0 {
-			lbEndpoints = cb.buildLocalityLbEndpoints(proxyNetworkView, service, opts.port.Port, []labels.Instance{subset.Labels})
-		} else {
-			lbEndpoints = cb.buildLocalityLbEndpoints(proxyNetworkView, service, opts.port.Port, nil)
```
howardjohn, line 134: where did this go?

howardjohn (commented):
debug log LGTM but this seems to change the behavior? I recall a release regression doing something very similar so a bit nervous about the change

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
ramaraochavali: lbEndpoints = cb.buildLocalityLbEndpoints(proxyNetworkView, service, opts.port.Port, []labels.Instance{subset.Labels})

It is replaced with this single call. Basically if else not needed here.

howardjohn: I see, so `nil` and `[]labels.Instance` are equivilent. Makes sense. Let me double check the original issue I am thinking of, but looks fine

howardjohn (approved):
Issue I was thinking of was https://github.com/istio/istio/issues/23821#issuecomment-629500686

The code has changed so this looks extremely similar, but is fine I think.

Issue #12503, opened 2019-03-14
Update gateways chart to list port 15443 (or some other always-on port) first in Service entries


venezia:
**Describe the feature request**
Update helm chart (or other possible vectors of installing istio) such that 15443 is the first port listed in the `ingressgateway` kubernetes service.

**Describe alternatives you've considered**
Out-of-band updating health check on ELB, is not fun

**Additional context**
In AWS environment, the first port on a service record is used for a health check.  In a default istio environment, the ingressgateway's behavior on port 80 is to reject connections (until a VirtualService references the ingressgateway at port 80) - this will result in the ELB showing up as unhealthy when this really is not the case.

In Istio 1.1, multicluster service meshes can be created on an AWS environment.  A remote server may not be exposing any "public" services through the ingress gateway, but might allow connections in through port 15443 for the mutual TLS connection between istio control planes.

Such an environment could be "configured as intended" not actually working "correctly" because the ingressgateway service will be down (as far as ELB health checks are concerned) because the ELB health check is looking for valid TCP connections on port 80 (by default)

The user experience might be lacking, as a person might get frustrated that examples in the documentation are not working, etc. and not appreciate that the problem is with the ELB's health check.

Whatever is listed as the first port in a Service of type LoadBalancer in an AWS cloud enabled environment is picked to be the port used for health checks.  If 15443 will always accept connections (or some other port) it would be desirable to list this port first for AWS customers

In other words, let's try to enhance the helm charts so that AWS customers will not be frustrated with their istio experience.

nrjpoddar:
Thanks for filing this @venezia. We had similar problems and was quite painful to pin down the issue. I have a PR out to address this.

nrjpoddar:
This can be closed as 1.1.3 has the fix.

TqrHsn:
How to disable 15020 to be exposed from cluster? In our scenario we cannot expose any ports other than 443.

iusergii:
> How to disable 15020 to be exposed from cluster? In our scenario we cannot expose any ports other than 443.

Same here. @nrjpoddar Is it possible?

nrjpoddar:
[...]

PR #33930, opened 2021-07-09
Perform strategic merge patch of service ports by port and protocol

### LINKED ISSUE #33841: Service manifest rendered for the gateway with same port number but different protocol (TCP & UDP) is incorrect
**Bug description**
I have the following `IstioOperator` manifest. As you can see the bookinfo-ingress has two ports - both 443, but different protocols. The use case is that I would like to have both TCP and UDP listeners exposed on that port so that for an HTTPS listener I can create a corresponding QUIC listener to support HTTP/3 (Context - https://github.com/istio/istio/pull/33817) 
```yml
kind: IstioOperator
metadata:
  name: istio-with-quic
  namespace: istio-system
spec:
  meshConfig:
    accessLogFile: /dev/stdout
  hub: localhost:5000
  tag: istio-quic
  profile: minimal
  components:
    ingressGateways:
    - name: bookinfo-ingress
      namespace: bookinfo
      enabled: true
      label:
        gateway: bookinfo
        quic: enabled
      k8s:
        service:
          type: LoadBalancer
          ports:
          - name: https-tcp
            port: 443
            targetPort: 8443
            protocol: TCP
          - name: http3-quic
            port: 443
            targetPort: 8443
            protocol: UDP
  values:
    global:
      imagePullPolicy: Always
      logging:
        level: "all:debug"
    pilot:
      env:
        PILOT_ENABLE_QUIC_LISTENERS: "true"
```

**Actual**: The corresponding service rendered
```yml
apiVersion: v1
kind: Service
metadata:
  annotations: null
  labels:
    app: istio-ingressgateway
    gateway: bookinfo
    install.operator.istio.io/owning-resource: unknown
    istio: ingressgateway
    istio.io/rev: default
    operator.istio.io/component: IngressGateways
    quic: enabled
    release: istio
  name: bookinfo-ingress
  namespace: bookinfo
spec:
  ports:
  - name: http3-quic
    port: 443
    protocol: UDP
    targetPort: 8443
  - name: http3-quic
    port: 443
    protocol: UDP
    targetPort: 8443
  selector:
    app: istio-ingressgateway
    gateway: bookinfo
    istio: ingressgateway
    quic: enabled
  type: LoadBalancer
```

**Expected**: Expected service
```yml
apiVersion: v1
kind: Service
metadata:
  annotations: null
  labels:
    app: istio-ingressgateway
    gateway: bookinfo
    install.operator.istio.io/owning-resource: unknown
    istio: ingressgateway
    istio.io/rev: default
    operator.istio.io/component: IngressGateways
    quic: enabled
    release: istio
  name: bookinfo-ingress
  namespace: bookinfo
spec:
  ports:
  - name: https-tcp
    port: 443
    protocol: TCP
    targetPort: 8443
  - name: http3-quic
    port: 443
    protocol: UDP
    targetPort: 8443
  selector:
    app: istio-ingressgateway
    gateway: bookinfo
    istio: ingressgateway
    quic: enabled
  type: LoadBalancer
```

[ ] Docs
[X] Installation
[ ] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

1.11-dev built from the master branch

**How was Istio installed?**
istioctl

su225:
Fixes #33841

The default Kubernetes behaviour is to do it only by port. In this case ports can be the same, but with different protocol. The output of Kubernetes' merge patch is incorrect. As it is much harder to fix in Kubernetes, this is a workaround to fix it at Istio manifest generation level.

[ ] Docs
[X] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/operator/cmd/mesh/testdata/manifest-generate/output/ingressgateway_k8s_settings.golden.yaml
+++ b/operator/cmd/mesh/testdata/manifest-generate/output/ingressgateway_k8s_settings.golden.yaml
@@ -486,10 +486,6 @@ metadata:
 spec:
   externalTrafficPolicy: Local
   ports:
-  - name: status-port
-    port: 15021
-    protocol: TCP
-    targetPort: 15021
   - name: http2
     port: 80
     protocol: TCP
@@ -498,6 +494,10 @@ spec:
     port: 443
     protocol: TCP
     targetPort: 8443
+  - name: status-port
+    port: 15021
+    protocol: TCP
+    targetPort: 15021
   selector:
     app: istio-ingressgateway
     istio: ingressgateway
@@ -521,10 +521,6 @@ metadata:
 spec:
   externalTrafficPolicy: Local
   ports:
-  - name: status-port
-    port: 15021
-    protocol: TCP
-    targetPort: 15021
   - name: http2
     port: 80
     protocol: TCP
@@ -533,6 +529,10 @@ spec:
     port: 443
     protocol: TCP
     targetPort: 8443
+  - name: status-port
+    port: 15021
+    protocol: TCP
+    targetPort: 15021
   selector:
     app: istio-ingressgateway
     istio: ingressgateway
--- a/operator/cmd/mesh/testdata/manifest-generate/output/ingressgateway_k8s_settings.golden.yaml
+++ b/operator/cmd/mesh/testdata/manifest-generate/output/ingressgateway_k8s_settings.golden.yaml
@@ -486,6 +486,10 @@ metadata:
 spec:
   externalTrafficPolicy: Local
   ports:
+  - name: status-port
+    port: 15021
+    protocol: TCP
+    targetPort: 15021
   - name: http2
     port: 80
     protocol: TCP
@@ -494,10 +498,6 @@ spec:
     port: 443
     protocol: TCP
     targetPort: 8443
-  - name: status-port
-    port: 15021
-    protocol: TCP
-    targetPort: 15021
   selector:
     app: istio-ingressgateway
     istio: ingressgateway
@@ -521,6 +521,10 @@ metadata:
 spec:
   externalTrafficPolicy: Local
   ports:
+  - name: status-port
+    port: 15021
+    protocol: TCP
+    targetPort: 15021
   - name: http2
     port: 80
     protocol: TCP
@@ -529,10 +533,6 @@ spec:
     port: 443
     protocol: TCP
     targetPort: 8443
-  - name: status-port
-    port: 15021
-    protocol: TCP
-    targetPort: 15021
   selector:
     app: istio-ingressgateway
     istio: ingressgateway
--- a/operator/pkg/translate/strategic_port_merge_test.go
+++ b/operator/pkg/translate/strategic_port_merge_test.go
@@ -1,3 +1,17 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package translate
 
 import (
--- a/operator/pkg/translate/strategic_port_merge_test.go
+++ b/operator/pkg/translate/strategic_port_merge_test.go
@@ -1,0 +1,99 @@
+package translate
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+var (
+	httpsPort = &v1.ServicePort{
+		Name:       "https",
+		Protocol:   v1.ProtocolTCP,
+		Port:       443,
+		TargetPort: intstr.IntOrString{IntVal: 8443},
+	}
+	quicPort = &v1.ServicePort{
+		Name:       "http3-quic",
+		Protocol:   v1.ProtocolUDP,
+		Port:       443,
+		TargetPort: intstr.IntOrString{IntVal: 8443},
+	}
+	httpPort = &v1.ServicePort{
+		Name:       "http-port",
+		Protocol:   v1.ProtocolTCP,
+		Port:       80,
+		TargetPort: intstr.IntOrString{IntVal: 8080},
+	}
+	mysqlPort = &v1.ServicePort{
+		Name:     "mysql-port",
+		Protocol: v1.ProtocolTCP,
+		Port:     3306,
+	}
+)
+
+func TestStrategicPortMergeByPortAndProtocol(t *testing.T) {
+	for _, tt := range []struct {
+		name                string
+		basePorts           []*v1.ServicePort
+		overlayPorts        []*v1.ServicePort
+		expectedMergedPorts []*v1.ServicePort
+	}{
+		{
+			name:                "both base and overlay are nil",
+			basePorts:           nil,
+			overlayPorts:        nil,
+			expectedMergedPorts: nil,
+		},
+		{
+			name:                "overlay is nil",
+			basePorts:           []*v1.ServicePort{httpPort, httpsPort, quicPort},
+			overlayPorts:        nil,
+			expectedMergedPorts: []*v1.ServicePort{httpPort, httpsPort, quicPort},
+		},
+		{
+			name:                "base is nil",
+			basePorts:           nil,
+			overlayPorts:        []*v1.ServicePort{httpPort, httpsPort, quicPort},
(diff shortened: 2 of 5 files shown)


operator/cmd/mesh/testdata/manifest-generate/output/ingressgateway_k8s_settings.golden.yaml
```diff
@@ -498,6 +494,10 @@ spec:
     port: 443
     protocol: TCP
     targetPort: 8443
+  - name: status-port
```
howardjohn, line 497: This is a breaking change, order matters for AWS I think. It will use the first port for health checks

howardjohn (commented):
I am a bit concerned if this is not working in Kubernetes upstream then Helm, Kustomize, or `kubectl apply` install instructions will not work?

operator/cmd/mesh/testdata/manifest-generate/output/ingressgateway_k8s_settings.golden.yaml
su225: Ok. So...Is `status-port` the only special case to keep in mind or are there similar cases to keep in mind while ordering the ports?

su225: Updated the PR to handle 15021 and 15020 specially. (ref - https://github.com/istio/istio/issues/12503)

operator/pkg/translate/translate.go
```diff
@@ -243,13 +246,150 @@ func (t *Translator) OverlayK8sSettings(yml string, iop *v1alpha1.IstioOperatorS
 		if err != nil {
 			return "", err
 		}
+
+		// Apply the workaround for merging service ports with (port,protocol) composite
+		// keys instead of just the merging by port.
+		if strings.HasSuffix(inPath, "Service") {
```
ostromart, line 252: better to do an exact match on the last path element here.

operator/pkg/translate/translate.go
```diff
@@ -243,13 +246,150 @@ func (t *Translator) OverlayK8sSettings(yml string, iop *v1alpha1.IstioOperatorS
...
+// strategicMergePorts merges the base with the given overlay considering both
+// port and the protocol as the merge keys. This is a workaround for the strategic
+// merge patch in Kubernetes which only uses port number as the key. This causes
+// an issue when we have to expose the same port with different protocols.
+// See - https://github.com/kubernetes/kubernetes/issues/103544
+func strategicMergePorts(base, overlay []*v1.ServicePort) []*v1.ServicePort {
+	sortFn := func(ps []*v1.ServicePort) func(int, int) bool {
+		return func(i, j int) bool {
+			pi, pj := ps[i].Port, ps[j].Port
+			// The ports 15020 (metrics) and 15021 (status) need to be handled
```
ostromart, line 339: Please leave a TODO to remove this workaround and reference this PR in the mentioned issue.

operator/pkg/translate/strategic_port_merge_test.go
```diff
@@ -0,0 +1,141 @@
...
+			basePorts:           []*v1.ServicePort{istioHealthcheckPort, httpsPort, httpPort},
+			overlayPorts:        []*v1.ServicePort{httpsPort, httpPort},
+			expectedMergedPorts: []*v1.ServicePort{istioHealthcheckPort, httpPort, httpsPort},
+		},
+		{
+			name:                "metrics port is present",
+			basePorts:           []*v1.ServicePort{istioMetricsPort, httpsPort, httpPort},
+			overlayPorts:        []*v1.ServicePort{httpsPort, httpPort},
+			expectedMergedPorts: []*v1.ServicePort{istioMetricsPort, httpPort, httpsPort},
+		},
```
ostromart, line 132: need another case here where an existing port number is changed by overlay i.e. ensure we're not breaking existing function.

ostromart (commented):
We currently have our own mirror of many k8s protos, including Service. Could the same behavior be accomplished through strategic merge annotations on those protos?

operator/pkg/translate/translate.go
su225: Done - added a reference from #12503 to this PR

operator/pkg/translate/translate.go
su225: Done.

operator/pkg/translate/strategic_port_merge_test.go
su225: I didn't get it. Here, merging is by port and protocol, not the name. Or did you mean using `overlays` section?

operator/pkg/translate/translate.go
su225: I also added a TODO referring the original issue.

su225:
**UPDATE:** Without this fix, I found a workaround to solve the problem using overlays. While the YAML mentioned in the issue renders duplicate entries, adding port using overlays works!
```yml
spec:
  components:
    ingressGateways:
    - name: istio-ingressgateway
      namespace: istio-system
      enabled: true
      k8s:
        service:
          type: LoadBalancer
          ports:
          - name: https-tcp
            port: 443
            targetPort: 8443
            protocol: TCP
        overlays:
        - apiVersion: v1
          kind: Service
          name: istio-ingressgateway
          patches:
          - path: spec.ports.[-1]
            value:
              name: http3-quic
              port: 443
              targetPort: 8443
              protocol: UDP
```
This renders the service manifest properly even with existing versions of `istioctl`
```yml
apiVersion: v1
kind: Service
metadata:
  annotations: null
  labels:
    app: istio-ingressgateway
    install.operator.istio.io/owning-resource: unknown
    istio: ingressgateway
    istio.io/rev: default
    operator.istio.io/component: IngressGateways
    release: istio
  name: istio-ingressgateway
  namespace: istio-system
spec:
  ports:
  - name: https-tcp
    port: 443
    protocol: TCP
    targetPort: 8443
  - name: http3-quic
    port: 443
    protocol: UDP
    targetPort: 8443
  selector:
    app: istio-ingressgateway
    istio: ingressgateway
  type: LoadBalancer
```
However, I'd still like to include this because overlays should be used sparingly. It is bad UX to add ports like this when there is API is available just because of some bug.

operator/pkg/translate/translate.go
```diff
@@ -243,13 +247,157 @@ func (t *Translator) OverlayK8sSettings(yml string, iop *v1alpha1.IstioOperatorS
...
+	port     int32
+	protocol v1.Protocol
+}
+
+// strategicMergePorts merges the base with the given overlay considering both
+// port and the protocol as the merge keys. This is a workaround for the strategic
+// merge patch in Kubernetes which only uses port number as the key. This causes
+// an issue when we have to expose the same port with different protocols.
+// See - https://github.com/kubernetes/kubernetes/issues/103544
+func strategicMergePorts(base, overlay []*v1.ServicePort) []*v1.ServicePort {
```
howardjohn, line 336: Instead of special hack, can we just sort it based on original order? Something like
```go
func indexOf(element portWithProtocol, data []portWithProtocol) int {
	for k, v := range data {
		if element == v {
			return k
		}
	}
	return len(data)
}

// strategicMergePorts merges the base with the given overlay considering both
// port and the protocol as the merge keys. This is a workaround for the strategic
// merge patch in Kubernetes which only uses port number as the key. This causes
// an issue when we have to expose the same port with different protocols.
// See - https://github.com/kubernetes/kubernetes/issues/103544
func strategicMergePorts(base, overlay []*v1.ServicePort) []*v1.ServicePort {
	portPriority := []portWithProtocol{}
	for _, p := range base {
		portPriority = append(portPriority, portWithProtocol{p.Port, p.Protocol})
	}
	for _, p := range overlay {
		portPriority = append(portPriority, portWithProtocol{p.Port, p.Protocol})
	}
	sortFn := func(ps []*v1.ServicePort) func(int, int) bool {
		return func(i, j int) bool {
			ii := indexOf(portWithProtocol{ps[i].Port, ps[i].Protocol}, portPriority)
			ji := indexOf(portWithProtocol{ps[i].Port, ps[i].Protocol}, portPriority)
			return ii < ji
		}
	}
```

I am just slightly worried that we will break users, or at the very least cause unexpected diffs, by changing order

su225: Thank you!! This looks much better and predictable from the PoV of users. Including this. Yeah. In manifest generate test I had to change the order to pass it - https://github.com/istio/istio/blob/02f7faf64f42b041ca8c8c3d188a4a426d0a1a4b/operator/cmd/mesh/manifest-generate_test.go#L185-L187 . With this feedback incorporated I don't have to do that!

operator/pkg/translate/strategic_port_merge_test.go
su225: I added more cases
* overlay with port name changed (in overlay)
* overlay with different protocol (people don't notice the difference in behavior as long as they are using only TCP for the same protocol. This affects the case when same port, but different protocol is used. It was incorrect before anyways)
* same base and overlay with protocol omitted for overlay (protocol TCP must be present post-merging)
* same base and overlay with protocol omitted for base (protocol TCP must be present post-merging)

Let me know if I have to add more cases.

howardjohn (approved):
Nice!

howardjohn:
Put a hold since @ostromart seems to have open comments

ostromart (approved):

PR #33932, opened 2021-07-09
Minor change: Reduce unnecessary judgment

ChenglongGao327:
This function is just for constructing mode.Service[], MeshExternal should be the same as other parameters, and can be used after passing it, without needing to judge again

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pilot/pkg/serviceregistry/serviceentry/conversion.go
+++ b/pilot/pkg/serviceregistry/serviceentry/conversion.go
@@ -170,6 +170,7 @@ func convertServices(cfg config.Config) []*model.Service {
 	if serviceEntry.WorkloadSelector != nil {
 		labelSelectors = serviceEntry.WorkloadSelector.Labels
 	}
+
 	hostAddresses := []*HostAddress{}
 	for _, hostname := range serviceEntry.Hosts {
 		if len(serviceEntry.Addresses) > 0 {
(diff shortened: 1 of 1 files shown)


pilot/pkg/serviceregistry/serviceentry/conversion.go
```diff
@@ -191,19 +192,20 @@ func convertServices(cfg config.Config) []*model.Service {
...
 	}
+	meshExternal := serviceEntry.Location == networking.ServiceEntry_MESH_EXTERNAL
 
-	return buildServices(hostAddresses, cfg.Namespace, svcPorts, serviceEntry.Location, resolution,
+	return buildServices(hostAddresses, cfg.Namespace, svcPorts, meshExternal, resolution,
 		exportTo, labelSelectors, serviceEntry.SubjectAltNames, creationTime, cfg.Labels)
 }
 
-func buildServices(hostAddresses []*HostAddress, namespace string, ports model.PortList, location networking.ServiceEntry_Location,
+func buildServices(hostAddresses []*HostAddress, namespace string, ports model.PortList, meshExternal bool,
```
ramaraochavali, line 201: What is the purpose of this change? Passing location seems better so that `buildServices` can use it for other purpose if needed?

howardjohn:
Please indicate what is being improved in your PRs in the PR description in the future to help us review better. Thanks!

pilot/pkg/serviceregistry/serviceentry/conversion.go
ChenglongGao327: > What is the purpose of this change? Passing location seems better so that `buildServices` can use it for other purpose if needed?

I think this function is just for constructing mode.Service[], MeshExternal should be the same as other parameters, and can be used after passing it, without needing to judge again. I haven't thought of other scenarios that Locality can use yet.

ChenglongGao327:
> Please indicate what is being improved in your PRs in the PR description in the future to help us review better. Thanks!

OK, i will add the detailed description for PR in the future

pilot/pkg/serviceregistry/serviceentry/conversion.go
ramaraochavali: I understand it is not used now but can be in future and I see no benefit of judging it before hand.

PR #33945, opened 2021-07-09
Special handling for the case agent is started as UID=0

costinm:
- DNS server can listen on port 53, and use the no-iptables mode (
alternateve resolv.conf and /etc/resolv.conf using localhost )
- envoy will be start with UID=1337
- DNS will answer to queries before XDS is connected ( so we can resolve
istiod address).

This mode is useful in environments where iptables is not available or
if iptables 'raw' is not available ( used for DNS interception ), for
example in docker containers.

Change-Id: I4d6e550f173056d198326d34aad21fa85e057af3

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


pkg/dns/client/dns.go
```diff
@@ -187,11 +208,17 @@ func (h *LocalDNSServer) ServeDNS(proxy *dnsProxy, w dns.ResponseWriter, req *dn
 	lp := h.lookupTable.Load()
 	hostname := strings.ToLower(req.Question[0].Name)
 	if lp == nil {
-		log.Debugf("dns request for host %q before lookup table is loaded", hostname)
-		response = new(dns.Msg)
-		response.SetReply(req)
-		response.Rcode = dns.RcodeServerFailure
-		_ = w.WriteMsg(response)
+		if os.Getuid() != 0 {
```
howardjohn, line 211: nit: Getuid makes a full syscall, seems like we can just check once

howardjohn: And it seems like this should be tied to using /var/lib/istio/resolv.conf as well?

howardjohn (commented):
Overall makes sense, although I am not sure UID=0 is the best indicator to use this mode? You could run as 0 and have the dns iptables, and you could run as non-0 and want this mode

pkg/dns/client/dns.go
costinm: In the big scheme of startup thing - a syscall is the last thing to optimize :-)

Running as UID=0 with iptables is pretty bad for envoy - will loop. We can add an extra env or flag - but if we start at root 
and the flag is not set, bad things can happen ( not for DNS, but for envoy exec ).

I'm a bit reluctant to add even more flags and envs - I think relying on the environment works better, less changes
to miss-configure. I think running as UID 0 is pretty clear user intent.

costinm:
Well, looking at the test failures ( it seems they run as root ) - I will add a setting :-)

pkg/dns/client/dns.go
```diff
@@ -187,11 +216,17 @@ func (h *LocalDNSServer) ServeDNS(proxy *dnsProxy, w dns.ResponseWriter, req *dn
 	lp := h.lookupTable.Load()
 	hostname := strings.ToLower(req.Question[0].Name)
 	if lp == nil {
-		log.Debugf("dns request for host %q before lookup table is loaded", hostname)
-		response = new(dns.Msg)
-		response.SetReply(req)
-		response.Rcode = dns.RcodeServerFailure
-		_ = w.WriteMsg(response)
+		if strings.HasSuffix(h.addr, ":53") {
```
howardjohn, line 219: nit: would be good to add a comment why its safe to resolve before loaded here

howardjohn: Also very nit: may prefer a function/struct field to store "directMode" (or better name). Or at least a comment that :53 == direct mode

pkg/dns/client/proxy.go
```diff
@@ -45,24 +48,33 @@ func newDNSProxy(protocol string, resolver *LocalDNSServer) (*dnsProxy, error) {
 		protocol: protocol,
 		resolver: resolver,
 	}
+	addr := resolver.addr
+	if addr == "" {
+		addr = "localhost:15053"
+	}
+	if strings.HasSuffix(addr, ":53") && os.Getuid() != 0 {
+		log.Error("DNS address :53 and not running as root, use default")
```
howardjohn, line 56: nit: isn't it technically possible to do this without root by giving the bind_serve cap?

pkg/dns/client/proxy.go
```diff
@@ -45,24 +48,33 @@ func newDNSProxy(protocol string, resolver *LocalDNSServer) (*dnsProxy, error) {
...
 	if protocol == "udp" {
-		p.server.PacketConn, err = net.ListenPacket("udp", "localhost:15053")
+		p.server.PacketConn, err = net.ListenPacket("udp", addr)
 	} else {
-		p.server.Listener, err = net.Listen("tcp", "localhost:15053")
+		p.server.Listener, err = net.Listen("tcp", addr)
 	}
 	if err != nil {
-		log.Errorf("Failed to listen on %s port 15053: %v", protocol, err)
+		log.Errorf("Failed to listen on %s port %s: %v", protocol, addr, err)
```
howardjohn, line 70:
```suggestion
		log.Errorf("Failed to listen on %s at %s: %v", protocol, addr, err)
```
very nit: its not just port now

howardjohn (approved):

Issue #33949, opened 2021-07-09
Support Kubernetes MCS host


nmittler:
The problem and proposed solution are defined in the [RFC](https://docs.google.com/document/d/1jTamFFXPhjZiLoQbspRaMW07nxE0tEN6qqqUUDqkTQI/edit#heading=h.lvpmdn3txu8s)

[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Additional context**

nmittler:
This is now effectively done. I've raised https://github.com/istio/istio/issues/35423 to track the Envoy work required to allow MCS to graduate from experimental.

PR #33950, opened 2021-07-09
Basic support for Kubernetes MCS host

nmittler:
This is the first phase of work for #33949.

This PR adds `clusterset.local` as an alias for `cluster.local` and resolves to the same IP. Since we're using the same IP, this change only impacts HTTP listeners and their routes.

The MCS host is also supported for DNS lookup.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/httproute.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute.go
@@ -419,33 +419,41 @@ func generateAltVirtualHosts(hostname string, port int, proxyDomain string) []st
 		return nil
 	}
 
+	// For Kubernetes services, If Kubernetes MCS host is enabled, also add a virtual host for 'clusterset.local'
+	if features.EnableMCSHost {
+		svcIndex := strings.LastIndex(hostname, ".svc.")
+		if svcIndex > 0 {
+			// It's a Kubernetes service. Add the virtual host for MCS.
+			mcsHost := hostname[:svcIndex+len(".svc.")] + mcsServiceDomain
+			vhosts = append(vhosts, mcsHost, domainName(mcsHost, port))
+		}
+	}
+
 	sharedDNSDomainParts := strings.Split(sharedDNSDomain, ".")
 	if len(strings.Split(uniqHostname, ".")) == 2 {
 		// This is the case of uniqHostname having namespace already.
 		dnsHostName := uniqHostname + "." + sharedDNSDomainParts[0]
 		vhosts = append(vhosts, uniqHostname, domainName(uniqHostname, port), dnsHostName, domainName(dnsHostName, port))
-	} else {
-		if strings.Contains(proxyDomain, ".svc.") {
-			// Derive the namespace from sharedDNSDomain and add virtual host.
-			namespace := sharedDNSDomainParts[0]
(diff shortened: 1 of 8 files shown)


pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -419,33 +419,41 @@ func generateAltVirtualHosts(hostname string, port int, proxyDomain string) []st
...
-			namespace := sharedDNSDomainParts[0]
-			if strings.HasPrefix(proxyDomain, namespace+".svc.") {
-				// Split the domain and add only for Kubernetes proxies.
-				vhosts = append(vhosts, uniqHostname, domainName(uniqHostname, port))
-				if len(sharedDNSDomainParts) > 1 {
-					dnsHostName := uniqHostname + "." + namespace + "." + sharedDNSDomainParts[1]
-					vhosts = append(vhosts, dnsHostName, domainName(dnsHostName, port))
-				}
-				hostNameWithNS := uniqHostname + "." + namespace
+	} else if strings.Contains(proxyDomain, ".svc.") {
```
nmittler, line 437: FYI this is just removing unnecessary nesting of `if` statements.

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -419,33 +419,41 @@ func generateAltVirtualHosts(hostname string, port int, proxyDomain string) []st
 		return nil
 	}
 
+	// For Kubernetes services, If Kubernetes MCS host is enabled, also add a virtual host for 'clusterset.local'
+	if features.EnableMCSHost {
+		svcIndex := strings.LastIndex(hostname, ".svc.")
```
stevenctl, line 424: So even if the `proxyDomain` is not `*.cluster.local` we will do this?

stevenctl (approved):
LGTM – curious how things would work for non http though

pilot/pkg/networking/core/v1alpha3/httproute.go
nmittler: yeah that's right ... even if there is a custom domain, we'll still create a `clusterset.local` host.

nmittler:
@stevenctl 

> LGTM – curious how things would work for non http though

Since the `clusterset.local` host resolves to the same IP, there should be effectively no change for TCP traffic. I can expand the test to verify.

pilot/pkg/networking/core/v1alpha3/httproute.go
howardjohn: I think this check is not fully safe. Probably fine for 99% of cases, but you could have `foo.svc.google.com`. It should only match if its `.svc.<cluster-domain>`. It looks like we are checking that a bit below?

howardjohn (commented):
LGTM, one small comment

howardjohn:
> Since the `clusterset.local` host resolves to the same IP, there should be effectively no change for TCP traffic. I can expand the test to verify.

*might* still need an SNI match? Although I think we only do that with explicit virtual service `sniHosts`, in which case its explicitly list, not expanded. so maybe not.

PR #33955, opened 2021-07-11
Add metrics for CNI installations and readiness.

bianpengyuan:
for https://github.com/istio/istio/issues/32437

### DIFF
--- a/cni/pkg/install/monitoring.go
+++ b/cni/pkg/install/monitoring.go
@@ -22,7 +22,7 @@ var (
 	resultCopyBinariesFailure     = "COPY_BINARIES_FAILURE"
 	resultReadSAFailure           = "READ_SERVICE_ACCOUNT_FAILURE"
 	resultCreateKubeConfigFailure = "CREATE_KUBECONFIG_FAILURE"
-	resultCreateCNIConfigFailure  = "CREATE_KUBECONFIG_FAILURE"
+	resultCreateCNIConfigFailure  = "CREATE_CNI_CONFIG_FAILURE"
 
 	stateLabel   = monitoring.MustCreateLabel("state")
 	stateReady   = "READY"
@@ -30,7 +30,7 @@ var (
 
 	cniInstalls = monitoring.NewSum(
 		"istio_cni_installs_total",
-		"Total number of CNI plugins installed by the ßIstio CNI installer",
+		"Total number of CNI plugins installed by the Istio CNI installer",
 		monitoring.WithLabels(resultLabel),
 	)
 
--- a/cni/pkg/install/monitoring.go
+++ b/cni/pkg/install/monitoring.go
@@ -30,7 +30,7 @@ var (
 		monitoring.WithLabels(resultLabel),
 	)
 
-	installState = monitoring.NewGauge(
+	installReady = monitoring.NewGauge(
 		"istio_cni_install_ready",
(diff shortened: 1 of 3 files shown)


cni/pkg/install/monitoring.go
```diff
@@ -0,0 +1,47 @@
...
+package install
+
+import "istio.io/pkg/monitoring"
+
+var (
+	resultLabel                   = monitoring.MustCreateLabel("result")
+	resultSuccess                 = "SUCCESS"
+	resultCopyBinariesFailure     = "COPY_BINARIES_FAILURE"
+	resultReadSAFailure           = "READ_SERVICE_ACCOUNT_FAILURE"
+	resultCreateKubeConfigFailure = "CREATE_KUBECONFIG_FAILURE"
```
howardjohn, line 24: Do you think these are too granular tags? I am guess these should ~never happen except if things are extremely broken. At that point, it may be more valuable to just diagnose based on logs?

I am not against this, just seems like we should have tags for expected errors, but not sure about unexpected ones?

cni/pkg/install/monitoring.go
```diff
@@ -0,0 +1,47 @@
...
+	resultCreateCNIConfigFailure  = "CREATE_KUBECONFIG_FAILURE"
+
+	stateLabel   = monitoring.MustCreateLabel("state")
+	stateReady   = "READY"
+	stateUnready = "UNREADY"
+
+	cniInstalls = monitoring.NewSum(
+		"istio_cni_installs_total",
+		"Total number of CNI plugins installed by the ßIstio CNI installer",
```
howardjohn, line 33:
```suggestion
		"Total number of CNI plugins installed by the Istio CNI installer",
```

cni/pkg/install/monitoring.go
```diff
@@ -0,0 +1,47 @@
...
+		monitoring.WithLabels(resultLabel),
+	)
+
+	installState = monitoring.NewGauge(
+		"istio_cni_install_state",
+		"Installation state of CNI plugin, one of [READY, UNREADY]",
```
howardjohn, line 39: Is there any standard prometheus pattern for exposing a boolean like this? I have seen `up` which feels a bit similar... seems like it might be more standard to have `istio_cni_install_ready` (or a better name) which is 0 for unready and 1 for ready (or reverse)?

I am not a prometheus expert, so maybe it doesn't make sense.

cni/pkg/install/monitoring.go
bianpengyuan: I think this will be useful for the case where log is not so easy to get, like in the vendor env.

I feel it is ok to track these errors at this granularity. If cardinality is totally OK (there aren't many) and the label is not confusing, make label a bit detailed seems handy to me.

cni/pkg/install/monitoring.go
bianpengyuan: Ideally the gauge could use string as value type, but unfortunately prom only allows numeric value. I cannot find any best practice, and we probably will only need a binary value here, and other binary metric just use 0/1 instead of labels so I updated this PR to match that. As long as we can easily get total number of ready/unready pod, we are good.

bianpengyuan:
@howardjohn Any more changes needed here?

bianpengyuan:
cc @richardwxn

howardjohn (approved):

Issue #33958, opened 2021-07-11
Go Routine leak in Test Agent 


ramaraochavali:
nt
    check.go:131: goroutine leak: Leaked goroutine: goroutine 4460 [sleep]:
        time.Sleep(0xee6b2800)
        	/usr/local/go/src/runtime/time.go:193 +0xd2
        istio.io/istio/security/pkg/nodeagent/sds.(*Server).initWorkloadSdsService.func1(0xc000db23a0, 0xc002728998, 0xc0019777d0)
        	/home/prow/go/src/istio.io/istio/security/pkg/nodeagent/sds/server.go:114 +0xc5
        created by istio.io/istio/security/pkg/nodeagent/sds.(*Server).initWorkloadSdsService
        	/home/prow/go/src/istio.io/istio/security/pkg/nodeagent/sds/server.go:92 +0x2dd
--- FAIL: TestAgent (19.33s)

PR #33959, opened 2021-07-11
fix goroutine leak in agent test

ramaraochavali:
Fixes https://github.com/istio/istio/issues/33958
This is not a real issue in prod but fails tests.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/security/pkg/nodeagent/sds/server.go
+++ b/security/pkg/nodeagent/sds/server.go
@@ -39,12 +39,12 @@ type Server struct {
 	grpcWorkloadListener net.Listener
 
 	grpcWorkloadServer *grpc.Server
-	stopped            atomic.Bool
+	stopped            *atomic.Bool
 }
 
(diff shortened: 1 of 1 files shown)


security/pkg/nodeagent/sds/server.go
```diff
@@ -38,11 +39,12 @@ type Server struct {
 	grpcWorkloadListener net.Listener
 
 	grpcWorkloadServer *grpc.Server
+	stopped            atomic.Bool
```
howardjohn, line 42: I think this is supposed to be a pointer created by atomic.NewBool?

ramaraochavali: oh yeah

howardjohn (approved):

PR #33966, opened 2021-07-12
CronJobs already GA in k8s V1.21

ChenglongGao327:
The latest version of Cronjob is batch/v1, which is GA in version 1.21 of k8s
it can be derived from pod metadata

k8s PR：Promote CronJobs to batch/v1
https://github.com/kubernetes/kubernetes/pull/98965

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pkg/kube/util.go
+++ b/pkg/kube/util.go
@@ -235,9 +235,6 @@ func GetDeployMetaFromPod(pod *kubeApiCore.Pod) (metav1.ObjectMeta, metav1.TypeM
 				if _, err := strconv.Atoi(controllerRef.Name[l-10:]); err == nil && string(controllerRef.Name[l-11]) == "-" {
 					deployMeta.Name = controllerRef.Name[:l-11]
 					typeMetadata.Kind = "CronJob"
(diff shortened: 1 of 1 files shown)


ChenglongGao327:
cc @bianpengyuan

howardjohn (commented):
How will this impact users not on 1.21? Most are not yet

bianpengyuan:
> How will this impact users not on 1.21? Most are not yet

+1, Also I think we will need to replace it, not removing it? But anyway, let's wait until istio support window covers 1.21+, which could take a while..

ChenglongGao327:
> > How will this impact users not on 1.21? Most are not yet
> 
> +1, Also I think we will need to replace it, not removing it? But anyway, let's wait until istio support window covers 1.21+ or majority of user is on 1.21+ before making this change, which could take a while..

I just found this, not sure when to modify cronjob to use v1 version. Proxy can be injected in an environment where k8s is less than 1.21

PR #33974, opened 2021-07-12
Bump dependencies

jwendell:


ericvn (approved):
LGTM, but I'm curious if the proxy change will actually pass the current tests.

docker/Dockerfile.distroless
```diff
@@ -8,13 +8,13 @@ FROM ubuntu:focal as ubuntu_source
 COPY --from=distroless_source /etc/ /home/etc
 COPY --from=distroless_source /home/nonroot /home/nonroot
 RUN echo istio-proxy:x:1337: >> /home/etc/group
-RUN echo isito-proxy:x:1337:1337:istio-proxy:/nonexistent:/sbin/nologin >> /home/etc/passwd
```
howardjohn, line 11: this is suspicious.. I wonder how it worked properly before?

howardjohn (approved):

jwendell:
> LGTM, but I'm curious if the proxy change will actually pass the current tests.

You're right. Should we skip proxy changes in this PR and let automator do its job when https://github.com/istio/proxy/pull/3395 is merged?

bianpengyuan:
Looks like there is proxy crashes: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/33974/integ-cni-k8s-tests_istio/1414772690526408704. An upstream fixes were just merged, which fixed a crash in the same function. https://github.com/envoyproxy/envoy/pull/17302

bianpengyuan:
Looks another crashes caused by the same change..: https://github.com/envoyproxy/envoy/pull/16948#issuecomment-879372229

bianpengyuan:
Hmm still does not work. They are reverting https://github.com/envoyproxy/envoy/pull/16948, so let's wait for that.

PR #33975, opened 2021-07-12
Remove proxy via agent option

howardjohn:
This has been stable for a number of releases now

I put a bunch of assignees since I am not sure if there is some usage of this I wasn't aware of

### DIFF
--- a/pilot/cmd/pilot-agent/options/agent.go
+++ b/pilot/cmd/pilot-agent/options/agent.go
@@ -45,17 +45,13 @@ func NewAgentOptions(proxy *model.Proxy, cfg *meshconfig.ProxyConfig) *istioagen
 		Platform:                 platform.Discover(),
 		GRPCBootstrapPath:        grpcBootstrapEnv,
 		DisableEnvoy:             disableEnvoyEnv,
+		ProxyXDSDebugViaAgent:    proxyXDSDebugViaAgent,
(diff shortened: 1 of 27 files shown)


adiprerepa (commented):
so after this, there is no way to connect to envoy directly? are we sure we want to do this?

It seems like a good "backup" option for if istio-agent fails

howardjohn:
> It seems like a good "backup" option for if istio-agent fails

At this I think we are confident in the stability, I hope. We have had it for quite a few releases, and for the most part its not even a good backup at this point. We have no tests and a large amount of functionality depends on the agent

adiprerepa:
> > It seems like a good "backup" option for if istio-agent fails
> 
> At this I think we are confident in the stability, I hope. We have had it for quite a few releases, and for the most part its not even a good backup at this point. We have no tests and a large amount of functionality depends on the agent

yeah, fair enough. I remember it being useful for initial delta stuff, but thats already pretty niche.

hzxuzhonghu (commented):
lgtm

ramaraochavali:
lgtm

lei-tang (commented):
lgtm

howardjohn:
Not stale @istio/wg-networking-maintainers

ramaraochavali (approved):

PR #33978, opened 2021-07-12
Add telemetry.istio.io to validator API groups

Monkeyanator:
Should validate resources for new Telemetry API.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


howardjohn (approved):
We should make the Telemetry validation actually good as well, not sure how robust it is today

Thanks for the fix!

PR #33527, opened 2021-06-18
Simplified external control plane charts

frankbu:
Fixes #33455

**NOTE:** Most of the changes in this PR are reverting what was removed in #33077. The only significant change is that this is also copying the needed `base` charts into `istiod-remote`, so external istiod now doesn't need `base.enabled=true` for a config cluster. IstiodRemote is now a standalone component, containing all the charts needed on a remote (or config) cluster.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #33982, opened 2021-07-12
Make external istiod crd charts templates

frankbu:
This is a followup to #33527. I made an ill-advised last minute change to move the copied crd charts from `templates` to `crds`, without testing :( The problem is that this won't work because I've added a conditional flag (`if configCluster`) to the copied charts, so they needs to be templates.

This fixes it by moving them back to the `templates` directory.


howardjohn (commented):
Wasn't there some Makefile that copies this? does that need to be updated?

frankbu:
@howardjohn, yes, of course, I'm rushing instead of thinking. Thanks.

Monkeyanator (approved):

PR #33983, opened 2021-07-12
chiron: Incorporating kubernetes csr v1 api

shankgan:
Add ability for chiron module of Istiod to use [Kubernetes v1 CSR api ](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/) to sign Istiod DNS certificates and optionally sign workload certificates when Istiod functions as an RA. Refer to [Issue](https://github.com/istio/istio/issues/34012) for context

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[X] Security
[ ] Test and Release
[] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/security/pkg/k8s/chiron/utils.go
+++ b/security/pkg/k8s/chiron/utils.go
@@ -34,61 +34,21 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	clientset "k8s.io/client-go/kubernetes"
 
-	"istio.io/istio/pkg/spiffe"
 	"istio.io/istio/security/pkg/pki/util"
 	"istio.io/pkg/log"
 )
 
 const (
-	maxNameLength       = 63
-	maxDomainNameLength = 6
-	maxNamespaceLength  = 8
-	maxSecretNameLength = 8
-	randomLength        = 18
-	csrRetriesMax       = 3
+	randomLength  = 18
+	csrRetriesMax = 3
 )
 
 type CsrNameGenerator func(string, string) string
 
 // GenCsrName : Generate CSR Name for Resource. Guarantees returning a resource name that doesn't already exist
-func GenCsrName(secretName, namespace string) string {
+func GenCsrName() string {
 	name := fmt.Sprintf("csr-workload-%s", rand.String(randomLength))
 	return name
-}
-
-// getRandomCsrName : returns a random name for CSR.
-func getRandomCsrName(secretName, namespace string) string {
-	domain := spiffe.GetTrustDomain()
-	if len(domain) > maxDomainNameLength {
-		domain = domain[:maxDomainNameLength]
-	}
-	if len(namespace) > maxNamespaceLength {
-		namespace = namespace[:maxNamespaceLength]
-	}
-	if len(secretName) > maxSecretNameLength {
-		secretName = secretName[:maxSecretNameLength]
-	}
-	name := fmt.Sprintf("csr-%s-%s-%s-%s",
-		domain, namespace, secretName, rand.String(randomLength))
-	if len(name) > maxNameLength {
-		name = name[:maxNameLength]
-	}
-	return name
-}
-
-func GetUniqueCsrName(csrCb CsrNameGenerator, client clientset.Interface, secretName, secretNamespace string) (string, error) {
-	var csrName string = ""
-	for retryCount := 0; retryCount < 10; retryCount++ {
-		csrName = csrCb(secretName, secretNamespace)
-		v1CsrReq, v1Beta1CsrReq := checkDuplicateCsr(client, csrName)
-		if v1CsrReq == nil && v1Beta1CsrReq == nil {
-			break
-		}
-	}
-	if csrName == "" {
-		return csrName, fmt.Errorf("exceeded retry attempt to generate unique csrName")
-	}
-	return csrName, nil
 }
 
 // GenKeyCertK8sCA : Generates a key pair and gets public certificate signed by K8s_CA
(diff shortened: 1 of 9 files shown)


howardjohn (commented):
Generally lgtm, I only had a chance to quickly skim it though

hzxuzhonghu:
One question: `kubernetes.io/legacy-unknown` is not allowed in v1 API, what will replace it

shankgan:
> One question: `kubernetes.io/legacy-unknown` is not allowed in v1 API, what will replace it

There is nothing that can replace it as is. Therefore this configuration :- "PILOT_CERT_PROVIDER = Kubernetes" will be obsoleted soon since there is no default signer. Users will have to configure a custom signer.

lei-tang (approved):
Based on offline discussion, this PR is a foundation for follow-up PRs and it does not change the current behaviors. The tests have passed and the PR lgtm.

hzxuzhonghu (changes requested):
Wait, i donot think we can stop supporting for k8s < 1.21 so quickly. 

@howardjohn This has big compatibility influence. 

I would recommend detecting k8s version and automatically choose what to do first.

howardjohn:
I completely agree we cannot drop support. My understanding is this is
detecting the supported API version and falling back to the old one if
needed?

If not then I definitely agree

shankgan:
We aren't dropping support. The v1 API is first invoked. If we get an API not detected error, then we switch to the v1beta1 api. note that all the integration tests that have passed have done so with the v1beta1 api.

security/pkg/k8s/chiron/utils.go
```diff
@@ -215,150 +228,118 @@ func reloadCACert(wc *WebhookController) (bool, error) {
...
+				Request:    csrData,
+				Usages:     usages,
+				SignerName: signerName,
+			},
 		}
-		// If an err other than the CSR exists is returned, re-try
-		if !kerrors.IsAlreadyExists(errRet) {
-			log.Debugf("failed to create CSR (%v): %v", csrName, errRet)
-			continue
+		if len(csr.Spec.Usages) > 0 && len(csr.Spec.SignerName) > 0 && csr.Spec.SignerName != "kubernetes.io/legacy-unknown" {
```
hzxuzhonghu, line 250: why usages need to be checked

security/pkg/k8s/chiron/utils.go
```diff
@@ -70,13 +76,30 @@ func getRandomCsrName(secretName, namespace string) string {
 	return name
 }
 
+func GetUniqueCsrName(csrCb CsrNameGenerator, client clientset.Interface, secretName, secretNamespace string) (string, error) {
```
hzxuzhonghu, line 79: `GetUniqueCsrName` `GenCsrName` `getRandomCsrName` seems very confusing, can you unify them? UUID seems good to largely reduce the conflict. 

Retrying here to detect an existing name is useless, the period between this and when we create a csr, there is a chance a csr with the same name created.

Would recommend corporating generating the csr name with creating csr.

security/pkg/k8s/chiron/utils.go
```diff
@@ -215,150 +228,118 @@ func reloadCACert(wc *WebhookController) (bool, error) {
...
+	csrData []byte, csrName string, signerName string,
+	usages []certv1.KeyUsage, numRetries int) (*certv1.CertificateSigningRequest, *certv1beta1.CertificateSigningRequest, error) {
+	var err error = fmt.Errorf("unable to submit csr %s", csrName)
+
 	for i := 0; i < numRetries; i++ {
 		log.Debugf("trial %v to create CSR (%v)", i, csrName)
-		reqRet, errRet = certClient.Create(context.TODO(), k8sCSR, metav1.CreateOptions{})
-		if errRet == nil && reqRet != nil {
-			break
+		csr := &certv1.CertificateSigningRequest{
```
hzxuzhonghu, line 238: move down to the if clause

security/pkg/k8s/chiron/utils.go
shankgan: Was borrowing the logic from here https://github.com/kubernetes/client-go/blob/master/util/certificate/csr/csr.go. Prefer to use logic already committed (and reviewed) in the kubernetes codebase. Can't use this package directly because it has some missing functionality (for approval/deletion of kubernetes csr requests)

security/pkg/k8s/chiron/utils.go
shankgan: Fixed this

security/pkg/k8s/chiron/utils.go
hzxuzhonghu: agreed

security/pkg/k8s/chiron/utils.go
```diff
@@ -215,150 +186,133 @@ func reloadCACert(wc *WebhookController) (bool, error) {
...
+				},
+			}
+			v1req, err := clientset.CertificatesV1().CertificateSigningRequests().Create(context.TODO(), csr, metav1.CreateOptions{})
+			if err == nil {
+				return csrName, v1req, nil, nil
+			} else if apierrors.IsAlreadyExists(err) {
+				csrName = ""
+				continue
+			} else if apierrors.IsNotFound(err) {
+				// don't attempt to use older api unless we get an API error
```
hzxuzhonghu, line 221: isn't it the opposite?

security/pkg/k8s/chiron/utils.go
```diff
@@ -215,150 +186,133 @@ func reloadCACert(wc *WebhookController) (bool, error) {
...
+
 	for i := 0; i < numRetries; i++ {
-		log.Debugf("trial %v to create CSR (%v)", i, csrName)
-		reqRet, errRet = certClient.Create(context.TODO(), k8sCSR, metav1.CreateOptions{})
-		if errRet == nil && reqRet != nil {
-			break
+		if csrName == "" {
+			csrName = GenCsrName()
+		}
+		if useV1 && len(usages) > 0 && len(signerName) > 0 && signerName != "kubernetes.io/legacy-unknown" {
```
hzxuzhonghu, line 200: nit: can  abstract the below logic to a function to prevent the `continue`, maybe as a follow up

hzxuzhonghu (approved):

PR #33985, opened 2021-07-12
Add policy readiness check in tests

JimmyCYJ:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/integration/security/authorization_test.go
+++ b/tests/integration/security/authorization_test.go
@@ -62,6 +62,7 @@ func TestAuthorization_mTLS(t *testing.T) {
 				policies := tmpl.EvaluateAllOrFail(t, args,
 					file.AsStringOrFail(t, "testdata/authz/v1beta1-mtls.yaml.tmpl"))
 				t.Config().ApplyYAMLOrFail(t, apps.Namespace1.Name(), policies...)
+				util.WaitForConfig(t, policies[0], apps.Namespace1)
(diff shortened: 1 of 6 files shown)


tests/integration/security/authorization_test.go
```diff
@@ -1290,6 +1296,7 @@ func TestAuthorization_Audit(t *testing.T) {
 						}
 						policies := tmpl.EvaluateAllOrFail(t, args, file.AsStringOrFail(t, tc.configFile))
 						t.Config().ApplyYAMLOrFail(t, ns.Name(), policies...)
+						util.WaitForConfig(t, policies[0], ns)
```
howardjohn, line 1299: should we switch WaitForConfig to be `WaitForConfig(t, ns, policies...)` so we can check all?

Probably doesn't matter, but could in some cases

howardjohn (approved):

tests/integration/security/authorization_test.go
JimmyCYJ: Good point! Let me update the PR later.

JimmyCYJ: Done.

bianpengyuan (approved):

Issue #33772, opened 2021-07-01
Add CNI dry run test 


bianpengyuan:
This is part of the CNI beta graduation requirements. CNI runs as cluster singleton and could configure IPTables for multiple Istio versions. A dry run test will be added to CNI in order to make any upgrade breaking changes easier to be identified and reviewed.

Design doc: https://docs.google.com/document/d/1CVRNAlDEsU6Lmssv01gWLt3uaaWJNBVLlIyZeGWh0KA/edit

PR #33992, opened 2021-07-13
Add CNI plugin dryrun test.

bianpengyuan:
fix https://github.com/istio/istio/issues/33772


cni/pkg/plugin/testdata/basic.txt.golden
```diff
@@ -0,0 +1,26 @@
+* nat
```
howardjohn, line 1: Not in scope for this PR, but it would be nice to move the core iptables tests to this format as well; its annoying right now that golden is in code so its hard to see diffs on failure and hard to update

cni/pkg/plugin/plugin_dryrun_test.go
```diff
@@ -0,0 +1,195 @@
...
+				t.Fatalf("CNI cmdAdd failed with error: %v", err)
+			}
+
+			generated, err := ioutil.ReadFile(outputFilePath)
+			if err != nil {
+				log.Fatalf("Cannot read generated IPTables rule file: %v", err)
+			}
+			generatedRules := getRules(generated)
+
+			if os.Getenv("REFRESH_GOLDEN") == "true" {
```
howardjohn, line 141: pilot/test/util/diff.go has some useful functions for golden files, including CompareContent, RefreshGoldenFile, etc that may be preferred. Not sure how well it would work here with the custom logic, but it seems like it could be plausible

howardjohn (approved):

howardjohn:
Put a hold in case you want to use the `diff` package, if not feel free to merge

cni/pkg/plugin/testdata/basic.txt.golden
bianpengyuan: Ack.

cni/pkg/plugin/plugin_dryrun_test.go
bianpengyuan: Thanks for the pointer! I used the RefreshGolden function, Compare function is a bit tricky because IPTables output tables order is not stable so need a bit of custom logic. IPTables output is small and I think we can live with raw got/want output.

howardjohn:
Surprised order is not stable, what part changes?

bianpengyuan:
@howardjohn it is the order of tables. Some times nat table comes first, sometimes raw table comes first. IPtables logic internally has a map for tables, which makes output order of tables not stable.

howardjohn:
Interesting, thanks

PR #33994, opened 2021-07-13
security/pkg/stsservice/tokenmanager/google: Fix the last connection was not closed when retrying

wzshiming:
Fix the last connection was not closed when retrying

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[x] Developer Infrastructure

Pull Request Attributes


howardjohn (commented):
I am a bit confused by the logic. It seems weird to close the body at the start of the loop and then do a defer as well..

I think it probably works, but its not very clear without a lot of studying. Can we simplify?

wzshiming:
I move the Close to the caller to do it, It looks simple now.

wzshiming:
@howardjohn Could you have time to take a look

Issue #33980, opened 2021-07-12
Usage of a deprecated go-control-plane struct


jwendell:
https://github.com/istio/istio/blob/0eb61cf4a6b12ef8918193e12d314f03d464e219/istioctl/pkg/writer/pilot/status.go#L173

`make lint` is complaining about `GetXdsConfig()` being deprecated with the upgrade to `github.com/envoyproxy/go-control-plane v0.9.10-0.20210708144103-3a95f2df6351` in https://github.com/istio/istio/pull/33976

howardjohn:
cc @therealmitchconnors

jwendell:
Fixed by https://github.com/istio/istio/pull/33997

PR #33997, opened 2021-07-13
fix deprecated xds config struct

monsteredc:
This PR is to fix the issue #33980. The old struct is deprecated, and i just revise it to the new one.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/istioctl/pkg/writer/pilot/status.go
+++ b/istioctl/pkg/writer/pilot/status.go
@@ -171,7 +171,7 @@ func (s *XdsStatusWriter) setupStatusPrint(drs map[string]*xdsapi.DiscoveryRespo
 				if err != nil {
 					return nil, nil, fmt.Errorf("could not unmarshal ClientConfig: %w", err)
 				}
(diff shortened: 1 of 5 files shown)


shamsher31 (commented):
@monsteredc Can you please add or update test accordingly?

monsteredc:
@shamsher31 PTAL.

hzxuzhonghu (approved):

howardjohn (commented):
Nice!

My only concern is this seems like its changing `istioctl` and `istiod` code together. does this mean a mismatch of versions between the two would result in this not working?

We may need to make istioctl support both the old and new version?

monsteredc:
@howardjohn Done. Good point, thanks!

howardjohn:
Thanks! I am not an expert on this code so defering to @therealmitchconnors

therealmitchconnors (approved):

PR #33998, opened 2021-07-13
Istio master

AliRezaTaleghani:
[X] Docs
[X] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[X] Test and Release
[X] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Using any **CRI**  except Docker engine required an explicit registry identifier with image path (URI).
- this PR tried to cover as many as images expected to be pulled from docker.io by explicit image path update where ever it was applicable.


craigbox:
/ok-to-test

Thanks for this; it might have a better chance of going through in several smaller PRs, but we'll see what we can do.

manifests/profiles/default.yaml
```diff
@@ -75,7 +75,7 @@ spec:
         autoInject: enabled
         tracer: "zipkin"
       proxy_init:
-        image: proxyv2
+        image: docker.io/proxyv2
```
jwendell, line 78: I don't think changes like this are right.
First, it lacks hub.
Second, we only put the image name to let hub and tag to be given as helm values. Our templates check if there's a hub in the image name, and if there's not, global hub and tag are applied.

AliRezaTaleghani: I see @jwendell 

you are right! I guess I pushed the boundary more that it should...
I went for CRIO as runtimes and found for stuff like **promethous** of **zipkin** and more others there is such a problems that would be nice to be fix. but it may be better to go for more smaller PRs as @craigbox said.

I guess I should not touch profiles manifest anyway...

AliRezaTaleghani: just let me know if you think it's to much to fix this way...
I will re-base my fork to have more smaller MR during the next couple of days then

PR #34000, opened 2021-07-13
update addons manifest with explicit image path

AliRezaTaleghani:
none-docker-engine CRI env expect an explicit image path contains registry as well

[X] Docs
[X] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[X] User Experience
[X] Developer Infrastructure

### DIFF
--- a/samples/addons/extras/prometheus_vm.yaml
+++ b/samples/addons/extras/prometheus_vm.yaml
@@ -452,7 +452,7 @@ spec:
       serviceAccountName: prometheus
       containers:
         - name: prometheus-server-configmap-reload
-          image: "jimmidyson/configmap-reload:v0.5.0"
+          image: "docker.io/jimmidyson/configmap-reload:v0.5.0"
           imagePullPolicy: "IfNotPresent"
           args:
             - --volume-dir=/etc/config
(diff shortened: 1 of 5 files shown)


howardjohn (changes requested):
This is generated from `./manifests/addons/gen.sh`, we need to update the inputs there then re-run the script.

That being said - these are basically generated by `helm template prometheus`, grafana, etc. Given all of these extremely popular charts are not putting docker.io, are we really expected to? It feels a bit like you need to boil the entire ocean to get this working...

howardjohn:
I am not against it, just a bit suspicious that its actually viable to run a kubernetes cluster that doesn't resolve this to docker.io already. Seems like a lot of applications would break

AliRezaTaleghani:
@howardjohn 
I see, if they are all just generation by `helm template` then really make sense to push them via original charts repo.

AliRezaTaleghani:
and that is very true! if you move away from Docker engine into any other CRI (ex: cotainerd / cri-o) then there is no magic auto registry definition under the hood and we need to explicitly say where should it pull from...

howardjohn:
Are you sure? I regularly run kind and GKE with containerd and it resolves
to docker.io. Not an expert here though, just seems to work for me

AliRezaTaleghani:
Regarding **cri-o**, yes as it's my current deployment mode...
Also, I can pretty sure about **containerd** there we have global config with specific directly:
- [[unqualified-search-registries]]
- [[registry]]

but this sounds to be a more broad case:
lets say we have containerd and array of trusted registries (ex: docker.io / quay.io)
sadly image called:  _proxy-xyz/v-1.2.3_ is available on both, but not maintained by the say team/co...

PR #34018, opened 2021-07-13
Align config cluster default remote-secret sa with istiod-remote install

frankbu:
Changes `istioctl x create-remote-secret --config` to use the `istiod` service-account by default, since `istiod-service-account` is now legacy for backward compat, and not even installed on an external istiod config cluster.

### DIFF
--- a/istioctl/pkg/multicluster/remote_secret.go
+++ b/istioctl/pkg/multicluster/remote_secret.go
@@ -334,7 +334,6 @@ func generateServiceAccountYAML(opt RemoteSecretOptions) (string, error) {
 	values := fmt.Sprintf(`
 global:
   istioNamespace: %s
-  externalIstiod: true
 `, opt.Namespace)
 
 	// Render the templates required for the service account and role bindings.
--- a/istioctl/pkg/multicluster/remote_secret.go
+++ b/istioctl/pkg/multicluster/remote_secret.go
@@ -334,6 +334,7 @@ func generateServiceAccountYAML(opt RemoteSecretOptions) (string, error) {
 	values := fmt.Sprintf(`
 global:
   istioNamespace: %s
+  externalIstiod: true
 `, opt.Namespace)
 
 	// Render the templates required for the service account and role bindings.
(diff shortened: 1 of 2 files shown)


pkg/config/constants/constants.go
```diff
@@ -93,8 +93,8 @@ const (
 	// DefaultServiceAccountName is the default service account to use for remote cluster access.
 	DefaultServiceAccountName = "istio-reader-service-account"
 
-	// DefaultConfigServiceAccountName is the default service account to use for external Istiod cluster access.
-	DefaultConfigServiceAccountName = "istiod-service-account"
+	// DefaultConfigServiceAccountName is the default service account to use for external Istiod config cluster access.
+	DefaultConfigServiceAccountName = "istiod"
```
Monkeyanator, line 97: isn't this service account name revisioned?

frankbu: It appends the revision if there is one:
```
name: istiod{{- if not (eq .Values.revision "") }}-{{ .Values.revision }}{{- end }}
```
But we install the external istiod without a revision, so it's just `istiod` in the initial setup. Is that not the intended name?

istioctl/pkg/multicluster/remote_secret.go
```diff
@@ -334,6 +334,7 @@ func generateServiceAccountYAML(opt RemoteSecretOptions) (string, error) {
 	values := fmt.Sprintf(`
 global:
   istioNamespace: %s
+  externalIstiod: true
```
stevenctl, line 337: Should this really be default for all flavors of multi-cluster? That doesn't seem right.

frankbu: Actually, I added this by accident in this PR. That said, it's a good question whether it should be added or not. In the external control plane doc, we set `--create-service-account=false` and instead install the service-account first, before we run `create-remote-secret`, which I think is the better way. So whether or not we set it here doesn't matter for that case.

In the general multicluster remote cluster case, however, I'm not sure that it shouldn't be set also. All it does is adds some additional privileges for istiod, esp. the configmap update permission which is needed to allow the istiod to add the ca-root-certificate cm. https://github.com/istio/istio/blob/2d99cd2817b2a939c85806020bd0acc29b261d5e/manifests/charts/istio-control/istio-discovery/templates/reader-clusterrole.yaml#L38

I think this will be needed when the multicluster instructions are changed to create a truly istiodless remote cluster. Currently, a local istiod is being used for the CA.

WDYT, should I remove it?

stevenctl: If we want it on for the general case, why have it gated by this value?

frankbu: I think it's not needed for primary-remote MC today, because it's running a local istiod CA on the remote cluster. Once that's changed to a proper istiodless remote (i.e., use the new "external" profile, instead of "remote"), I think the remote istiod may need it. But then it may not be needed at all if certs are configured separately, then istiod won't install the ca-root-cert configmap, right?

Anybody else have more insight on why this is the way it is? The chart has been this way for a quite a while, but then multicluster wasn't working properly with external control plane until recently.

frankbu: I removed the `extenralIstiod: true` setting, since changes to the flag and/or chart is outside the scope of this PR. This PR is only intended to align the default generated sa name with the default istio-remote install.

Monkeyanator (approved):

nmittler (approved):

PR #34025, opened 2021-07-14
[FIX]default mesh config file path

scaat:
The current path is `./etc/istio/config/mesh`, use absolute path instead, keep it consistent with `networksConfig`.

[ X ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

[ X ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

### DIFF
--- a/pilot/cmd/pilot-discovery/app/cmd.go
+++ b/pilot/cmd/pilot-discovery/app/cmd.go
@@ -127,7 +127,7 @@ func addFlags(c *cobra.Command) {
 		serverArgs.RegistryOptions.ClusterRegistriesNamespace, "Namespace for ConfigMap which stores clusters configs")
 	c.PersistentFlags().StringVar(&serverArgs.RegistryOptions.KubeConfig, "kubeconfig", "",
 		"Use a Kubernetes configuration file instead of in-cluster configuration")
-	c.PersistentFlags().StringVar(&serverArgs.MeshConfigFile, "meshConfig", "./etc/istio/config/mesh",
+	c.PersistentFlags().StringVar(&serverArgs.MeshConfigFile, "meshConfig", "/etc/istio/config/mesh",
 		"File name for Istio mesh configuration. If not specified, a default mesh will be used.")
 	c.PersistentFlags().StringVar(&serverArgs.NetworksConfigFile, "networksConfig", "/etc/istio/config/meshNetworks",
 		"File name for Istio mesh networks configuration. If not specified, a default mesh networks will be used.")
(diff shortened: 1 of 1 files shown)


howardjohn:
This is intended to allow running without access to the root filesystem. In the standard docker images we run with WD=/ so it's the same in most cases

scaat:
@howardjohn I see. Do I need to close this pr?

pilot/cmd/pilot-discovery/app/cmd.go
```diff
@@ -127,7 +127,7 @@ func addFlags(c *cobra.Command) {
 		serverArgs.RegistryOptions.ClusterRegistriesNamespace, "Namespace for ConfigMap which stores clusters configs")
 	c.PersistentFlags().StringVar(&serverArgs.RegistryOptions.KubeConfig, "kubeconfig", "",
 		"Use a Kubernetes configuration file instead of in-cluster configuration")
-	c.PersistentFlags().StringVar(&serverArgs.MeshConfigFile, "meshConfig", "./etc/istio/config/mesh",
+	c.PersistentFlags().StringVar(&serverArgs.MeshConfigFile, "meshConfig", "/etc/istio/config/mesh",
```
ramaraochavali, line 130: I think @costinm added it intentionally. Any issue why you want to change?

scaat: It may be easy to mislead.
Or adjust the default configuration path of networksConfig to a relative path `./etc/istio/config/meshNetworks`?

hzxuzhonghu: looks reasonable

howardjohn: As mentioned in the other comment: "This is intended to allow running without access to the root filesystem. In the standard docker images we run with WD=/ so it's the same in most cases"

We can make the mesh networks one a relative path too if we want, but shouldn't change the meshconfig path.

scaat: ok, adjust the default configuration path of networksConfig to a relative path `./etc/istio/config/meshNetworks`.

howardjohn (approved):

PR #34026, opened 2021-07-14
fix delete secret bug: delete secretKey cluster only once

gy95:
only after fininsh range `remoteClusters` using `secretKey`, we could delete from  `remoteClusters` map using key `secretKey`, and delete only once
first delete using secretKey and clusterID
finally delete all using secretKey

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/kube/secretcontroller/secretcontroller.go
+++ b/pkg/kube/secretcontroller/secretcontroller.go
@@ -486,8 +486,9 @@ func (c *Controller) deleteSecret(secretKey string) {
 				clusterID, secretKey, err)
 		}
 		close(cluster.Stop)
-		delete(c.cs.remoteClusters, secretKey)
+		delete(c.cs.remoteClusters[secretKey], clusterID)
 	}
+	delete(c.cs.remoteClusters, secretKey)
 }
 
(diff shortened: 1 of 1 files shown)


pkg/kube/secretcontroller/secretcontroller.go
```diff
@@ -486,8 +486,9 @@ func (c *Controller) deleteSecret(secretKey string) {
 				clusterID, secretKey, err)
 		}
 		close(cluster.Stop)
-		delete(c.cs.remoteClusters, secretKey)
+		delete(c.cs.remoteClusters[secretKey], clusterID)
```
hzxuzhonghu, line 489: Maybe donot need this at all, not sure which is best for go gc

hzxuzhonghu (approved):
/ok-to-test

pkg/kube/secretcontroller/secretcontroller.go
gy95: yes, maybe just need delete from map using `secretKey` finally after range

PR #34028, opened 2021-07-14
Update helm integration tests to use recommended approach when using helm upgrades

jacob-delgado:
Create integration test to use stable revision labels with Helm

Do the following:
1. install helm with 1.10.0 with --set revision=1-10-0 --set
   revisionTags={prod}
2. Install client/server app and communicate between the two of them;
   this will use proxyv2 from 1.10.0
3. install helm with master branch with --set revision=latest --set
   revisionTags={canary}
4. Install client/server app and communicate between the two of them
   (both using latest version)
5. Verify that communication can occur between 1.10.0 and latest with
   echo app
6. Change prod tag to point to the latest revision
7. Change the namespace used for 1-10-0 revision to point to 'latest'
8. Restart the pods in that old namespace. They should now get the
   latest proxyv2 image on restart.
9. Verify that communication between these restarted pods can occur

[ ] Docs
[X] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[X] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


jacob-delgado:
cc @Monkeyanator I worked with @nrjpoddar last week to reverse engineer how stable label revisions work. We were able to get it working manually and would like for this to get in so we can add it to the helm documentation.

We also believe that this would help both Helm stabilization to Beta and the Revisions work you've done.

pkg/test/kube/util.go
```diff
@@ -253,3 +254,24 @@ func newRetryOptions(opts ...retry.Option) []retry.Option {
 	out = append(out, opts...)
 	return out
 }
+
+// MutatingWebhookConfigurationsExists returns true if all of the given mutating webhook configs exist.
+func MutatingWebhookConfigurationsExists(a kubernetes.Interface, names []string) bool {
```
shamsher31, line 259: nit:
```suggestion
func IsMutatingWebhookConfigurationsExists(a kubernetes.Interface, names []string) bool {
```

pkg/test/kube/util.go
```diff
@@ -253,3 +254,24 @@ func newRetryOptions(opts ...retry.Option) []retry.Option {
...
+
+	sort.Strings(names)
+	for _, cfg := range cfgs.Items {
+		if idx := sort.SearchStrings(names, cfg.Name); idx == len(names) {
+			return false
+		}
+	}
+
+	return true
+}
```
shamsher31, line 277: new line.
```suggestion
}

```

tests/integration/helm/upgrade/util.go
```diff
@@ -227,14 +226,94 @@ func performRevisionUpgradeFunc(previousVersion string) func(framework.TestConte
...
+		//t.ConditionalCleanup(func() {
+		//	err := deleteIstioRevision(h, revisionChartSuffix)
+		//	if err != nil {
+		//		t.Fatalf("could not delete istio: %v", err)
+		//	}
+		//	err = deleteIstio(cs, h)
+		//	if err != nil {
+		//		t.Fatalf("could not delete istio: %v", err)
+		//	}
+		//})
```
shamsher31, line 257:
```suggestion
```

tests/integration/helm/util.go
```diff
@@ -156,20 +166,47 @@ func deleteIstio(t framework.TestContext, h *helm.Helm, cs *kube.Cluster) {
...
+			if _, err := kubetest.CheckPodsAreReady(kubetest.NewPodFetch(cs, IstioNamespace, "app=istio-egressgateway")); err != nil {
+				return fmt.Errorf("istio egress gateway pod is not ready: %v", err)
+			}
 		}
 		return nil
 	}, retry.Timeout(RetryTimeOut), retry.Delay(RetryDelay))
 	scopes.Framework.Infof("=== succeeded ===")
 }
+
+func SetRevisionTag(ctx framework.TestContext, h *helm.Helm, fileSuffix, revision, revisionTag string) {
```
Monkeyanator, line 190: So will we be documenting to template from the revision tag chart and the `kubectl apply` to Helm users? I think when switching tags this is the only way that works(?) since otherwise there'd be chart overlap conflicts

Maybe in the future we have a dedicated chart for tags or just put it all in base so users could `helm upgrade` instead.

tests/integration/helm/upgrade/util.go
```diff
@@ -227,14 +226,94 @@ func performRevisionUpgradeFunc(previousVersion string) func(framework.TestConte
...
+
+		// install 1.10.0 charts with revision set to "1-10-0"
+		previousRevision := strings.ReplaceAll(previousVersion, ".", "-")
+		overrideValuesFile := getValuesOverrides(t, gcrHub, previousVersion, previousRevision)
+		helmtest.InstallIstioWithRevision(t, cs, h, tarGzSuffix, "-"+previousRevision, overrideValuesFile, true)
```
Monkeyanator, line 262: why `"-"+previousRevision` here?

pkg/test/kube/util.go
```diff
@@ -253,3 +254,24 @@ func newRetryOptions(opts ...retry.Option) []retry.Option {
 	out = append(out, opts...)
 	return out
 }
+
+// MutatingWebhookConfigurationsExists returns true if all of the given mutating webhook configs exist.
+func IsMutatingWebhookConfigurationsExists(a kubernetes.Interface, names []string) bool {
```
Monkeyanator, line 259:
```suggestion
func MutatingWebhookConfigurationsExists(a kubernetes.Interface, names []string) bool {
```

tests/integration/helm/upgrade/util.go
```diff
@@ -227,14 +226,94 @@ func performRevisionUpgradeFunc(previousVersion string) func(framework.TestConte
...
+		_, newClient, newServer := sanitycheck.SetupTrafficTest(t, t, latestRevisionTag)
+		sanitycheck.RunTrafficTestClientServer(t, newClient, newServer)
+
+		// now check that we are compatible with N-1 proxy with N proxy
+		sanitycheck.RunTrafficTestClientServer(t, oldClient, newServer)
+
+		// set the revision tag for prod to point to the latest istiod
+		helmtest.SetRevisionTag(t, h, tarGzSuffix, latestRevisionTag, prodTag)
+
+		err = oldNs.RemoveLabel(istioRevLabel)
```
Monkeyanator, line 296: nit: should be able to just `SetLabel` without having to remove first?

tests/integration/helm/upgrade/util.go
```diff
@@ -227,14 +226,94 @@ func performRevisionUpgradeFunc(previousVersion string) func(framework.TestConte
...
+		if err != nil {
+			t.Fatal("could not remove istio.io/rev from old namespace")
+		}
+
+		err = oldNs.SetLabel(istioRevLabel, latestRevisionTag)
```
Monkeyanator, line 301: don't we want to be setting `istio.io/rev=prod` here to test the revision tag upgrade?

tests/integration/helm/upgrade/util.go
```diff
@@ -227,14 +226,94 @@ func performRevisionUpgradeFunc(previousVersion string) func(framework.TestConte
...
+		err = oldClient.Restart()
+		if err != nil {
+			t.Fatal("could not restart old client")
+		}
+		err = oldServer.Restart()
+		if err != nil {
+			t.Fatal("could not restart old server")
+		}
+
+		// now check traffic still works
```
Monkeyanator, line 315: should we check the actual container image somewhere to make sure it updated as part of the revision tag upgrade?

tests/integration/helm/upgrade/util.go
jacob-delgado: I removed this, but the intent is to install with `istiod-REVISION`. It just appends the revision for the helm chart. I modified the code to defer this until it needs to be done.

tests/integration/helm/upgrade/util.go
jacob-delgado: fixed

tests/integration/helm/upgrade/util.go
jacob-delgado: Good catch. Yes, this was "working" but is not the intended behavior we wish to test.

tests/integration/helm/util.go
jacob-delgado: For now yes. Users are also able to specify using `istioctl`, but for those that cannot this is a current workaround. When you template it only does it for a specific template that you provide. This way its declarative, but isn't kept to a helm chart.

We'd have to think through how users should use helm install/upgrade, but @nrjpoddar and I were ok with this approach. We can always re-evaluate it after we get feedback from users. For now, this is about creating parity for Helm users with `istioctl` users.

tests/integration/helm/upgrade/util.go
jacob-delgado: I needed this code. Fixed

tests/integration/helm/upgrade/util.go
jacob-delgado: Added code to verify the image does not contain the previous tag "1.10.0"

tests/integration/helm/util.go
nrjpoddar: While this is a workaround to get parity, if everyone is ok with restructuring Helm chart or adding/moving templates around I’m up for that. @Monkeyanator i don’t quite get how installing a new mutating webhook can be achieved though.

Only way I can think of is having 2 copies of mutating webhook where the second one is guarded by a Helm value “defaultRevision” being set? We can do that now and then just the Helm upgrade workflow would work.

Monkeyanator (approved):

stevenctl (approved):

tests/integration/helm/util.go
jacob-delgado: After talking about this with @Monkeyanator we agreed to write up an RFC for addressing the chart structure. The current base chart has deprecated items and in https://github.com/istio/istio/pull/33773 the webhooks are being moved from discovery to base.

Being consistent would be helpful for devs and users to get the experience they desire. Helm will probably go to Beta in 1.11, but without a clear plan it'll be harder to reach the Stable stage.

Issue #31638, opened 2021-03-23
Delta xDS tracking


adiprerepa:
An issue to track the progress of delta xDS within istio.

[Design doc](https://docs.google.com/document/d/1hwC81_jS8qARBDcDE6VTxx6fA31In96xAZWqfwnKhpQ/edit#)

cc @howardjohn 

- [x] Initial implementation (https://github.com/istio/istio/pull/31212)
- [x] start with deltas for CDS (least CRDS impact) (https://github.com/istio/istio/pull/34029)
- [x] integration tests
- [ ] enable by default
- [ ] Possibly explore on-demand?
- [x] Get current integration tests to work when delta is enabled (https://github.com/istio/istio/pull/31832)
- [x] https://github.com/istio/istio/issues/32565

howardjohn:
I would add "get all existing integration tests to pass when enabled". That is probably the most important, as it shows that we have feature parity with SotW

adiprerepa:
Not stale

linsun:
reopen

howardjohn:
Delta xDS is on by default in 1.22 :partying_face:

PR #34029, opened 2021-07-14
delta: initial implementation for CDS

adiprerepa:
delta for CDS -- very minimal initial implementation
This only generates delta clusters whenever *just* services change, and for outbound clusters only.

this PR is mainly to provide a mechanism to implement deltas as well as provide an initial implementation.

Issue: https://github.com/istio/istio/issues/31638

cc @howardjohn 

added some assignees for people who might be interested


pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -71,35 +75,76 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
...
+	delta := len(configKeys) == 1 && configKeys[0].Kind == gvk.Service && attemptDeltas
+	if delta {
+		// use delta -- only services changed
+		configs := model.ConfigsOfKind(updates.ConfigsUpdated, gvk.Service)
+		for s := range configs {
+			// get service that changed
+			service := push.ServiceForHostname(proxy, host.Name(nameFromFQDN(s.Name)))
+			// is the service visible to envoy?
+			// does envoy care about this service changing? (watchedResources)
+			if push.IsServiceVisible(service, proxy.ConfigNamespace) {
```
howardjohn, line 96: I don't think this fully takes
```
	if features.FilterGatewayClusterConfig && cb.proxy.Type == model.Router {
		services = cb.push.GatewayServices(cb.proxy)
	} else {
		services = cb.push.Services(cb.proxy)
	}
```

logic into account.

For gateways, its slightly different. For sidecar, `push.IsServiceVisible` only accounts for if its _exported_, not if its imported. I think we need to look at `proxy.SidecarScope.serviceByHostname`

pilot/pkg/model/context.go
```diff
@@ -208,7 +208,7 @@ var DefaultXdsLogDetails XdsLogDetails = XdsLogDetails{}
 // Note: any errors returned will completely close the XDS stream. Use with caution; typically and empty
 // or no response is preferred.
 type XdsResourceGenerator interface {
-	Generate(proxy *Proxy, push *PushContext, w *WatchedResource, updates *PushRequest) (Resources, XdsLogDetails, error)
+	Generate(proxy *Proxy, push *PushContext, w *WatchedResource, updates *PushRequest, delta bool) (Resources, XdsLogDetails, error)
```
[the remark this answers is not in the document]
adiprerepa: yeah, thats how I did it originally, but with a separate `DeltaCdsGenerator`. Will do.

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -71,35 +75,76 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
...
+func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *model.PushContext, attemptDeltas bool, updates *model.PushRequest, watched *model.WatchedResource) ([]*discovery.Resource, model.XdsLogDetails) {
 	clusters := make([]*cluster.Cluster, 0)
 	resources := model.Resources{}
+	services := make([]*model.Service, 0)
 	envoyFilterPatches := push.EnvoyFilters(proxy)
 	cb := NewClusterBuilder(proxy, push, configgen.Cache)
 	instances := proxy.ServiceInstances
-	cacheStats := cacheStats{}
+	configKeys := configMapKeys(updates.ConfigsUpdated)
+	delta := len(configKeys) == 1 && configKeys[0].Kind == gvk.Service && attemptDeltas
```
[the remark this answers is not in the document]
adiprerepa: ah, so I need to filter by `gvk`

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -71,35 +75,76 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
...
-		ob, cs := configgen.buildOutboundClusters(cb, outboundPatcher)
-		cacheStats = cacheStats.merge(cs)
-		resources = append(resources, ob...)
+		var ob []*discovery.Resource
+		var cs cacheStats
+		if delta {
+			ob, cs = configgen.buildOutboundClustersWithServices(cb, outboundPatcher, services)
+			// only add to list when we know the proxy is subscribed to this resource
+			for _, resource := range ob {
+				if contains(watched.ResourceNames, resource.Name) {
```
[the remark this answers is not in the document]
adiprerepa: ah, so we shouldn't care about `watchedResources` in this context?

pilot/pkg/model/context.go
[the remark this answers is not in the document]
adiprerepa: > Then in findGenerator we can have logic to lookup delta/typeUrl first for delta, and fallback to typeUrl if not present

Wouldn't this mean that delta would be enabled by default? if it first looks for the delta generator
Maybe we should pass in a delta flag to `findGenerator`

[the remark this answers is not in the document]
howardjohn: > Maybe we should pass in a delta flag to findGenerator

Yep, that is what I meant. Lookup delta/typeUrl only for delta

pilot/pkg/model/context.go
```diff
@@ -784,6 +784,11 @@ func (node *Proxy) SetSidecarScope(ps *PushContext) {
 	node.PrevSidecarScope = sidecarScope
 }
 
+func (node *Proxy) ServiceInSidecarScope(service *Service) bool {
+	_, exists := node.SidecarScope.servicesByHostname[service.Hostname]
```
howardjohn, line 788: nit: I would make this a function of sidecar scope

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -30,14 +30,18 @@ import (
 
 	meshconfig "istio.io/api/mesh/v1alpha1"
 	networking "istio.io/api/networking/v1alpha3"
+
+	// galleyutil "istio.io/istio/galley/pkg/config/analysis/analyzers/util"
```
howardjohn, line 34: very nit: remove

pilot/pkg/networking/apigen/apigen.go
```diff
@@ -62,8 +62,7 @@ func NewGenerator(store model.IstioConfigStore) *APIGenerator {
 // This provides similar functionality with MCP and :8080/debug/configz.
 //
 // Names are based on the current resource naming in istiod stores.
-func (g *APIGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *model.WatchedResource,
-	updates *model.PushRequest) (model.Resources, model.XdsLogDetails, error) {
+func (g *APIGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *model.WatchedResource, updates *model.PushRequest) (model.Resources, model.XdsLogDetails, error) {
```
howardjohn, line 65: very nit: remove unneeded changes

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -113,10 +144,27 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
...
 
-	if cacheStats.empty() {
+	if clusterCacheStats.empty() {
 		return resources, model.DefaultXdsLogDetails
 	}
-	return resources, model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", cacheStats.hits, cacheStats.hits+cacheStats.miss)}
+	return resources, model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", clusterCacheStats.hits, clusterCacheStats.hits+clusterCacheStats.miss)}
+}
+
+func allConfigKeysOfType(keys []model.ConfigKey, cfg config.GroupVersionKind) bool {
```
howardjohn, line 153:
```suggestion
func allConfigKeysAreSameType(keys []model.ConfigKey, cfg config.GroupVersionKind) bool {
```

currently sounds like its filtering the config keys to only the `cfg` type

pilot/pkg/model/context.go
howardjohn: And I would make this take in a hostname. Then the caller doesn't need to call push.ServiceForHostname(proxy, host.Name(s.Name))?

Actually it should probably return the service, not a bool. It may not return the same thing as ServiceForHostname, and it makes the usage easier for the caller

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -71,20 +75,44 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
...
 	case model.SidecarProxy:
 		// Setup outbound clusters
 		outboundPatcher := clusterPatcher{efw: envoyFilterPatches, pctx: networking.EnvoyFilter_SIDECAR_OUTBOUND}
-		ob, cs := configgen.buildOutboundClusters(cb, outboundPatcher)
-		cacheStats = cacheStats.merge(cs)
+		var ob []*discovery.Resource
+		var cs cacheStats
+		if delta {
+			ob, cs = configgen.buildOutboundClustersWithServices(cb, outboundPatcher, services)
+			// only add to list when we know the proxy is subscribed to this resource
```
howardjohn, line 110: nit: stale comment

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -71,20 +75,44 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
 // For outbound: Cluster for each service/subset hostname or cidr with SNI set to service hostname
 // Cluster type based on resolution
 // For inbound (sidecar only): Cluster for each inbound endpoint port and for each service port
-func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *model.PushContext) ([]*discovery.Resource, model.XdsLogDetails) {
+// if attemptDeltas is set to true, then BuildClusters will attempt to build delta clusters for that proxy
+func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *model.PushContext, attemptDeltas bool, updates *model.PushRequest, watched *model.WatchedResource) ([]*discovery.Resource, model.XdsLogDetails) {
```
howardjohn, line 79: `watched` is not used.

Do we need `attemptDeltas` and `updates`? Maybe we can just make non-nil `updates` meant to use delta

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -71,20 +75,44 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
...
 	switch proxy.Type {
 	case model.SidecarProxy:
 		// Setup outbound clusters
 		outboundPatcher := clusterPatcher{efw: envoyFilterPatches, pctx: networking.EnvoyFilter_SIDECAR_OUTBOUND}
-		ob, cs := configgen.buildOutboundClusters(cb, outboundPatcher)
-		cacheStats = cacheStats.merge(cs)
+		var ob []*discovery.Resource
+		var cs cacheStats
+		if delta {
+			ob, cs = configgen.buildOutboundClustersWithServices(cb, outboundPatcher, services)
```
howardjohn, line 109: I wonder if this code would be a bit simpler if we just always use a single buildOutboundClusters call, but allow passing nil `services` to generate all?

Although need to be careful to not accidentally send all if the delta logic returns no services, so maybe too tricky

pilot/pkg/xds/sds_test.go
```diff
@@ -364,16 +363,14 @@ func TestCaching(t *testing.T) {
...
+	secrets, _, _ := gen.Generate(s.SetupProxy(istiosystem), s.PushContext(), &model.WatchedResource{ResourceNames: []string{"kubernetes://generic"}}, fullPush)
 	raw := xdstest.ExtractTLSSecrets(t, model.ResourcesToAny(secrets))
 	if len(raw) != 1 {
 		t.Fatalf("failed to get expected secrets for authorized proxy: %v", raw)
 	}
 
 	// We should not get secret returned, even though we are asking for the same one
-	secrets, _, _ = gen.Generate(s.SetupProxy(otherNamespace), s.PushContext(),
-		&model.WatchedResource{ResourceNames: []string{"kubernetes://generic"}}, fullPush)
+	secrets, _, _ = gen.Generate(s.SetupProxy(otherNamespace), s.PushContext(), &model.WatchedResource{ResourceNames: []string{"kubernetes://generic"}}, fullPush)
```
howardjohn, line 373: can you revert these changes

howardjohn (commented):
Looks good

cc @hzxuzhonghu @ramaraochavali

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -71,20 +75,44 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
...
 	instances := proxy.ServiceInstances
-	cacheStats := cacheStats{}
+	// delta when only services are modified and we attempt deltas
+	delta := updates != nil && allConfigKeysOfType(configMapKeys(updates.ConfigsUpdated), gvk.Service) && attemptDeltas
+	if delta {
+		// use delta -- only services changed
+		configs := model.ConfigsOfKind(updates.ConfigsUpdated, gvk.Service)
+		for s := range configs {
+			// get service that changed
+			service := push.ServiceForHostname(proxy, host.Name(s.Name))
```
hzxuzhonghu, line 93: This call is redudant

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -71,20 +75,44 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
...
+func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *model.PushContext, attemptDeltas bool, updates *model.PushRequest, watched *model.WatchedResource) ([]*discovery.Resource, model.XdsLogDetails) {
 	clusters := make([]*cluster.Cluster, 0)
 	resources := model.Resources{}
+	services := make([]*model.Service, 0)
 	envoyFilterPatches := push.EnvoyFilters(proxy)
 	cb := NewClusterBuilder(proxy, push, configgen.Cache)
 	instances := proxy.ServiceInstances
-	cacheStats := cacheStats{}
+	// delta when only services are modified and we attempt deltas
+	delta := updates != nil && allConfigKeysOfType(configMapKeys(updates.ConfigsUpdated), gvk.Service) && attemptDeltas
```
hzxuzhonghu, line 87: I think you can iterate the map within one function, no need to declare two functions

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -152,14 +200,18 @@ func buildClusterKey(service *model.Service, port *model.Port, cb *ClusterBuilde
...
 	var services []*model.Service
 	if features.FilterGatewayClusterConfig && cb.proxy.Type == model.Router {
 		services = cb.push.GatewayServices(cb.proxy)
 	} else {
 		services = cb.push.Services(cb.proxy)
 	}
+	return configgen.buildOutboundClustersWithServices(cb, cp, services)
+}
+
+func (configgen *ConfigGeneratorImpl) buildOutboundClustersWithServices(cb *ClusterBuilder, cp clusterPatcher, services []*model.Service) ([]*discovery.Resource, cacheStats) {
```
hzxuzhonghu, line 212: unrelated change

hzxuzhonghu, line 212: Where do you handle the deleted service?

hzxuzhonghu: I see it in delta.go, but with this, i donot think it is right

hzxuzhonghu (commented):
After a look, I am not sure if this is the right direction and if it is general. This model is very simple, and only covers services change, and invalidated with combined changes.

I think our ultimate target is to handle all configs and services changes, so do they conflict?

adiprerepa:
> After a look, I am not sure if this is the right direction and if it is general. This model is very simple, and only covers services change, and invalidated with combined changes.
> 
> I think our ultimate target is to handle all configs and services changes, so do they conflict?

I think delta is a very incremental optimization. So, it is not necessary to have **everything** done at once. I do not believe that starting with this implementation will conflict with future optimizations (@howardjohn correct me if im wrong).  I think it is good to start simple, and optimize later on.

pilot/pkg/model/context.go
```diff
@@ -784,6 +784,11 @@ func (node *Proxy) SetSidecarScope(ps *PushContext) {
 	node.PrevSidecarScope = sidecarScope
 }
 
+func (node *Proxy) ServiceInSidecarScope(service *Service) bool {
```
ramaraochavali, line 787: Moving this to sidecar makes sense

pilot/pkg/xds/xdsgen.go
```diff
@@ -60,11 +60,15 @@ func init() {
 	controlPlane = &corev3.ControlPlane{Identifier: string(byVersion)}
 }
 
-func (s *DiscoveryServer) findGenerator(typeURL string, con *Connection) model.XdsResourceGenerator {
+func (s *DiscoveryServer) findGenerator(typeURL string, con *Connection, delta bool) model.XdsResourceGenerator {
 	if g, f := s.Generators[con.proxy.Metadata.Generator+"/"+typeURL]; f {
 		return g
 	}
 
+	if g, f := s.Generators["delta/"+typeURL]; f && delta {
```
ramaraochavali, line 68: nit: Check for delta/typeURL in generators only if it is delta

pilot/pkg/xds/xdsgen.go
```diff
@@ -60,11 +60,15 @@ func init() {
 	controlPlane = &corev3.ControlPlane{Identifier: string(byVersion)}
 }
 
-func (s *DiscoveryServer) findGenerator(typeURL string, con *Connection) model.XdsResourceGenerator {
+func (s *DiscoveryServer) findGenerator(typeURL string, con *Connection, delta bool) model.XdsResourceGenerator {
```
ramaraochavali, line 63: It does not right changing signature of this method. The better way would be to append "delta" to typeURl when it is invoked from delta.go

pilot/pkg/networking/core/configgen.go
```diff
@@ -35,7 +35,7 @@ type ConfigGenerator interface {
 	BuildListeners(node *model.Proxy, push *model.PushContext) []*listener.Listener
 
 	// BuildClusters returns the list of clusters for the given proxy. This is the CDS output
-	BuildClusters(node *model.Proxy, push *model.PushContext) ([]*discovery.Resource, model.XdsLogDetails)
```
ramaraochavali, line 38: Having to pass boolean to detect delta or not to BuildClusters does not seem right in the interface. Can we add additional method for delta like BuildDeltaClusters that takes additional updates and cdsGenerator can call that when it is delta mode?

ramaraochavali: In the implementation we can reuse with an internal method.

ramaraochavali (changes requested):
Added few interface comments. Let us first align on that. Then I can take another look.

pilot/pkg/xds/xdsgen.go
adiprerepa: well, it only returns the generator if its delta, are you saying we shouldn't even attempt access?

pilot/pkg/xds/xdsgen.go
adiprerepa: yeah, we could do that

pilot/pkg/networking/core/configgen.go
adiprerepa: I think adding a delta flag would involve less changes -- and is just simpler overall. `BuildDeltaClusters` would require no-op implementations

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: how so?

pilot/pkg/networking/core/configgen.go
ramaraochavali: I think at the interface level I feel it is better to have a separate function for BuildDeltaClusters or a generic optionsConfig struct (that defaults for normal cases) instead of forcing every caller to pass some value for delta. Some API design considerations https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

But I would like to hear from @hzxuzhonghu @howardjohn  also.

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: ping on this question

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -119,6 +124,103 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
...
+			services = append(services, service)
+		}
+		// we also need to update proxies of deleted config for delta
+		if updates.UpdateTypes[s] == model.EventDelete {
+			deletedServices = append(deletedServices, service)
+		}
+	}
+	clusterCacheStats := cacheStats{}
+	switch proxy.Type {
+	case model.SidecarProxy:
```
adiprerepa, line 165: ill have an internal method that has most of this logic to avoid duplicates between `BuildClusters` and `BuildDeltaClusters`

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: i think i solved it

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -71,6 +77,7 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
 // For outbound: Cluster for each service/subset hostname or cidr with SNI set to service hostname
 // Cluster type based on resolution
 // For inbound (sidecar only): Cluster for each inbound endpoint port and for each service port
+// if attemptDeltas is set to true, then BuildClusters will attempt to build delta clusters for that proxy
```
howardjohn, line 80: stale comment

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -119,6 +126,106 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
...
+	deletedServices := make([]*model.Service, 0)
+	envoyFilterPatches := push.EnvoyFilters(proxy)
+	cb := NewClusterBuilder(proxy, push, configgen.Cache)
+	instances := proxy.ServiceInstances
+	// delta when only services are modified and we attempt deltas
+	// we can progressively optimize this
+	log.Infof("adiprerepa: update types: %v", updates.UpdateTypes)
+	delta := updates != nil && allConfigKeysOfType(updates.ConfigsUpdated, gvk.ServiceEntry)
+	// if we can't use delta, fall back to generate all
+	if !delta || updates == nil || len(updates.ConfigsUpdated) == 0 {
```
howardjohn, line 145: Why would `updates == nil`?

pilot/pkg/networking/core/configgen.go
```diff
@@ -37,6 +37,10 @@ type ConfigGenerator interface {
 	// BuildClusters returns the list of clusters for the given proxy. This is the CDS output
 	BuildClusters(node *model.Proxy, push *model.PushContext) ([]*discovery.Resource, model.XdsLogDetails)
 
+	// BuildDeltaClusters returns both a list of resources that need to be pushed for a given proxy and a list of resources
+	// that have been deleted and should be removed from a given proxy. This is Delta CDS output.
+	BuildDeltaClusters(node *model.Proxy, push *model.PushContext, updates *model.PushRequest) ([]*discovery.Resource, []*discovery.Resource, model.XdsLogDetails)
```
howardjohn, line 42: I don't think returning `[]*discovery.Resource` for deleted is the right approach. There is no need to build the entire cluster just to tell Envoy to remove it . I think it would not work anyways, as if the service is deleted, then `proxy.SidecarScope.Service` will return `nil` and we skip the whole thing.

We probably just want to return the names

pilot/pkg/model/context.go
howardjohn: can you remove now that its unused? Makes the PR review eaieeasier

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -260,11 +260,12 @@ func getUpdatedConfigs(services []*model.Service) map[model.ConfigKey]struct{} {
 func (s *ServiceEntryStore) serviceEntryHandler(old, curr config.Config, event model.Event) {
 	cs := convertServices(curr)
 	configsUpdated := map[model.ConfigKey]struct{}{}
+	updateTypes := make(map[model.ConfigKey]model.Event, 0)
 
 	// If it is add/delete event we should always do a full push. If it is update event, we should do full push,
 	// only when services have changed - otherwise, just push endpoint updates.
 	var addedSvcs, deletedSvcs, updatedSvcs, unchangedSvcs []*model.Service
-
+	lg.Infof("adiprerepa: YO FAM IN THE SERVICE ENTRY HANDLER. event: %v", event.String())
```
howardjohn, line 268: should remove all of these at some point

pilot/pkg/model/push_context.go
```diff
@@ -280,6 +280,9 @@ type PushRequest struct {
 	// The kind of resources are defined in pkg/config/schemas.
 	ConfigsUpdated map[ConfigKey]struct{}
 
+	// UpdateTypes tracks the action (add, update, delete) that was taken on a configuration.
+	UpdateTypes map[ConfigKey]Event
```
howardjohn, line 284: I think we don't need this. For Service, if the Service doesn't exist (ie SidecarScope.Service returns `nil`), then we know the cluster is not needed, and we can remove it.

For other resources, like DR, whether its added/removed/updated doesn't matter at all. We just regenerate the clusters and send them out.

pilot/pkg/xds/cds.go
```diff
@@ -18,10 +18,13 @@ import (
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pkg/config"
 	"istio.io/istio/pkg/config/schema/gvk"
+
 )
 
+
 type CdsGenerator struct {
 	Server *DiscoveryServer
+	Delta  bool
```
howardjohn, line 27: If we do GenerateDeltas, I think this is not needed?

howardjohn (commented):
Hopefully https://github.com/istio/istio/pull/34144 can merge before this

pilot/pkg/networking/core/configgen.go
hzxuzhonghu: Yes, the delta xds response only need to know the resource names for the deleted resources

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -119,6 +126,106 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
...
+
+func allConfigKeysOfType(cfgs map[model.ConfigKey]struct{}, cfg config.GroupVersionKind) bool {
+	keys := make([]model.ConfigKey, 0)
+	for k := range cfgs {
+		keys = append(keys, k)
+	}
+	for _, m := range keys {
+		if m.Kind != cfg {
+			return false
+		}
```
hzxuzhonghu, line 224: nit: move up, you need only one iteration

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: It is a little hard to understand, please abstract to a function. And calling it in the front of `BuildDeltaClusters`

hzxuzhonghu: updates == nil can only happen for test

pilot/pkg/networking/core/configgen.go
hzxuzhonghu: Maybe call `DeltaBuildClusters` to be consistent with `DeltaAggregatedResources`

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -119,6 +126,106 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
...
+	// delta when only services are modified and we attempt deltas
+	// we can progressively optimize this
+	log.Infof("adiprerepa: update types: %v", updates.UpdateTypes)
+	delta := updates != nil && allConfigKeysOfType(updates.ConfigsUpdated, gvk.ServiceEntry)
+	// if we can't use delta, fall back to generate all
+	if !delta || updates == nil || len(updates.ConfigsUpdated) == 0 {
+		cl, lg := configgen.BuildClusters(proxy, push)
+		return cl, make([]*discovery.Resource, 0), lg
+	}
+	configs := model.ConfigsOfKind(updates.ConfigsUpdated, gvk.ServiceEntry)
```
hzxuzhonghu, line 149: As all the `ConfigsUpdated` are SE, why need to select again?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -119,6 +126,106 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
...
+	for s := range configs {
+		// get service that changed
+		service := proxy.SidecarScope.Service(host.Name(s.Name))
+		// SidecarScope.Service will return nil if the proxy doesn't care about the service.
+		if service == nil {
```
hzxuzhonghu, line 154: it could also be that the service deleted

hzxuzhonghu: I think you donot need `UpdateTypes`, you can decide which services are deleted by proxy `WatchedResource`

hzxuzhonghu:
We need to revisit after all the comments addressed. @adiprerepa feel free to ping me on slack if any confusion

pilot/pkg/model/context.go
```diff
@@ -209,6 +209,8 @@ var DefaultXdsLogDetails XdsLogDetails = XdsLogDetails{}
 // or no response is preferred.
 type XdsResourceGenerator interface {
 	Generate(proxy *Proxy, push *PushContext, w *WatchedResource, updates *PushRequest) (Resources, XdsLogDetails, error)
+
+	GenerateDeltas(proxy *Proxy, push *PushContext, updates *PushRequest) (Resources, Resources, XdsLogDetails, error)
```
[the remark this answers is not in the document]
adiprerepa: > Also since most are not implementing it, doesn't that mean only CDS works with delta (meaning no pods can run?)?

I think I'll make it so the rest of the implementations fall back on `Generate`

pilot/pkg/model/push_context.go
adiprerepa: I see -- I will make it so `service=nil` means it was deleted

pilot/pkg/networking/core/configgen.go
adiprerepa: > We probably just want to return the names
Got it. I'll return a `[]string`
> Maybe call DeltaBuildClusters to be consistent with DeltaAggregatedResources
sure

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: so, if the service is in `WatchedResources` but the service is `nil`, we will then mark as deleted?

Do we also need to remove from `WatchedResources` ourselves?

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: good catch

pilot/pkg/networking/core/v1alpha3/cluster.go
howardjohn: We should not touch WatchedResources here I think. If we need to touch WatchedResources (which, isn't it empty for CDS? Or is that only for sotw), we can do it in delta.go?

pilot/pkg/xds/cds.go
adiprerepa: yes

pilot/pkg/networking/core/configgen.go
adiprerepa: @ramaraochavali addressed

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: @howardjohn WatchedResources isnt actually empty, I checked and `WatchedResources.ResourceNames` is populated with all the cluster names. I used it to cross-reference and see which cluster name was deleted, and returned that.

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -119,6 +120,93 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
...
+		return cl, make([]string, 0), lg
+	}
+	for s := range updates.ConfigsUpdated {
+		// get service that changed
+		service := proxy.SidecarScope.Service(host.Name(s.Name))
+		// SidecarScope.Service will return nil if the proxy doesn't care about the service OR it was deleted.
+		// we can just send that it was a removed_resource, because even if it wasn't, envoy will silently ignore.
+		if service == nil {
+			for _, n := range watched.ResourceNames {
+				if strings.Contains(n, s.Name) {
```
howardjohn, line 148: strings.Contains is not sufficient (think foo.com vs example.foo.com), should probably use model.ParseSubsetKey to exact match the hostname.

adiprerepa: good catch, applied

pilot/pkg/model/sidecar.go
```diff
@@ -473,6 +473,13 @@ func (sc *SidecarScope) Services() []*Service {
 	return sc.services
 }
 
+func (sc *SidecarScope) Service(serviceHostname host.Name) *Service {
```
hzxuzhonghu, line 476: I think this is same as `PushContext.ServiceForHostname`

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -119,6 +121,98 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
...
+	cb := NewClusterBuilder(proxy, push, configgen.Cache)
+	instances := proxy.ServiceInstances
+	// delta when only services are modified and we attempt deltas
+	// we can progressively optimize this
+	delta := updates != nil && allConfigKeysOfType(updates.ConfigsUpdated, gvk.ServiceEntry)
+	// if we can't use delta, fall back to generate all
+	if !delta || len(updates.ConfigsUpdated) == 0 {
+		cl, lg := configgen.BuildClusters(proxy, push)
+		return cl, make([]string, 0), lg
+	}
```
hzxuzhonghu, line 142: I did suggest moving this front so that we will not redundantly call `push.EnvoyFilters` and `NewClusterBuilder`

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -119,6 +121,98 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
...
+	for s := range updates.ConfigsUpdated {
+		// get service that changed
+		service := proxy.SidecarScope.Service(host.Name(s.Name))
+		// SidecarScope.Service will return nil if the proxy doesn't care about the service OR it was deleted.
+		// we can just send that it was a removed_resource, because even if it wasn't, envoy will silently ignore.
```
hzxuzhonghu, line 147: are the comments still valid, i can see you compare with watched resources below

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: maybe as a followup, the below code should be reused, otherwise it will be hard to maintain

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: ack

pilot/pkg/xds/nds.go
```diff
@@ -79,3 +79,9 @@ func (n NdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *m
 	resources := model.Resources{&discovery.Resource{Resource: util.MessageToAny(nt)}}
 	return resources, model.DefaultXdsLogDetails, nil
 }
+
+func (n *NdsGenerator) GenerateDeltas(proxy *model.Proxy, push *model.PushContext, updates *model.PushRequest,
+	w *model.WatchedResource) (model.Resources, []string, model.XdsLogDetails, error) {
+	res, logs, err := n.Generate(proxy, push, w, updates)
+	return res, make([]string, 0), logs, err
```
hzxuzhonghu, line 86: nit: s/make([]string, 0)/nil

hzxuzhonghu (commented):
overall lg

pilot/pkg/networking/apigen/apigen.go
```diff
@@ -170,6 +170,12 @@ func (g *APIGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *
 	return resp, model.DefaultXdsLogDetails, nil
 }
 
+func (g *APIGenerator) GenerateDeltas(proxy *model.Proxy, push *model.PushContext, updates *model.PushRequest,
+	w *model.WatchedResource) (model.Resources, []string, model.XdsLogDetails, error) {
+	res, logs, err := g.Generate(proxy, push, w, updates)
+	return res, make([]string, 0), logs, err
```
ramaraochavali, line 176: What is the second return value? What does empty []string indicate?

ramaraochavali: I see it is list of deleted resources ? Do we plan to implement it later? If yes, can you please leave a TODO here?

pilot/pkg/networking/core/configgen.go
ramaraochavali: Thank you

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -119,6 +121,98 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
 	return resources, model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", cacheStats.hits, cacheStats.hits+cacheStats.miss)}
 }
 
+// BuildDeltaClusters will generate the deltas (add and delete) for a given proxy. Currently, only service changes are reflected with deltas.
```
ramaraochavali, line 124: nit: /s/BuildDeltaClusters will generate the deltas/BuildDeltaClusters generates

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -119,6 +121,98 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
...
+	removedClusterNames := make([]string, 0)
+	services := make([]*model.Service, 0)
+	envoyFilterPatches := push.EnvoyFilters(proxy)
+	cb := NewClusterBuilder(proxy, push, configgen.Cache)
+	instances := proxy.ServiceInstances
+	// delta when only services are modified and we attempt deltas
+	// we can progressively optimize this
+	delta := updates != nil && allConfigKeysOfType(updates.ConfigsUpdated, gvk.ServiceEntry)
+	// if we can't use delta, fall back to generate all
+	if !delta || len(updates.ConfigsUpdated) == 0 {
```
ramaraochavali, line 139: can we fold this len(updates.ConfigsUpdated) == 0 also to delta so that do not have to revisit here like 
`delta := updates != nil && len(updates.ConfigsUpdated) > 0 && allConfigKeysOfType(updates.ConfigsUpdated, gvk.ServiceEntry)`

pilot/pkg/networking/core/v1alpha3/cluster.go
ramaraochavali: Can you please leave a TODO to refactor this to a common method?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -119,6 +121,98 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, push *mo
...
+	clusters := make([]*cluster.Cluster, 0)
+	resources := model.Resources{}
+	removedClusterNames := make([]string, 0)
+	services := make([]*model.Service, 0)
+	envoyFilterPatches := push.EnvoyFilters(proxy)
+	cb := NewClusterBuilder(proxy, push, configgen.Cache)
+	instances := proxy.ServiceInstances
+	// delta when only services are modified and we attempt deltas
+	// we can progressively optimize this
+	delta := updates != nil && allConfigKeysOfType(updates.ConfigsUpdated, gvk.ServiceEntry)
```
ramaraochavali, line 137: allConfigKeysOfType - This check is not clear to me. I may be missing some thing. So if PushRequest contains other configs like VS, DR other than service - we do not generate deltas? Any reason??

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -152,14 +246,18 @@ func buildClusterKey(service *model.Service, port *model.Port, cb *ClusterBuilde
...
 	var services []*model.Service
 	if features.FilterGatewayClusterConfig && cb.proxy.Type == model.Router {
 		services = cb.push.GatewayServices(cb.proxy)
 	} else {
 		services = cb.push.Services(cb.proxy)
 	}
+	return configgen.buildOutboundClustersWithServices(cb, cp, services)
+}
+
+func (configgen *ConfigGeneratorImpl) buildOutboundClustersWithServices(cb *ClusterBuilder, cp clusterPatcher, services []*model.Service) ([]*discovery.Resource, cacheStats) {
```
ramaraochavali, line 258: Any specific reason why this has been refactored to separate method? I am  fine with this but just curious about the reason? Do we intend to reuse some where?

pilot/pkg/xds/cds.go
```diff
@@ -68,11 +68,23 @@ func cdsNeedsPush(req *model.PushRequest, proxy *model.Proxy) bool {
 	return false
 }
 
-func (c CdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *model.WatchedResource,
-	req *model.PushRequest) (model.Resources, model.XdsLogDetails, error) {
-	if !cdsNeedsPush(req, proxy) {
+func (c CdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *model.WatchedResource, updates *model.PushRequest) (model.Resources, model.XdsLogDetails, error) {
+	// todo -- right now, we push some clusters (inbound for sidecars, passthrough and blackhole for sidecars and gateways) regardless.
```
ramaraochavali, line 72: Is this still valid in regular Generate?

pilot/pkg/networking/apigen/apigen.go
ramaraochavali: The same comment applies for all other than CdsGenerator where we have actually implementation?

ramaraochavali (commented):
This looks generally good. Few more comments.

pilot/pkg/networking/apigen/apigen.go
adiprerepa: yeah, []string indicates that no resources were removed, which is what we want. I'm not sure if we should add a TODO because we may not want to support delta for resoures other than LDS and CDS.

With the rest of the resources, we just fall back on the SOTW implementation, and just use the `resources` field in the `DeltaDiscoveryResponse`, with nothing set in `removed_resources` (hence returning an empty slice)

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: the reason is so we don't have a very large change in one PR. For now, we decided it would be best to start of with services, and incrementally move onto DestinationRule, VirtualService, etc.

We will eventually generate deltas for all, at which point we will not need this check.

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: yes. the default implementation (`buildOutboundClusters`), would build all outbound clusters, using all the services (see https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/v1alpha3/cluster.go#L161). For delta, we need to generate only based on a set of services, which are defined as which services changed. So, I needed a way to pass in the services based on which it generates the clusters.

pilot/pkg/xds/cds.go
adiprerepa: good catch

pilot/pkg/xds/nds.go
adiprerepa: got it

pilot/pkg/xds/cds.go
```diff
@@ -68,11 +68,21 @@ func cdsNeedsPush(req *model.PushRequest, proxy *model.Proxy) bool {
...
 }
+
+// GenerateDeltas for CDS currently only builds deltas when services change. todo implement changes for DestinationRule, etc
+func (c CdsGenerator) GenerateDeltas(proxy *model.Proxy, push *model.PushContext, updates *model.PushRequest,
+	w *model.WatchedResource) (model.Resources, []string, model.XdsLogDetails, error) {
+	if !cdsNeedsPush(updates, proxy) {
+		return nil, nil, model.DefaultXdsLogDetails, nil
+	}
+	updatedClusters, removedClusters, logs := c.Server.ConfigGenerator.BuildDeltaClusters(proxy, push, updates, w)
+	log.Infof("UPDATED: %v REMOVED %v", updatedClusters, removedClusters)
```
howardjohn, line 86: remove

pilot/pkg/xds/delta.go
```diff
@@ -448,11 +448,7 @@ func (s *DiscoveryServer) pushDeltaXds(con *Connection, push *model.PushContext,
 		Nonce:             nonce(push.LedgerVersion),
 		Resources:         res,
 	}
-	// We take the set of watched resources and anything not in the response is sent as RemovedResources
-	// This is similar to SotW, but done on the server side instead of the client.
-	cur := sets.NewSet(w.ResourceNames...)
```
howardjohn, line 453: I think you still need this code for the generators that don't have first class delta support? otherwise we never delete

pilot/pkg/networking/apigen/apigen.go
ramaraochavali: Ok. Makes sense. Slight preference to return `[]string{}` instead of make. but it is a huge nit.

pilot/pkg/networking/core/v1alpha3/cluster.go
ramaraochavali: Got it. So delta would kick in only when updates have all services? In that case I would suggest to add method like 

`shouldUseDelta` and pass in updates to it. and there you check if updates has only service changes. So slowly we can make changes to that method - we can reuse that method for listeners etc. So we do not have to change individual builders

ramaraochavali (commented):
Looks good - few more minor comments. Let me know once that is done, I am good with this as first step.

pilot/pkg/xds/delta.go
adiprerepa: fixed

howardjohn (approved):

pilot/pkg/model/sidecar.go
hzxuzhonghu: @adiprerepa This method is not used any more

hzxuzhonghu (approved):

pilot/pkg/networking/core/v1alpha3/cluster.go
ramaraochavali: @adiprerepa This is missed. But I am ok to do this as follow-up

ramaraochavali (approved):

PR #34031, opened 2021-07-14
fix kube client logs format

gy95:
1. fix some logs: should use `namespace/name` format but not use `name/namespace` format

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/kube/portforwarder.go
+++ b/pkg/kube/portforwarder.go
@@ -30,7 +30,7 @@ import (
 
 // PortForwarder manages the forwarding of a single port.
 type PortForwarder interface {
-	// Run this forwarder.
+	// Start runs this forwarder.
 	Start() error
 
 	// Address returns the local forwarded address. Only valid while the forwarder is running.
@@ -39,7 +39,7 @@ type PortForwarder interface {
 	// Close this forwarder and release an resources.
 	Close()
 
-	// Block until connection closed (e.g. control-C interrupt)
+	// WaitForStop blocks until connection closed (e.g. control-C interrupt)
 	WaitForStop()
 }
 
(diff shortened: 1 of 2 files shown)


pkg/kube/portforwarder.go
```diff
@@ -39,7 +39,7 @@ type PortForwarder interface {
 	// Close this forwarder and release an resources.
 	Close()
 
-	// Block until connection closed (e.g. control-C interrupt)
+	// WaitForStop block until connection closed (e.g. control-C interrupt)
```
hzxuzhonghu, line 42: blocks

pkg/kube/portforwarder.go
```diff
@@ -30,7 +30,7 @@ import (
 
 // PortForwarder manages the forwarding of a single port.
 type PortForwarder interface {
-	// Run this forwarder.
+	// Start run this forwarder.
```
hzxuzhonghu, line 33: runs

hzxuzhonghu (commented):
Is there any difference?

pkg/kube/portforwarder.go
gy95: ok

gy95:
> Is there any difference?

namespace should precede name, separeated by '/', more conventional

hzxuzhonghu (approved):
/ok-to-test

hzxuzhonghu:
@gy95 You can apply for istio member, so that your pr can be tested without other members to label `/ok-to-test`

PR #34033, opened 2021-07-14
In the scene where service entry select workload entry：set up istio endpoint port

ChenglongGao327:
In the scene where service entry select workload entry
The priority of the model.IstioEndPoint.EndpointPort port should be : unixAddress > we.ports > se.port.targetPort > se.port.number

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pilot/pkg/serviceregistry/serviceentry/conversion.go
+++ b/pilot/pkg/serviceregistry/serviceentry/conversion.go
@@ -228,14 +228,12 @@ func convertEndpoint(service *model.Service, servicePort *networking.Port,
 	wle *networking.WorkloadEntry, configKey *configKey, clusterID cluster.ID) *model.ServiceInstance {
 	var instancePort uint32
 	addr := wle.GetAddress()
+	// priority level: unixAddress > we.ports > se.port.targetPort > se.port.number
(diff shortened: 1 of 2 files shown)


howardjohn (dismissed):
Thanks for the PR! can you add a test? We have a bunch of similar ones in https://github.com/istio/istio/blob/eea231b5a2a7d88258a73de903fe556d7c970bce/pilot/pkg/serviceregistry/serviceregistry_test.go#L262 I think

ChenglongGao327:
> Thanks for the PR! can you add a test? We have a bunch of similar ones in
> 
> https://github.com/istio/istio/blob/eea231b5a2a7d88258a73de903fe556d7c970bce/pilot/pkg/serviceregistry/serviceregistry_test.go#L262

add test for this pr

stevenctl (approved):

stevenctl:
LGTM - #33774 changes the tests to better simulate k8s a bit, so that _may_ change the behavior of this or work in combination with it

hzxuzhonghu (approved):

ChenglongGao327:
@howardjohn  please approvals

Issue #33977, opened 2021-07-12
Improve documentation on namespace scoping of routes


howardjohn:
Gateways allow namespace bound virtual service selection (eg `foo/foo.example.com`). This is a good security best practice, but not present in docs or samples. We should spread this usage, as most users are copy+pasting from our examples.

Ruoyu-y:
@howardjohn Are we trying to create a new example to reflect this? Or we just want to add the gateway namespaces specified?

howardjohn:
Probably not new example, just make existing examples use it where possible

PR #34035, opened 2021-07-14
Apply namespace scoping to Gateway servers in samples

kailun-qin:
Gateways allow namespace bound virtual service selection (e.g.,
foo/foo.example.com). This is a good security best practice.

This patch spreads this usage to all samples to restrict the set of
virtual services that can bind to a gateway server.

Fixes https://github.com/istio/istio/issues/33977

Signed-off-by: Kailun Qin <kailun.qin@intel.com>

[x] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/samples/bookinfo/networking/bookinfo-gateway.yaml
+++ b/samples/bookinfo/networking/bookinfo-gateway.yaml
@@ -2,6 +2,7 @@ apiVersion: networking.istio.io/v1alpha3
 kind: Gateway
 metadata:
   name: bookinfo-gateway
+  namespace: default
 spec:
   selector:
     istio: ingressgateway # use istio default controller
@@ -17,6 +18,7 @@ apiVersion: networking.istio.io/v1alpha3
 kind: VirtualService
 metadata:
   name: bookinfo
+  namespace: default
 spec:
   hosts:
   - "*"
--- a/samples/bookinfo/networking/bookinfo-gateway.yaml
+++ b/samples/bookinfo/networking/bookinfo-gateway.yaml
@@ -11,7 +11,7 @@ spec:
       name: http
       protocol: HTTP
     hosts:
-    - "*"
+    - "default/*"
 ---
 apiVersion: networking.istio.io/v1alpha3
 kind: VirtualService
--- a/samples/bookinfo/networking/certmanager-gateway.yaml
+++ b/samples/bookinfo/networking/certmanager-gateway.yaml
@@ -12,7 +12,7 @@ spec:
       name: http
       protocol: HTTP
     hosts:
-    - "*"
+    - "istio-system/*"
 ---
 apiVersion: networking.istio.io/v1alpha3
 kind: VirtualService
--- a/samples/helloworld/helloworld-gateway.yaml
+++ b/samples/helloworld/helloworld-gateway.yaml
@@ -2,6 +2,7 @@ apiVersion: networking.istio.io/v1alpha3
 kind: Gateway
 metadata:
   name: helloworld-gateway
+  namespace: default
 spec:
   selector:
     istio: ingressgateway # use istio default controller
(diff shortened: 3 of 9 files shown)


nrjpoddar (commented):
Love these changes! Thanks for doing this @kailun-qin.

We would need to change the istio.io examples/tasks too or else the docs and bundled sample resources will be out of sync?

craigbox:
We should also give though to whether it makes it more complicated to understand by looking at the samples, as these are some of the things we use to teach Istio to people.

samples/bookinfo/networking/bookinfo-gateway.yaml
```diff
@@ -11,7 +11,7 @@ spec:
       name: http
       protocol: HTTP
     hosts:
-    - "*"
+    - "default/*"
```
howardjohn, line 14: Problem with this is the YAML itself is not namespace bound. So they can apply anywhere, and then this won't work.

Maybe we should have allowed `./*` in the code...

kailun-qin: Yep, an alternative is to explicitly specify the namespaces for the YAMLs so that they can be *bound* together. The namespace to bind should align w/ the sample READMEs.

In this way, when `kubectl apply` w/ no explicit ns set, the resources will by default be applied to the exact ns specified in the Pod spec. And when `kubectl apply` w/ explicit `-n <ns_name>`, there'll be an error returned to users when they try to apply it elsewhere (message like `the namespace from the provided object "istio-system" does not match the namespace "default". You must pass '--namespace=istio-system' to perform this operation.`), which is guaranteed by kubectl.

Does this approach make sense to you @howardjohn? Thanks!

kailun-qin:
> Love these changes! Thanks for doing this @kailun-qin.
> 
> We would need to change the istio.io examples/tasks too or else the docs and bundled sample resources will be out of sync?

Thanks! Yes, we do need to update the related if this change itself makes sense.

kailun-qin:
> We should also give though to whether it makes it more complicated to understand by looking at the samples, as these are some of the things we use to teach Istio to people.

Agree. IMO this looks fine for readability. Additional doc/comments can be updated if not clear enough.

samples/bookinfo/networking/bookinfo-gateway.yaml
frankbu: Some of these resources (e.g., bookinfo ones) are intended to be used in whatever namespace the user wants. Even the Istio tasks use different namepaces (e.g., some use default, others use ns bookinfo). Outside blogs and other docs also rely on the namespace flexibility, so this would be unacceptable breaking change IMO.

kailun-qin: Thanks for the comments!

If it's the case, I think maybe we should have allowed `*` in the code. We should only apply ns scoping to the samples where the YAMLs have already bound w/ specific ns, e.g. `samples/bookinfo/networking/certmanager-gateway.yaml`, `samples/multicluster` etc.

WDYT @frankbu @howardjohn?

frankbu: > We should only apply ns scoping to the samples where the YAMLs have already bound w/ specific ns

That makes sense but I don't think there are very many of those. The `samples/multicluster` east-west gateways are not always in `istio-system`. For example, see the external control plane doc: https://preliminary.istio.io/latest/docs/setup/install/external-controlplane/#setup-east-west-gateways

frankbu: btw, another recommended best practice is to use fully-qualified hosts in VirtualServices (e.g., `reviews.default.svc.cluster.local` insead of just `reviews`), but we haven't done that in the samples either, for the reason reason.

Maybe we need to add more [best practices](https://preliminary.istio.io/latest/docs/ops/best-practices/traffic-management/) to the docs and put some tips in the tasks/examples that point to them.

PR #34037, opened 2021-07-14
Fuzzing: Add v1alpha3 fuzzer

AdamKorcz:
Adds a v1alpha3 fuzzer.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/fuzz/v1alpha3_fuzzer.go
+++ b/tests/fuzz/v1alpha3_fuzzer.go
@@ -1,0 +1,47 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// nolint: golint
+package fuzz
+
+import (
+	"testing"
+
+	fuzz "github.com/AdaLogics/go-fuzz-headers"
+
+	"istio.io/istio/pilot/pkg/model"
+	"istio.io/istio/pilot/pkg/networking/core/v1alpha3"
+)
+
+func init() {
+	testing.Init()
+}
+
+func FuzzValidateClusters(data []byte) int {
+	t := &testing.T{}
+	proxy := model.Proxy{}
+	f := fuzz.NewConsumer(data)
+	err := f.GenerateStruct(&proxy)
+	if err != nil {
+		return 0
+	}
+	to := v1alpha3.TestOptions{}
+	err = f.GenerateStruct(&to)
+	if err != nil {
+		return 0
+	}
+	cg := v1alpha3.NewConfigGenTest(t, to)
+	_ = cg.Clusters(cg.SetupProxy(&proxy))
+	return 1
+}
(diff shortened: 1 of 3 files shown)


tests/fuzz/v1alpha3_fuzzer.go
```diff
@@ -0,0 +1,57 @@
...
+	t := &testing.T{}
+	proxy := model.Proxy{}
+	f := fuzz.NewConsumer(data)
+	err := f.GenerateStruct(&proxy)
+	if err != nil {
+		return 0
+	}
+	to := v1alpha3.TestOptions{}
+	cg := v1alpha3.NewConfigGenTest(t, to)
+	clusters := cg.Clusters(cg.SetupProxy(&proxy))
```
howardjohn, line 42: I think this will always generate empty response?

howardjohn:
I am not sure I quite understand the rationale of fuzzing our test helpers. It seems like if they crash during fuzzing of non-test code we can fix it, and if not then we don't care if they can crash on some obscure input since its just for tests?

AdamKorcz:
> I am not sure I quite understand the rationale of fuzzing our test helpers. It seems like if they crash during fuzzing of non-test code we can fix it, and if not then we don't care if they can crash on some obscure input since its just for tests?

If the test helpers cannot run continuously we will not be able to use them in the fuzzing suite. `NewConfigGenTest` for example would be useful to target several different v1alpha3 apis, but if `NewConfigGenTest` itself is a fuzz-blocker, this will not be possible.
A concrete scenario could be if a test helper runs out of memory after 1 minute or it triggers a runtime error. This will be a blocker for using a given test helper in fuzzer.

howardjohn:
Sure, but if that is the case won't we find out when we try to use it?
Basically indirectly testing it rather than directly

AdamKorcz:
> Sure, but if that is the case won't we find out when we try to use it? Basically indirectly testing it rather than directly

Yes

AdamKorcz:
@howardjohn Would you prefer to write fuzzers _with_ the test helpers rather than _for_ the test helpers? If so, then I would be glad to amend this PR.

howardjohn:
Yeah that is what I was suggesting.

Although the goal is not "to write fuzzers *with* the test helpers", but
rather "to write useful fuzzers, and using the test helpers may assist in
doing that". I don't care if the test helpers are tested, or used - they
are just another tool.

AdamKorcz:
Sounds good. Let's keep this PR open, I will amend the fuzzers to reflect that.

tests/fuzz/v1alpha3_fuzzer.go
AdamKorcz: My observation based on local tests is that `len(clusters)` is not `0` about 15% of the time.

AdamKorcz:
The fuzzer that targeted `NewConfigGenTest` has been removed.

tests/fuzz/v1alpha3_fuzzer.go
howardjohn: I think when its non-zero its also probably not getting meaningful config generation. I would guess we hit https://github.com/istio/istio/blob/fb62ce54058a1749ab747d5dd2780288525d1b76/pilot/pkg/networking/core/v1alpha3/cluster.go#L89 which is just building a static cluster that never changes. The real coverage would come in at https://github.com/istio/istio/blob/fb62ce54058a1749ab747d5dd2780288525d1b76/pilot/pkg/networking/core/v1alpha3/cluster.go#L163 which would require the push context to be non-empty to ever enter the loop

AdamKorcz: I see. You are correct that the fuzzer does not make into [this loop](https://github.com/istio/istio/blob/19fb530b0c6176404fdba48de8d104f76d81cdea/pilot/pkg/networking/core/v1alpha3/cluster.go#L163)  on shorter runs. 

OSS-fuzz provides tooling that is great for tracking this, so what do you think about not having a branch in the fuzzer for the cluster validation and improve coverage once we get feedback from OSS-fuzz based on a few hundreds of hours of fuzzing? The current version reflects this suggestion.

howardjohn: Ahh, I was missing that we fuzz TestOptions which can generate Services. I thought we only fuzzed Proxy which would never enter. SGTM then.

Maybe if the fuzzer doesn't get much overage we can add some seed info

howardjohn (approved):

PR #34047, opened 2021-07-14
Uninstall should remove validating webhooks

Monkeyanator:
In 1.11 we install per-revision validating webhooks. These should be removed on `istioctl uninstall`.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/operator/pkg/helmreconciler/prune.go
+++ b/operator/pkg/helmreconciler/prune.go
@@ -73,6 +73,7 @@ var (
 	// ClusterCPResources lists cluster scope resources types which should be deleted during uninstall command.
 	ClusterCPResources = []schema.GroupVersionKind{
 		{Group: "admissionregistration.k8s.io", Version: "v1", Kind: name.MutatingWebhookConfigurationStr},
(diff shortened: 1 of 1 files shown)


operator/pkg/helmreconciler/prune.go
```diff
@@ -73,6 +73,7 @@ var (
 	// ClusterCPResources lists cluster scope resources types which should be deleted during uninstall command.
 	ClusterCPResources = []schema.GroupVersionKind{
 		{Group: "admissionregistration.k8s.io", Version: "v1", Kind: name.MutatingWebhookConfigurationStr},
+		{Group: "admissionregistration.k8s.io", Version: "v1", Kind: name.ValidatingWebhookConfigurationStr},
```
howardjohn, line 76: Will this also remove the ones from `base`?

Monkeyanator: The one in base gets `istio.io/rev` from the last revision installed, so if `istioctl uninstall`ing the last revision installed it will be removed... so maybe this is dangerous...

Issue #30834, opened 2021-02-12
Istiod CRD client: support dynamically added CRDs


howardjohn:
Currently, Istiod supports ignoring CRDs if they are not present. However, if a CRD is added later it will not be picked up.

This is especialy troublesome with multiple instances. We may have 1 Istiod with a CRD being read, and another without. For example, I install k8s gateway CRD, then scale up istiod. my gateway works until it connects to the old Istiod and breaks.

instead, we can just watch CRDs. We probably only need to watch them to  *add* CRDs - I don't think we have any requirement to support dynamic removal, although we can look into that as well.

Requirement: This should use metadata-only watch. CRDs are *massive* and we have had leaks in Galley causing OOM and cluster outages due to the size of CRDs that constantly change.

PR #34056, opened 2021-07-14
Handle CRDs created after Istio starts

howardjohn:
Fixes https://github.com/istio/istio/issues/30834

This is especially useful as we add support for more third party CRDs
(gateway-api and MCS). Currently, using these resources requires a
restart of Istiod. In normal cases, this is not ideal. In worst case
(external Istiod) this is impossible.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/config/kube/crdclient/cache_handler.go
+++ b/pilot/pkg/config/kube/crdclient/cache_handler.go
@@ -70,7 +70,7 @@ func (h *cacheHandler) onEvent(old interface{}, curr interface{}, event model.Ev
 }
 
 func createCacheHandler(cl *Client, schema collection.Schema, i informers.GenericInformer) *cacheHandler {
-	scope.Debugf("registered CRD %v", schema.Resource())
+	scope.Debugf("registered CRD %v", schema.Resource().GroupVersionKind())
 	h := &cacheHandler{
 		client:   cl,
 		schema:   schema,
--- a/pilot/pkg/config/kube/crdclient/cache_handler.go
+++ b/pilot/pkg/config/kube/crdclient/cache_handler.go
@@ -36,7 +36,6 @@ import (
 type cacheHandler struct {
 	client   *Client
 	informer cache.SharedIndexInformer
-	handlers []func(config.Config, config.Config, model.Event)
 	schema   collection.Schema
 	lister   func(namespace string) cache.GenericNamespaceLister
 }
@@ -64,13 +63,14 @@ func (h *cacheHandler) onEvent(old interface{}, curr interface{}, event model.Ev
 	}
 
 	// TODO we may consider passing a pointer to handlers instead of the value. While spec is a pointer, the meta will be copied
-	for _, f := range h.handlers {
+	for _, f := range h.client.handlers[h.schema.Resource().GroupVersionKind()] {
 		f(oldConfig, currConfig, event)
 	}
 	return nil
 }
 
 func createCacheHandler(cl *Client, schema collection.Schema, i informers.GenericInformer) *cacheHandler {
+	scope.Debugf("registered CRD %v", schema.Resource())
 	h := &cacheHandler{
 		client:   cl,
 		schema:   schema,
--- a/pilot/pkg/config/kube/crdclient/client.go
+++ b/pilot/pkg/config/kube/crdclient/client.go
@@ -57,6 +57,7 @@ import (
 	"istio.io/istio/pkg/config"
 	"istio.io/istio/pkg/config/schema/collection"
 	"istio.io/istio/pkg/config/schema/collections"
+	"istio.io/istio/pkg/config/schema/gvk"
 	"istio.io/istio/pkg/kube"
 	"istio.io/istio/pkg/queue"
 	"istio.io/pkg/log"
@@ -78,8 +79,12 @@ type Client struct {
 	revision string
 
 	// kinds keeps track of all cache handlers for known types
-	kinds map[config.GroupVersionKind]*cacheHandler
-	queue queue.Instance
+	kinds   map[config.GroupVersionKind]*cacheHandler
+	kindsMu sync.RWMutex
+	queue   queue.Instance
+
+	// handlers defines a list of event handlers per-type
+	handlers map[config.GroupVersionKind][]model.EventHandler
 
 	// The istio/client-go client we will use to access objects
 	istioClient istioclient.Interface
@@ -90,7 +95,10 @@ type Client struct {
 	// beginSync is set to true when calling SyncAll, it indicates the controller has began sync resources.
 	beginSync *atomic.Bool
 	// initialSync is set to true after performing an initial processing of all objects.
-	initialSync *atomic.Bool
+	initialSync         *atomic.Bool
+	schemasByCRDName    map[string]collection.Schema
+	client              kube.Client
+	crdMetadataInformer cache.SharedIndexInformer
 }
 
 var _ model.ConfigStoreCache = &Client{}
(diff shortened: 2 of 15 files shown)


tests/integration/pilot/main_test.go
```diff
@@ -52,16 +51,6 @@ func supportsCRDv1(t resource.Context) bool {
 func TestMain(m *testing.M) {
 	framework.
 		NewSuite(m).
-		Setup(func(t resource.Context) (err error) {
```
howardjohn, line 55: This effectively tests this change - now we create the CRDs after install

pkg/config/schema/metadata.yaml
```diff
@@ -256,7 +256,7 @@ snapshots:
 resources:
   # Kubernetes specific configuration.
   - kind: "CustomResourceDefinition"
-    plural: "CustomResourceDefinitions"
+    plural: "customresourcedefinitions"
```
adiprerepa, line 259: why this change?

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -349,6 +368,23 @@ func (cl *Client) objectInRevision(o *config.Config) bool {
...
+	defer cl.kindsMu.RUnlock()
+	ret := make([]*cacheHandler, 0, len(cl.kinds))
+	for _, k := range cl.kinds {
+		ret = append(ret, k)
+	}
+	return ret
+}
+
+func (cl *Client) kind(r config.GroupVersionKind) (*cacheHandler, bool) {
+	cl.kindsMu.RLock()
```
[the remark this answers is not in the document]
howardjohn: Inside handleCRDAdd, which can now happen after startup at any time

pkg/config/schema/metadata.yaml
howardjohn: The old way was wrong, but in a way that didn't matter because we didn't use it anywhere. Expected `plural` is all lower case, which is used by all the others.

For example:
```
$ kubectl get crd -oyaml | rg plural:
      plural: authorizationpolicies
      plural: backendconfigs
....
```

howardjohn:
ping @hzxuzhonghu if you have a chance

hzxuzhonghu:
MIssed it

pilot/pkg/config/kube/crdclient/cache_handler.go
```diff
@@ -64,13 +63,14 @@ func (h *cacheHandler) onEvent(old interface{}, curr interface{}, event model.Ev
...
 	// TODO we may consider passing a pointer to handlers instead of the value. While spec is a pointer, the meta will be copied
-	for _, f := range h.handlers {
+	for _, f := range h.client.handlers[h.schema.Resource().GroupVersionKind()] {
 		f(oldConfig, currConfig, event)
 	}
 	return nil
 }
 
 func createCacheHandler(cl *Client, schema collection.Schema, i informers.GenericInformer) *cacheHandler {
+	scope.Debugf("registered CRD %v", schema.Resource())
```
hzxuzhonghu, line 73: schema.Resource().GroupVersionKind()

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -423,3 +459,56 @@ func genPatchBytes(oldRes, modRes runtime.Object, patchType types.PatchType) ([]
...
+	if !f {
+		scope.Debugf("added resource that we are not watching: %v", name)
+		return
+	}
+	resourceGVK := s.Resource().GroupVersionKind()
+	gvr := s.Resource().GroupVersionResource()
+
+	cl.kindsMu.Lock()
+	defer cl.kindsMu.Unlock()
+	if ch, f := cl.kinds[s.Resource().GroupVersionKind()]; f {
```
hzxuzhonghu, line 475: cl.kinds[resourceGVK]

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -187,14 +190,29 @@ func (cl *Client) Run(stop <-chan struct{}) {
...
+
+	cl.crdMetadataInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			crd, ok := obj.(*metav1.PartialObjectMetadata)
+			if !ok {
+				// Shouldn't happen
+				scope.Errorf("wrong type %T: %v", obj, obj)
+				return
+			}
+			handleCRDAdd(cl, crd.Name, stop)
```
hzxuzhonghu, line 202: This is time costing, it is not a good practice to synchronously handle event

howardjohn: I think the only place this may block is in the waitforcachesync part. I may be able to remove that - if so there is nothing doing any IO here so it should be fast enough?

if not maybe can move this into go routine?

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -423,3 +459,56 @@ func genPatchBytes(oldRes, modRes runtime.Object, patchType types.PatchType) ([]
...
+		// Shouldn't happen
+		scope.Errorf("failed to create informer for %v", resourceGVK)
+		return
+	}
+	cl.kinds[s.Resource().GroupVersionKind()] = createCacheHandler(cl, s, i)
+	if stop != nil {
+		// Start informer factory, only if stop is defined. In startup case, we will not start here as
+		// we will start all factories once we are ready to initialize.
+		// For dynamically added CRDs, we need to start immediately though
+		ifactory.Start(stop)
```
hzxuzhonghu, line 508: looks there can start multi informer for a single crd, if it is created before istiod start

howardjohn: the Start code does nothing if called multiple times , so it shouldn't be an issue. It's really "start all unstarted informers"

hzxuzhonghu:
The most critical part is if the crd does not exist, istiod still start the informer, and it list-watch that cr, but with error like this 
```
2495:2021-07-21T03:09:21.565139Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)
2496:2021-07-21T03:09:27.141529Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)
2497:2021-07-21T03:09:36.565058Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)
2498:2021-07-21T03:09:49.502129Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)
2499:2021-07-21T03:10:40.019219Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)
2500:2021-07-21T03:11:15.959366Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)

```

Then after WorkloadEntry crd created, no error thrown?

howardjohn:
> The most critical part is if the crd does not exist, istiod still start the informer, and it list-watch that cr, but with error like this
> 
> ```
> 2495:2021-07-21T03:09:21.565139Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)
> 2496:2021-07-21T03:09:27.141529Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)
> 2497:2021-07-21T03:09:36.565058Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)
> 2498:2021-07-21T03:09:49.502129Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)
> 2499:2021-07-21T03:10:40.019219Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)
> 2500:2021-07-21T03:11:15.959366Z	error	watch error in cluster Kubernetes: failed to list *v1alpha3.WorkloadEntry: the server could not find the requested resource (get workloadentries.networking.istio.io)
> ```
> 
> Then after WorkloadEntry crd created, no error thrown?

Is this from my PR? The expectation is:
* at startup,start all informers for know CRDs. This means if one is missing we do not start
* If CRD is later created start it

So I wouldn't expect that to happen (but could be a bug)

hzxuzhonghu:
This is from the master branch

hzxuzhonghu:
>This means if one is missing we do not start

How do you do this?

howardjohn:
> > This means if one is missing we do not start
> 
> How do you do this?
The code for this already exists,unchanged here


> This is from the master branch

That should not happen,let me investigate. We do have a unit test of this but no integ test

hzxuzhonghu:
Yes, that's weird

hzxuzhonghu:
My test way was not right, i deleted the crd after istiod started

howardjohn:
Ah makes sense. Yeah we don't handle deletion well. I think it's a bit harder and not that common of use case

howardjohn:
@hzxuzhonghu I think I have addressed the comments

hzxuzhonghu (approved):

Issue #34058, opened 2021-07-14
[Revision] Incorrect discoveryAddress for warm cert in Virtual Machine Installation


xulingqing:
**Bug description**

Use Case: 
Install VM following [guide](https://istio.io/latest/docs/setup/install/virtual-machine/) to a controlPlane with revision label.

In the workload entry generation step,  the generated hosts is correctly pointing to `<Eastwestgateway External_IP> istiod-revision-label.istio-system.svc`, but Effective config: binaryPath: /usr/local/bin/envoy  discoveryAddress is pointing to  istiod.istio-system.svc.
```
istioctl x workload entry configure --revision revision-label -f workloadgroup.yaml -o "${WORK_DIR}" --clusterID "${CLUSTER}" --autoregister
```

Observation:
Apply mesh config from file defaultConfig is correctly labeled. Apply proxy config from env will override discoveryAddress to unlabeled svc. 

**Affected product area (please put an X in all that apply)**

Virtual Machine Installation to revision labeled controlPlane.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Affected features (please put an X in all that apply)**

[X] Multi Cluster
[X] Virtual Machine
[X] Multi Control Plane

**Expected behavior**

Apply proxy config from env should not override mesh config from defaultConfig

**Steps to reproduce the bug**

Install Vm to istiod with revision label.

**Version** (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version --short` if you used Helm)

1.10.2

**How was Istio installed?**

IstioOperator with revision label in spec.

**Environment where the bug was observed (cloud vendor, OS, etc)**

- GCP [BareMetal Cluster 1.7](https://cloud.google.com/anthos/clusters/docs/bare-metal/1.7/installing/creating-clusters/hybrid-cluster-creation). 
- centos8 VM

xulingqing:
@stevenctl

xulingqing:
@liminw

xulingqing:
Fix https://github.com/istio/istio/pull/35290

xulingqing:
VM installation webpage https://istio.io/latest/docs/setup/install/virtual-machine/#configure-the-virtual-machine
change: https://github.com/istio/istio.io/pull/10380

stevenctl:
#35340

PR #34060, opened 2021-07-14
make cluster env not Override defaultConfig for PILOT_ADDRESS value

xulingqing:
Fix of https://github.com/istio/istio/issues/34058

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


xulingqing:
https://github.com/istio/istio/issues/34058

tools/packaging/common/istio-start.sh
```diff
@@ -99,10 +99,6 @@ if [ "${ISTIO_INBOUND_INTERCEPTION_MODE}" = "TPROXY" ] ; then
   EXEC_USER=root
 fi
 
-if [ -z "${PILOT_ADDRESS:-}" ]; then
```
howardjohn, line 102: I am not sure its safe to always remove this, as user may customize it. Maybe we can only not set it if ISTIO_SYSTEM_NAMESPACE and ISTIO_PILOT_PORT are unset?

tools/packaging/common/istio-start.sh
```diff
@@ -99,10 +99,6 @@ if [ "${ISTIO_INBOUND_INTERCEPTION_MODE}" = "TPROXY" ] ; then
...
-  PILOT_ADDRESS=istiod.${ISTIO_SYSTEM_NAMESPACE}.svc:${ISTIO_PILOT_PORT}
```
howardjohn, line 103: I think even ifthis is not set, `discoveryAddress` will be present in L122 and cause issues? It would be great if we have a VM+revisions test to prove this works

Issue #33965, opened 2021-07-11
Improve iptables debugability


howardjohn:
* Iptables allows inserting comments (`--comment`) which can help explain why we have certain rules.
* Iptables allows inserting logs (`iptables -A INPUT -j LOG`, for example). This could be helpful to debug the flow of traffic. I don't think you can add a log a specific line, but for every rule we create we can make a log statement with the same match I think.

howardjohn:
Like need `echo 1 | sudo tee /proc/sys/net/netfilter/nf_log_all_netns` to get logs in containers

ramaraochavali:
Huge +1 for this

howardjohn:
implementation idea:
* Setup a map of shortname -> long description.  example: comment gets `--comment "IPT001: redirect app calls back to self"`, log gets `--log-prefix "IPT001:".
* Each call to append a rule also (conditionally) adds a `-j LOG` rule before with the same match
* Each call has a `--comment` attached
* Likely need to better abstract iptables calls to do this. Probably useful for ipv4+ipv6 consistency anyways

howardjohn:
```
	//if len(params) > 2 && params[len(params)-2] == "-j" {
	//	p := append([]string{"-A", chain}, params...)
	//	p = p[:len(p)-2]
	//	p = append(p, "-j", "LOG", "--log-uid", "--log-prefix", "istio\t")
	//	rb.rules.rulesv4 = append(rb.rules.rulesv4, &Rule{
	//		chain:  chain,
	//		table:  table,
	//		params: p,
	//	})
	//}
```

as a hack to AppendV4Rules actually works pretty well

aojea:
@howardjohn small nit, the field is bounded  to 256 char ;)
```
comment

Allows you to add comments (up to 256 characters) to any rule.
--comment comment
Example:
iptables -A INPUT -s 192.168.0.0/16 -m comment --comment "A privatized IP block"
```
https://linux.die.net/man/8/iptables

PR #34061, opened 2021-07-15
Improve iptables tests

howardjohn:
This changes tests to follow a standard golden file test pattern. This makes seeing diffs, seeing expected output, making changes, etc all easier. Additionally, a ton of test code is deduplicated.

The biggest change here is we always call `run()` instead of some tests which called targeted functions. IMO this is a better testing methodology, and the increased test output sizes are worth it to keep things simpler, more realistic, and detect any unexpected changes.

For some of the tests, I modified the job to write the `expected []string` content to the golden file, to ensure the new test is identical. For a few that did non-standard things, I spot checked the result to ensure they were the same.


howardjohn:
Motivation is to help test https://github.com/istio/istio/issues/33965

tools/istio-iptables/pkg/builder/iptables_builder_impl.go
```diff
@@ -34,22 +34,22 @@ type Rules struct {
 	rulesv6 []*Rule
 }
 
-// IptablesBuilderImpl is an implementation for IptablesBuilder interface
-type IptablesBuilderImpl struct {
+// IptablesBuilder is an implementation for IptablesBuilder interface
```
hzxuzhonghu, line 37: The comments should be changed to `adding iptables rules`

And may rename the file name to iptables_builder.go

hzxuzhonghu (commented):
lgtm. i like this style

ramaraochavali (commented):
like this approach. Gold files definitely give better idea.

bianpengyuan (approved):
Verified that there is no lost of test coverage. Thanks!

PR #34064, opened 2021-07-15
istiod-hostnetwork

chandrakanthkannam:
Ability to deploy istiod pods on hostnetwork when using helm chart.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
+++ b/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
@@ -58,6 +58,9 @@ spec:
 {{- if .Values.global.priorityClassName }}
       priorityClassName: "{{ .Values.global.priorityClassName }}"
 {{- end }}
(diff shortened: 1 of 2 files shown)


howardjohn (changes requested):
We have been undergoing an effort to limit the scope of the values.yaml API to values that are fully supported,  tested, and able to be promoted to beta. I don't think this meets that bar currently. For this, and similar changes, it seems like the best option is for users with these custom requirements to post-process the helm charts through various tools (kustomize, etc)

chandrakanthkannam:
hi @howardjohn, thanks for looking at this. If this PR cannot be accepted, can you take this as a feat maybe? to include this option for users who use these helm charts directly in their environments?

Issue #34065, opened 2021-07-15
empty regex cause envoy warnings


kebe7jun:
**Bug description**

Envoy output the following logs:
```
2021-07-15T03:52:44.189014Z	warning	envoy config	gRPC config for type.googleapis.com/envoy.config.route.v3.RouteConfiguration rejected: Proto constraint validation failed (RouteConfigurationValidationError.VirtualHosts[i]: ["embedded message failed validation"] | caused by VirtualHostValidationError.Routes[i]: ["embedded message failed validation"] | caused by RouteValidationError.Match: ["embedded message failed validation"] | caused by RouteMatchValidationError.Headers[i]: ["embedded message failed validation"] | caused by HeaderMatcherValidationError.SafeRegexMatch: ["embedded message failed validation"] | caused by RegexMatcherValidationError.Regex: ["value length must be at least " '\x01' " runes"]): name: "80"
```

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Steps to reproduce the bug**
Virtual Service psec:
```yaml
  spec:
    hosts:
    - s1.default.svc.cluster.local
    http:
    - match:
      - method:
          regex: ""
        name: 80_default
      route:
      - destination:
          host: s1.default.svc.cluster.local
```

**Version** (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version --short` if you used Helm)
1.9

PR #34066, opened 2021-07-15
Add empty regex validator

kebe7jun:
Fix https://github.com/istio/istio/issues/34065


kebe7jun:
Similar to https://github.com/istio/istio/pull/33786, should I use warnings instead?

howardjohn (approved):
I think this case of adding validation is ok because the previous configuration was 100% invalid and resulted in a broken cluster. It is in the case where a configuration is not recommended/ineffective or other non-critical issues where we should not add validation I think.

Thanks for the PR!

costinm (approved):

PR #34070, opened 2021-07-15
Add validation for service entry port.targetPort

ChenglongGao327:
1）if targetport is set, add validation for targetPort
2）move ports parameter validation to the same for loop

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pkg/config/validation/validation_test.go
+++ b/pkg/config/validation/validation_test.go
@@ -4317,6 +4317,37 @@ func TestValidateServiceEntries(t *testing.T) {
 			},
 			valid: false,
 		},
+		{
+			name: "repeat target port", in: networking.ServiceEntry{
+				Hosts:            []string{"google.com"},
+				WorkloadSelector: &networking.WorkloadSelector{Labels: map[string]string{"key": "bar"}},
+				Ports: []*networking.Port{
+					{Number: 80, Protocol: "http", Name: "http-valid1", TargetPort: 80},
+					{Number: 81, Protocol: "http", Name: "http-valid2", TargetPort: 80},
+				},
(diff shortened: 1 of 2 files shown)


pkg/config/validation/validation_test.go
```diff
@@ -4317,6 +4317,37 @@ func TestValidateServiceEntries(t *testing.T) {
 			},
 			valid: false,
 		},
+		{
+			name: "repeat target port", in: networking.ServiceEntry{
+				Hosts:            []string{"google.com"},
+				WorkloadSelector: &networking.WorkloadSelector{Labels: map[string]string{"key": "bar"}},
+				Ports: []*networking.Port{
+					{Number: 80, Protocol: "http", Name: "http-valid1", TargetPort: 80},
+					{Number: 81, Protocol: "http", Name: "http-valid2", TargetPort: 80},
```
howardjohn, line 4326: This should not be invalid. Its legal in Kubernetes, for example.

howardjohn (changes requested):
I am not sure this is right. Also, we should be careful about adding validation. Any configuration that was previously working on older versions must continue to work or upgrades will be blocked

pkg/config/validation/validation_test.go
ChenglongGao327: > This should not be invalid. Its legal in Kubernetes, for example.

I think too much, it should be consistent with k8s

howardjohn (approved):

PR #34073, opened 2021-07-15
fix spiffe possible memory leak

gy95:
only when the http Get operation is successful, `defer resp.Body.Close()` will be triggered.
And during the `for` loop, if timeout occurs, we will return directly without close `resp.Body`,
And this return may be triggered if the `resp.StatusCode != http.StatusOK`, and on this circumstance `resp.Body` is also not-nil, so we should release `Body` resource

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/spiffe/spiffe.go
+++ b/pkg/spiffe/spiffe.go
@@ -214,6 +214,11 @@ func RetrieveSpiffeBundleRootCerts(config map[string]string, caCertPool *x509.Ce
 		retryBackoffTime := firstRetryBackOffTime
 		startTime := time.Now()
 		var resp *http.Response
+		defer func() {
+			if resp != nil {
+				resp.Body.Close()
+			}
+		}()
 		for {
 			resp, err = httpClient.Get(endpoint)
 			var errMsg string
@@ -239,7 +244,6 @@ func RetrieveSpiffeBundleRootCerts(config map[string]string, caCertPool *x509.Ce
 			time.Sleep(retryBackoffTime)
 			retryBackoffTime *= 2 // Exponentially increase the retry backoff time.
 		}
-		defer resp.Body.Close()
 
 		doc := new(bundleDoc)
 		if err := json.NewDecoder(resp.Body).Decode(doc); err != nil {


pkg/spiffe/spiffe.go
```diff
@@ -231,6 +231,9 @@ func RetrieveSpiffeBundleRootCerts(config map[string]string, caCertPool *x509.Ce
 			}
 
 			if startTime.Add(retryTimeout).Before(time.Now()) {
+				if resp != nil && resp.Body != nil {
+					resp.Body.Close()
```
shankgan, line 235: If this is cause of the leak, can this be added to a defer statement somewhere, to prevent more leaks in the future?

shankgan: Ignore previous statement. This mem will probably be cleaned up by GC. Was any leak observed because of this?

gy95: no, i only find this, and we can modify as below to prevent more leaks.
we can remove `defer` statement to just after declaration of the variable `resp`, but with the judgement `resp != nil && resp.Body != nil` to prevent possible panic. And according to the standard lib comment, 
```
// The http Client and Transport guarantee that Body is always
// non-nil, even on responses without a body or responses with
// a zero-length body. 
```
So we can simplify this judgement to `resp != nil`
```
var resp *http.Response
defer func() {
	if resp != nil {
		resp.Body.Close()
	}
}()
```

gy95: And one more question, the `resp` variable is declared in the for-range loop https://github.com/istio/istio/blob/e2b2ce2758a94f19e0aa65dc1cc8166bacf4e74b/pkg/spiffe/spiffe.go#L196
So i'm not quite sure whether using `defer` in loop will lead to resource leaks

gy95: @shankgan how do you think about it? Could you please give me some advice?

shankgan: Based on your intent, I think the defer statement needs to be inside the loop, since ever call to Http.Get will yield a different *http.Response that (as per your analysis) will need to be cleaned up.

gy95: done, use defer just after the variable declaration

PR #34085, opened 2021-07-15
Fix crash in tpath

howardjohn:
I have no idea what this code does, so this may be a bad fix...


jacob-delgado:
cc @howardjohn do you want approval? Did you find this issue due to the fuzzer?

howardjohn:
Would be good for someone that understands this code to review it. This was from fuzzer

operator/pkg/tpath/tree.go
```diff
@@ -337,6 +337,10 @@ func setPathContext(nc *PathContext, value interface{}, merge bool) error {
 func setValueContext(nc *PathContext, value interface{}, merge bool) (bool, error) {
 	vv, mapFromString := tryToUnmarshalStringToYAML(value)
 
+	if nc.Parent == nil {
+		return false, nil
```
richardwxn, line 341: move to the beginning?

richardwxn (approved):
feel free to remove label and merge after addressing the comment

Issue #34097, opened 2021-07-15
Memory leak in xds_proxy


howardjohn:
https://github.com/istio/istio/blob/efef026fa83791b917d064c1ec91d4d95b9b1a32/pkg/istio-agent/xds_proxy.go#L391 is susceptible to a memory leak. This can occur when all reader are blocked, and `responsesChan` has hit capacity (10).

All readers can be blocked if the consumer is slow: https://github.com/istio/istio/blob/efef026fa83791b917d064c1ec91d4d95b9b1a32/pkg/istio-agent/xds_proxy.go#L455. This can be triggered by making the downstream.Send slow (reproduced in tests, with artificial delays), or by `requestsChan` being full and blocking.

This is far worse in 1.8, which can also be blocked by slow upstream.Send (https://github.com/istio/istio/blob/1.8.6/pkg/istio-agent/xds_proxy.go#L294). 

Every time the readers have exited (from conn.stopped closing), but we are still attempting to write on `responsesChan`, we will leak a goroutine+a channel with 10 discovery responses.

In real world (on 1.8, which again, is more susceptible to this), we saw roughly 200 occurrences of such a leak, leading to (presumably, no hard data for this) 2k DiscoveryResponses kept in memory, which took up 20gb of RAM and ultimately led to repeated OOMs.

mandarjog:
@kyessenov can take a look at this ?

PR #34098, opened 2021-07-15
Fix memory leak in xds_proxy when channels are blocked

howardjohn:
Fixes https://github.com/istio/istio/issues/34097

~I have some tests locally but its really hacky, will cleanup and add before submitting~ Tests added. This was a bit tricky to test, but the test fails without the PR and passes with it, so I think its all good.

### DIFF
--- a/pilot/test/xdstest/grpc.go
+++ b/pilot/test/xdstest/grpc.go
@@ -1,0 +1,74 @@
+package xdstest
+
+import (
+	"context"
+	"time"
+
+	"google.golang.org/grpc"
+
+	"istio.io/pkg/log"
+)
+
+func safeSleep(ctx context.Context, t time.Duration) {
+	select {
+	case <-time.After(t):
+	case <-ctx.Done():
+	}
+}
+
+type slowClientStream struct {
+	grpc.ClientStream
+	recv, send time.Duration
+}
+
+func (w *slowClientStream) RecvMsg(m interface{}) error {
+	if w.recv > 0 {
+		safeSleep(w.Context(), w.recv)
+		log.Infof("delayed recv for %v", w.recv)
+	}
+	return w.ClientStream.RecvMsg(m)
+}
+
+func (w *slowClientStream) SendMsg(m interface{}) error {
+	if w.send > 0 {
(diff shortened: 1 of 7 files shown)


lei-tang (approved):

lei-tang (commented):
Thanks for the PR! There are multiple test failures (unit test, lint test, etc). Please fix them.

pilot/test/xdstest/grpc.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+}
+
+type slowClientStream struct {
+	grpc.ClientStream
+	recv, send time.Duration
+}
+
+func (w *slowClientStream) RecvMsg(m interface{}) error {
+	if w.recv > 0 {
```
lei-tang, line 25: Where is w.recv updated? Does it need a mutex protection?

howardjohn: Its not mutated, just set once at construction

howardjohn:
Thanks, removed my logs

pkg/istio-agent/xds_proxy.go
```diff
@@ -282,6 +289,15 @@ type ProxyConnection struct {
 	upstreamDeltas     discovery.AggregatedDiscoveryService_DeltaAggregatedResourcesClient
 }
 
+// safeSendRequest is a small wrapper around sending to con.requestsChan. This ensures that we do not
+// block forever on
```
stevenctl, line 293: nit: "block forever on a full requestsChan"?

stevenctl (approved):

pkg/istio-agent/xds_proxy.go
```diff
@@ -282,6 +289,15 @@ type ProxyConnection struct {
...
+func (con *ProxyConnection) safeSendRequest(req *discovery.DiscoveryRequest) {
```
ramaraochavali, line 294: nit: We can just probably call this `sendRequest`

hzxuzhonghu:
/cherrypick release-1.10


It seems backport to 1.9 and 1.10 failed

hzxuzhonghu:
sorry, it is done manually

PR #34103, opened 2021-07-16
In the scene where k8s service select workload entry: set up istio endpoint port

ChenglongGao327:
First use the targetPort name, then use the service port name to look up in the portMap
both of them are not found then use targetPort port
This is consistent with serviceEntry

TODO：add test

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -856,27 +856,25 @@ func (c *Controller) serviceInstancesFromWorkloadInstances(svc *model.Service, r
 		if wi.Namespace != svc.Attributes.Namespace {
 			continue
 		}
-		if selector.SubsetOf(wi.Endpoint.Labels) {
-			// create an instance with endpoint whose service port name matches
-			istioEndpoint := *wi.Endpoint
-			if targetPortName != "" {
-				// This is a named port, find the corresponding port in the port map
-				matchedPort := wi.PortMap[targetPortName]
-				if matchedPort == 0 {
-					// No match found, skip this endpoint
-					continue
-				}
-				istioEndpoint.EndpointPort = matchedPort
-			} else {
-				istioEndpoint.EndpointPort = uint32(targetPort)
-			}
-			istioEndpoint.ServicePortName = servicePort.Name
-			out = append(out, &model.ServiceInstance{
-				Service:     svc,
-				ServicePort: servicePort,
-				Endpoint:    &istioEndpoint,
-			})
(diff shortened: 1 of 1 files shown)


ChenglongGao327:
https://github.com/istio/istio/pull/33774

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -856,27 +856,25 @@ func (c *Controller) serviceInstancesFromWorkloadInstances(svc *model.Service, r
...
-				Endpoint:    &istioEndpoint,
-			})
+		if !selector.SubsetOf(wi.Endpoint.Labels) {
+			continue
+		}
+		// create an instance with endpoint whose service port name matches
+		istioEndpoint := *wi.Endpoint
+		if matchedPort, ok := wi.PortMap[targetPortName]; ok && matchedPort > 0 {
+			istioEndpoint.EndpointPort = matchedPort
+		} else if matchedPort, ok := wi.PortMap[servicePort.Name]; ok && matchedPort > 0 {
```
hzxuzhonghu, line 866: I donot think service port name should match workloadentry port name

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -856,27 +856,25 @@ func (c *Controller) serviceInstancesFromWorkloadInstances(svc *model.Service, r
...
-				Service:     svc,
-				ServicePort: servicePort,
-				Endpoint:    &istioEndpoint,
-			})
+		if !selector.SubsetOf(wi.Endpoint.Labels) {
+			continue
+		}
+		// create an instance with endpoint whose service port name matches
+		istioEndpoint := *wi.Endpoint
+		if matchedPort, ok := wi.PortMap[targetPortName]; ok && matchedPort > 0 {
```
hzxuzhonghu, line 864: In the case `targetPortName` exist, but not found it in workloadentry, it should ignore this port, however with this patch, the result is unknown

pilot/pkg/serviceregistry/kube/controller/controller.go
ChenglongGao327: > I donot think service port name should match workloadentry port name

TargetPortName is rarely used by people, I think it can be used to define the port of a specific workloadEntry resource, just like when se.port.name matches we.port, we.port will be used for this workloadEntry resource rather than se.port

Issue #34084, opened 2021-07-15
Invalid cipherSuites lead to NACKs


howardjohn:
```
ADS:LDS: ACK ERROR istio-ingressgateway-588b68dbb6-l89ss.istio-system-8 Internal:Error adding/updating listener(s) 0.0.0.0_8443: Failed to initialize cipher suites fake. The following ciphers were rejected when tried individually: fake
```

```yaml
apiVersion: networking.istio.io/v1beta1
kind: Gateway
metadata:
  name: echo
  namespace: default
spec:
  selector:
    istio: ingressgateway
  servers:
  - hosts:
    - '*'
    port:
      name: https
      number: 443
      protocol: HTTPS
    tls:
      cipherSuites:
      - fake
      credentialName: istio-ingressgateway-certs
      mode: SIMPLE
```

Depending on order of the Gateway and secret creation, sometimes it is the SDS that NACKs, not LDS.

cc @lambdai @mandarjog

PR #34109, opened 2021-07-16
Add validation to prevent invalid cipher suites

howardjohn:
Fixes https://github.com/istio/istio/issues/34084

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/gateway.go
+++ b/pilot/pkg/networking/core/v1alpha3/gateway.go
@@ -43,6 +43,7 @@ import (
 	"istio.io/istio/pkg/config/gateway"
 	"istio.io/istio/pkg/config/host"
 	"istio.io/istio/pkg/config/protocol"
+	"istio.io/istio/pkg/config/security"
 	"istio.io/istio/pkg/proto"
 	"istio.io/istio/pkg/util/istiomultierror"
 	"istio.io/pkg/log"
@@ -637,11 +638,24 @@ func buildGatewayListenerTLSContext(
 		ctx.CommonTlsContext.TlsParams = &tls.TlsParameters{
 			TlsMinimumProtocolVersion: convertTLSProtocol(server.Tls.MinProtocolVersion),
 			TlsMaximumProtocolVersion: convertTLSProtocol(server.Tls.MaxProtocolVersion),
-			CipherSuites:              server.Tls.CipherSuites,
+			CipherSuites:              filteredCipherSuites(server.Tls.CipherSuites),
 		}
 	}
 
 	return ctx
+}
+
+// Invalid cipher suites lead Envoy to NACKing. This filters the list down to just the supported set.
+func filteredCipherSuites(suites []string) []string {
+	ret := make([]string, 0, len(suites))
+	for _, s := range suites {
+		if security.IsValidCipherSuite(s) {
+			ret = append(ret, s)
+		} else {
+			log.Debugf("ignoring unsupported cipherSuite: %q", s)
+		}
+	}
+	return ret
 }
 
 func convertTLSProtocol(in networking.ServerTLSSettings_TLSProtocol) tls.TlsParameters_TlsProtocol {
--- a/pkg/config/security/security.go
+++ b/pkg/config/security/security.go
@@ -20,9 +20,11 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"unicode"
 
 	"github.com/hashicorp/go-multierror"
 
+	"istio.io/istio/pilot/pkg/util/sets"
 	"istio.io/istio/pkg/config/host"
 )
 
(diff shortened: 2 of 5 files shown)


pkg/config/security/security.go
```diff
@@ -181,3 +182,11 @@ func validateMapKey(key string) error {
 	}
 	return fmt.Errorf("bad key (%s): should have format a[b]", key)
 }
+
+// validCipherSuites contains a list of all ciphers supported in Gateway.server.tls.cipherSuites
+// TODO: DO-NOT-MERGE: extract full supported cipher suite list
+var validCipherSuites = sets.NewSet("ECDHE-ECDSA-AES128-SHA")
```
lambdai, line 188: 1. "ALL" magical words. user should not use though.
2. `bssl ciphers -openssl-name ALL | rg -v PSK`
ECDHE-ECDSA-AES128-GCM-SHA256
ECDHE-RSA-AES128-GCM-SHA256
ECDHE-ECDSA-AES256-GCM-SHA384
ECDHE-RSA-AES256-GCM-SHA384
ECDHE-ECDSA-CHACHA20-POLY1305
ECDHE-RSA-CHACHA20-POLY1305
ECDHE-ECDSA-AES128-SHA
ECDHE-RSA-AES128-SHA
ECDHE-ECDSA-AES256-SHA
ECDHE-RSA-AES256-SHA
AES128-GCM-SHA256
AES256-GCM-SHA384
AES128-SHA
AES256-SHA
DES-CBC3-SHA

pkg/config/security/security.go
```diff
@@ -181,3 +182,11 @@ func validateMapKey(key string) error {
...
+
+func IsValidCipherSuite(cs string) bool {
+	return validCipherSuites.Contains(cs)
```
lambdai, line 191: Can you add that if the cs doesn't start with alphabet, log but let it pass the validation?
The idea is that these list is supposed to be a operator(e.g. exclude blabla) and operand.
I think we the validation is to figure out the obvious error but not block advanced users.

howardjohn: Done

pkg/config/security/security.go
howardjohn: Done ,thanks!

pkg/config/security/security.go
```diff
@@ -181,3 +183,35 @@ func validateMapKey(key string) error {
...
+	"AES128-SHA",
+	"AES256-SHA",
+	"DES-CBC3-SHA",
+)
+
+func IsValidCipherSuite(cs string) bool {
+	if cs == "" || cs == "ALL" {
+		return true
+	}
+	if unicode.IsSymbol(rune(cs[0])) {
```
lambdai, line 211: Great. Thanks!

lambdai (approved):
LGTM. I have a poor knowledge on openssl. 

This covers all the easy use cases that I am aware of.

ramaraochavali:
Why do not we add this in gateway validation instead of at listener creation? Since this is an invalid config - I think it is OK to add that new validation as any one would have used it like that?

howardjohn:
We add warning at validation because if we make it hard reject, then some
configurations that are previously valid will suddenly stop working when
the  upgrade. Kubernetes API guidance suggests to not make validation
stricter without a new api version

ramaraochavali:
> We add warning at validation because if we make it hard reject, then some
configurations that are previously valid will suddenly stop working when
the  upgrade. 

But this config would have never worked right? Envoy would have rejected it any way so people might not have used it like this. So was thinking it would be good to enforce this. But thinking about this more, when we have mutiple servers in gateway and one is bad, we should not reject all now.

ramaraochavali:
Another idea, how about doing it in 'MergeGateways' and drop that server?

howardjohn:
> Another idea, how about doing it in 'MergeGateways' and drop that server?

Dropping the whole server seems not great, since they may just add a cipher that is invalid and suddenly it breaks? Especially since we only warn about it.

Otherwise the UX right now would be they add a bad cipher, get a warning, and then update the yaml and re-apply. But during that time they have an outage.


> But this config would have never worked right? 

Its actually a bit trickier. The cipher suite rejection logic is not always enacted when we push LDS. If the cert is loaded latter, we may push LDS successfully without any NACK, then later the cert is added and it starts breaking.

ramaraochavali:
> Its actually a bit trickier. The cipher suite rejection logic is not always enacted when we push LDS. If the cert is loaded latter, we may push LDS successfully without any NACK, then later the cert is added and it starts breaking.

Got it. 

My only concern with this is we are doing it every time listener creation happens. But not a huge issue - would have been good if it is done once.

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -637,13 +638,26 @@ func buildGatewayListenerTLSContext(
...
 }
 
+// Invalid cipher suites lead Envoy to NACKing. This filters the list down to just the supported set.
+func filteredCipherSuites(suites []string) []string {
+	ret := make([]string, 0, len(suites))
+	for _, s := range suites {
+		if security.IsValidCipherSuite(s) {
+			ret = append(ret, s)
+		} else {
+			log.Debugf("ignoring unsupported cipherSuite: %q", s)
```
stevenctl, line 655: nit: including the server name may make it easier to find the offending config

pkg/config/security/security.go
```diff
@@ -181,3 +183,35 @@ func validateMapKey(key string) error {
...
+)
+
+func IsValidCipherSuite(cs string) bool {
+	if cs == "" || cs == "ALL" {
+		return true
+	}
+	if !unicode.IsNumber(rune(cs[0])) && !unicode.IsLetter(rune(cs[0])) {
+		// Not all of these are correct, but this is needed to support advanced cases like - and + operators
+		// without needing to parse the full expression
+		return true
```
stevenctl, line 214: Would `return IsValidCipherSuite(cs[1:])` be valid? Still doesn't catch all cases but it's closer.

stevenctl (approved):

pkg/config/security/security.go
```diff
@@ -181,3 +183,35 @@ func validateMapKey(key string) error {
...
+	"ECDHE-RSA-CHACHA20-POLY1305",
+	"ECDHE-ECDSA-AES128-SHA",
+	"ECDHE-RSA-AES128-SHA",
+	"ECDHE-ECDSA-AES256-SHA",
+	"ECDHE-RSA-AES256-SHA",
+	"AES128-GCM-SHA256",
+	"AES256-GCM-SHA384",
+	"AES128-SHA",
+	"AES256-SHA",
+	"DES-CBC3-SHA",
```
nibahl, line 204: Hi @howardjohn, we were trying to enable some of the cipher suites in our Istio config but we face 2 issues. 
1. Even after specifying these values in our gateway.yaml file, we do not see some of them when we run the scanner on our endpoint here for TLS 1.2: https://www.ssllabs.com/ssltest/
2. Is there a way to specify the cipher suites for TLS 1.3? We do see some of the cipher suites show up on the above scanner even after not specifying them in our config. Some of them are not allowed as per our company policy. Is there a way to fix this?

howardjohn: 1. Hard to say without seeing the configdump, logs, etc
2. boringssl doesn't allow custom cipher suites in TLS 1.3. So if you want this in Istio you would need to change boringssl.

PR #34112, opened 2021-07-16
enable mTLS from proxyless gRPC clients

stevenctl:
* Set the experimental flag
* Build proper transport credential
* Cleanup CDS
* Cleanup bootstrap:
  * correct plugin name
  * properly marshal durationpb  

---

Tested with:

```
kind: PeerAuthentication
metadata:
  name: "default"
spec:
  mtls:
    mode: STRICT
```

Before, we'd not even attempt TLS and the request would fail. This change gets me a successful request to a workload running envoy, with TLS. 

Screen Shot 2021-07-16 at 12 18 46 PM

---

Planning to add proxyless to the security reachabilty suite in a follow up

### DIFF
--- a/pkg/istio-agent/grpcxds/grpc_bootstrap.go
+++ b/pkg/istio-agent/grpcxds/grpc_bootstrap.go
@@ -17,13 +17,13 @@ package grpcxds
 import (
 	"encoding/json"
 	"fmt"
-	"google.golang.org/protobuf/encoding/protojson"
 	"io/ioutil"
 	"os"
 	"path"
 	"time"
 
 	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/structpb"
 
@@ -148,7 +148,7 @@ func GenerateBootstrap(opts GenerateBootstrapOptions) (*Bootstrap, error) {
 					PrivateKeyFile:    path.Join(opts.CertDir, "key.pem"),
 					CertificateFile:   path.Join(opts.CertDir, "cert-chain.pem"),
 					CACertificateFile: path.Join(opts.CertDir, "root-cert.pem"),
-					RefreshDuration: refresh,
+					RefreshDuration:   refresh,
 				},
 			},
 		}
--- a/pkg/istio-agent/grpcxds/grpc_bootstrap.go
+++ b/pkg/istio-agent/grpcxds/grpc_bootstrap.go
@@ -17,6 +17,7 @@ package grpcxds
 import (
 	"encoding/json"
 	"fmt"
+	"google.golang.org/protobuf/encoding/protojson"
 	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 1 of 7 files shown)


howardjohn:
How does auto mtls work to grpc? I assume we cannot tell if it will accept TLS or not since *grpc ignores the TLS configuration sent over XDS*

pkg/istio-agent/grpcxds/grpc_bootstrap.go
```diff
@@ -51,17 +52,17 @@ type XdsServer struct {
...
 
 type FileWatcherCertProviderConfig struct {
-	CertificateFile   string               `json:"certificate_file,omitempty"`
-	PrivateKeyFile    string               `json:"private_key_file,omitempty"`
-	CACertificateFile string               `json:"ca_certificate_file,omitempty"`
-	RefreshDuration   *durationpb.Duration `json:"refresh_interval,omitempty"`
+	CertificateFile   string          `json:"certificate_file,omitempty"`
+	PrivateKeyFile    string          `json:"private_key_file,omitempty"`
+	CACertificateFile string          `json:"ca_certificate_file,omitempty"`
+	RefreshDuration   json.RawMessage `json:"refresh_interval,omitempty"`
```
howardjohn, line 65: does just time.Duration work?

howardjohn (approved):

pkg/istio-agent/grpcxds/grpc_bootstrap.go
stevenctl: Possibly? I rewrote this to match their unmarshaling

howardjohn: I think it would but matching there's is better

stevenctl:
> grpc ignores the TLS configuration sent over XDS

Do you mean for gRPC servers? That has to be dealt with separately, and it also requires application code changes to use https://pkg.go.dev/google.golang.org/grpc/xds#NewGRPCServer.

Clients won't ignore it if the flag is set and right client code is used, and without the peerAuthentication this change causes failures, not sure why exactly; I figured even without STRICT the server-side envoy would still perform TLS. 

```
WARNING: 2021/07/17 17:01:41 [core] grpc: addrConn.createTransport failed to connect to {10.120.1.14:17070 a.default.svc.cluster.local:7070 0xc00496a550 0 <nil>}. Err: connection error: desc = "transport: authentication handshake failed: read tcp 10.120.1.17:48262->10.120.1.14:17070: read: connection reset by peer". Reconnecting...
```

If you mean how do we know if the client will actually use the TLS configuration we send it... that I'm not sure. If its client-side, things will probably just break for the most part. For servers ignoring TLS config we should definitely find a way to raise a warning. 

I'll update this [doc](https://docs.google.com/document/d/1UG-s2mG-Sjt9_OzTLFPc-WJiZ09ins48kS5lnMD2V6o/edit?usp=sharing) as I figure more things out.

howardjohn:
> Do you mean for gRPC servers? 

What I mean is in k8s, we know if the server will accept mTLS traffic by logic which is basically `hasSidecar && !PeerAuth=DISABLE`. This works in almost all cases (one case it does not is multiprimary - we don't know which PeerAuth is used).

For grpc its a bit different. First, we do not have an equivalent `hasSidecar` check we can do. You could kind of do one from injection but not really a good idea. Even if they are using XDS, they may still not have the tls config set in their application, so the presences of PeerAuth is not sufficient as they may just ignore it.

So I am not sure we have a good way to implement this core functionality

stevenctl:
By defaulting everything to plaintext, and only claiming support for STRICT mode, if either side is miconfigured requests will break rather than a user expecting mTLS but actually doing plaintext... ? Agree we don't have a good way to implement this.

Issue #33809, opened 2021-07-02
Make auto mTLS work with workload selector peer authn


incfly:
**Describe the feature request**

Thanks to https://github.com/envoyproxy/envoy/issues/16536 being resolved, now we can revisit the known auto mTLS limitation around the workload peer authn. This used to be due to the fact that workload selector and upstream cluster mapping are not 1:1. Plus updating the endpoint label can't trigger data plane transport socket config updates.

But now given that limitation is resolved in Envoy, we should try to fix this, therefore users never need to set DetinationRule regardles what peer authentication policy is, even for [workload level peer authn](https://istio.io/latest/docs/tasks/security/authentication/authn-policy/#enable-mutual-tls-per-workload). This can be achieved by:

1. We calculate the endpoints label, `tlsMode: istio` only when the endpoint has the sidecar & the peer authn policy says server side enables permissive or strict.
1. If a PeerAuthn in namespace `foo` is changed/added/deleted, we selectively pushed all the EDS belonging to the namespace `foo`.

Note that we still have to push all the EDS in namespace `foo`, this is because for a peer authn `selector: {env: prod}`, we don't know what set of the services in namespace foo will be affected. (unless there're some new optimizations in istiod that I'm not aware of).

**Describe alternatives you've considered**

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[X ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

incfly:
I will start a prototype recently. @howardjohn @hzxuzhonghu FYI Let me know what you think. thanks.

howardjohn:
I have a prototype that I *think* was only missing the envoy PR. Not at my computer now to find the branch name but I can find it next week and send it. Happy to have someone take it over if anyone has time

hzxuzhonghu:
>If a PeerAuthn in namespace foo is changed/added/deleted, we selectively pushed all the EDS belonging to the namespace foo.

You don't need to consider this, it has already been this.

PR #34118, opened 2021-07-17
Make auto mTLS works with workload level peer authentication policy.

### LINKED ISSUE #32788: auto mtls: add support for workload selectors on PeerAuthentication
Currently, auto mtls only works for namespace or cluster scoped PeerAuthentications. This makes creating a DISABLE PeerAuthentication for a specific workload really hard, since now you need a DR which is confusing.

cc @incfly

incfly:
Fixes #32788  #33809

- Creates the mTLS checker in the endpoint builder when the feature enabled.
- When policy enables mTLS plus the workload has the sidecar, we add metadata tlsMode: istio.
- Feature guarded via an env var. Integration test covered.

credits to @cpakulski for the work on envoy side


pilot/pkg/xds/endpoint_builder.go
```diff
@@ -315,9 +323,16 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
...
 				localityEpMap[ep.Locality.Label] = locLbEps
 			}
-			if ep.EnvoyEndpoint == nil {
-				ep.EnvoyEndpoint = buildEnvoyLbEndpoint(ep)
+			// TODO(incfly): check with reviewer to see if we should only touches the updated metadata field
+			// or rebuild from scratch (better readability).
+			// if ep.EnvoyEndpoint == nil {
+			tlsMode := ep.TLSMode
+			if b.mtlsChecker != nil && b.mtlsChecker.mtlsDisabledByPeerAuthentication(ep) {
+				log.Infof("incfly pocliy disabled proxy %v, true or false %v",
```
yangminzhu, line 331: update the log?

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -315,9 +323,16 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
...
+					ep.Address, b.mtlsChecker.mtlsDisabledByPeerAuthentication(ep))
+				tlsMode = ""
```
yangminzhu, line 333: just curious, what's the difference between `""` and `model.DisabledTLSModeLabel` for the `tlsMode`?

tests/integration/security/testdata/beta-mtls-automtls-workload.yaml
```diff
@@ -0,0 +1,42 @@
...
+  annotations:
+    test-suite: "beta-mtls-automtls-workload"
+spec:
+  selector:
+    matchLabels:
+      app: c
+  mtls:
+    mode: STRICT
+  portLevelMtls:
+    # 8090 is the targetPort for service http(80)
```
yangminzhu, line 40: 8090 or 8080?

tests/integration/security/reachability_test.go
```diff
@@ -89,6 +89,26 @@ func TestReachability(t *testing.T) {
...
+						// Sidecar injected client always succeed.
+						if apps.D.Contains(src) {
+							return true
+						}
+						// For naked app as client, only requests targeted to mTLS disabled endpoints succeed:
+						// A are disabled by workload selector for entire service.
+						// C port 8090 http port are disabled.
+						return apps.A.Contains(opts.Target) || (apps.C.Contains(opts.Target) && opts.PortName == "http")
+					},
+					ExpectMTLS:             Never,
```
yangminzhu, line 109: should the `ExpectMTLS` be used for the mTLS cases?

yangminzhu (commented):
should also include a release note.

pilot/pkg/features/pilot.go
```diff
@@ -146,6 +146,11 @@ var (
 		false,
 		"Skip validating the peer is from the same trust domain when mTLS is enabled in authentication policy").Get()
 
+	EnableAutomTLSCheckPolicies = env.RegisterBoolVar(
+		"ENABLE_AUTO_MTLS_CHECK_POLICIES", false,
```
hzxuzhonghu, line 150: shouldn't it always be true?

pkg/test/framework/components/echo/annotations.go
```diff
@@ -36,17 +36,18 @@ type Annotation struct {
...
+	SidecarBootstrapOverride            = workloadAnnotation(annotation.SidecarBootstrapOverride.Name, "")
+	SidecarVolumeMount                  = workloadAnnotation(annotation.SidecarUserVolumeMount.Name, "")
+	SidecarVolume                       = workloadAnnotation(annotation.SidecarUserVolume.Name, "")
+	SidecarConfig                       = workloadAnnotation(annotation.ProxyConfig.Name, "")
+	SidecarInterceptionMode             = workloadAnnotation(annotation.SidecarInterceptionMode.Name, "REDIRECT")
+	SidecarIncludeInboundPorts          = workloadAnnotation(annotation.SidecarTrafficIncludeInboundPorts.Name, "")
+	SidecarIncludeOutboundIPRanges      = workloadAnnotation(annotation.SidecarTrafficIncludeOutboundIPRanges.Name, "")
+	SidecarProxyConfig                  = workloadAnnotation(annotation.ProxyConfig.Name, "")
+	SidecarInjectTemplates              = workloadAnnotation(inject.TemplatesAnnotation, "")
+	WorkloadEnableAutoMtlsWithPeerAuthn = workloadAnnotation("security.istio.io/autoMtlsWithPeerAuthn", "true")
```
howardjohn, line 50: what is this for?

tests/integration/security/testdata/beta-mtls-automtls-workload.yaml
incfly: 8090 is the http port, would just works fine, right?
https://github.com/istio/istio/blob/08f3ac1e1b7b8204dcc8fba2c0e0da5f6039536a/tests/integration/security/util/framework.go#L82

tests/integration/security/reachability_test.go
incfly: done. use mtlsOnExpect.

pkg/test/framework/components/echo/annotations.go
incfly: removed. i originally want to add support for per workload config option,but later find that might be not necessary and cause some trouble instead.

pilot/pkg/xds/endpoint_builder.go
incfly: done.

pilot/pkg/xds/endpoint_builder.go
incfly: it seems like only used in workload entry and service entry. not much difference in terms of the end effect AFAIK.

pilot/pkg/features/pilot.go
incfly: done.

tests/integration/security/testdata/beta-mtls-automtls-workload.yaml
yangminzhu: oh I thought this is commenting for the port number 8080 in line 41, maybe update the comment to make it clear

incfly: good point. updated to make the same.

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -315,9 +323,11 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
 				}
 				localityEpMap[ep.Locality.Label] = locLbEps
 			}
-			if ep.EnvoyEndpoint == nil {
-				ep.EnvoyEndpoint = buildEnvoyLbEndpoint(ep)
+			tlsMode := ep.TLSMode
+			if b.mtlsChecker != nil && b.mtlsChecker.mtlsDisabledByPeerAuthentication(ep) {
+				tlsMode = ""
 			}
+			ep.EnvoyEndpoint = buildEnvoyLbEndpoint(ep, tlsMode)
```
howardjohn, line 330: should check the impact of this change (ie running benchmark, which I triggered)

howardjohn (commented):
LGTM

pilot/pkg/xds/endpoint_builder.go
incfly: sure. anything else you would suggest besides the benchmark results?

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -315,9 +323,11 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
 				}
 				localityEpMap[ep.Locality.Label] = locLbEps
 			}
-			if ep.EnvoyEndpoint == nil {
-				ep.EnvoyEndpoint = buildEnvoyLbEndpoint(ep)
+			tlsMode := ep.TLSMode
+			if b.mtlsChecker != nil && b.mtlsChecker.mtlsDisabledByPeerAuthentication(ep) {
+				tlsMode = ""
```
stevenctl, line 328: Does this also make sense to do with the client settings from DR?

stevenctl (approved):
lgtm! added hold pending benchmark results or other reviewers

pilot/pkg/xds/endpoint_builder.go
howardjohn: If we do have client DR we shouldn't use auto mtls so it probably doesn't matter?

incfly: you mean when the client side DR disable the mTLS? I think in that case the cds.go code would already configure the cluster as plaintext, no `transport_socket_matches` at all.

pilot/pkg/xds/endpoint_builder.go
howardjohn:
```
EndpointGeneration/1/100-8                           376kB ± 0%         536kB ± 0%  +42.52%  (p=0.016 n=5+4)
EndpointGeneration/10/10-8                           246kB ± 0%         364kB ± 0%  +48.44%  (p=0.008 n=5+5)
EndpointGeneration/100/10-8                         2.33MB ± 0%        3.47MB ± 0%  +48.80%  (p=0.008 n=5+5)
EndpointGeneration/1000/1-8                         2.34MB ± 0%        3.47MB ± 0%  +48.53%  (p=0.008 n=5+5)
```

Not great... do you think its feasible to rebuild only when tlsMode changes (which is very rare)?

incfly:
The benchmark result is now available. Here are some cpu and mem delta for the EndpointGeneration. Noticed since there's no policy change in the test code  AFAIS, https://github.com/istio/istio/blob/39e6e3922878f292e89c8267aaa3c64cff3f6d38/pilot/pkg/xds/bench_test.go#L299. So these delta would only be the baseline delta cost for just checking the server endpoint mtls status, I think. The potential optimization, to only update metadata when the policy changed, would not change the numbers here.

not sure the criteria for such change and data. WDTY? @howardjohn @stevenctl 

```
EndpointGeneration/1000/1-8                         2.34MB ± 0%        3.47MB ± 0%  +48.53%  (p=0.008 n=5+5)
EndpointGeneration/1000/1-8                         10.2ms ± 2%        12.4ms ± 3%  +21.15%  (p=0.008 n=5+5)

```

pilot/pkg/xds/endpoint_builder.go
incfly: We definitely can do that(only rebuild when tls Mode changes), but that might not be that rare since now, `tlsMode` value can change caused by a peer authn config changed, right? See my previous comments.

Current code is building policy checker for each endpoint, which underneaths creates the PolicyApplier. I think reuse a single one for a single PushContext would reduce the memory significantly. From config perpective i think this would work as well since the policy applier is not endpoint specific.

howardjohn: I think it is rare - we push endpoints FAR more often then peer auth changes. Likely many many orders of magnitude more often.

For PolicyApplier - doesn't it take `ep.Labels` into account?

incfly: > I think it is rare in relationship to how often we hit this code path. We push endpoints FAR more often then peer auth changes. Likely many many orders of magnitude more often.

I see. you mean like when a single pod comes up we hit this code(rebuild the entire endpoint list for a given cluster maybe) which is far more often than the peer authn changes. In that case, if we only add these cost(cpu/memory) when peer authn changes, that could be acceptable.

true, the applier indeed take the labels into account, though we can build another map from labels to the applier instance to de-dup among the same set of endpoints.

incfly: did some profilng againt the benchmark test. it appears that [1] the delta inc for the memory indeed comes the `buildEnvoyEndpoint`. conditionally construct the endpoint solves the problem. benchmark numbers should go back to the same level before.

image

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -258,6 +266,19 @@ func (e *LocLbEndpointsAndOptions) AssertInvarianceInTest() {
...
+	if e == nil {
+		return true
+	}
+	epTLSMode := ""
+	if e.Metadata != nil && e.Metadata.FilterMetadata == nil {
+		if v, ok := e.Metadata.FilterMetadata[util.EnvoyTransportSocketMetadataKey]; ok {
+			epTLSMode = v.Fields[model.TLSModeLabelShortname].GetStringValue()
+		}
+	}
+	return tlsMode == epTLSMode
```
hzxuzhonghu, line 279: why not construct when tlsMode != epTLSMode ?

pilot/pkg/xds/endpoint_builder.go
hzxuzhonghu: I think we need to unify "" and DisabledTLSModeLabel, maybe later.

hzxuzhonghu (approved):

pilot/pkg/xds/endpoint_builder.go
incfly: now fixed. see benchmark test results.

above comment is not enough. we still need to ensure when tlsMode == "", do not set filter metadata, just leave empty values. Because later on EndpointByNetworkFilters would use `proto.Clone`, increase the consumption as well.

incfly:
Can you take another look? The benchmark data now looks okay to me. https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/34118/benchmark_istio/1417780463652573184/build-log.txt

howardjohn (approved):

stevenctl:
@incfly does this need to be in 1.11?

howardjohn:
Do not merge to 1.11: see https://github.com/istio/istio/issues/34227

PR #34119, opened 2021-07-17
fix the way of getting url host

gy95:
1. It's not correct to use function `strings.Trim` to remove prefix `"http://"` of URL, using `url/net` to `Parse` URL to get host
   according to go standard lib, `Trim` is used for 
```
// Trim returns a slice of the string s with all leading and
// trailing Unicode code points contained in cutset removed.
```
2. fix some non-standard comments

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/security/security.go
+++ b/pkg/security/security.go
@@ -51,8 +51,10 @@ const (
 	// TODO: change all the pilot one reference definition here instead.
 	WorkloadKeyCertResourceName = "default"
 
-	// Credential fetcher type
-	GCE  = "GoogleComputeEngine"
+	// GCE is Credential fetcher type of Google plugin
+	GCE = "GoogleComputeEngine"
+
+	// Mock is Credential fetcher type of mock plugin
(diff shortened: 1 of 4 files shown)


morvencao (commented):
why don't we use `net.URL` type to get the host?

gy95:
> why don't we use `net.URL` type to get the host?

you mean using package `net/url` ?
```
u, _ := url.Parse(httpServer.URL)
if err := os.Setenv("GCE_METADATA_HOST", u.Host); err != nil
...
```

morvencao:
yes, does it make sense?

gy95:
> yes, does it make sense? @gy95

of course, done

morvencao (approved):

howardjohn (approved):
This is just test code so not a big concern...

PR #34120, opened 2021-07-17
modify some codes of k8s/chiron

gy95:
1. modify some comments
2. There is no need to using `defer` func to Close Connection, just Close directly
3. fix some logs

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/security/pkg/k8s/chiron/controller.go
+++ b/security/pkg/k8s/chiron/controller.go
@@ -39,14 +39,12 @@ import (
 	"istio.io/pkg/log"
 )
 
-/* #nosec: disable gas linter */
 const (
-	// The Istio DNS secret annotation type
+	// IstioDNSSecretType is the Istio DNS secret annotation type
 	IstioDNSSecretType = "istio.io/dns-key-and-cert"
 
 	// For debugging, set the resync period to be a shorter period.
 	secretResyncPeriod = 10 * time.Second
-	// secretResyncPeriod = time.Minute
 
 	recommendedMinGracePeriodRatio = 0.2
 	recommendedMaxGracePeriodRatio = 0.8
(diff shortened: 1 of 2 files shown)


security/pkg/k8s/chiron/controller.go
```diff
@@ -136,7 +134,7 @@ func NewWebhookController(gracePeriodRatio float32, minGracePeriod time.Duration
 	if err != nil {
 		return nil, err
 	}
-	if len(dnsNames) == 0 {
+	if len(serviceNamespaces) == 0 {
```
gy95, line 137: Although `len(dnsNames)` is equal to `len(serviceNamespaces)`, I think it's better to use `len(serviceNamespaces)`. Because if length is not 0,  we need to build `namespace` `List-Watch`

security/pkg/k8s/chiron/utils.go
```diff
@@ -232,7 +232,7 @@ func submitCSR(certClient certclient.CertificateSigningRequestInterface,
 	var reqRet *cert.CertificateSigningRequest
 	var errRet error
 	for i := 0; i < numRetries; i++ {
-		log.Debugf("trial %v to create CSR (%v)", i, csrName)
+		log.Debugf("trial %v to create CSR (%v)", i+1, csrName)
```
gy95, line 235: trail times(1 ~ numRetries) is `i+1`(0 ~ numRetries-1)

security/pkg/k8s/chiron/utils.go
```diff
@@ -195,7 +195,7 @@ func isTCPReachable(host string, port int) bool {
 		// No connection yet, so no need to conn.Close()
 		return false
 	}
-	defer conn.Close()
+	conn.Close()
```
gy95, line 198: No need to use defer to Close connection

howardjohn (approved):

Issue #31745, opened 2021-03-28
gateway-apis: HTTPRoute with 0 weight still forward the request to backend


nak3:
**Bug description**

- Istio still forwards the request to backend even when HTTPRoute sets `weight: 0`.

[X] Networking

**Expected behavior**

- `weight: 0` should not forward the request.

**Steps to reproduce the bug**

#### 1. Prepare k8s cluster with non-default cluster domain
<details>
<summary>2. Install Istio</summary>

For example, I am using the following manifest.
```
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  values:
    global:
      proxy:
        autoInject: disabled
        clusterDomain: cluster.local
      useMCP: false
      # The third-party-jwt is not enabled on all k8s.
      # See: https://istio.io/docs/ops/best-practices/security/#configure-third-party-service-account-tokens
      jwtPolicy: first-party-jwt
    pilot:
      autoscaleMin: 1
      autoscaleMax: 3
      cpu:
        targetAverageUtilization: 60
      env:
        PILOT_ENABLED_SERVICE_APIS: true

    gateways:
      istio-ingressgateway:
        autoscaleMin: 1
        autoscaleMax: 3
        type: NodePort

  addonComponents:
    pilot:
      enabled: true

  components:
    ingressGateways:
      - name: istio-ingressgateway
        enabled: true
        k8s:
          resources:
            requests:
              cpu: 100m
              memory: 256Mi
```

</details>

<details>
<summary>3. deploy Gateway,GatewayClass,HTTPRoute (with weight=0) ,Deployments</summary>

```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Namespace
metadata:
  name: gateway-demo
---
apiVersion: networking.x-k8s.io/v1alpha1
kind: GatewayClass
metadata:
  name: istio
spec:
  controller: istio.io/gateway-controller
---
kind: Gateway
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: simple-gateway
  namespace: gateway-demo
spec:
  gatewayClassName: istio
  listeners:  # Use GatewayClass defaults for listener definition.
  - protocol: HTTP
    port: 80
    routes:
      kind: HTTPRoute
      namespaces:
        from: "All"
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: foo
  namespace: gateway-demo
spec:
  replicas: 1
  selector:
    matchLabels:
      app: foo
  template:
[...]

PR #34129, opened 2021-07-18
Do not forward the traffic to that backend when weight is 0

nak3:
As per https://github.com/kubernetes-sigs/gateway-api/issues/596,
the traffic should not be forwarded when weight is 0.

This patch updates it.

Fix https://github.com/istio/istio/issues/31745

[X] Networking


howardjohn (commented):
I think the API specifies that we must process the requests, but not forward them. I think this prevents processing the request entirely?

nak3:
Thank you. I missed it.
But I have no idea to handle the zero weight traffic except for using `falutInjection` :sweat_smile: Updated the PR with falutInjection for now.

howardjohn (approved):

PR #34114, opened 2021-07-16
Fix non-deterministic behavior in multi-network code

nmittler:
Discovered while investigating #34051.

Enabling [UNSAFE_PILOT_ENABLE_RUNTIME_ASSERTIONS](https://github.com/istio/istio/blob/dbf6a9e1c123bf6abbab45683b1c6c1f7660916c/pilot/pkg/features/pilot.go#L461)
Revealed that recent changes to managing multi-network gateways caused issues with EDS caching.

This PR fixes a couple of things:

- Kube controller now correctly identifies when network gateways have changed
- EDS filtering logic for split horizon now generates deterministic results

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #34131, opened 2021-07-19
Fix network gateway change judge

hzxuzhonghu:
The `gatewaysChanged` should be opposite


pilot/pkg/serviceregistry/kube/controller/network.go
```diff
@@ -204,7 +204,7 @@ func (c *Controller) extractGatewaysInner(svc *model.Service) bool {
 	}
 
 	previousGateways := c.networkGateways[svc.Hostname][nw]
-	gatewaysChanged := newGateways.equals(previousGateways)
+	gatewaysChanged := !newGateways.equals(previousGateways)
```
ramaraochavali, line 207: This is also fixed in https://github.com/istio/istio/pull/34114

PR #34137, opened 2021-07-19
generate virtual host domains based on virtual service hosts

ramaraochavali:
This is one of the issues we have identified in https://github.com/istio/istio/pull/33904. Since that is getting delayed, I just push this PR to generate domains based on virtual service hosts.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/httproute.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute.go
@@ -215,7 +215,7 @@ func BuildSidecarOutboundVirtualHosts(node *model.Proxy, push *model.PushContext
 		if listenerPort == 0 {
 			// Take all ports when listen port is 0 (http_proxy or uds)
 			// Expect virtualServices to resolve to right port
-			servicesByName[svc.Hostname] = svc
+			servicesByName[svc.Hostname] = svc.DeepCopy()
 		} else if svcPort, exists := svc.Ports.GetByPort(listenerPort); exists {
 			servicesByName[svc.Hostname] = &model.Service{
 				Hostname:     svc.Hostname,
--- a/pilot/pkg/networking/core/v1alpha3/httproute_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute_test.go
@@ -308,14 +308,16 @@ func TestSidecarOutboundHTTPRouteConfigWithWildcardsInVirtualService(t *testing.
 		t.Fatalf("got nil route for %s", routeName)
 	}
 	expectedHosts := map[string][]string{
-		"allow_any":     {"*"},
-		"*.com:7442":    {"*.com", "*.com:7442"},
-		"*.global:7442": {"*.global", "*.global:7442"},
+		"allow_any":         {"*"},
+		"*.com:7442":        {"*.com", "*.com:7442"},
+		"*.foo.global:7442": {"*.foo.global", "*.foo.global:7442"},
+		"*.global:7442":     {"*.global", "*.global:7442"},
 	}
 	expectedDestination := map[string]string{
-		"allow_any":     "PassthroughCluster",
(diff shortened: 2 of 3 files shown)


howardjohn:
Can you explain more why we are doing this? I can see the XDS changes:
```
        map[*.global:7442:[*.global *.global:7442] allow_any:[*] test-service-one.com:7442:[test-service-one.com test-service-one.com:7442] test-service-two.com:7442:[test-service-two.com test-service-two.com:7442]], wanted
        map[*.com:7442:[*.com *.com:7442] *.foo.global:7442:[*.foo.global *.foo.global:7442] *.global:7442:[*.global *.global:7442] allow_any:[*]]
```

But not sure what the positive and negative impacts are of this change.

ramaraochavali:
> Can you explain more why we are doing this? I can see the XDS changes:

I think it is because when some one specifies virtual services should not we generate domains based on configured host? Right now the domains are random based on what service it matches.

howardjohn:
> I think it is because when some one specifies virtual services should not we generate domains based on configured host? Right now the domains are random based on what service it matches.

Ah I think I get it. So if there are no services matching, we put `*.example.com`. But then suddenly if someone adds a service `foo.example.com`, we start matching only foo.example.com, which is unexpected.

The flipside is that now a user can create `*.cluster.local` (or with your other PR, `*`), and capture ALL traffic - with higher precedence than explicitly FQDN. That is extremely dangerous, and a major breaking API change.

So I can see some benefits of this, but I think we need to be more careful here around security and backwards compatibility.

howardjohn (changes requested):
Also seeing a bug here:

```
  - domains:
    - '*.cluster.local'
    - '*.cluster.local:7070'
    - 10.96.116.22
    - 10.96.116.22:7070
```

from VS:
```
kind: VirtualService
metadata:
  name: route
spec:
  hosts:
  - "*.cluster.local"
  http:
  - route:
    - destination:
        host: echo
```

So its grabbing some random matching cluster IP

ramaraochavali:
> The flipside is that now a user can create *.cluster.local (or with your other PR, *), and capture ALL traffic - with higher precedence than explicitly FQDN. That is extremely dangerous, and a major breaking API change.

I will close my other PR because it does not make sense to allow `*` after thinking about it for some more time. But my question about why do we allow for gateways still remains - I think we should add a warning in validation for sure. If you are OK, I can repurpose that PR to add that warning.

> with higher precedence than explicitly FQDN. 
I guess this in general a problem with wildcard hosts in virtual service. But since we allow it now we must add some guardrails if we want to prevent that  - because I think we have never said in API that domains would be created with first matching host.
People expect domains to be same as what they specify in VS hosts for wildcards also.

> That is extremely dangerous, and a major breaking API change.

I agree it is dangerous. What API expectation this breaks? I am not fully following that part?

howardjohn:
> But my question about why do we allow for gateways still remains - I think we should add a warning in validation for sure. If you are OK, I can repurpose that PR to add that warning.

Because with gateway everything is opt-in/explicitly configured. So if I configure example.com, and you configure `*.com`, example.com takes precedence.

For sidecar, because if the way we setup hostnames for each Service automatically, the semantics get a bit blurry. Making `*.cluster.local` take precedence over all Services is both dangerous and unique to sidecars

howardjohn:
Btw I think `*` is just a specialized form of `*.cluster.local` or any other broad wildcards. The same applies concerns apply to either really

ramaraochavali:
> Btw I think * is just a specialized form of *.cluster.local or any other broad wildcards. The same applies concerns apply to either really

Yes. This was my point.

ramaraochavali:
I am closing as there is no agreement on whether this is the right the thing to do or not.

PR #34138, opened 2021-07-19
Add more validation for DestinationRule

ChenglongGao327:
1. add more validation for DestinationRule
2. remove duplicate functions

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -1185,7 +1185,7 @@ func validateOutlierDetection(outlier *networking.OutlierDetection) (errs Valida
 	}
 
 	if outlier.BaseEjectionTime != nil {
-		errs = appendValidation(errs, ValidateDurationGogo(outlier.BaseEjectionTime))
+		errs = appendValidation(errs, ValidateDuration(outlier.BaseEjectionTime))
 	}
 	if outlier.ConsecutiveErrors != 0 {
 		warn := "outlier detection consecutive errors is deprecated, use consecutiveGatewayErrors or consecutive5xxErrors instead"
@@ -1197,7 +1197,7 @@ func validateOutlierDetection(outlier *networking.OutlierDetection) (errs Valida
 		errs = appendValidation(errs, errors.New(err))
 	}
 	if outlier.Interval != nil {
-		errs = appendValidation(errs, ValidateDurationGogo(outlier.Interval))
+		errs = appendValidation(errs, ValidateDuration(outlier.Interval))
 	}
 	errs = appendValidation(errs, ValidatePercent(outlier.MaxEjectionPercent), ValidatePercent(outlier.MinHealthPercent))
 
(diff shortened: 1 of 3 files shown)


pkg/config/validation/validation.go
```diff
@@ -1296,6 +1304,18 @@ func validateTLS(settings *networking.ClientTLSSettings) (errs error) {
 		}
 	}
 
+	if settings.Mode == networking.ClientTLSSettings_ISTIO_MUTUAL {
+		if settings.ClientCertificate != "" {
```
howardjohn, line 1308: This violates API backwards compatibility. A previously valid Destinationrule will now be rejected. This causes issues on upgrade, as a configuration will suddenly start being rejected.

In general, we cannot make validation stricter after an API has been introduced.

ChenglongGao327: The added part of the check for tls settings.Mode has been deleted. It is difficult to balance , in order to be compatible with the old version, users are allowed to use non-standard configurations in the new version.

pkg/config/validation/validation.go
```diff
@@ -3187,6 +3183,12 @@ func validateLocalityLbSetting(lb *networking.LocalityLoadBalancerSetting) error
 	}
 
 	for _, failover := range lb.GetFailover() {
+		if failover.From == "" || failover.To == "" {
+			return fmt.Errorf("locality lb failover settings must specify both From and To")
+		}
+		if len(strings.SplitN(failover.From, "/", 3)) > 1 || len(strings.SplitN(failover.To, "/", 3)) > 1 {
```
hzxuzhonghu, line 3189: strings.Contains

pkg/config/validation/validation.go
```diff
@@ -3187,6 +3183,12 @@ func validateLocalityLbSetting(lb *networking.LocalityLoadBalancerSetting) error
...
+			return fmt.Errorf("locality lb failover settings only different regions can be set and should not be start or end with '/'")
```
hzxuzhonghu, line 3190: locality lb failover settings must not contain `/`

ChenglongGao327: done

pkg/config/validation/validation.go
ChenglongGao327: done

hzxuzhonghu (approved):

pkg/config/validation/validation.go
```diff
@@ -3164,6 +3157,9 @@ func validateLocalityLbSetting(lb *networking.LocalityLoadBalancerSetting) error
 
 	srcLocalities := make([]string, 0, len(lb.GetDistribute()))
 	for _, locality := range lb.GetDistribute() {
+		if locality.From == "" || locality.To == nil {
+			return fmt.Errorf("locality lb distribute settings must specify both From and To")
```
howardjohn, line 3161: I don't think this is required. Its not documented in the API and its possible to configure it this way without issue.

howardjohn: BTW, I have opened https://github.com/istio/api/pull/2061 to clarify whether we should accept changes like this. It would be good to get consensus there first.

It is a tradeoff - but I think we should favor backwards compatibility over stricter validation.

howardjohn (changes requested):

pkg/config/validation/validation.go
ChenglongGao327: ok

PR #34142, opened 2021-07-19
Fix external CP integration test

ericvn:
Work on re-enabling this test with changes. The original test didn't ever work properly as its SUCCESS was a false positive. The test will verify that the test-gateway is actually running and is updated for the External Istiod changes that are in 1.11.

Thanks to many: @GregHanson, @frankbu, @howardjohn, and @stevenctl who contributed to getting this working.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/integration/multicluster/centralremotekubeconfig/main_test.go
+++ b/tests/integration/multicluster/centralremotekubeconfig/main_test.go
@@ -96,8 +96,8 @@ components:
         - path: spec.template.spec.volumes[100]
           value: |-
             name: inject-volume
-              configMap:
-                name: istio-sidecar-injector
+            configMap:
+              name: istio-sidecar-injector
         - path: spec.template.spec.containers[0].volumeMounts[100]
           value: |-
             name: config-volume
--- a/tests/integration/multicluster/centralremotekubeconfig/main_test.go
+++ b/tests/integration/multicluster/centralremotekubeconfig/main_test.go
@@ -70,11 +70,7 @@ values:
     omitSidecarInjectorConfigMap: true
     configCluster: true
   pilot:
-    configMap: true
-  istiodRemote:
-    injectionURL: https://istiod.istio-system.svc:443/inject
-  base:
-    validationURL: https://istiod.istio-system.svc:443/validate`
+    configMap: true`
 			cfg.ControlPlaneValues = `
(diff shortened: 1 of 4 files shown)


ericvn:
Cleanup doesn't seem complete as trying to run a second time on the clusters yields:
```
Test setup error: services "istiod" already exists
```

tests/integration/multicluster/centralremotekubeconfig/main_test.go
```diff
@@ -32,13 +32,13 @@ var ist istio.Instance
...
 		NewSuite(m).
-		Skip("https://github.com/istio/istio/pull/33045").
+		// Skip("https://github.com/istio/istio/pull/33045").
 		Label(label.Multicluster).
 		RequireMinClusters(2).
 		Setup(func(ctx resource.Context) error {
 			// TODO, this should be exclusively configurable outside of the framework
-			configCluster := ctx.Clusters()[0]
-			externalControlPlaneCluster := ctx.Clusters()[1]
+			configCluster := ctx.Clusters()[1]
```
howardjohn, line 40: using index seems flaky. I think `ctx.Clusters().{Remotes,Primaries}()` can do this better?

howardjohn: if not - we should make functions that do it better :slightly_smiling_face:

stevenctl: I'm not sure this test should stay around much longer anyway - it doesn't cover what we do for most multicluster tests (pod to pod traffic, telemetry, etc)

This is all configurable at the framework level, I figured we would add new jobs that set this in the topology file.

ericvn: Yes, this could be updated, but I basically left it as it was, changing what was installed on each to match the istio.io documentation. The documentation has the user install the config cluster on the `remote` cluster, and the control plane on the `external` cluster (used the `primary` cluster since that is what we have).

stevenctl (approved):

PR #34144, opened 2021-07-19
Add workload using delta xds

howardjohn:
Should give us some good e2e testing, for pretty small cost

### DIFF
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -681,10 +681,7 @@ func (s *DiscoveryServer) shouldProcessRequest(proxy *model.Proxy, req *discover
 // The delta protocol changes the request, adding unsubscribe/subscribe instead of sending full
 // list of resources. On the response it adds 'removed resources' and sends changes for everything.
 func (s *DiscoveryServer) DeltaAggregatedResources(stream discovery.AggregatedDiscoveryService_DeltaAggregatedResourcesServer) error {
-	if features.DeltaXds {
-		return s.StreamDeltas(stream)
-	}
-	return status.Errorf(codes.Unimplemented, "not implemented")
+	return s.StreamDeltas(stream)
 }
 
 // Compute and send the new configuration for a connection.
--- a/tests/integration/pilot/common/apps.go
+++ b/tests/integration/pilot/common/apps.go
@@ -229,7 +229,7 @@ func SetupApps(t resource.Context, i istio.Instance, apps *EchoDeployments) erro
 				Ports:     common.EchoPorts,
 				Subsets: []echo.SubsetConfig{{
 					Annotations: echo.NewAnnotations().Set(echo.SidecarProxyConfig, `proxyMetadata:
-  ISTIO_META_ISTIO_DELTA_XDS: "true"`),
+  ISTIO_DELTA_XDS: "true"`),
 				}},
 				WorkloadOnlyPorts: common.WorkloadPorts,
 			})
--- a/tests/integration/pilot/common/apps.go
+++ b/tests/integration/pilot/common/apps.go
@@ -219,17 +219,21 @@ func SetupApps(t resource.Context, i istio.Instance, apps *EchoDeployments) erro
 			AutoRegisterVM:    true,
 			Subsets:           []echo.SubsetConfig{{}},
 			WorkloadOnlyPorts: common.WorkloadPorts,
-		}).
-		WithConfig(echo.Config{
-			Service:   DeltaSvc,
-			Namespace: apps.Namespace,
-			Ports:     common.EchoPorts,
-			Subsets: []echo.SubsetConfig{{
-				Annotations: echo.NewAnnotations().Set(echo.SidecarProxyConfig, `proxyMetadata:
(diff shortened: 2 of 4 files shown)


pilot/pkg/xds/ads.go
```diff
@@ -681,10 +681,7 @@ func (s *DiscoveryServer) shouldProcessRequest(proxy *model.Proxy, req *discover
 // The delta protocol changes the request, adding unsubscribe/subscribe instead of sending full
 // list of resources. On the response it adds 'removed resources' and sends changes for everything.
 func (s *DiscoveryServer) DeltaAggregatedResources(stream discovery.AggregatedDiscoveryService_DeltaAggregatedResourcesServer) error {
-	if features.DeltaXds {
```
adiprerepa, line 684: why do we want to remove this? I think we still want to return not implemented if we don't fully support/people don't actively enable the flag.

howardjohn: I don't see why we need to block it on control plane and data plane side. If data plane is not configured to use delta, it would never be used. The only reason I guess would be if we think allowing clients to connect to delta introduces risk to the control plane, but I am not sure that is too large of a concern.

The reason I did it is that we would need a custom deployment of Istiod to test delta as is; with this removed we can just deploy a single workload using delta

adiprerepa: fair enough

adiprerepa (approved):

tests/integration/pilot/common/apps.go
```diff
@@ -216,6 +219,16 @@ func SetupApps(t resource.Context, i istio.Instance, apps *EchoDeployments) erro
...
 			Subsets:           []echo.SubsetConfig{{}},
 			WorkloadOnlyPorts: common.WorkloadPorts,
+		}).
+		WithConfig(echo.Config{
+			Service:   DeltaSvc,
+			Namespace: apps.Namespace,
+			Ports:     common.EchoPorts,
+			Subsets: []echo.SubsetConfig{{
+				Annotations: echo.NewAnnotations().Set(echo.SidecarProxyConfig, `proxyMetadata:
+  ISTIO_META_ISTIO_DELTA_XDS: "true"`),
```
hzxuzhonghu, line 229: What does `ISTIO_META_ISTIO_DELTA_XDS` do?

adiprerepa: I think it directs xds traffic to the delta api

tests/integration/pilot/common/apps.go
```diff
@@ -255,6 +268,9 @@ func SetupApps(t resource.Context, i istio.Instance, apps *EchoDeployments) erro
 	if !t.Settings().SkipVM {
 		apps.VM = echos.Match(echo.Service(VMSvc))
 	}
+	if !t.Settings().SkipDelta {
+		apps.DeltaXDS = echos.Match(echo.Service(DeltaSvc))
```
stevenctl, line 272: 1. We don't need to deploy the app if this flag is on
2. "workload agnostic" tests won't filter this out

The only reason we don't deal with this for VMs is here:
https://github.com/istio/istio/blob/943ba0765876590d5c6da89d5df034fe0ea0808a/pkg/test/framework/components/echo/echoboot/echoboot.go#L98

howardjohn: I fixed (1), I am not sure I understand how (2) works. if there are zero instances, then how/why are we sending traffic to it?

stevenctl (approved):

tests/integration/pilot/common/apps.go
hzxuzhonghu: xdsType := "GRPC"
	if features.DeltaXds {
		xdsType = "DELTA_GRPC"
	}

	opts = append(opts,
		option.NodeID(cfg.ID),
		option.NodeType(cfg.ID),
		option.PilotSubjectAltName(cfg.Metadata.PilotSubjectAltName),
		option.ProxyViaAgent(cfg.Metadata.ProxyViaAgent),
		option.OutlierLogPath(cfg.Metadata.OutlierLogPath),
		option.ProvCert(cfg.Metadata.ProvCert),
		option.DiscoveryHost(discHost),
		option.XdsType(xdsType))

I donot understand how `ISTIO_META_ISTIO_DELTA_XDS` translated to xdsType

hzxuzhonghu:
Sorry i have a last question

tests/integration/pilot/common/apps.go
howardjohn: I think you are right, should be ISTIO_DELTA_XDS

hzxuzhonghu: We donot have doc about this env either

howardjohn: its in https://preliminary.istio.io/latest/docs/reference/commands/pilot-agent/#envvars. its an experimental feature for now, so not any good docs for (or ever - its not a use facing functionality, just performance optimization)

howardjohn: Updated and verified that it correctly uses delta now

hzxuzhonghu: Thanks

hzxuzhonghu (approved):

PR #34166, opened 2021-07-20
[WIP] handle NodePort services with external traffic policy set to Local

su225:
Signed-off-by: su225 <suchithjn22@gmail.com>

Currently, for node port services Istio assumes that the external traffic policy is set to `Cluster`. It is the Kubernetes default as well. So for external addresses field, it adds all external node IPs. However, there is another option called `Local` which is not handled correctly leading to traffic drops. This PR adds support to it.

**Problem**
We have a use case where gateway's NodePort service's external traffic policy is set to `Local` instead of the default `Cluster`. This is done to preserve the original source IP address as seen by the gateway workload (Setting it to `Cluster` applies SNAT which we want to avoid). However, the problem with setting to Local is that if the traffic hits the node which does not host the workload then it is dropped. So the list of external addresses in Istio's internal model should contain only those nodes which actually host the workload.

Consider the following example
The cluster contains the following nodes (IPs here are for illustration only)
- node1 (10.100.100.10)
- node2 (10.100.100.20)
- node3 (10.100.100.30)

To start with we have service instances in the following nodes
- node1 (10.100.100.10)

Current behavior - external addresses would be (valid when external traffic policy is `Cluster`)
- 10.100.100.10
- 10.100.100.20
- 10.100.100.30

Expected behavior for policy = `Local`
- 10.100.100.10

Say we scaled up to 2 instances so that it is scheduled on 2 different nodes
- node1 (10.100.100.10)
- node2 (10.100.100.20)

Now expected behavior for policy = `Local`
- 10.100.100.10
- 10.100.100.20
- 10.100.100.30 <---- This should not be in the list

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


howardjohn:
How does this behave without Istio at all? I was under the impression if you use `Local` its your responsibility to make sure it runs on all nodes (ie Daemonset)?

Is kube-proxy smart enough to do this as well, or are we trying to make Istio behave better than kube-proxy?

vikaschoudhary16:
> How does this behave without Istio at all? I was under the impression if you use `Local` its your responsibility to make sure it runs on all nodes (ie Daemonset)?

no no... with daemonset is just a particular use case. Without istio NLB would do healthchecks to figure out which nodes have pods for the service running and will forward to those only. In our use case, i.e with Istio, in place of NLB there is another envoy. Cluster size very big and envoy is not able to keep up with the active health checking at this scale.  our tooling listens on xds updates which we use as a lib. So with this change, out of band, it would be possible to filter out nodes which are not relevant. 
This is also a relevant on `local` policy i think, https://www.youtube.com/watch?v=y2bhV81MfKQ&t=1823s
> 
> Is kube-proxy smart enough to do this as well, or are we trying to make Istio behave better than kube-proxy?

with `local` policy we just ask kube-proxy not to show smartness  to avoid extra network  hops :)

howardjohn:
I guess I am not understanding the request flow. If we have gateway=Local, and NLB in front, then NLB will do the health checking, which should work fine today? The description says "We have a use case where gateway's NodePort service's external traffic policy is set to Local" so it seems like this is the target use case? If so, I don't understand how XDS can impact this at all.

Is this for the case we have NLB (or anything, really) -> Gateway -> backend service (policy=Local)? If so, this still doesn't make 100% sense to me, since we are forwarding to pod IPs not node IPs.

I am probably just missing something - let me know

vikaschoudhary16:
> I guess I am not understanding the request flow. If we have gateway=Local, and NLB in front, then NLB will do the health checking, which should work fine today? The description says "We have a use case where gateway's NodePort service's external traffic policy is set to Local" so it seems like this is the target use case? If so, I don't understand how XDS can impact this at all.
> 
> Is this for the case we have NLB (or anything, really) -> Gateway -> backend service (policy=Local)? If so, this still doesn't make 100% sense to me, since we are forwarding to pod IPs not node IPs.
> 
> I am probably just missing something - let me know

We are trying to address the "multi-primay multi-network" case where east-west gw is nodeport service and externalTrafficPolicy for this service is local. Without PR changes, multicluster controller on client cluster will populate all the nodeips in the networkGateway addresses for this east-west gw. If policy is local, some of those will not work

howardjohn:
Thanks for the explanation, I think I understand this now. Assigned folks familiar with this area

pilot/pkg/serviceregistry/kube/controller/network.go
```diff
@@ -242,25 +244,65 @@ func (c *Controller) updateServiceNodePortAddresses(svcs ...*model.Service) bool
...
 	}
 	for _, svc := range svcs {
+		if c.isNodePortGatewayService(svc) && svc.Attributes.ExternalTrafficPolicy == model.ExternalTrafficPolicyLocal {
+			endpoints := c.endpoints.buildIstioEndpointsWithService(svc.Attributes.Name, svc.Attributes.Namespace, svc.Hostname)
+			if features.EnableK8SServiceSelectWorkloadEntries {
+				fep := c.collectWorkloadInstanceEndpoints(svc)
+				endpoints = append(endpoints, fep...)
+			}
+			nodesWithEndpoints := getWorkloadNodeLocations(endpoints)
+			log.Debugf("---> svc:%s nodesWithEndpoints = %#v", svc.Hostname, nodesWithEndpoints)
```
ramaraochavali, line 254: nit: remove or change it to correct format

su225: Removed it.

hzxuzhonghu:
@su225 I donot think using NodePort type service like this is a good practice, in k8s cluster, workloads are dynamic down and up, the scheduler could schedule them to any node.  So I wonder how you control the workloads all scheduled to the nodes that used as east-west gateways

su225:
@hzxuzhonghu
>  So I wonder how you control the workloads all scheduled to the nodes that used as east-west gateways

Simple answer - **we don't**. This is related to `externalTrafficPolicy` field in Kubernetes service. It accepts two possible values - `Cluster` (default) and `Local`. When it is cluster, you can route traffic to any node and it will bounce it off to the node containing the workload. So you have an additional hop - from the node in another cluster to the node that actually has the workload. Currently, Istio assumes that traffic policy would be `Cluster` and adds addresses of all the nodes in the other cluster as you can see here - https://github.com/istio/istio/blob/c4881a7bce05ab288ecd34f9cc6be9347aa39729/pilot/pkg/serviceregistry/kube/controller/network.go#L251-L255

This does not work when traffic policy is `Local` which is used to save an additional hop - the consequence of this is that if you route traffic to the node that does not host the workload, then is dropped leading to additional latency due to retries or the request might error out. It is an annoying, non-deterministic and hard-to-debug issue. **So we have to track which nodes host east-west gateway workloads and only add the addresses of those nodes to the list of endpoints ONLY when traffic policy is `Local`**. As you can see we are not messing with Kubernetes scheduling in any way. It is just handling the case where traffic policy is `Local` correctly. This is in addition to what Istio does already. 

So we have an additional filter for node addresses - https://github.com/istio/istio/blob/72825655277d2ca618e1cf838cea1f9835d2e0bf/pilot/pkg/serviceregistry/kube/controller/network.go#L288-L296

hzxuzhonghu:
I mean, for example, you have 3 nodes that have external IP, which can be used as gateways, but workloads can be scheduled to other nodes as well, in this case if the Local mode is set, it definitely is not expected. So that's why I doubt if this is a real use case

vikaschoudhary16:
> I mean, for example, you have 3 nodes that have external IP, which can be used as gateways, but workloads can be scheduled to other nodes as well, in this case if the Local mode is set, it definitely is not expected. 

taints and tolerations can be used in such case to restrict gw pods to the specified set of nodes. Though in our case, clusters are on-prem where all the nodeips are reachable from peer cluster.

> So that's why I doubt if this is a real use case

use case is absolutely real :)

hzxuzhonghu:
make sense for users that have no load balancer.

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -246,6 +246,11 @@ type Controller struct {
 	workloadInstancesByIP map[string]*model.WorkloadInstance
 	// Stores a map of workload instance name/namespace to address
 	workloadInstancesIPsByName map[string]string
+	// serviceToWorkloadNodesMap is a mapping of NodePort services to the list of kubernetes node names
+	// on which there is at least one workload belonging to the service is running. This is needed to handle
+	// node-port services with external traffic policy type set to Local (default is Cluster). This is because
+	// when it is set to local and traffic is sent to a node not hosting a workload, it will be dropped
+	serviceToWorkloadNodesMap map[host.Name]map[string]struct{}
```
ramaraochavali, line 253: Does it contain services  with Local externalTrafficPolicy only? Atleast That is what I understood based on my code reading. If yes, can you update comments?

ramaraochavali (commented):
Few comments. Generally looks good.

pilot/pkg/serviceregistry/kube/controller/controller.go
stevenctl: Can we also make the name reflect this is used specifically for network gateways (service is labeled `topolgy.istio.io/network=something`) and not all NodePort services in the cluster

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -503,6 +509,11 @@ func (c *Controller) onNodeEvent(obj interface{}, event model.Event) error {
 		updatedNeeded = true
 		c.Lock()
 		delete(c.nodeInfoMap, node.Name)
+		// We should remove the node for all the services that
+		// had a workload there as we don't want to route traffic
+		for h := range c.serviceToWorkloadNodesMap {
+			delete(c.serviceToWorkloadNodesMap[h], node.Name)
```
stevenctl, line 515: It's probably possible for `c.serviceToWorkloadNodesMap[h]` to be nil?

pilot/pkg/serviceregistry/kube/controller/endpointcontroller.go
```diff
@@ -109,6 +110,15 @@ func updateEDS(c *Controller, epc kubeEndpointsController, ep interface{}, event
 		}
 	}
 
+	// Handle external address update for NodePort services whose external
```
stevenctl, line 113: Would we enter this func for these updates (do NodePort services update Endpoints/EndpointSlice)?

pilot/pkg/serviceregistry/kube/controller/network.go
```diff
@@ -306,3 +347,10 @@ func (s gatewaySet) toArray() []*model.NetworkGateway {
 	// Sort the array so that it's stable.
 	return model.SortGateways(gws)
 }
+
+func (c *Controller) isNodePortGatewayService(svc *model.Service) bool {
```
stevenctl, line 351: Now we have a func and a method with the same name but different logic:

https://github.com/istio/istio/blob/ed5761ebe373cf158b16f033e69a216997ea30c9/pilot/pkg/serviceregistry/kube/controller/util.go#L191

hzxuzhonghu:
I will take a look today as I have investigated node local service routing.

pilot/pkg/model/service.go
```diff
@@ -487,6 +492,38 @@ type ServiceAttributes struct {
...
 	// We translate that to the appropriate node port here.
 	ClusterExternalPorts map[cluster.ID]map[uint32]uint32
+
+	// ExternalTrafficPolicy affects the list of available endpoints available in case
+	// of NodePort services. This is useful to preserve source IP address. If the traffic
+	// is sent to a node which does not have a workload then it would be dropped.
+	ExternalTrafficPolicy ExternalTrafficPolicy
+}
+
+type ExternalTrafficPolicy string
```
hzxuzhonghu, line 502: It looks we donot need to redefine them again

pilot/pkg/serviceregistry/kube/controller/controller.go
hzxuzhonghu: At first glance i could not know what does this mean, will read through

pilot/pkg/serviceregistry/kube/controller/controller.go
hzxuzhonghu, line 515: Concerned with the performance of the  loop

pilot/pkg/serviceregistry/kube/controller/endpointcontroller.go
hzxuzhonghu: Why do we need to handle node port service when endpoint update?

pilot/pkg/serviceregistry/kube/controller/network.go
```diff
@@ -242,25 +244,64 @@ func (c *Controller) updateServiceNodePortAddresses(svcs ...*model.Service) bool
...
 	for _, svc := range svcs {
+		if c.isNodePortGatewayService(svc) && svc.Attributes.ExternalTrafficPolicy == model.ExternalTrafficPolicyLocal {
+			endpoints := c.endpoints.buildIstioEndpointsWithService(svc.Attributes.Name, svc.Attributes.Namespace, svc.Hostname)
+			if features.EnableK8SServiceSelectWorkloadEntries {
+				fep := c.collectWorkloadInstanceEndpoints(svc)
+				endpoints = append(endpoints, fep...)
+			}
+			nodesWithEndpoints := getWorkloadNodeLocations(endpoints)
+			c.Lock()
+			c.serviceToWorkloadNodesMap[svc.Hostname] = nodesWithEndpoints
```
hzxuzhonghu, line 255: I get to know serviceToWorkloadNodesMap stores the svc --> nodes (where the service instance resides) 

 Same performance concern to do it here.

hzxuzhonghu (changes requested):
We get gateway address from NetworkManager. But I do not see where the service is considered when acquiring the gw address. So i doubt the rightness.
And another problem is loadbalancet weight for each endpoint is not right, as you do not consider how many endpoints reside on a single node

pilot/pkg/model/service.go
```diff
@@ -487,6 +492,38 @@ type ServiceAttributes struct {
 	// The port that the user provides in the meshNetworks config is the service port.
 	// We translate that to the appropriate node port here.
 	ClusterExternalPorts map[cluster.ID]map[uint32]uint32
+
+	// ExternalTrafficPolicy affects the list of available endpoints available in case
```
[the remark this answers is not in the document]
su225: Oops. Thank you. Fixed it.

pilot/pkg/serviceregistry/kube/controller/network.go
```diff
@@ -242,25 +244,64 @@ func (c *Controller) updateServiceNodePortAddresses(svcs ...*model.Service) bool
...
+				!c.containsWorkloadForLocalTrafficService(svc, nodeName) {
+				continue
+			}
+			nodeAddresses = append(nodeAddresses, n.address)
+		}
+	}
+	svc.Attributes.ClusterExternalAddresses = map[cluster.ID][]string{c.opts.ClusterID: nodeAddresses}
+}
+
+func (c *Controller) containsWorkloadForLocalTrafficService(svc *model.Service, nodeName string) bool {
```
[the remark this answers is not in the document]
su225: It is used in only one place and the caller (`updateClusterExternalAddressesForNodePortServices`) already had a lock. Anyways, I removed the function altogether and added a comment explaining why we have to handle this type of external traffic policy specially.

pilot/pkg/serviceregistry/kube/controller/controller.go
su225: @stevenctl - I changed it to `nodePortGwLocalServiceToWorkloadNodesMap`. It is long and wordy. Please feel free to suggest better alternatives 

@ramaraochavali - Yes. It is for those NodePort gateway services with external traffic policy set to `Local`. And yes, I updated comments.
```go
	// nodePortGwLocalServiceToWorkloadNodesMap is a mapping of NodePort services with external traffic policy set to Local
	// to the list of kubernetes node names on which there is at least one workload belonging to the service is
	// running. This is needed to handle node-port services with external traffic policy type set to Local
	// (default is Cluster). This is because when it is set to local and traffic is sent to a node not hosting
	// a workload, it will be dropped
	nodePortGwLocalServiceToWorkloadNodesMap map[host.Name]map[string]struct{}
```

pilot/pkg/serviceregistry/kube/controller/endpointcontroller.go
su225: @stevenctl @hzxuzhonghu - No. `NodePort` services don't update Endpoints/EndpointSlice. But when external traffic policy is Local, we need to handle endpoint update. Let me give you an example

Imagine a cluster with 3 nodes on which we can schedule gateway pods (let us call them node-1, node-2 and node-3). Let node's external IPs be 
```
node-1: 34.34.78.78
node-2: 34.34.78.79
node-3: 34.34.78.80
```
Now let us assume that the pod is initially scheduled in node-2. So the list of external addresses would only have `34.34.78.79`. When the pod is rescheduled to node-1 (say, because node-2 is being upgraded/patched), the list of external addresses should be `34.34.78.78`. 

As you can see, the list of external addresses depends on the nodes where endpoints are scheduled. So we should also handle endpoint events to update the mapping from service-hostname to the list of nodes containing the workload. In the example, I gave it would be initially `istio-eastwestgateway -> [node-2]`. On endpoint event due to reschedule, it would be `istio-eastwestgateway -> [node-1]`. The maps are updated by `updateServiceNodePortAddresses` called due to endpoint event.

pilot/pkg/serviceregistry/kube/controller/network.go
su225: Removed this.

pilot/pkg/model/service.go
su225: Removed it.

pilot/pkg/serviceregistry/kube/controller/controller.go
su225: This is when a node go down (which itself is not too frequent). The map contains only the gateway NodePort services whose external traffic policy is set to `Local`. I don't expect thousands of them to be present. So I thought a loop is fine. If you think we need to optimise this then it is possible to have an inverse map `node -> [nodePortWithLocalServices]` and only pick those keys while iterating.

pilot/pkg/serviceregistry/kube/controller/network.go
su225: This cost is limited to node port gateways with external traffic policy set to Local. As I mentioned earlier, in practice, we don't expect a lot of them. So I think it should be fine. One potential concern could be the call on EDS update which is more frequent than service or node updates. However, this is applicable only for specific kind of Gateway services. AFAIK, it would not be too many. So it should be fine?

pilot/pkg/serviceregistry/kube/controller/endpointcontroller.go
stevenctl: When our set of node addresses change, this will trigger us to recompute the mappings:

https://github.com/istio/istio/blob/943ba0765876590d5c6da89d5df034fe0ea0808a/pilot/pkg/serviceregistry/kube/controller/controller.go#L531

https://github.com/istio/istio/blob/943ba0765876590d5c6da89d5df034fe0ea0808a/pilot/pkg/serviceregistry/kube/controller/controller.go#L456

Assuming the correct set of addresses are updated here:

https://github.com/istio/istio/blob/1fe1ec39d19502ebef06b1c1e22daba213175f30/pilot/pkg/serviceregistry/kube/controller/network.go#L159

I think everything is compatible with that already.

stevenctl:
> We get gateway address from NetworkManager. But I do not see where the service is considered when acquiring the gw address. So i doubt the rightness.
> And another problem is loadbalancet weight for each endpoint is not right, as you do not consider how many endpoints reside on a single node

I think aside from the endpoints handler it looks good; network manager pulls from our `c.networkGateways` map; weight computation is done in EDS.

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -503,6 +510,13 @@ func (c *Controller) onNodeEvent(obj interface{}, event model.Event) error {
 		updatedNeeded = true
 		c.Lock()
 		delete(c.nodeInfoMap, node.Name)
+		// We should remove the node for all the services that
+		// had a workload there as we don't want to route traffic.
+		// For performance reasons, we do it only for gateway services
+		// whose external traffic policy is set to Local.
+		for h := range c.nodePortGwLocalServiceToWorkloadNodesMap {
```
hzxuzhonghu, line 517: Only need to do this when this node is a gateway node, right? So u can do this only when the node is in `nodeInfoMap`

pilot/pkg/serviceregistry/kube/controller/network.go
```diff
@@ -242,25 +244,65 @@ func (c *Controller) updateServiceNodePortAddresses(svcs ...*model.Service) bool
 		return false
 	}
 	for _, svc := range svcs {
+		c.RLock()
+		_, found := c.nodeSelectorsForServices[svc.Hostname]
+		isNodePortGateway := found && c.servicesMap[svc.Hostname] != nil
```
hzxuzhonghu, line 249: why not reuse `isNodePortGatewayService`

pilot/pkg/serviceregistry/kube/controller/endpointcontroller.go
```diff
@@ -96,11 +96,12 @@ func updateEDS(c *Controller, epc kubeEndpointsController, ep interface{}, event
 		endpoints = epc.buildIstioEndpoints(ep, host)
 	}
 
+	c.RLock()
+	svc := c.servicesMap[host]
```
nmittler, line 100: nit: the 2 blocks below both only execute if `svc != nil`. Maybe we could do something like:

```golang
if svc != nil {
    // handling k8s service selecting workload entries
    if features.EnableK8SServiceSelectWorkloadEntries {
        fep := c.collectWorkloadInstanceEndpoints(svc)
        endpoints = append(endpoints, fep...)
    }

    // Handle external address update for NodePort services whose external
    // traffic policy is set to Local.
    if len(svc.Attributes.ClusterExternalPorts) > 0 &&
        svc.Attributes.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeLocal &&
        c.updateServiceNodePortAddresses(svc) {
        c.opts.XDSUpdater.SvcUpdate(string(c.opts.ClusterID), string(svc.Hostname), svc.Attributes.Namespace, event)
    }
}
```

pilot/pkg/serviceregistry/kube/controller/network.go
```diff
@@ -242,25 +244,65 @@ func (c *Controller) updateServiceNodePortAddresses(svcs ...*model.Service) bool
...
+		c.RUnlock()
+
+		if isNodePortGateway && svc.Attributes.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeLocal {
+			endpoints := c.buildEndpointsForService(svc)
```
nmittler, line 253: Should `buildEndpointsForService` just do this filtering instead? I'm a little concerned that other code paths might miss this logic. Maybe it's not an issue?

releasenotes/notes/34246.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+issue:
+  - 34246
+releaseNotes:
+  - |
+    **Added** support for NodePort eastwest gateway when
```
nmittler, line 8: Maybe expand this a bit to add more context for someone unfamiliar with the problem?

pilot/pkg/serviceregistry/kube/controller/controller.go
su225: Yes. Only when the node hosts some gateway with Local policy this map is filled. Added the check.

pilot/pkg/serviceregistry/kube/controller/network.go
su225: `isNodePortGatewayService` takes a Kubernetes service as parameter. So this had to be done.

pilot/pkg/serviceregistry/kube/controller/endpointcontroller.go
su225: That's right. I updated it. However, I would like to keep `updateServiceNodePortAddresses` separate for readability. 
```go
if len(svc.Attributes.ClusterExternalPorts) > 0 &&
	svc.Attributes.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeLocal {

	// It is written this way for clarity. updateServiceNodePortAddresses has side-effect
	// where it can update the node port addresses of the passed service. If it is updated,
	// then "addressUpdated" would be true and we have to trigger Service Update [XDS]
	if addressesUpdated := c.updateServiceNodePortAddresses(svc); addressesUpdated {
		c.opts.XDSUpdater.SvcUpdate(shard, string(svc.Hostname), svc.Attributes.Namespace, event)
	}
}
```

pilot/pkg/serviceregistry/kube/controller/network.go
su225: I didn't get it. Sorry. There, endpoint computation is done to update `nodePortGwLocalServiceToWorkloadNodesMap` which tracks the nodes with at least one endpoint for a given Service (keyed by hostname). Ideally `buildEndpointsForService` should not be concerned with Node addresses which could be different from Workload addresses (in Kubernetes, it is node IP and pod IPs would be different).

In that particular `if` block, list of node addresses are being updated. For this particular case, it is the subset of the cluster nodes. When external traffic policy is `Cluster` it would be all nodes of the cluster.

releasenotes/notes/34246.yaml
su225: Done.

mindtraveller:
I see it takes some time to get this PR merged, might this help to review it https://viezly.com/change_requests/4139?

PR #34170, opened 2021-07-20
add server to filtered cipher suites log

ramaraochavali:
[ ] Docs
[ ] Installation
[X ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/model/gateway.go
+++ b/pilot/pkg/model/gateway.go
@@ -169,11 +169,11 @@ func MergeGateways(gateways []gatewayWithInstances) *MergedGateway {
 						autoPassthrough = true
 					}
 					if len(s.Tls.CipherSuites) > 0 {
-						serverName := s.Name
-						if len(serverName) == 0 {
-							serverName = s.Port.Name
+						serverIdentifier := s.Name
+						if len(serverIdentifier) == 0 {
+							serverIdentifier = s.Port.Name
 						}
-						s.Tls.CipherSuites = filteredCipherSuites(s.Tls.CipherSuites, serverName, gatewayName)
+						s.Tls.CipherSuites = filteredCipherSuites(s.Tls.CipherSuites, serverIdentifier, gatewayName)
 					}
 				}
 				serverPort := ServerPort{resolvedPort, s.Port.Protocol, s.Bind}
@@ -473,13 +473,13 @@ func getTargetPortMap(serversByRouteName map[string][]*networking.Server) Gatewa
 }
 
 // Invalid cipher suites lead Envoy to NACKing. This filters the list down to just the supported set.
-func filteredCipherSuites(suites []string, serverName string, gatewayName string) []string {
+func filteredCipherSuites(suites []string, serverIdentifier string, gatewayName string) []string {
 	ret := make([]string, 0, len(suites))
 	for _, s := range suites {
 		if security.IsValidCipherSuite(s) {
 			ret = append(ret, s)
 		} else {
-			log.Debugf("ignoring unsupported cipherSuite: %q for server %s in gateway %s", s, serverName, gatewayName)
+			log.Debugf("ignoring unsupported cipherSuite: %q for server %s in gateway %s", s, serverIdentifier, gatewayName)
 		}
 	}
 	return ret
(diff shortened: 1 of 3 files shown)


ramaraochavali:
@howardjohn would this work?

pilot/pkg/model/gateway.go
```diff
@@ -167,6 +168,13 @@ func MergeGateways(gateways []gatewayWithInstances) *MergedGateway {
 					if s.Tls.Mode == networking.ServerTLSSettings_AUTO_PASSTHROUGH {
 						autoPassthrough = true
 					}
+					if len(s.Tls.CipherSuites) > 0 {
+						serverName := s.Name
+						if len(serverName) == 0 {
+							serverName = s.Port.Name
```
hzxuzhonghu, line 174: This is hacky

hzxuzhonghu (commented):
lgtm

hzxuzhonghu:
Just noticed the issue. For the previous gateway with invalid cipher suites, it is invalid with nack. I would rather prevent it in validation. We can assume no such gateways exist in mesh.

pilot/pkg/model/gateway.go
ramaraochavali: So we either identify server with unique port name or server name. Why do you think it is hacky?

ramaraochavali:
https://github.com/istio/istio/pull/34109 - We discussed about that here and decided to not prevent such gateways but add warning because a gateway may have many servers and only a single cipher suite may be invalid in server.

pilot/pkg/model/gateway.go
```diff
@@ -167,6 +168,13 @@ func MergeGateways(gateways []gatewayWithInstances) *MergedGateway {
 					if s.Tls.Mode == networking.ServerTLSSettings_AUTO_PASSTHROUGH {
 						autoPassthrough = true
 					}
+					if len(s.Tls.CipherSuites) > 0 {
+						serverIdentifier := s.Name
+						if len(serverIdentifier) == 0 {
+							serverIdentifier = s.Port.Name
+						}
+						s.Tls.CipherSuites = filteredCipherSuites(s.Tls.CipherSuites, serverIdentifier, gatewayName)
```
howardjohn, line 176: is it safe to mutate here? Will this impact the underlying client-go cache?

hzxuzhonghu: I didnot realize it is mutating the original object

ramaraochavali: So is it OK if filteredCipherSuite returns a different set, we clone the server and set?

hzxuzhonghu: What is this pr fixing?

ramaraochavali: #34109 added validation and logic to skip invalid cipher suites. But the logic to skip is executed on every listener creation. I am trying to see if we can do it once with this PR.

ramaraochavali:
I tried to clone when ciphersuites are invalid - but selective cloning does not work as the same server is referenced in multiple places like serversByRouteName. I felt cloning every server is not efficient - rather we pay the cost of comparison at listener creation. So I reverted that change and just added server to log to indicate which server is causing issue.

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -992,3 +979,17 @@ func buildGatewayVirtualHostDomains(hostname string, port int) []string {
...
+
+// Invalid cipher suites lead Envoy to NACKing. This filters the list down to just the supported set.
+func filteredCipherSuites(server *networking.Server) []string {
+	suites := server.Tls.CipherSuites
+	ret := make([]string, 0, len(suites))
+	for _, s := range suites {
+		if security.IsValidCipherSuite(s) {
+			ret = append(ret, s)
+		} else {
+			log.Debugf("ignoring unsupported cipherSuite: %q for server %s", s, server.String())
```
howardjohn, line 991: nit: server.String is fairly expensive. Maybe we can wrap in `if log.DebugEnabled` so we do not pay for it when debug log is off

hzxuzhonghu (approved):

PR #34186, opened 2021-07-20
Fix container builds on arm64

AWSjswinney:

[ ] Docs
[X] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

This change removes the default value from the usages of `${TARGETARCH}` argument in docker files. It's not immediately clear why this is necessary, but on when building on `arm64` with `make dockerx`, this default value prevents Docker from locating the correct binary version. According to [Docker documentation](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope), that `ARG` is always provided when using the Buildkit backend, which I *think* is the case, but please correct me if I am wrong.

### DIFF
--- a/pilot/docker/Dockerfile.pilot
+++ b/pilot/docker/Dockerfile.pilot
@@ -15,7 +15,7 @@ FROM gcr.io/istio-release/distroless:${BASE_VERSION} as distroless
 FROM ${BASE_DISTRIBUTION}
 
 ARG TARGETARCH
-COPY ${TARGETARCH:-amd64}/pilot-discovery /usr/local/bin/pilot-discovery
+COPY ${TARGETARCH}/pilot-discovery /usr/local/bin/pilot-discovery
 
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
This will break `make docker.pilot` (ie no buildkit). I think we have some usage without buildkit due to bugs that have persisted in buildx to this day, as well as lack of support in some environments.

Is it possible to set it only if its not configured by buildx or something?

Also if we do this we should change all the Dockerfiles, there are many others with the default

AWSjswinney:
> Is it possible to set it only if its not configured by buildx or something?

I'm not sure, but I will look into it to see what I can find out.

AWSjswinney:
I can build the containers with buildx... at some point I may look into fixing the the non-buildkit case.

PR #34189, opened 2021-07-20
Improve iptables debugability

### LINKED ISSUE #33965: Improve iptables debugability

howardjohn:
Fixes https://github.com/istio/istio/issues/33965

This adds two big changes to iptables:
* comment each iptables rule, to explain what it is doing
* Add (optional) trace logging to each rule. This allows seeing which rules each packet hits. The istio-agent will emit these logs, which requires NET_ADMIN. This is meant for spot-debugging only and not intended to run continually -- it is very noisy.

This PR has the base foundation complete, but it is not fully implemented - only a few of the chains will emit logs currently. Opening early to get feedback.

Example outbound request:
```
2021-07-20T23:32:16.925827Z     info    iptables        direct all traffic through ISTIO_OUTBOUND chain command=IPT02 uid=0 gid=0 inDev= outDev=eth0 src=10.244.0.42 dst=10.96.116.22
2021-07-20T23:32:16.926390Z     info    iptables        direct all traffic through ISTIO_OUTBOUND chain command=IPT02 uid=0 gid=1337 inDev= outDev=eth0 src=10.244.0.42 dst=10.244.0.6
```
Here you can see request to envoy, and the request from envoy.

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/files/injection-template.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/injection-template.yaml
@@ -310,6 +310,20 @@ spec:
       failureThreshold: {{ annotation .ObjectMeta `readiness.status.sidecar.istio.io/failureThreshold` .Values.global.proxy.readinessFailureThreshold }}
     {{ end -}}
     securityContext:
+      {{- if eq (index .ProxyConfig.ProxyMetadata "IPTABLES_TRACE_LOGGING") "true" }}
+      allowPrivilegeEscalation: true
+      capabilities:
+        add:
+        - NET_ADMIN
+        drop:
+        - ALL
+      privileged: true
+      readOnlyRootFilesystem: {{ ne (annotation .ObjectMeta `sidecar.istio.io/enableCoreDump` .Values.global.proxy.enableCoreDump) "true" }}
+      runAsGroup: 1337
+      fsGroup: 1337
+      runAsNonRoot: false
+      runAsUser: 0
+      {{- else }}
       allowPrivilegeEscalation: {{ .Values.global.proxy.privileged }}
       capabilities:
         {{ if or (eq (annotation .ObjectMeta `sidecar.istio.io/interceptionMode` .ProxyConfig.InterceptionMode) `TPROXY`) (eq (annotation .ObjectMeta `sidecar.istio.io/capNetBindService` .Values.global.proxy.capNetBindService) `true`) -}}
@@ -333,6 +347,7 @@ spec:
       {{- else -}}
       runAsNonRoot: true
       runAsUser: 1337
+      {{- end }}
       {{- end }}
     resources:
   {{- if or (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyCPU`) (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyMemory`) (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyCPULimit`) (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyMemoryLimit`) }}
--- a/tools/istio-iptables/pkg/builder/iptables_builder_impl.go
+++ b/tools/istio-iptables/pkg/builder/iptables_builder_impl.go
@@ -20,6 +20,7 @@ import (
 
 	"istio.io/istio/tools/istio-iptables/pkg/config"
 	"istio.io/istio/tools/istio-iptables/pkg/constants"
+	"istio.io/istio/tools/istio-iptables/pkg/log"
 )
 
 // Rule represents iptables rule - chain, table and options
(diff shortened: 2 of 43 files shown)


tools/istio-iptables/pkg/builder/iptables_builder_impl.go
```diff
@@ -67,11 +70,37 @@ func (rb *IptablesBuilder) InsertRuleV6(chain string, table string, position int
...
+	return -1 // not found.
+}
+
+func (rb *IptablesBuilder) AppendRuleV4(command log.CommandID, chain string, table string, params ...string) *IptablesBuilder {
+	idx := indexOf("-j", params)
+	if rb.loggingEnable && idx >= 0 && command != log.UndefinedCommand {
+		match := params[:idx]
+		// 1337 group is just a random constant to be matched on the log reader side
+		// Size of 20 allows reading the IPv4 IP header.
+		match = append(match, "-j", "NFLOG", "--nflog-prefix", fmt.Sprintf(`"%s"`, command), "--nflog-group", "1337", "--nflog-size", "20")
```
jacob-delgado, line 88: Do you have to also modify AppendRuleV4?

jacob-delgado:
This looks great as troubleshooting IP tables is rather tedious.

howardjohn:
@jacob-delgado @lambdai finally got this complete and green

tools/istio-iptables/pkg/config/config.go
```diff
@@ -41,7 +41,7 @@ type Config struct {
 	OutboundPortsExclude    string        `json:"OUTBOUND_PORTS_EXCLUDE"`
 	OutboundIPRangesInclude string        `json:"OUTBOUND_IPRANGES_INCLUDE"`
 	OutboundIPRangesExclude string        `json:"OUTBOUND_IPRANGES_EXCLUDE"`
-	KubevirtInterfaces      string        `json:"KUBEVIRT_INTERFACES"`
+	KubevirtInterfaces      string        `json:"KUBE_VIRT_INTERFACES"`
```
lambdai, line 44: That's not consistent 
KubevirtInterfaces vs KUBE_VIRT_INTERFACES
Why?

tools/istio-iptables/pkg/log/command.go
```diff
@@ -0,0 +1,42 @@
...
+	JumpOutbound            = Command{"IPT02", "direct all traffic through ISTIO_OUTBOUND chain"}
+	ExcludeInboundPort      = Command{"IPT03", "exclude inbound port from capture"}
+	IncludeInboundPort      = Command{"IPT04", "include inbound port for capture"}
+	InboundCapture          = Command{"IPT05", "redirect inbound request to proxy"}
+	KubevirtCommand         = Command{"IPT06", "Kubevirt outbound redirect"}
+	ExcludeInterfaceCommand = Command{"IPT07", "Excluded interface"}
+	UndefinedCommand        = Command{"IPT00", ""}
+)
+
+var IDToCommand = map[string]Command{
```
lambdai, line 33: Why don't you just put the raw string or abbreviation in the NFLOG rule? Even abbr is better than number

lambdai (commented):
Looks awesome!

It would be nice if you can split the 
-m 
-nflog
various refactor

The unit tests provides confidence though

tools/istio-iptables/pkg/config/config.go
howardjohn: This is just making it match the reality - the input flag is `kube-virt-inferaces`. but should probably make it KubeVirtInterfaces

tools/istio-iptables/pkg/log/command.go
howardjohn: There is a very very small length allowed IIRC. I can look into it more, maybe its possible

tools/istio-iptables/pkg/config/config.go
lambdai: I see...
Let's rename to `KubeVirtInterfaces` as well. IDE should help :)

tools/istio-iptables/pkg/log/command.go
lambdai: Feel free to ignore. No doubt this is far better than what we have nowadays.

howardjohn: LOG has small limit, 64 for NFLOG though. lets do it

howardjohn:
@lambdai ptal

lambdai (approved):
LGTM. Thank you!

morvencao (approved):

nmittler (approved):

jacob-delgado (approved):

manifests/charts/istio-control/istio-discovery/files/injection-template.yaml
```diff
@@ -310,6 +310,20 @@ spec:
       failureThreshold: {{ annotation .ObjectMeta `readiness.status.sidecar.istio.io/failureThreshold` .Values.global.proxy.readinessFailureThreshold }}
     {{ end -}}
     securityContext:
+      {{- if eq (index .ProxyConfig.ProxyMetadata "IPTABLES_TRACE_LOGGING") "true" }}
```
costinm, line 313: Not sure if IPTABLES_TRACE_LOGGING should have all this broad effect.

Maybe have a more precise setting that enables NET_ADMIN/coredump ?

tools/istio-iptables/pkg/cmd/root.go
```diff
@@ -373,6 +379,8 @@ func init() {
 
 	rootCmd.Flags().BoolP(constants.DryRun, "n", false, "Do not call any external dependencies like iptables")
 
+	rootCmd.Flags().Bool(constants.TraceLogging, false, "Insert tracing logs for each iptables rules, using the LOG chain.")
```
costinm, line 382: Would be great to use env variables for any new feature - flags have a lot of problems.

tools/istio-iptables/pkg/config/config.go
```diff
@@ -41,7 +41,7 @@ type Config struct {
 	OutboundPortsExclude    string        `json:"OUTBOUND_PORTS_EXCLUDE"`
 	OutboundIPRangesInclude string        `json:"OUTBOUND_IPRANGES_INCLUDE"`
 	OutboundIPRangesExclude string        `json:"OUTBOUND_IPRANGES_EXCLUDE"`
-	KubevirtInterfaces      string        `json:"KUBEVIRT_INTERFACES"`
+	KubeVirtInterfaces      string        `json:"KUBE_VIRT_INTERFACES"`
```
costinm, line 44: Is this backward compatible ? Did we ever use the old name ?

tools/istio-iptables/pkg/log/nflog.go
```diff
@@ -0,0 +1,117 @@
...
+package log
+
+import (
+	"context"
+	golog "log"
+	"net"
+	"os"
+	"time"
+
+	"github.com/florianl/go-nflog/v2"
```
costinm, line 24: New depenency and logger ? Or is this something specific to nflog ?

tools/istio-iptables/pkg/log/nflog.go
```diff
@@ -0,0 +1,117 @@
...
+	"IPTABLES_TRACE_LOGGING",
+	false,
+	"When enable, all iptables actions will be logged. "+
+		"This requires NET_ADMIN privilege and has noisy logs; as a result, this is intended for debugging only").Get()
+
+var iptablesTrace = log.RegisterScope("iptables", "trace logs for iptables", 0)
+
+// ReadNFLOGSocket reads from the nflog socket, sending output to logs.
+// This is intended for debugging only.
+func ReadNFLOGSocket(ctx context.Context) {
```
costinm, line 41: Is this that could be done by a standalone command or helper ( or iptables ), without bloating pilot-agent ?

howardjohn: I thought about making it in istioctl or something but its a bit tricky since it needs to run on the actual node. I think best case for standalone would be ephemeral containers but that isn't really ready unfortunately... It is a bit nice to have the access logs + iptables logs side by side in some cases

tools/istio-iptables/pkg/log/nflog.go
howardjohn: its an nflog specific thing. Basically nflog is serving some logs over a uds socket so the library is just a bit of code to read from this. We use the standard Istio logger to write out the data read by the library

tools/istio-iptables/pkg/config/config.go
howardjohn: The config is incredibly confusing but I am almost certain the json tags here are just for logging. The real config uses `viper` to read, which has always had the separator https://github.com/istio/istio/blob/4ca427a5f1483e4280dd8c7f45c83ff0d7eca81a/tools/istio-iptables/pkg/constants/constants.go#L91.

It would probably be good to clean the config up here, as I cannot wrap my head around it easily

tools/istio-iptables/pkg/cmd/root.go
howardjohn: It is using viper, so has env var + flag. I don't have a strong preference to one or the other but wanted to be consistent with the other configs

tools/istio-iptables/pkg/log/nflog.go
costinm: Given the direction is to reduce permissions and isolate network - I'm not so sure. But it's not a huge dependency, compared with the others.

tools/istio-iptables/pkg/config/config.go
costinm: If we are sure it was never used - sounds good. It's the type of changes that may have subtle upgrade breakages if someone ever had to set it.

tools/istio-iptables/pkg/cmd/root.go
costinm: Sounds good, at some point we should kill viper and just use env - all other feature flags are env.

Issue #33906, opened 2021-07-08
integration test support for k8s CSR


irisdingbj:
Add a component into our integration test framework which will contain a controller to reconsile k8s CSR and sign certs with full cert-chain

zhlsunshine:
all related PRs have been merged, so I think it can be closed.

PR #34207, opened 2021-07-21
integrate CSR controller to integration test framework for Istio

zhlsunshine:
Work on [issue#33906](https://github.com/istio/istio/issues/33906)

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


pkg/test/framework/components/csrctrl/README.md
```diff
@@ -0,0 +1,66 @@
...
+Check the binary file under the folder of `istio/out/linux_amd64/`, then build the image based on it:
+
+```console
+$ make docker.csrctrl
+```
+
+User can find the CSR controller image by `docker images`, and see the $HUB/csrctrl with tag value $TAG.
+
+## How to deploy CSR controller in Kubernetes cluster in test code?
+
```
irisdingbj, line 24: Can we add  a section to notify users that it only supports signer from `istio.io/`?

pkg/test/framework/components/csrctrl/integration-ca.yaml
```diff
@@ -0,0 +1,186 @@
...
+  - secrets
+  verbs:
+  - create
+  - patch
+  - delete
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
irisdingbj, line 69: This is not needed for us as istio will be responsible for authorization.

pkg/test/framework/components/csrctrl/integration-ca.yaml
```diff
@@ -0,0 +1,186 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: signer-ca-manager-role
+subjects:
+- kind: ServiceAccount
+  name: default
+  namespace: sig-ca-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
irisdingbj, line 113: this is not needed either, WDYT?

pkg/test/framework/components/csrctrl/integration-ca.yaml
```diff
@@ -0,0 +1,186 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: signer-ca-proxy-role
+subjects:
+- kind: ServiceAccount
+  name: default
+  namespace: sig-ca-system
+---
+apiVersion: v1
```
irisdingbj, line 126: Do we need a Service? Suppose only the deployment is enough.

pkg/test/framework/components/csrctrl/integration-ca.yaml
```diff
@@ -0,0 +1,186 @@
...
+        name: manager
+        resources:
+          limits:
+            cpu: 100m
+            memory: 30Mi
+          requests:
+            cpu: 100m
+            memory: 20Mi
+        volumeMounts:
+        - mountPath: /tmp/pki/signer/istio.io/foo/
```
irisdingbj, line 177: Is it possible to remove this?

irisdingbj:
Thanks @zhlsunshine, this will make us support  to  k8s CSR conveniently.

pkg/test/framework/components/csrctrl/README.md
zhlsunshine: Sure

pkg/test/framework/components/csrctrl/kube.go
```diff
@@ -0,0 +1,111 @@
...
+	_ io.Closer = &kubeComponent{}
+)
+
+type kubeComponent struct {
+	id      resource.ID
+	ns      namespace.Instance
+	cluster cluster.Cluster
+}
+
+func newKube(ctx resource.Context, cfg Config) (Instance, error) {
```
howardjohn, line 47: I think we do not need a new component. This will make the test slower, and its more complex to maintain.

Instead, the test can just run the CSR controller directly from the go test code. The go test already has access to the cluster of course, and its just acting on k8s resources, so there is no need to run in cluster.

howardjohn (changes requested):

pkg/test/framework/components/csrctrl/kube.go
irisdingbj: Hi John,  suppose we should have multiple test that will use this CSR controller to sign cert back. If we put this into test code, it means every test needs to write this logic, right? using current way, it will make it a pluggable mode that each test can use it easily.    This will deployed as a seperate pod and will sign cert for the signer you defined.  Can you please explain why this will make test slower? Thanks.

zhlsunshine:
Hi @howardjohn & @irisdingbj, based on your discussion, I did some changes below:

- no binary generation and image build for CSR controller
- removed the new component `csrctrl` for CSR controller
- expose an interface named `RunCSRController` for user to start the CSR controller

And thank you for your reply and suggestion!

zhlsunshine:
Hi @howardjohn,  according to discussion [above](https://github.com/istio/istio/pull/34207#discussion_r674878283),  there is a API named `RunCSRController`, however,  we have encountered 2 problems for it. 
1. This API need signing permission to sign CSR, but how should it be assigned the permission if it's just a goroutine?
2. This API need to be running as a signing server in a pod, but which pod should it be placed in the test framework during test?
Based on current situation,  it seem that the CSR controller does not look like generic function or API which can be called one time and need no permission.  It may be proper to introduce a new component in test framework, any idea from you? Thanks!

pkg/test/csrctrl/controllers/start_csrctrl.go
```diff
@@ -0,0 +1,91 @@
...
+	_              = corev1.AddToScheme(scheme)
+)
+
+func RunCSRController(signerNames string) {
+	// Config Istio log
+	if err := log.Configure(loggingOptions); err != nil {
+		log.Infof("Unable to configure Istio log error: %v", err)
+		os.Exit(-1)
+	}
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
```
irisdingbj, line 53: Need to reuse the same k8s client

pkg/test/csrctrl/controllers/start_csrctrl.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	if err := (&CertificateSigningRequestSigningReconciler{
+		Client:        mgr.GetClient(),
+		SignerRoot:    signerRoot,
+		CtrlCertTTL:   certificateDuration,
+		Scheme:        mgr.GetScheme(),
+		SignerNames:   arrSingers,
+		Signers:       signersMap,
+		EventRecorder: mgr.GetEventRecorderFor("CSRSigningReconciler"),
```
irisdingbj, line 80: please remove the eventRecorder

pkg/test/csrctrl/signer/signer.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	rootCerts, err := util.AppendRootCerts(der, s.caProvider.caLoader.CertFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to append root certificates (%v)", err)
+	}
+
+	return pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: rootCerts}), nil
```
irisdingbj, line 66: this needs to change to pem format

howardjohn (commented):
Changes look great, thanks!

@shankgan can you take a look?

shankgan:
> Changes look great, thanks!
> 
> @shankgan can you take a look?

LGTM those parts that I understand. The csrcontroller is invoked in the test scripts,  will these test scripts have the permission to read and sign the csr requests?

zhlsunshine (commented):
> > Changes look great, thanks!
> > @shankgan can you take a look?
> 
> LGTM those parts that I understand. The csrcontroller is invoked in the test scripts, will these test scripts have the permission to read and sign the csr requests?

Yes, @shankgan, that's a question,  in fact, I create a new component named `csrctrl` to do this in my previous commits, including the component's build binary and image, then deployment. In new component, we can contains the permission etc, but for now, we need to know how to get the permission to list, watch, patch and sign CSR. So I also recommend that this should be implemented as a new component. Any idea from you @howardjohn & @shankgan? Thanks for your time!

howardjohn:
> LGTM those parts that I understand. The csrcontroller is invoked in the test scripts, will these test scripts have the permission to read and sign the csr requests?

The person running tests needs to be cluster admin generally. If they cannot sign CSR requests they can skip this test.

Putting it in cluster doesn't help, it actually makes it worse because you would need permission to give permissions, which is more privileged than signing CSR

zhlsunshine:
> > LGTM those parts that I understand. The csrcontroller is invoked in the test scripts, will these test scripts have the permission to read and sign the csr requests?
> 
> The person running tests needs to be cluster admin generally. If they cannot sign CSR requests they can skip this test.
> 
> Putting it in cluster doesn't help, it actually makes it worse because you would need permission to give permissions, which is more privileged than signing CSR

Hi @howardjohn, the CSR controller privilege is under the control by clusterrole in yaml, only include necessary privilege, looks like below:

```
- apiGroups:
  - certificates.k8s.io
  resources:
  - certificatesigningrequests
  verbs:
  - get
  - list
  - watch
  - patch
- apiGroups:
  - certificates.k8s.io
  resources:
  - certificatesigningrequests/status
  verbs:
  - update
- apiGroups:
  - certificates.k8s.io
  resourceNames:
  - istio.io/*
  resources:
  - signers
  verbs:
  - sign
  ```

File: pkg/test/csrctrl/authority/policies.go [hunk 10]
```diff
@@ -0,0 +1,140 @@
...
+	// Usages are the allowed usages of a certificate.
+	Usages []capi.KeyUsage
+}
+
+func (p PermissiveSigningPolicy) apply(tmpl *x509.Certificate) error {
+	usage, extUsages, err := keyUsagesFromStrings(p.Usages)
+	if err != nil {
+		return err
+	}
+	tmpl.KeyUsage = usage
```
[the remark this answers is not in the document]
zhlsunshine: Hi @shankgan,  the usage has been validated in before calling function `keyUsagesFromStrings` and the line range is betwee 106 and 112. Thank you!

irisdingbj (approved):
Thanks @zhlsunshine changes LGTM now. wating for @howardjohn @shankgan 's final nod.

irisdingbj:
Using the CSRcontroller provided in this PR,  it successds sign cert back in our integration test.

zhlsunshine:
> Using the CSRcontroller provided in this PR, it successds sign cert back in our integration test.

@irisdingbj Thank you for your effort for such integration! @howardjohn could you please help to approve this PR if possible? Thanks!

zhlsunshine:
> Using the CSRcontroller provided in this PR, it successds sign cert back in our integration test.

@irisdingbj Thank you for your effort for such integration! @howardjohn & @shankgan  could you please help to approve this PR if possible? Thanks!

shankgan (approved):

howardjohn (approved):

PR #34217, opened 2021-07-21
Allow disabling tracing

howardjohn:
This builds on https://github.com/istio/istio/pull/34216, should merge that first to reduce size of this.

Currently, there is no way to disable tracing view istioctl. This allows disabling it.

This is important for users that are concerned about the overhead of adding tracing by default (especially DNS), as well as for migration to Telemetry API.


douglas-reid (approved):
Looks reasonable to me.

Monkeyanator (approved):

melichron:
Any instrustuctions didn't works for me, but after all I disabled zipkin by explicity add podAnnotation to every pod with sidecar and istio gateways:
```
        podAnnotations:
          proxy.istio.io/config: |
            tracing: {}
```
That will overwrite any upstream settings.

PR #34228, opened 2021-07-21
Add validation for Telemetry

howardjohn:
This add validation logic for new Telemetry API. This also fixes the webhook test to properly detect that there was no validation logic, and switch to `go:embed` to avoid excessive generated code


pkg/config/validation/validation.go
```diff
@@ -3308,3 +3325,126 @@ func validateNetwork(network *meshconfig.Network) (errs error) {
...
+				}
+			}
+		}
+	}
+	return
+}
+
+func validateTelemetryMetrics(metrics []*telemetry.Metrics) (v Validation) {
+	if len(metrics) > 1 {
+		v = appendWarningf(v, "multiple metrics is not currently supported")
```
douglas-reid, line 3404: do we want this restriction to start?  unlike tracing, for metrics we can have multiple providers at the same time (prom and stackdriver).

howardjohn: I intentionally made this a warning, not an error. So it won't block; same with tracing and logging.

I guess for metrics it currently supports *nothing* (not yet implemented though), and plausibly the initial implementation will support multiple providers at the same time, so maybe this one doesn't make sense. if anything, we could add a warning that Telemetry is not even implemented yet

douglas-reid: fair enough. we'll want to add addressing this warning in the alpha promotion plan.

tests/integration/pilot/validation_test.go
```diff
@@ -78,10 +78,8 @@ func TestValidation(t *testing.T) {
 				// We are only checking the string literals of the rejection reasons
 				// from the webhook and the k8s api server as the returned errors are not
 				// k8s typed errors.
-				return strings.Contains(err.Error(), "denied the request") ||
-					strings.Contains(err.Error(), "error validating data") ||
-					strings.Contains(err.Error(), "Invalid value") ||
-					strings.Contains(err.Error(), "is invalid")
+				// Note: this explicitly does NOT catch OpenAPI schema rejections - only validating webhook rejections.
+				return strings.Contains(err.Error(), "denied the request")
```
Monkeyanator, line 82: surprised removing these didn't cause other validation cases to start failing (or else why were they there in the first place?)

pkg/config/validation/validation.go
```diff
@@ -3308,3 +3327,134 @@ func validateNetwork(network *meshconfig.Network) (errs error) {
...
+		v = appendValidation(v, validateTelemetryProviders(l.Providers))
+		if l.RandomSamplingPercentage.GetValue() < 0 || l.RandomSamplingPercentage.GetValue() > 100 {
+			v = appendErrorf(v, "randomSamplingPercentage must be in range [0.0, 100.0]")
+		}
+		for name, tag := range l.CustomTags {
+			if name == "" {
+				v = appendErrorf(v, "tag name may not be empty")
+			}
+			switch t := tag.Type.(type) {
+			case *telemetry.Tracing_CustomTag_Literal:
```
Monkeyanator, line 3386: Any additional restrictions on `CustomTag` value we can validate for? Or all strings make valid tags?

tests/integration/pilot/validation_test.go
howardjohn: I am 90% the other ones were (incorrectly) added to catch open api schema rejections - when we explicitly only want webhook rejections. The 10% chance is that other versions of kubernetes had different error message...

pkg/config/validation/validation.go
howardjohn: I am pretty sure its a CEL expression. It could be nice to validate that, but i am not sure the details on that. @kyessenov @douglas-reid any thoughts?

douglas-reid: Here, `Literal` represents a static value that gets added to each span. The other two choices are an environment variable and a request header.

howardjohn: oh I was thinking of MetricsOverrides_TagOverride.Value which is a CEL expression?

douglas-reid: yes.

kyessenov: We should be parsing all the various languages here (custom targs, access logs, CEL). They are highly error-prone at syntax level. cel-go can be used for validation, but the others one need new parsers in golang.

howardjohn: What is the format for custom tags?

Access logs we don't currently support format in the API so we can defer that I think.

This PR is already fairly large so may want to do this in a followup if we need a new library

stevenctl (approved):

stevenctl:
lgtm, but allowing others more familiar with the API to approve before merge

Issue #34192, opened 2021-07-21
TestRemoteClusters test flake


howardjohn:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/34144/integ-pilot-multicluster-tests_istio/1417642057895251968

Looking at istiod log (https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/34144/integ-pilot-multicluster-tests_istio/1417642057895251968/artifacts/pilot-0614cc9aea2143a692ea49d7b/TestRemoteClusters/primary/_test_context/istio-state412117859/primary/istiod-5c4d49d975-dqf8x_discovery.log):
```
472:2021-07-21T00:40:59.065367Z info    Processing add: istio-system/istio-remote-secret-cross-network-primary
473:2021-07-21T00:40:59.067298Z info    Processing add: istio-system/istio-remote-secret-cross-network-primary
474:2021-07-21T00:40:59.067457Z info    secret controller got event from queue for secret istio-system/istio-remote-secret-cross-network-primary
475:2021-07-21T00:40:59.067471Z info    processing secret event for secret istio-system/istio-remote-secret-cross-network-primary
476:2021-07-21T00:40:59.067482Z info    Adding cluster cross-network-primary from secret istio-system/istio-remote-secret-cross-network-primary
482:2021-07-21T00:40:59.101448Z info    secret controller got event from queue for secret istio-system/istio-remote-secret-cross-network-primary
483:2021-07-21T00:40:59.101475Z info    processing secret event for secret istio-system/istio-remote-secret-cross-network-primary
484:2021-07-21T00:40:59.101527Z info    Adding cluster cross-network-primary from secret istio-system/istio-remote-secret-cross-network-primary=
```

they look reasonable

howardjohn:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/33898/integ-pilot-multicluster-tests_istio/1417618101368786944

howardjohn:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/33245/integ-pilot-multicluster-tests_istio/1417609501795684352

PR #34229, opened 2021-07-21
fix TestRemoteClusters flake

stevenctl:
It seems like `primary` is showing the results from `cross-network-primary`'s istiod... I captured the loop var in a new closure so I'm not sure why that's happening. Switching away from `RunParallel` to see if it fixes https://github.com/istio/istio/issues/34192


howardjohn (approved):

howardjohn:
lgtm ,can you retest a few times to make sure its not still happening?

Issue #34227, opened 2021-07-21
`concurrent map read and map write` in Istiod


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/integ-cni-k8s-tests_istio_postsubmit/1417923144731070464

howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-120_istio_postsubmit/1417923115362553856

howardjohn:
```
==================
WARNING: DATA RACE
Write at 0x00c001527a40 by goroutine 390:
  runtime.mapdelete_faststr()
      runtime/map_faststr.go:297 +0x0
  istio.io/istio/pilot/pkg/networking/util.MaybeUpdateTLSModeLabel()
      istio.io/istio/pilot/pkg/networking/util/util.go:548 +0x44f
  istio.io/istio/pilot/pkg/xds.(*EndpointBuilder).buildLocalityLbEndpointsFromShards()
      istio.io/istio/pilot/pkg/xds/endpoint_builder.go:334 +0x15ca
  istio.io/istio/pilot/pkg/xds.(*DiscoveryServer).llbEndpointAndOptionsForCluster()
      istio.io/istio/pilot/pkg/xds/eds.go:309 +0x257
  istio.io/istio/pilot/pkg/xds.(*DiscoveryServer).generateEndpoints()
      istio.io/istio/pilot/pkg/xds/eds.go:313 +0xa4
  istio.io/istio/pilot/pkg/xds.(*EdsGenerator).Generate()
      istio.io/istio/pilot/pkg/xds/eds.go:405 +0x6bb
  istio.io/istio/pilot/pkg/xds.(*DiscoveryServer).pushXds()
      istio.io/istio/pilot/pkg/xds/xdsgen.go:101 +0x2b6
  istio.io/istio/pilot/pkg/xds.(*DiscoveryServer).processRequest()
      istio.io/istio/pilot/pkg/xds/ads.go:236 +0x36f
  istio.io/istio/pilot/pkg/xds.(*DiscoveryServer).Stream()
      istio.io/istio/pilot/pkg/xds/ads.go:311 +0xa5c
  istio.io/istio/pilot/pkg/xds.(*DiscoveryServer).StreamAggregatedResources()
      istio.io/istio/pilot/pkg/xds/ads.go:241 +0x4c
[...]

PR #34231, opened 2021-07-21
Disable EnableAutomTLSCheckPolicies due to race condition

howardjohn:
See https://github.com/istio/istio/issues/34227

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -147,7 +147,7 @@ var (
 		"Skip validating the peer is from the same trust domain when mTLS is enabled in authentication policy").Get()
 
 	EnableAutomTLSCheckPolicies = env.RegisterBoolVar(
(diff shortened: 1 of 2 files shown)


incfly (approved):
might need to add this ENABLE_AUTO_MTLS_CHECK_PICKES: true to the integration tests main_test.go.

howardjohn:
I do not think we want it running in tests, it has a race condition that will crash the istiod and make tests flake (which is how I found it)

howardjohn:
Ah... but I need to disable the test that depends on it

incfly:
sure. then you can comment out the `automtls-*-workload.yaml` test introduced in the original PR.

incfly (approved):

adiprerepa (approved):

PR #34240, opened 2021-07-22
Bump HPA resource version to autoscaling/v2beta2

### LINKED ISSUE #32005: Update HorizontalPodAutoscaler to autoscaling/v2beta1
`W0407 10:54:12.530768 2443155 warnings.go:70] autoscaling/v2beta1 HorizontalPodAutoscaler is deprecated in v1.22+, unavailable in v1.25+; use autoscaling/v2beta2 HorizontalPodAutoscaler`

kebe7jun:
Fixes https://github.com/istio/istio/issues/32005

[X] Installation


howardjohn (commented):
For helm - LGTM

For istioctl - how does this interact with KubernetesResourcesSpec (which embeds v2beta1 API) and arbitrary overlay that we support? I noticed no tests fail which makes me think we have lack of test coverage here maybe?

morvencao (commented):
you should also update this at least https://github.com/istio/istio/blob/master/operator/pkg/helmreconciler/prune.go#L45

kebe7jun:
TODO: update v1alpha1.ResourceMetricSource struct in https://github.com/istio/api/blob/36794c332d018290fdbc9441f36c6103cb36aabf/operator/v1alpha1/operator.proto#L539-L542

It was beacause the v2beta2.HPA's struct is not same as v2beta2.HPA.

BTW, if we not use hpaSpec in profile, this fix will work well.

ostromart:
Apart from changing the code and proto, we should also update the tests e.g. 
https://github.com/istio/istio/blob/master/operator/cmd/mesh/testdata/manifest-generate/input/pilot_k8s_settings.yaml#L24

howardjohn:
Not stale, this is still high priority

ostromart:
Talked about this with @richardwxn and the solution would be the following:
- update the templates to use the new HPA. may need two versions if we need backwards compatibility
- update our copy of the HPA schema to include the new field. This will cover both versions and even though validation will pass if the wrong version is used, k8s will complain upon apply which seems like a reasonable tradeoff.

Issue #34179, opened 2021-07-20
CVE-2017-11468 in docker/distribution triggers security scanners


brendarearden:
**Bug description**

After reaching out to the security team, I was given to go-ahead to open this issue. 
|Library| Vulnerability ID | Severity | Installed Version | Fixed Version | Title | 
|-----------|---------|-------|--------|---------|----------|
|github.com/docker/distribution|CVE-2017-11468|HIGH|v0.0.0-20191216044856-a8371794149d|v2.7.0-rc.0+incompatible|docker-distribution: Does not properly restrict the amount of content accepted from a...-->avd.aquasec.com/nvd/cve-2017-11468|

This CVE does not impact the security of Istio and the portion of the library that can be exploited is not in use by this repo, meaning Istio has a very distant transitive dependency on this. 

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**
No HIGH or CRITICAL CVEs should be present in libraries used

**Steps to reproduce the bug**
Scan most recent istioctl image ( version istio/istioctl:1.11.0-beta.0 was used for this example) 
See HIGH severity CVE result `CVE-2017-11468`

liminw:
cc. @myidpt

PR #34244, opened 2021-07-22
bump github.com/docker/distribution version to v2.7.1+incompatible

kebe7jun:
Fix https://github.com/istio/istio/issues/34179

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,8 +7,7 @@ replace github.com/spf13/viper => github.com/istio/viper v1.3.3-0.20190515210538
 // Old version had no license
 replace github.com/chzyer/logex => github.com/chzyer/logex v1.1.11-0.20170329064859-445be9e134b2
 
-// Avoid pulling in incompatible libraries
-replace github.com/docker/distribution => github.com/docker/distribution v0.0.0-20191216044856-a8371794149d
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -7,8 +7,7 @@ replace github.com/spf13/viper => github.com/istio/viper v1.3.3-0.20190515210538
 // Old version had no license
 replace github.com/chzyer/logex => github.com/chzyer/logex v1.1.11-0.20170329064859-445be9e134b2
 
-// Avoid pulling in incompatible libraries
-replace github.com/docker/distribution => github.com/docker/distribution v0.0.0-20191216044856-a8371794149d
+replace github.com/docker/distribution => github.com/docker/distribution v2.7.1+incompatible
```
howardjohn, line 10: I have some concerns about blindly replacing with a new library, and how it may impact dependencies that user other versions. It seems ideally we would get the downstream projects to update? Seems like we mostly need https://github.com/helm/helm/pull/9675

howardjohn: Opened https://github.com/kubernetes/kubernetes/issues/103849 as well.

In particular, I tried changing helm to use `v2.7.1+incompatible`, and it no longer passes their tests. This makes me concerned we may be breaking some functionality depended on

kebe7jun: fine, I close this.

PR #34247, opened 2021-07-22
move select virtual services logic to sidecar

ramaraochavali:
Merges selectVirtualServices logic in http route to IstioEgressListeners selectVirtualServices
Few more minor refactors

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


ramaraochavali:
@hzxuzhonghu can you PTAL?

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -256,7 +215,7 @@ func BuildSidecarOutboundVirtualHosts(node *model.Proxy, push *model.PushContext
 		if listenerPort == 0 {
 			// Take all ports when listen port is 0 (http_proxy or uds)
 			// Expect virtualServices to resolve to right port
-			servicesByName[svc.Hostname] = svc
+			servicesByName[svc.Hostname] = svc.DeepCopy() // To ensure further changes does not affect service registry.
```
hzxuzhonghu, line 218: why?

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -291,6 +245,38 @@ func BuildSidecarOutboundVirtualHosts(node *model.Proxy, push *model.PushContext
 		}
 	}
 
+	buildVirtualHost := func(hostname string, vhwrapper istio_route.VirtualHostWrapper, svc *model.Service) *route.VirtualHost {
```
hzxuzhonghu, line 248: seems unrelated change

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -271,11 +230,6 @@ func BuildSidecarOutboundVirtualHosts(node *model.Proxy, push *model.PushContext
 		}
 	}
 
-	// This is hack to keep consistent with previous behavior.
-	if listenerPort != 80 {
-		// only select virtualServices that matches a service
-		virtualServices = selectVirtualServices(virtualServices, servicesByName)
```
hzxuzhonghu, line 277: `IstioEgressListenerWrapper.VirtualServices` is not equal to this selection, this is port related

pilot/pkg/networking/core/v1alpha3/httproute.go
ramaraochavali: This is not needed for this PR. But in general added it fo consistency with else block. If you think it is not needed, 
I will remove it

pilot/pkg/networking/core/v1alpha3/httproute.go
ramaraochavali: As i mentioned in PR - I did few more refactoring to simplify the code

pilot/pkg/networking/core/v1alpha3/httproute.go
ramaraochavali: I added that logic inside filterVirtualServices

pilot/pkg/networking/core/v1alpha3/httproute_test.go
```diff
@@ -862,7 +862,6 @@ func TestSidecarOutboundHTTPRouteConfig(t *testing.T) {
 				"test-private.com:80": {
 					"test-private.com": true, "test-private.com:80": true, "9.9.9.9": true, "9.9.9.9:80": true,
 				},
-				"match-no-service.not-default:80": {"match-no-service.not-default": true, "match-no-service.not-default:80": true},
```
ramaraochavali, line 865: @hzxuzhonghu This test case correctly fails when we select virtual services with matching services

howardjohn: Why do we want this to change? I don't think a VS is required to match a service?

howardjohn (commented):
What are the impacts on the generated XDS configuration here? Its hard to grok 600 line change, want to make sure this is not impacting users

ramaraochavali:
> What are the impacts on the generated XDS configuration here? Its hard to grok 600 line change, want to make sure this is not impacting users

This should not impact any XDS config change. This is a fair point. Let me move plain refactorings to a different PR so that it is easy to review this PR. Sorry about that.

pilot/pkg/networking/core/v1alpha3/httproute_test.go
ramaraochavali: https://github.com/istio/istio/pull/34247 - This change if VS host does not find a service, we drop it.

pilot/pkg/networking/core/v1alpha3/httproute.go
hzxuzhonghu: https://github.com/istio/istio/blob/77366e7a2b077d426df043669f09fe23d4a22c7e/pilot/pkg/networking/core/v1alpha3/httproute.go#L254-L277 

You can see here it selects the services that has specific port, but i do not see there is port related logic in sidecar

hzxuzhonghu (changes requested):
As i said they are not equal, TestSelectVirtualService should be fixed

pilot/pkg/networking/core/v1alpha3/httproute_test.go
howardjohn: The link is just to this PR, was it meant to be somewhere else?

I am a bit concerned we are making a backwards incompatible change.

ramaraochavali: https://github.com/istio/istio/pull/33764 - Sorry this PR has changed that.

howardjohn: Is that part actually merged?

> If you have concern of breaking some compatibility, we can leave 1, and probihit 2.

seems like it wasn't? Or I do not understand it :slightly_smiling_face:

pilot/pkg/model/sidecar.go
```diff
@@ -634,6 +642,50 @@ func (ilw *IstioEgressListenerWrapper) selectVirtualServices(virtualServices []c
 	return importedVirtualServices
 }
 
+// filterVirtualServices selects the virtual services by matching given services' host names.
+func (ilw *IstioEgressListenerWrapper) filterVirtualServices(virtualServices []config.Config, servicesByName map[host.Name]*Service) []config.Config {
+	// This is hack to keep consistent with previous behavior.
+	if ilw.IstioListener != nil && ilw.IstioListener.Port != nil && ilw.IstioListener.Port.Number == 80 {
```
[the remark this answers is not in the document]
hzxuzhonghu: yes, this is not enough. Say we want to select virtualservice for route `8080`, first we need to get only the services that has port `8080`. And then select the vs according to the services that has 8080.

But in this patch, it selects the vses by all the services, am i misreading?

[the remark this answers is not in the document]
ramaraochavali: That is what this patch does. This selects services based on egress listener config and then apply this filter logic. Please check one more time - If I am missing some flow.

pilot/pkg/networking/core/v1alpha3/httproute_test.go
ramaraochavali: I am confused now. @hzxuzhonghu Is n't this expected if we select VS that only have services in the registry? Basically your fix in #33674 ?

pilot/pkg/model/sidecar.go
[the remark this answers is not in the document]
hzxuzhonghu: Maybe i missed， can you point me where the service port is matched? I mean not this hacky 80 port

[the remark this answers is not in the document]
ramaraochavali: https://github.com/istio/istio/blob/b30f8ab30ab35e6e5608d2c407a17230d6d4bf65/pilot/pkg/model/sidecar.go#L679 - This is an existing code. Egresslistener matching services by port and adds them

pilot/pkg/networking/core/v1alpha3/httproute_test.go
hzxuzhonghu: As i said for route 80, it is hacky, and there is a integration test that depend on the behavior. So i kept it.

pilot/pkg/model/sidecar.go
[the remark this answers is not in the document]
hzxuzhonghu: Not sure I understand how this is related to my question? Think about when there is a default sidecar.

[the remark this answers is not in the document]
ramaraochavali: I think there is some confusion here. This does not change the behaviour of your patch. Your fix of selectVirtualServices is getting services and virtualServices from egressListenerWrapper only https://github.com/istio/istio/blob/bd9ecccaaa83b0fd6a388855d17b313b0f19b627/pilot/pkg/networking/core/v1alpha3/httproute.go#L242 - This patch also is doing the same. Are some thing is missed? I am not able to understand the difference.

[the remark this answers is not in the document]
ramaraochavali: I see what you mean. Let me fix that

pilot/pkg/networking/core/v1alpha3/httproute_test.go
ramaraochavali: Ok. I fixed and this has not changed

ramaraochavali:
@hzxuzhonghu fixed PTAL

hzxuzhonghu:
Thanks for the fix, will be on my way home. Will take a look tomorrow

pilot/pkg/model/sidecar.go
```diff
@@ -634,6 +636,46 @@ func (ilw *IstioEgressListenerWrapper) selectVirtualServices(virtualServices []c
 	return importedVirtualServices
 }
 
+// filterVirtualServices selects the virtual services by matching given services' host names.
+func (ilw *IstioEgressListenerWrapper) filterVirtualServices(virtualServices []config.Config, servicesByName map[host.Name]*Service) []config.Config {
```
hzxuzhonghu, line 640: @ramaraochavali This is still a separate function from that one.

ramaraochavali: Yeah. I am wondering how to combine this? Because it is dependent on servicesByPort chosen for that particular RDS. Any suggestions?

ramaraochavali:
@hzxuzhonghu This should be ready. Now I refactored a bit so that we do select virtual services as part of Listener wrapper creation - not in the same `selectVirtualServices` method because after virtual services are selected we add few more services for matching vs destination. Now this is done once per listener conversion rather than every route gen. PTAL

hzxuzhonghu:
I am thinking of a more simple way. Could show you tomorrow, hope it works

ramaraochavali:
@hzxuzhonghu changed. PTAL

pilot/pkg/model/sidecar.go
```diff
@@ -457,20 +456,17 @@ func convertIstioListenerToWrapper(ps *PushContext, configNamespace string,
...
-	out.virtualServices = out.selectVirtualServices(vses, out.listenerHosts)
+	out.virtualServices = SelectVirtualServices(vses, out.listenerHosts)
 	svces := ps.Services(&dummyNode)
 	out.services = out.selectServices(svces, configNamespace, out.listenerHosts)
 
 	return out
 }
 
-// Services returns the list of services imported across all egress listeners by this
-// Sidecar config
```
hzxuzhonghu, line 468: why remove

pilot/pkg/model/virtualservice_test.go
```diff
@@ -18,17 +18,26 @@ import (
...
 	"istio.io/istio/pkg/config"
+	"istio.io/istio/pkg/config/host"
+	"istio.io/istio/pkg/config/protocol"
 	"istio.io/istio/pkg/config/schema/collections"
 	"istio.io/istio/pkg/config/visibility"
 )
 
+const wildcardIP = "0.0.0.0"
+
+var tnow = time.Now()
```
hzxuzhonghu, line 39: nit：move down to the function

hzxuzhonghu (commented):
lgtm, will approve later

PR #34254, opened 2021-07-22
Switch our direct dependency from docker/distribution to v3

howardjohn:
See https://github.com/kubernetes/kubernetes/issues/103849 for why

We still have a lot of indirect deps on the old one, but hopefully when or if those
move we can finally drop it


howardjohn:
cc @nmittler

nmittler (approved):

PR #34257, opened 2021-07-22
Filter container with exit code 0 from crashed container.

bianpengyuan:
Got a case where test log suggests container crash, but pod state and log does not have any indication about that: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/34202/integ-cni-k8s-tests_istio/1418257862072733696 feels like there might be a race between resource clean up and container crash detection, but before digging into that it would be great to print more terminate state and understand if it is just because container exited normally. We might want to filter out container with exit code 0 from crashed container.

### DIFF
--- a/pkg/test/kube/dump.go
+++ b/pkg/test/kube/dump.go
@@ -214,13 +214,13 @@ func containerRestarts(pod corev1.Pod, container string) int {
 	return 0
 }
 
-func containerCrashed(pod corev1.Pod, container string) bool {
+func containerCrashed(pod corev1.Pod, container string) (bool, *corev1.ContainerStateTerminated) {
 	for _, cs := range pod.Status.ContainerStatuses {
-		if cs.Name == container {
-			return cs.State.Terminated != nil
+		if cs.Name == container && cs.State.Terminated != nil {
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

howardjohn:
lgtm. fwiw https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/34202/integ-cni-k8s-tests_istio/1418257862072733696/artifacts/pilot-cni-d56274045c2b43038fa33/_suite_context/istio-state-162149128/cluster-0/istio-cni-node-wqzd5_pod-events.yaml and https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/34202/integ-cni-k8s-tests_istio/1418257862072733696/artifacts/pilot-cni-d56274045c2b43038fa33/_suite_context/istio-state-162149128/cluster-0/istio-cni-node-wqzd5_pod-state.yaml show more info

howardjohn:
Yeah that is probably fine, assuming K8s restart policy is based on exit
code != 0? We want to detect if it would crash loop , not if it terminated
I think

PR #34259, opened 2021-07-22
Revert "[release-1.10] istio-agent csr flow: add api for caClient plugins to explicitly provide CA TrustAnchors if available"

sergii-ssh:
Reverts istio/istio#34226


howardjohn (approved):
This PR is co-dependant on https://github.com/istio/istio/pull/34255/files, which is likely not being backported (unless it goes through exception process). So revert the backport of this one

Monkeyanator (approved):

PR #34262, opened 2021-07-22
Move external control plane test to it's own istiodremote directory 

ericvn:
Move the test to it's own directory. The test will use a new topology json which removes the changes in the test to the multicluster json.

Edit: Actually copying the files to a new directory with updates. A test-infra PR is needed to test them, making it optional until the tests pass, and then we can make it required. A follow on PR here to delete the files.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


prow/config/topology/external-istiod.json
```diff
@@ -0,0 +1,32 @@
...
+    "network": "network-1",
+    "primaryClusterName": "external",
+    "configClusterName": "config",
+    "meta": {
+      "fakeVM": false
+    }
+  },
+  {
+    "kind": "Kubernetes",
+    "clusterName": "additional",
```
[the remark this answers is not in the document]
ericvn: @frankbu  mentioned `remote` although since it is on another network, `cross-network` might be better.

prow/config/topology/external-istiod.json
```diff
@@ -0,0 +1,32 @@
...
+  {
+    "kind": "Kubernetes",
+    "clusterName": "config",
+    "podSubnet": "10.20.0.0/16",
+    "svcSubnet": "10.255.20.0/24",
+    "network": "network-1",
+    "primaryClusterName": "external",
+    "configClusterName": "config",
+    "meta": {
+      "fakeVM": false
```
frankbu, line 20: What's this? Do we need it?
cc @stevenctl

tests/integration/istiodremote/base/main_test.go
```diff
@@ -0,0 +1,117 @@
...
+            mountPath: /var/lib/istio/inject
+      env:
+      - name: INJECTION_WEBHOOK_CONFIG_NAME
+        value: "istio-sidecar-injector-istio-system"
+      - name: VALIDATION_WEBHOOK_CONFIG_NAME
+        value: "istio-istio-system"
+      - name: EXTERNAL_ISTIOD
+        value: "true"
+      - name: CLUSTER_ID
+        value: config
```
frankbu, line 97: Do we need this? Hasn't the framework set this one already, from the topology?

prow/config/topology/external-istiod.json
stevenctl: I would put this on the `remote` cluster.

stevenctl: And maybe on the `external` cluster or this `cluster` as well. It only makes sense to have the fake VM for the "primary" or "config".

tests/integration/istiodremote/base/main_test.go
```diff
@@ -0,0 +1,117 @@
...
+    operatorManageWebhooks: true
+    configValidation: false
+`
+		})).
+		Run()
+}
+
+func TestIngressGateway(t *testing.T) {
+	istiodremote.GatewayTest(t, "installation.istiodremote")
+}
```
stevenctl, line 117: We should also copy/re-use the reachability test from `multicluster/`

stevenctl:
I assume the plan is:
1. to add a new Prow job that targets this new directory + topology file
2. move more of the stuff from TestMain to the framework/user install process
3. change the prow job to target the `pilot` tests and get better coverage?

frankbu:
Moved to #34331

PR #34272, opened 2021-07-22
Refactor ipv6 iptables to reduce code duplication

howardjohn:
This allows most calls (except ones with IPs) to use a single call for
v4 and v6. This will hopefully make things simpler and reduce drift.

Still not identical: tproxy and dns. Will do that in a followup


bianpengyuan (approved):
Nice! thanks!

PR #34274, opened 2021-07-22
Add ip6tables support for TPROXY and DNS

howardjohn:


jacob-delgado:
LGTM so far. The refactor looks great and would better help with parity between ipv4 and ipv6.

tools/istio-iptables/pkg/cmd/testdata/dns-uid-gid.golden
```diff
@@ -35,4 +35,41 @@ iptables -t raw -A OUTPUT -p udp --sport 15053 -m owner --gid-owner 1 -j CT --zo
...
+ip6tables -t nat -A ISTIO_OUTPUT -o lo ! -d ::1/128 -p tcp ! --dport 53 -m owner --uid-owner 4 -j ISTIO_IN_REDIRECT
+ip6tables -t nat -A ISTIO_OUTPUT -o lo -p tcp ! --dport 53 -m owner ! --uid-owner 4 -j RETURN
+ip6tables -t nat -A ISTIO_OUTPUT -m owner --uid-owner 4 -j RETURN
+ip6tables -t nat -A ISTIO_OUTPUT -o lo ! -d ::1/128 -m owner --gid-owner 1 -j ISTIO_IN_REDIRECT
+ip6tables -t nat -A ISTIO_OUTPUT -o lo -p tcp ! --dport 53 -m owner ! --gid-owner 1 -j RETURN
+ip6tables -t nat -A ISTIO_OUTPUT -m owner --gid-owner 1 -j RETURN
+ip6tables -t nat -A ISTIO_OUTPUT -o lo ! -d ::1/128 -m owner --gid-owner 2 -j ISTIO_IN_REDIRECT
+ip6tables -t nat -A ISTIO_OUTPUT -o lo -p tcp ! --dport 53 -m owner ! --gid-owner 2 -j RETURN
+ip6tables -t nat -A ISTIO_OUTPUT -m owner --gid-owner 2 -j RETURN
+ip6tables -t nat -A ISTIO_OUTPUT -d ::1/128 -j RETURN
```
jacob-delgado, line 60: Did a comparison between `iptables` and `ip6tables` and the only difference is iptables has `I noticed that iptables has `iptables -t nat -A ISTIO_OUTPUT -p tcp --dport 53 -d 127.0.0.53/32 -j REDIRECT --to-ports 15053` where that is missing from ipv6.

Since pilot-agent runs a local dns server on that port, would we want to extend it to 15053? I'd see that as new feature that would then be pushed back to iptables.

howardjohn: This is actually working as intended, just the test is bad. We only match traffic destined to the nameserver in resolv.conf - in this case didn't update the test to include a ipv6 server

tools/istio-iptables/pkg/builder/iptables_builder_impl.go
```diff
@@ -186,3 +186,22 @@ func (rb *IptablesBuilder) BuildV4Restore() string {
...
 }
+
+// AppendVersionedRule takes is a wrapper around AppendRule that substitutes an ipv4/ipv6 specific value
+// in place in the params. This allows appending a dual-stack rule that has an IP value in it.
+func (rb *IptablesBuilder) AppendVersionedRule(ipv4 string, ipv6 string, chain string, table string, params ...string) {
+	rb.AppendRuleV4(chain, table, replaceVersionSpecific(ipv4, params...)...)
+	rb.AppendRuleV6(chain, table, replaceVersionSpecific(ipv6, params...)...)
+}
+
+func replaceVersionSpecific(contents string, inputs ...string) []string {
```
lambdai, line 197: Looks like it's a formatter function. This works as well!

lambdai (approved):
LGTM. Thanks!

PR #34275, opened 2021-07-22
Fix race condition for auto mTLS plus workload level peer authn.

### LINKED ISSUE #34227: `concurrent map read and map write` in Istiod

incfly:
Fix #34227

The root cause is because two xds stream are both operating on the same `endpointBuilder.IstioEndpoint (a pointer)`. When one stream is sending endpoints to sidecar the other one is mutating the Endpiont.Metadata.FilterMetadata （a map).

We avoid this by making a copy of the original one. The copy is only on demand when necessary. This is similar to what we did in  https://github.com/istio/istio/blob/e490aed372744254e622c6da27edd02f03dbe968/pilot/pkg/xds/endpoint_builder.go#L203

### DIFF
--- a/common/scripts/gobuild.sh
+++ b/common/scripts/gobuild.sh
@@ -47,7 +47,6 @@ LDFLAGS=${LDFLAGS:--extldflags -static}
 GOBUILDFLAGS=${GOBUILDFLAGS:-""}
 # Split GOBUILDFLAGS by spaces into an array called GOBUILDFLAGS_ARRAY.
 IFS=' ' read -r -a GOBUILDFLAGS_ARRAY <<< "$GOBUILDFLAGS"
-echo $GOBUILDFLAGS
 
 GCFLAGS=${GCFLAGS:-}
 export CGO_ENABLED=${CGO_ENABLED:-0}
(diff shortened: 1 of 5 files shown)


common/scripts/gobuild.sh
```diff
@@ -47,6 +47,7 @@ LDFLAGS=${LDFLAGS:--extldflags -static}
 GOBUILDFLAGS=${GOBUILDFLAGS:-""}
 # Split GOBUILDFLAGS by spaces into an array called GOBUILDFLAGS_ARRAY.
 IFS=' ' read -r -a GOBUILDFLAGS_ARRAY <<< "$GOBUILDFLAGS"
+echo $GOBUILDFLAGS
```
incfly, line 50: revert this file before submit.

pilot/pkg/networking/util/util.go
```diff
@@ -525,29 +525,40 @@ func BuildLbEndpointMetadata(networkID network.ID, tlsMode, workloadname, namesp
...
+		epTLSMode = ""
+	}
+	if tlsMode == model.DisabledTLSModeLabel {
+		tlsMode = ""
+	}
 	if epTLSMode == tlsMode {
-		return false
+		return nil, false
 	}
+	log.Warnf("jianfeih debug diff clone, tlsMode %v, epTLSMode %v", tlsMode, epTLSMode)
```
incfly, line 550: revert before submit.

incfly:
tested that before and after this pr the`-race && CGO_ENABLED=1` have different output in istiod log. race report changes from exist to gone.

howardjohn (commented):
Change LGTM, just need to see how perf is. proto.Clone is not very fast. I think it should be fine since this is not commonly called code

incfly:
https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/34275/benchmark_istio/1418360904457654272/build-log.txt

benchmark looks good

incfly:
@howardjohn ping for approval?

hzxuzhonghu (commented):
lgtm. not much performance degradation

howardjohn (approved):

PR #34278, opened 2021-07-23
Fix unexpected error when using mcp-over-xds

### LINKED ISSUE #34126: Receive unexpected error when using mcp-over-xds
**Bug description**

I'm try to set up a demo with mcp-over-xds, here is the code [mcpoverxds](https://github.com/zirain/mcpoverxds)
but got some unexpected warn log, and there's no update event got

image

[ ] Docs
[ ] Installation
[x] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**
no warn log and get all istio crd change event

- install istio via istioctl 1.8.4
- install the mcpoverxds demo
- apply the helloworld demo app

image

**How was Istio installed?**

istioctl

windows 10 + docker + kubernetes on docker desktop

ChenglongGao327:
Fixes #34126 

use the master branch code to reproduce this issue, there will also be these warn logs, but can receive update events
```
2021-07-22T07:25:58.033224Z	info	adsc	Received istiod.istio-system.svc:15010 type networking.istio.io/v1alpha3/VirtualService cnt=2 nonce=94AAXEWSWkU=5e49198f-e507-4e38-8d34-d0793609aaa7
2021-07-22T07:25:58.033555Z	info	config-controller	receive istio event update, curr istio-system/tcp-stats-filter-1.9
2021-07-22T07:25:58.033559Z	info	config-controller	receive istio event update, curr istio-system/metadata-exchange-1.10
2021-07-22T07:25:58.033562Z	info	config-controller	receive istio event update, curr istio-system/metadata-exchange-1.9
2021-07-22T07:25:58.033565Z	info	config-controller	receive istio event add, curr default/testwr
2021-07-22T07:25:58.033570Z	info	config-controller	receive istio event update, curr default/synthetic-kubernetes
2021-07-22T07:25:58.033573Z	info	config-controller	receive istio event update, curr istio-system/synthetic-istiod
```

for client, if the cfg resourceVersion to be updated exists and is the same as the existing cfg, there is no need to update, otherwise update.
resource version is only used to distinguish the version, if not set, use the current time.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pilot/pkg/config/memory/store.go
+++ b/pilot/pkg/config/memory/store.go
@@ -33,6 +33,8 @@ var (
 	errConflict = errors.New("conflicting resource version, try again")
 )
 
+const ResourceVersion string = "ResourceVersion"
+
 // Make creates an in-memory config store from a config schemas
 // It is with validation
 func Make(schemas collection.Schemas) model.ConfigStore {
@@ -162,8 +164,9 @@ func (cr *store) Create(cfg config.Config) (string, error) {
 
 	if !exists {
 		tnow := time.Now()
-		cfg.ResourceVersion = tnow.String()
-
+		if cfg.ResourceVersion == "" {
+			cfg.ResourceVersion = tnow.String()
+		}
 		// Set the creation timestamp, if not provided.
 		if cfg.CreationTimestamp.IsZero() {
 			cfg.CreationTimestamp = tnow
@@ -206,44 +209,19 @@ func (cr *store) Update(cfg config.Config) (string, error) {
 	if hasConflict(existing.(config.Config), cfg) {
 		return "", errConflict
 	}
+	if cfg.Annotations != nil && cfg.Annotations[ResourceVersion] != "" {
+		cfg.ResourceVersion = cfg.Annotations[ResourceVersion]
+		delete(cfg.Annotations, ResourceVersion)
+	} else {
+		cfg.ResourceVersion = time.Now().String()
+	}
 
-	rev := time.Now().String()
-	cfg.ResourceVersion = rev
 	ns.Store(cfg.Name, cfg)
-	return rev, nil
+	return cfg.ResourceVersion, nil
 }
 
 func (cr *store) UpdateStatus(cfg config.Config) (string, error) {
-	cr.mutex.Lock()
-	defer cr.mutex.Unlock()
-	kind := cfg.GroupVersionKind
-	s, ok := cr.schemas.FindByGroupVersionKind(kind)
-	if !ok {
-		return "", errors.New("unknown type")
-	}
-	if !cr.skipValidation {
-		if _, err := s.Resource().ValidateConfig(cfg); err != nil {
-			return "", err
-		}
-	}
-
-	ns, exists := cr.data[kind][cfg.Namespace]
-	if !exists {
-		return "", errNotFound
-	}
-
-	existing, exists := ns.Load(cfg.Name)
-	if !exists {
-		return "", errNotFound
-	}
-	if hasConflict(existing.(config.Config), cfg) {
-		return "", errConflict
-	}
-
-	rev := time.Now().String()
-	cfg.ResourceVersion = rev
-	ns.Store(cfg.Name, cfg)
-	return rev, nil
+	return cr.Update(cfg)
 }
 
 func (cr *store) Patch(orig config.Config, patchFn config.PatchFunc) (string, error) {
--- a/pkg/adsc/adsc.go
+++ b/pkg/adsc/adsc.go
@@ -46,6 +46,7 @@ import (
 
 	mcp "istio.io/api/mcp/v1alpha1"
 	"istio.io/api/mesh/v1alpha1"
+	mem "istio.io/istio/pilot/pkg/config/memory"
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pilot/pkg/networking/util"
 	"istio.io/istio/pilot/pkg/serviceregistry/memory"
@@ -584,6 +585,9 @@ func mcpToPilot(m *mcp.Resource) (*config.Config, error) {
 			Labels:          m.Metadata.Labels,
 			Annotations:     m.Metadata.Annotations,
 		},
+	}
+	if c.Meta.Annotations == nil {
+		c.Meta.Annotations = make(map[string]string)
 	}
 	nsn := strings.Split(m.Metadata.Name, "/")
 	if len(nsn) != 2 {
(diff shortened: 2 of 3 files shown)


pilot/pkg/config/memory/store.go
```diff
@@ -194,56 +191,18 @@ func (cr *store) Update(cfg config.Config) (string, error) {
...
-	}
-
-	existing, exists := ns.Load(cfg.Name)
+	oldCfg, exists := ns.Load(cfg.Name)
 	if !exists {
 		return "", errNotFound
 	}
-	if hasConflict(existing.(config.Config), cfg) {
-		return "", errConflict
+	if cfg.ResourceVersion == "" || oldCfg.(config.Config).ResourceVersion != cfg.ResourceVersion {
```
howardjohn, line 198: Why remove `hasConflict` which is self documenting with a more complex boolean?

pilot/pkg/config/memory/store.go
```diff
@@ -161,12 +161,9 @@ func (cr *store) Create(cfg config.Config) (string, error) {
 	_, exists = ns.Load(cfg.Name)
 
 	if !exists {
-		tnow := time.Now()
-		cfg.ResourceVersion = tnow.String()
```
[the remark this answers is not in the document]
howardjohn: The store is meant to simulate k8s store, so it should be setting resource version I think

pilot/pkg/config/memory/store.go
```diff
@@ -272,22 +231,7 @@ func (cr *store) Patch(orig config.Config, patchFn config.PatchFunc) (string, er
...
-}
-
-// hasConflict checks if the two resources have a conflict, which will block Update calls
-func hasConflict(existing, replacement config.Config) bool {
-	if replacement.ResourceVersion == "" {
-		// We don't care about resource version, so just always overwrite
-		return false
-	}
-	// We set a resource version but its not matched, it is a conflict
-	if replacement.ResourceVersion != existing.ResourceVersion {
```
ChenglongGao327, line 289: Set the resourceVersion to the current time when creating cfg. If resourceVersion is not empty it will always return true when executed in this place.

pilot/pkg/config/memory/store.go
ChenglongGao327: https://github.com/istio/istio/blob/ce3ba502930b8cc7fa8c41ddcf1f8c3c8d9debc0/pilot/pkg/config/memory/store.go#L289
Set the resourceVersion to the current time when creating cfg. If resourceVersion is not empty it will always return true when executed in this place.

pilot/pkg/config/memory/store.go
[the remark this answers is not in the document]
ChenglongGao327: > The store is meant to simulate k8s store, so it should be setting resource version I think

I added the debug log and found that the resourceVersion exists.
When I update the resource, the resourceVersion will change, otherwise it will not change. This is not the same as k8s store.

pilot/pkg/config/memory/store.go
ChenglongGao327: > Why remove `hasConflict` which is self documenting with a more complex boolean?

Modified a bit

pilot/pkg/config/memory/store.go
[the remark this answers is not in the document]
howardjohn: > I added the debug log and found that the resourceVersion exists.

Maybe in current MCP over XDS code - the memory store is used for more than just that though?

hzxuzhonghu (approved):

pilot/pkg/config/memory/store.go
```diff
@@ -161,12 +159,13 @@ func (cr *store) Create(cfg config.Config) (string, error) {
...
-		cfg.ResourceVersion = tnow.String()
-
 		// Set the creation timestamp, if not provided.
 		if cfg.CreationTimestamp.IsZero() {
-			cfg.CreationTimestamp = tnow
+			cfg.CreationTimestamp = time.Now()
+		}
+
+		if cfg.ResourceVersion == "" {
+			cfg.ResourceVersion = time.Now().String()
```
howardjohn, line 168: nit: may be best to use the same time for creation time and resource version. otherwise could be a bit confusing

pilot/pkg/config/memory/store.go
```diff
@@ -194,56 +193,25 @@ func (cr *store) Update(cfg config.Config) (string, error) {
...
 		return "", errNotFound
 	}
 
-	existing, exists := ns.Load(cfg.Name)
-	if !exists {
-		return "", errNotFound
+	if cfg.ResourceVersion == oldCfg.(config.Config).ResourceVersion {
+		return cfg.ResourceVersion, nil
 	}
-	if hasConflict(existing.(config.Config), cfg) {
```
howardjohn, line 206: We still need the conflict check. See https://github.com/istio/istio/commit/b53449263b1b23d68a04ddef64a9b6fe64946c6d for motivation.

The store is meant to act like Kubernetes. Updating with the wrong resource version is not legal, and by removing this check we make our tests which mock Kubernetes with memory store invalid.

howardjohn (changes requested):

pilot/pkg/config/memory/store.go
hzxuzhonghu: >Updating with the wrong resource version is not legal

This violates the client store even k8s client store does not work like this. Think about this scenario, client(mcp sink) only gets notified when resources changed, for virtualService, DestinationRule, ect, the resource versions are inherited from k8s, and every update means resource version change. It is right to update the store only when resource version differs or unset.


Haven't look into https://github.com/istio/istio/commit/b53449263b1b23d68a04ddef64a9b6fe64946c6d. i guess there is a different case.

pilot/pkg/config/memory/store.go
ChenglongGao327: > nit: may be best to use the same time for creation time and resource version. otherwise could be a bit confusing

yes，thx

pilot/pkg/config/memory/store.go
ChenglongGao327: Agree with you @hzxuzhonghu 

@howardjohn  If we must use conflict check, only update cfg when the resource version is the same or not set, I can use other methods to fix this bug

howardjohn: > This violates the client store even k8s client store does not work like this

My understanding is you need to set resourceVersion=0 or resourceVersion=oldCfg.resourceVersion in k8s or update will be rejected with "conflict error"

I see.. I guess there are two use cases for the store:
* Simulate k8s store, with exact semantics of k8s. This needs conflict check, as well as auto-generate creation timestamp, etc
* Use as a cache for another store, where we want to submit items as-is and have no modifications

So there is some tension here between these two use cases

hzxuzhonghu: I think for a mcp server, the store will cache configs from k8s or other registries, we can update the cache regardless of the existing cache. In k8s，update means resource version must change. But i have to remind you that k8s is an eventual consistent model, we may not get all the update events(if the watch cache full, older events will be first dropped). 

For a MCP sink, the store will cache configs from a MCP server. Whether to update cache, there is two cases:

1. if a resource has resource version, update it when rv changes. So we can reduce many event handling if mcp server does not do delta push.
2. If a resource has no resource version, it is possible, because mcp protocol does not enforce it. In this case, we can not do delta handling, so we should update it in case miss any event.

pilot/pkg/config/memory/store.go
```diff
@@ -272,11 +251,12 @@ func (cr *store) Patch(orig config.Config, patchFn config.PatchFunc) (string, er
 		return "", errNotFound
 	}
 
-	rev := time.Now().String()
-	cfg.ResourceVersion = rev
+	if cfg.ResourceVersion == "" {
+		cfg.ResourceVersion = time.Now().String()
```
howardjohn, line 255: shouldn't patch update the resource version?

pilot/pkg/config/memory/store.go
```diff
@@ -206,44 +209,20 @@ func (cr *store) Update(cfg config.Config) (string, error) {
 	if hasConflict(existing.(config.Config), cfg) {
 		return "", errConflict
 	}
+	if cfg.Annotations != nil {
+		cfg.ResourceVersion = cfg.Annotations[ResourceVersion]
+		delete(cfg.Annotations, ResourceVersion)
+	}
+	if cfg.ResourceVersion == "" {
+		cfg.ResourceVersion = time.Now().String()
```
howardjohn, line 217: don't we need to update resource version?

howardjohn: That is how k8s works - when you apply an update the resource version changes.

howardjohn: And we can skip it if we hit L212 (annotation with resource version)

ChenglongGao327: > That is how k8s works - when you apply an update the resource version changes.

Yes, resource version should change when updating

ChenglongGao327: > And we can skip it if we hit L212 (annotation with resource version)

Yes, using if...else may be logically clearer.
L216  The reason for judging whether  resource version is empty is because I am worried that annotations are not empty, but annotations[ResourceVersion] is empty.

pilot/pkg/config/memory/store.go
ChenglongGao327: > shouldn't patch update the resource version?

Should be updated

howardjohn (approved):

Issue #34238, opened 2021-07-22
Kubernetes httpGet liveness/readiness probes fail if Istio-proxy has to follow redirect


SeanWallace:
**Bug description**

After upgrading from 1.9.5 to 1.10.3 I've noticed that I have some pods that are in crash loop backoff and the thing they all have in common is their readiness/liveness probes point to an endpoint that returns a 301 redirect.

They have the following readiness/liveness probes:

```
        livenessProbe:
          failureThreshold: 3
          httpGet:
            path: /
            port: 80
            scheme: HTTP
          initialDelaySeconds: 1
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 10
        name: apache
        ports:
        - containerPort: 80
          protocol: TCP
        readinessProbe:
          failureThreshold: 3
          httpGet:
            path: /
            port: 80
            scheme: HTTP
          initialDelaySeconds: 1
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 10
```

Pod description:

```
Readiness probe failed: Get "http://172.16.147.130:15020/app-health/apache/readyz": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

Proxy logs:

```
2021-07-22T00:53:39.186449Z	error	Request to probe app failed: Get "https://www.mydomain.com/new/location": context deadline exceeded (Client.Timeout exceeded while awaiting headers), original URL path = /app-health/apache/readyz
app URL path = /
```

So, it appears that proxy is attempting to follow this redirect however from the pod in question the networking policies prohibit that traffic, so it would never be able to.

The Kubernetes spec does say:

> Any code greater than or equal to 200 and less than 400 indicates success. Any other code indicates failure.

So a 301 would be a perfectly acceptable return code however proxy is failing with a `HTTP/1.1 500 Internal Server Error` when curl'd directly.

I'm sure this is extreme edge case and really it's super easy to work around but in the event it's unexpected or unwanted behavior I thought it worth the bug report.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

In previous versions pods that returned 301 at their probes did not fail via proxy when queried by Kubernetes.

**Steps to reproduce the bug**

[...]

Issue #29468, opened 2020-12-07
Rewrite app probes doesn't handle probes which have redirects


howardjohn:
kubelet will follow redirects on probes. Its pretty odd to have a redirect, but for HTTPs redirect it may make sense.

The proper fix is for the user to change the probe to be HTTPS probably. However, we shouldn't break if they haven't set this.

howardjohn:
not stale

PR #34291, opened 2021-07-23
Properly handle redirects in app prober

howardjohn:
This is large copy+paste from kubelet code. Before, we would just
blindly follow redirects. This has some DOS implications (not a big deal
since it just kills your own pod, unlike kubelet which is shared) and
also breaks when the redirect goes external because sourceIP=127.0.0.6
does not work for external call.

The new logic, matching kubelet:
* If we exceed 10 redirects, the probe fails
* If we redirect somewhere external, the probe succeeds immediately
* If we redirect to the same address, the probe will follow the redirect

Fixes https://github.com/istio/istio/issues/29468
Fixes https://github.com/istio/istio/issues/34238

### DIFF
--- a/pilot/cmd/pilot-agent/status/server_test.go
+++ b/pilot/cmd/pilot-agent/status/server_test.go
@@ -56,7 +56,9 @@ const (
 var liveServerStats = "cluster_manager.cds.update_success: 1\nlistener_manager.lds.update_success: 1\nserver.state: 0\nlistener_manager.workers_started: 1"
 
 func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if r.URL.Path == "/header" {
+	segments := strings.Split(r.URL.Path[1:], "/")
+	switch segments[0] {
+	case "header":
 		if r.Host != testHostValue {
 			log.Errorf("Missing expected host header, got %v", r.Host)
 			w.WriteHeader(http.StatusBadRequest)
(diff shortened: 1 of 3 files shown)


pilot/cmd/pilot-agent/status/server_test.go
```diff
@@ -556,48 +567,100 @@ func TestAppProbe(t *testing.T) {
...
+			t.Errorf("[%v] unexpected status code, want = %v, got = %v", tc.probePath, tc.statusCode, resp.StatusCode)
+		}
 	}
+	for _, tc := range testCases {
+		t.Run(tc.probePath, func(t *testing.T) { testFn(t, tc) })
+	}
+	// Next we check ever
+	t.Run("status codes", func(t *testing.T) {
+		for code := http.StatusOK; code <= http.StatusNetworkAuthenticationRequired; code++ {
+			if http.StatusText(code) == "" { // Not a valid HTTP code
```
adiprerepa, line 640: same as code=0?

adiprerepa (approved):

pilot/cmd/pilot-agent/status/server_test.go
howardjohn: No, this makes sure that a code like `399` which is not defined is not tested.

pilot/cmd/pilot-agent/status/server_test.go
```diff
@@ -65,11 +67,19 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 			w.WriteHeader(http.StatusBadRequest)
 		}
-	}
-	if r.URL.Path != "/hello/sunnyvale" && r.URL.Path != "/" {
+	case "redirect":
+		http.Redirect(w, r, "/", http.StatusMovedPermanently)
+	case "remote-redirect":
+		http.Redirect(w, r, "http://example.com/foo", http.StatusMovedPermanently)
+	case "", "hello/sunnyvale":
+		w.Write([]byte("welcome, it works"))
```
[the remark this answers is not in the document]
howardjohn: No clue. just kept the existing logic and refactored

Monkeyanator (approved):
Approving for release note.

pilot/cmd/pilot-agent/status/server_test.go
```diff
@@ -65,11 +67,19 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			log.Errorf("Missing expected Some-Header, got %v", r.Header)
 			w.WriteHeader(http.StatusBadRequest)
 		}
-	}
-	if r.URL.Path != "/hello/sunnyvale" && r.URL.Path != "/" {
+	case "redirect":
+		http.Redirect(w, r, "/", http.StatusMovedPermanently)
+	case "remote-redirect":
```
incfly, line 72: maybe add a test case for 10 exceeds? hosting a /forever-loop path and redirect it over there to achieve that effect?

incfly (approved):

PR #34295, opened 2021-07-23
Fix Kubernetes Ingress and Gateway prefix match

Monkeyanator:
Imagine a Kubernetes Ingress with `path: /sample` and `pathType: Prefix`:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: minimal-ingress
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /
spec:
  rules:
  - http:
      paths:
      - path: /sample
        pathType: Prefix
        backend:
          service:
            name: test
            port:
              number: 80
```

In our current implementation we append a `/` to  `/sample` which turns into an exact string prefix match for `/sample/`:

https://github.com/istio/istio/blob/db4afa01f9a696a2c77528d75dacccffa565ea13/pilot/pkg/config/kube/ingressv1/conversion.go#L166-L172

We do this to prevent `/sample` from matching `/samplerandomtext`. However, it also prevents us form matching on `/sample`, which must match per the Ingress spec. Need to use a smarter regex instead to make it so that we match the prefix without trailing slash AND match with additional path components after the prefix.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/config/kube/gateway/conversion.go
+++ b/pilot/pkg/config/kube/gateway/conversion.go
@@ -16,6 +16,7 @@ package gateway
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -852,6 +853,10 @@ func createHeadersMatch(match k8s.HTTPRouteMatch) (map[string]*istio.StringMatch
 	return res, nil
 }
 
+// prefixMatchRegex optionally matches "/..." at the end of a path.
+// regex taken from https://github.com/projectcontour/contour/blob/2b3376449bedfea7b8cea5fbade99fb64009c0f6/internal/envoy/v3/route.go#L59
+const prefixMatchRegex = `((\/).*)?`
+
 func createURIMatch(match k8s.HTTPRouteMatch) (*istio.StringMatch, *ConfigError) {
 	tp := k8s.PathMatchPrefix
 	if match.Path.Type != nil {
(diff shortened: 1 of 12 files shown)


howardjohn (approved):
Thanks!

can you set https://github.com/istio/istio/blob/c526ddcbb37edee38df09290796af7bcb7cb22a1/pilot/pkg/config/kube/gateway/conversion.go#L856? Double check the spec but I think it has the same logic

Monkeyanator:
All the Gateway API spec mentions for `PathMatchType.Prefix`: 

> For prefix paths, a trailing ‘/’ character in the Path is ignored, e.g. /abc and /abc/ specify the same match ([here](https://gateway-api.sigs.k8s.io/references/spec/#networking.x-k8s.io/v1alpha1.PathMatchType)).

This is less specific than what the old `Ingress` spec describes:

> Matches based on a URL path prefix split by /. Matching is case sensitive and done on a path element by element basis. A path element refers to the list of labels in the path split by the / separator. A request is a match for path p if every p is an element-wise prefix of p of the request path.

I _think_ that since the spec doesn't say either way it's safe to go with the old Ingress behavior...

howardjohn (approved):

PR #34309, opened 2021-07-24
fix possible panic of slice index out of range

gy95:
when slice index out of range `len(gs.Listeners)`, we should append `until` length is big enough to store a new value, but not just append only once,  this may cause panic due to index out of range, for example, index is 10, and `len(gs.Listeners)` is 1.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


pilot/pkg/config/kube/gateway/conditions.go
```diff
@@ -178,7 +178,7 @@ func reportGatewayCondition(obj config.Config, conditions map[string]*condition)
 func reportListenerCondition(index int, l k8s.Listener, obj config.Config, conditions map[string]*condition) {
 	obj.Status.(*kstatus.WrappedStatus).Mutate(func(s config.Status) config.Status {
 		gs := s.(*k8s.GatewayStatus)
-		if index >= len(gs.Listeners) {
+		for index >= len(gs.Listeners) {
```
hzxuzhonghu, line 181: This is safer

hzxuzhonghu (approved):

Issue #33600, opened 2021-06-23
Modifying arbitrary labels on a DestinationRule results in a full config push


Stono:
**Bug description**
https://github.com/istio/istio/issues/33569 highlighted to us how often we are doing full pushes because the impact of a push as of 1.10.1 is a large memory spike.  As our config rarely changes, this confused me.

Digging into it some more, it appears that we get a full config push even when something as arbitrary as a label on a `DestinationRule` changes (any key-value pair).

This feels wasteful, is it the desired behaviour for some reason, or a bug/potential optimisation?

[ ] Docs
[ ] Installation
[ ] Networking
[x] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**
Should only be pushing config around when it's really required.

**Steps to reproduce the bug**
Add a random label to a `DestinationRule`.

**Version** (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version --short` if you used Helm)
1.10.1

**How was Istio installed?**

**Environment where the bug was observed (cloud vendor, OS, etc)**

Additionally, please consider running `istioctl bug-report` and attach the generated cluster-state tarball to this issue.
Refer [cluster state archive](http://istio.io/help/bugs/#generating-a-cluster-state-archive) for more details.

PR #34312, opened 2021-07-24
trigger push only spec has changed for destination rules

ramaraochavali:
Fixes https://github.com/istio/istio/issues/33600. 

Currently implemented for VS and DR. Left few TODOs as followup

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


pilot/pkg/bootstrap/config_compare.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func init() {
+	// TODO(ramaraochavali): add functions for other configs.
+	registerCompareFunc(gvk.DestinationRule.String(), compareDestinationRule)
+	registerCompareFunc(gvk.VirtualService.String(), compareVirtuaService)
+}
+
+func equals(prev config.Config, curr config.Config) bool {
+	if prev.GroupVersionKind != curr.GroupVersionKind {
```
howardjohn, line 55: why is gvk mismatch equal?

pilot/pkg/bootstrap/config_compare.go
```diff
@@ -0,0 +1,62 @@
...
+		return true
+	}
+	if cf, exists := compareFuncs[prev.GroupVersionKind.String()]; exists {
+		return cf(prev, curr)
+	}
+	return true
```
howardjohn, line 61: if this is no compare function then we just assume they are equal? This doesn't seem to match stand 'equals' functionality so maybe we should have a new name or comment?

pilot/pkg/bootstrap/server.go
```diff
@@ -882,7 +882,12 @@ func (s *Server) initRegistryEventHandlers() {
 	s.ServiceController().AppendServiceHandler(serviceHandler)
 
 	if s.configController != nil {
-		configHandler := func(_ config.Config, curr config.Config, event model.Event) {
+		configHandler := func(prev config.Config, curr config.Config, event model.Event) {
+			// For update events, trigger push only if spec has changed.
+			if event == model.EventUpdate && !equals(prev, curr) {
+				log.Debugf("skipping push for %s as spec has not changed", prev.Key())
```
howardjohn, line 888: s.statusReporter.AddInProgressResource(curr)

this part might be broken since we skip it? Not 100% sure but I am thinking `istioctl wait` maybe get mad the config was never sent?

pilot/pkg/bootstrap/server.go
```diff
@@ -882,7 +882,12 @@ func (s *Server) initRegistryEventHandlers() {
 	s.ServiceController().AppendServiceHandler(serviceHandler)
 
 	if s.configController != nil {
-		configHandler := func(_ config.Config, curr config.Config, event model.Event) {
+		configHandler := func(prev config.Config, curr config.Config, event model.Event) {
+			// For update events, trigger push only if spec has changed.
+			if event == model.EventUpdate && !equals(prev, curr) {
```
hzxuzhonghu, line 887: should be equals

pilot/pkg/bootstrap/config_compare.go
hzxuzhonghu: return false if no compare function found

pilot/pkg/bootstrap/config_compare.go
```diff
@@ -0,0 +1,62 @@
...
+
+func init() {
+	// TODO(ramaraochavali): add functions for other configs.
+	registerCompareFunc(gvk.DestinationRule.String(), compareDestinationRule)
+	registerCompareFunc(gvk.VirtualService.String(), compareVirtuaService)
+}
+
+func equals(prev config.Config, curr config.Config) bool {
+	if prev.GroupVersionKind != curr.GroupVersionKind {
+		return true
```
hzxuzhonghu, line 56: false

pilot/pkg/bootstrap/config_compare.go
ramaraochavali: I will change the semantics

ramaraochavali:
@howardjohn @hzxuzhonghu changed semantics to be more clear. PTAL.

pilot/pkg/bootstrap/server.go
ramaraochavali: I was not sure about this. So moved it to defer func so that it gets always executing

pilot/pkg/bootstrap/server.go
```diff
@@ -882,7 +882,19 @@ func (s *Server) initRegistryEventHandlers() {
 	s.ServiceController().AppendServiceHandler(serviceHandler)
 
 	if s.configController != nil {
-		configHandler := func(_ config.Config, curr config.Config, event model.Event) {
+		configHandler := func(prev config.Config, curr config.Config, event model.Event) {
+			defer func() {
+				if event != model.EventDelete {
+					s.statusReporter.AddInProgressResource(curr)
+				} else {
+					s.statusReporter.DeleteInProgressResource(curr)
```
hzxuzhonghu, line 890: I think this is not right to call this when there is no push

pilot/pkg/bootstrap/server.go
hzxuzhonghu: BTW, I think we need to provide a way to completely disable `statusReporter`, currently it is always initiated, though disabled to write status.

pilot/pkg/bootstrap/server.go
ramaraochavali: https://github.com/istio/istio/pull/34312#discussion_r676164117 Changed it based on that comment.  But can change back if @howardjohn confirms that it is not an issue.

pilot/pkg/bootstrap/config_compare.go
```diff
@@ -0,0 +1,68 @@
...
+
+func registerEqualsFunc(name string, equals EqualsFunc) {
+	equalsFuncs[name] = equals
+}
+
+// destinationRuleEquals checks proxy policies
+var destinationRuleEquals = func(prev config.Config, curr config.Config) bool {
+	prevspec := prev.Spec
+	currspec := curr.Spec
+	// TODO(ramaraochavali): optimize this by comparing relevant fields only.
```
adiprerepa, line 37: why not in this PR?

adiprerepa: and maybe we should have a generic specEquals if that is all the functionality in virtualServiceEquals and destinationRuleEquals, and register the specEquals function under the DR and VS GVK

ramaraochavali: My main concern is using `reflect.DeepEquals` for performance reasons. So I think we need an optimized version of specEquals that compares each field.

That is what I meant by relevant fields. 

@howardjohn  yes. I also think all fields are relevant - but I want to optimize the field comparisions though. That is why I left that TODO.

howardjohn:
Ah I see. I wonder if some auto generated config could automatically
maintain efficient Equals functions for all API types...

That being said, I am not sure deepequals on dest rules will be a large
bottleneck - could be wrong though. I think we do it for endpoints already
which is higher scale (not 100% sure we do deepequals)

ramaraochavali:
For endpoints we use a custom equals func https://github.com/istio/istio/blob/d33ee9afc49bf533187c507e04b299b163c632cb/pilot/pkg/serviceregistry/kube/controller/endpoints.go#L240

Looking at auto generated code seems like a good option - Let me explore that

howardjohn:
proto.Equals may be fast as well

ramaraochavali:
Moved to `proto.Equal` and removed all TODOs. If it becomes a perf issue, we can take it later as followup. But with this we can support all types.

ramaraochavali:
@howardjohn Do you know why Ingress test is failing with this change? I thought k8s ingress depend on other resource change so I skipped all non istio resources from this - but still it fails. I do not see Ingress update being processed by the config Handler if I add log lines. Is this working/depend on some other istio resource?

pilot/pkg/bootstrap/config_compare.go
```diff
@@ -0,0 +1,46 @@
...
+		// This should never happen.
+		return false
+	}
+	// If the config is not Istio, let us just push.
+	if !strings.HasSuffix(prev.GroupVersionKind.Group, "istio.io") {
+		return true
+	}
+	prevspec, ok := prev.Spec.(proto.Message)
+	if !ok {
+		return false
```
howardjohn, line 39: should be true? That way if we fail we always push (essentially same as L35)?

pilot/pkg/bootstrap/server.go
```diff
@@ -883,7 +883,20 @@ func (s *Server) initRegistryEventHandlers() {
...
+				}
+			}()
+			log.Infof("event %s for config %s %s", event.String(), prev.Key(), curr.Key())
```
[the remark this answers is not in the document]
howardjohn:
```
2021-07-27T14:15:53.492519Z	info	event update for config //// networking.istio.io/v1alpha3/VirtualService//
2021-07-27T14:15:53.492685Z	info	event update for config //// networking.istio.io/v1alpha3/Gateway//
2021-07-27T14:15:57.015235Z	info	event add for config //// networking.istio.io/v1alpha3/VirtualService//
2021-07-27T14:15:57.015386Z	info	event add for config //// networking.istio.io/v1alpha3/Gateway//
2021-07-27T14:15:57.046012Z	info	event update for config //// networking.istio.io/v1alpha3/VirtualService//
2021-07-27T14:15:57.046253Z	info	event update for config //// networking.istio.io/v1alpha3/Gateway//
2021-07-27T14:15:57.091418Z	info	event update for config //// networking.istio.io/v1alpha3/VirtualService//
2021-07-27T14:15:57.091534Z	info	event update for config //// networking.istio.io/v1alpha3/Gateway//
```
how come its all empty?

pilot/pkg/bootstrap/server.go
howardjohn: Its still used by `istioctl wait` even if write status is disabled. We could still provide a way to disable if needed though

pilot/pkg/bootstrap/server.go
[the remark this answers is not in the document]
ramaraochavali: I will remove this log line. Added for debugging this problem

howardjohn:
Looked at it briefly, no clues for Ingress failing right now...

pilot/pkg/bootstrap/server.go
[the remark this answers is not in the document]
ramaraochavali: Where did you see that empty ones? I am surprised. Is that in unit tests and integration?

ramaraochavali:
@howardjohn found the gateway test issue and fixed. PTAL.

pilot/pkg/bootstrap/server.go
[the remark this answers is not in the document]
howardjohn: That was the integ test logs from the Ingress test failures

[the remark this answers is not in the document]
howardjohn: Oh from this maybe? no names: https://github.com/istio/istio/blob/617d4884e2023558ffef680bbcfb0bd20ad09667/pilot/pkg/config/kube/ingressv1/controller.go#L218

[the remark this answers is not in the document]
ramaraochavali: Yes. I fixed that

pilot/pkg/bootstrap/server.go
```diff
@@ -883,7 +883,19 @@ func (s *Server) initRegistryEventHandlers() {
...
+		configHandler := func(prev config.Config, curr config.Config, event model.Event) {
+			defer func() {
+				if event != model.EventDelete {
+					s.statusReporter.AddInProgressResource(curr)
+				} else {
+					s.statusReporter.DeleteInProgressResource(curr)
+				}
+			}()
+			// For update events, trigger push only if spec has changed.
+			if event == model.EventUpdate && !needsPush(prev, curr) {
```
howardjohn, line 895: Trying to look for places where metadata may matter:

https://github.com/istio/istio/blob/83b692a1ea44dca0f27c745f4c2970fc4d847332/pilot/pkg/config/kube/crdclient/client.go#L343 - I think not relevant since this is *before* the push trigger

https://github.com/istio/istio/blob/7d4a1366f5ce1ca8eef67ccd05975f4c02850e8b/pilot/pkg/config/kube/gateway/conversion.go#L93 - This is a real usage of labels. We need to make label changes trigger push for TLSRoute, TCPRoute, HTTPRoute. Can we also put a comment on L93 that if we add support for others (IE virtual service) we need to update it maybe?

Name/Namespace change - not possible

AuthzPolicy dry run: https://github.com/istio/istio/blob/f393a2dde0910ed71749b1604b6e093caa723547/pilot/pkg/security/authz/builder/builder.go#L158 - this needs a change

Possibly workload entry annotations matter? not certain: https://github.com/istio/istio/blob/358a5da1a43146502c5b563c7870d3d8d2f18ada/pilot/pkg/controller/workloadentry/workloadentry_controller.go#L250

Gateway annotation: https://github.com/istio/istio/blob/77366e7a2b077d426df043669f09fe23d4a22c7e/pilot/pkg/model/push_context.go#L1740

howardjohn (changes requested):
Overall I like the change and think its a good improvement. I am a bit worried about the impact it may have, both now (see other comment) or in the future (if we take a dependency on config.meta somewhere else and miss this). The bug would be very subtle and hard to catch.

Do you have any ideas on how we can improve it? One thing I was thinking of is if `*.istio.io*` label or annotation changes we push.. but that may still not capture all cases

pkg/config/constants/constants.go
```diff
@@ -113,6 +113,9 @@ const (
 
 	TestVMVersionLabel = "istio.io/test-vm-version"
 
+	// Label to skip config comparison.
+	AlwaysPushLabel = "istio.io/alwayspush"
```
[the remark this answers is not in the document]
ramaraochavali: yes. That is correct

pilot/pkg/bootstrap/server.go
ramaraochavali: I will take a look

ramaraochavali:
> One thing I was thinking of is if *.istio.io* label or annotation changes we push.. but that may still not capture all cases

Not sure if it covers all cases. Let me review the label and annotation usage you shared then we can take a call.

pilot/pkg/bootstrap/config_compare.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"istio.io/istio/pkg/config"
+	"istio.io/istio/pkg/config/constants"
+)
+
+// needsPush checks whether the passed in config has same spec and hence push needs
+// to be triggered. This is to avoid unnecessary pushes only when labels have changed
+// for example.
+func needsPush(prev config.Config, curr config.Config) bool {
+	if _, exists := prev.Meta.Labels[constants.AlwaysPushLabel]; exists {
```
hzxuzhonghu, line 28: what is `AlwaysPushLabel` for?

pkg/config/constants/constants.go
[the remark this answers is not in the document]
hzxuzhonghu: why we need this label?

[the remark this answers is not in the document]
ramaraochavali: There is no spec change here but still we need to push. This label identifies that we need to push.  We will have to make this generic based on other comments.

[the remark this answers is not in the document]
hzxuzhonghu: I donot think we should add a hack label which should be used by user, but it is hidden in code.  And not sure what is the use case here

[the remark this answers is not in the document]
ramaraochavali: This is not to be used by user BTW.

ramaraochavali:
@howardjohn I think looking at the usages - I feel this problem is only limited to the cases where we internally convert k8s service api/ingress apis to istio resources i.e. gateway, vs , dr.  In all other cases we need not worry. One thing I wonder is since these resources are internally converted by Istio code, we could add a label/annotation like `internal.istio.io/external-config` which indicates this config created as result of non istio config - and we should skip checks if we see this label. I think that should work.

Now coming to how we ensure conversion adds this label by adding a test to ensure conversion adds this label and properly commenting it - I can not think of any better way.

Let me know what you think. I can change accordingly.

howardjohn:
IoIstioDryRun and the TCP/TLS/HTTPRoute ones are not internal configurations  - these are provided by user

ramaraochavali:
> TCP/TLS/HTTPRoute ones are not internal configurations - these are provided by user

But I thought we generate VS& DR based on that right? I meant those are internally generated

howardjohn:
> But I thought we generate VS& DR based on that right? I meant those are internally generated

I think the flow (today) would be: TCPRoute adds label. Trigger push. Init push context. Read TCPRoute (and new label), and generate VS. This VS does NOT generate a push (unlike how ingress works); its just added to push context.

With this change: TCPRoute adds label. Skip push, nothing happens.

> IoIstioDryRun

For this one its just like any other change to spec

ramaraochavali:
> With this change: TCPRoute adds label. Skip push, nothing happens.

Got it. So is it fair to say if we skip non Istio resources checks this should be OK?

And for Ingress we convert it to VS, DR and Istio Gateway and there the problem exists because we are not setting the prev and curr meta correctly - Can we address this by the intergen label that I proposed above?

And if we do those two changes and skip push for any metadata that has "*istio.io" label/annotation would it cover all cases? Is it safe?

ramaraochavali:
@howardjohn What do you we do with this? I am open for any suggestions? If you think we should wait, I am happy to close this PR.

pilot/pkg/bootstrap/config_compare.go
```diff
@@ -0,0 +1,44 @@
...
+		return true
+	}
+	if prev.GroupVersionKind != curr.GroupVersionKind {
+		// This should never happen.
+		return false
```
howardjohn, line 33: nit: if it somehow does, might as well be safe and return `true`? although it should not..

pilot/pkg/bootstrap/config_compare.go
howardjohn: IMO should be true - if we do not understand it we should assume it changed and needs push

howardjohn:
I think it will be good with the suggestions you had above. A bit concerned something may break we didn't think of though

pilot/pkg/bootstrap/config_compare.go
ramaraochavali: Yes. Agree

ramaraochavali:
@howardjohn makes sense. Changed. PTAL

pilot/pkg/bootstrap/server.go
```diff
@@ -882,7 +882,19 @@ func (s *Server) initRegistryEventHandlers() {
...
+				log.Infof("skipping push for %s as spec has not changed", prev.Key())
```
howardjohn, line 895: likely too spammy?

howardjohn (approved):

pilot/pkg/bootstrap/server.go
ramaraochavali: will change to debug

Issue #33476, opened 2021-06-16
Add integration tests for external istiod


ericvn:
**Describe the feature request**

Create integration tests for an external control plane and remote clusters for promotion of external istiod to beta.

**Describe alternatives you've considered**

[ ] Docs
[x] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Additional context**

zhlsunshine:
Hi @ericvn & @GregHanson, shall we move this issue to 1.12?

PR #34331, opened 2021-07-26
External control plane integration testing

frankbu:
Fixes https://github.com/istio/istio/issues/33476

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


ericvn:
@stevenctl @howardjohn Would you please review. I have gotten these tests to run successfully locally selecting `./tests/integration/istiodremote/...`. I will verify the test-infra job PR as well.

tests/integration/iop-externalistiod-remote-integration-test-defaults.yaml
```diff
@@ -0,0 +1,27 @@
...
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+metadata:
+  name: install
+spec:
+  meshConfig:
+    accessLogFile: "/dev/stdout"
+    defaultConfig:
+      proxyMetadata:
+        ISTIO_META_DNS_CAPTURE: "true"
```
howardjohn, line 11: I do not like copying these in 4 places, they will fall out of sync (actually I think its starting out of sync?). istioctl can take multiple IOPs, we should do `-f base.yaml -f remote.yaml` so we get the defaults?

tests/integration/istiodremote/base/main_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"istio.io/istio/pkg/test/framework"
+	"istio.io/istio/pkg/test/framework/components/istio"
+	"istio.io/istio/pkg/test/framework/label"
+	"istio.io/istio/tests/integration/istiodremote"
+)
+
+var ist istio.Instance
+
+func TestMain(m *testing.M) {
```
howardjohn, line 29: why do we have istiodremote/base and istiod/multicluster? Not clear what these do differently without documentation

pkg/test/framework/components/istio/eastwest.go
```diff
@@ -81,6 +81,7 @@ func (i *operatorComponent) deployEastWestGateway(cluster cluster.Cluster, revis
 		"--set", "tag=" + imgSettings.Tag,
 		"--set", "values.global.imagePullPolicy=" + imgSettings.PullPolicy,
 		"--set", "values.gateways.istio-ingressgateway.autoscaleEnabled=false",
+		"--set", "values.gateways.istio-ingressgateway.injectionTemplate=gateway",
```
howardjohn, line 84: We should test what we ship to customers. I am 100% fine with this change, but it should be in `genGatewayScript`, not the test code

ericvn:
It didn't pick up the prow change. Trying to rerun again.

tests/integration/istiodremote/base/main_test.go
stevenctl: I think with this change we can get rid of the `tests/integration/multicluster` package. It doesn't provide further coverage than the individual multi-cluster jobs; we were mostly keeping it around for the centralremotekubeconfig/ test.

ericvn: Looking at the `multicluster/base` test, I see it has a number of tests in addition to the `reachability` test Frank added in `istiodremote/multicluster`. I'm not sure if Frank tried to pull the other tests over or not. I think we might want to keep the tests in `multicluster/base` until they are all pulled over. Thoughts on that?

pkg/test/framework/components/istio/eastwest.go
ericvn: Updated

tests/integration/istiodremote/base/main_test.go
ericvn: I did try pulling the other three tests from `multicluster/base` and  `TestCrossClusterLoadbalancing` and `TestClusterLocalServic` passed while ` TestTelemetry` didn't:
```
2021-08-03T20:18:42.651998Z	info	tf	=== DONE (passed):  Test: 'istiodremote_multicluster[TestTelemetry/telemetry] (27.7283ms)' ===
=== CONT  TestTelemetry/telemetry/config-1->http://config-1:http/
    telemetry.go:54: istio_requests_total not found in stats: map[envoy_cluster_assignment_stale:name:"envoy_cluster_assignment_stale" type:COUNTER metric:<label:<name:"cluster_name" value:"xds-grpc" > counter:<value:0 > >  envoy_cluster_assignment_timeout_received:name:"envoy_cluster_assignment_timeout_received" type:COUNTER metric:<label:<name:"cluster_name" value:"xds-grpc" > counter:<value:0 > >  envoy_cluster_bind_errors:na
```

ericvn:
Running the hidden job for these these, they are passing: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/34331/integ-istiodremote-k8s-tests_istio/1422649869926928384

tests/integration/istiodremote/base/main_test.go
```diff
@@ -0,0 +1,40 @@
...
+	framework.
+		NewSuite(m).
+		Label(label.Multicluster).
```
ericvn, line 32: Per https://github.com/istio/istio/issues/34517, I think we can remove this line.

tests/integration/istiodremote/multicluster/main_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+var (
+	ist    istio.Instance
+	appCtx multicluster.AppContext
+)
+
+func TestMain(m *testing.M) {
+	framework.
+		NewSuite(m).
+		Label(label.Multicluster).
```
ericvn, line 36: Per https://github.com/istio/istio/issues/34517, I think we can remove this line.

pkg/test/framework/components/cluster/cluster.go
```diff
@@ -102,6 +102,13 @@ func (c Clusters) Remotes(excluded ...Cluster) Clusters {
 	}, exclude(excluded...))
 }
 
+// MeshClusters returns the subset that are not external control plane clusters.
+func (c Clusters) MeshClusters(excluded ...Cluster) Clusters {
+	return c.filterClusters(func(cc Cluster) bool {
+		return !cc.IsExternalControlPlane()
```
stevenctl, line 108: Should we also have an `isMeshCluster` on the singular `Cluster` interface?

pkg/test/framework/components/cluster/cluster.go
```diff
@@ -102,6 +102,13 @@ func (c Clusters) Remotes(excluded ...Cluster) Clusters {
 	}, exclude(excluded...))
 }
 
+// MeshClusters returns the subset that are not external control plane clusters.
+func (c Clusters) MeshClusters(excluded ...Cluster) Clusters {
```
stevenctl, line 106: I did sort of like `DataPlaneClusters` as well; not hard-set on that though.

stevenctl (approved):
lgtm - and can continue to iterate

stevenctl:
let's trigger a few times to start ruling out flakiness; can probably turn it on with `optional` after this merges

pkg/test/framework/components/cluster/cluster.go
frankbu: I'm fine with either name. I just thought it's more clear to think about it as clusters that are part of the mesh vs. an external control plane cluster, which is just a kubernetes cluster hosting istiod but not itself part of the mesh.

pkg/test/framework/components/cluster/cluster.go
frankbu: I would leave it for now, since it's just the inverse of `IsExternalControlPlane`. In the future, if we come up with other kinds of external clusters, we can add it.

PR #34348, opened 2021-07-27
fix some problems of pilot/serviceregistry

gy95:
1. when `convertToConfigMap` failed, we should return, or may result in panic due to nil pointer.
2. fix some comments
 

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/endpoint_builder.go
+++ b/pilot/pkg/serviceregistry/kube/controller/endpoint_builder.go
@@ -29,7 +29,7 @@ import (
 	"istio.io/istio/pkg/network"
 )
 
-// A stateful IstioEndpoint builder with metadata used to build IstioEndpoint
+// EndpointBuilder is a stateful IstioEndpoint builder with metadata used to build IstioEndpoint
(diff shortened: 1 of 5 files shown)


pilot/pkg/serviceregistry/kube/controller/fake.go
```diff
@@ -44,6 +44,7 @@ func (fx *FakeXdsUpdater) ConfigUpdate(req *model.PushRequest) {
 	if req != nil && len(req.ConfigsUpdated) > 0 {
 		for key := range req.ConfigsUpdated {
 			id = key.Name
+			break
```
hzxuzhonghu, line 47: This is not quite correct, maybe we should send all the config updated to the Events channel. Could you debug?

gy95: ok, i will open another PR to correct this.

hzxuzhonghu (approved):

PR #34361, opened 2021-07-27
[do not merge] Debug multicluster ingress test failures

Monkeyanator:
Having trouble reproducing flakes locally which seem consistent in CI, getting some debug info


Monkeyanator:
`integ-pilot-multicluster-tests` has not failed on `TestIngress` across across 6 runs despite failing ~half the time in job history from what I can tell...

PR #34373, opened 2021-07-27
Fix TestIngress update flake

### LINKED ISSUE #34359: `TestIngress/update-class-not-istio` flake
Example failures: [https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/34346/integ-pilot-multicluster-tests_istio/1419908128928960512](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/34346/integ-pilot-multicluster-tests_istio/1419908128928960512) 
[https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/34312/integ-pilot-multicluster-tests_istio/1419976967003836416](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/34312/integ-pilot-multicluster-tests_istio/1419976967003836416)

Likely from #34295

Monkeyanator:
Fixes #34359.

Before we'd expect a `404` when hitting an ingress endpoint where the corresponding `Ingress` resource had a bogus ingress class. This is wrong since when the Ingress class doesn't match we shouldn't get a response at all (should not even have a generated Gateway). The test relies on a race condition and only passing when the new `Ingress` has not yet taken effect.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/tests/integration/pilot/ingress_test.go
+++ b/tests/integration/pilot/ingress_test.go
@@ -296,7 +296,7 @@ spec:
 			}
 
 			successValidator := echo.And(echo.ExpectOK(), echo.ExpectReachedClusters(apps.PodB.Clusters()))
-			failureValidator := echo.And(echo.ExpectCode("404"))
+			failureValidator := echo.ExpectCode("404")
 			count := 1
 			if t.Clusters().IsMulticluster() {
 				count = 2 * len(t.Clusters())
@@ -538,7 +538,7 @@ spec:
 						Headers: map[string][]string{
 							"Host": {"server"},
 						},
-						Validator: echo.ExpectCode("404"),
+						Validator: echo.ExpectError(),
 					},
 				},
 				{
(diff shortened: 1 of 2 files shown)


tests/integration/pilot/ingress_test.go
```diff
@@ -538,7 +538,7 @@ spec:
 						Headers: map[string][]string{
 							"Host": {"server"},
 						},
-						Validator: echo.ExpectCode("404"),
+						Validator: echo.ExpectError(),
```
howardjohn, line 541: seems like we may want error or 404? 

I think this only works under the assumption there are zero other Gateways in the entire cluster. Not sure how safe that assumption is...

Monkeyanator: Makes sense, added `echo.Or` and used here.

howardjohn (approved):

Issue #34356, opened 2021-07-27
CDS cache maybe not right when there is a envoyfilter that selects different proxies


hzxuzhonghu:
Currently, the cluster key does not include proxy labels/version, nor envoyfilters.

With the below envoyfilter, if there are 1.10 and 1.9 sidecars, then there is a chance some sidecars can get an error cluster.

Same happens if there is a workloadSelector specified in envoyfilter.

```
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: tcp-metadata-exchange-1.10
  namespace: istio-system
  labels:
    istio.io/rev: default
spec:
  configPatches:
    - applyTo: NETWORK_FILTER
      match:
        context: SIDECAR_INBOUND
        proxy:
          proxyVersion: '^1\.10.*'
        listener: {}
      patch:
        operation: INSERT_BEFORE
        value:
          name: istio.metadata_exchange
          typed_config:
            "@type": type.googleapis.com/udpa.type.v1.TypedStruct
            type_url: type.googleapis.com/envoy.tcp.metadataexchange.config.MetadataExchange
            value:
              protocol: istio-peer-exchange
    - applyTo: CLUSTER
      match:
        context: SIDECAR_OUTBOUND
        proxy:
          proxyVersion: '^1\.10.*'
        cluster: {}
      patch:
        operation: MERGE
        value:
          filters:
          - name: istio.metadata_exchange
            typed_config:
              "@type": type.googleapis.com/udpa.type.v1.TypedStruct
              type_url: type.googleapis.com/envoy.tcp.metadataexchange.config.MetadataExchange
              value:
                protocol: istio-peer-exchange
    - applyTo: CLUSTER
      match:
        context: GATEWAY
        proxy:
          proxyVersion: '^1\.10.*'
        cluster: {}
      patch:
        operation: MERGE
        value:
          filters:
          - name: istio.metadata_exchange
            typed_config:
              "@type": type.googleapis.com/udpa.type.v1.TypedStruct
              type_url: type.googleapis.com/envoy.tcp.metadataexchange.config.MetadataExchange
              value:
                protocol: istio-peer-exchange
```

hzxuzhonghu:
cc @howardjohn

howardjohn:
Hmm. I had envoyfilter code at some point, not sure why I removed it..

howardjohn:
We may need to turn off by default if so

PR #34376, opened 2021-07-28
temparaily disable cds cache as there is potential inconsitence

hzxuzhonghu:
Related issue: #34356

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


ramaraochavali (approved):

adiprerepa:
What inconsistency? When was this decided?

howardjohn:
Should try to get a fix in soon, its pretty easy. I actually had it at one point - somehow I removed it.. not sure why.

Also shows we have a test gap we need to address

PR #34377, opened 2021-07-28
In the mcp-over-xds scene: add resource version for service entry

ChenglongGao327:
When I use the master branch code to reproduce this issue #34126 ，I found that service entry converted from k8s service does not have resource version. 

for client, function handles the received resources
https://github.com/istio/istio/blob/8efca561efcab2d1322ebd2d50b7ecc01bfe21e5/pkg/adsc/adsc.go#L1262

If the resource version is not set, the client store always updates the resource, even if the resource has not changed.

Add resource version: if service entry is converted from k8s service, it inherits the resource version of k8s service. update the client store only when the resource changed.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pilot/pkg/model/service.go
+++ b/pilot/pkg/model/service.go
@@ -107,6 +107,9 @@ type Service struct {
 	// MeshExternal (if true) indicates that the service is external to the mesh.
 	// These services are defined using Istio's ServiceEntry spec.
 	MeshExternal bool
+
+	// ResourceVersion represents the internal version of this object.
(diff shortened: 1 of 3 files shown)


ChenglongGao327:
@howardjohn PTAL

howardjohn (approved):

PR #34378, opened 2021-07-28
Fix prefix match for root path (`/`) to match all paths

nak3:
7d4a1366f5ce1ca8eef67ccd05975f4c02850e8b fixed the prefix match behavior.

While the fix is correct for `/<something>`, the root path (`/`) does not work as expected.
The root path `/` with prefix match needs to match "all paths" as per doc[1][2].

This patch changed `/` to match all paths.

[1] https://kubernetes.io/docs/concepts/services-networking/ingress/#examples
[2] https://gateway-api.sigs.k8s.io/references/spec/

[X] Networking

Pull Request Attributes

/cc @Monkeyanator @howardjohn

### DIFF
--- a/pilot/pkg/config/kube/gateway/conversion.go
+++ b/pilot/pkg/config/kube/gateway/conversion.go
@@ -869,7 +869,7 @@ func createURIMatch(match k8s.HTTPRouteMatch) (*istio.StringMatch, *ConfigError)
 	switch tp {
 	case k8s.PathMatchPrefix, k8s.PathMatchImplementationSpecific:
 		return &istio.StringMatch{
-			MatchType: &istio.StringMatch_Regex{Regex: regexp.QuoteMeta(dest) + prefixMatchRegex},
+			MatchType: &istio.StringMatch_Regex{Regex: regexp.QuoteMeta(strings.TrimSuffix(dest, "/")) + prefixMatchRegex},
 		}, nil
 	case k8s.PathMatchExact:
 		return &istio.StringMatch{
--- a/pilot/pkg/config/kube/gateway/conversion.go
+++ b/pilot/pkg/config/kube/gateway/conversion.go
@@ -870,7 +870,7 @@ func createURIMatch(match k8s.HTTPRouteMatch) (*istio.StringMatch, *ConfigError)
 	case k8s.PathMatchPrefix, k8s.PathMatchImplementationSpecific:
 		if dest == "/" {
 			return &istio.StringMatch{
-				MatchType: &istio.StringMatch_Prefix{Prefix: dest},
+				MatchType: &istio.StringMatch_Regex{Regex: regexp.QuoteMeta(dest)},
 			}, nil
 		}
 		return &istio.StringMatch{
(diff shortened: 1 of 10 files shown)


nak3:
/test release-notes_istio
Set `release-notes-none` label as this change roll backs the behavior for root path as the current release (before 7d4a136 ).

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -868,6 +868,11 @@ func createURIMatch(match k8s.HTTPRouteMatch) (*istio.StringMatch, *ConfigError)
 	}
 	switch tp {
 	case k8s.PathMatchPrefix, k8s.PathMatchImplementationSpecific:
+		if dest == "/" {
+			return &istio.StringMatch{
+				MatchType: &istio.StringMatch_Regex{Regex: regexp.QuoteMeta(dest)},
```
Monkeyanator, line 873: should this be `&istio.StringMatch_Prefix`?

Monkeyanator (commented):
Thanks for reporting @nak3. Could you give an example of a case that doesn't work for `Prefix: /`?  Also adding an integration test for the broken case [here](https://github.com/istio/istio/blob/42b9a0342ef9ea4c38e39821332b41af1c7a96cf/tests/integration/pilot/ingress_test.go#L313) would be helpful.

From experimenting just now I actually found that paths consisting of just backslashes (e.g. `//`, `///`) don't match prefix `/` which is probably an issue since I believe that's technically valid.

nak3:
> Could you give an example of a case that doesn't work for Prefix: /?

Sure, I am using Gateway API and, for example, if I changed HTTPRoute in
https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/ to:

```
apiVersion: networking.x-k8s.io/v1alpha1
kind: HTTPRoute
metadata:
  name: http
  namespace: default
  labels:
    selected: "yes"
spec:
  gateways:
    allow: All
  hostnames: ["httpbin.example.com"]
  rules:
  - matches:
    - path:
        type: Prefix
        value: /           ## Replace with root path
    filters:
    - type: RequestHeaderModifier
      requestHeaderModifier:
        add:
          my-added-header: added-value
    forwardTo:
    - serviceName: httpbin
      port: 8000
```

Then, the curl gets `404`.

```
$ curl -s -I -H "Host: httpbin.example.com" 172.20.0.3:30419/get
HTTP/1.1 404 Not Found
date: Thu, 29 Jul 2021 00:08:53 GMT
server: istio-envoy
transfer-encoding: chunked
```

The gateway api doc says [1], so the curl for `/get` should work with prefix `/`, correct?

[1] https://gateway-api.sigs.k8s.io/references/spec/
> If no matches are specified, the default is a prefix path match on “/”, which has the effect of matching every HTTP request.

pilot/pkg/config/kube/gateway/conversion.go
nak3: Thank you. Updated.

Monkeyanator:
Ah I think the real problem here is we don't trim the `/` suffix in:

https://github.com/istio/istio/blob/master/pilot/pkg/config/kube/gateway/conversion.go#L872

Maybe we should have `regexp.QuoteMeta(strings.TrimSuffix(dest, "/"))` or similar rather than having special handling for `Prefix: /`? From local testing that seems to get things working.

nak3:
Oh! Yes, that works on my env. Thank you! Updated.

Monkeyanator (approved):

nak3:
@howardjohn Could you please take a look?

howardjohn (approved):

Issue #34393, opened 2021-07-28
benchmark tests: add a quick test to validate there are no bugs


howardjohn:
We occasionally see breakages like https://prow.istio.io/view/gs/istio-prow/logs/benchmark-report_istio_postsubmit/1420440003825111040 since we don't run benchmark in presubmit. We should have minimal case tests as unit tests (ie single iteration on small config) to ensure there is no breakages.

PR #34398, opened 2021-07-28
Fix SDS benchmark and ensure no more regressions

howardjohn:
Fixes https://github.com/istio/istio/issues/34393


adiprerepa (approved):
lgtm

PR #34402, opened 2021-07-28
delta: destinationrule deltas for CDS

adiprerepa:
Delta xDS currently only detects deltas of services, this will expand that to DestinationRules.

see #31638


pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -137,23 +137,42 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
+					break
+				}
+				dr := prevCfg.Spec.(*networking.DestinationRule)
+				removedClusterNames = append(removedClusterNames, getRemovedClusterName(watched.ResourceNames, dr.Host))
+			} else {
+				// destination exists, was updated
+				// generate cluster based on service associated with destinationrule
+				dr := cfg.Spec.(*networking.DestinationRule)
+				service := push.ServiceForHostname(proxy, host.Name(dr.Host))
+				services = append(services, service)
```
howardjohn, line 174: I don't think this accounts for duplicates - either a Service and DR for the same service change, or multiple DRs for the same service changing

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -137,23 +137,42 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
+					// something went wrong, skip -- we shouldn't be nil here
+					break
+				}
+				dr := prevCfg.Spec.(*networking.DestinationRule)
+				removedClusterNames = append(removedClusterNames, getRemovedClusterName(watched.ResourceNames, dr.Host))
+			} else {
+				// destination exists, was updated
+				// generate cluster based on service associated with destinationrule
+				dr := cfg.Spec.(*networking.DestinationRule)
+				service := push.ServiceForHostname(proxy, host.Name(dr.Host))
```
howardjohn, line 173: dr.Host can be a wildcard, so this will not work in a lot of cases

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -202,10 +221,35 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
 		model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", clusterCacheStats.hits, clusterCacheStats.hits+clusterCacheStats.miss)}, true
 }
 
+// getRemovedClusterName finds the name of the removed cluster given a hostname and list of cluster names.
+func getRemovedClusterName(resNames []string, target string) string {
+	for _, n := range resNames {
+		_, _, svcHost, _ := model.ParseSubsetKey(n)
+		if svcHost == host.Name(target) {
```
howardjohn, line 228: same wildcard comment as below applies here

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -202,10 +221,35 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
 		model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", clusterCacheStats.hits, clusterCacheStats.hits+clusterCacheStats.miss)}, true
 }
 
+// getRemovedClusterName finds the name of the removed cluster given a hostname and list of cluster names.
+func getRemovedClusterName(resNames []string, target string) string {
+	for _, n := range resNames {
+		_, _, svcHost, _ := model.ParseSubsetKey(n)
```
howardjohn, line 227: say we have dr for foo.com with subsets=[a,b]. We remove it.

This would also remove `outbound|80||foo.com` I think? when it shouldn't, should only remove  `outbound|80|a|foo.com`  and  `outbound|80|b|foo.com` 

On that note - do we handle the subsets list changing? That would end up with some deleted/added clusters

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: yeah worst case scenario we generate duplicate clusters I think, but will fix.

pilot/pkg/model/push_context.go
```diff
@@ -695,6 +695,18 @@ func (ps *PushContext) ServiceForHostname(proxy *Proxy, hostname host.Name) *Ser
 	return nil
 }
 
+// ServicesForHostname returns a list of services that fall under the hostname provided. This hostname
+// can be a wildcard.
+func (ps *PushContext) ServicesForHostname(proxy *Proxy, hostname host.Name) []*Service {
```
ramaraochavali, line 700: Why not use `ServiceForHostname(proxy *Proxy, hostname host.Name)` in push context?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -137,23 +138,59 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
+				// check with the name of our service (cluster names are in the format outbound|<port>||<hostname>
+				// so, we can check if the cluster names contains the service name, and determine if it is deleted by that
+				removedClusterNames = append(removedClusterNames, getRemovedClusterNames(watched.ResourceNames, nil, key.Name)...)
+			} else {
+				services = append(services, service)
 			}
 		} else {
-			services = append(services, service)
+			// destination rules
+			cfg := proxy.SidecarScope.DestinationRuleByName(key.Name)
```
ramaraochavali, line 160: Are we assuming updates will only have service and DR?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -137,23 +138,59 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
+				}
+				dr := prevCfg.Spec.(*networking.DestinationRule)
+				removedClusterNames = append(removedClusterNames, getRemovedClusterNames(watched.ResourceNames, nil, dr.Host)...)
+			} else {
+				// destination exists, was updated
+				// generate cluster based on service associated with destinationrule
+				// hostName can be a wildcard, and can match to multiple services
+				dr := cfg.Spec.(*networking.DestinationRule)
+				matchedSvcs := push.ServicesForHostname(proxy, host.Name(dr.Host))
+				// service := push.ServiceForHostname(proxy, host.Name(dr.Host))
```
ramaraochavali, line 176: nit: remove

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -137,23 +138,59 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
+				services = append(services, matchedSvcs...)
+				// check for removed subsets
+				prevCfg := proxy.PrevSidecarScope.DestinationRuleByName(key.Name)
+				if prevCfg == nil {
+					break
+				}
+				prevDr := prevCfg.Spec.(*networking.DestinationRule)
+				// detecting added/updated subsets works, we just need to track deletions
+				if !reflect.DeepEqual(prevDr.GetSubsets(), dr.GetSubsets()) {
```
ramaraochavali, line 185: checking len is sufficient? Do not have to do complete equals check?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -202,10 +239,53 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
+		}
+	}
+	return removed
+}
+
+// getDeletedSubsets detects deleted subsets and returns those names
+func getDeletedSubsets(orig []*networking.Subset, new []*networking.Subset) []string {
+	subsetNames := make([]string, 0)
+	// deleted if exists in orig but not in new
+	ne := make(map[string]struct{})
```
ramaraochavali, line 259: use sets.Set?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -202,10 +239,53 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
 	// as we enhance delta code to handle other complex/mixed config changes.
-	return updates != nil && allConfigKeysOfType(updates.ConfigsUpdated, gvk.ServiceEntry) && len(updates.ConfigsUpdated) > 0
+	return updates != nil && AllConfigKeysAreOfTypes(updates.ConfigsUpdated, []config.GroupVersionKind{gvk.ServiceEntry, gvk.DestinationRule}) && len(updates.ConfigsUpdated) > 0
+}
+
+// AllConfigKeysAreOfTypes returns whether or not the configs provided only consist of the gvks in target.
+func AllConfigKeysAreOfTypes(cfgs map[model.ConfigKey]struct{}, target []config.GroupVersionKind) bool {
+	deduped := make(map[config.GroupVersionKind]struct{}, 0)
+
+	for k := range cfgs {
```
ramaraochavali, line 281: nit: You can return if you find a cfg that is not part of target. It would simplify things if target is a `map[gvk]struct{}`

ramaraochavali (commented):
generally looks good. Left some comments

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: yeah, that is the clause in `shouldUseDelta`

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: if a subset is removed but then a new one is added, and then configuration is applied,  length will be the same, right? I think we need deepequals

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: maybe, but then I would have to extract everything into []string...is it ok to keep this way?

ramaraochavali: https://github.com/istio/istio/blob/master/pilot/pkg/util/sets/string.go#L19 - it is same as what you are doing here. But a cleaner api

pilot/pkg/networking/core/v1alpha3/cluster.go
ramaraochavali: Yeah. Let us leave it as DeepEquals.

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -137,23 +138,59 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
-		if service == nil {
-			// WatchedResources.ResourceNames will contain the names of the clusters it is subscribed to. We can
-			// check with the name of our service (cluster names are in the format outbound|<port>||<hostname>
-			// so, we can check if the cluster names contains the service name, and determine if it is deleted by that
-			for _, n := range watched.ResourceNames {
-				_, _, svcHost, _ := model.ParseSubsetKey(n)
-				if svcHost == host.Name(key.Name) {
-					removedClusterNames = append(removedClusterNames, n)
-				}
+		if key.Kind == gvk.ServiceEntry {
```
howardjohn, line 144: nit: this is getting big, would be good to break it out

pilot/pkg/xds/cds.go
```diff
@@ -84,5 +86,14 @@ func (c CdsGenerator) GenerateDeltas(proxy *model.Proxy, push *model.PushContext
 		return nil, nil, model.DefaultXdsLogDetails, false, nil
 	}
 	updatedClusters, removedClusters, logs, usedDelta := c.Server.ConfigGenerator.BuildDeltaClusters(proxy, push, updates, w)
+	log.Infof("CDS Delta Generation (used delta=%v) UPDATED: %v\nREMOVED: %v", usedDelta, names(updatedClusters), removedClusters)
```
howardjohn, line 89: should remove before merge

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: wait, is there actual harm in this? seems like we detect duplicates with `pilot_duplicate_envoy_clusters`

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: just into separate methods? did you have anything in mind

pilot/pkg/model/push_context.go
hzxuzhonghu, line 700: can this be called when SidecarScope has not been initiated

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: Donot you need to add namesapce?

pilot/pkg/model/sidecar.go
```diff
@@ -474,6 +474,15 @@ func (sc *SidecarScope) Services() []*Service {
 	return sc.services
 }
 
+func (sc *SidecarScope) DestinationRuleByName(name string) *config.Config {
```
hzxuzhonghu, line 477: I think we should limit SidecarScope's method calling, PushContext should be the right place. Otherwise, new contributors will be confused

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -137,23 +138,59 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
+				// WatchedResources.ResourceNames will contain the names of the clusters it is subscribed to. We can
+				// check with the name of our service (cluster names are in the format outbound|<port>||<hostname>
+				// so, we can check if the cluster names contains the service name, and determine if it is deleted by that
+				removedClusterNames = append(removedClusterNames, getRemovedClusterNames(watched.ResourceNames, nil, key.Name)...)
+			} else {
+				services = append(services, service)
 			}
 		} else {
-			services = append(services, service)
+			// destination rules
```
hzxuzhonghu, line 159: It makes the implementation complicated, someday we want to add other kinds, we need to add all of them.

hzxuzhonghu: This makes me doubt whether this is a good way to do delta xds, @howardjohn

pilot/pkg/model/push_context.go
adiprerepa: I can do a nil check

pilot/pkg/networking/core/v1alpha3/cluster.go
howardjohn: I generally agree this is pretty complicated but I am not sure of better options... any ideas?

adiprerepa: All I can think of is a large switch/case or something, breaking it out into further abstraction than a separate method seems like overkill

pilot/pkg/model/sidecar.go
adiprerepa: do you mean `*model.Proxy`

pilot/pkg/model/push_context.go
adiprerepa: @hzxuzhonghu done

pilot/pkg/model/sidecar.go
hzxuzhonghu: I mean move the method to PushContext

pilot/pkg/model/push_context.go
```diff
@@ -870,6 +885,30 @@ func (ps *PushContext) DestinationRule(proxy *Proxy, service *Service) *config.C
 	return nil
 }
 
+func (ps *PushContext) DestinationRuleByName(proxy *Proxy, name string) *config.Config {
```
hzxuzhonghu, line 888: namespace is unnecessary? name is not unique?

pilot/pkg/model/push_context.go
```diff
@@ -870,6 +885,30 @@ func (ps *PushContext) DestinationRule(proxy *Proxy, service *Service) *config.C
...
+	}
+	for _, v := range proxy.SidecarScope.destinationRules {
+		if v.Name == name {
+			return v
+		}
+	}
+	return nil
+}
+
+func (ps *PushContext) PrevDestinationRuleByName(proxy *Proxy, name string) *config.Config {
```
hzxuzhonghu, line 900: ditto

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -202,19 +242,52 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
 		model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", clusterCacheStats.hits, clusterCacheStats.hits+clusterCacheStats.miss)}, true
 }
 
+// getRemovedClusterName finds the name of the removed cluster given a hostname and list of cluster names.
+// subsets is ignored if it is nil
+func getRemovedClusterNames(resNames []string, subsets []string, target string) []string {
+	removed := make([]string, 0)
+	for _, n := range resNames {
```
hzxuzhonghu, line 249: I think this loop is not efficient

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -202,19 +242,52 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
+		_, sub, svcHost, _ := model.ParseSubsetKey(n)
+		if svcHost == host.Name(target) && (len(subsets) == 0 || contains(subsets, sub)) {
```
hzxuzhonghu, line 251: As for a dr deleted, it satisfies `len(subsets) == 0`, if a cluster svcHost = dr.Host, the cluster will be removed. 

Think about the service maybe still there, in this case envoy will miss this cluster.

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -137,23 +140,60 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
+				prevCfg := push.PrevDestinationRuleByName(proxy, key.Name)
+				if prevCfg == nil {
+					log.Infof("Prev DestinationRule form PrevSidecarScope is nil for %v", key.String())
+					break
+				}
+				prevDr := prevCfg.Spec.(*networking.DestinationRule)
+				// detecting added/updated subsets works, we just need to track deletions
+				log.Infof("Prev DR Subsets: %v\nNew DR subsets: %v", getSubsetNames(prevDr.GetSubsets()), getSubsetNames(dr.GetSubsets()))
+				if !reflect.DeepEqual(prevDr.GetSubsets(), dr.GetSubsets()) {
+					// subsets changed, did something delete?
```
hzxuzhonghu, line 189: what if dr.Host changed?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -137,23 +140,60 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, pus
...
+					deletedSubsetNames := getDeletedSubsets(prevDr.GetSubsets(), dr.GetSubsets())
+					if len(deletedSubsetNames) > 0 {
+						removedSubsets := getRemovedClusterNames(watched.ResourceNames, deletedSubsetNames, dr.Host)
```
hzxuzhonghu, line 192: The subset host can be totally different from dr.Host, it looks this pr does not take into consideration about this case.

hzxuzhonghu (changes requested):
I left some bug concerns, which can happen

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: so if a destinationRule is deleted, do we send an update instead of delete @hzxuzhonghu?

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: then it's a different destinationrule right? which is applied to a different cluster

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: I think if a subset is deleted, we need to send the subset cluster deletion

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: when does that happen?

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: I mean, watchedResources is not very large and this is O(n)

maybe a `map[host.Name]SubsetValue` where `SubsetValue` is
```go
type SubsetValue struct {
  direction TrafficDirection
  subset string
  host host.Name
  port int
}
```
and we populate/modify whenever watchedresources changes?

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: I am asking the same destinationrule.Host update

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: NVM, my memory messed.

Another question: what if the dr.Host is wildcarded

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: Yes it is O(n), but multiplex the number of proxies , it is O(n*n)

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: Istio support destinationrules merge, let's say subset v1 can be in dr1, and subset v2 can be in dr2.

adiprerepa:
@howardjohn @ramaraochavali @hzxuzhonghu ptal

samples/bookinfo/platform/kube/bookinfo-details.yaml
```diff
@@ -57,3 +58,28 @@ spec:
         securityContext:
           runAsUser: 1000
 ---
+apiVersion: networking.istio.io/v1alpha3
```
adiprerepa, line 61: will remove, this was just for testing

pilot/pkg/xds/cds.go
adiprerepa: will do

adiprerepa:
ping @howardjohn @ramaraochavali

ramaraochavali:
@adiprerepa Sorry was busy with some other tomorrow. Will finish it tomorrow.

samples/bookinfo/platform/kube/bookinfo-details.yaml
```diff
@@ -22,6 +22,7 @@ metadata:
   labels:
     app: details
     service: details
+    foo: bar
```
howardjohn, line 25: can you remove

pilot/pkg/model/push_context.go
```diff
@@ -695,6 +695,21 @@ func (ps *PushContext) ServiceForHostname(proxy *Proxy, hostname host.Name) *Ser
 	return nil
 }
 
+// ServicesForHostname returns a list of services that fall under the hostname provided. This hostname
+// can be a wildcard.
+func (ps *PushContext) ServicesForHostname(proxy *Proxy, hostname host.Name) []*Service {
```
ramaraochavali, line 700: See the changes I am making in https://github.com/istio/istio/pull/34694. Pass SidecarScope and sidecar to get this not proxy. We will remove dependency of proxy in cluster/cluster builder.

pilot/pkg/model/push_context.go
```diff
@@ -870,6 +885,30 @@ func (ps *PushContext) DestinationRule(proxy *Proxy, service *Service) *config.C
 	return nil
 }
 
+func (ps *PushContext) DestinationRuleByName(proxy *Proxy, name, namespace string) *config.Config {
+	if proxy == nil || proxy.SidecarScope == nil {
+		return nil
+	}
+	for _, v := range proxy.SidecarScope.destinationRules {
+		if v.Name == name && v.Namespace == namespace {
```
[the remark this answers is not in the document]
ramaraochavali: Same here

pilot/pkg/model/push_context.go
```diff
@@ -870,6 +885,30 @@ func (ps *PushContext) DestinationRule(proxy *Proxy, service *Service) *config.C
...
+			return v
+		}
+	}
+	return nil
+}
+
+func (ps *PushContext) PrevDestinationRuleByName(proxy *Proxy, name, namespace string) *config.Config {
```
ramaraochavali, line 900: Same.. Do not pass proxy. Pass sidecarscope directly

pilot/pkg/networking/core/v1alpha3/cluster.go
ramaraochavali: Switch case and separate methods seems better so that we can write proper unit tests for each config type

pilot/pkg/networking/core/v1alpha3/cluster.go
ramaraochavali: What are other possible different config types that come here for cluster?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -101,22 +103,65 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, upd
...
-			// so, if this service is part of watched resources, we can conclude that it is a removed cluster.
-			for _, n := range watched.ResourceNames {
-				_, _, svcHost, _ := model.ParseSubsetKey(n)
-				if svcHost == host.Name(key.Name) {
-					deletedClusters = append(deletedClusters, n)
-				}
+		// get service that changed
+		if key.Kind == gvk.ServiceEntry {
+			service := updates.Push.ServiceForHostname(proxy, host.Name(key.Name))
+			// SidecarScope.ServiceForHostname will return nil if the proxy doesn't care about the service OR it was deleted.
```
ramaraochavali, line 109: nit: push.ServiceForHostName

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -101,22 +103,65 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, upd
...
+				// so, we can check if the cluster names contains the service name, and determine if it is deleted by that
+				deletedClusters = append(deletedClusters, getRemovedClusterNames(watched.ResourceNames, nil, key.Name)...)
+			} else {
+				services = append(services, service)
 			}
 		} else {
-			services = append(services, service)
+			// destination rules
+			cfg := updates.Push.DestinationRuleByName(proxy, key.Name, key.Namespace)
+			log.Infof("adiprerepa: destinationRule name: %v namespace: %v", key.Name, key.Namespace)
```
ramaraochavali, line 124: nit: remove

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -101,22 +103,65 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, upd
...
+			if cfg == nil {
+				// subset clusters need to be removed, actual cluster needs to be updated
+				// a destinationrule was deleted, get the populated destinationRule from PrevSidecarScope
+				prevCfg := updates.Push.PrevDestinationRuleByName(proxy, key.Name, key.Namespace)
+				if prevCfg == nil {
+					// something went wrong, skip -- we shouldn't be nil here
```
ramaraochavali, line 130: can you add a warning log line here?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -101,22 +103,65 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, upd
...
+					break
```
ramaraochavali, line 131: is break correct? Should n't we continue with other configs?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -101,22 +103,65 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, upd
...
+				// generate cluster based on service associated with destinationrule
+				// hostName can be a wildcard, and can match to multiple services
+				dr := cfg.Spec.(*networking.DestinationRule)
+				matchedSvcs := updates.Push.ServicesForHostname(proxy, host.Name(dr.Host))
+				services = append(services, matchedSvcs...)
+				// check for removed subsets
+				prevCfg := updates.Push.PrevDestinationRuleByName(proxy, key.Name, key.Namespace	)
+				if prevCfg == nil {
+					log.Infof("Prev DestinationRule form PrevSidecarScope is nil for %v", key.String())
+					break
```
ramaraochavali, line 151: is break correct? Should n't we continue with other configs?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -101,22 +103,65 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, upd
...
+				}
+				prevDr := prevCfg.Spec.(*networking.DestinationRule)
+				// detecting added/updated subsets works, we just need to track deletions
+				log.Infof("Prev DR Subsets: %v\nNew DR subsets: %v", getSubsetNames(prevDr.GetSubsets()), getSubsetNames(dr.GetSubsets()))
```
ramaraochavali, line 155: will be a log spam. change it to debug

pilot/pkg/networking/core/v1alpha3/cluster.go
ramaraochavali: Can add a TODO to build a more optimized equals for subsets?

ramaraochavali (commented):
Added some comments. I think we need to add lots of tests for each function include getRemovedResources etc...

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: done

adiprerepa:
@ramaraochavali added unit tests

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: correct, fixed

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -173,6 +169,113 @@ func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *mod
 	return resources, model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", cacheStats.hits, cacheStats.hits+cacheStats.miss)}
 }
 
+func (configgen* ConfigGeneratorImpl) generateDeltaServices(updatedService model.ConfigKey, proxy *model.Proxy, push *model.PushContext, watched []string) ([]*model.Service, []string) {
+	deletedClusters := make([]string, 0)
+	services := make([]*model.Service, 0)
+	service := push.ServicesForHostname(proxy, host.Name(updatedService.Name))
+	// SidecarScope.ServiceForHostname will return nil if the proxy doesn't care about the service OR it was deleted.
```
ramaraochavali, line 176: SidecarScope.ServiceForHostname -> push.ServicesForHostname

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -173,6 +169,113 @@ func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *mod
...
+		// check with the name of our service (cluster names are in the format outbound|<port>||<hostname>
+		// so, we can check if the cluster names contains the service name, and determine if it is deleted by that
+		deletedClusters = append(deletedClusters, getRemovedClusterNames(watched, nil, updatedService.Name)...)
+	} else {
+		services = append(services, service...)
+	}
+	return services, deletedClusters
+}
+
+func (configgen *ConfigGeneratorImpl) generateDeltaDestinationRules(updatedDr model.ConfigKey, proxy *model.Proxy, push *model.PushContext, watched []string) ([]*model.Service, []string) {
```
ramaraochavali, line 191: nit: we are not generating delta destinationrules may be some thing like deltaServicesFromDestinationRules?

pilot/pkg/model/push_context.go
adiprerepa: no longer applicable

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: done

pilot/pkg/model/push_context.go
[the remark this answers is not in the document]
adiprerepa: fixed

hzxuzhonghu:
I will review again to see whether my comments are addressed.

pilot/pkg/model/push_context.go
```diff
@@ -695,6 +695,21 @@ func (ps *PushContext) ServiceForHostname(proxy *Proxy, hostname host.Name) *Ser
 	return nil
 }
 
+// ServicesForHostname returns a list of services that fall under the hostname provided. This hostname
+// can be a wildcard.
+func (ps *PushContext) ServicesForHostname(proxy *Proxy, hostname host.Name) []*Service {
+	if proxy == nil || proxy.SidecarScope == nil {
+		return nil
+	}
+	services := make([]*Service, 0)
```
ramaraochavali, line 704: should we cache this in `sidecarscope` like `matchingServicesByHostName` and use it here instead of recomputing every time?

ramaraochavali (approved):
LGTM. Apart from minor optimization. I am fine to do it as follow-up

pilot/pkg/model/push_context.go
adiprerepa: will do in follow-up

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: can address optimization in follow up

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: not really sure what you mean, could you elaborate?

hzxuzhonghu: Please take a look at PushContext.PushContext

There is a case, more than one destinationRule can be defined for a single service, though this is not common


```
// apiVersion: networking.istio.io/v1alpha3
// kind: DestinationRule
// metadata:
//   name: bookinfo-ratings
// spec:
//   host: ratings.prod.svc.cluster.local
//   trafficPolicy:
//     loadBalancer:
//       simple: LEAST_CONN
//   subsets:
//   - name: testversion
//     labels:
//       version: v3
//     trafficPolicy:
//       loadBalancer:
//         simple: ROUND_ROBIN


---

// apiVersion: networking.istio.io/v1alpha3
// kind: DestinationRule
// metadata:
//   name: bookinfo-ratings-v2
// spec:
//   host: ratings.prod.svc.cluster.local
//   trafficPolicy:
//     loadBalancer:
//       simple: LEAST_CONN
//   subsets:
//   - name: testversion
//     labels:
//       version: v2
//     trafficPolicy:
//       loadBalancer:
//         simple: ROUND_ROBIN
```

pilot/pkg/model/push_context.go
```diff
@@ -695,6 +695,21 @@ func (ps *PushContext) ServiceForHostname(proxy *Proxy, hostname host.Name) *Ser
...
+	for _, svc := range proxy.SidecarScope.services {
```
hzxuzhonghu, line 705: I would rather do this only when the given `hostname` is wildcarded

pilot/pkg/model/push_context.go
```diff
@@ -870,6 +885,28 @@ func (ps *PushContext) DestinationRule(proxy *Proxy, service *Service) *config.C
 	return nil
 }
 
+func (ps *PushContext) DestinationRuleByName(proxy *Proxy, name, namespace string) *config.Config {
+	if proxy == nil || proxy.SidecarScope == nil {
+		return nil
+	}
+	return proxy.SidecarScope.destinationRulesByNames[ConfigKey{
+		Kind:      gvk.DestinationRule,
```
hzxuzhonghu, line 893: nit: kind is redundant

pilot/pkg/model/sidecar.go
```diff
@@ -403,16 +410,19 @@ func ConvertToSidecarScope(ps *PushContext, sidecarConfig *config.Config, config
...
-			out.destinationRules[s.Hostname] = dr
-			out.AddConfigDependencies(ConfigKey{
+			ckey := ConfigKey{
 				Kind:      gvk.DestinationRule,
 				Name:      dr.Name,
 				Namespace: dr.Namespace,
-			})
+			}
+			out.destinationRules[s.Hostname] = dr
+			out.destinationRulesByNames[ckey] = dr
```
hzxuzhonghu, line 424: This is not right, In L416 `dr := ps.DestinationRule(&dummyNode, s)`, the destinationrule got is merged, so the name only include one, but i think here you want all the original names.

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -101,26 +103,20 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, upd
 	services := make([]*model.Service, 0)
 	// In delta, we only care about the services that have changed.
```
hzxuzhonghu, line 104: legacy?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -173,6 +169,113 @@ func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *mod
...
+	service := push.ServicesForHostname(proxy, host.Name(updatedService.Name))
+	// push.ServiceForHostname will return nil if the proxy doesn't care about the service OR it was deleted.
+	// we can cross-reference with WatchedResources to figure out which services were deleted.
+	if service == nil {
+		// we assume a service was deleted, it doesn't hurt to have positives for removedResources when a resource included
+		// does not exist.
+		// WatchedResources.ResourceNames will contain the names of the clusters it is subscribed to. We can
+		// check with the name of our service (cluster names are in the format outbound|<port>||<hostname>
+		// so, we can check if the cluster names contains the service name, and determine if it is deleted by that
+		deletedClusters = append(deletedClusters, getRemovedClusterNames(watched, nil, updatedService.Name)...)
```
hzxuzhonghu, line 185: don't we need to remove all the subset clusters as well if a service is deleted

hzxuzhonghu (changes requested):

hzxuzhonghu:
The most critical point is destinationrule merge, this pr does not take into consideration of that

pilot/pkg/model/push_context.go
adiprerepa: I don't understand, what's the alternative? a lookup?

pilot/pkg/model/sidecar.go
adiprerepa: could you point me to where the merging is happening?

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: what do you mean?

pilot/pkg/networking/core/v1alpha3/cluster.go
adiprerepa: gtm

pilot/pkg/model/push_context.go
hzxuzhonghu: I mean the perf can be improved

pilot/pkg/model/sidecar.go
hzxuzhonghu: I think i answered above. It is mergeDestinationRule, ptal

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: I mean both service and DestinarionRule cares

howardjohn:
What are the next steps? Waiting for @hzxuzhonghu to re-review or @adiprerepa to address comments?

pilot/pkg/model/sidecar.go
```diff
@@ -89,6 +89,7 @@ type SidecarScope struct {
 	// Union of services imported across all egress listeners for use by CDS code.
 	services           []*Service
 	servicesByHostname map[host.Name]*Service
+	servicesByNames    map[ConfigKey]*Service
```
howardjohn, line 92: not clear why we need this when we already have `servicesByHostname`? If we do have a need we should very clearly explain in the code comments why we need both

pilot/pkg/model/push_context.go
howardjohn: agree, make it take a NamespacedName type. There are some defined, not sure if in `model` or if we need to just make a new one (doesn't even need to be externally visible).

pilot/pkg/model/sidecar.go
```diff
@@ -210,8 +214,18 @@ func DefaultSidecarScopeForNamespace(ps *PushContext, configNamespace string) *S
...
+			Kind: gvk.Service,
+			Name: s.Attributes.Name,
+			Namespace: s.Attributes.Namespace,
+		}] = s
 		if dr := ps.DestinationRule(&dummyNode, s); dr != nil {
 			out.destinationRules[s.Hostname] = dr
+			out.destinationRulesByNames[ConfigKey{
+				Kind:      gvk.DestinationRule,
+				Name:      dr.Name,
+				Namespace: s.Attributes.Namespace,
```
howardjohn, line 227:
```suggestion
				Namespace: dr.Namespace,
```

pilot/pkg/model/sidecar.go
howardjohn: +1, seems like merging is not accounted for here

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -99,28 +101,22 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, upd
...
+		var deleted []string
+		if key.Kind == gvk.ServiceEntry {
+			// services
+			svcs, deleted = configgen.deltaServices(key, proxy, updates.Push, watched.ResourceNames)
 		} else {
-			services = append(services, service)
+			// destination rules
+			svcs, deleted = configgen.deltaServicesFromDestinationRules(key, proxy, updates.Push, watched.ResourceNames)
 		}
+		services = append(services, svcs...)
```
howardjohn, line 115: this will result in duplicate in both `svcs` and `deleted` I think

pilot/pkg/model/sidecar.go
```diff
@@ -184,15 +186,17 @@ func DefaultSidecarScopeForNamespace(ps *PushContext, configNamespace string) *S
...
-		RootNamespace:      ps.Mesh.RootNamespace,
-		Version:            ps.PushVersion,
+		Name:                    defaultSidecar,
+		Namespace:               configNamespace,
+		EgressListeners:         []*IstioEgressListenerWrapper{defaultEgressListener},
+		services:                defaultEgressListener.services,
+		destinationRules:        make(map[host.Name]*config.Config),
+		destinationRulesByNames: make(map[ConfigKey]*config.Config),
+		servicesByHostname:      make(map[host.Name]*Service, len(defaultEgressListener.services)),
+		servicesByNames: 		 make(map[ConfigKey]*config.Config, len(defaultEgressListener.services)),
```
howardjohn, line 196:
```suggestion
		servicesByNames: 		 make(map[ConfigKey]*Service, len(defaultEgressListener.services)),
```

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -173,6 +169,122 @@ func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *mod
...
 }
 
+func (configgen *ConfigGeneratorImpl) deltaServices(updatedService model.ConfigKey, proxy *model.Proxy, push *model.PushContext,
+	watched []string) ([]*model.Service, []string) {
+	deletedClusters := make([]string, 0)
+	services := make([]*model.Service, 0)
+	service := push.ServicesForHostname(proxy, host.Name(updatedService.Name), updatedService.Namespace)
+	// push.ServiceForHostname will return nil if the proxy doesn't care about the service OR it was deleted.
+	// we can cross-reference with WatchedResources to figure out which services were deleted.
+	if service == nil {
```
howardjohn, line 179: `ServicesForHostname` returns a list. This means this will ~never be `nil`, and we end up inserting nil entries and panic later on.

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -173,6 +169,122 @@ func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *mod
...
+		// does not exist.
+		// WatchedResources.ResourceNames will contain the names of the clusters it is subscribed to. We can
+		// check with the name of our service (cluster names are in the format outbound|<port>||<hostname>
+		// so, we can check if the cluster names contains the service name, and determine if it is deleted by that
+		deletedClusters = append(deletedClusters, getRemovedClusterNames(watched, nil, updatedService.Name)...)
+		// remove all subsets whenever a service is deleted
+		// there is usually only one service here
+		for _, s := range service {
+			cfg := push.DestinationRule(proxy, s)
+			dr := cfg.Spec.(*networking.DestinationRule)
```
howardjohn, line 190: this assumes `cfg` is non-nil

howardjohn (changes requested):

ramaraochavali:
@adiprerepa Are you still interested in reviving this? If not, I would like to take over this PR to make progress on Delta

PR #34404, opened 2021-07-29
Added envoyfilters dependency to cds cache

hzxuzhonghu:
Fix: #34356

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/model/envoyfilter.go
+++ b/pilot/pkg/model/envoyfilter.go
@@ -21,6 +21,7 @@ import (
 	"github.com/gogo/protobuf/proto"
 
 	networking "istio.io/api/networking/v1alpha3"
+	"istio.io/istio/pilot/pkg/util/sets"
 	"istio.io/istio/pkg/config"
 	"istio.io/istio/pkg/config/labels"
 	"istio.io/istio/pkg/config/xds"
@@ -163,11 +164,18 @@ func proxyMatch(proxy *Proxy, cp *EnvoyFilterConfigPatchWrapper) bool {
 	return true
 }
 
-func (efw *EnvoyFilterWrapper) Key() string {
+// Returns all the envoyfilters that wrapped.
+func (efw *EnvoyFilterWrapper) Keys() []string {
 	if efw == nil {
-		return ""
+		return nil
 	}
-	return efw.Namespace + "/" + efw.Name
+	keys := sets.Set{}
+	for _, patches := range efw.Patches {
+		for _, patch := range patches {
+			keys.Insert(patch.Key())
+		}
+	}
+	return keys.SortedList()
 }
 
 func (cpw *EnvoyFilterConfigPatchWrapper) Key() string {
--- a/pilot/pkg/model/envoyfilter.go
+++ b/pilot/pkg/model/envoyfilter.go
@@ -164,7 +164,7 @@ func proxyMatch(proxy *Proxy, cp *EnvoyFilterConfigPatchWrapper) bool {
 	return true
 }
 
-// Returns all the envoyfilters that wrapped.
+// Returns all the wrapped envoyfilters keys in increasing order .
 func (efw *EnvoyFilterWrapper) Keys() []string {
 	if efw == nil {
 		return nil
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -232,10 +232,11 @@ func (c cacheStats) merge(other cacheStats) cacheStats {
 	}
 }
 
-func buildClusterKey(service *model.Service, port *model.Port, cb *ClusterBuilder) *clusterCache {
+func buildClusterKey(service *model.Service, port *model.Port, cb *ClusterBuilder, efKeys []string) *clusterCache {
 	clusterName := model.BuildSubsetKey(model.TrafficDirectionOutbound, "", service.Hostname, port.Port)
 	clusterKey := &clusterCache{
 		clusterName:     clusterName,
+		proxyVersion:    cb.proxy.Metadata.IstioVersion,
 		locality:        cb.proxy.Locality,
 		proxyClusterID:  string(cb.proxy.Metadata.ClusterID),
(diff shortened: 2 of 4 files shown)


hzxuzhonghu:
@howardjohn Ready for review now

pilot/pkg/model/envoyfilter.go
```diff
@@ -163,11 +164,18 @@ func proxyMatch(proxy *Proxy, cp *EnvoyFilterConfigPatchWrapper) bool {
 	return true
 }
 
-func (efw *EnvoyFilterWrapper) Key() string {
+// Returns all the envoyfilters that wrapped.
+func (efw *EnvoyFilterWrapper) Keys() []string {
```
ramaraochavali, line 168: can Keys be constructed as part of EnvoyFilterWrapper creation? instead of computing them whenever it is called?

hzxuzhonghu: This can save some cpu cycles, but I think it is not very good to cache any attributes that can be inferred from the existing fields.

pilot/pkg/model/envoyfilter.go
```diff
@@ -163,11 +164,18 @@ func proxyMatch(proxy *Proxy, cp *EnvoyFilterConfigPatchWrapper) bool {
 	return true
 }
 
-func (efw *EnvoyFilterWrapper) Key() string {
+// Returns all the envoyfilters that wrapped.
```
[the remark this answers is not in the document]
hzxuzhonghu: sgtm

pilot/pkg/model/envoyfilter.go
ramaraochavali: I think it depends on how frequently you call and what is being cached. Since there wont be may patches and patches wont change after wrapper is constructed - it feels safe to cache it as there wont be any memory pressure. But up to you

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -232,10 +232,11 @@ func (c cacheStats) merge(other cacheStats) cacheStats {
 	}
 }
 
-func buildClusterKey(service *model.Service, port *model.Port, cb *ClusterBuilder) *clusterCache {
+func buildClusterKey(service *model.Service, port *model.Port, cb *ClusterBuilder, efKeys []string) *clusterCache {
 	clusterName := model.BuildSubsetKey(model.TrafficDirectionOutbound, "", service.Hostname, port.Port)
 	clusterKey := &clusterCache{
 		clusterName:     clusterName,
+		proxyVersion:    cb.proxy.Metadata.IstioVersion,
```
howardjohn, line 239: Why do we need this one?

howardjohn: Since I think its already pre-filtered in https://github.com/istio/istio/blob/2f84ae2646764f4bea6a1afcf455c95b824375b3/pilot/pkg/model/push_context.go#L1663?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -368,7 +368,20 @@ func (t *clusterCache) Key() string {
...
+		params = append(params, t.destinationRule.Name+"/"+t.destinationRule.Namespace)
+	}
+	params = append(params, t.envoyFilterKeys...)
+
+	hash := md5.New()
+	for _, param := range params {
+		hash.Write([]byte(param))
+	}
+	sum := hash.Sum(nil)
+	return hex.EncodeToString(sum)
```
howardjohn, line 384: This makes it a bit harder to debug.. maybe ok for performance though.

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: In theory each envoyfilter can include multi patches, and we allow proxy match for each proxy. 

So even though a envoyfilter can be matched by two proxies, but the patches can be different.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
hzxuzhonghu: Yes, that's also what I think. I could not balance them.

pilot/pkg/model/envoyfilter.go
hzxuzhonghu: Let me rethink about it.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
ramaraochavali: Can we add a debugKey() method that is used in debug logs?

hzxuzhonghu: We can, but currently after the key is generated, we pass it to xds cache, so we have no chance to call `debugKey`

pilot/pkg/model/envoyfilter.go
hzxuzhonghu: @ramaraochavali Cache the keys won't help here. 
As for cds, we generate a envoyfilterwrapper for each proxy once.  https://github.com/istio/istio/blob/888434e9c6f7817fa4d004ce450fcb0069365f4f/pilot/pkg/networking/core/v1alpha3/cluster.go#L79

hzxuzhonghu:
@ramaraochavali @howardjohn ptal, any other comments

pilot/pkg/networking/core/v1alpha3/cluster.go
howardjohn: I am not sure I understand. EnvoyFilterWrapper comes AFTER filtering the patches to only the ones used by the proxy. Then we look at all the patches in EnvoyFilterWrapper and form a key.

So how can they be different?

howardjohn (approved):
I still have some comments, but better than current state

PR #34409, opened 2021-07-29
fix body close leaking when ioutil.ReadAll failed or response.StatusCode is not 2xx

silenceshell:
When `client.Get(uri)` is success, but `ioutil.ReadAll` is failed or `resp.StatusCode` is not 2xx, the defer function will only calls `networkFetchFailCounter.Increment()` and return, missing call to `resp.Body.Close()`.
This will result in a leaking of resp.Body.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


hzxuzhonghu (approved):

howardjohn (commented):
lgtm

Issue #34407, opened 2021-07-29
Application Health checks fail in VMs when specifying ISTIO_SVC_IP


rinormaloku:
**Bug description**

Generating the configuration for a VM that uses a publicly accessible IP (by specifying `ISTIO_SVC_IP`) fails the health checks.

Steps:
1. create the VM configuration (PAY ATTENTION to the flag --workloadIP)
```
istioctl x workload entry configure \
    --name vm --namespace vm-services \
    --clusterID "clusterX" \
    --workloadIP $VM_IP \
    --autoregister \
    -o ./workloadfiles
```
This command creates the sidecar.env file, with the following content:
```
ISTIO_SVC_IP='40.85.149.87'
```

2. Starting the sidecar proxy in the VM will use the above IP to register the correct Workload Entry address.
3. In the virtual machine make sure that the app is running and make a request to the health endpoint to verify that it works. It does!
4. Check the WorkloadEntry `kubectl get workloadentry <name> -n vm-services -o yaml`

Expected result: Application passes health checks
Actual result: Application fails health checks with the following message:
```
  - lastProbeTime: "2021-07-29T09:33:50.281295466Z"
    lastTransitionTime: "2021-07-29T09:33:50.281296166Z"
    message: 'Get "http://40.85.149.87:8080/api/healthz": dial tcp 127.0.0.6:0->40.85.149.87:8080:
      connect: invalid argument'
    status: "False"
    type: Healthy
```

Making a request to the application fails. Meanwhile, trying the same from both local computer and the VM itself succeeds.
```
curl -v http://40.85.149.87:8080/api/healthz
```

[x] Installation

[x] Virtual Machine

PR #32383, opened 2021-04-22
adding workloadIP to istioctl config workload

rinormaloku:
This PR resolves #32462 

[x] User Experience

[x] Does not have any changes that may affect Istio users.

---
This update adds the ability to specify the workload IP when configuring a new VM-based workload.

By default, when creating the configuration for a workload:
```
istioctl --context="cluster1" x workload entry configure \
    -f workloadgroup.yaml -o ./workload-files/ \
    --clusterID "cluster1" \
    --autoregister
```

We set up all the configurations needed for registering to the remote cluster. The issue is that as an IP address is determined by:
```
ISTIO_SVC_IP=$(hostname --all-ip-addresses | cut -d ' ' -f 1)
```
This works fine when the workload is in the same network but not when it is in a separate network. 

With the update you can define the workload IP manually:
```
istioctl --context="cluster1" x workload entry configure \
    -f workloadgroup.yaml -o ./workload-files/ \
    --clusterID "cluster1" \
    --workloadIP "108.47.9.33" \
    --autoregister
```

PR #34411, opened 2021-07-29
fix health probes for VM's

rinormaloku:
This resolves the issue explained here: #34407

tl;dr 

VM's that specify the `ISTIO_SVC_IP` fail the health checks (this wasn't an issue in Istio 1.9.x when manually specifying the env variable)

With this PR we update the `istioctl workload entry configure` command to set the env var `REWRITE_PROBE_LEGACY_LOCALHOST_DESTINATION` to true, which will use localhost instead of IP address for the health probes.

### DIFF
--- a/istioctl/cmd/workload.go
+++ b/istioctl/cmd/workload.go
@@ -62,7 +62,8 @@ var (
 	outputDir      string
 	clusterID      string
 	ingressIP      string
-	workloadIP     string
+	internalIP     string
+	externalIP     string
 	ingressSvc     string
 	autoRegister   bool
 	dnsCapture     bool
@@ -221,10 +222,16 @@ Configure requires either the WorkloadGroup artifact path or its location on the
 					return fmt.Errorf("workloadgroup %s not found in namespace %s: %v", name, namespace, err)
 				}
 			}
-			if err = createConfig(kubeClient, wg, clusterID, ingressIP, workloadIP, outputDir, cmd.OutOrStderr()); err != nil {
+			if err = createConfig(kubeClient, wg, clusterID, ingressIP, internalIP, externalIP, outputDir, cmd.OutOrStderr()); err != nil {
 				return err
 			}
 			fmt.Printf("configuration generation into directory %s was successful\n", outputDir)
+			return nil
+		},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if len(internalIP) > 0 && len(externalIP) > 0 {
+				return fmt.Errorf("the flags --internalIP and --externalIP are mutually exclusive")
+			}
 			return nil
 		},
 	}
@@ -239,7 +246,8 @@ Configure requires either the WorkloadGroup artifact path or its location on the
 	configureCmd.PersistentFlags().StringVar(&ingressIP, "ingressIP", "", "IP address of the ingress gateway")
 	configureCmd.PersistentFlags().BoolVar(&autoRegister, "autoregister", false, "Creates a WorkloadEntry upon connection to istiod (if enabled in pilot).")
 	configureCmd.PersistentFlags().BoolVar(&dnsCapture, "capture-dns", true, "Enables the capture of outgoing DNS packets on port 53, redirecting to istio-agent")
-	configureCmd.PersistentFlags().StringVar(&workloadIP, "workloadIP", "", "IP address of the workload used in the WorkloadEntry")
+	configureCmd.PersistentFlags().StringVar(&internalIP, "internalIP", "", "Internal IP address of the workload")
+	configureCmd.PersistentFlags().StringVar(&externalIP, "externalIP", "", "External IP address of the workload")
 	opts.AttachControlPlaneFlags(configureCmd)
 	return configureCmd
 }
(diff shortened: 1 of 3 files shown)


rinormaloku:
@stevenctl and @howardjohn, tagging you guys as you helped review the previous story regarding supporting VMs that specifying the workloadIP: #32383

howardjohn:
I understand the problem but am not sure this is the correct fix. Will think about it more and get back to you soon

howardjohn:
Specifically because this requires applications to bind to `localhost`, which historically has proven to be something many applications do not do

howardjohn:
Thought about this some more:

One idea is to keep using external IP, but auto-detect the IP is external and not use the `::6` bind. This does not work; the request would loop back through envoy inbound and be rejected (strict mtls, etc).

Other option is to have a separate setting for internal IP and external IP. This would work but I think a bit complex.

Third option is to do exactly what you did here, but under a flag (say `--external-ip` or something) so it only impacts the (somewhat small) set of external IP users. For these users, they need to take care to ensure their applications bind to `localhost`.

I am leaning towards the third option, WDYT?

Other option is to find an alternative to "tag" traffic we don't want iptables captured besides source IP, but I have not found a good way

rinormaloku:
For the third option would these be the changes: 
1. rename workloadIP flag to --external-ip (this appends the REWRITE_PROBE_LEGACY_LOCALHOST_DESTINATION env)
2. add a new --internal-ip flag that adds only the ISTIO_SVC_IP

Or only the first point?

howardjohn:
> For the third option would these be the changes:
> 
> 1. rename workloadIP flag to --external-ip (this appends the REWRITE_PROBE_LEGACY_LOCALHOST_DESTINATION env)
> 2. add a new --internal-ip flag that adds only the ISTIO_SVC_IP
> 
> Or only the first point?

In a vacuum those both sound good. But would be slightly worried about renaming existing flag? I suppose it could be an alias. I don't have a strong opinion here, will leave to @istio/wg-user-experience-maintainers (owners of this code) to decide on best flag names

rinormaloku:
Most likely nobody uses the flag `workloadIP` as Istio 1.11 is the first release containing it — and if someone would have used it they would have filed a ticket as the feature is not working. I think it is safe to rename those. However, I'll wait for a confirmation before I make and submit the updates

howardjohn:
Oh I was not aware that it's new! No wonder no one else had this issue :-).

In that case, I think `--internal-ip` and `--external-ip` are good choices.

rinormaloku:
@howardjohn made and tested the changes. 


Should I update this release note? https://github.com/istio/istio/blob/master/releasenotes/notes/32462.yaml

Or should I create a new one?

istioctl/cmd/workload.go
```diff
@@ -360,6 +366,18 @@ func createSidecarEnv(workloadIP string, dir string) error {
 	return ioutil.WriteFile(filepath.Join(dir, "sidecar.env"), []byte(mapToString(sidecarEnv)), filePerms)
 }
 
+func generateSidecarEnvAsMap(internalIP string, externalIP string) map[string]string {
+	sidecarEnv := make(map[string]string)
+
+	if len(internalIP) > 0 {
```
howardjohn, line 372: should add some validation that both external and internal are not set. Otherwise one is silently ignored.

howardjohn (approved):
LGTM aside from one small comment. Would be great to update existing release note, as it is now no longer accurate

esnible (changes requested):
This seems reasonable to me.  I do not use the `workload` commands, so I am not sure.

You are removing the `workloadIP` parameter, replacing it with `internalIP` and `externalIP`.  Please keep `workloadIP`, but mark it as deprecated.  If it is supplied, print a warning then either fail the command or use it to set `internalIP`.

howardjohn:
> You are removing the workloadIP parameter, replacing it with internalIP and externalIP. Please keep workloadIP, but mark it as deprecated. If it is supplied, print a warning then either fail the command or use it to set internalIP.


The flag is new in 1.11, so we never released it. That is why we should make sure to merge this in 1.11 (or not at all, in the current state)

esnible (approved):

shamsher31 (approved):
Please add a release notes.

istioctl/cmd/workload.go
rinormaloku: It is okay if both are not set, then we won't create a sidecar env configuration (sidecar.env) and by default, it is set by the istio-start.sh script:
```
ISTIO_SVC_IP=$(hostname --all-ip-addresses | cut -d ' ' -f 1)
```

rinormaloku:
btw @howardjohn thinking about this, does it make more sense to have these flags:

--workloadIP and
--probe-on-localhost
?

I.e. workloadIP configures the IP that the workload uses to auto register and for health checks,

Meanwhile --probe-on-localhost overwrites so that the health check is performed on localhost for apps bound to it. (I think the --internal-ip and --external-ip might confuse users)

howardjohn:
> Meanwhile --probe-on-localhost overwrites so that the health check is performed on localhost for apps bound to it. (I think the --internal-ip and --external-ip might confuse users)

idk, personally I think `--probe-on-localhost` is more confusing to users (then again, I am not a user :slightly_smiling_face: ). It specifies what we want to do, but a user needs to know *why* they want to do that - I doubt anyone will know that unless they get a broken VM and stumble upon this PR/issue?

Whereas with external/internal, they will see both flags and realize they need to decide which their IP is. Under the hood we make it work - which could be through different means in the future.

rinormaloku:
Makes sense...

So, I am asking as I don't know the implementation details and what I am
not sure is:

Do we want to always set  REWRITE_PROBE_LEGACY_LOCALHOST_DESTINATION to
true when we have an external IP?

If yes then the current implementation is good :))

howardjohn:
Today we do. Maybe in the future we have a better solution

rinormaloku:
@howardjohn do we want to `/cherrypick release-1.11` this change?

howardjohn:
hm, it was supposed to.. from the label

/cherrypick release-1.11

PR #34423, opened 2021-07-29
grpcgen: inbound configuration

stevenctl:
Adds mTLS support for gRPC _servers_

Usage:
* Server must use `xds.NewGRPCServer`
* For security, the env flag must be enabled _and_ an XDS specific `TransportCredentials` must be provided when creating the server

--- 

Generator changes:

* share CommonTlsContext gen from cds
* make listener resource names also filter at the port-level
* generate inbound listeners only if there are corresponding ServiceInstances  on the proxy
* only support DISABLE or STRICT; can't really do much "detection" due to filter chain match limitations

See https://github.com/grpc/proposal/blob/master/A36-xds-for-servers.md for details

---

Test changes:

* Added a flag to echo to mark a gRPC port as an XDS server

* Had to regenerate echo Go code from newer protoc; old version only accepts `*grpc.Server` and we need `grpc.ServiceRegistrar` to accept the new XDS version of the server

* Echo gRPC readiness has to authenticate with itself. We can't just utilize xDS to figure out the auth because we'd need to use an `xds:///` url with FQDN and possibly hit other instances. Instead we workaround by reading cert paths from the bootstrap. 

* `grpcgen` tests now use the `grpcxds` package to generate bootstrap instead of using a static one

* Added a local-only test for mTLS with gRPC server and client

* // TODO add an integration test to security suite with mTLS enabled on proxyless server and client (manually tested; no DR or mTLS disabled client side results in "did not look like TLS" on the server; no PA or disabled serverside results in client-side failures. 

---

Some possibly noisy benchmark results:

image

image


pilot/pkg/xds/endpoint_builder.go
```diff
@@ -292,12 +292,13 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
 	}
 	// The shards are updated independently, now need to filter and merge for this cluster
 	for _, clusterID := range keys {
-		endpoints := shards.Shards[clusterID]
 		// If the downstream service is configured as cluster-local, only include endpoints that
-		// reside in the same cluster.
-		if isClusterLocal && (cluster.ID(clusterID) != b.clusterID) {
+		// reside in the same cluster. A clusterID of "" may indicate this is a ServiceEntry registry and should be
```
costinm, line 296: When would this happen ? Can you add more comments and example. Unlike the grpcgen changes, this is critical path code.

pkg/istio-agent/grpcxds/grpc_bootstrap.go
```diff
@@ -31,13 +31,23 @@ import (
 	"istio.io/istio/pkg/file"
 )
 
+const (
+	ServerListenerNamePrefix = "lds/inbound|"
```
costinm, line 35: Is this a new pattern ? This is turning a bit into an API - the 'lds/' and inbound| are a bit strange. Could we have a more intuitive/natural naming ?

costinm (approved):
Seems reasonably safe - I would still like to understand when clusterid can be empty, and we should have a separate discussion on which listener/cluster naming scheme we want to support considering some is exposed to user now and it may impact federation. But nothing wrong with this PR that I can see.

pilot/pkg/xds/endpoint_builder.go
stevenctl: See the discussion in https://github.com/istio/istio/pull/34480

The tests hit istiod and we only send cluster local endpoints for that service. The test worked before because we didn't set a proper cluster Id in metadata but now we do.

pkg/istio-agent/grpcxds/grpc_bootstrap.go
stevenctl: Something like `xds.istio.io/listener/inbound/%s`?

hzxuzhonghu:
This should be after #34480

pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -52,6 +52,9 @@ import (
 )
 
 const (
+	// for proxyless we add a special gRPC server that doesn't get configured with xDS for test-runner use
+	grpcMagicPort = 117171
```
howardjohn, line 56: Isn't this greater than max port number?

pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -172,6 +175,9 @@ spec:
           - "{{ $cluster }}"
 {{- range $i, $p := $.ContainerPorts }}
 {{- if eq .Protocol "GRPC" }}
+{{- if and $.ProxylessGRPC (ne $p.Port $.GRPCMagicPort) }}
```
howardjohn, line 178: why not just add an option like LocalhostIP, InstanceIP, etc?

stevenctl: Done (in getContainerPorts rather than template)

stevenctl:
> This should be after #34480

I worked around that bug in this PR so it's not blocked by testing against a non-cluster-local service.

pilot/pkg/networking/grpcgen/grpcecho_test.go
```diff
@@ -66,34 +68,54 @@ type configGenTest struct {
 //      ports:
 //        grpc: {generated portnum}
 func newConfigGenTest(t *testing.T, discoveryOpts xds.FakeOptions, servers ...echoCfg) *configGenTest {
-	if runtime.GOOS == "darwin" {
-		// TODO always skip if this breaks anywhere else
-		t.Skip("cannot use 127.0.0.x on OSX")
-	}
+	//if runtime.GOOS == "darwin" {
```
howardjohn, line 71: Is this no longer required?

pilot/pkg/networking/grpcgen/grpcecho_test.go
```diff
@@ -206,6 +228,65 @@ spec:
...
+  mtls:
+    mode: STRICT
+`,
+	}, echoCfg{version: "v1"})
+
+	retry.UntilSuccessOrFail(tt.T, func() error {
+		cw := tt.dialEcho("xds:///echo-app.default.svc.cluster.local:7070")
+		for i := 0; i < 10; i++ {
+			res, err := cw.Echo(context.Background(), &proto.EchoRequest{Message: "needle"})
+			fmt.Println(res)
```
howardjohn, line 281: possibly do not want println

pilot/pkg/networking/grpcgen/grpcgen_test.go
```diff
@@ -26,30 +28,57 @@ import (
...
 	"istio.io/istio/pkg/test"
+	"istio.io/istio/pkg/test/env"
 )
 
 var (
 	grpcXdsAddr = "127.0.0.1:14057"
 
 	// Address of the Istiod gRPC service, used in tests.
-	istiodSvcAddr = "istiod.istio-system.svc.cluster.local:14057"
+	// TODO `notistiod` is a hack to workaround a bug in cluster-local code
```
howardjohn, line 48: any more info on the bug would be good?

stevenctl: added link to issue

howardjohn (approved):

PR #34429, opened 2021-07-30
Reset refreshTicker instead of stop and create a new Ticker

silenceshell:
As the `refreshTicker` has expired here, we could reset the ticker instead of stop and create a new Ticker. 

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/model/jwks_resolver.go
+++ b/pilot/pkg/model/jwks_resolver.go
@@ -394,8 +394,7 @@ func (r *JwksResolver) refresher() {
 				r.refreshInterval = r.refreshDefaultInterval
 			}
 			lastHasError = currentHasError
-			r.refreshTicker.Stop()
-			r.refreshTicker = time.NewTicker(r.refreshInterval)
+			r.refreshTicker.Reset(r.refreshInterval)
 		case <-closeChan:
 			r.refreshTicker.Stop()
 			return


pilot/pkg/model/jwks_resolver.go
```diff
@@ -394,8 +394,7 @@ func (r *JwksResolver) refresher() {
 				r.refreshInterval = r.refreshDefaultInterval
 			}
 			lastHasError = currentHasError
-			r.refreshTicker.Stop()
-			r.refreshTicker = time.NewTicker(r.refreshInterval)
+			r.refreshTicker.Reset(r.refreshInterval)
```
hzxuzhonghu, line 397: It is not safe to use reset like this. 
https://github.com/golang/go/blob/7d2473dc81c659fba3f3b83bc6e93ca5fe37a898/src/time/sleep.go#L96-L119

hzxuzhonghu: NVM, it is ticker not timer.  Looks good

silenceshell: Thanks for your response.

```
// If a program has already received a value from t.C, the timer is known
// to have expired and the channel drained, so t.Reset can be used directly.
```

According to the comment above, since we have received value from `<-r.refreshTicker.C` (L379), I think it is safe to reset? I am not sure if there are other conditions, please let me know that if so. Thanks again.

hzxuzhonghu (approved):

PR #34437, opened 2021-07-30
remove duplicate mutex lock from atomic bool

gy95:
field `refreshIndexes` is already `atomic bool` value, so we do not need to use mutex to protect its read/write.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
@@ -347,9 +347,7 @@ func (s *ServiceEntryStore) serviceEntryHandler(old, curr config.Config, event m
 
 	// Recomputing the index here is too expensive - lazy build when it is needed.
 	// Only recompute indexes if services have changed.
-	s.storeMutex.Lock()
 	s.refreshIndexes.Store(true)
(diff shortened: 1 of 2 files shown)


ramaraochavali (approved):

howardjohn:
This is not always safe. Consider old code:

```go
mu.Lock
a := refreshIndexes.Load()
b := refreshIndexes.Load()
mu.Unlock()
```

Before, we are guaranteed `a == b`. Now they may change.

I am a bit concerned by this change without thorough review because its not fixing any known issues, and there have been a number of very subtle race conditions here.

ramaraochavali:
Sorry I missed this part. Will revert this

howardjohn:
this change may actually be fine, just noting that it's not always wrong to
lick an atomic var

ramaraochavali:
Ah,. Ok Got it - I thought we are using it like the way you mentioned. I agree this part of the code has been real trouble some with locking :-)

hzxuzhonghu (approved):

PR #34446, opened 2021-07-30
[release-1.11] Remove old validating webhook configuration from base

### LINKED ISSUE #34452: Validating webhook installation regression
We introduced per-revision validating webhook configurations in [https://github.com/istio/istio/pull/33044/files](https://github.com/istio/istio/pull/33044). As part of this, we had to keep the old validating webhook around to keep from purging it on revisioned upgrade with Helm. However, having a lame-duck validating webhook configuration around, even if it never moves to `FailClose`, has a cost:

1. Produces error logs about missing CA bundle on each resource application
2. Webhook still gets called on each resource application

In addition, with Helm-based install, we never install the revisioned webhook that is meant to actually handle validation since `global: configValidation` is not set in the `istio-discovery` chart.

Monkeyanator:
Fixes #34452 

This resource was kept around to keep from purging the old validating webhook configuration on revisioned upgrade with Helm. However, having a lame-duck validating webhook configuration around, even if it never moves to `FailClose`, has a cost:

1. Produces error logs about missing CA bundle on each resource application
2. Webhook still gets called on each resource application

Given these drawbacks it makes sense to remove the old validator to prevent a regression. Should have a corresponding warning in Helm revisioned upgrade doc that  for 1.10->1.11 the old validator will be purged on new base installation.

[ ] Docs
[x] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


jwendell (commented):
Is there a corresponding PR in master for this? Is it too risky to have this on 1.11?

Monkeyanator:
> Is there a corresponding PR in master for this?

For master this will be solve in [https://github.com/istio/istio/pull/33773](https://github.com/istio/istio/pull/33773) but backporting that entire PR is too risky at this stage.

> Is it too risky to have this on 1.11?

The main risk is on Helm upgrade- validation will stop working for old revisions and switch to the new revision on upgrade. However, since we know master is having problems on fresh install with this webhook, the change makes sense. Will leave for others to comment.

howardjohn:
> The main risk is on Helm upgrade- validation will stop working for old revisions and switch to the new revision on upgrade. However, since we know master is having problems on fresh install with this webhook, the change makes sense. Will leave for others to comment.

What about in place helm upgrade?

Monkeyanator:
> What about in place helm upgrade?

Users will lose validation in the window between upgrade their base chart and upgrading their pilot chart.

nrjpoddar (commented):
This change LGTM.

jwendell (approved):

PR #34464, opened 2021-08-01
refactor build clusters logic

ramaraochavali:
This PR does the following
- Removes `GenerateDeltas` - I did not see a need for it. It is actually lot of boiler plate and many generates do not care about it. Changed the existing `Generate` signature to return `DeletedResources` also which will be used in case of incremental Xds like Delta
- Removed `usedDelta` flag that indicates whether deltas have been actually used. All we care about is whether there are removedResources. We can just rely on length of removedResources
- Refactor buildCluster logic so that is used by both `BuildClusters` and `BuildDeltaClusters`
- Some other minor refactors to simplify the code.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -212,7 +212,8 @@ var DefaultXdsLogDetails = XdsLogDetails{}
 type XdsResourceGenerator interface {
 	// Generate builds the xds resources for given proxy. Resources contain the newly added/updated resources.
 	// DeletedResources typically used in incremental Xds protocols contain the resources that have been deleted.
-	Generate(proxy *Proxy, push *PushContext, w *WatchedResource, updates *PushRequest) (Resources, DeletedResources, XdsLogDetails, error)
+	// It also indicates whether the response is delta or SoTW.
+	Generate(proxy *Proxy, push *PushContext, w *WatchedResource, updates *PushRequest) (Resources, DeletedResources, bool, XdsLogDetails, error)
 }
 
 // Proxy contains information about an specific instance of a proxy (envoy sidecar, gateway,
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -172,6 +172,8 @@ func (e *Environment) SetLedger(l ledger.Ledger) {
 // Resources is an alias for array of marshaled resources.
 type Resources = []*discovery.Resource
 
+type DeletedResources = []string
+
 func AnyToUnnamedResources(r []*any.Any) Resources {
 	a := make(Resources, 0, len(r))
 	for _, rr := range r {
@@ -208,10 +210,9 @@ var DefaultXdsLogDetails = XdsLogDetails{}
 // Note: any errors returned will completely close the XDS stream. Use with caution; typically and empty
 // or no response is preferred.
 type XdsResourceGenerator interface {
-	Generate(proxy *Proxy, push *PushContext, w *WatchedResource, updates *PushRequest) (Resources, XdsLogDetails, error)
-
-	// GenerateDeltas returns the changed and removed resources, along with whether or not delta was actually used.
-	GenerateDeltas(proxy *Proxy, push *PushContext, updates *PushRequest, w *WatchedResource) (Resources, []string, XdsLogDetails, bool, error)
+	// Generate builds the xds resources for given proxy. Resources contain the newly added/updated resources.
+	// DeletedResources typically used in incremental Xds protocols contain the resources that have been deleted.
+	Generate(proxy *Proxy, push *PushContext, w *WatchedResource, updates *PushRequest) (Resources, DeletedResources, XdsLogDetails, error)
 }
 
 // Proxy contains information about an specific instance of a proxy (envoy sidecar, gateway,
--- a/pilot/pkg/networking/core/configgen.go
+++ b/pilot/pkg/networking/core/configgen.go
@@ -39,8 +39,7 @@ type ConfigGenerator interface {
 
 	// BuildDeltaClusters returns both a list of resources that need to be pushed for a given proxy and a list of resources
 	// that have been deleted and should be removed from a given proxy. This is Delta CDS output.
-	BuildDeltaClusters(node *model.Proxy, push *model.PushContext, updates *model.PushRequest,
-		watched *model.WatchedResource) ([]*discovery.Resource, []string, model.XdsLogDetails, bool)
+	BuildDeltaClusters(node *model.Proxy, updates *model.PushRequest, watched *model.WatchedResource) ([]*discovery.Resource, []string, model.XdsLogDetails)
 
 	// BuildHTTPRoutes returns the list of HTTP routes for the given proxy. This is the RDS output
 	BuildHTTPRoutes(node *model.Proxy, push *model.PushContext, routeNames []string) []*route.RouteConfiguration
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -78,6 +78,7 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
 // Cluster type based on resolution
 // For inbound (sidecar only): Cluster for each inbound endpoint port and for each service port
 func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, req *model.PushRequest) ([]*discovery.Resource, model.XdsLogDetails) {
+	// In Sotw, we care about all services.
 	var services []*model.Service
 	if features.FilterGatewayClusterConfig && proxy.Type == model.Router {
 		services = req.Push.GatewayServices(proxy)
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -123,7 +123,8 @@ func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, upd
 }
 
 // buildClusters builds clusters for the proxy with the services passed.
-func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *model.PushRequest, services []*model.Service) ([]*discovery.Resource, model.XdsLogDetails) {
+func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *model.PushRequest,
+	services []*model.Service) ([]*discovery.Resource, model.XdsLogDetails) {
 	clusters := make([]*cluster.Cluster, 0)
 	resources := model.Resources{}
 	envoyFilterPatches := req.Push.EnvoyFilters(proxy)
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -36,7 +36,7 @@ import (
 	"istio.io/istio/pilot/pkg/networking/core/v1alpha3/loadbalancer"
 	"istio.io/istio/pilot/pkg/networking/util"
 	"istio.io/istio/pilot/pkg/serviceregistry/provider"
-	"istio.io/istio/pkg/config"
+	"istio.io/istio/pilot/pkg/util/sets"
 	"istio.io/istio/pkg/config/host"
 	"istio.io/istio/pkg/config/protocol"
 	"istio.io/istio/pkg/config/schema/gvk"
@@ -52,6 +52,10 @@ var defaultTransportSocketMatch = &cluster.Cluster_TransportSocketMatch{
 		Name: util.EnvoyRawBufferSocketName,
 	},
 }
+
+// deltaConfigTypes are used to detect changes and trigger delta calculations. When config updates has ONLY entries
+// in this map, then delta calculation is triggered.
+var deltaConfigTypes = sets.NewSet(gvk.ServiceEntry.Kind)
 
 // getDefaultCircuitBreakerThresholds returns a copy of the default circuit breaker thresholds for the given traffic direction.
 func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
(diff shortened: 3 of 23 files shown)


pilot/pkg/xds/delta.go
```diff
@@ -446,7 +446,7 @@ func (s *DiscoveryServer) pushDeltaXds(con *Connection, push *model.PushContext,
 		Nonce:             nonce(push.LedgerVersion),
 		Resources:         res,
 	}
-	if usedDelta {
+	if len(deletedRes) > 0 {
```
adiprerepa, line 449: We can't rely on the length of `DeletedResources` to determine whether or not delta was used. The case could just be that delta generation was used, but nothing was deleted.

pilot/pkg/xds/cds.go
```diff
@@ -69,20 +70,14 @@ func cdsNeedsPush(req *model.PushRequest, proxy *model.Proxy) bool {
...
-}
-
-// GenerateDeltas for CDS currently only builds deltas when services change. todo implement changes for DestinationRule, etc
-func (c CdsGenerator) GenerateDeltas(proxy *model.Proxy, push *model.PushContext, updates *model.PushRequest,
-	w *model.WatchedResource) (model.Resources, []string, model.XdsLogDetails, bool, error) {
-	if !cdsNeedsPush(updates, proxy) {
-		return nil, nil, model.DefaultXdsLogDetails, false, nil
+	if c.delta {
+		updatedClusters, removedClusters, logs := c.Server.ConfigGenerator.BuildDeltaClusters(proxy, updates, w)
+		return updatedClusters, removedClusters, logs, nil
```
adiprerepa, line 79: If we are going this direction, I would change the signature of `Generate` to return a `bool` for whether delta generation was actually used, and return true here.

adiprerepa (commented):
Should we be changing the existing `Generate` API at all?

Also, could you test delta out locally to make sure it still works?

adiprerepa:
cc @howardjohn

pilot/pkg/xds/delta.go
ramaraochavali: But what we care about is sending RemovedResources in Delta Response right? Or do we care about some thing else also?

pilot/pkg/xds/cds.go
ramaraochavali: I did not understand this part. What we care is there are removed resources or not right? Then why do we need this explicit flag ? May be I am missing some thing here

ramaraochavali:
> Should we be changing the existing Generate API at all?

Since this is internal to Istio I think it should be fine unless some external dependency is there. Is there any contract from Istio side for this? @howardjohn may know better.

howardjohn:
> > Should we be changing the existing Generate API at all?
> 
> Since this is internal to Istio I think it should be fine unless some external dependency is there. Is there any contract from Istio side for this? @howardjohn may know better.

fine to change

pilot/pkg/xds/delta.go
howardjohn: you need a way to distinguish "these are the SotW" and "I generated some new partial updates, with no deletes"
The problem is many generators (including cds depending on the update) return SotW style so we need to convert that to delta (ie determine what were deleted, as done here)

hzxuzhonghu (commented):
Need to be more careful, I prefer to leave two separate paths and resue some functions.

pilot/pkg/xds/delta.go
ramaraochavali: Ok. Is this needed for intermediate state? And if for example cds completely moves to delta, we can get rid of this flag?

ramaraochavali:
> Need to be more careful, I prefer to leave two separate paths and resue some functions.

Two seperate paths at actual cluster, listener makes sense but at Generator level it is lot of boiler plate code and most of the generators not implement delta at all. And We can simplify `Generate` interface once we fully implement delta.

ramaraochavali:
@howardjohn @adiprerepa brought usedDelta flag back. PTAL

pilot/pkg/xds/cds.go
```diff
@@ -69,20 +70,14 @@ func cdsNeedsPush(req *model.PushRequest, proxy *model.Proxy) bool {
...
-}
-
-// GenerateDeltas for CDS currently only builds deltas when services change. todo implement changes for DestinationRule, etc
-func (c CdsGenerator) GenerateDeltas(proxy *model.Proxy, push *model.PushContext, updates *model.PushRequest,
-	w *model.WatchedResource) (model.Resources, []string, model.XdsLogDetails, bool, error) {
-	if !cdsNeedsPush(updates, proxy) {
-		return nil, nil, model.DefaultXdsLogDetails, false, nil
+	if features.DeltaXds {
+		updatedClusters, removedClusters, logs := c.Server.ConfigGenerator.BuildDeltaClusters(proxy, updates, w)
+		return updatedClusters, removedClusters, true, logs, nil
```
howardjohn, line 79: I don't think `true` is always correct here. BuildDeltaClusters can produce deltas OR sotw. if it does produce sotw, then we need the ads logic to automatically determine the removedClusters.

pilot/pkg/model/context.go
```diff
@@ -208,10 +210,10 @@ var DefaultXdsLogDetails = XdsLogDetails{}
...
 // or no response is preferred.
 type XdsResourceGenerator interface {
-	Generate(proxy *Proxy, push *PushContext, w *WatchedResource, updates *PushRequest) (Resources, XdsLogDetails, error)
-
-	// GenerateDeltas returns the changed and removed resources, along with whether or not delta was actually used.
-	GenerateDeltas(proxy *Proxy, push *PushContext, updates *PushRequest, w *WatchedResource) (Resources, []string, XdsLogDetails, bool, error)
+	// Generate builds the xds resources for given proxy. Resources contain the newly added/updated resources.
+	// DeletedResources typically used in incremental Xds protocols contain the resources that have been deleted.
+	// It also indicates whether the response is delta or SoTW.
+	Generate(proxy *Proxy, push *PushContext, w *WatchedResource, updates *PushRequest) (Resources, DeletedResources, bool, XdsLogDetails, error)
```
howardjohn, line 216: I almost wonder if we should return a struct as the response at this point... I am fine either way.

howardjohn (commented):
lgtm aside from my comment

pilot/pkg/xds/cds.go
howardjohn: Also this brings up an interesting point - I am pretty sure this means we will be missing removals. But no tests will ever fail when we miss removals...

pilot/pkg/model/context.go
adiprerepa: I actually did this in the initial implementation but decided not to because it seemed too tedious. up to @ramaraochavali

pilot/pkg/networking/core/configgen.go
```diff
@@ -39,8 +39,7 @@ type ConfigGenerator interface {
 
 	// BuildDeltaClusters returns both a list of resources that need to be pushed for a given proxy and a list of resources
 	// that have been deleted and should be removed from a given proxy. This is Delta CDS output.
-	BuildDeltaClusters(node *model.Proxy, push *model.PushContext, updates *model.PushRequest,
-		watched *model.WatchedResource) ([]*discovery.Resource, []string, model.XdsLogDetails, bool)
+	BuildDeltaClusters(node *model.Proxy, updates *model.PushRequest, watched *model.WatchedResource) ([]*discovery.Resource, []string, model.XdsLogDetails)
```
adiprerepa, line 42: We need to return whether or not delta was used form `BuildDeltaClusters`, as was done in the inital PR. We do not know if delta was actually used, generation may have fallen back on sotw

adiprerepa (commented):
I'm fine with this as long as you can confirm delta still works

pilot/pkg/networking/core/configgen.go
ramaraochavali: Ack. I am thinking about what is the best way to do it. Will make that change.

pilot/pkg/model/context.go
ramaraochavali: I agree it seems tedious to access from struct but I see @howardjohn's point. Let us leave it like this for now. We can do it later if we have add another return value.

ramaraochavali:
> I'm fine with this as long as you can confirm delta still works

Are there any manual testing needs to be done to confirm this?

pilot/pkg/model/context.go
```diff
@@ -208,10 +210,10 @@ var DefaultXdsLogDetails = XdsLogDetails{}
 // Note: any errors returned will completely close the XDS stream. Use with caution; typically and empty
 // or no response is preferred.
 type XdsResourceGenerator interface {
-	Generate(proxy *Proxy, push *PushContext, w *WatchedResource, updates *PushRequest) (Resources, XdsLogDetails, error)
-
-	// GenerateDeltas returns the changed and removed resources, along with whether or not delta was actually used.
-	GenerateDeltas(proxy *Proxy, push *PushContext, updates *PushRequest, w *WatchedResource) (Resources, []string, XdsLogDetails, bool, error)
+	// Generate builds the xds resources for given proxy. Resources contain the newly added/updated resources.
+	// DeletedResources typically used in incremental Xds protocols contain the resources that have been deleted.
+	// It also indicates whether the response is delta or SoTW.
```
hzxuzhonghu, line 215: can we say it is delta if no deleted resources? I understand delta xds can also contain no deleted resources, maybe i am wrong

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -74,96 +78,67 @@ func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
...
 			services = append(services, service)
 		}
 	}
-	clusterCacheStats := cacheStats{}
+	clusters, log := configgen.buildClusters(proxy, updates, services)
+	return clusters, deletedClusters, log
+}
+
+// buildClusters builds clusters for the proxy with the services passed.
+func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *model.PushRequest,
```
hzxuzhonghu, line 127: I like this abstraction

pilot/pkg/model/context.go
```diff
@@ -208,10 +210,10 @@ var DefaultXdsLogDetails = XdsLogDetails{}
 // Note: any errors returned will completely close the XDS stream. Use with caution; typically and empty
 // or no response is preferred.
 type XdsResourceGenerator interface {
-	Generate(proxy *Proxy, push *PushContext, w *WatchedResource, updates *PushRequest) (Resources, XdsLogDetails, error)
-
-	// GenerateDeltas returns the changed and removed resources, along with whether or not delta was actually used.
-	GenerateDeltas(proxy *Proxy, push *PushContext, updates *PushRequest, w *WatchedResource) (Resources, []string, XdsLogDetails, bool, error)
```
hzxuzhonghu, line 214: I donot see much benefit merging the two functions

hzxuzhonghu: It bring so many updates

pilot/pkg/model/context.go
ramaraochavali: Right now we can not say if deleted resources are empty that is because we do partial delta processing in CDS i.e. we do only delta logic when all updates are of either Service or DR. So that explicit flag is needed.

pilot/pkg/model/context.go
ramaraochavali: The benefit and the main reason I merged is to not force every generator to implement this. Another option I can think of is introduce a NoOpDeltaGenerator struct that just throws "Not implement" but that does not work as delta.go blindly  calls GenerateDeltas if we have that method but it expects some response. 

Is there a way to avoid that boiler plate code for every generator?

hzxuzhonghu: But actually, you need to change all the existing empty implementations

hzxuzhonghu: Compared to a bool value to indicate delta, I am more for the separate function

ramaraochavali: This bool value is not a permanent thing. Since we are doing incrementally we needed this. After all is done, we can remove this and mark a generator as "Delta" generator so that we just process "removed resources". Even for CDS today it is needed.

howardjohn: IMO it is permanent. Even years from now, there will still be cases where the generators need to just rebuild everything (ex - mesh config changes)

pilot/pkg/xds/cds.go
howardjohn: comment still applies

pilot/pkg/model/context.go
ramaraochavali: But in that case can not we rely on removedResources?

pilot/pkg/xds/cds.go
ramaraochavali: even after moving the logic to cds? Are you saying in future it can still return sotw and hence we need to have that flag of usedDelta?

pilot/pkg/model/context.go
howardjohn: I do not think so. Say a mesh config updates. In CDS, we would enter the current logic for buildOutboundClusters (iterate over all services/DR and make clusters). At this point we have a list of clusters; removedResources is empty.

However, in reality, we may have removed some clusters! Since delta=true and removedResources=empty, we do not remove any. This is wrong.

Instead, we need to compute the removedResources from `watchedResources.Names - newClusters.Names`

howardjohn: Basically _someone_ needs to compute which clusters are removed. This means either the generators are smart enough to perfectly handle 100% of of possible pushes (never going to happen IMO), or we have a fallback logic of `watchedResources.Names - newClusters.Names` in delta.go

ramaraochavali: Let me go through in detail.  But if it is permanent flag it is  even needed GenerateDeltas. So we can not avoid the boolean by having two separate functions @hzxuzhonghu

adiprerepa: however, we can avoid having `RemovedResources` in `Generate` if we have a separate API. IMO it's cleaner to keep the existing `Generate` as it is API and not have `RemovedResources` there, delta seems like a separate thing altogether.

ramaraochavali: Ok. If every one things that is better way - I can leave it like that. I have no strong preference. Let me carve out the general refactors in to a separate PR and close this one.

ramaraochavali:
Closing this in favour of https://github.com/istio/istio/pull/34513

PR #34480, opened 2021-08-02
allow cluster-local shards from Service/WorkloadEntry store

stevenctl:
When we build endpoint shards we key them with `registry.Cluster` for the ServiceEntry store, which is always `""`:

https://github.com/istio/istio/blob/13be95e3b974cbaf1e962ed7f7ab8ed5a8db2ba8/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go#L471-L475

That means any proxy that properly sets the `CLUSTER_ID` meta will exclude any serviceentry store endpoints for a cluster-local service. 

I _think_ the places in the code where we would use `clusterID == ""` to indicate ServiceEntry store have mostly been cleaned up – it may be better if we instead start to return a clusterID from the ServiceEntry store and rely on `Provider()` to differentiate. 

Another discussion is whether a WorkloadEntry endpoint is "cluster local"; it's registered in the cluster. But it doesn't reside in the cluster; with this change cross-cluster WorkloadEntry would be allowed under "cluster local".


howardjohn (commented):
> Another discussion is whether a WorkloadEntry endpoint is "cluster local"; it's registered in the cluster. But it doesn't reside in the cluster; with this change cross-cluster WorkloadEntry would be allowed under "cluster local".

This doesn't seem quite right to me... Is that how it would work though? I thought we read WorkloadEntry from all cluster now, so they will go to the cluster they belong to?

stevenctl:
Which part:
1. A `WorkloadEntry` based endpoint is cluster-local to the cluster it is registered in
2. with this change cross-cluster WorkloadEntry would be allowed under "cluster local".

2 is definitely not right – this quick fix just makes it so we don't drop _everything_ from serviceentry store when we use cluster.local. The proper fix means setting the cluster ID.

howardjohn:
How I think it should work is:

* A VM declares its a part of some cluster, say cluster-x
* When it connects, we create the Workload in cluster-x. All other clusters read it
* The WorkloadEntry is cluster-local to other workloads in cluster-x only.

Is that what you are thinking as well?

stevenctl:
Agreed  – I will say though that it would "connect to cluster-x" rather than declare it's part of cluster-x; but same thing really.

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -292,12 +292,13 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
 	}
 	// The shards are updated independently, now need to filter and merge for this cluster
 	for _, clusterID := range keys {
-		endpoints := shards.Shards[clusterID]
 		// If the downstream service is configured as cluster-local, only include endpoints that
-		// reside in the same cluster.
-		if isClusterLocal && (cluster.ID(clusterID) != b.clusterID) {
+		// reside in the same cluster. A clusterID of "" may indicate this is a ServiceEntry registry and should be
+		// included.
+		if clusterID != "" && isClusterLocal && (cluster.ID(clusterID) != b.clusterID) {
```
hzxuzhonghu, line 298: With this condition, all WLE can be visible even the service is cluster local.

hzxuzhonghu:
https://github.com/istio/istio/blob/13be95e3b974cbaf1e962ed7f7ab8ed5a8db2ba8/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go#L471-L475
could we return real clusterID which the wle connects?

pilot/pkg/xds/endpoint_builder.go
stevenctl: Yeah; won't merge as is. Planning on actually assigning a cluster-id.

stevenctl:
need to add a test - probably also need to set an actual cluster ID on the MemoryStore

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -471,7 +471,7 @@ func (s *ServiceEntryStore) Provider() provider.ID {
 func (s *ServiceEntryStore) Cluster() cluster.ID {
 	// DO NOT ASSIGN CLUSTER ID to non-k8s registries. This will prevent service entries with multiple
 	// VIPs or CIDR ranges in the address field
-	return ""
+	return s.clusterID
```
hzxuzhonghu, line 474: above comment is stale

stevenctl:
So I think we support a Service in cluster-2 selecting a WorkloadEntry from cluster-1; at least that's what our FakeDiscoveryServer does. This means the endpoint shards for cluster-2 will actually have a cluster-1 endpoint. 

https://github.com/istio/istio/blob/072fe94219bddbeb7aeec32698e7feddab4164e1/pilot/pkg/xds/endpoint_builder.go#L295-L298

Currently the cluster-local logic is just based on the shard key; with https://github.com/istio/istio/pull/34487 we can use the `topology.istio.io/cluster` internal label to do the filtering though. 

WDYT @hzxuzhonghu? Or is the cross-cluster selection already wrong..

stevenctl:
Also, I'm not 100% sure the change is safe as is. I think we could end up overwriting endpoint shards; we need to change the shard-keys to not be cluster-ID but also include provider type.

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -293,12 +294,19 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
...
+
 		// If the downstream service is configured as cluster-local, only include endpoints that
 		// reside in the same cluster.
 		if isClusterLocal && (cluster.ID(clusterID) != b.clusterID) {
 			continue
 		}
 		for _, ep := range endpoints {
+			// cross-cluster WorkloadEntry may come from a shard with a different cluster ID
+			// AugmentLabels adds an internal label to indicate individual endpoint origin
+			if epCluster := ep.Labels[label.TopologyCluster.Name]; epCluster != "" && epCluster != b.clusterID {
```
nmittler, line 306: Should we just set the cluster field on the `IstioEndpoint`? Seems like the existing cluster-local logic would just work ... or am I missing something?

stevenctl: The above cluster-local logic is purely based off the shard names and doesn't look at istio endpoints. I don't think there is a cluster field, just this label.

stevenctl: I might try to push this into the DiscoverabilityPolicy so we don't have to dupe this logic across here and https://github.com/istio/istio/blob/462811d9d1503fdc6069b81f1812276b68dff2e7/pilot/pkg/networking/core/v1alpha3/cluster_builder.go#L502

nmittler: Ah I see ..  so the issue is that `b` can contain endpoints from multiple clusters? Then yes, I think we should remove the cluster-local check outside of the loop and instead have the check on a per-endpoint basis.

I had considered adding this to `DiscoverabilityPolicy`, but the issue is that the policy would have to do the check to see if the service is cluster-local each time, which is a bit costly.

stevenctl: `b.clusterId` is the proxy's cluster

The original issue was that ` shards.Shards` had all WorkloadEntry endpoints in a shard `""` which would never be usable in cluster-local mode. 

Now we set a cluster ID on that registry, but a new issue is that shard names need more info to separate k8s cluster-1 and serviceentry cluster-1; shard name can't be considered a safe cluster id..

Even when that's solved, we update the shards in cluster-1 containing a service with an endpoint from cluster-2's workloadentry (see AppendWorkloadHandler calls).

tl;dr, shards become an invalid way to determine which cluster an IstioEndpoint is a part of

nmittler: > Now we set a cluster ID on that registry, but a new issue is that shard names need more info to separate k8s cluster-1 and serviceentry cluster-1; shard name can't be considered a safe cluster id..

Ah got it ... so there are multiple service registries that attempt to update the same shard but with different endpoints, causing collisions. Makes sense. We should definitely break any reliance on shard being a cluster ID.

> Even when that's solved, we update the shards in cluster-1 containing a service with an endpoint from cluster-2's workloadentry (see AppendWorkloadHandler calls).

I'm not quite sure I understand what you mean here. Can you clarify?

stevenctl: I thought we could have a Service in cluster-a select a WorkloadEntry in cluster b

I don't think it can actually happen:

https://github.com/istio/istio/blob/b03691f9e0cc1d8f9560ab72ad75cbe326ff5f36/pilot/pkg/serviceregistry/kube/controller/multicluster.go#L204-L205

But our FakeDiscoveryServer in tests does that

https://github.com/istio/istio/blob/e24ae0a21c892c604195f8bc7125a7cc2746f568/pilot/pkg/xds/fake.go#L251-L258

stevenctl: So in that case – rather than relying completely on the label, maybe we could have something like 

```
type shardKey struct {cluster cluster.ID provider provider.ID}
``

Comment (nmittler) [reply pilot/pkg/xds/endpoint_builder.go:306]: > I thought we could have a Service in cluster-a select a WorkloadEntry in cluster b

Not sure I follow ... Doesn't Istio have a single merged definition of a given service, with endpoints spread across multiple clusters?

**Comment by hzxuzhonghu** [comment]
I think service select pods/workloadentry should be cluster centric, otherwise the select logic would be super complex and may even bring duplicate endpoints.

File: pilot/pkg/serviceregistry/aggregate/controller.go [hunk 4]
```diff
@@ -167,19 +167,22 @@ func (c *Controller) GetService(hostname host.Name) (*model.Service, error) {
...
+func mergeService(dst, src *model.Service, srcRegistry serviceregistry.Instance) {
 	dst.Mutex.Lock()
+	defer dst.Mutex.Unlock()
 	if dst.ClusterVIPs == nil {
 		dst.ClusterVIPs = make(map[cluster.ID]string)
 	}
-	dst.ClusterVIPs[srcCluster] = src.Address
-	dst.Mutex.Unlock()
+	// prefer the k8s VIP where possible
+	if srcRegistry.Provider() == provider.Kubernetes || dst.ClusterVIPs[srcRegistry.Cluster()] == "" {
```
hzxuzhonghu, line 183: nit: We have checked provider in L163 https://github.com/istio/istio/pull/34480/files#diff-4ab8bb2188afa1dc914f11b5eaab7e1a544454457d25ef3841fdb6f2c23b1d7bR163

hzxuzhonghu (commented):
What is  `ShardKey` provider used for? I donot look at the test, is it the only use?

stevenctl:
> I think service select pods/workloadentry should be cluster centric, otherwise the select logic would be super complex and may even bring duplicate endpoints.

Agreed – we had code in the tests that would allow this to happen, but the actual implementation doesn't. 

> What is ShardKey provider used for? I donot look at the test, is it the only use?

The ShardKey must include both cluster and provider. 

If we have a ServiceEntry store for `cluster-1`, the endpoint shards can't be keyed by `registry.Cluster()`, otherwise the k8s registry and serviceentry store could overwrite each other.

pilot/pkg/serviceregistry/aggregate/controller.go
stevenctl: Unless it's confusing to leave this unused branch, can we keep it for safety? No strong preference.

hzxuzhonghu:
What's the current behavior when hostnames of SE and Service are equal? I guess we may generate duplicate eds since cross-registry selection is enabled. 

Not related though

hzxuzhonghu (approved):

PR #34484, opened 2021-08-03
respect cert TTL in client side for k8s RA

irisdingbj:
user can specify the workload cert TTL via `SECRET_TTL`, however this option is ignored if we use K8s RA as cert provider. 
cert-manager supports to use `experimental.cert-manager.io/request-duration` annotation in  k8s CSR, So this PR will add the annotation to the CSR istio generated to support the eco-system.

### DIFF
--- a/pkg/config/constants/constants.go
+++ b/pkg/config/constants/constants.go
@@ -125,4 +125,6 @@ const (
 	// CertProviderNone does not create any certificates for the control plane. It is assumed that some external
 	// load balancer, such as an Istio Gateway, is terminating the TLS.
 	CertProviderNone = "none"
+
+	RequestLifeTimeAnnotationForCertManager = "experimental.cert-manager.io/request-duration"
 )
--- a/pkg/config/constants/constants.go
+++ b/pkg/config/constants/constants.go
@@ -125,6 +125,4 @@ const (
 	// CertProviderNone does not create any certificates for the control plane. It is assumed that some external
 	// load balancer, such as an Istio Gateway, is terminating the TLS.
 	CertProviderNone = "none"
-
-	RequestLifeTimeAnnotationForCertManager = "experimental.cert-manager.io/request-duration"
 )
--- a/security/pkg/k8s/chiron/controller.go
+++ b/security/pkg/k8s/chiron/controller.go
@@ -205,8 +205,9 @@ func (wc *WebhookController) upsertSecret(secretName, dnsName, secretNamespace s
 		return nil
 	}
 
+	requestedLifetime := time.Duration(0)
 	// Now we know the secret does not exist yet. So we create a new one.
-	chain, key, caCert, err := GenKeyCertK8sCA(wc.clientset, dnsName, secretName, secretNamespace, wc.k8sCaCertFile, wc.certIssuer, true)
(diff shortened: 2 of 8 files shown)


shankgan:
Are we sure the Kubernetes CSR API gives us no mechanism to specify the cert TTL (I wasn't able to find the relevant API as well). We can use certManager annotations if there is no other canonical way to configure this.

security/pkg/k8s/chiron/utils.go
```diff
@@ -211,6 +213,9 @@ func submitCSR(clientset clientset.Interface,
 					SignerName: signerName,
 				},
 			}
+			if requestedLifetime.String() != "0s" {
```
howardjohn, line 216:
```suggestion
			if requestedLifetime != time.Duration(0) {
```

irisdingbj:
> Are we sure the Kubernetes CSR API gives us no mechanism to specify the cert TTL (I wasn't able to find the relevant API as well). We can use certManager annotations if there is no other canonical way to configure this.

Yes, the CSR API itself has no way to sepcify the cert TTL this is also the reason why cert-manager is using annotation in the CSR to support this.

security/pkg/k8s/chiron/controller.go
```diff
@@ -205,8 +205,9 @@ func (wc *WebhookController) upsertSecret(secretName, dnsName, secretNamespace s
 		return nil
 	}
 
+	requestedLifetime := time.Duration(0)
```
hzxuzhonghu, line 208: What does 0 mean?



It seems not quite right https://github.com/jetstack/cert-manager/blob/75d91bcb297d8a1506abd376ea378cf760c4f7f1/pkg/util/pki/csr.go#L380

irisdingbj: This is for Backward compatiblity : we will use  the `kubernetes.io/legacy-unknow' by default. In that case, no TTL can be specified.  So using a  '0' means we do not want to add related annotation to the CSR.  See : https://github.com/istio/istio/pull/34484/files#diff-6fa317e12eadb34d01780d845992c68fb30c869c9cc9127fd6458e728dacec1fR247

hzxuzhonghu: make sense

hzxuzhonghu (approved):

irisdingbj:
@shankgan @howardjohn PTAL, Thanks!

pkg/config/constants/constants.go
```diff
@@ -125,4 +125,6 @@ const (
 	// CertProviderNone does not create any certificates for the control plane. It is assumed that some external
 	// load balancer, such as an Istio Gateway, is terminating the TLS.
 	CertProviderNone = "none"
+
+	RequestLifeTimeAnnotationForCertManager = "experimental.cert-manager.io/request-duration"
```
shankgan, line 129: Please add comment here to explain why we are using CertManager Annotation. Also, could we move this constant somewhere inside the chiron package?

shankgan (approved):

pkg/config/constants/constants.go
irisdingbj: done

irisdingbj:
@howardjohn PTAL, Thanks!

howardjohn (approved):
approve t&r change

PR #34486, opened 2021-08-03
WIP: Adds ParsePEMCertificateChain for parsing certificate PEM chain. Used in istiod verification on Kube CSR responses

### LINKED ISSUE #34218: istiod RA Kubernetes CSR strategy does not parse certificate chains correctly
**Bug description**

Currently when using Kubernetes CSR resources to sign certificates, istiod only parses the leaf certificate, not the full signed chain returned. This causes verification errors when a  signer is using an intermediate certificate and will never work unless the intermediate certificate _is also_ propagated with the root CA. This negates the point of using an intermediate in the first place, and is impracticable for signers who's intermediates are only known during runtime.

The code causing this issue is [here](https://github.com/istio/istio/blob/77258b37054a4c33b657e61f6133308600dd300d/security/pkg/k8s/chiron/utils.go#L340).

This function should be replaced with a func which will parse the full chain, and any intermediates be included in the Verify further down. The following diff should produce the correct result:

```git
diff --git a/security/pkg/k8s/chiron/utils.go b/security/pkg/k8s/chiron/utils.go
index a9320851a3..00b349a760 100644
--- a/security/pkg/k8s/chiron/utils.go
+++ b/security/pkg/k8s/chiron/utils.go
@@ -293,10 +293,16 @@ func readSignedCertificate(client clientset.Interface, csrName string,
                return []byte{}, []byte{}, nil
        }
 
+       certsParsed, err := util.ParsePemEncodedCertificateChain(certPEM)
+       if err != nil {
+               return nil, nil, fmt.Errorf("decoding certificate failed: %w", err)
+       }
+
        caCert, err := readCACert(caCertPath)
        if err != nil {
                return nil, nil, fmt.Errorf("error when retrieving CA cert: (%v)", err)
        }
+
        // Verify the certificate chain before returning the certificate
        roots := x509.NewCertPool()
        if roots == nil {
@@ -305,12 +311,17 @@ func readSignedCertificate(client clientset.Interface, csrName string,
        if ok := roots.AppendCertsFromPEM(caCert); !ok {
                return nil, nil, fmt.Errorf("failed to append CA certificate")
        }
-       certParsed, err := util.ParsePemEncodedCertificate(certPEM)
-       if err != nil {
-               return nil, nil, fmt.Errorf("decoding certificate failed")
+
+       intermediates := x509.NewCertPool()
+       if len(certsParsed) > 1 {
+               for _, cert := range certsParsed[1:] {
+                       intermediates.AddCert(cert)
+               }
        }
-       _, err = certParsed.Verify(x509.VerifyOptions{
-               Roots: roots,
+
+       _, err = certsParsed[0].Verify(x509.VerifyOptions{
+               Roots:         roots,
+               Intermediates: intermediates,
        })
        if err != nil {
                return nil, nil, fmt.Errorf("failed to verify the certificate chain: %v", err)
diff --git a/security/pkg/pki/util/crypto.go b/security/pkg/pki/util/crypto.go
index f9cc30307d..fbe996d4ac 100644
--- a/security/pkg/pki/util/crypto.go
+++ b/security/pkg/pki/util/crypto.go
@@ -20,6 +20,7 @@ import (
        "crypto/rsa"
        "crypto/x509"
        "encoding/pem"
+       "errors"
        "fmt"
        "reflect"
 )
@@ -46,6 +47,35 @@ func ParsePemEncodedCertificate(certBytes []byte) (*x509.Certificate, error) {
        return cert, nil
 }
 
+// ParsePemEncodedCertificateChain constructs a slice of `x509.Certificate`
+// objects using the given a PEM-encoded certificate chain.
+func ParsePemEncodedCertificateChain(certBytes []byte) ([]*x509.Certificate, error) {
+       var (
+               certs []*x509.Certificate
+               cb    *pem.Block
+       )
+
+       for {
+               cb, certBytes = pem.Decode(certBytes)
+               if cb == nil {
+                       break
+               }
+
+               cert, err := x509.ParseCertificate(cb.Bytes)
+               if err != nil {
+                       return nil, fmt.Errorf("failed to parse X.509 certificate")
+               }
+
+               certs = append(certs, cert)
+       }
+
+       if len(certs) == 0 {
+               return nil, errors.New("no PEM encoded X.509 certificates parsed")
+       }
+
+       return certs, nil
+}
+
 // ParsePemEncodedCSR constructs a `x509.CertificateRequest` object using the
 // given PEM-encoded certificate signing request.
 func ParsePemEncodedCSR(csrBytes []byte) (*x509.CertificateRequest, error) {
```

[x] Security

**Expected behavior**

Using the Kubernetes CSR strategy should allow intermediates to be correctly parsed in `status.Certificate` CertificateSigningRequests.

Setup istio using Kubernetes CSR and use a signer who signs with an intermediate. The propagated CA contains the root CA, whilst `status.Certificate` contains the intermediate and leaf certificate.

---

I have a WIP branch with the above changes and am currently working through updating the tests. Please advice is the issue is correct for istio and I will open up a PR to fix once ready.

JoshVanL:

Fixes #34218 

This PR changes the istiod certificate verification when using the Kubernetes CSR strategy to include intermediates. Without this, most typical signers can't work as intermediates are shared during signing as the signed certificate chain.

NOTE: This PR hasn't updated unit tests yet.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [X] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/security/pkg/k8s/chiron/utils.go
+++ b/security/pkg/k8s/chiron/utils.go
@@ -293,13 +293,13 @@ func readSignedCertificate(client clientset.Interface, csrName string,
 	if len(certPEM) == 0 {
 		return []byte{}, []byte{}, nil
 	}
-	certParsed, err := util.ParsePemEncodedCertificate(certPEM)
+	certsParsed, err := util.ParsePemEncodedCertificateChain(certPEM)
 	if err != nil {
-		return nil, nil, fmt.Errorf("decoding certificate failed")
+		return nil, nil, fmt.Errorf("decoding certificates failed: %w", err)
 	}
+
 	caCert := []byte{}
-	certChain := []byte{}
-	certChain = append(certChain, certPEM...)
+
 	if appendCaCert && caCertPath != "" {
 		caCert, err = readCACert(caCertPath)
 		if err != nil {
@@ -313,15 +313,25 @@ func readSignedCertificate(client clientset.Interface, csrName string,
 		if ok := roots.AppendCertsFromPEM(caCert); !ok {
 			return nil, nil, fmt.Errorf("failed to append CA certificate")
 		}
-		_, err = certParsed.Verify(x509.VerifyOptions{
-			Roots: roots,
+
+		intermediates := x509.NewCertPool()
+		if len(certsParsed) > 1 {
+			for _, cert := range certsParsed[1:] {
+				intermediates.AddCert(cert)
+			}
+		}
+
+		_, err = certsParsed[0].Verify(x509.VerifyOptions{
+			Roots:         roots,
+			Intermediates: intermediates,
 		})
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to verify the certificate chain: %v", err)
 		}
-		certChain = append(certChain, caCert...)
+		certPEM = append(certPEM, caCert...)
 	}
-	return certChain, caCert, nil
+
+	return certPEM, caCert, nil
 }
 
 func checkDuplicateCsr(client clientset.Interface, csrName string) (*certv1.CertificateSigningRequest, *certv1beta1.CertificateSigningRequest) {
--- a/security/pkg/pki/util/crypto.go
+++ b/security/pkg/pki/util/crypto.go
@@ -20,6 +20,7 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"reflect"
 )
@@ -44,6 +45,35 @@ func ParsePemEncodedCertificate(certBytes []byte) (*x509.Certificate, error) {
 	}
 
 	return cert, nil
+}
+
+// ParsePemEncodedCertificateChain constructs a slice of `x509.Certificate`
+// objects using the given a PEM-encoded certificate chain.
+func ParsePemEncodedCertificateChain(certBytes []byte) ([]*x509.Certificate, error) {
+	var (
+		certs []*x509.Certificate
+		cb    *pem.Block
+	)
+
+	for {
+		cb, certBytes = pem.Decode(certBytes)
+		if cb == nil {
+			break
+		}
+
+		cert, err := x509.ParseCertificate(cb.Bytes)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse X.509 certificate")
+		}
+
+		certs = append(certs, cert)
+	}
+
+	if len(certs) == 0 {
+		return nil, errors.New("no PEM encoded X.509 certificates parsed")
+	}
+
+	return certs, nil
 }
 
 // ParsePemEncodedCSR constructs a `x509.CertificateRequest` object using the


security/pkg/k8s/chiron/utils.go
```diff
@@ -313,15 +313,25 @@ func readSignedCertificate(client clientset.Interface, csrName string,
...
+
+		_, err = certsParsed[0].Verify(x509.VerifyOptions{
+			Roots:         roots,
+			Intermediates: intermediates,
 		})
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to verify the certificate chain: %v", err)
 		}
-		certChain = append(certChain, caCert...)
+		certPEM = append(certPEM, caCert...)
```
JoshVanL, line 331: Question: Why are we appending the CA certificate to the chain here? Is this not shared out of band with cert gRPC clients?

shankgan: Not really. Today, the only way that the mesh infers trustAnchors for in-mesh mTLS is by looking at the root of the signed workload cert chain. Since the initial signer was the default "legacy-unknown" signer who did not return the entire  cert-chain (including the CA-cert), this code was relevant. 

For most custom signers, we don't need to append the ca cert since it will be a part of the cert-chain. We could have a env configuration to control this

shankgan (approved):

shankgan:
approve if it passes tests. There is an e2e test in tests/integration/security/external_ca which starts an external signer in a go-routine. Would be great to amend that test to use intermediate certs. @irisdingbj

security/pkg/k8s/chiron/utils.go
JoshVanL: I see. It is surprising to me (and concerning?) that clients are intended to blindly trust whatever CA is returned here. Would this not be a vector for abuse?

I assumed all CA trust was done via the config map propagation by istiod or other means.

Very well could be well off here.

JoshVanL:
Anyone, please feel free to pick it up from here. I likely won't have time to cross the ts on this PR this week.

shankgan:
> Anyone, please feel free to pick it up from here. I likely won't have time to cross the ts on this PR this week.

Will take a stab at this

security/pkg/k8s/chiron/utils.go
JoshVanL: If a signer was able to hijack CSRs and istiod did not have `appendCaCert` enabled, can a malicious actor start making clients trust whatever they want?

Could this also be true even if `appendCACert` is enabled?

shankgan: It is a valid concern, although not sure how it necessarily opens up a larger attack surface area. Also, our CA functionality was limited until now, since we mainly relied on citadel in OSS. With the RA functionality, this question becomes more relevant. 

Note that we also configure additional trustAnchors through the MeshConfig config-map, though this is newer functionality. Perhaps our default behavior should be to check if the root of the returned cert-chain from the CA matches with one of these trustAnchors...

shankgan: going point reg the malicious signer. Reinforces my idea that when using Istiod as RA, we mandate that the signer should affix their CA cert to the signed certChain and we compare this against our own configured TrustAnchors (in meshConfig or mounted to Istiod)

JoshVanL: > Perhaps our default behavior should be to check if the root of the returned cert-chain from the CA matches with one of these trustAnchors...

This makes sense to me :+1:

JoshVanL: > going point reg the malicious signer. Reinforces my idea that when using Istiod as RA, we mandate that the signer should affix their CA cert to the signed certChain and we compare this against our own configured TrustAnchors (in meshConfig or mounted to Istiod) to confirm the identity of the CA

This doesn't require the signer to send out their CA though right? The RA should be able to correctly verify a cert chain without the CA since it already has it as an entry in the TrustAnchors.. Just the same as normal TLS.

JoshVanL: Coming from the cert-manager side, we are apprehensive about including the CA in the chain response as it basically allows people to shoot themselves in the foot by allowing them to incorrectly blindly trust that CA.

Whilst sending the CA isn't bad in isolation, not sending it forces users to do the right thing.

irisdingbj:
> LGTM it passes existing tests + unit tests. There is an e2e test in tests/integration/security/external_ca which starts an external signer in a go-routine. Would be great to amend that test to use intermediate certs (not a requirement for this PR). @irisdingbj

sure, will enhance our csrctrl to take care of this.

security/pkg/k8s/chiron/utils.go
irisdingbj: > It is a valid concern, although not sure how it necessarily opens up a larger attack surface area. Also, our CA functionality was limited until now, since we mainly relied on istiod. With the RA functionality, this question becomes more relevant.
> 
> Note that we also configure additional trustAnchors through the MeshConfig config-map, though this is newer functionality. Perhaps our default behavior should be to check if the root of the returned cert-chain from the CA matches with one of these trustAnchors...

Does this mean admin needs to upload all of their root certs in meshconfig with mapping for the signers? seems our current mesh config API does not support this, we may need to change it and think about how to distribute them efficiently.

irisdingbj:
> > LGTM it passes existing tests + unit tests. There is an e2e test in tests/integration/security/external_ca which starts an external signer in a go-routine. Would be great to amend that test to use intermediate certs (not a requirement for this PR). @irisdingbj
> 
> sure, will enhance our csrctrl to take care of this.

talked with @shankgan and I will work on this later to get this moved on.

PR #28126, opened 2020-10-20
label system namespace with default network for cluster

stevenctl:
[RFC: Default network for cluster](https://docs.google.com/document/d/19PPGzw04gjk4evQA85oscYbJBOS6G8DMwc2q_UNBsZA/edit#)

This PR should make [integ-pilot-multicluster-tests_istio](https://prow.istio.io/?repo=istio%2Fistio&pull=28126&job=integ-pilot-multicluster-tests_istio) pass.

PR #34487, opened 2021-08-03
Added network label

hzxuzhonghu:

With this, we can support weighted traffic routing to different networks.

```
kind: DestinationRule
metadata:
  name: httpbin
spec:
  host: httpbin
  trafficPolicy:
    loadBalancer:
      simple: LEAST_CONN
  subsets:
  - name: network1
    labels:
      topology.istio.io/network: network1
  - name: kubernetes
    labels:
      topology.istio.io/network: network2

```

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller_test.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller_test.go
@@ -44,6 +44,7 @@ import (
 	"istio.io/istio/pkg/config/mesh"
 	"istio.io/istio/pkg/config/protocol"
 	kubelib "istio.io/istio/pkg/kube"
+	"istio.io/istio/pkg/network"
 	"istio.io/istio/pkg/spiffe"
 	"istio.io/istio/pkg/test"
 	"istio.io/istio/pkg/test/util/retry"
@@ -315,6 +316,7 @@ func TestController_GetPodLocality(t *testing.T) {
 
 func TestGetProxyServiceInstances(t *testing.T) {
 	clusterID := cluster.ID("fakeCluster")
+	networkID := network.ID("fakeNetwork")
 	for mode, name := range EndpointModeNames {
 		mode := mode
 		t.Run(name, func(t *testing.T) {
@@ -322,6 +324,8 @@ func TestGetProxyServiceInstances(t *testing.T) {
 				Mode:      mode,
 				ClusterID: clusterID,
 			})
+			// add a network ID to test endpoints include topology.istio.io/network label
+			controller.network = networkID
 			defer controller.Stop()
 			p := generatePod("128.0.0.1", "pod1", "nsa", "foo", "node1", map[string]string{"app": "test-app"}, map[string]string{})
 			addPods(t, controller, fx, p)
@@ -404,7 +408,6 @@ func TestGetProxyServiceInstances(t *testing.T) {
 			})
 
 			expected := &model.ServiceInstance{
-
 				Service: &model.Service{
 					Hostname:        "svc1.nsa.svc.company.com",
 					Address:         "10.0.0.1",
@@ -426,9 +429,11 @@ func TestGetProxyServiceInstances(t *testing.T) {
 						NodeRegionLabelGA:          "r",
 						NodeZoneLabelGA:            "z",
 						label.TopologyCluster.Name: clusterID.String(),
+						label.TopologyNetwork.Name: networkID.String(),
 					},
 					ServiceAccount:  "account",
 					Address:         "1.1.1.1",
+					Network:         networkID,
 					EndpointPort:    0,
 					ServicePortName: "tcp-port",
 					Locality: model.Locality{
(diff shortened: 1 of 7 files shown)


stevenctl:
Generally LGTM.

How does this work for WorkloadEntry? We have an explicit `network` field there that should be translated into this label. Not sure if it is today

stevenctl:
I think we'd need similar logic here: 

https://github.com/istio/istio/blob/eea231b5a2a7d88258a73de903fe556d7c970bce/pilot/pkg/serviceregistry/serviceentry/conversion.go#L367

hzxuzhonghu:
Yes, will add them

hzxuzhonghu:
@stevenctl Another look?

pilot/pkg/serviceregistry/util/label/label.go
```diff
@@ -0,0 +1,53 @@
...
+
+// AugmentLabels adds additional labels to the those provided.
+func AugmentLabels(in labels.Instance, clusterID cluster.ID, locality string, networkID network.ID) labels.Instance {
+	// Copy the original labels to a new map.
+	out := make(labels.Instance, len(in)+5)
+	for k, v := range in {
+		out[k] = v
+	}
+
+	// Don't need to add label.TopologyNetwork.Name, since that's already added by injection.
```
stevenctl, line 35: Confused what we mean by this comment? Seems like we do add it on L50.

pilot/pkg/serviceregistry/kube/controller/controller_test.go
```diff
@@ -556,6 +563,7 @@ func TestGetProxyServiceInstances(t *testing.T) {
 				ServicePort: &model.Port{Name: "tcp-port", Port: 8080, Protocol: protocol.TCP},
 				Endpoint: &model.IstioEndpoint{
 					Address:         "129.0.0.2",
+					Network:         networkID,
```
stevenctl, line 566: Can we also test that when a Pod already has the network label set, it gets preserved (rather than using the controller-default)

stevenctl: Probably easier to do so in `TestNewEndpointBuilderTopologyLabels`

pilot/pkg/serviceregistry/serviceentry/conversion.go
```diff
@@ -241,28 +243,30 @@ func convertEndpoint(service *model.Service, servicePort *networking.Port,
...
 
-	tlsMode := getTLSModeFromWorkloadEntry(endpoint)
+	tlsMode := getTLSModeFromWorkloadEntry(wle)
 	sa := ""
-	if endpoint.ServiceAccount != "" {
-		sa = spiffe.MustGenSpiffeURI(service.Attributes.Namespace, endpoint.ServiceAccount)
+	if wle.ServiceAccount != "" {
+		sa = spiffe.MustGenSpiffeURI(service.Attributes.Namespace, wle.ServiceAccount)
 	}
+	labels := labelutil.AugmentLabels(wle.Labels, clusterID, wle.Locality, network.ID(wle.Network))
```
stevenctl, line 251: We probably need a lot of the same logic that we embed into the `serviceregistry/kube/controller/` to determine which network the endpoint is on:
* cidr range (sort of legacy, part of meshNetworks, but still valid)
* topology label. in this case the `wle.Network` is probably more valid, but we might want to support both
* default network for cluster, if not set on WLE (determined from `topology.istio.io/network` label on the system namespace; see https://github.com/istio/istio/pull/28126)

This would probably require a bit of restructuring to allow re-using this. And now we can run `serviceentry` registries, so we have to support this per-cluster for each registry type. Probably best as a follow-up task since this is already brand new functionality.

stevenctl:
LGTM – the workloadentry part still needs a few things but we can iterate; this doesn't break anything

stevenctl (approved):

pilot/pkg/serviceregistry/util/label/label.go
hzxuzhonghu: This is copied from previous implement, i will address separately

pilot/pkg/serviceregistry/kube/controller/controller_test.go
hzxuzhonghu: I have added cases there

pilot/pkg/serviceregistry/serviceentry/conversion.go
hzxuzhonghu: >cidr range (sort of legacy, part of meshNetworks, but still valid)

Good point

>topology label. in this case the wle.Network is probably more valid, but we might want to support both

This is already supported, wle.network override that one from wle.labels.

>default network for cluster, if not set on WLE (determined from topology.istio.io/network label on the system namespace; see label system namespace with default network for cluster #28126)

It implicitly means wle network should be same as the cluster.  I am not sure if this is right.

I think the order should be: wle.Network > wle.Labels >  cluster namespace label. > cidr(maybe should not support it as it has been deprecated)

PR #34494, opened 2021-08-03
Allow cached response in initial discovery request

howardjohn:
This is a small regression in the change to the token invalidation - .Start is not set on push requests so we lose out on one of the biggest sources of cache hits.


adiprerepa (approved):
:goat:

howardjohn:
Hold for assert test to pass

PR #34353, opened 2021-07-27
support multi signer in istiod

irisdingbj:

PR #34496, opened 2021-08-03
Fix CSR test to run on 1.19+

howardjohn:
https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/

Introduced in 1.19, not 1.18
Regression in https://github.com/istio/istio/pull/34353/files


lei-tang (approved):

shankgan:
thanks for catching this

PR #34498, opened 2021-08-03
Re-enable CDS cache now that envoyfilter key is fixed

howardjohn:
Since https://github.com/istio/istio/pull/34404


pilot/pkg/features/pilot.go
```diff
@@ -60,7 +60,7 @@ var (
 
 	PushThrottle = env.RegisterIntVar(
 		"PILOT_PUSH_THROTTLE",
-		100,
+		1,
```
hzxuzhonghu, line 63: I guess this is for debug

howardjohn: oops, not intended to merge. Thanks!

hzxuzhonghu (approved):
Suggest not backport to 1.11, in case potential bug.

howardjohn:
agreed, this will give 3mo to test CDs and rds

PR #34512, opened 2021-08-04
adding debug/proxy-info to get proxy istio version

ChenglongGao327:
adding debug/proxy-info to get proxy istio version

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/proxy/proxyinfo.go
+++ b/pkg/proxy/proxyinfo.go
@@ -19,45 +19,32 @@ import (
 	"encoding/json"
 	"fmt"
 
-	"istio.io/istio/pilot/pkg/xds"
 	"istio.io/istio/pkg/kube"
-	istioVersion "istio.io/pkg/version"
+	"istio.io/pkg/version"
 )
 
-type sidecarSyncStatus struct {
-	// nolint: structcheck, unused
-	pilot string
-	xds.SyncStatus
-}
-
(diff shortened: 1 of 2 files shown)


pkg/proxy/proxyinfo.go
```diff
@@ -19,45 +19,32 @@ import (
 	"encoding/json"
 	"fmt"
 
-	"istio.io/istio/pilot/pkg/xds"
```
hzxuzhonghu, line 22: Great, this can decouple importing `istio.io/istio/pilot`

hzxuzhonghu (approved):

howardjohn (commented):
doesn't istioctl ps already show this info?

hzxuzhonghu:
I think it is to remove dependency on the pilot package. But `istioctl/pkg/writer/pilot/status.go` also depend on `xds.SyncStatus`.

howardjohn (changes requested):
Please explain why we need this since proxy status already has it. Thanks!

ChenglongGao327:
> Please explain why we need this since proxy status already has it. Thanks!

sorry, missing it , i think it decouple importing ```istio.io/istio/pilot```, did not notice that istioctl ps can get these.

PR #34519, opened 2021-08-04
Fix debug assert due to missing start time

howardjohn:


hzxuzhonghu:
I did not add the StartTime because i am not sure if it can cause assert failure

howardjohn:
> I did not add the StartTime because i am not sure if it can cause assert failure

I did not think it can, its exactly the same as a XDS request?

If it does we have a test now :slightly_smiling_face:

hzxuzhonghu (approved):

PR #34536, opened 2021-08-05
When the istioctl -h command is executed, available commands displays the alias

ChenglongGao327:
When users use istioctl, if executed **istioctl -h** echos the alias of the command, it will be more convenient to use.

1. old version： When the **istioctl -h** command is executed available commands does not show aliases.
```
...
Usage:
  istioctl [command]
Available Commands:
  proxy-config   Retrieve information about proxy configuration from Envoy [kube only]
  proxy-status   Retrieves the synchronization status of each Envoy in the mesh [kube only]
  upgrade        Upgrade Istio control plane in-place
  validate       Validate Istio policy and rules files
  verify-install Verifies Istio Installation Status
  version        Prints out build version information
...
```

2. for this PR ：When the **istioctl -h** command is executed available commands display alias.
```
...
Usage:
  istioctl [command]
Available Commands:
  admin, istiod           Manage control plane (istiod) configuration
  proxy-config, pc        Retrieve information about proxy configuration from Envoy [kube only]
  proxy-status, ps        Retrieves the synchronization status of each Envoy in the mesh [kube only]
...
```

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/istioctl/cmd/root.go
+++ b/istioctl/cmd/root.go
@@ -76,6 +76,34 @@ var (
 
 	// scope is for dev logging.  Warning: log levels are not set by --log_output_level until command is Run().
 	scope = log.RegisterScope("cli", "istioctl", 0)
+
+	// usageTemplate is nearly identical to the default template.
+	// only change available commands name to name and aliases.
+	usageTemplate = `Usage:{{if .Runnable}}
+  {{.UseLine}}{{end}}{{if .HasSubCommands}}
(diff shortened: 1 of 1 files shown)


ChenglongGao327:
@nmittler PTAL

Issue #27650, opened 2020-09-30
CDS is not updated as destination changes in VirtualService 


weibohe:
**Bug description**
After https://github.com/istio/istio/pull/20408, Istio is supposed to include all the destinations specified in VirtualService as part of the CDS. 

However, currently as VirtualService changes, Istio doesn't trigger a CDS update:
- on 1.6.11 https://github.com/istio/istio/blob/1.6.11/pilot/pkg/proxy/envoy/v2/ads_common.go#L115-L117
- on master https://github.com/istio/istio/blob/master/pilot/pkg/xds/cds.go#L33

As a result, when a new destination is added to the VirtualService, the route will be updated while the cluster won't, which will lead to 503 NR on the client side proxy.

We encounter this issue in production:
- As part of release pipeline, we perform ACA (automatic canary analysis), during which we shift a percent of traffic from `svc.svc-production` to `svc.svc-canary` & `svc.svc-canary-baseline` (two new namespaces spin up during the ACA and terminates after ACA).
- the caller only specify `svc-production/*` in its sidecar egress hosts and rely on Istio to figure out the clusters
- our deploy pipeline changes the VS to route a small percent of traffic to the new destinations, but we found that the clusters for the new destinations were not pushed to the client side proxy, while the route was updated to send traffic to the new clusters
- as a result, all the traffic allocated to the new clusters saw 503 NR
- only when the old pilot connection terminates and a new connection is established, will the client proxy get the new clusters

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Expected behavior**
VirtualService update triggers CDS pushes.

**Steps to reproduce the bug**
Download 3 files from https://gist.github.com/weibohe/8e325bb5c1c331404fe597b4a7d1545b

run `repro.sh`

It basically set up a client & a server and adds a new destination to the VS. 

**Version (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version` if you used Helm)**
1.6.11
1.7.3

**How was Istio installed?**
Demo installation

**Environment where bug was observed (cloud vendor, OS, etc)**
k8s

PR #34537, opened 2021-08-05
Skip push cds for sidecar when virtualService updated

hzxuzhonghu:
**Please provide a description of this PR:**

VirtualService and Gateway both donot influence sidecar cds.

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [x] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #27651, opened 2020-09-30
xds: push cds on vs update

brianwolfe:
Currently, VirtualService changes do not trigger CDS updates.
Istio is supposed to transitively include all destinations specified in
the VirtualService as part of the CDS. This is correctly calculated
after https://github.com/istio/istio/pull/20408 was included, but if the
change is actually made to the virtualservice and no subsequent changes
are made to the DestinationRule or Service, the new cluster will not
actually get pushed.

Only LDS & RDS trigger, so we will route to a nonexistent cluster,
resulting in 503 NR errors in envoy until a full push of clusters is
triggered again.

Fixes https://github.com/istio/istio/issues/27650

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any changes that may affect Istio users.

PR #34552, opened 2021-08-05
Revert "Skip push cds for sidecar when virtualService updated"

howardjohn:
Reverts istio/istio#34537

This is to add back https://github.com/istio/istio/pull/27651 which fixed https://github.com/istio/istio/issues/27650

We can probably filter many VS updates, but need a bit more complex logic I think


hzxuzhonghu (approved):

hzxuzhonghu:
I will investigate. I think with this, it can escalate the egress hosts of `sidecar`. It looks like a security concern.

howardjohn:
4 years later update... https://github.com/istio/istio/pull/55757

Issue #34566, opened 2021-08-06
bug: no labels on SE+DR subset == no match


howardjohn:
```
---
apiVersion: networking.istio.io/v1beta1
kind: ServiceEntry
metadata:
  name: redis
  namespace: default
spec:
  hosts:
  - httpbin.org
  ports:
  - number: 6978
    name: tcp
    protocol: TCP
  resolution: DNS
  location: MESH_EXTERNAL
  endpoints:
    - address: httpbin.org
      labels:
        foo: bar
---
apiVersion: networking.istio.io/v1beta1
kind: DestinationRule
metadata:
  name: redis-originate-tls
  namespace: default
spec:
  host: httpbin.org
  subsets:
  - name: redis-originate-TLS
    trafficPolicy:
      loadBalancer:
        simple: ROUND_ROBIN
      portLevelSettings:
      - port:
          number: 6978
        tls:
          mode: SIMPLE
          sni: httpbin.org
```

If remove
```
      labels:
        foo: bar
```

we end up with no endpoints matched

PR #34568, opened 2021-08-06
Fix subsets with no labels

howardjohn:
Fixes https://github.com/istio/istio/issues/34566
Regression from https://github.com/istio/istio/pull/33929

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -130,7 +130,11 @@ func (cb *ClusterBuilder) buildSubsetCluster(opts buildClusterOpts, destRule *co
 		clusterType = cluster.Cluster_ORIGINAL_DST
 	}
 	if !(isPassthrough || clusterType == cluster.Cluster_EDS) {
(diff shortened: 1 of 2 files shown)


stevenctl (approved):

stevenctl:
LGTM; but wouldn't the `len(0)` logic be better to have in ServiceInstancesByPort

ramaraochavali:
@howardjohn Thanks for fixing it - Sorry, we discussed this in that PR but I thought it is same. Is this failing because of it is missing this https://github.com/istio/istio/blob/master/pilot/pkg/model/push_context.go#L1918 when there are no labels?

And one more question on the issue you created, if subset does not have have any labels, what is the use of subsets? Just for my understanding

ramaraochavali:
> LGTM; but wouldn't the len(0) logic be better to have in ServiceInstancesByPort

I also think it is better to add there - otherwise every caller would have to do this. But given this subtle regression - I am scared to touch it :-)

howardjohn:
> And one more question on the issue you created, if subset does not have have any labels, what is the use of subsets? Just for my understanding


Its all about applying different settings for different routes. Not super common but we do document it in our egress gw docs (despite it not being required)

PR #34580, opened 2021-08-07
Fuzzing: Add security fuzzer

AdamKorcz:
Adds a fuzzer that targets `istio.io/istio/security/pkg/pki/util.GenCSR`

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience

### DIFF
--- a/tests/fuzz/security_fuzzer.go
+++ b/tests/fuzz/security_fuzzer.go
@@ -1,0 +1,133 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// nolint: golint
+package fuzz
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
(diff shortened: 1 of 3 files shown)


tests/fuzz/security_fuzzer.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func FuzzGenCSR(data []byte) int {
+	f := fuzz.NewConsumer(data)
+	certOptions := util.CertOptions{}
+	err := f.GenerateStruct(&certOptions)
+	if err != nil {
+		return 0
+	}
+	_, _, _ = util.GenCSR(certOptions)
```
howardjohn, line 30: One concern: I think this is fairly expensive call (in CPU time) so we may end up not getting many iterations

howardjohn (commented):
GenCSR is called on the client side rather than on the server side I believe, so it may not be as interesting as some of the other codepaths that run on the server side, which involve parsing arbitrary input (and are of course super security sensitive). Those may be a bit harder to fuzz possibly

tests/fuzz/security_fuzzer.go
AdamKorcz: Locally it runs alright. I will keep an eye out for the stats from runs by OSS-fuzz.

AdamKorcz:
> GenCSR is called on the client side rather than on the server side I believe, so it may not be as interesting as some of the other codepaths that run on the server side, which involve parsing arbitrary input (and are of course super security sensitive). Those may be a bit harder to fuzz possibly

I have included a fuzzer for server side targets.

AdamKorcz:
Checking in to see if this one can be considered ready to be merged.

AdamKorcz:
@howardjohn @jacob-delgado Kindly requesting review of this to see if it can be approved.

howardjohn (approved):

PR #34590, opened 2021-08-09
Fully disable CDS cache

howardjohn:
See https://github.com/istio/istio/pull/34589 for context


stevenctl (approved):

stevenctl:
Have you tested this in master to see if removing the tokens fixed this?

howardjohn:
> Have you tested this in master to see if removing the tokens fixed this?
No, the leak only appeared after a few days. Otherwise hard to tell

PR #34595, opened 2021-08-09
Make the adduser evaluation in post-install less OS dependant

Nino-K:
This change makes the evaluation of `useradd` and `groupadd` less dependant on the OS type and more reliant on the availability of the command itself. The current `if` statement does not work in SUSE distros since SUSE also uses  `groupadd` and `useradd` same as redhat and centos.

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/tools/packaging/postinst.sh
+++ b/tools/packaging/postinst.sh
@@ -20,7 +20,7 @@ set -e
 umask 022
 
 if ! getent passwd istio-proxy >/dev/null; then
-    if [ -x /usr/sbin/useradd ]; then
+    if command -v useradd >/dev/null; then
         groupadd --system istio-proxy
         useradd --system --gid istio-proxy --home-dir /var/lib/istio istio-proxy
     else
--- a/tools/packaging/postinst.sh
+++ b/tools/packaging/postinst.sh
@@ -20,7 +20,7 @@ set -e
 umask 022
 
 if ! getent passwd istio-proxy >/dev/null; then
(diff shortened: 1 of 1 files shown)


tools/packaging/postinst.sh
```diff
@@ -20,7 +20,7 @@ set -e
 umask 022
 
 if ! getent passwd istio-proxy >/dev/null; then
-    if  find /etc/*release | grep -Eiq 'centos|redhat'; then
+    if [ -x /usr/sbin/useradd ]; then
```
howardjohn, line 23: is `command -v useradd` more robust? useradd could be in any folder?

Nino-K: Both approaches are shell builtin which is ideal. While the -x looks for both the existence of the file and makes sure the file is executable. The command -v only displays the filename. Also, it might be a little cleaner to go with the -x in this context.
All the major distros such as Redhat, centos, and SUSE that leverage `useradd` and `groupadd` install these utilities in /usr/sbin/.

howardjohn: IMO relying on the path is a bad pattern. https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script recommends `command` (or even both in the 2nd answer)

howardjohn (approved):

PR #34604, opened 2021-08-10
add metadata to cluster cache key

ramaraochavali:
With out the proxy metadata certs are not considered in cache key and wrong cluster is delivered with incorrect UpstreamTLSContext.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -373,6 +367,14 @@ func (t *clusterCache) Key() string {
 		sort.Strings(nv)
 		params = append(params, nv...)
 	}
+	if t.proxyMetadata != nil {
+		md := make([]string, 0, len(t.proxyMetadata))
+		for mk, mv := range t.proxyMetadata {
+			md = append(md, mk+"~"+mv)
+		}
```
hzxuzhonghu, line 374: It can be a huge map, can you add the one that really matters

ramaraochavali: currently certificates matter. But not sure what else matters - I can trim down to certs but do we know/sure we do not care about any thing else?

howardjohn:
what we should ideally do is never access the metadata (or anything) from
the Proxy object. Instead transform the Proxy into the key type and pass
that around.

Then it's impossible to use something not in the key.

We do this for EDS

ramaraochavali:
> We do this for EDS

Can you point me to that? I will take a look.

howardjohn:
> Can you point me to that? I will take a look.

https://github.com/istio/istio/blob/072fe94219bddbeb7aeec32698e7feddab4164e1/pilot/pkg/xds/endpoint_builder.go#L87

OK it actually does include proxy.. but we only use it in 2 places, and I bet we could clean those up.

ramaraochavali:
@howardjohn @hzxuzhonghu moved metadata access to separate keys. PTAL

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -87,20 +87,55 @@ type MutableCluster struct {
...
-	cache model.XdsCache
+	proxy         *model.Proxy
+	req           *model.PushRequest
+	cache         model.XdsCache
+	metadataCerts MetadataCerts
+	clusterID     string
 }
 
 // NewClusterBuilder builds an instance of ClusterBuilder.
 func NewClusterBuilder(proxy *model.Proxy, req *model.PushRequest, cache model.XdsCache) *ClusterBuilder {
```
howardjohn, line 116: in EDS "EndpointBuilder" (equivilent here) *is* the cache key. This makes it ~impossible to misuse. We don't have to do it in this PR but I think thats the best long term direction

howardjohn (approved):

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
ramaraochavali: Yes. I agree. We will have to move to that model and remove proxy altogether.

PR #34614, opened 2021-08-10
grpc-agent template sets holdApplicationUntilProxyStarts

stevenctl:
It's possible for an app to race with `server.Serve(listener)` and the agent generating the bootstrap/xds proxy being ready.

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
@@ -962,6 +962,8 @@ data:
           {{- range $index, $container := .Spec.Containers  }}
           {{ if not (eq $container.Name "istio-proxy") }}
           - name: {{ $container.Name }}
+            annotations:
+              proxy.istio.io/config: '{"holdApplicationUntilProxyStarts": true}'
             env:
             - name: "GRPC_XDS_BOOTSTRAP"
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
Why do we need this here but not in non-proxyless? Shouldn't the app be robust against the race condition?

stevenctl:
We could make things a bit better by moving the bootstrap generation to an init container, but we would still likely fail if we can't create a connection to the xDS server specified by the bootstrap. 

Envoy handles this a lot better with retries, but for gRPC _servers_ we will hard fail. 

https://github.com/grpc/grpc-go/blob/01bababd83492b6eb1c7046ab4c3a4b1bcc5e9d6/xds/server.go#L185-L187

https://github.com/grpc/grpc-go/blob/01bababd83492b6eb1c7046ab4c3a4b1bcc5e9d6/xds/internal/xdsclient/client.go#L604-L608

Yes applications should be more robust against that (see #34594), but many won't be. I'm not 100% sure if this should be in a template, or a recommendation on how to get things working without modifying application code. Right now, if the app crashes, and that one container restarts, things will still sort of work. Other apps like `echo` fail to listen then move on.

stevenctl:
After thinking about this some more, it's probably better to let users set this themselves. We can document that this _may_ be necessary, unless they have the proper retry logic surrounding their `Serve` call.

PR #34624, opened 2021-08-11
Set timeout for http server

hzxuzhonghu:

To make http server resilient

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/security/pkg/stsservice/server/server.go
+++ b/security/pkg/stsservice/server/server.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httputil"
+	"time"
 
 	"istio.io/istio/pkg/security"
 	"istio.io/istio/security/pkg/stsservice"
@@ -81,8 +82,10 @@ func NewServer(config Config, tokenManager security.TokenManager) (*Server, erro
 	mux.HandleFunc(TokenPath, s.ServeStsRequests)
 	mux.HandleFunc(StsStatusPath, s.DumpStsStatus)
 	s.stsServer = &http.Server{
-		Addr:    fmt.Sprintf("%s:%d", config.LocalHostAddr, config.LocalPort),
-		Handler: mux,
+		Addr:        fmt.Sprintf("%s:%d", config.LocalHostAddr, config.LocalPort),
+		Handler:     mux,
+		IdleTimeout: 90 * time.Second, // matches http.DefaultTransport keep-alive timeout
+		ReadTimeout: 30 * time.Second,
 	}
 	ln, err := net.Listen("tcp", fmt.Sprintf("%s:%d", config.LocalHostAddr, config.LocalPort))
 	if err != nil {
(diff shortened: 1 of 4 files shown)


hzxuzhonghu:
There are security issues like a malicious user can attack the server by using up all the connections if no timeout set.

hzxuzhonghu:
@ramaraochavali If we declare a const for timeout, the constants package maybe imported everywhere. I did not see any project do it this way.

hzxuzhonghu:
FYI: https://blog.cloudflare.com/exposing-go-on-the-internet/

ramaraochavali (approved):
I did not mean constants in a

hzxuzhonghu:
@istio/wg-security-maintainers ptal

shankgan (approved):

shankgan:
@hzxuzhonghu - did you try using ReadHeaderTimeout? Also, is the 30 sec timeout large enough for XDS?

hzxuzhonghu:
xds is through grpc, this is for the http server.  `ReadHeaderTimeout` and `ReadTimeout` have slight difference, both works here,  `ReadTimeout` is the maximum duration for reading the entire request, including the body. I think it is safer, we have no large body sent to server.

howardjohn:
We do actually serve grpc using the http2 stack when using the port
multiplex (ie set gRPC port to "" and it will serve on the http server).
should make sure timeout doesn't apply to that (or applies to first byte vs
full stream)

hzxuzhonghu:
Let me look into that

security/pkg/stsservice/server/server.go
```diff
@@ -81,8 +82,10 @@ func NewServer(config Config, tokenManager security.TokenManager) (*Server, erro
 	mux.HandleFunc(TokenPath, s.ServeStsRequests)
 	mux.HandleFunc(StsStatusPath, s.DumpStsStatus)
 	s.stsServer = &http.Server{
-		Addr:    fmt.Sprintf("%s:%d", config.LocalHostAddr, config.LocalPort),
-		Handler: mux,
+		Addr:        fmt.Sprintf("%s:%d", config.LocalHostAddr, config.LocalPort),
+		Handler:     mux,
+		IdleTimeout: 90 * time.Second, // matches http.DefaultTransport keep-alive timeout
```
ryan-dayrit, line 87: Does it make sense for the 90/30 to come from config.IdleTimeout/config.ReadTimeout vs having it baked into the compiled code?

hzxuzhonghu: @ryan-dayrit What do you mean?

ryan-dayrit: - Extending the config struct to have these IdleTimeout / ReadTimeout props. 
- Setting default vals to 90 / 30 respectively 

```
IdleTimeout: config.IdleTimeoutSecs * time.Second
ReadTimeout: config.ReadTimeoutSecs * time.Second
```

hzxuzhonghu:
@howardjohn From https://golang.org/src/net/http/server.go#L937  the readtimeout and readheaderTimeout is same if only one is specified.  And it does not influence http2 stream, as we have enabled keep alive by default 30s.

PR #34625, opened 2021-08-11
Fix frequent detecting JWT policy

### LINKED ISSUE #34453: "detecting third-party JWT support" in cluster operator triggers way to much
This call is not cheap! it makes hundreds of requests to the API server. We should call it just once

kebe7jun:
Fix https://github.com/istio/istio/issues/34453

- [X] Installation

### DIFF
--- a/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
+++ b/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
@@ -313,17 +313,20 @@ func (r *ReconcileIstioOperator) Reconcile(_ context.Context, request reconcile.
 	if err != nil {
 		return reconcile.Result{}, nil
 	}
-	var jwtPolicy util.JWTPolicy
-	if jwtPolicy, err = util.DetectSupportedJWTPolicy(kubeClient); err != nil {
-		// TODO(howardjohn): add to dictionary. When resolved, replace this sentence with Done or WontFix - if WontFix, add reason.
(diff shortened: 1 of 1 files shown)


operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
```diff
@@ -313,17 +313,20 @@ func (r *ReconcileIstioOperator) Reconcile(_ context.Context, request reconcile.
...
-	var jwtPolicy util.JWTPolicy
-	if jwtPolicy, err = util.DetectSupportedJWTPolicy(kubeClient); err != nil {
-		// TODO(howardjohn): add to dictionary. When resolved, replace this sentence with Done or WontFix - if WontFix, add reason.
-		scope.Warnf("Failed to detect third-party JWT support: %v", err)
-	} else {
-		if jwtPolicy == util.FirstPartyJWT {
-			scope.Info("Detected that your cluster does not support third party JWT authentication. " +
-				"Falling back to less secure first party JWT. " +
-				"See " + url.ConfigureSAToken + " for details.")
+	if iopMerged.Status == nil || iopMerged.Status.Status == v1alpha1.InstallStatus_NONE {
```
howardjohn, line 316: I am not sure how this works? globalValues is just a local variable, it will not mutate the spec in the cluster (?). This means it will not actually be updated? So on the next reconcile we will have the wrong value I would think

PR #34633, opened 2021-08-11
Fix adsc wrong syncCh size

YonkaFang:

This PR fix a bug in adsc. Which will cause adsc hang when having received initial resource push.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/releasenotes/notes/34633.yaml
+++ b/releasenotes/notes/34633.yaml
@@ -4,4 +4,4 @@ area: traffic-management
 
 releaseNotes:
 - |
-  **Fixed** adsc wrong syncCh size which will make it be certain to hang
+  **Fixed** an issue in which ADS would hang due to the wrong `syncCh` size being provided.
--- a/releasenotes/notes/34633.yaml
+++ b/releasenotes/notes/34633.yaml
@@ -1,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+
+releaseNotes:
+- |
+  **Fixed** adsc wrong syncCh size which will make it be certain to hang
--- a/pkg/adsc/adsc.go
+++ b/pkg/adsc/adsc.go
@@ -20,6 +20,7 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
+	"math"
 	"net"
 	"os"
 	"sort"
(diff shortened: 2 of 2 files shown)


hzxuzhonghu (approved):

hzxuzhonghu:
@YonkaFangIt looks `syncCh` is read from `WaitConfigSync` and not used in istio, do you use it in your own code base?

YonkaFang:
> @YonkaFangIt looks `syncCh` is read from `WaitConfigSync` and not used in istio, do you use it in your own code base?

Currently adsc is used in mcp-over-xds configcontroller is istio(rather than our own code), but the func  `WaitConfigSync` is not called. I think there's some work for configController to do.

If syncCh is set to a correct size, adsc can normally works even when nobody cares abound the sync status/read the syncCh.

releasenotes/notes/34633.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+
+releaseNotes:
+- |
+  **Fixed** adsc wrong syncCh size which will make it be certain to hang
```
brian-avery, line 7:
```suggestion
  **Fixed** an issue in which ADS would hang due to the wrong `syncCh` size being provided.
```

brian-avery (approved):

brian-avery:
Looks good to me. Please take a look at the release note suggestion before removing the `do-not-merge/hold` label.

YonkaFang:
@brian-avery plz take a look. suggested release note change has been applied.

hzxuzhonghu:
@YonkaFang I come to think maybe it's better to remove the `syncCh` and the `WaitConfigSync `

hzxuzhonghu:
With just a HasSync method, it allows users make use of it easily like other controllers and config store do

YonkaFang:
> and

I also thinks the user-side should do the wait thing and keep the consistency among the controllers. I'm ajust afraid that anyone will use this func later.

hzxuzhonghu (approved):
This is not a api, we didnot promise our package does not change. Actually, we are always changing internally.

PR #34634, opened 2021-08-11
istioctl: fix skip confirmation with dryrun while upgrading istio

zufardhiyaulhaq:

I encounter bug behavior while executing istioctl upgrade with skip confirmation and dry-run in 1.9.7 and master branch.
```
./out/linux_amd64/istioctl upgrade --force --dry-run --skip-confirmation -f /root/infrastructure/community-ops/istio_mesh/cluster.yaml 

.
.
Upgrade version check passed: 1.9.7 -> 1.12.0.
.
.
.

Confirm to proceed [y/N]? ^C
```

`waitForConfirmation()` function should not be depends on `dryRun` args. This PR removing the bad logic when calling `waitForConfirmation()`.

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/operator/cmd/mesh/upgrade.go
+++ b/operator/cmd/mesh/upgrade.go
@@ -215,7 +215,7 @@ func upgrade(rootArgs *rootArgs, args *upgradeArgs, l clog.Logger) (err error) {
 	}
 	checkUpgradeIOPS(currentProfileIOPSYaml, targetIOPYaml, overrideIOPYaml, l)
 
-	waitForConfirmation(args.skipConfirmation, l)
+	waitForConfirmation(args.skipConfirmation || rootArts.dryRun, l)
 
 	// Apply the Istio Control Plane specs reading from inFilenames to the cluster
 	iop, err := InstallManifests(targetIOP, args.force, rootArgs.dryRun, restConfig, client, args.readinessTimeout, l)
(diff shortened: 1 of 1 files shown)


operator/cmd/mesh/upgrade.go
```diff
@@ -215,7 +215,7 @@ func upgrade(rootArgs *rootArgs, args *upgradeArgs, l clog.Logger) (err error) {
 	}
 	checkUpgradeIOPS(currentProfileIOPSYaml, targetIOPYaml, overrideIOPYaml, l)
 
-	waitForConfirmation(args.skipConfirmation && !rootArgs.dryRun, l)
+	waitForConfirmation(args.skipConfirmation, l)
```
howardjohn, line 218:
```suggestion
	waitForConfirmation(args.skipConfirmation || rootArts.dryRun, l)
```

WDYT? I am not sure it ever makes sense to confirm in dryrun, and it seems to match https://github.com/istio/istio/blob/8032be2664d55b55fd013857a50bb60390b32ec3/operator/cmd/mesh/install.go#L170

zufardhiyaulhaq: I think it will make the user confused, what happens if the user wants to execute a dry run with confirmation? I think we also need to refactor the install.go in this case.

so in the end, the user will have 4 options
1. dry-run with confirmation
2. dry-run without confirmation
3. install/upgrade with confirmation
4. install/upgrade without confirmation

howardjohn: I think we should be consistent with `install`. There is no need to confirm a dry run - the point of confirmation is to avoid accidentally making changes to your cluster

zufardhiyaulhaq: got it, commit your suggestion.

zufardhiyaulhaq:
@howardjohn let me know if we can merge this. thanks!

howardjohn (approved):

elfinhe (approved):

Issue #34577, opened 2021-08-07
serviceEntry addresses only the last address is effective,can not find other address config in envoy config


13567436138:
(**NOTE**: This is used to report product bugs:
  To report a security vulnerability, please visit <https://istio.io/about/security-vulnerabilities>
  To ask questions about how to use Istio, please visit <https://discuss.istio.io>)

**Bug description**
serviceEntry addresses only the last address is effective,can not find other address config in envoy config

[ ] Docs
[ ] Installation
[x ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**
192.168.198.177 and 192.168.198.178 both effective
**Steps to reproduce the bug**
```
apiVersion: networking.istio.io/v1beta1
kind: ServiceEntry
metadata:
  name: details-se
spec:
  hosts:
  - details.bookinfo.com
  addresses:
  - 192.168.198.177/32
  - 192.168.198.178/32
  location: MESH_INTERNAL
  ports:
  - number: 9080
    name: http
    protocol: HTTP
  resolution: STATIC
  workloadSelector:
    labels:
      app: details
```

![1628311481(1)](https://user-images.githubusercontent.com/31508490/128588243-0ebe3e59-b618-417a-8e0b-54d58a20b294.jpg)

there is no 192.168.198.177's config

**Version** (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version --short` if you used Helm)
istio 1.10.0
**How was Istio installed?**

**Environment where the bug was observed (cloud vendor, OS, etc)**

Additionally, please consider running `istioctl bug-report` and attach the generated cluster-state tarball to this issue.
Refer [cluster state archive](http://istio.io/help/bugs/#generating-a-cluster-state-archive) for more details.

ChenglongGao327:
It can also be reproduced using the master branch

13567436138:
so is this a problem

ChenglongGao327:
i think so, let me fix it

howardjohn:
Hypothetically yes - the API allows it although it probably shouldn't as it
clearly doesn't work. It will be useful when we support dual stack , but we
don't right now.

ChenglongGao327:
When do we plan to do dual stack

howardjohn:
@jacob-delgado and a few others are working on it

[...]

PR #34635, opened 2021-08-11
For service entry add all spec.addresses to the envoy config

ChenglongGao327:
fix: https://github.com/istio/istio/issues/34577
For service entry resources, add all addresses of service entry spec.addresses to envoy route domains

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/gateway.go
+++ b/pilot/pkg/networking/core/v1alpha3/gateway.go
@@ -230,9 +230,9 @@ func (configgen *ConfigGeneratorImpl) buildGatewayListeners(builder *ListenerBui
 }
 
 func buildNameToServiceMapForHTTPRoutes(node *model.Proxy, push *model.PushContext,
-	virtualService config.Config) map[host.Name]*model.Service {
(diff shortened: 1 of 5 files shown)


howardjohn (commented):
I am not sure this is sufficient. The vhost match is only one of many many places the VIP is used, where it will not support multiple. For example, on TCP we should be generating multiple listeners, for DNS we need to return both of them, etc.

If we are going to do this we should probably add integration tests for these case

ChenglongGao327:
> I am not sure this is sufficient. The vhost match is only one of many many places the VIP is used, where it will not support multiple. For example, on TCP we should be generating multiple listeners, for DNS we need to return both of them, etc.
> 

This PR only fixes the domains in the envoy config route, and does not involve other places, such as listeners, dns

ChenglongGao327:
> If we are going to do this we should probably add integration tests for these case

yes，there are more relevant places

PR #34639, opened 2021-08-11
grpc xds bootstrap: include full node metadata

stevenctl:
The previous code used `Node.RawMetadata`

For everything except the following call, we extract into `untypedMeta`, but here we extract into typed meta.

https://github.com/istio/istio/blob/6da7a3193970e46ebe6cf748f6411f1de1d286ec/pkg/bootstrap/config.go#L492

This means the gRPC bootstrap ends up missing out on  a few critical fields:
* POD_NAME
* POD_NAMESPACE
* SERVICE_ACCOUNT
* ISTIO_METAJSON_LABELS

Without this fix we can't set namespace-level policy like PeerAuthn

### DIFF
--- a/pkg/istio-agent/testdata/grpc-bootstrap.json
+++ b/pkg/istio-agent/testdata/grpc-bootstrap.json
@@ -23,9 +23,9 @@
       "PROXY_CONFIG": {
         "binaryPath": "/Users/landow/go/src/istio.io/istio/out/darwin_amd64/envoy",
         "concurrency": 2,
-        "configPath": "/var/folders/43/dd38c_f97lvdt_p9rgj7hvq400r4rr/T/TestAgent_gRPC_XDS_bootstrap3603799447/001",
+        "configPath": "/something/predictable",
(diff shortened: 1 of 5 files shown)


pkg/istio-agent/testdata/grpc-bootstrap.json
```diff
@@ -14,7 +14,33 @@
   ],
   "node": {
     "id": "sidecar~127.0.0.1~pod1.fake-namespace~fake-namespace.svc.cluster.local",
-    "metadata": {},
+    "metadata": {
+      "INSTANCE_IPS": "127.0.0.1",
+      "PILOT_SAN": [
+        "istiod.istio-system.svc"
+      ],
+      "PROV_CERT": "/Users/landow/go/src/istio.io/istio/tests/testdata/certs/pilot/root-cert.pem",
```
howardjohn, line 22: seems possibly not determinstic?

stevenctl: fixed - just ignoring the fields with these

howardjohn (approved):

PR #34654, opened 2021-08-12
fix proxyConfig overrides

Regevd:
Subject: Fix a bug in ProxyConfig overrides that causes overrides to be applied to the default config.

Description:
+ When using proxy config overrides via pod annotation the overrides are applied to the global default config so all new pod gets the overrides.
+ The bug is caused by a lack of deep copying of the relevant map.
+ The test were updated to check the default config to remain as the old one.

- [X ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/config/mesh/mesh.go
+++ b/pkg/config/mesh/mesh.go
@@ -128,13 +128,16 @@ func ApplyProxyConfig(yaml string, meshConfig meshconfig.MeshConfig) (*meshconfi
 }
 
 func applyProxyConfig(yaml string, proxyConfig *meshconfig.ProxyConfig) (*meshconfig.ProxyConfig, error) {
-	origMetadata := proxyConfig.ProxyMetadata
-	if err := gogoprotomarshal.ApplyYAML(yaml, proxyConfig); err != nil {
+	result, err := deepCopyProxyConfig(proxyConfig)
+	if err != nil {
+		return nil, fmt.Errorf("could not deep copy proxy config: %v", err)
+	}
+	if err := gogoprotomarshal.ApplyYAML(yaml, result); err != nil {
 		return nil, fmt.Errorf("could not parse proxy config: %v", err)
 	}
-	newMetadata := proxyConfig.ProxyMetadata
-	proxyConfig.ProxyMetadata = mergeMap(origMetadata, newMetadata)
-	return proxyConfig, nil
+	origMetadata := proxyConfig.ProxyMetadata
+	result.ProxyMetadata = mergeMap(origMetadata, result.ProxyMetadata)
+	return result, nil
 }
 
 func extractYamlField(key string, mp map[string]interface{}) (string, error) {
@@ -231,13 +234,14 @@ func mergeMap(original map[string]string, merger map[string]string) map[string]s
 	if original == nil && merger == nil {
 		return nil
 	}
-	if original == nil {
-		original = map[string]string{}
+	result := map[string]string{}
+	for k, v := range original {
+		result[k] = v
 	}
 	for k, v := range merger {
-		original[k] = v
+		result[k] = v
 	}
-	return original
+	return result
 }
 
 // ApplyMeshConfigDefaults returns a new MeshConfig decoded from the
@@ -256,6 +260,18 @@ func DeepCopyMeshConfig(mc *meshconfig.MeshConfig) (*meshconfig.MeshConfig, erro
 		return nil, err
 	}
 	return nmc, nil
+}
+
+func deepCopyProxyConfig(mc *meshconfig.ProxyConfig) (*meshconfig.ProxyConfig, error) {
+	j, err := gogoprotomarshal.ToJSON(mc)
+	if err != nil {
+		return nil, err
+	}
+	npc := &meshconfig.ProxyConfig{}
+	if err := gogoprotomarshal.ApplyJSON(j, npc); err != nil {
+		return nil, err
+	}
+	return npc, nil
 }
 
 // EmptyMeshNetworks configuration with no networks
--- a/pkg/config/mesh/mesh.go
+++ b/pkg/config/mesh/mesh.go
@@ -128,16 +128,13 @@ func ApplyProxyConfig(yaml string, meshConfig meshconfig.MeshConfig) (*meshconfi
 }
 
 func applyProxyConfig(yaml string, proxyConfig *meshconfig.ProxyConfig) (*meshconfig.ProxyConfig, error) {
-	result, err := deepCopyProxyConfig(proxyConfig)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


Regevd:
@googlebot I signed it!

pkg/config/mesh/mesh.go
```diff
@@ -231,13 +234,14 @@ func mergeMap(original map[string]string, merger map[string]string) map[string]s
 	if original == nil && merger == nil {
 		return nil
 	}
-	if original == nil {
-		original = map[string]string{}
+	result := map[string]string{}
```
ChenglongGao327, line 237: the new map` result` is not required

ramaraochavali (changes requested):
I think this is working as expected. @howardjohn Can confirm

pkg/config/mesh/mesh.go
Regevd: this function usages expect `original` param to remain the same. this isn't possible without creating a new map for result

Regevd:
@ramaraochavali see https://github.com/istio/istio/issues/34657

ramaraochavali:
Is n't this test validating that https://github.com/istio/istio/blob/dff2dacde9a5eb84e439b9aa10ddeb4d4667a4e8/pkg/config/mesh/mesh_test.go#L61? You can add the additional check that you had without your changes. I think it will still pass.

Regevd:
> Is n't this test validating that
> 
> https://github.com/istio/istio/blob/dff2dacde9a5eb84e439b9aa10ddeb4d4667a4e8/pkg/config/mesh/mesh_test.go#L61
> 
> ? You can add the additional check that you had without your changes. I think it will still pass.

fails

Regevd:
@ramaraochavali  can you ok-it-test it?

ramaraochavali:
```
	t.Run("apply proxy metadata", func(t *testing.T) {
		config := mesh.DefaultMeshConfig()
		config.DefaultConfig.ProxyMetadata = map[string]string{
			"merged":  "original",
			"default": "foo",
		}
		mc, err := mesh.ApplyProxyConfig(`proxyMetadata: {"merged":"override","override":"bar"}`, config)
		if err != nil {
			t.Fatal(err)
		}
		// Ensure we didn't modify the passed in mesh config
		if !reflect.DeepEqual(mc.DefaultConfig.ProxyMetadata, map[string]string{

			"merged":   "override",
			"default":  "foo",
			"override": "bar",
		}) {
			t.Fatalf("unexpected proxy metadata: %+v", mc.DefaultConfig.ProxyMetadata)
		}
		if !reflect.DeepEqual(config.DefaultConfig.ProxyMetadata, map[string]string{
			"merged":  "original",
			"default": "foo",
		}) {
			t.Fatalf("unexpected proxy metadata: %+v", config.DefaultConfig.ProxyMetadata)
		}
	})
	
```
This works for me on master

ramaraochavali (approved):

Regevd:
> ```
> 	t.Run("apply proxy metadata", func(t *testing.T) {
> 		config := mesh.DefaultMeshConfig()
> 		config.DefaultConfig.ProxyMetadata = map[string]string{
> 			"merged":  "original",
> 			"default": "foo",
> 		}
> 		mc, err := mesh.ApplyProxyConfig(`proxyMetadata: {"merged":"override","override":"bar"}`, config)
> 		if err != nil {
> 			t.Fatal(err)
> 		}
> 		// Ensure we didn't modify the passed in mesh config
> 		if !reflect.DeepEqual(mc.DefaultConfig.ProxyMetadata, map[string]string{
> 
> 			"merged":   "override",
> 			"default":  "foo",
> 			"override": "bar",
> 		}) {
> 			t.Fatalf("unexpected proxy metadata: %+v", mc.DefaultConfig.ProxyMetadata)
> 		}
> 		if !reflect.DeepEqual(config.DefaultConfig.ProxyMetadata, map[string]string{
> 			"merged":  "original",
> 			"default": "foo",
> 		}) {
> 			t.Fatalf("unexpected proxy metadata: %+v", config.DefaultConfig.ProxyMetadata)
> 		}
> 	})
> 	
> ```
> 
> This works for me on master which shows default mesh config is not modified. Am I missing some thing?

Your test seems fine but as you can see the test job fails.
I'll bring my changes back to fix this test

pkg/config/mesh/mesh_test.go
```diff
@@ -89,14 +96,22 @@ func TestApplyProxyConfig(t *testing.T) {
...
-			t.Fatalf("unexpected proxy metadata: %+v", mc.DefaultConfig.ProxyMetadata)
+		}
+		if !reflect.DeepEqual(mc.DefaultConfig.ProxyMetadata, expected) {
+			t.Fatalf("unexpected proxy metadata. expected: %+v, actual: %+v", expected, config.DefaultConfig.ProxyMetadata)
+		}
+		// Ensure we didn't modify the default mesh config
+		expected = map[string]string{
+			"merged":  "original",
+			"default": "foo",
+		}
```
ramaraochavali, line 112: Sorry I missed this. This test is calling `ApplyMeshConfig` with changed config - So it will override. Which is correct. I guess no changes needed? This expectation is incorrect.

ramaraochavali (changes requested):
Second test expectation is incorrect.

pkg/config/mesh/mesh_test.go
Regevd: this test calls `ApplyProxyConfig` which the comment above it says 
`// ApplyProxyConfig applies the give proxy config yaml to a mesh config object. The passed in mesh config
// will not be modified.`

ramaraochavali: Is n't this calling  ApplyMeshConfig https://github.com/istio/istio/blob/dff2dacde9a5eb84e439b9aa10ddeb4d4667a4e8/pkg/config/mesh/mesh_test.go#L87?

Regevd: Sorry you're right. Missed it.
And yet   ApplyMeshConfig comment ->`ApplyMeshConfig returns a new MeshConfig decoded from the
 input YAML with the provided defaults applied to omitted configuration values.`
 ApplyMeshConfig suppose to return new instance, not to edit-in-place. that's why there was a check in test to '// Ensure we didn't modify the passed in mesh config'

Regevd: ApplyMeshConfig should base it's output upon the second param, `defaultConfig`, no to modify it!

howardjohn:
I want to take a close look at this but haven't had time. Hoping to review by EOD

PR #34655, opened 2021-08-12
use sets.Set to replace the array

ChenglongGao327:
1. use sets.Set
2. fix a typo

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/security/trustdomain/bundle.go
+++ b/pilot/pkg/security/trustdomain/bundle.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"strings"
 
+	"istio.io/istio/pilot/pkg/util/sets"
 	"istio.io/istio/pkg/config/constants"
 	istiolog "istio.io/pkg/log"
 )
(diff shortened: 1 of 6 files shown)


pilot/pkg/security/trustdomain/bundle.go
```diff
@@ -103,11 +104,11 @@ func (t Bundle) replaceTrustDomains(principal, trustDomainFromPrincipal string)
...
 		// has a * prefix. For example, "*-td" can match with "old-td" and "new-td", but we only want
 		// to keep the principal as-is in the generated config, .i.e. *-td.
-		if !isKeyInList(newPrincipal, principalsForAliases) {
-			principalsForAliases = append(principalsForAliases, newPrincipal)
+		if !principalsForAliases.Contains(newPrincipal) {
+			principalsForAliases.Insert(newPrincipal)
 		}
 	}
-	return principalsForAliases
+	return principalsForAliases.SortedList()
```
ChenglongGao327, line 111: sort newPrincipal array, need adjust the order of test data

howardjohn (changes requested):
I am suspicious that this is faster.. we need to alloc a map (slow) and sort it (slow).

if we are going to micro-optimize we should have benchmarks proving its actually an improvement IMO

pilot/pkg/security/trustdomain/bundle.go
```diff
@@ -84,9 +85,9 @@ func (t Bundle) ReplaceTrustDomainAliases(principals []string) []string {
 // replaceTrustDomains replace the given principal's trust domain with the trust domains from the
 // trustDomains list and return the new principals.
 func (t Bundle) replaceTrustDomains(principal, trustDomainFromPrincipal string) []string {
-	principalsForAliases := []string{}
+	principalsForAliases := sets.Set{}
```
yangminzhu, line 88: In most cases there will be only a few items in the `principalsForAliases` and `t.TrustDomains` list, this likely won't be any faster especially you have to do an extra `SortedList()` in the end.

ChenglongGao327:
yes，if there only a few items this won't be faster

PR #34656, opened 2021-08-12
Impl HasSynced for memory controller and improve HasSynced of ADSC

YonkaFang:

In current istio codebase, memory config controller does not impl the `HasSynced` method which means we will wait for readiness of config source using mem store (such as mcp-over-xds). And this will break the integrity of istiod state control.

This PR impl `HasSynced` for mem controller and use it in the XDS configsource case.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/bootstrap/configcontroller.go
+++ b/pilot/pkg/bootstrap/configcontroller.go
@@ -221,7 +221,9 @@ func (s *Server) initConfigSources(args *PilotArgs) (err error) {
 				return fmt.Errorf("failed to dial XDS %s %v", configSource.Address, err)
 			}
 			store := memory.MakeSkipValidation(collections.Pilot)
-			configController := memory.NewController(store)
+			configController := memory.NewControllerWithOptions(store, &memory.Config{
+				HasSynced: xdsMCP.HasSynced,
+			})
 			xdsMCP.Store = model.MakeIstioStore(configController)
 			err = xdsMCP.Run()
 			if err != nil {
--- a/pilot/pkg/bootstrap/configcontroller.go
+++ b/pilot/pkg/bootstrap/configcontroller.go
@@ -221,9 +221,7 @@ func (s *Server) initConfigSources(args *PilotArgs) (err error) {
 				return fmt.Errorf("failed to dial XDS %s %v", configSource.Address, err)
 			}
 			store := memory.MakeSkipValidation(collections.Pilot)
-			configController := memory.NewControllerWithOptions(store, &memory.Config{
-				HasSynced: xdsMCP.HasSynced,
-			})
+			configController := memory.NewController(store, memory.PatchHasSynced(xdsMCP.HasSynced))
 			xdsMCP.Store = model.MakeIstioStore(configController)
 			err = xdsMCP.Run()
 			if err != nil {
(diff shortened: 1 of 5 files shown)


hzxuzhonghu (commented):
Can you add some tests

YonkaFang:
> Can you add some tests

ok. i'll push them a little later

YonkaFang:
@hzxuzhonghu I've pushed the test cases. Plz help take another look.

pilot/pkg/bootstrap/configcontroller.go
```diff
@@ -221,7 +221,9 @@ func (s *Server) initConfigSources(args *PilotArgs) (err error) {
 				return fmt.Errorf("failed to dial XDS %s %v", configSource.Address, err)
 			}
 			store := memory.MakeSkipValidation(collections.Pilot)
-			configController := memory.NewController(store)
+			configController := memory.NewControllerWithOptions(store, &memory.Config{
+				HasSynced: xdsMCP.HasSynced,
```
hzxuzhonghu, line 225: why not directly pass `func HasSynced() bool `

pilot/pkg/config/memory/controller.go
```diff
@@ -26,9 +26,35 @@ import (
 	"istio.io/istio/pkg/config/schema/collection"
 )
 
+type Config struct {
+	HasSynced func() bool
+	SyncProc  bool
```
hzxuzhonghu, line 31: what's this?

YonkaFang: It keeps the ability to new a controller with sync monitor. U can see the code below.

pilot/pkg/config/memory/controller.go
```diff
@@ -26,9 +26,35 @@ import (
...
+	if opts == nil {
+		opts = &Config{}
+	}
+
+	out := &controller{
+		configStore: cs,
+		opts:        opts,
+	}
+
+	if opts.SyncProc {
```
hzxuzhonghu, line 51: It looks no better than before

hzxuzhonghu (commented):
Basically lg, but need some improve

pilot/pkg/bootstrap/configcontroller.go
YonkaFang: I consider the fact that there're several `new` functions already. Then I add a general one which can do the complex init things.

pilot/pkg/config/memory/controller.go
hzxuzhonghu: Yes, understood. Looks below is more readable

func NewController(cs model.ConfigStore, hasSynced func() bool) model.ConfigStoreCache {

pilot/pkg/config/memory/controller.go
YonkaFang: just keep a compability with existing usage.

pilot/pkg/config/memory/controller.go
YonkaFang: I think the `hasSynced` func is not a common args. It's more like a customization. So I didn't add it to the **common** `NewController` func.
BTW, there've been tons of usage of `NewController`.

YonkaFang:
@hzxuzhonghu Plz help review the updated code. Thanks.

hzxuzhonghu (approved):
Nit: The patch function is still very complex, i will clean up afterwards.

PR #34662, opened 2021-08-12
grpc xds: support calling service using shortnames

stevenctl:
When a proxyless client dials `xds:///foo:123`, it expects a listener named `foo:123` exactly. 

Currently, we look at egress listener services and generate a listener if the filtered name and port matches the FQDN of the service. 

To support making requests using shortnames (foo, foo.ns, foo.ns.svc) we generate the list of AltVirtualHost domains. If the requested listener matches one of these, we build the listener with that exact name. 

Since it's possible for a gRPC app to have watches against both `foo` and `foo.ns`, we must duplicate the listener with each of the requested names.


stevenctl:
It's possible this could be improved by instead taking the input names, and converting them to the FQDN, rather than generating alt hosts for every service. 

### Example:

Input: 
  foo:80, foo.ns:81
Filter: 
```json
{
  "foo.ns.svc.cluster.local": {
    "foo": {
      "80": {}
    },
    "foo.ns": {
      "81": {}
    }
  }
}
```

This lets us do the initial match on the FQDN, then the port-level matches based on the exact name.

costinm (approved):
Changes are specific to grpcgen package, so safe.

PR #34664, opened 2021-08-12
support istio skip ugrade testing

Aakash2017:
This PR is part of future work to support istio skip-level upgrade testing from versions below 1.8 (which do not re-inject a pod if istio-proxy is already present).

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [X] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/test/framework/components/echo/kube/deployment.go
+++ b/pkg/test/framework/components/echo/kube/deployment.go
@@ -148,6 +148,7 @@ spec:
 {{- if and
   (ne ($subset.Annotations.GetByName "sidecar.istio.io/inject") "false")
   (ne ($subset.Annotations.GetByName "inject.istio.io/templates") "grpc")
+  ($.OverlayIstioProxy)
 }}
       - name: istio-proxy
         image: auto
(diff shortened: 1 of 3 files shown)


pkg/test/framework/components/echo/kube/testdata/multiple-istio-versions.yaml
```diff
@@ -65,7 +65,7 @@ spec:
           - --version
           - "bar"
           - --istio-version
-          - "1.8.2"
```
nmittler, line 68: did you need the changes in this file?

Aakash2017: If these versions were left to be '1.8.2', the deployment yaml would not include the istio-proxy overlay. This change ensures that we are fully testing the added conditional statement in the template.

nmittler: ah got it.

howardjohn (approved):

PR #34673, opened 2021-08-13
Add resource quota support to CNI DaemonSet.

### LINKED ISSUE #34371: Add default system-node-critical ResourceQuota to the namespace istio-cni installed
Istio CNI runs with `system-node-critical` which prevents it from being evicted when node is under resource pressure. However most cloud providers only set node critical `ResourceQuota` in `kube-system` and istio installation does not create `ResourceQuota` by default in CNI namespace, which prevents istio-cni pods being scheduled outside kube-system out of the box. We should consider adding the following `ResourceQuota` when CNI is enabled.

```
apiVersion: v1
kind: ResourceQuota
metadata:
  name: istio-critical-pods
  namespace: <cni installed-namespace>
spec:
  hard:
    pods: 1G
  scopeSelector:
    matchExpressions:
    - operator: In
      scopeName: PriorityClass
      values:
      - system-node-critical
```

cc @mandarjog @howardjohn

bianpengyuan:
fix https://github.com/istio/istio/issues/34371


manifests/charts/istio-cni/templates/resourcequota.yaml
```diff
@@ -0,0 +1,16 @@
...
+  namespace: {{ .Release.Namespace }}
+spec:
+  hard:
+    pods: {{ .Values.cni.resourceQuotas.pods | quote }}
+  scopeSelector:
+    matchExpressions:
+    - operator: In
+      scopeName: PriorityClass
+      values:
+      - system-node-critical
```
howardjohn, line 15: should be based on values.priorityClassName?

howardjohn (commented):
looks good

manifests/charts/istio-cni/templates/resourcequota.yaml
bianpengyuan: This priority class targets only cni, and We hard coded CNI priority class to be `system-node-critical`: https://github.com/istio/istio/blob/bb280f568f4f5369f3732130cd1bd8481b5fdd3c/manifests/charts/istio-cni/templates/daemonset.yaml#L55

costinm (approved):

Issue #34679, opened 2021-08-13
Improve Auto-detect HTTP Headless Services


howardjohn:
Route config for auto headless service:
```
      "name": "alertmanager-operated.default.svc.cluster.local:9093",
      "virtual_hosts": [
       {
        "name": "alertmanager-operated.default.svc.cluster.local:9093",
        "domains": [
         "alertmanager-operated.default.svc.cluster.local",
         "alertmanager-operated.default.svc.cluster.local:9093",
         "alertmanager-operated",
         "alertmanager-operated:9093",
         "alertmanager-operated.default.svc",
         "alertmanager-operated.default.svc:9093",
         "alertmanager-operated.default",
         "alertmanager-operated.default:9093",
         "*.alertmanager-operated.default.svc.cluster.local",
         "*.alertmanager-operated.default.svc.cluster.local:9093",
         "*.alertmanager-operated",
         "*.alertmanager-operated:9093",
         "*.alertmanager-operated.default.svc",
         "*.alertmanager-operated.default.svc:9093",
         "*.alertmanager-operated.default",
         "*.alertmanager-operated.default:9093"
        ],
```

Route config for auto non-headless:
```
       {
        "name": "echo.default.svc.cluster.local:80",
        "domains": [
         "echo.default.svc.cluster.local",
         "echo.default.svc.cluster.local:80",
         "echo",
         "echo:80",
         "echo.default.svc",
         "echo.default.svc:80",
         "echo.default",
         "echo.default:80",
         "10.96.75.204",
         "10.96.75.204:80"
        ],
```

This means that with non-headless, I can do `curl IP`. With headless, I cannot.

In the real world, this breaks alertmanager, which is an extremely common deployment.

Because the route is already scoped to the Pod IP (for headless) and cluster IP (for non-headless), I think in all cases of auto-HTTP we should have domains be `*` only.

howardjohn:
BUG: For `http` headless, we generate routes like:
```
       {
        "name": "headless.echo.svc.cluster.local:18080",
        "domains": [
         "headless.echo.svc.cluster.local",
         "headless.echo.svc.cluster.local:18080",
         "headless",
         "headless:18080",
         "headless.echo.svc",
         "headless.echo.svc:18080",
         "headless.echo",
         "headless.echo:18080",
         "*.headless.echo.svc.cluster.local",
         "*.headless.echo.svc.cluster.local:18080",
         "*.headless",
         "*.headless:18080",
         "*.headless.echo.svc",
[...]

PR #34680, opened 2021-08-13
Do not check HTTP Hostname for single-VIP listeners (auto protocol) and add tests

howardjohn:
Fixes https://github.com/istio/istio/issues/34679

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/httproute.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute.go
@@ -463,7 +463,14 @@ func getVirtualHostsForSniffedServicePort(vhosts []*route.VirtualHost, routeName
 	}
 
 	if len(virtualHosts) == 0 {
-		virtualHosts = vhosts
+		return virtualHosts
+	}
+	if len(virtualHosts) == 1 {
+		virtualHosts[0].Domains = []string{"*"}
+		return virtualHosts
+	}
+	if features.EnableUnsafeAssertions {
+		panic(fmt.Sprintf("unexpectedly matched multiple virtual hosts for %v: %v", routeName, virtualHosts))
 	}
 	return virtualHosts
 }
--- a/pilot/pkg/networking/core/v1alpha3/httproute_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute_test.go
@@ -1082,7 +1082,7 @@ func TestSidecarOutboundHTTPRouteConfig(t *testing.T) {
 			virtualServiceConfigs: nil,
 			expectedHosts: map[string]map[string]bool{
 				"test-private.com:70": {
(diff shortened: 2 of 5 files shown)


howardjohn:
@ramaraochavali @hzxuzhonghu WYDT? I need to clean up the PR a bit but seems to be working

ramaraochavali:
Makes sense to me.

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -463,7 +463,13 @@ func getVirtualHostsForSniffedServicePort(vhosts []*route.VirtualHost, routeName
 	}
 
 	if len(virtualHosts) == 0 {
-		virtualHosts = vhosts
+		return virtualHosts
+	} else if len(virtualHosts) == 1 {
+		virtualHosts[0].Domains = []string{"*"}
```
hzxuzhonghu, line 468: I understand it happens when routename is in vhost.domains. And this could happen on the condition

```
if listenerProtocol == istionetworking.ListenerProtocolAuto &&
			sniffingEnabled && listenerOpts.bind != actualWildcard && listenerOpts.service != nil {
			rdsName = string(listenerOpts.service.ClusterLocal.Hostname) + ":" + strconv.Itoa(listenerOpts.port.Port)
		}
```

But i could not understand how this could fix the original issue

howardjohn: Basically the original issue the route is bound to a listener on <vip>. Their request matches that vip, but the host header doesn't match any domains.
This makes the route match ALL host headers so it doesn't matter that they didn't have  a match

hzxuzhonghu: You mean the original issue is the svc port is `auto-xxx` and for headless with it , we build a listener for each instance with name `podip:port`.

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -463,7 +463,13 @@ func getVirtualHostsForSniffedServicePort(vhosts []*route.VirtualHost, routeName
...
+	} else {
+		if features.EnableUnsafeAssertions {
+			panic(fmt.Sprintf("unexpectedly matched multiple virtual hosts for %v: %v", routeName, virtualHosts))
```
hzxuzhonghu, line 471: nit: from the above this could never happen

howardjohn: Why not? The conditions are `0`, `1`, and `>1`. It _shouldn't_ happen but due to a bug it could, which is why there is the assertion

hzxuzhonghu (approved):

jacob-delgado (approved):

PR #34681, opened 2021-08-13
Test framework fixes to support pilot tests in external istiod topology

frankbu:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/test/framework/components/istio/operator.go
+++ b/pkg/test/framework/components/istio/operator.go
@@ -370,16 +370,21 @@ func deploy(ctx resource.Context, env *kube.Environment, cfg Config) (Instance, 
 		return nil, fmt.Errorf("%d errors occurred deploying remote clusters: %v", errs.Len(), errs.ErrorOrNil())
 	}
 
-	// Configure discovery and east-west gateways for remote clusters.
+	// Configure discovery and gateways for remote clusters.
 	for _, c := range ctx.Clusters().Kube().Remotes() {
 		c := c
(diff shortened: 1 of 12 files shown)


pkg/test/framework/components/istio/operator.go
```diff
@@ -370,16 +370,21 @@ func deploy(ctx resource.Context, env *kube.Environment, cfg Config) (Instance,
...
+	// Configure discovery and gateways for remote clusters.
 	for _, c := range ctx.Clusters().Kube().Remotes() {
 		c := c
 		if i.isExternalControlPlane() || cfg.IstiodlessRemotes {
 			if err = configureRemoteClusterDiscovery(i, cfg, c); err != nil {
 				return i, err
 			}
+
+			// Install ingress and egress gateways
+			if err = installRemoteClusterGateways(i, c); err != nil {
```
stevenctl, line 382: Shouldn't we be able to do this with the same install manifest we already have?

stevenctl:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/34681/integ-pilot-istiodremote-tests_istio/1427343006570123264

stevenctl (approved):
Seems like this gets us past any initial setup failures, but definitely need some work. I think test writers calling `Primary` now need to take special care to make sure they want that and not `Config`. Could be tricky to communicate.  Hope we can get this job running as blocking sooner than later so new tests aren't added that break this.

PR #34685, opened 2021-08-13
Use safe timeout and proxy for spiffe bundle

howardjohn:


shankgan (approved):
LFTM as long as it passes e2e

Issue #34609, opened 2021-08-10
HTTPS redirect doesn't work and 404 is returned


seed-248:
**Bug description**
HTTPS redirects configured in the Gateway resource don't work anymore if a virtual service that only matches HTTPS traffic is configured. Further explanations can be found in the example comments.

```
apiVersion: networking.istio.io/v1alpha3
kind: Gateway
metadata:
  name: my-gateway
  namespace: my-namespace
spec:
  selector:
    istio: ingressgateway
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - "example.com"
    tls:
      httpsRedirect: true
  - port:
      number: 443
      name: https
      protocol: HTTPS
    hosts:
    - "example.com"
    tls:
      # TLS termination at the gateway
      mode: SIMPLE
      credentialName: my-cert
      minProtocolVersion: TLSV1_2

---

apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: my-service
  namespace: my-namespace
spec:
  gateways:
  - my-gateway
  hosts:
  - "example.com"
  http:
  - name: example
    # This matching rule causes HTTP requests to 'example.com'. To return 404 (Not Found) responses instead of 301 (Moved Permanently).
    # If the matching rule is commented out, the HTTPS redirects work as expected. HTTPS requests work in both cases.
    match:
    - port: 443
    route:
    - destination:
        host: my-kubernetes-service
        port:
          number: 443
```
 

[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

HTTPS redirects shouldn't be influenced by VirtualServices. 

**Steps to reproduce the bug**

1. Install istio with the default configuration
2. Create a Gateway and a VirtualService resource similar to the ones described above

**Version** (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version --short` if you used Helm)
1.9.7

**How was Istio installed?**
istioctl operator init + IstioOperator resource

**Environment where the bug was observed (cloud vendor, OS, etc)**
AKS

howardjohn:
I can reproduce this on master

anyname2:
I've hit this same issue with 1.10.1.
I added this to my http routes as a workaround:
```yaml
...
  http:
  - match:
[...]

Issue #30501, opened 2021-01-29
How to Redirect HTTP to HTTPS with specificed domain name?


hillbun:
https://istio.io/latest/docs/reference/config/networking/gateway/

```
apiVersion: networking.istio.io/v1alpha3
kind: Gateway
metadata:
  name: my-gateway
  namespace: some-config-namespace
spec:
  selector:
    app: my-gateway-controller
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - uk.bookinfo.com
    - eu.bookinfo.com
    tls:
      httpsRedirect: true
```

if I only wanto to redirect eu.bookinfo.com to HTTPS, how to setup?

howardjohn:
```
apiVersion: networking.istio.io/v1alpha3
kind: Gateway
metadata:
  name: my-gateway
  namespace: some-config-namespace
spec:
  selector:
    app: my-gateway-controller
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - uk.bookinfo.com
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - eu.bookinfo.com
    tls:
      httpsRedirect: true
```

zengyuxing007:
@howardjohn 

If I want to redirect all exmaple.com subdomains except test.example.com, it seems that 1.8, 1.9 and the current version are not supported, I have tested it in 1.9.7 and confirmed that it is not supported.


```yaml
apiVersion: networking.istio.io/v1beta1
kind: Gateway
metadata:
  name: my-gateway
  namespace: default
spec:
  selector:
    app: istio-ingressgateway
  servers:
   - hosts:
     - '*.example.com'
     name: others
     port:
       name: http
       number: 80
       protocol: HTTP
     tls:
       httpsRedirect: true
   - hosts:
     - "test.example.com"
     name: test
     port:
       name: http1
       number: 80
       protocol: HTTP
     tls:
       httpsRedirect: false
   - hosts:
     - '*.example.com'
     name: httpbin-https
     port:
       name: https
       number: 443
       protocol: HTTPS
     tls:
      mode: SIMPLE
      credentialName: httpbin-credential
```

osswangxining:
@howardjohn  I have tested using version 1.8.6 and found that it didn't work for above sample.  But it can work on version 1.7.5 and 1.6.8. Any breaking changes for this feature?

osswangxining:
[...]

PR #34689, opened 2021-08-14
301 redirect in VirtualHost when server HttpsRedirect true 

zufardhiyaulhaq:
fix https://github.com/istio/istio/issues/34609

Signed-off-by: Zufar Dhiyaulhaq <zufardhiyaulhaq@gmail.com>

previously, we only create custom virtual hosts with 301 redirections when there is **no virtual service available** for a specific server in the Gateway object with HttpsRedirect set to true. 

**Gateway server** & **virtual service** will create a **virtual host**. **virtual host** for HTTP might be empty if we create something like this:
```
kind: Gateway
metadata:
  name: my-gateway
  namespace: my-namespace
spec:
  selector:
    istio: ingressgateway
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - "example.com"
  - port:
      number: 443
      name: https
      protocol: HTTPS
    hosts:
    - "example.com"
kind: VirtualService
metadata:
  name: my-service
  namespace: my-namespace
spec:
  gateways:
  - my-gateway
  hosts:
  - "example.com"
  http:
  - name: example
    match:
    - port: 443
    route:
    - destination:
        host: my-kubernetes-service
        port:
          number: 443
```

but when the HTTP server enabled TLS redirection, this PR will create dummy **virtual host** for HTTP also.
```
kind: Gateway
metadata:
  name: my-gateway
  namespace: my-namespace
spec:
  selector:
    istio: ingressgateway
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    tls:
      httpsRedirect: true
    hosts:
    - "example.com"
  - port:
      number: 443
      name: https
      protocol: HTTPS
    hosts:
    - "example.com"
 ```

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/gateway.go
+++ b/pilot/pkg/networking/core/v1alpha3/gateway.go
@@ -310,30 +310,7 @@ func (configgen *ConfigGeneratorImpl) buildGatewayHTTPRouteConfig(node *model.Pr
 			virtualServices = push.VirtualServicesForGateway(node, gatewayName)
 			gatewayVirtualServices[gatewayName] = virtualServices
 		}
-		if len(virtualServices) == 0 && server.Tls != nil && server.Tls.HttpsRedirect {
-			// this is a plaintext server with TLS redirect enabled. There is no point in processing the
-			// virtual services for this server because all traffic is anyway going to get redirected
-			// to https. short circuit the route computation by generating a virtual host with no routes
-			// but with tls redirect enabled
-			for _, hostname := range server.Hosts {
-				if vHost, exists := vHostDedupMap[host.Name(hostname)]; exists {
-					if server.Tls != nil && server.Tls.HttpsRedirect {
-						vHost.RequireTls = route.VirtualHost_ALL
-					}
-				} else {
-					newVHost := &route.VirtualHost{
-						Name:                       domainName(hostname, port),
-						Domains:                    buildGatewayVirtualHostDomains(hostname, port),
-						IncludeRequestAttemptCount: true,
-					}
-					if server.Tls != nil && server.Tls.HttpsRedirect {
-						newVHost.RequireTls = route.VirtualHost_ALL
-					}
-					vHostDedupMap[host.Name(hostname)] = newVHost
-				}
-			}
-			continue
-		}
+
 		for _, virtualService := range virtualServices {
 			virtualServiceHosts := host.NewNames(virtualService.Spec.(*networking.VirtualService).Hosts)
 			serverHosts := host.NamesForNamespace(server.Hosts, virtualService.Namespace)
@@ -393,7 +370,10 @@ func (configgen *ConfigGeneratorImpl) buildGatewayHTTPRouteConfig(node *model.Pr
 		// check all hostname in vHostDedupMap and if is not exist with HttpsRedirect set to true
 		// create VirtualHost to redirect
 		for _, hostname := range server.Hosts {
-			if _, exists := vHostDedupMap[host.Name(hostname)]; exists {
+			if vHost, exists := vHostDedupMap[host.Name(hostname)]; exists {
+				if server.Tls != nil && server.Tls.HttpsRedirect {
+					vHost.RequireTls = route.VirtualHost_ALL
+				}
 				continue
 			} else if server.Tls != nil && server.Tls.HttpsRedirect {
 				newVHost := &route.VirtualHost{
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pilot/pkg/networking/core/v1alpha3/gateway.go [hunk 1]
```diff
@@ -389,6 +389,22 @@ func (configgen *ConfigGeneratorImpl) buildGatewayHTTPRouteConfig(node *model.Pr
...
+			if _, exists := vHostDedupMap[host.Name(hostname)]; exists {
+				continue
+			} else if server.Tls != nil && server.Tls.HttpsRedirect {
+				newVHost := &route.VirtualHost{
+					Name:                       domainName(hostname, port),
+					Domains:                    buildGatewayVirtualHostDomains(hostname, port),
+					IncludeRequestAttemptCount: true,
+					RequireTls:                 route.VirtualHost_ALL,
+				}
+				vHostDedupMap[host.Name(hostname)] = newVHost
```
hzxuzhonghu, line 405: It seems these are copied from L318-L333, so i guess we could remove the above `if len(virtualServices) == 0 && server.Tls != nil && server.Tls.HttpsRedirect {` branch

zufardhiyaulhaq: yeah make sense, removing the duplicate code

howardjohn (commented):
This seems like it will re-introduce the regression fixed by https://github.com/istio/istio/pull/29895

howardjohn (changes requested):
I don't think this is safe, see comment above.

zufardhiyaulhaq:
> I don't think this is safe, see comment above.

I cannot see your comment.

howardjohn:
> The other case is that a user can have an HTTP server but still "bypass" the httpsRedirect with X-Forward-Proto and num_trusted_hops, so removing the routes breaks this.

If you remove all of the vhosts then this shouldn't work I would think?

zufardhiyaulhaq:
Hi @howardjohn sorry, forgot to change the description, already updated it to not making confusion.

howardjohn (commented):
Can you add an integration test? We already have a similar one, we can expand it for this case:
```
tests/integration/pilot/common/routing.go
904-                    opts: echo.CallOptions{
905-                            Count: 1,
906-                            Port: &echo.Port{
907-                                    Protocol: protocol.HTTP,
908-                            },
909:                            Validator: echo.ExpectCode("301"),
```

zufardhiyaulhaq:
@howardjohn create two integration tests, let me know what do you think and what do you expect.

howardjohn:
FWIW https://github.com/istio/istio/issues/30501#issuecomment-903925200 appears to be not fixed by this PR (but it also was broken before the PR). Maybe we can get that one after this lands. Thanks for the PR, will give it another review shortly

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -389,6 +366,25 @@ func (configgen *ConfigGeneratorImpl) buildGatewayHTTPRouteConfig(node *model.Pr
...
+			}
+		}
```
howardjohn, line 387:
```suggestion
		for _, hostname := range server.Hosts {
			if !server.GetTls().GetHttpsRedirect() {
				continue
			}
			if vHost, exists := vHostDedupMap[host.Name(hostname)]; exists {
				vHost.RequireTls = route.VirtualHost_ALL
				continue
			}
			newVHost := &route.VirtualHost{
				Name:                       domainName(hostname, port),
				Domains:                    buildGatewayVirtualHostDomains(hostname, port),
				IncludeRequestAttemptCount: true,
				RequireTls:                 route.VirtualHost_ALL,
			}
			vHostDedupMap[host.Name(hostname)] = newVHost
		}
	}
```

Feels a little cleaner than duplicating the check multiple times

howardjohn: Not required though, just personal preference

howardjohn (approved):

pilot/pkg/networking/core/v1alpha3/gateway.go
zufardhiyaulhaq: yeah seems cleaner, let me try to commit the changes

PR #34690, opened 2021-08-14
test framework support connection via proxy

jtrbs:
This PR is to extend the test framework to support connection via proxy. This is for testing with clusters which requires http proxy to connect to the kubernetes API server or connect to the ingress gateway.  

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [X] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/test/framework/components/cluster/cluster.go
+++ b/pkg/test/framework/components/cluster/cluster.go
@@ -209,6 +209,6 @@ type Cluster interface {
 	// ConfigName returns the name of the config cluster for this cluster.
 	ConfigName() string
 
-	// Proxy returns the http proxy config to connect to the cluster
-	Proxy() string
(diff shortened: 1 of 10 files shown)


pkg/test/framework/components/cluster/cluster.go
```diff
@@ -208,4 +208,7 @@ type Cluster interface {
 
 	// ConfigName returns the name of the config cluster for this cluster.
 	ConfigName() string
+
+	// Proxy returns the http proxy config to connect to the cluster
+	Proxy() string
```
nmittler, line 213: nit: Should it be called `HTTPProxy`, just to be more clear?

nmittler (approved):

PR #34691, opened 2021-08-14
Revert "fix portforwarder Close time"

gy95:
Reverts istio/istio#34034
not sure  whether istio/istio#34034 is correct or not, 
not sure when `func (f *forwarder) Close()` should be called? only when PortFowarder  `Start()` function is called successfully, or 
just when it being created successfully
https://github.com/istio/istio/blob/0ba661aea9d7e264b7b5f1670d2d3bbfe4fd569d/pkg/kube/portforwarder.go#L74-L78


hzxuzhonghu (approved):
After a close look, the port forwarder only start forwarding after `Start`

PR #34694, opened 2021-08-14
remove proxy dependency in cluster builder

ramaraochavali:
Removes proxy dependency in cluster builder

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -683,16 +683,16 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 // Services returns the list of services that are visible to a Proxy in a given config namespace
 func (ps *PushContext) Services(proxy *Proxy) []*Service {
 	if proxy != nil {
-		return ps.ServicesBySidecar(proxy.SidecarScope, proxy.Type == SidecarProxy, proxy.ConfigNamespace)
+		return ps.ServicesBySidecar(proxy.SidecarScope, proxy.ConfigNamespace)
 	}
-	return ps.ServicesBySidecar(nil, true, "")
+	return ps.ServicesBySidecar(nil, "")
 }
 
 // Services returns the list of services that are visible to a Proxy in a given config namespace
-func (ps *PushContext) ServicesBySidecar(sidecarScope *SidecarScope, sidecar bool, configNamespace string) []*Service {
+func (ps *PushContext) ServicesBySidecar(sidecarScope *SidecarScope, configNamespace string) []*Service {
 	// If proxy has a sidecar scope that is user supplied, then get the services from the sidecar scope
 	// sidecarScope.config is nil if there is no sidecar scope for the namespace
-	if sidecarScope != nil && sidecar {
+	if sidecarScope != nil {
 		return sidecarScope.services
 	}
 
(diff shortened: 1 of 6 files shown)


adiprerepa (commented):
why are we doing this? what was wrong with the proxy dependency?

ramaraochavali:
#34604 - Refer to the conversation in the PR. The short answer is having a dependency on proxy is too wide and it is difficult to confidently build cache key.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -532,7 +561,9 @@ func (cb *ClusterBuilder) buildLocalityLbEndpoints(proxyNetworkView map[network.
 			continue
 		}
 		// TODO(nmittler): Consider merging discoverability policy with cluster-local
-		if !instance.Endpoint.IsDiscoverableFromProxy(cb.proxy) {
+		// TODO(ramaraochavali): Find a better way here so that we do not have build proxy.
+		// Currently it works because we only determine discoverability only by cluster.
+		if !instance.Endpoint.IsDiscoverableFromProxy(&model.Proxy{Metadata: &model.NodeMetadata{ClusterID: istio_cluster.ID(cb.clusterID)}}) {
```
ramaraochavali, line 566: @nmittler Is proxy mandatory for DiscoverableFromProxy? Today it is always based on cluster (or network?) ? Can we change this to string parameter instead of proxy or add a new struct encapsulating what we need instead of proxy for that function parameter?

hzxuzhonghu (commented):
I donot think this is a good idea to change a lot of other not correlated funtions

ramaraochavali:
> I donot think this is a good idea to change a lot of other not correlated funtions

Can you point me to which functions? I think I have changed relevant functions only which are needed to remove the proxy dependency.

pilot/pkg/model/push_context.go
```diff
@@ -682,22 +682,30 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
...
 // Services returns the list of services that are visible to a Proxy in a given config namespace
 func (ps *PushContext) Services(proxy *Proxy) []*Service {
+	if proxy != nil {
+		return ps.ServicesBySidecar(proxy.SidecarScope, proxy.ConfigNamespace)
+	}
+	return ps.ServicesBySidecar(nil, "")
+}
+
+// Services returns the list of services that are visible to a Proxy in a given config namespace
+func (ps *PushContext) ServicesBySidecar(sidecarScope *SidecarScope, configNamespace string) []*Service {
```
hzxuzhonghu, line 692: I mean this kind of method could be called directly, which is confusing

pilot/pkg/model/push_context.go
```diff
@@ -809,16 +817,21 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Colle
 
 // DestinationRule returns a destination rule for a service name in a given domain.
 func (ps *PushContext) DestinationRule(proxy *Proxy, service *Service) *config.Config {
+	return ps.DestinationRuleBySidecar(service, proxy.SidecarScope, proxy.ConfigNamespace)
+}
+
+// DestinationRuleBySidecar returns a destination rule for a service name in a given domain with passed in sidecar.
+func (ps *PushContext) DestinationRuleBySidecar(service *Service, sidecarScope *SidecarScope, configNamespace string) *config.Config {
```
hzxuzhonghu, line 824: ditto

pilot/pkg/model/push_context.go
ramaraochavali: Do you mean by contructing proxy object before calling?

ramaraochavali: Ok. Found a way to revert these changes

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -99,19 +100,41 @@ type metadataCerts struct {
...
+	proxyType         model.NodeType
+	sidecarScope      *model.SidecarScope
+	passThroughBindIP string
+	supportsIPv4      bool
+	supportsIPv6      bool
+	locality          *core.Locality
+	networkView       map[network.ID]bool
+	proxyIPAddresses  []string
+	configNamespace   string
+	req               *model.PushRequest
```
hzxuzhonghu, line 117: Add some comments

hzxuzhonghu (commented):
LGTM

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
ramaraochavali: added

adiprerepa (approved):
lgtm -- my main concern was addressed by @hzxuzhonghu's comment

PR #34701, opened 2021-08-15
add resource names in debug cache

ramaraochavali:
Since we are moving hash as cache keys we need a better way to represent them in debug. This PR displays typeUrl and resource names in debug mode.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/debug.go
+++ b/pilot/pkg/xds/debug.go
@@ -318,12 +318,13 @@ func (s *DiscoveryServer) cachez(w http.ResponseWriter, req *http.Request) {
 		snapshot := s.Cache.Snapshot()
 		res := make(map[string]string, len(snapshot))
 		totalSize := 0
-		for k, v := range snapshot {
-			if v == nil {
+		for _, resource := range snapshot {
+			if resource == nil {
 				continue
 			}
-			sz := len(v.Resource.GetValue())
-			res[k] = util.ByteCount(sz)
+			resourceType := resource.Resource.TypeUrl
+			sz := len(resource.Resource.GetValue())
+			res[resourceType] += util.ByteCount(sz)
 			totalSize += sz
 		}
 		res["total"] = util.ByteCount(totalSize)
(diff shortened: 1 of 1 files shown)


pilot/pkg/xds/debug.go
```diff
@@ -330,9 +330,13 @@ func (s *DiscoveryServer) cachez(w http.ResponseWriter, req *http.Request) {
 		writeJSON(w, res)
 		return
 	}
-	keys := s.Cache.Keys()
-	sort.Strings(keys)
-	writeJSON(w, keys)
+	snapshot := s.Cache.Snapshot()
+	resources := make(map[string][]string) // Key is typeUrl and value is resource names.
```
howardjohn, line 334: we may want similar logic for the "sizes" logic. IMO it's important there as well so we can see which type is using up space

pilot/pkg/xds/debug.go
```diff
@@ -330,9 +330,13 @@ func (s *DiscoveryServer) cachez(w http.ResponseWriter, req *http.Request) {
...
+	for _, resource := range snapshot {
+		resourceType := resource.Resource.TypeUrl
+		resources[resourceType] = append(resources[resourceType], resource.Name)
```
howardjohn, line 337: maybe `name/hash`? Seems the hash could be useful in some scenarios.. although not as useful as unobfuscated

hzxuzhonghu (commented):
Good for debugging, but as 8080 is exposed publicly, i think we should prevent calling from non-local

howardjohn:
It's not exposed publicly anymore and we already have far more sensitive
data (configz)

hzxuzhonghu:
Good to know.

ramaraochavali:
@hzxuzhonghu @howardjohn changed PTAL.

howardjohn (approved):

Issue #34705, opened 2021-08-15
istioctl 1.11.0 build issue


chenrui333:
**Bug description**

istioctl 1.11.0 build issue when running `make istioctl istioctl.completion`

And error is below:

```
==> make istioctl istioctl.completion
make: *** No rule to make target `/private/tmp/istioctl-20210813-16296-1l75wdb/out/darwin_amd64/istioctl', needed by `/private/tmp/istioctl-20210813-16296-1l75wdb/out/darwin_amd64/release/istioctl.bash'.  Stop.
make: *** Waiting for unfinished jobs....
ISTIO_OUT=/private/tmp/istioctl-20210813-16296-1l75wdb/out/darwin_amd64 ISTIO_BIN=/Users/brew/Library/Caches/Homebrew/go_mod_cache/bin GOOS_LOCAL=darwin bin/retry.sh SSL_ERROR_SYSCALL bin/init.sh
Skipping envoy debug. Set DEBUG_IMAGE to download.
```

[ ] Docs
[x] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

- macOS build passed

**Steps to reproduce the bug**

- git checkout the `1.11.0` tag
- setup couple env vars (like `TARGET_ARCH`)
- run `make istioctl istioctl.completion`

**Version** (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version --short` if you used Helm)

**How was Istio installed?**

**Environment where the bug was observed (cloud vendor, OS, etc)**

macOS

**Additional details**

This is the regression upgrade build from 1.10.3 to 1.11.0

relates to https://github.com/Homebrew/homebrew-core/pull/83195

howardjohn:
cc @ericvn since I thought you had some PRs here recently

ericvn:
I have not had any issues with doing the build on my Mac on both `master` and `release-1.11`, and verified by checking out the 1.11.0 label as well. Looking at the brew script, I notice `ENV["BUILD_WITH_CONTAINER"] = "0"`, which is causing the build to be done differently than normal, so tried adding that to my environment. Build still works.

[...]

PR #34720, opened 2021-08-16
Add additional targets for use outside container with multiple threads.

ericvn:

Fixes https://github.com/istio/istio/issues/34705

https://github.com/istio/istio/issues/34705 notes an issue with brew creating the istioctl and completion files on the new 11.0 thread.  See the issue for more discussion, but the problem is that the hombrew make command uses a `-j6` parameter to use multiple threads and does not build in the build container. Because multiple threads are being used, the dependency for the targets are not yet completed and we get an error about a missing target. This PR adds the targets so there is no missing dependency.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience


howardjohn (approved):

howardjohn:
This broke `BUILD_ALL=false`, it builds everything now unfortunately

PR #34730, opened 2021-08-17
fix e2e test when go test TestGateway with flag istio.test.kube.loadbalancer=false

ChenglongGao327:
local test e2e , if go test with flag istio.test.kube.loadbalancer=false ,there will get port error because of get port twice.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn (changes requested):
this ends up calling the wrong port now. It should be the tcp port

ChenglongGao327:
> this ends up calling the wrong port now. It should be the tcp port

yes，it's wrong

PR #34737, opened 2021-08-17
Fix eds debounce CommittedUpdates

andrewshaoyu:
if `PILOT_ENABLE_EDS_DEBOUNCE` equals false , commitupdate should also be increased.
otherwise pilot server may not be ready.

``` go
func (s *DiscoveryServer) ConfigUpdate(req *model.PushRequest) {
	inboundConfigUpdates.Increment()
	s.InboundUpdates.Inc()
	s.pushChannel <- req
}
```
During `ConfigUpdate` inboundUpdate++,then push model.request,

``` go
push := func(req *model.PushRequest, debouncedEvents int) {
    pushFn(req)
    updateSent.Add(int64(debouncedEvents))
    freeCh <- struct{}{}
}
```

but when enableEDSDebounce, only do `pushFn`.  So InboundUpdates is greater than CommittedUpdates.

``` go
if !opts.enableEDSDebounce && !r.Full {
    // trigger push now, just for EDS
    go pushFn(r)
    continue
}
```

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/pilot/pkg/xds/discovery.go
+++ b/pilot/pkg/xds/discovery.go
@@ -439,7 +439,10 @@ func debounce(ch chan *model.PushRequest, stopCh <-chan struct{}, opts debounceO
 			}
 			if !opts.enableEDSDebounce && !r.Full {
 				// trigger push now, just for EDS
-				go pushFn(r)
+				go func(req *model.PushRequest) {
+					pushFn(req)
+					updateSent.Inc()
+				}(r)
 				continue
 			}
 


hzxuzhonghu:
@andrewshaoyu Could you elaborate the `bug`?

pilot/pkg/model/push_context.go
```diff
@@ -320,6 +320,9 @@ type PushRequest struct {
 	// There should only be multiple reasons if the push request is the result of two distinct triggers, rather than
 	// classifying a single trigger as having multiple reasons.
 	Reason []TriggerReason
+
+	// DebounceCount represents the times a push was merged. Default value should be 1.
+	DebounceCount int64
```
hzxuzhonghu, line 325: nit: you can make it default 0, and add one when  added to `CommittedUpdates`

hzxuzhonghu: So  it  means the debounce merge times, and can get rid of the below unreadable code

hzxuzhonghu (commented):
lgtm, i am thinking if you can add a ut

pilot/pkg/model/push_context.go
andrewshaoyu: yes,i will remove `Default value should be 1.`

andrewshaoyu:
> lgtm, i am thinking if you can add a ut

I have not found the ut about `CommittedUpdates`.  So make a ut which can indicate the bug?

howardjohn (changes requested):
Why not just replace
```
go pushFn(r)
```
with 
```go
go func() {
  pushFn(r)
  updateSent.Add(int64(debouncedEvents))
}
```
?

andrewshaoyu:
``` go
	if !opts.enableEDSDebounce && !r.Full {
		// trigger push now, just for EDS
                go func() {
                  pushFn(r)
                  updateSent.Inc()
                }
		continue
	}
```
If !enableEDSDebounce, pushFn will only commit current PushRequest r, now  just do `updateSent.Inc()`. But How many PushRequests In r ? currently it is one. 

Calculating debounce times when `req.merge(r)`, it will be more clear. We do not care about where to clear debouncedEvents.

I think  how many times the PushRequest was merged is one of PushRequest's Properties. then we can add a metric to expose it.

howardjohn:
> If !enableEDSDebounce, pushFn will only commit current PushRequest r, now just do updateSent.Inc(). But How many PushRequests In r ? currently it is one.

Its always 1, right? So you can do
```
      go func() {
                  pushFn(r)
                  updateSent.Inc()
                }
```

then we don't need all the other changes

hzxuzhonghu (commented):
Is the merge right? When two pushrequest merged, the debouncedEvents = 1

hzxuzhonghu:
And updateSent is incremented by one

andrewshaoyu:
now when !enableedsdebounce, pushrequest will be sent right now without merged. so it should be 1.

hzxuzhonghu:
Then how about the other case debounce enabled?

andrewshaoyu:
now using debounceevents as event mered times. If we add a new debounce function， we need to caculate the merged times.

Xunzhuo (commented):
LGTM

hzxuzhonghu:
What if two push request merged, the debounceEvents=1

hzxuzhonghu:
IC, req is nil when the first time it runs `req = req.Merge(r)`

hzxuzhonghu (approved):
LGTM, please add a ut

hzxuzhonghu:
it looks this is not needed any more

andrewshaoyu:
Have been fixed in latest release?

hzxuzhonghu:
I am wrong, nope. 

@howardjohn Another look

howardjohn (approved):

PR #34747, opened 2021-08-17
Deep merge trust domain aliases

howardjohn:
This is useful with the share mesh config feature, as it allows this
    field to be essentially append-only. This follows the pattern used by
    `providers`.

### DIFF
--- a/pkg/config/mesh/mesh.go
+++ b/pkg/config/mesh/mesh.go
@@ -27,6 +27,7 @@ import (
 
 	meshconfig "istio.io/api/mesh/v1alpha1"
 	"istio.io/api/networking/v1alpha3"
+	"istio.io/istio/pilot/pkg/util/sets"
 	"istio.io/istio/pkg/config/constants"
 	"istio.io/istio/pkg/config/validation"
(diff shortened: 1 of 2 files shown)


pkg/config/mesh/mesh.go
```diff
@@ -220,6 +222,8 @@ func ApplyMeshConfig(yaml string, defaultConfig meshconfig.MeshConfig) (*meshcon
 		}
 	}
 
+	defaultConfig.TrustDomainAliases = sets.NewSet(append(defaultConfig.TrustDomainAliases, prevTrustDomainAliases...)...).SortedList()
```
hzxuzhonghu, line 225: What about the other fields? Like CaCertificates DiscoverySelectors

shankgan (approved):

pkg/config/mesh/mesh.go
shankgan: Since we are selecting default value for a field, can that be done in a separate PR?

hzxuzhonghu: make sense

hzxuzhonghu (approved):

PR #34774, opened 2021-08-19
Allow http proxy to be enabled per workload instead of global.

costinm:
Using http proxy for workloads without iptables is a huge pain - requires mesh-wide setting that affects the size of the config for all other workloads.

Eventually we should deprecate the MeshConfig.HttpProxyPort - but since that may break stuff
we can just leave it there. 

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu:
lgtm, this should be per proxy config

pilot/pkg/networking/core/v1alpha3/listener_test.go
```diff
@@ -1643,6 +1643,32 @@ func TestHttpProxyListener(t *testing.T) {
 	}
 }
 
+func TestHttpProxyListenerPerWorkload(t *testing.T) {
+	p := &fakePlugin{}
+	configgen := NewConfigGenerator([]plugin.Plugin{p}, &model.DisabledCache{})
```
ramaraochavali, line 1648: Can you please add a test case to cover the case that has mesh value and show that sidecar one overrides it as well - just for completeness?

costinm: The intent was to eventually deprecate the MeshConfig value - in reality both tests are pretty superficial. I will add more testing - I am working on some environments without iptables. It's not a big deal to cut&paste this test again and set both, but not sure it adds much value.

ramaraochavali: If it is permanent it makes sense to add test. If it is temporary and mesh config value is deprecated, I am fine without test.

costinm: Yes, mesh config HttpProxyPort needs to be deprecated - it's not really usable and doesn't make sense. You may have few machines or docker images where iptables is missing - but setting the global flag will increase the config size for ALL  the other workloads in the mesh.

Technically the entire MeshConfig needs to be deprecated and replaced with narrower-scoped APIs, we're starting with ProxyConfig.

ramaraochavali (approved):

PR #34778, opened 2021-08-19
make dns record ttl configurable

ChenglongGao327:
If dns capture is enabled, different scenarios have different requirements for dns recording ttl, So making this configurable.

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/dns/client/dns.go
+++ b/pkg/dns/client/dns.go
@@ -27,10 +27,16 @@ import (
 	"istio.io/istio/pilot/pkg/serviceregistry/provider"
 	"istio.io/istio/pkg/config/host"
 	dnsProto "istio.io/istio/pkg/dns/proto"
+	"istio.io/pkg/env"
 	istiolog "istio.io/pkg/log"
 )
 
-var log = istiolog.RegisterScope("dns", "Istio DNS proxy", 0)
+var (
+	log = istiolog.RegisterScope("dns", "Istio DNS proxy", 0)
+
+	dnsRecordTTL = env.RegisterDurationVar("DNS_RECORD_TTL", 30*time.Second,
+		"DNS record TTL, in case the client decides to honor the TTL, keep it low so that we can always serve the latest IP for a host.").Get().Seconds()
+)
 
 // LocalDNSServer holds configurations for the DNS downstreamUDPServer in Istio Agent
 type LocalDNSServer struct {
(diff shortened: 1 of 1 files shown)


pkg/istio-agent/agent.go
```diff
@@ -467,7 +467,7 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
 }
 
 func (a *Agent) initLocalDNSServer() (err error) {
-	// we dont need dns server on gateways
+	// we don't need dns server on gateways
```
howardjohn, line 470: Please avoid refactoring in a PR with real changes.. it makes it hard to review

howardjohn (changes requested):
Can you discuss the use case a bit more? I think we want to ensure there are good reasons to configure things before we add more hacky env vars

ChenglongGao327:
> Can you discuss the use case a bit more? I think we want to ensure there are good reasons to configure things before we add more hacky env vars

For scenarios that run stably and require frequent domain name resolution, set the ttl to a larger value to reduce the number of times to access DNS; in other scenarios, ttl can be set to be more general.

pkg/istio-agent/agent.go
ChenglongGao327: > Please avoid refactoring in a PR with real changes.. it makes it hard to review

Sorry,  i got it

howardjohn:
is it something that makes sense as a global option or per service? I
cannot imagine needing low TTL for a cluster ip - it literally never
changes. But for headless service smaller TTL makes more sense

ChenglongGao327:
> is it something that makes sense as a global option or per service? I cannot imagine needing low TTL for a cluster ip - it literally never changes. But for headless service smaller TTL makes more sense

Yes, the clusterIP ttl needs to be set larger, and the headless ttl needs to be set smaller. 
I want to distinguish headless service and clusterIP，and set different ttl. such as through kubernetes and kube-dns clusterIP
When connecting to k8s, if the IP address corresponding to the service domain name is not in the same network segment as the kubernetes and kube-dns service cluster IP addresses, then the service is considered a headless service.

howardjohn:
Do you have a real use case for this, or just saw a TODO? I would avoid fixing TODOs without a strong motivation, usually they mean "TODO [if we ever have a compelling use case])".

> IP address corresponding to the service domain name is not in the same network segment as the kubernetes
We do not need this implicit logic, we know if its headless or not in the control plane

ChenglongGao327:
> Do you have a real use case for this, or just saw a TODO? I would avoid fixing TODOs without a strong motivation, usually they mean "TODO [if we ever have a compelling use case])".

I saw a TODO and think that it is better to set different ttl for different scenarios, so making it configurable

howardjohn:
I highly recommend not implementing TODOs like this - the reason they were
not done is that they add a lot of complexity and we haven't seen any real
world demand to warrant that complexity

ChenglongGao327:
> I highly recommend not implementing TODOs like this - the reason they were not done is that they add a lot of complexity and we haven't seen any real world demand to warrant that complexity

ok, i see. Thanks for your reviews

PR #34784, opened 2021-08-19
Fix endpoint filter

hzxuzhonghu:

It could be this case, 

cluster1, neytwork1:

     -  3 eastwest gateway
     -  1 endpoint (10.0.0.1)

cluster2, network2:

     -  2 eastwest gateway (20.0.0.1) (20.0.0.2)
     -  1 endpoint (20.0.0.100)

Take this as an example, for a proxy in cluster1:

** Without this pr:

endpoints:
  - 10.0.0.1 weight 3
  - 20.0.0.1 weight 2
  - 20.0.0.2 weight 1

We can see the two gateways of networks get unbalanced weight. As the real endpoints in network2 grows, the unbalance becomes severe. 

** With this pr:

endpoints:
  - 10.0.0.1 weight 6
  - 20.0.0.1 weight 3
  - 20.0.0.2 weight 3

The gateways of network2 have equal lb weight

    

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/network.go
+++ b/pilot/pkg/model/network.go
@@ -15,6 +15,7 @@
 package model
 
 import (
+	"fmt"
 	"net"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


hzxuzhonghu:
Follow #34781 to better understand what it is doing.

stevenctl:
```
** Without this pr:

10.0.0.1 weight 6
20.0.0.1 weight 3
20.0.0.2 weight 3
```

```
** With this pr:

10.0.0.1 weight 3
20.0.0.1 weight 2
20.0.0.2 weight 1
```

It seems like the original works as intended? There is 1 actual endpoint behind each of the `20.`, so the weights of `20.` gateways should sum to the weight of the local endpoint. 

Even the `With this pr` change isn't _terrible_. The  total weight to the underlying endpoint is the same, but we use one gateway more than the other.

nmittler:
I agree with @stevenctl ... I'm not sure I understand the problem. The description of this issue appears to be correct behavior.

hzxuzhonghu:
It's the opposite, i wrote wrong, will update

pilot/pkg/xds/ep_filters.go
```diff
@@ -203,6 +183,20 @@ func (b *EndpointBuilder) scaleEndpointLBWeight(ep *endpoint.LbEndpoint, scaleFa
...
+// Apply the weight for this endpoint to the network gateways.
+func splitWeightAmongGateways(weight uint32, gateways []*model.NetworkGateway, gatewayWeights map[model.NetworkGateway]uint32) {
+	// Spread the remaining weight across the gateways.
+	weightPerGateway := weight / uint32(len(gateways))
+	remain := weight % uint32(len(gateways))
+	for i, gateway := range gateways {
+		gatewayWeights[*gateway] += weightPerGateway
+		if uint32(i) < remain {
+			// This should not happen, just in case
+			gatewayWeights[*gateway]++
```
hzxuzhonghu, line 195: This is the difference, `remain := weight % uint32(len(gateways))` the remain must be 0.

pilot/pkg/model/network.go
```diff
@@ -180,3 +184,22 @@ func SortGateways(gws []*NetworkGateway) []*NetworkGateway {
...
+			x = y
+			y = tmp
+		} else {
+			return y
+		}
+	}
+}
+
+// least common multiple of x and y
+func lcm(x, y int) int {
```
stevenctl, line 203: +1

pilot/pkg/xds/ep_filters_test.go
```diff
@@ -161,15 +164,16 @@ var networkFiltered = []networkFilterCase{
 			{
 				lbEps: []LbEpInfo{
 					// 1 local endpoint on network4
-					{address: "40.0.0.1", weight: 2},
+					{address: "40.0.0.1", weight: 6},
```
stevenctl, line 167: Why is our  scale much larger now? I thought the lcm logic would help us keep these values lower.

hzxuzhonghu: I added gateway numbers in the test.

pilot/pkg/xds/ep_filters_test.go
```diff
@@ -681,7 +685,7 @@ func xdsConnection(nw network.ID, c cluster.ID) *Connection {
 
 // environment defines the networks with:
 //  - 1 gateway for network1
-//  - 2 gateway for network2
+//  - 3 gateway for network2
 //  - 1 gateway for network3
 //  - 0 gateways for network4
```
hzxuzhonghu, line 690: I added gateway numbers here, and the lcm is 3, because we calculated based on per network gws, per cluster gws

hzxuzhonghu: Cluster2a has 1 gw, cluster2b has 2 now

pilot/pkg/xds/ep_filters_test.go
stevenctl: saw that once I scrolled a bit more – thanks!

pilot/pkg/xds/ep_filters.go
```diff
@@ -44,10 +44,14 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocLbEndpointsAn
...
-	// Scale all weights by the maximum number of gateways that might appear for a network.
+	// Scale all weights by the lcm of gateways per network and gateways per cluster.
 	// This will allow us to more easily spread traffic to the endpoint across multiple
 	// network gateways, increasing reliability of the endpoint.
-	scaleFactor := b.push.NetworkManager().GetMaxGatewaysPerNetwork()
+	divisor := uint32(len(b.push.NetworkManager().GatewaysForNetwork(b.network)))
+	if divisor == 0 {
+		divisor = 1
+	}
+	scaleFactor := b.push.NetworkManager().GetLCM() / divisor
```
stevenctl, line 54: Can we just condense these 4 lines into `NetworkManager().GetScaleFactor(b.network)`

hzxuzhonghu: Good, can do that

hzxuzhonghu: done

stevenctl (approved):

PR #34789, opened 2021-08-19
Implement gateway-api v1alpha2 API

howardjohn:
This implements the new gateway-api v1alpha2 API. Per guidance from the project, we will be dropping support for v1alpha1 and directly moving to v1alpha2 without supporting both APIs simultaneously. 

For release notes, see https://github.com/kubernetes-sigs/gateway-api/releases/tag/v0.4.0-rc1

The biggest change in this version is the new Gateway<->Route binding mechanism.

This PR is built upon https://github.com/istio/istio/pull/34910; we may want to merge that first. This also depends on (already merged) PR #34852.

Note to reviewers: sorry its huge... its pretty hard to split this up. I recommend just pulling up the `pilot/pkg/config/kube/gateway` folder in your local editor.


pilot/pkg/features/pilot.go
```diff
@@ -300,7 +300,7 @@ var (
 		return durationpb.New(defaultRequestTimeoutVar.Get())
 	}()
 
-	EnableServiceApis = env.RegisterBoolVar("PILOT_ENABLED_SERVICE_APIS", true,
+	EnableGatewayAPI = env.RegisterBoolVar("PILOT_ENABLED_SERVICE_APIS", true,
```
costinm, line 303: Should we rename the env varialbe as well ?

costinm (approved):
Most of the PR is tests and gateway-specific code. I looked in particular for changes to the non-gateway code - and didn't find anything interesting. Will start testing the new gateway code after it is submitted, it is new API so won't affect existing users even if there are bugs.

pilot/pkg/features/pilot.go
howardjohn: Probably not a bad idea to do it now as its already a discrete migration to the new api. Its at least opt-out now so the flag is hopefully very uncommon. will send an update

howardjohn:
TODO : change leader election identity

PR #34800, opened 2021-08-20
add echo-grpc sample for doc

stevenctl:
see https://github.com/istio/istio.io/pull/10168


ericvn:
The one question I have is do you want to use `samples/echo` (as in the blog post PR), or `samples/echo-grpc` as shown here?

stevenctl:
Good point - also should we make this more generic and open up the same set of ports we have for integration tests? Then we can use the same sample for tcp, http and a bunch of other stuff.

ericvn:
> Not really. This is actually a different app, and it supports pretty much any protocol (tcp, gRPC, http, etc) - while my doc is gRPC based, this could be used for other things if needed

I notice that the yaml does call out the echo-grpc namespace as well as other grpc values.

stevenctl:
I guess with all the special annotations it's not worth trying to make this re-usable. Renamed back to echo-grpc, will update the blogpost.

ericvn:
I'm fine with `echo-grpc`, although I understand @rvennam comment about switching the order so echo is `grpc` is first as `tcp` is first

PR #34818, opened 2021-08-22
Support multiple hosts and Cluster VIPs

nmittler:
This is laying the groundwork for future features:

- IPv6, which requires that services support multiple cluster VIPs.
- MCS, which requires a separate host and cluster VIP for the clusterset.

Summary of changes:

- Introducing `cluster.AddressMap`, which handles the synchronization for the VIPs. This is now used both for ClusterVIPs and ClusterExternalAddresses.
- Introducing `model.HostVIPs`, which combines a hostname with the ClusterVIPs.
- Added service field `ClusterLocal`, which is the `HostVIPs` for the `<svc>.<namespace>.svc.cluster.local` host. This is currently the only host until a second is added for the MCS `clusterset.local` host.
- Removed `Hostname`, `ClusterVIPs`, and `Mutex` from service, as they were replaced by `ClusterLocal`.

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu:
Before looking through the refactor, i want to first ask a question:  

according to MCS kep, it is optional but not recommended to use MCS DNS (https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#dns)

What's the difference between `<service>.<ns>.svc.clusterset.local` and `<service>.<ns>.svc.cluster.local.` in the end? Will you change the cluster.local to access only cluster local endpoints?

nmittler:
@hzxuzhonghu 

> according to MCS kep, it is optional but not recommended to use MCS DNS (https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#dns)

There's really no way to make this work MCS work properly without `clusterset.local`. I'm not entirely sure why the KEP lists it as optional.

> What's the difference between <service>.<ns>.svc.clusterset.local and <service>.<ns>.svc.cluster.local. in the end? Will you change the cluster.local to access only cluster local endpoints?

Yes, we will gradually transition `cluster.local` over to being actually cluster local. See the [RFC](https://docs.google.com/document/d/1jTamFFXPhjZiLoQbspRaMW07nxE0tEN6qqqUUDqkTQI/edit) for details.

pilot/pkg/model/service.go
```diff
@@ -70,11 +79,11 @@ type Service struct {
 	// CreationTime records the time this service was created, if available.
 	CreationTime time.Time `json:"creationTime,omitempty"`
 
-	// Name of the service, e.g. "catalog.mystore.com"
-	Hostname host.Name `json:"hostname"`
```
hzxuzhonghu, line 74: Please revert, it is so widely used.  If you want to support mcs, it is better to add a new field

nmittler: I don't think the fact that it's used widely is a sufficient reason to keep it. If we're introducing multiple hostnames, leaving this creates confusion and room for misuse.

I'd like to get thoughts from others: @howardjohn @jacob-delgado thoughts?

jacob-delgado:
cc @ycai-aspen Let's review this as it pertains to dual stack changes.

pilot/pkg/model/service.go
howardjohn: One thing we could do is:
```
	// ClusterLocal specifies the cluster.local host and cluster VIPs for this service.
	HostVIPs `json:"clusterLocalHostVIPs,omitempty"`
```

Then access can just be like `svc.Hostname`. When we add clusterset, it would be `ClusterSet HostVIPs` so access is like `svc.ClusterSet.Hostname`. The benefit here is its simpler and minimal changes; the downside is its less explicit.

So I guess if we want all callers to ask themselves: "Should I use the ClusterLocal or the ClusterSet IP?", then they way we have now is probably right. If instead using ClusterLocal is right in most cases and we only need the ClusterSet one in some isolated multicluster code, then I think embedded or other approach may be simpler.

I think right now we are in the latter case (using ClusterLocal is right in most cases - really all cases today as ClusterSet does not exist). Do we expect that to change in the near future?

pilot/pkg/model/service.go
```diff
@@ -58,7 +67,7 @@ import (
 type Service struct {
 	// Attributes contains additional attributes associated with the service
 	// used mostly by RBAC for policy enforcement purposes.
-	Attributes ServiceAttributes
+	Attributes *ServiceAttributes
```
howardjohn, line 70: I think making this a metric degrades this a bit. I assume this is to make the mutex work. Would it be simpler to instead make L54 `ClusterVIPs *cluster.AddressMap`?

pilot/pkg/model/service.go
nmittler: > So I guess if we want all callers to ask themselves: "Should I use the ClusterLocal or the ClusterSet IP?", then they way we have now is probably right.

This is exactly my thinking. By making each call site use `svc.ClusterLocal.xxx`, we give the fields themselves way more context, so it will be easier to identify bugs going forward.

> I think right now we are in the latter case (using ClusterLocal is right in most cases - really all cases today as ClusterSet does not exist). Do we expect that to change in the near future?

I do expect this to change in the future. I expect us to get to a point where MCS is a default feature of k8s and Istio uses it by default to drive service discoverability. IMO it would be better to just take the refactoring hit now, and remove all of the assumptions of a logical default.

pilot/pkg/model/service.go
nmittler: I assume you meant "pointer"? In what way are you thinking it degrades this? Are you thinking from a perf perspective?

howardjohn: (likely very) minor perf hit and we either need to make sure all Services have this set or all callers have nil checks. Its nice to just have a plain struct that we never need to worry about these

nmittler: Yeah I struggled with this too ... it is nice to just have a plain struct. The reason I went down this path is that the existence of creation of the ClusterVIPs seems to need protection. If we make the structure dynamically created, we would have to move the mutex up a level which gets ugly. By making it a plain struct within `ServiceAttributes`, we avoid the issue. The down side is that every time we create service, we have to also create `ServiceAttributes`, which is a bit annoying but doable.

pilot/pkg/model/service.go
hzxuzhonghu: I point it because, model.Service is a core API used internally.  We are in Day-2 stage As i know, how would people think?


Another reason is MCS is still an alpha feature, not sure how many users. It is not wise to introduce an alpha support but may break a stable one.

pilot/pkg/model/service.go
```diff
@@ -651,23 +653,22 @@ func ParseSubsetKey(s string) (direction TrafficDirection, subsetName string, ho
 	return
 }
 
-// GetServiceAddressForProxy returns a Service's IP address specific to the cluster where the node resides
-func (s *Service) GetServiceAddressForProxy(node *Proxy) string {
-	clusterIP := func() string {
-		if node.Metadata == nil || node.Metadata.ClusterID == "" {
-			return ""
+// GetClusterLocalIPv4AddressForProxy returns a Service's IPv4 address specific to the cluster where the node resides
+func (s *Service) GetClusterLocalIPv4AddressForProxy(node *Proxy) string {
```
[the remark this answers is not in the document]
howardjohn: +1, I think its better to not say ipv4 - the comment was inaccurate. We do have an ipv6 test

howardjohn (approved):

nmittler:
@howardjohn @ycai-aspen @hzxuzhonghu PTAL

pilot/pkg/model/service.go
jacob-delgado: `model.Service` is only used internally. Adding it to be cluster-aware will help devs know the specific places where MCS is different. We've had issues trying to modifying `model.Service` by just adding new, similar fields, but it has made finding subtle issues hard to track. I think this is the right approach.

pilot/pkg/model/service.go
```diff
@@ -45,6 +46,16 @@ import (
...
 )
 
+// HostVIPs provides the VIPs in each cluster for a given host.
+type HostVIPs struct {
+	// Name of the service, e.g. "catalog.mystore.com"
+	Hostname host.Name `json:"hostname"`
+
+	// ClusterVIPs specifies the service address of the load balancer
+	// in each of the clusters where the service resides
+	ClusterVIPs cluster.AddressMap `json:"clusterVIPs,omitempty"`
```
hzxuzhonghu, line 56: nit: why not define AddressMap in this pkg

pilot/pkg/model/service.go
hzxuzhonghu: I prefer @howardjohn 's suggestion inlined field makes minor change to istio, but since you have manually changed all places, i am ok with either.

pkg/cluster/addressmap.go
```diff
@@ -0,0 +1,130 @@
...
+import "sync"
+
+// AddressMap provides a thread-safe mapping of addresses for each Kubernetes cluster.
+type AddressMap struct {
+	// Addresses hold the underlying map. Visible only for testing, for the purposes of simplified construction.
+	// Production code should never access this directly.
+	Addresses map[ID][]string
+
+	// NOTE: The copystructure library is not able to copy unexported fields, so the mutex will not be copied.
+	mutex sync.RWMutex
```
hzxuzhonghu, line 26: You can add a tag ``copy:"ignore"``, so that the mutex will be ignored when copied

nmittler: It's not needed. Private fields are not copied by copystructure.

pilot/pkg/model/service.go
nmittler: It uses `cluster.Hostname`. I chose to move it to that package since it doesn't really have any dependencies on pilot. No real strong motivation.

nmittler:
@hzxuzhonghu 

> I prefer @howardjohn 's suggestion inlined field makes minor change to istio, but since you have manually changed all places, i am ok with either.

@howardjohn is onboard with the new approach since this will be quickly followed up with the introduction of a second field for MCS. He's already approved.

PR #34820, opened 2021-08-23
add imagePullPolicy configurable for istio-proxy

ChenglongGao327:
local test e2e, container istio-proxy always pull image even though go test with flag  `istio.test.pullpolicy=IfNotPresent` and local node has this image. So add imagePullPolicy configurable for istio-proxy, convenient for local testing.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience


howardjohn (commented):
don't we pass it to the install? if it's not being set sounds like it might be a real bug

ChenglongGao327:
> don't we pass it to the install? if it's not being set sounds like it might be a real bug

yes, there is no `imagePullPolicy` set in the test app deployment yaml and the default value `Always` is used.

hzxuzhonghu (approved):

howardjohn (changes requested):
https://github.com/istio/istio/blob/16debb5891fd3fcc91bccd3837df7a93de500fc5/pkg/test/framework/components/istio/config.go#L312

Its because of ^. If we want to change it, we should change there

ChenglongGao327:
https://github.com/istio/istio/blob/16debb5891fd3fcc91bccd3837df7a93de500fc5/pkg/test/framework/components/istio/config.go#L312
> Its because of ^. If we want to change it, we should change there

L312 No need to modify, this is correct, it only works for containers other than istio-proxy.
i use this command to test 
```
export HUB="istio-test"
export TAG="v1"
go test -tags integ  -v ./tests/integration/pilot/  -istio.test.env kube -istio.test.hub=${HUB} -istio.test.tag=${TAG}  -istio.test.pullpolicy=IfNotPresent  -istio.test.kube.loadbalancer=false  -run TestXFFGateway
```
set `istio.test.tag=v1` `istio.test.pullpolicy=IfNotPresent` , the tag is not latest
1.  Container istio-proxy always pull image ,not valid for istio-proxy, but valid for other containers, such as app a-v1/b-v1 etc.
2. Application specific to testing TestXFFGateway, if imagPullPolicy is not set , will use default policy `Always`.

so, for container istio-proxy and application specific to testing TestXXX i add imagePullPolicy to use istio.test.pullpolicy. In this way, the imagePullPolicy of all containers can be configured through flag `istio.test.pullpolicy`.

howardjohn (approved):
Ah I understand now... k8s defaults the imagePullPolicy _before_ the webhook is created. So its kind of a bug/limitation in Istio injector

Issue #26932, opened 2020-08-31
istio-init container has huge resources limits


MarianZoll:
**Bug description**

After upgrading to Istio 1.7 (we also switched to operator managed mode), the `istio-init` container resources limits are 2 CPUs and 1Gi memory. This is very resource-intensive for deployments that are expected to have very less traffic inside the mesh. 

As result, the node-based resource allocations would cause operations alerts in production.  

We already limit the resources of the `istio-proxy` by providing the following annotations:
```
sidecar.istio.io/proxyCPU: 75m
sidecar.istio.io/proxyCPULimit: 75m
sidecar.istio.io/proxyMemory: 75Mi
sidecar.istio.io/proxyMemoryLimit: 75Mi
sidecar.istio.io/rewriteAppHTTPProbers: "true"
```

We are looking for a similar way to restrict the limits of the `istio-init` container. From reading the docs of previous Istio releases, the limit annotations are quite new to `istio-init` as it was around (50m, 100Mi in 1.4). Additionally, you were able to configure them using:
```
global.proxy.init.resources.limits.cpu	100m	
global.proxy.init.resources.limits.memory	50Mi	
global.proxy.init.resources.requests.cpu	10m	
global.proxy.init.resources.requests.memory	10Mi	
```

Maybe I am just looking at the wrong place, but I at least hope that the issue if it is just a link to a documentation might help someone else as well. So sorry in case it is a no brainer 👍 

[ ] Docs
[ X ] Installation
[ ] Networking
[ X ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Expected behavior**

`istio-init` container should not have huge resources limits defined.

**Steps to reproduce the bug**

1. Deploy Istio 1.7 with the provided IstioOperator
2. Deploy a deployment in a istio-injected namespace
3. Run `kubectl get po -n $NAMESPACE $POD -o yaml` and check the resources k8s limits

**Version (include the output of `istioctl version --remote` and `kubectl version` and `helm version` if you used Helm)**

```
$ ./istioctl version --remote 
client version: 1.7.0
control plane version: 1.7.0
data plane version: 1.7.0 (17 proxies), 1.6.5 (5 proxies)
```

```
$ kubectl version
[...]

PR #34821, opened 2021-08-23
Fix init container resources reference error

kebe7jun:
- [X] Installation

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/files/injection-template.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/injection-template.yaml
@@ -109,8 +109,8 @@ spec:
         {{ end }}
     {{- end }}
   {{- else }}
-    {{- if .Values.global.proxy.resources }}
-      {{ toYaml .Values.global.proxy.resources | indent 6 }}
+    {{- if .Values.global.proxy_init.resources }}
+      {{ toYaml .Values.global.proxy_init.resources | indent 6 }}
     {{- end }}
   {{- end }}
     securityContext:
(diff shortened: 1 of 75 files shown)


manifests/charts/istio-control/istio-discovery/files/injection-template.yaml
```diff
@@ -110,7 +110,7 @@ spec:
     {{- end }}
   {{- else }}
     {{- if .Values.global.proxy.resources }}
```
bianpengyuan, line 112: This should also be changed to proxy_init.

kebe7jun: Thanks! : )

howardjohn (changes requested):
I don't think this is right. It looks right at first glance, but:

* It is a breaking change
* There is no reason I know of to have a different request for init vs container; if we request 1 CPU in the sidecar then its fine to request 1 CPU in the init container as well, as it will not have any negative impact on scheduling
* The request must be higher. IIRC the reason we made the change ^ is specifically because the 10m CPU requests led to the container getting throttled and taking >10s to setup iptables.
* Annotations like `sidecar.istio.io/proxyCPU` already apply to both.

See https://github.com/istio/istio/issues/26932

I would prefer if we just warn if its set maybe

kebe7jun:
> I don't think this is right. It looks right at first glance, but:
> 
> * It is a breaking change
> * There is no reason I know of to have a different request for init vs container; if we request 1 CPU in the sidecar then its fine to request 1 CPU in the init container as well, as it will not have any negative impact on scheduling
> * The request must be higher. IIRC the reason we made the change ^ is specifically because the 10m CPU requests led to the container getting throttled and taking >10s to setup iptables.
> * Annotations like `sidecar.istio.io/proxyCPU` already apply to both.
> 
> See #26932
> 
> I would prefer if we just warn if its set maybe

Right, so should we remove the `global.proxy_init.resources` setting to avoid misunderstanding it?

PR #34823, opened 2021-08-23
Fix sidecar's resources if set resource annotation and global values

### LINKED ISSUE #34822: sidecar-injection: global.proxy.resources cannot affect sidecar's resources if `sidecar.istio.io/proxyCPULimit` annotations exists
**Bug description**
If set `sidecar.istio.io/proxyCPULimit` annotation for a pod manually, the pod's memory limit will setted as empty, which makes the value `global.proxy.resources.limits.memory` not affect the sidecar's resources.

[X] Installation

**Expected behavior**
Pod's resources follow with global.proxy.resources even if set `sidecar.istio.io/proxyCPULimit` annotations.

1.11

kebe7jun:
Fix https://github.com/istio/istio/issues/34822

- [X] Installation

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
@@ -304,30 +304,20 @@ data:
             {{- end }}
           {{- end }}
             resources:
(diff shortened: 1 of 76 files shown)


kebe7jun:
@istio/wg-environments-maintainers PTAL.

howardjohn:
I don't think we can change this, its a breaking change. https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#customizing-injection is the preferred way to customize this anyways

PR #34825, opened 2021-08-23
add retriable status codes to retry policy

ramaraochavali:
If retry policy specifies Http Status codes, we add them to retriable-status-codes. But if user does not add `retriable-status-codes`  in the retry policy but just adds Http status they will not be retries. This PR fixes it.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/route/retry/retry.go
+++ b/pilot/pkg/networking/core/v1alpha3/route/retry/retry.go
@@ -81,6 +81,10 @@ func ConvertPolicy(in *networking.HTTPRetry) *route.RetryPolicy {
 		// Allow the incoming configuration to specify both Envoy RetryOn and RetriableStatusCodes. Any integers are
 		// assumed to be status codes.
 		out.RetryOn, out.RetriableStatusCodes = parseRetryOn(in.RetryOn)
+		// If user has just specified HTTP status codes in retryOn but have not specified "retriable-status-codes", let us add it.
+		if len(out.RetriableStatusCodes) > 0 && !strings.Contains(out.RetryOn, "retriable-status-codes") {
+			out.RetryOn += ",retriable-status-codes"
+		}
 	}
 
 	if in.PerTryTimeout != nil {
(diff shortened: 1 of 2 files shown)


pilot/pkg/networking/core/v1alpha3/route/retry/retry.go
```diff
@@ -81,6 +81,10 @@ func ConvertPolicy(in *networking.HTTPRetry) *route.RetryPolicy {
 		// Allow the incoming configuration to specify both Envoy RetryOn and RetriableStatusCodes. Any integers are
 		// assumed to be status codes.
 		out.RetryOn, out.RetriableStatusCodes = parseRetryOn(in.RetryOn)
+		// If user has just specified HTTP status codes in retryOn but have not specified "retriable-status-codes", let us add it.
+		if len(out.RetriableStatusCodes) > 0 && !strings.Contains(out.RetryOn, "retriable-status-codes") {
+			out.RetryOn += ",retriable-status-codes"
```
hzxuzhonghu, line 86: Isn't `RetryPolicy.RetriableStatusCodes` enough?

ramaraochavali: It is not enough

See api docs 

https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter.html?highlight=retriable%20status%20codes#x-envoy-retriable-status-codes

"Setting this header informs Envoy about what status codes should be considered retriable when used in conjunction with the retriable-status-code retry policy." - when used in conjunction with the retriable-status-code retry policy.

https://github.com/envoyproxy/envoy/blob/e071417f12809eddbf02052504fd9953eedce8f0/envoy/router/router.h#L245 (Envoy code)
https://github.com/envoyproxy/envoy/blob/main/source/common/router/retry_state_impl.cc#L345 - Considers retriable status code only when the policy is set

hzxuzhonghu: good, reading the code is more clear.

hzxuzhonghu (approved):

Issue #3308, opened 2018-02-08
Galley v1: Ensure that MCP based config distribution is controlled via a Helm flag


ozevren:
It is very likely that Galley will be in Alpha phase when Istio v1 ships. As such, the users should be able to opt-in to using Galley using a (standardized?) Istio mechanism.

(See https://github.com/istio/istio/issues/2082 for more general info).

ayj:
We probably ought to have standardized support for feature gates across Istio. Baking this into istioct/helm install is one option. Finer granularity may require flag support at the component level. An option to selectively enable/disable API groups may also be desirable.

refs

https://kubernetes.io/docs/reference/feature-gates/
https://kubernetes.io/docs/tasks/administer-cluster/cluster-management/#turn-on-or-off-an-api-version-for-your-cluster

PR #34828, opened 2021-08-23
add env to control watching cni configuration and binary files

### LINKED ISSUE #34777: Better support for CNI namespace migration
There can be cases where CNI DaemonSet is moved from one namespace to another, which means there can be brief period where two CNI installer runs simultaneously. After https://github.com/istio/istio/issues/34524 being fixed, the CNI installer will continuously reconcile the CNI artifacts on the node if there is any change. To avoid two installer steps on each other's toe, we need to provide a mechanism to stop watch & reinstall.

silenceshell:

fix #34777
Add CNI_FILE_WATCH_ENABLED to control watching cni configuration and binary files, default true.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn (commented):
Not sure I understand why you would stop watching vs delete the DS. Is it so it installs 1x but doesn't keep updating?

Don't we need the watch+update in case other CNIs like calico, etc mess with the files?

manifests/charts/istio-cni/values.yaml
```diff
@@ -16,6 +16,7 @@ cni:
   cniBinDir: /opt/cni/bin
   cniConfDir: /etc/cni/net.d
   cniConfFileName: ""
+  cniFileWatchEnabled: true
```
bianpengyuan, line 19: No need to add a value for this. I would prefer to just pass it through via istio operator.

bianpengyuan:
@howardjohn Yeah it is for migration, where the lifecycle of the duplicated CNI Daemonset is not under our control: #34777

cni/pkg/install/install.go
```diff
@@ -51,34 +51,39 @@ func NewInstaller(cfg *config.InstallConfig, isReady *atomic.Value) *Installer {
...
+		return
+	}
 
+	if in.cfg.CNIFileWatchEnabled {
 		if err = sleepCheckInstall(ctx, in.cfg, in.cniConfigFilepath, in.isReady); err != nil {
 			return
 		}
 		// Invalid config; pod set to "NotReady"
 		installLog.Info("Restarting...")
+		goto install
```
bianpengyuan, line 83: This would make container just quit, but the install-cni container still needs to do race repairing (i.e. we want it to be long running). An easy way is to make the new env variable control the install-cni to still watch in a loop, but only install the CNI for one time.

cni/pkg/config/config.go
```diff
@@ -39,6 +39,8 @@ type InstallConfig struct {
 	CNINetworkConfigFile string
 	// CNI config template string
 	CNINetworkConfig string
+	// Whether to watch and reinstall CNI configuration and binary files
+	CNIFileWatchEnabled bool
```
bianpengyuan, line 43: Can we add this new option to the configuration print at line 116?

cni/pkg/install/install.go
silenceshell: Thanks for the review.
After `goto install`, the container will reinstall cni files and waits for fsnotify event again, it wont quit in my opinion.
I am not sure whether we should watch the file or not if set env CNI_FILE_WATCH_ENABLED to false, maybe we should only wait for `ctx.Done`?

cni/pkg/config/config.go
silenceshell: done

manifests/charts/istio-cni/values.yaml
silenceshell: This is a little tricky to avoid a helm problem( [#3308](https://github.com/helm/helm/issues/3308)), it is quite complicated to use default value of a boolean variable.
So I set the default value in the `values.yaml`, and when migration, user could set value via istio operator like below.

```yaml
kind: IstioOperator
metadata:
  name: default
  namespace: istio-system
spec:
  components:
    cni:
      enabled: true
  values:
    cni:
      cniFileWatchEnabled: false
```

bianpengyuan: I was actually hoping not to expose this as part of values at all, since this is only useful for one time migration. User could either pass in the env var via IstioOperator k8s overlay like:

```
kind: IstioOperator
metadata:
  name: default
  namespace: istio-system
spec:
  components:
    cni:
      enabled: true
      k8s:
       env:
       - name: CNI_FILE_WATCH_ENABLED
          value: "false"
```

Or even manually edit the daemonset to add that env var.

cni/pkg/install/install.go
bianpengyuan: I built with your patch and tried it out. If we set `CNI_FILE_WATCH_ENABLED` to false, CNI container will just exit after first install, which is not expected. I think we can rename the env var to `CNI_ENABLE_REINSTALL` and only execute installation logic once if this is set to false: https://github.com/istio/istio/blob/e2d05310fe3b3daba1d9fbfd3bf0561b265c6a14/cni/pkg/install/install.go#L55-L75

silenceshell:
oh, I see. Will fix it later.
----------

该邮件从移动设备发送



--------------原始邮件--------------
发件人："Pengyuan Bian ***@***.***&gt;;
发送时间：2021年8月25日(星期三) 凌晨2:30
收件人："istio/istio" ***@***.***&gt;;
抄送："silenceshell ***@***.***&gt;;"Mention ***@***.***&gt;;
主题：Re: [istio/istio] add env to control watching cni configuration and binary files (#34828)
-----------------------------------

 

 
@bianpengyuan commented on this pull request.
 
 
In manifests/charts/istio-cni/values.yaml:
 &gt; @@ -16,6 +16,7 @@ cni:    cniBinDir: /opt/cni/bin    cniConfDir: /etc/cni/net.d    cniConfFileName: "" +  cniFileWatchEnabled: true  
I was actually hoping not to expose this as part of values at all, since this is only useful for one time migration. User could either pass in the env var via IstioOperator k8s overlay like:
 apiVersion: install.istio.io/v1alpha1 kind: IstioOperator metadata:   name: default   namespace: istio-system spec:   components:     cni:       enabled: true       k8s:        env:        - name: CNI_FILE_WATCH_ENABLED           value: "false"
 
Or even manually edit the daemonset to add that env var.

cni/pkg/install/install.go
```diff
@@ -48,35 +48,50 @@ func NewInstaller(cfg *config.InstallConfig, isReady *atomic.Value) *Installer {
...
+			if err = in.install(ctx); err != nil {
+				return
+			}
 		}
 
-		if err = sleepCheckInstall(ctx, in.cfg, in.cniConfigFilepath, in.isReady); err != nil {
-			return
-		}
 		// Invalid config; pod set to "NotReady"
 		installLog.Info("Restarting...")
```
bianpengyuan, line 96: Let's make the log more clear. We need to move this to line 88, and let's change it to be more informative (`Detect changes to the CNI configuration and binaries, attempt reinstalling...`). Then add an `else` block at line 93 with a log line saying (`Skip reinstalling CNI configuration and binaries.`)

bianpengyuan (commented):
LGTM, just a minor log comment. Thanks!

bianpengyuan (approved):

PR #9688, opened 2018-11-01
Support multiple network interfaces(#9441)

zhaohuabing:
Allows proxy to send multiple IP addresses to Pilot, Pilot uses all the
IP addresses to tell which services are located with the proxy and build
the corresponding inbound listeners for that proxy.

Fixes #9441

Signed-off-by: Huabing Zhao <zhaohuabing@gmail.com>

PR #34832, opened 2021-08-23
iptables: add exclude nic

lambdai:
Add a parameter in istio-iptables command to exclude local NIC from istio traffic capture.

Note that there is no corresponding pod annotation because there is no such use case in k8s yet.

Signed-off-by: Yuchen Dai <silentdai@gmail.com>

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


tools/istio-iptables/pkg/cmd/run.go
```diff
@@ -125,6 +125,7 @@ func (iptConfigurator *IptablesConfigurator) logConfig() {
 	b.WriteString(fmt.Sprintf("ISTIO_INBOUND_PORTS=%s\n", os.Getenv("ISTIO_INBOUND_PORTS")))
 	b.WriteString(fmt.Sprintf("ISTIO_OUTBOUND_PORTS=%s\n", os.Getenv("ISTIO_OUTBOUND_PORTS")))
 	b.WriteString(fmt.Sprintf("ISTIO_LOCAL_EXCLUDE_PORTS=%s\n", os.Getenv("ISTIO_LOCAL_EXCLUDE_PORTS")))
+	b.WriteString(fmt.Sprintf("ISTIO_EXCLUDE_NIC=%s\n", os.Getenv("ISTIO_EXCLUDE_NIC")))
```
howardjohn, line 128: Would recommend naming it `ISTIO_EXCLUDE_INTERFACES` for consistency

hzxuzhonghu:
But from pilot side , it builds rules for each adress of a proxy

lambdai:
> But from pilot side , it builds rules for each adress of a proxy

Thank you for the feedback!

Are you referring to VM or k8s? I don't think it's true for VM. 

I see pilot-agent collect a few ips but [barely one by default](https://github.com/istio/istio/blob/master/tools/packaging/common/istio-start.sh#L57)

We definitely have to override the above var SVC_IP to stripped the captured NIC ips besides this iptables exclusion rules. That's a separated PR though

hzxuzhonghu:
@lambdai #9688

lambdai:
> @lambdai #9688

Thanks for the pointer! So any workload, pod or VM, could have multiple ips and NICs.

I am not sure if I understand your point.

The goal is to exclude the capture from specified NICs

we need 
a. Skip capture at iptables(both inbound and outbound), must have so the traffic bypasses proxy before and after the proxy receives the xDS
b. Report fewer IPs to istiod so that fewer listener entries are generated. It's nice to have since this reduces the config size. 

Is there a conflict between a and b?

howardjohn:
Are you sure we still put the IP in xds? At least in normal case we do not,
maybe some edge cases (which are probably bugs)

lambdai:
@howardjohn Are we a little bit off the topic?

I am pretty sure I don't have multiple IPs at either inbound or outbound(in eds dump). However, regardless what the magic there, the traffic from/to the excluded NICs is captured without this PR.

hzxuzhonghu:
After looking, i doubt the same, seems it is meaningless now

tools/istio-iptables/pkg/cmd/run.go
```diff
@@ -247,6 +248,14 @@ func (iptConfigurator *IptablesConfigurator) shortCircuitKubeInternalInterface()
 	}
 }
 
+func (iptConfigurator *IptablesConfigurator) shortCircuitExcludeInterfaces() {
+	for _, excludeInterface := range split(iptConfigurator.cfg.ExcludeInterfaces) {
+		iptConfigurator.iptables.AppendRule(
+			constants.PREROUTING, constants.NAT, "-i", excludeInterface, "-j", constants.RETURN)
+		iptConfigurator.iptables.AppendRule(constants.OUTPUT, constants.NAT, "-i", excludeInterface, "-j", constants.RETURN)
```
hzxuzhonghu, line 255: What about inbound short circuit?

hzxuzhonghu, line 255: For TPROXY， maybe better apply it to mangle table

lambdai:
> After looking, i doubt the same, seems it is meaningless now

Yeah, some stitch is needed to generate the roles of NICs in the VM registration flow.
Theoretically in other service registraty as well. I will create an issue for other registry.

tools/istio-iptables/pkg/cmd/run.go
lambdai: Added.

hzxuzhonghu: nit:  only set to nat when it is not TPROXY

lambdai: > nit: only set to nat when it is not TPROXY

@hzxuzhonghu Can you explain a little bit?

I add "shortcircuit" to NAT in TPROXY because I see TPROXY does use NAT table: See https://github.com/istio/istio/blob/master/tools/istio-iptables/pkg/cmd/testdata/tproxy.golden#L18-L19

The NAT shortcircuit should be added to NAT unless all existing NAT are not hit, correct?

hzxuzhonghu (approved):

PR #34844, opened 2021-08-24
Remove deprecated option

Ruoyu-y:
* Remove unused option while find CA
* use const for well-known cert provision path

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/istio-agent/agent.go
+++ b/pkg/istio-agent/agent.go
@@ -84,6 +84,9 @@ const (
 	// this may be replaced with ./etc/certs, if a root-cert.pem is found, to
 	// handle secrets mounted from non-citadel CAs.
 	CitadelCACertPath = "./var/run/secrets/istio"
+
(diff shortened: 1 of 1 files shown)


Ruoyu-y:
@shankgan I removed the deprecated 'systemRootCert' option for checking root ca. Can you help with review? Thanks.

howardjohn (changes requested):
Can you explain more? I would not call `SystemRootCerts` deprecated - we are actively using this and its the only way currently to get "system root certs". Using `/etc/ssl/certs/ca-certificates.crt` is NOT an alternative - its is not portable to other OS.

In fact, we should update https://istio.io/latest/docs/setup/install/external-controlplane/#set-up-the-control-plane-in-the-external-cluster to use SYSTEM imo

Issue #34807, opened 2021-08-20
Investigate model.Service.Address


howardjohn:
This field is a bit suspicious - all references should probably be through ClusterVIPs. I suspect anything using it is actually a bug...

The one legit usage seems to be https://github.com/istio/istio/blob/21af48e524ed581e77a75385788b2c608e6f9caf/pilot/pkg/serviceregistry/aggregate/controller.go#L184, but I think we could replace it with `dst.ClusterVIPs[srcRegistry.Cluster()] = src.ClusterVIPs[srcRegistry.Cluster()]`

hzxuzhonghu:
src is the original svc converted from a single k8s registry, the `ClusterVIPs` is empty

hzxuzhonghu:
func ServiceToServiceEntry(svc *model.Service) *config.Config {

use svc.Address directly , maybe not right

hzxuzhonghu:
`buildOutboundListeners`

hzxuzhonghu:
From my investigation, the most common case does not use svc.Address directly

PR #34846, opened 2021-08-24
Fix service address misuse

hzxuzhonghu:

Fix: #34807 get the service address for the cluster

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/apigen/apigen.go
+++ b/pilot/pkg/networking/apigen/apigen.go
@@ -145,7 +145,7 @@ func (g *APIGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *
 			if s.Attributes.ServiceRegistry == provider.External {
 				continue
 			}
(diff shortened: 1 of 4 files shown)


nmittler:
@hzxuzhonghu I don't think this quite fixes the underlying cause of the issue. The field is named poorly or simply shouldn't be publicly accessible. This PR fixes the call sites for now, but new ones might very well crop up because it's not immediately clear what the field represents or how it should be used.

howardjohn:
I agree we should make it private or rename it

hzxuzhonghu:
I agree, will try to make it private. I was seeing some tests use this field directly.

hzxuzhonghu:
Make it private seems not viable.

adiprerepa (approved):

Issue #34758, opened 2021-08-18
istioctl manifest generate inconsistent default behavior


xinlaini:
Hello, I'm trying to use istioctl (1.10.3) to generate default istio manifests, but I'm seeing a behavior that can't be explained.
basically, if I run
```istioctl manifest generate > /tmp/no-arg.yaml```
vs
```istioctl manifest generate -f ./manifests/profiles/default.yaml > /tmp/with-default.yaml```
I expect them to generate identical manifests. However, the two outputs seem to differ at exactly one location - the former's istiod deployment doesn't have `spec.replicas` specified, but the latter has `spec.replicas` explicitly set to 1.

From the helm chart in manifests/charts/istio-control/istio-discovery/templates/deployment.yaml, it seems like the condition
 ```
{{- if not .Values.pilot.autoscaleEnabled }}
{{- if .Values.pilot.replicaCount }}
  replicas: {{ .Values.pilot.replicaCount }}
{{- end }}
{{- end }}
```
didn't take effect when `-f default.yaml` (which sets values.pilot.autoscaleEnabled to true) is explicitly passed.

howardjohn:
Interesting, I would also expect those to be the same... That being said, the end result show he functionally equivalent, right? replicas default is 1

xinlaini:
For the k8s behavior, yes, whether unsetting replicas or explicitly setting it to 1 is equivalent.
But the confusing part is why the `istioctl manifest generate command` did not seem to respect the if conditions set in the charts.
It's behaving as if the profile's `values.pilot.replicaCount` is set to `spec.replicas` unconditionally.

xinlaini:
To test my hypothesis further, I removed the two ifs from the chart, and change the replicas line to
```
replicas: "foo"
```

Then if default.yaml has `values.pilot.replicaCount: 42`, then the final manifest sets `replicas: 42`, if default.yaml has `values.pilot.replicaCount: 0`, then the final manifest keeps `replicas: foo`

Clearly there are some hard-coded behavior in istioctl binary, regardless of what the chart's logic seems to dictate.

howardjohn:
I think it somehow overlays the IstioOperator settings on top of the
values.yaml files inside the chart itself. Not 100% sure

xinlaini:
To add a little bit more confusion to the issue
Apparently, the results are different if you do 
```
istioctl manifest generate -f ./default.yaml -d ../
```
vs 
```
istioctl manifest generate --set profile=./default.yaml -d ../
[...]

PR #34849, opened 2021-08-24
fix istioctl manifest generate default inconsistent behavior

hanxiaop:
This PR is to fix issue https://github.com/istio/istio/issues/34758

Applying this fix will result in

`istioctl manifest generate`
and
`istioctl manifest generate -f ./manifests/profiles/default.yaml`
have the identical result.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/operator/cmd/mesh/manifest-generate.go
+++ b/operator/cmd/mesh/manifest-generate.go
@@ -99,6 +99,11 @@ func manifestGenerate(args *rootArgs, mgArgs *manifestGenerateArgs, logopts *log
 		return fmt.Errorf("could not configure logs: %s", err)
 	}
 
+	profile := manifest.GetValueForSetFlag(mgArgs.set, "profile")
+	if len(mgArgs.inFilename) == 0 && profile == "" || profile == name.DefaultProfileName {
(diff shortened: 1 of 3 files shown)


operator/cmd/mesh/manifest-generate.go
```diff
@@ -99,6 +99,15 @@ func manifestGenerate(args *rootArgs, mgArgs *manifestGenerateArgs, logopts *log
 		return fmt.Errorf("could not configure logs: %s", err)
 	}
 
+	inFileName := make([]string, 0)
+	for _, p := range mgArgs.inFilename {
+		if strings.Contains(p, fmt.Sprintf("%s.yaml", name.DefaultProfileName)) {
```
howardjohn, line 104: This seems like a fragile workaround. I could have a file `default.yaml` that has a ton of non-default customizations

monsteredc: @howardjohn Agree. Please take a look at the newer revision. Thanks!

monsteredc:
not stale

PR #34852, opened 2021-08-24
Refactor SDS credential references for reference policy

howardjohn:
This is a subset of https://github.com/istio/istio/pull/34789. It is broken out as that PR is already massive, this can be done for v1alpha1 and v1alpha2, and this is security sensitive code so it deserves a close review that is hard in a giant PR.

Overall changes:
* Change secret authorization to a per-credential basis. This allows a bad secret access to not block the entire SDS response. This is critical now that its possible to have invalid references on a per-credential basis
* Create a new `kubernetes-gateway://` secret type. This is an internal implementation detail - user facing `credentialName`/`certificateRef` API is unchanged. This is added so that we know if we should use the proxy's cluster (legacy behavior) or the config cluster (new behavior) for accessing the cert.
* Update gateway API tests to also test HTTPS. This also tests ^, as we only deploy the secret in the config clusters but test all ingresses.
* Add stub for ReferencePolicy. This just always returns false (Reference not allowed


pilot/pkg/model/credentials/resource.go
```diff
@@ -0,0 +1,100 @@
...
+	} else if strings.HasPrefix(resourceName, kubernetesGatewaySecretTypeURI) {
+		// Valid formats:
+		// * kubernetes-gateway://secret-namespace/secret-name
+		// Namespace is required. The secret is read from the config cluster; this is the primary difference from KubernetesSecretType.
+		res := strings.TrimPrefix(resourceName, kubernetesGatewaySecretTypeURI)
+		split := strings.Split(res, sep)
+		if len(split) <= 1 {
+			return SecretResource{}, fmt.Errorf("invalid resource name %q. Expected namespace and name", resourceName)
+		}
+		namespace := split[0]
```
stevenctl, line 95: what about empty cases "kubernetes-gateway:///name" or "kuberntes-gateway://ns/"

pilot/pkg/xds/fake.go
```diff
@@ -197,7 +197,7 @@ func NewFakeDiscoveryServer(t test.Failer, opts FakeOptions) *FakeDiscoveryServe
 		kubesecrets.DisableAuthorizationForTest(defaultKubeClient.Kube().(*fake.Clientset))
 	}
 	sc := kubesecrets.NewMulticluster(defaultKubeClient, "", "", stop)
-	s.Generators[v3.SecretType] = NewSecretGen(sc, s.Cache)
+	s.Generators[v3.SecretType] = NewSecretGen(sc, s.Cache, "")
```
stevenctl, line 200: nit, pass `opts.DefaultClusterName`

pilot/pkg/xds/sds.go
```diff
@@ -190,6 +130,15 @@ func (s *SecretGen) Generate(proxy *model.Proxy, push *model.PushContext, w *mod
 			}
 		}
 
+		// Fetch the appropriate cluster's secrets, based on the credential type
+		var secretController secrets.Controller
+		switch sr.Type {
+		case credentials.KubernetesGatewaySecretType:
```
stevenctl, line 136: Should/would this API ever allow for cross-cluster references? Where the proxy is co-located with it's secrets.

pilot/pkg/xds/sds.go
```diff
@@ -226,6 +175,72 @@ func (s *SecretGen) Generate(proxy *model.Proxy, push *model.PushContext, w *mod
...
+			if sameNamespace && isAuthorized() {
+				allowedResources = append(allowedResources, r)
+			}
+		default:
+			// Should never happen
+			log.Warnf("unknown credential type %q", r.Type)
+			pilotSDSCertificateErrors.Increment()
+		}
+		if verified || (sameNamespace && isAuthorized()) {
+			allowedResources = append(allowedResources, r)
```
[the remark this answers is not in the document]
stevenctl: I suppose we'd see some negative number in the log below if that's the case.

pilot/pkg/xds/sds.go
howardjohn: Good question. IMO no, it should always be the config cluster. A secret is just another config in this model.
I think this is how MCI works, which is the only real prior art I know of. Verifying with them now.

WDYT?

pilot/pkg/model/credentials/resource.go
howardjohn: Added a test case and error for these cases. Current callers of ParseResourceName appropriate handle errors, so I think its safe to make this return an error for these cases.

pilot/pkg/xds/sds.go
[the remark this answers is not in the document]
howardjohn: Thats a bug, thanks!

PR #34874, opened 2021-08-25
Simple Gateway chart for Helm

howardjohn:


jacob-delgado (approved):

manifests/charts/gateway/templates/role.yaml
```diff
@@ -0,0 +1,23 @@
+{{/*Set up roles for Istio Gateway. Note required for gateway-api*/}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "gateway.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
+rules:
```
costinm, line 7: Do we still need this ? Can we add an option to not create this role / binding ?

manifests/charts/gateway/templates/service.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: v1
```
costinm, line 1: Can you add an option to not create the service ? For example if service.type == 'NONE'

manifests/charts/gateway/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
costinm, line 1: If the option to not create the role bindings is set - maybe we can also use default SA instead of creating a dedicated SA.

costinm (approved):
Looks great, added few comments that can be addressed in sparate PRs ( or left as TODO)

manifests/charts/gateway/templates/serviceaccount.yaml
howardjohn: I will look into common practice. I think its still useful for things like WI, AuthorizationPolicies, etc.

manifests/charts/gateway/templates/service.yaml
howardjohn: ack, will add it soon

manifests/charts/gateway/templates/role.yaml
howardjohn: We need it if they are using Istio Gateways in another namespace only. For k8s gateway, or even Istio gateway if they are all in istio-system, its not required.

howardjohn: So maybe we can remove. or at least option

Monkeyanator (approved):

mick1627:
Is it possible to add a PodDisruptionBudget ressource.
service.yaml is not using  .Values.service.annotations to set specifc annotation

howardjohn:
> service.yaml is not using .Values.service.annotations to set specifc annotation

You can use `.Values.annotations`, is that sufficient?

> in service.yaml {{- if .Values.service.externalTrafficPolicy }} should be replace by {{- with .Values.service.externalTrafficPolicy }}

thanks!

mick1627:
> > service.yaml is not using .Values.service.annotations to set specifc annotation
> 
> You can use `.Values.annotations`, is that sufficient?
> 

In fact, we need to set annotations specific to the service to configure the loadbalancer.
For exemple, in aws, we use this kind of annotations : 

  annotations:
    service.beta.kubernetes.io/aws-load-balancer-internal: "true"
    service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "150"
    service.beta.kubernetes.io/aws-load-balancer-target-node-labels: "asm=ingress"

It's probably better to be able to specify this kind of annotation only for the service.

PR #34910, opened 2021-08-27
Enhance authority rewrite support to allow port and scheme

howardjohn:
Currently, the redirect.authority can take any input, such as `new-host`, `https://new-host:80`, etc. However, only `host` and `host:port` actually work. For example `https://new-host` would end up with envoy setting Location to `http://https://new-host`.

The new Gateway API (https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPRequestRedirect) has port, hostname, and scheme as valid options. Any combination of these three may be set.

This PR makes `redirect.authority` more flexible, by natively understanding `scheme://host:port`, where any of the three are optional, and converting directly into Envoy's support for these in redirects rather than always doing host redirect.

Backwards compatibility:
* `host` form: not impacted
* `host:port` form: not impacted. The envoy config will now be `hostRedirect: host; portRedirect: port` but the end result is the same
* All other forms: before it would generate invalid location like  `http://https://new-host`, now it works
Therefore, all previously valid configs are still valid

Alternatives:
* Modify the API to have port and scheme first class
* Do something similar to this PR, but make it internal-facing only so it can _only_ be access by gateway-api objects.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/route/route.go
+++ b/pilot/pkg/networking/core/v1alpha3/route/route.go
@@ -16,6 +16,7 @@ package route
 
 import (
 	"fmt"
+	"net"
 	"sort"
 	"strconv"
 	"strings"
@@ -410,13 +411,18 @@ func translateRoute(node *model.Proxy, in *networking.HTTPRoute,
 	}
 
 	if redirect := in.Redirect; redirect != nil {
+		scheme, redirectHost, redirectPort := splitRedirectAuthority(redirect.Authority)
 		action := &route.Route_Redirect{
 			Redirect: &route.RedirectAction{
-				HostRedirect: redirect.Authority,
+				HostRedirect: redirectHost,
+				PortRedirect: redirectPort,
 				PathRewriteSpecifier: &route.RedirectAction_PathRedirect{
 					PathRedirect: redirect.Uri,
 				},
 			},
+		}
+		if scheme != "" {
+			action.Redirect.SchemeRewriteSpecifier = &route.RedirectAction_SchemeRedirect{SchemeRedirect: scheme}
 		}
 
 		switch in.Redirect.RedirectCode {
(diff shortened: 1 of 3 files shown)


pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -539,6 +545,28 @@ func translateRoute(node *model.Proxy, in *networking.HTTPRoute,
...
+	for _, match := range []string{"https://", "http://"} {
+		if strings.HasPrefix(redirect, match) {
+			scheme = match[:len(match)-3]
+			redirect = strings.TrimPrefix(redirect, match)
+			break
+		}
+	}
+	h, p, err := net.SplitHostPort(redirect)
+	if err != nil {
+		host = redirect
```
adiprerepa, line 561: I don't really understand this part...host is set to the redirect URL if the redirect is malformed?

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -539,6 +545,28 @@ func translateRoute(node *model.Proxy, in *networking.HTTPRoute,
...
+		return
+	}
+	host = h
+	portNum, _ := strconv.Atoi(p)
```
adiprerepa, line 565: we should error handle here, right?

pilot/pkg/networking/core/v1alpha3/route/route.go
howardjohn: If there is no port -> the full thing is the host. If there is a port, host=h, port=p. I can expand the comments

pilot/pkg/networking/core/v1alpha3/route/route.go
howardjohn: SplitHostPort can never return a port that cannot be Atoi converted

hzxuzhonghu:
@howardjohn This would make the VirtualService.HTTPRedirect.Authority confusing, according to the rfc, authority does not include scheme. 

So prefer make it a first-class API.

ramaraochavali:
I also feel a first class api like `scheme` may be better and with that we can also support [scheme-redirect](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-redirectaction) in Envoy?

howardjohn:
ok I can add it as an API. We need port as well

howardjohn:
obsolete: https://github.com/istio/istio/pull/35146

PR #34728, opened 2021-08-16
Test golang 1.17 in presubmit

jacob-delgado:
Test golang 1.17 in presubmit before we accept https://github.com/istio/tools/pull/1678

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [X] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [X] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #34912, opened 2021-08-28
go 1.17 compatibility

jacob-delgado:
go 1.17 no longer supports fallback from h2 to http1.1 when using ALPN. See https://github.com/golang/go/issues/46310

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [X] Test and Release
- [ ] User Experience

- [ X Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

### DIFF
--- a/pkg/test/echo/server/forwarder/protocol.go
+++ b/pkg/test/echo/server/forwarder/protocol.go
@@ -197,7 +197,7 @@ func newProtocol(cfg Config) (protocol, error) {
 			}
 		} else if cfg.Request.Http2 && scheme.Instance(urlScheme) == scheme.HTTPS {
 			if cfg.Request.Alpn == nil {
-				tlsConfig.NextProtos = []string{"h2"}
+				tlsConfig.NextProtos = []string{"h2", "http/1.1"}
 			}
 			proto.client.Transport = &http2.Transport{
 				TLSClientConfig: tlsConfig,


jacob-delgado:
Depends on https://github.com/istio/istio/pull/34728 to pass all CI/CD

pkg/test/echo/server/forwarder/protocol.go
```diff
@@ -197,7 +197,7 @@ func newProtocol(cfg Config) (protocol, error) {
 			}
 		} else if cfg.Request.Http2 && scheme.Instance(urlScheme) == scheme.HTTPS {
 			if cfg.Request.Alpn == nil {
-				tlsConfig.NextProtos = []string{"h2"}
+				tlsConfig.NextProtos = []string{"h2", "http/1.1"}
```
howardjohn, line 200: > As an exception, when the value "h2" is included in the server's Config.NextProtos, HTTP/1.1 clients will be allowed to connect as if they didn't support ALPN. See issue 46310 for more information.

I thought this is not needed here? And the problem is that `istio-peer-exchange` is rejected

PR #34913, opened 2021-08-28
change envoy filter metrics from counter to gauge

ramaraochavali:
We did a trial run of these stats in dev clusters and this being a counter is getting incremented on every push and the skip counter is almost useless as it is being skipped on every listener/filter chain.

So I am proposing to change that stat to gauge and set only if it ever applied once so we can easily track filters that were never applied (which was our usecase)
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/envoyfilter/monitoring.go
+++ b/pilot/pkg/networking/core/v1alpha3/envoyfilter/monitoring.go
@@ -14,6 +14,8 @@
 package envoyfilter
 
 import (
+	"sync"
+
 	"istio.io/istio/pilot/pkg/features"
 	"istio.io/pkg/monitoring"
 )
@@ -23,6 +25,7 @@ type Result string
 const (
 	Error   Result = "error"
 	Applied Result = "applied"
+	Skipped Result = "skipped"
 )
 
 type PatchType string
@@ -51,9 +54,16 @@ var (
 	)
 )
 
+var (
+	envoyFilterStatusMap map[string]map[string]bool // Map of Envoy filter name, patch and status.
+	envoyFilterMutex     sync.RWMutex
+)
+
 func init() {
 	if features.EnableEnvoyFilterMetrics {
 		monitoring.MustRegister(envoyFilterStatus)
+		envoyFilterStatusMap = make(map[string]map[string]bool)
+		envoyFilterMutex = sync.RWMutex{}
 	}
 }
 
@@ -62,10 +72,13 @@ func IncrementEnvoyFilterMetric(name string, pt PatchType, applied bool) {
 	if !features.EnableEnvoyFilterMetrics {
 		return
 	}
-	// Only set the Gauge when the filter is atleast applied once.
+	envoyFilterMutex.Lock()
+	defer envoyFilterMutex.Unlock()
+	if _, exists := envoyFilterStatusMap[name]; !exists {
+		envoyFilterStatusMap[name] = make(map[string]bool)
+	}
 	if applied {
-		envoyFilterStatus.With(nameType.Value(name)).With(patchType.Value(string(pt))).
-			With(resultType.Value(string(Applied))).Record(1)
+		envoyFilterStatusMap[name][string(pt)] = true
 	}
 }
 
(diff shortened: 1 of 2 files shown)


adiprerepa (commented):
lgtm but yielding

pilot/pkg/networking/core/v1alpha3/envoyfilter/monitoring.go
```diff
@@ -63,30 +62,17 @@ func IncrementEnvoyFilterMetric(name string, pt PatchType, applied bool) {
 	if !features.EnableEnvoyFilterMetrics {
 		return
 	}
-	result := Applied
-	if !applied {
-		result = Skipped
+	// Only set the Gauge when the filter is atleast applied once.
+	if applied {
+		envoyFilterStatus.With(nameType.Value(name)).With(patchType.Value(string(pt))).
```
howardjohn, line 67: don't you need to set it to 0 when its not applied?

howardjohn (commented):
IMO this is an abuse of prometheus and violates their metrics guidelines. It would be better suited for a debug endpoint.

hzxuzhonghu:
Donot we support crd status write into k8s?

pilot/pkg/networking/core/v1alpha3/envoyfilter/monitoring.go
ramaraochavali: So we want to set it to only 1 if is every applied if not it would be zero by default right? That was my assumption.

howardjohn:
Just because it's applied once doesn't mean it will not later not apply due
to some changes, so I feel like we would need to be able to "undo" setting
it

ramaraochavali:
> IMO this is an abuse of prometheus and violates their metrics guidelines. It would be better suited for a debug endpoint.

Very fair point. I was going back and forth on whether to support this via metric or via debug endpoint. The main reason for metric is for service owners to write tests and monitor if their filter is applied correctly.  Another option could be to add proxyID as another dimension - but counter seems not right thing to do - it gets incremented every time is applied (on push context) and skipped gets incremented every time it is skipped for every listener/filter - so it is also not useful.

But definitely open for other suggestions

ramaraochavali:
> Just because it's applied once doesn't mean it will not later not apply due
to some changes, so I feel like we would need to be able to "undo" setting
it

I see what you mean. Let me think about it.

ramaraochavali:
> Donot we support crd status write into k8s?

I do not know if that helps - CRD status only validates schema and it is difficult to with the current model. But we can maintain a map of filter and status (applied or not) and can add it to status?

ramaraochavali:
@howardjohn changed it to compute these stats and update on every push. PTAL

ramaraochavali:
@howardjohn @hzxuzhonghu can you please see if this is an acceptable solution?

pilot/pkg/networking/core/v1alpha3/envoyfilter/monitoring.go
```diff
@@ -45,16 +47,23 @@ var (
...
+	envoyFilterStatusMap map[string]map[string]bool // Map of Envoy filter name, patch and status.
+	envoyFilterMutex     sync.RWMutex
+)
+
 func init() {
 	if features.EnableEnvoyFilterMetrics {
-		monitoring.MustRegister(totalEnvoyFilters)
+		monitoring.MustRegister(envoyFilterStatus)
+		envoyFilterStatusMap = make(map[string]map[string]bool)
+		envoyFilterMutex = sync.RWMutex{}
```
howardjohn, line 66: nit: zero value doesn't need to be explicitly set

pilot/pkg/networking/core/v1alpha3/envoyfilter/monitoring.go
```diff
@@ -63,30 +72,39 @@ func IncrementEnvoyFilterMetric(name string, pt PatchType, applied bool) {
...
 	}
-	totalEnvoyFilters.With(patchType.Value(string(pt))).With(resultType.Value(string(Error))).Record(1)
+	envoyFilterStatus.With(patchType.Value(string(pt))).With(resultType.Value(string(Error))).Record(1)
 }
 
-// RecordEnvoyFilterMetric increments the filter metric with the given value.
-func RecordEnvoyFilterMetric(name string, pt PatchType, success bool, value float64) {
-	if !features.EnableEnvoyFilterMetrics {
+func RecordMetrics() {
+	if features.EnableEnvoyFilterMetrics {
```
howardjohn, line 94:
```suggestion
	if !features.EnableEnvoyFilterMetrics {
```
?

pilot/pkg/networking/core/v1alpha3/envoyfilter/monitoring.go
```diff
@@ -63,30 +72,39 @@ func IncrementEnvoyFilterMetric(name string, pt PatchType, applied bool) {
...
-		result = Skipped
+	envoyFilterMutex.RLock()
+	defer envoyFilterMutex.RUnlock()
+	for name, pmap := range envoyFilterStatusMap {
+		for pt, applied := range pmap {
+			result := Applied
+			if !applied {
+				result = Skipped
+			}
+			envoyFilterStatus.With(nameType.Value(name)).With(patchType.Value(pt)).
```
howardjohn, line 105: I still don't see where we ever record 0 if we remove it?

ramaraochavali: I thought setting Skipped with 1 covers that case. But it does not make sense - I will set it to 0 with Applied result.

ramaraochavali:
@howardjohn addressed comments. PTAL.

ramaraochavali:
@howardjohn can you please take a look when you get chance?

howardjohn (approved):

PR #34915, opened 2021-08-29
Make chiron CSR submission return real error.

bianpengyuan:
Current logic just returns and logs `unable to submit csr`. This change makes it return and log the real error from k8s server.

### DIFF
--- a/security/pkg/k8s/chiron/utils.go
+++ b/security/pkg/k8s/chiron/utils.go
@@ -193,7 +193,7 @@ func submitCSR(clientset clientset.Interface,
 	csrData []byte, signerName string,
 	usages []certv1.KeyUsage, numRetries int, requestedLifetime time.Duration) (string, *certv1.CertificateSigningRequest,
 	*certv1beta1.CertificateSigningRequest, error) {
-	var err error = fmt.Errorf("unable to submit csr")
+	var lastErr error
 	var useV1 bool = true
 	var csrName string = ""
 	for i := 0; i < numRetries; i++ {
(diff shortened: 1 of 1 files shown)


security/pkg/k8s/chiron/utils.go
```diff
@@ -252,12 +254,14 @@ func submitCSR(clientset clientset.Interface,
...
-		} else if apierrors.IsAlreadyExists(err) {
+		}
+		lastErr = err
+		if apierrors.IsAlreadyExists(err) {
 			csrName = ""
 		}
 	}
-	log.Errorf("retry attempts exceeded when creating csr request %v: %v", csrName, err)
-	return "", nil, nil, err
+	log.Errorf("retry attempts exceeded when creating csr request %v", csrName)
```
lei-tang, line 263: The original log statement outputs err in Errorf. Why does the new log statement not output the lastErr in Errof?

bianpengyuan: It will eventually be logged at https://github.com/istio/istio/blob/23c56d13515f8c415bc914082a88e4c296d4efde/pilot/cmd/pilot-discovery/app/cmd.go#L94, e.g. this the log lines if I modify the code in this function to return a dummy error:

```
2021-08-30T22:37:03.612350Z     info    initializing Istiod DNS certificates host: istiod.istio-system.svc, custom host: 
2021-08-30T22:37:03.612410Z     info    Generating K8S-signed cert for [istiod.istio-system.svc istiod-remote.istio-system.svc istio-pilot.istio-system.svc]
2021-08-30T22:37:03.906321Z     error   failed to create discovery service: failed generating key and cert by kubernetes: unable to submit CSR request (). Error: dummy error
Error: failed to create discovery service: failed generating key and cert by kubernetes: unable to submit CSR request (). Error: dummy error
```

lei-tang (approved):

security/pkg/k8s/chiron/utils.go
lei-tang: Thanks! Approved.

Issue #34932, opened 2021-08-30
https://github.com/istio/istio/pull/33773 causes failures in istio.io doc tests


ericvn:
### Bug Description

If istio.io doc tests pulled an Istio build with the https://github.com/istio/istio/pull/33773, several of the tests fail. See https://github.com/istio/istio.io/pull/10208.

Looking at the PR, it seems we may need some Doc updates, or we may need some additional code changes.

### Version

```prose
Commits including and after https://github.com/istio/istio/pull/33773.
```

### Affected product area

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [ ] Upgrade
- [ ] Multi Cluster
- [ ] Virtual Machine
- [X] Control Plane Revisions

### Is this the right place to submit this?

- [X] This is not a security vulnerability
- [X] This is not a question about how to use Istio

PR #34939, opened 2021-08-30
Remove previous default tag on non-revisioned installation

Monkeyanator:
Part of the fix for #34932 

Installing Istio without a revision should be considered as installing a default revision, and therefore we should remove the previous default tag on install to avoid mutating webhook overlap issues (otherwise analyzer will detect duplicate between default injector and default tag and fail out).

Noticed another issue when looking into this: since we rely on the default tag to provide resource validation now, and the in-cluster operator does not use this, we have no validation when installing with in-cluster operator currently. Need to address this before 1.12 release (either document to use `istioctl tag set default` or do it in `istioctl operator init` similar to how we autogenerate tag for `istioctl install`).

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/operator/cmd/mesh/install.go
+++ b/operator/cmd/mesh/install.go
@@ -184,12 +184,16 @@ func runApplyCmd(cmd *cobra.Command, rootArgs *rootArgs, iArgs *installArgs, log
 
 	// Detect whether previous installation exists prior to performing the installation.
 	exists := revtag.PreviousInstallExists(context.Background(), kubeClient)
+	pilotEnabled := iop.Spec.Components.Pilot != nil && iop.Spec.Components.Pilot.Enabled.Value
+	rev := iop.Spec.Revision
+	if rev == "" && pilotEnabled {
+		_ = revtag.DeleteTagWebhooks(context.Background(), kubeClient, revtag.DefaultRevisionName)
+	}
 	iop, err = InstallManifests(iop, iArgs.force, rootArgs.dryRun, restConfig, client, iArgs.readinessTimeout, l)
 	if err != nil {
 		return fmt.Errorf("failed to install manifests: %v", err)
 	}
 
-	rev := iop.Spec.Revision
 	if !exists || rev == "" {
 		cmd.Println("Making this installation the default for injection and validation.")
 		if rev == "" {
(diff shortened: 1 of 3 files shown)


operator/cmd/mesh/install.go
```diff
@@ -184,12 +184,15 @@ func runApplyCmd(cmd *cobra.Command, rootArgs *rootArgs, iArgs *installArgs, log
 
 	// Detect whether previous installation exists prior to performing the installation.
 	exists := revtag.PreviousInstallExists(context.Background(), kubeClient)
+	rev := iop.Spec.Revision
+	if rev == "" {
+		_ = revtag.DeleteTagWebhooks(context.Background(), kubeClient, revtag.DefaultRevisionName)
```
howardjohn, line 189: This introduces a (likely small) downtime that may lead to a pod not being injected. I think this is something that is never acceptable, even for a small time, as failiure to inject a pod can lead to authz policy bypass

howardjohn (changes requested):
Overwrite sounds fine, but  delete+create is suspicious

operator/cmd/mesh/install.go
Monkeyanator: Hm there is already a small window of injection downtime (between calling `DeactivateIstioInjectionWebhook` and creating the generated revision tag).... will have to rework this too.

howardjohn: ok, I think that is worth doing but we can merge this in the mean time if desired as its not making the situation any worse I guess

Monkeyanator: Yeah this is meant to unblock docs test update, I think we should either merge or revert the breaking PR

Monkeyanator: Bit of a brain dump but another thing we're trying to avoid here is having a window of double-injection by different revisions.  For example, default tag points to revision `foo`, we install without a revision so then we also have the istio-sidecar-injector webhook injecting for default labels until at the end of the installation we deactivate that webhook.

Maybe this is taping over the fact that there shouldn't be an injector for `istio-injection=enabled` installed through the `istio-discovery` chart anymore now that we control this through default tag... would fix the issue of in-cluster operator reconciling over the deactivated webhook too

howardjohn (approved):

Monkeyanator:
cc @istio/wg-user-experience-maintainers for review

shamsher31 (approved):

Issue #34949, opened 2021-08-31
delta: EDS and SDS wrongfully remove resources


adiprerepa:
With EDS and SDS, we will have PushRequest.Full = false, which is technically an "incremental" update, so we send a few resources.  However, delta takes this to be the SOTW with EDS and SDS, and removes the rest of the resources.

howardjohn:
Ah... the reason it doesn't break the world is envoy just ignores the removal.

For RDS it will give an error log, for SDS/EDS it is silent

PR #34950, opened 2021-08-31
delta: eds/sds removedresources fix

adiprerepa:
for https://github.com/istio/istio/issues/34949

This makes it so that for EDS and SDS, when an incremental update is sent, we don't assume it isn't by checking PushRequest.Full.

cc @howardjohn

### DIFF
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -379,16 +379,17 @@ func edsNeedsPush(updates model.XdsUpdates) bool {
 	return false
 }
 
-func (eds *EdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *model.WatchedResource,
-	req *model.PushRequest) (model.Resources, model.XdsLogDetails, error) {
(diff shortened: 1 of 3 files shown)


pilot/pkg/xds/eds.go
```diff
@@ -425,10 +427,16 @@ func (eds *EdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
 			eds.Server.Cache.Add(builder, req, resource)
 		}
 	}
-	return resources, model.XdsLogDetails{
+	return resources, removed, model.XdsLogDetails{
```
howardjohn, line 430: The behavior here is pretty subtle. I think some comments here would help a lot, to explain why we have Delta for EDS/SDS despite them not actually being "delta"

howardjohn (commented):
lgtm but want to get clarification on https://envoyproxy.slack.com/archives/CEFDKQ3RQ/p1630424884013900 first

howardjohn:
not stale

PR #34955, opened 2021-08-31
Make priorityClass of cni in helm chart config

venkatk-25:
The priorityClassName is hardcoded in helm charts of cni. We need it to be configurable to deploy in our custom enviroment.

### DIFF
--- a/manifests/charts/istio-cni/templates/daemonset.yaml
+++ b/manifests/charts/istio-cni/templates/daemonset.yaml
@@ -52,7 +52,7 @@ spec:
           operator: Exists
         - effect: NoExecute
           operator: Exists
-      priorityClassName: system-node-critical
+      priorityClassName: {{ default 'system-node-critical' .Values.priorityClassName }}
       serviceAccountName: istio-cni
       # Minimize downtime during a rolling upgrade or deletion; tell Kubernetes to do a "force
       # deletion": https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.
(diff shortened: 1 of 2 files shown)


bianpengyuan (commented):
Could you explain a bit why `system-node-critical` does not work within your environment?

venkatk-25:
> Could you explain a bit why `system-node-critical` does not work within your environment?

We have calico as cni daemonset in on premise cluster and want to run istio-cni as chained cni to that. 
We want to make sure calico-cni gets scheduled first and istio-cni after that on addition of new nodes to cluster.
So we are planning to use `system-cluster-critical` to ensure slightly lesser priority

bianpengyuan:
@venkatk-25 istio-cni should block and wait for main CNI plugin to be ready. Is that not the case? Did you observe some problem? If yes we need to fix that instead of relying on priority class.

venkatk-25:
> @venkatk-25 istio-cni should block and wait for main CNI plugin to be ready. Is that not the case? Did you observe some problem? If yes we need to fix that instead of relying on priority class.

I didn't know it will wait, our internal team has raised it as a race condition for adoption. Can you point to related docs/code

bianpengyuan:
@venkatk-25 Wait logic is around https://github.com/istio/istio/blob/8d40d0fd16a9c8401f607eab61d6116a76c43e31/cni/pkg/install/cniconfig.go#L195.

venkatk-25:
> @venkatk-25 Wait logic is around
> 
> https://github.com/istio/istio/blob/8d40d0fd16a9c8401f607eab61d6116a76c43e31/cni/pkg/install/cniconfig.go#L195
> 
> .

@bianpengyuan thank you

PR #34956, opened 2021-08-31
add support for specifying default host in http10

ramaraochavali:
Host Header is optional in Http1.0 and Envoy recommends setting [DefaultHttpHostHeader ](https://github.com/envoyproxy/envoy/blob/3a4baf64a5c300739b7ce47128c1e62fd2a38f33/api/envoy/config/core/v3/protocol.proto#L215) when Http1.0 is enabled. We might need this feature for migrating our legacy customers. This PR adds that config support via Metadata.
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -582,6 +582,10 @@ type NodeMetadata struct {
 	// Alpha in 1.1, based on feedback may be turned into an API or change. Set to "1" to enable.
 	HTTP10 string `json:"HTTP10,omitempty"`
 
+	// DefaultHTTP10Host indicates the default host to be be used when HTTP/1.0 is enabled and request does
+	// not have host header as host header is not mandatory for HTTP/1.0.
+	DefaultHTTP10Host string `json:"DEFAULT_HTTP10_HOST,omitempty"`
+
 	// Generator indicates the client wants to use a custom Generator plugin.
 	Generator string `json:"GENERATOR,omitempty"`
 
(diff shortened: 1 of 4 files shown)


pilot/pkg/model/context.go
```diff
@@ -582,6 +582,10 @@ type NodeMetadata struct {
 	// Alpha in 1.1, based on feedback may be turned into an API or change. Set to "1" to enable.
 	HTTP10 string `json:"HTTP10,omitempty"`
 
+	// DefaultHTTP10Host indicates the default host to be be used when HTTP/1.0 is enabled and request does
+	// not have host header as host header is not mandatory for HTTP/1.0.
+	DefaultHTTP10Host string `json:"DEFAULT_HTTP10_HOST,omitempty"`
```
hzxuzhonghu, line 587: Maybe i am not understanding the use case, Why isn't the host header provided by clients if they want it?

ramaraochavali: Because it is not mandated by Http1.0 and legacy clients might not have set it but Envoy needs it without that it rejects the http1.0 requests. Now we can not go back and change all legacy clients to pass host header when they are coming on to mesh. That is why Envoy added this support to specify default host header.

howardjohn: What would you even set as the default host header?

hzxuzhonghu: I think it is more reasonable to file a fix in envoy to allow http 1.0 without host header.

ramaraochavali: I do not think that is easy in Envoy - because Envoy upgrading it to http1.1 for upstream is the default flow. It does not support end -> end http1.0.

So if we do not want to move forward with this in Istio, we can do it via Envoy filter at our side. 

Just pushed this PR thinking that this might be generally useful. If you guys think, it is not I can close this PR.

hzxuzhonghu: From my point, this is not common.

ramaraochavali: Ok. Fair enough. I will close this.

Issue #29482, opened 2020-12-08
Remote pilot address is bind to port 15012 (for xDs). It would be great if we can override this port, making it able to support NodePort for remode istiod.


12beseuahmad:
**Describe the feature request**
I am following this guide for primary remote on different network.

https://istio.io/latest/docs/setup/install/multicluster/primary-remote_multi-network/

Instead of using the load balancer on primary to expose the istiod, I am trying to use NodePort. Also verified it using netcat and it works. Now on remote machine when I configure using IstioOperator there is an option for remotePilotAddress but no option for the port i.e it get binded to default 15012.

**Describe alternatives you've considered**

[ ] Docs
[x] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

**Additional context**

wojtek-viirtue:
Looks like this is also related to: https://github.com/istio/istio/issues/17801

stevenctl:
not stale

PR #34960, opened 2021-08-31
override port for remotePilotAddress to support NodePort

stevenctl:
untested but possible solution to access the control plane via node-port (https://github.com/istio/istio/issues/29482)


manifests/charts/base/templates/endpoints.yaml
```diff
@@ -9,7 +9,7 @@ subsets:
 - addresses:
   - ip: {{ .Values.global.remotePilotAddress }}
   ports:
-  - port: 15012
+  - port: {{ .Values.global.remotePilotPort | default "15012" }}
```
morvencao, line 12: why add quote to port number?

Issue #34944, opened 2021-08-31
Destination Rule Subsets don't update LDS for AUTO_PASSTHROUGH


EItanya:
### Bug Description

After the behavior of AUTH_PASSTHROUGH was changed recently as a result of the following [CVE](https://istio.io/latest/news/security/istio-security-2021-006/), the routing behavior of AUTO_PASSTHROUGH is now based on `pilot` generating al of the listeners, rather than using a `tcp_cluster_rewrite`. However, these listeners do not update when `DestinationRules` because of the [following code](https://github.com/istio/istio/blob/4a77b9817fa84306cbbfe8e3fd7f07bf8e200acc/pilot/pkg/xds/lds.go#L32-L37). This means that adding subsets to DestinationRules will not update the listener config, and require changes to other resources, or a restart of `istiod` to start working.

After I created a build commenting out the `DestinationRule` skip it started to work properly.

### Version

```prose
istioctl version
client version: 1.10.4
control plane version: 1.10.2
data plane version: 1.10.2 (9 proxies)
```

hzxuzhonghu:
@EItanya  Can you elaborate a little bit more?

EItanya:
Sure, when a gateway is configured with [`AUTH_PASSTHROUGH`](https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings-TLSmode) it now creates individual filter chains for each reachable service, for example:
```
{
         "filter_chain_match": {
          "application_protocols": [
           "istio",
           "istio-peer-exchange",
           "istio-http/1.0",
           "istio-http/1.1",
           "istio-h2"
          ],
          "server_names": [
           "outbound_.9080_._.reviews.bookinfo.svc.remote-cluster.soloio"
          ]
         },
         "filters": [
          {
           "name": "istio.stats",
           "typed_config": {
            "@type": "type.googleapis.com/udpa.type.v1.TypedStruct",
            "type_url": "type.googleapis.com/envoy.extensions.filters.network.wasm.v3.Wasm",
            "value": {
             "config": {
              "root_id": "stats_outbound",
              "vm_config": {
               "vm_id": "tcp_stats_outbound",
               "runtime": "envoy.wasm.runtime.null",
               "code": {
                "local": {
                 "inline_string": "envoy.wasm.stats"
                }
               }
              },
              "configuration": {
               "@type": "type.googleapis.com/google.protobuf.StringValue",
[...]

PR #34963, opened 2021-08-31
DestinationRule triggers lds for gateways

stevenctl:
Partial fix https://github.com/istio/istio/issues/34944 although this only affects a small number of cases. We should still have a way to optimize when there is no autopassthrough. 

This is better than config not being reflected though.


howardjohn:
didn't have a chance to read the code, we can make it only for gateways
though if we didn't already

howardjohn (approved):

EItanya:
Will this be back ported as well? This is broken on all versions with this CVE fix.

PR #34971, opened 2021-08-31
use EndpointSlices by default in 1.21

stevenctl:
Not sure if we have other examples changing default behavior conditionally on the k8s version, this seems like it would wor.

### DIFF
--- a/tests/integration/pilot/endpointslice/endpointslice_test.go
+++ b/tests/integration/pilot/endpointslice/endpointslice_test.go
@@ -18,6 +18,8 @@
 package pilot
 
 import (
+	"fmt"
+	kubelib "istio.io/istio/pkg/kube"
 	"testing"
 
 	"istio.io/istio/pkg/test/framework"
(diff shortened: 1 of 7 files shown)


stevenctl:
Should this have a release note? I would think it's sort of an implementation detail of pilot.

tests/integration/pilot/endpointslice/endpointslice_test.go
```diff
@@ -41,11 +43,13 @@ func TestMain(m *testing.M) {
...
 		Setup(istio.Setup(&i, func(t resource.Context, cfg *istio.Config) {
-			cfg.ControlPlaneValues = `
+			cfg.ControlPlaneValues = fmt.Sprintf(`
 values:
   pilot:
     env:
-      PILOT_USE_ENDPOINT_SLICE: "true"`
+      PILOT_USE_ENDPOINT_SLICE: "true"`,
+      			// for k8s 1.21+, this suite should test disabling EndpointSlice mode
+				kubelib.IsLessThanVersion(t.Clusters().Kube().Default(), 21))
```
howardjohn, line 52: the sprintf doesn't seem to have any `%s`?

howardjohn:
Blocker: https://github.com/istio/istio/issues/34995

howardjohn:
> Should this have a release note? I would think it's sort of an implementation detail of pilot.

IMO yes. It is an implementation detail but a big enough one users should know about it for debugging, etc.

Also note users may manually create EndpointSlices that are currently ignored by Istiod, which will now not be ignored. So it can be user facing in edge cases

hzxuzhonghu (approved):

howardjohn (changes requested):
This LGTM but would prefer https://github.com/istio/istio/issues/34995 to be fixed first. While its a relatively rare edge case, the impact is extremely bad and it is a regression

stevenctl:
With #35990 do we want to squeeze this into 1.12?

howardjohn:
This seems like a pretty large change to get this late into the release. I
think its ideal to get it in now for master so we get 4 months of soak time
before shipping?

ramaraochavali:
+1 for master

howardjohn (approved):

PR #34974, opened 2021-09-01
support routing based on jwt claims

yangminzhu:
Adds support of the special header `x-jwt-claims` in virtual service for routing based on JWT claims.

This must be used with RequestAuthentication otherwise the routing will never match (covered in e2e tests), see more details in design doc: https://docs.google.com/document/d/1s8Lykf8-KUZTLiG4YMSCHvxBDwZzHlrrHyZ7cJymc_8/edit#heading=h.6p29w9kcduab

Note: will add release notes to make this feature in a follow PR

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [x] Security
- [ ] Test and Release
- [ ] User Experience


yangminzhu:
@howardjohn @costinm @mandarjog PTAL, this is ready for review. Note I will have some follow-up PRs for things like code refactor of the authn filter go code generation, the release notes and probably more tests, thanks.

There also seems quite some tests fail due to the Envoy update (see https://github.com/istio/istio/pull/34188 for the failed envoy update) but the newly added tests for jwt route matching should pass.

istio.deps
```diff
@@ -4,6 +4,6 @@
     "name": "PROXY_REPO_SHA",
     "repoName": "proxy",
     "file": "",
-    "lastStableSHA": "74393cf764c167b545f32eb895314e624186e5b6"
+    "lastStableSHA": "b6e2543232ea486685292a07f8ec68775c0f6b7d"
```
howardjohn, line 7: Ideally we split this out if possible before merging. Maybe even the g-c-p part

yangminzhu: sure, I can split it out in separate PRs (including the go-control-plane part). There seems some test failures due to the Envoy update though. Could you review other part in the PR, we can hold the merge until the other PRs are merged. thanks.

yangminzhu: added the hold label.

yangminzhu:
@howardjohn @costinm could you review the PR? thanks.

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -859,6 +871,53 @@ func isCatchAllHeaderMatch(in *networking.StringMatch) bool {
 	return catchall
 }
 
+// translateMetadataMatch translates a header match to dynamic metadata matcher. Returns nil if the header is not supported
+// or the header format is invalid for generating metadata matcher.
+//
+// The currently only supported header is x-jwt-claim for JWT claims matching. Nested claims is also supported using
+// the separator "." or "-" between claim names. For example,
```
howardjohn, line 878: This is not ideal if if we can avoid it... is there any character that is a valid HTTP header but not a jwt claim? I am guessing no because claim can be any json key?

I feel like:
* Have 1 separator handles 99% of cases
* Have 2 separators handles 99.5% of cases but doubles the complexity

For example, if my claim is `"foo-bar.com"` it would not work.

Maybe an alternative is to have users URL encode any `.`s?

howardjohn (commented):
lgtm, one small comment

pilot/pkg/networking/core/v1alpha3/route/route.go
yangminzhu: > This is not ideal if if we can avoid it... is there any character that is a valid HTTP header but not a jwt claim? I am guessing no because claim can be any json key?

yeah, I don't think there would be any character that is valid in HTTP but not in jwt, it's actually the other way, you can technically put anything in the jwt claim (but practically it's mostly only ASCII characters).


> I feel like:
> 
> * Have 1 separator handles 99% of cases
> * Have 2 separators handles 99.5% of cases but doubles the complexity
> 
> For example, if my claim is `"foo-bar.com"` it would not work.
> 
> Maybe an alternative is to have users URL encode any `.`s?

I think `.` and `-` addresses the most common use cases (99.5%), Is the complexity mainly for user to understand there are 2 options for the separator? (it does not add too much complexity in implementation). I'm not sure if supporting URL encode in header is a good idea, AFAIK the header is not supposed to be url encoded.

The other alternative is probably just to support **all** valid header characters (except ALPHA and DIGIT, see the tchar definition in https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6) as the separator, this is a bit similar to the multiple delimiter usage in the linux command `sed`.

@costinm @mandarjog any thoughts on this?

yangminzhu: updated to use `.` as separator.

PR #34978, opened 2021-09-01
rm PILOT_USE_TARGET_PORT_FOR_GATEWAY_ROUTES

hzxuzhonghu:

According to the comments, it can be removed in 1.12

### DIFF
--- a/pilot/pkg/model/gateway.go
+++ b/pilot/pkg/model/gateway.go
@@ -351,12 +351,9 @@ func udpSupportedPort(number uint32, instances []*ServiceInstance) bool {
 // port and translating to the targetPort in addition to just directly referencing a port. In this
 // case, we just make a best effort guess by picking the first match.
 func resolvePorts(number uint32, instances []*ServiceInstance, legacyGatewaySelector bool) []uint32 {
-	if !features.UseTargetPortForGatewayRoutes {
-		return []uint32{number}
-	}
 	ports := map[uint32]struct{}{}
 	for _, w := range instances {
-		if _, directPortTranslation := w.Service.Attributes.Labels[DisableGatewayPortTranslationLabel]; directPortTranslation && legacyGatewaySelector {
+		if _, disablePortTranslation := w.Service.Attributes.Labels[DisableGatewayPortTranslationLabel]; disablePortTranslation && legacyGatewaySelector {
 			// Skip this Service, they opted out of port translation
 			// This is only done for legacyGatewaySelector, as the new gateway selection mechanism *only* allows
 			// referencing the Service port, and references are un-ambiguous.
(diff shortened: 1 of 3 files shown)


pilot/pkg/model/gateway.go
```diff
@@ -323,6 +319,9 @@ func resolvePorts(number uint32, instances []*ServiceInstance, legacyGatewaySele
 			ports[w.Endpoint.EndpointPort] = struct{}{}
 		}
 	}
+	if len(ports) > 1 {
+		log.Warnf("gateway: found %d services on port %d with different target port: %v", len(ports), number, ports)
```
howardjohn, line 323: why warn?

howardjohn: it's totally valid case to have multiple

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -190,7 +154,7 @@ func (configgen *ConfigGeneratorImpl) buildGatewayListeners(builder *ListenerBui
 		pluginParams := &plugin.InputParams{
 			Node:            builder.node,
 			Push:            builder.push,
-			ServiceInstance: si,
+			ServiceInstance: nil,
 		}
```
howardjohn, line 158: is this valid? If so should we remove SseviceInsrances from plugin Paramus entirely?

pilot/pkg/model/gateway.go
hzxuzhonghu: Copied from https://github.com/istio/istio/pull/34978/files#diff-7abd8e9bc7226623690e85b5acc0280def2d72df7960b1324349b08dcb186ca1L100

Do you mean it is valid below:
2 services for gateway.

1. 80-->8080

2. 80-->8081

pilot/pkg/networking/core/v1alpha3/gateway.go
hzxuzhonghu: InputParams is not used by `OnOutboundListener`, but necessary for `inboundXXX`

If we want to remove this, i think we should define different `InputParams` for them

hzxuzhonghu: I can update it as a separate one, if you feels that

howardjohn: I see. Yeah let's not bundle it with this PR but it may be a good idea. We had bugs in the past where plugin expected certain field but not everywhere set it

pilot/pkg/model/gateway.go
howardjohn: In the old code it was invalid. In the new code it is valid

hzxuzhonghu: Then in this case, which port will be used as listener/route

hzxuzhonghu: removed

howardjohn (approved):

PR #34981, opened 2021-09-01
Extract yaml code to increase function readability

ChenglongGao327:
1. Extract yaml code to increase function readability
2. ProxylessGRPCSvc is only created when t.Clusters().IsMulticluster() is false
https://github.com/istio/istio/blob/8d40d0fd16a9c8401f607eab61d6116a76c43e31/tests/integration/pilot/common/apps.go#L244

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/tests/integration/pilot/common/apps.go
+++ b/tests/integration/pilot/common/apps.go
@@ -87,6 +87,46 @@ const (
 	externalHostname = "fake.external.com"
 )
 
+var sidecarTemplate = `
+apiVersion: networking.istio.io/v1alpha3
+kind: Sidecar
+metadata:
+  name: restrict-to-namespace
+spec:
+  egress:
+  - hosts:
+    - "./*"
+    - "istio-system/*"
+`
+
+var serviceEntryTemplate = `
+apiVersion: networking.istio.io/v1alpha3
+kind: ServiceEntry
+metadata:
+  name: external-service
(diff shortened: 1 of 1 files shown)


tests/integration/pilot/common/apps.go
```diff
@@ -274,58 +314,30 @@ func SetupApps(t resource.Context, i istio.Instance, apps *EchoDeployments) erro
 	apps.StatefulSet = echos.Match(echo.Service(StatefulSetSvc))
 	apps.Naked = echos.Match(echo.Service(NakedSvc))
 	apps.External = echos.Match(echo.Service(ExternalSvc))
-	apps.ProxylessGRPC = echos.Match(echo.Service(ProxylessGRPCSvc))
+	if !t.Clusters().IsMulticluster() {
```
ChenglongGao327, line 317: ProxylessGRPCSvc is only created when t.Clusters().IsMulticluster() is false
istio/tests/integration/pilot/common/apps.go

howardjohn (commented):
IMO this makes it less readable, since now I need to go look try to find the template to understand it. If its not used in multiple places then moving it out doesn't help much really.

PR #34983, opened 2021-09-01
Optimize: reduce calling mtlsChecker.mtlsDisabledByPeerAuthentication

hzxuzhonghu:

make use of `computeForEndpoint` calculated result instead of call mtlsDisabledByPeerAuthentication again

### DIFF
--- a/pilot/pkg/xds/endpoint_builder.go
+++ b/pilot/pkg/xds/endpoint_builder.go
@@ -328,21 +328,21 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
 			if ep.EnvoyEndpoint == nil {
 				ep.EnvoyEndpoint = buildEnvoyLbEndpoint(ep)
 			}
-			if features.EnableAutomTLSCheckPolicies {
-				tlsMode := ep.TLSMode
-				if b.mtlsChecker != nil && b.mtlsChecker.mtlsDisabledByPeerAuthentication(ep) {
-					tlsMode = ""
-				}
-				if nep, modified := util.MaybeApplyTLSModeLabel(ep.EnvoyEndpoint, tlsMode); modified {
-					ep.EnvoyEndpoint = nep
-				}
-			}
 			locLbEps.append(ep, ep.EnvoyEndpoint, ep.TunnelAbility)
 
 			// detect if mTLS is possible for this endpoint, used later during ep filtering
 			// this must be done while converting IstioEndpoints because we still have workload labels
 			if b.mtlsChecker != nil {
 				b.mtlsChecker.computeForEndpoint(ep)
+				if features.EnableAutomTLSCheckPolicies {
+					tlsMode := ep.TLSMode
+					if b.mtlsChecker != nil && b.mtlsChecker.isMtlsDisabled(ep.EnvoyEndpoint) {
+						tlsMode = ""
+					}
+					if nep, modified := util.MaybeApplyTLSModeLabel(ep.EnvoyEndpoint, tlsMode); modified {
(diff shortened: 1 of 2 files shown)


hzxuzhonghu:
@incfly  Do you know what i missed?

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -328,21 +322,21 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
...
 			// this must be done while converting IstioEndpoints because we still have workload labels
 			if b.mtlsChecker != nil {
 				b.mtlsChecker.computeForEndpoint(ep)
+				if features.EnableAutomTLSCheckPolicies {
+					tlsMode := ep.TLSMode
+					if b.mtlsChecker != nil && b.mtlsChecker.isMtlsDisabled(ep.EnvoyEndpoint) {
+						tlsMode = ""
+					}
+					if nep, modified := util.MaybeApplyTLSModeLabel(ep.EnvoyEndpoint, tlsMode); modified {
+						ep.EnvoyEndpoint = nep
```
howardjohn, line 337: A bit confused how this works. Doesn't L325 store the old EnvoyEndpoint, not the updated one now?

incfly: correct. i was about to say same thing. this is probably why the test fails?

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -328,22 +322,21 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
...
-			}
-			locLbEps.append(ep, ep.EnvoyEndpoint, ep.TunnelAbility)
-
 			// detect if mTLS is possible for this endpoint, used later during ep filtering
 			// this must be done while converting IstioEndpoints because we still have workload labels
 			if b.mtlsChecker != nil {
 				b.mtlsChecker.computeForEndpoint(ep)
+				if features.EnableAutomTLSCheckPolicies {
+					tlsMode := ep.TLSMode
+					if b.mtlsChecker != nil && b.mtlsChecker.isMtlsDisabled(ep.EnvoyEndpoint) {
```
incfly, line 331: got it. the idea is that computeForEndpoint will contain a call of mtlsDisabledByPeerAuthentication, so by the line of 331, we don't have to re-calculate again, is that right?

that SG to me.

incfly, line 331: you don't have to check b.mTLSchecker != nil again here right?

incfly: thinking more. to avoid this re-invocation again, now i see all the calling of mTLSDisabledByPeerAuthn is used within computeEndpionts.

If we are putting the mTLSDisabledByPeerAuthn a local func variable within computeEndpiont, other code won't have access to the function right?

to me, it seems we only need look up via isMTLSdisabled.

incfly:
just curious what drives you find this? some real performance measurement of istiod?

pilot/pkg/xds/endpoint_builder.go
hzxuzhonghu: good catch

hzxuzhonghu:
>just curious what drives you find this? some real performance measurement of istiod?

When i investigated xds cache, i found this part seems duplicate.

pilot/pkg/xds/endpoint_builder.go
hzxuzhonghu: sgtm

incfly (approved):

howardjohn (approved):

PR #34986, opened 2021-09-01
add flag to control cluster telemetry metadata

ramaraochavali:
This is not a huge perf impact but with many clusters (especially in large k8s clusters) this is adding some overhead. So adding a flag to control this behaviour. Ideally we can deprecate existing `PILOT_ENDPOINT_TELEMETRY_LABEL` and remove it later. For now I have added a new flag that works for both.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -400,7 +400,7 @@ var (
 			"Setting the timeout to 0 disables this behavior.",
 	).Get()
 
-	EnableTelemetry = env.RegisterBoolVar("PILOT_ENABLE_TELEMETRY", true,
+	EnableTelemetry = env.RegisterBoolVar("PILOT_ENABLE_TELEMETRY_LABEL", true,
 		"If true, pilot will add telemetry related metadata to cluster and endpoint resources, which will be consumed by telemetry filter.",
 	).Get()
 
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -400,6 +400,10 @@ var (
 			"Setting the timeout to 0 disables this behavior.",
 	).Get()
 
+	EnableTelemetry = env.RegisterBoolVar("PILOT_ENABLE_TELEMETRY", true,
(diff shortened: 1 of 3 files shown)


pilot/pkg/features/pilot.go
```diff
@@ -400,6 +400,10 @@ var (
 			"Setting the timeout to 0 disables this behavior.",
 	).Get()
 
+	EnableTelemetry = env.RegisterBoolVar("PILOT_ENABLE_TELEMETRY", true,
```
bianpengyuan, line 403: Maybe `ENABLE_TELEMETRY_LABEL`? `ENABLE_TELEMETRY` sound ambiguous to me.

bianpengyuan (commented):
Disable cluster label will make your server side metrics missing destination service related information. Are you sure you would like to disable this?

ramaraochavali:
> Disable cluster label will make your server side metrics missing destination service related information. Are you sure you would like to disable this?

Yes. We do not use telemetry at all. We use Envoy stats for our telemetry

pilot/pkg/features/pilot.go
ramaraochavali: Ok. Makes sense. Changed.

pilot/pkg/features/pilot.go
```diff
@@ -400,7 +400,7 @@ var (
 			"Setting the timeout to 0 disables this behavior.",
 	).Get()
 
-	EnableTelemetry = env.RegisterBoolVar("PILOT_ENABLE_TELEMETRY", true,
+	EnableTelemetry = env.RegisterBoolVar("PILOT_ENABLE_TELEMETRY_LABEL", true,
```
bianpengyuan, line 403: Let's change the variable name as well.

ramaraochavali: sorry. Done now.

bianpengyuan (approved):

ramaraochavali:
@hzxuzhonghu @howardjohn can you PTAL?

howardjohn (approved):

PR #34993, opened 2021-09-01
Bump up default grpc transport settings for adsc.

SpecialYang:
Our mcp resources server has lots of services with huge number of service instances. When we use mcp-over-xds protocol to get service entry resources by adsc client, we got the following error.

```
2021-09-01T20:13:39.787444Z info    adsc    Connection closed for node sidecar~22.1.72.106~test-
1.default~default.svc.cluster.local with err: rpc error: 
code = ResourceExhausted desc = grpc: received message larger than max (19542550 vs. 4194304)
```

The default received message size of grpc is 4mb which is really not enough for huge services. So I think we can simply set it to math.MaxInt32 which does same with `XdsProxy`.
https://github.com/istio/istio/blob/dceadfc91e5d7aafefd12cac87c7ffc27858dc9b/pkg/istio-agent/xds_proxy.go#L630-L655

### DIFF
--- a/pkg/adsc/adsc.go
+++ b/pkg/adsc/adsc.go
@@ -20,6 +20,7 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
+	"math"
 	"net"
 	"os"
 	"sort"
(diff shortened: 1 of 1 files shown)


hzxuzhonghu (approved):
lgtm, I am thinking maybe we can extract all these settings later

PR #35000, opened 2021-09-02
Allow specifying an IP:port for istiod address.

costinm:
For VM and other non-k8s environments, allow specifying an IP:port address without changing /etc/hosts

The DiscoveryAddress was also used to set the ServerName, for cert validation - typically istiod-REV.istio-system.svc - and gets
resolved in k8s to the cluster IP, and on VMs was using /etc/hosts. 

After this change, the DiscoveryAddress can be an IP:port - either the pilot endpoint if it is stable, or the east/west gateway IP -
without having to touch /etc/hosts. The Istiod cert will be validated with the custom SAN.

For VMs and similar environments it is the IP of the east-west gateway or a similar load balancer. Current instructions
modify /etc/hosts to map the name to IP, but sometimes it is not possible or convenient to change /etc/hosts. 

This is for internal use, for tools that setup/launch Istiod. 

Change-Id: Ibcc1f5a6bd4235d21f4e278d1197749ed6463922


pkg/istio-agent/xds_proxy.go
```diff
@@ -142,6 +143,7 @@ func initXdsProxy(ia *Agent) (*XdsProxy, error) {
 	}
 	proxy := &XdsProxy{
 		istiodAddress:         ia.proxyConfig.DiscoveryAddress,
+		istiodAddressOverride: ia.cfg.IstiodAddrOverride,
```
yangminzhu, line 146: why not just override the existing `istiodAddress`?

pkg/istio-agent/xds_proxy.go
```diff
@@ -367,7 +369,11 @@ func (p *XdsProxy) handleStream(downstream adsStream) error {
 
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
-	upstreamConn, err := grpc.DialContext(ctx, p.istiodAddress, p.istiodDialOptions...)
+	upstreamAddr := p.istiodAddress
+	if p.istiodAddressOverride != "" {
+		upstreamAddr = p.istiodAddressOverride
+	}
+	upstreamConn, err := grpc.DialContext(ctx, upstreamAddr, p.istiodDialOptions...)
```
yangminzhu, line 376: the proxyLog.Errorf() below is still using the `p.istiodAddress`

stevenctl:
Why can't we just use proxy config?

I suppose we should also document that this is an option if you can access pilot directly and don't need to go through a gateway. 

Do we run into any TLS issue due to using the IP instead of a valid name?

pkg/istio-agent/xds_proxy.go
costinm: Will fix.

costinm:
Re ProxyConfig - it's not supposed to be user-configurable. Some launcher or installer will get the endpoint IP ( or the LB address ), and set it before launching the agent. We currently configure it in /etc/hosts. 

I think there is a better way to achieve the same result, will rewrite the PR - we need both IP:port and the 'SAN' to validate, I started with keeping DiscoveryAddress be the URL and overriding the actual IP, but it seems better to go the other way, make the DiscoveryAddress be IP:port and overriding the SAN.

It seems I also need to update Citadel client ( I initially tested with MeshCA, but if citadel is used it needs similar override ).

Sorry for the false start, hold on for few days...

costinm:
PTAL - switched to overriding the SAN, and added citadel client too. This is far more useful - and may allow more flexibility and consistency if we want to allow Istiod to also support workload identity (again). I don't remember how we lost it, but older releases did support using spiffee identity for pilot.

pilot/cmd/pilot-agent/options/options.go
```diff
@@ -115,4 +115,7 @@ var (
 	// certSigner is cert signer for workload cert
 	certSigner = env.RegisterStringVar("ISTIO_META_CERT_SIGNER", "",
 		"The cert signer info for workload cert")
+
+	istiodSAN = env.RegisterStringVar("ISTIOD_SAN", "",
+		"Override the ServerName used to validate Istiod certificate. Can be used as an alternative to setting /etc/hosts for VMs - discovery address will be an IP:port")
```
howardjohn, line 120: should we switch our VM docs/code to use this instead

howardjohn (commented):
looks good, will leave approval to others as they had comments

pilot/cmd/pilot-agent/options/options.go
costinm: I would wait a bit, let's see how it goes with 'docker' - this mode is convenient if we use a custom launcher that can lookup the LB IP (or the Endpoint IP), and set up all env variables.

pkg/istio-agent/xds_proxy.go
costinm: Reverted to use discovery address, log is good now. If the error is related to validating the cert - will be logged in the err message.

yangminzhu (approved):

stevenctl (approved):

Issue #33558, opened 2021-06-21
Consistent hashing not supported for tcp/tls proxies


klarose:
**Bug description**

Due to a restriction in Envoy’s http 1.1 parser, I’m forced to SNI route to one of my backend services using TCP. Unfortunately, this service is stateful: I need to ensure multiple flows from the same source arrive at the same pod. I decided to use consistent hashing based on the source IP to achieve this

It does not work. Successive requests from the same source IP are sent to different pods. Looking through the proxy-config, I didn't see the relevant configuration in the tcpProxy, so it looks like istio just didn't generate the corresponding config. I also couldn't find anything in the istio source aside from http-related configuration.

I posted about this on the istio discussion board, but didn't hear anything back: https://discuss.istio.io/t/consistent-hashing-not-supported-for-tls-tcp-proxies/10629

[ ] Docs
[ ] Installation
[ X ] Networking
[ X ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

Requests from the same IP will go to the same pod, barring any changes in cluster state. 

Note: Envoy definitely supports this (I was able to work around by using an `EnvoyFilter` to patch the tcp proxy).

**Steps to reproduce the bug**

Configure a TCP gateway and virtual service. Configure source IP based consistent hashing to a tcp service with multiple instances. Send traffic hitting that route from the same IP. See that the connections are sent to different pods.

E.g. destination rule:
```
  trafficPolicy:
    connectionPool:
      tcp:
        connectTimeout: 1500ms
    loadBalancer:
      consistentHash:
        useSourceIp: true
    outlierDetection:
      baseEjectionTime: 2s
      consecutiveGatewayErrors: 3
      interval: 5s
```

**Version** (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version --short` if you used Helm)

```
client version: 1.9.5
control plane version: 1.9.5
data plane version: 1.9.5
```

**How was Istio installed?**

`istioctl install`

**Environment where the bug was observed (cloud vendor, OS, etc)**

GKE

**Workaround**

```
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: source-ip-hash
  namespace: istio-system
spec:
  configPatches:
  - applyTo: NETWORK_FILTER
    match:
      context: GATEWAY
      listener:
        portNumber: 8443
[...]

PR #35014, opened 2021-09-03
add support for sourceip hashing in tcp proxy

ramaraochavali:
Fixes https://github.com/istio/istio/issues/33558

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/networkfilter.go
+++ b/pilot/pkg/networking/core/v1alpha3/networkfilter.go
@@ -82,7 +82,7 @@ func buildOutboundNetworkFiltersWithSingleDestination(push *model.PushContext, n
 	}
 
 	if destinationRule != nil {
-		useSourceIp := destinationRule.GetTrafficPolicy().GetLoadBalancer().GetConsistentHash() != nil &&
+		useSourceIP := destinationRule.GetTrafficPolicy().GetLoadBalancer().GetConsistentHash() != nil &&
 			destinationRule.GetTrafficPolicy().GetLoadBalancer().GetConsistentHash().GetUseSourceIp()
 
 		for _, subset := range destinationRule.Subsets {
@@ -92,16 +92,16 @@ func buildOutboundNetworkFiltersWithSingleDestination(push *model.PushContext, n
 			// If subset has load balancer - see if it is also consistent hash source IP
 			if subset.TrafficPolicy != nil && subset.TrafficPolicy.LoadBalancer != nil {
 				if subset.TrafficPolicy.LoadBalancer.GetConsistentHash() != nil {
-					useSourceIp = subset.TrafficPolicy.LoadBalancer.GetConsistentHash().GetUseSourceIp()
+					useSourceIP = subset.TrafficPolicy.LoadBalancer.GetConsistentHash().GetUseSourceIp()
 				} else {
 					// This means that subset has defined non sourceIP consistent hash load balancer.
-					useSourceIp = false
+					useSourceIP = false
 				}
 			}
 			break
 		}
 		// If destinationrule has consistent hash source ip set, use it for tcp proxy.
-		if useSourceIp {
+		if useSourceIP {
 			tcpProxy.HashPolicy = []*hashpolicy.HashPolicy{{PolicySpecifier: &hashpolicy.HashPolicy_SourceIp_{
 				SourceIp: &hashpolicy.HashPolicy_SourceIp{},
 			}}}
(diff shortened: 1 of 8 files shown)


howardjohn:
We need to update the API docs as well. Also add an integration test to https://github.com/istio/istio/blob/f2909db45681292f4acaf550d83cc92f2be47828/tests/integration/pilot/common/routing.go#L1409.

The other thing is backwards compatibility. I think we are probably ok here, but it could represent a behavioral change for some users

pilot/pkg/networking/core/v1alpha3/networkfilter.go
```diff
@@ -167,7 +174,9 @@ func buildOutboundNetworkFilters(node *model.Proxy,
 			statPrefix = util.BuildStatPrefix(push.Mesh.OutboundClusterStatName, routes[0].Destination.Host,
 				routes[0].Destination.Subset, port, &service.Attributes)
 		}
-		return buildOutboundNetworkFiltersWithSingleDestination(push, node, statPrefix, clusterName, port)
+		destRule := push.DestinationRule(node, service)
+		destinationRule := CastDestinationRule(destRule)
+		return buildOutboundNetworkFiltersWithSingleDestination(push, node, statPrefix, clusterName, port, destinationRule)
 	}
 	return buildOutboundNetworkFiltersWithWeightedClusters(node, routes, push, port, configMeta)
```
hzxuzhonghu, line 181: Doesn't it need src ip hash?

pilot/pkg/networking/core/v1alpha3/networkfilter.go
```diff
@@ -69,18 +70,25 @@ func setAccessLogAndBuildTCPFilter(push *model.PushContext, config *tcp.TcpProxy
...
 	}
 
 	idleTimeout, err := time.ParseDuration(node.Metadata.IdleTimeout)
 	if err == nil {
 		tcpProxy.IdleTimeout = durationpb.New(idleTimeout)
 	}
 
+	// If destinationrule has consistent hash source ip set, use it for tcp proxy.
+	if destinationRule != nil && destinationRule.GetTrafficPolicy().GetLoadBalancer().GetConsistentHash() != nil &&
+		destinationRule.GetTrafficPolicy().GetLoadBalancer().GetConsistentHash().GetUseSourceIp() {
```
hzxuzhonghu, line 86: You should consider subset hash here

pilot/pkg/networking/core/v1alpha3/networkfilter.go
ramaraochavali: I do not know if it makes sense for weighted clusters - Does it?

pilot/pkg/networking/core/v1alpha3/networkfilter.go
ramaraochavali: Yes. I will add

ramaraochavali:
> We need to update the API docs as well. Also add an integration test to

Yes. Planning to update API docs once this is merged. Will add an integration test

ramaraochavali:
@hzxuzhonghu @howardjohn any more comments on this?

pilot/pkg/networking/core/v1alpha3/networkfilter.go
```diff
@@ -69,18 +70,44 @@ func setAccessLogAndBuildTCPFilter(push *model.PushContext, config *tcp.TcpProxy
...
 		ClusterSpecifier: &tcp.TcpProxy_Cluster{Cluster: clusterName},
-		// TODO: Need to set other fields such as Idle timeouts
 	}
 
 	idleTimeout, err := time.ParseDuration(node.Metadata.IdleTimeout)
 	if err == nil {
 		tcpProxy.IdleTimeout = durationpb.New(idleTimeout)
 	}
 
+	if destinationRule != nil {
```
howardjohn, line 84:
```suggestion
	if destinationRule.GetTrafficPolicy().GetLoadBalancer().GetConsistentHash().GetUseSourceIp() {
```

protobuf adds nil checks into each Get call

howardjohn: nevermind it doesn't quite work due to subsets. But we can simplify L85-L86

pilot/pkg/networking/core/v1alpha3/networkfilter.go
howardjohn: Don't we support it for HTTP? We should be consistent at least I think

howardjohn (commented):
LGTM aside from weighted cluster concerns

pilot/pkg/networking/core/v1alpha3/networkfilter.go
ramaraochavali: I just looked at envoy code and it picks the cluster based on weights first applies the hash - so we should be able to do it here as well. Let me try

howardjohn (approved):

PR #35022, opened 2021-09-03
Fix istio build in travis CI

libesz:
Add TRAVIS env vars to the excluded env var list, when building istio components and containers in docker. TRAVIS CI injects a number of env vars to the build environment, including the actual GIT commit message. When the commit message is a multiline string, docker fails to run with the provided environment settings.
Refs:
* https://github.com/moby/moby/issues/12997
* https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/common/scripts/setup_env.sh
+++ b/common/scripts/setup_env.sh
@@ -88,7 +88,7 @@ export IMG="${IMG:-gcr.io/istio-testing/${IMAGE_NAME}:${IMAGE_VERSION}}"
 
 export CONTAINER_CLI="${CONTAINER_CLI:-docker}"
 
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
Would it be more robust to actually filter out (or escape, if it works?) multi-line env vars?

In any event, this file is synced from istio/common-files repo so you would need to submit there

libesz:
@howardjohn thanks for the quick check, was not aware of the other repo.
I though I just continue the list of excluded vars as others did in the past, but yes, dynamic filtering would be definitely more robust.

howardjohn:
The excluded ones are more about specific variables from the host that
don't make sense in the container. TRAVIS doesn't have anything
specifically in that category other than the general issue of spaces. I am
fine with thisapproach but could be more flexible if you have time.

PR #35023, opened 2021-09-03
Remove explicit NegotiatedSerializer which breaks controller-runtime

howardjohn:
See https://github.com/kubernetes-sigs/controller-runtime/issues/1654

### DIFF
--- a/pkg/kube/util.go
+++ b/pkg/kube/util.go
@@ -25,7 +25,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/kubernetes/scheme"
 
 	//  allow out of cluster authentication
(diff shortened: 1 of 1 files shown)


pkg/kube/util.go
```diff
@@ -144,12 +143,6 @@ func SetRestDefaults(config *rest.Config) *rest.Config {
 	if len(config.ContentType) == 0 {
 		config.ContentType = runtime.ContentTypeJSON
 	}
-	if config.NegotiatedSerializer == nil {
-		// This codec factory ensures the resources are not converted. Therefore, resources
-		// will not be round-tripped through internal versions. Defaulting does not happen
-		// on the client.
-		config.NegotiatedSerializer = scheme.Codecs.WithoutConversion()
```
hzxuzhonghu, line 151: This is not the right setup, we should set each groups' sheme.

But this field is also a must.

howardjohn:
Yeah I will make a different change

howardjohn:
obsolete by https://github.com/istio/istio/pull/35049

PR #35024, opened 2021-09-03
Implement a new controller to deploy a Service/Deploy from a Gateway

howardjohn:
This implements https://docs.google.com/document/d/1PfT7jdXSJ5iU_kU771YpV8Jq6l017Y65crOHJxXZORg/edit#

Open question: how do we want to enable it? I want it always-on, since using the API is very explicit, but that does mean all users will get elevated role (for svc/deployment write).


hzxuzhonghu (changes requested):
I donot think this is what istiod should do, i feels maybe this can goes to istio-operator

howardjohn:
> I donot think this is what istiod should do, i feels maybe this can goes to istio-operator

IMO this must be in Istiod or not at all. The reason is this implements a core networking API (Gateway) and is primarily  to benefit the simpler use cases.

Requiring a core API to require a separate deployment goes against the Istiod design goals.

Requiring a separate deployment to do this defeats the simplicity goals of this design, as now they have to deploy a whole new component just to operate their gateways.


I think we have made a conscious decision with the Istiod movement to have a modularized monolith. All components (that reasonable can) should belong in Istiod; but they should be segmented and disable-able. This proposal meets that goal - if a user is not comfortable with this running in Istiod they are free to disable it, and I am happy to abstract it a bit more so it can run alone in a standalone `main.go` if users are interested in deploying just this controller as another deployment.

hzxuzhonghu:
I would also want @istio/wg-product-security-maintainers @istio/wg-security-maintainers to have a look,

howardjohn:
@hzxuzhonghu I have brought this up to PSWG/Security WG and got approval on the document.

howardjohn:
@hzxuzhonghu I have also brought this up to TOC and there were not concerns

manifests/charts/istiod-remote/templates/clusterrole.yaml
```diff
@@ -36,6 +36,19 @@ rules:
     verbs: [ "get", "watch", "list", "update", "patch", "create", "delete" ]
     resources: [ "workloadentries/status" ]
 
+  # gateway controller
+{{- if not (eq (toString .Values.pilot.env.PILOT_ENABLE_GATEWAY_API_DEPLOYMENT_CONTROLLER) "false") }}
```
costinm, line 40: If possible, can you add a new ClusterRole object with the extra stuff, and a separate ClusterRoleBinding ? 
Seems better than having cluster roles that have different content based on flags.

Usually the cluster roles should be narrow and documented - the SA can have multiple bindings.

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -111,6 +111,24 @@ func New(client kube.Client, revision, domainSuffix string) (model.ConfigStoreCa
 	return NewForSchemas(context.Background(), client, revision, domainSuffix, schemas)
 }
 
+var crdWatches = map[config.GroupVersionKind]chan struct{}{
+	gvk.KubernetesGateway: make(chan struct{}),
+}
+
+func WaitForCRD(k config.GroupVersionKind, stop <-chan struct{}) bool {
```
costinm, line 118: Some comment would be nice. Is this a new way to enable features only when the CRD is created ? Wouldn't a callback be better ( i.e. a map of CRD types to init functions ? )

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -492,7 +510,11 @@ func handleCRDAdd(cl *Client, name string, stop <-chan struct{}) {
 		scope.Errorf("failed to create informer for %v", resourceGVK)
 		return
 	}
-	cl.kinds[s.Resource().GroupVersionKind()] = createCacheHandler(cl, s, i)
+	cl.kinds[resourceGVK] = createCacheHandler(cl, s, i)
+	if w, f := crdWatches[resourceGVK]; f {
```
costinm, line 514: I like the idea of starting various controllers if/when the CRD is detected, we should generalize it.

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -1129,6 +1116,35 @@ func convertGateways(r *KubernetesResources) ([]config.Config, map[parentKey]map
 	return result, gwMap, namespaceLabelReferences
 }
 
+func isManaged(gw *k8s.GatewaySpec) bool {
+	return len(gw.Addresses) == 0
```
costinm, line 1120: Comments. Why would it be managed if addresses are empty ?

pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+	if err := d.ApplyObject(gws, "status"); err != nil {
+		return fmt.Errorf("update gateway status: %v", err)
+	}
+	log.Info("gateway updated")
+	return nil
+}
+
+// ApplyTemplate renders a template with the given input and (server-side) applies the results to the cluster.
+func (d *DeploymentController) ApplyTemplate(template string, input interface{}, subresources ...string) error {
```
costinm, line 219: Seems like a pretty generic function. 
- don't we have it already ?
- should we move to /pkg ?

pilot/pkg/config/kube/gateway/templates.go
```diff
@@ -0,0 +1,61 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+// Templates embeds the templates
+//go:embed templates/*
+var Templates embed.FS
+
+func funcMap() template.FuncMap {
+	f := sprig.TxtFuncMap()
+	f["strdict"] = func(v ...string) map[string]string {
```
costinm, line 32: 'strdict' ? At least a comment explaining what it does. 

Does helm have a similar function ? It wouldn't be bad to reuse the names.

pilot/pkg/config/kube/gateway/templates/deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+  labels:
+    {{ toYamlMap .Labels
+      (strdict "gateway.istio.io/managed" "istio.io-gateway-controller")
+      | nindent 4}}
+  name: {{.Name}}
+  namespace: {{.Namespace}}
+  ownerReferences:
+  - apiVersion: gateway.networking.k8s.io/v1alpha2
+    kind: Gateway
+    name: {{.Name}}
```
costinm, line 15: Would be really nice to use the same template elements/names as helm, it is harder to read/understand otherwise. And if this gets more complicated - we may reuse the helm-applying code we use in istioctl.

pilot/pkg/config/kube/gateway/templates/service-export.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: multicluster.x-k8s.io/v1alpha1
+kind: ServiceExport
```
costinm, line 2: Why is ServiceExport here ?

costinm: I mean: we should not export it automatically AFAIK.

pkg/kube/client.go
```diff
@@ -401,7 +401,19 @@ func NewClient(clientConfig clientcmd.ClientConfig) (Client, error) {
 	return newClientInternal(newClientFactory(clientConfig), "")
 }
 
+// NewDefaultClient creates a Kubernetes client from the current Kubernetes context.
```
costinm, line 404: Do we need this ? Seems strange to not have needed it so far.

costinm (commented):
Generally looks good, left a couple of comments but must can be addressed by adding TODOs and extra comments in the code, nothing fundamentally wrong.

kyessenov:
Services are not intended as an automation API. There are quadratic costs associated, and vendors put a hard cap on number of services permitted. Is there an analysis of how many gateways are normally deployed?

howardjohn:
> Services are not intended as an automation API.

Why not? There seems to be a large number of controllers existing that create Service, and all current implementations of gateway-api are planning the same.

> Is there an analysis of how many gateways are normally deployed?

Good question. We cannot get hard data since no one is using the API in prod but it would be reasonable to assume its less than Istio Gateways given our deployment mode is now 1:1 with Envoy whereas with Istio gateway its N:1.

For Istio gateways, we see:
p50 = 1 gateways
p95 = 4 gateways
p99 = 25 gateways

I do not expect problems here. Users will be well aware that a Gateway resource has potential underlying cost, as ALL implementations of gateway-api will have this behavior.

pilot/pkg/config/kube/gateway/templates/service-export.yaml
howardjohn: Its not automatic, but allow users a way to export from the Gateway, vs needing to create a ServiceExport on their own (which somewhat leaks the fact there is a Service created).

Thinking about it more its probably not very useful

pilot/pkg/config/kube/crdclient/client.go
howardjohn: Yes, added some comments.

A callback seems nice but I think it may add some complications around the startup logic that makes it a bit more complex. Right now the startup handles releasing leader lock on shutdown, etc that would be a bit harder with callback. If there are more use cases  that callback would be better we can always add it later I think?

pilot/pkg/config/kube/gateway/conversion.go
howardjohn: Added comment in code, copying here:
```go
// isManaged checks if a Gateway is managed (ie we create the Deployment and Service) or unmanaged.
// This is based on the address field of the spec. If address is set, it should point to an existing
// Service that handles the gateway traffic. If it is not set, we will consider it managed and provision the Service.
// While there is no defined standard for this in the API yet, it is tracked in https://github.com/kubernetes-sigs/gateway-api/issues/892.
// So far, this mirrors how out of clusters work (address set means to use existing IP, unset means to provision one),
// and there has been growing consensus on this model for in cluster deployments.
func isManaged(gw *k8s.GatewaySpec) bool {
	return len(gw.Addresses) == 0
}
```

pilot/pkg/config/kube/gateway/deploymentcontroller.go
howardjohn: We don't. The only thing really apply templates to cluster is the operator code, which use controller-runtime which is incompatible with our needs for Istiod (discussed at length in the DeploymentController comments).

There is the new `controllers` package so moving it there seems reasonable

pilot/pkg/config/kube/gateway/templates.go
howardjohn: Yeah the problem is it returns the wrong type unfortunately. added a comment

pkg/kube/client.go
howardjohn: no, was for debugging

costinm (approved):
On creating Service and its cost: KNative and others do the same thing, and AFAIK the consensus is that subsets will also be represented by Service ( like KNative does).

We may start using IPV6, and k8s may need to optimize how they implement Service - but in case of Gateway it is really not a problem since we would create the same service if helm was used.

howardjohn:
@hzxuzhonghu ptal.

hzxuzhonghu (commented):
The implement looks good to me, I am concerned about the number of gateways as well

howardjohn:
> The implement looks good to me, I am concerned about the number of gateways as well

Just like today, the user can choose how many they want. If they only need 1 then they should just create 1 gateway.

hzxuzhonghu (approved):
approve since it is optional

PR #35029, opened 2021-09-05
reduce cpu usage when pilot agent get envoy stats

holooooo:

When prom scrap stats of pilot agent, Pilot agent will use io.ReadAll to read all the byte in response.But response will use a 512 bytes slice as  buffer.It will take a lot of time to grow slice and clean up the old slice when envoy stats is huge(such as gateway or some naming service).

So I replace io.ReadAll with io.Copy. Copy uses constant memory space, it can reduce CPU usage time.


pilot/cmd/pilot-agent/status/server.go
```diff
@@ -502,8 +514,56 @@ func negotiateMetricsFormat(contentType string) expfmt.Format {
...
+			continue
+		}
+		if n+len(line)+1 > cap(p) {
+			for range line {
+				_ = r.buf.UnreadByte()
+			}
+			return
+		}
+		copy(p[n:], line)
+		p[n+len(line)] = '\n'
```
howardjohn, line 559: If we are going to add this custom parsing logic I think we need a LOT of testing and benchmarks to prove this is worth the complexity

howardjohn (changes requested):
Thanks for the PR! I think there are some good improvements to make here

holooooo:
In Go 1.17 ,use custom filter to replace "\n\n" with '\n' is cheaper than replaceAll after readAll, and friendly to gc

image
image

image
image

image
image

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -502,8 +514,51 @@ func negotiateMetricsFormat(contentType string) expfmt.Format {
 	return expfmt.FmtText
 }
 
-func processMetrics(metrics []byte) []byte {
-	return bytes.ReplaceAll(metrics, []byte("\n\n"), []byte("\n"))
+// metricReader used to replace "\n\n" in stream with '\n'.
```
howardjohn, line 517: I think this is missing an explanation about *why* we do this?

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -502,8 +514,51 @@ func negotiateMetricsFormat(contentType string) expfmt.Format {
...
+			length, err = w.Write(newLine)
+			n += int64(length)
+		}
+	}
+	return
+}
+
+func (r *metricReader) Read(p []byte) (n int, err error) {
+	buf := bytes.NewBuffer(p)
+	buf.Reset()
```
howardjohn, line 555: why reset?

holooooo: Oh 😢 It is a mistake. Because read is never used.So I want to multiplexing writeTo to implement read.

Implementation using bytes.buffer looks like easy. But byte.buffer can only be written from the start of the slice when using an empty slice as the buffer.So I reset it to make sure that slice has no part filled with 0 value.

But bytes.buffer has the side effect of make slice to grow up.I forget that 🤔

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -570,17 +630,12 @@ func (s *Server) scrape(url string, header http.Header) ([]byte, string, error)
...
 		return nil, "", fmt.Errorf("error scraping %s, status code: %v", url, resp.StatusCode)
 	}
-	metrics, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return nil, "", fmt.Errorf("error reading %s: %v", url, err)
-	}
 
 	format := resp.Header.Get("Content-Type")
-	return metrics, format, nil
+	return resp.Body, format, nil
```
holooooo, line 638: needs to handle cancel

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -482,15 +490,19 @@ func (s *Server) handleStats(w http.ResponseWriter, r *http.Request) {
 		log.Errorf("failed to write agent metrics: %v", err)
 		metrics.AgentScrapeErrors.Increment()
 	}
-	if _, err := w.Write(envoy); err != nil {
-		log.Errorf("failed to write envoy metrics: %v", err)
-		metrics.EnvoyScrapeErrors.Increment()
+	if envoy != nil {
+		if _, err := io.Copy(w, envoy); err != nil {
```
holooooo, line 494: This introduces a new problem. That is, the request for envoy stats may be interrupted because the response is written too slowly.

pilot/cmd/pilot-agent/status/server.go
holooooo: done

pilot/cmd/pilot-agent/status/server.go
holooooo: @howardjohn Here is refactored, Read will be degraded to read all before reading. It make sure io.Reader is implementation.

pilot/cmd/pilot-agent/status/server.go
holooooo: @howardjohn The code comments have been completed

holooooo:
The writeto is reconstructed, the performance is increased by 10% compared with the previous implementation, and can deal with some envoy interrupt requests caused by the client side reading metrics too slowly.

At the same time, in order to correctly handle openmetrics, the call order of handlestats is modified. At the beginning of the request, the indicator of the scrape app metrics will be first and the corresponding content type will be set

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -502,8 +493,68 @@ func negotiateMetricsFormat(contentType string) expfmt.Format {
...
+		}
+	}
+	if err == io.EOF {
+		err = nil
+	}
+	return
+}
+
+// Read will not be used often, so WriteTo is reused
+func (r *metricReader) Read(p []byte) (n int, err error) {
```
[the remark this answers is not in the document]
holooooo: True,I'm stuck in the original idea 🤣

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -502,8 +493,68 @@ func negotiateMetricsFormat(contentType string) expfmt.Format {
...
+		// Once get unexpected error, drop last line that may be incomplete
+		if err != nil && err != io.EOF && err != bufio.ErrBufferFull {
+			break
+		}
+		// Remove "# EOF" avoid terminates the full exposition
+		if bytes.HasPrefix(line, EOFLine) {
+			line = line[len(EOFLine):]
+		}
+		// If last line size equal buffer size, need to add "\n"
+		isEmptyLine = bytes.Equal(line, newLine) && !isBufferFull
```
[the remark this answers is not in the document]
holooooo: Yeah, it is not necessary.There is a version that needs to be under the function scope, but the logic is too complex. I forgot to fix these scopes after modify.

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -502,8 +493,68 @@ func negotiateMetricsFormat(contentType string) expfmt.Format {
...
+		if isEmptyLine {
+			continue
+		}
+		wLength, werr := w.Write(line)
+		n += int64(wLength)
+
+		isBufferFull = false
+		if werr == nil && err == bufio.ErrBufferFull {
+			isBufferFull = true
```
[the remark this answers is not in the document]
holooooo: If there is a very large line (larger than buffer size), it will lead to wrong line feed.Although bufio.reader get 4k buffer, I still worry about it(current, our line is about 1.1k in size).

Once line length over the buffer size, ReadSlice while return a bufio.ErrBufferFull error and not '\n' in line.So we need to avoid exiting functions early because of errors. And add "\n" if last line size equal buffer size.

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -502,8 +493,68 @@ func negotiateMetricsFormat(contentType string) expfmt.Format {
...
 }
 
-func processMetrics(metrics []byte) []byte {
-	return bytes.ReplaceAll(metrics, []byte("\n\n"), []byte("\n"))
+// metricReader used to remove all the blank, "# EOF" or incomplete lines in envoy metrics.
+// It makes the envoy metric compatible with FmtOpenMetrics (https://github.com/istio/istio/pull/33550)
+type metricReader struct {
+	reader  io.ReadCloser
+	buf     *bufio.Reader
+	readBuf *bytes.Buffer
```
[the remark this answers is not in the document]
holooooo: ReadBuf is used to implementation.Now it useless.

PR #35037, opened 2021-09-06
Remove the corresponding file when remove deleted resources from cache

devincd:
Signed-off-by: devincd <505259926@qq.com>

We should remove the corresponding file when remove deleted resources from cache.


hzxuzhonghu (approved):
LGTM, maybe should remove local cache later.

PR #35039, opened 2021-09-06
Not allocate address for dns type serviceentry with endpoints

hzxuzhonghu:

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/serviceregistry/serviceentry/conversion.go
+++ b/pilot/pkg/serviceregistry/serviceentry/conversion.go
@@ -195,13 +195,24 @@ func convertServices(cfg config.Config) []*model.Service {
 		}
 	}
 
+	allocateAddress := true
+	// donot allocate address for DNS type serviceentry without endpoints specified.
+	if resolution == model.DNSLB && len(serviceEntry.Endpoints) == 0 && serviceEntry.WorkloadSelector == nil {
+		allocateAddress = false
+	}
+
 	return buildServices(hostAddresses, cfg.Namespace, svcPorts, serviceEntry.Location, resolution,
-		exportTo, labelSelectors, serviceEntry.SubjectAltNames, creationTime, cfg.Labels)
+		exportTo, labelSelectors, serviceEntry.SubjectAltNames, creationTime, cfg.Labels, allocateAddress)
 }
 
 func buildServices(hostAddresses []*HostAddress, namespace string, ports model.PortList, location networking.ServiceEntry_Location,
 	resolution model.Resolution, exportTo map[visibility.Instance]bool, selectors map[string]string, saccounts []string,
-	ctime time.Time, labels map[string]string) []*model.Service {
+	ctime time.Time, labels map[string]string, autoAllocateAddr bool) []*model.Service {
+	autoAllocatedAddress := ""
+	if !autoAllocateAddr {
+		// set to "0.0.0.0" here to prevent allocate a E class address later for it.
+		autoAllocatedAddress = constants.UnspecifiedIP
+	}
 	out := make([]*model.Service, 0, len(hostAddresses))
 	for _, ha := range hostAddresses {
 		out = append(out, &model.Service{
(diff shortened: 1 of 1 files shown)


pilot/pkg/serviceregistry/serviceentry/conversion.go
```diff
@@ -195,13 +195,24 @@ func convertServices(cfg config.Config) []*model.Service {
 		}
 	}
 
+	allocateAddress := true
+	// donot allocate address for DNS type serviceentry without endpoints specified.
+	if resolution == model.DNSLB && len(serviceEntry.Endpoints) == 0 && serviceEntry.WorkloadSelector == nil {
```
howardjohn, line 200: Why? I do not understand. `address` and `endpoints` are completely orthogonal. `address` is how we match incoming requests. `endpoints` is where we send outgoing requests

howardjohn (changes requested):

pilot/pkg/serviceregistry/serviceentry/conversion.go
hzxuzhonghu: The intention here is not to allocate address for ServiceEntry without endpoints, it has nothing to do with address. 

Whether it has address set or not, donot allocate it for the below SE.

```
kind: ServiceEntry
metadata:
  name: jsontime-entry
spec:
  hosts:
  - time.jsontest.com
  ports:
  - number: 80
    name: http-port
    protocol: HTTP
  resolution: DNS
```

howardjohn:
Sorry I am confused. How can allocate address have nothing to do with
address? The whole feature is about automatically setting `address`? Sorry
I am not understanding

hzxuzhonghu:
1. for SE with addreses explicitly set, when convert it to model.Service, it has an adresss. We donot allocate for this type.
2. for Se of DNS resolution with endpoints, the address of the endpoint can be ip or fqdn. In this case, allocate address for SE's host does not redirect traffic to none existing backend.
3. But for  Se of DNS resolution without endpoints, allocate address for it will redirect traffic to unavailable address.

howardjohn:
I Still don't understand why address and endpoint (or resolution type) have anything to do with each other. Address configures how we match the traffic. Endpoints is where we send the traffic to.. so what is the relationship?

Can you provide a specific example that is broken without this PR and fixed with it?

hzxuzhonghu:
```
kind: ServiceEntry
metadata:
  name: external-svc-redirect
spec:
  hosts:
  - www.google.com
  location: MESH_EXTERNAL
  ports:
  - number: 443
    name: https
    protocol: TLS
  resolution: DNS
```

howardjohn:
And what is behavior before and after? I can try it out tomorrow as well. But I would expect that without auto allocate it would match all SNI for google.com on any IP, and with it it would match the auto allocated IP. Seems like it should work fine but maybe a different bug..

howardjohn:
Sorry I cannot try it out this moment, don't have access to a cluster

hzxuzhonghu:
Without this pr, if dns capture enabled, istio will allocate a 240.xxx address for www.google.com. google.com is definitely inaccessible.

howardjohn:
My understanding is google.com is resolved by Envoy. Envoy DNS resolution
is not capture by istio-agent, so it goes to upstream resolver and gets
normal DNS response?

hzxuzhonghu:
But `curl www.google.com:80` failed in this case

howardjohn:
I think that is a completely different problem and don't see how this PR helps it. Note this config (which has endpoints) set is broken as well:
```
kind: ServiceEntry
metadata:
  name: external-svc-redirect
spec:
  hosts:
  - www.google.com
  location: MESH_EXTERNAL
  ports:
  - number: 443
    name: https
    protocol: TLS
  resolution: DNS
  endpoints:
  - address: www.google.com

```

Which I don't think this PR would fix, since it only handles this special case.

The problem is a fundamental flaw of using fake IP address. `curl www.google.com:80` is working only because its going to the passthrough cluster. When we create the SE above, we configure only port 443. So now we get IP 240.240.240.240:80; there is no match on port 80 (its only on 443, as configured in the SE). This hits passthrough clsuter which of course breaks.

hzxuzhonghu:
Right, the example is very special.

Issue #34855, opened 2021-08-24
envoy drain duration should adapt to pod spec


kyessenov:
There are four (!) settings controlling the sidecar termination duration:
1. `drainDuration` is envoy's graceful drain duration (default 45s).
2. `parentShutdownDuration` is envoy's parent shutdown delay (default 60s).
3. `terminationDrainDuration` is pilot-agent's termination duration delay after SIGTERM (default 5s).
4. `terminationGracefulPeriodSeconds` is pod's graceful termination delay after SIGTERM before SIGKILL (default 30s).

This is very confusing and is actually inconsistent since (1)-(3) are normally mesh-level and (4) is pod-level. (4) overrides (3), and (3) overrides (1), while (2) is not used. (3) is also too short to be meaningful since envoy's drain is passive.

The proposal is to use (4) as the main deadline since it's orchestrated by kubelet irrespective of mesh-level settings. In this world, there is just one graceful deadline value (4) that dictates how long envoy gracefully drains open connections before shutting down. That means envoy is not terminated abruptly without sending final telemetry, for example. The process would look like this:

1. On SIGTERM, agent initiates inbound listener graceful drain using (4) as the deadline. GOAWAY and connection:close are sent on all inbound connections, no more inbound connections are accepted, container readiness fails, endpoint removed from the pools. The max drain deadline is set by (4), but envoy can be stopped before that once all outstanding requests/connections are finished (including outbound app-driven). It's probably good to have a fixed minimum delay because detecting if an app (peer container) has outstanding requests is difficult (alternatively, CNI can undo outbound iptables on termination).
2. After (4) seconds, SIGKILL is issued if envoy does not stop before that.

cc @ramaraochavali @mandarjog

kyessenov:
Ref https://github.com/envoyproxy/envoy/issues/17056

howardjohn:
We use `--drain-strategy immediate` which should somewhat help "envoy's drain is passive" if you weren't aware - doesn't work in all cases though.

Overall I agree we should do this. A few concerns I think we will need to work out:
* what we should do if users set `terminationDrainDuration` (or just have the old default of 5s)?
* Will we hit any cases where we get stuck? For example, I have seen some users set terminationGracefulPeriodSeconds to extreme values (>10 minutes) - if there are cases where their pod would have shut down and now it doesn't we may run into issues

The benefits largely outweigh the issues, but I we should see what we can do to address those

kyessenov:
[...]

PR #35043, opened 2021-09-07
terminate envoy when number of active connections is zero

ramaraochavali:
For https://github.com/istio/istio/issues/34855 except that it still uses terminationDrainDuration.

It gets `downstream_cx_active` of all listeners (ignores admin and stats listeners) and when the total becomes zero, it terminates Envoy or waits for terminationDrainDuration. It starts this process after waiting for minDrainDuration i.e. 5s.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali:
@howardjohn @kyessenov I was experimenting this a bit and here is draft PR. Please take a look - if the approach makes sense - I will add tests.

pkg/config/mesh/mesh.go
```diff
@@ -43,7 +43,7 @@ func DefaultProxyConfig() meshconfig.ProxyConfig {
 		ServiceCluster:           constants.ServiceClusterName,
 		DrainDuration:            types.DurationProto(45 * time.Second),
 		ParentShutdownDuration:   types.DurationProto(60 * time.Second),
-		TerminationDrainDuration: types.DurationProto(5 * time.Second),
+		TerminationDrainDuration: types.DurationProto(45 * time.Second),
```
howardjohn, line 46: Discussed a bit more in the issue, but we need to make sure we don't make this a breaking API change

pkg/envoy/agent.go
```diff
@@ -17,22 +17,43 @@ package envoy
...
 )
 
 var errAbort = errors.New("epoch aborted")
 
 const errOutOfMemory = "signal: killed"
 
+// minDrainDuration is the minimum duration for which agent waits before it actively checks
+// for outstanding connections. After this period, agent terminates proxy whenever active
+// connections become zero.
+var minDrainDuration = 5 * time.Second
```
howardjohn, line 37: same comment about breaking change. You should be able to override to 0s

pkg/envoy/agent.go
```diff
@@ -17,22 +17,43 @@ package envoy
...
+
+var activeConnectionCheckDelay = 1 * time.Second
+
+var knownIstioListeners = sets.NewSet(
+	"listener.0.0.0.0_15006.downstream_cx_active", "listener.0.0.0.0_15021.downstream_cx_active",
```
howardjohn, line 42: The ports can be configured

pkg/envoy/agent.go
```diff
@@ -105,13 +131,72 @@ func (a *Agent) terminate() {
 	if e != nil {
 		log.Warnf("Error in invoking drain listeners endpoint %v", e)
 	}
-	log.Infof("Graceful termination period is %v, starting...", a.terminationDrainDuration)
-	time.Sleep(a.terminationDrainDuration)
-	log.Infof("Graceful termination period complete, terminating remaining proxies.")
-	a.abortCh <- errAbort
+	time.Sleep(minDrainDuration)
```
howardjohn, line 134: on L129 log what we are going to do maybe. Like `Agent draining proxy for 5s, then waiting for active connections to terminate...`

pkg/envoy/agent.go
```diff
@@ -17,22 +17,43 @@ package envoy
...
+	"listener.admin.main_thread.downstream_cx_active",
+)
+
 // NewAgent creates a new proxy agent for the proxy start-up and clean-up functions.
-func NewAgent(proxy Proxy, terminationDrainDuration time.Duration) *Agent {
+func NewAgent(proxy Proxy, terminationDrainDuration time.Duration, localhost string, adminPort uint16) *Agent {
 	return &Agent{
 		proxy:                    proxy,
 		statusCh:                 make(chan exitStatus, 1), // context might stop drainage
+		drainCh:                  make(chan bool),
```
howardjohn, line 52:
```suggestion
		drainCh:                  make(chan struct{}),
```

then use `close(drainCh)` to signal, not send

kyessenov:
The cleanest solution is to wait for other containers to complete IMHO (and dealing with their organic restarts). 

What you have here seems reasonable. The danger is some app might want to open a connection at the end (think heavy app generating logs and dumping them somewhere else). So this won't work in general.

I'm actually would be totally OK if we stopped using envoy for readiness and prometheus ports. I don't like the complexity incurred by having to deal with those listeners. It's complexity for complexity sake.

ramaraochavali:
@howardjohn @kyessenov Some how this branch is messedup. Closing this and opening a new PR https://github.com/istio/istio/pull/35059.

@howardjohn addressed all your comments. If we are good with the approach I will add tests.

howardjohn:
> The cleanest solution is to wait for other containers to complete IMHO (and dealing with their organic restarts).

I agree but don't know how to do this. Any ideas?

PR #35044, opened 2021-09-07
Fixed Unable to add an item to a list using overlays in IstioOperator

litong01:
Fixes#34024 
The issue was that the code comments do not match up with the actual
code behavior. This patch fixed the comments so that the claimed and
the actual behavior will match

### DIFF
--- a/releasenotes/notes/34024.yaml
+++ b/releasenotes/notes/34024.yaml
@@ -1,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: installation
+issue:
+- 34024
+releaseNotes:
+- |
+  **Fixed** Unable to add an item to a list using overlays in IstioOperator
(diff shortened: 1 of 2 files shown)


releasenotes/notes/34024.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: installation
+issue:
+- 34024
+releaseNotes:
+- |
+  **Fixed** Unable to add an item to a list using overlays in IstioOperator
```
howardjohn, line 8: I don't think we should have release notes for a comment in the code, its not using facing

ostromart (approved):

releasenotes/notes/34024.yaml
litong01: @howardjohn would you like me to remove the releasenotes? I was not really sure if we need one or not. The issue is that the comment in the code read by users and they followed the comment to author their manifest file. when they follow the what the comments said, they got error, that is why we have this issue open. I can remove it but this is actually user facing in my view. Please let me know what you think. Thanks.

PR #35046, opened 2021-09-07
Add a rate limit to incoming XDS requests

howardjohn:
In a cluster we saw thundering herd problem. When a new istiod spins up, it instantly receives 1k requests and attempts to process them all. This leads to OOM, so all of them disconnect and try again to new instance. While the instance can handle 1k connections, it cannot handle 1k requests all at once. Requests are the most expensive (always total full push) and there is no limiting.

With rate limit, we can limit max memory and avoid throttling, allowing us to slowly ramp up number of connections.

An alternative or addition we can do is to have a semaphore limiting max *concurrent* requests. The reason I went with this is:
* Rate limit makes it easier to block requests immediately - before we do anything expensive at all or cause churn with connects+disconnects, workload entry registration, etc. If we had a semaphore that early then we would be blocking all requests on synchronous api-server calls (for VM registration) which could block everyone. This doesn't seem right.
* I thought about *also* adding a semaphore. We have an existing one for pushes; I don't think we want to reuse it though, as requests must take precedence over pushes. If we had a way to have a shared semaphore that favors requests over pushes that would be ideal IMO but seems a bit too complex for now

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -64,6 +64,12 @@ var (
 		"Limits the number of concurrent pushes allowed. On larger machines this can be increased for faster pushes",
 	).Get()
 
+	RequestLimit = env.RegisterFloatVar(
+		"PILOT_MAX_REQUESTS_PER_SECOND",
+		100.0,
+		"Limits the number of incoming XDS requests per second. On larger machines this can be increased to handle more proxies concurrently.",
+	).Get()
+
 	// MaxRecvMsgSize The max receive buffer size of gRPC received channel of Pilot in bytes.
 	MaxRecvMsgSize = env.RegisterIntVar(
 		"ISTIO_GPRC_MAXRECVMSGSIZE",
--- a/pilot/pkg/xds/discovery.go
+++ b/pilot/pkg/xds/discovery.go
@@ -15,6 +15,7 @@
 package xds
 
 import (
+	"context"
 	"fmt"
 	"strconv"
 	"sync"
@@ -23,6 +24,7 @@ import (
 	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	"github.com/google/uuid"
 	"go.uber.org/atomic"
+	"golang.org/x/time/rate"
 	"google.golang.org/grpc"
 
 	"istio.io/istio/pilot/pkg/controller/workloadentry"
@@ -92,7 +94,10 @@ type DiscoveryServer struct {
 	// may also choose to not send any updates.
 	ProxyNeedsPush func(proxy *model.Proxy, req *model.PushRequest) bool
 
+	// concurrentPushLimit is a semaphore that limits the amount of concurrent XDS pushes.
 	concurrentPushLimit chan struct{}
+	// requestRateLimit limits the number of new XDS requests allowed. This helps prevent thundering hurd of incoming requests.
+	requestRateLimit *rate.Limiter
 
 	// InboundUpdates describes the number of configuration updates the discovery server has received
 	InboundUpdates *atomic.Int64
@@ -180,6 +185,7 @@ func NewDiscoveryServer(env *model.Environment, plugins []string, instanceID str
 		ProxyNeedsPush:          DefaultProxyNeedsPush,
 		EndpointShardsByService: map[string]map[string]*EndpointShards{},
 		concurrentPushLimit:     make(chan struct{}, features.PushThrottle),
+		requestRateLimit:        rate.NewLimiter(rate.Limit(features.RequestLimit), 1),
 		InboundUpdates:          atomic.NewInt64(0),
 		CommittedUpdates:        atomic.NewInt64(0),
 		pushChannel:             make(chan *model.PushRequest, 10),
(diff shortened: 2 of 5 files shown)


pilot/pkg/xds/discovery.go
```diff
@@ -649,3 +655,11 @@ func (s *DiscoveryServer) ClientsOf(typeUrl string) []*Connection {
 
 	return pending
 }
+
+func (s *DiscoveryServer) WaitForRequestLimit(ctx context.Context) error {
```
costinm, line 659: Should we make the interval configurable too ( fill_interval - and may rename the other setting tokens_per_fill - for consistency with envoy naming in rate limit ) ?

I was not sure if it should wait 1 sec if bucket is empty - or return immediately so it can try another backend. But for this case I think waiting is better, not sure if the exponential backoff is properly implemented in clients and other intermediaries.

costinm (approved):
Looks good, it's something we should strongly consider for backporting - but if we do it should be off by default.

howardjohn:
Put a hold to make sure 0s == disable instead of always reject

pilot/pkg/xds/discovery.go
```diff
@@ -649,3 +655,11 @@ func (s *DiscoveryServer) ClientsOf(typeUrl string) []*Connection {
...
+	// Give a bit of time for queue to clear out, but if not fail fast. Client will connect to another
+	// instance in best case, or retry with backoff.
+	wait, cancel := context.WithTimeout(ctx, time.Second)
+	defer cancel()
+	return s.requestRateLimit.Wait(wait)
```
hzxuzhonghu, line 664: @howardjohn I think this can not handle 1k xds requests being handled concurrently. because of the duration of each request handling.

In other word, why not handle concurrent xds processing with buffered channel like what we have done for xds push.

howardjohn: I discussed this a bit in the issue description. The problem we face in doing that is the request initialization has a number of expensive actions that are not done in the push. In particular, there is an API server call to create workload entry for VMs. If we have that locked by a semaphore, then even a small latency spike in api-server will cause xds to be blocked.

With this implementation it will not be. While it may not be ideal in every possible case it is very simple and should make every case behave better. In the extreme rare case that someone actually needs to serve over 100qps with a single instance, they can adjust it disable this

hzxuzhonghu: This scenario is like kube-apiserver, it denies any request incoming if the total requests being precessed exceed the max inflight limit.  Which is far more safer for a server. Even with this, k8s introduced flow control recently https://kubernetes.io/docs/concepts/cluster-administration/flow-control/

howardjohn:
Thanks for the link. I think when you need a 10 page doc to explain your
rate limiting it's far too complex. I don't think what we have here is the
perfect solution, but it's dead simple and is effective. Since we have long
lived connections and we are trying to limit only the initial burst (since
start of connection is expensive, but after that it's not) it's not very
clear how we can implement a "max in flight" without substantial complexity.

hzxuzhonghu:
Let's talk about this from another side. Lets's say we have 10k sidecars connected to a single pilot, when a full xds push triggered and later pushed 4 kind of xds to all sidecars in 2s(suppose the push is extremly fast).  That will casue 4 * 10K ack, right?

With this limit, how long will it take for pilot to handle all all the ack, IIUC, it is 400s.

howardjohn:
This limit is on new XDS connections, it's not on the push requests. ACK is
not impacted at all?

hzxuzhonghu:
Ah, sorry to be mistaken.  I would approve.

hzxuzhonghu (approved):

ramaraochavali:
One of things we are experimenting internally is to use https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/connection_limit/v3/connection_limit.proto.html?highlight=connection%20limit%20filter on istiod's proxy and limit number of connections? This would not only help thundering herd problem with either controlplane restarts or mass proxy restarts and seems to potentially help with even load balancing of proxies across istiod instances.  But the only challenge I am seeing with that is how do we come up with that max number when the cluster is growing.

ramaraochavali:
LGTM

pilot/pkg/features/pilot.go
```diff
@@ -64,6 +64,12 @@ var (
 		"Limits the number of concurrent pushes allowed. On larger machines this can be increased for faster pushes",
 	).Get()
 
+	RequestLimit = env.RegisterFloatVar(
+		"PILOT_MAX_REQUESTS_PER_SECOND",
+		100.0,
```
nrjpoddar, line 69: How was the default decided?

howardjohn:
Rama - we don't have an Envoy in front of Istiod though in the standard
install. Otherwise we could push this there.

Neeraj - we did some testing on our side on a large config and smallish
istiod instances. In our tests we found 25qps was the sweet spot to prevent
OOM while also allowing a reasonable speed of new connection. However, I
was worried about making such a small default since users may have massive
istiod instances with 64cpu that can handle more. As a result, I was very
conservative here and picked 100.

PR #35047, opened 2021-09-07
Configure revision tags through base

Monkeyanator:
Requirement for revision tags alpha promotion. Lets users configure revision tags through base like:

```yaml
base:
  tags:
    default:
      revision: "1-10-3"
      injectionURL: "https://istiod.inject.io"
    canary:
      revision: "1-11-0"
```

Moves revision tags from the istio-discovery chart into base and removes the istio-default chart. This lends itself to more natural Helm usage since users can do `helm upgrade istio-base` with the new tag values to change them.

A few unclear points:

1. Helm users had to set `--set defaultRevision=true` at install to get validation previously. Now, Helm users will have to set the uglier `--set base.tags.default.revision=default` to get the same behavior
2. Not specific to this PR, but I'm starting to wonder if we should ever install `istio-injection=enabled` webhooks outside of the default tag.

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


manifests/charts/base/templates/mutatingwebhooks.yaml
```diff
@@ -34,13 +31,13 @@ metadata:
...
     istio.io/tag: {{ $tagName }}
-    istio.io/rev: {{ $.Values.revision | default "default" }}
+    istio.io/rev: {{ $tag.revision | default "default" }}
     install.operator.istio.io/owning-resource: {{ $.Values.ownerName | default "unknown" }}
     operator.istio.io/component: "Pilot"
     app: sidecar-injector
     release: {{ $.Release.Name }}
 webhooks:
-{{- include "core" (mergeOverwrite (deepCopy $) (dict "Prefix" "rev.namespace.") ) }}
+{{- include "core" (mergeOverwrite (deepCopy $) (deepCopy $tag) (dict "Prefix" "rev.namespace.") ) }}
```
Monkeyanator, line 40: these are getting really ugly

PR #35048, opened 2021-09-07
istioctl workload sets discovery address for revision

stevenctl:
If the revision is set, we try to use the same discovery address that we setup in /etc/hosts

fixes https://github.com/istio/istio/issues/34058


howardjohn (commented):
Can we add a integ test?

howardjohn (approved):

shamsher31 (approved):

PR #35052, opened 2021-09-07
Add MCS ClusterSet IPs to Service

nmittler:
This is the first step toward adding support for the Kubernetes Multi-Cluster Services (MCS) ClusterSet host (i.e. name.ns.svc.clusterset.local).

Overview of changes:

- New `ServiceImportCache` that reads `ServiceImport` and adds the ClusterSet IPs to cached `model.Service`.
- Updated aggregate controller to merge ClusterSet IPs from various clusters.
- Updated service conversion to populate ClusterSet host and IPs.

Changes to XDS processing will be made in a follow-on PR in an attempt to aid review.


pilot/pkg/serviceregistry/aggregate/controller.go
```diff
@@ -177,7 +177,17 @@ func mergeService(dst, src *model.Service, srcRegistry serviceregistry.Instance)
...
 	if srcRegistry.Provider() == provider.Kubernetes || len(dst.ClusterLocal.ClusterVIPs.GetAddressesFor(clusterID)) == 0 {
-		dst.ClusterLocal.ClusterVIPs.SetAddressesFor(clusterID, []string{src.Address})
+		dst.ClusterLocal.ClusterVIPs.SetAddressesFor(clusterID, src.ClusterLocal.ClusterVIPs.GetAddressesFor(clusterID))
+	}
+	if srcRegistry.Provider() == provider.Kubernetes {
+		// Copy over Kubernetes Multi-Cluster Services (MCS) ClusterSet Host and IPs.
+		if len(dst.ClusterSetLocal.Hostname) == 0 {
+			dst.ClusterSetLocal.Hostname = src.ClusterSetLocal.Hostname
+		}
+		if len(dst.ClusterSetLocal.ClusterVIPs.GetAddressesFor(clusterID)) == 0 {
```
howardjohn, line 187: I am not sure I understand how this can be false or why we need it?

pilot/pkg/serviceregistry/aggregate/controller.go
```diff
@@ -177,7 +177,17 @@ func mergeService(dst, src *model.Service, srcRegistry serviceregistry.Instance)
 	// prefer the k8s VIP where possible
 	clusterID := srcRegistry.Cluster()
 	if srcRegistry.Provider() == provider.Kubernetes || len(dst.ClusterLocal.ClusterVIPs.GetAddressesFor(clusterID)) == 0 {
-		dst.ClusterLocal.ClusterVIPs.SetAddressesFor(clusterID, []string{src.Address})
+		dst.ClusterLocal.ClusterVIPs.SetAddressesFor(clusterID, src.ClusterLocal.ClusterVIPs.GetAddressesFor(clusterID))
+	}
+	if srcRegistry.Provider() == provider.Kubernetes {
+		// Copy over Kubernetes Multi-Cluster Services (MCS) ClusterSet Host and IPs.
+		if len(dst.ClusterSetLocal.Hostname) == 0 {
```
howardjohn, line 184: How come we need hostname here but not for L180 ClusterLocal?

howardjohn (commented):
Over LGTM. Not hitting approve since I had a few tiny nits and wanted to give others a chance

pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
```diff
@@ -0,0 +1,156 @@
...
+		}: {}},
+		Reason: []model.TriggerReason{model.ServiceUpdate},
+	}
+	ic.opts.XDSUpdater.ConfigUpdate(pushReq)
+}
+
+func (ic *serviceImportCacheImpl) GetClusterSetIPs(name types.NamespacedName) []string {
+	si, err := ic.lister.ServiceImports(name.Namespace).Get(name.Name)
+	if err == nil && si.Spec.Type == mcsCore.ClusterSetIP && len(si.Spec.IPs) > 0 {
+		return si.Spec.IPs
```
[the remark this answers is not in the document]
nmittler: I actually added that in the previous PR based on @jacob-delgado's description of their work on ipv6 support. It was only a partial implementation to just change the structure to an array, but I didn't do all the plumbing to make it work.

pilot/pkg/model/service.go
```diff
@@ -56,6 +56,15 @@ type HostVIPs struct {
 	ClusterVIPs cluster.AddressMap `json:"clusterVIPs,omitempty"`
 }
 
+func (h *HostVIPs) DeepCopy() HostVIPs {
+	return HostVIPs{
+		Hostname: h.Hostname,
+		ClusterVIPs: cluster.AddressMap{
```
hzxuzhonghu, line 62: There is `AddressMap.DeepCopy`

pilot/pkg/serviceregistry/aggregate/controller.go
hzxuzhonghu: same doubt

pilot/pkg/serviceregistry/aggregate/controller.go
hzxuzhonghu: I guess this is copied from cluster local logic, but the logic is provider = k8s || address is enpty there.

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -444,7 +450,8 @@ func (c *Controller) onServiceEvent(curr interface{}, event model.Event) error {
 
 	log.Debugf("Handle event %s for service %s in namespace %s", event, svc.Name, svc.Namespace)
 
-	svcConv := kube.ConvertService(*svc, c.opts.DomainSuffix, c.Cluster())
+	svcConv := kube.ConvertService(*svc, c.opts.DomainSuffix, c.Cluster(),
+		c.imports.GetClusterSetIPs(kube.NamespacedNameForK8sObject(svc)))
```
hzxuzhonghu, line 454: What happens when serviceImport does not exist when running this?

pilot/pkg/serviceregistry/kube/conversion.go
```diff
@@ -183,11 +198,30 @@ func ExternalNameServiceInstances(k8sSvc *coreV1.Service, svc *model.Service) []
...
+}
+
 // ServiceHostname produces FQDN for a k8s service
+// TODO(nmittler): Rename this to ServiceClusterLocalHostname to differentiate with ServiceClusterSetLocalHostname.
 func ServiceHostname(name, namespace, domainSuffix string) host.Name {
 	return host.Name(name + "." + namespace + "." + "svc" + "." + domainSuffix) // Format: "%s.%s.svc.%s"
 }
 
+// ServiceHostnameForKR calls ServiceHostname with the name and namespace of the given kubernetes resource.
+func ServiceHostnameForKR(obj metav1.Object, domainSuffix string) host.Name {
```
hzxuzhonghu, line 216: I suggest calling `ServiceHostname` directly, this may cause misuse

pilot/pkg/serviceregistry/kube/conversion.go
```diff
@@ -183,11 +198,30 @@ func ExternalNameServiceInstances(k8sSvc *coreV1.Service, svc *model.Service) []
...
+	return ServiceHostname(obj.GetName(), obj.GetNamespace(), domainSuffix)
+}
+
+// ServiceClusterSetLocalHostname produces Kubernetes Multi-Cluster Services (MCS) ClusterSet FQDN for a k8s service
+func ServiceClusterSetLocalHostname(nn types.NamespacedName) host.Name {
```
hzxuzhonghu, line 221: NamespacedName causes an extra conversion

pilot/pkg/xds/debug.go
```diff
@@ -204,6 +204,7 @@ func (s *DiscoveryServer) AddDebugHandlers(mux, internalMux *http.ServeMux, enab
 	s.addDebugHandler(mux, internalMux, "/debug/clusterz", "List remote clusters where istiod reads endpoints", s.clusterz)
 	s.addDebugHandler(mux, internalMux, "/debug/networkz", "List cross-network gateways", s.networkz)
 	s.addDebugHandler(mux, internalMux, "/debug/exportz", "List endpoints that been exported via MCS", s.exportz)
+	s.addDebugHandler(mux, internalMux, "/debug/importz", "List endpoints that been exported via MCS", s.importz)
```
hzxuzhonghu, line 207: Wrong description

pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
```diff
@@ -0,0 +1,156 @@
...
+			Namespace: si.Namespace,
+		}: {}},
+		Reason: []model.TriggerReason{model.ServiceUpdate},
+	}
+	ic.opts.XDSUpdater.ConfigUpdate(pushReq)
+}
+
+func (ic *serviceImportCacheImpl) GetClusterSetIPs(name types.NamespacedName) []string {
+	si, err := ic.lister.ServiceImports(name.Namespace).Get(name.Name)
+	if err == nil && si.Spec.Type == mcsCore.ClusterSetIP && len(si.Spec.IPs) > 0 {
```
hzxuzhonghu, line 117: IMO, you donot need so many conditions here, only  ClusterSetIP type can have ips set
if si !=nil {
 return si.Spec.IPs
}

hzxuzhonghu (commented):
After adding serviceImport, i doubt do we really need serviceExport controller simultaneously. In my view, the function can be satisfied by serviceImport too.

hzxuzhonghu:
What's more, who is responsible for generating endpointslice for serviceImport/serviceExport

pilot/pkg/serviceregistry/kube/conversion.go
nmittler: What sort of misuse are you referring to?

pilot/pkg/serviceregistry/kube/conversion.go
nmittler: Yes, but it's pretty minor and significantly reduces chances of accidental misuse of the method when passing in two strings.

nmittler:
@hzxuzhonghu 

> After adding serviceImport, i doubt do we really need serviceExport controller simultaneously. In my view, the function can be satisfied by serviceImport too.

We actually need both. The `ServiceImport` is only used for getting the clusterset IPs. We're using `ServiceExport` for everything else (i.e. determining export state of service). We have to read all services anyway for a variety of reasons, so there is no real advantage currently to driving service discovery from `ServiceImport`/`EndpointSlice` exclusively. For example, the MCS API does not currently support multi-network use cases.

pilot/pkg/serviceregistry/kube/controller/controller.go
nmittler: It will get a `nil` array.

pilot/pkg/model/service.go
nmittler: Ah right ... good catch.

pilot/pkg/serviceregistry/aggregate/controller.go
nmittler: Yeah that's right. Dst starts out empty and we just add to it. So the logic is effectively, if it hasn't been written to, write to it ... or if it's coming from a k8s cluster, always overwrite. The comment at the top I think is the key ... we always prefer vips from k8s.

pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
nmittler: done.

pkg/config/constants/constants.go
```diff
@@ -60,8 +60,14 @@ const (
 	IstioIngressNamespace = "istio-system"
 
 	// DefaultKubernetesDomain the default service domain suffix for Kubernetes, if not overridden in config.
+	// TODO(nmittler): Rename this to DefaultClusterLocalDomain.
+	// TODO(nmittler): Search/replace explicit usages of the string with this constant.
 	DefaultKubernetesDomain = "cluster.local"
 
+	// DefaultClusterSetLocalDomain is the default domain suffix for Kubernetes Multi-Cluster Services (MCS)
+	// used for load balancing requests against endpoints across the ClusterSet (i.e. mesh).
+	DefaultClusterSetLocalDomain = "clusterset.local"
```
howardjohn, line 69: very nit/question: is it *default* or is it always clusterset.local in the MCS spec?

howardjohn (approved):

pkg/config/constants/constants.go
nmittler: Always via the spec, but you can always provide an alias.

On a related note, I think we should technically be supporting `cluster.local` even when the user specifies `domainSuffix`.  We should follow this up with a bit of a re-think on `domainSuffix` as a whole.

hzxuzhonghu:
>>What's more, who is responsible for generating endpointslice for serviceImport/serviceExport

>An MCS controller. It's also responsible for generating ServiceImport.

Then with it, maybe istio does not need to watch multiple clusters. And get service and endpoints from ServiceImport and EndpointSlice.   So i understand the serviceExport is used by an MCS controller, which could be an separate component or one that resides in istio?

nmittler:
@hzxuzhonghu 

> Then with it, maybe istio does not need to watch multiple clusters. And get service and endpoints from ServiceImport and EndpointSlice. So i understand the serviceExport is used by an MCS controller, which could be an separate component or one that resides in istio?

That was our hope a while back, but it's just not possible given the current state of the spec. See the discussion of [gaps in the MCS API](https://docs.google.com/document/d/125JiDxXjvqHXu_678mUXKy3pKOvfSw4zKwXJjulU_ug/edit?resourcekey=0-cwuRNzqukgPyk0ILC2pxMw#heading=h.y5alospizrt4).

PR #35059, opened 2021-09-08
terminate envoy when number of active connections is zero

ramaraochavali:
For #34855 except that it still uses terminationDrainDuration.

It gets downstream_cx_active of all listeners (ignores admin and stats listeners) and when the total becomes zero, it terminates Envoy or waits for terminationDrainDuration. It starts this process after waiting for minDrainDuration i.e. 5s.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/cmd/pilot-agent/options/options.go
+++ b/pilot/cmd/pilot-agent/options/options.go
@@ -119,4 +119,9 @@ var (
 	istiodSAN = env.RegisterStringVar("ISTIOD_SAN", "",
 		"Override the ServerName used to validate Istiod certificate. "+
 			"Can be used as an alternative to setting /etc/hosts for VMs - discovery address will be an IP:port")
+
+	minimumDrainDurationEnv = env.RegisterDurationVar("MINIMUM_DRAIN_DURATION",
+		5*time.Second,
+		"The minimum duration for which agent waits before it checks for active connections and terminates proxy"+
+			"when number of active connections become zero").Get()
 )
--- a/pkg/envoy/agent.go
+++ b/pkg/envoy/agent.go
@@ -17,8 +17,13 @@ package envoy
 import (
 	"context"
 	"errors"
+	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
+	"istio.io/istio/pilot/pkg/util/sets"
+	"istio.io/istio/pkg/http"
 	"istio.io/pkg/log"
 )
 
@@ -26,13 +31,30 @@ var errAbort = errors.New("epoch aborted")
 
 const errOutOfMemory = "signal: killed"
 
+var activeConnectionCheckDelay = time.Duration(1 * time.Second)
+
 // NewAgent creates a new proxy agent for the proxy start-up and clean-up functions.
-func NewAgent(proxy Proxy, terminationDrainDuration time.Duration) *Agent {
+func NewAgent(proxy Proxy, terminationDrainDuration, minDrainDuration time.Duration, localhost string,
+	adminPort, statusPort, prometheusPort int) *Agent {
+
+	knownIstioListeners := sets.NewSet(
+		fmt.Sprintf("listener.0.0.0.0_%d.downstream_cx_active", statusPort),
+		fmt.Sprintf("listener.0.0.0.0_%d.downstream_cx_active", prometheusPort),
+		"listener.admin.downstream_cx_active",
+		"listener.admin.main_thread.downstream_cx_active",
+	)
 	return &Agent{
 		proxy:                    proxy,
 		statusCh:                 make(chan exitStatus, 1), // context might stop drainage
+		drainCh:                  make(chan struct{}),
 		abortCh:                  make(chan error, 1),
 		terminationDrainDuration: terminationDrainDuration,
+		minDrainDuration:         minDrainDuration,
+		adminPort:                adminPort,
+		statusPort:               int(statusPort),
+		prometheusPort:           int(prometheusPort),
+		localhost:                localhost,
+		knownIstioListeners:      knownIstioListeners,
 	}
 }
 
@@ -59,10 +81,21 @@ type Agent struct {
 	// channel for proxy exit notifications
 	statusCh chan exitStatus
 
+	drainCh chan struct{}
+
 	abortCh chan error
 
 	// time to allow for the proxy to drain before terminating all remaining proxy processes
 	terminationDrainDuration time.Duration
+	minDrainDuration         time.Duration
+
+	adminPort int
+	localhost string
+
+	statusPort     int
+	prometheusPort int
+
+	knownIstioListeners sets.Set
 }
 
 type exitStatus struct {
@@ -105,11 +138,69 @@ func (a *Agent) terminate() {
 	if e != nil {
 		log.Warnf("Error in invoking drain listeners endpoint %v", e)
 	}
-	log.Infof("Graceful termination period is %v, starting...", a.terminationDrainDuration)
-	time.Sleep(a.terminationDrainDuration)
-	log.Infof("Graceful termination period complete, terminating remaining proxies.")
-	a.abortCh <- errAbort
+	log.Infof("Agent draining proxy for %v, then waiting for active connections to terminate...", a.minDrainDuration)
+	time.Sleep(a.minDrainDuration)
+	log.Infof("Termination drain duration period is %v, checking for active connections...", a.terminationDrainDuration)
+	go a.waitForDrain()
+	select {
+	case <-a.drainCh:
+		log.Info("There are no more active connections. terminating proxy...")
+		a.abortCh <- errAbort
+	// TODO: remove terminationDrainDuration and rely on "terminationGracefulPeriodSeconds" of pod?
+	case <-time.After(a.terminationDrainDuration):
+		log.Info("Termination period complete, terminating proxy...")
+		a.abortCh <- errAbort
+		close(a.drainCh)
+	}
 	log.Warnf("Aborted all epochs")
+}
+
+func (a *Agent) waitForDrain() {
+	for {
+		select {
+		case <-time.After(activeConnectionCheckDelay):
+			if a.activeProxyConnections() == 0 {
+				close(a.drainCh)
+				return
+			}
+		case <-a.drainCh:
+			return
+		}
+	}
+}
+
+func (a *Agent) activeProxyConnections() int {
+	activeConnectionsURL := fmt.Sprintf("http://%s:%d/stats?usedonly&filter=downstream_cx_active", a.localhost, a.adminPort)
+	stats, err := http.DoHTTPGet(activeConnectionsURL)
+	if err != nil {
+		log.Warnf("Unable to get listener stats from Envoy : %v", err)
+		return -1
+	}
+	activeConnections := 0
+	for stats.Len() > 0 {
+		line, _ := stats.ReadString('\n')
+		parts := strings.Split(line, ":")
+		if len(parts) != 2 {
+			log.Warnf("envoy stat line is missing separator. line:%s", line)
+			continue
+		}
+		// downstream_cx_active is accounted under "http." and "listener." for http listeners.
+		// Only consider listener stats.
+		if !strings.HasPrefix(parts[0], "listener.") {
+			continue
+		}
+		// If the stat is for a known Istio listener skip it.
+		if a.knownIstioListeners.Contains(parts[0]) {
+			continue
+		}
+		val, err := strconv.ParseUint(strings.TrimSpace(parts[1]), 10, 64)
+		if err != nil {
+			log.Warnf("failed parsing Envoy stat %s (error: %s) line: %s", parts[0], err.Error(), line)
+			continue
+		}
+		activeConnections += int(val)
+	}
+	return activeConnections
 }
 
 // runWait runs the start-up command as a go routine and waits for it to finish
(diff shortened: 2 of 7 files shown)


pkg/envoy/agent.go
```diff
@@ -105,13 +138,71 @@ func (a *Agent) terminate() {
...
+				return
+			}
+		case <-a.drainCh:
+			return
+		}
+	}
+}
+
+func (a *Agent) activeProxyConnections() int {
+	activeConnectionsURL := fmt.Sprintf("http://%s:%d/stats?usedonly&filter=downstream_cx_active", a.localhost, a.adminPort)
```
howardjohn, line 173: very very nit: IIRC regex performance here has a big impact. It is only on shutdown so not a huge deal, but does adding things like `downstream_cx_active$` make it cheaper?

howardjohn: I recall something in the past replacing a `.` with a `\.` or similar made it like 4x faster or something crazy is all.

pkg/envoy/agent.go
```diff
@@ -105,13 +138,71 @@ func (a *Agent) terminate() {
...
+	stats, err := http.DoHTTPGet(activeConnectionsURL)
+	if err != nil {
+		log.Warnf("Unable to get listener stats from Envoy : %v", err)
+		return -1
+	}
+	activeConnections := 0
+	for stats.Len() > 0 {
+		line, _ := stats.ReadString('\n')
```
howardjohn, line 181: If we are adding custom parsing we should add some unit tests before merging. Aside from testing the code it also shows an example of what the inputs we are expecting to parse look like

pkg/envoy/agent.go
```diff
@@ -105,13 +138,71 @@ func (a *Agent) terminate() {
...
+				close(a.drainCh)
+				return
+			}
+		case <-a.drainCh:
+			return
+		}
+	}
+}
+
+func (a *Agent) activeProxyConnections() int {
```
howardjohn, line 172: Is it possible users are excluding `downstream_cx_active` stats? I actually am not seeing it at all in a default install...

ramaraochavali: yes. We will have to add them to default inclusion

pkg/envoy/agent.go
ramaraochavali: Will take a look

pkg/envoy/agent.go
ramaraochavali: yes. We need to add tests. I will add before I mark this PR ready.

ramaraochavali:
@howardjohn PTAL when you get chance

pkg/envoy/agent.go
howardjohn: It seems like we haven't done that. And I am a bit concerned about adding `downstream_cx_active` for *all* listeners. Can we just add it for the 2 listeners we need it for?

hzxuzhonghu:
cc @kyessenov

pkg/envoy/agent.go
ramaraochavali: I have not added it yet. I will do that. Yes. We can add it for two listeners

pkg/bootstrap/config.go
```diff
@@ -55,6 +55,8 @@ const (
 
 	rbacEnvoyStatsMatcherInclusionSuffix = "rbac.allowed,rbac.denied,shadow_allowed,shadow_denied"
 
+	requiredEnvoyStatsMatcherInclusionSuffixes = rbacEnvoyStatsMatcherInclusionSuffix + "downstream_cx_active" // Needed for draining.
```
howardjohn, line 58: @mandarjog @kyessenov ptal

howardjohn:
```suggestion
	requiredEnvoyStatsMatcherInclusionSuffixes = rbacEnvoyStatsMatcherInclusionSuffix + ",downstream_cx_active" // Needed for draining.
```

?

pkg/envoy/agent_test.go
```diff
@@ -127,10 +164,46 @@ func TestRecovery(t *testing.T) {
...
+func TestActiveConnections(t *testing.T) {
+	cases := []struct {
+		name     string
+		stats    string
+		expected int
+	}{
+		{
+			"invalid stats",
+			invalidStats,
+			-1,
```
howardjohn, line 184: If someone disables all stats we should have sane behavior. Probably the old behavior? wdyt?

pkg/envoy/agent.go
mandarjog: We don’t have that many inbound listeners so adding this metric is fine.
We were considering adding downstream connection metrics for a long time anyway.
It is especially useful at ingress.

mandarjog: Can you check active vs open, and make sure it captures the correct quantity as the name implies.

pkg/envoy/agent_test.go
mandarjog: Either that, or we add this to the stats that are always and automatically added.
I think having these stats is important anyway.
Several users enable these in production.

pilot/cmd/pilot-agent/options/options.go
```diff
@@ -119,4 +119,9 @@ var (
 	istiodSAN = env.RegisterStringVar("ISTIOD_SAN", "",
 		"Override the ServerName used to validate Istiod certificate. "+
 			"Can be used as an alternative to setting /etc/hosts for VMs - discovery address will be an IP:port")
+
+	minimumDrainDurationEnv = env.RegisterDurationVar("MINIMUM_DRAIN_DURATION",
+		5*time.Second,
```
mandarjog, line 124: May be covered elsewhere, why is this an env var at all?
Is it controlling the “wait for connection termination “ behaviour or just when it starts polling for connection count?
If this time is 0 then we will start checking for connections right away which does not sound too bad either. Do we really need this var?

pkg/envoy/agent_test.go
ramaraochavali: I have added them by default. Some one can not disable them I guess.

pkg/envoy/agent.go
ramaraochavali: If you mean total vs active - I verified. It captures the correct quantity

pilot/cmd/pilot-agent/options/options.go
ramaraochavali: > or just when it starts polling for connection count?

This is when It starts polling for connection count. This is to avoid race condition described here https://github.com/envoyproxy/envoy/issues/17056#issuecomment-886267587

mandarjog: If we are worried about that race then we should not let users set this env var to 0. at what time do we deem that the race will not be there? I am still wondering about keeping this configurable. Again in remote istiod this is not configurable so it is important to just set the right value.

howardjohn: Proxy env vars are configurable regardless of where istiod runs

mandarjog: Ok my bad, I was reading this as a pilot env var.

How did we arrive at 5, and what process should a user go thru to decide on another value ?

ramaraochavali: > How did we arrive at 5, and what process should a user go thru to decide on another value ?

5s was chosen just as we felt it is reasonable. We do not expect this to be tuned by users. The reason why we provided env var is so that users can set it to 0  - to get back the old behaviour.

howardjohn (commented):
I was thinking this through a bit...
Current we have two knobs:
* Termination drain, default 5s. Users override this to 0 for fast exit (not super common? I do it though...) and to set it higher to avoid premature exit. 
* terminationGracePeriodSeconds (k8s), default 30s. Users override for fast exit or higher for slow exit.

Largely these two knobs are the same.

This PR config:
* minimumDrainDurationEnv - 5s; always wait this time
* exit after min(all connections close, terminationDrainDuration)

So the default is 5s + up to 5s.

I think there are a few use cases for configuration:
* I want the proxy to never exit before my app (primary case)
    * We cannot do this perfectly, so there are two subcases:
    * My app keeps all connections open until it exits (common case)
    * My app does a bunch of shutdown work, then at the last second opens a connection to save data somewhere
* I want to exit ASAP, no waiting

Possible alternatives:

1. Optimize for backwards compatibility of terminationDrainDuration:
* If terminationDrainDuration is configured, use the old behavior
* If terminationDrainDuration is not configured, always sleep minimumDrainDurationEnv (5s) then exit after min(all connections close, terminationGracePeriodSeconds-minimumDrainDurationEnv)

This satisfies all use cases, with some config. For "exit ASAP", users need to set minimumDrainDurationEnv=true and terminationGracePeriodSeconds=0s. For "last minute write" cases, they need to set minimumDrainDurationEnv=some-high-number. For common case, it works OOTB.

The primary benefit here is backwards compatibility is 100% preserved.

---

2. Optimize for respecting small terminationDrainDuration:
* Always sleep terminationDrainDuration
* Exit after min(connections close, terminationDrainDuration)

For terminationDrainDuration=0, we exit immediately. For terminationDrainDuration>0, we exit sometime between terminationDrainDuration-2*terminationDrainDuration

This satisfies "exit ASAP" but not "last minute write" or "exit after app" (just makes the time 2x for these cases).

---

3. Optimize for respecting large terminationDrainDuration (this PR):
* Always sleep minimumDrainDurationEnv (5s)
* Exit after min(connections close, terminationDrainDuration-5s)

We respect users desire to be *able* to stay alive for terminationDrainDuration, but do not satisfy cases "exit ASAP" or "last minute write"

----

Overall no option is perfect. IMO, `Optimize for backwards compatibility of terminationDrainDuration` seems like a good option though since we don't want to break users on upgrade.

I have also asked knative team to give feedback since they currently set terminationDrainDuration=600s

ramaraochavali:
I am fine with changing this `Optimize for backwards compatibility of terminationDrainDuration` - Let us know the feedback of knative folks is.

ramaraochavali:
@howardjohn added option 1. PTAL

pilot/cmd/pilot-agent/options/options.go
jacob-delgado: > 5s was chosen just as we felt it is reasonable. We do not expect this to be tuned by users. The reason why we provided env var is so that users can set it to 0 - to get back the old behaviour.

The comment for this env should reflect the user can change it to 0 if they wish to get back the old behavior.

ramaraochavali: @howardjohn with the change to the new approach do you think this is still needed or we can just internally hard code it to `5s`?

ramaraochavali:
@howardjohn can you PTAL when you get chance?

ramaraochavali:
@howardjohn Please take a look when you get chance - no rush.

howardjohn:
@ramaraochavali sorry for the delay - there was some discussion about an alternative option that @costinm was working on that would also solve the startup ordering.

I think likely the best option is to have both as experimental in 1.12 (if Costin's makes it in time) - no need to block one experimental feature for another, and we can compare how effective they are.

howardjohn (approved):

kyessenov:
Seems like my comment was lost. What do we do if application waits to dump its state? It's common to have to not open connections until the very last step when its logs / buffer are sent. Waiting for 0 connections will break those use cases.

howardjohn:
Put a hold for ^

I would agree this does not solve that use case which is a valid and somewhat common use case. However, it also doesn't make it _worse_ I think. Both before and after this we always wait 5s minimum; now we just may wait longer.

kyessenov:
I don't mind it as an experiment. But definitely not something on by default. The failure is too subtle, and it's really can cause some damage to some customer who is not careful.

howardjohn:
> I don't mind it as an experiment. But definitely not something on by default. The failure is too subtle, and it's really can cause some damage to some customer who is not careful.

I am not against making it off by default, probably is a safer option at least for now. What damage do you expect though? My expectation is this will only ever make a pod taker longer to shutdown, not shorter?

kyessenov:
As I mentioned any app that waits to open connection to send last bits of telemetry/logs will not have a proxy. JVM apps can take up to 2 minutes to shut down. So if a large app waits a minute before sending data (or just gets interrupted by GC), envoy is not there.

howardjohn:
I agree but don't see how this PR makes it worse.

Today: wait 5s and exit
This PR: wait 5s,then wait until connections close

In both cases slow java app is broken the same

kyessenov:
If you put it that way, yes. But no serious user should be setting drain duration to 5s. It should be at least 30s.

ramaraochavali:
So what do we want to do? Should I disable it by default?

ramaraochavali:
Added logic to control via flag.

rittneje:
@ramaraochavali @howardjohn Are these new properties configurable per pod via annotations? Or are they just pilot env vars? If the latter, this doesn't seem useful for us.

> My app keeps all connections open until it exits (common case)

@howardjohn I find this claim to be dubious. During pod shutdown, there may be some number of in-flight requests that need to be completed. These requests may require talking with other services (both inside and outside the cluster), and it is unreasonable to assume there already would be a connection. To be honest, until Kubernetes accounts for container dependencies, you really need a way for the main container to directly tell istio-proxy when it is done.

ramaraochavali:
> Are these new properties configurable per pod via annotations? Or are they just pilot env vars? If the latter, this doesn't seem useful for us

These are per pod metadata - not pilot properties

rittneje:
@ramaraochavali How exactly do you set them per pod? What are the metadata keys? I don't see anything in the docs.
https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/
https://istio.io/latest/docs/reference/config/annotations/

ramaraochavali:
@rittneje See https://github.com/istio/istio/issues/34855#issuecomment-951379884 on how to set those values per pod.

PR #35061, opened 2021-09-08
fix sidecar listener creation with bind and catchall in sidecar

ramaraochavali:
When we have the following sidecar
```
spec:
  egress:
  - bind: 0.0.0.0
    hosts:
    - '*/test.testns.svc.cluster.local'
    port:
      name: redis-cluster
      number: 12000
      protocol: REDIS
   - hosts:
    - '*/*'
```
with a port and bind with a non HTTP protocol and a wildcard catchall - we end up creating listener on wildcard because of bind and service address because listener key does not match for TCP. This PR fixes it.

- [ ] Docs
- [ ] Installation
- [X ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/listener.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener.go
@@ -841,7 +841,12 @@ func (configgen *ConfigGeneratorImpl) buildSidecarOutboundTCPListenerOptsForPort
 	// in the map.
 	//
 	// Check if this TCP listener conflicts with an existing HTTP listener
-	if *currentListenerEntry, exists = listenerMap[*listenerMapKey]; exists {
+	*currentListenerEntry, exists = listenerMap[*listenerMapKey]
+
+	if !exists {
+		*currentListenerEntry, exists = listenerMap[listenerKey(actualWildcard, listenerOpts.port.Port)]
+	}
+	if exists {
 		// NOTE: This is not a conflict. This is simply filtering the
 		// services for a given listener explicitly.
 		// When the user declares their own ports in Sidecar.egress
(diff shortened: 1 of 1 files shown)


pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -841,7 +841,20 @@ func (configgen *ConfigGeneratorImpl) buildSidecarOutboundTCPListenerOptsForPort
 	// in the map.
 	//
 	// Check if this TCP listener conflicts with an existing HTTP listener
-	if *currentListenerEntry, exists = listenerMap[*listenerMapKey]; exists {
+	*currentListenerEntry, exists = listenerMap[*listenerMapKey]
+
+	// If there is no listener with serviceListenAddress, check if a wildcard bind exists.
+	// This can happen if user specified a restricted service set on this port with wildcard
+	// bind and also adds a catchall('*/*' with no ports) listener. We should not create
+	// listener with serviceListenAddress in that case.
```
howardjohn, line 849: I think I understand what this PR does, but I am not sure I understand why. What is wrong with creating a listener for

```
0.0.0.0_12000
<VIP>_12000
```
?

it seems like it should have no impact? is it just some perf optimization?

ramaraochavali: Yes. In one of the usecase, we have 20 services on the same port but we want only one service with wildcard bind so we added sidecar and now if I add '*/*' at the end - it creates 20 additional listeners.

ramaraochavali:
@howardjohn @hzxuzhonghu what do you think about this? Is this reasonable to restrict or should we just create both wildcard and ip listeners?

pilot/pkg/networking/core/v1alpha3/listener.go
howardjohn: I am a bit worried keeping the wildcard one will mean that we do not get customizations we may have otherwise gotten for the `<VIP>_12000` one

ramaraochavali: Sure. No issues.

PR #35078, opened 2021-09-09
Rename charts we will publish to help repo

howardjohn:
Before, this name was ~meaningless. With
https://github.com/istio/release-builder/pull/757, this will become the
"API" that users install under. We should use this as an opportunity to
rename.

UX will be:
```
helm install istio/base
helm install istio/istiod
helm install istio/gateway
helm install istio/cni
```
Note: this is a blocker for https://github.com/istio/release-builder/pull/757 to pass

### DIFF
--- a/manifests/charts/istio-cni/Chart.yaml
+++ b/manifests/charts/istio-cni/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-name: istio-cni
+name: cni
 version: 1.1.0
 description: Helm chart for istio-cni components
 keywords:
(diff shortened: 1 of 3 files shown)


Monkeyanator (approved):

Monkeyanator:
One question: what's the point of having the nested `istio-control/istiod`? Can we just move `istiod` to top-level?

howardjohn:
We can but it may impact users slightly. They will need to change their
`helm install` commands.

Once we move to a helm repo the folder structure won't matter (for users
using it) and it would be less impactful

PR #35080, opened 2021-09-09
fix analyze having the unexpected log to the stdout

monsteredc:
If no analyzing message is provided, the json formatter and the yaml formatter will still print a pair of brackets, like below:
```
istioctl analyze --use-kube=false --failure-threshold=Info -o json /Users/xiaopenghan/Documents/testIstio/bad.yaml
Error(s) adding files: 1 error occurred:
	* errors parsing content "/Users/xiaopenghan/Documents/testIstio/bad.yaml": 1 error occurred:
	* error processing /Users/xiaopenghan/Documents/testIstio/bad.yaml[0]: failed converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context



[]
No validation issues found when analyzing namespace: default (but 1 file could not be parsed).
```
which is inconsistent with the log format.

Just revise the code to not print the brackets if there is no message.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/istioctl/pkg/util/formatting/formatter_test.go
+++ b/istioctl/pkg/util/formatting/formatter_test.go
@@ -147,8 +147,8 @@ func TestFormatter_PrintEmpty(t *testing.T) {
 	g.Expect(logOutput).To(Equal(""))
 
 	jsonOutput, _ := Print(msgs, JSONFormat, false)
-	g.Expect(jsonOutput).To(Equal("[]"))
+	g.Expect(jsonOutput).To(Equal(""))
 
 	yamlOutput, _ := Print(msgs, YAMLFormat, false)
-	g.Expect(yamlOutput).To(Equal("[]\n"))
+	g.Expect(yamlOutput).To(Equal(""))
 }
--- a/istioctl/pkg/util/formatting/formatter_test.go
+++ b/istioctl/pkg/util/formatting/formatter_test.go
@@ -147,8 +147,8 @@ func TestFormatter_PrintEmpty(t *testing.T) {
 	g.Expect(logOutput).To(Equal(""))
 
 	jsonOutput, _ := Print(msgs, JSONFormat, false)
-	g.Expect(jsonOutput).To(Equal(""))
+	g.Expect(jsonOutput).To(Equal("[]"))
 
 	yamlOutput, _ := Print(msgs, YAMLFormat, false)
-	g.Expect(yamlOutput).To(Equal(""))
+	g.Expect(yamlOutput).To(Equal("[]\n"))
 }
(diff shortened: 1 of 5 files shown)


shamsher31 (commented):
LGTM, please fix the test.

istioctl/pkg/util/formatting/formatter_test.go
```diff
@@ -147,8 +147,8 @@ func TestFormatter_PrintEmpty(t *testing.T) {
 	g.Expect(logOutput).To(Equal(""))
 
 	jsonOutput, _ := Print(msgs, JSONFormat, false)
-	g.Expect(jsonOutput).To(Equal("[]"))
+	g.Expect(jsonOutput).To(Equal(""))
```
howardjohn, line 150: I don't understand? `[]` is valid JSON for an empty list. The result is an empty list. Why do we not want to output the result?

monsteredc: Yes it's valid JSON format. What I was trying to do is to make all three output formats have the same output results if there is no message needed to be printed. The original results will be below:
```
➜  testIstio cat bad.yaml | istioctl analyze --use-kube=false -olog -
Error(s) adding files: 1 error occurred:
	* errors parsing content "-": 1 error occurred:
	* error processing -[0]: failed converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context




No validation issues found when analyzing namespace: default (but 1 file could not be parsed).
Error: Some files couldn't be parsed.
➜  testIstio cat bad.yaml | istioctl analyze --use-kube=false -ojson -
Error(s) adding files: 1 error occurred:
	* errors parsing content "-": 1 error occurred:
	* error processing -[0]: failed converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context



[]
No validation issues found when analyzing namespace: default (but 1 file could not be parsed).
➜  testIstio cat bad.yaml | istioctl analyze --use-kube=false -oyaml -
Error(s) adding files: 1 error occurred:
	* errors parsing content "-": 1 error occurred:
	* error processing -[0]: failed converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context



[]

No validation issues found when analyzing namespace: default (but 1 file could not be parsed).
```
The JSON format will print an extra string `[]` and the YAML format will print `[]\n`.

howardjohn: > What I was trying to do is to make all three output formats have the same output results if there is no message needed to be printed.

I don't think this is a good goal. JSON output is for machines. Machines like the `[]`.

monsteredc: Alright I see. Thanks

PR #35084, opened 2021-09-09
support IP verification and de duplication

zc2638:

support `proxy.IPAddresses` verification and de duplication

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn (changes requested):
Can you help explain what bug is being fixed by this? Thanks!

zc2638:
> Can you help explain what bug is being fixed by this? Thanks!

Hello, this is not a bug. 
On the original basis, it supports IP repeat filtering, and makes the code look more lovely.

howardjohn:
I would strongly prefer not to refactor code that has been stable for 3
years. It introduced substantial risk ina  critical code path and offers no
benefit to users

zc2638:
`network.getprivateips` will obtain all IP addresses on the node. 
If the IP addresses of multiple network cards are the same, duplicate IP addresses will be generated.
Although this network configuration may not be common, we should avoid it.

howardjohn:
https://serverfault.com/questions/705919/the-same-ip-on-multiple-interfaces#:~:text=You%20cannot%20use%20the%20same,address%20on%20the%20bridge%20itself.

Maybe we can do this if we see a real world case where this would help? Otherwise I am worried about making changes that do not add benefit but add risk.

PR #35087, opened 2021-09-09
remove port double check

zc2638:

remove port double check


hzxuzhonghu (approved):

howardjohn (commented):
I feel this actually makes the code much less readable. I would encourage you to focus on fixing bugs impacting users instead of refactorings.

PR #35103, opened 2021-09-09
Fuzzing: Add missing `-i` in sed command in oss-fuzz build script

AdamKorcz:
Applies the sed command instead of just printing the result out.

### DIFF
--- a/tests/fuzz/oss_fuzz_build.sh
+++ b/tests/fuzz/oss_fuzz_build.sh
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-sed 's/out.initJwksResolver()/\/\/out.initJwksResolver()/g' "${SRC}"/istio/pilot/pkg/xds/discovery.go
+sed -i 's/out.initJwksResolver()/\/\/out.initJwksResolver()/g' "${SRC}"/istio/pilot/pkg/xds/discovery.go
 
 compile_go_fuzzer istio.io/istio/tests/fuzz FuzzParseInputs fuzz_parse_inputs
 compile_go_fuzzer istio.io/istio/tests/fuzz FuzzParseAndBuildSchema fuzz_parse_and_build_schema


ericvn (changes requested):
This change won't work on a Mac, as you need to specify the extension after the -i. I did this same thing recently at https://github.com/istio/common-files/blob/master/bin/update-build-image.sh#L52.

AdamKorcz:
> This change won't work on a Mac, as you need to specify the extension after the -i. I did this same thing recently at https://github.com/istio/common-files/blob/master/bin/update-build-image.sh#L52.

Thanks for the heads up. This is probably not relevant, as I don't think Google ever runs the fuzzers on Mac infrastructure.
I have used sed dozens of times in production on OSS-fuzz and have never had any problems with it.

howardjohn (approved):

ericvn:
> > This change won't work on a Mac, as you need to specify the extension after the -i. I did this same thing recently at https://github.com/istio/common-files/blob/master/bin/update-build-image.sh#L52.
> 
> Thanks for the heads up. This is probably not relevant, as I don't think Google ever runs the fuzzers on Mac infrastructure.
> I have used sed dozens of times in production on OSS-fuzz and have never had any problems with it.

I guess the question is if we ever expect anyone else to want to run fuzzers, and will it be on Mac software. sed for the most part works on Macs, it just has some weirdness to it, -I is one of them.

AdamKorcz:
> I guess the question is if we ever expect anyone else to want to run fuzzers, and will it be on Mac software. sed for the most part works on Macs, it just has some weirdness to it, -I is one of them.

In that case, I would suggest offering a second, more general build-script rather than including compatibility for building the fuzzers in the OSS-fuzz build script. The OSS-fuzz build script is only meant to be run by OSS-fuzz and should not be seen as a general utility for local fuzzing.

If it is of interest to offer mac support to the community then I can look into it. As I see it, improving coverage has higher priority at this moment and I would rather use dev cycles for that. But I am open ears nonetheless.

ericvn:
> The case where I consider it a problem is if lack of mac support breaks anything in the CI test suite.

This is what I wanted to prevent. We shouldn't be merging code which we anticipate people may want to run locally as part their testing before merging a PR for example. Since you mention that this is not something to be run locally, I'm fine merging it in.

ericvn (approved):

PR #35107, opened 2021-09-09
Add test to validate --force flag ignores IST0139

deveshkandpal24121990:
- This PR adds a test that validates that IST0139 is ignored when `--force` flag is set to `true` and complains by default.
- Add logic for mocking k8s api server which is not dependent on `kubebuilder` and is more easily testable. 
- Adds `clientSet` wherever `client` and `kubeconfig` (`rest.Config`) are expected so that mock clientSet can be passed externally for testing purpose. 
- Motivation for adding this flag is based on this thread -> https://github.com/istio/istio/issues/33537#issuecomment-897746725

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


Monkeyanator (commented):
Having `--skipWebhookSelectorCheck` as a top-level install flag feels a bit awkward and overly specific IMO. Maybe we can generalize this into `--skipValidation` or similar in case we add other install-time validations?

ramaraochavali:
- --skipValidation - Are you saying we should fold `SkipServiceCheck` also to that?

How about this? --validationMode=Enforce/logOnly 
- Enforce means it validates and returns errors and stops the installation
- logOnly (or Report) it still does validation but just logs it as warning and proceeds with installation?

Monkeyanator:
I think having a flag to log validation failures as warnings instead of failing out makes the most sense to me.

deveshkandpal24121990:
@Monkeyanator  - it seems the `force` flag which is available as an argument serves the purpose of logging the error without completely failing. I was looking at adding a test for this scenario where I can set the `force` flag to true and have an extra MWC so that failure due to webhookAnalyzer can be [ignored](https://github.com/istio/istio/blob/3593fd3e0769be1ee18f6aad2cc6958e90b80da3/operator/pkg/helmreconciler/reconciler.go#L167) in this case.

I can see there are existing tests defined [here](https://github.com/istio/istio/blob/3593fd3e0769be1ee18f6aad2cc6958e90b80da3/operator/cmd/mesh/manifest-generate_test.go#L149) that I could add to for this particular scenario.

In order to deploy the duplicate MWC, I can see there is a provision available [here](https://github.com/istio/istio/blob/3593fd3e0769be1ee18f6aad2cc6958e90b80da3/operator/cmd/mesh/manifest_shared_test.go#L139), however, execution of that piece of code is dependent on setting up [kubebuilder](https://github.com/istio/istio/blob/3593fd3e0769be1ee18f6aad2cc6958e90b80da3/operator/cmd/mesh/manifest_shared_test.go#L82) it seems.

Do you have thoughts/suggestions on how can I add a test for this ?

deveshkandpal24121990:
@Monkeyanator - gentle reminder :) Whenever you have a moment, could you please take a look at my comment [above](https://github.com/istio/istio/pull/35107#issuecomment-931927707) ? Thanks!

Monkeyanator:
So are we thinking the `--force` flag is sufficient for this case and that we don't need a separate flag? SGTM.

In terms of adding additional resources for the test, not super familiar, but it seems like adding the webhook config to `manifest-generate/input-extra-resources` should do it?

deveshkandpal24121990:
@Monkeyanator - I tried adding the webhook config to `manifest-generate/input-extra-resources`, but the issue I think is that the test infrastructure seems to be dependent on having kubebuilder in place. The way these extra resources get applied is via -> https://github.com/istio/istio/blob/3593fd3e0769be1ee18f6aad2cc6958e90b80da3/operator/cmd/mesh/manifest_shared_test.go#L183

which depends on `testClient ` and `testRestConfig ` which actually get initialized inside this -> https://github.com/istio/istio/blob/3593fd3e0769be1ee18f6aad2cc6958e90b80da3/operator/cmd/mesh/manifest_shared_test.go#L92, which expects `kubebuilder` to be installed.

I could try installing kubebuilder locally, but i'm also wondering if this will work as is in the CI and kubebuilder will be present ?!

operator/cmd/mesh/manifest-generate_test.go
```diff
@@ -195,6 +197,74 @@ func TestManifestGenerateGateways(t *testing.T) {
...
+			assertFunc: func(g *WithT, objs *ObjectSet, err error) {
+				g.Expect(err).Should(BeNil())
+				g.Expect(objs.kind(name.MutatingWebhookConfigurationStr).size()).Should(Equal(2))
+			},
+		},
+		{
+			name:  "Duplicate MutatingWebhookConfiguration should not be allowed when --force is disabled",
+			force: false,
+			assertFunc: func(g *WithT, objs *ObjectSet, err error) {
+				g.Expect(err.Error()).Should(BeEquivalentTo("creating default tag would conflict:\nError [IST0139] (MutatingWebhookConfiguration" +
```
howardjohn, line 220: can we just do some strings.Containers match for 'Webhook overlaps with others'? We don't need to test the exact output here, may be hard to maintain

deveshkandpal24121990: makes sense, let me fix that.

ramaraochavali:
Thank you LGTM. @Monkeyanator Can you please take a look?

Monkeyanator (approved):
LGTM, thanks! Approved with a hold, feel free to ignore the comments and remove

ramaraochavali:
@istio/wg-user-experience-maintainers Can you PTAL?

PR #35109, opened 2021-09-09
Fuzzing: Improve inject fuzzer

AdamKorcz:

***The problem:***
The fuzzer currently does not make it into `IntoObject` here: 
https://github.com/istio/istio/blob/a89b03f81a4994bca43166675dba143b5bcc5217/pkg/kube/inject/inject.go#L504-L518

***The solution:***
Add a few more tokens in the dictionary to help the fuzzer mutate inputs to reach that code.


tests/fuzz/inject_fuzzer.go
```diff
@@ -20,6 +20,8 @@ import (
 
 	fuzz "github.com/AdaLogics/go-fuzz-headers"
 
+	_ "k8s.io/apimachinery/pkg/api/apitesting"
```
howardjohn, line 23: what is this for?

howardjohn (approved):

tests/fuzz/inject_fuzzer.go
AdamKorcz: shouldn't have been there. Sorry for that.

PR #35128, opened 2021-09-10
Remove useless code and define the errUnknownType.

devincd:
Signed-off-by: devincd <505259926@qq.com>

### DIFF
--- a/pilot/pkg/config/memory/store.go
+++ b/pilot/pkg/config/memory/store.go
@@ -17,7 +17,6 @@ package memory
 
 import (
 	"errors"
-	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pilot/pkg/config/memory/store.go
```diff
@@ -147,7 +147,7 @@ func (cr *store) Create(cfg config.Config) (string, error) {
 	kind := cfg.GroupVersionKind
 	s, ok := cr.schemas.FindByGroupVersionKind(kind)
 	if !ok {
-		return "", fmt.Errorf("unknown type %v", kind)
```
howardjohn, line 150: this loses context and makes it hard to understand why it failed

SpecialYang: Please don't remove this. It is very useful to detect unexpected kind resource when we try to store it.

devincd: ok

howardjohn (approved):

PR #35144, opened 2021-09-10
Lazy-ily compute sidecar scope for namespaces without Sidecar

howardjohn:
Fixes https://github.com/istio/istio/issues/32034

This is both an optimization and fix for ^.

Before, we iterated over all namespaces that had services and computed
sidecar scopes for them. This is inaccurate and inefficient: a namespace
may have 0 services (inaccurate), and it may have services but no
proxies (inefficient).

This changes it such that on each request, we lazy load the sidecar
scope for that namespace.

NOTE: this represents a fairly substantial change to push context in
that its not as much of a "snapshot" anymore. It still is deterministic
what the output is given a push version (ie computing this up front, or
lazy loaded will return the same result), but the actual PushContext
object is mutated. While I don't love this, I think its ok because its
an optimization matter (ie we could remove it entirely, it would just
make things super slow).

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -124,6 +124,26 @@ func newDestinationRuleIndex() destinationRuleIndex {
 	}
 }
 
+// sidecarIndex is the index of sidecar rules
+type sidecarIndex struct {
+	// sidecars for each namespace
+	sidecarsByNamespace map[string][]*SidecarScope
+	// the Sidecar for the root namespace (if present). This applies to any namespace without its own Sidecar.
+	rootConfig *config.Config
+	// defaultSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar.
+	// These are lazy-loaded. Access protected by defaultSidecarMu
+	defaultSidecarsByNamespace map[string]*SidecarScope
+	defaultSidecarMu           *sync.Mutex
+}
+
+func newSidecarIndex() sidecarIndex {
+	return sidecarIndex{
+		sidecarsByNamespace:        map[string][]*SidecarScope{},
+		defaultSidecarsByNamespace: map[string]*SidecarScope{},
+		defaultSidecarMu:           &sync.Mutex{},
+	}
+}
+
 // gatewayIndex is the index of gateways by various fields.
 type gatewayIndex struct {
 	// namespace contains gateways by namespace.
@@ -171,8 +191,8 @@ type PushContext struct {
 	// clusterLocalHosts extracted from the MeshConfig
 	clusterLocalHosts ClusterLocalHosts
 
-	// sidecars for each namespace
-	sidecarsByNamespace map[string][]*SidecarScope
+	// sidecarIndex stores sidecar resources
+	sidecarIndex sidecarIndex
 
 	// envoy filters for each namespace including global config namespace
 	envoyFiltersByNamespace map[string][]*EnvoyFilterWrapper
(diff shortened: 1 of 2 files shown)


pilot/pkg/model/push_context.go
```diff
@@ -130,17 +130,24 @@ type sidecarIndex struct {
...
 	// the Sidecar for the root namespace (if present). This applies to any namespace without its own Sidecar.
 	rootConfig *config.Config
-	// defaultSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar.
+	// computedSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar.
+	// These may be DefaultSidecarScopeForNamespace if rootConfig is empty or ConvertToSidecarScope if not.
 	// These are lazy-loaded. Access protected by defaultSidecarMu
-	defaultSidecarsByNamespace map[string]*SidecarScope
-	defaultSidecarMu           *sync.Mutex
+	computedSidecarsByNamespace map[string]*SidecarScope
+	// computedSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar,
```
ramaraochavali, line 137: nit computedSidecarsByNamespace -> gatewayDefaultSidecarsByNamespace

pilot/pkg/model/push_context.go
```diff
@@ -820,7 +847,36 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Colle
 		}
 	}
 
-	return DefaultSidecarScopeForNamespace(ps, proxy.ConfigNamespace)
+	// We didn't have a Sidecar in the namespace. This means we should use the default - either an implicit
+	// default selecting everything, or pulling from the namespace.
```
ramaraochavali, line 851: nit: pulling from the namespace. -> pulling from the root namespace?

pilot/pkg/model/push_context.go
```diff
@@ -820,7 +847,36 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Colle
...
+		// We have already computed the scope for this namespace, just fetch it
+		return sc
+	}
+	// We need to compute this namespace
+	var computed *SidecarScope
+	if ps.sidecarIndex.rootConfig != nil {
+		computed = ConvertToSidecarScope(ps, ps.sidecarIndex.rootConfig, proxy.ConfigNamespace)
+	} else {
+		computed = DefaultSidecarScopeForNamespace(ps, proxy.ConfigNamespace)
+		// Even though we were a gateway, we can store this as a gateway one since it could be used by a gateway
```
ramaraochavali, line 875: nit:  Even though we were a gateway ->  Even though this is for sidecar?

ramaraochavali (commented):
some minor comment nits. 

And one question - if we do it lazily - do we have to maintain this in push context? Should we just do it when proxy is initialized ?

howardjohn:
> some minor comment nits.
> 
> And one question - if we do it lazily - do we have to maintain this in push context? Should we just do it when proxy is initialized ?

It's basically an optimization. Instead of computing 1000 times for 1000 proxies in namespace we just do it once and share the result

ramaraochavali:
> It's basically an optimization. Instead of computing 1000 times for 1000 proxies in namespace we just do it once and share the result

nvm.. I understood this part was thinking about some thing else .

ramaraochavali (approved):
LGTM. Put a hold to address the minor comments. Feel free to remove it once they are done.

pilot/pkg/model/push_context.go
```diff
@@ -820,7 +847,36 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Colle
...
+	if f {
+		// We have already computed the scope for this namespace, just fetch it
+		return sc
+	}
+	// We need to compute this namespace
+	var computed *SidecarScope
+	if ps.sidecarIndex.rootConfig != nil {
+		computed = ConvertToSidecarScope(ps, ps.sidecarIndex.rootConfig, proxy.ConfigNamespace)
+	} else {
+		computed = DefaultSidecarScopeForNamespace(ps, proxy.ConfigNamespace)
```
hzxuzhonghu, line 874: isn't  this duplicate with L860

howardjohn: the code is duplicated but only this or L860 will run, never both

hzxuzhonghu: Correct, it confuses me with the below line.  We may not have gateways reside in same namespace as normal sidecar proxy

PR #35148, opened 2021-09-11
fix incorrect use of fmt placeholder

zc2638:

The placeholder `%v` parameter is a function, we should use `%p` to print the pointer.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn (changes requested):
I don't think we care about printing out the address of a pointer.. If we want to print something we should give it a useful String() representation or just not print it

zc2638:
> I don't think we care about printing out the address of a pointer.. If we want to print something we should give it a useful String() representation or just not print it

I can explain first.
`InterceptRuleMgrTypes["mock"]` is a func, there will be no useful String().And we can only get its memory address.

So if you feel that it is still inappropriate, please close this PR.

PR #35151, opened 2021-09-11
ServiceEntry use port map to make instances of pod

pangsq:
Signed-off-by: pangshaoqiang <pangsq9413@gmail.com>

For now, serviceEntry use port map to make instances of workloadEntries but does not use port map to make instances of pods.
This pr aims to let serviceEntry can use port names when selecting pod's ports as its instances.

### DIFF
--- a/pilot/pkg/serviceregistry/serviceentry/conversion.go
+++ b/pilot/pkg/serviceregistry/serviceentry/conversion.go
@@ -343,16 +343,21 @@ func getTLSModeFromWorkloadEntry(wle *networking.WorkloadEntry) string {
 
 // The workload instance has pointer to the service and its service port.
 // We need to create our own but we can retain the endpoint already created.
-func convertWorkloadInstanceToServiceInstance(workloadInstance *model.IstioEndpoint, serviceEntryServices []*model.Service,
+func convertWorkloadInstanceToServiceInstance(workloadInstance *model.WorkloadInstance, serviceEntryServices []*model.Service,
 	serviceEntry *networking.ServiceEntry) []*model.ServiceInstance {
 	out := make([]*model.ServiceInstance, 0)
+	if workloadInstance.Endpoint == nil {
+		return out
+	}
 	for _, service := range serviceEntryServices {
 		for _, serviceEntryPort := range serviceEntry.Ports {
-			ep := *workloadInstance
+			ep := *workloadInstance.Endpoint
 			ep.ServicePortName = serviceEntryPort.Name
-			// if target port is set, use the target port else fallback to the service port
-			// TODO: we need a way to get the container port map from k8s
-			if serviceEntryPort.TargetPort > 0 {
+			// if port is defined in portMap, use the port
+			portMap := workloadInstance.PortMap
+			if len(portMap) > 0 && portMap[ep.ServicePortName] > 0 {
+				ep.EndpointPort = portMap[ep.ServicePortName]
+			} else if serviceEntryPort.TargetPort > 0 { // if target port is set, use the target port else fallback to the service port
 				ep.EndpointPort = serviceEntryPort.TargetPort
 			} else {
 				ep.EndpointPort = serviceEntryPort.Number
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery_test.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery_test.go
@@ -876,7 +876,7 @@ func TestServiceDiscoveryWorkloadInstance(t *testing.T) {
 		expectEvents(t, events, Event{kind: "eds", host: "selector.com", namespace: selector.Namespace, endpoints: 0})
 
 		// Add a workload instance, and use port map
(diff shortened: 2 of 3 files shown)


pilot/pkg/serviceregistry/serviceentry/servicediscovery_test.go
```diff
@@ -866,6 +866,29 @@ func TestServiceDiscoveryWorkloadInstance(t *testing.T) {
...
+	t.Run("workload instance with port map", func(t *testing.T) {
+		// Delete the instance
+		callInstanceHandlers([]*model.WorkloadInstance{fi1}, sd, model.EventDelete, t)
+		instances := make([]*model.ServiceInstance, 0)
+		expectServiceInstances(t, sd, selector, 0, instances)
+		expectProxyInstances(t, sd, instances, "2.2.2.2")
+		expectEvents(t, events, Event{kind: "eds", host: "selector.com", namespace: selector.Namespace, endpoints: 0})
+
+		// Add a workload instance, and use port map
+		fi1.PortMap = map[string]uint32 {"tcp-444": 4444, "http-445": 4445}
```
stevenctl, line 879: If we modify `fi1` deep copy it since other tests could use it in the future

stevenctl (commented):
LGTM outside of 1 nit and formatting failures

pilot/pkg/serviceregistry/serviceentry/servicediscovery_test.go
pangsq: The PortMap will be reset to nil by the `defer func()` in the next line.

pilot/pkg/serviceregistry/serviceentry/conversion.go
```diff
@@ -343,16 +343,21 @@ func getTLSModeFromWorkloadEntry(wle *networking.WorkloadEntry) string {
...
+			ep := *workloadInstance.Endpoint
 			ep.ServicePortName = serviceEntryPort.Name
-			// if target port is set, use the target port else fallback to the service port
-			// TODO: we need a way to get the container port map from k8s
-			if serviceEntryPort.TargetPort > 0 {
+			// if port is defined in portMap, use the port
+			portMap := workloadInstance.PortMap
+			if len(portMap) > 0 && portMap[ep.ServicePortName] > 0 {
+				ep.EndpointPort = portMap[ep.ServicePortName]
+			} else if serviceEntryPort.TargetPort > 0 { // if target port is set, use the target port else fallback to the service port
```
hzxuzhonghu, line 360: I think TargetPort should be taken first, then serviceport name

hzxuzhonghu:
@stevenctl For serviceentry port we donot say the port name should be searched from pod's port.

pilot/pkg/serviceregistry/serviceentry/conversion.go
pangsq: In the [convertEndpoint](https://github.com/istio/istio/blob/1.12.0-alpha.1/pilot/pkg/serviceregistry/serviceentry/conversion.go#L238), it considers the port name at first. I think they should keep the same logic.

hzxuzhonghu: Good catch, i think that is not right either. Explicit target port should be first priority

stevenctl:
@hzxuzhonghu so the portMap should only apply to WorkloadEntry/endpoints embedded in the SE? That doesn't seem right to me.

hzxuzhonghu:
no, i mean the port name in SE does not equal the port name in workloadEntry. Unlike k8s service and pod, which has a targetPort field which could be string or int, if it is a string, then it means using it to find the target port number in pod.

stevenctl:
So if we have:

```yaml
kind: ServiceEntry
...
  ports:
  - number: 443
    name: https

---

kind: Pod
...
ports:
  - name: https
    containerPort: 8443


---

kind: WorkloadEntry
...
ports:
   https: 15443
```

You're saying we shouldn't do mappings based on the name at all? Only if the ServiceEntry explicitly has targetPort? 
It seems more expected that we map on the name. Am I misunderstanding your comment?

hzxuzhonghu:
We should map the target port in this case, what i meant is, if the se has explicitly set targetport, it takes preference.

pangsq:
@stevenctl K8s service's targetPort is different from ServiceEntry's. 
In k8s's service, targetPort can be a string. For example service may has a port with `targetPort: <port_name>`,  then k8s will find the port number whose name is `<port_name>` in pod. In that logic, port map can only be built between service's targetPort(as a string) and pod's port name. And if the targetPort is an int, k8s will directly use that targetPort to build endpoints.

However, ServiceEntry can only define `targetPort` in int type. If we choose to keep same with k8s, istio should use the targetPort to build instances if targetPort is defined without considering the port map.

@hzxuzhonghu Should we change the `port priority level` in [convertEndpoint](https://github.com/istio/istio/blob/1.12.0-alpha.1/pilot/pkg/serviceregistry/serviceentry/conversion.go#L238) ?

PR #35166, opened 2021-09-13
Fix pushqueue memory leak

howardjohn:
Currently, the queue has a memory leak. This leads to retaining
Connection objects after the proxy has disconnected. This object
includes references to PushContext, gRPC rw buffers (100kb), and can
store the DiscoveryResp/DiscoveryRequest as well. Overall, this adds up
a LOT - in many cases memory usage over 2gb wasted.

https://github.com/grpc/grpc-go/issues/4758 discusses this a bit more in
depth.

The root cause is the queue using `slice[1:]`. While this moves the
slice, it the underlying array may still store a reference to the
`Connection` at (formerly) `slice[0]`. This will get freed only if the
slice is completely copied to a new array (during an append, for
example), allowing the underlying array to be freed.

Before (memory not freed after client disconnect):

After (memory freed after client disconnect):

### DIFF
--- a/pilot/pkg/xds/pushqueue.go
+++ b/pilot/pkg/xds/pushqueue.go
@@ -88,7 +88,11 @@ func (p *PushQueue) Dequeue() (con *Connection, request *model.PushRequest, shut
 		return nil, nil, true
 	}
 
-	con, p.queue = p.queue[0], p.queue[1:]
+	con = p.queue[0]
+	// The underlying array will still exist, despite the slice changing, so the object may not GC without this
+	// See https://github.com/grpc/grpc-go/issues/4758
+	p.queue[0] = nil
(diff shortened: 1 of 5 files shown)


stevenctl (approved):

hzxuzhonghu:
good catch!

mandarjog:
Great catch. Was this reported by a customer? 
wondering why was this was not caught using the controlplane tests using the controlplane load tester.

howardjohn:
This was actually caught by our tests almost a year ago (https://github.com/grpc/grpc-go/issues/4013) but then was forgotten because we couldn't root cause it despite sinking a few hours into it. This wasnt directly reported by a user but I suspect it played a role in some OOMs in a real users cluster I have been working on - but I don't have hard proof of that.

PR #35187, opened 2021-09-14
Skip security tests that fail in multicluter and external-controlplane topologies

frankbu:

These three tests, which fail in a multicluster topology, are also failing with a remoteistiod topology. Same errors as described in the open issues referenced in the TODO comments.

### DIFF
--- a/tests/integration/security/ca_custom_root/secure_naming_test.go
+++ b/tests/integration/security/ca_custom_root/secure_naming_test.go
@@ -107,7 +107,7 @@ func TestSecureNaming(t *testing.T) {
 		Features("security.peer.secure-naming").
 		Run(func(t framework.TestContext) {
 			// TODO https://github.com/istio/istio/issues/32292
-			if t.Clusters().IsMulticluster() || t.Clusters().Default().Primary().IsExternalControlPlane() {
+			if t.AllClusters().IsMulticluster() {
 				t.Skip()
 			}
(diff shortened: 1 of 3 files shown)


tests/integration/security/ca_custom_root/secure_naming_test.go
```diff
@@ -107,7 +107,7 @@ func TestSecureNaming(t *testing.T) {
 		Features("security.peer.secure-naming").
 		Run(func(t framework.TestContext) {
 			// TODO https://github.com/istio/istio/issues/32292
-			if t.Clusters().IsMulticluster() {
+			if t.Clusters().IsMulticluster() || t.Clusters().Default().Primary().IsExternalControlPlane() {
```
howardjohn, line 110: Why does `IsMulticluster` not cover external control plane? external control plane is by definition multicluster, isn't it?

frankbu: No. `IsMulticluster` only refers to "mesh clusters" (it doesn't consider clusters outside the mesh). I'm currently trying to run these tests in a single cluster (config cluster) w external control plane topology and running into the same issue(s).

frankbu: That gives me an idea though. `t.AllClusters().IsMulticluster()` is a simpler check.

howardjohn (approved):

PR #35188, opened 2021-09-14
[release-1.11] Env Var allowing proxies to utilize OS CA Cert (#33472)

Kmoneal:
* Env Var allowing proxies to utilize OS CA Cert

* Introduce VERIFY_CERTIFICATE_AT_CLIENT as a bool env var
* Select an existing CA cert bundle from a set of known file locations
* If bool set and DR does not set CaCertificates, replace the proxies'
CA cert with an OS CA cert found in the file system

* Move OS cert replace logic to proxy

* Fix typo

* Move function in pilot code and check for OS cert once

* Move SdsCertificateConfigFromResourceName out of pilot authenticate
code and into utils for agent.
* Determine OS CA certificate in pilot-agent to pass down instead of
searching for the file more often

* Linter and change setting off by default

* Move OS CA Path find to option generation

* Add release notes

* Lock pointers when accessing or using them

* Remove mutex inside Options

* Fix secret resource choice

* Add unit tests

* Fix tests and linter errors

* Improve tests, releasenotes and logging

* Add back code that was needed

* Fix linter issues

* Use string constant in place of file-root:system

* Update to remove proxy from ClusterBuilder

* Fix auto import location

* Refactor code

* Refactor SdsCertificateConfig to common package

* Move security tests from util.go

* Fix releasenotes

* Add documentation to public methods

* Improve VERIFY_CERT_AT_CLIENT testing

* initialize CAFilePath in security instead of main

* Refactor and remove dead code

* Remove code that is no longer needed with CA file path being generated
on init
* Move CACertFilePath init outside of shared package so it only needs to
be made for sdsservice.go and secretcache.go

* Refactor to move cafile package

* Refactor toEnvoySecret funciton

* Revert changes that were needed for istiod to know the OS CA file path

* Fix linter problems

* Correct tests and introduce caRootPath to SecretManager

* SecretManager uses GenerateSecret and to avoid adding another
perameter to the function, SecretManager implementations need to have it
in the type.
* `file-root:` was missing from secretcache_test.go to pull the correct
certificate information

* Fix missing argument for mock initializaiton

* Reverse mock changes and replace string with const

* Fix SdsCertificateConfig to use file-root:system cert.

* Fix tests by removing file-root: prefix

* Fix empty OS CA cert causing an error

* Remove testing print code

* Rebase cluster_builder_test update struct

* Fix rebase changes made to DestiantionRule


jwendell:
Shouldn't this target master?

Kmoneal:
> Shouldn't this target master?

This is already in master and this is the cherry-pick. Sorry for the delay, I was just trying to test and make sure everything is still behaving as expected.

brian-avery:
The cherrypick confused me at first as well. This was originally merged into master in https://github.com/istio/istio/pull/33472. This adds it to 1.11. 

The goal in the original design doc for this was to ship it in 1.9 and 1.10. In that time, 1.11 was published and it makes sense to me to merge it here as well. 

This behavior is also disabled by default and must be enabled using the `VERIFY_CERTIFICATE_AT_CLIENT` flag.

jwendell:
This looks like a big change to include in a patch release.

jacob-delgado:
The product security working group would like for this to be included, however, we understand that it's up to the Release Managers if such a patch gets included. I can work with Kenan to make sure that this is working as intended for 1.11 (we did this for master) if that would give you confidence in accepting such a patch. If not, that's fine and we will proceed with changes to master.

jacob-delgado:
We decided not to cherry pick this back into 1.9-1.11

PR #35193, opened 2021-09-14
perf: reuse PushRequest object during Merge

howardjohn:
On a cluster with ~10k configs (svcs, eps, etc):
Before: `All caches have been synced up in 7.978471048s, marking server ready`
After: `All caches have been synced up in 2.101581551s, marking server ready`

Currently all usages already are safe, so mutation should not cause any issue.

This is so much faster mostly because ConfigUpdates; before we are allocating a ton of maps

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -353,6 +353,9 @@ const (
 )
 
 // Merge two update requests together
+// Merge behaves similarly to a list append; usage should in the form `a = a.merge(b)`.
+// Importantly, Merge may decide to allocate a new PushRequest object or reuse the existing one - both
+// inputs should not be used after completion.
 func (pr *PushRequest) Merge(other *PushRequest) *PushRequest {
 	if pr == nil {
 		return other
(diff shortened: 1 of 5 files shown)


hzxuzhonghu (commented):
lgtm, only use merge before processing each connection.

hzxuzhonghu:
It is amazing to see so much improvement by slice and map append. Maybe better to provide a benchmark.

howardjohn:
There is one in https://github.com/istio/istio/pull/35189 I added. But will need to run it against this pr

howardjohn:
@hzxuzhonghu the benchmark doesn't show as much improvement since we only have `pushesMerged := 10`. With `pushesMerged := 10000`:

```
name           old time/op    new time/op    delta
PushRequest-6     639ms ± 4%     413ms ±27%  -35.37%  (p=0.002 n=6+6)

name           old alloc/op   new alloc/op   delta
PushRequest-6     950MB ± 0%     107MB ± 0%  -88.68%  (p=0.004 n=6+5)

name           old allocs/op  new allocs/op  delta
PushRequest-6     3.08M ± 0%     3.04M ± 0%   -1.32%  (p=0.002 n=6+6)
``

hzxuzhonghu (approved):

PR #35208, opened 2021-09-15
Add portforward to operator ClusterRole

litong01:

When revision is used, while delete istio resources, the operator will use GetProxyInfo to retrieve resource information, it will need the pods/portforward resource to be in the clusterrole. Without it, the method fails, and the reconciler goes into infinite loop. This PR fixes issue/#35016


morvencao (approved):
/lgtm

howardjohn (approved):

ostromart (approved):
LGTM but please change the title to reflect the change e.g. "Add portforward to operator ClusterRole"

PR #35211, opened 2021-09-15
Improve the error handling of STS token manager

lei-tang:
- When STS CreateTokenManager() fails, return the error.
- Add error detection to constructAudience().
- When NewSecureTokenServiceExchanger() fails, return the error.

- [ ] Docs
- [ ] Installation
- [X ] Networking
- [ ] Policies and Telemetry
- [X ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/cmd/pilot-agent/options/security.go
+++ b/pilot/cmd/pilot-agent/options/security.go
@@ -67,9 +67,6 @@ func NewSecurityOptions(proxyConfig *meshconfig.ProxyConfig, stsPort int, tokenM
 		// tokenManager is gcp token manager when using the default token manager plugin.
 		tokenManager, err = tokenmanager.CreateTokenManager(tokenManagerPlugin,
 			tokenmanager.Config{CredFetcher: o.CredFetcher, TrustDomain: o.TrustDomain})
-		if err != nil {
-			log.Errorf("failed to create token manager: %v", err)
-		}
 	}
 	o.TokenManager = tokenManager
 
--- a/pilot/cmd/pilot-agent/options/security.go
+++ b/pilot/cmd/pilot-agent/options/security.go
@@ -116,7 +116,11 @@ func SetupSecurityOptions(proxyConfig *meshconfig.ProxyConfig, secOpt *security.
 	}
 	// TODO extract this logic out to a plugin
 	if o.CAProviderName == security.GoogleCAProvider || o.CAProviderName == security.GoogleCASProvider {
-		o.TokenExchanger = stsclient.NewSecureTokenServiceExchanger(o.CredFetcher, o.TrustDomain)
+		var err error
+		o.TokenExchanger, err = stsclient.NewSecureTokenServiceExchanger(o.CredFetcher, o.TrustDomain)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	if o.ProvCert != "" && o.FileMountedCerts {
--- a/pilot/cmd/pilot-agent/options/security.go
+++ b/pilot/cmd/pilot-agent/options/security.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"path/filepath"
 	"strings"
-	"time"
 
 	meshconfig "istio.io/api/mesh/v1alpha1"
 	"istio.io/istio/pilot/pkg/features"
@@ -31,11 +30,6 @@ import (
 	"istio.io/istio/security/pkg/nodeagent/plugin/providers/google/stsclient"
 	"istio.io/istio/security/pkg/stsservice/tokenmanager"
 	"istio.io/pkg/log"
-)
-
-const (
-	tokenManagerMaxNumRetries = 3
-	tokenManagerRetryInterval = 1 * time.Second
 )
 
 func NewSecurityOptions(proxyConfig *meshconfig.ProxyConfig, stsPort int, tokenManagerPlugin string) (*security.Options, error) {
@@ -71,19 +65,10 @@ func NewSecurityOptions(proxyConfig *meshconfig.ProxyConfig, stsPort int, tokenM
 	var tokenManager security.TokenManager
 	if stsPort > 0 || xdsAuthProvider.Get() != "" {
 		// tokenManager is gcp token manager when using the default token manager plugin.
-		for retryCount := 1; retryCount <= tokenManagerMaxNumRetries; retryCount++ {
-			tokenManager, err = tokenmanager.CreateTokenManager(tokenManagerPlugin,
-				tokenmanager.Config{CredFetcher: o.CredFetcher, TrustDomain: o.TrustDomain})
-			if err != nil {
-				if retryCount < tokenManagerMaxNumRetries {
-					log.Errorf("failed to create token manager: %v, retry %v ...", err, retryCount)
-					time.Sleep(tokenManagerRetryInterval)
-				} else { // no more retries
-					log.Errorf("failed to create token manager: %v, after trying %v times", err, retryCount)
-				}
-			} else {
-				break
-			}
+		tokenManager, err = tokenmanager.CreateTokenManager(tokenManagerPlugin,
+			tokenmanager.Config{CredFetcher: o.CredFetcher, TrustDomain: o.TrustDomain})
+		if err != nil {
+			log.Errorf("failed to create token manager: %v", err)
 		}
 	}
 	o.TokenManager = tokenManager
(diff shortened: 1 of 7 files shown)


security/pkg/stsservice/tokenmanager/tokenmanager.go
```diff
@@ -80,22 +80,24 @@ func GetGCPProjectInfo() GCPProjectInfo {
...
 	case GoogleTokenExchange:
 		if projectInfo := GetGCPProjectInfo(); len(projectInfo.Number) > 0 {
 			if p, err := google.CreateTokenManagerPlugin(config.CredFetcher, config.TrustDomain,
 				projectInfo.Number, projectInfo.clusterURL, true); err == nil {
 				tm.plugin = p
 			}
+			// When err != nil, the error will be returned at the end
 		} else {
-			log.Warnf("%v token manager specified but failed to ready GCP project information", GoogleTokenExchange)
+			err = fmt.Errorf("%v token manager specified but failed to ready GCP project information", GoogleTokenExchange)
```
howardjohn, line 97:
```suggestion
			return nil, fmt.Errorf("%v token manager specified but failed to ready GCP project information", GoogleTokenExchange)
```

pilot/cmd/pilot-agent/options/security.go
```diff
@@ -65,8 +71,20 @@ func NewSecurityOptions(proxyConfig *meshconfig.ProxyConfig, stsPort int, tokenM
...
 	if stsPort > 0 || xdsAuthProvider.Get() != "" {
 		// tokenManager is gcp token manager when using the default token manager plugin.
-		tokenManager = tokenmanager.CreateTokenManager(tokenManagerPlugin,
-			tokenmanager.Config{CredFetcher: o.CredFetcher, TrustDomain: o.TrustDomain})
+		for retryCount := 1; retryCount <= tokenManagerMaxNumRetries; retryCount++ {
+			tokenManager, err = tokenmanager.CreateTokenManager(tokenManagerPlugin,
+				tokenmanager.Config{CredFetcher: o.CredFetcher, TrustDomain: o.TrustDomain})
+			if err != nil {
+				if retryCount < tokenManagerMaxNumRetries {
+					log.Errorf("failed to create token manager: %v, retry %v ...", err, retryCount)
```
howardjohn, line 79: I don't think a retry helps. The GCP metadata fetch is cached, if it fails once we would reuse the empty result.

Also I think there are two different STS token managers - probably both has the same issue.

howardjohn (changes requested):
Thanks for the fix! IMO this is complex enough that we likely need a test to cover this and ensure that all functionality works as expected in the presence of metadata server issues.

pilot/cmd/pilot-agent/options/security.go
lei-tang: - Based on the comment, the retry mechanism is removed.
- The error message “token manager failed to generate access token: no plugin is found” indicates that it results from CreateTokenManager(). If other token managers encounter the same error messages, we can address them in a follow-up PR.

security/pkg/stsservice/tokenmanager/tokenmanager.go
lei-tang: Revised based on the comment.

pilot/cmd/pilot-agent/options/security.go
howardjohn: That does not make sense to me. We can be 100% certain if the bug exists there it also exists here: https://github.com/istio/istio/blob/562cdb883aa613efdf6c2584709f6a1f7adf99a0/pilot/cmd/pilot-agent/options/security.go#L116 because its calling the exact same code.

If we do tests with GCE metadata server degraded we will quickly find ^ is broken as well. 

The error logs from the bug reports show the code above also has issues.

lei-tang: The code has been updated to:
- Add error detection to constructAudience().
- When NewSecureTokenServiceExchanger() fails, return the error.

lei-tang:
CreateTokenManager() and NewSecureTokenServiceExchanger() are tested in sts_test.go and stsclient_test.go. To facilitate the testing of the scenarios that metadata server fails, some functions may need to be refactored such that the metadata server calls are abstracted as an input parameter to allow mocked calls, instead of hardcoded calling the metadata servers in the function (e.g., constructAudience() calls platform.NewGCP().Metadata()[platform.GCPClusterURL]).  I created a separate issue https://github.com/istio/istio/issues/35214 to improve the testing of STS token operations when metadata calls fail.

howardjohn:
Can we at least do a manual test? you can add firewall rule for the GCE
metadata ip or something maybe?

lei-tang:
Currently I am booked with other tasks and I will try a manual test when I get bandwidth. Meanwhile, it will be great if someone can help on testing (manual tests and https://github.com/istio/istio/issues/35214). To split the work into smaller increments, can we merge this PR and address the remaining issues (e.g., mock testing etc) in follow-up PRs?

pilot/cmd/pilot-agent/options/security.go
```diff
@@ -65,8 +65,11 @@ func NewSecurityOptions(proxyConfig *meshconfig.ProxyConfig, stsPort int, tokenM
 	var tokenManager security.TokenManager
 	if stsPort > 0 || xdsAuthProvider.Get() != "" {
 		// tokenManager is gcp token manager when using the default token manager plugin.
-		tokenManager = tokenmanager.CreateTokenManager(tokenManagerPlugin,
+		tokenManager, err = tokenmanager.CreateTokenManager(tokenManagerPlugin,
 			tokenmanager.Config{CredFetcher: o.CredFetcher, TrustDomain: o.TrustDomain})
+		if err != nil {
+			log.Errorf("failed to create token manager: %v", err)
```
howardjohn, line 71: Why do we log here but return err below? Seems inconsistent

security/pkg/stsservice/tokenmanager/tokenmanager.go
```diff
@@ -80,22 +80,24 @@ func GetGCPProjectInfo() GCPProjectInfo {
 
 // CreateTokenManager creates a token manager with specified type and returns
 // that token manager
-func CreateTokenManager(tokenManagerType string, config Config) security.TokenManager {
+func CreateTokenManager(tokenManagerType string, config Config) (security.TokenManager, error) {
 	tm := &TokenManager{
 		plugin: nil,
 	}
+	var err error
```
howardjohn, line 87:
```suggestion
```

security/pkg/stsservice/tokenmanager/tokenmanager.go
```diff
@@ -80,22 +80,24 @@ func GetGCPProjectInfo() GCPProjectInfo {
...
 				tm.plugin = p
 			}
+			// When err != nil, the error will be returned at the end
 		} else {
-			log.Warnf("%v token manager specified but failed to ready GCP project information", GoogleTokenExchange)
+			return nil, fmt.Errorf("%v token manager specified but failed to ready GCP project information", GoogleTokenExchange)
 		}
 	}
-	return tm
+	return tm, err
```
howardjohn, line 100:
```suggestion
	return tm, nil
```

pilot/cmd/pilot-agent/options/security.go
lei-tang: The log here has been removed.

security/pkg/stsservice/tokenmanager/tokenmanager.go
lei-tang: This error variable is needed because "if p, err := google.CreateTokenManagerPlugin(...)" may return an error.

security/pkg/stsservice/tokenmanager/tokenmanager.go
lei-tang: Because "if p, err := google.CreateTokenManagerPlugin(...)" may return an error, here returns "tm, err", instead of "tm, nil".

security/pkg/stsservice/tokenmanager/tokenmanager.go
lei-tang: I have refactored the code to remove this variable.

security/pkg/stsservice/tokenmanager/tokenmanager.go
lei-tang: I have refactored code to return "tm, nil" here.

howardjohn (approved):

PR #35212, opened 2021-09-15
Add metadata exchange as native code

howardjohn:
Benchmark:
```
name                            old time/op        new time/op        delta
ClusterGeneration/telemetry-6          168µs ± 1%         155µs ± 2%  -7.94%  (p=0.029 n=4+4)
ListenerGeneration/telemetry-6         512µs ± 3%         501µs ± 1%  -1.98%  (p=0.114 n=4+4)

name                            old kb/msg         new kb/msg         delta
ClusterGeneration/telemetry-6           5.53 ± 0%          5.34 ± 0%  -3.42%  (p=0.029 n=4+4)
ListenerGeneration/telemetry-6          32.9 ± 0%          31.9 ± 0%  -3.16%  (p=0.029 n=4+4)

name                            old resources/msg  new resources/msg  delta
ClusterGeneration/telemetry-6           14.0 ± 0%          14.0 ± 0%    ~     (all equal)
ListenerGeneration/telemetry-6          6.00 ± 0%          6.00 ± 0%    ~     (all equal)

name                            old alloc/op       new alloc/op       delta
ClusterGeneration/telemetry-6         76.1kB ± 0%        70.8kB ± 0%  -6.96%  (p=0.029 n=4+4)
ListenerGeneration/telemetry-6         254kB ± 0%         251kB ± 0%  -1.20%  (p=0.029 n=4+4)

name                            old allocs/op      new allocs/op      delta
ClusterGeneration/telemetry-6          1.19k ± 0%         1.10k ± 0%  -7.24%  (p=0.029 n=4+4)
ListenerGeneration/telemetry-6         3.15k ± 0%         3.12k ± 0%  -1.14%  (p=0.029 n=4+4)
```

Broken out of https://github.com/istio/istio/pull/33583

This moves the MX code from EnvoyFilter to native code. This is an important part of the Telemetry API implementation.

In a followup, the remaining filters will be converted as well.

The MX filter is now unconditionally enabled, similar to other Istio features. This is because we need the MX filter on both the client *and* server in order for telemetry to work. Because we cannot know the entirety of the Telemetry config for client/servers (especially cross cluster), there is no reliable way to detect if its safe to elide MX filter. For users that have extreme high performance use cases and do not use Istio telemetry, nor any other filters depending on MX, there is an opt out flag left in Istiod. We expect this to be the minority use case (@ramaraochavali :slightly_smiling_face:).


pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -275,6 +281,16 @@ func (cb *ClusterBuilder) applyDestinationRule(mc *MutableCluster, clusterMode C
 	return subsetClusters
 }
 
+func (cb *ClusterBuilder) applyMetadataExchange(c *cluster.Cluster) {
+	// We configure MX only if we have some Telemetry configured. This essentially acts as a feature flag
```
douglas-reid, line 285: does this comment still match the intent/code?  It seems like it is now just gated on the actual feature flag.

pilot/pkg/networking/core/v1alpha3/networkfilter.go
```diff
@@ -39,6 +41,18 @@ import (
 // redisOpTimeout is the default operation timeout for the Redis proxy filter.
 var redisOpTimeout = 5 * time.Second
 
+func buildTelemetryNetworkFilters(class istionetworking.ListenerClass) []*listener.Filter {
+	filterstack := make([]*listener.Filter, 0)
+
+	// We add metadata exchange on inbound if any telemetry exists, as a client may expect metadata exchange.
+	// For outbound, we only need to set it if our own proxy has telemetry enabled.
```
douglas-reid, line 48: Should this still be conditional? naively, it seems like we to allow for metadata exchange for outbound calls even if we aren't collecting telemetry locally. otherwise, won't server-side telemetry potentially be incomplete?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
howardjohn: Oops, will need to go through the code. I pulled directly from the original PR and hacked on it until it compiled - bad comments compile :slightly_smiling_face:

pilot/pkg/networking/core/v1alpha3/networkfilter.go
howardjohn: I think the comment is wrong but the code is right. Envoyfilter only adds to inbound: https://github.com/istio/istio/pull/35212/files/19b4c7cc957b0627d0fc3c45b8f83626202dddec#diff-6f01bd95b2effca99e0e138bfe1d57c761ec27e506a6f88ec69d4a64467c48d8L139

douglas-reid: Ah, lost context. cluster outbound, filter inbound, correct?

howardjohn: Yeah. Let me clean up the comments and code a bit - I was confused myself

ramaraochavali:
Should we also fold PILOT_ENABLE_TELEMETRY_LABEL and PILOT_ENDPOINT_TELEMETRY_LABEL in this config? Or MetadataExchange can be independently enabled?

howardjohn:
> Should we also fold PILOT_ENABLE_TELEMETRY_LABEL and PILOT_ENDPOINT_TELEMETRY_LABEL in this config? Or MetadataExchange can be independently enabled?

Like have 1 env var to control all of them or something? I would be ok with that I think. You can technically use MX without Istio telemetry but that is so obscure I don't think we care too much (?)

bianpengyuan:
> Should we also fold PILOT_ENABLE_TELEMETRY_LABEL and PILOT_ENDPOINT_TELEMETRY_LABEL in this config? Or MetadataExchange can be independently enabled?

I think there can be case where user uses MX with envoy access log service. Better have separated control if it is not too much of burden. Setting TELEMETRY_LABEL to false removes MX feels a bit too magic to me.

ramaraochavali:
Ok. I am fine with leaving them as is. I was thinking we use the new flag
that was added here to control telemetry labels also I.e if metadata
exchange is enabled we add labels.

douglas-reid (approved):
Cool. Telemetry changes look good to me.

pilot/pkg/model/context.go
```diff
@@ -1066,3 +1067,11 @@ type GatewayController interface {
 	// this would return true only if there was a policy allowing `ingress-ns` to access Secrets in the `ns-name` namespace.
 	SecretAllowed(resourceName string, namespace string) bool
 }
+
+// OutboundClassFromType is a help to turn a NodeType for outbound config into a Gateway or SidecarOutbound
```
ramaraochavali, line 1071: nit: is a helper

ramaraochavali (approved):

morvencao (approved):

PR #35219, opened 2021-09-16
Revert "Bump addon versions"

ericvn:
Reverts istio/istio#35196

The original PR causes the https://preliminary.istio.io/latest/docs/tasks/observability/metrics/using-istio-dashboard/ istio.io test to fail, preventing merges in the istio.io master branch.  Part of that test is to do a curl to the dashboard which instead of getting a `200` now gets a `404`:

++access_grafana_istio_mesh_dashboard(): curl -L -s -o /dev/null -w '%{http_code}' http://localhost:3000/dashboard/db/istio-mesh-dashboard
+__verify_with_retry(): out=404

I'm fine not reverting this and updating the test a replacement for the `http://localhost:3000/dashboard/db/istio-mesh-dashboard` can be found for the test case.

A smaller revert of just the `samples/addons/grafana.yaml` might be acceptable, but not sure if that's a good idea or not.

### DIFF
--- a/manifests/addons/gen.sh
+++ b/manifests/addons/gen.sh
@@ -30,7 +30,7 @@ TMP=$(mktemp -d)
 {
 helm3 template kiali-server \
   --namespace istio-system \
-  --version 1.40.0 \
+  --version 1.38.0 \
   --include-crds \
   --set nameOverride=kiali \
   --set fullnameOverride=kiali \
@@ -42,7 +42,7 @@ helm3 template kiali-server \
 # Set up prometheus
 helm3 template prometheus prometheus \
   --namespace istio-system \
-  --version 14.6.1 \
+  --version 14.3.0 \
   --repo https://prometheus-community.github.io/helm-charts \
   -f "${WD}/values-prometheus.yaml" \
   > "${ADDONS}/prometheus.yaml"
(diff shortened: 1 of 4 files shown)


ericvn:
I hand-crafted the test to use `samples/addons/grafana.yaml` from the prior commit, `5811650b5f79632071afeada9b1f47496dc1eea9` and the tests passed. Moving to `907440903fcb191d32ded47bb4dba6a5367ea60e` causing the test to timeout with the 404's.

howardjohn:
I'll poke around at it for a few minutes and approve if I cannot figure it
out. Thanks !

howardjohn:
`http://localhost:3000/dashboard/db/istio-mesh-dashboard` works for me...

ericvn:
Slacked with John since it was failing for me. Determined, the Grafana update changed the URL for the dashboard. It is now `http://localhost:3000/d/G8wLrJIZk/istio-mesh-dashboard`.

Testing this change in istio.io now.

ericvn:
https://github.com/istio/istio.io/pull/10349 was created to point to the new URLs.

https://github.com/istio/istio/issues/35226 was created to see about the possibility of moving the Grafana URLs back to the old version.

PR #35233, opened 2021-09-16
Add flag to skip TProxy related parts in test.

bianpengyuan:
In some env, like GKE autopilot, TProxy mode cannot work (Echo cannot become ready) because NET_ADMIN is disallowed. This follows VM and Delta xDS to add a flag to skip TProxy echo set up and test.

### DIFF
--- a/tests/integration/pilot/common/apps.go
+++ b/tests/integration/pilot/common/apps.go
@@ -208,12 +208,10 @@ func SetupApps(t resource.Context, i istio.Instance, apps *EchoDeployments) erro
 			WorkloadOnlyPorts: common.WorkloadPorts,
 		}).
 		WithConfig(echo.Config{
-			Service:   PodTproxySvc,
-			Namespace: apps.Namespace,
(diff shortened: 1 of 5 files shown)


howardjohn (approved):

tests/integration/pilot/common/apps.go
```diff
@@ -208,12 +208,10 @@ func SetupApps(t resource.Context, i istio.Instance, apps *EchoDeployments) erro
 			WorkloadOnlyPorts: common.WorkloadPorts,
 		}).
 		WithConfig(echo.Config{
-			Service:   PodTproxySvc,
-			Namespace: apps.Namespace,
-			Ports:     common.EchoPorts,
-			Subsets: []echo.SubsetConfig{{
```
stevenctl, line 214: The annotation shouldn't be removed, right?

bianpengyuan: Oops Left off from test.. thanks for catching this!

PR #35235, opened 2021-09-16
flag to skip arbitrary types of workloads

stevenctl:
Alternative/addition to #35233

### DIFF
--- a/pilot/pkg/util/sets/string.go
+++ b/pilot/pkg/util/sets/string.go
@@ -14,14 +14,7 @@
 
 package sets
 
-import (
-	"flag"
-	"fmt"
-	"sort"
-	"strings"
-)
-
-var _ flag.Value = Set{}
+import "sort"
 
 type Set map[string]struct{}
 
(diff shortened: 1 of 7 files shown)


howardjohn (approved):

pilot/pkg/util/sets/string.go
```diff
@@ -139,3 +146,15 @@ func (s Set) Equals(other Set) bool {
 func (s Set) Empty() bool {
 	return len(s) == 0
 }
+
+func (s Set) Set(value string) error {
+	parts := strings.Split(value, ",")
+	for _, part := range parts {
```
hzxuzhonghu, line 152: What's this, i haven't looked at kind of set like this

hzxuzhonghu: It will lead to misuse.

howardjohn: Oh yeah I didn't notice this was in the core code. This should probably just be done in the test code to make sure no one uses this in core code?

hzxuzhonghu: Agree

stevenctl: Like `type setFlag sets.Set` or something +1

PR #35237, opened 2021-09-17
config capture-all-dns via env ISTIO_META_DNS_CAPTURE_ALL

liuxu623:

support config capture-all-dns via env ISTIO_META_DNS_CAPTURE_ALL

### DIFF
--- a/tools/istio-iptables/pkg/cmd/root.go
+++ b/tools/istio-iptables/pkg/cmd/root.go
@@ -39,6 +39,8 @@ var (
 	// Enable interception of DNS.
 	dnsCaptureByAgent = env.RegisterBoolVar("ISTIO_META_DNS_CAPTURE", false,
 		"If set to true, enable the capture of outgoing DNS packets on port 53, redirecting to istio-agent on :15053").Get()
+	dnsCaptureAll = env.RegisterBoolVar("ISTIO_META_DNS_CAPTURE_ALL",false,
+		"If set to true, capture all DNS traffic at port 53 instead of only capturing DNS traffic to DNS server IP.").Get()
 )
 
 var rootCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
can you give more info why? This will not work in many environments

liuxu623:
> can you give more info why? This will not work in many environments

@howardjohn We'd like to make CNI and init-container use same iptabels rules, but CNI capture-all-dns is true, we couldn't set capture-all-dns when use init-container.

tools/istio-iptables/pkg/cmd/root.go
```diff
@@ -39,6 +39,8 @@ var (
 	// Enable interception of DNS.
 	dnsCaptureByAgent = env.RegisterBoolVar("ISTIO_META_DNS_CAPTURE", false,
 		"If set to true, enable the capture of outgoing DNS packets on port 53, redirecting to istio-agent on :15053").Get()
+	dnsCaptureAll = env.RegisterBoolVar("ISTIO_META_DNS_CAPTURE_ALL",false,
```
hzxuzhonghu, line 42: I anm not sure what's the diff with `ISTIO_META_DNS_CAPTURE`

liuxu623:
And sometimes app may use other DNS Server not in /etc/resolv.conf，this PR can config capture dns traffic to other DNS Server.

tools/istio-iptables/pkg/cmd/root.go
liuxu623: When use init-container to config iptables rules，
set ISTIO_META_DNS_CAPTURE true will only capturing DNS traffic to DNS server IP in /etc/resolv.conf.
set ISTIO_META_DNS_CAPTURE and ISTIO_META_DNS_CAPTURE_ALL treu will capturing all DNS traffic, even it not in /etc/resolv.conf.

liuxu623: When use CNI config iptables rules, because CNI can not get DNS server IP in /etc/resolv.conf, set ISTIO_META_DNS_CAPTURE true will capturing all DNS traffic.

hzxuzhonghu: That's really confusing literally

liuxu623: How about `ISTIO_META_CAPTURE_ALL_DNS_TRAFFIC` ?

PR #35240, opened 2021-09-17
fix 'misspellings

dsy3502:
fix misspellings


howardjohn (changes requested):
Thanks for the PR! this code is generated from istio/api so it would need to be sent there.

PR #35243, opened 2021-09-17
add exponential backoff for webhook patch retry

ramaraochavali:
Recently we had an issue where the root certificate on istiod nodes was inconsistent and as a result of that both istiod instances were competing with each other to update the webhook certificate which resulted in retry loop 
with the following error - spamming k8s api server and almost bringing down etcd.

"Work item handle failed (Operation cannot be fulfilled on mutatingwebhookconfigurations.admissionregistration.k8s.io "istio-sidecar-injector-mesh-control-plane": the object has been modified; please apply your changes to the latest version and try again), retry after delay 2s

This PR adds exponential backoff with random jitter for patching retries so that all istiod instances do not compete with a fixed delay 2s.
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/webhooks/webhookpatch.go
+++ b/pkg/webhooks/webhookpatch.go
@@ -80,7 +80,7 @@ func NewWebhookCertPatcher(
 	retryBackoff.InitialInterval = 2 * time.Second
 	retryBackoff.MaxInterval = 15 * time.Minute // Max Backoff interval.
 	// If certificate has not been patched with in an hour of retries, let us stop retrying.
-	retryBackoff.MaxElapsedTime = 1 * time.Hour
+	retryBackoff.MaxElapsedTime = 0
 	return &WebhookCertPatcher{
 		client:          client,
 		revision:        revision,
--- a/pkg/webhooks/webhookpatch.go
+++ b/pkg/webhooks/webhookpatch.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff"
 	v1 "k8s.io/api/admissionregistration/v1"
 	kubeErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -75,12 +76,17 @@ func NewWebhookCertPatcher(
 		func(options *metav1.ListOptions) {
 			options.LabelSelector = fmt.Sprintf("%s=%s", label.IoIstioRev.Name, revision)
 		})
+	retryBackoff := backoff.NewExponentialBackOff()
+	retryBackoff.InitialInterval = 2 * time.Second
+	retryBackoff.MaxInterval = 15 * time.Minute // Max Backoff interval.
+	// If certificate has not been patched with in an hour of retries, let us stop retrying.
+	retryBackoff.MaxElapsedTime = 1 * time.Hour
 	return &WebhookCertPatcher{
 		client:          client,
 		revision:        revision,
 		webhookName:     webhookName,
 		CABundleWatcher: caBundleWatcher,
-		queue:           queue.NewQueue(time.Second * 2),
+		queue:           queue.NewBackOffQueue(retryBackoff),
 		whcLw:           whcLw,
 	}, nil
 }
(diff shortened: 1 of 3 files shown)


howardjohn:
I agree its good to have resiliancy but shouldn't leader election ensure only one patches?

pkg/webhooks/webhookpatch.go
```diff
@@ -75,12 +76,17 @@ func NewWebhookCertPatcher(
 		func(options *metav1.ListOptions) {
 			options.LabelSelector = fmt.Sprintf("%s=%s", label.IoIstioRev.Name, revision)
 		})
+	retryBackoff := backoff.NewExponentialBackOff()
+	retryBackoff.InitialInterval = 2 * time.Second
+	retryBackoff.MaxInterval = 15 * time.Minute // Max Backoff interval.
+	// If certificate has not been patched with in an hour of retries, let us stop retrying.
+	retryBackoff.MaxElapsedTime = 1 * time.Hour
```
howardjohn, line 83: This means we may never recover from a temporary outage or permission error, I am not sure if that is safe. 15 mins is already really long internal anyhow?

Monkeyanator:
> shouldn't leader election ensure only one patches?

Mutating webhook patching isn't protected by leader election since each revision should only patch its own webhooks. In general no two istiod's should try patching the same webhook unless they have the same revision name (maybe an in-cluster and external control plane having the same revision name could lead to to this?).

Monkeyanator:
I just realized this is flawed, we can have multiple replicas on same revision that fight with each other for updates...

ramaraochavali:
Yeah multiple replicas of same revision are fighting with  each other.
I will remove the 1 hour cap and push it?

hzxuzhonghu:
This could not fix the patch conflict if the certificates is un equal in different istiod servers

hzxuzhonghu:
In order to fix the conflict, i think we need to do leader election

hzxuzhonghu (commented):
What I proposed before in a pr is puh a unique key for each task as well.

Pure backoff retry can make things worse, for exampl, if two tasks operating one object, one is v1 and one is v2, with retry we could not guarantee the execution order on the two versions.
If v1 is executed later, the worst case is the system down.

ramaraochavali:
> This could not fix the patch conflict if the certificates is un equal in different istiod servers

Yes. We are not expecting this to solve that problem. If during some period if the certificates are different, this should not bring down etcd. That is what we are trying to fix.

> In order to fix the conflict, i think we need to do leader election

Leader election can not be done because of multiple revisions. Each Istiod has to patch its own revision.

> Pure backoff retry can make things worse, for exampl, if two tasks operating one object, one is v1 and one is v2, with retry we could not guarantee the execution order on the two versions.
If v1 is executed later, the worst case is the system down.

In general that may be true. That is why I just enabled it for patching so that atleast it wont bring down etcd during the inconsistency.

ramaraochavali:
@howardjohn removed max elapsed time limit. PTAL

hzxuzhonghu:
I am ok with this, leaves to @howardjohn to have a look. I will file an issue about the potential  bug

pkg/webhooks/webhookpatch.go
```diff
@@ -75,12 +76,17 @@ func NewWebhookCertPatcher(
 		func(options *metav1.ListOptions) {
 			options.LabelSelector = fmt.Sprintf("%s=%s", label.IoIstioRev.Name, revision)
 		})
+	retryBackoff := backoff.NewExponentialBackOff()
+	retryBackoff.InitialInterval = 2 * time.Second
+	retryBackoff.MaxInterval = 15 * time.Minute // Max Backoff interval.
+	// If certificate has not been patched with in an hour of retries, let us stop retrying.
```
howardjohn, line 82: nit: comment is wrong now

howardjohn (approved):

itsLucario:
@ramaraochavali Any info on why this issue causes memory usage to bulk up and terminate pod?
We observed the same issue but instead of etcd crash, we saw high memory usage and istiod scale up constantly.

ramaraochavali:
@itsLucario Sorry. No idea.,

PR #28228, opened 2020-10-23
Update rather than patch webhook configuration

howardjohn:
This is a far more flexible pattern, allowing us to have multiple
webhooks and patch them successful. This pattern follows what the
cert-manager does in their webhook patcher (see
pkg/controller/cainjector), which I consider to be top quality code.

This also aligns with the validating webhook patcher

PR #35245, opened 2021-09-17
use patch instead of update in webhook patching

ramaraochavali:
"Patch" seems ideal in webhook patching instead of "update". This PR moves it to use patch.

Any reasons why we use "Update"?

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/webhooks/webhookpatch.go
+++ b/pkg/webhooks/webhookpatch.go
@@ -17,6 +17,7 @@ package webhooks
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
Whatever we do we ought to do the same for validation as mutating IMO.

In the past we switch from patch -> update. I think there was some issues with lists of webhooks behaving unexpectedly

ramaraochavali:
@howardjohn Thanks. I see you have moved it from patch -> update https://github.com/istio/istio/pull/28228/files. I will close this PR.

PR #35253, opened 2021-09-17
SetupMonitoring should return immediately when error occurred

silenceshell:

`SetupMonitoring` should return immediately when error occurred, otherwise we will use nil value In the following processing.

### DIFF
--- a/cni/pkg/monitoring/monitoring.go
+++ b/cni/pkg/monitoring/monitoring.go
@@ -35,10 +35,12 @@ func SetupMonitoring(port int, path string, stop <-chan struct{}) {
 	var err error
 	if listener, err = net.Listen("tcp", fmt.Sprintf(":%d", port)); err != nil {
 		log.Errorf("unable to listen on socket: %v", err)
+		return
 	}
 	exporter, err := ocprom.NewExporter(ocprom.Options{Registry: prometheus.DefaultRegisterer.(*prometheus.Registry)})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cni/pkg/monitoring/monitoring.go
```diff
@@ -35,10 +35,12 @@ func SetupMonitoring(port int, path string, stop <-chan struct{}) {
 	var err error
 	if listener, err = net.Listen("tcp", fmt.Sprintf(":%d", port)); err != nil {
 		log.Errorf("unable to listen on socket: %v", err)
+		return
```
bianpengyuan, line 38: I don't think we want to skip prometheus set up if UDS listen fails.

silenceshell: here is an [example](https://goplay.space/#5heCMngj6Dq) show that when `net.Listen` failed and we still use the `listener`(which is nil in this case) , we will get an panic:
`panic: runtime error: invalid memory address or nil pointer dereference`

If metrics listening is critical, maybe we should just exit the `install-cni` command, like [StartUDSLogServer](https://github.com/istio/istio/blob/261c55897e83acaa0108d15c9fd4455a37c141c0/cni/pkg/cmd/root.go#L71)?

bianpengyuan: Oh sorry, I was confused. I thought this is about setting up listener for UDS logging. LGTM

bianpengyuan (approved):

PR #35261, opened 2021-09-17
Add Hostname to leader election name for improve visibility

howardjohn:
This is only used in cases when pod name is not set, to help with
debugging. Typically this would happen in a local dev setting, but could
happen in external control plane.

### DIFF
--- a/pilot/pkg/leaderelection/leaderelection.go
+++ b/pilot/pkg/leaderelection/leaderelection.go
@@ -16,6 +16,8 @@ package leaderelection
 
 import (
 	"context"
+	"fmt"
+	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


hzxuzhonghu (commented):
you donot need to pass in name, os.Hostname equals that value for pod

howardjohn:
> you donot need to pass in name, os.Hostname equals that value for pod

I think it can be overriden. I slightly prefer to keep passing in `name` with fallback to `os.Hostname` for cases where users do not have a useful os.Hostname set (running off-k8s).

This is all for debug-ability though so its not the most critical what the name is

hzxuzhonghu (approved):

PR #35272, opened 2021-09-18
Add cabundle watcher to ns controller

hzxuzhonghu:

So when cabundle changes, it will reconcile all namespace's configmap

### DIFF
--- a/pilot/pkg/bootstrap/certcontroller.go
+++ b/pilot/pkg/bootstrap/certcontroller.go
@@ -25,8 +25,6 @@ import (
 	"time"
 
 	"istio.io/istio/pilot/pkg/features"
-	"istio.io/istio/pilot/pkg/keycertbundle"
-	"istio.io/istio/pilot/pkg/serviceregistry/kube/controller"
 	"istio.io/istio/pkg/config/constants"
 	"istio.io/istio/pkg/security"
 	"istio.io/istio/security/pkg/k8s/chiron"
@@ -41,6 +39,9 @@ const (
 
 	// defaultMinCertGracePeriod is the default minimum grace period for workload cert rotation.
 	defaultMinCertGracePeriod = 10 * time.Minute
+
+	// the interval polling root cert and re sign istiod cert when it changes.
+	rootCertPollingInterval = 60 * time.Second
 
 	// Default CA certificate path
 	// Currently, custom CA path is not supported; no API to get custom CA cert yet.
@@ -167,7 +168,7 @@ func (s *Server) watchRootCertAndGenKeyCert(stop <-chan struct{}) {
 		select {
 		case <-stop:
 			return
-		case <-time.After(controller.NamespaceResyncPeriod):
+		case <-time.After(rootCertPollingInterval):
 			newRootCert := s.CA.GetCAKeyCertBundle().GetRootCertPem()
 			if !bytes.Equal(caBundle, newRootCert) {
 				caBundle = newRootCert
@@ -239,13 +240,13 @@ func (s *Server) initCertificateWatches(tlsOptions TLSOptions) error {
 	return nil
 }
 
-func (s *Server) reloadIstiodCert(watchCh <-chan keycertbundle.KeyCertBundle, stopCh <-chan struct{}) {
+func (s *Server) reloadIstiodCert(watchCh <-chan struct{}, stopCh <-chan struct{}) {
 	for {
 		select {
 		case <-stopCh:
 			return
-		default:
-			if err := s.loadIstiodCert(watchCh, stopCh); err != nil {
+		case <-watchCh:
+			if err := s.loadIstiodCert(); err != nil {
 				log.Errorf("reload istiod cert failed: %v", err)
 			}
 		}
@@ -253,13 +254,8 @@ func (s *Server) reloadIstiodCert(watchCh <-chan keycertbundle.KeyCertBundle, st
 }
 
 // loadIstiodCert load IstiodCert received from watchCh once
-func (s *Server) loadIstiodCert(watchCh <-chan keycertbundle.KeyCertBundle, stopCh <-chan struct{}) error {
-	var keyCertBundle keycertbundle.KeyCertBundle
-	select {
-	case keyCertBundle = <-watchCh:
-	case <-stopCh:
-		return nil
-	}
+func (s *Server) loadIstiodCert() error {
+	keyCertBundle := s.istiodCertBundleWatcher.GetKeyCertBundle()
 	keyPair, err := tls.X509KeyPair(keyCertBundle.CertPem, keyCertBundle.KeyPem)
 	if err != nil {
 		return fmt.Errorf("istiod loading x509 key pairs failed: %v", err)
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -939,11 +939,10 @@ func (s *Server) initRegistryEventHandlers() {
 }
 
 func (s *Server) initIstiodCertLoader() error {
-	neverStop := make(chan struct{})
-	watchCh := s.istiodCertBundleWatcher.AddWatcher()
-	if err := s.loadIstiodCert(watchCh, neverStop); err != nil {
+	if err := s.loadIstiodCert(); err != nil {
 		return fmt.Errorf("first time load IstiodCert failed: %v", err)
 	}
+	_, watchCh := s.istiodCertBundleWatcher.AddWatcher()
 	s.addStartFunc(func(stop <-chan struct{}) error {
 		go s.reloadIstiodCert(watchCh, stop)
 		return nil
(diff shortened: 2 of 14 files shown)


pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
```diff
@@ -135,6 +132,39 @@ func (nc *NamespaceController) Run(stopCh <-chan struct{}) {
...
+		if err != nil {
+			return err
+		}
+
+		for _, ns := range namespaceList {
+			cm, err := nc.configmapLister.ConfigMaps(ns.Name).Get(CACertNamespaceConfigMap)
+			if err != nil {
+				continue
+			}
+			err = nc.configMapChange(cm)
```
howardjohn, line 151: Seems like we should push them on the queue. Although your issue about queue improvement will make it better

hzxuzhonghu: it is in the `task` cb, which is pushed to the queue

howardjohn: Yeah but we are pushing one task which writes 1000s of events, instead of pushing 1000 tasks that write 1 event.

If any 1 of the 1000 namespaces fails we would retry the whole thing

pilot/pkg/bootstrap/certcontroller.go
```diff
@@ -167,7 +166,7 @@ func (s *Server) watchRootCertAndGenKeyCert(stop <-chan struct{}) {
 		select {
 		case <-stop:
 			return
-		case <-time.After(controller.NamespaceResyncPeriod):
+		case <-time.After(time.Second * 60):
```
lei-tang, line 169: Why replace controller.NamespaceResyncPeriod with the hardcoded value of time.Second * 60?

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
shankgan: Agree with John's comment - think its a good idea to have each namespace update as a separate task. Also, does it make sense to change the task definition to actually include the caBundle as an argument? Wondering what happens if two changes to the caBundle happen quickly, so that one set of namespaces get updated with update1 and the other with update2? Eventually, all of them will reconcile to the latest, but could this be an issue?

shankgan (approved):

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
```diff
@@ -146,7 +176,7 @@ func (nc *NamespaceController) insertDataForNamespace(ns string) error {
 		Namespace: ns,
 		Labels:    configMapLabel,
 	}
-	return k8s.InsertDataToConfigMap(nc.client, nc.configmapLister, meta, nc.getData())
+	return k8s.InsertDataToConfigMap(nc.client, nc.configmapLister, meta, nc.caBundleWatcher.GetCABundle())
```
shankgan, line 179: This is a fairly critical update without which workloads will not be able to communicate with istiod. Does it make sense to re-add the task item to the queue if there is a failure?

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
hzxuzhonghu: IC,  i am ok with this

pilot/pkg/bootstrap/certcontroller.go
hzxuzhonghu: Will make it a const in the pkg

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -227,7 +227,7 @@ func (m *Multicluster) AddMemberCluster(clusterID cluster.ID, rc *secretcontroll
 				NewLeaderElection(options.SystemNamespace, m.serverID, leaderelection.NamespaceController, client.Kube()).
 				AddRunFunction(func(leaderStop <-chan struct{}) {
 					log.Infof("starting namespace controller for cluster %s", clusterID)
-					nc := NewNamespaceController(m.fetchCaRoot, client)
+					nc := NewNamespaceController(client, m.caBundleWatcher)
```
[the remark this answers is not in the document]
hzxuzhonghu: good catch， removed as we use the caBundleWatcher

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
hzxuzhonghu: Have already done for `insertDataForNamespace`

pilot/pkg/bootstrap/server.go
```diff
@@ -1130,30 +1130,6 @@ func (s *Server) startCA(caOpts *caOptions) {
...
-	}
-
-	// For Kubernetes CA, we don't distribute it; it is mounted in all pods by Kubernetes.
-	if features.PilotCertProvider == constants.CertProviderKubernetes {
-		return nil
-	}
-	// For no CA we don't distribute it either, as there is no cert
-	if features.PilotCertProvider == constants.CertProviderNone {
-		return nil
-	}
```
howardjohn, line 1150: Sounds like we are losing this functionality which will break a lot of users?

howardjohn (changes requested):
lgtm except the lost PilotCertProvider parts

pilot/pkg/serviceregistry/kube/controller/multicluster.go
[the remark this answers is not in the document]
shankgan: not sure if its okay to completely remove it though. Not sure if the two remaining flags encapsulate all the cases;  even for the simplest single cluster installations, depending on the value of PILOT_CERT_PROVIDER, we need logic to decide whether to disseminate the root cert or not - logic that was encapsulated in fetchCARoot.

pilot/pkg/bootstrap/server.go
hzxuzhonghu: IC, in this case, ns controller should not be started

pilot/pkg/keycertbundle/watcher.go
```diff
@@ -67,7 +84,9 @@ func (w *Watcher) SetAndNotify(key, cert, caBundle []byte) {
 	}
 	w.initDone.Store(true)
 	for _, ch := range w.watchers {
-		ch <- w.bundle
+		go func() {
+			ch <- w.bundle
```
howardjohn, line 88: This seems risky; if we have a bug where a channel is never read we may leak goroutine forever. Probably low risk though.

One idea to make it safe would be to use `ch` as just a `struct{}` carrier to signal to the consumer to then call `w.FetchBundle()` (or similar). Then we can make it have a buffered size of 1 and skip if its full.

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
```diff
@@ -135,6 +132,31 @@ func (nc *NamespaceController) Run(stopCh <-chan struct{}) {
 	}
 	log.Infof("Namespace controller started")
 	go nc.queue.Run(stopCh)
+	go nc.startCaBundleWatcher(stopCh)
+}
+
+// startCaBundleWatcher listens for updates to the CA bundle and update cm in each namespace
+func (nc *NamespaceController) startCaBundleWatcher(stop <-chan struct{}) {
+	id, watchCh := nc.caBundleWatcher.AddWatcher()
```
howardjohn, line 140: nit:
```suggestion
	watchCh, cancel := nc.caBundleWatcher.AddWatcher()
	defer cancel()
```

May be more idiomatic interface

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
```diff
@@ -135,6 +132,31 @@ func (nc *NamespaceController) Run(stopCh <-chan struct{}) {
...
+	defer nc.caBundleWatcher.RemoveWatcher(id)
```
howardjohn, line 141: Why do we need the remove watch now? Is it in case we lose leader election, we stop ?

pilot/pkg/bootstrap/server.go
```diff
@@ -1108,28 +1108,7 @@ func (s *Server) maybeCreateCA(caOpts *caOptions) error {
...
-			s.RunCA(grpcServer, s.CA, caOpts)
-		}
-		return nil
-	})
-}
-
+// TODO: remove it, currently used to indicate whether to start namespace controller
 func (s *Server) fetchCARoot() map[string]string {
 	if s.isDisableCa() {
 		return map[string]string{
```
howardjohn, line 1114: Not very familiar with this code but I am not certain that we actually respect the `string(s.RA.GetCAKeyCertBundle().GetRootCertPem())`  part?

We probably need more test coverage for this code..

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
hzxuzhonghu: Yes, to prevent leak, not only leader election, but also multi-cluster cleanup

pilot/pkg/bootstrap/server.go
hzxuzhonghu: Yes, this code is very tricky, i want to cleanup it later

pilot/pkg/keycertbundle/watcher.go
```diff
@@ -30,26 +30,43 @@ type KeyCertBundle struct {
...
+func (w *Watcher) AddWatcher() (int32, chan KeyCertBundle) {
 	ch := make(chan KeyCertBundle, 1)
 	w.mutex.Lock()
 	defer w.mutex.Unlock()
 	if w.initDone.Load() {
 		ch <- w.bundle
 	}
-	w.watchers = append(w.watchers, ch)
-	return ch
+	id := w.watcherId.Load()
```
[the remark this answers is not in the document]
hzxuzhonghu: I missed the lock

pilot/pkg/keycertbundle/watcher.go
hzxuzhonghu: Good suggestion, just use the channel to notify signal

hzxuzhonghu:
ping @howardjohn

howardjohn (approved):
thanks!

PR #35273, opened 2021-09-18
Remove unused error from GetService method

nmittler:
This has been a source of confusion in the code. No controller actually returns an error and there is no description in the interface for when an error should be returned.


howardjohn (approved):
Much better

PR #35279, opened 2021-09-20
Istio-Agent SecretCache: Dedup additional trustAnchors in dataplane f…

shankgan:
…or backward compatibility with filebased certs

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -721,29 +721,48 @@ func concatCerts(certsPEM []string) []byte {
...
 }
 
 func (sc *SecretManagerClient) mergeConfigTrustBundle(rootCert []byte) []byte {
-	return pkiutil.AppendCertByte(sc.getConfigTrustBundle(), rootCert)
+	duplicate := false
+	rootCertBytes := []byte{}
+	sc.configTrustBundleMutex.RLock()
+	_, duplicate = sc.configTrustBundle[string(rootCert)]
+	for cert := range sc.configTrustBundle {
+		rootCertBytes = pkiutil.AppendCertByte(rootCertBytes, []byte(cert))
```
howardjohn, line 761: order here is not determinstic

howardjohn (commented):
IIUC the original issue was about ansible and istio-agent constantly reconciling the same root-cert file?

So won't we still have that issue, even if we dedupe it? I assume ansible is just looking for any changes to file?

security/pkg/nodeagent/cache/secretcache.go
shankgan: sorry, that's correct. I can add in a sorted manner, that should resolve the issue right?

shankgan: Could also add code prevent a write to the file if its contents are not changing. That will resolve the issue

shankgan: fixed this

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -721,29 +722,57 @@ func concatCerts(certsPEM []string) []byte {
...
+	for cert := range newTbSet {
+		if _, ok := sc.configTrustBundle[cert]; !ok {
+			update = true
+		}
+	}
+	for cert := range sc.configTrustBundle {
+		if _, ok := newTbSet[cert]; !ok {
+			update = true
+		}
+	}
```
myidpt, line 746: Can we simply use reflect.DeepEqual(newTbSet, sc.configTrustBundle)?

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -721,29 +722,57 @@ func concatCerts(certsPEM []string) []byte {
...
 
 func (sc *SecretManagerClient) mergeConfigTrustBundle(rootCert []byte) []byte {
-	return pkiutil.AppendCertByte(sc.getConfigTrustBundle(), rootCert)
+	duplicate := false
+	certs := []string{}
+	rootCertBytes := []byte{}
+
+	sc.configTrustBundleMutex.RLock()
+	if len(rootCert) != 0 {
+		_, duplicate = sc.configTrustBundle[string(rootCert)]
```
myidpt, line 763: Seeing the places mergeConfigTrustBundle() is called, it seems rootCert could contain multiple certs, right?

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -721,29 +722,57 @@ func concatCerts(certsPEM []string) []byte {
...
+		certs = append(certs, cert)
+	}
+	sc.configTrustBundleMutex.RUnlock()
+
+	if !duplicate && len(rootCert) != 0 {
+		certs = append(certs, string(rootCert))
+	}
+	sort.Strings(certs)
+	for _, cert := range certs {
+		rootCertBytes = pkiutil.AppendCertByte(rootCertBytes, []byte(cert))
```
myidpt, line 775: I see this function is returning the merged trust bundle to the caller, but configTrustBundle isn't updated. So configTrustBundle only stores the root from dynamic proxyConfig, is that true?

PR #35280, opened 2021-09-20
Enable telemetry with external control plane

frankbu:

### DIFF
--- a/tests/integration/iop-externalistiod-remote-integration-test-defaults.yaml
+++ b/tests/integration/iop-externalistiod-remote-integration-test-defaults.yaml
@@ -25,3 +25,5 @@ spec:
   values:
     global:
       externalIstiod: true
+    telemetry:
+      enabled: false
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -373,9 +373,10 @@ gen: \
 gen-check: gen check-clean-repo
 
 copy-templates:
+	rm manifests/charts/istiod-remote/templates/*
+	rm manifests/charts/gateways/istio-egress/templates/*
+
 	# gateway charts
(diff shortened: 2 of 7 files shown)


tests/integration/iop-externalistiod-remote-integration-test-defaults.yaml
```diff
@@ -25,3 +25,5 @@ spec:
   values:
     global:
       externalIstiod: true
+    telemetry:
+      enabled: false
```
frankbu, line 29: This shouldn't be needed, because it's the default (https://github.com/istio/istio/blob/790d9eef8badd0c9a70caf3676c9c06fc1571ca5/manifests/charts/istiod-remote/values.yaml#L119), but without it the install fails while trying to apply the EnvoyFilters because the Istio CRDs are not defined in a remote cluster.

^^^ @howardjohn Any idea what I'm missing?

howardjohn: I am pretty sure `profiles/default.yaml` has precedence over values.yaml

frankbu: Hmm, since `profiles/default.yaml` explicitly sets every setting, what can be changed in values.yaml? This also makes me wonder if we need to explicitly set enabled=false in `profiles/external.yaml`. I guess it's not a bad idea to set it there, whether or not it's needed, just for clarity.

morvencao: `profiles/external.yaml` > `profiles/default.yaml` > `values.yaml`
so set the `enabled=false` in `profiles/external.yaml` makes sense, IMO.

howardjohn (approved):

PR #35281, opened 2021-09-20
Add per-revision leader election for webhook patching

Monkeyanator:
Potential fix for the problem addressed in #35243, adds a new kind of leader election that is winnable by a single replica from each revision (for operations like webhook patching).

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/bootstrap/sidecarinjector.go
+++ b/pilot/pkg/bootstrap/sidecarinjector.go
@@ -91,7 +91,7 @@ func (s *Server) initSidecarInjector(args *PilotArgs) (*inject.Webhook, error) {
 	// This requires RBAC permissions - a low-priv Istiod should not attempt to patch but rely on
 	// operator or CI/CD
 	if features.InjectionWebhookConfigName != "" {
-		s.addStartFunc(func(stop <-chan struct{}) error {
+		s.addTerminatingStartFunc(func(stop <-chan struct{}) error {
 			go leaderelection.NewPerRevisionLeaderElection(args.Namespace, args.PodName, args.Revision, leaderelection.MutatingWebhookController, s.kubeClient.Kube()).
 				AddRunFunction(func(stop <-chan struct{}) {
 					// update webhook configuration by watching the cabundle
--- a/pilot/pkg/bootstrap/sidecarinjector.go
+++ b/pilot/pkg/bootstrap/sidecarinjector.go
@@ -95,12 +95,7 @@ func (s *Server) initSidecarInjector(args *PilotArgs) (*inject.Webhook, error) {
 			go leaderelection.NewPerRevisionLeaderElection(args.Namespace, args.PodName, args.Revision, leaderelection.MutatingWebhookController, s.kubeClient.Kube()).
 				AddRunFunction(func(stop <-chan struct{}) {
 					// update webhook configuration by watching the cabundle
-					patcher, err := webhooks.NewWebhookCertPatcher(s.kubeClient, args.Revision, webhookName, s.istiodCertBundleWatcher)
-					if err != nil {
-						// HANDLE THIS ERROR BETTER, CANNOT JUST RETURN?
-						log.Errorf("failed to create webhook cert patcher: %v", err)
-						return
-					}
+					patcher := webhooks.NewWebhookCertPatcher(s.kubeClient, args.Revision, webhookName, s.istiodCertBundleWatcher)
 					patcher.Run(stop)
 				}).Run(stop)
 			return nil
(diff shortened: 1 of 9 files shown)


pilot/pkg/bootstrap/sidecarinjector.go
```diff
@@ -91,15 +92,17 @@ func (s *Server) initSidecarInjector(args *PilotArgs) (*inject.Webhook, error) {
...
-				return nil
-			}
-
-			patcher.Run(stop)
+			go leaderelection.NewPerRevisionLeaderElection(args.Namespace, args.PodName, args.Revision, leaderelection.MutatingWebhookController, s.kubeClient.Kube()).
+				AddRunFunction(func(stop <-chan struct{}) {
+					// update webhook configuration by watching the cabundle
+					patcher, err := webhooks.NewWebhookCertPatcher(s.kubeClient, args.Revision, webhookName, s.istiodCertBundleWatcher)
+					if err != nil {
+						// HANDLE THIS ERROR BETTER, CANNOT JUST RETURN?
```
howardjohn, line 100: afaik NewWebhookCertPatcher never returns an error, maybe we can just change the function to not return one

pilot/pkg/bootstrap/sidecarinjector.go
```diff
@@ -91,15 +92,17 @@ func (s *Server) initSidecarInjector(args *PilotArgs) (*inject.Webhook, error) {
...
-			// No leader election - different istiod revisions will patch their own cert.
-			// update webhook configuration by watching the cabundle
-			patcher, err := webhooks.NewWebhookCertPatcher(s.kubeClient, args.Revision, webhookName, s.istiodCertBundleWatcher)
-			if err != nil {
-				log.Errorf("failed to create webhook cert patcher: %v", err)
-				return nil
-			}
-
-			patcher.Run(stop)
+			go leaderelection.NewPerRevisionLeaderElection(args.Namespace, args.PodName, args.Revision, leaderelection.MutatingWebhookController, s.kubeClient.Kube()).
```
howardjohn, line 95: addTerminatingStartFunc is slightly preferred, will give us a better chance at releasing the lock on shutdown (so another instance can immediately start).

pilot/pkg/leaderelection/leaderelection.go
```diff
@@ -136,3 +142,25 @@ func NewLeaderElection(namespace, name, electionID string, client kubernetes.Int
...
+// For instance, in webhook patching, we want each revision to handle its own patching, but only a
+// single replica within that revision.
+func NewPerRevisionLeaderElection(namespace, name, revision, electionID string, client kubernetes.Interface) *LeaderElection {
+	if name == "" {
+		name = "unknown"
+	}
+	if revision != "" {
+		electionID = fmt.Sprintf("%s-%s", electionID, revision)
+	}
+	return &LeaderElection{
```
howardjohn, line 156: nit: make this call `NewLeaderElection` and just have the function modify the `electionID` arg

pilot/pkg/bootstrap/sidecarinjector.go
Monkeyanator: Done.

pilot/pkg/bootstrap/sidecarinjector.go
Monkeyanator: Cool did not know about this, done.

pilot/pkg/leaderelection/leaderelection.go
Monkeyanator: Done.

pilot/pkg/leaderelection/leaderelection.go
```diff
@@ -136,3 +142,15 @@ func NewLeaderElection(namespace, name, electionID string, client kubernetes.Int
 		cycle: atomic.NewInt32(0),
 	}
 }
+
+// NewPerRevisionLeaderElection calls a leader election scoped to the replicas of a given revision.
+// For instance, in webhook patching, we want each revision to handle its own patching, but only a
+// single replica within that revision.
+func NewPerRevisionLeaderElection(namespace, name, revision, electionID string, client kubernetes.Interface) *LeaderElection {
```
hzxuzhonghu, line 149: `NewLeaderElection` is already a general construction,  prefer reuse it instead of being aware of revision

hzxuzhonghu (commented):
I think all  leader elections in istio should be per revision

howardjohn:
> I think all leader elections in istio should be per revision
Sme of them are modifying global things so we should only have 1 leader, even across revisions. The idea we have for those is that the "default" revision will have precedence

Monkeyanator:
From Environments meeting this isn't a big issue for now since the main result is error logs and we should defer until default leader election work.

hzxuzhonghu:
I remember some configs like webhookconfiguration name has revision in it, not sure all of them.

howardjohn:
There are others that are truely global though, like the namespace
configmap writer or status updater

Monkeyanator:
Reopening since it seems like there are real issues that can be fixed with having this: https://github.com/istio/istio/issues/36478.

PR #35282, opened 2021-09-20
Use a separate service for the MCS host

nmittler:
The initial attempt at adding the Kubernetes Multi-Cluster Services (MCS) ClusterSet host (clusterset.local) revealed that Istio relies heavily on the property that there be 1 hostname per service.

This PR reverts some recent work that added the MCS hostname to `model.Service`. This PR splits out the MCS host into a separate `model.Service` that is synthetically generated by the `ServiceImportCache`. This synthetic service is updated whenever the real service is updated, or whenever there is a change to the associated `ServiceImport` resource.


tests/integration/pilot/mcs/discoverability/discoverability_test.go
```diff
@@ -266,3 +288,83 @@ func createAndCleanupServiceExport(t framework.TestContext, service string, clus
...
+	}
+
+	dummySvcName := "clusterset-vip-" + serviceB
+	dummySvc := &kubeCore.Service{
+		ObjectMeta: kubeMeta.ObjectMeta{
+			Name:      dummySvcName,
+			Namespace: testNS,
+			Labels: map[string]string{
+				// TODO(nmittler): Implement this in Istio.
+				"testing.istio.io/ignored-service": "true",
```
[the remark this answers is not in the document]
nmittler: > Not sure I understand why we need this. The real MCS will not have this, so why does the test?

The reference MCS controller uses services and it's expected that real implementations may as well. The MCS spec is going to have to make recommendations in this regard. But in the meantime we can probably just use the same label that MCS defines for `EndpointSlice`.

> I would also think we can just pick a random VIP for the test?

Possibly. But the code here should work and, if so, it's probably better just to do this, no?

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -450,70 +489,88 @@ func (c *Controller) onServiceEvent(curr interface{}, event model.Event) error {
...
-			c.nodeSelectorsForServices[svcConv.ClusterLocal.Hostname] = nodeSelector
-			c.Unlock()
-			needsFullPush = c.updateServiceNodePortAddresses(svcConv)
-		}
+		c.addOrUpdateService(svc, svcConv, event)
+	}
 
-		// instance conversion is only required when service is added/updated.
-		instances := kube.ExternalNameServiceInstances(svc, svcConv)
+	return nil
```
howardjohn, line 501: very nit: seems like 
```
	c.opts.XDSUpdater.SvcUpdate(shard, string(svcConv.ClusterLocal.Hostname), svc.Namespace, event)
	// Notify service handlers.
	for _, f := range c.serviceHandlers {
		f(svcConv, event)
	}
``

could remain here since its the same for all events

howardjohn (approved):
lgtm. While it looks a bit complex, its mostly because we are undoing other complexity we introduced previously

pilot/pkg/serviceregistry/kube/controller/controller.go
nmittler: I copied to each of the separate methods so that I could call them independently without calling `onServiceEvent` manually, which seems icky.

jacob-delgado (approved):
lgtm, looks like a lot of things are getting reverted and cleaned up

nmittler:
@jacob-delgado 

> lgtm, looks like a lot of things are getting reverted and cleaned up

Yeah, this actually simplifies things a lot. I'll also be following this with a PR to revert the introduction of `Service.ClusterLocal`, moving the hostname and vips back to the top-level. Didn't want to do it here since it would make a big ugly diff even bigger and uglier.

PR #35285, opened 2021-09-20
Fix tests so they can run off-GCP

howardjohn:
This was relying on real GCP metadata. Should be using all mock data for
tests


lei-tang (approved):
Thanks John for fixing this test!

PR #35287, opened 2021-09-21
Remove `-m comment` from iptables output

howardjohn:
We are already emiting the logs which has the command name as part of
it, so its a bit redundant to store the full comment in the kernel. This
has caused some compatibility issues with kernels not supporting
`comment` module
We never shipped this so its backwards compat


costinm (approved):
Thanks !

PR #35290, opened 2021-09-21
allow discoveryAddress to be set via proxyConfig for VMs with revision

stevenctl:
similar to #34060 – need to do more investigation on the safety and triple check the test does what I think it does 

* This should make `CA_ADDR`, `PILOT_ADDRESS` and `ISTIO_PILOT_PORT` vars still work in a backward compatible way, but avoid overriding with the default if the user didn't customize the envs. 
* The workload command now properly sets the port for the discovery address
* The workload command sets the CA_ADDR to match the discovery address (if rev is specified)
* The tests actually copy `sidecar.env` 
* The tests actually deploy a VM
* The test send traffic from the VM to make sure it connected to istiod properly
* The tests deploy a VirtualService pointing to the revisioned version of the eastwest gateway (this may need some cleanup to match the specific SNI)

### DIFF
--- a/samples/multicluster/expose-istiod-rev.yaml
+++ b/samples/multicluster/expose-istiod-rev.yaml
@@ -26,7 +26,7 @@ spec:
 apiVersion: networking.istio.io/v1alpha3
 kind: VirtualService
 metadata:
-  name: istiod-vs-{{.Revision}}
+  name: istiod-vs
 spec:
   hosts:
   - "*"
--- a/samples/multicluster/expose-istiod-rev.yaml
+++ b/samples/multicluster/expose-istiod-rev.yaml
@@ -1,0 +1,54 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: Gateway
+metadata:
+  name: istiod-gateway
+spec:
+  selector:
+    istio: eastwestgateway
+  servers:
+    - port:
+        name: tls-istiod
+        number: 15012
+        protocol: tls
+      tls:
+        mode: PASSTHROUGH        
+      hosts:
+        - "*"
+    - port:
+        name: tls-istiodwebhook
+        number: 15017
+        protocol: tls
+      tls:
+        mode: PASSTHROUGH          
+      hosts:
+        - "*"
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: istiod-vs-{{.Revision}}
+spec:
+  hosts:
+  - "*"
+  gateways:
+  - istiod-gateway
+  tls:
+  - match:
+    - port: 15012
+      sniHosts:
+      - "*"
+    route:
+    - destination:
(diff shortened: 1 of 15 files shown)


tools/packaging/common/istio-start.sh
```diff
@@ -116,11 +117,18 @@ ISTIO_AGENT_FLAGS=${ISTIO_AGENT_FLAGS:-}
...
-export PROXY_CONFIG=${PROXY_CONFIG:-"
+DEFAULT_PROXY_CONFIG="
 serviceCluster: $SVC
 controlPlaneAuthPolicy: ${CONTROL_PLANE_AUTH_POLICY}
-discoveryAddress: ${PILOT_ADDRESS}
-"}
+"
+if [ -n "${CUSTOM_PILOT_ADDRESS}" ]; then
+  PROXY_CONFIG="$PROXY_CONFIG
+  discoveryAddress: ${CUSTOM_PILOT_ADDRESS}
```
howardjohn, line 126:
```suggestion
discoveryAddress: ${CUSTOM_PILOT_ADDRESS}
```
spacing is suspicious here?

tests/integration/pilot/revisions/revisions_test.go
```diff
@@ -95,9 +95,10 @@ func TestMultiRevision(t *testing.T) {
 				}).
 				// tests bootstrap
 				WithConfig(echo.Config{
-					Service:   "vm",
-					Namespace: canary,
-					Ports:     []echo.Port{},
+					Service:    "vm",
+					Namespace:  canary,
+					DeployAsVM: true,
```
howardjohn, line 100: oof

howardjohn (approved):

tests/integration/pilot/revisions/revisions_test.go
stevenctl: I still feel like the test should fail. We use https://github.com/istio/istio/blob/master/samples/multicluster/expose-istiod.yaml which doesn't use the revisioned version. I'm running manually to verify

samples/multicluster/expose-istiod-rev.yaml
```diff
@@ -0,0 +1,54 @@
...
+spec:
+  hosts:
+  - "*"
+  gateways:
+  - istiod-gateway
+  tls:
+  - match:
+    - port: 15012
+      sniHosts:
+      - "*"
```
stevenctl, line 39: Assuming  this should be `istiod-rev.*`? @howardjohn

howardjohn: I don't think prefix match is supported

Monkeyanator (approved):

xulingqing:
@istio/wg-user-experience-maintainers could u take a look? Thanks!

therealmitchconnors (approved):

PR #35293, opened 2021-09-21
compute sidecar for gateway when service entry is added

ramaraochavali:
When service entry and gateway are added at the same time, service entry is not sent. Regression from this https://github.com/istio/istio/commit/ec1766dd7e315b2e43bb5497e008a8cd36271a9c.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/releasenotes/notes/35293.yaml
+++ b/releasenotes/notes/35293.yaml
@@ -2,7 +2,7 @@ apiVersion: release-notes/v2
 kind: bug-fix
 area: traffic-management
 issue:
-- 35068
+- 35172
 releaseNotes:
 - |
-  **Fixed** An issue when service entry and gateway created at same time, the service entry is ignored some times.
+  **Fixed** an issue when creating a Service and Gateway at the same time, causing the Service to be ignored.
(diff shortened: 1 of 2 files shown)


releasenotes/notes/35293.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+issue:
+- 35068
+releaseNotes:
+- |
+  **Fixed** An issue when service entry and gateway created at same time, the service entry is ignored some times.
```
ericvn, line 8: Maybe **Fixed** an issue when creating a ServiceEntry and Gateway at the same time where the ServiceEntry is occasionally not created.

releasenotes/notes/35293.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+issue:
+- 35068
```
ericvn, line 5: 35068 doesn't seem to be correct.

PR #35295, opened 2021-09-21
workload cmd: set discoveryaddress via ISTIOD_ADDRESS

stevenctl:
safer, less robust, alternative to #35290

### DIFF
--- a/istioctl/cmd/workload.go
+++ b/istioctl/cmd/workload.go
@@ -71,6 +71,7 @@ var (
 	labels         []string
 	annotations    []string
 	svcAcctAnn     string
+	xdsPort        string
 )
 
 const (
@@ -606,7 +607,8 @@ func revisionedIstiodAddress(rev string) string {
 	if rev == "" || rev == "default" {
 		return ""
 	}
-	return fmt.Sprintf("istiod-%s.%s.svc.cluster.local", rev, istioNamespace)
+	// TODO make port configurable
+	return fmt.Sprintf("istiod-%s.%s.svc.cluster.local:15012", rev, istioNamespace)
 }
 
 // Returns a map with each k,v entry on a new line
--- a/istioctl/cmd/workload.go
+++ b/istioctl/cmd/workload.go
@@ -301,10 +301,12 @@ func createConfig(kubeClient kube.ExtendedClient, wg *clientv1alpha3.WorkloadGro
 	if proxyConfig, err = createMeshConfig(kubeClient, wg, clusterID, outputDir); err != nil {
 		return err
 	}
+	rev := kubeClient.Revision()
+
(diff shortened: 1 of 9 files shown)


pkg/test/framework/components/namespace/kube.go
```diff
@@ -76,6 +79,34 @@ func (n *kubeNamespace) Prefix() string {
...
+	}
+	if err := errG.Wait().ErrorOrNil(); err != nil {
+		return nil, err
+	}
+	for i, clusterLabels := range perCluster {
+		if i == 0 {
+			continue
+		}
+		if diff := cmp.Diff(perCluster[0], clusterLabels); diff != "" {
+			log.Warnf("namespace labels are different accross clusters:\n%s", diff)
```
stevenctl, line 104:
```suggestion
			log.Warnf("namespace labels are different across clusters:\n%s", diff)
```

istioctl/cmd/workload.go
```diff
@@ -378,9 +381,12 @@ func createSidecarEnv(internalIP string, externalIP, dir string) error {
 	return os.WriteFile(filepath.Join(dir, "sidecar.env"), []byte(mapToString(sidecarEnv)), filePerms)
 }
 
-func generateSidecarEnvAsMap(internalIP string, externalIP string) map[string]string {
+func generateSidecarEnvAsMap(internalIP string, externalIP string, rev string) map[string]string {
 	sidecarEnv := make(map[string]string)
 
+	if addr := revisionedIstiodAddress(rev); addr != "" {
+		sidecarEnv["PILOT_ADDRESS"] = addr
```
howardjohn, line 388: Why do we need this if we have the EXPLICITLY_SET_PILOT_ADDRESS thing?

howardjohn (commented):
Did we ever figure out why the test wasn't effective?

istioctl/cmd/workload.go
stevenctl: We don't -- I didn't revert it correctly.

stevenctl:
> Did we ever figure out why the test wasn't effective?

One issue is that we didn't actually pass `--revision` to `istioctl`, so we weren't even setting the value. 
This means the discovery address was defaulting to `istiod.istio-system...` which I guess still exists in the revisions test.

tests/integration/pilot/revisions/revisions_test.go
```diff
@@ -95,9 +95,10 @@ func TestMultiRevision(t *testing.T) {
 				}).
 				// tests bootstrap
 				WithConfig(echo.Config{
-					Service:   "vm",
-					Namespace: canary,
-					Ports:     []echo.Port{},
+					Service:    "vm",
+					Namespace:  canary,
+					DeployAsVM: true,
```
stevenctl, line 100: This is also why the test didn't give coverage...

PR #35308, opened 2021-09-22
Refactor istio queue to support eventual consistent model

### LINKED ISSUE #35269: Refactor istio queue
Currently, istio queue allows use push a task handler when a config updated, and the queue is run as a daemon thread.

If one task handler executes fails, then it will be retried with a fixed delay.

This could cause severe bug,  let's say

1. t0:  enqueue a task handler, which handles an objectXX with v0
2. t1: enqueue a task handler, which handles an objectXX with v1

The queue does not know the two handlers will handle the same object. So if 

3. t3:  execute on v0 objectXX  failed, it will retry later
4. t4:  execute on v1 objectXX succeeds
5. t5:  retry handling v0 objectXX , then it will violate eventual consistence

hzxuzhonghu:

Fix: #35269

### DIFF
--- a/pilot/pkg/config/kube/ingress/controller.go
+++ b/pilot/pkg/config/kube/ingress/controller.go
@@ -80,7 +80,7 @@ type controller struct {
 	meshWatcher  mesh.Holder
 	domainSuffix string
 
-	queue                  queue.Instance
+	queue                  queue.Queue
 	virtualServiceHandlers []model.EventHandler
 	gatewayHandlers        []model.EventHandler
 
@@ -138,7 +138,7 @@ func NetworkingIngressAvailable(client kube.Client) bool {
 func NewController(client kube.Client, meshWatcher mesh.Holder,
 	options kubecontroller.Options) model.ConfigStoreCache {
 	// queue requires a time duration for a retry delay after a handler error
-	q := queue.NewQueue(1 * time.Second)
+	q := queue.NewFixedDelayQueue(1 * time.Second)
 
 	if ingressNamespace == "" {
 		ingressNamespace = constants.IstioIngressNamespace
@@ -170,19 +170,31 @@ func NewController(client kube.Client, meshWatcher mesh.Holder,
 	ingressInformer.AddEventHandler(
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) {
-				q.Push(func() error {
+				key, err := keyFunc(obj)
+				if err != nil {
+					return
+				}
+				q.Push(key, func() error {
 					return c.onEvent(nil, obj, model.EventAdd)
 				})
 			},
 			UpdateFunc: func(old, cur interface{}) {
 				if !reflect.DeepEqual(old, cur) {
-					q.Push(func() error {
+					key, err := keyFunc(cur)
+					if err != nil {
+						return
+					}
+					q.Push(key, func() error {
 						return c.onEvent(old, cur, model.EventUpdate)
 					})
 				}
 			},
 			DeleteFunc: func(obj interface{}) {
-				q.Push(func() error {
+				key, err := keyFunc(obj)
+				if err != nil {
+					return
+				}
+				q.Push(key, func() error {
 					return c.onEvent(nil, obj, model.EventDelete)
 				})
 			},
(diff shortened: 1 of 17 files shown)


stevenctl:
need to take a deeper look, but in general i think its good

hzxuzhonghu:
The basic implement is referenced from k8s workqueue. One concern is the risk about it.

pilot/pkg/config/kube/ingress/controller.go
```diff
@@ -170,19 +170,31 @@ func NewController(client kube.Client, meshWatcher mesh.Holder,
 	ingressInformer.AddEventHandler(
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) {
-				q.Push(func() error {
+				key, err := keyFunc(obj)
+				if err != nil {
+					return
+				}
+				q.Push(key, func() error {
```
howardjohn, line 177: I don't think this is safe. 

There are three queue models:
* Enqueue tasks. This just means we naively process each event, in order. This existing Istio impl
* Enqueue keys. When we process the key, we _reconcile_ the state based on the current state of the world. This is k8s impl
* Enqueue keys and tasks. This is the PR here. This will not work because we are mixing the two states.

For example, say I have:
1. Update ingress update, from class=istio to class=ignored.
2. Delete ingress.

Before, we would notice we stopped watching the class, and trigger an update. Then later the delete comes in, and we see the oldObj is class=ignore, so we do nothing.

With this new change, the delete event will override the update event. This causes us to only process a "delete class=ignored" event, which we do nothing for. This will lead to stale config.

While its valid to enqueue tasks and its valid to enqueue keys, they are not drop in replacements. The actual queue processors need to behave quite differently for the different cases.

I am fine with refactoring our queue usage, but I think it should be incremental and we should not mix keys+tasks.

howardjohn (changes requested):

pilot/pkg/config/kube/ingress/controller.go
hzxuzhonghu: I donot think the example you put up is a problem. For an ingress, we only cares about whether we should process it and what gateway/vs it can be converted.

Changing the class --> delete it, should result in no gateway/vs associated with it. 

I think with/without this pr, the ingress event handling is both not quite right.


> Update ingress update, from class=istio to class=ignored. 

will not trigger vs/gateway delete event, i think we need an separate fix

hzxuzhonghu: Other than that, i could not find why cannot we mix keys+tasks. The task is just to get the latest object since we are unifying the queue reconcile.

howardjohn: > Other than that, i could not find why cannot we mix keys+tasks. The task is just to get the latest object since we are unifying the queue reconcile.

I suppose we can but if the task isn't always the same its probably a bug... so seems like we should associate the task with the queue instance, not the task. This is more like how k8s queue works?

pkg/queue/queue.go
```diff
@@ -0,0 +1,210 @@
...
+			log.Info("queue is shutting down")
+			return
+		}
+
+		err := innerTask.task()
+		q.done(innerTask.item)
+		if err != nil {
+			delay := innerTask.NextBackOff()
+			if delay == backoff.Stop {
+				log.Errorf("Drop work item (%v) because of exceeding MaxElapsedTime")
```
pangsq, line 150: `%v` is lack of the `innerTask.item`

pkg/queue/queue.go
```diff
@@ -0,0 +1,210 @@
...
+	cond    *sync.Cond
+	closing bool
+}
+
+func NewBackOffQueue(backoff BackOffOption) Queue {
+	return &queue{
+		backoffOpts: backoff,
+		tasks:       make([]*innerTask, 0),
+		processing:  make(map[t]empty),
+		dirty:       make(map[interface{}]*innerTask),
```
pangsq, line 89: it's wired to see that using `t` as `interface{}` in `processing`, but using `interface{}` in `dirty`. Or just make `processing: make(set)`?

hzxuzhonghu: Need the latest task recorded in dirty

pkg/queue/queue.go
hzxuzhonghu: right

howardjohn:
Aside from my above comment, it seems overly risky to change everything at once. Maybe we can have two queue types and slowly move controllers over. Especially the high risk ones like NS controller, while keeping low/zero risk ones (due to no errors in queue tasks) like kube controller which have huge blast radius untouched for now.

hzxuzhonghu:
Totally agree， this is too risky as all the related usage are hot path.

hzxuzhonghu:
Close, we should fix one by one if neccessary

Issue #34395, opened 2021-07-28
WorkloadEntry telemetry does not report correct canonical_revision


nrfox:
**Bug description**

Given a `WorkloadEntry` with a `version: v1` label and generated from `istioctl workload entry configure ...`, telemetry for the workload reports `source_canonical_revision="latest"` instead of `v1`. Telemetry does correctly report `source_version="v1"` and `version="v1"` though.

WorkloadEntry
```
apiVersion: networking.istio.io/v1beta1
kind: WorkloadEntry
metadata:
  name: details-v1
  namespace: bookinfo
spec:
  address: 10.10.10.10
  labels:
    app: details
    version: v1
  network: Kubernetes
  serviceAccount: details-service-account
```

When the workload entry VM files are created with `istioctl workload entry configure`, the `"service.istio.io/canonical-version"` label gets generated but not the `"service.istio.io/canonical-revision"` label.

mesh.yaml
```
defaultConfig:
  discoveryAddress: istiod.istio-system.svc:15012
  meshId: mesh1
  proxyMetadata:
    CANONICAL_REVISION: v1
    CANONICAL_SERVICE: details
    ISTIO_META_CLUSTER_ID: Kubernetes
    ISTIO_META_DNS_CAPTURE: "true"
    ISTIO_META_MESH_ID: mesh1
    ISTIO_META_NETWORK: Kubernetes
    ISTIO_META_WORKLOAD_NAME: details-v1
    ISTIO_METAJSON_LABELS: '{"app":"details","service.istio.io/canonical-name":"details","service.istio.io/canonical-version":"v1","version":"v1"}'
    POD_NAMESPACE: bookinfo
    SERVICE_ACCOUNT: vm-service-account
    TRUST_DOMAIN: cluster.local
  tracing:
    zipkin:
      address: zipkin.istio-system:9411
```

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[X] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

Expected `canonical_revision` in telemetry to match `version` label.

**Steps to reproduce the bug**

1. Create a `WorkloadGroup` template with a `version: v1` label.
2. Run `istioctl workload entry configure ...` to create the `WorkloadEntry` and VM config files (mesh.yaml, cluster.env, etc.)
[...]

PR #35318, opened 2021-09-22
fix canonical revision label for workload groups

douglas-reid:
An incorrect label was used in the `workload` command, leading to incorrectly labeled `WorkloadGroup`s. This PR corrects the issue, first identified in #34395.


bianpengyuan (approved):
Do we have any integration test for this? Also probably worth a release notes.

istioctl/cmd/workload.go
```diff
@@ -530,7 +530,7 @@ func createMeshConfig(kubeClient kube.ExtendedClient, wg *clientv1alpha3.Workloa
 	}
 	md["ISTIO_META_WORKLOAD_NAME"] = wg.Name
 	labels["service.istio.io/canonical-name"] = md["CANONICAL_SERVICE"]
-	labels["service.istio.io/canonical-version"] = md["CANONICAL_REVISION"]
+	labels["service.istio.io/canonical-revision"] = md["CANONICAL_REVISION"]
```
howardjohn, line 533: do we have this in istio.io/api/labels package? would avoid this

howardjohn (approved):

douglas-reid:
> Do we have any integration test for this? Also probably worth a release notes.

I don't believe we have any integration tests for the `workload` command (or generated telemetry from the impacted workloads). There are the golden file-based unit tests updated in this PR, but that's all I'm aware of.

Release notes have been added.

stevenctl (approved):

stevenctl:
The VMs in integration tests are set up using this code, so it should be possible to test the generated telemetry.

douglas-reid:
> The VMs in integration tests are set up using this code, so it should be possible to test the generated telemetry.

Do you have a quick pointer?  The telemetry tests on VMs are limited, and seem to manually label instances ([example](https://github.com/istio/istio/blob/master/tests/integration/telemetry/stackdriver/vm/main_test.go#L166)).

stevenctl:
The things in `vmEnv` just get appended to the output from `workload` command:

https://github.com/istio/istio/blob/4ccc36879f85faba3e78ed0b96c8230559be6d63/pkg/test/framework/components/echo/kube/deployment.go#L983

You can (probably) remove a good amount of that customization and cover what we generate in the command.

douglas-reid:
> You can (probably) remove a good amount of that customization and cover what we generate in the command.

Sounds good. Filed #35319 to track follow-up to remove customization.

therealmitchconnors (approved):

PR #35321, opened 2021-09-22
Remove ClusterLocal structure from Service

nmittler:
This was originally added to support a second set of host/vips for Kubernetes Multi-Cluster Services (MCS). We've since gone with a different approach and this is no longer needed.

Moving the hostname and cluster VIPs to the top-level in Service.


hzxuzhonghu:
@nmittler what's the different approach?

nmittler:
@hzxuzhonghu  we're now using a separate (synthetic) service to represent the MCS `clusterset.local` host. So many of the previous changes are effectively now reverted.

hzxuzhonghu:
Seems just one change: svc hostname revert. LGTM

howardjohn (approved):

Issue #31590, opened 2021-03-19
Expose mTLS state in access logs


howardjohn:
Its a common request to see if a request was using mTLS or not. While there are various ways to do that, one great way would be access logs. This is especially handy in auto-mtls world when we may not know if we will use mTLS or not.

JimmyCYJ:
cc @incfly

PR #35337, opened 2021-09-23
Add downstream SAN to access logs

howardjohn:
Example:
```
default/shell-7854df9c5-9zfnp[istio-proxy]: [2021-09-23T23:40:40.289Z] "GET / HTTP/1.1" 200 - via_upstream - "-" 0 603 8 8 "-" "curl/7.69.1" "51531b24-e16f-4ac9-ab6c-667711893a60" "echo" "10.244.0.12:80" outbound|80||echo.default.svc.cluster.local 10.244.0.11:35560 10.96.116.201:80 10.244.0.11:39818 - - default
default/echo-cb96f8d94-722gq[istio-proxy]: [2021-09-23T23:40:40.289Z] "GET / HTTP/1.1" 200 - via_upstream - "-" 0 603 0 0 "-" "curl/7.69.1" "51531b24-e16f-4ac9-ab6c-667711893a60" "echo" "10.244.0.12:80" inbound|80|| 127.0.0.6:58943 10.244.0.12:80 10.244.0.11:35560 outbound_.80_._.echo.default.svc.cluster.local spiffe://cluster.local/ns/default/sa/default default
```

Note we only get this on the server side.

Fixes https://github.com/istio/istio/issues/31590
Mostly opening for discussion, I am not convinced we should do this. I think its useful to tell if mtls is used but may not be worth the churn

cc @douglas-reid @yangminzhu


pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -47,7 +47,7 @@ const (
 		"%DURATION% %RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)% \"%REQ(X-FORWARDED-FOR)%\" " +
 		"\"%REQ(USER-AGENT)%\" \"%REQ(X-REQUEST-ID)%\" \"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\" " +
 		"%UPSTREAM_CLUSTER% %UPSTREAM_LOCAL_ADDRESS% %DOWNSTREAM_LOCAL_ADDRESS% " +
-		"%DOWNSTREAM_REMOTE_ADDRESS% %REQUESTED_SERVER_NAME% %ROUTE_NAME%\n"
+		"%DOWNSTREAM_REMOTE_ADDRESS% %REQUESTED_SERVER_NAME% %DOWNSTREAM_PEER_URI_SAN% %ROUTE_NAME%\n"
```
yangminzhu, line 50: Is the current `REQUESTED_SERVER_NAME` enough for telling the mTLS status or maybe `DOWNSTREAM_TLS_VERSION` is better as it is more straightforward and covers both TLS and mTLS?

`DOWNSTREAM_PEER_URI_SAN` is also fine to me as it gives more information that can be useful in general debugging.

Either way, please also remember to update the documentation of the default access log: https://istio.io/latest/docs/tasks/observability/logs/access-log/#default-access-log-format

howardjohn: > REQUESTED_SERVER_NAME

That's a good point actually.. That makes this less compelling

howardjohn:
Probably not worth it since mtls info is already present in SNI (more or less)

Issue #31833, opened 2021-03-31
Ingress resources are not keyed on namespace, leading to conflicts


howardjohn:
https://github.com/istio/istio/blob/release-1.6/pilot/pkg/config/kube/ingress/conversion.go#L126-L127

`ingressNamespace` here should probably be `ingress.Namespace`. Alternatively, we could put `ingress.Namespace` in the name.

howardjohn:
not stale

PR #35339, opened 2021-09-23
Fix ingress name conflict

howardjohn:
Fixes https://github.com/istio/istio/issues/31833

I am fairly sure, but not 100% sure, that this will not break anyone as the name is internal impl detail only.


stevenctl (approved):

hzxuzhonghu (commented):
why not make the ingress ns as gateway ns

howardjohn:
Our recommendation has been to have the Gateway live in the same namespace
as the Deployment so I figured I would keep it. There is even a env var to
enforce it happens

hzxuzhonghu (commented):
As you can see the controller.List method is not quite right, the ns passed in is not equal to returned.
Not a big matter, i believe this is a legacy issue.

Issue #35332, opened 2021-09-23
Multi Cluster EKS Istio submitting invalid Envoy config 


jwilner:
### Bug Description

I'm trying to set up a multi-network multi-primary mesh in EKS between two regions. Per https://github.com/istio/istio/issues/29359, I am setting up EIPs strapped onto my EW NLB and notifying Istio about them via the istio cm. When I add workloads following the [multi cluster verify runbook](https://istio.io/latest/docs/setup/install/multicluster/verify/) though, envoy gets very mad at the config Istio is sending it:

```
2021-09-23T13:08:01.396294Z warn    ads ADS:EDS: ACK ERROR sleep-557747455f-8mtxg.sample-105 Internal:Proto constraint validation failed (ClusterLoadAssignmentValidationError.Endpoints[0]: embedded message failed validation | caused by LocalityLbEndpointsValidationError.LbEndpoints[1]: embedded message failed validation | caused by LbEndpointValidationError.LoadBalancingWeight: value must be greater than or equal to 1): cluster_name: "outbound|80||sleep.sample.svc.cluster.local"
endpoints {
  locality {
    region: "us-east-1"
    zone: "us-east-1a"
  }
  lb_endpoints {
    endpoint {
      address {
        socket_address {
          address: "US-EAST-1A-IP"
          port_value: 15443
        }
      }
    }
    metadata {
      filter_metadata {
        key: "envoy.transport_socket_match"
        value {
          fields {
            key: "tlsMode"
            value {
              string_value: "istio"
            }
          }
        }
      }
      filter_metadata {
        key: "istio"
        value {
          fields {
            key: "workload"
            value {
              string_value: ";;;;us-west-2-cluster"
            }
          }
        }
      }
    }
    load_balancing_weight {
      value: 1
    }
  }
  lb_endpoints {
    endpoint {
      address {
        socket_address {
          address: "US-EAST-1B-IP"
          port_value: 15443
        }
      }
    }
    metadata {
      filter_metadata {
        key: "envoy.transport_socket_match"
        value {
          fields {
            key: "tlsMode"
            value {
              string_value: "istio"
            }
          }
        }
      }
      filter_metadata {
        key: "istio"
        value {
          fields {
            key: "workload"
            value {
              string_value: ";;;;us-west-2-cluster"
            }
          }
        }
      }
    }
    load_balancing_weight {
    }
  }
...
```

[...]

PR #35346, opened 2021-09-24
[WIP] Make use of NetworkGateway value

hzxuzhonghu:

Try to fix: https://github.com/istio/istio/issues/29359#issuecomment-925823768

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu:
@sarvaniv @jwilner can you please try this, and be care of this log

https://github.com/istio/istio/pull/35346/files#diff-04af6226156fa3d0ed87d7da7a52526e9653acf4b2f7ef9522f2e7ca510b0efeR119

jwilner:
Thanks @hzxuzhonghu, I'll try this out! 

Just a heads up though -- GH issues is suggesting this PR fixes #29359 (supporting hostname-based gateways), but that's a much bigger issue that this PR definitely doesn't fix. I reported #35332 which describes this bug.

stevenctl:
@jwilner my bad, I missed that bug so I linked to the comment within the issue. Thanks for opening the other bug. 

@hzxuzhonghu I imagine the root of the issue is we now track the `cluster` a gateway is associated with, and the meshNetworks approach doesn't include that?

pilot/pkg/model/network.go
```diff
@@ -64,29 +64,28 @@ func NewNetworkManager(env *Environment) *NetworkManager {
...
-	for _, gw := range env.NetworkGateways() {
-		if gwIP := net.ParseIP(gw.Addr); gwIP != nil {
-			// - the internal map of label gateways - these get deleted if the service is deleted, updated if the ip changes etc.
-			// - the computed map from meshNetworks (triggered by reloadNetworkLookup, the ported logic from getGatewayAddresses)
-			gatewaySet[*gw] = struct{}{}
-		} else {
-			log.Warnf("Failed parsing gateway address %s from Service Registry. "+
-				"Hostnames are not supported for gateways",
-				gw.Addr)
+	} else {
```
stevenctl, line 67: I think we definitely want the ability to use both.

pilot/pkg/xds/ep_filters.go
```diff
@@ -106,16 +106,19 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocLbEndpointsAn
...
+			gateways = append(gateways, gw)
 		}
 		gateways = model.SortGateways(gateways)
 
 		// Create endpoints for the gateways.
 		for _, gw := range gateways {
-			epWeight := gatewayWeights[*gw]
+			epWeight := gatewayWeights[gw]
+			if epWeight == 0 {
+				log.Warnf("gateway weight must not be 0, scaleFactor is %d", scaleFactor)
```
stevenctl, line 119: Should this be `continue`? This seems to occur if our total `weight` to distribute across the gateways isn't large enough to actually utilize all the gateways. 

I think this should be `continue` for safety purposes, with a warning log, but the underlying issue is that scaleFactor isn't making the total weight large enough to utilize all the gws. I think we can test this by making gws > endpoints by some degree?

pilot/pkg/model/network.go
hzxuzhonghu: right

pilot/pkg/xds/ep_filters.go
hzxuzhonghu: This is just a debug info, in theory it could not enter, because the scaleFactor is the lcm of all gateways

hzxuzhonghu:
@jwilner Could you have a test with this patch?

savealive:
I've tested the patch. Everything looks good. 
```
2021-10-09T00:16:21.017157Z	info	version: 1.12-dev-5f0ba472e9c4119b9d46388f95834569b77cd550-Clean
```

hzxuzhonghu:
good to know, so we should make it in

PR #35347, opened 2021-09-24
add omit empty to metadata

ramaraochavali:
When we display /debug/connections we do not display metadata

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn (approved):

pilot/pkg/xds/debug.go
```diff
@@ -97,7 +97,7 @@ type AdsClient struct {
 	ConnectionID string              `json:"connectionId"`
 	ConnectedAt  time.Time           `json:"connectedAt"`
 	PeerAddress  string              `json:"address"`
-	Metadata     *model.NodeMetadata `json:"metadata"`
+	Metadata     *model.NodeMetadata `json:"metadata,omitempty"`
```
hzxuzhonghu, line 100: When can this field be empty

ramaraochavali: when /connections is displayed https://github.com/istio/istio/blob/master/pilot/pkg/xds/debug.go#L511

PR #35351, opened 2021-09-24
Allow publishing explicit `-debug` tagged images

### LINKED ISSUE #31711: distroless: publish xyz-ubuntu tagged images
Currently, we have tags like `pilot` and `pilot-distroless`. We should instead have: `pilot`, `pilot-debug` (or `pilot-ubuntu`), and `pilot-distroless`.

The plain `pilot` one will be floating - we can choose to make it `distroless` or `ubuntu` based on our preferences. In the future, this would likely become distroless.
Users that want to explicitly use one or the other should use one of the explicitly tagged ones.

howardjohn:
Fixes https://github.com/istio/istio/issues/31711

This shifts our images from three options:
* `-distroless` - explicit distroless
* `-debug` - explicit debug
* no suffix - whatever Istio project choses. Default now, distroless in
the future

Honestly this is pretty complex, the makefile is a bit of a mess..

### DIFF
--- a/cni/deployments/kubernetes/Dockerfile.install-cni
+++ b/cni/deployments/kubernetes/Dockerfile.install-cni
@@ -12,7 +12,7 @@ FROM gcr.io/istio-release/distroless:${BASE_VERSION} as distroless
 
 # This will build the final image based on either debug or distroless from above
 # hadolint ignore=DL3006
-FROM ${BASE_DISTRIBUTION}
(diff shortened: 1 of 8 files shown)


ericvn (approved):

morvencao (approved):

craigbox (approved):
only docs-relevant change is `samples/tcp-echo/src/Dockerfile` and it LGTM

nmittler (approved):

Issue #33034, opened 2021-05-22
Excessive test dump size


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/integ-security-multicluster-tests_istio_postsubmit/1396122398708731904 is one example, its >4gb. I think this may be leading to high IO which may impact other jobs on the same node as well possible

cc @stevenctl I think you were looking at this a bit

howardjohn:
That job took 14 minutes to upload! 6.23gb total

stevenctl:
It's mostly `proxy-config.json` each taking up ~1.5MB in that example. These types of things get duplicated with deeply nested subtests. Working on something to help reduce this by skipping the dump for leaf tests that don't have unique config changes.

stevenctl:
`| grep core.proxy` results in a lot of files around 180mb. We probably don't always want these.

stevenctl:
```
...
63651246/primary/statefulset-v1-0_core.proxy
173.63 MiB   gs://istio-prow/pr-logs/pull/istio_istio/33483/integ-pilot-multicluster-tests_istio/1410018472557023232/artifacts/pilot-endpointslice-93577615b20/TestTraffic/virtualservice/set_header/vm/to_vm/from_primary/_test_context/echo-1-69067-state663651246/primary/tproxy-v1-664db5df55-spr2w_core.proxy
172.62 MiB   gs://istio-prow/pr-logs/pull/istio_istio/33483/integ-pilot-multicluster-tests_istio/1410018472557023232/artifacts/pilot-endpointslice-93577615b20/TestTraffic/virtualservice/set_header/vm/to_vm/from_primary/_test_context/echo-1-69067-state663651246/remote/a-v1-6bfb5d5b9d-pwv8x_core.proxy
145.68 MiB   gs://istio-prow/pr-logs/pull/istio_istio/33483/integ-pilot-multicluster-tests_istio/1410018472557023232/artifacts/pilot-endpointslice-93577615b20/TestTraffic/virtualservice/set_header/vm/to_vm/from_primary/_test_context/echo-1-69067-state663651246/remote/b-v1-5db4fc987c-xjzl7_core.proxy
[...]

PR #35355, opened 2021-09-24
prevent writing large and identical core dumps in test artifacts

stevenctl:
related to https://github.com/istio/istio/issues/33034

### DIFF
--- a/pkg/test/kube/dump.go
+++ b/pkg/test/kube/dump.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 
 	"github.com/hashicorp/go-multierror"
+	"go.uber.org/atomic"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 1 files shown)


pkg/test/kube/dump.go
```diff
@@ -152,13 +161,27 @@ func DumpPods(ctx resource.Context, workDir, namespace string, selectors []strin
...
+		if coreDumpedPods.Load() >= maxCoreDumpedPods {
+			return
+		}
+		wroteDumpsForPod := false
 		containers := append(pod.Spec.Containers, pod.Spec.InitContainers...)
 		for _, container := range containers {
 			if container.Name != "istio-proxy" {
 				continue
 			}
+			restarts := containerRestarts(pod, "istio-proxy")
```
stevenctl, line 178: Will the core dumps even exist? I'm not sure this check is necessary.

pkg/test/kube/dump.go
```diff
@@ -36,6 +37,14 @@ import (
 	"istio.io/istio/pkg/test/scopes"
 )
 
+const (
+	maxCoreDumpedPods = 5
+)
+
+var (
+	coreDumpedPods = atomic.NewInt32(0)
```
howardjohn, line 45: I think this is fine but fwiw this will be a per-package limit I think, since each package is its own executable

pkg/test/kube/dump.go
howardjohn: There will only be a dump if it crashed I think, so this will reduce exec calls which I think is good.

howardjohn (approved):

pkg/test/kube/dump.go
stevenctl: > max 5 * suites is still only like 30

I may optimize later, didn't want to add complexity such as storing this in a file

PR #35357, opened 2021-09-24
Implement InsecureSkipVerify from DestinationRule

Kmoneal:
Implement InsecureSkipVerify from the DestinationRule to not verify certificates using a CA certificate or SAN.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -1052,7 +1052,7 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 		if opts.serviceRegistry == provider.External && len(tls.SubjectAltNames) == 0 {
 			tls.SubjectAltNames = opts.serviceAccounts
 		}
-		if tls.CredentialName != "" {
+		if tls.InsecureSkipVerify.Value == true || tls.CredentialName != "" {
 			tlsContext = &auth.UpstreamTlsContext{
 				CommonTlsContext: &auth.CommonTlsContext{},
 				Sni:              tls.Sni,
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -1052,7 +1052,15 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 		if opts.serviceRegistry == provider.External && len(tls.SubjectAltNames) == 0 {
 			tls.SubjectAltNames = opts.serviceAccounts
 		}
-		if tls.InsecureSkipVerify.Value == true || tls.CredentialName != "" {
+		if tls.InsecureSkipVerify.Value {
+			tls.CaCertificates = ""
+			tls.SubjectAltNames = []string{}
+
+			tlsContext = &auth.UpstreamTlsContext{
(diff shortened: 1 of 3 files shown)


pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1052,7 +1052,15 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 		if opts.serviceRegistry == provider.External && len(tls.SubjectAltNames) == 0 {
 			tls.SubjectAltNames = opts.serviceAccounts
 		}
-		if tls.CredentialName != "" {
+		if tls.InsecureSkipVerify != nil && tls.InsecureSkipVerify.Value {
+			tls.CaCertificates = ""
+			tls.SubjectAltNames = []string{}
```
hzxuzhonghu, line 1057: why do you change the input tls settings

hzxuzhonghu (commented):
/ok-to-test

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
Kmoneal: This is for testing but I am thinking it will be best to remove this and make an integration test to replace the unit test

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1052,7 +1052,15 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 		if opts.serviceRegistry == provider.External && len(tls.SubjectAltNames) == 0 {
 			tls.SubjectAltNames = opts.serviceAccounts
 		}
-		if tls.CredentialName != "" {
+		if tls.InsecureSkipVerify != nil && tls.InsecureSkipVerify.Value {
```
howardjohn, line 1055: Should we just make `InsecureSkipVerify` not allowed when CaCertificates or SubjectAltNames is set? In validation

Kmoneal: I have no problem with `InsecureSkipVerify` only taking effect when `VerifyCertificateAtClient` is in use

Kmoneal: After revisiting this, I am a bit more concerned about `InsecureSkipVerify` not doing what is expected from the name and how it is used in programs like curl. If you feel strongly about it only effecting `VerifyCertificateAtClient`, I would be happy to go ahead and make changes.

howardjohn: I don't mean anything to do with VerifyCertificateAtClient, I mean:
```
tls:
  caCertificates: /foo/bar
  insecureSkipVerify: true
```

make that invalid in the webhook. There is no need to set the SAN/ca certs and then not verify - it just opens room for confusion that they are actually used?

Kmoneal: Thanks for the clarification. I'll work on getting that up. Just wanted to make sure it is fine that it doesn't do exactly what [the RFC](https://docs.google.com/document/d/100Jvwjr9KkURiF2HtiOtRlAON8TqgskUyLfT17_xoRM/edit#heading=h.xw1gqgyqs5b) specified.

SpecialYang:
Is anyone still working on this? I want to support this new api field.
@howardjohn

Issue #35356, opened 2021-09-24
Distroless does not work with ipv6 or tproxy


howardjohn:
Missing `ip` command

PR #35359, opened 2021-09-24
More fixes for TPROXY + ipv6

howardjohn:
With this and a few other patches I will send a PR for, I have
IPv6+TPROXY+Distroless+CNI all working together. Currently we don't have
any tests of this but working to expand our environmental testing.

**Please provide a description of this PR:**

PR #35360, opened 2021-09-24
WIP: drop dependency on `iproute2` binary and libraries

howardjohn:
Fixes https://github.com/istio/istio/issues/35356

This adds a new dependency on `github.com/vishvananda/netlink` which was formerly an indirect dependency. In return, we can avoid pulling into multiple mbs of libraries unrelated to netlink just to run a few netlink commands.


howardjohn:
This is WIP just until https://github.com/istio/istio/pull/35359 merges to split up the PR a bit

bianpengyuan (commented):
lgtm

howardjohn:
> Generally LGTM.
> 
> The missing piece is the fwmark for ipv6. See below. The ipv4 and ipv6 iptables are independent
> 
> Feel free to ignore if the missing ipv6 is not a regression.
> 
> ```
> $ sudo ip rule add fwmark 123 lookup 123
> $ sudo ip rule
> 0:      from all lookup local
> 32765:  from all fwmark 0x7b lookup 123
> 32766:  from all lookup main
> 32767:  from all lookup default
> $ sudo ip -6 rule
> 0:      from all lookup local
> 32766:  from all lookup main
> ```

Not a regression. Will fix anyways

PR #35361, opened 2021-09-24
Update to grpc 1.41

howardjohn:

### DIFF
--- a/pilot/pkg/networking/grpcgen/lds.go
+++ b/pilot/pkg/networking/grpcgen/lds.go
@@ -33,6 +33,7 @@ import (
 	"istio.io/istio/pilot/pkg/security/authn"
 	"istio.io/istio/pilot/pkg/security/authn/factory"
 	"istio.io/istio/pilot/pkg/util/sets"
+	xdsfilters "istio.io/istio/pilot/pkg/xds/filters"
 	"istio.io/istio/pkg/config/labels"
 	"istio.io/istio/pkg/istio-agent/grpcxds"
 )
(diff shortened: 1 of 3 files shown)


stevenctl (approved):

pilot/pkg/networking/grpcgen/lds.go
```diff
@@ -147,7 +148,7 @@ func buildFilterChain(nameSuffix string, tlsContext *tls.DownstreamTlsContext) *
 			Name: "inbound-hcm" + nameSuffix,
 			ConfigType: &listener.Filter_TypedConfig{
 				TypedConfig: util.MessageToAny(&hcm.HttpConnectionManager{
-					// TODO gRPC doesn't support httpfilter yet; sending won't cause a NACK but they don't do anything
+					HttpFilters: []*hcm.HttpFilter{xdsfilters.Router},
```
stevenctl, line 151: Is there any reason we choose to set this? Since we don't have other filters, it shouldn't matter, but it seems like this just acts like a terminator to stop processing filters:

https://github.com/grpc/grpc-go/blob/ef66d13abb84ad6c6d99c8cbf3697607b7891f32/xds/internal/resolver/serviceconfig.go#L263-L266

howardjohn: they may it required: https://github.com/grpc/grpc-go/blame/458ea7640a92039aad37edc67b63e6d040a93320/xds/internal/xdsclient/xds.go#L260

costinm (approved):

PR #35364, opened 2021-09-25
Fixing MCS for EndpointSlice mode

nmittler:
When receiving a `ServiceExport`, the `serviceExportCache` calls into the kube endpoint controller to build the endpoints. However, when in `EndpointSlice` mode, this simply returns the pre-built endpoints in an internal cache. This means that the endpoints will never get the updated `EndpointDiscoverabilityPolicy`, allowing them to be discovered across the mesh.

This PR allows the `serviceExportCache` to request a rebuild of the cache for the service.

Also does some other minor improvements around debuggability (mcsz), etc.

### DIFF
--- a/pilot/pkg/model/service.go
+++ b/pilot/pkg/model/service.go
@@ -436,27 +436,53 @@ func (ep *IstioEndpoint) GetLoadBalancingWeight() uint32 {
 	return 1
 }
 
-// IsDiscoverableFromProxy indicates whether or not this endpoint is discoverable from the given Proxy.
+// IsDiscoverableFromProxy indicates whether this endpoint is discoverable from the given Proxy.
 func (ep *IstioEndpoint) IsDiscoverableFromProxy(p *Proxy) bool {
 	if ep == nil || ep.DiscoverabilityPolicy == nil {
 		// If no policy was assigned, default to discoverable mesh-wide.
+		// TODO(nmittler): Will need to re-think this default when cluster.local is actually cluster-local.
 		return true
 	}
-	return ep.DiscoverabilityPolicy(ep, p)
+	return ep.DiscoverabilityPolicy.IsDiscoverableFromProxy(ep, p)
 }
 
 // EndpointDiscoverabilityPolicy determines the discoverability of an endpoint throughout the mesh.
-type EndpointDiscoverabilityPolicy func(*IstioEndpoint, *Proxy) bool
+type EndpointDiscoverabilityPolicy interface {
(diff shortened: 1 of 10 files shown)


tests/integration/pilot/mcs/discoverability/discoverability_test.go
```diff
@@ -153,6 +160,7 @@ func enableMCSServiceDiscovery(_ resource.Context, cfg *istio.Config) {
 values:
   pilot:
     env:
+      PILOT_USE_ENDPOINT_SLICE: "true"
```
stevenctl, line 163: Not sure if we should revert before merging or not? 

#34971 will make it so this runs in ES mode for k8s 1.21+ and EP mode for earlier.

tests/integration/pilot/mcs/discoverability/discoverability_test.go
```diff
@@ -141,8 +142,14 @@ func installMCSCRDs(t resource.Context) error {
 		if err != nil {
 			return err
 		}
-		if err := t.Config().ApplyYAML("", string(crd)); err != nil {
-			return err
+		if t.Settings().NoCleanup {
```
stevenctl, line 145: Do we need an `ApplyYamlConditionalCleanup`?

pilot/pkg/model/service.go
```diff
@@ -436,27 +436,53 @@ func (ep *IstioEndpoint) GetLoadBalancingWeight() uint32 {
...
 }
 
 // EndpointDiscoverabilityPolicy determines the discoverability of an endpoint throughout the mesh.
-type EndpointDiscoverabilityPolicy func(*IstioEndpoint, *Proxy) bool
+type EndpointDiscoverabilityPolicy interface {
+	// IsDiscoverableFromProxy indicates whether an endpoint is discoverable from the given Proxy.
+	IsDiscoverableFromProxy(*IstioEndpoint, *Proxy) bool
+
+	// String returns name of this policy.
+	String() string
```
stevenctl, line 455: +1

tests/integration/pilot/mcs/discoverability/discoverability_test.go
nmittler: Let's revert in #34971. MCS relies on EndpointSlice anyway, so I think forcing this test to use EndpointSlice makes sense until your PR lands.

tests/integration/pilot/mcs/discoverability/discoverability_test.go
nmittler: Yeah that might help a bit. We can always refactor later.

stevenctl (approved):

PR #35379, opened 2021-09-27
ServiceEntry ignores inactive pods at EventAdd

pangsq:

ServiceEntry should ignore inactive pods on pod EventAdd.

If not, inactive pods (whose phase is Succeeded or Failed) will be used by ServiceEntry to make instances when pilot firstly received these pods' Add events.

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/pod.go
+++ b/pilot/pkg/serviceregistry/kube/controller/pod.go
@@ -134,6 +134,8 @@ func (pc *PodCache) onEvent(curr interface{}, ev model.Event) error {
 				if !IsPodReady(pod) {
 					ev = model.EventDelete
 				}
+			default:
+				return nil
 			}
 		case model.EventUpdate:
 			if pod.DeletionTimestamp != nil {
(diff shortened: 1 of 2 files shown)


pangsq:
A specific example to produce that bug:

1. Create resources:
-  a job with pod label(app=demo) and command(["sleep", "10"])
-  a serviceEntry  with workloadSelector.labels(app=demo)
-  a pod with sidecar injected

2. Wait until the pod created by the job is completed(whose phase is Succeeded). That completed pod will still have a PodIP
3. Restart istiod
4. `istioctl proxy-config endpoints <pod with sidecar injected>.<pod's namespace> | grep <completed pod's ip>`. We will see the ip still exist.

hzxuzhonghu (commented):
lgtm, compared with k8s endpoints controller, i think we still miss some considerations here.

It would be better to improve

pangsq:
I noticed that https://github.com/istio/istio/issues/35404 may be caused by similar reasons. 

And pod events are not only generated by informer, according to https://github.com/istio/istio/blob/1.12.0-alpha.1/pilot/pkg/serviceregistry/kube/controller/discoverycontrollers.go#L156.

PR #35380, opened 2021-09-27
Fix syntax error in stackdriver test file templates

frankbu:

I'm getting a syntax error in some of the stackdriver yaml files when running the telemetry tests using the `istiodremote` topology. One example of the problem is that `OnGCE` is false [here](https://github.com/istio/istio/blob/e24ded225705aeb4849014f0942260eee55b717b/tests/integration/telemetry/stackdriver/testdata/server_request_count.json.tmpl#L29). As a result, unmashalling the expanded template complains about the `,` on line 28. It seems that the comma needs to be moved inside the `{{- if .OnGCE }}`.

@douglas-reid I was also wondering if it sounds right to you that `OnGCE` is false when I run the tests? Where is that set? Have I missed something in the configuration for running the tests?

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience

### DIFF
--- a/tests/integration/telemetry/stackdriver/testdata/client_request_count.json.tmpl
+++ b/tests/integration/telemetry/stackdriver/testdata/client_request_count.json.tmpl
@@ -25,8 +25,7 @@
           "source_workload_name": "clt-{{ .ClusterName }}-v1",
           "source_workload_namespace": "{{ .EchoNamespace }}"
        }
-    },
-    {{- if .OnGCE  }}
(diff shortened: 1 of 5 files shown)


bianpengyuan (approved):
`OnGCE` is set by test pod looking up the metadata server. It will be true if the test pod running on GKE.

PR #35382, opened 2021-09-27
Integrate with GKE workload certificates

lei-tang:
https://github.com/istio/istio/issues/35385

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [X ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/cmd/pilot-agent/options/security.go
+++ b/pilot/cmd/pilot-agent/options/security.go
@@ -112,14 +112,23 @@ func SetupSecurityOptions(proxyConfig *meshconfig.ProxyConfig, secOpt *security.
 		o.CredFetcher = credFetcher
 	}
 
-	if UseGkeWorkloadCertificate(security.GkeWorkloadCertChainFilePath,
-		security.GkeWorkloadKeyFilePath, security.GkeWorkloadRootCertFilePath) {
+	if o.CAProviderName == security.GoogleGkeWorkloadCertificateProvider {
+		if !CheckGkeWorkloadCertificate(security.GkeWorkloadCertChainFilePath,
+			security.GkeWorkloadKeyFilePath, security.GkeWorkloadRootCertFilePath) {
+			return nil, fmt.Errorf("GKE workload certificate files (%v, %v, %v) not present",
+				security.GkeWorkloadCertChainFilePath, security.GkeWorkloadKeyFilePath, security.GkeWorkloadRootCertFilePath)
+		}
+		if o.ProvCert != "" {
+			return nil, fmt.Errorf(
+				"invalid options: PROV_CERT and FILE_MOUNTED_CERTS of GKE workload cert are mutually exclusive")
+		}
 		o.FileMountedCerts = true
 		o.CertChainFilePath = security.GkeWorkloadCertChainFilePath
 		o.KeyFilePath = security.GkeWorkloadKeyFilePath
 		o.RootCertFilePath = security.GkeWorkloadRootCertFilePath
 		return o, nil
 	}
+
 	// Default the CA provider where possible
 	if strings.Contains(o.CAEndpoint, "googleapis.com") {
 		o.CAProviderName = security.GoogleCAProvider
@@ -139,9 +148,9 @@ func SetupSecurityOptions(proxyConfig *meshconfig.ProxyConfig, secOpt *security.
 	return o, nil
 }
 
-// UseGkeWorkloadCertificate returns true when the GKE workload certificate
-// files are present under the well-known path. Otherwise, return false.
-func UseGkeWorkloadCertificate(certChainFilePath, keyFilePath, rootCertFilePath string) bool {
+// CheckGkeWorkloadCertificate returns true when the GKE workload certificate
+// files are present under the path for GKE workload certificate. Otherwise, return false.
+func CheckGkeWorkloadCertificate(certChainFilePath, keyFilePath, rootCertFilePath string) bool {
 	if _, err := os.Stat(certChainFilePath); err != nil {
 		return false
 	}
--- a/pilot/cmd/pilot-agent/options/security.go
+++ b/pilot/cmd/pilot-agent/options/security.go
@@ -16,6 +16,7 @@ package options
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -54,6 +55,9 @@ func NewSecurityOptions(proxyConfig *meshconfig.ProxyConfig, stsPort int, tokenM
 		STSPort:                        stsPort,
 		CertSigner:                     certSigner.Get(),
 		CARootPath:                     cafile.CACertFilePath,
+		CertChainFilePath:              security.DefaultCertChainFilePath,
+		KeyFilePath:                    security.DefaultKeyFilePath,
+		RootCertFilePath:               security.DefaultRootCertFilePath,
 	}
 
 	o, err := SetupSecurityOptions(proxyConfig, o, jwtPolicy.Get(),
@@ -107,6 +111,15 @@ func SetupSecurityOptions(proxyConfig *meshconfig.ProxyConfig, secOpt *security.
 		log.Infof("using credential fetcher of %s type in %s trust domain", credFetcherTypeEnv, o.TrustDomain)
 		o.CredFetcher = credFetcher
 	}
+
+	if UseGkeWorkloadCertificate(security.GkeWorkloadCertChainFilePath,
+		security.GkeWorkloadKeyFilePath, security.GkeWorkloadRootCertFilePath) {
+		o.FileMountedCerts = true
+		o.CertChainFilePath = security.GkeWorkloadCertChainFilePath
+		o.KeyFilePath = security.GkeWorkloadKeyFilePath
+		o.RootCertFilePath = security.GkeWorkloadRootCertFilePath
+		return o, nil
+	}
 	// Default the CA provider where possible
 	if strings.Contains(o.CAEndpoint, "googleapis.com") {
 		o.CAProviderName = security.GoogleCAProvider
@@ -124,4 +137,19 @@ func SetupSecurityOptions(proxyConfig *meshconfig.ProxyConfig, secOpt *security.
 		return nil, fmt.Errorf("invalid options: PROV_CERT and FILE_MOUNTED_CERTS are mutually exclusive")
 	}
 	return o, nil
+}
+
+// UseGkeWorkloadCertificate returns true when the GKE workload certificate
+// files are present under the well-known path. Otherwise, return false.
+func UseGkeWorkloadCertificate(certChainFilePath, keyFilePath, rootCertFilePath string) bool {
+	if _, err := os.Stat(certChainFilePath); err != nil {
+		return false
+	}
+	if _, err := os.Stat(keyFilePath); err != nil {
+		return false
+	}
+	if _, err := os.Stat(rootCertFilePath); err != nil {
+		return false
+	}
+	return true
 }
--- a/pkg/istio-agent/agent_test.go
+++ b/pkg/istio-agent/agent_test.go
@@ -93,11 +93,11 @@ func TestAgent(t *testing.T) {
 	t.Run("Kubernetes defaults", func(t *testing.T) {
 		// XDS and CA are both using JWT authentication and TLS. Root certificates distributed in
 		// configmap to each namespace.
-		Setup(t, nil).Check(t, security.WorkloadKeyCertResourceName, security.RootCertReqResourceName)
+		Setup(t).Check(t, security.WorkloadKeyCertResourceName, security.RootCertReqResourceName)
 	})
 	t.Run("RSA", func(t *testing.T) {
 		// All of the other tests use ECC for speed. Here we make sure RSA still works
-		Setup(t, nil, func(a AgentTest) AgentTest {
+		Setup(t, func(a AgentTest) AgentTest {
 			a.Security.ECCSigAlg = ""
 			return a
 		}).Check(t, security.WorkloadKeyCertResourceName, security.RootCertReqResourceName)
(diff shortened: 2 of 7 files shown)


pkg/security/security.go
```diff
@@ -42,6 +42,15 @@ const (
 	// DefaultRootCertFilePath is the well-known path for an existing root certificate file
 	DefaultRootCertFilePath = "./etc/certs/root-cert.pem"
 
+	// GkeWorkloadCertChainFilePath is the well-known path for the GKE workload certificate chain file
+	GkeWorkloadCertChainFilePath = "/var/run/secrets/workload-spiffe-credentials/certificates.pem"
```
howardjohn, line 46: would be nice to have a link to the gke docs for this?

pilot/cmd/pilot-agent/options/security.go
```diff
@@ -107,6 +111,15 @@ func SetupSecurityOptions(proxyConfig *meshconfig.ProxyConfig, secOpt *security.
 		log.Infof("using credential fetcher of %s type in %s trust domain", credFetcherTypeEnv, o.TrustDomain)
 		o.CredFetcher = credFetcher
 	}
+
+	if UseGkeWorkloadCertificate(security.GkeWorkloadCertChainFilePath,
```
howardjohn, line 115: This seems problematic unless its 100% guaranteed that the files will be present *before* the pod is started. Does the implementation guarantee that? In current usage, typically the files may be populated after the pod starts.

pkg/security/security.go
```diff
@@ -42,6 +42,15 @@ const (
...
+
+	// GkeWorkloadKeyFilePath is the well-known path for the GKE workload certificate key file
+	GkeWorkloadKeyFilePath = "/var/run/secrets/workload-spiffe-credentials/private_key.pem"
```
howardjohn, line 49:
```suggestion
	GkeWorkloadKeyFilePath = "./var/run/secrets/workload-spiffe-credentials/private_key.pem"
```

Along with the other three. This allows running outside of root and makes things more testable, etc.

In particular, once we do this we can and should have a test for this flow in TestAgent

howardjohn (commented):
overall looks good

pilot/cmd/pilot-agent/options/security.go
ahmedtd: GKE workload certificates are provided by a CSI driver.  The pod will not start until the CSI driver has successfully written the key, certificate, and trust anchor set to the pod filesystem.

pilot/cmd/pilot-agent/options/security.go
```diff
@@ -107,6 +111,15 @@ func SetupSecurityOptions(proxyConfig *meshconfig.ProxyConfig, secOpt *security.
...
+		security.GkeWorkloadKeyFilePath, security.GkeWorkloadRootCertFilePath) {
+		o.FileMountedCerts = true
```
shankgan, line 117: nit: It appears that workload certificates are going to be signed by a root that Istiod does not have in its peer cert verifier. If the pod is going to present its certificate to Istiod for control plane XDS mTLS, Istiod may not recognize it. Could this cause issues?

pkg/security/security.go
lei-tang: Added a link. Quoted from https://cloud.google.com/traffic-director/docs/security-proxyless-setup#create-service:
"On creation, each Pod gets a volume at /var/run/secrets/workload-spiffe-credentials."

pkg/security/security.go
```diff
@@ -198,6 +209,13 @@ type Options struct {
 
 	// Root Cert read from the OS
 	CARootPath string
+
+	// The path for an existing certificate chain file
+	CertChainFilePath string
+	// The path for an existing key file
+	KeyFilePath string
+	// The path for an existing root certificate file
+	RootCertFilePath string
```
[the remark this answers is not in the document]
lei-tang: Revised to "The path for an existing root certificate bundle".

pilot/cmd/pilot-agent/options/security.go
shankgan: based on offline discussion: this shouldn't be a problem - the agent will not be presenting this cert to Istiod for XDS connection bringup.

howardjohn: Test seems to pass but I would have thought this would stop XDS from using proper auth... https://github.com/istio/istio/blob/925e6e61704893fdbaa1e5bdf4b1b86115cdea37/pkg/istio-agent/xds_proxy.go#L654?

I guess tests pass since we never exercise this code path in our integ tests?

lei-tang: Can you elaborate on why the quoted code in xds_proxy.go would stop XDS from using proper auth?

howardjohn: Because you set FileMountedCerts and if that is set then we don't send a jwt to istiod for xds so it will break.

pkg/security/security.go
```diff
@@ -42,6 +42,17 @@ const (
 	// DefaultRootCertFilePath is the well-known path for an existing root certificate file
 	DefaultRootCertFilePath = "./etc/certs/root-cert.pem"
 
+	// GkeWorkloadCertChainFilePath is the well-known path for the GKE workload certificate chain file.
+	// Quoted from https://cloud.google.com/traffic-director/docs/security-proxyless-setup#create-service:
+	// "On creation, each Pod gets a volume at /var/run/secrets/workload-spiffe-credentials."
+	GkeWorkloadCertChainFilePath = "./var/run/secrets/workload-spiffe-credentials/certificates.pem"
```
howardjohn, line 48: This path is actually not a special value that is relevant to Istio. That only impacts the gke WI webhook - something we would have no need for given we already have our own webhook and will configure the volume ourself.

I suggest we use a more standard path - there is no need for anything GKE specific at all as far as I can tell. I recall we even have some similar code to automatically read from /etc/certs if it's present. It's possible we may not even need any code changes if we do that - or if we do, like less (perhaps different filename if we cannot configure that)

howardjohn (changes requested):

pilot/cmd/pilot-agent/options/security.go
lei-tang: Based on the comment in the following test for "File mounted certs" in https://github.com/istio/istio/blob/925e6e61704893fdbaa1e5bdf4b1b86115cdea37/pkg/istio-agent/agent_test.go#L136, when FileMountedCerts is used, _"mTLS is always used for authentication with
Istiod, never JWT."_ Futhermore,  ISTIO_META_TLS_CLIENT* need to be specified for the file paths of the certificate used to authenticate to istiod.

t.Run("File mounted certs", func(t *testing.T) {
		// User sets FileMountedCerts. They also need to set ISTIO_META_TLS_CLIENT* to specify the
		// file paths. CA communication is disabled. mTLS is always used for authentication with
		// Istiod, never JWT.

howardjohn: I agree. And you are setting FileMountedCerts, which means mTLS will be used to Istiod for XDS which is *not* what we want. 

The workload certificate is totally irrelevant to the XDS connection. 

I encourage you to add a test for this, which I believe will fail.

pilot/cmd/pilot-agent/options/security.go
lei-tang: The PR has been revised such that if o.CAProviderName == security.GoogleGkeWorkloadCertificateProvider, the code will check the presence of GKE workload certificate files.

pkg/security/security.go
lei-tang: Revised based on the comment and a test has been added for this flow in TestAgent.

pilot/cmd/pilot-agent/options/security.go
lei-tang: A test has been added for this flow in TestAgent.

pkg/security/security.go
lei-tang: The code has been revised to read from ./etc/certs.  The cert and key file names of GKE workload certificates differ from those of the default file mounted cert:  the cert and key file names of GKE workload certificate are "certificates.pem", "private_key.pem", and "ca_certificates.pem", whereas those of the default file mounted cert are "cert-chain.pem", "key.pem", and "root-cert.pem".

ahmedtd: The path /var/run/secrets/workload-spiffe-credentials/{certifcates.pem,private_key.pem,ca_certificates.pem} is shaping up to be a cross-product standard (envoy and grpc directly support that path).  If possible, we should keep that path the same.

lei-tang: The PR has been revised to use the standard path of GKE workload certificate (./var/run/secrets/workload-spiffe-credentials).

pkg/istio-agent/agent_test.go
```diff
@@ -159,6 +159,34 @@ func TestAgent(t *testing.T) {
...
+		// the default file mounted cert and key names.
+		copyGkeWorkloadCerts(t, dir)
+		cfg := security.SdsCertificateConfig{
+			CertificatePath:   filepath.Join(dir, "certificates.pem"),
+			PrivateKeyPath:    filepath.Join(dir, "private_key.pem"),
+			CaCertificatePath: filepath.Join(dir, "ca_certificates.pem"),
+		}
+		Setup(t, func(a AgentTest) AgentTest {
+			// Ensure we use the mTLS certs for XDS
+			a.XdsAuthenticator.Set("", preProvisionID)
```
howardjohn, line 179: We must NOT use mTLS for XDS.

howardjohn (changes requested):
XDS should continue to use token.

pkg/istio-agent/agent_test.go
lei-tang: Fixed. PTAL.

lei-tang:
@howardjohn @ahmedtd All review comments have been addressed. Can you take another look? Thanks!

pkg/security/security.go
howardjohn: Can you provide some reference to it becoming standard? I don't see any mention of it in the envoy or grpc repos. If it is a standard then sounds great, but we should link to a doc about it.

pkg/istio-agent/xds_proxy.go
```diff
@@ -650,9 +650,7 @@ func (p *XdsProxy) buildUpstreamClientDialOpts(sa *Agent) ([]grpc.DialOption, er
 		keepaliveOption, initialWindowSizeOption, initialConnWindowSizeOption, msgSizeOption,
 	}
 
-	if !sa.secOpts.FileMountedCerts {
-		dialOptions = append(dialOptions, grpc.WithPerRPCCredentials(caclient.NewXDSTokenProvider(sa.secOpts)))
-	}
+	dialOptions = append(dialOptions, grpc.WithPerRPCCredentials(caclient.NewXDSTokenProvider(sa.secOpts)))
```
howardjohn, line 653: @ramaraochavali  will this break you?

Before:
* Only mTLS used
* After: mTLS *and* token used

pkg/security/security.go
lei-tang: @ahmedtd Do you have a reference link for it?

pkg/istio-agent/agent_test.go
```diff
@@ -445,7 +465,7 @@ type AgentTest struct {
 	agent *Agent
 }
 
-func Setup(t *testing.T, opts ...func(a AgentTest) AgentTest) *AgentTest {
+func Setup(t *testing.T, sdsCertConfig *security.SdsCertificateConfig, opts ...func(a AgentTest) AgentTest) *AgentTest {
```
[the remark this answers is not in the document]
lei-tang: Thanks! Revised accordingly.

pkg/istio-agent/xds_proxy.go
howardjohn: I am a bit worried its actually `After: mTLS and token used and if token cannot be fetched fail hard` which may break VMs...

lei-tang: If one of the authenticators pass, the client authentication will pass. After adding a token authenticator, the worst case for the authentication at the newly added token authenticator is that the authentication at the token authenticator fails, which is equivalent to before adding the token authenticator. 
Therefore, I am not sure why adding a token authenticator will break the authentication. Can you elaborate?

howardjohn: I am not 100% sure it will but my concern was on the client side, not the server side. That gRPC will call the token provider which will send an error and break the whole request.

Maybe it never returns an error , would need to double check.

lei-tang: Thanks for the clarification. Based on the review comment that the token authentication must be used for XDS, attaching a token on the client side is inevitable.

howardjohn (approved):

pkg/istio-agent/xds_proxy.go
howardjohn: Thanks Lei. I did some testing and verified that we never return an error on the token provider - even if we cannot read the token. Ity would be great if you can make this change so that we are testing this:
```
diff --git a/pkg/istio-agent/agent_test.go b/pkg/istio-agent/agent_test.go
index 84aeafae1b..e80e469fd5 100644
--- a/pkg/istio-agent/agent_test.go
+++ b/pkg/istio-agent/agent_test.go
@@ -154,6 +154,7 @@ func TestAgent(t *testing.T) {
                        a.Security.CertChainFilePath = cfg.CertificatePath
                        a.Security.KeyFilePath = cfg.PrivateKeyPath
                        a.Security.RootCertFilePath = cfg.CaCertificatePath
+                       a.Security.JWTPath = "bogus"
                        a.ProxyConfig.ProxyMetadata = map[string]string{}
                        a.ProxyConfig.ProxyMetadata[MetadataClientCertChain] = filepath.Join(dir, "cert-chain.pem")
                        a.ProxyConfig.ProxyMetadata[MetadataClientCertKey] = filepath.Join(dir, "key.pem")
```

Since right now the test has a valid JWT token so its not ideal test.

lei-tang: A new test case with (a.Security.JWTPath = "bogus") has been added.

PR #35070, opened 2021-09-08
Drop --bootstrap-version flag

sergii-ssh:
Drop --bootstrap-version flag for envoy binary. Flag has been deprecated https://github.com/envoyproxy/envoy/pull/17724

- [ ] Configuration Infrastructure
- [ ] Docs
- [ x ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [ x ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #35389, opened 2021-09-27
[Cherry-pick] Drop --bootstrap-version flag (#35070)

sergii-ssh:
Drop --bootstrap-version flag for envoy binary. Flag has been deprecated envoyproxy/envoy#17724

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/envoy/proxy.go
+++ b/pkg/envoy/proxy.go
@@ -125,7 +125,6 @@ func (e *envoy) args(fname string, epoch int, bootstrapConfig string) []string {
 		"--drain-strategy", "immediate", // Clients are notified as soon as the drain process starts.
 		"--parent-shutdown-time-s", fmt.Sprint(int(convertDuration(e.ParentShutdownDuration) / time.Second)),
 		"--local-address-ip-version", proxyLocalAddressType,
-		"--bootstrap-version", "3",
 		// Reduce default flush interval from 10s to 1s. The access log buffer size is 64k and each log is ~256 bytes
 		// This means access logs will be written once we have ~250 requests, or ever 1s, which ever comes first.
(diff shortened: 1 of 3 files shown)


jwendell:
Why do we need to backport this to 1.11? We won't update envoy version anyway...

sergii-ssh:
hi John, we consume it downstream and it very much desirable to update it in istio. Also for any future CVE patching it might come handy. It seems to be no-op change for istio, appreciate if you could approve it.

ramaraochavali:
> we consume it downstream and it very much desirable to update it in istio

Can you please explain more about this?

I do not CVE patches will be on a newer version. We will get patches on the same version. So I kind of agree with @jwendell that this seems to be not needed in 1.11

shuoyang2016:
> > we consume it downstream and it very much desirable to update it in istio
> 
> Can you please explain more about this?
> 
> I do not CVE patches will be on a newer version. We will get patches on the same version. So I kind of agree with @jwendell that this seems to be not needed in 1.11

Downstream we have a custom gateway which uses newer version of envoy. And it's not released on 1.11 yet.

jwendell:
Deprecated doesn't mean removed. You may get a warning in envoy logs, but it should not be harmful.

We don't commit to supporting a newer version of Envoy than what's being used in a release.

sergii-ssh:
Sorry, should've used more precise wording.  Envoy actually removed the flag - https://github.com/envoyproxy/envoy/pull/17724
That's ok, we'll workaround downstream. Thanks

PR #33447, opened 2021-06-15
Remove istio.io/rev from injection template

Monkeyanator:
Fixes #33237 by not overwriting `istio.io/rev` on injection. Adds revision to the `sidecar.istio.io/status` annotation.

[ ] Configuration Infrastructure
[ ] Docs
[x] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #35390, opened 2021-09-27
Add istio.io/injectedBy injection label

Monkeyanator:
Since https://github.com/istio/istio/pull/33447 we haven't had a good way to identify pods injected by a given revision.

That PR removed the `istio.io/rev` label since it didn't play nice with revision tags (`istio.io/rev` label would get overwritten on injection) but there's no reason we can't add another, non-overlapping label (`istio.io/injectedBy`) to serve the same purpose. This is extra-useful when using revision tags since there's an extra layer of indirection between what `istio.io/rev` label you have and what revision actually does the injection.

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn (commented):
I am not 100% convinced this warrents a new label. These have a runtime and cognitive cost.

It seems the benefit here is for client-side querying, for debugging etc. This is not something that I think needs to be performant. 

is this something that could instead just be a istioctl command that does the filtering on the client side?

I don't feel very strongly about this but leaning towards that

Monkeyanator:
Agreed a label adds some cognitive load but I'm not sure it's a bad piece of information to have visible in this case. In terms of performance implications, I know it's an index in etcd but having a hard time finding more details on how detrimental a single pod label is really.

We could certainly add an istioctl command to do the same, but IMO doing this with k8s tooling is preferable, and we'd also be adding some bloat/cognitive load/maintenance burden to istioctl if we put it there.

I also don't feel too strongly one way or the other though. Since it seems partially like a UX discussion, anybody on @istio/wg-user-experience-maintainers have any thoughts here?

PR #35392, opened 2021-09-28
webhook patcher: replace istio queue with k8s queue

hzxuzhonghu:

To prevent out of order patch ops cause config outdated

### DIFF
--- a/pkg/webhooks/webhookpatch.go
+++ b/pkg/webhooks/webhookpatch.go
@@ -22,17 +22,19 @@ import (
 	"strings"
 	"time"
 
-	"github.com/cenkalti/backoff"
 	v1 "k8s.io/api/admissionregistration/v1"
 	kubeErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	admissioninformer "k8s.io/client-go/informers/admissionregistration/v1"
 	"k8s.io/client-go/kubernetes"
 	admissionregistrationv1client "k8s.io/client-go/kubernetes/typed/admissionregistration/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 
 	"istio.io/api/label"
 	"istio.io/istio/pilot/pkg/keycertbundle"
-	"istio.io/istio/pkg/queue"
+	kubelib "istio.io/istio/pkg/kube"
 	"istio.io/istio/pkg/webhooks/util"
 	"istio.io/pkg/log"
 )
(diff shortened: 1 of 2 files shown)


pkg/webhooks/webhookpatch.go
```diff
@@ -50,64 +52,79 @@ type WebhookCertPatcher struct {
...
+}
+
+// Run runs the WebhookCertPatcher
+func (w *WebhookCertPatcher) Run(stopChan <-chan struct{}) {
+	go w.runWebhookController(stopChan)
+	go w.startCaBundleWatcher(stopChan)
+	go wait.Until(w.worker, time.Second, stopChan)
+}
+
+// Run runs the WebhookCertPatcher
```
ramaraochavali, line 101: nit: incorrect comment

pkg/webhooks/webhookpatch.go
```diff
@@ -50,64 +52,79 @@ type WebhookCertPatcher struct {
...
+func (w *WebhookCertPatcher) worker() {
+	for w.processNextItem() {
+	}
+}
+
+// Run runs the WebhookCertPatcher
+func (w *WebhookCertPatcher) processNextItem() bool {
```
ramaraochavali, line 108: nit: same. incorrect comment

ramaraochavali:
LGTM. Couple of nits

pkg/webhooks/webhookpatch.go
```diff
@@ -50,64 +52,79 @@ type WebhookCertPatcher struct {
...
+	}
+	informer := admissioninformer.NewFilteredMutatingWebhookConfigurationInformer(client, 0, cache.Indexers{}, func(options *metav1.ListOptions) {
+		options.LabelSelector = fmt.Sprintf("%s=%s", label.IoIstioRev.Name, revision)
+	})
+	p.informer = informer
+	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			oldConfig := oldObj.(*v1.MutatingWebhookConfiguration)
+			newConfig := newObj.(*v1.MutatingWebhookConfiguration)
+			p.updateWebhookHandler(oldConfig, newConfig)
```
howardjohn, line 83: nit: I think its slightly more idiomatic k8s controller to just get the key (name) of newObj here and enqueue it. The queue will handle the things in updateWebhookHandler (skipping if no update is needed).

I don't see any issues with this - should work - just nitpicks

howardjohn (approved):

pkg/webhooks/webhookpatch.go
```diff
@@ -50,64 +52,79 @@ type WebhookCertPatcher struct {
...
+	})
+
+	return p, nil
+}
+
+// Run runs the WebhookCertPatcher
+func (w *WebhookCertPatcher) Run(stopChan <-chan struct{}) {
+	go w.runWebhookController(stopChan)
+	go w.startCaBundleWatcher(stopChan)
+	go wait.Until(w.worker, time.Second, stopChan)
```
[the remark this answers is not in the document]
ramaraochavali: +1. I had the same doubt but looked at other controllers and they were using the same logic - so thought it is pattern that was agreed up on. But would be good to know.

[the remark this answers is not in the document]
hzxuzhonghu: If you look into the details of wait.Until. you will find the it will recover crash from `worker`. So it is safe to always run any worker.

PR #35396, opened 2021-09-28
Skip stackdriver vm test in istiodremote topology

frankbu:


bianpengyuan (commented):
Can we add a bit more context in the description about why it does not work?

frankbu:
> Can we add a bit more context in the description about why it does not work?

@bianpengyuan I think the VM tests could be made to work, but currently the way the echo services are set up, it doesn't work quite right when there are multiple clusters. Rather than fixing the setup, we decided to not run any of the VM pilot or security tests in MC or external control plane. This test is already disabled for MC (`RequireSingleCluster`), but the extra check `RequireLocalControlPlane` is also needed to make sure there's also no external control plane cluster in the test topology.

bianpengyuan (approved):

PR #35397, opened 2021-09-28
Generate MCS service from merged Service

nmittler:
Currently, the MCS `serviceImportCache` requires that the kube `Service` exist in each cluster containing a `ServiceImport` resource. This is an overly strict requirement and goes against the spirit of the MCS spec.

This PR modifies the `serviceImportCache` to use the merged, mesh-wide `Service` from the `aggregate.Controller` instead. This means that as long as the service exists somewhere in the mesh, it can be imported in any cluster and can generate the synthetic MCS service for the `clusterset.local` host.

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
+++ b/pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
@@ -43,11 +43,20 @@ type importedService struct {
 	clusterSetVIP  string
 }
 
-// serviceImportCache reads Kubernetes Multi-Cluster Services (MCS) ServiceImport resources in the
-// cluster and generates a synthetic service for the MCS host (i.e. clusterset.local) that contains
-// the ClusterSet VIP. The synthetic service is a copy of the regular kube Service (e.g. cluster.local)
-// with the same namespaced name, but with the hostname and VIPs changed to the appropriate values for
-// the ClusterSet.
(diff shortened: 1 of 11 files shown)


pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
```diff
@@ -83,30 +92,39 @@ type serviceImportCacheImpl struct {
...
 
-	if exists {
+	if prevMcsService != nil {
 		event = model.EventUpdate
 	} else {
 		event = model.EventAdd
 	}
 
+	mcsService := ic.genMCSService(svc, mcsHost, vips)
 	ic.addOrUpdateService(nil, mcsService, event)
```
howardjohn, line 128: I think there is a bug here, or I am misunderstanding the flow:

* Cluster 1 creates Service `foo` with port 80. We trigger this function and store the `model.Service` with port 80
* Cluster 2 creates Service `foo` with port 8080. We trigger this function and _overwrite_ the `model.Service` with port 8080

nmittler: So this is actually something I discussed with @stevenctl yesterday.

Service merging assumes sameness across clusters. If a service is configured differently across clusters, I suspect multi-cluster is just generally broken.

nmittler:
@howardjohn @stevenctl mind taking a look?

howardjohn (approved):

PR #35409, opened 2021-09-29
replace deprecated HeaderMatchSpecifier with HeaderMatcher_StringMatch

pangsq:

`HeaderMatcher_ExactMatch`/`HeaderMatcher_SafeRegexMatch`/`HeaderMatcher_PrefixMatch`/`HeaderMatcher_SuffixMatch` are deprecated, which can be replaced with `HeaderMatcher_StringMatch`.

### DIFF
--- a/pilot/pkg/security/authz/matcher/header.go
+++ b/pilot/pkg/security/authz/matcher/header.go
@@ -36,22 +36,34 @@ func HeaderMatcher(k, v string) *routepb.HeaderMatcher {
 	} else if strings.HasPrefix(v, "*") {
 		return &routepb.HeaderMatcher{
 			Name: k,
-			HeaderMatchSpecifier: &routepb.HeaderMatcher_SuffixMatch{
-				SuffixMatch: v[1:],
+			HeaderMatchSpecifier: &routepb.HeaderMatcher_StringMatch{
+				StringMatch: &matcherpb.StringMatcher{
+					MatchPattern: &matcherpb.StringMatcher_Suffix{
+						Suffix: v[1:],
+					},
+				},
 			},
 		}
 	} else if strings.HasSuffix(v, "*") {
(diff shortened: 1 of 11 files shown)


ramaraochavali (approved):

howardjohn:
what envoy version was this added in?

pilot/pkg/security/authz/matcher/header.go
```diff
@@ -36,22 +36,34 @@ func HeaderMatcher(k, v string) *routepb.HeaderMatcher {
 	} else if strings.HasPrefix(v, "*") {
 		return &routepb.HeaderMatcher{
 			Name: k,
-			HeaderMatchSpecifier: &routepb.HeaderMatcher_SuffixMatch{
-				SuffixMatch: v[1:],
+			HeaderMatchSpecifier: &routepb.HeaderMatcher_StringMatch{
+				StringMatch: &matcherpb.StringMatcher{
```
yangminzhu, line 40: The `StringMatcher` is added in Envoy recently (https://github.com/envoyproxy/envoy/pull/17119, the upcoming 1.20.0) and not available in older versions.

I think we should wait a little longer before using it (or guard it with version check) otherwise it will cause problems for older Envoy for in-place upgrade.

yangminzhu (changes requested):

pilot/pkg/security/authz/matcher/header.go
pangsq: you are right, it may be too early to update the `HeaderMatchSpecifier` to use `StringMatch` and that may cause compatibility issues with old version envoy

ramaraochavali:
@yangminzhu Thanks for catching. I some how thought that deprecation happened a while ago. Sorry for that.

Issue #33071, opened 2021-05-25
vs match queryParams prefix not work,any can 


13567436138:
(**NOTE**: This is used to report product bugs:
  To report a security vulnerability, please visit <https://istio.io/about/security-vulnerabilities>
  To ask questions about how to use Istio, please visit <https://discuss.istio.io>)

**Bug description**

[ ] Docs
[ ] Installation
[ x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

**Steps to reproduce the bug**
```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: bookinfo
spec:
  gateways:
  - bookinfo-gateway
  hosts:
  - '*'
  http:
  - match:
    - queryParams:
        test:
          prefix: test
    route:
    - destination:
        host: productpage
        port:
          number: 9080
```
**Version** (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version --short` if you used Helm)
[root@master01 match]# istioctl version
client version: 1.10.0
control plane version: 1.10.0
data plane version: 1.10.0 (11 proxies)
**How was Istio installed?**

**Environment where the bug was observed (cloud vendor, OS, etc)**

Additionally, please consider running `istioctl bug-report` and attach the generated cluster-state tarball to this issue.
Refer [cluster state archive](http://istio.io/help/bugs/#generating-a-cluster-state-archive) for more details.

13567436138:
why nobody  look at this

PR #35415, opened 2021-09-29
QueryParam support prefixMatch

pangsq:

Fix https://github.com/istio/istio/issues/33071. 
Though [istio/api](https://github.com/istio/api/blob/1.12.0-alpha.1/networking/v1alpha3/virtual_service.pb.go#L1581) said `prefix` is not supported, but actually envoy can support prefixMatch on QueryParam.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/route/route.go
+++ b/pilot/pkg/networking/core/v1alpha3/route/route.go
@@ -857,6 +857,14 @@ func translateQueryParamMatch(name string, in *networking.StringMatch) *route.Qu
 				},
 			},
 		}
+	case *networking.StringMatch_Prefix:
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
should we add a test to https://github.com/istio/istio/blob/af583e94e4e5e8783b66276e7c99e8d743aa115a/tests/integration/pilot/common/routing.go#L128?

pangsq:
> should we add a test to
> 
> https://github.com/istio/istio/blob/af583e94e4e5e8783b66276e7c99e8d743aa115a/tests/integration/pilot/common/routing.go#L128
> 
> ?

ok, I will take a look into that. 
:( for now, I am not familiar with the Integration Tests

PR #35417, opened 2021-09-29
proxyless grpc: support fault injection

stevenctl:
Change our response to include the filters required for https://github.com/grpc/proposal/blob/master/A33-Fault-Injection.md

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/route/route.go
+++ b/pilot/pkg/networking/core/v1alpha3/route/route.go
@@ -475,7 +475,7 @@ func translateRoute(node *model.Proxy, in *networking.HTTPRoute,
 			d = features.DefaultRequestTimeout
 		}
 		action.Timeout = d
-		if node.Metadata.Generator == "grpc" {
+		if node.IsProxylessGrpc() {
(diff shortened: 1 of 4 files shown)


stevenctl:
cc @ramaraochavali wrt the MaxStreamDuration

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -475,9 +475,15 @@ func translateRoute(node *model.Proxy, in *networking.HTTPRoute,
 			d = features.DefaultRequestTimeout
 		}
 		action.Timeout = d
-		// Use deprecated value for now as the replacement MaxStreamDuration has some regressions.
-		// nolint: staticcheck
-		action.MaxGrpcTimeout = d
+		if node.Metadata.Generator == "grpc" {
```
howardjohn, line 478: nit: If we are going to add these checks I feel like we should have some function abstracting this - how we identify its for grpc may change in the future, etc. Seems more clear to have `node.IsGrpc()` or similar

howardjohn (approved):

PR #35426, opened 2021-09-30
Bumping up timeouts in serviceimportcache_test

nmittler:
Attempt to address a flake seen in a [PR](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/35416/unit-tests_istio/1443265686993375232). Does not seem to be reproducible locally.


howardjohn:
Still seeing this, i was able to reproduce before and after this PR following https://github.com/istio/istio/wiki/Test-Flakes#reproducing-test-flakes

howardjohn:
> failed processing add event for ServiceImport test-ns/test-svc in cluster test-cluster. No matching service found in cluster

I am guessing it doesn't like updates coming in the wrong order

nmittler:
Closing this since the issue appears to be in the fake k8s client.

Issue #35436, opened 2021-10-01
SAN is not checked unless supplied in a DR


Kmoneal:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**
For `SIMPLE` and `MUTUAL` TLS mode, the host name is never checked in the presented certificate unless SAN is specified. SAN should be automatically filled to properly check the host certificate.

**Describe alternatives you've considered**
SAN being required by default.

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[X] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**
Below are SEs, DRs, and VSs that will illustrate the problem. `wrong.host.badssl.com` communication should be rejected because the certificate is issued to `*.badssl.com` rather than `*.host.badssl.com`. If no SAN is presented, even with a CA certificate specified, the TLS connection is accepted. 

`wrong.host.badssl.com` communication accepted. This is insecure behavior.
```
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
  name: wrong-host-badssl-service-entry
  namespace: demo
spec:
  hosts:
  - wrong.host.badssl.com
  location: MESH_EXTERNAL
  ports:
  - number: 443
    name: https
    protocol: TLS
  resolution: DNS
---
apiVersion: networking.istio.io/v1beta1
kind: DestinationRule
metadata:
  name: wrong-host-badssl-destination-rule
  namespace: demo
spec:
  host: wrong.host.badssl.com
  trafficPolicy:
    tls:
      caCertificates: /etc/ssl/certs/ca-certificates.crt
      mode: SIMPLE
      sni: wrong.host.badssl.com
---
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: wrong-host-badssl-virtual-service
  namespace: demo
spec:
  hosts:
  - wrong.host.badssl.com
  http:
  - route:
    - destination:
        host: wrong.host.badssl.com
        port:
          number: 443
```

`wrong.host.badssl.com` communication refused.
```
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
[...]

PR #35437, opened 2021-10-01
Verify cert at client san

Kmoneal:
Introduce SAN checking by default. This is possibly breaking behavior so I propose using `VerifyCertAtClient` to introduce this behavior.

If no SAN field is specified in the ServiceEntry or DestinationRule, the host name will be used as a SAN. This should handle certificates more appropriately.

This addresses #35436

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -241,7 +241,7 @@ func (cb *ClusterBuilder) applyDestinationRule(mc *MutableCluster, clusterMode C
 
 	host := ""
 	if destinationRule != nil {
-		host = destinationRule.Host
+		host = service.Hostname.String()
 	}
 	// merge applicable port level traffic policy settings
 	trafficPolicy := MergeTrafficPolicy(nil, destinationRule.GetTrafficPolicy(), port)
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -238,6 +238,11 @@ func (cb *ClusterBuilder) buildSubsetCluster(opts buildClusterOpts, destRule *co
 func (cb *ClusterBuilder) applyDestinationRule(mc *MutableCluster, clusterMode ClusterMode, service *model.Service,
 	port *model.Port, proxyNetworkView map[network.ID]bool, destRule *config.Config, serviceAccounts []string) []*cluster.Cluster {
 	destinationRule := CastDestinationRule(destRule)
+
+	host := ""
+	if destinationRule != nil {
+		host = destinationRule.Host
+	}
 	// merge applicable port level traffic policy settings
 	trafficPolicy := MergeTrafficPolicy(nil, destinationRule.GetTrafficPolicy(), port)
 	opts := buildClusterOpts{
@@ -249,6 +254,7 @@ func (cb *ClusterBuilder) applyDestinationRule(mc *MutableCluster, clusterMode C
 		clusterMode:      clusterMode,
 		direction:        model.TrafficDirectionOutbound,
 		cache:            cb.cache,
+		host:             host,
 	}
 
 	if clusterMode == DefaultClusterMode {
@@ -1068,9 +1074,16 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 			if !res.IsRootCertificate() {
 				tlsContext.CommonTlsContext.ValidationContextType = &auth.CommonTlsContext_ValidationContext{}
 			} else {
+				subjectAltNames := make([]string, 0)
+				if tls.SubjectAltNames != nil && len(tls.SubjectAltNames) != 0 {
+					subjectAltNames = tls.SubjectAltNames
+				} else if features.VerifyCertAtClient && opts != nil && opts.host != "" {
+					subjectAltNames = append(subjectAltNames, opts.host)
+					tls.SubjectAltNames = subjectAltNames
+				}
 				tlsContext.CommonTlsContext.ValidationContextType = &auth.CommonTlsContext_CombinedValidationContext{
 					CombinedValidationContext: &auth.CommonTlsContext_CombinedCertificateValidationContext{
-						DefaultValidationContext:         &auth.CertificateValidationContext{MatchSubjectAltNames: util.StringToExactMatch(tls.SubjectAltNames)},
+						DefaultValidationContext:         &auth.CertificateValidationContext{MatchSubjectAltNames: util.StringToExactMatch(subjectAltNames)},
 						ValidationContextSdsSecretConfig: authn_model.ConstructSdsSecretConfig(res.GetRootResourceName()),
 					},
 				}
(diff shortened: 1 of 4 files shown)


pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2967,3 +2967,273 @@ func TestApplyDestinationRuleOSCACert(t *testing.T) {
...
+						SubjectAltNames: []string{"*.default.svc.cluster.local"},
+						Mode:            networking.ClientTLSSettings_SIMPLE,
+					},
+				},
+			},
+			expectedSubjectAltNames:  []string{"*.default.svc.cluster.local"},
+			enableVerifyCertAtClient: false,
+		},
+		{
+			name:            "VerifyCertAtClient false with SAN and MUTUAL TLS mode",
```
Kmoneal, line 3038: VerifyCertAtClient false and no setting SAN

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2967,3 +2967,273 @@ func TestApplyDestinationRuleOSCACert(t *testing.T) {
...
+						ClientCertificate: "some-value",
+						PrivateKey:        "some-value",
+					},
+				},
+			},
+			expectedSubjectAltNames:  []string{"*.default.svc.cluster.local"},
+			enableVerifyCertAtClient: false,
+		},
+		{
+			name:            "VerifyCertAtClient set with SAN and SIMPLE TLS mode",
```
Kmoneal, line 3067: Not overwriting

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2967,3 +2967,273 @@ func TestApplyDestinationRuleOSCACert(t *testing.T) {
...
+						SubjectAltNames: []string{"*.default.svc.cluster.local"},
+						Mode:            networking.ClientTLSSettings_SIMPLE,
+					},
+				},
+			},
+			expectedSubjectAltNames:  []string{"*.default.svc.cluster.local"},
+			enableVerifyCertAtClient: true,
+		},
+		{
+			name:            "VerifyCertAtClient set with SAN and MUTUAL TLS mode",
```
Kmoneal, line 3094: Update name to explain SAN is in SE not DR

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2967,3 +2967,273 @@ func TestApplyDestinationRuleOSCACert(t *testing.T) {
...
+						ClientCertificate: "some-value",
+						PrivateKey:        "some-value",
+					},
+				},
+			},
+			expectedSubjectAltNames:  []string{"*.default.svc.cluster.local"},
+			enableVerifyCertAtClient: true,
+		},
+		{
+			name:            "VerifyCertAtClient set without SAN in SIMPLE TLS mode",
```
Kmoneal, line 3121: give SE, DR, VerifyCertAtClient, and expected behavior

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2967,3 +2967,273 @@ func TestApplyDestinationRuleOSCACert(t *testing.T) {
...
+		port                     *model.Port
+		networkView              map[network.ID]bool
+		serviceAccounts          []string
+		destRule                 *networking.DestinationRule
+		expectedSubjectAltNames  []string
+		enableVerifyCertAtClient bool
+	}{
+
+		{
+			name:            "VerifyCertAtClient false with SAN and SIMPLE TLS mode",
```
Kmoneal, line 3011: Add test with SE and DR in conflict

howardjohn:
@hzxuzhonghu any concern now?

Kmoneal:
@hzxuzhonghu, hopefully you don't see this until Monday, but any addition feedback would be appreciated after the switch from destiantionRule.Host to service.Hostname.

Kmoneal:
> Will an user guide be added to istio.io for this feature? It will be helpful to provide some examples of using/verifying the feature in the user guide.

Yes, I have been updating [`Security Best Practices`](https://preliminary.istio.io/latest/docs/ops/best-practices/security/#configure-tls-verification-in-destination-rule-when-using-tls-origination) but would be happy to add this wherever it is also appropriate

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -239,6 +239,11 @@ func (cb *ClusterBuilder) buildSubsetCluster(opts buildClusterOpts, destRule *co
 func (cb *ClusterBuilder) applyDestinationRule(mc *MutableCluster, clusterMode ClusterMode, service *model.Service,
 	port *model.Port, proxyNetworkView map[network.ID]bool, destRule *config.Config, serviceAccounts []string) []*cluster.Cluster {
 	destinationRule := CastDestinationRule(destRule)
+
+	host := ""
+	if destinationRule != nil {
+		host = service.Hostname.String()
```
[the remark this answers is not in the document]
howardjohn: I think if user had this config today (without the explicit SAN/caCerts), they would be broken for `*.google.com` since its exact match though?

Exact match on `*.x.y` doesn't make sense at all. IMO we can either not automatically set the SAN check for wildcard (as short term action), and long term allow prefix. 

We could go directly to long term, just don't want to scope creep this more...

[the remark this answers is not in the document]
Kmoneal: That's right. I thought you meant applying an SE with a wildcard is broken (cannot be applied). I didn't catch the example above did not have `resolution: DNS`. 
I think we will go with not setting SAN when wildcard is set and work toward auto_san/auto_sni for long term

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -238,6 +238,8 @@ func (cb *ClusterBuilder) buildSubsetCluster(opts buildClusterOpts, destRule *co
 func (cb *ClusterBuilder) applyDestinationRule(mc *MutableCluster, clusterMode ClusterMode, service *model.Service,
 	port *model.Port, proxyNetworkView map[network.ID]bool, destRule *config.Config, serviceAccounts []string) []*cluster.Cluster {
 	destinationRule := CastDestinationRule(destRule)
+
+	host := service.Hostname.String()
```
hzxuzhonghu, line 242: I think this is not right for dns resolution, think about ServiceEntry with DNS type.

Kmoneal: I added a logging statement and applied a ServiceEntry with `resolution: DNS` and it was printing out the expected hostnames individually (for multiple hosts this still works). Is there anything I am missing? If this is not right for DNS resolution, do you have any helpful resources I could use to choose the hostname?
```
kind: ServiceEntry
metadata:
  name: badssl-service-entry
  namespace: default
spec:
  hosts:
  - badssl.com
  - self-signed.badssl.com
  location: MESH_EXTERNAL
  ports:
  - number: 443
    name: https
    protocol: TLS
  resolution: DNS
```

hzxuzhonghu: Forexample

···
// apiVersion: networking.istio.io/v1alpha3
// kind: ServiceEntry
// metadata:
//   name: external-svc-dns
// spec:
//   hosts:
//   - foo.bar.com
//   location: MESH_EXTERNAL
//   ports:
//   - number: 80
//     name: http
//     protocol: HTTP
//   resolution: DNS
//   endpoints:
//   - address: us.foo.bar.com
//     ports:
//       http: 8080
//   - address: uk.foo.bar.com
//     ports:
//       http: 9080
//   - address: in.foo.bar.com
//     ports:
//       http: 7080
···

PR #35447, opened 2021-10-01
Allow turning off ALPN in echo server

howardjohn:
Sad we have to resort to this but I don't see any other good options 

This is required for Go 1.17. They removed the ability to have unknown ALPN treated as HTTP 1 *and* h2 negotiation at the same time. For our tests, we rely on one or the other of these functionalities, so we can just make the test opt-in. Needing unknown ALPN is very rare (only when we are simulating Envoy mTLS via echo app) so its not a huge burden

### DIFF
--- a/pkg/test/echo/cmd/server/main.go
+++ b/pkg/test/echo/cmd/server/main.go
@@ -48,6 +48,7 @@ var (
 	crt              string
 	key              string
 	istioVersion     string
+	disableALPN      bool
 
 	loggingOptions = log.DefaultOptions()
 
@@ -123,6 +124,7 @@ var (
 				Cluster:               cluster,
 				IstioVersion:          istioVersion,
 				UDSServer:             uds,
+				DisableALPN:           disableALPN,
 			})
 
 			if err := s.Start(); err != nil {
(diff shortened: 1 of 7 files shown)


jacob-delgado:
While not ideal turning off alpn will have to be the solution for this test.

I had attempted to try a different hack where I modified https://github.com/istio/istio/blob/fe0baaf77ef044d6968aa898f660052071ddb88b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go#L1039 (we would *NOT* want this)

```golang
		// `istio-peer-exchange` alpn is only used when using mtls communication between peers.
		// We add `istio-peer-exchange` to the list of alpn strings.
		// The code has repeated snippets because We want to use predefined alpn strings for efficiency.
		if cb.IsHttp2Cluster(c) {
			// This is HTTP/2 in-mesh cluster, advertise it with ALPN.
			tlsContext.CommonTlsContext.AlpnProtocols = util.ALPNInMeshH2WithMxc
		} else {
			// This is in-mesh cluster, advertise it with ALPN.
			tlsContext.CommonTlsContext.AlpnProtocols = util.ALPNInMeshWithMxc
		}
```

However, I wasn't able to get this to work unless I did 

```golang
		...
		else {
			// This is in-mesh cluster, advertise it with ALPN.
			tlsContext.CommonTlsContext.AlpnProtocols = []string{"http/1.1"}
		}
```

as setting it to `[]string{"h2", "http/1.1"}` didn't work for me. I see you've reached out to the golang team. https://github.com/golang/go/issues/46310#issuecomment-932504562

I can't think of any way for us to test the behavior we want.

jacob-delgado (approved):

myidpt (approved):

Issue #27826, opened 2020-10-08
Ensure EnvoyFilter doesn't patch shared structs


howardjohn:
In Istio 1.4 I came across a bug where we mutated a shared struct from EnvoyFilter merge. It doesn't exist anymore, but we should ensure there are no similar bugs lurking in our code today.

cc @ramaraochavali @rshriram

howardjohn:
not stale

howardjohn:
Not stale, but hard to do without killing perf

howardjohn:
not stale

howardjohn:
not stale

howardjohn:
Notstale

vikaschoudhary16:
not stale

yann-soubeyrand:
Hi @howardjohn, do you know if this issue has been fixed or not?

howardjohn:
adhoc it has but not universally

PR #35449, opened 2021-10-02
do not build catch all virtual host every time

ramaraochavali:
CatchAllVirtualHost only changes when sidecar for a proxy is changed. We do not have to build it for every route every time.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	route "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
 	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	gogojsonpb "github.com/gogo/protobuf/jsonpb"
 	"github.com/golang/protobuf/jsonpb"
@@ -34,6 +35,7 @@ import (
 	structpb "github.com/golang/protobuf/ptypes/struct"
 
 	meshconfig "istio.io/api/mesh/v1alpha1"
+	networking "istio.io/api/networking/v1alpha3"
 	istionetworking "istio.io/istio/pilot/pkg/networking"
 	"istio.io/istio/pilot/pkg/trustbundle"
 	"istio.io/istio/pkg/cluster"
@@ -303,6 +305,8 @@ type Proxy struct {
 
 	// XdsNode is the xDS node identifier
 	XdsNode *core.Node
+
+	CatchAllVirtualHost *route.VirtualHost
 }
 
 // WatchedResource tracks an active DiscoveryRequest subscription.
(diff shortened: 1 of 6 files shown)


howardjohn:
Benchmark test shows about a 3% increase maybe. Is that what you expected or do you have setups where its even more impactful?

I am 100% fine with this but am slightly worried about this making https://github.com/istio/istio/issues/27826 a bit worse. We need to fix that regardless for sure though...

The other question I had was - where do we stop with this? A natural next step would be to cache passthrough cluster, etc. I guess since those are only 1 instance its not a big deal vs this being 1 per route config.
It would be slightly preferred if we just generate the 1 catch all in the RDS gen (instead of in push context), but maybe that would require way to much refactoring to make it viable? That way all the route generation happens in the "right" spot, which is *slightly* useful for code organization and runtime performance. For example, we are now pushing route generation out of the route generation code, so telemetry is slightly skewed for xds gen timings, etc. Not a big deal though.

ramaraochavali:
> Benchmark test shows about a 3% increase maybe. Is that what you expected or do you have setups where its even more impactful?

No. But I was hoping when we scale to more routes and proxies, it might help.

> I am 100% fine with this but am slightly worried about this making #27826 a bit worse. We need to fix that regardless for sure though...

Yes Agree. 

> The other question I had was - where do we stop with this? A natural next step would be to cache passthrough cluster, etc. I guess since those are only 1 instance its not a big deal vs this being 1 per route config.
It would be slightly preferred if we just generate the 1 catch all in the RDS gen (instead of in push context), but maybe that would require way to much refactoring to make it viable? That way all the route generation happens in the "right" spot, which is slightly useful for code organization and runtime performance. For example, we are now pushing route generation out of the route generation code, so telemetry is slightly skewed for xds gen timings, etc. Not a big deal thoug

I was going back and forth on this specific thing.  That would mean we have to change a lot.

ramaraochavali:
@howardjohn WDYT about this? Should we move forward or should we try to refactor it to rds? I can try that if you prefer that approach - but not sure how much refactoring it will be.

YonkaFang:
> > Benchmark test shows about a 3% increase maybe. Is that what you expected or do you have setups where its even more impactful?
> 
> No. But I was hoping when we scale to more routes and proxies, it might help.
> 
> > I am 100% fine with this but am slightly worried about this making #27826 a bit worse. We need to fix that regardless for sure though...
> 
> Yes Agree.
> 
> > The other question I had was - where do we stop with this? A natural next step would be to cache passthrough cluster, etc. I guess since those are only 1 instance its not a big deal vs this being 1 per route config.
> > It would be slightly preferred if we just generate the 1 catch all in the RDS gen (instead of in push context), but maybe that would require way to much refactoring to make it viable? That way all the route generation happens in the "right" spot, which is slightly useful for code organization and runtime performance. For example, we are now pushing route generation out of the route generation code, so telemetry is slightly skewed for xds gen timings, etc. Not a big deal thoug
> 
> I was going back and forth on this specific thing. That would mean we have to change a lot.

@howardjohn It looks like what u worried has happened according to #36357

PR #35461, opened 2021-10-04
Handle empty ratings collection when using Mongo

luksa:
If the mongo db was empty, the ratings service would throw the following error:

TypeError: Cannot read property 'rating' of undefined
at /opt/microservices/ratings.js:155:37


craigbox (approved):
Lgtm, thanks!

PR #35478, opened 2021-10-05
add 'Kubernetes' as valid cluster ID for skipping proxy instances

ramaraochavali:
The following sequence fails to build proxy services instances for old proxies
- Deployed Istio with `ISTIO_META_CLUSTER_ID` default (i.e. `Kubernetes`). All proxies now belong to `Kubernetes`.
- Add support for another cluster - now updated the `ISTIO_META_CLUSTER_ID` to point to the right cluster name
- Now old proxies which have `Kubernetes` does not get proxy service instances till they are restarted.

This could be a miss on our end to populate the cluster ID correctly even for single cluster to start with. - But since we are in this (may be others?) proposing this change. 

This PR tries to check if the proxy's cluster is `Kubernetes` (default value) and allows to get proxy service instances.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -1214,7 +1214,8 @@ func (c *Controller) onSystemNamespaceEvent(obj interface{}, ev model.Event) err
 // isControllerForProxy should be used for proxies assumed to be in the kube cluster for this controller. Workload Entries
 // may not necessarily pass this check, but we still want to allow kube services to select workload instances.
 func (c *Controller) isControllerForProxy(proxy *model.Proxy) bool {
-	return proxy.Metadata.ClusterID == "" || proxy.Metadata.ClusterID == c.Cluster()
+	return proxy.Metadata.ClusterID == "" || proxy.Metadata.ClusterID == cluster.ID(provider.Kubernetes) ||
+		proxy.Metadata.ClusterID == c.Cluster()
 }
 
 // getProxyServiceInstancesFromMetadata retrieves ServiceInstances using proxy Metadata rather than
--- a/pilot/pkg/serviceregistry/aggregate/controller.go
+++ b/pilot/pkg/serviceregistry/aggregate/controller.go
@@ -226,7 +226,7 @@ func skipSearchingRegistryForProxy(nodeClusterID cluster.ID, r serviceregistry.I
 		return false
 	}
 
-	return !r.Cluster().Equals(nodeClusterID)
+	return !(r.Cluster().Equals(nodeClusterID) || nodeClusterID == cluster.ID(provider.Kubernetes))
 }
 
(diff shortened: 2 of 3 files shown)


ramaraochavali:
@stevenctl Please see if this reasonable or any other better suggestions?

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -1214,7 +1214,8 @@ func (c *Controller) onSystemNamespaceEvent(obj interface{}, ev model.Event) err
 // isControllerForProxy should be used for proxies assumed to be in the kube cluster for this controller. Workload Entries
 // may not necessarily pass this check, but we still want to allow kube services to select workload instances.
 func (c *Controller) isControllerForProxy(proxy *model.Proxy) bool {
-	return proxy.Metadata.ClusterID == "" || proxy.Metadata.ClusterID == c.Cluster()
+	return proxy.Metadata.ClusterID == "" || proxy.Metadata.ClusterID == cluster.ID(provider.Kubernetes) ||
```
ramaraochavali, line 1217: I think it is same as empty ClusterID so should be fine.

ramaraochavali:
> In the description we say Add support for another cluster but do we really mean rename the cluster from the default name to something else

Yes. When adding another cluster - we rename the cluster name. But you are right it can happen even if we just rename it.

> This could cause some obscure bugs.. but maybe better than current state?

I think so.

> Regardless of what we do in this PR we should also heavily recommend setting a cluster name even for single cluster installs.

Huge +1 for this. We missed it and possibly many others.

stevenctl:
If we go this direction do we also need to update this:

https://github.com/istio/istio/blob/42b03d9860736ae4cc23c149e0c21e287377810d/pilot/pkg/model/context.go#L664


Also want @howardjohn and @nmittler opinions. My one concerns is how we deal with cases where a user actually wants to keep using the default name, and then they add ex. `Kubernetes2` for their second cluster.

ramaraochavali:
Yes. I agree it is an aliasing issue. How about this? Add a server org to
define aliases and when node connects switch to definitive name so that no
other code needs to change? I do not think we need an API for this and we
strongly encourage people to name correctly?

hzxuzhonghu:
> IIUC, the issue is that the mesh admin forgot to set the clusterID when installing Istio on a single-cluster mesh, so it defaulted to "Kubernetes". Later on, they decided to expand the mesh to include other clusters and therefore need to migrate the cluster to a proper clusterID

I guess this is not a valid use case,  they don't share one CA, how can they communicate cross clusters

ramaraochavali:
> Actually, I don't think this should be an issue though, because old proxies with the old cluster id value will also have the old discovery address
That is correct.

Issue #30632, opened 2021-02-03
istioctl install: no warning on applying multiple IOPs in a single file


elfinhe:
**Bug description**
istioctl install: no warning on applying multiple IOPs in a single file.
Only the first IOP is used, but customers are expecting all of them are merged. 

I'd suggest we give warning messages and stop the installation.

[ ] Docs
[x] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**

**Steps to reproduce the bug**

iop.yaml

```
---
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  meshConfig:
    enableTracing: false
---
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  meshConfig:
    enableTracing: true
```

Output: https://gist.github.com/elfinhe/bc2c4440c6be59e5089718313a20c027

`istioctl` still use `enableTracing: false` to install without warning.

**Environment where the bug was observed (cloud vendor, OS, etc)**
Linux

esnible:
@istio/wg-environments-maintainers To me it makes sense to for the tool to install them all if the revision tags are different.  For example, there could be a file that sets up Istio mid-canary with two control planes.

If the file contains multiple IOPs for the same `revision` should we 
- Issue warning and skip?
- Issue warning and update the control plane twice?
- Merge all IOPs for the revision before applying?
- Reject the entire file as likely bogus?

howardjohn:
I do not think we should install multiple. We already support `-f a.yaml -f b.yaml` which does *not* mean install all of them, but rather merge all of them. So to have multiple in a single file lead to multiple installations is wrong in my opinion.

If there are multiple revisions in `-f` or in a single file it should probably be rejected

morvencao:
The IOP api support `IstioOperatorList`, see: https://github.com/istio/istio/blob/2bef9460f448242bc49ade969037cd5b7205ac68/operator/pkg/apis/istio/v1alpha1/deepcopy.go#L63-L68
So we should installed multiple istio if theses IOPs is specified in the same `IstioOperatorList ` 
If multiple IOPs in the same file, but not in the same `IstioOperatorList` we should merged them.

elfinhe:
Some examples we've already supported:

```
istioctl install -f a.yaml -b.yaml
```

```
[...]

PR #35494, opened 2021-10-06
Adds warning for multiple documents in overlay file. #30632

clsacramento:

Adds a warning when the overlay file contain multiple documents.

https://github.com/istio/istio/issues/30632

### DIFF
--- a/operator/pkg/manifest/shared.go
+++ b/operator/pkg/manifest/shared.go
@@ -541,7 +541,7 @@ func overlaySetFlagValues(iopYAML string, setFlags []string) (string, string, er
 	if err != nil {
 		return "", "", err
 	}
-	manifests := strings.Split(iopYAML, "---")
+	manifests := strings.Split(iopYAML, "\n---")
 	if len(manifests) > 1 {
 		return string(out), "Multiple documents in one manifest file. First document assumed, the rest is ignored.", nil
 	}
--- a/operator/pkg/manifest/shared.go
+++ b/operator/pkg/manifest/shared.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/fatih/color"
 	"github.com/ghodss/yaml"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -168,9 +169,14 @@ func GenIOPFromProfile(profileOrPath, fileOverlayYAML string, setFlags []string,
 
 	// Combine file and --set overlays and translate any K8s settings in values to IOP format. Users should not set
 	// these but we have to support this path until it's deprecated.
-	overlayYAML, err := overlaySetFlagValues(fileOverlayYAML, setFlags)
+	overlayYAML, warning, err := overlaySetFlagValues(fileOverlayYAML, setFlags)
 	if err != nil {
 		return "", nil, err
+	} else if warning != "" {
+		warnMarker := color.New(color.FgYellow).Add(color.Italic).Sprint("WARNING:")
+		l.PrintErr(warnMarker + " " + warning)
+		l.PrintErr("\n* Overlay file received:\n" + fileOverlayYAML)
+		l.PrintErr("\n* Overlay file assumed:\n" + overlayYAML)
 	}
 	t := translate.NewReverseTranslator()
 	overlayYAML, err = t.TranslateK8SfromValueToIOP(overlayYAML)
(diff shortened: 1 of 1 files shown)


clsacramento:
With this PR, the warning looks like this on my test:
```
$ ./istioctl install -f operator/iop.yaml --dry-run
2021-10-06T14:16:41.883760Z	info	proto: tag has too few fields: "-"
2021-10-06T14:16:42.496124Z	info	installer	Applying Cluster specific settings: spec:
  values:
    global:
      jwtPolicy: third-party-jwt
    pilot:
      env:
        ENABLE_LEGACY_FSGROUP_INJECTION: false

WARNING: Multiple documents in one manifest file. First document assumed, the rest is ignored.
* Overlay file received:
---
kind: IstioOperator
spec:
  meshConfig:
    enableTracing: false
---
kind: IstioOperator
spec:
  meshConfig:
    enableTracing: true

* Overlay file assumed:
kind: IstioOperator
spec:
  meshConfig:
    enableTracing: false
2021-10-06T14:16:43.535524Z	info	processing	Runtime.run: Starting new session id:1
2021-10-06T14:16:43.535600Z	info	processing	session[1] "inactive" => "starting"
2021-10-06T14:16:43.535842Z	info	source	Beginning CRD Discovery, to figure out resources that are available...
2021-10-06T14:16:43.836318Z	info	processing	session.handleMeshEvent: Received initial mesh event, applying it: [Event](Added: istio/mesh/v1alpha1/MeshConfig/istio-system/meshconfig)
2021-10-06T14:16:43.836331Z	info	source	CRD Discovery complete, starting listening to resources...
```

operator/pkg/manifest/shared.go
```diff
@@ -523,20 +529,24 @@ func overlaySetFlagValues(iopYAML string, setFlags []string) (string, error) {
...
+			return "", "", err
 		}
 	}
 
 	out, err := yaml.Marshal(iop)
 	if err != nil {
-		return "", err
+		return "", "", err
+	}
+	manifests := strings.Split(iopYAML, "---")
```
mandarjog, line 544: we are looking for `\n---` .  a `---` in the middle of the yaml file is still fine.

operator/pkg/manifest/shared.go
```diff
@@ -523,20 +529,24 @@ func overlaySetFlagValues(iopYAML string, setFlags []string) (string, error) {
...
 	}
 
 	out, err := yaml.Marshal(iop)
 	if err != nil {
-		return "", err
+		return "", "", err
+	}
+	manifests := strings.Split(iopYAML, "\n---")
+	if len(manifests) > 1 {
+		return string(out), "Multiple documents in one manifest file. First document assumed, the rest is ignored.", nil
```
howardjohn, line 546: Its valid and common to have a leading or trailing `---` in the file, but only have a single IOP, which may lead to unexpected warnings here

clsacramento: I just tested it and after adding mjog@'s suggestion of checking `\n---`, I am not getting a warning with this file:
```
$ cat operator/singleiop.yaml
---
kind: IstioOperator
spec:
  meshConfig:
    enableTracing: false
```
So leading `---` does not warn. Are you sure you don't want to warn for a trailing `---`?
It looks like this:
```
WARNING: Multiple documents in one manifest file. First document assumed, the rest is ignored.
* Overlay file received:
kind: IstioOperator
spec:
  meshConfig:
    enableTracing: false
---

* Overlay file assumed:
kind: IstioOperator
spec:
  meshConfig:
    enableTracing: false
```    
It looks clear that it is safe to ignore the warning but at the same time it educates people that they are not supposed to add anything extra to the file. WDYT?
I can otherwise just check if the `---` is in the end if you prefer.

howardjohn: I am not sure I agree with "they are not supposed to add anything extra to the file". Its pretty common from what I have seen. Adding warnings that are not actually an issue can lead to noise + ignoring them

morvencao (commented):
thanks for this enhancement, what if `IstioOperatorList` type is used?

morvencao:
@clsacramento An example of `IstioOperatorList `:

```
apiVersion: v1
kind: List
items:
- apiVersion: install.istio.io/v1alpha1
  kind: IstioOperator
  metadata:
    name: foo
  spec:
    meshConfig:
      enableTracing: false
- apiVersion: install.istio.io/v1alpha1
  kind: IstioOperator
  metadata:
    name: bar
  spec:
    meshConfig:
      enableTracing: true
```

clsacramento:
For the above file I get following output:
```
Error: fetch Kubernetes config file: Unauthorized
```
This is the same as what I get when I checkout master (i.e: this PR is not changing this behavior).

Please note that for this PR I am only concerned with the use of multiple documents in one YAML. I would like to address any other parsing problems separately.

morvencao (commented):
thank you @clsacramento fpr the PR, but I saw @ostromart also raised a PR for the same issue, see: https://github.com/istio/istio/pull/35508

clsacramento:
oh thanks, I didn't notice Martin's PR. Ok I will close this one then. :)

Issue #35495, opened 2021-10-06
Official gateway chart cannot be used as a subchart


remi-gelinas:
### Bug Description

Of all the official Istio charts in the[ official repo](https://artifacthub.io/packages/helm/istio-official/gateway), `gateway` is the only one with an enforced values schema - because `global` is not an allowed property in the schema, but is one that is injected to subcharts. one cannot use `istio/gateway` as a subchart.

I am setting up Istio to deploy through ArgoCD, which requires me to use proxy Helm charts when generating `ApplicationSets`.

### Version

```prose
`istio/gateway 1.12.0-alpha.1`
```

QuestionAndAnswer:
Hey, does anyone know about workaround for now?

nc-ruth:
Please check out https://github.com/helm/helm/issues/10392 where there is a discussion

billiaz:
Added a pull request, with one of the solutions: 
https://github.com/istio/istio/pull/36744 might need changes and discussion

PR #35496, opened 2021-10-06
Allow global property in istio/gateway values schema

remi-gelinas:
Fix for #35495 - allows the official `istio/gateway` chart to be used as a subchart.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience

### DIFF
--- a/manifests/charts/gateway/values.schema.json
+++ b/manifests/charts/gateway/values.schema.json
@@ -3,6 +3,9 @@
   "type": "object",
   "additionalProperties": false,
   "properties": {
+    "global": {
+      "type": "object"
+    },
     "affinity": {
       "type": "object"
     },


howardjohn (commented):
Interesting.. this seems like a general flaw with subcharts and schema. While this seems reasonable I would like to look at what other charts do

zoonoo:
@howardjohn 
can you look into this issue?
It does seem to me that the lack of `global` attribute in the schema prevents us from using is as helm subchart.

Other charts within istio repository does not seem to declare the schema, but they do have `global` attributes within the default `values.yaml`.

Thanks.

howardjohn:
I am still not quite sure what setup triggers the issue. The proposed fix in the original issue seems like its not the right approach. I asked in the Helm slack and got no response, nor could I find any other helm charts doing this.

It would help if someone could add a reproducer on how to get intot his state?

zoonoo:
So the issue is reproduced when I create a chart that takes this gateway as a subchart. For example, my chart with the following spec would reproduce the issue : 

```
apiVersion: v2
name: istio-gateway
version: 0.0.1
appVersion: 1.12.0
description: "This Helm chart is an installation of istio-gateway, with additional components."
home: https://istio.io/latest/docs/setup/install/helm/

maintainers:
  - name: Park Junwoo
    email: junwoo@cloa.io

dependencies:
  - name: gateway
     version: 1.12.0
     repository: https://istio-release.storage.googleapis.com/charts
```

```
istio-gateway git:main ❯ helm template . 
Error: values don't meet the specifications of the schema(s) in the following chart(s):
gateway:
- (root): Additional property global is not allowed
```


The reason as I understand it is that [the Values data type has a reserved section called `Values.global`](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/), and this global section is populated when there is a dependency subchart.

As `values.schema.json` file in helm chart enforces Values data type to have the specified attributes only, helm lint fails.

```
istio-gateway git:main ❯ helm version
version.BuildInfo{Version:"v3.6.3", GitCommit:"d506314abfb5d21419df8c7e7e68012379db2354", GitTreeState:"dirty", GoVersion:"go1.16.6"}
```

howardjohn:
Opened https://github.com/helm/helm/issues/10392 to discuss

jortkoopmans:
Also encountered this issue today (unfortunately). Having read the discussion, my choice would be to use `"additionalProperties": true` or remove the line altogether (as it is true by default). Alternatively, this PR which excludes `Values.global` explicitly would also do the job.
@howardjohn ; this PR is going stale shortly, are you in a position to evaluate this?

howardjohn (approved):
I am not a huge fan of this but seems the best way to unblock folks for now

jacob-delgado:
cc @remi-gelinas Please add a release note similar to https://github.com/istio/istio/pull/36452/files to get this merged in. Thanks!

emwalker:
I just encountered this today, for what it's worth.  I had assumed I was doing something wrong or missing a step.  It seems the problem is in the chart itself.

howardjohn:
Merged in https://github.com/istio/istio/pull/36896 - thanks for getting this started @remi-gelinas !

PR #35510, opened 2021-10-07
Remove unused integration tests

frankbu:


frankbu:
@stevenctl I think the `istiodremote` ones are useless, but I'll put the multicluster ones back then. WDYT?

stevenctl:
This should actually be fine..

PR #35514, opened 2021-10-07
[Test Framework] Fix host header when Address specified

nmittler:
This was discovered as an issue while testing the Kubernetes Multi-Cluster Services (MCS) host (clusterset.local). When setting `Address`, it's expected that the host header would match by default.


stevenctl:
lgtm! maybe we should cleanup the HostHeader() on echo.Config? Feel free to remote DNM

PR #35528, opened 2021-10-09
Fix TestServiceDiscoveryServices

hzxuzhonghu:
**Please provide a description of this PR:**

Fix https://prow.istio.io/view/gs/istio-prow/logs/unit-tests_istio_postsubmit/1446377076654346240

PR #35542, opened 2021-10-11
Rework "refactor serviceentry store"

hzxuzhonghu:
I will add https://github.com/istio/istio/pull/35538 https://github.com/istio/istio/pull/35528 into this pr.


hzxuzhonghu:
@howardjohn kindly ping in case you missed it

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -165,72 +166,79 @@ func (s *ServiceEntryStore) workloadEntryHandler(old, curr config.Config, event
...
 	}
-
-	if event != model.EventDelete {
-		s.updateExistingInstances(key, instancesUpdated)
+	selected, unSelected := compareServiceEntries(oldSes, currSes)
+	log.Debugf("workloadEntry %v select serviceEntry, selected %v, unSelected %v", selected, unSelected)
+	s.mutex.Lock()
+	for _, namespacedName := range selected {
+		services := s.services.getServices(namespacedName)
+		cfg := s.store.Get(gvk.ServiceEntry, namespacedName.Name, namespacedName.Namespace)
```
howardjohn, line 212: I do not really understand this.

First, we list the SEs []config.Config. Then, we fetch all of the names/ns of the SEs that are relevant. Then, for each name/ns we look up the []config.Config... but we already had them before. Can't we just use the ones we already listed?

howardjohn: I guess we are inside the lock now, but seems like we could have just `List` inside the lock?

At the very least this could use some comments explaining why we do it, if there is a good reason.

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -165,72 +166,79 @@ func (s *ServiceEntryStore) workloadEntryHandler(old, curr config.Config, event
...
+			oldSes = currSes
+		} else {
+			oldSes = getWorkloadServiceEntries(cfgs, oldWle)
+		}
 	}
-
-	if event != model.EventDelete {
-		s.updateExistingInstances(key, instancesUpdated)
+	selected, unSelected := compareServiceEntries(oldSes, currSes)
+	log.Debugf("workloadEntry %v select serviceEntry, selected %v, unSelected %v", selected, unSelected)
```
howardjohn, line 208: this log is invalid, has more `%v` than parameters

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -320,46 +315,68 @@ func (s *ServiceEntryStore) serviceEntryHandler(old, curr config.Config, event m
...
-		currentServiceEntry := curr.Spec.(*networking.ServiceEntry)
-		oldServiceEntry := old.Spec.(*networking.ServiceEntry)
 		// If this service entry had endpoints with IPs (i.e. resolution STATIC), then we do EDS update.
 		// If the service entry had endpoints with FQDNs (i.e. resolution DNS), then we need to do
 		// full push (as fqdn endpoints go via strict_dns clusters in cds).
-		// Non DNS service entries are sent via EDS. So we should compare and update if such endpoints change.
+		// Non DNS service entries are sent via EDS.
+		// Trigger full push when DNS resolution ServiceEntry in case endpoint changes.
 		if currentServiceEntry.Resolution == networking.ServiceEntry_DNS || currentServiceEntry.Resolution == networking.ServiceEntry_DNS_ROUND_ROBIN {
-			if !reflect.DeepEqual(currentServiceEntry.Endpoints, oldServiceEntry.Endpoints) {
```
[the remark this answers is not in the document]
howardjohn: At the least the comment needs to change - it indicates the endpoints changed but we have no idea if that is true or not anymore

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -398,72 +392,81 @@ func (s *ServiceEntryStore) WorkloadInstanceHandler(wi *model.WorkloadInstance,
...
-		s.updateExistingInstances(key, instances)
+	if event == model.EventDelete {
+		s.serviceInstances.deleteInstances(key, instances)
 	} else {
-		s.deleteExistingInstances(key, instances)
+		s.serviceInstances.updateInstances(key, instances)
 	}
+	s.mutex.Unlock()
 
 	s.edsUpdate(instances, true)
```
howardjohn, line 471: Slight concern we need edsUpdate inside the lock. I will look at other usages

howardjohn: I think not. In edsUpdate we extract the keys. Then in edsUpdateByKeys, we fetch the instances in a lock and fetch the latest instances by key. However, we then release the lock and call EDSUpdate.

One concern might be multiple threads call edsUpdateByKeys:

1. Thread one fetches allInstances@v1
2. Thread two fetches allInstances@v2
3. Thread 2 calls EDSUpdate@v2
4. Thread 1 calls EDSUpdate@v1
Then we lose the v2 update.

WDYT?

This is an existing problem in the current code though

howardjohn (approved):
Thanks for your patience here, I finally got time to stay late and review this and it looks really good :+1: . Super happy to see the refreshIndex stuff gone!

I left a few comments, but overall they are all pretty minor. I think they can be addressed in followups, so I will approve and add a hold so you have a chance to address them. If you don't mind, can you include the fixes in a separate commit so its easier to review?

Thanks!!

hzxuzhonghu:
Thanks @howardjohn I will address the comments later

pilot/pkg/serviceregistry/serviceentry/store.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (s *serviceInstancesStore) deleteAllServiceEntryInstances(key types.NamespacedName) {
+	delete(s.instancesBySE, key)
+}
+
+// stores all the workload instances from pods
+type workloadInstancesStore struct {
+	// Stores a map of workload instance name/namespace to workload instance
+	instancesByKey map[string]*model.WorkloadInstance
```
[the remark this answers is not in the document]
hzxuzhonghu: sgtm

pilot/pkg/serviceregistry/serviceentry/store.go
```diff
@@ -0,0 +1,175 @@
...
+		delete(s.instancesBySE[key], cKey)
+	}
+}
+
+func (s *serviceInstancesStore) deleteAllServiceEntryInstances(key types.NamespacedName) {
+	delete(s.instancesBySE, key)
+}
+
+// stores all the workload instances from pods
+type workloadInstancesStore struct {
```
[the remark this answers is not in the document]
hzxuzhonghu: Yes, also wle

pilot/pkg/serviceregistry/serviceentry/store.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (w *workloadInstancesStore) update(wi *model.WorkloadInstance) {
+	if wi == nil {
+		return
+	}
+	key := keyFunc(wi.Namespace, wi.Name)
+	w.instancesByKey[key] = wi
+}
+
+// stores all the services and serviceEntries
```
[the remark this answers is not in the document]
hzxuzhonghu: right, should be services converted from se

pilot/pkg/serviceregistry/serviceentry/store.go
```diff
@@ -0,0 +1,175 @@
...
+	servicesBySE map[types.NamespacedName][]*model.Service
+}
+
+func (s *serviceStore) getAllServices() []*model.Service {
+	var out []*model.Service
+	for _, svcs := range s.servicesBySE {
+		out = append(out, svcs...)
+	}
+
+	return model.SortServicesByCreationTime(out)
```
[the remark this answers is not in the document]
hzxuzhonghu: We need to allocate the ip address in `func (s *ServiceEntryStore) Services() ([]*model.Service, error) ` after calling this method. So need an ordered list.

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -165,72 +166,79 @@ func (s *ServiceEntryStore) workloadEntryHandler(old, curr config.Config, event
...
-			}
+	wi := s.convertWorkloadEntryToWorkloadInstance(curr, s.Cluster())
+	if wi != nil {
+		// fire off the k8s handlers
+		for _, h := range s.workloadHandlers {
+			h(wi, event)
 		}
+	} else {
+		// in case workloadInstances store mem leak
+		event = model.EventDelete
```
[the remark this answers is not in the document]
hzxuzhonghu: yes, i thought about the case: user updates the address from an IP to uds. This is a very rare case, Maybe we should prevent this case in validation.

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -165,72 +166,79 @@ func (s *ServiceEntryStore) workloadEntryHandler(old, curr config.Config, event
...
-	if len(instancesDeleted) > 0 {
-		s.deleteExistingInstances(key, instancesDeleted)
+	cfgs, _ := s.store.List(gvk.ServiceEntry, curr.Namespace)
+	currSes := getWorkloadServiceEntries(cfgs, wle)
+	var oldSes map[types.NamespacedName]struct{}
+	if oldWle != nil {
+		if reflect.DeepEqual(oldWle.Labels, wle.Labels) {
+			oldSes = currSes
+		} else {
+			oldSes = getWorkloadServiceEntries(cfgs, oldWle)
```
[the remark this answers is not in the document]
hzxuzhonghu: If you mean the se selector updated simultaneously, it does not matter because we update `serviceInstances` as well in serviceentry handler.

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
hzxuzhonghu: Need an improvement, there is a race here, `Get` can return nil otherwise.

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
hzxuzhonghu: Yeah looks like a race. 
I guess previously it is the performance consideration that it is not protected.

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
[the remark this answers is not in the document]
hzxuzhonghu: done

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
hzxuzhonghu: Need to think about it carefully, maybe we can run one thread updating eds.

hzxuzhonghu: done

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -503,7 +505,17 @@ func (s *ServiceEntryStore) Services() ([]*model.Service, error) {
...
-	return autoAllocateIPs(allServices), nil
+
+	autoAllocateIPs(allServices)
+	out := make([]*model.Service, len(allServices))
+	for _, svc := range allServices {
+		// TODO: eliminate the shallowcopy here
+		// autoAllocateIPs will re-allocate ips for the service,
+		// if return the pointer directly, there will be a race with `BuildNameTable`
+		tmp := *svc
+		out = append(out, &tmp)
```
hzxuzhonghu, line 516: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/35542/unit-tests_istio/1458711022863912960

This is for the race test. I think we need to improve later to prevent allocate again for all the services

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -503,7 +505,16 @@ func (s *ServiceEntryStore) Services() ([]*model.Service, error) {
 	s.mutex.RLock()
 	allServices := s.services.getAllServices()
 	s.mutex.RUnlock()
-	return autoAllocateIPs(allServices), nil
+
+	out := make([]*model.Service, 0, len(allServices))
+	for _, svc := range allServices {
+		// TODO: eliminate the deepcopy here
+		// autoAllocateIPs will re-allocate ips for the service,
+		// if return the pointer directly, there will be a race with `BuildNameTable`
```
howardjohn, line 513: hmm.. this makes sense but is odd it didn't happen it current code?

pilot/pkg/serviceregistry/serviceentry/servicediscovery_test.go
```diff
@@ -672,7 +672,10 @@ func TestServiceDiscoveryWorkloadUpdate(t *testing.T) {
 		}
 		expectProxyInstances(t, sd, instances, "2.2.2.2")
 		expectServiceInstances(t, sd, selector, 0, instances)
-		expectEvents(t, events, Event{kind: "eds", host: "selector.com", namespace: selector.Namespace, endpoints: 2})
+		expectEvents(t, events,
+			Event{kind: "xds", proxyIP: "2.2.2.2"},
```
howardjohn, line 676: why does this change?

howardjohn (commented):
:+1:

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
hzxuzhonghu: previously, the services are all recreated with maybeRefreshIndex each time this function is called.

pilot/pkg/serviceregistry/serviceentry/servicediscovery_test.go
hzxuzhonghu:
```
		// trigger full xds push to the related sidecar proxy
		if event == model.EventAdd {
			s.XdsUpdater.ProxyUpdate(s.Cluster(), wle.Address)
		}
```

Previously, this event is ignored, if you run the test without this pr,  should see the ignore log. But now eds event is sent asynchrously, so i make it happen after proxy update event

howardjohn: :+1:

hzxuzhonghu:
@howardjohnCan u take the last look and remove the hold

pilot/pkg/xds/xds_test.go
```diff
@@ -543,11 +544,14 @@ spec:
 					fakeOpts.MeshConfig = &meshConfig
 					s := NewFakeDiscoveryServer(t, fakeOpts)
 					for clusterID := range want {
-						p := &model.Proxy{Metadata: &model.NodeMetadata{ClusterID: clusterID}}
-						eps := xdstest.ExtractLoadAssignments(s.Endpoints(p))[tt.serviceCluster]
-						if want := want[clusterID]; !listEqualUnordered(eps, want) {
-							t.Errorf("got %v but want %v for %s", eps, want, clusterID)
-						}
+						retry.UntilSuccess(func() error {
```
howardjohn, line 547: I am concerned by this change... does this mean we had some guarantees in the old code that are not present in the new code?

Biggest concern is this impacts prod users. Secondary impact is this impacts ALL our unit tests, and we are just fixing a single one.

howardjohn: Also need to use UntilSuccessOrFail otherwise this doesn't test anything.

howardjohn (changes requested):
LGTM besides one comment

pilot/pkg/xds/xds_test.go
hzxuzhonghu: In theory, original all tests should wait for some time. before got the expectations. Because the handlers processings are asych.  What makes the test failure i think is the async eds update. 

It does not influence users much.

howardjohn: If that was the case I would expect this test (and NDS, etc) to fail before this PR. As is, it passes 100% of the time without this PR and fails regularly with this PR. That leads me to believe the behavior has changed.

hzxuzhonghu: Chnaged

howardjohn: A bunch of other tests fail now with this PR, such as TestServiceScoping/STATIC. I suspect this code: https://github.com/istio/istio/blob/1c82a227a4e258b0db9bb2922484dbd277fd79ac/pilot/pkg/xds/fake.go#L315-L319 is now broken.

hzxuzhonghu: where did it fail?

howardjohn:
I ran it locally

hzxuzhonghu:
Not sure why the ci does not fail, i will run locally

pilot/pkg/xds/xds_test.go
hzxuzhonghu: >If that was the case I would expect this test (and NDS, etc) to fail before this PR

I figured it out, the tests ` NewConfigGenTest ` use the sync memory controller, `memory.NewSyncController`.

hzxuzhonghu:
Fixed

pilot/pkg/xds/xds_test.go
```diff
@@ -543,11 +544,14 @@ spec:
 					fakeOpts.MeshConfig = &meshConfig
 					s := NewFakeDiscoveryServer(t, fakeOpts)
 					for clusterID := range want {
-						p := &model.Proxy{Metadata: &model.NodeMetadata{ClusterID: clusterID}}
-						eps := xdstest.ExtractLoadAssignments(s.Endpoints(p))[tt.serviceCluster]
-						if want := want[clusterID]; !listEqualUnordered(eps, want) {
-							t.Errorf("got %v but want %v for %s", eps, want, clusterID)
-						}
+						retry.UntilSuccessOrFail(t, func() error {
```
howardjohn, line 547: We should revert his. Test should not need retry around XDS generation, it is intended to be handledby the NewFakeDiscoveryServer

howardjohn (changes requested):
I still see errors.

With this PR: `141 runs so far, 5 failures (96.58% pass rate). 6.33093786s avg, 9.572585688s max, 3.1743803s min`
Without this PR: `138 runs so far, 0 failures (100.00% pass rate). 6.470739882s avg, 8.844566221s max, 3.344019226s min`

On TestServiceScoping

hzxuzhonghu:
I guess we need to restore the sleep, otherwise we donot know when the resources are handled completely

howardjohn:
I am not okay with adding sleep to our tests. This will introduce flakes
and/or test slowness. The code in master branch handles it, I would not
want to regress from that.

hzxuzhonghu:
The master uses a totally different model `memory.NewSyncController` from production. Now we are asynchronous, we need some notification mechanism.

howardjohn:
why can't we use the sync controller here?

hzxuzhonghu:
Please take a look at 
```
func (s *ServiceEntryStore) edsUpdate(instances []*model.ServiceInstance, push bool) {
	// Find all keys we need to lookup
	keys := map[instancesKey]struct{}{}
	for _, i := range instances {
		keys[makeInstanceKey(i)] = struct{}{}
	}
	s.edsQueue.Push(func() error {
		s.edsUpdateByKeys(keys, push)
		return nil
	})
}
```

Now i send the eds update to a task queue, which is run async

howardjohn:
Maybe we need to make it synchronous then. We added the queue to enforce
ordering, right? we can have ordering and still he synchronous

hzxuzhonghu:
My concern is it will have a big influence on performance. This is a multi-producer one consumer model.

hzxuzhonghu:
I have an idea, how about making `ResyncEDS` keep running synchronously as before.  Since it is only for tests, i think it does not matter

hzxuzhonghu:
With the fix, no failure running 300 times on TestServiceScoping

howardjohn (approved):
Thanks for all your improvements you have made here!

hzxuzhonghu:
Thanks for your comments, which make it better

ramaraochavali:
Great work @hzxuzhonghu on this. So happy to see that `maybeRefreshIndexes` is gone now :-)

PR #35548, opened 2021-10-11
Another attempt at fixing DOCKER_BUILD_VARIANTS

howardjohn:
Our goal is to have the ability to build a -debug suffix from ubuntu, a
"" suffix from ubuntu, and -distroless suffix from distroless.

This has caused quite a few issues. I think this is the right fix.

Previously, we loop over DOCKER_BUILD_VARIANTS which is empty, so we do
nothing. Now, we have `default` in the loop, then just do some
processing later to ensure that we correctly rewrite it to `debug` in
the dockerfiles, and `""` in the tag.

I have tested `docker`, `dockerx`, and the release-builder
At this point I am convinced we should rewrite this in golang, but that is for another day...

### DIFF
--- a/tools/istio-docker.mk
+++ b/tools/istio-docker.mk
@@ -308,19 +308,22 @@ docker.distroless: docker/Dockerfile.distroless
 define variant-tag
 $(if $(filter-out default,$(1)),-$(1),)
 endef
+define normalize-tag
+$(subst default,debug,$(1))
+endef
 
 # DOCKER_BUILD_VARIANTS ?=debug distroless
 # Base images have two different forms:
 # * "debug", suffixed as -debug. This is a ubuntu based image with a bunch of debug tools
 # * "distroless", suffixed as -distroless. This is distroless image - no shell. proxyv2 uses a custom one with iptables added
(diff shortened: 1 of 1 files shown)


kfaseela:
I am building istio for first time.
my "make docker.push" is resulting in below error, wondering if this has anything to do with the current change, or I am missing some environment variable being set? "-default" is coming in tag, however the image built is not having this this suffix in the tag.

```sys	0m0.637s
time (set -e && for distro in default; do tag=depUpgrade-${distro}; docker push kfaseela/pilot:${tag%-debug}; done)
The push refers to repository [docker.io/kfaseela/pilot]
tag does not exist: kfaseela/pilot:depUpgrade-default

real	0m0.673s
user	0m0.028s
sys	0m0.034s
make[1]: *** [tools/istio-docker.mk:367: push.docker.pilot] Error 1
make: *** [docker.push] Error 2

kfaseela:
> Sorry about the breakage. will try to dfix asap. In the meantime, dockerx.pushx should work (and be faster)

Thanks @howardjohn

PR #35550, opened 2021-10-11
always generate grpc bootstrap

stevenctl:
Agent will generate a file if this var is set.

By always generating it, users can use the default injection with Envoy, and set `GRPC_XDS_BOOTSTRAP` on their app container + use `traffic.sidecar.istio.io/<in|ex>clude<In|Out>boundPorts` to allow their gRPC app to use proxyless features while still capturing TCP/HTTP traffic.


costinm (approved):
Is /var/lib/istio/data the right directory ? Shouldn't we instead use /etc/istio/envoy ( which is mounted and writeable - we write the other bootstrap ) ? 

/var/run/istio/bootstrap/ may be a better option, usually /var/run is tmpfs and writeable, and the name 'data' is far too broad for just one file.

Issue #35404, opened 2021-09-28
ServiceEntry has stale endpoints


yingzhuivy:
### Bug Description

We have a ServiceEntry in an external Istiod cluster that selects k8s pods in a remote cluster, when the pods are restarted, sometimes there will be stale endpoints in the cluster corresponding to the ServiceEntry. Looking at Istiod's `debug/endpointz`, all of the istiod pods have the same stale states, restarting the Istiod pods fix the problem.

To more easily reproduce, I just scale the deployment to 100 replicas, and then scale down to 1, sometimes some endpoints are not correctly deleted and there will be more than 1 endpoint under the cluster corresponding to the ServiceEntry. It do appear that it's easier to reproduce in more heavy loaded k8s clusters.

Here is the code for reproduction:
In the external Istiod cluster:
```---
apiVersion: v1
kind: Namespace
metadata:
  name: airmesh-test-ying
---
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
  name: fortioserver
  namespace: airmesh-test-ying
spec:
  addresses:
  - 100.124.99.206
  hosts:
  - fortioserver.airmesh-test-ying.svc.ea1.us.a
  location: MESH_INTERNAL
  ports:
    - name: http-port
      number: 8080
      protocol: HTTP
  resolution: STATIC
  workloadSelector:
    labels:
      app: fortioserver
```

In the remote cluster:
```
---
apiVersion: v1
kind: Namespace
metadata:
  name: airmesh-test-ying
  labels:
    istio-injection: enabled
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: fortioclient
  name: fortioclient
  namespace: airmesh-test-ying
spec:
  selector:
    matchLabels:
      app: fortioclient
  template:
    metadata:
      labels:
        app: fortioclient
    spec:
      containers:
        - name: fortioclient
          image: docker-artifactory:443/airmesh-fortio:1.9.1-airbnb1
          args:
            - report
          volumeMounts:
            - name: shared-data
              mountPath: /var/lib/fortio
      volumes:
        - name: shared-data
          emptyDir: {}
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: fortioserver
  name: fortioserver
  namespace: airmesh-test-ying
spec:
  replicas: 1
  selector:
    matchLabels:
      app: fortioserver
  template:
    metadata:
      labels:
        app: fortioserver
    spec:
      containers:
[...]

PR #35369, opened 2021-09-27
remove dependency on old serviceEntry and workloadEntry

hzxuzhonghu:
**Please provide a description of this PR:**

In event handler, remove the dependency on old resources(Se, Wle). So we could do this obey eventual consistent model, and support queue task merge later.

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #35559, opened 2021-10-11
istio: fix service entry stale endpoints

yingzhuivy:

fixes the problem described in this issue:
https://github.com/istio/istio/issues/35404

workloadInstancesByIP is replaced by workloadInstancesByName, i.e.
we don't assume one to one correspondance between wi and ip.
workloadInstancesIPsByName is removed since it's covered by the
new map.

There is a refactor going on for the ServiceEntryStore which should (?)
also fix the problem: https://github.com/istio/istio/pull/35369. But
that probably won't get backported to 1.11, so I created this smaller
fix to be patched onto 1.11.

ptal @howardjohn @hzxuzhonghu


hzxuzhonghu:
@yingzhuivy all the backport should be from master branch

yingzhuivy:
i see. will your refactor fix the stale endpoint issue? can it be backport to 1.11? @hzxuzhonghu

Issue #32056, opened 2021-04-08
Update to EndpointSlices v1


howardjohn:
`discovery.k8s.io/v1beta1 EndpointSlices are deprecated in favor of discovery.k8s.io/v1, and will no longer be served in Kubernetes v1.25. `

Need to take care when we add it for backwards compat

howardjohn:
Not stale

howardjohn:
@dgn I don't know why I put it as a release blocker, just dropped it. But it is good to get it in early, otherwise we end up with the 1.10 problem where users on a supported version (1.9) cannot upgrade to k8s 1.22

PR #35577, opened 2021-10-12
use EndpointSlice v1 in Kubernetes 1.21+

stevenctl:
The EndpointSlice controller will read using the new v1 if the server is running 1.21 or newer. 

fixes https://github.com/istio/istio/issues/32056

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -1093,7 +1093,7 @@ func (c *Controller) hydrateWorkloadInstance(si *model.WorkloadInstance) []*mode
 		ObjectMeta: metav1.ObjectMeta{Namespace: si.Namespace, Labels: si.Endpoint.Labels},
 	}
 
-	// find the services that map to this workload entry, fire off eds updates if the service is of type createClient-side lb
+	// find the services that map to this workload entry, fire off eds updates if the service is of type client-side lb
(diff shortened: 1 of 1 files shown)


stevenctl:
still need to update the usage of v1beta1 structs... trying to avoid a full fork.

stevenctl:
From https://kubernetes.io/docs/reference/using-api/deprecation-guide/

> All existing persisted objects are accessible via the new API

It seems that at they very least the fake k8s client doesn't exhibit this behavior, or that I don't understand this statement. Does "existing" mean, things that already existed before an upgrade? 

> make sure all of our unit tests exercise both paths

Thinking about the best way to go about this. A lot of tests call `.DiscoveryV1Beta1()` directly. Wondering if I should just add something that mirrors between the old/new versions to the fake kube client code.

howardjohn:
If it was easy, I think ideally we would actually do the version mirror for
*all* types. Its pretty annoying writing a test and accidentally using
v1beta1 istio API and it does nothing. Super low priority if its not easy.

Mirror looks good assuming its just for tests. As i am sure you
know, convertRes may not exactly meet api-server behavior (at least
generically - maybe in ES case its ok), but seems good enough for our tests

stevenctl:
Added the mirror for tests.. didn't make it two way, so test writers will have to use v1, but we can trigger the v1beta1 compat path by overriding the k8s version in the client... This means I need to update several tests to check the backwards compat.

PR #35615, opened 2021-10-14
JWKS with empty keys or invalid fields should fail validations

aweis89:
Fixes: #35582 

- Previously we were validating the JWKS using JWT parse [function](https://github.com/istio/istio/pull/35615/files#diff-1488ede8a94d615faf5bbbd17f81fdb16b0409a3cc21ae6525a91991e55df34fL1874), which seemed to only validate that the JSON is valid.
- This creates a new `ValidateJwks` function that relies on the same lib but using the jwk parse [function](https://github.com/lestrrat-go/jwx/blob/4c29e6cfaeb22ae3cbd6a7d3a95424a33dff5079/jwk/jwk.go#L381) function
- Attempts to unify `RequestAuthentication`'s validation logic between static `jwks` and resolved jwks from the `jwksUri` by invoking the above function in both instances


howardjohn:
If the validation changes make configuration that did not get NACKed by
envoy no longer valid - even if they do nothing, are bad, etc generally we
consider it a breaking changes.

Note we can add warnings as well, not sure errors - but should make sure
there are not false positives

aweis89:
TBH I'm not quite sure what the desired behavior should be when the key returned by the `jwksUri` returns an invalid key according to Envoy's validation. Excluding the JWKS from future updates can have obvious security implications. On the other hand, including the key will result in Envoy rejecting all updates which seems even worse

pkg/config/validation/validation.go
```diff
@@ -1851,6 +1851,39 @@ var ValidateRequestAuthentication = registerValidateFunc("ValidateRequestAuthent
...
+		Keys []json.RawMessage `json:"keys"`
+	}{}
+	if err := json.Unmarshal(raw, &jwks); err != nil {
+		return err
+	}
+	// empty keys can cause an invalid Envoy config
+	if len(jwks.Keys) == 0 {
+		return fmt.Errorf("missing keys in jwks: \"%s\"", string(raw))
+	}
+	keyErrs := []error{}
```
yangminzhu, line 1868: I'm still not sure if it's safe to rely on the golang JWT library which does not behave the same as the Envoy c++ JWT library when parsing jwks, a mismatch could potentially cause production issues for some unlucky customers. 

Should we take a step back to only return error for the obvious invalid case? For example, return error only if `jwks.Keys == 0`?

@incfly @liminw @myidpt wdyt?

mandarjog: @yangminzhu don you know of specific differences between the two?

aweis89: For what it's worth, after thinking about this a bit more, I've added my opinion on what I'd like to encompass with the validation logic [here](https://github.com/istio/istio/issues/35582#issuecomment-978118762)

PR #35625, opened 2021-10-15
fix: serviceCluster configalways from serviceCluster flag set but oth…

ysl940114:
…er config never take effect


ysl940114:
> what is the user facing impact of this change?

the config of serviceCluster in pod annotation can take effect as expect

PR #35630, opened 2021-10-15
Guard prioritized leader election with envvar

Monkeyanator:
Adds the `PRIORITIZED_LEADER_ELECTION` environment variable to disable https://github.com/istio/istio/pull/35286


costinm:
Yes, in some clusters istiod may not have permissions to watch mutating webhooks.

PR #35635, opened 2021-10-15
Bump presubmit tests to 1.22

howardjohn:


jacob-delgado:
We have some integration tests that do not run on 1.22. Should we enable them? Put a DNM. Remove if you think it's fine.

howardjohn:
Shankar is working on making the  compatible for the security ones. For the
upgrade stuff,we would be fine iff we only test 1.10+, not sure we can do
much else. We do run them in postsubmit stuff though which is why I felt
this was the right thing to do

PR #35643, opened 2021-10-17
add MultipleYAMLDiff func and fix invalid unit test

nicole-lihui:
- add multiple YAML diff fun
- fix operator_test invalid unit test


nicole-lihui:
not stale

morvencao (commented):
thanks!
this looks good, but we have many other places still using the `YAMLDiff` function, did you tried to replace all of them?

nicole-lihui:
> thanks! this looks good, but we have many other places still using the `YAMLDiff` function, did you tried to replace all of them?
@morvencao 
Good idea, I will finish it next

morvencao (approved):

Issue #31666, opened 2021-03-23
Migrate to google.golang.org/protobuf


howardjohn:
This is the new proto API: https://blog.golang.org/protobuf-apiv2

howardjohn:
Blocker: https://github.com/cncf/udpa/issues/47
Motivation: https://github.com/golang/protobuf/pull/1217/

howardjohn:
not stale

zhlsunshine:
Hi @howardjohn,  shall we move this issue to 1.12?

PR #35659, opened 2021-10-19
Migrate to google.golang.org/protobuf 

howardjohn:
Fixes https://github.com/istio/istio/issues/31666

This pr was WAY bigger than I expected. 

It mostly consists of:
* Boilerplate replacing imports/function names
* Fixing instances where we mixed golang/protobuf and gogo/protobuf, which are now entirely incompatible
* jsonpb changed a bit, both in the call signature and now the output is not deterministic so where we care we need to use caution
* In istioctl we had some hacks to allow unknown Any which required copying some internal code from proto library into our own
* Bumping cncf/xds and cncf/udpa libraries which had some breakages with importing a newer library

Benchmarks show no difference in perf

### DIFF
--- a/istioctl/cmd/precheck.go
+++ b/istioctl/cmd/precheck.go
@@ -28,7 +28,6 @@ import (
 	"github.com/spf13/cobra"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
-	"google.golang.org/protobuf/encoding/protojson"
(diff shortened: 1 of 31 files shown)


jacob-delgado:
Is the plan to also change istio/api too? Should we wait until we branch for 1.12? Simple enough change so I'm thinking we should if possible.

howardjohn:
Blocked by https://github.com/golang/protobuf/issues/1374

PR #35662, opened 2021-10-19
fix operator prune resources for action required case.

### LINKED ISSUE #35657: operator not cleaning up resources
### Bug Description

When using the istio operator as an installation method it doesn't clean up the underlying resources (e.g. services, deployments) when the `IstioOperator ` for ingress is deleted as an example. This was also experienced when testing a canary upgrade from 1.9.7 to 1.10.3. The old 1.9.7 ingress and control plane resources were left around but the custom resource were cleaned up so its difficult to understand why the resources were not cleaned up. 

### CNI
```yaml
kind: IstioOperator
metadata:
  name: cni
  namespace: istio-system
spec:
  profile: empty
  revision: 1-10-3
  components:
    cni:
      enabled: true
  values:
    cni:
      excludeNamespaces:
        - istio-system
        - kube-system
```

### Control Plane
```yaml
kind: IstioOperator
metadata:
  namespace: istio-system
  name: controlplane-1-10-3
spec:
  profile: minimal
  revision: 1-10-3
  values:
    istio_cni:
      enabled: true
  components:
    pilot:
      k8s:
        replicaCount: 2
        resources:
          requests:
            cpu: 100m 
            memory: 512Mi
```

### IngressGateway

```yaml
kind: IstioOperator
metadata:
  namespace: istio-system
  name: ingressgateway-1-10-3
spec:
  profile: empty
  revision: 1-10-3
  components:
    ingressGateways:
      - enabled: true
        name: ingressgateway-1-10-3
        k8s:
          replicaCount: 3
          serviceAnnotations:
            service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: "instance"
            service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
            service.beta.kubernetes.io/aws-load-balancer-type: external
```

### Version

```prose
istioctl 1.10.3

kubectl
Client Version: v1.22.2
Server Version: v1.21.2-eks-0389ca3
```

### Additional Information

More info can be found in the comments on this issue #35016.

morvencao:
fix: https://github.com/istio/istio/issues/35657

Currently the in-cluster operator will check if there are active proxies connecting to the istio control plane before the istio control plane is deleted. But the operator didn't check if the deleted IOP instance actually contains the istiod component, for example if the deleting IOP instance only contains Gateways components, it should be safe to be deleted.

### DIFF
--- a/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
+++ b/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
@@ -235,6 +235,16 @@ func (r *ReconcileIstioOperator) Reconcile(_ context.Context, request reconcile.
 		return reconcile.Result{}, nil
 	}
 
+	var err error
+	iopMerged := &iopv1alpha1.IstioOperator{}
+	*iopMerged = *iop
+	// get the merged values in iop on top of the defaults for the profile given by iop.profile
+	iopMerged.Spec, err = mergeIOPSWithProfile(iopMerged)
+	if err != nil {
+		scope.Errorf(errdict.OperatorFailedToMergeUserIOP, "failed to merge base profile with user IstioOperator CR %s, %s", iopName, err)
+		return reconcile.Result{}, err
+	}
+
 	deleted := iop.GetDeletionTimestamp() != nil
 	finalizers := sets.NewString(iop.GetFinalizers()...)
 	if deleted {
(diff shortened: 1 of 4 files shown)


jacob-delgado (approved):
Should this also get cherry picked into 1.10?

morvencao:
@istio/wg-environments-maintainers Please take a look, the in-cluster operator is broken now.

Monkeyanator (approved):

PR #35668, opened 2021-10-19
Add WasmPlugin integration test

dgn:
Adds an integration test for WasmPlugin.

Also fixes https://github.com/istio/istio/issues/35948


dgn:
Okay, here's a completely different take on this now. Instead of adding a new test, I changed the `customizemetrics` integration test to use the WasmPlugin API instead of manually configuring two EnvoyFilters to load the .wasm module- this also allowed me to test the manifest->json configuration conversion. Works like a charm. Let me know what you think.

dgn:
@mandarjog @bianpengyuan can we get this in? I'd really like to see this cherry-picked to 1.12, too to get some coverage of the Wasm API implementation

dgn:
okay it turns out that without the naming fixes in this PR, the WasmPlugin support is broken in 1.12 - so this is now a bit more urgent

howardjohn (approved):

mandarjog (approved):
lgtm

PR #35674, opened 2021-10-19
multicluster: run a single remote secrets controller

### LINKED ISSUE #33669: Run a single controller for watching multicluster secrets
This is briefly described in [this doc](https://docs.google.com/document/d/1eS72Ozu_tREk3Zw5PTrPA8oGeAMis1VwvkgTgLgUr7g/edit?resourcekey=0-0OnlRQXkFqL09ZXyqMhpog).

Currently we run two controllers watching for remote-secrets for cross-cluster API server access. This on it's own is odd, but it also means that we create 2x kube clients for the same remote clusters.

stevenctl:
fixes https://github.com/istio/istio/issues/33669

* Consolidate to a single instance of the controller watching k8s Secrets, so the k8s registry and credentials controller can share their informer/watch and for each cluster share a k8s client and stop ch. 
* Move metric recording to the top level controller
* Fix an issue where the credential secretcontroller didn't register event handlers for clusters created after the initial sync
* Add a guard to ignore clusters in remote secrets that have the same name as the "main"/"local" cluster to istiod

note to reviewers – there are some renames so it's easier to review if you filter by commit range

Renames:
- `pilot/pkg/secrets` -> `credentials` per John's comment.
- `pkg/kube/secretcontroller` -> `remoteclusters` - this could be backed by something besides Secrets in the future but should share the same interface


howardjohn (commented):
Awesome! LGTM

howardjohn:
Left off approval since I think some others may be interested, I had some nits (do not need to be addressed), and tests were failing. Ping me if you need a stamp.

hzxuzhonghu (commented):
lgtm

stevenctl:
@hzxuzhonghu thanks for the review. 

I ended up:
* Removing more code from bootstrap for local cluster init, rename remoteclusters -> multicluster and have that trigger each handler
* Get rid of clusterID as a param where we already take *Cluster
* don't expose the writable stop channel in the callbacks

howardjohn (approved):

howardjohn:
Should we cherrypick to 1.12 once merged?

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -191,25 +190,22 @@ func (m *Multicluster) AddMemberCluster(clusterID cluster.ID, rc *secretcontroll
...
+				m.remoteKubeControllers[cluster.ID].workloadEntryStore = serviceentry.NewServiceDiscovery(
 					configStore, model.MakeIstioStore(configStore), options.XDSUpdater,
-					serviceentry.DisableServiceEntryProcessing(), serviceentry.WithClusterID(clusterID),
+					serviceentry.DisableServiceEntryProcessing(), serviceentry.WithClusterID(cluster.ID),
 					serviceentry.WithNetworkIDCb(kubeRegistry.Network))
-				m.opts.MeshServiceController.AddRegistry(m.remoteKubeControllers[clusterID].workloadEntryStore)
 				// Services can select WorkloadEntry from the same cluster. We only duplicate the Service to configure kube-dns.
-				m.remoteKubeControllers[clusterID].workloadEntryStore.AppendWorkloadHandler(kubeRegistry.WorkloadInstanceHandler)
+				m.remoteKubeControllers[cluster.ID].workloadEntryStore.AppendWorkloadHandler(kubeRegistry.WorkloadInstanceHandler)
+				m.opts.MeshServiceController.AddRegistryAndRun(m.remoteKubeControllers[cluster.ID].workloadEntryStore, clusterStopCh)
```
hzxuzhonghu, line 199: generally it is better not to mix two ops in one function

stevenctl: Agreed; can split into `AddRegistry` and `RunDeferred`. 

I'd like to do so in a follow-up, the PR is already quite large and I want to get it into 1.12.

therealmitchconnors (approved):
Approve for UX

PR #35675, opened 2021-10-19
Cleanup config manager usage in integration tests

frankbu:

This PR changes the test framework config manager to only apply config to `config` clusters by default, instead of all clusters, because many more callers expect that than don't. The callers that want to apply to all clusters now need to call `ctx.ConfigAll()`, instead of `ctx.Config()` to create a config manager.


frankbu:
@stevenctl @howardjohn I think I've addressed your comments and this is ready to merge now.

douglas-reid (approved):

howardjohn (approved):

Issue #35676, opened 2021-10-19
Wrong routes rendering in relation to clusterDomain


apolovov:
### Bug Description

In some cases, istio cuts off first-level domain in routes domains. Reproducing:

* Create cluster with `clusterDomain: aaa.bbb.not-example.wtf`.
* In namespace `test`, enable istio and deploy the sleep app to it.
* Create two `ServiceEntry`-es:
```
---
kind: ServiceEntry
apiVersion: networking.istio.io/v1alpha3
metadata:
  name: aaa
  namespace: test
spec:
  endpoints:
    - address: aaa.example.wtf
      ports:
        http-7777: 7777
  exportTo:
    - .
  hosts:
    - aaa.example.wtf
  location: MESH_EXTERNAL
  ports:
    - name: http-7777
      number: 7777
      protocol: HTTP
  resolution: DNS
---
kind: ServiceEntry
apiVersion: networking.istio.io/v1alpha3
metadata:
  name: bbb
  namespace: test
spec:
  endpoints:
    - address: bbb.example.net
      ports:
        http-8888: 8888
  exportTo:
    - .
  hosts:
    - bbb.example.net
  location: MESH_EXTERNAL
  ports:
    - name: http-8888
      number: 8888
      protocol: HTTP
  resolution: DNS
```
* Check the routes for port 8888 (everything is fine and domains are as expected):
```
# istioctl -n test pc routes sleep-557747455f-95ff6 --name 8888 -o yaml
- name: "8888"
  validateClusters: false
  virtualHosts:
  - domains:
    - bbb.example.net
    - bbb.example.net:8888
    includeRequestAttemptCount: true
    name: bbb.example.net:8888
    routes:
...
```
* Check the routes for port 7777 (the domains are inconsistent, first-level "wtf" is missing):
```
# istioctl -n test pc routes sleep-557747455f-95ff6 --name 7777 -o yaml
- name: "7777"
  validateClusters: false
  virtualHosts:
  - domains:
    - aaa.example
    - aaa.example:7777
    includeRequestAttemptCount: true
    name: aaa.example.wtf:7777
    routes:
...
```

### Version

```prose
# istioctl version
client version: 1.11.2
control plane version: 1.10.1
data plane version: 1.10.1 (6 proxies)

# kubectl version
[...]

PR #35691, opened 2021-10-20
Fix alt-host generation when hostname has a partial overlap with cluster domain

howardjohn:
For example, a service `foo.bar.local` behaves incorrectly because we
incorrectly match the `.local` up with `.cluster.local`

Fixes https://github.com/istio/istio/issues/35676


ramaraochavali (approved):

ericvn (approved):

howardjohn:
This is currently only fixed in 1.13.0+

fams:
Yes, I saw this. My question is: Do you guys have plans to fix this issue in 1.11.X? This release is still in maintenance, I think.

PR #35694, opened 2021-10-20
Add default to revision tag in manifests

jacob-delgado:

In the master branch, to install using helm you need to do `--set defaultRevision=default` in a fresh install to get a validating webhook. Users should not have to do this to make it simpler for them to install.

### DIFF
--- a/manifests/charts/base/files/gen-istio-cluster.yaml
+++ b/manifests/charts/base/files/gen-istio-cluster.yaml
@@ -6322,3 +6322,38 @@ subjects:
   - kind: ServiceAccount
     name: istiod-service-account
     namespace: istio-system
+---
+# Source: base/templates/default.yaml
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: istiod-default-validator
+  labels:
+    app: istiod
+    release: istio
+    istio: istiod
+    istio.io/rev: default
+webhooks:
+  - name: validation.istio.io
+    clientConfig:
+      service:
+        name: istiod
+        namespace: istio-system
+        path: "/validate"
+    rules:
+      - operations:
+          - CREATE
+          - UPDATE
+        apiGroups:
+          - security.istio.io
+          - networking.istio.io
+        apiVersions:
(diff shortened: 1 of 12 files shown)


howardjohn (commented):
Makes sense to me

Monkeyanator (commented):
Looks good to me, just one comment.

I wonder if we should make `helm install base --defaultRevision` install the default tag webhook too (similar to https://github.com/istio/istio/pull/35047)? The reason we didn't do this is to avoid bringing all the webhook values into the base charts, but because of this, we don't get the prioritized leader election or `istioctl tag list` interop.

Either way, we can document to run `istioctl tag set default` manually after Helm install to get those things, but it'd be nice if it worked out of the box.

jacob-delgado:
> Looks good to me, just one comment.
> 
> I wonder if we should make `helm install base --defaultRevision` install the default tag webhook too (similar to #35047)? The reason we didn't do this is to avoid bringing all the webhook values into the base charts, but because of this, we don't get the prioritized leader election or `istioctl tag list` interop.
> 
> Either way, we can document to run `istioctl tag set default` manually after Helm install to get those things, but it'd be nice if it worked out of the box.

I would say yes. If we want Helm to go Beta this should also be done. We can discuss this more in Environments. Having `istioctl` be a part of Helm installs seems counter-intuitive to me. I think there may be some testing we'd want to add to make sure we get the behavior we want.

Ideally we wanted to get rid of the base chart, but with the validating webhook moved into the base chart t I don't see how it can get removed and cleaned up anyway. It feels like a poor UX to have them separated at this time.

howardjohn (commented):
Makes sense to me, leaving to @Monkeyanator to review closely

Monkeyanator (approved):
LGTM. We'll need to be sure to document that for revisioned Helm install, explicit `--defaultRevision` is still required, and the case this PR improves is just non-revisioned Helm install.

zawachte-c:
Is it possible this pr is causing the istiod-default-validator to always expect installation in istio-system? 

I have been playing around with installing istio in another namespace, but this webhook seems to be the only blocker.

PR #35699, opened 2021-10-21
Skip zero size gateway array checks.

kswapd:


kswapd:
Skip unnecessary  gateway validations when there is no gateway info in this array and code seems much more readable.

PR #35719, opened 2021-10-25
Import from k8s.io/utils/clock instead

jonyhy96:

Change import path to `k8s.io/utils/clock `

Due to [kubernetes/kubernetes#94738](https://github.com/kubernetes/kubernetes/issues/94738)

### DIFF
--- a/pilot/pkg/leaderelection/k8sleaderelection/healthzadaptor_test.go
+++ b/pilot/pkg/leaderelection/k8sleaderelection/healthzadaptor_test.go
@@ -23,7 +23,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


jonyhy96:
ping @hzxuzhonghu @howardjohn  should we push this forward?

howardjohn:
I don't think we should diverge. If the upsteam leaderelection package
changes we can move it but otherwise I don't see a good reason to diverge.
If Kubernetes wants us to get off this library they should update all of
their own usages and mark the package deprecated. Until that happens I
don't see why we should bother?

jonyhy96:
Most of k8s projects has migrated to use new lib but i agree with ur opinion

PR #35727, opened 2021-10-25
Fuzzing: Add more coverage

AdamKorcz:
Adds more fuzzing to increase coverage with no particular prioritization of packages or parts of the code.

A few comments:

1. There are ONLY changes to fuzzing. No code outside of `tests/fuzz` has been modified.
2. I have added a few fuzzers that need to be moved to their respective packages prior to being built. This is all done automatically when OSS-fuzz builds the fuzzers. This imposes a change on the regression testing of the fuzzers that is **not** included in this PR. The change is to add support for regression testing of these fuzzers as well.


AdamKorcz:
Can this one be merged?

howardjohn (approved):

PR #35739, opened 2021-10-26
WIP:  fix svcA->gateway->svcB calling failure if not using fqdn

hzxuzhonghu:
```
spec:
  gateways:
  - mesh
  - istio-system/helloworld-gateway
  hosts:
  - helloworld
  http:
  - match:
    - authority:
        regex: helloworld.*:80
    route:
    - destination:
        host: helloworld
        port:
          number: 80
```

If set up the above vs bound to a gateway, and we setup vs + dr to route traffic calling helloworld  to the gateway, the gateway can not route it.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/gateway.go
+++ b/pilot/pkg/networking/core/v1alpha3/gateway.go
@@ -40,6 +40,7 @@ import (
 	"istio.io/istio/pilot/pkg/networking/plugin"
 	"istio.io/istio/pilot/pkg/networking/util"
 	authn_model "istio.io/istio/pilot/pkg/security/model"
+	"istio.io/istio/pilot/pkg/serviceregistry/provider"
 	"istio.io/istio/pkg/config"
 	"istio.io/istio/pkg/config/gateway"
 	"istio.io/istio/pkg/config/host"
@@ -411,9 +412,10 @@ func (configgen *ConfigGeneratorImpl) buildGatewayHTTPRouteConfig(node *model.Pr
 						vHost.RequireTls = route.VirtualHost_ALL
 					}
 				} else {
+					svc := push.ServiceForHostname(node, hostname)
 					newVHost := &route.VirtualHost{
 						Name:                       util.DomainName(string(hostname), port),
-						Domains:                    buildGatewayVirtualHostDomains(string(hostname), port),
+						Domains:                    buildGatewayVirtualHostDomains(svc, string(hostname), port),
 						Routes:                     routes,
 						IncludeRequestAttemptCount: true,
 					}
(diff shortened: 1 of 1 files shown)


hzxuzhonghu:
@howardjohn anything unclear? How should we handle this case

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -1028,8 +1031,21 @@ func isGatewayMatch(gateway string, gatewayNames []string) bool {
 	return false
 }
 
-func buildGatewayVirtualHostDomains(hostname string, port int) []string {
+func buildGatewayVirtualHostDomains(service *model.Service, hostname string, port int) []string {
 	domains := []string{hostname}
+	if service != nil && service.Attributes.ServiceRegistry == provider.Kubernetes {
```
howardjohn, line 1036: I think the `service.Attributes.ServiceRegistry == provider.Kubernetes` part is problematic, at least wrt to current code.

Use your VS example but replace `helloworld` with something that is NOT a Service. You will get only `fake-name.ns.svc.cluster.local`.

It seems like if we expand it, we should always fully expand it. But we probably shouldn't expand shortname at all in this case since we don't know its a k8s service...

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -1028,8 +1031,21 @@ func isGatewayMatch(gateway string, gatewayNames []string) bool {
...
+		domains = append(domains,
+			service.Attributes.Name,
```
howardjohn, line 1038: This is dangerous. Consider VS in `bar` namespace AND namespace in `foo` namespace for `helloworld`.

1. We may get duplicate domains. I think we filter them out, but will give a ton of warnings + behavior based on creation time of service which is not great.
2. Say the `foo` namepace wins the conflict. If the `bar` namespace calls `helloworld`, they will unexpectedly go to `foo`.

Knative does something very similar with their cluster-local-gateway, and they do NOT support this form.

howardjohn (commented):
Overall I am a bit concerned about this.. I feel like expanding names at the gateway is a bit of a concern - even the current expansion we do. You can still get the behavior by specifying all the domains (just 3) in the virtual service. We actually optimized this as well - it used to create 3 routes but now creates just one. Knative does this extensively.

pilot/pkg/networking/core/v1alpha3/gateway.go
hzxuzhonghu: Yes, it is documented shortnames can be used only for k8s.

pilot/pkg/networking/core/v1alpha3/gateway.go
hzxuzhonghu: Yes, for different namespaces, when there are two same name service.

hzxuzhonghu:
> You can still get the behavior by specifying all the domains (just 3) in the virtual service. 

It is not perfect:

1. we can not enforce user to write 3 same 
 - productpage
 - productpage.default
 - productpage.default.svc
2. productpage will be extended to FQDN, so an access with http://productpage:port is still impossible

pilot/pkg/networking/core/v1alpha3/gateway.go
howardjohn: Right.. its not safe.

If I have `helloworld` in a sidecar, its fine because we know the namespace of the client (its in the same pod!). For a gateway, we have clients from all namespaces.

howardjohn:
> we can not enforce user to write 3 same


Well you can if they want this behavior right?

> productpage will be extended to FQDN, so an access with http://productpage:port is still impossible

I don't think this is a good idea in general, but FQDN is explicitly _not_ supported at Gateway. So this would also only work with `Gateway.host=*` which is a fairly dangerous + uncommon pattern

PR #35746, opened 2021-10-26
Enable TestKubeInject in istiodremote topology

frankbu:


costinm (approved):
In a future PR, it would be great to also pass the root cert of the XDS server and make sure istioctl is checking it.

PR #35762, opened 2021-10-27
update JWT claim routing implementation

yangminzhu:

The PR updates the JWT claim routing implementation with the following changes:

- changes the special header name to `@request.auth.names`, see details in https://github.com/istio/api/pull/2134
- adds analyzer for virtual service using JWT claim based routing and returns error if the gateway pod does not have request authentication policies.
- adds the validation webhook to reject the virtual service using JWT claim based routing without gateways

This is needed for Istio 1.12.

cc @mandarjog @howardjohn


yangminzhu:
@howardjohn @mandarjog could you take a look again, I have updated the PR to match the latest API, needed for 1.12, thanks.

howardjohn (approved):

yangminzhu:
@istio/wg-user-experience-maintainers ptal, this is needed for 1.12, thanks.

yangminzhu:
@istio/wg-user-experience-maintainers ping

esnible (approved):
Galley code LGTM

galley/pkg/config/analysis/analyzers/virtualservice/jwtclaimroute.go
```diff
@@ -0,0 +1,142 @@
...
+			// The gateway does not exist, this should already be covered by the gateway analyzer.
+			continue
+		}
+
+		gw := gwRes.Message.(*v1alpha3.Gateway)
+		gwSelector := k8s_labels.SelectorFromSet(gw.Selector)
+
+		// Check each pod selected by the gateway.
+		c.ForEach(collections.K8SCoreV1Pods.Name(), func(rPod *resource.Instance) bool {
+			pod := rPod.Message.(*v1.Pod)
```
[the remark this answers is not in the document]
therealmitchconnors: The ForEach function allows for the possibility of iterators, which will not load the entire collection contents into memory, though I do not recall if this is used in the current implementation.  Generally, analyzers often have O(M*N) because we are joining two unindexed collections together.  I have not yet seen any user pain related to the slow performance of individual analyzers.

galley/pkg/config/analysis/msg/messages.yaml
```diff
@@ -535,3 +535,16 @@ messages:
     description: "user namespace should be injectable if Istio is installed with enableNamespacesByDefault enabled and neither injection label is set."
     template: "is enabled for Istio injection, as Istio is installed with enableNamespacesByDefault as true."
     url: "https://istio.io/latest/docs/reference/config/analysis/ist0148/"
+
+  - name: "JwtClaimBasedRoutingWithoutRequestAuthN"
+    code: IST0149
+    level: Error
```
therealmitchconnors, line 541: Is this condition always an error?  Is it possible to use this header for some other purpose, which would trigger this analyzer?

yangminzhu: it shouldn't, the `@` prefix already makes it an invalid HTTP header, if a customer puts it in virtual service today, it will not work at all, and Envoy will simply reject the header with HTTP code 400.

PR #35763, opened 2021-10-27
Add support for recursive file watch to pilot

howardjohn:
This is useful when users are trying to aggregate multiple sources into
a single folder to watch all of them. For example, mount multiple
configmaps to a common parent directory.


howardjohn:
That library is totally overkill IMO. I think it was working around lack of proper fsnotify support when it was made 3 years ago. Now there is no reason to introduce so much complexity - fsnotify library handles everything we need for us.

hzxuzhonghu:
>The watch limit is set from kernel..

@costinm I confirmed the answer is wrong, the limit can be set in container.

hzxuzhonghu:
@howardjohn Looks it is resolved by `AddRaw`  in https://github.com/fsnotify/fsnotify/pull/289

hzxuzhonghu (approved):

PR #22598, opened 2020-03-30
Properly stop/start namespace controller for leader election

howardjohn:
Might fix https://github.com/istio/istio/issues/22463, was found during
debugging this.

https://github.com/istio/istio/pull/21669 made a change that set the
stop channel to the global stop channel instead of leader election stop
channel, so it doesn't stop when leader is lost. Once that is fixed we
need to create new informers, or client-go will crash due to closing the
same informer twice

PR #35768, opened 2021-10-28
Wait for cache sync before proceeding with leader election

### LINKED ISSUE #35765: Prioritized leader election: race condition at start leads to stealing lock unexpectedly
Everytime my non-default revision starts it is taking the leader election lock. I think the problem is:
https://github.com/istio/istio/blob/7d5070373f99613e59b3aa6f0a425a3fe115a21f/pilot/pkg/leaderelection/leaderelection.go#L161

We do not wait for the informer to sync. As a result, `l.defaultWatcher.GetDefault()` is `""` and KeyComparison returns `true` the first attempt.

I think we need to:
* Add a WaitForCacheSync.
* Possibly just use one default watcher so we don't need to sync N times

Monkeyanator:
Makes it so that we wait for caches to sync before proceeding with leader election. This prevents a race condition where revision with name "" would steal the leader lock on startup

fixes #35765

### DIFF
--- a/pilot/pkg/leaderelection/leaderelection.go
+++ b/pilot/pkg/leaderelection/leaderelection.go
@@ -24,6 +24,7 @@ import (
 	"go.uber.org/atomic"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
 
 	"istio.io/istio/pilot/pkg/features"
 	"istio.io/istio/pilot/pkg/leaderelection/k8sleaderelection"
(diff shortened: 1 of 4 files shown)


Monkeyanator:
Looks like this introduced a real bug resulting in `'MountVolume.SetUp failed for volume "istiod-ca-cert" : configmap "istio-ca-root-cert"`. Seems like it could be related to the namespace controller and some weird interactions it's had in the past with leaderelection (https://github.com/istio/istio/pull/22598).

Monkeyanator:
> Looks like this introduced a real bug resulting in 'MountVolume.SetUp failed for volume "istiod-ca-cert" : configmap "istio-ca-root-cert". Seems like it could be related to the namespace controller and some weird interactions it's had in the past with leaderelection (#22598). Could be because we use the remote cluster's client in the default watcher in the multicluster leader election (just a hunch)

This should not have merged. I tried removing the `cache.WaitForCacheSync` to see if it fixed the race condition above and it looks like it did! But now we're not waiting for queue to process before running the leader election.

PR #35795, opened 2021-10-29
configure request without query access log formatter

ramaraochavali:
Configures access log formatter for REQ_WITHOUT_QUERY - Without this if user specifies it access log format, the listener will be rejected.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/accesslog.go
+++ b/pilot/pkg/networking/core/v1alpha3/accesslog.go
@@ -18,14 +18,13 @@ import (
 	"sync"
 
 	accesslog "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v3"
-	formatters "github.com/envoyproxy/go-control-plane/envoy/extensions/formatter/req_without_query/v3"
-
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	listener "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 	fileaccesslog "github.com/envoyproxy/go-control-plane/envoy/extensions/access_loggers/file/v3"
 	grpcaccesslog "github.com/envoyproxy/go-control-plane/envoy/extensions/access_loggers/grpc/v3"
 	hcm "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
 	tcp "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/tcp_proxy/v3"
+	formatters "github.com/envoyproxy/go-control-plane/envoy/extensions/formatter/req_without_query/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
 	structpb "google.golang.org/protobuf/types/known/structpb"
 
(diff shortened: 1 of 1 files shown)


ramaraochavali:
> Why doesn't envoy just do this automatically? seems excessively verbose and error prone...

I had the same doubt - Reading this https://github.com/envoyproxy/envoy/issues/7583#issuecomment-820215673 seems it is deliberate choice but I agree this is a very common need so should have been built in formatter.

@dio Do you have better insight on why this was chosen as custom formatter?

dio:
Yeah, I think we should continue the effort on the envoy side.

ramaraochavali:
@howardjohn given that we may need more discussions on the Envoy, Do you think we can go with this for now and if/when makes that change we can remove it?

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -102,6 +103,15 @@ var (
 
 	// accessLogBuilder is used to set accessLog to filters
 	accessLogBuilder = newAccessLogBuilder()
+
+	// accessLogFormatters configures additional formatters needed for some of the format strings like "REQ_WITHOUT_QUERY"
+	accessLogFormatters = []*core.TypedExtensionConfig{
```
howardjohn, line 108: shouldn't we only add it if they use `REQ_WITHOUT_QUERY`? Otherwise we add this filter to everyone when 99.9% of users do not use it?

ramaraochavali: I initially did that but later removed it because the check.is not needed as this won't add any overhead and envoy will use it if the format has that field

howardjohn: Well it does have overhead as the XDS will become even more bloated. The access logs are one of the biggest parts of XDS already

ramaraochavali: I am not sure if this formatter array would cause significant bloating. But good to add it only when needed. Added PTAL.

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -231,6 +245,9 @@ func buildFileAccessLogHelper(path string, mesh *meshconfig.MeshConfig) *accessl
 				jsonLogStruct = &parsedJSONLogStruct
 			}
 		}
+		if _, exists := jsonLogStruct.Fields[requestWithoutQuery]; exists {
```
howardjohn, line 248: wouldn't requestWithoutQuery be in the values, not the keys?

howardjohn (approved):

pilot/pkg/networking/core/v1alpha3/accesslog.go
ramaraochavali: Oh. I thought is field name with empty value. I have changed. PTAL

ramaraochavali:
@howardjohn PTAL changed to use value.

PR #35797, opened 2021-10-29
Fuzzing: Add CIFuzz

AdamKorcz:

This adds CIFuzz to Istios OSS-fuzz integration. Upon merging this PR, Istios fuzzers will run in the CI for 1200 seconds when a pull request is made. 

When https://github.com/istio/istio/pull/35727 gets merged, the 1200 seconds should be increased significantly.

Documentation for CIFuzz can be found at: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'istio'
+        dry-run: false
+        language: go
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'istio'
+        fuzz-seconds: 1200
+        dry-run: false
+        language: go
+    - name: Upload Crash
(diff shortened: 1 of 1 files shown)


AdamKorcz:
If this PR can be rejected, feel free to close it.

howardjohn:
I think we can close this. Having fuzzers run in our CI I think is useful, but it would need to be in prow. IMO we can wait 3 months for go 1.18 which will make it much simpler. Thanks for the PR!

jonathanmetzman:
> I think we can close this. Having fuzzers run in our CI I think is useful, but it would need to be in prow. IMO we can wait 3 months for go 1.18 which will make it much simpler. Thanks for the PR!

We actually do support Prow. If youre interested in running CIFuzz, I can work on docs and we could even do a video meeting so I can help you set it up if needed.

howardjohn:
Ah cool. I assume this would be https://google.github.io/clusterfuzzlite/running-clusterfuzzlite/prow/?

Will there be support for go 1.18 fuzzing eventually? If so it may be desirable to wait for that to avoid too many changes in this area. Our core fuzzing is https://github.com/envoyproxy/envoy which is completely separate from this project (in terms of CI), so the fuzzers here are not the top priority that would make sense to invest much effort ifwe will change things in 2 months.

AdamKorcz:
> Will there be support for go 1.18 fuzzing eventually?

Yes. This will be added in 2022.

jonathanmetzman:
> Ah cool. I assume this would be https://google.github.io/clusterfuzzlite/running-clusterfuzzlite/prow/?

Right. I just need to change that document to be CIFuzz specific instead of ClusterFuzzLite specific (both projects use the same code under the hood).

PR #35827, opened 2021-11-02
change istiod svc 15012 port name from https-dns to tls-istiod

zackzhangkai:
Signed-off-by: zackzhangkai <zhangkaiamm@gmail.com>

The port 15021 of svc istiod should not be `https-dns`, for this port is `grpc + tls`.

I think that changing istiod svc 15012 port name from `https-dns` to `tls-istiod` is more appropriate.


hzxuzhonghu (commented):
why `operator/cmd/mesh/testdata/manifest-generate/data-snapshot.tar.gz`

zackzhangkai:
> why `operator/cmd/mesh/testdata/manifest-generate/data-snapshot.tar.gz`

@hzxuzhonghu In the test, will extract the gz package, and patch the port `tls-istiod`. The old gz pkg is still `https-dns`. So should modify it.

[test code](https://github.com/istio/istio/blob/master/operator/cmd/mesh/manifest-generate_test.go#L402)

```golang
	// Snapshot charts are in testdata/manifest-generate/data-snapshot
	snapshotCharts = func() chartSourceType {
		d, err := os.MkdirTemp("", "data-snapshot-*")
		if err != nil {
			panic(fmt.Errorf("failed to make temp dir: %v", err))
		}
		f, err := os.Open("testdata/manifest-generate/data-snapshot.tar.gz")
		if err != nil {
			panic(fmt.Errorf("failed to read data snapshot: %v", err))
		}
		if err := tgz.Extract(f, d); err != nil {
			panic(fmt.Errorf("failed to extract data snapshot: %v", err))
		}
		return chartSourceType(filepath.Join(d, "manifests"))
```

```golang
func TestManifestGeneratePilot(t *testing.T) {
	runTestGroup(t, testGroup{
		{
			desc:       "pilot_default",
			diffIgnore: "CustomResourceDefinition:*:*,ConfigMap:*:istio",
		},
		{
			desc:       "pilot_k8s_settings",
			diffSelect: "Deployment:*:istiod,HorizontalPodAutoscaler:*:istiod",
		},
		{
			desc:       "pilot_override_values",
			diffSelect: "Deployment:*:istiod,HorizontalPodAutoscaler:*:istiod",
		},
		{
			desc:       "pilot_override_kubernetes",
			diffSelect: "Deployment:*:istiod, Service:*:istiod,MutatingWebhookConfiguration:*:istio-sidecar-injector,ClusterRoleBinding::istio-reader-istio-system",
		},
```

howardjohn (changes requested):
I don't think we should change this. We are serving TLS encrypted http traffic which is HTTPS. This is an appropriate name

Even if TLS was a slightly better name, this would add churn that may cause user disruption on upgrade.

zackzhangkai:
Got it. Closed this PR.

PR #35832, opened 2021-11-02
Wait for pods to be deleted when cleaning up operators

### LINKED ISSUE #35821: TestPostInstallControlPlaneVerification test flake
Example: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/35354/integ-ipv6-k8s-tests_istio/1455215065903927296

Basically what happens is TestController runs and install revision=v2. Then TestPostInstallControlPlaneVerification runs and runs `istioctl verify-install`. This triggers this code: https://github.com/istio/istio/blob/2177595595a191f8812c56f1376fe152974585f2/istioctl/pkg/verifier/verifier.go#L151 which detects the revision based on istiod pods. The pods from v2 are still terminating, causing this to incorrectly pick `v2` and then later fail

dddddai:

Wait for pods to be deleted when cleaning up operators to avoid impacting `TestPostInstallControlPlaneVerification`

Fix #35821

### DIFF
--- a/tests/integration/operator/switch_cr_test.go
+++ b/tests/integration/operator/switch_cr_test.go
@@ -255,6 +255,22 @@ func cleanupInClusterCRs(t framework.TestContext, cs cluster.Cluster) {
 	} else {
 		t.Logf("failed to list existing CR: %v", err.Error())
 	}
+
+	// wait for pods in istio-system to be deleted
+	err = retry.UntilSuccess(func() error {
+		podList, err := cs.Kube().CoreV1().Pods(IstioNamespace).List(context.TODO(), kubeApiMeta.ListOptions{})
+		if err != nil {
+			return err
+		}
+		if len(podList.Items) == 0 {
+			return nil
+		}
+		return fmt.Errorf("pods still remain in %s", IstioNamespace)
+	}, retry.Timeout(retryTimeOut), retry.Delay(retryDelay))
+
+	if err != nil {
+		t.Logf("failed to delete pods in %s: %v", IstioNamespace, err)
+	}
 }
 
 func installWithCRFile(t framework.TestContext, ctx resource.Context, cs cluster.Cluster, s *image.Settings,


tests/integration/operator/switch_cr_test.go
```diff
@@ -255,6 +255,22 @@ func cleanupInClusterCRs(t framework.TestContext, cs cluster.Cluster) {
 	} else {
 		t.Logf("failed to list existing CR: %v", err.Error())
 	}
+
+	// wait for pods in istio-system to be deleted
+	err := retry.UntilSuccess(func() error {
+		podList, err := cs.Kube().CoreV1().Pods(IstioNamespace).List(context.TODO(), kubeApiMeta.ListOptions{})
```
howardjohn, line 261: We maybe want to add a label selector to only match the appropriate `istio.io/rev`. But I guess if this passes tests it is ok

dddddai: This function deletes all operators no matter if istio.io/rev is set, so it should delete all pods, shouldn't it? Please correct me if i'm wrong

howardjohn: I was thinking maybe there are non-istio pods that happen to be deployed in the users cluster. But probably not worth bothering with it

howardjohn (approved):
ty!

tests/integration/operator/switch_cr_test.go
```diff
@@ -255,6 +255,22 @@ func cleanupInClusterCRs(t framework.TestContext, cs cluster.Cluster) {
...
+			return err
+		}
+		if len(podList.Items) == 0 {
+			return nil
+		}
+		return fmt.Errorf("pods still remain in %s", IstioNamespace)
+	}, retry.Timeout(retryTimeOut), retry.Delay(retryDelay))
+
+	if err != nil {
+		t.Logf("failed to delete pods in %s: %v", IstioNamespace, err)
```
therealmitchconnors, line 272: Why is this not a failure condition?

PR #35834, opened 2021-11-02
Expose canonical-name and canonical-revision to grpc-agent container

lidizheng:

In the Envoy injection template, we specifies 3 resource labels manually (instead of using the downward API) as metadata. These labels are critical to allow gRPC join the Istio observability.

https://github.com/istio/istio/blob/bbf184e3bcfac8ffaae796f4038bc7bf495ae7b3/manifests/charts/istio-control/istio-discovery/files/injection-template.yaml#L3-L7

These labels in the template has revisioned many time, I couldn't find the original PR adding them. If you think we should add them at another abstraction level, please advice. Thanks ahead.

CC @howardjohn @costinm

---
This change is [Reviewable](https://reviewable.io/reviews/istio/istio/35834)

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
@@ -1,6 +1,9 @@
 {{- $containers := list }}
 {{- range $index, $container := .Spec.Containers }}{{ if not (eq $container.Name "istio-proxy") }}{{ $containers = append $containers $container.Name }}{{end}}{{- end}}
 metadata:
+  labels:
+    service.istio.io/canonical-name: {{ index .ObjectMeta.Labels `service.istio.io/canonical-name` | default (index .ObjectMeta.Labels `app.kubernetes.io/name`) | default (index .ObjectMeta.Labels `app`) | default .DeploymentMeta.Name  | quote }}
+    service.istio.io/canonical-revision: {{ index .ObjectMeta.Labels `service.istio.io/canonical-revision` | default (index .ObjectMeta.Labels `app.kubernetes.io/version`) | default (index .ObjectMeta.Labels `version`) | default "latest"  | quote }}
   annotations: {
     {{- if eq (len $containers) 1 }}
     kubectl.kubernetes.io/default-logs-container: "{{ index $containers 0 }}",
--- a/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
@@ -992,6 +992,9 @@ data:
         {{- $containers := list }}
         {{- range $index, $container := .Spec.Containers }}{{ if not (eq $container.Name "istio-proxy") }}{{ $containers = append $containers $container.Name }}{{end}}{{- end}}
         metadata:
+          labels:
+            service.istio.io/canonical-name: {{ index .ObjectMeta.Labels `service.istio.io/canonical-name` | default (index .ObjectMeta.Labels `app.kubernetes.io/name`) | default (index .ObjectMeta.Labels `app`) | default .DeploymentMeta.Name  | quote }}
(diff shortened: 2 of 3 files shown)


manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
```diff
@@ -1,6 +1,10 @@
 {{- $containers := list }}
 {{- range $index, $container := .Spec.Containers }}{{ if not (eq $container.Name "istio-proxy") }}{{ $containers = append $containers $container.Name }}{{end}}{{- end}}
 metadata:
+  labels:
+    security.istio.io/tlsMode: {{ index .ObjectMeta.Labels `security.istio.io/tlsMode` | default "istio"  | quote }}
```
howardjohn, line 5: How will this be used? AFAIK auto mtls is not really compatible with proxyless gRPC since the application determines if it uses mTLS, not the control plane?

howardjohn (commented):
lgtm aside from my question
Thanks!

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
lidizheng: Removed, thanks for the quick review.

To turn on mTLS, gRPC applications need to: 1. pass xDS credentials to client/server; 2. enable the mTLS in the control plane.

In the existing [mTLS setup guide](https://cloud.google.com/traffic-director/docs/security-proxyless-setup#configure-td-security) and (the [blog post](https://istio.io/latest/blog/2021/proxyless-grpc/#enabling-mtls)), I didn't find a pointer to this label (well, this is an Istio specific label). So, I removed it for now, as gRPC doesn't understand it.

In the long-term, could this flag help controlling the security feature for gRPC applications? Can I ask how does Envoy perform auto-mTLS?

CC @sanjaypujare

howardjohn: In envoy, the configuration of the proxy is _100% controlled by XDS_. This is not the case in gRPC. 

Because we control the proxy 100%, the control plane is smart enough to know which servers at supporting mTLS and which are not. With gRPC we lose this.

I personally think it makes PSM a less effective product, although the reasons for doing it are reasonable. See discussion in https://github.com/grpc/grpc-go/issues/4123 and https://github.com/grpc/proposal/pull/184#discussion_r557617865.

This means that one of the largest friction points of a service mesh (configuring TLS properly) which has been solved in Istio already, will still occur on proxyless gRPC.

howardjohn (approved):
/ok-to-test

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
lidizheng: Thanks for the context. I see now how gRPC applications need an extra step to enable mTLS.

Opt-in xDS at the library level is more invasive than opt-in the sidecar model. For gRPC clients, PSM requires the schema change, so they need to update anyway. For gRPC servers, you are right, this will be an extra step.

sanjaypujare: > For gRPC clients, PSM requires the schema change, so they need to update anyway.

Strictly speaking the target string (which contains the schema `xds:`) can come from a run-time value such as a command line argument so strictly speaking even in the client it may not require code change. The specification of the channel credential (whether xDS creds or plaintext or TLS creds) can also depend on some run time value but more likely to be hardcoded in the code. The xDS-creds enables the app developer to control whether they want to rely on xDS provided security or use their own security. 

> For gRPC servers, you are right, this will be an extra step.

This is strictly speaking not true either. If they want an xDS-managed server they need to use the XdsServerBuilder (instead of the previous ServerBuilder) and the XdsServerBuilder then takes XdsServer creds or other types of creds but using XdsServer creds with fallback creds seems to be the standard option to use.

costinm:
I mean: the PR is good, but we may want to do this in the actual go code, and not depend on the template. 

We want to support proxyless gRPC on VMs and CloudRun, where injection is not available - so ideally the template should be minimal/empty.

lidizheng:
In the Envoy template, we do set default values to these two labels:

- `service.istio.io/canonical-name`: deployment name
- `service.istio.io/canonical-revision`: "latest"

By Go code, do you mean the bootstrap generator? That's a good point. This PR can be one of the source that the bootstrap generator reads.

PR #35354, opened 2021-09-24
Test environments more

howardjohn:
**Please provide a description of this PR:**

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #35846, opened 2021-11-02
gateway controller: run as privileged

howardjohn:
The alternative here is to automagically select a >1024 port if they use
one like 80. I think this is a really bad idea, since its complicated
*and cannot ever be changed without downtime*. In the near future,
Kubernetes will make NET_BIND_SERVICE the default (or runnable without
priv, I forget the details), so we should rely on that plan long term.

This works today because our image implicitly runs as root. However, the
distroless image doesn't, so distroless is broken. This is just
formalizing the fact that we run as root - it is not escalating
privileges beyond what are used today already. Also, this is only for
the alpha gateway implementation, not impacting prod usage.

Tested in https://github.com/istio/istio/pull/35354/files.

### DIFF
--- a/pilot/pkg/config/kube/gateway/templates/deployment.yaml
+++ b/pilot/pkg/config/kube/gateway/templates/deployment.yaml
@@ -32,6 +32,10 @@ spec:
           .Labels
           | nindent 8}}
     spec:
+      securityContext:
+        runAsUser: 0
+        runAsGroup: 1337
+        runAsNonRoot: false
       containers:
       - image: auto
         name: istio-proxy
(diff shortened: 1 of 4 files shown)


pilot/pkg/config/kube/gateway/templates/deployment.yaml
```diff
@@ -32,13 +32,26 @@ spec:
           .Labels
           | nindent 8}}
     spec:
+      securityContext:
```
costinm, line 35: Flags, options.

costinm (commented):
Looks good, but we should have an option to control.

And I don't think we need both run as root and NET_BIND_SERVICE. If net_bind_service is supported, it should be used. If not - fallback to run as root.

howardjohn:
Opening a new PR

PR #35849, opened 2021-11-02
echo: support dualstack --bind-localhost

howardjohn:
Tested in https://github.com/istio/istio/pull/35354/

### DIFF
--- a/pkg/test/echo/server/instance.go
+++ b/pkg/test/echo/server/instance.go
@@ -15,8 +15,10 @@
 package server
 
 import (
+	"context"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/test/echo/server/instance.go
```diff
@@ -122,6 +132,41 @@ func (s *Instance) Start() (err error) {
...
+	v4, v6 := false, false
+	// Obtain all the IPs from the node
+	ipAddrs, ok := network.GetPrivateIPs(context.Background())
+	if !ok {
+		return []string{ip}
+	}
+	for _, ip := range ipAddrs {
+		addr := net.ParseIP(ip)
+		if addr == nil {
+			// Should not happen
```
nmittler, line 150: log? panic?

nmittler (approved):

PR #35851, opened 2021-11-02
ipv6: fix VM healthcheck ipv6 detection

howardjohn:
Currently its 100% broken. SupportsIPv6 ALWAYS returns false since we
never called DiscoverIPVersions. Even if we fix that, we still should
not use it since we need to be consistent with the code that selects
where Envoy admin server binds, which uses `network.IsIPv6Proxy`


lambdai (approved):
LGTM

Issue #34051, opened 2021-07-14
TestServiceExportedInOneCluster test flake


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/integ-pilot-multicluster-tests_istio_postsubmit/1415399944373669888
https://prow.istio.io/view/gs/istio-prow/logs/integ-pilot-multicluster-tests_istio_postsubmit/1412820797625470976

nmittler:
I suspect something must not be getting updated/pushed when we [updateXDS](https://github.com/istio/istio/blob/81a739548db1218563822c8fe31101d5108f4658/pilot/pkg/serviceregistry/kube/controller/serviceexportcache.go#L97). Perhaps the issue is that it's only clearing the EDS cache for the cluster where the `ServiceExport` was created? Maybe we need to clear EDS for the service in all clusters?

stevenctl:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/34086/integ-pilot-multicluster-tests_istio/1417269147036815360

howardjohn:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/34258/integ-pilot-multicluster-tests_istio/1418407700542263296

Looks like https://github.com/istio/istio/pull/34143 is not the fix

howardjohn:
fyi https://prow.istio.io/view/gs/istio-prow/logs/integ-pilot-multicluster-tests_istio_postsubmit/1425152917559578624 is still occuring

nmittler:
Quick update ... still appears to be happening after a variety of recent updates to the MCS logic: https://prow.istio.io/view/gs/istio-prow/logs/integ-pilot-multicluster-tests_istio_postsubmit/1447618309238296576

nmittler:
not stale

PR #35852, opened 2021-11-02
mcs test: backoff more aggresively

howardjohn:
Might help with https://github.com/istio/istio/issues/34051

I think we just retry to much, exhausting all ephemeral ports...

Also make it so you can rerun the test for local dev


nmittler (approved):

hzxuzhonghu:
seems no help

PR #35853, opened 2021-11-02
tf: fix backoff

howardjohn:
Currently we only backoff 2x, since we always use `base * 2` instead of
`previous * 2`. This leads to most calls occuring every 20ms for tests
not overriding defaults, which is a bit fast.

Might help with https://github.com/istio/istio/issues/34051, which is
caused by overloading


ericvn (approved):

ericvn:
It doesn't appear to have totally helped https://github.com/istio/istio/issues/34051.

PR #35855, opened 2021-11-02
Fix filterchain match's incorrect early match

mandarjog:
Fixing incorrectly matched filterchain if `lMatch.FilterChain==nil`.
This moves filtering unrelated to `lMatch.FilterChain` before the nil check.

Signed-off-by: Mandar U Jog <mjog@google.com>

manual cherry pick for https://github.com/istio/istio/pull/35836

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/envoyfilter/listener_patch.go
+++ b/pilot/pkg/networking/core/v1alpha3/envoyfilter/listener_patch.go
@@ -617,6 +617,13 @@ func filterChainMatch(listener *xdslistener.Listener, fc *xdslistener.FilterChai
 		return true
 	}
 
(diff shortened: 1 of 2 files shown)


jwendell:
@mandarjog Is this worth a release note entry?

mandarjog:
@jwendell This is a bug fix. it just ensures that multiple chains are no inadvertently matched.

mandarjog:
@jwendell ping !

jwendell:
@mandarjog I know this is a bug fix. That's exactly why I asked about a release note. Given .z releases mostly contain bug fixes, users like to know what's new in a .z release.

Let me know if you still don't think a release note entry is not needed in this case and I'll approve it. Thanks.

jwendell (approved):

Issue #35860, opened 2021-11-03
vm health check: unexpected error if probe response happens before first XDS request


howardjohn:
https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/35847/integ-ipv6-k8s-tests_istio/1455650257617555456/artifacts/security-013ad14b4942492d845bd2/TestReachability/beta-mtls-on/a_in_primary-0/http_to_vm:http_positive/_test_context/test-ns1-1-20588-state1325041659/cluster-0/vm-v1-8677957464-hl62h_istio-proxy.envoy.log

```
2021-11-02T21:42:27.403263Z	info	healthcheck	failure threshold hit, marking as unhealthy: Get "http://127.0.0.1:15000/stats?usedonly&filter=^(cluster_manager\\.cds|listener_manager\\.lds)\\.(update_success|update_rejected)$": dial tcp 127.0.0.1:15000: connect: connection refused
2021-11-02T21:42:27.405496Z	warn	xdsproxy	upstream [1] terminated with unexpected error rpc error: code = InvalidArgument desc = missing node information
```

Root cause is that the Health DiscoveryRequest comes before the initial CDS request, so we have no node info.

I don't think this would cause any real world issues but is not ideal.

PR #35863, opened 2021-11-03
Send healthcheck after lds request has been sent

hzxuzhonghu:

Fix https://github.com/istio/istio/issues/35860

### DIFF
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -169,6 +169,7 @@ func (s *DiscoveryServer) receive(con *Connection) {
 		if firstRequest {
 			// probe happens before envoy sends first xDS request
 			if req.TypeUrl == v3.HealthInfoType {
+				log.Warnf("ADS: %q %s send health check probe before normal xDS request", con.PeerAddr, con.ConID)
 				continue
 			}
 			firstRequest = false
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -167,6 +167,10 @@ func (s *DiscoveryServer) receive(con *Connection) {
 		}
 		// This should be only set for the first request. The node id may not be set - for example malicious clients.
 		if firstRequest {
+			// probe happens before envoy sends first xDS request
+			if req.TypeUrl == v3.HealthInfoType {
+				continue
+			}
 			firstRequest = false
 			if req.Node == nil || req.Node.Id == "" {
 				con.errorChan <- status.New(codes.InvalidArgument, "missing node information").Err()
--- a/pkg/istio-agent/xds_proxy.go
+++ b/pkg/istio-agent/xds_proxy.go
@@ -324,50 +324,6 @@ func (p *XdsProxy) handleStream(downstream adsStream) error {
 	p.RegisterStream(con)
 	defer p.UnregisterStream(con)
 
-	// Handle downstream xds
-	initialRequestsSent := false
-	go func() {
-		for {
-			// From Envoy
-			req, err := downstream.Recv()
-			if err != nil {
-				select {
-				case con.downstreamError <- err:
-				case <-con.stopChan:
-				}
-				return
-			}
-			if req.TypeUrl == v3.HealthInfoType && !initialRequestsSent {
-				// only send healthcheck probe after LDS request has been sent
-				continue
-			}
-			// forward to istiod
-			con.sendRequest(req)
-			if !initialRequestsSent && req.TypeUrl == v3.ListenerType {
-				// fire off an initial NDS request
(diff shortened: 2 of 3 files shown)


ramaraochavali:
LGTM

pilot/pkg/xds/ads.go
```diff
@@ -167,6 +167,10 @@ func (s *DiscoveryServer) receive(con *Connection) {
 		}
 		// This should be only set for the first request. The node id may not be set - for example malicious clients.
 		if firstRequest {
+			// probe happens before envoy sends first xDS request
+			if req.TypeUrl == v3.HealthInfoType {
+				continue
```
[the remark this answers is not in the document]
hzxuzhonghu: Warn sgtm


>We also might want to allow it if the Node info is there for some reason
We are quite sure it is not allowed healthcheck to be the first request.

pkg/istio-agent/xds_proxy.go
```diff
@@ -337,6 +337,10 @@ func (p *XdsProxy) handleStream(downstream adsStream) error {
 				}
 				return
 			}
+			if req.TypeUrl == v3.HealthInfoType && !initialRequestsSent {
```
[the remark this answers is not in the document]
hzxuzhonghu: Right, it is weird to have two separate threads sending xds requests. I can merge them

hzxuzhonghu:
@howardjohn another look

pilot/pkg/xds/ads.go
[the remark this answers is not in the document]
howardjohn: Seems valid to me but I think its a very very niche case and the client code doesn't support it anyways.

pkg/istio-agent/xds_proxy.go
```diff
@@ -449,10 +409,55 @@ func (p *XdsProxy) HandleUpstream(ctx context.Context, con *ProxyConnection, xds
 }
 
 func (p *XdsProxy) handleUpstreamRequest(con *ProxyConnection) {
+	// Handle downstream xds
```
howardjohn, line 412: Having a function named `handleUpstreamRequest` that does `Handle downstream xds` seems misleading?

pkg/istio-agent/xds_proxy.go
```diff
@@ -449,10 +409,55 @@ func (p *XdsProxy) HandleUpstream(ctx context.Context, con *ProxyConnection, xds
...
+			}
+		}
+	}()
+
 	defer con.upstream.CloseSend() // nolint
 	for {
 		select {
 		case req := <-con.requestsChan:
+			if req.TypeUrl == v3.HealthInfoType && !initialRequestsSent.Load() {
+				// only send healthcheck probe after LDS request has been sent
```
howardjohn, line 458: can we add a test to make sure it is actually correctly sent after LDS? Worried we may drop the update

pkg/istio-agent/xds_proxy.go
```diff
@@ -449,10 +409,55 @@ func (p *XdsProxy) HandleUpstream(ctx context.Context, con *ProxyConnection, xds
...
+					})
+				}
+				// Fire of a configured initial request, if there is one
+				p.connectedMutex.RLock()
+				initialRequest := p.initialRequest
+				if initialRequest != nil {
+					con.sendRequest(initialRequest)
+				}
+				p.connectedMutex.RUnlock()
+				initialRequestsSent.Store(true)
```
howardjohn, line 448: I think there is a race condition that will lead to droppped requests:

```
a) con.SendRequest
a) load p.initialRequest, it is nil. Unlock connectedMutex
b) PersistRequest, store initialRequest and send on req chan
c) read from requestChan. initialRequestsSent=false, do nothing
a) `initialRequestsSent.Store(true)`
```

Letter == the goroutine.

I think we can just move this Store into the connectedMutex lock?

pkg/istio-agent/xds_proxy.go
hzxuzhonghu: maybe should recomment it

pkg/istio-agent/xds_proxy.go
hzxuzhonghu: good catch, i think move the initialRequestsSent.Store(true) before the `con.sendRequest(initialRequest)` should be ok.

pkg/istio-agent/xds_proxy.go
hzxuzhonghu: ok, will add to it

hzxuzhonghu: done

howardjohn (approved):
ty!

PR #35882, opened 2021-11-03
docker: rewrite make+bash with a simple golang tool

howardjohn:
This rewrites the amalgamation of make+bash that we use to build docker files today into a fairly simple go program. The bash and make combination is horrible to use - it has a bash HCL generation, and some extreme make complexity (example: `${ISTIO_DOCKER_TAR}/$(subst docker.,,$(TGT))$(call variant-tag,$(VARIANT)).tar $(HUB)/$(subst docker.,,$(TGT)):$(subst -$(DEFAULT_DISTRIBUTION),,$(TAG)-$(VARIANT)) && \`). This has caused a few issues:
* Code is hard to understand
* Code is even harder to change. I tried to make a simple change and it took all day, and popped up many new bugs. When we fix one bug, a new one pops up. This is directly related to the complexity of the code.
* Code is not optimal because its too painful to do the right thing. This leads to people like me having tons of crazy aliases to optimize workflows, while others are left with poor UX.

This re-implementation (attempts) to fix these. Along with simply rewriting in a more readable manner, there are a few other benefits:
* More efficient: automatically determine how to invoke `go build` for optimal performance
* More efficient: for --save mode, emit to tar file in buildx instead of double invocation+copy
* More efficient: for --save mode, only do it once for -debug and default
* Easier to understand. There are 3 variants, debug default and distroless. Do not need to manage INCLUDE_UNTAGGED_DEFAULT as it happens automatically.

In the initial iteration, this simply wraps existing docker commands like `make docker`, `make dockerx`, `make docker.push` to use the new tool. An env var DOCKER_V2_BUILDER is added as an opt out in case this breaks something (which is why the PR has no code removed).

In the future, we can consider the long term plan. Part of the goal of this was to provide a better CLI experience for developers. For example, instead of trying to remember the right make command, I can run things like `tools/docker --push --targets pilot`. Long term, we will need to consider if we start using the tool directly, both in our own usage and in developer recommendations, and sunset the old `make` commands. However, we can decide that after merging this.

### DIFF
--- a/tools/docker-builder/main.go
+++ b/tools/docker-builder/main.go
@@ -1,16 +1,49 @@
 package main
 
 import (
+	"encoding/json"
+	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
 	"k8s.io/utils/env"
 
-	"istio.io/pkg/log"
+	"istio.io/istio/pilot/pkg/util/sets"
 	testenv "istio.io/istio/pkg/test/env"
+	"istio.io/pkg/log"
 )
+
+type Group struct {
+	Targets []string `json:"targets" hcl:"targets"`
+}
+
+type BakeFile struct {
+	Target map[string]Target `json:"target,omitempty"`
+	Group  map[string]Group `json:"group,omitempty"`
+}
+
+type Target struct {
+	Context          *string           `json:"context,omitempty" hcl:"context,optional"`
+	Dockerfile       *string           `json:"dockerfile,omitempty" hcl:"dockerfile,optional"`
+	DockerfileInline *string           `json:"dockerfile-inline,omitempty" hcl:"dockerfile-inline,optional"`
+	Args             map[string]string `json:"args,omitempty" hcl:"args,optional"`
+	Labels           map[string]string `json:"labels,omitempty" hcl:"labels,optional"`
+	Tags             []string          `json:"tags,omitempty" hcl:"tags,optional"`
+	CacheFrom        []string          `json:"cache-from,omitempty"  hcl:"cache-from,optional"`
+	CacheTo          []string          `json:"cache-to,omitempty"  hcl:"cache-to,optional"`
+	Target           *string           `json:"target,omitempty" hcl:"target,optional"`
+	Secrets          []string          `json:"secret,omitempty" hcl:"secret,optional"`
+	SSH              []string          `json:"ssh,omitempty" hcl:"ssh,optional"`
+	Platforms        []string          `json:"platforms,omitempty" hcl:"platforms,optional"`
+	Outputs          []string          `json:"output,omitempty" hcl:"output,optional"`
+	Pull             *bool             `json:"pull,omitempty" hcl:"pull,optional"`
+	NoCache          *bool             `json:"no-cache,omitempty" hcl:"no-cache,optional"`
+
+	// IMPORTANT: if you add more fields here, do not forget to update newOverrides and README.
+}
 
 // docker-builder
 // * --push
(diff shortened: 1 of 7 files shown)


ericvn:
Just trying a `make docker` on my Mac:
```
./tools/docker-copy.sh /work/out/linux_amd64/operator manifests operator/docker/Dockerfile.operator /work/out/linux_amd64/dockerx_build/docker.operator
make[2]: *** No rule to make target 'docker.install-ci'.  Stop.
Error: exit status 2
```
Investigating that and other targets.

howardjohn:
`install-ci` is just a typo for install-cni. not sure how that passed tests... I think they explicitly set the targets maybe

tools/docker-builder/main.go
```diff
@@ -0,0 +1,332 @@
...
+func createBuildxBuilderIfNeeded(a Args) error {
+	if !a.Save {
+		return nil
+	}
+	if _, f := os.LookupEnv("CI"); !f {
+		// For now only do this for CI, we do not want to mess with users config. And users rarely use --save
+		return nil
+	}
+	return exec.Command("sh", "-c", `
+export DOCKER_CLI_EXPERIMENTAL=enabled
```
mandarjog, line 112: Doesn’t docker tool have a go client?

douglas-reid (approved):
the changes in the E&T owned files look ok to me.

tools/docker-builder/main.go
howardjohn: Yes, but I didn't really want to introduce a new dependency.

We could also use a bunch of exec.Cmd calls for this but its a lot simple to just do this :-)

ericvn:
I know running on a Mac will cause issues. Next issue:
```
2021-11-04T15:20:11.525897Z	info	Running command: docker buildx bake -f /work/out/darwin_amd64/dockerx_build/docker-bake.json all
[+] Building 1.7s (1/1) FINISHED
 => [internal] booting buildkit                                                                                                                                                                                                         1.7s
 => => pulling image gcr.io/istio-testing/buildkit:buildx-stable-1                                                                                                                                                                      1.2s
 => => creating container buildx_buildkit_container-builder0                                                                                                                                                                            0.5s
error: unable to prepare context: path "/work/out/darwin_amd64/dockerx_build/docker.app_sidecar_ubuntu_xenial" not found
```
Need to figure out if the docker-bake.json should be under `darwin` or `linux`, noting it does exist under `darwin` right now.

howardjohn:
> > Need to figure out if the docker-bake.json should be under `darwin` or `linux`, noting it does exist under `darwin` right now.
> 
> Making an assumption it was `linux`, the following changes got `make docker` to complete locally:
> 
> ```
> --- a/tools/docker-builder/main.go
> +++ b/tools/docker-builder/main.go
> @@ -89,8 +89,8 @@ var rootCmd = &cobra.Command{
>  }
> 
>  func RunBake(args Args) error {
> -       out := filepath.Join(testenv.IstioOut, "dockerx_build", "docker-bake.json")
> -       _ = os.MkdirAll(filepath.Join(testenv.IstioOut, "release", "docker"), 0o755)
> +       out := filepath.Join(testenv.LocalOut, "dockerx_build", "docker-bake.json")
> +       _ = os.MkdirAll(filepath.Join(testenv.LocalOut, "release", "docker"), 0o755)
>         if err := createBuildxBuilderIfNeeded(args); err != nil {
>                 return err
>         }
> @@ -205,7 +205,7 @@ func ConstructBakeFile(a Args) (map[string]string, error) {
>                         if strings.HasPrefix(target, "app_") && variant == DistrolessVariant {
>                                 continue
>                         }
> -                       p := filepath.Join(testenv.IstioOut, "dockerx_build", fmt.Sprintf("docker.%s", target))
> +                       p := filepath.Join(testenv.LocalOut, "dockerx_build", fmt.Sprintf("docker.%s", target))
>                         t := Target{
>                                 Context:    sp(p),
>                                 Dockerfile: sp(fmt.Sprintf("Dockerfile.%s", target)),
> @@ -267,12 +267,12 @@ func ConstructBakeFile(a Args) (map[string]string, error) {
>                 Target: targets,
>                 Group:  groups,
>         }
> -       out := filepath.Join(testenv.IstioOut, "dockerx_build", "docker-bake.json")
> +       out := filepath.Join(testenv.LocalOut, "dockerx_build", "docker-bake.json")
>         j, err := json.MarshalIndent(bf, "", "  ")
>         if err != nil {
>                 return nil, err
>         }
> -       _ = os.MkdirAll(filepath.Join(testenv.IstioOut, "dockerx_build"), 0o755)
> +       _ = os.MkdirAll(filepath.Join(testenv.LocalOut, "dockerx_build"), 0o755)
>         return tarFiles, os.WriteFile(out, j, 0o644)
>  }
> ```

Nice! so its fully working with that?

ericvn:
I did verify that without this change, the files are under out/linux.

ericvn:
Verified `make docker(x)` and `make dockerx.save` targets worked on a make (just verifying images/tars are built).

ericvn (approved):

PR #35883, opened 2021-11-03
enable multicluster headless by default

stevenctl:
We've had this integration tested for some time, and it has been tried by [some users](https://istio.slack.com/archives/C01AT2S8B7B/p1630861283017700).

The only concern is users who relied on these being cluster-local, the release note partially covers that. Some of the docs for multi-cluster in 1.12 will show better mechanisms to do cluster-local settings on a per-service basis.

### DIFF
--- a/releasenotes/notes/35883.yaml
+++ b/releasenotes/notes/35883.yaml
@@ -1,5 +1,6 @@
 area: traffic-management
 releaseNotes:
 - |
-  **Enabled** discovering cross-cluster, same network, endpoints for headless services. This can be disabled by setting
+  **Enabled** discovering cross-cluster, same network, endpoints for headless services when `ISTIO_META_DNS_CAPTURE` and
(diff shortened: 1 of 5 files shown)


releasenotes/notes/35883.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+releaseNotes:
+- |
+  **Enabled** discovering cross-cluster, same network, endpoints for headless services. This can be disabled by setting
```
howardjohn, line 6: nit: may mention this is tied to the (off by default) DNS settings.

howardjohn (approved):

releasenotes/notes/35883.yaml
stevenctl: right, thanks!

PR #35884, opened 2021-11-03
enable multicluster workload entry by default

stevenctl:
This has been enabled since January in integration tests, it should be proven stable.


stevenctl:
cc @istio/release-managers-1-12

PR #35893, opened 2021-11-03
app=helloworld is right label

AbdullahBM:

### DIFF
--- a/samples/helloworld/README.md
+++ b/samples/helloworld/README.md
@@ -24,7 +24,7 @@ kubectl apply -f helloworld.yaml
 Alternatively, you can run just one version at a time by first defining the service:
 
 ```bash
-kubectl apply -f helloworld.yaml -l service=helloworld
+kubectl apply -f helloworld.yaml -l app=helloworld
 ```
 
 and then deploying version v1, v2, or both:


AbdullahBM:
> @googlebot I signed it!

samples/helloworld/README.md
```diff
@@ -24,7 +24,7 @@ kubectl apply -f helloworld.yaml
 Alternatively, you can run just one version at a time by first defining the service:
 
 ```bash
-kubectl apply -f helloworld.yaml -l service=helloworld
+kubectl apply -f helloworld.yaml -l app=helloworld
```
kebe7jun, line 27: Here it is apply service first, only service has `service: helloworld` tag in yaml, so it should be reasonable.
https://github.com/istio/istio/blob/a48d843bd3302ede4a9dadd491742b562f3f376f/samples/helloworld/helloworld.yaml#L7

kebe7jun (changes requested):

samples/helloworld/README.md
frankbu: @kebe7jun is right. This command is trying to apply **only** the service, so it is correctly using the `service=helloworld` label. Using the `app=helloworld` label would apply the service and the deployments as well, which is not what is intended.

PR #35895, opened 2021-11-04
support root cert specified in mesh config

irisdingbj:
- do not create istio ca if `PILOT-CERT-PROVIDER` is `k8s.io/**` and `EXTERNAL_CA` is `ISTIOD_RA_KUBERNETES_API`
- Use root certs specified in meshconfig for workloads and istiod
- validate the root cert for signed cert from cert-chain against root cert specified in meshconfig

### DIFF
--- a/pilot/pkg/bootstrap/certcontroller.go
+++ b/pilot/pkg/bootstrap/certcontroller.go
@@ -110,7 +110,32 @@ func (s *Server) initDNSCerts(hostname, namespace string) error {
 
 	var certChain, keyPEM, caBundle []byte
 	var err error
-	if features.PilotCertProvider == constants.CertProviderKubernetes {
+	pilotCertProviderName := features.PilotCertProvider
+	if strings.HasPrefix(pilotCertProviderName, constants.CertProviderKubernetesSignerPrefix) && s.RA != nil {
+		signerName := strings.TrimPrefix(pilotCertProviderName, constants.CertProviderKubernetesSignerPrefix)
+		log.Infof("Generating K8S-signed cert for %v using signer %v", s.dnsNames, signerName)
+		certChain, keyPEM, _, err = chiron.GenKeyCertK8sCA(s.kubeClient,
+			strings.Join(s.dnsNames, ","), hostnamePrefix+".csr.secret", namespace, "", signerName, true, SelfSignedCACertTTL.Get())
+		if err != nil {
+			return fmt.Errorf("failed generating key and cert by kubernetes: %v", err)
+		}
+		caBundle, err = s.RA.GetRootCertFromMeshConfig(signerName)
+		if err != nil {
+			return err
+		}
+		// MeshConfig:Add callback for mesh config update
+		s.environment.AddMeshHandler(func() {
+			newCaBundle, _ := s.RA.GetRootCertFromMeshConfig(signerName)
+			if newCaBundle != nil && string(newCaBundle) != string(s.istiodCertBundleWatcher.GetKeyCertBundle().CABundle) {
+				newCertChain, newKeyPEM, _, err := chiron.GenKeyCertK8sCA(s.kubeClient,
+					strings.Join(s.dnsNames, ","), hostnamePrefix+".csr.secret", namespace, "", signerName, true, SelfSignedCACertTTL.Get())
+				if err != nil {
+					log.Fatalf("failed regenerating key and cert for istiod by kubernetes: %v", err)
+				}
+				s.istiodCertBundleWatcher.SetAndNotify(newKeyPEM, newCertChain, newCaBundle)
+			}
+		})
+	} else if pilotCertProviderName == constants.CertProviderKubernetes {
 		log.Infof("Generating K8S-signed cert for %v", s.dnsNames)
 		certChain, keyPEM, _, err = chiron.GenKeyCertK8sCA(s.kubeClient,
 			strings.Join(s.dnsNames, ","), hostnamePrefix+".csr.secret", namespace, defaultCACertPath, "", true, SelfSignedCACertTTL.Get())
@@ -121,7 +146,7 @@ func (s *Server) initDNSCerts(hostname, namespace string) error {
 		if err != nil {
 			return fmt.Errorf("failed reading %s: %v", defaultCACertPath, err)
 		}
-	} else if features.PilotCertProvider == constants.CertProviderIstiod {
+	} else if pilotCertProviderName == constants.CertProviderIstiod {
 		certChain, keyPEM, err = s.CA.GenKeyCert(s.dnsNames, SelfSignedCACertTTL.Get(), false)
 		if err != nil {
 			return fmt.Errorf("failed generating istiod key cert %v", err)
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -1259,8 +1259,17 @@ func (s *Server) initWorkloadTrustBundle(args *PilotArgs) error {
 // It return true only if istiod certs is signed by Kubernetes and
 // workload certs are signed by external CA
 func (s *Server) isDisableCa() bool {
-	return (features.PilotCertProvider == constants.CertProviderKubernetes ||
-		strings.HasPrefix(features.PilotCertProvider, constants.CertProviderKubernetesSignerPrefix)) && s.RA != nil
+	if s.RA != nil {
+		// do not create CA server if PilotCertProvider is `kubernetes` and RA server exists
+		if features.PilotCertProvider == constants.CertProviderKubernetes {
+			return true
+		}
+		// do not create CA server if PilotCertProvider is `k8s.io/*` and RA server exists
+		if strings.HasPrefix(features.PilotCertProvider, constants.CertProviderKubernetesSignerPrefix) {
+			return true
+		}
+	}
+	return false
 }
 
 func (s *Server) initStatusManager(_ *PilotArgs) {
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -989,6 +989,12 @@ func (s *Server) initIstiodCerts(args *PilotArgs, host string) error {
 		if err == nil {
 			err = s.initIstiodCertLoader()
 		}
+	} else if strings.HasPrefix(features.PilotCertProvider, constants.CertProviderKubernetesSignerPrefix) {
+		log.Infof("initializing Istiod DNS certificates host: %s, custom host: %s", host, features.IstiodServiceCustomHost)
+		err = s.initDNSCerts(host, args.Namespace)
+		if err == nil {
+			err = s.initIstiodCertLoader()
+		}
 	}
 
 	return err
@@ -1009,7 +1015,13 @@ func (s *Server) createPeerCertVerifier(tlsOptions TLSOptions) (*spiffe.PeerCert
 		}
 	} else {
 		if s.RA != nil {
-			rootCertBytes = append(rootCertBytes, s.RA.GetCAKeyCertBundle().GetRootCertPem()...)
+			if strings.HasPrefix(features.PilotCertProvider, constants.CertProviderKubernetesSignerPrefix) {
+				signerName := strings.TrimPrefix(features.PilotCertProvider, constants.CertProviderKubernetesSignerPrefix)
+				caBundle, _ := s.RA.GetRootCertFromMeshConfig(signerName)
+				rootCertBytes = append(rootCertBytes, caBundle...)
+			} else {
+				rootCertBytes = append(rootCertBytes, s.RA.GetCAKeyCertBundle().GetRootCertPem()...)
+			}
 		}
 		if s.CA != nil {
 			rootCertBytes = append(rootCertBytes, s.CA.GetCAKeyCertBundle().GetRootCertPem()...)
@@ -1247,7 +1259,8 @@ func (s *Server) initWorkloadTrustBundle(args *PilotArgs) error {
 // It return true only if istiod certs is signed by Kubernetes and
 // workload certs are signed by external CA
 func (s *Server) isDisableCa() bool {
-	return features.PilotCertProvider == constants.CertProviderKubernetes && s.RA != nil
+	return (features.PilotCertProvider == constants.CertProviderKubernetes ||
+		strings.HasPrefix(features.PilotCertProvider, constants.CertProviderKubernetesSignerPrefix)) && s.RA != nil
 }
 
 func (s *Server) initStatusManager(_ *PilotArgs) {
--- a/security/pkg/pki/ca/ca.go
+++ b/security/pkg/pki/ca/ca.go
@@ -320,8 +320,8 @@ func (ca *IstioCA) Sign(csrPEM []byte, certOpts CertOpts) (
 
 // SignWithCertChain is similar to Sign but returns the leaf cert and the entire cert chain.
 func (ca *IstioCA) SignWithCertChain(csrPEM []byte, certOpts CertOpts) (
-	[]byte, error) {
-	return ca.signWithCertChain(csrPEM, certOpts.SubjectIDs, certOpts.TTL, true, certOpts.ForCA)
+	[]string, error) {
+	return ca.getSignedCertChain(csrPEM, certOpts.SubjectIDs, certOpts.TTL, true, certOpts.ForCA)
 }
 
 // GetCAKeyCertBundle returns the KeyCertBundle for the CA.
@@ -414,12 +414,31 @@ func (ca *IstioCA) sign(csrPEM []byte, subjectIDs []string, requestedLifetime ti
 	return cert, nil
 }
 
+func (ca *IstioCA) getSignedCertChain(csrPEM []byte, subjectIDs []string, requestedLifetime time.Duration, lifetimeCheck,
+	forCA bool) ([]string, error) {
+	cert, err := ca.sign(csrPEM, subjectIDs, requestedLifetime, lifetimeCheck, forCA)
+	if err != nil {
+		return nil, err
+	}
+	respCertChain := []string{string(cert)}
+	chainPem := ca.GetCAKeyCertBundle().GetCertChainPem()
+	if len(chainPem) > 0 {
+		respCertChain = append(respCertChain, string(chainPem))
+	}
+	_, _, _, rootCertBytes := ca.GetCAKeyCertBundle().GetAll()
+	if len(rootCertBytes) != 0 {
+		respCertChain = append(respCertChain, string(rootCertBytes))
+	}
+	return respCertChain, nil
+}
+
 func (ca *IstioCA) signWithCertChain(csrPEM []byte, subjectIDs []string, requestedLifetime time.Duration, lifetimeCheck,
 	forCA bool) ([]byte, error) {
 	cert, err := ca.sign(csrPEM, subjectIDs, requestedLifetime, lifetimeCheck, forCA)
 	if err != nil {
 		return nil, err
 	}
+
 	chainPem := ca.GetCAKeyCertBundle().GetCertChainPem()
 	if len(chainPem) > 0 {
 		cert = append(cert, chainPem...)
--- a/security/pkg/pki/ca/ca.go
+++ b/security/pkg/pki/ca/ca.go
@@ -321,7 +321,11 @@ func (ca *IstioCA) Sign(csrPEM []byte, certOpts CertOpts) (
 // SignWithCertChain is similar to Sign but returns the leaf cert and the entire cert chain.
 func (ca *IstioCA) SignWithCertChain(csrPEM []byte, certOpts CertOpts) (
 	[]string, error) {
-	return ca.getSignedCertChain(csrPEM, certOpts.SubjectIDs, certOpts.TTL, true, certOpts.ForCA)
+	cert, err := ca.signWithCertChain(csrPEM, certOpts.SubjectIDs, certOpts.TTL, true, certOpts.ForCA)
+	if err != nil {
+		return nil, err
+	}
+	return []string{string(cert)}, nil
 }
 
 // GetCAKeyCertBundle returns the KeyCertBundle for the CA.
@@ -412,24 +416,6 @@ func (ca *IstioCA) sign(csrPEM []byte, subjectIDs []string, requestedLifetime ti
 	cert := pem.EncodeToMemory(block)
 
 	return cert, nil
-}
-
-func (ca *IstioCA) getSignedCertChain(csrPEM []byte, subjectIDs []string, requestedLifetime time.Duration, lifetimeCheck,
-	forCA bool) ([]string, error) {
-	cert, err := ca.sign(csrPEM, subjectIDs, requestedLifetime, lifetimeCheck, forCA)
-	if err != nil {
-		return nil, err
-	}
-	respCertChain := []string{string(cert)}
-	chainPem := ca.GetCAKeyCertBundle().GetCertChainPem()
-	if len(chainPem) > 0 {
-		respCertChain = append(respCertChain, string(chainPem))
-	}
-	_, _, _, rootCertBytes := ca.GetCAKeyCertBundle().GetAll()
-	if len(rootCertBytes) != 0 {
-		respCertChain = append(respCertChain, string(rootCertBytes))
-	}
-	return respCertChain, nil
 }
 
 func (ca *IstioCA) signWithCertChain(csrPEM []byte, subjectIDs []string, requestedLifetime time.Duration, lifetimeCheck,
--- a/security/pkg/pki/ra/k8s_ra.go
+++ b/security/pkg/pki/ra/k8s_ra.go
@@ -16,11 +16,14 @@ package ra
 
 import (
 	"fmt"
+	"strings"
+	"sync"
 	"time"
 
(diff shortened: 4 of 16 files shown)


pilot/pkg/bootstrap/certcontroller.go
```diff
@@ -279,3 +305,27 @@ func (s *Server) loadIstiodCert() error {
...
+	if len(caCertificates) == 0 {
+		return nil, fmt.Errorf("no caCertificates defined in mesh config")
+	}
+	for _, caCertificate := range caCertificates {
+		signers := caCertificate.CertSigners
+		if len(signers) == 0 {
+			return nil, fmt.Errorf("no cert signers defined in mesh config")
+		}
+		sort.Strings(signers)
+		index := sort.SearchStrings(signers, signerName)
```
howardjohn, line 321: Its faster + simpler to just iterate over the list and find it

pilot/pkg/bootstrap/server.go
```diff
@@ -1011,7 +1011,11 @@ func (s *Server) createPeerCertVerifier(tlsOptions TLSOptions) (*spiffe.PeerCert
 		}
 	} else {
 		if s.RA != nil {
-			rootCertBytes = append(rootCertBytes, s.RA.GetCAKeyCertBundle().GetRootCertPem()...)
+			if strings.HasPrefix(features.PilotCertProvider, constants.CertProviderKubernetesSignerPrefix) {
```
howardjohn, line 1014: if this is "RA" then why aren't we using s.RA here?

pilot/pkg/bootstrap/server.go
```diff
@@ -1249,5 +1269,6 @@ func (s *Server) initWorkloadTrustBundle(args *PilotArgs) error {
 // It return true only if istiod certs is signed by Kubernetes and
 // workload certs are signed by external CA
 func (s *Server) isDisableCa() bool {
-	return features.PilotCertProvider == constants.CertProviderKubernetes && s.RA != nil
+	return features.PilotCertProvider == constants.CertProviderKubernetes ||
```
howardjohn, line 1272: can you put parenthesis or split it up? its confusing to have `a || b && c` since you need to remember precedence rules

pilot/pkg/bootstrap/server.go
irisdingbj: There are two cases for root certs when s.RA is not nil:
1.  Using root certs in meshconfig 
1.  Mount `external-ca-cert` secret and use it as the RA's root certs  -- previous supported https://istio.io/latest/docs/tasks/security/cert-management/custom-ca-k8s/#part-2-using-custom-ca

pilot/pkg/bootstrap/certcontroller.go
irisdingbj: Thanks! Fixed

pilot/pkg/bootstrap/server.go
irisdingbj: Sure, fixed

irisdingbj:
@shankgan @howardjohn PTAL, Thanks!

pilot/pkg/bootstrap/certcontroller.go
```diff
@@ -111,7 +111,32 @@ func (s *Server) initDNSCerts(hostname, namespace string) error {
 
 	var certChain, keyPEM, caBundle []byte
 	var err error
-	if features.PilotCertProvider == constants.CertProviderKubernetes {
+	pilotCertProviderName := features.PilotCertProvider
+	if strings.HasPrefix(pilotCertProviderName, constants.CertProviderKubernetesSignerPrefix) {
```
shankgan, line 115: Can we make this a function saying either "Kubernetes" or "k8s.io"?

pilot/pkg/bootstrap/server.go
shankgan: Not sure if this is the best logic: the bundle here is supposed to represent the mesh cabundle. The "istiodCertBundleWatcher" is only meant to watch the caBundle corresponding to the Istiod cert, which may be different from the RA's cabundle. There are use cases where we could use Istiod as an RA to sign workload certificates, but use self-signed Istiod certificates.

pilot/pkg/bootstrap/server.go
```diff
@@ -1101,11 +1105,27 @@ func (s *Server) maybeCreateCA(caOpts *caOptions) error {
...
 			if s.CA, err = s.createIstioCA(corev1, caOpts); err != nil {
 				return fmt.Errorf("failed to create CA: %v", err)
 			}
+			s.environment.AddMeshHandler(func() {
+				meshConfig := s.environment.Mesh()
+				caCertificates := meshConfig.CaCertificates
+				if len(caCertificates) == 0 {
+					return
+				}
+				s.CA.SetCACertificatesFromMeshConfig(caCertificates)
```
shankgan, line 1127: Given that this logic is quite specific to how Istiod functions as an RA, we should keep the logic out of the CA code as much as possible IMO

pilot/pkg/trustbundle/trustbundle.go
```diff
@@ -212,11 +212,14 @@ func (tb *TrustBundle) AddMeshConfigUpdate(cfg *meshconfig.MeshConfig) error {
 		certs := []string{}
 		endpoints := []string{}
 		for _, pemCert := range cfg.GetCaCertificates() {
-			cert := pemCert.GetPem()
-			if cert != "" {
-				certs = append(certs, cert)
-			} else if pemCert.GetSpiffeBundleUrl() != "" {
-				endpoints = append(endpoints, pemCert.GetSpiffeBundleUrl())
+			// if there is cert signer or trust domains defined for the cert, we should not use it as global trust anchor
+			if pemCert.CertSigners == nil && pemCert.TrustDomains == nil {
```
shankgan, line 216: Will this logic change sometime in the future? Is there a strong requirement that workloads whose certis signed by CertSigner 1 should not talk to workloads whose trustAnchor is signed by CertSigner 2?

security/pkg/server/ca/server.go
```diff
@@ -94,24 +100,58 @@ func (s *Server) CreateCertificate(ctx context.Context, request *pb.IstioCertifi
...
+	var rootCert, rootCertFromMeshConfig []byte
 	if len(rootCertBytes) == 0 {
-		rootCert, err := util.FindRootCertFromCertificateChainBytes(cert)
+		rootCertFromCertChain, err := util.FindRootCertFromCertificateChainBytes(cert)
 		if err != nil {
 			serverCaLog.Errorf("failed to find root cert from signed cert-chain (%v)", err.Error())
 			s.monitoring.GetCertSignError(err.(*caerror.Error).ErrorType()).Increment()
 			return nil, status.Errorf(err.(*caerror.Error).HTTPErrorCode(), "failed to find root cert from signed cert-chain (%v)", err.(*caerror.Error))
 		}
+		caCertificates := s.ca.GetCACertificatesFromMeshConfig()
```
shankgan, line 116: nit: I wonder if there is a way to move all this logic into the RA code as part of the ca.sign() call? The existing logic in this function -> (a) extract the external signer that is supposed to sign the cert (b) find the correct CA cert that maps to the signer (c) verify that the cert is indeed signed by the CA Cert - this logic is highly specific to the RA code and makes the existing Istiod CA code confusing to read IMHO. Perhaps ca and ra servers could both implement a signWithCertChain that could return the entire cert-chain as a string?

pilot/pkg/bootstrap/server.go
irisdingbj: Yes, so this is the reason why we add a condition here  `if strings.HasPrefix(features.PilotCertProvider, constants.CertProviderKubernetesSignerPrefix)`, so when istiod is acting as RA and istiod's certificate is signed by k8s signer we will use `istiodCertBundleWatcher` . The `istiodCertBundleWatcher` is set in https://github.com/istio/istio/blob/f26bfc01f2ce4254b0988e48980dbb9809be4397/pilot/pkg/bootstrap/certcontroller.go#L137 in this case to use the one from meshconfig.

pilot/pkg/trustbundle/trustbundle.go
irisdingbj: Make sense, Let me remove this condition check.

pilot/pkg/bootstrap/server.go
irisdingbj: Thanks! This should not be needed

security/pkg/server/ca/server.go
irisdingbj: Yes, Thanks for the suggestion, Let me refactor this to scoped this to RA code base.

pilot/pkg/bootstrap/server.go
irisdingbj: remvoed

pilot/pkg/trustbundle/trustbundle.go
irisdingbj: removed

security/pkg/server/ca/server.go
irisdingbj: moved to SignWithCertChain in RA code

irisdingbj:
@shankgan PTAL, Thanks!

irisdingbj:
Hi @Shankar Ganesan I submitted the changes to address your review comments, PTAL, Thanks!

pilot/pkg/bootstrap/server.go
shankgan: can we just pick up the certificates from the ra only with the latest set of changes?

security/pkg/server/ca/server.go
```diff
@@ -39,18 +40,23 @@ type CertificateAuthority interface {
 	// Sign generates a certificate for a workload or CA, from the given CSR and cert opts.
 	Sign(csrPEM []byte, opts ca.CertOpts) ([]byte, error)
 	// SignWithCertChain is similar to Sign but returns the leaf cert and the entire cert chain.
-	SignWithCertChain(csrPEM []byte, opts ca.CertOpts) ([]byte, error)
-	// GetCAKeyCertBundle returns the KeyCertBundle used by CA.
+	SignWithCertChain(csrPEM []byte, opts ca.CertOpts) ([]string, error)
+	// SignWithCertChain is similar to Sign but returns the leaf cert and the entire cert chain.
 	GetCAKeyCertBundle() *util.KeyCertBundle
+	// GetCACertificatesFromMeshConfig returns the CACertificates in mesh config.
+	GetCACertificatesFromMeshConfig() []*meshconfig.MeshConfig_CertificateData
```
shankgan, line 47: trying to understand why do these need to be exposed by the interface? this is functionality that is specific to the RA. I don't see this being used anywhere in the ca path

pilot/pkg/bootstrap/server.go
```diff
@@ -1097,6 +1107,14 @@ func (s *Server) maybeCreateCA(caOpts *caOptions) error {
 			if s.RA, err = s.createIstioRA(s.kubeClient, caOpts); err != nil {
 				return fmt.Errorf("failed to create RA: %v", err)
 			}
+			s.environment.AddMeshHandler(func() {
```
shankgan, line 1110: IMO, this can be done inside the createIstioRA function. This would also require the "SetCACertificatesFromMeshConfig" method to implemented by the RA code and not by the the ca interface

security/pkg/pki/ca/ca.go
```diff
@@ -415,14 +417,37 @@ func (ca *IstioCA) sign(csrPEM []byte, subjectIDs []string, requestedLifetime ti
 }
 
 func (ca *IstioCA) signWithCertChain(csrPEM []byte, subjectIDs []string, requestedLifetime time.Duration, lifetimeCheck,
```
shankgan, line 419: nit: IMO, It might be better idea to implement and use a new function with new functionality rather than change the functionality of an older function. Perhaps the new function could be called "getSignedCertChain" which returns the string slice

security/pkg/pki/ca/ca.go
```diff
@@ -415,14 +417,37 @@ func (ca *IstioCA) sign(csrPEM []byte, subjectIDs []string, requestedLifetime ti
...
 	}
+
 	chainPem := ca.GetCAKeyCertBundle().GetCertChainPem()
 	if len(chainPem) > 0 {
 		cert = append(cert, chainPem...)
 	}
 	return cert, nil
 }
+
+func (ca *IstioCA) GetCACertificatesFromMeshConfig() []*meshconfig.MeshConfig_CertificateData {
```
shankgan, line 447: See earlier comment - don't think there is a need to implement this function in the struct, nor alter the existing interface definition to include it

pilot/pkg/bootstrap/server.go
shankgan: In the Mesh handler, I think it might be safer to copy the certificates onto a newer array and use that to "SetCACertificatesFromMeshConfig".

shankgan:
Would be good to add some changes to k8s_ra_test to test the RA functionality changes. For instance  - Verify that when we change the meshConfig.CaCerts, the new signed certificate  has the new cacert and the older signed certificate verification fails (since the cert corresponding to the signer has changed)

pilot/pkg/bootstrap/server.go
irisdingbj: changed to getRootCertFromRA

security/pkg/server/ca/server.go
irisdingbj: removed from ca path

pilot/pkg/bootstrap/server.go
irisdingbj: This is done inside RA only

security/pkg/pki/ca/ca.go
irisdingbj: fixed

security/pkg/pki/ca/ca.go
irisdingbj: yep, Thanks, removed from ca path

irisdingbj:
@shankgan for the valuable comments! I have submitted a new verion and a new UT has been added into k8s_ra_test.go to cover the case you suggested.  PTAL, Thanks!

security/pkg/server/ca/server.go
```diff
@@ -81,38 +81,22 @@ func (s *Server) CreateCertificate(ctx context.Context, request *pb.IstioCertifi
...
-			s.monitoring.GetCertSignError(err.(*caerror.Error).ErrorType()).Increment()
-			return nil, status.Errorf(err.(*caerror.Error).HTTPErrorCode(), "failed to find root cert from signed cert-chain (%v)", err.(*caerror.Error))
-		}
-		if verifyErr := util.VerifyCertificate(nil, cert, rootCert, nil); verifyErr != nil {
-			serverCaLog.Errorf("root cert from signed cert-chain is invalid (%v)", err.Error())
-			s.monitoring.GetCertSignError(err.(*caerror.Error).ErrorType()).Increment()
-			return nil, status.Errorf(err.(*caerror.Error).HTTPErrorCode(), "root cert from signed cert-chain is invalid (%v)", err.(*caerror.Error))
-		}
-		rootCertBytes = rootCert
+	if len(rootCertBytes) != 0 {
```
shankgan, line 97: nit: is this still needed if the function invoked is "SignWithCertChain" as opposed to "Sign". Can this be handled inside the function

pilot/pkg/bootstrap/server.go
shankgan: Don't see the change here yet..were you planning to add it?

security/pkg/pki/ca/ca.go
```diff
@@ -414,12 +414,27 @@ func (ca *IstioCA) sign(csrPEM []byte, subjectIDs []string, requestedLifetime ti
...
+	cert, err := ca.sign(csrPEM, subjectIDs, requestedLifetime, lifetimeCheck, forCA)
+	if err != nil {
+		return nil, err
+	}
+	respCertChain := []string{string(cert)}
+	chainPem := ca.GetCAKeyCertBundle().GetCertChainPem()
+	if len(chainPem) > 0 {
+		respCertChain = append(respCertChain, string(chainPem))
+	}
+	return respCertChain, nil
```
shankgan, line 428: should we also add the rootCert here?

security/pkg/pki/ra/k8s_ra.go
```diff
@@ -94,10 +98,73 @@ func (r *KubernetesRA) SignWithCertChain(csrPEM []byte, certOpts ca.CertOpts) ([
...
+	return respCertChain, nil
 }
 
 // GetCAKeyCertBundle returns the KeyCertBundle for the CA.
 func (r *KubernetesRA) GetCAKeyCertBundle() *util.KeyCertBundle {
 	return r.keyCertBundle
 }
+
+func (r *KubernetesRA) GetCACertificatesFromMeshConfig() []*meshconfig.MeshConfig_CertificateData {
+	return r.caCertificatesFromMeshConfig
```
shankgan, line 141: Two points: I am not sure it is safe to hold a reference to meshconfig.MeshConfig_CertificateData here. Would be safer to copy over the certificates into an array here IMO. Secondly, the meshHandler code and GetCACertificatesFromMeshConfig are called from different goroutines. We need to have a mutex here IMO.

tests/integration/security/external_ca/main_test.go
```diff
@@ -91,41 +92,58 @@ func TestMain(m *testing.M) {
...
+    caCertificates:
+    - pem: |
+{{.rootcert1 | indent 8}}
+      certSigners:
+      - {{.signer1}}
+    - pem: |
+{{.rootcert2 | indent 8}}
+      certSigners:
+      - {{.signer2}}
+  components:
```
shankgan, line 121: Would be great to add a negative test case here - have the signer1's caCertificate mapped into signer2 in meshConfig and ensure that the cert Signing fails. This can be handled later.

shankgan (commented):
great progress. A few minor changes and should be able to approve this.

security/pkg/server/ca/server.go
irisdingbj: yes, fixed

security/pkg/pki/ca/ca.go
irisdingbj: yes, combined

tests/integration/security/external_ca/main_test.go
irisdingbj: Let's add this  in later PR

pilot/pkg/bootstrap/server.go
irisdingbj: done

pilot/pkg/bootstrap/server.go
irisdingbj: changed to get from RA directly now.

irisdingbj:
@shankgan PTAL, Thanks!

irisdingbj:
@shankgan A ping on this^^

pilot/pkg/bootstrap/certcontroller.go
```diff
@@ -110,7 +110,32 @@ func (s *Server) initDNSCerts(hostname, namespace string) error {
 
 	var certChain, keyPEM, caBundle []byte
 	var err error
-	if features.PilotCertProvider == constants.CertProviderKubernetes {
+	pilotCertProviderName := features.PilotCertProvider
+	if strings.HasPrefix(pilotCertProviderName, constants.CertProviderKubernetesSignerPrefix) && s.RA != nil {
```
shankgan, line 114: nit: don't think this should be related to whether the RA is enabled or not, we should be able to do this regardless

shankgan (approved):

pilot/pkg/bootstrap/certcontroller.go
irisdingbj: yep. we can do this later.

irisdingbj:
@howardjohn Can you please give an approval here if no issues?  Thanks~

security/pkg/server/ca/server.go
```diff
@@ -81,38 +81,18 @@ func (s *Server) CreateCertificate(ctx context.Context, request *pb.IstioCertifi
...
 		CertSigner: certSigner,
 	}
-	cert, signErr := s.ca.Sign([]byte(request.Csr), certOpts)
+	respCertChain, signErr := s.ca.SignWithCertChain([]byte(request.Csr), certOpts)
 	if signErr != nil {
 		serverCaLog.Errorf("CSR signing error (%v)", signErr.Error())
 		s.monitoring.GetCertSignError(signErr.(*caerror.Error).ErrorType()).Increment()
 		return nil, status.Errorf(signErr.(*caerror.Error).HTTPErrorCode(), "CSR signing error (%v)", signErr.(*caerror.Error))
 	}
-	respCertChain := []string{string(cert)}
```
howardjohn, line 97: I don't see why this whole block can be removed. We don't have the same logic in `SignWithCertChain`.

This is changing core code central to the most security critical code in all of Istio, which has been stable for ~3 years. I strongly recommend we do not change this code in order to improve an experimental feature.

If we need a new code path, please make it only for the RA stuff.

`getSignedCertChain` is not sufficient as it does not have the robust error handling and telemetry that we have here

security/pkg/server/ca/server.go
```diff
@@ -39,8 +39,8 @@ type CertificateAuthority interface {
 	// Sign generates a certificate for a workload or CA, from the given CSR and cert opts.
 	Sign(csrPEM []byte, opts ca.CertOpts) ([]byte, error)
 	// SignWithCertChain is similar to Sign but returns the leaf cert and the entire cert chain.
-	SignWithCertChain(csrPEM []byte, opts ca.CertOpts) ([]byte, error)
-	// GetCAKeyCertBundle returns the KeyCertBundle used by CA.
+	SignWithCertChain(csrPEM []byte, opts ca.CertOpts) ([]string, error)
```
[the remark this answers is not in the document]
howardjohn: Also seems we should not have `Sign` if its never used...

howardjohn (changes requested):
LGTM overall, but one critical comments

security/pkg/server/ca/server.go
shankgan: My initial suggestion was to not have different code-paths here for the RA and CA. Perhaps we could check if the certSigner was not null and based on that trigger code-paths - for the CA, we could keep the old code, for the RA, the newer code path. @irisdingbj , would this work?

security/pkg/server/ca/server.go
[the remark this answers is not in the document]
shankgan: this is mainly because the grpc security service responds with three strings, the root-cert, the cert-chain and the main cert. its difficult to contruct this from just the byte array. IF this is purely used for the RA code, I think it should be ok to change the signature?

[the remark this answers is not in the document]
howardjohn: Oh I didn't realize it was a `[]string`. Misread as `string`

security/pkg/pki/util/verify_cert.go
```diff
@@ -185,7 +185,7 @@ func FindRootCertFromCertificateChainBytes(certBytes []byte) ([]byte, error) {
 	}
 	rootBlock, _ := pem.Decode(cert)
 	if rootBlock == nil {
-		return nil, fmt.Errorf("error decoding root certificate")
+		return nil, nil
```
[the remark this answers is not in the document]
irisdingbj: In some cases especially when using external CA(e.g. cert-manager), they will not return a full cert-chain-- means no root cert will be returned. So in such case, we will find the root cert in meshconfig instead. So we should not return error here to allow both cases.

security/pkg/server/ca/server.go
irisdingbj: Yes.  This works!  Let me keep the ca code path unchanged.

pilot/pkg/bootstrap/server.go
```diff
@@ -1247,7 +1259,8 @@ func (s *Server) initWorkloadTrustBundle(args *PilotArgs) error {
 // It return true only if istiod certs is signed by Kubernetes and
 // workload certs are signed by external CA
 func (s *Server) isDisableCa() bool {
-	return features.PilotCertProvider == constants.CertProviderKubernetes && s.RA != nil
+	return (features.PilotCertProvider == constants.CertProviderKubernetes ||
```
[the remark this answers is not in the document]
irisdingbj: Fixed

security/pkg/server/ca/server.go
```diff
@@ -88,29 +87,24 @@ func (s *Server) CreateCertificate(ctx context.Context, request *pb.IstioCertifi
...
 	if signErr != nil {
 		serverCaLog.Errorf("CSR signing error (%v)", signErr.Error())
 		s.monitoring.GetCertSignError(signErr.(*caerror.Error).ErrorType()).Increment()
 		return nil, status.Errorf(signErr.(*caerror.Error).HTTPErrorCode(), "CSR signing error (%v)", signErr.(*caerror.Error))
 	}
-	respCertChain := []string{string(cert)}
-	if len(certChainBytes) != 0 {
-		respCertChain = append(respCertChain, string(certChainBytes))
-	}
-	if len(rootCertBytes) == 0 {
```
irisdingbj, line 101: this is for the multi-signer case only, so it is safe we remove the monitoring part as it is newly added as well.

irisdingbj:
@howardjohn @shankgan  PTAL, Thanks~

howardjohn (approved):

PR #35898, opened 2021-11-04
Make envoy hotRestart configurable

### LINKED ISSUE #32696: expose envoy hot restart in pilot-agent
we are using custom lua code to generate a RSA signature in envoy filter. the private key is distributed via secure mount to a VM, but because the envoy lua filter doesn't posses asynchronous capability, we need to be able to reload the filter when the file containing private key changes.
it seems the envoy's hot restart capability is the easiest way to achieve this currently, without a downtime (systemd path watcher can be then setup to trigger this)

this feature already exists in envoy, and was even used by pilot agent prior sds implementation change, so it shouldn't be a large effort to expose it to custom use by the agent

* cold restart, which doesn't work well, since the mount changes on all participating VMs in the same time, causing a service downtime
* manual rolling restart on secret change (not user friendly)

[ ] Docs
[ ] Installation
[x] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

**Additional context**

cwdsuzhou:
Make envoy hotRestart configurable

Fix https://github.com/istio/istio/issues/32696


cwdsuzhou:
@googlebot I signed it!

howardjohn (commented):
Can you provide some more context?

cwdsuzhou:
> Can you provide some more context?

1. We are trying to implement `hot upgrade` of `envoy`. When this `hotRestart` is disabled, it is impossible to do this.
2. I find `--disable-hot-restart` was added from codes , this is not flexible enough.

So we would like make it `disabled` by default, but also can be enabled through flag.

Issue #35489, opened 2021-10-06
VirtualService: HeaderMatch fails with prefix=""


philipsparrow:
Hi folks,

Using Anthos Service Mesh 198-1 (Istio 1.9.8)

Tried deploying a VirtualService with header matching.

According to docs (https://istio.io/v1.9/docs/reference/config/networking/virtual-service/#HTTPMatchRequest):
> If the value is empty and only the name of header is specfied, presence of the header is checked.

e.g.
```
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: api
spec:
  hosts:
    - "*"
  gateways:
    - gw
  http:
    - match:
        - uri:
            exact: /health
          headers:
            x-my-header:
              prefix: "" # Empty value means only the presence of the header is checked
      route:
        - destination:
            host: echo-service.default.svc.cluster.local
            port:
              number: 80
```

This doesn't work because I used `prefix: ""` rather than `exact: ""`.

Perhaps this is simply a documentation issue?

The error in my ingress gateway:

```
2021-10-05T23:41:28.472641Z	warning	envoy config	gRPC config for type.googleapis.com/envoy.config.route.v3.RouteConfiguration rejected: Proto constraint validation failed (RouteConfigurationValidationError.VirtualHosts[i]: ["embedded message failed validation"] | caused by VirtualHostValidationError.Routes[i]: ["embedded message failed validation"] | caused by RouteValidationError.Match: ["embedded message failed validation"] | caused by RouteMatchValidationError.Headers[i]: ["embedded message failed validation"] | caused by HeaderMatcherValidationError.PrefixMatch: ["value length must be at least " '\x01' " runes"]): name: "http.80"
virtual_hosts {
  name: "*:80"
  domains: "*"
  routes {
    match {
      path: "/health"
      case_sensitive {
        value: true
      }
      headers {
        name: "x-my-header"
        prefix_match: ""
      }
    }
    route {
      cluster: "outbound|80||echo-service.default.svc.cluster.local"
      timeout {
      }
      retry_policy {
        retry_on: "connect-failure,refused-stream,unavailable,cancelled,retriable-status-codes"
        num_retries {
          value: 2
        }
...
```

Thanks, I'm loving Istio!

howardjohn:
```
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: api
spec:
  hosts:
    - "echo"
  gateways:
    - mesh
  http:
    - match:
[...]

PR #35899, opened 2021-11-04
WIP relove empty value "" in headers

MouceL:

resolve the problems describe here https://github.com/istio/istio/issues/35489

### DIFF
--- a/istio.deps
+++ b/istio.deps
@@ -4,6 +4,6 @@
     "name": "PROXY_REPO_SHA",
     "repoName": "proxy",
     "file": "",
-    "lastStableSHA": "fa1b725bdddb18222fa0d676147ba46f49fbaed3"
+    "lastStableSHA": "79efbe6829d95344469b5a05eec8ca09e2e74595"
   }
 ]


MouceL:
@googlebot I signed it!

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -988,17 +988,23 @@ func translateHeaderMatch(name string, in *networking.StringMatch) *route.Header
 
 	switch m := in.MatchType.(type) {
 	case *networking.StringMatch_Exact:
-		out.HeaderMatchSpecifier = &route.HeaderMatcher_ExactMatch{ExactMatch: m.Exact}
+		if m.Exact != "" {
```
howardjohn, line 991: this doesn't seem to solve it, just makes it not do anything.

https://github.com/istio/istio/blob/90e61d212dce2550e29250afb053ed45099adb6a/pilot/pkg/networking/core/v1alpha3/route/route.go#L923 is the code intended to solve this I believe. I think its incorrectly checking `nil` when `""` can occur as well?

howardjohn: Although I think maybe this should be invalid and rejected in the webhook.

MouceL: > this doesn't seem to solve it, just makes it not do anything.
> 
> https://github.com/istio/istio/blob/90e61d212dce2550e29250afb053ed45099adb6a/pilot/pkg/networking/core/v1alpha3/route/route.go#L923
> 
> is the code intended to solve this I believe. I think its incorrectly checking `nil` when `""` can occur as well?

  In the absence of any header match specifier, match will default to present_match, so it seems work in envoy when you set  prefix: ""  
headers:
    x-my-header:
      prefix: "" # Empty value means only the presence of the header is checked

==>
          "match": {
           "path": "/productpage",
           "case_sensitive": true,
           "headers": [
            {
             "name": "x-my-header"
            }
           ]
          }

MouceL: according `(repeated config.route.v3.HeaderMatcher) Specifies a set of headers that the route should match on. The router will check the request’s headers against all the specified headers in the route config. A match will happen if all the headers in the route are present in the request with the same values (or based on presence if the value field is not in the config)`  ==>  so  if headers only contain name, the match will default to present_match , it's envoy feature

MouceL: @cywang1905 hello

cywang1905: = =

cywang1905:
@MouceL = =

PR #35900, opened 2021-11-04
Fuzzing: Improve coverage

AdamKorcz:
Adds more fuzzers to improve test coverage.
From a high level, more coverage is being added for the following areas of the code:

- Kube CRD
- Aggregate controller
- Kube gateway controller
- Status reconciler
- Workload entry controller
- CA Server
- Security utilities
- Security pki utilities
- Pilot status

This PR does not modify any production code. It only adds new fuzz tests.

The PR does remove the analyzer fuzzer from the regression tests since it broke it because of the recent major refactoring. I will take it upon myself to add it back but would prefer to do this in a separate PR, since I have some things to add to this fuzzer as well.

### DIFF
--- a/tests/fuzz/analyzer_fuzzer.go
+++ b/tests/fuzz/analyzer_fuzzer.go
@@ -93,7 +93,7 @@ func runAnalyzer(sa *local.IstiodAnalyzer) (local.AnalysisResult, error) {
 }
 
 // FuzzAnalyzer implements the fuzzer
-func FFuzzAnalyzer(data []byte) int {
+func FuzzAnalyzer(data []byte) int {
 	f := fuzz.NewConsumer(data)
 	analyzerIndex, err := f.GetInt()
 	if err != nil {
--- a/tests/fuzz/analyzer_fuzzer.go
+++ b/tests/fuzz/analyzer_fuzzer.go
@@ -93,7 +93,7 @@ func runAnalyzer(sa *local.IstiodAnalyzer) (local.AnalysisResult, error) {
 }
 
 // FuzzAnalyzer implements the fuzzer
-func FuzzAnalyzer(data []byte) int {
+func FFuzzAnalyzer(data []byte) int {
 	f := fuzz.NewConsumer(data)
 	analyzerIndex, err := f.GetInt()
 	if err != nil {
--- a/tests/fuzz/pki_fuzzer.go
+++ b/tests/fuzz/pki_fuzzer.go
@@ -1,0 +1,222 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// nolint: golint
+package fuzz
+
+import (
+	"crypto/x509/pkix"
+	"os"
+
+	fuzz "github.com/AdaLogics/go-fuzz-headers"
+
+	"istio.io/istio/security/pkg/pki/util"
+)
+
+// FuzzVerifyCertificate implements a fuzzer
+// that tests util.VerifyCertificate
+func FuzzVerifyCertificate(data []byte) int {
+	f := fuzz.NewConsumer(data)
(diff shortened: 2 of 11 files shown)


AdamKorcz:
Could we move forward with this PR as well?

tests/fuzz/analyzer_fuzzer.go
```diff
@@ -93,7 +93,7 @@ func runAnalyzer(sa *local.IstiodAnalyzer) (local.AnalysisResult, error) {
 }
 
 // FuzzAnalyzer implements the fuzzer
-func FuzzAnalyzer(data []byte) int {
+func FFuzzAnalyzer(data []byte) int {
```
howardjohn, line 96: Why is this changed?

howardjohn (commented):
lgtm except the FuzzAnalyzer part I am not sure of

howardjohn (approved):

howardjohn:
Missed the comment

> The PR does remove the analyzer fuzzer from the regression tests since it broke it because of the recent major refactoring. I will take it upon myself to add it back but would prefer to do this in a separate PR, since I have some things to add to this fuzzer as well.

Is this just a merge conflict maybe? The expected state currently is the fuzzer is enabled, and has no 'brokenCases' as we fixed those in the main codebase

AdamKorcz:
> Is this just a merge conflict maybe?

No, it is related to the recent refactoring. If you would rather not have a fuzzer that _should_ be tested but _is not_ being tested, then I am fine with removing/deactivating it completely and add it back later.
Fixing this fuzzer will fit nicer with adding the updated analyzer fuzzer I had initially uploaded in this PR but which now needs some adjustments.

tests/fuzz/analyzer_fuzzer.go
AdamKorcz: Marking as solved because of [this comment](https://github.com/istio/istio/pull/35900#issuecomment-971786345).

howardjohn:
> No, it is related to the recent refactoring. There were also merge conflicts, but it seemed to be broken in several ways. If you would rather not have a fuzzer that should be tested but is not being tested, then I am fine with removing/deactivating it completely and add it back later.

How is it broken or not tested?

All of the fuzzers are run in the oss-fuzz, and all fuzzers are tested in unit tests with a set of simple smoke-test cases. In addition, tests that have ever failed are run against those cases. Some also have *known broken cases* (ie bugs in our code). The recent PR to TestAnalyzer *fixed* a bug, causing the known broken cases to be reduced. The test coverage is still the same

AdamKorcz:
> How is it broken or not tested?

I meant that this PR will result in the analyzer fuzzer not being included in the regression tests.

AdamKorcz:
Adding the analyzer fuzzer back to the regression test.

tests/fuzz/pki_fuzzer.go
```diff
@@ -0,0 +1,222 @@
...
+func FuzzParsePemEncodedCertificateChain(data []byte) int {
+	_, _ = util.ParsePemEncodedCertificateChain(data)
+	return 1
+}
+
+// FuzzUpdateVerifiedKeyCertBundleFromFile implements a
+// fuzzer that tests UpdateVerifiedKeyCertBundleFromFile
+func FuzzUpdateVerifiedKeyCertBundleFromFile(data []byte) int {
+	f := fuzz.NewConsumer(data)
+	certFile, err := os.Create("certfile")
```
howardjohn, line 100: can we delete it on cleanup?

Also is using files for fuzzing really appropriate? Seems like we should just fuzz NewVerifiedKeyCertBundleFromPem instead?

tests/fuzz/regression_test.go
```diff
@@ -35,7 +35,10 @@ var baseCases = [][]byte{
 
 // brokenCases contains test cases that are currently failing. These should only be added if the
 // failure is publicly disclosed!
-var brokenCases = map[string]string{}
+var brokenCases = map[string]string{
+	"6169070276837376": "https://github.com/go-yaml/yaml/issues/666",
+	"6087702507290624": "https://github.com/go-yaml/yaml/issues/768",
+}
```
howardjohn, line 41: If you revert this change I think things will get into the expected state.

AdamKorcz: thanks!

tests/fuzz/pki_fuzzer.go
AdamKorcz: > can we delete it on cleanup?

Sure, added.

> Also is using files for fuzzing really appropriate?

Yes, it is fine as long as it is fast enough. I suggest we try `UpdateVerifiedKeyCertBundleFromFile` and observe if the speed is poor after a few weeks. Otherwise we stick with `NewVerifiedKeyCertBundleFromPem`.

PR #35921, opened 2021-11-05
Enable/Fix multi-arch build for DOCKER_V2_BUILDER

TrevorTaoARM:
When DOCKER_V2_BUILDER=true, the build on arm64 would get an error:
./tools/docker: line 29: out/linux_amd64/docker-builder: cannot execute
binary file: Exec format error

which is due to lack of arch and os information in tools/docker when
building with common/scripts/gobuild.sh(so it would lead to an amd64
binary on arm64 platform).

Now we set the right OS and Arch to enable and fix the building both
on arm64 and amd64.

### DIFF
--- a/tools/docker
+++ b/tools/docker
@@ -20,17 +20,22 @@ ROOT=$(dirname "$WD")
 
 set -eu
 
+TARGET_OS=${TARGET_OS:-linux}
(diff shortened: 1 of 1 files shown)


TrevorTaoARM:
I will submit a little bit later after our internal review.

PR #35925, opened 2021-11-05
experiment with headless filter chains

howardjohn:


howardjohn:
It works, proving we could move headless to FC. Closing for now

PR #35928, opened 2021-11-05
added e2e test for jwksuri

aryan16:
- Added e2e test for jwksResolverExtraRootCA
- Updated the deployment file to support https jwk server


tests/integration/security/testdata/requestauthn/remote.yaml.tmpl
```diff
@@ -9,7 +9,7 @@ spec:
       app: {{ .dst }}
   jwtRules:
   - issuer: "test-issuer-1@istio.io"
-    jwksUri: "http://jwt-server.{{ .Namespace }}.svc.cluster.local:8000/jwks"
+    jwksUri: "https://jwt-server.default.svc.cluster.local:8443/jwks"
```
lei-tang, line 12: Can you comment on why replacing {{ .Namespace }} with a hardcoded namespace "default"?
Should a new jwksUri be created for the new httpbs-based test case, instead of revising an existing jwksUri?

tests/integration/security/main_test.go
```diff
@@ -49,6 +49,25 @@ func setupConfig(ctx resource.Context, cfg *istio.Config) {
 	cfg.ControlPlaneValues = `
 values:
   pilot: 
+    jwksResolverExtraRootCA: |
+      MIIDRTCCAi2gAwIBAgIUBeTBvkDJcwG0pIYWP/VphzEGWVEwDQYJKoZIhvcNAQEL
```
lei-tang, line 53: Consider adding a comment on how this cert is generated, similar to the following comment in the PR
#command to generate certificate and a key
  #openssl req \ -x509 \ -nodes \ -newkey rsa:2048 \ -keyout server.key \ -out server.crt \ -days 3650 \

tests/integration/security/testdata/requestauthn/remote.yaml.tmpl
aryan16: We haven't specified any ns in the deployment, it will always use default namespace. That's my reasoning, correct me if I'm wrong here.

tests/integration/security/main_test.go
aryan16: sure, thanks for the suggestion !!

tests/integration/security/testdata/requestauthn/remote.yaml.tmpl
lei-tang: As {{ .Namespace }} will always be populated with "default", I suggest leave it as is. Meanwhile, to cover both http and https, should a new https-based test case be created, instead of revising the existing http test case?

aryan16: sure I will revert this change. @yangminzhu do we need to test https separately here ? We have done the unit testing separately though.

samples/jwt-server/jwt-server.yaml
```diff
@@ -25,9 +25,106 @@ spec:
...
     port: 8000
     targetPort: 8000
+  - name: https
+    port: 8443
+    targetPort: 8443
   selector:
     app: jwt-server
 ---
+apiVersion: v1
+kind: ConfigMap
```
howardjohn, line 35: very nit: store as a Secret.

I know its not "secret" in the sense its in a public github repo, but Secret is literally designed to store TLS certs (there is a specific type for it)

tests/integration/security/testdata/requestauthn/remote.yaml.tmpl
howardjohn: It is using a namespace actually I think: https://github.com/istio/istio/blob/424695809a6c77e45642f5ec2090e66dd88b89b1/tests/integration/security/jwt_test.go#L324

samples/jwt-server/jwt-server.yaml
```diff
@@ -25,9 +25,106 @@ spec:
...
+metadata:
+  name: jwt-server-cert-key
+data:
+  #command to generate certificate and a key
```
howardjohn, line 39: are current certs sufficient? like `tests/testdata/certs/cert.crt`?

Not a big concern, but it gets confusing to have a bunch of test certs around.

howardjohn (commented):
overall looks good

samples/jwt-server/jwt-server.yaml
```diff
@@ -25,9 +25,106 @@ spec:
...
+  #openssl req \ -x509 \ -nodes \ -newkey rsa:2048 \ -keyout server.key \ -out server.crt \ -days 3650 \
```
yangminzhu, line 40: remove the `\` in the command, you are already putting it in a single line, you don't need the `\` any more.

tests/integration/security/testdata/requestauthn/remote.yaml.tmpl
yangminzhu: yeah, I think the namespace is used, and it makes sense to have a separate test.

samples/jwt-server/jwt-server.yaml
aryan16: good point, I will check that.

samples/jwt-server/jwt-server.yaml
aryan16: I agree but I would require SAN enabled certs and for this "tests/testdata/certs/cert.crt", SAN is empty.

samples/jwt-server/jwt-server.yaml
aryan16: sure !!

tests/integration/security/testdata/requestauthn/remotehttps.yaml.tmpl
```diff
@@ -0,0 +1,15 @@
...
+kind: RequestAuthentication
+metadata:
+  name: request-authn
+spec:
+  selector:
+    matchLabels:
+      app: {{ .dst }}
+  jwtRules:
+  - issuer: "test-issuer-1@istio.io"
+    jwksUri: "https://jwt-server.{{ .Namespace }}.svc.cluster.local:8443/jwks"
```
howardjohn, line 12: on the cert I see `DNS:*.localhost.com, DNS:*.local, DNS:localhost`. So how does this work?

aryan16: request made to *.local or *.localhost.com or localhost can be validated with this cert.

howardjohn: `*.local` is not supposed to match `x.y.local` IIUC. It should only matches up to one label?

aryan16: '*' would signify any number of char, so I guess it will allow multiple dots.

yangminzhu: good catch, the `*.local` won't match with the `jwt-server.<namespace.svc.cluster.local` in certificate, see https://en.wikipedia.org/wiki/Wildcard_certificate

I think the easiest fix is to generate the certificate with DNS name `jwt-server` and `localhost` and modify to deploy it in the same namespace as istiod, which should be istio-system by default, and change the RequestAuthentication jwksUri to just `https://jwt-server:8443`, you could omit the namespace part as long as they are in the same namespace.

tests/integration/security/main_test.go
```diff
@@ -45,10 +45,30 @@ func setupConfig(ctx resource.Context, cfg *istio.Config) {
 	if cfg == nil {
 		return
 	}
-
+	// command to generate certificate
+	// openssl req -x509 -nodes -newkey rsa:2048 -keyout server.key -out server.crt -days 3650
```
yangminzhu, line 49: I don't think this command generates the following certificate, could you just mention to follow the `server.crt` instructions in `samples/jwt-server/testdata/README.md`

tests/integration/security/testdata/requestauthn/remotehttps.yaml.tmpl
```diff
@@ -0,0 +1,15 @@
...
+    outputPayloadToHeader: "x-test-payload"
```
yangminzhu, line 13: Could you change this to a different value (e.g. `x-test-payload-for-https-jwksuri) to make sure the jwt config is in effective for test? You should also update the test code with the new value.

tests/integration/security/testdata/requestauthn/remotehttps.yaml.tmpl
howardjohn: I think you can do `*.*.svc.cluster.local` too. Probably not a great idea though... Also would be ideal to not encode `cluster.local` if possible because we eventually want to make it tested with that customized. But that is probably a long way off, do not need to solve it here.

yangminzhu: I thought only the left most `*` will work as the wildcard matching? I did a quick test by modifying the unit test and it seems so.

aryan16: @yangminzhu I tried to do it by adding jwt-server in the dns, but it is failing. And one more thing, in the test we are doing the deployment as well right ? https://github.com/istio/istio/blob/master/tests/integration/security/jwt_test.go#:~:text=jwtserver%20%3A%3D%20applyyaml(%22..%2F..%2F..%2Fsamples%2Fjwt-server%2Fjwt-server.yaml%22%2C%20ns)

tests/integration/security/main_test.go
aryan16: cool. I'll do that.

tests/integration/security/testdata/requestauthn/remotehttps.yaml.tmpl
yangminzhu: could you clarify what is failing? you need to change that line to pass the `istio-system` ns.

aryan16: this is the error which I'm receiving - `failed calling a (cluster=cluster-0)->'http://b.test-ns1-1-37638.svc.cluster.local:8095//valid-token-forward-remote-jwks': call failed from a (cluster=cluster-0) to http://b.test-ns1-1-37638.svc.cluster.local:8095/valid-token-forward-remote-jwks (using http)`

aryan16: @howardjohn correct me if I'm wrong, I just read about wildcards and found out multiple wildcard doesn't work - https://en.wikipedia.org/wiki/Wildcard_certificate#:~:text=A%20cert%20with,*.*.domain.com

tests/integration/security/jwt_test.go
```diff
@@ -46,17 +46,17 @@ const (
 func TestRequestAuthentication(t *testing.T) {
 	payload1 := strings.Split(jwt.TokenIssuer1, ".")[1]
 	payload2 := strings.Split(jwt.TokenIssuer2, ".")[1]
+
 	framework.NewTest(t).
 		Features("security.authentication.jwt").
 		Run(func(t framework.TestContext) {
-			ns := apps.Namespace1
+			ns := istio.ClaimSystemNamespaceOrFail(t, t)
```
yangminzhu, line 53: hmm, probably better to use a different variable for the istio-system namespace, otherwise you will also create the policies in the istio-system namespace which would work since it's the root namespace but I'm not sure this is needed for this test.

samples/jwt-server/jwt-server.yaml
```diff
@@ -25,9 +25,70 @@ spec:
...
     app: jwt-server
 ---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: jwt-cert-key-secret
+# command to generate certificate
+# https://github.com/istio/istio/blob/master/samples/jwt-server/testdata/README.MD
+stringData: 
+  example.crt: |
```
yangminzhu, line 41: Let's use the `server.crt` and `server.key` to match with the name used in the README.

tests/integration/security/main_test.go
```diff
@@ -45,10 +45,30 @@ func setupConfig(ctx resource.Context, cfg *istio.Config) {
 	if cfg == nil {
 		return
 	}
-
+	// command to generate certificate
+	// https://github.com/istio/istio/blob/master/samples/jwt-server/testdata/README.MD
```
yangminzhu, line 49: Is this using the `ca.crt`? Better to clarify which cert is used since the REAME generates two certs.

yangminzhu (commented):
Overall looks great, just some small comments, thanks!

tests/integration/security/main_test.go
aryan16: oh it's server.crt. I'll add more comments here.

samples/jwt-server/jwt-server.yaml
aryan16: sure!!

samples/jwt-server/testdata/README.MD
```diff
@@ -15,5 +15,5 @@ openssl req -newkey rsa:2048 -nodes -keyout server.key -subj "/C=US/ST=AZ/O=Acme
 
 ### Signing Certificate with CA
 ```
-openssl x509 -req -extfile <(printf "subjectAltName=DNS:*.localhost.com,DNS:*.local,DNS:localhost") -days 365 -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt
+openssl x509 -req -extfile <(printf "subjectAltName=DNS:jwt-server,DNS:localhost") -days 365 -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt
```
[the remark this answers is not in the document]
aryan16: sure!!

tests/integration/security/jwt_test.go
aryan16: so the same namespace is being used while iterating the testcases. That's why it was failing previously. If I add a new var for istio-system, I need to add a condition when we are iterating the testcases.

yangminzhu: why, you just create a new var for istio-system and use it for the jwt-server, the policy will still use the old ns, how could this fail? The namespace of the jwt-server should not affect the policy namespace.

tests/integration/security/https_jwt/https_jwt_test.go
```diff
@@ -0,0 +1,148 @@
...
+			applyYAML := func(filename string, ns namespace.Instance) []string {
+				policy := tmpl.EvaluateAllOrFail(t, args, file.AsStringOrFail(t, filename))
+				for _, cluster := range t.AllClusters() {
+					t.ConfigKube(cluster).ApplyYAMLOrFail(t, ns.Name(), policy...)
+				}
+				return policy
+			}
+
+			jwtServer := applyYAML("../../../../samples/jwt-server/jwt-server.yaml", istioSystemNS)
+			for _, cluster := range t.AllClusters() {
```
yangminzhu, line 64: you can just:

```go
defer func() {
  for _, cluster := range t.AllClusters() {
    t.ConfigKube(cluster).DeleteYAMLOrFail(t, istioSystemNS.Name(), jwtServer...)
  }
}()

File: tests/integration/security/testdata/requestauthn/remotehttps.yaml.tmpl (hunk 8 above)
Comment (yangminzhu) [reply tests/integration/security/testdata/requestauthn/remotehttps.yaml.tmpl:13]: can you address the comment?

File: tests/integration/security/https_jwt/main_test.go [hunk 14]
```diff
@@ -0,0 +1,84 @@
...
+      AH6ybGMGwM3cXqLbZtR1ayJJhJx/v2CWK6WOc7T73nm+oQWyb5byvfMPa4G7JRmc
+      XWQbIoB9w5Gu62Vo3CFCleLq7uSKKjKw8XNMpKNm2MsL7pKPJs13XbjXRoyn/OP3
+      IjrAi3+wks5Wt8BhwJDp1JJLiRF1FGKLquuO7alm8zAs6PSXRePDPXQdSfDCGy+q
+      StG8hekFKF+BomUBRbuuMjSKPlnD3eMIdmWXHsTW8Gg1Anua7ddKD9aZUAxbNHk4
+      31paKAmE2v8g/ZUtFKYLJyFhVqe8pB6IhQZm3/3C8xvJpy1sTj7u+ydqkzHbIUfh
+      Iaw5/US65AS6weIJcztH2Zw=
+      -----END CERTIFICATE-----
+    env: 
+      PILOT_JWT_ENABLE_REMOTE_JWKS: false
+meshConfig:
```
yangminzhu, line 78: nit: the `accessLogEncoding` and `defaultConfig` can just be removed, you can keep the `accessLogFile` to make the debug easier.

tests/integration/security/https_jwt/https_jwt_test.go
```diff
@@ -0,0 +1,148 @@
...
+				}
+
+				if testCase.SkipMultiCluster && t.Clusters().IsMulticluster() {
+					t.Skip()
+				}
+				echotest.New(t, apps.All).
+					SetupForDestination(func(t framework.TestContext, dst echo.Instances) error {
+						if testCase.Config != "" {
+							policy := yml.MustApplyNamespace(t, tmpl.MustEvaluate(
+								file.AsStringOrFail(t, fmt.Sprintf("../testdata/requestauthn/%s.yaml.tmpl", testCase.Config)),
```
yangminzhu, line 119: please just create a new testdata directory together with the test and put the yaml file there, you're already creating a new directory (`http_jwt`) so there is no need for using the old one.

aryan16: so is it okay if I just move **remotehttps.yaml.tmpl** file to **https_jwt** directory ?

yangminzhu: yes, I meant you should only move remotehttps.yaml.tmpl to https_jwt/testdata.

aryan16: sure, understood.

yangminzhu (approved):
Looks good!

aryan16:
@istio/wg-docs-maintainers could you please help me merge this PR. The PR is already reviewed by a member of Istio Security team @yangminzhu.

tests/integration/security/https_jwt/main_test.go
```diff
@@ -0,0 +1,80 @@
...
+      87hsljSiV7IHsmF0MA8GA1UdEwEB/wQFMAMBAf8wDQYJKoZIhvcNAQELBQADggEB
+      AH6ybGMGwM3cXqLbZtR1ayJJhJx/v2CWK6WOc7T73nm+oQWyb5byvfMPa4G7JRmc
+      XWQbIoB9w5Gu62Vo3CFCleLq7uSKKjKw8XNMpKNm2MsL7pKPJs13XbjXRoyn/OP3
+      IjrAi3+wks5Wt8BhwJDp1JJLiRF1FGKLquuO7alm8zAs6PSXRePDPXQdSfDCGy+q
+      StG8hekFKF+BomUBRbuuMjSKPlnD3eMIdmWXHsTW8Gg1Anua7ddKD9aZUAxbNHk4
+      31paKAmE2v8g/ZUtFKYLJyFhVqe8pB6IhQZm3/3C8xvJpy1sTj7u+ydqkzHbIUfh
+      Iaw5/US65AS6weIJcztH2Zw=
+      -----END CERTIFICATE-----
+    env: 
+      PILOT_JWT_ENABLE_REMOTE_JWKS: false
```
howardjohn, line 77: Its a bit confusing to me that we test HTTPS with `false` and HTTP with `true`. I would expect we test the full matrix of the settings (ie 4 tests). Is this since we did not have time, or are we explicitly eliding the other 2 tests cases? Either way, a comment or TODO may help

tests/integration/security/https_jwt/main_test.go
```diff
@@ -0,0 +1,80 @@
...
+		return
+	}
+
+	// command to generate certificate
+	// use the generated ca.crt by following https://github.com/istio/istio/blob/master/samples/jwt-server/testdata/README.MD
+	cfg.ControlPlaneValues = `
+values:
+  pilot: 
+    jwksResolverExtraRootCA: |
+      -----BEGIN CERTIFICATE-----
```
howardjohn, line 55: nit: would be slightly nicer to read from file, so its not copied

tests/integration/security/https_jwt/https_jwt_test.go
```diff
@@ -0,0 +1,150 @@
...
+			args := map[string]string{"Namespace": istioSystemNS.Name()}
+			applyYAML := func(filename string, ns namespace.Instance) []string {
+				policy := tmpl.EvaluateAllOrFail(t, args, file.AsStringOrFail(t, filename))
+				for _, cluster := range t.AllClusters() {
+					t.ConfigKube(cluster).ApplyYAMLOrFail(t, ns.Name(), policy...)
+				}
+				return policy
+			}
+
+			jwtServer := applyYAML("../../../../samples/jwt-server/jwt-server.yaml", istioSystemNS)
```
howardjohn, line 63: nit: `filepath.Join(env.IstioSrc,...`, so its not dependant on file structure

tests/integration/security/https_jwt/main_test.go
aryan16: I guess we can't read it from file afaik, we are creating yaml template inside golang.

tests/integration/security/https_jwt/https_jwt_test.go
aryan16: I'm sorry but I couldn't understand the comment. Could you please elaborate it little bit.

tests/integration/security/https_jwt/main_test.go
aryan16: sure, I will add a TODO about it.

aryan16:
cc: @yangminzhu

zirain:
run `make lint` to fix lint_istio error

tests/integration/security/https_jwt/main_test.go
yangminzhu: +1 it's better to read from the file instead of hard coding the cert.

tests/integration/security/https_jwt/main_test.go
yangminzhu: yes please add the TODO, ideally we should run the JWT test cases with both true/false values of the flag. Currently it was only tested with one value.

aryan16: oh I added that then I reverted this file to previous commit and forgot to add that again. I will fix this. Thanks

tests/integration/security/https_jwt/main_test.go
aryan16: I tried doing that way, but formatting got messed up which resulted in errors.

yangminzhu: this should be doable using the go template and the indent function, check the example here: https://github.com/istio/istio/blob/master/tests/integration/pilot/locality_test.go#L75

aryan16:
cc : @yangminzhu

yangminzhu (approved):

howardjohn (approved):

nmittler (approved):

PR #35930, opened 2021-11-05
Set sysctls for gateway ports

howardjohn:
Currently, we have two options for gateway ports:
* Use targetPort to bind >1024
* Run as root

Since Kubernetes 1.22, there is another option available, by
configuring the sysctls on the pod.

This is especially important in the gateway controller, as users do not
control the Service. This means we would either need to automagically
configure the targetPort, which becomes tricky (what if we set it to
8080 and then they also want to bind to 8080?) - especially since things
like AuthorizationPolicy use the target port! Even worse, we can never
change it without downtime, so we are locking ourselves into using
targetPort when its not even needed anymore on newer Kubernetes
versions.

This changes the *new* gateway injection template and the Gateway
controller to use this new option when 1.22+ is used. Both of these are
net-new in Istio 1.12, so there is no compatibility concerns at all.

Note: currently the gateway controller is running as root. So this is
dropping privileges, not adding them.

 Prior attempt: https://github.com/istio/istio/pull/35846

### DIFF
--- a/manifests/charts/gateway/templates/deployment.yaml
+++ b/manifests/charts/gateway/templates/deployment.yaml
@@ -1,11 +1,3 @@
-{{/*
-unPriv port determines if net.ipv4.ip_unprivileged_port_start should be set, which allows binding to low ports like 80.
-Because its not safe for all versions, we allow explicitly configuring or automatically setting based on version
-This is safe since 1.22: https://github.com/kubernetes/kubernetes/pull/103326.
-The expression is complex but is basically `unpriv = explicitly enabled || (auto && recent enough version)`
-If not true, users should bind to ports >1024 with targetPort or set runAsRoot.
-*/}}
-{{ $unprivPort := or (eq (toString .Values.unprivilegedPort) "true") (and (eq (toString .Values.unprivilegedPort) "auto") (semverCompare ">=1.22-0" .Capabilities.KubeVersion.GitVersion)) }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -40,36 +32,45 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "gateway.serviceAccountName" . }}
-      {{- if not .Values.runAsRoot }}
       securityContext:
-        {{- if $unprivPort }}
+      {{- if .Values.securityContext }}
+        {{- toYaml .Values.securityContext | nindent 8 }}
+      {{- else if (semverCompare ">=1.22-0" .Capabilities.KubeVersion.GitVersion) }}
         # Safe since 1.22: https://github.com/kubernetes/kubernetes/pull/103326
         sysctls:
         - name: net.ipv4.ip_unprivileged_port_start
           value: "0"
-        {{- end }}
-        runAsUser: 1337
-        runAsGroup: 1337
-        runAsNonRoot: true
-        fsGroup: 1337
       {{- end }}
       containers:
         - name: istio-proxy
           image: auto
-          {{- if not .Values.runAsRoot }}
           securityContext:
-            allowPrivilegeEscalation: false
+          {{- if .Values.containerSecurityContext }}
+            {{- toYaml .Values.containerSecurityContext | nindent 12 }}
+          {{- else if (semverCompare ">=1.22-0" .Capabilities.KubeVersion.GitVersion) }}
+            # Safe since 1.22: https://github.com/kubernetes/kubernetes/pull/103326
             capabilities:
               drop:
               - ALL
+            allowPrivilegeEscalation: false
             privileged: false
+            readOnlyRootFilesystem: true
+            runAsUser: 1337
+            runAsGroup: 1337
+            runAsNonRoot: true
+          {{- else }}
+            capabilities:
+              drop:
+              - ALL
+              add:
+              - NET_BIND_SERVICE
+            runAsUser: 0
+            runAsGroup: 1337
+            runAsNonRoot: false
+            allowPrivilegeEscalation: true
             readOnlyRootFilesystem: true
           {{- end }}
           env:
-          {{- if and (not .Values.runAsRoot) (not $unprivPort) }}
-          - name: ISTIO_META_UNPRIVILEGED_POD
-            value: "true"
-          {{- end }}
           {{- with .Values.networkGateway }}
           - name: ISTIO_META_REQUESTED_NETWORK_VIEW
             value: "{{.}}"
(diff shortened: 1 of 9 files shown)


manifests/charts/gateway/templates/deployment.yaml
```diff
@@ -52,7 +66,7 @@ spec:
             readOnlyRootFilesystem: true
           {{- end }}
           env:
-          {{- if not .Values.runAsRoot }}
+          {{- if and (not .Values.runAsRoot) (not $unprivPort) }}
```
costinm, line 69: Will this change the semantics on GKE1.22 plus ? I.e. will it result in Istiod binding to port 80 instead of 8080 ? And if yes, any backward incompat problems ( for examples logs showing a different port after upgrade, or authz policies no longer matching ) ?

manifests/charts/gateway/templates/deployment.yaml
```diff
@@ -34,6 +42,12 @@ spec:
       serviceAccountName: {{ include "gateway.serviceAccountName" . }}
       {{- if not .Values.runAsRoot }}
       securityContext:
+        {{- if $unprivPort }}
```
costinm, line 45: I would set this if "semverCompare > 1.22" - regardless of the values.unpriviledged port.
Simpler, no harm in having this.

manifests/charts/gateway/values.yaml
```diff
@@ -28,6 +28,11 @@ podAnnotations:
 
 # Whether to run the gateway in a privileged container
 runAsRoot: false
+# Whether to enable net.ipv4.ip_unprivileged_port_start. Generally if this is set, runAsRoot should be false,
+# unless root is desired for other reasons.
+# This is only supported by default on 1.22+; when auto is used this will be set based on the Kubernetes version.
+# Note: `helm template` cannot determine the version, so `auto` may yield incorrect results.
+unprivilegedPort: auto
```
costinm, line 35: I would default to false, to avoid risk of upgrade problems. 

Also please comments that are user-focused - i.e. this option will cause gateway pods to bind on port 80/443 and other unprivileged ports. And some info on the impact on telemetry and authz/other policies on port.

manifests/charts/gateway/templates/deployment.yaml
costinm: It is also consistent with the auto-created ones.

costinm (commented):
Great change - but not sure if safe on by default.

manifests/charts/gateway/templates/deployment.yaml
howardjohn: No. on 1.21-, they will be broken if they set a low port. So the chart defaults to 80->8080. When they enable this, it does not _automatically_ switch the 8080 to 80. They could, but its not OOTB.

howardjohn: The alternative would be to have:
1.21-: root
1.22: sysctl

and never set targetPort by default. If users on 1.21 and older want nonRoot, they can explicitly configure the ports.

Pros:
* Matches gateway controller
* Better long term - users don't need to change port #s and get the long term desired behavior (bind to 80)

Cons:
* Users have degraded security posture on 1.21 by default

howardjohn:
**Gateway API**
Option 1: only use direct port mapping. Run as root or use sysctl depending on version.
Pros:
* Ideal long term state. Authorization policy makes sense (uses port 80)
Cons:
* Migration from default usage (8080) may be more complex, since they may need to change Authz policy from 8080 -> 80. Migrating to gateway-api will already be a fairly involved change though; its not automatic. Using automated deployment is optional; users can always do it manually and use port 8080 if they desire.
Option 2: define a port mapping. Likely something like 80 -> 15080
Pros:
* No need to worry about permissions
Cons:
* Weird Authz Policy usage. We could make Authz policy aware of this mapping and translate as well, but it adds complexity to security sensitive code.
* Can never change it, as its backwards incompatible
* Makes logs/debugging/internal XDS a bi confusing

**Helm**

Values options:
   1a: With runAsRoot=true|false sysctl=false|auto|true. Fail if <1024 and both are false
   1b: With securityMode=none|sysctl|root|auto. Fail if <1024 and none. Auto is sysctl if >1.22, root if using port <1024, and none if ports are all >1024.

Option 1: use direct port mapping by default. 
Pros:
* Consistency with Gateway API (1).
* Long term best option (with sysctl mode)
Cons:
* Likely migration done when moving to this chart from old helm charts or istioctl. Possible not from plain YAML migration.

Option 2: use port translation by default
Pro:
* Consistency for helm/istioctl _default_ users adopting
Cons:
* No way to migrate to direct port mapping without downtime 

Overall I think we should use direct port mapping for both, and option (1b) for the values API. Overall this is better longer term and will give a consistent story across types. While there may be migration required for some users, especially on authz, there will be with any mechanism. We should bite the bullet now to take advantage of these 2 new greenfield deployment methods to go with the more modern preferred option. To mitigate authz issues, we can make documentation clear and add warning to analyze if configs apply to a port not in the Gateway at all. For non-k8s cases (like run on VM), this isn't directly relevant since this is only configuring our K8s deployments mechanisms - targetPort is still supported by the Pilot code, we just don't utilize it in our k8s deployments.

costinm (approved):

manifests/charts/gateway/templates/deployment.yaml
```diff
@@ -32,30 +32,45 @@ spec:
...
-        runAsUser: 1337
-        runAsGroup: 1337
-        runAsNonRoot: true
-        fsGroup: 1337
+      {{- if .Values.securityContext }}
+        {{- toYaml .Values.securityContext | nindent 8 }}
+      {{- else if (semverCompare ">=1.22-0" .Capabilities.KubeVersion.GitVersion) }}
+        # Safe since 1.22: https://github.com/kubernetes/kubernetes/pull/103326
+        sysctls:
+        - name: net.ipv4.ip_unprivileged_port_start
```
jacob-delgado, line 41: cc @howardjohn @Kmoneal 
does it matter for ipv6?

howardjohn: I am about 60% sure this applies to ipv6 and is poorly named. I did not test it though

pilot/pkg/config/kube/gateway/templates/deployment.yaml
```diff
@@ -32,9 +32,40 @@ spec:
           .Labels
           | nindent 8}}
     spec:
+      {{- if .KubeVersion122 }}
```
jacob-delgado, line 35: cc @howardjohn @Kmoneal 

Will this work for 1.23+? I thought the plan was to support 1.23 in testing after it is released and we can add it to the verified checks.

I'm wondering since above you had

```
	else if (semverCompare ">=1.22-0" .Capabilities.KubeVersion.GitVersion)
```

howardjohn: This is the same, it comes from `kube.IsAtLeastVersion(d.client, 22)`. It is poorly named though

PR #35934, opened 2021-11-05
Fix DefaultHTTPRetryPolicy for gateways

howardjohn:
We did not set mesh config so it did nothing...


hzxuzhonghu (commented):
lgtm, curious isn't there a test

stevenctl (approved):

PR #35945, opened 2021-11-08
Delete redundant service handler

dddddai:

It seems redundant to append service handler in `ClusterAdded`, because the handler has already been appended in `initRegistryEventHandlers`

This patch would reduce unnecessary push requests

related snippet:
https://github.com/istio/istio/blob/fe652368feb59068c6693452deaf5f6d8664c205/pilot/pkg/bootstrap/server.go#L861-L877
https://github.com/istio/istio/blob/fe652368feb59068c6693452deaf5f6d8664c205/pilot/pkg/serviceregistry/kube/controller/multicluster.go#L208
https://github.com/istio/istio/blob/fe652368feb59068c6693452deaf5f6d8664c205/pilot/pkg/serviceregistry/aggregate/controller.go#L71-L80

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/multicluster.go
+++ b/pilot/pkg/serviceregistry/kube/controller/multicluster.go
@@ -34,7 +34,6 @@ import (
 	"istio.io/istio/pkg/cluster"
 	"istio.io/istio/pkg/config/schema/collection"
 	"istio.io/istio/pkg/config/schema/collections"
-	"istio.io/istio/pkg/config/schema/gvk"
 	kubelib "istio.io/istio/pkg/kube"
 	"istio.io/istio/pkg/kube/multicluster"
 	"istio.io/istio/pkg/webhooks"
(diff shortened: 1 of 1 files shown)


stevenctl:
I think this patch would make it so we only fire the push for service updates in the "main" cluster this istiod is watching. 

Instead we could either:
1. Remove the handler we setup in `initRegistryEventHandlers`, since we do it in `multicluster.go`
2. Gate the `AppendServiceHandler` call in `multicluster.go` with `if localCluster`

dddddai:
@stevenctl Thanks for your review :)
> I think this patch would make it so we only fire the push for service updates in the "main" cluster this istiod is watching.

Why? When a "remote" cluster added, it would callback `ClusterAdded` and call `AddRegistryAndRun` to append the service handler to remote registry.

So it should fire the push for remote services, shouldn't it? Please correct me if I'm wrong

stevenctl:
Yes it should fire the push

hzxuzhonghu:
/ok-to-test

seems good

stevenctl (approved):

stevenctl:
misunderstood at first pass – this looks good

PR #35953, opened 2021-11-08
Switch port naming for generated Service

howardjohn:
Gateway Listener.name is already distinct, so it makes sense to mirror
that exactly. The current scheme may flip-flop depending on unimportant
YAML ordering

### DIFF
--- a/pilot/pkg/config/kube/gateway/deploymentcontroller.go
+++ b/pilot/pkg/config/kube/gateway/deploymentcontroller.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"strings"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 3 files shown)


stevenctl (approved):

pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -286,14 +285,13 @@ func extractServicePorts(gw gateway.Gateway) []corev1.ServicePort {
...
-	for i, l := range gw.Spec.Listeners {
+	for _, l := range gw.Spec.Listeners {
 		if _, f := portNums[int32(l.Port)]; f {
 			continue
 		}
 		portNums[int32(l.Port)] = struct{}{}
-		name := fmt.Sprintf("%s-%d", strings.ToLower(string(l.Protocol)), i)
 		svcPorts = append(svcPorts, corev1.ServicePort{
-			Name: name,
+			Name: string(l.Name),
```
costinm, line 294: Is Name a required field ( validated, etc) ? Should we have a fallback if not set ?

costinm (approved):

pilot/pkg/config/kube/gateway/deploymentcontroller.go
howardjohn: It is required, part of the CRD schema. Fallback is a good idea though, just in case..

PR #35967, opened 2021-11-09
fix the problem that missing remote pilot name in dns names

### LINKED ISSUE #35939: istiod in remote cluster lacks env ISTIO_CUSTOM_HOST to add its self host to altNames
### Bug Description

As the title. This will cause the ingress gateway in the remote cluster to report an error when initiating a CSR request to istiod-1-11-4 in its own cluster with the following message.

```
failed to warm certificate: failed to generate workload certificate: create certificate: rpc error: code = Unavailble desc = connection error: desc = "transport: authentication handshake failed: x509: certificate is valid for istiod-remote.istio-system.svc, istiod.istio-system.svc, istio-pilot.istio-system.svc, not istiod-1-11-4.istio-system.svc"
```

**It can be fixed by adding below env to the remote istiod**, 

```yaml
        env:
        - name: ISTIOD_CUSTOM_HOST
          value: istiod-1-11-4.istio-system.svc
```

I've tried to propose a PR to fix it. But maybe it's because I'm not familiar with the manifest part, I couldn't get expected result when I apply the change to `manifests/charts/istio-control/istio-discovery/templates/deployment.yaml` as below:

```yaml
{{- if not (eq $.Values.global.remotePilotAddress "") }}
          - name: ISTIOD_CUSTOM_HOST
            value: istiod{{- if not (eq .Values.revision "") }}-{{ .Values.revision }}{{- end }}
{{- end }}
```

I even found that no changes to this deployment template had any effect from the output of `./bin/istioctl manifest generate  --context="${CTX_CLUSTER2}" --revision=1-11-4 -f op_cluster2.yaml`.

### Version

```prose
./bin/istioctl version
client version: 1.11.4
istiod version: 1.11.4
istiod version: 1.9.9
data plane version: 1.11.4 (3 proxies)

istio-1.11.4 kubectl version --short
Client Version: v1.20.4
Server Version: v1.20.4
```

### Additional Information

_No response_

### Affected product area

- [ ] Docs
- [X] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Upgrade
- [ ] Multi Cluster
- [ ] Virtual Machine
- [ ] Control Plane Revisions

cywang1905:
fix https://github.com/istio/istio/issues/35939
Today, the remote profile will install an istiod server in the remote cluster which will be used for CA and webhook injection for workloads in that cluster. Service discovery, however, will be directed to the control plane in the primary cluster.
Remote pilot sets istiodHost to "istiod-remote" for envoy, but the real remote pilot address for sds connection is "istiod" or "istiod-{{ .Values.revision }}".The problem is "istiod-{{ .values.revison }}" is not in the default dns names and will cause server certificate verifying problem when initiating a remote CSR request.

### DIFF
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -960,7 +960,14 @@ func (s *Server) initIstiodCerts(args *PilotArgs, host string) error {
 
 	// The first is the recommended one, also used by Apiserver for webhooks.
 	// add a few known hostnames
-	for _, altName := range []string{"istiod", "istiod-remote", "istio-pilot"} {
+	knownHosts := []string{"istiod", "istiod-remote", "istio-pilot"}
+	// In some conditions, pilot address for sds is different from other xds,
+	// like multi-cluster primary-remote mode with revision.
+	if args.Revision != "" && args.Revision != "default" {
+		knownHosts = append(knownHosts, "istiod"+"-"+args.Revision)
+	}
+
+	for _, altName := range knownHosts {
 		name := fmt.Sprintf("%v.%v.svc", altName, args.Namespace)
 		if name == host || name == customHost {
 			continue


cywang1905:
@googlebot I signed it!

manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
```diff
@@ -151,6 +151,10 @@ spec:
             value: "{{ .Values.global.istiod.enableAnalysis }}"
           - name: CLUSTER_ID
             value: "{{ $.Values.global.multiCluster.clusterName | default `Kubernetes` }}"
+{{- if not (eq .Values.global.remotePilotAddress "") }}
+          - name: ISTIOD_CUSTOM_HOST
+            value: istiod{{- if not (eq .Values.revision "") }}-{{ .Values.revision }}{{- end }}
```
hzxuzhonghu, line 156: donot we have same value in L149, and it is already signed in the certificate?

cywang1905: In this multi-cluster scenario, remote cluster normally has different host with primary cluster, so it's necessary to add customHost to server's dns names. With the bootstrap logic of pilot, this value is setted by env ISTIOD_CUSTOM_HOST.
Also, I think ISTIOD_ADDR in L149 is not used in code, so ISTIOD_CUSTOM_HOST is necessary here.

manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
```diff
@@ -151,6 +151,10 @@ spec:
             value: "{{ .Values.global.istiod.enableAnalysis }}"
           - name: CLUSTER_ID
             value: "{{ $.Values.global.multiCluster.clusterName | default `Kubernetes` }}"
+{{- if not (eq .Values.global.remotePilotAddress "") }}
+          - name: ISTIOD_CUSTOM_HOST
+            value: istiod{{- if not (eq .Values.revision "") }}-{{ .Values.revision }}{{- end }}.{{ .Release.Namespace }}.svc
```
howardjohn, line 156: I am a bit confused. `remotePilotAddress` can be anything, like `istiod.example.com`. How does adding `istio-....` fix it?

howardjohn (changes requested):

manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
cywang1905: Thanks for your suggestion. I realize that it is not good to overwrite the env ISTIOD_CUSTOM_HOST. And I fix this problem through a new way.

cywang1905:
@howardjohn In multi-cluster primary-remote mode, remote envoys connect to remote pilot for sds and connect to primary pilot for other xds. The root cause of this problem is Remote pilot sets istiodHost to "istiod-remote" for envoy, but the real remote pilot address for sds connection is "istiod" or "istiod-{{ .Values.revision }}". So it is better to add "istiod-{{ .values.revison }}" to default dns names if revision is used.

howardjohn:
> @howardjohn In multi-cluster primary-remote mode, remote envoys connect to remote pilot for sds and connect to primary pilot for other xds. The root cause of this problem is Remote pilot sets istiodHost to "istiod-remote" for envoy, but the real remote pilot address for sds connection is "istiod" or "istiod-{{ .Values.revision }}". So it is better to add "istiod-{{ .values.revison }}" to default dns names if revision is used.

Following what docs? https://istio.io/latest/docs/setup/install/multicluster/primary-remote/ doesn't do this and it seems a bit odd to use this deployment model.

cywang1905:
> > @howardjohn In multi-cluster primary-remote mode, remote envoys connect to remote pilot for sds and connect to primary pilot for other xds. The root cause of this problem is Remote pilot sets istiodHost to "istiod-remote" for envoy, but the real remote pilot address for sds connection is "istiod" or "istiod-{{ .Values.revision }}". So it is better to add "istiod-{{ .values.revison }}" to default dns names if revision is used.
> 
> Following what docs? https://istio.io/latest/docs/setup/install/multicluster/primary-remote/ doesn't do this and it seems a bit odd to use this deployment model.

@howardjohn 

[Install Primary-Remote](https://istio.io/latest/docs/setup/install/multicluster/primary-remote/)
截屏2021-11-11 上午10 36 22

This doc has a tip. So I say "remote envoys connect to remote pilot for sds and connect to primary pilot for other xds."

https://github.com/istio/istio/blob/8aea949750d59a51b322d36ec7cac2c60cea5ac8/manifests/charts/istio-control/istio-discovery/templates/configmap.yaml#L55-L63
Here the discoveryAddress of remote cluster service will be set to "istiod-remote..."

https://github.com/istio/istio/blob/8aea949750d59a51b322d36ec7cac2c60cea5ac8/manifests/charts/istiod-remote/files/injection-template.yaml#L220-L225
Here, CA_ADDR will be "istiod-[revision]..."

cywang1905:
@howardjohn  Also, here is a example of remote cluster in my actual environment. CA_ADDR and discoveryAddress is not the same. So I think it is necessary to add istiod-[revision] to dns names.

```yaml
#sample pod:
- name: CA_ADDR
  value: istiod-1-11-4.istio-system.svc:15012

#configmap istio-1-11-4/istio-system:
  discoveryAddress: istiod-remote.istio-system.svc:15012
```yaml

**Comment by cywang1905** [comment]
@howardjohn @hzxuzhonghu any suggestions? thank you

**Comment by howardjohn** [comment]
cc @stevenctl @frankbu

**Comment by stevenctl** [comment]
I think I'm a bit confused on how this is different from a single-cluster instance connecting to a revisioned control plane for SDS or injection..

**Comment by cywang1905** [comment]
> I think I'm a bit confused on how this is different from a single-cluster instance connecting to a revisioned control plane for SDS or injection..

@stevenctl  Let's assume that istiod revision is 1-11-4, in single-cluster condition, CA_ADDR and discoveryAddress are the same, "istiod-1-11-4.istio-system.svc:15012". But in primary-remote cluster condition, the CA_ADDR of the remote cluster is "istiod-1-11-4.istio-system.svc:15012",  discoveryAddress is "istiod-remote.istio-system.svc:15012". The code of initIstiodCerts cannot ensure revison is in the host list.
So https://github.com/istio/istio/issues/35939 reports the problem that 
`failed to warm certificate: failed to generate workload certificate: create certificate: rpc error: code = Unavailble desc = connection error: desc = "transport: authentication handshake failed: x509: certificate is valid for istiod-remote.istio-system.svc, istiod.istio-system.svc, istio-pilot.istio-system.svc, not istiod-1-11-4.istio-system.svc"`

**Comment by cywang1905** [comment]
@stevenctl hi, have I made myself clear?

**Comment by stevenctl** [comment]
```
┌───Primary───┐   ┌────Remote───┐
│             │   │             │
│   istiod◄───┼─┐ │    istiod   │
│     ▲       │ │ │       ▲     │
│ SDS + xDS   │xDS│       │     │
│     │       │ │ │      SDS    │
│     │       │ └─┼────── ┼     │
│   proxy     │   │     proxy   │
└─────────────┘   └─────────────┘
```


So the proxy in cluster `Primary` is likely using CA_ADDR=istiod-{rev}. I'm surprised that would work, but this wouldn't. It seems like you've installed using `--revision` on both of the clusters. I will try to repro this.

**Comment by cywang1905** [comment]
@stevenctl Yes, use `--revision` for both clusters. Then, the ingress gateway in the remote cluster to will report an error when initiating a CSR request, like 
`failed to warm certificate: failed to generate workload certificate: create certificate: rpc error: code = Unavailble desc = connection error: desc = "transport: authentication handshake failed: x509: certificate is valid for istiod-remote.istio-system.svc, istiod.istio-system.svc, istio-pilot.istio-system.svc, not istiod-1-11-4.istio-system.svc"`

**Comment by cywang1905** [comment]
pilot/pkg/bootstrap/server.go

```go
	// Parse and validate Istiod Address.
	istiodHost, _, err := e.GetDiscoveryAddress()
	if err != nil {
		return nil, err
	}

	// Create Istiod certs and setup watches.
	if err := s.initIstiodCerts(args, string(istiodHost)); err != nil {
		return nil, err
	}
```

This code make sure discoveryAddress will be valid for csr. The discoveryAddress and CA_ADDR address of primary cluster is the same, so it does not have any CSR error. But they are not the same for remote cluster. The CA_ADDR of remote cluster is "istiod-revison", discoveryAddress is "istiod-remote". 

```go
for _, altName := range []string{"istiod", "istiod-remote", "istio-pilot"} {
		name := fmt.Sprintf("%v.%v.svc", altName, args.Namespace)
		if name == host || name == customHost {
			continue
		}
		s.dnsNames = append(s.dnsNames, name)
	}
```

The previous code of function initIstiodCerts as above only make sure "istiod", "istiod-remote", "istio-pilot" are valid. So "istiod-revison" is invalid now. 
What I want to do is to ensure that "istiod-revision" is also valid when using revison. If you don't use revision, it will not be added to dnsNames to keep security. I think this is a proper fix, and no additional risks are introduced.

stevenctl:
Your fix does make sense, but what _doesn't_ make sense to me is the fact that in the _primary_ cluster we are also using `istiod-rev` as the CA_ADDR and not having issues.

stevenctl:
So the main issue is `discoveryAddress` is NOT always the desired value for `caAddress`. 

We can fix this either by:
* Simply always allowing `istiod-{rev}`; seems safe to me, open to opinions @howardjohn @Monkeyanator 

OR

* Providing a new way to say "this discovery address is also valid". This is was the purpose of `ISTIO_CUSTOM_HOST`. I think your original solution is actually _almost_ valid. We'd only add the custom host if `remoteDiscoveryAddress && rev`. It's also possible that this is only one of multiple "CUSTOM_HOST" values but I do think it's a safe change.

stevenctl (approved):
This LGTM. I added a hold, please take a look @Monkeyanator @hzxuzhonghu @howardjohn

hzxuzhonghu (approved):

cywang1905:
can anyone help to set the lable 'release-notes-none' ?
@howardjohn @hzxuzhonghu @stevenctl

howardjohn (approved):

cywang1905:
could you help to remove the lable 'do-not-merge/hold ' ? thx~
@howardjohn @hzxuzhonghu @stevenctl

hzxuzhonghu:
done

PR #35971, opened 2021-11-09
[multi cluster]use node internalIP if no external IP found when using node port mode

fatalc:

In multi cluster and using Nodeport mode for cross cluster traffic,EW gateway's `.Attributes.ClusterExternalAddresses` using node's ExternalIP address and svc node port.
While In kubernetes cluster which setup by kubeadm has nodes with Internal Address only(no ExternalIP). Those nodes doesn't cached by controller. This gateways has a empty ClusterExternalAddresses, the Istio multi cluster traffic doesn't work.

Could we consider using node InternalIP as an alternative if no ExternalIP found.

This change may affect the existing functions, if  a feature-gate needed？

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -639,7 +639,16 @@ func (c *Controller) onNodeEvent(obj interface{}, event model.Event) error {
 			}
 		}
 		if k8sNode.address == "" {
-			return nil
+			// use internal address if available
+			for _, address := range node.Status.Addresses {
+				if address.Type == v1.NodeInternalIP && address.Address != "" {
+					k8sNode.address = address.Address
+					break
+				}
+			}
+			if k8sNode.address == "" {
+				return nil
+			}
 		}
 
 		c.Lock()


fatalc:
@googlebot I signed it!

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -639,7 +639,16 @@ func (c *Controller) onNodeEvent(obj interface{}, event model.Event) error {
 			}
 		}
 		if k8sNode.address == "" {
-			return nil
+			// use internal address if available
+			for _, address := range node.Status.Addresses {
+				if address.Type == v1.NodeInternalIP && address.Address != "" {
```
howardjohn, line 644: Can you explain why _internal_ address makes sense? We are trying to access this node externally, so I would expect internal to not work?

fatalc: Just an alt for no External IP nodes, node's internalIP always used for "out of cluster" communication under the **private cloud** architecture.

howardjohn: Per https://kubernetes.io/docs/concepts/architecture/nodes/#addresses, this field should be for _reaching it within the cluster_. In this case, we explicitly want to reach it from OUT of the cluster, so this feels wrong to me.

PR #35972, opened 2021-11-09
pilot: support access token for apache skywalking

### LINKED ISSUE #35968: The accessToken of skywalking cannot be sent to the sidecar
### Bug Description

configmap: istio
```
apiVersion: v1
data:
  mesh: |-
    extensionProviders:
    - name: envoy.tracers.skywalking
      skywalking:
        port: 11800
        service: xxxx
        accessToken: xxxxx        # <--- Here!  When I dump config from Envoy, I can not find it.
    rootNamespace: istio-system
    trustDomain: cluster.local
  meshNetworks: 'networks: {}'
kind: ConfigMap
```

### Version

```prose
$ istioctl version
client version: 1.11.2
control plane version: 1.11.2
data plane version: 1.11.2 (15 proxies)
```

### Additional Information

_No response_

Shikugawa:
Close #35968 
This PR adds support for access token for Apache SkyWalking, which is not implemented on the previous implementation.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [x] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/tracing.go
+++ b/pilot/pkg/networking/core/v1alpha3/tracing.go
@@ -138,6 +138,14 @@ func configureFromProviderConfig(pushCtx *model.PushContext, meta *model.NodeMet
 	case *meshconfig.MeshConfig_ExtensionProvider_Skywalking:
 		tracing, err = buildHCMTracing(pushCtx, providerCfg.Name, provider.Skywalking.Service,
 			provider.Skywalking.Port, 0, func(clusterName string) (*anypb.Any, error) {
+				clientConfig := &tracingcfg.ClientConfig{}
+
+				if provider.Skywalking.AccessToken != "" {
+					clientConfig.BackendTokenSpecifier = &tracingcfg.ClientConfig_BackendToken{
+						BackendToken: provider.Skywalking.AccessToken,
+					}
+				}
+
 				s := &tracingcfg.SkyWalkingConfig{
 					GrpcService: &envoy_config_core_v3.GrpcService{
 						TargetSpecifier: &envoy_config_core_v3.GrpcService_EnvoyGrpc_{
@@ -146,6 +154,7 @@ func configureFromProviderConfig(pushCtx *model.PushContext, meta *model.NodeMet
 							},
 						},
 					},
+					ClientConfig: clientConfig,
 				}
 
 				return anypb.New(s)
--- a/pilot/pkg/networking/core/v1alpha3/tracing_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/tracing_test.go
@@ -102,9 +102,9 @@ func TestConfigureTracing(t *testing.T) {
 		},
 		{
 			name:      "basic config (with skywalking provicer)",
-			inSpec:    fakeTracingSpec(fakeSkywalking(), 99.999, false),
+			inSpec:    fakeTracingSpec(fakeSkywalking("token"), 99.999, false),
 			opts:      fakeOptsOnlySkywalkingTelemetryAPI(),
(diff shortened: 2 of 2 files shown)


pilot/pkg/networking/core/v1alpha3/tracing.go
```diff
@@ -138,6 +138,14 @@ func configureFromProviderConfig(pushCtx *model.PushContext, meta *model.NodeMet
 	case *meshconfig.MeshConfig_ExtensionProvider_Skywalking:
 		tracing, err = buildHCMTracing(pushCtx, providerCfg.Name, provider.Skywalking.Service,
 			provider.Skywalking.Port, 0, func(clusterName string) (*anypb.Any, error) {
+				clientConfig := &tracingcfg.ClientConfig{}
+
+				if provider.Skywalking.AccessToken != "" {
+					clientConfig.BackendTokenSpecifier = &tracingcfg.ClientConfig_BackendToken{
+						BackendToken: provider.Skywalking.AccessToken,
```
howardjohn, line 145: Is this value obfuscated in /config_dump?

I guess istiod debug interface will also show it.... holding access tokens in mesh config seems like a bad idea.

howardjohn (approved):
LGTM since this implements the API we approved, but i have concerns about the API...

pilot/pkg/networking/core/v1alpha3/tracing.go
hzxuzhonghu: Agree, it has security issue to expose token in configmap

howardjohn (changes requested):
after thinking about it more I am strongly against putting security critical information in mesh config. We need to revisit the API

Shikugawa:
@howardjohn I agree with it. I think that we need to add SDS support on Envoy side to resolve this problem... Do you have another idea?

howardjohn:
I am not against istiod sending the access token over some XDS that isn't
SDS but:
- it should be redacted in envoy and istiod config dump
- it should not be in mesh config which can be accessed by many insecure
ways

howardjohn:
SDS seems like a reasonable way to get there although I think it would need
some tweaks, at least on istiod, as it's fairly TLS cert oriented

Shikugawa:
For example, lightstep tracer supports datasource to extract atoken with more secure way than hardcoded. (It seems istiod doesn't use this feature to alter using `access_token_file` which is deprecated way)
- https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/trace/v3/lightstep.proto#L52
- https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/v1alpha3/tracing.go#L121

I think it is better to use similar way, using datasource for access token.

howardjohn:
My concern with DataSource is you need either a file or env var then? But its pretty hard to get a file onto ~all pods. SDS would definitely be preferred I think if possible.

Shikugawa:
@howardjohn We added a support for the Envoy side in the above PR. But, we need GenericSecret SDS support is required in istiod to enable this feature. My idea is to be able to handle Tokens as Kubernetes Secret and send them from Istiod to Envoy via SDS. This way, I think the change in Istio's API will be small and the implementation will be simple. What do you think?

howardjohn:
I think that is the right direction. However I am not sure it will be so
easy.. I think the challenge will be figuring out authentication; which
sidecars do we allow access to which Secrets. We definitely cannot just
give access to any secret or they could request any secret in the cluster
which is insecure. Today we only have this problem solved for Gateway.

One approach is we can just allow access to the secret referenced in the
mesh config. However, ideally we solve it more generally - there is a
desire to have sidecars also have access to TLS certs for various cases
(today Destination Rule to originate TLS, there is also a open PR to add
TLS termination). One idea is to use ReferencePolicy from gateway-api but
it hasn't been clearly defined.

There was a doc about this 'auth for Istiod SDS' but it's a bit dated

Issue #34995, opened 2021-09-01
Implement Endpointslice FQDN addresses


howardjohn:
### Bug Description

```yaml
addressType: FQDN
apiVersion: discovery.k8s.io/v1
endpoints:
- addresses:
  - httpbin.org
  conditions:
    ready: true
    serving: true
    terminating: false
kind: EndpointSlice
metadata:
  labels:
    kubernetes.io/service-name: httpbin
  name: httpbin-manualy
  namespace: default
ports:
- name: http
  port: 80
  protocol: TCP
---
apiVersion: v1
kind: Service
metadata:
  name: httpbin
  namespace: default
spec:
  clusterIP: None
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 80
```

With `set env deployment/istiod PILOT_USE_ENDPOINT_SLICE=true`

Leads to 
```
2021-09-01T15:30:58.882375Z     warn    ads     ADS:EDS: ACK ERROR istio-ingressgateway-6b48bbf7d6-q4m28.istio-system-1 Internal:malformed IP address: httpbin.org. Consider setting resolver_name or setting cluster type to 'STRICT_DNS' or 'LOGICAL_DNS'
```

This actually works without PILOT_USE_ENDPOINT_SLICE because coredns generates a CNAME but Istiod doesn't see the endpoint at all. So it works just like any other headless. This would NOT work for TCP, since it would expect to know the IPs.

### Version

```prose
master
```

howardjohn:
We should figure out what we actually need to do in this case though. Do we treat it as strict-dns? Does kubernetes allow mixed FQDN and IP and what do we do in that case (I think envoy is ok with IPs in strict DNS cluster although it's not ideal)? 

Can this happen with services with selector, headless and cluster ip, etc. What does coredns do if we have multiple fqdn address not just one?

This is looking more complex than I thought... Whatever we do is better than current state though

dprotaso:
Probably worth revisiting this as it's been a year and new k8s versions are out.

Also related
https://github.com/kubernetes-sigs/gateway-api/issues/1718

howardjohn:
What exactly does it mean to support it? Afaik there is no defined meaning for fqdn?

dprotaso:
From my point of view it’s analogous to Service with type ExternalName but
it can be RBAC’d so it doesn’t let users craft services that can compromise
the system

https://github.com/projectcontour/contour/security/advisories/GHSA-5ph6-qq5x-7jwc

[...]

PR #35990, opened 2021-11-09
ignore FQDN endpointslice

stevenctl:
If I understand the comments in https://github.com/kubernetes/kubernetes/issues/105172 and #34995 , we should be able to safely ignore these EndpointSlices for now, so `PILOT_USE_ENDPOINT_SLICE` doesn't cause NACKs.

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/endpointslice.go
+++ b/pilot/pkg/serviceregistry/kube/controller/endpointslice.go
@@ -149,6 +149,10 @@ func serviceNameForEndpointSlice(labels map[string]string) string {
 func (esc *endpointSliceController) sliceServiceInstances(c *Controller, slice interface{}, proxy *model.Proxy) []*model.ServiceInstance {
 	var out []*model.ServiceInstance
 	ep := wrapEndpointSlice(slice)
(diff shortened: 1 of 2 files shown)


pilot/pkg/serviceregistry/kube/controller/endpointslice.go
```diff
@@ -149,6 +149,10 @@ func serviceNameForEndpointSlice(labels map[string]string) string {
 func (esc *endpointSliceController) sliceServiceInstances(c *Controller, slice interface{}, proxy *model.Proxy) []*model.ServiceInstance {
 	var out []*model.ServiceInstance
 	ep := wrapEndpointSlice(slice)
+	if ep.AddressType() == v1.AddressTypeFQDN {
+		// TODO support FQDN endpointslice
```
howardjohn, line 153:
```suggestion
		// TODO(https://github.com/istio/istio/issues/34995) support FQDN endpointslice
```

howardjohn (approved):
Would be nice to have a unit test. Definitely better than current state!

howardjohn (approved):

PR #35994, opened 2021-11-10
Enable building for base images

TrevorTaoARM:
Now for DOCKER_V2_BUILDER, it does not support the explicit building for
base images, for example, when building with
make dockerx.app_sidecar_base_ubuntu_xenial
it would show some error information for the wrong VM_IMAGE_NAME and
VM_IMAGE_VERSION, now we just fix it here.

And add other base images in the istio-docker.mk in the default list too.

### DIFF
--- a/tools/docker-builder/main.go
+++ b/tools/docker-builder/main.go
@@ -287,6 +287,10 @@ func vmImageName(target string) string {
 		// Not a VM
 		return ""
 	}
+	if strings.HasPrefix(target, "app_sidecar_base") {
+		return strings.Split(target, "_")[3]
+	}
+
 	return strings.Split(target, "_")[2]
 }
 
@@ -295,6 +299,10 @@ func vmImageVersion(target string) string {
 		// Not a VM
 		return ""
 	}
+	if strings.HasPrefix(target, "app_sidecar_base") {
+		return strings.Split(target, "_")[4]
+	}
+
 	return strings.Split(target, "_")[3]
 }
 


tools/docker-builder/types.go
```diff
@@ -93,6 +93,15 @@ func DefaultArgs() Args {
 		"istioctl",
 		"operator",
 		"install-cni",
+		"base",
```
howardjohn, line 96: The rest of the change is great - thanks for the fix - but I don't think we want to build all of these every time. Currently we are only rebuilding these when there are CVEs in the base image since they are slow + make a bunch of external calls. There is also a `distroless` one fwiw

TrevorTaoARM: OK, I think it can be removed from the list for what you said.

TrevorTaoARM: Thanks.

tools/docker-builder/types.go
```diff
@@ -94,6 +94,14 @@ func DefaultArgs() Args {
 		"operator",
 		"install-cni",
 
+		"app_sidecar_base_ubuntu_xenial",
```
howardjohn, line 97: sorry my comment about base wasn't clear - I meant all of the base ones including these.

That will give matching behavior in the old logic and new logic

TrevorTaoARM: They are removed now. Thanks.

howardjohn (approved):

ericvn (approved):

PR #36016, opened 2021-11-11
Remove useless shell script in makefile

SpecialYang:
Remove useless shell script `set_up_env.sh` in makefile. When we use make command with mode of `BUILD_WITH_CONTAINER=1`, the shell script `set_up_env.sh` will be executed twice. The first execution of `set_up_env.sh` in makefile does no effect and is unnecessary.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,11 +32,9 @@ export BUILD_WITH_CONTAINER ?= 0
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
 
-# An export free of arugments in a Makefile places all variables in the Makefile into the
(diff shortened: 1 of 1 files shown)


ericvn (changes requested):
This file is copied from stio.io/common-files/files/Makefile. A change would need to be made there which will be copied to all repos. A change here would be overridden with the next common-files update.

PR #36041, opened 2021-11-12
Fix docker.save tar targets

sergii-ssh:
Similar to https://github.com/istio/istio/pull/35761 if building non-dockerX targets with `DOCKER_BUILD_VARIANTS=debug distroless default` builder attempts to save `<REPO>:<TAG>-default` tar which doesn't exist


ericvn:
Verifying this works with `DOCKER_V2_BUILDER=false make docker.save`.

ericvn (approved):

howardjohn (approved):

ericvn:
Note: https://github.com/istio/istio/pull/36061 is needed to fix this with current commits which no longer use the Makefile logic unless `DOCKER_V2_BUILDER=false`

PR #36042, opened 2021-11-12
Rebuilding docker-builder if code changed

TrevorTaoARM:
Rebuilding docker-builder if code changed, especially for
development phase.


ericvn (approved):

howardjohn (commented):
Nice! I was thinking the same would be useful and did not come up with such an elegant solution :+1:

PR #36043, opened 2021-11-12
do not generate debug output for ndsz if dns is not enabled

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [ X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/debug.go
+++ b/pilot/pkg/xds/debug.go
@@ -782,7 +782,7 @@ func (s *DiscoveryServer) List(w http.ResponseWriter, req *http.Request) {
 // It is mapped to /debug/ndsz on the monitor port (15014).
 func (s *DiscoveryServer) ndsz(w http.ResponseWriter, req *http.Request) {
 	proxyID, con := s.getDebugConnection(req)
-	if !con.proxy.Metadata.DNSCapture {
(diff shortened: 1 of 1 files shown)


pilot/pkg/xds/debug.go
```diff
@@ -781,11 +781,16 @@ func (s *DiscoveryServer) List(w http.ResponseWriter, req *http.Request) {
 // ndsz implements a status and debug interface for NDS.
 // It is mapped to /debug/ndsz on the monitor port (15014).
 func (s *DiscoveryServer) ndsz(w http.ResponseWriter, req *http.Request) {
+	proxyID, con := s.getDebugConnection(req)
+	if !con.proxy.Metadata.DNSCapture {
```
hzxuzhonghu, line 785: should check if con == nil

hzxuzhonghu: and con.proxy ==nil

hzxuzhonghu (approved):
lgtm, con.proxy must not be nil guaranteed by `getProxyConnection`

hzxuzhonghu (commented):
/hold

@ramaraochavali please move down after existing con == nil

ramaraochavali:
Moved

Issue #36052, opened 2021-11-12
`make docker` related targets now failing on a Mac 


ericvn:
### Bug Description

```
 make docker
./tools/docker
./tools/docker: line 34: out/darwin_amd64/docker-builder: cannot execute binary file: Exec format error
```

This was working for me last week.

### Version

```prose
master branch
```

ericvn:
Broken by https://github.com/istio/istio/pull/35965

PR #36055, opened 2021-11-12
Fix `make docker` related target on MacOS

ericvn:

Fixes https://github.com/istio/istio/issues/36052

A prior PR, https://github.com/istio/istio/pull/35965, updated the `docker` script to use the TARGET_OS and TARGET_ARCH env vars, however these point to the host's OS/ARCH and not that running locally in the container (or locally if not in the container). This causes the `docker-builder` to be built as a darwin executable which isn't able to be run locally in the container. This PR changes the script to use LOCAL_OUT which is the architecture where we are currently running in and should be used for build related artifacts.

### DIFF
--- a/tools/docker
+++ b/tools/docker
@@ -23,14 +23,12 @@ set -eux
 # LOCAL_OUT is set when runing in a Makefile and set to current architecture (not host)
 if [[ ! -z "${LOCAL_OUT}" ]]; then
   OS_ARCH=${LOCAL_OUT##*/}
-  LOCAL_TARGET_ARCH=${OS_ARCH##*_}
   LOCAL_TARGET_OS=${OS_ARCH%_*}
+  LOCAL_TARGET_ARCH=${OS_ARCH##*_}
+else
+  LOCAL_TARGET_OS=${TARGET_OS:-linux}
+  LOCAL_TARGET_ARCH=${TARGET_ARCH:-amd64}
 fi
-
-LOCAL_TARGET_OS=${LOCAL_TARGET_OS:-linux}
-LOCAL_TARGET_ARCH=${LOCAL_TARGET_ARCH:-amd64}
-LOCAL_GOOS_LOCAL=${LOCAL_TARGET_OS}
-LOCAL_GOARCH_LOCAL=${LOCAL_TARGET_ARCH}
 
 cd "${ROOT}"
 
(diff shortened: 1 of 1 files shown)


ericvn:
This should be ready to go. Hopefully @howardjohn or another can verify this doesn't break linux.

tools/docker
```diff
@@ -20,23 +20,18 @@ ROOT=$(dirname "$WD")
...
-TARGET_ARCH=${TARGET_ARCH:-amd64}
-GOOS_LOCAL=${TARGET_OS}
-GOARCH_LOCAL=${TARGET_ARCH}
-
 cd "${ROOT}"
 
 export REPO_ROOT="${ROOT}"
 
-if [[ -f out/${TARGET_OS}_${TARGET_ARCH}/docker-builder ]]; then
+if [[ -f ${LOCAL_OUT}/docker-builder ]]; then
```
howardjohn, line 27: can we keep the defaults? I would prefer this to be runnable directly since IMO the UX is better than the make variables.

Or is it not reasonable since we have so much logic to set up the vars? Perhaps we can do `source ./common/scripts/setup_env.sh envfile` or similar if they are not set?

ericvn: Problem is that the TARGET variables point to the HOST environment and in the case of using the build container on a Mac (any non-linux host), those values are Darwin/other values, and we need to build/run the docker-builder in the linux container. 

I could try looking to see if the LOCAL_OUT is set and use that (would have to parse to get OS/ARCH) and if not set fall back to the linux values. I'm not sure about moving LOCAL_OUT out to the setup_env.sh (where we set the TARGET*) and calling that, but that might actually be the best. Thoughts?

howardjohn: Ah its not already there, is it...

It could be tricky since it depends on IN_BUILD_CONTAINER. maybe it would work though... 

Either way works fine for me.

ericvn: @howardjohn I tried to simply set TARGET_ARCH and TARGET_OS from the LOCAL_OUT. This resulted in the follow-on `make` having an issue with it thinking it needed to use the `linux` side of the `out` directory instead of the `darwin` side as the TARGET* env vars were overridden. I suspect that trying to run setup_env again will have the same effect. This isn't necessarily an elegant solution, but does set the variables to build docker-build and not affect the makefile call.

ericvn:
This is about as clean as I can make it to use the LOCAL_OUT if running in a build container on a different host, and still allow the TARGET_OS and TARGET_ARCH to be set locally and use those. As noted, we have to use a different variable in the script so as not to override the TARGET* env vars that might have been set locally in the build container.

howardjohn (approved):

Issue #36059, opened 2021-11-12
`make docker.save` with DOCKER_V2_BUILDER have default in the tar file name


ericvn:
### Bug Description

Running `make docker.save` on a current commit using docker-builder has `default` in the tar file names:
```
2021-11-12T17:03:44.382315Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/app_sidecar_ubuntu_xenial-default.tar
2021-11-12T17:03:50.374347Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/app_sidecar_ubuntu_bionic-default.tar
2021-11-12T17:03:55.146230Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/app_sidecar_ubuntu_focal-default.tar
2021-11-12T17:04:00.715592Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/app_sidecar_debian_9-default.tar
2021-11-12T17:04:06.364810Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/app_sidecar_debian_10-default.tar
2021-11-12T17:04:12.139635Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/app_sidecar_centos_8-default.tar
2021-11-12T17:04:19.661861Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/pilot-default.tar
2021-11-12T17:04:22.115334Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/operator-default.tar
2021-11-12T17:04:24.376854Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/istioctl-default.tar
2021-11-12T17:04:26.632515Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/install-cni-default.tar
2021-11-12T17:04:29.682106Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/app_sidecar_centos_7-default.tar
2021-11-12T17:04:36.623750Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/proxyv2-default.tar
2021-11-12T17:04:40.072065Z	info	Running command: gzip --fast --force /work/out/linux_amd64/release/docker/app-default.tar
```

This differs from the behavior in the release-1.12 branch or pre V2 where the `-default` was removed.

### Version

```prose
current master branch
```

PR #36061, opened 2021-11-12
Fix docker.save having `-default` in tar name

ericvn:

Fixes https://github.com/istio/istio/issues/36059

### DIFF
--- a/tools/docker-builder/main.go
+++ b/tools/docker-builder/main.go
@@ -246,13 +246,13 @@ func ConstructBakeFile(a Args) (map[string]string, error) {
 				t.Outputs = []string{"type=registry"}
 			} else if args.Save {
 				n := target
-				if variant != "" {
+				if variant != "" && variant != DefaultVariant { // For default variant, we do not add it.
 					n += "-" + variant
 				}
 
 				tarFiles[n] = ""
 				if variant == PrimaryVariant && hasDoubleDefault {
-					tarFiles[n] = target
+					tarFiles[n] = target + "-" + variant
 				}
 				t.Outputs = []string{"type=docker,dest=" + filepath.Join(testenv.LocalOut, "release", "docker", n+".tar")}
 			} else {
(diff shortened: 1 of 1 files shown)


tools/docker-builder/main.go
```diff
@@ -246,13 +246,13 @@ func ConstructBakeFile(a Args) (map[string]string, error) {
 				t.Outputs = []string{"type=registry"}
 			} else if args.Save {
 				n := target
-				if variant != "" {
+				if variant != "" && variant != DefaultVariant { // For default variant, we do not add it.
```
ericvn, line 249: This line removes the `-default`.

tools/docker-builder/main.go
```diff
@@ -246,13 +246,13 @@ func ConstructBakeFile(a Args) (map[string]string, error) {
...
 					n += "-" + variant
 				}
 
 				tarFiles[n] = ""
 				if variant == PrimaryVariant && hasDoubleDefault {
-					tarFiles[n] = target
+					tarFiles[n] = target + "-" + variant
```
ericvn, line 255: We append the tag if doubled up. Using the alias to create another tar file. This line might not be needed or conflict with the alias elsewhere.

ericvn: Looks like this change can be removed.

ericvn:
What should `./tools/docker --save --variants=debug,distroless` generate for tar files. Should it generated bug, distroless and empty variant files? default,debug,distroless does generate all 3. It seems debug.distroless only creates debug and distroless and no empty variants.

jacob-delgado (approved):

howardjohn:
> What should ./tools/docker --save --variants=debug,distroless generate for tar files. Should it generate debug, distroless and empty variant files? default,debug,distroless does generate all 3. It seems debug.distroless only creates debug and distroless and no empty variants. Testing prior to this change, I don't get the empty variant so this change doesn't change that.

It should generate debug and distroless ONLY.

PR #36062, opened 2021-11-12
Istioctl x describe add PeerAuthentication debug info

liwenhao0810:
This PR is to add PeerAuthentication debug info into `istioctl x describe pod` command. 
1. If client has access to root namespace, debug info will includes all applied PA to the specified pod the active mTLS mode.
2. If client doesn't have access to root namespace, debug info will only includes all applied PA in the workload's namespace.

Related design doc: [link](https://docs.google.com/document/d/1apxj7duekYneZL3BCczCfQM0cpTJARdsYswNC7EIG8c)

### DIFF
--- a/istioctl/cmd/describe.go
+++ b/istioctl/cmd/describe.go
@@ -1207,11 +1207,9 @@ func describePeerAuthentication(writer io.Writer, configClient istioclient.Inter
 		return fmt.Errorf("failed to fetch workload namespace PeerAuthentication: %v", err)
 	}
 
-	hasRootNamespaceAccess := true
 	rootPAList, err := configClient.SecurityV1beta1().PeerAuthentications(istioNamespace).List(context.Background(), metav1.ListOptions{})
 	if err != nil {
-		hasRootNamespaceAccess = false
-		fmt.Fprintf(writer, "Cannot analyze active mTLS mode since failed to fetch root namespace PeerAuthentication: %v:\n", err)
+		return fmt.Errorf("failed to fetch root namespace PeerAuthentication: %v", err)
 	}
 
 	allPAs := append(rootPAList.Items, workloadPAList.Items...)
@@ -1226,10 +1224,8 @@ func describePeerAuthentication(writer io.Writer, configClient istioclient.Inter
 	matchedPA := findMatchedConfigs(podsLabels, cfgs)
 	printConfigs(writer, matchedPA)
 
-	if hasRootNamespaceAccess {
-		activePA := authnv1beta1.ComposePeerAuthentication(istioNamespace, matchedPA)
-		printPeerAuthentication(writer, activePA)
-	}
+	activePA := authnv1beta1.ComposePeerAuthentication(istioNamespace, matchedPA)
+	printPeerAuthentication(writer, activePA)
 
 	return nil
 }
@@ -1249,15 +1245,10 @@ func findMatchedConfigs(podsLabels k8s_labels.Set, configs []*config.Config) []*
 
 	for _, cfg := range configs {
 		cfg := cfg
-		spec := cfg.Spec.(Workloader)
-		if spec.GetSelector() == nil || len(spec.GetSelector().MatchLabels) == 0 {
+		labels := cfg.Spec.(Workloader).GetSelector().GetMatchLabels()
+		selector := k8s_labels.SelectorFromSet(labels)
+		if selector.Matches(podsLabels) {
 			cfgs = append(cfgs, cfg)
-		} else {
-			selector := spec.GetSelector().GetMatchLabels()
-			slr := k8s_labels.SelectorFromSet(selector)
-			if slr.Matches(podsLabels) {
-				cfgs = append(cfgs, cfg)
-			}
 		}
 	}
 
--- a/istioctl/cmd/describe.go
+++ b/istioctl/cmd/describe.go
@@ -151,7 +151,7 @@ the configuration objects that affect that pod.`,
 
 			// render PeerAuthentication info
 			fmt.Fprintf(writer, "--------------------\n")
-			err = describePeerAuthentication(writer, kubeClient, configClient, istioNamespace, ns, k8s_labels.Set(pod.ObjectMeta.Labels))
+			err = describePeerAuthentication(writer, kubeClient, configClient, ns, k8s_labels.Set(pod.ObjectMeta.Labels))
 			if err != nil {
 				return err
 			}
@@ -1203,8 +1203,7 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
 // describePeerAuthentication fetches all PeerAuthentication in workload and root namespace.
 // It lists the ones applied to the pod, and the current active mTLS mode.
 // When the client doesn't have access to root namespace, it will only show workload namespace Peerauthentications.
-func describePeerAuthentication(writer io.Writer, kubeClient kube.ExtendedClient, configClient istioclient.Interface, istioNamespace, workloadNamespace string, podsLabels k8s_labels.Set) error { // nolint: lll
-	// get root namespace
+func describePeerAuthentication(writer io.Writer, kubeClient kube.ExtendedClient, configClient istioclient.Interface, workloadNamespace string, podsLabels k8s_labels.Set) error { // nolint: lll
 	meshCfg, err := getMeshConfig(kubeClient)
 	if err != nil {
 		return fmt.Errorf("failed to fetch mesh config: %v", err)
--- a/istioctl/cmd/describe.go
+++ b/istioctl/cmd/describe.go
@@ -1215,9 +1215,6 @@ func describePeerAuthentication(writer io.Writer, configClient istioclient.Inter
 	}
 
 	allPAs := append(rootPAList.Items, workloadPAList.Items...)
-	if len(allPAs) == 0 {
-		return nil
-	}
 
 	var cfgs []*config.Config
 	for _, pa := range allPAs {
@@ -1243,7 +1240,8 @@ type Workloader interface {
 }
 
 // findMatchedConfigs should filter out unrelated configs that are not matched given podsLabels.
-// Configs passed into this method should only contains workload's namespaces configs
+// When the config has no selector labels, this method will treat it as qualified namespace level
+// config. So configs passed into this method should only contains workload's namespaces configs
 // and rootNamespaces configs, caller should be responsible for controlling configs passed
 // in.
 func findMatchedConfigs(podsLabels k8s_labels.Set, configs []*config.Config) []*config.Config {
@@ -1268,12 +1266,16 @@ func findMatchedConfigs(podsLabels k8s_labels.Set, configs []*config.Config) []*
 
 func printConfigs(writer io.Writer, configs []*config.Config) {
 	if len(configs) == 0 {
+		fmt.Fprintf(writer, "No config found.\n")
 		return
 	}
 	fmt.Fprintf(writer, "Applied %s:\n", configs[0].Meta.GroupVersionKind.Kind)
 	var cfgNames string
-	for _, cfg := range configs {
-		cfgNames += cfg.Meta.Name + "." + cfg.Meta.Namespace + ", "
+	for i, cfg := range configs {
+		cfgNames += cfg.Meta.Name + "." + cfg.Meta.Namespace
+		if i < len(configs)-1 {
+			cfgNames += ", "
+		}
 	}
 	fmt.Fprintf(writer, "   %s\n", cfgNames)
 }
--- a/istioctl/cmd/describe.go
+++ b/istioctl/cmd/describe.go
@@ -36,6 +36,7 @@ import (
 	k8s_labels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 
+	meshconfig "istio.io/api/mesh/v1alpha1"
 	"istio.io/api/networking/v1alpha3"
 	"istio.io/api/security/v1beta1"
 	typev1beta1 "istio.io/api/type/v1beta1"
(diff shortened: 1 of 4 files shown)


istioctl/cmd/describe.go
```diff
@@ -1186,3 +1197,94 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
...
+// When the client doesn't have access to root namespace, it will only show workload namespace Peerauthentications.
+func describePeerAuthentication(writer io.Writer, configClient istioclient.Interface, istioNamespace, workloadNamespace string, pod *v1.Pod, podsLabels k8s_labels.Set) error { // nolint: lll
+	workloadPAList, err := configClient.SecurityV1beta1().PeerAuthentications(workloadNamespace).List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to fetch workload namespace PeerAuthentication: %v", err)
+	}
+
+	hasRootNamespaceAccess := true
+	rootPAList, err := configClient.SecurityV1beta1().PeerAuthentications(istioNamespace).List(context.Background(), metav1.ListOptions{})
+	if err != nil {
```
yangminzhu, line 1212: I think you can just return the error, it seems the only possible case that this would fail is due to some network issues (or very unlikely some RBAC issues), either case you don't need to continue the analyze because you don't have enough information to give the accurate result.

istioctl/cmd/describe.go
```diff
@@ -1186,3 +1197,94 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
...
+		hasRootNamespaceAccess = false
+		fmt.Fprintf(writer, "Cannot analyze active mTLS mode since failed to fetch root namespace PeerAuthentication: %v:\n", err)
+	}
+
+	allPAs := append(rootPAList.Items, workloadPAList.Items...)
+	if len(allPAs) == 0 {
+		return nil
```
yangminzhu, line 1219: I think it would be better to print the default mode (PERMISSIVE as returned by `ComposePeerAuthentication()`) even if there are no policies applied explicitlly.

istioctl/cmd/describe.go
```diff
@@ -1186,3 +1197,94 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
 	}
 	return false, fmt.Errorf("no container %q in pod", containerName)
 }
+
+// describePeerAuthentication fetches all PeerAuthentication in workload and root namespace.
+// It lists the ones applied to the pod, and the current active mTLS mode.
+// When the client doesn't have access to root namespace, it will only show workload namespace Peerauthentications.
+func describePeerAuthentication(writer io.Writer, configClient istioclient.Interface, istioNamespace, workloadNamespace string, pod *v1.Pod, podsLabels k8s_labels.Set) error { // nolint: lll
```
yangminzhu: the argument `pod` is not used in the function.

istioctl/cmd/describe.go
```diff
@@ -1186,3 +1197,94 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
...
+	var cfgs []*config.Config
+
+	for _, cfg := range configs {
+		cfg := cfg
+		spec := cfg.Spec.(Workloader)
+		if spec.GetSelector() == nil || len(spec.GetSelector().MatchLabels) == 0 {
+			cfgs = append(cfgs, cfg)
+		} else {
+			selector := spec.GetSelector().GetMatchLabels()
+			slr := k8s_labels.SelectorFromSet(selector)
```
yangminzhu, line 1259: you do not need the if-else check, `k8s_labels.SelectorFromSet(nil)` will return a selector that matches everything, see the comment of the function.

istioctl/cmd/describe.go
```diff
@@ -1186,3 +1197,94 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
...
+// Workloader is used for matching all configs
+type Workloader interface {
+	GetSelector() *typev1beta1.WorkloadSelector
+}
+
+// findMatchedConfigs should filter out unrelated configs that are not matched given podsLabels.
+// Configs passed into this method should only contains workload's namespaces configs
+// and rootNamespaces configs, caller should be responsible for controlling configs passed
+// in.
+func findMatchedConfigs(podsLabels k8s_labels.Set, configs []*config.Config) []*config.Config {
```
yangminzhu, line 1249: Could you clarify what does the `Configs passed into this method should only contains workload's namespaces configs and rootNamespaces configs` mean? What's the negative cases?

istioctl/cmd/describe.go
```diff
@@ -1186,3 +1197,94 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
...
+}
+
+func printConfigs(writer io.Writer, configs []*config.Config) {
+	if len(configs) == 0 {
+		return
+	}
+	fmt.Fprintf(writer, "Applied %s:\n", configs[0].Meta.GroupVersionKind.Kind)
+	var cfgNames string
+	for _, cfg := range configs {
+		cfgNames += cfg.Meta.Name + "." + cfg.Meta.Namespace + ", "
```
yangminzhu, line 1276: this will always print the `, ` after the last policy

istioctl/cmd/describe.go
```diff
@@ -1186,3 +1197,94 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
...
+	}
+	fmt.Fprintf(writer, "   %s\n", cfgNames)
+}
+
+func printPeerAuthentication(writer io.Writer, pa *v1beta1.PeerAuthentication) {
+	fmt.Fprintf(writer, "Active PeerAuthentication:\n")
+	fmt.Fprintf(writer, "   Workload mTLS: %s\n", pa.Mtls.Mode.String())
```
yangminzhu, line 1283: It's worth to also print the namespace/name of the active policy.

istioctl/cmd/describe.go
liwenhao0810: How about is the user doesn't have root namespace access?

istioctl/cmd/describe.go
liwenhao0810: That makes sense, will do.

istioctl/cmd/describe.go
liwenhao0810: The composed PA might not be from only one PA, to print all the active PA would involve bigger change, probably not reuse ComposePeerAuthentication(). And that would have potential sync problem with istiod's logic. 
Thats why end up printing out all applied PAs and show the active mode. The method name might be misleading though, will change that.

istioctl/cmd/describe.go
liwenhao0810: How about the edge case of `len(spec.GetSelector().MatchLabels) == 0`, is that a valid case?

istioctl/cmd/describe.go
yangminzhu: yeah, that makes sense, we can add this later.

istioctl/cmd/describe.go
yangminzhu: I think it also handles this, you can double check the function comment and test it a little bit.

istioctl/cmd/describe.go
yangminzhu: then you don't need to continue the analysis since you cannot get the accurate result anyway.

istioctl/cmd/describe.go
liwenhao0810: Added more comments, basically the idea is when there is no workload selector, it will treat as qualified, meaning it is either workload namespace or root namespace.

istioctl/cmd/describe.go
liwenhao0810: I've tested in cluster and that works, but there is no 0 size labels map.
Don't think in normal case, there is a 0 size map, it normally nil or size > 0. 
However, not sure whether there is such a case.

istioctl/cmd/describe.go
liwenhao0810: So what I did is, if user doesn't have access to root namespace, only show the matched PA in workload namespace, but will not show active mtls mode. Thoughts are that might be helpful rather than show nothing in that case.

yangminzhu: I think that is unnecessary and make the code complicated without much benefits.

istioctl/cmd/describe.go
yangminzhu: I'm not sure what you mean, are you saying the `k8s_labels.SelectorFromSet()` can not handle 0 size map?

The function comment clearly says that `A nil and empty Sets are considered equivalent to Everything().` so it should handle 0 size map without problems.

```
// SelectorFromSet returns a Selector which will match exactly the given Set. A
// nil and empty Sets are considered equivalent to Everything().
// It does not perform any validation, which means the server will reject
// the request if the Set contains invalid values.
func SelectorFromSet(ls Set) Selector {
```

liwenhao0810: Ah that's right, cool will do that.

istioctl/cmd/describe.go
liwenhao0810: Sg. Still think it might be useful, but not sure how much though. Will do the error for now, we can revisit if there is need.

istioctl/cmd/describe.go
```diff
@@ -143,6 +147,13 @@ the configuration objects that affect that pod.`,
 				return err
 			}
 
+			// render PeerAuthentication info
+			fmt.Fprintf(writer, "--------------------\n")
+			err = describePeerAuthentication(writer, configClient, istioNamespace, ns, k8s_labels.Set(pod.ObjectMeta.Labels))
```
yangminzhu, line 152: is the `istioNamespace` the root namespace? Note it could be different and configurable in meshconfig.

istioctl/cmd/describe.go
```diff
@@ -1186,3 +1197,87 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
 	}
 	return false, fmt.Errorf("no container %q in pod", containerName)
 }
+
+// describePeerAuthentication fetches all PeerAuthentication in workload and root namespace.
+// It lists the ones applied to the pod, and the current active mTLS mode.
+// When the client doesn't have access to root namespace, it will only show workload namespace Peerauthentications.
+func describePeerAuthentication(writer io.Writer, configClient istioclient.Interface, istioNamespace, workloadNamespace string, podsLabels k8s_labels.Set) error { // nolint: lll
```
yangminzhu, line 1204: Could you also update the `tests/integration/pilot/istioctl_test.go::TestDescribe()` to cover the PeerAuthN?

yangminzhu (commented):
Looks good! Just a few small comments, thanks!

istioctl/cmd/describe.go
liwenhao0810: Updated to use root namespace.

istioctl/cmd/describe.go
liwenhao0810: Is this test running in any job? I'm actually not sure the describe is showing DR correctly since in my env it didn't.
And will it make sense to update integration test in a separate pr?

istioctl/cmd/describe.go
```diff
@@ -1186,3 +1199,111 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
...
+		cfgNames += cfg.Meta.Name + "." + cfg.Meta.Namespace
+		if i < len(configs)-1 {
+			cfgNames += ", "
+		}
+	}
+	fmt.Fprintf(writer, "   %s\n", cfgNames)
+}
+
+func printPeerAuthentication(writer io.Writer, pa *v1beta1.PeerAuthentication) {
+	fmt.Fprintf(writer, "Active PeerAuthentication:\n")
```
howardjohn, line 1282: Format looks like:
```
Applied PeerAuthentication:
   default.istio-system, default.default
Active PeerAuthentication:
   Workload mTLS: PERMISSIVE
```

I think it may look better like
```
Effective PeerAuthentication: PERMISSIVE
   Configs applied: default.istio-system, default.default
```
Or similar. its not clear what 'applied' and 'active' mean to me

istioctl/cmd/describe.go
```diff
@@ -1186,3 +1199,111 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
...
+	if len(pa.PortLevelMtls) != 0 {
+		fmt.Fprintf(writer, "   Port Level mTLS:\n")
+		for port, mode := range pa.PortLevelMtls {
+			fmt.Fprintf(writer, "      %d: %s\n", port, mode.Mode.String())
+		}
+	}
+}
+
+func getMeshConfig(client kubernetes.Interface) (*meshconfig.MeshConfig, error) {
+	meshConfigMap, err := client.CoreV1().ConfigMaps(istioNamespace).Get(context.TODO(), defaultMeshConfigMapName, metav1.GetOptions{})
```
howardjohn, line 1293: this won't work if revisions are used. At least it should not fail if there is an error

howardjohn (approved):
lgtm aside from some comments. I am not an istioctl owner so will give approval on the other parts

istioctl/cmd/describe.go
liwenhao0810: Updated to support revision.

istioctl/cmd/describe.go
```diff
@@ -1186,3 +1199,111 @@ func containerReady(pod *v1.Pod, containerName string) (bool, error) {
...
+			cfgs = append(cfgs, cfg)
+		}
+	}
+
+	return cfgs
+}
+
+func printConfigs(writer io.Writer, configs []*config.Config) {
+	if len(configs) == 0 {
+		fmt.Fprintf(writer, "No config found.\n")
```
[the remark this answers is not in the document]
liwenhao0810: Agreed, removed this log, and added a check in caller that only enter this method when the `configs` is not empty.

istioctl/cmd/describe.go
liwenhao0810: Updated the output accordingly.

istioctl/cmd/describe.go
yangminzhu: yes it's running in the presubmit, search for `TestDescribe` in https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/36062/integ-pilot-k8s-tests_istio/1461881654225145856/build-log.txt

It's fine if you want to add it in a separate PR.

yangminzhu (approved):

therealmitchconnors (approved):

PR #36078, opened 2021-11-15
Fire off eds update on pod label change

### LINKED ISSUE #36056: EDS missing endpoints after resource changed
### Bug Description

 Hello everyone!
 I'm opening this issue as I noticed what seems to be some misbehaviour in the EDS.
 In particular, I noticed that if you change the label to a pod after it has been created it will not be taken into consideration when building the endpoints.
 This will result in 503 errors as I have a "subset" in the DestinationRule that is supposed to match that specific label.
 
 I managed to reproduce the issue with Istio 1.11.4 and Istio 1.12-beta2

 The step to reproduce it are quite easy:

 0. Download the yaml manifests which you can find here: [yaml_resources.zip](https://github.com/istio/istio/files/7529010/yaml_resources.zip)
 1. Install Istio with controller + Ingress Gateway
 2. Deploy httpbin.yaml that comes with podapp=httpbin label by default
 3. Wait for httpbin to be ready.
 4. Apply the broken=httpbin label to the httpbin podapp via `kubectl label pod HTTPBIN_POD_NAME broken=httpbin`
 5. Deploy vs_dr.yaml (note that the file includes a gw as well, sorry for the misleading name) and check the endpoints in the gateway (**only endpoints for podapp are available**)
 ```ENDPOINT                         STATUS      OUTLIER CHECK     CLUSTER
10.63.31.6:8443                  HEALTHY     OK                outbound|443||kubernetes.default.svc.cluster.local
127.0.0.1:15000                  HEALTHY     OK                prometheus_stats
127.0.0.1:15020                  HEALTHY     OK                agent
172.17.0.12:44134                HEALTHY     OK                outbound|44134||tiller-deploy.kube-system.svc.cluster.local
172.17.0.2:53                    HEALTHY     OK                outbound|53||kube-dns.kube-system.svc.cluster.local
172.17.0.2:9153                  HEALTHY     OK                outbound|9153||kube-dns.kube-system.svc.cluster.local
172.17.0.5:4443                  HEALTHY     OK                outbound|443||metrics-server.kube-system.svc.cluster.local
172.17.0.6:8080                  HEALTHY     OK                outbound|80||istio-ingressgateway.istio-system.svc.cluster.local
172.17.0.6:8443                  HEALTHY     OK                outbound|443||istio-ingressgateway.istio-system.svc.cluster.local
172.17.0.6:15021                 HEALTHY     OK                outbound|15021||istio-ingressgateway.istio-system.svc.cluster.local
172.17.0.7:80                    HEALTHY     OK                outbound|8000|podapp|httpbin.default.svc.cluster.local
172.17.0.7:80                    HEALTHY     OK                outbound|8000||httpbin.default.svc.cluster.local
172.17.0.8:15010                 HEALTHY     OK                outbound|15010||istiod.istio-system.svc.cluster.local
172.17.0.8:15012                 HEALTHY     OK                outbound|15012||istiod.istio-system.svc.cluster.local
172.17.0.8:15014                 HEALTHY     OK                outbound|15014||istiod.istio-system.svc.cluster.local
172.17.0.8:15017                 HEALTHY     OK                outbound|443||istiod.istio-system.svc.cluster.local
unix://./etc/istio/proxy/SDS     HEALTHY     OK                sds-grpc
unix://./etc/istio/proxy/XDS     HEALTHY     OK                xds-grpc
```

 6. You can confirm that things are not working by making a `curl GATEWAY_URL/status/200 -i` and `curl GATEWAY_URL/status/201 -i`. Only the first one will work as that is handled by the "podapp" subset. The status/201 request that is handled by the "broken" subset label is returning 503 instead of 201.
 6. Restart the controller and check the endpoints again. You will find soon enough that both the one for "podapp" and "broken" are present now. Note that restarting the ingress-gateway will not yield to any improvement (only the restarting of the controller will do).
 `curl GATEWAY_URL/status/201 -i` will now return the expected 201 status code and not 503.
  ```
ENDPOINT                         STATUS      OUTLIER CHECK     CLUSTER
10.63.31.6:8443                  HEALTHY     OK                outbound|443||kubernetes.default.svc.cluster.local
127.0.0.1:15000                  HEALTHY     OK                prometheus_stats
127.0.0.1:15020                  HEALTHY     OK                agent
172.17.0.12:44134                HEALTHY     OK                outbound|44134||tiller-deploy.kube-system.svc.cluster.local
172.17.0.2:53                    HEALTHY     OK                outbound|53||kube-dns.kube-system.svc.cluster.local
172.17.0.2:9153                  HEALTHY     OK                outbound|9153||kube-dns.kube-system.svc.cluster.local
172.17.0.3:15010                 HEALTHY     OK                outbound|15010||istiod.istio-system.svc.cluster.local
172.17.0.3:15012                 HEALTHY     OK                outbound|15012||istiod.istio-system.svc.cluster.local
172.17.0.3:15014                 HEALTHY     OK                outbound|15014||istiod.istio-system.svc.cluster.local
172.17.0.3:15017                 HEALTHY     OK                outbound|443||istiod.istio-system.svc.cluster.local
172.17.0.5:4443                  HEALTHY     OK                outbound|443||metrics-server.kube-system.svc.cluster.local
172.17.0.6:8080                  HEALTHY     OK                outbound|80||istio-ingressgateway.istio-system.svc.cluster.local
172.17.0.6:8443                  HEALTHY     OK                outbound|443||istio-ingressgateway.istio-system.svc.cluster.local
172.17.0.6:15021                 HEALTHY     OK                outbound|15021||istio-ingressgateway.istio-system.svc.cluster.local
172.17.0.7:80                    HEALTHY     OK                outbound|8000|broken|httpbin.default.svc.cluster.local
172.17.0.7:80                    HEALTHY     OK                outbound|8000|podapp|httpbin.default.svc.cluster.local
172.17.0.7:80                    HEALTHY     OK                outbound|8000||httpbin.default.svc.cluster.local
unix://./etc/istio/proxy/SDS     HEALTHY     OK                sds-grpc
unix://./etc/istio/proxy/XDS     HEALTHY     OK                xds-grpc
```

 **Note that the issue does not appear if the label is added before the Istio controller is even started.**

### Version

```prose
$ istioctl version
client version: 1.12.0-beta.2
control plane version: 1.12.0-beta.2
data plane version: 1.12.0-beta.2 (1 proxies)
$ kubectl version --short
Client Version: v1.18.2
Server Version: v1.20.2
$ helm version --short
v3.4.0+g7090a89
```

### Additional Information

 [bug-report.zip](https://github.com/istio/istio/files/7529029/bug-report.zip)

dddddai:

Now the control plane does not fire off an eds update when pod label changes, which may lead to inconsistency, please refer to https://github.com/istio/istio/issues/36056#issue-1052149403

Fix #36056

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -399,7 +399,7 @@ func NewController(kubeClient kubelib.Client, options Options) *Controller {
 			})
 		}
 	})
-	c.registerHandlers(c.pods.informer, "Pods", c.pods.onEvent, nil)
+	c.registerHandlers(c.pods.informer, "Pods", c.pods.onEvent, c.pods.checkPodLabelChange)
 
 	c.exports = newServiceExportCache(c)
 	c.imports = newServiceImportCache(c)
--- a/pilot/pkg/serviceregistry/kube/controller/pod.go
+++ b/pilot/pkg/serviceregistry/kube/controller/pod.go
@@ -25,6 +25,7 @@ import (
 	"istio.io/istio/pilot/pkg/serviceregistry/kube"
 	"istio.io/istio/pilot/pkg/serviceregistry/kube/controller/filter"
 	"istio.io/istio/pilot/pkg/util/sets"
+	"istio.io/istio/pkg/config/labels"
 )
 
 // PodCache is an eventually consistent pod cache
@@ -43,7 +44,8 @@ type PodCache struct {
 	// needResync is map of IP to endpoint namespace/name. This is used to requeue endpoint
 	// events when pod event comes. This typically happens when pod is not available
 	// in podCache when endpoint event comes.
-	needResync         map[string]sets.Set
+	needResync map[string]sets.Set
+	// queueEndpointEvent is a callback that triggers endpoint events to update eds
 	queueEndpointEvent func(string)
 
 	c *Controller
@@ -177,6 +179,27 @@ func (pc *PodCache) notifyWorkloadHandlers(pod *v1.Pod, ev model.Event) {
 		PortMap:   getPortMap(pod),
 	}
 	pc.c.handlers.NotifyWorkloadHandlers(workloadInstance, ev)
+}
+
+// checkPodLabelChange fires off eds update on pod label change
+func (pc *PodCache) checkPodLabelChange(old, cur interface{}) bool {
+	oldPod, ok := old.(*v1.Pod)
+	if !ok {
+		return true
+	}
+	curPod, ok := cur.(*v1.Pod)
+	if !ok {
+		return true
+	}
+	if !labels.Instance(oldPod.Labels).Equals(curPod.Labels) {
+		// fire off eds update on pod label change
+		if services, err := getPodServices(pc.c.serviceLister, curPod); err == nil {
+			for _, svc := range services {
+				pc.queueEndpointEvent(kube.KeyFunc(svc.Name, svc.Namespace))
+			}
+		}
+	}
+	return false
 }
 
 func getPortMap(pod *v1.Pod) map[string]uint32 {
--- a/pilot/pkg/serviceregistry/kube/controller/controller_test.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller_test.go
@@ -2139,6 +2139,88 @@ func TestEndpointUpdateBeforePodUpdate(t *testing.T) {
 	}
 }
 
+func TestEdsUpdateOnPodLabelChange(t *testing.T) {
+	controller, fx := NewFakeControllerWithOptions(FakeControllerOptions{})
+	go controller.Run(controller.stop)
+	// Wait for the caches to sync, otherwise we may hit race conditions where events are dropped
+	cache.WaitForCacheSync(controller.stop, controller.HasSynced)
+	defer controller.Stop()
+
+	// Create initial pods with a service, and endpoints.
+	pod1 := generatePod("172.0.1.1", "pod1", "nsA", "", "node1", map[string]string{"app": "prod-app"}, map[string]string{})
+	pod2 := generatePod("172.0.1.2", "pod2", "nsA", "", "node1", map[string]string{"app": "prod-app"}, map[string]string{})
+	pods := []*coreV1.Pod{pod1, pod2}
+	nodes := []*coreV1.Node{
+		generateNode("node1", map[string]string{NodeZoneLabel: "zone1", NodeRegionLabel: "region1", label.TopologySubzone.Name: "subzone1"}),
+	}
+	addNodes(t, controller, nodes...)
+	addPods(t, controller, fx, pods...)
+
+	createService(controller, "svc1", "nsA", nil,
+		[]int32{8080}, map[string]string{"app": "prod-app"}, t)
+	if ev := fx.Wait("service"); ev == nil {
+		t.Fatal("Timeout creating service")
+	}
+
+	pod1Ips := []string{"172.0.1.1"}
+	portNames := []string{"tcp-port"}
+	createEndpoints(t, controller, "svc1", "nsA", portNames, pod1Ips, nil, nil)
+	if ev := fx.Wait("eds"); ev == nil {
+		t.Fatal("Timeout incremental eds")
+	}
+
+	// update eds on pod label change
+	pod1.SetLabels(map[string]string{
+		"app": "prod-app",
+		"foo": "bar",
+	})
+	_, err := controller.client.CoreV1().Pods(pod1.Namespace).Update(context.TODO(), pod1, metaV1.UpdateOptions{})
+	if err != nil {
+		t.Fatalf("Cannot update %s in namespace %s (error: %v)", pod1.ObjectMeta.Name, pod1.ObjectMeta.Namespace, err)
+	}
+	if ev := fx.Wait("eds"); ev == nil {
+		t.Fatal("Timeout incremental eds")
+	}
+
+	// update eds on pod label change
+	pod1.SetLabels(map[string]string{
+		"app": "prod-app",
+	})
+	_, err = controller.client.CoreV1().Pods(pod1.Namespace).Update(context.TODO(), pod1, metaV1.UpdateOptions{})
+	if err != nil {
+		t.Fatalf("Cannot update %s in namespace %s (error: %v)", pod1.ObjectMeta.Name, pod1.ObjectMeta.Namespace, err)
+	}
+	if ev := fx.Wait("eds"); ev == nil {
+		t.Fatal("Timeout incremental eds")
+	}
+
+	// don't update eds if there is no matched service
+	pod1.SetLabels(map[string]string{
+		"app": "prod-app2",
+	})
+	_, err = controller.client.CoreV1().Pods(pod1.Namespace).Update(context.TODO(), pod1, metaV1.UpdateOptions{})
+	if err != nil {
+		t.Fatalf("Cannot update %s in namespace %s (error: %v)", pod1.ObjectMeta.Name, pod1.ObjectMeta.Namespace, err)
+	}
+	if ev := fx.Wait("eds"); ev != nil {
+		t.Fatal("Update eds unexpectedly")
+	}
+
+	// don't update eds for other fields change
+	pod2.SetAnnotations(map[string]string{
+		"foo": "bar",
+	})
+	pod2.Spec.Containers[0].ImagePullPolicy = coreV1.PullAlways
+	pod2.Spec.RestartPolicy = coreV1.RestartPolicyOnFailure
+	_, err = controller.client.CoreV1().Pods(pod2.Namespace).Update(context.TODO(), pod2, metaV1.UpdateOptions{})
+	if err != nil {
+		t.Fatalf("Cannot update %s in namespace %s (error: %v)", pod2.ObjectMeta.Name, pod2.ObjectMeta.Namespace, err)
+	}
+	if ev := fx.Wait("eds"); ev != nil {
+		t.Fatal("Update eds unexpectedly")
+	}
+}
+
 func TestWorkloadInstanceHandlerMultipleEndpoints(t *testing.T) {
 	controller, fx := NewFakeControllerWithOptions(FakeControllerOptions{})
 	go controller.Run(controller.stop)
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
@@ -408,13 +408,11 @@ func (s *ServiceEntryStore) WorkloadInstanceHandler(wi *model.WorkloadInstance, 
 	redundantEventForPod := false
 
 	var addressToDelete string
+	var oldWorkload *model.WorkloadInstance
+	k := types.NamespacedName{Namespace: wi.Namespace, Name: wi.Name}
 
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
-
-	// this is from a pod. Store it in separate map so that
-	// the refreshIndexes function can use these as well as the store ones.
-	k := types.NamespacedName{Namespace: wi.Namespace, Name: wi.Name}
 	switch event {
 	case model.EventDelete:
 		if s.workloadInstances.get(k) == nil {
@@ -425,14 +423,16 @@ func (s *ServiceEntryStore) WorkloadInstanceHandler(wi *model.WorkloadInstance, 
 		}
 	default: // add or update
 		if old := s.workloadInstances.get(k); old != nil {
-			if old.Endpoint.Address != wi.Endpoint.Address {
-				addressToDelete = old.Endpoint.Address
-			}
 			// If multiple k8s services select the same pod or a service has multiple ports,
 			// we may be getting multiple events ignore them as we only care about the Endpoint IP itself.
 			if model.WorkloadInstancesEqual(old, wi) {
 				// ignore the update as nothing has changed
 				redundantEventForPod = true
+			} else {
+				oldWorkload = old
+				if old.Endpoint.Address != wi.Endpoint.Address {
+					addressToDelete = old.Endpoint.Address
+				}
 			}
 		}
 		s.workloadInstances.update(wi)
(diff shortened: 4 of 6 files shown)


pilot/pkg/serviceregistry/kube/controller/pod.go
```diff
@@ -157,6 +165,18 @@ func (pc *PodCache) onEvent(curr interface{}, ev model.Event) error {
 			} else if shouldPodBeInEndpoints(pod) {
 				if key != pc.podsByIP[ip] {
 					pc.update(ip, key)
+				} else if !model.WorkloadInstancesEqual(pc.c.workloadInstancesByPods[key], instance) {
+					// fire off eds update on pod change
+					svcList, err := pc.c.serviceLister.Services(pod.Namespace).List(k8sLabels.Everything())
```
howardjohn, line 170: Currently the pod labels for a connected pod are static as they are part of the XDS node metadata which does not update.

When changing this, are we only solving one end of the problem that may lead to unexpected state?

pilot/pkg/serviceregistry/kube/controller/pod.go
```diff
@@ -157,6 +165,18 @@ func (pc *PodCache) onEvent(curr interface{}, ev model.Event) error {
...
+					if err != nil {
+						return err
+					}
+					for _, svc := range svcList {
+						// look for matched services and update eds
+						if labels.Instance(svc.Spec.Selector).SubsetOf(pod.Labels) {
```
howardjohn, line 176: there may be performance concerns about O(n) lookup on services for every pod change

dddddai: Yes, but I didn't figure out a better solution, the same pattern can be found in workload instance handlers

https://github.com/istio/istio/blob/bc83862fbc62ca83493545ab332bdd68fe4b7e6b/pilot/pkg/serviceregistry/kube/controller/controller.go#L1168-L1177
https://github.com/istio/istio/blob/bc83862fbc62ca83493545ab332bdd68fe4b7e6b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go#L441-L445

pilot/pkg/serviceregistry/kube/controller/pod.go
dddddai: Do you mean we should update XDS node metadata when pod label changes?

hzxuzhonghu: We do not consider changing pod labels as a good practice, and so did not make it at first. I doubt it is not worth to support it with so much performance sacrifice.

dddddai: I see, but it sounds a bit odd that the rule would be different after istiod restarted

Generally pod update events(**which we care**) do not happen frequently, does it really kill performance a lot?
And actually `ServiceEntryStore.WorkloadInstanceHandler` does similar things as this patch:
https://github.com/istio/istio/blob/master/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go#L441-L468

Anyway I'm ok if we really don't need this feature

pilot/pkg/model/service_test.go
```diff
@@ -192,11 +192,23 @@ func TestWorkloadInstanceEqual(t *testing.T) {
...
 	}{
+		{
+			comparer: nil,
+			comparee: nil,
+			shouldEq: true,
+			name:     "two null endpoints",
+		},
+		{
+			comparer: nil,
+			comparee: exampleInstance.DeepCopy(),
```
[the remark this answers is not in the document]
dddddai: I add the nil guard because `pc.c.workloadInstancesByPods[key]` would be nil when the pod becomes ready for the first time, and this is the util test to verify it works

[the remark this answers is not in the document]
dddddai: Oh I think it's not needed cos the ready event would be handled by endpoints controller, thanks for reminding, I will fix it later

pilot/pkg/serviceregistry/kube/controller/pod.go
howardjohn: > Do you mean we should update XDS node metadata when pod label changes?

AFAIK this requires restarting Envoy which is not really viable

dddddai: > AFAIK this requires restarting Envoy which is not really viable

Yeah, it's bootstrap configuration

Anyway I think we should "keep the same behavior as workload handler" which does [fire eds update on workload change](https://github.com/istio/istio/blob/master/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go#L468), WDYT?

dddddai: Or maybe give an env variable(default to false) to control this?

dddddai:
Hello @howardjohn @hzxuzhonghu, any ideas?

hzxuzhonghu:
As a short term fix, i think it is viable. But long term fix we need to support meta label update. 
Need another look though

howardjohn:
Generally in favor of the change directionally, just not sure it is sufficient to cover all cases and the test coverage is not high enough to prove this as is. Maybe this is step 1 though and we can followup.

Just high level review, I haven't had a chance to review the code yet but plan to soon.

dddddai:
> Generally in favor of the change directionally, just not sure it is sufficient to cover all cases and the test coverage is not high enough to prove this as is.

Thanks! I added some negative tests to cover more cases

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -412,7 +412,7 @@ func NewController(kubeClient kubelib.Client, options Options) *Controller {
 			})
 		}
 	})
-	c.registerHandlers(c.pods.informer, "Pods", c.pods.onEvent, nil)
+	c.registerHandlers(c.pods.informer, "Pods", c.pods.onEvent, c.pods.checkPodLabelChange)
```
hzxuzhonghu, line 415: why check the labels in filter func

pilot/pkg/serviceregistry/kube/controller/pod.go
```diff
@@ -170,19 +171,42 @@ func (pc *PodCache) onEvent(curr interface{}, ev model.Event) error {
...
 		}
 		// fire instance handles for workload
-		ep := NewEndpointBuilder(pc.c, pod).buildIstioEndpoint(ip, 0, "", model.AlwaysDiscoverable)
-		workloadInstance := &model.WorkloadInstance{
-			Name:      pod.Name,
-			Namespace: pod.Namespace,
-			Kind:      model.PodKind,
-			Endpoint:  ep,
-			PortMap:   getPortMap(pod),
+		if len(pc.c.handlers.GetWorkloadHandlers()) > 0 {
```
hzxuzhonghu, line 174: it looks redundant, isn't this  always > 0

pilot/pkg/serviceregistry/kube/controller/pod.go
```diff
@@ -170,19 +171,42 @@ func (pc *PodCache) onEvent(curr interface{}, ev model.Event) error {
...
+func (pc *PodCache) checkPodLabelChange(old, cur interface{}) bool {
+	oldPod, ok := old.(*v1.Pod)
+	if !ok {
+		return true
+	}
+	curPod, ok := cur.(*v1.Pod)
+	if !ok {
+		return true
+	}
+	if !labels.Instance(oldPod.Labels).Equals(curPod.Labels) {
```
hzxuzhonghu, line 199: is this better than reflect.DeepEqual

pilot/pkg/serviceregistry/kube/controller/pod.go
```diff
@@ -170,19 +171,42 @@ func (pc *PodCache) onEvent(curr interface{}, ev model.Event) error {
...
+		// fire off eds update on pod label change
+		if services, err := getPodServices(pc.c.serviceLister, curPod); err == nil {
+			for _, svc := range services {
+				pc.queueEndpointEvent(kube.KeyFunc(svc.Name, svc.Namespace))
```
hzxuzhonghu, line 203: Not to use `queueEndpointEvent`, it is aimed to do resync when pod comes later than endpoint

pilot/pkg/serviceregistry/kube/controller/pod.go
```diff
@@ -170,19 +171,42 @@ func (pc *PodCache) onEvent(curr interface{}, ev model.Event) error {
...
+	if !ok {
+		return true
+	}
+	curPod, ok := cur.(*v1.Pod)
+	if !ok {
+		return true
+	}
+	if !labels.Instance(oldPod.Labels).Equals(curPod.Labels) {
+		// fire off eds update on pod label change
+		if services, err := getPodServices(pc.c.serviceLister, curPod); err == nil {
```
hzxuzhonghu, line 201: i think need to process old pod service too

hzxuzhonghu: Oh, maybe not , if it is endpoints removal, it will be triggered by ep informer

pilot/pkg/serviceregistry/kube/controller/pod.go
dddddai: I'm not sure, shouldn't it be 0 if `ServiceEntrySelectPod` is disabled?

pilot/pkg/serviceregistry/kube/controller/controller.go
dddddai: Because the handler func does not include the old object, and filter func is the only place to fetch it.

I know it looks kind of hacky to do this in the filter func, but I believe it's way cheaper than maintaining a map just for fetching old pods

pilot/pkg/serviceregistry/kube/controller/pod.go
dddddai: > Oh, maybe not , if it is endpoints removal, it will be triggered by ep informer

Yes

pilot/pkg/serviceregistry/kube/controller/pod.go
dddddai: I'm not sure but it should be cheaper than reflect, cos reflect has to figure out the type and casting the objects

pilot/pkg/serviceregistry/kube/controller/pod.go
dddddai: Does it really matter? Please correct me if I'm wrong

pilot/pkg/serviceregistry/kube/controller/controller.go
hzxuzhonghu: IC

pilot/pkg/serviceregistry/kube/controller/pod.go
hzxuzhonghu: it is still >0, just the handler does nothing

pilot/pkg/serviceregistry/kube/controller/pod.go
hzxuzhonghu: It is confusing, or you have to update the comment In L43

hzxuzhonghu (commented):
Generally, LGTM, one more question, looks not consider pods only selected by SE.

pilot/pkg/serviceregistry/kube/controller/pod.go
dddddai: https://github.com/istio/istio/blob/1202dc2a681b4b6a9e98f5414249f346129a2188/pilot/pkg/serviceregistry/kube/controller/multicluster.go#L177-L180

Why? Is there any other place where appending workload handlers?

pilot/pkg/serviceregistry/kube/controller/pod.go
dddddai: OK, thanks for reminding

dddddai:
> 

Thanks for your review! pods selected by SE are handled by `ServiceEntryStore.WorkloadInstanceHandler`

https://github.com/istio/istio/blob/bc83862fbc62ca83493545ab332bdd68fe4b7e6b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go#L468

pilot/pkg/serviceregistry/kube/controller/pod.go
dddddai: I've added a comment for `queueEndpointEvent`

pilot/pkg/serviceregistry/kube/controller/pod.go
hzxuzhonghu: right

hzxuzhonghu (approved):

dddddai:
@howardjohn sorry to bother you, could you please take a look?

howardjohn:
Sorry have been on holiday. Would be great to have a test that applies a config with workloadSelector, verifies it has no impact, then labels pod, verifies it has impact, then removes label and verifies it no longer has impact.

Will take a closer look soon

dddddai:
> Would be great to have a test that applies a config with workloadSelector, verifies it has no impact, then labels pod, verifies it has impact, then removes label and verifies it no longer has impact.

Thanks for the suggestion, I've added a test for se

kfaseela:
@dddddai  thanks for the PR. We have tested this locally as well to check if the problem mentioned in #36056 is solved according to our usecase. And seems to work fine. @howardjohn @hzxuzhonghu

hzxuzhonghu (approved):

howardjohn:
I am a bit worried that this PR only partially address the problem. This gives the mistaken idea that dynamic pod labels are supported. For example, this policy will not support dynamic updates even with this PR:

```yaml
kind: AuthorizationPolicy
metadata:
  name: name
spec:
  selector:
    matchLabels:
      custom: label
  action: DENY
  rules:
  - to:
    - operation:
        methods: ["GET"]
        paths: ["/user"]
```

I have no problem adding support for dynamic pod labels, but only if we do it completely. We also need integration tests that do things like:
* Apply Config X with label selector Y not matching pod
* Send request, assert config is not applied
* Label pod with Y
* Send request, assert config is applied
* Unlabel pod
* Send request, assert config is not applied

We would need this test for a variety of types as well - Dest rule, authz policy, etc.

PR #36081, opened 2021-11-15
add downstream_cx_active stats only when ExitOnZeroActiveConnections is used

ramaraochavali:
add downstream_cx_active stats only when  ExitOnZeroActiveConnections is enabled.

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -625,6 +625,9 @@ type NodeMetadata struct {
 	// Envoy prometheus port redirecting to admin port prometheus endpoint.
 	EnvoyPrometheusPort int `json:"ENVOY_PROMETHEUS_PORT,omitempty"`
 
+	// ExitOnZeroActiveConnections terminates Envoy if there are no active connections if set.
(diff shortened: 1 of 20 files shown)


ramaraochavali:
Put a hold because https://github.com/istio/istio/blob/master/pkg/envoy/agent.go#L187 needs to be handled for prometheus as well. Might need regex change

ramaraochavali:
@howardjohn This as a fix to avoid `downstream_cx_active` metrics when the feature is not used is fine. 

I thought we can avoid http.downstream_cx_active also since we can just rely on listener stats only as was done in the code but it is not possible to do that because we can not add/change the regex. So this is the best we can do. WDYT?

howardjohn (approved):

mandarjog:
I think this is ok as a strict feature flag based solution.
However customers do want to know number of downstream active connections specifically at ingress.
Previously this was possible by adding a new inclusion, not it will be magically filtered out unless you also enable this feature.

mandarjog:
Cc @douglas-reid

Issue #35902, opened 2021-11-04
istio-cni plugin fails when installing Istio with non-root user


johnma14:
### Bug Description

For users who build their own Istio images  with non-root user, Istio fails to install when CNI plugin is enabled. 

In 1.11, as part of the beta graduation of [Istio CNI plugin](https://github.com/istio/istio/issues/24416], a new feature was added to [enable CNI plugin to send logs to the CNI container](https://github.com/istio/istio/pull/33926). It creates a hostPath volume to write the logs at `/var/run/istio-cni` which requires root privileges. Since this is a hostPath volume, it is created when the pod comes up and before the app containers start running these non-root  images. This directory is created by the root user and owned by root:root with default permission of 755 [`rwxr-xr-x 1 root root  log.sock`].

From the Kubernetes docs about [hostPaths](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath):

> The files or directories created on the underlying hosts are only writable by root. You either need to run your
process` as root in a privileged Container or modify the file permissions on the host to be able to write to a 
hostPath volume.

When the install-cni container starts with the non-root user, it tries to start the UDS log server at `/var/run/istio-cni` directory which is owned by the root user and fails because it does not have sufficient permissions to write to log.sock

Two approaches I tried that worked - In [daemonset.yaml](https://github.com/istio/istio/blob/master/manifests/charts/istio-cni/templates/daemonset.yaml) :

1. **Explicitly set the securityContext in the install-cni container to run as the root user.**
    ~~~
        securityContext:
            runAsUser: 0
    ~~~
2. **Create initContainers to chown (from root:root -> root:1337) and chmod (user in the group 1337 can rwx/775) the hostPath mount point.**
    ~~~
     initContainers:
        - name: istio-volume-permissions-chown
          image: busybox
          command: ["/bin/chown","-R", "root:1337" , "/var/run/istio-cni"]
          volumeMounts:
            - name: cni-log-dir
              mountPath: /var/run/istio-cni
        - name: istio-volume-permissions-chmod
          image: busybox
          command: ["/bin/chmod","-R","775", "/var/run/istio-cni"]
          volumeMounts:
            - name: cni-log-dir
              mountPath: /var/run/istio-cni
      ~~~

### Version

```prose
- istioctl version: 

client version: 1.12-dev
control plane version: 1.12-dev-8a03c8529b90e382984f65fe43cc14ec2e831948
data plane version: 1.12-dev (1 proxies)
```

### Additional Information

[...]

PR #35986, opened 2021-11-09
Add securityContext to CNI DaemonSet to run as root.

bianpengyuan:
fix https://github.com/istio/istio/issues/35902

PR #36097, opened 2021-11-15
[release-1.11] revert /host prefix from CNI log volume mount path. (#36038)

johnma14:
(cherry picked from commit 8dcf53141d6d50efef56da1506b3d1076a93450a)

 Fixes: #35902

### DIFF
--- a/cni/pkg/cmd/root.go
+++ b/cni/pkg/cmd/root.go
@@ -134,7 +134,7 @@ func init() {
 	registerStringArrayParameter(constants.SkipCNIBinaries, []string{},
 		"Binaries that should not be installed. Currently Istio only installs one binary `istio-cni`")
 	registerIntegerParameter(constants.MonitoringPort, 15014, "HTTP port to serve prometheus metrics")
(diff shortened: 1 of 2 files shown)


johnma14:
@bianpengyuan not sure if I did this cherry pick correctly. Can you help verify. Thanks.

bianpengyuan (commented):
Looks like no file changes. I think you will need to cherry-pick the securityContext change first, then add this commit on top of that.

johnma14:
Done @bianpengyuan. That actually worked. Thank you for the pointer.

bianpengyuan (approved):

johnma14:
Adding a little context to help with reviewing. 

We first made this change #35986 to fix the issue with the CNI plugin not starting for non-root users. In that PR, we also made an additional change to change the mountPath for the logUDSPath. We had to revert that specific change in another PR [#36038], since it was breaking the logging feature. So I manually cherry picked the commit id for both these PR's and you can see that it removed the changes related to the mountPath. Hope that helps. Thanks for the review.

jwendell (approved):

PR #36098, opened 2021-11-15
[release-1.12] revert /host prefix from CNI log volume mount path. (#36038)

johnma14:
(cherry picked from commit 8dcf53141d6d50efef56da1506b3d1076a93450a)

Fixes: #35902

### DIFF
--- a/cni/pkg/cmd/root.go
+++ b/cni/pkg/cmd/root.go
@@ -136,7 +136,7 @@ func init() {
 	registerStringArrayParameter(constants.SkipCNIBinaries, []string{},
 		"Binaries that should not be installed. Currently Istio only installs one binary `istio-cni`")
 	registerIntegerParameter(constants.MonitoringPort, 15014, "HTTP port to serve prometheus metrics")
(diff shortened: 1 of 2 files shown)


bianpengyuan (approved):

johnma14:
Thanks @bianpengyuan

dgn:
strangely, the changes in here and those in the [referenced PR](https://github.com/istio/istio/pull/35986) look wildly different?

johnma14:
@dgn I should have added some comments in here about what we were doing. Apologies.

So we first made this change #35986 to fix the issue with the CNI plugin not starting for non-root users. In that PR, we also made an additional change to change the mountPath for the logUDSPath. We had to revert that specific change in another PR [#36038], since it was breaking the logging feature. So I manually cherry picked the commit id for both these PR's and you can see that it removed the changes related to the mountPath. Hope that helps. Thanks for the review.

dgn:
Ah, thanks-- I was looking at the complete diff and it obviously didn't show any changes to the mount path (as you added and removed them both in this PR). Thanks for the explanation

dgn:
Adding a hold until rc1 is confirmed for release

johnma14:
Thanks @dgn. Appreciate it.

Issue #36108, opened 2021-11-17
istio-cni-node used too much memory


cocotyty:
### Bug Description

istio-cni-node used too much memory because of repair controller.

Every istio-cni-node used nearly 9GB memory in my cluster.

Maybe we can move repair controller to istiod ?

### Version

```prose
$ istioctl version
client version: 1.11.4
istiod version: 1.10.2
istiod version: 1.10.2
istiod version: 1.10.2
istiod version: 1.11.4
istiod version: 1.11.4
istiod version: 1.11.4
istiod version: 1.11.4
istiod version: 1.11.4
data plane version: 1.11.4 (5564 proxies), 1.10.2 (399 proxies)

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.10", GitCommit:"1bea6c00a7055edef03f1d4bb58b773fa8917f11", GitTreeState:"clean", BuildDate:"2020-02-11T20:13:57Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-13T11:13:49Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

cocotyty:
@howardjohn

howardjohn:
wow. Why is it so much? I thought we were only watching pods on the same node (ie up to 110 max).

cc @bianpengyuan @stewartbutler

bianpengyuan:
Looks like the node name is never used to filter..

bianpengyuan:
@cocotyty does the memory grow over time? If so it is very likely that this is caused by the queue backlog.

cocotyty:
![image](https://user-images.githubusercontent.com/2241753/142148322-c9a411e3-4f40-4a98-83e3-86df9dd687ce.png)
@bianpengyuan 
top in a new pod.

bianpengyuan:
Ok, it is relatively small comparing to 9G, but 0.7G is still bigger than I would expect with a new pod. Would be nice if you can try out https://github.com/istio/istio/pull/36109 and see if that can improve or not.

cocotyty:
[...]

PR #36109, opened 2021-11-17
CNI: filter pod event from different node and detect pod that needs to be repaired before adding it to work queue.

bianpengyuan:
fix https://github.com/istio/istio/issues/36108

### DIFF
--- a/cni/pkg/repair/repaircontroller.go
+++ b/cni/pkg/repair/repaircontroller.go
@@ -15,6 +15,7 @@
 package repair
 
 import (
+	"fmt"
 	"strings"
 	"time"
 
@@ -62,6 +63,8 @@ func NewRepairController(reconciler brokenPodReconciler) (*Controller, error) {
 					fieldSelectors = append(fieldSelectors, fs)
 				}
 			}
+			// filter out pod events from different nodes
+			fieldSelectors = append(fieldSelectors, fmt.Sprintf("spec.nodeName=%v", reconciler.cfg.NodeName))
 			options.LabelSelector = strings.Join(labelSelectors, ",")
 			options.FieldSelector = strings.Join(fieldSelectors, ",")
 		},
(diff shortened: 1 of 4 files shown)


cni/pkg/repair/repaircontroller.go
```diff
@@ -69,16 +69,27 @@ func NewRepairController(reconciler brokenPodReconciler) (*Controller, error) {
...
 	return c, nil
 }
 
+func (rc *Controller) mayAddToWorkQueue(obj interface{}) {
+	pod, ok := obj.(*v1.Pod)
+	if !ok {
+		repairLog.Error("Cannot convert object to pod. Skip adding it to the repair working queue.")
+		return
+	}
+	if rc.reconciler.detectPod(*pod) {
```
howardjohn, line 88: confused how this can happen  - shouldn't the fieldselector already filter these out?

Doing this type of clientside filtering will likely not have a huge impact, as we still need to process all pods in memory. FieldSelector is server side

cni/pkg/repair/repaircontroller.go
```diff
@@ -62,23 +63,36 @@ func NewRepairController(reconciler brokenPodReconciler) (*Controller, error) {
 					fieldSelectors = append(fieldSelectors, fs)
 				}
 			}
+			// filter out pod events from different nodes
+			fieldSelectors = append(fieldSelectors, fmt.Sprintf("spec.nodeName=%v", reconciler.cfg.NodeName))
```
howardjohn, line 67: is this going to be an issue if they have already defined this in `reconciler.cfg.FieldSelectors`? Seems like its actually ok to have multiple duplicate selector although its a bit weird

howardjohn (approved):

cni/pkg/repair/repaircontroller.go
bianpengyuan: Yeah it should be ok to have duplicated selectors.

PR #36121, opened 2021-11-17
Integration test framwork improvements and enable istiodless remote testing

frankbu:


tests/integration/telemetry/stats/prometheus/nullvm/dashboard_test.go
```diff
@@ -67,7 +67,8 @@ var dashboards = []struct {
 			// In default install, we have no proxy
 			"istio-proxy",
 			// https://github.com/istio/istio/issues/22674 this causes flaky tests
-			"galley_validation_passed",
+			//"galley_validation_passed",
+			"galley_validation_failed",
```
frankbu, line 71: Oddly, this test fails looking for `galley_validation_failed` on the primary cluster in the multicluster topology, even though all this PR is changing is making remote clusters, which are excluded (`requirePrimary == true`) from this test anyway, istiodless. The istio installation on the primary cluster, where it fails, is unchanged.

I also noticed that the test seems to work fine without excluding `galley_validation_passed` on both single and multicluster topologies. Maybe the bug is fixed?

@howardjohn @douglas-reid Any ideas? Is this change OK?

stevenctl:
Although it would be nice to simplify this code a bit, should we get rid of the `Istiodless` check? It may be good to keep both paths tested until we've got consensus that we're going this direction. When we first added istiodless tests, the with-istiod path was broken and we didn't' find it until bug bash.

tests/integration/telemetry/stats/prometheus/nullvm/dashboard_test.go
douglas-reid: clarification: it sounds like we want to remove `galley_validation_passed` despite it now seeming to work?  and we want to add `galley_validation_failed` despite it breaking the test?

Or have I misread that?

what in the test would cause a validation failure?

douglas-reid: ah, wait, i see. These are the excluded.

douglas-reid: This seems fine to me -- unless there is something that should be causing validation failures.

frankbu: RIght, before my change it seemed to be expecting a validation failure, but now there is none (at least in the multicluster env), so I had to exclude it. Seems strange that it was expecting some kind of failure anyway.

The validation_passed one was previously excluded because it was flaky according to @howardjohn, but after re-enabling it, I don't see any failures, so I wondering if the flakiness is fixed now?

howardjohn: It only failed in postsubmit on old k8s versions. No clue why. So it may pass in presubmit and fail in postsubmit

frankbu:
@stevenctl I'm not sure it makes sense to run all 3 multicluster jobs for both istiodless and not istiodless. There's a lot of noise in the checking, extra config files, etc., to support both, not to mention 3 more test jobs to run. I could start by sending a PR that just gets the istiodless tests running, but I hate the idea of making them required, running 3 more jobs with every commit. Personally I don't see why we want to propagate the remotes with istiod pattern anyway. I'd rather just bite the bullet and move to the proper implementation for remote clusters. WDYT?

tests/integration/telemetry/stats/prometheus/nullvm/dashboard_test.go
frankbu: I guess it's worth trying to reenable it and see if the postsubmit failure returns. If it does I'll send a followup PR to exclude it again.

frankbu:
I tried to add additional cleanup in this PR, the way the framework sets up access to the remote istiod, but it's more complicated than I thought, so I reverted that change. I'm thinking it's best to merge this one as is and work on more cleanup in a followup future PR.

pkg/test/framework/components/istio/operator.go
```diff
@@ -569,19 +553,19 @@ func installControlPlaneCluster(i *operatorComponent, cfg Config, c cluster.Clus
...
 
-	// Set the clusterName for the local cluster.
-	// This MUST match the clusterName in the remote secret for this cluster.
 	if i.environment.IsMulticluster() {
-		if i.isExternalControlPlane() || cfg.IstiodlessRemotes {
-			// enable namespace controller writing to remote clusters
-			installSettings = append(installSettings, "--set", "values.pilot.env.EXTERNAL_ISTIOD=true")
-		}
+		// Enable namespace controller writing to remote clusters
+		installSettings = append(installSettings, "--set", "values.pilot.env.EXTERNAL_ISTIOD=true")
```
stevenctl, line 558: I think this will have some unintended effects. For _every_ cluster an istiod has secrets for (remote, or another primary), we'll try to do things like webhook cert patch and namespace controller:

https://github.com/istio/istio/blob/1393b4d596fd4d37abdc3d2834842c741efb11fd/pilot/pkg/serviceregistry/kube/controller/multicluster.go#L207-L208

We need a way for the control plane to differentiate.

frankbu: I see what you mean. Interestingly all three istiodless test jobs seem to work with this bug (just lucky timing I suppose?). Anyway, I added back the istiodless checks, so we still run the tests without istiodless remotes, but the postsubmit istiodless jobs seem to be passing now.
 
I suggest we merge this PR now since it includes enough progress and then we can fix the cert patch problem and figure out how to proceed with istiodless in a future PR. If you agree, please approve.

stevenctl: 👍  Agree. I think https://github.com/istio/istio/pull/35286 may be the solution here. Things _should_ still work, but it's not ideal.

manifests/charts/istiod-remote/templates/services.yaml
```diff
@@ -0,0 +1,37 @@
...
+  name: istiod-remote
+  namespace: {{ .Release.Namespace }}
+spec:
+  ports:
+  - port: 15012
+    name: tcp-istiod
+    protocol: TCP
+  clusterIP: None
+  {{- else }}
+# when istiod isn't enabled, we can use istiod service name
```
stevenctl, line 16:
```suggestion
# when local istiod isn't enabled, we can use istiod service name to reach the remote control plane
```

stevenctl (approved):

douglas-reid (approved):
E&T-owned files LGTM.

Issue #35831, opened 2021-11-02
Can no longer build some Istio binaries on a Mac (istio-cni, istio-iptables)


ericvn:
### Bug Description

It appears that https://github.com/istio/istio/pull/35360 has made some changes that now prevent builds from completing error free when using a Mac. The VS Code editor also shows the same 3 errors:
```
# istio.io/istio/tools/istio-iptables/pkg/capture
tools/istio-iptables/pkg/capture/run.go:357:15: undefined: netlink.RuleAdd
tools/istio-iptables/pkg/capture/run.go:376:17: undefined: netlink.SCOPE_HOST
tools/istio-iptables/pkg/capture/run.go:377:17: undefined: unix.RTN_LOCAL
```

Note that the other `build` binaries do seem to build OK.

### Version

```prose
Current commit, 0baff96984f6f7114e2de9763114b757678ca90d, on the `master` branch.
```

esnible:
I am having the same problem.  It is the same if I use `BUILD_WITH_CONTAINER=1 make istioctl`.

therealmitchconnors:
I am able to work around this problem by running the build inside of `sudo make shell`, with the home, hub, and tag env vars persisted.

ericvn:
@esnible I think istioctl will be built even though the error appears (I think only the istio-iptables and istio-cni executables aren't).

I'm not sure if @therealmitchconnors suggestion will work, since you'll still be trying to build the Darwin version of istio-iptables which I would think would fail. If you set TARGET_OS=linux, then you are causing the linux version of the executables to be built so if you don't need the darwin version, you can could use that.

PR #36122, opened 2021-11-17
Fix `make build` failing on MacOS

ericvn:

Fixes https://github.com/istio/istio/issues/35831

This PR moves replaces some of the Linux used resources with stubbed resources on non-Linux hosts. This is a first pass.
The binaries using this code, istio-iptables and istio-cni will not function normally on non-Linux hosts.

Alternative: Submit a PR against the upstream netlinks package so that the call, const used is moved into their _linux.go files. That leaves one unix.RTN_LOCAL change that needed to be worked around in the Istio code.


tools/istio-iptables/pkg/capture/run_linux.go
```diff
@@ -0,0 +1,841 @@
+// Copyright Istio Authors
```
howardjohn, line 1: Do we need to mvoe all of this to run_linux.go?  Could only `ConfigureRoutes` be moved to a new file?

ericvn: Pass #2 with just moving the `configureTProxyRoutes` to a _linux file and stubbing it out for all non-linux builds.

howardjohn (approved):

PR #36132, opened 2021-11-18
Correct bug-report `--include|--exclude` params order

dddddai:

bug-report `--include|--exclude` params order seems incorrect, which may result in parse error
The correct order should be: namespaces/deployments/pods/labels/annotations/**containers**

Please refer to:
https://github.com/istio/istio/blob/03dff9ab756a3058145349d5c52659a3b69bb6b0/tools/bug-report/pkg/config/config.go#L28-L31
https://github.com/istio/istio/blob/03dff9ab756a3058145349d5c52659a3b69bb6b0/tools/bug-report/pkg/config/config.go#L213-L217


dddddai:
@Monkeyanator Excuse me, could you please take a look?

Monkeyanator (approved):

PR #36143, opened 2021-11-18
validation: add validation for re2 max regex size

howardjohn:
Currently, regex over 1024 program size will NACK envoy.

This introduces a new validation error if the program *length* exceeds
1024. Length is always less than program size. This ensures that anyone
this validation error fails for is already broken (by NACKs); we must
approximate as golang doesn't expose a program size.

In order to ensure that users with length<1024 but size>1024, we also
bump the max size enforced by envoy. This is set to a number 4x higher
than I was able to make any regex with length<1024, but with some margin
in case I missed some cases where there are larger sizes, while also not
being so high to allow absurdly high regexs in the worst case.

Overall, this should result in:
* All users with working config: continue to work
* Users with NACK and length<1024: will start working
* Users with NACK and length>1024: will start rejecting in webhook

### DIFF
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -2508,6 +2508,17 @@ func validateStringMatchRegexp(sm *networking.StringMatch, where string) error {
 		return fmt.Errorf("%q: regex string match should not be empty", where)
 	}
 
+	// Envoy enforces a re2.max_program_size.error_level re2 program size is not the same as length,
+	// but it is always *larger* than length. Because goland does not have a way to evaluate the
+	// program size, we approximate by the length. To ensure that a program that is smaller than 1024
+	// length but larger than 1024 size does not enter the system, we program Envoy to allow very large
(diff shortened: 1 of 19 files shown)


pkg/config/validation/validation.go
```diff
@@ -2508,6 +2508,17 @@ func validateStringMatchRegexp(sm *networking.StringMatch, where string) error {
 		return fmt.Errorf("%q: regex string match should not be empty", where)
 	}
 
+	// Envoy enforces a re2.max_program_size.error_level re2 program size is not the same as length,
+	// but it is always *larger* than length. Because goland does not have a way to evaluate the
```
jacob-delgado, line 2512: nit: golang

jacob-delgado (approved):

jacob-delgado:
Should we cherry-pick this to other releases?

linsun (approved):
LGTM thanks John!

Issue #36051, opened 2021-11-12
Too many default metrics in envoy


Stono:
### Bug Description

Hi, 
"back in the day" there was a lot of noise/junk metrics being exported from the sidecar.  All this adds up to:

- increased memory footprint of the sidecar
- increased scrape duration & data transfer to prometheus

Particularly "at scale".

Stats inclusion matches were added (https://istio.io/latest/docs/ops/configuration/telemetry/envoy-stats), and many default metrics that weren't considered essential were removed which helped a lot, but I feel like there's slowly more junk creeping back in that isn't "essential", https://github.com/istio/istio/issues/36049 as an example of what i'm seeing in 1.12.

However pre 1.12, I still think there's _loads_ of stuff that surely isn't essential, for example in this output [prometheus.txt](https://github.com/istio/istio/files/7527885/prometheus.txt), which was generated with:

```
      proxyStatsMatcher:
        inclusionSuffixes:
        - outlier_detection.ejections_active
        - upstream_cx_active
        - upstream_cx_total
```

We have loads of `envoy_cluster_http2.*` metrics, `envoy_server_.*` stuff, `envoy_listener_manager` and `envoy_cluster_manager` type histogram buckets.  Are they really all required?

Considering the work that went into getting the amount of metrics exposed by the sidecar right down, are there any tests tracking that it doesn't regress (like all those additional ones in 1.12)?

> Note including more Envoy stats might increase number of time series collected by prometheus significantly. Care needs to be taken on Prometheus resource provision and configuration to reduce cardinality.

https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#ProxyConfig

### Version

```prose
1.12-beta2
```

howardjohn:
We should at the very least have a test that captures the set of metrics we expose to make sure we are conscious of changes. Maybe its not strictly related to this issue though

Issue #20235, opened 2020-01-16
locality aware routing metrics strange format


Stono:
**Bug description**
The format of the locality aware envoy metrics are a bit odd, and make it difficult to work with in prometheus, and doesn't fit the typical model of exposing cardinality in the labels rather than the metric name

As an example, why is this:
```
envoy_cluster_zone_europe_west4_c_europe_west4_a_upstream_rq_time_sum{cluster_name="outbound|80||app.search-one.svc.cluster.local"} 48534298.75
```

Not just a label on the existing metric, like this:
```
envoy_cluster_upstream_rq_time_sum{source_zone="europe-west4-c", destination_zone="europe-west4-a", cluster_name="outbound|80||app.search-one.svc.cluster.local"} 48534298.75
```

**Expected behavior**
Metrics to expose their cardinality in labels, rather than in the metric name.

**Steps to reproduce the bug**
Enable http metrics on envoy

**Version (include the output of `istioctl version --remote` and `kubectl version` and `helm version` if you used Helm)**
1.4.3

**How was Istio installed?**
Helm

**Environment where bug was observed (cloud vendor, OS, etc)**
GKE 1.15

howardjohn:
I don't think we have any control over this as it's an envoy metric?

brianwolfe:
Stats are parsed out in the bootstrap configuration to create the tags. An additional field in the bootstrap should work to parse those stats out to tags correctly: https://github.com/istio/istio/blob/master/tools/packaging/common/envoy_bootstrap_v2.json#L21-L24

The zone format is specified here: https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats.html?highlight=stats%20zone#per-service-zone-dynamic-http-statistics

howardjohn:
Ah very cool, I didn't realize that config existed

ckcd:
@Stono

ckcd:
Hi @Stono 

> `envoy_cluster_zone_europe_west4_c_europe_west4_a_upstream_rq_time_sum`

just want to know how to generate this metrics in istio 1.4 ?  I'm using below in `meshConfig` and cannot get this metrics in istio-proxy. I already using `sidecar.istio.io/statsInclusionPrefixes: cluster` to allow generate metrics.

```
    localityLbSetting:
      distribute:
      - from: "cn-shanghai/cn-shanghai-e"
        to:
          "cn-shanghai/cn-shanghai-e": 80
[...]

PR #36145, opened 2021-11-18
Add test to validate full set of envoy and istio metrics exposed by istio-agent.  

douglas-reid:
This test will be used to enforce intentional growth in metrics footprint across Istio releases. In prior releases, we have seen growth in the metrics surface, primarily from new envoy metrics exposed via istio-agent. This test (and subsequent changes to it) should help enforce better stability and care over increases in surface.

Related issue: https://github.com/istio/istio/issues/36051


douglas-reid:
@Stono would be interested in your comments, especially on the listed set of existing metrics.

tests/integration/telemetry/stats/prometheus/footprint/footprint_test.go
```diff
@@ -0,0 +1,474 @@
...
+		"envoy_server_version":                                                         true,
+		"envoy_server_wip_protos":                                                      true,
+		"envoy_server_worker_0_watchdog_mega_miss":                                     true,
+		"envoy_server_worker_0_watchdog_miss":                                          true,
+		"envoy_server_worker_1_watchdog_mega_miss":                                     true,
+		"envoy_server_worker_1_watchdog_miss":                                          true,
+		"envoy_server_worker_2_watchdog_mega_miss":                                     true,
+		"envoy_server_worker_2_watchdog_miss":                                          true,
+		"envoy_server_worker_3_watchdog_mega_miss":                                     true,
+		"envoy_server_worker_3_watchdog_miss":                                          true,
```
howardjohn, line 302: This seems likely flaky - I think you will get one per core for ingress, so on a !4 core machine it will fail(?)

Side note: is this a bug? seems worker should be a tag...

tests/integration/telemetry/stats/prometheus/footprint/footprint_test.go
```diff
@@ -0,0 +1,474 @@
...
+		"envoy_cluster_metadata_exchange_alpn_protocol_found":                          true,
+		"envoy_cluster_metadata_exchange_alpn_protocol_not_found":                      true,
+		"envoy_cluster_metadata_exchange_header_not_found":                             true,
+		"envoy_cluster_metadata_exchange_initial_header_not_found":                     true,
+		"envoy_cluster_metadata_exchange_metadata_added":                               true,
+		"envoy_cluster_original_dst_host_invalid":                                      true,
+		"envoy_cluster_retry_or_shadow_abandoned":                                      true,
+		"envoy_cluster_ssl_ciphers_ECDHE_RSA_AES256_GCM_SHA384":                        true,
+		"envoy_cluster_ssl_connection_error":                                           true,
+		"envoy_cluster_ssl_curves_P_256":                                               true,
```
howardjohn, line 155: This will likely work in an isolated test environment but when ran against an Envoy using serving real traffic may differ

tests/integration/telemetry/stats/prometheus/footprint/footprint_test.go
```diff
@@ -0,0 +1,474 @@
...
+		Setup(istio.Setup(common.GetIstioInstance(), setupConfig)).
+		Setup(common.TestSetup).
+		Run()
+}
+
+func setupConfig(_ resource.Context, cfg *istio.Config) {
+	if cfg == nil {
+		return
+	}
+	// enable telemetry v2 with nullvm
```
howardjohn, line 471: do we need a new test suite? Or could we use the existing nullvm one?

tests/integration/telemetry/stats/prometheus/footprint/footprint_test.go
Stono: I raised this here: https://github.com/istio/istio/issues/20235#issuecomment-971538339 because I think it's a bug too

Stono:
Hey, so i'm not sure how your test framework will run this - but the metrics that come out can be heavily influenced by _how_ you use istio.   For example when you turn on locality aware routing, you get additional metrics **once traffic flows through it**?

tools/packaging/common/envoy_bootstrap.json
```diff
@@ -115,6 +115,22 @@
       {
         "tag_name": "authz_dry_run_result",
         "regex": "(\\.shadow_(allowed|denied))"
+      },
+      {
```
Stono, line 119: Is part of this PR adding new default tags too?

tests/integration/telemetry/stats/prometheus/footprint/footprint_test.go
douglas-reid: The actual underlying problem is an Envoy bug-turned-feature, iirc, because it wasn't caught for at least one release.

But yes, this test is also probably flaky, and needs to be fixed.

tests/integration/telemetry/stats/prometheus/footprint/footprint_test.go
douglas-reid: Probably don't need a full test suite. It was cleaner/easier to separate at first, but I can collapse it.

tools/packaging/common/envoy_bootstrap.json
douglas-reid: Yes. I'm trying to correct for some known issues as well. I can separate the PRs as needed.

douglas-reid:
> Hey, so i'm not sure how your test framework will run this - but the metrics that come out can be heavily influenced by _how_ you use istio. For example when you turn on locality aware routing, you get additional metrics **once traffic flows through it**?

Yes. I'm not quite sure how best to exercise *all* (or even most) of the cases where new metrics might pop up. Open to suggestions.

tests/integration/telemetry/stats/prometheus/footprint/footprint_test.go
douglas-reid: moved it.

PR #36168, opened 2021-11-19
Make istioctl upgrade an alias for istioctl install

### LINKED ISSUE #35951: istioctl upgrade: consolidate with install
Currently, upgrade has a few issues:

* Calls to github external - this breaks MANY users
* Unclear when/why to use it vs `install`, given `install` can upgrade and is declarative. Especially with revisions
* Gives bad advice, such as no skip level upgrades.
* Detects future version using `tag`, which may be a random string. `install` has a different method
* Output formatting is very inconsistent with `install` and less polished
* Doesn't actually add any useful checks over `install` other than showing a diff of changes.
* Diff logic is flaky, due to defaulting values you often get false positives.
* Has custom wait logic (`waitUpgradeComplete`). Not sure why this is needed when `install` already has very robust waiting.

I propose we rework it.

Changes I would make:
* Deprecate `upgrade` (but do not remove), make it an alias for `install --show-diff`
* Remove the entire existing `upgrade` path (since its not an alias for `install`)
* Add `--show-diff` option to `install`. This follows the current diff logic for `upgrade`, but should NOT pull from github.com. Instead, create a function to fetch the `installed-state` from the cluster. Also make sure we hit the right code paths to resolve the defaulting issues.

Monkeyanator:
Per discussion in environments working group, we want to remove the existing logic from istioctl upgrade and make it an alias for istioctl install. Follow-ups should add (1) running pre-check command in the upgrade command (requires some refactoring) and (2) reimplementing the diff functionality to not pull from github.

Fixes https://github.com/istio/istio/issues/35951

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience


operator/cmd/mesh/upgrade.go
```diff
@@ -15,633 +15,47 @@
...
+	}
 	cmd := &cobra.Command{
 		Use:   "upgrade",
 		Short: "Upgrade Istio control plane in-place",
-		Long: "The upgrade command checks for upgrade version eligibility and," +
-			" if eligible, upgrades the Istio control plane components in-place. Warning: " +
-			"traffic may be disrupted during upgrade. Please ensure PodDisruptionBudgets " +
-			"are defined to maintain service continuity.",
+		Long: "The upgrade command is an alias for the install command" +
+			" that performs additional upgrade-related checks.",
```
howardjohn, line 53: isn't this not accurate?

howardjohn (approved):

PR #36173, opened 2021-11-19
Fuzzing: Add CRD roundtrip fuzzer

AdamKorcz:
Adds 1 fuzzer.

The crd roundtrip fuzzer is a slightly reduced version of this roundtrip fuzzer. How it has been reduced:

The fuzzer will not check if deep copying is altering the object. The reason here is that I have simply not found an good way to ignore the `XXX_` fields.
The test for the internal and external types are removed with the expectation that it will be added later.
The reason for adding a new version of this fuzzer is that the one in this PR will be able to run continuously through OSS-fuzz.

### DIFF
--- a/tests/fuzz/crd_roundtrip_fuzzer.go
+++ b/tests/fuzz/crd_roundtrip_fuzzer.go
@@ -1,0 +1,210 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


tests/fuzz/pilot_model_fuzzer_to_move.go
```diff
@@ -0,0 +1,240 @@
...
+	ns, err := f.GetString()
+	if err != nil {
+		return 0
+	}
+	wl := make(map[string]string)
+	err = f.FuzzMap(&wl)
+	if err != nil {
+		return 0
+	}
+	result := authzPolicies.ListAuthorizationPolicies(ns, []labels.Instance{wl})
```
howardjohn, line 154: I am not sure what we are trying to catch here? This function seems nearly trivial and can easily get 100% test coverage from unit tests. Is our goal to fuzz 100% of the code base? My expectation is that fuzzing is typically more useful on things like parsing arbitrary inputs, etc.

it seems like we practically re-implement ListAuthorizationPolicies in the test code and check the output is the same.

AdamKorcz: Removed that fuzzer.

AdamKorcz:
It would be great to get this reviewed and merged.

howardjohn (approved):

PR #36175, opened 2021-11-19
Test and add support for Gateway API for eastwest gateway

howardjohn:
This introduces a new API - under tls.options to use AUTO_PASSTHROUGH.
This works around the tls.mode not being extensible. Other than that,
just adds tests and some smartness to set the REQUIRED_NETWORK_VIEW
based on the (already defined) network label.
Put a hold to get feedback from all reviewers


Monkeyanator (approved):
Nice! Next I wonder if we can add the option to use this in test framework for eastwest deployment.

costinm (commented):
Not sure. Autopass as a label to be used on any port sounds super scary. Even on the east west gateway it is scary.

Great to see the east/west converted to new api, but can we just have a narrower way to enable this ? Should be only on a specific port, for workloads in a whitelisted set of namespaces. Nothing else should use this hack - and we need to replace it asap.

Also if we're converting east/west - it may be the time to start terminating istiod conn in the gateway ( and restore spiffe auth of istiod ). But we can discuss it separately.

howardjohn:
> Not sure. Autopass as a label to be used on any port sounds super scary. Even on the east west gateway it is scary.
> 
> Great to see the east/west converted to new api, but can we just have a narrower way to enable this ? Should be only on a specific port, for workloads in a whitelisted set of namespaces. Nothing else should use this hack - and we need to replace it asap.
> 
> Also if we're converting east/west - it may be the time to start terminating istiod conn in the gateway ( and restore spiffe auth of istiod ). But we can discuss it separately.

I think there may be some confusion with what is automated vs my example of how a user would use it?

This PR adds 2 changes to the code:
* If you set `topology.istio.io/network` label, you also get `ISTIO_META_REQUESTED_NETWORK_VIEW` set. This is not dependent on any port, applies to the whole pod
* On a _per port basis_, in `listener.tls.options`, you can set `istio.io/passthrough-mode=AUTO` (note: the field is a `map[string]string` and the spec suggests we prefix with our domain), which uses AUTO_PASSTHROUGH. This is because listener.tls.mode does not have 'auto' mode.

The example in https://github.com/istio/istio/blob/1a343b453aaf7d889ef1a8fb039f2bb4ced0da34/pilot/pkg/config/kube/gateway/testdata/eastwest.yaml is what the user would actually apply. There is no magic ports, and we are not locking into terminating vs passthrough; the user can configure that still

stevenctl (approved):
So much prettier than gen-eastwest-gateway

nmittler (approved):
Nice!

howardjohn:
Per @costinm's recommendation (which I agree with), I have removed the explicit AUTO_PASSTHROUGH option. Instead, we can infer this based on the already established labels that indicate a gateway is a multinetwork gateway - which already includes the port (15443, overrideable by a different label).

@nmittler @stevenctl @Monkeyanator can you take another look?

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -1255,12 +1258,31 @@ func buildListener(r *KubernetesResources, obj config.Config, l k8s.Listener, li
...
+	_, networkSet := obj.Labels[label.TopologyNetwork.Name]
+	if !networkSet {
+		return false
+	}
+	expectedPort := "15443"
+
+	if port, f := obj.Labels[controller.IstioGatewayPortLabel]; f {
+		expectedPort = port
+	}
+	return fmt.Sprint(l.Port) == expectedPort
```
nmittler, line 1277: nit: would `strconv.Itoa` be more efficient than `fmt.Sprint`?

nmittler (approved):

pilot/pkg/config/kube/gateway/testdata/eastwest.yaml
```diff
@@ -0,0 +1,56 @@
...
+    port: 15012
+    protocol: TLS
+    tls:
+      mode: Passthrough
+  - name: istiod-webhook
+    port: 15017
+    protocol: TLS
+    tls:
+      mode: Passthrough
+  - name: cross-network
```
stevenctl, line 21: Would we break if someone specified multiple listeners with the same port but different hosts? e.g. `*.local` and `serviceentry.com`

howardjohn: Good point. In today code they can set a list of hostnames. I will see what happens here

costinm (approved):
Thanks

stevenctl (approved):

PR #36183, opened 2021-11-20
Remove duplicate Grafana Pod port definition.

mt-inside:
Keep the one named "service" because that's what the port's called on the _Service_

Minor fix to the Grafana addon resources.
Note that while `kubectl` is happy to apply this, some systems' pre-checks (eg Flux 2) reject the file because of this error.

- [ ] Docs
- [X] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/samples/addons/grafana.yaml
+++ b/samples/addons/grafana.yaml
@@ -157,9 +157,6 @@ spec:
             - name: service
               containerPort: 3000
               protocol: TCP
-            - name: grafana
-              containerPort: 3000
-              protocol: TCP
           env:
             
             - name: GF_PATHS_DATA


mandarjog:
This seems reasonable @douglas-reid doo you know why we use same port with two names? Is another system looking for name “grafana”?

howardjohn (changes requested):
While the change itself looks reasonable, we are generating this from helm chart: manifests/addons/gen.sh

My guess is the default is port 80+3000, and we want just 3000 for legacy reasons, which ends up duplicated.

mt-inside:
Ah I wasn't aware of that mechanism. This is a bit weirder than I expected.

Grafana with default config only listens on 3000, doesn't open 80. Verified with netstat after running docker.io/grafana/grafana:8.2.5 locally.

By default the helm chart causes the Deployment to declare port 3000 (hard-coded) AND port 80 ([source](https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/_pod.tpl#L316-L322)).
* Port number 3000 is hard-coded
* Port 80 comes from the Service part of the values object. This looks like a straight-up bug, a spurious copy-and-paste from the [Service definition](https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/service.yaml#L38-L40)

Sadly, getting the Service on 3000, as Istio needs, will always result in a duplicate 3000 port on the Deployment, because the value is re-used.

I've raised upstream: https://github.com/grafana/helm-charts/pull/844

Suggest we leave this bug open until they merge?

mt-inside:
FYI I'm still chasing this error upstream with Grafana - it was closed without merge or comment but I've asked what the deal is.

PR #36200, opened 2021-11-22
Remove dependency on k8s.io/kubernetes

howardjohn:
This is not actually used, just a 100 step removed indirect dependency
that upstreams are not accepting fixes to remove. So we can just
explicitly drop it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,9 @@ module istio.io/istio
 
 go 1.17
 
+// https://github.com/containerd/containerd/issues/5781
+exclude k8s.io/kubernetes v1.13.0
+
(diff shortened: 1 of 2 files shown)


ericvn (approved):

costinm:
Why do we depend on containerd in the first place ? Can we just exclude them.

But the bigger question - are we ready to start having modular deps, i.e. a go.mod in pkg/, one in operator/, one for istioctl, etc ?

howardjohn:
We depend on containerd because of helm (which has it for OCI images) and
go-containerregistry (which we import for WASM).

For the bigger question - I don't know, would need to do some research of
the implications

PR #36205, opened 2021-11-22
exposing ports on pod

kevin-lindsay-1:

I'm sure you guys have plans here already, but I was unable to get an ingress gateway to work without a strategic merge patch because these ports were not exposed, and didn't appear to be configurable.

Feel free to do whatever you want with the PR, I'm just creating it as more-or-less a `todo`.

### DIFF
--- a/manifests/charts/gateway/templates/deployment.yaml
+++ b/manifests/charts/gateway/templates/deployment.yaml
@@ -80,6 +80,18 @@ spec:
             value: {{ $val }}
           {{- end }}
           ports:
+          - containerPort: 15021
+            protocol: TCP
+          - containerPort: 8080
+            protocol: TCP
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
I am a bit confused why this is needed for merging to work?

My understanding is in general containerPort is almost never required, unless some third party tools are using it (like prometheus).

if we do this, we would at least need to derive it from the ports config in values.yaml

kevin-lindsay-1:
No idea, all I know is upgraded, waited for the ELB in AWS to provision, and it wasn't working.

Diffed the old deployment, and those ports used to be exposed.

I have no attachment to this PR, I only created it because it wasn't working and this seemed to fix my issue pretty much instantly.

PR #36142, opened 2021-11-18
Sidecar in Hybrid Mode

shriramsharma:
**Please provide a description of this PR:**
Fixes https://github.com/istio/istio/issues/35111
Extended sidecar to perform gateway functions in order to enable TLS from outside of mesh.
RFC: [Support hybrid sidecar mode](https://docs.google.com/document/d/15Qhr7errbylXEzxxCK7ij_oUpn4E5SFU2uDdl_n2GIc/edit#heading=h.xw1gqgyqs5b)

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [X] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #36210, opened 2021-11-23
Sidecar in Hybrid Mode

### LINKED ISSUE #35111: Enable TLS termination on the envoy sidecar

Enable TLS termination on the envoy sidecar.
Currently, the server side envoy proxy running as a sidecar can only allow mTLS for secure connection. In a regular Istio mesh deployment the onus of one way TLS termination for downstream requests is on the ingress gateway. Though this satisfies most of the use cases, for some use cases (like an API Gateway on mesh) the ingress gateway resource is not necessarily needed.

This feature request is to introduce an annotation "sidecar.istio.io/proxyMode: true" on a deployment that runs the Istio proxy on the pod in hybrid mode, i.e., it performs a function of both a router/gateway and a sidecar.

1. Create a custom EnvoyFilter configuration as described in [this](https://github.com/istio/istio/issues/21368) issue. The problem with this approach is that this could lead to maintenance overhead as EnvoyFilters does not guarantee backwards compatibility. 
2. Use the Ingress Gateway resource to offload the TLS termination. Going this route forces us to add an unnecessary network hop which can be avoided by this feature.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**
Related Issue: https://github.com/istio/istio/issues/21368
RFC (WIP): https://docs.google.com/document/d/15Qhr7errbylXEzxxCK7ij_oUpn4E5SFU2uDdl_n2GIc/edit?usp=sharing

shriramsharma:
Fixes https://github.com/istio/istio/issues/35111
Extended sidecar to perform gateway functions in order to enable TLS from outside of mesh.
RFC: [Support hybrid sidecar mode](https://docs.google.com/document/d/15Qhr7errbylXEzxxCK7ij_oUpn4E5SFU2uDdl_n2GIc/edit#)

---
**NOTE**

I had mistakenly created a PR on release-1.12 branch. Here is the old [PR](https://github.com/istio/istio/pull/36142) for historical reference.

---


go.mod
```diff
@@ -15,6 +15,8 @@ replace github.com/docker/docker => github.com/moby/moby v17.12.0-ce-rc1.0.20200
 // Client-go does not handle different versions of mergo due to some breaking changes - use the matching version
 replace github.com/imdario/mergo => github.com/imdario/mergo v0.3.5
 
+replace istio.io/api => github.com/shriramsharma/api v0.0.0-20211120145828-7ad2eb296068
```
adilfulara, line 18: TODO: Remove

go.sum
```diff
@@ -1089,6 +1089,8 @@ github.com/sergi/go-diff v1.1.0 h1:we8PVUC3FE2uYfodKH/nBHMSetSfHDR6scGdBi+erh0=
 github.com/sergi/go-diff v1.1.0/go.mod h1:STckp+ISIX8hZLjrqAeVduY0gWCT9IjLuqbuNXdaHfM=
 github.com/shopspring/decimal v1.2.0 h1:abSATXmQEYyShuxI4/vyW3tV1MrKAJzCZ/0zLUXYbsQ=
 github.com/shopspring/decimal v1.2.0/go.mod h1:DKyhrW/HYNuLGql+MJL6WCR6knT2jwCFRcu2hWCYk4o=
+github.com/shriramsharma/api v0.0.0-20211122132636-909a14111939 h1:syrBv17W+82p28MhJhxeXbEUgMDubO+FHrTsSGMPXQo=
+github.com/shriramsharma/api v0.0.0-20211122132636-909a14111939/go.mod h1:lavaUNsnT7RGyMFNOGgV5XvOgP3fkTSZkxP/0H/ISt4=
```
adilfulara, line 1093: TODO: Remove

howardjohn:
This doesn't currently build, if you can get it compiling it will be easier to review. Thanks!

manifests/charts/istio-control/istio-discovery/files/injection-template.yaml
```diff
@@ -209,6 +209,8 @@ spec:
           - wait
   {{- end }}
     env:
+    - name: SIDECAR_IN_HYBRID_MODE
```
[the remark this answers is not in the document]
shriramsharma: thanks @howardjohn  ... I'll remove this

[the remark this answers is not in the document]
shriramsharma: @howardjohn I have removed this

pilot/pkg/model/context.go
```diff
@@ -836,7 +836,7 @@ func (node *Proxy) BuildCatchAllVirtualHost() {
 // fetch all the gateways and invoke the merge call in multiple places (lds/rds).
 // Must be called after ServiceInstances are set
 func (node *Proxy) SetGatewaysForProxy(ps *PushContext) {
-	if node.Type != Router {
+	if node.Type == SidecarProxy {
```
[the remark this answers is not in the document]
shriramsharma: this is done

shriramsharma:
> This doesn't currently build, if you can get it compiling it will be easier to review. Thanks!

@howardjohn , I added the [istio/api replace directive ](https://github.com/istio/istio/pull/36210/commits/4468f3378d7256140413b0546b8e0111717d0a88) in the `go.mod` file so that the build passes. Once [this PR](https://github.com/istio/api/pull/2165) is merged, I'll remove this and update the version. I'm not sure if there is better way to make this work in the interim.

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -115,6 +117,70 @@ func (configgen *ConfigGeneratorImpl) BuildListeners(node *model.Proxy,
 	return builder.getListeners()
 }
 
+func (configgen *ConfigGeneratorImpl) BuildListenerTLSContext(serverTLSSettings *networking.ServerTLSSettings, proxy *model.Proxy, transportProtocol istionetworking.TransportProtocol) *auth.DownstreamTlsContext {
+	alpnByTransport := util.ALPNHttp
+	if transportProtocol == istionetworking.TransportProtocolQUIC {
+		alpnByTransport = util.ALPNHttp3OverQUIC
```
howardjohn, line 123: I am not sure we can support QUIC, we don't have redirection support

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -598,6 +598,24 @@ func (configgen *ConfigGeneratorImpl) buildInboundFilterchains(in *plugin.InputP
 	matchingIP string, clusterName string, passthrough bool) []*filterChainOpts {
 	mtlsConfigs := getMtlsSettings(configgen, in, passthrough)
 	newOpts := []*fcOpts{}
+	isHybrid := false
+	if listenerOpts.tlsSettings != nil {
+		downstreamTLSContext := configgen.BuildListenerTLSContext(listenerOpts.tlsSettings, in.Node, istionetworking.TransportProtocolTCP)
+		gatewayTLSConfig := &plugin.MTLSSettings{
+			Port: in.ServiceInstance.Endpoint.EndpointPort,
+			Mode: model.MTLSPermissive,
+			HTTP: downstreamTLSContext,
```
howardjohn, line 607: Setting `HTTP` here doesn't seem right. Supported modes are:
* `TLS`: tcp over TLS
* `HTTPS`: http over TLS

This is also a bit of an abuse since this is *m*TLS settings. The end result is also wrong, as we end up with a filter chain match like: `{"destinationPort":80,"transportProtocol":"tls","applicationProtocols":["http/1.0","http/1.1","h2c"]}` - match `h2c` over TLS doesn't seem right, should be `h2`?

Current config looks like:
```
{"fcm":{"destinationPort":80,"transportProtocol":"raw_buffer"},"tls":"default"}
{"fcm":{"destinationPort":80,"transportProtocol":"tls","applicationProtocols":["http/1.0","http/1.1","h2c"]},"tls":"file-cert:/etc/certs/user/tls.crt~/etc/certs/user/tls.key"}
{"fcm":{"destinationPort":80,"transportProtocol":"tls","applicationProtocols":["istio","istio-peer-exchange","istio-http/1.0","istio-http/1.1","istio-h2"]},"tls":"default"}
{"fcm":{"destinationPort":80,"transportProtocol":"tls"},"tls":"default"}
```
(from `istioctl pc l svc/echo --port 15006 -ojson | jq '.[].filterChains[] | {fcm: .filterChainMatch, tls: .transportSocket.typedConfig.commonTlsContext.tlsCertificateSdsSecretConfigs[0].name}' -c | sort | jq -c`)

So:
* plaintext we terminate mTLS. This cannot work.
* For tls with `applicationProtocols":["http/1.0","http/1.1","h2c"]` we terminate with user certs. 
* For anything else we use mTLS.

I am not sure how we should handle plaintext. It seems like it should be allowed, provided they are in STRICT mode. That makes the `tls` config additive. 

`tls` should not required specific ALPNs, IMO. If they are Istio ALPNs, we can treat it as mTLS. if not, it should use the user config.

Ideally all the cases of: `{incoming TLS, HTTPS, HTTP, TCP} x {ALPN} x {Sidecar config}` are clearly defined - probably in `TestInbound` of sidecar_simulation_test.go

pkg/config/validation/validation.go
```diff
@@ -1025,6 +1025,18 @@ var ValidateSidecar = registerValidateFunc("ValidateSidecar",
...
+
+			if i.Tls != nil {
+				if i.Tls.Mode == networking.ServerTLSSettings_ISTIO_MUTUAL || i.Tls.Mode == networking.ServerTLSSettings_AUTO_PASSTHROUGH {
+					errs = appendValidation(errs, fmt.Errorf("configuration is invalid: cannot set mode to %s in sidecar ingress tls", i.Tls.Mode.String()))
+				}
+				protocol := protocol.Parse(i.Port.Protocol)
+				if !protocol.IsTLS() {
+					errs = appendValidation(errs, fmt.Errorf("server cannot have TLS settings for non HTTPS/TLS ports"))
+				}
+				errs = appendValidation(errs, validateTLSOptions(i.Tls))
```
howardjohn, line 1037: Should also reject `httpsRedirect`? or support it, but would prefer to reject.

`subjectAltNames` - should probably also be rejected. I don't think it makes much sense to match on SAN when we only allow a single TLS config to apply (and thus there is no SNI routing, unlike at gateway).

manifests/charts/istio-control/istio-discovery/values.yaml
```diff
@@ -407,6 +416,9 @@ global:
   # Currently, two options are supported: "third-party-jwt" and "first-party-jwt".
   jwtPolicy: "third-party-jwt"
 
+  # This flag can be set true in order to be able to run the sidecar in gateway mode.
+  sidecarInHybridMode: false
```
kfaseela, line 420: are we using this flag still?

pilot/pkg/networking/core/v1alpha3/listener.go
shriramsharma: @howardjohn Actually I just yanked this code from gateway and created a new func so that both GW and Sidecar can call it. So this may make sense for GW.

manifests/charts/istio-control/istio-discovery/values.yaml
shriramsharma: @kfaseela , I'll remove this. You're right, this flag is not used anymore.

kfaseela:
@shriramsharma : Would it make sense to have a documentation PR on istio.io as well, to explain the sample config and demo on how things work end to end?

pilot/pkg/networking/core/v1alpha3/listener_builder.go
shriramsharma: @howardjohn , I made the changes based on our offline conversations.
So, unless the PeerAuth for is set to DISABLE either ns wide or on the specific port, the TLS settings won’t take effect.
Something like this
```
kind: PeerAuthentication
metadata:
  name: httpbin-pa
  namespace: app1
spec:
  selector:
    matchLabels:
      app: httpbin
  mtls:
    mode: STRICT
  portLevelMtls:
    80:
      mode: DISABLE
```
in this ☝️  case, the filterchain would look like this
```
{"fcm":{"destinationPort":15006},"tls":null}
{"fcm":{"destinationPort":80,"transportProtocol":"tls"},"tls":"file-cert:/etc/istio/sidecargateway-certs/tls.crt~/etc/istio/sidecargateway-certs/tls.key"}
```
I dropped the plaintext match
If the PeerAuth is set to STRICT/PERMISSIVE NS wide, then TLS won't take effect. The filterchain match would look something like this.
```
{"fcm":{"destinationPort":15006},"tls":null}
{"fcm":{"destinationPort":80,"transportProtocol":"raw_buffer"},"tls":null}
{"fcm":{"destinationPort":80,"transportProtocol":"tls","applicationProtocols":["istio","istio-peer-exchange","istio-http/1.0","istio-http/1.1","istio-h2"]},"tls":"default"}
```
for STRICT you wont have ALPN match) 🙂  … basically it works as is
Please let me know if this makes sense

pkg/config/validation/validation.go
shriramsharma: @howardjohn ... added the following validations

shriramsharma:
Here is the sample yaml configuration for this feature

```
apiVersion: v1
kind: Service
metadata:
  name: httpbin
  labels:
    app: httpbin
    service: httpbin
spec:
  ports:
  - name: http
    port: 8443
    targetPort: 80
  selector:
    app: httpbin
---

kind: Sidecar
metadata:
  name: httpbin
  namespace: app1
spec:
  workloadSelector:
    labels:
      app: httpbin
  ingress:
    - port:
        number: 80
        name: http
        protocol: HTTPS
      defaultEndpoint: 127.0.0.1:80
      tls:
        mode: SIMPLE
        privateKey: "/etc/certs/tls.key"
        serverCertificate: "/etc/certs/tls.crt"
---

kind: PeerAuthentication
metadata:
  name: httpbin-pa
  namespace: app1
spec:
  selector:
    matchLabels:
      app: httpbin
  mtls:
    mode: STRICT
  portLevelMtls:
    80:
      mode: DISABLE
```

go.mod
```diff
@@ -18,6 +18,8 @@ replace github.com/docker/docker => github.com/moby/moby v17.12.0-ce-rc1.0.20200
 // Client-go does not handle different versions of mergo due to some breaking changes - use the matching version
```
howardjohn, line 18: Overall looks really good, I think this is close. Main concern is the cluster building. We are building the clusters like HTTPS/TLS clusters, which have minimal config. This leads to weird behavior like `curl https://foo --http2` negiotating `h2` ALPN, but then forwarding http/1.1 to the app which seems problematic

howardjohn: Thinking `HttpProtocolOptions_AutoHttpConfig` could plausibly be used for this. Never used before in Istio though so may be suspicious.

I am ok merging in current state if we make it experimental behind an off-by-default flag (so we don't need to worry about breaking change when we fix it).

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -588,15 +588,37 @@ func buildInboundCatchAllFilterChains(configgen *ConfigGeneratorImpl,
...
+	mtlsConfigs := getMtlsSettings(configgen, in, passthrough)
 	for _, mtlsConfig := range mtlsConfigs {
 		for _, match := range getFilterChainMatchOptions(mtlsConfig, listenerOpts.protocol) {
 			opt := fcOpts{matchOpts: match}.populateFilterChain(mtlsConfig, mtlsConfig.Port, matchingIP)
 			newOpts = append(newOpts, &opt)
 		}
 	}
 
+	// unless the PeerAuthentication is set to "DISABLE",
+	// TLS settings won't take effect
```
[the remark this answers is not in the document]
shriramsharma: @howardjohn , I can add debug/warn logs that could spit out some informative message?

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -588,15 +588,37 @@ func buildInboundCatchAllFilterChains(configgen *ConfigGeneratorImpl,
...
+	hasMTLs := false
```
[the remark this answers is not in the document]
shriramsharma: @howardjohn , I've fixed this

go.mod
shriramsharma: hey @howardjohn , since we are getting too close to the release, would it be ok if we can create an issue for this and try to tackle it in future releases?
Based on your comment, I added the feature flag.
https://github.com/istio/istio/pull/36210/commits/54948eb40db0af6353dfdde4238486aee8cf24ab

shriramsharma: @howardjohn , I've added `hide-from-docs` for now

go.mod
```diff
@@ -18,6 +18,8 @@ replace github.com/docker/docker => github.com/moby/moby v17.12.0-ce-rc1.0.20200
...
 replace github.com/imdario/mergo => github.com/imdario/mergo v0.3.5
 
+replace istio.io/api => github.com/shriramsharma/api v0.0.0-20220111202215-d4772d585f73
```
kfaseela, line 21: now that your api changes are merged, change these?

pilot/pkg/features/pilot.go
```diff
@@ -571,6 +571,9 @@ var (
 	PrioritizedLeaderElection = env.RegisterBoolVar("PRIORITIZED_LEADER_ELECTION", true,
```
howardjohn, line 571: can't comment on it directly - but go.mod and go.sum permission got changed. can you revert that?

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -656,60 +655,15 @@ func buildGatewayConnectionManager(proxyConfig *meshconfig.ProxyConfig, node *mo
...
-	if len(server.Tls.CipherSuites) > 0 ||
-		server.Tls.MinProtocolVersion != networking.ServerTLSSettings_TLS_AUTO ||
-		server.Tls.MaxProtocolVersion != networking.ServerTLSSettings_TLS_AUTO {
-		ctx.CommonTlsContext.TlsParams = &tls.TlsParameters{
-			TlsMinimumProtocolVersion: convertTLSProtocol(server.Tls.MinProtocolVersion),
-			TlsMaximumProtocolVersion: convertTLSProtocol(server.Tls.MaxProtocolVersion),
-			CipherSuites:              filteredCipherSuites(server),
-		}
-	}
+	server.Tls.CipherSuites = filteredGatewayCipherSuites(server)
```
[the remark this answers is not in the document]
howardjohn: would be good to fix this up

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -588,15 +588,36 @@ func buildInboundCatchAllFilterChains(configgen *ConfigGeneratorImpl,
...
 	newOpts := []*fcOpts{}
+
+	// unless the PeerAuthentication is set to "DISABLE",
+	// TLS settings won't take effect
+	hasMTLs := true
+
+	mtlsConfigs := getMtlsSettings(configgen, in, passthrough)
 	for _, mtlsConfig := range mtlsConfigs {
+		if mtlsConfig.Mode == model.MTLSDisable {
+			hasMTLs = false
```
howardjohn, line 600: I think this is not quite right. `mtlsConfigs` could contain [STRICT, DISABLE]

howardjohn (approved):

Issue #36160, opened 2021-11-19
istioctl analyze samples/bookinfo/networking/ has a number of new errors


ericvn:
### Bug Description

The current istio.io tests run `istioctl analyze samples/bookinfo/networking/ ` on the `master` branch yields a number of new errors (this is one of the commands in the documentation (https://preliminary.istio.io/latest/docs/ops/diagnostic-tools/istioctl-analyze/#analyzing-live-clusters-local-files-or-both)):
```
istioctl analyze samples/bookinfo/networking/ 
Skipping file samples/bookinfo/networking/bookinfo-gateway.yaml.bak, recognized file extensions are: [.json .yaml .yml]
Error(s) adding files: 17 errors occurred:
	* cannot store config {networking.istio.io/v1alpha3/DestinationRule  productpage default  map[] map[istiofilefieldmap:{"{.apiVersion}":1,"{.kind}":2,"{.metadata.name}":4,"{.spec.host}":6,"{.spec.subsets[0].labels.version}":10,"{.spec.subsets[0].name}":8} istiosource:{"Filename":"samples/bookinfo/networking/destination-rule-all.yaml","Line":1}] v9 0001-01-01 00:00:00 +0000 UTC [] 0} from reader: item already exists
	* cannot store config {networking.istio.io/v1alpha3/DestinationRule  reviews default  map[] map[istiofilefieldmap:{"{.apiVersion}":1,"{.kind}":2,"{.metadata.name}":4,"{.spec.host}":6,"{.spec.subsets[0].labels.version}":13,"{.spec.subsets[0].name}":11,"{.spec.subsets[1].labels.version}":16,"{.spec.subsets[1].name}":14,"{.spec.subsets[2].labels.version}":19,"{.spec.subsets[2].name}":17,"{.spec.trafficPolicy.loadBalancer.simple}":9} istiosource:{"Filename":"samples/bookinfo/networking/destination-rule-reviews.yaml","Line":1}] v10 0001-01-01 00:00:00 +0000 UTC [] 0} from reader: item already exists
[...]

PR #36214, opened 2021-11-23
Fix istioctl analyze having new config store errors

hanxiaop:
Fix https://github.com/istio/istio/issues/36160

This fix is consistent with the old `ApplyContent` logic.

/cc @therealmitchconnors


howardjohn (commented):
CAn you add a test?

monsteredc:
@howardjohn PTAL. Thanks

monsteredc:
Revised with the logic: if the config exists, set the `ResourceVersion` of the new config to the old one in order to perform successful update.

hanxiaop:
not stale

hanxiaop:
not stale

hanxiaop:
not stale

hanxiaop:
https://github.com/istio/istio/blob/d2b96bb9572875fee561c7c34e1e42f58e04c88c/pilot/pkg/config/memory/store.go#L204-L212
Looked through the code, the error is due to the conflict in update. Set the newer config `ResourceVersion` to the annotation part can avoid of `ResourceVersion` conflicts.

pilot/pkg/config/file/store.go
```diff
@@ -204,7 +204,14 @@ func (s *KubeSource) ApplyContent(name, yamlText string) error {
...
 		if !found || oldSha != r.sha {
 			s.versionCtr++
-			r.config.ResourceVersion = fmt.Sprintf("v%d", s.versionCtr)
+			if !found {
+				r.config.ResourceVersion = fmt.Sprintf("v%d", s.versionCtr)
+			} else {
+				if r.config.Annotations == nil {
+					r.config.Annotations = map[string]string{}
+				}
+				r.config.Annotations[memory.ResourceVersion] = fmt.Sprintf("v%d", s.versionCtr)
```
howardjohn, line 213: I get why we need L208, but not why we need this. Can you add test/comments

hanxiaop:
not stale

zirain:
not stale

pilot/pkg/config/file/store.go
```diff
@@ -204,7 +204,17 @@ func (s *KubeSource) ApplyContent(name, yamlText string) error {
 		oldSha, found := s.shas[key]
 		if !found || oldSha != r.sha {
 			s.versionCtr++
-			r.config.ResourceVersion = fmt.Sprintf("v%d", s.versionCtr)
+			if !found {
+				r.config.ResourceVersion = fmt.Sprintf("v%d", s.versionCtr)
+			} else {
+				// If the old resource version exists, and we want to overwrite the old object to the new object,
+				// we should set the resource version in the annotation to avoid of conflicting resource reported
```
therealmitchconnors, line 211: If I understand this correctly, we are allowing subsequent copies of a resource to override the initial copies.  Is that correct?  If so, what determines the order?

hanxiaop: @therealmitchconnors Yes correct. The newest content will be the one to be analyzed. The logic I refered to was from the old branch before this https://github.com/istio/istio/pull/35764 got merged. Which explained here:

https://github.com/istio/istio/blob/842115039209d938d6bc9f96736f1ca62e2f8f6f/galley/pkg/config/source/kube/inmemory/kubesource.go#L142-L146

And the code is:

https://github.com/istio/istio/blob/842115039209d938d6bc9f96736f1ca62e2f8f6f/galley/pkg/config/source/kube/inmemory/kubesource.go#L160-L166

hanxiaop:
not stale

Issue #36155, opened 2021-11-19
Workload Names fail on DeploymentConfig/ReplicationController controllers


lucasponce:
The workload names used in Istio telemetry are provided by environment variables on injection time.

When a user deploys a DeploymentConfig controller (this controller creates a ReplicationController child that is in charged to generate the Pods) the logic to identify the workload name doesn't work correctly [1], and it introduces some false workloads.

So, I think this logic should be updated to contemplate these workloads as well.

This looks a regression from the old code used in the mixer component [2], so, in recent versions it should apply some similar way to resolve the parent workload name.

[1] https://github.com/istio/istio/blob/master/pkg/kube/util.go#L230
[2] https://github.com/istio/istio/blob/1.6.14/mixer/adapter/kubernetesenv/cache.go#L165

PR #36220, opened 2021-11-23
Add DeploymentConfig to GetDeployMetaFromPod

monsteredc:
Related to https://github.com/istio/istio/issues/36155.

@bianpengyuan @lucasponce


pkg/kube/util.go
```diff
@@ -231,6 +231,9 @@ func GetDeployMetaFromPod(pod *kubeApiCore.Pod) (metav1.ObjectMeta, metav1.TypeM
 				name := strings.TrimSuffix(controllerRef.Name, "-"+pod.Labels["pod-template-hash"])
 				deployMeta.Name = name
 				typeMetadata.Kind = "Deployment"
+			} else if typeMetadata.Kind == "ReplicationController" && pod.Labels["deploymentconfig"] != "" {
+				deployMeta.Name = pod.Labels["deploymentconfig"]
```
howardjohn, line 235: what is this 'deploymentconfig' label? I don't see any reference is kubernetes.

howardjohn (commented):
Can you add a test case for this?

pkg/kube/util.go
bianpengyuan: It is an openshift resource: https://docs.openshift.com/container-platform/4.1/applications/deployments/what-deployments-are.html#deployments-and-deploymentconfigs_what-deployments-are We need to comment it well here and point to the openshift reference for this label.

bianpengyuan:
@lucasponce please take a look at this change. thanks!

pkg/kube/util.go
hzxuzhonghu: `ReplicationController` was replaced by deployment several years ago. Should be no use anymore

monsteredc: For the openshift platform, pods which are controlled by `ReplicationController` can still be created by `DeploymentConfig`.

lucasponce:
> @lucasponce please take a look at this change. thanks!

Patch looks ok, thanks !

DeploymentConfig are old controllers, but there are a large population of OpenShift users that still use them, and they want to add them to the mesh without migrating, so this patch will help them to get the proper name of the workload.

pkg/kube/util.go
```diff
@@ -231,6 +231,19 @@ func GetDeployMetaFromPod(pod *kubeApiCore.Pod) (metav1.ObjectMeta, metav1.TypeM
...
+				//
+				// nolint: lll
+				// For DeploymentConfig details, refer to
+				// https://docs.openshift.com/container-platform/4.1/applications/deployments/what-deployments-are.html#deployments-and-deploymentconfigs_what-deployments-are
+				//
+				// For the reference to the pod label 'deploymentconfig', refer to
+				// https://github.com/openshift/library-go/blob/7a65fdb398e28782ee1650959a5e0419121e97ae/pkg/apps/appsutil/const.go#L25
+				deployMeta.Name = pod.Labels["deploymentconfig"]
+				typeMetadata.Kind = "DeploymentConfig"
+				delete(deployMeta.Labels, "deploymentconfig")
```
bianpengyuan, line 246: Why do we delete this label here?

bianpengyuan: Actually it makes sense, since this label comes should only exist at the pod.

bianpengyuan (approved):

Issue #36182, opened 2021-11-20
istioctl proxy-config log fails when it is used with label selector


bseenu:
### Bug Description

```
bash-5.1$ istioctl pc log -l app=istio-ingressgateway -n istio-system
Error: failed to execute command on Envoy: failed retrieving: resource name may not be empty in the "istio-system" namespace
```

### Version

```prose
bash-5.1$ istioctl version
client version: 1.11.4
control plane version: 1.11.4
data plane version: 1.11.4 (17 proxies)
```

ldemailly:
regression from #27490 it seems

PR #36223, opened 2021-11-23
Fix  label selector not select default pod

kebe7jun:
Fix https://github.com/istio/istio/issues/36182

### DIFF
--- a/istioctl/cmd/proxyconfig.go
+++ b/istioctl/cmd/proxyconfig.go
@@ -591,6 +591,9 @@ func logCmd() *cobra.Command {
 				if err != nil {
 					return err
 				}
+				if len(podNames) > 0 {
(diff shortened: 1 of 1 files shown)


istioctl/cmd/proxyconfig.go
```diff
@@ -591,6 +591,11 @@ func logCmd() *cobra.Command {
 				if err != nil {
 					return err
 				}
+				if len(podNames) > 1 {
+					return fmt.Errorf("multiple pods(%v) selected, you should chose one specified pod", podNames)
```
shamsher31, line 595: Instead of returning an error, you should select the first pod by default, or ask a user to pass pod name as a selector?

shamsher31 (approved):

PR #36233, opened 2021-11-24
Cleanup and remove listwatch usage

howardjohn:
This is simply a refactoring; no behavioral changes for users.

Currently, Chiron code takes a list of namespaces. To support this, it needs to use this multins listwatch code. This code is complex and has no other usage. However, inspection of the chiron code shows that the list is always [SystemNamespace, SystemNamespace, SystemNamespace, ..]. So instead we can just pass a single namespace item and clean up the whole package.


pilot/pkg/bootstrap/certcontroller.go
```diff
@@ -70,14 +70,13 @@ func (s *Server) initCertController(args *PilotArgs) error {
 			// Chiron will generate the key and certificate and save them in a secret
 			secretNames = append(secretNames, c.GetSecretName())
 			dnsNames = append(dnsNames, name)
-			namespaces = append(namespaces, args.Namespace)
```
howardjohn, line 73: Basically we had a bunch of code to support watching multiple namespaces. But passed it a list of the same namespace repeated over and over

hzxuzhonghu (approved):

nmittler (approved):

lei-tang (approved):

PR #36234, opened 2021-11-24
Cleanup legacy galley code

howardjohn:
In a recent PR, we removed a huge portion of galley. However, quite a bit was left behind in legacy/ folder due to a tangled usage.

The primary linchpin here was https://github.com/istio/istio/blob/d0748d7c46c02e1fddf77d7a38f90cb3620e4358/pkg/config/analysis/local/istiod_analyze.go#L108

Essentially, for analyze we want to be able to filter to a subset of collections to look at. This was drawing from a number of sources; one of these was `transformerProviders`. This is ALWAYS all of the collections, so we can just replace it with `collections.All`. With that cleaned up, there is essentially no real usage of galley code - it is highly connected to itself through extensive unit tests, but can be untangled without impacting any real code.

This cleans up the following:
* Everything under legacy/. This is things like Source, Event, Transformer, Provider, etc - all of the core galley data model
* Cleanup metadata. No more transformers, providers
* Cleanup metadata more. No more duplicate k8s/xyz and istio/xyz collections. Each kind has one collection. I am quite sure we can simplify this more but didn't want to feature creep too much.
* Analyzers - you will notice SkippedAnalyzers is removed as it is impossible for it to ever be non-empty
* Stop all usage of 'schema.Get()' - there is no need to access the raw AST stuff when we already have collections.gen.go with the same information. This is simpler and more efficient

Note to reviewers: Hit 'filter filter' and uncheck 'deleted files'


pkg/config/analysis/local/local.go
```diff
@@ -29,6 +29,5 @@ const (
 // AnalysisResult represents the returnable results of an analysis execution
 type AnalysisResult struct {
 	Messages          diag.Messages
-	SkippedAnalyzers  []string
```
therealmitchconnors, line 32: SkippedAnalyzers can be set when an analyzer requires an input that is not provided by any of the specified sources.  These cases are rare, but when they occur, this field is the only indication that the analyzer did not run and pass.  I can see how this should be more rare, now that we have arbitraryStore, but I'm nervous about missing an edge case, and reporting skipped analyzers as having passed...

pkg/config/schema/collections/collections.gen.go
```diff
@@ -517,25 +517,6 @@ var (
 		}.MustBuild(),
 	}.MustBuild()
 
-	// K8SExtensionsIstioIoV1Alpha1Wasmplugins describes the collection
-	// k8s/extensions.istio.io/v1alpha1/wasmplugins
-	K8SExtensionsIstioIoV1Alpha1Wasmplugins = collection.Builder{
```
therealmitchconnors, line 522: +100!  Working with istio config will be so much easier without these.

pkg/config/analysis/analyzer_test.go
```diff
@@ -79,63 +73,12 @@ func TestCombinedAnalyzer(t *testing.T) {
 	avalableSchemas.Add(&testSchemaImpl{col2.Name()})
```
therealmitchconnors, line 73: availableSchemas appears to be unused...

pkg/config/analysis/analyzer_test.go
```diff
@@ -79,63 +73,12 @@ func TestCombinedAnalyzer(t *testing.T) {
...
-	expectCollections(g, getDisabledOutputs(collection.Names{in4.Name()}, xformProviders), collection.Names{out4.Name()})
-}
-
-func expectCollections(g *GomegaWithT, actualSet map[collection.Name]struct{}, expectedCols collection.Names) {
-	g.Expect(actualSet).To(HaveLen(len(expectedCols)))
-	for _, col := range expectedCols {
-		g.Expect(actualSet).To(HaveKey(col))
-	}
+	g.Expect(a3.ran).To(BeTrue())
+	g.Expect(a4.ran).To(BeTrue())
```
therealmitchconnors, line 81: This is concerning.  The point of this test is to have analyzers requiring collections 1-4, but only provide collections 1-3, meaning analyzer 4 should not be run.  Inverting the success criteria of a test looks like perhaps a bug?

pkg/config/analysis/incluster/controller.go
```diff
@@ -45,20 +43,15 @@ type Controller struct {
...
-	// TODO: many of the types in PilotGatewayAPI (watched above) are duplicated
-	// I'm not sure why, but we shouldn't watch them twice.
-	duplicates := []collection.Schema{}
-	for k := range analysis.ContainmentMapSchema(rwConfigStore.Schemas()) {
-		duplicates = append(duplicates, k)
-	}
+	// Filter out configs watched by rwConfigStore
 	store, err := arbitraryclient.NewForSchemas(ctx, kubeClient, "default",
-		domainSuffix, collections.All.Remove(duplicates...))
+		domainSuffix, collections.All.Remove(rwConfigStore.Schemas().All()...))
```
therealmitchconnors, line 54: +1 so much cleaner!

pkg/config/analysis/local/istiod_analyze.go
```diff
@@ -41,27 +41,26 @@ import (
...
-	stores               []model.ConfigStoreCache
-	analyzer             *analysis.CombinedAnalyzer
-	transformerProviders transformer.Providers
-	namespace            resource.Namespace
-	istioNamespace       resource.Namespace
+	// internalStore stores synthetic configs for analysis (mesh config, etc)
+	internalStore model.ConfigStore
+	// stores contains all the (non file) config sources to analyze
+	stores []model.ConfigStoreCache
+	// fileSource contains all file bases srouces
```
therealmitchconnors, line 58: nit: typo

pkg/config/schema/collection/names_test.go
```diff
@@ -19,15 +19,14 @@ import (
...
-	data "istio.io/istio/pkg/config/legacy/testing/data"
 	"istio.io/istio/pkg/config/schema/collection"
+	"istio.io/istio/pkg/config/schema/collections"
 )
 
 func TestNames_Clone(t *testing.T) {
 	g := NewWithT(t)
 
-	n := collection.Names{basicmeta.K8SCollection1.Name(), basicmeta.Collection2.Name()}
+	n := collection.Names{collections.All.All()[0].Name(), collections.All.All()[1].Name()}
```
therealmitchconnors, line 29: nit: stutters

therealmitchconnors (commented):
Thanks for the PR!  I didn't realize how much of a mess I had left behind.   Thank you for cleaning it up.  Comments are minor, except the conversation about skipped analyzers.  Happy to chat about that if it helps.

pkg/config/analysis/local/local.go
howardjohn: As discussed offline, I misunderstood this. Added this back. Thanks!

therealmitchconnors (approved):

nmittler (approved):
Generally LGTM. I'm just going to approve since Mitch already took a closer look.

stevenctl (approved):

PR #36236, opened 2021-11-24
fix incorrect use log and optimized log output

nicole-lihui:
fix incorrect use log ：pilot/pkg/xds/discovery.go
optimized log output：other


pilot/pkg/xds/sds.go
```diff
@@ -154,7 +154,7 @@ func (s *SecretGen) Generate(proxy *model.Proxy, push *model.PushContext, w *mod
 			secret, err := secretController.GetCaCert(sr.Name, sr.Namespace)
 			if err != nil {
 				pilotSDSCertificateErrors.Increment()
-				log.Warnf("failed to fetch ca certificate for %v: %v", sr.ResourceName, err)
+				log.Warnf("failed to fetch ca certificate for %s: %v", sr.ResourceName, err)
```
howardjohn, line 157: I don't mind this but don't think its important to refactor around, as it will just add more noise with little benefit?

howardjohn (approved):

PR #36189, opened 2021-11-22
fix-gateway-svc-annotations

steevi:
**Please provide a description of this PR:**
The annotations added to the service of the gateway, are not the ones from the service yaml block but the global ones. This PR addresses that issue.

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [X] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [X] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #36247, opened 2021-11-24
Fix #36246

### LINKED ISSUE #36246: Official Istio gateway chart does not implement service annotations as described
### Bug Description

Hi there

In the Istio `gateway` chart `values.yaml` and chart documentation, it says that I can set the gateway annotations by setting `.Values.service.annotations`, however inside the chart the gateway service uses an undeclared `.Values.annotations` instead (not under `service` subkey).

`values.yaml`:
https://github.com/istio/istio/blob/1.12.0/manifests/charts/gateway/values.yaml#L51

`service.yaml`:
https://github.com/istio/istio/blob/master/manifests/charts/gateway/templates/service.yaml#L12

This had the impact of not deploying the ingress gateway service with the correct annotations for my environment.

### Version

```prose
client version: 1.11.4
control plane version: 1.12.0
data plane version: 1.12.0 (16 proxies)
```

### Additional Information

_No response_

lukeplausin:

Fix #36246

This change alters a slight detail in the `gateway` chart so that it uses the correct annotations source (in line with what's in the `values.yaml` file).

### DIFF
--- a/manifests/charts/gateway/templates/service.yaml
+++ b/manifests/charts/gateway/templates/service.yaml
@@ -10,7 +10,7 @@ metadata:
     topology.istio.io/network: "{{.}}"
     {{- end }}
   annotations:
-    {{- .Values.annotations | toYaml | nindent 4 }}
+    {{- .Values.service.annotations | toYaml | nindent 4 }}
 spec:
 {{- with .Values.service.loadBalancerIP }}
   loadBalancerIP: "{{ . }}"


manifests/charts/gateway/templates/service.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
     topology.istio.io/network: "{{.}}"
     {{- end }}
   annotations:
-    {{- .Values.annotations | toYaml | nindent 4 }}
+    {{- .Values.service.annotations | toYaml | nindent 4 }}
```
Monkeyanator, line 13: Do we want both `.Values.service.annotations` and `.Values.annotations` here @howardjohn? Same goes for other resources in this chart. Description of the annotations field:

```yaml
# Annotations to apply to all resources
annotations: {}
```

morvencao: I think we need both, I see another PR for this: https://github.com/istio/istio/pull/36189

morvencao:
I'll close this as another [PR](https://github.com/istio/istio/pull/36384 ) was merged. Thanks!

PR #36251, opened 2021-11-25
Validation: do not warn if latter VS prefix match rule covers former one

### LINKED ISSUE #36245: False IST0131 warning when using match / as catch all after 1.12.0 upgrade
### Bug Description

Getting this warning when applying the below manifest.  However accessing /blocked_stuff does result in the 400 error so it is working.

Warning: virtualService rule "20_blocked_urls" match #0 of prefix /blocked_stuff is not used (duplicate/overlapping match in rule #0 of prefix / on "90_main")

Istioctl analyze similarly reports 
Info [IST0131] (VirtualService default/productpage) VirtualService rule "20_blocked_urls" match #0 of prefix /blocked_stuff is not used (duplicate/overlapping match in rule #0 of prefix / on "90_main").

This didn't happen with 1.10.5, skipped 1.11.x and getting these warnings in 1.12.0
We 
Sample Manifest
`---
kind: VirtualService
metadata:
  name: productpage
spec:
  hosts:
  - productpage
  http:
  - fault:
      abort:
        httpStatus: 400
        percentage:
          value: 100
    match:
    - uri:
        prefix: /blocked_stuff
    name: 20_blocked_urls
    route:
    - destination:
        host: productpage
        subset: v1
  - match:
    - uri:
        prefix: /
    name: 90_main
    route:
    - destination:
        host: productpage
        subset: v1`

It seems the warning goes away if instead of having a final match on '/' we just have a default route.  This is probably the correct definition looking at the current docs anyways.

Just wondering if the definition as we have it is actually valid? or could break at any time, or is this more a bug with the analyzer?

### Version

```prose
istioctl
client version: 1.12.0
control plane version: 1.12.0
data plane version: 1.12.0 (6 proxies)

kubectl
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:41:02Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.13", GitCommit:"61be87ea533002fdf9251891b1a6e616c26a5cfd", GitTreeState:"clean", BuildDate:"2021-10-09T05:09:23Z", GoVersion:"go1.15.14", Compiler:"gc", Platform:"linux/amd64"}
```

### Additional Information

_No response_

### Affected product area

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Security
- [ ] Test and Release
- [X] User Experience
- [X] Upgrade
- [ ] Multi Cluster
- [ ] Virtual Machine
- [ ] Control Plane Revisions

bianpengyuan:
fix #36245

### DIFF
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -2273,20 +2273,11 @@ func analyzeUnreachableHTTPRules(routes []*networking.HTTPRoute,
 				if strings.Compare(matchHTTPRoutes[rIndex].Prefix, routePrefix.Prefix) == 0 {
 					continue
 				}
-				// Valid from A to B for matchHTTPRoutes
-				isAtoBCover := coveredValidation(routePrefix, matchHTTPRoutes[rIndex])
-				if isAtoBCover {
+				// Validate former prefix match does not cover the latter one.
+				if coveredValidation(routePrefix, matchHTTPRoutes[rIndex]) {
 					prefixMatchA := matchHTTPRoutes[rIndex].MatchStr + " of prefix " + matchHTTPRoutes[rIndex].Prefix
 					prefixMatchB := routePrefix.MatchStr + " of prefix " + routePrefix.Prefix + " on " + routePrefix.RouteStr
 					reportIneffective(matchHTTPRoutes[rIndex].RouteStr, prefixMatchA, prefixMatchB)
-				}
-
-				// Valid from B to A for matchHTTPRoutes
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
unit test would be ideal here

PR #36256, opened 2021-11-25
Add explicit authorization interface

hzxuzhonghu:
Provide a common interface, to support extending authorize in future
And also prevent the identities from being persisted, so saves a little memory

### DIFF
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -15,7 +15,6 @@
 package xds
 
 import (
-	"fmt"
 	"strconv"
 	"strings"
 	"sync/atomic"
@@ -37,7 +36,6 @@ import (
 	v3 "istio.io/istio/pilot/pkg/xds/v3"
 	"istio.io/istio/pkg/cluster"
 	"istio.io/istio/pkg/config/schema/gvk"
-	"istio.io/istio/pkg/spiffe"
 	"istio.io/pkg/env"
 	istiolog "istio.io/pkg/log"
 )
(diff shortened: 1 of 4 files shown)


pilot/pkg/xds/ads.go
```diff
@@ -321,6 +315,10 @@ func (s *DiscoveryServer) Stream(stream DiscoveryStream) error {
 	// reqChannel and the connection not being enqueued for pushes to pushChannel until the
 	// initialization is complete.
 	<-con.initialized
+	// authorize client
```
ramaraochavali, line 318: Should n't this be called before we start go routine for receive? May be move authenticate and authorize to initConnection?

hzxuzhonghu: no, the receive go routine initialize proxy, authorize depend on it.

ramaraochavali: Ah. Got it

ramaraochavali (approved):

pilot/pkg/xds/ads.go
```diff
@@ -321,6 +315,10 @@ func (s *DiscoveryServer) Stream(stream DiscoveryStream) error {
...
+	if err = s.authorize(con, ids); err != nil {
```
howardjohn, line 319: This must happen before RegisterWorkload I feel like? Now its happening afterwards

howardjohn (commented):
Abstracting the function SGTM. The new location is concerning.

pilot/pkg/xds/ads.go
hzxuzhonghu: will check

PR #36259, opened 2021-11-25
add fault filter to top of filters

ramaraochavali:
Fault Injection should be added as the first filter https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/fault_filter.html?highlight=fault%20injection#configuration

This PR fixes that ordering

- [ ] Docs
- [ ] Installation
- [ X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/listener.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener.go
@@ -1322,9 +1322,13 @@ func buildHTTPConnectionManager(listenerOpts buildListenerOpts, httpOpts *httpLi
 		filters = append(filters, xdsfilters.Alpn)
 	}
 
-	filters = append(filters, xdsfilters.Cors, xdsfilters.Fault)
+	filters = append(filters, xdsfilters.Cors)
 	filters = append(filters, listenerOpts.push.Telemetry.HTTPFilters(listenerOpts.proxy, listenerOpts.class)...)
 	filters = append(filters, xdsfilters.BuildRouterFilter(routerFilterCtx))
+
+	// Add fault filter at the top.
+	filters = append(filters[:1], filters[0:]...)
+	filters[0] = xdsfilters.Fault
 
 	connectionManager.HttpFilters = filters
 
--- a/pilot/pkg/networking/core/v1alpha3/listener.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener.go
@@ -1302,8 +1302,10 @@ func buildHTTPConnectionManager(listenerOpts buildListenerOpts, httpOpts *httpLi
 
 	routerFilterCtx := configureTracing(listenerOpts, connectionManager)
 
-	filters := make([]*hcm.HttpFilter, len(httpFilters))
-	copy(filters, httpFilters)
+	// Add fault filter at the top.
+	filters := make([]*hcm.HttpFilter, len(httpFilters)+1)
+	filters[0] = xdsfilters.Fault
+	copy(filters[1:], httpFilters)
 
 	if features.MetadataExchange {
 		filters = append(filters, xdsfilters.HTTPMx)
@@ -1325,10 +1327,6 @@ func buildHTTPConnectionManager(listenerOpts buildListenerOpts, httpOpts *httpLi
 	filters = append(filters, xdsfilters.Cors)
 	filters = append(filters, listenerOpts.push.Telemetry.HTTPFilters(listenerOpts.proxy, listenerOpts.class)...)
 	filters = append(filters, xdsfilters.BuildRouterFilter(routerFilterCtx))
-
-	// Add fault filter at the top.
-	filters = append(filters[:1], filters[0:]...)
-	filters[0] = xdsfilters.Fault
 
 	connectionManager.HttpFilters = filters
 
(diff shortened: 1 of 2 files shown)


pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -1322,10 +1322,14 @@ func buildHTTPConnectionManager(listenerOpts buildListenerOpts, httpOpts *httpLi
 		filters = append(filters, xdsfilters.Alpn)
 	}
 
-	filters = append(filters, xdsfilters.Cors, xdsfilters.Fault)
+	filters = append(filters, xdsfilters.Cors)
 	filters = append(filters, listenerOpts.push.Telemetry.HTTPFilters(listenerOpts.proxy, listenerOpts.class)...)
 	filters = append(filters, xdsfilters.BuildRouterFilter(routerFilterCtx))
 
+	// Add fault filter at the top.
+	filters = append(filters[:1], filters[0:]...)
```
hzxuzhonghu, line 1330: Instead of this writing, i prefer 

```
filters := make([]*hcm.HttpFilter, len(httpFilters) + 1)
filters[0] = fault filter
copy(filters[1:], httpFilters)
```

hzxuzhonghu: wouldn't this output duplicate filters？
append(filters[:1], filters[0:]...)

ramaraochavali: Changed

hzxuzhonghu (approved):

howardjohn (commented):
What is the behavioral difference?

ramaraochavali:
There is no behavioural difference OOB. But if faults are configured by using Envoy filter (merging to the existing filter) faults will be executed first before any authn/authz happens which is not needed when faults are added.

howardjohn:
What about without EnvoyFilter, will fault now apply before authz and
previously it applied after? If so that seems like both a breaking change
and one I am not sure is intended. I would expect authz to apply quite
early, if we show the fault response we may leak info that was intended to
be protected by authZ

ramaraochavali:
No. Right now it is a noop because there are no faults configured

ramaraochavali:
Ok. One case it might change the order if faults are controlled by envoy
specific header. But what information fault will that is protected by authz
and see envoys recommendation that I linked in the pr as well

howardjohn:
Even the fact an endpoint exists can be something that you want protected by authz. Status code can additionally leak information beyond just the binary exists/not exists. 

What is the motivation for this? Is it just because Envoy docs say to do it or do we have a specific use case?

ramaraochavali:
> What is the motivation for this? Is it just because Envoy docs say to do it or do we have a specific use case?

The usecase is we want do fault injection before validating JWT and with the current model RequestAuthentication adds JWT before fault injection. And When I referred Envoy docs found that line.

howardjohn:
It seems just as likely that you want it before as that you want it after
though... I don't think the Envoy doc matters much, it's user specific
where they want faults to apply. In this case, we are breaking
compatibility to accommodate your use case (which is a very reasonable use
case), but at the expense of others use cases (which are also reasonable).

Maybe there is a way to modify the order with envoy filter or something
that doesn't change the behavior for existing users?

ramaraochavali:
> Maybe there is a way to modify the order with envoy filter or something
that doesn't change the behavior for existing users?

If we strictly want the existing order - the other option is document this and folks can remove the existing Envoy and add a new filter at the top using Envoy filter.
Or add a flag to control this?

Let me know - I change like that.

howardjohn:
Flag seems tricky since its often not really a global decision. Envoyfilter
may be the best option?

ramaraochavali:
Another thought - What is the motivation/history for adding fault filter by default? I am wondering if we can remove it completely and ask folks to add their own fault filter at listener (via envoy filter) if they need because any way do not support any default config and if they need route specific they can use VS feature or VS feature needs this?

@hzxuzhonghu might remember the history

hzxuzhonghu:
It is too old, but seems we are relying on route fault injection.

ramaraochavali:
I think Envoy does not need to be added to top level filters

https://github.com/envoyproxy/envoy/blob/91c2e6ba81f6892e4ab132f934d9f6179a50bbc2/source/common/router/config_impl.cc#L1708 here is where it builds the per filter config for route and it will try to get by name and if not then only it will throw exception
https://github.com/envoyproxy/envoy/blob/0fae6970ddaf93f024908ba304bbd2b34e997a51/source/common/config/utility.h#L256  

So for fault filter we do not have to really add it to the top level filters to make this work for per route config.

hzxuzhonghu:
code shows everything

ramaraochavali:
Validated the behaviour with this sample https://github.com/envoyproxy/envoy/tree/a20c7fcffd02cf046bce2c04e196681108eebf93/examples/csrf - it works correctly even if we remove the top level filter

ramaraochavali:
@lizan Curious what is rationale behind this https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/fault_filter.html?highlight=fault%20injection#configuration recommendation? Is it just to ensure there is in general no need to execute other filters when we are doing chaos testing or any thing else?

PR #36260, opened 2021-11-25
Fix omitted enableNamespacesByDefault in the istio-revision-tag-default injector mutating webhook

### LINKED ISSUE #36258: Omitted setting sidecarInjectorWebhook.enableNamespacesByDefault in istio-revision-tag-default mutating webhook.
### Bug Description

In the Istio 1.12.0 installed by istioctl with default profile, there are two mutating webhooks. 
[mutatingwebhook.yaml](https://github.com/istio/istio/blob/1.12.0/manifests/charts/default/templates/mutatingwebhook.yaml) is missing the part of injector's logic when neither label is set:
```
{{- if .Values.sidecarInjectorWebhook.enableNamespacesByDefault }}
{{- /* Special case 3: no labels at all */}}
{{- include "core" (mergeOverwrite (deepCopy .) (dict "Prefix" "auto.") ) }}
  namespaceSelector:
    matchExpressions:
    - key: istio-injection
      operator: DoesNotExist
    - key: istio.io/rev
      operator: DoesNotExist
  objectSelector:
    matchExpressions:
    - key: sidecar.istio.io/inject
      operator: DoesNotExist
    - key: istio.io/rev
      operator: DoesNotExist
{{- end }}
```
Neither the latest [docs](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy) nor the upgrade notes did not state changes to this logic.

### Version

```prose
istioctl version
client version: 1.12.0
control plane version: 1.12.0
data plane version: 1.12.0 (3 proxies)
```

### Additional Information

Installed the Istio default profile with`istioctl install --set values.sidecarInjectorWebhook.enableNamespacesByDefault=true` using istioctl-1.12.0.
Then deployed workload without pod and namespace annotations.

cnvergence:
Fixes issue #36258, where special case is omitted in the mutating webhook template.

### DIFF
--- a/manifests/charts/default/values.yaml
+++ b/manifests/charts/default/values.yaml
@@ -14,3 +14,8 @@ istiodRemote:
 
 # Revision is set as 'version' label and part of the resource names when installing multiple control planes.
 revision: ""
+
+sidecarInjectorWebhook:
+  # This enables injection of sidecar in all namespaces,
+  enableNamespacesByDefault: false
+
--- a/releasenotes/notes/36258.yaml
+++ b/releasenotes/notes/36258.yaml
@@ -1,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: installation
+issue:
+  - https://github.com/istio/istio/issues/36258
+releaseNotes:
+  - |
+    **Fixed** omitted special case in the revision default mutating webhook.
--- a/releasenotes/notes/36258.yaml
+++ b/releasenotes/notes/36258.yaml
@@ -5,5 +5,5 @@ issue:
   - https://github.com/istio/istio/issues/36258
 releaseNotes:
   - |
-    **Fixed** omitted setting .Values.sidecarInjectiorWebhook.enableNamespacesByDefault in the revision
-    mutating webhook and added --auto-namespace-injection flag to `istioctl tag` controlling this setting.
+    **Fixed** omitted setting .Values.sidecarInjectiorWebhook.enableNamespacesByDefault in the default revision
+    mutating webhook and added --auto-inject-namespaces flag to `istioctl tag` controlling this setting.
--- a/releasenotes/notes/36258.yaml
+++ b/releasenotes/notes/36258.yaml
@@ -5,5 +5,5 @@ issue:
   - https://github.com/istio/istio/issues/36258
 releaseNotes:
   - |
-    **Fixed** omitted setting .Values.sidecarInjectiorWebhook.enableNamespacesByDefault in the revision default mutating webhook
-    and added --auto-namespace-injection flag to `istioctl tag` controlling this setting.
+    **Fixed** omitted setting .Values.sidecarInjectiorWebhook.enableNamespacesByDefault in the revision
+    mutating webhook and added --auto-namespace-injection flag to `istioctl tag` controlling this setting.
--- a/releasenotes/notes/36258.yaml
+++ b/releasenotes/notes/36258.yaml
@@ -5,4 +5,5 @@ issue:
   - https://github.com/istio/istio/issues/36258
 releaseNotes:
   - |
-    **Fixed** omitted special case in the revision default mutating webhook.
+    **Fixed** omitted setting .Values.sidecarInjectiorWebhook.enableNamespacesByDefault in the revision default mutating webhook
+    and added --auto-namespace-injection flag to `istioctl tag` controlling this setting.
--- a/istioctl/cmd/tag.go
+++ b/istioctl/cmd/tag.go
@@ -47,17 +47,19 @@ overwrite existing revision tags.`
 	tagCreatedStr   = `Revision tag %q created, referencing control plane revision %q. To enable injection using this
 revision tag, use 'kubectl label namespace <NAMESPACE> istio.io/rev=%s'
 `
-	webhookNameHelpStr = "Name to use for a revision tag's mutating webhook configuration."
+	webhookNameHelpStr          = "Name to use for a revision tag's mutating webhook configuration."
+	autoInjectNamespacesHelpStr = "If set to true, the sidecars should be automatically injected into all namespaces by default"
 )
 
 // options for CLI
 var (
 	// revision to point tag webhook at
-	revision         = ""
-	manifestsPath    = ""
-	overwrite        = false
-	skipConfirmation = false
-	webhookName      = ""
+	revision             = ""
+	manifestsPath        = ""
+	overwrite            = false
+	skipConfirmation     = false
+	webhookName          = ""
+	autoInjectNamespaces = false
 )
 
 type tagDescription struct {
@@ -148,6 +150,7 @@ injection labels.`,
 	cmd.PersistentFlags().BoolVarP(&skipConfirmation, "skip-confirmation", "y", false, skipConfirmationFlagHelpStr)
 	cmd.PersistentFlags().StringVarP(&revision, "revision", "r", "", revisionHelpStr)
 	cmd.PersistentFlags().StringVarP(&webhookName, "webhook-name", "", "", webhookNameHelpStr)
+	cmd.PersistentFlags().BoolVar(&autoInjectNamespaces, "auto-namespace-injection", false, autoInjectNamespacesHelpStr)
 	_ = cmd.MarkPersistentFlagRequired("revision")
 
 	return cmd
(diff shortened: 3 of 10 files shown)


manifests/charts/default/values.yaml
```diff
@@ -14,3 +14,8 @@ istiodRemote:
 
 # Revision is set as 'version' label and part of the resource names when installing multiple control planes.
 revision: ""
+
+sidecarInjectorWebhook:
```
Monkeyanator, line 18: The default chart is rendered through code with [hardcoded values](https://github.com/istio/istio/blob/master/istioctl/pkg/tag/generate.go#L208), so this won't get set (even if setting `sidecarInjectorWebhook.enableNamespacesByDefault=true` on the pilot chart). I think we need to add another [generate option](https://github.com/istio/istio/blob/master/istioctl/pkg/tag/generate.go#L56) with a new flag to `istioctl tag` for this.

Monkeyanator (changes requested):

manifests/charts/default/values.yaml
cnvergence: Thanks for explaining, I have added the generate option, PTAL

releasenotes/notes/36258.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: installation
+issue:
+  - https://github.com/istio/istio/issues/36258
+releaseNotes:
+  - |
+    **Fixed** omitted special case in the revision default mutating webhook.
```
linsun, line 8: recommend adding this config `enableNamespacesByDefault` in the release note.

Monkeyanator (commented):
Nice! One last thing, I think we want to make it so that when the `enableNamespacesByDefault` is set on the IOP we set this option in https://github.com/istio/istio/blob/master/operator/cmd/mesh/install.go#L197.

linsun:
hey @Monkeyanator  this function is a bit more complicated than I had thought. Is this the expected behavior?  

case 1) Fresh install of a Istio with revision (or no revision), user can enable auto injection namespace enabled.  default is disabled.
case 2) User has Istio 1.11 installed with auto injection namespace disabled. User installs Istio 1.12(canary) revision with auto injection namespace enabled.  This injection configuration is only applicable to pods or namespaces that are labelled with matching canary revision.  Or when canary is promoted to be the default, then it will change all namespaces behavior to auto inject with 1.12's sidecar? What is the behavior if a namespace is tagged specifically to the old 1.11 revision?  Do we still inject 1.11's sidecar?

Or should we recommend users not change this configuration during upgrade to avoid complication?   They can change it after the upgrade of cp and dp is complete.

linsun:
@cnvergence  thank you for your contribution, your PR LGTM.  I'll defer to @Monkeyanator for the final approval.

Monkeyanator:
> case 1) Fresh install of a Istio with revision (or no revision), user can enable auto injection namespace enabled. default is disabled.

Right, for this we'll respect the option after this change.

> case 2) User has Istio 1.11 installed with auto injection namespace disabled. User installs Istio 1.12(canary) revision with auto injection namespace enabled. This injection configuration is only applicable to pods or namespaces that are labelled with matching canary revision. Or when canary is promoted to be the default, then it will change all namespaces behavior to auto inject with 1.12's sidecar?

Since there's an existing revision (1.11) we won't automatically install the default tag in this case. The user will have to take the action to do `istioctl tag set default --revision 1-12-revision --auto-inject-namespaces` if they want to get auto injection for 1.12.

> What is the behavior if a namespace is tagged specifically to the old 1.11 revision? Do we still inject 1.11's sidecar?

The selector for the auto-injection webhook only matches if there aren't any of the other `istio.io/rev`, `sidecar.istio.io/inject`, `istio-injection` labels. So 1.11 injector will still inject the old revisions.

Monkeyanator (approved):

linsun:
@Monkeyanator  thank you for the clarification!  I think the behavior you described is reasonable and expected.

linsun (approved):
LGTM

@therealmitchconnors @esnible @irisdingbj could we get an UX approval from one of you?

jacob-delgado (approved):

howardjohn (commented):
Say my current state is:

```
revision prod, default injector
revision canary
```

I want to change canary to default injector. How can I do it?

If I set canary=default then prod=non-default, I have a period of 2x default injectors.

If I do the opposite, I have a period where pods will not be injected

therealmitchconnors (commented):
I'm struggling to tell if this code is fixing a regression, where a prior version of `tag` had this behavior and we are restoring it, which is what the linked issue seems to say, or if we are adding an entirely new feature.  @cnvergence can you explain?

howardjohn:
> Say my current state is:
> 
> ```
> revision prod, default injector
> revision canary
> ```
> 
> I want to change canary to default injector. How can I do it?
> 
> If I set canary=default then prod=non-default, I have a period of 2x default injectors.
> 
> If I do the opposite, I have a period where pods will not be injected

Spoke to Sam, I think we want to allow this only on the `default` tag, which is a singleton

Monkeyanator:
> I'm struggling to tell if this code is fixing a regression, where a prior version of tag had this behavior and we are restoring it, which is what the linked issue seems to say, or if we are adding an entirely new feature.

@therealmitchconnors it both introduces a new feature and fixes a regression; we need the new tag feature to fix the regression. The regression is that the `enableNamespacesByDefault` option no longer does anything since we generate the default tag separately rather than rendering from the pilot charts. The feature is that the default tag can be made to inject for workloads without labels on them which it previously couldn't.

Monkeyanator (changes requested):
@cnvergence as John pointed out, I think we want logic to only allow this option when creating/modifying the default tag.

Monkeyanator:
And one more thing: I think we may want to remove the `enableNamespacesByDefault` section from https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/templates/mutatingwebhook.yaml#L126? @therealmitchconnors pointed out that if we don't we'll get the no-labels injector for both the per-revision webhook and the default tag.

cnvergence:
@therealmitchconnors I have started with the fix related to the issue and with the review of @Monkeyanator we have also introduced a new option in the `istioctl tag`

cnvergence:
@howardjohn @Monkeyanator This option is only allowed for the default tag in both: --auto-inject-namespaces and istioctl install --set values.sidecarInjectorWebhook.enableNamespacesByDefault=true.
https://github.com/istio/istio/blob/50b858b5eca9e798d82042149a9f136ada4a8501/manifests/charts/istio-control/istio-discovery/templates/revision-tags.yaml#L78
`istioctl tag set canary --revision bar `won't generate no labels selector

cnvergence:
> And one more thing: I think we may want to remove the `enableNamespacesByDefault` section from https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/templates/mutatingwebhook.yaml#L126? @therealmitchconnors pointed out that if we don't we'll get the no-labels injector for both the per-revision webhook and the default tag.

I believe they are always disabled for the default tag
https://github.com/istio/istio/blob/83563f2b5070349d7a052ce1b5bb572038fd1d74/istioctl/pkg/tag/generate.go#L119

Monkeyanator:
> I believe they are always disabled for the default tag

`DeactivateIstioInjectionWebhook` will deactivate the webhook for non-revisioned installations but not for revisioned webhooks. So IIUC,`istioctl install --revision orange --set enableNamespacesByDefault=true` on a fresh cluster will result in two "no label" webhooks: (1) the istio-revision-tag-default webhook will have it (because of this PR) and (2) the istio-sidecar-injector-orange webhook will have it (because it gets templated in and we never deactivate it).

I think we can remove the section in the per-revision webhook template (https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/templates/mutatingwebhook.yaml#L126) that has the "no label" selector and enfroce that only the default tag can contain this "no label" selector.

Monkeyanator (approved):
Was wrong about the previous concerns, LGTM!

cnvergence:
@Monkeyanator thanks for the review!
Hopefully, we can include this soon, as it is holding us from the upgrade.

therealmitchconnors (approved):

PR #36263, opened 2021-11-25
distinguis logs between ads and dads

nicole-lihui:


hzxuzhonghu (changes requested):
I donot think we need this, as we could only enable one mode

nicole-lihui:
> I donot think we need this, as we could only enable one mode

If this is just a transitional stage of a new model, it is really not necessary, thank you for your suggestion

PR #36269, opened 2021-11-26
MInor cleanup and fix for delta xds

hzxuzhonghu:

1. merge responseNonce=nil with `previousInfo =nil` processing
2. fix con.proxy lock for shouldRespondDelta
3. cleanup some comments

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -373,25 +373,16 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 		return false
 	}
 
-	// This is first request - initialize typeUrl watches.
-	if request.ResponseNonce == "" {
-		log.Debugf("ADS:%s: INIT %s %s %s", stype, con.ConID, request.VersionInfo, request.ResponseNonce)
-		con.proxy.Lock()
-		con.proxy.WatchedResources[request.TypeUrl] = &model.WatchedResource{TypeUrl: request.TypeUrl, ResourceNames: request.ResourceNames, LastRequest: request}
-		con.proxy.Unlock()
-		return true
-	}
-
 	con.proxy.RLock()
 	previousInfo := con.proxy.WatchedResources[request.TypeUrl]
 	con.proxy.RUnlock()
 
-	// This is a case of Envoy reconnecting Istiod i.e. Istiod does not have
+	// This is a case of Envoy init/reconnecting Istiod i.e. Istiod does not have
 	// information about this typeUrl, but Envoy sends response nonce - either
 	// because Istiod is restarted or Envoy disconnects and reconnects.
 	// We should always respond with the current resource names.
 	if previousInfo == nil {
-		log.Debugf("ADS:%s: RECONNECT %s %s %s", stype, con.ConID, request.VersionInfo, request.ResponseNonce)
+		log.Debugf("ADS:%s: INIT/RECONNECT %s %s %s", stype, con.ConID, request.VersionInfo, request.ResponseNonce)
 		con.proxy.Lock()
 		con.proxy.WatchedResources[request.TypeUrl] = &model.WatchedResource{TypeUrl: request.TypeUrl, ResourceNames: request.ResourceNames, LastRequest: request}
 		con.proxy.Unlock()
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -378,7 +378,7 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 	con.proxy.RUnlock()
 
 	// This can happen in two cases:
-	// 1. Envoy initially send request to istiod
+	// 1. Envoy initially send request to Istiod
 	// 2. Envoy reconnect to Istiod i.e. Istiod does not have
 	// information about this typeUrl, but Envoy sends response nonce - either
 	// because Istiod is restarted or Envoy disconnects and reconnects.
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -377,11 +377,13 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 	previousInfo := con.proxy.WatchedResources[request.TypeUrl]
 	con.proxy.RUnlock()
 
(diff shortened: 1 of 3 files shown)


pilot/pkg/xds/ads.go
```diff
@@ -374,25 +373,16 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
...
 	con.proxy.RUnlock()
 
-	// This is a case of Envoy reconnecting Istiod i.e. Istiod does not have
+	// This is a case of Envoy init/reconnecting Istiod i.e. Istiod does not have
 	// information about this typeUrl, but Envoy sends response nonce - either
 	// because Istiod is restarted or Envoy disconnects and reconnects.
 	// We should always respond with the current resource names.
 	if previousInfo == nil {
-		log.Debugf("ADS:%s: RECONNECT %s %s %s", stype, con.ConID, request.VersionInfo, request.ResponseNonce)
+		log.Debugf("ADS:%s: INIT/RECONNECT %s %s %s", stype, con.ConID, request.VersionInfo, request.ResponseNonce)
```
ramaraochavali, line 385: This is not same as above. This is the case where Istiod does not have information (prevInfo) but request has ResponseNonce.

May be check request.responseNonce here and change debug accordingly to be very clear.

Also update the above comment for both cases if we are merging

hzxuzhonghu: Sorry i donot understand how can this happen  "This is the case where Istiod does not have information (prevInfo) but request has ResponseNonce."

hzxuzhonghu: Do you mean i should update the comments above?

ramaraochavali: Envoy sends a request when it is reconnecting which has responseNonce but Istiod since that proxy is not connected to it - it does not have the prevInfo.

hzxuzhonghu: understood, will update the comments

pilot/pkg/xds/ads.go
```diff
@@ -377,11 +377,13 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 	previousInfo := con.proxy.WatchedResources[request.TypeUrl]
 	con.proxy.RUnlock()
 
-	// This is a case of Envoy init/reconnecting Istiod i.e. Istiod does not have
+	// This can happen in two cases:
+	// 1. Envoy initially send request to istiod
```
ramaraochavali, line 381: nit: istiod -> Istiod to be consistent

hzxuzhonghu: ok, will fixlater

ramaraochavali (approved):

pilot/pkg/xds/ads.go
```diff
@@ -374,25 +373,18 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
...
-	// This is a case of Envoy reconnecting Istiod i.e. Istiod does not have
+	// This can happen in two cases:
+	// 1. Envoy initially send request to Istiod
+	// 2. Envoy reconnect to Istiod i.e. Istiod does not have
 	// information about this typeUrl, but Envoy sends response nonce - either
 	// because Istiod is restarted or Envoy disconnects and reconnects.
 	// We should always respond with the current resource names.
-	if previousInfo == nil {
-		log.Debugf("ADS:%s: RECONNECT %s %s %s", stype, con.ConID, request.VersionInfo, request.ResponseNonce)
+	if request.ResponseNonce == "" || previousInfo == nil {
```
howardjohn, line 386: can we please add some tests when we are changing core functionality?

pilot/pkg/xds/ads.go
howardjohn: What is the motivation for this change? It seems like we have worse debug ability now...

pilot/pkg/xds/delta.go
```diff
@@ -383,27 +381,18 @@ func (s *DiscoveryServer) shouldRespondDelta(con *Connection, request *discovery
...
-		if features.EnableUnsafeAssertions {
-			panic(fmt.Sprintf("dADS:%s: New ACK and old ACK check mismatch: %v vs %v", stype, oldAck, newAck))
-		}
-	}
+
 	// Envoy can send two DiscoveryRequests with same version and nonce
 	// when it detects a new resource. We should respond if they change.
-	if oldAck {
-		log.Debugf("dADS:%s: ACK %s %s", stype, con.ConID, request.ResponseNonce)
+	if listEqualUnordered(previousResources, con.proxy.WatchedResources[request.TypeUrl].ResourceNames) {
```
howardjohn, line 394: Why is this changed? The Delta XDS protocol explicitly says to use `newAck` as the determination. This is why we had the check to assert they are the same previously - in some tests we have found they are NOT which indicates we have a misunderstanding somewhere.

Even if they are always identical, we should use newAck as it is simpler and follows the spec and more efficient

pilot/pkg/xds/delta.go
```diff
@@ -451,9 +440,10 @@ func (s *DiscoveryServer) pushDeltaXds(con *Connection, push *model.PushContext,
 		// If subscribe is set, client is requesting specific resources. We should just give it the
 		// new resources it needs, rather than the entire set of known resources.
 		subres := sets.NewSet(subscribe...)
-		filteredResponse := []*discovery.Resource{}
+		filteredResponse := make([]*discovery.Resource, 0, len(subscribe))
 		for _, r := range res {
 			if subres.Contains(r.Name) {
+				log.Debugf("----- %s", r.Name)
```
howardjohn, line 446: ?

pilot/pkg/xds/delta.go
hzxuzhonghu:
```
Spontaneous DeltaDiscoveryRequests from the client. This can be done to dynamically add or remove elements from the tracked resource_names set. In this case response_nonce must be omitted.
```

If this is what you mean, then we don't need to compare the previous watched resources and current sub ones.


```
Note that while a response_nonce may be set on the request, the server must honor changes to the subscription state even if the nonce is stale. The nonce may be used to correlate an ack/nack with a server response, but should not be used to reject stale requests.
```

Issue #21228, opened 2020-02-18
Ability to gradually warm services


Stono:
**Describe the feature request**
We write a lot of applications that rely heavily on JIT/Hot spot code compilation.  As a result, they require a certain amount of warming before they're really good to serve load.

The typical kubernetes rollout brings in new instances, and then istio (and kubernetes in general) will almost immediately send a fair percentage of traffic to that instance.   This is an issue as it means a large amount of requests will hit a cold pod and the overall service will encounter a nice latency spike.

It would be brilliant if that percentage was ramped up over a short window, say 15 seconds, rather than firehosed, which means only a smaller amount of traffic is used to "warm" the pod minimising consumer impact during rollouts or traffic switching.

EG:
application has instances 1, 2, 3, 4 - each getting 25% load.
#5 comes online, instead of the weighting going to 20/20/20/20/20 as fast as possible, it'd be great if #5 was gradually ramped up to 20.

**Describe alternatives you've considered**
Complicated deployment strategies and using traffic mirroring.  However then i'm using "fake" traffic which would need to be excluded from key business metrics (hard).

Convincing my entire organisation to write in other languages (unlikely).

Readiness probes that send a few fake requests at the service before it becomes ready, but they're not really doing a decent job (giving enough variance), and also require maintenance.

**Additional context**
It'd be a nice added differentiator for Istio, to also improve deployments! 

Red line indicates deployment:
<img width="808" alt="Screenshot 2020-02-18 at 22 13 09" src="https://user-images.githubusercontent.com/1486729/74782700-e09cc380-529b-11ea-8582-7b080bb99395.png">

hzxuzhonghu:
Don't u have readiness probe? It can work in this case.

Stono:
Yes of course we do, it doesn’t help.  As I explained in the “alternatives we have considered”

BradErz:
We have the same problem @Stono, we are looking into deploying https://github.com/weaveworks/flagger so that we can do automated gradual rollouts of the new version starting with like 1,5,10% of traffic etc to warm up the code path before giving all the traffic.

(We haven't deployed it to prod yet but we are evaluating it)

BradErz:
Ah ok yes i misunderstood sorry.

zhongfox:
as described in https://github.com/istio/istio/issues/28271, some user already hack istio, use simple strategy to warm up: increase weight of endpoint by ready-time.

any change we add the new feature to DR?

```
kind: DestinationRule
spec:
[...]

PR #36274, opened 2021-11-26
send unready endpoints as unhealthy

ramaraochavali:
PR to send not ready endpoints as unhealthy.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali:
@hzxuzhonghu @howardjohn PTAL. If this change is fine, I will add few tests

howardjohn:
Didn't have a chance to look but quick comment - one concern is we
introduce 2 pushes on each new pod now. The initial not ready push doesn't
seem to matter quite as much . Ideally we would skip the initial unready ,
but that may be complex.

Btw - with warming if we send it as unhealthy will it start warming
already?

ramaraochavali:
> Btw - with warming if we send it as unhealthy will it start warming
already?

I do not think so. 

> one concern is we
introduce 2 pushes on each new pod now. The initial not ready push doesn't
seem to matter quite as much . Ideally we would skip the initial unready ,
but that may be complex.

I was wondering about this too. But do not know how to fix it. Does K8s previous state by any chance? I could n't find

ramaraochavali:
@howardjohn @hzxuzhonghu This is ready for review.

@howardjohn  handled this case as well. PTAL
```
one concern is we
introduce 2 pushes on each new pod now. The initial not ready push doesn't
seem to matter quite as much . Ideally we would skip the initial unready ,
but that may be complex.
```

ramaraochavali:
@howardjohn @hzxuzhonghu can you PTAL this?

pilot/pkg/xds/eds.go
```diff
@@ -124,27 +126,28 @@ func (s *DiscoveryServer) EDSCacheUpdate(shard model.ShardKey, serviceName strin
...
 		// unnecessary full push which can become a real problem if a pod is in crashloop and thus endpoints
 		// flip flopping between 1 and 0.
 		s.deleteEndpointShards(shard, hostname, namespace)
 		log.Infof("Incremental push, service %s has no endpoints", hostname)
-		return false
+		return false, true
 	}
 
-	fullPush := false
+	needPush := false
```
howardjohn, line 140: nit: set this on L 178 where it is used?

pilot/pkg/xds/eds.go
```diff
@@ -165,9 +168,37 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
...
+	}
+	if created {
+		return true, true
+	}
+	// Check if new Endpoints are ready to be pushed. This check
+	// will ensure that if a new pod comes with a non ready endpoint,
+	// we do not unnecessarily push that config to Envoy.
+	for _, nie := range istioEndpoints {
+		exists := false
+		for _, oie := range oldIstioEndpoints {
```
howardjohn, line 181: O(N^2) check may be problematic here? we can have huge numbers

pilot/pkg/xds/eds_test.go
```diff
@@ -338,6 +339,119 @@ func TestEDSOverlapping(t *testing.T) {
 	testOverlappingPorts(s, adscon, t)
 }
 
+func TestEDSUnhealthyEndpoints(t *testing.T) {
```
howardjohn, line 342: This test takes 5s to run, can we make it not wait for negative case?

howardjohn (commented):
Since this is a behavioral change I think we need upgrade note + feature flag

hzxuzhonghu:
Can you link the issue, we didnot send any unhealthy endpoints before.

ramaraochavali:
@hzxuzhonghu This for https://github.com/istio/istio/issues/21228 and https://github.com/istio/api/pull/2153

hzxuzhonghu:
Do they have some dependencies?

ramaraochavali:
Yes. See comments from below https://github.com/istio/api/pull/2153#issuecomment-969839933

pilot/pkg/serviceregistry/kube/controller/endpoints.go
```diff
@@ -209,6 +209,21 @@ func (e *endpointsController) buildIstioEndpoints(endpoint interface{}, host hos
...
+			pod, expectedPod := getPod(e.c, ea.IP, &metav1.ObjectMeta{Name: ep.Name, Namespace: ep.Namespace}, ea.TargetRef, host)
+			if pod == nil && expectedPod {
+				continue
+			}
+			builder := NewEndpointBuilder(e.c, pod)
+
+			// EDS and ServiceEntry use name for service port - ADS will need to map to numbers.
+			for _, port := range ss.Ports {
+				istioEndpoint := builder.buildIstioEndpoint(ea.IP, port.Port, port.Name, discoverabilityPolicy)
+				istioEndpoint.HealthStatus = model.UnHealthy
```
hzxuzhonghu, line 226: Looks this is duplicate with above , and can be abstracted. 

I am thinking maybe we should have a feature flag for it.

pilot/pkg/xds/eds.go
```diff
@@ -124,27 +126,28 @@ func (s *DiscoveryServer) EDSCacheUpdate(shard model.ShardKey, serviceName strin
 // edsCacheUpdate updates EndpointShards data by clusterID, hostname, IstioEndpoints.
 // It also tracks the changes to ServiceAccounts. It returns whether a full push
 // is needed or incremental push is sufficient.
-func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string, namespace string, istioEndpoints []*model.IstioEndpoint) bool {
+func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string, namespace string, istioEndpoints []*model.IstioEndpoint) (bool, bool) {
```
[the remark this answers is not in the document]
hzxuzhonghu: Yeah, also need to update the description

pilot/pkg/xds/eds.go
```diff
@@ -165,9 +168,37 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
...
 	if saUpdated {
 		log.Infof("Full push, service accounts changed, %v", hostname)
-		fullPush = true
+		return true, true
+	}
+	if created {
+		return true, true
+	}
+	// Check if new Endpoints are ready to be pushed. This check
+	// will ensure that if a new pod comes with a non ready endpoint,
```
hzxuzhonghu, line 177: how do you distinguish if it is from a new pod

ramaraochavali: Check the logic below. If it is not there in shards

pilot/pkg/xds/eds.go
ramaraochavali: I thought about it but could not find a better way. Let me see.

ramaraochavali: I found a better way - will fix.

pilot/pkg/xds/eds.go
```diff
@@ -165,9 +168,37 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
 	// For existing endpoints, we need to do full push if service accounts change.
 	if saUpdated {
 		log.Infof("Full push, service accounts changed, %v", hostname)
-		fullPush = true
+		return true, true
+	}
+	if created {
```
bianpengyuan, line 173: For newly created shards, are we going to push the endpoints regardless of whether it is healthy or not? Seems like the logic below try to avoid pushing new endpoints which are not healthy yet, but here we are pushing regardless of that.

ramaraochavali: Yes. That is correct. Because it is a full push, we want to send endpoints as well. The below is an optimization to reduce one additional endpoint push when a new pod comes up with with unready and changes to ready immediately.

ramaraochavali:
@hzxuzhonghu @howardjohn addressed all comments, added feature flag and release notes. PTAL.

pilot/pkg/serviceregistry/kube/controller/endpoints.go
```diff
@@ -199,21 +200,35 @@ func (e *endpointsController) buildIstioEndpoints(endpoint interface{}, host hos
...
 	for _, ss := range ep.Subsets {
-		for _, ea := range ss.Addresses {
-			pod, expectedPod := getPod(e.c, ea.IP, &metav1.ObjectMeta{Name: ep.Name, Namespace: ep.Namespace}, ea.TargetRef, host)
-			if pod == nil && expectedPod {
-				continue
-			}
-			builder := NewEndpointBuilder(e.c, pod)
+		endpoints = append(endpoints, e.buildIstioEndpointFromAddress(ep, ss, ss.Addresses, host, discoverabilityPolicy, true)...)
+		if features.SendUnhealthyEndpoints {
+			endpoints = append(endpoints, e.buildIstioEndpointFromAddress(ep, ss, ss.NotReadyAddresses, host, discoverabilityPolicy, true)...)
```
hzxuzhonghu, line 205: false

pilot/pkg/features/pilot.go
```diff
@@ -105,6 +105,12 @@ var (
 			" EDS pushes may be delayed, but there will be fewer pushes. By default this is enabled",
 	).Get()
 
+	SendUnhealthyEndpoints = env.RegisterBoolVar(
+		"PILOT_SEND_UNHEALTHY_ENDPOINTS",
+		true,
```
hzxuzhonghu, line 110: Suggest false by default, this may have performance influence

pilot/pkg/xds/eds.go
```diff
@@ -165,9 +167,35 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
...
+		} else {
+			// If the endpoint does not exist in shards that means it is a
+			// new endpoint. Only send if it is healthy to avoid pushing endpoints
+			// that are not ready to start with.
+			needPush = nie.HealthStatus == model.Healthy
+		}
+		if needPush {
+			break
+		}
+	}
```
hzxuzhonghu, line 196: Concerned with the performance, also should check the env flag

pilot/pkg/features/pilot.go
ramaraochavali: The reason I set default to true is this is needed for features like warmup etc which we are going to build. When that is used we can not ask customers to set this. If they are not using those features and they see it as perf issue, they can disable it.

pilot/pkg/xds/eds.go
ramaraochavali: What is the concern  - it is not n^2 any more and breaks at the earliest opportunity? Any suggestions?

pilot/pkg/serviceregistry/kube/controller/endpoints.go
```diff
@@ -199,21 +200,35 @@ func (e *endpointsController) buildIstioEndpoints(endpoint interface{}, host hos
...
 	for _, ss := range ep.Subsets {
-		for _, ea := range ss.Addresses {
-			pod, expectedPod := getPod(e.c, ea.IP, &metav1.ObjectMeta{Name: ep.Name, Namespace: ep.Namespace}, ea.TargetRef, host)
-			if pod == nil && expectedPod {
-				continue
-			}
-			builder := NewEndpointBuilder(e.c, pod)
+		endpoints = append(endpoints, e.buildIstioEndpointFromAddress(ep, ss, ss.Addresses, host, discoverabilityPolicy, true)...)
+		if features.SendUnhealthyEndpoints {
+			endpoints = append(endpoints, e.buildIstioEndpointFromAddress(ep, ss, ss.NotReadyAddresses, host, discoverabilityPolicy, false)...)
```
howardjohn, line 205: nit: instead of `ready bool` use `health model.HealthStatus`. Simpler and more readable

pilot/pkg/xds/eds.go
```diff
@@ -165,9 +167,38 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
...
+	if !features.SendUnhealthyEndpoints {
+		return true, true
 	}
-	return fullPush
+	// Check if new Endpoints are ready to be pushed. This check
+	// will ensure that if a new pod comes with a non ready endpoint,
+	// we do not unnecessarily push that config to Envoy.
+	emap := make(map[string]*model.IstioEndpoint, len(oldIstioEndpoints))
+	for _, oie := range oldIstioEndpoints {
+		emap[oie.Address] = oie
```
howardjohn, line 183: I would note (in the comments as well) that address is not a unique key. So this may push too many. Which is ok since its an optimization, but should be clearly noted.

ramaraochavali:
@howardjohn addressed comments. PTAL

ramaraochavali:
@howardjohn @hzxuzhonghu can you PTAL when you get chance?

howardjohn (approved):

howardjohn:
seems huge regression in postsubmit since this such as https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-120_istio_postsubmit/1486472822791344128. Suddenly failed 10x

ramaraochavali:
I am surprised - what is different in postsubmit?

Please revert this while I figure out the problem

ramaraochavali:
Or I can disable it by default - Should I do that?

howardjohn:
Yeah I have no clue really tbh. The only thing is it runs all the test
sequential instead of split up into different jobs.

I'll peek at it a bit more tomorrow and revert if i can't figure it out.
Could also be a coincidental timing

Issue #36278, opened 2021-11-26
Enable proxying Expect: 100-Continue by default


sschepens:
**Describe the feature request**

Default to enabling Envoy's `proxy_100_continue` in HttpConnectionManagers.

This is needed to prevent issues while uploading files and being redirected by the upstream, for example by using S3 pre signed requests.

There shouldn't be any downsides or issues to enable this by default, as suggested in [#21382](https://github.com/istio/istio/issues/21382)

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Additional context**

Issue #21382, opened 2020-02-23
Istio sidecar stripping Expect: 100-Continue header. 


JagadeeshSreeram:
**Bug description**

While uploading a large file we want to first authenticate the request before accepting the data. When client send Expect: 100-Continue, we want to validate authenticate the request and return 401 if request is not valid . Since istio proxy strips off the Expect headers and sent 100-continue back by default, our server doesn't even know if client sent the Expect header.

It looks like envoy supports this feature as part of 
https://github.com/envoyproxy/envoy/pull/2497

But there is no way in istio to enable this. I tried writing Envoy filter , but it fails with below error

```
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: enable-expect-header-envoyfilter
  namespace: confighub
spec:
  configPatches:
  - applyTo: NETWORK_FILTER
    match:      
      listener:
        filterChain:
          filter:
            name: "envoy.http_connection_manager"
    patch:
      operation: MERGE
      value:
        typed_config:
          "@type": "type.googleapis.com/envoy.config.filter.network.http_connection_manager.v2.HttpConnectionManager"
          proxy_100_continue: true

```

Error 

```
admission webhook "pilot.validation.istio.io" denied the request: configuration is invalid: envoy filter: missing filters
```

**Expected behavior**

**Steps to reproduce the bug**

**Version (include the output of `istioctl version --remote` and `kubectl version` and `helm version` if you used Helm)**

**How was Istio installed?**

**Environment where bug was observed (cloud vendor, OS, etc)**

howardjohn:
I have tried the Envoyfilter on latest code, it works. You must have an old Istio version.

@rshriram @lambdai is this something we should consider enabling by default? Not sure of the implications of this

lambdai:
100-continue must be implemented in each hop.

My understanding is that we could turn on `proxy_100_continue` everywhere: it doesn't guarantee deliver to client, but it should not raise any trouble either, for http1.0, http1.1, http2

@PiotrSikora Is above true?

PR #36279, opened 2021-11-27
set proxy 100 continue to true by default, so that envoy will not str…

hzxuzhonghu:
…ip Expect: 100-continue header

Fix: #36278 #21382

Will add ut if this is the right direction.


howardjohn (commented):
will this have negative impact on any users?

hzxuzhonghu:
From api doc, there is no influence. cc @lambdai

hzxuzhonghu:
@lambdai WDYT?

hzxuzhonghu:
ping @lambdai

Issue #35429, opened 2021-09-30
service with externalname overrides istio-proxy 443 rules


ricosega:
### Bug Description

I have some services without selector pointing to endpoints for external services to the cluster in order to monitor them.
At the moment I create the following object:

```
apiVersion: v1
kind: Service
metadata:
  name: nexus-metrics-monitoring
  namespace: external-monitoring
  labels:
    app: nexus-metrics-monitoring
spec:
  externalName: nexus.internal
  ports:
  - name: nexus
    port: 4443
    protocol: TCP
    targetPort: 443
  - name: node
    port: 9100
    protocol: TCP
    targetPort: 9100   
  - name: containers
    port: 8080
    protocol: TCP
    targetPort: 8080       
  type: ExternalName
```
The TLS connections originated from inside the cluster in a pod with the istio-proxy sidecar starts to fail returning me 404 errors.
This is the ```istioctl proxy-config listeners toolkit-68754d4d6b-d7hm5 | grep 443 ``` output:
```
0.0.0.0        443   Trans: raw_buffer; App: HTTP                                             Route: 443
0.0.0.0        443   ALL                                                                      Cluster: outbound|443||nexus-metrics-monitoring.external-monitoring.svc.cluster.local
10.100.107.185 443   ALL                                                                      Cluster: outbound|443||istiod.istio-system.svc.cluster.local
10.107.150.172 443   ALL                                                                      Cluster: outbound|443||istio-ingressgateway.istio-system.svc.cluster.local
10.107.212.31  443   ALL                                                                      Cluster: outbound|443||keycloak.keycloak.svc.cluster.local
10.108.30.165  443   ALL                                                                      Cluster: outbound|443||istio-egressgateway.istio-system.svc.cluster.local
10.96.0.1      443   ALL                                                                      Cluster: outbound|443||kubernetes.default.svc.cluster.local
10.97.216.209  443   ALL                                                                      Cluster: outbound|443||step-certificates.step-certificates.svc.cluster.local
10.97.92.163   443   ALL                                                                      Cluster: outbound|443||monitoring-operator.monitoring.svc.cluster.local
```

And this is the output of the same command when the service is deleted and all works fine:
```
0.0.0.0        443   ALL                                                                      PassthroughCluster
10.100.107.185 443   ALL                                                                      Cluster: outbound|443||istiod.istio-system.svc.cluster.local
[...]

PR #36281, opened 2021-11-28
Fixes Issue #35429 

eliavem:
This PR fixes issue #35429. It adds an error issue for ExternalName service type when service does not follow naming convention or has an unnamed port 

- [x] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience


galley/pkg/config/analysis/analyzers/service/portname.go
```diff
@@ -72,6 +72,11 @@ func (s *PortNameAnalyzer) analyzeService(r *resource.Instance, c analysis.Conte
 			m := msg.NewPortNameIsNotUnderNamingConvention(
 				r, port.Name, int(port.Port), port.TargetPort.String())
 
+			if svc.Type == "ExternalName" {
+				m = msg.NewExternalNameServiceTypeInvalidPortName(
+					r, port.Name, int(port.Port), port.TargetPort.String())
+			}
```
linsun, line 78: I'm thinking only to generate the error for port 443 since it is a common protocol for HTTPS.  Declaring `protocol: tcp` could be valid if tcp is the actual protocol, for example on port 9090.

galley/pkg/config/analysis/analyzers/testdata/incorrect-port-name-external-name-service-type.yaml
```diff
@@ -0,0 +1,28 @@
...
+kind: Service
+metadata:
+  name: nginx
+  namespace: nginx-ns
+spec:
+  externalName: nginx.example.com
+  ports:
+  - name: nginx
+    port: 443
+    protocol: TCP
```
linsun, line 13: Note protocol here is declared, it is TCP. [-suffix] is optional when declaring protocol.

cc @bianpengyuan for input also

galley/pkg/config/analysis/analyzers/service/portname.go
```diff
@@ -72,6 +72,11 @@ func (s *PortNameAnalyzer) analyzeService(r *resource.Instance, c analysis.Conte
 			m := msg.NewPortNameIsNotUnderNamingConvention(
 				r, port.Name, int(port.Port), port.TargetPort.String())
 
+			if svc.Type == "ExternalName" {
```
bianpengyuan, line 75: Should we skip warning on `NewPortNameIsNotUnderNamingConvention` if svc type is external name? Otherwise we are generating two somewhat duplicated warnings for an ExternalName svc.

galley/pkg/config/analysis/msg/messages.yaml
```diff
@@ -535,3 +535,12 @@ messages:
     description: "user namespace should be injectable if Istio is installed with enableNamespacesByDefault enabled and neither injection label is set."
     template: "is enabled for Istio injection, as Istio is installed with enableNamespacesByDefault as true."
     url: "https://istio.io/latest/docs/reference/config/analysis/ist0148/"
+
+  - name: "ExternalNameServiceTypeInvalidPortName"
+    code: IST0150
+    level: Error
+    description: "Port name is not under naming convention. Proxy blocks communication to external services if service is of type ExternalName."
+    template: "Port name %s (port: %d, targetPort: %s) doesn't follow the naming convention of Istio port. Proxy will block communication to defined external service"
```
bianpengyuan, line 543: I don't think this is accurate. It does not necessarily block traffic to external service, but rather incorrectly route unmatched traffic. Probably something like "Proxy could potentially prevent TCP traffic on the same port from being forwarded correctly" would be better.

galley/pkg/config/analysis/analyzers/analyzers_test.go
```diff
@@ -608,6 +608,15 @@ var testGrid = []testCase{
 			{msg.ImageAutoWithoutInjectionError, "Pod default/injected-pod"},
 		},
 	},
+	{
+		name:       "ExternalNameServiceTypeInvalidPortName",
+		inputFiles: []string{"testdata/incorrect-port-name-external-name-service-type.yaml"},
+		analyzer:   &service.PortNameAnalyzer{},
+		expected: []message{
+			{msg.ExternalNameServiceTypeInvalidPortName, "Service nginx-ns/nginx"},
+			{msg.ExternalNameServiceTypeInvalidPortName, "Service nginx-ns2/nginx-svc2"},
```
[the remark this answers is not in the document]
eliavem: I will work on those, thanks for the feedback!

galley/pkg/config/analysis/analyzers/service/portname.go
eliavem: It is an error message should I convert it to a warning only?. 

Actually, NewPortNameIsNotUnderNamingConvention is skipped is service type is ExternalName and port does not follow the naming convention. No duplicate warnings are displayed

galley/pkg/config/analysis/msg/messages.yaml
eliavem: Yes, I can make that change. I just wonder will it "potentially prevent TCP traffic"? I cannot think of a case where it will actually succeed

howardjohn (commented):
This seems to have changed a bunch of unexpected things. Maybe a merge conflict?

eliavem:
@linsun Istio fork was outdated for me. I have galley while galley has been removed. I will close this PR and submit a new one

galley/pkg/config/analysis/analyzers/service/portname.go
eliavem: Tested this and no duplicated message shows. The message gets assigned issue ExternalNameServiceTypeInvalidPortName when svc is of type external name

galley/pkg/config/analysis/msg/messages.yaml
linsun: +1 to what @bianpengyuan said.

linsun:
Thank you @eliavem !!

Issue #36162, opened 2021-11-19
serviceCluster behaviour change in 1.12


Stono:
### Bug Description

Historically the `serviceCluster` on istio-proxy correlated to the span name (see https://github.com/istio/istio/issues/12644#issuecomment-973867802)

However since upgrading to 1.12, this no longer seems to be the case:

<img width="209" alt="142591824-a445a7cb-cc49-40f4-9071-7426e7486d40" src="https://user-images.githubusercontent.com/1486729/142664917-231a6fc2-4498-42a1-b07f-abb283cc1fde.png">

<img width="413" alt="142591849-53ed12cd-7f92-46e7-9317-03b1d742a9d8" src="https://user-images.githubusercontent.com/1486729/142664925-b48cb1b7-736c-491b-b5c3-22073cc2269b.png">

So this feels like a regression.

@howardjohn pointed towards https://github.com/istio/istio/pull/35625 as a potential cause?

### Version

```prose
1.12
```

douglas-reid:
@kyessenov thoughts here?  it looks you moved control over configurability in https://github.com/istio/istio/pull/32407.

Stono:
Ah so serviceCluster is no longer used?  How would I go about controlling the service name in spans now then?

douglas-reid:
@howardjohn @Stono wdyt about supporting something like this (as a workaround for setting proxyconfig annotations on each pod, etc.): https://github.com/istio/istio/compare/master...douglas-reid:experiment-service-cluster?expand=1 ?

kyessenov:
We're trying to remove customization via templates (envoy and pod templates) and use actual API for such customizations. Doesn't ProxyConfig.serviceCluster work for you? That field is still honored I think. It can be injected as an annotation and probably migrate to a separate API later.

Stono:
@kyessenov I get that - but it certainly feels like this has been removed and an alternative isn't really there (or it isn't clear to me what to do).  

We could go and add that annotation to all of our workloads I suppose, but we're talking 500+ of them, previously this was nicely handled by the injector and dynamically (in our case) set to the app label by doing (`--serviceCluster {{ index .ObjectMeta.Labels `app` }}"`).

@douglas-reid not sure what your PR shows, I don't know golang really and don't have context of the code base :)

kyessenov:
[...]

PR #36285, opened 2021-11-29
Support templates in serviceCluster configuration

douglas-reid:
This PR was created in response to https://github.com/istio/istio/issues/36162. It attempts to allow user control over `serviceCluster` names, based on golang templates over `model.BootstrapNodeMetadata`. This would avoid having to create per-workload-instance annotations across an installation.

WARNING: This would be exposing, in some fashion, Istio implementation details as part of a "hidden" API.

With this PR, the user/admin would install with config like:

```yaml
defaultConfig:
  proxyConfig:
    serviceCluster: {{ .Labels.app }}
```

to customize the `serviceCluster` behavior for an entire mesh.

- [ X ] User Experience


pkg/bootstrap/config.go
```diff
@@ -278,9 +280,20 @@ func getServiceCluster(metadata *model.BootstrapNodeMetadata) string {
...
+	tmpl, err := template.New("serviceCluster").Parse(serviceCluster)
+	if err != nil {
+		log.Infof("could not parse template: %v", err)
+		return serviceCluster
+	}
+	buf := &bytes.Buffer{}
+	err = tmpl.Execute(buf, metadata)
+	if err != nil {
+		log.Warnf("could not execute template: %v", err)
+		return serviceCluster
```
kyessenov, line 294: Returning the template does not sound right.

kyessenov:
I think this is really an API change. While I'm generally OK with this, I think it would be better to have a list of pre-determined and reasonable options.

douglas-reid:
> I think this is really an API change. While I'm generally OK with this, I think it would be better to have a list of pre-determined and reasonable options.

I'm fairly conflicted about this myself.

How do you propose to expose the options?

pkg/bootstrap/config.go
douglas-reid: fixed.

douglas-reid:
@kyessenov @Stono @howardjohn wdyt about something like: https://github.com/istio/api/compare/master...douglas-reid:service-cluster-config?expand=1 as an alternative?

Stono:
I'm happy with either, the latter puts more emphasis on istio to ensure each of those enums remains functional across releases however whereas the former puts more control in the users hands at the risk of future breaking 🤷 

In my personal case the latter would be preferred as it's then almost a guaranteed API.  It could always be revisited with an additional `TEMPLATE` option or something in the future, should the requirement for more flexibility arise?

PR #36295, opened 2021-11-30
Add a building script for Arm Platform

TrevorTaoARM:
Now most Istio components/images can be built on arm platform and installed with
the Istioctl by the built images but needs some special commands
to process all things.
This script is based on our practice of Istio work on arm.
We think it would greatly benefit the development and propagation of Istio and its
application at the current phase and can be removed in the future if everytihing is ok for
arm platform as others.

Refer:
https://github.com/aws/aws-graviton-getting-started/blob/main/containers-workarounds.md#Istio

### DIFF
--- a/tools/build-for-arm.sh
+++ b/tools/build-for-arm.sh
@@ -1,0 +1,75 @@
+#!/bin/bash
+#
+# Copyright 2021 Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Note: Experimental only for the use of building your own on arm64 platform
+
+set -ex
+
+CURR_DIR=$(dirname "${BASH_SOURCE[0]}")
+O_HUB=${1:-istio}
+
+if [ ! -d ${CURR_DIR}/istio ]; then
+    git clone https://github.com/istio/istio.git &&
+    git clone https://github.com/istio/proxy.git &&
+    git clone https://github.com/istio/tools.git
+fi
+
+# Build build-tools and build-tools-proxy
+pushd ${CURR_DIR}/tools/docker/build-tools
+if [ -z "$(docker images |grep "build-tools-proxy" | grep "master-latest")" ]; then
+    DRY_RUN=1 time ./build-and-push.sh
+fi
+popd
+
+# Build Istio binary
+pushd ${CURR_DIR}/istio
+TARGET_ARCH=arm64 IMAGE_VERSION=master-latest make build
+popd
+
+
+export IMAGE_TAG=master-latest
+export IMAGE_VERSION=$IMAGE_TAG
+export DOCKER_ARCHITECTURES="linux/arm64"
+export TARGETARCH="arm64"
+export TAG=$IMAGE_TAG
+export BASE_VERSION=$IMAGE_TAG
+export TARGET_ARCH=arm64
+export CI=true
+
+# Build Envoy and install envoy
+export PROXY_REPO_SHA=$(cat ${CURR_DIR}/istio/istio.deps |grep "lastStableSHA" | cut -d '"' -f 4)
+pushd ${CURR_DIR}/proxy
+git checkout $PROXY_REPO_SHA
+popd
+
+aarch64_bin=$(file ${CURR_DIR}/istio/out/linux_arm64/release/envoy | grep aarch64) || true
+if [ -z "${aarch64_bin}" ]; then
+    pushd proxy
+    BUILD_WITH_CONTAINER=1 make build
+    BUILD_WITH_CONTAINER=1 make exportcache
+    popd
+    cp ${CURR_DIR}/proxy/out/linux_arm64/envoy ${CURR_DIR}/istio/out/linux_arm64/release
+fi
+
+
+# Build Istio base containers
+cd ${CURR_DIR}/istio
+DOCKER_TARGETS='docker.base docker.distroless' \
+         HUBS="gcr.io/istio-release" TARGET_ARCH="arm64" make docker
+
+# Build Istio core containers
+DOCKER_TARGETS='docker.pilot docker.proxyv2 docker.app docker.pilot docker.install-cni docker.istioctl docker.operator' \
+         HUBS=$O_HUB make docker


tools/build-for-arm.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
howardjohn, line 1: We should probably have some notice that this is experimental only, etc?

tools/build-for-arm.sh
```diff
@@ -0,0 +1,84 @@
...
+if [ ! -d ${CURR_DIR}/istio ]; then
+    git clone https://github.com/istio/istio.git &&
+    git clone https://github.com/istio/proxy.git &&
+    git clone https://github.com/istio/tools.git
+fi
+
+# Build build-tools and build-tools-proxy
+pushd ${CURR_DIR}/tools/docker/build-tools
+if [ -z "$(docker images |grep "build-tools-proxy" | grep "master-latest")" ]; then
+    DRY_RUN=1 time ./build-and-push.sh
```
howardjohn, line 30: Seems like we should assert users have HUB set to something? Otherwise it will default to Istio official GCR which will either fail or if some devs have permission unexpectedly push

tools/build-for-arm.sh
```diff
@@ -0,0 +1,84 @@
...
+popd
+
+# Build Istio binary
+pushd ${CURR_DIR}/istio
+TARGET_ARCH=arm64 IMAGE_VERSION=master-latest make build
+popd
+
+# Build Envoy and install envoy
+pushd ${CURR_DIR}/proxy
+export PROXY_REPO_SHA=$(git rev-parse HEAD)
```
howardjohn, line 41: building from HEAD of istio/proxy is not quite right, should use the SHA in istio.deps

tools/build-for-arm.sh
```diff
@@ -0,0 +1,84 @@
...
+popd
+if [ ! -f ${CURR_DIR}/istio/out/linux_arm64/release/envoy ]; then
+    docker rm -f istio-proxy-build || true
+    docker run --name istio-proxy-build -it -v $(realpath ${CURR_DIR}/proxy):/work -w /work -u $(id -u) gcr.io/istio-testing/build-tools-proxy:master-latest make build
```
howardjohn, line 45: ``BUILD_WITH_CONTAINER=1 make build`

tools/build-for-arm.sh
```diff
@@ -0,0 +1,84 @@
...
+    docker cp istio-proxy-build:/work/bazel-bin/src/envoy/envoy ${CURR_DIR}/istio/out/linux_arm64/release/envoy-${PROXY_REPO_SHA}
```
howardjohn, line 46: I think `exportcache` target is supposed to do this. Not sure that works as expected though

tools/build-for-arm.sh
```diff
@@ -0,0 +1,84 @@
...
+cd ${CURR_DIR}/istio
+for i in 'dockerx.base' 'dockerx.distroless' \
+         'dockerx.app_sidecar_base_ubuntu_xenial' \
+         'dockerx.app_sidecar_base_ubuntu_bionic' \
+         'dockerx.app_sidecar_base_ubuntu_focal' \
+         'dockerx.app_sidecar_base_debian_9' \
+         'dockerx.app_sidecar_base_debian_10' \
+         'dockerx.app_sidecar_base_centos_8' \
+         'dockerx.app_sidecar_base_centos_7'; \
+          do HUBS="gcr.io/istio-release" TARGET_ARCH="arm64" make $i; done
```
howardjohn, line 74: This seems unexpected (HUB=gcr.io/istio-release). Shouldn't user be pushing to their own hubs?

tools/build-for-arm.sh
```diff
@@ -0,0 +1,84 @@
...
+export BASE_VERSION=$IMAGE_TAG
+pushd ${CURR_DIR}/proxy
+export PROXY_REPO_SHA=$(git rev-parse HEAD)
+popd
+export TARGET_ARCH=arm64
+export CI=true
+
+# Build Istio base containers
+cd ${CURR_DIR}/istio
+for i in 'dockerx.base' 'dockerx.distroless' \
```
howardjohn, line 66: `DOCKER_TAGRETS='docker.base docker.pilot .....' make docker` is substantial more efficient than this as it parallelizes everything

tools/build-for-arm.sh
```diff
@@ -0,0 +1,84 @@
...
+         'dockerx.app_sidecar_base_ubuntu_xenial' \
+         'dockerx.app_sidecar_base_ubuntu_bionic' \
+         'dockerx.app_sidecar_base_ubuntu_focal' \
+         'dockerx.app_sidecar_base_debian_9' \
+         'dockerx.app_sidecar_base_debian_10' \
+         'dockerx.app_sidecar_base_centos_8' \
+         'dockerx.app_sidecar_base_centos_7'; \
```
howardjohn, line 73: These are likely not needed, for most cases anyhow. Its just for VM tests

tools/build-for-arm.sh
howardjohn: Actually why do we need to build it at all? The base images are already built for arm64?

howardjohn (commented):
cool! left a few comments

tools/build-for-arm.sh
TrevorTaoARM: Yes, 1 note would be added here.

tools/build-for-arm.sh
TrevorTaoARM: BUILD_WITH_CONTAINER seems to be default to 1 in Makefile.overrides.mk.
Anyway, I would add it.

TrevorTaoARM: Here the building should already be done in the container, so I think there is no need to add a BUILD_WITH_CONTAINER=1 here.

tools/build-for-arm.sh
TrevorTaoARM: Yes, I would change it.

tools/build-for-arm.sh
TrevorTaoARM: Yes, we just use the default value of "gcr.io/istio-testing" to build an image named as gcr.io/istio-testing/build-tools-proxy:master-latest locally and use it the base building image for other images.

tools/build-for-arm.sh
TrevorTaoARM: yes, exportcache would copy envoy binary to its out/linux_arm64, and we still need to copy it to istio/out/linux_arm64.

tools/build-for-arm.sh
TrevorTaoARM: ok, I would change it.

tools/build-for-arm.sh
TrevorTaoARM: The base images building would be removed if not needed.

tools/build-for-arm.sh
howardjohn: I mean replace this entire line with this. We should not need to invoke `docker` directly

tools/build-for-arm.sh
TrevorTaoARM: Currently I think we can just choose the default one because the following building would rely on this image, or we need to explicitly set the IMG variable for subsequent building. 
I can do it subsequently if it's actual needed. But currently it works fine for a local building on arm.

tools/build-for-arm.sh
TrevorTaoARM: Updated.

tools/build-for-arm.sh
TrevorTaoARM: Updated.

tools/build-for-arm.sh
TrevorTaoARM: Updated.

tools/build-for-arm.sh
TrevorTaoARM: Updated.

tools/build-for-arm.sh
TrevorTaoARM: Updated. Thanks.

tools/build-for-arm.sh
TrevorTaoARM: Updated. Thanks.

jacob-delgado:
cc @TrevorTaoARM please consider joining the istio community (see the above comment). I know you've put up a few PRs recently that have gotten merged. While it is unlikely that we can support ARM in open source, for now, if you have access to your own prow cluster running on ARM you can build a fork of istio until GCP lands ARM vm support. Thanks!

TrevorTaoARM:
> cc @TrevorTaoARM please consider joining the istio community (see the above comment). I know you've put up a few PRs recently that have gotten merged. While it is unlikely that we can support ARM in open source, for now, if you have access to your own prow cluster running on ARM you can build a fork of istio until GCP lands ARM vm support. Thanks!
@jacob-delgado Thank you. Your comments are of great value to us and we will consider the CI issues under them. 
Hope more support and cooperation from you and Istio community in the future.

TrevorTaoARM:
@craigbox Thank you.

howardjohn:
This is supported natively now. thanks!!

PR #36300, opened 2021-11-30
revert fault filter change

ramaraochavali:
As being discussed https://github.com/istio/istio/pull/36259#issuecomment-982286784 and below, I think we do not need to add filter at top level and per route config will take care of it.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu (commented):
LGTM, give a chance to @howardjohn to take a look

ramaraochavali:
Looked at Envoy code again, the original filter chain is built based on the filters defined in HCM and when the filter is executed it resolves the most specific config. So adding it to http filters is needed. Not sure how csrf test passed for me yesterday.

ramaraochavali:
@howardjohn @hzxuzhonghu can PTAL? Any concerns with this?

howardjohn (approved):

PR #36307, opened 2021-11-30
multicluster: add allowlist for kubeconfig auth methods

howardjohn:

### DIFF
--- a/pkg/kube/multicluster/secretcontroller.go
+++ b/pkg/kube/multicluster/secretcontroller.go
@@ -35,9 +35,11 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/client-go/util/workqueue"
 
 	"istio.io/istio/pilot/pkg/features"
+	"istio.io/istio/pilot/pkg/util/sets"
 	"istio.io/istio/pkg/cluster"
 	"istio.io/istio/pkg/kube"
 	"istio.io/pkg/log"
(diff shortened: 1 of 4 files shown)


stevenctl (approved):

pkg/kube/multicluster/secretcontroller.go
```diff
@@ -444,6 +449,66 @@ var BuildClientsFromConfig = func(kubeConfig []byte) (kube.Client, error) {
...
+					return fmt.Errorf("auth provider %s is not allowed", ap.Name)
+				}
+			}
+		}
+		if auths.ClientKey != "" && !allowlist.Contains("clientKey") {
+			return fmt.Errorf("clientKey is not allowed")
+		}
+		if auths.ClientCertificate != "" && !allowlist.Contains("clientCertificate") {
+			return fmt.Errorf("clientCertificate is not allowed")
+		}
```
rohitagarwal003, line 475: Hey @howardjohn,

Can you provide more information about how having this (paths to a tls client cert and key) is insecure?

I looked at https://github.com/kubernetes/kubectl/issues/697 but couldn't see anything related to how these can be leaked.

howardjohn: Attack vector is a user creates a kubeconfig with `clientKey: /some/path/in/istiod/container`. Then they set the API server URL to some server that records the key sent. This allows ex-filtrating arbitrary files from the Istio filesystem.

rohitagarwal003: `clientKey` is the path to the TLS private key, it's not sent over the wire, correct?

---

The argument may apply to `clientCertificate` which does get sent over the wire. Though even there I think the TLS stack will not sent arbitrary bytes over the wire and we will fail validation (e.g. [in here](https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/crypto/tls/tls.go;l=265))

howardjohn: Correct, I should have said clientCertificate - it may not allow _arbitrary_ exfiltration in this case, but is still a SSRF. For example, Istiod may have high privileged mtls certs, and now a user can have them send requests using those credentials to arbitrary services.

If you have legitimate use cases you should be able to re-enable this, right?

rohitagarwal003: Yeah, I can re-enable this. I just wanted to understand what I am re-enabling. Thanks for the explanation.

howardjohn: :+1: . Token auth is much worse than mTLS

PR #36309, opened 2021-11-30
Add find root cert fuzzer to tests

howardjohn:

### DIFF
--- a/security/pkg/pki/util/verify_cert.go
+++ b/security/pkg/pki/util/verify_cert.go
@@ -174,6 +174,9 @@ func FindRootCertFromCertificateChainBytes(certBytes []byte) ([]byte, error) {
 		if len(certBytes) == 0 {
 			break
 		}
+		if block == nil {
+			return nil, fmt.Errorf("error decoding certificate")
+		}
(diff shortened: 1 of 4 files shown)


lei-tang (commented):
"tests/fuzz/testdata/FuzzFindRootCertFromCertificateChainBytes/6118602084843520" is an empty file. Is it needed?

howardjohn:
Yes, its a regression test case. Generally the idea so far as has been that
test cases the fuzzer generates that fail the test are added as regression
tests. it being empty is a bit trivial, but in this case it did trigger a
failure previously so its good to keep around as reference and to ensure it
doesn't break in the future, IMO.

lei-tang:
Nit: why is the file named as "6118602084843520"? Will a name such as "empty-file" also work?

howardjohn:
Its correlated to the OSS fuzz test case. This is accessible only to PSWG.
This is the same pattern https://go.dev/blog/fuzz-beta uses

lei-tang (approved):

jacob-delgado (approved):

PR #36310, opened 2021-11-30
stop processing events for deleted cluster

stevenctl:
Fixes stuck endpoint shards due a service/endpoint being added in the middle of cluster deletion, before the queue is stopped. This will cause the endpoint to linger forever. 

- [ ] TODO:There is still a slight issue here where we don't know if the queue is finished processing the last event it picked up. 

---

In a follow up I'd like to address how we can prevent removing then re-adding endpoints when doing an `Update` on a cluster secret. 

We also need a way to set priority across multiple secrets that provide access to the same cluster.


ramaraochavali (commented):
LGTM. Leaving it open for others to take a look

stevenctl:
Seems like this can easily cause some panics, need to rework it a bit.

hzxuzhonghu:
Is it possible to stop first and then call `c.handleDelete`
https://github.com/istio/istio/blob/8835b8da622a62bb55e47a46b2b2a26c1be53d83/pkg/kube/multicluster/secretcontroller.go#L589

And then i gues we need to modify the istio queue instance  to not process any more after it is stopped

stevenctl:
@hzxuzhonghu reworked this a bit, PTAL.

hzxuzhonghu:
Sorry for delay, will take a look

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -516,15 +515,12 @@ func (c *Controller) Network(endpointIP string, labels labels.Instance) network.
...
+	if err := queue.WaitForClose(c.queue, 30*time.Second); err != nil {
+		log.Warnf("queue for removed kube registry %q may not be done processing: %v", c.Cluster(), err)
 	}
-	for _, s := range svcs {
-		for _, svc := range c.servicesForNamespacedName(kube.NamespacedNameForK8sObject(s)) {
-			c.opts.XDSUpdater.SvcUpdate(model.ShardKeyFromRegistry(c), svc.Hostname.String(),
-				s.Namespace, model.EventDelete)
-		}
+	if c.opts.XDSUpdater != nil {
+		c.opts.XDSUpdater.RemoveShard(model.ShardKeyFromRegistry(c))
```
hzxuzhonghu, line 522: As i see `SvcUpdate` already does remove shards, except this is a perf improvement, is there any diff?

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -516,15 +515,12 @@ func (c *Controller) Network(endpointIP string, labels labels.Instance) network.
 }
 
 func (c *Controller) Cleanup() error {
-	svcs, err := c.serviceLister.List(klabels.Everything())
-	if err != nil {
-		return fmt.Errorf("error listing services for deletion: %v", err)
+	if err := queue.WaitForClose(c.queue, 30*time.Second); err != nil {
```
hzxuzhonghu, line 518: What is this for?

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -297,12 +297,12 @@ func (m *Multicluster) ClusterDeleted(clusterID cluster.ID) error {
 		log.Infof("cluster %s does not exist, maybe caused by invalid kubeconfig", clusterID)
 		return nil
 	}
-	if err := kc.Cleanup(); err != nil {
-		log.Warnf("failed cleaning up services in %s: %v", clusterID, err)
-	}
 	if kc.workloadEntryStore != nil {
 		m.opts.MeshServiceController.DeleteRegistry(clusterID, provider.External)
 	}
+	if err := kc.Cleanup(); err != nil {
```
hzxuzhonghu, line 303: Does this order matter?

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -516,15 +515,12 @@ func (c *Controller) Network(endpointIP string, labels labels.Instance) network.
...
+		log.Warnf("queue for removed kube registry %q may not be done processing: %v", c.Cluster(), err)
 	}
-	for _, s := range svcs {
-		for _, svc := range c.servicesForNamespacedName(kube.NamespacedNameForK8sObject(s)) {
-			c.opts.XDSUpdater.SvcUpdate(model.ShardKeyFromRegistry(c), svc.Hostname.String(),
-				s.Namespace, model.EventDelete)
-		}
+	if c.opts.XDSUpdater != nil {
+		c.opts.XDSUpdater.RemoveShard(model.ShardKeyFromRegistry(c))
+		c.opts.XDSUpdater.ConfigUpdate(&model.PushRequest{Full: true, Reason: []model.TriggerReason{model.ClusterUpdate}})
```
hzxuzhonghu, line 523: This is duplicate with Multicluster.ClusterDeleted

pilot/pkg/serviceregistry/kube/controller/multicluster.go
stevenctl: Not really, as long as this is called _after_ the stopCh for `kc` is closed.

stevenctl: We could actually make it so `Cleanup` isn't an exported method, and instead schedule it from within `Run`.

pilot/pkg/serviceregistry/kube/controller/controller.go
stevenctl: We _must_ make sure the queue is actually done. If we don't we could end up removing endpointshards, then processing an event _after_ that, and the endpoint for the removed cluster will linger forever.

pilot/pkg/serviceregistry/kube/controller/controller.go
stevenctl: We rely on service lister which isn't 100% guaranteed to have everything that's in shards.

hzxuzhonghu: make sense, if a svc is deleted and hasn't yet been processed, there will be mem leak

pilot/pkg/serviceregistry/kube/controller/controller.go
hzxuzhonghu: Got it, suddenly i think of the below closing of  `stop` better move in front of `handleDelete`

```
func (c *Controller) deleteCluster(secretKey string, clusterID cluster.ID) {
	c.cs.Lock()
	defer func() {
		c.cs.Unlock()
		log.Infof("Number of remote clusters: %d", c.cs.Len())
	}()
	log.Infof("Deleting cluster_id=%v configured by secret=%v", clusterID, secretKey)
	err := c.handleDelete(clusterID)
	if err != nil {
		log.Errorf("Error removing cluster_id=%v configured by secret=%v: %v",
			clusterID, secretKey, err)
	}
	close(c.cs.remoteClusters[secretKey][clusterID].stop)
	delete(c.cs.remoteClusters[secretKey], clusterID)
}
```

pkg/queue/instance.go
```diff
@@ -97,7 +113,18 @@ func (q *queueImpl) pushRetryTask(item *BackoffTask) {
...
 }
 
+func (q *queueImpl) Closed() <-chan struct{} {
+	return q.closed
+}
+
 func (q *queueImpl) Run(stop <-chan struct{}) {
+	log.Debugf("started queue %s", q.id)
+	defer func() {
+		q.closeOnce.Do(func() {
```
hzxuzhonghu, line 123: maybe remove closeOnce

pilot/pkg/serviceregistry/kube/controller/controller.go
hzxuzhonghu: remove this line

pilot/pkg/xds/eds.go
```diff
@@ -170,6 +170,19 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
 	return fullPush
 }
 
+func (s *DiscoveryServer) RemoveShard(shardKey model.ShardKey) {
```
hzxuzhonghu, line 173: This is not enough, looks at the `deleteService`, also need UpdateServiceAccount

stevenctl: re-using `deleteService` now

pilot/pkg/serviceregistry/kube/controller/controller.go
stevenctl: moved

pkg/queue/instance.go
stevenctl: This helps with the goroutine leak test.. the closeOnce helps in scenarios where we `Run` more than once (which we should never do).. but i'd rather the queue be robust to improper usage.

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -518,10 +518,6 @@ func (c *Controller) Cleanup() error {
 	if err := queue.WaitForClose(c.queue, 30*time.Second); err != nil {
 		log.Warnf("queue for removed kube registry %q may not be done processing: %v", c.Cluster(), err)
 	}
-	if c.opts.XDSUpdater != nil {
-		c.opts.XDSUpdater.RemoveShard(model.ShardKeyFromRegistry(c))
```
hzxuzhonghu: @stevenctl Removed this op by mistake?

hzxuzhonghu (approved):
Thanks

PR #36311, opened 2021-11-30
Added scripts to perform following tasks

litong01:
1. Setup k8s cluster using kind and metallb
2. Create self signed certs and k8s secrets for Istio

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/listener_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener_builder.go
@@ -616,10 +616,7 @@ func (configgen *ConfigGeneratorImpl) buildInboundFilterchains(in *plugin.InputP
 			}
 		}
 	}
-	// only inject WasmPlugins for Sidecar inbound listeners and routers outbound listeners
-	if in.Node.Type == model.SidecarProxy {
-		extension.AddWasmPluginsToMutableObjects(mutable, in.Push.WasmPlugins(in.Node))
-	}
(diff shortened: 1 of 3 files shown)


frankbu:
Haven't tried them, but looks good.

Should add a couple of new sections to the README. For example: `Use kind clusters for multicluster testing` and `Configure certs for multicluster trust`.

irisdingbj (commented):
https://github.com/istio/istio/tree/master/tools/certs   Can this be used to perform par of your task?

litong01:
> https://github.com/istio/istio/tree/master/tools/certs Can this be used to perform par of your task?
@irisdingbj Iris, that document only creates the certs, adding the certs to k8s as a generic secret takes manual steps. The new scripts allow a user to issue one command which will produce right root and intermediate certs and produce the k8s secrets, the new script does not rely on make, it only consists of that script.

frankbu:
@litong01 I think it would make sense to change your script to use the makefiles to perform its task, instead of calling openssl directly. That way, your script is just a convenience layer over the makefiles, rather than a different implementation, which will just confuse people.

litong01:
> @litong01 I think it would make sense to change your script to use the makefiles to perform its task, instead of calling openssl directly. That way, your script is just a convenience layer over the makefiles, rather than a different implementation, which will just confuse people.

I thought that the idea of doing this is to have less dependency and less clutter. using make file requires a user to install make and other dependencies. vs this is just using bash. the make also uses openssl.

ericvn:
I agree with @frankbu to try and limit the implementations being duplicated between the makefile and scripts. We already have some changes where someone will update one and not both and they get out of sync.

PR #36315, opened 2021-12-01
controllers: abstract Kubernetes queue into common package

howardjohn:
We have made a lot of new controllers use the Kubernetes queue. Generally this is a good thing, but the API is way too low level. This leads to complex logic throughout much of our code, and its all slightly different with slightly different bugs.

This abstracts this into a new `controllers.Queue` type. In addition, secretcontroller, webhook, gateway deployment, and default revision controllers are moved over to the new queue

### DIFF
--- a/pilot/pkg/bootstrap/sidecarinjector.go
+++ b/pilot/pkg/bootstrap/sidecarinjector.go
@@ -99,7 +99,7 @@ func (s *Server) initSidecarInjector(args *PilotArgs) (*inject.Webhook, error) {
 				return nil
 			}
 
-			patcher.Run(stop)
+			go patcher.Run(stop)
 			return nil
 		})
 	}
--- a/pilot/pkg/config/kube/gateway/deploymentcontroller.go
+++ b/pilot/pkg/config/kube/gateway/deploymentcontroller.go
@@ -91,7 +91,9 @@ func NewDeploymentController(client kube.Client) *DeploymentController {
 			return err
 		},
 	}
-	dc.queue = controllers.NewQueue(controllers.WithReconciler(dc.Reconcile))
+	dc.queue = controllers.NewQueue("gateway deployment",
+		controllers.WithReconciler(dc.Reconcile),
+		controllers.WithMaxAttempts(5))
 
 	// Set up a handler that will add the parent Gateway object onto the queue.
 	// The queue will only handle Gateway objects; if child resources (Service, etc) are updated we re-add
--- a/pilot/pkg/config/kube/gateway/deploymentcontroller.go
+++ b/pilot/pkg/config/kube/gateway/deploymentcontroller.go
@@ -96,7 +96,7 @@ func NewDeploymentController(client kube.Client) *DeploymentController {
 	// Set up a handler that will add the parent Gateway object onto the queue.
 	// The queue will only handle Gateway objects; if child resources (Service, etc) are updated we re-add
 	// the Gateway to the queue and reconcile the state of the world.
-	handler := controllers.LatestVersionHandlerFuncs(controllers.EnqueueForParentHandler(dc.queue, gvk.KubernetesGateway))
+	handler := controllers.ObjectHandler(controllers.EnqueueForParentHandler(dc.queue, gvk.KubernetesGateway))
 
 	// Use the full informer, since we are already fetching all Services for other purposes
 	// If we somehow stop watching Services in the future we can add a label selector like below.
@@ -115,7 +115,7 @@ func NewDeploymentController(client kube.Client) *DeploymentController {
 
 	// Use the full informer; we are already watching all Gateways for the core Istiod logic
 	client.GatewayAPIInformer().Gateway().V1alpha2().Gateways().Informer().
-		AddEventHandler(controllers.LatestVersionHandlerFuncs(controllers.EnqueueForSelf(dc.queue)))
+		AddEventHandler(controllers.ObjectHandler(dc.queue.AddObject))
 
 	return dc
 }
--- a/pilot/pkg/config/kube/gateway/deploymentcontroller.go
+++ b/pilot/pkg/config/kube/gateway/deploymentcontroller.go
@@ -32,7 +32,6 @@ import (
 	appsinformersv1 "k8s.io/client-go/informers/apps/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/util/workqueue"
 	gateway "sigs.k8s.io/gateway-api/apis/v1alpha2"
 	"sigs.k8s.io/yaml"
 
@@ -68,7 +67,7 @@ import (
 // * This leaves YAML templates, converted to unstructured types and Applied with the dynamic client.
 type DeploymentController struct {
 	client    kube.Client
-	queue     workqueue.RateLimitingInterface
+	queue     controllers.Queue
 	templates *template.Template
 	patcher   patcher
 }
@@ -79,11 +78,25 @@ type patcher func(gvr schema.GroupVersionResource, name string, namespace string
 // NewDeploymentController constructs a DeploymentController and registers required informers.
 // The controller will not start until Run() is called.
 func NewDeploymentController(client kube.Client) *DeploymentController {
-	q := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+	dc := &DeploymentController{
+		client:    client,
+		templates: processTemplates(),
+		patcher: func(gvr schema.GroupVersionResource, name string, namespace string, data []byte, subresources ...string) error {
+			c := client.Dynamic().Resource(gvr).Namespace(namespace)
+			t := true
+			_, err := c.Patch(context.Background(), name, types.ApplyPatchType, data, metav1.PatchOptions{
+				Force:        &t,
+				FieldManager: ControllerName,
+			}, subresources...)
+			return err
+		},
+	}
+	dc.queue = controllers.NewQueue(controllers.WithReconciler(dc.Reconcile))
+
 	// Set up a handler that will add the parent Gateway object onto the queue.
 	// The queue will only handle Gateway objects; if child resources (Service, etc) are updated we re-add
 	// the Gateway to the queue and reconcile the state of the world.
-	handler := controllers.LatestVersionHandlerFuncs(controllers.EnqueueForParentHandler(q, gvk.KubernetesGateway))
+	handler := controllers.LatestVersionHandlerFuncs(controllers.EnqueueForParentHandler(dc.queue, gvk.KubernetesGateway))
 
 	// Use the full informer, since we are already fetching all Services for other purposes
 	// If we somehow stop watching Services in the future we can add a label selector like below.
@@ -102,56 +115,13 @@ func NewDeploymentController(client kube.Client) *DeploymentController {
 
 	// Use the full informer; we are already watching all Gateways for the core Istiod logic
 	client.GatewayAPIInformer().Gateway().V1alpha2().Gateways().Informer().
-		AddEventHandler(controllers.LatestVersionHandlerFuncs(controllers.EnqueueForSelf(q)))
+		AddEventHandler(controllers.LatestVersionHandlerFuncs(controllers.EnqueueForSelf(dc.queue)))
 
-	return &DeploymentController{
-		client:    client,
-		queue:     q,
-		templates: processTemplates(),
-		patcher: func(gvr schema.GroupVersionResource, name string, namespace string, data []byte, subresources ...string) error {
-			c := client.Dynamic().Resource(gvr).Namespace(namespace)
-			t := true
-			_, err := c.Patch(context.Background(), name, types.ApplyPatchType, data, metav1.PatchOptions{
-				Force:        &t,
-				FieldManager: ControllerName,
-			}, subresources...)
-			return err
-		},
-	}
+	return dc
 }
 
 func (d *DeploymentController) Run(stop <-chan struct{}) {
-	defer d.queue.ShutDown()
-	log.Infof("starting gateway deployment controller")
-	go func() {
-		// Process updates until we return false, which indicates the queue is terminated
-		for d.processNextItem() {
-		}
-	}()
-	<-stop
-}
-
-func (d *DeploymentController) processNextItem() bool {
-	// Wait until there is a new item in the working queue
-	key, quit := d.queue.Get()
-	if quit {
-		return false
-	}
-
-	log.Debugf("handling update for %v", key)
-
-	defer d.queue.Done(key)
-
-	err := d.Reconcile(key.(types.NamespacedName))
-	if err != nil {
-		if d.queue.NumRequeues(key) < 5 {
-			log.Errorf("error handling %v, retrying: %v", key, err)
-			d.queue.AddRateLimited(key)
-		} else {
-			log.Errorf("error handling %v, and retry budget exceeded: %v", key, err)
-		}
-	}
-	return true
+	d.queue.Run(stop)
 }
 
 // Reconcile takes in the name of a Gateway and ensures the cluster is in the desired state
--- a/pilot/pkg/serviceregistry/kube/controller/multicluster.go
+++ b/pilot/pkg/serviceregistry/kube/controller/multicluster.go
@@ -239,7 +239,7 @@ func (m *Multicluster) ClusterAdded(cluster *multicluster.Cluster, clusterStopCh
 			if err != nil {
 				log.Errorf("could not initialize webhook cert patcher: %v", err)
 			} else {
-				patcher.Run(clusterStopCh)
+				go patcher.Run(clusterStopCh)
 			}
 		}
 		// Patch validation webhook cert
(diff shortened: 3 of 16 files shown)


pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -79,11 +78,25 @@ type patcher func(gvr schema.GroupVersionResource, name string, namespace string
...
+			c := client.Dynamic().Resource(gvr).Namespace(namespace)
+			t := true
+			_, err := c.Patch(context.Background(), name, types.ApplyPatchType, data, metav1.PatchOptions{
+				Force:        &t,
+				FieldManager: ControllerName,
+			}, subresources...)
+			return err
+		},
+	}
+	dc.queue = controllers.NewQueue(controllers.WithReconciler(dc.Reconcile))
```
stevenctl, line 94: nit: can be part of the previous statement

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,168 @@
...
+	return func(q *Queue) {
+		q.work = func(key interface{}) error {
+			return f(key.(types.NamespacedName))
+		}
+	}
+}
+
+// NewQueue creates a new queue
+func NewQueue(options ...func(*Queue)) Queue {
+	q := Queue{
```
stevenctl, line 70: nit: Is there a reason we don't want pointer receiver or return a pointer here?

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,168 @@
...
+	return func(q *Queue) {
+		q.maxAttempts = n
+	}
+}
+
+// WithReconciler defines the to handle items on the queue
+func WithReconciler(f func(name types.NamespacedName) error) func(q *Queue) {
+	return func(q *Queue) {
+		q.work = func(key interface{}) error {
+			return f(key.(types.NamespacedName))
```
stevenctl, line 63: Is it better to type assert here and return an error? 
If there is no way to enqueue a `work` func that doesn't use `NamespacedName`, should we even export `Add(item interface{}`?

pilot/pkg/config/kube/gateway/deploymentcontroller.go
howardjohn: It references `dc` so I don't think it can?

pkg/kube/controllers/queue.go
howardjohn: I was going to have a function that allowed `interface{}` but turned out that nothing needed it (yet??). Likely its a bad idea to use anything else, so maybe we should just clean this up and not export Add

pkg/kube/controllers/queue.go
howardjohn: I don't know. Is there a reason we should return a pointer? Seems most Queues do so maybe thats reason enough

pilot/pkg/config/kube/gateway/deploymentcontroller.go
stevenctl: Misread, thought `Reconcile` was a struct member on first readthrough.

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,168 @@
...
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	})
+}
+
+// Run the queue. This is synchronous, so should typically be called in a goroutine.
+func (q Queue) Run(stop <-chan struct{}) {
+	defer utilruntime.HandleCrash()
+	defer q.queue.ShutDown()
+	log.Infof("starting %v", q.name)
```
Monkeyanator, line 105: nit: could store a logger on the struct with queue name prefix instead of prepending everywhere

pkg/revisions/default_watcher.go
```diff
@@ -159,20 +100,19 @@ func (p *defaultWatcher) notifyHandlers() {
...
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	p.defaultRevision = revision
 	p.notifyHandlers()
+	return nil
 }
 
-func filterUpdate(obj metav1.Object) bool {
-	return obj.GetName() != defaultTagWebhookName
+func filterUpdate(obj controllers.Object) bool {
```
Monkeyanator, line 116: very nit: counterintuitive naming, filterUpdate should be true when the object is not the default tag

pkg/revisions/default_watcher.go
```diff
@@ -53,35 +49,29 @@ type defaultWatcher struct {
...
 func NewDefaultWatcher(client kube.Client, revision string) DefaultWatcher {
 	p := &defaultWatcher{
-		revision:    revision,
-		queue:       workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
-		initialSync: atomic.NewBool(false),
-		mu:          sync.RWMutex{},
+		revision: revision,
+		mu:       sync.RWMutex{},
 	}
+	p.queue = controllers.NewQueue(controllers.WithName("default revision"), controllers.WithReconciler(p.setDefault))
```
Monkeyanator, line 62: How do we detect default webhook deletion now? Before if we got a delete event, we'd call `setDefault` with `""`, now it seems like we'd set the default to the last active default?

pkg/revisions/default_watcher.go
howardjohn: I would say `filter` is a fairly well known pattern in functional programing https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#filter-java.util.function.Predicate- 

But I will admit I confused myself when writing the code though so I mostly agree

pkg/revisions/default_watcher.go
howardjohn:
```
	revision := ""
	wh, _, _ := p.webhookInformer.GetIndexer().GetByKey(key.Name)
	if wh != nil {
		revision = wh.(metav1.Object).GetLabels()[label.IoIstioRev.Name]
	}
```

if its delete `wh` is nil and therefor revision is `""`

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,168 @@
+// Copyright Istio Authors
```
hzxuzhonghu, line 1: how about putting it under ` pkg/kube/queue`

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// Queue defines an abstraction around Kubernetes' workqueue.
+// Items enqueued are deduplicated; this generally means relying on ordering of events in the queue is not feasible.
+type Queue struct {
+	queue       workqueue.RateLimitingInterface
+	initialSync *atomic.Bool
+	name        string
+	maxAttempts int
+	work        func(key interface{}) error
```
hzxuzhonghu, line 35: s/work/worker

pkg/kube/controllers/queue.go
hzxuzhonghu: why not the args consistent?

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,168 @@
...
+// NewQueue creates a new queue
+func NewQueue(options ...func(*Queue)) Queue {
+	q := Queue{
+		initialSync: atomic.NewBool(false),
+	}
+	for _, o := range options {
+		o(&q)
+	}
+	if q.name == "" {
+		q.name = "queue"
```
hzxuzhonghu, line 77: nit: It seems meaningless

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,168 @@
...
+
+// Run the queue. This is synchronous, so should typically be called in a goroutine.
+func (q Queue) Run(stop <-chan struct{}) {
+	defer utilruntime.HandleCrash()
+	defer q.queue.ShutDown()
+	log.Infof("starting %v", q.name)
+	q.Add(defaultSyncSignal)
+	done := make(chan struct{})
+	go func() {
+		// Process updates until we return false, which indicates the queue is terminated
```
hzxuzhonghu, line 109: defer utilruntime.HandleCrash() should be here

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,168 @@
...
+		for q.processNextItem() {
+		}
+		close(done)
```
hzxuzhonghu, line 112: do we have cases to return?

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,168 @@
...
+	case <-done:
+	}
+	log.Infof("stopped %v", q.name)
+}
+
+// syncSignal defines a dummy signal that is enqueued when .Run() is called. This allows us to detect
+// when we have processed all items added to the queue prior to Run().
+type syncSignal struct{}
+
+var defaultSyncSignal = syncSignal{}
```
hzxuzhonghu, line 125: Still confused with the sync stuff just seeing the queue itself

pkg/kube/controllers/queue.go
howardjohn: it's for logging so we can distinguish queues

pkg/kube/controllers/queue.go
howardjohn: What is the confusion? I can help update the comments but wasn't sure. This is taking the sync logic we already have in a lot of queues

pkg/kube/controllers/queue.go
howardjohn: I think we need to in case the queue is closed. TBH this should probably never happen but we cannot

pkg/kube/controllers/queue.go
howardjohn: Hm, I am not sure. Controllers in kubernetes does not do this; they look like the code here

pkg/kube/controllers/queue.go
howardjohn: A bit worried about confusion with `pkg/queue`... maybe we can move it after we move everything over?

pkg/kube/controllers/queue.go
howardjohn: I don't know what you mean by args consistent?

hzxuzhonghu: interface{} vs types.NamespacedName

pkg/kube/controllers/queue.go
hzxuzhonghu: sgtm

pkg/kube/controllers/queue.go
hzxuzhonghu: If keep defer outside the go routine, it could not capture the crash users passed in and executed in this routine

pkg/kube/controllers/queue.go
hzxuzhonghu: It means to process all the buffering items enqueued before serving ready i think.

pilot/pkg/bootstrap/sidecarinjector.go
```diff
@@ -99,7 +99,7 @@ func (s *Server) initSidecarInjector(args *PilotArgs) (*inject.Webhook, error) {
 				return nil
 			}
 
-			patcher.Run(stop)
+			go patcher.Run(stop)
```
hzxuzhonghu, line 102: unrelated?

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -239,7 +239,7 @@ func (m *Multicluster) ClusterAdded(cluster *multicluster.Cluster, clusterStopCh
 			if err != nil {
 				log.Errorf("could not initialize webhook cert patcher: %v", err)
 			} else {
-				patcher.Run(clusterStopCh)
+				go patcher.Run(clusterStopCh)
```
hzxuzhonghu, line 242: ditto

pkg/kube/controllers/queue.go
howardjohn: NamespacedName is the external interface. Internally we pass sync signal as well so it's generic

pkg/kube/controllers/queue.go
howardjohn: Maybe we don't need it then.. every controller I looked at does it this way

pkg/kube/controllers/queue.go
howardjohn: Yep

pkg/revisions/default_watcher.go
```diff
@@ -53,35 +49,30 @@ type defaultWatcher struct {
...
 func NewDefaultWatcher(client kube.Client, revision string) DefaultWatcher {
 	p := &defaultWatcher{
-		revision:    revision,
-		queue:       workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
-		initialSync: atomic.NewBool(false),
-		mu:          sync.RWMutex{},
+		revision: revision,
+		mu:       sync.RWMutex{},
 	}
+	log.Errorf("howardjohn: default watch 1")
```
hzxuzhonghu, line 62: REMOVE

pilot/pkg/bootstrap/sidecarinjector.go
howardjohn: Just keeping consistent pattern for controllers to have Run block

stevenctl (approved):

hzxuzhonghu (approved):

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,173 @@
...
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	})
+}
+
+// Run the queue. This is synchronous, so should typically be called in a goroutine.
+func (q Queue) Run(stop <-chan struct{}) {
+	defer utilruntime.HandleCrash()
+	defer q.queue.ShutDown()
+	q.log.Infof("starting")
```
nmittler, line 109: This doesn't seem like a very useful message. Should we be able to name our queues? Or maybe log this from the call sites instead?

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,173 @@
...
+func (q Queue) AddObject(obj Object) {
+	q.queue.Add(types.NamespacedName{
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	})
+}
+
+// Run the queue. This is synchronous, so should typically be called in a goroutine.
+func (q Queue) Run(stop <-chan struct{}) {
+	defer utilruntime.HandleCrash()
```
nmittler, line 107: What does this do?

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,173 @@
...
+		// Process updates until we return false, which indicates the queue is terminated
+		for q.processNextItem() {
+		}
+		close(done)
+	}()
+	select {
+	case <-stop:
+	case <-done:
+	}
+	q.log.Infof("stopped")
```
nmittler, line 122: Same comment as above

pkg/kube/controllers/queue.go
howardjohn: `q.log` already embeds about the queues name and adds it to all messages

pkg/kube/controllers/queue.go
howardjohn: I think for our case effectively nothing. I will remove it

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,173 @@
...
+func (q Queue) processNextItem() bool {
+	// Wait until there is a new item in the working queue
+	key, quit := q.queue.Get()
+	if quit {
+		// We are done, signal to exit the queue
+		return false
+	}
+
+	// We got the sync signal. This is not a real event, so we exit early after signaling we are synced
+	if key == defaultSyncSignal {
```
nmittler, line 149: So theoretically, any empty struct would evaluate to true here, correct? I guess it wouldn't make sense for a user to add an empty struct, but doing so could lead to a false positive for syncing?

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,173 @@
...
+		q.log.Debugf("synced")
```
nmittler, line 150: Same comment ... starting to feel like we should just provide a name to all of the queues?

pkg/kube/controllers/queue.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	q.log.Debugf("handling update: %v", key)
+
+	// 'Done marks item as done processing' - should be called at the end of all processing
+	defer q.queue.Done(key)
+
+	err := q.workFn(key)
+	if err != nil {
+		if q.queue.NumRequeues(key) < q.maxAttempts {
```
nmittler, line 162: Do we know how many of our events are really retriable? If one attempt fails, will it ever succeed later?

pkg/kube/controllers/queue.go
howardjohn: I think it should be impossible for any user outside of the `controllers` package. The queue only exposes interfaces to add Object or NamespacedName, so they cannot pass it in. And even if they did somehow, `syncSignal` is private as well so cannot be constructed by anyone else

pkg/kube/controllers/queue.go
howardjohn: Also going to change it so name is required though

pkg/kube/controllers/queue.go
howardjohn: Some of them will (api server writes). I think a safer default is to not retry, and then opt into retries for those that makes sense

nmittler (approved):

PR #36320, opened 2021-12-01
Add timeout to the analysis

hanxiaop:
Currently executing `istioctl analyze` with the `--timeout` option does not take effect after https://github.com/istio/istio/pull/35764 got merged.

This PR is to add back the `--timeout` option.


monsteredc:
Add `release-notes-none` label since #35764 was merged after 1.12.

pkg/config/analysis/local/context_test.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"testing"
+	"time"
+
+	. "github.com/onsi/gomega"
+
+	"istio.io/istio/pkg/config/schema/collection"
+)
+
+func TestIstiodContextTimeout(t *testing.T) {
```
howardjohn, line 25: This test is flaky (`111 runs so far, 15 failures (88.10% pass rate). 1.949718985s avg, 3.644403612s max, 677.973719ms min`). using time.Sleep is not a great idea in tests

monsteredc:
Did not write the integration test since there is already one [TestTimeout](https://github.com/istio/istio/blob/769e03a499452f5d1f8ac47719acc409bd177d6e/tests/integration/pilot/analyze_test.go#L336). However, looks like it's been skipped so that the refactoring of analysis PR passed the test without the timeout. Will investigate the `TestTimeout` integration test later.

pkg/config/analysis/local/context_test.go
monsteredc: Fixed

monsteredc:
not stale

hanxiaop:
not stale

hanxiaop:
not stale

PR #36325, opened 2021-12-01
Istio install precheck only checks the min k8s version supported

johnma14:
When running the precheck command before installing Istio,  the `MinK8SVersion`
we are using to test is still the older 1.17. Istio 1.12 is officially supported
on Kubernetes versions 1.19 to 1.22.

Added the following changes:
- updated the `MinK8SVersion` from 1.17 to 1.19
- added a new testcase in the version unit test to check the kubernetes version

### DIFF
--- a/istioctl/pkg/install/k8sversion/version.go
+++ b/istioctl/pkg/install/k8sversion/version.go
@@ -28,7 +28,10 @@ import (
 const (
 	// MinK8SVersion is the minimum k8s version required to run this version of Istio
 	// https://istio.io/docs/setup/platform-setup/
-	MinK8SVersion = 17
+	MinK8SVersion               = 19
+	UnSupportedK8SVersionLogMsg = "\nThe Kubernetes version %s is not supported by Istio %s. The minimum supported Kubernetes version is 1.%d.\n" +
+		"Proceeding with the installation, but you might experience problems. " +
+		"See https://istio.io/latest/docs/setup/platform-setup/ for a list of supported versions.\n"
 )
 
 // CheckKubernetesVersion checks if this Istio version is supported in the k8s version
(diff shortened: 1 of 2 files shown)


istioctl/pkg/install/k8sversion/version_test.go
```diff
@@ -152,3 +157,36 @@ func TestExtractKubernetesVersion(t *testing.T) {
 		})
 	}
 }
+
+func TestCheckKubernetesVersion(t *testing.T) {
+	cases := []struct {
+		version  *version.Info
+		expected bool
+		errMsg   error
```
irisdingbj, line 165: nit: Can we remove errMsg as it is not needed at all?

istioctl/pkg/install/k8sversion/version.go
```diff
@@ -38,7 +39,7 @@ func CheckKubernetesVersion(versionInfo *version.Info) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	return MinK8SVersion <= v, nil
+	return (MinK8SVersion <= v && MaxK8SVersion >= v), nil
```
irisdingbj, line 42: Not sure we want to add maxK8sVersion check or not. @therealmitchconnors @esnible  any thoughts for this?

irisdingbj (commented):
Thanks for the fix! overall LGTM but need @therealmitchconnors  and @esnible thoughts about whether we need add the max version check.

linsun:
I am a bit against adding max k8s version check unless we know for sure the max version breaks the Istio version, it could be a new version that is perfectly fine most cases.  Most software only checks for min version anyway.

istioctl/pkg/install/k8sversion/version_test.go
johnma14: true, will do that. Thanks @irisdingbj

istioctl/pkg/install/k8sversion/version.go
ericvn: Problem might be that TOC has stated that we should try supporting the next k8s release after release if it's a short time. Example is 1.12 should be expanded to support 1.23 (discussion was for the RMs to have a long running qualification test done on the new k8s version when it comes out).  Certainly, the main branch should support 1.23, and would need to likely support 1.24 at it's release (likely on the istio 1.13 branch).

istioctl/pkg/install/k8sversion/version_test.go
```diff
@@ -152,3 +157,51 @@ func TestExtractKubernetesVersion(t *testing.T) {
 		})
 	}
 }
+
```
ericvn, line 160: I would add a comment in here (and maybe in the version.go) that IsK8VersionSupported only returns an error if it can't parse the version and only outputs a log if the version isn't supported (not sure if there is a way to verify if the text was output, but would be a good check). My first thought looking at this test is that the 1.19 checks should be different than the 1.17 checks (1.19 being good and 1.17 being bad), but the case data looks the same.

irisdingbj (approved):
Thanks for this!

PR #36350, opened 2021-12-02
operator: use shared kube client

howardjohn:
Currently we create a TON of clients, and call a ton of things
redundantly. This consolodates the code so we use a single shared client
(almost?) everywhere.

Before: 180 GET calls to api server for `istioctl install`
After: 40 GET calls

### DIFF
--- a/operator/cmd/mesh/shared.go
+++ b/operator/cmd/mesh/shared.go
@@ -24,10 +24,10 @@ import (
 	"time"
 
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"istio.io/istio/istioctl/pkg/install/k8sversion"
 	"istio.io/istio/operator/pkg/apis/istio/v1alpha1"
 	"istio.io/istio/operator/pkg/cache"
 	"istio.io/istio/operator/pkg/helmreconciler"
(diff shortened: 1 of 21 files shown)


hzxuzhonghu (approved):

irisdingbj (approved):
LGTM, Thanks!

shamsher31 (approved):

operator/cmd/mesh/shared.go
```diff
@@ -95,41 +95,26 @@ func confirm(msg string, writer io.Writer) bool {
 	return false
 }
 
-// K8sConfig creates a rest.Config, Clientset and controller runtime Client from the given kubeconfig path and context.
-func K8sConfig(kubeConfigPath string, context string) (*rest.Config, *kubernetes.Clientset, client.Client, error) {
-	restConfig, clientset, err := InitK8SRestClient(kubeConfigPath, context)
+func KubernetesClients(kubeConfigPath, context string, l clog.Logger) (kube.ExtendedClient, client.Client, error) {
+	rc, err := kube.DefaultRestConfig(kubeConfigPath, context, func(config *rest.Config) {
+		config.QPS = 200
```
jacob-delgado, line 100: Does the comment below still apply?

```golang
	// We are running a one-off command locally, so we don't need to worry too much about rate limitting
	// Bumping this up greatly decreases install time
```

operator/pkg/helmreconciler/reconciler.go
```diff
@@ -252,19 +237,14 @@ func (h *HelmReconciler) processRecursive(manifests name.ManifestMap) *v1alpha1.
...
-		// check k8s minor version
-		k8sVer, err := k8sversion.GetKubernetesVersion(h.restConfig)
-		if err != nil {
-			scope.Errorf("failed to get k8s version: %s", err)
-		}
+	if h.kubeClient != nil {
 		// There is a mutatingwebhook in gke that would corrupt the managedFields, which is fixed in k8s 1.18.
 		// See: https://github.com/kubernetes/kubernetes/issues/96351
-		if k8sVer >= 18 {
+		if kube.IsAtLeastVersion(h.kubeClient, 18) {
```
jacob-delgado, line 243: nit: We are going to only support 1.19+ for Istio 1.13, correct? This can be changed once that happens.

jacob-delgado (approved):
LGTM. Put a DNM since QPS/Burst was changed. If you don't see that being an issue you can remove it. Nice change.

operator/cmd/mesh/shared.go
howardjohn: Oops, copy+pasted from somewhere else. nice catch. Will revert numbers back and add the comment back

pkg/kube/client.go
```diff
@@ -506,10 +506,19 @@ func (c *client) RunAndWait(stop <-chan struct{}) {
...
+			c.version = v
+		}
 	})
-	return c.version, c.versionErr
+	if c.version != nil {
+		return c.version, nil
+	}
+	// Initial attempt failed, retry on each call to this function
+	v, err := c.Discovery().ServerVersion()
+	if err != nil {
```
able8, line 518: Is this a bug?  Should be `if err == nil {` ?

howardjohn: Good catch https://github.com/istio/istio/pull/41175

PR #36371, opened 2021-12-03
Specify a default metrics format

davidmortiz:
This fixes an issue where a blank Content-Type header would
be returned in cases where application metrics were disabled 
as would be the case pilot-agent is running as a gateway.

This addresses https://github.com/istio/istio/issues/3633


howardjohn (approved):

davidmortiz:
I’ve signed the CLA if someone wouldn’t mind triggering a re-scan

PR #36379, opened 2021-12-06
e2e test for opt-in label/annotation istio injection

deveshkandpal24121990:
- This PR adds additional test to validate opt-in istio injection.
-  We run in a shared k8s cluster, where not all services are istio injected so we depend on an opt-in model.
- Additionally, some service owners use annotations while newer services are using label based injection.
- These tests validate following scenario: opt-in/out label/annotation injection + no injection when label/annotation not present.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/kube/inject/inject.go
+++ b/pkg/kube/inject/inject.go
@@ -550,7 +550,7 @@ func parseTemplate(tmplStr string, funcMap map[string]interface{}, data SidecarT
 // IntoResourceFile injects the istio proxy into the specified
 // kubernetes YAML file.
 // nolint: lll
-func IntoResourceFile(injector Injector, sidecarTemplate Templates,
+func IntoResourceFile(injector Injector, sidecarTemplate *Config,
 	valuesConfig string, revision string, meshconfig *meshconfig.MeshConfig, in io.Reader, out io.Writer, warningHandler func(string)) error {
 	reader := yamlDecoder.NewYAMLReader(bufio.NewReaderSize(in, 4096))
 	for {
@@ -611,7 +611,7 @@ func FromRawToObject(raw []byte) (runtime.Object, error) {
 
 // IntoObject convert the incoming resources into Injected resources
 // nolint: lll
-func IntoObject(injector Injector, sidecarTemplate Templates, valuesConfig string,
+func IntoObject(injector Injector, sidecarTemplate *Config, valuesConfig string,
 	revision string, meshconfig *meshconfig.MeshConfig, in runtime.Object, warningHandler func(string)) (interface{}, error) {
 	out := in.DeepCopyObject()
 
@@ -738,7 +738,7 @@ func IntoObject(injector Injector, sidecarTemplate Templates, valuesConfig strin
 		}
 	}
 	if patchBytes == nil {
-		if !injectRequired(IgnoredNamespaces, &Config{Policy: InjectionPolicyEnabled}, &pod.Spec, pod.ObjectMeta) {
+		if !injectRequired(IgnoredNamespaces, &Config{Policy: InjectionPolicyEnabled, NeverInjectSelector: sidecarTemplate.NeverInjectSelector}, &pod.Spec, pod.ObjectMeta) {
 			warningHandler(fmt.Sprintf("===> Skipping injection because %q has sidecar injection disabled\n", name))
 			return out, nil
 		}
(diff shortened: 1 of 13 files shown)


ramaraochavali (approved):

pkg/kube/inject/inject.go
```diff
@@ -737,8 +738,12 @@ func IntoObject(injector Injector, sidecarTemplate Templates, valuesConfig strin
 			return nil, merr
 		}
 	}
+	var neverInjectSelector []metav1.LabelSelector
+	if sidecarConfig != nil {
+		neverInjectSelector = sidecarConfig.NeverInjectSelector
+	}
 	if patchBytes == nil {
-		if !injectRequired(IgnoredNamespaces, &Config{Policy: InjectionPolicyEnabled}, &pod.Spec, pod.ObjectMeta) {
+		if !injectRequired(IgnoredNamespaces, &Config{Policy: InjectionPolicyEnabled, NeverInjectSelector: neverInjectSelector}, &pod.Spec, pod.ObjectMeta) {
```
howardjohn, line 746: I am not sure why we are changing the main code when the PR is adding tests. Is this refactoring to make it more testable or fixing a bug?

It would be preferred to only touch test code.

howardjohn (changes requested):

pkg/kube/inject/inject.go
deveshkandpal24121990: @howardjohn - we need to make a change to make it more testable (let me update PR description). In my test I am supplying `NeverInjectSelector` from the operator spec but because this specific code path is executed from the test, where we create a new Config object with just  `InjectionPolicyEnabled` property set on it, its just getting ignored.

Having said that - Since we are supplying `Config` to `injectRequired` , I think it might be cleaner to pass the `Config` object as is instead of specifically passing `Config.Templates` from `IntoResources` and `IntoObject`. WDYT ?

deveshkandpal24121990:
@howardjohn - could you PTAL ? I've updated the PR description to provide clarity on why changes to `inject.go` is needed for supporting these tests.

pkg/kube/inject/testdata/inject/opt-in-injection.iop.yaml
```diff
@@ -0,0 +1,22 @@
...
+      - "{{ valueOrDefault .DeploymentMeta.Namespace \"default\" }}.global"
+    sidecarInjectorWebhook:
+      neverInjectSelector:
+      - matchExpressions:
+        - key: sidecar.istio.io/inject
+          operator: DoesNotExist
+      enableNamespacesByDefault: true
+      objectSelector:
+        enabled: true
+        autoInject: true
```
howardjohn, line 22: This config doesn't make any sense to me. I am fine adding a test of course, but I would strongly discourage anyone from doing this...

pkg/kube/inject/inject.go
```diff
@@ -761,8 +762,12 @@ func IntoObject(injector Injector, sidecarTemplate Templates, valuesConfig strin
 			return nil, merr
 		}
 	}
+	var neverInjectSelector []metav1.LabelSelector
+	if sidecarConfig != nil {
+		neverInjectSelector = sidecarConfig.NeverInjectSelector
+	}
 	if patchBytes == nil {
-		if !injectRequired(IgnoredNamespaces.UnsortedList(), &Config{Policy: InjectionPolicyEnabled}, &pod.Spec, pod.ObjectMeta) {
+		if !injectRequired(IgnoredNamespaces.UnsortedList(), &Config{Policy: InjectionPolicyEnabled, NeverInjectSelector: neverInjectSelector}, &pod.Spec, pod.ObjectMeta) {
```
howardjohn, line 770: why pass `NeverInjectSelector` but not `AlwaysInjectSelector`?

istioctl/cmd/kubeinject.go
```diff
@@ -535,7 +535,7 @@ It's best to do kube-inject when the resource is initially created.
 			retval := inject.IntoResourceFile(injector, sidecarTemplate, valuesConfig, rev, meshConfig,
 				reader, writer, func(warning string) {
 					warnings = append(warnings, warning)
-				})
+				}, nil)
```
howardjohn, line 538: If we only allow it for tests then, IMO we should make `IntoResourceFile` unchanged and make a new internal private `intoResourceFileWithConfig` that `IntoResourceFile` calls with `nil`. Otherwise we leak test detaIls into the API

Issue #32880, opened 2021-05-14
AWS platform/locality detection is not working when istio-proxy is running in non-root mode


alx13:
**Bug description**
If istio-proxy is running in non root mode on AWS EKS optimized AMI it's not detecting platform hence not providing locality information to envoy.

The code [here](https://github.com/istio/istio/blob/master/pkg/bootstrap/platform/aws.go#L120) is trying to read files from `/sys` filesystem, but it fails:

```
Container: istio-proxy
istio-proxy@pod-75d7bbf784-qvntg:/$ cat /sys/hypervisor/uuid
cat: /sys/hypervisor/uuid: No such file or directory
istio-proxy@pod-75d7bbf784-qvntg:/$ cat /sys/class/dmi/id/product_uuid
cat: /sys/class/dmi/id/product_uuid: Permission denied
istio-proxy@pod-75d7bbf784-qvntg:/$ ls -la /sys/class/dmi/id/product_uuid
-r-------- 1 root root 4096 May 13 05:08 /sys/class/dmi/id/product_uuid
istio-proxy@pod-75d7bbf784-qvntg:/$ cat /sys/class/dmi/id/product_version
istio-proxy@pod-75d7bbf784-qvntg:/$
```

[ ] Docs
[ ] Installation
[X ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**
platform and locality detection works with non-root privileges

**Version** (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version --short` if you used Helm)
- same code is used in istio 1.8, 1.9
- kubernetes 1.18 on EKS with AMI id `ami-07ab1fd3a20f0cfbf`

howardjohn:
cc @douglas-reid

douglas-reid:
Is there anyone with AWS expertise that can comment on a way to detect platform without root privileges?

alx13:
Currently I solved it with EC2 userdata and chmod a+r on

/sys/class/dmi/id/product_uuid

But feels a little bit dirty.

ramaraochavali:
@douglas-reid @howardjohn  WDYT about users specifying the platform via Env var or some other way and if specified can take precedence over auto detection?

ramaraochavali:
[...]

PR #36380, opened 2021-12-06
add env var for platform and fix aws

ramaraochavali:
This PR
- Adds env var support for specifying platform. So if user specifies a platform which he knows what platform is better, we do not have to discover and rely on user given value.
- Fixes AWS detection and treats permission error as AWS so that we can use metadata client rather than treating it as non AWS. This is needed for proxy running as  non root.

Fixes https://github.com/istio/istio/issues/32880

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/bootstrap/platform/aws.go
+++ b/pkg/bootstrap/platform/aws.go
@@ -15,6 +15,8 @@
 package platform
 
 import (
+	"errors"
+	"io/fs"
 	"os"
 	"strings"
 
@@ -22,6 +24,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	"istio.io/pkg/log"
 )
 
 const (
@@ -36,11 +39,14 @@ func IsAWS() bool {
 	if !systemInfoSuggestsAWS() {
 		// fail-fast for local cases
 		// WARN: this may lead to some cases of false negatives.
+		log.Debug("system info suggests this is not an AWS environment")
 		return false
 	}
 
 	if client := getEC2MetadataClient(); client != nil {
-		return client.Available()
+		available := client.Available()
+		log.Debugf("EC2Metadata client available :%v", available)
+		return available
 	}
 	return false
 }
@@ -118,16 +124,24 @@ func getEC2MetadataClient() *ec2metadata.EC2Metadata {
 // Note: avoided importing the satellite package directly to reduce number of
 // dependencies, etc., required.
 func systemInfoSuggestsAWS() bool {
-	hypervisorUUIDBytes, _ := os.ReadFile("/sys/hypervisor/uuid")
+	hypervisorUUIDBytes, uerr := os.ReadFile("/sys/hypervisor/uuid")
 	hypervisorUUID := strings.ToLower(string(hypervisorUUIDBytes))
 
-	productUUIDBytes, _ := os.ReadFile("/sys/class/dmi/id/product_uuid")
+	productUUIDBytes, perr := os.ReadFile("/sys/class/dmi/id/product_uuid")
 	productUUID := strings.ToLower(string(productUUIDBytes))
 
 	hasEC2Prefix := strings.HasPrefix(hypervisorUUID, "ec2") || strings.HasPrefix(productUUID, "ec2")
 
-	version, _ := os.ReadFile("/sys/class/dmi/id/product_version")
+	version, verr := os.ReadFile("/sys/class/dmi/id/product_version")
 	hasAmazonProductVersion := strings.Contains(string(version), "amazon")
 
-	return hasEC2Prefix || hasAmazonProductVersion
+	// If the error is a permission error, treat it as AWS as the files exist but user does not have
+	// permissions - we can try with EC2 metadata client instead of totally failing with false positive.
+	hasPermissionError := isPermissionError(uerr) || isPermissionError(perr) || isPermissionError(verr)
+
+	return hasPermissionError || hasEC2Prefix || hasAmazonProductVersion
+}
+
+func isPermissionError(err error) bool {
+	return !errors.Is(err, fs.ErrNotExist) && errors.Is(err, fs.ErrPermission)
 }
--- a/pkg/bootstrap/platform/aws.go
+++ b/pkg/bootstrap/platform/aws.go
@@ -46,7 +46,7 @@ func IsAWS() bool {
 
 	if client := getEC2MetadataClient(); client != nil {
 		available := client.Available()
-		log.Debugf("EC2Metadata client available :%v", available)
+		log.Debugf("EC2Metadata client available: %v", available)
 		return available
 	}
 	return false
--- a/pkg/bootstrap/platform/aws.go
+++ b/pkg/bootstrap/platform/aws.go
@@ -24,6 +24,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+
 	"istio.io/pkg/log"
 )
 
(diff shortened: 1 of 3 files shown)


ramaraochavali:
@douglas-reid @howardjohn PTAL

pkg/bootstrap/platform/aws.go
```diff
@@ -36,11 +40,14 @@ func IsAWS() bool {
...
 		// fail-fast for local cases
 		// WARN: this may lead to some cases of false negatives.
+		log.Debug("system info suggests this is not an AWS environment")
 		return false
 	}
 
 	if client := getEC2MetadataClient(); client != nil {
-		return client.Available()
+		available := client.Available()
+		log.Debugf("EC2Metadata client available :%v", available)
```
howardjohn, line 49:
```suggestion
		log.Debugf("EC2Metadata client available: %v", available)
```

pkg/bootstrap/platform/aws.go
```diff
@@ -118,16 +125,24 @@ func getEC2MetadataClient() *ec2metadata.EC2Metadata {
...
 	productUUID := strings.ToLower(string(productUUIDBytes))
 
 	hasEC2Prefix := strings.HasPrefix(hypervisorUUID, "ec2") || strings.HasPrefix(productUUID, "ec2")
 
-	version, _ := os.ReadFile("/sys/class/dmi/id/product_version")
+	version, verr := os.ReadFile("/sys/class/dmi/id/product_version")
 	hasAmazonProductVersion := strings.Contains(string(version), "amazon")
 
-	return hasEC2Prefix || hasAmazonProductVersion
+	// If the error is a permission error, treat it as AWS as the files exist but user does not have
```
howardjohn, line 139: These files exist everywhere?
```
$ docker run ubuntu cat /sys/class/dmi/id/product_uuid
9a8ecd0f-ec04-e911-9c8b-f4939fef9664
```

So this seems unsafe.

pkg/bootstrap/platform/discovery.go
```diff
@@ -17,16 +17,34 @@ package platform
...
+	"istio.io/pkg/log"
 )
 
 const (
 	defaultTimeout = 5 * time.Second
 	numPlatforms   = 3
 )
 
+var CloudPlatform = env.RegisterStringVar("CLOUD_PLATFORM", "", "Clound Platformn on which proxy is running, if not specified,"+
+	"Istio will try to discover the platform. Valid platform vaulues aws,azure,gcp.").Get()
```
howardjohn, line 31:
```suggestion
	"Istio will try to discover the platform. Valid platform values aws,azure,gcp.").Get()
```

pkg/bootstrap/platform/discovery.go
```diff
@@ -41,20 +59,23 @@ func DiscoverWithTimeout(timeout time.Duration) Environment {
 
 	go func() {
 		if IsGCP() {
+			log.Info("platform detected is Azure")
```
howardjohn, line 62:
```suggestion
			log.Info("platform detected is GCP")
```

pkg/bootstrap/platform/discovery.go
```diff
@@ -41,20 +59,23 @@ func DiscoverWithTimeout(timeout time.Duration) Environment {
...
 			plat <- NewGCP()
 		}
 		wg.Done()
 	}()
 
 	go func() {
 		if IsAWS() {
+			log.Info("platform detected is Aws")
```
howardjohn, line 70:
```suggestion
			log.Info("platform detected is AWS")
```

howardjohn (changes requested):
I think the /sys reading part is unsafe. The rest generally LGTM

pkg/bootstrap/platform/discovery.go
```diff
@@ -17,16 +17,34 @@ package platform
...
+	"istio.io/pkg/env"
+	"istio.io/pkg/log"
 )
 
 const (
 	defaultTimeout = 5 * time.Second
 	numPlatforms   = 3
 )
 
+var CloudPlatform = env.RegisterStringVar("CLOUD_PLATFORM", "", "Clound Platformn on which proxy is running, if not specified,"+
```
douglas-reid, line 30:
```suggestion
var CloudPlatform = env.RegisterStringVar("CLOUD_PLATFORM", "", "Cloud Platform on which proxy is running, if not specified,"+
```

pkg/bootstrap/platform/discovery.go
```diff
@@ -17,16 +17,34 @@ package platform
...
+	"Istio will try to discover the platform. Valid platform vaulues aws,azure,gcp.").Get()
+
 // Discover attempts to discover the host platform, defaulting to
 // `Unknown` if a platform cannot be discovered.
 func Discover() Environment {
+	// First check if user has specified platform - use it if provided.
+	if len(CloudPlatform) > 0 {
+		switch CloudPlatform {
```
douglas-reid, line 38: should we `toLower()` CloudPlatform to allow for `AWS|GCP|Azure` in addition to `aws|gcp|azure` ?

pkg/bootstrap/platform/aws.go
douglas-reid: Yeah, I think some of these are checked in the Azure (and GCP maybe?) checks as well.

ramaraochavali: Yes. These are checked already. 

@howardjohn Did not understand the "unsafe" part? Are you saying we should not read them at all? Or some problem with permission checking?

howardjohn: I mean this statement is basically always true. So we will always go down the AWS path even outside if AWS?

ramaraochavali:
@howardjohn @douglas-reid addressed all comments. PTAL.

pkg/bootstrap/platform/aws.go
ramaraochavali: If there are permissions errors  we just skip system info check but still go ahead and check with EC2 metadata https://github.com/istio/istio/blob/master/pkg/bootstrap/platform/aws.go#L42 which will fail if it is not AWS. Without this if there is a permission error on a real AWS, it just thinks it is not AWS.

howardjohn: I think the extra check is because if you have restrictive network policies the EC2 metadata check may hang forever (except we have a timeout, but its not very low). My memory here may be wrong though @douglas-reid ?

ramaraochavali: if EC2 metadata can not be accessed, even if it is detected as AWS based on sys files - we wont get locality info also ?

howardjohn: I mean if you call getEC2MetadataClient/client.Available off-AWS with restrictive network policies it may hang forever. So this may break non-AWS users

douglas-reid: This was an issue we had with off-GCP testing, iirc. Slow detection leading to long(er) proxy start times. The extra check was meant to be a rough cut short-circuit for off-platform users. If we don't have file perms, however, i'm not sure what choice we have. Should we have a `other` or `none` option for the env var that skips platform detection entirely to allow for quicker startups in those cases?

ramaraochavali: > The extra check was meant to be a rough cut short-circuit for off-platform users. If we don't have file perms, however, i'm not sure what choice we have.

Yes. Exactly. That is why I felt we could take a chance and call AWS api if file permissions are not there - then it would be slow but people would be fine if it detects correctly. And even if we have file permissions and it detects AWS, does not the call to metadata for locality would run in to same slowness?

> Should we have a other or none option for the env var that skips platform detection entirely to allow for quicker startups in those cases?

We can have and in this case where would it get locality from ? Is it from `istio-locality` label?

douglas-reid: RE: locality. `istio-locality` seems like the right place, I guess. that should work, right?

howardjohn: Did some brief testing and restrictive network policy already takes 5s before this PR anyways.

So IMO this is ok. We can/should add NONE to allow skipping. Its valid to be "none" - locality info is not detected in the proxy but can be grabbed from the nodes.

ramaraochavali:
@howardjohn @douglas-reid added "none". PTAL.

howardjohn (approved):

PR #36389, opened 2021-12-06
Revert "remove unnecessary sort"

howardjohn:
Reverts istio/istio#36382

This broke a bunch of tests. Some were obvious like https://github.com/istio/istio/pull/36388 but for the NDS I am not sure. May be more complex


stevenctl (approved):

ramaraochavali:
Surprised to see tests relying on this sort.

PR #36399, opened 2021-12-07
Make status and xds status have the consistent "NOT_SENT"

hanxiaop:
To change the proxy status `NOT SENT` to consistent with the envoy [NOT_SENT](https://github.com/envoyproxy/go-control-plane/blob/c2a905638ae8c2245137c7749df5663ae2a90e28/envoy/service/status/v3/csds.pb.go#L66) .
Which will make something like `istioctl ps` and `istioctl x ps` have the same format.


monsteredc:
not stale

PR #36400, opened 2021-12-07
[Optimize] reduce pilot memory in large degree

hzxuzhonghu:

pre processing request in receive thread as the main stream thread can be stuck when pushing xds. This is obvious when doing large scale performance test with pilot-load.

 We created total 15000 pods and 2400 services in 150 namespaces. And we have three pilot instances which means we make xds connections per pilot 5000.

The memory usage of pilot-discovery comparison is updated with latest code :

|  | Before | After |
| ------ | ------ | ------ |
| Peek | 13.96GB|  11.7 GB |
| Stable | 7.56 GiB | 7.36 GiB |

### DIFF
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -211,7 +211,7 @@ func (s *DiscoveryServer) receive(con *Connection, identities []string) {
 func deepCopyNode(node *core.Node) *core.Node {
 	out := &core.Node{}
 	buffer, _ := protomarshal.Marshal(node)
-	_ = protomarshal.Unmarshal(buffer, node)
+	_ = protomarshal.Unmarshal(buffer, out)
 	return out
 }
 
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -36,6 +36,7 @@ import (
 	v3 "istio.io/istio/pilot/pkg/xds/v3"
 	"istio.io/istio/pkg/cluster"
 	"istio.io/istio/pkg/config/schema/gvk"
+	"istio.io/istio/pkg/util/protomarshal"
 	"istio.io/pkg/env"
 	istiolog "istio.io/pkg/log"
 )
@@ -181,7 +182,8 @@ func (s *DiscoveryServer) receive(con *Connection, identities []string) {
 				con.errorChan <- status.New(codes.InvalidArgument, "missing node information").Err()
 				return
 			}
-			if err := s.initConnection(req.Node, con, identities); err != nil {
+			node := deepCopyNode(req.Node)
+			if err := s.initConnection(node, con, identities); err != nil {
 				con.errorChan <- err
 				return
 			}
@@ -204,6 +206,13 @@ func (s *DiscoveryServer) receive(con *Connection, identities []string) {
 			return
 		}
 	}
+}
+
+func deepCopyNode(node *core.Node) *core.Node {
+	out := &core.Node{}
+	buffer, _ := protomarshal.Marshal(node)
+	_ = protomarshal.Unmarshal(buffer, node)
+	return out
 }
 
 // StreamAggregatedResources implements the ADS interface.
@@ -403,8 +412,9 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 	// We should always respond with the current resource names.
 	if request.ResponseNonce == "" || previousInfo == nil {
 		log.Debugf("ADS:%s: INIT/RECONNECT %s %s %s", stype, con.ConID, request.VersionInfo, request.ResponseNonce)
+		resourceNames := deepCopyResourceNames(request.ResourceNames)
 		con.proxy.Lock()
-		con.proxy.WatchedResources[request.TypeUrl] = &model.WatchedResource{TypeUrl: request.TypeUrl, ResourceNames: request.ResourceNames}
+		con.proxy.WatchedResources[request.TypeUrl] = &model.WatchedResource{TypeUrl: request.TypeUrl, ResourceNames: resourceNames}
 		con.proxy.Unlock()
 		return true
 	}
@@ -427,7 +437,6 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 	previousResources := previousInfo.ResourceNames
 	previousInfo.NonceAcked = request.ResponseNonce
 	previousInfo.NonceNacked = ""
-	previousInfo.ResourceNames = request.ResourceNames
 
 	// Envoy can send two DiscoveryRequests with same version and nonce
 	// when it detects a new resource. We should respond if they change.
@@ -435,10 +444,19 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 		log.Debugf("ADS:%s: ACK %s %s %s", stype, con.ConID, request.VersionInfo, request.ResponseNonce)
 		return false
 	}
+
+	resourceNames := deepCopyResourceNames(request.ResourceNames)
+	previousInfo.ResourceNames = resourceNames
 	log.Debugf("ADS:%s: RESOURCE CHANGE previous resources: %v, new resources: %v %s %s %s", stype,
 		previousResources, request.ResourceNames, con.ConID, request.VersionInfo, request.ResponseNonce)
 
 	return true
+}
+
+func deepCopyResourceNames(in []string) []string {
+	resourceNames := make([]string, len(in))
+	copy(resourceNames, in)
+	return resourceNames
 }
 
 // shouldUnsubscribe checks if we should unsubscribe. This is done when Envoy is
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -100,7 +100,7 @@ type Connection struct {
 	stop chan struct{}
 
 	// reqChan is used to receive discovery requests for this connection.
-	reqChan      chan *discovery.DiscoveryRequest
+	reqChan      chan Request
 	deltaReqChan chan *discovery.DeltaDiscoveryRequest
 
 	// errorChan is used to process error during discovery request processing.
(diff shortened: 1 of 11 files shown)


pilot/pkg/xds/ads.go
```diff
@@ -893,6 +917,8 @@ func (conn *Connection) send(res *discovery.DiscoveryResponse) error {
 	sendHandler := func() error {
 		start := time.Now()
 		defer func() { recordSendTime(time.Since(start)) }()
+		conn.streamMutex.Lock()
+		defer conn.streamMutex.Unlock()
```
howardjohn, line 921: when does concurrent send happen? is it just for DebugType?

hzxuzhonghu: right

pilot/pkg/xds/ads.go
```diff
@@ -126,7 +138,7 @@ func newConnection(peerAddr string, stream DiscoveryStream) *Connection {
 		pushChannel:   make(chan *Event),
 		initialized:   make(chan struct{}),
 		stop:          make(chan struct{}),
-		reqChan:       make(chan *discovery.DiscoveryRequest, 1),
+		reqChan:       make(chan Request, 4), // 4 kinds of most frequently xds requests
```
ramaraochavali, line 141: The comment is not clear

pilot/pkg/xds/ads.go
```diff
@@ -180,69 +192,23 @@ func (s *DiscoveryServer) receive(con *Connection, identities []string) {
 			log.Infof("ADS: new connection for node:%s", con.ConID)
 		}
 
+		shouldResponde, blockedPushRequest := s.preProcessRequest(req, con)
```
ramaraochavali, line 195: shouldResponde -> shouldRespond

pilot/pkg/xds/ads.go
```diff
@@ -338,6 +306,62 @@ func (s *DiscoveryServer) Stream(stream DiscoveryStream) error {
...
+func (s *DiscoveryServer) postProcessRequest(req Request, push *model.PushContext, con *Connection) *model.PushRequest {
+	request := req.pushRequest
+	// this is client xds request
+	if request == nil {
+		request = &model.PushRequest{Full: true, Push: push}
+		request.Reason = append(request.Reason, model.ProxyRequest)
+		request.Start = time.Now()
+	}
+	// SidecarScope for the proxy may not have been updated based on this pushContext.
+	// It can happen when `processRequest` comes after push context has been updated(s.initPushContext),
```
ramaraochavali, line 357: nit: comment needs to be updated

hzxuzhonghu: good

howardjohn:
I am a bit confused by the performance numbers. How can this impact pilot_xds_push_time_bucket? we are still running the same push code...? Just less GC pressure?

Also confused how stable memory usage is reduced. If its stable there should be no requests so where is the memory differences?

pilot/pkg/xds/ads.go
howardjohn: TBH I am quite concerned by this. We have had quite a lot of success serializing the request/push such that we don't need mutexes, etc everywhere. I am concerned that allowing so much concurrency from various places will cause a lot of complexity, if not bugs.

It seems the main benefit here is that we do not hold onto `discovery.Request` object as long, which I guess must be around 200kb in your tests. It seems like this is because `pushChannel` is triggered, blocking `reqChan`, causing it to retain the memory long? Is that accurate? Some heap pprofs would be handy.

If the issue is the channel blocking maybe we can prioritize the reqChan. Go doesn't have a perfect solution for this but this should be good enough: https://stackoverflow.com/a/11121616/374797. There is a small race but should be strictly better than today. Can you try that out and compare?

hzxuzhonghu: We did the tests weeks ago, i could not distinguish which pprof belong to which case. 

But I have a simple report on the tests. Will send you

hzxuzhonghu: @howardjohn The lock here can be removed, as the debug type is only from istioctl, it doesnot req concurrently with other xds

pilot/pkg/xds/ads.go
```diff
@@ -126,7 +138,7 @@ func newConnection(peerAddr string, stream DiscoveryStream) *Connection {
 		pushChannel:   make(chan *Event),
 		initialized:   make(chan struct{}),
 		stop:          make(chan struct{}),
-		reqChan:       make(chan *discovery.DiscoveryRequest, 1),
+		reqChan:       make(chan Request, 4), // 4 kinds of most frequently requested xds[CDS,EDS,LDS,RDS]
```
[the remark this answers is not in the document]
hzxuzhonghu: The original thought is to make receive thread not blocked by reqChan. So that it will not refer the discoveryRequest.

hzxuzhonghu:
>How can this impact pilot_xds_push_time_bucket? 
 i guess it has something todo with the client receiver, 

```
	// SendMsg blocks until:
	//   - There is sufficient flow control to schedule m with the transport, or
	//   - The stream is done, or
	//   - The stream breaks.
```


>Also confused how stable memory usage is reduced. If its stable there should be no requests so where is the memory differences?

I am not sure either, it is a bonne surprise

hzxuzhonghu:
With 1.12 the result is different,  

|  | Before | After |
| ------ | ------ | ------ |
| Peek | 13.96GB|  11.7 GB |
| Stable | 7.56 GiB | 7.36 GiB |

howardjohn (commented):
How are you triggering the high memory? I cannot reproduce it by opening many clients with large DiscoveryRequests

howardjohn:
is it possible the `reqChan` is not garbage collected? and it has 1 request stuck in it.

I tested this - fill up reqChan buffer, then close it without reading. This is not sufficient to free the DiscoveryRequest if the `reqChan` is still in scope.

Test:

```go
	c := make(chan *discovery.DiscoveryRequest, 1)
	t.Run("many goroutine", func(t *testing.T) {
		var r1 func() bool
		var r2 func() bool
		go func() {
			req := &discovery.DiscoveryRequest{}
			//leak.MustGarbageCollect(t, req, "req")
			r1 = leak.CheckCollected(req, "req")
			req2 := &discovery.DiscoveryRequest{}
			r2 = leak.CheckCollected(req2, "req2")
			//leak.MustGarbageCollect(t, req2, "req2")
			t.Log(1)
			c <- req
			t.Log(2)
			c <- req2
			t.Log(3)
			time.Sleep(time.Second*10)
		}()
		time.Sleep(time.Millisecond * 100)
		t.Log("read")
		<-c
		time.Sleep(time.Millisecond * 100)
		t.Log("close", len(c))
		close(c)
		log.Errorf("howardjohn: wait for finalize")
		t.Log(r1())
		t.Log(r2())
		t.Log(len(c)) // Make sure c does not go out of scope
		for len(c) > 0 {
			t.Log("drain")
			<-c
		}
		t.Log(r1())
		t.Log(r2())
		t.Log(len(c)) // Make sure c does not go out of scope
	})
```

Result:
```
=== RUN   TestLeak/many_goroutine
    leaky_test.go:38: 1
    leaky_test.go:40: 2
    leaky_test.go:46: read
    leaky_test.go:42: 3
    leaky_test.go:49: close 1
2021-12-16T21:07:56.092021Z	error	ads	howardjohn: wait for finalize
2021-12-16T21:07:56.096015Z	error	howardjohn: finalized [req]
    leaky_test.go:52: true
    leaky_test.go:53: false
    leaky_test.go:54: 1
    leaky_test.go:56: drain
2021-12-16T21:07:56.101445Z	error	howardjohn: finalized [req2]
    leaky_test.go:59: true
    leaky_test.go:60: true   ### ONLY GC AFTER DRAIN
    leaky_test.go:61: 0
    --- PASS: TestLeak/many_goroutine (0.21s)
PASS

Process finished with the exit code 0
```

But here I am artificially keeping the channel in scope...

pilot/pkg/xds/ads.go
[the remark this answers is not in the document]
hzxuzhonghu: The buffer is only of some string not the entire DiscoveryRequest

hzxuzhonghu:
>How are you triggering the high memory? I cannot reproduce it by opening many clients with large DiscoveryRequests

I am running test with pilot-load forked and modeified a little ( to run fake pod in real single cluster), in this case we use this config

```
apiVersion: v1
data:
  config.yaml: |
    nodeMetadata: {}
    gracePeriod: 10s
    jitter:
      workloads: "24h"
      config: "24h"
    adsPercentage: 33  
    namespaces:
    - name: mesh
      replicas: 150
      applications:
      - name: big
        replicas: 1
        instances: 20
      - name: medium
        replicas: 5
        instances: 10
      - name: small
        replicas: 10
        instances: 3

```

hzxuzhonghu:
`adsPercentage` i remember is one i added, means percentage of clients that connects to istiod. We use this because we have only one istiod to save machine mem.

pilot/pkg/xds/ads.go
```diff
@@ -112,6 +113,15 @@ type Connection struct {
...
 
+// Request is the wrapper of requested xds type or blocked pushRequest
+type Request struct {
+	// pushRequest represents the blocked push request.
+	pushRequest *model.PushRequest
+
+	// resource type
+	typeURL string
+}
+
```
hzxuzhonghu, line 124: This is what i send to the reqChan, it has no reference to DiscoveryRequest any more. Let alone, i could not think of when we did not read the buffered channel.

hzxuzhonghu:
I guess you misunderstand the stable here: I mean there is no xds requests/responsed on the fly, but the xds connection is still there.

hzxuzhonghu:
Close as the resource names allocated by grpc recv can are shallow copied to `WatchedResource`, but golang profile reported it is still held by grpc lib.

Sp preprocess does not help this, not sure why it worked on 1.8 release.

The only way that can sloves this is maintain a cluster name mem pool, and for every xds client, we copy the value from this mem pool. In this way, all clients' WatchedResource shares common memory.

Suppose we have `10000` clusters, and each pilot has 5000 connections, and each cluster name is 50Byte like `outbound|8000||httpbin.default.svc.cluster.local`. The total memory cost is at least 10k * 5K * 50B = 2.5GB. As the go memory management is not so efficient, the real memory cost is far large.

Issue #36301, opened 2021-11-30
Istio remote TLS failure, SSLV3_ALERT_HANDSHAKE_FAILURE


Stono:
### Bug Description

We have a remote host that we're unable to proxy through Istio, in the istio-proy logs show:

```
{"level":"debug","time":"2021-11-30T14:01:32.672691Z","scope":"envoy connection","msg":"[C327111] connecting to 148.139.17.75:443"}
{"level":"debug","time":"2021-11-30T14:01:32.672919Z","scope":"envoy connection","msg":"[C327111] connection in progress"}
{"level":"debug","time":"2021-11-30T14:01:32.690127Z","scope":"envoy connection","msg":"[C327111] connected"}
{"level":"debug","time":"2021-11-30T14:01:32.706355Z","scope":"envoy connection","msg":"[C327111] TLS error: 268436496:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE 268435610:SSL routines:OPENSSL_internal:HANDSHAKE_FAILURE_ON_CLIENT_HELLO"}
{"level":"debug","time":"2021-11-30T14:01:32.706390Z","scope":"envoy connection","msg":"[C327111] closing socket: 0"}
{"level":"debug","time":"2021-11-30T14:01:32.706420Z","scope":"envoy connection","msg":"[C327111] TLS error: 268436496:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE 268435610:SSL routines:OPENSSL_internal:HANDSHAKE_FAILURE_ON_CLIENT_HELLO"}
{"level":"debug","time":"2021-11-30T14:01:32.706450Z","scope":"envoy client","msg":"[C327111] disconnect. resetting 0 pending requests"}
{"level":"debug","time":"2021-11-30T14:01:32.706461Z","scope":"envoy pool","msg":"[C327111] client disconnected, failure reason: TLS error: 268436496:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE 268435610:SSL routines:OPENSSL_internal:HANDSHAKE_FAILURE_ON_CLIENT_HE
LLO"}
[...]

PR #36401, opened 2021-12-07
add default ecdh curves to upstream tls context

ramaraochavali:
Fixes https://github.com/istio/istio/issues/36301

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -61,6 +61,9 @@ var istioMtlsTransportSocketMatch = &structpb.Struct{
 	},
 }
 
+// SupportedEcdhCurves for upstream TLS configuration.
+var SupportedEcdhCurves = []string{"X25519", "P-256", "secp384r1"}
+
 // h2UpgradeMap specifies the truth table when upgrade takes place.
 var h2UpgradeMap = map[upgradeTuple]bool{
 	{meshconfig.MeshConfig_DO_NOT_UPGRADE, networking.ConnectionPoolSettings_HTTPSettings_UPGRADE}:        true,
@@ -1134,6 +1137,11 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 		if cb.IsHttp2Cluster(c) {
 			// This is HTTP/2 cluster, advertise it with ALPN.
 			tlsContext.CommonTlsContext.AlpnProtocols = util.ALPNH2Only
+		}
+	}
+	if tlsContext != nil && tlsContext.CommonTlsContext != nil {
+		tlsContext.CommonTlsContext.TlsParams = &auth.TlsParameters{
+			EcdhCurves: SupportedEcdhCurves,
 		}
 	}
 	return tlsContext, nil
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
@@ -1802,6 +1802,7 @@ func TestBuildUpstreamClusterTLSContext(t *testing.T) {
 			result: expectedResult{
 				tlsContext: &tls.UpstreamTlsContext{
 					CommonTlsContext: &tls.CommonTlsContext{
+						TlsParams: &tls.TlsParameters{EcdhCurves: SupportedEcdhCurves},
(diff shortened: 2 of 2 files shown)


pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -61,6 +61,9 @@ var istioMtlsTransportSocketMatch = &structpb.Struct{
 	},
 }
 
+// SupportedEcdhCurves for upstream TLS configuration.
+var SupportedEcdhCurves = []string{"X25519", "P-256", "secp384r1"}
```
ramaraochavali, line 65: first two are Envoy default list. Added the third one.  We do not add default cipher suites/ecdh curves for upstream tls context today - we rely on Envoy defaults.

Wondering is this the right thing to do or should we atleast make this configurable ?
@howardjohn @hzxuzhonghu  WDYT?

hzxuzhonghu: I am not an expert on this. Curious: there are many other types of curves, why only add these three?

ramaraochavali: first two are current defaults in Envoy and the third one is needed for the issue mentioned.

hzxuzhonghu: I know this, my question is why not the others

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -61,6 +61,9 @@ var istioMtlsTransportSocketMatch = &structpb.Struct{
 	},
 }
 
+// SupportedEcdhCurves for upstream TLS configuration.
```
howardjohn, line 64: I think this may break users compiling with FIPS mode in envoy - it will likely reject these. I can test it.

Have we looked into getting envoy to increase the default? That will make our config size smaller, trivially add the check for ^. And if they ah no maybe they give us a good reason that we shouldn't?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1136,6 +1139,11 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 			tlsContext.CommonTlsContext.AlpnProtocols = util.ALPNH2Only
 		}
 	}
+	if tlsContext != nil && tlsContext.CommonTlsContext != nil {
+		tlsContext.CommonTlsContext.TlsParams = &auth.TlsParameters{
+			EcdhCurves: SupportedEcdhCurves,
```
howardjohn, line 1144: probably don't need it for istio mutual

Stono:
hey @howardjohn @ramaraochavali - any update on this?  don't want it to get lost in the ether :)

hzxuzhonghu:
Need to clarify why these three are the default `{"X25519", "P-256", "secp384r1"}`

Stono:
@hzxuzhonghu `X25519` and `P-256` are the current defaults in envoy and `secp384r1` is what we'd like adding.  One risk here would be that you'd be making these ciphers and explicit istio configuration rather than pulling in the default envoy config (eg if they change that in the future, you'd drift).  

If there are concerns about that, another option would be to make this more easily configurable (eg a first class api for what i'm doing here: https://github.com/istio/istio/issues/36301#issuecomment-983413360)

without it, there are some host we can't connect to with istio.

ramaraochavali:
My preference would be to add a first class API in DestinationRule ClientTlsSettings similar to how we do for gateway cipher suites https://github.com/istio/api/blob/master/networking/v1alpha3/gateway.proto#L732 and only apply that logic for  MUTUAL as ISTIO_MUTUAL does not need these settings.

@howardjohn @hzxuzhonghu WDYT? If there are no objections., I can make that API change.

howardjohn:
I do have a PR to add configurable cipher suites to DR: https://github.com/istio/api/pull/2014. But there is some concern if it should be an enum (DEFAULT|COMPATIBILITY|SECURE).

Also GW has cipher suites but ecdh curves would need to be a separate setting, right?

ramaraochavali:
> Also GW has cipher suites but ecdh curves would need to be a separate setting, right?

Yes

PR #36404, opened 2021-12-07
fix filter and virtual host patch removal

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu:
can you add some tests, since i can see the previous remove handling seems ok

```
	if networkFiltersRemoved {
		tempArray := make([]*xdslistener.Filter, 0, len(fc.Filters))
		for _, filter := range fc.Filters {
			if filter.Name != "" {
				tempArray = append(tempArray, filter)
			}
		}
		fc.Filters = tempArray
	}
```

ramaraochavali:
I will add. But previous removing logic is not fine, it is considering the last filter status only

ramaraochavali:
Will add tests then we can review

howardjohn (commented):
seems good but can we have a test?

howardjohn (commented):
VHOST remove is broken as well:
```
kind: EnvoyFilter
metadata:
  name: connect-match
  namespace: default
spec:
  configPatches:
  - applyTo: VIRTUAL_HOST
    match:
      context: SIDECAR_OUTBOUND
      routeConfiguration:
        vhost:
          name: allow_any
    patch:
      operation: REMOVE
```

ramaraochavali:
Listener patch removal seems to be working fine. So just added a test

ramaraochavali:
@howardjohn VHOST remove is broken as well - It may be because we are catching the catchAllVirtualHost? Or are you thinking it is because of some other problem?

howardjohn:
No I had the same issue I think? With empty name. But maybe _also_ related to vhost cache. But if we remove it should not be a problem for the cache/

ramaraochavali:
Ok. Let me try that Envoy filter and see what is going on.

pilot/pkg/networking/core/v1alpha3/envoyfilter/listener_patch.go
```diff
@@ -524,7 +528,7 @@ func patchHTTPFilters(patchContext networking.EnvoyFilter_PatchContext,
 func patchHTTPFilter(patchContext networking.EnvoyFilter_PatchContext,
 	patches map[networking.EnvoyFilter_ApplyTo][]*model.EnvoyFilterConfigPatchWrapper,
 	listener *xdslistener.Listener, fc *xdslistener.FilterChain, filter *xdslistener.Filter,
-	httpFilter *hcm.HttpFilter, httpFilterRemoved *bool) {
+	httpFilter *hcm.HttpFilter) bool {
```
hzxuzhonghu, line 531: Better document the meaning of return value

ramaraochavali:
I was able to reproduce virtual host removal issue in our environment. But not able to do it via unit test. It works correctly. I will continue to spend some time to reproduce it via unit test.

ramaraochavali:
@howardjohn @hzxuzhonghu Fixed the issue for vhost and listeners. Added test for vhost - with out this change, the last test case in `TestApplyRouteConfigurationPatches` will have virtual host with empty name.

howardjohn (approved):

pilot/pkg/networking/core/v1alpha3/envoyfilter/listener_patch.go
```diff
@@ -506,11 +506,11 @@ func patchHTTPFilters(patchContext networking.EnvoyFilter_PatchContext,
 		}
 		IncrementEnvoyFilterMetric(lp.Key(), HttpFilter, applied)
 	}
-	if httpFiltersRemoved {
+	if len(removedFilters) > 0 {
 		tempArray := make([]*hcm.HttpFilter, 0, len(httpconn.HttpFilters))
```
hzxuzhonghu, line 510: nit: len(httpconn.HttpFilters)-len(removedFilters)

hzxuzhonghu (approved):

PR #36416, opened 2021-12-07
Remove arbitraryclient to have a single consistent client

howardjohn:
In the recent change from galley -> pilot code for the analysis, we added a new `arbitraryclient`. This is a config store that implements the `ConfigStore` interface, as required by the analysis code, but uses a different code base than the crdclient (and uses the dynamic client, which is less efficient than generated code).

This leads to quite a bit of code duplication, as well as some performance penalty.

This PR dedupes these clients. The old `crdclient` (probably a poor name now) can handle all Kubernetes resources (that we care about, anyways), by gaining an understanding of what is a "CRD" and what is a native type. There is a bit of boilerplate here, but with Go 1.18 generics eventually it will go away IMO.


hzxuzhonghu:
What does this pr do

howardjohn:
Updated  the description

hzxuzhonghu:
nit: maybe we can rename crdClient to dynamicClient to align with k8s

pilot/pkg/config/file/store.go
```diff
@@ -436,10 +435,47 @@ func ToConfig(object metav1.Object, schema collection.Schema, source resource.Re
...
 	}
-	result := arbitraryclient.TranslateObject(u, "", schema)
+	result := TranslateObject(u, "", schema)
 	return result, nil
 }
 
+func TranslateObject(obj *unstructured.Unstructured, domainSuffix string, schema collection.Schema) *config.Config {
+	mv2, err := schema.Resource().NewInstance()
+	if err != nil {
+		panic(err)
```
hzxuzhonghu, line 445: return error

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -174,10 +174,19 @@ func NewForSchemas(ctx context.Context, client kube.Client, revision, domainSuff
 	for _, s := range schemas.All() {
 		// From the spec: "Its name MUST be in the format <.spec.name>.<.spec.group>."
 		name := fmt.Sprintf("%s.%s", s.Resource().Plural(), s.Resource().Group())
-		if _, f := known[name]; f {
+		crd := true
+		switch s.Resource().Group() {
+		case gvk.Pod.Group, gvk.Deployment.Group, gvk.MutatingWebhookConfiguration.Group, gvk.CustomResourceDefinition.Group:
```
hzxuzhonghu, line 179: This is the hack point

hzxuzhonghu: Don't we have other k8s native type? Like Node, ValidatingWebhookConfiguration?

I wonder we can cache all the k8s api resources, and if one resource kind  is in,  it is native.

howardjohn:
Ack, will make it more robust.

howardjohn (commented):
@hzxuzhonghu hacks are removed now

hzxuzhonghu (approved):

PR #36422, opened 2021-12-07
resolve gateway hostnames in control plane

### LINKED ISSUE #29359: Multi-Cluster/Multi-Network - Cannot use a hostname-based gateway for east-west traffic
**Bug description**

Following the guide [Install Multi-Primary on different networks](https://istio.io/latest/docs/setup/install/multicluster/multi-primary_multi-network), everything seems to install as expected without errors and is running in the cluster.  For `secret/cacerts` I am using the example certificate material from [samples/certs/*.pem](https://github.com/istio/istio/tree/release-1.8/samples/certs) in both cluster1 and cluster2.

When I attempt to verify the installation using the guide [Verify the installation](https://istio.io/latest/docs/setup/install/multicluster/verify) the requests are not getting routed to the remote cluster as expected.  I am only getting responses from the service on the local cluster:
```bash
# From Cluster 1 [where helloworld v1 is deployed]
$ while true; do kubectl exec --context="${CTX_CLUSTER1}" -n sample -c sleep     "$(kubectl get pod --context="${CTX_CLUSTER1}" -n sample -l \
    app=sleep -o jsonpath='{.items[0].metadata.name}')"     -- curl -s helloworld.sample:5000/hello; done
Hello version: v1, instance: helloworld-v1-578dd69f69-r9lkz
Hello version: v1, instance: helloworld-v1-578dd69f69-r9lkz
Hello version: v1, instance: helloworld-v1-578dd69f69-r9lkz
Hello version: v1, instance: helloworld-v1-578dd69f69-r9lkz
Hello version: v1, instance: helloworld-v1-578dd69f69-r9lkz
...
# From Cluster 2 [where helloworld v2 is deployed]
$ while true; do kubectl exec --context="${CTX_CLUSTER2}" -n sample -c sleep     "$(kubectl get pod --context="${CTX_CLUSTER2}" -n sample -l \
    app=sleep -o jsonpath='{.items[0].metadata.name}')"     -- curl -s helloworld.sample:5000/hello; done
Hello version: v2, instance: helloworld-v2-776f74c475-h5j2q
Hello version: v2, instance: helloworld-v2-776f74c475-h5j2q
Hello version: v2, instance: helloworld-v2-776f74c475-h5j2q
Hello version: v2, instance: helloworld-v2-776f74c475-h5j2q
Hello version: v2, instance: helloworld-v2-776f74c475-h5j2q
...
```

`istioctl proxy-config endpoint` for the `sleep` pod in cluster1 and cluster2 to `helloworld` destination service:
```bash
# Cluster 1
$ istioctl -n sample --context=${CTX_CLUSTER1} proxy-config endpoint "$(kubectl get pod --context="${CTX_CLUSTER1}" -n sample -l app=sleep -o jsonpath='{.items[0].metadata.name}')" | grep helloworld
10.100.1.12:5000                 HEALTHY     OK                outbound|5000||helloworld.sample.svc.cluster.local

# Cluster 2
$ istioctl -n sample --context=${CTX_CLUSTER2} proxy-config endpoint "$(kubectl get pod --context="${CTX_CLUSTER2}" -n sample -l app=sleep -o jsonpath='{.items[0].metadata.name}')" | grep helloworld
10.100.2.188:5000                HEALTHY     OK                outbound|5000||helloworld.sample.svc.cluster.local
```

It seems like maybe something is missing from the docs or example configs, but I understand that there are tests for the docs/examples, which is why I’ve been troubleshooting my own cluster… but just seems like something small is missing

[X] Docs
[X] Installation
[X] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**
I expected to be able to follow the guide and get the same behavior that the guide expects

Following the guides for [multi-primary, multi-network install](https://istio.io/latest/docs/setup/install/multicluster/multi-primary_multi-network) and [verify the install](https://istio.io/latest/docs/setup/install/multicluster/verify)

```bash
$ istioctl version
client version: 1.8.0
control plane version: 1.8.0
data plane version: 1.8.0 (4 proxies)

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-12T01:09:16Z", GoVersion:"go1.15.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.12-eks-7684af", GitCommit:"7684af4ac41370dd109ac13817023cb8063e3d45", GitTreeState:"clean", BuildDate:"2020-10-20T22:57:40Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

**How was Istio installed?**
Istio Operator installed with `istioctl operator init` and the rest of the installation of istio in `istio-system` is done as steps in the guide [using example manifests & scripts to compile example manifests]

AWS EKS with Kubernetes v1.17
Istio 1.8.0 on Mac
```bash
$ uname -a
Darwin OAK-MAC-HLJHD2 19.6.0 Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64 x86_6
```

stevenctl:
This seems like the best we can do for now... 

1. Strict DNS clusters would mean we do a ton of CDS updates as endpoints change
2. Aggregate cluster between EDS and Strict DNS only fails-over, doesn't actually LB
3. In-agent solutions seem even more complex. We need envoy to contain IPs in its config... 

fixes #29359

---

1. Move network manager to `Environment`
2. When gateways change, trigger a callback to make NetworkManager reload all the gateways
3. When we're aggregating the gateways, we check if it's a hostname and try to resolve it
4. TODO if the hostname TTL expires, we re-trigger the networkmanager reload
5. If gateways changed during a reload, trigger a push

---

Manually tested (CTX1 is EKS, CTX2 is GKE)

```
$ k --context $CTX1 get svc -n istio-system istio-eastwestgateway
NAME                    TYPE           CLUSTER-IP      EXTERNAL-IP                                                               PORT(S)                                                           AGE
istio-eastwestgateway   LoadBalancer   10.100.215.77   a777f08b1a517490b9b6175166717390-1490576810.us-east-2.elb.amazonaws.com   15021:32093/TCP,15443:32497/TCP,15012:31981/TCP,15017:32072/TCP   34m
 $ dig +nocmd +noall +answer a777f08b1a517490b9b6175166717390-1490576810.us-east-2.elb.amazonaws.com.
a777f08b1a517490b9b6175166717390-1490576810.us-east-2.elb.amazonaws.com. 60 IN A 3.17.125.95
a777f08b1a517490b9b6175166717390-1490576810.us-east-2.elb.amazonaws.com. 60 IN A 18.116.184.220
a777f08b1a517490b9b6175166717390-1490576810.us-east-2.elb.amazonaws.com. 60 IN A 3.13.34.113
$ stioctl --context $CTX2 x internal-debug networkz | jq
[
  {
    "Network": "network1",
    "Cluster": "cluster1",
    "Addr": "18.116.184.220",
    "Port": 15443
  },
  {
    "Network": "network1",
    "Cluster": "cluster1",
    "Addr": "3.13.34.113",
    "Port": 15443
  },
  {
    "Network": "network1",
    "Cluster": "cluster1",
    "Addr": "3.17.125.95",
    "Port": 15443
  },
  {
    "Network": "network2",
    "Cluster": "cluster2",
    "Addr": "34.72.86.18",
    "Port": 15443
  }
]
```


pilot/pkg/features/pilot.go
```diff
@@ -547,6 +547,9 @@ var (
 	MulticlusterHeadlessEnabled = env.RegisterBoolVar("ENABLE_MULTICLUSTER_HEADLESS", true,
 		"If true, the DNS name table for a headless service will resolve to same-network endpoints in any cluster.").Get()
 
+	HostnameGateways = env.RegisterBoolVar("ALLOW_HOSTNAME_GATEWAYS", false,
```
howardjohn, line 550: should we default true?

pilot/pkg/model/network_gw_names.go
```diff
@@ -0,0 +1,128 @@
...
+	*dns.Client
+	resolvConfServers []string
+}
+
+func newClient() (*dnsClient, error) {
+	dnsConfig, err := dns.ClientConfigFromFile("/etc/resolv.conf")
+	if err != nil {
+		return nil, err
+	}
+	c := &dnsClient{
```
howardjohn, line 96: Any reason to use miekg vs `net.Resolver.LookupIP`?

howardjohn: I think the former is much lower level

howardjohn: Ah but then we don't get the TTL. https://github.com/golang/go/issues/24796.

https://github.com/rs/dnscache does add this, but its still not using TTL so doesn't seem ideal

Seems like there ought to be a better way but I don't see any

pilot/pkg/model/network_gw_names.go
```diff
@@ -0,0 +1,128 @@
...
+	if entry, ok := n.cache[name]; ok && entry.expiry.After(time.Now()) {
+		return entry.value
+	}
+	return n.resolveAndCache(name)
+}
+
+func (n *networkGatewayNameCache) resolveAndCache(name string) []string {
+	delete(n.cache, name)
+	names, ttl := n.resolve(name)
+	// don't cache for very small TTL or empty list of names
```
howardjohn, line 44: For very small TTL, maybe we should just set a minimum bar we cap at. Like never resolve more than 1x/minute?

I don't think any cloud is expecting that all clients are re-resolving DNS every single query and will break if they don't - if they change IP I would hope they would have both IPs working for some period of time. Just a guess - i do not know for sure.

pilot/pkg/model/network_gw_names.go
```diff
@@ -0,0 +1,128 @@
...
+	if len(names) == 0 || ttl < 10*time.Second {
+		return names
+	}
+	expiry := time.Now().Add(ttl)
+	time.AfterFunc(ttl, func() {
+		log.Infof("triggering update due to gateway hostname TTL expiry")
```
howardjohn, line 50: should have more context on which one probably?

Can we only do it if the response is different as well?

howardjohn: Also we should probably normalize the response order. Some DNS's will rotate them to get dumb clients to load balance; the order does not matter for envoy and will only cause update churn

pilot/pkg/model/context.go
```diff
@@ -160,6 +155,11 @@ func (e *Environment) Init() {
 	e.clusterLocalServices = NewClusterLocalProvider(e)
 }
 
+func (e *Environment) InitGatewayNameCache(updater XDSUpdater) (err error) {
+	e.gatewayNameCache, err = newNetworkGatewayNameCache(updater)
```
howardjohn, line 159: I am a bit concerned with this synchronous flow

I think a better model may be to have something async.

* When a GW is added with hostname, we add it to 'dnsCache'. It is not effective until dnsCache has resolved.
* For all items in dnsCache, we keep them up to date in background, re-querying every once and while (influenced by TTL). If the result changes, we trigger a push.
  * One concern is the DNS response is constantly changing as a poor-mans load balancer. Hopefully no one is doing this but we may need to add some check for this...

In the current model I am a bit concerned with slow DNS updates blocking an XDS push. Also if there are no pushes for a long time, we will go long past the TTL without re-resolving

WDYT?

howardjohn: Actually I see the time.AfterFunc. So my concern about not re-resolving is not valid.

IMO its better to have more DNS resolves than more pushes. The latter is a lot more expensive.

Do we know the (default) TTL of AWS LB? I am guessing its pretty low

stevenctl: This sounds good, but means I'll need to refactor a few things first since we aggregate all the gateways during push-context init.

howardjohn: One middle ground would be to only do synchronous lookup on the first add. This should happen rarely. Then future updates are async.

I don't think this is the best solution, but its likely way easier. And its opt in - currently anyone with Hostname is broken entirely. This doesn't negatively impact users not using hostname (right?)

stevenctl: > This doesn't negatively impact users not using hostname (right?)

Right. 

>  only do synchronous lookup on the first add

We should probably refactor this to be notify-up anyway... something like: 

```go
env.ServiceDiscovery.AddNetworkGatewaysHandler(networkManager.onGatewaysChanged())
```

stevenctl:
The last commit added a lot of complexity and made TestMeshNetworking take 5s per-testcase.

pilot/pkg/model/context.go
stevenctl: > only do synchronous lookup on the first add

I had to change a ton of code to avoid this.. worth it?

pilot/pkg/features/pilot.go
nrjpoddar: I think so.

stevenctl:
I think it would be a lot of work, but I can give it a shot.

pilot/pkg/model/network.go
```diff
@@ -108,40 +155,101 @@ func NewNetworkManager(env *Environment) *NetworkManager {
...
+	hostnameGateways := map[string][]NetworkGateway{}
+	names := sets.NewSet()
+	for gw := range gatewaySet {
+		if gwIP := net.ParseIP(gw.Addr); gwIP != nil {
+			continue
+		}
+		delete(gatewaySet, gw)
+		if !features.HostnameGateways {
+			log.Warnf("Failed parsing gateway address %s from Service Registry. "+
+				"Hostnames are not supported for gateways",
```
linsun, line 176: We should teach user how to enable this in the log if we don't feel confident to enable this support in day 1.

pilot/pkg/model/network.go
```diff
@@ -108,40 +155,101 @@ func NewNetworkManager(env *Environment) *NetworkManager {
...
+		}
+		hostnameGateways[gw.Addr] = append(hostnameGateways[gw.Addr], gw)
+		names.Insert(gw.Addr)
+	}
+
+	// resolve each hostname
+	for host, addrs := range mgr.nameCache.Resolve(names) {
+		gwsForHost := hostnameGateways[host]
+		if len(addrs) == 0 {
+			log.Warnf("could not resolve hostname %q for %d gateways", host, len(gwsForHost))
```
linsun, line 188: What is the implication of this?  data plane traffic through that gw won't function?

linsun: or no traffic would be sent to that gw?

stevenctl: Right, we won't have an endpoint for that gateway so we drop it. This likely means that the control plane can't hit whichever nameservers actually resolve the hostname.

pilot/pkg/model/network.go
stevenctl: Good point – although I'm thinking we can just turn the feature flag on by default.

stevenctl:
Last things to do:

1. Write some tests covering the DNS part
2. Possibly break this into some smaller PRs

cc @nmittler @howardjohn - lmk if you'd rather see this as 1 PR for the NetworkManager callbacks refactor and 1 for the DNS stuff

stevenctl:
hold: seems like bench fails due to an import cycle

pilot/pkg/model/network_gw_names.go
stevenctl: We sort the gateways later, no need to sort here.

pilot/pkg/model/network.go
```diff
@@ -202,3 +326,239 @@ func gcd(x, y int) int {
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// resolve gets all the A and AAAA records for the given name
+func (n *networkGatewayNameCache) resolve(name string) ([]string, time.Duration) {
+	// TODO figure out how to query only A + AAAA
```
howardjohn, line 485: I think you need to do two queries.

pilot/pkg/model/network.go
```diff
@@ -108,40 +159,103 @@ func NewNetworkManager(env *Environment) *NetworkManager {
...
+
+	// resolve each hostname
+	for host, addrs := range mgr.NameCache.Resolve(names) {
+		gwsForHost := hostnameGateways[host]
+		if len(addrs) == 0 {
+			log.Warnf("could not resolve hostname %q for %d gateways", host, len(gwsForHost))
+		}
+		// expand each resolved address into a NetworkGateway
+		for _, gw := range gwsForHost {
+			for _, resolved := range addrs {
```
howardjohn, line 196: A bit confused by this. So if there are multiple addrs, we just use the last one? Especially concerned with this WRT to the A/AAAA queries - its plausible the LB is available under both but the cluster is only v4/v6 so one or the other will fail?

pilot/pkg/model/network.go
howardjohn: Various go programs likely do this if we want to peek at their code. Coredns, tailscale,

pilot/pkg/model/network.go
howardjohn: Also ideally long term we can have multiple IPs for a single gw and adjust the weight proportionally? But certainly not a P0

pilot/pkg/model/network.go
```diff
@@ -202,3 +326,239 @@ func gcd(x, y int) int {
...
+
+func (c *dnsClient) Query(req *dns.Msg) *dns.Msg {
+	var response *dns.Msg
+	for _, upstream := range c.resolvConfServers {
+		cResponse, _, err := c.Exchange(req, upstream)
+		if err == nil {
+			response = cResponse
+			break
+		} else {
+			log.Infof("upstream dns failure: %v", err)
```
howardjohn, line 555: warnf maybe?

howardjohn (approved):
LGTM, certainly an improvement from today - we can fix issues with the resolution in followups if the pop up

pilot/pkg/model/network.go
stevenctl: This expands into a gateway instance per address in the answer. The weighting logic is already handled in our EDS code.

howardjohn: I am not sure I understand. For each iteration of `addrs`, we set `resolvedGw.Addr = resolved`. Won't that just overwrite each time and leave only the last `addrs`?

nrjpoddar:
@stevenctl thanks for getting this merged, do you plan to cherry-pick this to 1.12?

PR #36429, opened 2021-12-08
Derive proxy image from values.global if it exist

deveshkandpal24121990:
- `imageName` is hardcoded to "proxyv2" as introduced in this PR -> https://github.com/istio/istio/pull/35996
- This PR fixes the bug and derives `imageName` from `values.global.proxy.image` if it exist
- Added unit test to verify the same

### DIFF
--- a/pkg/kube/inject/inject.go
+++ b/pkg/kube/inject/inject.go
@@ -305,6 +305,10 @@ func ProxyImage(values *opconfig.Values, image *proxyConfig.ProxyImage, annotati
 		imageType = image.ImageType
 	}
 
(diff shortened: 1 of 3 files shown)


ramaraochavali (commented):
LGTM. @mandarjog @howardjohn can you PTAL?

mandarjog:
we had left out image name ( and proxy init image name) because it has not been used in a long time.
Do we have  evidence that this is still used ?

ramaraochavali:
Yes. We use it

mandarjog (approved):
Lgtm

howardjohn (approved):
ty!

PR #36436, opened 2021-12-08
Run update_deps; migrate to use crane

jacob-delgado:
Run bin/update_deps.sh
Use crane now that it is part of the build image

### DIFF
--- a/bin/update_deps.sh
+++ b/bin/update_deps.sh
@@ -47,8 +47,10 @@ go mod tidy
 sed -i "s/^BUILDER_SHA=.*\$/BUILDER_SHA=$(getSha release-builder)/" prow/release-commit.sh
 sed -i '/PROXY_REPO_SHA/,/lastStableSHA/ { s/"lastStableSHA":.*/"lastStableSHA": "'"$(getSha proxy)"'"/  }' istio.deps
 
-LATEST_DEB11_DISTROLESS_SHA256=$(docker image pull gcr.io/distroless/static-debian11 | grep Digest | awk -F\\: '{print $3}')
+# shellcheck disable=SC1001
+LATEST_DEB11_DISTROLESS_SHA256=$(crane digest gcr.io/distroless/static-debian11 | awk -F\: '{print $2}')
 sed -i -E "s/sha256:[a-z0-9]+/sha256:${LATEST_DEB11_DISTROLESS_SHA256}/g" docker/Dockerfile.distroless
 
-LATEST_IPTABLES_DISTROLESS_SHA256=$(docker image pull gcr.io/istio-release/iptables | grep Digest | awk -F\\: '{print $3}')
+# shellcheck disable=SC1001
+LATEST_IPTABLES_DISTROLESS_SHA256=$(crane digest gcr.io/istio-release/iptables | awk -F\: '{print $2}')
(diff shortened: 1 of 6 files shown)


jacob-delgado:
cc @howardjohn Looks like https://github.com/istio/pkg/pull/438 got rid of the `probe` module and https://github.com/istio/istio/blob/0d240a70072f5e365f4fc24194db1fdfb87dadb1/security/pkg/pki/ca/ca.go#L34 depends on it.

howardjohn:
> cc @howardjohn Looks like [istio/pkg#438](https://github.com/istio/pkg/pull/438) got rid of the `probe` module and
> 
> https://github.com/istio/istio/blob/0d240a70072f5e365f4fc24194db1fdfb87dadb1/security/pkg/pki/ca/ca.go#L34
> 
> depends on it.

It was intended, the dependency is from dead code. I thought I merge a removal already, sorry. Want to fix it here, or I can in another PR?

Basically remove all usage of `probe`, its just a struct field but not used.

Sorry, forgot about this.

jacob-delgado:
> 

I can do it here. Just wanted to make sure it was intended.

bin/update_deps.sh
```diff
@@ -47,8 +47,8 @@ go mod tidy
 sed -i "s/^BUILDER_SHA=.*\$/BUILDER_SHA=$(getSha release-builder)/" prow/release-commit.sh
 sed -i '/PROXY_REPO_SHA/,/lastStableSHA/ { s/"lastStableSHA":.*/"lastStableSHA": "'"$(getSha proxy)"'"/  }' istio.deps
 
-LATEST_DEB11_DISTROLESS_SHA256=$(docker image pull gcr.io/distroless/static-debian11 | grep Digest | awk -F\\: '{print $3}')
+LATEST_DEB11_DISTROLESS_SHA256=$(crane digest gcr.io/distroless/static-debian11 | grep Digest | awk -F\\: '{print $3}')
```
howardjohn, line 50: I think the grep part is now wrong?

howardjohn (approved):

howardjohn:
hold for the crane issue

bin/update_deps.sh
ericvn: Agree that on my Mac the `Digest` is now missing.

crane digest gcr.io/distroless/static-debian11 | awk -F\\: '{print $2}' seems to be correct.

bin/update_deps.sh
```diff
@@ -47,8 +47,8 @@ go mod tidy
...
 sed -i -E "s/sha256:[a-z0-9]+/sha256:${LATEST_DEB11_DISTROLESS_SHA256}/g" docker/Dockerfile.distroless
 
-LATEST_IPTABLES_DISTROLESS_SHA256=$(docker image pull gcr.io/istio-release/iptables | grep Digest | awk -F\\: '{print $3}')
+LATEST_IPTABLES_DISTROLESS_SHA256=$(crane digest gcr.io/istio-release/iptables | grep Digest | awk -F\\: '{print $3}')
```
ericvn, line 53: Same as above.

bin/update_deps.sh
jacob-delgado: Removed the grep and updated the awk command to get rid of the warning.

bin/update_deps.sh
jacob-delgado: Fixed

PR #36446, opened 2021-12-09
only send eds for updated services 

hzxuzhonghu:

Service、endpoint are the most frequently updated resources that trigger xds push.

We only need to push eds for the updated clusters, here we calc it based on configsUpdated only contains ServiceEntry.

### DIFF
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -409,15 +409,21 @@ func (eds *EdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
 
 	cached := 0
 	regenerated := 0
-	for _, clusterName := range w.ResourceNames {
-		if edsUpdatedServices != nil {
+
+	var filter []string
+	if edsUpdatedServices != nil {
+		for _, clusterName := range w.ResourceNames {
 			_, _, hostname, _ := model.ParseSubsetKey(clusterName)
-			if _, ok := edsUpdatedServices[string(hostname)]; !ok {
-				// Cluster was not updated, skip recomputing. This happens when we get an incremental update for a
-				// specific Hostname. On connect or for full push edsUpdatedServices will be empty.
-				continue
+			if _, ok := edsUpdatedServices[string(hostname)]; ok {
+				filter = append(filter, clusterName)
 			}
 		}
+		log.Infof("======eds cluster filter: %v", filter)
+	} else {
+		filter = w.ResourceNames
+	}
+
+	for _, clusterName := range filter {
 		builder := NewEndpointBuilder(clusterName, proxy, push)
 		if marshalledEndpoint, f := eds.Server.Cache.Get(builder); f && !features.EnableUnsafeAssertions {
 			// We skip cache if assertions are enabled, so that the cache will assert our eviction logic is correct
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -381,7 +381,7 @@ func edsNeedsPush(updates model.XdsUpdates) bool {
 	return false
 }
 
-func allConfigsOfKind(configs map[model.ConfigKey]struct{}, kind string) bool {
+func allConfigsWithKind(configs map[model.ConfigKey]struct{}, kind string) bool {
 	for k := range configs {
 		if kind != k.Kind.Kind {
 			return false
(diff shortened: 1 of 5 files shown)


operator/cmd/mesh/install.go
```diff
@@ -247,7 +247,13 @@ func InstallManifests(iop *v1alpha12.IstioOperator, force bool, dryRun bool, kub
 		return iop, fmt.Errorf("errors occurred during operation: %v", err)
 	}
 	if status.Status != v1alpha1.InstallStatus_HEALTHY {
-		return iop, fmt.Errorf("errors occurred during operation")
+		additionalInfo := ""
```
ramaraochavali, line 250: This seems unrelated to this PR

pilot/pkg/xds/eds.go
```diff
@@ -379,28 +379,35 @@ func edsNeedsPush(updates model.XdsUpdates) bool {
...
+
 func (eds *EdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *model.WatchedResource,
 	req *model.PushRequest) (model.Resources, model.XdsLogDetails, error) {
 	if !edsNeedsPush(req.ConfigsUpdated) {
 		return nil, model.DefaultXdsLogDetails, nil
 	}
 	var edsUpdatedServices map[string]struct{}
-	if !req.Full {
+	if !req.Full ||
+		// all configs updated are ServiceEntry type, then we only need to send eds for these updated services.
```
ramaraochavali, line 398: When will we need this check - it seems if it is not a full push we are only sending eds?

operator/cmd/mesh/install.go
hzxuzhonghu: Yes, i used it to know why the install failed

pilot/pkg/xds/eds.go
hzxuzhonghu: I want to calculate which eds need to push when se updated.  Envoy does not re-warm clusters that are not changed, so we donot need to push eds for them too?

ramaraochavali: > I want to calculate which eds need to push when se updated

Where are you doing that? edsUpdatedServices has list of all ses?

hzxuzhonghu: Get all the `updated services` and  

ignore those eds for unchanged clusters: w.ResourceNames - `updated services`

ramaraochavali: sorry. my question was not clear - it did not change with this PR though, we were already doing that, wondering what is the use of this new condition? 

Are you saying this will be used if req is Full and ConfigUpdated has only serviceEntry update even then you want to send eds only?

Not sure if that correct behaviour because if service is changed should n't we send updated cluster?

hzxuzhonghu: This is working on full push

hzxuzhonghu: If svc is changed, it will be added to `edsUpdatedServices`

and i meant to gen eds for it

pilot/pkg/xds/eds.go
```diff
@@ -387,7 +396,12 @@ func (eds *EdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
 	var edsUpdatedServices map[string]struct{}
 	if !req.Full {
 		edsUpdatedServices = model.ConfigNamesOfKind(req.ConfigsUpdated, gvk.ServiceEntry)
+	} else if len(req.ConfigsUpdated) > 0 && allConfigsWithKind(req.ConfigsUpdated, gvk.ServiceEntry.Kind) {
+		// when full push
+		// if all configs updated are ServiceEntry type, then we only need to send eds for these updated services.
+		edsUpdatedServices = model.ConfigNamesOfKind(req.ConfigsUpdated, gvk.ServiceEntry)
```
howardjohn, line 402: What about multinetwork Gateways? is that a concern?

hzxuzhonghu: Should be no. Gateway acts the same as sidecars.

hzxuzhonghu:
After some investigation, the failure seems random. And I filed an issue https://github.com/envoyproxy/envoy/issues/19262

hzxuzhonghu:
This is currently not right given to envoy bug https://github.com/envoyproxy/envoy/issues/19262

Issue #36448, opened 2021-12-09
fatal error: concurrent map iteration and map write in /debug/registryz


maxbischoff:
### Bug Description

Since upgrading to istio 1.12.0 our istiod instances crash regularly with following error:

```
fatal error: concurrent map iteration and map write

goroutine 13767 [running]:
runtime.throw({0x2e0daff, 0x462999})
	runtime/panic.go:1198 +0x71 fp=0xc021c32ee8 sp=0xc021c32eb8 pc=0x435891
runtime.mapiternext(0x17)
	runtime/map.go:858 +0x4eb fp=0xc021c32f58 sp=0xc021c32ee8 pc=0x40f90b
reflect.mapiternext(0x195)
	runtime/map.go:1346 +0x19 fp=0xc021c32f70 sp=0xc021c32f58 pc=0x4628d9
reflect.(*MapIter).Next(0xc01805d420)
	reflect/value.go:1618 +0x99 fp=0xc021c32f90 sp=0xc021c32f70 pc=0x4d4399
encoding/json.mapEncoder.encode({0x0}, 0xc00adcc100, {0x28462e0, 0xc00947cfc8, 0xc021c33208}, {0xc, 0x0})
	encoding/json/encode.go:799 +0x305 fp=0xc021c33150 sp=0xc021c32f90 pc=0x74aec5
encoding/json.mapEncoder.encode-fm(0x29c0380, {0x28462e0, 0xc00947cfc8, 0x27}, {0xf8, 0x31})
	encoding/json/encode.go:779 +0x45 fp=0xc021c33190 sp=0xc021c33150 pc=0x7556e5
encoding/json.structEncoder.encode({{{0xc009e2de60, 0x2f06798, 0x2974500}, 0xc01d43b680}}, 0xc00adcc100, {0x29c0380, 0xc00947cfc8, 0xe}, {0x0, 0x1})
	encoding/json/encode.go:761 +0x1f4 fp=0xc021c33240 sp=0xc021c33190 pc=0x74a914
encoding/json.structEncoder.encode-fm(0x29c0380, {0x29c0380, 0xc00947cfc8, 0xf3}, {0xc0, 0xc0})
	encoding/json/encode.go:732 +0x69 fp=0xc021c33298 sp=0xc021c33240 pc=0x755629
[...]

PR #36449, opened 2021-12-09
> fix service list concurrent read and write

kebe7jun:
Fix https://github.com/istio/istio/issues/36448

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -874,7 +874,7 @@ func (c *Controller) Services() ([]*model.Service, error) {
 	c.RLock()
 	out := make([]*model.Service, 0, len(c.servicesMap))
 	for _, svc := range c.servicesMap {
-		out = append(out, svc)
+		out = append(out, svc.DeepCopy())
 	}
 	c.RUnlock()
 	sort.Slice(out, func(i, j int) bool { return out[i].Hostname < out[j].Hostname })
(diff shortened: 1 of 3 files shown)


howardjohn (changes requested):
Do we need to deepcopy in the prod code? That will be costly. Maybe just in registryz you can copy it only

kebe7jun:
> Do we need to deepcopy in the prod code? That will be costly. Maybe just in registryz you can copy it only

DeepCopy does lose performance, but without DeepCopy, concurrent reads and writes to objects do cause problems, and we can't just do DeepCopy inside `registryz`, because it's not fundamentally different from calling writeJSON directly, they all go to read the contents inside the object..

PS: ServiceEntry  https://github.com/istio/istio/blob/9f449d325e325d4b0e3f281586ca988c404d273f/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go#L511-L514
also do DeepCopy, if we don't use DeepCopy, we may have to make some bigger changes to possibly solve this problem.

howardjohn:
Where are we writing to the map after creation?

kebe7jun:
Service.Attributes.Labels ExportTo LabelSelectors ClusterExternalPorts.

https://github.com/istio/istio/blob/9f449d325e325d4b0e3f281586ca988c404d273f/pilot/pkg/model/service.go#L508-L541

kebe7jun:
Good question, I'll try to find it out later.

zijianz:
@kebe7jun Does it mean that at this time, `registryz` can not be called in an intensive manner? Thanks!

rcernich:
I think the issue here is that we're simply serializing these objects, but the objects themselves are being modified in another thread.  I suspect the code that is doing the modification is most likely holding a lock (e.g. the kube registry), but the code doing the reading (json serializer) is not.  I think the solution would be to lock the object before serializing it to json (`Service` has a publicly accessible mutex).

PR #36450, opened 2021-12-09
Fix bug-report label and anno match

monsteredc:
From the bug-report help command:
```
istioctl bug-report -h
bug-report selectively captures cluster information and logs into an archive to help diagnose problems.
Proxy logs can be filtered using:
  --include|--exclude ns1,ns2.../dep1,dep2.../pod1,pod2.../cntr1,cntr.../lbl1=val1,lbl2=val2.../ann1=val1,ann2=val2...
where ns=namespace, dep=deployment, cntr=container, lbl=label, ann=annotation

The filter spec is interpreted as 'must be in (ns1 OR ns2) AND (dep1 OR dep2) AND (cntr1 OR cntr2)...'
```

Labels and annotations should be matched with `OR` condition. If the pod has one of the label satisfied in the `--include`, and no label matched in the `--exclude`, it should not be skipped.

### DIFF
--- a/tools/bug-report/pkg/cluster/cluster.go
+++ b/tools/bug-report/pkg/cluster/cluster.go
@@ -124,16 +124,13 @@ func shouldSkip(deployment string, config *config2.BugReportConfig, pod *corev1.
 			}
 
 			if len(ild.Labels) > 0 {
-				isLabelsMatch := true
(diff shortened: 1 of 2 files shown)


jacob-delgado (approved):

jacob-delgado:
Verified this as working

PR #36452, opened 2021-12-09
Add wildcard pattern match for bug-report --include and --exclude

monsteredc:
From the bug report help command:
```
The filter spec is interpreted as 'must be in (ns1 OR ns2) AND (dep1 OR dep2) AND (cntr1 OR cntr2)...'
The log will be included only if the container matches at least one include filter and does not match any exclude filters.
All parts of the filter are optional and can be omitted e.g. ns1//pod1 filters only for namespace ns1 and pod1.
All names except label and annotation keys support '*' glob matching pattern.

e.g.
--include ns1,ns2 (only namespaces ns1 and ns2)
--include n*//p*/l=v* (pods with name beginning with 'p' in namespaces beginning with 'n' and having label 'l' with value beginning with 'v'.)
```

The include and exclude parts should also support wildcard match instead of only using string contains.


jacob-delgado:
cc @monsteredc can you please rebase this and ping me when you are done?

monsteredc:
@jacob-delgado PTAL. Thanks for the review!

jacob-delgado:
I've been on PTO. I'll do this on Tuesday @monsteredc

monsteredc:
@jacob-delgado Alright thanks, just rebased.

jacob-delgado (approved):
verified it works

PR #36454, opened 2021-12-09
fix(dashboards): workloads dashboard should display namespaces with only sent bytes

timcosta:
Currently if you are running workloads that read and write from an external destination (for example reading and writing from kafka topics on MSK) the namespace those workloads run in will not be displayed in the namespaces dropdown. This PR includes namespaces that only have `istio_tcp_received_bytes_total` metrics and not just `istio_tcp_sent_bytes_total` or `istio_requests_total`.

I've also included a change that causes the `Reporter` dropdown to select both `source` and `destination` by default rather than the current behavior of only `destination` which I personally think makes more sense as it causes charts in use cases like I mentioned above to show data by default rather than be empty.

### DIFF
--- a/manifests/addons/dashboards/istio-workload-dashboard.json
+++ b/manifests/addons/dashboards/istio-workload-dashboard.json
@@ -2513,11 +2513,11 @@
         "multi": false,
         "name": "namespace",
         "options": [],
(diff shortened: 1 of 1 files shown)


timcosta:
CLA has been signed.

PR #36458, opened 2021-12-10
Drop import of github.com/stretchr/testify 

howardjohn:
Trim dependency set a bit. Instead, provide our own 20 line library that is Istio opinionated and:
* Supports protobuf properly
* Takes in our `test.Failer` type

Also refactored some helps that use `t` in a non-standard position.

### DIFF
--- a/cni/pkg/install/install_test.go
+++ b/cni/pkg/install/install_test.go
@@ -23,7 +23,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
+	"istio.io/istio/pkg/test/util/assert"
 
 	"istio.io/istio/cni/pkg/config"
 	testutils "istio.io/istio/pilot/test/util"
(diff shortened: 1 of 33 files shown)


nmittler (approved):

stevenctl (approved):

liamawhite (approved):

cni/pkg/install/install_test.go
```diff
@@ -221,7 +220,7 @@ func TestSleepCheckInstall(t *testing.T) {
 
 			select {
 			case <-readyChan:
-				assert.Truef(t, isReady.Load().(bool), "isReady should have been set to true")
+				assert.Equal(t, isReady.Load(), true)
```
lei-tang, line 223: Nit: why remove the message "isReady should have been set to true"?

howardjohn: This message doesn't have any value. Equal will already emit that its 'false' if the test case fails and the line that fails.

lei-tang (approved):

Issue #35177, opened 2021-09-14
Get error message in kubernetes IPv6 only cluster when executing command `istioctl proxy-config route`


zhlsunshine:
### Bug Description

I get error message when executing istioctl proxy-config route in my cluster which is kubernetes IPv6 only.
And the detail as below:

```
$ istioctl proxy-config route httpbin-74fb669cc6-bpm7b --name 80 -ojson
2021-09-14T06:19:58.748624Z	error	klog	an error occurred forwarding 45017 -> 15000: error forwarding port 15000 to pod df101dc6c65effd3b0904bf8d5a9b3cc124163cefac68106ce348e1247ebd647, uid : exit status 1: 2021/09/14 14:19:58 socat[27452] E connect(5, AF=2 127.0.0.1:15000, 16): Connection refused
Error: failed to execute command on httpbin-74fb669cc6-bpm7b.default sidecar: failure running port forward process: Get "http://127.0.0.1:45017/config_dump": EOF
```

### Version

```prose
$ istioctl version
client version: 1.11.1
control plane version: ce6205d503e5c5e41af496ebbe01ece7dc6c3547
data plane version: 1.11.1 (4 proxies)
$ kubectl version --short
Client Version: v1.22.1
Server Version: v1.22.1
```

### Additional Information

Found it when working on [issue#](https://github.com/istio/istio/issues/34966)

### Affected product area

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [X] User Experience
- [ ] Developer Infrastructure
- [ ] Upgrade
- [ ] Multi Cluster
- [ ] Virtual Machine
- [ ] Control Plane Revisions

### Is this the right place to submit this?

- [X] This is not a security vulnerability
- [X] This is not a question about how to use Istio

pseudonator:
@zhlsunshine Il just throw my findings here as per https://github.com/istio/istio/issues/34358.
Also I do notice that your port-forward is using `socat`. Its possible that your runtime doesn't have the fix I talk about.

zhlsunshine:
Hi @pseudonator, I did some investigation again for this issue, as you know, Istio implement its own port-forwarder function, and it's very similar with `kubectl`, and it's okay for `kubectl port-forward`, but not works for `Istio`. So this makes me more confuse.

zhlsunshine:
[...]

Issue #34358, opened 2021-07-27
istioctl commands fail in an IPv6 only cluster


pseudonator:
**Bug description**
In an IPv6 only cluster if `istioctl pc` is executed then the following error will pop-up
```
2021-07-27T12:15:12.330900Z	error	klog	an error occurred forwarding 52054 -> 15000: error forwarding port 15000 to pod 516bcfbe033ea20d6db0151ccc7e073c25661a3569ff6671677e2a3c30219883, uid : failed to execute portforward in network namespace "/var/run/netns/cni-e8dd1f78-40d9-f83c-c61c-740add119d6a": failed to dial 15000: dial tcp4 127.0.0.1:15000: connect: connection refused
Error: failed to execute command on istio-ingressgateway-7f66dbb97c-6m62m.rcp-istio-system sidecar: failure running port forward process: Get "http://127.0.0.1:52054/config_dump": EOF
```

This is because admin port `15000` is bound to `::1`
```
tcp6       0      0 ::1:15000               :::*                    LISTEN      10/envoy
```

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**
`istioctl` to work regardless of the underlying cluster running on IPv4, IPv6 or dualstack.

**Steps to reproduce the bug**

**Version**
client version: 1.10.2
control plane version: 1.10.2
data plane version: 1.10.2 (3 proxies)

**Environment where the bug was observed (cloud vendor, OS, etc)**
In an IPv6 cluster.

esnible:
I don't have an IPv6 only cluster to troubleshoot or test this issue.

pseudonator:
@esnible Il see if i can dig into it further and create a PR

zhlsunshine:
Hi @pseudonator,  I have a pure IPv6 k8s cluster, and I also can reproduce this error easily, so could I take this issue if you are not working on it?

pseudonator:
@esnible @zhlsunshine Investigating this further it seems that this is a generic issue relating to IPv6 and CRI.
[...]

PR #36461, opened 2021-12-10
add notification for istioctl with IPv6 only cluster that port-forward is broken for specified container runtime

zhlsunshine:
Add notification for IPv6 only cluster that it will break port-forward if container runtime is Docker or container runtime is containerd and its version < 1.5.
please refer to [here](https://docs.google.com/document/d/1a6eZ6ldF2l7De3vX-SBAHmuJ00uxlgpPznHocG7zcfs/edit?usp=sharing) for more detail.
Related issues: #35177 and #34358

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


zhlsunshine:
Output as below:

```
istioctl pc route httpbin-74fb669cc6-g7jhk
2021-12-10T07:45:46.161949Z	error	klog	an error occurred forwarding 40379 -> 15000: error forwarding port 15000 to pod e1e8b65195a6a8f4ea4bf7eb429290c1a062ad2143977fa46318c98f7318012d, uid : exit status 1: 2021/12/10 15:45:46 socat[23447] E connect(5, AF=2 127.0.0.1:15000, 16): Connection refused
Error: failed to execute command on httpbin-74fb669cc6-g7jhk.default sidecar: failure running port forward process: Get "http://127.0.0.1:40379/config_dump": EOF

Notification: the cluster is IPv6 only cluster and it will break port-forward if container runtime is Docker or container runtime is containerd and its version < 1.5
```

pkg/kube/client.go
```diff
@@ -695,7 +696,22 @@ func (c *client) EnvoyDo(ctx context.Context, podName, podNamespace, method, pat
 
 func (c *client) portForwardRequest(ctx context.Context, podName, podNamespace, method, path string, port int) ([]byte, error) {
 	formatError := func(err error) error {
-		return fmt.Errorf("failure running port forward process: %v", err)
+		pods, _ := c.GetIstioPods(context.TODO(), podNamespace, map[string]string{
```
howardjohn, line 699: Can we just parse the error message or something? Doing a lookup on pods just to return a better error message seems like a high cost

zhlsunshine: sure, that's okay

zhlsunshine: Hi @howardjohn,  I have checked that
Firstly,  there may be different output for 3 cases:
1. Docker as a container runtime
2. containerd as a container runtime but version is less than 1.50
3. containerd as a container runtime but version is equal or great than 1.50
It's easy to identify case 1 which use **socat**, but parse the error message may be not enough to distinguish 2 and 3 which just tell user the connection refuse. 
Secondly, this lookup on pods could be happened only when error is occurring, not every time.
Thirdly, it may be not reliable to depend on the error output from other, such as container runtime or kubelet.
So can we keep the current logic for this?

howardjohn: I feel very uncomfortable running remote calls in error handling.. Its quite unexpected that when I do a port forward call suddenly we are calling LIST PODS (very expensive, may not have permissions). IMO, I would rather make no change than do this lookup.

zhlsunshine: Hi @howardjohn,  how about just print out the notification when there is error? There would be no more other cost and risk. does it make sense?

howardjohn: sgtm

zhlsunshine: okay

pkg/kube/client.go
```diff
@@ -695,7 +695,9 @@ func (c *client) EnvoyDo(ctx context.Context, podName, podNamespace, method, pat
 
 func (c *client) portForwardRequest(ctx context.Context, podName, podNamespace, method, path string, port int) ([]byte, error) {
 	formatError := func(err error) error {
-		return fmt.Errorf("failure running port forward process: %v", err)
+		message := "\n\nNotification: the cluster is IPv6 only cluster and it will break port-forward " +
```
howardjohn, line 698: 'the cluster is IPv6 only cluster' - this is probably not true for 99.9% of users seeing this error...

zhlsunshine: Hi @howardjohn it might be confused, I change the message content, how about now?

zhlsunshine:
Hi @howardjohn, could you please help to approve this PR if possible?  Thanks!

Issue #36434, opened 2021-12-08
Http1.DPE error for tcp incoming traffic occassionally


arunpmohan:
Istio version : 1.9.5

k8s : 1.20 & 1.19

Our service definition is  as follows and the port of interest is 10101 which is a tcp port. 

```
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2021-12-05T04:01:07Z"
  labels:
    app: centralmanagement` 
  name: smcentralmanagement-service
  namespace: sm
spec:
  ports:
  - name: tcp-ltm
    port: 10101
    protocol: TCP
    targetPort: tcp-ltm
  - name: grpc-ltm
    port: 10102
    protocol: TCP
    targetPort: tcp-ltm
  - name: tcp-rollbackmgr
    port: 10103
    protocol: TCP
    targetPort: tcp-rollbackmgr
  selector:
    app: smcentralmanagement
    haActive: "true"
  sessionAffinity: None
  type: ClusterIP
 
```
 

  Our outbound policy is REGISTRY_ONLY and this is an internal service in the same namespace.
 
  Now when other pods tries to connect to this centralmanagement service , we get this http1.dpe error in the istio proxy
  logs of this centralmanagement pods.
 
```
[2021-11-22T21:54:57.850Z] "- - HTTP/1.1" 400 DPE http1.codec_error - "-" 0 11 0 - "-" "-" "-" "-" "-" - - 192.168.162.154:10101 192.168.162.191:36656 outbound_.10101_._.smcentralmanagement-service.sm.svc.cluster.local -
[2021-11-22T21:54:57.949Z] "- - HTTP/1.1" 400 DPE http1.codec_error - "-" 0 11 0 - "-" "-" "-" "-" "-" - - 192.168.162.154:10101 192.168.162
```
 
 
Any help is appreciated. This goes away sometimes on pod restarts or istiod restarts. It doesnt happen always as well.

howardjohn:
> . This goes away sometimes on pod restarts or istiod restarts. It doesnt happen always as well.

So when it does happen, it happens repeatedly on requests in a row?

It sounds like somehow the config is getting incorrectly labeled as HTTP instead of TCP. Based on the config:
```
  - name: tcp-ltm
    port: 10101
    protocol: TCP
    targetPort: tcp-ltm
```

It _should_ be TCP (https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/). But for some reason its using HTTP.

[...]

PR #36462, opened 2021-12-10
Make service instances order the same as ports defined at service spec

bianpengyuan:
fix https://github.com/istio/istio/issues/36434

One port can only talk one protocol. This make selection of protocol stable if user have multiple services port pointing to the same target port with different protocols.


pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -1340,11 +1340,10 @@ func (c *Controller) getProxyServiceInstancesByPod(pod *v1.Pod,
...
 			}
 			// Dedupe the target ports here - Service might have configured multiple ports to the same target port,
-			// we will have to create only one ingress listener per port and protocol so that we do not endup
-			// complaining about listener conflicts.
+			// we will have to create only one ingress listener per port so that we do not endup complaining about
+			// listener conflicts.
 			targetPort := model.Port{
-				Port:     portNum,
-				Protocol: svcPort.Protocol,
+				Port: portNum,
```
howardjohn, line 1346: I wonder if we should have some precedence order: http < tcp < auto.

That way if user has tcp + http, they get downgraded to tcp vs one of their ports breaking.WDYT?

bianpengyuan: Yeah that makes sense.

hzxuzhonghu: Why it fails if we use http instead of tcp?  The real protocol is tcp?

pkg/config/protocol/instance.go
```diff
@@ -56,6 +56,12 @@ const (
 	Unsupported Instance = "UnsupportedProtocol"
 )
 
+// ProtocolOrder defines precedence of protocol in case there is protocol conflict for a port.
+var ProtocolOrder = map[Instance]int{
+	UDP: 1, Unsupported: 2, TCP: 3, TLS: 4, HTTPS: 5, MySQL: 6, Redis: 7, Mongo: 8,
+	Thrift: 9, HTTP: 10, HTTP_PROXY: 11, HTTP2: 12, GRPC: 13, GRPCWeb: 14,
```
hzxuzhonghu, line 62: It looks confusing about this, should we downgrade all ports to tcp if they are not equal

bianpengyuan: If user have http2 and grpc, which works perfectly fine before, and should not be downgraded.

I agree this is confusing, and I cannot find a clear way to set protocol precedences. Probably we should go back to the original commit to just remove protocol from target port dedup: https://github.com/istio/istio/pull/36462/commits/7e9835f29aa9889d0fbfd1baa36c4a2d24cc9676. It's a clean change and makes user service behave consistently, which is strictly better than before where service could be broken or work in different ways randomly. @howardjohn wdyt?

howardjohn:
One thing I do not fully understand there - which one wins? is it just the
first one?

bianpengyuan:
> One thing I do not fully understand there - which one wins? is it just the
first one?

Yeah, with https://github.com/istio/istio/commit/7e9835f29aa9889d0fbfd1baa36c4a2d24cc9676 it will be the first one at service spec.

pkg/config/protocol/instance.go
hzxuzhonghu: SG, as this is only working on inbound, which does not have much diff

bianpengyuan: Reverted to [7e9835f](https://github.com/istio/istio/commit/7e9835f29aa9889d0fbfd1baa36c4a2d24cc9676).

pilot/pkg/serviceregistry/kube/controller/controller_test.go
```diff
@@ -670,7 +670,7 @@ func TestGetProxyServiceInstancesWithMultiIPsAndTargetPorts(t *testing.T) {
 					TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: 8080},
 				},
 			},
-			wantNum: 2,
+			wantNum: 1,
```
howardjohn, line 673: Might be worth adding higher level tests that validate the listeners/clusters are right. Maybe in sidecar_simulation_test

howardjohn (approved):

howardjohn:
Also probably worth a release note

pilot/pkg/serviceregistry/kube/controller/controller_test.go
bianpengyuan: I tried to, but simulation tests are all based on service entry, which does not suffer from this dedup logic.

bianpengyuan:
> Also probably worth a release note

added a release note.

howardjohn (approved):

Issue #36468, opened 2021-12-10
WorkloadEntry accompanied by ServiceEntry giving "STRICT_DNS cluster without endpoints" error


Arsen-Uulu:
### Bug Description

I have created a  following WorkloadEntry and ServiceEntry
```---
apiVersion: networking.istio.io/v1beta1
kind: WorkloadEntry
metadata:
  name:  my-app
  namespace: my-namespace
spec:
  serviceAccount: vm-workload
  address:  myapp.example.com
  labels:
    app: myapp

---
apiVersion: networking.istio.io/v1beta1
kind: ServiceEntry
metadata:
  name: my-app
  namespace: my-namespace
spec:
  hosts:
  - myapp.my-namespace.svc.cluster.local
  location: MESH_INTERNAL
  ports:
  - number: 443
    name: https
    protocol: HTTPS
    targetPort: 8003
  resolution: DNS
  workloadSelector:
    labels:
      app: myapp
```
After checking pilot logs 
I see 
```"STRICT_DNS cluster without endpoints outbound|443||myapp.my-namespace.svc.cluster.local found while pushing CDS"```

I thought that WorkloadEntry should provide Endpoints.

### Version

```prose
istio 1.11.0
kubernetes 1.19
```

bianpengyuan:
I can reproduce. When creating WE before SE, it won't be selected.

Arsen-Uulu:
@bianpengyuan Thank you for your quick reply, is there any workarounds? 

I just added the ServiceEntry with endpoints and removed WorkloadEntry, I still get the same error.

bianpengyuan:
It works for me if ServiceEntry is created before WorkloadEntry.

Arsen-Uulu:
so workaround is to create ServiceEntry then WorkloadEntry?

Arsen-Uulu:
@bianpengyuan  Thank you

Arsen-Uulu:
Workaround didn't work for me

bianpengyuan:
> Workaround didn't work for me

@Arsen-Uulu did you create them together by one kube create? That might not work. What works for me is to create the ServiceEntry, wait a bit, then create the WorkloadEntry.

Arsen-Uulu:
@bianpengyuan Thanks it works. 

is there a reason I don't see a cluster for my workloadEntry when I use istioctl proxy-config cluster, but I see route for it? 
and route is configured to take me to outbound|myworkloadentry|443.

PR #36480, opened 2021-12-11
Fix ServiceEntry selecting WorkloadEntry with host & UDS address.

bianpengyuan:
Fix https://github.com/istio/istio/issues/36468

For a workload entry, it won't be added to workload instances if address is not IP, which will cause it not being selected if a ServiceEntry being added later. This change will add WorkloadEntry with host and UDS address to workload instances, and mark it explicitly only selectable by ServiceEntry.

### DIFF
--- a/pilot/pkg/model/service.go
+++ b/pilot/pkg/model/service.go
@@ -266,6 +266,8 @@ type WorkloadInstance struct {
 	Kind     workloadKind      `json:"kind"`
 	Endpoint *IstioEndpoint    `json:"endpoint,omitempty"`
 	PortMap  map[string]uint32 `json:"portMap,omitempty"`
+	// Can only be selected by service entry.
+	ServiceEntryOnly bool `json:"serviceEntryOnly,omitempty"`
 }
 
 // DeepCopy creates a copy of WorkloadInstance.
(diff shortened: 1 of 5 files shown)


pilot/pkg/model/service.go
```diff
@@ -266,6 +266,8 @@ type WorkloadInstance struct {
 	Kind     workloadKind      `json:"kind"`
 	Endpoint *IstioEndpoint    `json:"endpoint,omitempty"`
 	PortMap  map[string]uint32 `json:"portMap,omitempty"`
+	// Can only be selected by service entry.
+	ServiceEntryOnly bool `json:"serviceEntryOnly,omitempty"`
```
hzxuzhonghu, line 270: Does it need to claim it can be only selected by DNS type SE?

pilot/pkg/serviceregistry/serviceentry/conversion_test.go
```diff
@@ -1001,7 +1017,23 @@ func TestConvertWorkloadEntryToWorkloadInstance(t *testing.T) {
...
 			},
-			out: nil,
+			out: &model.WorkloadInstance{
+				Namespace: "ns1",
+				Kind:      model.WorkloadEntryKind,
+				Endpoint: &model.IstioEndpoint{
+					Labels: map[string]string{
+						"topology.istio.io/cluster": clusterID,
+					},
+					Address:        "scooby.com",
```
howardjohn, line 1027: Will this be sent over EDS and then fail?

bianpengyuan: Updated to avoid selecting by non-DNS SE.

hzxuzhonghu (approved):

PR #36482, opened 2021-12-12
Update eds cache on service add/update only if `K8S_SELECT_WORKLOAD_ENTRIES` is enabled

dddddai:
There is no need to update eds cache on service add/update if `K8S_SELECT_WORKLOAD_ENTRIES` is disabled, as the cache would be updated by k8s endpoints events


hzxuzhonghu (commented):
lgtm. is there any test case

dddddai:
@hzxuzhonghu Thanks for your review, I added a unit test, PTAL :)

hzxuzhonghu (approved):

PR #36485, opened 2021-12-13
Validate locality lb setting: failover only specify region

nicole-lihui:
config locality loadbalance of failover always config invalid config like: region1/zone1
should add validate notes with locality failover 

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [X] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/config/validation/validation_test.go
+++ b/pkg/config/validation/validation_test.go
@@ -5897,6 +5897,54 @@ func TestValidateLocalityLbSetting(t *testing.T) {
 			},
 			valid: false,
 		},
+		{
+			name: "invalid failover src contain '*' wildcard",
+			in: &networking.LocalityLoadBalancerSetting{
+				Failover: []*networking.LocalityLoadBalancerSetting_Failover{
+					{
+						From: "*",
+						To:   "region2",
+					},
+				},
+			},
+			valid: false,
+		},
+		{
+			name: "invalid failover dst contain '*' wildcard",
+			in: &networking.LocalityLoadBalancerSetting{
+				Failover: []*networking.LocalityLoadBalancerSetting_Failover{
+					{
+						From: "region1",
+						To:   "*",
+					},
+				},
+			},
+			valid: false,
+		},
+		{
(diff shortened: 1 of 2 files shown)


pkg/config/validation/validation_test.go
```diff
@@ -5897,6 +5897,54 @@ func TestValidateLocalityLbSetting(t *testing.T) {
 			},
 			valid: false,
 		},
+		{
+			name: "invalid failover src contain '*' wildcard",
+			in: &networking.LocalityLoadBalancerSetting{
+				Failover: []*networking.LocalityLoadBalancerSetting_Failover{
+					{
+						From: "*",
+						To:   "region2",
```
howardjohn, line 5906: Why is this invalid now?

nicole-lihui: From the source code analysis: 
The `failoverSetting.From`  only matching the `locality.Region`, and when matching without using the canonical matching, so thai this invalid now.
```go
		// region not match, apply failover settings when specified
		// update localityLbEndpoints' priority to 4 if failover not match
		if priority == 3 {
			for _, failoverSetting := range failover {
				if failoverSetting.From == locality.Region {
					if localityEndpoint.Locality == nil || localityEndpoint.Locality.Region != failoverSetting.To {
						priority = 4
					}
					break
				}
			}
		}
```

howardjohn (approved):

pkg/config/validation/validation_test.go
wanghonglei5181: in my case,i have a lot of regions and zones in a single cluster,  i only want to failover a special region/zone to another region/zone in another cluster ,i use workloadentry, if do not allow '/', how let it work in my case?

hzxuzhonghu: Why would you need failover from region1.zone1 to region2.zone2, rather than first failover to those ones in same region, like(region1.zone2, region1.zone3)

PR #36486, opened 2021-12-13
Add kind and namespace to kube-inject log

monsteredc:
Provide more specific info for each resource.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [x] User Experience


hzxuzhonghu (commented):
I am confused, cann't we get the ns from pod metadata?

monsteredc:
> I am confused, cann't we get the ns from pod metadata?


@hzxuzhonghu The file to be injected may have no namespace specified like the [helloworld exmaple](https://github.com/istio/istio/blob/master/samples/helloworld/helloworld.yaml). In this case, the metadata namespace will be empty.

hzxuzhonghu:
Empty namespace is equal to `default`

monsteredc:
> Empty namespace is equal to `default`

@hzxuzhonghu The default namespace can also be set via the kube context `namespace` field like https://github.com/istio/istio/blob/cc07c57ef4aae8d095e7166c234a6b94111e7566/istioctl/cmd/root.go#L367
I was trying to provide the namespace info for each of the resources since the resources may have the same name but in different namespaces. Do you think it is better to only print the namespace/name if the resource's namespace is explicitly set?

pkg/kube/inject/inject.go
```diff
@@ -695,15 +699,28 @@ func IntoObject(injector Injector, sidecarTemplate Templates, valuesConfig strin
 	if name == "" {
 		name = deploymentMetadata.Name
 	}
+	namespace := metadata.Namespace
+	if namespace == "" {
+		namespace = defaultNamespace
```
howardjohn, line 704: This is not a safe assumption for kubeinject IMO. I think people will run `istioctl kube-inject -f <file-without-namespace> | kubectl apply -f - -n my-namespace`. This would get set to 'default' here which doesn't seem right?

monsteredc: Thanks for the point

monsteredc: Removed the `default` value, and only print the namespace if it is set in the yaml config.

howardjohn (approved):
Nice!

PR #36490, opened 2021-12-13
Add ignore parse error flag to analyze

### LINKED ISSUE #36471: Analyze always errors on unrecognized files.
### Bug Description

In a gitops repo, there may be no simple way of telling Istio Resource files from other Kubernetes resources.  When all these files are passed to analyze, analyze will always return an error code if any file contains an unrecognized resource.  We need a flag to ignore such errors, to make analyze more useful for gitops and CI workflows.

### Version

```prose
1.12
```

### Additional Information

_No response_

hanxiaop:
Fix https://github.com/istio/istio/issues/36471

### DIFF
--- a/istioctl/cmd/analyze.go
+++ b/istioctl/cmd/analyze.go
@@ -73,6 +73,7 @@ var (
 	suppress          []string
 	analysisTimeout   time.Duration
 	recursive         bool
+	ignoreParseError  bool
 
 	fileExtensions = []string{".json", ".yaml", ".yml"}
 )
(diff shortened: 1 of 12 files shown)


jacob-delgado:
cc @monsteredc how do I test this? I see it already only checks *.yaml/*.json/*.yml files and ignores everything else

monsteredc:
@jacob-delgado Yes it only read `.yaml/.json/*.yml` files. However, if a part of the chunks is ill-formed or unrecognized by the reader, analyze will ignore the entire file. You may try to use the yaml like the one I provided in the test, and the analysis output will still contain some of the Istio recognized resources if the flag `--ignore-parse-errors=true` is specified. Also, no parsing error message will be added to the analysis output.

jacob-delgado (approved):
LGTM but get @therealmitchconnors to agree that this is the solution he wanted in the original PR

jacob-delgado:
I think this is a duplicate of https://github.com/istio/istio/pull/36778

therealmitchconnors:
@hanxiaop I didn't see that you had already created this pull request until I had already merged mine.  Thank you for taking the time to write this, and my apologies for not noticing sooner.  As an alternate PR has already merged I will close this pull request.

Issue #29284, opened 2020-11-30
[Testing] Support real stackdriver


nmittler:
We currently have a fake SD test component and a selection of tests that use it. We should add support for real SD as well. To do this, we'll want an abstraction to the SD component so that tests don't care if they're using real/fake SD. The choice can be controlled by a flag (i.e. either all SD tests run against the real or fake SD ... no mix/match in the same job).

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Additional context**

nmittler:
fyi @douglas-reid @mandarjog since we've talked about this in the past.

bianpengyuan:
not stale.

nmittler:
@bianpengyuan @douglas-reid is this still something on your roadmap?

bianpengyuan:
@nmittler yes, will get back on this once I have more time.

nmittler:
@mandarjog @douglas-reid @bianpengyuan any updates here? Are we prioritizing this work?

bianpengyuan:
Still on my todo list. Need to refresh my memory on this but I think the test can pass now, it is just to enable it at postsubmit.

nmittler:
@bianpengyuan ah great ... so this is essentially fixed?

bianpengyuan:
> @bianpengyuan ah great ... so this is essentially fixed?

Well, I did a quick check yesterday and looks like test was broken, likely because of some new changes: https://github.com/istio/istio/pull/36497. Will need some work to fix that.

douglas-reid:
Was there more work here?

bianpengyuan:
Looks like prow pods now actually lacks permission to report to stackdriver (test failure: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/38281/integ-telemetry_istio/1512123541196115968 from https://github.com/istio/istio/pull/38281) Things left to do in this issue:

1. Figure out why prow cannot make direct telemetry report to stackdriver. This is likely due to shift to workload identity, and SD report permission needs to be bind to prow service account.
2. Adjust query duration to not rely on timing. Basically address this comment: https://github.com/istio/istio/pull/36497#discussion_r772617563. We should use log name to distinguish logs from different cases.
[...]

PR #36497, opened 2021-12-13
Fix stackdriver integ test with real backend.

bianpengyuan:
for https://github.com/istio/istio/issues/29284

Fix stackdriver test with real backend. Once this is merged, we can enable sd test with real backend: https://github.com/istio/test-infra/pull/3718.

### DIFF
--- a/pkg/test/framework/components/stackdriver/google.go
+++ b/pkg/test/framework/components/stackdriver/google.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	cloudtrace "google.golang.org/api/cloudtrace/v1"
@@ -75,7 +76,7 @@ var (
 			resourceType: "k8s_pod",
 		},
 	}
-	queryInterval = -5 * time.Minute
+	queryInterval = -90 * time.Second
 )
 
 func newRealStackdriver(_ resource.Context, _ Config) (Instance, error) {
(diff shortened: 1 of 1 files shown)


pkg/test/framework/components/stackdriver/google.go
```diff
@@ -75,7 +76,7 @@ var (
 			resourceType: "k8s_pod",
 		},
 	}
-	queryInterval = -5 * time.Minute
+	queryInterval = -90 * time.Second
```
howardjohn, line 79: Any reason for this? seems reasonable since we shouldn't need 5min but am curious

howardjohn (approved):

pkg/test/framework/components/stackdriver/google.go
bianpengyuan: We recently added some variant to the access log test (audit log). Fetching with too wide time interval + limit number of entries returned would make the returned entries all from previous test cases.

howardjohn: Can we add some other distinguisher? Assuming our tests are slow so there won't be overlap is not great...

bianpengyuan: Yeah we should be able to filter by log name, this PR is just to make the least amount of change to make test pass :). I will add that in a follow up PR.

douglas-reid: @howardjohn the logs query is based on namespace. do we feel that the echo ns will be reused within a time period?

howardjohn: In real tests it shouldn't since its randomized. However, for local runs `--istio.test.stableNamespaces` removes randomization and is super useful for quick iteration.

douglas-reid: Ok. The tricky bit is that logName is not configurable for Stackdriver. Making that configurable, just for testing, is a big lift.

PR #32469, opened 2021-04-26
Add external profile

frankbu:
Fixes #32370 

[ ] Configuration Infrastructure
[x] Docs
[x] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any changes that may affect Istio users.

PR #36502, opened 2021-12-14
fix default values for external profile

zirain:
Refer to [istio.io/pull/9616](https://github.com/istio/istio.io/pull/9616/files) and https://github.com/istio/istio/pull/32469 
CRD must be installed on remote config cluster

### DIFF
--- a/manifests/profiles/external.yaml
+++ b/manifests/profiles/external.yaml
@@ -5,7 +5,7 @@ kind: IstioOperator
 spec:
   components:
     base:
-      enabled: false
+      enabled: true
     pilot:
       enabled: false
     ingressGateways:


zirain:
kindly ping @frankbu

manifests/profiles/external.yaml
```diff
@@ -5,7 +5,7 @@ kind: IstioOperator
 spec:
   components:
     base:
-      enabled: false
+      enabled: true
```
frankbu, line 8: This is not needed. The `external` profile will install the CRDs if `values.global.configCluster` is `true`.

frankbu, line 8: This is not needed. The `external` profile will install the CRDs if `values.global.configCluster` is `true`.

PR #36505, opened 2021-12-14
add grpc tcp als support

zirain:

1. add grpc tcp als provider
2. respect `FilterStateObjectsToLog` in `MeshConfig`


hzxuzhonghu (commented):
/ok-to-test

You should apply for istio member

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -163,6 +163,13 @@ func buildAccessLogFromTelemetry(push *model.PushContext, mesh *meshconfig.MeshC
 			if al := buildHTTPGrpcAccessLogHelper(push, prov.EnvoyHttpAls); al != nil {
 				als = append(als, al)
```
howardjohn, line 164: how come there is no forListener here? since its only HTTP?

howardjohn (approved):
Now that ALS is in the first class API I think we really need integration tests for this added in the future.

Thanks for your recent PRs here!

PR #36506, opened 2021-12-14
Use buffered channel with signal.Notify

jxlwqq:
The channel used with signal.Notify should be buffered.


pkg/istio-agent/agent.go
```diff
@@ -465,7 +465,7 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
 		}()
 	} else if a.WaitForSigterm() {
 		// wait for SIGTERM and perform graceful shutdown
-		stop := make(chan os.Signal)
+		stop := make(chan os.Signal, 1)
```
lei-tang, line 468: Thanks for the fix! LGTM.

lei-tang (approved):

PR #36508, opened 2021-12-14
>Fix: httpProtocolOptions  be reset when apply connection pool settings

nicole-lihui:
There is only one strategy in idleTimeout and maxRequestsPerConnection, there is another strategy is reset.


howardjohn (commented):
Its not clear what the goal is here? Is it a refactoring or bug fix? If its a bug fix can we add a test?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -940,7 +939,9 @@ func (cb *ClusterBuilder) applyConnectionPool(mesh *meshconfig.MeshConfig, mc *M
 			mc.httpProtocolOptions = &http.HttpProtocolOptions{}
 		}
 		commonOptions := mc.httpProtocolOptions
-		commonOptions.CommonHttpProtocolOptions = &core.HttpProtocolOptions{}
+		if commonOptions.CommonHttpProtocolOptions == nil {
```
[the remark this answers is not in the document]
nicole-lihui: I think this is necessary, because there is no guarantee that the future is always a nil. 
Just like to my subsequent add unit tests.

nicole-lihui:
> Its not clear what the goal is here? Is it a refactoring or bug fix? If its a bug fix can we add a test?

@howardjohn 
Thank you, this is a good idea，add test cases can more clearly.

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -3062,3 +3066,118 @@ func TestApplyDestinationRuleOSCACert(t *testing.T) {
...
+					IdleTimeout: gogo.DurationToProtoDuration(&types.Duration{
+						Seconds: 22,
+					}),
+					MaxRequestsPerConnection: &wrappers.UInt32Value{Value: 22},
+				},
+			},
+		},
+	}
+
+	g := gomega.NewGomegaWithT(t)
```
howardjohn, line 3160: needs to be inside `t.Run` or it will panic if it were to fail. Although we don't need this, can/should use `assert.Equals()`

howardjohn (approved):

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
nicole-lihui: This is a very suggestion, before I didn't find istio has its own `assert` pkg!

PR #36518, opened 2021-12-15
Fix endpoint slice cache memory leak

dddddai:
If an endpoint was deleted or the endpoint key was updated (e.g. `ServicePortName` changed), then the old endpoint would always remain in `endpointByKey` and never get freed

You can reproduce it with [this test](https://github.com/dddddai/istio/blob/memory-leak/pilot/pkg/serviceregistry/kube/controller/endpointslice_test.go#L125-L154)

So this pr drops `endpointByKey` to fix the problem

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/endpointslice.go
+++ b/pilot/pkg/serviceregistry/kube/controller/endpointslice.go
@@ -379,15 +379,13 @@ type endpointKey struct {
 }
 
 type endpointSliceCache struct {
-	mu                            sync.RWMutex
-	endpointKeysByServiceAndSlice map[host.Name]map[string][]endpointKey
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

jacob-delgado:
cc @dddddai Should we try and get this cherry-picked into 1.11 and 1.12? Thanks.

dddddai:
> cc @dddddai Should we try and get this cherry-picked into 1.11 and 1.12? Thanks.

Yeah I think so, thanks for reminding

/cherry-pick release-1.11
/cherry-pick release-1.12

dddddai:
@jacob-delgado, sorry I'm a noob in cherry-picking, could you please help me? Thank you :)

PR #36534, opened 2021-12-16
terminate proxy after termination drain duration even if there are active connections

ramaraochavali:
With ExitOnActiveConnections is enabled, if active connections never become zero, the proxy waits forever. This PR fixes it by exiting when terminationDrainDuration is reached even if there are active connections.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali:
@howardjohn PTAL

pkg/envoy/agent.go
```diff
@@ -148,17 +148,22 @@ func (a *Agent) terminate() {
...
 		ticker := time.NewTicker(activeConnectionCheckDelay)
-		for range ticker.C {
+		select {
+		case <-ticker.C:
 			if a.activeProxyConnections() == 0 {
 				log.Info("There are no more active connections. terminating proxy...")
 				a.abortCh <- errAbort
 				return
 			}
+		case <-time.After(a.terminationDrainDuration):
```
howardjohn, line 158: With defaults of minDrainDuration=5s and terminationDrainDuration=5s, this means the activeProxyConnections is effectively never used...?

It seems like this value should equal to terminationGracePeriodSeconds on pod spec (even then its not *really* required, since k8s will force kill it at that point)?

ramaraochavali: Oh. That is a good point. I did not think about defaults. It makes sense to equal it to terminationGracePeriodSeconds. But in our environment k8s never killed even after 20 mins

ramaraochavali: looks like terminationGracePeriodSeconds is available only in 1.20 - that may be the reason why k8s did not kill our pod.  On 1.20 looks like k8s might kill it - So should we just ignore this for now?

howardjohn: Where did you see that? I think its wrong. I saw it on https://sysdig.com/blog/kubernetes-1-21-whats-new/ but its not accurate, the new feature is node graceful shutdown, not pod, AFAIK

ramaraochavali: Do not remember where I have seen. Let me see if I can get the link. Let me also figure out why k8s never killed the pod - btw it is a job not sure if it matters.

howardjohn:
Jobs are different, when the job container exits Kubernetes will not
SIGTERM the proxy

howardjohn:
...Because it doesn't know one is a "sidecar" and one is the "job" - they
are just two containers.

ramaraochavali:
I see. I remember now. That makes sense.

ramaraochavali:
@howardjohn since we need to terminate this properly for jobs when they call "/quitquitquit" I have changed to terminate at pod spec's terminationGracePeriodSeconds as the max limit.

ramaraochavali:
looks like terminationGracePeriodSeconds can not be used on downward API https://github.com/kubernetes/kubernetes/issues/50032

pkg/envoy/agent.go
```diff
@@ -148,17 +151,27 @@ func (a *Agent) terminate() {
...
+			ac := a.activeProxyConnections()
+			if ac == 0 {
 				log.Info("There are no more active connections. terminating proxy...")
 				a.abortCh <- errAbort
 				return
 			}
+			log.Infof("There are still %d connections active", ac)
+		// This is needed for cases where Jobs's proxy is not terminated by Kubernetes
+		// and Job kill the proxy using agent's /quitquitquit endpoint.
+		case <-time.After(podTerminationGracePeriod):
```
hzxuzhonghu, line 166: why we need this, after `podTerminationGracePeriod`, if this process exist, kubelet will send kill -9 to it.

ramaraochavali: Please see the comment above ^^. For Jobs it is not doing it.

hzxuzhonghu: ic

hzxuzhonghu: After job completes, who will terminate the sidecar?

ramaraochavali:
I am closing this as we can not inject `terminationGracePeriodSeconds` as env var using downward API and it is not worth exposing this as env var just for Jobs.

pkg/envoy/agent.go
ramaraochavali: Job triggers `/quitquitquit` endpoint after it finishes.

Issue #36489, opened 2021-12-13
Sidecar send unexpected RST packet to ingress gateway, which result in intermittent 503 at client side during upload files.


flyingyang:
### Bug Description

We have a storage pod running in our cluster with istio injection enabled and get some intermittent 503 UC errors at client side during upload 300M files. 

Our topology is below, we have an ingress gateway to manage all the traffic entering in our mesh.
![image](https://user-images.githubusercontent.com/22345415/145781983-c30ab55f-85b5-47df-a612-93d50e53c385.png)

First, we check the envoy access log in gateway and sidecar, we see the 503 UC response in gateway side, but not see any 503 in sidecar side. So we think the problem maybe occur in the connection between ingress gateway and sidecar.
Then, we tune the evnoy log level to debug and found this: The ingress gateway's connection is terminated by sidecar, the sidecar receive the "Connection reset by peer" msg.
- ingress gateway envoy log
```
2021-12-07T08:37:28.358739Z debug   envoy http2 [C488678497] updating connection-level initial window size to 268435456
2021-12-07T08:37:28.358906Z debug   envoy http  [C488678497] new stream
2021-12-07T08:37:28.358940Z debug   envoy http  [C488678497][S10432259314075048153] request headers complete (end_stream=false):
':method', 'PUT'
':path', '/random300M_1'
':scheme', 'https'
':authority', ''test.example.com:8443'
'user-agent', 'curl/7.58.0'
'accept', '*/*'
'content-length', '314572800'
'content-type', 'application/x-www-form-urlencoded'

2021-12-07T08:37:28.359015Z debug   envoy router    [C488678497][S10432259314075048153] cluster 'outbound|443||storage-internal-svc.test.svc.kubernetes.io' match for URL '/random300M_1'
2021-12-07T08:37:28.359053Z debug   envoy router    [C488678497][S10432259314075048153] router decoding headers:
':method', 'PUT'
':path', '/random300M_1'
':scheme', 'https'
':authority', 'test.example.com:8443'
'user-agent', 'curl/7.58.0'
'accept', '*/*'
'content-length', '314572800'
'content-type', 'application/x-www-form-urlencoded'
'x-forwarded-proto', 'https'
'x-envoy-internal', 'true'
'x-request-id', 'd0784d05-2b65-4121-b309-a28fe69c91b0'
2021-12-07T08:37:28.367909Z debug   envoy router    [C488678497][S10432259314075048153] pool ready
[...]

PR #36536, opened 2021-12-16
Drop packets in INVALID state to avoid intermittent connection reset from sidecar

shonecyx:
This is to fix bug https://github.com/istio/istio/issues/36489 
Similar upstream issue in k8s https://kubernetes.io/blog/2019/03/29/kube-proxy-subtleties-debugging-an-intermittent-connection-reset/

We find the root cause that sidecar will intermittent reset the connection to ingress gateway during high throughput request(like upload large files 300MB) and the root cause is that there are some out of order packet comes to backend application directly without being DNAT by iptables, so the application reset it directly.

Here is the overflow:

Screen Shot 2021-12-16 at 8 33 51 PM

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ x] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


Kmoneal:
Is there any reason this is not a PR for master to be cherry-picked into 1.12?

shonecyx:
Thanks @Kmoneal, let me close this one.

PR #36539, opened 2021-12-16
Add UNINITIALIZED state to `istioctl proxy-config secret`

howardjohn:
This is a state Envoy can get into when it fails to parse a secret.
Today, everything in istioctl reports its fine, when it really is just
as blocking as `WARMING`

Can reproduce with a cert that has the cert field in both the cert AND key section. which is not valid of course


irisdingbj (approved):
Thanks for the fix!

Issue #36541, opened 2021-12-16
istiod nil pointer exception


ian-mi:
### Bug Description

After installing istio through helm with the following commands, istiod crash loops with a nil pointer exception:

```
$ kubectl create namespace istio-system
$ helm upgrade --install istio-base istio/base -n istio-system --devel
$ helm upgrade --install \
                -n istio-system \
                istiod \
        istio/istiod \
        --devel \
        --set global.operatorManageWebhooks=true --set pilot.env.PRIORITIZED_LEADER_ELECTION=false --set pilot.env.INJECTION_WEBHOOK_CONFIG_NAME=''  \
        --set global.hub=gcr.io/istio-testing \
        --set global.tag=latest \
        --set telemetry.enabled=true \
        --set global.sds.token.aud="ianmi-prod.svc.id.goog" \
        --set meshConfig.trustDomain="ianmi-prod.svc.id.goog" \
        \
        --set meshConfig.proxyHttpPort=15007 \
        --set meshConfig.accessLogFile=/dev/stdout \
        \
        --set pilot.replicaCount=1 \
        --set pilot.autoscaleEnabled=false \
        \
        --set pilot.env.TOKEN_AUDIENCES="ianmi-prod.svc.id.goog\,istio-ca" \
        --set pilot.env.ISTIO_MULTIROOT_MESH=true \
        --set pilot.env.PILOT_ENABLE_WORKLOAD_ENTRY_AUTOREGISTRATION=true \
        --set pilot.env.PILOT_ENABLE_WORKLOAD_ENTRY_HEALTHCHECKS=true
```

Error:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x24da0f1]

goroutine 101 [running]:
istio.io/istio/pilot/pkg/leaderelection.(*LeaderElection).Run(0xc000cce840, 0xc000910fc0)
        istio.io/istio/pilot/pkg/leaderelection/leaderelection.go:78 +0x31
istio.io/istio/pilot/pkg/bootstrap.(*Server).initK8SConfigStore.func1(0x0)
        istio.io/istio/pilot/pkg/bootstrap/configcontroller.go:182 +0x15e
istio.io/istio/pilot/pkg/server.(*instance).RunComponentAsyncAndWait.func1.1()
        istio.io/istio/pilot/pkg/server/instance.go:139 +0x2e
[...]

PR #36542, opened 2021-12-16
Guard triggering of default watch go routine with enablement flag.

bianpengyuan:
fix https://github.com/istio/istio/issues/36541

### DIFF
--- a/pilot/pkg/leaderelection/leaderelection.go
+++ b/pilot/pkg/leaderelection/leaderelection.go
@@ -75,7 +75,9 @@ type LeaderElection struct {
 
 // Run will start leader election, calling all runFns when we become the leader.
 func (l *LeaderElection) Run(stop <-chan struct{}) {
-	go l.defaultWatcher.Run(stop)
+	if l.prioritized {
+		go l.defaultWatcher.Run(stop)
+	}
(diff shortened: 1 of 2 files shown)


Monkeyanator (approved):

pilot/pkg/leaderelection/leaderelection.go
```diff
@@ -75,7 +75,9 @@ type LeaderElection struct {
 
 // Run will start leader election, calling all runFns when we become the leader.
 func (l *LeaderElection) Run(stop <-chan struct{}) {
-	go l.defaultWatcher.Run(stop)
+	if l.prioritized {
+		go l.defaultWatcher.Run(stop)
```
howardjohn, line 79: very nit: should we check `l.defaultWatcher != nil`? This is for sure correct, but a change somewhere else could make the logic slightly off in the future?

howardjohn (approved):
cc @Monkeyanator

pilot/pkg/leaderelection/leaderelection.go
bianpengyuan: Updated to also check on l.defaultWatcher is not nil.

PR #36549, opened 2021-12-17
Bump min k8s version to 1.19

kebe7jun:
According to https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases, the min k8s version is 1.19.


johnma14:
@kebe7jun there is already a PR out there addressing this: https://github.com/istio/istio/pull/36325

ericvn:
Also not that although we only officially support 1.19 and above (and that's a Istio 1.12 statement), we do test and allow k8s versions back to 1.16.

PR #36551, opened 2021-12-17
Fix rootca-compare cmd validate logic error

kebe7jun:
`configDumpFile` not inited in this cmd... we should not use it...


kebe7jun:
@hanxiaop PTAL.

hanxiaop (approved):
LGTM

PR #36553, opened 2021-12-17
Override gateway deployment selector with k8s Ingress's annotation

RolandMa1986:
Signed-off-by: Roland.Ma <rolandma@kubesphere.io>

Override default gateway deployment selector with k8s Ingress's annotation, so that users can selector which one to use. 
Currently, I found only one deployment can be used to serve the k8s ingress, even I setup up with multiple Gateway deployments.  The default Selector can be either "isito:ingressgateway " or from the mesh config.
So I added an annotation to override the default setting in the ingress to allow users to override the selector config.
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: foo
  namespace: ns
  annotations:
    istio.io/gateway.selector: "istio=ingressgateway-foo"
```

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn:
I am not a huge fan of adding this complexity. If you need this, you would likely be better suited with Istio gateway or https://gateway-api.sigs.k8s.io/. Istio generally is not trying to be an advanced `Ingress` controller, and this is functionality that other ingress controllers do not even have.

PR #36563, opened 2021-12-19
Bugfix: changes in delegate virtual service do not take effect

### LINKED ISSUE #36525: changes in delegate virtual service do not take effect
### Bug Description

Similar to #29123 seems to be happening with Istio 1.12.

Suppose we have following setup:

```
kind: VirtualService
metadata:
  name: foo
  namespace: foo
spec:
  hosts:
  - bar.bar
  http:
  - match:
    - uri:
        prefix: "/prefix"
    delegate:
      name: bar
      namespace: bar
---
kind: VirtualService
metadata:
  name: bar
  namespace: bar
spec:
  http:
  - route:
    - destination:
        host: bar.bar
        subset: stable
      weight: 90
    - destination:
        host: bar.bar
        subset: canary
      weight: 10
```

When we change the `bar` virtual service, for example, change the weight from 90,10 to 80,20, routes are not updated on the foo sidecar proxy. Restarting the pod does not get the routes updated either. Triggering some sort of update over the `foo` virtual service does update the routes.

### Version

```prose
$ istioctl version
client version: 1.12.1
control plane version: 1.12.1
data plane version: 1.12.1

$ kubectl version --short
Client Version: v1.21.1
Server Version: v1.21.2-eks-0389ca3
```

### Additional Information

From Istiod logs I can see that VirtualServices are indeed being merged, and an update pushed, but if I check a config dump of the proxy, the weights are not updated.

dddddai:
The root cause is that the rds cache is not cleared when delegate virtual services were updated

This pr adds delegate virtual services to dependent configs of the rds cache entry, so that we can clear the cache when delegate virtual services are updated

Fix #36525

### DIFF
--- a/releasenotes/notes/refresh-rds-cache.yaml
+++ b/releasenotes/notes/refresh-rds-cache.yaml
@@ -1,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+issue:
+- https://github.com/istio/istio/issues/36525
+releaseNotes:
+- |
+  **Fixed** changes in delegate virtual service do not take effect when rds cache enabled
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -842,6 +842,9 @@ func (ps *PushContext) VirtualServicesForGateway(proxy *Proxy, gateway string) [
 // DelegateVirtualServicesConfigKey lists all the delegate virtual services configkeys associated with the provided virtual services
 func (ps *PushContext) DelegateVirtualServicesConfigKey(vses []config.Config) []ConfigKey {
 	var out []ConfigKey
+	if !features.EnableVirtualServiceDelegate {
+		return out
+	}
 	for _, vs := range vses {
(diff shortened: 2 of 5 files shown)


releasenotes/notes/refresh-rds-cache.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+issue:
+- https://github.com/istio/istio/issues/36525
+releaseNotes:
+- |
+  **Fixed** changes in delegate virtual service do not take effect
```
hzxuzhonghu, line 8: **Fixed** changes in delegate virtual service do not take effect when rds cache enabled

hzxuzhonghu (approved):

releasenotes/notes/refresh-rds-cache.yaml
dddddai: Done

howardjohn (approved):

hzxuzhonghu:
This should be backported to 1.12

PR #36566, opened 2021-12-20
Drop packets in INVALID state to avoid intermittent connection reset from sidecar

shonecyx:
Please provide a description of this PR:
This is to fix bug #36489
Similar upstream issue in k8s https://kubernetes.io/blog/2019/03/29/kube-proxy-subtleties-debugging-an-intermittent-connection-reset/

We find the root cause that sidecar will intermittent reset the connection to ingress gateway during high throughput request(like upload large files 300MB) and the root cause is that there are some out of order packet comes to backend application directly without being DNAT by iptables, so the application reset it directly.

Here is the overflow:
Screen Shot 2021-12-22 at 1 54 59 PM

- [ ] Docs
- [ ] Installation
- [ x] Networking
- [ x] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/tools/istio-iptables/pkg/capture/run.go
+++ b/tools/istio-iptables/pkg/capture/run.go
@@ -131,7 +131,8 @@ func (cfg *IptablesConfigurator) logConfig() {
 	b.WriteString(fmt.Sprintf("ISTIO_EXCLUDE_INTERFACES=%s\n", os.Getenv("ISTIO_EXCLUDE_INTERFACES")))
 	b.WriteString(fmt.Sprintf("ISTIO_SERVICE_CIDR=%s\n", os.Getenv("ISTIO_SERVICE_CIDR")))
 	b.WriteString(fmt.Sprintf("ISTIO_SERVICE_EXCLUDE_CIDR=%s\n", os.Getenv("ISTIO_SERVICE_EXCLUDE_CIDR")))
-	b.WriteString(fmt.Sprintf("ISTIO_META_DNS_CAPTURE=%s", os.Getenv("ISTIO_META_DNS_CAPTURE")))
+	b.WriteString(fmt.Sprintf("ISTIO_META_DNS_CAPTURE=%s\n", os.Getenv("ISTIO_META_DNS_CAPTURE")))
+	b.WriteString(fmt.Sprintf("ISTIO_META_INVALID_DROP=%s\n", os.Getenv("ISTIO_META_INVALID_DROP")))
 	log.Infof("Istio iptables environment:\n%s", b.String())
 	cfg.cfg.Print()
 }
@@ -354,6 +355,13 @@ func (cfg *IptablesConfigurator) Run() {
 
 	// Do not capture internal interface.
 	cfg.shortCircuitKubeInternalInterface()
+
+	// Create a rule for invalid drop in PREROUTING chain in mangle table, so the iptables will drop the out of window packets instead of reset connection .
+	dropInvalid := cfg.cfg.DropInvalid
+	if dropInvalid {
+		cfg.iptables.AppendRule(iptableslog.UndefinedCommand, constants.PREROUTING, constants.MANGLE, "-m", "conntrack", "--ctstate",
+			"INVALID", "-j", constants.DROP)
+	}
 
 	// Create a new chain for to hit tunnel port directly. Envoy will be listening on port acting as VPN tunnel.
 	cfg.iptables.AppendRule(iptableslog.UndefinedCommand, constants.ISTIOINBOUND, constants.NAT, "-p", constants.TCP, "--dport",
--- a/releasenotes/notes/36566.yaml
+++ b/releasenotes/notes/36566.yaml
@@ -1,0 +1,10 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+issue:
+  - https://github.com/istio/istio/pull/36566
+releaseNotes:
+  - |
+    **Fixed** an issue that sidecar iptables will cause intermittent connection reset due to the out of window packet.
+    Introduced a flag meshConfig.defaultConfig.proxyMetadata.ISTIO_META_INVALID_DROP to control this setting.
+
--- a/tools/istio-iptables/pkg/capture/run_test.go
+++ b/tools/istio-iptables/pkg/capture/run_test.go
@@ -228,6 +228,12 @@ func TestIptables(t *testing.T) {
 				cfg.TraceLogging = true
 			},
 		},
+		{
+			"drop-invalid",
+			func(cfg *config.Config) {
+				cfg.DropInvalid = true
+			},
+		},
 	}
 	for _, tt := range cases {
 		t.Run(tt.name, func(t *testing.T) {
--- a/tools/istio-iptables/pkg/capture/testdata/drop-invalid.golden
+++ b/tools/istio-iptables/pkg/capture/testdata/drop-invalid.golden
@@ -1,0 +1,17 @@
+iptables -t nat -N ISTIO_INBOUND
+iptables -t nat -N ISTIO_REDIRECT
+iptables -t nat -N ISTIO_IN_REDIRECT
+iptables -t nat -N ISTIO_OUTPUT
+iptables -t mangle -A PREROUTING -m conntrack --ctstate INVALID -j DROP
+iptables -t nat -A ISTIO_INBOUND -p tcp --dport 15008 -j RETURN
+iptables -t nat -A ISTIO_REDIRECT -p tcp -j REDIRECT --to-ports 15001
+iptables -t nat -A ISTIO_IN_REDIRECT -p tcp -j REDIRECT --to-ports 15006
+iptables -t nat -A OUTPUT -p tcp -j ISTIO_OUTPUT
+iptables -t nat -A ISTIO_OUTPUT -o lo -s 127.0.0.6/32 -j RETURN
+iptables -t nat -A ISTIO_OUTPUT -o lo ! -d 127.0.0.1/32 -m owner --uid-owner 1337 -j ISTIO_IN_REDIRECT
+iptables -t nat -A ISTIO_OUTPUT -o lo -m owner ! --uid-owner 1337 -j RETURN
+iptables -t nat -A ISTIO_OUTPUT -m owner --uid-owner 1337 -j RETURN
+iptables -t nat -A ISTIO_OUTPUT -o lo ! -d 127.0.0.1/32 -m owner --gid-owner 1337 -j ISTIO_IN_REDIRECT
+iptables -t nat -A ISTIO_OUTPUT -o lo -m owner ! --gid-owner 1337 -j RETURN
+iptables -t nat -A ISTIO_OUTPUT -m owner --gid-owner 1337 -j RETURN
+iptables -t nat -A ISTIO_OUTPUT -d 127.0.0.1/32 -j RETURN
--- a/tools/istio-iptables/pkg/cmd/root.go
+++ b/tools/istio-iptables/pkg/cmd/root.go
@@ -39,6 +39,9 @@ var (
 	// Enable interception of DNS.
 	dnsCaptureByAgent = env.RegisterBoolVar("ISTIO_META_DNS_CAPTURE", false,
 		"If set to true, enable the capture of outgoing DNS packets on port 53, redirecting to istio-agent on :15053").Get()
+	// Enable invalid drop iptables rule to drop the out of window packets
+	invalidDropByIptables = env.RegisterBoolVar("ISTIO_META_INVALID_DROP", false,
+		"If set to true, enable the invalid drop iptables rule, default false will cause iptables reset out of window packets").Get()
 )
 
 var rootCmd = &cobra.Command{
@@ -121,6 +124,7 @@ func constructConfig() *config.Config {
 		SkipRuleApply:           viper.GetBool(constants.SkipRuleApply),
 		RunValidation:           viper.GetBool(constants.RunValidation),
 		RedirectDNS:             viper.GetBool(constants.RedirectDNS),
+		DropInvalid:             viper.GetBool(constants.DropInvalid),
 		CaptureAllDNS:           viper.GetBool(constants.CaptureAllDNS),
 		OutputPath:              viper.GetString(constants.OutputPath),
 		NetworkNamespace:        viper.GetString(constants.NetworkNamespace),
@@ -321,6 +325,11 @@ func bindFlags(cmd *cobra.Command, args []string) {
 	}
 	viper.SetDefault(constants.RedirectDNS, dnsCaptureByAgent)
 
+	if err := viper.BindPFlag(constants.DropInvalid, cmd.Flags().Lookup(constants.DropInvalid)); err != nil {
+		handleError(err)
+	}
+	viper.SetDefault(constants.DropInvalid, invalidDropByIptables)
+
 	if err := viper.BindPFlag(constants.CaptureAllDNS, cmd.Flags().Lookup(constants.CaptureAllDNS)); err != nil {
 		handleError(err)
 	}
@@ -415,6 +424,8 @@ func bindCmdlineFlags(rootCmd *cobra.Command) {
 	rootCmd.Flags().Bool(constants.RunValidation, false, "Validate iptables")
 
 	rootCmd.Flags().Bool(constants.RedirectDNS, dnsCaptureByAgent, "Enable capture of dns traffic by istio-agent")
+
+	rootCmd.Flags().Bool(constants.DropInvalid, invalidDropByIptables, "Enable invalid drop in the iptables rules")
 
 	rootCmd.Flags().Bool(constants.CaptureAllDNS, false,
 		"Instead of only capturing DNS traffic to DNS server IP, capture all DNS traffic at port 53. This setting is only effective when redirect dns is enabled.")
--- a/tools/istio-iptables/pkg/config/config.go
+++ b/tools/istio-iptables/pkg/config/config.go
@@ -50,6 +50,7 @@ type Config struct {
 	SkipRuleApply           bool          `json:"SKIP_RULE_APPLY"`
 	RunValidation           bool          `json:"RUN_VALIDATION"`
 	RedirectDNS             bool          `json:"REDIRECT_DNS"`
+	DropInvalid             bool          `json:"DROP_INVALID"`
 	CaptureAllDNS           bool          `json:"CAPTURE_ALL_DNS"`
 	EnableInboundIPv6       bool          `json:"ENABLE_INBOUND_IPV6"`
 	DNSServersV4            []string      `json:"DNS_SERVERS_V4"`
@@ -87,6 +88,7 @@ func (c *Config) Print() {
 	b.WriteString(fmt.Sprintf("KUBE_VIRT_INTERFACES=%s\n", c.KubeVirtInterfaces))
 	b.WriteString(fmt.Sprintf("ENABLE_INBOUND_IPV6=%t\n", c.EnableInboundIPv6))
 	b.WriteString(fmt.Sprintf("DNS_CAPTURE=%t\n", c.RedirectDNS))
+	b.WriteString(fmt.Sprintf("DROP_INVALID=%t\n", c.DropInvalid))
 	b.WriteString(fmt.Sprintf("CAPTURE_ALL_DNS=%t\n", c.CaptureAllDNS))
 	b.WriteString(fmt.Sprintf("DNS_SERVERS=%s,%s\n", c.DNSServersV4, c.DNSServersV6))
 	b.WriteString(fmt.Sprintf("OUTPUT_PATH=%s\n", c.OutputPath))
--- a/tools/istio-iptables/pkg/constants/constants.go
+++ b/tools/istio-iptables/pkg/constants/constants.go
@@ -53,6 +53,7 @@ const (
 	REDIRECT = "REDIRECT"
 	MARK     = "MARK"
 	CT       = "CT"
+	DROP     = "DROP"
 )
 
 const (
@@ -98,6 +99,7 @@ const (
 	IptablesProbePort         = "iptables-probe-port"
 	ProbeTimeout              = "probe-timeout"
 	RedirectDNS               = "redirect-dns"
+	DropInvalid               = "drop-invalid"
 	CaptureAllDNS             = "capture-all-dns"
 	OutputPath                = "output-paths"
 	NetworkNamespace          = "network-namespace"


shonecyx:
@howardjohn  Would you share some insights there? The invalid drop seems to be a kernel issue but since sidecar is using iptables, we can also add invalid drop in the iptables rules. 
BTW there is high chance to reproduce this issue with high throughput, like us in a storage service, if we keep uploading 300 MB file for several times it's almost 100% reproducible.

hzxuzhonghu (commented):
/ok-to-test

tools/istio-iptables/pkg/capture/run.go
```diff
@@ -359,6 +359,10 @@ func (cfg *IptablesConfigurator) Run() {
 	cfg.iptables.AppendRule(iptableslog.UndefinedCommand, constants.ISTIOINBOUND, constants.NAT, "-p", constants.TCP, "--dport",
 		cfg.cfg.InboundTunnelPort, "-j", constants.RETURN)
 
+	// Create a rule for invalid drop
```
howardjohn, line 362: I think this rule is not self explanatory, and could use some extensive comments on why we do this. I actually don't understand it myself. Ideally linking to some docs recommending we do this

tools/istio-iptables/pkg/capture/run.go
```diff
@@ -359,6 +359,10 @@ func (cfg *IptablesConfigurator) Run() {
...
+	cfg.iptables.AppendRule(iptableslog.UndefinedCommand, constants.ISTIOINBOUND, constants.MANGLE, "-m", "conntrack", "--ctstate",
```
howardjohn, line 363: Why MANGLE? AFAIK we only use it for tproxy today?

tools/istio-iptables/pkg/capture/run.go
howardjohn: Actually after reading the k8s blog it explains it pretty well, could link there in the docs.

howardjohn (commented):
I am a bit concerned about impacts on stability from this. It seems reasonable but this is the most critical code in Istio and has hard to understand behavior. In particular, users on many platforms may already set nf_conntrack_tcp_be_liberal which the doc listed suggests is a suitable alternative. I worry we will destabilize these users, who are not impacted by this bug. Maybe we should have it in a feature flag for now?

howardjohn:
Awesome job debugging this

tools/istio-iptables/pkg/capture/run.go
shonecyx: Thanks @howardjohn, here is the doc link  https://kubernetes.io/blog/2019/03/29/kube-proxy-subtleties-debugging-an-intermittent-connection-reset/

tools/istio-iptables/pkg/capture/run.go
hzxuzhonghu: Is it because it is the first table that the rule will be dropped as early as possible

hzxuzhonghu:
Which packet is not dnated and invalid

shonecyx:
@hzxuzhonghu would you go through this upstream issue and you'll get more background:  https://kubernetes.io/blog/2019/03/29/kube-proxy-subtleties-debugging-an-intermittent-connection-reset/

@howardjohn I'l like to share more details of the debugging, it's a little bit tricky than the kube-proxy issue because tcpdump didn't help reveal the root cause since it kicks in before the iptables DNAT due to the sidecar and app are in same network namespace, so we wrote some eBFP tool to dig into socket state in the kernel.

**Packet capture analysis**

Ingress Gateway IP: 10.0.0.1
Sidecar/Application Pod IP: 10.0.0.2

The ip and port should change as :

**Ingress Gateway:** 
```
10.0.0.1 -> 10.0.0.2:9000
```

**Sidecar :**  
```
10.0.0.1 -> 10.0.0.2:15006 (iptables): DNAT to change the port to istio inbound
10.0.0.1 -> 10.0.0.2:15006 (envoy): Forward to envoy
127.0.0.1 ->127.0.0.1:9000 -> application: Envoy picks up a new TCP connection to application
```

Here is the tcpdump in backend pod:
```
05:34:34.814254 IP 10.0.0.2.9000 > 10.0.0.1.53304: Flags [R], seq 3142962686, win 0, length 0
05:34:34.814989 IP 10.0.0.2.9000 > 10.0.0.1.53304: Flags [R], seq 3142962686, win 0, length 0
05:34:34.815045 IP 10.0.0.1.53304 > 10.0.0.2.9000: Flags [R], seq 1351926695, win 0, length 0
05:34:34.815536 IP 127.0.0.1.57088 > 127.0.0.1.9000: Flags [F.], seq 1128675839, ack 912453698, win 256, options [nop,nop,TS val 397913316 ecr 397912839], length 0
05:34:34.826703 IP 127.0.0.1.57088 > 127.0.0.1.9000: Flags [R], seq 1128675840, win 0, length 0
```
When the issue happens, the packet flow is as follows:

1. Sidecar send tcp rst to ingress  gateway
2. ingress  gateway send tcp rst to sidecar
3. Sidecar send tcp fin to application server
4. Sidecar send tcp rst to application server

But for now we can't tell who sent the rst, envoy or the application. Then we enabled trace log in envoy:
```
2021-12-13T03:04:11.315747Z trace   envoy connection    [C1779] socket event: 3
2021-12-13T03:04:11.315750Z trace   envoy connection    [C1779] write ready
2021-12-13T03:04:11.315752Z trace   envoy connection    [C1779] read ready. dispatch_buffered_data=false
2021-12-13T03:04:11.315766Z debug   envoy misc  Unknown error code 104 details Connection reset by peer
2021-12-13T03:04:11.315771Z trace   envoy connection    [C1779] ssl read returns: -1
2021-12-13T03:04:11.315773Z trace   envoy connection    [C1779] ssl error occurred while read: SYSCALL
2021-12-13T03:04:11.315775Z trace   envoy connection    [C1779] ssl read 0 bytes
2021-12-13T03:04:11.315778Z debug   envoy connection    [C1779] remote close
2021-12-13T03:04:11.315781Z debug   envoy connection    [C1779] closing socket: 0
2021-12-13T03:04:11.315793Z debug   envoy misc  Unknown error code 32 details Broken pipe
2021-12-13T03:04:11.315797Z debug   envoy connection    [C1779] SSL shutdown: rc=-1
2021-12-13T03:04:11.315974Z trace   envoy connection    [C1779] raising connection event 0
2021-12-13T03:04:11.315993Z debug   envoy http  [C1779][S1987488964837614969] stream reset
```
We checked the code, the stack trace is:
```
ConnectionImpl::onReadReady()
-> SslSocket::doRead(Buffer::Instance & read_buffer)
    -> SslSocket::sslReadIntoSlice(Buffer::RawSlice& slice)
        -> int SSL_read(SSL *ssl, void *buf, int num); // in lib “openssl/ssl.h”
```
When we enter into the onReadReady function,  the connection must be connected because of the ASSERT statement. Then it calls the SslCocket::doRead function, and the doRead function also ensures the socket is connected.
From the envoy trace log above, the error occurred when we called SSL_read to read data from the socket.

So the order is “The socket is reset unexpectedly -> SSL_read return -1 -> doRead failed -> envoy close the connection”. 
When ssl_socket was ready to read data, the socket received the rst packet from the ingress gateway. And the SSL_read function returns the error code and error message. BTW we tried to disable SSL to use plain text, issue is still there when do read.

So far we still haven't found the reason why sidecar sent rst packet to the ingress gateway pod. It could be something wrong in the kernel of the backend pod, hence we tried to use the eBFP.

From previous analysis,  we can conclude that the sidecar envoy sent a TCP reset to ingress gateway which is an unexpected behavior.  In many cases from experience,  the TCP reset happens when the socket has been closed and then still receives the packet to that socket.  So the next step  is to verify whether tcp_close is called before TCP reset is sent. 
Checking kernel code, we can see that there are two functions to send TCP reset.
tcp_send_active_reset() and tcp_v4_send_reset()

```
void tcp_send_active_reset(struct sock *sk, gfp_t priority) {
     …
     ...
     tcp_init_nondata_skb(skb, tcp_acceptable_seq(sk),
                  TCPHDR_ACK | TCPHDR_RST);   //  For this case, the TCP reset packet must have ACK and RST flags. But from  the packet capture,  we can see there is no ACK flag for the  TCP reset packet. So we can conclude that this is not the case.

}
```
So the TCP reset should be generated by the function tcp_v4_send_reset.  In order to verify whether the socket is closed unexpectedly, we can use eBPF to kprobe tcp_v4_send_reset , tcp_close, inet_sk_set_state and inet_sk_state_store.

We know that the issue happens between Ingress Gateway(10.0.0.1) and sidecar (10.0.0.2).

The eBPF will filter the packet (10.0.0.1<->10.0.0.2). When the issue happens, from eBPF output, we can know that there is no tcp_close before tcp_v4_send_reset. So we can exclude the assumption that  the socket is closed unexpectedly. 
Checking the sock details when TCP reset is sent,  the sock state is TCP_LISTEN. From kernel code, we can see that if a socket is in TCP_LISTEN, it will only accept the TCP SYN handshaking packet. But now it receives the data packet from Ingress Gateway, that’s why it will send TCP reset. 

Why is the sock state TCP_LISTEN? Checking the sock address, we can see that the sock for the TCP reset is not the original sock to transmit data.  And also we found the destination port is 9000 instead of 15006.  Now we know the truth. Actually the upstream server will listen to port 9000 and envoy will listen to port 15006. The packet is supposed to do destination port NAT from 9000 to 15006. Now the packet misses the NAT and hits the listening sock of the upstream server and got rst.

@hzxuzhonghu  I hope you get the answer from those details. 

In our case there is pretty high throughput in ingress gateway and sidecar, for example we need to upload 3TB file within 30 mins through ingress gateway and almost every day we can see some resets, although client retry can help tolerate, it would be better we can handle this in sidecar.

tools/istio-iptables/pkg/capture/run.go
shonecyx: @howardjohn  I added more details for the debugging. In our case we are still using the default value 0 nf_conntrack_tcp_be_liberal, so reset the Out-of-Order (OFO) packet is the expected behavior and we have similar concern to enable it at node level and we don't know why kernel use 0 as default value. 
So config the DROP rule in app pod network namespace would be lower risk than config it in node level.

tools/istio-iptables/pkg/capture/run.go
shonecyx: Per my understanding, we need to configure iptables to drop invalid packets instead of forwarding them without address translation. Create it in INPUT chain would also help, I'll verify it in our case first to see if the reset is gone after this rule configured.
```
iptables -I ISTIOINBOUND -m conntrack --ctstate INVALID -j DROP
```

hzxuzhonghu:
Great work, thanks it is much clearer.

lambdai:
@shonecyx Thank you for sharing the diagnose procedure. That's awesome.

I have a couple of questions on the concrete phenomenon. 

> In our case there is pretty high throughput in ingress gateway and sidecar, for example we need to upload 3TB file within 30 mins through ingress gateway and almost every day we can see some resets, although client retry can help tolerate, it would be better we can handle this in sidecar.

1. Is the long tcp connection a potential cause of the issue or a break http transaction is more damaging? 

2. Kind of duplication to my first question. Do you see the reset packet only at the beginning of the tcp stream, or it's possible all over the life of the tcp request? 


Thanks

tools/istio-iptables/pkg/capture/run.go
lambdai: The drop decision can be made anytime after the conntrack. Now that table MANGLE is the first after the conntrack, it's the indeed the best table

noah8713:
Thanks @shonecyx. To add on to more details,  our k8s version is v1.18  and upstream k8s proxy already supports setting this for proxy with iptables mode https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/iptables/proxier.go#L1468  . However, since we are using ipvs mode, seems this logic of dropping invalid packet didn't kick in for any pods. We did a quick test by setting dropInvalidConntrack with iptables mode for k8s proxy and the logic kicks in for all pods.

 ```
iptables-save  | grep INVA
-A KUBE-FORWARD -m conntrack --ctstate INVALID -j DROP
```
Hence,  taking this route which will only fix pods that are in mesh and regular k8s pods will still see this issue as per https://github.com/kubernetes/kubernetes/issues/74839 . If no other alternative, +1 to merge this too and may be we can propose a fix for explicitly setting this in k8s even in ipvs mode. Of course, on the side we can still explore setting nf_conntrack_tcp_be_liberal =1  on node level directly which will work with mesh/non-mesh pods.

File: tools/istio-iptables/pkg/capture/run.go (hunk 2 above)
Comment (shonecyx) [reply tools/istio-iptables/pkg/capture/run.go]: @lambdai  Based on below diagram, I think we can configure the invalid drop in PREROUTING chain. 
proxy

Actually I tired to add to ISTIO_INBOUND but failed with error:
```
iptables -t mangle -A  ISTIO_INBOUND -p TCP    -m conntrack --ctstate INVALID -j DROP
iptables: No chain/target/match by that name.
```
And the reason seems like the ISTIO related chains are not in `mangle` table and `INPUT` chain won't work since only non-TCP traffic will go through it as per above diagram. 

```
iptables -t mangle  -L --line-numbers -n
Chain PREROUTING (policy ACCEPT)
num  target     prot opt source               destination
1    DROP       all  --  0.0.0.0/0            0.0.0.0/0            ctstate INVALID

Chain INPUT (policy ACCEPT)
num  target     prot opt source               destination

Chain FORWARD (policy ACCEPT)
num  target     prot opt source               destination

Chain OUTPUT (policy ACCEPT)
num  target     prot opt source               destination

Chain POSTROUTING (policy ACCEPT)
num  target     prot opt source               destination
```

After adding this rule I can see it now in the PREROUTING:
```
 iptables-save
# Generated by iptables-save v1.4.21 on Wed Dec 22 02:53:04 2021
*raw
:PREROUTING ACCEPT [25195:6428378]
:OUTPUT ACCEPT [22269:6017603]
COMMIT
# Completed on Wed Dec 22 02:53:04 2021
# Generated by iptables-save v1.4.21 on Wed Dec 22 02:53:04 2021
*mangle
:PREROUTING ACCEPT [23364:6031899]
:INPUT ACCEPT [23364:6031899]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [20556:5473772]
:POSTROUTING ACCEPT [20556:5473772]
**-A PREROUTING -m conntrack --ctstate INVALID -j DROP**
COMMIT
# Completed on Wed Dec 22 02:53:04 2021
# Generated by iptables-save v1.4.21 on Wed Dec 22 02:53:04 2021
*filter
:INPUT ACCEPT [65109:15777691]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [58256:15968553]
COMMIT
# Completed on Wed Dec 22 02:53:04 2021
# Generated by iptables-save v1.4.21 on Wed Dec 22 02:53:04 2021
*nat
:PREROUTING ACCEPT [358200:21495109]
:INPUT ACCEPT [374732:22297943]
:OUTPUT ACCEPT [830897:64816986]
:POSTROUTING ACCEPT [1520591:120009333]
:ISTIO_INBOUND - [0:0]
:ISTIO_IN_REDIRECT - [0:0]
:ISTIO_OUTPUT - [0:0]
:ISTIO_REDIRECT - [0:0]
-A PREROUTING -p tcp -j ISTIO_INBOUND
-A OUTPUT -p tcp -j ISTIO_OUTPUT
-A OUTPUT -p udp -m udp --dport 53 -m owner --uid-owner 1337 -j RETURN
-A OUTPUT -p udp -m udp --dport 53 -m owner --gid-owner 1337 -j RETURN
-A OUTPUT -p udp -m udp --dport 53 -j REDIRECT --to-ports 15053
-A ISTIO_INBOUND -p tcp -m tcp --dport 15008 -j RETURN
-A ISTIO_INBOUND -p tcp -m tcp --dport 22 -j RETURN
-A ISTIO_INBOUND -p tcp -m tcp --dport 9090 -j RETURN
-A ISTIO_INBOUND -p tcp -m tcp --dport 12020 -j RETURN
-A ISTIO_INBOUND -p tcp -m tcp --dport 8083 -j RETURN
-A ISTIO_INBOUND -p tcp -m tcp --dport 15020 -j RETURN
-A ISTIO_INBOUND -p tcp -m tcp --dport 15021 -j RETURN
-A ISTIO_INBOUND -p tcp -m tcp --dport 15090 -j RETURN
-A ISTIO_INBOUND -p tcp -j ISTIO_IN_REDIRECT
-A ISTIO_IN_REDIRECT -p tcp -j REDIRECT --to-ports 15006
-A ISTIO_OUTPUT -p tcp -m tcp --dport 443 -j RETURN
-A ISTIO_OUTPUT -p tcp -m tcp --dport 80 -j RETURN
-A ISTIO_OUTPUT -s 127.0.0.6/32 -o lo -j RETURN
-A ISTIO_OUTPUT ! -d 127.0.0.1/32 -o lo -p tcp -m tcp ! --dport 53 -m owner --uid-owner 1337 -j ISTIO_IN_REDIRECT
-A ISTIO_OUTPUT -o lo -p tcp -m tcp ! --dport 53 -m owner ! --uid-owner 1337 -j RETURN
-A ISTIO_OUTPUT -m owner --uid-owner 1337 -j RETURN
-A ISTIO_OUTPUT ! -d 127.0.0.1/32 -o lo -m owner --gid-owner 1337 -j ISTIO_IN_REDIRECT
-A ISTIO_OUTPUT -o lo -p tcp -m tcp ! --dport 53 -m owner ! --gid-owner 1337 -j RETURN
-A ISTIO_OUTPUT -m owner --gid-owner 1337 -j RETURN
-A ISTIO_OUTPUT -p tcp -m tcp --dport 53 -j REDIRECT --to-ports 15053
-A ISTIO_OUTPUT -d 127.0.0.1/32 -j RETURN
-A ISTIO_OUTPUT -j ISTIO_REDIRECT
-A ISTIO_REDIRECT -p tcp -j REDIRECT --to-ports 15001
COMMIT
# Completed on Wed Dec 22 02:53:04 2021
```

**Comment by shonecyx** [comment]
@lambdai 
> * Is the long tcp connection a potential cause of the issue or a break http transaction is more damaging?

We do have a global DestinationRule for TCP keepalive as below. Here the problematic connection is between ingressgateway and backend sidecar and there is a connection pool with tcp keepalive. Originally we suspect it was caused by the tcp keepalive but it turned out to be irrelevant.  Also we tried to remove the sidecar then no RST observed during the testing. So I don't think long tcp connection could cause this issue but something introduced by the sidecar/iptables with the out of order packet, without the sidecar/iptables, those OfO packets won't cause trouble. And I think this problem has been there long time even without high throughput we could also observe RST from the sidecar.
```
  trafficPolicy:
    connectionPool:
      http:
        h2UpgradePolicy: DEFAULT
        http1MaxPendingRequests: 20000
        http2MaxRequests: 20000
        idleTimeout: 60s
        maxRequestsPerConnection: 0
        maxRetries: 20000
      tcp:
        connectTimeout: 1s
        maxConnections: 20000
        tcpKeepalive:
          interval: 10s
          probes: 10
          time: 700s
```

> * Kind of duplication to my first question. Do you see the reset packet only at the beginning of the tcp stream, or it's possible all over the life of the tcp request?

The problematic connection is in the sidecar connection pool to upstream(backend application), every time envoy will pick up a connection from the connection pool and talk to backend. The reset is not at the beginning of the tcp stream since we didn't see the 3-way handshake but it's in the middle of the tcp transmitting packet. If there is an out of window packet during the transmission, the tcp connection will be reset and need client to retry the request.

File: tools/istio-iptables/pkg/capture/run.go (hunk 2 above)
Comment (shonecyx) [reply tools/istio-iptables/pkg/capture/run.go]: BTW we are using `sidecar.istio.io/interceptionMode: REDIRECT`, seems in this mode the invalid drop it not supported in the ISTIO_INBOUND chain.

Comment (shonecyx) [reply tools/istio-iptables/pkg/capture/run.go]: The ISTIOINBOUND chain is in `nat` table with REDIRECT mode and the `nat` table is not intended for filtering, the use of DROP is therefore inhibited. So we'll need to add this rule in PREROUTING chain,
```
		if cfg.cfg.InboundInterceptionMode == constants.TPROXY {
			// When using TPROXY, create a new chain for routing all inbound traffic to
			// Envoy. Any packet entering this chain gets marked with the ${INBOUND_TPROXY_MARK} mark,
			// so that they get routed to the loopback interface in order to get redirected to Envoy.
			// In the ISTIOINBOUND chain, '-j ISTIODIVERT' reroutes to the loopback
			// interface.
			// Mark all inbound packets.
			cfg.iptables.AppendRule(iptableslog.UndefinedCommand, constants.ISTIODIVERT, constants.MANGLE, "-j", constants.MARK, "--set-mark",
				cfg.cfg.InboundTProxyMark)
			cfg.iptables.AppendRule(iptableslog.UndefinedCommand, constants.ISTIODIVERT, constants.MANGLE, "-j", constants.ACCEPT)

			// Create a new chain for redirecting inbound traffic to the common Envoy
			// port.
			// In the ISTIOINBOUND chain, '-j RETURN' bypasses Envoy and
			// '-j ISTIOTPROXY' redirects to Envoy.
			cfg.iptables.AppendVersionedRule("127.0.0.1/32", "::1/128", iptableslog.UndefinedCommand,
				constants.ISTIOTPROXY, constants.MANGLE, "!", "-d", constants.IPVersionSpecific,
				"-p", constants.TCP, "-j", constants.TPROXY,
				"--tproxy-mark", cfg.cfg.InboundTProxyMark+"/0xffffffff", "--on-port", cfg.cfg.InboundCapturePort)
			table = constants.MANGLE
		} else {
			table = constants.NAT
		}
		cfg.iptables.AppendRule(iptableslog.JumpInbound, constants.PREROUTING, table, "-p", constants.TCP,
			"-j", constants.ISTIOINBOUND)
```

**Comment by shonecyx** [comment]
> I am a bit concerned about impacts on stability from this. It seems reasonable but this is the most critical code in Istio and has hard to understand behavior. In particular, users on many platforms may already set nf_conntrack_tcp_be_liberal which the doc listed suggests is a suitable alternative. I worry we will destabilize these users, who are not impacted by this bug. Maybe we should have it in a feature flag for now?

@howardjohn  I'm fine to add a feature flag for it. BTW as I explained above, this drop policy need to be added into PREROUTING chain since in REDIRECT mode ISTIOINBOUND chain is in `nat` table and doesn't support filtering. Another thing I would like to highlight is that the intermittent reset due to out of window packet is introduced by sidecar/iptables, we tried to disable the sidecar injection and ran same test, the reset was gone. That means if there is no sidecar/iptables, the out of window packet won’t cause trouble since no DNAT needed. I believe this issue has been in upstream for a long time(like the kube-proxy one) and client side retry helped tolerate this issue.

**Comment by mandarjog** [comment]
@shonecyx what is the mode of failure if the new rule is hit when we did it intend it to?  Since the new rule only applies to invalid packets, are there any situations when we do want an invalid packet to get back all the way to the client?  ( i don’t see a specific risk based on the description)

**Comment by shonecyx** [comment]
> @shonecyx what is the mode of failure if the new rule is hit when we did not intend it to? Since the new rule only applies to invalid packets, are there any situations when we do want an invalid packet to get back all the way to the client? ( i don’t see a specific risk based on the description)

@mandarjog  Currently the behavior is that sidecar will reset the connection since iptables won't DNAT the out of window packet, which means even without this invalid drop rule, the invalid packet still can not get to the backend application, nor the client. Another trick we can do is to set nf_conntrack_tcp_be_liberal to 1 which means iptables will still DNAT the out of window packet but it still can't get to the backend application nor client since it will be dropped by kernel.

In other words, the invalid packet won't get to application/client with or without this invalid drop policy, the difference is that the sidecar will reset the connection then the client have to retry or just drop the packet and leverage on TCP retransmission and it will be transparent to client.

**Comment by mandarjog** [comment]
Relying on tcp retransmission seems fine. So far it looks like this fix is unambiguously better than before, do you agree? @howardjohn is there a scenario where this fix makes things worse, so that we need to be careful?
Basically, is it a feature or a bug fix.

**Comment by shonecyx** [comment]
@mandarjog @howardjohn 
I just tested out this feature in our Production to reproduce and verify the fix and here is the result.

Basically we have two options for the invalid drop:
Option 1: Directly add the invalid drop to PREROUTING chain in mangle table, the iptables cmd as below:
```
iptables -t mangle -A ISTIO_INBOUND   -m conntrack --ctstate INVALID -j DROP
```
Option 2: Create a new chain in mangle and add the invalid drop, the iptables cmd as below. The third command is mandatory otherwise there is no reference to this new chain:
```
iptables -t mangle -N ISTIO_INBOUND
iptables -t mangle -A ISTIO_INBOUND   -m conntrack --ctstate INVALID -j DROP
iptables -t mangle -A PREROUTING -p tcp -j ISTIO_INBOUND 
```
And I would like to share the test result. For each test, there is a for loop to upload a 300MB file 50 times through IngressGateway, the request reset number out of the 50 ranges from 1 to 13. For example, you can see there are 2 upstream connect error out of 27 requests.
```
for i in {1..50}; do  curl -k --location --request PUT  https://ingressgateway.example.com:8443/reproduce503/round1/random300M_1 --data-binary @/etc/istio/proxy/tmp.300M --resolve ingressgateway.example.com:8443:10.0.0.1 ; echo "done $i"; done
done 1
done 2
done 3
done 4
done 5
done 6
done 7
done 8
done 9
done 10
done 11
done 12
done 13
done 14
done 15
done 16
done 17
upstream connect error or disconnect/reset before headers. reset reason: connection terminationdone 18
done 19
upstream connect error or disconnect/reset before headers. reset reason: connection terminationdone 20
done 21
done 22
done 23
done 24
done 25
done 26
done 27
```

Here is one test result for **option 1**, we can see there are 8 invalid drops. 
```
Chain PREROUTING (policy ACCEPT 4637K packets, 18G bytes)
 pkts bytes target     prot opt in     out     source               destination
    8  154K DROP       all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate INVALID

Chain INPUT (policy ACCEPT 4637K packets, 18G bytes)
 pkts bytes target     prot opt in     out     source               destination

Chain FORWARD (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination

Chain OUTPUT (policy ACCEPT 1883K packets, 18G bytes)
 pkts bytes target     prot opt in     out     source               destination

Chain POSTROUTING (policy ACCEPT 1883K packets, 18G bytes)
 pkts bytes target     prot opt in     out     source               destination
```

Here is one test result for **option 2**, we can see there are 13 invalid drops. 
```
iptables -L -n -v -t mangle  -w
Chain PREROUTING (policy ACCEPT 8163K packets, 33G bytes)
 pkts bytes target     prot opt in     out     source               destination
8163K   33G ISTIO_INBOUND  tcp  --  *      *       0.0.0.0/0            0.0.0.0/0

Chain INPUT (policy ACCEPT 8163K packets, 33G bytes)
 pkts bytes target     prot opt in     out     source               destination

Chain FORWARD (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination

Chain OUTPUT (policy ACCEPT 3274K packets, 33G bytes)
 pkts bytes target     prot opt in     out     source               destination

Chain POSTROUTING (policy ACCEPT 3274K packets, 33G bytes)
 pkts bytes target     prot opt in     out     source               destination

Chain ISTIO_INBOUND (1 references)
 pkts bytes target     prot opt in     out     source               destination
   13  389K DROP       all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate INVALID
```
So both options are working and I would prefer option 1 because no extra chain added, the performance should be better

**Review by howardjohn — APPROVED** [review]

File: tools/istio-iptables/pkg/capture/run.go [hunk 3]
```diff
@@ -131,7 +131,8 @@ func (cfg *IptablesConfigurator) logConfig() {
 	b.WriteString(fmt.Sprintf("ISTIO_EXCLUDE_INTERFACES=%s\n", os.Getenv("ISTIO_EXCLUDE_INTERFACES")))
 	b.WriteString(fmt.Sprintf("ISTIO_SERVICE_CIDR=%s\n", os.Getenv("ISTIO_SERVICE_CIDR")))
 	b.WriteString(fmt.Sprintf("ISTIO_SERVICE_EXCLUDE_CIDR=%s\n", os.Getenv("ISTIO_SERVICE_EXCLUDE_CIDR")))
-	b.WriteString(fmt.Sprintf("ISTIO_META_DNS_CAPTURE=%s", os.Getenv("ISTIO_META_DNS_CAPTURE")))
+	b.WriteString(fmt.Sprintf("ISTIO_META_DNS_CAPTURE=%s\n", os.Getenv("ISTIO_META_DNS_CAPTURE")))
+	b.WriteString(fmt.Sprintf("ISTIO_META_INVALID_DROP=%s\n", os.Getenv("ISTIO_META_INVALID_DROP")))
```
hzxuzhonghu, line 135: maybe better to remove ISTIO_META? so we do not need to send it to pilot

shonecyx: @hzxuzhonghu  What's the downside to add ISTIO_META prefix? It seems it's used to pass env vars as node metadata so we can get it in the pod environment so it will be straightforward to check in the sidecar if this flag is enabled or not, IMHO that's a good thing.
```
istio-proxy@swissknife-test-1:/$ echo $ISTIO_META_INVALID_DROP
true
```

hzxuzhonghu: I mean we donot want to send this value via node metadata to pilot

dgn:
any chance this could be backported to 1.12?

tools/istio-iptables/pkg/capture/run.go
shonecyx: Ack, let me have another PR to remove the ISTIO_META prefix  @hzxuzhonghu

shonecyx:
> any chance this could be backported to 1.12?
@dgn  I'll backport to 1.12 once this PR gets merged https://github.com/istio/istio/pull/36667/files

tools/istio-iptables/pkg/capture/run.go
shonecyx: Removed the prefix in PR https://github.com/istio/istio/pull/36667

shonecyx:
@Kmoneal  Thanks for cherry-pick, please help do the auto cherry-pick for https://github.com/istio/istio/pull/36667,  this feature requires both PRs.

samene:
Can this behavior still happen when `nf_conntrack_tcp_be_liberal` is already set to 1?

I have it set to 1 but see same behavior when uploading a big file to a pod via ingress gateway. Get 503. Istio 1.10.3

on ingress gateway
```
2022-03-21T05:18:01.299568Z	debug	envoy http	[C774][S10992807495329005775] Read-enabling downstream stream due to filter callbacks.
2022-03-21T05:18:01.303297Z	debug	envoy connection	[C737] onAboveWriteBufferHighWatermark
2022-03-21T05:18:01.303316Z	debug	envoy http	[C774][S10992807495329005775] Read-disabling downstream stream due to filter callbacks.
2022-03-21T05:18:01.304859Z	debug	envoy connection	[C737] onBelowWriteBufferLowWatermark
2022-03-21T05:18:01.304870Z	debug	envoy http	[C774][S10992807495329005775] Read-enabling downstream stream due to filter callbacks.
2022-03-21T05:18:01.306479Z	debug	envoy connection	[C737] closing socket: 0
2022-03-21T05:18:01.306490Z	debug	envoy connection	[C737] SSL shutdown: rc=-1
2022-03-21T05:18:01.306515Z	debug	envoy client	[C737] disconnect. resetting 1 pending requests
2022-03-21T05:18:01.306520Z	debug	envoy client	[C737] request reset
2022-03-21T05:18:01.306528Z	debug	envoy router	[C774][S10992807495329005775] upstream reset: reset reason: connection termination, transport failure reason: 
2022-03-21T05:18:01.306643Z	debug	envoy http	[C774][S10992807495329005775] Sending local reply with details upstream_reset_before_response_started{connection termination}
2022-03-21T05:18:01.306704Z	debug	envoy http	[C774][S10992807495329005775] encoding headers via codec (end_stream=false):
':status', '503'
'content-length', '95'
'content-type', 'text/plain'
'date', 'Mon, 21 Mar 2022 05:18:01 GMT'
'server', 'istio-envoy'
'connection', 'close'

2022-03-21T05:18:01.306752Z	debug	envoy http	[C774][S10992807495329005775] doEndStream() resetting stream
2022-03-21T05:18:01.306755Z	debug	envoy http	[C774][S10992807495329005775] stream reset
```

and on the pod envoy

```
2022-03-21T05:18:01.312224Z	debug	envoy http	[C23][S1317549543541473321] Read-disabling downstream stream due to filter callbacks.
2022-03-21T05:18:01.317552Z	debug	envoy connection	[C24] onBelowWriteBufferLowWatermark
2022-03-21T05:18:01.317567Z	debug	envoy http	[C23][S1317549543541473321] Read-enabling downstream stream due to filter callbacks.
2022-03-21T05:18:01.319353Z	debug	envoy connection	[C24] onAboveWriteBufferHighWatermark
2022-03-21T05:18:01.319364Z	debug	envoy http	[C23][S1317549543541473321] Read-disabling downstream stream due to filter callbacks.
2022-03-21T05:18:01.319373Z	debug	envoy connection	[C23] remote close
2022-03-21T05:18:01.319376Z	debug	envoy connection	[C23] closing socket: 0
2022-03-21T05:18:01.319386Z	debug	envoy connection	[C23] SSL shutdown: rc=-1
2022-03-21T05:18:01.319406Z	debug	envoy http	[C23][S1317549543541473321] stream reset
```

shonecyx:
Where did you set `nf_conntrack_tcp_be_liberal` to 1? It should be in upstream pod network namespace and it should work as we tested before.

samene:
Actually I also tried setting `INVALID_DROP` to true in meshConfig.defaultConfig.proxyMetadata but I didn't see the iptables entry. When I changed it to `DROP_INVALID` I was able to see it. Is there some mismatch somewhere in the env variable? Because the doc states it as `INVALID_DROP`
This is istio 1.13.2

with DROP_INVALID :
```
> kubectl logs -n xxx xxx -c istio-init | grep conntrack
-A PREROUTING -m conntrack --ctstate INVALID -j DROP
-A PREROUTING -m conntrack --ctstate INVALID -j DROP
```

and the same with INVALID_DROP
```
> kubectl logs -f -n xxx xxx -c istio-init | grep conntrack
```

This is the logs of istio-init with INVALID_DROP true

```
2022-03-24T17:39:58.365220Z     info    Istio iptables environment:
ENVOY_PORT=
INBOUND_CAPTURE_PORT=
ISTIO_INBOUND_INTERCEPTION_MODE=
ISTIO_INBOUND_TPROXY_ROUTE_TABLE=
ISTIO_INBOUND_PORTS=
ISTIO_OUTBOUND_PORTS=
ISTIO_LOCAL_EXCLUDE_PORTS=
ISTIO_EXCLUDE_INTERFACES=
ISTIO_SERVICE_CIDR=
ISTIO_SERVICE_EXCLUDE_CIDR=
ISTIO_META_DNS_CAPTURE=true
INVALID_DROP=true

2022-03-24T17:39:58.365278Z     info    Istio iptables variables:
PROXY_PORT=15001
PROXY_INBOUND_CAPTURE_PORT=15006
PROXY_TUNNEL_PORT=15008
PROXY_UID=1337
PROXY_GID=1337
INBOUND_INTERCEPTION_MODE=REDIRECT
INBOUND_TPROXY_MARK=1337
INBOUND_TPROXY_ROUTE_TABLE=133
INBOUND_PORTS_INCLUDE=*
INBOUND_PORTS_EXCLUDE=15090,15021,15020
OUTBOUND_IP_RANGES_INCLUDE=*
OUTBOUND_IP_RANGES_EXCLUDE=10.43.0.1/32
OUTBOUND_PORTS_INCLUDE=
OUTBOUND_PORTS_EXCLUDE=
KUBE_VIRT_INTERFACES=
ENABLE_INBOUND_IPV6=false
DNS_CAPTURE=true
DROP_INVALID=false
CAPTURE_ALL_DNS=false
DNS_SERVERS=[10.43.0.10],[]
OUTPUT_PATH=
NETWORK_NAMESPACE=
CNI_MODE=false
EXCLUDE_INTERFACES=

2022-03-24T17:39:58.365474Z     info    Writing following contents to rules file: /tmp/iptables-rules-1648143598365342644.txt1768066335
* raw
-A OUTPUT -p udp --dport 53 -m owner --uid-owner 1337 -j CT --zone 1
-A OUTPUT -p udp --sport 15053 -m owner --uid-owner 1337 -j CT --zone 2
-A OUTPUT -p udp --dport 53 -m owner --gid-owner 1337 -j CT --zone 1
-A OUTPUT -p udp --sport 15053 -m owner --gid-owner 1337 -j CT --zone 2
-A OUTPUT -p udp --dport 53 -d 10.43.0.10/32 -j CT --zone 2
-A PREROUTING -p udp --sport 53 -d 10.43.0.10/32 -j CT --zone 1
COMMIT
* nat
-N ISTIO_INBOUND
-N ISTIO_REDIRECT
-N ISTIO_IN_REDIRECT
-N ISTIO_OUTPUT
-A ISTIO_INBOUND -p tcp --dport 15008 -j RETURN
-A ISTIO_REDIRECT -p tcp -j REDIRECT --to-ports 15001
-A ISTIO_IN_REDIRECT -p tcp -j REDIRECT --to-ports 15006
-A PREROUTING -p tcp -j ISTIO_INBOUND
-A ISTIO_INBOUND -p tcp --dport 15090 -j RETURN
-A ISTIO_INBOUND -p tcp --dport 15021 -j RETURN
-A ISTIO_INBOUND -p tcp --dport 15020 -j RETURN
-A ISTIO_INBOUND -p tcp -j ISTIO_IN_REDIRECT
-A OUTPUT -p tcp -j ISTIO_OUTPUT
-A ISTIO_OUTPUT -o lo -s 127.0.0.6/32 -j RETURN
-A ISTIO_OUTPUT -o lo ! -d 127.0.0.1/32 -p tcp ! --dport 53 -m owner --uid-owner 1337 -j ISTIO_IN_REDIRECT
-A ISTIO_OUTPUT -o lo -p tcp ! --dport 53 -m owner ! --uid-owner 1337 -j RETURN
-A ISTIO_OUTPUT -m owner --uid-owner 1337 -j RETURN
-A ISTIO_OUTPUT -o lo ! -d 127.0.0.1/32 -m owner --gid-owner 1337 -j ISTIO_IN_REDIRECT
-A ISTIO_OUTPUT -o lo -p tcp ! --dport 53 -m owner ! --gid-owner 1337 -j RETURN
-A ISTIO_OUTPUT -m owner --gid-owner 1337 -j RETURN
-A ISTIO_OUTPUT -p tcp --dport 53 -d 10.43.0.10/32 -j REDIRECT --to-ports 15053
-A ISTIO_OUTPUT -d 127.0.0.1/32 -j RETURN
-A ISTIO_OUTPUT -d 10.43.0.1/32 -j RETURN
-A ISTIO_OUTPUT -j ISTIO_REDIRECT
-A OUTPUT -p udp --dport 53 -m owner --uid-owner 1337 -j RETURN
-A OUTPUT -p udp --dport 53 -m owner --gid-owner 1337 -j RETURN
-A OUTPUT -p udp --dport 53 -d 10.43.0.10/32 -j REDIRECT --to-port 15053
COMMIT
2022-03-24T17:39:58.365512Z     info    Running command: iptables-restore --noflush /tmp/iptables-rules-1648143598365342644.txt1768066335
2022-03-24T17:39:58.371161Z     info    Writing following contents to rules file: /tmp/ip6tables-rules-1648143598371119464.txt908361801
```
In the first few lines I see both
```
INVALID_DROP=true
DROP_INVALID=false
```

shonecyx:
We are using istio-cni in our production environment using the flag `INVALID_DROP` and the invalid drop rule is there. Are you using istio-cni or sidecar injection?  Could be a bug(inconsistent flag name) in the sidecar injection flow here https://github.com/istio/istio/blob/master/tools/istio-iptables/pkg/config/config.go#L55 and need to change the flag to `INVALID_DROP` too.
```
apiVersion: v1
data:
  mesh: |-
    accessLogEncoding: JSON
    accessLogFile: /var/log/proxy/access.log
    defaultConfig:
      concurrency: 20
      holdApplicationUntilProxyStarts: true
      proxyMetadata:
        INVALID_DROP: "true"
```

anannaya:
We are seeing the 503 issue , even after adding 'DROP_INVALID and INVALID_DROP` both .

huanghuangzym:
will this backport to istio old release?

hzxuzhonghu:
Older releases have EOL

Issue #27425, opened 2020-09-19
Implement “REPLACE" operation in the Envoyfilter API


zhaohuabing:
**Describe the feature request**

We want to use EnvoyFIlter API to add a Redis Proxy network filter in the listener, but it conflicts with the existing TCP proxy. Adding a "REPLACE" operation in the Envoyfilter API would solve this problem.

We can't use the Istio build-in Redis xDS implementation because it's too simple to support our Redis use case such as read policy and mirroring.

**Describe alternatives you've considered**

[ ] Docs
[ ] Installation
[*] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Additional context**

howardjohn:
#27426

hzxuzhonghu:
@zhaohuabing What's the background of implementing Replace? Especially only for `NETWORK_FILTER` and `NETWORK_FILTER`, isn't `Remove` & `Add` satisfying you?

zhaohuabing:
@hzxuzhonghu The original reason is the one in the description of this issue:

We want to use EnvoyFIlter API to add a Redis Proxy network filter in the listener, but it conflicts with the existing TCP proxy. Adding a "REPLACE" operation in the Envoyfilter API would solve this problem.

Basically, sometimes we need to replace the TCP proxy in the listener with a more specific filter.

hzxuzhonghu:
Yeah, i understand it. What i am confused is can `Remove` and `Add` do that too?

zhaohuabing:
The semantic of "Remove & Add" is not exactly the same as "Replace", because order matters in a filter chain.

hzxuzhonghu:
Got it, how about InsertBefore/After

zhaohuabing:
Might work. But then the "Replace" logic would depend on other filters in the filter chain. I prefer "Replace" because it‘s more elegant. Is there any reason we want to eliminate the "Replace" operation?

zhaohuabing:
Currently, Aeraki is heavily using "Replace" operation to replace the default TCP filter with other filters to support other layer-7 protocols in Istion.

hzxuzhonghu:
Not eliminate, just to be clear about the use case

mandarjog:
@hzxuzhonghu replace is a cleaner version of merge, when you do not want merge semantics. Add-remove makes you handle handling, but replace does not.

PR #36582, opened 2021-12-21
Support REPLACE for virtual host in envoy filter

### LINKED ISSUE #36415: envoyfilter: support REPLACE for VHOST
Currently you can merge but not REPLACE

dddddai:
Fixes #36415

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/envoyfilter/rc_patch_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/envoyfilter/rc_patch_test.go
@@ -808,7 +808,7 @@ func TestReplaceVhost(t *testing.T) {
 		},
 	}
 
-	serviceDiscovery := memory.NewServiceDiscovery(nil)
+	serviceDiscovery := memory.NewServiceDiscovery()
 	env := newTestEnvironment(serviceDiscovery, testMesh, buildEnvoyFilterConfigStore(configPatches))
 	push := model.NewPushContext()
 	push.InitContext(env, nil, nil)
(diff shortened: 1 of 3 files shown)


pilot/pkg/networking/core/v1alpha3/envoyfilter/rc_patch_test.go
```diff
@@ -524,6 +556,40 @@ func TestApplyRouteConfigurationPatches(t *testing.T) {
 			},
 		},
 	}
+	sidecarInboundRCToBeReplaced := &route.RouteConfiguration{
+		Name: "inbound|http|80",
+		VirtualHosts: []*route.VirtualHost{
+			{
```
mandarjog, line 562: Please include configuration elements in the to-be-replaced vhost.
I am not sure if the present test case will fail if you just change the operation to merge instead of replace.

dddddai: Thanks for your review :)
> Please include configuration elements in the to-be-replaced vhost that you do not expect in the output.

Done, and I move this to a separate test to avoid confilcts

hzxuzhonghu:
/wait

Why need replace ops? Remove and add can express same semantic.

hzxuzhonghu:
Original `Replace` proposal https://github.com/istio/istio/issues/27425

dddddai:
> Why need replace ops? Remove and add can express same semantic.

IMHO it sounds kinda weird to use both `REMOVE` and `ADD`, cos they are semantically conflicting, and users may be confused about the operation order
Instead, `REPLACE` is more clear and user-friendly

hzxuzhonghu:
OK, I am asking this because i doubt if we need a general implement for all patch types (not only vhost) we have.

mandarjog:
> OK, I am asking this because i doubt if we need a general implement for all patch types (not only vhost) we have.

REPLACE relies on ordering decisions that are made earlier so it does not need to specify or disturb ordering.

I expect RePlace to be useful anywhere merge is used.

dddddai:
Hello @mandarjog, any comments?

dddddai:
@howardjohn Could you please take a look?

howardjohn (approved):

PR #36602, opened 2021-12-21
gateway: fix bug with managed gateway status and class

howardjohn:
Two bugs:
* `setOnce` was supposed to only set status once. The problem is the
status is always set by Kubernetes by default due to CRD defaulting.
Handle that using the special NotReconciled status reason
* GatewayClass only looks at the *name*, not the *controller*. Update
this to use the same (correct) logic the main code uses.

We definitely need more test coverage here. I am looking into how we can get
defaults for the first case, for the latter we just need more time.
Thanks @robscott for finding these

### DIFF
--- a/pilot/pkg/config/kube/gateway/conditions.go
+++ b/pilot/pkg/config/kube/gateway/conditions.go
@@ -134,8 +134,8 @@ type condition struct {
 	// error defines an error state; the reason and message will be replaced with that of the error and
 	// the status inverted
 	error *ConfigError
-	// setOnce, if enabled, will only set the condition if it is not yet present
-	setOnce bool
+	// setOnce, if enabled, will only set the condition if it is not yet present or set to this reason
+	setOnce string
 }
 
 // setConditions sets the existingConditions with the new conditions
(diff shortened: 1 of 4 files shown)


ramaraochavali (approved):

pilot/pkg/config/kube/gateway/conditions.go
```diff
@@ -134,8 +134,8 @@ type condition struct {
 	// error defines an error state; the reason and message will be replaced with that of the error and
 	// the status inverted
 	error *ConfigError
-	// setOnce, if enabled, will only set the condition if it is not yet present
-	setOnce bool
+	// setOnce, if enabled, will only set the condition if it is not yet present or set to this reason
+	setOnce string
```
hzxuzhonghu, line 138: This is more like a bool var

hzxuzhonghu (commented):
Do you need to add some tests? For those who haven't used k8s gateway it is not clear.

pilot/pkg/config/kube/gateway/conditions.go
howardjohn: Yeah let me rename it. Its kind of a bool+string - it only takes effect if its not `""`

howardjohn:
> Do you need to add some tests? For those who haven't used k8s gateway it is not clear.

Will add some tests next week before merging

Issue #36513, opened 2021-12-14
Implement EnvoyFileAccessLogProvider.logFormat


howardjohn:
https://preliminary.istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-ExtensionProvider-EnvoyFileAccessLogProvider

Currently this is merged in the API but not implemented in Istiod

howardjohn:
@zirain you sent some great PRs in this area recently if you are interested in working on this and/or https://github.com/istio/istio/issues/36514

zirain:
@howardjohn ok, I will take a look at this.

zirain:
@howardjohn are you free to reviews https://github.com/istio/istio/pull/36604?

PR #36604, opened 2021-12-22
Implement EnvoyFileAccessLogProvider.logFormat

zirain:

fix https://github.com/istio/istio/issues/36513


hzxuzhonghu (approved):

ramaraochavali:
@zirain This PR is breaking Istio upgrades for us. 


If we create telemetry CRD with accesslog provider set to "envoy", it used to builds the access log with defaults from mesh config if logFormat is nil which is the case in default provider https://github.com/istio/istio/blob/37a7475427976e38ac11f4584ac3d02889405b0d/pilot/pkg/networking/core/v1alpha3/accesslog.go#L169 and now it defaults to default Istio formats https://github.com/istio/istio/blob/d85a2685bd58c5a7f57a35cd09151202ad763700/pilot/pkg/networking/core/v1alpha3/accesslog.go#L332

So we loose the ability to use default extension provider "Envoy" with mesh config defaults. - Is that intentional change?

I think we should continue with the previous behaviour if there are no issues with it.

Let me know. I can fix if it is not intended change.

PR #36605, opened 2021-12-22
Gracefully delete cluster

hzxuzhonghu:

Alternative of #36310 with minor change


hzxuzhonghu:
cc @stevenctl

stevenctl:
For the most part LGTM. 

We still need to make sure a currently running task on the queue is finished before triggering `handleDelete`. Otherwise we could `handleDelete` (cleanup endpoint shards), then the current task completes, adding to endpointshards and never getting cleaned up. TL;DR `close(stop)` needs to be replaced with something that blocks until we've truly stopped processing queue tasks. 

And we also need the part of #36310 that properly cleans up endpontshards.

stevenctl (approved):
This is still an overall improvement. I'll follow-up with the fixes I mentioned.

PR #36606, opened 2021-12-22
Refactor deleta xds subscribed resource names handling

hzxuzhonghu:

This is a minor improvement, instead of generating all watched resources, it only generates the new subscribed ones.

### DIFF
--- a/pilot/pkg/xds/delta.go
+++ b/pilot/pkg/xds/delta.go
@@ -421,6 +421,14 @@ func (s *DiscoveryServer) pushDeltaXds(con *Connection, push *model.PushContext,
 	}
 	t0 := time.Now()
 
+	// If subscribe is set, client is requesting specific resources. We should just generate the
+	// new resources it needs, rather than the entire set of known resources.
+	if subscribe != nil {
+		w = &model.WatchedResource{
+			ResourceNames: subscribe,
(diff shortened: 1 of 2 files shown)


adiprerepa (approved):

pilot/pkg/xds/delta.go
```diff
@@ -475,7 +468,8 @@ func (s *DiscoveryServer) pushDeltaXds(con *Connection, push *model.PushContext,
 	if len(resp.RemovedResources) > 0 {
 		log.Debugf("ADS:%v %s REMOVE %v", v3.GetShortType(w.TypeUrl), con.ConID, resp.RemovedResources)
 	}
-	if isWildcardTypeURL(w.TypeUrl) {
+	// normally wildcard xds `subscribe` is always nil, just in case there are some extended type not handled correctly.
+	if subscribe == nil && isWildcardTypeURL(w.TypeUrl) {
 		// this is probably a bad idea...
```
howardjohn, line 473: can wildcard ever have non nil subscribe? That seems like it should be impossible..?

hzxuzhonghu: Generally yes, just in case a wrong xds client implement, especially our extended type

Issue #36297, opened 2021-11-30
Support delta xds in adsc


hzxuzhonghu:
ADSC is heavily depended by xds tests, but it only works in sotw mode now, so we donot have any complete ut for delta xds.

howardjohn:
I have a client here: https://github.com/howardjohn/pilot-load/blob/master/adsc/delta.go

It has a lot of bugs :-)

hzxuzhonghu:
Good, do you have bandwidth on it?

howardjohn:
Probably not any time soon

hzxuzhonghu:
Still lack delta in adsc

howardjohn:
I've been working on this 15min here and there for a few months. Not a high priority but if anyone else is thinking about picking this up please talk to me first so we are on the same page

tanujd11:
Hi @howardjohn , I would like to work on this if it is not close to completion. It will be great if you can point me to work that you have done. Thanks.

howardjohn:
https://github.com/howardjohn/istio/tree/adsc/v2-delta. And we basically have a rust version in https://github.com/istio/ztunnel/blob/master/src/xds/client.rs

hzxuzhonghu:
https://github.com/istio/istio/pull/48031

hanxiaop:
> #48031

Didn't realize we had an issue and that someone is already working on it... I noticed the recent delta xDS issue and thought we could support delta adsc for commands. So I was trying to start working on this.

hzxuzhonghu:
@hanxiaop I donot think anyone is working on it now. Just refer to the draft implement from @howardjohn

hanxiaop:
> @hanxiaop I donot think anyone is working on it now. Just refer to the draft implement from @howardjohn

Thanks, I'm going to have a look.

tonya11en:
Is there still anything that needs to be done here after https://github.com/istio/istio/pull/48031 merged?

hanxiaop:
> Is there still anything that needs to be done here after #48031 merged?

Maybe tests that can relay delta adsc. I think this can be closed for now.

PR #36610, opened 2021-12-22
support delta eds, currently only support serviceentry

hzxuzhonghu:

### DIFF
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -391,51 +391,8 @@ func (eds *EdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
 	if !edsNeedsPush(req.ConfigsUpdated) {
 		return nil, model.DefaultXdsLogDetails, nil
 	}
-	var edsUpdatedServices map[string]struct{}
-	if !req.Full {
-		edsUpdatedServices = model.ConfigNamesOfKind(req.ConfigsUpdated, gvk.ServiceEntry)
-	}
-	resources := make(model.Resources, 0)
-	empty := 0
-
-	cached := 0
-	regenerated := 0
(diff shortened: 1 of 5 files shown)


ramaraochavali:
Can you please add a test case for it?

hzxuzhonghu:
I thought it depends on adsc https://github.com/istio/istio/issues/36297, after a  look at delta_test.go, it does not use that package.

hzxuzhonghu:
Will add some cases

howardjohn (commented):
A bit confused what the benefit is. Don't we already basically do this with EDS with `full=false`, even in SotW?

Is this to make full push more efficient?

hzxuzhonghu:
> Is this to make full push more efficient?

Yes

howardjohn:
Thanks, makes sense now. Will take a look soon after I am back from holidays

hzxuzhonghu:
pig @howardjohn

pilot/pkg/xds/eds.go
```diff
@@ -484,3 +441,138 @@ func buildEmptyClusterLoadAssignment(clusterName string) *endpoint.ClusterLoadAs
...
+			cached++
+		} else {
+			// if a service is not found, it means the cluster is removed
+			if builder.service == nil {
+				removed = append(removed, clusterName)
+				continue
+			}
+			l := eds.Server.generateEndpoints(builder)
+			if l == nil {
+				continue
```
howardjohn, line 560: does this mean removed as well?

howardjohn (approved):
LGTM, one comment but can get this in

pilot/pkg/xds/eds.go
hzxuzhonghu: yes, i made this change also to suite stow.

PR #36613, opened 2021-12-22
Add dry-run test for operator init

hanxiaop:
Made the kube client creation function overridable to perform unit tests. Also, I've noticed `TODO` here:
https://github.com/istio/istio/blob/e0207105bb09baf1299a5752b98d83240ba981af/operator/cmd/mesh/operator_test.go#L30
I'll look into this one since the client creation is overridable now.

### DIFF
--- a/operator/cmd/mesh/operator_test.go
+++ b/operator/cmd/mesh/operator_test.go
@@ -20,11 +20,21 @@ import (
 	"strings"
 	"testing"
 
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
 	"istio.io/istio/operator/pkg/util"
 	"istio.io/istio/operator/pkg/util/clog"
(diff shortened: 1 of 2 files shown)


operator/cmd/mesh/operator_test.go
```diff
@@ -158,3 +168,57 @@ func TestOperatorInit(t *testing.T) {
...
+				args = append(args, "--watchedNamespaces", test.watchedNamespaces)
+			}
+
+			rootCmd := GetRootCmd(args)
+			err := rootCmd.Execute()
+			assert.NoError(t, err)
+
+			actions := extendedClient.Kube().(*fake.Clientset).Actions()
+			for _, action := range actions {
+				if action.GetVerb() != "get" {
```
howardjohn, line 218: list/watch is not a problem. Or does action consider these 'get'?

howardjohn (approved):

operator/cmd/mesh/operator_test.go
monsteredc: Thanks. Fixed in https://github.com/istio/istio/pull/36641

PR #36618, opened 2021-12-23
add additional logging for termination drain logic

ramaraochavali:
Minor log additions
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu (approved):
LGTM. One question:  we only drain inbound listeners for sidecar, if there is connection pool for outbound without idle timeout, it should never becomes 0.

ramaraochavali:
> One question: we only drain inbound listeners for sidecar, if there is connection pool for outbound without idle timeout, it should never becomes 0.

Yes.

PR #36622, opened 2021-12-23
refactor keep alive settings

ramaraochavali:
Keep Alive settings logic checks the same conditions at multiple places. This PR simplifies that logic. There is no functional change.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -594,36 +594,35 @@ type upgradeTuple struct {
 }
 
 func applyTCPKeepalive(mesh *meshconfig.MeshConfig, c *cluster.Cluster, tcp *networking.ConnectionPoolSettings_TCPSettings) {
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (commented):
I think we need test cases covering this refactor.

ramaraochavali:
I looked at existing test cases and they seem sufficient. Do you have any specific test in mind? I can add.

hzxuzhonghu:
You mean `TestApplyTCPKeepalive`? It lacks specify keepalive both in mesh and connectionPool setting

ramaraochavali:
I mean this https://github.com/istio/istio/blob/c017654753cd8f4254abbe0567ff9c728d2034d2/pilot/pkg/networking/core/v1alpha3/cluster_test.go#L621 - It covers those cases

hzxuzhonghu (approved):
LGTM, hard to trace the test cases

PR #36623, opened 2021-12-23
Cleanup agent running

hzxuzhonghu:

### DIFF
--- a/pilot/cmd/pilot-agent/main.go
+++ b/pilot/cmd/pilot-agent/main.go
@@ -158,7 +158,6 @@ var (
 				return err
 			}
 			wait()
-			<-ctx.Done()
 			return nil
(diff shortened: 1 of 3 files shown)


pilot/cmd/pilot-agent/main.go
```diff
@@ -158,6 +158,7 @@ var (
 				return err
 			}
 			wait()
+			<-ctx.Done()
```
howardjohn, line 161: This seems like we will hang forever if `wait()` terminates, and it looks like wait can terminate without ctx.Done

hzxuzhonghu: good catch, i will fix

howardjohn (approved):
nice

PR #15123, opened 2019-06-25
Implement /quitquitquit in pilot-agent to support k8s job exit

hzxuzhonghu:
Please provide a description for what this PR is for.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

Added a `/quitquitquit` to handle quit request from client, which mainly provide a way for k8s jobs to exit.

It only handles  http POST method,  by default pilot-agent listens on 15020 port.

fixes: #https://github.com/istio/istio/issues/15041

PR #36624, opened 2021-12-23
Fix sidecar not exit when job completes

hzxuzhonghu:

This handle quits forcibly as this is called from application container, we donot need to gracefully terminate, not to mention the graceful termination can not work now.

### DIFF
--- a/pilot/cmd/pilot-agent/status/server.go
+++ b/pilot/cmd/pilot-agent/status/server.go
@@ -603,8 +603,8 @@ func (s *Server) handleQuit(w http.ResponseWriter, r *http.Request) {
 	}
 	w.WriteHeader(http.StatusOK)
 	_, _ = w.Write([]byte("OK"))
-	log.Infof("handling %s, notifying pilot-agent to exit", quitPath)
-	notifyExit()
+	log.Infof("handling %s, exiting...", quitPath)
+	go os.Exit(0)
 }
 
 func (s *Server) handleAppProbe(w http.ResponseWriter, req *http.Request) {
--- a/pilot/cmd/pilot-agent/status/server.go
+++ b/pilot/cmd/pilot-agent/status/server.go
@@ -134,6 +134,7 @@ type Server struct {
 	envoyStatsPort        int
 	fetchDNS              func() *dnsProto.NameTable
 	upstreamLocalAddress  *net.TCPAddr
+	stop                  chan struct{}
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


hzxuzhonghu:
@ramaraochavali

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -603,8 +603,8 @@ func (s *Server) handleQuit(w http.ResponseWriter, r *http.Request) {
 	}
 	w.WriteHeader(http.StatusOK)
 	_, _ = w.Write([]byte("OK"))
-	log.Infof("handling %s, notifying pilot-agent to exit", quitPath)
-	notifyExit()
+	log.Infof("handling %s, exiting...", quitPath)
+	go os.Exit(0)
```
hzxuzhonghu, line 607: I need to start a new routine, otherwise the handler will have no chance to return

ramaraochavali:
> we donot need to gracefully terminate, not to mention the graceful termination can not work now.

But /quitquitquit can be called for other use cases where we need graceful termination and it works now. But I agree that graceful termination is not needed for jobs but we can distinguish that is job which called it?

hzxuzhonghu:
For long-running pod, it will not call /quitquitquit

ramaraochavali:
So if it will not call /quitquitquit when will this code gets executed?

hzxuzhonghu:
The long running pod use k8s pod graceful shutdown mechanism, user app is not even aware of this interface

ramaraochavali:
> The long running pod use k8s pod graceful shutdown mechanism

That is handled today. Why do we need to terminate it forcefully? Sorry. I may not be understanding the intent.

hzxuzhonghu:
`handleQuit` is for k8s job, it requires calling it explicitly before application returns. 

For pods created by deployment, user will no need to be aware of this api.

hzxuzhonghu:
#15123

ramaraochavali:
> Previously, with this job can exit successfully, but sometime later we added active connection check, which breaks the job container exit.

Active Connection check is not on by default. It needs to be enabled using this `EXIT_ON_ZERO_ACTIVE_CONNECTIONS` this metadata. And jobs can exit successfully as earlier by default now and that flag can be enabled only for regular sidecar apps.

hzxuzhonghu:
Whatever it is a regression, and  active connection check can not work in most cases

PR #36628, opened 2021-12-24
During SyncAll(), do syncDiscoveryNamespaces before sync pods,services,endpoints

andrewshaoyu:
Before sync Services,pods,endpoints, discoveryNamespacesFilter.discoveryNamespaces should be updated first. Otherwise 
if we have many namespaces,some resources may be ignored when namespace have not been inserted into discoveryNamespacesFilter.discoveryNamespaces.

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go
+++ b/pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go
@@ -156,16 +156,17 @@ func (d *discoveryNamespacesFilter) SyncNamespaces() error {
 		return err
 	}
 
+	// omitting discoverySelectors indicates discovering all namespaces
+	if len(d.discoverySelectors) == 0 {
+		for _, ns := range namespaceList {
+			newDiscoveryNamespaces.Insert(ns.Name)
+		}
+	}
+
 	// range over all namespaces to get discovery namespaces
 	for _, ns := range namespaceList {
 		for _, selector := range d.discoverySelectors {
 			if selector.Matches(labels.Set(ns.Labels)) {
-				newDiscoveryNamespaces.Insert(ns.Name)
-			}
-		}
-		// omitting discoverySelectors indicates discovering all namespaces
-		if len(d.discoverySelectors) == 0 {
-			for _, ns := range namespaceList {
 				newDiscoveryNamespaces.Insert(ns.Name)
 			}
 		}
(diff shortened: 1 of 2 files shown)


pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go
```diff
@@ -142,6 +144,39 @@ func (d *discoveryNamespacesFilter) SelectorsChanged(
...
+	for _, ns := range namespaceList {
+		for _, selector := range d.discoverySelectors {
+			if selector.Matches(labels.Set(ns.Labels)) {
+				newDiscoveryNamespaces.Insert(ns.Name)
+			}
+		}
+		// omitting discoverySelectors indicates discovering all namespaces
+		if len(d.discoverySelectors) == 0 {
+			for _, ns := range namespaceList {
+				newDiscoveryNamespaces.Insert(ns.Name)
```
hzxuzhonghu, line 169: can you move this section above

andrewshaoyu: done

hzxuzhonghu (approved):

pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go
```diff
@@ -142,6 +144,40 @@ func (d *discoveryNamespacesFilter) SelectorsChanged(
...
+	for _, ns := range namespaceList {
+		for _, selector := range d.discoverySelectors {
+			if selector.Matches(labels.Set(ns.Labels)) {
+				newDiscoveryNamespaces.Insert(ns.Name)
+			}
+		}
+	}
+
+	// update filter state
+	d.discoveryNamespaces = newDiscoveryNamespaces
```
howardjohn, line 176: can we deduplicate this code with SelectorsChanged? Seems like they are a copy?

andrewshaoyu: SelectorsChanged means discoveryNamespacesFilter's selectors have been changed, it is designed for mesh.watcher.selectors changed. During syncAll, we can use current discoveryNamespacesFilter's selectors, don't need to know where is `selectors`, although we can get from `c.opts.MeshWatcher.Mesh().DiscoverySelectors`

PR #36633, opened 2021-12-25
Sidecar grace shutdown: respect both active connection and graceful drain duration

hzxuzhonghu:

I believe this is better than https://github.com/istio/istio/pull/36624 in case some user depend /quitquitquit in their long-running application

### DIFF
--- a/pilot/cmd/pilot-agent/options/agent.go
+++ b/pilot/cmd/pilot-agent/options/agent.go
@@ -44,6 +44,7 @@ func NewAgentOptions(proxy *model.Proxy, cfg *meshconfig.ProxyConfig) *istioagen
 		ServiceNode:                 proxy.ServiceNode(),
 		EnvoyStatusPort:             envoyStatusPortEnv,
 		EnvoyPrometheusPort:         envoyPrometheusPortEnv,
+		MinimumDrainDuration:        minimumDrainDurationEnv,
 		ExitOnZeroActiveConnections: exitOnZeroActiveConnectionsEnv,
 		Platform:                    platform.Discover(),
 		GRPCBootstrapPath:           grpcBootstrapEnv,
--- a/pilot/cmd/pilot-agent/options/agent.go
+++ b/pilot/cmd/pilot-agent/options/agent.go
@@ -44,7 +44,6 @@ func NewAgentOptions(proxy *model.Proxy, cfg *meshconfig.ProxyConfig) *istioagen
 		ServiceNode:                 proxy.ServiceNode(),
 		EnvoyStatusPort:             envoyStatusPortEnv,
 		EnvoyPrometheusPort:         envoyPrometheusPortEnv,
-		MinimumDrainDuration:        minimumDrainDurationEnv,
 		ExitOnZeroActiveConnections: exitOnZeroActiveConnectionsEnv,
 		Platform:                    platform.Discover(),
 		GRPCBootstrapPath:           grpcBootstrapEnv,
--- a/pkg/envoy/agent.go
+++ b/pkg/envoy/agent.go
@@ -140,22 +140,30 @@ func (a *Agent) terminate() {
 	if e != nil {
 		log.Warnf("Error in invoking drain listeners endpoint %v", e)
 	}
-	// If exitOnZeroActiveConnections is enabled, always sleep minimumDrainDuration then exit
-	// after min(all connections close, terminationGracePeriodSeconds-minimumDrainDuration).
-	// exitOnZeroActiveConnections is disabled (default), retain the existing behavior.
+	// If exitOnZeroActiveConnections is enabled, we will check active connections periodically.
+	// If all connections closed or terminationDrainDuration has passed, proxy will be killed.
+	// If exitOnZeroActiveConnections is disabled (default), retain the existing behavior.
 	if a.exitOnZeroActiveConnections {
 		log.Infof("Agent draining proxy for %v, then waiting for active connections to terminate...", a.minDrainDuration)
 		time.Sleep(a.minDrainDuration)
 		log.Infof("Checking for active connections...")
 		ticker := time.NewTicker(activeConnectionCheckDelay)
-		for range ticker.C {
-			ac := a.activeProxyConnections()
-			if ac == 0 {
-				log.Info("There are no more active connections. terminating proxy...")
+		graceDrainCh := time.After(a.terminationDrainDuration)
+		for {
+			select {
+			case <-graceDrainCh:
+				log.Infof("Graceful termination period complete, terminating remaining proxies.")
 				a.abortCh <- errAbort
 				return
+			case <-ticker.C:
+				ac := a.activeProxyConnections()
+				if ac == 0 {
+					log.Info("There are no more active connections. terminating proxy...")
+					a.abortCh <- errAbort
+					return
+				}
+				log.Infof("There are still %d active connections", ac)
 			}
-			log.Infof("There are still %d active connections", ac)
 		}
 	} else {
 		log.Infof("Graceful termination period is %v, starting...", a.terminationDrainDuration)
@@ -163,7 +171,6 @@ func (a *Agent) terminate() {
 		log.Infof("Graceful termination period complete, terminating remaining proxies.")
 		a.abortCh <- errAbort
 	}
-	log.Warnf("Aborted all epochs")
 }
 
 func (a *Agent) activeProxyConnections() int {
--- a/pkg/config/mesh/mesh.go
+++ b/pkg/config/mesh/mesh.go
@@ -43,7 +43,7 @@ func DefaultProxyConfig() meshconfig.ProxyConfig {
 		ServiceCluster:           constants.ServiceClusterName,
 		DrainDuration:            types.DurationProto(45 * time.Second),
 		ParentShutdownDuration:   types.DurationProto(60 * time.Second),
-		TerminationDrainDuration: types.DurationProto(5 * time.Second),
+		TerminationDrainDuration: types.DurationProto(30 * time.Second),
 		ProxyAdminPort:           15000,
 		Concurrency:              &types.Int32Value{Value: 2},
 		ControlPlaneAuthPolicy:   meshconfig.AuthenticationPolicy_MUTUAL_TLS,
--- a/pkg/istio-agent/agent.go
+++ b/pkg/istio-agent/agent.go
@@ -180,6 +180,7 @@ type AgentOptions struct {
 	// proxy config.
 	EnvoyPrometheusPort int
 
+	MinimumDrainDuration        time.Duration
 	ExitOnZeroActiveConnections bool
 
 	// Cloud platform
--- a/pkg/istio-agent/agent.go
+++ b/pkg/istio-agent/agent.go
@@ -180,8 +180,6 @@ type AgentOptions struct {
 	// proxy config.
 	EnvoyPrometheusPort int
 
-	MinimumDrainDuration time.Duration
-
 	ExitOnZeroActiveConnections bool
 
 	// Cloud platform


ramaraochavali:
LGTM. Will leave it to @howardjohn to take a look

pilot/cmd/pilot-agent/options/agent.go
```diff
@@ -44,7 +44,6 @@ func NewAgentOptions(proxy *model.Proxy, cfg *meshconfig.ProxyConfig) *istioagen
 		ServiceNode:                 proxy.ServiceNode(),
 		EnvoyStatusPort:             envoyStatusPortEnv,
 		EnvoyPrometheusPort:         envoyPrometheusPortEnv,
-		MinimumDrainDuration:        minimumDrainDurationEnv,
```
ramaraochavali, line 47: Please see here on why it was added https://github.com/istio/istio/pull/35059#pullrequestreview-758922774

pkg/envoy/agent.go
```diff
@@ -140,30 +140,37 @@ func (a *Agent) terminate() {
...
 	if a.exitOnZeroActiveConnections {
 		log.Infof("Agent draining proxy for %v, then waiting for active connections to terminate...", a.minDrainDuration)
 		time.Sleep(a.minDrainDuration)
 		log.Infof("Checking for active connections...")
 		ticker := time.NewTicker(activeConnectionCheckDelay)
-		for range ticker.C {
-			ac := a.activeProxyConnections()
-			if ac == 0 {
-				log.Info("There are no more active connections. terminating proxy...")
+		graceDrainCh := time.After(a.terminationDrainDuration)
```
ramaraochavali, line 151: Please refer to comments here https://github.com/istio/istio/pull/36534#discussion_r770739087 on why it was not done. Since you are changing the default terminationDrainDuration value it may be fine - but I am not sure changing the default value (you are setting it to right value though) is the right thing to do wrt BC?

hzxuzhonghu:
```
select {
		case <-ticker.C:
			if a.activeProxyConnections() == 0 {
				log.Info("There are no more active connections. terminating proxy...")
				a.abortCh <- errAbort
				return
			}
		case <-time.After(a.terminationDrainDuration):
```

This is not right, i am not sure what you mean

rittneje: I think in order to preserve backwards compatibility, the default value of `terminationDrainDuration` needs to be handled differently in the two cases.
- If `exitOnZeroActiveConnections` is true, then the default `terminationDrainDuration` is (effectively) infinity.
- If `exitOnZeroActiveConnections` is false, then the default `terminationDrainDuration` is 5s.

@ramaraochavali Your original code that you linked had a bug. Since you forget to put a for loop, it would have checked the active connections once after one second, and then given up. I'm assuming what you intended to write is what @hzxuzhonghu has written here, and so the heart of your question is whether `minDrainDuration` should count against `terminationDrainDuration`. Personally I feel that it should not, as this resolves the question around what it means if `terminationDrainDuration` <= `minDrainDuration`.

That said, I don't know under what circumstances you would want `terminationDrainDuration` to be less than the pod's termination grace period when `exitOnZeroActiveConnections` is true, so it's unclear what the benefit of this change is.

howardjohn: > If exitOnZeroActiveConnections is true, then the default terminationDrainDuration is (effectively) infinity.

One concern with this is it means we can essentially never make exitOnZeroActiveConnections default to true, or we will break `terminationDrainDuration` users?

rittneje: This again comes back to the question of why you would even want to specify drain duration with `exitOnZeroActiveConnections` enabled. But you could in theory say that `exitOnZeroActiveConnections` defaults to false if `terminationDrainDuration` is set, and true if it isn't.

hzxuzhonghu: >If exitOnZeroActiveConnections is true, then the default terminationDrainDuration is (effectively) infinity.

I am afraid this may break job termination if the active connections is never gown to 0.

rittneje: I am confused. Why do you need this kind of graceful termination for a job? Can't you just tell istio-proxy to stop immediately once your container is finished (i.e., `exitOnZeroActiveConnections=false`)?

hzxuzhonghu: You are right, this means we can never make `exitOnZeroActiveConnections = true` by default

hzxuzhonghu: If so, i also doubt when `exitOnZeroActiveConnections` is useful

rittneje: > If so, i also doubt when `exitOnZeroActiveConnections` is useful

For a regular pod (not job), both the sidecar and main container receive SIGTERM simultaneously. But if the sidecar exits immediately, then all in-flight requests will fail. So first `terminationDrainDuration` was introduced to artificially keep the the sidecar alive for some time after SIGTERM. But the negative consequence of this is that it always took that long to terminate, even if the main container finished earlier. So then `exitOnZeroActiveConnections` was introduced, to allow it to exit earlier if the connection count reaches zero. But then, there was no real need for `terminationDrainDuration` in that case, because it might as well just wait for the pod termination grace period in the worst case. However, neither of this approaches really make pod termination graceful in all cases, they just make it a little better. This is why I'm hoping they can be replaced with a synchronous mechanism that takes out the guesswork. It should even work for job pods.

> You are right, this means we can never make exitOnZeroActiveConnections = true by default

This is not necessarily true. I assume you are manually calling into the sidecar from your main container at the end? If so, perhaps istio-proxy could make it so that endpoint causes it to exit immediately regardless of these settings.

hzxuzhonghu: I almost agree with you, but why do we pursue termination so quickly?

>This is not necessarily true. I assume you are manually calling into the sidecar from your main container at the end? If so, perhaps istio-proxy could make it so that endpoint causes it to exit immediately regardless of these settings.

This way is not good for me. Users should not be aware of the drain details. And chaing their software to adapt to it.

For job, it is different though.. I guess the synch mechanism you want is something like https://github.com/istio/istio/pull/36624/files#diff-8a331529c387b787b4d84fab3273afb6ce52355c663f4dee58aad15a1e832c18R600-R612

rittneje: > I almost agree with you, but why do we pursue termination so quickly?

I don't understand the question. Can you re-state it?

> This way is not good for me. Users should not be aware of the drain details. And chaing their software to adapt to it.

Ideally this would be the case. And with the `istio-wait` wrapper that @howardjohn touched on [here](https://github.com/istio/istio/issues/36640#issuecomment-1001778681), it would ultimately be completely transparent. For instance, when wrapping a job container, `istio-wait` could automatically signal the sidecar container to exit once your container exits. In fact, my suspicion is all of these other levers (`exitOnZeroActiveConnections `, `terminationDrainDuration`, possibly even `holdApplicationUntilProxyStarts`) will be obsolete if that wrapper is implemented.

However, as things currently stand, Kubernetes itself is unaware that the sidecar container is not important, and the job is not considered completed until all containers exit. So for the time being you will have to explicitly signal the sidecar container when you are done.

hzxuzhonghu: >I almost agree with you, but why do we pursue termination so quickly?

I mean does pod terminate quickly or slowly really matter

For the second issue, I don't think `istio-wait` is more user friendly than a rest api `/quitquitquit`.

rittneje: > I mean does pod terminate quickly or slowly really matter

It matters during rolling updates and node drains, as pods that are slow to terminate will delay those operations, possibly resulting in other issues, or at least developer frustration.

> For the second issue, I don't think istio-wait is more user friendly than a rest api /quitquitquit.

`istio-wait` is not user- or developer-facing. It would be injected automatically by the mutating webhook. From your perspective everything would work the same as if you weren't using Istio at all.

hzxuzhonghu: This is a little hack, it changes user's container, i am not sure if there is a security issue since i am not an security expert.

From what I can tell, our users can not accept changing their YAML.

hzxuzhonghu: Another point I want to share is that for http the connection idle timeout is default 1 hour, so for 99% cases the active connection check will not be satisfied

ramaraochavali:
@rittneje 
>> Your original code that you linked had a bug. Since you forgot to put a for loop, it would have checked the active connections once after one second, and then given up.

No. I do not think so. In our environment jobs did not exit when there are still active connections - https://github.com/istio/istio/blob/7ab13d9a13e338bec31c00fe36f95bdbddc1f640/pkg/envoy/agent.go#L151.

So now coming to actual problem, for regular pods `terminationDrainDuration` is bit redundant with k8s introducing `terminationGracePeriodSeconds`. So the goal of `exitOnActiveConnections` is to exit as soon as active connections become zero with out worrying about `terminationDrainDuration` or `terminationGracePeriodSeconds`

But for Jobs, K8s does not kill after `terminationGracePeriodSeconds`. Ideally we get `terminationGracePeriodSeconds` to pods env using downward API and use it terminate after that period even if active connections are not zero. Tried to do that here https://github.com/istio/istio/pull/36534 - but k8s does not support that field downwardAPI.

One way is to `terminationDrainDuration` change the default to match k8s (this PR) but I do not know if it is a backward compatible change. Second wait for `terminationGracePeriodSeconds` to be exposed as downward API and use it to terminate jobs and till that time add documentation that for Jobs ExitOnActiveConnections should not be enabled.

pkg/envoy/agent.go
rittneje: > From what I can tell, our users can not accept changing their YAML.

It already does change their yaml today. For example, it has to rewrite the ports to go to the sidecar, and it has to rewrite the liveness/readiness probes as well. But note that these changes are applied automatically by the mutation webhook. You are not expected to change the yaml you are applying via `kubectl` (or similar).

> This is a little hack, it changes user's container, i am not sure if there is a security issue since i am not an security expert.

First, note that by its very nature istio-proxy already has the ability the snoop on and potentially even modify all your unencrypted traffic. In other words, you already have to trust all the Istio-provided executables not to be malicious.

Also, there is no solution to this issue that isn't a hack at some level. Adding container dependencies to Kubernetes has been in limbo for years now. The only workaround in the interim is for the two containers to coordinate during shutdown. Arbitrary sleeps cannot work properly for this in general. And having istio-proxy ignore SIGTERM until it hits the pod termination grace period would be a rather poor experience.

> Another point I want to share is that for http the connection idle timeout is default 1 hour, so for 99% cases the active connection check will not be satisfied

I'm not sure what you mean by this. The active connection count is being checked during pod termination, which means all of your connections are going to be closed one way or the other, since the kernel automatically closes all file descriptors when the process exits.

rittneje:
> But for Jobs, K8s does not kill after terminationGracePeriodSeconds.

@ramaraochavali My understanding of the issue is that for job pods, Kubernetes simply waits for all the containers to finish. It does not terminate or signal the other containers when the first one is done. This is why the /quitquitquit endpoint is used, to tell istio-proxy that the main container is finished, in which case it can just exit immediately, because there wouldn't be anything to wait for.

ramaraochavali:
> My understanding of the issue is that for job pods, Kubernetes simply waits for all the containers to finish. It does not terminate or signal the other containers when the first one is done. This is why the /quitquitquit endpoint is used, to tell istio-proxy that the main container is finished, in which case it can just exit immediately, because there wouldn't be anything to wait for.

I agree that we can change /quitquitquit to exit immediately if it is used just for jobs. I am not sure about that. If we assume that it is only used by jobs, I am fine with that.

pkg/envoy/agent.go
hzxuzhonghu: Changing application's running args is far more complex and risky than other corner changes like probe. 

Even with istio-wait, you still face the same issue when to terminate. Same question as what we have now.

Issue #36472, opened 2021-12-10
kube-inject should accept iop file


therealmitchconnors:
when running kube-inject locally, it expects separate files for meshconfig, proxyconfig, etc.  Many gitops users simply store their iop in git, which is capable of generating these files at runtime.  If kube-inject would accept an iop file, the workflow would be much simpler for users.

PR #36634, opened 2021-12-25
Add iop file support to kube-inject

hanxiaop:
Fix https://github.com/istio/istio/issues/36472

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/istioctl/cmd/kubeinject.go
+++ b/istioctl/cmd/kubeinject.go
@@ -419,10 +419,8 @@ func setupKubeInjectParameters(sidecarTemplate *inject.Templates, valuesConfig *
 				return nil, nil, err
 			}
 			*valuesConfig = string(valuesConfigBytes)
-		} else {
-			if *valuesConfig, err = getValuesFromConfigMap(kubeconfig, revision); err != nil {
-				return nil, nil, err
-			}
+		} else if *valuesConfig, err = getValuesFromConfigMap(kubeconfig, revision); err != nil {
+			return nil, nil, err
 		}
 	}
 	return injector, meshConfig, err
@@ -445,6 +443,9 @@ func setConfigsFromIOPIfNotPresent(valuesConfig *string, logger clog.Logger) (*m
 		}
 		if iops.Spec.MeshConfig != nil {
 			meshConfigYaml, err := yaml.Marshal(iops.Spec.MeshConfig)
+			if err != nil {
+				return nil, err
+			}
 			meshConfig, err = mesh.ApplyMeshConfigDefaults(string(meshConfigYaml))
 			if err != nil {
 				return nil, err
--- a/istioctl/cmd/kubeinject.go
+++ b/istioctl/cmd/kubeinject.go
@@ -48,6 +48,8 @@ import (
 	meshconfig "istio.io/api/mesh/v1alpha1"
 	"istio.io/istio/istioctl/pkg/clioptions"
 	"istio.io/istio/istioctl/pkg/tag"
+	"istio.io/istio/operator/pkg/manifest"
+	"istio.io/istio/operator/pkg/util/clog"
 	"istio.io/istio/pkg/config/mesh"
 	"istio.io/istio/pkg/kube"
 	"istio.io/istio/pkg/kube/inject"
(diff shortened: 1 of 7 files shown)


istioctl/cmd/kubeinject.go
```diff
@@ -474,6 +517,8 @@ It's best to do kube-inject when the resource is initially created.
 			}
 			var reader io.Reader
 
+			l := clog.NewConsoleLogger(c.OutOrStdout(), c.ErrOrStderr(), scope)
```
howardjohn, line 520: logging anything seems like it will break `kube-inject | kubectl apply` ?

istioctl/cmd/kubeinject.go
```diff
@@ -391,26 +394,65 @@ func setupKubeInjectParameters(sidecarTemplate *inject.Templates, valuesConfig *
...
 		if err != nil {
-			return nil, nil, err
+			return nil, err
+		}
+		if iops.Spec.Values != nil {
+			values, err := json.Marshal(iops.Spec.Values)
+			if err != nil {
+				return nil, err
+			}
+			*valuesConfig = string(values)
```
howardjohn, line 442: we should return this, not set a pointer value IMO

istioctl/cmd/kubeinject.go
```diff
@@ -391,26 +394,65 @@ func setupKubeInjectParameters(sidecarTemplate *inject.Templates, valuesConfig *
...
-	if valuesFile != "" {
-		valuesConfigBytes, err := os.ReadFile(valuesFile) // nolint: vetshadow
+	return injector, meshConfig, err
+}
+
+// setConfigsFromIOPIfNotPresent sets the configs with configs in IOPs.
+func setConfigsFromIOPIfNotPresent(valuesConfig *string, logger clog.Logger) (*meshconfig.MeshConfig, error) {
+	var meshConfig *meshconfig.MeshConfig
+	if iopFilename != "" {
+		_, iops, err := manifest.GenerateConfig([]string{iopFilename}, nil, false, nil, logger)
```
howardjohn, line 433: GenerateConfig seems like its doing way more than what we want. Can we not just unmarshal it ?

istioctl/cmd/kubeinject_test.go
```diff
@@ -48,6 +48,29 @@ func TestKubeInject(t *testing.T) {
...
+			args: strings.Split(
+				"kube-inject --operatorFileName testdata/istio-operator.yaml"+
+					" --injectConfigFile testdata/inject-config-inline-iop.yaml -f testdata/deployment/hello.yaml",
+				" "),
+			goldenFilename: "testdata/deployment/hello.yaml.iop.injected",
+		},
+		{ // case 6 with iops and values override
+			args: strings.Split(
+				"kube-inject --operatorFileName testdata/istio-operator.yaml"+
+					" --injectConfigFile testdata/inject-config-inline-iop.yaml"+
```
howardjohn, line 68: seems like `injectConfigFile` shouldn't be *required* with operatorFileName? Maybe optional, but it can be derived?

howardjohn (commented):
Nice

istioctl/cmd/kubeinject.go
monsteredc: Removed since `GenerateConfig ` is not used anymore.

istioctl/cmd/kubeinject.go
monsteredc: Done.

monsteredc: Done.

istioctl/cmd/kubeinject_test.go
monsteredc: Yes, `injectConfigFile` is not required. If it is not provided, either `ExternalInjector` will be used or the injection template will be derived from `istio-sidecar-injector` config map. Providing `injectConfigFile` flag will help make the tests' want results become consistent because the templates are always the same.

therealmitchconnors (approved):
Thanks for tackling this!

Issue #36627, opened 2021-12-24
DiscoverySelectors should also work on CRs


hzxuzhonghu:
Earlier this year we introduced discovery selectors to configure namespaces controlled by istio. But for istio's own API which is defined by CRD, this feature does not work. 

As a supplement, we should make DiscoverySelectors work on CRD as well.

Issue #35092, opened 2021-09-09
Discovery Selectors do not apply to Gateway resources


emike922:
### Bug Description

When using discovery selectors, e.g. `--set meshConfig.discoverySelectors[0].matchLabels.name=istio-system`, I would expect istiod to disregard more or less everything happening in non-matching namespaces. But it appears that it still watches for and reads/uses Gateway resources from every namespace.
This could conceivably result in wrong or undesired configuration from an unlabelled namespace getting enabled. Of course one could argue, there should not be any wrong/undesired configuration lying around to begin with, but hey...

```
$ k logs -n istio-system istiod-649b6cfb4b-bn2p9
...
2021-09-09T06:21:47.612062Z     info    mesh configuration updated to: (*v1alpha1.MeshConfig)(0xc0008d4b40)(proxy_listen_port:15001 connect_timeout:<seconds:10 > protocol_detection_timeout:<> ingress_class:"istio" ingress_service:"istio-ingressgateway" ingress_controller_mode:STRICT enable_tracing:true default_config:<config_path:"./etc/istio/proxy" binary_path:"/usr/local/bin/envoy" service_cluster:"istio-proxy" drain_duration:<seconds:45 > parent_shutdown_duration:<seconds:60 > discovery_address:"istiod.istio-system.svc:15012" proxy_admin_port:15000 control_plane_auth_policy:MUTUAL_TLS stat_name_length:189 concurrency:<value:2 > tracing:<zipkin:<address:"zipkin.istio-system:9411" > > status_port:15020 termination_drain_duration:<seconds:5 > > outbound_traffic_policy:<mode:ALLOW_ANY > enable_auto_mtls:<value:true > trust_domain:"cluster.local" default_service_export_to:"*" default_virtual_service_export_to:"*" default_destination_rule_export_to:"*" root_namespace:"istio-system" locality_lb_setting:<enabled:<value:true > > dns_refresh_rate:<seconds:5 > thrift_config:<> enable_prometheus_merge:<value:true > extension_providers:<name:"envoy" envoy_file_access_log:<path:"/dev/stdout" > > default_providers:<access_logging:"envoy" >
  discovery_selectors:<matchLabels:<key:"name" value:"istio-system" >> )

$ istioctl proxy-config listener istio-ingressgateway-7cfff779cc-492mn.istio-system
ADDRESS PORT  MATCH DESTINATION
0.0.0.0 8080  ALL   Route: http.8080 <<<
[...]

Issue #31115, opened 2021-02-26
Multi-Istio deployments in same k8s cluster


l8huang:
**Describe the feature request**

After issue #26679 resolved, istiod can be configured to watch a subset of namespaces in a k8s cluster by using `meshConfig.discoverySelectors`. But for deploying multi-istio in a k8s cluster, there are still some gaps:

- namespace controller injects ConfigMap istio-root-ca-cert to namespaces, need to scoping it to the namespaces which are part of mesh managed by one Istio deployment, otherwise, it would override istio-root-ca-cert injected by another Istiod deployment.
- mutating/validating webhook should be configured to run on an object based on which Istio deployment manages its namespace.
- secrets controller watches all namespaces currently, it should also be configured to avoid trigger XDS push unnecessarily when a Secret doesn't belong to a Istio deployment. This is not mandatory, but better to have.

PR #29802 implemented a DiscoveryNamespacesFilter, which can be reused by namespace controller and secrets controller, but looks like a different set of namespace selectors should be defined, because "discovery - namespaces watched by Istio" and "namespaces in a mesh" are different concept.

Per my understanding, another namespace selector("meshNamespaceSelectors"?) in MeshConfig should be defined and items listed above should be modified accordingly. Please correct me if I misunderstand anything.

**Describe alternatives you've considered**

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

[] Multi Cluster
[ ] Virtual Machine
[X] Multi Control Plane

**Additional context**

liamawhite:
I think you would also need to filter out Istio CRs to avoid cross-contamination. i.e. If I create a service entry with `exportTo: ['*']` in a namespace not managed by a given mesh, it shouldn't appear in Envoys for that mesh. Having spoken to @harveyxia, #29802 doesn't ignore Istio resources created in a given namespace.

kfaseela:
FYI: https://github.com/istio/istio/pull/41198 has been merged

PR #36639, opened 2021-12-27
Support crd discovery selector

hzxuzhonghu:

Fix https://github.com/istio/istio/issues/36627


howardjohn (commented):
This is a breaking change. The API explicitly only applies to discovery, so making it do more will be incompatible.

It would be good to revisit the design doc. I recall we explicitly decided to make it only discovery. If we are changing our minds there we ought to understand why our decisions have changed.

hzxuzhonghu:
The api does not explicitly say it does not limit CR, and also it does not explicitly say it works on it. Will search the design

hzxuzhonghu:
Unfortunately, there is even no mention of crd in the design https://docs.google.com/document/d/1S3GPCUBmTmvh09NmpuMCG3vwDTNkVeRhGT3lbLyA4qs/edit?resourcekey=0-a8iMutRwjQElYN5dlgEDdw#heading=h.h3lxcxfhqndp

SO I think there is no compatibility issue, this can be treated as a feature request

howardjohn:
Regardless of if it's in the doc dramatically changing the functionality of an existing API is incompatible.

Let's take a step back here. I haven't seen any discussion of the motivations for this change. We already have a lot of 'scoping' mechanisms. What problems are we trying to solve here?

hzxuzhonghu:
https://github.com/istio/istio/issues/35092

hzxuzhonghu:
@howardjohn @ramaraochavali What do i need to promote this feature?

howardjohn:
I still don't understand the use case. The issue you linked does not explain *why* they want this.

hzxuzhonghu:
An special case i can think of is in multi tenant case, multi istio can be deployed in a single cluster. We donot want the crd{CR} to interference each other.

hzxuzhonghu:
@howardjohn https://github.com/istio/istio/issues/31115

howardjohn:
I think this is a pretty important topic and should probably have a proper design. At the very least, we cannot change the behavior of the existing field or there will be incompatible behavior change when users of discoverySelector upgrade to this commit

hzxuzhonghu:
I donot think there are many users of discoverySelector who have CR created in namespaces that are not selected. 

For sake, I am fine to add a knob for this.  We can also propose a discussion about it.

emike922:
> Doesn't Sidecar scoping handle this?

I haven't tested this, but the doc for sidecar workload selector (this is what you mean, right?) does not mention scoping for all CRs but only a handful? Does scoping work on all CRs?

howardjohn:
> I haven't tested this, but the doc for sidecar workload selector (this is what you mean, right?) does not mention scoping for all CRs but only a handful? Does scoping work on all CRs?

All the CRs it does *not* scope are already scoped by namespace. For example, if you have a PeerAuthentication if `foobar` namespace, and have no proxies in `foobar`, it does nothing.

emike922:
For us (non-vanilla) this might be a convenient way to avoid configuration leakage between multiple tenants/istiod control planes. Tenant boundaries could be defined in terms of immutable `kubernetes.io/metadata.name` labels, allowing to dynamically extend or shrink the scope of each control plane without too much hassle or risk.

For vanilla users, this might be useful to include or exclude namespaces for configuration discovery, for example to ignore "less trusted" namespaces when operating in a shared environment.
It could also be useful to enable testing of revisions with incompatible or maybe just experimental (globally-scoped) configurations without affecting the productive instance.

kfaseela:
> Doesn't Sidecar scoping handle this?
> 
> In multitenant situations I would imagine its typical to not actually have _permission_ to read from all namespaces?

sidecar scoping solves part of the issue, but what about the gateways?

howardjohn:
@kfaseela Gateways have namespace scoping options already: PILOT_SCOPE_GATEWAY_TO_NAMESPACE or use new Kubernetes Gateway API which has this as default

kfaseela:
> @kfaseela Gateways have namespace scoping options already: PILOT_SCOPE_GATEWAY_TO_NAMESPACE or use new Kubernetes Gateway API which has this as default

Would try this end to end along with sidecar scoping and get back. But wouldn't it simplify the configuration a lot when we have this supported out of the box, while using discovery selectors? Ideally we don't want to configure additional sidecar config to limit the CRD scope.

emike922:
This might also be useful as an option to lock down certain CR discovery settings that **cannot** be overridden by the user (default mesh-wide sidecar w/o workload selector in root namespace is easily overridden by another sidecar if I understand correctly)

SpecialYang:
We have implemented this feature in production but by little dirty way. We can deploy multiple istiod instances in one K8s cluster for tenant case. These istiod instances can not only locate at individual namespaces, but also locate at the same namespace isolated by labels of CR.

Part of code changes like below.

Server.
```
if alifeatures.WatchResourcesByNamespaceForPrimaryCluster != "" {
	informerOptions := kubelib.InformerOptions{
		WatchedNamespace: alifeatures.WatchResourcesByNamespaceForPrimaryCluster,
		Label:            alifeatures.WatchResourcesByLabelForPrimaryCluster,
	}
	s.kubeClient, err = kubelib.NewClientWithInformerOptions(kubelib.NewClientConfigForRestConfig(kubeRestConfig), informerOptions)
} else {
	s.kubeClient, err = kubelib.NewClient(kubelib.NewClientConfigForRestConfig(kubeRestConfig))
}
```
KubeClient.
```
func NewClientWithInformerOptions(clientConfig clientcmd.ClientConfig, options InformerOptions) (*client, error) {
	var c client
	var err error
	clientFactory := newClientFactory(clientConfig)
	c.clientFactory = clientFactory
	c.config, err = clientFactory.ToRESTConfig()
	if err != nil {
		return nil, err
	}
	c.restClient, err = clientFactory.RESTClient()
	if err != nil {
		return nil, err
	}
	c.discoveryClient, err = clientFactory.ToDiscoveryClient()
	if err != nil {
		return nil, err
	}
	c.mapper = restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(c.discoveryClient))
	c.Interface, err = kubernetes.NewForConfig(c.config)
	c.kube = c.Interface
	if err != nil {
		return nil, err
	}
	c.kubeInformer = informers.NewSharedInformerFactoryWithOptions(c.Interface, resyncInterval, informers.WithNamespace(options.WatchedNamespace))
	c.metadata, err = metadata.NewForConfig(c.config)
	if err != nil {
		return nil, err
	}
	c.metadataInformer = metadatainformer.NewFilteredSharedInformerFactory(c.metadata, resyncInterval, options.WatchedNamespace, nil)
	c.dynamic, err = dynamic.NewForConfig(c.config)
	if err != nil {
		return nil, err
	}
	c.dynamicInformer = dynamicinformer.NewFilteredDynamicSharedInformerFactory(c.dynamic, resyncInterval, options.WatchedNamespace, nil)
	c.istio, err = istioclient.NewForConfig(c.config)
	if err != nil {
		return nil, err
	}
	if options.Label != "" {
		c.istioInformer = istioinformer.NewSharedInformerFactoryWithOptions(c.istio, resyncInterval, istioinformer.WithNamespace(options.WatchedNamespace), istioinformer.WithTweakListOptions(func(listOpts *metav1.ListOptions) {
			listOpts.LabelSelector = options.Label
		}))
	} else {
		c.istioInformer = istioinformer.NewSharedInformerFactoryWithOptions(c.istio, resyncInterval, istioinformer.WithNamespace(options.WatchedNamespace))
	}
	c.gatewayapi, err = gatewayapiclient.NewForConfig(c.config)
	if err != nil {
		return nil, err
	}
	if options.Label != "" {
		c.gatewayapiInformer = gatewayapiinformer.NewSharedInformerFactoryWithOptions(c.gatewayapi, resyncInterval, gatewayapiinformer.WithNamespace(options.WatchedNamespace), gatewayapiinformer.WithTweakListOptions(func(listOpts *metav1.ListOptions) {
			listOpts.LabelSelector = options.Label
		}))
	} else {
		c.gatewayapiInformer = gatewayapiinformer.NewSharedInformerFactoryWithOptions(c.gatewayapi, resyncInterval, gatewayapiinformer.WithNamespace(options.WatchedNamespace))
	}
	c.mcsapis, err = mcsapisClient.NewForConfig(c.config)
	if err != nil {
		return nil, err
	}
	if options.Label != "" {
		c.mcsapisInformers = mcsapisInformer.NewSharedInformerFactoryWithOptions(c.mcsapis, resyncInterval, mcsapisInformer.WithNamespace(options.WatchedNamespace), mcsapisInformer.WithTweakListOptions(func(listOpts *metav1.ListOptions) {
			listOpts.LabelSelector = options.Label
		}))
	} else {
		c.mcsapisInformers = mcsapisInformer.NewSharedInformerFactoryWithOptions(c.mcsapis, resyncInterval, mcsapisInformer.WithNamespace(options.WatchedNamespace))
	}
	c.extSet, err = kubeExtClient.NewForConfig(c.config)
	if err != nil {
		return nil, err
	}
	return &c, nil
}
```

hzxuzhonghu:
@howardjohn  Can this be used for multi tenancy RFC https://docs.google.com/document/d/1y4liRJbQW0NCMeQtqMma46flVqs-izV1/edit?usp=sharing&ouid=105743966233121608166&rtpof=true&sd=true

howardjohn:
There are really 3 parts - discovery, CRDs, and configmap writing
So do we want 3 selectors, 1 selector that does all of them (with temp flag for backwards compat), or 1 selector and an option to select which of the 3 it should apply to?
I tend to think 1 selector is probably sufficient for most cases.. wdyt?

If there is no use cases for splitting these, then I think we can just make the selector apply to all 3 of these and add a flag to allow opt-out for backwards compat

cc @linsun @GregHanson @nrjpoddar @harveyxia folks involved in the original discovery selector

hzxuzhonghu:
Great， one selector makes much sense to me

kfaseela:
> Great， one selector makes much sense to me

Great! thanks @hzxuzhonghu @howardjohn : 1 selector is good enough for us as well!

pilot/pkg/config/file/store.go
```diff
@@ -78,7 +81,20 @@ func (s *KubeSource) Get(typ config.GroupVersionKind, name, namespace string) *c
 }
 
 func (s *KubeSource) List(typ config.GroupVersionKind, namespace string) ([]config.Config, error) {
-	return s.inner.List(typ, namespace)
+	configs, err := s.inner.List(typ, namespace)
+	if err != nil {
+		return nil, err
+	}
+	if namespace == model.NamespaceAll && s.namespacesFilter != nil {
```
howardjohn, line 88: Why does the filter only apply to NamespaceAll?

pilot/pkg/config/file/store.go
```diff
@@ -78,7 +81,20 @@ func (s *KubeSource) Get(typ config.GroupVersionKind, name, namespace string) *c
...
+		var out []config.Config
```
howardjohn, line 89: List if pretty big sometimes.. can/should we in-place filter the list instead of copy?

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -208,6 +211,12 @@ func (m *Multicluster) addCluster(cluster *multicluster.Cluster) (*Controller, *
 	m.remoteKubeControllers[cluster.ID] = &kubeController{
 		Controller: kubeRegistry,
 	}
+	// localCluster may also be the "config" cluster, in an external-istiod setup.
+	localCluster := m.opts.ClusterID == cluster.ID
+	if features.EnableConfigurationDiscovery && localCluster && m.configController != nil {
+		// register filter
+		m.configController.RegisterNameSpaceDiscoveryFilter(kubeRegistry.opts.DiscoveryNamespacesFilter.Filter)
```
howardjohn, line 218: I am not sure I understand this. Isn't this called per-cluster, but `m.configController` is a singleton shared by all clusters?

pilot/pkg/config/file/store.go
```diff
@@ -78,7 +81,20 @@ func (s *KubeSource) Get(typ config.GroupVersionKind, name, namespace string) *c
...
+		for _, config := range configs {
+			if s.namespacesFilter(config) {
```
howardjohn, line 91: Does this scale...? List() is processing a LOT of items - it was a bottleneck in the code prior to some performance improvements. `Filter` uses a lock on each call - we may end up needing to acquire 10k locks for 1 call to List potentially.

The current code is passing the filter to NewFilteredSharedIndexInformer, instead - this is very different since it is only processing each event, not all items on every call to List

pilot/pkg/serviceregistry/kube/controller/multicluster.go
hzxuzhonghu: It is protected by L216, so it is only registered once.

The intention to do it here is we want to reuse the `DiscoveryNamespacesFilter`

pilot/pkg/features/pilot.go
```diff
@@ -646,6 +646,10 @@ var (
 		"If enabled, the cluster secret watcher will watch the namespace of the external cluster instead of config cluster").Get()
 
 	SidecarIgnorePort = env.Register("SIDECAR_IGNORE_PORT_IN_HOST_MATCH", true, "If enabled, port will not be used in vhost domain matches.").Get()
+
+	EnableConfigurationDiscovery = env.Register("ENABLE_DISCOVERY_SELECTOR_FOR_CUSTOM_RESOURCE", false,
```
kfaseela, line 650: Since this feature flag is going to be used generically for other resource scoping as well as mentioned in the RFC, would it make sense to rename it in a bit more generic way?

pilot/pkg/serviceregistry/kube/controller/multicluster.go
howardjohn: Ah I think I missed the `localCluster` part

pilot/pkg/features/pilot.go
howardjohn: +1, but we should also make all the changes we are going to make in 1 release - so there is only 1 user facing behavior change.

hzxuzhonghu: updated

pilot/pkg/features/pilot.go
```diff
@@ -646,6 +646,10 @@ var (
 		"If enabled, the cluster secret watcher will watch the namespace of the external cluster instead of config cluster").Get()
 
 	SidecarIgnorePort = env.Register("SIDECAR_IGNORE_PORT_IN_HOST_MATCH", true, "If enabled, port will not be used in vhost domain matches.").Get()
+
+	EnableEnhancedResourceScoping = env.Register("ENABLE_ENHANCED_RESOURCE_SCOPING", false,
+		"If enabled, meshConfig.discoverySelectors will also limit the configurations(like Gateway,VirtualService,DestinationRule,Ingress, etc)"+
```
kfaseela, line 651: description also should be enhanced

hzxuzhonghu: ANY OTHERS?

pilot/pkg/config/kube/ingress/controller.go
```diff
@@ -178,7 +180,13 @@ func NewController(client kube.Client, meshWatcher mesh.Holder,
 	c.queue = controllers.NewQueue("ingress",
 		controllers.WithReconciler(c.onEvent),
 		controllers.WithMaxAttempts(5))
-	c.ingressInformer.AddEventHandler(controllers.ObjectHandler(c.queue.AddObject))
+	c.ingressInformer.AddEventHandler(
+		controllers.FilteredObjectHandler(c.queue.AddObject, func(o controllers.Object) bool {
```
howardjohn, line 184: shouldn't we be filtering the informer itself, not the handler? We may be reading Ingresses outside of the handler events

howardjohn: I see. I was looking at NewFilteredSharedIndexInformer and thought it was in core client-go doing it under the hood. Looks like not.

Can we re-use NewFilteredSharedIndexInformer though? It will do the same thing but avoid re-implementing the same (somewhat complex) logic everywhere?

Probably can move it to `pkg/kube` or similar

pilot/pkg/config/kube/ingress/controller.go
```diff
@@ -361,6 +373,9 @@ func (c *controller) List(typ config.GroupVersionKind, namespace string) ([]conf
 	ingressByHost := map[string]*config.Config{}
 
 	for _, ingress := range sortIngressByCreationTime(c.ingressInformer.GetStore().List()) {
+		if c.namespacesFilter != nil && !c.namespacesFilter(ingress) {
```
howardjohn, line 376: I think if we use NewFilteredSharedIndexInformer List will handle this automatically?

pilot/pkg/config/kube/ingress/controller.go
hzxuzhonghu: That looks good, i will do that

pilot/pkg/config/kube/ingress/controller.go
hzxuzhonghu: good catch, i will move it to a common place

pilot/pkg/config/kube/crdclient/cache_handler.go
```diff
@@ -82,16 +82,11 @@ func createCacheHandler(cl *Client, schema collection.Schema, i informers.Generi
...
 		schema:   schema,
-		informer: i.Informer(),
-	}
-	h.lister = func(namespace string) cache.GenericNamespaceLister {
-		if schema.Resource().IsClusterScoped() {
-			return i.Lister()
-		}
-		return i.Lister().ByNamespace(namespace)
+		informer: informer.NewFilteredSharedIndexInformer(cl.namespacesFilter, i.Informer()),
 	}
```
howardjohn, line 86: `lister: cache.NewGenericLister(h.informer.GetIndexer(), schema.Resource().GroupVersionResource().GroupResource())`

can we do this?

hzxuzhonghu: Definitely we can implement it in theory, but I think currently it is enough to use

NewGenericLister does not provide much benefit to our case

hzxuzhonghu: As you can see with generic lister for clusterscoped resource, we need to handle differently. however with underlying indexer, we do not need to care about it, that is why I believe this way is more simple

hzxuzhonghu: Actually, all the indexer related code are in pkg/kube, the fiteredInformer does only provide Get List interface.

howardjohn: I am not sure how it's simpler. The lister requires 1 single if statement here - the indexer approach has like 50 lines of added complexity throughout the PR of more complex error handling, type casting, and still has the if for namespaces but at each List call instead of in one single place 

Listers are the Kubernetes standard for his to list objects, I don't see a compelling reason to move away from that. It also seems unrelated to the rest of the PR.

hzxuzhonghu: GenericLister returns runtime.Object, which also need type cast, error handling. I do not have strong preference though

pkg/kube/namespace/filter.go
```diff
@@ -74,6 +74,10 @@ func (d *discoveryNamespacesFilter) Filter(obj any) bool {
 		return true
 	}
 
+	if ns, ok := obj.(string); ok {
```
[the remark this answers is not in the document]
hzxuzhonghu: This is to filter out namespaces, which is useful for namespace controller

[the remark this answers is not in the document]
howardjohn: Its not used in this PR, right? Just the future?

In the namespace controllre PR we discussed making a `FilterObject(obj Object)` and `FilterNamespace(namespace string)`. Instead of doing all the type casting, etc. This is more explicit API

pilot/pkg/config/kube/crdclient/cache_handler.go
howardjohn: yes but we don't need this: https://github.com/istio/istio/pull/36639/files#diff-eff3d408ac506e9395b51f480c999a588f9fa408e392ce77506f25ceae00d6a1R328


and we now have to do:
```
	item, _, err := h.informer.GetIndexer().GetByKey(KeyFunc(namespace, name))
	if item == nil || err != nil {
```

instead of just 

```
	obj, err := h.lister(namespace).Get(name)
	if err != nil {
```

pkg/kube/namespace/filter.go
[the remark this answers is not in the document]
hzxuzhonghu: removed

pilot/pkg/config/kube/ingress/controller.go
```diff
@@ -168,17 +169,22 @@ func NewController(client kube.Client, meshWatcher mesh.Holder,
...
 		ingresses:       make(map[types.NamespacedName]*ingress.Ingress),
-		ingressInformer: ingressInformer.Informer(),
 		ingressLister:   ingressInformer.Lister(),
 		classes:         classes,
 		serviceInformer: serviceInformer.Informer(),
 		serviceLister:   serviceInformer.Lister(),
 	}
 
+	if options.DiscoveryNamespacesFilter != nil {
+		c.filteredIngressInformer = informer.NewFilteredSharedIndexInformer(options.DiscoveryNamespacesFilter.Filter, ingressInformer.Informer())
```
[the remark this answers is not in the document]
howardjohn: still applicable but not blocking

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -802,7 +803,7 @@ func (c *Controller) syncNodes() error {
 
 func (c *Controller) syncServices() error {
 	var err *multierror.Error
-	services := c.serviceInformer.GetIndexer().List()
+	services, _ := c.serviceInformer.List("")
```
howardjohn, line 806: nit: use metav1.NamespaceAll

pilot/pkg/serviceregistry/kube/controller/serviceexportcache.go
```diff
@@ -58,8 +58,11 @@ func newServiceExportCache(c *Controller) serviceExportCache {
 		_ = dInformer.Informer().SetTransform(kube.StripUnusedFields)
 		ec := &serviceExportCacheImpl{
 			Controller: c,
-			informer:   dInformer.Informer(),
-			lister:     dInformer.Lister(),
+		}
+		if c.opts.DiscoveryNamespacesFilter != nil {
+			ec.filteredInformer = informer.NewFilteredSharedIndexInformer(c.opts.DiscoveryNamespacesFilter.Filter, dInformer.Informer())
+		} else {
+			ec.filteredInformer = informer.NewFilteredSharedIndexInformer(nil, dInformer.Informer())
```
howardjohn, line 65: same comment about reducing `if`

pkg/kube/multicluster/secretcontroller.go
```diff
@@ -147,6 +150,10 @@ func NewController(kubeclientset kube.Client, namespace string, clusterID cluste
 		informer:            secretsInformer,
 	}
 
+	nsInformer := kubeclientset.KubeInformer().Core().V1().Namespaces().Informer()
+	_ = nsInformer.SetTransform(kube.StripUnusedFields)
+	nsLister := kubeclientset.KubeInformer().Core().V1().Namespaces().Lister()
+	controller.DiscoveryNamespacesFilter = filter.NewDiscoveryNamespacesFilter(nsLister, meshWatcher.Mesh().GetDiscoverySelectors())
```
howardjohn, line 156: do we want to protect this with the feature flag?

howardjohn (approved):
lgtm, we can followup on the rest

pilot/pkg/serviceregistry/kube/controller/serviceexportcache.go
hzxuzhonghu: c.opts.DiscoveryNamespacesFilter can be nil, this can be optimized later, because we need to filter namespace string as well

PR #36651, opened 2021-12-28
Make includeInboundPorts values override work.

### LINKED ISSUE #36644: Can't override includeInboundPorts from values.yaml file
### Bug Description

I can confirm that istiod helm chart (1.12.1 version) have by default includeInboundPorts: "*" in values.yaml.
But overriding the parameters doesn't take place since files/injection-template.yaml has this args:

"{{ annotation .ObjectMeta traffic.sidecar.istio.io/includeInboundPorts * }}"
That's why the sidecar is added with default and doesn't takes the override value
INBOUND_PORTS_INCLUDE=*

### Version

```prose
$ kubectl version --short
Client Version: v1.19.3
Server Version: v1.19.16-gke.1500
$ helm list --namespace istio-system
NAME                            	NAMESPACE   	REVISION	UPDATED                             	STATUS  	CHART        	APP VERSION
istio-base	istio-system	5       	2021-12-27 18:13:24.675766 +0200 IST	deployed	base-1.12.1  	1.12.1
istiod    	istio-system	4       	2021-12-27 18:15:38.536671 +0200 IST	deployed	istiod-1.12.1	1.12.1
```

### Additional Information

_No response_

bianpengyuan:
fix https://github.com/istio/istio/issues/36644


ericvn:
I would think that this PR would need a release note as we are fixing something in the release and likely needing to fix this in 1.12.

bianpengyuan:
> I would think that this PR would need a release note as we are fixing something in the release and likely needing to fix this in 1.12.

Hmm a release notes is actually already included. Not sure why I added `release-notes-none` (I might want to add cherry-pick release 1.12 but somehow added the `release-notes-none` label). Sorry about that.

ericvn (approved):

jacob-delgado:
Cherry pick to 1.11?

bianpengyuan:
> Cherry pick to 1.11?

The option is not available at 1.11.

bianpengyuan:
@jacob-delgado any chance to take another look? thanks!

morvencao (approved):

Issue #36658, opened 2021-12-29
EDS data race


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/unit-tests_istio_postsubmit/1476183392830623744

howardjohn:
cc @hzxuzhonghu I didn't take a look but I recall there were some changes in this area

PR #36659, opened 2021-12-29
Revert "Use read lock when building endpoints from shards"

howardjohn:
Reverts istio/istio#36657
fixes https://github.com/istio/istio/issues/36658

### DIFF
--- a/pilot/pkg/xds/endpoint_builder.go
+++ b/pilot/pkg/xds/endpoint_builder.go
@@ -274,7 +274,7 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
 	// and should, therefore, not be accessed from outside the cluster.
 	isClusterLocal := b.clusterLocal
 
-	shards.mutex.RLock()
+	shards.mutex.Lock()
 	// Extract shard keys so we can iterate in order. This ensures a stable EDS output. Since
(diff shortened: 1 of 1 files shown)


dddddai:
The root cause is:
https://github.com/istio/istio/blob/c151f81ab7e0934269ad660b49bb2f1317ef55fa/pilot/pkg/xds/endpoint_builder.go#L323

The endpoint is modified and then read by another goroutine
But in terms of the result, it should not matter as `ep.EnvoyEndpoint` is always the same for each endpoint

hzxuzhonghu (approved):

ramaraochavali:
@howardjohn Sorry did not understand what is the problem with just read lock there? It does not seem to be modifying shards?

dddddai:
> Sorry did not understand what is the problem with just read lock there? It does not seem to be modifying shards?

The root cause is:
https://github.com/istio/istio/blob/995ded3d2b93a2dbf70a1bc51a71057f5f03d318/pilot/pkg/xds/endpoint_builder.go#L323

The endpoint is modified and then read by another goroutine
But in terms of the result, it does not matter since `ep.EnvoyEndpoint` is always the same for each endpoint, maybe we should revert it to get rid of the race before we come up with a better solution though

howardjohn:
It is not safe to write concurrently to the same location in memory even if
you are writing the same data, assignments are not always atomic. Even if
it was it is breaking tests at the very least

PR #30831, opened 2021-02-12
Support for smart dns using istio-cni

nikhilag09:
[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any changes that may affect Istio users.

Solves: https://github.com/istio/istio/issues/29511

PR #32416, opened 2021-04-23
Add DNS capture into CNI.

bianpengyuan:
fix #29511

PR #36667, opened 2021-12-30
Added flag to enable capture of invalid drop into CNI

shonecyx:

This is related with https://github.com/istio/istio/pull/36566 which introduced a flag `ISTIO_META_INVALID_DROP` to enable the invalid drop iptables rule while there was a bug that this flag won't be passed from Pod environment variable to istio CNI.

**Steps to reproduce the bug:**
1. Install istio with cni and deploy the istio-cni DaemonSet with the cherry-picked fix https://github.com/istio/istio/pull/36566
2. Config the ISTIO_META_INVALID_DROP into proxyMetaData :
```
  mesh: |-
    accessLogEncoding: JSON
    accessLogFile: /var/log/proxy/access.log
    defaultConfig:
      concurrency: 40
      discoveryAddress: istiod.istio-pre-production.svc.87.tess.io:15012
      proxyMetadata:
        ISTIO_META_DNS_AUTO_ALLOCATE: "true"
        ISTIO_META_DNS_CAPTURE: "true"
        ISTIO_META_INVALID_DROP: "true"
```
2. Create a new pod with sidecar injection and we can see the env is in pod
```
    env:
    - name: ISTIO_META_DNS_AUTO_ALLOCATE
      value: "true"
    - name: ISTIO_META_DNS_CAPTURE
      value: "true"
    - name: ISTIO_META_INVALID_DROP
      value: "true"
```
3. nsenter the pod network namespace and check the iptables rule, the invalid drop policy is missing:
```
# Generated by xtables-save v1.8.2 on Thu Dec 30 01:31:55 2021
*mangle
:PREROUTING ACCEPT [0:0]
:INPUT ACCEPT [0:0]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
:POSTROUTING ACCEPT [0:0]
COMMIT
# Completed on Thu Dec 30 01:31:55 2021
# Generated by xtables-save v1.8.2 on Thu Dec 30 01:31:55 2021
*nat
:PREROUTING ACCEPT [0:0]
:INPUT ACCEPT [0:0]
:POSTROUTING ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
:ISTIO_INBOUND - [0:0]
:ISTIO_REDIRECT - [0:0]
:ISTIO_IN_REDIRECT - [0:0]
:ISTIO_OUTPUT - [0:0]
-A PREROUTING -p tcp -j ISTIO_INBOUND
-A OUTPUT -p tcp -j ISTIO_OUTPUT
-A OUTPUT -p udp -m udp --dport 53 -m owner --uid-owner 1337 -j RETURN
-A OUTPUT -p udp -m udp --dport 53 -m owner --gid-owner 1337 -j RETURN
-A OUTPUT -p udp -m udp --dport 53 -j REDIRECT --to-ports 15053
-A ISTIO_INBOUND -p tcp -m tcp --dport 15008 -j RETURN
-A ISTIO_INBOUND -p tcp -m tcp --dport 22 -j RETURN
-A ISTIO_INBOUND -p tcp -m tcp --dport 15020 -j RETURN
-A ISTIO_INBOUND -p tcp -m tcp --dport 15021 -j RETURN
-A ISTIO_INBOUND -p tcp -m tcp --dport 15090 -j RETURN
-A ISTIO_INBOUND -p tcp -j ISTIO_IN_REDIRECT
-A ISTIO_REDIRECT -p tcp -j REDIRECT --to-ports 15001
-A ISTIO_IN_REDIRECT -p tcp -j REDIRECT --to-ports 15006
-A ISTIO_OUTPUT -p tcp -m tcp --dport 15020 -j RETURN
-A ISTIO_OUTPUT -s 127.0.0.6/32 -o lo -j RETURN
-A ISTIO_OUTPUT ! -d 127.0.0.1/32 -o lo -p tcp -m tcp ! --dport 53 -m owner --uid-owner 1337 -j ISTIO_IN_REDIRECT
-A ISTIO_OUTPUT -o lo -p tcp -m tcp ! --dport 53 -m owner ! --uid-owner 1337 -j RETURN
-A ISTIO_OUTPUT -m owner --uid-owner 1337 -j RETURN
-A ISTIO_OUTPUT ! -d 127.0.0.1/32 -o lo -m owner --gid-owner 1337 -j ISTIO_IN_REDIRECT
-A ISTIO_OUTPUT -o lo -p tcp -m tcp ! --dport 53 -m owner ! --gid-owner 1337 -j RETURN
-A ISTIO_OUTPUT -m owner --gid-owner 1337 -j RETURN
-A ISTIO_OUTPUT -p tcp -m tcp --dport 53 -j REDIRECT --to-ports 15053
-A ISTIO_OUTPUT -d 127.0.0.1/32 -j RETURN
-A ISTIO_OUTPUT -j ISTIO_REDIRECT
COMMIT
# Completed on Thu Dec 30 01:31:55 2021
```

- [ ] Docs
- [ ] Installation
- [ x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/cni/pkg/plugin/iptables.go
+++ b/cni/pkg/plugin/iptables.go
@@ -54,6 +54,7 @@ func (ipt *iptables) Program(podName, netns string, rdrct *Redirect) error {
 	viper.Set(constants.OutputPath, drf)
 	viper.Set(constants.RedirectDNS, rdrct.dnsRedirect)
 	viper.Set(constants.CaptureAllDNS, rdrct.dnsRedirect)
+	viper.Set(constants.DropInvalid, rdrct.invalidDrop)
 	iptablesCmd := cmd.GetCommand()
 	log.Infof("============= Start iptables configuration for %v =============", podName)
 	defer log.Infof("============= End iptables configuration for %v =============", podName)
--- a/cni/pkg/plugin/plugin_dryrun_test.go
+++ b/cni/pkg/plugin/plugin_dryrun_test.go
@@ -32,6 +32,7 @@ import (
 	"istio.io/istio/pilot/cmd/pilot-agent/options"
 	diff "istio.io/istio/pilot/test/util"
 	"istio.io/istio/pkg/test/env"
+	"istio.io/istio/tools/istio-iptables/pkg/cmd"
 )
 
 type k8sPodInfoFunc func(*kubernetes.Clientset, string, string) (*PodInfo, error)
@@ -113,6 +114,16 @@ func TestIPTablesRuleGeneration(t *testing.T) {
 				ProxyEnvironments: map[string]string{options.DNSCaptureByAgent.Name: "true"},
 			},
 			golden: filepath.Join(env.IstioSrc, "cni/pkg/plugin/testdata/dns.txt.golden"),
+		},
+		{
+			name: "invalid-drop",
+			input: &PodInfo{
+				Containers:        []string{"test", "istio-proxy"},
+				InitContainers:    map[string]struct{}{"istio-validate": {}},
+				Annotations:       map[string]string{annotation.SidecarStatus.Name: "true"},
+				ProxyEnvironments: map[string]string{cmd.InvalidDropByIptables.Name: "true"},
+			},
+			golden: filepath.Join(env.IstioSrc, "cni/pkg/plugin/testdata/invalid-drop.txt.golden"),
 		},
 	}
 
--- a/cni/pkg/plugin/redirect.go
+++ b/cni/pkg/plugin/redirect.go
@@ -23,6 +23,7 @@ import (
 
 	"istio.io/api/annotation"
 	"istio.io/istio/pilot/cmd/pilot-agent/options"
+	"istio.io/istio/tools/istio-iptables/pkg/cmd"
 	"istio.io/pkg/log"
 )
 
@@ -78,6 +79,7 @@ type Redirect struct {
 	kubevirtInterfaces   string
 	excludeInterfaces    string
 	dnsRedirect          bool
+	invalidDrop          bool
 }
 
 type annotationValidationFunc func(value string) error
@@ -255,6 +257,12 @@ func NewRedirect(pi *PodInfo) (*Redirect, error) {
 			log.Warnf("cannot parse DNS capture environment variable %v", valErr)
 		}
 	}
-
+	if v, found := pi.ProxyEnvironments[cmd.InvalidDropByIptables.Name]; found {
+		// parse and set the bool value of invalidDrop
+		redir.invalidDrop, valErr = strconv.ParseBool(v)
+		if valErr != nil {
+			log.Warnf("cannot parse invalid drop environment variable %v", valErr)
+		}
+	}
 	return redir, nil
 }
--- a/cni/pkg/plugin/testdata/invalid-drop.txt.golden
+++ b/cni/pkg/plugin/testdata/invalid-drop.txt.golden
@@ -1,0 +1,28 @@
+* mangle
+-A PREROUTING -m conntrack --ctstate INVALID -j DROP
+COMMIT
+* nat
+-N ISTIO_INBOUND
+-N ISTIO_REDIRECT
+-N ISTIO_IN_REDIRECT
+-N ISTIO_OUTPUT
+-A ISTIO_INBOUND -p tcp --dport 15008 -j RETURN
+-A ISTIO_REDIRECT -p tcp -j REDIRECT --to-ports 15001
+-A ISTIO_IN_REDIRECT -p tcp -j REDIRECT --to-ports 15006
+-A PREROUTING -p tcp -j ISTIO_INBOUND
+-A ISTIO_INBOUND -p tcp --dport 15020 -j RETURN
+-A ISTIO_INBOUND -p tcp --dport 15021 -j RETURN
+-A ISTIO_INBOUND -p tcp --dport 15090 -j RETURN
+-A ISTIO_INBOUND -p tcp -j ISTIO_IN_REDIRECT
+-A OUTPUT -p tcp -j ISTIO_OUTPUT
+-A ISTIO_OUTPUT -p tcp --dport 15020 -j RETURN
+-A ISTIO_OUTPUT -o lo -s 127.0.0.6/32 -j RETURN
+-A ISTIO_OUTPUT -o lo ! -d 127.0.0.1/32 -m owner --uid-owner 1337 -j ISTIO_IN_REDIRECT
+-A ISTIO_OUTPUT -o lo -m owner ! --uid-owner 1337 -j RETURN
+-A ISTIO_OUTPUT -m owner --uid-owner 1337 -j RETURN
+-A ISTIO_OUTPUT -o lo ! -d 127.0.0.1/32 -m owner --gid-owner 1337 -j ISTIO_IN_REDIRECT
+-A ISTIO_OUTPUT -o lo -m owner ! --gid-owner 1337 -j RETURN
+-A ISTIO_OUTPUT -m owner --gid-owner 1337 -j RETURN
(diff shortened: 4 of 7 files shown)


shonecyx:
@howardjohn PTAL

releasenotes/notes/36667.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
```
bianpengyuan, line 1: I think there is already a release notes in the original fix? If so no need to repeat it for CNI since the config knob is the same.

pilot/cmd/pilot-agent/options/options.go
```diff
@@ -89,7 +89,9 @@ var (
 	// DNSCaptureByAgent is a copy of the env var in the init code.
 	DNSCaptureByAgent = env.RegisterBoolVar("ISTIO_META_DNS_CAPTURE", false,
 		"If set to true, enable the capture of outgoing DNS packets on port 53, redirecting to istio-agent on :15053")
-
+	// InvalidDropByIptables is a copy of the env var in the init code.
+	InvalidDropByIptables = env.RegisterBoolVar("ISTIO_META_INVALID_DROP", false,
```
bianpengyuan, line 93: I don't think we need to actually define it? We only need it to present on pod spec to let CNI consume it.

releasenotes/notes/36667.yaml
shonecyx: There is a release-notes_istio prow job for each PR, so the release notes is mandatory.

pilot/cmd/pilot-agent/options/options.go
shonecyx: @bianpengyuan CNI consumes it from here 
https://github.com/istio/istio/blob/master/tools/istio-iptables/pkg/capture/run.go#L360 
But the config was constructed here https://github.com/istio/istio/blob/master/tools/istio-iptables/pkg/cmd/root.go#L101 
That's why we need to set it from here https://github.com/istio/istio/pull/36667/files#diff-b59f1b2d634969faeda5096d06c7626e4de94d73c437391d08e0e91bbe9a1e21R57

releasenotes/notes/36667.yaml
bianpengyuan: You can add `release-notes-none` label.

pilot/cmd/pilot-agent/options/options.go
bianpengyuan: Not sure I follow. I think you just need the env variable name at CNI when it checks env from pod? Can you just use the existing one defined at https://github.com/istio/istio/blob/master/tools/istio-iptables/pkg/cmd/root.go#L43? The agent does not really have any logic based on this env variable, so feels not right to define it here.

releasenotes/notes/36667.yaml
bianpengyuan: I added the label manually.

shonecyx: Removed, thanks

pilot/cmd/pilot-agent/options/options.go
shonecyx: @bianpengyuan  Actually this PR is quite similar with this one https://github.com/istio/istio/pull/32416/files.
I went through the code again and this is the current flow:

1. Istio CNI redirect. go will read the `ISTIO_META_INVALID_DROP` from pod env variable and assign to `rdrct.invalidDrop`
2. Istio CNI iptables.go read this flag from `#1` `rdrct.invalidDrop` and set to `viper.Set(constants.DropInvalid, rdrct.invalidDrop)`
3. Istio CNI iptables.go will program the iptables rules based on the parameters provided in Redirect, it will get the iptables command from Istio tools root.go. While the root.go will `constructConfig()` based on the `viper.GetBool(constants.DropInvalid)` and the value was set in `#2`.
4. Istio CNI generates the iptables with DropInvalid = true, so the invalid drop policy is there.

If my understanding is correct, your concern is in `#3`, the Istio CNI can directly read the flag from [here](https://github.com/istio/istio/blob/master/tools/istio-iptables/pkg/cmd/root.go#L43) , that would possibly work but it will not follow the `viper` style.

bianpengyuan: No, the flow is right. All I mean is that we do not need to define it at agent. At `#1`, after adding the env var via PROXY_METADATA, it will show up at istio proxy env var list. CNI can parse that env var even without actually defining it at agent. Not sure if I miss something here?

shonecyx: @bianpengyuan  Just tested again without this PR Istio CNI will always get that env as false even specified it as true in the proxy container spec and in the printenv, so the invalid drop rule is not there:
```
kubectl exec  -n httpbin-pre-production swissknife-test-1 -c istio-proxy  -- printenv |  grep ISTIO_META_INVALID_DROP
ISTIO_META_INVALID_DROP=true
```
So seems like the CNI can't parse env var here and more precisely, the istio-iptables binary can't read it from the env since istio-cni calls istio-iptables with params to setup pod netns.

So maybe I can refactor this PR to remove the ISTIO_META_INVALID_DROP in CNI as below but keep it in agent:
https://github.com/istio/istio/blob/master/tools/istio-iptables/pkg/cmd/root.go#L43

bianpengyuan: Yeah I understand we need this PR. I meant you don't need to define it again in the agent code (i.e. revert change at this options.go file and use the init container env var that you already have to get the name of invalid_drop env var.) CNI does not read the env var at proxy container runtime. It reads from the pod spec directly (i.e. cni is doing something like `kubectl get pod -oyaml | grep ISTIO_META_INVALID_DROP` before pod starts up). If you keep this all the changes you have in this PR and revert this options.go file. Things should still work.

shonecyx: And I think the reason why `istio-iptables` can't read the `ISTIO_META_INVALID_DROP` from proxy container env var is that in CNI mode this binary runs in host network and then do nsenter and inject iptables to pod network ns, so it can't read the container env var. 
https://github.com/istio/istio/blob/master/tools/istio-iptables/pkg/cmd/root.go#L331
In other words, CNI can only set the default value and need agent to get it from the container env var in spec and pass to CNI.

bianpengyuan: Agent does not pass env var to CNI. All of CNI logic executes before pod starts up.

shonecyx: >CNI can only set the default value and need agent to get it from the container env var in spec and pass to CNI.

It's actually istio-iptables can only set the default value and need CNI to get it from the container env var and pass to the istio-iptables with the addition of an CLI argument.

It reminds me of another upstream issue that the smart dns proxy was not working in CNI mode that's why there was a fix https://github.com/istio/istio/pull/30831.  It's similar issue there since with CNI the iptables binary runs on the node, it's getting the env var from the node, not from the container. With the addition of an istio-iptables CLI argument, CNI can read it from the pod info and pass to istio-iptables.

I think I'm clear now. There are two flows need to take care, one is the sidecar injection through init-container and another is the CNI and this PR is particular for the CNI flow.

Thanks @bianpengyuan  for you comments, I just refactor the PR.

shonecyx: @bianpengyuan  I've tested this PR and it's also working, please let me know if you have any comments, thanks!

shonecyx:
Removed prefix ISTIO_META_ to avoid sending it to pilot as per comments:
https://github.com/istio/istio/pull/36566#discussion_r776124395

shonecyx:
@bianpengyuan @hzxuzhonghu  @howardjohn  Kindly help take a look to move forward.

bianpengyuan (approved):

PR #36674, opened 2021-12-31
Fix service entry instances store memory leak

dddddai:
If a workload was deleted, it would still remain in `serviceInstances.instancesBySE`
It does leak memory when users enable `ServiceEntrySelectPods`(which is enabled by default) and rely on it

You can reproduce it with [the test](https://github.com/dddddai/istio/blob/aca3a467528e02a1099235652d46cc4a2307d5c7/pilot/pkg/serviceregistry/serviceentry/servicediscovery_test.go#L984) in this pr

### DIFF
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
@@ -463,6 +463,8 @@ func (s *ServiceEntryStore) WorkloadInstanceHandler(wi *model.WorkloadInstance, 
 				instancesDeleted = append(instancesDeleted, di)
 			}
 			s.serviceInstances.deleteServiceEntryInstances(seNamespacedName, key)
+		} else if event == model.EventDelete {
+			s.serviceInstances.deleteServiceEntryInstances(seNamespacedName, key)
 		} else {
 			s.serviceInstances.updateServiceEntryInstancesPerConfig(seNamespacedName, key, instance)
 		}
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery_test.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery_test.go
@@ -22,6 +22,8 @@ import (
 	"testing"
 	"time"
 
+	"k8s.io/apimachinery/pkg/types"
+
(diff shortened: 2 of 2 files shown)


releasenotes/notes/service-entry-instances-store-memory-leak.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+releaseNotes:
+- |
+  **Fixed** service entry instances store memory leak
```
hzxuzhonghu, line 6: We donot need this as this bug is not in previous releases, so no need to expose out.

hzxuzhonghu (approved):

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -463,6 +463,8 @@ func (s *ServiceEntryStore) WorkloadInstanceHandler(wi *model.WorkloadInstance,
 				instancesDeleted = append(instancesDeleted, di)
 			}
 			s.serviceInstances.deleteServiceEntryInstances(seNamespacedName, key)
+		} else if event == model.EventDelete {
+			s.serviceInstances.deleteServiceEntryInstances(seNamespacedName, key)
```
hzxuzhonghu, line 467: good catch, how do you find it

dddddai: Thanks @hzxuzhonghu! I found this when working on #36078

BTW, could you please help add `release-note-none` label, the [comment](https://github.com/istio/istio/pull/36674#issuecomment-1003285961) seems not work

jacob-delgado:
Thanks again @dddddai. Do you know if this should by cherry picked to 1.11 and 1.12?

dddddai:
> Thanks again @dddddai. Do you know if this should by cherry picked to 1.11 and 1.12?

Hello @jacob-delgado, I've checked the code in 1.11 and 1.12, neither needs cherry picking, because `instancesBySE` is introduced in 1.13, as @hzxuzhonghu [metioned above](https://github.com/istio/istio/pull/36674#discussion_r776936698)

PR #36677, opened 2021-12-31
add logging integration test

zirain:

add logging integration test


bianpengyuan (commented):
For the sake of reducing the test time (Telemetry test is already the slowest test suite), can we combine this to the existing telemetry test to avoid adding a new suite? The prometheus test seems to be a better fit.

zirain:
> For the sake of reducing the test time (Telemetry test is already the slowest test suite), can we combine this to the existing telemetry test to avoid adding a new suite? The prometheus test seems to be a better fit.

IMHO, logging is the first class api, it's not a good idea to combine with other test suite.
this PR is the first way, I'll add more test cases about logging provider and telemetry in the near future。

bianpengyuan:
> IMHO, logging is the first class api, it's not a good idea to combine with other test suite. this PR is the first way, I'll add more test cases about logging provider and telemetry in the near future。

First class API does not mean that we need to create a new suite. As an analogy, all networking APIs are tested with one suite with `TestTraffic`. Creating a new suite will increase test time significantly. The existing prometheus test does not conflict with log at all, so would be a better place to augment. If the name of test package is not inclusive enough, it can be renamed to `telemetry` etc.

zirain:
> > IMHO, logging is the first class api, it's not a good idea to combine with other test suite. this PR is the first way, I'll add more test cases about logging provider and telemetry in the near future。
> 
> First class API does not mean that we need to create a new suite. As an analogy, all networking APIs are tested with one suite with `TestTraffic`. Creating a new suite will increase test time significantly. The existing prometheus test does not conflict with log at all, so would be a better place to augment. If the name of test package is not inclusive enough, it can be renamed to `telemetry` etc.

ok, did you means `tests/integration/telemetry/stats/prometheus` ?

bianpengyuan:
> ok, did you means tests/integration/telemetry/stats/prometheus ?

Yep, the package name and test directory might need some reorganization to better accommodate new tests that will come with telemetry API.

zirain:
> > ok, did you means tests/integration/telemetry/stats/prometheus ?
> 
> Yep, the package name and test directory might need some reorganization to better accommodate new tests that will come with telemetry API.

ok, I'll take a overall look after holiday.

zirain:
@bianpengyuan PTAL, it's better to do re-org with another PR.

tests/integration/telemetry/stats/prometheus/extsionproviders/accesslogs_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"testing"
+
+	"istio.io/istio/pkg/test/framework"
+	"istio.io/istio/pkg/test/framework/components/istio"
+	"istio.io/istio/pkg/test/framework/label"
+	"istio.io/istio/pkg/test/framework/resource"
+	common "istio.io/istio/tests/integration/telemetry/stats/prometheus"
+)
+
+func TestMain(m *testing.M) {
```
bianpengyuan, line 30: Can we just add the access log mesh config to the existing test suite and avoid adding new suite https://github.com/istio/istio/blob/a3a07b85065a1ebd6eef2ba7e633eb311a628b60/tests/integration/telemetry/stats/prometheus/nullvm/stats_filter_test.go#L48 or is there some reason we cannot do that?

bianpengyuan: Probably use https://github.com/istio/istio/blob/4a28c571b84adf2eab44d009ef338a2686501510/pkg/test/framework/components/istio/util.go#L180.

zirain: I'll try to refactor the test case ASAP

zirain:
@bianpengyuan request for a review again.

tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
```diff
@@ -33,33 +19,51 @@ package nullvm
...
+extensionProviders:
+- name: file-log
+  envoyFileAccessLog:
+    path: /dev/stdout
+defaultProviders:
+  accessLogging:
+  - file-log
+`
+			ist := *(common.GetIstioInstance())
+			istio.PatchMeshConfig(t, ist.Settings().IstioNamespace, t.Clusters(), cfg)
```
bianpengyuan, line 60: Does mesh config get written back to the pre-patched version after test finishes? Otherwise we are relying on the order of test which seems to be not ideal.

tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
```diff
@@ -33,33 +19,51 @@ package nullvm
...
+			time.Sleep(time.Minute) // wait new meshconfig come into force
```
bianpengyuan, line 61: Let's not idle wait on config, which slows down the test quite a bit. RunAccessLogTests below should check log entries based on test id, so busy wait there should be good enough.

tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
zirain: yes, you're right, it may cause fail when another test case added, I'll fixed it.

Issue #36686, opened 2022-01-03
Istio should exit immediately during shutdown if envoy exits


rittneje:
**Describe the feature request**

If envoy exits after istio-proxy receives SIGTERM (perhaps because someone sends /quitquitquit to it), then istio-proxy should exit immediately.

Currently it will get stuck in `Agent.terminate`.
1. If `exitOnZeroActiveConnections` is true, then it will keep polling the number of active connections. Since this will fail, it will return -1 all the time, which will cause it to loop infinitely until the pod termination grace period expires.
2. If `exitOnZeroActiveConnections` is false, then it will pointlessly keep waiting until the `terminationDrainDuration` expires.

To fix this, it must pull from `a.statusCh` and take appropriate action.

If this is fixed, then use cases like ingress-nginx that cannot rely on `exitOnZeroActiveConnections` can at least issue /quitquitquit just before exiting to abort the `terminationDrainDuration` sleep prematurely.

**Describe alternatives you've considered**

None

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

ramaraochavali:
Are you asking if `/quitquitquit` should forcefully exit? If yes - I think that is fair given that it is currently only used in Jobs. But if it is used else where, we can also add `/quitquitquit?shutdownImmediate` .

rittneje:
No, I'm saying right now there is arguably a bug if envoy exits for any reason while istio-proxy is in the middle of shutdown (in `Agent.terminate`).

Also, I think you may have misunderstood. I am referring to the [envoy endpoint](https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--quitquitquit). From what I understand, istio-proxy itself also has a `/quitquitquit` endpoint that causes it to send SIGTERM to itself, but that is separate.

hzxuzhonghu:
I get to understand the issue, it is a potential race here, let me file a fix

rittneje:
@ramaraochavali It seems the istio-policy-bot needs you to do something else.

ramaraochavali:
no stale

rittneje:
[...]

PR #36691, opened 2022-01-04
Fix potential stuck when envoy exits before SIGTERM received

hzxuzhonghu:

Fix https://github.com/istio/istio/issues/36686

### DIFF
--- a/pkg/envoy/agent.go
+++ b/pkg/envoy/agent.go
@@ -124,13 +124,6 @@ func (a *Agent) Run(ctx context.Context) {
 		log.Infof("No more active epochs, terminating")
 	case <-ctx.Done():
 		a.terminate()
-		status := <-a.statusCh
-		if status.err == errAbort {
-			log.Infof("Epoch %d aborted normally", status.epoch)
-		} else {
-			log.Warnf("Epoch %d aborted abnormally", status.epoch)
-		}
-		log.Info("Agent has successfully terminated")
 	}
 }
 
@@ -153,11 +146,12 @@ func (a *Agent) terminate() {
 			if ac == 0 {
 				log.Info("There are no more active connections. terminating proxy...")
 				a.abortCh <- errAbort
-				return
+				break
 			}
+			// There is a race that envoy proxy can exit just after ctx done, and in this case ac is -1.
 			select {
 			case status := <-a.statusCh:
-				a.statusCh <- status
+				log.Errorf("Epoch %d exited with error: %v", status.epoch, status.err)
 				return
 			default:
 				log.Infof("There are still %d active connections", ac)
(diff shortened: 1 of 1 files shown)


pkg/envoy/agent.go
```diff
@@ -155,7 +155,13 @@ func (a *Agent) terminate() {
 				a.abortCh <- errAbort
 				return
 			}
-			log.Infof("There are still %d active connections", ac)
+			select {
+			case status := <-a.statusCh:
+				a.statusCh <- status
+				return
```
rittneje, line 161: To simplify this, can `terminate` always pull from `a.statusCh` and return the result, rather than making `Run` do it afterwards?
Also, it would be nice to pull from `a.statusChan` during the `terminationDrainDuration` sleep (which should then switch to a timer).

hzxuzhonghu: I have no strong preference, done in https://github.com/istio/istio/pull/36691/commits/c8a6a428d4e77ea412ca7e4c738c37aac90738c6

howardjohn:
> If exitOnZeroActiveConnections is false, then it will pointlessly keep waiting until the terminationDrainDuration expires.

In my testing, the behavior with this PR is unchanged from `master`:
* If we kill Envoy first, everything immediately exits.
* If we send SIGTERM to pilot-agent, then kill envoy, it will wait the full terminationDrainDuration

Is this expected to change?

rittneje:
@howardjohn The current changeset is not handling the case correctly. I have left some comments to that effect.

hzxuzhonghu:
This can happen in very rare case

rittneje:
Why not just fix it though?

hzxuzhonghu:
What do you mean? I am fixing it.

rittneje:
@hzxuzhonghu I interpreted your comment to mean you didn't think it was worth fixing the `terminationDrainDuration` case. Apologies if that's not what you intended.

hzxuzhonghu:
`terminationDrainDuration ` is under hot discussion without any consensus.

Here the issue can happen in a very rare condition with nothing to do with duration. so make this a little change.

hzxuzhonghu:
ping @howardjohn

pkg/envoy/agent.go
howardjohn: instead of `range ticker.C` should we select over `a.statusCh` and `ticker.C`? Otherwise we must always wait for a tick.

howardjohn:
Is it feasible to add unit tests for this? Its a bit tricky to follow

rittneje:
@howardjohn @hzxuzhonghu I wanted to follow up on this. Currently we have an issue where Istio cannot possibly detect when ingress-nginx is done, because the inbound connections are not proxied via Istio (so `exitOnZeroActiveConnections` is useless). Hence we have no choice but to have istio delay shutdown for an arbitrary amount of time, which is extremely undesirable. For this reason, I want the ingress-nginx pre-stop hook to force kill Envoy as soon as nginx terminates. Consequently, the `terminationDrainDuration` case also needs to be fixed to account for Envoy getting killed "prematurely".

PR #36693, opened 2022-01-04
Fix analyze finding namespace error return

hanxiaop:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/istioctl/cmd/analyze.go
+++ b/istioctl/cmd/analyze.go
@@ -134,9 +134,12 @@ func Analyze() *cobra.Command {
 					return err
 				}
 				_, err = client.CoreV1().Namespaces().Get(context.TODO(), namespace, v1.GetOptions{})
-				if errors.IsNotFound(err) {
-					fmt.Fprintf(cmd.ErrOrStderr(), "namespace %q not found\n", namespace)
-					return nil
+				if err != nil {
+					if errors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


istioctl/cmd/analyze.go
```diff
@@ -134,9 +134,12 @@ func Analyze() *cobra.Command {
 					return err
 				}
 				_, err = client.CoreV1().Namespaces().Get(context.TODO(), namespace, v1.GetOptions{})
-				if errors.IsNotFound(err) {
-					fmt.Fprintf(cmd.ErrOrStderr(), "namespace %q not found\n", namespace)
-					return nil
+				if err != nil {
```
zhlsunshine, line 137: Why additional nest here for err message?

hanxiaop: If only checking the `NotFound` error here, the other types of errors like `Forbidden` error will be escaped, and it will continue to run the rest of the logic. The nested logic here is to prevent other errors being escaped.

zhlsunshine: I understand, but I don't think that the error should be triggered here if the analyzer client has no privilege to fetch a namespace from the cluster.

hanxiaop: You right. I think it doesn't matter if there are other types of errors. Thanks!

PR #36697, opened 2022-01-04
auto allocate ip while updating the store

ramaraochavali:
Auto allocate IPs during store update to avoid DeepCopy in Services() call as Services() is frequently used.

### DIFF
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
@@ -735,10 +735,9 @@ func servicesDiff(os []*model.Service, ns []*model.Service) ([]*model.Service, [
 }
 
 func servicesEqual(os *model.Service, ns *model.Service) bool {
-	// TODO(ramaraochavali): do a field level comparison and exclude autoallocated addresses.
-	s := os.DeepCopy()
+	s := *os
 	s.AutoAllocatedAddress = ""
-	return reflect.DeepEqual(s, ns)
+	return reflect.DeepEqual(&s, ns)
 }
 
 // Automatically allocates IPs for service entry services WITHOUT an
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
@@ -735,6 +735,8 @@ func servicesDiff(os []*model.Service, ns []*model.Service) ([]*model.Service, [
 }
 
 func servicesEqual(os *model.Service, ns *model.Service) bool {
+	// Disabling `go vet` warning since this is actually safe in this case.
+	// nolint: vet
 	s := *os
 	s.AutoAllocatedAddress = ""
 	return reflect.DeepEqual(&s, ns)
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
@@ -516,16 +516,7 @@ func (s *ServiceEntryStore) Services() ([]*model.Service, error) {
 	s.mutex.RLock()
 	allServices := s.services.getAllServices()
 	s.mutex.RUnlock()
-
-	out := make([]*model.Service, 0, len(allServices))
-	for _, svc := range allServices {
-		// TODO: eliminate the deepcopy here
-		// autoAllocateIPs will re-allocate ips for the service,
-		// if return the pointer directly, there will be a race with `BuildNameTable`
-		out = append(out, svc.DeepCopy())
-	}
-	autoAllocateIPs(out)
-	return out, nil
+	return allServices, nil
 }
 
 // GetService retrieves a service by host name if it exists.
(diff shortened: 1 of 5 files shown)


ramaraochavali:
I am still working on this. not ready for review yet.

howardjohn (dismissed):
My understanding is:

Before: `autoAllocateIPs` is called with ALL services
After: `autoAllocateIPs` is called many times with a few services

`autoAllocateIPs` starts at the same base IP each time. Calling it multiple times will lead to overlapping allocated IPs?

ramaraochavali:
> autoAllocateIPs starts at the same base IP each time. Calling it multiple times will lead to overlapping allocated IPs?

Yes. Conversion is not the right place. I am changing to do at the time when we add a new service for all services.

hzxuzhonghu:
If we want to allocate once per service, the right direction should be ipadm, and also be care of sync between multi instances

ramaraochavali:
> If we want to allocate once per service, the right direction should be ipadm, and also be care of sync between multi instances

I am not planning to do that but allocate them all at once today but only do when services are added to store rather than when they are retrieved to avoid deepcopy and repeated autoallocations as gets are lot more than updates

ramaraochavali:
@hzxuzhonghu @howardjohn This is ready for review except for one TODO I left in the code which I will finish it before merging. But I want you to look at overall approach and see if it makes sense?

howardjohn (commented):
Makes sense to me. Do we have any good coverage of this feature though? Not sure if we have tests that would catch assigning the same IP twice?

hzxuzhonghu:
Will have a look, this is hot path

pilot/pkg/serviceregistry/serviceentry/store.go
```diff
@@ -170,4 +170,5 @@ func (s *serviceStore) deleteServices(key types.NamespacedName) {
 
 func (s *serviceStore) updateServices(key types.NamespacedName, services []*model.Service) {
 	s.servicesBySE[key] = services
+	autoAllocateIPs(s.getAllServices())
```
hzxuzhonghu, line 173: lgtm

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -744,6 +734,13 @@ func servicesDiff(os []*model.Service, ns []*model.Service) ([]*model.Service, [
 	return added, deleted, updated, unchanged
 }
 
+func servicesEqual(os *model.Service, ns *model.Service) bool {
+	// TODO(ramaraochavali): do a field level comparison and exclude autoallocated addresses.
+	s := os.DeepCopy()
```
hzxuzhonghu, line 739: no need deepcopy here, a shallow copy is ok


copied:=*os
copied.AutoAllocatedAddress = ""

ramaraochavali: i tried - shallow copy is not working as it has locks and also some tests are failing with shallow copy

ramaraochavali: I actually think DeepCopy is not costly here as we do only during add/updates - not during gets.

hzxuzhonghu: With the patch below, i can not find any failure in servicediscovery_test.go
```
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
@@ -728,7 +728,7 @@ func servicesDiff(os []*model.Service, ns []*model.Service) ([]*model.Service, [
                newSvc, f := newServiceHosts[s.Hostname]
                if !f {
                        deleted = append(deleted, s)
-               } else if !reflect.DeepEqual(s, newSvc) {
+               } else if !servicesEqual(s, newSvc) {
                        updated = append(updated, newSvc)
                } else {
                        unchanged = append(unchanged, newSvc)
@@ -744,6 +744,12 @@ func servicesDiff(os []*model.Service, ns []*model.Service) ([]*model.Service, [
        return added, deleted, updated, unchanged
 }

+func servicesEqual(os *model.Service, ns *model.Service) bool {
+       s := *os
+       s.AutoAllocatedAddress = ""
+       return reflect.DeepEqual(&s, ns)
+}
+

```

ramaraochavali: Strange... It failed some tests and now it passes.

ramaraochavali:
@howardjohn @hzxuzhonghu I think this is ready for review. Can you PTAL?

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -744,6 +734,14 @@ func servicesDiff(os []*model.Service, ns []*model.Service) ([]*model.Service, [
 	return added, deleted, updated, unchanged
 }
 
+func servicesEqual(os *model.Service, ns *model.Service) bool {
+	// Disabling `go vet` warning since this is actually safe in this case.
+	// nolint: vet
+	s := *os
+	s.AutoAllocatedAddress = ""
+	return reflect.DeepEqual(&s, ns)
```
howardjohn, line 742: Is this going to compare things like mutex, etc? Seems odd... maybe mutex just is always considered equal though. I guess we already have it

howardjohn (commented):
lgtm but @hzxuzhonghu can approve since I think they wanted to review

ramaraochavali:
@howardjohn can you please remove the requested changes as well?

PR #36706, opened 2022-01-05
Support rewriting gRPC probes

### LINKED ISSUE #36314: Add support for gRPC probe
Our probe rewrite needs to handle https://github.com/kubernetes/enhancements/issues/2727

dddddai:
Support rewriting gRPC probes for Kubernetes v1.23
Fix #36314
 
I'm not very familiar with the integration test, but I've tested it in my local environment, and here's the result:

#### 1. Apply strict mtls:
```sh
kubectl apply -f - <<EOF
kind: PeerAuthentication
metadata:
  name: "default"
  namespace: "istio-system"
spec:
  mtls:
    mode: STRICT
EOF
```
#### 2. Create a pod with a grpc probe

Before rewriting([an example in k8s doc](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe "example in k8s doc")):
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: etcd-with-grpc
spec:
  containers:
  - name: etcd
    image: k8s.gcr.io/etcd:3.5.1-0
    command: [ "/usr/local/bin/etcd", "--data-dir",  "/var/lib/etcd", "--listen-client-urls", "http://0.0.0.0:2379", "--advertise-client-urls", "http://127.0.0.1:2379", "--log-level", "debug"]
    ports:
    - containerPort: 2379
    readinessProbe:
      grpc:
        port: 2379
      initialDelaySeconds: 10
```

After rewriting:
```yaml
...........
      readinessProbe:
        failureThreshold: 3
        httpGet:
          path: /app-health/etcd/readyz
          port: 15020
          scheme: HTTP
        initialDelaySeconds: 10
        periodSeconds: 10
        successThreshold: 1
        timeoutSeconds: 1
...........
      - name: ISTIO_KUBE_APP_PROBERS
        value: '{"/app-health/etcd/readyz":{"grpc":{"port":2379,"service":""},"timeoutSeconds":1}}'
      image: docker.io/istio/proxyv2:1.12.1
...........
```

And the container stays ready
```sh
[root@new grpc-health]# kubectl get po
NAME             READY   STATUS    RESTARTS   AGE
etcd-with-grpc   2/2     Running   0          5m3s
```

#### 3. Set rewrite to `false` and recreate the pod
```yaml
......
metadata:
  annotations: 
    sidecar.istio.io/rewriteAppHTTPProbers: "false"
......
```

Turns out the probe would not be rewrited:
```yaml
      readinessProbe:
        failureThreshold: 3
        grpc:
          port: 2379
          service: ""
        initialDelaySeconds: 10
        periodSeconds: 10
        successThreshold: 1
        timeoutSeconds: 1
```

The container would become unready
```sh
[root@new grpc-health]# kubectl get po
NAME                             READY   STATUS    RESTARTS   AGE
etcd-7c7d558878-bv8dd            1/2     Running   0          4m22s
```

### DIFF
--- a/pilot/cmd/pilot-agent/status/server.go
+++ b/pilot/cmd/pilot-agent/status/server.go
@@ -40,6 +40,11 @@ import (
 	"github.com/prometheus/client_golang/prometheus/collectors"
 	"github.com/prometheus/common/expfmt"
 	"go.opencensus.io/stats/view"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/insecure"
+	grpcHealth "google.golang.org/grpc/health/grpc_health_v1"
+	grpcStatus "google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	"istio.io/istio/pilot/cmd/pilot-agent/metrics"
@@ -99,6 +104,7 @@ type KubeAppProbers map[string]*Prober
 type Prober struct {
 	HTTPGet        *apimirror.HTTPGetAction   `json:"httpGet,omitempty"`
 	TCPSocket      *apimirror.TCPSocketAction `json:"tcpSocket,omitempty"`
+	GRPC           *apimirror.GRPCAction      `json:"grpc,omitempty"`
 	TimeoutSeconds int32                      `json:"timeoutSeconds,omitempty"`
 }
 
@@ -278,11 +284,18 @@ func validateAppKubeProber(path string, prober *Prober) error {
 	if !appProberPattern.Match([]byte(path)) {
 		return fmt.Errorf(`invalid path, must be in form of regex pattern %v`, appProberPattern)
 	}
-	if prober.HTTPGet == nil && prober.TCPSocket == nil {
-		return fmt.Errorf(`invalid prober type, must be of type httpGet or tcpSocket`)
+	count := 0
+	if prober.HTTPGet != nil {
+		count++
 	}
-	if prober.HTTPGet != nil && prober.TCPSocket != nil {
-		return fmt.Errorf(`invalid prober, type must be either httpGet or tcpSocket`)
+	if prober.TCPSocket != nil {
+		count++
+	}
+	if prober.GRPC != nil {
+		count++
+	}
+	if count != 1 {
+		return fmt.Errorf(`invalid prober type, must be one of type httpGet, tcpSocket or gRPC`)
 	}
 	if prober.HTTPGet != nil && prober.HTTPGet.Port.Type != intstr.Int {
 		return fmt.Errorf("invalid prober config for %v, the port must be int type", path)
@@ -621,11 +634,13 @@ func (s *Server) handleAppProbe(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	if prober.HTTPGet != nil {
+	switch {
+	case prober.HTTPGet != nil:
 		s.handleAppProbeHTTPGet(w, req, prober, path)
-	}
-	if prober.TCPSocket != nil {
+	case prober.TCPSocket != nil:
 		s.handleAppProbeTCPSocket(w, prober)
+	case prober.GRPC != nil:
+		s.handleAppProbeGRPC(w, req, prober)
 	}
 }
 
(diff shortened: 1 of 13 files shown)


hzxuzhonghu:
Cool, i just learned grpc prove has been supported in k8s 1.23

howardjohn (commented):
Nice! Just took a quick look and looks good at a high level. We will want an integ test for this - we have ones for the other protocols so should be able to expand them for this. I can give some pointers for that

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -714,6 +728,66 @@ func (s *Server) handleAppProbeTCPSocket(w http.ResponseWriter, prober *Prober)
 	}
 }
 
+func (s *Server) handleAppProbeGRPC(w http.ResponseWriter, req *http.Request, prober *Prober) {
+	opts := []grpc.DialOption{
+		grpc.WithBlock(),
+		// nolint: staticcheck
```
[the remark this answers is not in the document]
dddddai: I was using `grpc.WithInsecure`, the lint check showed:
```
pilot/cmd/pilot-agent/status/server.go:734:3: SA1019: grpc.WithInsecure is deprecated: use insecure.NewCredentials() instead. Will be supported throughout 1.x. (staticcheck)
		grpc.WithInsecure(), // credentials are currently not supported
		^
```

Then I changed to `insecure.NewCredentials()`, but it reports error:
```
pilot\cmd\pilot-agent\status\server.go:736:26: cannot use insecure.NewCredentials() (type credentials.TransportCredentials) as type grpc.DialOption in slice literal:
	credentials.TransportCredentials does not implement grpc.DialOption (missing grpc.apply method)
```

And [kubelet](https://github.com/kubernetes/kubernetes/blob/v1.23.1/pkg/probe/grpc/grpc.go#L58) also uses `grpc.WithInsecure()`

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -714,6 +728,66 @@ func (s *Server) handleAppProbeTCPSocket(w http.ResponseWriter, prober *Prober)
 	}
 }
 
+func (s *Server) handleAppProbeGRPC(w http.ResponseWriter, req *http.Request, prober *Prober) {
+	opts := []grpc.DialOption{
```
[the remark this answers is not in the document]
dddddai: Yeah it's exactly what Kubernetes does, the link is added below:
https://github.com/istio/istio/blob/99342beb223d1aa53368cb2f72ea5994639bf854/pilot/cmd/pilot-agent/status/server.go#L738-L741

dddddai:
@howardjohn Thank you! I've addressed your comments
> We will want an integ test for this - we have ones for the other protocols so should be able to expand them for this.

OK, I will take a stab

pilot/cmd/pilot-agent/status/server.go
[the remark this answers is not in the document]
hzxuzhonghu: We should explicitly comment the code is copied from k8s, otherwise there is license issue

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -714,6 +728,66 @@ func (s *Server) handleAppProbeTCPSocket(w http.ResponseWriter, prober *Prober)
...
+		grpc.WithBlock(),
+		// nolint: staticcheck
+		grpc.WithInsecure(), // credentials are currently not supported
+	}
+	if userAgent := req.Header["User-Agent"]; len(userAgent) > 0 {
```
hzxuzhonghu, line 737: Where is the header come from?

dddddai: It's from [kubelet](https://github.com/kubernetes/kubernetes/blob/v1.23.1/pkg/probe/http/http.go#L103)

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -714,6 +728,66 @@ func (s *Server) handleAppProbeTCPSocket(w http.ResponseWriter, prober *Prober)
...
+			}
+		} else {
+			log.Errorf("grpc probe failed: %v", err)
+		}
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	if resp.GetStatus() == grpcHealth.HealthCheckResponse_SERVING {
+		w.WriteHeader(http.StatusOK)
```
hzxuzhonghu, line 785: return

pkg/kube/apimirror/probe.go
```diff
@@ -70,3 +70,17 @@ type TCPSocketAction struct {
 	// +optional
 	Host string `json:"host,omitempty" protobuf:"bytes,2,opt,name=host"`
 }
+
+// GRPCAction describes an action based on GRPC health check
+type GRPCAction struct {
```
hzxuzhonghu, line 75: redundant, why not use k8s directly?

pilot/cmd/pilot-agent/status/server.go
dddddai: Why is return needed? there's no more instructions below

hzxuzhonghu (commented):
overall lgtm

pkg/kube/apimirror/probe.go
dddddai: As the comment says above:
https://github.com/istio/istio/blob/1f6144eb21303a6aff9a1214eae564a81b603adb/pkg/kube/apimirror/probe.go#L15-L18

pilot/cmd/pilot-agent/status/server.go
hzxuzhonghu: So can get rid of else

pilot/cmd/pilot-agent/status/server.go
hzxuzhonghu: IC

pilot/cmd/pilot-agent/status/server.go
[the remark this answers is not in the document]
dddddai: > We should explicitly comment the code is copied from k8s, otherwise there is license issue

OK, I've added reference comments

dddddai:
@hzxuzhonghu Thank you!

pkg/kube/inject/app_probe_test.go
```diff
@@ -110,3 +113,272 @@ func TestShouldRewriteAppHTTPProbers(t *testing.T) {
...
+		},
+	} {
+		patchRewriteProbe(annotations, tc.pod, 15020)
+		if !reflect.DeepEqual(tc.pod, tc.expected) {
+			t.Errorf("[%v] failed, want %v, got %v", tc.name, tc.expected, tc.pod)
+		}
+	}
+}
+
+func jsonEqual(s1, s2 string) bool {
```
hzxuzhonghu, line 370: why is this so complex? 

The k8s api does not have omitempty? 
> Service *string `json:"service" protobuf:"bytes,2,opt,name=service"`

dddddai: > why is this so complex?

Sorry, what do you mean by complex?

> The k8s api does not have omitempty?

Oh yes, it's my bad

hzxuzhonghu (approved):

pilot/cmd/pilot-agent/status/server.go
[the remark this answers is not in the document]
howardjohn: Should be `grpc.WithTransportCredentials(insecure.NewCredentials())`. gRPC only uses it since they are on an older gRPC version I think, they are functionally equivilent.

pkg/kube/inject/webhook.go
```diff
@@ -590,17 +590,19 @@ func reorderPod(pod *corev1.Pod, req InjectionParameters) error {
...
 
 	rewrite := ShouldRewriteAppHTTPProbers(pod.Annotations, valuesStruct.GetSidecarInjectorWebhook().GetRewriteAppHTTPProbe())
-	sidecar := FindSidecar(pod.Spec.Containers)
-
 	// We don't have to escape json encoding here when using golang libraries.
-	if rewrite && sidecar != nil {
+	if rewrite {
 		if prober := DumpAppProbers(&pod.Spec, req.meshConfig.GetDefaultConfig().GetStatusPort()); prober != "" {
 			sidecar.Env = append(sidecar.Env, corev1.EnvVar{Name: status.KubeAppProberEnvName, Value: prober})
 		}
```
howardjohn, line 608: For adding integ test:

* Update prow/config/mixedlb-service.yaml to add `GRPCContainerProbe` feature gate. I wouldn't mind if this file name was changed to just be `default` or something.
* Add a new test like `TestTcpProbe` but for gRPC.

dddddai:
There's a problem that the rewrited probe does not work with strict mtls, the agent reports the [error](https://github.com/istio/istio/blob/d08ed217b216fbec0e3abd5c70f65fb1d1e37cff/pilot/cmd/pilot-agent/status/server.go#L753):
```
Failed to create grpc connection to probe app: context deadline exceeded
```
I'm looking into it, so far have no idea what's going wrong

pkg/kube/inject/webhook.go
dddddai: @howardjohn Thanks for elaborating! I've added an integ test and it works fine, PTAL

tests/integration/pilot/grpc_probe_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+var etcdWithGRPC = `
+apiVersion: v1
+kind: Pod
+metadata:
+  name: etcd-with-grpc
+spec:
+  containers:
+    - name: etcd
+      image: 'k8s.gcr.io/etcd:3.5.1-0'
```
howardjohn, line 41: Should not use etcd, we have a standardized test image used in ~all tests. It supports gRPC - although probably not the health service. We can definitely add that to `pkg/test/echo/server/endpoint/grpc.go`, and then use that like the tcp probe

dddddai: Done

pkg/test/echo/server/endpoint/grpc.go
```diff
@@ -113,6 +115,11 @@ func (s *grpcInstance) Start(onReady OnReadyFunc) error {
 	}
 	s.server = s.newServer(opts...)
 
+	// add the standard grpc health check
+	healthServer := health.NewServer()
+	healthServer.SetServingStatus("", grpcHealth.HealthCheckResponse_SERVING)
```
howardjohn, line 120: should this be set in `onReady`? otherwise it will return `HEALTHY` before it is

howardjohn (approved):

pkg/test/echo/server/endpoint/grpc.go
dddddai: Yes, fixed

Issue #36709, opened 2022-01-05
Unify xds proxy and sds client tls config setting


hzxuzhonghu:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**

Currently the the grpc dial options they are using are built similarly, which is very duplicate, so we should merge them. 
And by doing this, it will be helpful for maintenance.

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

howardjohn:
In theory they could share the same connection as well. Although it only works in some subset of cases so probably super complex

hzxuzhonghu:
not stale

hzxuzhonghu:
not stale

howardjohn:
https://github.com/istio/istio/pull/42720

PR #36730, opened 2022-01-06
unify key/certs process used to connect to Citadel and xds server

hzxuzhonghu:

Try to address https://github.com/istio/istio/issues/36709

### DIFF
--- a/security/pkg/nodeagent/caclient/providers/citadel/client.go
+++ b/security/pkg/nodeagent/caclient/providers/citadel/client.go
@@ -31,6 +31,7 @@ import (
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/metadata"
+
 	pb "istio.io/api/security/v1alpha1"
 	"istio.io/istio/pkg/security"
 	"istio.io/istio/security/pkg/nodeagent/caclient"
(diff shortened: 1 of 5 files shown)


security/pkg/nodeagent/caclient/providers/citadel/client.go
```diff
@@ -112,58 +117,37 @@ func (c *CitadelClient) CSRSign(csrPEM []byte, certValidTTLInSec int64) ([]strin
...
-				}
+			key, cert := c.tlsOpts.Key, c.tlsOpts.Cert
+			if cert != "" {
 				var isExpired bool
-				isExpired, err = c.isCertExpired(filepath.Join(c.opts.ProvCert, "cert-chain.pem"))
+				isExpired, err = c.isCertExpired(cert)
 				if err != nil {
 					citadelClientLog.Warnf("cannot parse the cert chain, using token instead: %v", err)
-					return &tls.Certificate{}, nil
+					return &certificate, nil
```
howardjohn, line 133: This seems to change behavior? Before it always returns empty cert. Now it may return `certificate` which is set by previous call to this function

howardjohn: Same on L137

hzxuzhonghu: sorry, my mistake, I meant to define `certificate` under the callback

howardjohn (approved):

PR #36740, opened 2022-01-07
Use a single defaultWatcher per cluster

howardjohn:
Avoid multiple runs doing the same thing

### DIFF
--- a/pkg/revisions/default_watcher.go
+++ b/pkg/revisions/default_watcher.go
@@ -60,6 +60,7 @@ func NewDefaultWatcher(client kube.Client, revision string) DefaultWatcher {
 		revision: revision,
 		mu:       sync.RWMutex{},
 	}
+	p.queue = controllers.NewQueue("default revision", controllers.WithReconciler(p.setDefault))
 	p.webhookInformer = client.KubeInformer().Admissionregistration().V1().MutatingWebhookConfigurations().Informer()
 	p.webhookInformer.AddEventHandler(controllers.FilteredObjectHandler(p.queue.AddObject, isDefaultTagWebhook))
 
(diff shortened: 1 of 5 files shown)


pkg/revisions/default_watcher.go
```diff
@@ -71,6 +72,12 @@ func (p *defaultWatcher) Run(stopCh <-chan struct{}) {
 		log.Errorf("failed to sync default watcher")
 		return
 	}
+	p.mu.Lock()
+	if p.started {
+		return
```
hzxuzhonghu, line 77: unlock

hzxuzhonghu (commented):
I read the default watcher carefully, and find that it maybe better to just remove the queue from it since it is only reading a revision.  And keep instantiating it in each leader election call

pkg/revisions/default_watcher.go
howardjohn: Thanks. refactored to use `defer`

howardjohn:
> I read the default watcher carefully, and find that it maybe better to just remove the queue from it since it is only reading a revision. And keep instantiating it in each leader election call

The work it does is trivial, definitely. Do you know if k8s recommends/uses queues _always_, or only for non-trivial tasks?

I do still think its better to not instantiate multiple times per call though. That would cause multiple handlers on the informer still, which in theory could fall out of sync and cause weird behavior. It _should_ work fine but doesn't seem ideal.

howardjohn:
I think we need to create it only if we are going to Run it. Creating the queue launches a goroutine and starts filling up the queue. if we never run it, the queue will grow unbounded

howardjohn:
closing since its harder than expected

PR #36744, opened 2022-01-07
global object needs to be accepted, when gateway is used as dependency

varet80:
In one of our setups, Gateway is used as a helm dependency.
Helmcharts populate an object `global` which is passed through all dependencies. By not including this in the schema the following error occures:
```
helm.go:88: [debug] values don't meet the specifications of the schema(s) in the following chart(s):
gateway:
- (root): Additional property global is not allowed

```

This change tries to fix that

- [x] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/manifests/charts/gateway/values.schema.json
+++ b/manifests/charts/gateway/values.schema.json
@@ -174,6 +174,9 @@
     },
     "networkGateway": {
       "type": "string"
+    },
+    "global": {
+      "type": "object"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


zirain:
Is this duplicate with https://github.com/istio/istio/pull/35496 ?

billiaz:
both achieve the same result @zirain but this one is more strict. As it accepts only global object. Which is the case of subchart.

this means, it keeps strict the rest of values. where the other patch, is removes the strict structure

billiaz:
Seems duplicate

PR #36746, opened 2022-01-07
WIP: Simplify egress mTLS config - Phase 1

kfaseela:

- [ ] Docs
- [ ] Installation
- [ X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu (commented):
LG, but i think the reference of secret should be put high priority and along with it.

hzxuzhonghu (approved):
As there is authorization for sds, i am ok with this. https://github.com/istio/istio/blob/master/pilot/pkg/credentials/kube/secrets.go#L190

howardjohn (changes requested):
I am worried about merging this in the current state. DestinationRules generally apply cluster wide. What this means is that if I set up a seemingly innocent DR like:
```
kind: DestinationRule
metadata:
  name: istiod
spec:
  host: istiod.istio-system.svc.cluster.local
  trafficPolicy:
    tls:
      mode: SIMPLE
      credentialName: sds-credential
```

Any pods that do not have `sds-credential` in their namespace (ie I need to copy it to ALL namespaces) and do not have RBAC setup (again, all namespaces will need this) will have 15s delayed startup:

```
2022-01-12T17:02:34.477756Z     warning envoy config    gRPC config: initial fetch timed out for type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.Secret
2022-01-12T17:02:35.036642Z     info    Readiness succeeded in 16.011891106s
```

This also will impact existing users of egressgateway TLS origination. Essentially when they upgrade all of their pods will suddenly take +15s to startup.

hzxuzhonghu:
This is something like kubelet does(it can watch all cm/secret/pvc), i am just not sure we should give the sidecar so high permission

Issue #36750, opened 2022-01-07
Traces being sampled when sample rate set to 0


Stono:
### Bug Description

We have the following config

```
data:
  mesh: |-
    defaultConfig:
      tracing:
        max_path_tag_length: 1024
        sampling: 0
        zipkin:
          address: zipkin.istio-system.svc.cluster.local:9411
```

However we're still seeing some spans being sampled.

Doing a config dump shows:
```
 "tracing": {
             "client_sampling": {
              "value": 100
             },
             "random_sampling": {
              "value": 1
             },
             "overall_sampling": {
              "value": 100
             },
```

So it appears the config isn't working

@howardjohn pointed me at: https://github.com/istio/istio/blame/8b5e6505c28feedea342b0b912741fe9c77bd904/pilot/pkg/networking/core/v1alpha3/tracing.go#L447 which explains it, that 0 effectively can't be used.

For context; we don't want istio to decide what we sample - that's what we do with our opentracing libraries at various points in our infrastructure.  All we want istio to do is report the spans if `x-b3-sampled: 1`. 

This is how we've had istio configured for years and it did used to work, because we only want traces for requests that originate through nginx (subsequently correlated to a client request), and we record 100% of those.  We don't care for internal tracing eg, some kafka consumer calling some other service.  And because that's recording just 1% anyway, it confuses our users when they're stumbling across spans for some, but not all of their apps requests (remember, they're used to it being 100% of client requests).

A somewhat cruddy work around is to set it to 0.1, which reduces the amount of noise we store, but obviously doesn't stop it.

Please can we regain the ability to have istio just report spans based on the tracing headers, but never attempt to create traces or spans (so effectively a sample rate of 0) - eg the ability to set `random_sampling` to 0.

Thanks

### Version

```prose
1.12
```

howardjohn:
cc @douglas-reid . Any easy fix is to use a protobuf wrapper to distinguish null from `0`, but I don't know if we can do that in a backwards compatible way?

douglas-reid:
@Stono @howardjohn that code was lifted from the existing code, so that bit should not have changed. Is the issue that it is defaulting to the pilot value of 1.0?  Can we (as a workaround) set the pilot traceSampling to 0 at install ?

Stono:
> Is the issue that it is defaulting to the pilot value of 1.0

[...]

PR #36751, opened 2022-01-07
tracing: allow specification of no sampling with reporting via Telemetry API

douglas-reid:
The Telemetry API documents that the default value for `randomSamplingPercentage` is `0.0`. This change will allow the control plane to respect that documentation (and user intent).

With this PR, Telemetry API users will no longer fallback to pilot environment variables, etc., when the value is not specified.

Related issue: https://github.com/istio/istio/issues/36750

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/tracing.go
+++ b/pilot/pkg/networking/core/v1alpha3/tracing.go
@@ -60,7 +60,7 @@ func configureTracingFromSpec(tracing *model.TracingConfig, opts buildListenerOp
 		}
 		// use the prior configuration bits of sampling and custom tags
 		hcm.Tracing = &hpb.HttpConnectionManager_Tracing{}
-		configureSampling(hcm.Tracing, 0.0, proxyCfg)
+		configureSampling(hcm.Tracing, proxyConfigSamplingValue(proxyCfg), proxyCfg)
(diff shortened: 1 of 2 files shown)


pilot/pkg/networking/core/v1alpha3/tracing.go
```diff
@@ -419,29 +419,19 @@ func buildServiceTags(metadata *model.NodeMetadata) []*tracing.CustomTag {
 	}
 }
 
-func configureSampling(hcmTracing *hpb.HttpConnectionManager_Tracing, providerPercentage float64, proxyCfg *meshconfig.ProxyConfig) {
+func configureSampling(hcmTracing *hpb.HttpConnectionManager_Tracing, providerPercentage float64) {
```
hzxuzhonghu, line 422: seems the second param passed in is not from provider

hzxuzhonghu:
cc @zirain

pilot/pkg/networking/core/v1alpha3/tracing.go
zirain: see proxyConfigSamplingValue(proxyCfg)

douglas-reid:
@howardjohn any thoughts here?  I'd like to get this functionality (whether in this form or not) in before the 1.13 branch cut.

howardjohn (approved):

PR #36756, opened 2022-01-08
add listener tcp metadata exchange filter only when enabled

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


bianpengyuan (commented):
I think this will not remove mx from alpn, which will cause issue? Also probably should also control the cluster network filter and alpn at upstream context.

ramaraochavali:
> I think this will not remove mx from alpn, which will cause issue

I think cluster has same alpns but we control with flag. It does not create issue. So should be fine

> Also probably should also control the cluster network filter and alpn at upstream context.

This is already controlled

bianpengyuan:
> I think cluster has same alpns but we control with flag. It does not create issue. So should be fine

But still it is incorrect config. The listener will claim something that it is not capable of. I can think of cases where someone switch this env var from true to false, and in the transition state, some cluster might still have mx in alpn, and connect with listeners which has mx in alpn but no mx filter. Request / TCP connection will fail during that brief period.

ramaraochavali:
> But still it is incorrect config. The listener will claim something that it is not capable of. I can think of cases where someone switch this env var from true to false, and in the transition state, some cluster might still have mx in alpn, and connect with listeners which has mx in alpn but no mx filter. Request / TCP connection will fail during that brief period.

Is n't that the case already? This PR is trying to fix that inconsistency.

bianpengyuan:
I don't think `features.MetadataExchange` controls listener alpn, so this PR actually make the filter chain and its alpn claim inconsistent? (i.e. alpn has `istio-peer-exchange` but no network mx filter in the chain). And looking at the code, it seems the same problem already exists at cluster, that `istio-peer-exchange` is always added to the alpn but client side tcp mx is controlled by `features.MetadataExchange`.

ramaraochavali:
> And looking at the code, it seems the same problem already exists at cluster, that istio-peer-exchange is always added to the alpn but client side tcp mx is controlled by features.MetadataExchange.

This is what I meant. If this is a problem, I think we should control ALPNs also with this flag?

bianpengyuan:
> This is what I meant. If this is a problem, I think we should control ALPNs also with this flag?

Yeah I think we should fix that instead of deepening the inconsistency.

ramaraochavali:
@bianpengyuan Fixed  - ALPN is also controlled by that flag. PTAL.

ramaraochavali:
@bianpengyuan Can you PTAL when you get chance?

bianpengyuan (approved):
LGTM, Sorry for the delay.

ramaraochavali:
@bianpengyuan Np

@howardjohn Can you please review this as well?

hzxuzhonghu (approved):

howardjohn (approved):

ramaraochavali:
/ test integ-pilot-k8s-tests_istio

Issue #36507, opened 2021-12-14
Istio Ingress Gateway is healthy in case of "certificate signed by unknown authority" error


qurname2:
### Bug Description

Hello!

We're using custom, issued by our own CA, certificates in Istio.
During the last certificates changing one of our istio-ingress-gateway pods wasn't restarted (it must be done for the correct work) due to human error. It led to the error messages in log:
`{"level":"error","time":"2021-12-13T13:24:04.890986Z","scope":"xdsproxy","msg":"failed to create upstream grpc client: rpc error: code = Unavailable desc = connection error: desc = \"transport: authentication handshake failed: x509: certificate signed by unknown authority\""}`
It's ok and I know how to fix it - just do a restart of this pod and it's all. But I noticed one interesting moment, the **readiness probe** of this pod is **healthy**. 
To my mind, it's not correct behaviour, cause in fact pod can't produce customer's traffic.
```
  readinessProbe:
    failureThreshold: 30
    httpGet:
      path: /healthz/ready
      port: 15021
      scheme: HTTP
     initialDelaySeconds: 10
     periodSeconds: 2
     successThreshold: 1
     timeoutSeconds: 1
```

As a workaround, I wanted to create an **alert** about such behaviour of istio-ingress-gateway pod, but I didn't find any istio metrics which can describe that pod isn't healthy due to "a certificate signed by unknown authority" error. 
If I missed smth, could somebody point me out to the right metric?

### Version

```prose
istioctl version
client version: 1.8.2
control plane version: 1.8.2
data plane version: 1.8.2 (13 proxies)

kubectl version --short
Client Version: v1.16.1
Server Version: v1.16.9
```

GregHanson:
@qurname2 Does `istioctl proxy-status` show everything as `SYNCED` for the ingress gateway in this failed state? Or do any secrets show as `WARMING` for `istioctl proxy-config secrets <ingress_pod_name> -n istio-system`

bianpengyuan:
I guess what would help here is a metric reported by agent on whether xds connection is up, which you can alert on if there is any unconnected proxy?

qurname2:
@GregHanson the affected ingress gateway doesn't show in istioctl ps output, it's not SYNCED

qurname2:
@bianpengyuan, yes, exactly, it will help a lot in such a case.
[...]

PR #36764, opened 2022-01-10
Increase metric when xds connection error

zirain:

Increase metric when xds connection error, fix https://github.com/istio/istio/issues/36507


hzxuzhonghu (commented):
can you also do this for delta xds?

zirain:
> can you also do this for delta xds?

forgot that, will do it ASAP.

hzxuzhonghu (approved):

PR #36774, opened 2022-01-10
Fix empty variable check

howardjohn:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/tools/docker
+++ b/tools/docker
@@ -21,7 +21,7 @@ ROOT=$(dirname "$WD")
 set -eu
 
 # LOCAL_OUT is set when runing in a Makefile and set to current architecture (not host)
-if [[ ! -z "${LOCAL_OUT}" ]]; then
+if [[ ! -z "${LOCAL_OUT+.}" ]]; then
   OS_ARCH=${LOCAL_OUT##*/}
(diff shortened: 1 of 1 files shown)


tools/docker
```diff
@@ -21,7 +21,7 @@ ROOT=$(dirname "$WD")
 set -eu
 
 # LOCAL_OUT is set when runing in a Makefile and set to current architecture (not host)
-if [[ ! -z "${LOCAL_OUT}" ]]; then
+if [[ ! -z "${LOCAL_OUT+.}" ]]; then
```
ericvn, line 24: Won't this always return true? "$LOCAL_OUT+." should always be a string do -z should return false, and then we make that true?

johnma14 (approved):

ericvn (changes requested):
Hold to understand the emptiness.

howardjohn:
```
$ echo "${x+.}"

$ x=1
$ echo "${x+.}"
.
```

seems right to me?

ericvn (approved):
Talked with @howardjohn and determined that the desired behavior is to use the else for the case where LOCAL_OUT is an empty string.

PR #36778, opened 2022-01-10
allow analyze to ignore malformed files

### LINKED ISSUE #36471: Analyze always errors on unrecognized files.

therealmitchconnors:
resolves #36471

### DIFF
--- a/istioctl/cmd/analyze.go
+++ b/istioctl/cmd/analyze.go
@@ -73,6 +73,7 @@ var (
 	suppress          []string
 	analysisTimeout   time.Duration
 	recursive         bool
+	ignoreInvalid     bool
 
 	fileExtensions = []string{".json", ".yaml", ".yml"}
 )
@@ -279,7 +280,7 @@ func Analyze() *cobra.Command {
 			var returnError error
 			if msgOutputFormat == formatting.LogFormat {
 				returnError = errorIfMessagesExceedThreshold(result.Messages)
-				if returnError == nil && parseErrors > 0 {
+				if returnError == nil && parseErrors > 0 && !ignoreInvalid {
 					returnError = FileParseError{}
 				}
 			}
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

istioctl/cmd/analyze.go
```diff
@@ -313,6 +314,8 @@ func Analyze() *cobra.Command {
 		"The duration to wait before failing")
 	analysisCmd.PersistentFlags().BoolVarP(&recursive, "recursive", "R", false,
 		"Process directory arguments recursively. Useful when you want to analyze related manifests organized within the same directory.")
+	analysisCmd.PersistentFlags().BoolVar(&ignoreInvalid, "ignore-invalid", false,
+		"Ignores invalid input documents, for cases where analyze should run only on k8s compliant inputs.")
```
esnible, line 318:
```suggestion
		"Don't complain about non-YAML, non-Kubernetes files.")
```

therealmitchconnors: In this case, we are handling both yaml and non-yaml files.  Files with non-yaml extensions are already excluded in the gather function.  This section handles files with the yaml extension which cannot be parsed by the k8s go-client.  One example would be an empty yaml document (usually part of a multi-document file, with sections separated by `---`).

esnible: @therealmitchconnors My suggestion didn't say non-.yaml, it said non-YAML.  Feel free to ignore.  It is a small point of style, that flag descriptions should say what the flag does rather than how it works (`Ignores`).

istioctl/cmd/analyze.go
```diff
@@ -313,6 +314,8 @@ func Analyze() *cobra.Command {
 		"The duration to wait before failing")
 	analysisCmd.PersistentFlags().BoolVarP(&recursive, "recursive", "R", false,
 		"Process directory arguments recursively. Useful when you want to analyze related manifests organized within the same directory.")
+	analysisCmd.PersistentFlags().BoolVar(&ignoreInvalid, "ignore-invalid", false,
```
hzxuzhonghu, line 317: i guess invalid here means unknown

hzxuzhonghu (approved):

istioctl/cmd/analyze.go
therealmitchconnors: I have updated the terminology to say what the flag does.

istioctl/cmd/analyze.go
therealmitchconnors: updated.

esnible (approved):

jacob-delgado (approved):

PR #36794, opened 2022-01-11
Do not fail test if envoy stats has non-float value

howardjohn:
Can be a string (and is in new Envoy build due to
d050420b6e8ebbe144e6548a616e65a0d873c27 which fixed `usedonly`).


ericvn (approved):
Initially confused changing the second variable from err to ok, but LGTM.

Issue #36796, opened 2022-01-11
Metadata exchange double injected causing disruption


bianpengyuan:
### Bug Description

When user upgrade control plane in place from 1.11 to 1.12, since 1.12 control plane injects tcp metadata exchange filter regardless of the version, 1.11 proxy will have TCP Mx filter double injected. When a double injected 1.11 talks to 1.12 proxy (which only have one tcp mx injected), the metadata will leak to the application, because metadata will be sent twice by 1.11 proxy, but only one chunk of metadata will be drained by the 1.12 peer.

### Version

```prose
1.12
```

bianpengyuan:
cc @rolandkool

hzxuzhonghu:
How does that happen?

bianpengyuan:
If the cluster is upgraded from 1.11 to 1.12, 1.11 proxy will be injected twice: one from EnvoyFilter and one from istiod.

hzxuzhonghu:
Thanks, understood

hzxuzhonghu:
Curious, didn't we test upgrade when 1.12 released

PR #36797, opened 2022-01-11
Only inject metadata exchange filters when proxy version is GE 1.12.

bianpengyuan:
fix https://github.com/istio/istio/issues/36796


howardjohn:
While istioctl doesn't prune the old EnvoyFilters, I worry some users (helm, etc) might be, which would break them. i asked @Stono about this

howardjohn:
I think my concerns about ^ are the case after talking with Karl.

I feel like there isn't much we can do to be fully robust without having awareness of if the EnvoyFilters are present at runtime in Istiod code. I don't know of any obvious ways to do that though... we could have special logic for the "well known" EnvoyFilter names but its a bit hacky - maybe the best we can do though.

WDYT?

bianpengyuan:
> I think my concerns about ^ are the case after talking with Karl.
> 
> I feel like there isn't much we can do to be fully robust without having awareness of if the EnvoyFilters are present at runtime in Istiod code. I don't know of any obvious ways to do that though... we could have special logic for the "well known" EnvoyFilter names but its a bit hacky - maybe the best we can do though.
> 
> WDYT?

Yeah I agree that is the best we can do. Will update this PR.

pilot/pkg/networking/util/util.go
```diff
@@ -100,6 +100,10 @@ const (
 
 	// Well-known header names
 	AltSvcHeader = "alt-svc"
+
+	// Well-known metadata exchange EnvoyFilter config name
+	MXName110 = "metadata-exchange-1.10"
+	MXName111 = "metadata-exchange-1.11"
```
hzxuzhonghu, line 106: Add a comment to remove them in 1.14

bianpengyuan: Sure, 1.14 should be good enough. We only support skip upgrade from n-2.

pilot/pkg/networking/util/util.go
```diff
@@ -100,6 +100,11 @@ const (
 
 	// Well-known header names
 	AltSvcHeader = "alt-svc"
+
+	// Well-known metadata exchange EnvoyFilter config name
+	// TODO: Remove these at 1.14.
+	MXName110 = "metadata-exchange-1.10"
```
howardjohn, line 106: what about tcp-mx?

pilot/pkg/networking/util/util.go
```diff
@@ -708,3 +713,19 @@ func TraceOperation(host string, port int) string {
...
 	return DomainName(host, port) + "/*"
 }
+
+// CheckProxyVerionForMX checks whether metadata exchange filters should be injected
+// based on proxy version and presence of the well known metadata exchange EnvoyFilter.
+// If those EnvoyFilter presents (which is the case if user in place upgrade with istioctl )
+// TODO: Remove this at 1.14 release, since we support skip version upgrade
+//       now and 1.11 proxy can talk to 1.13 control plane.
+func CheckProxyVerionForMX(push *model.PushContext, proxyVersion *model.IstioVersion) bool {
+	hasDefaultEnvoyFilter := push.HasEnvoyFilters(MXName111, push.Mesh.RootNamespace) || push.HasEnvoyFilters(MXName110, push.Mesh.RootNamespace)
```
howardjohn, line 723: I realize this is a huge edge case to have mx10 but not mx11 or vise versa, but wouldn't ideal behavior be:

```
if ver >= 1.12 { return true }
if ver >= 1.11 { return !has(mx11) }
if ver >= 1.10 { return !has(mx10) }
...
```

(where true = use native code)

howardjohn: Also want to note there could be `metadata-exchange-1.10-<revision>`, hypothetically. This should be impossible unless someone in place upgraded a revision or something which is not supported. I don't think we need to handle the case, just mentioning it.

howardjohn (approved):
LGTM, just had one minor comment - not sure we need to change anything though.

I know we are missing test coverage here, did you get a chance to test the two upgrade cases though (prune/not prune)? Logic looks good but fairly complex so just want to verify.

will approve with a hold, feel free to remove

pilot/pkg/networking/util/util.go
bianpengyuan: Sure, I've updated the code as you suggested, and yeah I just ignored revision assuming user will always use revision.

pilot/pkg/networking/util/util.go
bianpengyuan: I just assumed that presence of http mx == presence of tcp one. I cannot imagine a case where user intentionally delete http mx but not tcp mx.

bianpengyuan:
> did you get a chance to test the two upgrade cases though (prune/not prune)?

I've verified test build would avoid double injection when EnvoyFilter presents, and injects when no EnvoyFilters is there.

Issue #36694, opened 2022-01-04
Get all the VIPs of a service entry through MCP over xds


zhaohuabing:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**
Get all the VIPs of a service entry through MCP over xds.

Currently, MCP over xds assumes the request comes from the data plane and only add the service VIP of the same cluster to the response of the request.  [the current implementation](https://github.com/istio/istio/blob/fef0938eb0570d81cfc0a7c6a877c7f0f7315603/pilot/pkg/model/service.go#L744)

[Aeraki](https://github.com/aeraki-framework/aeraki) is using MCP over xds to fetch the service registry from Istio. Aeraki runs on the control plane and needs all the VIPs. 

**Describe alternatives you've considered**
Deploying an Aeraki instance in each cluster could be an alternative solution. However, I prefer using the same deploy mode as Istiod since they're both in the control plane.

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[*] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[*] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

costinm:
Sounds very reasonable. Main concern is the protocol - the quick solution would be to use a node metadata to indicate the node is a 'controller', and we may start with that.

However there is a larger issue - we may need to have some better filtering. One option is to treat the resource name as a URL, i.e. use query parameters. It is backward compatible to add this, and may allow per-resource filtering. For example we may want only resources from a namespace, or from a namespace + all imported/visible resources, or from specific clusters. 
MCP over XDS is supposed to be aligned with K8S API semantically, so it makes sense to provide at least the same granularity that K8S provides, plus the cross cluster option.

zhaohuabing:
Related Aeraki issue: https://github.com/aeraki-framework/aeraki/issues/149

zhaohuabing:
> Sounds very reasonable. Main concern is the protocol - the quick solution would be to use a node metadata to indicate the node is a 'controller', and we may start with that.
> 

[...]

PR #36820, opened 2022-01-13
Get all the VIPs of a service entry through MCP over xDS

zhaohuabing:
Related issue is here: https://github.com/istio/istio/issues/36694

Signed-off-by: zhaohuabing <huabingzhao@tencent.com>

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/service.go
+++ b/pilot/pkg/model/service.go
@@ -750,6 +750,17 @@ func ParseSubsetKey(s string) (direction TrafficDirection, subsetName string, ho
 	return
 }
 
+// GetAddresses returns a Service's addresses.
+// This method returns all the VIPs of a service if the ClusterID is explicitly set to "", otherwise only return the VIP
+// specific to the cluster where the node resides
+func (s *Service) GetAddresses(node *Proxy) []string {
+	if node.Metadata != nil && node.Metadata.ClusterID == "" {
+		return s.getAllAddresses()
+	}
+
+	return []string{s.GetAddressForProxy(node)}
+}
+
 // GetAddressForProxy returns a Service's address specific to the cluster where the node resides
 func (s *Service) GetAddressForProxy(node *Proxy) string {
 	if node.Metadata != nil {
@@ -767,6 +778,20 @@ func (s *Service) GetAddressForProxy(node *Proxy) string {
 	}
 
 	return s.DefaultAddress
+}
+
+// getAllAddresses returns a Service's all addresses.
+func (s *Service) getAllAddresses() []string {
+	var addresses []string
+	addressMap := s.ClusterVIPs.GetAddresses()
+	for _, clusterAddresses := range addressMap {
+		addresses = append(addresses, clusterAddresses...)
+	}
+
+	if len(addresses) == 0 && s.DefaultAddress != "" {
+		addresses = append(addresses, s.DefaultAddress)
+	}
+	return addresses
 }
 
 // GetTLSModeFromEndpointLabels returns the value of the label
--- a/pilot/pkg/model/service.go
+++ b/pilot/pkg/model/service.go
@@ -788,9 +788,6 @@ func (s *Service) getAllAddresses() []string {
 		addresses = append(addresses, clusterAddresses...)
 	}
 
-	if len(addresses) == 0 && s.DefaultAddress != "" {
-		addresses = append(addresses, s.DefaultAddress)
-	}
 	return addresses
 }
 
--- a/pilot/pkg/serviceregistry/serviceentry/conversion.go
+++ b/pilot/pkg/serviceregistry/serviceentry/conversion.go
@@ -63,7 +63,7 @@ func ServiceToServiceEntry(svc *model.Service, proxy *model.Proxy) *config.Confi
 		// ServiceEntry can represent multiple - but we are not using that. SE may be merged.
 		// Will be 0.0.0.0 if not specified as ClusterIP or ClusterIP==None. In such case resolution is Passthrough.
 		//
-		Addresses: []string{svc.GetAddressForProxy(proxy)},
+		Addresses: svc.GetAddresses(proxy),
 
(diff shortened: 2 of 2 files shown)


zhaohuabing:
I guess we could, but it would be confusing, because its role is not a proxy.

[Huabing Zhao - Chat @ Spike](https://spikenow.com/r/a/?ref=spike-organic-signature&_ts=1dmtlo)	[1dmtlo]

On January 13, 2022 at 13:50 GMT, Rama Chavali ***@***.***> wrote:

But this use case is another node type, which is a
"controller" running on the control plane.

Why cant we distinguish with proxy metadata?

pilot/pkg/model/context.go
```diff
@@ -744,9 +744,11 @@ const (
 
 	// Router type is used for standalone proxies acting as L7/L4 routers
 	Router NodeType = "router"
+
+	ControllerNode NodeType = "controller"
```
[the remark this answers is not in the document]
zhaohuabing: Are there any concerns about introducing a new node type?

zhaohuabing:
@costinm @hzxuzhonghu Could you please share your thoughts about this? Should we just use Proxy Metadata or introduce a "controller" node type?

zhaohuabing:
If we choose to use MetaData, we have a few options: Labels, Annotations, Raw. Neither Labels nor Annotations fit the semantic of the use case. May I just define a key "isController" in the Raw map?

ramaraochavali:
> May I just define a key "isController" in the Raw map?

My preference would be this. as introducing a new node type and not all code paths honouring that type (we mostly check if it is router or sidecar in majority of the code) seems incorrect to me. But I will wait for others feedback

zhaohuabing:
> > May I just define a key "isController" in the Raw map?
> 
> My preference would be this. as introducing a new node type and not all code paths honouring that type (we mostly check if it is router or sidecar in majority of the code) seems incorrect to me. But I will wait for others feedback

@costinm @hzxuzhonghu  How do you think about it?

hzxuzhonghu:
Option 2: there is a generator field in `NodeMetadata`, we can make use of it.

zhaohuabing:
> Option 2: there is a generator field in `NodeMetadata`, we can make use of it.

I need to pull the MCP resources, so the generator field has already been set as "API".

hzxuzhonghu:
`API` seems enough

zhaohuabing:
> `API` seems enough

The Aeraki use case is that Aeraki needs to pull MCP resources(Service Entry, VIrtual Service, etc.) from the Istio, just the same as the "api" generator currently does, but from a "controller" point of view, so it needs all the VIPs in a Service Entry, not just the address of the same cluster of the "proxy".

ramaraochavali:
Thank you. LGTM. Will leave it for others to review

costinm:
I think we should look more at the feature we want - not at the particular role.

You seem to want a custom filtering of the resources that are returned. It's not specific to controllers.

We can start with a Result_filter=none metadata. Later we can refine it, or even add it the the resource name.

zhaohuabing:
This approach is more generic and can be applied to more use cases, I'll change the code.

hzxuzhonghu:
MCP client does not want to filter out some resources, instead, it wants to know all resources.  There is a lack of use cases to implement more logic in pilot to support generic filtering.

hzxuzhonghu:
Put a hold until come to an agreement

zhaohuabing:
Should we put all the addresses in the result by default? Will the behavior cause incompatible change?

zhaohuabing:
@hzxuzhonghu @ramaraochavali @costinm Hi guys, is anything I can do to move this forward?

stevenctl:
> Should we put all the addresses in the result by default? Will the behavior cause incompatible change?

I like the idea, but depends on how consumers are using this.. If they're just building things that match these VIPs, they'll have a few irrelevant ones, but there could easily be cases where traffic starts getting sent to VIPs that are invalid for a given cluster. 

Do consumers properly set `CLUSTER_ID` meta? They should... and maybe if that's unspecified `""` (_not_ the default `"Kubernetes"' value) we don't filter VIPs based on CLUSTER_ID?

zhaohuabing:
> Do consumers properly set `CLUSTER_ID` meta? They should... and maybe if that's unspecified `""` (_not_ the default `"Kubernetes"' value) we don't filter VIPs based on CLUSTER_ID?

I think this approach is the best so far. It works for my use case and also has the least impacts on existing users.

@hzxuzhonghu @ramaraochavali @costinm if you also agree with this, I can change this PR this way.

stevenctl:
Another idea is `cluster_id=*` – less likely to break other clients. Only issue is making that wildcard have consistent behavior in all the places that we use cluster_id..

pilot/pkg/model/service.go
```diff
@@ -769,6 +780,20 @@ func (s *Service) GetAddressForProxy(node *Proxy) string {
...
 
+// getAllAddresses returns a Service's all addresses.
+func (s *Service) getAllAddresses() []string {
+	var addresses []string
+	addressMap := s.ClusterVIPs.GetAddresses()
+	for _, clusterAddresses := range addressMap {
+		addresses = append(addresses, clusterAddresses...)
+	}
+
+	if len(addresses) == 0 && s.DefaultAddress != "" {
```
hzxuzhonghu, line 791: DefaultAddress  can be "0.0.0.0"? 
In this case, there maybe autoallocated address, do we need to return it?

zhaohuabing: > DefaultAddress can be "0.0.0.0"? In this case, there maybe autoallocated address, do we need to return it?

We shouldn't return “0.0.0.0" or the auto-allocated address.

hzxuzhonghu (approved):

PR #36832, opened 2022-01-13
Fix build base images script

Monkeyanator:
since we moved to the go-based builder we don't support the `HUBS` envvar, the script takes a single hub at a time. should fix failures in https://github.com/istio/istio/pull/36830 after triggering rebuild


ericvn:
Looks good, but want to confirm that @howardjohn isn't updating the go builder to also look at the `HUBS` env var.

ericvn:
I believe this PR is no longer needed. Please rebase and reopen this PR if we need to review.

PR #36834, opened 2022-01-14
Set connection_balance_config to exact_balance...

### LINKED ISSUE #18152: Istio control plane supports the envoy function (tcp listener connection balancing)
The envoy community has released about tcp listener connection balancing(https://github.com/envoyproxy/envoy/pull/8422), which need pilot-discovery to support its configure
``` yaml
connection_balance_config:
  exact_balance: {}
```

This feature will improve the istio performance, and we verify that the fluctuation is reduced under the fortio scenario test.

hope istio  can add this feature to the istio 1.4 version. thanks 

**Additional context**

jacob-delgado:
for inbound and outbound connections. This does not apply to
virtualOutbound and for listeners that use QUIC.

Fixes #18152

Based on lessons from https://aspenmesh.io/sailing-faster-with-istio
we should set connectionBalanceConfig to exactBalance.


jacob-delgado:
We could either do this as an environment var to opt-in, but given the performance numbers we see I'd suggest we make this the default except in the case of mixed tcp/udp (quic) gateway entries We could also create an API change (not sure what CRD would be best for this, maybe sidecar?), but that usually takes a long time to get through and implemented and we are seeing good results using this across the board. I'll include Yuchen in this discussion.

Istio's half of https://github.com/envoyproxy/envoy/issues/15146

howardjohn (commented):
Any reason we shouldn't do this 100% of the time? Edge cases it performs worse in, etc? Or put another way - why didn't envoy make it default?

We should probably have a feature flag just in case - but on may be a reasonable default

jacob-delgado:
> Any reason we shouldn't do this 100% of the time? Edge cases it performs worse in, etc? Or put another way - why didn't envoy make it default?
> 
> We should probably have a feature flag just in case - but on may be a reasonable default

Maybe @lambdai would have a better idea why this wasn't the default. With the introduction of QUIC support we do need to be a bit cautious as the envoy docs say it only works with TCP listeners (and we don't want to set it for virtualOutbound per docs).

> The listener’s connection balancer configuration, currently only applicable to TCP listeners. If no configuration is specified, Envoy will not attempt to balance active connections between worker threads.
>
> In the scenario that the listener X redirects all the connections to the listeners Y1 and Y2 by setting use_original_dst in X and bind_to_port to false in Y1 and Y2, it is recommended to disable the balance config in listener X to avoid the cost of balancing, and enable the balance config in Y1 and Y2 to balance the connections among the workers.

ramaraochavali:
May be we should have a feature flag/ or config at proxy level  - https://github.com/envoyproxy/envoy/blob/1260c5c2d1c5b85639a8e566a6b153198ddbf109/docs/root/intro/arch_overview/intro/threading_model.rst#listener-connection-balancing - It depends on workload and Envoy claims for most workloads kernel does a better Job. If this works for all workloads, why can not we make this default in Envoy so that we do not have send this additional config every listener?

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -387,14 +387,25 @@ func (lb *ListenerBuilder) buildVirtualInboundListener(configgen *ConfigGenerato
 
 	actualWildcard, _ := getActualWildcardAndLocalHost(lb.node)
 	// add an extra listener that binds to the port that is the recipient of the iptables redirect
-	filterChains, passthroughInspector := buildInboundCatchAllFilterChains(configgen, lb.node, lb.push)
+	filterChains, passthroughInspector, usesQUIC := buildInboundCatchAllFilterChains(configgen, lb.node, lb.push)
+
+	// exact balance used in Envoy is only supported over TCP connections
+	var connectionBalance *listener.Listener_ConnectionBalanceConfig
+	if !usesQUIC && lb.node.GetInterceptionMode() == model.InterceptionRedirect {
```
hzxuzhonghu, line 394: why the mode should be redirect

jacob-delgado: changed that restriction

jacob-delgado:
> May be we should have a feature flag/ or config at proxy level - [envoyproxy/envoy@`1260c5c`/docs/root/intro/arch_overview/intro/threading_model.rst#listener-connection-balancing](https://github.com/envoyproxy/envoy/blob/1260c5c2d1c5b85639a8e566a6b153198ddbf109/docs/root/intro/arch_overview/intro/threading_model.rst#listener-connection-balancing) - It depends on workload and Envoy claims for most workloads kernel does a better Job. If this works for all workloads, why can not we make this default in Envoy so that we do not have send this additional config every listener?

Added a feature flag for both inbound and outbound connections. It's doubtful Envoy would change this behavior, as the behavior on Windows seems to definitely favor kernel, while on Linux our testing showed it's throughput would be better. Having different defaults for different platforms would be difficult to get settled.

Given that this can possibly also be workload-dependent I think a better long-term solution would be to add this to the new ProxyConfig CRD. But that is still a WIP for the moment. I can bring that up with Environments at the next meeting, but I'll hold off onto this for now.

I'd like to get this cherry-picked into `release-1.13` so we can start accumulating some feedback.

ramaraochavali:
Another approach could be to drive this by PROXY METADATA in short term while we add it ProxyConfig. Setting some thing at Pilot for this does not seem right to me

jacob-delgado:
> Another approach could be to drive this by PROXY METADATA in short term while we add it ProxyConfig. Setting some thing at Pilot level for this does not seem right to me

That actually sounds very reasonable and would lead to better adoption. I'll push that up in this PR and then request a re-review.

jacob-delgado:
Cc @ramaraochavali @hzxuzhonghu anything I can do to move this forward?

pilot/pkg/model/context.go
```diff
@@ -609,6 +609,14 @@ type NodeMetadata struct {
 	// ExitOnZeroActiveConnections terminates Envoy if there are no active connections if set.
 	ExitOnZeroActiveConnections StringBool `json:"EXIT_ON_ZERO_ACTIVE_CONNECTIONS,omitempty"`
 
+	// InboundListenerExactBalance sets connection balance config to use exact_balance for virtualInbound,
+	// as long as QUIC, since it uses UDP, isn't also used.
+	InboundListenerExactBalance StringBool `json:"INBOUND_LISTENER_EXACT_BALANCE,omitempty"`
+
+	// InboundListenerExactBalance sets connection balance config to use exact_balance for redirected
```
ramaraochavali, line 616: nit: fix.comment

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -1537,17 +1537,31 @@ func buildListener(opts buildListenerOpts, trafficDirection core.TrafficDirectio
...
+		var connectionBalance *listener.Listener_ConnectionBalanceConfig
 		if !opts.bindToPort {
 			bindToPort = proto.BoolFalse
 		}
+
+		if opts.proxy.Metadata.OutboundListenerExactBalance && trafficDirection == core.TrafficDirection_OUTBOUND {
+			// virtualOutbound listener should not use exact_balance as it is using
+			// original dst and the associated redirected listeners are not bound
+			// to a port, per envoy docs
+			connectionBalance = &listener.Listener_ConnectionBalanceConfig{
```
ramaraochavali, line 1549: Do not understand comment at this place - are you saying since this is not VirtualOutbound you are adding the exact balance? It is little confusing to me.

ramaraochavali (commented):
LGTM - left couple of minor comments
@hzxuzhonghu @howardjohn PTAL

pilot/pkg/model/context.go
jacob-delgado: Fixed

ramaraochavali (approved):

ericvn (approved):
Approval of release notes.

PR #36836, opened 2022-01-14
replace explicit usages of cluster.local with this constant

huanghuangzym:
in the todo list
we shoud replace cluster.local with  constant DefaultKubernetesDomain to avoid if it changed


Arhell (commented):
/ok-to-test

ramaraochavali (approved):

PR #36847, opened 2022-01-16
Correct route precedence logic for k8s ingress

### LINKED ISSUE #35033: The behavior of istio ingress is different from the ingress documentation.

According to the [ingress documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/#multiple-matches), it seems that routing is applied with longest matching path first. But the istio ingress(`kubernetes.io/ingress.class: istio`) seems to be affected by the order described in `paths` . For interoperability, I expect it to work as written in the Ingress documentation.

"In some cases, multiple paths within an Ingress will match a request. In those cases precedence will be given first to the longest matching path. If two paths are still equally matched, precedence will be given to paths with an exact path type over prefix path type."

https://github.com/goharbor/harbor-helm/issues/485#issuecomment-686222551

high priority to the longest matching path

[X] Docs
[ ] Installation
[X] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[X] User Experience
[ ] Developer Infrastructure

[X] Multi Cluster
[X] Virtual Machine
[X] Multi Control Plane

**Additional context**

dddddai:
Sort generated vs routes to meet ingress route precedence requirements, please refer to https://kubernetes.io/docs/concepts/services-networking/ingress/#multiple-matches

> In some cases, multiple paths within an Ingress will match a request. In those cases **precedence will be given first to the longest matching path. If two paths are still equally matched, precedence will be given to paths with an exact path type over prefix path type.**

Fixes #35033

### DIFF
--- a/pilot/pkg/config/kube/ingress/conversion.go
+++ b/pilot/pkg/config/kube/ingress/conversion.go
@@ -29,6 +29,7 @@ import (
 
 	meshconfig "istio.io/api/mesh/v1alpha1"
 	networking "istio.io/api/networking/v1alpha3"
+	"istio.io/istio/pilot/pkg/features"
 	"istio.io/istio/pilot/pkg/serviceregistry/kube"
 	"istio.io/istio/pkg/config"
 	"istio.io/istio/pkg/config/constants"
@@ -150,13 +151,11 @@ func ConvertIngressVirtualService(ingress v1beta1.Ingress, domainSuffix string, 
 			host = "*"
 		}
 		virtualService := &networking.VirtualService{
-			Hosts:    []string{},
+			Hosts:    []string{host},
 			Gateways: []string{fmt.Sprintf("%s/%s-%s-%s", ingressNamespace, ingress.Name, constants.IstioIngressGatewayName, ingress.Namespace)},
 		}
 
-		virtualService.Hosts = []string{host}
-
-		httpRoutes := make([]*networking.HTTPRoute, 0)
+		httpRoutes := make([]*networking.HTTPRoute, 0, len(rule.HTTP.Paths))
 		for _, httpPath := range rule.HTTP.Paths {
 			httpMatch := &networking.HTTPMatchRequest{}
 			if httpPath.PathType != nil {
@@ -216,19 +215,36 @@ func ConvertIngressVirtualService(ingress v1beta1.Ingress, domainSuffix string, 
 		if f {
 			vs := old.Spec.(*networking.VirtualService)
 			vs.Http = append(vs.Http, httpRoutes...)
-			sort.SliceStable(vs.Http, func(i, j int) bool {
-				r1 := vs.Http[i].Match[0].GetUri()
-				r2 := vs.Http[j].Match[0].GetUri()
-				_, r1Ex := r1.GetMatchType().(*networking.StringMatch_Exact)
-				_, r2Ex := r2.GetMatchType().(*networking.StringMatch_Exact)
-				// TODO: default at the end
-				if r1Ex && !r2Ex {
-					return true
-				}
-				return false
-			})
+			if features.LegacyIngressBehavior {
+				sort.SliceStable(vs.Http, func(i, j int) bool {
+					r1 := vs.Http[i].Match[0].GetUri()
+					r2 := vs.Http[j].Match[0].GetUri()
+					_, r1Ex := r1.GetMatchType().(*networking.StringMatch_Exact)
+					_, r2Ex := r2.GetMatchType().(*networking.StringMatch_Exact)
+					// TODO: default at the end
+					if r1Ex && !r2Ex {
+						return true
+					}
+					return false
+				})
+			}
 		} else {
 			ingressByHost[host] = &virtualServiceConfig
+		}
+
+		if !features.LegacyIngressBehavior {
+			// sort routes to meet ingress route precedence requirements
+			// see https://kubernetes.io/docs/concepts/services-networking/ingress/#multiple-matches
+			vs := ingressByHost[host].Spec.(*networking.VirtualService)
+			sort.SliceStable(vs.Http, func(i, j int) bool {
+				r1Len, r1Ex := getMatchURILength(vs.Http[i].Match[0])
+				r2Len, r2Ex := getMatchURILength(vs.Http[j].Match[0])
+				// TODO: default at the end
+				if r1Len == r2Len {
+					return r1Ex && !r2Ex
+				}
+				return r1Len > r2Len
+			})
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


howardjohn (commented):
Nice. Can we are a feature flag to retain the old behavior for backwards compatibility?

Also what happens for 2 ingress with same hostname?

dddddai:
Thank you!
> Can we are a feature flag to retain the old behavior for backwards compatibility?

OK, so should the default value be false?

> Also what happens for 2 ingress with same hostname?

They will be merged to one vs whose routes are sorted, as the unit test shows:

ingress 1:
https://github.com/istio/istio/blob/af5e25e8615342de44f6816265574553f521848e/pilot/pkg/config/kube/ingressv1/conversion_test.go#L142-L149

ingress 2:
https://github.com/istio/istio/blob/af5e25e8615342de44f6816265574553f521848e/pilot/pkg/config/kube/ingressv1/conversion_test.go#L215-L222

result:
https://github.com/istio/istio/blob/af5e25e8615342de44f6816265574553f521848e/pilot/pkg/config/kube/ingressv1/conversion_test.go#L281-L283

pilot/pkg/config/kube/ingress/conversion.go
```diff
@@ -216,20 +216,22 @@ func ConvertIngressVirtualService(ingress v1beta1.Ingress, domainSuffix string,
 		if f {
 			vs := old.Spec.(*networking.VirtualService)
 			vs.Http = append(vs.Http, httpRoutes...)
-			sort.SliceStable(vs.Http, func(i, j int) bool {
```
dddddai, line 219: And I'm a little confused about the old behavior, it seems to wanna sort the routes, but only when merging, so if there's only one ingress the generated vs routes will not be sorted, is that right?

hzxuzhonghu:
nginx ingress-controller: 
https://github.com/nginxinc/kubernetes-ingress/tree/v2.1.0/examples/mergeable-ingress-types  

Also have CRD: https://docs.nginx.com/nginx-ingress-controller/configuration/virtualserver-and-virtualserverroute-resources/

Not all ingress controllers support ingress merge

dddddai:
Thanks for the info
> Not all ingress controllers support ingress merge

Yes but we do merge ingresses
https://github.com/istio/istio/blob/994621ac3efbebbbadf1ed661a32d93e854d7508/pilot/pkg/config/kube/ingress/conversion.go#L134-L136

pilot/pkg/config/kube/ingress/conversion.go
```diff
@@ -216,20 +216,22 @@ func ConvertIngressVirtualService(ingress v1beta1.Ingress, domainSuffix string,
...
-				// TODO: default at the end
-				if r1Ex && !r2Ex {
-					return true
-				}
-				return false
-			})
 		} else {
 			ingressByHost[host] = &virtualServiceConfig
 		}
+
```
howardjohn, line 222: for backwards compat we will need a flag for old behavior. Can combine it with https://github.com/istio/istio/pull/36855#pullrequestreview-855690740 maybe

pilot/pkg/config/kube/ingress/conversion.go
howardjohn: I think so. I think the idea was to ensure if we have 2 ingress with something like:
```
match0: /bar
match1: /
```

and

```
match0: /foo
match1: /
```

we put the `/` matches at the bottom so /foo and /bar can match

pilot/pkg/config/kube/ingress/conversion.go
```diff
@@ -216,20 +216,22 @@ func ConvertIngressVirtualService(ingress v1beta1.Ingress, domainSuffix string,
...
+		// sort routes to meet ingress route precedence requirements
+		// see https://kubernetes.io/docs/concepts/services-networking/ingress/#multiple-matches
+		vs := ingressByHost[host].Spec.(*networking.VirtualService)
+		sort.SliceStable(vs.Http, func(i, j int) bool {
+			r1Len, r1Ex := getMatchURILength(vs.Http[i].Match[0])
+			r2Len, r2Ex := getMatchURILength(vs.Http[j].Match[0])
+			// TODO: default at the end
```
howardjohn, line 229: shouldn't default be at the end due to len=0?

pilot/pkg/config/kube/ingress/conversion.go
```diff
@@ -216,20 +216,22 @@ func ConvertIngressVirtualService(ingress v1beta1.Ingress, domainSuffix string,
...
+			if r1Len == r2Len {
```
howardjohn, line 230: > If two paths are still equally matched, precedence will be given to paths with an exact path type over prefix path type.

This is technically still ambiguous on which one wins... I guess here we are doing it based on creation time which seems reasonable

pilot/pkg/config/kube/ingressv1/conversion.go
```diff
@@ -240,6 +242,22 @@ func ConvertIngressVirtualService(ingress knetworking.Ingress, domainSuffix stri
...
 }
 
+// getMatchURILength returns the length of matching path, and whether the match type is EXACT
+func getMatchURILength(match *networking.HTTPMatchRequest) (length int, exact bool) {
+	uri := match.GetUri()
+	switch uri.GetMatchType().(type) {
+	case *networking.StringMatch_Exact:
+		return strings.Count(uri.GetExact(), "/"), true
+	case *networking.StringMatch_Prefix:
+		return strings.Count(uri.GetPrefix(), "/"), false
```
howardjohn, line 252: Is it the number of `/` or the raw `len(uri.GetPrefix())` we want? Can we look at some others like nginx and countor and see what they do?

pilot/pkg/config/kube/ingress/conversion.go
dddddai: We already sorted the ingresses based on creation time:
https://github.com/istio/istio/blob/994621ac3efbebbbadf1ed661a32d93e854d7508/pilot/pkg/config/kube/ingress/controller.go#L348

pilot/pkg/config/kube/ingress/conversion.go
dddddai: The comment is derived from [here](https://github.com/istio/istio/pull/36847/files#diff-b01367d947f649ce053caa5a9ffda6141cf242cfda73cfd62687fe185603e689L224)

I guess "default" means `pathType: ImplementationSpecific`?

pilot/pkg/config/kube/ingress/conversion.go
dddddai: Done

pilot/pkg/config/kube/ingressv1/conversion.go
dddddai: Oh yes, I've checked [ingress-nginx doc](https://kubernetes.github.io/ingress-nginx/user-guide/ingress-path-matching/#example_1), should use the raw length.
Fixed

howardjohn (approved):

PR #36855, opened 2022-01-17
[WIP] Update k8s ingress spec

Mangaal:


Arhell (commented):
/ok-to-test

pilot/pkg/config/kube/ingress/conversion.go
```diff
@@ -80,26 +80,39 @@ func ConvertIngressV1alpha3(ingress v1beta1.Ingress, mesh *meshconfig.MeshConfig
...
+			gateway.Servers = append(gateway.Servers, &networking.Server{
+				Port: &networking.Port{
+					Number:   443,
+					Protocol: string(protocol.HTTPS),
+					Name:     fmt.Sprintf("https-443-ingress-%s-%s-%d", ingress.Name, ingress.Namespace, i),
+				},
+				Hosts: tls.Hosts,
+				Tls: &networking.ServerTLSSettings{
+					HttpsRedirect:  false,
+					Mode:           networking.ServerTLSSettings_AUTO_PASSTHROUGH,
```
howardjohn, line 92:
```suggestion
					Mode:           networking.ServerTLSSettings_PASSTHROUGH,
```

Auto passthrough is very different, its for multinetwork

howardjohn (commented):
Nice. We will need some tests but this looks like a great start.

Since this is a behavioral change it will also be nice to add a flag in pilot/pkg/features to conditionally enable it. There is another Ingress change like this open (to sort routes), IMO we can use one flag and release both changes in one release. Then release notes can be something like 'Improved Ingress conformance (tls + route ordering). To use old behavior set OLD_INGRESS_BEHAVIOR=true" or similar

pilot/pkg/config/kube/ingress/conversion.go
```diff
@@ -80,26 +80,39 @@ func ConvertIngressV1alpha3(ingress v1beta1.Ingress, mesh *meshconfig.MeshConfig
 
 	for i, tls := range ingress.Spec.TLS {
 		if tls.SecretName == "" {
-			log.Infof("invalid ingress rule %s:%s for hosts %q, no secretName defined", ingress.Namespace, ingress.Name, tls.Hosts)
-			continue
+			gateway.Servers = append(gateway.Servers, &networking.Server{
+				Port: &networking.Port{
+					Number:   443,
+					Protocol: string(protocol.HTTPS),
+					Name:     fmt.Sprintf("https-443-ingress-%s-%s-%d", ingress.Name, ingress.Namespace, i),
```
hzxuzhonghu, line 87: some of this behaviors should be documented

Issue #36865, opened 2022-01-18
Support to disable traffic for WorkloadEntry


shonecyx:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**
WorkloadEntry can be used to represent external mesh VIP and select by ServiceEntry. There are some user cases that we need to disable traffic to a WorkloadEntry:
1. On purpose region/az exit, for example we need to disable the traffic to a specific VIP(WorkloadEntry) in one region/az.
2. AZ failover/Disaster recover. Let's say if all the backend servers behind a external VIP go down, we need to disable the traffic to that VIP(WorkloadEntry).

The capability to disable VIP traffic is very important for the traffic management, while currently Workload spec doesn't support based on current spec and there are some options for that:
1. Disable WorkloadEntry traffic if its weight is 0. As per upstream, the endpoint minimum weight should be 1 https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto,  so Istio will configure weight 1 for the Endpoint if the WorkloadEntry weight is 0, we can leverage on this 0 weight and set the corresponding lbEndpoint HealthStatus to be `HealthStatus_UNHEALTHY` to achieve  this. 
```
// GetLoadBalancingWeight returns the weight for this endpoint, normalized to always be > 0.
func (ep *IstioEndpoint) GetLoadBalancingWeight() uint32 {
	if ep.LbWeight > 0 {
		return ep.LbWeight
	}
	return 1
}

```
2. Introduce a new filed `adminState` in WorkloadEntry, it will also set the corresponding lbEndpoint HealthStatus to be `HealthStatus_UNHEALTHY`
```
apiVersion: networking.istio.io/v1alpha3
kind: WorkloadEntry
metadata:
  name: details-svc
spec:
  serviceAccount: details-legacy
  address: 2.2.2.2
  adminState: Down
  labels:
    app: details-legacy
```

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[ x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

hzxuzhonghu:
[...]

PR #36866, opened 2022-01-18
[WIP]Support to disable traffic for WorkloadEntry if weight is 0

shonecyx:

fix https://github.com/istio/istio/issues/36865 for option 1

- [ ] Docs
- [ ] Installation
- [ x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -25,6 +25,7 @@ import (
 
 	cluster "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	endpoint "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
 	auth "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
 	http "github.com/envoyproxy/go-control-plane/envoy/extensions/upstreams/http/v3"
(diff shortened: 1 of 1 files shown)


Arhell (commented):
/ok-to-test

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -633,6 +635,13 @@ func (cb *ClusterBuilder) buildLocalityLbEndpoints(proxyNetworkView map[network.
 	return localityLbEndpoints
 }
 
+func getEndpointHealthStatus(instance *model.ServiceInstance) v3.HealthStatus {
+	if instance.Endpoint.LbWeight == 0 {
```
howardjohn, line 639: how can you tell if weight is `0` because user explicitly set it to 0 or because they did not set it at all?

I think this will mean all users must explicitly set `weight: 1` to work.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -595,6 +596,7 @@ func (cb *ClusterBuilder) buildLocalityLbEndpoints(proxyNetworkView map[network.
 			LoadBalancingWeight: &wrappers.UInt32Value{
 				Value: instance.Endpoint.GetLoadBalancingWeight(),
 			},
+			HealthStatus: getEndpointHealthStatus(instance),
```
howardjohn, line 599: I would prefer we never set this field. We can filter out the endpoints instead.

howardjohn (changes requested):

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
shonecyx: We can have a flag for this feature based on this flag we'll know if WLE 0 weight means no traffic or default weight 1.
But I also think introduce the adminState in WLE is a better solution, it aligns with the Envoy LBEndpoint health status.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
shonecyx: Perhaps we can set it based on the WLE `adminState` so we can on purpose disable the traffic to that endpoint.

shonecyx:
Got it. To summarize, Istio does support to mark WLE unhealthy and the contract is WLE annotation

proxy.istio.io/health-checks-enabled: "true"
And manipulate WLE healthy status to be False:

status:
  conditions:
  - lastProbeTime: "2019-12-26T22:06:34Z"
    lastTransitionTime: "2019-12-26T22:06:34Z"
    message: 'Code 500: VIP Unhealthy'
    reason: unhealthy
    status: "False"
    type: Healthy
It's decoupled with readinessProbe and we can have external controller to mark the status unhealthy to disable traffic or do readiness check.

PR #36869, opened 2022-01-18
JWKS: No longer print entire response body on error

### LINKED ISSUE #35663: "Failed to fetch jwt public key from" prints whole response body to logs
### Bug Description

From the logs:
```
2021-10-19T07:51:48.035517Z	error	Failed to fetch jwt public key from "https://redacted.example.com/.well-known/openid-configuration/jwks": status 404, <!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" [continues to print the whole page...]

```

https://github.com/istio/istio/blob/5e7a398bcb3e3a67e1a8324ff5e5791e64684340/pilot/pkg/model/jwks_resolver.go#L354

Expected behaviour: Istio prints a succinct error message
Actual behaviour: Istio prints the whole body of the page to logs

This fills up the logs with noise. I would suggest to only print the first line of the body and limit the number of characters to print.

https://github.com/istio/istio/blob/5e7a398bcb3e3a67e1a8324ff5e5791e64684340/pilot/pkg/model/jwks_resolver.go#L348

Also reading the whole response body at all seems like a risk. If the response body is very large it could OOM istiod. Is there a safe upper bound on how large a valid public key can be?

### Version

```prose
master branch
```

### Additional Information

_No response_

nielsole:
* Fixes #35663 
* Limits log output to 100 bytes of the first line of response body to reduce log verbosity.
* Refined take on https://github.com/istio/istio/pull/35723

### DIFF
--- a/pilot/pkg/model/jwks_resolver.go
+++ b/pilot/pkg/model/jwks_resolver.go
@@ -346,15 +346,9 @@ func (r *JwksResolver) getRemoteContentWithRetry(uri string, retry int) ([]byte,
 		}
 		defer resp.Body.Close()
 
-		// RFC7517 does not specify a maximum length, but unbounded input may lead to resource exhaustion.
-		const maxBytesToRead int64 = 1048577 // 1MiB + 1Byte
-		limitedBody := io.LimitReader(resp.Body, maxBytesToRead)
-		body, err := io.ReadAll(limitedBody)
+		body, err := io.ReadAll(resp.Body)
 		if err != nil {
 			return nil, err
-		}
-		if int64(len(body)) == maxBytesToRead {
-			return nil, fmt.Errorf("status %d, response exceeded maximum length of %d", resp.StatusCode, maxBytesToRead-1)
 		}
 
 		if resp.StatusCode < 200 || resp.StatusCode >= 300 {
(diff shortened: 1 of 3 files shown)


pilot/pkg/model/jwks_resolver.go
```diff
@@ -351,7 +352,12 @@ func (r *JwksResolver) getRemoteContentWithRetry(uri string, retry int) ([]byte,
 		}
 
 		if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-			return nil, fmt.Errorf("status %d, %s", resp.StatusCode, string(body))
+			message := strconv.Quote(string(body))
```
howardjohn, line 355: should we also use https://pkg.go.dev/io#LimitedReader on L349? Since we still read the entire response into memory right now....

nielsole: The closed PR https://github.com/istio/istio/pull/35723 still includes those changes.
I did a couple of tests and I could not get my istiod to crash, because of the 5 second timeout. I simply didn't manage to download enough json within 5 seconds to OOM istiod. 
Since there wasn't a consensus on how high we should limit it, I removed it.

howardjohn (approved):

PR #36878, opened 2022-01-18
Changes to make pilot-agent generate correct bootstrap to connect to GCP

vishalpowar:
TrafficDirector

The changes in the PR fall in two categories
1) bootstrap template (gcp_envoy_bootstrap.json)
  - Change to support the new xDS V3 nodeID format used by
  TrafficDirector
  - Change to explictly support V3 API when connecting to
  TrafficDirector
  - Changes to support istio-proxy inbound interception.

2) supporting changes in pilot-agent to set bootstrap template
varaibles.

cla: yes

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pkg/bootstrap/config.go
```diff
@@ -79,15 +79,28 @@ func (cfg Config) toTemplateParams() (map[string]interface{}, error) {
...
 		option.OutlierLogPath(cfg.Metadata.OutlierLogPath),
 		option.ProvCert(cfg.Metadata.ProvCert),
 		option.DiscoveryHost(discHost),
+		option.GCPMeshScope(meshScope),
 		option.XdsType(xdsType))
 
+	// Add GCPProjectNumber to access in bootstrap template.
+	md := cfg.Metadata.PlatformMetadata
+	if projectNumber, found := md[platform.GCPProjectNumber]; found {
+		opts = append(opts, option.GCPProjectNumber(projectNumber))
```
howardjohn, line 101: maybe should set it in the same place project name is? https://github.com/istio/istio/blob/0a5a486d1c69ea6c8bf96a22dc1b8daf254cf646/pkg/bootstrap/config.go#L99

pkg/bootstrap/config.go
```diff
@@ -79,15 +79,28 @@ func (cfg Config) toTemplateParams() (map[string]interface{}, error) {
 		xdsType = "DELTA_GRPC"
 	}
 
+	// Add the meshScope
+	meshScope := "default"
+	if scope, found := cfg.Metadata.Labels["gke.io/mesh"]; found {
```
howardjohn, line 84: would prefer to just pass `cfg.Metadata` then in the template do this. To keep the gcp stuff in the gcp bootstrap file.

pilot/cmd/pilot-agent/options/agent.go
```diff
@@ -44,6 +44,9 @@ func NewAgentOptions(proxy *model.Proxy, cfg *meshconfig.ProxyConfig) *istioagen
 		ServiceNode:                 proxy.ServiceNode(),
 		EnvoyStatusPort:             envoyStatusPortEnv,
 		EnvoyPrometheusPort:         envoyPrometheusPortEnv,
+		TdInterceptionPort:          tdInterceptionPortEnv.Get(),
```
howardjohn, line 47: I don't see where this is actually used..?

howardjohn: I don't think we need to define it in code then. Anything with ISTIO_META_ is passed

pkg/bootstrap/config.go
vishalpowar: I wanted this to be unconditional, but I can put it there if you think that is the right place.

pkg/bootstrap/config.go
vishalpowar: sent Metadata.Labels instead as i was not able to make Metadata work. This change does move the gcp stuff in the gcp bootstrap file. LMK WYT.

pkg/bootstrap/config.go
howardjohn: Do you set STSPort in your case? I would think so but not certain.

It does seem weird to have ID in one place and Name in another

Arhell (commented):
/ok-to-test

pilot/cmd/pilot-agent/options/options.go
```diff
@@ -106,6 +106,12 @@ var (
 		"Envoy health status port value").Get()
 	envoyPrometheusPortEnv = env.RegisterIntVar("ENVOY_PROMETHEUS_PORT", 15090,
 		"Envoy prometheus redirection port value").Get()
+	tdInterceptionPortEnv = env.RegisterStringVar("TRAFFICDIRECTOR_INTERCEPTION_PORT","15001",
+		"TrafficDirector Interception port value")
+	tdInboundInterceptionPortEnv = env.RegisterStringVar("TRAFFICDIRECTOR_INBOUND_INTERCEPTION_PORT", "15006",
+		"TrafficDirector Inbound Interception port value")
+	tdInboundBackendPortsEnv = env.RegisterStringVar("TRAFFICDIRECTOR_INBOUND_BACKEND_PORTS", "8000",
+		"TrafficDirector Inbound Backend ports value")
```
hzxuzhonghu, line 114: I think it is better not to expose these special options to normal users

pkg/bootstrap/config.go
vishalpowar: I tried moving it below, and it failed for me on default configuration and deployment.

pkg/bootstrap/config.go
```diff
@@ -86,8 +86,15 @@ func (cfg Config) toTemplateParams() (map[string]interface{}, error) {
 		option.OutlierLogPath(cfg.Metadata.OutlierLogPath),
 		option.ProvCert(cfg.Metadata.ProvCert),
 		option.DiscoveryHost(discHost),
+		option.Metadata(cfg.Metadata),
```
bianpengyuan, line 89: Probably should scope this down to cfg.Metadata.Labels?

pkg/bootstrap/config.go
bianpengyuan: Curious, how does it fail exactly?

pkg/bootstrap/config.go
howardjohn: This is just making it available to the template, should not have any performance impact or anything. The main risk I see would be someone may customize bootstrap and rely on fields we later take away - but we already do not have a strong API here. Are there other concerns?

bianpengyuan: Mainly to be consistent with existing code, no functional concern. All other existing metadata fields are passed in as specific options, but here we are passing in the whole metadata for just labels. But anyway this is just a subjective suggestion, I guess up to @vishalpowar whether to update it or not ;).

pkg/bootstrap/config.go
vishalpowar: The nodeId does not find the ProjectName as the logic to populate does not get executed.

pkg/bootstrap/config.go
vishalpowar: My initial draft had changes with just Metadata.Labels being passed :) 
But i agree with @howardjohn that passing Metadata in its entirety is quite reasonable.

pkg/bootstrap/config.go
bianpengyuan: Ok so sts port is indeed not always used. Thanks!

bianpengyuan (approved):

pilot/cmd/pilot-agent/options/agent.go
vishalpowar: Removed this,

pilot/cmd/pilot-agent/options/options.go
vishalpowar: Ack, removed this.

pilot/pkg/model/context.go
```diff
@@ -606,6 +606,11 @@ type NodeMetadata struct {
 	// Envoy prometheus port redirecting to admin port prometheus endpoint.
 	EnvoyPrometheusPort int `json:"ENVOY_PROMETHEUS_PORT,omitempty"`
 
+	// TD ports
+	TdInterceptionPort string `json:"TRAFFICDIRECTOR_INTERCEPTION_PORT,omitempty"`
+	TdInboundInterceptionPort string `json:"TRAFFICDIRECTOR_INBOUND_INTERCEPTION_PORT,omitempty"`
+	TdInboundBackendPorts string `json:"TRAFFICDIRECTOR_INBOUND_BACKEND_PORTS,omitempty"`
+
```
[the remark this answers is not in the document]
vishalpowar: Removed

howardjohn (approved):

Issue #36465, opened 2021-12-10
OPENSSL_internal:CERTIFICATE_VERIFY_FAILED


Noksa:
### Bug Description

Hello.

I have three primary clusters that were installed using this manual: https://istio.io/latest/docs/setup/install/multicluster/multi-primary/

Everything works as expected but I have a problem with one service (it is couchdb).

For some reason when it restarts other workloads can't connect to it because of the error.

Let's start from the very beginning.

The problem occurs in all clusters but let's do it in the first cluster.

SVC:
```
➜ ~ k get svc -n kazoo-db --context first | grep "storage-db-svc"
storage-db-svc                  ClusterIP   10.100.26.163    <none>        5984/TCP,5986/TCP   3d16h
```

Endpoints from another workload:
```
➜ ~ istioctl pc endpoint --context first crossbar-844cd68fd9-g8q9c | grep -Ei "5984.*storage-db-svc"
10.1.9.200:5984                  HEALTHY     OK                outbound|5984||storage-db-svc.kazoo-db.svc.cluster.local
10.2.1.109:5984                  HEALTHY     OK                outbound|5984||storage-db-svc.kazoo-db.svc.cluster.local
192.168.58.110:5984              HEALTHY     FAILED            outbound|5984||storage-db-svc.kazoo-db.svc.cluster.local
```

The error in the envoy log of crossbar pod:

```
[2021-12-10T12:36:02.063Z] "GET / HTTP/1.1" 503 UF,URX upstream_reset_before_response_started{connection_failure,TLS_error:_268435581:SSL_routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED} - "TLS error: 268435581:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED" 0 195 45 - "-" "hackney/1.6.4" "9b3d6495-fa43-4054-8d0c-c6f51308b288" "storage-db-svc.kazoo-db:5984" "192.168.58.110:5984" outbound|5984||storage-db-svc.kazoo-db.svc.cluster.local - 10.100.26.163:5984 192.168.46.122:43165 - default
```

So it looks like we have bad certificate, right? But why?  Other workloads work fine between clusters...

Moreover, the `192.168.58.110:5984` address is in the same cluster where we have `crossbar-844cd68fd9-g8q9c`. So it is even not between clusters, they are located in the same cluster.

And I have the following DR:
```
│ spec:                                                                                                                                                                                                                                                                        
[...]

Issue #31534, opened 2021-03-17
Passthrough traffic with auto mtls doesn't handle workloads scaling causing TestReachability/automtls-passthrough failures


bianpengyuan:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/31531/integ-ipv6-k8s-tests_istio/1372292143661453312

cc @howardjohn

howardjohn:
This is a real bug

howardjohn:
The problem is EDS changes do not trigger this code: https://github.com/istio/istio/blob/79d56525633f13084241c6da2e3f948098242ce9/pilot/pkg/model/push_context.go#L1813.

So we start with 0 instances, and TLS is disabled. Then we scale to 1 instance, but don't update CDS, so we have no TLS configured.
If we made 0 instances enable mTLS, then it would still be broken, since we could scale to 1 instance without a sidecar.

I don't know a great way to resolve this. Any thoughts @hzxuzhonghu @incfly 

Note: this is not a regression, its a new test that catches an existing bug that has been around for a while.

hzxuzhonghu:
Similar to https://github.com/istio/istio/issues/29405

incfly:
Yeah. Seems we'll have to trigger a CDS push/updates in this case right? If no pods instances, we don't even know the service account associated with the server pods. Would not be able to set the SAN verification fields correctly.

howardjohn:
Good point about the SANs.. not sure how it works for. non passthrough but doesn't here. need to look into it more

hzxuzhonghu:
I am concerned about the istiod load though. After all app scale-up down is so common.

howardjohn:
Yeah we definitely cannot push cds on all endpoint changes. But maybe just
when the cds will actually change. we already have similar logic for the
SANs changing I think? It just doesn't work here for some reason

PR #36882, opened 2022-01-18
Recompute service account at endpoint shard when all endpoints are removed

bianpengyuan:
fix https://github.com/istio/istio/issues/36465 and https://github.com/istio/istio/issues/31534.

The sequence to trigger the issue:

1. Scale down a service's workloads to zero, either manually, or some upgrade or reschedule event. This will trigger an EDS push, which does not update Cluster / push CDS.
2. Update any istio config or any service, which will trigger a Cluster update and CDS push for all services (Note CDS cache is invalidated at step 1, which means Cluster for this service will be recomputed), and for this service, since there is no endpoint, the default validation context / SAN match will be empty `"default_validation_context": {}`.
3. Now scale up the workloads, this will trigger another EDS push, and we have the logic to check if the service accounts have any update (https://github.com/istio/istio/blob/4dea7d99eba0ab42e84056d50e1a96151f301887/pilot/pkg/xds/eds.go#L150). if there is, a CDS update will also be triggered. However, whether the service account change is checked against the EndpointShards cache, which is not cleaned up when all endpoints get removed at step 1, so no CDS update will happen.

### DIFF
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -127,11 +127,11 @@ func (s *DiscoveryServer) EDSCacheUpdate(shard model.ShardKey, serviceName strin
 func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string, namespace string, istioEndpoints []*model.IstioEndpoint) bool {
 	if len(istioEndpoints) == 0 {
 		// Should delete the service EndpointShards when endpoints become zero to prevent memory leak,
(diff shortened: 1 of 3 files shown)


howardjohn (commented):
Nice, LGTM but will let @ramaraochavali or @hzxuzhonghu take a look since I think they are familiar with this code deeply.

hzxuzhonghu (approved):

bianpengyuan:
@howardjohn ping for release notes approval. thanks!

howardjohn (approved):

PR #32928, opened 2021-05-18
agent: move magic concurrency equation to injector

kyessenov:

PR #36884, opened 2022-01-19
auto set concurrency for gateways from cpu limit

timonwong:

Currently, the gateway uses all hosts cpu, however, when CPU limit is set, and when during load the gateway gives us poor performance (high latency) due to CFS aggresive throttling.

This proposal uses cpu limit when concurrency is not set, which should mitigate this problem

### DIFF
--- a/manifests/charts/gateways/istio-egress/templates/deployment.yaml
+++ b/manifests/charts/gateways/istio-egress/templates/deployment.yaml
@@ -286,6 +286,11 @@ spec:
             - path: "annotations"
               fieldRef:
                 fieldPath: metadata.annotations
+            - path: "cpu-limit"
+              resourceFieldRef:
+                containerName: istio-proxy
+                resource: limits.cpu
+                divisor: 1m
       - name: istio-envoy
         emptyDir: {}
       - name: istio-data
--- a/manifests/charts/gateways/istio-ingress/templates/deployment.yaml
+++ b/manifests/charts/gateways/istio-ingress/templates/deployment.yaml
@@ -286,6 +286,11 @@ spec:
             - path: "annotations"
               fieldRef:
                 fieldPath: metadata.annotations
+            - path: "cpu-limit"
+              resourceFieldRef:
+                containerName: istio-proxy
+                resource: limits.cpu
+                divisor: 1m
       - name: istio-envoy
         emptyDir: {}
       - name: istio-data
--- a/pilot/cmd/pilot-agent/config/config.go
+++ b/pilot/cmd/pilot-agent/config/config.go
@@ -16,7 +16,6 @@ package config
 
 import (
 	"fmt"
-	"io/ioutil"
 	"math"
 	"os"
 	"strconv"
@@ -63,7 +62,7 @@ func ConstructProxyConfig(meshConfigFile, serviceCluster, proxyConfigEnv string,
 	}
 
 	if concurrency == 0 && role.Type == model.Router {
-		// If --concurrency is unset and not override in env & annotations, we will automatically set this
+		// If --concurrency is unset and not override in file, env, or annotations, we will automatically set this
 		// based on CPU limits.
 		if proxyConfig.Concurrency == nil {
 			byResources := determineConcurrencyOption()
@@ -164,7 +163,7 @@ func fileExists(path string) bool {
 }
 
 func readPodCPULimits() (int, error) {
-	b, err := ioutil.ReadFile(constants.PodInfoCPULimitsPath)
+	b, err := os.ReadFile(constants.PodInfoCPULimitsPath)
 	if err != nil {
 		return 0, err
 	}
--- a/pilot/cmd/pilot-agent/config/config.go
+++ b/pilot/cmd/pilot-agent/config/config.go
@@ -16,6 +16,8 @@ package config
 
 import (
 	"fmt"
+	"io/ioutil"
+	"math"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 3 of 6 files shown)


manifests/charts/gateways/istio-ingress/templates/deployment.yaml
```diff
@@ -278,6 +278,11 @@ spec:
             - path: "annotations"
               fieldRef:
                 fieldPath: metadata.annotations
+            - path: "cpu-limit"
+              resourceFieldRef:
+                containerName: istio-proxy
+                resource: limits.cpu
```
hzxuzhonghu, line 284: does downward api support resource now?

hzxuzhonghu (commented):
/ok-to-test

LGTM

manifests/charts/gateways/istio-ingress/templates/deployment.yaml
hzxuzhonghu: Found it [resourceFieldRef](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#capabilities-of-the-downward-api)

timonwong: @hzxuzhonghu Yeah, actually this take previous art (for sidecar proxy only) which was removed sometime ago: https://github.com/istio/istio/pull/32928

zirain:
this PR change the default behavior, @timonwong can you add a release notes?

manifests/charts/gateways/istio-ingress/templates/deployment.yaml
howardjohn: It wasn't fully *removed* it was *moved*. The recommended gateway install methods all use https://istio.io/latest/docs/setup/additional-setup/gateway/ which does injection which can do this in the same way.

I would prefer to just add the

```
          {{- if gt .EstimatedConcurrency 0 }}
            - --concurrency
            - "{{ .EstimatedConcurrency }}"
          {{- end -}}
```

bit to https://github.com/istio/istio/blob/c2d1f7aff60f2e7f3265839208ba0960c20f1091/manifests/charts/istio-control/istio-discovery/files/gateway-injection-template.yaml?

In current code the `--concurrency` logic in agent has caused SO many bugs, so if we are going to change it again I would have concerns there and need more thorough investigation (and testing)

timonwong: > I would prefer to just add the
> 
> ```
>           {{- if gt .EstimatedConcurrency 0 }}
>             - --concurrency
>             - "{{ .EstimatedConcurrency }}"
>           {{- end -}}
> ```

@howardjohn Yeah, but maybe requires some modification since there are some minor differences in concurrency logic between gateways and sidecar proxies.
I think it's easier for sidecar because by default the manifests in our repo still do not use injection templates, and deploy gateways like old days (egress and ingress).

timonwong:
> this PR change the default behavior, @timonwong can you add a release notes?

@zirain releasenote is now added.

pilot/cmd/pilot-agent/config/config.go
```diff
@@ -59,7 +61,16 @@ func ConstructProxyConfig(meshConfigFile, serviceCluster, proxyConfigEnv string,
 		proxyConfig = *meshConfig.DefaultConfig
 	}
 
-	if concurrency != 0 {
+	if concurrency == 0 && role.Type == model.Router {
+		// If --concurrency is unset and not override in file, env, or annotations, we will automatically set this
+		// based on CPU limits.
```
costinm, line 66: Why not for sidecars ?

timonwong: Sidecars and gateways are treated differently:
https://github.com/istio/istio/pull/32928/files

hzxuzhonghu (approved):

zirain:
ping @istio/wg-test-and-release-maintainers @istio/release-managers

howardjohn:
Adding @kyessenov since this is essentially reverting their PR

zirain:
not stale

zirain:
ping @kyessenov

manifests/charts/gateways/istio-ingress/templates/deployment.yaml
kyessenov: I think the long term plan is to "manage" gateways via injection and no longer support static YAML templates for gateways. I'd prefer the default to stay what it is today (no guessing at runtime).

kyessenov: Looks like this is not a default path, so it's fine.

GregHanson:
not stale

zirain:
ping @istio/wg-test-and-release-maintainers , seems @kyessenov approved

manifests/charts/gateways/istio-egress/templates/deployment.yaml
```diff
@@ -286,6 +286,11 @@ spec:
             - path: "annotations"
               fieldRef:
                 fieldPath: metadata.annotations
+            - path: "cpu-limit"
```
howardjohn, line 289: These are the legacy charts, what about manifests/charts/gateway?

pilot/cmd/pilot-agent/config/config.go
```diff
@@ -85,6 +96,20 @@ func ConstructProxyConfig(meshConfigFile, serviceCluster, proxyConfigEnv string,
...
 
+// determineConcurrencyOption determines the correct setting for --concurrency based on CPU limits
+func determineConcurrencyOption() *wrapperspb.Int32Value {
+	// If limit is set, us that
+	// The format in the file is a plain integer. `100` in the file is equal to `100m` (based on `divisor: 1m`
+	// in the pod spec).
+	// With the resource setting, we round up to single integer number; for example, if we have a 500m limit
+	// the pod will get concurrency=1. With 6500m, it will get concurrency=7.
+	limit, err := readPodCPULimits()
+	if err == nil && limit > 0 {
```
howardjohn, line 107: The old logic used requests and limit. Its not clear if or why only limits is sufficient now: https://github.com/istio/istio/pull/32928/files. Can you expand?

howardjohn (changes requested):

Issue #36804, opened 2022-01-12
grpcgen: TestMtls failed


hzxuzhonghu:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/36789/unit-tests_istio/1481079407186546688

```
=== RUN   TestMtls
2022-01-12T01:49:38.720187Z	info	kube	Controller terminated
2022-01-12T01:49:38.750235Z	info	initializing Kubernetes credential reader for cluster Kubernetes
2022-01-12T01:49:38.750626Z	info	Skipping IngressClass, resource not supported
2022-01-12T01:49:38.759102Z	info	model	reloading network gateways
2022-01-12T01:49:38.760556Z	info	controllers	starting	controller=ingress
2022-01-12T01:49:38.761138Z	info	ads	Full push, new service default/echo-app.default.svc.cluster.local
2022-01-12T01:49:38.761433Z	info	kube	kube controller for Kubernetes synced after 887.855µs
2022-01-12T01:49:38.761453Z	info	ads	Push debounce stable[1] 1 for config ServiceEntry/default/echo-app.default.svc.cluster.local: 50.456µs since last change, 49.803µs since last push, full=false
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5a00ea6]
goroutine 553 [running]:
istio.io/istio/pilot/pkg/xds.(*DiscoveryServer).AdsPushAll(0xc0014d71e0, {0xc000080960, 0x16}, 0xc000d980f0)
	/home/prow/go/src/istio.io/istio/pilot/pkg/xds/ads.go:840 +0x346
istio.io/istio/pilot/pkg/xds.(*DiscoveryServer).Push(0x1b91ac5, 0xc000d980f0)
	/home/prow/go/src/istio.io/istio/pilot/pkg/xds/discovery.go:337 +0x40c
istio.io/istio/pilot/pkg/xds.debounce.func1(0x5c1f650, 0x7fd35fc00008)
	/home/prow/go/src/istio.io/istio/pilot/pkg/xds/discovery.go:418 +0x54
created by istio.io/istio/pilot/pkg/xds.debounce.func2
	/home/prow/go/src/istio.io/istio/pilot/pkg/xds/discovery.go:440 +0x619
FAIL	istio.io/istio/pilot/pkg/networking/grpcgen	1.798s
```

PR #36887, opened 2022-01-19
Fix grpcgen TestMtls

hzxuzhonghu:

Fix https://github.com/istio/istio/issues/36804

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/fake.go
+++ b/pilot/pkg/networking/core/v1alpha3/fake.go
@@ -141,7 +141,7 @@ func NewConfigGenTest(t test.Failer, opts TestOptions) *ConfigGenTest {
 		serviceDiscovery.AddRegistry(reg)
 	}
 
-	env := &model.Environment{}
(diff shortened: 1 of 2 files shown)


pilot/pkg/networking/core/v1alpha3/fake.go
```diff
@@ -172,7 +172,6 @@ func NewConfigGenTest(t test.Failer, opts TestOptions) *ConfigGenTest {
 		if err := env.InitNetworksManager(&FakeXdsUpdater{}); err != nil {
 			t.Fatal(err)
 		}
-		env.PushContext = model.NewPushContext()
 		if err := env.PushContext.InitContext(env, nil, nil); err != nil {
```
hzxuzhonghu, line 175: SkipRun is very error prone, @howardjohn Maybe we should remove this flag

howardjohn (approved):

howardjohn:
Still a few flakes: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/37112/unit-tests_istio/1488922988769185792. Need to look into it

PR #36892, opened 2022-01-19
gateway chart: support envoy's bootstrap configuration override

ryodocx:
I implemented gateway's HelmChart mostly about this article:
https://medium.com/@nizam./envoy-layered-runtime-configuration-override-ac3499f82f6f

After this change, we can configure additonal bootstrap configuration of EnvoyProxy.
In order to override admin setting, runtime configuration, and so on.

### DIFF
--- a/manifests/charts/gateway/templates/configmap.yaml
+++ b/manifests/charts/gateway/templates/configmap.yaml
@@ -1,0 +1,14 @@
+{{- if .Values.customBootstrap }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "gateway.name" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "gateway.labels" . | nindent 4 }}
+data:
+  {{- with .Values.customBootstrap }}
+  custom_bootstrap.json: |-
+    {{- . | mustToPrettyJson | nindent 4 }}
+  {{- end }}
+{{- end }}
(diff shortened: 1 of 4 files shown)


zirain:
/ok-to-test

@ryodocx please sign CLA

howardjohn (changes requested):
Our intent has been to keep this helm chart fairly minimal - similar to the core of other popular charts and the output of `helm chart create`. I think this drifts a bit from that, and introduces substantial complexity to expose a super advanced use case, which is also not the long term implementation of this feature (we have an experimental BOOTSTRAP_XDS_AGENT mode which requests bootstrap from XDS, which can override this more appropriately).

I appreciate the effort put into this but I am not sure its the right path. I would recommend BOOTSTRAP_XDS_AGENT or using last mile customizations (https://tech.aabouzaid.com/2020/09/3-ways-to-customize-off-the-shelf-helm-charts-with-kustomize-kubernetes.html). Cheers

ryodocx:
@howardjohn Thank you!
`BOOTSTRAP_XDS_AGENT=true` was worked with EnvoyFilter (`applyTo: BOOTSTRAP`).
Can override Envoy's bootstrap configuration at will.
Close this PullRequest.

However, I found bug of HelmChart.
Charts can't accept boolean at env now (v1.12.2). 
I fixed it. If it appropriate, I make new PR.
https://github.com/ryodocx/istio/commit/0c0d702d15a71e4b1d4ce381a2cf3659672cc002

PR #36898, opened 2022-01-19
control operator replica sets history

tareksha:


howardjohn (commented):
Can you explain why?

Arhell (commented):
/ok-to-test

litong01 (approved):
Changes look good to me. The changes keep the behavior the same if not using this new configuration parameter.

jacob-delgado (approved):
You aren't changing defaults, but allowing users to specify. LGTM.

PR #36900, opened 2022-01-19
bump kiali to v1.45

jmazzitelli:

Upgrade the Kiali addon to v1.45.0

Note that this corrects the image version also (the previous change forgot to do that - you must remember to modify the values-kiali.yaml when doing this kind of update to kiali).

### DIFF
--- a/manifests/addons/gen.sh
+++ b/manifests/addons/gen.sh
@@ -31,6 +31,7 @@ TMP=$(mktemp -d)
 helm3 template kiali-server \
   --namespace istio-system \
   --version 1.45.0 \
+  --set deployment.image_version=v1.45 \
   --include-crds \
   --set nameOverride=kiali \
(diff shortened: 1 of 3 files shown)


bianpengyuan (approved):

howardjohn (approved):

jmazzitelli:
I made a slight change - the `gen.sh` script now defines the version to be used for both kiali settings (the helm chart `version` and the kiali configuration setting `deployment.image_version`. This will make it easier for people to remember how to do this change in the future. Now you just change `gen.sh` rather than having to remember to also change `values-kiali.yaml` (it seems like everytime we do this update, whoever does the update forgets to change the values file which in turn means you will be running the older kiali image with the newer kiali config - bad things will happen if you do that. This hopefully makes it easier to avoid this problem again in the future).

PR #36901, opened 2022-01-19
added a fix for https-jwt test

aryan16:
1) Updated the jwt-server deployment by adding the readiness probe.

### DIFF
--- a/samples/jwt-server/jwt-server.yaml
+++ b/samples/jwt-server/jwt-server.yaml
@@ -119,6 +119,13 @@ spec:
         ports:
         - containerPort: 8000
         - containerPort: 8443
+        readinessProbe:
+          initialDelaySeconds: 10
+          periodSeconds: 10
+          httpGet:
+            scheme: HTTPS
+            path: /
+            port: 8443
       volumes:
       - name: certkeysecretresource
         secret:


aryan16:
cc : @yangminzhu

samples/jwt-server/jwt-server.yaml
```diff
@@ -119,6 +119,13 @@ spec:
         ports:
         - containerPort: 8000
         - containerPort: 8443
+        readinessProbe:
+          initialDelaySeconds: 10
```
howardjohn, line 123: Do we need initial delay? This means we will never become ready until 10+rand(0,10) seconds (the random part is the probe interval)

aryan16: I added this because the same issue was appearing again, https jwt-server was taking time to fully get ready. Initial delay resolved that problem.

howardjohn: That seems wrong? We are basically adding time.Sleep(10s) then. If its marked as ready then why is it failing?

aryan16: I believe it just gets ready but not still in running state. Is there a way to check that without using the initial delay ?

yangminzhu: I guess this is to fix the occasional flaky in the test? Is a smaller delay and period (e.g. 3s) good enough for the fix?

yangminzhu: So here is my guess:
1. the jwt-server takes some unusual long time to be ready to serve the https traffic
2. istiod tried to fetch the jwks public key and failed all the retries, the jwt-server pod is READY but the https part is not, this caused the test to fail
3. after adding the readiness, the jwt-server gets more time to become ready for the https part so when the pod is READY, the https part is also ready

I'm not sure why 1) would happen in the first place though, @aryan16 please add if I missed anything

aryan16: I also believe the issue is with the https server endpoint is not getting ready though cluster is ready to serve as it has multiple endpoints.

Issue #36571, opened 2021-12-20
WasmPlugin: How to add insecure-registry to fetch the wasm binary


leyao-daily:
In my WasmPlugin, my oci url is point to an address which need to be added as a insecure-registry. And when I add it to host docker, I can pull the image, But when I apply this WasmPlugin CR, the istio-proxy will throw a error:
`x509: certificate signed by unknown authority`

Is there anything wrong with my configuration?

bianpengyuan:
Can you provide a bit more detail about your config here? Hope to take a look at what exactly is the oci uri you are using.

leyao-daily:
@bianpengyuan  url: oci://registry.fi.intel.com/leyao/envoy-modsecurity-wasm-plugin:0.1.0

leyao-daily:
And does WASMPLUGIN work for istio-system/istio-ingressgateway?

bianpengyuan:
We will need an option to agent to skip verify insecure, similar to https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/root.go#L68

Yes, see example at https://istio.io/latest/docs/reference/config/proxy_extensions/wasm-plugin/

poussa:
Another way is to teach the agent the CA who signed the registry certificate. That is, somehow let envoy know that it trust the signing CA. Using insecure option is just saying I don't care who signed it.

bianpengyuan:
@poussa I did not try, but would it just work if you mount the root cert to `/etc/ssl/certs` at Istio proxy?

leyao-daily:
Will this be developed as a new feature in the future to accepte insecure repo?

bianpengyuan:
Yeah I will add an env variable to the istio agent to allow pulling from insecure repo.

hzxuzhonghu:
It looks kubelet does not check the ca, should we allow insecure by default?

bianpengyuan:
> It looks kubelet does not check the ca, should we allow insecure by default?

@hzxuzhonghu  Are you sure? That seems unlikely to me... And is that is not clear, we are actually talking about pull with TLS but private CA, not pull with plain text.

hzxuzhonghu:
@bianpengyuan I am wrong, i looked at the code of kubelet, actually it calls the `CRI`. So should look at CRI implementation. And i go through docker/moby repo, looks MTLS by default.

bryanwux:
@bianpengyuan  Are you working on this right now?

[...]

PR #36903, opened 2022-01-20
add insecure oci support

zirain:

fixed https://github.com/istio/istio/issues/36571

### DIFF
--- a/pilot/cmd/pilot-agent/options/options.go
+++ b/pilot/cmd/pilot-agent/options/options.go
@@ -99,6 +99,9 @@ var (
 	enableProxyConfigXdsEnv = env.RegisterBoolVar("PROXY_CONFIG_XDS_AGENT", false,
 		"If set to true, agent retrieves dynamic proxy-config updates via xds channel").Get()
 
+	wasmInsecureRegistries = env.RegisterStringVar("WASM_INSECURE_REGISTRIES", "",
+		"allow agent pull wasm plugin from insecure registries, for example: 'localhost:5000,docker-registry:5000'").Get()
+
 	// Ability of istio-agent to retrieve bootstrap via XDS
 	enableBootstrapXdsEnv = env.RegisterBoolVar("BOOTSTRAP_XDS_AGENT", false,
 		"If set to true, agent retrieves the bootstrap configuration prior to starting Envoy").Get()
--- a/pkg/istio-agent/xds_proxy.go
+++ b/pkg/istio-agent/xds_proxy.go
@@ -142,6 +142,8 @@ func initXdsProxy(ia *Agent) (*XdsProxy, error) {
 			LocalHostAddr: localHostAddr,
 		}
 	}
+
+	cache := wasm.NewLocalFileCache(constants.IstioDataDir, wasm.DefaultWasmModulePurgeInterval, wasm.DefaultWasmModuleExpiry, ia.cfg.WASMInsecureRegistries)
 	proxy := &XdsProxy{
 		istiodAddress:         ia.proxyConfig.DiscoveryAddress,
 		istiodSAN:             ia.cfg.IstiodSAN,
@@ -151,7 +153,7 @@ func initXdsProxy(ia *Agent) (*XdsProxy, error) {
 		healthChecker:         health.NewWorkloadHealthChecker(ia.proxyConfig.ReadinessProbe, envoyProbe, ia.cfg.ProxyIPAddresses, ia.cfg.IsIPv6),
 		xdsHeaders:            ia.cfg.XDSHeaders,
 		xdsUdsPath:            ia.cfg.XdsUdsPath,
-		wasmCache:             wasm.NewLocalFileCache(constants.IstioDataDir, wasm.DefaultWasmModulePurgeInterval, wasm.DefaultWasmModuleExpiry),
+		wasmCache:             cache,
 		proxyAddresses:        ia.cfg.ProxyIPAddresses,
 		downstreamGrpcOptions: ia.cfg.DownstreamGrpcOptions,
 	}
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -28,6 +28,7 @@ import (
 	"sync"
 	"time"
 
+	"istio.io/istio/pilot/pkg/util/sets"
 	"istio.io/pkg/log"
 )
 
@@ -62,8 +63,9 @@ type LocalFileCache struct {
 	mux sync.Mutex
 
 	// Duration for stale Wasm module purging.
-	purgeInterval    time.Duration
-	wasmModuleExpiry time.Duration
+	purgeInterval      time.Duration
+	wasmModuleExpiry   time.Duration
+	insecureRegistries sets.Set
 
 	// stopChan currently is only used by test
 	stopChan chan struct{}
@@ -86,14 +88,15 @@ type cacheEntry struct {
 }
 
 // NewLocalFileCache create a new Wasm module cache which downloads and stores Wasm module files locally.
-func NewLocalFileCache(dir string, purgeInterval, moduleExpiry time.Duration) *LocalFileCache {
+func NewLocalFileCache(dir string, purgeInterval, moduleExpiry time.Duration, insecureRegistries []string) *LocalFileCache {
 	cache := &LocalFileCache{
-		httpFetcher:      NewHTTPFetcher(),
-		modules:          make(map[cacheKey]cacheEntry),
-		dir:              dir,
-		purgeInterval:    purgeInterval,
-		wasmModuleExpiry: moduleExpiry,
-		stopChan:         make(chan struct{}),
+		httpFetcher:        NewHTTPFetcher(),
+		modules:            make(map[cacheKey]cacheEntry),
+		dir:                dir,
+		purgeInterval:      purgeInterval,
+		wasmModuleExpiry:   moduleExpiry,
+		stopChan:           make(chan struct{}),
+		insecureRegistries: sets.NewSet(insecureRegistries...),
 	}
 	go func() {
 		cache.purge()
@@ -143,8 +146,17 @@ func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration
 	case "oci":
 		ctx, cancel := context.WithTimeout(context.Background(), timeout)
 		defer cancel()
+
+		insecure := false
+		if c.insecureRegistries.Contains(u.Host) {
+			insecure = true
+		}
 		// TODO: support imagePullSecret and pass it to ImageFetcherOption.
-		fetcher := NewImageFetcher(ctx, ImageFetcherOption{})
+		imgFetcherOps := ImageFetcherOption{
+			Insecure: insecure,
+		}
+		wasmLog.Debugf("wasm oci fetch options: %v", imgFetcherOps)
+		fetcher := NewImageFetcher(ctx, imgFetcherOps)
 		b, err = fetcher.Fetch(u.Host+u.Path, checksum)
 		if err != nil {
 			if errors.Is(err, errWasmOCIImageDigestMismatch) {
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -155,7 +155,7 @@ func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration
 		imgFetcherOps := ImageFetcherOption{
 			Insecure: insecure,
 		}
-		wasmLog.Debugf("wasm oci fetch options: %v", imgFetcherOps)
+		wasmLog.Debugf("wasm oci fetch %s with options: %v", downloadURL, imgFetcherOps)
 		fetcher := NewImageFetcher(ctx, imgFetcherOps)
 		b, err = fetcher.Fetch(u.Host+u.Path, checksum)
 		if err != nil {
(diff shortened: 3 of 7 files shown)


pkg/wasm/imagefetcher.go
```diff
@@ -41,6 +42,9 @@ type ImageFetcherOption struct {
 	Username string
 	Password string
 	// TODO(mathetake) Add signature verification stuff.
+
+	// TODO: how to set this? WasmPlugin, Pilot ENV or MeshConfig
```
daixiang0, line 46: Better to set this in WasmPlugin as it is only for Wasm.

zirain:
@bianpengyuan as we talk on Slack, ptal

pkg/wasm/imagefetcher.go
```diff
@@ -41,6 +42,9 @@ type ImageFetcherOption struct {
 	Username string
 	Password string
 	// TODO(mathetake) Add signature verification stuff.
+
+	// TODO: make this configurable by WasmPlugin
```
bianpengyuan, line 46: I am not sure we will ever want to make this first class. Like kubelet, it is just an runtime option for testing/dev purpose.

pilot/cmd/pilot-agent/options/options.go
```diff
@@ -98,6 +98,9 @@ var (
 	enableProxyConfigXdsEnv = env.RegisterBoolVar("PROXY_CONFIG_XDS_AGENT", false,
 		"If set to true, agent retrieves dynamic proxy-config updates via xds channel").Get()
 
+	enableInsecureWasmPluginEnv = env.RegisterBoolVar("INSECURE_WASM_PLUGIN", false,
```
bianpengyuan, line 101: I thought we are planning to do a per registry configuration? i.e. it should be a list of registries which are allowlisted for insecure pull.

zirain: IMHO, if we plan to do `per registry configuration`, `WasmPlugin` seems suitable for this scenario. there's why I put a TODO on `ImageFetcherOption `.

bianpengyuan: I don't think we should ever expose an insecure option in the first class API.

zirain: add `InsecureRegistries` in `MeshConfig`?

costinm: Meahconfig is not visible to the agent, and if we 'wire' it via injection will end up as an env variable anyways.

Unless you mean ProxyConfig.

Either way, please qualify it - wasmInsecure... - there are other registries used.

zirain: > Meahconfig is not visible to the agent, and if we 'wire' it via injection will end up as an env variable anyways.
> 
> Unless you mean ProxyConfig.
> 
> Either way, please qualify it - wasmInsecure... - there are other registries used.

we need some config like `insecure-registries` in `/etc/docker/daemon.json`, @costinm do you have any suggestion?

pkg/wasm/imagefetcher.go
daixiang0: Please think about the private network case, own private register often uses certs signed by an unknown authority(as it is private), which is used more than testing/dev, it usually is used in production, especially in the financial sector.

daixiang0: It is not like kubelet, as kubelet will use config from container runtime, but WasmPlugin does not.

bianpengyuan: For those cases, we should add image pull cert to the API, instead of exposing insure options.

pilot/cmd/pilot-agent/options/options.go
bianpengyuan: > add InsecureRegistries in MeshConfig?

Adding it to ProxyConfig seems like an overkill to me. If the concern is to configure some registries to be insecure mesh wide, you can already do that by setting ProxyMetadata at mesh wide ProxyConfig.

pkg/wasm/imagefetcher.go
daixiang0: Unfortunately, now wasm pull secret is also not supported :(

zirain:
@daixiang0 @bianpengyuan @costinm 
Let me summarize:
- for long-term, we should support `imagePullSecret` on `WasmPlugin`, IIRC, @bianpengyuan is working on
- for test/no-production env, users can simply turn `INSECURE_WASM_PLUGIN`  on.

pkg/wasm/imagefetcher.go
```diff
@@ -61,6 +65,15 @@ func NewImageFetcher(ctx context.Context, opt ImageFetcherOption) *ImageFetcher
 	} else {
 		fetchOpts = append(fetchOpts, remote.WithAuth(&authn.Basic{Username: opt.Username}))
 	}
+
+	if opt.Insecure {
+		t := remote.DefaultTransport.Clone()
+		t.TLSClientConfig = &tls.Config{
+			InsecureSkipVerify: opt.Insecure, //nolint: gosec
```
howardjohn, line 72: What about plaintext repos?

zirain: seems some error for now
```console
2022-01-26T02:48:54.028581Z	error	wasm	cannot fetch Wasm module oci://kind-registry:5000/zirain/autocomplete-authority:v1: could not fetch Wasm OCI image: could not fetch image: Get "https://kind-registry:5000/v2/": http: server gave HTTP response to HTTPS client

```

howardjohn: I didn't look at the code but I would note that Helm just added OCI support and plaintext works out of the box.

But it's possible it read some local config I have on my machine and I had configure the registry as plaintext... not sure

zirain: I think we should support plaintext repo, especially local develop with kind.

bianpengyuan: Yeah sounds good to allow pull from plain text by default.

pilot/cmd/pilot-agent/options/options.go
bianpengyuan: Let's make it a registry whitelist env for now?

zirain: > Let's make it a registry whitelist env variable for now?

pass array by `ProxyConfig` sounds a bit odd

daixiang0: > Let's make it a registry whitelist env variable for now?

How about using `WASM_INSECURE_REGISTRIES`, to disable the insecure feature if it is empty.

hzxuzhonghu: +1 ProxyConfig is a good place for it.

zirain:
updated, @bianpengyuan ptal
- replace `INSECURE_WASM_PLUGIN` with `WASM_INSECURE_REGISTRIES`
- support plain text repo

pkg/wasm/cache.go
```diff
@@ -28,6 +28,8 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/sets"
```
howardjohn, line 31: nit: use `pilot/pkg/util/sets`

pkg/wasm/imagefetcher.go
```diff
@@ -61,14 +67,30 @@ func NewImageFetcher(ctx context.Context, opt ImageFetcherOption) *ImageFetcher
...
 	}
 }
 
 // Fetch is the entrypoint for fetching Wasm binary from Wasm Image Specification compatible images.
 func (o *ImageFetcher) Fetch(url, expManifestDigest string) ([]byte, error) {
-	ref, err := name.ParseReference(url)
+	nameOpts := make([]name.Option, 0, 1)
+	// support plain text repo, see more detail [here](https://github.com/google/go-containerregistry/blob/master/pkg/name/registry.go#L75)
+	if o.insecure {
+		nameOpts = append(nameOpts, name.Insecure)
```
howardjohn, line 90: a bit confused - here we have plaintext and above we have `InsecureSkipVerify`. Do we support both TLS without verify _and_ plaintext? If so, how does the library decide which to use?

zirain: without this, `ref` still use `https` protocol default.

https://github.com/google/go-containerregistry/blob/master/pkg/name/registry.go#L75
```golang
// Scheme returns https scheme for all the endpoints except localhost or when explicitly defined.
func (r Registry) Scheme() string {
	if r.insecure {
		return "http"
	}
	if r.isRFC1918() {
		return "http"
	}
	if strings.HasPrefix(r.Name(), "localhost:") {
		return "http"
	}
	if reLocal.MatchString(r.Name()) {
		return "http"
	}
	if reLoopback.MatchString(r.Name()) {
		return "http"
	}
	if reipv6Loopback.MatchString(r.Name()) {
		return "http"
	}
	return "https"
}
```

howardjohn: So `if insecure use http`. But doesn't that mean if I have a registry at https://foo.bar that I want to use TLS with insecureSkipVerify, now it will use HTTP, not TLS?

zirain: updated, ptal

howardjohn (approved):
LGTM, do we have the ability to run unit tests against this?

zirain:
> LGTM, do we have the ability to run unit tests against this?

not sure about it, will take a look in the near future. 

@bianpengyuan need one more approval, PTAL

pkg/wasm/cache.go
```diff
@@ -143,8 +146,17 @@ func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration
...
+		insecure := false
+		if c.insecureRegistries.Contains(u.Host) {
+			insecure = true
+		}
 		// TODO: support imagePullSecret and pass it to ImageFetcherOption.
-		fetcher := NewImageFetcher(ctx, ImageFetcherOption{})
+		imgFetcherOps := ImageFetcherOption{
+			Insecure: insecure,
+		}
+		wasmLog.Debugf("wasm oci fetch options: %v", imgFetcherOps)
```
bianpengyuan, line 158: nit: add image url to the log line?

pkg/wasm/imagefetcher.go
```diff
@@ -122,6 +135,22 @@ func (o *ImageFetcher) Fetch(url, expManifestDigest string) ([]byte, error) {
 	)
 }
 
+func parseReference(url string) (name.Reference, error) {
+	ref, err := name.ParseReference(url)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse url in image reference: %v", err)
```
bianpengyuan, line 141: nit: could just return the error here, otherwise the error will be wrapped again with the same text after returning.

pkg/wasm/imagefetcher.go
```diff
@@ -122,6 +135,22 @@ func (o *ImageFetcher) Fetch(url, expManifestDigest string) ([]byte, error) {
...
+	}
+
+	// fallback to http based request, inspired by [helm](https://github.com/helm/helm/blob/12f1bc0acdeb675a8c50a78462ed3917fb7b2e37/pkg/registry/client.go#L594)
+	_, err = remote.Get(ref)
+	if err != nil && strings.Contains(err.Error(), "server gave HTTP response") {
+		wasmLog.Infof("fetch plain text from %s", url)
```
bianpengyuan, line 147:
```suggestion
		wasmLog.Infof("fetch with plain text from %s", url)
```

bianpengyuan (approved):

pkg/istio-agent/xds_proxy.go
```diff
@@ -142,6 +142,8 @@ func initXdsProxy(ia *Agent) (*XdsProxy, error) {
 			LocalHostAddr: localHostAddr,
 		}
 	}
+
+	cache := wasm.NewLocalFileCache(constants.IstioDataDir, wasm.DefaultWasmModulePurgeInterval, wasm.DefaultWasmModuleExpiry, ia.cfg.WASMInsecureRegistries)
```
hzxuzhonghu, line 146: nit:   wasm.DefaultWasmModulePurgeInterval, wasm.DefaultWasmModuleExpiry no need to pass

hzxuzhonghu (approved):

daixiang0:
Still need doc for this.

zirain:
> Still need doc for this.

any idea about the right place for a doc? maybe `istio.io`?

hzxuzhonghu:
I think the document can be with imagepull secret

daixiang0:
Failed to enable this feature:
```
istioctl install --set meshConfig.defaultConfig.proxyMetadata.WASM_INSECURE_REGISTRIES=my-registery
```

bianpengyuan:
Any interesting log at proxy? Turning on debug log would help debugging.

daixiang0:
Same log as below:
```
could not fetch Wasm OCI image: could not parse url in image reference: Get "https://registry.fi.intel.com/v2/": x509: certificate signed by unknown authority
```

zirain:
> Same log as below:
> 
> ```
> could not fetch Wasm OCI image: could not parse url in image reference: Get "https://registry.fi.intel.com/v2/": x509: certificate signed by unknown authority
> ```

let me take a look at this, seems something error with the plaintext support on 
```golang
// fallback to http based request, inspired by [helm](https://github.com/helm/helm/blob/12f1bc0acdeb675a8c50a78462ed3917fb7b2e37/pkg/registry/client.go#L594)
	_, err = remote.Get(ref)
	if err != nil && strings.Contains(err.Error(), "server gave HTTP response") {
		wasmLog.Infof("fetch with plain text from %s", url)
		return name.ParseReference(url, name.Insecure)
	}
```

zirain:
I'll take a look at this

bianpengyuan:
@zirain I think https://github.com/istio/istio/blob/50c0cc307ee273e34e35e324fc130371bd536bdf/pkg/wasm/imagefetcher.go#L151 should return nil here instead of err.

zirain:
> @zirain I think
> 
> https://github.com/istio/istio/blob/50c0cc307ee273e34e35e324fc130371bd536bdf/pkg/wasm/imagefetcher.go#L151
> 
> should return nil here instead of err.

seems I forgot to respect `fetchOpts` when doing `remote.Get`, I need to rebuild my test env to verify.

PR #36908, opened 2022-01-20
remove aggregator AppendWorkloadHandler implement to prevent misuse

hzxuzhonghu:


howardjohn:
I am a bit confused how this works. Is the issue we are registering things 2x or something?

stevenctl:
I don't see where we actually append handlers at the aggregate level today, so this seems safe to me. Are you seeing an actual issue right now, or is this code cleanup?

hzxuzhonghu:
@stevenctl You are right, there is no append now. This is to make sure no developer in the future will misuse it unless we find it is useful.

stevenctl (approved):

Issue #36907, opened 2022-01-20
Cluster Deletion does not clean the handler it registered to the aggregator controller


hzxuzhonghu:
This can happen for MCS, because serviceImport registered service handler for each registry.

PR #36910, opened 2022-01-20
Remove cluster handlers from aggregate controller when cluster deleted

hzxuzhonghu:

This is a bug fix, and I will add a test if this way works.

fix https://github.com/istio/istio/issues/36907

cc @stevenctl @nmittler

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
+++ b/pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
@@ -75,7 +75,7 @@ func newServiceImportCache(c *Controller) serviceImportCache {
 		}
 
 		// Register callbacks for Service events anywhere in the mesh.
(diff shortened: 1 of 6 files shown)


pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
```diff
@@ -75,7 +75,7 @@ func newServiceImportCache(c *Controller) serviceImportCache {
 		}
 
 		// Register callbacks for Service events anywhere in the mesh.
-		c.opts.MeshServiceController.AppendServiceHandler(sic.onServiceEvent)
+		c.opts.MeshServiceController.AppendServiceHandlerForCluster(c.Cluster(), sic.onServiceEvent)
```
hzxuzhonghu, line 78: This is the root cause

hzxuzhonghu:
ping @stevenctl

stevenctl (approved):

PR #36922, opened 2022-01-21
gateway: expose mechanism to use spiffe workload cert at gateway

howardjohn:
There are various times when users need to deploy TLS on a workload, but
doesn't care at all what certificate is used. One example could be
istio.io docs, but that is a minor case. The more important is things
like
https://aws.amazon.com/blogs/containers/secure-end-to-end-traffic-on-amazon-eks-using-tls-certificate-in-acm-alb-and-istio/.

Folks will run `client ---TLS---> Cloud LB (terminate) --TLS---> Istio
Ingress (terminate) ---> backend`.

The connection between cloud LB and Istio Ingress doesn't do any
verification, which leads to using self signed certs. While there is
technically nothing wrong with this, it adds extra steps and draws
scrutiny from security-conscious users.

As a step to make this easier, we can allow using the workloads mTLS
certs. This works in `MUTUAL` and `SIMPLE` cases.

Opening to get feedback, will add tests if there is agreement. @ramaraochavali @hzxuzhonghu

### DIFF
--- a/pilot/pkg/model/credentials/resource.go
+++ b/pilot/pkg/model/credentials/resource.go
@@ -30,6 +30,9 @@ const (
 	// take the form kubernetes-gateway://namespace/name. They are pulled from the config cluster.
 	KubernetesGatewaySecretType    = "kubernetes-gateway"
 	kubernetesGatewaySecretTypeURI = KubernetesGatewaySecretType + "://"
+	// BuiltinGatewaySecretType is the name of a SDS secret that uses the workloads own mTLS certificate
+	BuiltinGatewaySecretType    = "builtin"
+	BuiltinGatewaySecretTypeURI = BuiltinGatewaySecretType + "://"
 )
 
 // SecretResource defines a reference to a secret
@@ -52,11 +55,17 @@ func (sr SecretResource) Key() string {
 }
 
 func ToKubernetesGatewayResource(namespace, name string) string {
+	if strings.HasPrefix(name, BuiltinGatewaySecretTypeURI) {
+		return BuiltinGatewaySecretTypeURI
+	}
 	return fmt.Sprintf("%s://%s/%s", KubernetesGatewaySecretType, namespace, name)
 }
 
 // ToResourceName turns a `credentialName` into a resource name used for SDS
 func ToResourceName(name string) string {
+	if strings.HasPrefix(name, BuiltinGatewaySecretTypeURI) {
+		return "default"
+	}
 	// If they explicitly defined the type, keep it
 	if strings.HasPrefix(name, kubernetesSecretTypeURI) || strings.HasPrefix(name, kubernetesGatewaySecretTypeURI) {
 		return name
--- a/pilot/pkg/model/credentials/resource_test.go
+++ b/pilot/pkg/model/credentials/resource_test.go
@@ -105,3 +105,42 @@ func TestParseResourceName(t *testing.T) {
 		})
 	}
 }
+
+func TestToResourceName(t *testing.T) {
+	tests := []struct {
+		name string
+		want string
+	}{
+		{"foo", "kubernetes://foo"},
+		{"kubernetes://bar", "kubernetes://bar"},
+		{"kubernetes-gateway://bar", "kubernetes-gateway://bar"},
+		{"builtin://", "default"},
+		{"builtin://extra", "default"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := ToResourceName(tt.name); got != tt.want {
+				t.Fatalf("got %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestToKubernetesGatewayResource(t *testing.T) {
+	tests := []struct {
+		name      string
+		namespace string
+		want      string
+	}{
+		{"foo", "ns", "kubernetes-gateway://ns/foo"},
+		{"builtin://", "anything", "builtin://"},
+		{"builtin://extra", "anything", "builtin://"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := ToKubernetesGatewayResource(tt.namespace, tt.name); got != tt.want {
+				t.Fatalf("got %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
--- a/pilot/pkg/security/model/authentication.go
+++ b/pilot/pkg/security/model/authentication.go
@@ -80,6 +80,12 @@ func ConstructSdsSecretConfigForCredential(name string) *tls.SdsSecretConfig {
 	if name == "" {
 		return nil
 	}
+	if name == credentials.BuiltinGatewaySecretTypeURI {
+		return ConstructSdsSecretConfig(SDSDefaultResourceName)
+	}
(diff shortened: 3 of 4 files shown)


johnzheng1975 (approved):
If BuiltinGatewaySecretType = "builtin", will return SDSDefaultResourceName (fetching normal key/cert).

johnzheng1975:
@howardjohn , Thanks a lot for your contribute.

How to use this then?
For example If user set **tls: mode** to **BUILDIN** (add a new item for mode), istio will create a certificate for all domain "*". Then aws application load balancer will connect it with https?

```
$ cat <<EOF | kubectl apply -f -
kind: Gateway
metadata:
  name: mygateway
spec:
  selector:
    istio: ingressgateway # use istio default ingress gateway
  servers:
  - port:
      number: 443
      name: https
      protocol: HTTPS
    tls:
      mode: BUILDIN
      # credentialName: httpbin-credential # remove
    hosts:
    - httpbin.example.com
EOF
```

hzxuzhonghu:
>client ---TLS---> Cloud LB (terminate) --TLS---> Istio Ingress (terminate) ---> backend

The certificate used between Cloud LB -> Istio ingress should not be the business of istio. If we expose a way to use internal cert in gateway, how should we send key/cert to Cloud LB?

johnzheng1975:
I test with https://aws.amazon.com/blogs/containers/secure-end-to-end-traffic-on-amazon-eks-using-tls-certificate-in-acm-alb-and-istio/, it works!
Just as @howardjohn  said, the aws load balancer did not verify at all, I change change to other self-signed cert for httpbin.example.com or helloworld-v1.example.com. I did not realize this before. 
This makes my requirements so easy to implement.
Thanks for your reminder.

howardjohn:
> The certificate used between Cloud LB -> Istio ingress should not be the business of istio. If we expose a way to use internal cert in gateway, how should we send key/cert to Cloud LB?

We don't need to - no cloud LB I know of verifies cert at all, its used only for encrypting not authentication. So this just makes it easier to expose "some cert, I don't care what".

We already expose a way to show the spiffe cert, we just don't have a way to use it for simple TLS.

johnzheng1975:
My question to AWS support:
Will application load balancer check the cert when its target group protocal is HTTPS?

As below target group, the alb is created in eks.
arn:aws:elasticloadbalancing:us-west-2:846014624907:targetgroup/k8s-istiosys-istioing-a0b8758b05/a5ed2e099b8fdfea
The backend service is istio ingress gateway.
I found:
No matter which cert I put in istio gateway, the connection is always fine.
Its real domain is test.hpbp.io, but it still works when I change cert to other self-signed cert for httpbin.example.com or helloworld-v1.example.com.
So, I guess the load balancer(application load balancer) do not check the cert domain at all. Even fine when it is overdue?
I want to confirm this, thus I can design my solution on my project.

The answer from AWS support:
[https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-routing-configuration](https://protect-us.mimecast.com/s/PuovCn5XpXTm1BWNnTP9BFt?domain=docs.aws.amazon.com)
If a target group is configured with the HTTPS protocol or uses HTTPS health checks, the TLS connections to the targets use the security settings from the ELBSecurityPolicy-2016-08 policy. The load balancer establishes TLS connections with the targets using certificates that you install on the targets. The load balancer does not validate these certificates. Therefore, you can use self-signed certificates or certificates that have expired. Because the load balancer is in a virtual private cloud (VPC), traffic between the load balancer and the targets is authenticated at the packet level, so it is not at risk of man-in-the-middle attacks or spoofing even if the certificates on the targets are not valid.

johnzheng1975:
Hope the answer from AWS support is useful for your PR.  @howardjohn  @hzxuzhonghu

howardjohn:
Thanks for the info. Fwiw all the same applies to GCP

hzxuzhonghu:
Understand, so this is a way to auto provide certificates for gateway, that any client can connect to it  via tls/https. And it does not require client cert at all. 

Sounds good

pilot/pkg/model/credentials/resource.go
```diff
@@ -30,6 +30,9 @@ const (
 	// take the form kubernetes-gateway://namespace/name. They are pulled from the config cluster.
 	KubernetesGatewaySecretType    = "kubernetes-gateway"
 	kubernetesGatewaySecretTypeURI = KubernetesGatewaySecretType + "://"
+	// BuiltinGatewaySecretType is the name of a SDS secret that uses the workloads own mTLS certificate
+	BuiltinGatewaySecretType    = "builtin"
```
hzxuzhonghu, line 34: Just a little hacky here, not a good ux

liyihuang: > Understand, so this is a way to auto provide certificates for gateway, that any client can connect to it via tls/https. And it does not require client cert at all.
> 
> Sounds good

Hey guys, I just would like to throw my 2 cents here since I have some chat with @johnzheng1975. I don't think providing a self-signed certs gonna be a good solution at least for AWS, and it may cause other confusion.

the problem that I can see is the AWS ALB strips off the SNI fields after terminating the TLS. if we configure any hosts in gateway resource, the gw envoy listener will not be able to handle this TLS connection. We will end up with a single wildcard gw but use different VS and hosts fields to differentiate different domains. That may break the mode if someone wants multiple gateways.

This can be the issue only to AWS ALB since I can see Azure application gateway can somehow preserve the SNI.

howardjohn: I don't see what that has to do with the certificate though. If you cannot accept that limitation then you cannot use AWS ALB -- TLS --> GW. Which is fine - but this is intended for people that _do_ want that. This is documented in https://istio.io/latest/docs/ops/common-problems/network-issues/#configuring-sni-routing-when-not-sending-sni btw.

liyihuang: Sorry, I didn't know this limitation is documented there. I think that's fair

costinm:
For API: I would use Secret and a special naming convention. 
We can also use a new Group/Kind - but that would be Istio specific, a naming convention could be implemented by others, and in some cases users could just create a real secret with that name and a workload identity.

It is certainly not a TLSMode - just a different source for the cert.

For MTLS - we could use an option. I think adding a new TLSMode is not ideal - we are terminating, but use a handshake option to request/require client cert.

costinm (approved):
Looks good as code, and it would be good to merge as is - we can fine tune the API later, since it's an extension anyways.

costinm:
I would at least change from 'builtin' to 'hbone' or 'mtls'. 

The semantics of the extended mode would be 'mTLS with spiffee workload identity, using default platform credentials and certs'.

navjotsinghji:
> Understand, so this is a way to auto provide certificates for gateway, that any client can connect to it via tls/https. And it does not require client cert at all.
> 
> Sounds good

Hi, 

Pardon me if my question is naive but isn't ISTIO_MUTUAL mode already a way to provide auto certificates for gateway. The certs are generated by istio itself and hence are not exposed via K8s secrets. @hzxuzhonghu/ @howardjohn / @johnzheng1975

howardjohn:
Yes but that means it's requiring mTLS and also a few other istio-isms like
custom ALPN, etc.

PR #36928, opened 2022-01-21
Ignore replicaCount when autoscale is enabled

timonwong:

When autoscale is enabled we should not overwrite replica count, consider following scenario:

1. Set `values.pilot.autoscaleEnabled=true`, `components.pilot.k8s.replicaCount=1`
2. In istio operator it "caches" the generated manifests (with istiod.replicas=1)
3. Provides: HPA scales our pilot replicas to 3
4. Set `values.pilot.autoscaleEnabled=false`
5. The generated manifests (with istiod.replicas=1) is same as istio operator "cache", the deployment will not get updated unless istio operator is restarted.

### DIFF
--- a/operator/pkg/apis/istio/v1alpha1/validation/validation.go
+++ b/operator/pkg/apis/istio/v1alpha1/validation/validation.go
@@ -168,19 +168,28 @@ func checkDeprecatedSettings(iop *v1alpha1.IstioOperatorSpec) (util.Errors, []st
 	return errs, messages
 }
 
+type FeatureValidator func(*valuesv1alpha1.Values, *v1alpha1.IstioOperatorSpec) (util.Errors, []string)
+
 // validateFeatures check whether the config sematically make sense. For example, feature X and feature Y can't be enabled together.
-func validateFeatures(values *valuesv1alpha1.Values, spec *v1alpha1.IstioOperatorSpec) (util.Errors, []string) {
-	errs := CheckServicePorts(values, spec)
-	warningMessages := checkAutoScaleAndReplicaCount(values, spec)
-	return errs, warningMessages
+func validateFeatures(values *valuesv1alpha1.Values, spec *v1alpha1.IstioOperatorSpec) (errs util.Errors, warnings []string) {
+	validators := []FeatureValidator{
+		CheckServicePorts,
+		CheckAutoScaleAndReplicaCount,
+	}
+
+	for _, validator := range validators {
+		newErrs, newWarnings := validator(values, spec)
+		errs = util.AppendErrs(errs, newErrs)
+		warnings = append(warnings, newWarnings...)
+	}
+
+	return
 }
 
-// checkAutoScaleAndReplicaCount warns when autoscaleEnabled is true and k8s replicaCount is set.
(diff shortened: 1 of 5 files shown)


timonwong:
Should this label "release-notes-none"?

zirain:
@timonwong  I prefer to add a release notes

timonwong:
> @timonwong I prefer to add a release notes
The release note is now added

releasenotes/notes/36928.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: installation
+
+releaseNotes:
+- |
+  **Fixed** an issue that was preventing the operator to update the deployments when `.autoscaleEnabled` is `true` and `.k8s.replicaCount` is nonzero.
+  Besides, when both autoscale is enabled and repliaCount is nonzero, warning messages will be generated during validation.
```
ericvn, line 8:
```suggestion
  **Fixed** an issue that was preventing the operator from updating deployments when `.autoscaleEnabled` is `true` and `.k8s.replicaCount` is nonzero.
  When both autoscale is enabled and replicaCount is nonzero, warning messages will be generated during validation.
```

ericvn (approved):

operator/pkg/apis/istio/v1alpha1/validation/validation.go
```diff
@@ -160,12 +165,43 @@ func checkDeprecatedSettings(iop *v1alpha1.IstioOperatorSpec) (util.Errors, stri
...
-func validateFeatures(values *valuesv1alpha1.Values, spec *v1alpha1.IstioOperatorSpec) util.Errors {
-	return CheckServicePorts(values, spec)
+func validateFeatures(values *valuesv1alpha1.Values, spec *v1alpha1.IstioOperatorSpec) (util.Errors, []string) {
+	errs := CheckServicePorts(values, spec)
+	warningMessages := checkAutoScaleAndReplicaCount(values, spec)
+	return errs, warningMessages
+}
+
+// checkAutoScaleAndReplicaCount warns when autoscaleEnabled is true and k8s replicaCount is set.
+func checkAutoScaleAndReplicaCount(values *valuesv1alpha1.Values, spec *v1alpha1.IstioOperatorSpec) []string {
```
litong01, line 179: can this check find not only warning but also errors? For example, the replicaCount set but also autoscale enabled, should that be considered an error? I wonder if it is actually better to return both errors and warning by this method? which will bear a similar interface as method validateFeatures? Since you are changing the signature of validateFeatures, do you mind getting this new method same interface?

timonwong: > can this check find not only warning but also errors? For example, the replicaCount set but also autoscale enabled, should that be considered an error? I wonder if it is actually better to return both errors and warning by this method? which will bear a similar interface as method validateFeatures? Since you are changing the signature of validateFeatures, do you mind getting this new method same interface?

@litong01 I think warnings should be better, because in general the change in replica count doesn't hurt too much, and it only occurred when you enable and then disable the autoscalEanbled. However using errors will just break the current IstioOperator CR, which is a compatibility issue.

litong01: @timonwong OK. Thanks. then I am ok with the changes. Thanks.

litong01 (approved):

PR #36930, opened 2022-01-21
customizable image for operator manifest

timonwong:


litong01:
Looks like that the commit is not including the `make gen` result, before commit, that command needs to run if some definitions were changed.

manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
```diff
@@ -53,7 +53,7 @@ data:
         zipkin:
           address: zipkin.istio-system:9411
     enablePrometheusMerge: true
-    rootNamespace: null
+    rootNamespace: istio-system
```
litong01, line 56: hmmm, is this really necessary? I thought that the code will default to istio-system if there is no specified.

manifests/charts/istio-operator/files/gen-operator.yaml
```diff
@@ -173,7 +173,7 @@ spec:
       serviceAccountName: istio-operator
       containers:
         - name: istio-operator
-          image: gcr.io/istio-testing/operator:1.14-dev
+          image: "gcr.io/istio-testing/operator:1.14-dev"
```
litong01, line 176: hmmm, do not think that the double quotes are needed, this change will make this one instance different from others, would not think this is needed.

litong01 (changes requested):

manifests/charts/istio-operator/files/gen-operator.yaml
timonwong: @litong01 I enclosed `image` with double quotes because other occurrences for `image` are double quoted.

releasenotes/notes/36930.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: feature
+area: installation
+
+releaseNotes:
+  - |
+    **Added** add customizable image to istio operator manifests, add `--image` to both `istioctl operator init` and `istioctl operator dump`.
```
ericvn, line 7:
```suggestion
    **Added** customizable image to istio operator manifests, adding `--image` to both `istioctl operator init` and `istioctl operator dump`.
```

ericvn (approved):

manifests/charts/istio-operator/files/gen-operator.yaml
litong01: Majority of the image values are not quoted. the only places that I can see quoted are places where scripts are needed. Other places that use mixed (quoted and unquoted) are from testdata which should not have been. I would prefer not to quote it. If you leave this one and the one above alone, your PR can have one less file changes.

jacob-delgado:
What is the use case for this? People want to use the istio operator, but not the image Istio creates?

PR #36931, opened 2022-01-21
Use leases lock for istio operator's resourcelock

timonwong:

Since 1.9 (https://github.com/istio/istio/commit/8ed67b3c4f2ef6691d65925b7c999652e0d95df8), the default resource lock is `configmapsleases`, now we can use "leases" only, which performs better than `configmaps` (for legacy reasons)

- [ ] Docs
- [X] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/operator/cmd/operator/server.go
+++ b/operator/cmd/operator/server.go
@@ -24,10 +24,10 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
 	"go.opencensus.io/stats/view"
-	"k8s.io/client-go/tools/leaderelection/resourcelock"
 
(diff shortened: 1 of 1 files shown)


Arhell (commented):
/ok-to-test

timonwong:
Is release note required?

And besides, why the @istio-policy-bot didn't run for some PRs?

morvencao:
@timonwong Is there logged issue for this fix? If yes, you'd better add release note.

howardjohn:
I would add a release note as it could be a breaking change if someone has
a bespoke config (since it uses new RBAC permission). Shouldn't be an issue
for users using the normal install method but good to call out just in case

PR #36933, opened 2022-01-21
Add warning when file mounted certs are used at Gateway.

howardjohn:
Using Secrets has been the recommended source of gateway credentials for
2+ years now. This introduces a warning when users use file mounted
certs, to suggest they instead use `credentialName`.

This is purely a warning. The file mounted certs will continue to work,
both now and in the future.

This also fixes a bug in the cipher validation. We exit early on warning, which means we may ignore errors later on.

### DIFF
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -574,12 +574,13 @@ func validateTLSOptions(tls *networking.ServerTLSSettings) (v Validation) {
 			}
 		}
 	}
+
 	if len(invalidCiphers) > 0 {
-		return WrapWarning(fmt.Errorf("ignoring invalid cipher suites: %v", invalidCiphers.SortedList()))
+		v = appendWarningf(v, "ignoring invalid cipher suites: %v", invalidCiphers.SortedList())
 	}
 
 	if len(duplicateCiphers) > 0 {
-		return WrapWarning(fmt.Errorf("ignoring duplicate cipher suites: %v", duplicateCiphers.SortedList()))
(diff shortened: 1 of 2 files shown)


pkg/config/validation/validation.go
```diff
@@ -598,6 +599,16 @@ func validateTLSOptions(tls *networking.ServerTLSSettings) (v Validation) {
 		return
 	}
 
+	if tls.ServerCertificate != "" {
```
ramaraochavali, line 602: This is actually a valid setup and people like us use it. Is there a specific reason why we want to introduce this new warning now?

hzxuzhonghu: Feel the same if the file mounted cert feature is supported forever, why warn it

howardjohn: Users get confused that they *should use* it, when 99% of users should not. And most other ingress do not have this either, so folks get mislead into thinking they need to *migrate* to this model. That is my impression anyhow. Our docs do not have filemounted referenecs anymore at all (except in reference)

hzxuzhonghu: AFAIK, sds could replace file mounted certs completely. And we should suggest changing to use it by changing the api doc.

howardjohn:
I did send an API PR as well. I am ok only doing that.

BTW for some edge cases file mount is really the only option - look CSI
driver mounting certs, etc. But not common

ramaraochavali:
API PR looks good to me with examples. My preference would be to just do that and not add warning.

PR #36959, opened 2022-01-24
respect prometheus annotations

### LINKED ISSUE #31187: metrics merging: detect prometheus.io.scrape format
Due to escaping logic in prom, this is equivilent to prometheus.io/scrape (`__meta_kubernetes_pod_annotation_prometheus_io_path`). So we probably want to detect both. See https://github.com/openfaas/faas-netes/issues/714.

zirain:

fixed  #31187

### DIFF
--- a/pkg/kube/inject/webhook.go
+++ b/pkg/kube/inject/webhook.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	kjson "k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"k8s.io/kube-openapi/pkg/util/sets"
 
 	"istio.io/api/annotation"
 	"istio.io/api/label"
@@ -73,6 +74,7 @@ const (
 	// for more details, please checkout [here](https://github.com/prometheus/prometheus/blob/71a0f42331566a8849863d77078083edbb0b3bc4/util/strutil/strconv.go#L40)
 	prometheusScrapeAnnotation = "prometheus_io_scrape"
 	prometheusPortAnnotation   = "prometheus_io_port"
+	prometheusPathAnnotation   = "prometheus_io_path"
 
 	watchDebounceDelay = 100 * time.Millisecond
 )
@@ -582,6 +584,10 @@ func applyRewrite(pod *corev1.Pod, req InjectionParameters) error {
 	return nil
 }
 
+var (
+	emptyScrape = status.PrometheusScrapeConfiguration{}
+)
+
 // applyPrometheusMerge configures prometheus scraping annotations for the "metrics merge" feature.
 // This moves the current prometheus.io annotations into an environment variable and replaces them
 // pointing to the agent.
@@ -596,14 +602,9 @@ func applyPrometheusMerge(pod *corev1.Pod, mesh *meshconfig.MeshConfig) error {
 				return nil
 			}
 		}
-		scrape := status.PrometheusScrapeConfiguration{
-			Scrape: pod.Annotations["prometheus.io/scrape"],
-			Path:   pod.Annotations["prometheus.io/path"],
-			Port:   pod.Annotations["prometheus.io/port"],
-		}
-		empty := status.PrometheusScrapeConfiguration{}
+		scrape := getPrometheusScrapeConfiguration(pod)
 		sidecar := FindSidecar(pod.Spec.Containers)
-		if sidecar != nil && scrape != empty {
+		if sidecar != nil && scrape != emptyScrape {
 			by, err := json.Marshal(scrape)
 			if err != nil {
 				return err
@@ -613,6 +614,10 @@ func applyPrometheusMerge(pod *corev1.Pod, mesh *meshconfig.MeshConfig) error {
 		if pod.Annotations == nil {
 			pod.Annotations = map[string]string{}
 		}
+		// if a user sets `prometheus/io/path: foo`, then we add `prometheus.io/path: /stats/prometheus`
+		// prometheus will pick a random one
+		// need to clear out all variants and then set ours
+		clearPrometheusAnnotations(pod)
 		pod.Annotations["prometheus.io/port"] = targetPort
 		pod.Annotations["prometheus.io/path"] = "/stats/prometheus"
 		pod.Annotations["prometheus.io/scrape"] = "true"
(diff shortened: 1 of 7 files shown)


pkg/kube/inject/webhook.go
```diff
@@ -608,10 +609,48 @@ func applyPrometheusMerge(pod *corev1.Pod, mesh *meshconfig.MeshConfig) error {
...
 
+var (
+	prometheusScrapeAnnotations = []string{
+		"prometheus.io/scrape",
+		"prometheus.io.scrape",
+	}
+
+	prometheusPortAnnotations = []string{
+		"prometheus.io/port",
+		"prometheus.io.port",
```
howardjohn, line 626: Prometheus has a variety of charecters that it escapes to the same thing. I don't know the fully list, but ideally we normalize the annotations in the same way they do.

For example `prometheus/io/port` would work with prometheus but this logic would not handle it

howardjohn (commented):
Looks really close, thanks for the PR

pkg/kube/inject/webhook.go
zirain: you're right, `prometheus/io/port` also will convert to `__meta_kubernetes_pod_annotation_prometheus_io_port`

zirain:
according to  [SanitizeLabelName](https://github.com/prometheus/prometheus/blob/71a0f42331566a8849863d77078083edbb0b3bc4/util/strutil/strconv.go#L40)

@howardjohn please review again.

zirain:
BTW, do we need a release notes?

pkg/kube/inject/webhook.go
```diff
@@ -608,10 +610,45 @@ func applyPrometheusMerge(pod *corev1.Pod, mesh *meshconfig.MeshConfig) error {
...
+		return nil
 	}
+
 	return nil
 }
 
+var (
+	prometheusScrapeAnnotation = "prometheus_io_scrape"
+	prometheusPortAnnotation   = "prometheus_io_port"
+)
```
hzxuzhonghu, line 622: what's this?

zirain: prometheus will convert annotation to this format see [here](https://github.com/prometheus/prometheus/blob/71a0f42331566a8849863d77078083edbb0b3bc4/discovery/kubernetes/pod.go#L205)

`prometheus.io/scrape` `prometheus.io.scrape` `prometheus-io/scrape` have the same meaning in Prometheus

howardjohn: Putting that comment in the code would be ideal for future reference

pkg/kube/inject/webhook.go
```diff
@@ -608,10 +610,45 @@ func applyPrometheusMerge(pod *corev1.Pod, mesh *meshconfig.MeshConfig) error {
 		pod.Annotations["prometheus.io/port"] = targetPort
 		pod.Annotations["prometheus.io/path"] = "/stats/prometheus"
 		pod.Annotations["prometheus.io/scrape"] = "true"
```
howardjohn, line 612: What if a user sets `prometheus/io/path: foo`. Then we add `prometheus.io/path: /stats/prometheus`. Then won't we end up with *both*, at which point prometheus may pick a random one?

We may need to clear out all variants and then set ours?

howardjohn (commented):
Nice!

pkg/kube/inject/webhook.go
zirain:
```golang
        for k, v := range pod.Annotations {
		ln := strutil.SanitizeLabelName(k)
		ls[model.LabelName(podAnnotationPrefix+ln)] = lv(v)
		ls[model.LabelName(podAnnotationPresentPrefix+ln)] = presentValue
	}
```
I think prometheus will take a random one due to range from a map 

I prefer to merge istio metric to the path which's already settled in annotations, wdyk @howardjohn

zirain:
```
We may need to clear out all variants and then set ours?
```
seems reasonable, Is there something told end users about this auto-merge behavior?

howardjohn: We do say https://istio.io/latest/docs/ops/integrations/prometheus/#option-1-metrics-merging they will be overriden so it seems ok to override? WDYT

zirain: updated

pkg/kube/inject/webhook.go
```diff
@@ -605,13 +612,84 @@ func applyPrometheusMerge(pod *corev1.Pod, mesh *meshconfig.MeshConfig) error {
...
+		if scrape, err := strconv.ParseBool(val); err == nil {
+			return scrape
+		}
+	}
+
+	return true
+}
+
+func clearPrometheusAnnotations(pod *corev1.Pod) {
+	prometheusAnnotations := sets.NewString(
```
hzxuzhonghu, line 645: This can be moved out

hzxuzhonghu (commented):
LGTM

releasenotes/notes/36959.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: bug-fix
+area: installation
+
+# issue is a list of GitHub issues resolved in this note.
+issue:
+  - 31187
+
+releaseNotes:
+- |
+  **Fixed** detect `prometheus.io.scrape` format when merging metrics
```
ericvn, line 11: Maybe:
```suggestion
  **Fixed** detection of `prometheus.io.scrape` annotations when merging metrics.
```
```suggestion
  **Fixed** detect `prometheus.io.scrape` format when merging metrics
```

ericvn, line 11: Maybe.
```suggestion
  **Fixed** detection of `prometheus.io.scrape` annotations when merging metrics.
```

ericvn (approved):

zirain:
ping @howardjohn @hzxuzhonghu

pkg/kube/inject/webhook.go
```diff
@@ -36,6 +37,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	kjson "k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"k8s.io/kube-openapi/pkg/util/sets"
```
howardjohn, line 40: nit: use `pilot/pkg/util/sets`

howardjohn (approved):
:+1:

Issue #29131, opened 2020-11-23
Re-enable PILOT_FILTER_GATEWAY_CLUSTER_CONFIG


howardjohn:
Blocking:
1. Fix https://github.com/istio/istio/issues/28315
2. Provide a way for ext_authz services to be sent. Probably also will need Telemetry provider as well when we have that?
3. Either automagically solve EnvoyFilter references as well, OR add upgrade note for this and provide some way to manually work around it
4. Fix https://github.com/istio/istio/issues/29310 (new cluster races reference to it)

howardjohn:
cc @yangminzhu for ext_authz, @douglas-reid for telemetry, @mandarjog since you have looked into this quite a bit.

This flag filers out unused clusters in the gateways, where unused is currently determined by VirtualService only

mandarjog:
For infra services. The choices are (previously discussed)
1. Use service entry to provide additional destinations . The destinations need to be available, but not routable.
2. If using EnvoyFilter, also add the cluster definition.
3. Add infra-services to proxyConfig / meshConfig . Now this looks similar to ExternalProvider.

howardjohn:
not stale

howardjohn:
@ramaraochavali proposed adding `exportTo: gateway` annotation to ensure a service is always included in gateway for the Envoyfilter issue

howardjohn:
not stale

zirain:
@howardjohn I think (2) is fixed by https://github.com/istio/istio/issues/37997, anything I missed?

howardjohn:
I think the best way to solve (4) is to only support it on delta XDS, and just delay 'deletes' on clusters (https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#eventual-consistency-considerations). Something like if there is a push that applies to LDS/RDS _and_ CDS, all CDS deletes are stored in `model.Proxy` (`PendingDeletes` or something). Then after we finish the full push chain we push those removals.

Its pretty hard without delta since we would need to be able to generate the old clusters. Could be done with OldPushContext probably though

howardjohn:
Tried this a bit in https://github.com/howardjohn/istio/tree/gateway/cluster-scope, didn't work out

thechristschn:
What is the current status of this issue?
We would really like to use this option, but only if we know, that it is safe to use.

I've seen some fixes in newer istio versions in regards of this option, but not sure if all know problems are now resolved.

howardjohn:
[...]

PR #36986, opened 2022-01-25
Enable gateway cluster config filtering 

hzxuzhonghu:

set PILOT_FILTER_GATEWAY_CLUSTER_CONFIG=true


hzxuzhonghu:
@howardjohn

howardjohn:
I think we need to fix https://github.com/istio/istio/issues/29131 first. Otherwise users will get 503s

hzxuzhonghu:
I thought is had been fixed

hzxuzhonghu:
Will double check

PR #37002, opened 2022-01-25
Fix linting issues with golangci-lint 1.44

howardjohn:
Most of these are actually from a bug where it would fail to lint some files, not new linters.


lei-tang (approved):

douglas-reid (approved):
approving p&t owned files.

jacob-delgado (approved):

nmittler (approved):

stevenctl (approved):

PR #37010, opened 2022-01-26
Remove duplicate networks handler registeration

hzxuzhonghu:

### DIFF
--- a/pilot/pkg/bootstrap/mesh.go
+++ b/pilot/pkg/bootstrap/mesh.go
@@ -85,8 +85,11 @@ func (s *Server) initMeshConfiguration(args *PilotArgs, fileWatcher filewatcher.
 	// Watch the istio ConfigMap for mesh config changes.
 	// This may be necessary for external Istiod.
 	configMapName := getMeshConfigMapName(args.Revision)
-	s.environment.Watcher = kubemesh.NewConfigMapWatcher(
+	multiWatcher := kubemesh.NewConfigMapWatcher(
 		s.kubeClient, args.Namespace, configMapName, configMapKey, multiWatch, s.internalStop)
+	s.environment.Watcher = multiWatcher
+	s.environment.NetworksWatcher = multiWatcher
+	log.Infof("initializing mesh networks from mesh config watcher")
 
 	if multiWatch {
 		kubemesh.AddUserMeshConfig(s.kubeClient, s.environment.Watcher, args.Namespace, configMapKey, features.SharedMeshConfig, s.internalStop)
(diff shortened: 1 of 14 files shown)


pilot/pkg/bootstrap/mesh.go
```diff
@@ -96,11 +99,7 @@ func (s *Server) initMeshConfiguration(args *PilotArgs, fileWatcher filewatcher.
 // initMeshNetworks loads the mesh networks configuration from the file provided
 // in the args and add a watcher for changes in this file.
 func (s *Server) initMeshNetworks(args *PilotArgs, fileWatcher filewatcher.FileWatcher) {
-	if mw, ok := s.environment.Watcher.(mesh.NetworksWatcher); ok {
```
hzxuzhonghu, line 99: removed this hack, and it is done in L91

pilot/pkg/bootstrap/server.go
```diff
@@ -1186,12 +1186,6 @@ func (s *Server) initMeshHandlers() {
 			Reason: []model.TriggerReason{model.GlobalUpdate},
 		})
 	})
-	s.environment.AddNetworksHandler(func() {
```
hzxuzhonghu, line 1189: This is duplicate with the one registered in `NetworkManager`

pkg/kube/configmapwatcher/configmapwatcher.go
```diff
@@ -64,10 +64,7 @@ func NewController(client kube.Client, namespace, name string, callback func(*v1
 		Core().V1().ConfigMaps()
 
 	c.queue = controllers.NewQueue("configmap "+name, controllers.WithReconciler(c.processItem))
-	c.informer.Informer().AddEventHandler(controllers.FilteredObjectSpecHandler(c.queue.AddObject, func(o controllers.Object) bool {
-		// Filter out configmaps
-		return o.GetName() == name && o.GetNamespace() == namespace
```
howardjohn, line 69: why remove the filter?

howardjohn (commented):
lgtm other than filter

pkg/kube/configmapwatcher/configmapwatcher.go
hzxuzhonghu: see above L60-L62, ns/name  is specified there

howardjohn: Oh yeah. Wonder why we had that

hzxuzhonghu: I know why we need it here. The tests failed because k8s fake client list-watch does not support field selector yet

PR #37019, opened 2022-01-26
refactor: merge JWTPath and credential fetcher

howardjohn:
Currently we have two ways to fetch credentials:
* JWTPath, which reads from a jwt token file
* CredentialFetcher, which is a generic interface to fetch credentials

This makes reading from a jwt token file a CredentialFetcher,
simplifying the code a bit

### DIFF
--- a/pkg/security/security.go
+++ b/pkg/security/security.go
@@ -67,6 +67,9 @@ const (
 	// GCE is Credential fetcher type of Google plugin
 	GCE = "GoogleComputeEngine"
 
+	// JWT is a Credential fetcher type that reads from a JWT token file
+	JWT = "JWT"
+
 	// Mock is Credential fetcher type of mock plugin
 	Mock = "Mock" // testing only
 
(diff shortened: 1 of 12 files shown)


stevenctl (approved):

pkg/security/security.go
```diff
@@ -67,6 +67,9 @@ const (
 	// GCE is Credential fetcher type of Google plugin
 	GCE = "GoogleComputeEngine"
 
+	// JWT is a Credential fetcher type that reads from a JWT token file
+	JWT = "JWT"
```
costinm, line 71: - I would not call it a fetcher type - maybe file
- I don't think we should explicitly set it, instead if file is found, use it else fallback to compute engine.

costinm (approved):
Strict improvement over what we have - but see my comment on naming, and ideally if we can have automatic it will simplify the setup.

pkg/security/security.go
howardjohn: Its tricky to do fallback - the GCE type actually uses the same file as part of its auth. Unless we have a smarter/different way to detect?

PR #37048, opened 2022-01-27
Revert "send unready endpoints as unhealthy"

howardjohn:
Reverts istio/istio#36274

Revert while we explore test failures

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -105,12 +105,6 @@ var (
 			" EDS pushes may be delayed, but there will be fewer pushes. By default this is enabled",
 	).Get()
 
-	SendUnhealthyEndpoints = env.RegisterBoolVar(
-		"PILOT_SEND_UNHEALTHY_ENDPOINTS",
-		true,
-		"If enabled, Pilot will include unhealthy endpoints in EDS pushes and even if they are sent Envoy does not use them for load balancing.",
-	).Get()
-
 	// HTTP10 will add "accept_http_10" to http outbound listeners. Can also be set only for specific sidecars via meta.
 	HTTP10 = env.RegisterBoolVar(
 		"PILOT_HTTP10",
--- a/pilot/pkg/model/service.go
+++ b/pilot/pkg/model/service.go
@@ -387,16 +387,6 @@ type Locality struct {
 	ClusterID cluster.ID
 }
 
-// Endpoint health status.
-type HealthStatus int32
-
-const (
-	// Healthy.
-	Healthy HealthStatus = 0
-	// Unhealthy.
-	UnHealthy HealthStatus = 1
-)
-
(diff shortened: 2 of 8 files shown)


ericvn (approved):

bianpengyuan (approved):

howardjohn:
Release notes is broken... force it

ramaraochavali:
@howardjohn curious Did you find any thing?

bianpengyuan:
I dug a bit at this, cannot find any clue. For this failure: https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-119_istio_postsubmit/1486623852787142656, in the config dump https://storage.googleapis.com/istio-prow/logs/integ-k8s-119_istio_postsubmit/1486623852787142656/artifacts/security-ca-custom-root-f813723/TestTrustDomainAliasSecureNaming/From_primary-0/server:server-naked-bar[fail]/_test_context/test-ns-1-85017-state294247395/cluster-0/client-v1-b5c4cdb65-dwr2t_proxy-config.json, cluster `outbound|443||server-naked-bar.test-ns-1-85017.svc.cluster.local` has endpoint misses SAN. However in the pilot log, I can see log lines suggested "service account changed" etc for `server-naked-bar`, which should trigger a cluster update with service accounts: https://storage.googleapis.com/istio-prow/logs/integ-k8s-119_istio_postsubmit/1486623852787142656/artifacts/security-ca-custom-root-f813723/TestTrustDomainAliasSecureNaming/_test_context/istio-state1696984509/cluster-0/istiod-7d5bd95456-cwwxw_discovery.log. I looked through the code path, but cannot see how could this happen...

ramaraochavali:
Ok. Thank you. I also could not find any thing based on above logs.

howardjohn:
Yeah I don't know _why_ it happens but somehow we miss the SANs. I am pretty confident it was this PR since it broke after merge and fixed after revert, but weird its only postsubmit... 

It is somewhat related though - this changes full vs incremental which could cause this

ramaraochavali:
>  It is somewhat related though - this changes full vs incremental which could cause this

The conditions under which it does full push have not been changed and it returns early in those conditions.

Only way this can happen is full push happens but endpoints does not have serviceaccount populated.

ramaraochavali:
I think I see what is going on 

In the old model where it is working, we hit this code https://github.com/istio/istio/blob/master/pilot/pkg/xds/eds.go#L128 multiple times before a full push happens and I _guess_  https://github.com/istio/istio/blob/master/pilot/pkg/xds/eds.go#L220 is ensuring SA are updated correctly.

With the new code, we only hit https://github.com/istio/istio/blob/master/pilot/pkg/xds/eds.go#L128  only once and rest of the times the len(IstioEndpoints) is not zero as it has unhealthy endpoints so it is skipping that logic. But still not clear to me why is this only happening in postsubmit and why it relies on deleteEndpointShards and how it is fully helping.

Need to dig more

Issue #36829, opened 2022-01-13
TestJWTHTTPS test flakes


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-123_istio_postsubmit/1481664052441649152

PR #37054, opened 2022-01-27
Skip flaky TestJWTHTTPS

howardjohn:
See https://github.com/istio/istio/issues/36829. This is a brand new test so we are not regression on coverage

### DIFF
--- a/tests/integration/security/https_jwt/main_test.go
+++ b/tests/integration/security/https_jwt/main_test.go
@@ -46,6 +46,7 @@ func loadCert(filename string) (string, error) {
 func TestMain(m *testing.M) {
 	framework.
 		NewSuite(m).
+		Skip("https://github.com/istio/istio/issues/36829").
 		Setup(istio.Setup(&ist, setupConfig)).
 		Setup(func(ctx resource.Context) error {
 			return util.SetupApps(ctx, ist, apps, true)


lei-tang (approved):

tests/integration/security/https_jwt/main_test.go
```diff
@@ -46,6 +46,7 @@ func loadCert(filename string) (string, error) {
 func TestMain(m *testing.M) {
 	framework.
 		NewSuite(m).
+		Skip("https://github.com/istio/istio/issues/36829").
```
yangminzhu, line 49: Is the test still flaky? it seems it only failed once in the past week: https://k8s-testgrid.appspot.com/istio_istio_postsubmit#integ-k8s-123_istio_postsubmit&width=20&include-filter-by-regex=TestJWTHTTPS

Also I think @aryan16 has a PR to improve the test by adding the readiness probe

howardjohn:
I don't see any PRs since the original. The testgrid is a bit misleading
since its split up between a variety of different jobs

https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-116_istio_postsubmit/1484396598837383168
https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-117_istio_postsubmit/1486407201449840640
https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-119_istio_postsubmit/1486431867178586112
https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-119_istio_postsubmit/1486560344758292480

aryan16:
Here is the PR - https://github.com/istio/istio/pull/36901, but it has a solution with initial delay. If I remove the delay from it, test will fail. We can try to add it to see if it fixes the flaky tests.

PR #37057, opened 2022-01-27
istio-iptables: support constraints on groups outgoing traffic from which should be redirected to Envoy

yskopets:

This PR gives users an option to put constraint on groups [outgoing] traffic from which should be redirected to Envoy.

The primary use case is Istio Sidecar running on a VM.

It is a common expectation for VM users to be able to redirect outgoing traffic only from certain applications rather than all outgoing traffic.

This PR introduces support for 2 scenarios:
* redirect [outgoing] traffic from all groups except for a given list
* redirect [outgoing] traffic only from a given list of groups

This configuration model is similar to `--istio-service-cidr / --istio-service-exclude-cidr`.

This PR introduces new environment variables into `istio-iptables` and `istio-clean-iptables` binaries:

* `ISTIO_OUTBOUND_OWNER_GROUPS`
* `ISTIO_OUTBOUND_OWNER_GROUPS_EXCLUDE`


howardjohn:
I can get a fuzzy idea of how this would work, but could you elaborate a bit? I am stuck on why they want outbound captured but not inbound. One example I can think of is we have app (uid=10) and management tool (uid=20). The management tool just makes outgoing requests to get some metadata for whatever, and we want to opt it out. It doesn't have inbound requests.

is that the use case?

tools/istio-clean-iptables/pkg/cmd/root.go
```diff
@@ -130,6 +156,26 @@ func init() {
...
+			"The wildcard character \"*\" can be used to configure redirection of traffic from all users. "+
+			"A user can be specified either by name or a numeric UID.")
+
+	rootCmd.Flags().String(constants.OwnerUsersExclude, "",
+		"Comma separated list of users whose [outgoing] traffic is to be excluded from redirection to Envoy. "+
+			"Only applies when traffic from all users (i.e. \"*\") is being redirected to Envoy. "+
+			"A user can be specified either by name or a numeric UID.")
+
+	rootCmd.Flags().String(constants.OwnerGroupsInclude, "",
+		"Comma separated list of groups whose [outgoing] traffic is to be redirected to Envoy. "+
```
howardjohn, line 171: nit: can use ` to do multiline string which can help avoid the need to escape

tools/istio-iptables/pkg/capture/testdata/outbound-owner-users.golden
```diff
@@ -0,0 +1,17 @@
...
+iptables -t nat -A ISTIO_IN_REDIRECT -p tcp -j REDIRECT --to-ports 15006
+iptables -t nat -A OUTPUT -p tcp -j ISTIO_OUTPUT
+iptables -t nat -A ISTIO_OUTPUT -o lo -s 127.0.0.6/32 -j RETURN
+iptables -t nat -A ISTIO_OUTPUT -o lo ! -d 127.0.0.1/32 -m owner --uid-owner 1337 -j ISTIO_IN_REDIRECT
+iptables -t nat -A ISTIO_OUTPUT -o lo -m owner ! --uid-owner 1337 -j RETURN
+iptables -t nat -A ISTIO_OUTPUT -m owner --uid-owner 1337 -j RETURN
+iptables -t nat -A ISTIO_OUTPUT -o lo ! -d 127.0.0.1/32 -m owner --gid-owner 1337 -j ISTIO_IN_REDIRECT
+iptables -t nat -A ISTIO_OUTPUT -o lo -m owner ! --gid-owner 1337 -j RETURN
+iptables -t nat -A ISTIO_OUTPUT -m owner --gid-owner 1337 -j RETURN
+iptables -t nat -A ISTIO_OUTPUT -m owner ! --uid-owner 101 -m owner ! --uid-owner tomcat -j RETURN
```
howardjohn, line 16: how big can we have a rule? If I set 100 users in the include list, is it ok?

howardjohn (commented):
Generally looks good, just one small question to help understand the use cases

costinm:
Can we:
- restrict this to only groups - I don't think we need both user and groups, too much complexity
- use only env variables ( if it is experimental ), or ProxyConfig - if not ( the plan was to move things like this in ProxyConfig AFAIK ?)
- I assume annotations will also need to be defined, for parity with the rest ? Or we'll just use the ProxyConfig annotation ? 

I think it's a good idea for VMs and not very harmful, and if there are strong use cases for UIDs that can't be addressed with GIDs 
we can also add the 2 options for users, but we can start with GID maybe ?

tools/istio-clean-iptables/pkg/cmd/root.go
costinm: Nit - we should stop adding flags, env variables are sufficient and easier to handle. 
And ProxyConfig too.

We need to start getting rid of the flags we have that duplicate ProxyConfig features too.

yskopets:
@howardjohn Hey

> I am stuck on why they want outbound captured but not inbound.

There are 2 reasons:
* restrictions on UID/GID are [only applicable](https://fossies.org/linux/iptables/extensions/libxt_owner.man) to the outgoing traffic (`OUTPUT` chain)
* for the incoming traffic there are already options for [including/excluding individual ports](https://github.com/istio/istio/blob/ab43f0b744ef2b9c1ba1778844f24d78e33575ea/tools/istio-iptables/pkg/cmd/root.go#L380-L386) from redirection; I think it should be enough

tools/istio-iptables/pkg/capture/testdata/outbound-owner-users.golden
yskopets: I will try to find any docs in Iptables and get back to you.

But, in any case, we can set our own limit. Let's say `20`.

tools/istio-clean-iptables/pkg/cmd/root.go
yskopets: @costinm I think your comment applies to both commands - `istio-iptables` and `istio-clean-iptables`.

But I'd like to ask a question about `istio-clean-iptables` specifically.

Ideally, I shouldn't be passing settings like a list of UID/GID into `istio-clean-iptables` at all.

As a user, I would expect that `istio-clean-iptables` should simply delete previously installed Iptables rules, even if I don't remember what parameters I used when I called `istio-iptables` earlier.

But, apparently, `istio-clean-iptables` must be given exactly the same parameters as `istio-iptables`. Otherwise, some Iptables rules might not be deleted.

The only reason why `istio-clean-iptables` needs a list of UID/GID, is because `istio-iptables` command appends rules for redirection of DNS over UDP traffic directly into the built-in `OUTPUT` chain instead of using a user-defined chain, e.g. `ISTIO_OUTPUT`, like it's done for all other inbound and outbound traffic.

Is it ok if I refactor `istio-iptables` command to use a user-defined chain for rules of the DNS over UDP traffic? E.g., as a separate PR.

yskopets:
@costinm Hey

> Can we restrict this to only groups?

Sure. I will update the PR.

> Can we use only env variables?

Sure.

> Can we use ProxyConfig? and define annotations ?

In our experience, users who want this level of precision over traffic redirection would not put this configuration into a k8s object but rather keep it local to the VM.

So, I would suggest to start from the environment variable only. And consider `ProxyConfig` and annotation after users explicitly request it.

> but we can start with GID maybe ?

Sure.

costinm:
By PrroxyConfig I meant the json object we pass in an env variable, not the CRD :-)

We can start with the env variable, and promote them if more users are interested.

howardjohn (approved):

tools/istio-iptables/pkg/capture/testdata/outbound-owner-users.golden
yskopets: @howardjohn I didn't find the docs on this subject. So, I tried in practice.

Apparently, the maximum number of match expressions varies across Linux distros:
* `Debian` `10`,`11` :  ~128
*  `Ubuntu` `18.04`, `20.04` : ~1150
* `CentOS 7` :  ~1150
* `RHEL 7` : ~1150
* `RHEL 8` : ~128

If the number of match expressions exceeds the limit, `iptables` command fails.

The error message varies, but mostly cryptic - `iptables: No space left on device.`

Sometimes there is a better message in the kernel log - `netlink: 'iptables': attribute type 3 has an invalid length.`

Do you want me to enforce some limit in the code?

howardjohn: Yeah maybe ~64? Seems like a reasonable upper bound and gives us some leway if its somehow less than 128?

yskopets: ok

yskopets: @howardjohn done.

Also added more unit tests and release notes.

PR #37063, opened 2022-01-28
fix flaky workload entry to ports

timonwong:

Fix flaky test introduced in #36504

https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/36930/unit-tests_istio/1486955106078298112

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [X] Test and Release
- [ ] User Experience

### DIFF
--- a/istioctl/cmd/workload.go
+++ b/istioctl/cmd/workload.go
@@ -537,15 +537,20 @@ func createMeshConfig(kubeClient kube.ExtendedClient, wg *clientv1alpha3.Workloa
 
 func marshalWorkloadEntryPodPorts(p map[string]uint32) string {
 	var out []model.PodPort
+
+	if len(p) == 0 {
+		return ""
+	}
+
 	for name, port := range p {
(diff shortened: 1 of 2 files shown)


timonwong:
should we label `release-notes-none` ?

howardjohn (commented):
Thanks! I think container Port may not be unique always so it could still be non deterministic

https://github.com/istio/istio/pull/37045 has similar change fwiw but happy to merge this as well if it gets approved. Thanks for caring about test stability

timonwong:
> Thanks! I think container Port may not be unique always so it could still be non deterministic
> 
> #37045 has similar change fwiw but happy to merge this as well if it gets approved. Thanks for caring about test stability

@howardjohn Thanks for pointing out the port number issue, now it's sorted by port name instead.

stevenctl (approved):

timonwong:
Since #37045 is merged, is this PR still useful?

PR #37075, opened 2022-01-29
 Credential controller: make use of generic event handler wrapper

hzxuzhonghu:

This is to prevent adding handler after secret informer has been started.

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/credentials/kube/multicluster.go
+++ b/pilot/pkg/credentials/kube/multicluster.go
@@ -91,7 +91,6 @@ func (m *Multicluster) ForCluster(clusterID cluster.ID) (credentials.Controller,
 	return agg, nil
 }
 
-// AddSecretHandler must be called before starting. Otherwise, the handler will not take effect.
 func (m *Multicluster) AddSecretHandler(h secretHandler) {
 	m.secretHandlers = append(m.secretHandlers, h)
 	for _, c := range m.remoteKubeControllers {
--- a/pilot/pkg/credentials/kube/multicluster.go
+++ b/pilot/pkg/credentials/kube/multicluster.go
@@ -24,14 +24,14 @@ import (
 	"istio.io/pkg/log"
 )
 
-type eventHandler func(name string, namespace string)
+type secretHandler func(name string, namespace string)
 
 // Multicluster structure holds the remote kube Controllers and multicluster specific attributes.
 type Multicluster struct {
 	remoteKubeControllers map[cluster.ID]*CredentialsController
 	m                     sync.Mutex // protects remoteKubeControllers
 	localCluster          cluster.ID
-	eventHandlers         []eventHandler
+	secretHandlers        []secretHandler
 }
 
 var _ credentials.MulticlusterController = &Multicluster{}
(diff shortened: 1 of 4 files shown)


pilot/pkg/credentials/kube/secrets.go
```diff
@@ -286,49 +318,3 @@ func extractRoot(scrt *v1.Secret) (cert []byte, err error) {
 	return nil, fmt.Errorf("found secret, but didn't have expected keys %s or %s; found: %s",
 		GenericScrtCaCert, TLSSecretCaCert, found)
 }
-
-func (s *CredentialsController) AddEventHandler(f func(name string, namespace string)) {
```
hzxuzhonghu, line 290: To prevent register handler after informer has been started

pilot/pkg/credentials/kube/secrets.go
```diff
@@ -96,8 +96,40 @@ func NewCredentialsController(client kube.Client, clusterID cluster.ID) *Credent
 		)
 	})
 
+	handler := func(obj interface{}) {
```
howardjohn, line 99: nit: replace L99-L129 with
```go
	controllers.ObjectHandler(func(o controllers.Object) {
		for _, h := range handlers {
			h(o.GetName(), o.GetNamespace())
		}
	})
```

pilot/pkg/credentials/kube/secrets.go
howardjohn: what is the problem with adding after start? you can add handlers and it will enqueue all objects

pilot/pkg/credentials/kube/multicluster.go
```diff
@@ -80,25 +77,19 @@ func (m *Multicluster) ForCluster(clusterID cluster.ID) (credentials.Controller,
...
 	agg.controllers = append(agg.controllers, m.remoteKubeControllers[m.localCluster])
 	return agg, nil
 }
 
-func (m *Multicluster) AddEventHandler(f eventHandler) {
-	m.eventHandlers = append(m.eventHandlers, f)
-	for _, c := range m.remoteKubeControllers {
-		c.AddEventHandler(f)
-	}
+func (m *Multicluster) AddSecretHandler(f secretHandler) {
```
howardjohn, line 91: nit: would be good to comment this MUST be called before starting?

howardjohn (commented):
Seems reasonable just not sure why its needed? And a small style nit

pilot/pkg/credentials/kube/secrets.go
hzxuzhonghu: First start, then register handler, it will miss the notification of some watched objects.

hzxuzhonghu: In the secret case, seems no influence. Just keep a good manner.

hzxuzhonghu: nope, IIRC the notification is based on recv from kube-apiserver

hzxuzhonghu: I will double check it

hzxuzhonghu: Ah, correct. https://github.com/kubernetes/client-go/blob/8f44946f6cbe967fbe2e2548e76987680a89428e/tools/cache/shared_informer.go#L561-L562)

hzxuzhonghu: I will revert some unnecessary changes

howardjohn (approved):
LGTM thanks!

PR #37086, opened 2022-01-31
vm: support health checks for VMs that are not using auto-registration

yskopets:

The PR implements support for health checking of those VMs that are not using auto-registration.

The notion of WorkloadEntry `"cleanup"` has been extended to accommodate 2 types VMs, namely:
* `"cleanup"` for VMs that do use auto-registration means "delete WorkloadEntry"
* `"cleanup"` for VMs that do NOT use auto-registration means "delete Health condition" (as health status is no longer known)

The overall purpose of this PR is to make `WorkloadEntry` the fundamental building block, so that it can be used even without higher level concepts such as auto-registration.

This PR does not introduce any API changes, however it makes possible a new use case that might need to be documented.

E.g., if a user wants to enable health-checking for a `WorkloadEntry` that is not using auto-registration, he/she needs to make the following changes to the configuration:

```yaml
kind: WorkloadEntry
metadata:
  annotations:
    # indicate that `istiod` should exclude this workload from load balancing
    # until `Healthy` condition is set to `True`
    proxy.istio.io/health-checks-enabled: "true"
spec:
  ...  
```


zirain:
`proxy.istio.io/config` is used to `Overrides for the proxy configuration for this specific proxy. Available options can be found at https://istio.io/docs/reference/config/istio.mesh.v1alpha1/#ProxyConfig.`

IMHO, we should not put `readinessProbe` configuration in annotations, prefer to extend `WorkloadEntry` definition

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -243,54 +244,76 @@ func (c *Controller) processNextWorkItem() bool {
...
-	// check if the WE already exists, update the status
-	entryName := autoregisteredWorkloadEntryName(proxy)
+	var entryName string
+	var autoCreate bool
+	if features.WorkloadEntryAutoRegistration &&
+		proxy.Metadata.AutoRegisterGroup != "" {
+		entryName = autoregisteredWorkloadEntryName(proxy)
+		autoCreate = true
+	} else if features.WorkloadEntryHealthChecks &&
+		proxy.Metadata.ProxyConfig != nil && proxy.Metadata.ProxyConfig.ReadinessProbe != nil {
```
howardjohn, line 272: Looking for `ProxyConfig.ReadinessProbe` in the metadata seems concerning to me. The control plane shouldn't care what the probe is - just if it should wait until its been marked "Ready" before using the WE or not (which the annotation does)

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -539,6 +616,17 @@ func (c *Controller) cleanupEntry(wle config.Config) {
 	log.Infof("cleaned up auto-registered WorkloadEntry %s/%s", wle.Namespace, wle.Name)
 }
 
+func (c *Controller) deleteHealthCondition(wle config.Config) {
+	wle = status.DeleteConfigCondition(wle, status.ConditionHealthy)
```
howardjohn, line 620: Is deleting it right? Doesn't that mean we mark it as unhealthy?

I would think that we would want to maintain last known state?

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -675,3 +763,46 @@ func (c *Controller) handleErr(err error, key interface{}) {
...
 	return string(proxy.Metadata.Network) + proxy.IPAddresses[0]
 }
+
+func lookupWorkloadEntryByProxyID(store model.ConfigStoreCache, proxy *model.Proxy) string {
+	if len(proxy.IPAddresses) == 0 {
+		return ""
+	}
+
+	parts := strings.Split(proxy.ID, ".")
+	if len(parts) != 2 || parts[1] != proxy.Metadata.Namespace {
```
howardjohn, line 773: can use `proxy.ConfigNamespace` instead of parts[1]

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -675,3 +763,46 @@ func (c *Controller) handleErr(err error, key interface{}) {
...
+		return ""
+	}
+	name := parts[0]
```
howardjohn, line 776: So this also adds a restriction that name of WE must match, right?

howardjohn (commented):
Only had a chance to look briefly, seems like a good idea overall but left a couple questions

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: Yes, my intention is to mark `WorkloadEntry` as unhealthy if sidecar doesn't reconnect within the grace period.

This behaviour is symmetrical to auto-registered `WorkloadEntry`(s) that get deleted if they don't reconnect within the grace period, which results in excluding this workload from load balancing, which is equivalent to marking workload unhealthy. 

I was choosing between:
1. setting `Healthy` condition to `False`
2. setting `Healthy` condition to `Unknown`
3. removing `Healthy` condition

I implemented removing `Healthy` condition since it requires less logic while the effect is the same.

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: Source code in this file uses `proxy.Metadata.Namespace` everywhere.

I can change to `proxy.ConfigNamespace`.

Should I change only this occurrence, or all occurrences in this file?

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: For the purposes of this feature, I need logic that finds `WorkloadEntry` corresponding to the connected proxy.

Right now this code, indeed, has only 1 strategy to find matching `WorkloadEntry`.

Ideally, I would like to reuse logic of [kube/controller](https://github.com/istio/istio/blob/06f42d2b95f14c00e6a989370c7975c552da8643/pilot/pkg/serviceregistry/kube/controller/controller.go#L1057-L1060), but I didn't find a way to achieve that.

Is it possible to introduce dependency on `kube/controller` in here?

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: Sorry, can you explain what would be the alternative to checking `proxy.Metadata.ProxyConfig != nil && proxy.Metadata.ProxyConfig.ReadinessProbe != nil` ?

I do this check to filter out early connected proxies that correspond to `Pods`.

Should I drop this check altogether and simply try to find `WorkloadEntry` ?

adiprerepa (approved):
lgtm but yield for @howardjohn

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: @howardjohn BTW, I've opened [another PR](https://github.com/istio/istio/pull/37129) to enhance how `kube/controller` searches for a matching `WorkloadEntry`.

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -243,54 +244,76 @@ func (c *Controller) processNextWorkItem() bool {
...
 
-	if err := c.registerWorkload(entryName, proxy, conTime); err != nil {
+	err := c.connectWorkload(entryName, proxy, conTime, autoCreate)
+	if err != nil {
 		log.Errorf(err)
+	}
+	return err
+}
+
+func (c *Controller) connectWorkload(entryName string, proxy *model.Proxy, conTime time.Time, autoCreate bool) error {
```
hzxuzhonghu, line 292: add an comment what does this function do

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -243,54 +244,76 @@ func (c *Controller) processNextWorkItem() bool {
...
-	entryName := autoregisteredWorkloadEntryName(proxy)
+	var entryName string
+	var autoCreate bool
+	if features.WorkloadEntryAutoRegistration &&
+		proxy.Metadata.AutoRegisterGroup != "" {
+		entryName = autoregisteredWorkloadEntryName(proxy)
+		autoCreate = true
+	} else if features.WorkloadEntryHealthChecks &&
+		proxy.Metadata.ProxyConfig != nil && proxy.Metadata.ProxyConfig.ReadinessProbe != nil {
+		entryName = lookupWorkloadEntryByProxyID(c.store, proxy)
```
hzxuzhonghu, line 273: What if the wle has not been registered when proxy connect to pilot

hzxuzhonghu: Or pilot has not yet discovered it

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: added comments to this and other functions affected by this PR

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: The feature introduced by this PR - health checks for VMs that are not using auto-registration - implies that `WorkloadEntry` resource must exist beforehand.

However, I see your point that there is a room for an error due to something that a user has no control over.

I think in a scenario where auto-registration is not used, a user expects that a proxy must correspond to an explicitly created `WorkloadEntry`. If the `WorkloadEntry` is missing, it's a configuration mistake.

So, I would prefer to return an error out of `RegisterWorkload()` in case the `WorkloadEntry` is not found, which will cause Istio Proxy to re-connect:
* if it's a config propagation delay, the issue will eventually resolve transparently to the user
* if it's a human mistake, it will be visible to the user (as proxy will never come up)

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -243,54 +244,100 @@ func (c *Controller) processNextWorkItem() bool {
...
 	}
-	// check if the WE already exists, update the status
-	entryName := autoregisteredWorkloadEntryName(proxy)
+	var entryName string
+	var autoCreate bool
+	if features.WorkloadEntryAutoRegistration &&
+		proxy.Metadata.AutoRegisterGroup != "" {
+		entryName = autoregisteredWorkloadEntryName(proxy)
+		autoCreate = true
+	} else if features.WorkloadEntryHealthChecks &&
```
hzxuzhonghu, line 282: It deserves a comment here

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -243,54 +244,100 @@ func (c *Controller) processNextWorkItem() bool {
...
 
-	if err := c.registerWorkload(entryName, proxy, conTime); err != nil {
+	err := c.onWorkloadConnect(entryName, proxy, conTime, autoCreate)
+	if err != nil {
 		log.Errorf(err)
+	}
+	return err
+}
+
+// onWorkloadConnect updates WorkloadEntry of the connecting workload.
```
hzxuzhonghu, line 303: creates/updates

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -342,56 +469,34 @@ func (c *Controller) QueueUnregisterWorkload(proxy *model.Proxy, origConnect tim
...
-		// TODO(@hzxuzhonghu): update the Get interface, fallback to calling apiserver.
-		return fmt.Errorf("workloadentry %s/%s is not found, maybe deleted or because of propagate latency", proxy.Metadata.Namespace, entryName)
-	}
-
-	// only queue a delete if this disconnect event is associated with the last connect event written to the worload entry
-	if mostRecentConn, err := time.Parse(timeFormat, cfg.Annotations[ConnectedAtAnnotation]); err == nil {
-		if mostRecentConn.After(origConnTime) {
-			// this disconnect event wasn't processed until after we successfully reconnected
-			return nil
+func (c *Controller) unregisterWorkload(entryName string, proxy *model.Proxy, disconTime, origConnTime time.Time, autoCreated bool) error {
```
hzxuzhonghu, line 484: why does it need to know autoCreated or not

pilot/pkg/controller/workloadentry/workloadentry_controller.go
```diff
@@ -342,56 +469,34 @@ func (c *Controller) QueueUnregisterWorkload(proxy *model.Proxy, origConnect tim
...
+	changed, err := c.changeWorkloadEntryStateToDisconnected(entryName, proxy, disconTime, origConnTime)
+	if err != nil {
+		if autoCreated {
+			autoRegistrationErrors.Increment()
```
hzxuzhonghu, line 488: If it is for this, i feel we can remove it.

yskopets: removed check for `autoCreated` here

but still checking in

```go
	if proxy.WorkloadEntryAutoCreated {
		autoRegistrationUnregistrations.Increment()
	}
```
to avoid confusing metrics

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: removed and checking `proxy.WorkloadEntryAutoCreated` instead

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: added a comment

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: @hzxuzhonghu Instead of trying to infer the name of the corresponding `WorkloadEntry` from `proxy.ID`, I would prefer to introduce a new field into the node metadata specifically for that purpose.

E.g.,

```go
type NodeMetadata struct {
...
	// WorkloadEntry is the name of the `WorkloadEntry` this proxy corresponds to.
	WorkloadEntry string `json:"WORKLOAD_ENTRY,omitempty"`
...
}
```

This setting would be set for a sidecar through environment variable `ISTIO_META_WORKLOAD_ENTRY`.

WDYT?

yskopets:
@zirain 

> IMHO, we should not put `readinessProbe` configuration in annotations, prefer to extend `WorkloadEntry` definition

I don't know what was the motivation not to include `readinessProbe` into `WorkloadEntry`.

I've noticed, though, that `WorkloadEntry` can be used not only as a standalone CRD but also inlined into `ServiceEntry`.

If `readinessProbe` field was a part of `WorkloadEntry`, it would be unenforceable in the context of `ServiceEntry`. Maybe that's the reason.

howardjohn:
> @zirain
> 
> > IMHO, we should not put `readinessProbe` configuration in annotations, prefer to extend `WorkloadEntry` definition
> 
> I don't know what was the motivation not to include `readinessProbe` into `WorkloadEntry`.
> 
> I've noticed, though, that `WorkloadEntry` can be used not only as a standalone CRD but also inlined into `ServiceEntry`.
> 
> If `readinessProbe` field was a part of `WorkloadEntry`, it would be unenforceable in the context of `ServiceEntry`. Maybe that's the reason.

ReadinessProbe is already in https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#ProxyConfig, not sure there is a clear reason to put it in WorkloadEntry?

pilot/pkg/controller/workloadentry/workloadentry_controller.go
hzxuzhonghu: Sounds good, but this should be more generic, I remember we have a test for vm use some flag @stevenctl may know.
 grpc now use a different generator, not sure some other proxies in the future need this.

pilot/pkg/controller/workloadentry/workloadentry_controller.go
howardjohn: What confused me is why `WorkloadEntry` has a `proxy.istio.io/config` annotation with the readiness probe in it... How does that actually end up on the VM or get used? My understanding is the VM knows nothing about the WorkloadEntry?

How I would expect the code:
```
if isVM() {
  if _,f := annotations[WorkloadEntryHealthCheckAnnotation]; !f { return // no health checks, no need to do anything }
  entryName = ...
}
```

For IsVM, I think the change you mentioned to explicitly declare it in NodeMetadata below is likely ideal?

howardjohn (commented):
Mostly LGTM, just need to get consensus on how to map proxy metadata <-> Workload Entry

cc @stevenctl

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: @howardjohn Hey

> What confused me is why `WorkloadEntry` has a `proxy.istio.io/config` annotation with the readiness probe in it...

You're right that Istio doesn't provide a command anymore to generate configuration for the Istio Sidecar based on a `WorkloadEntry` rather than on a `WorkloadGroup`.

Yeah, I can remove `proxy.istio.io/config` from the example to avoid confusion.

The workflow that is possible right now:
* a user creates a `WorkloadGroup`
* then uses `istioctl x workload entry` command to generate configuration for the sidecar
* then manually creates  a `WorkloadEntry` to represent a single VM

> My understanding is the VM knows nothing about the `WorkloadEntry`?

Correct.

At the moment, the only link possible is if a user will set environment variable `POD_NAME` to the name of `WorkloadEntry` he/she created manually.

>   if _,f := annotations[WorkloadEntryHealthCheckAnnotation]; !f { return // no health checks, no need to do anything }

Ok, I will rework to check for `annotations[WorkloadEntryHealthCheckAnnotation]`. 

> For IsVM, I think the change you mentioned to explicitly declare it in NodeMetadata below is likely ideal?

Since I've been focused on the VM use case, I proposed to name this field `WORKLOAD_ENTRY` and use it only for `WorkloadEntry`(s).

But maybe you prefer a neutral field that could be used for either `Pod` name or `WorkloadEntry` name?

On the downside, it won't be enough to distinguish a VM.

howardjohn: Use WORKLOAD_ENTRY, SGTM

yskopets: added `WORKLOAD_ENTRY` and  started checking `WorkloadEntryHealthCheckAnnotation`.

@howardjohn PTAL

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: this function has been removed

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: this function has been removed

pilot/pkg/controller/workloadentry/workloadentry_controller.go
yskopets: added `WORKLOAD_ENTRY` to proxy metadata and return error if there is no such WorkloadEntry

@hzxuzhonghu PTAL

hzxuzhonghu: Will take another look, sorry for the delay

hzxuzhonghu (approved):

PR #37102, opened 2022-02-01
fix flake: queue.TestClosed

### LINKED ISSUE #37085: TestClosed flakes
https://github.com/istio/istio/blob/4dea7d99eba0ab42e84056d50e1a96151f301887/pkg/queue/instance_test.go#L157

This job fails due to not always processing the task. I think this is a test that doesn't match the implementation. The impl closes immediately once close is called; the test seems to be testing that we have processed all finished tasks before we close.

stevenctl:
fixes #37085 

The way this is written now better matches the impl. 
Older impls would fail this (running 1 more task after close, or trying to drain the queue). 

This should be good enough, and at least it isn't flaky.

### DIFF
--- a/pkg/queue/instance_test.go
+++ b/pkg/queue/instance_test.go
@@ -162,11 +162,14 @@ func TestClosed(t *testing.T) {
 			close(stop)
 			return nil
 		})
+		go q.Run(stop)
+		if err := WaitForClose(q, 10*time.Second); err != nil {
+			t.Error()
+		}
 		q.Push(func() error {
 			taskComplete.Store(true)
 			return nil
 		})
-		go q.Run(stop)
 		if taskComplete.Load() {
 			t.Error("task ran on closed queue")
 		}
--- a/pkg/queue/instance_test.go
+++ b/pkg/queue/instance_test.go
@@ -154,21 +154,21 @@ func TestClosed(t *testing.T) {
 			t.Error(err)
 		}
 	})
-	t.Run("close after tasks", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
This is still flaky. When stop is closed its handled asyncronously

stevenctl:
made us wait then push and ensure nothing else runs.. not sure how valuable it is at this point though. 

I could push 2 tasks, 1 that takes ~1 second and another that marks `taskComplete`. It's okay if the one task gets complete? That sounds likely to have flakes. Not really sure how to test this effectively.

pkg/queue/instance_test.go
```diff
@@ -154,21 +154,24 @@ func TestClosed(t *testing.T) {
...
 		q.Push(func() error {
-			taskComplete.Store(true)
+			close(stop)
 			return nil
 		})
-		go q.Run(stop)
-		close(stop)
 		if err := WaitForClose(q, 10*time.Second); err != nil {
-			t.Error(err)
+			t.Error()
```
howardjohn, line 166:
```suggestion
			t.Fatal(err)
```

pkg/queue/instance_test.go
```diff
@@ -154,21 +154,24 @@ func TestClosed(t *testing.T) {
...
 		}
-		if !taskComplete.Load() {
-			t.Error("task did not complete")
+		q.Push(func() error {
+			taskComplete.Store(true)
+			return nil
+		})
+		go q.Run(stop)
```
howardjohn, line 172: Move to L164?

howardjohn (approved):

PR #37111, opened 2022-02-01
Break instead of return on empty resources at xds proxy.

bianpengyuan:
Otherwise xds proxy will just stop if for some reason control plane is passing empty resource, which is catastrophic.


howardjohn (approved):
Nice find - cherrypick it?

PR #37116, opened 2022-02-02
Various test framework improvements

howardjohn:
* fix flake: queue.TestClosed

Change-Id: Ibddc488302badd9ef2c8568640e9bdff68ad99f5

* wait for close then push

Change-Id: I4cda8cefe325622a26e18920e8b2fb8fd4514bc8

### DIFF
--- a/pkg/test/framework/components/echo/kube/sidecar.go
+++ b/pkg/test/framework/components/echo/kube/sidecar.go
@@ -47,14 +47,14 @@ type sidecar struct {
 	cluster      cluster.Cluster
 }
 
-func newSidecar(pod kubeCore.Pod, cluster cluster.Cluster) (*sidecar, error) {
+func newSidecar(pod kubeCore.Pod, cluster cluster.Cluster) *sidecar {
 	sidecar := &sidecar{
 		podNamespace: pod.Namespace,
 		podName:      pod.Name,
 		cluster:      cluster,
 	}
 
-	return sidecar, nil
+	return sidecar
 }
 
 func (s *sidecar) Info() (*envoyAdmin.ServerInfo, error) {
--- a/pkg/test/framework/components/echo/kube/sidecar.go
+++ b/pkg/test/framework/components/echo/kube/sidecar.go
@@ -16,7 +16,6 @@ package kube
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 
@@ -43,7 +42,6 @@ const (
 var _ echo.Sidecar = &sidecar{}
 
 type sidecar struct {
-	nodeID       string
 	podNamespace string
 	podName      string
 	cluster      cluster.Cluster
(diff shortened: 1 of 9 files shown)


howardjohn:
TestIngressRequestAuthentication times out in CI but not locally. add more logs

howardjohn:
Current theory is event is sent on distributionEventQueue but not fully processed (in time? or never.. not clear).

Out of order is also possible but unlikely. Actually, maybe we do this:
* Send v1
* Send v2 (skip)
* Real ACK v1 (override ^)

pkg/test/framework/components/echo/kube/sidecar.go
```diff
@@ -56,31 +54,9 @@ func newSidecar(pod kubeCore.Pod, cluster cluster.Cluster) (*sidecar, error) {
 		cluster:      cluster,
 	}
 
-	// Extract the node ID from Envoy.
-	if err := sidecar.WaitForConfig(func(cfg *envoyAdmin.ConfigDump) (bool, error) {
```
howardjohn, line 60: This is only to get node ID, which is only used to get pod name, which we already have. This adds >1s to each run -- this adds up because re-running locally only takes like 3s total so its 33% overhead.

pkg/test/framework/components/istioctl/kube.go
```diff
@@ -60,8 +61,10 @@ func (c *kubeComponent) WaitForConfigs(defaultNamespace string, configs string)
 		if ns == "" {
 			ns = defaultNamespace
 		}
-		if _, _, err := c.Invoke([]string{"x", "wait", cfg.GroupVersionKind.Kind, cfg.Name + "." + ns}); err != nil {
-			return err
+		// TODO(https://github.com/istio/istio/issues/37148) increase timeout. Right now it fails often, so
+		// set it to low timeout to reduce impact
+		if out, stderr, err := c.Invoke([]string{"x", "wait", "-v", "--timeout=5s", cfg.GroupVersionKind.Kind, cfg.Name + "." + ns}); err != nil {
```
howardjohn, line 66: getting more logging and decrease timeout to avoid wasted time

pkg/test/framework/testcontext.go
```diff
@@ -215,8 +214,7 @@ func (c *testContext) AllClusters() cluster.Clusters {
 }
 
 func (c *testContext) CreateDirectory(name string) (string, error) {
-	dir := filepath.Join(c.workDir, name)
-	err := os.Mkdir(dir, os.ModePerm)
+	dir, err := os.MkdirTemp(c.workDir, name)
```
howardjohn, line 217: avoid issue here we run test twice and use the same thing. code copied from suite context

tests/integration/security/jwt_test.go
```diff
@@ -437,38 +437,38 @@ func TestIngressRequestAuthentication(t *testing.T) {
...
-							return nil
-						}).
-						From(util.SourceFilter(t, apps, ns.Name(), false)...).
-						ConditionallyTo(echotest.ReachableDestinations).
-						ConditionallyTo(func(from echo.Instance, to echo.Instances) echo.Instances {
-							return to.Match(echo.InCluster(from.Config().Cluster))
-						}).
-						To(util.DestFilter(t, apps, ns.Name(), false)...).
-						Run(func(t framework.TestContext, src echo.Instance, dest echo.Instances) {
+				echotest.New(t, apps.All).
```
howardjohn, line 440: clean up test naming

nmittler (approved):

lei-tang (approved):

stevenctl:
/retest

This is a separate flake, not affected by your changes.

PR #37129, opened 2022-02-02
vm support: support multiple WorkloadEntry(s) with the same IP

yskopets:

This PR adds support for the use case with multiple `WorkloadEntry`(s) with the same IP address.

Apparently, some users would like to run multiple Istio Proxies on the same VM (e.g., multiple `sidecars` without Iptables or multiple `routers`).

This PR changes behaviour in case there are multiple `WorkloadEntry`(s) with the same IP.

Previously, "GetProxyServiceInstances()" would only consider one of the `WorkloadEntry`(s) with the same IP (AFAIU, the choice was not stable/predictable).

After this PR, "GetProxyServiceInstances()" will consider all `WorkloadEntry`(s) and favour the one that matches `proxy.ID`. If there is no `WorkloadEntry` that matches `proxy.ID`, the `WorkloadEntry`(s) are sorted by name and the first one is chosen.


hzxuzhonghu (commented):
Need to do the same for service entry registry

hzxuzhonghu:
>Need to do the same for service entry registry

Can you split this refactor to another pr? It seems this change is huge and maybe risky.  I got a review yesterday, it looks good.

yskopets:
@hzxuzhonghu Sure, I will split `serviceregistry/serviceentry` changes into a separate PR.

I'm still trying to understand all the source code of `serviceregistry/serviceentry`.

pilot/pkg/serviceregistry/kube/controller/controller_test.go
```diff
@@ -847,18 +849,62 @@ func TestGetProxyServiceInstances_WorkloadInstance(t *testing.T) {
...
+
 	createService(ctl, "reviews", "bookinfo-reviews",
 		map[string]string{
 			annotation.AlphaKubernetesServiceAccounts.Name: "reviews",
 			annotation.AlphaCanonicalServiceAccounts.Name:  "reviews@gserviceaccount2.com",
 		},
 		[]int32{7070}, map[string]string{"app": "reviews"}, t)
 
-	ev := fx.Wait("service")
+	ev = fx.Wait("service")
```
howardjohn, line 883: very nit - may be easier to just pass in `t` to Wait if we are always just going to fail the same way

pilot/pkg/serviceregistry/util/workloadinstances/util.go
```diff
@@ -0,0 +1,87 @@
...
+	if len(parts) == 2 && proxy.ConfigNamespace == parts[1] {
+		return types.NamespacedName{Name: parts[0], Namespace: parts[1]}
+	}
+	return types.NamespacedName{}
+}
+
+// GetInstanceForProxy returns a workload instance that
+// corresponds to a given proxy, if any.
+func GetInstanceForProxy(index Index, proxy *model.Proxy, proxyIP string) *model.WorkloadInstance {
+	if !sets.NewSet(proxy.IPAddresses...).Contains(proxyIP) {
```
howardjohn, line 49: nit: might be faster after L56 to avoid so many map creations

pilot/pkg/serviceregistry/util/workloadinstances/util.go
```diff
@@ -0,0 +1,87 @@
...
+		return wi.Namespace == proxy.ConfigNamespace
+	})
+	if instance != nil {
+		return instance
+	}
+
+	// fall back to choosing one of the workload instances
+
+	// NOTE: for the sake of backwards compatibility, we don't enforce
+	//       instance.Namespace == proxy.ConfigNamespace
```
howardjohn, line 85: Isn't this net-new  since backwards compat case would be handled by len=0 or len=1 which returns early?

hzxuzhonghu: good catch

pilot/pkg/serviceregistry/kube/controller/controller_test.go
yskopets: done

pilot/pkg/serviceregistry/util/workloadinstances/util.go
yskopets: replaced `sets.NewSet()` with `slices.ContainsString()`

pilot/pkg/serviceregistry/util/workloadinstances/util.go
yskopets: Let's assume in the past a user had 2 `WorkloadEntry`(s) with the same IP and in the wrong namespace(s).

Previously, it worked because index kept at most 1 `WorkloadEntry` per IP and namespace wasn't checked.

To make sure it will still work, now that the index keeps all `WorkloadEntry`(s) per IP, we need to pick one of the `WorkloadEntry`(s) and still don't check the namespace.

Issue #36621, opened 2021-12-23
Multiple VirtualServices for `Mesh`  and ingress gateway behaves differently 


hzxuzhonghu:
If there are two virtualServices: A and B, A is created before B. Both A and B is defing routes fro same host.

If the gateway is ingress gateway, both of them are reflected in gateway's route config.

But if the gateway is Mesh,  only A is effecting. B is ignorred.

It looks we donot document this difference anywhere.

howardjohn:
There is no merging for mesh. Maybe we need more clear docs though

hzxuzhonghu:
Personally, i tend more to make them consistent. Users can be super confused about this.

ramaraochavali:
I think there is an issue open for mesh VS merge https://github.com/istio/istio/issues/22997

Issue #22997, opened 2020-04-16
Merging of split VirtualServices using the mesh gateway (sidecars)


erSitzt:
Hi,

we deploy lots of different version of the same services and i have to split up the deployments into multiple VirtualServices and DestinationRules using the same host but routing http traffic via the url prefix.

like so:

- svc-a.domain.com/v1.0
- svc-a.domain.com/v1.1
- svc-a.domain.com/v1.2

All services use an istio gateway and the "_mesh_" gateway, because the services are talking to each other a lot.
As merging does not work in the sidecars only one of the three entries above will work as long as i keep them attached to the "_mesh_"-gateway.

If i remove the "_mesh_"-gateway from the VirtualServices everything works as planned but all traffic goes through my ingress-gateway and the "Visibility" benefit of the Service Mesh especially in Kiali is gone because i cant see svc-a talking to svc-b directly

I didnt find any clues or get any answers why host merging works in gateways but not the sidecar, but i think this is a feature a lot of people will need/use
I read a comment somewhere here about preventing unpredictable behavior which can result from merging, but i think this should be up to the user. These are configs we used with haproxy for years, they are only unpredictable if you f##k up, but that goes for every config ;)

Any feedback on this is welcome

rshriram:
Hi.. virtual service merging is not implemented for the `mesh` gateways as of now. So, unfortunately to get the behavior you desire, you need to have one large virtual service/destination rule. We can add this to the feature request list and see if we can knock this off in the next release. cc @nrjpoddar to track.

erSitzt:
thanks for the info @rshriram this would be great.
Creating one large VirtualService/destinationRule in our deployments is really hard because we are running up to 10 versions of the same service. because we have to support older app version for a long time here.

robert-blackman:
> Hi.. virtual service merging is not implemented for the `mesh` gateways as of now. So, unfortunately to get the behavior you desire, you need to have one large virtual service/destination rule. We can add this to the feature request list and see if we can knock this off in the next release. cc @nrjpoddar to track.

Hey @rshriram any movement on this?

erSitzt:
Is there any more info on why VirtualService merging was/is done differently in gateways and sidecars ?

luisdalves:
Hey Guys,

I'm stuck with the same issue. Is there any advance on this or maybe a workaround?

Thanks

erSitzt:
Please dont close this issue.

Leo7654:
[...]

PR #37139, opened 2022-02-03
Correct route precedence logic for gateway-api

### LINKED ISSUE #36719: gateway-api: correct route precedence logic
See https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPRouteRule

We match in ordered list, they have more defined rules.

We should also be merging `mesh` routes, IMO, using the same logic in these cases. We can collapse multiple HTTPRoute down into 1 VS that has a big ordered list (if they have the same hostname).

dddddai:
Fixes #36719


pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -165,17 +165,31 @@ func convertVirtualService(r *KubernetesResources, gatewayMap map[parentKey]map[
...
 	}
 	return result
 }
 
-func buildHTTPVirtualServices(obj config.Config, gateways map[parentKey]map[k8s.SectionName]*parentInfo, domain string) *config.Config {
-	route := obj.Spec.(*k8s.HTTPRouteSpec)
-
+func buildHTTPVirtualServices(obj config.Config, gateways map[parentKey]map[k8s.SectionName]*parentInfo,
+	domain string, vsByHostAndGateway map[string]map[string]*config.Config) {
+	route := obj.Spec.(*k8s.HTTPRouteSpec).DeepCopy()
```
howardjohn, line 192: maybe only deep copy if matches >1?

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -190,31 +204,43 @@ func buildHTTPVirtualServices(obj config.Config, gateways map[parentKey]map[k8s.
 
 	httproutes := []*istio.HTTPRoute{}
 	hosts := hostnameToStringList(route.Hostnames)
-	for _, r := range route.Rules {
+	for i := 0; i < len(route.Rules); i++ {
```
howardjohn, line 207:
```suggestion
	for i := range route.Rules {
```

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -190,31 +204,43 @@ func buildHTTPVirtualServices(obj config.Config, gateways map[parentKey]map[k8s.
...
 			}
 			method, err := createMethodMatch(match)
 			if err != nil {
 				reportError(err)
-				return nil
+				return
 			}
 			vs.Match = append(vs.Match, &istio.HTTPMatchRequest{
+				// the name is a hack to fetch the original path type
+				Name:        generateMatchName(obj, i, *match.Path.Type),
```
howardjohn, line 243: why do we need Type? isn't index sufficient?

pilot/pkg/config/kube/gateway/testdata/route-precedence.yaml
```diff
@@ -0,0 +1,100 @@
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
howardjohn, line 1: Cases to think about is merging routes that are across namespace. Especially with `mesh`

howardjohn: Yeah this is problematic..

```yaml
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: HTTPRoute
metadata:
  name: echo
  namespace: default
spec:
  parentRefs:
  - kind: Mesh
    name: istio
  hostnames: ["echo.default.svc.cluster.local"]
  rules:
  - backendRefs:
    - name: echo
      port: 80
---
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: HTTPRoute
metadata:
  name: echo
  namespace: consumer
spec:
  parentRefs:
  - kind: Mesh
    name: istio
  hostnames: ["echo.default.svc.cluster.local"]
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /get
    backendRefs:
    - name: echo
      port: 80
```

Ends up getting merged like:
```yaml
kind: VirtualService
metadata:
  annotations:
    internal.istio.io/parent: HTTPRoute/echo.default
  name: echo-istio-autogenerated-k8s-gateway-0
  namespace: default
spec:
  gateways:
  - mesh
  hosts:
  - echo.default.svc.cluster.local
  http:
  - match:
    - name: echo-0-PathPrefix
      uri:
        regex: /get((\/).*)?
    route:
    - destination:
        host: echo.consumer.svc.cluster.local
        port:
          number: 80
  - match:
    - name: echo-0-PathPrefix
      uri:
        prefix: /
    route:
    - destination:
        host: echo.default.svc.cluster.local
        port:
          number: 80
```

howardjohn (commented):
Nice this is a really good start!

pilot/pkg/config/kube/gateway/conversion.go
dddddai: Because we are using `regex` to implement both `prefix` and `regex`, if we don't have the type, we would not be able to determine whether to trim the regex suffix for `prefix` when calculating the uri length

For example:
https://github.com/istio/istio/pull/37139/files/8cb067ba9c70430f11a7ef942b4cf9e19a640915#diff-13176e82f5f2350940ea3ea83d0bfeb3147c172af5808771f6006be8393e1632R33-R45

The index is to make match names not duplicated

howardjohn: ah I didn't realize we were later parsing it in getOriginalPathType. Hmm.. it seems more ideal if we can have it somewhere else instead of parsing but that may not be possible

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -165,17 +165,31 @@ func convertVirtualService(r *KubernetesResources, gatewayMap map[parentKey]map[
 		}
 	}
 
+	// sort HTTPRoutes by creation timestamp and namespace/name
```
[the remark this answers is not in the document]
dddddai: `sortConfigByCreationTime` uses `name.namespace` but here I'm using `namespace/name`, I'm not sure if changing the order would break something, if it's ok I will change it

pilot/pkg/config/kube/gateway/conversion.go
dddddai: This is intentional because `range` can not reach the index appended in the loop
For example:
```go
for i := range arr {
	if i == 0 {
		arr = append(arr, 3)
	}
	println(arr[i])
}
```
It will not print the appended "3", but it will if using `i < len(arr)`

pilot/pkg/config/kube/gateway/testdata/route-precedence.yaml
dddddai: Sorry I didn't get the problem, could you please explain the right behavior? Thanks!

howardjohn: I am not quite sure - but not this I think. I'll need to think about it. I am thinking we shouldn't merge across namespace for mesh but will need to think about all cases

dddddai: Seems related to #22997 and #36621

howardjohn: Ok sorry for the delay. I had a complete misunderstand of how VirtualService even works! so I think we may need some broader changes. Will investigate a bit more and get back to you

howardjohn: Need a bit more time. sorry

howardjohn: Apologies for the delay - can't believe this was 5 weeks! I was thinking about this the entire time believe it or not, its a lot trickier problem than I originally anticipated. I am planning to write up a full design doc on how I think Mesh routes should work, but I think for now it would be safe to make it so we just do not merge across namespaces. I think we will need more changes than that, but that is out of scope for this PR.

BTW, for Gateway case, can we add a test like:

```
Gateway:
  allowedRoutes: [ns1, ns2]
Route in ns1
Route in ns2
Route in ns3
```

Result should be ns1+ns2 merge, and ns3 is properly ignore

dddddai: No problem :)
Fixed, thanks for elaborating

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -263,36 +304,144 @@ func buildHTTPVirtualServices(obj config.Config, gateways map[parentKey]map[k8s.
...
+					},
+					Spec: &istio.VirtualService{
+						Hosts:    []string{host},
+						Gateways: []string{gw},
+						Http:     httproutes,
+					},
+				}
+				count++
+			}
+		}
```
howardjohn, line 388:
```suggestion
for _, gw := range gatewayNames {
		routeMap := gatewayRoutes
		routeKey := gw
		if gw == "mesh" {
			routeMap = meshRoutes
			routeKey = ns
		}

		if _, f := routeMap[routeKey]; !f {
			routeMap[routeKey] = make(map[string]*config.Config)
		}
		for _, host := range hosts {
			if cfg := routeMap[routeKey][host]; cfg != nil {
				// merge http routes
				vs := cfg.Spec.(*istio.VirtualService)
				vs.Http = append(vs.Http, httproutes...)
				sortHTTPRoutes(vs.Http)
			} else {
				sortHTTPRoutes(httproutes)
				// build one VS per namespace+host
				routeMap[routeKey][host] = &config.Config{
					Meta: config.Meta{
						CreationTimestamp: obj.CreationTimestamp,
						GroupVersionKind:  gvk.VirtualService,
						Name:              fmt.Sprintf("%s-%d", name, count),
						Annotations:       parentMeta(obj, nil),
						Namespace:         ns,
						Domain:            domain,
					},
					Spec: &istio.VirtualService{
						Hosts:    []string{host},
						Gateways: []string{gw},
						Http:     httproutes,
					},
				}
				count++
			}
		}
```

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -263,36 +304,144 @@ func buildHTTPVirtualServices(obj config.Config, gateways map[parentKey]map[k8s.
...
+		// it's a gateway route
+		if _, f := gatewayRoutes[gw]; !f {
+			gatewayRoutes[gw] = make(map[string]*config.Config)
+		}
+		for _, host := range hosts {
+			if cfg := gatewayRoutes[gw][host]; cfg != nil {
+				// merge http routes
+				vs := cfg.Spec.(*istio.VirtualService)
+				vs.Http = append(vs.Http, httproutes...)
+				sortHTTPRoutes(vs.Http)
```
howardjohn, line 367: sortHTTPRoutes is n^2 because we call it each time we append. Maybe we can just call it once on all routes after?

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -263,36 +304,144 @@ func buildHTTPVirtualServices(obj config.Config, gateways map[parentKey]map[k8s.
...
+}
+
+func getOriginalPathType(match *istio.HTTPMatchRequest) k8s.PathMatchType {
+	i := strings.LastIndexByte(match.Name, '-')
+	return k8s.PathMatchType(match.Name[i+1:])
+}
+
+// sortHTTPRoutes sort generated vs routes to meet gateway-api requirements
+// see https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPRouteRule
+func sortHTTPRoutes(routes []*istio.HTTPRoute) {
```
howardjohn, line 407: I think we may be missing the fallback part:

```
If ties still exist across multiple Routes, matching precedence MUST be determined in order of the following criteria, continuing on ties:

The oldest Route based on creation timestamp.
The Route appearing first in alphabetical order by “{namespace}/{name}”.
```

Also importantly is the CreationTimestamp of the generate VirtualService, as that is used in a number of places in the later XDS config generation - right now I guess its just the timestamp of the first one?

dddddai: > I think we may be missing the fallback part:

The routes [have already been sorted based on "{namespace}/{name}"](https://github.com/istio/istio/commit/1622e7f68148d90560c9011744177bb69b55f4b1#diff-a0e8831b6aefb0ef9b2cd269fcd726b26fd1104120950952b5217a7b104ba153R84-R92) before they are converted to VS

> Also importantly is the CreationTimestamp of the generate VirtualService, as that is used in a number of places in the later XDS config generation - right now I guess its just the timestamp of the first one?

Yes it's the first one

howardjohn: First one so the oldest right? If so that make sense to me

dddddai: Yeah

howardjohn (commented):
Sorry didn't get a chance to review this today. @ruigulala is trying this out and its my priority tomorrow to approve

dddddai:
OK, no rush

Issue #35607, opened 2021-10-14
Wasm API tracking issue


dgn:
Needed for 1.12:

- [x] api: Add WasmPlugin (https://github.com/istio/api/pull/1940)
- [x] Initial WasmPlugin implementation (https://github.com/istio/istio/pull/33374)
- [x] api: make WasmPlugin sha256 field non-optional. see https://github.com/istio/istio/pull/33374#discussion_r717028025 ( https://github.com/istio/api/pull/2125 and https://github.com/istio/istio/pull/35608)
- [x] add positive integration test for WasmPlugin (https://github.com/istio/istio/pull/35668)
- [x] use networkingutil.MessageToAny instead of gogo.MessageToAny in implementation. see https://github.com/istio/istio/pull/33374#discussion_r717026690 (https://github.com/istio/istio/pull/35628)
- [x] add support for ext-authz. current implementation behaves in unexpected ways when ext-authz is used. see https://github.com/istio/istio/pull/33374#discussion_r721807423
- [x] blog post(https://istio.io/latest/blog/2021/wasm-api-alpha/)
- [x] documentation(https://istio.io/latest/docs/tasks/observability/telemetry/)

Future:
- [x] ImagePullSecrets implementation(https://github.com/istio/istio/pull/37161)
- [ ] https://github.com/istio/istio/issues/32002

mathetake:
How about "image pull secrets"? Is that removed from the scope of 1.12? cc @mandarjog @bianpengyuan

dgn:
> If we get it in, that would be great. But it is not blocking .

+1 to that. I'll add it to the list

mathetake:
and signature verification https://github.com/istio/istio/issues/32002 (cc @asraa) should also be tracked here

mathetake:
https://github.com/istio/istio/pull/35683 this should also be merged in 1.12

mathetake:
blog PR: https://github.com/istio/istio.io/pull/10530

mathetake:
I realized that we haven't implemented "ImagePullPolicy" as well in imagefetcher.go https://github.com/istio/api/blob/master/extensions/v1alpha1/wasm.proto#L214

mathetake:
anyone working on imagePullSecrets? (I won't be able to work neither on Istio or Envoy for a next few months or maybe a year, it would be great if someone can work on it)

bianpengyuan:
I am working on it now, will also implement the pull policy.

mathetake:
[...]

PR #37161, opened 2022-02-04
Implement Wasm pull secret.

bianpengyuan:
https://github.com/istio/istio/issues/35607

Implement [WasmPlugin image pull secret ](https://github.com/istio/api/blob/master/extensions/v1alpha1/wasm.proto#L248)by sending WasmPull secret with the ECDS resource push. The extension config will use an env variable at wasm VM config to store the secret, and the env variable will be cleaned up before forwarding to Envoy.


pilot/pkg/xds/ecds.go
```diff
@@ -59,17 +69,169 @@ func (e *EcdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
...
+	for _, wps := range wasmPlugins {
+		for _, wp := range wps {
+			if watched.Contains(wp.ExtensionConfiguration.Name) && wp.ImagePullSecret != "" {
+				referencedSecrets.Insert(wp.ImagePullSecret)
+			}
+		}
+	}
+	var filtered []SecretResource
+	for rn := range referencedSecrets {
+		sr, err := parseSecretName(rn, proxy.Metadata.ClusterID)
```
howardjohn, line 183: So the spec says `Name of a K8s Secret in the same namespace as the WasmPlugin that` but here we expect `kubernetes://secret-namespace/secret-name`. This seems problematic that they can request any namespace

pilot/pkg/xds/ecds.go
```diff
@@ -59,17 +69,169 @@ func (e *EcdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
...
+func getDockerCredential(name, namespace string, secretController credscontroller.Controller) ([]byte, error) {
+	// Check if the secret is of type `kubernetes.io/dockerconfigjson`.
+	t, err := secretController.GetType(name, namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get type for secret %v/%v: %v", namespace, name, err)
+	}
+	if t != v1.SecretTypeDockerConfigJson {
+		return nil, fmt.Errorf("secret %v/%v type %v is not kubernetes.io/dockerconfigjson", namespace, name, t)
+	}
+	return secretController.GetDockerCredential(name, namespace)
```
howardjohn, line 203: seems like this can just return error if its not the right Type

pkg/istio-agent/xds_proxy.go
```diff
@@ -545,7 +545,7 @@ func (p *XdsProxy) handleUpstreamResponse(con *ProxyConnection) {
 }
 
 func (p *XdsProxy) rewriteAndForward(con *ProxyConnection, resp *discovery.DiscoveryResponse) {
-	sendNack := wasm.MaybeConvertWasmExtensionConfig(resp.Resources, p.wasmCache)
+	converedResources, sendNack := wasm.MaybeConvertWasmExtensionConfig(resp.Resources, p.wasmCache)
```
howardjohn, line 548:
```suggestion
	convertedResources, sendNack := wasm.MaybeConvertWasmExtensionConfig(resp.Resources, p.wasmCache)
```

pkg/wasm/proto/wsds.proto
```diff
@@ -0,0 +1,29 @@
...
+
+option go_package = "istio.io/istio/pkg/dns/proto/istio_networking_wsds_v1";
+
+// WasmSecret is a secret used to pull Wasm image.
+message WasmSecret {
+    // Name of the Wasm pull secret resource.
+    string name = 1;
+
+    // The secret that are used to pull Wasm images.
+    bytes secret = 2;
```
howardjohn, line 27: I am curious the pros/cons compared to:
* Using Envoy standard SDS type
* Just putting it in the ECDS message (metadata?). Doesn't seem like there is any obvious metadata section but maybe there should be...

howardjohn (changes requested):

pkg/wasm/proto/wsds.proto
bianpengyuan: > Using Envoy standard SDS type

Main problem with either SDS or separated resource is complexity and edge cases which is hard to cover. We need to coordinate ECDS update and SDS update (e.g. ECDS arrives at the agent, we read secret it needs, trigger a SDS request, then wait for secret to arrive etc). It will be even more complicated with gateway, since we need to intercept and mutate the request.

> Just putting it in the ECDS message (metadata?). Doesn't seem like there is any obvious metadata section but maybe there should be...

That is my preference, like you noticed, I did not find a good place to put it. We can potentially just use the VM env variables, but that requires unmarshal and marshal again at xds gen time, which would be very inefficient? I can explore more on this route if sending different resources are too hacky.

howardjohn: > Main problem with either SDS or separated resource is complexity and edge cases which is hard to cover. 

Couldn't we still just put it in the same response message, like we are doing here

bianpengyuan: > Couldn't we still just put it in the same response message, like we are doing here

Oh I see what you are saying. I thought you meant to use the existing SDS channel. Yeah, there is no particular reason to have a new type of resource, I can change it to Envoy Secret type.

ramaraochavali:
@bianpengyuan I could not fully follow - but is this implementing support for https://github.com/istio/api/blob/master/extensions/v1alpha1/wasm.proto#L248?

pilot/pkg/xds/ecds.go
bianpengyuan: If you look at `toSecretResourceName` implementation, it update the secret resource name to make sure it uses plugin's namespace.

pilot/pkg/xds/ecds.go
```diff
@@ -59,17 +69,169 @@ func (e *EcdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
...
+
+		cachedItem, f := e.secretCache.Get(sr)
+		if f && !features.EnableUnsafeAssertions {
+			// If it is in the Cache, add it and continue
+			// We skip cache if assertions are enabled, so that the cache will assert our eviction logic is correct
+			results = append(results, cachedItem)
+			continue
+		}
+
+		cred, err := getDockerCredential(sr.Name, sr.Namespace, secretController)
```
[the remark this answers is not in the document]
bianpengyuan: Do you mean validate type of secret? or some other validations?

pilot/pkg/xds/ecds.go
```diff
@@ -59,17 +69,169 @@ func (e *EcdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
...
+		secrets = e.GeneratePullSecrets(proxy, updatedSecrets, secretResources, secretController, req)
+	}
+	resources := make(model.Resources, 0, len(ec)+len(secrets))
 	for _, c := range ec {
 		resources = append(resources, &discovery.Resource{
 			Name:     c.Name,
 			Resource: util.MessageToAny(c),
 		})
 	}
+	resources = append(resources, secrets...)
```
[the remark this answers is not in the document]
bianpengyuan: Updated to only use ExtensionConfig. Now when constructing push context, WasmConfig is built but not marshaled to any. Later at xds gen time, we will add the secret to VM env and marshal to any.

bianpengyuan:
> @bianpengyuan I could not fully follow - but is this implementing support for https://github.com/istio/api/blob/master/extensions/v1alpha1/wasm.proto#L248?

Yep, that is what this PR for. Updated PR description to make it more clear.

pilot/pkg/xds/ecds.go
```diff
@@ -71,5 +117,127 @@ func (e *EcdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
...
+			res := toWasmSecret(sr.ResourceName, cred)
+			e.secretCache.Add(sr, req, res)
+		}
+	}
+	return results
+}
+
+func (e *EcdsGenerator) SetCredController(creds credscontroller.MulticlusterController, cache model.XdsCache) {
+	e.secretController = creds
+	e.secretCache = cache
```
hzxuzhonghu, line 164: Duplicate, EcdsGenerator.Server.Cache already has it

pilot/pkg/credentials/kube/secrets.go
```diff
@@ -218,6 +219,20 @@ func (s *CredentialsController) GetCaCert(name, namespace string) (cert []byte,
 	return extractRoot(k8sSecret)
 }
 
+func (s *CredentialsController) GetDockerCredential(name, namespace string) ([]byte, error) {
+	k8sSecret, err := s.secretLister.Secrets(namespace).Get(name)
+	if err != nil {
+		return nil, fmt.Errorf("secret %v/%v not found", namespace, name)
+	}
+	if k8sSecret.Type != v1.SecretTypeDockerConfigJson {
```
hzxuzhonghu, line 227: Though from the undeylying implement k8sSecret can not be nil, it is better to check

pilot/pkg/xds/ecds.go
```diff
@@ -71,5 +117,127 @@ func (e *EcdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
 			Resource: util.MessageToAny(c),
 		})
 	}
+
 	return resources, model.DefaultXdsLogDetails, nil
 }
+
+func (e *EcdsGenerator) GeneratePullSecrets(proxy *model.Proxy, updatedSecrets map[model.ConfigKey]struct{}, secretResources []SecretResource,
```
hzxuzhonghu, line 124: The difference from sds is that there is no authorize here, isn's there a risk of secret leak

pilot/pkg/xds/ecds.go
```diff
@@ -71,5 +117,127 @@ func (e *EcdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
...
+		if updatedSecrets != nil {
+			if _, found := updatedSecrets[model.ConfigKey{Kind: gvk.Secret, Name: sr.Name, Namespace: sr.Namespace}]; !found {
+				// This is an incremental update, filter out credscontroller that are not updated.
+				continue
+			}
+		}
+
+		if cachedItem, f := e.secretCache.Get(sr); f {
+			// If it is in the Cache, add it and continue
+			sb, err := toSecretBytes(cachedItem)
```
hzxuzhonghu, line 142: It seems the cache does not help for performance, and it adds conversion

pkg/wasm/imagefetcher.go
```diff
@@ -40,15 +43,20 @@ import (
 var errWasmOCIImageDigestMismatch = errors.New("fetched image's digest does not match the expected one")
 
 type ImageFetcherOption struct {
-	Username string
-	Password string
 	// TODO(mathetake) Add signature verification stuff.
-
-	Insecure bool
+	PullSecret []byte
+	Insecure   bool
```
hzxuzhonghu, line 48: Is there a conflict with PullSecret  and insecure?

bianpengyuan: Not really, insecure decides whether to verified server cert at client, pull secret is used at server side auth.

pilot/pkg/xds/ecds.go
bianpengyuan: I thought it will reduce API list call for secret, although it will be indeed meaningless if credential controller already have a cache. Not sure if that is case?

pilot/pkg/xds/ecds.go
bianpengyuan: WasmPlugin is a bit different from gateway secret, that root namespace WasmPlugin config could apply to all namespaces, so specify that the secret has to be in the same namespace as the WasmPlugin.

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -142,18 +145,41 @@ func toEnvoyHTTPFilter(wasmPlugin *model.WasmPluginWrapper) *hcm_filter.HttpFilt
...
+			if !hasName.Contains(p.ResourceName) {
 				continue
 			}
-			result = append(result, proto.Clone(p.ExtensionConfiguration).(*envoy_config_core_v3.TypedExtensionConfig))
+			wasmExtensionConfig := proto.Clone(p.WasmExtensionConfig).(*extensionsv3.Wasm)
+			// find the pull secret resource name from wasm vm env variables.
+			envs := wasmExtensionConfig.GetConfig().GetVmConfig().GetEnvironmentVariables().GetKeyValues()
+			var secretName string
+			if envs != nil {
+				secretName = envs[model.WasmSecretEnv]
```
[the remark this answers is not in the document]
bianpengyuan: Updated with comments at several places about this.

pilot/pkg/model/extensions.go
```diff
@@ -108,22 +113,36 @@ func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
 				VmConfig: &envoy_extensions_wasm_v3.VmConfig{
 					Runtime: defaultRuntime,
 					Code:    datasource,
+					EnvironmentVariables: &envoy_extensions_wasm_v3.EnvironmentVariables{
+						KeyValues: map[string]string{
```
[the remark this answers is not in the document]
bianpengyuan: Yeah it is not obfuscated, although since agent will remove this before forwarding, so it won't be in config dump and we just need to be careful that the resource is not logged by us at istiod and istio-agent.

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -157,12 +157,14 @@ func InsertedExtensionConfigurations(
 				continue
 			}
 			wasmExtensionConfig := proto.Clone(p.WasmExtensionConfig).(*extensionsv3.Wasm)
-			// find the pull secret resource name from wasm vm env variables.
+			// Find the pull secret resource name from wasm vm env variables.
+			// The Wasm extension config should already have a `ISTIO_META_WASM_IMAGE_PULL_SECRET` env variable
+			// at in the VM env variables, with value being the secret resource name. We try to find the actual
+			// secret, and replace the env variable value with it. When ECDS config update reaches the proxy,
+			// agent will extract out the secret from env variable, use it for image pulling, and strip the
+			// env variable from VM config before forwarding it to env.
```
ramaraochavali, line 165: nit: before forwarding it to envoy

pilot/pkg/xds/ecds.go
ramaraochavali: I think credential store will have informer cache

bianpengyuan: Ok, removed the cache here. thanks!

hzxuzhonghu (approved):

pilot/pkg/xds/ecds.go
```diff
@@ -59,7 +65,44 @@ func (e *EcdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
...
+		if err != nil {
+			log.Warnf("proxy %s is from an unknown cluster, cannot retrieve certificates for Wasm image pull: %v", proxy.ID, err)
+			return nil, model.DefaultXdsLogDetails, nil
+		}
+	}
+	var secrets map[string][]byte
+	if len(secretResources) > 0 {
+		// Inserts Wasm pull secrets in ECDS response, which will be used at xds proxy for image pull.
+		// Before forwarding to Envoy, xds proxy will remove the secret from ECDS response.
+		secrets = e.GeneratePullSecrets(proxy, updatedSecrets, secretResources, secretController, req)
```
howardjohn, line 102: Currently it seems we fail open if we cannot fetch secret. IE we send the extension without a secret. It seems like maybe we should fail closed..? wdyt?

pilot/pkg/xds/ecds.go
```diff
@@ -48,6 +52,8 @@ func ecdsNeedsPush(req *model.PushRequest) bool {
 			return true
 		case gvk.WasmPlugin:
 			return true
+		case gvk.Secret:
```
howardjohn, line 55: comment on L48 is stale. But more importantly I am seeing Secret pushes filtered out entirely at a higher level (ProxyNeedsPush). Because of https://github.com/istio/istio/blob/8ce0defcc905873dadf3fa1d8c3f3629cd39895c/pilot/pkg/xds/proxy_dependencies.go#L26

pkg/wasm/imagefetcher.go
```diff
@@ -311,3 +319,39 @@ func extractOCIArtifactImage(img v1.Image) ([]byte, error) {
...
+
+type wasmKeyChain struct {
+	data []byte
+}
+
+// Resolve an image reference to a credential.
+// The function code is borrowed from https://github.com/google/go-containerregistry/blob/v0.8.0/pkg/authn/keychain.go#L65,
+// by making it take dockerconfigjson directly as bytes instead of reading from files.
+func (k *wasmKeyChain) Resolve(target authn.Resource) (authn.Authenticator, error) {
+	reader := bytes.NewReader(k.data)
```
howardjohn, line 331: This is not safe without https://github.com/docker/cli/pull/3434. Maybe we can add an explicit check for `null` in the mean time?

pilot/pkg/model/extensions.go
```diff
@@ -75,32 +80,48 @@ func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
 	if u.Scheme == "" {
 		u.Scheme = ociScheme
 	}
-
+	// Normalize the image pull secret to the full resource name.
+	wasmPlugin.ImagePullSecret = toSecretResourceName(wasmPlugin.ImagePullSecret, plugin.Namespace)
```
howardjohn, line 84: This is mutating the object that actually lives in the k8s informer which seems problematic. In VS we do similar but make a deepcopy. Worried as-is it may lead to race or something?

pilot/pkg/model/extensions.go
```diff
@@ -75,32 +80,48 @@ func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
...
+		return ""
+	}
+	// If the secret already has type prefix, trim it for now.
+	name = strings.TrimPrefix(name, credentials.KubernetesSecretTypeURI)
+	// If the secret does not have namespace prefix, concat the secret name to be namespace/name.
+	if !strings.HasPrefix(name, pluginNamespace+"/") {
+		name = pluginNamespace + "/" + name
+	}
+	// Finally, add the kubernetes:// secret type.
+	return credentials.KubernetesSecretTypeURI + name
```
howardjohn, line 122: instead of our own logic here, should we utilize some of the parsing in credentials/resource.go? Then it would be `resource.Namespace = pluginnamespace; return resource.String()` which feels a lot more robust.

I am very worried about the propogation of parsing logic for secret resources when any bugs will likely result in vulnerabilities

pilot/pkg/model/extensions.go
hzxuzhonghu: better not mutate underlying store,

pilot/pkg/xds/ecds.go
```diff
@@ -71,5 +114,87 @@ func (e *EcdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
...
+	results := make(map[string][]byte)
+	for _, sr := range secretResources {
+		if updatedSecrets != nil {
+			if _, found := updatedSecrets[model.ConfigKey{Kind: gvk.Secret, Name: sr.Name, Namespace: sr.Namespace}]; !found {
+				// This is an incremental update, filter out credscontroller that are not updated.
+				continue
+			}
+		}
+
+		cred, err := secretController.GetDockerCredential(sr.Name, sr.Namespace)
```
hzxuzhonghu, line 137: does it mean that the proxy can get secret from any namespaces?

zirain: this is hack, `sr.Namespace` actually come form `plugin.Namespace`
```
wasmPlugin.ImagePullSecret = toSecretResourceName(wasmPlugin.ImagePullSecret, plugin.Namespace)
```

zirain: if `wasmplugin` config in `rootNamespace`, all sidecar will pull secret from that namespace, there's a risk of secret leak.

pilot/pkg/xds/ecds.go
```diff
@@ -71,5 +114,87 @@ func (e *EcdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w
...
+			log.Warnf("Failed to parse secret resource name %v: %v", rn, err)
+			continue
+		}
+		filtered = append(filtered, sr)
+	}
+
+	return filtered
+}
+
+func parseSecretName(resourceName string, proxyCluster cluster.ID) (SecretResource, error) {
```
zirain, line 180: `parseSecretName` decode `resourceName` generated by `toSecretResourceName`, can we put them together or add some comment?

pilot/pkg/xds/ecds.go
hzxuzhonghu: This is something similar to sidecar sds, if a user can define DR for a sidecar, and the sidecar have permission to read the cert defined.

pkg/wasm/imagefetcher.go
bianpengyuan: Thanks for the pointer! Added an explicit check. Let me know if I did it right.

pilot/pkg/xds/ecds.go
bianpengyuan: Fixed. I was only testing with Gateway and did not find this :).

pilot/pkg/xds/ecds.go
bianpengyuan: I think it is fine to fail open, which should result in a fetch failure at proxy, which would be more visible to the user.

pilot/pkg/xds/ecds.go
bianpengyuan: Added some comments.

pilot/pkg/xds/ecds.go
bianpengyuan: Yeah, this is a known behavior. We need to make it clear at the reference doc that referencing a secret at WasmPlugin == secret will be distributed to the proxies that it applies to.

hzxuzhonghu: IMO, we should not support mesh global wasm plugin in root namespace.  In the current case, it could not get image pull secret as the `namespace ` may differ.

hzxuzhonghu: @howardjohn you opinion?

bianpengyuan:
@howardjohn @zirain This is ready for review. Please take another look. Thanks!

pilot/pkg/xds/ecds.go
bianpengyuan: Sorry just noticed this comment. Then how do you achieve global WasmPlugin config? Seems really hard to manage if secret needs to be copied into every namespace?

bianpengyuan: Alternative is to let control plane pull and push Wasm bytes, which could take some effort to implement.

If this is a concern, we probably could have an env var at pilot to disable secret distribution from root namespace and only allow it when user explicitly set it.

pilot/pkg/credentials/kube/multicluster.go
```diff
@@ -146,3 +146,19 @@ func (a *AggregateController) Authorize(serviceAccount, namespace string) error
 func (a *AggregateController) AddEventHandler(f func(name string, namespace string)) {
 	// no ops
 }
+
+func (a *AggregateController) GetDockerCredential(name, namespace string) ([]byte, error) {
+	// Search through all clusters, find first non-empty result
```
howardjohn, line 151: I wonder if we should make this a bit stricter and make it only the same cluster as the requesting proxy?

A downside is it is possibly more work to have someone replicate this. The positive is its less permissive, matches K8s image pulling. Also the downside, IMO, is not that bad since they already need to replicate it to each NS

pilot/pkg/model/credentials/resource.go
```diff
@@ -54,6 +54,10 @@ func (sr SecretResource) Key() string {
 	return "sds://" + sr.Type + "/" + sr.Name + "/" + sr.Namespace + "/" + string(sr.Cluster)
 }
 
+func (sr SecretResource) KubernetesResourceName() string {
```
howardjohn, line 57: This seems a bit awkward since `Type` could be something other than `KubernetesSecretType`. This seems like it _could_ lead to type-confusion bugs

pilot/pkg/model/extensions.go
```diff
@@ -36,19 +37,26 @@ const (
 	defaultRuntime = "envoy.wasm.runtime.v8"
 	fileScheme     = "file"
 	ociScheme      = "oci"
+
+	// name of environment variable at Wasm VM, which will carry the Wasm image pull secret.
+	WasmSecretEnv = "ISTIO_META_WASM_IMAGE_PULL_SECRET"
```
howardjohn, line 42: This leaks out through ecdsz to any authenticated user

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
```diff
@@ -142,18 +145,43 @@ func toEnvoyHTTPFilter(wasmPlugin *model.WasmPluginWrapper) *hcm_filter.HttpFilt
...
+			// secret, and replace the env variable value with it. When ECDS config update reaches the proxy,
+			// agent will extract out the secret from env variable, use it for image pulling, and strip the
+			// env variable from VM config before forwarding it to envoy.
+			envs := wasmExtensionConfig.GetConfig().GetVmConfig().GetEnvironmentVariables().GetKeyValues()
+			secretName := envs[model.WasmSecretEnv]
+			if secretName != "" {
+				if sec, found := pullSecrets[secretName]; found {
+					envs[model.WasmSecretEnv] = string(sec)
+				} else {
+					envs[model.WasmSecretEnv] = ""
```
howardjohn, line 172: As far as I can tell, currently this is ALWAYS set. Not sure if this is the exact place that causes it though.

Should we change it so that:

* Set and non-empty: successfully got secret
* Set and empty: want secret, but failed. Currently this will fail in the agent with obscure error (401 instead of "failed to fetch secret). It actually probably shouldn't even try to pull the plugin without auth -- a server could return different response to authenticated vs unauthenticated for example. Or we should do what k8s does
* Not set: didn't want a secret at all

pilot/pkg/credentials/kube/multicluster.go
howardjohn: Actually nvm, noticed we actually allow pulling root credentials directly

howardjohn: And looks like we do call `secretController.ForCluster(proxy)` already

howardjohn (approved):

pilot/pkg/model/credentials/resource.go
bianpengyuan: Updated to use `sr.Type` for type.

pilot/pkg/model/extensions.go
bianpengyuan: Make it redacted at ecdsz.

pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
bianpengyuan: Yeah makes sense. I've updated the behavior to what you described. Set and empty will not result in fetch. Agent will short circuit and nack in that case.

bianpengyuan:
@douglas-reid @zirain could you please take a look? Thanks!

pkg/wasm/convert.go
```diff
@@ -119,6 +120,27 @@ func convert(resource *any.Any, cache Cache) (newExtensionConfig *any.Any, sendN
...
+	if envs != nil {
+		if sec, found := envs.KeyValues[model.WasmSecretEnv]; found {
+			if sec == "" {
+				status = fetchFailure
+				wasmLog.Errorf("cannot fetch Wasm module %v: missing image pulling secret", wasmHTTPFilterConfig.Config.Name)
+				return
+			}
+			pullSecret = []byte(sec)
+		}
+		// Strip all internal env variables from VM env variable.
```
douglas-reid, line 134: nit: do we want to explain why we are stripping the variables here?

douglas-reid (approved):
the E&T owned file changes look reasonable to me. approving those changes.

pkg/wasm/convert.go
bianpengyuan: Addressed in https://github.com/istio/istio/pull/38228

PR #37167, opened 2022-02-04
Add option to replace Round Robin LB with Least Request

nmittler:
This is the first part of the work described in [Better Load Balancing by Default](https://docs.google.com/document/d/1SspUSsOt2GXWmSl8EmyVuvXROIHLAkD6XkcUsmw1y7w/edit?resourcekey=0-kXCZGVsqypEE9YZB5lVOFw#heading=h.xw1gqgyqs5b).

Various comparisons have shown Least Request (LEAST_CONN) LB to be strictly better and safer than Round Robin. With this change, LR will always be used instead of RR in all cases except gRPC, which does not currently support LR.

This option is expected to eventually be enabled by default (likely in the next release).

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -726,12 +726,16 @@ func applyLoadBalancer(c *cluster.Cluster, lb *networking.LoadBalancerSettings, 
 
 	// DO not do if else here. since lb.GetSimple returns a enum value (not pointer).
 	switch lb.GetSimple() {
+	case networking.LoadBalancerSettings_ROUND_ROBIN:
+		if features.EnableBetterLoadBalancing {
+			c.LbPolicy = cluster.Cluster_LEAST_REQUEST
+		} else {
+			c.LbPolicy = cluster.Cluster_ROUND_ROBIN
+		}
 	case networking.LoadBalancerSettings_LEAST_CONN:
 		c.LbPolicy = cluster.Cluster_LEAST_REQUEST
 	case networking.LoadBalancerSettings_RANDOM:
 		c.LbPolicy = cluster.Cluster_RANDOM
-	case networking.LoadBalancerSettings_ROUND_ROBIN:
-		c.LbPolicy = cluster.Cluster_ROUND_ROBIN
 	case networking.LoadBalancerSettings_PASSTHROUGH:
 		c.LbPolicy = cluster.Cluster_CLUSTER_PROVIDED
 		c.ClusterDiscoveryType = &cluster.Cluster_Type{Type: cluster.Cluster_ORIGINAL_DST}
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -278,6 +278,14 @@ var (
 			"ENABLE_MCS_HOST also be enabled.").Get() &&
 		EnableMCSHost
 
+	EnableBetterLoadBalancing = env.RegisterBoolVar(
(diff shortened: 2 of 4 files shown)


pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -726,12 +726,16 @@ func applyLoadBalancer(c *cluster.Cluster, lb *networking.LoadBalancerSettings,
 
 	// DO not do if else here. since lb.GetSimple returns a enum value (not pointer).
 	switch lb.GetSimple() {
+	case networking.LoadBalancerSettings_ROUND_ROBIN:
```
howardjohn, line 729: I get why you are doing this but it seems problematic that a user cannot set ROUND_ROBIN explicitly

nmittler: The reason being that RR is not generally safe and should never really be used. I don't think we should even allow it.

howardjohn: I don't think removing options is a good thing. Changing the default is good but we don't need to remove options just because they are not the ideal in standard scenarios.

nmittler: I generally understand where you're coming from, but I suspect nobody chooses RR. They just either use the default (probably 99.9% of cases) or select it because they don't particularly know or care which algorithm is used. Why would anyone ever want RR?

nmittler: How would this break backward compatibility?

howardjohn: Today: I set ROUND_ROBIN and get ROUND_ROBIN
Future: I set ROUND_ROBIN and do not get ROUND_ROBIN

I suppose it doesn't strictly break compat right now since its off by default but the plan is to make it default

nmittler: If the user needs RR, they will still be able to disable this option (assuming it's eventually enabled by default).

@louiscryan @smawson @costinm any thoughts?

ramaraochavali: So when it is eventually enabled by default - do we allow ROUND_ROBIN? I think there are valid usecases for ROUND_ROBIN. In our environment we set LEAST_CONN by default but certain services (one usecase is  one service has to connect its peers in a round robin way) use this. So removing the option is not good.

hzxuzhonghu:
Curious since Weighted Least is better, why not envoy make it default?

pilot/pkg/networking/core/v1alpha3/cluster.go
tonya11en: FWIW I have been preaching the `LEAST_CONN` gospel for years over in Envoy-land and still don't think it's a good idea to disallow WRR completely. Without the option to configure RR, there's no way to guarantee strict weighted host selection.

Just my 2 cents- I don't really have a horse in this race.

tonya11en:
@hzxuzhonghu @nmittler I don't think there was really any discussion about changing the default LB that I was involved with. This is probably something set from the start that nobody found troublesome enough to bother changing. My guess is that we didn't know if it would break anything for anyone and folks that have a problem with RR could always configure LR if they wanted.

There are other poor defaults like this such as retry circuit breaker defaults and retry budgets that nobody really revisited after learning more and implementing better alternatives.

nmittler:
Closing this in favor of https://github.com/istio/api/pull/2232

Issue #37056, opened 2022-01-27
Probes intended for proxyless gRPC can break other setups


stevenctl:
When `GRPC_XDS_BOOTSTRAP` is set we perform probes to make sure the certs exist. 

The certs are expected to be at `OUTPUT_CERT_DIR/...`. 

At some point we set a default value for the `GRPC_XDS_BOOTSTRAP` so that agent would always generate the bootstrap, and applications could just use it if they wanted. 

We either need to:

* Require enabling this check explicitly
* Make the assumed location of certs more intelligent 

cc @howardjohn

stevenctl:
https://istio.slack.com/archives/GHKNPELMS/p1643312321979929

stevenctl:
Basically this issue makes it so readiness fails when `OUTPUT_CERTS` is set but we don't write the expected files there.

stevenctl:
@howardjohn 

There are a lot of paths in this part of the code.. do you know off the top of your head a scenario where `OUTPUT_CERTS` is set, but we don't care that the certs aren't output? 

This failure will only occur if `OUTPUT_CERTS != ""`

howardjohn:
pkg/istio-agent/README.md could be useful if you haven't seen it. Do you
know *why* they are not written? Doesn't that mean we never fetched any
certs? That seems odd.

PR #37168, opened 2022-02-04
don't allow filemountedcerts to be set alongside output certs

stevenctl:
https://github.com/istio/istio/issues/37056

I believe this is the root cause

### DIFF
--- a/pilot/cmd/pilot-agent/options/security.go
+++ b/pilot/cmd/pilot-agent/options/security.go
@@ -140,6 +140,11 @@ func SetupSecurityOptions(proxyConfig *meshconfig.ProxyConfig, secOpt *security.
 	if o.ProvCert != "" && o.FileMountedCerts {
 		return nil, fmt.Errorf("invalid options: PROV_CERT and FILE_MOUNTED_CERTS are mutually exclusive")
 	}
+
+	if o.FileMountedCerts && o.OutputKeyCertToDir != "" {
+		return nil, fmt.Errorf("invalid options: OUTPUT_CERTS and FILE_MOUNTED_CERTS are mutually exclusive")
+	}
+
 	return o, nil
 }
 


howardjohn (commented):
seems reasonable but let me check with original reporter to make sure it was not intended to set both

pilot/cmd/pilot-agent/options/security.go
```diff
@@ -140,6 +140,11 @@ func SetupSecurityOptions(proxyConfig *meshconfig.ProxyConfig, secOpt *security.
 	if o.ProvCert != "" && o.FileMountedCerts {
 		return nil, fmt.Errorf("invalid options: PROV_CERT and FILE_MOUNTED_CERTS are mutually exclusive")
 	}
+
+	if o.FileMountedCerts && o.OutputKeyCertToDir != "" {
+		return nil, fmt.Errorf("invalid options: OUTPUT_CERTS and FILE_MOUNTED_CERTS are mutually exclusive")
```
howardjohn, line 145: I think istio VM is setting OUTPUT_CERTS by default which is probably why

stevenctl: Rather than the UX of:

1. I install on VMs with some special CA or file mounted certs setup
2. I get this error, I go manually remove `OUTPUT_CERTS` from the VMs config. 
3. Now it works. 

Instead maybe:

1. I install on VMs with something that triggers `o.FileMountedCerts`
2. Logs warn that outputcerts is ineffective due to this conflict (possibly the above conflict as well? i'd rather keep the change small)
3. We unset OUTPUT_CERTS here

stevenctl: wdyt?

howardjohn: SGTM

PR #37174, opened 2022-02-04
workloadSelector support in DR

kfaseela:

Support workloadSelector in DestinationRule so that usecases as mentioned in RFC [Simplify Sidecar Egress For MTLS](https://docs.google.com/document/d/1UXpT3rZpE2uFeMh5KffAZAPAoZBLNSHc/) can be better handled.

API changes: https://github.com/istio/api/pull/2207

This change involves making the sidecarScope maintain a list of DestinationRules per host, so that DRs can be filtered based on a workloadSelector. If a DR has a workloadSelector specified, it will be always namespaceLocal and won't be exported to other namespaces. 

- [ ] Docs
- [ ] Installation
- [ X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/destination_rule.go
+++ b/pilot/pkg/model/destination_rule.go
@@ -21,6 +21,7 @@ import (
 
 	networking "istio.io/api/networking/v1alpha3"
 	"istio.io/istio/pkg/config"
+	"istio.io/istio/pkg/config/labels"
 	"istio.io/istio/pkg/config/visibility"
 )
 
@@ -37,18 +38,29 @@ import (
 func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleConfig config.Config, exportToMap map[visibility.Instance]bool) {
 	rule := destRuleConfig.Spec.(*networking.DestinationRule)
 	resolvedHost := ResolveShortnameToFQDN(rule.Host, destRuleConfig.Meta)
-	newDr := true
 	if mdrList, exists := p.destRules[resolvedHost]; exists {
 		// Deep copy destination rule, to prevent mutate it later when merge with a new one.
 		// This can happen when there are more than one destination rule of same host in one namespace.
+		newDr := true
 		for i, mdr := range mdrList {
 			copied := mdr.DeepCopy()
 			p.destRules[resolvedHost][i] = &copied
 			mergedRule := copied.Spec.(*networking.DestinationRule)
-			if !workloadSelectorEquals(mergedRule.GetWorkloadSelector().GetMatchLabels(), rule.GetWorkloadSelector().GetMatchLabels()) {
+
+			// The new destination rule can be merged with the existing one, only if both has no
+			// workload selectors or a matching workload selector.
+			if mergedRule.GetWorkloadSelector() != nil && rule.GetWorkloadSelector() != nil &&
+				!labels.Instance(mergedRule.GetWorkloadSelector().GetMatchLabels()).Equals(rule.GetWorkloadSelector().GetMatchLabels()) {
 				continue
 			}
-			newDr = false
+			// Decide whether the incoming destination rule would become a new unique entry to the drList.
+			// If a new rule is created without any workload selector, it has to be merged to all the existing entries with workload selectors,
+			// and at the same time appended to the list as a global/namespaceLocal DR.
+			if (rule.GetWorkloadSelector() == nil && mergedRule.GetWorkloadSelector() == nil) || (rule.GetWorkloadSelector() != nil &&
+				labels.Instance(mergedRule.GetWorkloadSelector().GetMatchLabels()).Equals(rule.GetWorkloadSelector().GetMatchLabels())) {
+				newDr = false
+			}
+
 			existingSubset := map[string]struct{}{}
 			for _, subset := range mergedRule.Subsets {
 				existingSubset[subset.Name] = struct{}{}
@@ -120,16 +132,4 @@ func (ps *PushContext) inheritDestinationRule(parent, child *config.Config) *con
 	}
 
 	return &merged
-}
-
-func workloadSelectorEquals(a, b map[string]string) bool {
-	if len(a) != len(b) {
-		return false
-	}
-	for k, v := range a {
-		if b[k] != v {
-			return false
-		}
-	}
-	return true
 }
--- a/pilot/pkg/model/destination_rule.go
+++ b/pilot/pkg/model/destination_rule.go
@@ -42,14 +42,10 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 		// `addRuleToProcessedDestRules` determines if the incoming destination rule would become a new unique entry in the processedDestRules list.
 		addRuleToProcessedDestRules := true
 		for i, mdr := range mdrList {
-			// Deep copy destination rule, to prevent mutate it later when merge with a new one.
-			// This can happen when there are more than one destination rule of same host in one namespace.
-			copied := mdr.DeepCopy()
-			p.destRules[resolvedHost][i] = &copied
-			mergedRule := copied.Spec.(*networking.DestinationRule)
-			bothWithoutSelector := rule.GetWorkloadSelector() == nil && mergedRule.GetWorkloadSelector() == nil
-			bothWithSelector := mergedRule.GetWorkloadSelector() != nil && rule.GetWorkloadSelector() != nil
-			selectorsMatch := labels.Instance(mergedRule.GetWorkloadSelector().GetMatchLabels()).Equals(rule.GetWorkloadSelector().GetMatchLabels())
+			existingRule := mdr.Spec.(*networking.DestinationRule)
+			bothWithoutSelector := rule.GetWorkloadSelector() == nil && existingRule.GetWorkloadSelector() == nil
+			bothWithSelector := existingRule.GetWorkloadSelector() != nil && rule.GetWorkloadSelector() != nil
+			selectorsMatch := labels.Instance(existingRule.GetWorkloadSelector().GetMatchLabels()).Equals(rule.GetWorkloadSelector().GetMatchLabels())
 
 			if bothWithSelector && !selectorsMatch {
 				// If the new destination rule and the existing one has workload selectors associated with them, skip merging
@@ -63,6 +59,12 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 			if bothWithoutSelector || (rule.GetWorkloadSelector() != nil && selectorsMatch) {
 				addRuleToProcessedDestRules = false
 			}
+
+			// Deep copy destination rule, to prevent mutate it later when merge with a new one.
+			// This can happen when there are more than one destination rule of same host in one namespace.
+			copied := mdr.DeepCopy()
+			p.destRules[resolvedHost][i] = &copied
+			mergedRule := copied.Spec.(*networking.DestinationRule)
 
 			existingSubset := map[string]struct{}{}
 			for _, subset := range mergedRule.Subsets {
--- a/pilot/pkg/model/destination_rule.go
+++ b/pilot/pkg/model/destination_rule.go
@@ -39,26 +39,29 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 	rule := destRuleConfig.Spec.(*networking.DestinationRule)
 	resolvedHost := ResolveShortnameToFQDN(rule.Host, destRuleConfig.Meta)
 	if mdrList, exists := p.destRules[resolvedHost]; exists {
-		// Deep copy destination rule, to prevent mutate it later when merge with a new one.
-		// This can happen when there are more than one destination rule of same host in one namespace.
-		newDr := true
+		// `addRuleToProcessedDestRules` determines if the incoming destination rule would become a new unique entry in the processedDestRules list.
+		addRuleToProcessedDestRules := true
 		for i, mdr := range mdrList {
+			// Deep copy destination rule, to prevent mutate it later when merge with a new one.
+			// This can happen when there are more than one destination rule of same host in one namespace.
 			copied := mdr.DeepCopy()
 			p.destRules[resolvedHost][i] = &copied
 			mergedRule := copied.Spec.(*networking.DestinationRule)
+			bothWithoutSelector := rule.GetWorkloadSelector() == nil && mergedRule.GetWorkloadSelector() == nil
+			bothWithSelector := mergedRule.GetWorkloadSelector() != nil && rule.GetWorkloadSelector() != nil
+			selectorsMatch := labels.Instance(mergedRule.GetWorkloadSelector().GetMatchLabels()).Equals(rule.GetWorkloadSelector().GetMatchLabels())
 
-			// The new destination rule can be merged with the existing one, only if both has no
-			// workload selectors or a matching workload selector.
-			if mergedRule.GetWorkloadSelector() != nil && rule.GetWorkloadSelector() != nil &&
-				!labels.Instance(mergedRule.GetWorkloadSelector().GetMatchLabels()).Equals(rule.GetWorkloadSelector().GetMatchLabels()) {
+			if bothWithSelector && !selectorsMatch {
+				// If the new destination rule and the existing one has workload selectors associated with them, skip merging
+				// if the selectors do not match
 				continue
 			}
-			// Decide whether the incoming destination rule would become a new unique entry to the drList.
-			// If a new rule is created without any workload selector, it has to be merged to all the existing entries with workload selectors,
-			// and at the same time appended to the list as a global/namespaceLocal DR.
-			if (rule.GetWorkloadSelector() == nil && mergedRule.GetWorkloadSelector() == nil) || (rule.GetWorkloadSelector() != nil &&
-				labels.Instance(mergedRule.GetWorkloadSelector().GetMatchLabels()).Equals(rule.GetWorkloadSelector().GetMatchLabels())) {
-				newDr = false
+
+			// If both the destination rules are without a workload selector or with matching workload selectors, simply merge them.
+			// If the incoming rule has a workload selector, it has to be merged with the existing rules with workload selector, and
+			// at the same time added as a unique entry in the processedDestRules.
+			if bothWithoutSelector || (rule.GetWorkloadSelector() != nil && selectorsMatch) {
+				addRuleToProcessedDestRules = false
 			}
 
 			existingSubset := map[string]struct{}{}
@@ -93,7 +96,7 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 				p.exportTo[resolvedHost] = exportToMap
 			}
 		}
-		if newDr {
+		if addRuleToProcessedDestRules {
 			p.destRules[resolvedHost] = append(p.destRules[resolvedHost], &destRuleConfig)
 		}
 		return
--- a/pilot/pkg/model/destination_rule.go
+++ b/pilot/pkg/model/destination_rule.go
@@ -39,12 +39,12 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 	rule := destRuleConfig.Spec.(*networking.DestinationRule)
 	resolvedHost := ResolveShortnameToFQDN(rule.Host, destRuleConfig.Meta)
 	newDr := true
-	if mdrList, exists := p.destRule[resolvedHost]; exists {
+	if mdrList, exists := p.destRules[resolvedHost]; exists {
 		// Deep copy destination rule, to prevent mutate it later when merge with a new one.
 		// This can happen when there are more than one destination rule of same host in one namespace.
 		for i, mdr := range mdrList {
 			copied := mdr.DeepCopy()
-			p.destRule[resolvedHost][i] = &copied
+			p.destRules[resolvedHost][i] = &copied
 			mergedRule := copied.Spec.(*networking.DestinationRule)
 			if !reflect.DeepEqual(mergedRule.GetWorkloadSelector().GetMatchLabels(), rule.GetWorkloadSelector().GetMatchLabels()) {
 				continue
@@ -75,7 +75,6 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 			if mergedRule.TrafficPolicy == nil && rule.TrafficPolicy != nil {
 				mergedRule.TrafficPolicy = rule.TrafficPolicy
 			}
-
 			// If there is no exportTo in the existing rule and
 			// the incoming rule has an explicit exportTo, use the
 			// one from the incoming rule.
@@ -84,12 +83,12 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 			}
 		}
 		if newDr {
-			p.destRule[resolvedHost] = append(p.destRule[resolvedHost], &destRuleConfig)
+			p.destRules[resolvedHost] = append(p.destRules[resolvedHost], &destRuleConfig)
 		}
 		return
 	}
 	// DestinationRule does not exist for the resolved host so add it
-	p.destRule[resolvedHost] = append(p.destRule[resolvedHost], &destRuleConfig)
+	p.destRules[resolvedHost] = append(p.destRules[resolvedHost], &destRuleConfig)
 	p.exportTo[resolvedHost] = exportToMap
 }
 
--- a/pilot/pkg/model/destination_rule.go
+++ b/pilot/pkg/model/destination_rule.go
@@ -16,7 +16,6 @@ package model
 
 import (
 	"fmt"
-	"reflect"
 
 	"google.golang.org/protobuf/proto"
 
@@ -46,7 +45,7 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 			copied := mdr.DeepCopy()
 			p.destRules[resolvedHost][i] = &copied
 			mergedRule := copied.Spec.(*networking.DestinationRule)
-			if !reflect.DeepEqual(mergedRule.GetWorkloadSelector().GetMatchLabels(), rule.GetWorkloadSelector().GetMatchLabels()) {
+			if !workloadSelectorEquals(mergedRule.GetWorkloadSelector().GetMatchLabels(), rule.GetWorkloadSelector().GetMatchLabels()) {
 				continue
 			}
 			newDr = false
@@ -121,4 +120,16 @@ func (ps *PushContext) inheritDestinationRule(parent, child *config.Config) *con
 	}
 
 	return &merged
+}
+
+func workloadSelectorEquals(a, b map[string]string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for k, v := range a {
+		if b[k] != v {
+			return false
+		}
+	}
+	return true
 }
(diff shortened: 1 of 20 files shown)


pilot/pkg/model/push_context.go
```diff
@@ -936,17 +936,30 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Colle
...
 		return nil
 	}
-
 	// If proxy has a sidecar scope that is user supplied, then get the destination rules from the sidecar scope
 	// sidecarScope.config is nil if there is no sidecar scope for the namespace
 	if proxy.SidecarScope != nil && proxy.Type == SidecarProxy {
 		// If there is a sidecar scope for this proxy, return the destination rule
 		// from the sidecar scope.
-		return proxy.SidecarScope.destinationRules[service.Hostname]
+		destinationRuleConfig := proxy.SidecarScope.destinationRules[service.Hostname]
```
howardjohn, line 948: I would expect I can have

DR: google.com, for app=foo
DR: google.com, for app=bar

And both work. However, the `destinationRules` map only holds ONE DR for a given hostname. Because of that, we will probably get `nil` for one of them. Both should be possible. I think in the sidecarscope construction (which is in the context of a single workload) we should do the label selection. At this point we have already lost weight

pilot/pkg/model/destination_rule.go
```diff
@@ -38,55 +39,63 @@ import (
...
-			} else {
-				// duplicate subset
-				ps.AddMetric(DuplicatedSubsets, string(resolvedHost), "",
-					fmt.Sprintf("Duplicate subset %s found while merging destination rules for %s",
-						subset.Name, string(resolvedHost)))
+		for i, mdr := range mdrList {
+			copied := mdr.DeepCopy()
+			p.destRule[resolvedHost][i] = &copied
+			mergedRule := copied.Spec.(*networking.DestinationRule)
+			if !reflect.DeepEqual(mergedRule.GetWorkloadSelector().GetMatchLabels(), rule.GetWorkloadSelector().GetMatchLabels()) {
```
kfaseela, line 50: this is the only major change in the method. remaining green lines are due to the right shift due to the new for loop

pilot/pkg/model/push_context.go
kfaseela: Have made the destionationRules hold a list per host, so that we can do the filtering based on workloadSelector

ramaraochavali (commented):
left some comments/questions. Will take a look later.

pilot/pkg/model/destination_rule.go
```diff
@@ -38,55 +39,63 @@ import (
...
-		for _, subset := range rule.Subsets {
-			if _, ok := existingSubset[subset.Name]; !ok {
-				// if not duplicated, append
-				mergedRule.Subsets = append(mergedRule.Subsets, subset)
-			} else {
-				// duplicate subset
-				ps.AddMetric(DuplicatedSubsets, string(resolvedHost), "",
-					fmt.Sprintf("Duplicate subset %s found while merging destination rules for %s",
-						subset.Name, string(resolvedHost)))
+		for i, mdr := range mdrList {
```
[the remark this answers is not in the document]
kfaseela: Hi Rama, Hope this is clarified now once you saw the changes in push_context.go

pilot/pkg/model/push_context.go
```diff
@@ -1641,8 +1642,10 @@ func (ps *PushContext) SetDestinationRules(configs []config.Config) {
 
 		rule.Host = string(ResolveShortnameToFQDN(rule.Host, configs[i].Meta))
 		exportToMap := make(map[visibility.Instance]bool)
-		for _, e := range rule.ExportTo {
-			exportToMap[visibility.Instance(e)] = true
+		if rule.GetWorkloadSelector() == nil {
+			for _, e := range rule.ExportTo {
+				exportToMap[visibility.Instance(e)] = true
```
howardjohn, line 1647: this makes sense but can we comment why we do this? its not super obvious

Also can we add validation that if workloadSelector is set exportTo must be `.` or unset? in validation.go

pilot/pkg/model/sidecar.go
```diff
@@ -216,12 +216,14 @@ func DefaultSidecarScopeForNamespace(ps *PushContext, configNamespace string) *S
...
-		out.AddConfigDependencies(ConfigKey{
-			Kind:      gvk.DestinationRule,
-			Name:      dr.Name,
-			Namespace: dr.Namespace,
-		})
+	if len(out.destinationRules) == 1 {
+		for _, dr := range out.destinationRules {
+			out.AddConfigDependencies(ConfigKey{
+				Kind:      gvk.DestinationRule,
+				Name:      dr[0].Name,
```
howardjohn, line 223: seems we should add all, not `[0]`

pilot/pkg/model/destination_rule.go
ramaraochavali: nit: you can compare maps directly instead of reflect.DeepEquals to imrprove performance.

pilot/pkg/model/push_context.go
```diff
@@ -256,7 +256,7 @@ type processedDestRules struct {
 	// Map of dest rule host to the list of namespaces to which this destination rule has been exported to
 	exportTo map[host.Name]map[visibility.Instance]bool
 	// Map of dest rule host and the merged destination rules for that host
-	destRule map[host.Name]*config.Config
+	destRule map[host.Name][]*config.Config
```
ramaraochavali, line 259: nit: destRule -> destRules

ramaraochavali:
few minor nits

pilot/pkg/model/sidecar.go
kfaseela: Done

pilot/pkg/model/sidecar.go
```diff
@@ -551,9 +554,30 @@ func (sc *SidecarScope) AddConfigDependencies(dependencies ...ConfigKey) {
...
-	return sc.destinationRules[svc]
+func (sc *SidecarScope) DestinationRule(direction TrafficDirection, proxy *Proxy, svc host.Name) *config.Config {
+	destinationRules := sc.destinationRules[svc]
+	var catchAllDr *config.Config
+	for _, destRule := range destinationRules {
+		destinationRule := destRule.Spec.(*networking.DestinationRule)
+		if destinationRule.GetWorkloadSelector() == nil {
+			catchAllDr = destRule
+		}
+		// If the DestinationRule has a workloadSelector, filter based on that only for outbound configuration
```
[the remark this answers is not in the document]
kfaseela: Done

pilot/pkg/model/push_context.go
kfaseela: Refactored

pilot/pkg/model/push_context.go
kfaseela: Done

pilot/pkg/model/push_context.go
```diff
@@ -1035,11 +1034,13 @@ func (ps *PushContext) getExportedDestinationRuleFromNamespace(owningNamespace s
 						parent = ps.destinationRuleIndex.inheritedByNamespace[ps.Mesh.RootNamespace]
 					}
 					if child := ps.destinationRuleIndex.exportedByNamespace[owningNamespace].destRule[specificHostname]; child != nil {
-						return ps.inheritDestinationRule(parent, child)
+						return ps.inheritDestinationRule(parent, child[0])
```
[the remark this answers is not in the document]
kfaseela: This is the ExportedDR method. But we discussed offline, I have made sure that SetDestinationRules() func inherits all the DRs for a host appropriately

pilot/pkg/model/push_context.go
```diff
@@ -1036,12 +1035,14 @@ func (ps *PushContext) getExportedDestinationRuleFromNamespace(owningNamespace s
...
-					if child := ps.destinationRuleIndex.exportedByNamespace[owningNamespace].destRule[specificHostname]; child != nil {
-						return ps.inheritDestinationRule(parent, child)
+					if child := ps.destinationRuleIndex.exportedByNamespace[owningNamespace].destRules[specificHostname]; child != nil {
+						return ps.inheritDestinationRule(parent, child[0])
 					}
 					return nil
 				}
-				return ps.destinationRuleIndex.exportedByNamespace[owningNamespace].destRule[specificHostname]
+				if dr, ok := ps.destinationRuleIndex.exportedByNamespace[owningNamespace].destRules[specificHostname]; ok {
+					return dr[0]
```
howardjohn, line 1044: why return just one here?

howardjohn: It seems like anywehre we do a `[0]` is a bug?

pilot/pkg/model/sidecar_test.go
```diff
@@ -868,6 +898,78 @@ var (
...
+	}
+	destinationRule2 = config.Config{
+		Meta: config.Meta{
+			Name:      "drRule2",
+			Namespace: "mynamespace",
+		},
+		Spec: &networking.DestinationRule{
+			Host: "httpbin.org",
+			WorkloadSelector: &v1beta1.WorkloadSelector{
+				MatchLabels: map[string]string{"app": "app2"},
```
howardjohn, line 937: Should add a test for when we have 2 DRs with the same labels,

howardjohn: and can we add one for one with no hostname (merging) and one with labels?

pkg/config/validation/validation.go
```diff
@@ -679,11 +678,15 @@ var ValidateDestinationRule = registerValidateFunc("ValidateDestinationRule",
 			v = appendValidation(v, validateSubset(subset))
 		}
 
-		v = appendValidation(v, validateExportTo(cfg.Namespace, rule.ExportTo, false))
+		v = appendValidation(v,
+			validateExportTo(cfg.Namespace, rule.ExportTo, false, rule.GetWorkloadSelector() != nil))
+
+		v = appendValidation(v, validateWorkloadSelector(rule.GetWorkloadSelector()))
```
howardjohn, line 684: can we also validate there is a hostname set with workload selector?

pilot/pkg/model/push_context.go
kfaseela: Fixed. Was thinking exported rules will be anyways one, so why iterate. However now have made the code uniform

pilot/pkg/model/sidecar_test.go
kfaseela: Done

pkg/config/validation/validation.go
kfaseela: Added

pilot/pkg/model/destination_rule.go
```diff
@@ -120,3 +127,15 @@ func (ps *PushContext) inheritDestinationRule(parent, child *config.Config) *con
 
 	return &merged
 }
+
+func workloadSelectorEquals(a, b map[string]string) bool {
```
ramaraochavali, line 131: looks like this already exists https://github.com/istio/istio/blob/35b6de45318e1f52544fd747f532732aac639f5d/pkg/config/labels/instance.go#L76

howardjohn: +1:

```
labels.Instance(a).Equals(b)
```

howardjohn (commented):
LGTM! aside from a couple minor comments

pilot/pkg/model/destination_rule.go
kfaseela: Changed

pkg/config/validation/validation.go
```diff
@@ -679,11 +678,20 @@ var ValidateDestinationRule = registerValidateFunc("ValidateDestinationRule",
...
+
+		if rule.Host == "" && rule.GetWorkloadSelector() != nil {
```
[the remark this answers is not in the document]
kfaseela: Done

pilot/pkg/model/destination_rule.go
```diff
@@ -38,55 +38,62 @@ import (
...
-			} else {
-				// duplicate subset
-				ps.AddMetric(DuplicatedSubsets, string(resolvedHost), "",
-					fmt.Sprintf("Duplicate subset %s found while merging destination rules for %s",
-						subset.Name, string(resolvedHost)))
+		for i, mdr := range mdrList {
+			copied := mdr.DeepCopy()
+			p.destRules[resolvedHost][i] = &copied
+			mergedRule := copied.Spec.(*networking.DestinationRule)
+			if !workloadSelectorEquals(mergedRule.GetWorkloadSelector().GetMatchLabels(), rule.GetWorkloadSelector().GetMatchLabels()) {
```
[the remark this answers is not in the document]
kfaseela: Enhanced the check with additional comments. Hope it looks good now?

pilot/pkg/model/destination_rule.go
```diff
@@ -38,18 +39,29 @@ import (
...
+		newDr := true
 		for i, mdr := range mdrList {
 			copied := mdr.DeepCopy()
 			p.destRules[resolvedHost][i] = &copied
 			mergedRule := copied.Spec.(*networking.DestinationRule)
-			if !workloadSelectorEquals(mergedRule.GetWorkloadSelector().GetMatchLabels(), rule.GetWorkloadSelector().GetMatchLabels()) {
+
+			// The new destination rule can be merged with the existing one, only if both has no
+			// workload selectors or a matching workload selector.
+			if mergedRule.GetWorkloadSelector() != nil && rule.GetWorkloadSelector() != nil &&
```
howardjohn, line 53: Where do we handle "both has no workload selector"?

pilot/pkg/model/destination_rule.go
```diff
@@ -38,18 +39,29 @@ import (
...
+				!labels.Instance(mergedRule.GetWorkloadSelector().GetMatchLabels()).Equals(rule.GetWorkloadSelector().GetMatchLabels()) {
 				continue
 			}
-			newDr = false
+			// Decide whether the incoming destination rule would become a new unique entry to the drList.
+			// If a new rule is created without any workload selector, it has to be merged to all the existing entries with workload selectors,
+			// and at the same time appended to the list as a global/namespaceLocal DR.
+			if (rule.GetWorkloadSelector() == nil && mergedRule.GetWorkloadSelector() == nil) || (rule.GetWorkloadSelector() != nil &&
```
howardjohn, line 60: nit: may help to split some of this up.
like
```
bothWithoutSelector := rule.GetWorkloadSelector() == nil && mergedRule.GetWorkloadSelector() == nil
selectorsMatch := ...
...
if bothWithoutSelector || selectorsMatch { ... }
```

kfaseela: Done

howardjohn:
@hzxuzhonghu putting a hold on a PR without explaination is quite demotivating, especially for a new contributor who has been working on this for. months.

Can you please list the criteria required for the hold to be removed?

hzxuzhonghu:
I want to have a look and have a DM with @kfaseela

howardjohn:
Ok, SGTM. Just want to make sure it's not unbounded

pilot/pkg/model/destination_rule.go
```diff
@@ -38,55 +39,76 @@ import (
...
 
-		// If there is no top level policy and the incoming rule has top level
-		// traffic policy, use the one from the incoming rule.
-		if mergedRule.TrafficPolicy == nil && rule.TrafficPolicy != nil {
-			mergedRule.TrafficPolicy = rule.TrafficPolicy
-		}
+			// If both the destination rules are without a workload selector or with matching workload selectors, simply merge them.
+			// If the incoming rule has a workload selector, it has to be merged with the existing rules with workload selector, and
+			// at the same time added as a unique entry in the processedDestRules.
+			if bothWithoutSelector || (rule.GetWorkloadSelector() != nil && selectorsMatch) {
```
hzxuzhonghu, line 64: have you checked when one has selector, but the other not.

pilot/pkg/model/destination_rule.go
```diff
@@ -38,55 +39,76 @@ import (
 func (ps *PushContext) mergeDestinationRule(p *processedDestRules, destRuleConfig config.Config, exportToMap map[visibility.Instance]bool) {
```
hzxuzhonghu, line 39: Please add ut for this, it is very complex now

pilot/pkg/model/push_context.go
```diff
@@ -256,7 +256,7 @@ type processedDestRules struct {
 	// Map of dest rule host to the list of namespaces to which this destination rule has been exported to
 	exportTo map[host.Name]map[visibility.Instance]bool
 	// Map of dest rule host and the merged destination rules for that host
-	destRule map[host.Name]*config.Config
+	destRules map[host.Name][]*config.Config
```
hzxuzhonghu, line 259: I am concerned about the searching performance

hzxuzhonghu: Also the merging performance

pilot/pkg/model/push_context.go
hzxuzhonghu: Do we need to specially check workload selector then?

pilot/pkg/model/push_context.go
```diff
@@ -1667,7 +1677,9 @@ func (ps *PushContext) SetDestinationRules(configs []config.Config) {
 		// We only honor . and *
 		if len(rule.ExportTo) == 0 && ps.exportToDefaults.destinationRule[visibility.Private] {
 			isPrivateOnly = true
-		} else if len(rule.ExportTo) == 1 && exportToMap[visibility.Private] {
+		} else if len(rule.ExportTo) == 1 && (exportToMap[visibility.Private]) {
+			isPrivateOnly = true
+		} else if rule.GetWorkloadSelector() != nil {
```
hzxuzhonghu, line 1682: ditto

pilot/pkg/model/sidecar.go
```diff
@@ -97,8 +98,7 @@ type SidecarScope struct {
 	// corresponds to a service in the services array above. When computing
 	// CDS, we simply have to find the matching service and return the
 	// destination rule.
-	destinationRules map[host.Name]*config.Config
-
+	destinationRules map[host.Name][]*config.Config
```
hzxuzhonghu, line 101: type map[host.Name][namespace]*config.Config ?

hzxuzhonghu: I have another question: if a DR without workload selector, and a DR with selector/ Both applies to a proxy. Then do they merge? Or the one with sector takes priority

pilot/pkg/model/sidecar.go
```diff
@@ -216,12 +216,14 @@ func DefaultSidecarScopeForNamespace(ps *PushContext, configNamespace string) *S
...
 
-	for _, dr := range out.destinationRules {
-		out.AddConfigDependencies(ConfigKey{
-			Kind:      gvk.DestinationRule,
-			Name:      dr.Name,
-			Namespace: dr.Namespace,
-		})
+	for _, drList := range out.destinationRules {
+		for _, dr := range drList {
+			out.AddConfigDependencies(ConfigKey{
```
hzxuzhonghu, line 221: Not related to this pr, i think from merge beginning the dependency is not right. We can ignore some drs, if they are merged in to a dr

pilot/pkg/model/sidecar.go
```diff
@@ -552,8 +555,32 @@ func (sc *SidecarScope) AddConfigDependencies(dependencies ...ConfigKey) {
 }
 
 // DestinationRule returns a destinationrule for a svc.
-func (sc *SidecarScope) DestinationRule(svc host.Name) *config.Config {
-	return sc.destinationRules[svc]
+func (sc *SidecarScope) DestinationRule(direction TrafficDirection, proxy *Proxy, svc host.Name) *config.Config {
```
hzxuzhonghu, line 558: same concern performance

pilot/pkg/model/push_context.go
kfaseela: I have added checks in validator.go to make sure that wrong exportTos won't get propagated. This was added as a double check.

pilot/pkg/model/sidecar.go
kfaseela: The filtering logic for the specific proxy will get a merged version of DR with selector + without selector. However other proxies in the same namespace will get the "without" selector DR.

pkg/config/validation/validation.go
```diff
@@ -679,11 +678,20 @@ var ValidateDestinationRule = registerValidateFunc("ValidateDestinationRule",
...
+		if features.EnableDestinationRuleInheritance && rule.Host == "" && rule.GetWorkloadSelector() != nil {
+			v = appendValidation(v,
+				fmt.Errorf("mesh/namespace destination rule cannot have workloadSelector configured"))
+		}
+
 		return v.Unwrap()
 	})
 
-func validateExportTo(namespace string, exportTo []string, isServiceEntry bool) (errs error) {
+func validateExportTo(namespace string, exportTo []string, isServiceEntry bool, isDestinationRuleWithWs bool) (errs error) {
```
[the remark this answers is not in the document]
kfaseela: Since both the functions will have to do similar exportTo parsing, I have just retained it this way

pkg/config/validation/validation.go
```diff
@@ -679,11 +678,20 @@ var ValidateDestinationRule = registerValidateFunc("ValidateDestinationRule",
 			v = appendValidation(v, validateSubset(subset))
 		}
 
-		v = appendValidation(v, validateExportTo(cfg.Namespace, rule.ExportTo, false))
+		v = appendValidation(v,
+			validateExportTo(cfg.Namespace, rule.ExportTo, false, rule.GetWorkloadSelector() != nil))
+
+		v = appendValidation(v, validateWorkloadSelector(rule.GetWorkloadSelector()))
+
+		if features.EnableDestinationRuleInheritance && rule.Host == "" && rule.GetWorkloadSelector() != nil {
```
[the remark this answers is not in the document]
kfaseela: Done

pkg/config/validation/validation.go
```diff
@@ -693,7 +701,10 @@ func validateExportTo(namespace string, exportTo []string, isServiceEntry bool)
 				// substitute this with the current namespace so that we
 				// can check for duplicates like ., namespace
 				key = namespace
+			} else if isDestinationRuleWithWs {
+				errs = appendErrors(errs, fmt.Errorf("destination rule with workload selector can only have exportTo: (.)"))
```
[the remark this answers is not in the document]
kfaseela: Missed this, Have changed it now

pilot/pkg/model/push_context.go
kfaseela: Same comment as above, I added this as a double check, even though we validate things beforehand.

pkg/config/validation/validation.go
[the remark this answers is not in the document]
hzxuzhonghu: ok, not a big deal

pkg/config/validation/validation.go
```diff
@@ -716,6 +724,13 @@ func validateExportTo(namespace string, exportTo []string, isServiceEntry bool)
 			}
 		}
 
+		// Make sure workloadSelector based destination rule does not use exportTo other than current namespace
+		if _, exists := exportToMap[namespace]; !exists {
+			if isDestinationRuleWithWs {
+				errs = appendErrors(errs, fmt.Errorf("destination rule with workload selector cannot have exportTo beyond current namespace"))
+			}
+		}
+
```
hzxuzhonghu, line 733: This looks a little hard to read, actually isDestinationRuleWithWs implicitly means dr with selector

pilot/pkg/model/destination_rule.go
```diff
@@ -38,55 +39,76 @@ import (
...
+		addRuleToProcessedDestRules := true
+		for i, mdr := range mdrList {
+			// Deep copy destination rule, to prevent mutate it later when merge with a new one.
+			// This can happen when there are more than one destination rule of same host in one namespace.
+			copied := mdr.DeepCopy()
+			p.destRules[resolvedHost][i] = &copied
+			mergedRule := copied.Spec.(*networking.DestinationRule)
+			bothWithoutSelector := rule.GetWorkloadSelector() == nil && mergedRule.GetWorkloadSelector() == nil
+			bothWithSelector := mergedRule.GetWorkloadSelector() != nil && rule.GetWorkloadSelector() != nil
+			selectorsMatch := labels.Instance(mergedRule.GetWorkloadSelector().GetMatchLabels()).Equals(rule.GetWorkloadSelector().GetMatchLabels())
```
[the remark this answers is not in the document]
hzxuzhonghu: SG

pilot/pkg/model/destination_rule.go
hzxuzhonghu: @kfaseela Can you point me?  Where do they merge?

hzxuzhonghu: So come back to original doubt https://github.com/istio/istio/pull/37174#discussion_r832791802

pilot/pkg/model/sidecar.go
hzxuzhonghu: THis logic is right.

pilot/pkg/model/destination_rule.go
kfaseela: if one has selector, and the other one does not have selector, we will have "addRuleToProcessedDestRules=true". And as the comment here says, in such cases, rule with selector merges to the rule without selector, as well as at line [102](https://github.com/istio/istio/pull/37174/files#diff-3320c5cd8bd7f37eca4dead440d5f0999739bd17c5939e4e2340f82a33f3f835R102), we add the new entry separately to the list. Is it clear now @hzxuzhonghu  ?

hzxuzhonghu: yep

kfaseela:
@hzxuzhonghu : Could you please take another look, and remove the do-not-merge label if the changes look okay to you now?

hzxuzhonghu:
Generally lg, only concerned with the performance because DR is so widely used.

hzxuzhonghu:
With this pr, it can influence the perf even no DR with selector created.  So I wonder if you can make it not a perf regression.

pilot/pkg/model/push_context.go
howardjohn: Yeah I don't see what the issue is. For all current cases, this will be a list of size 1.

If a user starts making use of this new feature, they could have a few items here. I cannot imagine this being a large number.

For example, we would have 10 items if _in the same namespace_, 10 distinct workload labels had customized DestinationRules for the same hostname. That seems extremely low probability, and even if it is, its a linear scale not n^2 so doesn't seem too bad?

Is there a concern I missed that makes this perform worse than this?

howardjohn (commented):
lgtm, just letting @hzxuzhonghu review/approve

hzxuzhonghu:
Once workload selector is in. normal users would set selector natively. 

Other than merging, the SidecarScope search `https://github.com/istio/istio/pull/37174/files#diff-42d26d46fa9490c70f6bd91080addf3ecc57aa44ebae320708ae58c824af8a9dR558` can be very bad.

Imagine a very bad usage, a mesh with `n` namespace, each namespace have `m` drs after merged, and totally `x`  proxies. 

The search will be o(n^2)?

howardjohn:
What amount of n, m, and x are you concerned about? I can setup a bench mark for this if we cannot figure it out theoretically

howardjohn (approved):

hzxuzhonghu:
The performance issue can only occur when set many DRs with different selectors for same service in the same namespace. So I compromised because  in real world, the size should be very small.

hzxuzhonghu (approved):

kfaseela:
> The performance issue can only occur when set many DRs with different selectors for same service in the same namespace. So I compromised because in real world, the size should be very small.

Thanks @howardjohn and @hzxuzhonghu !

PR #37192, opened 2022-02-07
Add env to configure the RSA key size of workload certificates

sha-rath:
This PR adds a new env, "WORKLOAD_RSA_KEY_SIZE" to make the RSA key size of workload certificates configurable. Note that there is already a similar env for configuring the RSA key size of self-signed CA. Both of these default to 2048 bits. Users who prefer to have a higher key size for better security can benefit from this env.

### DIFF
--- a/pilot/cmd/pilot-agent/options/options.go
+++ b/pilot/cmd/pilot-agent/options/options.go
@@ -75,6 +75,8 @@ var (
 
 	secretRotationGracePeriodRatioEnv = env.RegisterFloatVar("SECRET_GRACE_PERIOD_RATIO", 0.5,
 		"The grace period ratio for the cert rotation, by default 0.5.").Get()
+	workloadRSAKeySizeEnv = env.RegisterIntVar("WORKLOAD_RSA_KEY_SIZE", 2048,
+		"Specify the RSA key size to use for workload certificates.").Get()
 	pkcs8KeysEnv = env.RegisterBoolVar("PKCS8_KEY", false,
 		"Whether to generate PKCS#8 private keys").Get()
 	eccSigAlgEnv        = env.RegisterStringVar("ECC_SIGNATURE_ALGORITHM", "", "The type of ECC signature algorithm to use when generating private keys").Get()
--- a/security/pkg/nodeagent/cache/secretcache.go
+++ b/security/pkg/nodeagent/cache/secretcache.go
@@ -45,9 +45,6 @@ var (
 )
 
 const (
-	// The size of a private key for a leaf certificate.
-	keySize = 2048
-
 	// firstRetryBackOffInMilliSec is the initial backoff time interval when hitting
 	// non-retryable error in CSR request or while there is an error in reading file mounts.
 	firstRetryBackOffInMilliSec = 50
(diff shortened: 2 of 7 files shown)


security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -560,7 +557,7 @@ func (sc *SecretManagerClient) generateNewSecret(resourceName string) (*security
 	cacheLog.Debugf("constructed host name for CSR: %s", csrHostName.String())
 	options := pkiutil.CertOptions{
 		Host:       csrHostName.String(),
-		RSAKeySize: keySize,
+		RSAKeySize: sc.configOptions.WorkloadRSAKeySize,
```
bianpengyuan, line 560: Do we need to put some validation on the configured keysize, warn and fallback to the default if it is not valid, to prevent user shoot themselves in the foot?

sha-rath: It would be good to have this, but I'm not sure if it is really needed. I don't see this being done for other envs. Also, isn't the user responsible for providing a reasonable value for this field? The GenCSR func currently checks for the minimumRSAKeySize to be 2048, similarly, maxRSAKeySize could be introduced? @jacob-delgado do you have any opinion on this?

jacob-delgado: No real opinion. If someone is doing this they are doing this intentionally so I hope they'd not misconfigure a system. A minimum check is recommended though so they cannot create an insecure system (e.g. 512).

hzxuzhonghu:
One confusion: If 2048 is not secure enough, why not change to 4096_

sha-rath:
@hzxuzhonghu It apparently impacts provisioning of many certs on large clusters. Also, there are some tls handshake concerns for latency and cpu time. Hence, the default was decided to be kept as 2048 bits. However, in my deployment, I have set the keysize to 3072 based on a security requirement we have.

jacob-delgado (approved):

sha-rath:
@myidpt could you please review/approve this PR?

kfaseela (approved):

sha-rath:
@howardjohn can you pls take a look into this?

sha-rath:
cc @ericvn

pilot/cmd/pilot-agent/options/options.go
```diff
@@ -75,6 +75,8 @@ var (
 
 	secretRotationGracePeriodRatioEnv = env.RegisterFloatVar("SECRET_GRACE_PERIOD_RATIO", 0.5,
 		"The grace period ratio for the cert rotation, by default 0.5.").Get()
+	workloadRSAKeySizeEnv = env.RegisterIntVar("WORKLOAD_RSA_KEY_SIZE", 2048,
```
hzxuzhonghu, line 78: How is it going with an invalid value

sha-rath: Container startup fails and therefore, the pod wouldn't be running.
But again, as mentioned earlier, this would be the case with most other envs right? It is the responsibility of the user to provide a valid value and not break the system.

sha-rath:
@ramaraochavali would you like to review this PR?

ramaraochavali (commented):
LGTM, I would like @istio/wg-security-maintainers to take a look at this

sha-rath:
Okay...thanks for the quick review!

sha-rath:
@jacob-delgado @lizan could you pls help in getting this merged?

sha-rath:
cc @howardjohn

sha-rath:
@istio/wg-networking-maintainers PTAL

zirain:
Not stale

jacob-delgado:
@myidpt can you take a look?

sha-rath:
Adding @lizan @liminw for review..

lizan (approved):

Issue #37196, opened 2022-02-08
Protocol sniffing selects wrong protocol across gateway hop


howardjohn:
Client --> ingress --> sidecar --> pod (with sniffing)

Requests always make it to pod as `http 1.1`, and are seen as TCP by the pod. This because we do not set USE_DOWNSTREAM_PROTOCOL _and_ we set `istio-peer-exchange` ALPN - so we have 2 issues

Issue #21230, opened 2020-02-18
Protocol Detection breaking H2 traffic to gateway


howardjohn:
https://github.com/istio/istio/commit/fd29fd362b7e0b6561734fd64444651baab43f2d introduced a change to use USE_DOWNSTREAM_PROTOCOL if sniffing is enabled. This seems to break gateway traffic.

Relevant configs:

443 listener:
```json
 {
     "version_info": "2020-02-18T18:34:05Z/3",
     "listener": {
      "name": "0.0.0.0_443",
      "address": {
       "socket_address": {
        "address": "0.0.0.0",
        "port_value": 443
       }
      },
      "filter_chains": [
       {
        "filter_chain_match": {
         "server_names": [
          "*.mydomain"
         ]
        },
        "tls_context": {
         "common_tls_context": {
          "alpn_protocols": [
           "h2",
           "http/1.1"
          ],
          "tls_certificate_sds_secret_configs": [
           {
            "name": "ingressgateway-wildcard-certs",
            "sds_config": {
             "api_config_source": {
              "api_type": "GRPC",
              "grpc_services": [
               {
                "google_grpc": {
                 "target_uri": "unix:/var/run/ingress_gateway/sds",
                 "stat_prefix": "sdsstat"
                }
               }
              ]
             }
            }
           }
          ]
         },
         "require_client_certificate": false
        },
        "filters": [
         {
          "name": "envoy.http_connection_manager",
          "typed_config": {
           "@type": "type.googleapis.com/envoy.config.filter.network.http_connection_manager.v2.HttpConnectionManager",
           "stat_prefix": "outbound_0.0.0.0_443",
           "http_filters": [
            {
             "name": "envoy.cors"
            },
            {
             "name": "envoy.fault"
            },
            {
             "name": "envoy.router"
            }
           ],
           "http_protocol_options": {},
           "server_name": "istio-envoy",
           "idle_timeout": "300s",
           "access_log": [
            {
             "name": "envoy.file_access_log",
             "typed_config": {
              "@type": "type.googleapis.com/envoy.config.accesslog.v2.FileAccessLog",
              "path": "/dev/stdout",
[...]

PR #37198, opened 2022-02-08
Fix client protocol detection at gateway

howardjohn:
Fixes https://github.com/istio/istio/issues/37196

WIP because I will add some more tests and release note. 

Related issue: https://github.com/istio/istio/issues/21230


howardjohn:
One thing I think may break is:
* Server is http1 only
* Client uses ALPN negotiation

Client will negotiate to http2, and this PR now makes that forward as http2 to the server instead of http1....

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1185,8 +1185,9 @@ func (cb *ClusterBuilder) setUpstreamProtocol(mc *MutableCluster, port *model.Po
 	// Since protocol detection is disabled for gateway and use_downstream_protocol is used
 	// under protocol detection for cluster to select upstream connection protocol when
 	// the service port is unnamed. use_downstream_protocol should be disabled for gateway.
-	if cb.sidecarProxy() && ((util.IsProtocolSniffingEnabledForInboundPort(port) && direction == model.TrafficDirectionInbound) ||
```
hzxuzhonghu, line 1188: It seems the current behavior is not work for pure http1.1 server either

hzxuzhonghu: Does the istio mtls influence the protocol selection?

ramaraochavali: > It seems the current behavior is not work for pure http1.1 server either

Why do you think so? It works. We are using this for http1.1 server and also http2.0 server

ramaraochavali: nm...I see what you are saying. If server only supports Http1.1, client can not use both http2/http1.1 - but that is not the usecase for `useClientProtocol` though.

howardjohn: Yes currently if mtls is used (which it is by default) the server side will process inbound as raw tcp regardless of htto1 or http2. So there are actually two distinct issues at play here

howardjohn:
After some discussion with @lambdai , I think we should not do this.

Consider HTTPS gateway, auto protocol backend

Before: always use HTTP1 to backend
After: use client protocol

Before its broken if their app requires http2
After, its broken if their app requires http1 AND clients use ALPN negotiation (because we always advertise h2)

Seems there is no perfect solution


Maybe we should keep current state. Users can explicitly configure new behavior already with DestinationRule `useClientProtocol` (or just configure the protocol to http/http2 explicitly). And all we should fix is the istio-peer-exchange ALPN -- since right now the server side will interpret as TCP which is wrong.

Will update the PR to just add the ALPN fix

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
hzxuzhonghu: How about a client without sidecar communicate with a server with sidecar then? 
In this case. does it work?
client -> http2-> sidecar(server)->server(serve only http1.1)

howardjohn: So this PR doesn't change that case at all. But what happens today depends on the protocol declared:
- http - works
- http2 - broken
- undeclared - broken since we use original protocol

hzxuzhonghu: Yes, clear now.

howardjohn:
Ok, updated. Now just fixing the ALPN part and adding tests.

hzxuzhonghu (approved):

ramaraochavali (approved):

ericvn (approved):

PR #37216, opened 2022-02-08
gateway: allow ServiceImport as backendRef

howardjohn:
This has two behaviors:
* if clusterset.local is enabled, we change it to forward to
  clusterset.local cluster
* else, we treat it the same as Service. This is still useful, so folks
  can have portable routes between implementations


pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -658,6 +659,26 @@ func buildDestination(to k8s.BackendRef, ns, domain string) (*istio.Destination,
 			Port: &istio.PortSelector{Number: uint32(*to.Port)},
 		}, nil
 	}
+	if nilOrEqual((*string)(to.Group), features.MCSAPIGroup) && nilOrEqual((*string)(to.Kind), "ServiceImport") {
+		// Service import
+		name := fmt.Sprintf("%s.%s.svc.clusterset.local", to.Name, namespace)
+		if !features.EnableMCSHost {
+			// They asked for ServiceImport, but actually don't have full support enabled...
+			// No problem, we can just treat it as Service, which is already cross-cluster in this mode anyways
+			name = fmt.Sprintf("%s.%s.svc.%s", to.Name, namespace, domain)
```
nmittler, line 668: nit: should we use the utility methods that exist for constructing the names?

nmittler (approved):

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -658,6 +659,26 @@ func buildDestination(to k8s.BackendRef, ns, domain string) (*istio.Destination,
 			Port: &istio.PortSelector{Number: uint32(*to.Port)},
 		}, nil
 	}
+	if nilOrEqual((*string)(to.Group), features.MCSAPIGroup) && nilOrEqual((*string)(to.Kind), "ServiceImport") {
+		// Service import
+		name := fmt.Sprintf("%s.%s.svc.clusterset.local", to.Name, namespace)
+		if !features.EnableMCSHost {
+			// They asked for ServiceImport, but actually don't have full support enabled...
+			// No problem, we can just treat it as Service, which is already cross-cluster in this mode anyways
```
hzxuzhonghu, line 667: It looks weird, since it does not supporte mcs, we should skip this config

hzxuzhonghu: Basically, we should not cross-reference resources that are not first-class citizens yet. Both are still alpha and having a dependency on each other is not good. As they evolve, the behavior can change.

howardjohn: My thought process here was that users are expecting route to be portable across many implementation. So if they are using, say nginx with MCS they would need to copy all their route and change ServiceImport to Service for no real reason.

This is not about coupling things or making ServiceImport the required or recommended type to use here

PR #37219, opened 2022-02-08
Only setup the credentials if there is one

jwendell:
Otherwise the credential helpers will be installed globally and this
might mess with pulls from `gcr.io`.


howardjohn (approved):
I am not 100% sure this is right since hypothetically we might need it and get credentials some other way, but this seems the best option right now. Can always revert if it explodes

PR #37220, opened 2022-02-08
Switch default LB algorithm to LEAST_REQUEST

nmittler:
Also adding an environment variable to revert the default globally.

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -278,6 +278,13 @@ var (
 			"ENABLE_MCS_HOST also be enabled.").Get() &&
 		EnableMCSHost
 
+	EnableLegacyLBAlgorithmDefault = env.RegisterBoolVar(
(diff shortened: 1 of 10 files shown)


howardjohn (commented):
lgtm but will others review since it is changing something prettty core

ramaraochavali:
LGTM as well

howardjohn (approved):

douglas-reid (approved):
p&t owned changes seem fine. i'm not sure that increasing request counts will ultimately ensure full cross-cluster traffic, but maybe 20 reqs is enough?

nmittler:
> p&t owned changes seem fine. i'm not sure that increasing request counts will ultimately ensure full cross-cluster traffic, but maybe 20 reqs is enough?

@douglas-reid It's actually 20 * number of clusters, so in our case we have 3 clusters = 60 requests.

PR #37227, opened 2022-02-09
support Wasmplugin vmconfig

zirain:

implement of `WasmPlugin.VmConfig`

### DIFF
--- a/pilot/pkg/model/extensions.go
+++ b/pilot/pkg/model/extensions.go
@@ -11,6 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
 package model
 
 import (
@@ -18,9 +19,9 @@ import (
 	"strings"
 	"time"
 
-	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
-	envoy_extensions_filters_http_wasm_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/wasm/v3"
-	envoy_extensions_wasm_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/wasm/v3"
+	envoyCoreV3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	envoyWasmFilterV3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/wasm/v3"
+	envoyExtensionsWasmV3 "github.com/envoyproxy/go-control-plane/envoy/extensions/wasm/v3"
 	"google.golang.org/protobuf/types/known/anypb"
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
@@ -33,6 +34,8 @@ import (
 
 const (
 	defaultRuntime = "envoy.wasm.runtime.v8"
+	fileScheme     = "file"
+	ociScheme      = "oci"
 )
 
 type WasmPluginWrapper struct {
@@ -41,7 +44,7 @@ type WasmPluginWrapper struct {
 	Name      string
 	Namespace string
 
-	ExtensionConfiguration *envoy_config_core_v3.TypedExtensionConfig
+	ExtensionConfiguration *envoyCoreV3.TypedExtensionConfig
 }
 
 func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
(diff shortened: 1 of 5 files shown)


zirain:
wait https://github.com/istio/api/pull/2227

zirain:
@bianpengyuan request for a review.

pilot/pkg/model/extensions.go
```diff
@@ -127,3 +100,63 @@ func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
...
+		},
+	}
+
+	if vm != nil && len(vm.Env) != 0 {
+		hostEnvKeys := make([]string, 0, len(vm.Env))
+		inlineEnvs := make(map[string]string, 0)
+		for _, e := range vm.Env {
+			switch e.ValueFrom {
+			case extensions.EnvValueSource_INLINE:
+				if e.Value != "" {
```
bianpengyuan, line 148: I think we should allow setting an env var to an empty string?

pkg/config/validation/validation.go
```diff
@@ -3623,3 +3624,28 @@ func validateWasmPluginSHA(plugin *extensions.WasmPlugin) error {
...
+		if env.Name == "" {
+			return fmt.Errorf("invalid env name")
+		}
+
+		if keys.Contains(env.Name) {
+			return fmt.Errorf("duplicate env")
+		}
+		keys.Insert(env.Name)
+
+		if env.ValueFrom == extensions.EnvValueSource_INLINE &&
```
bianpengyuan, line 3644: Same question here. I think empty string is perfectly valid value.

pkg/config/validation/validation.go
```diff
@@ -3623,3 +3624,28 @@ func validateWasmPluginSHA(plugin *extensions.WasmPlugin) error {
...
+
+func validateWasmPluginVmConfig(vm *extensions.VmConfig) error {
+	if vm == nil {
+		return nil
+	}
+
+	keys := sets.NewSet()
+	for _, env := range vm.Env {
+		if env.Name == "" {
+			return fmt.Errorf("invalid env name")
```
bianpengyuan, line 3636: What the error message going to look like? Want to make sure the output has enough information that could point user to the `env` at Wasm config.

releasenotes/notes/37227.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: feature
+area: telemetry
+issue: []
+releaseNotes:
+  - |
+    **Added** implement of `WasmPlugin.VmConfig`.
```
bianpengyuan, line 7:
```suggestion
    **Added** environment variable support at Wasm extension via  VM configuration in WasmPlugin API.
```

pilot/pkg/model/extensions.go
zirain: why user need setup an empty string ENV?

pkg/config/validation/validation.go
zirain: it will looks like
```console
Error from server: error when creating "wasmplugin.yaml": admission webhook "validation.istio.io" denied the request: configuration is invalid: invalid env name
```

pilot/pkg/model/extensions.go
```diff
@@ -127,3 +100,63 @@ func convertToWasmPluginWrapper(plugin *config.Config) *WasmPluginWrapper {
 		ExtensionConfiguration: ec,
 	}
 }
+
+func getDataSource(u *url.URL, wasmPlugin *extensions.WasmPlugin) *envoyCoreV3.AsyncDataSource {
```
hzxuzhonghu, line 104: nit: prefer buildDataSource

pilot/pkg/model/extensions.go
hzxuzhonghu: get the  existence of an env?

pkg/config/validation/validation.go
hzxuzhonghu: +1

pkg/config/validation/validation.go
hzxuzhonghu: add the name back

bianpengyuan: > we don't do that for most resources. If I understood right that you mean to give the name/namespace of the WasmPlugin

Yeah. Any reason why we don't do that? What if there are multiple wasm plugins being updated, isn't it hard to tell which one is having problem?

pilot/pkg/model/extensions.go
zirain: updated

pilot/pkg/model/extensions.go
zirain: fixed

hzxuzhonghu (approved):

PR #37230, opened 2022-02-09
minor refactor `util/sets`

zirain:

minor refactor `util/sets`


hzxuzhonghu (approved):
maybe we should consider moving sets out of pilot

PR #37253, opened 2022-02-10
set imagePullSecrets with values.yaml

abrahamjoc:
This PR is for allowing to pull image from different registry

- [ x] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

- [ x] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

### DIFF
--- a/manifests/charts/gateway/templates/deployment.yaml
+++ b/manifests/charts/gateway/templates/deployment.yaml
@@ -27,9 +27,11 @@ spec:
         {{- end }}
         {{- include "gateway.selectorLabels" . | nindent 8 }}
     spec:
-      {{- with .Values.imagePullSecrets }}
+      {{- if .Values.global.imagePullSecrets }}
       imagePullSecrets:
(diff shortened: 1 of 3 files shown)


howardjohn (changes requested):
Why is this under global.imagePullSecrets? Its already there in `imagePullSecrets`?

abrahamjoc:
oh sorry that's true but the other charts are with global.imagePullSecrets and I just left it the same

howardjohn:
That is intended. The other charts are all nested under `global.xyz.bar` which is due to legacy reasons. This chart was explicitly intended to do things the right way.

howardjohn:
Ah yeah I see that is accidentally forgotten. Thanks

PR #18111, opened 2019-10-21
sidecar: trim service ports based on egress listeners and merge services

ramaraochavali:
This PR does two things for sidecar
1. When Egress listener specifies a port, it trims other ports of the service
2. When each egress listener produces a partial service, it merges services across listeners in the sidecar scope
[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[X ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #37257, opened 2022-02-10
Make sidecarcope services cache consistent

hzxuzhonghu:

Before: there is a diverge for default scope and sidecar scope

### DIFF
--- a/pilot/pkg/model/sidecar.go
+++ b/pilot/pkg/model/sidecar.go
@@ -183,7 +183,6 @@ func DefaultSidecarScopeForNamespace(ps *PushContext, configNamespace string) *S
 		Name:               defaultSidecar,
 		Namespace:          configNamespace,
 		EgressListeners:    []*IstioEgressListenerWrapper{defaultEgressListener},
-		services:           defaultEgressListener.services,
 		destinationRules:   make(map[host.Name]*config.Config),
 		servicesByHostname: make(map[host.Name]*Service, len(defaultEgressListener.services)),
 		configDependencies: make(map[uint64]struct{}),
@@ -194,7 +193,7 @@ func DefaultSidecarScopeForNamespace(ps *PushContext, configNamespace string) *S
 	// Now that we have all the services that sidecars using this scope (in
 	// this config namespace) will see, identify all the destinationRules
 	// that these services need
-	for _, s := range out.services {
+	for _, s := range defaultEgressListener.services {
 		if !out.addService(s) {
 			continue
 		}
--- a/pilot/pkg/model/sidecar.go
+++ b/pilot/pkg/model/sidecar.go
@@ -195,17 +195,9 @@ func DefaultSidecarScopeForNamespace(ps *PushContext, configNamespace string) *S
 	// this config namespace) will see, identify all the destinationRules
 	// that these services need
 	for _, s := range out.services {
-		// In some scenarios, there may be multiple Services defined for the same hostname due to ServiceEntry allowing
-		// arbitrary hostnames. In these cases, we want to pick the first Service, which is the oldest. This ensures
-		// newly created Services cannot take ownership unexpectedly.
-		// However, the Service is from Kubernetes it should take precedence over ones not. This prevents someone from
-		// "domain squatting" on the hostname before a Kubernetes Service is created.
-		// This relies on the assumption that
-		if existing, f := out.servicesByHostname[s.Hostname]; f &&
-			!(existing.Attributes.ServiceRegistry != provider.Kubernetes && s.Attributes.ServiceRegistry == provider.Kubernetes) {
+		if !out.addService(s) {
 			continue
 		}
-		out.servicesByHostname[s.Hostname] = s
 		if dr := ps.destinationRule(configNamespace, s); dr != nil {
 			out.destinationRules[s.Hostname] = dr
 		}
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
do you know what user facing impact this will have? Do we need upgrade notice?

pilot/pkg/model/sidecar.go
```diff
@@ -199,17 +199,9 @@ func DefaultSidecarScopeForNamespace(ps *PushContext, configNamespace string) *S
...
-		// newly created Services cannot take ownership unexpectedly.
-		// However, the Service is from Kubernetes it should take precedence over ones not. This prevents someone from
-		// "domain squatting" on the hostname before a Kubernetes Service is created.
-		// This relies on the assumption that
-		if existing, f := out.servicesByHostname[s.Hostname]; f &&
-			!(existing.Attributes.ServiceRegistry != provider.Kubernetes && s.Attributes.ServiceRegistry == provider.Kubernetes) {
+		if !out.addService(s) {
 			continue
 		}
-		out.servicesByHostname[s.Hostname] = s
```
hzxuzhonghu, line 212: there was no port merging if they are in same nses

pilot/pkg/model/sidecar.go
```diff
@@ -288,46 +331,14 @@ func ConvertToSidecarScope(ps *PushContext, sidecarConfig *config.Config, config
...
-		}
-		if foundSvc, found := servicesAdded[s.Hostname]; !found {
-			servicesAdded[s.Hostname] = s
-			out.AddConfigDependencies(ConfigKey{
-				Kind:      gvk.ServiceEntry,
-				Name:      string(s.Hostname),
-				Namespace: s.Attributes.Namespace,
-			})
-			out.services = append(out.services, s)
-		} else if foundSvc.Attributes.Namespace == s.Attributes.Namespace && s.Ports != nil && len(s.Ports) > 0 {
```
hzxuzhonghu, line 308: There was no k8s svc override se here

pilot/pkg/model/sidecar.go
howardjohn: So before with sidecar and without you got merging or not?

hzxuzhonghu: with or without the behavior is opposite. After git blame, these two mergings are not implemented at the same time.

hzxuzhonghu: Here for the default sidecarscope, no merge at all. 
And the below for sidecar, there is.

hzxuzhonghu:
ping @howardjohn

hzxuzhonghu:
@howardjohn can you take a look

pilot/pkg/model/sidecar.go
```diff
@@ -248,6 +239,55 @@ func DefaultSidecarScopeForNamespace(ps *PushContext, configNamespace string) *S
...
+		for _, p := range s.Ports {
+			found := false
+			for _, osp := range existing.Ports {
+				if p.Port == osp.Port {
+					found = true
+					break
+				}
+			}
+			if !found {
+				existing.Ports = append(existing.Ports, p)
```
howardjohn, line 278: I believe this mutates the Service which makes this sticky even when the port is later removed.

howardjohn: Note this was broken in `master` as well it appears. But still should fix it

pilot/pkg/model/sidecar.go
howardjohn: So in the before case, we did not _merge_, but we did actually keep the service. Note we don't filter `out.services`, but only out.servicesByHostname - which seems a bit weird.

So we would actually end up generating for the second Service, but not EDS which is super weird:
```
$ ik pc c shell-6874dcdd7f-htgw2 | rg shell
shell.default.svc.cluster.local                          9087      -          outbound      EDS
shell.default.svc.cluster.local                          27018     -          outbound      EDS
$ ik pc ep shell-6874dcdd7f-htgw2 | rg shell
10.244.0.53:9087                 HEALTHY     OK                outbound|9087||shell.default.svc.cluster.local
```

pilot/pkg/model/sidecar_test.go
```diff
@@ -1465,6 +1466,82 @@ func TestCreateSidecarScope(t *testing.T) {
...
+		},
+		{
+			name:     "two services from SE with same host, same namespace, same port",
+			services: []*Service{svc3, svc3},
+			expected: []*Service{svc3},
+		},
+		{
+			name:     "two services from SE with same host, same namespace, different ports",
+			services: []*Service{svc3, svc3DiffPort},
+			expected: []*Service{svc3twoPorts},
```
howardjohn, line 1516: I have some concerns about just merging port. Consider a Service with port 8080. I define a SE for the same hostname:

```
kind: ServiceEntry
metadata:
  name: shell
spec:
  hosts:
  - shell.default.svc.cluster.local
  ports:
  - number: 12345
    name: cstum
    protocol: TCP
  - number: 8080
    name: exists
    protocol: HTTP
  location: MESH_INTERNAL
  resolution: STATIC
  endpoints:
  - address: 2.2.2.2
```

Now I get a cluster on ports 12345 and 8080. But the endpoints for 8080 are just the k8s pods, not `2.2.2.2`. This seems inconstent.

It seems like we should not merge at all. It is too tricky, with different resolution modes, different protocols on the same port, etc.

It seems like the original PR (https://github.com/istio/istio/pull/18111) was actuallying doing it for synthetic Sidecar services. So something like this:
```
  egress:
  - port:
      number: 9080
      protocol: HTTP
      name: egresshttp
    hosts:
    - "prod-us1/httpbin"
  - port:
      number: 12345
      protocol: HTTP
      name: egresshttp
    hosts:
    - "prod-us1/httpbin"
```

Not actually merging 'real' Services.

cc @ramaraochavali

pilot/pkg/model/sidecar.go
hzxuzhonghu: make sense

pilot/pkg/model/sidecar_test.go
hzxuzhonghu: SE itself has already make istio very complex, as we allow creating SE with same hosts in any namespace with any ports.

Not to mention it can be used to define k8s svc as well.  Here I am not meant to solve the mess. I want to fix the inconsistent. 

Regarding to the mess, we should draft a higher design, and maybe need to break some compatibility if we want to make it clean.

hzxuzhonghu: BTW, I have talked about the merging difficulty again and again with @zirain. And we haven't come up with a better solution.

howardjohn: Totally agree, but I think in solving the inconsistency we may be expanding the wrong behavior? I think the merging was only intended to merge sidecar.egress synthetic imports , but now we fully extend it to all services which may not be great...

howardjohn: Seems like there is no real use case for merging ports in real services?

hzxuzhonghu: Agreed, if we want to remove merging ports we should have a way to warn users if they did use SE in this way.

PR #37264, opened 2022-02-10
Add ability to add extra arguments and mount extra volumes to Pilot 

### LINKED ISSUE #36916: Provide the ability to load tls cert volume and parameters for the pilot deployment for pilot.
Need to be able to add extraVolumes, extraVolumemounts and arguments to pilot when using the helm chart installation. (also within Redhat Service Mesh context which leverages that).
This is to load a certificate to pilot from different Kubernetes secrets (one for tls cert and one for a root of trust certificate).

This could work for example with keys:
"pilot.extraVolumes"
"pilot.extraVolumeMounts"
and 
"pilot.extraArgs"

These will be loaded as expected as many other charts typically do.

It is not possible for us to use overlays or the IstioOperator in this context, this is what we'd typically do in vanilla istio installations, for example: https://github.com/cert-manager/istio-csr/blob/main/hack/istio-config-1.10.0.yaml

[ ] Docs
[x] Installation
[ ] Networking
[ ] Extensions and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[x] Developer Infrastructure

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**
This is for adding support for a certificate solution to RedHat ServiceMesh, their operator leverages the helm installation method but we have no way to set required flags.

Happy to go for an alternative or direct approach to load these arguments, but I think these should be generally useful, also can submit a PR with what should be a relatively straightforward implementation.

SpectralHiss:
- These parameters are needed to load a cert-manager generated tls certificate at runtime
- This is a requirement for istio-csr installation
- While it's possible to use the override mechanism in IstioOperator some setups don't have such a facility
- resolves https://github.com/istio/istio/issues/36916

- [] Configuration Infrastructure
- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
```diff
@@ -97,6 +97,11 @@ spec:
 {{- end }}
           - --keepaliveMaxServerConnectionAge
           - "{{ .Values.pilot.keepaliveMaxServerConnectionAge }}"
+{{- if .Values.pilot.extraArgs }}
+  {{-  range .Values.pilot.extraArgs }}
+          - {{ . }}
```
howardjohn, line 102: may want to pass through `| quote` or numerical/bool values will not work

manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
```diff
@@ -217,4 +225,7 @@ spec:
         configMap:
           name: pilot-jwks-extra-cacerts{{- if not (eq .Values.revision "") }}-{{ .Values.revision }}{{- end }}
   {{- end }}
+{{- if .Values.pilot.extraVolumes }}
```
howardjohn, line 228: maybe this test could be updated to use the new fields: https://github.com/istio/istio/blob/0392abbdbf549f44fa6d093d0029e44c119ae081/tests/integration/security/file_mounted_certs/main_test.go#L140

perhaps outside of scope for this PR, but the gateway ones in the same test could also be using the existing `secretVolumes` options present the gateway chart.

adding those to the test will fail until added to `operator/pkg/apis/istio/v1alpha1/values_types.proto` as well, which is required for fields using istioctl

SpectralHiss: makes sense,  i will do the gateways one in a separate commit to not confuse things.

As for the protobuf changes, should i restrict the proto definition to the limited set of secret volumes or allow for example for configmap volumes or real volumes?
if i go for the broad version that's a lot of protos to add:
https://github.com/kubernetes/api/blob/8fb9658ecf05b1e952883268f63ba699ea1e1ad7/core/v1/generated.proto#L5450

if i go for restricted i wonder if there might be some error if people expect to (ab)use it in certain ways and the best way to show that.

SpectralHiss: I have taken your advice to use the more generic proto types, makes perfect sense in this context :+1: 
I will add the gateway bit right away.

SpectralHiss: Ok all tests passed in commit `[96c2b44](https://github.com/istio/istio/pull/37264/commits/96c2b444d6018225c63ffb8d3e00ea4c9d5d3040)`
added the gateway change , and i think that addresses everything?

SpectralHiss: looks like all tests are passing now :+1:

SpectralHiss:
Any chance for a re-review on this, looks ready my side

jacob-delgado (approved):

SpectralHiss:
Hey, from what i can see this PR was ready to go and all comments were addressed, it's important for us to get these helm changes so we don't have to maintain overlays downstream in Maistra where this is used, https://github.com/maistra/istio-operator/pull/890
On top of  being useful for the Maistra use-case this will enable a better installation  of istio-csr for helm chart installation method which we can see is becoming the standard.

Could you please let us if this PR will proceed? and if so get a re-review?

zadiman:
Hey @SpectralHiss did you face any issues with the merge? Is it something I can help you with to get things moving or is it just to get a re-review to be able to merge again?

SpectralHiss:
> Hey @SpectralHiss did you face any issues with the merge? Is it something I can help you with to get things moving or is it just to get a re-review to be able to merge again?

Just needed enough approving reviews really.. the PR was ready for merging.

zadiman:
Should we reopen the issues as the [istio-policy-bot](https://github.com/istio-policy-bot) says? That maybe give us some attention again?

SpectralHiss:
> Should we reopen the issues as the [istio-policy-bot](https://github.com/istio-policy-bot) says? That maybe give us some attention again?

Ah yeah, this would be my first contribution on here, so that's why It looks like I missed a step! 
I guess I need to "assign" someone right? not sure what to base this on other than John has been helpful from the start as always :) 

/assign @howardjohn

adabuleanu:
why wasn't this PR merged?

SpectralHiss:
> why wasn't this PR merged?

There is now a new way to make this work that doesn't require extraargs and extravolumes , cf related PR: 
`https://github.com/istio/istio/pull/39375`

adabuleanu:
thanks for the info. and this feature is available starting 1.16.x ?

PR #37277, opened 2022-02-10
chore: adds vanity import.

jcchavezs:

This PR adds vanity import to all Go files to make it easier to consume them and to search. All this codebase is hosted at github.com/istio/istio but consumed as istio.io/istio which isn't easy to find out.


howardjohn (changes requested):
Thanks for the PR. I don't see a compelling reason to do this, and have not seen it on a number of projects with vanity URLs.

The URL is clearly defined in `go.mod`. This comment syntax seems to be a legacy from old days (https://docs.google.com/document/d/1jVFkZTcYbNLaTxXD9OcGfn7vYv5hWtPx9--lTx1gPMs/edit) before go.mod.

IF we are going to do this, we certainly need automation to enforce + automatically do it. That being said, I don't think we should since I don't think it adds value nor is it an established best practice in the golang community (from what i have seen)

Arhell (commented):
/ok-to-test

PR #37286, opened 2022-02-11
feat(crdclient): populate istio crds with field path map

YLonely:

istio analyzer construct a field path map for the local yaml files to be analyzed before analyzing, it is store in the annotation field of the [Config](https://github.com/istio/istio/blob/284def0c93cc2dc37b07a1d7acd74f26eb24fb63/pkg/config/model.go#L97) struct and the error line is generated based on it

it would be useful that the analyzer generates error lines for incluster istio resources, this pr adds field path map for istio crds

### DIFF
--- a/pilot/pkg/config/kube/crdclient/gen/types.go.tmpl
+++ b/pilot/pkg/config/kube/crdclient/gen/types.go.tmpl
@@ -22,18 +22,23 @@ package crdclient
 // as declared in the Istio config model.
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"encoding/json"
 	metav1alpha1 "istio.io/api/meta/v1alpha1"
 
 	versionedclient "istio.io/client-go/pkg/clientset/versioned"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	yamlv3 "gopkg.in/yaml.v3"
+	"sigs.k8s.io/yaml"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	gatewayapiclient "sigs.k8s.io/gateway-api/pkg/client/clientset/gateway/versioned"
 
 	"istio.io/istio/pkg/config"
 	"istio.io/istio/pkg/config/schema/collections"
+	"istio.io/istio/pilot/pkg/config/file"
 
 	extensionsv1alpha1 "istio.io/api/extensions/v1alpha1"
 	networkingv1alpha3 "istio.io/api/networking/v1alpha3"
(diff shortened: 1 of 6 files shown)


YLonely:
> Haven't looked carefully, if this is a new feature, you should provide a test. It helps understanding what you are doing

ok, i will add proper tests for this feature

go.mod
```diff
@@ -33,6 +33,7 @@ require (
 	github.com/fatih/color v1.13.0
 	github.com/florianl/go-nflog/v2 v2.0.1
 	github.com/fsnotify/fsnotify v1.5.1
+	github.com/ghodss/yaml v1.0.0
```
howardjohn, line 36: Please do not use this library.

pilot/pkg/config/kube/crdclient/gen/types.go.tmpl
```diff
@@ -22,18 +22,23 @@ package crdclient
...
+	"encoding/json"
 	metav1alpha1 "istio.io/api/meta/v1alpha1"
 
 	versionedclient "istio.io/client-go/pkg/clientset/versioned"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	yamlv3 "gopkg.in/yaml.v3"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	gatewayapiclient "sigs.k8s.io/gateway-api/pkg/client/clientset/gateway/versioned"
+	"github.com/ghodss/yaml"
```
howardjohn, line 37: as mentioned, do not need 2 yaml libraries. Use `gopkg.in/yaml.v3` in very rare cases when we need complex things and sigs.k8s.io/yaml for everything else

pilot/pkg/config/kube/crdclient/gen/types.go.tmpl
```diff
@@ -152,17 +157,29 @@ func delete(ic versionedclient.Interface, sc gatewayapiclient.Interface, typ con
...
+		{{- if .WithFieldPathMap }}
+		annots := obj.Annotations 
+		fieldPathMap, err := buildFieldPathMap(obj)
+		if err != nil {
+			scope.Errorf("failed to build field path map for object %s/%s: %v", obj.Namespace, obj.Name, err)
+		} else {
+			if annots == nil {
+				annots = map[string]string{}
+			}
+			annots[file.FieldMapKey] = fieldPathMap
```
howardjohn, line 173: I don't see why this "field map" belongs on an annotation

howardjohn (changes requested):
Is there any context/design docs/tests? Its not clear what this does at all

go.mod
YLonely: library removed

pilot/pkg/config/kube/crdclient/gen/types.go.tmpl
YLonely: package removed

pilot/pkg/config/kube/crdclient/gen/types.go.tmpl
YLonely: it is the same trick that when the istio analyzer analyses a local resource file, it constructs the "field map" for the resource and stores it in the annotation(https://github.com/istio/istio/blob/284def0c93cc2dc37b07a1d7acd74f26eb24fb63/pilot/pkg/config/file/store.go#L433)

YLonely:
> Haven't looked carefully, if this is a new feature, you should provide a test. It helps understanding what you are doing

test provided

pkg/config/analysis/local/analyze_test.go
```diff
@@ -162,6 +169,85 @@ func TestAddRunningKubeSource(t *testing.T) {
...
+	if _, err := mk.Istio().NetworkingV1alpha3().VirtualServices(namespace).Create(context.TODO(), vs, metav1.CreateOptions{}); err != nil {
+		t.Fatalf("Error creating virtual service: %v", err)
+	}
+	sa := NewSourceAnalyzer(analysis.Combine("vs", &vsanalyzer.DestinationHostAnalyzer{}), resource.Namespace(namespace), "", nil, true, timeout)
+	sa.AddRunningKubeSource(mk)
+	result, err := sa.Analyze(cancel)
+	if err != nil {
+		t.Fatal(err)
+	}
+	line, err := locateErrorLine(mk, namespace, name, "host")
```
howardjohn, line 215: why not just write out the string? Otherwise we don't know if the test code is correct or not

howardjohn (changes requested):
Sorry, I do not understand at all. How can you get "line number" from configs? There are no line numbers... I can write
```
foo: bar
baz: bana
```

or

```
baz: bana


foo: bar
```

and they will show up the same in kubernetes.

If we want to do this, please do not put it in config.Config. This is a highly optimized structure use to store all configs, we do not want to bloat it for analyze.

PR #37290, opened 2022-02-11
add support for start tls mode

ramaraochavali:
As discussed in https://github.com/istio/api/pull/2196 - adds support for start tls for postgres protocol
Fixes https://github.com/istio/istio/issues/29761
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ericvn (approved):

pkg/config/protocol/instance.go
```diff
@@ -85,6 +87,8 @@ func Parse(s string) Instance {
 		return Redis
 	case "mysql":
 		return MySQL
+	case "postgres":
```
howardjohn, line 90: need to make it opt-in with an env var, like mysql and others

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1002,6 +1013,10 @@ func (cb *ClusterBuilder) applyUpstreamTLSSettings(opts *buildClusterOpts, tls *
 	}
 }
 
+func startTLSProtocol(port *model.Port) bool {
+	return port != nil && port.Protocol == protocol.Postgres
```
treyhyde, line 1017: bot 5432 and 6432 (pg bouncer convention) are common in pg world

hobbytp, line 1017: does that mean only postgres use startTLS?

pkg/config/protocol/instance.go
ramaraochavali: in mysql and others - we are actually adding the specific filters like mysql. Here we are not adding them. Do you still want to control whether startTLS is added based on that?

pkg/config/protocol/instance.go
```diff
@@ -50,6 +50,8 @@ const (
 	Redis Instance = "Redis"
 	// MySQL declares that the port carries MySQL traffic.
 	MySQL Instance = "MySQL"
+	// Postgres declares that the port carries Postgres traffic.
+	Postgres Instance = "Postgres"
```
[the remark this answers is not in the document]
ramaraochavali: We can add that in the release notes. Adding postgres filter is not trivial as it is in contrib. Do you have any alternate suggestions?

pkg/config/protocol/instance.go
howardjohn: I think it may be backwards incompatible otherwise? Plus, I would guess in the future we would add the filters(?)

ramaraochavali: Sure. that makes sense. I can add that flag

pkg/config/protocol/instance.go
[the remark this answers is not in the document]
hzxuzhonghu: Wondering we are just removing thrift but adding postgres?  `postgress` feature is still not completely supported, how do you think about that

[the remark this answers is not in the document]
ramaraochavali: We are not adding full postgres support yet. We are just adding startTLS functionality needed for postgres as discussed https://github.com/istio/api/pull/2196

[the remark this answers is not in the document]
hzxuzhonghu: Very confusing(after first read, think it is Postgres support), and also there is a very corner case that need this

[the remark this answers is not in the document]
ramaraochavali: The PR description mentions clearly that it is startTLS support - not full postgres support. And this is not a corner case, without this - we can not use postgres with Istio (even with tcp protocol). We had to create envoy filter to do this.

howardjohn:
can't you use postgres, just not tls origination?

ramaraochavali:
> can't you use postgres, just not tls origination?

You mean postgres filter? It is not in main envoy - it is in contrib at this point.

ramaraochavali:
> can't you use postgres, just not tls origination?

Oh you mean just use it on server?

ramaraochavali:
> No I mean postgres should work fine, right? The only issue is if you try to use DestinationRule SIMPLE TLS origination?
Yes

ramaraochavali:
@howardjohn @hzxuzhonghu So how do we want to proceed on this? I am certainly open for suggestions.

ramaraochavali:
@howardjohn @hzxuzhonghu what do you think about this? How should we proceed?

hzxuzhonghu:
TBH, i donot feel strongly about such a hacky fix for a so small scenario. envoyfilter is intended for such a rare use case

hzxuzhonghu:
I am not saying starttls is not a standard protocol, I am concerned it is strongly coupled with Postgres

hzxuzhonghu:
Once we make a hacky implement in, we can hardly remove without breaking compatibility

ramaraochavali:
>  I am not saying starttls is not a standard protocol, I am concerned it is strongly coupled with Postgres

But that is what was agreed here https://github.com/istio/api/pull/2196 - we will base it on protocol.

Do you have any better ideas to delink it from protocol and add this support? I would be happy to change.

hzxuzhonghu:
Sorry for the delay, I feel it should be in API as it is common and should not be coupled with protocol.

ramaraochavali:
@costinm @howardjohn WDYT?

hobbytp:
seems different application protocols use STARTTLS with different flows (different flow before launching real TLS, it looks like a common protocol and it can be adapted to different application protocol. refer https://en.wikipedia.org/wiki/Opportunistic_TLS), so maybe it will be better to put the implementation to protocol? It does not mean coupling postgres with starttls, it means we focus on postgres first even it also need to configure starttls.

hzxuzhonghu:
> seems different application protocols use STARTTLS with different flows

If this is the case, I am confused how envoy does starttls without knowing the protocol/no protocol-specific filter configured

ramaraochavali:
> If this is the case, I am confused how envoy does starttls without knowing the protocol/no protocol-specific filter configured

That is why Envoy does not do it by default till we ask it to do for the protocols we know need startTLS. That is why they have added that additional support for controlplanes to configure it. 
See this postgres integration test https://github.com/envoyproxy/envoy/blob/ef82a02bed0e6ab3de45e87b365cb18b55be6010/contrib/postgres_proxy/filters/network/test/postgres_integration_test.cc#L28 - The config is being added explicitly

hzxuzhonghu:
To be clear: there is no postgres filter, how could starttls work for Postgres? I am asking because you say startls flow is different for different protocols

hobbytp:
> To be clear: there is no postgres filter, how could starttls work for Postgres? I am asking because you say startls flow is different for different protocols

good point.
For downstream starttls config, it has to be paired with a network filter to support it (e.g. postgres_filter)
For upstream, seems there is not related postgres filter to do that job. refer comment https://github.com/envoyproxy/envoy/pull/15443/files#r603453529, if there is not anything changed, does that mean upstream socket can not support postgres protocol now?  Copy @lizan for comments

ramaraochavali:
it works because postgres JDBC driver -> talks to Envoy -> Postgres server. Envoy does not have to have postgres filter. The driver implements postGres protocol. Envoy treats it as TCP and originates TLS. This is the flow we have tested with EnvoyFilter injecting the startTLS config.

hobbytp:
> it works because postgres JDBC driver -> talks to Envoy -> Postgres server. Envoy does not have to have postgres filter. The driver implements postGres protocol. Envoy treats it as TCP and originates TLS. This is the flow we have tested with EnvoyFilter injecting the startTLS config.

Thanks for clarification, so starttls flow looks as below? any comments? 
client --------SSLReq----->envoy(upstream)  ---SSLReq---> PG    (envoy forward SSLReq to upstream PG)
client <---------"N"--------envoy(upstream) <---"S"-------- PG    (PG wants to TLS, so it send back "S" byte. )
client -------plaintext---->envoy(upstream) <----TLS HS--->PG    (envoy tls origin TLS when client send plaintext Startup pkt)
client <--Res/plaintext--- envoy(upstream)  <---Res/TLS-- PG      (PG TLS response envoy in client )

I am not sure about step 2, but it can explain why client will use plaintext in later flow.

ramaraochavali:
Yes I think that flow sounds right to me

lizan:
> > To be clear: there is no postgres filter, how could starttls work for Postgres? I am asking because you say startls flow is different for different protocols
> 
> good point. For downstream starttls config, it has to be paired with a network filter to support it (e.g. postgres_filter) For upstream, seems there is not related postgres filter to do that job. refer comment https://github.com/envoyproxy/envoy/pull/15443/files#r603453529, if there is not anything changed, does that mean upstream socket can not support postgres protocol now? Copy @lizan for comments

That's correct. We don't have upstream socket support of Postgres protocol and there is no plan for that. However the upstream socket need to be configured with supporting filter, which doesn't exist in upstream Envoy today, so we shouldn't configure that.

hzxuzhonghu:
From the comments, i get the conclusion： upstream starttls is not relevant to postgres, so why it has to be coupled with postgres

ramaraochavali:
> upstream starttls is not relevant to postgres, so why it has to be coupled with postgres

If you have to originate TLS from proxy to protocls that expect startTLS like postgres, IMAP - you need startTLS to be configured for upstream. So  it is not very specific to just one protocol, we need to do it for all protocols that need this.

There are couple of ways we can do
- Istio knows which protocols need that and configure it for that protocol if DR specifies that TLS needs to be used (this PR)
- Do not base on protocol but let DR tell to configure start TLS for upstream connections. Then it becomes service owners responsibility to configure it. That is what this PR tried to do https://github.com/istio/api/pull/2196

But the discussion in api PR suggested to use protocol.

I am fine with either approach or any other approach that solves this problem.

ramaraochavali:
That was my original proposal to add this to DR but @costinm was suggesting it to tie to protocol (which kind of made sense to me as well) because if a protcol needs it, why should user explicitly specify it). Having said that I am open to that as well, but I think configuring part of TLS via Istio and asking users to add startTLS (with a complicated envoy filter because it needs to have complete tls settings) is not correct.

hobbytp:
> > it works because postgres JDBC driver -> talks to Envoy -> Postgres server. Envoy does not have to have postgres filter. The driver implements postGres protocol. Envoy treats it as TCP and originates TLS. This is the flow we have tested with EnvoyFilter injecting the startTLS config.
> 
> Thanks for clarification, so starttls flow looks as below? any comments? 
> client --------SSLReq----->envoy(upstream) ---SSLReq---> PG (envoy forward SSLReq to upstream PG) 
> client <---------"N"--------envoy(upstream) <---"S"-------- PG (PG wants to TLS, so it send back "S" byte. ) 
> client -------plaintext---->envoy(upstream) <----TLS HS--->PG (envoy tls origin TLS when client send plaintext Startup pkt) 
> client <--Res/plaintext--- envoy(upstream) <---Res/TLS---- PG (PG TLS response envoy in client )
> 
> I am not sure about step 2, but it can explain why client will use plaintext in later flow.

So from lizan's comment, I think currently the step2 is not supported by envoy proxy for there is not any Postgres Proxy to support upstream/client logic (switch "S" to "N" to client and switch transport socket from plaintext to Tls), the existed [Postgres Proxy](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/postgres_proxy_filter#config-network-filters-postgres-proxy) only support downstream/server side. 

So my question now is, which of the following envoy proxy configuration will this PR support?
1. Only starttls for both downstream and upstream configuration without any network filter. This in fact does not support switching from plaintext to tls.
2. downstream startls + existed Postgres Proxy. This only supports server side postgre protocol.

ramaraochavali:
upstream startls -This is what This PR supports. We can not support postgres filter in Istio directly because it is not part of main envoy build. It is still in contrib

hobbytp:
@ramaraochavali 
My client is also Postgres JDBC driver, but seems it always send out SSL Request first, and then when it receive the "S" from Postgres, it will try to launch TLS handshake by itself instead of letting sidecar to do TLS origination. After that, the TLS handshake fails for the client (not the sidecar) does not own any certificate. The TCPDump pcap shows the client certificate size is 0 which cause the TLS handshake failure.

So my question is what is the difference between your postgres JDBC driver and mine? 
And my envoyfilter is as below, could you please help to check if it is correct?
```yaml
kind: EnvoyFilter
metadata:
  name: starttls-pg
spec:
  workloadSelector:
   labels:
     app: notification-service
  configPatches:
  - applyTo: CLUSTER
    match:
      cluster:
        portNumber: 5432
        service: notification-service-database-pg.moon.svc.cluster.local
      context: SIDECAR_OUTBOUND
    patch:
      operation: MERGE
      value:
        transport_socket:
          name: envoy.transport_sockets.starttls
          typed_config:
            '@type': type.googleapis.com/envoy.extensions.transport_sockets.starttls.v3.UpstreamStartTlsConfig
            tls_socket_config:
              allow_renegotiation: true
              common_tls_context:
                tls_certificate_sds_secret_configs:
                - name: 'file-cert:/etc/istio/tls-test/cert.pem~/etc/istio/tls-test/key.pem'
                  sds_config:
                    api_config_source:
                      api_type: GRPC
                      transport_api_version: V3
                      grpc_services:
                      - envoy_grpc:
                          cluster_name: sds-grpc
                      set_node_on_first_message_only: true
                    resource_api_version: V3
                combined_validation_context:
                  default_validation_context: {}
                  validation_context_sds_secret_config:
                    name: 'file-root:/etc/istio/tls-ca-test/cacertbundle.pem'
                    sds_config:
                      api_config_source:
                        api_type: GRPC
                        transport_api_version: V3
                        grpc_services:
                        - envoy_grpc:
                            cluster_name: sds-grpc
                        set_node_on_first_message_only: true
                      resource_api_version: V3

```

hobbytp:
Our team changes the JDBC url as "jdbc:postgresql://database-pg:5432/catalog?sslmode=disable" to disable SSLRequest message, and then apply the DestinationRule and EnvoyFilter as below, but the result is : no SSL request is sent by client which is what I want , but also not any TLS packets exchange between server (PG) and client (with sidecar injected) pod, from the tcpdump in PG side, the client (with sidecar) sends out "Start Up" message to PG directly without TLS origination first. Shall I change the PG service port (5432) name with prefix "tls-xxx" also? PG is not in mesh currently.

```yaml
kind: EnvoyFilter
metadata:
  name: starttls-pg
spec:
  workloadSelector:
   labels:
     app: notification-service
  configPatches:
  - applyTo: CLUSTER
    match:
      cluster:
        portNumber: 5432
        service: notification-service-database-pg.moon.svc.cluster.local
      context: SIDECAR_OUTBOUND
    patch:
      operation: MERGE
      value:
        transport_socket:
          name: envoy.transport_sockets.starttls
          typed_config:
            '@type': type.googleapis.com/envoy.extensions.transport_sockets.starttls.v3.UpstreamStartTlsConfig
            tls_socket_config:
              allow_renegotiation: true
              common_tls_context:
                tls_certificate_sds_secret_configs:
                - name: 'file-cert:/etc/istio/tls-test/cert.pem~/etc/istio/tls-test/key.pem'
                  sds_config:
                    api_config_source:
                      api_type: GRPC
                      transport_api_version: V3
                      grpc_services:
                      - envoy_grpc:
                          cluster_name: sds-grpc
                      set_node_on_first_message_only: true
                    resource_api_version: V3
                combined_validation_context:
                  default_validation_context: {}
                  validation_context_sds_secret_config:
                    name: 'file-root:/etc/istio/tls-ca-test/cacertbundle.pem'
                    sds_config:
                      api_config_source:
                        api_type: GRPC
                        transport_api_version: V3
                        grpc_services:
                        - envoy_grpc:
                            cluster_name: sds-grpc
                        set_node_on_first_message_only: true
                      resource_api_version: V3
```
```yaml
kind: DestinationRule
metadata:
  name: pg-dr
spec:
  host: notification-service-database-pg
  trafficPolicy:
    portLevelSettings:
    - port:
        number: 5432
      tls:
        mode: MUTUAL
        clientCertificate: /etc/istio/tls-test/cert.pem
        privateKey: /etc/istio/tls-test/key.pem
        caCertificates: /etc/istio/tls-ca-test/cacertbundle.pem
```

ramaraochavali:
Can you please link the correct stackoverflow link?

ramaraochavali:
`sslmode=disabled` for client seems good. That is what we also did. And for us also postgres is not on Mesh. We use 5432.

Our startTLS envoy filter is 
```
                  "operation": "MERGE",
                 "value": {
                   "transport_socket": {
                     "name": "envoy.transport_sockets.starttls",
                     "typed_config": {
                       "@type": "type.googleapis.com/envoy.extensions.transport_sockets.starttls.v3.UpstreamStartTlsConfig",
                         "tls_socket_config": {
                           "allow_renegotiation": true,
                         }
                     }
                   }
                 }
```

Destination rule just has this - because we use simple TLS as it is aws managed rds

```
      trafficPolicy: {
        tls: {
          mode: tlsMode,
          caCertificates: certLocation,
        }
      }
```

Can you try this?

hobbytp:
@ramaraochavali from my previous testing, the envoy proxy does not do any TLS origination. So I doubt if it is correct to use UpstreamStartTlsConfig which seems need to do socket switching. But in fact, in this case, we don't need any socket switch, the only thing the envoy proxy need to do is start executing TLS origination directly when it receive "Start Up" message from client, and then forward this "Start up" message to PG in this TLS tunnel.  So maybe I need to use Tls socket directly instead of StartTls.

ramaraochavali:
> So maybe I need to use Tls socket directly instead of StartTls, and change the PG port name like "tcp-xxx"? So the envoy proxy can trigger TLS origination?

We use service entry to point to AWS managed postgres and we name the port is 5432 as tcp-

hobbytp:
@ramaraochavali 
Have you captured the pkt by tcpdump tool? How can you be sure the communication is TLS instead of plaintext? Today I install bitnami/postgresql as below, but seems the PG still accept plaintext "Start Up" and then continue with following plaintext "Authentication Request", "Password message", I use psql with sslmode=disable for this test, and I apply startTlsUpstream by envoyfilter, I think only the clean-text socket in startTlsSocket work in the whole flow.

```
helm upgrade --install my-release bitnami/postgresql \
--set volumePermissions.enabled=true \
--set tls.enabled=true \
--set tls.certificatesSecret="certificates-tls-secret" \
--set tls.certFilename="my-release-postgresql.pg.svc.cluster.local.crt" \
--set tls.certKeyFilename="my-release-postgresql.pg.svc.cluster.local.key" \
--set tls.certCAFilename="example.com.crt" \
 -n pg
```

ramaraochavali:
> . I think only the clean-text socket in startTlsSocket works in the whole flow.

Sorry. I did not understand. Can you share the filter?

ramaraochavali:
As we do not have consensus on what to do here, I am closing this PR.

iprasla:
Can this be reopen?

PR #37314, opened 2022-02-13
fix locking issue in ads

ramaraochavali:
WatchedResources is accessed without lock. this PR fixes it.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu:
I donot think this has effect, the map is still read without lock

PR #37316, opened 2022-02-13
fix type assertion

sivchari:
Fix to do type assertion.

### DIFF
--- a/pilot/pkg/config/file/store.go
+++ b/pilot/pkg/config/file/store.go
@@ -366,7 +366,7 @@ func (s *KubeSource) parseChunk(r *collection.Schemas, name string, lineNum int,
 	}
 	objMeta, ok := obj.(metav1.Object)
 	if !ok {
-		return kubeResource{}, errors.New("failed to type assertion")
+		return kubeResource{}, errors.New("failed to assert type of object metadata")
 	}
 
 	// If namespace is blank and we have a default set, fill in the default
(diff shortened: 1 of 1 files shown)


pilot/pkg/config/file/store.go
```diff
@@ -364,7 +364,10 @@ func (s *KubeSource) parseChunk(r *collection.Schemas, name string, lineNum int,
 	if err != nil {
 		return kubeResource{}, fmt.Errorf("failed parsing JSON for built-in type: %v", err)
 	}
-	objMeta := obj.(metav1.Object)
+	objMeta, ok := obj.(metav1.Object)
+	if !ok {
+		return kubeResource{}, errors.New("failed to type assertion")
```
bianpengyuan, line 369:
```suggestion
		return kubeResource{}, errors.New("failed to assert type of object metadata")
```

bianpengyuan (commented):
/ok-to-test

pilot/pkg/config/file/store.go
sivchari: I fixed it.

pilot/.golangci.yml
```diff
@@ -0,0 +1,61 @@
+linters-settings:
```
bianpengyuan, line 1: accidental check in?

sivchari: Oh, sorry. I fixed it.

bianpengyuan (approved):

PR #37319, opened 2022-02-14
do not compare cert bytes directly

irisdingbj:
We should not compare the root certs bytes directly, instead we can rely on the https://github.com/istio/istio/compare/master...irisdingbj:verifyfix?expand=1#diff-df16ac4e2cc1c048a4ce70aa3f27ec9c173573fc5288fb029fdc19b93c0ae259L133 to make sure the cert chain retured is correctly signed from root cert.

### DIFF
--- a/security/pkg/pki/ra/k8s_ra.go
+++ b/security/pkg/pki/ra/k8s_ra.go
@@ -15,7 +15,6 @@
 package ra
 
 import (
-	"bytes"
 	"fmt"
 	"strings"
 	"sync"
@@ -116,20 +115,11 @@ func (r *KubernetesRA) SignWithCertChain(csrPEM []byte, certOpts ca.CertOpts) ([
 		if err != nil {
 			return nil, fmt.Errorf("failed to find root cert from mesh config (%v)", err.Error())
 		}
-		if rootCertFromCertChain != nil && rootCertFromMeshConfig != nil {
-			if !bytes.Equal(rootCertFromCertChain, rootCertFromMeshConfig) {
-				return nil, fmt.Errorf("root cert from signed cert-chain" +
-					" is conflicting with the one specified in mesh config")
-			}
-		}
 		if rootCertFromMeshConfig != nil {
 			rootCert = rootCertFromMeshConfig
-		}
-
-		if rootCertFromCertChain != nil {
+		} else if rootCertFromCertChain != nil {
 			rootCert = rootCertFromCertChain
 		}
-
 		if verifyErr := util.VerifyCertificate(nil, cert, rootCert, nil); verifyErr != nil {
 			return nil, fmt.Errorf("root cert from signed cert-chain is invalid %v ", verifyErr)
 		}
--- a/tests/integration/security/external_ca/main_test.go
+++ b/tests/integration/security/external_ca/main_test.go
@@ -77,7 +77,6 @@ func SetupApps(ctx resource.Context, apps *EchoDeployments) error {
 func TestMain(m *testing.M) {
 	// Integration test for testing interoperability with external CA's that are integrated with K8s CSR API
 	// Refer to https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/
-	// nolint: staticcheck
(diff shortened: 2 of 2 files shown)


security/pkg/pki/ra/k8s_ra.go
```diff
@@ -116,24 +116,17 @@ func (r *KubernetesRA) SignWithCertChain(csrPEM []byte, certOpts ca.CertOpts) ([
...
-				return nil, fmt.Errorf("root cert from signed cert-chain" +
-					" is conflicting with the one specified in mesh config")
-			}
-		}
 		if rootCertFromMeshConfig != nil {
 			rootCert = rootCertFromMeshConfig
-		}
-
-		if rootCertFromCertChain != nil {
+		} else if rootCertFromCertChain != nil {
```
howardjohn, line 121: this changes the behavior. before cert chain > mesh config, now its the inverse?

howardjohn (commented):
I don't think I really understand this, maybe @shankgan does

security/pkg/pki/ra/k8s_ra.go
```diff
@@ -116,24 +116,17 @@ func (r *KubernetesRA) SignWithCertChain(csrPEM []byte, certOpts ca.CertOpts) ([
 		if err != nil {
 			return nil, fmt.Errorf("failed to find root cert from mesh config (%v)", err.Error())
 		}
-		if rootCertFromCertChain != nil && rootCertFromMeshConfig != nil {
```
shankgan, line 119: Some comments here would be super helpful. IMO, this logic is quite hard to read. Specifically, 

- IF there is a caCert specified in meshConfig, why we are we expecting that the CertChain of the signed cert from the external CA is signed by it
- What are the alternatives here - it appears that either the rootCert can be present in the certChain or not. What are the relevant scenarios where these could happen. Why does the MeshConfig root take precedence over the root in the certChain. 
- If we are extracting the "root" from the CertChain bytes, how come we are appending another "root" to it from MeshConfig. 
-

shankgan: the logic does look correct to me though, still wondering if more comments + refactoring would make this more readable

security/pkg/pki/ra/k8s_ra_test.go
```diff
@@ -161,6 +161,41 @@ func TestK8sSignWithMeshConfig(t *testing.T) {
...
+		SubjectIDs: []string{subjectID},
+		TTL:        60 * time.Second, ForCA: false,
+		CertSigner: "kube-apiserver-client",
+	}
+	// expect to sign back successfully
+	_, err = ra.SignWithCertChain(csrPEM, certOptions)
+	if err != nil {
+		t.Errorf("K8s CA Signing CSR With Root Cert In Meshconfig failed")
+	}
+}
```
shankgan, line 197: Would be good to add a few more test cases here - 
(a) Scenario where certificate received from external CA is signed by a different root compared to meshConfig root and therefore certificate verification fails
(b)  Scenario where certificate received from external CA is signed by an intermediate cert that chains to the common root in the Meshconfig. IMO, this is what the newer code changes are for correct?

irisdingbj: added more test cases

irisdingbj:
@howardjohn @shankgan  Thanks a lot for the review comments.  I have added comments to explain the possible scenarios in the code and also added more test cases to cover these secnarios.    The new code are majorly to handle the situation that cert chain contains only intermediate CA. So we can not compare bytes directly. PTAL and let me know if the current rules are good.

security/pkg/pki/ra/k8s_ra.go
irisdingbj: We now will verify cert-chain against the root cert in meshconfig not simply comare bytes. - in some cases, the root cert in cert-chain might be only an intermediate ca.

shankgan (approved):

shankgan:
@irisdingbj - looks good with the additional comments and the test cases

PR #37320, opened 2022-02-14
fix map locking issue in ads properly

ramaraochavali:
Signed-off-by: Rama Chavali <rama.rao@salesforce.com>
Follow up of #37314 

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -300,9 +300,6 @@ type Proxy struct {
 	// WatchedResources contains the list of watched resources for the proxy, keyed by the DiscoveryRequest TypeUrl.
 	WatchedResources map[string]*WatchedResource
 
-	// WatchedResourcesList contains the ordered list of watched resources for the proxy, ordered in accordance with known push order.
-	WatchedResourcesList []*WatchedResource
-
 	// XdsNode is the xDS node identifier
 	XdsNode *core.Node
 
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -300,6 +300,9 @@ type Proxy struct {
 	// WatchedResources contains the list of watched resources for the proxy, keyed by the DiscoveryRequest TypeUrl.
 	WatchedResources map[string]*WatchedResource
 
+	// WatchedResourcesList contains the ordered list of watched resources for the proxy, ordered in accordance with known push order.
+	WatchedResourcesList []*WatchedResource
+
 	// XdsNode is the xDS node identifier
 	XdsNode *core.Node
 
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -696,13 +696,9 @@ func (s *DiscoveryServer) pushConnection(con *Connection, pushEv *Event) error {
 		return nil
 	}
 
-	con.proxy.RLock()
-	wr := con.proxy.WatchedResources
-	con.proxy.RUnlock()
-
 	// Send pushes to all generators
 	// Each Generator is responsible for determining if the push event requires a push
-	for _, w := range orderWatchedResources(wr) {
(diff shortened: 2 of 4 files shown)


ramaraochavali:
@hzxuzhonghu PTAL

pilot/pkg/xds/ads.go
```diff
@@ -992,6 +987,24 @@ func (conn *Connection) Watched(typeUrl string) *model.WatchedResource {
 	return nil
 }
 
+// eventTypes return the event types that need to pushed when config changes.
+func (conn *Connection) eventTypes() []EventType {
```
hzxuzhonghu, line 991: The function does not return expected all types, so looks confusing. Maybe `AllEventTypes` is not right. @howardjohn

ramaraochavali: Sorry I did not understand. This is to support this call https://github.com/istio/istio/blob/master/pilot/pkg/xds/ads.go#L694 where we need to push all eventTypes.

pilot/pkg/xds/ads.go
```diff
@@ -696,13 +696,9 @@ func (s *DiscoveryServer) pushConnection(con *Connection, pushEv *Event) error {
...
 	}
 
-	con.proxy.RLock()
-	wr := con.proxy.WatchedResources
-	con.proxy.RUnlock()
-
 	// Send pushes to all generators
 	// Each Generator is responsible for determining if the push event requires a push
-	for _, w := range orderWatchedResources(wr) {
+	for _, w := range orderWatchedResources(con) {
```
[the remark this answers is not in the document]
ramaraochavali: The problem is it is read without lock earlier and ordered resources might be different. I see your point though. I think in both cases we just need typeUrls. Let me refactor this a bit.

ramaraochavali:
@howardjohn @hzxuzhonghu refactored a bit more so that we order resources only when we add a new typeUrl not on every push. PTAL.

pilot/pkg/model/context.go
```diff
@@ -300,6 +300,9 @@ type Proxy struct {
 	// WatchedResources contains the list of watched resources for the proxy, keyed by the DiscoveryRequest TypeUrl.
 	WatchedResources map[string]*WatchedResource
 
+	// WatchedResourcesList contains the ordered list of watched resources for the proxy, ordered in accordance with known push order.
+	WatchedResourcesList []*WatchedResource
```
hzxuzhonghu, line 304: I donot like adding a new confusing field to Proxy, this struct is already very complex

ramaraochavali: What is confusing here? Do you want to order it every time? If that is the preference I can change it back. @howardjohn WDYT?

ramaraochavali: Since you mentioned this is confusing, I just reverted this part of the change and doing ordering every time as we were doing earlier. PTAL.

howardjohn (approved):

hzxuzhonghu:
Why remove, we set one actually. Allows service select wle, but not allow se selects pods

ramaraochavali:
> Why remove, we set one actually. Allows service select wle, but not allow se selects pods

Is this for a  differentPR?

hzxuzhonghu:
Sorry, for the env removal

ramaraochavali:
Ok. Why do you have to disable this "but not allow se selects pods"? Just curious what is the problem with it?

hzxuzhonghu:
There is a duplicate if we enable both

ramaraochavali:
> There is a duplicate if we enable both

Should we fix that? Or do you want me reintroduce that flag?

hzxuzhonghu:
bring the flag back sgtm. Because it is not easy to fix.

PR #37325, opened 2022-02-14
[TF] Prevent concurrent istioctl invoke

nmittler:

### DIFF
--- a/pkg/test/framework/components/istioctl/kube.go
+++ b/pkg/test/framework/components/istioctl/kube.go
@@ -18,6 +18,8 @@ import (
 	"bytes"
 	"fmt"
 	"strings"
+	"sync"
+	"time"
 
 	"istio.io/istio/istioctl/cmd"
 	"istio.io/istio/pilot/pkg/config/kube/crd"
@@ -25,6 +27,10 @@ import (
 	"istio.io/istio/pkg/test/framework/resource"
 	"istio.io/istio/pkg/test/scopes"
 )
+
+// We cannot invoke the istioctl library concurrently due to the number of global variables
+// https://github.com/istio/istio/issues/37324
+var invokeMutex sync.Mutex
 
 type kubeComponent struct {
 	config     Config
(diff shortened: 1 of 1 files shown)


pkg/test/framework/components/istioctl/kube.go
```diff
@@ -26,6 +27,12 @@ import (
 	"istio.io/istio/pkg/test/scopes"
 )
 
+var (
+	// We cannot invoke the istioctl library concurrently due to the number of global variables
+	// https://github.com/istio/istio/issues/37324
+	invokeMutex sync.Mutex
```
howardjohn, line 33: at one point I thought we installed multi clusters in parallel. do we not do that already, or does this PR stop making that happen?

howardjohn (commented):
lgtm if it works. Was there issues?

nmittler:
@howardjohn I'm doing this purely as a reaction to realizing that this is fundamentally broken. If we are currently doing things in parallel, I suspect it's an issue waiting to happen.

@stevenctl any concerns with this?

stevenctl:
+1, There is a lot of global use in istioctl.

pkg/test/framework/components/istioctl/kube.go
stevenctl: Good point..

nmittler: @stevenctl  How is our installation safe if we're invoking istioctl concurrently?

nmittler:
@stevenctl It looks like `integ-pilot-multicluster_istio` takes about 5 min longer with this PR. If that's a concern, let's discuss possible work-arounds.

pkg/test/framework/components/istioctl/kube.go
stevenctl: It's not.. but this part also doesn't seem to be horribly affected. 

We could also workaround by having a separate `Invoke` impl that execs the istioctl binary. Downside is things are harder to debug (we can control the impl with a flag?)

howardjohn:
> It's not.. but this part also doesn't seem to be horribly affected.
> 
> We could also workaround by having a separate Invoke impl that execs the istioctl binary. Downside is things are harder to debug (we can control the impl with a flag?)

It also makes it harder to run since the `go test` is no longer hermetic - depends on istioctl being built correctly, in the path, and not the wrong version, etc. Would prefer to avoid if possible

nmittler:
@howardjohn 

> It also makes it harder to run since the go test is no longer hermetic - depends on istioctl being built correctly, in the path, and not the wrong version, etc. Would prefer to avoid if possible

+1 ... we specifically moved to the in-memory model because calling out to the command line for istioctl was incredibly confusing and error-prone when running locally.

stevenctl:
My vote is to turn off the locking mechanism for install only, at least until we determine things area breaking there as well.

nmittler:
@stevenctl 

> My vote is to turn off the locking mechanism for install only, at least until we determine things area breaking there as well.

I've already changed the installation code (#37332) to avoid using `istioctl.Invoke`, so this *should* be safe as-is.

stevenctl:
This lgtm then.

stevenctl (approved):

nmittler:
@stevenctl FYI with https://github.com/istio/istio/pull/37332 in place, this change seems to have virtually no impact on job run duration

PR #37328, opened 2022-02-14
[TF] Create common WaitForConfig implementation

nmittler:
Security tests have a custom implementation to deal with issues related to WaitForConfig. Moving to a common implementation.

### DIFF
--- a/pkg/test/framework/config.go
+++ b/pkg/test/framework/config.go
@@ -18,8 +18,11 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/hashicorp/go-multierror"
+
 	"istio.io/istio/pkg/test"
 	"istio.io/istio/pkg/test/framework/components/cluster"
+	"istio.io/istio/pkg/test/framework/components/istioctl"
 	"istio.io/istio/pkg/test/framework/resource"
 	"istio.io/istio/pkg/test/scopes"
 )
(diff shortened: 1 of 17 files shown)


howardjohn (commented):
LGTM but I think we should make all places that should, in the future, fail if wait fails, use `OrFail` so that we just swap out the code to actually fail if it fails.

As is, and usage of `OrFail` represents a bug/flaky test waiting to happen

nmittler:
@howardjohn PTAL ... `OrFail` now just logs the failure.

lei-tang (approved):
LGTM.

pkg/test/framework/config.go
```diff
@@ -112,6 +115,40 @@ func (c *configManager) DeleteYAMLOrFail(t test.Failer, ns string, yamlText ...s
...
+	for _, c := range c.ctx.Clusters() {
+		ik, err := istioctl.New(ctx, istioctl.Config{Cluster: c})
+		if err != nil {
+			return err
+		}
+
+		for _, config := range yamlText {
+			config := config
+
+			// TODO(https://github.com/istio/istio/issues/37324): It's currently unsafe
```
douglas-reid, line 129: this sounds like we could be in for surprises?  How big of a concern is this issue?

douglas-reid (approved):

pkg/test/framework/config.go
nmittler: I'm not entirely sure TBH. I just realized it was an issue and raised https://github.com/istio/istio/pull/37325 as a potential mitigation. I suspect fixing istioctl would be a major undertaking and is unlikely to ever happen.

howardjohn (approved):

PR #37335, opened 2022-02-15
wip: remove mutex from model.Service

howardjohn:
Currently we are using a mutex to modify a model.Service. This seems wrong to me; PushContext is supposed to be a snapshot of the current state of the world, but we are mutating it at will. Instead, it seems more appropriate to create a new service with the desired state and trigger a push when we change this.

I am not sure if this fixes any bugs or is simply a refactor...

Currently tests are failing so still a WIP

### DIFF
--- a/pilot/pkg/serviceregistry/aggregate/controller.go
+++ b/pilot/pkg/serviceregistry/aggregate/controller.go
@@ -178,7 +178,10 @@ func (c *Controller) Services() ([]*model.Service, error) {
 		}
 
 		if r.Provider() != provider.Kubernetes {
-			services = append(services, svcs...)
+			for _, s := range svcs {
+				services = append(services, s)
+				smap[s.Hostname] = s
+			}
 		} else {
 			for _, s := range svcs {
 				sp, ok := smap[s.Hostname]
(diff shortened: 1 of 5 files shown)


pilot/pkg/serviceregistry/aggregate/controller.go
```diff
@@ -178,7 +178,10 @@ func (c *Controller) Services() ([]*model.Service, error) {
 		}
 
 		if r.Provider() != provider.Kubernetes {
-			services = append(services, svcs...)
+			for _, s := range svcs {
+				services = append(services, s)
+				smap[s.Hostname] = s
```
hzxuzhonghu, line 183: Isn't this changing the previous behavior say if two SEs with same hostname

howardjohn:
need to fix some stuff so I don't expect the tests to pass yet. But looking
for I'd this looks right directionally?

pilot/pkg/serviceregistry/aggregate/controller.go
```diff
@@ -192,11 +195,15 @@ func (c *Controller) Services() ([]*model.Service, error) {
...
 					// If it is seen second time, that means it is from a different cluster, update cluster VIPs.
-					mergeService(sp, s, r)
+					smap[sp.Hostname] = mergeService(sp, s, r)
 				}
 			}
 		}
 	}
+	for i := range services {
+		// TODO what about conflicts..? This probably breaks same hostname in multiple registry
+		services[i] = smap[services[i].Hostname]
```
hzxuzhonghu, line 205: duplicate with append before

pilot/pkg/serviceregistry/aggregate/controller.go
howardjohn: Yes - this is incorrect. Need to update it

PR #37341, opened 2022-02-15
Introduce delta xds log scope to distinguish ads logs

hzxuzhonghu:

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -25,7 +25,7 @@ import (
 	endpoint "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
 	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	xdstype "github.com/envoyproxy/go-control-plane/envoy/type/v3"
-	structpb "google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/structpb"
 	wrappers "google.golang.org/protobuf/types/known/wrapperspb"
 
 	meshconfig "istio.io/api/mesh/v1alpha1"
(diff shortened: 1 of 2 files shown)


pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -173,7 +174,8 @@ func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *mod
 	return resources, model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", cacheStats.hits, cacheStats.hits+cacheStats.miss)}
 }
 
-func shouldUseDelta(updates *model.PushRequest) bool {
```
ramaraochavali, line 176: This is supposed to be generic and used for all xDS?

hzxuzhonghu: no, it is only for cds

hzxuzhonghu: Otherwise it should not be in this file

ramaraochavali: I think I only added this and my intention was to reuse this across all xDS

howardjohn (commented):
I can see why you sent the PR but I actually have a very slight preference to keep things the way they are. My thinking is that you will likely be all delta or all sotw, so we don't really need to distinguish then much. On initial connection it shows if it's delta or not so if you really need to you can trace it back.

That being said it's only a slight preference and I see benefits here as well so not opposed to merging

hzxuzhonghu:
Yes， we can infer from the context, even some are full xds, some are delta. but not obvious for users.

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: I will revert this  change, do it when necessary

ramaraochavali (approved):

PR #37350, opened 2022-02-15
Fix rendering of injectedAnnotations in Helm charts

luksa:
If an injected annotation value contained a double-quote character, the rendered istiod-injector-configmap YAML would be invalid. Unmarshalling the inject config would fail with: update error: failed to unmarshal injection template: error converting YAML to JSON: yaml: line 10: did not find expected key


howardjohn (approved):

howardjohn:
@zbintliff it did not, I don't think we realized that case would be impacted

zbintliff:
no worries, glad I was able to find _why_ it changed though.

thanks

Issue #37331, opened 2022-02-14
Better support ExternalName


howardjohn:
Example config:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: b-ext
  namespace: echo
spec:
  ports:
  - name: tcp
    port: 80
    targetPort: 18080
  type: ExternalName
  externalName: b.echo.svc.cluster.local
```

Kubernetes behavior: `b-ext` creates a CNAME record for `b`. Port is not relevant.

Istio (today): Create a DNS cluster for each port, which ends up resolving to Cluster IP. Match on Host header (http), SNI (tls), or just port (TCP) (meaning it captures all traffic on that port).

Issues:
* Port matters in Istio but not Kubernetes
* We get minimal functionality because we only have the cluster IP. No load balancing, auto mtls, etc

Proposal option 1: just add additional matchers to existing Service definitions. For example, above we would add a match for `b-ext` everywhere we match for `b`. This works for HTTP and TLS; for TCP it is already not needed as it works today (envoy will see traffic to `b` and `b-ext` as the same). If there was no Service/ServiceEntry defined at all for the target `externalName`, then we would end up just passing through.
Pros: simple and efficient to implement.
Pros: Behaves identically to Kubernetes.
Cons: cannot use ExternalName service to apply different policy to different resources (subset replacement)

Proposal option 2: create new listeners/routes/clusters for the Service. Instead of making a DNS cluster, create and EDS cluster and resolve the `externalName` in the control plane (not literally a DNS lookup, but make it an alias). If no ports are defined, inherent from the referenced Service. If ports are defined, we should probably use them.
Pros: allows override policy (subset replacement)
Cons: more expensive due to additional XDS config
Cons: only works for HTTP, unless we do DNS auto allocation

Past issues:
* https://github.com/istio/istio/issues/31640
* https://github.com/istio/istio/issues/30547
* https://github.com/istio/istio/issues/29846
* https://github.com/istio/istio/issues/24886
* https://github.com/istio/istio/issues/23463
* https://github.com/istio/istio/issues/20703
* https://github.com/istio/istio/issues/13479

howardjohn:
note: knative uses ExternalName but I think it doesn't actually matter in this case. For mesh they override it with VS, and for non-mesh none of this applies.

costinm:
IMO the safest path is for Service.ExternalName to behave exactly like in K8S if there is no Istio-specific config.

[...]

PR #37365, opened 2022-02-16
WIP: Better ExternalName support

howardjohn:
Fixes https://github.com/istio/istio/issues/37331

Example config:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: b-ext
  namespace: echo
spec:
  ports:
  - name: tcp
    port: 80
    targetPort: 18080
  type: ExternalName
  externalName: b.echo.svc.cluster.local
```

Kubernetes behavior: `b-ext` creates a CNAME record for `b`. Port is not relevant.

Istio (today): Create a DNS cluster for each port, which ends up resolving to Cluster IP. Match on Host header (http), SNI (tls), or just port (TCP) (meaning it captures all traffic on that port).

Issues:
* Port matters in Istio but not Kubernetes
* We get minimal functionality because we only have the cluster IP. No load balancing, auto mtls, etc
* TCP doesn't work well since there is no VIP - instead we capture ALL traffic on the configured port 

This PR changes the implementation to more closely align with Kubernetes.

ExternalName becomes a new "Alias" service type internally. An alias simply adds extra matching logic to _existing_ Services.  For example, above we would add a match for `b-ext` everywhere we match for `b`. This works for HTTP and TLS; for TCP it is already not needed as it works today (envoy will see traffic to `b` and `b-ext` as the same). If there was no Service/ServiceEntry defined at all for the target `externalName`, then we would end up just passing through.

Note that port from ExternalName is not used at all, just like in Kubernetes.

If we have a `destination` in VirtualService referring to an ExternalName service explicitly, this will now no longer work unless they have also defined the referenced `externalName` as a concrete service. For example:
```
apiVersion: v1
kind: Service
metadata:
  name: google
spec:
  type: ExternalName
  externalName: google.com
```
If I forward to this in a VS, it will work IFF I have a ServiceEntry defined for google.com. As such, it is generally not recommended to do this (you could just point to google.com directly), but can be useful when use third party Ingress configs, etc. Other Ingress implementations have extremely inconsistent support for ExternalName -- I couldn't find two implementations that behaved the same way -- so there isn't much of a standard to follow here.

One major difference in functionality with this change is that the ExternalName service is now just an alias, not its own "thing" (Envoy cluster). This means we cannot apply different rules for it. However, this can be done with standard DR subsets.

TODO:
* Add tests. Critical ones are Sidecar scoping
* Remove all my debug logs
* Add feature flag to retain old behavior
* Add release note


mikebell90:
@howardjohn is this PR dead?

howardjohn:
Basically. Its tricky since its a breaking change. This might be something we adopt in ambient mesh (as a point where we are already making breaking changes)

mikebell90:
@howardjohn  could this not be behind a Config switch? But yeah, I know the slippery slope of too many of those. We are kind of stuck in this situation, migrating to istio from a setup where have a namespace with all services "mirrored" as external names. This was to have "flat namespacing". And there's lots of code depending on this. While we currently hack around it with SE/VS (or more accurately a RouteTable since we use Solo), it would be nice for this to "just work"

howardjohn:
github is acting weird; re-opening under https://github.com/istio/istio/pull/46332. Has a feature flag this time

Issue #37356, opened 2022-02-15
Old service IP address present in Istio Proxy route config


JWT95:
## Overview

If I delete a service and then recreate it, the deleted IP address for the service is present in the list of domains in the Istio Proxy route config for the service but the new IP address is not present.

This means that requests directly to the service IP (i.e. requests with the service IP in the host header) are not handled correctly. In particular they fail if the mutual tls policy is STRICT as they go via the PassthroughCluster.

#### Services
```
$ kubectl get services -n app
NAME                                   TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                         AGE
alt-http-service                       ClusterIP      10.255.193.97   <none>        7001/TCP                        3s
internal-http                          ClusterIP      10.255.32.53    <none>        7001/TCP,8082/TCP,8083/TCP      52s
```

####  Istio-Proxy route config on a client pod for route "7001"
```yaml
      - version_info: 2022-02-15T15:48:46Z/44
        route_config:
          '@type': type.googleapis.com/envoy.config.route.v3.RouteConfiguration
          name: "7001"
          virtual_hosts:
           - name: allow_any
             ...
           - name: alt-http-service.app.svc.cluster.local:7001
              ...
           - name: internal-http.app.svc.cluster.local:7001
              domains:
                - internal-http.app.svc.cluster.local
                - internal-http.app.svc.cluster.local:7001
                - internal-http
                - internal-http:7001
                - internal-http.app.svc
                - internal-http.app.svc:7001
                - internal-http.app
                - internal-http.app:7001
                - 10.255.32.53
                - 10.255.32.53:7001
```
The Service IP address, `10.255.32.53`, is present in the list of domains

### After deleting the internal-http service

Service deleted by:
```
kl delete service internal-http -n app
```

#### Services
```
$ kubectl get services -n app
alt-http-service                       ClusterIP      10.255.193.97   <none>        7001/TCP                        7m28s
```

#### Istio-Proxy route config on a client pod for route "7001"
```yaml
      - version_info: 2022-02-15T15:55:56Z/45
        route_config:
          '@type': type.googleapis.com/envoy.config.route.v3.RouteConfiguration
          name: "7001"
          virtual_hosts:
            - name: allow_any
             ...
            - name: alt-http-service.app.svc.cluster.local:7001
             ...
```
The service has been removed from the route_config for route "7001"

### After recreating the internal-http service

The service is recreated by doing a Helm upgrade on the Helm release that installed it.

#### Services
[...]

PR #37367, opened 2022-02-16
Fix route cache service key.

bianpengyuan:
Fix https://github.com/istio/istio/issues/37356

Route cache uses a copy of services when constructing key, which does not have attributes populated. Also cache assertion does not really work for rds cache for now and this PR fixes that.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/route/route_cache.go
+++ b/pilot/pkg/networking/core/v1alpha3/route/route_cache.go
@@ -78,7 +78,7 @@ func (r *Cache) DependentConfigs() []model.ConfigKey {
 	configs := make([]model.ConfigKey, 0, len(r.Services)+len(r.VirtualServices)+
 		len(r.DelegateVirtualServices)+len(r.DestinationRules)+len(r.EnvoyFilterKeys))
 	for _, svc := range r.Services {
-		configs = append(configs, model.ConfigKey{Kind: gvk.ServiceEntry, Name: svc.Attributes.Namespace, Namespace: svc.Attributes.Namespace})
+		configs = append(configs, model.ConfigKey{Kind: gvk.ServiceEntry, Name: svc.Attributes.Name, Namespace: svc.Attributes.Namespace})
 	}
 	for _, vs := range r.VirtualServices {
 		configs = append(configs, model.ConfigKey{Kind: gvk.VirtualService, Name: vs.Name, Namespace: vs.Namespace})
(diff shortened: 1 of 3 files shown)


pilot/pkg/networking/core/v1alpha3/route/route_cache.go
```diff
@@ -78,7 +78,7 @@ func (r *Cache) DependentConfigs() []model.ConfigKey {
 	configs := make([]model.ConfigKey, 0, len(r.Services)+len(r.VirtualServices)+
 		len(r.DelegateVirtualServices)+len(r.DestinationRules)+len(r.EnvoyFilterKeys))
 	for _, svc := range r.Services {
-		configs = append(configs, model.ConfigKey{Kind: gvk.ServiceEntry, Name: string(svc.Hostname), Namespace: svc.Attributes.Namespace})
+		configs = append(configs, model.ConfigKey{Kind: gvk.ServiceEntry, Name: svc.Attributes.Namespace, Namespace: svc.Attributes.Namespace})
```
howardjohn, line 81: `svc.Attributes.Namespace` for `Name` seems wrong. I do think Hostname is actually proper here; ConfigKey ServiceEntry actually uses hostname (which is a bit odd)

bianpengyuan: Yeah sorry, this is a typo.

>  I did think Hostname is actually proper here; ConfigKey ServiceEntry actually uses hostname (which is a bit odd)

Oh.. hmm I see, I guess the cache is hit just because namespace is empty then. Let me update that.

pilot/pkg/networking/core/v1alpha3/route/route_cache.go
```diff
@@ -78,7 +78,7 @@ func (r *Cache) DependentConfigs() []model.ConfigKey {
 	configs := make([]model.ConfigKey, 0, len(r.Services)+len(r.VirtualServices)+
 		len(r.DelegateVirtualServices)+len(r.DestinationRules)+len(r.EnvoyFilterKeys))
 	for _, svc := range r.Services {
-		configs = append(configs, model.ConfigKey{Kind: gvk.ServiceEntry, Name: string(svc.Hostname), Namespace: svc.Attributes.Namespace})
+		configs = append(configs, model.ConfigKey{Kind: gvk.ServiceEntry, Name: svc.Attributes.Name, Namespace: svc.Attributes.Namespace})
```
hzxuzhonghu, line 81: I donot think i understand this change, 

as from below, we use hostname to clear

```
	s.Cache.Clear(map[model.ConfigKey]struct{}{{
		Kind:      gvk.ServiceEntry,
		Name:      hostname,
		Namespace: namespace,
	}: {}})
```

bianpengyuan: Yeah sorry for the confusion. I think the root cause is that namespace at key is not populated when inserting new entry.

hzxuzhonghu (approved):

ericvn (approved):
Release note looks good.

howardjohn (approved):

PR #37371, opened 2022-02-16
Allow outbound passthrough on reserved ports

### LINKED ISSUE #36040: Allow outbound passthrough on reserved ports

Istio currently blocks outbound traffic to external systems on a subset of its reserved ports when traffic is proxied through the sidecar using ALLOW_ANY outbound traffic policy. For example, if a workload would like to access an external system on port 15021, this traffic is currently blocked by the sidecar.

In the interest of making the proxy as transparent as possible, we should do away with these restrictions/conflicts by reducing the scope of listeners down to the minimal set of IP addresses needed (e.g. instead of 0.0.0.0, listen on 127.0.0.1 and pod_ip). Please see the "Additional Context" for specific proposals per reserved port. 

- Create ServiceEntry configurations for each external service that uses an Istio reserved port
  - This creates Envoy listeners that will be prioritized above the default reserved-port listeners due to address specificity
  - This is not optimal, since many operators will use ALLOW_ANY specifically to avoid having to maintain service entries for external systems. Additionally, this is cumbersome when destination systems leverage port ranges as opposed to single, well-defined ports.
- Bypass the proxy for systems using reserved ports
  - This is not optimal, since it means traffic to those systems will be excluded from the benefits of the service mesh
- Reconfigure destination systems to use other ports
  - This is sometimes possible, sometimes not, depending on system ownership and configurability of the destination

[ ] Docs
[ ] Installation
[x] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

I have done some analysis on the changes required for this:

- 15000
  - Currently listens only on localhost https://github.com/istio/istio/blob/452101a4d549e580bcf17e4b6ef053e968a52c2b/tools/packaging/common/envoy_bootstrap.json#L153
  - No change needed
- 15001
  - Currently blackholes any traffic with original destination on port 15001 https://github.com/istio/istio/blob/da6178604559bdf2c707a57f452d16bee0de90c8/pilot/pkg/networking/core/v1alpha3/listener_builder.go#L744-L753
  - **Proposal:** Leverage sourceType in the FilterChainMatch to scope the blackhole filter chain down to only traffic coming from the proxy back to itself on 15001
- 15004
  - Currently listens only on localhost https://github.com/istio/istio/blob/452101a4d549e580bcf17e4b6ef053e968a52c2b/pkg/istio-agent/xds_proxy.go#L839-L840
  - No change needed
- 15006
  - Currently blackholes any traffic with original destination on port 15006 https://github.com/istio/istio/blob/da6178604559bdf2c707a57f452d16bee0de90c8/pilot/pkg/networking/core/v1alpha3/listener_builder.go#L558-L570 
  - **Proposal:** Leverage sourceType in the FilterChainMatch to scope the blackhole filter chain down to only traffic coming from the proxy back to itself on 15006
- 15008
  - Couldn't find an actual listener for this. Appears to have been preemptively added as part of https://github.com/istio/istio/issues/24540 via https://github.com/istio/istio/pull/25392 
  - Seems no change needed for now, but would need to ensure reserved ports introduced as part of "Better transport security for HTTP" are unobtrusive
- 15009
  - Like 15008, appears to be WIP, part of https://github.com/istio/istio/issues/24540 via https://github.com/istio/istio/pull/28244
  - Seems no change needed for now, but would need to ensure reserved ports introduced as part of "Better transport security for HTTP" are unobtrusive
- 15020
  - Listens on 0.0.0.0. However, this is not an Envoy listener, therefore, there shouldn't be any bad interactions when there is passthrough traffic targeting this port on an external system. https://github.com/istio/istio/blob/c017654753cd8f4254abbe0567ff9c728d2034d2/pilot/cmd/pilot-agent/status/server.go#L317
  - Note: It appears that with istio-init, this port is redirected to the proxy on OUTPUT (seems correct), but on CNI, it is excluded from redirect/proxy on OUTPUT (seems incorrect) https://github.com/istio/istio/blob/437970b95b6a67b1df4d2211095f1e0ef3f9a9e6/cni/pkg/plugin/redirect.go#L32-L38 https://github.com/istio/istio/blob/437970b95b6a67b1df4d2211095f1e0ef3f9a9e6/cni/pkg/plugin/redirect.go#L63
  - **Proposal:** Remove the outbound exclusion for this port in CNI iptables rules so outbound traffic on 15020 will be redirected to 15001 for passthrough
- 15021
  - Currently listens on 0.0.0.0. Traffic w/ original dst on port 15021 will be redirected to 15001. That listener will delegate to this one, preventing outbound passthrough to external services on 15021. https://github.com/istio/istio/blob/452101a4d549e580bcf17e4b6ef053e968a52c2b/tools/packaging/common/envoy_bootstrap.json#L520
  - **Proposal:** Instead of one listener on 0.0.0.0, let's split this into 2: 1 listening on pod_ip and 1 listening on localhost. This will not match traffic to external services, thus allowing it to passthrough.
- 15053
  - Currently listens on localhost https://github.com/istio/istio/blob/dc98beef00c9b851ee8c79e6dfb89fea8f19c64a/pilot/cmd/pilot-agent/options/options.go#L94
  - No change needed
- 15090
  - Currently listens on 0.0.0.0. Traffic w/ original dst on port 15090 will be redirected to 15001. That listener will delegate to this one, preventing outbound passthrough to external services on 15090. https://github.com/istio/istio/blob/452101a4d549e580bcf17e4b6ef053e968a52c2b/tools/packaging/common/envoy_bootstrap.json#L471
  - **Proposal:** Instead of one listener on 0.0.0.0, let's split this into 2: 1 listening on pod_ip and 1 listening on localhost. This will not match traffic to external services, thus allowing it to passthrough.

**Proof of Concept**

Additionally, I have put together a proof-of-concept implementation of the above proposals to be further developed if this approach makes sense: https://github.com/ricec/istio/commit/14f7ff4b25723215c28e3879a0b17fd61afa2ab6

ricec:
Fixes #36040

**Goal:** Allow outbound traffic on reserved ports (15000, 15001, etc.). Currently, outbound traffic to external systems on these ports fails.
**High-Level Approach:** For each reserved port, instead of listening on 0.0.0.0, listen only on the addresses that are truly necessary (pod IP, localhost, or both).

**Changes:**
- Updated inbound listener to bind to pod IP only
- Updated inbound iptables redirects to always send traffic to pod IP
- Updated outbound listener to bind to localhost only
- Refined traffic-loop-prevention logic to avoid blocking outbound traffic on 15001 and 15006
- Split other reserved-port listeners in 2, one on localhost and one on pod IP
- Change istio-iptables validator to more accurately validate TPROXY (needed as a consequence of binding on pod IP)

_Refreshing goldens added a lot of changes to this PR. To simplify review, the meat of the change is in the first commit, and refreshing goldens in the last._


pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -523,26 +532,11 @@ func buildInboundCatchAllFilterChains(configgen *ConfigGeneratorImpl,
 		ipVersions = append(ipVersions, util.InboundPassthroughClusterIpv6)
 	}
 
-	var filters []*listener.Filter
```
ricec, line 526: Moved this block to `inboundBlackholeFilterChain` below

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -729,29 +727,23 @@ func buildOutboundCatchAllNetworkFiltersOnly(push *model.PushContext, node *mode
 	return filterStack
 }
 
-// TODO: This code is still insufficient. Ideally we should be parsing all the virtual services
-// with TLS blocks and build the appropriate filter chain matches and routes here. And then finally
-// evaluate the left over unmatched TLS traffic using allow_any or registry_only.
-// See https://github.com/istio/istio/issues/21170
-func buildOutboundCatchAllNetworkFilterChains(_ *ConfigGeneratorImpl,
```
ricec, line 736: Moved the logic from this function up to `buildVirtualOutboundListener` and refactored to move the catchall filter chain to `DefaultFilterChains`

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -729,29 +727,23 @@ func buildOutboundCatchAllNetworkFiltersOnly(push *model.PushContext, node *mode
...
+			// Prevent traffic loops on the listener port while still allowing outbound passthrough
+			// to external endpoints that use this same port.
+			//
+			// This only block traffic loops of the form localhost:15001 -> localhost:15001.
+			// Traffic loops of the forms pod_ip:15001 -> pod_ip:15001, pod_ip:15006 -> pod_ip:15006,
+			// and localhost:15006 -> localhost:15006 are blocked on the inbound listener.
 			DestinationPort: &wrappers.UInt32Value{Value: uint32(push.Mesh.ProxyListenPort)},
+			PrefixRanges: []*core.CidrRange{
+				{
+					AddressPrefix: localhost,
```
ricec, line 743: With the outbound blackhole filter chain scoped to only match localhost as destination, it allows outbound to non-localhost on 15001, but blocks localhost loops. This does not block loops on Pod IP, but those are handled by the inbound listener, which blocks _all_ traffic with original_dst on ports 15001 or 15006.

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -523,26 +532,11 @@ func buildInboundCatchAllFilterChains(configgen *ConfigGeneratorImpl,
...
-	filterChains := make([]*listener.FilterChain, 0, 1+5*len(ipVersions))
-	filterChains = append(filterChains, &listener.FilterChain{
-		Name: model.VirtualInboundBlackholeFilterChainName,
-		FilterChainMatch: &listener.FilterChainMatch{
-			DestinationPort: &wrappers.UInt32Value{Value: ProxyInboundListenPort},
-		},
-		Filters: filters,
-	})
+	filterChains := make([]*listener.FilterChain, 0, 2+5*len(ipVersions))
+	filterChains = append(filterChains, inboundBlackholeFilterChain(uint32(ProxyInboundListenPort), push, node))
```
ricec, line 538: Block inbound traffic on both 15001 and 15006 to prevent traffic loops

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -360,18 +361,26 @@ func (lb *ListenerBuilder) buildVirtualOutboundListener(configgen *ConfigGenerat
 		isTransparentProxy = proto.BoolTrue
 	}
 
-	filterChains := buildOutboundCatchAllNetworkFilterChains(configgen, lb.node, lb.push)
+	_, localhost := getActualWildcardAndLocalHost(lb.node)
 
-	actualWildcard, _ := getActualWildcardAndLocalHost(lb.node)
+	filterStack := buildOutboundCatchAllNetworkFiltersOnly(lb.push, lb.node)
+	defaultFilterChain := &listener.FilterChain{
```
ricec, line 367: Move the catchall filter chain out of `FilterChains` and into `DefaultFilterChain`. This is done because of Envoy's matching logic. If a filter chain specifies a destination port but the rest of the `FilterChainMatch` is not a match, Envoy _does not_ fallback to filter chains with wildcard ports. Instead, it falls back to the default filter chain. We need this now since the blackhole filter chain now has additional `FilterChainMatch` criteria beyond just destination port.

See https://github.com/envoyproxy/envoy/blob/main/source/server/filter_chain_manager_impl.cc#L482

tools/packaging/common/envoy_bootstrap.json
```diff
@@ -450,12 +450,14 @@
       {{ end }}
     ],
     "listeners":[
+    {{range $index, $local_ip := .local_ips}}
```
ricec, line 453: Splitting prometheus and status port listeners into 2 each... 1 on localhost and 1 on pod IP. This prevents them from disrupting outbound traffic on ports 15021 and 15090.

.gitignore
```diff
@@ -55,3 +55,4 @@ var/
 ._*
 # MacOS Desktop Services Store
 .DS_Store
+.venv
```
ricec, line 58: Only needed because I'm working from GitHub Codespaces. I can remove once PR is otherwise done, if needed.

tools/istio-iptables/pkg/capture/run.go
```diff
@@ -376,8 +377,8 @@ func (cfg *IptablesConfigurator) Run() {
 	// Use this chain also for redirecting inbound traffic to the common Envoy port
 	// when not using TPROXY.
 
-	cfg.iptables.AppendRule(iptableslog.InboundCapture, constants.ISTIOINREDIRECT, constants.NAT, "-p", constants.TCP, "-j", constants.REDIRECT,
-		"--to-ports", cfg.cfg.InboundCapturePort)
+	cfg.iptables.AppendRule(iptableslog.InboundCapture, constants.ISTIOINREDIRECT, constants.NAT, "-p", constants.TCP, "-j", constants.DNAT,
```
ricec, line 380: See https://github.com/istio/istio/issues/36040#issuecomment-1045360454 for details on why this change was made

ricec:
Lots of test failures. Many look like they're from changes to the test framework that aren't in this branch. Will rebase and take a look...

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -391,7 +400,8 @@ func (lb *ListenerBuilder) buildVirtualInboundListener(configgen *ConfigGenerato
 		isTransparentProxy = proto.BoolTrue
 	}
 
-	actualWildcard, _ := getActualWildcardAndLocalHost(lb.node)
+	podIP := lb.node.IPAddresses[0]
```
howardjohn, line 403: A bit concerned about this. Workload may have many IPs. Can this break anything? @lambdai

howardjohn: One thing that is also unfortunate is it completely kills our ability to cache. We aren't caching this today but in the future we could

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -729,29 +729,23 @@ func buildOutboundCatchAllNetworkFiltersOnly(push *model.PushContext, node *mode
...
-			// ensures we do not passthrough back to the listen port.
+			// Prevent traffic loops on the listener port while still allowing outbound passthrough
+			// to external endpoints that use this same port.
+			//
+			// This only block traffic loops of the form localhost:15001 -> localhost:15001.
+			// Traffic loops of the forms pod_ip:15001 -> pod_ip:15001, pod_ip:15006 -> pod_ip:15006,
+			// and localhost:15006 -> localhost:15006 are blocked on the inbound listener.
 			DestinationPort: &wrappers.UInt32Value{Value: uint32(push.Mesh.ProxyListenPort)},
+			PrefixRanges: []*core.CidrRange{
+				{
```
howardjohn, line 744: nit: use util.ConvertAddressToCidr, will handle ipv6

tests/integration/pilot/common/routing.go
```diff
@@ -867,6 +867,41 @@ func trafficLoopCases(apps *EchoDeployments) []TrafficTestCase {
 	return cases
 }
 
+// reservedPortPassthroughCases contains tests to ensure outbound traffic on
+// reserved ports (e.g. 15000, 15001, etc.) is not blocked
+func reservedPortPassthroughCases(apps *EchoDeployments) []TrafficTestCase {
```
howardjohn, line 872: nice!

tools/packaging/common/envoy_bootstrap.json
howardjohn: https://github.com/envoyproxy/envoy/pull/19367 would make this trivial IIUC

howardjohn (commented):
Overall I think this is really nice. Since this touches the most critical parts of Istio, the bar is super high here, so it would be good to get eyes from @lambdai @ramaraochavali @hzxuzhonghu @costinm if possible...

One thing we should do to de-risk this is to guard it behind a feature flag so it can be turned off if it causes issues. Can wait until we get consensus though if you want to save some effort.

pilot/pkg/networking/core/v1alpha3/listener_builder.go
ricec: For workloads that have many IPs in the same family (e.g. with Multus), the iptables DNAT rules will ensure traffic on any of those IPs is forwarded to the IP on which Envoy is listening.

I'm more concerned about dual-stack IPv4/IPv6 use cases. For these, we would need to duplicate the inbound listener entirely or wait for https://github.com/envoyproxy/envoy/pull/19367 as you had mentioned. Because of this, I'm considering descoping dual-stack from this change. Dual stack use cases would continue listening on wildcard for now. This would mean outbound on 15006 would remain broken for dual-stack use cases, but we would avoid mass listener duplication. Thoughts?

pilot/pkg/networking/core/v1alpha3/listener_builder.go
ricec: Will get this updated

tests/integration/pilot/common/routing.go
ricec: Thanks!

ricec:
Thanks, @howardjohn. Looking forward to feedback from @lambdai, @ramaraochavali, @hzxuzhonghu, and @costinm as well.

manifests/charts/istio-control/istio-discovery/files/injection-template.yaml
```diff
@@ -124,6 +124,9 @@ spec:
         add:
         - NET_ADMIN
         - NET_RAW
+    {{- else if eq (annotation .ObjectMeta `sidecar.istio.io/interceptionMode` .ProxyConfig.InterceptionMode) `TPROXY` }}
+        add:
+        - NET_ADMIN
```
ricec, line 129: The istio-validator container needs NET_ADMIN in order to work with TPROXY. This didn't cause any issues previously, since the validation traffic was localhost-to-localhost, which does not require IP_TRANSPARENT to be set. Now, the traffic is localhost-to-podIP, so this is needed.

The istio-proxy container already runs with NET_ADMIN for TPROXY mode, so this change simply makes the istio-validator more accurate in its validation.

pilot/pkg/networking/core/v1alpha3/listener_builder.go
ricec: @howardjohn, I realize now that dual stack isn't yet supported by Istio, so this shouldn't be a concern for now, right?

pilot/pkg/networking/core/v1alpha3/listener_builder.go
ricec: Updated

ricec:
@howardjohn, @lambdai, @ramaraochavali, @hzxuzhonghu, and @costinm, I believe this change should be functionally complete. All tests are passing, just awaiting your review and signoff. Since your first look at it @howardjohn, I made a couple more changes to address failing IPv6 and CNI scenarios. I've left those commits separate for ease of review.

Please let me know what you think.

ricec:
@howardjohn, this has been marked stale and would be close end of this week. As you can see, I haven't gotten any traction on review from others. Do you want to keep this open?

lambdai (commented):
`If we only listen on localhost, inbound traffic from external will break`

Do we have to support inbound service listening on reserved port?
I think the title is to support istio-proxy as outbound proxy, not inbound proxy for reserved port

ricec:
> `If we only listen on localhost, inbound traffic from external will break`
> 
> Do we have to support inbound service listening on reserved port? I think the title is to support istio-proxy as outbound proxy, not inbound proxy for reserved port

Hey, @lambdai. No, the intent here is only to support outbound on reserved ports, not inbound. The reason inbound listener needs to change is this... If the inbound listener binds to 0.0.0.0:15006, then outbound traffic to 1.2.3.4:15006 will be incorrectly handled by the inbound listener. It would go something like this:

app --- 1.2.3.4:15006 ---> iptables REDIRECT --- 127.0.0.1:15001 ---> outbound listener ---> orig dst filter --- 1.2.3.4:15006 ---> inbound listener

This change prevents this^ scenario by avoiding wildcard bind addresses so that 1.2.3.4:15006 does not match the inbound listener.

lambdai:
Thank you for clarify the intention is to capture outbound traffic. 

What you described aligns with the fact that Envoy Listener 0.0.0.0:15001 would redirect to 0.0.0.0:15006 Envoy listener.  

What if the inbound Envoy listener binds to 127.0.0.1:15006 ? 


```
app --- 1.2.3.4:15006 ---> iptables REDIRECT --- 127.0.0.1:15001 ---> outbound listener ---> orig dst filter --- 1.2.3.4:15006 ---> inbound listener
```

becomes
```
a) the prefix doesn't change until the last step
```
app --- 1.2.3.4:15006 ---> iptables REDIRECT --- 127.0.0.1:15001 ---> outbound listener ---> orig dst filter --- 1.2.3.4:15006 
````
2) outbound listener will redirect to original 1.2.3.4:15006. It doesn't match 127.0.0.1:15006 so it won't hit Envoy Listener from EnvoyListener. 

As long as we tag this connection should not be re-captured by iptables, we are good to go, right?

ricec:
Thanks, @lambdai.

> What if the inbound Envoy listener binds to 127.0.0.1:15006 ?

If Envoy were only listening on localhost, we would need to ensure iptables redirects all traffic to localhost. However, when using REDIRECT, iptables will send inbound traffic to pod IP. From my comment [here](https://github.com/istio/istio/issues/36040#issuecomment-1045360454):
> For all traffic inbound to the pod from external, the REDIRECT from PREROUTING will result in Pod IP being the destination

This means we would still need to move away from REDIRECT to use DNAT instead to force all traffic to 127.0.0.1.

However, there is still another problem with this. If we DNAT packets in this way, they will be seen as martian and will be dropped (unless route_localnet=1, but unless I'm mistaken, setting this causes security issues).

ricec:
Hey, @lambdai. Did this^ answer your questions? Any other questions/concerns?

ricec:
Hi, @lambdai & @howardjohn. This has been marked stale and will auto-close end of this week.

Do you have any other questions/concerns on this change?

costinm:
Sorry, missed this PR - I think some of it is very useful, but as John mentioned it is high risk. What is the reason to allow traffic to those ports in the first place ? 

Re. NET_ADMIN permissions on 'validator' - the main purpose of CNI was to avoid granting net permissions to pods, because in some secure K8S deployments policies prevent pods from getting the permissions.

pilot/pkg/networking/core/v1alpha3/listener_builder.go
costinm: We are working to add it, and this will be a concern. 

I would also be concerned about the change from REDIRECT to DNAT - we also use this for VMs, where AFAIK the IP address may change (on Pods it's almost always fixed - at least the first one). IPv6 addresses do change dynamically quite often ( network prefix ). 

Anything that involves using the IP address is dangerous - in particular since we don't regenerate the bootstrap and can't send updates to istiod.

ricec:
Thanks, @costinm. To answer your questions...

> What is the reason to allow traffic to those ports in the first place ?

Some external systems can (and do) listen on these ports, and ideally, workloads should be able to access them outbound through the proxy.

> Re. NET_ADMIN permissions on 'validator' - the main purpose of CNI was to avoid granting net permissions to pods

Understood, but CNI or not, NET_ADMIN is needed for TPROXY. It is currently set on the istio-proxy container [here](https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/files/injection-template.yaml#L343-L345) only when TPROXY is enabled. Likewise, in this change to the istio-validation container, NET_ADMIN is granted only when TPROXY is enabled.

pilot/pkg/networking/core/v1alpha3/listener_builder.go
ricec: I see. If we can't rely on a fixed IP address, then this solution would require istiod to watch for changes in IP address and update the envoy config accordingly. Doesn't seem ideal.

The only reason this is needed is to prevent outbound on port 15006 from being redirected to the inbound listener. Perhaps a better solution would be for Envoy to support a "don't redirect to me" flag on listeners? If such a thing existed, we could continue to listen on wildcard address for the inbound listener without disrupting outbound traffic.

howardjohn:
I have thought about this a lot over the past few days/months, and I don't think we should move forward with this, personally. I feel pretty bad saying that, as this is an exceptionally well written PR, especially for a new contributor - I am very impressed by the depth of this PR, and that it lingered for so long, and that it _is_ solving a real problem.

Ultimately, though, I feel like this introduces too much risk and complexity. This touches a number of areas we have had CVEs in the past and still continue to find major bugs (https://github.com/istio/istio/issues/38982), and diverges from years of experience running REDIRECT in production (including in other meshes). Additionally, we had a major effort to avoid putting IP addresses in XDS for various reasons, which this backtracks a bit.

Because of these risks, the change would need to be protected by a feature flag. However, given the risk I am not sure we would ever be able to turn this on by default, leaving this to only niche users at a large maintenance cost to the project.

I do think the feature itself (or bug fix, really) is useful. However, I suspect the path forward is more transformational than incremental, doing things like running the proxy in another network namespace which obviates the problem, using a tunneling protocol, etc. While all of this is a bit fuzzy right now, there has been some discussion in this area in the WGs and I think over the next few months will solidify.

nagavijayan-nagarathinam:
I have used this annotation to fix this issue. but does it make any side effect in service mesh network ? `traffic.sidecar.istio.io/excludeOutboundPorts: `

howardjohn:
> I have used this annotation to fix this issue. but does it make any side effect in service mesh network ? `traffic.sidecar.istio.io/excludeOutboundPorts: 15001`

Well traffic going out on port 15001 will bypass the mesh.  But otherwise probably fine

Issue #37291, opened 2022-02-11
Metrics on workload certificate expiry


Christoph-Raab:
**Describe the feature request**

Since istio 1.7 workload certificates are not stored as kubernetes secrets anymore, but rather on the sidecar proxy container. Viewing them works with the `istioctl pc secret <pod>` command. They are also exposed over the admin api under `localhost:15000/certs`, where the istioctl command seams to use that endpoint in combination with port-forwarding.

We want to be able to monitor the workload certificates for possible expiry, e.g. by scraping that information from istiod.

**Describe alternatives you've considered**

An option we considered was defining `values.sidecarInjectorWebhook.templates: certificate-monitor` and maybe set the `values.sidecarInjectorWebhook.defaultTemplates: [sidecar, certificate-monitor]`. An app in the container would basically query `localhost:15000/certs` and extract and exposes the certificate expiry, being scraped by prometheus on a regular basis. Got a working proof of concept, but the overhead on both prometheus and running a second sidecar for every istio enabled workload is unproportional, especially when having thousands of workloads. 

Also one could think about opening the admin endpoint for a specific monitoring application that collects the information and exposes them to prometheus. This was discarded because of the obvious security concerns.

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[x] Extensions and Telemetry
[x] Security
[ ] Test and Release
[] User Experience
[ ] Developer Infrastructure

bianpengyuan:
This looks more like a metric at istio proxy instead of istiod? Envoy has a metric exposed `envoy_server_days_until_first_cert_expiring` although it will almost always be 0 since the default expiry 24 hours.

anannaya:
@bianpengyuan   Is there any way i can get this metrics from pilot-agent ? (In seconds )

bianpengyuan:
The metric is not yet available, it needs to be added, probably via a periodic goroutine running against the secret cache at istio-agent: https://github.com/istio/istio/blob/284def0c93cc2dc37b07a1d7acd74f26eb24fb63/security/pkg/nodeagent/cache/secretcache.go#L93

zirain:
@bianpengyuan IIRC, we can add metric here?

https://github.com/istio/istio/blob/199b33bf6d0515c995cde922973a2ec2a4dd4c20/security/pkg/nodeagent/cache/secretcache.go#L600

bianpengyuan:
[...]

PR #37372, opened 2022-02-16
pilot-agent add cert expiry metric

zirain:

add a Gauge metric of `default` cert expiry minutes on `pilot-agent`

```console
# HELP istio_agent_cert_expiry_minutes The left minutes, when cert chain will expire. A negative value indicates the cert is expired.
# TYPE istio_agent_cert_expiry_minutes gauge
istio_agent_cert_expiry_minutes{resource_name="default"} 701
```

fixed: https://github.com/istio/istio/issues/37291

***warning***

see from 
https://github.com/census-instrumentation/opencensus-go/blob/49838f207d61097fc0ebb8aeef306913388376ca/metric/registry.go#L237

the registry use metric.Name as key, I think we can not have multi DerivedGauge (same name but different in labels) before we modify the implement of upstream.

In short-term, this PR will allow user monitor `default` cert.
for a long-term, we should make `DerivedGauge ` support metrics with different labels.

BTW, `opencensus-go` is merged to `OpenTelemetry`, so that would take a little more time to achieve this.

### DIFF
--- a/security/pkg/nodeagent/cache/monitoring.go
+++ b/security/pkg/nodeagent/cache/monitoring.go
@@ -51,7 +51,7 @@ var (
 		"cert_expiry_seconds",
 		"The left time, in seconds, when cert chain will expire. "+
 			"A negative value indicates the cert is expired.",
-			monitoring.WithLabelKeys("resource_name"))
+		monitoring.WithLabelKeys("resource_name"))
 )
 
 func init() {
--- a/security/pkg/nodeagent/cache/monitoring.go
+++ b/security/pkg/nodeagent/cache/monitoring.go
@@ -14,7 +14,9 @@
 
 package cache
 
-import "istio.io/pkg/monitoring"
+import (
+	"istio.io/pkg/monitoring"
+)
 
 var RequestType = monitoring.MustCreateLabel("request_type")
 
@@ -44,6 +46,12 @@ var (
 	numFileSecretFailures = monitoring.NewSum(
 		"num_file_secret_failures_total",
 		"Number of times secret generation failed for files")
+
+	certExpirySeconds = monitoring.NewDerivedGauge(
+		"cert_expiry_seconds",
+		"The left time, in seconds, when cert chain will expire. "+
+			"A negative value indicates the cert is expired.",
+			monitoring.WithLabelKeys("resource_name"))
 )
 
 func init() {
--- a/security/pkg/nodeagent/cache/secretcache.go
+++ b/security/pkg/nodeagent/cache/secretcache.go
@@ -628,6 +628,7 @@ func (sc *SecretManagerClient) rotateTime(secret security.SecretItem) time.Durat
 
 func (sc *SecretManagerClient) registerSecret(item security.SecretItem) {
 	delay := sc.rotateTime(item)
+	certExpirySeconds.ValueFrom(func() float64 { return time.Until(item.ExpireTime).Seconds() }, item.ResourceName)
 	item.ResourceName = security.WorkloadKeyCertResourceName
 	// In case there are two calls to GenerateSecret at once, we don't want both to be concurrently registered
 	if sc.cache.GetWorkload() != nil {
--- a/pilot/pkg/bootstrap/monitoring.go
+++ b/pilot/pkg/bootstrap/monitoring.go
@@ -47,7 +47,7 @@ var (
 	)
 )
 
-func init() {
+func init(){
 	uptime.ValueFrom(func() float64 {
 		return time.Since(serverStart).Seconds()
 	})
--- a/pilot/pkg/bootstrap/monitoring.go
+++ b/pilot/pkg/bootstrap/monitoring.go
@@ -47,7 +47,7 @@ var (
 	)
 )
 
-func init(){
+func init() {
 	uptime.ValueFrom(func() float64 {
 		return time.Since(serverStart).Seconds()
 	})


anannaya:
Thanks for the PR .much needed one

security/pkg/nodeagent/cache/monitoring.go
```diff
@@ -44,6 +51,12 @@ var (
 	numFileSecretFailures = monitoring.NewSum(
 		"num_file_secret_failures_total",
 		"Number of times secret generation failed for files")
+
+	certExpiryMinutes = monitoring.NewGauge(
+		"cert_expiry_minutes",
```
howardjohn, line 56: seconds may be more canonical: https://prometheus.io/docs/practices/naming/#base-units

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -186,11 +186,23 @@ func NewSecretManagerClient(caClient security.Client, options *security.Options)
...
 	return ret, nil
 }
 
+func (sc *SecretManagerClient) refreshMetrics() {
+	t := time.NewTicker(time.Minute)
+	for range t.C {
+		if sc.cache.GetWorkload() != nil {
+			item := sc.cache.GetWorkload()
+			delay := sc.rotateTime(*item)
+			RecordCertExpiry(item.ResourceName, delay)
```
howardjohn, line 201: I think we want NewDerivedGauge, so we do not need to manually run a goroutine. Can compute it during each prometheus fetch

zirain: https://github.com/istio/pkg/blob/4fff90a48d5b5b21f7ab31a18db3b94272249b54/monitoring/monitoring.go#L189

why `NewDerivedGauge` not support labels? for now, I can just put resource name on metric name, and it will looks like as following

```console
# HELP istio_agent_default_cert_expiry_seconds The left time, in seconds, when cert chain will expire. A negative value indicates the cert is expired.
# TYPE istio_agent_default_cert_expiry_seconds gauge
istio_agent_default_cert_expiry_seconds 86242.038222406

```

the format looks wired

security/pkg/nodeagent/cache/monitoring.go
```diff
@@ -55,3 +60,13 @@ func init() {
 		numFileSecretFailures,
 	)
 }
+
+func RecordCertExpiry(resourceName string, expiry time.Time) {
+	monitoring.NewDerivedGauge(
```
bianpengyuan, line 65: Can we add labels to derived metric? I know it is currently supported but I don't think there should be anything blocking?

bianpengyuan: Feels wrong to have resource name at stats name.

zirain: need to take a look at istio/pkg package

zirain: looks good, need to update `istio/pkg` first.

```console
# HELP istio_agent_cert_expiry_seconds The left time, in seconds, when cert chain will expire. A negative value indicates the cert is expired.
# TYPE istio_agent_cert_expiry_seconds gauge
istio_agent_cert_expiry_seconds{resource_name="default"} 86108.240811166
```

zirain:
https://github.com/istio/pkg/pull/483 @howardjohn PTAL

zirain:
here is the latest output:
```console
# HELP istio_agent_cert_expiry_seconds The left time, in seconds, when cert chain will expire. A negative value indicates the cert is expired.
# TYPE istio_agent_cert_expiry_seconds gauge
istio_agent_cert_expiry_seconds{resource_name="default"} 86279.188321444

```

bianpengyuan (approved):

security/pkg/nodeagent/cache/monitoring.go
```diff
@@ -55,3 +59,17 @@ func init() {
 		numFileSecretFailures,
 	)
 }
+
+func recordCertExpiry(resourceName string, expiry time.Time) {
+	labels := map[string]string{
+		"resource_name": resourceName,
+	}
+	monitoring.NewDerivedGaugeWithLabels(
```
douglas-reid, line 67: what happens if this is called more than once with different resourceNames?

zirain: the metric will be updated.
```golang
	recordCertExpiry(item.ResourceName, item.ExpireTime)
	recordCertExpiry(item.ResourceName+"bak", item.ExpireTime)
```
output:
```console
# HELP istio_agent_cert_expiry_seconds The left time, in seconds, when cert chain will expire. A negative value indicates the cert is expired.
# TYPE istio_agent_cert_expiry_seconds gauge
istio_agent_cert_expiry_seconds{resource_name="defaultbak"} 86355.853291707

```

douglas-reid: Yes. I guess I'm asking if there is ever a valid case for two cert expiry metrics in the same process.

zirain: IIRC, only `default` cert, @howardjohn can you confirm this?

howardjohn: We support file mounted ones, so may be unbounded number actually. Also ROOTCA may be something we want to measure as well..?

douglas-reid: I think the workaround we want is something like: https://github.com/istio/pkg/pull/483/files#r810176904

zirain: https://github.com/census-instrumentation/opencensus-go/blob/49838f207d61097fc0ebb8aeef306913388376ca/metric/registry.go#L237

the `registry` use `metric.Name` as key, I think we can not have multi `DerivedGauge` (same name but different in labels) before we modify the implement of upstream.

costinm:
I assume the main goal here is to generate an alert if something is wrong - certs not renewed and close to expiration, and critical for expired certs.

For this purpose, time in seconds may not be the best metric.

We know when we are supposed to renew, so adding metric on how many certs are beyond the expected renewal time, and one for expired certs may be more useful. Or number of renewal failure. 

It is a critical area, we should also have some best practice, example alerts and in general a higher level view on this.

Not arguing against this metric, we can also expose this - but it's harder to page or know something is about to be wrong without some guessing.

Christoph-Raab:
> I assume the main goal here is to generate an alert if something is wrong - certs not renewed and close to expiration, and critical for expired certs.

As requester of the feature I want to clarify: Our rootca certificate lifetime is rather short, speaking of days (not the default 10 years). We want to be able to monitor the workload certificates expiry to ensure that they do not expire after the root/intermediate cacerts, which would make it impossible for the sidecar to renew the workload certificate.

For that use case the time in seconds until expiry is exactly what we're looking for.

zirain:
@howardjohn @douglas-reid updated the desc, please review this PR again.

security/pkg/nodeagent/cache/monitoring.go
howardjohn: What is current status? Last one wins?

If so that is not a great option IMO. Either we only do `default`, do `default`+`ROOTCA`, only do the oldest, or do all of them. But doing the last one will give unexpected results

douglas-reid: @zirain, I was OOO so I may have lost context, but the metric.Name as the key shouldn't be an issue. If the function is applied with labels, it should be possible to have multiple values. I'll try to add the functionality with tests and ping back here.

zirain: > What is current status? Last one wins?
> 
> If so that is not a great option IMO. Either we only do `default`, do `default`+`ROOTCA`, only do the oldest, or do all of them. But doing the last one will give unexpected results

the last win, but right now it will only be called at `registerSecret`.

zirain: > @zirain, I was OOO so I may have lost context, but the metric.Name as the key shouldn't be an issue. If the function is applied with labels, it should be possible to have multiple values. I'll try to add the functionality with tests and ping back here.

if we can make `DerivedGauge` support metrics with same name and labels keys but different label values, everything will be more easier.

douglas-reid: @zirain @howardjohn please take a look at https://github.com/istio/pkg/pull/490. i think that (or similar) should allow us to define the metric once and have multiple calls like:

```golang
certExpirySeconds = monitoring.NewDerivedGauge(
       "cert_expiry_seconds",
       "The left time, in seconds, when cert chain will expire. ",
       monitoring.WithLabelKeys("resource_name"),
)
...
certExpirySeconds.ValueFrom(func() float64 { return time.Until(expiry).Seconds() }, resourceName)
```

go.mod
```diff
@@ -2,6 +2,8 @@ module istio.io/istio
 
 go 1.17
 
+replace istio.io/pkg => github.com/douglas-reid/pkg v0.0.0-20220301021643-00c7529ce85a
```
howardjohn, line 5: can remove now I think ?

howardjohn (approved):

go.mod
zirain: yes, that's a nit.

security/pkg/nodeagent/cache/monitoring.go
```diff
@@ -44,6 +46,12 @@ var (
 	numFileSecretFailures = monitoring.NewSum(
 		"num_file_secret_failures_total",
 		"Number of times secret generation failed for files")
+
+	certExpirySeconds = monitoring.NewDerivedGauge(
+		"cert_expiry_seconds",
+		"The left time, in seconds, when cert chain will expire. "+
```
douglas-reid, line 52:
```suggestion
		"The time remaining, in seconds, before the certificate chain will expire. "+
```

nrjpoddar:
Is this enabled by default?

howardjohn:
Yes

PR #37374, opened 2022-02-16
remove feature flags for se selecting pods

ramaraochavali:
These have been enabled since long and have been stable. I do not see any need for for disabling them. So removing them.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -389,14 +389,6 @@ var (
 		"If enabled, pilot will authorize XDS clients, to ensure they are acting only as namespaces they have permissions for.",
 	).Get()
 
-	EnableServiceEntrySelectPods = env.RegisterBoolVar("PILOT_ENABLE_SERVICEENTRY_SELECT_PODS", true,
-		"If enabled, service entries with selectors will select pods from the cluster. "+
-			"It is safe to disable it if you are quite sure you don't need this feature").Get()
-
-	EnableK8SServiceSelectWorkloadEntries = env.RegisterBoolVar("PILOT_ENABLE_K8S_SELECT_WORKLOAD_ENTRIES", true,
-		"If enabled, Kubernetes services with selectors will select workload entries with matching labels. "+
-			"It is safe to disable it if you are quite sure you don't need this feature").Get()
-
 	InjectionWebhookConfigName = env.RegisterStringVar("INJECTION_WEBHOOK_CONFIG_NAME", "istio-sidecar-injector",
 		"Name of the mutatingwebhookconfiguration to patch, if istioctl is not used.").Get()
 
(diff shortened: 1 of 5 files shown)


djannot:
This change is causing problem for some users.

Here is the scenario.

A user is creating a ServiceEntry which selects both a Kubernetes Service and a WorkloadEntry which corresponds to a remote Service (from another Kubernetes cluster).
Using the hostname defined in the ServiceEntry, the user can send requests to both the local and the remote service.

But now when the user tries to send a request to the local service (using service.namespace.svc.cluster.local), requests are also sent to the remote service, which isn't expected.

Setting `PILOT_ENABLE_K8S_SELECT_WORKLOAD_ENTRIES` to `false` was allowing the user to fix the issue, but since Istio 1.14, there's no solution.

linsun:
@ramaraochavali what is your thoughts?  Perhaps what @djannot describes is possible with some configs?  If not, could we restore this function?

linsun:
> cc @djannot would you please open an new ticket?

let's gather Rama's thoughts first before we open a new ticket.

ramaraochavali:
Can you please explain the use case? Why do you want with one host (SE host) both accessible and with another host (k8s host)  only k8s pods? In which case you need one vs. other? This more seems like a subset use case rather than these variables?

hzxuzhonghu:
Not strongly from me. I guess some users may even not know this feature at all, so with the env default set, they could meet unexpected results.  

Since pods and workloadentry are not the same resources, to prevent misbehavior I am ok to keep the feature env.

ramaraochavali:
> I guess some users may even not know this feature at all, so with the env default set, they could meet unexpected results.

I think here the case is reverse though. They are not using the defaults, so they have intentionally changed.  My worry is these env vars are intended for short duration like to avoid surprises during upgrades as you said  but supporting them forever may not be good.

Any way let us wait for the use case and @howardjohn's thoughts and we can decide.

djannot:
@ramaraochavali the goal of this ServiceEntry is to have a global name which corresponds to both local and remote endpoints.
Then, outlier detection and locality settings are configured with a DestinationRule to send only the requests to the local endpoints and automatically failover to the remote endpoints when the local endpoints aren't available.

ramaraochavali:
That is very interesting use of this feature. But 

> only the requests to the local endpoints and automatically failover to the remote endpoints when the local endpoints aren't available.

So you configure the remote WLE with a different locality and configure Locality LB to fail over to that endpoint?

And when do you send a request to the local service (using service.namespace.svc.cluster.local)? Based on what you described they should go to remote only when all local endpoints are down?  So there are some flows where you want to access only local endpoints?

djannot:
Yes, the remote WLE has a different locality.

The goal is to have the global name (service.global) having this behavior (local when possible, failover to remote), but when sending a request to service.namespace.svc.cluster.local it should only be local.

ramaraochavali:
IMO, Workload Entries and these specific flags are not meant to represent remote **kubernetes** cluster pods and  are intended for collating end points of k8s service and VM workloads during migration. 

Let me think about if there is any existing feature can solve your use case.

@howardjohn for this thoughts

Issue #36791, opened 2022-01-11
TestClusterLocal flakes


hzxuzhonghu:
Recently i met several failure of TestClusterLocal 

https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/36788/integ-pilot-multicluster-tests_istio/1480827322700402688

hzxuzhonghu:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/36908/integ-pilot-multicluster_istio/1484342318780649472

howardjohn:
So the test applies this config:
```yaml
apiVersion: networking.istio.io/v1beta1
kind: DestinationRule
metadata:
  name: mysvc-dr
  namespace: echo
spec:
  host: b.echo.svc.cluster.local
  subsets:
  - name: primary
    labels:
      topology.istio.io/cluster: primary
  - name: remote
    labels:
      topology.istio.io/cluster: remote
  - name: cross-network-primary
    labels:
      topology.istio.io/cluster: cross-network-primary
---
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: mysvc-vs
  namespace: echo
spec:
  hosts:
  - b.echo.svc.cluster.local
  http:
  - name: "primary-local"
    match:
    - sourceLabels:
        topology.istio.io/cluster: primary
    route:
    - destination:
        host: b.echo.svc.cluster.local
        subset: primary
  - name: "remote-local"
    match:
    - sourceLabels:
        topology.istio.io/cluster: remote
    route:
    - destination:
        host: b.echo.svc.cluster.local
        subset: remote
  - name: "cross-network-primary-local"
    match:
    - sourceLabels:
        topology.istio.io/cluster: cross-network-primary
    route:
    - destination:
        host: b.echo.svc.cluster.local
        subset: cross-network-primary
```

Then later removes it and sends requests. The failures show the `remote-local` route still present. Istiod configz shows it does not know the VS, and you can see Istiod triggers a push. 


XDS logs shows push of v129 then v130. Only the a pod (client in this case, the only one that matters) has v129 and has the `remote-local` route. Others are already on v130. `a` pod starts the push after v129 is started but before v130 has. RDS push is 0/64 cache so unlikely a cache issue. For some reason, once it completes the full XDS push cycle, it never pushes again - so v130 is missing.

Other pods, such as `b`, also start before v130, and were able to push again after completion

howardjohn:
[...]

PR #37381, opened 2022-02-16
[TF] Temporarily switch back to RR LB for tests

nmittler:
This is a temporary work-around for #36791 while we investigate.

### DIFF
--- a/tests/integration/iop-integration-test-defaults-with-quic.yaml
+++ b/tests/integration/iop-integration-test-defaults-with-quic.yaml
@@ -62,6 +62,7 @@ spec:
         UNSAFE_ENABLE_ADMIN_ENDPOINTS: true
         PILOT_REMOTE_CLUSTER_TIMEOUT: 15s
         PILOT_ENABLE_QUIC_LISTENERS: true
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

nmittler:
@howardjohn the errors don't seem related to this PR. Perhaps https://github.com/istio/istio/pull/37378 broke the security tests somehow?

hzxuzhonghu:
seems no effect at all

nmittler:
Seems like this isn't the cause of the test flakes. Closing.

howardjohn:
On this failure it was a 503 actually: `    multicluster_test.go:144: failed calling a (cluster=remote)->'http://b.echo-1-80610.svc.cluster.local:80/': call failed from a (cluster=remote) to http://b.echo-1-80610.svc.cluster.local:80 (using http): timeout while waiting after 105 attempts, 0/3 successes (last error: response[19] Status Code: 503)`. The flakes are not.

A bit worried turning off the flag breaks everything, and I don't think we have proven it is not the flake cause yet per ^

PR #37393, opened 2022-02-17
[release-1.13] Image version should point to release-1.13 image

jacob-delgado:


ericvn:
@jacob-delgado This image is correctly a 1.13 image, even though it was named master. It's master since there was a problem in test-infra that I fixed so the next time a base-image is built, it will have the correct name.

PR #37405, opened 2022-02-17
Headless service should use target port in istio

hzxuzhonghu:

Without this, a headless service with

 port: 80
 targetPort: 8080

Istio does not generate a listener for 8080, not event to manage its traffic.

TODO: should add a release note

### DIFF
--- a/pilot/pkg/serviceregistry/kube/conversion.go
+++ b/pilot/pkg/serviceregistry/kube/conversion.go
@@ -45,7 +45,15 @@ const (
 	NodeSelectorAnnotation = "traffic.istio.io/nodeSelector"
 )
 
-func convertPort(port coreV1.ServicePort) *model.Port {
+func convertPort(port coreV1.ServicePort, override bool) *model.Port {
+	if override {
+		if port.TargetPort.IntVal != 0 {
+			// set the target port as port for headless service to line up with k8s.
+			// kube-proxy does not redirect traffic to headless service port.
+			// So it is only viable to access with the real target port.
+			port.Port = port.TargetPort.IntVal
+		}
+	}
 	return &model.Port{
 		Name:     port.Name,
 		Port:     int(port.Port),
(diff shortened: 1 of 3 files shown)


howardjohn (changes requested):
This is not how Kubernetes handles Headless services, they ignore target port as well

hzxuzhonghu:
> This is not how Kubernetes handles Headless services, they ignore target port as well


Kube proxy does not have any rules set for headless service, just provide a way to resolve the domain to specific pod. So it only allows access by podip:targetport. This is what we should respect in istio as well. However, we didn't before.

howardjohn:
Say I have a svc port 80 target port 8080. First,to make sure we are on the same page this config is dumb and users should never do this - I don't know why k8s doesn't reject.

When a client calls a service, they should call `curl service:servicePort`. So in this case port 80.

IMO we should either keep current behavior or support all ports (in that doc I shared). Maybe we can support both ports. But just switching this I think is more likely to break folks.

Ultimately in Kubernetes you can put literally anything in either port and it doesn't matter, so anything we derive from port is pretty much arbitrary. I would favor either fixing it properly or keep status quo for backwards compatibility

hzxuzhonghu:
>When a client calls a service, they should call curl service:servicePort. So in this case port 80.

In headless case, they should curl service:targetPort

hzxuzhonghu:
Headless service is very tricky, as you say user can not define any port or even define the wrong ports. User can curl svc:port if the port is listened by pod.

hzxuzhonghu:
So I would rather say what we did before in istio is totally wrong.

howardjohn:
Totally agree the current state is wrong. But concern is that this PR is
also wrong for the same reason, and changes long standing istio before. I
think we should wait until we can support "all port" which I am hoping is
very soon. WDYT?

hzxuzhonghu:
@howardjohn In volcano, the ports are totally a placeholder, no real meaning at all.

https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/job/plugins/svc/svc.go#L242-L243

howardjohn:
weird, wonder why not no ports

hzxuzhonghu:
I filed a pr to remove

PR #37413, opened 2022-02-17
push context: refactor DestinationRule to split usages

howardjohn:
Currently, we actually have two distinct code paths in a single
function. This is pretty confusing to follow, and makes other PRs more
complex (https://github.com/istio/istio/pull/37174/files).

This splits out to an internal function, which is to be called when
creating SidecarScope, and a method on SidecarScope to be called by
consumers (eg Cluster builder code).

This can be cleaned up further to stop doing the `dummyNode` thing,
which is a major source of bugs historically, if we do the same with
Service. But trying to keep this minimal.

This PR also has a lot of testing changes, as many of our tests do
things that are not realistic.

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -936,17 +936,9 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Colle
 }
 
 // DestinationRule returns a destination rule for a service name in a given domain.
-func (ps *PushContext) DestinationRule(proxy *Proxy, service *Service) *config.Config {
+func (ps *PushContext) destinationRuleForSidecarScope(proxy *Proxy, service *Service) *config.Config {
 	if service == nil {
 		return nil
-	}
-
-	// If proxy has a sidecar scope that is user supplied, then get the destination rules from the sidecar scope
-	// sidecarScope.config is nil if there is no sidecar scope for the namespace
-	if proxy.SidecarScope != nil && proxy.Type == SidecarProxy {
-		// If there is a sidecar scope for this proxy, return the destination rule
-		// from the sidecar scope.
-		return proxy.SidecarScope.destinationRules[service.Hostname]
 	}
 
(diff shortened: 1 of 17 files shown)


kfaseela:
thank you so much @howardjohn ! This really helps my other PR.

kfaseela (approved):

stevenctl (approved):

pilot/pkg/model/push_context.go
```diff
@@ -936,19 +936,11 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Colle
...
 	if service == nil {
 		return nil
 	}
 
-	// If proxy has a sidecar scope that is user supplied, then get the destination rules from the sidecar scope
-	// sidecarScope.config is nil if there is no sidecar scope for the namespace
-	if proxy.SidecarScope != nil && proxy.Type == SidecarProxy {
-		// If there is a sidecar scope for this proxy, return the destination rule
-		// from the sidecar scope.
-		return proxy.SidecarScope.destinationRules[service.Hostname]
```
hzxuzhonghu, line 949: not sure i understand

hzxuzhonghu: IC, is there a plan to do the same for services/virtualservices?

pilot/pkg/model/sidecar.go
```diff
@@ -557,6 +557,10 @@ func (sc *SidecarScope) AddConfigDependencies(dependencies ...ConfigKey) {
 	}
 }
 
+func (sc *SidecarScope) DestinationRule(svc host.Name) *config.Config {
```
hzxuzhonghu, line 560: Maybe keep PushContext.DestinationRule name as well, it calls this function

pilot/pkg/model/push_context.go
hzxuzhonghu: I vote yes, they looks simliar

PR #37414, opened 2022-02-17
Remove mock service discovery

howardjohn:
We currently have two fake service discovery mechanisms. This reduces it to one. This should only impact tests.

Also some function signatures are cleaned up for simplicty.

This is purely refactoring of test code. There should be no impact on production code.


ramaraochavali:
Nice

jacob-delgado (approved):

Issue #35835, opened 2021-11-02
Use a local HTTP server for serving files in integration tests


howardjohn:
Some integration tests use public URLs to access files. This is problematic for a few reasons:

* Adds a dependency on external service, making tests not hermetic. This has broken us before
* Does not work in  airgap env
* Does not work for ipv6 testing.
   * In our current setup we have no external connectivity for ipv6 clusters
   * Even if we did, github does not support ipv6 it seems

Currently I know we use this for the JWKS and attributegen, but there may be others.

Proposed solution: Deploy a trivial file server that serves plaintext + HTTPS into istio-system for tests that require it

yangminzhu:
Regarding the JWT test, we actually already have a local jwks server deployed in the cluster but it's not used for all tests, do we want to change all jwt tests to use it?

howardjohn:
Seems reasonable.

It might be nice to also server generic bytes so we can server attributegen out of it as well (really the server is 99% there other than it serves a single hardcoded const), but we can do that in future. It might be easier to reach attributegen as well since the google API does have IPv6 connectivity at least (although that only solves the 3rd use case, but its the one I am focusing on currently).

yangminzhu:
@aryan16 this could be a good follow-up after your last PR that adds the https support to the jwks server.

aryan16:
Nice !! correct me if I'm wrong here, we need to add services which would minimize the external services dependencies for integration tests right ?

nmittler:
@aryan16 @yangminzhu any update?

nmittler:
@aryan16 is this something you'll be working on? Should we assign to you?

aryan16:
Yeah I can look into this. I am adding one https jwks server, we can try to use it for other tests as well.

howardjohn:
This would fix https://github.com/istio/istio/issues/38390

aryan16:
should we use https jwks server instead of github links for pubkey ?

PR #37420, opened 2022-02-18
Add a fake container registry.

bianpengyuan:
Add a fake container registry which will be used in the future Wasm integration test. The registry preloads the given images, and serve images with hard coded username and password.

### DIFF
--- a/pkg/test/fakes/imageregistry/Makefile
+++ b/pkg/test/fakes/imageregistry/Makefile
@@ -1,0 +1,33 @@
+# Copyright Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+.PHONY: build_and_push clean all
+
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+MD_PATH := $(dir $(MKFILE_PATH))
+IMG := gcr.io/istio-testing/fake-registry
+
+# NOTE: TAG should be updated whenever changes are made in this directory
+# This should also be updated in dependent components
+TAG := 1.0
+
+all: build_and_push clean
+
+build_and_push:
+	cd $(MD_PATH) && CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -tags netgo -ldflags '-w -extldflags "-static"' main.go
+	docker build $(MD_PATH) -t $(IMG):$(TAG)
+	docker push $(IMG):$(TAG)
+
+clean:
+	rm $(MD_PATH)/main
--- a/pkg/test/fakes/imageregistry/main.go
+++ b/pkg/test/fakes/imageregistry/main.go
@@ -1,0 +1,113 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"crypto/tls"
+	"encoding/base64"
+	"flag"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"istio.io/pkg/log"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	containerregistry "github.com/google/go-containerregistry/pkg/registry"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+var port = flag.Int("port", 1338, "port to run registry on")
+var images = flag.String("images", "", "comma separated list of images that will be preloaded and served")
+var registry = flag.String("registry", "localhost:1338", "name of registry to push and pull image")
+
+const (
+	User   = "user"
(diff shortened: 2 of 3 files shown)


pkg/test/fakes/imageregistry/main.go
```diff
@@ -0,0 +1,113 @@
...
+const (
+	User   = "user"
+	Passwd = "passwd"
+)
+
+type Handler struct {
+	registry http.Handler
+}
+
+func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
howardjohn, line 46: can we make this forward compatible with https://github.com/istio/istio/issues/35835? so we can just have one "file server" that can serve blobs or images?

howardjohn: Do not need to scope creep this PR (unless you want...), just if there are things we would do differently

pkg/test/fakes/imageregistry/main.go
```diff
@@ -0,0 +1,113 @@
...
+var port = flag.Int("port", 1338, "port to run registry on")
+var images = flag.String("images", "", "comma separated list of images that will be preloaded and served")
+var registry = flag.String("registry", "localhost:1338", "name of registry to push and pull image")
+
+const (
+	User   = "user"
+	Passwd = "passwd"
+)
+
+type Handler struct {
```
howardjohn, line 42: can we add /ready for probe

pkg/test/fakes/imageregistry/main.go
```diff
@@ -0,0 +1,113 @@
...
+	"istio.io/pkg/log"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	containerregistry "github.com/google/go-containerregistry/pkg/registry"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+var port = flag.Int("port", 1338, "port to run registry on")
+var images = flag.String("images", "", "comma separated list of images that will be preloaded and served")
```
howardjohn, line 34: should we bake them into the docker image? so startup is improved?

pkg/test/fakes/imageregistry/main.go
```diff
@@ -0,0 +1,113 @@
...
+		t.TLSClientConfig = &tls.Config{
+			InsecureSkipVerify: true, //nolint: gosec
+		}
+		img, err := remote.Image(ref, remote.WithTransport(t))
+		if err != nil {
+			log.Errorf("Could not fetch image %v", err)
+			continue
+		}
+
+		log.Infof("Pusing image %v", newImage)
```
howardjohn, line 102: I don't quite follow why we are pushing an image..?

pkg/test/fakes/imageregistry/main.go
bianpengyuan: I am planning to test with images generated from istio/proxy (e.g. attributes gen), which could often change, so that's why I am making it configured dynamically here.

pkg/test/fakes/imageregistry/main.go
bianpengyuan: I don't find an interface to retag an image programmatically with this library, so taking this pushing detour here.

bianpengyuan: Let me dig a bit more into this.

pkg/test/fakes/imageregistry/main.go
howardjohn: If they are already pushed to some registry I don't really get what we need this for then...? Just to add user and passwd check? If just that it seems like we could just 301 redirect to the real registry (not 100% sure its possible but I think it is)?

pkg/test/fakes/imageregistry/main.go
howardjohn: `crane cp` does it, so I would look there. it might be what you did though, there is no native "retag" mechanism in docker AFAIK

howardjohn: oh so this takesthe original images, then pushes it to localhost:1338 which is serving our own ServeHTTP above in the same process?

bianpengyuan: Yeah, so how it will be used is that we configure it to pull `gcr.io/istio-testing/blah`, then push it with `local-registry/blah`, so that it can be served as `local-registry/blah` and pulled by proxies in the test cluster. Not sure if there is a cleaner way to achieve this (probably the redirection you mentioned, I can give it a try).

pkg/test/fakes/imageregistry/main.go
```diff
@@ -0,0 +1,113 @@
...
+		w.Header().Set("WWW-Authenticate", "Basic")
+		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+		return
+	}
+	h.registry.ServeHTTP(w, r)
+}
+
+func main() {
+	flag.Parse()
+	s := &http.Server{
```
zirain, line 62: do we need support SSL?

bianpengyuan: I think it is fine to serve and test with only plaintext? SSL or not SSL is not the really relevant to Wasm image pulling.

bianpengyuan: I don't think we need integration test to verify that? Should be easy to just have a unit test with a test server?

pkg/test/fakes/imageregistry/main.go
bianpengyuan: Updated to redirect.

pkg/test/fakes/imageregistry/main.go
bianpengyuan: Yeah same feeling. If the concern is whether env variable can be applied to proxy via ProxyMetadata, that path should already well covered by other tests.

pkg/test/fakes/imageregistry/Makefile
```diff
@@ -0,0 +1,33 @@
...
+# NOTE: TAG should be updated whenever changes are made in this directory
+# This should also be updated in dependent components
+TAG := 1.0
+
+all: build_and_push clean
+
+build_and_push:
+	cd $(MD_PATH) && CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -tags netgo -ldflags '-w -extldflags "-static"' main.go
+	docker build $(MD_PATH) -t $(IMG):$(TAG)
+	docker push $(IMG):$(TAG)
```
howardjohn, line 30:
```suggestion
	docker buildx build $(MD_PATH) -t $(IMG):$(TAG) --push
```

howardjohn (approved):

PR #37451, opened 2022-02-21
add `/debug/ecdsz`

zirain:

right now ecds is not a part of configdump, this PR make you can debug with endpoint `/debug/ecdsz` 

demo usage:

```console
kubectl exec -it istiod-7db4959cf9-btrj2 -nistio-system -- pilot-discovery request get /debug/ecdsz?proxyID=httpbin-74fb669cc6-mxhpx
```

```json
[
    {
        "config": {
            "name": "default.httpbin-rate-limiting",
            "Vm": {
                "VmConfig": {
                    "runtime": "envoy.wasm.runtime.v8",
                    "code": {
                        "Specifier": {
                            "Remote": {
                                "http_uri": {
                                    "uri": "oci://ghcr.io/tetratelabs/wasm-rate-limiting:v1",
                                    "HttpUpstreamType": {
                                        "Cluster": "_"
                                    },
                                    "timeout": {
                                        "seconds": 30
                                    }
                                }
                            }
                        }
                    }
                }
            },
            "configuration": {}
        }
    },
    {
        "config": {
            "name": "default.display-metadata",
            "Vm": {
                "VmConfig": {
                    "runtime": "envoy.wasm.runtime.v8",
                    "code": {
                        "Specifier": {
                            "Remote": {
                                "http_uri": {
                                    "uri": "oci://ghcr.io/zirain/display-metadata:latest",
                                    "HttpUpstreamType": {
                                        "Cluster": "_"
                                    },
                                    "timeout": {
                                        "seconds": 30
                                    }
                                }
                            }
                        }
                    },
                    "environment_variables": {
                        "host_env_keys": [
                            "POD_NAMESPACE"
                        ]
                    }
                }
            },
            "configuration": {}
        }
    }
]
```

@bianpengyuan  please take a first reviews.

### DIFF
--- a/pilot/pkg/xds/debug.go
+++ b/pilot/pkg/xds/debug.go
@@ -171,6 +171,7 @@ func (s *DiscoveryServer) AddDebugHandlers(mux, internalMux *http.ServeMux, enab
 		s.addDebugHandler(mux, internalMux, "/debug/force_disconnect", "Disconnects a proxy from this Pilot", s.forceDisconnect)
 	}
 
+	s.addDebugHandler(mux, internalMux, "/debug/ecdsz", "Status and debug interface for ECDS", s.ecdsz)
 	s.addDebugHandler(mux, internalMux, "/debug/edsz", "Status and debug interface for EDS", s.Edsz)
 	s.addDebugHandler(mux, internalMux, "/debug/ndsz", "Status and debug interface for NDS", s.ndsz)
 	s.addDebugHandler(mux, internalMux, "/debug/adsz", "Status and debug interface for ADS", s.adsz)
@@ -578,6 +579,40 @@ func (s *DiscoveryServer) adsz(w http.ResponseWriter, req *http.Request) {
 		return adsClients.Connected[i].ConnectionID < adsClients.Connected[j].ConnectionID
 	})
 	writeJSON(w, adsClients)
+}
+
+// ecdsz implements a status and debug interface for ECDS.
+// It is mapped to /debug/ecdsz
+func (s *DiscoveryServer) ecdsz(w http.ResponseWriter, req *http.Request) {
+	if s.handlePushRequest(w, req) {
+		return
+	}
+	proxyID, con := s.getDebugConnection(req)
+	if con == nil {
+		s.errorHandler(w, proxyID, con)
+		return
+	}
+
+	if s.Generators[v3.ExtensionConfigurationType] != nil {
+		// TODO: make ECDS generator support empty ResourceNames
+		wasmPlugins := s.globalPushContext().WasmPlugins(con.proxy)
+		names := make([]string, 0, len(wasmPlugins))
+		for _, list := range wasmPlugins {
+			for _, p := range list {
+				names = append(names, p.ExtensionConfiguration.Name)
+			}
+		}
+		r := &model.WatchedResource{
+			ResourceNames: names,
+		}
+		ecds, _, _ := s.Generators[v3.ExtensionConfigurationType].Generate(con.proxy, s.globalPushContext(), r, nil)
+		if len(ecds) == 0 {
+			w.WriteHeader(http.StatusBadRequest)
+			_, _ = w.Write([]byte(fmt.Sprintf("get len(ecds)==0, names: %s\n", names)))
+			return
+		}
+		writeJSON(w, ecds)
+	}
 }
 
 // ConfigDump returns information in the form of the Envoy admin API config dump for the specified proxy
--- a/pilot/pkg/xds/debug.go
+++ b/pilot/pkg/xds/debug.go
@@ -594,21 +594,17 @@ func (s *DiscoveryServer) ecdsz(w http.ResponseWriter, req *http.Request) {
 	}
 
 	if s.Generators[v3.ExtensionConfigurationType] != nil {
-		// TODO: make ECDS generator support empty ResourceNames
-		wasmPlugins := s.globalPushContext().WasmPlugins(con.proxy)
-		names := make([]string, 0, len(wasmPlugins))
-		for _, list := range wasmPlugins {
-			for _, p := range list {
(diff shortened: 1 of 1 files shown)


bianpengyuan (commented):
Can we decode and unmarshal if the type is Wasm extension? If it is other type, it would also be useful to get the type of the extension, although probably not possible to unmarshal. Also notice that we are using env at VM config to passing image pull secret, so will need some redaction at the dump.

zirain:
> Can we decode and unmarshal if the type is Wasm extension? If it is other type, it would also be useful to get the type of the extension, although probably not possible to unmarshal. Also notice that we are using env at VM config to pass image pull secret, so will need some redaction at the dump.

I think we can decode and unmarshal, but maybe some perfermance issue when there're too much ecds configs.

BTW, I think `EcdsGenerator` do not respect the rule `If empty, all resources of the
 TypeUrl type are watched`, can you confirm that?
https://github.com/istio/istio/blob/56edd90bb930ff715474769e8c9587f8c74700eb/pilot/pkg/model/context.go#L320

zirain:
@bianpengyuan updated with unmarshal

pilot/pkg/xds/debug.go
```diff
@@ -580,6 +583,60 @@ func (s *DiscoveryServer) adsz(w http.ResponseWriter, req *http.Request) {
...
+	if s.handlePushRequest(w, req) {
+		return
+	}
+	proxyID, con := s.getDebugConnection(req)
+	if con == nil {
+		s.errorHandler(w, proxyID, con)
+		return
+	}
+
+	if s.Generators[v3.ExtensionConfigurationType] != nil {
```
howardjohn, line 598: maybe we should just make config_dump able to take in a list of generator keys. We can have shortcuts sa well. So something like `config_dump?types=ecds`.

Then we do not need a bunch of new methods for every type we add?

pilot/pkg/xds/debug.go
```diff
@@ -580,6 +583,60 @@ func (s *DiscoveryServer) adsz(w http.ResponseWriter, req *http.Request) {
...
+}
+
+func unmarshalToWasm(r *discovery.Resource) (*wasm.Wasm, error) {
+	tce := &core.TypedExtensionConfig{}
+	if err := r.GetResource().UnmarshalTo(tce); err != nil {
+		return nil, err
+	}
+
+	w := &wasm.Wasm{}
+	if err := tce.TypedConfig.UnmarshalTo(w); err != nil {
```
bianpengyuan, line 633: I think user probably would use EnvoyFilter to add ECDS resource which is not of type Wasm? We probably should not warn, but instead just dump it as unmarshalled typedconfig?

bianpengyuan:
> I think we can decode and unmarshal, but maybe some perfermance issue when there're too much ecds configs.

This is a debug interface, so performance is not much of a concern.

howardjohn:
We should be careful to not leak Secrets once the pullSecret PR is merged

bianpengyuan:
> We should be careful to not leak Secrets once the pullSecret PR is merged

Yeah, I added to the head of my priority list :).

pilot/pkg/xds/debug.go
zirain: right now, `adsz` have something like what your said, `confug_dump` will dump everything .

https://github.com/istio/istio/blob/e3904269b7bd95207fdccd2be6f14276a3ad470a/pilot/pkg/xds/debug.go#L177

https://github.com/istio/istio/blob/e3904269b7bd95207fdccd2be6f14276a3ad470a/pilot/pkg/xds/debug.go#L195

I'm not sure that `config_dump?types=ecds` is a good idea or not.

pilot/pkg/xds/debug.go
zirain: updated

pilot/pkg/xds/debug.go
```diff
@@ -580,6 +583,69 @@ func (s *DiscoveryServer) adsz(w http.ResponseWriter, req *http.Request) {
...
+			}
+		}
+
+		writeJSON(w, wasmCfgs)
+	}
+}
+
+const (
+	apiTypePrefix      = "type.googleapis.com/"
+	wasmHTTPFilterType = apiTypePrefix + "envoy.extensions.filters.http.wasm.v3.Wasm"
```
bianpengyuan, line 628: Looks like these const are repeated at several places. Can we consolidate them into model.go?

pilot/pkg/xds/debug.go
```diff
@@ -580,6 +583,69 @@ func (s *DiscoveryServer) adsz(w http.ResponseWriter, req *http.Request) {
...
+	if con == nil {
+		s.errorHandler(w, proxyID, con)
+		return
+	}
+
+	if s.Generators[v3.ExtensionConfigurationType] != nil {
+		r, ok := con.proxy.WatchedResources[v3.ExtensionConfigurationType]
+		if !ok {
+			w.WriteHeader(http.StatusNotFound)
+			_, _ = w.Write([]byte(fmt.Sprintf("not watched ExtensionConfigurationType found, proxyID: %s\n", proxyID)))
```
bianpengyuan, line 602:
```suggestion
			_, _ = w.Write([]byte(fmt.Sprintf("no watched ExtensionConfigurationType found, proxyID: %s\n", proxyID)))
```

pilot/pkg/xds/debug.go
zirain: actually yes, but I want refactor all this in a another single PR.

zirain: there's something same in `pkg/wasm/convert.go`, we'd find a better place where we can share these const.

pilot/pkg/xds/debug.go
zirain: tracking in https://github.com/istio/istio/issues/37504

bianpengyuan (approved):

PR #37452, opened 2022-02-21
`InsertedExtensionConfigurations` should return all configs when `nam…

zirain:
`InsertedExtensionConfigurations` should return all configs when `names` is empty

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/envoyfilter/extension_configuration_patch.go
+++ b/pilot/pkg/networking/core/v1alpha3/envoyfilter/extension_configuration_patch.go
@@ -40,9 +40,10 @@ func InsertedExtensionConfigurations(efw *model.EnvoyFilterWrapper, names []stri
 			log.Errorf("extension config patch %+v does not match TypeExtensionConfig type", p.Value)
 			continue
 		}
-		if hasName.Contains(ec.GetName()) {
-			result = append(result, proto.Clone(p.Value).(*core.TypedExtensionConfig))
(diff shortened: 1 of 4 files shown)


pilot/pkg/networking/core/v1alpha3/envoyfilter/extension_configuration_patch.go
```diff
@@ -40,9 +40,10 @@ func InsertedExtensionConfigurations(efw *model.EnvoyFilterWrapper, names []stri
 			log.Errorf("extension config patch %+v does not match TypeExtensionConfig type", p.Value)
 			continue
 		}
-		if hasName.Contains(ec.GetName()) {
-			result = append(result, proto.Clone(p.Value).(*core.TypedExtensionConfig))
+		if len(hasName) != 0 && !hasName.Contains(ec.GetName()) {
```
hzxuzhonghu, line 43: Actually, empty resource names are only for lds and cds, not apply to other xDS. Maybe we should update the comments to reflect it

bianpengyuan: Yeah ECDS should follow RDS and EDS, where according to spec: `the client must specify all resource names it is interested in with each request`: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol

zirain: OK, let's close this PR and update the comment.

PR #37453, opened 2022-02-21
[WIP]: Initial delta rds support for sidecars

hzxuzhonghu:

Currently, depend on rds cache,

// 1. if the route config exists in the cache, it is not updated.
// 2. if not exists in cache, it is added/updated
// 3. if not exists in cache and it is empty, it is removed.

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/httproute.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute.go
@@ -96,6 +96,51 @@ func (configgen *ConfigGeneratorImpl) BuildHTTPRoutes(
 	return routeConfigurations, model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", hit, hit+miss)}
 }
 
+// BuildDeltaHTTPRoutes generates the deltas (add and delete) for a given proxy.
+// It depends on the cache:
+// 1. if the route config exists in the cache, it is not updated.
+// 2. if not exists in cache, it is added/updated
+// 3. if not exists in cache and it is empty, it is removed.
+func (configgen *ConfigGeneratorImpl) BuildDeltaHTTPRoutes(proxy *model.Proxy, req *model.PushRequest,
+	routeNames []string) ([]*discovery.Resource, []string, model.XdsLogDetails, bool) {
+	var delta bool
+	var removed []string
+	routeConfigurations := make([]*discovery.Resource, 0)
+	efw := req.Push.EnvoyFilters(proxy)
+	switch proxy.Type {
+	case model.SidecarProxy:
+		delta = true
+		vHostCache := make(map[int][]*route.VirtualHost)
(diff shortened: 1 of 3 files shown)


pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -96,6 +96,51 @@ func (configgen *ConfigGeneratorImpl) BuildHTTPRoutes(
 	return routeConfigurations, model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", hit, hit+miss)}
 }
 
+// BuildDeltaHTTPRoutes generates the deltas (add and delete) for a given proxy.
+// It depends on the cache:
+// 1. if the route config exists in the cache, it is not updated.
```
hzxuzhonghu, line 101: not right here

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -96,6 +96,51 @@ func (configgen *ConfigGeneratorImpl) BuildHTTPRoutes(
...
+	efw := req.Push.EnvoyFilters(proxy)
+	switch proxy.Type {
+	case model.SidecarProxy:
+		delta = true
+		vHostCache := make(map[int][]*route.VirtualHost)
+		// dependent envoyfilters' key, calculate in front once to prevent calc for each route.
+		envoyfilterKeys := efw.Keys()
+		for _, routeName := range routeNames {
+			rc, cached := configgen.buildSidecarOutboundHTTPRouteConfig(proxy, req, routeName, vHostCache, efw, envoyfilterKeys)
+			// cache miss means http route is updated
```
howardjohn, line 118: not sure using cache is safe since its shared across many proxies

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -96,6 +96,51 @@ func (configgen *ConfigGeneratorImpl) BuildHTTPRoutes(
...
+			if !cached {
+				if rc != nil {
+					routeConfigurations = append(routeConfigurations, rc)
+				} else {
+					removed = append(removed, routeName)
+				}
+			}
+		}
+	case model.Router:
+		delta = false
```
adiprerepa, line 128: Why not using delta with proxy being a Router?

howardjohn:
Seems like we can/should do "incremental RDS" with SotW even. We currently
always send all routes - we can do like EDS and just send changes.

We don't actually get any benefits from Delta here unless we switch to
vhds, as far as I understand it.

hzxuzhonghu:
It is hard to get the similar behavior like eds, route is influenced by many types of resources, service/vs/dr. So it is not easy to infer which route change by config updates. I was thinking about we use cache token to infer which route changed, and i will file a proposal.

PR #37457, opened 2022-02-21
[TF] Simplify call validation

nmittler:
There is currently validation logic in several places (ParsedResponse, ParsedResponses, eecho.Validator).

This change moves all of the validation logic into the echo client library. In addition, it allows an optimization for being able to run several checks on each message, rather than having to process a single check across all messages.


pkg/test/echo/response/check/checker.go
```diff
@@ -0,0 +1,46 @@
...
+type Func func(r *response.Instance) error
+
+func (c Func) Check(r *response.Instance) error {
+	return c(r)
+}
+
+func (c Func) CheckCall(rs response.Instances, _ error) error {
+	for i, r := range rs {
+		if err := c.Check(r); err != nil {
+			return fmt.Errorf("response[%d]: %v", i, err)
```
howardjohn, line 42: is multierror useful or are we intentionally short-circuiting? seems like it could be useful to know 1 request failed vs all of them did?

pkg/test/echo/response/checkcall/checkers.go
```diff
@@ -0,0 +1,123 @@
...
+	return Func(func(r response.Instances, err error) error {
+		var lasterr error
+		for _, c := range checkers {
+			if e := c.CheckCall(r, err); e != nil {
+				lasterr = err
+				continue
+			}
+			return nil
+		}
+		return fmt.Errorf("all checks failed. Last error: %v", lasterr)
```
howardjohn, line 47: nit: says "all" failed but could be only one..?

depending on usage it might look weird too. if I say `or(code(200), code(201), code(202))`, I would get "all checks failed. Last error: expected code 202, got code 300".

pkg/test/echo/response/instances.go
```diff
@@ -0,0 +1,55 @@
...
+//  limitations under the License.
+
+package response
+
+import (
+	"fmt"
+)
+
+// Instances is an ordered list of parsed response objects.
+type Instances []*Instance
```
howardjohn, line 22: nit: do we need pointers here and ~everywhere else? I think these shouldn't be mutable and are not large so not sure pointer is needed?

tests/integration/pilot/common/routing.go
```diff
@@ -155,13 +157,11 @@ spec:
...
 				PortName: "http",
 				Count:    1,
-				Validator: echo.And(
-					echo.ExpectOK(),
-					echo.ValidatorFunc(
-						func(response echoclient.ParsedResponses, _ error) error {
-							return response.Check(func(_ int, response *echoclient.ParsedResponse) error {
-								return ExpectString(response.RawResponse["Istio-Custom-Header"], "user-defined-value", "request header")
-							})
+				Checker: check.And(
```
howardjohn, line 160: super nit: its an extremely common pattern to have `And(ok, some-checker`). We could introduce a struct field that always does this (eg `CheckerOk: some-checker`), a helper (`Checker: check.OKAnd(some-checker)`), or even make checkall smarter and somehow always do "OK" if there is not an explicit "expect error"?

tests/integration/pilot/common/routing.go
```diff
@@ -155,13 +157,11 @@ spec:
...
+					check.OK(),
+					check.Func(
```
howardjohn, line 162: nit: is it possible to not have the checker interface and instead just have it be a `func(r *response.Instance) error` type alias? Then we don't need to wrap everything?

pkg/test/echo/response/check/checker.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"fmt"
+
+	"istio.io/istio/pkg/test/echo/response"
+	"istio.io/istio/pkg/test/echo/response/checkcall"
+)
+
+// Checker inspects echo responses for errors.
+type Checker interface {
```
howardjohn, line 25: I am not sure I understand the point of this interface. It seems all usages are checkcall. I understand the distinction (some checks can apply to a single request (and therefor also all requests) vs others need all requests) but it seems like an internal distinction. 

This is more confusing when we have the same method in both. I wouldn't know how to use check.And vs checkall.And, and I suspect many others would not.

Would it make sense to consolidate these to a single `check.Checker` interface which acts on response.Instances? For the checkers that act on only one response at a time that can be abstracted out with a common (private) function that upconverts single checkers to multi checkers.

Not sure that makes sense - LMK what you think

pkg/test/echo/client/client.go
```diff
@@ -98,21 +99,21 @@ func (c *Instance) Close() error {
 	return nil
 }
 
-func (c *Instance) Echo(ctx context.Context, request *proto.EchoRequest) (*ParsedResponse, error) {
+func (c *Instance) Echo(ctx context.Context, request *proto.EchoRequest) (*response.Instance, error) {
```
howardjohn, line 102: mentioned earlier, but might be an opportunity to stop using pointers.

Also while I am not the nitpicks, I would have a slight prefernce for this to be `echo.Response` and `echo.Responses`... not a big deal but we have a lot of "Instance"s that get confusing

tests/integration/pilot/common/routing.go
```diff
@@ -155,13 +157,11 @@ spec:
...
+						func(r *echoclient.Instance) error {
+							return ExpectString(r.RawResponse["Istio-Custom-Header"], "user-defined-value", "request header")
```
howardjohn, line 164: nit: could use check.Key

howardjohn: that does lose the "help" bit.

Options:
* Just lose the help bit
* Add new check.WithHelp(check.Key(foo, bar), "some help message")
* Make ExpectString the same but return a check.Func so we don't always need to wrap this two levels deep.

With some of the other comments on "check.OK" wthis would look like `Checker: check.Key("Istio-Custom-Header", "user-defined-value")` or similar which is super clean

tests/integration/security/ca_custom_root/trust_domain_validation_test.go
```diff
@@ -27,9 +27,10 @@ import (
 	"testing"
 	"time"
 
-	client2 "istio.io/istio/pkg/test/echo/client"
 	"istio.io/istio/pkg/test/echo/common/scheme"
 	epb "istio.io/istio/pkg/test/echo/proto"
+	client2 "istio.io/istio/pkg/test/echo/response"
```
howardjohn, line 32: nit: make this not `client2`

howardjohn (approved):
Nice! This brought out a lot of nit picks but this is definitely an improvement and step in the right direction

pkg/test/echo/response/check/checker.go
nmittler: I waffled back and forth on this for a while as well. The main motivation for the `checker` package was failing fast on messages, but I'm not actually sure how much of an optimization that is in practice. Let me try merging the two as you suggested.

pkg/test/echo/response/checkcall/checkers.go
nmittler: I think `Or` is confusing altogether. I've removed it and modified the only call site that uses it.

tests/integration/pilot/common/routing.go
nmittler: I think I'll defer this particular change for now. We can always add this later.

bianpengyuan (approved):

PR #37465, opened 2022-02-22
Enhance delta xds

hzxuzhonghu:

FYI: https://docs.google.com/document/d/10l6wsJP2swM_IWGb8mrKA-Vargd5xToeELVmLdfbBJY/edit#

First step:  refactor xds cache to return token.
Second:  Delta CDs build

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -39,13 +39,8 @@ import (
 	"istio.io/istio/pilot/pkg/serviceregistry/provider"
 	"istio.io/istio/pkg/config/host"
 	"istio.io/istio/pkg/config/protocol"
-	"istio.io/istio/pkg/config/schema/gvk"
 	"istio.io/istio/pkg/util/sets"
 )
-
-// deltaConfigTypes are used to detect changes and trigger delta calculations. When config updates has ONLY entries
-// in this map, then delta calculation is triggered.
-var deltaConfigTypes = sets.NewWith(gvk.ServiceEntry.Kind)
 
 // getDefaultCircuitBreakerThresholds returns a copy of the default circuit breaker thresholds for the given traffic direction.
 func getDefaultCircuitBreakerThresholds() *cluster.CircuitBreakers_Thresholds {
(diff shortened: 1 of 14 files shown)


pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -90,62 +85,17 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, req *mod
...
-				oldPorts := servicePorts[service.Hostname.String()]
-				for port, cluster := range oldPorts {
-					// if this service port is removed, we can conclude that it is a removed cluster.
-					if _, exists := service.Ports.GetByPort(port); !exists {
-						deletedClusters = append(deletedClusters, cluster)
-					}
-				}
-			}
-		}
+	// In Sotw, we care about all services.
```
howardjohn, line 91: a bit confused since this is not Sotw?

hzxuzhonghu: updated

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -80,62 +75,18 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, req *mod
 	return configgen.buildClusters(proxy, req, services)
 }
 
-// BuildDeltaClusters generates the deltas (add and delete) for a given proxy. Currently, only service changes are reflected with deltas.
-// Otherwise, we fall back onto generating everything.
-func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, updates *model.PushRequest,
+// BuildDeltaClusters generates the deltas (add and delete) for a given proxy.
+func (configgen *ConfigGeneratorImpl) BuildDeltaClusters(proxy *model.Proxy, req *model.PushRequest,
```
ramaraochavali, line 79: Couple of questions:
- So in Delta we generate all the resources as usual and later filter based on cache token? Then why do not we call regular BuildClusters rather than repeating the logic here?
- Do we still need, usedDelta?

howardjohn: > So in Delta we generate all the resources as usual and later filter based on cache token? Then why do not we call regular BuildClusters rather than repeating the logic here?

If this is the case (didn't read too much), that would be not great. Since control plane load will still be just as high as without delta mostly.

Can we just construct the key (cheap) and see if they key depends on the recent push?

ramaraochavali: I agree. That was my next questions if the answer to my first question is "Yes". It is not definetly not great, we are only optimizing on network (with additional overhead of filtering in control plane) if I read the code correctly?

hzxuzhonghu: >So in Delta we generate all the resources as usual and later filter based on cache token? Then why do not we call regular BuildClusters rather than repeating the logic here?

We can not use `BuildClusters without any change, but can refactor it to return the token info. Then we can decide the delta changes.

>Can we just construct the key (cheap) and see if they key depends on the recent push?

I did not thinking too much about this, worth looking into. 

>If this is the case (didn't read too much), that would be not great. Since control plane load will still be just as high as without delta mostly.

Yeah, currently it only saves bandwidth between pilot and sidecar, which is still great from our performance test last year.  

And also helps a lot for envoy config hot reloading.

hzxuzhonghu: >>Can we just construct the key (cheap) and see if they key depends on the recent push?

This will introduce another cache, for CDS the key is cluster name, the value is the clusterkey. Then is it easy to get the updated clusternames by comparing the clusterkeys.

How about deleted clusters? How could we know?

ramaraochavali: > Yeah, currently it only saves bandwidth between pilot and sidecar, which is still great from our performance test last year.

While it still great, what is our longer term plan? Our longer term plan is to be more "delta" aware so that we optimize on controlplane also along with network optimization. So in that sense, taking this approach now - seems not a step in our long term direction or do you think we can go back to long term approach easily?

hzxuzhonghu: Why do you think depending on the cache is not ` more "delta" aware`? Or what do you think is `  more "delta" aware`?

Come back to the first question: `So in Delta we generate all the resources as usual and later filter based on cache token? `

When we generate all resources, if one need to regenerate, it absolutely means it is updated. So for it, no matter what kind of way we use, the effort is same.

For one in the cache, I means it has no change at all, and since it is in the cache, why it is heavy?

ramaraochavali: > Why do you think depending on the cache is not more "delta" aware? Or what do you think is more "delta" aware?

I am not saying using cache is not "delta" aware. All I am saying is irrespective of mechanism we choose
- We have to optimize compute in control plane as well, we can not just say it will be OK if we optimize on just network.

Based on what you described above it seems there is no additional overhead of Building all clusters and filtering them for delta in controlplane which is different from your original statement that we are only getting network optimization - That confused me.

hzxuzhonghu: >Based on what you described above it seems there is no additional overhead of Building all clusters and filtering them for delta in controlplane which is different from your original statement that we are only getting network optimization - That confused me.

sorry to make you confused, the truth is with cache, cpu/badwidth can both be saved for delta xds.

hzxuzhonghu:
@howardjohn @ramaraochavali @adiprerepa ptal

howardjohn:
Explored some alternative ideas in https://github.com/istio/istio/pull/37926. Let me know what you think. I would like to avoid building all the clusters and throwing them away

hzxuzhonghu:
ok, willl take a look

howardjohn:
BTW I don't think that will work exactly out of the box. it doesn't handle deletes and I don't think it handles cases where we just do `ClearCache` to wipe out everything (although this may not matter in practice...). Meant to be an example rather than working code

hzxuzhonghu:
@howardjohn Do you have any other idea, since we can not get a way to figure out the removed xds. I believe this is the best way currently we can think of.

hzxuzhonghu:
ping @ramaraochavali @howardjohn

howardjohn:
@hzxuzhonghu I think Rama and I have already mentioned we are not comfortable with an approach that depends on the contents of the cache for correctness

hzxuzhonghu:
As I said it is an improvement, though not perfect. Can you suggest a good way to handle it?

howardjohn:
I would prefer the current state than this in the mean time. I did have
some ideas for handling deletes by storing a map of the cache key inside
WatchedResources which would allow us to handle deletions I think. Planning
to try it out this week.

hzxuzhonghu:
We are planning to evaluate the performance on large scale multi-cluster before adopting delta to handle large scalability.

howardjohn:
Actually for deletions handling I think we can just look at what config we
sent last and compare

hzxuzhonghu:
The tricky part is for some resource kind,  it references to other resources, like VS, DR could point to a totally different host.

howardjohn:
ah I was mostly thinking only of CDS

hzxuzhonghu:
@howardjohn can you take another look, below is some drawbacks

- xDS generator should deal with all xDS resources rather than only the updates ones.  Fortunately, we have xDS cache, so only need to build xDSCache key actually.  
- As xds cache clear and xds generate are asynchronously at different times. So at one time if xds generate for push0 happens after xds cache cleared for push1, the delta xds generation procedure may not be as efficient as it should be. But it recovers eventually when the xds cache aligns with push context.
image


Hope it is not a blocker for this.

howardjohn:
On "generate xds for t1" - the cache is initially empty. So it is inefficient the first push for t1 as well, isn't it?

hzxuzhonghu:
> On "generate xds for t1" - the cache is initially empty. So it is inefficient the first push for t1 as well, isn't it?

It could be if the clear action is empty all cache.  In this case, I think the current delta algorithm based on SE change is also generate all.   No difference. 

For ther case, clear just a small part of cache, the cost is relatively light weight

howardjohn:
I have re-implemented my prototype: https://github.com/istio/istio/pull/37926. This now handles, as far as I know, 100% of cases including subsets and deletion. It does not depend on the XDS cache at all (it could be disabled and still function), but uses the same invalidation logic.

I strongly prefer an approach like this since it is deterministic (not based on things like whether we happened to have the cached filled or not) which is more consistent and efficient. It is more flexible since it does not depend on the XDS cache, and also more scalable since the cache has a max size. It has similar benefits to this approach that we do not need to re-invent a bunch of logic to determin config dependencies. We only generate the cache key for each object, rather than the entire object (the expensive part!).

hzxuzhonghu:
WIll take a look, hope it lands down asap.

PR #37358, opened 2022-02-15
[TF] Remove security Checker

nmittler:
**Please provide a description of this PR:**

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #37475, opened 2022-02-22
[TF] Cleaning up status code logic

nmittler:
Echo currently use a string to represent the status code throughout.

This PR switches to using an integer code (to the extend possible) and using the constants defined in the HTTP library.


bianpengyuan (approved):

lei-tang (approved):

pkg/test/echo/check/checkers.go
```diff
@@ -94,19 +95,52 @@ func Error() Checker {
...
-			return fmt.Errorf("expected response code %s, got %q", expected, r.Code)
+		if r.Code != expectedStr {
+			return fmt.Errorf("expected response code `%s`, got %q", expectedStr, r.Code)
 		}
 		return nil
 	})
 }
 
+// TooManyRequests checks that at least one message receives a StatusTooManyRequests status code.
+func TooManyRequests() Checker {
```
howardjohn, line 131: Not concerned yet but in general we should be conscious of localized Checkers vs ones here. If we end up with too many niche ones it will make usage harder.

nmittler: Agree in general. The ones here seem pretty general-purpose. TBH I'm not sure how many specialized ones we'll have at the end of the day. I guess we'll see as we continue to refactor.

tests/integration/security/util/authn/authn_util.go
```diff
@@ -52,30 +52,29 @@ func (c *TestCase) String() string {
...
-		matcher := fmt.Sprintf("%s=%s", k, v)
-		if len(v) == 0 {
-			if strings.Contains(responses[0].Body, matcher) {
-				return fmt.Errorf("%s: expect header %s does not exist, got response\n%s", c, k, responses[0].Body)
+	return check.And(
+		check.StatusCode(c.ExpectResponseCode),
+		check.Each(func(r echoclient.Response) error {
+			// Checking if echo backend see header with the given value by finding them in response body
+			// (given the current behavior of echo convert all headers into key=value in the response body)
+			for k, v := range c.ExpectHeaders {
```
howardjohn, line 60: outside scope but this should use Key() probably

howardjohn (approved):

tests/integration/security/util/authn/authn_util.go
nmittler: Yeah, I'm going to address this in https://github.com/istio/istio/pull/37358

PR #37484, opened 2022-02-23
[TF] Top-level header fields in echo response

nmittler:
We currently don't have a way to distinguish between request and response headers when checking echo results. This promotes request and response headers to a top-level field in echo.Response and provides tools for validating headers.

### DIFF
--- a/tests/integration/pilot/common/routing.go
+++ b/tests/integration/pilot/common/routing.go
@@ -158,7 +158,7 @@ spec:
 				Count:    1,
 				Check: check.And(
 					check.OK(),
-					check.WithInfo("request header", check.Key("Istio-Custom-Header", "user-defined-value"))),
+					check.RequestHeader("Istio-Custom-Header", "user-defined-value")),
 			},
 			workloadAgnostic: true,
 		},
@@ -185,7 +185,7 @@ spec:
 				Count:    1,
 				Check: check.And(
 					check.OK(),
-					check.WithInfo("added request header", check.Key("X-Custom", "some-value"))),
+					check.RequestHeader("X-Custom", "some-value")),
 			},
 			workloadAgnostic: true,
 		},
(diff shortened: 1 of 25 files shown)


tests/integration/pilot/common/routing.go
```diff
@@ -520,10 +524,12 @@ spec:
 							Count:    1,
 							Check: check.And(
 								check.OK(),
-								check.WithInfo("CORS preflight header", check.Key("Access-Control-Allow-Origin", "cors.com")),
-								check.WithInfo("CORS preflight header", check.Key("Access-Control-Allow-Methods", "POST,GET")),
-								check.WithInfo("CORS preflight header", check.Key("Access-Control-Allow-Headers", "X-Foo-Bar,X-Foo-Baz")),
-								check.WithInfo("CORS preflight header", check.Key("Access-Control-Max-Age", "86400"))),
+								check.ResponseHeaders(map[string]string{
```
howardjohn, line 527: nit: slightly prefer taking in 2 strings as input. This is likely one of the only cases we will want to check multiple headers and can just call the Response Header checker multiple times

tests/integration/telemetry/outboundtrafficpolicy/helper.go
```diff
@@ -259,29 +256,10 @@ func RunExternalRequest(cases []*TestCase, prometheus prometheus.Instance, mode
...
-								for k, v := range tc.Expected.Metadata {
-									if got := r.RawResponse[k]; got != v {
-										return fmt.Errorf("expected metadata %v=%v, got %q", k, v, got)
-									}
-								}
-							}
-							return nil
-						},
+						Check: check.And(
+							check.NoError(),
```
howardjohn, line 260: I think we expect error in some cases

tests/integration/pilot/common/routing.go
howardjohn: Ok actually there was one other usage it looks like. Still my preference but I'm fine either way

howardjohn (commented):
did not approve due to the telemetry test thing, not the map vs string input - let me know if I just misunderstood

tests/integration/pilot/common/routing.go
nmittler: I've added both and one calls the other.

bianpengyuan (approved):

PR #37498, opened 2022-02-23
fixed the https jwt test issues

aryan16:
1) The major issue was with the networking, istiod tries to fetch the public key from jwsuri when the containers are not ready to serve though it shows the status `running` using kubectl command, but it is not ready in real. So following is the solution for that - 

Increased the number of retries in `istiod` when it tries to fetch public key from `jwksuri` - The number is set after rigorous testing and it is the least possible retry number we can select.

2) Added readiness probe for making the process better.


pilot/pkg/model/jwks_resolver.go
```diff
@@ -67,7 +67,7 @@ const (
 	// How many times should we retry the failed network fetch on main flow. The main flow
 	// means it's called when Pilot is pushing configs. Do not retry to make sure not to block Pilot
 	// too long.
-	networkFetchRetryCountOnMainFlow = 0
+	networkFetchRetryCountOnMainFlow = 5
```
yangminzhu, line 70: This is on the listener generation path, not sure if `5` is too high for it, probably set it to a smaller value `1` or `2`? 

I think @howardjohn has better insights on this.

aryan16: I have tried 1, 2, 3 and it still fails with that.

aryan16: If we don't want to increase the retries, other option is to reduce the refresh interval so that we can try to fetch the public key in background.

howardjohn: We should not have any retries on the push path - I don't even think we should have any calls at all. We should do it all async in the background. When a user creates a new ReqAuth, we do not push it immediately but schedule a background job to fetch the jwks. Then once it fetches we push.

This is the pattern in https://github.com/istio/istio/blob/e734af4be2312475bea8c9367abb4138cb44ae16/pilot/pkg/model/network.go#L396.

howardjohn (changes requested):

pilot/pkg/model/jwks_resolver.go
yangminzhu: The other option is as john suggested, completely refactor the code, remove the call on the push flow and fetch the jwksUri asynchronously. I think this probably does not need much code changes since we already have the JwksResolver here (it is already started as a separate job for the jwks refresh, just need to update it to handle the push path use case as well): https://github.com/istio/istio/blob/cd004f1ba7f9bab3951824912af4119074ec9a00/pilot/pkg/model/jwks_resolver.go#L154

aryan16: the time for retry in async process is so high - https://github.com/istio/istio/blob/cd004f1ba7f9bab3951824912af4119074ec9a00/pilot/pkg/model/jwks_resolver.go#L58
By that time out test ends, is it okay to decrease this to some smaller time ?

aryan16: @yangminzhu correct me if I am wrong do we not have this logic already ? We have a async function which hits the jwks uri continuously after a fixed interval of time and add the key to the cache.

yangminzhu: The idea here is to do some refactor, let the main listener push flow to trigger the refresh() job to fetch the jwksUri asynchronously, once the refresh() job fetched the jwks, it needs to trigger an extra config push. The refresh() can retry with a much smaller interval if it failed, there is a separate refresh interval for the refresher() job to rotate jwks keys (that are already fetched successfully).

yangminzhu: @howardjohn By the way, if we remove the fetch from the main flow completely, it seems we will always trigger a second config push right after the first one (when the jwksUri fetching is done), is this kind of consecutive config push desired?

howardjohn: Well it would only be when there is a new JWKS added. That must be extremely rare. If the code is smart enough it will not push on the first creation and only when the jwks is resolved

aryan16: we can modify the push, if it doesn't have any pubkey we can stop that. Right ?

pilot/pkg/model/jwks_resolver.go
```diff
@@ -247,19 +251,22 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
...
 		var resp []byte
-		resp, err = r.getRemoteContentWithRetry(jwksURI, networkFetchRetryCountOnMainFlow)
+		resp, err = r.getRemoteContentWithRetry(jwksURI, 0)
 		if err != nil {
 			log.Errorf("Failed to fetch public key from %q: %v", jwksURI, err)
+			// running the process of fetching public key in background
+			r.refreshTicker.Stop()
+			r.refreshIntervalOnFailure = time.Second
+			r.refreshInterval = time.Second
+			go r.refresher()
```
yangminzhu, line 261: I think you should reuse the same refresher routine instead of starting a new one.

You can do some refactor to let you just add the new jwksUri to the refresher. The fetching/retry/etc. (e.g. internal) should be internal logic to the refresher and not exposed here.

pilot/pkg/model/jwks_resolver.go
```diff
@@ -380,13 +387,17 @@ func (r *JwksResolver) getRemoteContentWithRetry(uri string, retry int) ([]byte,
 func (r *JwksResolver) refresher() {
 	// Wake up once in a while and refresh stale items.
 	r.refreshTicker = time.NewTicker(r.refreshInterval)
-	lastHasError := false
+	lastHasError := tickerCounter - 1
+	if r.refreshIntervalOnFailure == time.Second {
```
yangminzhu, line 391: This looks like horrible hack to compare with a constant `time.Second` ...

pilot/pkg/model/jwks_resolver.go
```diff
@@ -474,6 +485,11 @@ func (r *JwksResolver) refresh() bool {
 				hasChange = true
 				log.Infof("Updated cached JWT public key from %q", jwksURI)
 			}
+
+			// making the refreshIntervalOnFailure to original state once public key is successfully fetched
+			if r.refreshIntervalOnFailure == time.Second {
```
yangminzhu, line 490: same here, please do not do it like this

pilot/pkg/model/jwks_resolver.go
```diff
@@ -247,19 +251,22 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
 		log.Errorf("Failed to jwks URI from %q: %v", issuer, err)
 	} else {
 		var resp []byte
-		resp, err = r.getRemoteContentWithRetry(jwksURI, networkFetchRetryCountOnMainFlow)
+		resp, err = r.getRemoteContentWithRetry(jwksURI, 0)
```
yangminzhu, line 254: This should be removed instead of just passing `0`.

pilot/pkg/model/jwks_resolver.go
aryan16: will that be okay if I add some extra variables in `JwksResolver` struct ?https://github.com/istio/istio/blob/58bf51baa71909c1b16a90d7d5acaf5b2c499a9a/pilot/pkg/model/jwks_resolver.go#L116

pilot/pkg/model/jwks_resolver.go
aryan16: cool, will change this!!

pilot/pkg/model/jwks_resolver.go
aryan16: cool, thanks!!

pilot/pkg/model/jwks_resolver.go
aryan16: And I have to do something with `refreshTicker` in this method as `refresher()` method is always running. Will try to rest the ticker like `r.refreshTicker.Reset(time.Second)`.

pilot/pkg/model/jwks_resolver.go
```diff
@@ -238,29 +244,28 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
...
-		pubKey = string(resp)
-	}
+		// reset the ticker to hit jwksuri more often from istiod
+		key.jwksURI = jwksURI
+		r.refreshTicker.Reset(time.Second)
+		r.refreshIntervalOnFailure = time.Second
+		r.tickerCounter = 5
 
+	}
 	r.keyEntries.Store(key, jwtPubKeyEntry{
```
aryan16, line 262: had to pass empty public key because we can't wait for the cache to be updated, if we are waiting for it then there is no difference in making request to jwksuri in the main flow.

yangminzhu (commented):
Some high level comments
1) we should just remove the networkFetchRetryCountOnRefreshFlow instead of setting it to 0
2) refactor the GetPublicKey(), it should no longer call getRemoteContentWithRetry() and instead it notifies the refresh job to fetch the new jwks uri (if not already found in cache)
3) some logging statements need to be updated to align with the new flow, for example, the error logging in BuildLocalJwks() will need to be updated otherwise it will always log an error 100% times

aryan16:
Please ignore the recent changes.

PR #37499, opened 2022-02-23
Remove unused fields from FilterChain struct

jewertow:
This pull request just removes unnecessary fields from `FilterChain` struct. Some of them are set, but not read and some are not used at all.


howardjohn (approved):
/ok-to-test

hzxuzhonghu (approved):

Issue #33335, opened 2021-06-08
WorkloadAgnostic tests have excessive config churn causing slow tests


howardjohn:
SetupForServicePair is called for each job. This means new config is applied for each pair.

For a typical job, say virtualservice/rewrite_uri, we end up with 9 pairs. This causes the test to take ~10s, which isn't _that_ slow but it adds up. Just 6 months ago pilot tests took 8 minutes, now they are 25min+.

I think just switching SetupForServicePair with SetupForDestination is not sufficient either unless we invert the ordering to group destinations (right now its source/destination)

This is also in a fast enviornment, config updates are extremely slow when running somewhere like GKE

cc @stevenctl

ryantking:
Moving to the 1.12 milestones due to recent inactivity

howardjohn:
It seems like almost all tests are just looking at dst. We could refactor it to be `to/from` instead of `from/to` and add an optimization to  not delete+re-apply config if its identical, or make it more explicit.

howardjohn:
Not stale

stevenctl:
not-stale

howardjohn:
not stale

howardjohn:
not stale

howardjohn:
not stale

PR #37501, opened 2022-02-23
WIP: Optimize config applying

howardjohn:
Fixes https://github.com/istio/istio/issues/33335

### DIFF
--- a/tests/integration/pilot/common/traffic.go
+++ b/tests/integration/pilot/common/traffic.go
@@ -86,6 +86,7 @@ type TrafficTestCase struct {
 
 	// minIstioVersion allows conditionally skipping based on required version
 	minIstioVersion string
+	topConfig       bool
 }
 
 var noProxyless = echotest.Not(echotest.FilterMatch(echo.IsProxylessGRPC()))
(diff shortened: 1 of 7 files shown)


tests/integration/pilot/common/traffic.go
```diff
@@ -193,6 +241,98 @@ func (c TrafficTestCase) RunForApps(t framework.TestContext, apps echo.Instances
...
+	v, f := p["_dst_written"]
+	return f && v != nil && v.(bool)
+}
+
+func (p tmplParams) WroteSrc() bool {
+	v, f := p["_src_written"]
+	return f && v != nil && v.(bool)
+}
+
+func findReferences(t test.Failer, n int, vars func(src echo.Callers, dest echo.Instances) map[string]interface{}, config string, echoT *echotest.T) (bool, bool) {
```
stevenctl, line 326: cool

stevenctl: I assume this works for nested access?

stevenctl:
Not sure if your run was lucky – looks like postsubmit is anywhere from 18m to ~25m, but seems like a really good improvement. 

Maybe a future cleanup would be moving some of our traffic test logic into echotest since we essentially have 2 frameworks here.

howardjohn:
> Maybe a future cleanup would be moving some of our traffic test logic into echotest since we essentially have 2 frameworks here.


Strongly agree

PR #37505, opened 2022-02-24
Make analysis interval configurable

litong01:


manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
```diff
@@ -153,6 +153,10 @@ spec:
             value: istiod{{- if not (eq .Values.revision "") }}-{{ .Values.revision }}{{- end }}.{{ .Release.Namespace }}.svc:15012
           - name: PILOT_ENABLE_ANALYSIS
             value: "{{ .Values.global.istiod.enableAnalysis }}"
+{{- if .Values.global.istiod.enableAnalysis }}
```
howardjohn, line 156: Should not add a config here, users can set arbitrary env vars already (values.pilot.env.foo=bar)

releasenotes/notes/30200.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: istioctl
+issue:
+  - 30200
+releaseNotes:
+  - |
+    **Fixed** wasteful re-runs of analyzer by making the analysis run interval configurable
```
howardjohn, line 8: Well its not really fixed.. if you configure it to be higher then you just get stale data.

Seems like ideally we would have event based analysis. Not suggesting we do that since its probably quite challenging...

litong01: @howardjohn Thanks for reviewing this PR so fast, the idea here is to provide an option to allow admin to check things according to their own need. Mitch and I discussed this a bit, event based idea seems require a more code and  watching a lot of resources can also add load to a cluster, the pros and cons certainly debatable.  As you've noted it may ask for a lot of code.  The timer based implementation actually is simpler, it may be the reason why it was implemented currently, with this PR, an admin can manipulate this parameter based on their understanding of their own cluster, they can observe their cluster by setting different values over a period of time until their cluster is relatively stable.

manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
litong01: add it here so that user can do --set values.global.istiod.analysisInterval=30s from command line like they set enableAnalysis. If you @howardjohn , I can certainly remove that.

howardjohn:
Yes, we have 100s of env vars, they are for experimental purpose so should
not float into our real APIs

litong01:
> Yes, we have 100s of env vars, they are for experimental purpose so should not float into our real APIs
> […](#)
> On Wed, Feb 23, 2022 at 5:32 PM Tong Li ***@***.***> wrote: ***@***.**** commented on this pull request. 
Will do. Thanks John.

releasenotes/notes/30200.yaml
therealmitchconnors: Let's change the `kind` of this release-note to `feature` to better reflect that we are not closing a bug here.

manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
therealmitchconnors: I agree with John.  Users can already set arbitrary environment variables usnig the `values.pilot.env` section, so there's no need to build a dedicated values.yaml field for this.

operator/pkg/validate/common.go
```diff
@@ -180,6 +181,25 @@ func validateIntRange(path util.Path, val interface{}, min, max int64) util.Erro
 	return util.NewErrs(err)
 }
 
+// validateAnalysisInterval checks whether val is valid time duration and greater than 1 second.
+func validateAnalysisInterval(path util.Path, val interface{}) util.Errors {
+	k := reflect.TypeOf(val).Kind()
```
therealmitchconnors, line 186: I believe type validation should not be necessary as `env.RegisterDurationVar().Get()` already returns a duration.  The only thing we need to check is that the value is greater than 1s, and that can be done in pilot, rather than the operator.

tests/integration/pilot/analysis/analysis_test.go
```diff
@@ -40,6 +41,21 @@ func TestStatusExistsByDefault(t *testing.T) {
 		NotImplementedYet(features.Usability_Observability_Status_DefaultExists)
 }
 
+func TestAnalysisIntervalExistsByDefault(t *testing.T) {
```
therealmitchconnors, line 44: Once we remove the operator code, I don't think we will need an integ test for this change.  We can assume that  the logic for setting environment variables has been effectively tested elsewhere.

therealmitchconnors (changes requested):
The changes to pilot look great.  I don't think we need the changes to the operator or manifests folders, as users can already set arbitrary environment variables without special entries in values.yaml.

operator/pkg/validate/common.go
litong01: will do.

pilot/pkg/features/pilot.go
```diff
@@ -298,6 +298,20 @@ var (
...
 	).Get()
 
+	AnalysisInterval = func() time.Duration {
+		val, _ := env.RegisterDurationVar(
+			"PILOT_ANALYSIS_INTERVAL",
+			10000*time.Millisecond,
+			"If analysis is enabled, pilot will run istio analyzers using this value as interval in seconds "+
+				"Istio Resources",
+		).Lookup()
+		if val == 0 {
```
therealmitchconnors, line 308: What if the duration is set less than 1 second but more than 0?

releasenotes/notes/30200.yaml
therealmitchconnors: Probably also change the word **Fixed** above.  You can look through the other release notes to see what new features usually look like...

operator/pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -455,6 +455,7 @@ message STSConfig {
 message IstiodConfig {
   // If enabled, istiod will perform config analysis
   google.protobuf.BoolValue enableAnalysis = 2;
+  google.protobuf.Duration analysisInterval = 3;
```
therealmitchconnors, line 458: This is no longer needed.

operator/pkg/apis/istio/v1alpha1/values_types.pb.go
```diff
@@ -4456,307 +4464,308 @@ func init() {
...
-	0x6f, 0x7f, 0x77, 0xa7, 0x7a, 0x03, 0x15, 0x61, 0x6a, 0x67, 0x77, 0xaf, 0xf6, 0xe2, 0xa0, 0x59,
-	0xcd, 0x21, 0x80, 0xc9, 0x46, 0x53, 0xdb, 0xdf, 0x6e, 0x56, 0xc7, 0xd0, 0x14, 0xe4, 0x9f, 0xef,
-	0xed, 0x55, 0xf3, 0x0f, 0x7e, 0x17, 0x04, 0x85, 0x0c, 0x2d, 0x2c, 0x62, 0xf5, 0x06, 0x2a, 0xc5,
-	0x0c, 0x71, 0x35, 0xc7, 0xc4, 0x48, 0x13, 0x5d, 0x1d, 0x63, 0x9d, 0xc4, 0x2c, 0x5f, 0x35, 0x8f,
-	0xe6, 0xa0, 0x42, 0x1d, 0x6c, 0x6f, 0x63, 0xdb, 0xeb, 0x78, 0xb5, 0x16, 0xb6, 0xfd, 0xea, 0x38,
-	0x9a, 0x86, 0x71, 0x9b, 0xda, 0xb8, 0x3a, 0xb1, 0xb5, 0xf8, 0xcf, 0x3f, 0xdc, 0xb9, 0xf1, 0x87,
-	0x1f, 0xee, 0xdc, 0xf8, 0xb7, 0x1f, 0xee, 0xdc, 0x78, 0x15, 0xfe, 0x33, 0x98, 0x93, 0x49, 0xbe,
-	0x16, 0x1f, 0xfe, 0x57, 0x00, 0x00, 0x00, 0xff, 0xff, 0xd9, 0xa3, 0xbc, 0x80, 0x4b, 0x46, 0x00,
-	0x00,
+	// 4836 bytes of a gzipped FileDescriptorProto
```
therealmitchconnors, line 4467: This is no longer needed.

operator/pkg/apis/istio/v1alpha1/values_types.pb.go
```diff
@@ -1699,6 +1700,13 @@ func (m *IstiodConfig) GetEnableAnalysis() *protobuf.BoolValue {
 	return nil
 }
 
+func (m *IstiodConfig) GetAnalysisInterval() *types.Duration {
```
therealmitchconnors, line 1703: This is no longer needed.

operator/pkg/apis/istio/v1alpha1/values_types.pb.go
```diff
@@ -1663,6 +1663,7 @@ func (m *STSConfig) GetServicePort() uint32 {
 type IstiodConfig struct {
 	// If enabled, istiod will perform config analysis
 	EnableAnalysis       *protobuf.BoolValue `protobuf:"bytes,2,opt,name=enableAnalysis,proto3" json:"enableAnalysis,omitempty"`
+	AnalysisInterval     *types.Duration     `protobuf:"bytes,3,opt,name=analysisInterval,proto3" json:"analysisInterval,omitempty"`
```
therealmitchconnors, line 1666: This is no longer needed.

manifests/profiles/default.yaml
```diff
@@ -46,6 +46,7 @@ spec:
       istioNamespace: istio-system
       istiod:
         enableAnalysis: false
+        analysisInterval: 10s
```
therealmitchconnors, line 49: This is no longer needed.

therealmitchconnors (commented):
Looks like there's still some artifacts in the manifests and operator folders which need to be cleaned up...

pilot/pkg/features/pilot.go
litong01: I was wondering if 0.5 second is an ok value, not really sure. what do you think?

operator/pkg/apis/istio/v1alpha1/values_types.proto
litong01: will remove

operator/pkg/apis/istio/v1alpha1/values_types.pb.go
litong01: this is generated, will be gone once the proto change is removed.

operator/pkg/apis/istio/v1alpha1/values_types.pb.go
litong01: same as above.

operator/pkg/apis/istio/v1alpha1/values_types.pb.go
litong01: same as above

manifests/profiles/default.yaml
litong01: will remove

tests/integration/pilot/analysis/analysis_test.go
litong01: removed

pilot/pkg/features/pilot.go
```diff
@@ -298,6 +298,20 @@ var (
...
+	AnalysisInterval = func() time.Duration {
+		val, _ := env.RegisterDurationVar(
+			"PILOT_ANALYSIS_INTERVAL",
+			10*time.Second,
+			"If analysis is enabled, pilot will run istio analyzers using this value as interval in seconds "+
+				"Istio Resources",
+		).Lookup()
+		if val < 1*time.Second {
+			log.Warnf("PILOT_ANALYSIS_INTERVAL %s is too small, it will be set to default 10 seconds", val.String())
+			return 10 * time.Second
```
howardjohn, line 310: why not set to 1s?

howardjohn (approved):

pilot/pkg/features/pilot.go
litong01: 10s was the default. We could do 1s as well, more of a judgement call. If you would like to use 1s, we could do that.

therealmitchconnors (approved):

PR #37540, opened 2022-02-25
Set upstream max tls version to TLS 1.3

### LINKED ISSUE #36271: possibility to adjust tls version towards backends
Hello,

in our project we use the istio-ingressgateway tied with kind: gateway, virtualservice, serviceentry and destinationrule to route traffic through our istio-ingressgateway to certain backends.

here, we are already able to enforce the usage of  TLS 1.2 and 1.3 on the gateway

```
          maxProtocolVersion: TLSV1_3
          minProtocolVersion: TLSV1_2
          mode: SIMPLE
```

However, we are not able to configure what tls mode to use when connecting from the istio-ingressgateway to one of our backends.
Right now, it seems like istio always uses tls 1.2? Is there a way to enforce tls 1.3? If it is not implemented, it would be a great addition!

If this is somehow possible already, I'd really like to know how. Thanks!

hzxuzhonghu:

Fix #36271

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -1020,7 +1020,7 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 		tlsContext = nil
 	case networking.ClientTLSSettings_ISTIO_MUTUAL:
 		tlsContext = &auth.UpstreamTlsContext{
-			CommonTlsContext: &auth.CommonTlsContext{},
+			CommonTlsContext: defaultUpstreamCommonTLSContext(),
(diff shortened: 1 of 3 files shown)


hzxuzhonghu:
Yes, will do later

lei-tang (approved):
LGTM. The Envoy's default TLS max version for clients is 1.2. Quoted from https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto: 
tls_maximum_protocol_version Maximum TLS protocol version. By default, it’s TLSv1_2 for clients and TLSv1_3 for servers.

ramaraochavali (approved):

jacob-delgado (approved):

PiotrSikora:
Is this configurable by the user or hardcoded without a way to change it?

PR #37549, opened 2022-02-25
[TF] Removing RBAC test utiities

nmittler:
This is removing one of the extra testing frameworks from the security tests. Preferring to use the raw echo libraries to the extent possible.


tests/integration/security/util/headers/builder.go
```diff
@@ -0,0 +1,66 @@
...
+func (b *Builder) WithHost(host string) *Builder {
+	return b.With(Host, host)
+}
+
+// WithXForwardedFor sets the origin IP of the request.
+func (b *Builder) WithXForwardedFor(ip string) *Builder {
+	return b.With(XForwardedFor, ip)
+}
+
+func (b *Builder) Build() http.Header {
```
howardjohn, line 63: can you make `type Builder http.Header` instead of a struct, then avoid the need for Build?

tests/integration/security/util/config/config.go
```diff
@@ -0,0 +1,50 @@
...
+		policy := tmpl.EvaluateAllOrFail(t, args, file.AsStringArrayOrFail(t, filePaths...)...)
+		t.ConfigIstio().ApplyYAMLOrFail(t, ns, policy...)
+	}
+}
+
+func WithWait(filePaths ...string) ApplyFn {
+	return func(t framework.TestContext, ns string, args map[string]string) {
+		t.Helper()
+		policy := tmpl.EvaluateAllOrFail(t, args, file.AsStringArrayOrFail(t, filePaths...)...)
+		// Have to apply/wait one at a time.
```
howardjohn, line 44: can we not apply all then wait for all?

since if you apply multiple quickly pilot will batch them

tests/integration/security/authorization_test.go
```diff
@@ -64,44 +61,46 @@ func TestAuthorization_mTLS(t *testing.T) {
...
 				}
-				policies := tmpl.EvaluateAllOrFail(t, args,
-					file.AsStringOrFail(t, "testdata/authz/v1beta1-mtls.yaml.tmpl"))
-				t.ConfigIstio().ApplyYAMLOrFail(t, apps.Namespace1.Name(), policies...)
-				t.ConfigIstio().WaitForConfigOrFail(t, t, apps.Namespace1.Name(), policies...)
+				config.WithWait("testdata/authz/v1beta1-mtls.yaml.tmpl")(t, apps.Namespace1.Name(), args)
+				callCount := 1
+				if dst.Clusters().IsMulticluster() {
+					// so we can validate all clusters are hit
+					callCount = util.CallsPerCluster * len(dst.Clusters())
```
howardjohn, line 68: nit: this 4 lines is in a lot of places, could just make it abstracted or even automatic

tests/integration/security/util/headers/builder.go
nmittler: By doing that, we lose the ability to call the http.Header methods, like Set/Get.

howardjohn: bummer

howardjohn: How about this:

```go
// Builder for HTTP headers.
type Builder http.Header

// With sets the given header value.
func (b Builder) With(key, value string) Builder {
	http.Header(b).Set(key, value)
	return b
}
func TestT(t *testing.T) {
	t.Log(Builder{}.With("foo", "bar"))
}
```

howardjohn: Oh then its not a http.Header though so you still cannot pass it to things that want http.Header. lame

tests/integration/security/authorization_test.go
```diff
@@ -64,44 +61,46 @@ func TestAuthorization_mTLS(t *testing.T) {
...
+						newTestCase := func(from echo.Instance, to echo.Instances, path string, expectAllowed bool) func(t framework.TestContext) {
+							return func(t framework.TestContext) {
+								opts := echo.CallOptions{
+									Target:   to[0],
+									PortName: "http",
+									Scheme:   scheme.HTTP,
+									Path:     path,
+									Count:    callCount,
+								}
+								opts.Check = check.If(expectAllowed).Then(scheck.Success(false, to, &opts)).
```
howardjohn, line 84: I am not seeing the value of `check.If` over a simple `if` statement here?

tests/integration/security/authorization_test.go
nmittler: Yeah I think that's something we should simply consider within the echo framework somehow. I'm punting on it for now.

nmittler:
@howardjohn PTAL

howardjohn (approved):

ericvn (approved):

JimmyCYJ (approved):

PR #37559, opened 2022-02-26
skip active connections check when downstream cx stats are not available

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/envoy/agent.go
+++ b/pkg/envoy/agent.go
@@ -149,6 +149,11 @@ func (a *Agent) terminate() {
 		ticker := time.NewTicker(activeConnectionCheckDelay)
 		for range ticker.C {
 			ac := a.activeProxyConnections()
+			if ac == -1 {
+				log.Info("Downstream connection stats(downstream_cx_active) are not enabled. skipping active connections check...")
+				a.abortCh <- errAbort
+				return
+			}
 			if ac == 0 {
 				log.Info("There are no more active connections. terminating proxy...")
 				a.abortCh <- errAbort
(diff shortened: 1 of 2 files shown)


pkg/envoy/agent.go
```diff
@@ -149,6 +149,11 @@ func (a *Agent) terminate() {
 		ticker := time.NewTicker(activeConnectionCheckDelay)
 		for range ticker.C {
 			ac := a.activeProxyConnections()
```
howardjohn, line 151: did you check it will still show up when zero? since usedonly was set I wasn't 100% sure

ramaraochavali: Yes.  That is correct. But my thought process was if downstream_cx_active is not there at all in the stats, we can skip active connections. May be we should reword the error message?

pkg/envoy/agent.go
```diff
@@ -149,6 +149,12 @@ func (a *Agent) terminate() {
...
+			if ac == -1 {
+				log.Info("downstream_cx_active are not available. This either means there are no downstream connection established yet" +
+					" or the stats are not enabled. Skipping active connections check...")
```
hzxuzhonghu, line 154: In theory it could be an error too

ramaraochavali: I did not change it to error because no client might have connected to it.

hzxuzhonghu: I mean it could be an error calling 127.0.0.1:15000

ramaraochavali: I see. Handled error differently

hzxuzhonghu (approved):

Issue #37565, opened 2022-02-26
When wasm plugin use OCI image, the `sha256` field should be the wasm binary's checksum instead of the image's digest.


johnlanni:
Example resource:
```yaml
apiVersion: extensions.istio.io/v1alpha1
kind: WasmPlugin
metadata:
  name: openid-connect
  namespace: istio-ingress
spec:
  selector:
    labels:
      istio: ingressgateway
  url: oci://private-registry:5000/openid-connect/openid:latest
  sha256: 1ef0c9a92b0420cf25f7fe5d481b231464bc88f486ca3b9c83ed5cc21d2f6210
```
 
When [fetching entries](https://github.com/istio/istio/blob/6654a4ab8b56c19ae0eb737652fa6a8876731dff/pkg/wasm/cache.go#L116) from the wasm file cache, the url + sha256 specified in WasmPlguin is used.

When [storing entries](https://github.com/istio/istio/blob/6654a4ab8b56c19ae0eb737652fa6a8876731dff/pkg/wasm/cache.go#L185) to the wasm file cache, the url + binary checksum  used.

Therefore, when specifying the `sha256` field as the  OCI image's digest, the cache here will never be hit.

zirain:
IMHO, we should fix the cache issue. 
Maybe the original thought is `share cache between `http` and `oci`, @bianpengyuan can you confirm that?

bianpengyuan:
+1 we should use url + provided sha for the cache.

johnlanni:
Another fix is related to this issue:
https://github.com/istio/istio/pull/37578

zirain:
@bianpengyuan @johnlanni what about change the key to `Namespace/Name` of `WasnPlugin` ?

johnlanni:
@zirain I don't recommend doing this because it needs to support the case when `imagePullPolicy` field is `Always`.

johnlanni:
> @bianpengyuan @johnlanni what about change the key to `Namespace/Name` of `WasnPlugin` ?

Now the logic that `imagePullPolicy` is `Always` has not been implemented, I do not recommend doing this, it is inconvenient for users who want to directly overwrite the image.

bianpengyuan:
[...]

PR #37566, opened 2022-02-26
Use image's digest instead of binary's checksum 

johnlanni:
Fixes: #37565

### DIFF
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -157,7 +157,7 @@ func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration
 		}
 		wasmLog.Debugf("wasm oci fetch %s with options: %v", downloadURL, imgFetcherOps)
 		fetcher := NewImageFetcher(ctx, imgFetcherOps)
-		b, err = fetcher.Fetch(u.Host+u.Path, checksum)
+		b, dChecksum, err = fetcher.Fetch(u.Host+u.Path, checksum)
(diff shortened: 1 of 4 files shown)


johnlanni:
@bianpengyuan  Implemented as discussed [here](https://github.com/istio/istio/issues/37565).

zirain (commented):
LGTM, leave to @bianpengyuan

pkg/wasm/cache.go
```diff
@@ -157,7 +157,7 @@ func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration
 		}
 		wasmLog.Debugf("wasm oci fetch %s with options: %v", downloadURL, imgFetcherOps)
 		fetcher := NewImageFetcher(ctx, imgFetcherOps)
-		b, err = fetcher.Fetch(u.Host+u.Path, checksum)
+		b, dChecksum, err = fetcher.Fetch(u.Host+u.Path, checksum)
```
bianpengyuan, line 160: If this function does not return error, the checksum should be matched right? So we don't really need to return the checksum, just `dChecksum = checksum` would be good?

johnlanni: Need to consider the case where the `sha256` field of the WasmPlugin resource is not set, and the checksum parameter here is an empty string.

bianpengyuan: Ah I see thanks!

bianpengyuan (approved):

PR #37576, opened 2022-02-27
Fix bookinfo k8s ingress example

hanxiaop:
`v1` is introduced in v1.19, and `v1beta1` is no longer served as of 1.22
https://kubernetes.io/docs/reference/using-api/deprecation-guide/


craigbox:
This probably has more impact than it seems.  +Environments leads for comment.

howardjohn (approved):

hanxiaop:
> 

@craigbox The doc may also need to be updated. I'll fix it as well.

Xunzhuo (approved):

PR #37578, opened 2022-02-28
fix a bug that cache entries were never updated

johnlanni:
 
The last touch time of the wasm cache entry was never updated.

### DIFF
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -226,6 +226,7 @@ func (c *LocalFileCache) getEntry(key cacheKey) string {
 	if ce, ok := c.modules[key]; ok {
 		// Update last touched time.
 		ce.last = time.Now()
+		c.modules[key] = ce
 		modulePath = ce.modulePath
 		cacheHit = true
 	}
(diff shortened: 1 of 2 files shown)


pkg/wasm/cache.go
```diff
@@ -226,6 +226,7 @@ func (c *LocalFileCache) getEntry(key cacheKey) string {
 	if ce, ok := c.modules[key]; ok {
 		// Update last touched time.
 		ce.last = time.Now()
+		c.modules[key] = ce
```
zirain, line 229: why need this?

johnlanni: the `ce` is a copy of the element in map

zirain: ce is not a point?

johnlanni: yes. you can see here:https://github.com/istio/istio/pull/37578/files#diff-b4b9abc466fa6e7cad54037be486711eda1582c289c1dcbdd13aebd776442f5cL94

zirain: how about change it to a point, wirting back looks wired.

johnlanni: ok

bianpengyuan (commented):
Thanks! Can we add a unit test for this?

johnlanni:
> Thanks! Can we add a unit test for this?

done

zirain (approved):
LGTM, thx!

PR #31356, opened 2021-03-10
Remove VALIDATION_ENABLED env

hzxuzhonghu:
We should enforce crd validation, otherwise there maybe unknown error.

PR #37579, opened 2022-02-28
Enable validation for mcp-over-xds store.

SpecialYang:
We should not skip validation for CR configs via mcp-over-xds.
We always do this check in previous versions until this pr https://github.com/istio/istio/pull/31356 has been merged.

PTAL @hzxuzhonghu


howardjohn (approved):

hzxuzhonghu:
LGTM

PR #37581, opened 2022-02-28
Istioctl analyze support beta version

hanxiaop:
Take over https://github.com/istio/istio/pull/37398.
/cc @therealmitchconnors 

Proposal: https://docs.google.com/document/d/1C6cpA7V240EPRCDxU19jfqMWWqseaHM-welmmotAVoY/edit?usp=sharing&resourcekey=0-9Il9oisoqX5oh8x9yScBXA

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience


istioctl/cmd/analyze.go
```diff
@@ -316,6 +317,9 @@ func Analyze() *cobra.Command {
 		"Process directory arguments recursively. Useful when you want to analyze related manifests organized within the same directory.")
 	analysisCmd.PersistentFlags().BoolVar(&ignoreUnknown, "ignore-unknown", false,
 		"Don't complain about un-parseable input documents, for cases where analyze should run only on k8s compliant inputs.")
+	analysisCmd.PersistentFlags().BoolVar(&allowAPIVersionConversion, "allowAPIVersionConversion", false,
```
howardjohn, line 320: Why would someone ever *not* want this?

pilot/pkg/config/file/store.go
```diff
@@ -408,6 +412,32 @@ func (s *KubeSource) parseChunk(r *collection.Schemas, name string, lineNum int,
...
+	}
+	sc, found := r.FindByGroupVersionKind(schemaresource.FromKubernetesGVK(chunkSchema))
+	if !found && allowVersionConversion {
+		gvk := &schema.GroupVersionKind{
+			Group: chunkSchema.Group,
+			Kind:  chunkSchema.Kind,
+		}
+		for version := range allowedVersion {
+			gvk.Version = version
+			if sc, found = r.FindByGroupVersionKind(schemaresource.FromKubernetesGVK(gvk)); found {
```
howardjohn, line 433: Looks like we are doing this for ANY type. But its only safe for some types -- really the `istio.io` ones. Maybe in metadata.yaml we can can put some field for version aliases, and for Istio types add them.

For other types, the versions are not inter-changable.

hanxiaop: Yep, I thought there might be configuration changes between packages with different versions. Thanks for the point, I'll look into it.

istioctl/cmd/analyze.go
hanxiaop: I thought it would be issues like the one you mentioned below, and so just let users decide whether to use the feature or not. Since it will be version aliases, the conversion will be safe, and I'll remove the flag.

howardjohn (approved):

Issue #33401, opened 2021-06-11
Unable to route IP with TLS traffic through egress gateway


wjwilson-ibm:
(**NOTE**: This is used to report product bugs:
  To report a security vulnerability, please visit <https://istio.io/about/security-vulnerabilities>
  To ask questions about how to use Istio, please visit <https://discuss.istio.io>)

**Bug description**
Trying to route traffic to an ELK instance in Softlayer using the provided IP over TLS and am unable to route the traffic through the Istio egress gateway.  Request goes directly to the IP/cluster defined by the ServiceEntry.  Was able to reproduce using another well-known IP, steps below

[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure
[ ] Upgrade

**Expected behavior**
Should be able to direct IP based traffic through egress gateway.
Expected cluster to be "outbound|443||istio-egressgateway.istio-system.svc.cluster.local", instead got "outbound|443||disney-placeholder.com", nothing in egress gateway log.

**Steps to reproduce the bug**
See attached egress-with-ip.yaml containing GW+SE+VS
[egress-with-ip.yaml.txt](https://github.com/istio/istio/files/6640561/egress-with-ip.yaml.txt)

**Setup**
1. Install the Sleep sample to your k8s cluster.
2. Open 3 terminal windows.
3. In all 3 terminals, execute:
`$ export SOURCE_POD=$(kubectl get pod -l app=sleep -o jsonpath={.items..metadata.name})`

In terminal 1, watch the egress log.  Execute: 
`$ kubectl logs -l istio=egressgateway -c istio-proxy -n istio-system -f`

In terminal 2, watch the Sleep app sidecar log.  Execute:
`$ kubectl logs ${SOURCE_POD} -c istio-proxy -f`

**Test with IP (130.211.198.204)**
In terminal 3, connect to the IP.  Execute:
```
$ kubectl create -f egress-with-ip.yaml
$ kubectl exec ${SOURCE_POD} -- curl -kLSs https://130.211.198.204
```

In the sidecar log, you will see:
"- - -" 0 - - - "-" 726 65086 111 - "-" "-" "-" "-" "130.211.198.204:443" outbound|443||disney-placeholder.com 172.30.33.92:38236 130.211.198.204:443 172.30.33.92:38234 - -

In the egress log, you will see:
No processing.

[...]

PR #37586, opened 2022-02-28
Add Warning for IPs in SNI Hosts in TLSRoute.Match criteria

GregHanson:

Follow-up to discussion in here https://github.com/istio/istio/issues/33401#issuecomment-971788146, to prevent users from specifying an IP address in the `sniHosts` field of [VirtualService.TLSRoute.Match](https://github.com/istio/api/blob/0511549a0fae34380fdf1399075ccf51bc240be8/networking/v1alpha3/virtual_service.pb.go#L2131) since most clients will not include an IP address in the SNI.  Example of a VirtualService that will no longer pass validation:

```yaml
kind: VirtualService
metadata:
  name: direct-cnn-through-egress-gateway
spec:
  hosts:
  - 1.2.3.4
  tls:
  - match:
    - gateways:
      - mesh
      port: 443
      sniHosts:
      - 1.2.3.4.  ## no longer valid
    route:
    - destination:
        host: istio-egressgateway.istio-system.svc.cluster.local
        port:
          number: 443
```

- [ ] Docs
- [ ] Installation
- [ X ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -2489,7 +2489,7 @@ func validateSniHost(sniHost string, context *networking.VirtualService) (errs V
 	if err := ValidateWildcardDomain(sniHost); err != nil {
 		ipAddr := net.ParseIP(sniHost) // Could also be an IP
 		if ipAddr != nil {
-			errs = appendValidation(errs, WrapWarning(fmt.Errorf("IP address %q is not a valid SNI host value", ipAddr)))
+			errs = appendValidation(errs, WrapWarning(fmt.Errorf("using an IP address (%q) goes against SNI spec and most clients do no support this", ipAddr)))
 			return
 		}
 		return appendValidation(errs, err)
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -2492,9 +2492,10 @@ func validateTLSMatch(match *networking.TLSMatchAttributes, context *networking.
 func validateSniHost(sniHost string, context *networking.VirtualService) error {
 	if err := ValidateWildcardDomain(sniHost); err != nil {
 		ipAddr := net.ParseIP(sniHost) // Could also be an IP
-		if ipAddr == nil {
-			return err
+		if ipAddr != nil {
+			return fmt.Errorf("IP address %q is not a valid SNI host value", ipAddr)
 		}
+		return err
 	}
 	sniHostname := host.Name(sniHost)
 	for _, hostname := range context.Hosts {
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -2443,68 +2443,65 @@ func requestName(match interface{}, matchn int) string {
 	return fmt.Sprintf("#%d", matchn)
 }
 
-func validateTLSRoute(tls *networking.TLSRoute, context *networking.VirtualService) error {
-	var errs error
+func validateTLSRoute(tls *networking.TLSRoute, context *networking.VirtualService) (errs Validation) {
 	if tls == nil {
-		return nil
+		return
 	}
 	if len(tls.Match) == 0 {
-		errs = appendErrors(errs, errors.New("TLS route must have at least one match condition"))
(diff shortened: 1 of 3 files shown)


pkg/config/validation/validation.go
```diff
@@ -2492,9 +2492,10 @@ func validateTLSMatch(match *networking.TLSMatchAttributes, context *networking.
 func validateSniHost(sniHost string, context *networking.VirtualService) error {
 	if err := ValidateWildcardDomain(sniHost); err != nil {
 		ipAddr := net.ParseIP(sniHost) // Could also be an IP
-		if ipAddr == nil {
-			return err
+		if ipAddr != nil {
+			return fmt.Errorf("IP address %q is not a valid SNI host value", ipAddr)
```
howardjohn, line 2496: I think we should make it a warning. I agree it matches the spec but its a breaking change if we do this; users could already have these configs and they will start breaking.

GregHanson: @howardjohn what do you think about the wording for the warning?
```
not all clients support IP address as an SNI value, use with caution
```

howardjohn: Does Envoy even support it? Its against SNI spec, so I think we should be a bit more strongly worded

GregHanson: so currently in Istio, if you specify `sniHosts` with an IP address you do get a listener with the proper SNI value set:
```
0.0.0.0   443  SNI: 1.2.3.4     Cluster: outbound|443|cnn|istio-egressgateway.istio-system.svc.cluster.local
```
So Envoy accepts an IP address in the generated XDS.  Using `curl` as the client still results in a request going through PassthroughCluster
```
[2022-02-28T21:34:00.905Z] "- - -" 0 UF,URX - - "-" 0 0 10015 - "-" "-" "-" "-" "1.2.3.4:443" PassthroughCluster - 1.2.3.4:443 172.30.141.10:59494 - -
```
I tried to configure a Go client example with IP based SNI but it too still goes through `PassthroughCluster`, so I'm not sure what client to test if Envoy functionally supports them

GregHanson: latest message:
```
using an IP address (%q) goes against SNI spec and most clients do no support this
```

pkg/config/validation/validation.go
```diff
@@ -2443,67 +2443,65 @@ func requestName(match interface{}, matchn int) string {
...
 }
 
-func validateSniHost(sniHost string, context *networking.VirtualService) error {
+func validateSniHost(sniHost string, context *networking.VirtualService) (errs Validation) {
 	if err := ValidateWildcardDomain(sniHost); err != nil {
 		ipAddr := net.ParseIP(sniHost) // Could also be an IP
-		if ipAddr == nil {
-			return err
+		if ipAddr != nil {
+			errs = appendValidation(errs, WrapWarning(fmt.Errorf("using an IP address (%q) goes against SNI spec and most clients do no support this", ipAddr)))
```
howardjohn, line 2492:
```suggestion
			errs = appendValidation(errs, WrapWarning(fmt.Errorf("using an IP address (%q) goes against SNI spec and most clients do not support this", ipAddr)))
```

pkg/config/validation/validation_test.go
```diff
@@ -2545,243 +2545,271 @@ func TestValidateVirtualService(t *testing.T) {
 		valid   bool
 		warning bool
 	}{
-		{name: "simple", in: &networking.VirtualService{
+		// {name: "simple", in: &networking.VirtualService{
```
howardjohn, line 2548: uncomment

howardjohn (commented):
lgtm besides the comments

pkg/config/validation/validation.go
GregHanson: fixed

pkg/config/validation/validation_test.go
GregHanson: fixed

howardjohn (approved):

PR #37589, opened 2022-02-28
Cleanup unneeded sleep

howardjohn:
`10167 runs so far, 0 failures (100.00% pass rate). 2.434597227s avg, 3.95269576s max, 660.065395ms min`

### DIFF
--- a/pilot/pkg/xds/ads_test.go
+++ b/pilot/pkg/xds/ads_test.go
@@ -648,7 +648,6 @@ func TestAdsPushScoping(t *testing.T) {
 	for _, c := range svcCases {
 		t.Run(c.desc, func(t *testing.T) {
 			// Let events from previous tests complete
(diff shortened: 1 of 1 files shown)


bianpengyuan (approved):

nmittler (approved):

hzxuzhonghu:
seems the failure related

ramaraochavali:
Yeah. It failed, when I tried to remove it. Could n't spend time on why it is actually needed though

howardjohn:
If 0s sleep fails then 10ms sleep is almost guaranteed to fail as well I
think, CI environment has a lot of hangs etc. Simple things can take >1s

ramaraochavali:
Earlier it was 100ms which wasn't failing. I have increased it further in another PR but let us see if we definitely need 100ms.

PR #37597, opened 2022-02-28
Reduce test duration for google mock CA

ichbinfrog:
Reduce the sleep delay for the mock CA server. The reason for the flakiness is the time it takes until the [first listener accept call](https://github.com/grpc/grpc-go/blob/v1.44.0/server.go#L779) which should not take over 1 second.

Tested on 1000 retries on my local machine.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience

### DIFF
--- a/security/pkg/nodeagent/caclient/providers/google/mock/ca_mock.go
+++ b/security/pkg/nodeagent/caclient/providers/google/mock/ca_mock.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"time"
 
 	"google.golang.org/grpc"
 
@@ -69,9 +68,6 @@ func CreateServer(addr string, service *CAService) (*CAServer, error) {
 		}
 	}()
 
-	// The goroutine starting the server may not be ready, results in flakiness
-	// This extra delay ensures that the future grpc dial does not fail in the worst case scenario
-	time.Sleep(100 * time.Millisecond)
 	if serveErr != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


security/pkg/nodeagent/caclient/providers/google/mock/ca_mock.go
```diff
@@ -62,15 +62,16 @@ func CreateServer(addr string, service *CAService) (*CAServer, error) {
...
 		if err := s.Server.Serve(lis); err != nil {
 			serveErr = err
 		}
 	}()
 
-	// The goroutine starting the server may not be ready, results in flakiness.
-	time.Sleep(1 * time.Second)
+	// The goroutine starting the server may not be ready, results in flakiness
+	// This extra delay ensures that the future grpc dial does not fail in the worst case scenario
+	time.Sleep(100 * time.Millisecond)
```
howardjohn, line 74: Any sleep like this at all is a code smell - do we need it at all? It seems to work even with 0s

howardjohn (approved):

security/pkg/nodeagent/caclient/providers/google/mock/ca_mock.go
ichbinfrog: In my tests 0s of delay works perfectly fine. There could be an edge case where the socket isn't registered yet in the routine leading to the test failing but I don't think it warrants this safeguard. 

Removing it completely

howardjohn (approved):

PR #37604, opened 2022-03-01
Pilot: store latest push instead of global

howardjohn:
Attempt to fix https://github.com/istio/istio/issues/36791

### DIFF
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -200,7 +200,7 @@ func (s *DiscoveryServer) processRequest(req *discovery.DiscoveryRequest, con *C
 
 	// For now, don't let xDS piggyback debug requests start watchers.
 	if strings.HasPrefix(req.TypeUrl, v3.DebugType) {
-		return s.pushXds(con, s.globalPushContext(), &model.WatchedResource{
+		return s.pushXds(con, con.proxy.LastPushContext, &model.WatchedResource{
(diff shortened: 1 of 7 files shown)


howardjohn:
cc @ramaraochavali @hzxuzhonghu WDYT? I cannot come up with any cases where it could break but want some more eyes

pilot/pkg/xds/ads.go
```diff
@@ -650,6 +653,7 @@ func (s *DiscoveryServer) computeProxyState(proxy *model.Proxy, request *model.P
 	if gateway && proxy.Type == model.Router {
 		proxy.SetGatewaysForProxy(push)
 	}
+	proxy.LastPushContext = push
```
hzxuzhonghu, line 656: Can you move this line up?

	if push == nil {
		push = s.globalPushContext()
                proxy.LastPushContext = push
	}

hzxuzhonghu (commented):
LGTM

ramaraochavali:
Where we using globalContext() now after this change? Probably in some tests only? Should we cleanup that also?

howardjohn:
We use it when we initiate Pushes. Let me add comments and clean things up

howardjohn:
> Where we using globalContext() now after this change? Probably in some tests only? Should we cleanup that also?

Cleaned up the remaining incorrect usages (there are a small number of correct usages) and added comments

ramaraochavali (approved):
Thankyou.

Issue #37533, opened 2022-02-25
Support Kubernetes dual-stack capability


jacob-delgado:
**Describe the feature request**

As of Kubernetes 1.23, dual-stack capability is deemed stable and is enabled by default (see https://kubernetes.io/docs/concepts/services-networking/dual-stack/)

This issue will be used a a place-holder for us to link all of the associated work necessary to get dual-stack working in Istio. In the near future we intend to write a checklist of things necessary for us to do before it is deemed experimental.

The RFC for this can be found on the Istio community drive, under the Networking Working Group: [Dual Stack Support](https://docs.google.com/document/d/1oT6pmRhOw7AtsldU0-HbfA0zA26j9LYiBD_eepeErsQ/edit?usp=sharing)

**Affected product area (please put an X in all that apply)**

[X] Docs
[X] Installation
[X] Networking
[X] Performance and Scalability
[X] Extensions and Telemetry
[ ] Security
[X] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[X] Multi Cluster
[X] Virtual Machine
[X] Multi Control Plane

**Additional context**

jacob-delgado:
cc @ycai-aspen @m-eaton

jacob-delgado:
Please link to associated issues/bugs and other information requested from our users.

https://github.com/istio/istio/issues/35781 and https://github.com/istio/istio/issues/35310

jacob-delgado:
The reference implementation done in Istio 1.9

[config_dump.json.zip](https://github.com/istio/istio/files/8138414/config_dump.json.zip)

PR #37606, opened 2022-03-01
Enable metallb LoadBalancer for Kind dual-stack

Kmoneal:
First steps in https://github.com/istio/istio/issues/37533. This will start allowing dual-stack integration tests to run.
Requires https://github.com/istio/common-files/pull/540

### DIFF
--- a/prow/integ-suite-kind.sh
+++ b/prow/integ-suite-kind.sh
@@ -107,8 +107,12 @@ grep 'model' /proc/cpuinfo
 # Default IP family of the cluster is IPv4
 export IP_FAMILY="${IP_FAMILY:-ipv4}"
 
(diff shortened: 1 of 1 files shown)


prow/integ-suite-kind.sh
```diff
@@ -107,8 +107,12 @@ grep 'model' /proc/cpuinfo
 # Default IP family of the cluster is IPv4
 export IP_FAMILY="${IP_FAMILY:-ipv4}"
 
+#Kind will not have a LoadBalancer, so we need to disable it
+export TEST_ENV=kind
```
howardjohn, line 111: seems like a regression, we support metallb for all cases today and you are actually disabling it for ipv6 here

Kmoneal:
Closing because I don't think this is necessary

Issue #37552, opened 2022-02-26
TestManifestGenerateFlags and TestInjection are extremely slow


howardjohn:
Example: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/37550/unit-tests_istio/1497362510515277824

howardjohn:
These two tests take  10minutes and use 100% CPU the entire time

cc @ostromart @litong01

howardjohn:
Part of this is from https://github.com/istio/istio/blob/bc9a9a013025438dacb48d9c475f0a011100470d/manifests/charts/default/templates/mutatingwebhook.yaml#L38

deepCopy of the entire values.yaml tree is super expensive. we can optimzie it

PR #37607, opened 2022-03-01
Optimize injection and operator jobs

howardjohn:
For https://github.com/istio/istio/issues/37552

This may seem like premature optimizations, but these jobs are **incredibly** slow and use 100% cpu the entire time. This, in turn, makes other tests slower/flaky.

In real world, we also see injector using 5% memory total

Changes done here:
* In injector, precompute values map/struct and pre-parse templates.
* In webhook yaml, avoid deep copy of the entire values.yaml tree
* Add ability to pre-filter manifests in operator generation tests. This allows skipping a bunch of resources we don't care about in tests
* Add ability to skip Prune, which is expensive
* In injector tests, store the rendered manifests in files to avoid recomputing each time

This PR should have no user-facing changes

Before:
ok         istio.io/istio/operator/cmd/mesh        163.397s
ok        istio.io/istio/pkg/kube/inject  40.178s

After:
ok         istio.io/istio/operator/cmd/mesh        90.035s
ok        istio.io/istio/pkg/kube/inject  11.369s

Note: these jobs now *temporarily* run on 15 CPU dedicated nodes (up from 5 core shared). When running on even smaller nodes, such as 3 nodes, the before runtime was much higher - I saw 40 **minutes** in my tests


howardjohn:
Todo: precompute manifest for injection. Also can/should use RenderManifestFiltered?
Look into why operator is so slow

litong01:
@howardjohn thanks John. Will look into this one.

litong01 (approved):
This is a big effort, thanks so much for the improvement.

nmittler (approved):

nmittler:
Awesome!

PR #37609, opened 2022-03-01
Skip recompute SidecarScope if push version equals SidecarScope version

hzxuzhonghu:

Fix [TestClusterLocal flakes bug 2](https://github.com/istio/istio/issues/36791#)https://github.com/istio/istio/issues/36791

cc @howardjohn 

- [ x] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -786,7 +786,7 @@ func (node *Proxy) ServiceNode() string {
 func (node *Proxy) SetSidecarScope(ps *PushContext) {
 	// If SidecarScope has been computed for the pushContext, skip recomputing.
 	// https://github.com/istio/istio/issues/36791#issuecomment-1054832155 (Bug 2)
-	if node.SidecarScope != nil && ps.PushVersion == node.SidecarScope.Version {
+	if node.SidecarScope != nil && ps.PushVersion != "" && ps.PushVersion == node.SidecarScope.Version {
 		return
 	}
 	node.PrevSidecarScope = node.SidecarScope
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -784,8 +784,12 @@ func (node *Proxy) ServiceNode() string {
 // Listener generation code will still use the SidecarScope object directly
 // as it needs the set of services for each listener port.
 func (node *Proxy) SetSidecarScope(ps *PushContext) {
-	sidecarScope := node.SidecarScope
-
+	// If SidecarScope has been computed for the pushContext, skip recomputing.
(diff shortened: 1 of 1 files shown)


pilot/pkg/model/context.go
```diff
@@ -784,16 +784,19 @@ func (node *Proxy) ServiceNode() string {
 // Listener generation code will still use the SidecarScope object directly
 // as it needs the set of services for each listener port.
 func (node *Proxy) SetSidecarScope(ps *PushContext) {
-	sidecarScope := node.SidecarScope
-
+	// If SidecarScope has been computed for the pushContext, skip recomputing.
+	// https://github.com/istio/istio/issues/36791#issuecomment-1054832155 (Bug 2)
+	if node.SidecarScope != nil && ps.PushVersion == node.SidecarScope.Version {
```
howardjohn, line 789: a concern I have is that push version can actually be LESS than sidecar scope version in extreme cases.

I think we may want to always send  requests using the last push PushCintext instead of the global latest push context. Or we could only update Previous sidecar scope for push (not request 0

hzxuzhonghu: >a concern I have is that push version can actually be LESS than sidecar scope version in extreme cases.

How could it be?

hzxuzhonghu: >I think we may want to always send requests using the last push PushCintext instead of the global latest push context. Or we could only update Previous sidecar scope for push (not request 0

The intention is opposite from the beginning, istio always generates xds from the latest pushContext.

howardjohn: 1. Dequeue push for version 0. Not yet hit https://github.com/istio/istio/blob/e3904269b7bd95207fdccd2be6f14276a3ad470a/pilot/pkg/xds/ads.go#L330
2. New update, global push set to version 1
3. New update, global push set to version 2
4. Receive push request, compute SidecarScope for version 2
5. The original push for v0 runs, now current is v0, previous is v2

> The intention is opposite from the beginning, istio always generates xds from the latest pushContext.
It is still the "latest", its just the per-proxy latest rather than the global "latest"

hzxuzhonghu: So, could we always use the global latest pushContext when `SetSidecarScope`

howardjohn: > So, could we always use the global latest pushContext when SetSidecarScope

Then the request.PushContext would not match the sidecarscope though

Issue #37529, opened 2022-02-24
Assertion failure in RDS cache


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/integ-assertion_istio_postsubmit/1496959423606689792

```
(*discoveryv3.Resource)(Inverse(protocmp.Transform, protocmp.Message{
  	"@type": s"envoy.service.discovery.v3.Resource",
  	"name":  string("8888"),
  	"resource": protocmp.Message{
  		"@type":    s"google.protobuf.Any",
  		"type_url": string("type.googleapis.com/envoy.config.route.v3.RouteConfiguration"),
  		"value": protocmp.Message{
  			"@type":             s"envoy.config.route.v3.RouteConfiguration",
  			"name":              string("8888"),
  			"validate_clusters": protocmp.Message{"@type": s"google.protobuf.BoolValue"},
  			"virtual_hosts": []protocmp.Message{
  				{"@type": s"envoy.config.route.v3.VirtualHost", "domains": []string{"fake.external.com", "fake.external.com:8888"}, "include_request_attempt_count": bool(true), "name": string("fake.external.com:8888"), ...},
  				{
  					... // 2 identical entries
  					"include_request_attempt_count": bool(true),
  					"name":                          string("allow_any"),
+ 					"request_headers_to_add": []protocmp.Message{
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
+ 						s`{header:{key:"x-vhost-outbound", value:"hello world"}}`,
[...]

PR #37612, opened 2022-03-01
Clone catch all vhost to prevent it from being mutated

hzxuzhonghu:

Fix https://github.com/istio/istio/issues/37529

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/httproute.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute.go
@@ -22,6 +22,7 @@ import (
 
 	route "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
 	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	protobuf "google.golang.org/protobuf/proto"
 
 	networking "istio.io/api/networking/v1alpha3"
 	"istio.io/istio/pilot/pkg/features"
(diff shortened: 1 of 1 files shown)


ramaraochavali (approved):

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -190,7 +191,9 @@ func (configgen *ConfigGeneratorImpl) buildSidecarOutboundHTTPRouteConfig(
 	util.SortVirtualHosts(virtualHosts)
 
 	if !useSniffing {
-		virtualHosts = append(virtualHosts, node.CatchAllVirtualHost)
+		// virtualhost envoyfilter can mutate this sharing config.
+		catchAll := protobuf.Clone(node.CatchAllVirtualHost).(*route.VirtualHost)
```
howardjohn, line 195: the whole point of `node.CatchAllVirtualHost` was to cache it. So if we are not caching it, why not just revert https://github.com/istio/istio/commit/b32710c73525c0ba0df8af1d4903d947da68e178?

My belief (untested!) is that `proto.Clone` is slower than rebuilding from scratch

hzxuzhonghu: Good question, I feel clone may be quicker than rebuild. let me test it

hzxuzhonghu: You are totally right. clone is more slower

BenchmarkClone-8                	  549873	      2212 ns/op
BenchmarkBuild-8                	 1844498	       643.6 ns/op



```
+
+
+func BenchmarkClone(b *testing.B) {
+       vhost := istionetworking.BuildCatchAllVirtualHost(true, "")
+       for i := 0; i < b.N; i++ {
+               _ = protobuf.Clone(vhost).(*route.VirtualHost)
+       }
+}
+
+func BenchmarkBuild(b *testing.B) {
+       for i := 0; i < b.N; i++ {
+               _ = istionetworking.BuildCatchAllVirtualHost(true, "")
+       }
+}

```

PR #37616, opened 2022-03-01
optimize: auto allocate addresses only when services updated

hzxuzhonghu:

Deepcopy is still necessary.

### DIFF
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
@@ -518,9 +518,10 @@ func (s *ServiceEntryStore) Services() []*model.Service {
 	}
 	s.mutex.RUnlock()
 	for _, svc := range allServices {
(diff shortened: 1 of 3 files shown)


pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -510,17 +510,19 @@ func (s *ServiceEntryStore) HasSynced() bool {
 // Services list declarations of all services in the system
 func (s *ServiceEntryStore) Services() []*model.Service {
 	s.mutex.RLock()
-	allServices := s.services.getAllServices()
-	s.mutex.RUnlock()
+	defer s.mutex.RUnlock()
```
ramaraochavali, line 513: can we move this before deepcopy loop?

ramaraochavali: yes

ramaraochavali (approved):

PR #37624, opened 2022-03-01
Fix: mcs incorrectly deleted

### LINKED ISSUE #36906: MCS serviceimport uncorrectly register service handler
### Bug Description

Currently, each cluster has one service importCache

And each one registers service handler to the aggregated mesh wide aggregator controller
it means when one service event will trigger `n` handler running. This is intended, but looking at serviceImportCacheImpl.onServiceEvent.

If a delete event is received, it will remove the generated mcs service from local cluster, this is wrong because the service can still exist in some other clusters.

```
	if event == model.EventDelete || len(vips) == 0 {
		if prevMcsService != nil {
			// There are no vips in this cluster. Just delete the MCS service now.
			ic.deleteService(prevMcsService)
		}
		return
	}

```

### Version

```prose
1.12.1
```

### Additional Information

_No response_

### Affected product area

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Upgrade
- [ ] Multi Cluster
- [ ] Virtual Machine
- [ ] Control Plane Revisions

dddddai:
Delete mcs only when all "real" services are deleted

Fixes #36906

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
+++ b/pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
@@ -110,8 +110,11 @@ func (ic *serviceImportCacheImpl) onServiceEvent(svc *model.Service, event model
 	// Get the ClusterSet VIPs for this service in this cluster. Will only be populated if the
 	// service has a ServiceImport in this cluster.
 	vips := ic.getClusterSetIPs(namespacedName)
+	name := namespacedName.Name
+	ns := namespacedName.Namespace
 
-	if event == model.EventDelete || len(vips) == 0 {
+	if len(vips) == 0 || (event == model.EventDelete &&
+		ic.opts.MeshServiceController.GetService(kube.ServiceHostname(name, ns, ic.opts.DomainSuffix)) == nil) {
 		if prevMcsService != nil {
 			// There are no vips in this cluster. Just delete the MCS service now.
 			ic.deleteService(prevMcsService)
--- a/pilot/pkg/serviceregistry/kube/controller/serviceimportcache_test.go
+++ b/pilot/pkg/serviceregistry/kube/controller/serviceimportcache_test.go
@@ -137,15 +137,30 @@ func TestDeleteImportedService(t *testing.T) {
 	for _, mode := range []EndpointMode{EndpointsOnly, EndpointSliceOnly} {
 		t.Run(mode.String(), func(t *testing.T) {
 			// Create and run the controller.
-			c, ic, cleanup := newTestServiceImportCache(mode)
+			c1, ic, cleanup := newTestServiceImportCache(mode)
 			defer cleanup()
 
-			ic.createKubeService(t, c)
+			// Create and run another controller.
+			c2, _ := NewFakeControllerWithOptions(FakeControllerOptions{
+				Stop:      c1.stop,
+				ClusterID: "test-cluster2",
+				Mode:      mode,
+			})
+			go c2.Run(c2.stop)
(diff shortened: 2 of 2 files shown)


pilot/pkg/serviceregistry/kube/controller/serviceimportcache_test.go
```diff
@@ -310,9 +325,18 @@ func (ic *serviceImportCacheImpl) updateKubeService(t *testing.T) {
...
 
-func (ic *serviceImportCacheImpl) deleteKubeService(t *testing.T) {
+func (ic *serviceImportCacheImpl) deleteKubeService(t *testing.T, anotherCluster *FakeController) {
 	t.Helper()
 
+	if err := anotherCluster.client.CoreV1().Services(serviceImportNamespace).Delete(context.TODO(), serviceImportName, kubeMeta.DeleteOptions{}); err != nil {
+		t.Fatal(err)
+	}
+	// Wait for the resources to be processed by the controller.
+	time.Sleep(1 * time.Second)
```
howardjohn, line 335: Sleeping during tests like this is not a good practice https://github.com/istio/istio/wiki/Test-Flakes#timing-dependencies

pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
```diff
@@ -111,7 +111,7 @@ func (ic *serviceImportCacheImpl) onServiceEvent(svc *model.Service, event model
 	// service has a ServiceImport in this cluster.
 	vips := ic.getClusterSetIPs(namespacedName)
 
-	if event == model.EventDelete || len(vips) == 0 {
+	if len(vips) == 0 || (event == model.EventDelete && ic.opts.MeshServiceController.GetService(kube.ServiceHostname(namespacedName.Name, namespacedName.Namespace, ic.opts.DomainSuffix)) == nil) {
```
howardjohn, line 114: do we event need to look for `model.EventDelete`? seems like we can/should just reconcile to the current state

dddddai: > seems like we can/should just reconcile to the current state

Yes but I think the idea is to clean up mcs
We [create mcs only if the real services exist](https://github.com/istio/istio/blob/1.13.1/pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go#L164-L172), so it makes sense to delete it when all real services are gone

pilot/pkg/serviceregistry/kube/controller/serviceimportcache.go
```diff
@@ -110,8 +110,11 @@ func (ic *serviceImportCacheImpl) onServiceEvent(svc *model.Service, event model
 	// Get the ClusterSet VIPs for this service in this cluster. Will only be populated if the
 	// service has a ServiceImport in this cluster.
 	vips := ic.getClusterSetIPs(namespacedName)
+	name := namespacedName.Name
+	ns := namespacedName.Namespace
 
-	if event == model.EventDelete || len(vips) == 0 {
+	if len(vips) == 0 || (event == model.EventDelete &&
+		ic.opts.MeshServiceController.GetService(kube.ServiceHostname(name, ns, ic.opts.DomainSuffix)) == nil) {
```
hzxuzhonghu, line 117: LGTM, though it may not efficient, but we have no better choice currently

hzxuzhonghu (approved):

Issue #37555, opened 2022-02-26
Slow unit tests


howardjohn:
The following tests take >1s:

```
TestReleaseOnCancellation_ConfigMaps 1.00s
TestReleaseOnCancellation_Endpoints 1.00s
TestReleaseOnCancellation_Leases 1.00s
TestTokenExchangePluginWithCache 1.00s
Test_tokenSupplier_GetRequestMetadata 1.00s
TestRunComponentsAfterStart/RunComponentAsyncAndWait 1.01s
TestSleepCheckInstall 1.01s
TestStsCache 1.01s
TestGetFederatedToken 1.02s
TestRunComponentsAfterStart 1.02s
TestGetCNIConfigFilepath 1.03s
TestSendPushesManyPushes 1.03s
TestClient 1.06s
TestStsFlow 1.06s
TestTrailingWhitespace 1.06s
TestLeaderElectionNoPermission 1.08s
TestSDS 1.08s
TestWorkloadAgentGenerateSecret/no_plugin 1.09s
TestAdsPushScoping/Add_an_unmatched_Sidecar_config 1.10s
TestAdsPushScoping/Add_a_unscopedname_not_match_service 1.10s
TestAdsPushScoping/Add_a_unscopedns_not_match_service 1.10s
TestAdsPushScoping/Add_destination_rule_to_an_unscoped_service 1.10s
TestAdsPushScoping/Add_instances_to_an_unscoped_service 1.10s
TestAdsPushScoping/Add_virtual_service_to_an_unscoped_service 1.10s
TestAdsPushScoping/Delete_destination_rule_of_a_unscoped_service 1.10s
TestAdsPushScoping/Delete_virtual_service_of_a_unscoped_service 1.10s
TestAdsPushScoping/Remove_a_unscopedname_not_match_service 1.10s
TestAdsPushScoping/Remove_a_unscopedns_not_match_service 1.10s
TestTokenExchangePlugin 1.15s
TestSidecarTemplate/sidecar_template 1.18s
TestStartEnvoyShouldSucceed 1.22s
TestInstallPackagePath/install_package_path#02 1.23s
TestJwtPubKeyRefreshWithNetworkError 1.25s
TestAgent/Token_exchange_with_credential_fetcher_downtime 1.32s
TestGetProxyServiceInstancesWithMultiIPsAndTargetPorts 1.32s
TestInstallPackagePath/install_package_path 1.36s
TestWorkloadAgentGenerateSecret 1.38s
[...]

PR #37640, opened 2022-03-01
Speed up TestK8sSignWithMeshConfig

howardjohn:
for https://github.com/istio/istio/issues/37555

Before this test took 25s. Now its ~instant. This was because we used Watches incorrectly, so it always hit 5s timeout


lei-tang (approved):
Thanks for the speed up!

PR #37647, opened 2022-03-02
tf: cleanup usages of Workload

howardjohn:
This introduces a new TargetWorkload entry to call *pods* rather than
*services*. And cleans up a bunch of usages where we use to do things
the hard way

### DIFF
--- a/pkg/test/framework/components/echo/calloptions.go
+++ b/pkg/test/framework/components/echo/calloptions.go
@@ -97,8 +97,14 @@ type Target interface {
 
 // CallOptions defines options for calling a Endpoint.
 type CallOptions struct {
-	// To is the Target to be called. Required.
+	// To is the Target to be called.
 	To Target
+
+	// ToWorkload will call a specific workload in this instance, rather than the Service.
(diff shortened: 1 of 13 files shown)


pkg/test/framework/components/echo/common/call.go
```diff
@@ -182,8 +182,15 @@ func ForwardEcho(srcName string, clientProvider EchoClientProvider, opts *echo.C
 }
 
 func fillInCallOptions(opts *echo.CallOptions) error {
-	if opts.Target != nil {
-		targetPorts := opts.Target.Config().Ports
+	if opts.Target != nil && opts.TargetWorkload != nil {
```
nmittler, line 185: There's a lot of branching here. Can we create 2 functions that handle either Target or TargetWorkload?

douglas-reid (approved):
Approving the telemetry test file change.

pkg/test/framework/components/echo/common/call.go
howardjohn: A bit worried about duplicating all the code which his pretty complex

lei-tang (approved):

PR #37648, opened 2022-03-02
Revert cache catch all virtualhost

hzxuzhonghu:

Because i found rebuild a new catch-all vhost is far more quickly than clone.


ramaraochavali:
Just put a hold to understand this better - is this caused by Rds cache or is this triggered by an envoy filter trying to change catchAllVirtualHost?

howardjohn:
it's from envoy filter. The cache just allows us to catch the bug

hzxuzhonghu:
>is this caused by Rds cache or is this triggered by an envoy filter trying to change catchAllVirtualHost?

envoyfilter changed the catchall vhost

ramaraochavali:
Ok. Thank you. That case was discussed and we felt catchAllVirtualHost is an implementation detail of Istio on how we handle unmatched traffic and moved forward with it.

Issue #37622, opened 2022-03-01
Support dual-stack capability for Istio resources


zhlsunshine:
**Describe the feature request**
As described in [dual-stack support](https://github.com/istio/istio/issues/37533), this issue is created to track the PR which focus on support dual-stack capability for Istio resources. And the Istio resources, such as cluster, endpoint, listener, httproute, gateway, will contains all necessary configuration to support dual-stack.
Please refer to [Dual Stack Support](https://docs.google.com/document/d/1oT6pmRhOw7AtsldU0-HbfA0zA26j9LYiBD_eepeErsQ/edit?usp=sharing) for detail information.

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[x] Docs
[x] Installation
[x] Networking
[x] Performance and Scalability
[x] Extensions and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[x] Multi Cluster
[x] Virtual Machine
[x] Multi Control Plane

**Additional context**

abhishekpatil9797:
Is dual stack support in Istio available in some Istio's version ?

zhlsunshine:
Hi @abhishekpatil9797,  there is a dual stack implementation for Istio in experimental branch [here](https://github.com/istio/istio/tree/experimental-dual-stack), and there is also an [issue](https://github.com/istio/istio/issues/40394) to track the dual stack solution in master branch.

devenderkamble2101:
Kubernetes cluster is dual stack, Internal calico network is on dual stack, Does Istio proxy tolerate this dual stack configuration and can work with only IPv4 or IPv46 only configuration

zhlsunshine:
Hi @devenderkamble2101 Yes, Istio proxy can support the only IPv4 or only IPv6 separately, however, dual stack is different with handle them separately. There already is an issue to trace this here: https://github.com/istio/istio/issues/40394

zhlsunshine:
Hi @abhishekpatil9797 this current master branch can support the dual stack

howardjohn:
Looking at the configs:

Gateway server bind (rare). I think you can just make multiple
Sidecar egress server bind. Same thing
Sidecar ingress server bind. Same thing
WorkloadEntry address. This is basically the one case where it would make sense to have DS. Since today you would need to create 2 WEs

So I still think this is valid maybe for WE.

PR #37652, opened 2022-03-02
add feature flag to support dual-stack feature

zhlsunshine:
Add feature flags to support dual-stack, including 2 flags, the one is for istiod, the other is for pilot-agent.
Sub task for [issue#37622](https://github.com/istio/istio/issues/37622)

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -232,6 +232,12 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
 	// used for both initKubeRegistry and initClusterRegistries
 	args.RegistryOptions.KubeOptions.EndpointMode = kubecontroller.DetectEndpointMode(s.kubeClient)
 
+	enableEndpointSlice, _ := features.EnableEndpointSliceController()
+	if features.EnableDualStack && !enableEndpointSlice {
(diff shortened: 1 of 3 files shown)


pilot/pkg/bootstrap/server.go
```diff
@@ -232,6 +232,12 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
 	// used for both initKubeRegistry and initClusterRegistries
 	args.RegistryOptions.KubeOptions.EndpointMode = kubecontroller.DetectEndpointMode(s.kubeClient)
 
+	enableEndpointSlice, _ := features.EnableEndpointSliceController()
```
ycai-aspen, line 235: Do you still need this when EndpointSlice is used by default?

jacob-delgado: So EndpointSlices are only turned on for K8s 1.21+. This acts as a way to also help us feature-gate dual stack from being run on earlier versions, although we should have a check similar to what was done for https://github.com/istio/istio/pull/34971 instead

zhlsunshine: Yeah, I think we need to consider a situation that there is a custom set to false, then the mistake can be caught by following logic.

jacob-delgado (approved):

PR #33108, opened 2021-05-26
Remove default caBundle value from Chart

etifontaine:
Fix #33052 

When using a GitOps approach where all generated files from `istioctl manifest generate -f .. -rev ..` are pushed to Git, keeping an empty string will trigger some infinite loop between Flux (for example) and the webhook in charge of setting the caBundle value.

By removing the key from the manifest, Flux (or any CD tool), won't "watch" this key and it will be set in the cluster.

[ ] Configuration Infrastructure
[ ] Docs
[X] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #37655, opened 2022-03-02
Remove default caBundle value from Chart to allow GitOps

etifontaine:
This is to fix a breaking change that has been already been patched with https://github.com/istio/istio/pull/33108
This sad to have to open it again to make the same changes..

When using a GitOps approach where all generated files from istioctl manifest generate -f .. -rev .. are pushed to Git, keeping an empty string will trigger some infinite loop between Flux (for example) and the webhook in charge of setting the caBundle value.

By removing the key from the manifest, Flux (or any CD tool), won't "watch" this key and it will be set in the cluster.

[ ] Docs
[X] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/manifests/charts/default/templates/mutatingwebhook.yaml
+++ b/manifests/charts/default/templates/mutatingwebhook.yaml
@@ -12,7 +12,6 @@
       namespace: {{ .Release.Namespace }}
       path: "/inject"
     {{- end }}
-    caBundle: ""
   sideEffects: None
   rules:
     - operations: [ "CREATE" ]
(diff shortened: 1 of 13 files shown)


etifontaine:
As it has been already discussed, it would be very nice to have it on the 1.12 and 1.13

howardjohn (commented):
Everything you changed here is just test data; the real charts are under manifests/ which is where we need the change

howardjohn:
But LGTM once we update those references

etifontaine:
@howardjohn Thank you, I have removed all references to `caBundle: ""`

howardjohn (approved):

etifontaine:
@howardjohn can you add the label to make it available on the next 1.12 and 1.13?

eliavem (approved):

PR #37669, opened 2022-03-02
Update --no-clobber to not check existence of tag `latest`

ericvn:

Job build-base-images_release-builder_periodic of type periodic ended with state failure fails with:
```
Error: image "docker.io/istio/app_sidecar_base_centos_7:latest" already exists
```

This was caused by a recent change to create :latest tags of the images (#https://github.com/istio/istio/pull/37417)


howardjohn (commented):
can we make no-clobber just allow `:latest` maybe? otherwise we lose some safety

ericvn:
> can we make no-clobber just allow `:latest` maybe? otherwise we lose some safety

Updated.

howardjohn (approved):

PR #37670, opened 2022-03-02
gateway: avoid recreating Deployment on each kubectl apply

howardjohn:
Currently, kubectl apply annotation from Gateway is propogated to
Deployment which triggers a restart everytime we change any field.

### DIFF
--- a/pilot/pkg/config/kube/gateway/templates.go
+++ b/pilot/pkg/config/kube/gateway/templates.go
@@ -52,13 +52,6 @@ func funcMap() template.FuncMap {
 		}
 		return strings.TrimSuffix(string(data), "\n")
 	}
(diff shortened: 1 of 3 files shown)


ruigulala (approved):

pilot/pkg/config/kube/gateway/templates.go
```diff
@@ -52,6 +52,28 @@ func funcMap() template.FuncMap {
 		}
 		return strings.TrimSuffix(string(data), "\n")
 	}
+	f["toYaml"] = func(mps map[string]interface{}) string {
```
bianpengyuan, line 55: Not used?

bianpengyuan (approved):

Issue #37433, opened 2022-02-18
ServiceEntry with Cross cluster VM workload


qudongfang:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**

A single network, Multi-cluster, Multi-control-plane connected with remote clusters.

For example:
Cluster `A` and Cluster `B`, they're remote clusters to each other.

VM workloads are registered to `A` cluster, `ServiceEntry` host `potato.vm` is working within the `A` cluster.
But `patato.vm` is not available in `B` cluster.

**Describe alternatives you've considered**

It works if replacing`ServiceEntry` with `Service`

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[X ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[X ] Multi Cluster
[X ] Virtual Machine
[X ] Multi Control Plane

**Additional context**

release: 1.11.6

hzxuzhonghu:
WorkloadEntryCrossCluster = env.RegisterBoolVar("PILOT_ENABLE_CROSS_CLUSTER_WORKLOAD_ENTRY", true,
		"If enabled, pilot will read WorkloadEntry from other clusters, selectable by Services in that cluster.").Get()


I thought you want this. It is there.

qudongfang:
> ```
> WorkloadEntryCrossCluster = env.RegisterBoolVar("PILOT_ENABLE_CROSS_CLUSTER_WORKLOAD_ENTRY", true,
> 	"If enabled, pilot will read WorkloadEntry from other clusters, selectable by Services in that cluster.").Get()
> ```
> 
> I thought you want this. It is there.

@hzxuzhonghu Thanks for your response.
You're right that WorkloardEntries are available across clusters with `Service`, but It's not working with `ServiceEntry` which is what this feature request is for.

qudongfang:
Any suggestions would be greatly appreciated! @stevenctl 

https://github.com/istio/istio/issues/29026

hzxuzhonghu:
Isn't SE selecting workload entries across clusters done by https://github.com/istio/istio/issues/29026?

qudongfang:
> Isn't SE selecting workload entries across clusters done by #29026?

Like I said, It's working with `Service`, not `ServiceEntry`.

[...]

PR #37675, opened 2022-03-03
Enable serviceentry wles cross clusters

hzxuzhonghu:

#37433


hzxuzhonghu:
assign @stevenctl

stevenctl (approved):
Should note that for primary-remote setups this won't work. Right now we have

`local cluster -> workload entry from remote clusters`, but not `all clusters -> all clusters`.

PR #37678, opened 2022-03-03
build default transport socketmatch each time instead of using shared

hzxuzhonghu:

To prevent cluster patch mutating the sharing default transport socket match

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -45,16 +45,6 @@ import (
 	"istio.io/istio/pkg/util/gogo"
 )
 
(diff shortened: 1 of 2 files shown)


ramaraochavali (approved):

howardjohn:
Wish we use a way to do copy on write. So we can do efficient code like this but if it's patched by envoy filter (which means advanced user) they pay a small penalty to copy it. Not sure how to do that

hzxuzhonghu:
I did think about this, original thought is: check the envoyfilter to decide whether copy or not.

ramaraochavali:
@hzxuzhonghu thinking about this, what would people change in the default socket match via envoy filter?

hzxuzhonghu:
Not intended, could change it by mistake. Say a patch could be applied to all without specifying match conditions.

Issue #37592, opened 2022-02-28
TestFileSecrets test flake


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/unit-tests_istio_postsubmit/1498362808511762432

howardjohn:
@ramaraochavali this is a real bug, can you take a look? The problem is we have two different methods of handling add and remove. In tests we do "add; Remove; add". Its not always processed in order

ramaraochavali:
Sure

ramaraochavali:
@howardjohn TBH, I did not fully understand the problem but thought you were referring to what I did in https://github.com/istio/istio/pull/37656 . Can u please explain more, if that is not the case?

ramaraochavali:
@howardjohn Can you please expand more on the issue ^^ - was my understanding correct that it is a test bug?

howardjohn:
I think it's a real bug. the problem is we may do an add+remove+add of a file. This is perfectly valid and has a defined ordering (ie we did these actions in sequence in single thread).The result should be we are watching the cert

But then in the agent we are handling adds and removes in seperately in different threads. So it turns out we may process them in order "add, add, remove". This ends up NOT watching the cert

ramaraochavali:
I see. Now I get it. Let me try to fix it.

PR #37679, opened 2022-03-03
file watcher fixes

ramaraochavali:
Fixes https://github.com/istio/istio/issues/37592

This fixes couple of things
- It does not check for already watching any more because we do not need to check for duplicate events. We have debounce implemented now.
- It does not handle REMOVE event. Remove event has a bug
     - We are triggering update callback for Remove which actually will fail because file is not there
     

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [X] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali:
@howardjohn PTAL

ramaraochavali:
Interesting in CI, atomicwrite generates CHMOD followed by REMOVE events. So earlier the REMOVE event logic was helping. https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/37679/unit-tests_istio/1499324591892336640/build-log.txt

Now changed to handle CHMOD for rename cases.

ramaraochavali:
@howardjohn PTAL when you get chance. Definitely not urgent. Just want to get this on your list.

howardjohn (commented):
I am still seeing TestFileSecrets failures when running locally

ramaraochavali:
ok. Let me take a look :-).

ramaraochavali:
With the recent fix I made to validate cert before delivering to Envoy, the REMOVE part is handled. So I am closing this PR.

PR #37681, opened 2022-03-03
Handle PrivateKeyProvider configuration

rveerama1:

When the PrivateKeyProvider configuration information is provided through Istio operator yaml file, parse and pass it to gateway or sidecars. Envoy will act based on the information provided by the configuration.

To set the mesh wide defaults, configure the `defaultConfig` section of `meshConfig`.
For example:

```
meshConfig:
  defaultConfig:
    privateKeyProvider:
      cryptomb:
        pollDelay: 0.01s
```

This can also be configured on a per-workload basis by configuring the `proxy.istio.io/config`
annotation on the pod.
For example:

```
annotations:
  proxy.istio.io/config: |
    privateKeyProvider:
      cryptomb:
        pollDelay: 0.01s
```

This requires https://github.com/istio/api/pull/2261

### DIFF
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -40,6 +40,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 
+	"istio.io/api/mesh/v1alpha1"
 	"istio.io/api/security/v1beta1"
 	kubecredentials "istio.io/istio/pilot/pkg/credentials/kube"
 	"istio.io/istio/pilot/pkg/features"
@@ -512,6 +513,14 @@ func (s *Server) WaitUntilCompletion() {
 	s.server.Wait()
 }
 
+func (s *Server) fetchPrivateKeyProviderConfig() *v1alpha1.PrivateKeyProvider {
+	meshConf := s.environment.Mesh()
+	proxyConf := meshConf.GetDefaultConfig()
+
+	pkpConf := proxyConf.GetPrivateKeyProvider()
+	return pkpConf
+}
+
 // initSDSServer starts the SDS server
 func (s *Server) initSDSServer() {
 	if s.kubeClient == nil {
@@ -535,7 +544,8 @@ func (s *Server) initSDSServer() {
 				Reason: []model.TriggerReason{model.SecretTrigger},
 			})
 		})
-		s.XDSServer.Generators[v3.SecretType] = xds.NewSecretGen(creds, s.XDSServer.Cache, s.clusterID)
+		pkpConf := s.fetchPrivateKeyProviderConfig()
+		s.XDSServer.Generators[v3.SecretType] = xds.NewSecretGen(creds, s.XDSServer.Cache, s.clusterID, pkpConf)
 		s.multiclusterController.AddHandler(creds)
 	}
 }
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -535,7 +535,7 @@ func (s *Server) initSDSServer() {
 				Reason: []model.TriggerReason{model.SecretTrigger},
 			})
 		})
-		s.XDSServer.Generators[v3.SecretType] = xds.NewSecretGen(creds, s.XDSServer.Cache, s.clusterID, s.environment.IstioConfigStore, s.environment.Mesh())
+		s.XDSServer.Generators[v3.SecretType] = xds.NewSecretGen(creds, s.XDSServer.Cache, s.clusterID, s.environment.Mesh())
 		s.multiclusterController.AddHandler(creds)
 	}
 }
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -40,7 +40,6 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 
-	"istio.io/api/mesh/v1alpha1"
 	"istio.io/api/security/v1beta1"
 	kubecredentials "istio.io/istio/pilot/pkg/credentials/kube"
 	"istio.io/istio/pilot/pkg/features"
@@ -513,14 +512,6 @@ func (s *Server) WaitUntilCompletion() {
 	s.server.Wait()
 }
 
-func (s *Server) fetchPrivateKeyProviderConfig() *v1alpha1.PrivateKeyProvider {
-	meshConf := s.environment.Mesh()
-	proxyConf := meshConf.GetDefaultConfig()
-
-	pkpConf := proxyConf.GetPrivateKeyProvider()
-	return pkpConf
-}
-
 // initSDSServer starts the SDS server
 func (s *Server) initSDSServer() {
 	if s.kubeClient == nil {
@@ -544,8 +535,7 @@ func (s *Server) initSDSServer() {
 				Reason: []model.TriggerReason{model.SecretTrigger},
 			})
 		})
-		pkpConf := s.fetchPrivateKeyProviderConfig()
-		s.XDSServer.Generators[v3.SecretType] = xds.NewSecretGen(creds, s.XDSServer.Cache, s.clusterID, pkpConf)
+		s.XDSServer.Generators[v3.SecretType] = xds.NewSecretGen(creds, s.XDSServer.Cache, s.clusterID, s.environment.IstioConfigStore, s.environment.Mesh())
 		s.multiclusterController.AddHandler(creds)
 	}
 }
--- a/pilot/pkg/xds/sds.go
+++ b/pilot/pkg/xds/sds.go
@@ -310,7 +310,8 @@ func toEnvoyCaSecret(name string, cert []byte) *discovery.Resource {
 
 func toEnvoyKeyCertSecretWithPrivateKeyProvider(name string, key, cert []byte, pkpConf *mesh.PrivateKeyProvider) *anypb.Any {
 	var res *anypb.Any
-	if pkpConf.Name == "cryptomb" {
+	switch pkpConf.GetProvider().(type) {
+	case *mesh.PrivateKeyProvider_Cryptomb:
 		crypto := pkpConf.GetCryptomb()
 		msg := util.MessageToAny(&cryptomb.CryptoMbPrivateKeyMethodConfig{
 			PollDelay: durationpb.New(time.Duration(crypto.GetPollDelay().Nanos)),
@@ -330,7 +331,7 @@ func toEnvoyKeyCertSecretWithPrivateKeyProvider(name string, key, cert []byte, p
 						},
 					},
 					PrivateKeyProvider: &envoytls.PrivateKeyProvider{
-						ProviderName: pkpConf.Name,
+						ProviderName: "cryptomb",
 						ConfigType: &envoytls.PrivateKeyProvider_TypedConfig{
 							TypedConfig: msg,
 						},
@@ -338,8 +339,8 @@ func toEnvoyKeyCertSecretWithPrivateKeyProvider(name string, key, cert []byte, p
 				},
 			},
 		})
-	} else {
-		log.Warnf("Unknown PrivateKeyProvider: %s", pkpConf.Name)
+	default:
+		log.Warnf("Unknown PrivateKeyProvider")
 		return nil
 	}
 
--- a/pilot/pkg/xds/sds.go
+++ b/pilot/pkg/xds/sds.go
@@ -140,7 +140,7 @@ func (s *SecretGen) Generate(proxy *model.Proxy, w *model.WatchedResource, req *
 			continue
 		}
 		regenerated++
-		res := s.generate(sr, configClusterSecrets, proxyClusterSecrets, proxy)
+		res := s.generate(sr, configClusterSecrets, proxyClusterSecrets)
 		if res != nil {
 			s.cache.Add(sr, req, res)
 			results = append(results, res)
@@ -149,7 +149,7 @@ func (s *SecretGen) Generate(proxy *model.Proxy, w *model.WatchedResource, req *
 	return results, model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", cached, cached+regenerated)}, nil
 }
 
-func (s *SecretGen) generate(sr SecretResource, configClusterSecrets, proxyClusterSecrets credscontroller.Controller, proxy *model.Proxy) *discovery.Resource {
+func (s *SecretGen) generate(sr SecretResource, configClusterSecrets, proxyClusterSecrets credscontroller.Controller) *discovery.Resource {
 	// Fetch the appropriate cluster's secret, based on the credential type
 	var secretController credscontroller.Controller
 	switch sr.Type {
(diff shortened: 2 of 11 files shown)


Arhell (commented):
/ok-to-test

rveerama1:
As it depends upon [api](https://github.com/istio/api/pull/2261) changes, build fails. I will move to WIP.

rveerama1:
@howardjohn @jacob-delgado @ericvn PTAL about implementation of https://github.com/istio/api/pull/2261private key provider addition.

security/pkg/nodeagent/sds/sdsservice.go
```diff
@@ -228,19 +271,24 @@ func toEnvoySecret(s *security.SecretItem, caRootPath string) *tls.Secret {
 			},
 		}
 	} else {
-		secret.Type = &tls.Secret_TlsCertificate{
-			TlsCertificate: &tls.TlsCertificate{
-				CertificateChain: &core.DataSource{
-					Specifier: &core.DataSource_InlineBytes{
-						InlineBytes: s.CertificateChain,
+		res := toEnvoySecretWithPrivateKeyProvider(s, pkpConf)
+		if res != nil {
```
yangminzhu, line 275: nit, a little bit more go idiom if `res` is not used beyond the if-statement:
```
if res := toEnvoySecretWithPrivateKeyProvider(s, pkpConf); res != nil {
 ...
} else {
 ...
}
```

pilot/pkg/xds/sds.go
```diff
@@ -304,8 +309,58 @@ func toEnvoyCaSecret(name string, cert []byte) *discovery.Resource {
 	}
 }
 
-func toEnvoyKeyCertSecret(name string, key, cert []byte) *discovery.Resource {
-	res := util.MessageToAny(&envoytls.Secret{
+func toEnvoyKeyCertSecretWithPrivateKeyProvider(name string, key, cert []byte, pkpConf *mesh.PrivateKeyProvider) *anypb.Any {
```
yangminzhu, line 312: this has quite some duplicate code as the other `toEnvoyKeyCertSecretWithPrivateKeyProvider()`, could we refactor and try to share the common logic

security/pkg/nodeagent/sds/server.go
```diff
@@ -43,9 +44,9 @@ type Server struct {
 }
 
 // NewServer creates and starts the Grpc server for SDS.
-func NewServer(options *security.Options, workloadSecretCache security.SecretManager) *Server {
+func NewServer(options *security.Options, workloadSecretCache security.SecretManager, pkpConf *mesh.PrivateKeyProvider) *Server {
```
yangminzhu, line 47: Just some general questions

1. Regarding the `PrivateKeyProvider` extension, is it actually providing any private key? It seems a bit confused to be because based on the PR and my understanding, it does not provide any private key but is just a new type (cryptomb) to wrap the TLS config that tells Envoy to  process it with some hardware assisted acceleration?

2. How are you going to e2e test the feature if it requires some special hardware support?

pilot/pkg/bootstrap/server.go
```diff
@@ -512,6 +513,14 @@ func (s *Server) WaitUntilCompletion() {
 	s.server.Wait()
 }
 
+func (s *Server) fetchPrivateKeyProviderConfig() *v1alpha1.PrivateKeyProvider {
+	meshConf := s.environment.Mesh()
+	proxyConf := meshConf.GetDefaultConfig()
```
jacob-delgado, line 518: I think you may want to use `EffectiveProxyConfig` as there is a merge issue with meshConfig vs a proxy annotation.

security/pkg/nodeagent/sds/sdsservice.go
```diff
@@ -178,7 +183,7 @@ func (s *sdsservice) Generate(_ *model.Proxy, _ *model.PushContext, w *model.Wat
 			names = append(names, i.Name)
 		}
 	}
-	resp, err := s.generate(names)
+	resp, err := s.generate(names, s.pkpConf)
```
jacob-delgado, line 186: Do you need to pass in `s.pkgConf` as you are calling a receiver on `s` anyway?

security/pkg/nodeagent/sds/server.go
rveerama1: > 1. Regarding the `PrivateKeyProvider` extension, is it actually providing any private key? The name is a bit confused to me because based on the PR and my understanding, it does not provide any private key but is just a new type (cryptomb) to wrap the TLS config that tells Envoy to  process it with some hardware assisted acceleration?

Name is bit confusing, but I tried to sync with Envoy naming pattern https://github.com/envoyproxy/envoy/tree/main/api/contrib/envoy/extensions/private_key_providers

security/pkg/nodeagent/sds/sdsservice.go
rveerama1: Done

security/pkg/nodeagent/sds/sdsservice.go
rveerama1: Done

pilot/pkg/xds/sds.go
rveerama1: It has some, but not completely. Anyway I am working on QAT private key provider extension support. I will fine tune in that PR.

security/pkg/nodeagent/sds/server.go
ipuustin: I think there are two things which can be e2e tested. One is the functionality of setting a private key provider configuration to Envoy, and the other is a specific private key provider such as CryptoMB.

The first case is probably not too bad -- for example Envoy uses a software-only private key provider for testing the interfaces (https://github.com/envoyproxy/envoy/blob/main/test/extensions/transport_sockets/tls/test_private_key_method_provider.cc) and we could do something similar. The test private key provider isn't exposed by Envoy though, but that code could probably be used or reimplemented by istio/proxy test/envoye2e directory.

The second case (CryptoMB) is tricky because that will indeed require a very modern Xeon processor. An e2e test could be written but it should be behind a switch, and we need to understand if it's possible to make the CI tests run in an environment which supports AVX-512. @yangminzhu do you have any suggestions how this could be accomplished?

pilot/pkg/bootstrap/server.go
howardjohn: As far as I can tell this is not for Istiod, this is for responding to Gateway requests. Which still have per-node Proxyconfig we need to respect. So I agree with Jacob

pilot/pkg/xds/sds.go
```diff
@@ -304,8 +309,58 @@ func toEnvoyCaSecret(name string, cert []byte) *discovery.Resource {
...
+		return nil
+	}
+
+	return res
+}
+
+func toEnvoyKeyCertSecret(name string, key, cert []byte, pkpConf *mesh.PrivateKeyProvider) *discovery.Resource {
+	var res *anypb.Any
+	if pkpConf != nil {
+		res = toEnvoyKeyCertSecretWithPrivateKeyProvider(name, key, cert, pkpConf)
```
howardjohn, line 354: style nit: I don't think we need these split code paths for `pkpConf != nil`. We can just have a big `switch` that has cases for `mesh.PrivateKeyProvider_Cryptomb`, <any future providers>, and `default`; the default case is the current code.

Essentially the default case *is* a PrivateKeyProvider

pkg/config/validation/validation.go
```diff
@@ -1623,6 +1623,32 @@ func validateServiceSettings(config *meshconfig.MeshConfig) (errs error) {
...
+	switch pkpConf.GetProvider().(type) {
+	case *meshconfig.PrivateKeyProvider_Cryptomb:
+		cryptomb := pkpConf.GetCryptomb()
+		if cryptomb == nil {
+			errs = multierror.Append(errs, errors.New("cryptomb confguration is required"))
+		} else {
+			pollDelay := cryptomb.GetPollDelay()
+			if pollDelay == nil {
+				errs = multierror.Append(errs, errors.New("pollDelay is required"))
+			} else if pollDelay.GetNanos() == 0 {
```
howardjohn, line 1641: isn't `Duration { seconds: 1, nanos: 0 }` valid?

pilot/pkg/bootstrap/server.go
rveerama1: I mean istiod to gateways. I will check how can I use EffectiveProxyconfig.

pkg/config/validation/validation.go
rveerama1: Right, sorry about that, my wooden eyes did see it as oneof{seconds, nanos}. I will fix it.

rveerama1:
@howardjohn @jacob-delgado updated the PR. Please take a look.

security/pkg/nodeagent/sds/server.go
yangminzhu: Thanks for the detail explanation, the naming part is a bit confusing but I think it is fine if you could document it clearly in the code and documentation so that other users does not confuse it with things like Plug in certificate that is really providing its own private/public key to Istio.

Regarding the e2e tests, it would be nice if you can somehow mock the CryptoMB functionality in Envoy then run it like a real e2e test but I have no idea how to do it easily so you probably will end up with tests verifying the Envoy TLS configs generated correctly (but not executing them). Also I have no idea how to require a very modern Xeon processor in the CI environment (it gives you some flexibility in machine type but probably not to the level for custom CPU), @howardjohn might know about about the CI configuration.

pilot/pkg/bootstrap/server.go
rveerama1: Found a reference usage [here](https://github.com/istio/istio/blob/master/pkg/kube/inject/webhook.go#L876). Updated the PR.

howardjohn:
How we do this with other custom integrations is OSS has unit level tests that check the generated XDS is correct. The owner of the platform runs tests on their own system with their own platform to verify it works in the real world.

For example we do this with things like Stackdriver, etc

rveerama1:
> The owner of the platform runs tests on their own system with their own platform to verify it works in the real world.

Sounds a bit bigger task to me to prepare a such setup and tests :). But I will investigate and work on it.

Are the rest of the changes are fine?

pilot/pkg/xds/sds.go
```diff
@@ -304,24 +309,72 @@ func toEnvoyCaSecret(name string, cert []byte) *discovery.Resource {
...
-						InlineBytes: cert,
+func fetchPrivateKeyProviderConfig(proxy *model.Proxy, configStore model.IstioConfigStore, meshConfig *mesh.MeshConfig) *mesh.PrivateKeyProvider {
+	if proxy == nil || meshConfig == nil || configStore == nil {
+		return nil
+	}
+	proxyConfs, err := model.GetProxyConfigs(configStore, meshConfig)
+	if err != nil {
+		return nil
+	}
+	if proxyConf := proxyConfs.EffectiveProxyConfig(proxy.Metadata, meshConfig); proxyConf != nil {
```
howardjohn, line 320: @jacob-delgado  @Monkeyanator this doesn't seem to take `meta.ProxyConfig` into account which makes me suspicious that its not quite right. WDYT?

I mean the existing EffectiveProxyConfig code, not the new stuff

howardjohn (commented):
Looks good except the one proxy config comment

pilot/pkg/xds/sds.go
rveerama1: Gentle ping @jacob-delgado @Monkeyanator

howardjohn: Maybe we should be using ProxyConfigOrDefault?

EffectiveProxyConfig is used only for injection

Monkeyanator: Yeah `ProxyConfigOrDefault` should be used here I think. `EffectiveProxyConfig` is used during injection which is where `meta.ProxyConfig` gets set from IIRC.

rveerama1: Thanks for the comments, I have now modified to use default.

rveerama1: PTAL @jacob-delgado

rveerama1:
> Pod annotations may not be working.

@jacob-delgado I tried to modify httpbin sample with pod annotation without mentioning in global meshconfig and it worked with on pod with cryptomb hardware.

`apiVersion: apps/v1
kind: Deployment
metadata:
  name: httpbin
spec:
  replicas: 1
  selector:
    matchLabels:
      app: httpbin
      version: v1
  template:
    metadata:
      labels:
        app: httpbin
        version: v1
      annotations:
        proxy.istio.io/config: |
          privateKeyProvider:
            cryptomb:
              pollDelay: 170ms
    spec:
      serviceAccountName: httpbin
      containers:
      - image: docker.io/kennethreitz/httpbin
        imagePullPolicy: IfNotPresent
        name: httpbin
        ports:
        - containerPort: 80
`

pilot/pkg/xds/sds.go
```diff
@@ -303,24 +308,57 @@ func toEnvoyCaSecret(name string, cert []byte) *discovery.Resource {
...
-	res := util.MessageToAny(&envoytls.Secret{
-		Name: name,
-		Type: &envoytls.Secret_TlsCertificate{
-			TlsCertificate: &envoytls.TlsCertificate{
-				CertificateChain: &core.DataSource{
-					Specifier: &core.DataSource_InlineBytes{
-						InlineBytes: cert,
+func toEnvoyKeyCertSecret(name string, key, cert []byte, meshConfig *mesh.MeshConfig) *discovery.Resource {
+	var res *anypb.Any
+	pkpConf := meshConfig.GetDefaultConfig().GetPrivateKeyProvider()
```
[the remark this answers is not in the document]
howardjohn: Yeah i tested annotation only works with workload cert, not gateway cert, because of this.

howardjohn (approved):
LGTM once we fix the gateway part, should be small fix

pilot/pkg/xds/sds.go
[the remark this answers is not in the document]
rveerama1: Thank you, fixed now.

pilot/pkg/xds/sds.go
```diff
@@ -308,9 +308,10 @@ func toEnvoyCaSecret(name string, cert []byte) *discovery.Resource {
 	}
 }
 
-func toEnvoyKeyCertSecret(name string, key, cert []byte, meshConfig *mesh.MeshConfig) *discovery.Resource {
+func toEnvoyKeyCertSecret(name string, key, cert []byte, proxy *model.Proxy, meshConfig *mesh.MeshConfig) *discovery.Resource {
 	var res *anypb.Any
-	pkpConf := meshConfig.GetDefaultConfig().GetPrivateKeyProvider()
+	pkpConf := proxy.Metadata.ProxyConfigOrDefault(meshConfig.GetDefaultConfig()).GetPrivateKeyProvider()
+	log.Infof("\n\n pkpconf: %v\n\n", pkpConf)
```
howardjohn, line 314: can you remove this? other lgtm

howardjohn (approved):

pilot/pkg/xds/sds.go
rveerama1: sorry about that.

jacob-delgado (approved):

jacob-delgado:
cc @rveerama1 Looks like you need to re-base, but have the appropriate approvals.

howardjohn (commented):
Thanks for the great work here!

Issue #37700, opened 2022-03-03
The Istio iptables chains created by `pilot-agent` don't get cleaned up when switching  to customize `iptables`


qudongfang:
### Bug Description

The Istio iptables chains created by `pilot-agent` don't get cleaned up when switching to customize `iptables` by setting `ISTIO_CUSTOM_IP_TABLES=true`.

### Version

```prose
client version: 1.13.0
control plane version: 1.11.7
data plane version: 1.11.0 (8 proxies), 1.11.7 (57 proxies)
```

PR #37702, opened 2022-03-03
`istio-start.sh clean` always tries to clean the Istio iptables chains

qudongfang:

This PR resolves https://github.com/istio/istio/issues/37700


jacob-delgado:
LGTM, but I don't know the history of this PR. Any @istio/wg-networking-maintainers have thoughts? If anyone approves then I'll sign it for T&R

howardjohn (changes requested):
This also means if I have custom iptables and istio-clean-iptables happens to match on similar names they will no no longer work. This seems like it could cause a regression for some users

PR #37711, opened 2022-03-04
Determine delta eds regardless of full push or not

hzxuzhonghu:

I donot remember why needed to check this

### DIFF
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -532,9 +532,6 @@ func (eds *EdsGenerator) GenerateDeltas(proxy *model.Proxy, push *model.PushCont
 var deltaConfigTypes = sets.NewSet(gvk.ServiceEntry.Kind)
 
 func shouldUseDeltaEds(req *model.PushRequest) bool {
-	if !req.Full {
-		return false
-	}
 	if len(req.ConfigsUpdated) > 0 {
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
What is the difference? Seems like it would be identical , is there a case it behaves different?

ramaraochavali:
For eds, I do not think it behaves differently?

hzxuzhonghu:
With the check, it means never do delta eds when full push. What we need is to make delta work even for full push.

howardjohn:
> With the check, it means never do delta eds when full push. What we need is to make delta work even for full push.

Isn't it "never do delta eds when *incremental* push."? And incremental is identical to delta

ramaraochavali:
I am confused why do we need "delta" for eds - Is n't it same as what we do now for non full push?

hzxuzhonghu:
Delta eds and incremental eds is not equal.

Say if service is created, we need to do full push. And In Stow, we need to push all the eds of all services.

But for delta, we only need to push eds for the single service.

howardjohn:
but this change only impacts  !full I thought

hzxuzhonghu:
Ok, to align with delta semantics, we also need to notify the removed resources

howardjohn:
But for `!full` there is no removed?

hzxuzhonghu:
I was confused by the double negative check.

Issue #37691, opened 2022-03-03
Cross namespaces VirtualServices have a different behavior since Istio 1.12


djannot:
### Bug Description

I'm deploying the bookinfo application in 2 different namespaces.
The `productpage` service in the `bookinfo-frontends` namespace.
The other services in the `bookinfo-backends` namespace.

Then, I create the following VirtualService in the `bookinfo-frontends` namespace:

```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: reviews
  namespace: bookinfo-frontends
spec:
  hosts:
    - reviews.bookinfo-backends.svc.cluster.local
  http:
  - route:
    - destination:
        host: reviews.bookinfo-backends.svc.cluster.local
    timeout: 0.5s
```

But it's not applied.

If I run the `istioctl pc route deploy/productpage-v1.bookinfo-frontends --name 9080` command, it confirms that the VirtualService isn't applied:

```
NAME     DOMAINS                                                     MATCH     VIRTUAL SERVICE
9080     details.bookinfo-backends, 10.2.100.220                     /*        
9080     productpage, productpage.bookinfo-frontends + 1 more...     /*        
9080     ratings.bookinfo-backends, 10.2.255.201                     /* 
9080     reviews.bookinfo-backends, 10.2.53.134                      /*
```

Now, if I create the same VirtualService in the `bookinfo-backends` namespace, it's correctly applied.

Here is the output of the same command:

```
NAME     DOMAINS                                                     MATCH     VIRTUAL SERVICE
9080     details.bookinfo-backends, 10.2.100.220                     /*        
9080     productpage, productpage.bookinfo-frontends + 1 more...     /*        
9080     ratings.bookinfo-backends, 10.2.255.201                     /*
9080     reviews.bookinfo-backends, 10.2.53.134                      /*        reviews.bookinfo-backends
```

In previous versions (tested in 1.11.7), it works with the VirtualService in `bookinfo-frontends`.
But in newer versions (tested in 1.12.4 and 1.13.1), it doesn't.

### Version

```prose
$ istioctl version
client version: 1.12.4
control plane version: 1.12.4
data plane version: 1.12.4 (11 proxies)

$ kubectl version --short
Client Version: v1.23.1
Server Version: v1.21.1
```

nrjpoddar:
This feels like a regression that we should have caught in our integration tests.

@howardjohn @stevenctl can you shed more light here?

howardjohn:
[...]

PR #37721, opened 2022-03-04
Fix virtualservice selecting: service and virtualservice namespace should not be enforced equally 

hzxuzhonghu:

Fix: https://github.com/istio/istio/issues/37691

### DIFF
--- a/pilot/pkg/model/sidecar.go
+++ b/pilot/pkg/model/sidecar.go
@@ -434,6 +434,7 @@ func convertIstioListenerToWrapper(ps *PushContext, configNamespace string,
 	}
 
 	out.listenerHosts = make(map[string][]host.Name)
+	listenerHosts := make([]host.Name, 0)
 	for _, h := range istioListener.Hosts {
 		parts := strings.SplitN(h, "/", 2)
 		if len(parts) < 2 {
@@ -447,10 +448,11 @@ func convertIstioListenerToWrapper(ps *PushContext, configNamespace string,
 			out.listenerHosts[parts[0]] = make([]host.Name, 0)
 		}
 		out.listenerHosts[parts[0]] = append(out.listenerHosts[parts[0]], host.Name(parts[1]))
+		listenerHosts = append(listenerHosts, host.Name(parts[1]))
 	}
 
 	vses := ps.VirtualServicesForGateway(configNamespace, constants.IstioMeshGateway)
-	out.virtualServices = SelectVirtualServices(vses, out.listenerHosts)
+	out.virtualServices = SelectVirtualServices(vses, listenerHosts)
 	svces := ps.servicesExportedToNamespace(configNamespace)
 	out.services = out.selectServices(svces, configNamespace, out.listenerHosts)
 
--- a/pilot/pkg/model/virtualservice.go
+++ b/pilot/pkg/model/virtualservice.go
@@ -47,9 +47,7 @@ func SelectVirtualServices(virtualServices []config.Config, hosts []host.Name) [
 				}
 			}
 		}
-
 	}
-
 	return out
 }
 
--- a/pilot/pkg/model/virtualservice.go
+++ b/pilot/pkg/model/virtualservice.go
@@ -30,48 +30,27 @@ import (
 
 // SelectVirtualServices selects the virtual services by matching given services' host names.
 // This is a common function used by both sidecar converter and http route.
-func SelectVirtualServices(virtualServices []config.Config, hosts map[string][]host.Name) []config.Config {
-	importedVirtualServices := make([]config.Config, 0)
-
-	vsset := sets.NewSet()
-	addVirtualService := func(vs config.Config, hosts host.Names) {
-		vsname := vs.Name + "/" + vs.Namespace
-		rule := vs.Spec.(*networking.VirtualService)
-		for _, ih := range hosts {
-			// Check if the virtual service is already processed.
-			if vsset.Contains(vsname) {
-				break
-			}
-			for _, h := range rule.Hosts {
-				// VirtualServices can have many hosts, so we need to avoid appending
-				// duplicated virtualservices to slice importedVirtualServices
-				if ih.Matches(host.Name(h)) {
-					importedVirtualServices = append(importedVirtualServices, vs)
-					vsset.Insert(vsname)
-					break
-				}
-			}
-		}
-	}
-	for _, c := range virtualServices {
-		configNamespace := c.Namespace
+// For performance, this should be called after hosts are filtered.
+func SelectVirtualServices(virtualServices []config.Config, hosts []host.Name) []config.Config {
+	out := make([]config.Config, 0)
+	for _, vs := range virtualServices {
 		// Selection algorithm:
 		// virtualservices have a list of hosts in the API spec
 		// if any host in the list matches one service hostname, select the virtual service
 		// and break out of the loop.
-
-		// Check if there is an explicit import of form ns/* or ns/host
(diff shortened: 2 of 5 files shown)


hzxuzhonghu:
@ramaraochavali @howardjohn I can not think of a case why check the namespace before?

ramaraochavali (commented):
LGTM

howardjohn:
I didn't have a chance to review this closely but I did write some unit tests locally that fail before and pass with this PR, so I assume it fixes the issue. will send a PR soon

hzxuzhonghu:
@[howardjohn](https://github.com/howardjohn) ptal, it need to be fixed ASAP

pilot/pkg/model/sidecar.go
```diff
@@ -447,10 +448,11 @@ func convertIstioListenerToWrapper(ps *PushContext, configNamespace string,
 			out.listenerHosts[parts[0]] = make([]host.Name, 0)
 		}
 		out.listenerHosts[parts[0]] = append(out.listenerHosts[parts[0]], host.Name(parts[1]))
+		listenerHosts = append(listenerHosts, host.Name(parts[1]))
 	}
 
 	vses := ps.VirtualServicesForGateway(configNamespace, constants.IstioMeshGateway)
-	out.virtualServices = SelectVirtualServices(vses, out.listenerHosts)
+	out.virtualServices = SelectVirtualServices(vses, listenerHosts)
```
howardjohn, line 455: Its not obvious why we use different listenerHosts. I sort of get it in the context of this PR, but the code would be confusing looking at it later. Can we add comments?

pilot/pkg/model/sidecar.go
```diff
@@ -447,10 +448,11 @@ func convertIstioListenerToWrapper(ps *PushContext, configNamespace string,
 			out.listenerHosts[parts[0]] = make([]host.Name, 0)
 		}
 		out.listenerHosts[parts[0]] = append(out.listenerHosts[parts[0]], host.Name(parts[1]))
+		listenerHosts = append(listenerHosts, host.Name(parts[1]))
```
howardjohn, line 451: By only taking into account the hostname and not namespace, we end up importing virtual services from namespaces we shouldn't. Take this test case for example:

```go
		{
			name: "sidecar filter",
			cfg: "" +
				vs(t, vsArgs{
					Namespace: "not-default",
					Match:     "*.example.com",
					Dest:      "arbitrary.example.com",
				}) +
				vs(t, vsArgs{
					Namespace: "default",
					Match:     "explicit.example.com",
					Dest:      "explicit.example.com",
				}) +
				sidecarScope(t, scArgs{
					Namespace: "not-default",
					Egress:    []string{"not-default/*.example.com", "not-default/not-default.example.org"},
				}),
			proxy:     proxy("not-default"),
			routeName: "80",
			expected: map[string][]string{
				// even though there is an *.example.com, since we do not import it we should create a wildcard matcher
				"*.example.com": {"outbound|80||arbitrary.example.com"},
				"explicit.example.com": nil,
			},
		},
```

With this PR, we end up with a route for `explicit.example.com`. This should not happen since its from `default` namespace which we do not import

howardjohn: I added all the tests in https://github.com/istio/istio/pull/37813. can you pull those into this PR?

hzxuzhonghu: >By only taking into account the hostname and not namespace, we end up importing virtual services from namespaces we shouldn't.

Listener hosts namespace has nothing todo with vritualservice namespace, see the bug reported.

hzxuzhonghu: >I don't think that's true. Sidecar scope DOES apply the VS.

Nope, Where is the doc saying that?

hzxuzhonghu:
The root cause is this is a legacy bug in selecting vs. But at that time, we do not select vs when build route.

https://github.com/istio/istio/commit/dea2be520667359e0f44e7d1db24b01f26b16741#diff-bbdf0a1c28ba55903555b88d31bf01e0a7a29bdd88e29142303687f78581aa0eR184

After i added select vs in above commit, you can see the selecting logic is not same as that one used in sidecarscope building.

Then @ramaraochavali resolved the TODO, and the selecting logic is changed to consider vs namespace.

pilot/pkg/model/sidecar.go
howardjohn: "or a service specified using a ServiceEntry or VirtualService" 
https://istio.io/latest/docs/reference/config/networking/sidecar/#IstioEgressListener

beside, regardless if doc this is how it has behaved since Istio 1.1

hzxuzhonghu: It does not say the namespace should equal to vs namespace

Because we can define a vs in ns A, but it can point to a service in namespace B

howardjohn: This is how it behaved in Istio 1.1 and beyond. If you want to change the behavior of the established API maybe it would be preferred to start with a design doc explaining the use cases, a migration path, etc. Without that I would recommend we fix the regression and go back to the original behavior as documented.

hzxuzhonghu: I donot think this is changing semantic, it is a bug since beginning.

hzxuzhonghu: The bugs is hidden because , sidecar egress hosts is not widely used in combine with vs

howardjohn: I don't understand why you think it's a bug? 

The ONLY place explicit.example.com.is defined is in ns "default"
 the proxy ONLY imports from namespace "not-default"

so why should that proxy see explicit.example.com?

howardjohn: I believe the behavior prior to these changes (eg in 1.11) was correct, and changes to it should to through a design doc

hzxuzhonghu: explicit.example.com is the vs host right, but how about the service is in `node-default` namespace.

hzxuzhonghu: cc @costinm may participate the original discussion

howardjohn: There is no service in not-default called "explicit.example.com". That is only defined in "default" VirtualService

hzxuzhonghu: Yes, but waht if the svc is in `not-default`

hzxuzhonghu: This is the scenario the issue reported

howardjohn: In the users case they are importing everything. In the test above we are trying to restrict what is imported

So in user case, we restricted too much. But now we do not restrict enough in this PR

hzxuzhonghu: Oh, i get to know the issue of this pr

hzxuzhonghu: @howardjohn I have to clarify there are 2 rounds of  vs selecting logic. 

1. when we build sidecarscope  --> selected `vs sets 1`
2.  when we build route, like for 8080, only for service with 8080. This time we  select vses from the  `vs sets 1`

hzxuzhonghu: In your tests, it is route 80, which is very different. You can see the comments I added here 

https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/v1alpha3/httproute.go#L274

hzxuzhonghu: https://github.com/istio/istio/pull/33764#issuecomment-873505981

hzxuzhonghu:
close

PR #37722, opened 2022-03-04
check key and cert are valid before sending to Envoy

ramaraochavali:
Currently we only watch cert and assume when cert changes key is also changed (wait for some debounce duration) and send it to Envoy. We have seen in some cases cert changes, but key is not changed with in the debounce period, leading to sending incorrect cert and key combination to Envoy resulting in cluster/listener NACKs.

This PR validates that key and cert form a correct combination with a retry (starting with debounce , upto max 1min). If the  key and cert are not a valid combination it returns error - to avoid NACKs.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [X] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/security/pkg/nodeagent/cache/secretcache_test.go
+++ b/security/pkg/nodeagent/cache/secretcache_test.go
@@ -377,13 +377,17 @@ func runFileAgentTest(t *testing.T, sds bool) {
 	if err := file.AtomicWrite(sc.existingCertificateFile.CertificatePath, testcerts.RotatedCert, os.FileMode(0o644)); err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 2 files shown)


security/pkg/nodeagent/cache/secretcache_test.go
```diff
@@ -377,21 +377,24 @@ func runFileAgentTest(t *testing.T, sds bool) {
 	if err := file.AtomicWrite(sc.existingCertificateFile.CertificatePath, testcerts.RotatedCert, os.FileMode(0o644)); err != nil {
 		t.Fatal(err)
 	}
+	if err := file.AtomicWrite(sc.existingCertificateFile.PrivateKeyPath, testcerts.RotatedKey, os.FileMode(0o644)); err != nil {
```
ramaraochavali, line 380: The new code caught this error :-)

howardjohn (approved):

PR #37730, opened 2022-03-04
gateway: fix error log when deleting a Gateway

markmc:
Removes error logs like:

```
  error   gateway unable to fetch Gateway: gateway.gateway.networking.k8s.io "gateway" not found  gateway=istio-ingress/gateway
```

Seems the intent here is to ignore not found errors.

- [X] Networking


howardjohn (approved):

Arhell (commented):
/ok-to-test

Issue #37663, opened 2022-03-02
Telemetry API: Support JSON format selection with default behavior


douglas-reid:
**Describe the feature request**

Right now, with the provider config for Envoy File Access Logs, if a `logFormat` is selected, it must be fully-specified. Unfortunately, the only way to get JSON serialization is to use `logFormat`. For users that would like JSON logs, it would be good to provide an easy way to get the default logging behavior.

Propose supporting an empty set of `labels` as a sentinel value for default logging behavior (with JSON serialization).

**Describe alternatives you've considered**

- Separate format selection from content description in format.  This would potentially be cleaner from a variety of perspectives, but requires an API change inside of MeshConfig. This is undesirable for a couple of reasons, including the long-standing need to move provider config out of MeshConfig.

**Affected product area (please put an X in all that apply)**

[ X ] Extensions and Telemetry
[ X ] User Experience

PR #37733, opened 2022-03-04
telemetry api: allow default json format for envoy file logger

douglas-reid:
Currently, with the Telemetry API and EnvoyFileAccessLogger extension provider, there is no easy way to select the Istio-default JSON logging format. This PR attempts to add support for that behavior by treating an empty `labels` set as an indication of selection of default behavior (as there is no point to having an empty access log).

- [ X ] User Experience

Fixes: #37663


howardjohn (approved):

zirain:
@douglas-reid need lint

sedflix:
@douglas-reid any example of how this can be used?

Anand25021998:
+1. I couldn't find any example usage for this.

PR #37744, opened 2022-03-04
[TF] More echo cleanup

nmittler:
- Removing WorkloadPort and just adding to the Ports list.
- Updated common ports to be a single list that includes both service and workload-only ports.
- Renaming CallOptions.Target->To. Just a bit more concise.
- Added Ports utility and moving aggregate port operations there.
- Added Configurable and WorkloadContainer interfaces in preparation for allowing different target types (e.g. Instances).
- Added Cluster to Workload in preparation to switching to use workloads for accessing all clusters for a set of Instances.


tests/integration/security/util/framework.go
```diff
@@ -101,58 +101,62 @@ func EchoConfig(name string, ns namespace.Instance, headless bool, annos echo.An
...
+				Protocol:     protocol.HTTPS,
+				TLS:          true,
 			},
 			{
-				Port:     8084,
-				Protocol: protocol.HTTPS,
-				TLS:      true,
+				Name:         "tcp-8084",
+				WorkloadPort: 8084,
+				Protocol:     protocol.HTTPS,
```
howardjohn, line 158: do these not need NO_SERVICE_PORT?

howardjohn (approved):

tests/integration/security/util/framework.go
nmittler: Ah good catch. Fixed.

zirain (approved):

PR #37762, opened 2022-03-06
bring back catch all virtual host caching

ramaraochavali:
This PR brings the caching of `catchallvirtualhost` back and only creates a copy when there is an envoy filter. So only users who use it pay the price.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu:
Agree with John, and I believe the catchall build is not a perf bottle neck now.  Minor improvement vs bad readability, we need to weigh off

ramaraochavali:
While I agree that we need a generic solution for all such cached things (again I do not know how to do - may be Envoy filters code should have its own generic merge logic like copyonwrite that checks if it is in the cached list, just do a clone and merge), What is the issue with readability there though?

ramaraochavali:
@howardjohn @hzxuzhonghu I tried to make the merge generic and implemented for routes. Please see if this makes sense. I can change other patch merges.

pilot/pkg/networking/core/v1alpha3/envoyfilter/rc_patch.go
```diff
@@ -251,7 +256,7 @@ func patchHTTPRoute(patchContext networking.EnvoyFilter_PatchContext,
 				*routesRemoved = true
 				return
 			} else if rp.Operation == networking.EnvoyFilter_Patch_MERGE {
-				proto.Merge(virtualHost.Routes[routeIndex], rp.Value)
+				safeMerge(virtualHost.Routes[routeIndex].Name, virtualHost.Routes[routeIndex], rp.Value)
```
howardjohn, line 259: `safeMerge` is not recursively checking the `name` to see if its safe, so don't we end up with an issue if we merge at a high level? For example:

```
knownCatchedObjects := ["foo"]
x := {bar:{foo:{}}}
safeMerge(x, ...)
```

We will check name (bar), but then merge and foo will get overwriten

ramaraochavali:
Yes. I want to add that complexity if we are 9k with this approach

hzxuzhonghu:
Can you test how much time does buildCatchallVhost saves?

hzxuzhonghu:
And then compare with deepcopy

ramaraochavali:
> Can you test how much time does buildCatchallVhost saves?

 DeepCopy will always be slower than rebuilding. But the point here,  user who have `catchAlllvirtualhost`  mutating envoy filter will only pay the price.

Another option I was thinking is to detect that an envoy filter exists for catchAll and then build it always rather than using this complicated safe merge.  The catchAllvirtualhost might yield trivial benefit with caching, but I am trying to come up with generic solution so that we get the optimizations for normal users and people who have such envoy filters will pay the price.

hzxuzhonghu:
I mean if a new `buildCatchallVhost `  just cost `ns` level, we do not care about it.

ramaraochavali:
> I mean if a new buildCatchallVhost just cost ns level, we do not care about it.

Yes for `buildCatchallVhost` - It is not a big deal. But I am thinking of generic solution to get benefits of both optimized code for this and socket matches + any thing that comes in future and making it work with envoy filters.

hzxuzhonghu:
The generic way looks meaningful if https://github.com/istio/istio/pull/37762#discussion_r830138503 can be resolved

ramaraochavali:
> The generic way looks meaningful if https://github.com/istio/istio/pull/37762#discussion_r830138503 can be resolved

I am exploring that option

ramaraochavali:
@howardjohn PTAL, if this makes sense

pkg/proto/merge/merge.go
```diff
@@ -58,18 +70,40 @@ var ReplaceMergeFn MergeFunction = func(dst, src protoreflect.Message) {
...
+		return proto.Clone(dst.Interface())
+	}
+	return dst.Interface()
+}
+
 var options = []OptionFn{
 	// Workaround https://github.com/golang/protobuf/issues/1359, merge duration properly
 	MergeFunctionOptionFn((&durationpb.Duration{}).ProtoReflect().Descriptor().FullName(), ReplaceMergeFn),
+	// We cache the catch all virtual host for performance reasons. This helps not to mutate cached object.
+	CloneFunctionOptionFn((&route.VirtualHost{}).ProtoReflect().Descriptor().FullName(), CatchAllVirtualHostCloneFn),
```
howardjohn, line 94: What if someone is legitimately trying to merge with VirtualHost?

howardjohn: oh nevermind, I see we check the name

pkg/proto/merge/merge.go
```diff
@@ -58,18 +70,40 @@ var ReplaceMergeFn MergeFunction = func(dst, src protoreflect.Message) {
 	})
 }
 
+// CatchAllVirtualHostCloneFn creates a clone if src is a catchall virtual host.
+var CatchAllVirtualHostCloneFn CloneFunction = func(dst, src protoreflect.Message) protoreflect.ProtoMessage {
+	catchall := false
+	dst.Range(func(fd protoreflect.FieldDescriptor, v protoreflect.Value) bool {
+		if v.String() == "allow_any" {
```
howardjohn, line 77: I think this means we are converting each field to a string - may be costly?

Can we just get the real `route.VirtualHost` and get the name field?

pkg/proto/merge/merge.go
howardjohn, line 94: Is there any way we can mark objects at creation that they are "must clone"? Then everything cached can be like

```
var cachedConfig = merge.MustClone(&SomeObject{...})
```

ramaraochavali:
@howardjohn PTAL. Addressed comments

pilot/pkg/networking/networking.go
```diff
@@ -204,3 +221,64 @@ func MessageToAny(msg proto.Message) *anypb.Any {
...
+			ClusterSpecifier: &route.RouteAction_Cluster{Cluster: egressCluster},
+			// Disable timeout instead of assuming some defaults.
+			Timeout: notimeout,
+			// Use deprecated value for now as the replacement MaxStreamDuration has some regressions.
+			// nolint: staticcheck
+			MaxGrpcTimeout: notimeout,
+		}
+
+		// Register catchall virtual host with merge logic, so that it clones while merging.
+		return merge.MustClone(&route.VirtualHost{
```
hzxuzhonghu, line 246: If we clone this each time. how could it be efficient?

ramaraochavali: It does not clone every time - it is only cloned if Envoy filter has virtualhost with allow_any that means a user is making change. So he will pay the price. Others it will be efficient.

ramaraochavali:
@howardjohn @hzxuzhonghu PTAL. Any further concerns on this? My goal was to to build this generically so that any cached message can use it - I realize that just for passThroughClusters, we may not need this complexity as the gains are not much.

pilot/pkg/networking/networking.go
hzxuzhonghu: I am still confused about the judgment. And not understand the change to merge at all. 

So I am concerned about two points:

1. the proto merge is already too complex, should we encourage making it more complex and even harder to maintain. I do not think so.
2. buildCatchAllVirtualHost is just constructing a simple struct with almost no inferring, I do not think it is a performance hot point.

ramaraochavali: > buildCatchAllVirtualHost is just constructing a simple struct with almost no inferring, I do not think it is a performance hot point.

See my point below. This is not just built for catchAllVirtualHost. I agree it is not a huge deal to build it every time. But this is meant to build a generic solution and prove that it works with this.

hzxuzhonghu: see point 1

ramaraochavali: > the proto merge is already too complex, should we encourage making it more complex and even harder to maintain. I do not think so.

IMO, at some point we have to take this hit because of envoy filters. We can not loose obvious optimizations (for some it may be very time consuming to build), we may need this complexity. Not sure if there is any simpler way to get both the things - optimzation + envoy filter working

hzxuzhonghu: I suggest we make it when we met some real performance issue.  This does not bring much benefit other than complexity to me.

ramaraochavali: fair point

PR #37767, opened 2022-03-07
support multiple custom hosts

shirleyyLi:

We have a requirement in grayscale release. The control plane needs to support multiple custom hosts. And we can separate multiple values ​​with commas. Of course it is still compatible with only one custom host.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [X] User Experience

### DIFF
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -954,9 +954,12 @@ func (s *Server) initIstiodCerts(args *PilotArgs, host string) error {
 	// Append custom hostname if there is any
 	customHost := features.IstiodServiceCustomHost
 	s.dnsNames = []string{host}
-	if customHost != "" && customHost != host {
-		log.Infof("Adding custom hostname %s", customHost)
-		s.dnsNames = append(s.dnsNames, customHost)
+	cHosts := strings.Split(customHost, ",")
+	for _, cHost := range cHosts {
+		if cHost != "" && cHost != host {
+			log.Infof("Adding custom hostname %s", cHost)
+			s.dnsNames = append(s.dnsNames, cHost)
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


hzxuzhonghu:
What is the use case?

shirleyyLi:
@hzxuzhonghu 
In grayscale release，we want to only change service's selector without restarting istiod pod.

For example，when we want to release a new version of istiod, if istiod pod supports multiple custom hosts which are `Service Stable` and `Service Canary`. At first, only Service Canary selects the new version. we can use Service Canary and releted envoys to test the new version of istiod pod. And then, the tests are successful, we can make the Service Stable select the new version. Then gradually scale up the new version pod and finally replace the old version.

Arhell (commented):
/ok-to-test

hzxuzhonghu:
Not clear to me, the recommended way is https://istio.io/latest/docs/setup/upgrade/canary/.

shirleyyLi:
@hzxuzhonghu
Your recommended way needs to migrate all of the traffic to the new version, which means the configuration of envoys should be updated and they will be restarted. But we do not want to have impact on the existing sidecar proxies or istio gateways. So using service selector and multiple custom hosts to control direction of the traffic is a better way to us.

costinm (approved):
Lgtm

hzxuzhonghu:
> which means the configuration of envoys should be updated and they will be restarted. 

Yes, but older envoy connects to new istiod could introduce unexpected error.  Though we always consider the version mismatch, be careful.

hzxuzhonghu (commented):
I will approve if you update the `ISTIOD_CUSTOM_HOST` help message as well

hzxuzhonghu (approved):

Issue #37709, opened 2022-03-04
RDS cache assertion failure 


howardjohn:
https://storage.googleapis.com/istio-prow/logs/integ-assertion_istio_postsubmit/1499541196886773760/artifacts/pilot-a1082c1b2930451fa875cc0f3/TestTraffic/_test_context/istio-state1044077535/primary-0/istiod-74cdd8db54-vzp7k_discovery.previous.log

cc @hzxuzhonghu

hzxuzhonghu:
a new failure?

howardjohn:
Yeah this was from today. Haven't had a chance to look at it yet

hzxuzhonghu:
OK, will look into it

hzxuzhonghu:
It looks the virtualhosts are not sorted, out of order

howardjohn:
I see. So that should not actually impact real users much, right? Its in one of the lists that order does not matter in envoy (?)

howardjohn:
Same issue in https://storage.googleapis.com/istio-prow/logs/integ-assertion_istio_postsubmit/1499627745921994752/artifacts/pilot-endpointslice-056847cd01a/_suite_context/istio-state-1862642344/primary-0/istiod-5b7c67974c-tnn8t_discovery.previous.log

and https://storage.googleapis.com/istio-prow/logs/integ-assertion_istio_postsubmit/1499759856070955008/artifacts/pilot-endpointslice-681a485a20c/_suite_context/istio-state-2853821471/primary-0/istiod-5b7c67974c-2m72j_discovery.previous.log

So this is either a regression OR caused by the go 1.18 upgrade

howardjohn:
It might be real bug since https://prow.istio.io/view/gs/istio-prow/logs/integ-pilot-istiodless-mc_istio_postsubmit/1499589787684704256 and https://prow.istio.io/view/gs/istio-prow/logs/integ-pilot-istiodremote-mc_istio_postsubmit/1499577121645793280 also fail for similar reasons

and https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/37708/integ-distroless_istio/1499793968668348416

could also be related to the push context change I made

howardjohn:
@hzxuzhonghu my theory was the recent change with push context broke because we still set PushTime to `now()` but used old push context. I changed that in https://github.com/istio/istio/pull/37732, but still seems to have some issues...

hzxuzhonghu:
[...]

PR #37800, opened 2022-03-08
Pilot: store latest push instead of global

howardjohn:
Attempt 2 at #37604, while not causing regression in #37709.

Fixes https://github.com/istio/istio/issues/37709

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -308,6 +308,10 @@ type Proxy struct {
 	// Historically, the latest was used which can cause problems when computing whether a push is
 	// required, as the computed sidecar scope version would not monotonically increase.
 	LastPushContext *PushContext
+	// LastPushTime records the time of the last push. This is used in conjunction with
+	// LastPushContext; the XDS cache depends on knowing the time of the PushContext to determine if a
+	// key is stale or not.
+	LastPushTime time.Time
 }
 
 // WatchedResource tracks an active DiscoveryRequest subscription.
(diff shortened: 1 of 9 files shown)


pilot/pkg/model/context.go
```diff
@@ -302,6 +302,16 @@ type Proxy struct {
...
+
+	// LastPushContext stores the most recent push context for this proxy. This will be monotonically
+	// increasing in version. Requests should send config based on this context; not the global latest.
+	// Historically, the latest was used which can cause problems when computing whether a push is
+	// required, as the computed sidecar scope version would not monotonically increase.
+	LastPushContext *PushContext
+	// LastPushTime records the time of the last push. This is used in conjunction with
+	// LastPushContext; the XDS cache depends on knowing the time of the PushContext to determine if a
+	// key is stale or not.
+	LastPushTime time.Time
```
hzxuzhonghu, line 314: instead of two separate fields, I am thinking we should make time/token in PushContext. Originally I want to use push version as token, however it is not an ever-increasing integer.

howardjohn: I was considering that.... one difference is this is per PushReuest vs per PushContext. I cannot think of any reason this would differ though so I can try that out

howardjohn:
This ran 7x in CI, and we have 16 integ tests. So 112 test iterations, 0 failures. I think it is safe to assume this works... @hzxuzhonghu

hzxuzhonghu (approved):

PR #37803, opened 2022-03-08
Don't add credHelper if there's no GCP credentials

TrevorTaoARM:

As we know, gcloud auth configure-docker adds the Docker credHelper
entry to Docker's configuration file, or creates the file
 if it doesn't exist.
But if there's no GCP credential file(defined by env
GOOGLE_APPLICATION_CREDENTIALS), which is often met in internal CI,
it would lead to a failure when doing
the normal docker operation:
"
ERROR: error getting credentials - err: exit status 1, out:
 `You do not currently have an active account selected.
See https://cloud.google.com/sdk/docs/authorizing for more information
"

So just ignore the credHelper adding operation if there's no
GCP credentials.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience


prow/lib.sh
```diff
@@ -41,12 +41,14 @@ function trace() {
 }
 
 function setup_gcloud_credentials() {
-  if [[ $(command -v gcloud) ]]; then
-    gcloud auth configure-docker -q
-  elif [[ $(command -v docker-credential-gcr) ]]; then
-    docker-credential-gcr configure-docker
-  else
-    echo "No credential helpers found, push to docker may not function properly"
+  if [[ -n "${GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
```
howardjohn, line 44: We should not check for `GOOGLE_APPLICATION_CREDENTIALS` ideally, this is the legacy mode where we mount SA token. Jobs are now using workload identity.

I agree we need some check but I am not sure it should be this. Unfortunately I don't know a great way to actually identity that. If we just add a `|| true` is it sufficient or do we need something more?

TrevorTaoARM: Yes, actually we need a way to bypass the setup the gcloud credentials for docker if it does not exist. We just choosed this way to bypass the adding credHelpers to docker in our internal CI.

I see the env GOOGLE_APPLICATION_CREDENTIALS value is still in istio/test-infra/prow/cluster/jobs/all-presets.yaml, so does the "workload identity" here mean?

howardjohn: It's defined there but no jobs use that anymore https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity

PR #37813, opened 2022-03-08
Revert selecting virtual services

howardjohn:
Revert https://github.com/istio/istio/pull/33764 to fix https://github.com/istio/istio/issues/37691. Alternative is https://github.com/istio/istio/pull/37721.

This was primarily to get the new tests I wrote passing. These same tests pass on the release-1.11 branch, ensuring we have no behavioral regression. I am ok moving forward with https://github.com/istio/istio/pull/37721 if we pull in these tests there and they all pass (currently they do not).

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/httproute.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute.go
@@ -48,7 +48,8 @@ const (
 func (configgen *ConfigGeneratorImpl) BuildHTTPRoutes(
 	node *model.Proxy,
 	req *model.PushRequest,
-	routeNames []string) ([]*discovery.Resource, model.XdsLogDetails) {
+	routeNames []string,
+) ([]*discovery.Resource, model.XdsLogDetails) {
 	routeConfigurations := make([]*discovery.Resource, 0)
 
 	efw := req.Push.EnvoyFilters(node)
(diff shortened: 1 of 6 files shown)


hzxuzhonghu:
The root cause is `SelectVirtualServices` is not right, if you define a sidecar with egress hosts set the bug is still same.

howardjohn:
My understanding is this is reverting the SelectVS which is,as you
mentioned,not right. We can forward-fix it like your PR but it doesn't
handle all cases (see comment on your PR)

hzxuzhonghu:
WIll import the tests

howardjohn:
SGTM. I would obviously prefer your PR btw and don't want to merge this.
Was just to develop the tests

pilot/pkg/networking/core/v1alpha3/httproute_test.go
```diff
@@ -1239,6 +1240,9 @@ func TestSidecarOutboundHTTPRouteConfig(t *testing.T) {
 				"test.com:8080": {
 					"test.com:8080": true, "8.8.8.8:8080": true,
 				},
+				"test-svc.testns.svc.cluster.local:80": {
+					"test-svc.testns.svc.cluster.local": true, "test-svc.testns.svc.cluster.local:80": true,
```
hzxuzhonghu, line 1244: which service is test-svc.testns.svc.cluster.local:80

hzxuzhonghu: I wonder this is not right

hzxuzhonghu:
https://github.com/istio/istio/pull/33764 is not the cause,  after https://github.com/istio/istio/pull/34247 it is broken.

GregHanson (approved):
waiting for final comments from @hzxuzhonghu

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -215,11 +218,55 @@ func (configgen *ConfigGeneratorImpl) buildSidecarOutboundHTTPRouteConfig(
...
+
+			if match {
+				break
+			}
+		}
+
+		if match {
+			out = append(out, c)
+		} else {
+			log.Errorf("howardjohn: drop %v", c.Name)
```
hzxuzhonghu, line 257: drop

hzxuzhonghu (commented):
LGTM

hzxuzhonghu (approved):

PR #37817, opened 2022-03-08
add dual stack support for echo server [TF]

ycai-aspen:
Modify the Echo Server to support Dual Stack. Based on the IP addresses of the Pod, Echo server will listen IPv4 or IPv6 addresses.
Part of https://github.com/istio/istio/issues/37533

- [ ] Docs
- [ ] Installation
- [x ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience

- [x ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

### DIFF
--- a/pkg/test/echo/server/endpoint/util.go
+++ b/pkg/test/echo/server/endpoint/util.go
@@ -28,7 +28,16 @@ import (
 var epLog = log.RegisterScope("endpoint", "echo serverside", 0)
 
 func listenOnAddress(ip string, port int) (net.Listener, int, error) {
-	ln, err := net.Listen("tcp", net.JoinHostPort(ip, strconv.Itoa(port)))
+	parsedIP := net.ParseIP(ip)
+	ipBind := "tcp"
+	if parsedIP != nil {
+		if parsedIP.To4() == nil && parsedIP.To16() != nil {
+			ipBind = "tcp6"
+		} else if parsedIP.To4() != nil {
+			ipBind = "tcp4"
+		}
+	}
+	ln, err := net.Listen(ipBind, net.JoinHostPort(ip, strconv.Itoa(port)))
 	if err != nil {
 		return nil, 0, err
 	}
@@ -38,11 +47,19 @@ func listenOnAddress(ip string, port int) (net.Listener, int, error) {
 }
 
 func listenOnAddressTLS(ip string, port int, cfg *tls.Config) (net.Listener, int, error) {
-	ln, err := tls.Listen("tcp", net.JoinHostPort(ip, strconv.Itoa(port)), cfg)
+	ipBind := "tcp"
+	parsedIP := net.ParseIP(ip)
+	if parsedIP != nil {
+		if parsedIP.To4() == nil && parsedIP.To16() != nil {
+			ipBind = "tcp6"
+		} else if parsedIP.To4() != nil {
+			ipBind = "tcp4"
+		}
+	}
+	ln, err := tls.Listen(ipBind, net.JoinHostPort(ip, strconv.Itoa(port)), cfg)
 	if err != nil {
 		return nil, 0, err
 	}
-
 	port = ln.Addr().(*net.TCPAddr).Port
 	return ln, port, nil
 }
--- a/pkg/test/echo/server/instance.go
+++ b/pkg/test/echo/server/instance.go
@@ -133,7 +133,7 @@ func (s *Instance) Start() (err error) {
 }
 
 func getBindAddresses(ip string) []string {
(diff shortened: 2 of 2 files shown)


pkg/test/echo/server/endpoint/util.go
```diff
@@ -28,7 +28,16 @@ import (
...
+	parsedIP := net.ParseIP(ip)
+	ipBind := "tcp"
+	if parsedIP != nil {
+		if parsedIP.To4() == nil && parsedIP.To16() != nil {
+			ipBind = "tcp6"
+		} else if parsedIP.To4() != nil {
+			ipBind = "tcp4"
+		}
+	}
+	ln, err := net.Listen(ipBind, net.JoinHostPort(ip, strconv.Itoa(port)))
```
howardjohn, line 40: Confused why any of this change is needed? On my machine it automatically handle dual stack?
```
$ netstat -ntlp | rg 8080
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)
tcp6       0      0 :::8080                 :::*                    LISTEN      3480191/server
```

jacob-delgado: In go when you bind to `0.0.0.0` and use `tcp` when listening on a port it binds to both ipv4 and ipv6 any addresses. When we were doing some of our testing we were trying to strictly enforce communication over IPv4 only or IPv6 only communication as part of the requirements we had.

We used the integration tests as a way to barricade the communication traffic from a client that performed an request over IPv4 to a server that accepted the IPv4 request and the same for IPv6, making sure that IPv4 or IPv6 was consistent from client app to sidecar to sidecar and to the server app. We didn't want to have the tests look at such things as pcaps or iptables counters and this was the solution we arrived at. By having the echo be capable of listening over IPv4 (0.0.0.0 with tcp4 attributes) and/or IPv6 (:: with tcp6 attributes) separately. This way it opens up 2 different listeners.

This also had the added benefit of making sure that IPv6 only and IPv4 only clusters weren't broken as we developed this capability.

In practice most users using go will never setup to listen on both `0.0.0.0` and `::` separately so it's confined to this narrow testing case.

howardjohn: Oh so we are not trying to support dual stack, really, but rather to support _single stack_ to test dual stack?

jacob-delgado: For this case, yes.

It can get a bit confusing since the CNI/CRI may support IPv6 and K8s may not be configured to support dual-stack capability. In the near-future we intend to push a dual-stack integration test that performs the following (or similar):

| Namespace                   | K8s Service config                                        | Echo listens on        |
|-----------------------------|-----------------------------------------------------------|------------------------|
| echo-ipv4-only              | ipFamilies: [IPv4]  / ipFamilyPolicy: SingleStack            | 0.0.0.0, tcp4          |
| echo-ipv6-only              | ipFamiles: [IPv6] / ipFamilyPolicy: SingleStack             | ::, tcp6               |
| echo-dual-stack-prefer-ipv4 | ipFamilies: [IPv4, IPv6] / ipFamilyPolicy: RequireDualStack | 0.0.0.0, tcp4 ::, tcp6 |
| echo-dual-stack-prefer-ipv6 | ipFamilies: [IPv4, IPv6] / ipFamilyPolicy: RequireDualStack | 0.0.0.0, tcp4 ::, tcp6 |

After this has been setup an echo app will perform the appropriate connectivity tests (IPv4-only, IPv6 only or dual-stack tests for each server) for both positive and negative test cases (connectivity passes and fails where expected). We do not intend to test K8s dual-stack capability (they have their own set of tests), but the lines can get blurry as istio manages endpoints for services. Headless services also have their own set of issues since there's no Service VIP.

There are also additional test cases could be added where we have K8s setup a service to listen on dual-stack, but the service (echo) is only listening on IPv4 or IPv6. I'm imaging that when this goes live such scenarios will happen as it's easy to misconfigure your K8s Services as well as your server apps; this can even be language dependent (I've seen this with some Java apps) as not all libraries/languages will bind to IPv4 and IPv6 any addresses when you bind to `0.0.0.0`.

For the other set of tests that won't exist in `tests/integration/dualstack` neither `ipFamilies` nor `ipFamilyPolicy` will not be set and the echo app will listen on `net.Listen(0.0.0.0, tcp)` as it currently does.

howardjohn (approved):

Issue #37756, opened 2022-03-05
TestEdsCache is flaky


ramaraochavali:
https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/37722/unit-tests_istio/1499786654670393344/build-log.txt

ramaraochavali:
@howardjohn could this be related to push context change? I am seeing this since two days?

howardjohn:
Probably so, need to look into it

howardjohn:
I don't think it is, I can reproduce with and without it

howardjohn:
1.13: `1053 runs so far, 0 failures (100.00% pass rate). 4.69240598s avg, 6.792750143s max, 1.860541828s min`
master: `522 runs so far, 2 failures (99.62% pass rate). 4.825105477s avg, 10.393853101s max, 2.02398546s min`

howardjohn:
Some findings is in master the update is a full push but in 1.13 its a incremental push. IDK why this results in the bug though. I think the diff occurs when we reflect.DeepEquals the services, but its unclear why right now. related to auto alloc

howardjohn:
cc @hzxuzhonghu

hzxuzhonghu:
// to make sure the eds update run in serial to prevent stale ones can override new ones
	// There are multiple threads calling edsUpdate.
	// If all share one lock, then all the threads can have an obvious performance downgrade.
	edsQueue queue.Instance

PR #37824, opened 2022-03-09
Fix TestEdsCache flake

hzxuzhonghu:

run eds cache update in serial and wait for completion

fix https://github.com/istio/istio/issues/37756

### DIFF
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
@@ -382,11 +382,8 @@ func (s *ServiceEntryStore) serviceEntryHandler(_, curr config.Config, event mod
 		keys[instancesKey{hostname: svc.Hostname, namespace: curr.Namespace}] = struct{}{}
 	}
 
-	// wait for the cache update finished
-	wg := sync.WaitGroup{}
-	wg.Add(1)
 	// trigger update eds endpoint shards
-	s.edsUpdateInSerial(keys, false)
+	s.edsUpdateByKeys(keys, false)
 
 	pushReq := &model.PushRequest{
 		Full:           true,
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
@@ -381,11 +381,12 @@ func (s *ServiceEntryStore) serviceEntryHandler(_, curr config.Config, event mod
 	for _, svc := range nonDNSServices {
 		keys[instancesKey{hostname: svc.Hostname, namespace: curr.Namespace}] = struct{}{}
 	}
+
+	// wait for the cache update finished
+	wg := sync.WaitGroup{}
+	wg.Add(1)
 	// trigger update eds endpoint shards
-	s.edsQueue.Push(func() error {
-		s.edsUpdateByKeys(keys, false)
-		return nil
-	})
+	s.edsUpdateInSerial(keys, false)
 
 	pushReq := &model.PushRequest{
 		Full:           true,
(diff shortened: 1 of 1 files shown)


pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -591,10 +577,27 @@ func (s *ServiceEntryStore) edsUpdate(instances []*model.ServiceInstance, push b
 	for _, i := range instances {
 		keys[makeInstanceKey(i)] = struct{}{}
 	}
+	s.edsUpdateInSerial(keys, push)
+}
+
+// edsUpdateInSerial run s.edsUpdateByKeys in serial and wait for complete.
+func (s *ServiceEntryStore) edsUpdateInSerial(keys map[instancesKey]struct{}, push bool) {
```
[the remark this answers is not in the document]
howardjohn: ok I saw comment from somewhere else actually

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -477,6 +474,7 @@ func (s *ServiceEntryStore) WorkloadInstanceHandler(wi *model.WorkloadInstance,
 	} else {
 		s.serviceInstances.updateInstances(key, instances)
 	}
+	s.mutex.Unlock()
 
 	s.edsUpdate(instances, true)
```
howardjohn, line 479: Not sure I understand this change. I thought the whole point of the EDS was to ensure that we process EDS events in order. But now we release the lock **then** enqueue the update -- so we lose strict ordering?

hzxuzhonghu: The order is guaranteed by enqueueing, in the function, the instances enqueue by order. So they are processed in strict order, I think

howardjohn: But between mutex unlock and s.edsUpdate, anything can happen. In extreme scenario the thread is not live for 10s, and 100s of events are processed in this time. Then finally we wake back up and wipe out the old configs?

hzxuzhonghu: In this case, the service instances store can be updated to the latest state. And in `edsUpdate`, the eds cache is then updated to latest state. I could not think how it can not work.

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -591,10 +577,27 @@ func (s *ServiceEntryStore) edsUpdate(instances []*model.ServiceInstance, push b
...
+		defer wg.Done()
 		s.edsUpdateByKeys(keys, push)
 		return nil
 	})
+	select {
+	// To prevent goroutine leak in tests.
+	case <-s.edsQueue.Closed():
+		return
+	default:
+		wg.Wait()
```
howardjohn, line 599: I did see this leak in tests when running locally. Not sure why - maybe a flake?

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
howardjohn: ok, I think I misunderstood. This is just telling it the keys it needs to update. The rest happens under a lock

howardjohn (commented):
after looking again I think it seems reasonable, just concerned about the leak commented.

cc @ramaraochavali can you look as well? it is complex so good to have 2 eyes

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
hzxuzhonghu: Maybe, the there is a race between wait and queue stopped. Need to fix

ramaraochavali:
LGTM as well. But the method naming is very confusing - some place we update cache and some places we do actual push. I do not have better alternatives though.

hzxuzhonghu:
I have the same feelings about the naming. Could do refactor separately

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
hzxuzhonghu: fixed

ramaraochavali (approved):

howardjohn:
I think this broke the benchmark https://storage.googleapis.com/istio-prow/logs/benchmark-report_istio_postsubmit/1502305181302263808/artifacts/benchmark-log.txt

hzxuzhonghu:
fix in https://github.com/istio/istio/pull/37901

PR #37825, opened 2022-03-09
Fix in resource settings in charts

kfaseela:

The current templates are in such a way that if all the four annotations
for resources are not set, the default values will not be set for the
remaining ones.

- [X ] Configuration Infrastructure
- [ ] Docs
- [X ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
@@ -222,18 +222,26 @@ data:
               requests:
                 {{ if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyCPU`) -}}
                 cpu: "{{ index .ObjectMeta.Annotations `sidecar.istio.io/proxyCPU` }}"
+                {{ else if index .Values.global.proxy.resources "requests" "cpu" }}
+                cpu: "{{ index .Values.global.proxy.resources "requests" "cpu" }}"
                 {{ end }}
                 {{ if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyMemory`) -}}
                 memory: "{{ index .ObjectMeta.Annotations `sidecar.istio.io/proxyMemory` }}"
+                {{ else if index .Values.global.proxy.resources "requests" "memory" }}
+                memory: "{{ index .Values.global.proxy.resources "requests" "memory" }}"
                 {{ end }}
             {{- end }}
             {{- if or (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyCPULimit`) (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyMemoryLimit`) }}
               limits:
                 {{ if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyCPULimit`) -}}
                 cpu: "{{ index .ObjectMeta.Annotations `sidecar.istio.io/proxyCPULimit` }}"
+                {{ else if index .Values.global.proxy.resources "limits" "cpu" }}
+                cpu: "{{ index .Values.global.proxy.resources "limits" "cpu" }}"
                 {{ end }}
                 {{ if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyMemoryLimit`) -}}
                 memory: "{{ index .ObjectMeta.Annotations `sidecar.istio.io/proxyMemoryLimit` }}"
+                {{ else if index .Values.global.proxy.resources "limits" "memory" }}
+                memory: "{{ index .Values.global.proxy.resources "limits" "memory" }}"
                 {{ end }}
             {{- end }}
           {{- else }}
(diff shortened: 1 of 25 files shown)


howardjohn (commented):
Having a field *not set* has a special meaning in Kubernetes. Is it possible to not set a field now?

kfaseela:
> Having a field _not set_ has a special meaning in Kubernetes. Is it possible to not set a field now?

https://github.com/istio/istio/pull/37825/files#diff-5d70e6b65d86515abc1bcdf047a0cbbadff8858086ba2ea07ad85b4c8c274bdeR220 is an "or" condition. So if any of the annotations being set, would avoid all other fields being present under resources. Was this really intentional? especially for cpu and memory limits, leaving it unspecified?

kfaseela:
> Having a field _not set_ has a special meaning in Kubernetes. Is it possible to not set a field now?

@howardjohn : I couldn't find an existing way to achieve this. Tried setting the annotation to "" as well. I can add an additional check to unset if the annotation has a value of "", if that is what you are asking for. Let me know if it makes sense.

kfaseela:
@howardjohn if leaving the annotation not set was intentionally done to make the resources unlimited, is it okay if I push a documentation PR to clarify this?

howardjohn:
My personal preference is just that we document this clear. I can see arguments for the behavior before or after this PR - they both make sense in different contexts. However, I am worried about this unexpectedly changing users behavior, so slightly lean towards the old behavior just because it is the incumbent. 

I believe in the current state, you can still achieve any combination of `{cpu,memory}x{request,limit}`, right? You may just need to set 2/4 annotations when you only care about one. I think that is a reasonable tradeoff.

Note https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#customizing-injection is the more modern way to control these things.

WDYT?

emike922:
> I couldn't find an existing way to achieve this.

I think something like this should do the trick:
```
{{- define "resources" -}}
  {{- $requests := dict -}}
  {{- if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyCPU`) -}}
    {{- $_ := set $requests "cpu" (index .ObjectMeta.Annotations `sidecar.istio.io/proxyCPU`) -}}
  {{- else if (((.Values.global.proxy).resources).requests).cpu -}}
    {{- $_ := set $requests "cpu" .Values.global.proxy.resources.requests.cpu -}}
  {{- end -}}
  {{- if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyMemory`) -}}
    {{- $_ := set $requests "memory" (index .ObjectMeta.Annotations `sidecar.istio.io/proxyMemory`) -}}
  {{- else if (((.Values.global.proxy).resources).requests).memory -}}
    {{- $_ := set $requests "memory" .Values.global.proxy.resources.requests.memory -}}
  {{- end -}}
  {{- range keys $requests -}}
    {{- if eq "" (index $requests .) -}}
      {{- $_ := unset $requests . -}}
    {{- end -}}
  {{- end -}}

  {{- $limits := dict -}}
  {{- if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyCPULimit`) -}}
    {{- $_ := set $limits "cpu" (index .ObjectMeta.Annotations `sidecar.istio.io/proxyCPULimit`) -}}
  {{- else if (((.Values.global.proxy).resources).limits).cpu -}}
    {{- $_ := set $limits "cpu" .Values.global.proxy.resources.limits.cpu -}}
  {{- end -}}
  {{- if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyMemoryLimit`) -}}
    {{- $_ := set $limits "memory" (index .ObjectMeta.Annotations `sidecar.istio.io/proxyMemoryLimit`) -}}
  {{- else if (((.Values.global.proxy).resources).limits).memory -}}
    {{- $_ := set $limits "memory" .Values.global.proxy.resources.limits.memory -}}
  {{- end -}}
  {{- range keys $limits -}}
    {{- if eq "" (index $limits .) -}}
      {{- $_ := unset $limits . -}}
    {{- end -}}
  {{- end -}}
  
  {{- toYaml (dict "requests" $requests "limits" $limits) | nindent 6 -}}
{{- end -}}

...
    resources:
    {{- template "resources" . }}
```

Whether or not you take it, that is for you to decide..

kfaseela:
> My personal preference is just that we document this clear. I can see arguments for the behavior before or after this PR - they both make sense in different contexts. However, I am worried about this unexpectedly changing users behavior, so slightly lean towards the old behavior just because it is the incumbent.
> 
> I believe in the current state, you can still achieve any combination of `{cpu,memory}x{request,limit}`, right? You may just need to set 2/4 annotations when you only care about one. I think that is a reasonable tradeoff.
> 
> Note https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#customizing-injection is the more modern way to control these things.
> 
> WDYT?

It makes sense, if you are worried about existing users facing a challenge with this changed behaviour. I have pushed this documentation PR clarifying the behaviour: https://github.com/istio/istio.io/pull/11066/files

kfaseela:
> > I couldn't find an existing way to achieve this.
> 
> I think something like this should do the trick:
> 
> ```
> {{- define "resources" -}}
>   {{- $requests := dict -}}
>   {{- if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyCPU`) -}}
>     {{- $_ := set $requests "cpu" (index .ObjectMeta.Annotations `sidecar.istio.io/proxyCPU`) -}}
>   {{- else if (((.Values.global.proxy).resources).requests).cpu -}}
>     {{- $_ := set $requests "cpu" .Values.global.proxy.resources.requests.cpu -}}
>   {{- end -}}
>   {{- if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyMemory`) -}}
>     {{- $_ := set $requests "memory" (index .ObjectMeta.Annotations `sidecar.istio.io/proxyMemory`) -}}
>   {{- else if (((.Values.global.proxy).resources).requests).memory -}}
>     {{- $_ := set $requests "memory" .Values.global.proxy.resources.requests.memory -}}
>   {{- end -}}
>   {{- range keys $requests -}}
>     {{- if eq "" (index $requests .) -}}
>       {{- $_ := unset $requests . -}}
>     {{- end -}}
>   {{- end -}}
> 
>   {{- $limits := dict -}}
>   {{- if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyCPULimit`) -}}
>     {{- $_ := set $limits "cpu" (index .ObjectMeta.Annotations `sidecar.istio.io/proxyCPULimit`) -}}
>   {{- else if (((.Values.global.proxy).resources).limits).cpu -}}
>     {{- $_ := set $limits "cpu" .Values.global.proxy.resources.limits.cpu -}}
>   {{- end -}}
>   {{- if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyMemoryLimit`) -}}
>     {{- $_ := set $limits "memory" (index .ObjectMeta.Annotations `sidecar.istio.io/proxyMemoryLimit`) -}}
>   {{- else if (((.Values.global.proxy).resources).limits).memory -}}
>     {{- $_ := set $limits "memory" .Values.global.proxy.resources.limits.memory -}}
>   {{- end -}}
>   {{- range keys $limits -}}
>     {{- if eq "" (index $limits .) -}}
>       {{- $_ := unset $limits . -}}
>     {{- end -}}
>   {{- end -}}
>   
>   {{- toYaml (dict "requests" $requests "limits" $limits) | nindent 6 -}}
> {{- end -}}
> 
> ...
>     resources:
>     {{- template "resources" . }}
> ```
> 
> Whether or not you take it, that is for you to decide..

Thanks for the suggestion @emike922 . Will see what is the best way forward!

Issue #37828, opened 2022-03-09
Deployment cannot be created again with the same name in IstioOperator


tuhao1020:
### Bug Description

sample yaml (iop-test.yaml):
```
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  name: ingressgw-test
  namespace: istio-system
spec:
  components:
    ingressGateways:
    - enabled: true
      k8s:
        replicaCount: 1
        resources:
          limits:
            cpu: "0.2"
            memory: 200Mi
          requests:
            cpu: "0.2"
            memory: 200Mi
      label:
        app: ingressgw-test
        istio: tuhao-test-ingressgw-test
      name: ingressgw-test
      namespace: tuhao-test
  profile: empty
  revision: 1-10
  values:
    gateways:
      istio-ingressgateway:
        autoscaleEnabled: false
        injectionTemplate: gateway
        replicaCount: 1
```

Reproduce steps:
1. `kubuectl apply -f iop-test.yaml`
2. wait for deployment and pod to be created
3. delete the deployment
4. repeat the first step

Expect:
deployment will be created again

Actual:
no deployment created.(you must change `ingressgw-test` to another name in iop-test.yaml) 

### Version

```prose
$ istioctl version
client version: 1.10.5
control plane version: 1.10.5-v3.7.5
data plane version: 1.10.5-v3.7.5 (10 proxies)
$ kubectl version --short
Client Version: v1.21.4
Server Version: v1.21.4
```

tuhao1020:
@zirain  I have closed the old pr and recreated a new one #40800

PR #37830, opened 2022-03-09
remove enabled components in object caches when exec prune

tuhao1020:
Signed-off-by: HaoTu <haotu@alauda.io>

flushes all object caches when exec prune, fix for #37828 

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


operator/pkg/helmreconciler/prune.go
```diff
@@ -105,6 +105,10 @@ func (h *HelmReconciler) Prune(manifests name.ManifestMap, all bool) error {
 // during reconciliation process of controller.
 // It returns the install status and any error encountered.
 func (h *HelmReconciler) PruneControlPlaneByRevisionWithController(iopSpec *v1alpha1.IstioOperatorSpec) (*v1alpha1.InstallStatus, error) {
+	// flushes all object caches when exec prune
+	scope.Info("Flush object caches")
+	cache.FlushObjectCaches()
```
morvencao, line 110: flushing the object cache will clean the cache for all objects when we only want to prune specific revision control plane resources.

morvencao (changes requested):

operator/pkg/helmreconciler/prune.go
```diff
@@ -143,6 +143,18 @@ func (h *HelmReconciler) PruneControlPlaneByRevisionWithController(iopSpec *v1al
...
+
+		crHash, err := h.getCRHash(c)
+		if err != nil {
+			return errStatus, err
+		}
+		scope.Infof("reset cache: %s for CR %s", c, crHash)
+		objCache := cache.GetCache(crHash)
+		if len(objCache.Cache) > 0 {
+			objCache.Mu.Lock()
+			objCache.Cache = map[string]*object.K8sObject{}
```
Chen-Xintong, line 155: Hi, one question here, why do we set an empty object here instead of directly setting it to nil?

tuhao1020: Just initialize outside `cache.GetCache`, if it is set to nil, then it will be initialized in the function

morvencao (commented):
/ok-to-test

tuhao1020:
@morvencao  Are there any other issues with this pr？

PR #37844, opened 2022-03-10
Support IPFamily for echo service

ycai-aspen:
Add support to configure IPFamily for echo service. Add "Addresses" for ClusterIPs. Add config function to get config dump.

### DIFF
--- a/pkg/test/framework/components/echo/config.go
+++ b/pkg/test/framework/components/echo/config.go
@@ -139,10 +139,10 @@ type Config struct {
 	// the same pod.
 	IncludeExtAuthz bool
 
-	// IPFamily for the service
+	// IPFamily for the service. This is optional field. Mainly is used for dual stack testing
 	IPFamilies string
 
-	// IPFamilyPolicy
+	// IPFamilyPolicy. This is optional field. Mainly is used for dual stack testing.
 	IPFamilyPolicy string
 }
 
--- a/pkg/test/framework/components/echo/config.go
+++ b/pkg/test/framework/components/echo/config.go
@@ -138,6 +138,12 @@ type Config struct {
 	// the CUSTOM authorization policy when the ext-authz server is deployed locally with the application container in
 	// the same pod.
 	IncludeExtAuthz bool
+
+	// IPFamily for the service
+	IPFamilies string
+
+	// IPFamilyPolicy
+	IPFamilyPolicy string
 }
 
 // SubsetConfig is the config for a group of Subsets (e.g. Kubernetes deployment).
--- a/pkg/test/framework/components/echo/kube/deployment.go
+++ b/pkg/test/framework/components/echo/kube/deployment.go
@@ -75,6 +75,12 @@ metadata:
 {{- end }}
 {{- end }}
 spec:
+{{- if .IPFamilies }}
+  ipFamilies: [ {{ .IPFamilies }} ]
+{{- end }}
+{{- if .IPFamilyPolicy }}
+  ipFamilyPolicy: {{ .IPFamilyPolicy }}
+{{- end }}
 {{- if .Headless }}
   clusterIP: None
 {{- end }}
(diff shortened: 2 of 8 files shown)


irisdingbj:
LGTM

pkg/test/framework/components/echo/config.go
```diff
@@ -138,6 +138,12 @@ type Config struct {
 	// the CUSTOM authorization policy when the ext-authz server is deployed locally with the application container in
 	// the same pod.
 	IncludeExtAuthz bool
+
+	// IPFamily for the service
```
jacob-delgado, line 142: It should be clarified what happens if the user does not set them to the echo service or that this should only be changed when doing dual-stack testing.

jacob-delgado:
LGTM, but we should also mention why we need the Config dump with EDS in the description

pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -75,6 +75,12 @@ metadata:
 {{- end }}
 {{- end }}
 spec:
+{{- if .IPFamilies }}
+  ipFamilies: [ {{ .IPFamilies }} ]
```
howardjohn, line 79: never want to set both?

pkg/test/framework/components/echo/sidecar.go
```diff
@@ -32,6 +32,11 @@ type Sidecar interface {
 	Config() (*envoyAdmin.ConfigDump, error)
 	ConfigOrFail(t test.Failer) *envoyAdmin.ConfigDump
 
+	// Config of the Envoy instance with EDS included.
+        // This will be used to verify the EDS configuration in dual stack testing.
```
howardjohn, line 36: I don't think we should be introspecting XDS config in e2e tests like I assume this is for - we have plenty of unit tests designed for these purposes

pkg/test/framework/components/echo/kube/deployment.go
ycai-aspen: Both are optional fields. Let the user to decide what to set for the testing.

pkg/test/framework/components/echo/sidecar.go
ycai-aspen: We did this in our testing to verify the configuration besides the traffic. I am ok to leave those out since we are not planning to verify the configuration at this point.

zhlsunshine:
LGTM

pkg/test/framework/components/echo/kube/deployment.go
howardjohn: I mean do we not ever want to test `ipFamilies: [ipv4, ipv6]`?

howardjohn (approved):

pkg/test/framework/components/echo/kube/deployment.go
ycai-aspen: We test those. Here is an example:
```
var deployConfig = map[string][3]string{
        "ipv4":      {"echo-ipv4", "IPv4", "SingleStack"},
        "ipv6":      {"echo-ipv6", "IPv6", "SingleStack"},
        "dual-ipv6": {"echo-dual-ipv6", "IPv6, IPv4", "RequireDualStack"},
        "dual-ipv4": {"echo-dual-ipv4", "IPv4, IPv6", "RequireDualStack"},
}
```
```
echo.Config{
          Service:            deployConfig[name][0],
          Namespace:     app.Namespace,
          Ports:                EchoPorts,
          Subsets:           []echo.SubsetConfig{{}},
          WorkloadOnlyPorts: WorkloadPorts,
          IPFamilies:        deployConfig[name][1],
          IPFamilyPolicy:    deployConfig[name][2],
}
```

howardjohn: I would make `IPFamilies` a `[]string` instead of hacking it in with a `string` with a comma then

Issue #37809, opened 2022-03-08
Investigate slow start sending traffic to unready pods


howardjohn:
@ramaraochavali I think this broke safe start. When new service starts up, I am now seeing `upstream connect error or disconnect/reset before headers. reset reason: connection failure, transport failure reason: delayed connect error: 111`. Seems we are sending traffic to not ready pod. This should never happen

_Originally posted by @howardjohn in https://github.com/istio/istio/issues/37122#issuecomment-1062003373_

PR #37854, opened 2022-03-10
do not send initial unhealthy endpoints on full push

ramaraochavali:
FIx https://github.com/istio/istio/issues/37809

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -159,11 +159,11 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string, 
 		pushType = FullPush
 	}
 
+	ep.mutex.Lock()
+	defer ep.mutex.UnLock()
 	newIstioEndpoints := []*model.IstioEndpoint{}
 	if features.SendUnhealthyEndpoints {
-		ep.mutex.RLock()
 		oldIstioEndpoints := ep.Shards[shard]
-		ep.mutex.RUnlock()
 
 		// Check if new Endpoints are ready to be pushed. This check
 		// will ensure that if a new pod comes with a non ready endpoint,
@@ -200,7 +200,6 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string, 
 
 	}
 
-	ep.mutex.Lock()
 	if features.SendUnhealthyEndpoints {
 		ep.Shards[shard] = newIstioEndpoints
 	} else {
@@ -228,7 +227,6 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string, 
 		Name:      hostname,
 		Namespace: namespace,
 	}: {}})
-	ep.mutex.Unlock()
 
 	return pushType
 }
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -103,7 +103,8 @@ func (s *DiscoveryServer) SvcUpdate(shard model.ShardKey, hostname string, names
 // the hostname-keyed map. And it avoids the conversion from Endpoint to ServiceEntry to envoy
 // on each step: instead the conversion happens once, when an endpoint is first discovered.
 func (s *DiscoveryServer) EDSUpdate(shard model.ShardKey, serviceName string, namespace string,
-	istioEndpoints []*model.IstioEndpoint) {
+	istioEndpoints []*model.IstioEndpoint,
+) {
 	inboundEDSUpdates.Increment()
 	// Update the endpoint shards
 	pushType := s.edsCacheUpdate(shard, serviceName, namespace, istioEndpoints)
@@ -129,7 +130,8 @@ func (s *DiscoveryServer) EDSUpdate(shard model.ShardKey, serviceName string, na
 //
 // Note: the difference with `EDSUpdate` is that it only update the cache rather than requesting a push
 func (s *DiscoveryServer) EDSCacheUpdate(shard model.ShardKey, serviceName string, namespace string,
-	istioEndpoints []*model.IstioEndpoint) {
+	istioEndpoints []*model.IstioEndpoint,
+) {
 	inboundEDSUpdates.Increment()
 	// Update the endpoint shards
 	s.edsCacheUpdate(shard, serviceName, namespace, istioEndpoints)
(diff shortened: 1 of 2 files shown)


pilot/pkg/xds/eds.go
```diff
@@ -183,7 +186,7 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
 		Namespace: namespace,
 	}: {}})
 	ep.mutex.Unlock()
-	if features.SendUnhealthyEndpoints && pushType == IncrementalPush {
```
ramaraochavali, line 186: This is the actual change. Rest all formating changes

howardjohn: Do we have any test coverage?

ramaraochavali: We have test for this function - probably not for this specific case. Let me take a look

hzxuzhonghu, line 186: Should elaborate on the flow under the bracket

ramaraochavali: just below bracket there are comments

ramaraochavali: @howardjohn added test. PTAL
@hzxuzhonghu  We can not set this feature to false by default because the new destination rule API depends on this. If users are sure that they do not use that api, they can set it to false.

pilot/pkg/xds/eds_test.go
```diff
@@ -890,10 +889,10 @@ func edsUpdateInc(s *xds.FakeDiscoveryServer, adsc *adsc.ADSC, t *testing.T) {
 
 	// Update the endpoint with different SA - expect full
 	s.Discovery.MemRegistry.SetEndpoints(edsIncSvc, "",
-		newEndpointWithAccount("127.0.0.3", "account2", "v1"))
+		newEndpointWithAccount("127.0.0.2", "account2", "v1"))
```
ramaraochavali, line 892: This is a test issue. Earlier we were sending a new endpoint.

pilot/pkg/xds/eds.go
```diff
@@ -183,36 +187,43 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
...
 		}
 
-		if !needPush {
+		if pushType != FullPush && !needPush {
 			log.Infof("No push, either old endpoint health status did not change or new endpoint came with unhealthy status, %v", hostname)
 			pushType = NoPush
 		}
+		ep.mutex.Lock()
+		ep.Shards[shard] = newIstioEndpoints
+		ep.mutex.Unlock()
```
howardjohn, line 226: I am not sure why its safe to set `ep.Shards[shard]` twice - here and L167, with different values? Can't the 'wrong' value be used in the mean time?

I would expect we set it once

pilot/pkg/xds/eds.go
```diff
@@ -139,7 +141,8 @@ func (s *DiscoveryServer) EDSCacheUpdate(shard model.ShardKey, serviceName strin
 // It also tracks the changes to ServiceAccounts. It returns whether endpoints need to be pushed and
 // it also returns if they need to be pushed whether a full push is needed or incremental push is sufficient.
 func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string, namespace string,
-	istioEndpoints []*model.IstioEndpoint) PushType {
+	istioEndpoints []*model.IstioEndpoint,
+) PushType {
```
hzxuzhonghu, line 145: suggest stop changing unneeded format within bug fix

pilot/pkg/xds/eds.go
```diff
@@ -183,36 +187,43 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
...
 		// will ensure that if a new pod comes with a non ready endpoint,
 		// we do not unnecessarily push that config to Envoy.
 		// Please note that address is not a unique key. So this may not accurately
 		// identify based on health status and push too many times - which is ok since its an optimization.
 		emap := make(map[string]*model.IstioEndpoint, len(oldIstioEndpoints))
+		// Add new endpoints only if they are ever ready once to shards
+		// so that full push does not send them from shards.
+		newIstioEndpoints := []*model.IstioEndpoint{}
 		for _, oie := range oldIstioEndpoints {
 			emap[oie.Address] = oie
```
hzxuzhonghu, line 201: Does not this matter, there can be many same address istioendpoints for multi port svc

pilot/pkg/xds/eds.go
ramaraochavali: This is unintentional because of automated gofmt (IT has this new format). I will revert.

pilot/pkg/xds/eds.go
ramaraochavali: Yeah. I did not think about that case. Thanks for catching it. Fixed it now.

pilot/pkg/xds/eds.go
ramaraochavali: If it exists in old (even with single port) we will push if its health status changes.

hzxuzhonghu (commented):
lgtm, will leave to @howardjohn

pilot/pkg/xds/eds.go
```diff
@@ -159,62 +159,77 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
...
 
-		if !needPush {
+		if pushType != FullPush && !needPush {
 			log.Infof("No push, either old endpoint health status did not change or new endpoint came with unhealthy status, %v", hostname)
 			pushType = NoPush
 		}
+
 	}
 
+	ep.mutex.Lock()
```
howardjohn, line 203: This is not safe AFAIK. Looks like a pretty classic race.

On L165, we read the data and drop the lock

On L203, we grab the lock again and set the data

Between that, another thread could wipe out the data, add endpoints, etc. Those would be reverted on L203

howardjohn (changes requested):

ramaraochavali:
moved to single lock. PTAL

howardjohn:
I don't think you pushed a change

pilot/pkg/xds/eds.go
```diff
@@ -160,61 +160,74 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
 	}
 
 	ep.mutex.Lock()
-	oldIstioEndpoints := ep.Shards[shard]
-	ep.Shards[shard] = istioEndpoints
-	// Check if ServiceAccounts have changed. We should do a full push if they have changed.
-	saUpdated := s.UpdateServiceAccount(ep, hostname)
+	defer ep.mutex.UnLock()
+	newIstioEndpoints := []*model.IstioEndpoint{}
```
howardjohn, line 164:
```suggestion
	newIstioEndpoints := oldIstioEndpoints
	if features.SendUnhealthyEndpoints {
	   newIstioEndpoints = make([]*model.IstioEndpoint, 0, len(oldIstioEndpoints))
	   
	   
```

Benefits are then on L203 we do not need the switch cases, we just always set to `newIstioEndpoints`. But the bigger thing is pre-allocating the array, since this can be 100s of endpoints and grow a lot

howardjohn (approved):

PR #37857, opened 2022-03-10
feat: helm-chart/gateway Apply all labels to pods also

langecode:

Currently when using the Helm chart for deploying Istio gateway the deployment resource is nicely labelled with all Kubernetes recommended labels, however, the actual pod is only labelled with the selector labels. Would be nice to have all labels on the actual pod also.

### DIFF
--- a/manifests/charts/gateway/templates/deployment.yaml
+++ b/manifests/charts/gateway/templates/deployment.yaml
@@ -25,7 +25,7 @@ spec:
         {{- with .Values.revision }}
         istio.io/rev: {{ . }}
         {{- end }}
-        {{- include "gateway.selectorLabels" . | nindent 8 }}
+        {{- include "gateway.labels" . | nindent 8 }}
     spec:
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:


manifests/charts/gateway/templates/deployment.yaml
```diff
@@ -25,7 +25,7 @@ spec:
         {{- with .Values.revision }}
         istio.io/rev: {{ . }}
         {{- end }}
-        {{- include "gateway.selectorLabels" . | nindent 8 }}
+        {{- include "gateway.labels" . | nindent 8 }}
```
morvencao, line 28: Gateway service is using `selectorLabels` in its selector to selector the gateway pods.

langecode: the `gateway.labels` template includes the `gateway.selectorLabels` (here https://github.com/istio/istio/blob/master/manifests/charts/gateway/templates/_helpers.tpl#L18) so both the service selector and and the selector `matchLabels` in the deployment should be all good. It is just a matter of the pod, which the deployment instantiated through the replica set, also gets the full set of labels as the deployment resource itself.

morvencao (commented):
/ok-to-test

langecode:
I’ll check up on the as soon as I can find a little bit of time.

fre. 11. mar. 2022 kl. 08.49 skrev Istio Automation <
***@***.***>:

PR #37862, opened 2022-03-10
Optimize single node kind cluster

howardjohn:
Recommended by https://github.com/kubernetes/minikube/issues/5682


prow/config/default.yaml
```diff
@@ -16,6 +16,11 @@ kubeadmConfigPatches:
       local:
         # Run etcd in a tmpfs (in RAM) for performance improvements
         dataDir: /tmp/kind-cluster-etcd
+    # We run since node, drop leader election to reduce overhead
```
ericvn, line 19:
```suggestion
    # We run single node, drop leader election to reduce overhead
```

PR #37870, opened 2022-03-10
[TF] Cleanup echo matchers

nmittler:
- Moves matchers to the `match` package, which aids in readability from the call sites.

- Moved match methods out of `echo.Instances` to avoid circular dependency.

- Added a few more options for `Filter`s and `Matcher`s that will be used in subsequent PRs.


pkg/test/framework/components/echo/echotest/echotest.go
```diff
@@ -41,7 +42,7 @@ func New(ctx framework.TestContext, instances echo.Instances) *T {
 	copy(d, instances)
 	t := &T{rootCtx: ctx, sources: s, destinations: d}
 	if ctx.Settings().Skip(echo.VM) {
-		noVM := Not(FilterMatch(echo.IsVirtualMachine()))
+		noVM := Not(FilterMatch(match.IsVirtualMachine()))
```
stevenctl, line 45: Overkill to add `ToNotMatch` and `FromNotMatch`?

stevenctl (approved):
LGTM.

howardjohn (approved):

douglas-reid (approved):

PR #37871, opened 2022-03-10
TestAgent: add GCP/STS tests and remove legacy tests

howardjohn:
[Add tests for the gcp envoy bootstrap](https://github.com/istio/istio/pull/37871/commits/bed0e91e768d9ceb1043a7ce5bea73c4d9821782) 

Currently this has no coverage.

The big changes required here was just regenerating the test certs to
have `localhost` as a SAN - the google_grpc does not allow fake SANs.

The other change needed was to allow a way to pass in a root cert, since
otherwise it is expecting OS trusted certs

[Remove legacy security tests](https://github.com/istio/istio/pull/37871/commits/7d7c45f7352410b5d533f363f3272096b0eb2b5b) 

See https://github.com/istio/istio/issues/37555

These tests have now been fully replaced by TestAgent, which offers far
more coverage and is much faster.

These legacy tests run a custom XDS server, custom bootstrap, etc. They
are extremely different from real world Istio usage, and as such provide
minimal coverage. Additionally, almost all of them have been skipped due
to being flaky, while the others are all in the top 10 slowest tests in
istio (https://github.com/istio/istio/issues/37555).

Since the previous commit completes the lack of coverage in TestAgent to
cover the STS server, this commit removes these legacy tests


nmittler (approved):

pkg/istio-agent/agent_test.go
```diff
@@ -517,6 +523,24 @@ func TestAgent(t *testing.T) {
 		}
 		a.Check(t, security.WorkloadKeyCertResourceName, security.RootCertReqResourceName)
 	})
+	t.Run("GCP", func(t *testing.T) {
```
shankgan, line 526: to clarify: this ensures that as part of agent startup, the bootstrap token is correctly picked up, exchanged with STS server. What does a.XDSAuthenticator.Success.Load() ensure?

shankgan:
LGTM. Not sure if the deleted certs were used in any other tests, but seeing as all tests pass we should be okay. thanks for cleanup

shankgan (approved):

stevenctl (approved):

PR #37872, opened 2022-03-10
update autoscaling version to v2beta2

richardwxn:
need to wait for  #https://github.com/istio/api/pull/2273 to get in first

address #32005


manifests/charts/gateways/istio-egress/templates/autoscale.yaml
```diff
@@ -1,6 +1,6 @@
 {{ $gateway := index .Values "gateways" "istio-egressgateway" }}
 {{- if and $gateway.autoscaleEnabled $gateway.autoscaleMin $gateway.autoscaleMax }}
-apiVersion: autoscaling/v2beta1
+apiVersion: autoscaling/v2beta2
```
howardjohn, line 3: v2beta2 just pushes us back 1 release to 1.26. So we may need to have support for v2beta2 and v2 unfortunately... Or just v2beta1 and v2 possibly. See https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v126

howardjohn: Which can just be a simple `if` in the api version I think - the schema is probably the same

howardjohn (commented):
Nice! what happens if someone has an overlay?

manifests/charts/gateways/istio-egress/templates/autoscale.yaml
richardwxn: yep I think the schema is the same between v2beta2 and v2

howardjohn:
@richardwxn the failures look like they are real

richardwxn:
> @richardwxn the failures look like they are real

Yep I think we need to check in the istio/api changes first, otherwise it cannot recognize the updated schema, which is the reason why the tests are failing

richardwxn:
> Nice! what happens if someone has an overlay?

for helm it should just work since the original values. settings just mapped to a different k8s field. 

A bit trickier for k8s overlay cases, since some fields are just removed in the v2beta2/v2 schema, two ideas:
1. check for k8s version, if older than xxx pick the old template otherwise pick the new template
2. check for deprecated fields, if using any deprecated fields pick the old template
but either way we need to keep two copies: both v2beta1 and v2

howardjohn:
2 seems slightly better since (1) would still be a breaking change to IOP when a user upgrades their underlying K8s cluster. With 2 they just need to move before they upgrade to 1.25 which removes the old API

richardwxn:
@howardjohn PTAL?

manifests/charts/gateway/templates/hpa.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.autoscaling.enabled }}
-apiVersion: autoscaling/v2beta2
+apiVersion: autoscaling/v2
```
howardjohn, line 2: This makes the minimum k8s version 1.23, which is way to high for now. We will need to support multiple versions concurrently and detect which to use

manifests/charts/gateways/istio-egress/templates/autoscale.yaml
```diff
@@ -22,6 +23,34 @@ spec:
...
 ---
+{{- else }}
+apiVersion: autoscaling/v2beta1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ $gateway.name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+  {{ $gateway.labels | toYaml | indent 4 }}
+release: {{ .Release.Name }}
```
howardjohn, line 38: indent is off.

Since the schema is identical can we just template the `apiVersion` part?

manifests/charts/gateways/istio-egress/values.yaml
```diff
@@ -279,6 +279,9 @@ global:
     # Setting this port to a non-zero value enables STS server.
     servicePort: 0
 
+  # whether to use autoscaling/v2 template for HPA settings
+  autoscalingv2API: true
```
howardjohn, line 283: Is a user expected to configure this? It would be nice to auto-detect based on their Kubernetes version.

See manifests/charts/gateway/templates/deployment.yaml `ip_unprivileged_port_start` for an example doing this

operator/pkg/helmreconciler/prune.go
```diff
@@ -43,7 +43,7 @@ import (
 var (
 	// NamespacedResources orders non cluster scope resources types which should be deleted, first to last
 	NamespacedResources = []schema.GroupVersionKind{
-		{Group: "autoscaling", Version: "v2beta1", Kind: name.HPAStr},
+		{Group: "autoscaling", Version: "v2beta2", Kind: name.HPAStr},
```
howardjohn, line 46: Will this break in 1.26? It seems confusing we use `v2` in some places and `v2beta2` in others

operator/pkg/translate/translate.go
```diff
@@ -211,19 +218,7 @@ func (t *Translator) OverlayK8sSettings(yml string, iop *v1alpha1.IstioOperatorS
 		if componentName == name.IstioBaseComponentName {
 			return "", fmt.Errorf("base component can only have k8s.overlays, not other K8s settings")
 		}
-		// for server-side apply, make sure service port has protocol defined.
-		// TODO(richardwxn): remove after https://github.com/kubernetes-sigs/structured-merge-diff/issues/130 is fixed.
 		inPathParts := strings.Split(inPath, ".")
-		if inPathParts[len(inPathParts)-1] == "Service" {
```
howardjohn, line 217: Why is this removed?

samples/addons/kiali.yaml
```diff
@@ -49,7 +49,7 @@ data:
       custom_secrets: []
       host_aliases: []
       hpa:
-        api_version: autoscaling/v2beta2
+        api_version: autoscaling/v2
```
howardjohn, line 52: should update the source (manifests/addons/values-kiali.yaml)

manifests/charts/gateways/istio-egress/templates/autoscale.yaml
richardwxn: Two templates are different, check the resource section, they are slightly off

manifests/charts/gateways/istio-egress/values.yaml
richardwxn: it is for internal usage, i.e. we check the existence of deprecated fields usages from user's IOP and set this autoscalingv2API to false, then render the v2beta1 template instead

operator/pkg/helmreconciler/prune.go
richardwxn: this should be v2, forgot to update

operator/pkg/translate/translate.go
richardwxn: This was a workaround for: https://github.com/kubernetes-sigs/structured-merge-diff/issues/130. Now it can be removed since that was fixed upstream a while ago

samples/addons/kiali.yaml
richardwxn: will do

manifests/charts/gateway/templates/hpa.yaml
howardjohn: I think this is still a concern?

richardwxn: You might be looking at a older revision? I actually updated it already, can you check again? @howardjohn

howardjohn: It seems to be the latest version - there are a lot of templates

douglas-reid (approved):
e&t owned changes seem fine. approving those.

manifests/charts/gateway/templates/hpa.yaml
richardwxn: ah I missed the helm specific one, updated

manifests/addons/values-kiali.yaml
```diff
@@ -7,5 +7,7 @@ deployment:
   accessible_namespaces:
   - '**'
   ingress_enabled: false
+  hpa:
+    api_version: autoscaling/v2
```
howardjohn, line 11: This one is a bit trickier since it ends up as plain YAML.. but this will break users. Maybe we should be more conservative for now and use v2beta2 for a bit here?

manifests/charts/gateway/templates/hpa.yaml
```diff
@@ -1,5 +1,6 @@
 {{- if .Values.autoscaling.enabled }}
-apiVersion: autoscaling/v2beta2
+{{- if (semverCompare ">=1.23-0" .Capabilities.KubeVersion.GitVersion)}}
```
howardjohn, line 2: nit: since the only change is API version should we just have the `if` around the `apiVersion`?

howardjohn (commented):
I couldn't test helm due to some syntax errors. 

```
$ go run ./istioctl/cmd/istioctl install -y -d manifests
✔ Istio core installed
- Processing resources for Istiod.                                                                                                                                                    2022-04-20T22:33:34.877626Z     error   installer       failed to update resource with server-side apply for obj HorizontalPodAutoscaler/istio-system/istiod: no matches for kind "HorizontalPodAutoscaler" in version "autoscaling/v2"
✘ Istiod encountered an error: failed to update resource with server-side apply for obj HorizontalPodAutoscaler/istio-system/istiod: no matches for kind "HorizontalPodAutoscaler" in version "autoscaling/v2"
- Processing resources for Ingress gateways.                                                                                                                                          2022-04-20T22:33:35.704757Z     error   installer       failed to update resource with server-side apply for obj HorizontalPodAutoscaler/istio-system/istio-ingressgateway: no matches for kind "HorizontalPodAutoscaler" in version "autoscaling/v2"
✘ Ingress gateways encountered an error: failed to update resource with server-side apply for obj HorizontalPodAutoscaler/istio-system/istio-ingressgateway: no matches for kind "HorizontalPodAutoscaler" in version "autoscaling/v2"
- Pruning removed resources                                                                                                                                                           2022-04-20T22:33:37.164714Z     warn    installer       retrieving resources to prune type autoscaling/v2, Kind=HorizontalPodAutoscaler: no matches for kind "HorizontalPodAutoscaler" in version "autoscaling/v2"
Error: failed to install manifests: errors occurred during operation
```

howardjohn:
Any updates? Unless I am reading this wrong this is pending updates on your side @richardwxn 

I am marking this a 1.14 release blocker cc @istio/release-managers-1-14

manifests/addons/values-kiali.yaml
richardwxn: done

manifests/charts/gateway/templates/hpa.yaml
richardwxn: done

manifests/charts/gateway/templates/hpa.yaml
```diff
@@ -1,5 +1,9 @@
-{{- if and (.Values.autoscaling.enabled) (eq .Values.kind "Deployment") }}
```
howardjohn, line 1: I think this got reverted accidentally

tests/integration/operator/switch_cr_test.go
```diff
@@ -403,7 +403,7 @@ func compareInClusterAndGeneratedResources(t framework.TestContext, cs cluster.C
 				_, err = cs.PolicyV1beta1().PodDisruptionBudgets(ns).Get(context.TODO(), name,
 					kubeApiMeta.GetOptions{})
```
howardjohn, line 404: (comment location is random)

I am seeing warning on install with 1.19:
```
warn    installer       retrieving resources to prune type autoscaling/v2, Kind=HorizontalPodAutoscaler: no matches for kind "HorizontalPodAutoscaler" in version "autoscaling/v2"
```

manifests/charts/gateways/istio-egress/templates/autoscale.yaml
```diff
@@ -1,6 +1,7 @@
 {{ $gateway := index .Values "gateways" "istio-egressgateway" }}
 {{- if and $gateway.autoscaleEnabled $gateway.autoscaleMin $gateway.autoscaleMax }}
-apiVersion: autoscaling/v2beta1
+{{- if and .Values.global.autoscalingv2API (semverCompare ">=1.23-0" .Capabilities.KubeVersion.GitVersion)}}
+apiVersion: autoscaling/v2
```
howardjohn, line 4: I think we still need the v2beta2 option?

* v2beta1: using legacy IOP config. Never used for helm (unless they touch the internal autoscalingv2API flag, but they shouldn't
* v2beta2: anyone else on <1.23
* v2: anyone else >=1.23

howardjohn:
Think this is close, nice work :+1:

manifests/charts/gateway/templates/hpa.yaml
richardwxn: added back

tests/integration/operator/switch_cr_test.go
richardwxn: it is expected on pre 1.23 versions, I thought of updating the pruning logic to check for versions as well but may prefer to do it in a follow up PR instead

manifests/charts/gateways/istio-egress/templates/autoscale.yaml
richardwxn: My original thought was to combine case1 and 2 together, so only if  >=1.23 and not using legacy IOP config would be v2, others are v2beta1. But what you suggest works too, I don't have strong preference

tests/integration/operator/switch_cr_test.go
howardjohn: I don't think we want this warning to trigger on all installs of <1.23. Most users will probably think something is broken

operator/pkg/helmreconciler/prune.go
```diff
@@ -43,6 +43,8 @@ import (
 var (
 	// NamespacedResources orders non cluster scope resources types which should be deleted, first to last
 	NamespacedResources = []schema.GroupVersionKind{
+		{Group: "autoscaling", Version: "v2", Kind: name.HPAStr},
```
howardjohn, line 46: instead of 2, can we have a single one and make NamespacedResources not a constant but rather function of k8s version?

As is we get that confusing warning and I think 2x the amount of API server calls.

richardwxn: done

tests/integration/operator/switch_cr_test.go
richardwxn: done

operator/pkg/helmreconciler/prune.go
```diff
@@ -84,10 +90,20 @@ var (
 		schema.GroupVersionKind{Group: "admissionregistration.k8s.io", Version: "v1", Kind: name.ValidatingWebhookConfigurationStr},
 		schema.GroupVersionKind{Group: "apiextensions.k8s.io", Version: "v1", Kind: name.CRDStr},
 	)
+	autoScalingV2GVK = schema.GroupVersionKind{Group: "autoscaling", Version: "v2", Kind: name.HPAStr}
 )
 
+// modifyPruneResourcesBasedOnK8SVersion modifies the pruning resources based on the k8s version
+func (h *HelmReconciler) modifyPruneResourcesBasedOnK8SVersion() {
```
howardjohn, line 97: Is it safe to mutate global variables in operator context? Not super familiar with its lifecycle

operator/pkg/helmreconciler/prune.go
```diff
@@ -193,6 +209,7 @@ func (h *HelmReconciler) GetPrunedResources(revision string, includeClusterResou
 		labels[IstioComponentLabelStr] = componentName
 	}
 	selector := klabels.Set(labels).AsSelectorPreValidated()
+	h.modifyPruneResourcesBasedOnK8SVersion()
```
howardjohn, line 212: Maybe could be `gvkList := h.GetResourcesByVersion()` instead of mutating globals?

operator/pkg/translate/translate.go
howardjohn: This is already a complex and risky PR, can we not couple it with other changes?

I don't think this is safe to be removed, as it requires the api-server to be updated so we will break old versions?

howardjohn (approved):

howardjohn:
Nice! Not sure if I mentioned this but I think we will also need PDB (v1 introduced in 1.21, v1beta1 removed in 1.25). That should be simple though since the schema is the same, just need the same conditionals on the apiVersion based on what we have here.

Thanks for the great work here

richardwxn:
Yep I will do similar work for the PDB version bump in a follow up PR

manifests/charts/gateways/istio-egress/values.yaml
Monkeyanator: Can we add a comment to the Helm description saying it's for internal usage? Users look to our Helm values as an API

manifests/charts/istiod-remote/values.yaml
```diff
@@ -450,6 +450,8 @@ global:
   # The default value is "" and when caName="", the CA will be configured by other
   # mechanisms (e.g., environmental variable CA_PROVIDER).
   caName: ""
+  # whether to use autoscaling/v2 template for HPA settings
+  autoscalingv2API: true
```
Monkeyanator, line 454: Do we need this in the istiod-remote chart values?

operator/pkg/helmreconciler/prune.go
```diff
@@ -84,10 +90,20 @@ var (
...
 		schema.GroupVersionKind{Group: "apiextensions.k8s.io", Version: "v1", Kind: name.CRDStr},
 	)
+	autoScalingV2GVK = schema.GroupVersionKind{Group: "autoscaling", Version: "v2", Kind: name.HPAStr}
 )
 
+// modifyPruneResourcesBasedOnK8SVersion modifies the pruning resources based on the k8s version
+func (h *HelmReconciler) modifyPruneResourcesBasedOnK8SVersion() {
+	// autoscaling v2 API is available on >=1.23
+	if kube.IsAtLeastVersion(h.kubeClient, autoscalingV2MinK8SVersion) {
+		NamespacedResources = append(NamespacedResources, autoScalingV2GVK)
```
Monkeyanator, line 100: Couldn't we end up appending this multiple times for in-cluster operator?

operator/pkg/helmreconciler/prune.go
Monkeyanator: +1, think we'd have to change this in a few places though?

manifests/charts/istiod-remote/values.yaml
richardwxn: Added comments. Not needed for remote, removed

operator/pkg/helmreconciler/prune.go
richardwxn: there are only two places for calling modifyPruneResourcesBasedOnK8SVersion(), one is during install, and the other is during uninstall, so they are used in different context and should be safe. But I can change it to avoid confusion

operator/pkg/helmreconciler/prune.go
richardwxn: there are only two places for calling modifyPruneResourcesBasedOnK8SVersion(), one is during install, and the other is during uninstall, so they are used in different context and should be safe. But I can change it to avoid confusion

operator/pkg/helmreconciler/prune.go
richardwxn: updated per above comment

lei-tang (approved):

Monkeyanator (approved):

PR #37887, opened 2022-03-12
using k8s rand string and cleanup unused utils

Chen-Xintong:


howardjohn (changes requested):
Can you describe the benefit of this? As is I prefer things the way they are.

Chen-Xintong:
> Can you describe the benefit of this? As is I prefer things the way they are.

I would like to unify the random string method in Istio, since I found that some parts of the code have adopted the standard library of k8s, so we don't have to recreate the function with the same functionality to simplifying the code.

douglas-reid (approved):

hzxuzhonghu (approved):

PR #37891, opened 2022-03-13
[WIP][experimental] UDP support at the gateways

su225:
Part of https://github.com/istio/istio/issues/1430

This PR adds a feature to support routing UDP traffic at the gateway. The VirtualService API is also enhanced with a new `udp` section. Here is the demo of it. All files used are available here - https://github.com/su225/istio-udp.

For full configs - see https://github.com/su225/istio-udp/blob/main/istio-configs.yaml
The corresponding API branch in my fork - https://github.com/su225/api/tree/udp-api (Commit: https://github.com/su225/api/commit/e25f95a6b25b5d95f26745f94e819e24e0095582)

Gateway 
```yaml
kind: Gateway
metadata:
  name: udp-echo-gateway
  namespace: istio-system
spec:
  selector:
    app: istio-ingressgateway
  servers:
  - port:
      number: 5000
      name: udp-echo
      protocol: UDP
    hosts:
    - echo.com
  - port:
      number: 5000
      name: tcp-echo
      protocol: TCP
    hosts:
    - echo.com
```

VirtualService
```yaml
kind: VirtualService
metadata:
  name: udp-routes
  namespace: istio-system
spec:
  hosts:
  - echo.com
  gateways:
  - udp-echo-gateway
  exportTo:
  - "*"
  tcp:
  - route:
    - destination:
        host: echo.udp-echo.svc.cluster.local
        subset: v1
        port:
          number: 2020
      weight: 80
    - destination:
        host: echo.udp-echo.svc.cluster.local
        subset: v2
        port:
          number: 2020
      weight: 20
  udp:
  - match:
    - port: 5000
    route:
    - destination:
        host: echo.udp-echo.svc.cluster.local
        port:
          number: 2020
```
Now, peeking into gateway config
```
$ istioctl proxy-config listeners istio-ingressgateway-768bd4c6d8-t5m6v.istio-system --port 5000 -o json
[
    {
        "name": "udp_0.0.0.0_5000",
        "address": {
            "socketAddress": {
                "protocol": "UDP",
                "address": "0.0.0.0",
                "portValue": 5000
            }
        },
        "listenerFilters": [
            {
                "name": "envoy.filters.udp_listener.udp_proxy",
                "typedConfig": {
                    "@type": "type.googleapis.com/envoy.extensions.filters.udp.udp_proxy.v3.UdpProxyConfig",
                    "statPrefix": "outbound|2020|v1|echo.udp-echo.svc.cluster.local",
                    "cluster": "outbound|2020|v1|echo.udp-echo.svc.cluster.local",
                    "upstreamSocketConfig": {
                        "preferGro": true
                    }
                }
            }
        ],
        "trafficDirection": "OUTBOUND",
        "udpListenerConfig": {
            "downstreamSocketConfig": {}
        },
        "enableReusePort": true,

// lots of other stuffs here
```

Let us route some traffic now
```
$ nc -v -u 172.18.210.1 5000
Ncat: Version 7.80 ( https://nmap.org/ncat )
Ncat: Connected to 172.18.210.1:5000.
hello
v1:b'hello\n'world
v1:b'world\n'[from istio] hello udp  
v1:b'[from istio] hello udp\n'Ncat: 35 bytes sent, 56 bytes received in 23.34 seconds.

$ nc -v -u 172.18.210.1 5000
Ncat: Version 7.80 ( https://nmap.org/ncat )
Ncat: Connected to 172.18.210.1:5000.
hello
v2:b'hello\n'hello
v2:b'hello\n'hello
v2:b'hello\n'hello
v2:b'hello\n'Ncat: 24 bytes sent, 52 bytes received in 8.82 seconds.
```
And it works. Notice `v1` and `v2` prefixes here - per-session stickiness and load balancing can be observed here...The `v1:b'<message>` is the reply from the server. 

From server logs
```
$ kubectl -n udp-echo logs udp-echo-v1-5cd457ddcd-dc4j2 -c udp-echo 
listening on 0.0.0.0:2000
received from ('10.244.0.42', 59898): b'hello\n'
received from ('10.244.0.42', 59898): b'world\n'
received from ('10.244.0.42', 59898): b'[from istio] hello udp\n'

$ kubectl -n udp-echo logs udp-echo-v2-6b68f8d9cd-z6fbw -c udp-echo 
listening on 0.0.0.0:2000
received from ('10.244.0.42', 54366): b'hello\n'
received from ('10.244.0.42', 54366): b'hello\n'
received from ('10.244.0.42', 54366): b'hello\n'
received from ('10.244.0.42', 54366): b'hello\n'
```
The IP `10.244.0.42` is that of the ingress gateway pod
```
$ kubectl -n istio-system get pods istio-ingressgateway-768bd4c6d8-t5m6v -o wide
NAME                                    READY   STATUS    RESTARTS   AGE    IP            NODE                    NOMINATED NODE   READINESS GATES
istio-ingressgateway-768bd4c6d8-t5m6v   1/1     Running   0          103m   10.244.0.42   istio-0-control-plane   <none>           <none>
```

Notice that TCP on the same port works (case where TCP and UDP ports are the same. Istio should handle it correctly)
```
$ echo hello | nc -v 172.18.210.1 5000
Ncat: Version 7.80 ( https://nmap.org/ncat )
Ncat: Connected to 172.18.210.1:5000.
v1  hello
Ncat: 6 bytes sent, 10 bytes received in 0.03 seconds.
```

**NOTE:** Currently, the UDP support is **very limited** in Envoy and comes with several caveats. Due to the limitations, we don't support many of the service-mesh features just yet. It is just a dumb UDP proxy which forwards traffic upstream. See [Envoy's UDP proxy documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/udp/udp_proxy/v3/udp_proxy.proto).
* No support for traffic splitting/subset-based routing. Notice the difference between `tcp_proxy` and `udp_proxy` - the `udp_proxy` does not even have `weighted_clusters` field that is present in `tcp_proxy` and `http_connection_manager`. While there are workarounds like messing around [load balancing weights](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto#config-endpoint-v3-localitylbendpoints), for the moment I'm not going that way. This could be fixable although I don't know much about Envoy internals.
* When I tested, access logging does not seem to work (what does that even mean in this context?)
* No network filter chains - just a `udp_proxy` listener filter. If the UDP listener is not QUIC (hence considered connectionless), then Envoy rejects configuration with filter chains complaining of its existence.
* `istioctl` tool doesn't seem to show UDP listener (but this is fixable) - but for later.
* Load balancing options look a bit different than for TCP based ones. It might need a separate section like `udpLoadBalancerSettings` with its own consistent hashing and `is_per_packet_balance` fields along with simple modes. This way, while specifying port-level settings we can configure both TCP and UDP ports in the same item. This is not yet done. Actually, **they are two different ports** - Here 5000/TCP and 5000/UDP.
* No `AuthorizationPolicy` support. Source IP can spoofed after all! 

This PR has a lot of work to be done on testing, validation, code quality front. As of opening the PR, it was littered with debug statements, there were no unit or integration tests, validations etc. The purpose was to make it work first.

**Areas**
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -241,3 +241,5 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.13.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 )
+
+replace istio.io/api => github.com/su225/api v0.0.0-20220311171230-df8456ce5e8b
--- a/manifests/charts/base/crds/crd-all.gen.yaml
+++ b/manifests/charts/base/crds/crd-all.gen.yaml
@@ -4507,6 +4507,49 @@ spec:
                       type: array
                   type: object
                 type: array
+              udp:
+                items:
+                  properties:
+                    match:
+                      items:
+                        properties:
+                          gateways:
+                            items:
+                              type: string
+                            type: array
+                          port:
+                            type: integer
+                        type: object
+                      type: array
+                    route:
+                      items:
+                        properties:
+                          destination:
+                            properties:
+                              host:
+                                description: The name of a service from the service
+                                  registry.
+                                type: string
+                              port:
+                                description: Specifies the port on the host that is
+                                  being addressed.
+                                properties:
+                                  number:
+                                    type: integer
+                                type: object
+                              subset:
+                                description: The name of a subset within the service.
+                                type: string
+                            type: object
+                          weight:
+                            description: Weight specifies the relative proportion
+                              of traffic to be forwarded to the destination.
+                            format: int32
+                            type: integer
+                        type: object
+                      type: array
+                  type: object
+                type: array
             type: object
           status:
             type: object
@@ -5228,6 +5271,49 @@ spec:
                     route:
                       description: The destination to which the connection should
                         be forwarded to.
+                      items:
+                        properties:
+                          destination:
+                            properties:
+                              host:
+                                description: The name of a service from the service
+                                  registry.
+                                type: string
+                              port:
+                                description: Specifies the port on the host that is
+                                  being addressed.
+                                properties:
+                                  number:
+                                    type: integer
+                                type: object
+                              subset:
+                                description: The name of a subset within the service.
+                                type: string
+                            type: object
+                          weight:
+                            description: Weight specifies the relative proportion
+                              of traffic to be forwarded to the destination.
+                            format: int32
+                            type: integer
+                        type: object
+                      type: array
+                  type: object
+                type: array
+              udp:
+                items:
+                  properties:
+                    match:
+                      items:
+                        properties:
+                          gateways:
+                            items:
+                              type: string
+                            type: array
+                          port:
+                            type: integer
+                        type: object
+                      type: array
+                    route:
                       items:
                         properties:
                           destination:
--- a/pilot/pkg/model/gateway.go
+++ b/pilot/pkg/model/gateway.go
@@ -62,6 +62,10 @@ type MergedGateway struct {
 	// using TCP protocols (like HTTP1.1, H2, mysql, redis etc)
 	MergedServers map[ServerPort]*MergedServers
 
+	// MergedUDPTransportServers map from physical UDP port to servers listening
+	// on the workload port. QUIC also uses UDP ports, so beware of conflicts
+	MergedUDPTransportServers map[ServerPort]*MergedServers
+
 	// MergedQUICTransportServers map from physical port to servers listening
 	// on QUIC (like HTTP3). Currently the support is experimental and
 	// is limited to HTTP3 only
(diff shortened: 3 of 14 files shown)


su225:
I'm not sure. The service-mesh related feature-set related to UDP is too limited. 

Currently, this is just a dumb UDP proxy. I checked the Envoy API and it seems that even canarying could not be supported easily. Authorization Policy based on L3/L4 attributes like source IP:port does not make sense because of the spoofing problem. Then Envoy does not seem to support DTLS. So, no transport security either. 

Plus, on the sidecar, it is passthrough traffic, not intercepted by Envoy. So the only things you would get at the moment are some kind of client side load balancing only from gateway-to-sidecar (not sidecar-to-sidecar as it requires interception at the client sidecar), some Envoy metrics at the gateway, configuring UDP through Istio.....That's it.

PR #37894, opened 2022-03-14
Fix selecting serviceentry by wle labels

hzxuzhonghu:

Before we do not merge labels when selecting se. Not consistent with `convertWorkloadEntryToServiceInstances`

https://github.com/istio/istio/blob/2a38aca26bf6d41dbd93ff9f3089eb1918d69a6e/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go#L208-L214

### DIFF
--- a/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
+++ b/pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
@@ -159,14 +159,36 @@ func NewServiceDiscovery(
 	return s
 }
 
+// convertWorkloadEntry convert wle from Config.Spec and populate the metadata labels into it.
+func convertWorkloadEntry(cfg config.Config) *networking.WorkloadEntry {
+	wle := cfg.Spec.(*networking.WorkloadEntry)
+	if wle == nil {
+		return nil
+	}
+
+	labels := make(map[string]string, len(wle.Labels)+len(cfg.Labels))
+	for k, v := range wle.Labels {
+		labels[k] = v
+	}
+	// we will merge labels from metadata with spec, with precedence to the metadata
+	for k, v := range cfg.Labels {
+		labels[k] = v
+	}
+	// shallow copy
+	copied := *wle
+	copied.Labels = labels
+	return &copied
+}
+
 // workloadEntryHandler defines the handler for workload entries
 func (s *ServiceEntryStore) workloadEntryHandler(old, curr config.Config, event model.Event) {
 	log.Debugf("Handle event %s for workload entry %s/%s", event, curr.Namespace, curr.Name)
 	var oldWle *networking.WorkloadEntry
 	if old.Spec != nil {
-		oldWle = old.Spec.(*networking.WorkloadEntry)
+		oldWle = convertWorkloadEntry(old)
 	}
-	wle := curr.Spec.(*networking.WorkloadEntry)
+	wle := convertWorkloadEntry(curr)
+	curr.Spec = wle
 	key := configKey{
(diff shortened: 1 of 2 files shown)


hzxuzhonghu:
cc @dddddai

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -159,13 +159,32 @@ func NewServiceDiscovery(
...
+	}
+	labels := make(map[string]string, len(wle.Labels)+len(wle.Labels))
+	for k, v := range wle.Labels {
+		labels[k] = v
+	}
+	// we will merge labels from metadata with spec, with precedence to the metadata
+	for k, v := range cfg.Labels {
+		labels[k] = v
+	}
+	wle.Labels = labels
```
dddddai, line 176: There's data race [as the test shows](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/37894/unit-tests_istio/1503205381462036480), maybe we can merge labels in `getWorkloadServiceEntries` and `convertWorkloadEntryToServiceInstances` (using config instead of WLE, like `convertWorkloadEntryToWorkloadInstance`) rather than modifying the WLE directly

Directionally I think it makes sense because it unifies WLE label handling

hzxuzhonghu: I want to merge labels once originally.

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
```diff
@@ -179,7 +199,7 @@ func (s *ServiceEntryStore) workloadEntryHandler(old, curr config.Config, event
 		event = model.EventDelete
 	}
 
-	wi := s.convertWorkloadEntryToWorkloadInstance(curr, s.Cluster())
+	wi := s.convertWorkloadEntryToWorkloadInstance(curr, currLabels, s.Cluster())
```
dddddai, line 202: What about `convertWorkloadEntryToServiceInstances`?

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
dddddai: Or you may create a shallow copied WLE whose labels are merged, either is fine
```go
func populateWorkloadEntryLabels(cfg config.Config) *networking.WorkloadEntry {
	wle := cfg.Spec.(*networking.WorkloadEntry)
	if wle == nil {
		return nil
	}
	labels := make(map[string]string, len(wle.Labels)+len(wle.Labels))
	for k, v := range wle.Labels {
		labels[k] = v
	}
	// we will merge labels from metadata with spec, with precedence to the metadata
	for k, v := range cfg.Labels {
		labels[k] = v
	}
	copy := *wle
	copy.Labels = labels
	return &copy
}
```

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
hzxuzhonghu: good catch, need to fix too

pilot/pkg/serviceregistry/serviceentry/servicediscovery.go
hzxuzhonghu: good suggestion, this will prevent updating anywhere

howardjohn (approved):

howardjohn:
cc @ramaraochavali

hzxuzhonghu:
rebased

PR #37904, opened 2022-03-14
add externalIPs in helm chart gateway service

### LINKED ISSUE #37902: Add spec.externalIPs for service in gateway helm charts for

According to https://istio.io/latest/docs/ops/diagnostic-tools/multicluster/ spec.externalIPs is required for environments which don't support kubernetes LoadBalancer services.

```
In this case, it may be necessary to customize the spec.externalIPs section of the Service to manually give the Gateway an IP reachable from outside the cluster.
```

The gateway helm chart doesn't support this field for the gateway service.

[ ] Docs
[x] Installation
[ ] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

tafkam:

Adds externalIPs to gateway service in helm chart.
Fixes #37902

Helm Chart default behaviour didn't change

I assumed there are no usecases to set externalIPs spec together with service type LoadBalancer.

### DIFF
--- a/manifests/charts/gateway/templates/service.yaml
+++ b/manifests/charts/gateway/templates/service.yaml
@@ -40,6 +40,11 @@ spec:
 {{- else }}
 {{ .Values.service.ports | toYaml | indent 4 }}
 {{- end }}
+{{- if .Values.service.externalIPs }}
+  externalIPs: {{- range .Values.service.externalIPs }}
+    - {{.}}
+  {{- end }}
+{{- end }}
   selector:
     {{- include "gateway.selectorLabels" . | nindent 4 }}
 {{- end }}
--- a/manifests/charts/gateway/values.yaml
+++ b/manifests/charts/gateway/values.yaml
@@ -52,6 +52,7 @@ service:
   loadBalancerIP: ""
   loadBalancerSourceRanges: []
   externalTrafficPolicy: ""
+  externalIPs: []
 
 resources:
   requests:


manifests/charts/gateway/templates/service.yaml
```diff
@@ -39,6 +39,11 @@ spec:
     targetPort: 15017
 {{- else }}
 {{ .Values.service.ports | toYaml | indent 4 }}
+{{- end }}
+{{- if and (.Values.service.externalIPs) (ne .Values.service.type "LoadBalancer") }}
```
howardjohn, line 43: For the `ne .Values.service.type "LoadBalancer"` part - what happens if both are set?  Does k8s reject, or is it just not standard?

Since it seems a bit confusing for a user to explicitly set `externalIPs` but you do not get it?

tafkam: If both are set the helm chart won't render the external IPs. According to https://istio.io/latest/docs/ops/diagnostic-tools/multicluster/ External IPs are required for istio, if you are running in an environment not support native kubernetes loadbalancer type services.
And if your environment doesn't support loadbalancer services, you shouldn't define your gateway service as type loadbalancer, but probably as nodeport service. But this is probably the wrong implementation for that nudge :)

That said I can't imagine a scenario using k8s LB service together with defined externalIPs, but maybe that's just me.
I have no idea if externalIPs work with or break LoadBalancer type services or the k8s api. 
At the end of the day it's probably better to remove the service type condition, if you want.

stevenctl:
LGTM

manifests/charts/gateway/templates/service.yaml
```diff
@@ -39,6 +39,11 @@ spec:
     targetPort: 15017
 {{- else }}
 {{ .Values.service.ports | toYaml | indent 4 }}
+{{- end }}
+{{- if .Values.service.externalIPs }}
+  externalIPs: {{- range .Values.service.externalIPs }}
+    - {{.}}
+  {{- end }}
```
jacob-delgado, line 46: is this an extra end?

tafkam: l46 ends the range iteration

jacob-delgado (approved):

morvencao (approved):

PR #37914, opened 2022-03-14
[TF] Refactoring authz tests

nmittler:
This is a complete rewrite of the authz tests to use a common echo deployment and the echotest framework.

The new tests deploy smaller configs and test more combinations of deployments and ports/protocols.


pkg/test/framework/components/echo/common/deployment/apps.go
```diff
@@ -0,0 +1,440 @@
...
+	"istio.io/istio/pkg/test/framework/components/echo/deployment"
+	"istio.io/istio/pkg/test/framework/components/echo/match"
+	"istio.io/istio/pkg/test/framework/components/istio"
+	"istio.io/istio/pkg/test/framework/components/istio/ingress"
+	"istio.io/istio/pkg/test/framework/components/namespace"
+	"istio.io/istio/pkg/test/framework/resource"
+	"istio.io/istio/pkg/test/util/tmpl"
+)
+
+type Namespace struct {
```
howardjohn, line 38: `A` within `Namespace` feels a bit odd to me?

howardjohn: Oh I guess it makes sense now

pkg/test/framework/components/echo/common/deployment/apps.go
```diff
@@ -0,0 +1,440 @@
...
+		}
+		res = append(res, p)
+	}
+	return res
+}
+
+func SetupApps(t resource.Context, i istio.Instance, apps *EchoDeployments) error {
+	var err error
+	apps.Ns1.Namespace, err = namespace.New(t, namespace.Config{
+		Prefix: "echo1",
```
howardjohn, line 233: really nit but can we keep this as `echo`? I have `echo` in my muscle memory...

pkg/test/framework/components/echo/common/deployment/apps.go
```diff
@@ -0,0 +1,440 @@
...
+		return err
+	}
+
+	// Load values from the deployed echo instances.
+	return apps.loadValues(t, echos)
+}
+
+// TODO: should t.Settings().Skip(echo.Delta) do all of this?
+func skipDeltaXDS(t resource.Context) bool {
+	return t.Settings().Skip(echo.Delta) || !t.Settings().Revisions.AtLeast("1.11")
```
howardjohn, line 279:
```suggestion
	return t.Settings().Skip(echo.Delta) || !t.Settings().Revisions.AtLeast("1.12")
```

changed in master

tests/integration/security/authz/authz_test.go
```diff
@@ -0,0 +1,338 @@
...
+)
+
+func TestAuthz_Principal(t *testing.T) {
+	framework.NewTest(t).
+		Features("security.authorization.mtls-local",
+			"security.authorization.grpc-protocol",
+			"security.authorization.tcp").
+		Run(func(t framework.TestContext) {
+			// Allow ns1.A to call every service in ns1.
+			allowed := apps.Ns1.A
```
howardjohn, line 48: bikeshedding a lot here, but should we embedd Ns1? 99% of tests shouldn't really care about namespace so they can then keep doing `apps.A`

tests/integration/security/authz/authz_test.go
```diff
@@ -0,0 +1,338 @@
...
+				ToNamespace(apps.Ns1.Namespace).
+				Build(t).
+				From(config.FileOrFail(t, "testdata/v1beta1/workload-ns.yaml.tmpl")).
+				ToNamespace(apps.Ns2.Namespace).
+				Build(t).
+
+				// Allow /policy-istio-system-<svc> for all services in all namespaces. Just using ns1 to avoid
+				// creating duplicate resources.
+				From(config.FileOrFail(t, "testdata/v1beta1/workload-system-ns.yaml.tmpl")).
+				ToSystemNamespace(t).
```
howardjohn, line 203: I cannot understand what this means. In the template they use `{{ .To.Config.Service }}`. How does that relate to SystemNamespace?

howardjohn: Oh Its where its applied to. I feel like we might have gone a little too far on the chaining here - I was a big fan of the current state in master.

I don't see the benefits of this abstraction. It seems optimized to apply a bunch of config all at once in parallel, but I don't think that is important? Perhaps I am missing context on the motivation

pkg/test/framework/components/echo/common/deployment/apps.go
nmittler: We need the prefix for each namespace to be unique for test names. I guess it could be `echo` and `echo1`, but the variables are Ns1 and Ns2 ... a bit odd.

tests/integration/security/authz/authz_test.go
nmittler: Yeah, some of the auth tests push configs that take a long time to complete. Really want to parallelize  it if possible.

tests/integration/security/authz/authz_test.go
nmittler: I don't disagree ... definitely seems like number of namespaces could be a config option when we build it. Let me think about it.

tests/integration/security/authz/authz_test.go
howardjohn: How is the `kubectl apply` slow? I would expect this to take milliseconds, its just an HTTP request?

nmittler: OK, so I dug in a bit and the problem with the security tests is that it ends up stacking a small delay (~ few seconds on GKE) across a lot of configs, so it can end up taking ~ 1 minute to push config. I've got a general fix here: https://github.com/istio/istio/pull/38024

pkg/test/framework/components/echo/common/deployment/apps.go
nmittler: Addressed by https://github.com/istio/istio/pull/38202

stevenctl (approved):
LGTM.. weird that the current failures are all cross-network-primary -> http2/grpc

stevenctl:
Looks like I already approved an earlier iteration, but I think this is awesome. 
My only comment is that we should update the wiki: https://github.com/istio/istio/wiki/Writing-Good-Integration-Tests.

nmittler:
@yangminzhu can you take a look at these changes? In general, I think this should greatly increase test coverage.

yangminzhu:
Adding @liwenhao0810  to also take a review of the PR, thanks.

nmittler:
@howardjohn @aryan16 PTAL ... this should be ready to go now.

howardjohn (approved):

tests/integration/security/authz/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"istio.io/istio/pkg/test/framework/resource"
+)
+
+var (
+	apps             deployment.TwoNamespaceView
+	authzServer      authz.Server
+	localAuthzServer authz.Server
+)
+
+func TestMain(m *testing.M) {
```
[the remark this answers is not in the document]
nmittler: I'm using a new suite temporarily because it adds an authz server and the common echo deployment. The plan is to follow this PR with changes to the rest of the top-level tests so that we can go back to a single suite.

tests/integration/security/authz/authz_test.go
```diff
@@ -0,0 +1,1547 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package security
+
+import (
```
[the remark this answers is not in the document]
aryan16: In most of the scenarios you're using the conditions with `notValues`. Shouldn't we try to check all the attributes which is there in authz policy (eg. https://istio.io/latest/docs/reference/config/security/authorization-policy/#Operation). I know under the hood it might use the same stuff (like conditions) but if anything gets changed with any of these attributes then tests based on this will help us analyze that in future.

aryan16 (approved):

tests/integration/security/authz/authz_test.go
```diff
@@ -0,0 +1,1800 @@
...
+	"istio.io/istio/pkg/test/framework/components/echo/match"
+	"istio.io/istio/pkg/test/framework/components/istio"
+	"istio.io/istio/pkg/test/framework/components/istio/ingress"
+	"istio.io/istio/pkg/test/framework/components/namespace"
+	"istio.io/istio/tests/common/jwt"
+)
+
+const (
+	// TODO(nmittler): Make this a baked-in default for echotest framework.
+	callMultiplier = 5
```
stevenctl, line 44: we have `util.CallsPerCluster`

stevenctl:
LGTM

PR #37916, opened 2022-03-14
Remove PILOT_ENABLE_FLOW_CONTROL flag

howardjohn:
This was added like 2 years ago and has no known usage and adds a lot of complexity to our core codepaths. I think its best to remove; if we want to re-vist we can always add it back


stevenctl (approved):
code LGTM, deferring to others on the intent

lei-tang (approved):

hzxuzhonghu:
It is to fix https://github.com/istio/istio/issues/25685

@[ZhuangYuZY](https://github.com/ZhuangYuZY) From IBM could use it.

howardjohn:
I checked that Knative, who the fix was originally developed for, does not actually use this flag. I am pretty sure it doesn't actually work. Can I remove the hold?

hzxuzhonghu:
LGTM

PR #37926, opened 2022-03-14
experimental delta: use cache key to decide if we should build

howardjohn:
Showing a sketch of what I recommended as an alternative to https://github.com/istio/istio/pull/37465 in the design doc.

How it works:

On each CDS generation (and this could apply to everything - but for now just CDS):
 * First we record all clusters already present in the proxy we are pushing to, call this `KnownClusters`. PR includes a few bug-fixes to make that possible.
 * We go through the start standard building phase, where we generate the clusterName we are about to build and the cache key. This is slightly complex to handle subsets but not a big issue.
   * For each key, we check the push info. If we depend on the push, we know the resource has changed. Otherwise, the resource hasn't changed, we mark this cluster as eligible to skip. This dependency check is the same check used in the XDS cache, but it does not depend on the _contents_ of the cache.
   * If we do skip it, remove the cluster from `KnownClusters`
   * This is not yet complete: we could still miss updates if they cluster _previously_ depended on the config. For example, removing a destination rule. To handle this, we amend WatchedResources to include a mapping of ResourceName -> CacheKey. This allows us to check if the cluster currently _or_ previously depending on the config, handling this case properly. This is similar to the PreviousSidecarScope stuff we have
* For all clusters we did generate, remove them from `KnownClusters` as well
* At this point, `KnownClusters` consists only of clusters we had previously and do not have now -- that is, the ones deleted. Return this as "deleted resource". Note: this is very similar to how the SotW -> Delta adapter works, but the reason we need to do it inline in the cluster code is we have additional context on "skipped" vs "deleted".

I have manually tested this with UNSAFE_PILOT_ENABLE_DELTA_TEST=true which passed TestTraffic.

Note: this is proof of concept currently. It will need a lot of work to be production ready

cc @ramaraochavali @hzxuzhonghu @adiprerepa

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -326,6 +326,8 @@ type WatchedResource struct {
 	// For Delta Xds, all resources of the TypeUrl that a client has subscribed to.
 	ResourceNames []string
 
+	CacheKeys map[string]XdsCacheEntry
+
 	// VersionSent is the version of the resource included in the last sent response.
 	// It corresponds to the [Cluster/Route/Listener]VersionSent in the XDS package.
 	VersionSent string
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -37,10 +37,12 @@ import (
 	"istio.io/istio/pilot/pkg/networking/core/v1alpha3/loadbalancer"
 	"istio.io/istio/pilot/pkg/networking/util"
 	"istio.io/istio/pilot/pkg/serviceregistry/provider"
+	v3 "istio.io/istio/pilot/pkg/xds/v3"
 	"istio.io/istio/pkg/config/host"
 	"istio.io/istio/pkg/config/protocol"
 	"istio.io/istio/pkg/config/schema/gvk"
 	"istio.io/istio/pkg/util/sets"
+	"istio.io/pkg/log"
 )
 
 // deltaConfigTypes are used to detect changes and trigger delta calculations. When config updates has ONLY entries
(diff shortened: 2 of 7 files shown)


pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -301,6 +262,25 @@ func (configgen *ConfigGeneratorImpl) buildOutboundClusters(cb *ClusterBuilder,
 	return resources, cacheStats{hits: hit, miss: miss}
 }
 
+func skipForDelta(key model.XdsCacheEntry, updated map[model.ConfigKey]struct{}) bool {
```
hzxuzhonghu, line 265: This is the main difference from #37465 

But the updated configs may include deleted configs, not sure how you will handle this

howardjohn: Yeah that will be the tricky part. Maybe if this functionsl returns `true` but the cluster generator doesn't build a cluster, that is a "delete". Would have to test it

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -301,6 +262,25 @@ func (configgen *ConfigGeneratorImpl) buildOutboundClusters(cb *ClusterBuilder,
...
+		for _, dt := range key.DependentTypes() {
+			if dt == cfg.Kind {
+				return false
+			}
+		}
+		for _, dc := range key.DependentConfigs() {
+			if dc == cfg {
+				return false
+			}
+		}
```
hzxuzhonghu, line 279: Compared with, this is stateful, while that one can be seen as stateless. 

```
 func (cb *ClusterBuilder) getAllCachedSubsetClusters(clusterKey clusterCache) ([]*discovery.Resource, bool) {
	if !features.EnableCDSCaching {
		return nil, false
	}
	destinationRule := CastDestinationRule(clusterKey.destinationRule)
	res := make([]*discovery.Resource, 0, 1+len(destinationRule.GetSubsets()))
	cachedCluster, f := cb.cache.Get(&clusterKey)
	allFound := f
	res = append(res, cachedCluster)
	dir, _, host, port := model.ParseSubsetKey(clusterKey.clusterName)
	for _, ss := range destinationRule.GetSubsets() {
		clusterKey.clusterName = model.BuildSubsetKey(dir, ss.Name, host, port)
		cachedCluster, f := cb.cache.Get(&clusterKey)
		if !f {
			allFound = false
		}
		res = append(res, cachedCluster)
	}
	return res, allFound
}
```

hzxuzhonghu: another point is that, the function looks not right. 
Look at `getAllCachedSubsetClusters`, the cluster key can be modified 
```
		clusterKey.clusterName = model.BuildSubsetKey(dir, ss.Name, host, port)

```

hzxuzhonghu: Also, i am not sure if this function is more efficient than `getAllCachedSubsetClusters`

howardjohn:
It definitely has the potential to be. For example if we rely on the cache,
the first unlucky proxy we push to will have 100% cache misses and we will
send all unneeded clusters. It's also inconsistent which makes scaling and
performance tricky

In this approach it's 100% deterministic and doesn't rely on global state
(cache) which make unexpectedly impact the clusters we send.

hzxuzhonghu:
>  For example if we rely on the cache,
the first unlucky proxy we push to will have 100% cache misses and we will
send all unneeded clusters.

this is not the advantage of this solution either. If  100% cache miss. 
1. Say meshconfig updated, we clear all the cache.  
The skipDelta will be false here, we still need to build all clusters.

Can you give a specific example, when this can have an advantage

howardjohn:
Yes, specific examples are while we are filling the cache we will send
redundant clusters, and if cache is overfilled we will give start sending
excess clusters as well. Additionally when this happens it's inconsistent.

In both approaches an mesh config update would trigger pushing all clusters
I thought?

hzxuzhonghu:
>while we are filling the cache we will send
redundant clusters, and if cache is overfilled we will give start sending
excess clusters as well.

Still not clear to me.

hzxuzhonghu:
While we are filling, it means these resources are changed. Do you mean this case:  at t0 (a service is added), and at t1 (meshconfig updated), and we cleared all cache.  Then when we push for t0, we need to push all.

In this case, we can push redundant clusters.  This is decided by asynchronous push, normally can not happen in most of the time.

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -77,71 +78,27 @@ func (configgen *ConfigGeneratorImpl) BuildClusters(proxy *model.Proxy, req *mod
...
+	var services []*model.Service
+	if features.FilterGatewayClusterConfig && proxy.Type == model.Router {
+		services = req.Push.GatewayServices(proxy)
+	} else {
+		services = proxy.SidecarScope.Services()
 	}
-	clusters, log := configgen.buildClusters(proxy, updates, services)
-	return clusters, deletedClusters, log, true
+	res, log := configgen.buildClusters(proxy, req, services, true)
+	// TODO handle deletes
```
adiprerepa, line 95: how would we handle deletes?

hzxuzhonghu: No ideas atm

pilot/pkg/model/context.go
```diff
@@ -326,6 +326,8 @@ type WatchedResource struct {
 	// For Delta Xds, all resources of the TypeUrl that a client has subscribed to.
 	ResourceNames []string
 
+	CacheKeys map[string]XdsCacheEntry
```
hzxuzhonghu, line 329: I think this could add great memory cost to istiod.  Last year we did a bunch of large scale performance test, and figured out that even the watched resource names of eds cost too much memory from pprof. The reason is even the eds request resources may have names ["cluster1", "cluster2"] same value as istiod inner generated one. But the string is unmarshalled from DiscoveryRequest, so in a process stack, they are not shared.  So I am not sure about this.

howardjohn: I had similar concerns actually. I will do some benchmarking to see what cost there is.

It's possible we can keep around not the full key but just enough to do the DependsOn part

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -239,8 +214,9 @@ func buildClusterKey(service *model.Service, port *model.Port, cb *ClusterBuilde
 }
 
 // buildOutboundClusters generates all outbound (including subsets) clusters for a given proxy.
-func (configgen *ConfigGeneratorImpl) buildOutboundClusters(cb *ClusterBuilder, proxy *model.Proxy, cp clusterPatcher,
-	services []*model.Service) ([]*discovery.Resource, cacheStats) {
+func (configgen *ConfigGeneratorImpl) buildOutboundClusters(cb *ClusterBuilder, proxy *model.Proxy,
+	cp clusterPatcher, services []*model.Service, updated map[model.ConfigKey]struct{}, known sets.Set, delta bool) ([]*discovery.Resource, cacheStats) {
```
hzxuzhonghu, line 218: nit: since delta can be inferred from `known`, is delta redundant

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -250,7 +226,23 @@ func (configgen *ConfigGeneratorImpl) buildOutboundClusters(cb *ClusterBuilder,
 				continue
 			}
 			clusterKey := buildClusterKey(service, port, cb, proxy, efKeys)
-			cached, allFound := cb.getAllCachedSubsetClusters(*clusterKey)
+			subsetKeys := cb.getSubsetKeys(*clusterKey)
+			known.Delete(clusterKey.clusterName)
+			skip := skipForDelta(clusterKey, updated) && skipForDelta(oldCacheKeys[clusterKey.clusterName], updated)
```
hzxuzhonghu, line 231: A little hard to understand, need some comments

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -250,7 +226,23 @@ func (configgen *ConfigGeneratorImpl) buildOutboundClusters(cb *ClusterBuilder,
...
+			for _, ss := range subsetKeys {
+				skipSubset := skipForDelta(&ss, updated) && skipForDelta(oldCacheKeys[ss.clusterName], updated)
+				known.Delete(ss.clusterName)
+				// We can only skip if we skip ALL subsets. This could probably be optimized in the future
+				skip = skip && skipSubset
```
hzxuzhonghu, line 236: same here

hzxuzhonghu (commented):
I think generally good. It adds a cache of  resource name->cache keys per proxy, so no other thread will remove this cache async.

pilot/pkg/model/context.go
hzxuzhonghu: One more concern, for eds the key is same as cds,

Issue #37724, opened 2022-03-04
Reduce Envoy CPU usage when disabling all metrics


rolandkool:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**
As part of investigating envoys CPU overhead, I performed some loadtests with various Istio features turned on/off. One of the features I disabled was metrics collection.

Our Istio setup has been configured to disable the telemetry (Envoyfilters for stats collection) component:
```
    values
      telemetry:
        v2:
          prometheus:
            enabled: false
          stackdriver:
            enabled: false
```

To disable metrics through the Telemetry API, there are two options:
1. Don't configure anything for metrics (omitting the `metrics` parameter)
2. Configure metrics and explicitly disable everything:
```
  metrics:
  - overrides:
    - disabled: true
      match:
        metric: ALL_METRICS
        mode: CLIENT_AND_SERVER
    providers:
    - name: prometheus
```

If you have, like we do, a default metrics configuration in the `istio-system` namespace that sets a proper default for all workloads, option 1 is no longer an option if you want to disable metrics for a particular workload. Option 2 is the only way to do that.

However, option 1 actually does reduce the CPU overhead of Envoy significantly compared to option 2. My guess is that this is because option 2 still configures some envoyfilter to generate the metrics only to then drop all of them, while option 1 does not configure the proxy to have any envoyfilter for generating metrics.

An improvement could be that when all metrics are explicitly disabled using option 2, that it actually removes all configuration from Envoy so no CPU cycles are wasted on generating metrics that are to be dropped anyway.

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[X] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

douglas-reid:
@howardjohn @zirain WDYT about @rolandkool 's suggestion here? With default enablement of `prometheus`, should we more efficiently handle `ALL_METRICS` + `disabled: true` by not installing the stats filter at all?

howardjohn:
[...]

PR #37928, opened 2022-03-15
not installing the stats filter when diabled all metrics

zirain:

fixed https://github.com/istio/istio/issues/37724

actually, this PR will not fix the `EnvoyFilter` part, get more info here https://github.com/istio/istio/issues/37724#issuecomment-1067464435 .

### DIFF
--- a/pilot/pkg/model/telemetry.go
+++ b/pilot/pkg/model/telemetry.go
@@ -540,11 +540,15 @@ func mergeMetrics(metrics []*tpb.Metrics, mesh *meshconfig.MeshConfig) map[strin
 					tpb.WorkloadMode_SERVER: {},
 				}
 			}
+
+			// root namespace disables all, but then enables them by namespace scoped
+			disabledAllMetricsProviders.Delete(provider)
+
 			mp := providers[provider]
 			// For each override, we normalize the configuration. The metrics list is an ordered list - latter
 			// elements have precedence. As a result, we will apply updates on top of previous entries.
 			for _, o := range m.Overrides {
-				// // if we disable all metrics, we should drop the entire filter
+				// if we disable all metrics, we should drop the entire filter
 				if isAllMetrics(o.GetMatch()) && o.Disabled.GetValue() {
 					disabledAllMetricsProviders.Insert(provider)
 					continue
--- a/pilot/pkg/model/telemetry.go
+++ b/pilot/pkg/model/telemetry.go
@@ -367,16 +367,18 @@ func (t *Telemetries) telemetryFilters(proxy *Proxy, class networking.ListenerCl
 
 	// First, take all the metrics configs and transform them into a normalized form
 	tmm := mergeMetrics(c.Metrics, t.meshConfig)
+	log.Debugf("merged metrics, proxyID: %s metrics: %+v", proxy.ID, tmm)
 	// Additionally, fetch relevant access logging configurations
 	tml, logsFilter := mergeLogs(c.Logging, t.meshConfig)
 
 	// The above result is in a nested map to deduplicate responses. This loses ordering, so we convert to
 	// a list to retain stable naming
-	m := []telemetryFilterConfig{}
 	allKeys := sets.NewSet(tml.UnsortedList()...)
 	for k := range tmm {
 		allKeys.Insert(k)
 	}
+
+	m := make([]telemetryFilterConfig, 0, len(allKeys))
 	for _, k := range allKeys.SortedList() {
 		p := t.fetchProvider(k)
 		if p == nil {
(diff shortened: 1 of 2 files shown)


zirain:
@douglas-reid PTAL

pilot/pkg/model/telemetry.go
```diff
@@ -542,6 +545,13 @@ func mergeMetrics(metrics []*tpb.Metrics, mesh *meshconfig.MeshConfig) map[strin
 			// For each override, we normalize the configuration. The metrics list is an ordered list - latter
 			// elements have precedence. As a result, we will apply updates on top of previous entries.
 			for _, o := range m.Overrides {
+				// // if we disable all metrics, we should drop the entire filter
+				if isAllMetrics(o.GetMatch()) && o.Disabled.GetValue() {
```
howardjohn, line 549: what if root namesapce disables all, but then a later config enables them?

zirain: it would depends on how `applicableTelemetries`  works, let me check that.

pilot/pkg/model/telemetry_test.go
```diff
@@ -669,6 +687,15 @@ func TestTelemetryFilters(t *testing.T) {
 			nil,
 			map[string]string{},
 		},
+		{
+			"disabled-prometheus",
+			[]config.Config{newTelemetry("istio-system", disbaledAllMetrics)},
```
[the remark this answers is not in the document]
zirain: updated

pilot/pkg/model/telemetry.go
zirain: updated

zirain:
ping @howardjohn @douglas-reid

pilot/pkg/model/telemetry_test.go
```diff
@@ -669,6 +691,29 @@ func TestTelemetryFilters(t *testing.T) {
...
+			[]config.Config{
+				newTelemetry("istio-system", disbaledAllMetrics),
+				newTelemetry("default", emptyPrometheus),
+			},
+			sidecar,
+			networking.ListenerClassSidecarOutbound,
+			networking.ListenerProtocolHTTP,
+			nil,
+			map[string]string{
+				"istio.stats": `{}`,
```
howardjohn, line 714: Prior to this PR, this would be **disabled** still. Since `disabled` is not set by the `emptyPrometheus`, it uses the parent settings (disabled: true). We need an explicit `disabled: false`

zirain: what should we do in this scenario:
```golang
newTelemetry("istio-system", emptyPrometheus),
newTelemetry("default", emptyPrometheus),
```

`it uses the parent settings` this sound a little wired.

howardjohn: that is how the merge works - unset fields pull from the parent's

zirain: sorry, I did not get what your concern about this test case. can you share more details?

howardjohn: I mean this should be disabled. If you apply this same config in Istio today you get no metrics enabled, but here we enable it

zirain: test on master branch with the following yaml:
```
kind: Telemetry
metadata:
  name: disable-metrics
  namespace: istio-system
spec:
  metrics:
    - overrides:
        - disabled: true
          match:
            metric: ALL_METRICS
            mode: CLIENT_AND_SERVER
      providers:
        - name: prometheus
---
kind: Telemetry
metadata:
  name: empty-prometheus
  namespace: default
spec:
  metrics:
    - providers:
        - name: prometheus
```
still get metrics
```
# TYPE istio_requests_total counter
istio_requests_total{response_code="200",reporter="destination",source_workload="sleep",source_workload_namespace="default",source_principal="spiffe://cluster.local/ns/default/sa/sleep",source_app="sleep",source_version="unknown",source_cluster="Kubernetes",destination_workload="httpbin",destination_workload_namespace="default",destination_principal="spiffe://cluster.local/ns/default/sa/httpbin",destination_app="httpbin",destination_version="v1",destination_service="httpbin.default.svc.cluster.local",destination_service_name="httpbin",destination_service_namespace="default",destination_cluster="Kubernetes",request_protocol="http",response_flags="-",grpc_response_status="",connection_security_policy="mutual_tls",source_canonical_service="sleep",destination_canonical_service="httpbin",source_canonical_revision="latest",destination_canonical_revision="v1"} 1
```

zirain:
> {"metrics":[{"name":"request_messages_total","drop":true},{"name":"response_messages_total","drop":true},{"name":"requests_total","drop":true},{"name":"request_duration_milliseconds","drop":true},{"name":"request_bytes","drop":true},{"name":"response_bytes","drop":true},{"name":"tcp_connections_closed_total","drop":true},{"name":"tcp_connections_opened_total","drop":true},{"name":"tcp_received_bytes_total","drop":true},{"name":"tcp_sent_bytes_total","drop":true}]}

I will verify again, did you install the `stats-*` EnvoyFilter?

howardjohn:
I removed the EnvoyFilters.

I was also able to reproduce by copying your new tests but not the code
changes

howardjohn (approved):

PR #37935, opened 2022-03-15
new docker build logic

howardjohn:
This PR refactors how our docker images are managed.

Previously, we used some magic Makefile incantations to declare build dependencies and files that should be in the docker context. This had a few issues:
* Super confusing - requires advanced make expertise to understand
* Requires build targets for all files in the dockerfile, which means we create weird targets for files that we don't do anything with
* Makes it impossible to do any sort of analysis of our images or programmatic changes. For example, building images with a non-docker system
* Weird things with moving Dockerfiles and copying files
 
Instead, we define a new simple YAML structure for all of our images, where they define inputs explicitly. This is then parsed by the builder.

This is not intended to have any user facing changes; the resulting images will be identical.


ericvn (approved):
These changes also fix up some of the changes I was making in another PR. Let's merge this and I'll simplify things in mine.

stevenctl:
LGTM! My only experience is hacking around in here and being confused. Deferring to Eric as the expert.

PR #37936, opened 2022-03-15
tf: skip VM suite when VM is disabled

howardjohn:
This test only makes sense when VMs are enabled. Currently it will fail in weird ways otherwise


tests/integration/telemetry/stackdriver/vm/main_test.go
```diff
@@ -121,6 +121,9 @@ func TestMain(m *testing.M) {
 		// https://github.com/istio/istio/issues/35923. Since IPv6 has no external connectivity, we are "not on GCP"
 		// in the sense that we cannot access the metadata server
 		Label(label.IPv4).
+		SkipIf("test requires VMs", func(ctx resource.Context) bool {
```
nmittler, line 124: Should probably include the issue in the message?

howardjohn: There is no issue - this suite is the `vm` suite, it doesn't make sense to run without VMs

PR #37941, opened 2022-03-15
[TF] Support multi-namespace in echotest

nmittler:
The current echotest tooling doesn't support multiple namespaces very well.  This changes a number of things:

- Include namespace prefix when multiple namepaces are used across source and destinations.
- Adding `RegularPod` logic as top-level property of config.
- Updating logic for selecting one regular pod to include one per namespace.
- Renaming `match.NamespacedName` to `match.SameService` and deprecating `match.Service`.
- Cleaning up naming of matchers.

### DIFF
--- a/pkg/test/framework/components/echo/config.go
+++ b/pkg/test/framework/components/echo/config.go
@@ -251,6 +251,16 @@ func (c Config) IsDelta() bool {
 	return len(c.Subsets) > 0 && c.Subsets[0].Annotations != nil && strings.Contains(c.Subsets[0].Annotations.Get(SidecarProxyConfig), "ISTIO_DELTA_XDS")
 }
 
+// IsRegularPod returns true if the echo pod is not any of the following:
+// - VM
+// - Naked
+// - Headless
+// - TProxy
+// - Multi-Subset
+func (c Config) IsRegularPod() bool {
+	return len(c.Subsets) == 1 && !c.IsVM() && !c.IsTProxy() && !c.IsNaked() && !c.IsHeadless() && !c.IsStatefulSet() && !c.IsProxylessGRPC()
+}
+
 // DeepCopy creates a clone of IstioEndpoint.
 func (c Config) DeepCopy() Config {
 	newc := c
--- a/pkg/test/framework/components/echo/match/matchers.go
+++ b/pkg/test/framework/components/echo/match/matchers.go
@@ -15,8 +15,6 @@
 package match
 
 import (
-	"strings"
-
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pkg/test/framework/components/cluster"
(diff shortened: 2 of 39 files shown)


pkg/test/framework/components/echo/config.go
```diff
@@ -251,6 +251,10 @@ func (c Config) IsDelta() bool {
 	return len(c.Subsets) > 0 && c.Subsets[0].Annotations != nil && strings.Contains(c.Subsets[0].Annotations.Get(SidecarProxyConfig), "ISTIO_DELTA_XDS")
 }
 
+func (c Config) IsRegularPod() bool {
```
douglas-reid, line 254: super nit: might be worth providing a comment on what is meant by "regular". It looks like it means "kubernetes pod holding a workload instance with an injected sidecar", but maybe it is easier to define it by what it is not ?

douglas-reid (approved):
the P&T changes look reasonable to me.

pkg/test/framework/components/echo/config.go
stevenctl: Can either reference `SingleSimplePodServiceAndAllSpecial`'s comment, or move that comment here and have that method reference this one.

tests/integration/pilot/common/routing.go
```diff
@@ -692,12 +692,14 @@ metadata:
...
-    - {{ ( index .dstSvcs 0) }}
+{{- range $idx, $svc := .dstSvcs }}
+    - {{ $svc.Name }}.{{ $svc.Namespace }}
+{{- end }}
   http:
   - route:
 {{- range $idx, $svc := .dstSvcs }}
     - destination:
-        host: {{ $svc }}
+        host: {{ $svc.Name }}.{{ $svc.Namespace }}
```
howardjohn, line 702: I didn't think  `{{ $svc.Name }}.{{ $svc.Namespace }}` was a valid destination...?

pkg/test/framework/components/echo/match/matchers.go
```diff
@@ -52,116 +50,118 @@ func Or(ms ...Matcher) Matcher {
...
 func Not(m Matcher) Matcher {
 	return func(i echo.Instance) bool {
 		return !m(i)
 	}
 }
 
-// ServicePrefix matches instances whose service name starts with the given prefix.
-func ServicePrefix(prefix string) Matcher {
+// SameService matches instances with the given namespace and service name.
+func SameService(n model.NamespacedName) Matcher {
```
howardjohn, line 61: nit: SameService was a confusing name to me. Maybe rename Service to ServiceName and name this one Service?

tests/integration/pilot/common/routing.go
nmittler: Reverted to just service name.

howardjohn (approved):

PR #37943, opened 2022-03-15
Add new ENVs for GCP to allow specification of project and location

douglas-reid:
There is a need to allow specification of location and project information for GCP outside of K8s context. This PR will allow this to be accomplished without using partial `GCP_METADATA` specification. In particular, it will now be possible to pass in locality information in bootstrap.

- [ X ] User Experience

/cc @liminw


howardjohn:
Is there any reason aside from UX/aesthetics that this differs from using `GCP_METADATA`? Just trying to make sure I understand

GregHanson:
not stale

Issue #37183, opened 2022-02-07
Add a new approach for CA integration through the Envoy SDS API.


maxlambrecht:
**Describe the feature request**

Add a new approach for CA integration through the Envoy SDS API.

Please see [RFC: Istio CA integration through Envoy SDS](https://docs.google.com/document/d/1zJP6QJukLzckTbdY42ZMLkulGXz4gWzH9SwOh4xoe0A/edit?usp=sharing)

**Describe alternatives you've considered**

The alternative that has been considered is fetching the workload identity materials through the _istio-agent_ calling a [SPIFFE Workload API](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Workload_API.md) implementation; this would require adding a new path to the _istio-agent_ Secrets Manager logic.

These ideas were discussed at Networking WG Meeting on [Nov 11, 2021](https://docs.google.com/document/d/1xHy2jQ8oiwMponMVY2zJr2eUAmHW_Hi9JK42a7cg5Pc/edit#heading=h.6wqx1ytjoh28). 

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[X] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

hzxuzhonghu:
@johnzheng1975 I guess this is what you want

PR #37947, opened 2022-03-15
CA integration through Envoy SDS

maxlambrecht:

This PR includes the following changes:

- Make istio-agent check the presence of a Unix Domain Socket on the fixed path `/var/run/secrets/workload-spiffe-uds/socket`:

        1. In the case where the socket exists, istio-agent does not start its own SDS Server.
        2. In the case where the socket does not exist, istio-agent checks whether certificate files on the fixed path folder (`/var/run/secrets/workload-spiffe-credentials`) exists.  If the certificate files exist, istio-agent starts its own SDS Server, listening on the fixed socket path (`/var/run/secrets/workload-spiffe-uds/socket`), to serve those certificate files, keeping file  watchers on them. Envoy proxy connects to the SDS Server run by istio-agent through the fixed socket path and gets the cryptographic materials of the certificate files served by the SDS API.
        3. In case istio-agent does not find either the socket, or the certificate files in the fixed path;  it starts its own SDS Server using a `caClient` to connect to istiod to get the cryptographic materials (as per default istio CA behavior).

- Add volumes and volumeMounts for the fixed paths to the chart manifests.
- Add new tests for istio-agent covering the scenarios where the socket is detected and where the certificates are detected.
- Update agent's README with the details of the new CA integration approach.
- Add release note. 

This PR solves the issue: [#37183](https://github.com/istio/istio/issues/37183)


manifests/charts/gateways/istio-ingress/templates/deployment.yaml
```diff
@@ -264,6 +268,10 @@ spec:
 {{ toYaml $gateway.additionalContainers | indent 8 }}
 {{- end }}
       volumes:
+      - emptyDir:
```
howardjohn, line 271:
```suggestion
      - emptyDir: {}
```

This is probably super nitpicky but its consistent with below and I have a very small worry that somehow something will treat them differently

manifests/charts/gateways/istio-ingress/templates/injected-deployment.yaml
```diff
@@ -118,6 +122,10 @@ spec:
 {{ toYaml $gateway.additionalContainers | indent 8 }}
 {{- end }}
       volumes:
+      - emptyDir:
```
howardjohn, line 125: not needed here since its in [gateway-injection-template.yaml](https://github.com/istio/istio/pull/37947/files#diff-34a675f455a18d4c77afb1cb2bc877a0a77315b3f8dfb8920ddf2f5af96cecb0)

manifests/charts/istio-control/istio-discovery/files/gateway-injection-template.yaml
```diff
@@ -131,10 +131,15 @@ spec:
       timeoutSeconds: 3
       failureThreshold: {{ .Values.global.proxy.readinessFailureThreshold }}
     volumeMounts:
+    - name: workload-socket
+      mountPath: /var/run/secrets/workload-spiffe-uds
     {{- if eq .Values.global.caName "GkeWorkloadCertificate" }}
-    - name: gke-workload-certificate
+    - name: workload-certs
       mountPath: /var/run/secrets/workload-spiffe-credentials
       readOnly: true
```
howardjohn, line 139: nit: I think we can probably drop the conditional and merge these, even though it means it will not be readOnly... @lei-tang any concerns?

tools/packaging/common/envoy_bootstrap.json
```diff
@@ -231,7 +231,7 @@
               "endpoint": {
                 "address":{
                   "pipe": {
-                    "path": "{{ .config.ConfigPath }}/SDS"
+                    "path": "./var/run/secrets/workload-spiffe-uds/socket"
```
howardjohn, line 234: My concern with using the same path for both is for workload restarts. I am pretty sure this can happen in Kubernetes since the socket is now in a volume mount, and it definitely can with VMs.

1. Run 1 - we detect no socket, so we create it and start server.
2. restart for various reasons
3. Run 2 - we detect a socket, so we connect to it and skip our server. Socket never sends anything

Do we prevent this somehow?

howardjohn (commented):
Nice! Just one main comment, otherwise looks great

pkg/istio-agent/agent.go
```diff
@@ -406,13 +406,30 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
...
 	}
 
-	a.secretCache, err = a.newSecretManager()
-	if err != nil {
-		return nil, fmt.Errorf("failed to start workload secret manager %v", err)
-	}
+	if socketFileExists(security.WorkloadIdentitySocketPath) {
+		log.Info("SDS socket found. Istio SDS Server won't be started")
+	} else {
+		if security.CheckWorkloadCertificate(security.WorkloadIdentityCertChainPath, security.WorkloadIdentityKeyPath, security.WorkloadIdentityRootCertPath) {
```
hzxuzhonghu, line 412: Not sure who is responsible to provide certs in env rather than gke

howardjohn: SPIRE is adding it, and GKE already does. Any CSI driver or similar could implement the same as long as they get the certs into the well known file path

hzxuzhonghu: Oh, that means spire can provide certs to the dir or serve them through  uds

manifests/charts/istio-control/istio-discovery/files/gateway-injection-template.yaml
maxlambrecht: >     1. In original code, the CSI driver volume "workloadcertificates.security.cloud.google.com" is created if global.caName is "GkeWorkloadCertificate". Why is "gke-workload-certificate" changed to "workload-certs"?
> 

As it's the same path we thought it makes sense to use the same volume name. I'll revert that change it you prefer to keep "gke-workload-certificate" when the caName is GkeWorkloadCertificate.
 
>     2. When caName is not "GkeWorkloadCertificate", why does the PR mount an emptyDir to /var/run/secrets/workload-spiffe-credentials which is used to mount GKE workload certificate? Which CA will provision cert files to this emptyDir? How to configure Istio to choose this CA as the CA provider?
> 

The idea was to reuse the same path for mounting certificates for the the Workload. There's no configuration needed, it detects the presence of the certs files on the path.

pkg/istio-agent/agent.go
maxlambrecht: SPIRE using a helper and other CAs could mount the certs on the dir. The main and recommended integration with SPIRE is through UDS.

pkg/security/security.go
```diff
@@ -403,6 +415,21 @@ func GetOSRootFilePath() string {
...
+	}
+	if _, err := os.Stat(keyFilePath); err != nil {
+		return false
+	}
+	if _, err := os.Stat(rootCertFilePath); err != nil {
+		return false
+	}
+	return true
+}
+
```
Chen-Xintong, line 432: Hi, which part is responsible for checking the certificate expiration time, is it Istio CA or Spire Agent Server? If the certificate expires and Istio's own SDS secretManager not started, what should we do in this scenario?

manifests/charts/istio-control/istio-discovery/files/gateway-injection-template.yaml
lei-tang: To be consistent with the caName "GkeWorkloadCertificate" and to avoid growing this PR into changing the existing GkeWorkloadCertificate workflow, I think it may be better to keep the original code for the caName "GkeWorkloadCertificate" as is (such as "gke-workload-certificate"). What do you think?
> 
> The idea was to reuse the same path for mounting certificates for the the Workload. There's no configuration needed, it detects the presence of the certs files on the path.
Which entity will provision the cert files on the path?

howardjohn: I am not sure there is much value in keeping the GKE-x specific names. The whole idea is that the GKE cert mounting is no longer a GKE specific thing, but a standard that is implemented by multiple vendors

maxlambrecht: Yes, I think it can be like this:

```
volumes:
  {{- if eq .Values.global.caName "GkeWorkloadCertificate" }}
  - name: gke-workload-certificate
    csi:
      driver: workloadcertificates.security.cloud.google.com
  {{- else}}
  - emptyDir:
    name: workload-certs
  {{- end }}

volumeMounts:
    {{- if eq .Values.global.caName "GkeWorkloadCertificate" }}
    - name: gke-workload-certificate
      mountPath: /var/run/secrets/workload-spiffe-credentials
      readOnly: true
    {{- else }}
    - name: workload-certs
      mountPath: /var/run/secrets/workload-spiffe-credentials
    {{- end }}
```
We still need to put the `workload-certs` on the `else` because we are reusing the same path.

lei-tang: @maxlambrecht The proposed code looks good to me. Thanks!

manifests/charts/gateways/istio-ingress/templates/deployment.yaml
maxlambrecht: I updated all the charts with that suggestion. Thanks

tools/packaging/common/envoy_bootstrap.json
maxlambrecht: In `3.`, do you mean that a socket injected by SPIRE never sends anything ?

howardjohn: In this flow I am talking only standard Istio. So no, 3 is the socket created by the first istio-agent run. The socket file is not removed when the process restarts

maxlambrecht: Got it, we are not handling that scenario. I think that scenario could be addressed by doing a test call to the socket from the istio agent using a SDS client, to check that the socket is responding. Not sure what would the best approach if the socket is not responding, though.

pkg/istio-agent/agent.go
hzxuzhonghu: got it thanks

manifests/charts/gateways/istio-egress/templates/deployment.yaml
```diff
@@ -264,6 +268,10 @@ spec:
 {{ toYaml $gateway.additionalContainers | indent 8 }}
 {{- end }}
       volumes:
+      - emptyDir: {}
+        name: workload-socket
+      - emptyDir: {}
+        name: workload-certs
```
hzxuzhonghu, line 274: should we only declare the volumes when spire agent is integrated.  Otherwise this is redundant

maxlambrecht: Actually the `workload-socket` volume is always used, either by spire or by the istio-agent to mount its own SDS server. And as the integration is not configured through any property but it's based on the detection of either the socket or the certs file, we could not avoid creating the volumes.

tools/packaging/common/envoy_bootstrap.json
irisdingbj: Should we make it as another path but not use the same path?  Imagine we should have other cases that might need use another sds address.

pkg/istio-agent/agent.go
```diff
@@ -406,13 +406,30 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
 		return nil, fmt.Errorf("failed to start local DNS server: %v", err)
 	}
 
-	a.secretCache, err = a.newSecretManager()
-	if err != nil {
-		return nil, fmt.Errorf("failed to start workload secret manager %v", err)
-	}
+	if socketFileExists(security.WorkloadIdentitySocketPath) {
+		log.Info("SDS socket found. Istio SDS Server won't be started")
+	} else {
```
ZackButcher, line 411: nit (I'm not really doing code maintenance anymore so I defer to the folks that actually are these days): This func is getting pretty unwieldy - I'd say pull the `else` out into an `a.initiateSDS()` func.

manifests/charts/gateways/istio-egress/templates/deployment.yaml
ZackButcher: I think there's a lot of value to users in making it "just work" in environments with SPIRE due to the socket existing/not (like we make the certs-on-disk "just work" already). My preferred approach would be to keep this as-is and have it "just work".

However, we _could_ push this up to a top-level option in the install API - "useSPIRE" or something - to configure the mounts and ensure we don't get into the case @howardjohn mentioned (workload restarted and istio-agent didn't clean up the socket so it doesn't start its own SDS server and Envoy never gets a cert). That top level flag would configure the template mounts, but also control the logic in istio-agent's `Run` to decide how the SDS server is started. It's only worthwhile to do this if we can't agree on a clean-enough solution to the socket/SDS server lifecycle stuff.

ZackButcher (commented):
Just wanted to chime in and say I'm excited to see this PR - it really unblocks some of our customers trying to use Istio+SPIRE. ...And I couldn't help but leave a drive-by comment or two while I was here 😉

tools/packaging/common/envoy_bootstrap.json
maxlambrecht: @howardjohn, I've looking into the problem with an old socket hanging around and I think we could check if it's responsive and use it, otherwise try to delete it and start the default SDS server:

```
socketExists := socketFileExists(security.WorkloadIdentitySocketPath)

// if the socket exists, check whether it's responsive:
// - in the case it's responsive: istio-agent doesn't start its own SDS server
// - in the case it's not responsive, it's probably an old socket from a previous istio-agent instance
//   and then try to remove the socket and set the flag "socketExist" to false
//   if the socket cannot be removed, its another problem and the startup should fail
if socketExists {
	conn, err := net.Dial("unix", security.WorkloadIdentitySocketPath)
	if err != nil {
		err = os.Remove(security.WorkloadIdentitySocketPath)
		if err != nil {
			return nil, fmt.Errorf("error connecting to existing socket and could not be removed: %v", err)
		}
		// setting socket existence flag to false
		socketExists = false
	} else {
		conn.Close()
		log.Info("SDS socket found. Istio SDS Server won't be started")
	}
}

if !socketExists { 
          ///  ...
 }
```

What do you think?

manifests/charts/gateways/istio-egress/templates/deployment.yaml
linsun: Is it correct to assume workload-socket and workload-certs can be used by non SPIRE env as well as long as it uses spiffe?  Or is this statement only for workload-socket?

pkg/istio-agent/README.md
```diff
@@ -14,6 +14,17 @@ requests to the agent, which will be forwarded to the configured discovery serve
 
 ## CA Flow
 
+istio-agent checks the presence of a socket file on the defined **socket path** `/var/run/secrets/workload-spiffe-uds/socket`.
+
+* If a socket is found, istio-agent will not start its own SDS Server and Envoy will be configured to use that socket as its source of cryptographic material.
+* If a socket is not found, istio-agent then checks whether certificate files are present or not on the defined **certificate path** `/var/run/secrets/workload-spiffe-credentials`. If certificate files are found, istio-agent will start its own SDS Server, listening and serving these certificates on the defined **socket path** `/var/run/secrets/workload-spiffe-uds/socket`,
+   while also keeping file watchers on them. Envoy proxy then connects to istio-agent's SDS Server through the defined socket path and gets the cryptographic materials of the certificate files served by the SDS API.
+* If istio-agent does not find either the socket, or the certificate files in their respective paths it will start its own SDS Server using a `caClient` to connect to istiod and to fetch cryptographic materials (See Default CA Flow).
```
linsun, line 22:
```suggestion
* If istio-agent does not find either the socket, or the certificate files in their respective paths it will start its own SDS Server using a `caClient` to connect to istiod or an external CA and to fetch cryptographic materials (See Default CA Flow).
```

linsun: @maxlambrecht can you check if this is correct?  ^^

pkg/istio-agent/agent.go
```diff
@@ -406,13 +406,30 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
...
+		if security.CheckWorkloadCertificate(security.WorkloadIdentityCertChainPath, security.WorkloadIdentityKeyPath, security.WorkloadIdentityRootCertPath) {
+			log.Info("workload certificate files detected, creating secret manager without caClient")
```
linsun, line 413: Do we need to check if the cert file is still valid, not expired yet?

linsun:
Hi @maxlambrecht great PR - nice job!!!

 I didn't see you touch any changes related to VM.  Are there any changes for VM integration coming in a future PR?

linsun:
Also, to minimize impact, is it possible to dark launch this first?

howardjohn:
> I didn't see you touch any changes related to VM. Are there any changes for VM integration coming in a future PR?

I think it just works? if the file is there, we use it. Its up to the user to get that file onto the VM, nothing for Istio to do there AFAIK

security/pkg/nodeagent/sds/server.go
```diff
@@ -81,12 +81,12 @@ func (s *Server) Stop() {
...
 
-func (s *Server) initWorkloadSdsService(options *security.Options) {
+func (s *Server) initWorkloadSdsService() {
 	s.grpcWorkloadServer = grpc.NewServer(s.grpcServerOptions()...)
 	s.workloadSds.register(s.grpcWorkloadServer)
 
 	var err error
-	s.grpcWorkloadListener, err = uds.NewListener(options.WorkloadUDSPath)
+	s.grpcWorkloadListener, err = uds.NewListener(security.WorkloadIdentitySocketPath)
 	if err != nil {
```
Chen-Xintong, line 90: Can we make this `WorkloadIdentitySocketPath` configurable to make it more generic to meet the needs of other SDS?

pkg/istio-agent/README.md
maxlambrecht: Yes, that is correct, istio-agent can connect either to istiod or an external CA, configuring a different `CA_ADDR`.

pkg/istio-agent/agent.go
maxlambrecht: We didn't check the cert expiration to keep consistency with the GkeWorkloadCertificate logic, where the certificates presence is detected and they are not checked for validity.

tools/packaging/common/envoy_bootstrap.json
howardjohn: Hmm, I am surprised that works. What is the socket returning? I would expect a GRPC server to not send any bytes until we send it some?

maxlambrecht: The socket is returning the settings frame on the first `Read()`, that's why it works. 

But I explored and tested another way that is more robust, using the `grpcDialContext` with a Deadline:

```
ctx, cancel := context.WithDeadline(ctx, time.Now().Add(time.Second))
defer cancel()

conn, err := grpc.DialContext(ctx, fmt.Sprintf("unix:%s", security.WorkloadIdentitySocketPath),
                        grpc.WithTransportCredentials(insecure.NewCredentials()),
		        grpc.FailOnNonTempDialError(true),
			grpc.WithReturnConnectionError(),
			grpc.WithBlock(),
		)
```

security/pkg/nodeagent/sds/server.go
Chen-Xintong: There is only one sds server for each pod which is no doubt, but my concern is the path names of uds: `/var/run/secrets/workload-spiffe-uds/socket` and cert: `/var/run/secrets/workload -spiffe-credentials` seems to be a bit specific. For HSM, we can reuse sockets in the same mount path, which is no problem, but I would like to suggest marking this path as a more generic path, such as `/var/run/ secrets/workload-external-uds/socket`, which seems more reasonable?

I think the best way is to make this name configurable by user who decide the name of the socket and secret, but no matter which option, each pod will only enable one SDS Server , such as Spire or HSM or other.

howardjohn: Why? Its still spiffe credentials. Spiffe != Spire, and "Spiffe" and HSM or orthogonal. Spiffe just refers to the type of certificate.

Even if the name is not perfect, I would prefer to use an existing standard that is suboptimal than a new one

howardjohn: And strongly prefer 'zero config' for users.

tools/packaging/common/envoy_bootstrap.json
howardjohn: Makes sense to me. In the happy case it is ~instant right, not 1s?

howardjohn (commented):
I am ready to approve this but want to get someone else to approve as well, since this spans multiple working groups; while I am CODEOWNER for all I do not want to tick all the approvals from just me.

I think we are also waiting on the stale socket detection? Unless I just missed it.

Just  an FYI, code freeze is april 18th: https://github.com/istio/istio/wiki/Istio-Release-1.14. So we have a lot of time but not forever :-). Lets try to make sure we can get this ready to go by then?

lei-tang (approved):

lei-tang:
> I am ready to approve this but want to get someone else to approve as well, since this spans multiple working groups; while I am CODEOWNER for all I do not want to tick all the approvals from just me.
> 
> I think we are also waiting on the stale socket detection? Unless I just missed it.
> 
> Just an FYI, code freeze is april 18th: https://github.com/istio/istio/wiki/Istio-Release-1.14. So we have a lot of time but not forever :-). Lets try to make sure we can get this ready to go by then?

@howardjohn Thanks for your planning for the PR! I have approved the PR on behalf of security WG and as a release manager.

tools/packaging/common/envoy_bootstrap.json
maxlambrecht: Yes, took around 400 µs in my tests in a kind cluster to check the SPIRE socket.

maxlambrecht: @howardjohn, I added the socket health check and did some refactors on my code to reduce the complexity of the `Run()` method.

pkg/istio-agent/agent.go
maxlambrecht: Thank you @ZackButcher, that's a good suggestion, I did that refactor as part of my latest changes where I also added the socket health check.

lei-tang:
> @lei-tang, the check for the release note always fails, even though the release note file is part of the PR, do you know what could be happening ?

The error message in the build log (https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/37947/release-notes_istio/1508932685504974848/build-log.txt) shows that "No release notes files found in '/releasenotes/notes/'."

Your PR https://github.com/istio/istio/pull/37947 contains "releasenotes/notes/37947.yaml".

The issue https://github.com/istio/istio.io/issues/37947 does not exist. Hence the failure.
You should name the release notes file based on an existing issue. Once that is fixed, the failure should be gone.

howardjohn:
I think the release note issue is a bug in the check, its maxing out at 100 files. So it doesn't see the release note file which is file >100.

security/pkg/nodeagent/sds/server.go
irisdingbj: I think we will have other cases that need rely on this external SDS server feature as well.   So it is better we make the path name more generic  like : `/var/run/secrets/external-sds/socket`  and  `/var/run/secrets/external-sds-credentials`   The current name is a little specific, right?

mchurichi:
> I think the release note issue is a bug in the check, its maxing out at 100 files. So it doesn't see the release note file which is file >100.

This is indeed a bug/limitation on gh cli (https://github.com/cli/cli/discussions/5359). Is there any workaround for this you folks can think of?

security/pkg/nodeagent/sds/server.go
irisdingbj: Talked with @howardjohn since all cert will be in spiffe format,  agree it is a reasonable name  :)

Chen-Xintong: I also think the name is reasonable enough because we won't change the spiffe standard cert for Istio certificates, same as @irisdingbj, thanks @howardjohn : )

pkg/istio-agent/agent.go
```diff
@@ -629,6 +662,54 @@ func fileExists(path string) bool {
...
+
+func socketHealthCheck(ctx context.Context, socketPath string) error {
+	ctx, cancel := context.WithDeadline(ctx, time.Now().Add(time.Second))
+	defer cancel()
+
+	conn, err := grpc.DialContext(ctx, fmt.Sprintf("unix:%s", socketPath),
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.FailOnNonTempDialError(true),
+		grpc.WithReturnConnectionError(),
+		grpc.WithBlock(),
```
hzxuzhonghu, line 703: only WithBlock is commonly used

hzxuzhonghu (approved):

pkg/istio-agent/agent.go
```diff
@@ -405,13 +406,20 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
...
+	socketExists, err := checkSocket(ctx, security.WorkloadIdentitySocketPath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to start workload secret manager %v", err)
+		return nil, fmt.Errorf("failed to check SDS socket: %v", err)
 	}
 
-	a.sdsServer = sds.NewServer(a.secOpts, a.secretCache)
-	a.secretCache.SetUpdateCallback(a.sdsServer.UpdateCallback)
+	if socketExists {
+		log.Info("SDS socket found. Istio SDS Server won't be started")
```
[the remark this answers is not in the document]
howardjohn: Would be ideal if we can add something in TestAgent to test this. May be as simple as just pre-creating the socket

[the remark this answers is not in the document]
maxlambrecht: Fixed the `checkSocket` method, seems that one return sentence was lost in refactoring. I tested killing the istio-proxy process and it started back up. 

Added a test in TestAgent with the scenario of an unhealthy socket.

howardjohn (approved):

pkg/istio-agent/agent.go
```diff
@@ -629,6 +662,56 @@ func fileExists(path string) bool {
...
+// Otherwise, returns (true, nil)
+func checkSocket(ctx context.Context, socketPath string) (bool, error) {
+	socketExists := socketFileExists(socketPath)
+	if !socketExists {
+		return false, nil
+	}
+
+	err := socketHealthCheck(ctx, socketPath)
+	if err != nil {
+		log.Errorf("SDS socket health check failed: %v", err)
```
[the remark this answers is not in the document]
maxlambrecht: Bumped to `Debug`

howardjohn:
Thanks for the great work here, this is really exciting and a very well-done PR!

PR #37948, opened 2022-03-15
Adds example under /samples for envoy SDS integration

jufantozzi:

Adds supporting example deployment for #37947 that allows the user easily test the newly implemented feature.

### DIFF
--- a/samples/security/spire/README.md
+++ b/samples/security/spire/README.md
@@ -1,9 +1,9 @@
 # Integrating SPIRE as a CA through Envoy's SDS API
 
 This sample deploys a setup of [SPIRE](https://github.com/spiffe/spire) (the SPIFFE Runtime Environment) as an example of integrating with [Envoy's SDS](https://www.envoyproxy.io/docs/envoy/latest/configuration/security/secret) API. For more information
-on SPIFFE and SPIRE specs, refer to the [SPIFFE Overview](https://spiffe.io/docs/latest/spiffe-about/overview/).
+on the SPIFFE specs, refer to the [SPIFFE Overview](https://spiffe.io/docs/latest/spiffe-about/overview/).
 
-Once SPIRE is deployed and integrated with Istio, this sample deploys a modified version of the [sleep](/samples/sleep/README.md) service and validates that its [identity](https://spiffe.io/docs/latest/spiffe-about/spiffe-concepts/#spiffe-verifiable-identity-document-svid) was issued by SPIRE.
+Once SPIRE is deployed and integrated with Istio, this sample deploys a modified version of the [sleep](/samples/sleep/README.md) service and validates that its [identity](https://spiffe.io/docs/latest/spiffe-about/spiffe-concepts/#spiffe-verifiable-identity-document-svid) was issued by SPIRE. Workload registration is automatically handled by the [k8s-workload-registrar](https://github.com/spiffe/spire/blob/main/support/k8s/k8s-workload-registrar/README.md).
 
 See [Istio CA Integration with SPIRE](https://istio.io/latest/docs/ops/integrations/spire) for further details about this integration.
 
@@ -35,7 +35,7 @@ See [Istio CA Integration with SPIRE](https://istio.io/latest/docs/ops/integrati
   $ kubectl apply -f sleep-spire.yaml
   ```
 
-  Otherwise, manually inject the sidecars before applying:
+  Otherwise, manually inject the sidecar before applying:
 
   ```bash
   $ kubectl apply -f <(istioctl kube-inject -f sleep-spire.yaml)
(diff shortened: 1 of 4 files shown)


samples/security/spire/README.md
```diff
@@ -0,0 +1,72 @@
+# Integrating SPIRE as a CA through Envoy's SDS API
```
ericvn, line 1: It would be good to add the steps in a test in the istio.io PR if possible. I think it would need to be in the set of tests where you install istio (as your require the spire Quickstart first).

samples/security/spire/README.md
```diff
@@ -0,0 +1,72 @@
...
+  $ kubectl apply -f spire-quickstart.yaml
+  ```
+
+1. Ensure that the deployment is completed before moving to the next step. This can be verified by waiting on the `spire-agent` pod to become ready:
+
+  ```bash
+  $ kubectl wait pod --for=condition=ready -n spire -l app=spire-agent
+  ```
+
+1. Use the configuration profile provided to install Istio (requires istioctl v1.15 or newer):
```
ericvn, line 24: 1.15?

samples/security/spire/spire-quickstart.yaml
```diff
@@ -0,0 +1,600 @@
...
+  # Don't change ownership on the contents of the mount since the Workload API
+  # Unix Domain Socket is typically open to all (i.e. 0777).
+  fsGroupPolicy: None
+
+  # Declare support for ephemeral volumes only.
+  volumeLifecycleModes:
+    - Ephemeral
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
```
mchurichi, line 29: Keep in mind that `CustomResourceDefinition` has been [removed](https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#api-changes) from the API Group `apiextensions.k8s.io/v1beta1` in Kubernetes 1.22, and there were a few changes in the spec. Therefore, this CRD doesn't work on the latest versions.

There's an updated CRD here: https://github.com/spiffe/spire/blob/main/support/k8s/k8s-workload-registrar/mode-crd/config/spiffeid.spiffe.io_spiffeids.yaml

samples/security/spire/README.md
jufantozzi: My apologies, I forgot I should've already left a comment about this:

I have two small concerns regarding this:

1. Since this is a new feature, is it a good idea to point out the requirement of the version? Or could that be omitted?

2. If deciding to go with the version requirement, as I didn't know at what version it would end up at, should a placeholder for the version be specified instead of a "guess" to avoid confusion?

linsun: I think it is reasonable to require a given istio version.  If this lands in 1.14, then it should require Istio 1.14.

ericvn: Since you have this usage in a 1.14 doc PR, I am making an assumption this works in 1.14 and that the comment should say 1.14. If this won't work in 1.14, but needs 1.15, then we shouldn't have a doc PR yet, wait until after istio.io branches before merging.  I hope this makes sense. In this repo, it might be hard to tell version, while the user should be able to tell version from the docs (preliminary.istio.io is currently 1.14 docs while istio.io is 1.13 docs).

jufantozzi: Thanks @linsun and @ericvn. I guess I was just being very cautious regarding how long it could take for the main PR to get approved/merged. I'll update this once we have the main PR merged to ensure the correct version.

jufantozzi: Okay I see what you mean now @ericvn, regarding the docs PR targeting 1.14. Seems like its well stablished then, I'll update accordingly.

samples/security/spire/spire-quickstart.yaml
jufantozzi: Nice catch, thanks @mchurichi

samples/security/spire/README.md
jufantozzi: I believe @alexandrealvino is taking care of this.

christian-posta:
I think this "quickstart" install of the spire server also includes the k8s-workload-registrar for auto-registration... we should add that to the doc somewhere and possibly indicate manual registration wouldn't be necessary here

jufantozzi:
> I think this "quickstart" install of the spire server also includes the k8s-workload-registrar for auto-registration... we should add that to the doc somewhere and possibly indicate manual registration wouldn't be necessary here

Hey @christian-posta, thanks for your review.

Thats a good point. One of my considerations when putting this together was to avoid as much as possible information redundancy with the [main document](https://github.com/istio/istio.io/pull/11057) of this integration, as well as keep it simple and short. 

With that said, from a usability stand point, as WL registration is an important concept, I agree it would be good to spare a few more words to note it.

craigbox:
@christian-posta can you propose any changes you'd like before we get this in for 1.14?

PR #37958, opened 2022-03-15
Fix update_deps.sh for use in Mac build-tools.

ericvn:

When running update_deps.sh on a Mac development machine using the build-tools image, the `crane` calls fails as it tries using the Mac's executable to get the credentials, which fails running in a Linux container:
```
build-tools:/work$ ./bin/update_deps.sh
...
2022/03/15 14:06:00 HEAD request failed, falling back on GET: error getting credentials - err: exec: "docker-credential-desktop": executable file not found in $PATH, out: ``
Error: error getting credentials - err: exec: "docker-credential-desktop": executable file not found in $PATH, out: ``
2022/03/15 14:06:00 HEAD request failed, falling back on GET: error getting credentials - err: exec: "docker-credential-desktop": executable file not found in $PATH, out: ``
Error: error getting credentials - err: exec: "docker-credential-desktop": executable file not found in $PATH, out: ``
build-tools:/work$
```

Update the code to run `crane` in a spun up container, in this case which doesn't fail trying to get creds from the Mac. Allow overriding the executable if desired.

Finally, run the script.


howardjohn (commented):
so does `./common/scripts/run.sh crane digest gcr.io/istio-release/iptables` fail for you?

howardjohn:
I would have thought https://github.com/google/go-containerregistry/blob/f1fa40b162a1601a863364e8a2f63bbb9e4ff36e/cmd/crane/cmd/root.go#L75-L83 handled it

ericvn:
I agree. Normally, when doing some Istio things like using release-builder I have the docker config file updated to use the plain test password file (as the creds are inaccessible in the linux container from the Mac host (how the team implemented this a while ago)). That continues to work for release-builder, but something in `crane` behaves differently than the `docker` command when interpreting the file, at least in the test I was doing yesterday.

I think I found another work-around I can do which leaves this code intact. I can doc that on the Mac wiki, and leave this code as is for now.

ericvn:
> so does `./common/scripts/run.sh crane digest gcr.io/istio-release/iptables` fail for you?

That command fails as it's trying to do more work than necessary:
```
./common/scripts/run.sh crane digest gcr.io/istio-release/iptables
docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /work.
See 'docker run --help'.
```
just running `crane` I get the same message as above:
```
crane digest gcr.io/istio-release/iptables
2022/03/16 08:56:39 HEAD request failed, falling back on GET: error getting credentials - err: exec: "docker-credential-desktop": executable file not found in $PATH, out: ``
Error: error getting credentials - err: exec: "docker-credential-desktop": executable file not found in $PATH, out: ``
```

PR #37967, opened 2022-03-16
add stat for total destination rules

ramaraochavali:
We have stat to indicate total VS but not DR. These two are most frequently used Istio resources. So adding a stat for DR.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -599,6 +599,12 @@ var (
 		"Total virtual services known to pilot.",
 	)
 
+	// totalDestinationRules tracks the total number of destination rules
+	totalDestinationRules = monitoring.NewGauge(
+		"pilot_destination_rules",
+		"Total destination rules known to pilot.",
+	)
+
 	// LastPushStatus preserves the metrics and data collected during lasts global push.
 	// It can be used by debugging tools to inspect the push event. It will be reset after each push with the
 	// new version.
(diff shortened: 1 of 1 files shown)


pilot/pkg/model/push_context.go
```diff
@@ -599,6 +599,12 @@ var (
 		"Total virtual services known to pilot.",
 	)
 
+	// totalDestinationRules tracks the total number of destination rules
+	totalDestinationRules = monitoring.NewGauge(
```
hzxuzhonghu, line 603: TBH, i donot know what's the use case about this. Also doubt about the vs monitor. 

If the use case is reasonable, why not we monitor all other kinds of configs.  This is not the business of istio, like when we are using k8s, should k8s provide metrics of how many pods/deployments/secrets. I doubt not.

ramaraochavali: The usecase we ran in to recently is because of a bug in our config generation logic, we accidentally deleted configs. Looking at VS count helped. I was thinking if we have the same for DR, we could alert on if we see drastic decrease in these two resources.

K8s does not provide them AFAIK.

I am fine to drop this if you think it is not needed. But we can not delete VS one though, as people might be using it.

ramaraochavali: I have added for DR because these two are most frequently used.

hzxuzhonghu: I think this is not core istio's business., but the responsibility of a company monitoring system. Otherwise we could include so many things in istio

howardjohn:
K8S actually does have these (or maybe it's etcd), so you could probably
monitor at that level

ramaraochavali:
Ok. I will close this then

PR #37974, opened 2022-03-16
EDS: send incremental pushes on full with only services changed

howardjohn:
This mirrors the Delta EDS code.

Basically, this allows the case of a Full push for a service to only
send the endpoints request. This happens when we create a new Service,
for example.

Note: "Delta EDS" and "Incremental EDS" are essentially exactly
identical other than the gRPC service. In both cases, its impossible to
remove resources (its all based on parent Cluster being removed). In
both cases, we can chose to send as many or as few resources as we want.

The same applies to RDS I believe.

### DIFF
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -568,8 +568,7 @@ func (eds *EdsGenerator) buildEndpoints(proxy *model.Proxy,
 	// ConfigsUpdated=ALL, so in this case we would not enable a partial push.
 	// Despite this code existing on the SotW code path, sending these partial pushes is still allowed;
 	// see https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#grouping-resources-into-responses
-	canSendPartialFullPushes := features.PartialFullPushes && onlyEndpointsChanged(req)
-	if !req.Full || canSendPartialFullPushes {
+	if !req.Full || (features.PartialFullPushes && onlyEndpointsChanged(req)) {
 		edsUpdatedServices = model.ConfigNamesOfKind(req.ConfigsUpdated, gvk.ServiceEntry)
 	}
 	resources := make(model.Resources, 0)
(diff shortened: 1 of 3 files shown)


hzxuzhonghu:
@lizan @lambdai Is the theory above correct?

pilot/pkg/xds/eds.go
```diff
@@ -551,7 +553,7 @@ func (eds *EdsGenerator) buildEndpoints(proxy *model.Proxy,
 	req *model.PushRequest,
 	w *model.WatchedResource) (model.Resources, model.XdsLogDetails) {
 	var edsUpdatedServices map[string]struct{}
-	if !req.Full {
+	if !req.Full || onlyEndpointsChanged(req) {
```
hzxuzhonghu, line 556: This is only a special case, we donot push the unchanged eds. 

lgtm if the above theory is correct.

howardjohn: Updated the PR with a flag in case we are wrong, and a comment explaining why its safe

howardjohn: I think this is definitely safe since we already do it - "Full" is an istio internal concept, to envoy it sees the same thing -- only a subset of load assignments sent.

I am working on another change that is more extremely and that I am less certain is safe

lambdai:
@hzxuzhonghu  I can confirmed the interested resource names have the same life across eds resource name and rds resource name.

hzxuzhonghu (commented):
Wondering should add tests since no one test fails

howardjohn:
@hzxuzhonghu added atest that passes with this change and fails without

ramaraochavali:
This is perfectly safe. IIRC. This is how we used to do in our in-house control plane

ramaraochavali:
LGTM. Will leave it for @hzxuzhonghu to take a look.

stevenctl (approved):

PR #37979, opened 2022-03-17
Update the clusterMatch port match condition

xichengliudui:
```go
// FIXME: Ports on a cluster can be 0. the API only takes uint32 for ports
```
the cluster port can be set to 0，so update the match condition

- [x] Networking


howardjohn (changes requested):
I don't think this will work since we cannot distinguish between unset and 0

Arhell (commented):
/ok-to-test

xichengliudui:
> I don't think this will work since we cannot distinguish between unset and 0

If cannot distinguish between unset and 0, I will turn off pr

PR #37989, opened 2022-03-17
pilot: unify and improve sub/unsub in delta and sotw

howardjohn:
Currently:
* SotW always does a full push on request
* Delta does a partial push only on subscribe, but not on unsubscribe.
* SotW will never skip a push, and will instead send a push with 0
  resources. Delta will skip a push only for EDS

New behavior:
* Both Delta and SotW will do partial pushes for subscribes.
* Both Delta and SotW, If there are only unsubcribes, we will not send a response at all

This optimizes request cases. For example, say we have 100 listeners and
remove 10. 45s later, those listeners will drain. This results in an XDS
requests with 99 routes, 98, 97 ... 90 routes - about 1000 routes
generated in total. With this optimization, we will send 0.

On the other side, say we add 1 listener. Before, we would push RDS
(100), and then immediately get a RDS request and send all 101 again.
With this change, we would send the same 100 in the initial RDS push,
but on the request we would just send the new 1 requested.

note: this builds on https://github.com/istio/istio/pull/37974

TODO:
* Add more tests
* Make sure all types are consistent with sending no response vs empty response. And get response from Envoy which is best
* Fix ADSC to support this
* Flag protect new behavior

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -357,15 +357,17 @@ type PushRequest struct {
 	// classifying a single trigger as having multiple reasons.
 	Reason []TriggerReason
 
+	// Delta defines the resources that were added or removed as part of this push request.
+	// This is set only on requests from the client which change the set of resources they (un)subscribe from.
 	Delta ResourceDelta
 }
 
 // ResourceDelta records the difference in requested resources by an XDS client
 type ResourceDelta struct {
 	// Subscribed indicates the client requested these additional resources
-	Subscribed []string
+	Subscribed sets.Set
 	// Unsubscribed indicates the client no longer requires these resources
-	Unsubscribed []string
+	Unsubscribed sets.Set
 }
 
 func (rd ResourceDelta) IsEmpty() bool {
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -356,6 +356,20 @@ type PushRequest struct {
 	// There should only be multiple reasons if the push request is the result of two distinct triggers, rather than
 	// classifying a single trigger as having multiple reasons.
 	Reason []TriggerReason
+
+	Delta ResourceDelta
+}
+
+// ResourceDelta records the difference in requested resources by an XDS client
+type ResourceDelta struct {
+	// Subscribed indicates the client requested these additional resources
+	Subscribed []string
+	// Unsubscribed indicates the client no longer requires these resources
+	Unsubscribed []string
+}
+
+func (rd ResourceDelta) IsEmpty() bool {
+	return len(rd.Subscribed) == 0 && len(rd.Unsubscribed) == 0
 }
 
 type TriggerReason string
--- a/pilot/pkg/xds/delta.go
+++ b/pilot/pkg/xds/delta.go
@@ -17,6 +17,7 @@ package xds
 import (
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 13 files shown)


howardjohn:
/test all

it works! just need to fix our ADS client which doesn't like it I guess

ramaraochavali (commented):
Approach looks good to me. We can try this since it is controlled by flag for sotw xds

pilot/pkg/model/push_context.go
```diff
@@ -360,6 +360,20 @@ type PushRequest struct {
 	// There should only be multiple reasons if the push request is the result of two distinct triggers, rather than
 	// classifying a single trigger as having multiple reasons.
 	Reason []TriggerReason
+
+	Delta ResourceDelta
```
hzxuzhonghu, line 364: Is this the current pushrequest's? Worth a comment

pilot/pkg/util/sets/string.go
```diff
@@ -139,3 +139,18 @@ func (s Set) Equals(other Set) bool {
 func (s Set) Empty() bool {
 	return len(s) == 0
 }
+
+// Diff takes a pair of Sets, and returns the elements that occur only on the left and right set.
+func (s Set) Diff(other Set) (left []string, right []string) {
```
hzxuzhonghu, line 144: nit: Like Intersection, add an example

hzxuzhonghu: But i think you can reuse `Difference`

pilot/pkg/model/push_context.go
howardjohn: Yes, let me add a comment

pilot/pkg/xds/delta.go
```diff
@@ -389,12 +394,17 @@ func (s *DiscoveryServer) pushDeltaXds(con *Connection,
 	t0 := time.Now()
 
 	originalW := w
-	// If subscribe is set, client is requesting specific resources. We should just generate the
+	// If delta is set, client is requesting new resources or removing old ones. We should just generate the
 	// new resources it needs, rather than the entire set of known resources.
-	if subscribe != nil {
+	// Note: we do not need to account for unsubscribed resources as these are handled by parent removal;
+	// See https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#deleting-resources.
+	// This means if there are only removals, we will send an empty response.
```
hzxuzhonghu, line 401: why not skip push?

howardjohn: We do actually, I need to update the comment

hzxuzhonghu (commented):
generally LGTM

hzxuzhonghu (approved):

pilot/pkg/util/sets/string.go
ramaraochavali, line 144: Looks like this is not being used any more? But it is OK to keep it for future use

pilot/pkg/xds/delta.go
```diff
@@ -389,12 +394,17 @@ func (s *DiscoveryServer) pushDeltaXds(con *Connection,
 	t0 := time.Now()
 
 	originalW := w
-	// If subscribe is set, client is requesting specific resources. We should just generate the
+	// If delta is set, client is requesting new resources or removing old ones. We should just generate the
 	// new resources it needs, rather than the entire set of known resources.
-	if subscribe != nil {
+	// Note: we do not need to account for unsubscribed resources as these are handled by parent removal;
```
ramaraochavali, line 399: Is this true of delta variant? Based on the xds link (the delete resources section you shared) this is the behaviour for sotW only.
May be just comment issue?

ramaraochavali:
LGTM. Couple of comments

pilot/pkg/xds/delta.go
ramaraochavali: Strange. We should update envoy docs then

shankgan (approved):

nmittler (approved):

PR #37992, opened 2022-03-17
wip: standardize environment variable setup

howardjohn:
Not ready for review


ericvn:
I did run several commands, `./tools/docker --help`, `./common/scripts/run.sh ./tools/docker --help` `DOCKER_TARGETS="" make docker`, and `make docker` on my Mac and all the command completed as expected so the later should be fixed, yet again, on Macs.

common/scripts/setup_env.sh
```diff
@@ -65,30 +73,27 @@ fi
 
 # Build image to use
 if [[ "${IMAGE_VERSION:-}" == "" ]]; then
-  export IMAGE_VERSION=master-2022-03-17T19-05-13
+  IMAGE_VERSION=master-2022-03-17T19-05-13
```
ericvn, line 76: This change might require a tweak in the automation that occurs post a `tools` image build.

ericvn: Might actually be OK: https://github.com/istio/common-files/blob/master/bin/update-build-image.sh#L52

common/scripts/setup_env.sh
```diff
@@ -163,21 +168,67 @@ fi
...
+# target.
+if [[ "${FOR_BUILD_CONTAINER:-0}" -eq "1" ]]; then
+  LOCAL_OUT="${TARGET_OUT_LINUX}"
+else
+  LOCAL_OUT="${TARGET_OUT}"
+fi
+
+if [[ "${FOR_BUILD_CONTAINER:-0}" -eq "1" ]]; then
+  # Override variables with container specific
+  TARGET_OUT=${CONTAINER_TARGET_OUT}
```
ericvn, line 186: Does it makes things simpler if we include this in the above than's?

common/scripts/setup_env.sh
```diff
@@ -163,21 +168,67 @@ fi
...
+  TARGET_OUT_LINUX=${CONTAINER_TARGET_OUT_LINUX}
+  REPO_ROOT=/work
+fi
+
+go_os_arch=${LOCAL_OUT##*/}
+# Golang OS/Arch format
+LOCAL_GO_OS=${go_os_arch%_*}
+LOCAL_GO_ARCH=${go_os_arch##*_}
+
+BUILD_WITH_CONTAINER=0
```
ericvn, line 196: Not sure if we want this overridable?

howardjohn: My expectation is you are either:

1. out of build container and setting up env, and planning to stay out of the build container
2. Out of the build container and setting up, but preparing to go into the container. FOR_BUILD_CONTAINER=1 is set. Once you are in the container, you do not want nested container, so BUILD_WITH_CONTAINER=0 is set.
It was the same prior to this PR btw

ericvn: A reason it may not need to be is that `BUILD_WITH_CONTAINER=1 make clean` does require the build-tools image while being 0 did not.

howardjohn: Not sure I understand. If you run `BUILD_WITH_CONTAINER=1 make clean` then:

* `./common/scripts/run.sh make clean` is run
* This sources setup_env with FOR_BUILD_CONTAINER=1
* Env is setup with all container specific things, and BUILD_WITH_CONTAINER=0
* Any future calls to make will have BUILD_WITH_CONTAINER=0 and not do nested containers

That is how I would expect it to work. What do you expect?

ericvn: That is what I expected. I was hoping that the BUILD_WITH_CONTAINER=1 wouldn't have been overriden and not happened in a container at all. It appears that it wasn't.

PR #37996, opened 2022-03-18
Generate grpc bootstrap only when DISABLE_ENVOY=true and GRPC_XDS_BOO…

hzxuzhonghu:
…TSTRAP set

Without this, it always generate grpc bootstrap for sidecar.

```
k exec -ti sleep-557747455f-n4lf8 -c istio-proxy -- ls etc/istio/proxy/
SDS  XDS  envoy-rev0.json  grpc-bootstrap.json

```

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/cmd/pilot-agent/options/options.go
+++ b/pilot/cmd/pilot-agent/options/options.go
@@ -113,7 +113,7 @@ var (
 
 	// Defined by https://github.com/grpc/proposal/blob/c5722a35e71f83f07535c6c7c890cf0c58ec90c0/A27-xds-global-load-balancing.md#xdsclient-and-bootstrap-file
 	grpcBootstrapEnv = env.RegisterStringVar("GRPC_XDS_BOOTSTRAP", filepath.Join(constants.ConfigPathDir, "grpc-bootstrap.json"),
-		"Path where gRPC expects to read a bootstrap file. Agent will generate one if set.").Get()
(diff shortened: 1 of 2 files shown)


stevenctl (changes requested):
This is intentional. If the application sets the bootstrap env it will pickup the bootstrap without needing to modify istio settings.

hzxuzhonghu:
Not sure the default setting though

stevenctl:
> Not sure the default setting though

The default setting of what? 

Also we can't have `only when DISABLE_ENVOY`; It's possible to have envoy enabled and exclude some gRPC ports from capture. That way if a Pod runs some regular TCP/HTTP stuff + gRPC, we can still use Envoy for non-grpc traffic.

hzxuzhonghu:
>The default setting of what?

The xds bootstrap env is not empty. 

The use case sounds very weird,

stevenctl:
This usecase has been cited by @costinm in the past. 

I don't really see the negative impact in generating the bootstrap. We don't have the actual env set, so the gRPC library won't actually use it unless the user modifies their app container to do so.

costinm:
The use case is an app that is using both proxyless gRPC and TCP/HTTP services with Envoy sidecar. 

Note that we are looking at other ways to generate the bootstrap - possibly as a library that can be linked to the app itself, so we may be able to remove this, but let's keep it until we have a replacement.

hzxuzhonghu:
SGTM

Issue #37997, opened 2022-03-18
missing cluster when enable `PILOT_FILTER_GATEWAY_CLUSTER_CONFIG`


zirain:
### Bug Description

- install istio with istioctl and following iop:
```yaml
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  components:
    pilot:
      k8s:
        env:
          - name: PILOT_FILTER_GATEWAY_CLUSTER_CONFIG
            value: "true"
  meshConfig:
    accessLogFile: /dev/stdout
    extensionProviders:
    - name: otel
      envoyOtelAls:
        service: otel-collector.istio-system.svc.cluster.local
        port: 4317
    defaultProviders:
      accessLogging:
      - otel
```
- create `httpbin` and expose with gateway
```shell
kubectl apply -f samples/httpbin/httpbin.yaml
kubectl apply -f samples/httpbin/httpbin-gateway.yaml
```
- access from ingressgateway receive warn log from pod

```console
2022-03-18T06:43:55.848204Z	warn	validation	found invalid extension provider (can be ignored if the given extension provider is not used): invalid extension provider otel: unsupported provider: &{service:"otel-collector.istio-system.svc.cluster.local" port:4317 } of type *v1alpha1.MeshConfig_ExtensionProvider_EnvoyOtelAls

```

### Version

```prose
client version: 1.14-dev
control plane version: 1.14
data plane version: 1.14-dev (4 proxies)
```

howardjohn:
FYI https://github.com/istio/istio/issues/29131

PR #37999, opened 2022-03-18
return ExtesionProvider clusters when PILOT_FILTER_GATEWAY_CLUSTER_CO…

zirain:

fixed: https://github.com/istio/istio/issues/37997

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -770,21 +770,20 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 // add services from MeshConfig.ExtensionProviders
 // TODO: include cluster from EnvoyFilter such as global ratelimit [demo](https://istio.io/latest/docs/tasks/policy-enforcement/rate-limit/#global-rate-limit)
 func (ps *PushContext) getHostsFromMeshConfig(proxy *Proxy) sets.Set {
-	hostsFromMeshConfig := sets.NewSet()
+	hostsFromMeshConfig, applicableProviders := sets.NewSet(), sets.NewSet()
 
-	applicablePrviders := sets.NewSet()
 	for _, p := range ps.Telemetry.AccessLogging(proxy).Providers {
 		if p.Name != "" {
-			applicablePrviders.Insert(p.Name)
+			applicableProviders.Insert(p.Name)
 		}
 	}
 
-	if traceingCfg := ps.Telemetry.Tracing(proxy); traceingCfg != nil {
-		applicablePrviders.Insert(traceingCfg.Provider.Name)
+	if tracingCfg := ps.Telemetry.Tracing(proxy); tracingCfg != nil {
+		applicableProviders.Insert(tracingCfg.Provider.Name)
 	}
 
 	for _, prov := range ps.Mesh.ExtensionProviders {
-		if !applicablePrviders.Contains(prov.Name) {
+		if !applicableProviders.Contains(prov.Name) {
 			continue
 		}
 
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -748,7 +748,7 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 	}
 
 	hostsFromMeshConfig := ps.getHostsFromMeshConfig(proxy)
-	hostsFromGateways.Insert(hostsFromMeshConfig.UnsortedList()...)
+	hostsFromGateways.Merge(hostsFromMeshConfig)
 
 	log.Debugf("GatewayServices: gateway %v is exposing these hosts:%v", proxy.ID, hostsFromGateways)
 
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -724,8 +724,6 @@ func virtualServiceDestinationHosts(v *networking.VirtualService) []string {
 // GatewayServices returns the set of services which are referred from the proxy gateways.
 func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 	svcs := proxy.SidecarScope.services
-	// host set.
-	hostsFromGateways := map[string]struct{}{}
 
 	// MergedGateway will be nil when there are no configs in the
 	// system during initial installation.
@@ -733,6 +731,8 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 		return nil
 	}
 
+	// host set.
+	hostsFromGateways := sets.NewSet()
 	for _, gw := range proxy.MergedGateway.GatewayNameForServer {
 		for _, vsConfig := range ps.VirtualServicesForGateway(proxy.ConfigNamespace, gw) {
 			vs, ok := vsConfig.Spec.(*networking.VirtualService)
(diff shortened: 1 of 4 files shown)


pilot/pkg/model/push_context.go
```diff
@@ -768,6 +771,53 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 	return gwSvcs
 }
 
+// add services from MeshConfig.ExtensionProviders
+// TODO: include cluster from EnvoyFilter such as global ratelimit [demo](https://istio.io/latest/docs/tasks/policy-enforcement/rate-limit/#global-rate-limit)
+func (ps *PushContext) getHostsFromMeshConfig(proxy *Proxy) sets.Set {
+	hostsFromMeshConfig := sets.NewSet()
+
+	applicablePrviders := sets.NewSet()
```
ramaraochavali, line 779: nit. applicablePrviders -> applicableProviders

pilot/pkg/model/push_context.go
```diff
@@ -768,6 +771,53 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
...
+	for _, p := range ps.Telemetry.AccessLogging(proxy).Providers {
+		if p.Name != "" {
+			applicablePrviders.Insert(p.Name)
+		}
+	}
+
+	if traceingCfg := ps.Telemetry.Tracing(proxy); traceingCfg != nil {
```
ramaraochavali, line 786: traceingCfg -> tracingCfg

ramaraochavali (commented):
some initial comments. Will take another look

pilot/pkg/model/push_context.go
zirain: updated

pilot/pkg/model/push_context.go
zirain: updated

pilot/pkg/model/push_context.go
```diff
@@ -768,6 +771,52 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
...
+	if tracingCfg := ps.Telemetry.Tracing(proxy); tracingCfg != nil {
+		applicableProviders.Insert(tracingCfg.Provider.Name)
+	}
+
+	for _, prov := range ps.Mesh.ExtensionProviders {
+		if !applicableProviders.Contains(prov.Name) {
+			continue
+		}
+
+		switch p := prov.Provider.(type) {
```
ramaraochavali, line 794: How do we keep this list in sync with API? any possibility?

zirain: maybe `refelect`? assume there's alway a field named `Address`.

pilot/pkg/model/push_context.go
```diff
@@ -746,11 +746,14 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
...
 
 			for _, host := range virtualServiceDestinationHosts(vs) {
-				hostsFromGateways[host] = struct{}{}
+				hostsFromGateways.Insert(host)
 			}
 		}
 	}
 
+	hostsFromMeshConfig := ps.getHostsFromMeshConfig(proxy)
+	hostsFromGateways.Insert(hostsFromMeshConfig.UnsortedList()...)
```
howardjohn, line 755: performance nit: its wasteful to convert to list just to add it. We could add some new method `func (s Set) Merge(other Set)`, which is like Union but it is in place...?

pilot/pkg/model/push_context.go
howardjohn: I think we can do

```
type Servicer interface {
  GetService() string
}
p, ok := prov.Provider.(Servicer)
if ok {
  insert(p.GetService())
}
```

Not 100% sure though.

If not, I would add a test that uses reflect and compares it here (rather than reflect in the prod code)

pilot/pkg/model/push_context.go
```diff
@@ -768,6 +771,52 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 	return gwSvcs
 }
 
+// add services from MeshConfig.ExtensionProviders
+// TODO: include cluster from EnvoyFilter such as global ratelimit [demo](https://istio.io/latest/docs/tasks/policy-enforcement/rate-limit/#global-rate-limit)
+func (ps *PushContext) getHostsFromMeshConfig(proxy *Proxy) sets.Set {
+	hostsFromMeshConfig, applicableProviders := sets.NewSet(), sets.NewSet()
+
+	for _, p := range ps.Telemetry.AccessLogging(proxy).Providers {
```
howardjohn, line 779: My initial thought was we would just get all the `ExtensionProviders`, not only the applicable ones. Since this is missing some other uses, like External Authz?

If not, it seems a bit wasteful to compute the full config (expensive) just to get the providers?

I think this is missing `metrics` as well?

howardjohn (commented):
Nice!

pilot/pkg/model/push_context.go
zirain: sounds good, will update the implements.

pilot/pkg/model/push_context.go
zirain: > My initial thought was we would just get all the ExtensionProviders, not only the applicable ones. Since this is missing some other uses, like External Authz?

that's what I do in the first version, then I take a look at `Metric` `Trace` and `Logging`, but I forgot `External Authz`. 
you're mostly right, it would be better for just get all the `ExtensionProviders`.

> I think this is missing metrics as well?

IIRC, there're no address on `prometheus` and `stackdriver`.

pilot/pkg/model/push_context.go
zirain: 1. first of all:
```
	for _, prov := range ps.Mesh.ExtensionProviders {
		if p, ok := prov.Provider.(ExtensionProvider); ok {
			hostsFromMeshConfig.Insert(p.GetService())
		}
	}
```
the code will not working.

```goalng
case *meshconfig.MeshConfig_ExtensionProvider_EnvoyExtAuthzHttp:
	hostsFromMeshConfig.Insert(p.EnvoyExtAuthzHttp.Service)
case *meshconfig.MeshConfig_ExtensionProvider_EnvoyExtAuthzGrpc:
	hostsFromMeshConfig.Insert(p.EnvoyExtAuthzGrpc.Service)
```

`GetService()`  is not function of `Provider`.

2. `I would add a test that uses reflect and compares it here (rather than reflect in the prod code)` this may block `Automator: update istio/api@master` work.

pilot/pkg/model/push_context.go
zirain: updated

pilot/pkg/model/push_context.go
zirain: updated

pilot/pkg/model/push_context.go
```diff
@@ -768,6 +771,42 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 	return gwSvcs
 }
 
+type ExtensionProvider interface {
```
howardjohn, line 774: dead code now. Agree it will not work after looking closer

pilot/pkg/model/push_context_test.go
```diff
@@ -1972,6 +1973,103 @@ func TestServiceWithExportTo(t *testing.T) {
 	}
 }
 
+func TestGetHostsFromMeshConfig(t *testing.T) {
```
howardjohn, line 1976: can we add the exhaustiveness check we discussed?

zirain: IMO, not.
once we add a test use reflect count there should have `N` ExtensionProviders, when we add a new `EntensionProvider` in the future, the `Automator: update istio/api@master` PR will be block.

pilot/pkg/model/push_context.go
zirain: fixed

pilot/pkg/model/push_context_test.go
zirain: ok, I will trying to add a test.

zirain: updated, please review again.

howardjohn (approved):

PR #38007, opened 2022-03-18
operator: handle tpath+protobuf consistently

howardjohn:
Currently, we have some higher level hacks to make this work. This drops the hacks to the lower level in tpath so we don' need special logic elsewhere


litong01 (approved):
Thanks very much!

PR #38024, opened 2022-03-20
Optimize config apply and delete

nmittler:
When pushing config to real clusters, each push can take several seconds. These times can stack pretty quickly when pushing a lot of config, which happens from some tests.

The new authz tests in https://github.com/istio/istio/pull/37914, push a lot of config at once. With this change, the config push time on GKE goes from ~1 minute to 5 seconds.

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/hashicorp/go-multierror"
 	"go.uber.org/atomic"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/credentials"
 	v1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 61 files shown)


howardjohn (commented):
I don't have a problem with this but one thing we could try is https://github.com/howardjohn/kubeconfig-proxy. Basically it multiplexs all requests over a single connection instead of a new connection per call. It's not really more than a POC yet so wouldn't be stable to drop in place in CI in current state

Since we are using client-go we could probably do similar directly, will peek at the code to see if anything stands out.

This could be in addition to this PR, wouldn't need to be either-or

howardjohn (commented):
Or I wonder if it's something with API discovery cache not working right in our config or something. GKE is slow compared to kind but 1 minute seems extreme. Will play around with it

nmittler:
@howardjohn do you have any particular concerns with the changes here? I don't think there is anything here that's particularly risky, and it helps with my testing on GKE. We can always make improvements to the internals later.

WDYT?

howardjohn:
no concerns just didn't have a chance to look at it

howardjohn:
fyi: https://github.com/istio/istio/pull/38031. We may still want to do this as well though

ericvn:
I agree that both sound correct.

douglas-reid (approved):
P&T changes LGTM

PR #38030, opened 2022-03-20
Add util functions for reading directories in tests

howardjohn:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/test/util/file/file.go
+++ b/pkg/test/util/file/file.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/test/util/file/file.go
```diff
@@ -128,3 +129,34 @@ func ReadTarFile(filePath string) (string, error) {
 	}
 	return "", fmt.Errorf("file not found %v", filePath)
 }
+
+// ReadDir returns the names of all files in the given directory. This is not recursive.
+// The base path is a appended; for example, ReadDir("dir") -> ["dir/file1", "dir/folder1"]
+func ReadDir(filePath string, extensions ...string) ([]string, error) {
+	dir, err := ioutil.ReadDir(filePath)
```
ericvn, line 136: A reason not to use os.ReadDir?

howardjohn: isn't it the same?

ericvn: They are similar, but I recall that  package io and os implementations are preferred in new code (you may recall when we tried moving to them earlier that the build tooling).

howardjohn: done. If we don't want ioutil used we should add a linter.

ericvn (approved):
Will look into seeing about a lint warning on new changes (we don't necessarily want to go back and change all the old items).

PR #38031, opened 2022-03-20
tf: optimize config apply by reusing discovery cache

howardjohn:
Relates to https://github.com/istio/istio/pull/38024 - I think we should
do both, since each config apply still takes about 250ms on my machine
to GKE with this. With https://github.com/howardjohn/kubeconfig-proxy it
is down to 125ms; I didn't specifically measure this PR + parallel
apply.

Basically the problem is we do resource discovery on EVERY call to apply
YAML, since our discovery cache is not shared. This discovery makes like
50 API server calls each time, which is very slow.

With this caching, the following test is >10x faster:

```go

func TestApplyYAML(t *testing.T) {
  cfgs := file.ReadDirOrFail(t, filepath.Join(testenv.IstioSrc, "tests/integration/security/testdata"), ".yaml")
  client, err := NewExtendedClient(BuildClientCmd("", ""), "")
  if err != nil {
    t.Fatal(err)
  }
  t.Log(client.ApplyYAMLFiles("testns", cfgs...))
}
```

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -48,7 +48,6 @@ import (
 	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/cli-runtime/pkg/resource"
 	"k8s.io/client-go/discovery"
-	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	dynamicfake "k8s.io/client-go/dynamic/fake"
@@ -60,7 +59,6 @@ import (
 	metadatafake "k8s.io/client-go/metadata/fake"
 	"k8s.io/client-go/metadata/metadatainformer"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/restmapper"
 	clienttesting "k8s.io/client-go/testing"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
@@ -367,7 +365,10 @@ func newClientInternal(clientFactory util.Factory, revision string) (*client, er
 	if err != nil {
 		return nil, err
 	}
-	c.mapper = restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(c.discoveryClient))
+	c.mapper, err = clientFactory.ToRESTMapper()
+	if err != nil {
+		return nil, err
+	}
 
 	c.Interface, err = kubernetes.NewForConfig(c.config)
 	c.kube = c.Interface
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (commented):
Why caching in the package, not  in the tf framework

nmittler (approved):

howardjohn:
note to self:

```go

var decUnstructured = yaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)

func (c *client) ssapplyYAMLFile(namespace string, dryRun bool, file string) error {
	d, err := os.ReadFile(file)
	if err != nil {
		return err
	}
	cfgs := yml.SplitString(string(d))
	for _, cfg := range cfgs {
		if err := c.ssapplyYAML(cfg, namespace, dryRun); err != nil {
			return err
		}
	}
	return nil
}

func (c *client) ssapplyYAML(cfg string, namespace string, dryRun bool) error {
	obj := &unstructured.Unstructured{}
	_, gvk, err := decUnstructured.Decode([]byte(cfg), nil, obj)
	if err != nil {
		return err
	}

	// Find GVR
	mapping, err := c.mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
	if err != nil {
		return err
	}

	// Obtain REST interface for the GVR
	var dr dynamic.ResourceInterface
	if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
		ns := obj.GetNamespace()
		if ns == "" {
			ns = namespace
		} else if namespace != "" && ns != namespace {
			return fmt.Errorf("object %v/%v provided namespace %q but apply called with %q", gvk, obj.GetName(), ns, namespace)
		}
		// namespaced resources should specify the namespace
		dr = c.dynamic.Resource(mapping.Resource).Namespace(ns)
	} else {
		// for cluster-wide resources
		dr = c.dynamic.Resource(mapping.Resource)
	}

	// 6. Marshal object into JSON
	data, err := json.Marshal(obj)
	if err != nil {
		return err
	}

	// 7. Create or Update the object with SSA
	//     types.ApplyPatchType indicates SSA.
	//     FieldManager specifies the field owner ID.
	var dryRunArgs []string
	if dryRun {
		dryRunArgs = []string{metav1.DryRunAll}
	}
	force := true
	_, err = dr.Patch(context.Background(), obj.GetName(), types.ApplyPatchType, data, metav1.PatchOptions{
		DryRun:       dryRunArgs,
		Force:        &force,
		FieldManager: "istio-ci",
	})

	return err
}

```

stevenctl (approved):

PR #38040, opened 2022-03-21
Changes log level to debug while fetching AWS info and typo fix

sha-rath:
This PR changes the log level from error to 'debug' for AWSInfo error messages. It is not preferred to see these error logs everytime during startup. 
Fixed typo in "Cloud Platform"

### DIFF
--- a/pkg/bootstrap/platform/aws.go
+++ b/pkg/bootstrap/platform/aws.go
@@ -98,7 +98,7 @@ func getAWSInfo(path string, ipv6 bool) (string, error) {
 
 	resp, err := http.DoHTTPGetWithTimeout(url, time.Millisecond*100)
 	if err != nil {
-		log.Errorf("error in getting aws info for %s : %v", path, err)
+		log.Warnf("error in getting aws info for %s : %v", path, err)
 		return "", err
 	}
 	return resp.String(), nil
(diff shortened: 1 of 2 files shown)


pkg/bootstrap/platform/aws.go
```diff
@@ -98,7 +98,7 @@ func getAWSInfo(path string, ipv6 bool) (string, error) {
 
 	resp, err := http.DoHTTPGetWithTimeout(url, time.Millisecond*100)
 	if err != nil {
-		log.Errorf("error in getting aws info for %s : %v", path, err)
+		log.Warnf("error in getting aws info for %s : %v", path, err)
```
hzxuzhonghu, line 101: prefer Debugf

sha-rath: Sure. I didn't know how to decide on warn or debug. 
I see that azure and gcp have log levels set to warn though..

hzxuzhonghu: ? If set to warn, it will print a confusing log in other platform

sha-rath: Sorry, warn messages are logged only if the cloud platform is explicitly specified.
During the discovery phase (when cloud platform not set), it is good to have these logs at debug level. Agree.

ramaraochavali (approved):

mbelang:
> It is not preferred to see these error logs everytime during startup

May I ask why?

PR #38045, opened 2022-03-21
Fix BUILD_WITH_CONTAINER=0 make docker.*

ericvn:

BUILD_WITH_CONTAINER=0 make docker.* on a Mac (and maybe other platforms) yields:
```
make: *** No rule to make target `docker.*'.  Stop.
```
It seems the various docker targets aren't being created when run locally.

I notice an extra `=` on the new define, and removing that fixes the issue on the Mac.

### DIFF
--- a/tools/istio-docker.mk
+++ b/tools/istio-docker.mk
@@ -57,7 +57,7 @@ dockerx: docker
 # Support individual images like `dockerx.pilot`
 
 # Docker commands defines some convenience targets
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

ericvn:
Merge https://github.com/istio/istio/pull/37935 as it will help with the Makefile command generating items in the correct LOCAL_OUT.

PR #38047, opened 2022-03-21
make KinD multi-cluster better on Mac

stevenctl:
This is WIP, haven't finished testing yet (pulling envoy, build-tools, etc is very slow for me on my current internet connection). 

Rather than modifying the kubeconfigs, here we pass the docker IPs to the framework, and the framework forwards them to the `create-remote-secret` command

Background:

We need the docker IP in the kubeconfig files since we use the kubeconfig file's server IP in `create-remote-secret`. 
The test runner also uses those files to apply configs and control the tests. 
A mac-hosted test-runner won't be able to use docker IPs.

### DIFF
--- a/common/scripts/kind_provisioner.sh
+++ b/common/scripts/kind_provisioner.sh
@@ -79,6 +79,7 @@ function load_cluster_topology() {
   fi
 
   export CLUSTER_NAMES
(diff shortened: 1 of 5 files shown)


stevenctl:
Also I have to make part of this change in common-files first.

PR #38050, opened 2022-03-21
Align Mesh Routes with gateway-api spec

howardjohn:
Currently, Routes are basically mapped 1:1 with VirtualService. This
includes semantics that do not align with the gateway-api spec.

In order to make these spec-compliant without totally rewriting the XDS
generation to read gateway types directly, this introduces a new
(*internal*) option on virtual services to give it "gateway-api
semantics".

Currently, this means:
* Wildcards currently will "Attach" to a FQDN service. Meaning if we
  have "helloworld.com" and `"*.com"`, our route applies to helloworld.com
and nothing else. In Gateway-API, FQDN service has high precedence, so
we will have "helloworld.com" (with no route config) and `"*.com"`
* Currently import logic is bi-directional. `VS: *.com` with `Import:
  foo.com` will be imported; this differs from Service and is
unexpected.
* Currently VS.Hosts are not independant. If you have a VS with [a, b]
  and import only a, you get the host for b.

These 3 cases have been fixed, and test cases have been added to
recognize these differences.

TODO:
* Currently VS picks the first resource across any namespace based on
  timestamp; we need to make it have a selection hierarchy.


costinm (commented):
Few general comments/questions:
1. I am a bit worried about different semantics for the migration from VS to HttpRoute. We need to have some very clear doc/process on how a user can do this, ideally with some automation possible.
2. What happens if I have VirtualService for example.ns.svc.cluster.local, example.ns, etc ( i.e. equivalent names ) ? I think I've seen some uses, not sure how common. 
3. The annotation seem to indicate specific selection semantics - what happens when we have a mix of HttpRoute and VirtualService ? Which semantic wins ? Do we want to put a boundary, like namespace level ? 'If a namespace has Gateway APIs, that semantic is used for the ns' ?
4. I assume it is possible to use the new annotation on regular VS ? Should we 'align' with GW, and start recommending users to use the new semantic, or opt-in to it at namespace level ? Or have some warnings if istio semantics are used ?

howardjohn:
> 1. I am a bit worried about different semantics for the migration from VS to HttpRoute. We need to have some very clear doc/process on how a user can do this, ideally with some automation possible.

ACK

> 2. What happens if I have VirtualService for example.ns.svc.cluster.local, example.ns, etc ( i.e. equivalent names ) ? I think I've seen some uses, not sure how common.

If they have multiple for the same name, they are merged (this is in the gateway-api spec). This is not implemented yet but close in https://github.com/istio/istio/pull/37139. 

> 3. The annotation seem to indicate specific selection semantics - what happens when we have a mix of HttpRoute and VirtualService ? Which semantic wins ? Do we want to put a boundary, like namespace level ? 'If a namespace has Gateway APIs, that semantic is used for the ns' ?

The intent is to only apply the new sematnics to HttpRoute's - VirtualServices remain unchanged so there is one migration (HTTPRoute -> VS) rather than an intermediate state where we try to fix VS which is challenging.

> 4. I assume it is possible to use the new annotation on regular VS ? Should we 'align' with GW, and start recommending users to use the new semantic, or opt-in to it at namespace level ? Or have some warnings if istio semantics are used ?

Hypothetically possible, but not documented or recommended. The fact we use an annotation is an implementation detail since we don't have another way to pass around internal-only API extensions. We can validate it to be not be allowed if we want, but I don't think we should encourage it to be used -- they should use the real API (HTTPRoute)

costinm (approved):
The code looks reasonable - we may still iterate based on real-world tests and feedback, but can be in separate PRs.

For example, all the annotations and names (including 'internal.istio.io) and generated names.

Looking at the names - isn't there a 64 bytes limit on the name ? 'istio-autogenerated-k8s-gateway' seems pretty long...

howardjohn:
All of these names are strictly internal - if we externalize them we should definitely revisit them. The fact they are internal also means we can use invalid names like BTW

PR #38055, opened 2022-03-21
Migrate gogo -> golang protobuf

### LINKED ISSUE #26751: Remove gogo entirely from Istio
removal of gogo. For many reasons.

howardjohn:
Fixes https://github.com/istio/istio/issues/26751

gogo/protobuf was originally introduced into Istio due to its performance benefits over the standard Go implementation. This was important for Mixer runtime performance. In the current state of Istio, Istio CR performance is negligible compared to XDS, which migrated to golang/protobuf a while back.

The gogo project is now unmaintained and no longer updated: https://github.com/gogo/protobuf/issues/691.

The plan is to get this approved, then merge istio/api, and unhold this. This ensures we do not have a delay where api repo merge is blocked.

* API PR: TODO
* client-go PR: https://github.com/istio/api/pull/2293

Changes required:
* Many minor tweaks to change library imports from gogo/protobuf -> golang/protobuf.
* Operator: we know longer have `IstioOperator` (the one defined in this repo that containers the top level K8s object, not the `IstioOperatorSpec` defined in istio/api as a protobuf, since there is no proto file. This means we replace some calls that expected UnmarshalWithJSONPB to just use plain YAML functions; these types have MarshalJSON functions defined so it behaves the same
* Move some tests to compare use `cmp` or `assert.Equal`; reflect.DeepEqual cannot be used with golang protobuf types.
* Make all references to types use pointers; golang protobuf types cannot be copied
* Add `nolint` in some places where things are (properly) marked Deprecated now
* Remove our custom fuzzing tests, used only for round trip testing; we have since replaced these with full OSS-fuzzer. This framework was effectively incompatible with the new protobuf types.


go.mod
```diff
@@ -8,6 +8,10 @@ exclude k8s.io/kubernetes v1.13.0
 // Client-go does not handle different versions of mergo due to some breaking changes - use the matching version
 replace github.com/imdario/mergo => github.com/imdario/mergo v0.3.5
 
+replace istio.io/api => github.com/howardjohn/api v0.0.0-20220321203251-8186a01edef9
```
nmittler, line 11: need to update?

howardjohn: This is intentional. The plan is to get approval for this with a hold, then merge API and update this. I don't want to just merge API now in case this one takes a while (need approval from everyone), so this let's us test in CI

operator/pkg/util/yaml.go
```diff
@@ -23,13 +23,52 @@ import (
...
+		v, err := yaml.Marshal(root)
+		if err != nil {
+			return nil, err
+		}
+		vs = v
+	}
+	return vs, nil
+}
+
+func MustToYAMLGeneric(root interface{}) string {
```
nmittler, line 54: why not just call the other method and panic?

howardjohn: I don't really have a strong opinion - I am not sure if this can be triggered from user input where panic is unsafe?

This is preserving existing functionality - my goal in this PR is to have 0 user facing behavior

nmittler (approved):
@howardjohn changes look great!

shankgan (approved):

ericvn (approved):

stevenctl (approved):

howardjohn:
/test all



github outage

jacob-delgado (approved):

lei-tang (approved):

PR #38057, opened 2022-03-21
tf: Use Server Side Apply for config application

howardjohn:


howardjohn:
Basically requires 1.20+ due to https://github.com/kubernetes-sigs/structured-merge-diff/issues/130. Not sure we should do this for now

PR #38066, opened 2022-03-22
prow scripts: workaround lack of GNU date on OSX

stevenctl:
Do we just want OSX users to alias `date` -> `gdate`? 

An alternative would be like:

```
date <some set of GNU date flags that we know work> || echo "Warning: use gdate as your standard date command"
```

### DIFF
--- a/prow/lib.sh
+++ b/prow/lib.sh
@@ -14,9 +14,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+function date_cmd() {
+  case "$(uname)" in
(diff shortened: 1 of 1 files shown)


stevenctl:
Even in make shell, the docker for mac restrictions apply (see https://github.com/istio/istio/pull/38047). If we can't check `uname`, then the new behavior that PR would introduce would need to be our default (which is _probably_ fine).

ericvn (approved):

howardjohn (approved):

PR #38069, opened 2022-03-22
eds: fix regression in removing healthy pods

howardjohn:
There is a currently an issue where if a endpoint goes from Healthy to
deleted, we do not trigger a push. This manifests in our CI as 503s.

Added a test which triggers this case.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -172,10 +172,14 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string, 
 		// Please note that address is not a unique key. So this may not accurately
 		// identify based on health status and push too many times - which is ok since its an optimization.
 		emap := make(map[string]*model.IstioEndpoint, len(oldIstioEndpoints))
+		nmap := make(map[string]*model.IstioEndpoint, len(newIstioEndpoints))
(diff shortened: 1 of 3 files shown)


GregHanson (approved):

pilot/pkg/xds/eds.go
```diff
@@ -193,6 +197,13 @@ func (s *DiscoveryServer) edsCacheUpdate(shard model.ShardKey, hostname string,
 				newIstioEndpoints = append(newIstioEndpoints, nie)
 			}
 		}
+		// Next, check for endpoints that were in old but no longer exist. If there are any, there is a
+		// removal so we need to push an update.
+		for _, oie := range oldIstioEndpoints {
+			if _, f := nmap[oie.Address]; f {
```
bianpengyuan, line 203: Sorry I am bit confused. Should this be `!f`?

howardjohn: That is a good point... now I am wondering how the test passed...

howardjohn: Ah lol, it passes because one endpoint is `f` and one if `!f`. Let me make the test more robust so it fails with the current code

Issue #38082, opened 2022-03-23
SidecarScope.DependsOnConfig is not correct for DestinationRule


hzxuzhonghu:
We do merge DRs for push context, so when sidecar scope AddConfigDependencies, we lost the some original configs info.  

Then when a dr(merged into another dr) is updated, we could skip push xds to a proxy.

```
// Compute and send the new configuration for a connection.
func (s *DiscoveryServer) pushConnection(con *Connection, pushEv *Event) error {
	pushRequest := pushEv.pushRequest

	if pushRequest.Full {
		// Update Proxy with current information.
		s.updateProxy(con.proxy, pushRequest)
	}

	if !s.ProxyNeedsPush(con.proxy, pushRequest) {
```

hzxuzhonghu:
This only happens on DR merge, so I believe no integration tests cover this case.

PR #38088, opened 2022-03-23
Fix dr dependency miss

hzxuzhonghu:

Fix https://github.com/istio/istio/issues/38082

For merged DRs, currently we have only one DR meta(name/namespace) and add only one DR dependency.  
And so we will miss xds push for some update.

The pr solves this by recording all DRs' meta, and adding all dependencies.

### DIFF
--- a/pilot/pkg/model/destination_rule.go
+++ b/pilot/pkg/model/destination_rule.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 
 	"google.golang.org/protobuf/proto"
+	"k8s.io/apimachinery/pkg/types"
 
 	networking "istio.io/api/networking/v1alpha3"
 	"istio.io/istio/pkg/config"
@@ -41,8 +42,8 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 	if mdrList, exists := p.destRules[resolvedHost]; exists {
 		// `addRuleToProcessedDestRules` determines if the incoming destination rule would become a new unique entry in the processedDestRules list.
 		addRuleToProcessedDestRules := true
-		for i, mdr := range mdrList {
-			existingRule := mdr.Spec.(*networking.DestinationRule)
+		for _, mdr := range mdrList {
+			existingRule := mdr.rule.Spec.(*networking.DestinationRule)
 			bothWithoutSelector := rule.GetWorkloadSelector() == nil && existingRule.GetWorkloadSelector() == nil
 			bothWithSelector := existingRule.GetWorkloadSelector() != nil && rule.GetWorkloadSelector() != nil
 			selectorsMatch := labels.Instance(existingRule.GetWorkloadSelector().GetMatchLabels()).Equals(rule.GetWorkloadSelector().GetMatchLabels())
@@ -52,7 +53,6 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 				// if the selectors do not match
 				continue
 			}
-
 			// If both the destination rules are without a workload selector or with matching workload selectors, simply merge them.
 			// If the incoming rule has a workload selector, it has to be merged with the existing rules with workload selector, and
 			// at the same time added as a unique entry in the processedDestRules.
@@ -62,8 +62,9 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 
 			// Deep copy destination rule, to prevent mutate it later when merge with a new one.
 			// This can happen when there are more than one destination rule of same host in one namespace.
-			copied := mdr.DeepCopy()
-			p.destRules[resolvedHost][i] = &copied
+			copied := mdr.rule.DeepCopy()
+			mdr.rule = &copied
+			mdr.from = append(mdr.from, types.NamespacedName{Namespace: destRuleConfig.Namespace, Name: destRuleConfig.Name})
 			mergedRule := copied.Spec.(*networking.DestinationRule)
 
 			existingSubset := map[string]struct{}{}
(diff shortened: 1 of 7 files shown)


hzxuzhonghu:
@howardjohn Ready now

hzxuzhonghu:
Kindly ping @howardjohn in case you missed it

pilot/pkg/model/destination_rule.go
```diff
@@ -80,37 +83,56 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
 	}
 
 	// DestinationRule does not exist for the resolved host so add it
-	p.destRule[resolvedHost] = &destRuleConfig
+	p.destRule[resolvedHost] = convertConsolidateDestRule(&destRuleConfig)
```
ramaraochavali, line 86: nit: convertConsolidateDestRule -> convertConsolidatedDestRule

pilot/pkg/model/destination_rule.go
```diff
@@ -80,37 +83,56 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
...
 	p.exportTo[resolvedHost] = exportToMap
 }
 
 // inheritDestinationRule child config inherits settings from parent mesh/namespace
-func (ps *PushContext) inheritDestinationRule(parent, child *config.Config) *config.Config {
+func (ps *PushContext) inheritDestinationRule(parent, child *consolidateDestRule) *consolidateDestRule {
```
ramaraochavali, line 91: consolidateDestRule -> consolidatedDestRule

pilot/pkg/model/destination_rule.go
```diff
@@ -80,37 +83,56 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
...
 	if parent == nil {
 		return child
 	}
 	if child == nil {
 		return parent
 	}
 
-	parentDR := parent.Spec.(*networking.DestinationRule)
+	if reflect.DeepEqual(parent.sourceRules, child.sourceRules) {
```
ramaraochavali, line 99: Is it possible to avoid reflect.DeepEqual?

pilot/pkg/model/push_context.go
```diff
@@ -115,14 +116,14 @@ type destinationRuleIndex struct {
 	exportedByNamespace map[string]*consolidatedDestRules
 	rootNamespaceLocal  *consolidatedDestRules
 	// mesh/namespace dest rules to be inherited
-	inheritedByNamespace map[string]*config.Config
+	inheritedByNamespace map[string]*consolidateDestRule
```
ramaraochavali, line 119: nit: consolidatedDestRule

pilot/pkg/model/push_context.go
```diff
@@ -115,14 +116,14 @@ type destinationRuleIndex struct {
...
 }
 
 func newDestinationRuleIndex() destinationRuleIndex {
 	return destinationRuleIndex{
 		namespaceLocal:       map[string]*consolidatedDestRules{},
 		exportedByNamespace:  map[string]*consolidatedDestRules{},
-		inheritedByNamespace: map[string]*config.Config{},
+		inheritedByNamespace: map[string]*consolidateDestRule{},
```
ramaraochavali, line 126: same here

pilot/pkg/model/push_context.go
```diff
@@ -252,7 +253,13 @@ type consolidatedDestRules struct {
 	// Map of dest rule host to the list of namespaces to which this destination rule has been exported to
 	exportTo map[host.Name]map[visibility.Instance]bool
 	// Map of dest rule host and the merged destination rules for that host
-	destRule map[host.Name]*config.Config
+	destRule map[host.Name]*consolidateDestRule
```
ramaraochavali, line 256: same

pilot/pkg/model/push_context.go
```diff
@@ -252,7 +253,13 @@ type consolidatedDestRules struct {
...
+}
+
+// consolidateDestRule represents a dr and from which it is consolidated.
+type consolidateDestRule struct {
```
ramaraochavali, line 260: consolidatedDestRules already exists. So it would be very confusing to have this struct. I think DestRuleConsolidationSource may be better. If you agree, change sourceRules -> fromRules?

ramaraochavali (commented):
some comments

ramaraochavali:
Also can you please add a description to the PR?

hzxuzhonghu:
Make sense, it should include a release note

pilot/pkg/model/push_context.go
hzxuzhonghu: consolidatedDestRules is a wrapper of rules from a single namespace. That one is plural.

How about calling `from`?

ramaraochavali: from makes sense

ramaraochavali:
LGTM will leave it for @howardjohn to take a look

pilot/pkg/model/destination_rule.go
```diff
@@ -80,37 +83,56 @@ func (ps *PushContext) mergeDestinationRule(p *consolidatedDestRules, destRuleCo
...
+func (ps *PushContext) inheritDestinationRule(parent, child *consolidatedDestRule) *consolidatedDestRule {
 	if parent == nil {
 		return child
 	}
 	if child == nil {
 		return parent
 	}
 
-	parentDR := parent.Spec.(*networking.DestinationRule)
+	if reflect.DeepEqual(parent.from, child.from) {
```
howardjohn, line 99: can we use a more efficient check for equality?

Especially concerned that this is an ordered list - does order really matter here?

ramaraochavali: https://github.com/istio/istio/pull/38088#discussion_r835743783 - that was my concern as well

hzxuzhonghu: The order is determined, we have sorted all the drs before merge

hzxuzhonghu: I can compare them one by one to improve efficiency

hzxuzhonghu: But not sure how much it improves, since i can see many places in k8s use reflect.DeepEqual to compare labels.
I have never tested the two ways. Either make sense to me.

hzxuzhonghu: improved

ramaraochavali (approved):

hzxuzhonghu:
@howardjohn Can you ptal

howardjohn (approved):

howardjohn:
does this impact CDS cache as well?

hzxuzhonghu:
It influences a little, previously the cds cache can not be right in some merging cases

GregHanson:
@hzxuzhonghu was this issue only present in 1.13?  Or does this need to be back ported to 1.14 as well?

hzxuzhonghu:
Yes, only exists in version <=1.13

hzxuzhonghu:
Also this should combine with https://github.com/istio/istio/pull/39730

PR #38094, opened 2022-03-23
Add istioctl proxy sample

### LINKED ISSUE #37217: Istiodless istioctl ps command broken when more than one remote istiod
### Bug Description

The `istioctl ps` command needs to aggregate the results from ALL istiod instances. In the non-remote istiod case, `istioctl` does this by sending requests to all the istiod pods in the cluster.

This is not possible when we have an external control plane because all `istioctl` has is the discovery endpoint which delegates requests to only ONE of the istiod instances. As a result `istioctl ps` only returns a partial answer whenever there is more than one istiod serving requests remotely.

We will need some kind of "aggregate status" API that istioctl can call where the target instance calls other instances and aggregate the result before returning, or something like that?

### Version

```prose
every version
```

### Additional Information

_No response_

frankbu:

Example proxy server needed to support istioctl running on a remote cluster.

Fixes #37217

Associated doc PR: https://github.com/istio/istio.io/pull/11099


howardjohn (commented):
This is pretty cool. I am not sure it belongs in samples/ though. Seems like this is not a sample, but a real component that someone deploys into their cluster. I would recommend hosting as a standalone project in istio-ecosystem OR to make this a first class component in Istio - strongly prefer the first though. WDYT

frankbu:
@howardjohn Yeah, I think you're right. I started out thinking this would be an example of what needs to be done, but as I progressed, it started looking more like a component people could just use. I'm not sure it really fits well in istio-ecosystem because it's not really much of a project, just a simple server. It seems like maybe it should be co-located with the istioctl code.

Would you be OK with merging it now in samples and I will take the TODO to move it elsewhere in a future PR. I'd like to make it, and the [associated blog post](https://github.com/istio/istio.io/pull/11099), available asap.

howardjohn:
I think this is exactly in line with istio-ecosystem, a lot of similar scoped projects are there. What I want to avoid is this server being the responsibility of the Istio project. Or if we do want that, then we will need to get buy in from WGs, performance testing, security review, ... which I don't think we want?

costinm:
I'm not sure about this... What I would like to see instead is the real implementation - even if it is as a go binary - of a HBONE proxy that allows forwarding to a specific pod, and supporting it as a first-class feature. It's the main missing feature for multi-network - Istiod is not special, just another stateful set in this case.

We can keep this as a sample - but I don't  know if we should promote it as ecosystem. 

Another option would be to link this code into Istiod itself, as a sort of 'federated' support, so one Istiod can also return the aggregated response ( it can also be an option in the request itself, to be able to get the data from one instance or cascade ).

frankbu:
@howardjohn @costinm I like the idea of linking this into istiod and making it an optional feature, but I don't have the cycles to drive that work right now. Are you OK with merging this PR as a sample for now? I will open an issue to move the functionality to Istio proper in a future PR.

I agree with @howardjohn that it's not "really" an Istio sample, but then either are some of the other components in the samples/ dir (e.g. httpbin service), so it doesn't seem like a total mismatch to me.

howardjohn:
I think something is only really a sample if there is a production
alternative.

What is the issue with istio-ecosystem? It's as simple as `git push`ing the
code

linsun:
https://github.com/istio-ecosystem/istioctl-proxy-sample is created for this

frankbu:
Moved to https://github.com/istio-ecosystem/istioctl-proxy-sample

PR #38100, opened 2022-03-24
use same ipv6 detection logic every where

ramaraochavali:
Currently there is inconsistency in how proxy ipv6 detection works. This PR unifies it and uses the same logic every where.


ramaraochavali:
@hzxuzhonghu @howardjohn can you PTAL?

pilot/pkg/util/network/ip.go
```diff
@@ -141,9 +141,26 @@ func ResolveAddr(addr string, lookupIPAddr ...lookupIPAddrType) (string, error)
 	return resolvedAddr, nil
 }
 
-// IsIPv6Proxy check the addresses slice and returns true for all addresses are valid IPv6 address
-// for all other cases it returns false
-func IsIPv6Proxy(ipAddrs []string) bool {
+// IsIPv6 checks the addresses slice and returns true if atleast one of the addresses
+// is a valid IPv6 address, for all other cases it returns false.
+func IsIPv6(ipAddrs []string) bool {
```
howardjohn, line 146: seems we got rid of IsLinkLocalUnicast check?

howardjohn (commented):
Are there user facing changes or just refactoring duplicate code? We just had to revert a very similar PR so a bit concerned

pilot/pkg/util/network/ip.go
ramaraochavali: yes. That was reverted yesterday. I merged the same code here

ramaraochavali:
There should not be user facing changes. But take a close look once.

pilot/cmd/pilot-agent/main.go
```diff
@@ -284,6 +287,8 @@ func initProxy(args []string) (*model.Proxy, error) {
 		proxy.IPAddresses = append(proxy.IPAddresses, localHostIPv4, localHostIPv6)
 	}
 
+	proxy.DiscoverIPVersions()
```
hzxuzhonghu, line 290: I donot read about this, is it safe to move this?

ramaraochavali: Yes. It is safe. It should have been here actually not later so that whenever proxy is initiated, we have ipversions discovered

pilot/pkg/util/network/ip.go
```diff
@@ -152,8 +169,24 @@ func IsIPv6Proxy(ipAddrs []string) bool {
...
+func GlobalUnicastIP(ipAddrs []string) string {
+	for i := 0; i < len(ipAddrs); i++ {
+		addr := net.ParseIP(ipAddrs[i])
+		if addr == nil {
+			// Should not happen, invalid IP in proxy's IPAddresses slice should have been caught earlier,
+			// skip it to prevent a panic.
+			continue
+		}
+		if addr.IsGlobalUnicast() {
+			return addr.String()
```
hzxuzhonghu, line 188: This looks right than before.

hzxuzhonghu (approved):

pilot/cmd/pilot-agent/main.go
zhlsunshine: @ramaraochavali  one question for this change,  how about the situation if both proxy.SupportsIPv4() and proxy.SupportsIPv6() are true, such as in dual-stack k8s cluster?  It seems that proxy.SupportsIPv4() has the higher priority?

ramaraochavali: Yes

pilot/pkg/util/network/ip.go
zhlsunshine: the original logic is that return true only if all IP addresses are IPv6, the new logic is that return true if there is at least one IP address is IPv6. And this function is widely used in many places, I think this change may cause some problem.

ramaraochavali: I verified at call sites to see if there would be a problem and at once place only it could and I handled. Can you please let me know if you see any potential issues?

zhlsunshine: In fact I git clone the repo from master branch, and can not correctly install my Istio with `istio-ingressgateway` which has the same situation before the IsIPv6Proxy enhancement was reverted yesterday.
image

ramaraochavali: Hmm. Means we are missing some test case here. Can you please see if this https://github.com/istio/istio/pull/38177 branch helps you? If not, I will take another look or we can revert this.

zhlsunshine: The same behavior
image

PR #38111, opened 2022-03-24
Update to do chmod on files after copy. Add read for non ELF files.

ericvn:

Running release-builder locally, I ran into a problem:
```
cp: cannot stat 'tests/testdata/certs/README.md': Permission denied
cp: cannot stat 'tests/testdata/certs/cert.crt': Permission denied
cp: cannot stat 'tests/testdata/certs/cert.key': Permission denied
cp: cannot stat 'tests/testdata/certs/default': Permission denied
cp: cannot stat 'tests/testdata/certs/dns': Permission denied
cp: cannot stat 'tests/testdata/certs/generate.sh': Permission denied
cp: cannot stat 'tests/testdata/certs/mountedcerts-client': Permission denied
cp: cannot stat 'tests/testdata/certs/mountedcerts-server': Permission denied
cp: cannot stat 'tests/testdata/certs/pilot': Permission denied
```

The issue is that we are removing the execute attribute on a directory. Update to do a `a+r` to set the read attribute on the files and update to do it on the destination files, not the source.

One possible issue is that if specifying a sub-directory, we only update the read attributes on the subdirectory, and not recursively. We could add the -R and do all the files (since we have already set the read attributes on the ELF files).

### DIFF
--- a/tools/docker-copy.sh
+++ b/tools/docker-copy.sh
@@ -35,9 +35,11 @@ function may_copy_into_arch_named_sub_dir() {
     case ${FILE_INFO} in
       *x86-64*)
         mkdir -p "${DOCKER_WORKING_DIR}/amd64/" && cp -rp "${FILE}" "${DOCKER_WORKING_DIR}/amd64/"
+        chmod 755 "${DOCKER_WORKING_DIR}/amd64/$(basename "${FILE}")"
         ;;
       *aarch64*)
         mkdir -p "${DOCKER_WORKING_DIR}/arm64/" && cp -rp "${FILE}" "${DOCKER_WORKING_DIR}/arm64/"
+        chmod 755 "${DOCKER_WORKING_DIR}/arm64/$(basename "${FILE}")"
         ;;
       *)
         cp -rp "${FILE}" "${DOCKER_WORKING_DIR}"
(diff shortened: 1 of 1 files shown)


tools/docker-copy.sh
```diff
@@ -42,6 +41,7 @@ function may_copy_into_arch_named_sub_dir() {
         ;;
       *)
         cp -rp "${FILE}" "${DOCKER_WORKING_DIR}"
+        chmod 755 "${DOCKER_WORKING_DIR}/$(basename "${FILE}")"
```
howardjohn, line 44: Are we missing L37 and L40?

ericvn: Yes. Fixed.

tools/docker-copy.sh
```diff
@@ -32,16 +32,18 @@ function may_copy_into_arch_named_sub_dir() {
   #   arm64/
   #   amd64/
   if [[ ${FILE_INFO} == *"ELF 64-bit LSB"* ]]; then
-    chmod 755 "${FILE}"
     case ${FILE_INFO} in
       *x86-64*)
         mkdir -p "${DOCKER_WORKING_DIR}/amd64/" && cp -rp "${FILE}" "${DOCKER_WORKING_DIR}/amd64/"
```
ericvn, line 37: Probably could also remove the -r on the cp as these are files.

howardjohn (approved):

PR #38114, opened 2022-03-24
chore: fixes httpbin yaml samples.

jcchavezs:

YAML files aren't right when it comes to indentation, for example simply running `kubectl apply -f https://github.com/istio/istio/blob/master/samples/httpbin/httpbin-gateway.yaml` would return an error 'error converting YAML to JSON: yaml: line 25: mapping values are not allowed in this context' which can be fixed if the list values have the right identation.

### DIFF
--- a/samples/httpbin/httpbin-gateway.yaml
+++ b/samples/httpbin/httpbin-gateway.yaml
@@ -6,12 +6,12 @@ spec:
   selector:
     istio: ingressgateway
   servers:
-  - port:
-      number: 80
-      name: http
-      protocol: HTTP
-    hosts:
-    - "*"
+    - port:
+        number: 80
+        name: http
+        protocol: HTTP
+      hosts:
+        - "*"
 ---
 apiVersion: networking.istio.io/v1alpha3
 kind: VirtualService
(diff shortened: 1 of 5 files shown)


frankbu (commented):
I'm not sure changing this one sample makes sense, since the formatting of every other one of our samples will also have the same issue. When we use the raw version of the file, it works fine: https://raw.githubusercontent.com/istio/istio/master/samples/httpbin/httpbin-gateway.yaml

jcchavezs:
I can change all samples or even better lint all yaml files in this repo
and have a check on CI. How does that look like?

tor. 24. mar. 2022, 21:34 skrev Frank Budinsky ***@***.***>:

howardjohn (changes requested):
The old indentation is fine, you are trying to apply HTML:

```
$ curl https://github.com/istio/istio/blob/master/samples/httpbin/httpbin-gateway.yaml -s | head






<!DOCTYPE html>
<html lang="en" data-color-mode="auto" data-light-theme="light" data-dark-theme="dark" >
  <head>
    <meta charset="utf-8">
```

PR #38116, opened 2022-03-24
gateway-api: run conformance tests

howardjohn:
This imports the conformance tests from gateway-api.

This involved updating the API import which changed some things, but otherwise pretty straightforward.


common/scripts/setup_env.sh
```diff
@@ -185,6 +185,7 @@ if [[ "${FOR_BUILD_CONTAINER:-0}" -eq "1" ]]; then
   # Override variables with container specific
   TARGET_OUT=${CONTAINER_TARGET_OUT}
   TARGET_OUT_LINUX=${CONTAINER_TARGET_OUT_LINUX}
+  REPO_ROOT=/work
```
costinm, line 188: Nit: it's usually good to use something like REPO_ROOT=${REPO_ROOT:-/work} so it can be overriden. It used to be possible to build Istio without a docker.

costinm (approved):

common/scripts/setup_env.sh
howardjohn: You can, this is specifically if its already detected you are in the docker container. I do most stuff outside docker as well. I will revert this from here, this line was added in another PR (added back - it was mistakenly removed, it has been there for years)

howardjohn:
I think this will fail until https://github.com/istio/istio/pull/38118 lands, but otherwise ready to go

nmittler (approved):

Issue #36971, opened 2022-01-24
auto create GatewayClass and IngressClass


howardjohn:
Logic should be that if it doesn't exist, create it. If it already exists, do not mess with it (so user can update it, etc).

rsalmond:
Dupe? #36718

PR #38118, opened 2022-03-24
gateway-api: auto-create GatewayClass if not present

howardjohn:
For https://github.com/istio/istio/issues/36971

Will help with https://github.com/istio/istio/pull/38116, although not strictly required (we can create it before we run)


ruigulala (approved):

costinm (approved):
We should also try to get the GW test suite run by CI/CD...

howardjohn:
> We should also try to get the GW test suite run by CI/CD...

Doing that after this lands (https://github.com/istio/istio/pull/38116)

PR #38122, opened 2022-03-25
experimental crane builder

howardjohn:
This introduces a new builder using the go-containerregistry package to efficiently construct images without docker at all. There are lot of details on how this works in the comments

Testing incremental build+push of a binary with `hyperfine --show-output -w 1 -r 5 -- 'echo var _ = \"\" >> pilot/cmd/pilot-discovery/main.go; echo var _ = \"\" >> pilot/cmd/pilot-agent/main.go; ./tools/docker  --push --targets=pilot --builder=crane'`

Docker builder:
```
  Time (mean ± σ):     10.581 s ±  0.372 s    [User: 4.366 s, System: 1.582 s]
  Range (min … max):   10.171 s … 11.181 s    5 runs
```

Crane builder:
```
  Time (mean ± σ):      5.207 s ±  0.178 s    [User: 4.532 s, System: 1.375 s]
  Range (min … max):    4.931 s …  5.386 s    5 runs
```

Note most of this time is spent in the binary builder; in the docker runs image pushing is ~6s and ~1s in the crane builder. There are a few other optimizations in upstream libraries that are planned that will bring the crane build to be sub-second even pushing all 10+ docker images we have (on modern hardware that supports sha256 acceleration).


pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -415,7 +422,7 @@ spec:
           name: istio-vm-bootstrap
         {{- range $name, $value := $subset.Annotations }}
         {{- if eq $name.Name "sidecar.istio.io/bootstrapOverride" }}
-        - mountPath: /etc/istio/custom-bootstrap
+        - mountPath: /etc/istio-custom-bootstrap
```
douglas-reid, line 425: just curious: what drives this change?

howardjohn: A fair question. Before, we had the `/etc/istio/config/mesh` file in the original docker image; now its added at runtime. But there is some conflict with the mount that behaves differently for these. I don't fully understand why, TBH, but this purely impacts tests with our fake VMs so I wasn't too worried

ericvn (commented):
LGTM. A few small nits.

ericvn:
One thing that I will mention is that crane when run from a build container on a Mac, ends up trying to use the hosts key store to get the registry credentials. Since we aren't replacing there normal build, and only used during testing, I don't think this will have a large impact. The current Istio common-files code copies the credentials into the container's config.json which in the past has been found and not tried failing back to the credsStore. I'll look at seeing about removing the CredStores line in the host's config which solves that issue. Secondarily, it might be good to add a warning in the case that no creds are found (Mac only?) and or check the credential store as well.

howardjohn:
/retest

@ericvn should be ready to go now

ericvn (approved):
LGTM. Thanks.

kyessenov (approved):
telemetry file

Issue #35310, opened 2021-09-22
Enable ip6tables interception in dual-stack clusters


emike922:
**Describe the feature request**
ip6tables should be enabled or at least be *possible* to enable/override (via annotation/env/CLI arg) if the host/pod has both IPv4 AND IPv6 addresses.
I realize this will not make Istio support dual-stack automagically, but it is the first prerequisite to work/develop towards this feature.

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Rough Solution Proposal**
```diff
diff --git a/tools/istio-iptables/pkg/cmd/root.go b/tools/istio-iptables/pkg/cmd/root.go
index 9942604847..cfaaa0a06e 100644
--- a/tools/istio-iptables/pkg/cmd/root.go
+++ b/tools/istio-iptables/pkg/cmd/root.go
@@ -63,12 +63,12 @@ var rootCmd = &cobra.Command{
                        iptConfigurator.run()
                }
                if cfg.RunValidation {
-                       hostIP, err := getLocalIP()
+                       hostIPs, err := getLocalIPs()
                        if err != nil {
                                // Assume it is not handled by istio-cni and won't reuse the ValidationErrorCode
                                panic(err)
                        }
-                       validator := validation.NewValidator(cfg, hostIP)
+                       validator := validation.NewValidator(cfg, hostIPs, hasIPv6(hostIPs))
 
                        if err := validator.Run(); err != nil {
                                handleErrorWithCode(err, constants.ValidationErrorCode)
@@ -124,12 +124,12 @@ func constructConfig() *config.Config {
                cfg.ProxyGID = cfg.ProxyUID
        }
 
-       // Detect whether IPv6 is enabled by checking if the pod's IP address is IPv4 or IPv6.
-       podIP, err := getLocalIP()
+       // Detect whether IPv6 is enabled by checking if the pod's IP addresses contain IPv6.
+       podIPs, err := getLocalIPs()
        if err != nil {
                panic(err)
        }
-       cfg.EnableInboundIPv6 = podIP.To4() == nil
+       cfg.EnableInboundIPv6 = hasIPv6(podIPs)
 
[...]

PR #38125, opened 2022-03-25
fix issue#35310 to enable ip6tables interception in dual-stack clusters

zhlsunshine:
Fix [issue#35310](https://github.com/istio/istio/issues/35310) to enable ip6tables interception in dual-stack clusters.

### DIFF
--- a/pilot/pkg/util/network/ip_test.go
+++ b/pilot/pkg/util/network/ip_test.go
@@ -229,6 +229,11 @@ func TestIsIPv6Address(t *testing.T) {
 			addr:     net.ParseIP("2001:db8::68"),
 			expected: true,
 		},
+		{
+			name:     "ipv6 linklocalunicast",
+			addr:     net.ParseIP("fe80::9656:d028:8652:66b6"),
+			expected: false,
+		},
 	}
 	for _, tt := range tests {
 		result := IsIPv6Address(tt.addr)
--- a/pilot/pkg/util/network/ip_test.go
+++ b/pilot/pkg/util/network/ip_test.go
@@ -212,3 +212,28 @@ func TestIsIPv6Proxy(t *testing.T) {
 		}
 	}
 }
+
+func TestIsIPv6Address(t *testing.T) {
+	tests := []struct {
+		name     string
+		addr     net.IP
+		expected bool
+	}{
+		{
+			name:     "ipv4",
+			addr:     net.IPv4(1, 2, 3, 4),
+			expected: false,
+		},
+		{
+			name:     "ipv6",
+			addr:     net.ParseIP("2001:db8::68"),
+			expected: true,
+		},
+	}
+	for _, tt := range tests {
+		result := IsIPv6Address(tt.addr)
+		if result != tt.expected {
+			t.Errorf("Test %s failed, expected: %t got: %t", tt.name, tt.expected, result)
+		}
+	}
+}
--- a/tools/istio-iptables/pkg/cmd/root.go
+++ b/tools/istio-iptables/pkg/cmd/root.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
+	"istio.io/istio/pilot/pkg/util/network"
 	"istio.io/istio/tools/istio-iptables/pkg/capture"
 	"istio.io/istio/tools/istio-iptables/pkg/config"
 	"istio.io/istio/tools/istio-iptables/pkg/constants"
@@ -105,6 +106,11 @@ var configureRoutesCommand = &cobra.Command{
 	},
 }
 
+// mock net.InterfaceAddrs to make its unit test become available
+var (
+	LocalIPAddrs = net.InterfaceAddrs
+)
+
 func constructConfig() *config.Config {
 	cfg := &config.Config{
 		DryRun:                  viper.GetBool(constants.DryRun),
@@ -157,15 +163,13 @@ func constructConfig() *config.Config {
 		cfg.ProxyGID = cfg.ProxyUID
 	}
 
-	// Detect whether IPv6 is enabled by checking if the pod's IP address is IPv4 or IPv6.
+	// Detect whether IPv6 is enabled by checking if the pod's IP address is IPv4, IPv6 or both
 	podIPs, err := getLocalIP()
 	if err != nil {
 		panic(err)
 	}
 	for _, podIP := range podIPs {
-		// need to check that a proxy can have an IPv6 address but configuration is not configured K8s for dual-stack support.
-		// In this case an ipv6 link local address will appear, but not one that is routable to with K8s
-		if podIP.To4() == nil && podIP.To16() != nil && !podIP.IsLinkLocalUnicast() {
+		if network.IsIPv6Address(podIP) {
 			cfg.EnableInboundIPv6 = true
 			break
 		}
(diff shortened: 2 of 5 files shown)


tools/istio-iptables/pkg/cmd/root.go
```diff
@@ -157,11 +158,18 @@ func constructConfig() *config.Config {
...
-	podIP, err := getLocalIP()
+	podIPs, err := getLocalIP()
 	if err != nil {
 		panic(err)
 	}
-	cfg.EnableInboundIPv6 = podIP.To4() == nil
+	for _, podIP := range podIPs {
+		// need to check that a proxy can have an IPv6 address but configuration is not configured K8s for dual-stack support.
+		// In this case an ipv6 link local address will appear, but not one that is routable to with K8s
+		if podIP.To4() == nil && podIP.To16() != nil && !podIP.IsLinkLocalUnicast() {
```
jacob-delgado, line 168: This is done so much in the code base that maybe it should be moved to its own function.

jacob-delgado (commented):
LGTM, but can you add any tests?

tools/istio-iptables/pkg/cmd/root.go
```diff
@@ -157,11 +158,18 @@ func constructConfig() *config.Config {
 	}
 
 	// Detect whether IPv6 is enabled by checking if the pod's IP address is IPv4 or IPv6.
-	podIP, err := getLocalIP()
+	podIPs, err := getLocalIP()
```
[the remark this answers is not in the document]
zhlsunshine: okay, done

tools/istio-iptables/pkg/cmd/root.go
```diff
@@ -178,18 +186,23 @@ func constructConfig() *config.Config {
 }
 
 // getLocalIP returns the local IP address
-func getLocalIP() (net.IP, error) {
+func getLocalIP() ([]net.IP, error) {
```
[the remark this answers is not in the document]
zhlsunshine: done

tools/istio-iptables/pkg/cmd/root.go
zhlsunshine: okay, good idea. done

zhlsunshine:
> LGTM, but can you add any tests?

Thanks @jacob-delgado, sure, I added some unit test.

howardjohn (changes requested):
I am not sure I understand this change. Shouldn't this be the very last IPv6 PR?\

Currently, IPv4 is captured and works and IPv6 is bypassed. That is not ideal.

However, with this PR IPv4 is captured and works, and IPv6 is broken. That seems worse than the current state?

pilot/pkg/util/network/ip_test.go
```diff
@@ -212,3 +212,28 @@ func TestIsIPv6Proxy(t *testing.T) {
...
+		{
+			name:     "ipv4",
+			addr:     net.IPv4(1, 2, 3, 4),
+			expected: false,
+		},
+		{
+			name:     "ipv6",
+			addr:     net.ParseIP("2001:db8::68"),
+			expected: true,
+		},
```
ycai-aspen, line 231: add test case with linklocalunicast address

zhlsunshine:
> I am not sure I understand this change. Shouldn't this be the very last IPv6 PR?\
> 
> Currently, IPv4 is captured and works and IPv6 is bypassed. That is not ideal.
> 
> However, with this PR IPv4 is captured and works, and IPv6 is broken. That seems worse than the current state?

Hi @howardjohn, 
Q: Shouldn't this be the very last IPv6 PR?
A: I change the last IPv6 PR, because I draw a common function `IsIPv6Address`  according to the comment [here](https://github.com/istio/istio/pull/38125#discussion_r835706032)

With this PR, IPv6 traffic can be captured as well, and it should go through the sidecar if Istio supports IPv6 the same as IPv4, not just be bypassed. Assume that there are IPv6 only client and server services in the same mesh and dual-stack k8s cluster, however, their traffic always be bypassed and can not be managed by Istio, that will make some features of Istio are unavailable for them, does it make sense?

howardjohn:
Let me clarify a bit. I sent `curl -6 google.com`.

Before this PR: it works.
After this PR: it fails.

This is a regression which will break users.

It doesn't even really have to do with k8s at all, you can send that same
request in an ipv4 only cluster.

zhlsunshine:
> 

```
Before this PR: it works.
After this PR: it fails.
```
In fact, the root cause is that the IPv6 configuration for the pod's proxy is incorrectly configured when the pod's service is deployed with IPv6 first in ip families. Such as [IPv6] or [IPv6, IPv4].
So the first step is to enable the IPv6's iptable rule, then correct the configuration.

howardjohn:
The first step cannot be "break all IP by traffic".

The number one priority is to not break existing users. Any PR that breaks
existing behavior should not be merged

zhlsunshine:
> The first step cannot be "break all IP by traffic". The number one priority is to not break existing users. Any PR that breaks existing behavior should not be merged

Okay, I will submit again to make sure that there is no any break for user and also include all necessary change to support IPv6

zhlsunshine:
> The first step cannot be "break all IP by traffic". The number one priority is to not break existing users. Any PR that breaks existing behavior should not be merged

Hi @howardjohn, I re-submitted again with the IsIPv6Proxy enhancement reverted, sorry for last PR which caused the trouble. And I also completed the user case tests. Please help to check it when available. Thanks!

pilot/pkg/util/network/ip_test.go
zhlsunshine: done

howardjohn:
With this PR I still see the same symptom (curl -s -6 -v google.com fails). It works without this PR.

tools/istio-iptables/pkg/cmd/root.go
```diff
@@ -156,12 +163,17 @@ func constructConfig() *config.Config {
...
-	podIP, err := getLocalIP()
+	// Detect whether IPv6 is enabled by checking if the pod's IP address is IPv4, IPv6 or both
+	podIPs, err := getLocalIP()
 	if err != nil {
 		panic(err)
 	}
-	cfg.EnableInboundIPv6 = podIP.To4() == nil
+	for _, podIP := range podIPs {
+		if network.IsIPv6Address(podIP) {
+			cfg.EnableInboundIPv6 = true
```
ycai-aspen, line 173: To avoid current behavior of not routing ipv6 in iptable for dual stack, should add check it's IPv6 only system, then enable ipv6 inbound.

zhlsunshine: Yeah,  this place is the root cause for command `curl -s -6 -v google.com` fails, but I need to support both families, not dual-stack support, so I will go on to work on the code change for the whole purpose.

zhlsunshine: you can refer to [PR#38042](https://github.com/istio/istio/pull/38042) for more detail

zhlsunshine:
According to some experiments, I think it may not be a good time to work on this PR, so I will close it at first. And please refer to [PR#38042 ](https://github.com/istio/istio/pull/38042) for details. Thanks!

leerin-ruby:
Hi @zhlsunshine ,
I am trying the istio dualstack and found the istio-iptables with --run-validation, it only verifies one ip address not both IPv4 and IPv6. Seems this PR can resolve this scenario. Will you continue work on this? thanks!

zhlsunshine:
Hi @leerin-ruby, part of changes in this PR had been merged in master branch, but it keeps the logic of only validating one ip address.

PR #38141, opened 2022-03-26
refactor xds code

ramaraochavali:
This PR does the following
- make some fields of connection unexported and renamed
-  Rename `ShouldProcessRequest` to `handleHealthRequest` and call it only for healthInfo types. This is to avoid repeated deltatosotw request conversion unnecessarily.
-  Move WaitForRequestLimit after authenticate so that we do not check or wait on request limit if the client is unauthenticated.
- Move `pushDeltaXds` to xdsgen to align with `pushXds`
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pilot/pkg/xds/ads.go
```diff
@@ -188,7 +188,8 @@ func (s *DiscoveryServer) receive(con *Connection, identities []string) {
 // handles 'push' requests and close - the code will eventually call the 'push' code, and it needs more mutex
 // protection. Original code avoided the mutexes by doing both 'push' and 'process requests' in same thread.
 func (s *DiscoveryServer) processRequest(req *discovery.DiscoveryRequest, con *Connection) error {
-	if !s.shouldProcessRequest(con.proxy, req) {
+	if req.TypeUrl == v3.HealthInfoType {
```
ramaraochavali, line 191: Handle HealthInfo explicitly to avoid unnecessary sotw->delta conversion for other request types

pilot/pkg/xds/ads.go
```diff
@@ -273,6 +269,11 @@ func (s *DiscoveryServer) Stream(stream DiscoveryStream) error {
 		log.Debugf("Unauthenticated XDS: %s", peerAddr)
 	}
 
+	if err := s.WaitForRequestLimit(stream.Context()); err != nil {
```
ramaraochavali, line 272: Move this after authenticate

pilot/pkg/xds/xdsgen.go
```diff
@@ -155,6 +157,118 @@ func (s *DiscoveryServer) pushXds(con *Connection, w *model.WatchedResource, req
 	return nil
 }
 
+// Push an Delta XDS resource for the given connection. Configuration will be generated
+// based on the passed in generator.
+func (s *DiscoveryServer) pushDeltaXds(con *Connection,
```
howardjohn, line 162: why move out of Delta code?

howardjohn (commented):
Makes sense to me

ramaraochavali:
Thought xdsgen is common place for both ads and delta push logic similar
to how pushXds is outside of ads.go. I can revert that change if you think
it is better to be in delta.go

pilot/pkg/xds/ads.go
```diff
@@ -68,15 +68,15 @@ type DeltaDiscoveryClient = discovery.AggregatedDiscoveryService_DeltaAggregated
...
+	peerAddr string
 
 	// Time of connection, for debugging
-	Connect time.Time
+	connectedAt time.Time
 
-	// ConID is the connection identifier, used as a key in the connection table.
+	// identifier is the connection identifier, used as a key in the connection table.
 	// Currently based on the node name and a counter.
-	ConID string
```
hzxuzhonghu, line 79: Suggest not changing this long-existing name

pilot/pkg/xds/ads.go
hzxuzhonghu: ? This assumes Authenticate is not a burden

pilot/pkg/xds/ads.go
```diff
@@ -641,11 +642,8 @@ func (s *DiscoveryServer) computeProxyState(proxy *model.Proxy, request *model.P
 	}
 }
 
-// shouldProcessRequest returns whether or not to continue with the request.
-func (s *DiscoveryServer) shouldProcessRequest(proxy *model.Proxy, req *discovery.DiscoveryRequest) bool {
-	if req.TypeUrl != v3.HealthInfoType {
-		return true
-	}
+// handleHealthRequest processes HealthInformation type Url.
+func (s *DiscoveryServer) handleHealthRequest(proxy *model.Proxy, req *discovery.DiscoveryRequest) {
```
hzxuzhonghu, line 646: how about handleWorkloadHealthcheck

pilot/pkg/xds/xdsgen.go
hzxuzhonghu: I think before delta stable we should keep it separate， after it is mature we can merge it with Stow, so reuse some code

pilot/pkg/xds/ads.go
ramaraochavali: Ok. I can revert but thought this is not used out side, it is OK to change

pilot/pkg/xds/ads.go
ramaraochavali: Yes. Also it does not make sense to wait for actual request limit, if the client is not authenticated.

hzxuzhonghu: I have a concern if the auth calls kube-apiserver for token review

ramaraochavali: I think request ratelimit was introduced for xds requests.

ramaraochavali: ok. reverted that change

ramaraochavali:
@hzxuzhonghu PTAL when you get chance

pilot/pkg/xds/delta.go
```diff
@@ -79,6 +74,11 @@ func (s *DiscoveryServer) StreamDeltas(stream DeltaDiscoveryStream) error {
 		deltaLog.Debugf("Unauthenticated XDS: %v", peerAddr)
 	}
 
+	if err := s.WaitForRequestLimit(stream.Context()); err != nil {
```
hzxuzhonghu, line 77: revert

ramaraochavali: I did revert it. Looks like file not saved. reverted now

howardjohn (approved):

Issue #36365, opened 2021-12-03
multicluster: problems when modifying remote secrets


stevenctl:
### Overlapping Secrets

Current behavior: Duplicate service registries.

Why we need to fix this: 
* Support migrating from one access method to another
* Better behavior in the case of accidental misconfig

We need to either ignore the new cluster, or treat this as an update. This should be conditional based on some “priority”.

The priority needs to be deterministic, so new istiod instances are consistent in which cluster-access secret they use. 
* Timestamp
* Priority Annotation
* Both: this is probably ideal. By default, the oldest secret (created timestamp, not update timestamp) takes precedence, but the user can force it to switch over by setting some “priority” annotation? 

### Updating a remote secret 

Current behavior: Remove then Add. This means we will delete then re-add a ton of endpoints. Could lead to a load-spike on the local cluster. 

Instead we should try to sync the new registry before we start reading from it. This has some implementation challenges because of the way we update endpoint shards. Possible solves:
* Add some generic EndpointCache that each ServiceDiscovery holds rather than storing the endpoint shards on DiscoveryServer.
* During the sync period, don’t update the main endpoint shards; once it finishes we go through and swap all the shards to use the new endpoints. 

What happens in the case that the newer registry fails to sync? Should we continue to use the old but (potentially) healthy one? 

### Removing a remote secret

As pointed out in #36310, it’s possible that after our Cleanup removes endpoint shards, events are processed that write to endpoint shards. These will never be cleaned up. 

The fix: 
* Stop processing events, and wait for an event that is currently being processed to finish before cleaning up endpoint shards. 
* We should also avoid using ServiceLister and instead just cleanup any shards matching the service registry’s shard key. 
* This cleanup needs to be implemented for cross-cluster-workload entry registry. We can probably make it generic.  

cc @hzxuzhonghu @ramaraochavali @nmittler @howardjohn

stevenctl:
Just wanted to get some feedback before I go about implementing some of the proposed fixes..

hzxuzhonghu:
### Overlapping Secrets

I think respecting old secrets and ignore new one could be ok, the new one can be seen as a misconfig.

### Updating a remote secret

this could be tricky, i am all for a new way. And I think we still need to update the cache in the backend, just to make the procedure graceful. I guess we could stand pushing some old xds compared to load spike

stevenctl:
not stale

stevenctl:
https://github.com/istio/istio/issues/37925

[...]

PR #38143, opened 2022-03-27
refactor endpointshards index into model

stevenctl:
1. Allows passing a reference to the formerly names EndpointShardsByService to generators living outside `xds` package. 
2. Prework for https://github.com/istio/istio/issues/36365. This will help in creating some way to write to one set of shards and swap in the new ones after sync (in updating remote secret). 

not 100% convinced this is needed for the latter case; hope the cleanup is worthwhile regardless. Only concern is some indirection that preserves some concurrency cache invalidation behavior.

### DIFF
--- a/pilot/pkg/model/endpointshards.go
+++ b/pilot/pkg/model/endpointshards.go
@@ -15,9 +15,7 @@
 package model
 
 import (
-	"fmt"
 	"istio.io/istio/pkg/config/schema/gvk"
-	"strings"
 	"sync"
 
 	"istio.io/istio/pilot/pkg/serviceregistry/provider"
@@ -31,24 +29,15 @@ type shardRegistry interface {
 	Provider() provider.ID
 }
 
-func NewShardKey(cluster cluster.ID, provider provider.ID) ShardKey {
-	return ShardKey(fmt.Sprintf("%s/%s", cluster, provider))
-}
-
 // ShardKeyFromRegistry computes the shard key based on provider type and cluster id.
 func ShardKeyFromRegistry(instance shardRegistry) ShardKey {
-	return NewShardKey(instance.Cluster(), instance.Provider())
+	return ShardKey{Cluster: instance.Cluster(), Provider: instance.Provider()}
 }
 
 // ShardKey is the key for EndpointShards made of a key with the format "cluster/provider"
-type ShardKey string
-
-func (sk ShardKey) Cluster() cluster.ID {
-	p := strings.Split(string(sk), "/")
-	if len(p) < 1 {
-		return ""
-	}
-	return cluster.ID(p[0])
+type ShardKey struct {
+	Cluster  cluster.ID
+	Provider provider.ID
 }
 
 // EndpointShards holds the set of endpoint shards of a service. Registries update
--- a/pilot/pkg/model/endpointshards.go
+++ b/pilot/pkg/model/endpointshards.go
@@ -81,6 +81,7 @@ func (e *EndpointIndex) SetCache(cache XdsCache) {
 	e.cache = cache
 }
 
+// must be called with lock
 func (e *EndpointIndex) clearCacheForService(svc, ns string) {
 	if e.cache == nil {
 		return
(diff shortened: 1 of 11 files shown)


hzxuzhonghu:
I think it's valuable to make `model` package exportable to other projects, even though I think this is not a must for target 2.

pilot/pkg/model/endpointshards.go
```diff
@@ -0,0 +1,199 @@
...
+
+// ShardKeyFromRegistry computes the shard key based on provider type and cluster id.
+func ShardKeyFromRegistry(instance shardRegistry) ShardKey {
+	return NewShardKey(instance.Cluster(), instance.Provider())
+}
+
+// ShardKey is the key for EndpointShards made of a key with the format "cluster/provider"
+type ShardKey string
+
+func (sk ShardKey) Cluster() cluster.ID {
```
howardjohn, line 46: This is called really often. Would it make sense to make ShardKey a `struct{cluster, provider}` so this is faster?

pilot/pkg/model/endpointshards.go
```diff
@@ -0,0 +1,199 @@
...
+		for nsKey, v := range v {
+			out[svcKey][nsKey] = v
+		}
+	}
+	return out
+}
+
+// ShardsForService returns the shards and true if they are found, or returns nil, false.
+func (e *EndpointIndex) ShardsForService(serviceName, namespace string) (*EndpointShards, bool) {
+	e.mu.Lock()
```
howardjohn, line 122: RLock?

pilot/pkg/model/endpointshards.go
```diff
@@ -0,0 +1,199 @@
...
+	for _, shard := range epShards.Shards {
+		for _, ep := range shard {
+			if ep.ServiceAccount != "" {
+				epShards.ServiceAccounts.Insert(ep.ServiceAccount)
+			}
+		}
+	}
+	// Clear the cache here to avoid race in cache writes.
+	e.clearCacheForService(serviceName, namespace)
+	epShards.Unlock()
```
howardjohn, line 190: why can we unlock before the delete?

howardjohn (commented):
Overall makes sense

pilot/pkg/xds/debug.go
```diff
@@ -297,9 +297,7 @@ func (s *DiscoveryServer) registryz(w http.ResponseWriter, req *http.Request) {
 // the full push.
 func (s *DiscoveryServer) endpointShardz(w http.ResponseWriter, req *http.Request) {
 	w.Header().Add("Content-Type", "application/json")
-	s.mutex.RLock()
-	out, _ := json.MarshalIndent(s.EndpointShardsByService, " ", " ")
-	s.mutex.RUnlock()
+	out, _ := json.MarshalIndent(s.EndpointIndex.Shardz, " ", " ")
```
howardjohn, line 300:
```suggestion
	out, _ := json.MarshalIndent(s.EndpointIndex.Shardz(), " ", " ")
```

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -278,25 +278,15 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
...
-	keys := make([]model.ShardKey, 0, len(shards.Shards))
-	for k := range shards.Shards {
-		keys = append(keys, k)
-	}
-	if len(keys) >= 2 {
-		sort.Slice(keys, func(i, j int) bool {
-			return keys[i] < keys[j]
-		})
-	}
+	shards.RLock()
```
howardjohn, line 281: is RLock sufficient? on L317 we mutate the endpoint

howardjohn (approved):

pilot/pkg/model/endpointshards.go
```diff
@@ -0,0 +1,216 @@
...
+	return keys
+}
+
+// EndpointIndex is a mutex protected index of endpoint shards
+type EndpointIndex struct {
+	mu sync.RWMutex
+	// keyed by svc then ns
+	shardsBySvc map[string]map[string]*EndpointShards
+	// We'll need to clear the cache in-sync with endpoint shards modifications.
+	cache XdsCache
```
hzxuzhonghu, line 95: I am not sure why cache need in sync with endpoint shards?

I am asking because this makes it hard to maintain

hzxuzhonghu: If this is coupled with xds cache, i donot think it should be moved to `model`

hzxuzhonghu: A struct in model should be a simple prototype, and can be reused, if coupled with xds cache, this is absolutely not general

PR #38149, opened 2022-03-28
move cache key build outside of critical section

ramaraochavali:
Some of the cache keys (implementation of Key()) has some non trivial logic. So better to build it outside of mutex to avoid locking xds cache for longer duration.

- [ ] Docs
- [ ] Installation
- [ X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu (approved):

blizard863 (commented):
=

PR #38150, opened 2022-03-28
PushRequest start time merge

andrewshaoyu:
when two push request merged , PushRequest.Start should keep the older one & non-zero one;

| t0 | t1 | result |
| -- | -- | -- |
| time.Now() | t0.Add(time.Minute) | t1 |
| nil | time.Now() | t1 |
| time.Now() | nil | t0 |

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -448,16 +448,16 @@ func (pr *PushRequest) CopyMerge(other *PushRequest) *PushRequest {
 		reason = append(reason, pr.Reason...)
 		reason = append(reason, other.Reason...)
 	}
-
+	startTime := pr.Start
 	if pr.Start.IsZero() {
-		pr.Start = other.Start
+		startTime = other.Start
 	} else if !other.Start.IsZero() && pr.Start.After(other.Start) {
-		pr.Start = other.Start
+		startTime = other.Start
 	}
 
 	merged := &PushRequest{
 		// Keep the first (older) start time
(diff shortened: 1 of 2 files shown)


howardjohn:
Is this fixing a bug? I didn't think we ever needed to merge start time

andrewshaoyu:
> Is this fixing a bug? I didn't think we ever needed to merge start time

when  merge two PushRequest, Now we use Pr.start. But I think when Pr.start is zero,it's not meaningful. We can choose the meaningful time as start time. At last when PushRequest was sent to xds client, the delta time from start are useful. Instead of set start time as time.Now() during `startPush`

howardjohn (changes requested):
I don't think we ever set Start before we merge, always after.

This field is extremely sensitive and caused 2 bugs in the past few months. If we are going to change it I would like to see concrete bugs that it fixes, rather than unproven refactorings

andrewshaoyu:
I won't insist on this PR. Now I set start as time.Now in kube informer handler , then pass to xxxhandler. At last I can get the time that every event used from being watched to being sent to envoy. Then I can have an overview about the whole xds chain.

howardjohn:
I see. The concern is Start is tightly coupled with push context cache so
it is actually become something very important for correctness

hzxuzhonghu:
Yes, in istio the Start is always set. We heavily rely on it.

andrewshaoyu:
I will take a look at latest version.

PR #38158, opened 2022-03-28
add enable install env var to be able to skip installation entirely

ZhengzheYang:
Change-Id: I9e303287bbdb82c22b40590fef9092e77b7d9df4

add another env var for CNI installation to skip the install entirely.

### DIFF
--- a/cni/pkg/config/config.go
+++ b/cni/pkg/config/config.go
@@ -39,6 +39,8 @@ type InstallConfig struct {
 	CNINetworkConfigFile string
 	// CNI config template string
 	CNINetworkConfig string
+	// Whether to install CNI configuration and binary files
+	CNIEnableInstall bool
 	// Whether to reinstall CNI configuration and binary files
 	CNIEnableReinstall bool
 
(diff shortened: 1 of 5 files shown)


cni/pkg/install/install.go
```diff
@@ -213,6 +217,10 @@ func sleepCheckInstall(ctx context.Context, cfg *config.InstallConfig, cniConfig
 
 // checkInstall returns an error if an invalid CNI configuration is detected
 func checkInstall(cfg *config.InstallConfig, cniConfigFilepath string) error {
+	// If the installation is skipped, don't check for invalid configurations.
+	if !cfg.CNIEnableInstall {
```
bianpengyuan, line 221: I feel it is probably more direct if we just make line 93 `!EnableInstall && !EnableReinstall`

ZhengzheYang: This line is still needed, otherwise the container's readiness check won't pass since it sees invalid configurations all the time and won't call `SetReady()`. Added the EnableInstall && EnableReinstall check to line 93. PTAL.

cni/pkg/config/config.go
```diff
@@ -40,6 +40,8 @@ type InstallConfig struct {
 	// CNI config template string
 	CNINetworkConfig string
 	// Whether to reinstall CNI configuration and binary files
```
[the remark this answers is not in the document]
ZhengzheYang: Done.

ericvn:
Interesting that this went in with no release notes. Is the env var supposed to be hidden?

bianpengyuan:
@ericvn it is to support one of our customers need. I don't expect this option to be used other than some really special use case. Let me know if you think this is really necessary to have a release notes.

PR #38170, opened 2022-03-29
Remove bad sample

frankbu:

Remove sample because it doesn't work on Mac or Windows, it only works on Linux.


zirain:
this sample is very useful for local development, can you keep them but add some warning message?

litong01:
I use it everyday for my dev and test. I think it is very useful. I do not mind at all to add some warnings if detected in non Linux env.

frankbu:
@litong Can you send a PR to add another prereq check in the script to make sure it's running on a platform that will work, and also add a warning in the README? I'm not sure it belongs in samples/ anyway, maybe better to move somewhere else (e.g., istio-ecosystem), but for now let's at least make sure that people don't run it on Mac or Windows just to find out later it isn't working when they try to use it.

frankbu:
We could also add something in the wiki about it, but the script, itself, still needs to live somewhere. I guess leaving it in samples is probably best, since it's there already :-)

frankbu:
Closing in favour of #38181

craigbox:
Just to check: _why_ doesn't this work on Mac or Windows?  `kind` is available there, is it `metallb` that's the problem?  I've seen a few posts talking about workarounds: https://www.thehumblelab.com/kind-and-metallb-on-mac/, for example

PR #38175, opened 2022-03-30
Convert deprecated filter names to canonical one

hzxuzhonghu:

Currently we can get a deprecated name in listener.  Below is an example

```
kind: EnvoyFilter
metadata:
  name: size-limit
  namespace: istio-system
spec:
  workloadSelector:
    labels:
      istio: ingressgateway
  configPatches:
    - applyTo: HTTP_FILTER
      match:
        context: GATEWAY
        listener:
          filterChain:
            filter:
              name: envoy.filters.network.http_connection_manager
      patch:
        operation: INSERT_BEFORE
        value:
          name: envoy.buffer
          typed_config:
            '@type': type.googleapis.com/udpa.type.v1.TypedStruct
            value:
              maxRequestBytes: 1073741824
```

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/envoyfilter/listener_patch.go
+++ b/pilot/pkg/networking/core/v1alpha3/envoyfilter/listener_patch.go
@@ -428,15 +428,18 @@ func patchHTTPFilters(patchContext networking.EnvoyFilter_PatchContext,
 			IncrementEnvoyFilterMetric(lp.Key(), HttpFilter, false)
 			continue
 		}
+		// clone instead of mutating the patch directly, otherwise it may influence other proxy
+		httpFilterPatch := proto.Clone(lp.Value).(*hcm.HttpFilter)
+		httpFilterPatch.Name = toCanonicalName(httpFilterPatch.Name)
 		if lp.Operation == networking.EnvoyFilter_Patch_ADD {
 			applied = true
-			httpconn.HttpFilters = append(httpconn.HttpFilters, proto.Clone(lp.Value).(*hcm.HttpFilter))
+			httpconn.HttpFilters = append(httpconn.HttpFilters, httpFilterPatch)
 		} else if lp.Operation == networking.EnvoyFilter_Patch_INSERT_FIRST {
-			httpconn.HttpFilters = append([]*hcm.HttpFilter{proto.Clone(lp.Value).(*hcm.HttpFilter)}, httpconn.HttpFilters...)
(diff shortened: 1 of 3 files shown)


ramaraochavali:
https://github.com/istio/istio/blob/da6178604559bdf2c707a57f452d16bee0de90c8/releasenotes/notes/envoy-filter.yaml - We mentioned we will not supported deprecated filter names going forward. I am wondering it is a continuous maintenance headache to covert back and forth.

hzxuzhonghu:
Yes,  i totally agree. But we can not eliminate this immediately, otherwise could break backward compatibility if a name is removed from the envoy. So this is better to have

ramaraochavali:
Should we add a feature flag to support old names so that the hit is taken by people who use it and we can default it to false and we can remove that flag couple of releases down the line?

howardjohn:
I am not sure I understand the motivation here. There should be no expectation that EnvoyFilters are compatible on upgrade

> NOTE 1: Some aspects of this API are deeply tied to the internal implementation in Istio networking subsystem as well as Envoy’s XDS API. While the EnvoyFilter API by itself will maintain backward compatibility, any envoy configuration provided through this mechanism should be carefully monitored across Istio proxy version upgrades, to ensure that deprecated fields are removed and replaced appropriately.

From https://istio.io/latest/docs/reference/config/networking/envoy-filter/

I am actually glad that we break compatiblity here; similar to the reasons in https://sre.google/sre-book/service-level-objectives/#xref_risk-management_global-chubby-planned-outage, it sets expectations that we **will** break EnvoyFilters. This is especially needed for the next few releases where I expect the internal XDS shape will change quite a bit.

hzxuzhonghu:
Do both of you feel we should remove the canonical name convert since we have the notes?

hzxuzhonghu:
Normally support deprecated name for some releases and removing it totally is ok. But as you can see, still many users are using the deprecated name.  This is the motivation here. 

If both of you agree, I am ok with completely clean up all the temporary conversions.

howardjohn:
Why not just keep things the way they are? Then we do not need to expand the workarounds but also do not break existing users?

hzxuzhonghu:
I could not understand this philosophy.

hzxuzhonghu:
With this theory, i doubt what's the effect of the deprecation release note.  Normally, we will announce the deprecation of something and then after some releases remove them, otherwise the deprecation stuff will become more and more. And eventually it would be more difficult to clean up

PR #38178, opened 2022-03-30
Revert "use same ipv6 detection logic every where"

ramaraochavali:
Reverts istio/istio#38100


ramaraochavali:
@howardjohn can you please approve this? I think definitely there are some missing test cases here. But I would like keep master sane while I investigate the issue.

Issue #38009, opened 2022-03-18
VM tests: refactor which OS's we tests


howardjohn:
Currently:

- ubuntu_xenial
- ubuntu_bionic
- ubuntu_focal
- debian_9
- debian_10
- centos_8
- centos_7

problem: we do not have full coverage of new versions, but adding a bunch of new ones will bloat things
proposal: carefully craft the set we use to maximise coverage will reducing tests

Proposal:

- ubuntu_xenial # test oldest ubuntu, ensure old GLIBC works
- ubuntu_jammy # test latest ubuntu
- debian_11 # test one debian, xenial covers older ones
- centos_8_stream # test latest centos
- centos_7 # test old centos (old GLIBC)

This drops us from 7 -> 5 cases while expanding the coverage (IMO - a bit subjective)

howardjohn:
@ericvn @stevenctl any concern?

mayyyyying:
> 

Hi Howard,

Do we have any plan to use the ubuntu jammy instead of focal as the base image, because there are bunch of VUL in focal, or did we already tested the latest ubuntu jammy can be safely upgraded from focal? any concern about this upgrading?

PR #38182, opened 2022-03-30
vm tests: update and deprcated OS's tested

howardjohn:
Fixes https://github.com/istio/istio/issues/38009

This changes the set of OS's we test to improve coverage while decreasing test count.

Before merging, I will need to build the new bases; I put a hold to do that. Will do so after approval.

There were a number of related fixes to make this (easily) work; comments inline describing why.

Testing: we only test all of these in postsubmit; I manually tested.

### DIFF
--- a/pkg/test/framework/components/echo/kube/deployment.go
+++ b/pkg/test/framework/components/echo/kube/deployment.go
@@ -613,13 +613,19 @@ func GenerateService(cfg echo.Config) (string, error) {
 
 var VMImages = map[echo.VMDistro]string{
 	echo.UbuntuXenial: "app_sidecar_ubuntu_xenial",
-	echo.UbuntuFocal:  "app_sidecar_ubuntu_focal",
-	echo.UbuntuBionic: "app_sidecar_ubuntu_bionic",
-	echo.Debian9:      "app_sidecar_debian_9",
-	echo.Debian10:     "app_sidecar_debian_10",
+	echo.UbuntuJammy:  "app_sidecar_ubuntu_jammy",
+	echo.Debian11:     "app_sidecar_debian_11",
 	echo.Centos7:      "app_sidecar_centos_7",
-	echo.Centos8:      "app_sidecar_centos_8",
+	echo.Rockylinux8:  "app_sidecar_rockylinux_8",
 }
+
+var RevVMImages = func() map[string]echo.VMDistro {
+	r := map[string]echo.VMDistro{}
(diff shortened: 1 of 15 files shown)


pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -613,14 +613,20 @@ func GenerateService(cfg echo.Config) (string, error) {
...
-	echo.Debian9:      "app_sidecar_debian_9",
-	echo.Debian10:     "app_sidecar_debian_10",
+	echo.UbuntuJammy:  "app_sidecar_ubuntu_jammy",
+	echo.Debian11:     "app_sidecar_debian_11",
 	echo.Centos7:      "app_sidecar_centos_7",
-	echo.Centos8:      "app_sidecar_centos_8",
+	echo.Rockylinux8:  "app_sidecar_rockylinux_8",
 }
 
+var RevVMImages = func() map[string]echo.VMDistro {
```
howardjohn, line 622: I think previously we just always tested DefaultVMDistro repeatedly... fixing this

tools/build-base-images.sh
```diff
@@ -23,13 +23,23 @@ ROOT=$(dirname "$WD")
...
+}
+
 # shellcheck source=prow/lib.sh
 source "${ROOT}/prow/lib.sh"
 buildx-create
 
 HUBS="${HUBS:?specify a space seperated list of hubs}"
 TAG="${TAG:?specify a tag}"
-DOCKER_TARGETS="${DOCKER_TARGETS:-base,distroless,app_sidecar_base_debian_9,app_sidecar_base_debian_10,app_sidecar_base_ubuntu_xenial,app_sidecar_base_ubuntu_bionic,app_sidecar_base_ubuntu_focal,app_sidecar_base_centos_7,app_sidecar_base_centos_8}"
+defaultTargets="$(cat "${ROOT}/tools/docker.yaml" | toJson | toJson | jq '[.images[] | select(.base) | .name] | join(",")' -r)"
```
howardjohn, line 41: auto-derive

tools/docker-builder/main.go
```diff
@@ -126,6 +126,27 @@ func CopyInputs(a Args) error {
 	return nil
 }
 
+func ReadPlanTargets() ([]string, []string, error) {
```
howardjohn, line 129: So we can automatically derive default targets

ericvn (approved):

PR #38186, opened 2022-03-30
TLS Configuration API (Phase 1)

lei-tang:
Issue: https://github.com/istio/api/issues/2285.

This PR implements the phase 1 TLS configuration API, including its unit tests and integration tests.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [X] Security
- [ ] Test and Release
- [ ] User Experience


tests/integration/security/tls_config/tls_config_test.go
```diff
@@ -0,0 +1,160 @@
...
+			err = checkTLS13Allowed(kubeConfig, retry, apps.Namespace1, fromPod, "istio-proxy", target)
+			if err != nil {
+				t.Errorf("checkTLS13Allowed() returns error: %v", err)
+			} else {
+				t.Logf("as expected, TLS 1.3 is allowed")
+			}
+		})
+}
+
+func checkTLS11Disallowed(kubeConfig string, retry tutil.Retrier, ns namespace.Instance, fromPod, fromContainer, connectTarget string) error {
```
yangminzhu, line 108: there seems quite some common part in the following 3 check functions, could probably refactor to share the common part.

pilot/pkg/security/authn/utils/utils.go
```diff
@@ -69,11 +71,29 @@ func BuildInboundTLS(mTLSMode model.MutualTLSMode, node *model.Proxy,
...
+func configureTLSVersion(minTLSVer meshconfig.MeshConfig_TLSConfig_TLSProtocol, tlsParam *tls.TlsParameters) {
+	// The min TLS version is at least TLS 1.2.
+	tlsParam.TlsMinimumProtocolVersion = tls.TlsParameters_TLSv1_2
+	// If user configures the min TLS version to be TLS 1.3, the min TLS version will be TLS 1.3.
+	switch minTLSVer {
+	case meshconfig.MeshConfig_TLSConfig_TLSV1_3:
+		tlsParam.TlsMinimumProtocolVersion = tls.TlsParameters_TLSv1_3
+	}
+	// The max TLS version is configured as TLS 1.3
+	tlsParam.TlsMaximumProtocolVersion = tls.TlsParameters_TLSv1_3
```
yangminzhu, line 98: this changes the behavior to always set the max TLS version to 1.3, the old code does not set it, is this intended?

pilot/pkg/security/authn/utils/utils.go
```diff
@@ -69,11 +71,29 @@ func BuildInboundTLS(mTLSMode model.MutualTLSMode, node *model.Proxy,
...
+		minTLSVersion = mc.GetMeshMTLS().MinProtocolVersion
+	}
+	configureTLSVersion(minTLSVersion, ctx.CommonTlsContext.TlsParams)
 
 	authn_model.ApplyToCommonTLSContext(ctx.CommonTlsContext, node, []string{}, /*subjectAltNames*/
 		trustDomainAliases, ctx.RequireClientCertificate.Value)
 	return ctx
 }
+
+func configureTLSVersion(minTLSVer meshconfig.MeshConfig_TLSConfig_TLSProtocol, tlsParam *tls.TlsParameters) {
```
yangminzhu, line 89: The code could be cleaner if you just pass the `mc` and return the min/max TLS version directly from the function, like `getTLSMinMaxVersion(mc *meshconfig.MeshConfig) (minVersion TlsParameters_TlsProtocol, maxVersion TlsParameters_TlsProtocol)`

tests/integration/security/tls_config/tls_config_test.go
lei-tang: I refactored 3 check functions into 1 check function.

pilot/pkg/security/authn/utils/utils.go
lei-tang: Yes, this is intended.

pilot/pkg/security/authn/utils/utils.go
lei-tang: The original function is refactored into getTLSVersions(mcMinTLSVer meshconfig.MeshConfig_TLSConfig_TLSProtocol) (tls.TlsParameters_TlsProtocol, tls.TlsParameters_TlsProtocol).

pilot/pkg/security/authn/utils/utils.go
```diff
@@ -69,11 +71,30 @@ func BuildInboundTLS(mTLSMode model.MutualTLSMode, node *model.Proxy,
...
+	minVersion := tls.TlsParameters_TLSv1_2
+	// If user configures the min TLS version to be TLS 1.3, the min TLS version will be TLS 1.3.
+	switch mcMinTLSVer {
+	case meshconfig.MeshConfig_TLSConfig_TLSV1_3:
+		minVersion = tls.TlsParameters_TLSv1_3
+	}
+	// The max TLS version is configured as TLS 1.3
+	maxVersion := tls.TlsParameters_TLSv1_3
+	return minVersion, maxVersion
+}
```
howardjohn, line 100:
```suggestion

	ctx.CommonTlsContext.TlsParams.TlsMinimumProtocolVersion,
		ctx.CommonTlsContext.TlsParams.TlsMaximumProtocolVersion = getTLSVersions(mc.GetMeshMTLS().GetMinProtocolVersion())
	authn_model.ApplyToCommonTLSContext(ctx.CommonTlsContext, node, []string{}, /*subjectAltNames*/
		trustDomainAliases, ctx.RequireClientCertificate.Value)
	return ctx
}

func getTLSVersions(ver meshconfig.MeshConfig_TLSConfig_TLSProtocol) (tls.TlsParameters_TlsProtocol, tls.TlsParameters_TlsProtocol) {
	switch ver {
	case meshconfig.MeshConfig_TLSConfig_TLSV1_3:
		return tls.TlsParameters_TLSv1_3, tls.TlsParameters_TLSv1_3
	default:
		return tls.TlsParameters_TLSv1_2, tls.TlsParameters_TLSv1_3
	}
}
```

Can simplify this a lot

tests/integration/security/tls_config/tls_config_test.go
```diff
@@ -0,0 +1,135 @@
...
+	var expResult string
+	if allow {
+		expResult = "allowed"
+	} else {
+		expResult = "disallowed"
+	}
+
+	retryFn := func(_ context.Context, i int) error {
+		execCmd := fmt.Sprintf(
+			"kubectl exec %s -c %s -n %s --kubeconfig %s -- openssl s_client -alpn istio %s -connect %s",
```
howardjohn, line 122: Depending on `openssl` in the container is not safe since we support open source. if you want to get the TLS certifiates from the handshake, the echo client can do this with the `TLS` protocol.

tests/integration/security/tls_config/main_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+func setupConfig(_ resource.Context, cfg *istio.Config) {
+	if cfg == nil {
+		return
+	}
+	cfg.ControlPlaneValues = `
+values:
+  meshConfig:
+    meshMTLS:
+      minProtocolVersion: TLSV1_3
```
howardjohn, line 55: I don't think we need a whole new test suite for this feature. A test suite is extremely expensive - about 2 min runtime. If we added one for every little flag in mesh config we would have tests that take a day to complete.

howardjohn (changes requested):

pilot/pkg/security/authn/utils/utils.go
lei-tang: The PR has been revised to simplify getTLSVersions(). The nil checkings of mc != nil && mc.GetMeshMTLS() != nil are needed because mc and mc.GetMeshMTLS() may be nil. For example, mc or mc.GetMeshMTLS() are nil in some existing tests and they crash without the nil checkings.

howardjohn: `GetMeshMTLS` (and any protobuf Get) handles nil

lei-tang: The getTLSVersions() in the comment failed at lint test with the following unparam error. The lint error is due to the second return parameter is always tls.TlsParameters_TLSv1_3. To fix the lint error, the original getTLSVersions() is used.

{pilot/pkg/security/authn/utils/utils.go:89:102: getTLSVersions - result 1 (github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3.TlsParameters_TlsProtocol) is always 4  : getTLSVersions - result 1 (github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3.TlsParameters_TlsProtocol) is always 4
Category: unparam
pilot/pkg/security/authn/utils/utils.go
Line: 89

howardjohn: Why not just not have the function return the max tls version?

All the code needs to do is get the mesh setting or a default - there is no need to make it 50 lines of complex code

tests/integration/security/tls_config/tls_config_test.go
lei-tang: The openssl is used here to verify whether a specific TLS version is allowed/disallowed during handshake through the -tls1_1, -tls1_2, and -tls1_3 flags of openssl s_client command. It is not used to get TLS certificates. AFAIK, the echo client does not support verifying a specific TLS version is allowed/disallowed during handshake.

tests/integration/security/tls_config/main_test.go
lei-tang: The integration test in the PR uses meshConfig to configure the mesh-wide minimum TLS version to be 1.3. When testing a configuration through meshConfig, the convention of all existing integration tests is as follows:
1. In TestMain(), call framework.NewSuite(m). Setup(..., setupConfig)).
2. In setupConfig(), configure the meshConfig.

The integration test in the PR follows this convention. If the meshConfig in the PR is merged into an existing integration test, it changes this existing test since the existing test does not configure minimum TLS version to be 1.3.

pilot/pkg/security/authn/utils/utils.go
lei-tang: Agreed. I have revised the function based on the comments.

tests/integration/security/tls_config/main_test.go
howardjohn: I would rather have no test than this test as is - adding this much runtime for such a small feature is not viable.

That being said, mesh config can be simply patched at runtime, there are a number of tests that do this: https://github.com/istio/istio/blob/1ade7ec2e569c2302d90a5f6b2bd86872e2314ad/tests/integration/security/normalization_test.go#L227 for example.

tests/integration/security/tls_config/tls_config_test.go
lei-tang: Based on the comments, the integration test has been removed.

tests/integration/security/tls_config/main_test.go
lei-tang: Based on the comments, the integration test has been removed.

pilot/pkg/security/authn/v1beta1/policy_applier.go
```diff
@@ -149,19 +150,26 @@ func (a *v1beta1PolicyApplier) AuthNFilter(forSidecar bool) *http_conn.HttpFilte
 func (a *v1beta1PolicyApplier) InboundMTLSSettings(endpointPort uint32, node *model.Proxy, trustDomainAliases []string) plugin.MTLSSettings {
 	effectiveMTLSMode := a.GetMutualTLSModeForPort(endpointPort)
 	authnLog.Debugf("InboundFilterChain: build inbound filter change for %v:%d in %s mode", node.ID, endpointPort, effectiveMTLSMode)
+	var mc *meshconfig.MeshConfig
+	if a.push != nil {
```
yangminzhu, line 154: nit: Could even simplify to call getMinTLSVersion() here (Only need to call once) and then pass the result `TlsParameters_TlsProtocol` directly to BuildInboundTLS().

yangminzhu (commented):
Looks good, just a small comment.

yangminzhu (approved):

stevenctl (approved):

PR #38187, opened 2022-03-30
updated istiod pilot to make async calls to jwksuri

aryan16:
1) Update the Istiod pilot logic to make async calls to jwksuri and removed the calls from main flow. 
2) Updated unit tests to support the new way of fetching public key from jwksuri


pilot/pkg/model/jwks_resolver.go
```diff
@@ -242,16 +245,12 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
 	if jwksURI == "" {
 		// Fetch the jwks URI if it is not hardcoded on config.
 		jwksURI, err = r.resolveJwksURIUsingOpenID(issuer)
+		key.jwksURI = jwksURI
```
howardjohn, line 248: using the return value (jwksURI) when an `err` was returned is not a great practice IMO. If you are intentionally setting it to `""` on err, then I would make it explicit.

pilot/pkg/model/jwks_resolver.go
```diff
@@ -242,16 +245,12 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
 	if jwksURI == "" {
 		// Fetch the jwks URI if it is not hardcoded on config.
 		jwksURI, err = r.resolveJwksURIUsingOpenID(issuer)
```
howardjohn, line 247: I am a bit confused. I thought this PR was to make the resolve async, but this seems to be synchronous?

pilot/pkg/model/jwks_resolver.go
```diff
@@ -266,17 +265,35 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
...
+		// var err error
 
 		// jwtKeyResolver should never be nil since the function is only called in Discovery Server request processing
 		// workflow, where the JWT key resolver should have already been initialized on server creation.
-		jwtPubKey, err = r.GetPublicKey(jwtIssuer, jwksURI)
-		if err != nil {
-			log.Errorf("Failed to fetch jwt public key from issuer %q, jwks uri %q: %s", jwtIssuer, jwksURI, err)
-			// This is a temporary workaround to reject a request with JWT token by using a fake jwks when istiod failed to fetch it.
-			// TODO(xulingqing): Find a better way to reject the request without using the fake jwks.
+		_, _ = r.GetPublicKey(jwtIssuer, jwksURI)
```
howardjohn, line 272: Ignoring both return values on a method called "Get" seems very confusing. if its purpose is something else, maybe it should be renamed like "update cache" or something more appropriate?

pilot/pkg/model/jwks_resolver.go
```diff
@@ -266,17 +265,35 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
...
+		key := jwtKey{issuer: jwtIssuer, jwksURI: jwksURI}
+		if val, found := r.keyEntries.Load(key); found {
+			e := val.(jwtPubKeyEntry)
+			if e.pubKey == "" {
+				log.Infof("no public key present in the cache, adding the key to the cache")
```
howardjohn, line 277: We lost the comments on why we use the fake jwks. Also the log says adding to cache but I don't see where we do that

pilot/pkg/model/jwks_resolver.go
```diff
@@ -384,7 +401,7 @@ func (r *JwksResolver) refresher() {
 	for {
 		select {
 		case <-r.refreshTicker.C:
-			currentHasError := r.refresh()
+			currentHasError := r.refresh(networkFetchRetryCountOnRefreshFlow)
```
howardjohn, line 404: if we are doing multiple retries here, then it seems like the refreshInterval modifications below will not be respected as expected anymore.

Also, we may be blocking here for a long time and not allow `jwksuriChannel` part to make progress

pilot/pkg/model/jwks_resolver.go
```diff
@@ -405,20 +422,30 @@ func (r *JwksResolver) refresher() {
 		case <-closeChan:
 			r.refreshTicker.Stop()
 			return
+		case jwtKeyData := <-jwksuriChannel:
+			hasErrors := true
+			for i := 0; i < 10; i++ {
```
howardjohn, line 427: why are we looping 10x?

pilot/pkg/model/jwks_resolver_test.go
```diff
@@ -144,13 +151,20 @@ func TestGetPublicKeyReorderedKey(t *testing.T) {
 	}
 	for _, c := range cases {
 		pk, err := r.GetPublicKey(c.in[0], c.in[1])
+		key := jwtKey{issuer: c.in[0], jwksURI: c.in[1]}
+		time.Sleep(time.Millisecond * 100)
```
howardjohn, line 155: can we avoid sleeping? https://github.com/istio/istio/wiki/Test-Flakes#timing-dependencies

aryan16: In middle of changes, will remove this soon.

pilot/pkg/model/jwks_resolver.go
aryan16: Planning the change this.

pilot/pkg/model/jwks_resolver.go
aryan16: We are already doing multiple retries for the refresh, we were not passing it in the method parameter. While making the call we are using these retries - https://github.com/istio/istio/blob/master/pilot/pkg/model/jwks_resolver.go#L454
I am not touching the existing flow instead I have created a new async process in the background.

pilot/pkg/model/jwks_resolver.go
aryan16: Because it's a safe bet. Since this is in the background we are good with this.

pilot/pkg/model/jwks_resolver.go
```diff
@@ -236,48 +242,28 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
...
 		if err != nil {
-			log.Errorf("Failed to fetch jwt public key from issuer %q, jwks uri %q: %s", jwtIssuer, jwksURI, err)
-			// This is a temporary workaround to reject a request with JWT token by using a fake jwks when istiod failed to fetch it.
-			// TODO(xulingqing): Find a better way to reject the request without using the fake jwks.
-			jwtPubKey = CreateFakeJwks(jwksURI)
+			log.Infof("Failed to fetch jwt public key from issuer %q, jwks uri %q: %s", jwtIssuer, jwksURI, err)
 		}
 	}
+	if jwtPubKey == "" {
+		log.Infof("no public key present in the cache, adding the key to the cache")
```
yangminzhu, line 262: the log is wrong, the following statement is returning a fake jwks and not adding it to any cache (at least not here).

pilot/pkg/model/jwks_resolver.go
yangminzhu: could you try to reuse the existing retry logic in `case <-r.refreshTicker.C:` and not duplicating it here?

pilot/pkg/model/jwks_resolver.go
```diff
@@ -384,41 +370,71 @@ func (r *JwksResolver) refresher() {
...
+			if hasErrors {
+				log.Errorf("Failed to get public key from  %q: %q", jwtKeyData.issuer, jwtKeyData.jwksURI)
+			}
+			stopBackgroundRefresh = false
 		}
 	}
 }
 
-func (r *JwksResolver) refresh() bool {
+func (r *JwksResolver) refresh(retryNumber int) bool {
```
yangminzhu, line 429: If you reuse the refresh logic (see comment above), you shouldn't need the `retryNumber` argument.

pilot/pkg/model/jwks_resolver.go
aryan16: If I do that, I need to increase `networkFetchRetryCountOnRefreshFlow` or change the `refresh interval`. Do you think that would be ideal to do that ?

pilot/pkg/model/jwks_resolver.go
```diff
@@ -236,48 +239,28 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
...
 }
 
 // BuildLocalJwks builds local Jwks by fetching the Jwt Public Key from the URL passed if it is empty.
 func (r *JwksResolver) BuildLocalJwks(jwksURI, jwtIssuer, jwtPubKey string) *envoy_jwt.JwtProvider_LocalJwks {
 	if jwtPubKey == "" {
 		var err error
-
 		// jwtKeyResolver should never be nil since the function is only called in Discovery Server request processing
 		// workflow, where the JWT key resolver should have already been initialized on server creation.
 		jwtPubKey, err = r.GetPublicKey(jwtIssuer, jwksURI)
```
hzxuzhonghu, line 253: With async resolve, isn't it always return an empty jwtPubKey?

hzxuzhonghu:
What's the issue it is fixing

pilot/pkg/model/jwks_resolver.go
aryan16: Only if the key is not present in the cache.

hzxuzhonghu: I read it and seems it will not wait for async resolve result, do I miss something

aryan16:
> What's the issue it is fixing

Currently In the main flow we are making one call to jwksuri without any retries and there are cases when jwks server is not fully ready and istiod is trying to make a request or we face some networking issues in between. So we don't want to increase the number of retries in the main flow to resolve this issue. Hence in this PR, we are fetching the public key in the background and directly updating the cache. Once the cache gets updated, we push the key from background flow.

pilot/pkg/model/jwks_resolver.go
aryan16: It won't wait instead we are pushing the key to the envoy in the backgroud.

hzxuzhonghu: How does it push to envoy in the background?

aryan16: I am using the same logic as `refresher` which is always running in the background - https://github.com/istio/istio/blob/master/pilot/pkg/model/jwks_resolver.go#L488

pilot/pkg/model/jwks_resolver.go
```diff
@@ -384,41 +372,49 @@ func (r *JwksResolver) refresher() {
...
-				r.refreshInterval = r.refreshDefaultInterval
-			}
-			lastHasError = currentHasError
-			r.refreshTicker.Reset(r.refreshInterval)
+			lastHasError = r.refreshCache(lastHasError)
 		case <-closeChan:
 			r.refreshTicker.Stop()
 			return
+		case jwtKeyData := <-jwksuriChannel:
+			hasErrors := r.refreshCache(lastHasError)
```
yangminzhu, line 380: should this be `lastHasError = r.refreshCache(lastHasError)`?

pilot/pkg/model/jwks_resolver.go
```diff
@@ -236,48 +239,33 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
...
-		resp, err = r.getRemoteContentWithRetry(jwksURI, networkFetchRetryCountOnMainFlow)
-		if err != nil {
-			log.Errorf("Failed to fetch public key from %q: %v", jwksURI, err)
-		}
-		pubKey = string(resp)
-	}
-
 	r.keyEntries.Store(key, jwtPubKeyEntry{
-		pubKey:            pubKey,
+		pubKey:            "",
```
yangminzhu, line 243: Please also update the comment of `GetPublicKey()`, the function name could also be updated to better reflect the new behavior.

pilot/pkg/model/jwks_resolver.go
```diff
@@ -458,6 +456,7 @@ func (r *JwksResolver) refresh() bool {
 				atomic.AddUint64(&r.refreshJobFetchFailedCount, 1)
 				return
 			}
+			log.Infof("Fetched the key successfully from the jwksuri, adding it to the cache")
```
yangminzhu, line 459: This log is not needed.

pilot/pkg/model/jwks_resolver.go
```diff
@@ -432,7 +428,9 @@ func (r *JwksResolver) refresh() bool {
 		}
 
 		oldPubKey := e.pubKey
-
+		if oldPubKey == "" {
+			oldKeyNull = true
+		}
```
yangminzhu, line 433: In line 441-450, you should update the cache for the jwksURI returned by the resolveJwksURIUsingOpenID() otherwise it will be called every time for an empty jwksURI.

aryan16: one confusion - in the case when jwksuri is not returning the valid public key, do we need to update the jwksuri in the cache with empty public key for next retries ?

pilot/pkg/model/jwks_resolver.go
hzxuzhonghu: IC

pilot/pkg/model/jwks_resolver.go
```diff
@@ -236,48 +240,33 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
...
-	} else {
-		var resp []byte
-		resp, err = r.getRemoteContentWithRetry(jwksURI, networkFetchRetryCountOnMainFlow)
-		if err != nil {
-			log.Errorf("Failed to fetch public key from %q: %v", jwksURI, err)
-		}
-		pubKey = string(resp)
-	}
-
 	r.keyEntries.Store(key, jwtPubKeyEntry{
```
hzxuzhonghu, line 243: Why do you store an empty jwt pub key

aryan16: I am storing it for `jwksuri`, we have to store some entry in the cache so that it will be available in the background.

hzxuzhonghu: So in this case, you can use the fake jwks

aryan16: there might be a case when `jwksuri` itself is not there, we are fetching is as well https://github.com/istio/istio/blob/master/pilot/pkg/model/jwks_resolver.go#L244
We are doing this as well in the background.

Issue #37665, opened 2022-03-02
Telemetry API: Consider adding Traffic Direction selection to Access Logging (and Tracing)


douglas-reid:
**Describe the feature request**

In the original Telemetry API proposals, there were designs for a generic `TelemetryRuleMatch` message that could live inside each domain type (metrics, access logs, tracing) and would allow tailoring of behavior based on the following attributes:
- traffic direction
- protocol
- target port

This was rejected and dropped from the initial API surface during API review. Subsequently, the `metrics` component of the API adopted a similar notion for `traffic_direction` with [`WorkloadMode`](https://istio.io/latest/docs/reference/config/telemetry/#WorkloadMode) inside of a [`MetricSelector`](https://istio.io/latest/docs/reference/config/telemetry/#MetricSelector).

There is no existing way to turn on/off logging (and/or tracing) for a workload based on traffic direction.  Access logging  (and, hopefully soon, Tracing) do provide filtering of behavior based on CEL expressions. Unfortunately, there is no Envoy attribute that covers traffic direction at the moment.

It would be useful to add a `LogSelector` message to the `AccessLogging` message that would allow matching based on traffic direction. The `LogSelector` message could be expanded to support `protocol` and other match criteria in future iterations.

**Describe alternatives you've considered**

- Add Envoy attributes for `traffic_direction` and use CEL filtering support instead. This would require Envoy-level changes and (potentially) push requirements for non-Envoy consumers of the Telemetry API to produce a similar attribute. In the original design proposal, there was a clear preference expressed for having both a structured way and an expression-based approach for this behavior.

**Affected product area (please put an X in all that apply)**

[ X ] Extensions and Telemetry
[ X ] User Experience

cc: @kyessenov @rolandkool @zirain @howardjohn

zirain:
+1 for add `traffic_direction`
In most cases, user will not set 100% sampling in the env, the cost of networking、cpu、storage is expensive.
if there's something wired, we can change the tracing filter do capture what we really want.
I hope telemetry will enable us do something like this:
```yaml
# some special rul
- traffic_direction: OUTBOUND
  url:
    prefix: /foo
  sampling: 100%
# the default rule
- url:
    prefix: /
  sampling: 1%
```

zirain:
@douglas-reid seems we can close this now?

douglas-reid:
Yes. We need to write some integration tests using this functionality (iirc), but I think we can close.

PR #38192, opened 2022-03-31
implement Telemetry LoggingSelector

zirain:

part of https://github.com/istio/istio/issues/37665

### DIFF
--- a/pilot/pkg/model/telemetry.go
+++ b/pilot/pkg/model/telemetry.go
@@ -180,27 +180,14 @@ type TracingSpec struct {
 }
 
 type LoggingConfig struct {
-	Providers []*meshconfig.MeshConfig_ExtensionProvider
-	Filter    *tpb.AccessLogging_Filter
-	Selector  *tpb.AccessLogging_LogSelector
+	Providers     []*meshconfig.MeshConfig_ExtensionProvider
+	ProviderNames sets.Set
+	Filter        *tpb.AccessLogging_Filter
 }
 
-func (cfg *LoggingConfig) MatchListenerClass(class networking.ListenerClass) bool {
-	if cfg.Selector == nil ||
-		class == networking.ListenerClassUndefined {
-		return true
-	}
-
-	switch cfg.Selector.Mode {
-	case tpb.WorkloadMode_CLIENT_AND_SERVER:
-		return true
-	case tpb.WorkloadMode_CLIENT:
-		return class == networking.ListenerClassSidecarOutbound || class == networking.ListenerClassGateway
-	case tpb.WorkloadMode_SERVER:
-		return class == networking.ListenerClassSidecarInbound
-	default:
-		return false
-	}
+type LoggingSpec struct {
+	Client *LoggingConfig
+	Server *LoggingConfig
 }
 
 // AccessLogging returns the logging configuration for a given proxy. If nil is returned, access logs
@@ -211,20 +198,30 @@ func (t *Telemetries) AccessLogging(proxy *Proxy, class networking.ListenerClass
 	if len(ct.Logging) == 0 && len(t.meshConfig.GetDefaultProviders().GetAccessLogging()) == 0 {
 		return nil
 	}
-	cfg := LoggingConfig{}
-	providers, f, s := mergeLogs(ct.Logging, t.meshConfig)
-	cfg.Filter = f
-	cfg.Selector = s
-	if !cfg.MatchListenerClass(class) {
-		return nil
+
+	cfg := mergeLogs(ct.Logging, t.meshConfig)
+
+	var spec *LoggingConfig
+	switch class {
+	case networking.ListenerClassGateway:
+		spec = cfg.Client
+	case networking.ListenerClassSidecarInbound:
+		spec = cfg.Server
+	case networking.ListenerClassSidecarOutbound:
+		spec = cfg.Client
+	case networking.ListenerClassUndefined:
+		// this should not happened, just in case
+		spec = cfg.Client
 	}
-	for _, p := range providers.SortedList() {
+
+	for _, p := range spec.ProviderNames.SortedList() {
 		fp := t.fetchProvider(p)
 		if fp != nil {
-			cfg.Providers = append(cfg.Providers, fp)
+			spec.Providers = append(spec.Providers, fp)
 		}
 	}
-	return &cfg
+
+	return spec
 }
 
 // Tracing returns the logging tracing for a given proxy. If nil is returned, tracing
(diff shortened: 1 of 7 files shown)


pilot/pkg/model/telemetry.go
```diff
@@ -434,6 +461,10 @@ func mergeLogs(logs []*tpb.AccessLogging, mesh *meshconfig.MeshConfig) (sets.Set
 		if m.Filter != nil {
 			loggingFilter = m.Filter
 		}
+
+		if m.Match != nil {
+			loggingSelector = m.Match
```
hzxuzhonghu, line 466: I donot think the merging is right, now this is merging all logging that could be applied to a proxy, but does not care whether the match selectors are same or not.

douglas-reid: Agreed. I think we need something that is aware of selector logic in merging.

Even with a single provider, the following scenario:

mesh-wide: server-side logging enabled, client-side logging disabled
namespace-wide: server-side logging enabled with filter `foo`
workload-specific: client-side logging enabled with filter `bar`

should result in:

server-side logging with filter `foo` and client-side logging with filter `bar`.

This would seem to set to result in only client-side logging with filter `bar`.

I _think_ the `mergeLogs` needs more complex updating as a result.

zirain: so you means we need merge `selector` but not merge `filter`?

zirain: updated, please take a review again @douglas-reid

pilot/pkg/model/telemetry.go
```diff
@@ -410,32 +450,46 @@ func (t *Telemetries) telemetryFilters(proxy *Proxy, class networking.ListenerCl
...
+	cfg := &LoggingSpec{
+		Client: &LoggingConfig{
+			ProviderNames: sets.New(),
+		},
+		Server: &LoggingConfig{
+			ProviderNames: sets.New(),
+		},
+	}
 
+	providerNames := mesh.GetDefaultProviders().GetAccessLogging()
```
hzxuzhonghu, line 464: move after if len(logs) == 0 {

pilot/pkg/model/telemetry.go
```diff
@@ -410,32 +450,46 @@ func (t *Telemetries) telemetryFilters(proxy *Proxy, class networking.ListenerCl
...
+			if m.Filter != nil {
+				s.Filter = m.Filter
+			}
+			s.ProviderNames = sets.New(providerNames...)
 		}
 	}
-	inScopeProviders := sets.New(providerNames...)
 
 	parentProviders := mesh.GetDefaultProviders().GetAccessLogging()
 	for _, m := range logs {
```
hzxuzhonghu, line 495: Seems there are two iterations on same logs, can they be merged?

pilot/pkg/model/telemetry.go
```diff
@@ -410,32 +450,46 @@ func (t *Telemetries) telemetryFilters(proxy *Proxy, class networking.ListenerCl
...
 
-// mergeLogs returns the set of providers for the given logging configuration.
-// This currently is just the names of providers as there is no access logging configuration, but
-// in the future it will likely be extended
-func mergeLogs(logs []*tpb.AccessLogging, mesh *meshconfig.MeshConfig) (sets.Set, *tpb.AccessLogging_Filter) {
-	providers := sets.New()
+// mergeLogs returns the merged configuration for the given telemetry logging configuration.
+func mergeLogs(logs []*tpb.AccessLogging, mesh *meshconfig.MeshConfig) *LoggingSpec {
+	cfg := &LoggingSpec{
+		Client: &LoggingConfig{
```
hzxuzhonghu, line 456: I do not see where you set LoggingConfig.Providers.  Is it useless?

hzxuzhonghu (commented):
Can you add an ut for mergeLogs. it is very hard to understand

pilot/pkg/model/telemetry.go
zirain: see L466, L502

hzxuzhonghu: ProviderNames there

pilot/pkg/model/telemetry.go
zirain: this merged this two
image

pilot/pkg/model/telemetry.go
hzxuzhonghu: IC, the providers are set in the caller, looks weird

pilot/pkg/model/telemetry.go
```diff
@@ -454,6 +469,13 @@ func mergeLogs(logs []*tpb.AccessLogging, mesh *meshconfig.MeshConfig) (sets.Set
 				providers.Delete(provider)
 				continue
 			}
+
+			if m.Match != nil &&
```
howardjohn, line 473: do we need to move this before `providers.Delete(provider)`

pilot/pkg/model/telemetry.go
```diff
@@ -179,16 +179,33 @@ type LoggingConfig struct {
 	Filter    *tpb.AccessLogging_Filter
 }
 
+func getWorkloadMode(class networking.ListenerClass) tpb.WorkloadMode {
```
douglas-reid, line 182: nit: just `workloadMode` is closer to typical go conventions.

pilot/pkg/model/telemetry.go
```diff
@@ -179,16 +179,33 @@ type LoggingConfig struct {
...
+		return tpb.WorkloadMode_CLIENT
+	case networking.ListenerClassUndefined:
+		// this should not happened, just in case
+		return tpb.WorkloadMode_CLIENT
+	}
+
+	return tpb.WorkloadMode_CLIENT
+}
+
 // AccessLogging returns the logging configuration for a given proxy. If nil is returned, access logs
```
douglas-reid, line 198:
```suggestion
// AccessLogging returns the logging configuration for a given proxy and listener class. If nil is returned, access logs
```

pilot/pkg/model/telemetry.go
douglas-reid, line 473: i'm wondering if we should refactor slightly to help readability.  Maybe something like:

```golang
if !logsSelected(m.Match, mode) {
   continue
}

...

func logsSelected(selector *tpb.LoggingSelector, mode tpb.WorkloadMode) bool {
  if selector == nil {
     return true
  }
  if selector.Mode == case tpb.WorkloadMode_CLIENT_AND_SERVER {
    return true
  }
  return selector.Mode == mode
}
```

Thoughts?

douglas-reid: It feels like it does. @zirain ? I think this means we also need a test for:

- mesh-wide: enabled
- ns-wide: disabled

and 

- mesh-wide: enabled
- ns-wide: disabled
- workload: enabled

just to validate we are deleting and re-adding appropriately.

pilot/pkg/model/telemetry_test.go
```diff
@@ -183,76 +226,159 @@ func TestAccessLogging(t *testing.T) {
...
+			[]config.Config{newTelemetry("istio-system", serverAndClient)},
+			networking.ListenerClassSidecarInbound,
+			sidecar,
+			nil,
+			[]string{"envoy"},
+		},
+		{
+			"server and client - outbound",
+			[]config.Config{newTelemetry("istio-system", serverAndClient)},
+			networking.ListenerClassSidecarOutbound,
```
douglas-reid, line 325: can we add a test case(s) for something with multiple telemetry resources at different levels in the hierarchy (mesh wide, ns wide)?

Maybe something like:

```golang
[]config.Config{newTelemetry("istio-system", server), newTelemetry("foo", client)}
networkingListenerClassSidecar{Inbound|Outbound}
```

to validate proper overrides with differing modes?

douglas-reid (commented):
Looking pretty good.  Let's get this in ASAP.

pilot/pkg/model/telemetry_test.go
zirain: done, add tests with multiple telemetry resources.

pilot/pkg/model/telemetry.go
zirain: sounds good, done.

howardjohn (approved):

douglas-reid (approved):

PR #38196, opened 2022-03-31
replace deprecated header match specifier

ramaraochavali:
Replaces deprecated header matcher with string matcher

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali:
looks like gRPC has not yet moved to https://github.com/grpc/grpc-go/blob/d47437c91e798b5370e019660cd74c9c0c8e32fb/internal/xds/rbac/matchers.go#L264 StringMatch for RBAC. I will revert the RBAC changes

pilot/pkg/networking/util/util.go
```diff
@@ -587,6 +590,37 @@ func StringToPrefixMatch(in []string) []*matcher.StringMatcher {
...
+	for _, im := range in {
+		if em := ConvertToEnvoyMatch(im); em != nil {
+			res = append(res, em)
+		}
+	}
+
+	return res
+}
+
+func ConvertToEnvoyMatch(in *networking.StringMatch) *matcher.StringMatcher {
```
hzxuzhonghu, line 605: I would suggest keeping it in route package if it is only used there

ramaraochavali: This is potentially useful for others as well similar to StringToExactMatch. That is why I kept here. Currently it is used only in route

hzxuzhonghu (approved):

Issue #36999, opened 2022-01-25
Unable to mount volumes to ingress gateway pod using istio/gateway helm chart 


freefood89:
### What I Tried
Deployed istio gateway using helm chart istio/gateway (version=1.12.2) using the configs:
```yaml
values:
  podAnnotations:                                                                                       
    sidecar.istio.io/userVolumeMount: |-
      [{ "name": "google-oauth2",  "mountPath": "/etc/istio/config",  "readonly": true }]
    sidecar.istio.io/userVolume: |-
      [{ "name": "google-oauth2", "secret": { "secretName": "istio-oauth2" } }]
```

### What I thought would happen
I expected istio to mutate the pod to add `volumes`/`volumeMounts`

### What actually happened
Istio added all the usual `volumes`/`volumeMounts` but did not add the ones specified in `podAnnotations`. The `template` of the Deployment and the Pod's `annotation` do have `sidecar.istio.io/userVolumeMount` and `sidecar.istio.io/userVolume`. Why isn't the volume being injected? Is the `inject.istio.io/templates: gateway` not allowing this?

### What else I tried
- I was able to confirm that volume mounting works on ordinary sidecar injections (on httpbin) with the same annotations.
- I was able to work around this by manually deploying the gateway with `volumes`/`volumeMounts` in the Deployment. Perhaps these can be exposed on the helm chart?

### Why I need this
This is needed to mount secrets necessary for the oauth2 envoyfilter. I would like to apply this envoyfilter at the gateway level.

### Version

```prose
$ istioctl version
client version: 1.10.2
control plane version: 1.12.2
data plane version: 1.12.2 (6 proxies)

$ kubectl version --short
Client Version: v1.21.3
Server Version: v1.21.5-gke.1302

$ helm version --short
v3.5.3+g041ce5a
```

### Additional Information

(redacted gke cluster name and gcp project name)
$ istioctl bug-report

Target cluster context: ***

Running with the following config: 

istio-namespace: istio-system
full-secrets: false
timeout (mins): 30
include: {  }
exclude: { Namespaces: kube-system, kube-public, kube-node-lease, local-path-storage } AND { Namespaces: kube-system, kube-public, kube-node-lease, local-path-storage }
end-time: 2022-01-25 11:31:19.622872 -1000 HST

The following Istio control plane revisions/versions were found in the cluster:
Revision default:
&version.MeshInfo{
    {
        Component: "pilot",
[...]

PR #38198, opened 2022-03-31
Support mounting volumes to gateway via Helm chart

maxhov:
This PR enables the mounting of volumes to a gateway via the Helm chart 'istio/gateway'. Mounting a volume is for instance necessary to supply the needed tokens for the oauth2 filter, such as described in issue #36999

### DIFF
--- a/manifests/charts/gateway/templates/deployment.yaml
+++ b/manifests/charts/gateway/templates/deployment.yaml
@@ -86,6 +86,10 @@ spec:
             name: http-envoy-prom
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+          {{- with .Values.volumeMounts }}
+          volumeMounts:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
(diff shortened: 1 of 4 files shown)


maxhov:
Not sure how to fix the issue with gencheck_istio. Seems like the makefile in common is having an issue, but the build output does not show what it is. Ran the helm lint check locally, but gave no issues.

zirain:
> Not sure how to fix the issue with gencheck_istio. Seems like the makefile in common is having an issue, but the build output does not show what it is. Ran the helm lint check locally, but gave no issues.

do your run `make gen` local

ristox1:
I'm facing the same issue and thus I'm "bumping" this issue to be reopened if possible :)

> **Please provide a description of this PR:** This PR enables the mounting of volumes to a gateway via the Helm chart 'istio/gateway'. Mounting a volume is for instance necessary to supply the needed tokens for the oauth2 filter, such as described in issue #36999

PR #38200, opened 2022-03-31
Fix the init target so users can use their own Envoy

jwendell:
If users override the variable `ISTIO_ENVOY_LINUX_RELEASE_PATH` in an
attempt to use their own version of Envoy, they will face an error in
the docker build stage, because it expects the binary to be in the
`ISTIO_ENVOY_LINUX_RELEASE_DIR` location (aka
`out/linux_amd64/release`).

This change makes sure we copy it to the expected place.

### DIFF
--- a/bin/init.sh
+++ b/bin/init.sh
@@ -80,6 +80,13 @@ function download_envoy_if_necessary () {
     echo "Copying $2 to $(dirname "$2")/${3}"
     cp -f "$2" "$(dirname "$2")/${3}"
     popd
+
+  else
+    # The Envoy binary already exists; this is the case when user overrides the
+    # ISTIO_ENVOY_LINUX_RELEASE_PATH variable to use their own Envoy.
+    # In this case, we must copy it to the release directory
+    mkdir -p "${ISTIO_ENVOY_LINUX_RELEASE_DIR}"
+    cp -f "$2" "${ISTIO_ENVOY_LINUX_RELEASE_DIR}"
   fi
 }
 


bin/init.sh
```diff
@@ -80,6 +80,13 @@ function download_envoy_if_necessary () {
     echo "Copying $2 to $(dirname "$2")/${3}"
     cp -f "$2" "$(dirname "$2")/${3}"
     popd
+
+  else
+    # The Envoy binary already exists; this is the case when user overrides the
+    # ISTIO_ENVOY_LINUX_RELEASE_PATH variable to use their own Envoy.
+    # In this case, we must copy it to the release directory
+    mkdir -p "${ISTIO_ENVOY_LINUX_RELEASE_DIR}"
+    cp -f "$2" "${ISTIO_ENVOY_LINUX_RELEASE_DIR}"
```
ericvn, line 89: It seems that you are setting the `ISTIO_ENVOY_LINUX_RELEASE_PATH` and not setting `ISTIO_ENVOY_LINUX_RELEASE_DIR` before calling. In that case, `ISTIO_ENVOY_LINUX_RELEASE_PATH` will default to ${TARGET_OUT_LINUX}/release.  In your scenario, setting the DIR as the parent of the PATH (and maybe the we should do that if only the PATH is defined) does not require this change.

jwendell: I see your point. In this case, we would have to update https://github.com/istio/istio/blob/master/tools/docker.yaml#L32 as it hardcodes the path to the sidecar, assuming it's under `${TARGET_OUT_LINUX}/release`.

ericvn: > Or another option would be to always copy from the download or specified PATH to TARGET_OUT_LINUX. That seems strange.

Well, the more I think about it, that's probably not that strange since we copy to TARGET_OUT at the end of the script. And I'm curious why the copy from ISTIOENVOY_NATIVE_PATH isn't working.

jwendell: We rely on files being present on `$ISTIO_OUT/release`. This makes sure we copy the envoy binary when the user overrides it. Overriding `$ISTIO_ENVOY_LINUX_RELEASE_DIR` along with `$ISTIO_ENVOY_LINUX_RELEASE_PATH` might not work well.

In my case, for example, I have `ISTIO_ENVOY_LINUX_RELEASE_PATH=/envoy`. Thus `ISTIO_ENVOY_LINUX_RELEASE_DIR` had to be `/`, which clearly is not correct. So, not touching this DIR variable in general works just fine. With this PR, the only change users need to do to use a custom Envoy is to override a single variable, `ISTIO_ENVOY_LINUX_RELEASE_PATH`.

ericvn (approved):
Based on our discussion, I think this is an acceptable solution. The only minor issue is that the old code would download envoy after a clean, and as long as another clean wasn't done, future init's would have found the file and passed over the download and copying to release subdirectory.  With this change the file is found, and it's copied to release again.

howardjohn (changes requested):
```
$ make init
TARGET_OUT=/work/out/linux_amd64 ISTIO_BIN=/gobin GOOS_LOCAL=linux bin/retry.sh SSL_ERROR_SYSCALL bin/init.sh
Skipping envoy debug. Set DEBUG_IMAGE to download.
cp: '/work/out/linux_amd64/release/envoy-af4ca235bf40fb11829f2f02cf678610641095a2' and '/work/out/linux_amd64/release/envoy-af4ca235bf40fb11829f2f02cf678610641095a2' are the same file
Unexpected failure
make[1]: *** [Makefile.core.mk:205: /work/out/linux_amd64/istio_is_init] Error 1
make: *** [Makefile:44: init] Error 2
```

when I try running now

PR #38140, opened 2022-03-26
Clean up unneeded method

hzxuzhonghu:
**Please provide a description of this PR:**

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #38204, opened 2022-03-31
Restore Set.InsertAll

nmittler:
This was removed by https://github.com/istio/istio/pull/38140, but should not have been. The Insert/InsertAll separation avoids a needless array allocation for the vast majority of Insert calls.


howardjohn (approved):
I think this is a good compromise between the two. I put a hold in case @hzxuzhonghu has concerns?

hzxuzhonghu:
OK. let's have a benchmark test to see the difference.

hzxuzhonghu:
```
goos: linux
goarch: amd64
pkg: istio.io/istio/pkg/util/sets
cpu: Intel(R) Xeon(R) Gold 6161 CPU @ 2.20GHz
BenchmarkSet_Insert-8      	48236504	        34.43 ns/op
BenchmarkSet_InsertOne-8   	47176616	        30.24 ns/op
PASS
ok  	istio.io/istio/pkg/util/sets	3.156s

```

hzxuzhonghu (commented):
LGTM with 10% improve

PR #38218, opened 2022-04-01
use same ipv6 detection logic every where (take 2)

ramaraochavali:
Follow-up to this https://github.com/istio/istio/pull/38100

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/cmd/pilot-agent/main.go
+++ b/pilot/cmd/pilot-agent/main.go
@@ -264,9 +264,6 @@ func initProxy(args []string) (*model.Proxy, error) {
 		proxy.IPAddresses = []string{podIP.String()}
 	}
 
-	// After IP addresses are set, let us discover IPMode.
-	proxy.DiscoverIPMode()
-
 	// Obtain all the IPs from the node
 	if ipAddrs, ok := network.GetPrivateIPs(context.Background()); ok {
 		if len(proxy.IPAddresses) == 1 {
@@ -286,6 +283,9 @@ func initProxy(args []string) (*model.Proxy, error) {
 	if len(proxy.IPAddresses) == 0 {
 		proxy.IPAddresses = append(proxy.IPAddresses, localHostIPv4, localHostIPv6)
 	}
+
+	// After IP addresses are set, let us discover IPMode.
+	proxy.DiscoverIPMode()
 
 	// Extract pod variables.
 	podName := options.PodNameVar.Get()
--- a/pilot/cmd/pilot-agent/main.go
+++ b/pilot/cmd/pilot-agent/main.go
@@ -221,7 +221,7 @@ func initStatusServer(ctx context.Context, proxy *model.Proxy, proxyConfig *mesh
 
 func initStsServer(proxy *model.Proxy, tokenManager security.TokenManager) (*stsserver.Server, error) {
 	localHostAddr := localHostIPv4
-	if network.IsIPv6Proxy(proxy.IPAddresses) {
+	if proxy.IsIPv6() {
 		localHostAddr = localHostIPv6
 	}
 	stsServer, err := stsserver.NewServer(stsserver.Config{
@@ -263,6 +263,9 @@ func initProxy(args []string) (*model.Proxy, error) {
 	if podIP != nil {
 		proxy.IPAddresses = []string{podIP.String()}
 	}
+
+	// After IP addresses are set, let us discover IPMode.
+	proxy.DiscoverIPMode()
 
 	// Obtain all the IPs from the node
 	if ipAddrs, ok := network.GetPrivateIPs(context.Background()); ok {
--- a/pilot/cmd/pilot-agent/options/agent.go
+++ b/pilot/cmd/pilot-agent/options/agent.go
@@ -21,7 +21,6 @@ import (
 
 	meshconfig "istio.io/api/mesh/v1alpha1"
 	"istio.io/istio/pilot/pkg/model"
-	"istio.io/istio/pilot/pkg/util/network"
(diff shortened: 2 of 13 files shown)


ramaraochavali:
@zhlsunshine Can you please try this commit and see if it works for you?

ramaraochavali:
@zhlsunshine gentle ping.. Can you PTAL?

pilot/cmd/pilot-agent/main.go
```diff
@@ -221,7 +221,7 @@ func initStatusServer(ctx context.Context, proxy *model.Proxy, proxyConfig *mesh
 
 func initStsServer(proxy *model.Proxy, tokenManager security.TokenManager) (*stsserver.Server, error) {
 	localHostAddr := localHostIPv4
-	if network.IsIPv6Proxy(proxy.IPAddresses) {
+	if proxy.IsIPv6() {
```
hzxuzhonghu, line 224: In dual stack mode, it will always listen to ipv6 address? If so, I do not  think it is right.

pilot/cmd/pilot-agent/options/agent.go
```diff
@@ -30,13 +29,13 @@ import (
 const xdsHeaderPrefix = "XDS_HEADER_"
 
 func NewAgentOptions(proxy *model.Proxy, cfg *meshconfig.ProxyConfig) *istioagent.AgentOptions {
-	proxy.DiscoverIPVersions()
+	proxy.DiscoverIPMode()
```
hzxuzhonghu, line 32: why we need this again, we should do it only once in `initProxy`

hzxuzhonghu (commented):
LGTM, will defer to @zhlsunshine

pilot/cmd/pilot-agent/options/agent.go
ramaraochavali: yes. This is not needed

ramaraochavali:
@zhlsunshine gentle ping. Can you PTAL and validate this?

pilot/cmd/pilot-agent/main.go
```diff
@@ -264,6 +264,9 @@ func initProxy(args []string) (*model.Proxy, error) {
 		proxy.IPAddresses = []string{podIP.String()}
 	}
 
+	// After IP addresses are set, let us discover IPMode.
+	proxy.DiscoverIPMode()
```
zhlsunshine, line 268: Why not call **DiscoverIPMode** after obtaining all the IPs from the node in later?

pilot/pkg/model/context.go
```diff
@@ -852,34 +860,31 @@ func (node *Proxy) SetWorkloadLabels(env *Environment) {
...
-		if node.GlobalUnicastIP == "" && addr.IsGlobalUnicast() {
-			node.GlobalUnicastIP = addr.String()
-		}
-		if addr.To4() != nil {
-			node.ipv4Support = true
-		} else {
-			node.ipv6Support = true
-		}
+// DiscoverIPMode discovers the IP Versions supported by Proxy based on its IP addresses.
+func (node *Proxy) DiscoverIPMode() {
```
zhlsunshine, line 864: I always has some confusion on this function, Istio discover IP Mode from `proxy.IPAddresses`, but how does it handle this situation? eg. a service is deployed in a dual-stack k8s kuberntes, but user define its ip family as below:
```
......
ipFamilies:
  - IPv6
  ipFamilyPolicy: SingleStack
......
```
This make thing become complicated, especially after k8s support dual-stack.

pilot/pkg/model/context.go
```diff
@@ -852,34 +860,31 @@ func (node *Proxy) SetWorkloadLabels(env *Environment) {
...
 // SupportsIPv4 returns true if proxy supports IPv4 addresses.
 func (node *Proxy) SupportsIPv4() bool {
-	return node.ipv4Support
+	return node.ipMode == IPv4 || node.ipMode == Dual
 }
 
 // SupportsIPv6 returns true if proxy supports IPv6 addresses.
 func (node *Proxy) SupportsIPv6() bool {
-	return node.ipv6Support
+	return node.ipMode == IPv6 || node.ipMode == Dual
```
zhlsunshine, line 882: I think `node.ipMode == IPv4 || node.ipMode == Dual` and `node.ipMode == IPv6 || node.ipMode == Dual` should not co-exist at the same time, otherwise, there will be a contradiction, especially when separately using `SupportsIPv4()` or `SupportsIPv6()`

pilot/pkg/model/context.go
ramaraochavali: It depends on what IP addresses proxy listens on and that is what we care about. BTW, it is not related to this PR, we will have to enhance it separately if we have to

pilot/pkg/model/context.go
ramaraochavali: SupportsIPv4() just tells whether proxyIP list has atleast one ipv4 address. Same for ipv6

pilot/cmd/pilot-agent/main.go
ramaraochavali: That makes sense. I will fix it.

zhlsunshine:
Hi @ramaraochavali and @hzxuzhonghu, I have checked this patch for `dual-stack`, `ipv4-only` and `ipv6-only` k8s cluster, it looks well for all of them.

hzxuzhonghu (approved):

PR #38219, opened 2022-04-01
Prioritize local over remote leader

frankbu:

Followup to https://github.com/istio/istio/pull/36121#discussion_r756199020


pilot/pkg/leaderelection/leaderelection_test.go
```diff
@@ -127,6 +136,38 @@ func TestPrioritizedLeaderElection(t *testing.T) {
...
+	_, stop2 := createElectionMulticluster(t, "pod2", "", false, watcher, false, true, client)
+	// A new remote pod cannot become leader
+	_, stop3 := createElectionMulticluster(t, "pod3", "", true, watcher, false, false, client)
+	close(stop3)
+	close(stop2)
+	close(stop)
+}
+
+func TestPrioritizedMulticlusterLeaderElection(t *testing.T) {
+	client := fake.NewSimpleClientset()
```
howardjohn, line 154: use `kube.NewFakeClient` - this may have test flakes because the k8s fake client is not safe by default

howardjohn (commented):
Didn't closely look so will ask instead: do we handle the case where old Istios are used (locally or remote) which don't use the new key?

And the inverse - do old versions handle the new key?

frankbu:
> Didn't closely look so will ask instead: do we handle the case where old Istios are used (locally or remote) which don't use > the new key?
>
> And the inverse - do old versions handle the new key?

An old key will look like that of a locally-running istiod, even if it's actually remote, so a new local one won't steal the lock, where it could if the key indicated the old leader was remote.

If we run an old istiod (local or remote), in the prioritized case, and it's the default revision, it could steal the lock from a remote new istiod that is also for the default revision, because the key prefix will make it think the other leader is a different revision.

I think these are reasonable expected behavior and unlikely corner cases anyway.

Otherwise, I think old and new works as before, regardless of the previous leader.

howardjohn:
> If we run an old istiod (local or remote), in the prioritized case, and
it's the default revision, it could steal the lock from a remote new istiod
that is also for the default revision, because the key prefix will make it
think the other leader is a different revision.


If I understand right, this means 2 instances think they should steal the
lock. This will result in them both constantly stealing the lock from each
other endlessly? If so, that seems not great

pilot/pkg/leaderelection/leaderelection_test.go
frankbu: I'm not following what you mean. I just copied this from the other tests.

Are you suggesting I change something here?

frankbu:
> If I understand right, this means 2 instances think they should steal the
> lock. This will result in them both constantly stealing the lock from each
> other endlessly? If so, that seems not great

No. Once the old one steals the lock, it will look like a locally running istiod for the default revision, so nobody can steal it after that.

pilot/pkg/leaderelection/leaderelection_test.go
frankbu: One of the tests calls `client.Fake.PrependReactor` which doesn't work with the `ExtendedClient` returned by `NewFakeClient()`, so I'll just leave all of them as `NewSimpleClientset()` which seems to work fine.

Monkeyanator (commented):
I'm not sure if this causes thrashing- let's look at when each kind of revision will steal the lock.

No matter what, old revisions will steal the leader lock if `l.revision != currentLeaderRevision && defaultRevision == l.revision` (i.e. the current lock holder has a revision that isn't the default and we have a revision that is), that's a given.

Let's also notice that new revisions will always see old revisions as local since old revisions won't set the special key prefix.

New remote revisions won't steal the lock if the current holder is local, so it never steals the lock except from newer remote revisions (since it sees all older ones as local). So IIUC with this change, even though a new remote instance is the default revision, it won't take the lock from any older revision. This seems wrong?

I'm not sure we have issues with thrashing between old and new revisions here that I could think of... when new is local and old is remote, the new will see both itself and the remote revision as local. Now we get back to a situation where each revision will steal just when it's the default revision and the current one isn't.

All this logic seems pretty fragile and implicit though, what's the logic behind local revisions winning elections in all cases over remote revisions? I think some explanatory comments in the leaderelection code would be useful.

frankbu:
> No matter what, old revisions will steal the leader lock if l.revision != currentLeaderRevision && defaultRevision == l.revision (i.e. the current lock holder has a revision that isn't the default and we have a revision that is), that's a given.

It's also possible that an old revision will steal the lock if the current lock holder has a revision that **IS** the default. This happens when the current lock holder is a new remote holder with the "^" prefix, so the old code will think `l.revision != currentLeaderRevision`, even though it is the same revision.

> Let's also notice that new revisions will always see old revisions as local since old revisions won't set the special key prefix.

Yes, this and the above are basically saying that old versions look and behave like they are local, regardless of where they are actually located. No thrashing will ever happen, just the old ones running remotely have higher priority than they deserve.

> New remote revisions won't steal the lock if the current holder is local, so it never steals the lock except from newer remote revisions (since it sees all older ones as local). So IIUC with this change, even though a new remote instance is the default revision, it won't take the lock from any older revision. This seems wrong?

This is related to why I also set `config.KeyComparison` even if `l.prioritized` is false. I added the local/remote prioritization code with the assumption that it is unrelated to the `prioritized leader election` feature, i.e., it applies all the time. The `prioritized` feature only applies to the revision prioritization, after the location decision has been made. Bottom line is we always favor a local istiod over a remote one, even if there is a more appropriate revision running remotely.

I did it this way because I think location prioritization should always be enabled, and I wasn't sure when the current prioritized leader election is enabled or not, but I see it is true by default, so I think it would be fine to handle location in the `prioritized` case only. That actually simplifies the algorithm a bit, leaving only the old code running locally but behaving like a local as the odd behavior, but that will go away in a few months after a couple more Istio releases.

I'll update the code with this changed.

pilot/pkg/leaderelection/leaderelection.go
```diff
@@ -142,8 +153,13 @@ func (l *LeaderElection) create() (*k8sleaderelection.LeaderElector, error) {
...
 		// Function to use to decide whether this revision should steal the existing lock.
 		config.KeyComparison = func(currentLeaderRevision string) bool {
+			var currentLeaderRemote bool
+			if currentLeaderRemote = strings.HasPrefix(currentLeaderRevision, remoteIstiodPrefix); currentLeaderRemote {
+				currentLeaderRevision = strings.TrimPrefix(currentLeaderRevision, remoteIstiodPrefix)
+			}
+			preferredLocation := !l.remote && currentLeaderRemote // this revison is running locally while the current one is remote
 			defaultRevision := l.defaultWatcher.GetDefault()
-			return l.revision != currentLeaderRevision &&
+			return (l.revision != currentLeaderRevision || preferredLocation) &&
```
howardjohn, line 162: confused by this. this means we will only steal it if we are preferredLocation AND the default revision. But what if the local and remote are both `rev=not-default`. Don't we still want stealing?

pilot/pkg/leaderelection/leaderelection.go
```diff
@@ -142,8 +153,13 @@ func (l *LeaderElection) create() (*k8sleaderelection.LeaderElector, error) {
 	if l.prioritized {
 		// Function to use to decide whether this revision should steal the existing lock.
 		config.KeyComparison = func(currentLeaderRevision string) bool {
```
howardjohn, line 155: This function is pretty complex and hard to understand and extremely critical. There is only a few inputs to it and its a pure, simple function. Can we split this out and unit test ~all the cases.

It would also be great to keep a list of all past `KeyComparison` functions, and run them against the current one, to ensure we do not have any inputs for which two different pods will both attempt to steal; if we are in this case its a critical bug. Would make me re-assured to have a test case to verify that is not the case.

pilot/pkg/leaderelection/leaderelection.go
frankbu: The current behavior is that the only way the lock is ever stolen is if the new one is a different revision and it's the default revision. I'm adding one more condition to also allow a local default revision to steal from a remote default revision. Non-default revisions will never steel the lock, before or after this change.

If we want to allow non-default local revisions to steal locks from non-default remote revisions, I think that would be fine, but it would need to also check that the local and remote are the same non-default revision, or we will end up with trashing.

I can add that change, if you'd like. This change is really just an optimization to try to have the local one be the preferred controller, but things still work without it, so that's why I was trying to change the behavior as little as possible for now, thinking we can always further optimize later, but I'm fine with adding the non-default revision remote->local override as well. Let me know what you think.

pilot/pkg/leaderelection/leaderelection.go
frankbu: Not sure I understand what you mean by `keep a list of all past KeyComparison functions`. Can you clarify?

I will add more unit tests to try to make sure that all combinations of remote/local, default/non-default, default/default and non-default/non-default are tested.

howardjohn: What I mean is that in a given cluster, we will have Istiods running with:
* No prioritized election
* The current HEAD  KeyComparison function
* This new KeyComparison fn
* Any future KeyComparison functions

If ANY of these different versions both think they should steal a leader lock, we will end up with an infinite loop of stealing the lock and probably break the cluster.

frankbu: I've improved the algorithm and added testing of all the combinations. PTAL.

pilot/pkg/leaderelection/leaderelection.go
```diff
@@ -60,7 +64,9 @@ type LeaderElection struct {
 
 	// Criteria to determine leader priority.
 	revision       string
+	remote         bool
 	prioritized    bool
+	ignoreLocation bool // turn on for testing with old key comparision
```
howardjohn, line 69: Sorry for the back and forth.. I don't think we should have the old code in the prod code path, since we introduce a lot of complex there.

I was thinking in the prod code we have
`config.KeyComparison = LocationPrioritizedComaprision // not an inline function now`

Then in the tests we have:

```go
compFunctions := []func(){
  LocationPrioritizedComaprision,
  func() { ... old logic ...}
}

Then we can just add simple unit tests for these and not need to do the full (somewhat heavy) leader election end-to-end flow we have now. We can make a table driven test for all cases

Comment (frankbu) [reply pilot/pkg/leaderelection/leaderelection.go]: > I was thinking in the prod code we have
>  config.KeyComparison = LocationPrioritizedComaprision // not an inline function now

The KeyComparison function just takes a single string argument, but our implementation also needs access to the both the config and the leaderelection struct, so it needs to be written in-scope, IIUC.

Comment (howardjohn) [reply pilot/pkg/leaderelection/leaderelection.go]:
```
config.KeyComparison = func(s string) {
  LocationPrioritizedComaprision(s, someOtherDep)
}
```
should work

Comment (frankbu) [reply pilot/pkg/leaderelection/leaderelection.go]: I split out the compare function and made the test simpler, table driven.

File: pilot/pkg/leaderelection/leaderelection_test.go [hunk 5]
```diff
@@ -127,6 +139,114 @@ func TestPrioritizedLeaderElection(t *testing.T) {
...
+
+func SimpleRevisionComparison(currentLeaderRevision string, l *LeaderElection) bool {
+	// Old key comparison impl for interoperablilty testing
+	defaultRevision := l.defaultWatcher.GetDefault()
+	return l.revision != currentLeaderRevision &&
+		// empty default revision indicates that there is no default set
+		defaultRevision != "" && defaultRevision == l.revision
+}
+
+func TestPrioritizedMulticlusterLeaderElection(t *testing.T) {
```
howardjohn, line 164: I was thinking something like this:
```

type instance struct {
	revision string
	remote   bool
	comp     string
}

func (i instance) GetComp() LeaderComparison {
	switch i.comp {
	case "local":
		return LocationPrioritizedComparison
	case "simple":
		return SimpleRevisionComparison
	default:
		panic("unknown comparison type")
	}
}
// TestPrioritizationCycles
func TestPrioritizationCycles(t *testing.T) {
	cases := []instance{}
	for _, rev := range []string{"", "default", "not-default"} {
		for _, loc := range []bool{false, true} {
			for _, comp := range []string{"local", "simple"} {
				cases = append(cases, instance{
					revision: rev,
					remote:   loc,
					comp:     comp,
				})
			}
		}
	}
	for _, start := range cases {
		t.Run(fmt.Sprint(start), func(t *testing.T) {
			checkCycles(t, start, cases, nil)
		})
	}
}

func alreadyHit(cur instance, chain []instance) bool {
	for _, cc := range chain {
		if cur == cc {
			return true
		}
	}
	return false
}

func checkCycles(t *testing.T, start instance, cases []instance, chain []instance) {
	if alreadyHit(start, chain) {
		t.Fatalf("cycle on leader election: cur %v, chain %v", start ,chain)
	}
	for _, nextHop := range cases {
		next := LeaderElection{
			remote:         nextHop.remote,
			defaultWatcher: &fakeDefaultWatcher{defaultRevision: "default"},
			revision:       nextHop.revision,
		}
		if nextHop.GetComp()(start.revision, &next) {
			nc := append([]instance{}, chain...)
			nc = append(nc, start)
			checkCycles(t, nextHop, cases, nc)
		}
	}
}
```

Then there are no changes in the core code required and its ~instant instead of taking ~20s or so. This ensures there are no cases where we steal the lock in an endless loop

frankbu: @howardjohn Thanks for the "crystal clear" explanation of your suggestion (i.e., writing the code). Makes sense.

Since your test handles the cycle checking, but not that the prioritization works as expected, I've also put back my original two other tests that did some basic checking of the functionality to make sure that remote functionality is at least making it through the system. PTAL.

howardjohn (approved):
Thanks! Sorry this was a bit of a painful review but I am pretty happy with the end result, hopefully you are as well. Thanks for pushing this through.

PR #38226, opened 2022-04-01
Use strict mode instead of plaintext if auto_mtls label is set.

costinm:
Small corner case, if an app is using both proxyless for gRPC ports and envoy for TCP/HTTP - Istio will
inject the auto mtls label, and that will make clients initiate mTLS connections. 

Right now we don't support this case ( WIP ), but it is the correct behavior.


howardjohn (commented):
Will this break any clients that are _not_ doing mTLS? Which is likely a lot, since you need to change your app (https://istio.io/latest/blog/2021/proxyless-grpc/#in-the-client)

costinm:
There is no 'magic' in proxyless gRPC - a server is either mTLS or plaintext. If a workload is part of the mesh and is using a sidecar, it is expected to use secure mode, no middle ground or confusion.

So workloads that are not using proxyless gRPC or a sidecar are NOT expected to talk with workloads that do use proxyless gRPC in secure mode. Users are expected to make changes - for migrations they can start a new service or use a different port for 'secure gRPC', and explicitly configure. 

This is what they would do if they were not using any mesh at all - if I have a regular gRPC app using plaintext and I want to 
start doing mTLS, it is expected that I would deploy mTLS on a different port, and modify the clients to use the new port with mTLS.

costinm:
Added an extra check, to allow users to explicitly select 'plain text' even if auto-mtls label is set.

pilot/pkg/networking/grpcgen/lds.go
```diff
@@ -122,6 +123,12 @@ func buildInboundListeners(node *model.Proxy, push *model.PushContext, names []s
 func buildInboundFilterChains(node *model.Proxy, push *model.PushContext, si *model.ServiceInstance, applier authn.PolicyApplier) []*listener.FilterChain {
 	mode := applier.GetMutualTLSModeForPort(si.Endpoint.EndpointPort)
 
+	// auto-mtls label is set - clients will attempt to connect using mtls, and
+	// gRPC doesn't support permissive.
+	if node.Metadata.Labels[label.SecurityTlsMode.Name] == "istio" && mode == model.MTLSPermissive {
+		mode = model.MTLSStrict
```
hzxuzhonghu, line 129: Doesn't this break clients that have no sidecar?

costinm: Why would it break them ? If a client without sidecar doesn't have the auto-mtls label, grpc will be configured as plain text - and  clients will use plain text as well ( as is the default behavior with auto-mtls ).

If the client has the auto-mtls label - we configure TLS, and clients will work as expected. 

It is still possible to set DestinationRule and PeerAuth explicitly to either STRICT or DISABLED - however 
permissive is not supported by gRPC, but we can do auto mtls.

costinm: If you mean regular gRPC that is not using XDS ( i.e. no sidecar, no proxyless gRPC ) - yes, they would be broken if 
auto-mtls label is set. 

Since proxyless doesn't support permissive - in order to migrate the user will need to use a separate port, and make code  changes if they want to transition. The purpose of this PR is to make sure Istio workloads work properly with automtls, having non-mesh gRPC clients talk with mesh enabled is not supported unless 2 ports are used.

Note that when we implement 'BTS/Hbone', we will have a dedicated port for mTLS and the original port will be plain text. This is not implemented, so we still need auto-mtls to work properly.

hzxuzhonghu: >If a client without sidecar doesn't have the auto-mtls label, grpc will be configured as plain text - and clients will use plain text as well ( as is the default behavior with auto-mtls ).

I understand `node.Metadata.Labels[label.SecurityTlsMode.Name] == "istio"`  is saying the grpc server is injected with a sidecar.

hzxuzhonghu: If grpc has sidecar though not common, then should always disable tls?

costinm: I think the label is saying 'workload supports istio TLS mode' - not that it has a sidecar ( or an envoy sidecar ). 
The mode really means mTLS connections with spifee certs, etc.

gRPC supports the same auth as Istio - same certs, verification, etc. We certainly don't want to disable it. 

It is possible for a proxyless gRPC application to have a sidecar - with the gRPC port excluded from interception. 
In this case - the label will be set, and gRPC ( like the intercepted ports ) will use the exact same mTLS.

hzxuzhonghu: I think i get to understand the case, grpc does not support permissive mode, so need to do a special processing

hzxuzhonghu (approved):

PR #38231, opened 2022-04-02
Fix root cert configmap err check

hzxuzhonghu:

This was a fix, not sure why we did not prevent it by review. 

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/security/pkg/k8s/configutil.go
+++ b/security/pkg/k8s/configutil.go
@@ -49,14 +49,14 @@ func InsertDataToConfigMap(client corev1.ConfigMapsGetter, lister listerv1.Confi
 		if _, err = client.ConfigMaps(meta.Namespace).Create(context.TODO(), configmap, metav1.CreateOptions{}); err != nil {
 			// Namespace may be deleted between now... and our previous check. Just skip this, we cannot create into deleted ns
 			// And don't retry a create if the namespace is terminating
-			if errors.IsNotFound(err) || errors.HasStatusCause(err, v1.NamespaceTerminatingCause) {
+			if errors.IsAlreadyExists(err) || errors.HasStatusCause(err, v1.NamespaceTerminatingCause) {
 				return nil
 			}
 			return fmt.Errorf("error when creating configmap %v: %v", meta.Name, err)
 		}
 	} else {
 		// Otherwise, update the config map if changes are required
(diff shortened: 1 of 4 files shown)


security/pkg/k8s/configutil.go
```diff
@@ -49,14 +49,14 @@ func InsertDataToConfigMap(client corev1.ConfigMapsGetter, lister listerv1.Confi
 		if _, err = client.ConfigMaps(meta.Namespace).Create(context.TODO(), configmap, metav1.CreateOptions{}); err != nil {
 			// Namespace may be deleted between now... and our previous check. Just skip this, we cannot create into deleted ns
 			// And don't retry a create if the namespace is terminating
-			if errors.IsNotFound(err) || errors.HasStatusCause(err, v1.NamespaceTerminatingCause) {
+			if errors.IsAlreadyExists(err) || errors.HasStatusCause(err, v1.NamespaceTerminatingCause) {
```
hzxuzhonghu, line 52: This is the fix.

ramaraochavali: Do we have a test for it?

hzxuzhonghu: This is a common usage of client-go. I am not sure if there is a test for all these kinds of code.

ramaraochavali: I understand but how do we ensure we captured the right error -  Can this test TestInsertDataToConfigMap be extended to have this case to ensure we are handling the right error

hzxuzhonghu: TBH, i donot know how to test this. Even in k8s `Create` error values are not tested thoroughly.

ramaraochavali: :-)

ramaraochavali (approved):

lei-tang (approved):

security/pkg/k8s/configutil.go
howardjohn: Does the fake client not return proper errors? If it doesn't we should file an upstream request. If it does we should add a test

hzxuzhonghu: WIll investigate on this

PR #38232, opened 2022-04-02
Refactor workloadentry controller 

hzxuzhonghu:

To use the generic queue instead of k8s lib.

### DIFF
--- a/pilot/pkg/config/kube/gateway/deploymentcontroller.go
+++ b/pilot/pkg/config/kube/gateway/deploymentcontroller.go
@@ -143,7 +143,8 @@ func (d *DeploymentController) Run(stop <-chan struct{}) {
 }
 
 // Reconcile takes in the name of a Gateway and ensures the cluster is in the desired state
-func (d *DeploymentController) Reconcile(req types.NamespacedName) error {
+func (d *DeploymentController) Reconcile(key interface{}) error {
+	req := key.(types.NamespacedName)
 	log := log.WithLabels("gateway", req)
 
 	gw, err := d.gatewayLister.Gateways(req.Namespace).Get(req.Name)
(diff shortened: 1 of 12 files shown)


pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -143,7 +143,8 @@ func (d *DeploymentController) Run(stop <-chan struct{}) {
 }
 
 // Reconcile takes in the name of a Gateway and ensures the cluster is in the desired state
-func (d *DeploymentController) Reconcile(req types.NamespacedName) error {
```
ramaraochavali, line 146: Why this change? Is there a problem if we leave it to have types/NamespacedName as earlier?

hzxuzhonghu: The function is the workfunc of the queue, it is not always of NamespacedName type. You can see the workloadentry controller

costinm: I looked - it seems Add() is adding a NamespacedName(), same for AddObject() -  I assume the object is retrieved is the same type - not sure why the Queue implementation uses interface{} in the function, maybe we should fix it ? 

But probably not a high priority, I can live with it as interface{}

costinm (approved):

pilot/pkg/config/kube/gateway/deploymentcontroller.go
hzxuzhonghu: I think it is to provide high flexibility. In most cases even for most k8s controllers name/ns is enough, but for some other cases, the item can be any customized struct.

PR #38233, opened 2022-04-04
Adopt efficient path based routing

### LINKED ISSUE #38222: Adopt efficient path based routing in Istio
https://github.com/envoyproxy/envoy/issues/18148#event-6348368516

This is useful for more efficient implementation of Ingress and k8s Gateway.

How I would implement it:
* https://github.com/istio/istio/blob/e11efc56130c1be35649b18b7eda6707aef35759/pilot/pkg/config/kube/gateway/conversion.go#L1057 remove this conversion, and the same one on the Ingress code
* For the internally generated VirtualServices, add an annotation indicating we should use PathPrefix mode.
* In the VS -> XDS Route code, add a version check. If >=1.14, set path_prefix. If old, add the same regex hack

dddddai:
Fixes #38222

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


dddddai:
We also need to update istio-proxy to get https://github.com/envoyproxy/envoy/pull/20145 included?

ericvn:
> We also need to update istio-proxy to get [envoyproxy/envoy#20145](https://github.com/envoyproxy/envoy/pull/20145) included?

We autoupdate to the latest commits daily. See https://github.com/istio/proxy/pull/3795. Unfortunately, there was a change upstream which is causing issues. Waiting on the team to fix it.

dddddai:
> We autoupdate to the latest commits daily. See https://github.com/istio/proxy/pull/3795. Unfortunately, there was a change upstream which is causing issues. Waiting on the team to fix it.

I see, thanks

> Also, it seems like this might need release notes.

There's no user-facing change as it's just an internal implementation change

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -816,7 +823,27 @@ func translateRouteMatch(in *networking.HTTPMatchRequest) *route.RouteMatch {
 		case *networking.StringMatch_Exact:
 			out.PathSpecifier = &route.RouteMatch_Path{Path: m.Exact}
 		case *networking.StringMatch_Prefix:
-			out.PathSpecifier = &route.RouteMatch_Prefix{Prefix: m.Prefix}
+			if (model.UseIngressSemantics(vs) || model.UseGatewaySemantics(vs)) && m.Prefix != "/" {
+				if node.IstioVersion.Compare(&istioVersion114) >= 0 {
```
dddddai, line 827: Is `proxy.IstioVersion` the version of associated control-plane or istio-proxy itself? 
If it's the former, this would not work when users only upgrade the control-plane while the istio-proxy version is still <1.14

howardjohn: The proxy. We have functions already for this -- IsIstioVersionGE112, etc. Would recommend using those

dddddai: Thanks but I couldn't find them, could you point out where they are?

ramaraochavali: We have removed such version checks in Master recently. You can find an example in 1.13 https://github.com/istio/istio/blob/c2461b13e11cabb82a0d6e7aa4bcd4e015b86ad6/pilot/pkg/networking/util/util.go#L276 - Please follow similar pattern and add that method in util

dddddai: Thank you! Will add it once https://github.com/istio/proxy/pull/3795 get merged

howardjohn:
fyi proxy is merged

dddddai:
I think we get stuck on https://github.com/istio/istio/pull/38310

PR #28085, opened 2020-10-19
allow configure RENEW_DEADLINE for operator manager

richardwxn:
resolve: https://github.com/istio/istio/issues/27509

Issue #27509, opened 2020-09-24
Managed Istio: Manager exited non-zero


luiof:
Hi team! We have the managed istio on multiple IKS clusters.
We are using the latest 1.7.2 version and sporadically on some clusters we can see logs as the following:
```
2020-09-24T13:23:12.697712Z	error	error retrieving resource lock ibm-operators/istio-operator-lock: Get "https://172.21.0.1:443/api/v1/namespaces/ibm-operators/configmaps/istio-operator-lock": context deadline exceeded
2020-09-24T13:23:12.697802Z	info	failed to renew lease ibm-operators/istio-operator-lock: timed out waiting for the condition
2020-09-24T13:23:12.697865Z	fatal	Manager exited non-zero: leader election lost
```

and as:
```
    Sep 24 15:23:13kube-brq9cjpd05d8tb2l7bfg-ghostfordoc-default-00000115syslogE0924 13:23:13.260809    6966 pod_workers.go:191] Error syncing pod bd4d62ea-e38d-4e67-98c0-48b339f06337 ("managed-istio-operator-88cc547b4-8p9x4_ibm-operators(bd4d62ea-e38d-4e67-98c0-48b339f06337)"), skipping: failed to "StartContainer" for "istio-operator" with CrashLoopBackOff: "back-off 10s restarting failed container=istio-operator pod=managed-istio-operator-88cc547b4-8p9x4_ibm-operators(bd4d62ea-e38d-4e67-98c0-48b339f06337)"

    Sep 24 15:23:13kube-brq9cjpd05d8tb2l7bfg-ghostfordoc-default-00000115kubelet.logE0924 13:23:13.260809    6966 pod_workers.go:191] Error syncing pod bd4d62ea-e38d-4e67-98c0-48b339f06337 ("managed-istio-operator-88cc547b4-8p9x4_ibm-operators(bd4d62ea-e38d-4e67-98c0-48b339f06337)"), skipping: failed to "StartContainer" for "istio-operator" with CrashLoopBackOff: "back-off 10s restarting failed container=istio-operator pod=managed-istio-operator-88cc547b4-8p9x4_ibm-operators(bd4d62ea-e38d-4e67-98c0-48b339f06337)"
```

This seems to appear only some days and everything seems to be fine (the system works as usually). 

The `istioctl analyze -n namespace` doesn't show any problem.
The `istioctl version` shows that every proxy is updated to latest 1.7.2 version.

Any ideas? Are there action that we can take to avoid those crash?

[...]

PR #38234, opened 2022-04-04
Expose RENEW_DEADLINE env variable in operator

nielsole:
Exposes RENEW_DEADLINE  env variable for configuration https://github.com/istio/istio/pull/28085
This is useful if one hits: https://github.com/istio/istio/issues/27509

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


nielsole:
I now see this was already attempted to merge once: https://github.com/istio/istio/pull/28085/commits/4a6fe73842689c8128736e1569d9eea8e135540e
From: https://github.com/istio/istio/pull/28085#discussion_r508796737:
> is this really needed? we don't want to introduce new values.yaml stuff
we could also wrap the env variable in an if statement, so that we don't need to change values.yaml. lmk what you prefer.

jacob-delgado:
cc @litong01 can you take a look at this?

litong01:
> I now see this was already attempted to merge once: [4a6fe73](https://github.com/istio/istio/commit/4a6fe73842689c8128736e1569d9eea8e135540e) From: [#28085 (comment)](https://github.com/istio/istio/pull/28085#discussion_r508796737):
> 
> > is this really needed? we don't want to introduce new values.yaml stuff
> 
> we could also wrap the env variable in an if statement, so that we don't need to change values.yaml. lmk what you prefer.

I think that the general consensus is to use the environment without changing the deployment template and values file. I also think changes to these files requires running `make gen`

nielsole:
> the general consensus is to use the environment without changing the deployment template and values file

How does one go about that? There is no `{{- toYaml .extraEnvs | nindent 12 }}` or similar. 
I would be open to changing the PR towards supporting changing env variables without changing the deployment template or values file.

nielsole:
sorry, to bump this. It would seem counterintuitive to me why the operator would have configuration options that are not exposed via the chart. Otherwise every configuration change requires a fork of the chart. 
I just took a look and it seems that `RENEW_DEADLINE` [is the only env variable that is not exposed in the chart](https://github.com/istio/istio/search?q=os.LookupEnv). `LEADER_ELECTION_NAMESPACE` and `WATCH_NAMESPACE` are exposed. Why would `RENEW_DEADLINE` be different?

I see three options going forward:
* refactor all three env variables into `{{- toYaml .extraEnvs | nindent 12 }}`
* Use the approach I chose currently.
* Close this ticket and keep `RENEW_DEADLINE` not exposed to users and require a fork of the chart.

Please let me know which approach you prefer and I will put the PR into good shape.

SebTardif:
Today, what's the faster way to change the value of RENEW_DEADLINE?
I have Image:docker.io/istio/operator:1.13.2

nielsole:
If you want to use the chart, you will have to maintain your own fork unfortunately.

howardjohn:
https://preliminary.istio.io/latest/docs/setup/additional-setup/customize-installation-helm/

Issue #37938, opened 2022-03-15
mention log_output_level flag in istioctl help


rittneje:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**

istioctl supports a `--log_output_level` flag to control the logging level of its various components. This flag is very useful for troubleshooting when istioctl is misbehaving. However, it is not documented in `istioctl -h` or `istioctl help`.

**Describe alternatives you've considered**

n/a

**Affected product area (please put an X in all that apply)**

[ ] Docs
[x] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

howardjohn:
Does `--log_output_level=all:debug` not work? They SetOutputLevel overrides those?

rittneje:
We can try that. However, that flag is not mentioned in the output from `istioctl help`.

litong01:
@howardjohn @rittneje the --log_output_level=all:debug does work, it will make `istioctl` show a lot of information.

rittneje:
Yes we were able to use it successfully. For this reason I've changed the request to instead document this flag.

hanxiaop:
@rittneje The flag is not shown because it prints several lines with many scopes. See https://github.com/istio/istio/issues/22061. There's a command `istioctl options` that can see all of these global flags. I think the point may become that users are not awaring of using this command. Maybe adding a more detailed command could be better?

rittneje:
@hanxiaop Personally, I don't understand why `istioctl help` arbitrarily prints some of the global flags (e.g., `--context`), and `istioctl options` prints the rest of them. Fundamentally, the documentation ought to be internally consistent, and put them all in a singular location.

In addition to that, `istioctl help` should definitely explicitly refer to `istiotctl options`. That is a fairly non-standard way to document flags (other than `kubectl` I don't know of any application that does that), so as it stands it is extremely unlikely that most developers will ever see that documentation. Also, to be frank, I don't understand what the motivation is for not simply including _all_ flags under `istioctl help`, regardless of whether they are global.

zirain:
fixed by https://github.com/istio/istio/pull/38236.

PR #38236, opened 2022-04-04
Improve istioctl options command description

hanxiaop:
https://github.com/istio/istio/issues/37938

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience


esnible (commented):
Improving this message is a good idea.  We might want to change the grammar of this message.

We wish to be similar to `kubectl`.  If `kubectl` is run without any arguments it prints the major options but also

```
Use "kubectl options" for a list of global command-line options (applies to all
commands).
```

We could do something similar.  We could hide the `istioctl options` command and add text at the bottom 

```
Use "istioctl options" for a list of global command-line options (applies to all
commands).
```

What do you think?

hanxiaop:
@esnible I think it's a great idea. I'll revise it.

GregHanson (approved):

istioctl/cmd/root.go
```diff
@@ -144,6 +146,9 @@ func GetRootCmd(args []string) *cobra.Command {
 		Long: `Istio configuration command line utility for service operators to
 debug and diagnose their Istio mesh.
 `,
+		Run: func(cmd *cobra.Command, args []string) {
+			_ = cmd.Help()
```
howardjohn, line 150: do we need this?

howardjohn: If so should we use RunE and handle the error?

hanxiaop: Yes, the command should be runnable. Otherwise the options cmd will not be recognized in the templater. Revised to RunE.

hanxiaop:
@esnible PTAL. Thanks.

hanxiaop:
not stale

hanxiaop:
@esnible PTAL.

esnible (approved):

howardjohn (commented):
This broke all `--help` commands, no flags are shown

Issue #38229, opened 2022-04-01
TestCustomizeMetrics flake


bianpengyuan:
### Bug Description

TestCustomizeMetrics is flaky after Wasm secret PR merged: https://prow.istio.io/view/gs/istio-prow/logs/integ-telemetry_istio_postsubmit/1509998564334374912

### Version

```prose
n/a
```

PR #38239, opened 2022-04-04
Deploy container registry before echo at customize metric test.

bianpengyuan:
Fix https://github.com/istio/istio/issues/38229

This makes sure the container registry is ready to serve in the cluster when test runs.


bianpengyuan:
Tested multiple times, all passed.

zirain (approved):
lgtm

Issue #38245, opened 2022-04-05
Allow the gateway helm chart to override lifecycle settings set at the mesh global config


ceastman-r7:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**
Update the gateway helm chart to allow the lifecycle section be overridden 

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[ ] Docs
[x] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**
I have the following currently set for the istiod helm install:
global:
  proxy:
    lifecycle:
      preStop:
        exec:
          command:
            - sh
            - -c
            - echo "Stopping container in 30 seconds...">/proc/1/fd/1 && sleep 30
      postStart:
        exec:
          command:
          - pilot-agent
          - wait

but i need the ability to override the preStop section due to aws nlb update speed with the following:

          lifecycle:
            preStop:
              exec:
                command:
                - sh
                - -c
                - echo "Stopping container in 300 seconds...">/proc/1/fd/1 ; sleep 300

howardjohn:
https://istio.io/latest/docs/setup/additional-setup/customize-installation-helm/ is the way to do this

PR #38247, opened 2022-04-05
Add lifecycle and terminationGracePeriodSeconds overrides to gateway(s) helm charts

ceastman-r7:
https://github.com/istio/istio/issues/38245

This PR adds the following optional values override to the gateway(s) helm charts:
lifecycle

This is mainly to allow people to have different values for lifecycle:preStop that are specifically needed for cloud provider load balancers that are slow to update.


howardjohn (commented):
Istio already provided this draining functionality natively with the termination drain duration mesh config option

ceastman-r7:
I removed the references to terminationGracePeriodSeconds

Issue #38246, opened 2022-04-05
getServiceAccountSecret assumes auto-created service account token secrets


liggitt:
### Bug Description

[getServiceAccountSecret](https://github.com/istio/istio/blob/5fd92a170f500462bb6ffd1d9462d6592968cd9a/istioctl/pkg/multicluster/remote_secret.go#L234) assumes 1) secret-based tokens will always be created for service accounts, and 2) a manually provided secret name will be referenced from the service account

Neither of those assumptions are universally correct.

In 1.21+ secret-based tokens are no longer used for mounting into pods (ephemeral time-limited tokens are).

In 1.24+, secret-based tokens are no longer auto-created by default for new service accounts.

Using ephemeral time-bound tokens is preferred in 1.21+ (see the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API) if possible.

If a secret-based token is still desired, one can be [created manually](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-service-account-api-token), but will not be referenced from the service account's .secrets list. If a secret name is explicitly provided to istioctl, it should be validated by looking at the type (`kubernetes.io/service-account-token`) and the `kubernetes.io/service-account.name` annotation (should match the service account name)

### Version

```prose
n/a
```

liggitt:
cc @monkeyanator

stevenctl:
cc @costinm 

Thanks for reporting!

howardjohn:
@stevenctl how come our CI didn't capture this? Or did it? We do test 1.24 - but possibly a build before the changes described here

PR #38249, opened 2022-04-05
create-remote-secret: handle k8s 1.24+ sa token behavior

stevenctl:
A fix for https://github.com/istio/istio/issues/38246 that maintains current behavior. 

IIUC, this part doesn't break us:

> In 1.21+ secret-based tokens are no longer used for mounting into pods (ephemeral time-limited tokens are).

wip: need to add tests and release note

### DIFF
--- a/istioctl/pkg/multicluster/remote_secret.go
+++ b/istioctl/pkg/multicluster/remote_secret.go
@@ -238,6 +238,74 @@ func getServiceAccountSecret(client kube.ExtendedClient, opt RemoteSecretOptions
 		return nil, err
 	}
 
+	if !kube.IsAtLeastVersion(client, 24) {
+		return legacyGetServiceAccountSecret(serviceAccount, client, opt)
+	}
+	return getOrCreateServiceAccountSecret(serviceAccount, client, opt)
+
+}
+
+// In 1.24+ we can't assume the secrets will be referenced in the ServiceAccount or be created automatically.
+// See https://github.com/istio/istio/issues/38246
+func getOrCreateServiceAccountSecret(
+	serviceAccount *v1.ServiceAccount,
+	client kube.ExtendedClient,
+	opt RemoteSecretOptions,
+) (*v1.Secret, error) {
+	ctx := context.TODO()
+
+	// manually specified secret, make sure it references the ServiceAccount
+	if opt.SecretName != "" {
+		secret, err := client.CoreV1().Secrets(opt.Namespace).Get(ctx, opt.SecretName, metav1.GetOptions{})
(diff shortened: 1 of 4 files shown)


istioctl/pkg/multicluster/remote_secret.go
```diff
@@ -238,6 +238,74 @@ func getServiceAccountSecret(client kube.ExtendedClient, opt RemoteSecretOptions
...
 	}
 
+	if !kube.IsAtLeastVersion(client, 24) {
+		return legacyGetServiceAccountSecret(serviceAccount, client, opt)
+	}
+	return getOrCreateServiceAccountSecret(serviceAccount, client, opt)
+
+}
+
+// In 1.24+ we can't assume the secrets will be referenced in the ServiceAccount or be created automatically.
```
nmittler, line 248: nit: In `Kubernetes` 1.24+...

istioctl/pkg/multicluster/remote_secret.go
```diff
@@ -238,6 +238,74 @@ func getServiceAccountSecret(client kube.ExtendedClient, opt RemoteSecretOptions
...
+	ctx := context.TODO()
+
+	// manually specified secret, make sure it references the ServiceAccount
+	if opt.SecretName != "" {
+		secret, err := client.CoreV1().Secrets(opt.Namespace).Get(ctx, opt.SecretName, metav1.GetOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("could not get specified secret %s/%s: %v",
+				opt.Namespace, opt.SecretName, err)
+		}
+		if err := validateServiceAccountSecret(serviceAccount, secret); err != nil {
```
nmittler, line 264: Do we need something added to the analyzer for this as well?

nmittler (approved):

istioctl/pkg/multicluster/remote_secret.go
stevenctl: I don't think so. 

1. We may not even use static tokens in some meshes. 
2. We really only read from this secret during remote-secret creation. The remote-secret won't have explicit references to this, just extracted content.

Purushotham233:
@stevenctl 
sa token behaviour change is starting from 1.22. [Reference](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-an-api-token-for-a-serviceaccount)

Why the min. cluster version is configured as 24?

>  	if !kube.IsAtLeastVersion(client, 24) {
> 		return legacyGetServiceAccountSecret(serviceAccount, client, opt)
> 	}

PR #38250, opened 2022-04-05
telemetry api: add support for TracingSelector

douglas-reid:
Adds control plane support for the new TracingSelector in the Telemetry API. This is needed to allow users to control differing sampling rates, providers, etc., for traffic that a workload originates rather than receives.

NOTE: I have not validated the behavior of this PR in cluster yet. This is being sent right now to allow for review and discussion over approach (as there is a parallel effort underway for logs).

Initial testing with:

```
kind: Telemetry
metadata:
  name: mesh-default
  namespace: istio-system
spec:
  tracing:
  - match:
      mode: SERVER
    providers:
    - name: zipkin
    randomSamplingPercentage: 100
  - match:
      mode: CLIENT
    providers:
    - name: zipkin
    randomSamplingPercentage: 13
```

Seems to indicate that proper `WorkloadMode` support exists and works.

### DIFF
--- a/pilot/pkg/model/telemetry.go
+++ b/pilot/pkg/model/telemetry.go
@@ -167,6 +167,11 @@ type computedTelemetries struct {
 }
 
 type TracingConfig struct {
+	ServerSpec TracingSpec
+	ClientSpec TracingSpec
+}
+
+type TracingSpec struct {
 	Provider                     *meshconfig.MeshConfig_ExtensionProvider
 	Disabled                     bool
 	RandomSamplingPercentage     float64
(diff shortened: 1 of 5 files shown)


pilot/pkg/model/telemetry.go
```diff
@@ -205,68 +210,84 @@ func (t *Telemetries) Tracing(proxy *Proxy) *TracingConfig {
...
-		UseRequestIDForTraceSampling: true,
-	}
-	if cfg.Provider == nil {
-		cfg.Disabled = true
-		return &cfg
+	clientSpec := TracingSpec{UseRequestIDForTraceSampling: true}
+	serverSpec := TracingSpec{UseRequestIDForTraceSampling: true}
+
+	if hasDefaultProvider {
+		// todo: what do we want to do with more than one default provider?
```
zirain, line 223: can we add some validation, maybe `istioctl analyze`?

pilot/pkg/model/telemetry.go
```diff
@@ -205,68 +210,84 @@ func (t *Telemetries) Tracing(proxy *Proxy) *TracingConfig {
...
+				specs = []*TracingSpec{&clientSpec}
+			case tpb.WorkloadMode_SERVER:
+				specs = []*TracingSpec{&serverSpec}
 			}
 		}
-		if !includeConfig {
-			break
+
+		if len(names) > 0 {
+			// NOTE: we only support a single provider per mode
```
zirain, line 244: add validation in webhook?

zirain:
link to https://github.com/istio/istio/issues/37665

pilot/pkg/model/telemetry.go
douglas-reid: I believe we have validation warnings for this scenario currently: https://github.com/istio/istio/blob/master/pkg/config/validation/validation.go#L3604

douglas-reid:
@howardjohn @zirain should we try to get this merged and onto 1.14 soon?

howardjohn (approved):

PR #38253, opened 2022-04-05
[tf] Refactor echo forwarder

nmittler:
A number of improvements for better overall performance and code cleanup.

Connection pooling now better supported for grpc and http.]


pkg/test/echo/server/forwarder/grpc.go
```diff
@@ -34,87 +36,204 @@ import (
 var _ protocol = &grpcProtocol{}
 
 type grpcProtocol struct {
-	conn func() (conn *grpc.ClientConn, err error)
+	q        *workQueue
+	connPool map[connectionPoolKey]*pooledGRPCConn
+	mu       sync.Mutex
```
stevenctl, line 41: nit: mu hat

pkg/test/echo/server/forwarder/http.go
```diff
@@ -22,197 +22,314 @@ import (
...
+	} else {
+		// Update the last time used.
+		pooledT.lastUsed = time.Now()
 	}
-	return raw[:schemeEnd+pathBegin], raw[schemeEnd+pathBegin:]
+	return pooledT.t, nil
 }
 
-func (c *httpProtocol) newClient() (*http.Client, error) {
+type pooledHTTPConn struct {
```
stevenctl, line 165: can we have a `pooledConn` interface and split out the `getFromPool` part?

pkg/test/echo/server/forwarder/config.go
```diff
@@ -158,9 +208,79 @@ func newTLSConfig(r *proto.ForwardEchoRequest, getClientCertificate func(info *t
...
+			if c.scheme == scheme.HTTP {
+				return nil, fmt.Errorf("http3 requires HTTPS")
+			}
+		case r.Http2:
+			if r.Alpn == nil {
+				tlsConfig.NextProtos = []string{"h2"}
+			}
+		default:
+			if r.Alpn == nil {
+				tlsConfig.NextProtos = []string{"http/1.1"}
```
costinm, line 226: AFAIK http/1.1 is the default and doesn't use NextProtos ( unless a new RFC added it - but even then it shouldn't be required).

pkg/test/echo/cmd/client/main.go
```diff
@@ -142,6 +140,12 @@ func init() {
 		"Treat as a server first protocol; do not send request until magic string is received")
 	rootCmd.PersistentFlags().BoolVarP(&followRedirects, "follow-redirects", "L", false,
 		"If enabled, will follow 3xx redirects with the Location header")
+	rootCmd.PersistentFlags().BoolVar(&newConnectionPerRequest, "new-connection-per-request", false,
```
costinm, line 143: If you added this to the proto, why does it need a flag ? It's a bit confusing - there is a default value in proto, but also one in flag - I doubt anyone will understand how they interact. 

What is the current behavior ? For gRPC it seems a new connection is used by default, so it would change the behavior.

costinm:
What is the purpose here ? Connection reuse seems dangerous/suspicious - tests may interfere with each other, ordering etc.

If we want to test multiple requests using same connection - maybe adding some 'repeat' or 'connection_id' that can be used explicitly for the test that is verifying something requiring reuse, but still doesn't interfere with other tests. 

I.e. I would instead add this connection_id to the ForwardRequest, and if set - use it to create or get a connection with that ID,
else make sure no connection reuse between tests.

pkg/test/echo/cmd/client/main.go
nmittler: This client app is just a main wrapper around the forwarder. This used to be the way that echo worked ... we would exec onto the pod and then run this command. TBH I'm not really sure if it's used anymore. This is just updating the app to allow configuring the new parameter. This is consistent with how all of the parameters are set by this app.

stevenctl (approved):

pkg/test/echo/server/forwarder/config.go
nmittler: Ack. This is just moved code. Not sure the original motivation here.

Issue #38083, opened 2022-03-23
Kube-inject panic when using revisions


Kalen-F:
### Bug Description

I Get a panic when I try to manual sidecar injection and revisions

My kubernetes is simple as I only do the following command

Installed istio 1-13-2 with revision
`istioctl install --set profile=demo -y --revision 1-13-2`

Installed the bookinfo sample application
`kubectl apply -f samples/bookinfo/platform/kube/bookinfo.yaml`

Finally try to do the manual injection using meshConfigMapName and injectConfigMapName
```
istioctl kube-inject -f samples/bookinfo/platform/kube/bookinfo.yaml --meshConfigMapName istio-1-13-2 --injectConfigMapName istio-sidecar-injector-1-13-2  | kubectl apply -f -

2022-03-23T00:58:42.302388Z	warn	failed to get injection config from mutatingWebhookConfigurations "istio-sidecar-injector", will fall back to get injection from the injection configmap "istio-sidecar-injector" : could not find valid mutatingWebhookConfiguration "sidecar-injector.istio.io" from cluster
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x178 pc=0x102b709b0]

goroutine 1 [running]:
istio.io/istio/pkg/kube/inject.RunTemplate({0x1400077f000, {{0x14000feb540, 0xa}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, ...}, ...}, ...})
	istio.io/istio/pkg/kube/inject/inject.go:412 +0x6f0
istio.io/istio/pkg/kube/inject.injectPod({0x1400077f000, {{0x14000feb540, 0xa}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, ...}, ...}, ...})
	istio.io/istio/pkg/kube/inject/webhook.go:330 +0xa4
istio.io/istio/pkg/kube/inject.IntoObject({0x103eea1c0, 0x14000fe8c00}, 0x1400080e660, {0x0, 0x0}, {0x0, 0x0}, 0x0, {0x103f28818, 0x140004cd680}, ...)
	istio.io/istio/pkg/kube/inject/inject.go:786 +0xcec
[...]

PR #38254, opened 2022-04-06
Fixes panic during injection from ConfigMap fallback

Kalen-F:
Fixes issue #38083 

The issue is actually more broad than the issue suggests.
As far as I can tell istio will always panic at
https://github.com/istio/istio/blob/72aa6cb678025fb34a82ab21e0a68df80f4aa25c/istioctl/cmd/kubeinject.go#L393-L398

Since it returns an injector with a nil config and nil for its meshconfig. (See issue for full explanation)
Change will require to actually grab a non nil meshconfig to use during fallback and prevent panic.

Alternate solutions would be to get rid of the fallback entirely and just display an error message, as the fallback itself in its current state seems to always fail to inject.

### DIFF
--- a/istioctl/cmd/kubeinject.go
+++ b/istioctl/cmd/kubeinject.go
@@ -393,8 +393,9 @@ func setupKubeInjectParameters(sidecarTemplate *inject.RawTemplates, valuesConfi
 			if *sidecarTemplate, err = getInjectConfigFromConfigMap(kubeconfig, revision); err != nil {
 				return nil, nil, err
 			}
+		} else {
+			return injector, nil, nil
 		}
-		return injector, nil, nil
 	}
 
 	// Get configs from IOP files firstly, and if not exists, get configs from files and configmaps.


istioctl/cmd/kubeinject.go
```diff
@@ -393,8 +393,9 @@ func setupKubeInjectParameters(sidecarTemplate *inject.RawTemplates, valuesConfi
 			if *sidecarTemplate, err = getInjectConfigFromConfigMap(kubeconfig, revision); err != nil {
 				return nil, nil, err
 			}
+		} else {
+			return injector, nil, nil
```
linsun, line 397: Could we improve error msg in this case?  something to let user know the mutating web hook config can't be found and alert user to use --revision if needed.

Kalen-F: The lack of a revision flag is not the only reason why the fallback may be taken. It does so on any error so the message would have to be more generic.

PR #38264, opened 2022-04-06
Add the prefix to td mesh name in gcp_envoy_bootstrap.json

yangson1:
TD use the following format for mesh resource and gateway resource respectively
 - mesh: /projects/project-name/networks/mesh:mesh-name/nodes/<node>
 - gateway: /projects/project-name/networks/scope:scope-name/nodes/<node>

However a valid k8s label must be an empty string or consist of alphanumeric characters, '-', '_' or '.' Thus we could not include the entire string `mesh:mesh-name`. The workaround is to add `mesh:` prefix here. And only include mesh-name in the label.


lambdai (approved):
LGTM

stevenctl (approved):

ericvn (approved):

PR #38270, opened 2022-04-06
[tf] Optimize scope close

nmittler:
We currently synchronously run all added closers, which can take a long time. Common use case for a closer is deleting a YAML file that was applied during a test.

This PR runs all closers in parallel. Significantly improves shutdown time on hosted k8s (e.g. GKE).

### DIFF
--- a/pkg/test/framework/resource/context.go
+++ b/pkg/test/framework/resource/context.go
@@ -112,7 +112,7 @@ type Context interface {
 
 	// Cleanup runs the given function when the test context completes.
 	// This function will always run, regardless of -istio.test.nocleanup. To run only when cleanup is enabled,
-	// use WhenDone.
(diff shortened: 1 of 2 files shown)


stevenctl (approved):
I think this will work, but it's possible we need to clean things up in order. We can add `CloseSync` which runs before the parallel ones.

howardjohn:
This caused a regression I think. do you have time to look into it or should we revert for now?

https://prow.istio.io/view/gs/istio-prow/logs/integ-pilot-multicluster_istio_postsubmit/1511753897725661184

Can see it gets stuck on `goroutine 8852 [select, 30 minutes]` which is the delete, so I assume this is what is blocking. It probably is a real bug in the test/Istio that its exposing though, so this is still likely a good change

PR #38274, opened 2022-04-07
chiron: add kubernetes csr version judgment function

Chen-Xintong:
This PR include two changes:

- add checkUseV1 to check kubernetes csr version first, this was also mentioned before [here](https://github.com/istio/istio/pull/33983#discussion_r670935838)

- Add the  v1Csr.Conditions object first in approveCSR()

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/security/pkg/k8s/chiron/utils.go
+++ b/security/pkg/k8s/chiron/utils.go
@@ -44,6 +44,7 @@ const (
 	csrRetriesMax = 3
 	// cert-manager use below annotation on kubernetes CSR to control TTL for the generated cert.
 	RequestLifeTimeAnnotationForCertManager = "experimental.cert-manager.io/request-duration"
+	KubernatesBuiltinSignerName             = "kubernetes.io/legacy-unknown"
 )
 
 type CsrNameGenerator func(string, string) string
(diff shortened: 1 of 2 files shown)


security/pkg/k8s/chiron/utils_test.go
```diff
@@ -397,8 +397,10 @@ func TestSubmitCSR(t *testing.T) {
 			cert.UsageClientAuth,
 		}
 
+		useV1 := true
```
zirain, line 400: can you add a negative test case?

Chen-Xintong: Sure, I'll add a test case for this scenario.

howardjohn (commented):
If I recall correctly, we either use (via library) k8s code to decide v1 or beta or copy+pasted their code. They must have had a reason for doing it that way. zi would recommend we find out why before diverging

security/pkg/k8s/chiron/utils.go
```diff
@@ -137,6 +135,13 @@ func SignCSRK8s(client clientset.Interface, csrData []byte, signerName string, u
 	return certChain, caCert, err
 }
 
+func checkUseV1(signerName string, usages []certv1.KeyUsage) bool {
+	if len(usages) > 0 && len(signerName) > 0 && signerName != KubernatesBuiltinSignerName {
```
irisdingbj, line 139: The original logic is to try v1 api first and if error, then swith to v1beta1.  So we should not change the version check logic here.

Chen-Xintong: That's make sense, thanks, I think we can close this PR because the original checking logic is correct :)

Chen-Xintong:
> If I recall correctly, we either use (via library) k8s code to decide v1 or beta or copy+pasted their code. They must have had a reason for doing it that way. zi would recommend we find out why before diverging

According to iris' comments, the original checking logic should be correct, so we can close this PR :)

PR #38282, opened 2022-04-07
Cleanup proxy view logic

nmittler:
This adds an abstraction `ProxyView` to restrict which endpoints are visible to a given gateway.

This was originally going to be used to additionally restrict views to same-cluster, but we decided against it for now.

So this is just a minor cleanup for now, but will make future changes to proxy view easier.

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -641,25 +641,11 @@ func (m NodeMetadata) ProxyConfigOrDefault(def *meshconfig.ProxyConfig) *meshcon
 	return def
 }
 
-// GetNetworkView returns the networks that the proxy requested.
-// When sending EDS/CDS-with-dns-endpoints, Pilot will only send
-// endpoints corresponding to the networks that the proxy wants to see.
+// GetView returns a restricted view of the mesh for this proxy. The view can be
+// restricted by network (via ISTIO_META_REQUESTED_NETWORK_VIEW).
 // If not set, we assume that the proxy wants to see endpoints in any network.
-func (node *Proxy) GetNetworkView() map[network.ID]bool {
-	if node == nil || node.Metadata == nil {
-		return nil
-	}
-	if len(node.Metadata.RequestedNetworkView) == 0 {
-		return nil
-	}
-
-	nmap := make(map[network.ID]bool)
-	for _, n := range node.Metadata.RequestedNetworkView {
-		nmap[network.ID(n)] = true
-	}
-	nmap[identifier.Undefined] = true
-
-	return nmap
+func (node *Proxy) GetView() ProxyView {
+	return newProxyView(node)
 }
 
 // InNetwork returns true if the proxy is on the given network, or if either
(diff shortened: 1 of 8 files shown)


howardjohn:
I don't think we need this. The gateway is a multi *network* gateway - not multi cluster. You do not need one in every cluster, just every network, so this will break users with those setups. Additionally, there is no reason I know of that you would want to keep traffic in the same cluster, as there is not expected to be additional cost cross clusters.

The "no L7 load balancing" problem wider than this, and we only fix 1 minor edge case in it that happens to be exposed in our e2e tests. If we had a test that was "assert all pods are reached" and had multi-replica pods, you would see similar failures even with this change

nmittler:
@howardjohn 

> The "no L7 load balancing" problem wider than this, and we only fix 1 minor edge case in it that happens to be exposed in our e2e tests. If we had a test that was "assert all pods are reached" and had multi-replica pods, you would see similar failures even with this change.

Agree in general, but it's not about fixing an edge case ... it's about fixing a flake that is preventing forward progress. The option won't break anyone unless they manually set, so we can just not document it for now if you'd prefer?

howardjohn:
I don't think we should make non-trivial changes to our product which allow our tests to assert incorrect invariants about the system (that load will be equally balanced). This also means our tests would be running code entirely different than real users, only so our incorrect assertion can pass. It doesn't seem like the right path

howardjohn:
Can we change the test to assert the actual reality of how Istio works? That is you can expect to reach all clusters in your network, and all networks, but not "all clusters" as asserted today

nmittler:
@howardjohn agreed. I've reduced the scope of this PR to just a bit of code cleanup. I'll open another PR to change the way we check for reached clusters in our tests.

nmittler:
@howardjohn the changes to the test framework are in https://github.com/istio/istio/pull/38315

pilot/pkg/model/context.go
```diff
@@ -641,25 +641,12 @@ func (m NodeMetadata) ProxyConfigOrDefault(def *meshconfig.ProxyConfig) *meshcon
...
-	nmap := make(map[network.ID]bool)
-	for _, n := range node.Metadata.RequestedNetworkView {
-		nmap[network.ID(n)] = true
-	}
-	nmap[identifier.Undefined] = true
-
-	return nmap
+// GetView returns a restricted view of the mesh for this proxy. The view can be
+// restricted by network and cluster (via ISTIO_META_REQUESTED_NETWORK_VIEW, and
+// ISTIO_META_REQUESTED_CLUSTER_VIEW, respectively).
```
hzxuzhonghu, line 646: Do we have this env? I donot think so

pilot/pkg/xds/ep_filters.go
```diff
@@ -67,6 +67,13 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocLbEndpointsAn
 
 		// Process all of the endpoints.
 		for i, lbEp := range ep.llbEndpoints.LbEndpoints {
+			istioEndpoint := ep.istioEndpoints[i]
+
+			// If the proxy can't view the network for this endpoint, exclude it entirely.
+			if !b.proxyView.IsVisible(istioEndpoint) {
```
howardjohn, line 73: This seems correct -- but also changes L80-86 to not mutate the weight.

Is this fixing a bug? Seems like it might be ... Or somehow it has no impact?

howardjohn (approved):

pilot/pkg/xds/ep_filters.go
nmittler: My thought as well. There hasn't been a bug filed that I'm aware of ... and even if there was a bug, it would likely not be very noticeable. However, I believe this code is (more) correct.

Issue #27420, opened 2020-09-18
Make istiod-less Remote Cluster the default for multicluster


nmittler:
For 1.8 (Multicluster Beta), we want to switch the default remote cluster type to be istiod-less. We also want to deprecate the old remote cluster type.

All tests should use the istiod-less remote by default as well.

[ ] Docs
[x] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Additional context**

stevenctl:
@irisdingbj @linsun can you point me to a doc or help me understand why we need to set `ISTIOD_CUSTOM_HOST`?

linsun:
@stevenctl ISTIOD_CUSTOM_HOST is an env var that is used to generate the webhook certificate for istiod to include the host.  This is useful when the host is not from a trust-able authority.

linsun:
Update on this: https://github.com/istio/istio/issues/27921

irisdingbj:
https://github.com/istio/istio/pull/28181/

ericvn:
@irisdingbj @nmittler  Should #28181 have satisfied this issue and it can be closed?

stevenctl:
No, we don't have a documented path for users to enable this in production. What we're doing in integration tests is a bit non-standard and actually is the cause of some of the flakes we see.

brian-avery:
@stevenctl @nmittler Who's doing the work here? What work is needed? This is marked as a p0 for 1.10. Is that still accurate?

stevenctl:
The work required:

* Tests
  * We have the capability to make ITs use this model, but that uses some hacks. If we can eliminate the CUSTOM_HOST thing by installing custom certs on our test containers, that may work.
  * We'll likely want to run 2 prow jobs for networking test, one that has this model and one with the old model of remotes. 
* Docs for how to set this up as a user.

As far as priority level, who is going to work on this and when, I'm not sure.

nmittler:
@linsun @GregHanson IBM had been leading this work previously. Is there any plan to help drive the feature to Beta?

nmittler:
not stale

ryantking:
@nmittler @linsun @irisdingbj @stevenctl Any opposition moving this into the 1.12 milestone?

Kmoneal:
@linsun @irisdingbj @stevenctl Any updates for getting this into 1.12?

[...]

PR #38283, opened 2022-04-07
Make istiod manage istiodless remote clusters by default

frankbu:

Changes needed to make istiodless remotes work by default.

Part of closing #27420 

Depends on #38219

Doc PR with corresponding multicluster doc changes https://github.com/istio/istio.io/pull/11155


pilot/pkg/features/pilot.go
```diff
@@ -383,7 +383,7 @@ var (
 	ClusterName = env.RegisterStringVar("CLUSTER_ID", "Kubernetes",
 		"Defines the cluster and service registry that this Istiod instance is belongs to").Get()
 
-	ExternalIstiod = env.RegisterBoolVar("EXTERNAL_ISTIOD", false,
+	ExternalIstiod = env.RegisterBoolVar("EXTERNAL_ISTIOD", true,
```
stevenctl, line 386: I didn't keep up with the leader election changes, but IIUC, turning this on means we  will join LE to patch the `istio-ca-root-cert` configmap. We don't differentiate a "istiodless remote" from a "remote primary" cluster, but our leader election logic will make it so the "most local" istiod always wins. _Is this correct?_

We also have a "webhook cert patcher". and a "validating webhook controller". Those may need to be guarded by leader election as well?

stevenctl:
My main concern with making this default is that we do a lot in test code which isn't really documented and we would never expect users to do. (i.e. [manual Endpoints creation](https://github.com/istio/istio/blob/37bc0851bc8a1bb2e13572f229760931dcca4b5c/pkg/test/framework/components/istio/operator.go#L1000)). 

I fear this will mask install-time issues and lead to lots of requests for support or concerns that `primary-remote` install is just broken.

stevenctl:
For things that _must_ be worked around for tests, we should at least add a comment giving context (and maybe link to a Github issue around fixing whatever is missing in our test infra). One example of something that's in tests but not the docs (injectionURL vs injectionPath):

https://github.com/istio/istio/blob/37bc0851bc8a1bb2e13572f229760931dcca4b5c/pkg/test/framework/components/istio/operator.go#L634-L635

pilot/pkg/features/pilot.go
frankbu: > We don't differentiate a "istiodless remote" from a "remote primary" cluster, but our leader election logic will make it so the "most local" istiod always wins. Is this correct?

Yes, that is correct.

> We also have a "webhook cert patcher". and a "validating webhook controller". Those may need to be guarded by leader election as well?

I had assumed that the istiod that gets the leader lock would be the one doing all these things, but maybe I'm missing something? Can you point me at the code where I should check and fix if necessary?

frankbu:
@stevenctl 

> My main concern with making this default is that we do a lot in test code which isn't really documented and we would never expect users to do. (i.e. [manual Endpoints creation](https://github.com/istio/istio/blob/37bc0851bc8a1bb2e13572f229760931dcca4b5c/pkg/test/framework/components/istio/operator.go#L1000)).

We do have a comment in the test framework to tell people that [users should do it properly](https://github.com/istio/istio/blob/37bc0851bc8a1bb2e13572f229760931dcca4b5c/pkg/test/framework/components/istio/operator.go#L927).

Also, in the external control plane doc we show how to [configure a proper URL for the remote istiod](https://istio.io/latest/docs/setup/install/external-controlplane/#set-up-the-remote-config-cluster). The Primary-Remote MC doc uses the Service/Endpoints approach (triggered by `values.global.remotePilotAddress`). I [haven't changed that in the doc](https://github.com/istio/istio.io/pull/11155), I just changed the Service and Endpoints yaml charts in this PR to include the webhook port (15017) which is needed if the remote cluster is istiodless.

frankbu:
@nmittler 

I don't think there's a doc, so let me summarize the latest status:

1. istiodless remote clusters have been working for several months now. They are used for both external control plane integration testing and in the external control plane user doc. We install Istio on the remote cluster using the `external` profile (we had to change the profile name from `remote` for backward-compatibility and deprecated `remote` back in 1.10).
2. We have also been running postsubmit multicluster integration tests with istiodless remotes for a while now as well. They work, but up until now there was an issue where we didn't support location-based leader election, so in the Primary-Primary case we didn't have any assurance that the local istiod would be the preferred leader in each primary cluster. That's been fixed now: #38219.
3. The last thing we need to do is update the MC docs to use istiodless remotes (i.e., install with the `external` profile instead of the `default` profile). The doc PR with those changes is here: https://github.com/istio/istio.io/pull/11155. It requires the changes in this PR that 1) add the webhook port 15017 to the Service/Endpoint that is used when `values.global.remotePilotAddress` is set and 2) makes the pilot's `EXTERNAL_ISTIOD` default to `true` so istiod will always try to get the leader lock for clusters it gets a remote key for. If you or @stevenctl feel the second part is risky, I can take it out of this PR (leave the default as `false`) and instead change the doc to explicitly set it to `true` when installing the primary.

nmittler:
@frankbu thanks for the overview

@frankbu @stevenctl are we in agreement that istiod-less remotes are ready to go GA? I think that would be the implication here by changing these docs. Did it go through the graduation process?

frankbu:
@nmittler I don't think this implies that we want to promote to GA. I think istiodless was supposed to be working when MC was promoted to Beta a while ago, with the non-istiodless remotes as a known issue. This is the issue I was hoping to close with this and the corresponding doc PR: https://github.com/istio/istio/issues/27420

pilot/pkg/features/pilot.go
stevenctl: https://github.com/istio/istio/blob/276fb292f30e3f20cc036fabba6328b76291c781/pilot/pkg/serviceregistry/kube/controller/multicluster.go#L235

There are a flew blocks in this method that check `features.ExternalIstiod` which. would be affected by this PR.

stevenctl:
I think as long as we're certain multi-primary istiods won't start stomping on each-other I'm good to merge this. The last thing is just making sure a few other controllers are leader elected. 

Should we do the init/teardown for these as part of a single leader-lock/an existing one, or is there a reason we'd want them to be independent?

stevenctl:
After thinking about it some more, I want to at the very least document _why_ we must do things differently in our tests than we want users to do. My gut feeling is we should be able to workaround it pretty easily in our test infra.

frankbu:
> After thinking about it some more, I want to at the very least document why we must do things differently in our tests than we want users to do. My gut feeling is we should be able to workaround it pretty easily in our test infra.

@stevenctl I'm not sure I'm following what you're suggesting. Can you elaborate? We don't really have any documentation for the integration tests, so I'm not sure where we would document that it does things differently and why.

stevenctl:
This PR changes some templates, but we then just manually create k8s objects in the framework. I would prefer if we got coverage of the changes we're making here. 

---

https://github.com/istio/istio/blob/37bc0851bc8a1bb2e13572f229760931dcca4b5c/pkg/test/framework/components/istio/operator.go#L927

This comment simply says "users will do things differently" but doesn't explain why the tests can't do the same. 
I don't really know why we must use `injectionUrl` here instead of using `injectionPath` and rely on the Service+Endpoints created in the template (like the docs PR does). 

My suggestion: 
* Start by avoiding `configureRemoteClusterDiscovery`
* Attempt to switch to `injectionPath` similar to docs. If it fails, I won't block this PR but we can better document the challenge in making it work. 

The docs test _will_ cover the install, but in the future as things shift around there is a lag between breaking something in istio/istio and seeing that our install steps no longer work. In the past this has led to tons of issues along the lines of "multicluster install is broken" being opened.

frankbu:
@stevenctl

> Start by avoiding configureRemoteClusterDiscovery

That absolutely makes sense, and should work now that I've fixed the Service/Endpoints templates to support more than just the discorvery port.

> Attempt to switch to injectionPath similar to docs. If it fails, I won't block this PR but we can better document the challenge in making it work.

I'm pretty sure this should just work. The reason it's using injectionURL currently, is because that's how it was originally written and changing it was out of scope for the external control plane changes I made which don't require the `ISTIOD_CUSTOM_HOST` header hack.

I would prefer to merge this PR first, and then make these fixes in a separate framework cleanup PR, unless you feel strongly that it should be done as a prereq to merging this one.

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -219,40 +218,39 @@ func (m *Multicluster) ClusterAdded(cluster *multicluster.Cluster, clusterStopCh
 					// recreate it again.
 					client.RunAndWait(clusterStopCh)
 					nc.Run(leaderStop)
-				}).Run(clusterStopCh)
+				})
+			// Set up injection webhook patching for remote clusters we are controlling.
+			// The local cluster has this patching set up elsewhere. We may eventually want to move it here.
```
frankbu, line 223: @stevenctl with this code added here, do you think we can remove it from the init code?
https://github.com/istio/istio/blob/ac5ae313be27941e6a2eb59f24cf68db2b9a5813/pilot/pkg/bootstrap/sidecarinjector.go#L89

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -219,40 +218,39 @@ func (m *Multicluster) ClusterAdded(cluster *multicluster.Cluster, clusterStopCh
...
+			if features.ExternalIstiod && !localCluster && m.caBundleWatcher != nil {
```
frankbu, line 224: Remove guard from here if we delete patching setup in bootstrap/sidecarinjector.go.
```suggestion
			if m.caBundleWatcher != nil {
```

pilot/pkg/serviceregistry/kube/controller/multicluster.go
frankbu: I tried removing this but it didn't work. Lots of problems with revision tests.

frankbu:
> I think it's preferable that we get things tested the way we want before merging docs PRs. If the work in istio/istio is split up that's fine.

@stevenctl Fair enough. If you are OK to merge this one now, I'll start on a second istio/istio cleanup PR right away. I'm also OK with not trying to get this (or the doc PR) into 1.14. We're OK with waiting for 1.15, no need to cherry pick any of this - just would like to make progress in the master branch.

stevenctl (approved):

PR #38286, opened 2022-04-08
added new authz tests

aryan16:
**Added new authz tests to improve the test coverage**
AuthZ.rules.to.operation.notPorts
AuthZ.rules.to.operation.notHosts
AuthZ.rules.to.operation.notMethods
Authz.rules.from.source.notRequestPrincipals
AuthZ.rules.from.source.ipBlocks
AuthZ.rules.from.source.notIpBlocks

### DIFF
--- a/tests/integration/security/authorization_test.go
+++ b/tests/integration/security/authorization_test.go
@@ -462,13 +462,11 @@ func TestAuthorization_NegativeMatch(t *testing.T) {
 				host   string // only set if not empty
 				port   string // default value is http if not specified
 			}
-
+			reqParam := request{method: "GET", port: "http"}
+			reqNotMethodParam := request{method: "PUT", port: "http"}
+			reqNotPortParam := request{method: "GET", port: "http-8091"}
+			reqNotHostParam := request{method: "GET", port: "http", host: "deny.com"}
 			for _, srcCluster := range t.Clusters() {
-				reqParam := request{method: "GET", port: "http"}
-				reqNotMethodParam := request{method: "PUT", port: "http"}
-				reqNotPortParam := request{method: "GET", port: "http-8091"}
-				reqNotHostParam := request{method: "GET", port: "http", host: "deny.com"}
-
 				a := match.And(match.Cluster(srcCluster), match.Namespace(apps.Namespace1.Name())).GetMatches(apps.A)
 				bInNS2 := match.And(match.Cluster(srcCluster), match.Namespace(apps.Namespace2.Name())).GetMatches(apps.B)
 				if len(a) == 0 || len(bInNS2) == 0 {
@@ -476,7 +474,7 @@ func TestAuthorization_NegativeMatch(t *testing.T) {
 				}
 
 				t.NewSubTestf("From %s", srcCluster.StableName()).Run(func(t framework.TestContext) {
-					newTestCase := func(from echo.Instance, to echo.Target, path string, expectAllowed bool, request request) func(t framework.TestContext) {
+					newTestCaseWithRequest := func(from echo.Instance, to echo.Target, path string, expectAllowed bool, request request) func(t framework.TestContext) {
 						callCount := util.CallsPerCluster * to.WorkloadsOrFail(t).Len()
 						return func(t framework.TestContext) {
 							opts := echo.CallOptions{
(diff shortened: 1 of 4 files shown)


tests/integration/security/testdata/authz/v1beta1-negative-match.yaml.tmpl
```diff
@@ -34,6 +49,22 @@ spec:
     - source:
         notNamespaces: ["{{ .Namespace }}"]
 ---
+# The following policy won't allow request to host "deny.com"
```
yangminzhu, line 52: `# The following policy allows request if the host is not "deny.com"` and add a newline below (just to be consistent with other examples in this file)

tests/integration/security/testdata/authz/v1beta1-negative-match.yaml.tmpl
```diff
@@ -16,9 +16,24 @@ spec:
...
+    matchLabels:
+      "app": "{{ .b }}"
+  action: ALLOW
+  rules:
+  - to:
+    - operation:
+        notMethods : ["GET"]
+---
 # The following policy denies access from other namespaces.
-
```
yangminzhu, line 21: revert this line.

tests/integration/security/testdata/authz/v1beta1-negative-match.yaml.tmpl
```diff
@@ -52,7 +83,23 @@ spec:
     - source:
         notPrincipals: ["*"]
 ---
+# The following policy won't allow request to port 8091
```
yangminzhu, line 86: `# The following policy allows request if the port is not "8091"`

tests/integration/security/authorization_test.go
```diff
@@ -494,46 +499,52 @@ func TestAuthorization_NegativeMatch(t *testing.T) {
...
 						// - path `/allow` should be allowed implicitly.
-						newTestCase(a[0], b, "/prefix", false),
-						newTestCase(a[0], b, "/prefix/other", false),
-						newTestCase(a[0], b, "/prefix/allowlist", true),
-						newTestCase(a[0], b, "/allow", true),
-						newTestCase(bInNS2[0], b, "/prefix", false),
-						newTestCase(bInNS2[0], b, "/prefix/other", false),
-						newTestCase(bInNS2[0], b, "/prefix/allowlist", true),
-						newTestCase(bInNS2[0], b, "/allow", true),
+						newTestCase(a[0], b, "/", false, "GET", "", "http"),
```
yangminzhu, line 507: the argument list of the function is way too long and you only need to customize the method/port/host for 1-2 test cases but need to duplicate the parameter in all other places, this is not good.

Can you refactor to pass a struct to the NewTestCase? For example

```
type struct request{
  method string // default value is GET if not specified
  host string / only set if not empty
  port string // default value is http if not specified
  path string // should always be set
}
```

You only specify the field in the struct if it's different from the default value.

tests/integration/security/authorization_test.go
```diff
@@ -455,24 +457,37 @@ func TestAuthorization_NegativeMatch(t *testing.T) {
...
 			}, "testdata/authz/v1beta1-negative-match.yaml.tmpl").ApplyOrFail(t)
 
+			type request struct {
+				method string // default value is GET if not specified
+				host   string // only set if not empty
+				port   string // default value is http if not specified
+			}
+
 			for _, srcCluster := range t.Clusters() {
+				reqParam := request{method: "GET", port: "http"}
```
yangminzhu, line 467: these can be defined out of the for-loop.

tests/integration/security/authorization_test.go
```diff
@@ -455,24 +457,37 @@ func TestAuthorization_NegativeMatch(t *testing.T) {
...
+
 				a := match.And(match.Cluster(srcCluster), match.Namespace(apps.Namespace1.Name())).GetMatches(apps.A)
 				bInNS2 := match.And(match.Cluster(srcCluster), match.Namespace(apps.Namespace2.Name())).GetMatches(apps.B)
 				if len(a) == 0 || len(bInNS2) == 0 {
 					continue
 				}
 
 				t.NewSubTestf("From %s", srcCluster.StableName()).Run(func(t framework.TestContext) {
-					newTestCase := func(from echo.Instance, to echo.Target, path string, expectAllowed bool) func(t framework.TestContext) {
+					newTestCase := func(from echo.Instance, to echo.Target, path string, expectAllowed bool, request request) func(t framework.TestContext) {
```
yangminzhu, line 479: you can create another wrapper function: `newTestCaseWithRequest(from echo.Instance, to echo.Target, path string expectAllowed bool, req *request)` and then change `newTestCase` to call it with the `reqParam` so that you can just pass the `reqParam` variable once and do not need to repeat it everywhere.

yangminzhu (commented):
Looks good, just a few small comments.

yangminzhu (approved):

Issue #38201, opened 2022-03-31
`golint` is deprecated. Replace with `revive`


ericvn:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**

From the linting logs:
```
WARN [runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive.
```
**Describe alternatives you've considered**

The accepted replacement for glint seems to be reviewing as noted.

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

I found one nolint: golint in `release-builder` and a number of them in `istio`.
I did to a quick sniff test. Even replacing `nolint: golint...` with `nolint: golint, revive...` (temporarily just add `revive` and remove `golint`   after the common-files are updated), there are a number of other failures noted in istio/istio and possible the other repos.

Once given the go-ahead I think the remaining changes can be made in the repos, the changes checked in, common-files updated, and then the nolint:golint removed.

zirain:
@ericvn I think it's ok to close this, WDYT?

ericvn:
Yes. The only thing we might want to do, but not sure, would be to remove the `nolint: golint` tag now that the build-tools are updated.

zirain:
> Yes. The only thing we might want to do, but not sure, would be to remove the `nolint: golint` tag now that the build-tools are updated.

I forgot that, will do that tomorrow.

zirain:
closed as everything might be ok now.

PR #38287, opened 2022-04-08
replace golint with revive

zirain:

part of https://github.com/istio/istio/issues/38201

### DIFF
--- a/common/config/.golangci.yml
+++ b/common/config/.golangci.yml
@@ -12,8 +12,8 @@ run:
   # timeout for analysis, e.g. 30s, 5m, default is 1m
   deadline: 20m
   build-tags:
-    - integ
-    - integfuzz
+  - integ
+  - integfuzz
   # which dirs to skip: they won't be analyzed;
   # can use regexp here: generated.*, regexp is applied on full path;
   # default value is empty list, but next dirs are always skipped independently
@@ -34,25 +34,25 @@ run:
 linters:
   disable-all: true
   enable:
-    - deadcode
-    - errcheck
-    - exportloopref
-    - gocritic
-    - gofumpt
-    - goimports
-    - revive
-    - gosimple
-    - govet
-    - ineffassign
-    - lll
-    - misspell
-    - staticcheck
-    - structcheck
-    - stylecheck
-    - typecheck
-    - unconvert
-    - unparam
-    - varcheck
+  - deadcode
+  - errcheck
+  - exportloopref
+  - gocritic
+  - gofumpt
+  - goimports
+  - golint
+  - gosimple
+  - govet
+  - ineffassign
+  - lll
+  - misspell
+  - staticcheck
+  - structcheck
+  - stylecheck
+  - typecheck
+  - unconvert
+  - unparam
+  - varcheck
   fast: false
 
 linters-settings:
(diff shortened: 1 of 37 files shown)


common/config/.golangci.yml
```diff
@@ -67,9 +67,67 @@ linters-settings:
   govet:
     # report about shadowed variables
     check-shadowing: false
-  golint:
-    # minimal confidence for issues, default is 0.8
-    min-confidence: 0.0
+  revive:
```
zirain, line 70: these changes will be move to common-files

common/config/.golangci.yml
```diff
@@ -67,9 +67,67 @@ linters-settings:
...
+      - name: atomic
+      - name: empty-lines
+      - name: call-to-gc
+      - name: duplicated-imports
+      - name: string-of-int
+      - name: defer
+        arguments: [["call-chain"]]
+      - name: unconditional-recursion
+      - name: identical-branches
+      # - name: confusing-results
```
zirain, line 115: these rules contains too much noise, maybe we can enable them in the future?

common/config/.golangci.yml
zirain: moved to https://github.com/istio/common-files

common/config/.golangci.yml
zirain: moved to https://github.com/istio/common-files

ericvn (commented):
Looks good other than a few nits about what looks like inconsistent empty lines.

howardjohn (changes requested):
Can we just move to the new linter but set the default golint rules? https://revive.run/docs#configuration

Some of these changes look worse to me and I would prefer we introduce new linter 1-by-1 so we can evaluate the impact.

My understanding is our goal with revive is to use a non-deprecated linter, not to introduce new rules?

ericvn:
> Can we just move to the new linter but set the default golint rules? https://revive.run/docs#configuration
> 
> Some of these changes look worse to me and I would prefer we introduce new linter 1-by-1 so we can evaluate the impact.
> 
> My understanding is our goal with revive is to use a non-deprecated linter, not to introduce new rules?

When I did my initial testing, I just renamed the `golint:` section `revive: ` and noted many of the same changes that are done here (~37 files seems familiar). I would be curious if these are indeed the same and these are all required with just the default.

zirain:
> Can we just move to the new linter but set the default golint rules? https://revive.run/docs#configuration

It's hard to do that.

> Some of these changes look worse to me and I would prefer we introduce new linter 1-by-1 so we can evaluate the impact.

can you point out which part became worse? we can disabled that rule.

pkg/config/validation/validation.go
```diff
@@ -2359,9 +2359,9 @@ func analyzeUnreachableHTTPRules(routes []*networking.HTTPRoute,
 				duplicateMatches++
 				// no need to handle for totally duplicated match rules
 				continue
-			} else {
```
zirain, line 2362: [revie rule]superfluous-else: if block ends with a continue statement, so drop this else and outdent its block

tools/bug-report/pkg/cluster/cluster.go
```diff
@@ -69,18 +69,18 @@ func shouldSkip(deployment string, config *config2.BugReportConfig, pod *corev1.
...
+			for key, val := range eld.Labels {
+				if evLablel, exists := pod.Labels[key]; exists {
+					if isExactMatchedOrPatternMatched(val, evLablel) {
 						return true
 					}
 				}
 			}
 		}
 		if len(eld.Annotations) > 0 {
-			for kAnnotation, vAnnotation := range eld.Annotations {
```
zirain, line 81: var-naming: don't use leading k in Go names; range var kAnnotation should be annotation

security/pkg/pki/util/san_test.go
```diff
@@ -86,7 +86,7 @@ func TestBuildAndExtractIdentities(t *testing.T) {
 		t.Errorf("Unmatched identities: before encoding: %v, after decoding %v", ids, actualIds)
 	}
 
-	if san.Critical == false {
+	if !san.Critical {
```
zirain, line 89: bool-literal-in-expr: omit Boolean literal in expression

security/pkg/pki/ra/k8s_ra_test.go
```diff
@@ -57,9 +57,9 @@ e+5z6MTAO6ktvHdQlSuH6ARn47bJrZOlkttAhg==
 )
 
 var (
-	testCsrHostName       string = spiffe.Identity{TrustDomain: "cluster.local", Namespace: "default", ServiceAccount: "bookinfo-productpage"}.String()
-	TestCACertFile        string = "../testdata/example-ca-cert.pem"
-	mismatchCertChainFile string = "../testdata/cert-chain.pem"
+	testCsrHostName       = spiffe.Identity{TrustDomain: "cluster.local", Namespace: "default", ServiceAccount: "bookinfo-productpage"}.String()
+	TestCACertFile        = "../testdata/example-ca-cert.pem"
```
zirain, line 61: var-declaration: should omit type string from declaration of var TestCACertFile; it will be inferred from the right-hand side

howardjohn (approved):

ericvn (approved):
Comments fixed. LGTM

hanxiaop (approved):

PR #38288, opened 2022-04-08
use deployment's namespace for AdmissionRequest in istioctl kube-inject

irisdingbj:
This fix will make `istioctl kube-inject` respect proxyconfig CR if the deployment has namespace defined

### DIFF
--- a/istioctl/cmd/kubeinject.go
+++ b/istioctl/cmd/kubeinject.go
@@ -71,7 +71,7 @@ type ExternalInjector struct {
 	injectorAddress string
 }
 
-func (e ExternalInjector) Inject(pod *corev1.Pod) ([]byte, error) {
+func (e ExternalInjector) Inject(pod *corev1.Pod, namespace string) ([]byte, error) {
 	cc := e.clientConfig
 	if cc == nil {
 		return nil, nil
@@ -138,6 +138,9 @@ func (e ExternalInjector) Inject(pod *corev1.Pod) ([]byte, error) {
 		},
 	}
 	podBytes, err := json.Marshal(pod)
+	if pod.Namespace != "" {
+		namespace = pod.Namespace
+	}
 	if err != nil {
 		return nil, err
 	}
@@ -159,7 +162,7 @@ func (e ExternalInjector) Inject(pod *corev1.Pod) ([]byte, error) {
 			RequestResource:    nil,
 			RequestSubResource: "",
 			Name:               pod.Name,
-			Namespace:          pod.Namespace,
+			Namespace:          namespace,
 		},
 		Response: nil,
 	}
--- a/istioctl/cmd/kubeinject.go
+++ b/istioctl/cmd/kubeinject.go
@@ -71,7 +71,7 @@ type ExternalInjector struct {
 	injectorAddress string
 }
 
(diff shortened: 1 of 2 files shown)


zirain:
what about root namespace?

irisdingbj:
> what about root namespace?

root namespace works well.  It has already been take care via https://github.com/istio/istio/blob/master/pkg/kube/inject/webhook.go#L876

istioctl/cmd/kubeinject.go
```diff
@@ -138,6 +138,9 @@ func (e ExternalInjector) Inject(pod *corev1.Pod) ([]byte, error) {
 		},
 	}
 	podBytes, err := json.Marshal(pod)
+	if pod.Namespace != "" {
+		namespace = pod.Namespace
```
hzxuzhonghu, line 142: When pod.Namespace can be empty?

hzxuzhonghu: IC,  You can set namespace to the pod before passing in. And then without changing this function

irisdingbj: the pod spec in the injected yaml will be changed if we set namespace to the pod before passing in.  - THis will be a behaviour change.

irisdingbj: one example as below:
```

apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  name: hello
  namespace: test
spec:
  replicas: 7
  selector:
    matchLabels:
      app: hello
      tier: backend
      track: stable
  strategy: {}
  template:
    metadata:
      annotations:
        kubectl.kubernetes.io/default-container: hello
        kubectl.kubernetes.io/default-logs-container: hello
        prometheus.io/path: /stats/prometheus
        prometheus.io/port: "15020"
        prometheus.io/scrape: "true"
        sidecar.istio.io/status: '{"initContainers":["istio-init"],"containers":["istio-proxy"],"volumes":["workload-socket","workload-certs","istio-envoy","istio-data","istio-podinfo","istio-token","istiod-ca-cert"],"imagePullSecrets":null,"revision":"default"}'
      creationTimestamp: null
      labels:
        app: hello
        security.istio.io/tlsMode: istio
        service.istio.io/canonical-name: hello
        service.istio.io/canonical-revision: latest
        tier: backend
        track: stable
      **namespace: test**

```
The ending namespace: test has been added if we set pod.ns -- this is a behaviour change as previous injected yaml will not have this line.

hzxuzhonghu: So any bad effects?

irisdingbj: not sure exactlly but there is no such namespace line in a normal deployment yaml.    And a bunch of test data needs be changed if we go with pod.ns way.

By the way,  let the adminRequest carry the  namespace info is also the behaviour of current webhook. So this change actually aligns them.

hzxuzhonghu: make sense. 
I am ok if rename the namespace to deploy namespace.

irisdingbj: done, PTAL, Thanks!

hzxuzhonghu (approved):

zhlsunshine:
LGTM

zhlsunshine (approved):
LGTM

PR #38289, opened 2022-04-08
minor sni fixes

ramaraochavali:
- Builds automTLS SNI only for ISTIO_MUTUAL
- use opts.istioMtlsSni instead of cluster name for SNI

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -172,10 +172,8 @@ func (cb *ClusterBuilder) buildSubsetCluster(opts buildClusterOpts, destRule *co
 	proxyNetworkView map[network.ID]bool) *cluster.Cluster {
 	opts.serviceMTLSMode = cb.req.Push.BestEffortInferServiceMTLSMode(subset.GetTrafficPolicy(), service, opts.port)
 	var subsetClusterName string
-	var defaultSni string
 	if opts.clusterMode == DefaultClusterMode {
 		subsetClusterName = model.BuildSubsetKey(model.TrafficDirectionOutbound, subset.Name, service.Hostname, opts.port.Port)
-		defaultSni = model.BuildDNSSrvSubsetKey(model.TrafficDirectionOutbound, subset.Name, service.Hostname, opts.port.Port)
 	} else {
 		subsetClusterName = model.BuildDNSSrvSubsetKey(model.TrafficDirectionOutbound, subset.Name, service.Hostname, opts.port.Port)
 	}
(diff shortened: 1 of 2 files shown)


pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1020,9 +1021,9 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 				ValidationContextSdsSecretConfig: authn_model.ConstructSdsSecretConfig(authn_model.SDSRootResourceName),
 			},
 		}
-		// Set default SNI of cluster name for istio_mutual if sni is not set.
+		// Set default SNI for istio_mutual if sni is not set.
 		if len(tlsContext.Sni) == 0 {
-			tlsContext.Sni = c.cluster.Name
+			tlsContext.Sni = opts.istioMtlsSni
```
hzxuzhonghu, line 1026: istioMtlsSni can be empty and what's the behavior of envoy. Does it send no sni?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -211,10 +209,12 @@ func (cb *ClusterBuilder) buildSubsetCluster(opts buildClusterOpts, destRule *co
 
 	// Apply traffic policy for subset cluster with the destination rule traffic policy.
 	opts.mutable = subsetCluster
-	opts.istioMtlsSni = defaultSni
 
 	// If subset has a traffic policy, apply it so that it overrides the destination rule traffic policy.
 	opts.policy = MergeTrafficPolicy(opts.policy, subset.TrafficPolicy, opts.port)
+	if opts.policy != nil && opts.policy.GetTls().GetMode() == networking.ClientTLSSettings_ISTIO_MUTUAL {
+		opts.istioMtlsSni = model.BuildDNSSrvSubsetKey(model.TrafficDirectionOutbound, subset.Name, service.Hostname, opts.port.Port)
```
hzxuzhonghu, line 216: I think this is not backward compatible if the server relies on this value.

ramaraochavali: We are not using this field while building upstreamTLSContext.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
ramaraochavali: yes. It does. But we should not use cluster name here though - it can be different from istioMTLSSni

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
hzxuzhonghu: OK, i read wrong

hzxuzhonghu:
@ramaraochavali What's the relationship with auto sni? You know that can be enabled by default in some time

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1020,9 +1021,9 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 				ValidationContextSdsSecretConfig: authn_model.ConstructSdsSecretConfig(authn_model.SDSRootResourceName),
 			},
 		}
-		// Set default SNI of cluster name for istio_mutual if sni is not set.
+		// Set default SNI for istio_mutual if sni is not set.
 		if len(tlsContext.Sni) == 0 {
```
hzxuzhonghu, line 1025: For istio mtls we always set sni from `BuildDNSSrvSubsetKey`? So is this redundant

ramaraochavali: That is what I am checking

ramaraochavali: Looks like this is not needed

ramaraochavali:
Looks like this is not safe

PR #38290, opened 2022-04-08
minor test fixes

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ericvn:
These don't look like test fixes. It looks like the removal of tests.

ramaraochavali:
No. It is just setting the VerifyAtClientCert feature back to its old value after test - otherwise we will run in test flakiness if we do not consider old value and the default is changed in future.

ericvn (approved):

ramaraochavali:
@hzxuzhonghu can you PTAL?

hzxuzhonghu (approved):

PR #38307, opened 2022-04-08
istiod async changes

aryan16:
**Istiod Pilot Changes**
1. Update the Istiod pilot logic to make async calls to jwksuri and removed the calls from main flow.
2. Updated unit tests to support the new way of fetching public key from jwksuri

### DIFF
--- a/pilot/pkg/model/jwks_resolver.go
+++ b/pilot/pkg/model/jwks_resolver.go
@@ -396,7 +396,9 @@ func (r *JwksResolver) refresher() {
 			r.refreshTicker.Stop()
 			return
 		case <-jwksuriChannel:
+			r.retryInterval = 10 * time.Millisecond
 			lastHasError = r.refreshCache(lastHasError)
+			r.retryInterval = JwtPubKeyRetryInterval
 		}
 	}
 }
@@ -575,16 +577,4 @@ func compareJWKSResponse(oldKeyString string, newKeyString string) (bool, error)
 	// If we aren't able to compare using keys, we should return true
 	// since we already checked exact equality of the responses
 	return true, nil
-}
-
-func (r *JwksResolver) CheckPubKeyExistInCache(issuer string, jwksURI string) bool {
-	key := jwtKey{issuer: issuer, jwksURI: jwksURI}
-	if val, found := r.keyEntries.Load(key); found {
-		e := val.(jwtPubKeyEntry)
-		r.keyEntries.Store(key, e)
-		if e.pubKey != "" {
-			return true
-		}
-	}
-	return false
 }
--- a/pilot/pkg/model/jwks_resolver.go
+++ b/pilot/pkg/model/jwks_resolver.go
@@ -372,6 +372,7 @@ func (r *JwksResolver) getRemoteContentWithRetry(uri string, retry int) ([]byte,
 
 	for i := 0; i < retry; i++ {
 		body, err := getPublicKey()
+		fmt.Println(r.retryInterval)
 		if err == nil {
 			return body, nil
 		}
@@ -396,9 +397,7 @@ func (r *JwksResolver) refresher() {
 			r.refreshTicker.Stop()
 			return
 		case <-jwksuriChannel:
-			r.retryInterval = 10 * time.Millisecond
 			lastHasError = r.refreshCache(lastHasError)
-			r.retryInterval = JwtPubKeyRetryInterval
 		}
 	}
 }
--- a/pilot/pkg/model/jwks_resolver.go
+++ b/pilot/pkg/model/jwks_resolver.go
@@ -434,12 +434,14 @@ func (r *JwksResolver) refresh() bool {
 				r.keyEntries.Delete(k)
 				k.jwksURI = jwksURI
 			}
-
 			resp, err := r.getRemoteContentWithRetry(jwksURI, networkFetchRetryCountOnRefreshFlow)
 			if err != nil {
 				hasErrors = true
 				log.Errorf("Failed to refresh JWT public key from %q: %v", jwksURI, err)
 				atomic.AddUint64(&r.refreshJobFetchFailedCount, 1)
+				if oldPubKey == "" {
+					r.keyEntries.Delete(k)
+				}
 				return
 			}
 			newPubKey := string(resp)
@@ -545,4 +547,16 @@ func compareJWKSResponse(oldKeyString string, newKeyString string) (bool, error)
 	// If we aren't able to compare using keys, we should return true
 	// since we already checked exact equality of the responses
 	return true, nil
+}
+
+func (r *JwksResolver) CheckPubKeyExistInCache(issuer string, jwksUri string) bool {
+	key := jwtKey{issuer: issuer, jwksURI: jwksUri}
+	if val, found := r.keyEntries.Load(key); found {
+		e := val.(jwtPubKeyEntry)
+		r.keyEntries.Store(key, e)
+		if e.pubKey != "" {
+			return true
+		}
+	}
+	return false
 }
--- a/pilot/pkg/model/jwks_resolver.go
+++ b/pilot/pkg/model/jwks_resolver.go
@@ -549,8 +549,8 @@ func compareJWKSResponse(oldKeyString string, newKeyString string) (bool, error)
 	return true, nil
 }
 
-func (r *JwksResolver) CheckPubKeyExistInCache(issuer string, jwksUri string) bool {
-	key := jwtKey{issuer: issuer, jwksURI: jwksUri}
+func (r *JwksResolver) CheckPubKeyExistInCache(issuer string, jwksURI string) bool {
+	key := jwtKey{issuer: issuer, jwksURI: jwksURI}
 	if val, found := r.keyEntries.Load(key); found {
 		e := val.(jwtPubKeyEntry)
 		r.keyEntries.Store(key, e)
--- a/pilot/pkg/model/jwks_resolver.go
+++ b/pilot/pkg/model/jwks_resolver.go
@@ -148,7 +148,7 @@ type JwksResolver struct {
 	// How many times refresh job failed to fetch the public key from network, used in unit test.
 	refreshJobFetchFailedCount uint64
 
-	// Whenever we fail to fetch pubkey from jwksuri in main flow this variable becomes true
+	// Whenever istiod fails to fetch the pubkey from jwksuri in main flow this variable becomes true for background trigger
 	jwksUribackgroundChannel bool
 }
 
--- a/pilot/pkg/model/jwks_resolver.go
+++ b/pilot/pkg/model/jwks_resolver.go
@@ -19,6 +19,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -222,9 +223,12 @@ func newJwksResolverWithCABundlePaths(
 	return ret
 }
 
+var errEmptyPubKeyFoundInCache = errors.New("empty public key found in cache")
+
 // GetOrUpdatePublicKey returns the  JWT public key if it is available in the cache
-// or adds the jwksURI in the cache to fetch the public key in the background process
-func (r *JwksResolver) GetOrUpdatePublicKey(issuer string, jwksURI string) string {
+// or fetch with from jwksuri if there is a error while fetching then it adds the
+// jwksURI in the cache to fetch the public key in the background process
+func (r *JwksResolver) GetOrUpdatePublicKey(issuer string, jwksURI string) (string, error) {
 	now := time.Now()
 	key := jwtKey{issuer: issuer, jwksURI: jwksURI}
 	if val, found := r.keyEntries.Load(key); found {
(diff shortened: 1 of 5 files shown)


howardjohn:
How does this related to the other PR? Seems the same?

pilot/pkg/model/jwks_resolver.go
```diff
@@ -449,6 +448,7 @@ func (r *JwksResolver) refresh() bool {
 					atomic.AddUint64(&r.refreshJobFetchFailedCount, 1)
 					return
 				}
+				k.jwksURI = jwksURI
```
yangminzhu, line 451: you're not storing the key back to the cache.

aryan16:
> How does this related to the other PR? Seems the same?

There were some issues with other branch , so I created this new one. Will close my older PR.

pilot/pkg/model/jwks_resolver.go
```diff
@@ -384,41 +373,49 @@ func (r *JwksResolver) refresher() {
...
-			}
-			lastHasError = currentHasError
-			r.refreshTicker.Reset(r.refreshInterval)
+			lastHasError = r.refreshCache(lastHasError)
 		case <-closeChan:
 			r.refreshTicker.Stop()
 			return
+		case jwtKeyData := <-jwksuriChannel:
+			lastHasError = r.refreshCache(lastHasError)
+			if lastHasError {
```
yangminzhu, line 382: I think you do not need to log this anymore, it should already be logged inside the refresh().

howardjohn:
I see. It would be ideal to just force push to that one to keep history of the review. You can do `git checkout broken-branch; git reset --hard istiod-sync` and it will make the old branch the same as this one. Then `git push origin --force-with-lease istiod-sync`

pilot/pkg/model/jwks_resolver.go
aryan16: we are storing it here - https://github.com/istio/istio/blob/master/pilot/pkg/model/jwks_resolver.go#L462. Do you think there's a point to store it if we can't fetch the pubkey from that ?

aryan16:
> I see. It would be ideal to just force push to that one to keep history of the review. You can do `git checkout broken-branch; git reset --hard istiod-sync` and it will make the old branch the same as this one. Then `git push origin --force-with-lease istiod-sync`

Tried to follow this, but that branch has added all the previous commits which made it look really hard to analyze. I will add that PR here for reference.

pilot/pkg/security/authn/v1beta1/policy_applier_test.go
```diff
@@ -691,18 +691,18 @@ func TestJwtFilter(t *testing.T) {
...
-		model.JwtPubKeyRefreshIntervalOnFailure, model.JwtPubKeyRetryInterval)
-	defer push.JwtKeyResolver.Close()
-
-	push.ServiceIndex.HostnameAndNamespace[host.Name("jwt-token-issuer.mesh")] = map[string]*model.Service{}
-	push.ServiceIndex.HostnameAndNamespace[host.Name("jwt-token-issuer.mesh")]["mesh"] = &model.Service{
-		Hostname: "jwt-token-issuer.mesh.svc.cluster.local",
-	}
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
+			push := model.NewPushContext()
```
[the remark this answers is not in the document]
aryan16: Since we are fetching `pubkey` in the background, the results could be flaky if we create one single object and use it for every test case becuase we don't know when the key will be ready. It's a safe bet to use it inside the for loop and use the `fakePubKey` for every scenario.

pilot/pkg/model/jwks_resolver.go
yangminzhu: the problem is the old key with empty jwksURI is not removed, you're not updating it but just adding a new key, by update I mean you should replace the old one with the new content.

pilot/pkg/security/authn/v1beta1/policy_applier_test.go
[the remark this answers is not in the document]
yangminzhu: could you clarify why it is flaky to use one object for every test cases? Isn't that how it is working in real cases? The test should not deviate from the real behavior.

pilot/pkg/model/jwks_resolver.go
```diff
@@ -232,51 +233,32 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
 		e.lastUsedTime = now
 		r.keyEntries.Store(key, e)
 		if e.pubKey == "" {
-			return e.pubKey, errEmptyPubKeyFoundInCache
+			return e.pubKey
```
yangminzhu, line 236: there is already a return below, you can just remove the if-statement together.

pilot/pkg/security/authn/v1beta1/policy_applier_test.go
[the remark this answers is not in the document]
aryan16: all the tests cases separately call the `buildJwksUri()` which calls the `GetOrUpdatePubKey()`. So for the first case it will always return fakejwks, but the other cases with same `jwksuri` and `issuer` will have uncertainty as the method will return fakejwks when cache is empty but if cache got filled in any of the previous cases it will return the real pub key. So we don't exactly know when it will return real pubkey.

aryan16:
cc : @yangminzhu, @howardjohn

pilot/pkg/security/authn/v1beta1/policy_applier_test.go
```diff
@@ -705,6 +704,15 @@ func TestJwtFilter(t *testing.T) {
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
 			istiotest.SetBoolForTest(t, &features.EnableRemoteJwks, c.enableRemoteJwks)
+			NewPolicyApplier("root-namespace", c.in, nil, push).JwtFilter()
+			for _, config := range c.in {
+				for _, jwtRule := range config.Spec.(*v1beta1.RequestAuthentication).JwtRules {
+					retryToCheckCache(5, 20*time.Millisecond, func() bool {
+						return push.JwtKeyResolver.CheckPubKeyExistInCache(jwtRule.Issuer, jwtRule.JwksUri)
+					})
```
howardjohn, line 712:
```suggestion
			retry.UntilOrFail(t, func() bool {
						return push.JwtKeyResolver.CheckPubKeyExistInCache(jwtRule.Issuer, jwtRule.JwksUri)
					})
```

and can update other methods + remove the retryToCheckCache function

pilot/pkg/security/authn/v1beta1/policy_applier_test.go
```diff
@@ -705,6 +704,15 @@ func TestJwtFilter(t *testing.T) {
...
+				}
+			}
+
 			if got := NewPolicyApplier("root-namespace", c.in, nil, push).JwtFilter(); !reflect.DeepEqual(c.expected, got) {
```
howardjohn, line 716: should we use the `JwtFilter` from L707? Or if we are testing its not changed, then check both times?

pilot/pkg/security/authn/v1beta1/policy_applier_test.go
howardjohn: Actually its not clear what this function was doing originally... since we do not do anything if/when it times out?

pilot/pkg/security/authn/v1beta1/policy_applier_test.go
```diff
@@ -405,7 +405,7 @@ func TestJwtFilter(t *testing.T) {
 									JwksSourceSpecifier: &envoy_jwt.JwtProvider_LocalJwks{
 										LocalJwks: &core.DataSource{
 											Specifier: &core.DataSource_InlineString{
-												InlineString: test.JwtPubKey1,
+												InlineString: test.JwtPubKey2,
```
howardjohn, line 408: why is this changed?

pilot/pkg/model/jwks_resolver.go
```diff
@@ -220,63 +222,40 @@ func newJwksResolverWithCABundlePaths(
...
-		if err != nil {
-			log.Errorf("Failed to fetch public key from %q: %v", jwksURI, err)
-		}
-		pubKey = string(resp)
-	}
-
 	r.keyEntries.Store(key, jwtPubKeyEntry{
-		pubKey:            pubKey,
+		pubKey:            "",
 		lastRefreshedTime: now,
```
howardjohn, line 239: it hasn't actually be refreshed 'now', so this seems inaccurate?

pilot/pkg/model/jwks_resolver.go
```diff
@@ -449,13 +431,17 @@ func (r *JwksResolver) refresh() bool {
...
+				r.keyEntries.Delete(k)
+				k.jwksURI = jwksURI
 			}
-
 			resp, err := r.getRemoteContentWithRetry(jwksURI, networkFetchRetryCountOnRefreshFlow)
 			if err != nil {
 				hasErrors = true
 				log.Errorf("Failed to refresh JWT public key from %q: %v", jwksURI, err)
 				atomic.AddUint64(&r.refreshJobFetchFailedCount, 1)
+				if oldPubKey == "" {
```
howardjohn, line 442: when this happens `compareJWKSResponse` will spam a confusing error now

pilot/pkg/model/jwks_resolver.go
```diff
@@ -220,63 +222,40 @@ func newJwksResolverWithCABundlePaths(
...
-			// TODO(xulingqing): Find a better way to reject the request without using the fake jwks.
-			jwtPubKey = CreateFakeJwks(jwksURI)
-		}
+		jwtPubKey = r.GetOrUpdatePublicKey(jwtIssuer, jwksURI)
+	}
+	if jwtPubKey == "" {
+		log.Infof("The JWKS key is not yet fetched for issuer %s (%s), using a fake JWKS for now", jwtIssuer, jwksURI)
+		// This is a temporary workaround to reject a request with JWT token by using a fake jwks when istiod failed to fetch it.
+		// TODO(xulingqing): Find a better way to reject the request without using the fake jwks.
+		jwtPubKey = CreateFakeJwks(jwksURI)
```
howardjohn, line 258: Are we sure that this is the right user facing behavior?

This means that if I apply some PeerAuthentication, my traffic will immediately break for a bit before it fetches later?

pilot/pkg/model/jwks_resolver.go
```diff
@@ -93,6 +92,9 @@ var (
 
 	// JwtPubKeyRefreshInterval is the running interval of JWT pubKey refresh job.
 	JwtPubKeyRefreshInterval = features.PilotJwtPubKeyRefreshInterval
+
+	// channel for making jwksuri request aynsc
+	jwksuriChannel = make(chan jwtKey, 5)
```
howardjohn, line 97: Slightly worried about size of 5.. users probably don't have many more but if they do it can block everything

pilot/pkg/model/jwks_resolver.go
```diff
@@ -365,6 +372,7 @@ func (r *JwksResolver) getRemoteContentWithRetry(uri string, retry int) ([]byte,
 
 	for i := 0; i < retry; i++ {
 		body, err := getPublicKey()
+		fmt.Println(r.retryInterval)
```
yangminzhu, line 375: This should be deleted from the PR.

aryan16: oh yeah, I forgot to remove this. Will do it.

pilot/pkg/security/authn/v1beta1/policy_applier_test.go
```diff
@@ -486,14 +486,15 @@ func TestJwtFilter(t *testing.T) {
 			},
 		},
 		{
-			name: "JWT policy with bad Jwks URI",
```
aryan16, line 489: Changed the position of `bad jwks uri` test because it initiates the background fetch and since we are trying to fetch `pubkey` for all the jwksuri in the cache we would update the key for other jwksuri (as our mock server returns pubkey1 in first hit and pubkey2 for rest of the hits). If we don't do this, it would create some flakiness in the code.

pilot/pkg/model/jwks_resolver.go
```diff
@@ -259,20 +264,22 @@ func (r *JwksResolver) GetPublicKey(issuer string, jwksURI string) (string, erro
 		lastRefreshedTime: now,
 		lastUsedTime:      now,
 	})
-
+	if err != nil {
+		// fetching the public key in the background
+		jwksuriChannel <- key
```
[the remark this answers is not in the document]
aryan16: It can be done in concurrent execution, but in that case we need to change `r.refreshInterval` (which is 20 mins rn) to some smaller value to make sure that our ticker executes instantly. If we do that there might be other cases which could go wrong.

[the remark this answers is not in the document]
aryan16: what type of edge cases you're talking about ?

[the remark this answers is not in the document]
howardjohn: oops, misunderstood ticker. Replace what I said with `r.refreshTicker.C <- time.Now()`.

The issue would be if the refreshTicker select block is running and slow (it could be making 100s of http calls). We do the send on jwksuriChannel >5 times, it will block until the refreshTicker part is done.

Alternatively we can just have the on-demand async calls decoupled entirely from the refreshTicker (and probably just refresh that 1 url, not loop over all of them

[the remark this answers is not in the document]
howardjohn: Actually the single call may be better, otherwise for every new jwks we need to refreshall other jwks - may not scale well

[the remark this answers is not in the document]
yangminzhu: > otherwise for every new jwks we need to refreshall other jwks - may not scale well

yes, we should address this.

The other way could be to update the refresh() logic so that it only fetches the jwks URI that are newly added from here.

yangminzhu (commented):
Overall looks good, please also address the remaining open comments from john.

pilot/pkg/model/jwks_resolver.go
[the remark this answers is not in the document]
aryan16: I have discussed with him, since we don't have many `jwksuri` in the cache in one single time. I will add a new logic in my follow-up work where we will just try to fetch the pubkey only for the jwksuri which is failing.

yangminzhu:
@aryan16 there are still many open comments in the PR, I'm not sure if they are already resolved or not needed, could you update in each comment to indicate if it's resolved or not needed?

pilot/pkg/model/jwks_resolver.go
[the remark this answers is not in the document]
yangminzhu: Could you just do it in this PR? I'm a bit worried about the performance even if the customer is only using a small number of `jwksUri` because each of the `jwksUri` will trigger a new config push right?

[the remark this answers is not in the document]
aryan16: even after we refresh one single `jwksuri`, it is gonna create a new config push and will push the whole cache again. The only drawback which I see in our current solution is that it will try to refresh all the `jwksuri` in the cache. 
@howardjohn do we have any option to push one single `jwksuri` - `pubkey` to envoy ?

aryan16:
> @aryan16 there are still many open comments in the PR, I'm not sure if they are already resolved or not needed, could you update in each comment to indicate if it's resolved or not needed?

sure, I forgot to close few comments

aryan16:
@yangminzhu @howardjohn PTAL

pilot/pkg/model/jwks_resolver.go
```diff
@@ -384,41 +394,53 @@ func (r *JwksResolver) refresher() {
...
+			if !r.jwksUribackgroundChannel {
+				lastHasError = r.refreshCache(lastHasError)
 			}
-			lastHasError = currentHasError
-			r.refreshTicker.Reset(r.refreshInterval)
 		case <-closeChan:
 			r.refreshTicker.Stop()
 			return
+		case <-jwksuriChannel:
+			r.jwksUribackgroundChannel = true
```
howardjohn, line 404: style: maybe we can just make `jwksUribackgroundChannel` a parameter to `refreshCache`? Using the struct level variable makes it pretty hard to follow

howardjohn (approved):

pilot/pkg/model/jwks_resolver.go
aryan16: With the struct level variable we can stop the main ticker execution when our background execution is going on. This would resolve the consistency issues, added a check on line no. 397

PR #38315, opened 2022-04-10
[tf] Make ReachedClusters match current behavior

nmittler:
Tests that use `check.ReachedClusters` today are flaky for cross-network calls.

To illustrate why, assume we have the following clusters: c1 and c2 on network 1, and c3 on network2. Each cluster has an east-west gateway. When a request from c3 is made, it will be load-balanced correctly across the gateways in c1 and c2. However, since each of those gateways see workloads in both c1 and c2, there is a chance that requests across the gateways will end up at the same backend pod. If the client connection is reused between requests, this would effectively mean that the other cluster would never be reached. This is because cross-network requests effectively establish an end-to-end TCP connection from client->gateway->pod, using SNI at the gateway for routing.

HBONE will ultimate fix this problem, but until then, we need to fix our tests so that they reflect Istio's behavior today.

This PR loosens the requirements for `ReachedClusters` to ensure that all expected networks were reached, and that all clusters within the same network as the client were reached.


hzxuzhonghu (commented):
Would be better to check reached networks for this test

nmittler:
@hzxuzhonghu 

> Would be better to check reached networks for this test

I'm not sure I understand your suggestion. Today (prior to this change) our tests verify that all cluster are reached, however this is not always guaranteed when using multiple networks. All of our multicluster prow jobs also use multinetwork, which means we get flakes.

tests/integration/pilot/ingress_test.go
```diff
@@ -383,175 +384,191 @@ spec:
...
+								Protocol: protocol.HTTPS,
+							},
+							HTTP: echo.HTTP{
+								Path:    "/test/namedport",
+								Headers: headers.New().WithHost("bar.example.com").Build(),
+							},
+							TLS: echo.TLS{
+								CaCert: ingressutil.IngressCredentialB.CaCert,
+							},
+							Check: check.And(check.OK(), check.ReachedClusters(t, ingr.Cluster(), apps.B.Clusters())),
```
howardjohn, line 569: It seems like this will break the 'agnostic' tests?

Is there some way we can pipe through more info into the check interface, like the opts and source?

Note opts is not actually enough in this particular case, since we have no `To` - the `apps.B` could never be inferred.

Note sure all that will be worth  the complexity... cc @stevenctl

howardjohn (approved):

tests/integration/pilot/ingress_test.go
nmittler: I considered that and I'm not against it. I didn't make the change here for 2 reasons:

1. the complexity of the change required
2. doing it this way allows some minor optimization so that we don't have to create the network map on every call.

Definitely open to making the bigger change, if everyone agrees that it's the right approach. @stevenctl WDYT?

nmittler:
@howardjohn @stevenctl I've removed the hold since I'd like to get this in. Let's discuss offline whether I should do the refactoring that @howardjohn suggested.

tests/integration/pilot/ingress_test.go
stevenctl: TrafficTestCase should already support this, if it's possible to use that here:

https://github.com/istio/istio/blob/11a37c6b029fb2e775ecde8c2e7314c36832b201/tests/integration/pilot/common/traffic.go#L68

stevenctl (approved):

hzxuzhonghu (approved):

nmittler:
@howardjohn @stevenctl I took a crack at refactoring `Caller` and `Checker`. PTAL

tests/integration/pilot/ingress_test.go
```diff
@@ -383,175 +383,191 @@ spec:
...
-						Port: echo.Port{
-							Protocol: protocol.HTTP,
-						},
-						HTTP: echo.HTTP{
-							Path:    "/prefix/should/match",
-							Headers: headers.New().WithHost("server").Build(),
-						},
-						Check: successChecker,
-						Count: count,
+					call: func(ingr ingress.Instance) echo.CallOptions {
```
howardjohn, line 417: This doesn't need to be a function anymore I think?

howardjohn (approved):

stevenctl (approved):

kyessenov (approved):

PR #38332, opened 2022-04-11
Test envoy update

ericvn:

Test for pulling in latest Envoy. Comment of change needed here: https://github.com/istio/istio/pull/38310#issuecomment-1094511914

Run update-reps.sh script to get latest dependencies including Envoy.

### DIFF
--- a/pkg/bootstrap/config.go
+++ b/pkg/bootstrap/config.go
@@ -283,6 +283,7 @@ func extractRuntimeFlags(cfg *model.NodeMetaProxyConfig) map[string]string {
 		"envoy.reloadable_features.require_strict_1xx_and_204_response_headers":                                "false",
 		"re2.max_program_size.error_level":                                                                     "32768",
 		"envoy.reloadable_features.http_reject_path_with_fragment":                                             "false",
+		"envoy.reloadable_features.no_extension_lookup_by_name":                                                "true",
 	}
 	if !StripFragment {
 		// Note: the condition here is basically backwards. This was a mistake in the initial commit and cannot be reverted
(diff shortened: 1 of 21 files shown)


pkg/bootstrap/config.go
```diff
@@ -283,6 +283,7 @@ func extractRuntimeFlags(cfg *model.NodeMetaProxyConfig) map[string]string {
 		"envoy.reloadable_features.require_strict_1xx_and_204_response_headers":                                "false",
 		"re2.max_program_size.error_level":                                                                     "32768",
 		"envoy.reloadable_features.http_reject_path_with_fragment":                                             "false",
+		"envoy.reloadable_features.no_extension_lookup_by_name":                                                "true",
```
kyessenov, line 286: override to false as true is default.

kyessenov: for each reloadable feature, there is an expiration date, so there needs to be some follow up work once the flag is gone.

ericvn: Updated.

ericvn: I'm not sure what to do about the expiration date. Is that something we should do now, knowing that we understand what reloadable feature is the causing the failure? Then maybe this PR isn't really relevant at this point.

kyessenov: I don't want to hold update now, but we probably need an issue, and reference it here.

kyessenov: Looks like some unit tests need the same change applied because they use random bootstrap.

ericvn:
Updated the Goldens and making it further. Seem to have an issue with Envoy starting?
```
Didn't find a registered implementation for 'envoy.filters.http.router' with type URL: ''
2022-04-11T22:09:27.087095Z	info	Envoy 'envoy' (epoch 0) exited with error: exit status 1
2022-04-11T22:09:27.087357Z	info	Failed freeing BaseID for Envoy 'envoy' (epoch 0): error deleting Envoy base ID 63631 shared memory /dev/shm/envoy_shared_memory_63631: remove /dev/shm/envoy_shared_memory_63631: no such file or directory
    instance_test.go:328: exit status 1
```

ericvn:
I'm not sure of the best way to try and get logs when running the unit-tests to see what might be wrong.

ericvn:
Added in  changes from @howardjohn

howardjohn (approved):
LGTM, I have the long term fix in https://github.com/istio/istio/pull/38372

PR #38335, opened 2022-04-11
Avoid use of `interface{}` in controllers where possible

howardjohn:
Partial revert of https://github.com/istio/istio/pull/38232/. Keep it
flexible for the rare case of `interface{}` key, without impacting all
other users that are using it the typical way.

Long term we can use go generics


hzxuzhonghu (approved):
I am ok with this as Namespace/Name is the most frequently used key

ericvn (approved):
LGTM

morvencao (approved):

PR #38336, opened 2022-04-12
Do not fuzz unexported fields for CRDs

howardjohn:
Mutating protobuf un-exported fields results in invalid `state`, which
cannot happen in real world and is not expected usage of the protobuf
library.


ericvn (approved):
LGTM

Issue #38366, opened 2022-04-13
/etc/cacerts watcher does not work as it is mounted from k8s secret


hzxuzhonghu:
File mounted from k8s cm or secret is a sym link fsnotify can not handle this. 

https://stackoverflow.com/questions/54220143/how-to-get-notification-inside-an-application-in-kubernetes-when-configmaps-get/55029514#55029514

hzxuzhonghu:
The root cause it that we check file after a event occur				`valid := isValidCACertsFile(event.Name)`

PR #38367, opened 2022-04-13
Fix cacerts file watch

hzxuzhonghu:

fix https://github.com/istio/istio/issues/38366

### DIFF
--- a/pilot/pkg/bootstrap/istio_ca.go
+++ b/pilot/pkg/bootstrap/istio_ca.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pilot/pkg/bootstrap/istio_ca.go
```diff
@@ -338,8 +322,7 @@ func (s *Server) handleCACertsFileWatch() {
 			}
 
 			if event.Op&fsnotify.Write == fsnotify.Write || event.Op&fsnotify.Create == fsnotify.Create {
-				valid := isValidCACertsFile(event.Name)
```
hzxuzhonghu, line 341: another way is to watch files one by one, but we have 4 files to watch. So that's why we use dir watch here.

ramaraochavali: I did not understand. What would be the `event.name` here?

hzxuzhonghu: Take `/var/run/secrets/istio` directory as an examplem, it is mounted from istio rootcert configmap. 
When we update the configmap, the name can be `..data`, `..2022_04_13_09_13_52.574172247`

```
$ ls -al /var/run/secrets/istio
total 12
drwxrwxrwx 3 root root 4096 Apr 13 09:13 .
drwxr-xr-x 7 root root 4096 Apr 13 09:08 ..
drwxr-xr-x 2 root root 4096 Apr 13 09:13 ..2022_04_13_09_13_52.574172247
lrwxrwxrwx 1 root root   31 Apr 13 09:13 ..data -> ..2022_04_13_09_13_52.574172247
lrwxrwxrwx 1 root root   20 Apr 13 09:08 root-cert.pem -> ..data/root-cert.pem

```

hzxuzhonghu: No event with name `root-cert.pem` come

ramaraochavali: I see

ramaraochavali (approved):

PR #38372, opened 2022-04-13
Update proxy binary

howardjohn:
Account for https://github.com/envoyproxy/envoy/pull/20397

### DIFF
--- a/pkg/bootstrap/config.go
+++ b/pkg/bootstrap/config.go
@@ -283,6 +283,7 @@ func extractRuntimeFlags(cfg *model.NodeMetaProxyConfig) map[string]string {
 		"envoy.reloadable_features.require_strict_1xx_and_204_response_headers":                                "false",
 		"re2.max_program_size.error_level":                                                                     "32768",
 		"envoy.reloadable_features.http_reject_path_with_fragment":                                             "false",
+		"envoy.reloadable_features.no_extension_lookup_by_name":                                                "false",
 	}
 	if !StripFragment {
 		// Note: the condition here is basically backwards. This was a mistake in the initial commit and cannot be reverted
--- a/pkg/bootstrap/config.go
+++ b/pkg/bootstrap/config.go
@@ -283,7 +283,6 @@ func extractRuntimeFlags(cfg *model.NodeMetaProxyConfig) map[string]string {
 		"envoy.reloadable_features.require_strict_1xx_and_204_response_headers":                                "false",
 		"re2.max_program_size.error_level":                                                                     "32768",
 		"envoy.reloadable_features.http_reject_path_with_fragment":                                             "false",
-		"envoy.reloadable_features.no_extension_lookup_by_name":                                                "false",
 	}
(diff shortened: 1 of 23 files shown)


howardjohn:
alternative to https://github.com/istio/istio/pull/38332/files

pkg/bootstrap/config.go
```diff
@@ -283,7 +283,6 @@ func extractRuntimeFlags(cfg *model.NodeMetaProxyConfig) map[string]string {
 		"envoy.reloadable_features.require_strict_1xx_and_204_response_headers":                                "false",
 		"re2.max_program_size.error_level":                                                                     "32768",
 		"envoy.reloadable_features.http_reject_path_with_fragment":                                             "false",
-		"envoy.reloadable_features.no_extension_lookup_by_name":                                                "false",
```
kyessenov, line 286: This will break EnvoyFilters so probably need to stay on to give enough warning.

kyessenov (approved):

pkg/bootstrap/config.go
howardjohn: Good point. Do you know if we can somehow warn based on looking at the actual EnvoyFilter? Alternatively, is there logs in envoy is this flag is enabled but they use extension-by-name?

kyessenov: You can parse the final xDS and find all missing typed_configs. I don't think there is a warning in envoy, probably worth adding one.

howardjohn: But is there some reasonable way to know something should have a typed_config? I guess some protoreflect magic possibly could do it, any simpler ideas?

kyessenov: No, because protobufs don't have subtyping. Yes, this requires some clunky reflection. But also, noone should expect EnvoyFilters to work as-is so this might be solving the wrong problem.

hzxuzhonghu: Then when can enable this flag?

pkg/bootstrap/testdata/all_golden.json
```diff
@@ -628,15 +628,15 @@
...
+  
+  
   ,
   "cluster_manager": {
     "outlier_detection": {
       "event_log_path": "/dev/stdout"
     }
   }
-
+
```
ericvn, line 641: Nit: when reverting the golden files back, a number of the white space corrections get reverted as well. I know it's a little bit of a chore to remove.

howardjohn: I just ran REFRESH_GOLDEN.. we shouldn't have manual changes here?

howardjohn:
cc @ramaraochavali @hzxuzhonghu

pkg/bootstrap/config.go
howardjohn: I added a warning in validation

ericvn (approved):

hzxuzhonghu (approved):

PR #38373, opened 2022-04-13
Enable hostNetwork support in istiod Helm chart

paulwitt:
When running in EKS and using a 3rd-party CNI provider like Weave Net the control plane is unable to call services inside the cluster. This prevents webhooks from working. Running istiod on the host network fixes this issue.

- [ ] Docs
- [X] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
+++ b/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
@@ -75,6 +75,9 @@ spec:
 {{- end }}
       securityContext:
         fsGroup: 1337
+      {{- if .Values.pilot.enableHostNetwork }}
+      hostNetwork: true
+      {{- end }}
       containers:
         - name: discovery
 {{- if contains "/" .Values.pilot.image }}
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
I don't think this is secure, IMO, to run in host network. Adding it to the helm chart essentially means we are endorsing it.

If folks want to run insecure options to workaround broken CNIs/Kubernetes clusters, then they are free to do so (using things like https://docs.solo.io/gloo-edge/latest/installation/gateway/kubernetes/helm_advanced/, IstioOperator patching, etc), but I don't think we should officially endorse this.

caliko61:
What about exposing only webhook port on Istiod to hostPort, instead of enabling hostNetwork.
Would exposing only webhook port on Istiod to host network cause some security issues?
Or can we consider that there will be no difference exposing webhook port on network host and not exposing?

howardjohn:
Its not just security, also reliability(cannot run multiple on one node, etc). IMO its a really dangerous idea. If individuals want to do this because they have a CNI that doesn't support core Kubernetes features that is one thing, but I don't think we should officially support it in any way.

In this scenario, if the CNI cannot be changed, IMO it would be better to expose the webhook over a api-server accessible mechanism (eg LoadBalancer service, etc)

howardjohn:
Closing due to security concerns and a decision in the WG that we would promote these types of customizations through last mile mutations, not chart modifcations (docs pending, see https://docs.google.com/document/d/1wsa06GGiq1LEGwhkiPP0FKIZJqdAiue-VeBonWAzAyk/edit#bookmark=id.4ru6nai2al4s). Thanks!

PR #38388, opened 2022-04-14
only do proxy check if istiod is part of enabled components

ramaraochavali:
Helm reconciler while pruining operator spec, checks for proxy info (connected proxies) even if `pilotEnabled` is false. That should be checked only operator spec lists pilot.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


morvencao (approved):
thanks!

Issue #37415, opened 2022-02-17
Add istioctl analyze warning about fragile EnvoyFilter patch order


frankbu:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**

An `EnvoyFilter` with an insert position relative to another filter (e.g., `INSERT_BEFORE`) can be
fragile because the default order of filter evaluation is based on the creation time of the filters.
If the insertion point for a filter is not found, the filter will not be added to the chain and nothing will
appear in the error log to indicate that it was skipped.

Documentation describing this subtlety has been added in https://github.com/istio/istio.io/pull/10935.
It would be good to also make `istioctl analyze` produce a warning message for this kind of
fragile configuration.

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

frankbu:
Reopening since the fix was prematurely merged. See comment here: https://github.com/istio/istio/pull/38294#discussion_r847393883

linsun:
@frankbu @AndreaM12345 @ericvn Can this be closed now? This is mentioned as something fixed in the release note of 1.14.

cc @lei-tang @GregHanson

PR #38389, opened 2022-04-14
Additional changes for istioctl analyzer when checking envoyFilter patch operations

AndreaM12345:
additional fixes for 37415

Changes to improve the istioctl analyzer for the envoyFilter Patch operations.  Makes checks to see what the filters are applied to and provides warnings/errors against the ADD, MERGE, REMOVE, REPLACE operations

### DIFF
--- a/pkg/config/analysis/analyzers/envoyfilter/envoyfilter.go
+++ b/pkg/config/analysis/analyzers/envoyfilter/envoyfilter.go
@@ -45,10 +45,8 @@ func (*EnvoyPatchAnalyzer) Metadata() analysis.Metadata {
 
 // Analyze implements analysis.Analyzer
 func (s *EnvoyPatchAnalyzer) Analyze(c analysis.Context) {
-	patchFilterNames := make([]string, 1)
 	c.ForEach(collections.IstioNetworkingV1Alpha3Envoyfilters.Name(), func(r *resource.Instance) bool {
-		names := s.analyzeEnvoyFilterPatch(r, c, patchFilterNames)
-		patchFilterNames = names
+		s.analyzeEnvoyFilterPatch(r, c)
 		return true
 	})
 }
@@ -65,13 +63,9 @@ func relativeOperationMsg(r *resource.Instance, c analysis.Context, index int, p
 	}
 }
 
-func (*EnvoyPatchAnalyzer) analyzeEnvoyFilterPatch(r *resource.Instance, c analysis.Context, patchFilterNames []string) []string {
+func (*EnvoyPatchAnalyzer) analyzeEnvoyFilterPatch(r *resource.Instance, c analysis.Context) {
 	ef := r.Message.(*network.EnvoyFilter)
 	for index, patch := range ef.ConfigPatches {
-		// collect filter names to figure out if the order is correct
-		tmpValue := patch.Patch.Value.GetFields()
-		instanceName := tmpValue["name"]
-
 		// check each operation type
 		if patch.Patch.Operation == network.EnvoyFilter_Patch_ADD {
 			// the ADD operation is an absolute operation but provide a warning
@@ -114,38 +108,17 @@ func (*EnvoyPatchAnalyzer) analyzeEnvoyFilterPatch(r *resource.Instance, c analy
 
 				c.Report(collections.IstioNetworkingV1Alpha3Envoyfilters.Name(), message)
 			} else {
-				count := 0
-				for _, name := range patchFilterNames {
-					if instanceName.String() == name {
-						count++
-						break
-					}
-				}
-				if count > 0 {
-					// There is more than one envoy filter using the same name with the REPLACE operation
-					// Also a relative operation (REPLACE) was used so check if priority is set and if not set provide a warning
-					relativeOperationMsg(r, c, index, ef.Priority)
-				}
+				// A relative operation (REPLACE) was used so check if priority is set and if not set provide a warning
+				relativeOperationMsg(r, c, index, ef.Priority)
 			}
 		} else if patch.Patch.Operation == network.EnvoyFilter_Patch_INSERT_BEFORE || patch.Patch.Operation == network.EnvoyFilter_Patch_INSERT_AFTER {
 			// Also a relative operation (INSERT_BEFORE or INSERT_AFTER) was used so check if priority is set and if not set provide a warning
 			relativeOperationMsg(r, c, index, ef.Priority)
 		} else if patch.Patch.Operation == network.EnvoyFilter_Patch_MERGE {
-			count := 0
-			for _, name := range patchFilterNames {
-				if instanceName.String() == name {
-					count++
-					break
-				}
-			}
-			if count > 0 {
-				// there is more than one envoy filter using the same name with the REPLACE relative operation
-				// provide a warning if no priority was set
-				relativeOperationMsg(r, c, index, ef.Priority)
-			}
+			// A relative operation (MERGE) was used so check if priority is set and if not set provide a warning
+			relativeOperationMsg(r, c, index, ef.Priority)
 		}
 		// append the patchValueStr to the slice for next iteration
-		patchFilterNames = append(patchFilterNames, instanceName.String())
 	}
-	return patchFilterNames
+	return
 }
--- a/pkg/config/analysis/analyzers/envoyfilter/envoyfilter.go
+++ b/pkg/config/analysis/analyzers/envoyfilter/envoyfilter.go
@@ -96,7 +96,6 @@ func (*EnvoyPatchAnalyzer) analyzeEnvoyFilterPatch(r *resource.Instance, c analy
 
 				c.Report(collections.IstioNetworkingV1Alpha3Envoyfilters.Name(), message)
 			}
-
 		} else if patch.Patch.Operation == network.EnvoyFilter_Patch_REPLACE {
 			// the REPLACE operation is only valid for HTTP_FILTER and NETWORK_FILTER.
 			if patch.ApplyTo != network.EnvoyFilter_NETWORK_FILTER && patch.ApplyTo != network.EnvoyFilter_HTTP_FILTER {
@@ -160,12 +159,10 @@ func (*EnvoyPatchAnalyzer) analyzeEnvoyFilterPatch(r *resource.Instance, c analy
 
 					c.Report(collections.IstioNetworkingV1Alpha3Envoyfilters.Name(), message)
 				}
-
 			}
 		}
 		// append the patchValueStr to the slice for next iteration
 		patchFilterNames = append(patchFilterNames, instanceName.String())
-
 	}
 	return patchFilterNames
 }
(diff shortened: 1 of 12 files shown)


releasenotes/notes/37415.yaml
```diff
@@ -6,3 +6,4 @@ issue:
 releaseNotes:
 - |
   **Added** a new analyzer for envoy filter patch operations to address issue 37415
+  **Updated** a new analyzer for envoy filter patch operations to address issue 37415
```
ericvn, line 9: Can we update the lines to not say #37415 and have some descriptive text. There will be a link that points to the issue added in the releases notes.

GregHanson: And perhaps updating the original release notes `**Added**` item instead of added an additional `**Updated**` line.  Since the code changes will still be in the same release it reads a little funny that something was `Added` and `Updated` in the same note.  Also, not sure if the `gen-release-notes` handles multiple releaseNotes lines

pkg/config/analysis/analyzers/envoyfilter/envoyfilter.go
```diff
@@ -45,20 +45,88 @@ func (*EnvoyPatchAnalyzer) Metadata() analysis.Metadata {
...
+				}
+				if count > 0 && ef.Priority == 0 {
+					// there is more than one envoy filter using the same name with the REPLACE operation
+					// provide a warning indicating that no priority was set and a relative operation was used
+					message := msg.NewEnvoyFilterUsesRelativeOperation(r)
+
+					if line, ok := util.ErrorLine(r, fmt.Sprintf(util.EnvoyFilterConfigPath, index)); ok {
+						message.Line = line
+					}
+					c.Report(collections.IstioNetworkingV1Alpha3Envoyfilters.Name(), message)
```
frankbu, line 126: These same lines are repeated in 4 places. Would be good to move to a separate function.

pkg/config/analysis/analyzers/envoyfilter/envoyfilter.go
```diff
@@ -71,7 +139,30 @@ func (*EnvoyPatchAnalyzer) analyzeEnvoyFilterPatch(r *resource.Instance, c analy
...
+					// provide a warning indicating that no priority was set and a relative operation was used
+
+					message := msg.NewEnvoyFilterUsesRelativeOperation(r)
+
+					if line, ok := util.ErrorLine(r, fmt.Sprintf(util.EnvoyFilterConfigPath, index)); ok {
+						message.Line = line
+					}
+
+					c.Report(collections.IstioNetworkingV1Alpha3Envoyfilters.Name(), message)
+				}
```
frankbu, line 161: These same lines are repeated in 2 places, so also would be good to move to a function?

pkg/config/analysis/analyzers/envoyfilter/envoyfilter.go
```diff
@@ -45,20 +45,88 @@ func (*EnvoyPatchAnalyzer) Metadata() analysis.Metadata {
...
+					// provide a warning indicating that no priority was set and a relative operation was used
+					message := msg.NewEnvoyFilterUsesRelativeOperation(r)
+
+					if line, ok := util.ErrorLine(r, fmt.Sprintf(util.EnvoyFilterConfigPath, index)); ok {
+						message.Line = line
+					}
+					c.Report(collections.IstioNetworkingV1Alpha3Envoyfilters.Name(), message)
+				}
+			}
+		} else if ef.Priority == 0 {
```
frankbu, line 129: This gets a little convoluted here. How about instead of checking `ef.Priority == 0` here, move the nested operation == checks up and then make sure priority is checked everywhere it's needed below. Looks to me like it is already redundantly checked in all but one places where it's needed. In fact, if you move the lines I suggested into a separate function, you could do the priority check in there. That would simplify the code a bit and make it more maintainable.

GregHanson (approved):
LGTM, adding a `hold` to wait for any last comments from @frankbu

pkg/config/analysis/analyzers/envoyfilter/envoyfilter.go
```diff
@@ -45,33 +45,107 @@ func (*EnvoyPatchAnalyzer) Metadata() analysis.Metadata {
...
+				c.Report(collections.IstioNetworkingV1Alpha3Envoyfilters.Name(), message)
+			} else {
+				count := 0
+				for _, name := range patchFilterNames {
+					if instanceName.String() == name {
+						count++
+						break
+					}
+				}
+				if count > 0 {
```
frankbu, line 124: This count check doesn't seem quite right to me. Can you explain what exactly you're trying to do?

AndreaM12345: was just trying to indicate if there is a patch being applied to the same envoy filter being used then should provide the warning indicating if a relative operation is being used.  If there is no previous envoy filter then don't need point out a relative is being used because then it would not matter since it will just get applied.

frankbu: Hmm, the doc for REPLACE says: `If the named filter is not found, this operation has no effect.` which is the same thing that happens if the matched filter for INSERT_BEFORE/AFTER is not found, but in those cases you do give a warning. Seems you should also warn here as well. Also, if the named filter is there, but after this one (instead of before) that seems like an error. Not sure how much you really should be checking here, but I think it should at least be consistent.

AndreaM12345: concerned if we are always giving warnings it will get ignored so was trying to narrow what  cases to indicate when its important to note things are relative.  Reason I didn't mark INSERT_BEFORE/AFTER was because the doc didn't mention what happened if the named filter was not found.  The doc mentions that if no filter is provided then with INSERT_BEFORE/AFTER then it will go either first or last.  From the issue it sounded like the doc was unclear if the filter is mentioned but not yet applied.  So it seemed we needed a warning since the doc was unclear as to behavior.  Since the doc explicitly says for REPLACE that the operation is ignored then was thinking  the behavior was known so no warning was needed.

AndreaM12345: if you think narrowing the cases is too confusing can also remove the checks so its always indicating a warning

AndreaM12345: okay made another check for just the proxyVersion so that if that was the issue it would stand out more while leaving the other warning messages.  Theory being that maybe more warnings are better and if we provide a different warning for the proxyVersion it will stand out more

pkg/config/analysis/msg/messages.yaml
```diff
@@ -560,6 +560,29 @@ messages:
   - name: "EnvoyFilterUsesRelativeOperation"
     code: IST0151
     level: Warning
-    description: "This envoy filter does not have a priority and has a relative patch operation set which can cause the envoyFilter not to be applied. Using the INSERT_FIRST option or setting the priority may help in ensuring the envoyFilter is applied correctly"
-    template: "This envoy filter does not have a priority and has a relative patch operation set which can cause the envoyFilter not to be applied. Using the INSERT_FIRST option or setting the priority may help in ensuring the envoyFilter is applied correctly"
-   
\ No newline at end of file
+    description: "This envoy filter does not have a priority and has a relative patch operation set which can cause the envoyFilter not to be applied. Using the INSERT_FIRST or ADD option or setting the priority may help in ensuring the envoyFilter is applied correctly"
```
esnible, line 563: Capitalized EnvoyFilter.  It should look like usage in https://istio.io/latest/docs/reference/config/networking/envoy-filter/

pkg/config/analysis/msg/messages.yaml
```diff
@@ -560,6 +560,29 @@ messages:
...
+    template: "This envoy filter does not have a priority and has a relative patch operation set which can cause the envoyFilter not to be applied. Using the INSERT_FIRST of ADD option or setting the priority may help in ensuring the envoyFilter is applied correctly"
```
esnible, line 564: `description` and `template` values usually end with a period.

releasenotes/notes/37415.yaml
```diff
@@ -5,4 +5,4 @@ issue:
   - 37415
 releaseNotes:
 - |
-  **Added** a new analyzer for envoy filter patch operations to address issue 37415
+  **Added** a new analyzer for envoy filter patch operations to provide warnings when relative patch operations are used without a priority set which can cause envoyFilters not to be applied correctly
```
esnible, line 8: Release notes should end with a period.

pkg/config/analysis/analyzers/testdata/relative-envoy-filter-operation.yaml
```diff
@@ -126,4 +126,85 @@ spec:
...
+kind: EnvoyFilter
+metadata:
+  name: test-reviews-lua-3
+  namespace: bookinfo
+spec:
+  workloadSelector:
+    labels:
+      app: reviews-3
+  configPatches:
+    # The first patch adds the lua filter to the listener/http connection manager
```
esnible, line 140: Capitalize Lua and all trademarks.

esnible (changes requested):
(So far) I have looked at the grammar of the messages output, not the logic of the analyzer.

esnible:
I tried testing this interactively.  First I checked it out with `gh pr checkout 38389`.  Then `make istioctl` and `istioctl -n bookinfo analyze --use-kube=false /tmp/envoy-filter-add-operation.yaml`.

In addition to the two new errors I see 

```
Warning [IST0133] (EnvoyFilter bookinfo/test-auth-2 /tmp/envoy-filter-add-operation.yaml:27) Schema validation warning: Envoy filter: unknown field "typed_config" in envoy.config.route.v3.RouteConfiguration
Warning [IST0133] (EnvoyFilter bookinfo/test-auth-3 /tmp/envoy-filter-add-operation.yaml:52) Schema validation warning: Envoy filter: unknown field "typed_config" in envoy.config.route.v3.Route
```

A bigger concern is that the analyzer isn't safe.  This EnvoyFilter

```yaml
kind: EnvoyFilter
metadata:
  name: test-auth-evil
  namespace: bookinfo
spec:
  configPatches:
  - applyTo: HTTP_FILTER
    match:
      context: SIDECAR_INBOUND
    patch:
```

will `istioctl analyze` to dump core when it tries to `patch.Patch.Value.GetFields()`.  This can be a denial-of-service if istiod validation is enabled.

esnible (approved):

GregHanson:
I think the last comments have been addressed, removing hold label

nmittler:
Looks like envoy filter analyzer tests are failing now. Here's an example from my PR: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/38253/unit-tests_istio/1523777549782487040

This PR sounds like it might be related. Can you take a look?

nmittler:
Looks like @howardjohn already raised [an issue for this](#38806)

PR #38398, opened 2022-04-15
cache sds secret configs

ramaraochavali:
Sds Secret configs are typically same for all clusters and listeners like default, RootCA / Metadata certs etc.

- This PR caches those configs so that they are not rebuilt every time. Specially useful for all listeners and new clusters (for existing clusters xds cache will cache it)
- Sets initial fetchtimeout to 0s for file based sds secret config. When there are lot of clusters/listeners, we see fetch timeouts happening. I think there is no harm to disable it for file based certs 

warning	envoy config	gRPC config: initial fetch timed out for type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.Secret

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pilot/pkg/security/model/authentication.go
```diff
@@ -72,6 +73,9 @@ var SDSAdsConfig = &core.ConfigSource{
 	ResourceApiVersion: core.ApiVersion_V3,
 }
 
+// Holds frequently used secret config to avoid excessive memory allocations.
+var sdsSecretConfigCache = lru.New(100)
```
hzxuzhonghu, line 77: I donot think caching a simple struct which can be built directly, like a construction function. Could help much. 
Actually, we have so many such small points that can be optimized, but introducing cache anywhere is not good AFAIK. Most obviously, it will make the project difficult to read and maintain.

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
```diff
@@ -2061,7 +2061,8 @@ func TestBuildUpstreamClusterTLSContext(t *testing.T) {
 												},
 											},
 										},
-										ResourceApiVersion: core.ApiVersion_V3,
+										ResourceApiVersion:  core.ApiVersion_V3,
+										InitialFetchTimeout: durationpb.New(time.Second * 0),
```
howardjohn, line 2065: This means a bad DR/Gateway config can block a pod indefinitely?

pkg/bootstrap/testdata/tracing_tls_golden.json
```diff
@@ -10,7 +10,7 @@
       "layers": [
           {
             "name": "global config",
-            "static_layer": {"envoy.deprecated_features:envoy.config.listener.v3.Listener.hidden_envoy_deprecated_use_original_dst":"true","envoy.reloadable_features.no_extension_lookup_by_name":"false","envoy.reloadable_features.http_reject_path_with_fragment":"false","envoy.reloadable_features.require_strict_1xx_and_204_response_headers":"false","overload.global_downstream_max_connections":"2147483647","re2.max_program_size.error_level":"32768"}
+            "static_layer": {"envoy.deprecated_features:envoy.config.listener.v3.Listener.hidden_envoy_deprecated_use_original_dst":"true","envoy.reloadable_features.http_reject_path_with_fragment":"false","envoy.reloadable_features.no_extension_lookup_by_name":"false","envoy.reloadable_features.require_strict_1xx_and_204_response_headers":"false","overload.global_downstream_max_connections":"2147483647","re2.max_program_size.error_level":"32768"}
```
howardjohn, line 13: is map iteration order impacting this and we need to sort it?

pilot/pkg/security/model/authentication.go
ramaraochavali: The point here, it is built over and over again with the same values similar to "default" and "RootCA" for metadata certs  - The same logic that we applied for "default" and "RootCA" (we cache them) applies for metadata certs as well as it is being built for every cluster and every listener.  I tried to make this generic by introducing cache here - but if you have concerns, I can limit it to just metadata certs similar to "default" and "RootCA" - but IMO that will be even more specific.

pkg/bootstrap/testdata/tracing_tls_golden.json
ramaraochavali: No. It changed the initial fetch timeout for static clusters referring to files.

pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
ramaraochavali: https://github.com/istio/istio/blob/f149180608e8e91ca7e686a0f7ffc01142f2a03d/security/pkg/nodeagent/cache/secretcache.go#L399 - We error out if file does not exist. So it should cover this case?

ramaraochavali:
@howardjohn @hzxuzhonghu should we purse this further or any concerns?

hzxuzhonghu:
TBH, I donot think we should do this

PR #38399, opened 2022-04-15
use empty composite filter when the ECDS of wasm filter is not arrived

### LINKED ISSUE #38313: HTTP  500 errors occurred during the creation of the `wasmplugin`

The `wasmplugin` resource triggers both ECDS and LDS updates, and when envoy receives the  LDS first, a new listener will start serving requests even if no wasm filter is found. Then all http requests fail with 500 status until the wasm filter loaded through ECDS.

Maybe we can add `default_config` and `apply_default_config_without_warming` to ECDS like this:

```json
             {
              "name": "wasm-test",
              "config_discovery": {
               "config_source": {
                "ads": {},
                "initial_fetch_timeout": "0s",
                "resource_api_version": "V3"
               },
               "default_config": {
                "@type": "type.googleapis.com/envoy.extensions.filters.http.composite.v3.Composite"
               },
               "apply_default_config_without_warming": true,
               "type_urls": [
                "type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm",
                "type.googleapis.com/envoy.extensions.filters.http.composite.v3.Composite"
               ]
              }
             }
```
The `default_config` here is  a empty composite filter config used to pass the requests without any processing.
I tested this configuration and no http 500 errors occurred during the creation of the `wasmplugin` resource.

[ ] Docs
[ ] Installation
[ ] Networking
[x] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

johnlanni:
fix #38313

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
+++ b/pilot/pkg/networking/core/v1alpha3/extension/wasmplugin.go
@@ -16,6 +16,7 @@ package extension
 
 import (
 	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	composite_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/composite/v3"
 	extensionsv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/wasm/v3"
 	hcm_filter "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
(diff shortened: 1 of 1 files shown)


kyessenov (commented):
This is not right in general if Wasm is used for auth. Using empty filter will let through requests that are disallowed by Wasm auth filters.

PR #38416, opened 2022-04-17
add alternate hosts to ndsz output in proxy

ramaraochavali:
Currently we do not add alternate hosts generated for k8s in ndsz output. Sometimes it would be useful to have that for debugging. It increases the ndsz output so if there are concerns with it (I can introduce a flag to url and only add if asked for). It is only added for proxy currently - as it involves more complication to do that at Istiod, but if it is useful, I can try that in a separate PR

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pkg/istio-agent/agent.go
```diff
@@ -575,7 +577,26 @@ func (a *Agent) Check() (err error) {
 
 func (a *Agent) GetDNSTable() *dnsProto.NameTable {
 	if a.localDNSServer != nil {
-		return a.localDNSServer.NameTable()
+		nt := a.localDNSServer.NameTable()
+		nt = proto.Clone(nt).(*dnsProto.NameTable)
+		a.localDNSServer.BuildAlternateHosts(nt, func(althosts map[string]struct{}, ipv4 []net.IP, ipv6 []net.IP, _ []string) {
```
howardjohn, line 582: why can't we use `buildDNSAnswers` here? instead of recreating it?

ramaraochavali: because the code is tightly couple to lookup table and populates a and aaaa records - we do not need to have that information here though.

May be possible to refactor both to reuse this piece of code.
Is that what you are suggestinb?

hzxuzhonghu (commented):
What's the ndsz before and after? IIUC, not any difference

ramaraochavali:
earlier we do not add alt hosts generated after nametable is built. This PR adds them

PR #38419, opened 2022-04-17
XDS: remove unused DependentTypes

Chen-Xintong:

XDS: remove unused DependentTypes since DependentConfigs can replace its functionality.

### DIFF
--- a/pilot/pkg/xds/endpoint_builder.go
+++ b/pilot/pkg/xds/endpoint_builder.go
@@ -170,12 +170,6 @@ func (b EndpointBuilder) DependentConfigs() []model.ConfigKey {
 	return configs
 }
 
-var edsDependentTypes = []config.GroupVersionKind{gvk.PeerAuthentication}
-
-func (b EndpointBuilder) DependentTypes() []config.GroupVersionKind {
-	return edsDependentTypes
-}
-
 // TODO(lambdai): Receive port value(15009 by default), builder to cover wide cases.
 type EndpointTunnelApplier interface {
 	// Mutate LbEndpoint in place. Return non-nil on failure.
--- a/pilot/pkg/model/xds_cache.go
+++ b/pilot/pkg/model/xds_cache.go
@@ -25,7 +25,6 @@ import (
 	"google.golang.org/protobuf/testing/protocmp"
 
 	"istio.io/istio/pilot/pkg/features"
-	"istio.io/istio/pkg/config"
 	"istio.io/istio/pkg/util/sets"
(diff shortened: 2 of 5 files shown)


pilot/pkg/xds/endpoint_builder.go
```diff
@@ -170,12 +170,6 @@ func (b EndpointBuilder) DependentConfigs() []model.ConfigKey {
 	return configs
 }
 
-var edsDependentTypes = []config.GroupVersionKind{gvk.PeerAuthentication}
-
-func (b EndpointBuilder) DependentTypes() []config.GroupVersionKind {
-	return edsDependentTypes
```
howardjohn, line 176: Isn't this breaking this? Now PA will not clear the EDS cache. They are NOT part of DependentConfigs.

DependentTypes is like DependentConfigs but when we want to say "all configs of this type", which we need for EDS.

howardjohn (changes requested):

pilot/pkg/xds/endpoint_builder.go
Chen-Xintong: Yes exactly, EDS DependentTypes really shouldn't be removed, but my understanding is based on previous changes, DependentTypes in XDS cache other than EDS should be removed, because they currently provide empty interfaces, is it right?

howardjohn: What do you mean 'removed'? They need the function to implement the interface. They currently just `return nil`.

What problem are you trying to solve?

Chen-Xintong: What I have some doubts is why do we need a `DependentTypes` that returns `nil`, can we remove this interface as my understanding is that `DependentConfigs` can functionally cover this part? Is the DependentTypes indispensable?

howardjohn: First, why? What problems is this solving?

Second - we need it because `DependentConfigs` needs a concrete list of the exact configs, but we don't know the full list of PA required by EDS.

Chen-Xintong: > First, why? What problems is this solving?

The starting point for this PR was that I saw that `DependentTypes` had no actual return value and reference, So I wanted to remove unneeded code. But as you said, `DependentConfigs` requires a concrete list of the exact configs, so we shouldn't need to make such changes for now.

> but we don't know the full list of PA required by EDS.

And for this, I can do some investigation on PA for EDS and find out what can do to make it more clearer for us.

PR #38438, opened 2022-04-18
Fix consistent hash based on source IP for TCP proxy

howardjohn:
Fixes test flakes like
https://prow.istio.io/view/gs/istio-prow/logs/integ-ipv6_istio_postsubmit/1514637916112949248

We recently added TCP sourceIP consistent hash. Even more recently, this
test started to fail often. I believe this is due to a change to apply
YAML in parallel, exposing this bug.

The root cause is that we do not push LDS for DR changes, but the config
impacts LDS.


hzxuzhonghu (approved):

hzxuzhonghu:
This is a bug fix, should be backported

PR #38446, opened 2022-04-19
Validate all local IPs of service proxy in dual-stack clusters

zhlsunshine:

Fixed [issue#35310](https://github.com/istio/istio/issues/35310) after [PR#38218](https://github.com/istio/istio/pull/38218) has been merged.
1. Validate all local IPs in service proxy, especially in dual stack cluster
2. Check all local IPs in service proxy, but keep the original logic for enabling IPv6 iptables only if all local IPs are IPv6

### DIFF
--- a/tools/istio-iptables/pkg/cmd/root.go
+++ b/tools/istio-iptables/pkg/cmd/root.go
@@ -85,8 +85,8 @@ var rootCmd = &cobra.Command{
 				panic(err)
 			}
 
-			for _, hostIP := range hostIPs {
-				validator := validation.NewValidator(cfg, hostIP)
+			if len(hostIPs) > 0 {
+				validator := validation.NewValidator(cfg, hostIPs[0])
 				if err := validator.Run(); err != nil {
 					handleErrorWithCode(err, constants.ValidationErrorCode)
 				}
@@ -175,6 +175,7 @@ func constructConfig() *config.Config {
 	for _, podIP := range podIPs {
 		arrPodIPs = append(arrPodIPs, podIP.String())
 	}
+	// TODO(zhlsunshine): this assertion logic should be changed if Istio dual-stack is ready
 	if networkutil.AllIPv6(arrPodIPs) {
 		cfg.EnableInboundIPv6 = true
 	}
--- a/tools/istio-iptables/pkg/cmd/root.go
+++ b/tools/istio-iptables/pkg/cmd/root.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
+	networkutil "istio.io/istio/pilot/pkg/util/network"
 	"istio.io/istio/tools/istio-iptables/pkg/capture"
 	"istio.io/istio/tools/istio-iptables/pkg/config"
 	"istio.io/istio/tools/istio-iptables/pkg/constants"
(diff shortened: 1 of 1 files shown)


howardjohn (changes requested):

tools/istio-iptables/pkg/cmd/root.go
```diff
@@ -78,16 +79,19 @@ var rootCmd = &cobra.Command{
...
+			hostIPs, err := getLocalIP()
 			if err != nil {
 				// Assume it is not handled by istio-cni and won't reuse the ValidationErrorCode
 				panic(err)
 			}
-			validator := validation.NewValidator(cfg, hostIP)
 
-			if err := validator.Run(); err != nil {
-				handleErrorWithCode(err, constants.ValidationErrorCode)
+			for _, hostIP := range hostIPs {
```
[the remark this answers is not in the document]
zhlsunshine: okay, sure, let's just validate the first non-Loopback, non-LinkLocalUnicast and non-LinkLocalMulticast IP as before.

howardjohn:
Actually I am not sure why we ever need to validate more than once.
Iptables either applied or didn't...

zhlsunshine:
> 

From the code, I notice that this validation after iptable rules set up is optional, and it's **false** by default, so it may be no big impact unless user required. However, I think this mechanism can help to do other thing, such as check the iptable rules are correct or expected, etc.

zhlsunshine:
> Actually I am not sure why we ever need to validate more than once. Iptables either applied or didn't...

From the code, I notice that this validation after iptable rules set up is optional, and it's **false** by default, so it may be no big impact unless user required. However, I think this mechanism can help to do other thing, such as check the iptable rules are correct or expected, etc.

howardjohn:
It's not intended to check the iptables rules are correct or expected and
does a bad job at that goal. It's goal - and only thing it is useful for,
IMO - is to check the rules are there without requiring permissions

zhlsunshine:
> It's not intended to check the iptables rules are correct or expected and does a bad job at that goal. It's goal - and only thing it is useful for, IMO - is to check the rules are there without requiring permissions

Good point, such as if it's in cni mode which set as **true** and **false** by default. This should be validated before build iptable rules, I will try to handle it if possible.

zhlsunshine:
> It's not intended to check the iptables rules are correct or expected and does a bad job at that goal. It's goal - and only thing it is useful for, IMO - is to check the rules are there without requiring permissions

@howardjohn New submit, there should be only one of IPv4 and IPv6 should be validated if it contains both IPv4 and IPv6. BTW, I understand `to check the rules are there without requiring permissions` means that this validation is used to check the impact of iptable rules with no any permissions required. Am I right?

tools/istio-iptables/pkg/cmd/root.go
```diff
@@ -69,25 +70,41 @@ var rootCmd = &cobra.Command{
...
 				// Assume it is not handled by istio-cni and won't reuse the ValidationErrorCode
 				panic(err)
 			}
-			validator := validation.NewValidator(cfg, hostIP)
-
-			if err := validator.Run(); err != nil {
-				handleErrorWithCode(err, constants.ValidationErrorCode)
+			var checkIPv4, checkIPv6 bool
+			for _, hostIP := range hostIPs {
+				// it's unnecessary to check all IP, only one of IPv4 and IPv6 should be validated
```
howardjohn, line 89: I don't understand this logic. Seems `checkIPv4 && checkIPv6` can never be `true` since `IPv4 || checkIPv6` will always skip things. Why not just keep the old logic, which was simpler and seems to do the same thing?

zhlsunshine: This logic can make sure that IPv4 and IPv6 addresses can be validated in dual-stack cluster, and just validate one IPv4 and one IPv6, no more than that.  `checkIPv4 && checkIPv6` can be true in dual-stack cluster, and `checkIPv4 || checkIPv6` can skip unnecessary IP validation.

howardjohn:
There is no need to validate ipv6 and ipv4. The only reason for the check
is to check if ANY rules were applied due to some bugs where the rule
application can be skipped

zhlsunshine:
> There is no need to validate ipv6 and ipv4. The only reason for the check is to check if ANY rules were applied due to some bugs where the rule application can be skipped

Okay, got it. If so, I can keep the old logic for this part.

PR #38447, opened 2022-04-19
Change update_deps.sh to use docker (default) or crane

ericvn:

This change is to better support developers on MacOS on silicon devices.

On Apple devices, commands like `sed` and `grep` behave differently. Some of the Istio scripts, like update_deps.sh, use these commands and fail when run on the MacOS. One can install software to make the commands work as they do on Linux. Another option is to simply to do `make shell`  to get a Linux command line. From this Linux command line, the scripts can be run.

One issue right now is that the `crane` command segfaults when run in the `make shell` container. An issue with qemu? 

This PR reverts a change that solely uses `crane` in `update_deps.sh`, instead allowing the developer to export `ISTIO_DOCKER_BUILDER=crane` to use `crane`. I made `docker` the default to mimic the behavior in `docker-builder`.

### DIFF
--- a/bin/update_deps.sh
+++ b/bin/update_deps.sh
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -e
+set -ex
 
 UPDATE_BRANCH=${UPDATE_BRANCH:-"master"}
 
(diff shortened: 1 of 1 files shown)


ericvn:
Adding a DNM to allow @GregHanson to test this on his Apple Silicon computer.

howardjohn:
This feels a bit wrong to me... do we expect arm machines to be able to run the build container successfully? I cannot imagine it works (or works well, at least) in many other places.

At the very least we should continue to use the proper way (crane) instead of hacking around it just for ARM.

Maybe its time to get multiarch build-tools

ericvn:
With qemu, there is a possibility of running intel/amd containers on arm machines. This works for Docker Desktop and podman. While I agree this isn't ideal and we should have arm versions of the build-tools and build arm images, that will be a longer process.

ericvn:
I'd like to leave this one set a little longer.

ericvn:
Closing this PR for now. Will reopen if the users are still seeing issues with crane on MacOS Silicon devices.

Issue #37463, opened 2022-02-22
WASM: config image pull interval


daixiang0:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**

I test it at local, the time that Istio proxy pull newer image is different, sometimes spend half an hour, sometimes spend 10 mins, looks like a random behavior.

I do not find any doc for its certain value, I think it would be better to make it configurable.

If I miss anything important, please tell me.

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[x] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

zirain:
there're two solution:
1. add a env(like `WASM_INSECURE_REGISTRIES` do) via `MeshConfig.ProxyMetadata`
2. make this configurable in `WasmPlugin` 

@bianpengyuan WDYT?

bianpengyuan:
Currently the image pulling happen on every listener update. We will implement pull policy to if not present, which should suffice what most people want (i.e. only pull once for the whole pod lifecycle).

daixiang0:
@bianpengyuan can I check the timing of the listener update? I want a certain value for pull interval, it helps for end users. If only pull once, the remaining update of WASM image would lack. Now, the random interval is not acceptable for end users.

bianpengyuan:
I think you should always tag your image differently if there is any change. You can specify `sha256` (https://istio.io/latest/docs/reference/config/proxy_extensions/wasm-plugin/#WasmPlugin) + IfNotPresent pull policy (Note this is not yet implemented) if you really want to use the same image url for different images.

daixiang0:
The case is that users always want to use the latest version.

bianpengyuan:
IMO I don't think it is good approach to let proxy pull new images whenever it is available in an uncontrolled way. If you are absolutely sure this is what you want, I think we can add an env variable to pilot-agent which will could set expiry for cached Wasm modules.

zirain:
[...]

PR #38475, opened 2022-04-20
Implement WasmPlugin image pull policy

ingwonsong:
https://github.com/istio/istio/issues/35607

Implement [WasmPlugin image pull policy ](https://github.com/istio/api/blob/b6a03a9e477eb918b7e3105aaa00529183e4f82a/extensions/v1alpha1/wasm.proto#L241).

Similar with [ImagePullSecret of Wasm](https://github.com/istio/istio/pull/37161), ImagePullPolicy of Wasm is sent to istio-agent through the environment variables of ECDS. 
The policy is used in the internal Wasm binary cache of istio-agent to follow the [definition](https://github.com/istio/api/blob/b6a03a9e477eb918b7e3105aaa00529183e4f82a/extensions/v1alpha1/wasm.proto#L234-L240)

### DIFF
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/name"
 
+	extensions "istio.io/api/extensions/v1alpha1"
 	"istio.io/istio/pkg/util/sets"
 	"istio.io/pkg/log"
 )
@@ -52,7 +53,7 @@ const (
 
 // Cache models a Wasm module cache.
 type Cache interface {
-	Get(url, checksum string, timeout time.Duration, pullSecret []byte) (string, error)
+	Get(url, checksum string, timeout time.Duration, pullSecret []byte, pullPolicy extensions.PullPolicy) (string, error)
 	Cleanup()
 }
 
@@ -134,8 +135,22 @@ func urlAsResourceName(fullURLStr string) string {
 	return fullURLStr
 }
 
+func needToUseCache(pullPolicy extensions.PullPolicy, u *url.URL) bool {
+	if u.Scheme == "oci" {
+		switch pullPolicy {
+		case extensions.PullPolicy_Always:
+			return false
+		case extensions.PullPolicy_IfNotPresent:
+			return true
+		default:
+			return !strings.HasSuffix(u.Path, ":latest")
+		}
+	}
+	return true
+}
+
 // Get returns path the local Wasm module file.
-func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration, pullSecret []byte) (string, error) {
+func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration, pullSecret []byte, pullPolicy extensions.PullPolicy) (string, error) {
 	// Construct Wasm cache key with downloading URL and provided checksum of the module.
 	key := cacheKey{
 		downloadURL: downloadURL,
@@ -145,16 +160,18 @@ func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration
 		},
 	}
 
-	// First check if the cache entry is already downloaded.
-	if modulePath := c.getEntry(key); modulePath != "" {
-		return modulePath, nil
-	}
-
-	// If not, fetch images.
 	u, err := url.Parse(downloadURL)
 	if err != nil {
 		return "", fmt.Errorf("fail to parse Wasm module fetch url: %s", downloadURL)
 	}
+
+	// First check if the cache entry is already downloaded and policy does not require to pull always.
+	if needToUseCache(pullPolicy, u) {
+		if modulePath := c.getEntry(key); modulePath != "" {
+			return modulePath, nil
+		}
+	}
+	// If not, fetch images.
 
 	// Byte array of Wasm binary.
 	var b []byte
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -53,7 +53,7 @@ const (
 
 // Cache models a Wasm module cache.
 type Cache interface {
-	Get(url, checksum string, timeout time.Duration, pullSecret []byte, pullPolicy extensions.PullPolicy) (string, error)
+	Get(url, checksum, resourceName, resourceVersion string, timeout time.Duration, pullSecret []byte, pullPolicy extensions.PullPolicy) (string, error)
 	Cleanup()
 }
 
@@ -62,7 +62,7 @@ type LocalFileCache struct {
 	// Map from Wasm module checksum to cache entry.
 	modules map[moduleKey]*cacheEntry
 	// Map from tagged URL to checksum
-	checksums map[string]string
+	checksums map[string]*checksumEntry
 
 	// http fetcher fetches Wasm module with HTTP get.
 	httpFetcher *HTTPFetcher
@@ -84,6 +84,12 @@ type LocalFileCache struct {
 
 var _ Cache = &LocalFileCache{}
 
+type checksumEntry struct {
+	checksum string
+	// Keeps the resource version per each resource for dealing with multiple resources which pointing the same image.
+	resourceVersionByResource map[string]string
+}
+
 type moduleKey struct {
 	// Identifier for the module. It should be neutral for the checksum.
 	// e.g.) oci://docker.io/test@sha256:0123456789 is not allowed.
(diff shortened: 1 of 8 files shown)


pkg/wasm/cache.go
```diff
@@ -104,25 +106,41 @@ func NewLocalFileCache(dir string, purgeInterval, moduleExpiry time.Duration, in
...
 		return "", fmt.Errorf("fail to parse Wasm module fetch url: %s", downloadURL)
 	}
 
+	// First check if the cache entry is already downloaded and policy does not require to pull always.
+	if needToUseCache(pullPolicy, u) {
+		if modulePath := c.getEntry(key); modulePath != "" {
+			return modulePath, nil
+		}
+	}
+	// If not, fetch images.
```
howardjohn, line 142: I think Kubernetes logic is a bit more efficient here. Even on Always it will not always re-pull the image and will instead just do a check against the current tag and make sure the sha256 of it matches what we have locally. So in the case it hasn't changed, it will be pretty cheap - just pulling the config of the image, not the full image.

pkg/wasm/cache.go
```diff
@@ -104,25 +106,41 @@ func NewLocalFileCache(dir string, purgeInterval, moduleExpiry time.Duration, in
...
-		return modulePath, nil
-	}
-
-	// If not, fetch images.
 	u, err := url.Parse(downloadURL)
 	if err != nil {
 		return "", fmt.Errorf("fail to parse Wasm module fetch url: %s", downloadURL)
 	}
 
+	// First check if the cache entry is already downloaded and policy does not require to pull always.
```
howardjohn, line 136: What does it mean to 'Always' pull? In Kubernetes it is clear - when we start up a pod. What does it mean for WASM though? Each time the xDS changes?

pkg/wasm/cache.go
ingwonsong: Good point. will take the way.

pkg/wasm/cache.go
douglas-reid: The API docs say simply: `'Always' is the default, mirroring K8s behaviour.` This _could_ be interpreted as lifetime of proxy container. That seems like it would essentially be the same as `IfNotPresent` then, however. 

The comment on `Always` suggests: `We will always pull the latest version of an image when applying this plugin.` Which I think means with each xDS push.

Thoughts?

howardjohn: Pulling on every xDS push seems not great - we push a lot

ingwonsong: Yes. Current implementation updates for each time of ECDS change for 'Always'.

If there is any desirable interpretation, please let me know that.

douglas-reid: > Pulling on every xDS push seems not great - we push a lot

Use of `Always` considered harmful ??

ingwonsong: Even with `Always`, if we check sha256, the number of xDS would not be problematic (of course, when the users provide sha256).

Please note that current implementation re-pulls the oci image or binary (via http) for each xDS change when the user does not provide sha256.

ingwonsong: We can use sha256 specified in `WasmPlugin` resource without pulling the config of images.
In normal case, I think it would be enough. 
Do you think that it is needed to care about mis-configuration as well?

ingwonsong: I am sorry. I realized that I am confused a bit.

As I mentioned above, currently if a user does not provide sha256 in WasmPlugin, it works like `Always` (pulling for each xDS changes). 

At this point, do you think current `ImagePullPolicy` is necessary? 

Rather, I guess that it would be better to remove `ImagePullPolicy` or add new property like `ImageExpiration` (or `ImageMinLifetime` whatever).

WDYT?

ingwonsong: One idea:

The comment for `Always` is `We will always pull the latest version of an image when applying this plugin`, 
so how about defining `Always` as `each time when WasmPlugin resource is changed`?

In implementation-wise, I think we can use "Namespace+Name(of WasmPlugin)+ResourceVersion" as a part of key in the local image cache.

howardjohn: I like the idea. We should check if ResourceVersion changes when annotation changes, since one way users may try to trigger a new pull is changing an annotation? It might, I just forget the semantics of the field

ingwonsong: Confirmed. The change of annotations causes ResourceVersion change. 

BTW, just for checking, Istiod reacts the change of annotation as well?

zirain:
@ingwonsong will this PR take cover https://github.com/istio/istio/issues/37463?

ingwonsong:
@zirain No. I don't have plan to support the feature yet. Rather, as @howardjohn and I discussed above, I am working on implementing `Always` semantics based on: pulling is happen for each time of `WasmPlugin` resource change.

zirain:
> @zirain No. I don't have plan to support the feature yet. Rather, as @howardjohn and I discussed above, I am working on implementing `Always` semantics based on: pulling is happen for each time of `WasmPlugin` resource change.

I know `ImagePullPolicy` come from k8s/docker, I'm thinking about reuse it on http, maybe rename to `PullPolicy`?

ingwonsong:
> I know `ImagePullPolicy` come from k8s/docker, I'm thinking about reuse it on http, maybe rename to `PullPolicy`?

That makes sense. Do we need to change API first?

ingwonsong:
@zirain I proposed the change ImagePullPolicy -> PullPolicy in https://github.com/istio/api/pull/2336.

pkg/wasm/cache.go
ingwonsong: As discussed here, I have changed this part to pull the image for each time that `ResourceVersion` is changed.

pkg/wasm/cache.go
ingwonsong: This is resolved by https://github.com/istio/istio/pull/38630

douglas-reid (approved):
LGTM. @howardjohn @zirain WDYT about approving this so we can get it in 1.14 with an eye towards Beta promotion in 1.15?

pkg/wasm/cache.go
```diff
@@ -52,7 +53,7 @@ const (
 
 // Cache models a Wasm module cache.
 type Cache interface {
-	Get(url, checksum string, timeout time.Duration, pullSecret []byte) (string, error)
+	Get(url, checksum, resourceName, resourceVersion string, timeout time.Duration, pullSecret []byte, pullPolicy extensions.PullPolicy) (string, error)
```
howardjohn, line 56: 4 strings/7 parameters is a lot.. maybe we should extract to a struct?

pkg/wasm/cache.go
```diff
@@ -61,7 +62,7 @@ type LocalFileCache struct {
 	// Map from Wasm module checksum to cache entry.
 	modules map[moduleKey]*cacheEntry
 	// Map from tagged URL to checksum
-	checksums map[string]string
+	checksums map[string]*checksumEntry
```
howardjohn, line 65: nit: do we need it to be a pointer? Seems to mostly introduce possibility for errors; copying the checksumEntry struct should be low cost

pkg/wasm/cache.go
```diff
@@ -278,7 +317,7 @@ func (c *LocalFileCache) addEntry(key cacheKey, wasmModule []byte, f string) err
 	return nil
 }
 
-func (c *LocalFileCache) getEntry(key cacheKey) string {
+func (c *LocalFileCache) getEntry(key *cacheKey, dontcareResourceVersion bool) string {
```
howardjohn, line 320: Mutating `cacheKey` and also returning something seems unexpected, could not tell this was the case without reading the implementation. Maybe we can return a new cacheKey?

Also maybe can replace `dontcareResourceVersion` with PullPolicy so its more clear when calling than a bool. Or if not, maybe can rename to `ignoreResourceVersion bool` or `alwaysPull bool` or something

howardjohn (approved):
Approving since concerns are stylistic

lei-tang (approved):

pkg/wasm/cache.go
ingwonsong: You are right. 

I am thinking about some sort of refactoring around the cache and fetchers after implementing the feature and integration tests.

pkg/wasm/cache.go
ingwonsong: Done to remove mutating `cacheKey` in `getEntry` and change to `ignoreResourceVersion`.

pkg/wasm/cache.go
ingwonsong: Will consider later. Thanks!

PR #38480, opened 2022-04-20
[tf] Various improvements

nmittler:
A number of improvements around the testing framework to help with work on https://github.com/istio/istio/pull/37914.

### DIFF
--- a/pkg/test/framework/components/echo/workload.go
+++ b/pkg/test/framework/components/echo/workload.go
@@ -66,6 +66,15 @@ func (ws Workloads) Len() int {
 	return len(ws)
 }
 
+// Addresses returns the list of addresses for all workloads.
+func (ws Workloads) Addresses() []string {
+	out := make([]string, 0, len(ws))
+	for _, w := range ws {
+		out = append(out, w.Address())
+	}
+	return out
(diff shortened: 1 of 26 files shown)


pkg/test/framework/components/echo/workload.go
```diff
@@ -66,6 +66,26 @@ func (ws Workloads) Len() int {
...
+// Addresses returns the list of addresses for all workloads.
+func (ws Workloads) Addresses() []string {
+	out := make([]string, 0, len(ws))
+	for _, w := range ws {
+		out = append(out, w.Address())
+	}
+	return out
+}
+
+// FormattedAddresses is a utility function that returns a formatted version of the Addresses
```
howardjohn, line 78: nit: slightly skeptical this should be used rather than something like `{{.FormattedAddresses|toYaml|indent x}}` or something?

howardjohn: oh especially if its like `notValues: [{{ .ipA }}, {{ .ipB }}, {{ .ipC }}]`. That could definitely be improved

howardjohn (approved):

PR #38482, opened 2022-04-20
pilot: do not add ALPN filter to inbound filters

howardjohn:
This is fixing a small regression in
https://github.com/istio/istio/pull/37198, which has not been shipped in
any Istio versions yet. Basically that accidentally added the ALPN
filter to inbound. This technically does nothing, since it only impacts
TLS origination which we don't do for inbound, but still better to keep
it out to retain old state + smaller config.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/listener_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener_builder.go
@@ -568,6 +568,7 @@ func buildInboundCatchAllFilterChains(configgen *ConfigGeneratorImpl,
 				Port:     15006,
 				Protocol: protocol.HTTP,
 			},
+			class:    istionetworking.ListenerClassSidecarInbound,
 			protocol: istionetworking.ListenerProtocolAuto,
 		}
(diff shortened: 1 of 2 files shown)


pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -568,6 +568,7 @@ func buildInboundCatchAllFilterChains(configgen *ConfigGeneratorImpl,
 				Port:     15006,
 				Protocol: protocol.HTTP,
 			},
+			class:    istionetworking.ListenerClassSidecarInbound,
```
hzxuzhonghu, line 571: Good, it's worth a test.

howardjohn: I have another PR refactoring all the tests. Will merge that and update the tests

howardjohn: done

hzxuzhonghu (approved):

PR #38483, opened 2022-04-20
Reduce https/http calls for getting an OCI image manifest for WASM plugin

ingwonsong:
In current implementation, to get OCI image for WASM, there are at least five times http/https calls happen.
By this patch, it is reduced to 3 times if the registry supports "https" which is usually expected. 

In details, `remote.Get` is always called two times. 
* [first](https://github.com/istio/istio/blob/7c583c4600decefc6bf7f8216dcd65edb0ed095a/pkg/wasm/imagefetcher.go#L162)
* [second](https://github.com/google/go-containerregistry/blob/570ba6c88a5041afebd4599981d849af96f5dba9/pkg/v1/remote/image.go#L53).

(each `remote.Get` calls two times http/https calls: `ping` and real request.)

- [ *] Policies and Telemetry


zirain:
@ingwonsong can you fix lint error? run `make lint`

zirain (approved):
lgtm, thanks!

PR #38489, opened 2022-04-21
Fix multi cluster sync

hzxuzhonghu:

We mark istiod ready before serving connections with this check

https://github.com/istio/istio/blob/8a5f8e115b3dd6c78b83d60359a17fb63a86053b/pilot/pkg/bootstrap/server.go#L849-L860

But ServiceController().HasSynced() respect `SyncTimeout`, this is not right, we spent a lot of effort to stop serving before cache and xds synced(CommittedUpdates).  This conflicts with our effort.

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -15,6 +15,7 @@
 package controller
 
 import (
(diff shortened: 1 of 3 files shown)


hzxuzhonghu:
replaced by https://github.com/istio/istio/pull/38491

PR #38491, opened 2022-04-21
Fix cluster and kube controller sync

hzxuzhonghu:

Before, every cluster use the same timeout boolean value, this is not right, thinking that a cluster can be added after the the boolean has been set

This pr also split the cluster and clusterstore out from secret controller, for good readability.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/bootstrap/servicecontroller.go
+++ b/pilot/pkg/bootstrap/servicecontroller.go
@@ -17,6 +17,7 @@ package bootstrap
 import (
 	"fmt"
 
+	"istio.io/istio/pilot/pkg/features"
 	"istio.io/istio/pilot/pkg/serviceregistry/aggregate"
 	kubecontroller "istio.io/istio/pilot/pkg/serviceregistry/kube/controller"
 	"istio.io/istio/pilot/pkg/serviceregistry/provider"
@@ -75,6 +76,7 @@ func (s *Server) initKubeRegistry(args *PilotArgs) (err error) {
 	args.RegistryOptions.KubeOptions.MeshWatcher = s.environment.Watcher
 	args.RegistryOptions.KubeOptions.SystemNamespace = args.Namespace
 	args.RegistryOptions.KubeOptions.MeshServiceController = s.ServiceController()
+	args.RegistryOptions.KubeOptions.SyncTimeout = features.RemoteClusterTimeout
 
 	s.multiclusterController.AddHandler(kubecontroller.NewMulticluster(args.PodName,
 		s.kubeClient,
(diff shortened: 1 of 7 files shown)


hzxuzhonghu:
@stevenctl

stevenctl:
Haven't looked at code yet, but 

> a cluster can be added after the the boolean has been set

The original idea was to prevent istiod from starting if one of the remote clusters were down.

I'm sure we can improve it though.

hzxuzhonghu:
Yes, i know the original idea. This does not break that

pilot/pkg/bootstrap/servicecontroller.go
```diff
@@ -75,6 +76,7 @@ func (s *Server) initKubeRegistry(args *PilotArgs) (err error) {
 	args.RegistryOptions.KubeOptions.MeshWatcher = s.environment.Watcher
 	args.RegistryOptions.KubeOptions.SystemNamespace = args.Namespace
 	args.RegistryOptions.KubeOptions.MeshServiceController = s.ServiceController()
+	args.RegistryOptions.KubeOptions.SyncTimeout = features.RemoteClusterTimeout
```
stevenctl, line 79: I don't think we want the timeout behavior for the "local" cluster.

stevenctl (commented):
LGTM except the one other comment.

pilot/pkg/bootstrap/servicecontroller.go
hzxuzhonghu: Reasonable, done

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -155,8 +155,6 @@ func (m *Multicluster) ClusterAdded(cluster *multicluster.Cluster, clusterStopCh
 	// clusterStopCh is a channel that will be closed when this cluster removed.
 	options := m.opts
 	options.ClusterID = cluster.ID
-	// the aggregate registry's HasSynced will use the k8s controller's HasSynced, so we reference the same timeout
-	options.SyncTimeout = cluster.SyncTimeout
```
stevenctl, line 159: I think we need to add this back, or make it

```
if !localCluster {
  options.SyncTimeout = features.RemoteClusterTimeout
}
```

I think the test still passes because the `pkg/kube/multicluster` blocks has synced, but the k8s controller doesn't.

hzxuzhonghu: good catch, i removed the value setting, should add it here for non local cluster

stevenctl (approved):
Thanks!

PR #38494, opened 2022-04-21
[charts] Allow to configure a sleep command to let the main app stop properly

etifontaine:

This command add a short sleep by default to the istio-proxy when the pod shutdowns. 
It can be overwritten with the annotation `sidecar.istio.io/preStopSleep` on each pods.
This can be useful for apps that takes some time to shutdown and needs to handle last requests.

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
@@ -521,6 +521,12 @@ data:
               value: "{{ $value }}"
             {{- end }}
             {{with .Values.global.imagePullPolicy }}imagePullPolicy: "{{.}}"{{end}}
+            {{- if (isset .ObjectMeta.Annotations `sidecar.istio.io/preStopSleep`) }}
+            lifecycle:
+              preStop:
+                exec:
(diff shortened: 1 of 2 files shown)


GregHanson (commented):
Is this something we want default ON at first pass?  Or should the default be to opt-in?

Also, have you verified that `TerminationDrainDuration` in ProxyConfig does not solve this problem?
```code
	// The amount of time allowed for connections to complete on proxy shutdown.
	// On receiving `SIGTERM` or `SIGINT`, `istio-agent` tells the active Envoy to start draining,
	// preventing any new connections and allowing existing connections to complete. It then
	// sleeps for the `termination_drain_duration` and then kills any remaining active Envoy processes.
	// If not set, a default of `5s` will be applied.
	TerminationDrainDuration *duration.Duration 
```
in the code at `pkg/envoy/agent.go`:
```
		log.Infof("Graceful termination period is %v, starting...", a.terminationDrainDuration)
		time.Sleep(a.terminationDrainDuration)
		log.Infof("Graceful termination period complete, terminating remaining proxies.")
```

howardjohn (changes requested):
TerminationDrainDuration and EXIT_ON_ZERO_ACTIVE_CONNECTIONS is meant for this. I am against any `preStop` solution

PR #38495, opened 2022-04-21
Add tcp MX filter at gateway listeners

vikaschoudhary16:
TCP MX filter is missing to be added for the gateway listeneres. This seems unintentional because gateway listener builder logic calls the api to add tcp MX filter but within the api there is a check which excludes gateways. 

- [ ] Docs
- [ ] Installation
- [ X ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu (commented):
LGTM, especially for egress gateway

pilot/pkg/networking/core/v1alpha3/gateway_test.go
```diff
@@ -2263,3 +2263,115 @@ func TestBuildNameToServiceMapForHttpRoutes(t *testing.T) {
...
+			} else {
+				proxy.Metadata = &proxyGatewayMetadata
+			}
+
+			builder := cg.ConfigGen.buildGatewayListeners(&ListenerBuilder{node: proxy, push: cg.PushContext()})
+			for _, listener := range builder.gatewayListeners {
+				for _, fc := range listener.GetFilterChains() {
+					gotNWFilters, gotHTTPFilters := xdstest.ExtractFilterNames(t, fc)
+
+					sort.Strings(gotHTTPFilters)
```
howardjohn, line 2363: nit: order is actually important, so I would not sort them and make sure the test has the right order

pilot/pkg/networking/core/v1alpha3/gateway_test.go
```diff
@@ -2263,3 +2263,115 @@ func TestBuildNameToServiceMapForHttpRoutes(t *testing.T) {
...
+						Servers: []*networking.Server{
+							{
+								Port: &networking.Port{Name: "http", Number: 80, Protocol: "HTTP"},
+							},
+						},
+					},
+				},
+			},
+			nil,
+			[]string{"istio.metadata_exchange", "istio.alpn", "envoy.filters.http.fault", "envoy.filters.http.cors", "envoy.filters.http.router"},
```
howardjohn, line 2292: nit: use the constants from xdsfilter package

pilot/pkg/networking/core/v1alpha3/gateway_test.go
```diff
@@ -2263,3 +2263,115 @@ func TestBuildNameToServiceMapForHttpRoutes(t *testing.T) {
...
+		name                   string
+		node                   *pilot_model.Proxy
+		gateways               []config.Config
+		virtualServices        []config.Config
+		expectedHTTPFilters    []string
+		expectedNetworkFilters []string
+	}{
+		{
+			"http",
+			&pilot_model.Proxy{Type: pilot_model.Router},
```
howardjohn, line 2278: nit: it is always the same, can just remove it

pilot/pkg/networking/core/v1alpha3/gateway_test.go
```diff
@@ -2263,3 +2263,115 @@ func TestBuildNameToServiceMapForHttpRoutes(t *testing.T) {
...
+			[]config.Config{
```
howardjohn, line 2279: nit: use key names for each struct value. It is hard to tell what is what since we have multiple elements of the same type. Goland can auto add them

pilot/pkg/networking/core/v1alpha3/gateway_test.go
```diff
@@ -2263,3 +2263,115 @@ func TestBuildNameToServiceMapForHttpRoutes(t *testing.T) {
...
+	}
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			Configs := make([]config.Config, 0)
+			Configs = append(Configs, tt.gateways...)
+			Configs = append(Configs, tt.virtualServices...)
+			cg := NewConfigGenTest(t, TestOptions{
+				Configs: Configs,
+			})
+			cg.MemRegistry.WantGetProxyServiceInstances = tt.node.ServiceInstances
```
howardjohn, line 2350: Not needed?

pilot/pkg/networking/core/v1alpha3/gateway_test.go
```diff
@@ -2263,3 +2263,115 @@ func TestBuildNameToServiceMapForHttpRoutes(t *testing.T) {
...
+			proxy := cg.SetupProxy(&proxyGateway)
+			if tt.node.Metadata != nil {
```
howardjohn, line 2352: This never occurs

pilot/pkg/networking/core/v1alpha3/gateway_test.go
```diff
@@ -2263,3 +2263,115 @@ func TestBuildNameToServiceMapForHttpRoutes(t *testing.T) {
...
+			&pilot_model.Proxy{Type: pilot_model.Router},
+			[]config.Config{
+				{
+					Meta: config.Meta{Name: "passthrough-gateway", Namespace: "testns", GroupVersionKind: gvk.Gateway},
+					Spec: &networking.Gateway{
+						Servers: []*networking.Server{
+							{
+								Port:  &networking.Port{Name: "tcp", Number: 9443, Protocol: "TLS"},
+								Hosts: []string{"barone.example.com"},
+								Tls:   &networking.ServerTLSSettings{CredentialName: "test", Mode: networking.ServerTLSSettings_PASSTHROUGH},
```
howardjohn, line 2306: why passthrough? I thought we had decided to do it for istio_mutual only?

vikaschoudhary16: yeah, that we discussed for setting alpn in another PR. And same rationale applies here as well while adding MX filter and should be added for ISTIO_MUTUAL only.

That led me to think why we are adding httpMx filter ALWAYS, https://github.com/istio/istio/blob/a48b0e7ded79bb16aceaf78cdd9bdd70cdb51934/pilot/pkg/networking/core/v1alpha3/listener.go#L1446 instead of adding only for ISTIO_MUTUAL case?

howardjohn (commented):
Per https://github.com/istio/istio/pull/38499#pullrequestreview-953692154 I don't feel comfortable making any changes here with out integration tests.

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -744,17 +746,21 @@ func buildGatewayNetworkFiltersFromTCPRoutes(node *model.Proxy, push *model.Push
...
+			out = append(out, xdsfilters.TCPListenerMx)
+		}
+
 		// ensure we satisfy the rule's l4 match conditions, if any exist
 		// For the moment, there can be only one match that succeeds
 		// based on the match port/server port and the gateway name
 		for _, tcp := range vsvc.Tcp {
 			if l4MultiMatch(tcp.Match, server, gateway) {
-				return buildOutboundNetworkFilters(node, tcp.Route, push, port, v.Meta)
+				out = append(out, buildOutboundNetworkFilters(node, tcp.Route, push, port, v.Meta)...)
```
hzxuzhonghu, line 758: is this fixing a bug?

vikaschoudhary16:
> Per [#38499 (review)](https://github.com/istio/istio/pull/38499#pullrequestreview-953692154) I don't feel comfortable making any changes here with out integration tests.

https://github.com/istio/istio/pull/38499#discussion_r862584483
adding the same comment here so that we can move discussion to this PR:

true. At sidecar inbound listeners(DownstreamTlsContext),  for HTTP we configure `h2, http/1.1` and for TCP we have `istio-peer-exchange, h2, http/1.1`, which means `istio-peer-exchange` is sent to clients only in the TCP case.
With this PR, ONLY IF ISTIO_MUTUAL,  gateways are sending `"istio-peer-exchange", "h2", "http/1.1"` in both the TCP and HTTP case to the downstream clients. Therefore you are correct on divergence because of additional `istio-peer-exchange` in HTTP case.
Sending `istio-peer-exchange` in HTTP case is harmless though because we are not exposing istio native alpns to the mesh external clients. IMHO, it should be ok.
If still we want to make gateways not to send `istio-peer-exchange` in HTTP case, this will require some interface changes. [BuildListenerTlsContext](https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/v1alpha3/gateway.go#L660) is not passed any hint today if server is HTTP or TLS or terminating-tcp. We will have to add that hint in the function arguments. Initially, as I said, it seemed harmless to me, wanted to avoid this churn.

In any case, +1 for integration tests.

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -744,17 +746,21 @@ func buildGatewayNetworkFiltersFromTCPRoutes(node *model.Proxy, push *model.Push
 			continue
 		}
 
+		if server.GetTls().GetMode() == networking.ServerTLSSettings_ISTIO_MUTUAL {
+			out = append(out, xdsfilters.TCPListenerMx)
```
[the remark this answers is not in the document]
vikaschoudhary16: to be able to add MX filter at gateways only in ISTIO_MUTUAL case, we will have to pass tls mode in all the function calls i.e to `buildOutboundNetworkFilters` , `buildOutboundNetworkFiltersWithSingleDestination`, `buildOutboundNetworkFiltersWithSingleDestination` and `buildMetadataExchangeNetworkFilters`
And I guess that would require fixing unit tests as well for this additional argument. To avoid these changes, thought of doing it here itself.

pilot/pkg/networking/core/v1alpha3/gateway.go
vikaschoudhary16: MX filter should be the first, hence doing append here.

vikaschoudhary16:
@howardjohn @hzxuzhonghu  Please take a look. Also as we discussed offline, I need help in pushing als-sink image, without that tests will fail at ci. Locally I have verified tests to be working. Thanks again!

samples/als-sink/README.md
```diff
@@ -0,0 +1,5 @@
+# als-sink
```
hzxuzhonghu, line 1: Can you query metrics rather than using access log?  Which should be simpler.

vikaschoudhary16: https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-ExtensionProvider-EnvoyHttpGrpcV3LogProvider
```
 "filterStateObjectsToLog": [
            "wasm.upstream_peer",
            "wasm.upstream_peer_id",
            "wasm.downstream_peer",
            "wasm.downstream_peer_id"
```
This metadata about peer envoy, along with some more details, is exchanged by tcp and http metadata exchange filters running on client and server envoy. Using the above shared config, these metadata fields get reported by envoy to grpc als sink service. Can please point me to what metrics can I use to confirm this MX? I know only als service method.

tests/integration/telemetry/metadataexchange/metadata_exchange_test.go
```diff
@@ -0,0 +1,407 @@
...
+
+				return nil
+			}, retry.Delay(framework.TelemetryRetryDelay), retry.Timeout(framework.TelemetryRetryTimeout))
+			if err != nil {
+				t.Fatalf("test failed: %v", err)
+			}
+		})
+}
+
+func TestTCPMetadataExchangeBetweenSidecarAndGateway(t *testing.T) {
```
vikaschoudhary16, line 105: If I run this test with latest `master` istiod, client gets stuck. Following I see in sidecar envoy logs:
```
2022-05-14T07:51:08.678952Z	debug	envoy filter	[C2410] Creating connection to cluster outbound|443||istio-egressgateway.istio-system.svc.cluster.local
2022-05-14T07:51:08.678972Z	debug	envoy misc	Allocating TCP conn pool
2022-05-14T07:51:08.678987Z	debug	envoy pool	trying to create new connection
2022-05-14T07:51:08.678992Z	debug	envoy pool	creating a new connection (connecting=0)
2022-05-14T07:51:08.679027Z	debug	envoy connection	[C2411] connecting to 10.244.0.149:8443
2022-05-14T07:51:08.679112Z	debug	envoy connection	[C2411] connection in progress
2022-05-14T07:51:08.679124Z	debug	envoy conn_handler	[C2410] new connection from 10.244.0.152:46940
2022-05-14T07:51:08.679135Z	debug	envoy connection	[C2411] connected
2022-05-14T07:51:08.679925Z	debug	envoy pool	[C2411] attaching to next stream
2022-05-14T07:51:08.679932Z	debug	envoy pool	[C2411] creating stream
2022-05-14T07:51:08.679942Z	debug	envoy filter	[C2410] TCP:onUpstreamEvent(), requestedServerName: edition.cnn.com
2022-05-14T07:51:08.703348Z	debug	envoy main	flushing stats
2022-05-14T07:51:08.714870Z	debug	envoy filter	Alpn Protocol matched. Waiting to read more initial header.
```
looks like client envoy keeps waiting and connection does not proceed because at gateway tcp-mx filter is not there to send the magic number in the application data

Then I reran the test with istiod(`c9c09d8620`), which is before adding `istio-peer-exchange` alpn at gateway, this test and http test sidecar to gateway works fine.

samples/als-sink/README.md
zirain: IIRC, there's no such metrics, maybe @douglas-reid know more.

hzxuzhonghu: For metadataexchange, it sends node meta to upstream, So i mean you can check the source_workload in upstream

pilot/pkg/networking/core/v1alpha3/gateway.go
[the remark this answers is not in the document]
hzxuzhonghu: My concern is this may cause diverge in the future. 
We should also respect the PILOT_ENABLE_METADATA_EXCHANGE env

How about 

```
if server.GetTls().GetMode() == networking.ServerTLSSettings_ISTIO_MUTUAL {
            buildMetadataExchangeNetworkFilters )() // 
}
```

pilot/pkg/networking/core/v1alpha3/gateway.go
howardjohn, line 758: What is `vsvc.Tcp` has multiple elements? or multiple virtualServices.. won't we get duplicates?

pilot/pkg/networking/core/v1alpha3/gateway_test.go
```diff
@@ -2284,3 +2286,194 @@ func TestBuildNameToServiceMapForHttpRoutes(t *testing.T) {
...
+					if !reflect.DeepEqual(gotHTTPFilters, tt.expectedHTTPFilters) {
+						t.Fatalf("Expected HTTPfilters: %v, got: %v", tt.expectedHTTPFilters, gotHTTPFilters)
+					}
+					if !reflect.DeepEqual(gotNWFilters, tt.expectedNetworkFilters) {
+						t.Fatalf("Expected network filters: %v, got: %v", tt.expectedNetworkFilters, gotNWFilters)
+					}
+				}
+			}
+		})
+	}
```
howardjohn, line 2478:
```suggestion
			listenertest.VerifyListeners(t, builder.gatewayListeners, listenertest.ListenersTest{
				Listener: listenertest.ListenerTest{FilterChains: []listenertest.FilterChainTest{
					{
						NetworkFilters: tt.expectedNetworkFilters,
						HTTPFilters:    tt.expectedHTTPFilters,
					},
				}}})
```

nit: could use the new matching framework for listeners. Doesn't save much code here but its nice to get more usage. Also the current test would pass if no listeners were returned.

samples/als-sink/README.md
howardjohn: Big +1 to @hzxuzhonghu 's comment. We should be testing end user behavior, not low level implementation details, in our e2e tests

Adding ALS tests is useful for other reasons, but should not be used to test this change IMO.

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -746,7 +746,12 @@ func buildGatewayNetworkFiltersFromTCPRoutes(node *model.Proxy, push *model.Push
 		// based on the match port/server port and the gateway name
 		for _, tcp := range vsvc.Tcp {
 			if l4MultiMatch(tcp.Match, server, gateway) {
-				return buildOutboundNetworkFilters(node, tcp.Route, push, port, v.Meta)
+				out := make([]*listener.Filter, 0)
+				if server.GetTls().GetMode() == networking.ServerTLSSettings_ISTIO_MUTUAL {
+					out = append(out,
+						buildMetadataExchangeNetworkFilters(istionetworking.ListenerClassTCPIstioMTLSGateway)...)
```
hzxuzhonghu, line 752: As tls mode is checked out of the function, does the new type needed? 

Asking this because the type can be passed to many functions.

hzxuzhonghu (approved):

tests/integration/telemetry/stats/prometheus/stats.go
```diff
@@ -207,6 +211,45 @@ func TestStatsTCPFilter(t *testing.T, feature features.Feature) {
...
+	framework.NewTest(t).
+		Features(feature).
+		Run(func(t framework.TestContext) {
+			g, _ := errgroup.WithContext(context.Background())
+			for _, cltInstance := range client {
+				cltInstance := cltInstance
+				g.Go(func() error {
+					err := retry.UntilSuccess(func() error {
+						t.Logf("sending tcp traffic to gateway from sidecar")
+						requestURL := "curl --insecure -s -o /dev/null -w '%{http_code}' https://edition.cnn.com/politics"
```
howardjohn, line 224: Once https://github.com/istio/istio/pull/39051 lands there is a much better way to call this, I will help set it up once that merges

zirain (approved):

tests/integration/telemetry/stats/prometheus/stats.go
howardjohn: Figured would be easier to show vs tell: https://github.com/howardjohn/istio/tree/fetch-38495

Can you apply those changes/similar?

vikaschoudhary16: really appreciate John!!!
updated.

pilot/pkg/networking/core/v1alpha3/gateway.go
howardjohn: Sorry for the late reply. I am very very strongly against a new ListenerClass. Adding new enum types is one of the biggest source of bugs in Istio historically.

If you don't want to change all callsites, just make a new function `buildMetadataExchangeNetworkFiltersForMTLSGateway` or something

howardjohn (commented):
LGTM with one small note. Sorry I didn't detect it earlier

howardjohn (approved):

PR #38499, opened 2022-04-21
For TCP listeners at gateways, configure ALPNDownstreamWithMxc (istio-peer-exchange)

vikaschoudhary16:
TCP MX filter looks for [`istio-peer-exchange` alpn](https://github.com/istio/istio/blob/master/pilot/pkg/xds/filters/filters.go#L153) before exchanging metadata. At gateways, for TCP transport protocol as well, `"h2", "http/1.1"` is being configured as alpn, which means tcp metadata exchange does not work between any sidecar and the (tcp)gateway.
This PR is configuring `istio-peer-exchange` as alpn for tcp case at gateways.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/listener.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener.go
@@ -132,7 +132,8 @@ func (configgen *ConfigGeneratorImpl) BuildListenerTLSContext(serverTLSSettings 
 	alpnByTransport := util.ALPNHttp
 	if transportProtocol == istionetworking.TransportProtocolQUIC {
 		alpnByTransport = util.ALPNHttp3OverQUIC
-	} else if transportProtocol == istionetworking.TransportProtocolTCP {
+	} else if transportProtocol == istionetworking.TransportProtocolTCP &&
+		serverTLSSettings.Mode == networking.ServerTLSSettings_ISTIO_MUTUAL {
 		alpnByTransport = util.ALPNDownstreamWithMxc
 	}
 
--- a/pilot/pkg/networking/core/v1alpha3/listener.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener.go
@@ -132,7 +132,10 @@ func (configgen *ConfigGeneratorImpl) BuildListenerTLSContext(serverTLSSettings 
 	alpnByTransport := util.ALPNHttp
 	if transportProtocol == istionetworking.TransportProtocolQUIC {
 		alpnByTransport = util.ALPNHttp3OverQUIC
+	} else if transportProtocol == istionetworking.TransportProtocolTCP {
+		alpnByTransport = util.ALPNDownstreamWithMxc
 	}
+
 	ctx := &auth.DownstreamTlsContext{
 		CommonTlsContext: &auth.CommonTlsContext{
 			AlpnProtocols: alpnByTransport,
--- a/pilot/pkg/networking/core/v1alpha3/gateway_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/gateway_test.go
@@ -62,7 +62,7 @@ func TestBuildGatewayListenerTlsContext(t *testing.T) {
 			},
 			result: &auth.DownstreamTlsContext{
 				CommonTlsContext: &auth.CommonTlsContext{
-					AlpnProtocols: util.ALPNHttp,
+					AlpnProtocols: util.ALPNDownstreamWithMxc,
 					TlsCertificateSdsSecretConfigs: []*auth.SdsSecretConfig{
 						{
 							Name: "default",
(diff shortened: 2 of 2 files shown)


howardjohn (commented):
Should it only be ISTIO_MUTUAL?

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -132,7 +132,10 @@ func (configgen *ConfigGeneratorImpl) BuildListenerTLSContext(serverTLSSettings
 	alpnByTransport := util.ALPNHttp
 	if transportProtocol == istionetworking.TransportProtocolQUIC {
 		alpnByTransport = util.ALPNHttp3OverQUIC
+	} else if transportProtocol == istionetworking.TransportProtocolTCP {
+		alpnByTransport = util.ALPNDownstreamWithMxc
```
hzxuzhonghu, line 136: What about HTTP

hzxuzhonghu: Does it use pure http/1.1 for metadata exchange?  @howardjohn 

Looks we set "istio"

vikaschoudhary16: for http metadata exchange, alpn is not checked. alpn is checked only in tcp MX case.

hzxuzhonghu (commented):
LGTM, can you add a test

hzxuzhonghu (approved):
/ok-to-test

pilot/pkg/networking/core/v1alpha3/listener.go
howardjohn: Sorry I do not understand. For sidecar inbound listeners, we do NOT set this for HTTP types. This seems like a divergence?

hzxuzhonghu:
Ithink we should remove `transportProtocol == istionetworking.TransportProtocolTCP`
Because in cluster upstream tls context, the alpnProtocols is always

```
                            "alpnProtocols": [
                                "istio-peer-exchange",
                                "istio"
                            ]

```

hzxuzhonghu:
Not sure how they communicate with mismatched alpn, searched an issue https://github.com/envoyproxy/envoy/issues/11909. A little confused though

hzxuzhonghu:
Found this `https://github.com/istio/istio/blob/master/pilot/pkg/security/authn/utils/utils.go#L65-L69`, it was intended not to set for HTTP protocol

pilot/pkg/networking/core/v1alpha3/listener.go
vikaschoudhary16: true. At sidecar inbound listeners(DownstreamTlsContext),  for HTTP we configure `h2, http/1.1` and for TCP we have `istio-peer-exchange, h2, http/1.1`, which means `istio-peer-exchange` is sent to clients only in the TCP case.
With this PR, ONLY IF ISTIO_MUTUAL,  gateways are sending `"istio-peer-exchange", "h2", "http/1.1"` in both the TCP and HTTP case to the downstream clients. Therefore you are correct on divergence because of additional `istio-peer-exchange` in HTTP case.
Sending `istio-peer-exchange` in HTTP case is harmless though because we are not exposing istio native alpns to the mesh external clients. IMHO, it should be ok.
If still we want to make gateways not to send `istio-peer-exchange` in HTTP case, this will require some interface changes. [BuildListenerTlsContext](https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/v1alpha3/gateway.go#L660) is not passed any hint today if server is HTTP or TLS or terminating-tcp. We will have to add that hint in the function arguments. Initially, as I said, it seemed harmless to me, wanted to avoid this churn.

vikaschoudhary16:
> Not sure how they communicate with mismatched alpn, searched an issue [envoyproxy/envoy#11909](https://github.com/envoyproxy/envoy/issues/11909). A little confused though

tls handshake would get completed. In istio case, if unmatrched/unresolved then it will break tcp metadata excahnge filter because there filter verifies if resolved alpn, whcih is saved in the connecttion context is same as the protocol configured in filter configuration, which is `istio-peer-exchange`,  or not.

vikaschoudhary16:
> Maybe we should remove `transportProtocol == istionetworking.TransportProtocolTCP` Because in cluster upstream tls context, the alpnProtocols is always
> 
> ```
>                             "alpnProtocols": [
>                                 "istio-peer-exchange",
>                                 "istio"
>                             ]
> ```

yeah, this check is of no use

pilot/pkg/networking/core/v1alpha3/listener.go
howardjohn: Due to the `istio.alpn` filter, the ALPN is actually overridden for HTTP. Its not `istio-peer-exchange`, and if we do send that it will break traffic on the receiving side.

vikaschoudhary16: As discussed offline, true that due to alpn override, `istio-peer-exchange` will not be sent in the tls client hello by http clients, but this will not break the traffic. gateway(server) will send the negotiated alpn(one of the http1.1 or h2) in the server hello.
I tried locally the above explained behavior. Following are the wireshark captures of tls handshake:
**tls client hello from the http client:**
image

**tls server hello from the gateway(istio-mtls server):**
image

Note that listener on server side had istio-peer-exchange in the downstreamTlsContext during this experiment:
image

If HCM is not there on the client side i.e no alpn override, in the that from the envoy cluster alpn configuration, `istio-peer-exchange` will be sent by client in tls client hello message. gateway server will respond to that with `istio-peer-exchange` and tcp metadata will get exchanged because of the `istio-peer-exchange` being the negotiated alpn.

Issue #38503, opened 2022-04-21
Dangerous global variable usage in istioctl


howardjohn:
Istioctl is full of global variables for things like kubeconfig, etc. This has caused a variety of issues in the past like:
* multiple calls to istioctl re-using the wrong variables between calls
* Parallel calls causing race conditions
* **connecting to users real clusters and performing mutation actions against their real cluster** in unit tests

As well as just generally being a bad practice.

We should refactor this a bit to follow best practices and avoid global mutable state.

hanxiaop:
I think this can be closed. Most of the refactor work in the istioctl package is done. There may still be some work to do, but it won't relate to this issue.

PR #38504, opened 2022-04-21
Improve istioctl tests

howardjohn:
TestWait waited the full 2s, no need to delay tests that much.

TestVersion is entirely unsafe, see
https://github.com/istio/istio/issues/38503. It contacts real live cluster. If someone wants to replace it -- awesome -- but until then its not an appropriate unit test

The log is broken, does match arguments and replacement


istioctl/cmd/workload.go
```diff
@@ -578,8 +578,7 @@ func createHosts(kubeClient kube.ExtendedClient, ingressIP, dir string, revision
 	if net.ParseIP(ingressIP) != nil {
 		hosts = fmt.Sprintf("%s %s\n", ingressIP, istiodHost(revision))
 	} else {
-		log.Warnf("Could not auto-detect IP for. Use --ingressIP to manually specify the Gateway address to reach istiod from the VM.",
-			istiodHost(revision), istioNamespace)
+		log.Warnf("Could not auto-detect IP for. Use --ingressIP to manually specify the Gateway address to reach istiod from the VM.")
```
stevenctl, line 581:
```suggestion
		log.Warnf("Could not auto-detect IP for %s/%s. Use --ingressIP to manually specify the Gateway address to reach istiod from the VM.",
		istiodHost(revision), istioNamespace)
```

liamawhite (approved):

Issue #38500, opened 2022-04-21
Expose the op_timeout setting for the envoy.extensions.filters.network.redis_proxy.v3.RedisProxy filter


diranged:
**Describe the feature request**

Currently the `op_timeout` setting [is hard-coded to 5s](https://github.com/istio/istio/blob/f72bf28a23033983912c0facb705d83019fa3aef/pilot/pkg/networking/core/v1alpha3/networkfilter.go#L41-L42). By hard-coding this setting, it cannot be overridden with an `EnvoyFilter` (https://github.com/istio/istio/issues/38501), and yet it can also not be adjusted by the user.

Please implement a setting so that we can adjust this timeout in some way. I don't know where it should be configurable - but even in the simplest way, configuring it at the `istiod` level as an environment variable would be acceptable to us. We want to lower it to 500ms, not 5s.

**Describe alternatives you've considered**

I wanted to use an EnvoyFilter to patch this.. but it turns out that when Golang merges the pre-built filter and then my custom filter, the numbers are _added_ together. See https://github.com/istio/istio/issues/38501 for details.

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

zirain:
hardcode is bad, but redis support is leak in Istio. maybe add a `NodeMetadata`?
@howardjohn WDYT?

PR #38522, opened 2022-04-22
make redis op timeout configurable

zirain:

fix: https://github.com/istio/istio/issues/38500

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -623,6 +623,10 @@ type NodeMetadata struct {
 	// redirected tcp listeners. This does not change the virtualOutbound listener.
 	OutboundListenerExactBalance StringBool `json:"OUTBOUND_LISTENER_EXACT_BALANCE,omitempty"`
 
+	// RedisOpTimeout specifies the operation timeout for the Redis proxy filter, in duration format (10s).
+	// If not set, default timeout is 5s.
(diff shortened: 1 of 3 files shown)


ramaraochavali (approved):

diranged:
@zirain Just so I understand the PR - this allows us to set the `REDIS_OP_TIMEOUT` environment variable on the pilot containers, to override the default of `5s` and instead lower it to something else. Right?

howardjohn:
IMO we shouldn't add so many special cases for these super super obscure cases. You can always just set it to 1s to workaround the merge issues:

```
kind: EnvoyFilter
metadata:
  name: envoy-patch-op-timeout
spec:
  configPatches:
  - applyTo: NETWORK_FILTER
    match:
      listener:
        filterChain:
          filter:
            name: envoy.filters.network.redis_proxy
    patch:
      operation: MERGE
      value:
        name: envoy.filters.network.redis_proxy
        typed_config:
          '@type': type.googleapis.com/envoy.extensions.filters.network.redis_proxy.v3.RedisProxy
          settings:
            op_timeout: 1.0s
```

diranged:
Wait .. really? That's crazy. Unfortunately that is still too high for us - we have done extensive testing, and in our environments we need it set to 0.5s or even lower.  Good to know though.. really strange merge behavior!

PR #38536, opened 2022-04-22
[tf] New authz server component

nmittler:
Moving the authz server creation to a new test component that can be easily reused.

This is supporting the work on https://github.com/istio/istio/pull/37914.

### DIFF
--- a/pkg/test/framework/components/authz/kube.go
+++ b/pkg/test/framework/components/authz/kube.go
@@ -1,0 +1,231 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package authz
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/go-multierror"
+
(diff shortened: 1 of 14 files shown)


pkg/test/framework/components/authz/kube.go
```diff
@@ -0,0 +1,231 @@
...
+		},
+	}
+	server.id = ctx.TrackResource(server)
+
+	// Deploy the authz server.
+	if err = server.deploy(ctx); err != nil {
+		return
+	}
+
+	// Patch MeshConfig to intall the providers.
```
yangminzhu, line 122: s/intall/install

pkg/test/framework/components/authz/provider.go
```diff
@@ -0,0 +1,205 @@
...
+
+// Provider for authz requests.
+type Provider interface {
+	Name() string
+
+	// API used by this provider.
+	API() API
+
+	// IsProtocolSupported returns true if the given request protocol is supported by this provider.
+	IsProtocolSupported(protocol.Instance) bool
```
yangminzhu, line 45: Are these methods actually used in the test? I found only the `Name()` and `API()` is used. If they are needed in a follow-up PR, it's better to add in the follow-up PR together for easier review.

howardjohn (approved):

pkg/test/framework/components/authz/server.go
```diff
@@ -0,0 +1,89 @@
...
+	t.Helper()
+	s, err := NewLocal(t, ns)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return s
+}
+
+// Setup is a utility function for configuring a global authz Server.
+func Setup(server *Server, ns func() namespace.Instance) resource.SetupFn {
```
yangminzhu, line 64: The `Setup()` and `SetupLocal()` is also not used, probably delete them in this PR and add later when actually used.

yangminzhu (commented):
looks good, just some small comments.

nmittler:
@howardjohn @ericvn @GregHanson FYI the build seems broken currently. I looked briefly at history but nothing stands out. Is anyone looking into this?

PR #38490, opened 2022-04-21
Uniform pilot-agent and pilot-discover 's code style

vince-0202:
**Please provide a description of this PR:**

I noticed that pilot-agent's code style was different from pilot-discovery's,I wonder if it would be better like that.

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [*] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

PR #38542, opened 2022-04-24
Uniform pilot-agent and pilot-discover 's code style

vince-0202:
I noticed that pilot-agent's code style was different from pilot-discovery's,I wonder if it would be better like that.

I uniformed pilot-agent and pilot-discover's code style in this PR #38490 ,and suggested that I should fix it in master branch.


go.mod
```diff
@@ -78,7 +78,7 @@ require (
 	gomodules.xyz/jsonpatch/v3 v3.0.1
 	google.golang.org/api v0.74.0
 	google.golang.org/genproto v0.0.0-20220414192740-2d67ff6cf2b4
-	google.golang.org/grpc v1.45.0
+	google.golang.org/grpc v1.46.0
```
hzxuzhonghu, line 81: Donot couple this

hzxuzhonghu (commented):
I like the way removing the long init function. Just a little worried about we are lack of tests

vince-0202:
> I like the way removing the long init function. Just a little worried about we are lack of tests

I have the same concern,It's my first PR for Istio.Is there anything I can do to help with the testing

PR #38548, opened 2022-04-24
[tf] Refactor cleanup

nmittler:
We currently have 3 strategies for cleaning up config in the test framework:

1. no cleanup
2. always cleanup after the context completes
3. cleanup after the context completes if `noCleanup` was not specified.

This PR tries to make these options more clear and consistent across the test infra. It also add the strategy for an option to MeshConfig changes.

### DIFF
--- a/pkg/test/framework/resource/context.go
+++ b/pkg/test/framework/resource/context.go
@@ -15,71 +15,11 @@
 package resource
 
 import (
-	"istio.io/istio/pkg/test"
 	"istio.io/istio/pkg/test/framework/components/cluster"
+	"istio.io/istio/pkg/test/framework/resource/config"
+	"istio.io/istio/pkg/test/framework/resource/config/cleanup"
 	"istio.io/istio/pkg/test/util/yml"
 )
-
-type ConfigOptions struct {
-	NoCleanup bool
-	Wait      bool
-}
-
-type ConfigOption func(o *ConfigOptions)
-
-// NoCleanup does not delete the applied Config once it goes out of scope.
-var NoCleanup ConfigOption = func(o *ConfigOptions) {
-	o.NoCleanup = true
-}
-
-// Wait for the Config to be applied everywhere.
-var Wait ConfigOption = func(o *ConfigOptions) {
-	o.Wait = true
-}
-
-// ConfigFactory is a factory for creating new Config resources.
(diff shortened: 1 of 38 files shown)


pkg/test/framework/resource/context.go
```diff
@@ -20,20 +20,37 @@ import (
...
-type ConfigOptions struct {
-	NoCleanup bool
-	Wait      bool
+type ApplyOptions struct {
+	Cleanup CleanupStrategy
+	Wait    bool
 }
 
-type ConfigOption func(o *ConfigOptions)
+type ApplyOption func(o *ApplyOptions)
```
hzxuzhonghu, line 28: LGTM, just one question: why not use struct directly,

nmittler: I could go either way, really. I think my original idea was that most of the time when applying config, you don't need to specify anything, so the call would just be `cfg.Apply()`. If it were a struct, the call site would look like `cfg.Apply(resource.ApplyOptions{})`. TBH I don't feel strongly either way.

pkg/test/framework/resource/context.go
```diff
@@ -20,20 +20,37 @@ import (
...
+
+func (o CleanupStrategy) Option() func(opts *ApplyOptions) {
+	return func(opts *ApplyOptions) {
+		opts.Cleanup = o
+	}
 }
 
+const (
+	// CleanupAlways will trigger a cleanup operation the test context completes.
+	CleanupAlways CleanupStrategy = iota
```
howardjohn, line 41: Should we just call Option() on all of these as the const (well, `var` technically)? Not sure the value of each caller needing to call .Option?

howardjohn: we use the raw one in CleanupStrategy but seems like we could just as easily use the `Cleanup` field? Its only for internal library use as well as far as I can tell so doesn't leak abstractions much.

pkg/test/framework/resource/context.go
hzxuzhonghu: That is many api does  , even client-go

nmittler:
@howardjohn @hzxuzhonghu PTAL

douglas-reid (approved):
P&T changes look good.

howardjohn (approved):

PR #38571, opened 2022-04-25
Support TLSVerifySkip in http fetcher, and use context-based timeout

ingwonsong:
Supports InsecureSkipVerify of TLSClientConfig option in http fetcher as well.
In addition, instead of setting timeout on HTTP Client, use context-based timeout in order to match the API pattern with image fetcher.

- [ * ] Policies and Telemetry

### DIFF
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -64,6 +64,9 @@ type LocalFileCache struct {
 	// Map from tagged URL to checksum
 	checksums map[string]*checksumEntry
 
+	// http fetcher fetches Wasm module with HTTP get.
+	httpFetcher *HTTPFetcher
+
 	// directory path used to store Wasm module.
 	dir string
 
@@ -118,6 +121,7 @@ type cacheEntry struct {
 // NewLocalFileCache create a new Wasm module cache which downloads and stores Wasm module files locally.
 func NewLocalFileCache(dir string, purgeInterval, moduleExpiry time.Duration, insecureRegistries []string) *LocalFileCache {
 	cache := &LocalFileCache{
+		httpFetcher:        NewHTTPFetcher(),
 		modules:            make(map[moduleKey]*cacheEntry),
 		checksums:          make(map[string]*checksumEntry),
 		dir:                dir,
@@ -201,7 +205,7 @@ func (c *LocalFileCache) Get(
 	switch u.Scheme {
 	case "http", "https":
 		// Download the Wasm module with http fetcher.
-		b, err = DefaultHTTPFetcher().Fetch(ctx, downloadURL, insecure)
+		b, err = c.httpFetcher.Fetch(ctx, downloadURL, insecure)
 		if err != nil {
 			wasmRemoteFetchCount.With(resultTag.Value(downloadFailure)).Increment()
 			return "", err
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -201,7 +201,7 @@ func (c *LocalFileCache) Get(
 	switch u.Scheme {
 	case "http", "https":
 		// Download the Wasm module with http fetcher.
-		b, err = NewHTTPFetcher(insecure).Fetch(ctx, downloadURL)
+		b, err = DefaultHTTPFetcher().Fetch(ctx, downloadURL, insecure)
 		if err != nil {
 			wasmRemoteFetchCount.With(resultTag.Value(downloadFailure)).Increment()
 			return "", err
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -44,6 +44,13 @@ const (
 	// DefaultWasmModuleExpiry is the default duration for least recently touched Wasm module to become stale.
 	DefaultWasmModuleExpiry = 24 * time.Hour
 
+	// Default timeout per a HTTP/HTTPS request for HTTP/HTTPS-based wasm pulling.
+	DefaultWasmHTTPRequestTimeout = 5 * time.Second
+
+	// Default maximum number of HTTP/HTTPS request retries for HTTP/HTTPS-based wasm pulling.
+	// Note that, if the timeout specified in WasmPlugin is reaching out, then the pulling is stopped even though the retry count is still less than this value.
+	DefaultWasmHTTPRequestMaxRetries = 5
+
 	// oci URL prefix
 	ociURLPrefix = "oci://"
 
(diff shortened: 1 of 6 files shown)


pkg/wasm/httpfetcher.go
```diff
@@ -23,39 +25,45 @@ import (
 	"github.com/cenkalti/backoff/v4"
 )
 
+var (
+	httpInitialBackoff = time.Millisecond * 500
```
douglas-reid, line 29: any particular reason to move this out, instead of keeping it as part of `HTTPFetcher`?

ingwonsong: It is just a sort of constant rather than instance-wise variable, so I refactored it.

pkg/wasm/cache.go
```diff
@@ -127,10 +123,17 @@ func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration
...
+		insecure = true
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
 	switch u.Scheme {
 	case "http", "https":
 		// Download the Wasm module with http fetcher.
-		b, err = c.httpFetcher.Fetch(downloadURL, timeout)
+		b, err = NewHTTPFetcher(insecure).Fetch(ctx, downloadURL)
```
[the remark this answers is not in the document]
ingwonsong: Got it. I have changed this to reuse the connections.

zirain (commented):
overall lgtm, add new tests would be better. https://github.com/istio/istio/pull/37420 what petter done before may give some help.

pkg/wasm/httpfetcher.go
```diff
@@ -23,39 +25,53 @@ import (
...
+	defaultHTTPFetcher = NewHTTPFetcher()
```
hzxuzhonghu, line 30: Why not make it a field of the cache?  package var is notencouraged

https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables

ingwonsong: As I mentioned before, `httpInitialBackoff` is basically just a constant. But when testing, it is need to be shorten for making easy the test. Therefore, it is defined as a package variable.
In case of `defaultHTTPFetcher` is a singleton. 

Do we need to include them for each instance of caches? 
WDYT?

hzxuzhonghu: I am not sure why each instance of caches, there is only one cache instance

ingwonsong: Yes. it is true. I just thought about loosening the coupling between cache and fetcher.
BTW, since I agree with that the change is not necessary and may be not-good in some sense, so I moved it back to the field of the cache.

Thanks!

pkg/wasm/httpfetcher_test.go
```diff
@@ -86,3 +90,62 @@ func TestWasmHTTPFetch(t *testing.T) {
...
+
+	cases := []struct {
+		name            string
+		handler         func(http.ResponseWriter, *http.Request, int)
+		insecure        bool
+		wantNumRequest  int
+		wantErrorSuffix string
+	}{
+		{
+			name: "download ok",
```
zirain, line 105: nit: download failed

pkg/wasm/httpfetcher.go
```diff
@@ -25,37 +27,43 @@ import (
...
 	attempts := 0
-
 	b := backoff.NewExponentialBackOff()
 	b.InitialInterval = f.initialBackoff
 	b.Reset()
 	var lastError error
 	for attempts < 5 {
 		attempts++
-		resp, err := c.Get(url)
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```
hzxuzhonghu, line 61: nit: More grace way is to check the attempts times and ctx.Done simultaneously

hzxuzhonghu: Or I guess the easier way is to set the Client.Timeout = 10s

ingwonsong: Good catch! I have added that.

hzxuzhonghu (approved):

pkg/wasm/cache.go
```diff
@@ -90,7 +90,7 @@ type cacheEntry struct {
 // NewLocalFileCache create a new Wasm module cache which downloads and stores Wasm module files locally.
 func NewLocalFileCache(dir string, purgeInterval, moduleExpiry time.Duration, insecureRegistries []string) *LocalFileCache {
 	cache := &LocalFileCache{
-		httpFetcher:        NewHTTPFetcher(),
+		httpFetcher:        NewHTTPFetcher(5 * time.Second),
```
jacob-delgado, line 93: Should this be user configurable via an environment variable? 5s seems reasonable for most cases, but may not be suitable for some users.

ingwonsong: I totally agree with the comment.

Currently, many variables are hard-coded regarding `cache` and `fetcher`. 
For example, at the top of pkg/wasm/cache.go, we can see `DefaultWasm.*` const and some of them seems to be configurable. 

One of the problem here is, if we want to make them configurable, then one of the functions of cache should take a long list of parameters, and it may causes poor readability. So, I am thinking about refactoring PR which is including `cache.go` as well as `imagefetcher.go`/`httpfetcher.go`.

Therefore, I would like to suggest to push making them configurable back, and at now, just keep the timeout per request and the number of retries as the constants.
(Please note that I pushed a commit to use the `DefaultWasmHTTPRequestTimeout` constant instead of directly using `5 * time.Second`.)

WDYT?

pkg/wasm/cache.go
```diff
@@ -161,10 +168,17 @@ func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration
 	// Hex-Encoded sha256 checksum of binary.
 	var dChecksum string
 	var binaryFetcher func() ([]byte, error)
+	insecure := false
+	if c.insecureRegistries.Contains(u.Host) {
+		insecure = true
+	}
```
howardjohn, line 174: nit
```suggestion
	insecure := c.insecureRegistries.Contains(u.Host)
```

pkg/wasm/httpfetcher.go
```diff
@@ -25,37 +27,50 @@ import (
...
-
 	b := backoff.NewExponentialBackOff()
 	b.InitialInterval = f.initialBackoff
 	b.Reset()
 	var lastError error
-	for attempts < 5 {
+	for attempts < DefaultWasmHTTPRequestMaxRetries {
 		attempts++
-		resp, err := c.Get(url)
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```
howardjohn, line 68: How does context timeout and requestTimeout correlate? if eiter is rtiggered we timeout?

howardjohn (approved):

pkg/wasm/httpfetcher.go
ingwonsong: "context timeout" is triggered when the specified time is passed after calling `WithTimeout`,
and requestTimeout is a timeout for each http request attempt.

For example, if we have `context timeout = 7s, requestTimeout = 5s` and server does not respond, then, with very high possibility:
* First http request attempt might be timed out by `requestTimeout`. So, move to next attempts.
* Second attempt might be timed out by `context timeout`, so all the operation using this context will fail.

While checking your comment, I realized that the retry loop is still going after context timeout reach. Of course, each attempt will fail after the context timeout, but meaningless. I have patched it.

zirain (approved):

PR #24874, opened 2020-06-22
Fix test cleanup of RunParallel

howardjohn:
I am guessing this will solve
https://github.com/istio/istio/issues/24028

I am not proud of this solution, but I spent quite a few hours on this
and its the best I could come up with. The sequencing here is quite
complicated...

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #38576, opened 2022-04-25
[tf] Fix deadlock for Parallel tests

nmittler:
I ran into a deadlock condition when running a number of nested parallel tests. Basically, some tests were stuck logging `Stuck waiting for parent test suites to terminate...`.

I recreated the issue in framework_test.go and the fix in this PR also fixes this test.

### DIFF
--- a/pkg/test/framework/integration/framework_test.go
+++ b/pkg/test/framework/integration/framework_test.go
@@ -16,125 +16,239 @@ package integration
 
 import (
 	"fmt"
-	"sync"
 	"testing"
 	"time"
-
-	"go.uber.org/atomic"
 
 	"istio.io/istio/pkg/test/framework"
 )
 
+func TestSynchronous(t *testing.T) {
+	// Root is always run sync.
+	newLevels().
+		// Level 1: bf=5(sync)
+		Add(5, false).
+		// Level 2: bf=2(sync)
+		Add(2, false).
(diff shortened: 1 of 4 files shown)


nmittler:
@howardjohn this reverts and modifies the solution in https://github.com/istio/istio/pull/24874

pkg/test/framework/integration/framework_test.go
```diff
@@ -16,125 +16,239 @@ package integration
...
+}
+
+func (tst *test) cleanup() {
+	tst.cleanupStart = time.Now()
+	doWork()
+	tst.cleanupEnd = time.Now()
+}
+
+func doWork() {
+	time.Sleep(time.Millisecond * 50)
```
howardjohn, line 241:
```suggestion
	time.Sleep(time.Millisecond
```

Do we need full 50ms sleep? Seems it is sufficient testing with very small and we do not get a ~15s runtime

howardjohn (commented):
LGTM if tests pass. Way simpler!

pkg/test/framework/integration/framework_test.go
nmittler: Reduced it to 10 millis

nmittler:
@howardjohn PTAL

howardjohn (approved):

PR #38577, opened 2022-04-25
Switch Istio base image to use Ubuntu Jammy

sergii-ssh:
**Please provide a description of this PR:**

Update base image to use ubuntu jammy

PR #38581, opened 2022-04-26
Revert "Update BASE_VERSION to master-2022-04-25T20-21-10"

ericvn:
Reverts istio/istio#38572

Let's revert this so that we get a new set of build images during tomorrow's automated check. This set of base images will be based on Jammy due to https://github.com/istio/istio/pull/38577.


ericvn:
Looks like we already have a new image coming.

PR #38595, opened 2022-04-26
pilot: remove concept of `plugins`

howardjohn:
The plugin concept is not something that is adding value to Istio
currently, and causes a number of problems. This PR is meant to be a
refactoring (no behavioral changes or XDS output changes) to remove the
plugin concept. I have a number of PRs that build upon this to further
improve the code around here, so this is meant to just be part 1.

Issues with plugin:
* Plugins are not actually pluggable - there is no mechanism to inject
  plugins at runtime. Almost all extensibility in Istio these days is
done through EnvoyFilters (eg `aeraki`). The one case I know using
plugins is Cilium; this is done by having a fork of Istio. This could be
handled just as easily by inserting 1 line in
`buildHTTPConnectionManager` rather than using a plugin. The plugin
interface was never stable, either, so there was no expectation that a
plugin could be zero-maintenance on upgrades/rebases/etc.
* Plugins are not actually generic - for example, we have a plugin for
  Authn. This implements the InboundMTLSConfiguration. There can only be
a single plugin that implements the InboundMTLSConfiguration interface,
and logically only the Authn plugin we have can meet it's requirements.
* Plugins work on the wrong abstractions. Currently we only have
  listener mutation. However, we do not even build 'Listeners' in most
cases - we build filter chains. Soon, even outbound will have one
listener most likely. This issue actually causes major code issues; we
build fake listeners for inbound then later convert them to filter
chains (I have a PR fixing this, building upon this PR).
* Plugin inputs are too large - this has been improved over the years,
  but still an issue. `InputParams` takes a lot of inputs, and many call
sites do not have all of them so they make up dummy inputs. This is
super fragile -- a plugin has access to a bunch of fields but many of
them are unsafe to use.
* Plugins are not flexible - we only expose a minimal amount of
  integration points.
* Plugins are fragile - all plugins work by just appending a filter to
  the end of the filter list. But filter order is extremely critical!

For these reasons, I suggest we finally remove the plugin concept
entirely. Instead, we directly insert the authn and authz filters where
needed.

Note: there is some duplication here, and leaky abstractions a bit in
the TCP filter application (HTTP is fine, IMO). I have some improvements
in another PR but do not want to make this PR too large.


ramaraochavali:
Huge +1 to this. The current plugins are not serving any purpose and are making code unnecessarily complex. I have not taken a look at the entire PR yet. But I am good to approve if others are good.

pilot/pkg/bootstrap/server.go
```diff
@@ -216,7 +207,7 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
 	}
 	// Initialize workload Trust Bundle before XDS Server
 	e.TrustBundle = s.workloadTrustBundle
-	s.XDSServer = xds.NewDiscoveryServer(e, args.Plugins, args.PodName, args.Namespace, args.RegistryOptions.KubeOptions.ClusterAliases)
```
ramaraochavali, line 219: is it intentional to remove namespace?

ramaraochavali (approved):

stevenctl (approved):

lei-tang (approved):

pilot/pkg/bootstrap/server.go
howardjohn: Yeah its not used (existing dead code). Already was cleaning up plugins so figured would do both

howardjohn:
/test unit-tests_istio
/test release-test_istio

For some reason never started

PR #38600, opened 2022-04-26
Fix bad validatingwebhookconfiguration gen

### LINKED ISSUE #38599: istioctl install generates bad validatingwebhookconfiguration
### Bug Description

Installing a gateway using an empty profile [like this](https://istio.io/latest/docs/setup/install/external-controlplane/#enable-gateways) on a config cluster (with an external control plane) creates a `istiod-default-validator` `ValidatingWebhookConfiguration` using the injection URL (`<external-host>/inject/cluster=...`), instead of the validation URL (`<external-host>/validate`) for the `clientConfig.url`.

I think there are two problems here:

1. Creating a gateway using an empty profile should not trigger webhook generation.
2. The webhook needs to use the proper URL for the validator.

### Version

```prose
1.13.2
```

### Additional Information

_No response_

frankbu:

Fixed #38599

### DIFF
--- a/istioctl/pkg/tag/generate.go
+++ b/istioctl/pkg/tag/generate.go
@@ -145,15 +145,15 @@ func Generate(ctx context.Context, client kube.ExtendedClient, opts *GenerateOpt
 	return tagWhYAML, nil
 }
 
-func fixWhConfig(injectionWhConfig *tagWebhookConfig) *tagWebhookConfig {
-	if injectionWhConfig.URL != "" {
-		webhookURL, err := url.Parse(injectionWhConfig.URL)
+func fixWhConfig(whConfig *tagWebhookConfig) *tagWebhookConfig {
+	if whConfig.URL != "" {
+		webhookURL, err := url.Parse(whConfig.URL)
 		if err == nil {
 			webhookURL.Path = "/validate"
-			injectionWhConfig.URL = webhookURL.String()
+			whConfig.URL = webhookURL.String()
 		}
 	}
-	return injectionWhConfig
+	return whConfig
 }
 
 // Create applies the given tag manifests.
--- a/istioctl/pkg/tag/generate.go
+++ b/istioctl/pkg/tag/generate.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 
@@ -130,6 +131,13 @@ func Generate(ctx context.Context, client kube.ExtendedClient, opts *GenerateOpt
 		// TODO(Monkeyanator) should extract the validationURL from revision's validating webhook here. However,
 		// to ease complexity when pointing default to revision without per-revision validating webhook,
 		// instead grab the endpoint information from the mutating webhook. This is not strictly correct.
+		if tagWhConfig.URL != "" {
+			webhookURL, err := url.Parse(tagWhConfig.URL)
+			if err == nil {
+				webhookURL.Path = "/validate"
+				tagWhConfig.URL = webhookURL.String()
+			}
+		}
 		vwhYAML, err := generateValidatingWebhook(tagWhConfig, opts.ManifestsPath)
 		if err != nil {
 			return "", fmt.Errorf("failed to create validating webhook: %w", err)
--- a/istioctl/pkg/tag/generate.go
+++ b/istioctl/pkg/tag/generate.go
@@ -131,14 +131,9 @@ func Generate(ctx context.Context, client kube.ExtendedClient, opts *GenerateOpt
 		// TODO(Monkeyanator) should extract the validationURL from revision's validating webhook here. However,
 		// to ease complexity when pointing default to revision without per-revision validating webhook,
 		// instead grab the endpoint information from the mutating webhook. This is not strictly correct.
-		if tagWhConfig.URL != "" {
-			webhookURL, err := url.Parse(tagWhConfig.URL)
-			if err == nil {
-				webhookURL.Path = "/validate"
-				tagWhConfig.URL = webhookURL.String()
-			}
-		}
-		vwhYAML, err := generateValidatingWebhook(tagWhConfig, opts.ManifestsPath)
+		validationWhConfig := fixWhConfig(tagWhConfig)
+
+		vwhYAML, err := generateValidatingWebhook(validationWhConfig, opts.ManifestsPath)
 		if err != nil {
 			return "", fmt.Errorf("failed to create validating webhook: %w", err)
 		}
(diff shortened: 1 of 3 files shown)


operator/cmd/mesh/install.go
```diff
@@ -204,7 +204,7 @@ func Install(rootArgs *RootArgs, iArgs *InstallArgs, logOpts *log.Options, stdOu
 		return fmt.Errorf("failed to install manifests: %v", err)
 	}
 
-	if !exists || rev == "" {
+	if !exists || rev == "" && pilotEnabled {
```
frankbu, line 207: This prevents webhook manipulation from being an unexpected side effect for things like simply deploying a gateway with an empty profile.

istioctl/pkg/tag/generate.go
```diff
@@ -130,6 +131,13 @@ func Generate(ctx context.Context, client kube.ExtendedClient, opts *GenerateOpt
 		// TODO(Monkeyanator) should extract the validationURL from revision's validating webhook here. However,
 		// to ease complexity when pointing default to revision without per-revision validating webhook,
 		// instead grab the endpoint information from the mutating webhook. This is not strictly correct.
+		if tagWhConfig.URL != "" {
+			webhookURL, err := url.Parse(tagWhConfig.URL)
+			if err == nil {
+				webhookURL.Path = "/validate"
+				tagWhConfig.URL = webhookURL.String()
```
frankbu, line 138: @Monkeyanator This fix is still "not strictly correct" either, but the original code is incorrect. At some point I think this should really be fixed properly.

litong01: feel that the fix will make the logic even more complex. don't we have some kind of very clear logic to indicate if the webhook should be touched? such as check if istiod is included or this is a remote cluster install or some other conditions? I have tried this case, install remote cluster which has webhook installed and the external istiod also injected the caBundles to the webhook, then when I install a gateway, it completely messed up the webhook, it made the external istiod think there is a leader election with different root ca and produced unauthorized requests. After awhile digging, I figured it was due to the install messing with the webhooks. was not easy to figure out why.

frankbu: @litong this change has nothing to do with what you're describing. It's simply fixing the incorrect URL in the default validating webhook configuration. Line 207 in the other file is trying to prevent generating the webhook when all that's being installed is a gateway.

litong01: @frankbu ok.

howardjohn (commented):
Any tests?

howardjohn (commented):
Test comment was not just about adding coverage - I do not understand how it can possibly work in any circumstances today?

The issue describes a very isolated case of Gateway only in config cluster, but doesn't this always use the MWH url in the VWH which is always wrong (currently)?

frankbu:
> The issue describes a very isolated case of Gateway only in config cluster, but doesn't this always use the MWH url in the VWH which is always wrong (currently)?

This should be tested in the istiodremote integration tests. Even though it's a bad webhook, the other installed one (which, unlike injection webhooks, is not deactivated) works. I think there may be more issues with this code, but this PR is just trying to just get it sortof correct but also disable generating it in the config cluster for now.

frankbu:
@howardjohn Actually, I see that we use injectionPath, not injectionURL, for the external control plane integration test, so this does not get tested. I noticed it when setting up an external control plane manually using the documented procedure, which does use the injectionURL (for a remote public DNS gateway). The doc test doesn't test validation, so it might not be working thanks to this bug?

howardjohn (commented):
Please add unit tests to generate_test.go. This is super complex, hard to tell what is going on without before/after test diff.

operator/cmd/mesh/install.go
howardjohn: Webhook injection is required for (some) gateways though, so this might cause issues...?

istioctl/pkg/tag/generate_test.go
```diff
@@ -141,6 +143,14 @@ func TestGenerateValidatingWebhook(t *testing.T) {
...
 			istioNamespace: "istio-system",
 			webhook:        revisionCanonicalWebhookRemote,
+			whURL:          remoteValidationURL,
+			whSVC:          "",
+			whCA:           "ca",
+		},
+		{
+			name:           "webhook-configured-with-bad-url",
+			istioNamespace: "istio-system",
+			webhook:        revisionCanonicalWebhookRemote,
```
howardjohn, line 153: I don't understand this test. The inputs are the same as `webhook-pointing-to-url` but the expectations are different and both pass... I don't understand why we conditionally skip the test on L200

frankbu: The URL field is different in this test: `whURL:          remoteInjectionURL,`. The `webhook-pointing-to-url` test uses: `remoteValidationURL`, which is a proper validation URL. This PR is needed because the code is incorrectly using an injection URL, which this PR is correcting. As Sam's TODO on line 131 points out, this is not strictly correct, but fixing the code to do it properly is a bigger effort (so I left Sam's TODO).

howardjohn: But the whURL is the _expected output_ of the test, not the inputs...? So what we are saying, in my understanding is for the same inputs, we expected 2 different outputs, and both tests pass. That only works since the test is not actually testing the expected output due to skippiung it on L200?

operator/cmd/mesh/install.go
frankbu: IIUC this should get done when an istiod is installed, not when just a gateway is being installed. Note that this isn't done at all when `istioctl manifest generate` is used instead of `istioctl install`, so it really should never matter? I must admit that this default webhook code seems pretty complicated and I'm not sure I understand if it's really working properly in all cases

istioctl/pkg/tag/generate_test.go
frankbu: You're right ... my brain isn't working so well today. I wanted to quickly test the case of a good URL, but that would require an injection webhook configured with a validation url, which never happens anyway. I think the only thing that matters is that this fix is changing injection URLs (which are always wrong), into a validation URLs. I changed the test. PTAL

operator/cmd/mesh/install.go
howardjohn: Yeah fair point on generate vs install. My concern was

control cluster: install full
remote cluster: install just gateway

But is the 'remote cluster' case still meeting `pilotEnabled` and it is just not literally setting up the Istiod deployment?

howardjohn (approved):

operator/cmd/mesh/install.go
frankbu: Hmm, not sure. In the docs, we are using `istioctl manifest generate` on the remote cluster, not sure why. Maybe workaround for another issue? If I have some spare cycles next week, I'll look into this some more.

esnible (approved):
LGTM

PR #38603, opened 2022-04-26
Using tag/digest trimmed URL for OCI images and having tagged URL to digest

ingwonsong:
This PR is proposing use of URL which the tag or digest suffix is trimmed as a cache key in terms of OCI image cache.

In terms of URL for OCI images, two forms can be used.
- tagged form: docker.io/foo/image:bar
- digest form: docker.io/foo/image@sha256:1234567890abcdef

Basically, since the tag is just an alias for a specific version or digest, using (URL having no tag/digest + Digest) is much natural approach. Even in current implementation, the cached file is named with the digest.
Of course, to provide cache functionality for tagged URL as well, we additionally need to keep the map from the tagged URL to the digest according to the original meaning of tag.

Practically, there are two benefits:

1. In current implementation, without use of the digest(sha256) field in WasmPlugin, tagged OCI URL has "Always" pull policy because we don't keep any recently used digest.
But, with this PR, we can easily implement "IfNotPresent" on the top of it, because the map from the tag to the digest can provide the cached digest which was used recently.

2. Let's assume that an image was pulled with a URL suffixed by a digest. If the tagged URL is used and the tag is pointing the image having "the digest", then the image might be found in cache with this PR.

Note that this is a part of implementing PullPolicy of WasmPlugin, this will be followed the other PR including https://github.com/istio/istio/pull/38475.

### DIFF
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -28,9 +28,10 @@ import (
 	"sync"
 	"time"
 
-	"github.com/google/go-containerregistry/pkg/name"
 	"istio.io/istio/pkg/util/sets"
 	"istio.io/pkg/log"
+
+	"github.com/google/go-containerregistry/pkg/name"
 )
 
 var wasmLog = log.RegisterScope("wasm", "", 0)
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -28,10 +28,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/go-containerregistry/pkg/name"
+
 	"istio.io/istio/pkg/util/sets"
 	"istio.io/pkg/log"
-
-	"github.com/google/go-containerregistry/pkg/name"
 )
 
 var wasmLog = log.RegisterScope("wasm", "", 0)
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -86,7 +86,7 @@ var _ Cache = &LocalFileCache{}
 type moduleKey struct {
 	// Identifier for the module. It should be neutral for the checksum.
 	// e.g.) oci://docker.io/test@sha256:0123456789 is not allowed.
-	//       oci://docker.io/test@latest (tagged form) is allowed.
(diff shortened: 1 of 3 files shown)


pkg/wasm/cache.go
```diff
@@ -211,20 +254,43 @@ func (c *LocalFileCache) addEntry(key cacheKey, wasmModule []byte, f string) err
...
 	wasmCacheEntries.Record(float64(len(c.modules)))
 	return nil
 }
 
 func (c *LocalFileCache) getEntry(key cacheKey) string {
 	modulePath := ""
 	cacheHit := false
+
+	// Only apply this for OCI image, not http/https because OCI image has ImagePullPolicy
+	// to control the pull policy, but http/https currently rely on existence of checksum.
```
zirain, line 274: what about tagged oci url with sha256?

ingwonsong: In that case, we already have sha256, so the sha256 digest will be used for finding an entry in the cache. If there is no entry in cache, the tagged URL will be used for retrieving an image from registry server.

So, the case is working as is.

zirain (commented):
lgtm, leave to @howardjohn .

pkg/wasm/cache.go
```diff
@@ -72,25 +83,35 @@ type LocalFileCache struct {
 
 var _ Cache = &LocalFileCache{}
 
+type moduleKey struct {
+	// Identifier for the module. It should be neutral for the checksum.
+	// e.g.) oci://docker.io/test@sha256:0123456789 is not allowed.
+	//       oci://docker.io/test@latest (tagged form) is allowed.
```
howardjohn, line 89:
```suggestion
	//       oci://docker.io/test:latest (tagged form) is allowed.
```

howardjohn: although in the code looks like we strip off the tag?

howardjohn (approved):

pkg/wasm/cache.go
ingwonsong: Yes, just for future changes, I took a note there to keep the intention.

zirain (approved):

PR #38605, opened 2022-04-26
stackdriver: add support for testing with real Cloud Monitoring backend

douglas-reid:
For "Real Stackdriver" tests, we must use the GCP resource hierarchy -- as data will be written against resources in various GCP projects. In certain testing environments, it may desirable to execute the tests in different contexts (including GKE clusters in distinct projects) from the projects that the data plane proxies are running in. In these cases, the Test Framework needs to support a way to "discover" the resource project for a cluster. This CL establishes a mechanism for that via `cluster.Config`'s `Meta` which is mapped into a new field in `Topology`.

With such a mechanism in place, the CL then expands the Test Framework stackdriver component interface to allow specification of a resource project as necessary (falling back to the discovered project of the running test code).

Test invocations can then establish the appropriate project information via `cluster.Config` as needed per environment. If left unspecified (as it is here) the project of the test will be used by default. As a result, this should be a no-op for the vast majority of Istio testing.


douglas-reid:
Attempting to recreate prior failures (which I was not able to duplicate in my personal testing environment)

nmittler (approved):

kyessenov (approved):

PR #38613, opened 2022-04-27
test not set inbound alpn

hzxuzhonghu:


hzxuzhonghu:
@howardjohn From the failure it seems inbould listener downstream tls context alpnProtocol can be any value. 

The only requirement is for TCP mx is it should include "istio-peer-exchange"

That's very weird.

howardjohn:
I think envoy doesn't really do ALPN negotiation. But the MX filter explicitly looks for that ALPN

hzxuzhonghu:
I guess you are right

Issue #38501, opened 2022-04-21
EnvoyFilter "MERGE" behavior will add values together, rather than replacing them!


diranged:
### Bug Description

When trying to solve https://github.com/istio/istio/issues/38500, I thought that I could use an `EnvoyFilter` to override the `op_timeout` setting like this:

```
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: envoy-patch-op-timeout
spec:
  configPatches:
  - applyTo: NETWORK_FILTER
    match:
      listener:
        filterChain:
          filter:
            name: envoy.filters.network.redis_proxy
    patch:
      operation: MERGE
      value:
        name: envoy.filters.network.redis_proxy
        typed_config:
          '@type': type.googleapis.com/envoy.extensions.filters.network.redis_proxy.v3.RedisProxy
          settings:
            op_timeout: 0.5s
```

However, when you apply this patch, the diff looks like this on the Envoy config:
```diff
-                op_timeout: 5s
+                op_timeout: 5.500s
```

The final filter config looks like this:
```
          - name: envoy.filters.network.redis_proxy
            typed_config:
              '@type': type.googleapis.com/envoy.extensions.filters.network.redis_proxy.v3.RedisProxy
              latency_in_micros: true
              prefix_routes:
                catch_all_route:
                  cluster: outbound|6379||test.test.svc.cluster.local
              settings:
                op_timeout: 5.500s
              stat_prefix: outbound|6379||test.test.svc.cluster.local
```

I cannot use `REPLACE` here to replace the filter, because of dynamic values like the `prefix_routes` and `stat_prefix` settings ... and the fact that there is no way to write an `EnvoyFilter` to patch a specific TCP listener because you cannot know the name that is going to be generated, since that is based on the IP address assigned.

### Version

```prose
$ istioctl version
client version: 1.13.3
control plane version: 1.13.2
data plane version: 1.13.0 (11 proxies), 1.13.2 (43 proxies)
```

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:51:05Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"darwin/arm64"}
[...]

PR #38617, opened 2022-04-27
EnvoyFilter: use custom merge logic

howardjohn:
The proto.Merge logic is problematic in a number of cases. In this instance, Duration does a blind merge of the `{Seconds, Nanoseconds}` struct, which means we end up adding these together. We have seen other areas where proto.Merge has caused problems as well.

This PR follows the path Kuma took, vendoring in the proto.Merge functinon from the protobuf library with support added for custom per-type merge functions. Code is mostly copied from them with attribution in the code (thanks!!). 

Currently, the only custom logic is Duration, but this gives us future flexibility

Fixes https://github.com/istio/istio/issues/38501
Fixes https://github.com/istio/istio/issues/38500 and reverts the original fix in https://github.com/istio/istio/pull/38522

cc @zirain @diranged

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -623,10 +623,6 @@ type NodeMetadata struct {
 	// redirected tcp listeners. This does not change the virtualOutbound listener.
 	OutboundListenerExactBalance StringBool `json:"OUTBOUND_LISTENER_EXACT_BALANCE,omitempty"`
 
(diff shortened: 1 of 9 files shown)


ramaraochavali (commented):
Generally LGTM. But test failures seem related

howardjohn:
will fix

ramaraochavali (approved):

nmittler (approved):

PR #38630, opened 2022-04-27
Check the cache once again just after getting the manifest of OCI image

ingwonsong:
This PR proposes to utilize the digest coming from the manifest of OCI image. 
Just after retrieving the manifest, we can calculate the digest before getting the image itself.
So, if there is a corresponding cache entry, we can use it instead of retrieving the image data.

This PR is originated from the [comment](https://github.com/istio/istio/pull/38475#discussion_r854532908) of @howardjohn.


pkg/wasm/cache.go
```diff
@@ -160,19 +156,34 @@ func (c *LocalFileCache) Get(downloadURL, checksum string, timeout time.Duration
 		}
 		wasmLog.Debugf("wasm oci fetch %s with options: %v", downloadURL, imgFetcherOps)
 		fetcher := NewImageFetcher(ctx, imgFetcherOps)
-		b, dChecksum, err = fetcher.Fetch(u.Host+u.Path, checksum)
+		binaryGetter, dChecksum, err = fetcher.Fetch(u.Host + u.Path)
```
zirain, line 159: a function named `Fetch` but actually you may done get binary on L180, how about add a new function check the cache, WDYT?

zirain: I think it's ok to let `HttpFetcher` different with `ImageFetcher`.

Issue #38554, opened 2022-04-25
Can't get the target port from the serviceentry generated by the xds api generator


zhaohuabing:
### Bug Description

Can't get the target port from the service entry generated by the xds API generator. The target port is not set in the code.

https://github.com/istio/istio/blob/b55bf1a7debc0ce0081e91ad25694e839737512b/pilot/pkg/serviceregistry/serviceentry/conversion.go#L118

### Version

```prose
client version: 1.10.0
control plane version: 1.10.0
data plane version: 1.10.0 (1 proxies), 1.10-dev (3 proxies), 65536.65536.65536 (1 proxies)

The latest version has this issue as well.
```

zhlsunshine:
Hi @zhaohuabing, according to the comment, do you have any idea about the PR [here](https://github.com/istio/istio/pull/38634#discussion_r860913433)?

PR #38634, opened 2022-04-28
add target port for internal server for service entry and k8s service

zhlsunshine:
Add target port for internal server for service entry and k8s service and fix [issue#38554](https://github.com/istio/istio/issues/38554)


pilot/pkg/model/service.go
```diff
@@ -193,6 +193,10 @@ type Port struct {
 
 	// Protocol to be used for the port.
 	Protocol protocol.Instance `json:"protocol,omitempty"`
+
+	// The port number on the endpoint where the traffic will be received.
+	// It can map the k8s service or ServiceEntries with targetPort specified
+	TargetPort int `json:"target_port,omitempty"`
```
howardjohn, line 199: This is not how the internal model is intended to be used. TargetPort only makes sense in the context of a ServiceInstance which already has this (EndpointPort)

A service does not just have one singular int target port

zhlsunshine: If so, target port will be missing when using `ServiceToServiceEntry` [here ](https://github.com/istio/istio/blob/b55bf1a7debc0ce0081e91ad25694e839737512b/pilot/pkg/networking/apigen/apigen.go#L129). 
One question, could I pass another parameter with ServiceInstance  for `ServiceToServiceEntry` to fetch the target port?

howardjohn: I think there are multiple levels of problems with that function

1. SE -> service -> is lossy
2. ServiceEntry cannot fully describe a Kubernetes Service (which allows target port by name)

zhlsunshine: 1. SE -> service -> is lossy  
Yes
2. ServiceEntry cannot fully describe a Kubernetes Service (which allows target port by name) 
Yes, that's why there are so many `TODO`s in this function, I think istio service and proxy can not really convert corresponding ServiceEntry.

zhlsunshine: BTW, I think all necessary information should be passed to `ServiceToServiceEntry` to complete this conversion, what's your suggestion? @howardjohn

hzxuzhonghu: I suggest we donot add this unless we need it internally, in case of abuse later.

zhlsunshine: Okay, @hzxuzhonghu, I will try another way to get the target port.

zhlsunshine:
Hi @howardjohn & @hzxuzhonghu,  new submit for this PR, any comment? Thanks! ^_^

pilot/pkg/serviceregistry/serviceentry/conversion.go
```diff
@@ -110,12 +129,13 @@ func ServiceToServiceEntry(svc *model.Service, proxy *model.Proxy) *config.Confi
 
 	// Port is mapped from ServicePort
 	for _, p := range svc.Ports {
+		targetPort := getTargetPortFromServiceInstances(p, svc, proxy.ServiceInstances)
```
hzxuzhonghu, line 132: This is not right. As the serviceinstance here is only this proxie's.  May not be the service's

zhlsunshine: u r right, the svc should be imported across all egress listeners. Thanks!

zhlsunshine: Hi @hzxuzhonghu, new submit for picking the service instances from PushRequest. Thanks!

pilot/pkg/serviceregistry/serviceentry/conversion.go
```diff
@@ -43,6 +43,25 @@ func convertPort(port *networking.Port) *model.Port {
...
+	if len(serviceInstances) == 0 {
+		return uint32(0)
+	}
+
+	for _, instance := range serviceInstances {
+		if svc != instance.Service {
+			continue
+		}
+		if port.Name == instance.ServicePort.Name {
+			return instance.Endpoint.EndpointPort
```
howardjohn, line 59: There may be 100 different "target ports" for one service. Trying to encode a single target port for a service is not consistent with Kubernetes Service

zhlsunshine: How about to iterate them one by one for the Service Entry, and append these ports into the Service Entry's Ports? Any idea?

zhlsunshine: Is there any problem for a Service Entry has multiple ports with the same port name but different target names?

howardjohn: I think the function is fundamentally broken and cannot be fixed... a ServiceEntry cannot represent a Service properly

zhaohuabing: @howardjohn @zhlsunshine It seems that there's no perfect solution for this issue. How about we do it this way:
If the k8s svc target port is a number, we put that number at the target port of the service entry, otherwise, leave it as it is now.  This at least would partly solve this problem. And I can ask the users to change the target port name to a number if they encounter this issue. 99% of svcs instances that I know have the same target number.  So this would work on most occasions.

zhaohuabing:
Related issue: https://github.com/istio/istio/issues/38554

pilot/pkg/model/service.go
zhaohuabing: > al model is intended to be used. TargetPort only makes sense in the context of a ServiceInstance which already has this (EndpointPort)

How about we make TargetPort can be either a number or a name? Just like k8s svc.

zhaohuabing: > This is not how the internal model is intended to be used. TargetPort only makes sense in the context of a ServiceInstance which already has this (EndpointPort)
> 
> A service does not just have one singular int target port

@howardjohn How about we make TargetPort can be either a number or a name? Just like k8s svc.

zhlsunshine:
Hi @howardjohn, @hzxuzhonghu and @zhaohuabing, based on current implementation,  re-submit again to add multiple ports to a service entry, and without duplicated target port. Any comment?

zhaohuabing:
> Hi @howardjohn, @hzxuzhonghu and @zhaohuabing, based on current implementation, re-submit again to add multiple ports to a service entry, and without duplicated target port. Any comment?

This would result in an incomplete target ports list in some cases. I would suggest getting the target port from the k8s service definition instead of inferring from service instances. Target port can be a fixed number or 0(not set) in case that k8s svc target port is a name.

pilot/pkg/networking/apigen/apigen.go
```diff
@@ -126,7 +126,7 @@ func (g *APIGenerator) Generate(proxy *model.Proxy, w *model.WatchedResource, re
 			if s.Attributes.ServiceRegistry == provider.External {
 				continue
 			}
-			c := serviceentry.ServiceToServiceEntry(s, proxy)
+			c := serviceentry.ServiceToServiceEntry(s, proxy, req.Push)
```
howardjohn, line 129: Taking a step back... what is the point of this? What use is a Service without the endpoints?

zhaohuabing: Let me explain my use case so we can understand why we need target port in the service entry:

To support non-HTTP protocols in Istio, Aeraki retrieves service entries from Istio via xds, using service entries to generate envoyfilters to replace the original TCP filter in the outbound and inbound listeners. Here is an example of an envoyfilter for inbound listeners:

```
kind: EnvoyFilter
metadata:
  labels:
    manager: Aeraki
  name: aeraki-inbound-thrift-sample-server.meta-thrift.svc.cluster.local-9090
  namespace: istio-system
spec:
  configPatches:
  - applyTo: NETWORK_FILTER
    match:
      listener:
        filterChain:
          destinationPort: 9090
          filter:
            name: envoy.filters.network.tcp_proxy
        name: virtualInbound
    patch:
      operation: REPLACE
      value:
        name: envoy.filters.network.meta_protocol_proxy
        typed_config:
          '@type': type.googleapis.com/udpa.type.v1.TypedStruct
          type_url: type.googleapis.com/aeraki.meta_protocol_proxy.v1alpha.MetaProtocolProxy
          value:
            applicationProtocol: thrift
            codec:
              name: aeraki.meta_protocol.codec.thrift
            metaProtocolFilters:
            - config:
                '@type': type.googleapis.com/aeraki.meta_protocol_proxy.filters.local_ratelimit.v1alpha.LocalRateLimit
                conditions:
                - match:
                    metadata:
                    - exactMatch: sayHello
                      name: method
                  tokenBucket:
                    fillInterval: 10s
                    maxTokens: 2
                    tokensPerFill: 2
                statPrefix: thrift-sample-server.meta-thrift.svc.cluster.local
                tokenBucket:
                  fillInterval: 60s
                  maxTokens: 5
                  tokensPerFill: 5
              name: aeraki.meta_protocol.filters.local_ratelimit
            - name: aeraki.meta_protocol.filters.router
            routeConfig:
              name: inbound|9090||
              routes:
              - route:
                  cluster: inbound|9090||
            statPrefix: inbound|9090||
  workloadSelector:
    labels:
      app: thrift-sample-server
```

Since we don't have target ports in the service entry, Aeraki is using the service port to generate the inbound envoyfilter(the destinationPort match condition in the envoyfilter), which is just a workaround and won't replace the original TCP filter in case the target port number is not the same as the service port.

The reported issue in Aeraki: https://github.com/aeraki-mesh/aeraki/issues/194

zhlsunshine: > Taking a step back... what is the point of this? What use is a Service without the endpoints?

@howardjohn I added the req.Push since commit: `cf4a022`, so should I step back to `ae8fae` ?

zhlsunshine:
Hi @howardjohn @hzxuzhonghu and @zhaohuabing,  as mentioned above, should we consider to add target port to Istio service's port(the first commit did), but make the target port has the same definition as k8s service `intstr.IntOrString`?

zhaohuabing:
I prefer this way.

howardjohn:
I would prefer not to pollute Istio's core data type (`model.Service`) with a field that is not usable in any context other than Service-over-XDS which is used by 0.01% of users. Maybe we can just work on localized fixes to that code path

howardjohn (changes requested):
I think we are fundamentally trying to solve the wrong problem. A `server` cannot represent a numeric target port, no matter how we try to slice it - it is fundamentally violating the model. This can only be exposed by having a Service _and_ workloads/endpoints/instances of that Service which give the final context.

What we are trying to represent is not aligning with reality of the system, so we end up with these issues.

pilot/pkg/serviceregistry/serviceentry/conversion.go
```diff
@@ -110,13 +110,32 @@ func ServiceToServiceEntry(svc *model.Service, proxy *model.Proxy) *config.Confi
...
-		se.Ports = append(se.Ports, &networking.Port{
-			Number: uint32(p.Port),
-			Name:   p.Name,
-			// Protocol is converted to protocol.Instance - reverse conversion will use the name.
-			Protocol: string(p.Protocol),
-			// TODO: target port
-		})
+		var targetPorts []uint32
+		if len(svc.Attributes.ClusterTargetPorts) > 0 {
+			if tpVal, ok := svc.Attributes.ClusterTargetPorts[p.Name]; ok {
```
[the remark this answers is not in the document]
zhlsunshine: I understand that `convertServices` transforms a Service Entry config to a list of Istio Service, and `ServiceToServiceEntry` do the reverses conversion. And the target port required here is in the field `TargetPort` of Service Entry's Ports, not the target port of field `Endpoints` in Service Entry. 
Please correct me if I am incorrect. @howardjohn & @zhaohuabing

zhlsunshine:
> I think we are fundamentally trying to solve the wrong problem. A `server` cannot represent a numeric target port, no matter how we try to slice it - it is fundamentally violating the model. This can only be exposed by having a Service _and_ workloads/endpoints/instances of that Service which give the final context.
> 
> What we are trying to represent is not aligning with reality of the system, so we end up with these issues.

Sure, I will close this PR if it is a wrong problem. However, I do not understand what is the meaning of `ServiceToServiceEntry` in `pilot/pkg/serviceregistry/serviceentry/conversion.go` if it can not be completely converted. Should I do something to it?

zhaohuabing:
If that's the case, should we change the type of target port in service
entry from number to a string?

PR #38654, opened 2022-04-28
Fixes #31946 Allow external Istiod to watch for local secret resources

### LINKED ISSUE #31946: External istiod should watch for local resource updates

When using an [external control plane configuration](https://preliminary.istio.io/latest/docs/setup/additional-setup/external-controlplane/), the external istiod is only watching for resource updates (including secrets) on the config cluster (i.e., the first remote cluster connected to the external control plane).

This has the undesirable effect that when adding a second cluster to the mesh, we need to apply the API server secret (for istiod to access the new remote cluster) on the first remote cluster, instead of on the external cluster, as [shown here](https://deploy-preview-9280--preliminary-istio.netlify.app/latest/docs/setup/additional-setup/external-controlplane/#register-the-new-cluster). Note that this is different from the first remote cluster, where the secret needed to be applied on the external cluster, to give it access to the config cluster in the first place.

Although this may be appropriate in some use cases, in a "managed control plane" scenario, it would be much better to apply and keep all the secrets in the external cluster, where istiod is running, instead:
```
istioctl x create-remote-secret \
  --context="${CTX_SECOND_CLUSTER}" \
  --type=remote \
  --namespace=external-istiod | \
  kubectl apply -f - --context="${CTX_EXTERNAL_CLUSTER}"
```

To fix this problem, we should allow the secret to be in either location, but having the external istiod watch for resources on both clusters, instead of just the first remote cluster, as it's doing today.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

litong01:
Fixes #31946 Allow Istiod to watch for local secret resources

Currently when Istiod runs as an external control plane, Istiod
only watches for secrets in a namespace of the config cluster. This
fix will also allow Istiod to watch for the secrets created in the
namespace where Istiod is actually running on the external control
plane cluster.

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -622,6 +622,9 @@ var (
 	CanonicalServiceForMeshExternalServiceEntry = env.RegisterBoolVar("LABEL_CANONICAL_SERVICES_FOR_MESH_EXTERNAL_SERVICE_ENTRIES", false,
 		"If enabled, metadata representing canonical services for ServiceEntry resources with a location of mesh_external will be populated"+
 			"in the cluster metadata for those endpoints.").Get()
+
+	LocalClusterSecretWatcher = env.RegisterBoolVar("LOCAL_CLUSTER_SECERT_WATCHER", false,
+		"If enabled, the cluster secret watcher will watch the namespace of the external cluster instead of config cluster").Get()
 )
 
 // EnableEndpointSliceController returns the value of the feature flag and whether it was actually specified.
--- a/pkg/kube/multicluster/secretcontroller.go
+++ b/pkg/kube/multicluster/secretcontroller.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
@@ -91,15 +92,37 @@ type Controller struct {
 
 // NewController returns a new secret controller
 func NewController(kubeclientset kube.Client, namespace string, clusterID cluster.ID) *Controller {
+	informerClient := kubeclientset
+
+	// When these two are set to true, Istiod will be watching the namespace in which
+	// Istiod is running on the external cluster. Use the inCluster credentials to
+	// create a kubeclientset
+	if features.LocalClusterSecretWatcher && features.ExternalIstiod {
+		config, err := rest.InClusterConfig()
+		if err != nil {
+			log.Errorf("Could not get istiod incluster configuration: %v", err)
+			return nil
+		}
+		log.Info("Successfully retrieved incluster config.")
+
+		localKubeClient, err := kube.NewClient(kube.NewClientConfigForRestConfig(config))
+		if err != nil {
+			log.Errorf("Could not create a client to access local cluster API server: %v", err)
+			return nil
+		}
+		log.Infof("Successfully created in cluster kubeclient at %s", localKubeClient.RESTConfig().Host)
+		informerClient = localKubeClient
+	}
+
 	secretsInformer := cache.NewSharedIndexInformer(
 		&cache.ListWatch{
 			ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
 				opts.LabelSelector = MultiClusterSecretLabel + "=true"
-				return kubeclientset.Kube().CoreV1().Secrets(namespace).List(context.TODO(), opts)
+				return informerClient.Kube().CoreV1().Secrets(namespace).List(context.TODO(), opts)
 			},
 			WatchFunc: func(opts metav1.ListOptions) (watch.Interface, error) {
 				opts.LabelSelector = MultiClusterSecretLabel + "=true"
-				return kubeclientset.Kube().CoreV1().Secrets(namespace).Watch(context.TODO(), opts)
+				return informerClient.Kube().CoreV1().Secrets(namespace).Watch(context.TODO(), opts)
 			},
 		},
 		&corev1.Secret{}, 0, cache.Indexers{},
--- a/pkg/test/framework/components/istio/operator.go
+++ b/pkg/test/framework/components/istio/operator.go
@@ -764,12 +764,11 @@ func getTargetClusterListForCluster(targetClusters []cluster.Cluster, c cluster.
 	var outClusters []cluster.Cluster
 	scopes.Framework.Infof("Secret from cluster: %s will be placed in following clusters", c.Name())
 	for _, cc := range targetClusters {
-		if c.IsConfig() && !cc.IsConfig() && !cc.IsRemote() {
-			// if cc is an external cluster, config cluster's secret should have already been
-			// placed on the cluster.
-			continue
-		} else if c.Name() != cc.Name() {
-			// if the given cluster is not the same as the cluster in the target list.
+		// if cc is an external cluster, config cluster's secret should have already been
+		// placed on the cluster, or the given cluster is the same as the cluster in
+		// the target list. Only when c is not config cluster, cc is not external cluster
+		// and the given cluster is not the same as the target, c's secret goes onto cc.
+		if (!c.IsConfig() || !cc.IsExternalControlPlane()) && c.Name() != cc.Name() {
 			scopes.Framework.Infof("Target cluster: %s", cc.Name())
 			outClusters = append(outClusters, cc)
 		}
--- a/pkg/test/framework/components/istio/operator.go
+++ b/pkg/test/framework/components/istio/operator.go
@@ -367,15 +367,29 @@ func deploy(ctx resource.Context, env *kube.Environment, cfg Config) (Instance, 
 		}
 	}
 
+	// Need to determine if there is a setting to watch cluster secret in config cluster
+	// or in external cluster. The flag is named LOCAL_CLUSTER_SECERT_WATCHER and set as
+	// an environment variable for istiod.
+	watchLocalNamespace := false
+	if istioctlConfigFiles.operatorSpec != nil && istioctlConfigFiles.operatorSpec.Values != nil {
+		localClusterSecretWatcher := GetConfigValue("pilot.env.LOCAL_CLUSTER_SECERT_WATCHER",
+			istioctlConfigFiles.operatorSpec.Values.Fields)
+		if localClusterSecretWatcher.GetStringValue() == "true" && i.isExternalControlPlane() {
+			watchLocalNamespace = true
+		}
+	}
+
+	if ctx.Clusters().IsMulticluster() {
+		if err := i.configureDirectAPIServerAccess(ctx, cfg, watchLocalNamespace); err != nil {
+			return nil, err
+		}
+	}
+
 	// Install (non-config) remote clusters.
 	errG = multierror.Group{}
 	for _, c := range ctx.Clusters().Kube().Remotes(ctx.Clusters().Configs()...) {
 		c := c
 		errG.Go(func() error {
-			// Configure API server access for the remote cluster's primary cluster control plane.
-			if err := i.configureDirectAPIServiceAccessBetweenClusters(ctx, cfg, c, c.Config()); err != nil {
-				return fmt.Errorf("failed providing primary cluster access for remote cluster %s: %v", c.Name(), err)
-			}
 			if err := installRemoteCluster(s, i, cfg, c, istioctlConfigFiles.remoteIopFile); err != nil {
 				return fmt.Errorf("failed installing remote cluster %s: %v", c.Name(), err)
 			}
@@ -384,13 +398,6 @@ func deploy(ctx resource.Context, env *kube.Environment, cfg Config) (Instance, 
 	}
 	if errs := errG.Wait(); errs != nil {
 		return nil, fmt.Errorf("%d errors occurred deploying remote clusters: %v", errs.Len(), errs.ErrorOrNil())
-	}
-
-	// For multicluster, configure direct access so each control plane can get endpoints from all API servers.
-	if ctx.Clusters().IsMulticluster() {
-		if err := i.configureDirectAPIServerAccess(ctx, cfg); err != nil {
-			return nil, err
-		}
 	}
 
 	// Configure gateways for remote clusters.
@@ -735,28 +742,8 @@ func waitForIstioReady(ctx resource.Context, c cluster.Cluster, cfg Config) erro
 	return nil
 }
 
-func (i *operatorComponent) configureDirectAPIServerAccess(ctx resource.Context, cfg Config) error {
-	// Configure direct access for each control plane to each APIServer. This allows each control plane to
-	// automatically discover endpoints in remote clusters.
-	for _, c := range ctx.Clusters().Kube() {
-		if err := i.configureDirectAPIServiceAccessForCluster(ctx, cfg, c); err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
-func (i *operatorComponent) configureDirectAPIServiceAccessForCluster(ctx resource.Context, cfg Config,
-	c cluster.Cluster,
-) error {
-	clusters := ctx.Clusters().Configs(c.Config())
-	if len(clusters) == 0 {
-		// giving 0 clusters to ctx.ConfigKube() means using all clusters
-		return nil
-	}
-	return i.configureDirectAPIServiceAccessBetweenClusters(ctx, cfg, c, clusters...)
-}
-
+// configureDirectAPIServiceAccessBetweenClusters - create a remote secret of cluster `c`` and place
+// the secret in all `from` clusters
 func (i *operatorComponent) configureDirectAPIServiceAccessBetweenClusters(ctx resource.Context, cfg Config,
 	c cluster.Cluster, from ...cluster.Cluster,
 ) error {
@@ -769,6 +756,47 @@ func (i *operatorComponent) configureDirectAPIServiceAccessBetweenClusters(ctx r
 		YAML(cfg.SystemNamespace, secret).
 		Apply(apply.NoCleanup); err != nil {
 		return fmt.Errorf("failed applying remote secret to clusters: %v", err)
+	}
+	return nil
+}
+
+func getTargetClusterListForCluster(targetClusters []cluster.Cluster, c cluster.Cluster) []cluster.Cluster {
+	var outClusters []cluster.Cluster
+	scopes.Framework.Infof("Secret from cluster: %s will be placed in following clusters", c.Name())
+	for _, cc := range targetClusters {
+		if c.IsConfig() && !cc.IsConfig() && !cc.IsRemote() {
+			// if cc is an external cluster, config cluster's secret should have already been
+			// placed on the cluster.
+			continue
+		} else if c.Name() != cc.Name() {
+			// if the given cluster is not the same as the cluster in the target list.
+			scopes.Framework.Infof("Target cluster: %s", cc.Name())
+			outClusters = append(outClusters, cc)
+		}
+	}
+	return outClusters
+}
+
+func (i *operatorComponent) configureDirectAPIServerAccess(ctx resource.Context, cfg Config, watchLocalNamespace bool) error {
+	var targetClusters []cluster.Cluster
+	if watchLocalNamespace {
+		// when configured to watch istiod local namespace, secrets go to the external cluster
+		// and primary clusters
+		targetClusters = ctx.AllClusters().Primaries()
+	} else {
+		// when configured to watch istiod config namespace, secrets go to config clusters
+		targetClusters = ctx.AllClusters().Configs()
+	}
+
+	// Now look through entire mesh, create secret for every cluster other than external control plane and
+	// place the secret into the target clusters.
+	for _, c := range ctx.Clusters().Kube().MeshClusters() {
+		theTargetClusters := getTargetClusterListForCluster(targetClusters, c)
+		if len(theTargetClusters) > 0 {
+			if err := i.configureDirectAPIServiceAccessBetweenClusters(ctx, cfg, c, theTargetClusters...); err != nil {
+				return fmt.Errorf("failed providing primary cluster access for remote cluster %s: %v", c.Name(), err)
+			}
+		}
 	}
 	return nil
 }
--- a/tests/integration/pilot/localwatcher/localsecretwatcher_test.go
+++ b/tests/integration/pilot/localwatcher/localsecretwatcher_test.go
@@ -1,0 +1,63 @@
+//go:build integ
+// +build integ
+
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package localwatcher
+
+import (
+	"testing"
+
+	"istio.io/istio/pkg/test/framework"
+	"istio.io/istio/pkg/test/framework/components/echo/common/deployment"
+	"istio.io/istio/pkg/test/framework/components/istio"
+	"istio.io/istio/pkg/test/framework/resource"
+	"istio.io/istio/tests/integration/pilot/common"
+)
+
+var (
+	i istio.Instance
+
+	// Below are various preconfigured echo deployments. Whenever possible, tests should utilize these
+	// to avoid excessive creation/tear down of deployments. In general, a test should only deploy echo if
+	// its doing something unique to that specific test.
+	apps = deployment.SingleNamespaceView{}
+)
+
+func TestMain(m *testing.M) {
+	// nolint: staticcheck
+	framework.
+		NewSuite(m).
+		RequireMinVersion(17).
+		RequireMinClusters(2).
+		Setup(istio.Setup(&i, func(t resource.Context, cfg *istio.Config) {
+			cfg.ControlPlaneValues = `
+values:
+  pilot:
+    env:
+      LOCAL_CLUSTER_SECERT_WATCHER: "true"`
+		})).
+		Setup(deployment.SetupSingleNamespace(&apps, deployment.Config{})).
+		Run()
+}
+
+func TestTraffic(t *testing.T) {
+	framework.
+		NewTest(t).
+		Features("traffic.routing", "traffic.reachability", "traffic.shifting").
+		Run(func(t framework.TestContext) {
+			common.RunAllTrafficTests(t, i, apps)
+		})
+}
--- a/pkg/test/framework/components/istio/util.go
+++ b/pkg/test/framework/components/istio/util.go
@@ -18,11 +18,14 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"regexp"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/hashicorp/go-multierror"
+	"google.golang.org/protobuf/types/known/structpb"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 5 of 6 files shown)


pilot/pkg/bootstrap/server.go
```diff
@@ -1281,3 +1287,60 @@ func (s *Server) initStatusManager(_ *PilotArgs) {
...
+	}
+	log.Info("Successfully retrieved incluster config.")
+
+	localKubeClient, err := kubelib.NewClient(kubelib.NewClientConfigForRestConfig(config))
+	if err != nil {
+		log.Errorf("Could not create a client to access local cluster API server: %v", err)
+		return
+	}
+	log.Infof("Successfully created in cluster kubeclient at %s", localKubeClient.RESTConfig().Host)
+	if s.kubeClient.RESTConfig().Host == localKubeClient.RESTConfig().Host {
```
howardjohn, line 1311: This check makes me feel a little uneasy...

I thought we already have an explicitly flag for "external control plane"?

litong01: @howardjohn thanks so much for reviewing this so quickly. I talked to Frank about this, the external control plane flag alone is not good enough. I will talk to Frank again why we cannot solely rely on that flag. Would like to know if you are ok we are also watching the namespace that istiod is running in so that the secrets do not have to be created in the config cluster to join a new cluster to the mesh? Thanks John.

frankbu: The `ExternalIstiod` flag says `If this is set to true, one Istiod will control remote clusters including CA`, which doesn't necessarily imply an official "external control plane". It could also be a primary cluster that is controlling an istiodless remote in a Primary-Remote multicluster deployment. [This PR](https://github.com/istio/istio/pull/38283) is actually trying to make `ExternalIstiod` the default, so it will almost always be true.

You need to check if the config is in a different cluster, so that you can add a second secret listener for the local cluster in that case.

litong01: That check does exactly that. @frankbu. Not sure if there is any other way for doing this.

pilot/pkg/bootstrap/server.go
```diff
@@ -1281,3 +1287,60 @@ func (s *Server) initStatusManager(_ *PilotArgs) {
...
+	namespace := env.RegisterStringVar("POD_NAMESPACE", "", "").Get()
+	if namespace == "" {
+		log.Error("Could not get istiod namespace.")
+		return
+	}
+	log.Infof("Successfully retrieved Istiod namespace: %s", namespace)
+
+	// create a new secret controller using the same store that the other secret controller
+	// also uses.
+	controller := multicluster.NewControllerWithClusterStore(localKubeClient, namespace,
```
hzxuzhonghu, line 1327: I donot think it is good to introduce another branch to init multi cluster secret controller

pilot/pkg/bootstrap/server.go
```diff
@@ -1281,3 +1287,60 @@ func (s *Server) initStatusManager(_ *PilotArgs) {
...
 }
+
+// initLocalSecretListener creates a new secret listener to listen for
+// secrets which represent new multi cluster participating clusters in a mesh.
+func (s *Server) initLocalSecretListener(args *PilotArgs) {
+	// Check if the istiod's kube client API endpoint is the same as the local cluster API endpoint
+	// If they are the same, then the Istiod is in cluster Istiod, otherwise, it will be an
+	// external istiod running outside of the config cluster, the localKubeClient and the
+	// istiod kube client endpoint should have been different.
+	config, err := rest.InClusterConfig()
```
hzxuzhonghu, line 1298: can you draw a diagram to show your solution?


What is `s.kubeClient` in istiod less?

litong01:
@hzxuzhonghu the whole point of doing this (please see the reference issue) is to listen for secret updates in the namespace where istiod is running. in an external control plane setup, we currently only watch the namespace in the config cluster which is a different cluster from where istiod is running. With this PR, we can watch the namespace for secret in the namespace where istiod is running, this is to allow new cluster secret to be created not in the config cluster. Thanks for reviewing.

pilot/pkg/bootstrap/server.go
frankbu: The question is why does @howardjohn feel uneasy about your kubeclient check? If that's an issue, I suggested a possible alternative way (last sentence in my previous comment).

howardjohn: `Host` (and probably most fields or combination of fields) is not a unique identifier for a cluster. You can have many clusters with the same host, and many hosts for the same cluster. It seems very unreliable.

pilot/pkg/bootstrap/server.go
```diff
@@ -1281,3 +1288,73 @@ func (s *Server) initStatusManager(_ *PilotArgs) {
 		return nil
 	})
 }
+
+// initLocalSecretListener creates a new secret listener to listen for
+// secrets which represent new multi cluster participating clusters in a mesh.
+func (s *Server) initLocalSecretListener(args *PilotArgs) {
+	// Check if this is an instance of Istiod running as external istiod
+	// The logic is based on the following:
+	// 1. command istioctl secret create will always create a secret named
```
frankbu, line 1297: `istioctl create-remote-secret` is just an example way to create the secret. Users can choose to create secrets for remote cluster without using istioctl.

pilot/pkg/bootstrap/server.go
```diff
@@ -1281,3 +1288,73 @@ func (s *Server) initStatusManager(_ *PilotArgs) {
...
+	//    istio-kubeconfig when secret type is config, to indicate that a cluster
+	//    gets used as a config cluster, the secret type must be config
+	// 2. this secret is required to be added to the namespace where istiod will be
+	//    running, otherwise, istiod wont run as an external istiod
+	// 3. this secret has to be mounted to a location to which the env KUBECONFIG points
```
frankbu, line 1302: Can't KUBECONFIG be used for completely different purposes too, so using it doesn't mean it's a config cluster for an external control plane?

pilot/pkg/bootstrap/server.go
```diff
@@ -1281,3 +1288,73 @@ func (s *Server) initStatusManager(_ *PilotArgs) {
...
+	// 4. due to the structure of the secret, this secret will be always mounted
+	//    as a file named config such as /var/run/secrets/remote/config
+	// 5. this kubeconfig is different from loading credentials off of a service account
+	//    which is how non-external istiod gets configured normally.
+	// Based on the above, this method will just check if the file which the value of
+	// KUBECONFIG points exists and if features.ExternalIstiod is true. With these two
```
frankbu, line 1308: features.ExternalIstiod is enabled by default. It controls whether an istiod should try to get the leader lock for remote clusters that it knows about. So it doesn't really mean it's an "external control plane", it could be a primary cluster, in which case it only needs to listen for secrets inCluster.

pilot/pkg/bootstrap/server.go
```diff
@@ -1281,3 +1288,73 @@ func (s *Server) initStatusManager(_ *PilotArgs) {
...
+	// conditions met we should be confident that this is an external Istiod instance.
+
+	// Inspect if the KUBECONFIG pointed location has a file
+	kubeconfigVar := env.StringVar{Var: env.Var{
+		Name: "KUBECONFIG", DefaultValue: "",
+		Description: "", Type: env.STRING,
+	}}
+	if !file.Exists(kubeconfigVar.Get()) {
```
frankbu, line 1316: What does istiod do during startup to decide if it is using an external config cluster or local cluster for config? Same check as this?

frankbu:
@litong01 Please change the PR title to something more descriptive.

pkg/test/framework/components/istio/operator.go
```diff
@@ -739,7 +739,7 @@ func (i *operatorComponent) configureDirectAPIServerAccess(ctx resource.Context,
 
 func (i *operatorComponent) configureDirectAPIServiceAccessForCluster(ctx resource.Context, cfg Config,
 	c cluster.Cluster) error {
-	clusters := ctx.Clusters().Configs(c.Config())
+	clusters := ctx.AllClusters().Primaries(c)
```
frankbu, line 742: I think you need to make this configurable (secretsOnConfig = true/false) to do it one way or the other. I think you also need a similar change where the primary cluster secret is being created (around line 370).

howardjohn:
Reassigned to Steven since I don't think I have enough context to give an adequate  review

hzxuzhonghu:
I think we need a topology figure to help understand, I was reviewing, but could not fully understand. Hard to understand

litong01:
@hzxuzhonghu this pr basically adds a secret controller to watch the namespace in which istiod runs. When a new secret is added to the namespace, the secret controller will filter through these secrets to see if they are secrets which represent a remote cluster. If it is, then cluster added process will be invoked. Then the cluster becomes part of the mesh. It will also handle accordingly when secrets get removed from the namespace. It works exactly same as how external control plane works currently. The only difference is that in an external control plane deployment istiod watches secrets in the namespace which is part of the config cluster. With this pr, istiod will also watch a namespace which is on the external control plane cluster. A admin will be able to add cluster representing secret either in config cluster or external cluster. That is all it does. I'll be very happy to discuss how external control plane works over slack if you like to.

litong01:
@hzxuzhonghu regarding a chart, is there any existing one depicting how external control plane works? If there is one I am happy to modify it to show the difference.

hzxuzhonghu:
[some thing like https://istio.io/latest/docs/setup/install/multicluster/primary-remote/arch.svg](https://istio.io/latest/docs/setup/install/external-controlplane/external-controlplane.svg) 
 Better with kube-apiserver and how istiod interact with it, basically some technique details

litong01:
@hzxuzhonghu do not think these charts really depict how a remote cluster gets added to the mesh but I think a chart describing how currently it works and how the new flow will work may help many who interested in. Thanks for taking time looking into this.

litong01:
@howardjohn @hzxuzhonghu @stevenctl Please see this chart and let me know if it helps on reviewing this PR.
image

hzxuzhonghu:
I donot understand the meaning that config  and multi-cluster secret come from different clusters

hzxuzhonghu:
My doubt is, multi external control planes for differen tenants can resides in same cluster but different namespaces. Then there will be no way to distinguish which cluster should be registered to one istiod.

BTW, one could ask to watch secrets from any cluster maybe different from config cluster and remote clusters too, how should  we tackle that case.

So I do not see it is a general solution.

litong01:
@hzxuzhonghu hmmm, the secret controller only watches the namespace for particular kind of secrets, only secrets representing a cluster will be handled. Adding a secret to the namespace means that cluster wanting to be part of the mesh. Otherwise, you cannot have a multiple cluster environment. Please read relevant document on how external control plane works with config cluster and remote cluster. I understand that the external control plane is a bit difficult to grasp but these document on Istio.io can certainly help.

pilot/pkg/bootstrap/server.go
hzxuzhonghu: Not sure i understand here, to istiod config cluster is a normal cluster. How could `KUBECONFIG` env related

frankbu:
> My doubt is, multi external control planes for differen tenants can resides in same cluster but different namespaces. Then there will be no way to distinguish which cluster should be registered to one istiod.

@hzxuzhonghu The diagram that @litong01 drew, above, is incorrect. The remote secrets to access the api server for (non-config) remote clusters must be in the same namespace in the config cluster as `istiod` is running in on the external cluster (i.e., `istio-system`, not `foo`, in this example). The first remote cluster attached to the external `istiod` is the config cluster, the secret for which would be in `istio-system` namespace of the `external` cluster when `istiod` is started. This causes `istiod` to subsequently use the config cluster, instead of its local cluster, to watch for config (as well as future remote secrets). So, if we want to add a second remote cluster, we currently need to add the secret for it in the `istio-system` namespace of the config cluster.

The purpose of this PR is allow the secret for the second remote cluster to be applied in the `istio-system` namespace of the external cluster instead, so that all the remote secrets can be in the same place (i.e., in the `istio-system` namespace on the external cluster).

hzxuzhonghu:
Yes, I totally understand external istiod model now.  It is great to provide managed mesh for multi tenants. I think listen to local ns/cluster multi-cluster secret make sense.

litong01:
@howardjohn @hzxuzhonghu @stevenctl added a bit more info in the graph for how this actually working and differ from existing implementation. ple
Screen Shot 2022-05-19 at 9 28 18 AM
ase comment.

pilot/pkg/bootstrap/server.go
stevenctl: Istiod doesn't really know the difference. Mainly just the EXTERNAL_ISTIOD feature flag

frankbu: @stevenctl in that case, how do you suggest istiod can decide whether or not it needs to listen to a different cluster for the remote secrets?

stevenctl: A new feature flag that gives the "local" kubeconfig, only to be used for this secret watching purpose? 

If _all_ clusters (including the config cluster) have to be referenced by secret, I think we will have to overhaul our bootstrap code.

stevenctl: I think a lot of the changes made in `bootstrap` should really be handlers for `secretcontroller`.

pkg/kube/multicluster/secretcontroller.go
```diff
@@ -126,13 +131,21 @@ func (c *Controller) AddHandler(h ClusterHandler) {
 	c.handlers = append(c.handlers, h)
 }
 
+func (c *Controller) GetClusterStore() *ClusterStore {
+	return c.cs
+}
+
 // Run starts the controller until it receives a message over stopCh
 func (c *Controller) Run(stopCh <-chan struct{}) error {
 	// run handlers for the local cluster; do not store this *Cluster in the ClusterStore or give it a SyncTimeout
```
stevenctl, line 140: Keep in mind the "local cluster" isn't the cluster housing an external istiod. Instead, it's the "config cluster". External istiod is ~mostly designed as if istiod is running outside _any_ cluster.

hzxuzhonghu: Big +1, `local` already used to indicate the configcluster

hzxuzhonghu: In this pr, the `KUBECONFIG` env is used and has a totally different meaning with `kubeconfig`, that's not align with developer concept.

pilot/pkg/bootstrap/server.go
hzxuzhonghu: My opinion is not tio use `KUBECONFIG` to indicate `local` here

pkg/kube/multicluster/secretcontroller.go
```diff
@@ -91,15 +92,36 @@ type Controller struct {
...
+	informerClient := kubeclientset
+	if features.LocalClusterSecretWatcher && features.ExternalIstiod {
+		config, err := rest.InClusterConfig()
+		if err != nil {
+			log.Errorf("Could not get istiod incluster configuration: %v", err)
+			return nil
+		}
+		log.Info("Successfully retrieved incluster config.")
+
+		localKubeClient, err := kube.NewClient(kube.NewClientConfigForRestConfig(config))
```
hzxuzhonghu, line 107: nit: maybe only init Kubernetes client with k8s.io.client-go

pkg/kube/multicluster/secretcontroller.go
```diff
@@ -91,15 +92,36 @@ type Controller struct {
...
+		if err != nil {
+			log.Errorf("Could not create a client to access local cluster API server: %v", err)
+			return nil
+		}
+		log.Infof("Successfully created in cluster kubeclient at %s", localKubeClient.RESTConfig().Host)
```
hzxuzhonghu, line 112: nit: localKubeClient.RESTConfig().Host is unneccessary

pkg/kube/multicluster/secretcontroller.go
```diff
@@ -130,9 +152,11 @@ func (c *Controller) AddHandler(h ClusterHandler) {
 func (c *Controller) Run(stopCh <-chan struct{}) error {
 	// run handlers for the local cluster; do not store this *Cluster in the ClusterStore or give it a SyncTimeout
 	// this is done outside the goroutine, we should block other Run/startFuncs until this is registered
-	localCluster := &Cluster{Client: c.localClusterClient, ID: c.localClusterID}
-	if err := c.handleAdd(localCluster, stopCh); err != nil {
-		return fmt.Errorf("failed initializing local cluster %s: %v", c.localClusterID, err)
+	if !features.LocalClusterSecretWatcher {
+		localCluster := &Cluster{Client: c.localClusterClient, ID: c.localClusterID}
```
hzxuzhonghu, line 156: have you changed the localClusterClient meaning, if not why it has something to do with `features.LocalClusterSecretWatcher`

pkg/kube/multicluster/secretcontroller.go
```diff
@@ -320,7 +344,7 @@ func (c *Controller) addSecret(name types.NamespacedName, s *corev1.Secret) {
 	}
 
 	for clusterID, kubeConfig := range s.Data {
-		if cluster.ID(clusterID) == c.localClusterID {
+		if cluster.ID(clusterID) == c.localClusterID && !features.LocalClusterSecretWatcher {
```
hzxuzhonghu, line 347: ditto, i think you should remove !features.LocalClusterSecretWatcher

tests/integration/iop-externalistiod-primary-integration-test-defaults.yaml
```diff
@@ -45,6 +45,8 @@ spec:
           value: "istio-istio-system"
         - name: EXTERNAL_ISTIOD
           value: "true"
+        - name: LOCAL_CLUSTER_SECERT_WATCHER
+          value: "true"
```
hzxuzhonghu, line 49: Why no test failure with this?

pkg/kube/multicluster/secretcontroller.go
```diff
@@ -91,15 +92,36 @@ type Controller struct {
 
 // NewController returns a new secret controller
 func NewController(kubeclientset kube.Client, namespace string, localClusterID cluster.ID) *Controller {
+	// When these two are set to true, Istiod will be watching the namespace in which
+	// Istiod is running on the external cluster. Use the inCluster credentials to
+	// create a kubeclientset
+	informerClient := kubeclientset
```
costinm, line 98: Not sure if creating k8s clients all over the place is a good pattern - I would guess at some point we may want to load other things from the 'local cluster'. 

Would it be possible to init the local cluster where we initialize config and remote clusters, and keep it in a field ?

pkg/kube/multicluster/secretcontroller.go
costinm: +1 - if we move this code close to the other k8s init, also use the same pattern/library/deps.

pkg/kube/multicluster/secretcontroller.go
```diff
@@ -91,15 +92,36 @@ type Controller struct {
...
+		log.Infof("Successfully created in cluster kubeclient at %s", localKubeClient.RESTConfig().Host)
+		informerClient = localKubeClient
+	}
+
 	secretsInformer := cache.NewSharedIndexInformer(
 		&cache.ListWatch{
 			ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
 				opts.LabelSelector = MultiClusterSecretLabel + "=true"
-				return kubeclientset.Kube().CoreV1().Secrets(namespace).List(context.TODO(), opts)
+				return informerClient.Kube().CoreV1().Secrets(namespace).List(context.TODO(), opts)
```
costinm, line 120: So we want to stop reading from the config cluster ? I thought you want to read from both, it may break some APIs or apps if we stop reading from the config cluster.

costinm (commented):
Besides initialization of the k8s client, I think we need to consider the requirements of the K8S API WRT Secret loading. 

It is a good change to allow reading secrets from 'local cluster' when different from config cluster.
In the past I remember we also had support from reading from remote clusters as well, and there is a thread about 'federating' so secrets are loaded from external SDS servers. 

I think we will need a bit more structured approach - in the other thread I was suggesting reusing the MeshConfig.ConfigSources, which enables getting configs from multiple sources ( k8s and 
XDS ), with an option on what info to pull from a source. We had this in the old galley - we preserved that for 'XDS federation'.  IMO that would be the right place to init and config.

costinm (commented):
To clarify: using MeshConfg.ConfigSources or similar federation model is 'long term', not asking you do it in this PR. 

For this PR I still think you should have the k8s client code fixed (use same pattern and place to init ). Not sure if reading from both local and config or just local is right, if you have a clear use case to ignore config cluster please add some comments to why ( and what impact it'll have on k8s Gateway secret references )

pkg/kube/multicluster/secretcontroller.go
hzxuzhonghu: It is controlled by the totally new `env`

tests/integration/pilot/localwatcher/localsecretwatcher_test.go
```diff
@@ -0,0 +1,63 @@
...
+	framework.
+		NewSuite(m).
+		RequireMinVersion(17).
+		RequireMinClusters(2).
+		Setup(istio.Setup(&i, func(t resource.Context, cfg *istio.Config) {
+			cfg.ControlPlaneValues = `
+values:
+  pilot:
+    env:
+      LOCAL_CLUSTER_SECERT_WATCHER: "true"`
```
frankbu, line 50: Just turning on LOCAL_CLUSTER_SECRET_WATCHER doesn't test the new feature. You also need to change the test framework to apply the secret locally and if you do that, all the tests that run without your feature turned on will fail

litong01: addressed that in the latest commit. Thanks for reviewing it.

pkg/test/framework/components/istio/operator.go
```diff
@@ -384,13 +398,6 @@ func deploy(ctx resource.Context, env *kube.Environment, cfg Config) (Instance,
 		return nil, fmt.Errorf("%d errors occurred deploying remote clusters: %v", errs.Len(), errs.ErrorOrNil())
 	}
 
-	// For multicluster, configure direct access so each control plane can get endpoints from all API servers.
-	if ctx.Clusters().IsMulticluster() {
-		if err := i.configureDirectAPIServerAccess(ctx, cfg); err != nil {
-			return nil, err
-		}
-	}
-
```
frankbu, line 393: Removing this will break [Multi-Primary](https://istio.io/latest/docs/setup/install/multicluster/multi-primary/)

pkg/test/framework/components/istio/operator.go
```diff
@@ -366,12 +367,31 @@ func deploy(ctx resource.Context, env *kube.Environment, cfg Config) (Instance,
...
 		c := c
 		errG.Go(func() error {
-			// Configure API server access for the remote cluster's primary cluster control plane.
-			if err := i.configureDirectAPIServiceAccessBetweenClusters(ctx, cfg, c, c.Config()); err != nil {
+			// Configure API server access for the remote cluster's primary cluster control plane
+			// which can be either config cluster or external cluster
+			var from []cluster.Cluster
+			if localClusterSecretWatcher.GetStringValue() == "true" {
+				// target clusters will be external cluster
+				from = append(from, ctx.AllClusters().Kube().Primaries(c.Config())...)
```
frankbu, line 389: What is this doing? Why not just c.Primary()?

pkg/test/framework/components/istio/operator.go
litong01: @frankbu multi-primary also share secrets? I did not see that from docs . if they do, what secrets get shared?

pkg/test/framework/components/istio/operator.go
litong01: this to get remote secrets to external (non-config) clusters.

frankbu: c.Primary()

pkg/test/framework/components/istio/operator.go
frankbu: https://istio.io/latest/docs/setup/install/multicluster/multi-primary/#enable-endpoint-discovery

pilot/pkg/features/pilot.go
```diff
@@ -622,6 +622,9 @@ var (
 	CanonicalServiceForMeshExternalServiceEntry = env.RegisterBoolVar("LABEL_CANONICAL_SERVICES_FOR_MESH_EXTERNAL_SERVICE_ENTRIES", false,
 		"If enabled, metadata representing canonical services for ServiceEntry resources with a location of mesh_external will be populated"+
 			"in the cluster metadata for those endpoints.").Get()
+
+	LocalClusterSecretWatcher = env.RegisterBoolVar("LOCAL_CLUSTER_SECERT_WATCHER", false,
+		"If enabled, the cluster secret watcher will watch the namespace in which Istiod runs").Get()
```
frankbu, line 627: This description is confusing. Even if not enabled, the cluster secret watcher will normally "watch the namespace in which Istiod runs". It's only in the external istiod case, that by default, it will watch a different config cluster. Maybe something like this:
```suggestion
		"If enabled, the cluster secret watcher for an external Istiod will watch the namespace in which Istiod runs instead of in the config cluster").Get()
```

pkg/kube/multicluster/secretcontroller.go
```diff
@@ -91,15 +92,36 @@ type Controller struct {
 
 // NewController returns a new secret controller
 func NewController(kubeclientset kube.Client, namespace string, clusterID cluster.ID) *Controller {
+	// When these two are set to true, Istiod will be watching the namespace in which
+	// Istiod is running on the external cluster. Use the inCluster credentials to
+	// create a kubeclientset
+	informerClient := kubeclientset
```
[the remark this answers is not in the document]
litong01: Done as suggested. Thanks.

pkg/test/framework/components/istio/operator.go
```diff
@@ -367,15 +367,29 @@ func deploy(ctx resource.Context, env *kube.Environment, cfg Config) (Instance,
...
+	if istioctlConfigFiles.operatorSpec != nil && istioctlConfigFiles.operatorSpec.Values != nil {
+		localClusterSecretWatcher := GetConfigValue("pilot.env.LOCAL_CLUSTER_SECERT_WATCHER",
+			istioctlConfigFiles.operatorSpec.Values.Fields)
+		if localClusterSecretWatcher.GetStringValue() == "true" && i.isExternalControlPlane() {
+			watchLocalNamespace = true
+		}
+	}
+
+	if ctx.Clusters().IsMulticluster() {
+		if err := i.configureDirectAPIServerAccess(ctx, cfg, watchLocalNamespace); err != nil {
```
frankbu, line 383: This is creating all the secrets before the remote clusters are started which won't guarantee that the correct primary cluster will be associated with each remote cluster. For example, assume we have two primary clusters, p1 and p2, and a remote cluster, r1, which is declared in the topology file to use p1 as its primary. Previously the code used to make sure that p1 controls r1 by:
1. create the r1 secret in p1
2. install r1
3. create the rest of the secrets

Now that you are creating all the secrets at once, both p1 and p2 will have the secret for r1 when it comes up, so they will both compete for control of the cluster, so depending on timing either p1 or p2 might become the primary cluster

pkg/test/framework/components/istio/operator.go
```diff
@@ -773,6 +760,47 @@ func (i *operatorComponent) configureDirectAPIServiceAccessBetweenClusters(ctx r
 	return nil
 }
 
+func getTargetClusterListForCluster(targetClusters []cluster.Cluster, c cluster.Cluster) []cluster.Cluster {
+	var outClusters []cluster.Cluster
+	scopes.Framework.Infof("Secret from cluster: %s will be placed in following clusters", c.Name())
+	for _, cc := range targetClusters {
+		if c.IsConfig() && !cc.IsConfig() && !cc.IsRemote() {
```
frankbu, line 767: Simpler check and also should handle case of more than one external control plane in the topology.

```suggestion
		if c.IsConfig() && cc.IsExternalControlPlane() && c.Primary().Name() == cc.Name() {
```

pkg/test/framework/components/istio/operator.go
litong01: @frankbu as you have said above, p1 and p2 are primary clusters, secret of r1 will need to be placed on both p1 and p2, regardless when it gets on p1 or p2, timing should not determines if p1 or p2 becomes primary cluster, they are primary clusters when they get setup, I do not think I follow your logic above, you already assumed p1 and p2 are primary clusters, why should r1's secret further determines if p1 or p2 becomes primary cluster?

pkg/test/framework/components/istio/operator.go
litong01: @frankbu ok. will change to the suggestion.

litong01: @frankbu Actually the suggestion and the original are not the same. The `else if` check was to make sure it is not adding secret to itself. but the original can be simplified. The point is that c.Primary().Name() does not always equal to c.Name().

litong01: @frankbu please see the new commit. Thanks.

frankbu: > The point is that c.Primary().Name() does not always equal to c.Name().

My suggestion, above, is comparing to `cc.Name()`, not `c.Name()`.

frankbu:
@stevenctl There are still a couple of issues with the secret creation (https://github.com/istio/istio/pull/38654#discussion_r890530438 and https://github.com/istio/istio/pull/38654#discussion_r891201674), but otherwise LGTM. If you are OK with this, please approve. I will take the TODO of sending a followup PR with fixes for the remaining issues

stevenctl:
LGTM

pkg/test/framework/components/istio/operator.go
frankbu: The only way to currently align the bi-directional association (1. remote -> remote-discover-address, 2. primary -> leadership) is by trying to make sure that the declared primary gets the remote secret earlier than other primaries that will compete. It works for the test framework, but I agree it's not reliable since the lock could always get dropped and picked up by another primary. That said, I'm not sure what else we can do for a short term fix.

I am a little confused about why tests don't fail when `remote` has remote discovery address pointing to `primary` but `cross-network-primary` is the leader patching certs in `remote`? I guess it's just because they are compatible certs (same root CA)?

Maybe this is too big a can of worms and we should revert the `feature.ExternalIstiod` to default false and also give up on istiodless defaults for now? Thoughts?

tests/integration/pilot/localwatcher/localsecretwatcher_test.go
howardjohn: Can we make this only run for external istiod? otherwise we are doing a bunch of wasteful testing

litong01: @howardjohn if I figure out a way like 'requiresExternalControlPlane' for the test, that will ease your concerns, right?

howardjohn: yes

frankbu: I think you want the opposite of this one: https://github.com/istio/istio/blob/232fa34a8dcc000b273ec0cdbb9259cb7cd239a7/pkg/test/framework/suite.go#L109

i.e., `RequireExternalControlPlaneTopology`

pkg/test/framework/components/istio/operator.go
frankbu: This problem is fixed in #39406

tests/integration/pilot/localwatcher/localsecretwatcher_test.go
litong01: This has been fixed in this PR https://github.com/istio/istio/pull/39470

Issue #38148, opened 2022-03-28
istioctl analyze: showing invalid error on release 1.13.2


soumyaparui:
### Bug Description

On the same cluster when running same command 
`istioctl analyze -A`

When the istioctl version is 1.13.2 it is showing the following as a error in the result:
`Error [IST0101] (Gateway istio-system/istio-ingressgateway) Referenced selector not found: "istio=ingressgateway"`

But when istioctl version is 1.12.2 it not showing anything as such.

Upon manual verification of the cluster the error looks to invalid. And all traffic are working fine in the cluster.

Upon verification of one of the gateway pod I can see the label:
```
$ kubectl describe pod istio-ingressgateway-abcd -n istio-system
Labels:       
              istio=ingressgateway
```
Also Istioctl validate does not seems to show any issues:
```
kubectl get gw istio-ingressgateway -n istio-system -o yaml | istioctl validate -f –
kubectl get deployments -o yaml -n istio-system | istioctl validate -f –
kubectl get services -o yaml -n istio-system | istioctl validate -f –
kubectl get pods -o yaml -n istio-system | istioctl validate -f –
```

### Version

```prose
$ istioctl version
client version: 1.13.2
control plane version: 1.13.2
data plane version: 1.13.2 (9 proxies)

$ kubectl version --short
Client Version: v1.21.0
Server Version: v1.21.9
```

### Additional Information

Related to:
https://github.com/istio/istio/issues/36160

### Affected product area

- [ ] Docs
- [X] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [X] Upgrade
- [ ] Multi Cluster
- [ ] Virtual Machine
- [ ] Control Plane Revisions

### Is this the right place to submit this?

- [X] This is not a security vulnerability
- [X] This is not a question about how to use Istio

soumyaparui:
Even I am getting invalid error on authorizationpolicy:
Error [IST0101] (AuthorizationPolicy sample-namespace/sample-authorizationpolicy) Referenced namespace not found: "sample-namespace"

zirain:
seems related to https://github.com/istio/istio/issues/38636.

zirain:
looks like a regression from https://github.com/istio/istio/pull/35764. the revision is const to `default`.
https://github.com/istio/istio/pull/38467 do not fix all.

[...]

PR #38659, opened 2022-04-29
istioctl analyze should add kube sources in all revisions

zirain:

fix https://github.com/istio/istio/issues/38148

### DIFF
--- a/pkg/config/analysis/local/istiod_analyze.go
+++ b/pkg/config/analysis/local/istiod_analyze.go
@@ -30,6 +30,7 @@ import (
 
 	"istio.io/api/annotation"
 	"istio.io/api/mesh/v1alpha1"
+	"istio.io/istio/istioctl/pkg/tag"
 	"istio.io/istio/pilot/pkg/config/aggregate"
 	"istio.io/istio/pilot/pkg/config/file"
 	"istio.io/istio/pilot/pkg/config/kube/crdclient"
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
Not sure about this. Should we have a `--revision` flag?

A bit concerned merging all of these:

1. 99.99% of resources are not revision-tagged. So we will just read each resource many times. This may confused analyzers in the worst case, and cost performance in best case
2. Users may have conflicting resources in different revisions. In real world, this is fine -- they are never used together. But now we are merging them together and analyzing, may cause incorrect errors
3. Same as above mostly, but imagine resource 1 depends on res 2. res 1 is in rev A and res 2 in rev B. Before, this would _correctly_ report an error. Now it will incorrectly report no issues.

zirain:
> Not sure about this. Should we have a `--revision` flag?
> 
> A bit concerned merging all of these:
> 
> 1. 99.99% of resources are not revision-tagged. So we will just read each resource many times. This may confused analyzers in the worst case, and cost performance in best case
> 2. Users may have conflicting resources in different revisions. In real world, this is fine -- they are never used together. But now we are merging them together and analyzing, may cause incorrect errors
> 3. Same as above mostly, but imagine resource 1 depends on res 2. res 1 is in rev A and res 2 in rev B. Before, this would _correctly_ report an error. Now it will incorrectly report no issues.

I think `--revision` flag will not work correct with `-A`.
this PR make the logic same as v1.12.

what you said in (3) is reasonable, maybe we should take a look about the whole logic.

arkodg:
> > Not sure about this. Should we have a `--revision` flag?
> > A bit concerned merging all of these:
> > 
> > 1. 99.99% of resources are not revision-tagged. So we will just read each resource many times. This may confused analyzers in the worst case, and cost performance in best case
> > 2. Users may have conflicting resources in different revisions. In real world, this is fine -- they are never used together. But now we are merging them together and analyzing, may cause incorrect errors
> > 3. Same as above mostly, but imagine resource 1 depends on res 2. res 1 is in rev A and res 2 in rev B. Before, this would _correctly_ report an error. Now it will incorrectly report no issues.
> 
> I think `--revision` flag will not work correct with `-A`. this PR make the logic same as v1.12.
> 
> what you said in (3) is reasonable, maybe we should take a look about the whole logic.


Hey @zirain @howardjohn, we have a use case where we have identical resources with different istio revisions in the same cluster. This is confusing `istioctl analyze` which is failing ( with errors such as `Conflict with gateways` & `VirtualService defines the same host` ) . Would it be possible to
* enhance `istioctl analyze` with `--revision` to only analyze resources with a specific revision set

Will create a GH issue, if you agree :) 

We would ideally like to refrain from using `galley.istio.io/analyze-suppress=IST0107,IST0002` annotations of the revisioned resources. TIA !

Issue #38642, opened 2022-04-28
Wrong endpoint when the WorkloadEntry is created before the ServiceEntry


djannot:
### Bug Description

Here are the 2 Istio objects I have created to reproduce the issue:

```
apiVersion: networking.istio.io/v1beta1
kind: WorkloadEntry
metadata:
  name: test
  namespace: default
spec:
  address: 1.2.3.4
  labels:
    foo: bar
  ports:
    http-1234: 5678
```

and

```
apiVersion: networking.istio.io/v1beta1
kind: ServiceEntry
metadata:
  name: test
  namespace: default
spec:
  addresses:
  - 254.154.115.171
  exportTo:
  - istio-gateways
  hosts:
  - test.global
  location: MESH_INTERNAL
  ports:
  - name: http-1234
    number: 1234
    protocol: HTTP
    targetPort: 1234
  resolution: STATIC
  workloadSelector:
    labels:
      foo: bar
```

If I create the WorkloadEntry before the ServiceEntry, I get the following endpoint when running `./istio-*/bin/istioctl --context cluster1 pc endpoints -n istio-gateways deploy/istio-ingressgateway | grep global`:

```
1.2.3.4:1234                     HEALTHY     OK                outbound|1234||test.global
```

This is wrong.

Now, if I create the ServiceEntry before the WorkloadEntry, I get the following endpoint:

```
1.2.3.4:5678                     HEALTHY     OK                outbound|1234||test.global
```

This one is right.

Obviously, the order shouldn't matter, that's why I create this issue.

From my test, it looks like this bug has been introduced in Istio 1.13.

```


### Version

```prose
Istio:
client version: 1.13.3
control plane version: 1.13.3
data plane version: 1.13.3-solo (11 proxies)
Kubernetes
Client Version: v1.21.1
Server Version: v1.21.1
```

djannot:
Also, if I restart istiod, the correct endpoint is created.
But if I restart it again, then the incorrect endpoint is back.
And if I continue to restart it, then the issue is randomly happening again.

hzxuzhonghu:
I will take a look

hzxuzhonghu:
```
  - name: http-1234
    number: 1234
    protocol: HTTP
    targetPort: 1234
```

This writing is not right, the target port should be 5678

linsun:
> ```
>   - name: http-1234
>     number: 1234
>     protocol: HTTP
>     targetPort: 1234
> ```
> 
[...]

PR #38661, opened 2022-04-29
Fix ServiceEntry endpoint port setting order

hzxuzhonghu:

Fix https://github.com/istio/istio/issues/38642

This is to make a determined order consistent with workload handler.

The priority is:

1. when see port name = wle port name. use the wle port.
2. use se target port
3. use se port number.


linsun:
Thanks @hzxuzhonghu, this LGTM.  Question:

I saw in [workload entry doc](https://istio.io/latest/docs/reference/config/networking/workload-entry/) for ports, we have this ```NOTE 2: endpoint port map takes precedence over targetPort.```.  With this PR, I think no change is needed for doc, can you please confirm it?

@howardjohn @ramaraochavali could you review? Thanks.

howardjohn (approved):

PR #38687, opened 2022-04-30
No need to store unconcerned ingresses.

SpecialYang:
In our k8s cluster, there are so many ingress resources. The istio ingress controller running in the cluster only care a little ingress resources whose ingress class is defined as `istio`. We found the ingress controller in istio still store the other ingress resources that has different ingress class, which wastes memory. This pr fix this issue to release memory for these unconcerned ingresses.
The added unit tests will fail without this pr.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/config/kube/ingress/controller.go
+++ b/pilot/pkg/config/kube/ingress/controller.go
@@ -95,6 +95,9 @@ type controller struct {
 	serviceLister   listerv1.ServiceLister
 	// May be nil if ingress class is not supported in the cluster
 	classes v1beta1.IngressClassInformer
+
+	// Only use for test case.
+	eventCompletedCallback func()
 }
 
 // TODO: move to features ( and remove in 1.2 )
@@ -214,8 +217,6 @@ func (c *controller) shouldProcessIngressUpdate(ing *ingress.Ingress) (bool, err
 	// previous processed but should not currently, delete it
 	if preProcessed && !shouldProcess {
 		delete(c.ingresses, item)
-	} else {
-		c.ingresses[item] = ing
 	}
 	c.mutex.Unlock()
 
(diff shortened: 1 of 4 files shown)


howardjohn (changes requested):
Thanks for the PR! The reason this was added was so we can properly cleanup when a Ingress goes from `istio` class to `not-istio` class. Do we handle that somehow?

Also, I would prefer not to impact prod code with test-only functions like `eventCompletedCallback`. Its not clear to me why its really needed anyways?

SpecialYang:
> Thanks for the PR! The reason this was added was so we can properly cleanup when a Ingress goes from `istio` class to `not-istio` class. Do we handle that somehow?

Yes. We only store the matched ingresses. There are two cases.
1. We receive the ingress with `istio` which match against the class defined in meshconfig. We will store this ingress into internal memory. When the class of this ingress from `istio` to other class, we will find this ingress should be delete from the internal memory store. 
2. We receive the ingress with `istio` which don't match against the class defined in meshconfig. We will ignore this ingress. When the class of this ingress from other class to `istio`, we will store this ingress into internal memory.

SpecialYang:
> Also, I would prefer not to impact prod code with test-only functions like `eventCompletedCallback`. Its not clear to me why its really needed anyways?

Just avoid to add time sleep for waiting the event of unmatched ingress to be processed. Have updated.

howardjohn (commented):
if its for optimization, maybe we want to just store the Objectmetadata and ingressclassname? Then it should be pretty small?

SpecialYang:
In our real user case, someone want to migrate other ingress-controller (nginx ingress controller) to istio ingress controller.  The common way is to set ingress class of meshconfig as the same with the exist ingress resource which was watched by the old ingress controller. Now, these ingress resources have been watched by old ingress controller and istio ingress controller. So they can add public ip of istio gateway to dns server with the exist domain to complete the traffic migration. But during migration, these two ingress controllers both alternatively modify the ingress status which will always fire the full push event. 

Hence, the ingress should be stored entirely so that we can compare old ingress spec and new ingress spec to avoid countless pull push event. I will fix this issue in the next pr.

SpecialYang:
Can we move on this pr? @howardjohn @hzxuzhonghu

howardjohn:
TBH I don't get how this helps. We are only storing a pointer to the ingress. Even if we don't store the pointer, the real object still lives in the k8s informer

PR #38692, opened 2022-05-02
allow regex in delegates

ramaraochavali:
This PR allows regex in delegates when there is one regex either at root or delegate and if regex is specified in either delegate or root, no other matching condition can be specified.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/virtualservice.go
+++ b/pilot/pkg/model/virtualservice.go
@@ -483,9 +483,17 @@ func stringMatchConflict(root, leaf *networking.StringMatch) bool {
 	if root == nil || leaf == nil {
 		return false
 	}
-	// regex match is not allowed
-	if root.GetRegex() != "" || leaf.GetRegex() != "" {
-		return true
+	// If root regex match is specified, delegate should not have other matches.
+	if root.GetRegex() != "" {
+		if leaf.GetRegex() != "" || leaf.GetPrefix() != "" || leaf.GetExact() != "" {
+			return true
+		}
+	}
+	// If delgate regex match is specified, root should not have other matches.
+	if leaf.GetRegex() != "" {
+		if root.GetRegex() != "" || root.GetPrefix() != "" || root.GetExact() != "" {
+			return true
+		}
 	}
 	// root is exact match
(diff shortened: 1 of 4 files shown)


ramaraochavali:
> One concern is now we remove obvious validation errors with subtle internal skips.

I can convert them to be warnings if it helps?

ramaraochavali:
seems @hzxuzhonghu is out. @howardjohn  do you mind taking a look and I can address any feedback that @hzxuzhonghu as follow-up if there are no major concerns here - otherwise we can wait till he is back.

howardjohn (approved):

hzxuzhonghu (commented):
LGTM

westwind-sc:
Could we backport this to 1.12?
During our testing, if skip ValidatingWebhook, it is actually working while not match headers in root, but use regex match in leaf headers.
It seems Validating is the only thing stop us using this feature.

hzxuzhonghu:
1.12 is EOL

fedyiv:
Hello @ramaraochavali , @hzxuzhonghu  
Please advise what is the use-case for this feature? I tried to use pattern of common VirtualService and multiple delegate based on prefix, but it does not work if regexp is used. I thought this PR was fixing it but it appeared that it allows using regexp only in case other type of matches are not used.

Here is an example of what I am trying to achieve:

Common VirtualService
```
kind: VirtualService
metadata:
  name: api-gateway
spec:
  ...
  http:
    - name: "ServiceA"
      match:
        - uri:
            prefix: /a
      delegate:
         name: a
````

Delegate Virtual Service

```
kind: VirtualService
metadata:
  name: a
spec:
  http:
    - name: "Route X"
      match:
        - uri:
            regex: /a/b/.*/x
      route:
        - destination:
            host: X
    - name: "Route Y"
      match:
        - uri:
            regex: /a/b/.*/y
      route:
        - destination:
            host: Y
```


So for me it looks unnatural that you can not combine prefix on root VS and regexp on Delegate VS. 
Please advice if it is possible to achieve

PR #38698, opened 2022-05-02
Grpc fixes

costinm:

GRPC c++ client (legitimately) rejects configs, go client seems more tolerant. 

For 'api mode' there is no need to send the address as well. 

Also found that in 'cert auth' mode, the trust aliases are not respected - I can move it to separate PR if needed.


costinm:
It seems grpc c++ needs to be fixed - validation does require address.

PR #38702, opened 2022-05-03
Telemetry: make sure istio-peer-exchange is configured only at terminating tcp servers

vikaschoudhary16:
This is a follow-up PR on https://github.com/istio/istio/pull/38499. In that PR we added `istio-peer-exchange` as the additional alpn in the downstreamTlsContext at the gateways. We ended up doing this alpn change for both the http and tcp servers.
As pointed out by @howardjohn [here](https://github.com/istio/istio/pull/38499#discussion_r862968318), this was a divergence from the inbound listeners at sidecars istio-peer-exchange is added only in the tcp case.
This PR has changes to ensure that inbound listeners are consistent at gateways and sidecars and no new alpn gets added for http servers at gateways.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/gateway.go
+++ b/pilot/pkg/networking/core/v1alpha3/gateway.go
@@ -658,7 +658,7 @@ func buildGatewayListenerTLSContext(
 	}
 
 	server.Tls.CipherSuites = filteredGatewayCipherSuites(server)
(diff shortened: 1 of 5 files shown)


howardjohn (approved):
Would be great to get integration tests for these cases

vikaschoudhary16:
> Would be great to get integration tests for these cases

I am on it John. Thanks a lot!

PR #38703, opened 2022-05-03
Cleanup operator cache on config deletion in revisioned control plane

vikaschoudhary16:
In revisioned control plane, while handling deletion of IstioOperator CR, pruned resources are not cleaned from the controller cache. That means recreation of the same IstioOperator resource does not result in k8s resources because rendered resources are already in controller cache.


jacob-delgado:
cc @litong01 can you take a look at this?

litong01 (approved):
Looks good. Thanks very much for your contributions.

GregHanson (approved):

PR #38713, opened 2022-05-03
Only expand vhost domains for local services

### LINKED ISSUE #38678: Some ServiceEntry hostnames can cause non-deterministic Envoy routes
### Bug Description

If a ServiceEntry `host` has a name that fits the pattern  `*.<namespace>.svc.*`, its hostname will undergo the same expansion as local Kubernetes `Service`s.  This can result in non-deterministic Envoy routes for short non-fully-qualified names.

An example patch to the existing unit tests will illustrate:
```diff --git a/pilot/pkg/networking/core/v1alpha3/httproute_test.go b/pilot/pkg/networking/core/v1alpha3/httproute_test.go
index 9ccd8c8b48..b6d5f65b73 100644
--- a/pilot/pkg/networking/core/v1alpha3/httproute_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute_test.go
@@ -111,6 +111,21 @@ func TestGenerateVirtualHostDomains(t *testing.T) {
                                "echo.default:8123",
                        },
                },
+               {
+                       name: "non-k8s service",
+                       service: &model.Service{
+                               Hostname:     "foo.default.svc.bar.baz",
+                               MeshExternal: false,
+                       },
+                       port: 8123,
+                       node: &model.Proxy{
+                               DNSDomain: "default.svc.cluster.local",
+                       },
+                       want: []string{
+                               "foo.default.svc.bar.baz",
+                               "foo.default.svc.bar.baz:8123",
+                       },
+               },
                {
                        name: "k8s service with default domain and different namespace",
                        service: &model.Service{

```

the result of this test is:
```
    ./github.com/istio/istio/pilot/pkg/networking/core/v1alpha3/httproute_test.go:246: unexpected virtual hosts:
        got  [foo.default.svc.bar.baz foo.default.svc.bar.baz:8123 foo foo:8123 foo.default.svc foo.default.svc:8123 foo.default foo.default:8123]
        want [foo.default.svc.bar.baz foo.default.svc.bar.baz:8123]
```

The problem is that the hostname `foo.default.svc.bar.baz` is being interpreted as the name of a Kubernetes service [here](https://github.com/istio/istio/blob/15d9d454792316c2f8fbda94074c1c584dfc4ea9/pilot/pkg/networking/core/v1alpha3/httproute.go#L537-L539), and therefore alternate service names are being generated for it.   

If there are any other `ServiceEntries` or Kubernetes `Service`s with a name of the form `foo.default.svc.*`, then the expanded alternate/short names will conflict and be de-duplicated.  The problem is that this is non-deterministic, so _one_ of them will win.  The net result is that upon re-starting `pilot`, requests to `http://foo` might end up routing to _any_ Service or ServiceEntry that matches.  

Inspecting Envoy's routes across re-starts can result in differences like:
Screen Shot 2022-04-28 at 2 24 56 PM

It looks like this bug was introduced as of 1.13

### Version

```prose
client version: 1.13.2
control plane version: 1.13
data plane version: 1.13.2 (2 proxies), 1.13.2 (3 proxies)
```

### Additional Information

_No response_

### Affected product area

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Upgrade
- [ ] Multi Cluster
- [ ] Virtual Machine
- [ ] Control Plane Revisions

birkland:

When building outbound sidecar config, generate an expanded set of hostnames (e.g. fqdn + short hostnames) only for local Kubernetes services in a given cluster.

resolves #38678

### DIFF
--- a/releasenotes/notes/38678.yaml
+++ b/releasenotes/notes/38678.yaml
@@ -2,7 +2,7 @@ apiVersion: release-notes/v2
 kind: bug-fix
 area: traffic-management
 issue:
-  - 38713
+  - 38678
 releaseNotes:
   - |
-    **Fixed** Some ServiceEntry hostnames can cause non-deterministic Envoy routes
+    **Fixed** some ServiceEntry hostnames can cause non-deterministic Envoy routes.


pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -532,8 +532,8 @@ func generateVirtualHostDomains(service *model.Service, port int, node *model.Pr
 //
 // - Given foo.local.campus.net on proxy domain "" or proxy domain example.com, this
 // function returns nil
-func GenerateAltVirtualHosts(hostname string, port int, proxyDomain string) []string {
-	if strings.Contains(proxyDomain, ".svc.") {
+func GenerateAltVirtualHosts(hostname string, port int, proxyDomain string, serviceIsLocal bool) []string {
+	if strings.Contains(proxyDomain, ".svc.") && serviceIsLocal {
```
howardjohn, line 536: I think it should be based on the suffix matching with DNSDomain? Especially if this was working on 1.12 and older, which used DNSDomain

birkland: Yeah, the resulting code when using DNSDomain suffix seems more explicit as to when hostnames are expanded or not.  It was easier to capture the nuance in comments at least!

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -729,3 +739,12 @@ func buildCatchAllVirtualHost(node *model.Proxy) *route.VirtualHost {
 		IncludeRequestAttemptCount: true,
 	}
 }
+
+// Simply removes everything before .svc, if present
+func removeSvcNamespace(domain string) string {
+	if idx := strings.Index(domain, ".svc."); idx > 0 {
+		return domain[idx:]
```
ramaraochavali, line 746: nit: I think this is redundant check for the context in which this is being used.

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -533,8 +533,18 @@ func generateVirtualHostDomains(service *model.Service, port int, node *model.Pr
...
+	// are indeed kubernetes services
 	if strings.Contains(proxyDomain, ".svc.") {
-		return generateAltVirtualHostsForKubernetesService(hostname, port, proxyDomain)
+
+		if strings.HasSuffix(hostname, removeSvcNamespace(proxyDomain)) {
+			return generateAltVirtualHostsForKubernetesService(hostname, port, proxyDomain)
+		}
+
+		// Hostname is not a kube service.  Since it contains a `.svc.`, it is not safe to expand the
+		// hostname as it could conflict with other kube services.
```
ramaraochavali, line 546: Is this comment correct? How do we know if the host has ".svc"?

birkland: Actually, it wasn't!  Updated.

pilot/pkg/networking/core/v1alpha3/httproute.go
birkland: It's actually possible for a hostname here not to have `.svc` in it (in which case, just returning the domain unchanged is the right thing to do).  I fixed the comments to make that more clear.   Sorry!

birkland: Ha, yes.  Ignore my original response - I had confused domain with hostname in my comment.  It is true that `domain` will always have `.svc`. As it turns out, though, despite the check not being necessary, the `gocritic` check/bot took issue with the fact that result could be `-1`.  So I think I have to keep the check in if I use `strings.Index` here at all?

linsun:
LGTM, nice work @birkland !!

releasenotes/notes/38713.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+issue:
+  - 38713
+releaseNotes:
+  - |
+    **Fixed** Some ServiceEntry hostnames can cause non-deterministic Envoy routes
```
ericvn, line 8: Nit. lower case some and add a period at the end.

releasenotes/notes/38713.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+issue:
+  - 38713
```
ericvn, line 5: Also point at the issue, https://github.com/istio/istio/issues/38678.

birkland: good catch, file re-named as well to match the issue number

ericvn (approved):
Release Note looks good

linsun (approved):

linsun:
@ramaraochavali friendly ping - can you please review?

ramaraochavali (approved):

PR #38722, opened 2022-05-04
WIP: auto_san_validation support

kfaseela:

RFC: [auto san validation support](https://docs.google.com/document/d/1pTUl-Ng3nXAWJb7UGJtalftznpxQEfID/)

- [ ] Docs
- [ ] Installation
- [X ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1143,15 +1142,24 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
...
+	if mc != nil && features.EnableAutoSni {
+		if len(tls.Sni) == 0 || (features.VerifyCertAtClient && len(tls.SubjectAltNames) == 0) {
+			if mc.httpProtocolOptions == nil {
+				mc.httpProtocolOptions = &http.HttpProtocolOptions{}
+			}
+			if mc.httpProtocolOptions.UpstreamHttpProtocolOptions == nil {
+				mc.httpProtocolOptions.UpstreamHttpProtocolOptions = &core.UpstreamHttpProtocolOptions{}
+			}
+			if len(tls.Sni) == 0 {
+				mc.httpProtocolOptions.UpstreamHttpProtocolOptions.AutoSni = true
```
howardjohn, line 1157: Don't we still want this even if VerifyCertAtClient is false?

kfaseela: Yes, line 1149 is an OR condition so that we don't unnecessarily set empty attrs..did I miss something?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1143,15 +1142,24 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
...
-		if mc.httpProtocolOptions == nil {
-			mc.httpProtocolOptions = &http.HttpProtocolOptions{}
-		}
-		if mc.httpProtocolOptions.UpstreamHttpProtocolOptions == nil {
-			mc.httpProtocolOptions.UpstreamHttpProtocolOptions = &core.UpstreamHttpProtocolOptions{}
+// Set auto_sni if sni field is not explicitly set and if EnableAutoSni feature flag is enabled.
+// Set auto_san_validation if VerifyCertAtClient feature flag is enabled
+func (cb *ClusterBuilder) setAutoSniAndAutoSanValidation(mc *MutableCluster, tls *networking.ClientTLSSettings) {
+	if mc != nil && features.EnableAutoSni {
+		if len(tls.Sni) == 0 || (features.VerifyCertAtClient && len(tls.SubjectAltNames) == 0) {
```
vikaschoudhary16, line 1149: should not we return with some warning message if EnableAutoSni is true and `len(tls.Sni) != 0`?

vikaschoudhary16, line 1149: Why are we coupling `AutoSanValidation` with `VerifyCertAtClient`? both seems unrelated to me. 
IMO, for setting AutoSanValidation, we should just look for EnableAutoSni(may be rename it to EnableAutoSniAndSANValidation) and `len(tls.SubjectAltNames) == 0)`.

vikaschoudhary16: consider the case, mode MUTUAL with user configured ca certificates. In this case VerifyCertAtClient could be false because client side does not need system ca for server cert verification, but AutoSanValidation can still be desired for auto-sni

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1143,15 +1142,24 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
...
-func (cb *ClusterBuilder) setAutoSni(mc *MutableCluster, tls *networking.ClientTLSSettings) {
-	if mc != nil && features.EnableAutoSni && len(tls.Sni) == 0 {
-		if mc.httpProtocolOptions == nil {
-			mc.httpProtocolOptions = &http.HttpProtocolOptions{}
-		}
-		if mc.httpProtocolOptions.UpstreamHttpProtocolOptions == nil {
-			mc.httpProtocolOptions.UpstreamHttpProtocolOptions = &core.UpstreamHttpProtocolOptions{}
+// Set auto_sni if sni field is not explicitly set and if EnableAutoSni feature flag is enabled.
+// Set auto_san_validation if VerifyCertAtClient feature flag is enabled
+func (cb *ClusterBuilder) setAutoSniAndAutoSanValidation(mc *MutableCluster, tls *networking.ClientTLSSettings) {
```
hzxuzhonghu, line 1147: This function is too complex, need to simplify

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1143,15 +1142,24 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
...
+			if mc.httpProtocolOptions == nil {
+				mc.httpProtocolOptions = &http.HttpProtocolOptions{}
+			}
+			if mc.httpProtocolOptions.UpstreamHttpProtocolOptions == nil {
+				mc.httpProtocolOptions.UpstreamHttpProtocolOptions = &core.UpstreamHttpProtocolOptions{}
+			}
+			if len(tls.Sni) == 0 {
+				mc.httpProtocolOptions.UpstreamHttpProtocolOptions.AutoSni = true
+			}
+			if features.VerifyCertAtClient && len(tls.SubjectAltNames) == 0 {
```
hobbytp, line 1159: I think we don't need to check tls.SubjectAltNames, because if the TLS layer SAN checking fails, the connection can not setup , so it will not impact to L7 SAN checking. And if this TLS layer SAN checking passed, it also does not impact L7 checking. But if you would like to only keep one check, you can do that.  The pseudo-code may look like this:

```go
func setAutoSniAndAutoSANValidation(){
	if len(tls.Sni) != 0 || !features.EnableAutoSni
	    return
		
	if mc.httpProtocolOptions == nil {
		mc.httpProtocolOptions = &http.HttpProtocolOptions{}
	}
	if mc.httpProtocolOptions.UpstreamHttpProtocolOptions == nil {
		mc.httpProtocolOptions.UpstreamHttpProtocolOptions = &core.UpstreamHttpProtocolOptions{}
	}
        mc.httpProtocolOptions.UpstreamHttpProtocolOptions.AutoSni = true
	
	if features.VerifyCertAtClient //&& len(tls.SubjectAltNames) == 0
	    mc.httpProtocolOptions.UpstreamHttpProtocolOptions.AutoSanValidation = true
}
```
BTW, I also think AutoSanValidation is not related to VerifyCertAtClient.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
howardjohn: SAN validation doesn't even work without VerifyCertAtClient unless you are extremely careful as far as I understand. 99% of users would end up not actually validating the SAN.

VerifyCertAtClient=false is a horrible default only for backwards compatibility

vikaschoudhary16: >SAN validation doesn't even work without VerifyCertAtClient unless you are extremely careful 

careful means configuring user ca certs in the DR correctly? 

>VerifyCertAtClient=false is a horrible default only for backwards compatibility

yeah, enabling it by default will mean SAN is validated using system ca certs if user has not provided ca certs and that seems better ok.

If we want to go ahead with relying on VerifyCertAtClient for AutoSanValidation, does it make sense to add validation logic to verify VerifyCertAtClient is set to true if EnableAutoSni is set to true?

kfaseela: @vikaschoudhary16 : We wanted to keep EnableAutoSni separate, as that is not tightly coupled with any validation. That way it gives flexibility to use only auto_sni part. However AutoSanValidation is turned on, only if VerifyCertAtClient is enabled, and that looks better?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
kfaseela: Could you please check if it looks better now?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
kfaseela: DR allows setting the sni explicitly, in that case we donot want to use auto_sni, and that is an allowed usecase. So I am not sure if we need a warning message in that case.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1143,15 +1142,35 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
...
 			mc.httpProtocolOptions = &http.HttpProtocolOptions{}
 		}
 		if mc.httpProtocolOptions.UpstreamHttpProtocolOptions == nil {
 			mc.httpProtocolOptions.UpstreamHttpProtocolOptions = &core.UpstreamHttpProtocolOptions{}
 		}
-		mc.httpProtocolOptions.UpstreamHttpProtocolOptions.AutoSni = true
+		if setAutoSni {
+			mc.httpProtocolOptions.UpstreamHttpProtocolOptions.AutoSni = true
+		}
+		if setAutoSanValidation {
```
hzxuzhonghu, line 1171: the auto san validation should be set with auto sni

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
hobbytp: > 
auto_san_validation only enabled when auto_sni is enabled. or I miss anything here?
refer [here](https://www.envoyproxy.io/docs/envoy/v1.22.0/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-upstreamhttpprotocoloptions-auto-san-validation)

hzxuzhonghu: Auto sni or auto san validation can be be set only when they are not set in `UpstreamTlsContext`, otherwise it will override.

So here must check tls. SubjectAltNames and SNI

kfaseela: @hobbytp envoy documentation says that auto_san_validation is "intended" to be set along with auto_sni as sni is needed for the validation. However since in istio we have a way to set SNI even otherwise through DR, I have just coded in such a way that if either auto_sni or explicit sni is present, we can use auto_san_validation, and that seem to work as well. Let me know what you think

hobbytp: Consider one case:

In DestinationRule
    host: *.example.com
    SANs: [a.example.com, *.example.com]

Client (with sidecar) sends HTTP with host header= b.example.com
1) if server responses cert with SANs: [a.example.com, b.example.com, *.example.com] ==> L4 SAN check passes, and L7 SAN validation pass.
2) if server responses cert SANs: [a.example.com, *.example.com] ==> will L4 check pass? If yes, L7 SAN validation will fail for it can not find b.example.com in SANs.
This means: len(tls.SubjectAltNames) != 0, but it does not cover L7 logic checking.

So my question is why should we couple L4 SAN validation and L7 SAN validation even though they both are about tls handshake.

howardjohn: The SAN check is not an `AND` operation - its an `OR`.

You should not be able to set explicitly SANs   and use auto san

hobbytp: But I think we can, at least from the envoy proxy perspective, because one is for L4, one is for L7, and their configurations are separated, right?
But maybe I got your point, you want to disable the auto_san_validation (in L7) when L4 configured SAN list in DR from istio perspective? Just like you said "You should not be able to set explicitly SANs and use auto san", right?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
kfaseela: @hzxuzhonghu Please take a look at my [comment](https://github.com/istio/istio/pull/38722#discussion_r870230964)

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
hobbytp: > 
@kfaseela  auto_san_validation is "intended" to be set with auto_sni field WHEN [override_auto_sni_header](https://www.envoyproxy.io/docs/envoy/v1.22.0/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-upstreamhttpprotocoloptions-override-auto-sni-header) is unset. When override_auto_sni_header is set, auto_san_validation will compare the SAN value(s) from the upstream certificate with the override_auto_sni_header. The SNI value in override_auto_sni_header will override the original HTTP header host/authority value. 
The auto_san_validation is only used to compare the SAN from upstream certificate with the SNI value from client HTTP host/authority header or configured override_auto_sni_header.

hzxuzhonghu: There is no L4  & L7, SAN check is during tls handshake. 

According to envoy document, auto_sni and auto_san_validation will override that specified in TransportSocketMatch, which is set from DR's TLS setting.

So here we definitely not to override user explicitly set ones.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
hzxuzhonghu: If only sni specified in DR, but not san. And then set auto_san_validation here, which san should it check>?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
hobbytp: yes, I also found that now, you are right. 
So the left questions are which SNI, SAN we want to take high priority. DR's or auto.
seems both you and John's comments prefer DR's to be the high priority, right?

hzxuzhonghu: Yep, we want to and must respect user explicitly setting, this is the convention of declaring API  Semantics.

hobbytp: That sounds reasonable.

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
kfaseela: envoy honors auto_sni and auto_san_validation config independent of each other: https://github.com/envoyproxy/envoy/blob/main/source/common/router/router.cc#L516

hzxuzhonghu: I am wondering how should we disable san check when `features.VerifyCertAtClient` = true, the env indicates checking ca bundle only.

kfaseela: https://istio.io/latest/docs/reference/config/networking/destination-rule/#ClientTLSSettings talks about insecureSkipVerify attribute @hzxuzhonghu

hzxuzhonghu: This is two different case, i mean how to only verify ca.

kfaseela: @howardjohn does it make sense to have a separate feature flag only for auto_san_validation? As per the envoy code mentioned above, both can be set independently.

kfaseela: Thanks @howardjohn . In that case, probably the current PR is good enough, and you can do one more round of review?

hobbytp (approved):

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
hzxuzhonghu: make sense.

howardjohn (approved):

PR #38724, opened 2022-05-04
Set ENABLE_LEGACY_FSGROUP_INJECTION to false by default

kfaseela:

Since k8s 1.18 is outside of the official istio support window,
changing the default value of the feature flag, though the changes
to automatically detect the version and set the flag are retained.

### DIFF
--- a/operator/pkg/manifest/shared.go
+++ b/operator/pkg/manifest/shared.go
@@ -430,8 +430,8 @@ func getClusterSpecificValues(client kube.Client, force bool, l clog.Logger) (st
 }
 
 func getFSGroupOverlay(config kube.Client) string {
-	if kube.IsAtLeastVersion(config, 19) {
-		return "values.pilot.env.ENABLE_LEGACY_FSGROUP_INJECTION=false"
+	if kube.IsLessThanVersion(config, 19) {
+		return "values.pilot.env.ENABLE_LEGACY_FSGROUP_INJECTION=true"
 	}
 	return ""
 }
--- a/operator/pkg/manifest/shared.go
+++ b/operator/pkg/manifest/shared.go
@@ -430,7 +430,12 @@ func getClusterSpecificValues(client kube.Client, force bool, l clog.Logger) (st
 }
 
 func getFSGroupOverlay(config kube.Client) string {
-	if kube.IsLessThanVersion(config, 19) {
+	// ENABLE_LEGACY_FSGROUP_INJECTION has first-party-jwt as allowed because we only
+	// need the fsGroup configuration for the projected service account volume mount,
+	// which is only used by first-party-jwt. The installer will automatically
+	// configure this on Kubernetes 1.18 or older.
+	jwtPolicy, _ := util.DetectSupportedJWTPolicy(config)
+	if kube.IsLessThanVersion(config, 19) && jwtPolicy != util.FirstPartyJWT {
 		return "values.pilot.env.ENABLE_LEGACY_FSGROUP_INJECTION=true"
 	}
 	return ""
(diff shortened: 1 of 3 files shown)


operator/pkg/manifest/shared.go
```diff
@@ -430,8 +430,8 @@ func getClusterSpecificValues(client kube.Client, force bool, l clog.Logger) (st
 }
 
 func getFSGroupOverlay(config kube.Client) string {
-	if kube.IsAtLeastVersion(config, 19) {
-		return "values.pilot.env.ENABLE_LEGACY_FSGROUP_INJECTION=false"
+	if kube.IsLessThanVersion(config, 19) {
```
howardjohn, line 433: should we add the JwtPolicy check maybe?

pilot/pkg/features/pilot.go
```diff
@@ -448,7 +448,7 @@ var (
 	// which is only used by first-party-jwt. The installer will automatically
 	// configure this on Kubernetes 1.19+.
 	// Note: while this appears unused in the go code, this sets a default which is used in the injection template.
-	EnableLegacyFSGroupInjection = env.RegisterBoolVar("ENABLE_LEGACY_FSGROUP_INJECTION", JwtPolicy != jwt.PolicyFirstParty,
+	EnableLegacyFSGroupInjection = env.RegisterBoolVar("ENABLE_LEGACY_FSGROUP_INJECTION", false,
```
howardjohn, line 451: nit: comment above is outdated

releasenotes/notes/disable-fs-group-injection.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: feature
+area: installation
+releaseNotes:
+- |
+  **Updated** default value of the feature flag `ENABLE_LEGACY_FSGROUP_INJECTION` to false
```
howardjohn, line 6:
```suggestion
  **Updated** default value of the feature flag `ENABLE_LEGACY_FSGROUP_INJECTION` to false; this may cause issues with sidecars when installing on Helm on Kubernetes versions prior to 1.19.
```

howardjohn (commented):
lgtm overall

pilot/pkg/features/pilot.go
kfaseela: Moved this to the newly added jwt check in shared.go

operator/pkg/manifest/shared.go
```diff
@@ -430,8 +430,12 @@ func getClusterSpecificValues(client kube.Client, force bool, l clog.Logger) (st
 }
 
 func getFSGroupOverlay(config kube.Client) string {
-	if kube.IsAtLeastVersion(config, 19) {
-		return "values.pilot.env.ENABLE_LEGACY_FSGROUP_INJECTION=false"
+	// Set ENABLE_LEGACY_FSGROUP_INJECTION to true only for Kubernetes 1.18 or older,
+	// together with third-party-jwt, as we need the fsGroup configuration for the projected
+	// service account volume mount, which is only used by third-party-jwt.
+	jwtPolicy, _ := util.DetectSupportedJWTPolicy(config)
```
howardjohn, line 436: Ah, this is actually a bit tricky.. this is only looking at the *detected* value - but users could override it. I think we may want to move `getFSGroupOverlay` to be after `getJwtTypeOverlay`, and use the result of `getJwtTypeOverlay` here?

operator/pkg/manifest/shared.go
kfaseela: Done

releasenotes/notes/disable-fs-group-injection.yaml
kfaseela: Done

operator/pkg/manifest/shared.go
kfaseela: Done

howardjohn (approved):

PR #38738, opened 2022-05-04
Cleanup istiodless multicluster and external istiod testing

frankbu:

This PR cleans up the istiodless and external istiod test framework and makes istiodless testing by default.

Followup to #38283

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
@@ -83,7 +83,7 @@ data:
             "cpu": "10m"
           }
         },
-        "externalIstiod": false,
+        "externalIstiod": true,
         "hub": "gcr.io/istio-testing",
         "imagePullPolicy": "",
         "imagePullSecrets": [],
@@ -1403,6 +1403,15 @@ rules:
   - apiGroups: ["authorization.k8s.io"]
     resources: ["subjectaccessreviews"]
     verbs: ["create"]
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["create", "get", "list", "watch", "update"]
+  - apiGroups: ["admissionregistration.k8s.io"]
+    resources: ["mutatingwebhookconfigurations"]
+    verbs: ["get", "list", "watch", "update", "patch"]
+  - apiGroups: ["admissionregistration.k8s.io"]
+    resources: ["validatingwebhookconfigurations"]
+    verbs: ["get", "list", "watch", "update"]
 ---
 # Source: istiod/templates/clusterrolebinding.yaml
 apiVersion: rbac.authorization.k8s.io/v1
--- a/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
@@ -83,7 +83,7 @@ data:
             "cpu": "10m"
           }
         },
-        "externalIstiod": true,
+        "externalIstiod": false,
         "hub": "gcr.io/istio-testing",
         "imagePullPolicy": "",
         "imagePullSecrets": [],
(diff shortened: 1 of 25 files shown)


litong01:
@frankbu I am actually having issues with this particular variable. But after reading some of the docs, the doc says this method is needed when the host name is actually an IP address such as an load balancer IP, then the cert will need to be recreated to include the IP address in SAN especially in a test env. are we not needing that any more? I have been messing with this parameter last few days. Can you confirm we indeed not need this? In quite few tests (I think at least 4 integration tests), we are patching the Istiod so that the Istiod also restarts. Please confirm.

litong01:
@frankbu are you using ISTIOCTL_XDS_ADDRESS to basically replace the ISTIOD_CUSTOM_HOST when generating cert?

manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
```diff
@@ -1403,6 +1403,15 @@ rules:
   - apiGroups: ["authorization.k8s.io"]
     resources: ["subjectaccessreviews"]
     verbs: ["create"]
+  - apiGroups: [""]
```
howardjohn, line 1406: is there any way to turn this off? This effectively gives root admin access to the cluster - a scary proposition when the role is called "reader". 

These permissions are only needed when following worst-practices - real users should never need these (public certs).

frankbu: TBH, I haven't really looked at what permissions are really necessary, maybe these aren't all needed. I think what is needed is enough permission for any istiod to aquire the leader lock on another cluster.

stevenctl: We need to make sure users understand how to do this, in relation to the istiodless remote cluster install docs. This means linking docs for setting up a proper CA (or file mounted certs) on the istiod side, and if they're not public certs tell them they somehow need to "install the root of trust on their system" on the remote cluster side. 

For testing, I think we can get a step closer to this. In the integ-suite-kind/bash layer we could generate certs and install them on the KinD nodes. We then pass flags to the Go part of the framework pointing to the certs. The istio-install logic can pick up on this to mount them and tell istiod to use them, similar to these tests: https://github.com/istio/istio/blob/f85509e8a90f4e7970524728c2fc99ebda7f11ed/tests/integration/security/file_mounted_certs/main_test.go#L140

howardjohn: I think mounting into kind node is not sufficient as the pods also need the cert for XDS and they do not inherit it from the node?

frankbu: @stevenctl @howardjohn I turned these extra permissions back off by default. I also realized that they aren't really needed for enabling istiodless remotes anyway, because they are just allowing external istiods to control other primary clusters. We don't want external istiods to control other primaries in the primary-primary MC setup. In fact, that's why I fixed the leadership election to make sure the local istiod wins the election. Not having these permissions short circuits that election process, so it's fine. We only need to turn on these extra permissions if we want HA control across primary clusters, but that's out of scope for this PR, which is just trying to get istiod out of remote clusters in the currently documented MC setups.

stevenctl (approved):

manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
stevenctl: My point is users probably won't do the right thing unless we document it. If we document a version where we give these broad permissions to patch the ca bundle into remote clusters, people will just do that. 

If we tell users "do things properly" without steps, they'll report bugs here saying primary-remote is broken. 

But, whatever we document we should try to replicate in our testing.

frankbu: I see, you're concerned about doing this in remotes too. The current pattern is now the same for all remote clusters, whether controlled by a primary istiod or an external one. If you feel that there's a better approach that we should document (for both I assume), we do have 3 months to figure it out because this is not going into 1.14 and will only be in the preliminary docs until 1.15. I'm happy to help work on another ("do things properly") pattern, but will definitely need help to understand what that should be.

PR #38750, opened 2022-05-05
pilot: fix EnovyFilter inline patch removal

icereval:
Found a bug where HTTP and Network Filters are not removed when a patch first creates a filter, and subsequently another patch removes the filter.

Use Case: Adding a namespace filter followed by a subsequent filter excluding the first with a workloadSelector label.

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu (approved):
Though this usage is not common, the fix is good

GregHanson (approved):

PR #38756, opened 2022-05-05
Fix name used in EnqueueForParentHandler

howardjohn:
Currently we use the object name, but supposed to be parent name

This doesn't matter in all current usages as they are always the same,
but for some in-development stuff it is important to use the right one.

### DIFF
--- a/pkg/kube/controllers/common_test.go
+++ b/pkg/kube/controllers/common_test.go
@@ -1,0 +1,83 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


pkg/kube/controllers/common_test.go
```diff
@@ -0,0 +1,83 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"istio.io/istio/pkg/config/schema/gvk"
+	"istio.io/istio/pkg/test"
+	"istio.io/istio/pkg/test/util/retry"
+)
+
+func TestEnqueueForParentHandler(t *testing.T) {
```
nmittler, line 31: @howardjohn just to be clear, in asking for the test I was thinking to test an actual call site and to make sure that the resulting behavior is expected. Does this do that?

howardjohn: Yes, its there on L39. And it fails without the code change in the PR

nmittler (approved):

Issue #38704, opened 2022-05-03
Eastwest gateway incorrectly applies DR


nmittler:
### Bug Description

If a mTLS for a destination is enabled via a DR in a namespace other than `istio-system`, the eastwest gateway will never get endpoints for the destination service.

This is due to the fact that the eastwest gateway still applies [endpoint filters](https://github.com/istio/istio/blob/6ec2a64c0b89f323cfaeb655b3a69ea196faf023/pilot/pkg/xds/eds.go#L322), which use the `mtlsChecker`. The mTLSChecker will have no DR for the service (since it was not applied to the `istio-system` namespace) and therefore assume the default non-mTLS setting, thus filtering out the endpoint from the east-west gateway.

Judging by the comment in that code, it appears this was intentional. However, I'm not sure that this particular case was considered or tested. [Other parts of the logic](https://github.com/istio/istio/blob/d8c3d6ed24f51ac87e554e3d37d24ebc9bf5b742/pilot/pkg/networking/core/v1alpha3/cluster_builder.go#L786) for eastwest gateway ignore DR, and I suspect this should as well.

### Version

```prose
head
```

### Affected product area

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [ ] Upgrade
- [ ] Multi Cluster
- [ ] Virtual Machine
- [ ] Control Plane Revisions

### Is this the right place to submit this?

- [X] This is not a security vulnerability
- [X] This is not a question about how to use Istio

nmittler:
@howardjohn do you know if there is anything that can be done here? I know that this logic was added to deal with a vul.

howardjohn:
@stevenctl I think the DR part specifically is actually not for the CVE, but rather to avoid accidentally sending non-mTLS traffic to EW gw (which fails). So to have that same check _at_ the EW gw may not make sense?

Just the DR part - the rest is required.

nmittler:
@stevenctl are you still looking into this? Can I assign to you?

stevenctl:
https://github.com/istio/istio/pull/39829#issuecomment-1179236417

stevenctl:
not-stale. Need to verify #40243 fixed this. I don' think it did.

PR #38762, opened 2022-05-05
multi-network: fix eastwest gateway endpoint filtering

stevenctl:
I think this fixes/helps with https://github.com/istio/istio/issues/38704

Gateways care about server settings only (authn, and tlsMode labels) when doing this filtering. 

We will filter out servers that we don't think accept mTLS (DISABLED or tlsMode: none). If a client somehow sent non-mTLS traffic, they would have to spoof SNI and then the server would deny them anyway.


nmittler:
@stevenctl can you also update [this test](https://github.com/istio/istio/blob/master/tests/integration/security/authorization_test.go#L887) to include a Count > 1? This test passing in multi-network actually relies on `Count==1`, which is incredibly fragile.

hzxuzhonghu (commented):
It needs a test

stevenctl:
Adjusted a test per @nmittler comment. His PR https://github.com/istio/istio/pull/37914/ improves coverage much more.

stevenctl:
> It needs a test 

Nathan's PR added tests. This PR removed the need for the global DestinationRule

hzxuzhonghu:
What;'s the behavior with tls disabled before and after? IIUC, for east-west gateway the cluster will include cluster local endpoints.

stevenctl:
This makes it so TLS disabled in DR doesn't affect passthrough gateways (the endpoint will remain). That DR based filtering makes it so clients can't send plaintext that will inevitably fail since we require TLS/SNI. 

If a server is plaintext, we still remove the endpoint from the gateway.

stevenctl:
>  will include cluster local endpoints.

Network local, non plaintext endpoints.

hzxuzhonghu:
>Network local, non plaintext endpoints.

Looks good

GregHanson (approved):

nmittler (approved):

lei-tang (approved):

nmittler:
I opened https://github.com/istio/istio/issues/39330

PR #38763, opened 2022-05-05
tests: update deployment.yaml for gRPC C++ PSM tests

menghanl:

This PR updates the integration tests to prepare for tests with gRPC C++
- Changed the flag format because C++'s flag parsing library only supports `--a=b`.
- Added a new environment variable `ECHO_IMAGE` for echo image, so it can be changed to point to the C++ image
  - Changed readiness probe to use TCP instead of HTTP when echo image is set because C++ doesn't have an easy way to start an HTTP server

### DIFF
--- a/pkg/test/env/istio.go
+++ b/pkg/test/env/istio.go
@@ -50,6 +50,10 @@ var (
 	// nolint: golint, revive, stylecheck
 	PULL_POLICY Variable = "PULL_POLICY"
 
+	// TEST_ECHO_IMAGE is the image to use when deploying echo services.
+	// nolint: golint, revive, stylecheck
+	TEST_ECHO_IMAGE Variable = "TEST_ECHO_IMAGE"
+
 	// KUBECONFIG is the list of Kubernetes configuration files. If configuration files are specified on
 	// the command-line, that takes precedence.
 	// nolint: golint, revive, stylecheck
--- a/pkg/test/env/istio.go
+++ b/pkg/test/env/istio.go
@@ -50,9 +50,9 @@ var (
 	// nolint: golint, revive, stylecheck
 	PULL_POLICY Variable = "PULL_POLICY"
 
-	// TEST_ECHO_IMAGE is the image to use when deploying echo services.
+	// ECHO_IMAGE is the image to use when deploying echo services.
 	// nolint: golint, revive, stylecheck
-	TEST_ECHO_IMAGE Variable = "TEST_ECHO_IMAGE"
+	ECHO_IMAGE Variable = "ECHO_IMAGE"
 
 	// KUBECONFIG is the list of Kubernetes configuration files. If configuration files are specified on
 	// the command-line, that takes precedence.
--- a/pkg/test/framework/components/echo/kube/deployment.go
+++ b/pkg/test/framework/components/echo/kube/deployment.go
@@ -171,7 +171,11 @@ spec:
         - containerPort: 9000
 {{- end }}
       - name: app
-        image: menghanl/interop-client:istio-echo-server
+{{- if $.ImageFullPath }}
+        image: {{ $.ImageFullPath }}
+{{- else }}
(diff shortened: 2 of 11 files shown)


pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -235,6 +235,7 @@ spec:
         - name: EXPOSE_GRPC_ADMIN
           value: "true"
 {{- end }}
+{{- if $.ImageFullPath }}
```
howardjohn, line 238: What is the relationship of image path and probe? And if there is one should it apply to liveness probe as well?

howardjohn: Oops, I see the comment now. Seems like the comment says tcp vs HTTP though but this is just disabling it entirely?

menghanl: Ahhh, sorry. Forgot to push. Please take another look.

And if you have better suggestions to do this, please let me know.

howardjohn: Only concern is what we want to say is "c++ doesn't support HTTP probe, so don't use it" but we end up saying "don't use HTTP probe for ImageFullPath [because we happen to use ImageFullPath for C++]".

if HTTP is not supported, what is done with the `--port` arguments? Just ignored? If so this is probably fine as is, we could add another flag but probably overkill

pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -668,6 +671,7 @@ func templateParams(cfg echo.Config, settings *resource.Settings) (map[string]in
 		"ImageTag":            strings.TrimSuffix(settings.Image.Tag, "-distroless"),
 		"ImagePullPolicy":     settings.Image.PullPolicy,
 		"ImagePullSecretName": imagePullSecretName,
+		"ImageFullPath":       settings.TestEchoImage, // This overrides image hub/tag if it's not empty.
```
nmittler, line 674: nit: instead of having multiple variables, just have a single variable `Image` and have it contain the appropriate value based on the settings.  You'll need a `EnableTCPHealthPort` (or something similar) as well.

pkg/test/env/istio.go
```diff
@@ -50,6 +50,10 @@ var (
 	// nolint: golint, revive, stylecheck
 	PULL_POLICY Variable = "PULL_POLICY"
 
+	// TEST_ECHO_IMAGE is the image to use when deploying echo services.
+	// nolint: golint, revive, stylecheck
+	TEST_ECHO_IMAGE Variable = "TEST_ECHO_IMAGE"
```
nmittler, line 55: nit: just call it `ECHO_IMAGE`. Everything here is for `test` :)

pkg/test/framework/components/echo/kube/deployment.go
menghanl: Yes. We are changing readiness probe to TCP for all custom images.
We will also have a Java image later. Not sure how easy it is to start an HTTP server in Java. So this may or may not be necessary for Java as well.

Another option we have is to change readiness probe to prefer gRPC whenever there is a gRPC server.

And right, the `--port` arguments are ignored by the C++ image right now.

howardjohn (approved):

pkg/test/env/istio.go
menghanl: Done!

pkg/test/framework/components/echo/kube/deployment.go
menghanl: Not sure if I understand this correctly.
Are you suggesting merging `"ImageTag"` and other image related keys in this map to one `"Image"` field?
But this map is also used by `vmDeploymentYaml`, right? Do you want to change that, too?

And I didn't get the comment about `EnableTCPHealthPort`

Issue #38497, opened 2022-04-21
[istioctl uninstall] Uninstalling the istio will not delete the resource configmap that has been created by istio.


tanjunchen:
### Bug Description

The resources created by istio should be deleted  when istio is uninstalled.

```
root@instance-0zt08gwc:~# k -n  tanjunchen get configmap
NAME                 DATA   AGE
istio-ca-root-cert   1      10s
kube-root-ca.crt     1      10s
```

```
root@instance-0zt08gwc:~# istioctl x uninstall --purge -y
All Istio resources will be pruned from the cluster

  Removed IstioOperator:istio-system:installed-state.
  Removed HorizontalPodAutoscaler:istio-system:istio-ingressgateway.
  Removed HorizontalPodAutoscaler:istio-system:istiod.
  Removed PodDisruptionBudget:istio-system:istio-ingressgateway.
  Removed PodDisruptionBudget:istio-system:istiod.
  Removed Deployment:istio-system:istio-ingressgateway.
  Removed Deployment:istio-system:istiod.
  Removed Service:istio-system:istio-ingressgateway.
  Removed Service:istio-system:istiod.
  Removed ConfigMap:istio-system:istio.
  Removed ConfigMap:istio-system:istio-sidecar-injector.
  Removed Pod:istio-system:istio-ingressgateway-7b555995fc-hbxwt.
  Removed Pod:istio-system:istiod-7dfd447c7c-blncc.
  Removed ServiceAccount:istio-system:istio-ingressgateway-service-account.
  Removed ServiceAccount:istio-system:istio-reader-service-account.
  Removed ServiceAccount:istio-system:istiod.
  Removed ServiceAccount:istio-system:istiod-service-account.
  Removed RoleBinding:istio-system:istio-ingressgateway-sds.
  Removed RoleBinding:istio-system:istiod.
  Removed RoleBinding:istio-system:istiod-istio-system.
  Removed Role:istio-system:istio-ingressgateway-sds.
  Removed Role:istio-system:istiod.
  Removed Role:istio-system:istiod-istio-system.
  Removed EnvoyFilter:istio-system:stats-filter-1.10.
  Removed EnvoyFilter:istio-system:stats-filter-1.11.
  Removed EnvoyFilter:istio-system:stats-filter-1.12.
  Removed EnvoyFilter:istio-system:tcp-stats-filter-1.10.
  Removed EnvoyFilter:istio-system:tcp-stats-filter-1.11.
[...]

PR #38774, opened 2022-05-06
remove istio-ca-root-cert configmap generated from istio namespace-controller when we uninstall istio cluster

tanjunchen:

https://github.com/istio/istio/issues/38497

remove istio-ca-root-cert configmap generated from istio namespace-controller when we uninstall istio cluster

### DIFF
--- a/operator/pkg/helmreconciler/prune.go
+++ b/operator/pkg/helmreconciler/prune.go
@@ -18,7 +18,9 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 
+	"k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


operator/pkg/helmreconciler/prune.go
```diff
@@ -177,6 +179,34 @@ func (h *HelmReconciler) DeleteObjectsList(objectsList []*unstructured.Unstructu
 	return errs.ToError()
 }
 
+// RemoveIstioCaRootCert remove the istio ca cert made by istio namespace controller
+func (h *HelmReconciler) RemoveIstioCaRootCert() error {
```
howardjohn, line 183: You may have multiple Istio installs, this is removing it from ALL of them

howardjohn (changes requested):

operator/pkg/helmreconciler/prune.go
tanjunchen: yes,it will remove all configmap *istio-ca-root-cert*. but I have some questions about it .
First:
It will leave the configmap *istio-ca-root-cert* in namespace if we run *istioctl x uninstall --purge*.
```
The optional --purge flag will remove all Istio resources, including cluster-scoped resources that may be shared with other Istio control planes.
``` 
This is not the same as the description.
Second:
Even if I delete the cm, if there are other control plane, the namespace controller will automatically create it immediately.
Thrid:
I tested and found that if a custom certificate is used and cm exists in the current namespace, the application under this namespace will fail to start.
Is there any other better way to solve this problem? thanks. @howardjohn

PR #38781, opened 2022-05-06
Fix resolving network gateway names with multiple and/or misbehaving DNS upstreams

### LINKED ISSUE #38689: Network gateway hostnames are not correctly resolved when DNS upstreams don't support ANY queries
### Bug Description

I've been building a proof-of-concept multi-cluster mesh with multi-primaries in different networks. In my case, one cluster is AWS EKS and another is DigitalOcean managed Kubernetes. The east-west gateway in the EKS cluster is exposed with an AWS NLB.

After linking the clusters as [per the documentation](https://istio.io/latest/docs/setup/install/multicluster/multi-primary_multi-network/#enable-endpoint-discovery), I've discovered that the domain name of the AWS NLB in the EKS cluster is not correctly resolved in the DO cluster. It turned out that the upstream DNS set up on DO KS nodes is returning `REFUSED` answers to `ANY` queries that are currently used in the Pilot code:
https://github.com/istio/istio/blob/e67c34b8c005ec13d2a3b2dd897b45124a02a5ca/pilot/pkg/model/network.go#L485-L486

```prose
$ dig -t ANY k8s-istiomul-istioeas-4d501f177f-9a9de7682aacbcd6.elb.us-west-2.amazonaws.com

; <<>> DiG 9.16.1-Ubuntu <<>> -t ANY k8s-istiomul-istioeas-4d501f177f-9a9de7682aacbcd6.elb.us-west-2.amazonaws.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: REFUSED, id: 44741
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
; COOKIE: c367afe288a65985 (echoed)
;; QUESTION SECTION:
;k8s-istiomul-istioeas-4d501f177f-9a9de7682aacbcd6.elb.us-west-2.amazonaws.com. IN ANY

;; Query time: 7 msec
;; SERVER: 10.245.0.10#53(10.245.0.10)
;; WHEN: Mon May 02 03:20:07 UTC 2022
;; MSG SIZE  rcvd: 118
```

`ANY` queries are not guaranteed to be consistently implemented in DNS servers. For example, Cloudflare [deems them deprecated](https://blog.cloudflare.com/rfc8482-saying-goodbye-to-any/) and their NS return `NOTIMP` to `ANY` queries.

I'd suggest replacing `ANY` with `A` and `AAAA` queries, as mentioned by the comment in the code. Though it is technically possible to craft a multi-type query with the library currently in use, such queries [also seem](https://stackoverflow.com/a/4083071) not guaranteed to be implemented consistently, so we'd likely have to make two separate queries and merge the results. I have a patch tested in my environment and can follow up with a PR.

### Version

```prose
$ istioctl version
client version: 1.13.3
control plane version: 1.13.3
data plane version: 1.13.3 (1 proxies)

$ kubectl version --short
Client Version: v1.23.6
Server Version: v1.22.8
```

### Additional Information

_No response_

### Affected product area

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Upgrade
- [x] Multi Cluster
- [ ] Virtual Machine
- [ ] Control Plane Revisions

sangwa:

* Fixes an issue with resolving network gateway names when multiple DNS upstreams are provided to the control plane (e.g. via resolv.conf) but only the first one is ever tried
* Fixes an issue when network gateway names cannot be resolved if DNS upstreams don't support ANY queries (e.g. DigitalOcean - returns REFUSED or Cloudflare - returns NOTIMPL)

Fixes #38689

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pilot/pkg/model/network.go
```diff
@@ -482,14 +482,41 @@ func stringSliceEqual(a, b []string) bool {
 
 // resolve gets all the A and AAAA records for the given name
 func (n *networkGatewayNameCache) resolve(name string) ([]string, time.Duration) {
-	// TODO figure out how to query only A + AAAA
-	res := n.client.Query(new(dns.Msg).SetQuestion(dns.Fqdn(name), dns.TypeANY))
-	if res == nil || len(res.Answer) == 0 {
+	var wg sync.WaitGroup
+	var resA, resAAAA *dns.Msg
+
+	wg.Add(2)
```
howardjohn, line 488: can we see what other clients do? parallel DNS resolution is tricky concept that has caused many bugs in other products I have seen. I don't feel comfortable doing much other than directly copying something established like coredns,consul,tailscale, etc. Or even stdlib

sangwa: Thanks for the catch. I introduced the parallel resolution trying to avoid an overhead when we only have one kind of records (either A or AAAA) but have to resolve both sequentially. However I haven't considered that we have two independent instances of the DNS client here and they can receive each other's datagrams mixed up, in this case the client library drops them. stdlib sends both datagrams sequentially and then waits for the both results specifically within the same handler function using request IDs to discern them. That is not possible to implement with the client library in use here, so I revert to sequential resolution.

pilot/pkg/model/network.go
```diff
@@ -544,15 +561,31 @@ func newClient() (*dnsClient, error) {
...
 		if err == nil {
 			response = cResponse
-			break
+			code := response.MsgHdr.Rcode
+			if code == dns.RcodeSuccess {
+				break
+			}
+			codeString := dns.RcodeToString[code]
+			// TODO make Debug as to not spam the logs on retries?
+			log.Infof("upstream dns error: %v: %v: %v", upstream, getReqNames(req), codeString)
```
hzxuzhonghu, line 585: Is there an error when only one type of DNS record supported?

sangwa: Well-behaved DNS servers should return NOERROR with an empty record set, as per RFC, but [that's not guaranteed](https://datatracker.ietf.org/doc/html/rfc4074#section-3). If an upstream DNS server has no proper IPv6 support then each AAAA lookup can lead to a logged NXDOMAIN error.

I think we can omit logging here if the status is NXDOMAIN. If there is either A or AAAA record the error for another one is redundant, and if both are missing the error will be logged in the downstream function NetworkManager.resolveHostnameGateways():
https://github.com/istio/istio/blob/e67c34b8c005ec13d2a3b2dd897b45124a02a5ca/pilot/pkg/model/network.go#L191-L193

However if there is a concern for other unexpected errors from the upstream I think it should be indeed downgraded to `log.Debugf`.

hzxuzhonghu: Yes, that's my concern

sangwa: Changed to `log.Debugf`. Not skipping on NXDOMAIN then as it may indicate a misbehaving upstream and it's on the debug log level now.

sangwa:
@howardjohn @stevenctl PTAL
Would be really nice to have this shipped in 1.14

hzxuzhonghu (approved):

sangwa:
Added another small fix in 42f3818: the TTL of a DNS record is returned in seconds but was treated as nanosecords because the derived time.Duration was not multiplied by time.Second and so the TTL was never actually used (always overridden by the default of 30 seconds).

Also I took a closer look at the DNS client library in use. It creates a new net.Dialer for every query unless a specific net.Dialer is pre-set directly by the calling code, so every new query uses a different local port binding and therefore parallel DNS resolution is safe, at least while the library's logic stays unmodified. I can revert the parallel resolution code back in place if it seems appropriate.

stevenctl:
LGTM

stevenctl (approved):

PR #38791, opened 2022-05-08
Allow reconciliation to resolve transient finalizer conflict errors

### LINKED ISSUE #38591: Istio operator doesn't re-reconcile on conflicts
### Bug Description

When the Istio operator attempts to reconcile an IstioOperator resource and encounters a temporary conflict (409), it does not return an error, and therefore does not attempt to re-reconcile the IstioOperator resource.

https://github.com/istio/istio/blob/master/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go#L291
https://github.com/istio/istio/blob/master/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go#L308

This can leave you in a state where the resource finalizer never gets created, and the resource ultimately never gets created.

### Version

```
istioctl-1.11.2 version
client version: 1.11.2
control plane version: 1.11-dev
data plane version: none

Note this issue is present in head branch.
```

### Minimal Reproduction

Create empty Kind cluster:
```
kind create cluster --name kind
Creating cluster "kind" ...
⢄⡱ Ensuring node image (kindest/node:v1.24.0) 🖼
...
```
Install the operator using istioctl:
```
istioctl operator init
Installing operator controller in namespace: istio-operator using image: docker.io/istio/operator:1.13.4
Operator controller will watch namespaces: istio-system
✔ Istio operator installed
✔ Installation complete                                                                                                                                    
...
```

In one terminal, tail the operator logs:
```
kubectl -n istio-operator logs -l name=istio-operator -f
```

In another terminal, kick off a loop to update an annotation on an IstioOperator resource:
```
cat <<EOF > loop.sh
set -x
for i in {1..1000}
do
   exists=$(kubectl get -n istio-system istiooperator base-operator-profile 2>/dev/null)
   echo $exists
   if [[ ! -z "${exists}" ]]; then
     timestamp=$(date +%s)
     kubectl -n istio-system patch istiooperator base-operator-profile \
       --type=merge -p '{"metadata":{"annotations":{"foo":'\"${timestamp}\"'}}}'
   fi
   sleep 1
done
EOF

chmod +x ./loop.sh

./loop.sh
```

In a third terminal, apply any IstioOperator to the cluster:
```
cat <<EOF | kubectl apply -f -
kind: IstioOperator
metadata:
  name: base-operator-profile
  namespace: istio-system
spec:
  profile: empty
  components:
    base:
      enabled: true
EOF
istiooperator.install.istio.io/base-operator-profile created
```

Observe that the istio-operator fails to apply the resource due to a conflict. Importantly, note that it never actually tries to reconcile again.

```
kc -n istio-operator logs -l name=istio-operator -f
2022-06-21T16:06:28.669708Z	info	installer	Reconciling IstioOperator
2022-06-21T16:06:29.772729Z	info	Adding finalizer istio-finalizer.install.istio.io to istio-system/base-operator-profile
2022-06-21T16:06:29.777632Z	info	installer	Could not add finalizer to base-operator-profile due to conflict. Operation will be retried in next reconcile attempt.
```

In our case we hit this issue when deploying an IstioOperator via a Helm chart, managed by Helm Operator. The Helm Operator itself attempts to immediately annotate and label each resource it applies to the cluster, causing an immediate conflict for any other entity in the cluster which attempts to modify a given resource.

rtmie:
Do not return 'nil' when transient conflict is detected when adding finalizer - this prevents subsequent reconciliation attempts on the istiooperator resource.

Resolves https://github.com/istio/istio/issues/38591

Signed-off-by: Rob Mason <robmie@gmail.com>

### DIFF
--- a/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
+++ b/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
@@ -288,6 +288,7 @@ func (r *ReconcileIstioOperator) Reconcile(_ context.Context, request reconcile.
 				return reconcile.Result{}, nil
 			} else if errors.IsConflict(finalizerError) {
 				scope.Infof("Could not remove finalizer from %s due to conflict. Operation will be retried in next reconcile attempt.", iopName)
+				return reconcile.Result{}, finalizerError
 			}
 			scope.Errorf(errdict.OperatorFailedToRemoveFinalizer, "error removing finalizer: %s", finalizerError)
 			return reconcile.Result{}, finalizerError
@@ -305,6 +306,7 @@ func (r *ReconcileIstioOperator) Reconcile(_ context.Context, request reconcile.
 				return reconcile.Result{}, nil
 			} else if errors.IsConflict(err) {
 				scope.Infof("Could not add finalizer to %s due to conflict. Operation will be retried in next reconcile attempt.", iopName)
+				return reconcile.Result{}, err
 			}
 			scope.Errorf(errdict.OperatorFailedToAddFinalizer, "Failed to add finalizer to IstioOperator CR %s: %s", iopName, err)
 			return reconcile.Result{}, err
--- a/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
+++ b/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
@@ -288,7 +288,7 @@ func (r *ReconcileIstioOperator) Reconcile(_ context.Context, request reconcile.
 				return reconcile.Result{}, nil
 			} else if errors.IsConflict(finalizerError) {
 				scope.Infof("Could not remove finalizer from %s due to conflict. Operation will be retried in next reconcile attempt.", iopName)
-				return reconcile.Result{}, finalizerError
+				return reconcile.Result{}, nil
 			}
 			scope.Errorf(errdict.OperatorFailedToRemoveFinalizer, "error removing finalizer: %s", finalizerError)
 			return reconcile.Result{}, finalizerError
--- a/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
+++ b/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
@@ -288,7 +288,6 @@ func (r *ReconcileIstioOperator) Reconcile(_ context.Context, request reconcile.
 				return reconcile.Result{}, nil
 			} else if errors.IsConflict(finalizerError) {
 				scope.Infof("Could not remove finalizer from %s due to conflict. Operation will be retried in next reconcile attempt.", iopName)
-				return reconcile.Result{}, nil
 			}
 			scope.Errorf(errdict.OperatorFailedToRemoveFinalizer, "error removing finalizer: %s", finalizerError)
 			return reconcile.Result{}, finalizerError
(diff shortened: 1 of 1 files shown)


operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
```diff
@@ -288,7 +288,6 @@ func (r *ReconcileIstioOperator) Reconcile(_ context.Context, request reconcile.
 				return reconcile.Result{}, nil
 			} else if errors.IsConflict(finalizerError) {
 				scope.Infof("Could not remove finalizer from %s due to conflict. Operation will be retried in next reconcile attempt.", iopName)
-				return reconcile.Result{}, nil
```
morvencao, line 291: I would suggest to replace `return reconcile.Result{}, nil` with `continue, so that the `finalizerMaxRetries` still works. If you remove `return reconcile.Result{}, nil`, it will be forced to next reconcile attempt.

rtmie: let me look at that

rtmie: @morvencao that codepoint is not in a loop so cannot do a continue.
I can return with the error immediately it occurs rather than let it drop through?
Is that ok?

morvencao: then the `finalizerMaxRetries` will take no effect, right?

rtmie: @morvencao apologies had misunderstood your original comment.

Really I only need the retry in the create side, should not have made change in the remove part.

LEt me resubmit

rtmie: AM I right in saying that the maxRetries is only concerned with conflict error in the finalizer removal, not the addition?

For my use case I really am concerned only with the create case.

I can revert the change in the remove part.

morvencao: So the change mean that you want to force the controller to start new reconcile loop once finalizer add conflict is found.

rtmie: @morvencao 

yes - otherwise a temporary conflict will never be resolved. The issue here occurs when delivering the operator profile by helm3.

Helm3 operator delivers the operator profile doc and adds a label to it `app.kubernetes.io/managed-by: Helm`. But there is a race as the resource state has now changed whillst the istio operator reconcile logic is running

Meanwhile istio-operator is trying to process causing an IOP error (which is not logged - I got this log by allowing the error to drop trhough:

```
2022-04-27T12:15:43.253727Z     info    installer       Reconciling IstioOperator
2022-04-27T12:15:43.858528Z     info    Adding finalizer istio-finalizer.install.istio.io to istio-system/istio-control-plane
2022-04-27T12:15:43.869335Z     info    installer       Could not add finalizer to istio-control-plane due to conflict. Operation will be retried in next reconcile attempt.
2022-04-27T12:15:43.869376Z     error   installer       Failed to add finalizer to IstioOperator CR istio-control-plane: Operation cannot be fulfilled on istiooperators.install.istio.io "istio-control-plane": the object has been modified; please apply your changes to the latest version and try again    moreInfo=A finalizer could not be added to the IstioOperator resource. The controller uses the finalizer to temporarily prevent the resource from being deleted while the Istio control plane is being deleted. impact=When the IstioOperator resource is deleted, the Istio control plane may not be fully removed. action=Delete and re-add the IstioOperator resource. If this error persists, see https://istio.io/latest/about/bugs for possible solutions. likelyCause=The likely cause is a problem with the Kubernetes API server.
```

As it stands without this change there is no subsequent attempt made to resolve this and the processing of the operator resource is never retried.

There are other possibilities for a transient  conflict to occur and the reconcile loop should resolve this by re-running

rtmie:
@morvencao any update?

litong01 (approved):
The PR looks good to me and have tested this case, currently when there is a conflict, even if you make changes to the resources, the controller wont reconcile so the resources wont be created. This fix should be merged.

PR #38793, opened 2022-05-08
[WIP]fix istioctl anaylze

zirain:

there's still some wrong with https://github.com/istio/istio/pull/38659 

I think the root cause is here
https://github.com/istio/istio/blob/d0748d7c46c02e1fddf77d7a38f90cb3620e4358/pkg/config/analysis/local/istiod_analyze.go#L278

another question is why do we set revision to "default" here.
https://github.com/istio/istio/blob/a0f6a9463feb5ffecf2ee23d4527b846260b7dcf/pkg/config/analysis/incluster/controller.go#L53

### DIFF
--- a/istioctl/cmd/analyze.go
+++ b/istioctl/cmd/analyze.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"istio.io/istio/istioctl/pkg/tag"
 	"istio.io/istio/istioctl/pkg/util/formatting"
 	"istio.io/istio/istioctl/pkg/util/handlers"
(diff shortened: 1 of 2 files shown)


GregHanson:
adding `release-notes-none` since this fix is still a part of https://github.com/istio/istio/pull/38659

istioctl/cmd/analyze.go
```diff
@@ -188,7 +189,19 @@ func Analyze() *cobra.Command {
 				if err != nil {
 					return err
 				}
-				sa.AddRunningKubeSource(k)
+				if allNamespaces {
+					sa.AddRunningKubeSource(k)
+					revisions, _ := tag.ListRevisions(context.TODO(), k.Kube())
```
howardjohn, line 194: injection revision and config revision are orthogonal. A namespace can be configured with rev1 and have CRs for rev2, rev3, etc.

PR #26838, opened 2020-08-27
XDS Proxy

rshriram:
Taking over from @howardjohn's prototype https://github.com/istio/istio/pull/26780

Currently disabled by default. Enabled only for pilot integration tests.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[x] Does not have any changes that may affect Istio users.

PR #38798, opened 2022-05-09
remove unnecessary comment in xds proxy

ramaraochavali:
Comment says to block but blockoption is not added.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pkg/istio-agent/xds_proxy.go
```diff
@@ -655,6 +655,7 @@ func (p *XdsProxy) buildUpstreamClientDialOpts(sa *Agent) ([]grpc.DialOption, er
 	dialOptions := []grpc.DialOption{
 		tlsOpts,
 		keepaliveOption, initialWindowSizeOption, initialConnWindowSizeOption, msgSizeOption,
+		grpc.WithBlock,
```
hzxuzhonghu, line 658: It was there In the original pr, can you checkout when and why this is removed. I can't remember

ramaraochavali: I looked at the history but could not find it in the original file also.

hzxuzhonghu: #26838

hzxuzhonghu (approved):

PR #38801, opened 2022-05-09
make use of networks to determine internal address config

ramaraochavali:
Alternate implementation as discussed in https://github.com/istio/api/pull/2312
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali:
@costinm PTAL. Internal address config implementation using networks API. If you are good with this, I will add tests.

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -634,6 +635,18 @@ func buildGatewayConnectionManager(proxyConfig *meshconfig.ProxyConfig, node *mo
 		httpConnManager.Http3ProtocolOptions = &core.Http3ProtocolOptions{}
 		httpConnManager.CodecType = hcm.HttpConnectionManager_HTTP3
 	}
+	if push != nil && push.Networks != nil {
```
howardjohn, line 638: nit: can use .GetNetworks.GetNetworks to avoid nil checks

costinm (commented):
Looks good except for the naming. 

I would think in a multi-network situation all defined networks should be used, and the env should be 'use_internal_networks' or something similar. Maybe hcm_internal_networks to be clear this is for hcm only ( bool, used as a feature flag).

pkg/config/mesh/networks_watcher.go
```diff
@@ -29,6 +29,7 @@ import (
 type NetworksHolder interface {
 	SetNetworks(*meshconfig.MeshNetworks)
 	Networks() *meshconfig.MeshNetworks
+	LastNetworks() *meshconfig.MeshNetworks
```
hzxuzhonghu, line 32: Not sure what is last network

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -1220,6 +1220,17 @@ func (lb *ListenerBuilder) buildHTTPConnectionManager(httpOpts *httpListenerOpts
...
 
+	if lb.push.Networks != nil {
+		if internalnetwork, exists := lb.push.Networks.Networks[features.InternalAddressMeshNetwork]; exists {
+			iac := &hcm.HttpConnectionManager_InternalAddressConfig{}
+			for _, ne := range internalnetwork.Endpoints {
+				if cidr := util.ConvertAddressToCidr(ne.GetFromCidr()); cidr != nil {
+					iac.CidrRanges = append(iac.CidrRanges)
+				}
+			}
+			httpOpts.connectionManager.InternalAddressConfig = iac
```
hzxuzhonghu, line 1231: Curious why outbound can work, according to envoy 
```
Envoy uses [XFF](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#config-http-conn-man-headers-x-forwarded-for) to determine this and then will set the header value to true.
```

hzxuzhonghu (commented):
A test should be better

ramaraochavali:
I will add tests and rename env var.

pkg/config/mesh/networks_watcher.go
ramaraochavali: I will rename this to PrevNetworks

ramaraochavali:
@costinm can you PTAL? All comments addressed.

ramaraochavali:
@costinm @howardjohn @hzxuzhonghu PTAL when you get chance

costinm (approved):

pilot/pkg/bootstrap/mesh.go
```diff
@@ -115,6 +118,24 @@ func (s *Server) initMeshNetworks(args *PilotArgs, fileWatcher filewatcher.FileW
 		log.Info("mesh networks configuration not provided")
 		s.environment.NetworksWatcher = mesh.NewFixedNetworksWatcher(nil)
 	}
+	s.environment.AddNetworksHandler(func() {
```
hzxuzhonghu, line 121: My question is we have already registered a network handle, which will trigger xds push

ramaraochavali: Those network handles does not trigger a push when endpoint changes. Only if gateway changes they push.

hzxuzhonghu: So can we merge them? Currently they are scattered and looks difficult to maintain

ramaraochavali: Yes. I agree. I will do a follow-up and merge them. It needs little refactoring because part of it is in network manager.

hzxuzhonghu: SGTM

hzxuzhonghu (approved):

ramaraochavali:
@istio/wg-test-and-release-maintainers @istio/release-managers PTAL

PR #38802, opened 2022-05-09
Fixtodo rename int32orstring

ramanujadasu:
Fix the rename int32orstring property changes
/release-notes-none


morvencao (commented):
kubernete still uses `IntOrString`: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/intstr#IntOrString

howardjohn (changes requested):
This "TODO" is from Kubernetes. We copy the Kubenetes protos and their TODOs.

I strongly discourage fixing random TODOs without context and would suggest instead looking through github issues with "help wanted" label

PR #38807, opened 2022-05-09
feat(charts/istio-discovery): add meshConfig.trustDomainAliases list

carnei-ro:
Signed-off-by: Leandro Carneiro <leandro@carnei.ro>

Add missing field meshConfig.trustDomainAliases to chart istio-discovery.

- [ ] Docs
- [X] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


costinm:
I don't think we want to list all fields from meshconfig - not sure why we have the few that got added, the defaults in code should work.

If there is a use case - please add a comment.

carnei-ro:
I need this! I’m installing a multi-cluster mesh with trustDomain per cluster (to identify which cluster the requesting).

PR #38818, opened 2022-05-10
istioctl: support `gateway` as input type

howardjohn:
This allows things like `istioctl proxy-config listeners gateway/my-gateway`. Also
included some bug fixes for namespace resolution (which is valid even
without the Gateway part).

### DIFF
--- a/istioctl/pkg/util/handlers/handlers_test.go
+++ b/istioctl/pkg/util/handlers/handlers_test.go
@@ -56,6 +56,42 @@ func TestInferPodInfo(t *testing.T) {
 			wantPodName:   "istio-security-post-install-1.2.2-bm9w2",
 			wantNamespace: "istio-system",
 		},
+		{
+			proxyName:     "service/istiod",
+			namespace:     "",
+			wantPodName:   "service/istiod",
+			wantNamespace: "",
+		},
+		{
+			proxyName:     "service/istiod",
+			namespace:     "namespace",
+			wantPodName:   "service/istiod",
+			wantNamespace: "namespace",
+		},
+		{
+			proxyName:     "service/istiod.istio-system",
+			namespace:     "namespace",
+			wantPodName:   "service/istiod",
+			wantNamespace: "istio-system",
+		},
+		{
+			proxyName:     "gateway.gateway.networking.k8s.io/istiod",
+			namespace:     "",
+			wantPodName:   "gateway.gateway.networking.k8s.io/istiod",
+			wantNamespace: "",
+		},
+		{
+			proxyName:     "gateway.gateway.networking.k8s.io/istiod",
+			namespace:     "namespace",
+			wantPodName:   "gateway.gateway.networking.k8s.io/istiod",
+			wantNamespace: "namespace",
+		},
+		{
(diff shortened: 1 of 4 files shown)


hanxiaop (approved):

hzxuzhonghu:
Seems no change to cmd to support `istioctl proxy-config listeners gateway/my-gateway`, do I miss something

howardjohn:
InferPodInfoFromTypedResource is changed to support gateway and proxy-config uses that.

istioctl/pkg/util/handlers/handlers_test.go
```diff
@@ -56,6 +56,42 @@ func TestInferPodInfo(t *testing.T) {
 			wantPodName:   "istio-security-post-install-1.2.2-bm9w2",
 			wantNamespace: "istio-system",
 		},
+		{
+			proxyName:     "service/istiod",
```
hzxuzhonghu, line 60: Not sure about this usage and "gateway.gateway.networking.k8s.io/istiod"

Curentl it is `istioctl proxy-config clusters <pod-name[.namespace]> `, right

howardjohn: No, current usage is `name.namespace` or `type/name.namespace`. 

The existing bug is type cannot have a `.` when it's perfectly valid syntax in k8s

hzxuzhonghu: Oh i noticed it, `listener [<type>/]<name>[.<namespace>]` looks very tricky

istioctl/pkg/util/handlers/handlers_test.go
```diff
@@ -56,6 +56,42 @@ func TestInferPodInfo(t *testing.T) {
...
+			namespace:     "namespace",
+			wantPodName:   "service/istiod",
+			wantNamespace: "namespace",
+		},
+		{
+			proxyName:     "service/istiod.istio-system",
+			namespace:     "namespace",
+			wantPodName:   "service/istiod",
```
hzxuzhonghu, line 74: I am confused about the pod names here you added

howardjohn: I think the problem is before this function returned "pod name" but was later changed to "resource name" without updating the tests. I can fix it but it's not really in scope of my PR.

hzxuzhonghu: Can you also explain what does below cmd do with this pr?

`istioctl pc listener gateway.gateway.networking.k8s.io/istiod`

howardjohn: The same thing `svc/xyz`, `deployment/xyz`, etc. Gets a pod associated with that resource. There is a direct mapping from Service to Pod, and similarly there is a direct mapping from Gateway to Pod which we use here.

This allows users to think about higher level construct (Gateway) and not worry about implementation detail (pod name)

hzxuzhonghu: IC, that is to say i can use istio gateway as well, though I do not think anyone will use it like this.

hzxuzhonghu (approved):

PR #38825, opened 2022-05-10
add nil check to authz builder

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


lei-tang (approved):

howardjohn (commented):
when is it nil?

howardjohn:
Should we be using the NewListenerBuilder that handles that though?

I don't mind this either but seems if we run into this case we. may be doing something wrong

ramaraochavali:
> Should we be using the NewListenerBuilder that handles that though?

Yes we can. Some tests are still using &ListenerBuilder and we can certainly move to NewListenerBuilder.

One of the builders(auth) already handles this. So thought it is better to be consistent and handle these as well.

PR #38837, opened 2022-05-10
Merge changes from istio template

costinm:
A merge of main injection template into the grpc template - they diverged a bit. 

Long term we should merge them completely, and use an annotation to control if we start envoy or not - the template 
will operate in both modes, depending on the application type ( grpc only or grpc + TCP/HTTP ). 

This PR:
- will fix lifecycle, which was not actually used. The agent port is used instead
- mount zatar and other forms of certificates (changes in main template)
- minimize the diff, so next time will be easier to sync

It does not merge the security section, no need to interfere with it since gRPC is not using iptables.


manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
```diff
@@ -35,17 +67,34 @@ spec:
...
+    ports:
+    - containerPort: 15090
+      protocol: TCP
+      name: http-envoy-prom
     args:
     - proxy
     - sidecar
     - --domain
     - $(POD_NAMESPACE).svc.{{ .Values.global.proxy.clusterDomain }}
+    - --proxyLogLevel={{ annotation .ObjectMeta `sidecar.istio.io/logLevel` .Values.global.proxy.logLevel }}
```
howardjohn, line 79: nit: this was probably intentionally kept out since it does nothing. I don't mind it being here for consistency though

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
```diff
@@ -1,17 +1,49 @@
...
+    {{- end }}
+    {{- if not (isset .ObjectMeta.Annotations `kubectl.kubernetes.io/default-container`) }}
     kubectl.kubernetes.io/default-container: "{{ index $containers 0 }}",
-    {{ end }}
+    {{- end }}
+    {{- end }}
     sidecar.istio.io/rewriteAppHTTPProbers: "false",
   }
 spec:
+  {{- $holdProxy := or .ProxyConfig.HoldApplicationUntilProxyStarts.GetValue .Values.global.proxy.holdApplicationUntilProxyStarts }}
```
howardjohn, line 46: not sure this works with grpc. Probably does though

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
```diff
@@ -145,42 +192,65 @@ spec:
...
+      fsGroup: 1337
+      runAsNonRoot: false
+      runAsUser: 0
+      {{- else }}
+      allowPrivilegeEscalation: {{ .Values.global.proxy.privileged }}
+      capabilities:
+        {{ if or (eq (annotation .ObjectMeta `sidecar.istio.io/interceptionMode` .ProxyConfig.InterceptionMode) `TPROXY`) (eq (annotation .ObjectMeta `sidecar.istio.io/capNetBindService` .Values.global.proxy.capNetBindService) `true`) -}}
+        add:
+        {{ if eq (annotation .ObjectMeta `sidecar.istio.io/interceptionMode` .ProxyConfig.InterceptionMode) `TPROXY` -}}
+        - NET_ADMIN
```
howardjohn, line 221: I don't think we want to add privs?

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
```diff
@@ -222,23 +310,44 @@ spec:
...
-{{- end }}
+  {{- end }}
   {{- if eq .Values.global.pilotCertProvider "istiod" }}
   - name: istiod-ca-cert
     configMap:
       name: istio-ca-root-cert
   {{- end }}
-  {{- if isset .ObjectMeta.Annotations `sidecar.istio.io/userVolume` }}
-  {{range $index, $value := fromJSON (index .ObjectMeta.Annotations `sidecar.istio.io/userVolume`) }}
+  {{- if .Values.global.mountMtlsCerts }}
```
howardjohn, line 327: Unless we want minimal diff: not needed

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
```diff
@@ -222,23 +310,44 @@ spec:
...
+    {{ toYaml $value | indent 4 }}
+    {{ end }}
+    {{ end }}
+  {{- if .Values.global.imagePullSecrets }}
+  imagePullSecrets:
+    {{- range .Values.global.imagePullSecrets }}
+    - name: {{ . }}
+    {{- end }}
+  {{- end }}
+  {{- if eq (env "ENABLE_LEGACY_FSGROUP_INJECTION" "true") "true" }}
```
howardjohn, line 350: Unless we want minimal diff: not needed

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
```diff
@@ -194,20 +264,38 @@ spec:
...
+  {{- if eq .Values.global.caName "GkeWorkloadCertificate" }}
+  - name: gke-workload-certificate
+    csi:
+      driver: workloadcertificates.security.cloud.google.com
+  {{- else }}
+  - emptyDir:
     name: workload-certs
-  # UDS channel between istioagent and gRPC client for XDS/SDS
+  {{- end }}
+  {{- if (isset .ObjectMeta.Annotations `sidecar.istio.io/bootstrapOverride`) }}
```
howardjohn, line 293: Unless we want minimal diff: not needed

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
```diff
@@ -194,20 +264,38 @@ spec:
     - mountPath: /var/run/secrets/tokens
       name: istio-token
     {{- end }}
+    {{- if .Values.global.mountMtlsCerts }}
```
howardjohn, line 267: Unless we want minimal diff: not needed

howardjohn (commented):
All comments non-blocking but the securitycontext one - I think we want to be sure we never have NET_ADMIN?

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
costinm: Easier to do merges when most lines are identical.

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
costinm: Right, will remove it. 

The direction of this is to fully merge the 2 templates - and to support a 'hybrid' mode, where envoy and interception are used for HTTP/TCP and gRPC is excluded.  But we'll also need a grpc-only mode, and no need to grant net_admin.

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
costinm: That was the reason I started this PR...

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
costinm: This one will be ( with few changes in agent ) - grpc should support mounted certs, zatar, etc.

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
costinm: I removed this line - for proxyless we need 'lifecycle' in all cases. It was not working - 'wait' was probing an envoy port, but I fixed the URL and now seems to be fine.

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
howardjohn: This is for the super legacy citadel secret mounting. the `/var/run/secrets/workload-spiffe-credentials` one is the new one

costinm: I know - if we still support it for sidecar, why not support it here too ? 

I don't think we have citadel code to sync secrets any more - but other tools can create similar secrets. Probably should be per-workload, not global.

howardjohn: Thats fair. My rationale was that for new code we should not carry forward tech debt

manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
```diff
@@ -1028,18 +1059,31 @@ data:
...
+            - --proxyLogLevel={{ annotation .ObjectMeta `sidecar.istio.io/logLevel` .Values.global.proxy.logLevel }}
+            - --proxyComponentLogLevel={{ annotation .ObjectMeta `sidecar.istio.io/componentLogLevel` .Values.global.proxy.componentLogLevel }}
             - --log_output_level={{ annotation .ObjectMeta `sidecar.istio.io/agentLogLevel` .Values.global.logging.level }}
           {{- if .Values.global.sts.servicePort }}
             - --stsPort={{ .Values.global.sts.servicePort }}
           {{- end }}
           {{- if .Values.global.logAsJson }}
             - --log_as_json
           {{- end }}
+            lifecycle:
```
howardjohn, line 1080: I think this doesn't do anything unless the holdProxy setting is also set, since that will ensure the injected container is first which is the critical component tothis

howardjohn (approved):

sanjaypujare:
Can this be ported to 1.14 branch so it's available in the next 1.14 patch release?

SkyfireFrancisZ:
Just confirmed with Istio 1.14 RM.  The owner (@costinm?) can cherry-pick this into release-1.14.  The RM will approve and it'll go out with the next patch.

sanjaypujare:
Opened PR https://github.com/istio/istio/pull/40410 to backport to 1.14.

PR #38838, opened 2022-05-10
Refactoring and extending `registryredirector` (formerly `containerregistry`) integration test component

ingwonsong:
To give more flexibility and clarity, this PR proposes following changes:

1. Renaming from `containerregistry` to `registryredirector` to prevent misconception.
2. Re-org Makefile for `fake/imageregistry`
3. Using JSON format for setting/retrieving the tag map
4. Providing a way to set the target registry (previously, fixed to `gcr.io`)

### DIFF
--- a/pkg/test/framework/components/registryredirector/kube.go
+++ b/pkg/test/framework/components/registryredirector/kube.go
@@ -12,14 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package containerregistry
+package registryredirector
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net"
 	"net/http"
-	"net/url"
+	"time"
 
 	"istio.io/istio/pkg/test/env"
 	"istio.io/istio/pkg/test/framework/components/cluster"
@@ -52,14 +54,14 @@ func newKube(ctx resource.Context, cfg Config) (Instance, error) {
 	}
 	c.id = ctx.TrackResource(c)
 	var err error
-	scopes.Framework.Info("=== BEGIN: Deploy container registry server ===")
+	scopes.Framework.Info("=== BEGIN: Deploy registry redirector server ===")
 	defer func() {
 		if err != nil {
 			err = fmt.Errorf("container registry deployment failed: %v", err)
-			scopes.Framework.Infof("=== FAILED: Deploy container registry server ===")
+			scopes.Framework.Infof("=== FAILED: Deploy registry redirector server ===")
 			_ = c.Close()
 		} else {
-			scopes.Framework.Info("=== SUCCEEDED: Deploy container registry server ===")
+			scopes.Framework.Info("=== SUCCEEDED: Deploy registry redirector server ===")
 		}
 	}()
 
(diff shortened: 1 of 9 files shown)


pkg/test/framework/components/registryredirector/registryredirector.go
```diff
@@ -12,43 +12,48 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
-// Package containerregistry provides basic utilities around configuring the fake
-// container registry server component for integration testing.
-package containerregistry
+// Package registryredirector provides basic utilities around configuring the fake
+// image registry server component for integration testing.
+package registryredirector
```
douglas-reid, line 17: is this functioning then as a proxy?  should we call this `registryproxy` instead ?

pkg/test/framework/components/registryredirector/registryredirector.go
```diff
@@ -12,43 +12,48 @@
...
 	SetupTagMap(map[string]string) error
 }
 
-// Config defines the options for creating an fake container registry component.
+// Config defines the options for creating an registry redirector component.
 type Config struct {
 	// Cluster to be used in a multicluster environment
 	Cluster cluster.Cluster
+	// Upstream registry. Default is "gcr.io".
+	TargetRegistry string
```
douglas-reid, line 40: can we add some details (in the CL description) on why `TargetRegistry` is needed now? I _think_ this is expanding the flexibility of the fake, which is alluded to in the current description. But I'm not clear on what is motivating that need.  Is it because we want to support alternative image registries in tests? Is there an open issue?

pkg/test/framework/components/registryredirector/registryredirector.go
```diff
@@ -12,43 +12,48 @@
...
+	// Docker image location of the fake registry. Default is "gcr.io/istio-testing/fake-registry:x.x".
+	// Please refer to registry_redirector_server.yaml for the exact default image.
+	RegistryRedirectorImage string
```
douglas-reid, line 43: nit: i'd just call this something simpler like `Image`.  this is already inside of `registryredirector.Config`.

pkg/test/framework/components/registryredirector/kube.go
```diff
@@ -12,14 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package containerregistry
+package registryredirector
```
douglas-reid, line 15: same question as below: could this be considered a proxy (`registryproxy`)?

douglas-reid (approved):
this seems generally fine to me. there shouldn't be any real issues, as this is a testing component. if you are going to be working on it/maintaining it, i trust your judgement on what naming convention is clearest. to me, it feels like `registryproxy` or `registrycache` may be easier to understand, but I won't block on that.

pkg/test/framework/components/registryredirector/registryredirector.go
ingwonsong: That is the reason why I decided to rename it.

This component does not do proxying, but just redirecting to the real registry.

The request for getting the docker manifest will be redirected.

pkg/test/framework/components/registryredirector/kube.go
ingwonsong: As I replied in the other file, this component is not a proxy server, but just doing redirection to the target registry.

pkg/test/framework/components/registryredirector/registryredirector.go
ingwonsong: Changed to `Image`.

stewartbutler (approved):

Issue #38785, opened 2022-05-07
when the istiod sending xds then update service port name will the EnvoyEndpoint  is null in endpointShardz


cobb-tx:
### Bug Description

when the istiod sending xds then update service port name will the EnvoyEndpoint  is null in endpointShardz

the old service  ports name is tcp-80   update to http-80 or the old service  ports name is http-80   update to tcp-80

when istiod sending  fisrt xds  the pushcontext is tcp-80
and now i update service port name to http-80 .
the EndpointShards updateing in the same time  but pushcontext is old  so 
svcPort.Name (tcp-80) != ep.ServicePortName (http-80)
https://github.com/istio/istio/blob/master/pilot/pkg/xds/endpoint_builder.go#L285

manifestation：
access istio-ingress give 503 no healthy upstream to me

look this envoy /clusters  inf-test-java.testing00.svc.cluster.local not have Endpoints

```
outbound|80|v1|inf-test-java.testing00.svc.cluster.local::observability_name::outbound|80|v1|inf-test-java.testing00.svc.cluster.local
outbound|80|v1|inf-test-java.testing00.svc.cluster.local::default_priority::max_connections::4294967295
outbound|80|v1|inf-test-java.testing00.svc.cluster.local::default_priority::max_pending_requests::4294967295
outbound|80|v1|inf-test-java.testing00.svc.cluster.local::default_priority::max_requests::4294967295
outbound|80|v1|inf-test-java.testing00.svc.cluster.local::default_priority::max_retries::4294967295
outbound|80|v1|inf-test-java.testing00.svc.cluster.local::high_priority::max_connections::1024
outbound|80|v1|inf-test-java.testing00.svc.cluster.local::high_priority::max_pending_requests::1024
outbound|80|v1|inf-test-java.testing00.svc.cluster.local::high_priority::max_requests::1024
outbound|80|v1|inf-test-java.testing00.svc.cluster.local::high_priority::max_retries::3
outbound|80|v1|inf-test-java.testing00.svc.cluster.local::added_via_api::true
sds-grpc::observability_name::sds-grpc
sds-grpc::default_priority::max_connections::1024
sds-grpc::default_priority::max_pending_requests::1024
sds-grpc::default_priority::max_requests::1024
sds-grpc::default_priority::max_retries::3
sds-grpc::high_priority::max_connections::1024
sds-grpc::high_priority::max_pending_requests::1024
sds-grpc::high_priority::max_requests::1024
[...]

PR #38776, opened 2022-05-06
feat：checkLbEndpoint

cobb-tx:
Signed-off-by: taoxin 464693850@qq.com

**Please provide a description of this PR:**

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ * ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [ x] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #38843, opened 2022-05-11
fix Service Port Name changce Signed-off-by: taoxin 464693850@qq.com

cobb-tx:

fix https://github.com/istio/istio/issues/38785 

- [ ] Docs
- [ ] Installation
- [  x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/endpoint_builder.go
+++ b/pilot/pkg/xds/endpoint_builder.go
@@ -282,7 +282,7 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
 			if !ep.IsDiscoverableFromProxy(b.proxy) {
 				continue
 			}
-			if svcPort.Name != ep.ServicePortName {
+			if svcPort.Port != int(ep.EndpointPort) {
 				continue
 			}
 			// Port labels


cobb-tx:
and can turn off https://github.com/istio/istio/pull/38776

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -282,7 +282,7 @@ func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
 			if !ep.IsDiscoverableFromProxy(b.proxy) {
 				continue
 			}
-			if svcPort.Name != ep.ServicePortName {
+			if svcPort.Port != int(ep.EndpointPort) {
```
hzxuzhonghu, line 285: This is not right, svc port can be different from endpoint port

cobb-tx: yes your are right。 i test find that

cobb-tx: Do you have any idea to fix the bug？

hzxuzhonghu: This is a good catcha, will investigate

cobb-tx: i find a way to fix the bug？ still hava 503 for a little while .send next xds is ok.

PR #38845, opened 2022-05-11
Helm extraargs vol pilot

zadiman:

- These parameters are needed to load a cert-manager generated tls certificate at runtime
- This is a requirement for istio-csr installation
- While it's possible to use the override mechanism in IstioOperator some setups don't have such a facility
resolves
https://github.com/istio/istio/issues/36916


operator/pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -706,6 +706,9 @@ message PilotConfig {
   string hub = 34;
 
   google.protobuf.Value tag = 35;
+  google.protobuf.ListValue extraArgs = 38;
```
zadiman, line 709: Are we using the correct types here?

SpectralHiss:
Duplicate of https://github.com/istio/istio/pull/37264 , I am happy to continue here, I just want to see the code merged..

zadiman:
> Duplicate of #37264 , I am happy to continue here, I just want to see the code merged..

morvencao (commented):
/ok-to-test

zadiman:
> /ok-to-test

Is there anything more I need to do here? Or should I just relax and wait?

costinm (commented):
What 'extraArgs' do you need for pilot ? I think current practice is to use MeshConfig ( reloadable ) or env variables for new features.  I hope to see the CLI gradually replaced with MeshConfig or proper setttings. 

Re. volumes - for pods we use a slightly different pattern. I'm not sure what is the common practice in help templates - can you check few popular charts and what they use ? Also this fits the pattern of 'patching' the config - could a 'post renderer' or kustomize be used instead ?

costinm (commented):
I think the volumes are common enough and we should add them, after some research on what is the common naming pattern/style in helm for adding volumes. 

I am mostly against extraArgs - if one of the existing args is needed, maybe add it as first class, but again it would be far better to change the code and move it to MeshConfig (or add an env to replace the CLI).

zadiman:
> What 'extraArgs' do you need for pilot ? I think current practice is to use MeshConfig ( reloadable ) or env variables for new features. I hope to see the CLI gradually replaced with MeshConfig or proper setttings.
> 
> Re. volumes - for pods we use a slightly different pattern. I'm not sure what is the common practice in help templates - can you check few popular charts and what they use ? Also this fits the pattern of 'patching' the config - could a 'post renderer' or kustomize be used instead ?

extraArgs for my use case we need --tlsCertFile=, --tlsKeyFile= and --caCertFile=. This we use together with [istio-csr](https://github.com/cert-manager/istio-csr). If you leverage the Helm installation, and not the operator, you do not get overlays for example. It's quite broad use case with the dynamic ending of each argument but maybe try to scope it a bit more to only cater for this three certificate specific args?

SpectralHiss:
> I think the volumes are common enough and we should add them, after some research on what is the common naming pattern/style in helm for adding volumes.
> 
> I am mostly against extraArgs - if one of the existing args is needed, maybe add it as first-class, but again it would be far better to change the code and move it to MeshConfig (or add an env to replace the CLI).

Hey, thanks for the review, I totally understand your point, it's opening a big surface, it was meant to set the existing CLI args for pilot serving TLS context as @zadiman explained.

I would need some more pointers about how to add to "MeshConfig API" if it's the only acceptable way forward. 
However, I can see a simple implementation of passing pilotCertFile, PilotKeyFile and PilotcaCertFile as helm values and only setting those args specifically instead of having a blank extraArgs being a good compromise?

I do recognise that this might conflict for example with mesh config/dynamic root of trust perhaps but users of these features usually accept that a restart of the pilot is required to reload such config at present?

This PR is a clone from the original changes, I will ping @zadiman of upcoming amends or move back to [the original PR](https://github.com/istio/istio/pull/37264)

costinm:
My suggestion for the 3 args is to make a small code change. We need to
keep them for some time, for backward compat - but the general pattern for
certificate for workloads is:
- look for 'well known' path - same we are looking in agent
- have an env variable with the directory

I don't know if we need the actual file, the directory should be sufficient
- we have some clear name patterns we use for key/cert/root in secrets and
other places, but
if there is a clear use case of something requiring different naming
pattern we can incorporate it as well.

The reason I am pushing to avoid args is exactly this - even with Kustomize
it is hard to manipulate them.

Once you add the 'well known' path - you can also add a volume mount ( with
optional: true - if the secret doesn't exist it is ignored ), and the UX
will be zero config, if the user creates the Secret with Istiod certs
they'll be used automatically.

We can look for few locations - the env variable can be a list of
directories to check, to accomodate for CSI drivers or other tools that may
auto-provision certs.

costinm:
If your use case is just the tls certs - see my other email. It is clearly
not a dynamic thing - neither the volume mounts nor the cert location.

I don't like passing the 3 args - but I am ok with passing a
'istiodCertDir', and requiring the key/cert/ca follow the naming pattern we
use for
the mounted ingress certificates.  However I don't see a lot of value in
user picking a directory name - it would be much simpler to
have a 'well-known' directory, /var/run/secrets/istiod or something like
that - a nice thing about volume mounts is that we can make them optional.

SpectralHiss:
> If your use case is just the tls certs - see my other email. It is clearly
not a dynamic thing - neither the volume mounts nor the cert location.

> I don't like passing the 3 args - but I am ok with passing a
'istiodCertDir', and requiring the key/cert/ca follow the naming pattern we
use for
> the mounted ingress certificates.  However I don't see a lot of value in
user picking a directory name - it would be much simpler to
have a 'well-known' directory, /var/run/secrets/istiod or something like
that - a nice thing about volume mounts is that we can make them optional.

Okay, i think i get the idea but just to clarify the use-case here's how it could be loaded, i will show this just in order to show that there is some convention mismatch causing paths to be a little unconventional:


example
```yaml

  # below configs set Pilot serving tls context
  # likely to be used with extraVolumes and extraVolumeMounts
  
  # sets Pilot tls serving cert
  # tlsCertFile: /etc/cert-manager/tls/tls.crt
  tlsCertFile: ""
  
  # sets Pilot tls private key
  # tlsKeyFile: /etc/cert-manager/tls/tls.key
  tlsKeyFile: ""
  
  # sets Pilot Root of trust
  # caCertFile: /etc/cert-manager/ca/root-cert.pem
  caCertFile: ""

  # Defines extra volumes
  # eg to load istio-csr pilot tls cert and root:
  # extraVolumes:
  #  - name: "cert-manager",
  #    secret:
  #          secretName: "istiod-tls"
  #  - name: "ca-root-cert"
  #    configMap:
  #      name: "istio-ca-root-cert" 
  extraVolumes: []

  # Defines extra volume mounts, to be used with extraVolumes,
  # eg to load previous secret volume:
  # extraVolumeMounts:
  # - name: "ca-root-cert"
  #   mountPath: "/etc/cert-manager/ca"
  #   readOnly: true
  # - name: "cert-manager"
  #   mountPath: "/etc/cert-manager/tls"
  extraVolumeMounts: []
```

The format of cert-manager certificates which provision this cert is a secret with 
-> ca.crt
-> tls.crt
-> tls.key

because it's a bad TOFU pattern we tend not to use the ca.crt file, instead the correct pattern is to load a seperate configmap managed by istio-csr, which is the standard istio root configmap:
with key:
istio-ca-root-cert.

Just wanted to put all the facts on the table while i try to understand how to make this more implicit as you said (with optional true volumes)

zadiman:
Was tinkering on the same looking setup, depending on what we decide here, is the correct pattern to hold them separate or contain them under it's own mapping?

```  
  # Used to supply your own certificates. Example:
  #
  #   tlsCerts:
  #     tlsCertFile: "/server-certs/cert-chain.pem"
  #     tlsKeyFile: "/server-certs/key.pem"
  #     caCertFile: "/server-certs/root-cert.pem"
  #
  # See https://istio.io/latest/docs/reference/commands/pilot-discovery/#pilot-discovery-discovery
  tlsCerts:
    tlsCertFile: ""
    tlsKeyFile: ""
    caCertFile: ""
```
In the end I just want to be able to load certificates into our istio deployments. I am up for any solution - move back to the original PR, continue here or close this all together and move in an other direction.

SpectralHiss:
> Was tinkering on the same looking setup, depending on what we decide here, is the correct pattern to hold them separate or contain them under it's own mapping?
> 
> ```
>   # Used to supply your own certificates. Example:
>   #
>   #   tlsCerts:
>   #     tlsCertFile: "/server-certs/cert-chain.pem"
>   #     tlsKeyFile: "/server-certs/key.pem"
>   #     caCertFile: "/server-certs/root-cert.pem"
>   #
>   # See https://istio.io/latest/docs/reference/commands/pilot-discovery/#pilot-discovery-discovery
>   tlsCerts:
>     tlsCertFile: ""
>     tlsKeyFile: ""
>     caCertFile: ""
> ```
> 
> In the end I just want to be able to load certificates into our istio deployments. I am up for any solution - move back to the original PR, continue here or close this all together and move in an other direction.

Yeah i don't mind either, but i think we want to go for like a "zeroconf" approach, where we change the pilot code to read for example /etc/server-certs/ always if it exists (currently checking what the existing behaviour is), we wouldn't even need extraVolumes etc.. (although i see no harm) because we can add volumes and volumemounts as optional and have it all work automagically if such pilot secret/configmap exist.

SpectralHiss:
I am having a look currently at the existing behaviour in regards to well known paths for the pilot cert:
If not using a custom or unidentified cert provider.
1) The [top level server initialisation logic for the ca](ttps://github.com/istio/istio/blob/36f988f68a40fb9aecbf84e0004805770fa9afaa/pilot/pkg/bootstrap/server.go#L266) doesn't try to generate an istiod cert if it doesn't have customTLSCerts or provider method is None, the predicate [hasCustomTLSCerts](https://github.com/istio/istio/blob/36f988f68a40fb9aecbf84e0004805770fa9afaa/pilot/pkg/bootstrap/server.go#L944) is actually basing this on the existence of the tlsoptions flags.

-> Therefore it seems that the current design assumed that custom tls certs which are different from the plugin `cacerts` method would be loaded using those flags only.

So we would need to either do the above approach or change the predicate to both check for flags and for a well known location, one possibility is to actually consider pilot as another "workload" and use other defaults we have of WorkloadIdentityKeyPath and WorkloadIdentityCertChainPath which are under ./var/run/secrets/workload-spiffe-credentials  or define other set of constants, or set PILOT_SERVING_TLS_DIR through env?

Alternatively we could accept the above solution, with @zadiman refinement on key map potentially..

costinm:
On Mon, May 30, 2022 at 7:54 AM Houssem El Fekih ***@***.***>
wrote:

> If your use case is just the tls certs - see my other email. It is clearly
> not a dynamic thing - neither the volume mounts nor the cert location.
>
> I don't like passing the 3 args - but I am ok with passing a
> 'istiodCertDir', and requiring the key/cert/ca follow the naming pattern we
> use for
> the mounted ingress certificates. However I don't see a lot of value in
> user picking a directory name - it would be much simpler to
> have a 'well-known' directory, /var/run/secrets/istiod or something like
> that - a nice thing about volume mounts is that we can make them optional.
>
> Okay, i think i get the idea but just to clarify the use-case here's how
> it could be loaded, i will show this just in order to show that there is
> some convention mismatch causing paths to be a little unconventional:
>
> example
>
>   # below configs set Pilot serving tls context
>   # likely to be used with extraVolumes and extraVolumeMounts
>
>   # sets Pilot tls serving cert
>   # tlsCertFile: /etc/cert-manager/tls/tls.crt
>   tlsCertFile: ""
>
>   # sets Pilot tls private key
>   # tlsKeyFile: /etc/cert-manager/tls/tls.key
>   tlsKeyFile: ""
>
>   # sets Pilot Root of trust
>   # caCertFile: /etc/cert-manager/ca/root-cert.pem
>   caCertFile: ""
>
>   # Defines extra volumes
>   # eg to load istio-csr pilot tls cert and root:
>   # extraVolumes:
>   #  - name: "cert-manager",
>   #    secret:
>   #          secretName: "istiod-tls"
>   #  - name: "ca-root-cert"
>   #    configMap:
>   #      name: "istio-ca-root-cert"
>   extraVolumes: []
>
>   # Defines extra volume mounts, to be used with extraVolumes,
>   # eg to load previous secret volume:
>   # extraVolumeMounts:
>   # - name: "ca-root-cert"
>   #   mountPath: "/etc/cert-manager/ca"
>   #   readOnly: true
>   # - name: "cert-manager"
>   #   mountPath: "/etc/cert-manager/tls"
>   extraVolumeMounts: []
>
> The format of cert-manager certificates which provision this cert is a
> secret with
> -> ca.crt
> -> tls.crt
> -> tls.key
>
> because it's a bad TOFU pattern we tend not to use the ca.crt file,
> instead the correct pattern is to load a seperate configmap managed by
> istio-csr, which is the standard istio root configmap:
> with key:
> istio-ca-root-cert.
>
> Just wanted to put all the facts on the table while i try to understand
> how to make this more implicit as you said (with optional true volumes)
>

Yes, there is some variation - the '/var/run/secrets/workload-certificate'
has yet another naming convention.
It is reasonable, for interoperability, to support multiple patterns and
check different locations - this is far
easier and cleaner in code, without forcing the user to have complex
config.

I think the recommended/default pattern should match the Secrets we use for
Gateway. Ideally
Istiod would not be exposed directly, but via a Gateway - which would load
the cert via SDS, just like
all other certs.  So whatever we do for Gateway - we should do for Istiod.

One thing I want to avoid is inventing a NEW pattern or config just for
this - if a pattern is already
in use - or in use by a broadly used system ( CertManager, Vault, GKE, etc
) - I'm all for interop
and integration, and we can add code ( but not complex config ) to detect
and adapt.

costinm:
On Tue, May 31, 2022 at 3:37 AM Houssem El Fekih ***@***.***>
wrote:

> I am having a look currently at the existing behaviour in regards to well
> known paths for the pilot cert:
> If not using a custom or unidentified cert provider.
>
>    1. The top level server initialisation logic for the ca doesn't try to
>    generate an istiod cert if it doesn't have customTLSCerts or provider
>    method is None, the predicate hasCustomTLSCerts
>    <https://github.com/istio/istio/blob/36f988f68a40fb9aecbf84e0004805770fa9afaa/pilot/pkg/bootstrap/server.go#L944>
>    is actually basing this on the existence of the tlsoptions flags.
>
> Yes, I think we need to fix this, just like we did for the agent.

And as I mentioned - I think the 'best practice' ( or at least a supported
deployment ) would be to have a proper Gateway in
front of Istio, with real DNS certificates ( and additional DOS/LB
capabilities).



>
>
> -> Therefore it seems that the current design assumed that custom tls
> certs which are different from the plugin cacerts method would be loaded
> using those flags only.
>
Or support for custom certs was added without a lot of design, or before we
made the changes to agent or realized the implications :-)

Unfortunately we may need to continue to support the flags for a long time,
for backward compat - but we can deprecate them and hide from docs.


> So we would need to either do the above approach or change the predicate
> to both check for flags and for a well known location, one possibility is
> to actually consider pilot as another "workload" and use other defaults we
> have of WorkloadIdentityKeyPath and WorkloadIdentityCertChainPath which are
> under ./var/run/secrets/workload-spiffe-credentials or define other set of
> constants, or set PILOT_SERVING_TLS_DIR through env?
>

Can I ask why do we even need to mount them, if they are in a Secret ? We
have code to read all secrets and serve them to gateways.
If they are not provided by a Secret ( but CSI/etc) - yes, I think we
should use the same pattern of a well-known path.
We may look back to older versions of Istio - we had this in the past (
/etc/istio and other locations ).

The workload-spieff-credentials is an interesting problem - IMO in a
separate PR we should detect if they are present, and if they
are start using it - possibly on 15011 ( the original Istiod port using
spiffee certs - we turned it off at some point ). Agent still has capability
to configure expected SAN for istiod - and if user has a platform-wide
workload cert we can use it for Istiod as well ( except webhooks ),
so the ns controller replicating istiod root is not needed.

zadiman:
Made the changes discussed above, incorporated your comments, and rebased. This works and is not as generic as before. Any comments?

SpectralHiss:
This is issue is not stale, cf the work in the related PR, which is almost ready to review: https://github.com/istio/istio/pull/39375

PR #38848, opened 2022-05-11
style: remove unused code

drivebyer:

This PR remove unused code.

### DIFF
--- a/pkg/util/sets/string.go
+++ b/pkg/util/sets/string.go
@@ -174,18 +174,3 @@ func (s Set) Len() int {
 func (s Set) IsEmpty() bool {
 	return len(s) == 0
 }
-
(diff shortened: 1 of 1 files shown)


howardjohn:
Is there more context for this? This is intended to use slice to avoid extra allocations of sets. Also I don't see how this can compile without changing call sites?

drivebyer:
> Is there more context for this? This is intended to use slice to avoid extra allocations of sets.

hi @howardjohn , in my view, if the return value is a slice, there is still a memory allocation. Maybe the return type of `Set` can provide more methods for caller for lating use.

>  Also I don't see how this can compile without changing call sites?

Actually, this method has no caller.

howardjohn:
I think we can just delete it then. The original PR adding it ended up just using Difference, I think it was merged in mistake

ericvn (approved):
LGTM

PR #38836, opened 2022-05-10
Feat/hpa memory

andrewhertog:
**Please provide a description of this PR:**
Add's the ability to scale on memory in the gateway chart. 

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [x] Configuration Infrastructure
- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [x] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #38856, opened 2022-05-11
feat: add deployment update strategy to gateway

andrewhertog:
Add's configurable deployment update strategy to the gateway helm chart

### DIFF
--- a/manifests/charts/gateway/templates/deployment.yaml
+++ b/manifests/charts/gateway/templates/deployment.yaml
@@ -11,6 +11,10 @@ spec:
   {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.replicaCount }}
   {{- end }}
+  {{- if .Values.deploymentStrategy }}
+  strategy:
+    {{ .Values.deploymentStrategy | toYaml | nindent 4 }}
+  {{- end }}
   selector:
     matchLabels:
       {{- include "gateway.selectorLabels" . | nindent 6 }}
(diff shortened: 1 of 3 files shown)


howardjohn (commented):
This seems like a pretty niche feature. We cannot possibly expose every field in the API. Maybe something like https://docs.solo.io/gloo-edge/latest/installation/gateway/kubernetes/helm_advanced/ is more suitable for this

andrewhertog:
> This seems like a pretty niche feature. We cannot possibly expose every field in the API. Maybe something like https://docs.solo.io/gloo-edge/latest/installation/gateway/kubernetes/helm_advanced/ is more suitable for this

So we are switching from using the ingress-gateway in the IstioOperator crd to use this chart as that is what is recommended in https://istio.io/latest/docs/setup/additional-setup/gateway/#deploying-a-gateway under the helm tab. We want to try step away from using the IstioOperator as it has been noted:
> Use of the operator for new Istio installations is discouraged in favor of the [Istioctl](https://istio.io/latest/docs/setup/install/istioctl) and [Helm](https://istio.io/latest/docs/setup/install/helm) installation methods. While the operator will continue to be supported, new feature requests will not be prioritized.

We are unable to use Istioctl to deploy as we are currently using flux v1 to bootstrap our clusters. This also prevents us from using the `--post-render` helm command.

If this feature isn't supported in this chart then a 1:1 migration path from the IstioOperator ingress-gateway to the gateway helm chart is not possible in this case as we have utilized this, and other features that i've created PR's for (https://github.com/istio/istio/pull/38849, https://github.com/istio/istio/pull/38836)

PR #38892, opened 2022-05-12
Dual stack intra-mesh integ tests

Kmoneal:
Integration tests to validate dual-stack intra-mesh communication. Mainly focusing on IPv4 to IPv6 and IPv6 to IPv4 traffic. These two tests are skipped for now but should not be in the future.


Kmoneal:
Step for https://github.com/istio/istio/issues/37533

pkg/test/framework/suite.go
```diff
@@ -302,6 +307,39 @@ func (s *suiteImpl) RequireMaxVersion(minorVersion uint) Suite {
 	return s
 }
 
+func (s *suiteImpl) RequireDualStackCluster() Suite {
+	fn := func(ctx resource.Context) error {
+		for _, c := range ctx.Clusters().Kube() {
```
stevenctl, line 312: super nit: can do this in the Environment setup as we create kube clusters _or_ cache this value on the `Cluster` struct somehow, just to avoid adding warmup time to each suite, even though it's probably very minimal.

tests/integration/dualstack/common/apps.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+var deploymentConfigs = []DeploymentConfig{
+	{Name: "ipv4", Service: "echo-ipv4", IPFamily: "IPv4", IPFamilyPolicy: "SingleStack"},
+	{Name: "ipv6", Service: "echo-ipv6", IPFamily: "IPv6", IPFamilyPolicy: "SingleStack"},
+	{Name: "dual-ipv4", Service: "echo-dual-ipv4", IPFamily: "IPv4, IPv6", IPFamilyPolicy: "RequireDualStack"},
+	{Name: "dual-ipv6", Service: "echo-dual-ipv6", IPFamily: "IPv6, IPv4", IPFamilyPolicy: "RequireDualStack"},
+}
+
+var EchoPorts = []echo.Port{
```
stevenctl, line 62: Can we re-use (a portion) of common echo deployments from other suites?

tests/integration/dualstack/common/traffic.go
```diff
@@ -0,0 +1,214 @@
...
+	// Single call. Cannot be used with children.
+	call func(t test.Failer, options echo.CallOptions) echo.CallResult
+	// opts specifies the echo call options. When using RunForApps, the Target will be set dynamically.
+	opts echo.CallOptions
+
+	// setting cases to skipped is better than not adding them - gives visibility to what needs to be fixed
+	skip bool
+}
+
+func (c TrafficTestCase) Run(ctx framework.TestContext, namespace string) {
```
stevenctl, line 56: For the most part, I think `tests/integration/pilot/common/` can support this. If there is a reason we re-implement it here, can you add a comment explaining why?

howardjohn (commented):
Should this be on the IPv6 experimental branch? Or on master?

tests/integration/dualstack/common/traffic.go
Kmoneal: For the most part it's the same. We do want to verify proxies are not sending http requests at the same time which is why I have a similar function: https://github.com/istio/istio/pull/38892/files#diff-8d1061ab02227bf02916079cac53e632843d46dcd82ce42a24e7ef20eeec6f89R75

Kmoneal:
> Should this be on the IPv6 experimental branch? Or on master?

Yep, I'll move it over to experimental-dual-stack

tests/integration/dualstack/common/apps.go
Kmoneal: I'll do that here. We did have our own version for other tests that we haven't upstreamed yet but for this case it makes sense to delay adding this.

howardjohn:
@leosarra might be something to look into to draw inspiration from

PR #38893, opened 2022-05-13
enabled reachability tests in multi-cluster env

aryan16:
**Enabled Reachability tests in multi-cluster environment**

### DIFF
--- a/tests/integration/security/reachability_test.go
+++ b/tests/integration/security/reachability_test.go
@@ -84,12 +84,15 @@ func TestReachability(t *testing.T) {
 					ExpectMTLS:    mtlsOnExpect,
 				},
 				{
-					ConfigFile:             "beta-mtls-off.yaml",
-					Namespace:              systemNM,
-					Include:                Always,
-					ExpectSuccess:          Always,
-					ExpectMTLS:             Never,
-					SkippedForMulticluster: true,
+					ConfigFile:    "beta-mtls-off.yaml",
+					Namespace:     systemNM,
+					Include:       Always,
+					ExpectSuccess: Always,
+					ExpectMTLS:    Never,
+					ExpectDestinations: func(from echo.Instance, to echo.Target) echo.Instances {
+						// Without TLS we can't perform SNI routing required for multi-network
+						return match.Network(from.Config().Cluster.NetworkName()).GetMatches(to.Instances())
+					},
 				},
 				{
 					ConfigFile: "beta-mtls-automtls-workload.yaml",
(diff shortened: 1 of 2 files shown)


tests/integration/security/reachability_test.go
```diff
@@ -112,15 +115,32 @@ func TestReachability(t *testing.T) {
...
-					Include:                Always,
-					ExpectSuccess:          Always,
-					ExpectMTLS:             Never,
-					SkippedForMulticluster: true,
+					ConfigFile:    "plaintext-to-permissive.yaml",
+					Namespace:     systemNM,
+					Include:       Always,
+					ExpectSuccess: Always,
+					ExpectMTLS:    Never,
+					ExpectDestinations: func(from echo.Instance, to echo.Target) echo.Instances {
```
howardjohn, line 139: nit: abstract these so its like `ExpectDestinations: SameNetwork` or similar

tests/integration/security/util/reachability/context.go
```diff
@@ -166,7 +165,9 @@ func Run(testCases []TestCase, t framework.TestContext, apps *util.EchoDeploymen
 								// Set the target on the call options.
 								opts.To = to
 								opts.Count = callCount
-
+								if c.ExpectReached != nil && !c.ExpectReached(opts) {
```
howardjohn, line 168: A bit confused by behavior of ExpectReached and the existing ExpectDestinations.

ExpectDestinations sounds like its changing the test _assertion_ - but its not... its actually changing where we send. It seems like this should be `FilterDestinations` (or just `Include` and just add `to` to that). This is existing code, but seems related

ExpectReached sounds like it is changing the _assertion_ as well (actually sounds identical to ExpectDestinations to me), but its not -- its skipping tests. Should this either just use `Include` or make it so we don't skip the test but rather run it and do not assert that all destinations are reached?

stevenctl: I thought ExpectDestinations skipped tests and ExpectedReached changed assertions. ExpectDestinations can probably just be handled in `Include`. ExpectReached has to have the `call opts` because expectations change for things like per-port settings.

stevenctl: Oh, misread your comment. We're on the same page.

aryan16: I agree with you John, I updated the code and added the logic of `ExpectedReached` inside `Include`.

tests/integration/security/reachability_test.go
```diff
@@ -84,19 +87,20 @@ func TestReachability(t *testing.T) {
...
+					// Without TLS we can't perform SNI routing required for multi-network
+					ExpectDestinations: SameNetwork,
 				},
 				{
 					ConfigFile: "beta-mtls-automtls-workload.yaml",
 					Namespace:  apps.Namespace1,
 					Include: func(from echo.Instance, opts echo.CallOptions) bool {
 						return (apps.B.Contains(from) || apps.IsNaked(from)) &&
-							(apps.A.ContainsTarget(opts.To) || apps.B.ContainsTarget(opts.To))
+							(apps.A.ContainsTarget(opts.To) || apps.B.ContainsTarget(opts.To)) && !(apps.B.ContainsTarget(opts.To) && opts.Port.Name == "http")
```
howardjohn, line 103: What is the significance of `http` here? The whole function here is pretty opaque, could use some comments on what its trying to do

howardjohn (approved):

howardjohn:
looking into tests seeming slow. I think its not caused by this PR but want to double check since this one  took 50min

PR #38899, opened 2022-05-13
Sds cleanup

hzxuzhonghu:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/credentials/kube/secrets.go
+++ b/pilot/pkg/credentials/kube/secrets.go
@@ -27,6 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	sa "k8s.io/apiserver/pkg/authentication/serviceaccount"
 	informersv1 "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/fake"
@@ -64,8 +65,6 @@ type CredentialsController struct {
 	secretLister   listersv1.SecretLister
 	sar            authorizationv1client.SubjectAccessReviewInterface
 
-	clusterID cluster.ID
-
 	mu                 sync.RWMutex
 	authorizationCache map[authorizationKey]authorizationResponse
 }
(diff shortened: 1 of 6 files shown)


pilot/pkg/credentials/kube/secrets.go
```diff
@@ -100,19 +96,13 @@ func NewCredentialsController(client kube.Client, clusterID cluster.ID) *Credent
...
-
+		secretInformer:     informer,
+		secretLister:       listersv1.NewSecretLister(informer.GetIndexer()),
 		sar:                client.Kube().AuthorizationV1().SubjectAccessReviews(),
-		clusterID:          clusterID,
 		authorizationCache: make(map[authorizationKey]authorizationResponse),
 	}
 }
 
-func toUser(serviceAccount, namespace string) string {
```
hzxuzhonghu, line 112: replace with k8s lib

pilot/pkg/credentials/kube/secrets.go
```diff
@@ -156,19 +146,8 @@ func (s *CredentialsController) insertCache(user string, response error) {
 	}
 }
 
-// DisableAuthorizationForTest makes the authorization check always pass. Should be used only for tests.
-func DisableAuthorizationForTest(fake *fake.Clientset) {
```
hzxuzhonghu, line 160: move to test, which uses it

ramaraochavali (approved):

pilot/pkg/credentials/kube/secrets.go
```diff
@@ -156,19 +146,8 @@ func (s *CredentialsController) insertCache(user string, response error) {
...
-			Status: authorizationv1.SubjectAccessReviewStatus{
-				Allowed: true,
-			},
-		}, nil
-	})
-}
-
 func (s *CredentialsController) Authorize(serviceAccount, namespace string) error {
-	user := toUser(serviceAccount, namespace)
+	user := sa.MakeUsername(namespace, serviceAccount)
```
howardjohn, line 150: I am not sure we want to bring in a complete new dependency to replace 1 line of code?

hzxuzhonghu: This is a good question, I am not sure when I am doing this.

But after, the package provides many utilities, istio could use for authz

howardjohn (approved):

pilot/pkg/credentials/kube/secrets.go
```diff
@@ -26,13 +26,11 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
-	"k8s.io/apimachinery/pkg/runtime"
+	sa "k8s.io/apiserver/pkg/authentication/serviceaccount"
```
costinm, line 29: Should we depend on apiserver ?

costinm (commented):
We need to be more careful with the dependencies, not sure apiserver package is meant as a library ( i.e. api stability, etc), and we are already having too many deps.

pilot/pkg/credentials/kube/secrets.go
hzxuzhonghu: It is not k/k, this lib is aimed to be used

costinm:
You're right, all good.

BTW - looking at go.mod in the library - it appears they already switched
to opentelemetry for the instrumentation, we should discuss a plan to do
the same, since
the library we use (opencensus) is mostly deprecated.

PR #38901, opened 2022-05-13
fix incorect sds logs

ramaraochavali:
initWorkloadSdsService checks for uds path async and before it returns we log the message "SDS server started". This PR fixes that along with other minor refactor.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [x] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/istio-agent/agent.go
+++ b/pkg/istio-agent/agent.go
@@ -505,16 +505,13 @@ func (a *Agent) initSdsServer() error {
 		a.secOpts.KeyFilePath = security.WorkloadIdentityKeyPath
 
 		a.secretCache, err = cache.NewSecretManagerClient(nil, a.secOpts)
-		if err != nil {
-			return fmt.Errorf("failed to start workload secret manager %v", err)
-		}
 	} else {
 		a.secretCache, err = a.newSecretManager()
-		if err != nil {
-			return fmt.Errorf("failed to start workload secret manager %v", err)
-		}
 	}
 
+	if err != nil {
+		return fmt.Errorf("failed to start workload secret manager %v", err)
+	}
 	pkpConf := a.proxyConfig.GetPrivateKeyProvider()
 	a.sdsServer = sds.NewServer(a.secOpts, a.secretCache, pkpConf)
 	a.secretCache.RegisterSecretHandler(a.sdsServer.OnSecretUpdate)
--- a/pkg/istio-agent/agent.go
+++ b/pkg/istio-agent/agent.go
@@ -494,11 +494,7 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
 
 func (a *Agent) initSdsServer() error {
 	var err error
-	// If proxy is using file mounted certs, we do not have to connect to CA.
-	if a.secOpts.FileMountedCerts {
-		log.Info("Workload is using file mounted certificates. Skipping connecting to CA")
-		a.secretCache, err = cache.NewSecretManagerClient(nil, a.secOpts)
-	} else if security.CheckWorkloadCertificate(security.WorkloadIdentityCertChainPath, security.WorkloadIdentityKeyPath, security.WorkloadIdentityRootCertPath) {
+	if security.CheckWorkloadCertificate(security.WorkloadIdentityCertChainPath, security.WorkloadIdentityKeyPath, security.WorkloadIdentityRootCertPath) {
 		log.Info("workload certificate files detected, creating secret manager without caClient")
 		a.secOpts.RootCertFilePath = security.WorkloadIdentityRootCertPath
 		a.secOpts.CertChainFilePath = security.WorkloadIdentityCertChainPath
@@ -796,6 +792,11 @@ func getKeyCertInner(certPath string) (string, string) {
 
 // newSecretManager creates the SecretManager for workload secrets
 func (a *Agent) newSecretManager() (*cache.SecretManagerClient, error) {
+	// If proxy is using file mounted certs, we do not have to connect to CA.
+	if a.secOpts.FileMountedCerts {
+		log.Info("Workload is using file mounted certificates. Skipping connecting to CA")
+		return cache.NewSecretManagerClient(nil, a.secOpts)
+	}
 	log.Infof("CA Endpoint %s, provider %s", a.secOpts.CAEndpoint, a.secOpts.CAProviderName)
 
 	// TODO: this should all be packaged in a plugin, possibly with optional compilation.
--- a/pkg/istio-agent/agent.go
+++ b/pkg/istio-agent/agent.go
@@ -494,7 +494,11 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
 
 func (a *Agent) initSdsServer() error {
 	var err error
-	if security.CheckWorkloadCertificate(security.WorkloadIdentityCertChainPath, security.WorkloadIdentityKeyPath, security.WorkloadIdentityRootCertPath) {
+	// If proxy is using file mounted certs, we do not have to connect to CA.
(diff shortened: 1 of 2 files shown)


pkg/istio-agent/agent.go
```diff
@@ -494,7 +494,11 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
 
 func (a *Agent) initSdsServer() error {
 	var err error
-	if security.CheckWorkloadCertificate(security.WorkloadIdentityCertChainPath, security.WorkloadIdentityKeyPath, security.WorkloadIdentityRootCertPath) {
+	// If proxy is using file mounted certs, we do not have to connect to CA.
```
ramaraochavali, line 497: moved this out so that newSecretManager can solely focus on CA flow

security/pkg/nodeagent/sds/server.go
```diff
@@ -48,7 +48,6 @@ func NewServer(options *security.Options, workloadSecretCache security.SecretMan
 	s := &Server{stopped: atomic.NewBool(false)}
 	s.workloadSds = newSDSService(workloadSecretCache, options, pkpConf)
 	s.initWorkloadSdsService()
-	sdsServiceLog.Infof("SDS server for workload certificates started, listening on %q", security.WorkloadIdentitySocketPath)
```
ramaraochavali, line 51: This log comes even there is a problem with SDS server

security/pkg/nodeagent/sds/server.go
```diff
@@ -85,13 +84,7 @@ func (s *Server) Stop() {
 func (s *Server) initWorkloadSdsService() {
 	s.grpcWorkloadServer = grpc.NewServer(s.grpcServerOptions()...)
 	s.workloadSds.register(s.grpcWorkloadServer)
-
 	var err error
-	s.grpcWorkloadListener, err = uds.NewListener(security.WorkloadIdentitySocketPath)
```
ramaraochavali, line 90: This seems redundant?

ramanujadasu (approved):

GregHanson (approved):

pkg/istio-agent/agent.go
```diff
@@ -494,23 +494,24 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
...
+	if a.secOpts.FileMountedCerts {
```
lei-tang, line 498: This "if ... else if" changes the program behavior, not just fixes logs. This "if ... else if" leads to the following problem: when workload certificate is used, FileMountedCerts can be true, which leads to the following three variables not being set.
                 a.secOpts.RootCertFilePath = security.WorkloadIdentityRootCertPath
		a.secOpts.CertChainFilePath = security.WorkloadIdentityCertChainPath
		a.secOpts.KeyFilePath = security.WorkloadIdentityKeyPath

The correct behavior is to first CheckWorkloadCertificate(security.WorkloadIdentityCertChainPath, security.WorkloadIdentityKeyPath, security.WorkloadIdentityRootCertPath), then check a.secOpts.FileMountedCerts:
if security.CheckWorkloadCertificate(security.WorkloadIdentityCertChainPath, security.WorkloadIdentityKeyPath, security.WorkloadIdentityRootCertPath) {
		log.Info("workload certificate files detected, creating secret manager without caClient")
		a.secOpts.RootCertFilePath = security.WorkloadIdentityRootCertPath
		a.secOpts.CertChainFilePath = security.WorkloadIdentityCertChainPath
		a.secOpts.KeyFilePath = **security.WorkloadIdentityKeyPath**
                 ....
} else if a.secOpts.FileMountedCerts {
....
}

pkg/istio-agent/agent.go
```diff
@@ -494,23 +494,24 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
...
+		log.Info("Workload is using file mounted certificates. Skipping connecting to CA")
+		a.secretCache, err = cache.NewSecretManagerClient(nil, a.secOpts)
+	} else if security.CheckWorkloadCertificate(security.WorkloadIdentityCertChainPath, security.WorkloadIdentityKeyPath, security.WorkloadIdentityRootCertPath) {
 		log.Info("workload certificate files detected, creating secret manager without caClient")
 		a.secOpts.RootCertFilePath = security.WorkloadIdentityRootCertPath
 		a.secOpts.CertChainFilePath = security.WorkloadIdentityCertChainPath
 		a.secOpts.KeyFilePath = security.WorkloadIdentityKeyPath
 
 		a.secretCache, err = cache.NewSecretManagerClient(nil, a.secOpts)
-		if err != nil {
```
lei-tang, line 504: Why delete this error checking?

pkg/istio-agent/agent.go
```diff
@@ -795,12 +796,6 @@ func getKeyCertInner(certPath string) (string, string) {
 
 // newSecretManager creates the SecretManager for workload secrets
 func (a *Agent) newSecretManager() (*cache.SecretManagerClient, error) {
-	// If proxy is using file mounted certs, we do not have to connect to CA.
```
lei-tang, line 798: Why delete this if a.secOpts.FileMountedCerts {...} and move it to initSdsServer()?

lei-tang (changes requested):

pkg/istio-agent/agent.go
lei-tang: The original code looks to be correct (i.e., first CheckWorkloadCertificate(), then check if a.secOpts.FileMountedCerts ). The description of this PR is to fix logs, not to change the workflow. Does the original workflow have bugs? Unless the original workflow has bugs, changing the original workflow does not seem to be necessary.

pkg/istio-agent/agent.go
ramaraochavali: moved below to cover for both if else

pkg/istio-agent/agent.go
ramaraochavali: I started with log changes only but thought it would be better to handle both fileMounted and checkWorkloadCertificate in the same manner at the top to avoid file checking unnecessarily for file mounted certs flow- Sorry I should have updated the PR.

I think it does not change behaviour because both are not typically used at the same time. But I do not have strong opinion on that. So reverted the change

adiprerepa (approved):

security/pkg/nodeagent/sds/server.go
```diff
@@ -85,13 +84,8 @@ func (s *Server) Stop() {
 func (s *Server) initWorkloadSdsService() {
 	s.grpcWorkloadServer = grpc.NewServer(s.grpcServerOptions()...)
 	s.workloadSds.register(s.grpcWorkloadServer)
-
 	var err error
 	s.grpcWorkloadListener, err = uds.NewListener(security.WorkloadIdentitySocketPath)
-	if err != nil {
```
lei-tang, line 91: Why is this err log deleted?

pkg/istio-agent/agent.go
lei-tang: Thanks! The PR looks much better now.

lei-tang:
@ramaraochavali Thanks for the PR and the revisions to the PR! I left one more comment. Please take a look.

security/pkg/nodeagent/sds/server.go
ramaraochavali: This NewListener call it self is only used in tests (to avoid timing issues) and the go routine below logs the same if it is the workloadListener is not setup. So this log line is redundant.

lei-tang (approved):
Thanks for the PR and the revisions! LGTM and approved.

PR #38903, opened 2022-05-13
remove remote profile and update precheck

### LINKED ISSUE #38832: Remove deprecated remote profile

The remote profile has been deprecated since 1.10 so we should remove it now.

Should also check for use of the remote profile in `istioctl x precheck`. Alternatively, maybe we should add a check in `istioctl install/generate` when it detects an error because a profile does not exist, if the profile is the `remote` profile generate a warning (instead of an error), and use the default profile instead (which is equivalent).

[ ] Docs
[x] Installation
[ ] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

[x] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

xichengliudui:
fix: https://github.com/istio/istio/issues/38832

* remove remote profile
* `istioctl x precheck` add remote profile check

- [x] Configuration Infrastructure
- [x] Installation

[] Does not have any changes that may affect Istio users.

### DIFF
--- a/istioctl/cmd/precheck.go
+++ b/istioctl/cmd/precheck.go
@@ -34,8 +34,6 @@ import (
 	"istio.io/istio/istioctl/pkg/clioptions"
 	"istio.io/istio/istioctl/pkg/install/k8sversion"
 	"istio.io/istio/istioctl/pkg/util/formatting"
-	"istio.io/istio/operator/pkg/apis/istio"
-	"istio.io/istio/operator/pkg/util"
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pkg/config/analysis"
 	"istio.io/istio/pkg/config/analysis/analyzers/maturity"
@@ -119,12 +117,6 @@ func checkControlPlane(cli kube.ExtendedClient) (diag.Messages, error) {
 
 	msgs = append(msgs, checkInstallPermissions(cli)...)
 
-	m, err = checkIOPProfile(cli)
-	if err != nil {
-		return nil, err
-	}
-	msgs = append(msgs, m...)
-
 	// TODO: add more checks
 
 	sa := local.NewSourceAnalyzer(analysis.Combine("upgrade precheck", &maturity.AlphaAnalyzer{}),
@@ -151,32 +143,6 @@ func checkControlPlane(cli kube.ExtendedClient) (diag.Messages, error) {
 	}
 
 	return msgs, nil
-}
-
-func checkIOPProfile(cli kube.ExtendedClient) (diag.Messages, error) {
-	iops, err := cli.Dynamic().Resource(istioOperatorGVR).
-		List(context.Background(), metav1.ListOptions{})
-	if err != nil {
-		return nil, fmt.Errorf("cannot retrieve IstioOperator CRs: %v", err)
-	}
-	for _, u := range iops.Items {
-		u.SetCreationTimestamp(metav1.Time{})
-		u.SetManagedFields([]metav1.ManagedFieldsEntry{})
-		iop, err := istio.UnmarshalIstioOperator(util.ToYAML(u.Object), true)
-		if err != nil {
-			return nil, fmt.Errorf("error while converting to IstioOperator CR - %s/%s: %v",
-				u.GetNamespace(), u.GetName(), err)
-		}
-		spec := fmt.Sprintf("%s", iop.Spec)
-		if spec == "" {
-			return []diag.Message{
-				msg.NewDeprecated(&resource.Instance{Origin: clusterOrigin{}},
-					fmt.Sprintf("%s is deprecated; use %s", "remote profile", "external profile")),
-			}, nil
-		}
-	}
-
-	return nil, nil
 }
 
 func checkInstallPermissions(cli kube.ExtendedClient) diag.Messages {
--- a/istioctl/cmd/precheck.go
+++ b/istioctl/cmd/precheck.go
@@ -34,6 +34,8 @@ import (
 	"istio.io/istio/istioctl/pkg/clioptions"
 	"istio.io/istio/istioctl/pkg/install/k8sversion"
 	"istio.io/istio/istioctl/pkg/util/formatting"
+	"istio.io/istio/operator/pkg/apis/istio"
+	"istio.io/istio/operator/pkg/util"
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pkg/config/analysis"
 	"istio.io/istio/pkg/config/analysis/analyzers/maturity"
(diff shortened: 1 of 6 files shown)


xichengliudui:
some test information
```
[root@test-build ~]# kubectl get iop -A
NAMESPACE      NAME                 REVISION     STATUS    AGE
istio-system   remote-profile                    HEALTHY   7h48m
[root@test-build ~]# ./istioctl x precheck
Warning [IST0002] (Cluster) Deprecated: remote profile is deprecated; use external profile
Error: Issues found when checking the cluster. Istio may not be safe to install or upgrade.
See https://istio.io/v1.15/docs/reference/config/analysis for more information about causes and resolutions.
```

ericvn:
This PR certainly needs a release note, removed the label.

istioctl/cmd/precheck.go
```diff
@@ -145,6 +153,32 @@ func checkControlPlane(cli kube.ExtendedClient) (diag.Messages, error) {
 	return msgs, nil
 }
 
+func checkIOPProfile(cli kube.ExtendedClient) (diag.Messages, error) {
```
howardjohn, line 156: do we want a precheck? Or just make a useful error message?

For example

currently:
```
$ istioctl manifest generate --set profile=remote
Error: failed to read profile foo from : open profiles/remote.yaml: file does not exist
```

after:
```
$ istioctl manifest generate --set profile=remote
Error: profile "remote" has been removed and is equivalent to "default"
```

Or just automatically translate remote to default?

frankbu: Don't automatically translate remote to default, because In the not-to-distant future we would like to reuse "remote" to be the "proper" remote profile, which is currently named "external" (which is a little confusing). When we re-introduce the remote profile (maybe in 1.16?) we will deprecate and eventually remove the external profile. Bottom line, is we would like to make sure that nobody continues to use the current incorrect remote profile. The nicer error message would be fine, but we will have to remove it when we reintroduce the proper remote profile.

It's not clear to me what will happen if users upgrade an existing installation that is using the "remote" profile. Isn't precheck the way to warn about the potential problem?

pkg/test/framework/components/istio/operator.go
```diff
@@ -934,7 +934,7 @@ func createIstioctlConfigFile(s *resource.Settings, workDir string, cfg Config)
 		cfg.RemoteClusterValues = cfg.ControlPlaneValues
 	}
 
-	configFiles.remoteIopFile = filepath.Join(workDir, "remote.yaml")
+	configFiles.remoteIopFile = filepath.Join(workDir, "external.yaml")
```
frankbu, line 937: This change is not correct. `remote.yaml` is not a profile name here, it is the name of the config file that the test framework is using for configuring the test cluster.

istioctl/cmd/precheck.go
howardjohn: It is the proper way to check about complex things, not simple schema validation. We already have deprecation _warnings_ in operator/pkg/apis/istio/v1alpha1/validation/validation.go; for making it a proper error we should just do it directly.

Not everyone will run `precheck`, but most people that do not will run `install` and get a super obscure error and not know what to do, probably.

frankbu: In that case, I agree that it would be best to just handle this by generating a more useful validation error message, and no need for a precheck.

pkg/test/framework/components/istio/operator.go
xichengliudui: revert

istioctl/cmd/precheck.go
xichengliudui: remove the current `checkIOPProfile` func?

xichengliudui: done

xichengliudui:
```
[root@test-build ~]# ./istioctl manifest generate --set profile=remote
Error: profile "remote" has been removed and is equivalent to "default"
```

xichengliudui:
> @xichengliudui please run `make gen` to fix the gencheck failure and fix the failing unit test in `profile-list_test.go` by removing the remote profile from the test.

done, update `profile-list_test.go` and `make gen`

GregHanson (approved):

operator/pkg/manifest/shared.go
```diff
@@ -340,13 +340,16 @@ func GetMergedIOP(userIOPStr, profile, manifestsPath, revision string, client ku
...
-// validateSetFlags validates that setFlags all have path=value format.
+// validateSetFlags 1: validates that setFlags all have path=value format, 2: check profile whether it is deprecated.
 func validateSetFlags(setFlags []string) error {
 	for _, sf := range setFlags {
 		pv := strings.Split(sf, "=")
 		if len(pv) != 2 {
 			return fmt.Errorf("set flag %s has incorrect format, must be path=value", sf)
 		}
+		if pv[0] == "profile" && pv[1] == "remote" {
+			return fmt.Errorf("profile \"remote\" has been removed and is equivalent to \"default\"")
```
frankbu, line 351:
```suggestion
			return fmt.Errorf("profile \"remote\" has been removed (use the \"default\" profile, which is equivalent)")
```

frankbu (changes requested):
Please make small changes to the wording of the error message and release note. Thanks.

operator/cmd/mesh/profile-list_test.go
```diff
@@ -38,7 +38,7 @@ func TestProfileList(t *testing.T) {
 		t.Fatalf("failed to execute istioctl profile command: %v", err)
 	}
 	output := out.String()
-	expectedProfiles := []string{"default", "demo", "empty", "minimal", "openshift", "preview", "remote"}
+	expectedProfiles := []string{"default", "demo", "empty", "minimal", "openshift", "preview"}
```
frankbu, line 41: Actually, this looks like it's also missing the external profile, which should have been there already.
```suggestion
	expectedProfiles := []string{"default", "demo", "empty", "minimal", "openshift", "preview", "external"}
```

operator/pkg/manifest/shared.go
xichengliudui: done

releasenotes/notes/remove-remote-profile.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: release-notes/v2
+kind: feature
+area: installation
+issue:
+  - 38832
+
+releaseNotes:
+  - |
+    **Removed** istio operator `remote.yaml` profile.
```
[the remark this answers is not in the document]
xichengliudui: done

operator/cmd/mesh/profile-list_test.go
xichengliudui: done

frankbu (approved):
LGTM. Thanks

esnible (approved):
LGTM

morvencao (approved):

PR #38910, opened 2022-05-13
gateway-api: update dependency and fix conformance test errors

howardjohn:
First commit fixes the sorting order to make the golden file diff cleaner. Rest of changes implement proper Hostname matching and fix AllowedKinds.

Hostname: before was just a simple string prefix match; now its a single label wildcard match (like SAN validation). Envoy doesn't support this natively in vhosts, so for now implement a `:authority` header match. Also requires changing generated VS to no longer just use `*` for hostname, and instead always use the intersection of Gateway+VS host.

AllowedKinds: before if there were no allowed kinds we allowed all.


pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -784,6 +784,8 @@ func translateHeadersOperations(headers *networking.Headers) headersOperations {
 	}
 }
 
+const singleDNSLabelWildcardRegex = "^[-a-zA-Z]*"
```
costinm, line 787: What is this used for ? DNS label includes digits too AFAIK.

pkg/config/host/name.go
```diff
@@ -58,6 +58,39 @@ func (n Name) Matches(o Name) bool {
 	return n == o
 }
 
+// MatchesSingleLabel is a variant of Matches that only will match a single label for wildcard.
+// This means "*.com", will match "a.b.com" for Matches, but not MatchesSingleLabel
```
costinm, line 62: Can you add more details - what is using this ? Gateway API ?

costinm (approved):
Looks good in general, few cosmetic comments.

pilot/pkg/networking/core/v1alpha3/route/route.go
howardjohn: Good point, let me find a more robust regex.

Basically we want to match `*.example.com` to `foo.example.com` but not `foo.bar.example.com`. Like TLS verification matching, its a 1 label wildcard.

Envoy doesn't have a native way to d othis

howardjohn:
Linter is blocked by https://github.com/istio/tools/pull/1996

PR #38911, opened 2022-05-13
WasmPlugin: Fix to update the checksum cache when the module cache hits

ingwonsong:
Currently, when the module cache of WasmPlugin hits, the checksum cache is not updated.
This may cause inefficiency when the same image was referenced by the two different tags as well as is not an intended behavior.

This PR fixes it by adding `touchEntry` to update the checksum cache when the module cache hits.

### DIFF
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -192,6 +192,7 @@ func (c *LocalFileCache) Get(
 	var modulePath string
 	modulePath, key.checksum = c.getEntry(key, pullIfNotPresent(pullPolicy, u))
 	if modulePath != "" {
+		c.touchEntry(key)
 		return modulePath, nil
 	}
 
@@ -242,6 +243,7 @@ func (c *LocalFileCache) Get(
 		key.checksum = dChecksum
 		// check again if the cache is having the checksum.
 		if modulePath, _ := c.getEntry(key, true); modulePath != "" {
+			c.touchEntry(key)
 			return modulePath, nil
 		}
 	} else if dChecksum != key.checksum {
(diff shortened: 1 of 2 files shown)


pkg/wasm/cache_test.go
```diff
@@ -535,13 +711,28 @@ func TestWasmCache(t *testing.T) {
...
 					t.Errorf("Wasm module cache entry's last access time not updated after get operation, key: %v", *cacheHitKey)
 				}
-				cache.mux.Unlock()
 			}
+
+			if diff := cmp.Diff(cache.modules, c.wantCachedModules,
+				cmpopts.IgnoreFields(cacheEntry{}, "last", "referencingURLs"),
+				cmp.AllowUnexported(cacheEntry{}),
+			); diff != "" {
+				t.Errorf("unexpected module cache is resulted:(+want, -got)\n%v", diff)
```
douglas-reid, line 725:
```suggestion
				t.Errorf("unexpected module cache: (+want, -got)\n%v", diff)
```

pkg/wasm/cache_test.go
```diff
@@ -535,13 +711,28 @@ func TestWasmCache(t *testing.T) {
...
+			}
+
+			if diff := cmp.Diff(cache.checksums, c.wantCachedChecksums,
+				cmp.AllowUnexported(checksumEntry{}),
+			); diff != "" {
+				t.Errorf("unexpected checksums is resulted:(+want, -got)\n%v", diff)
```
douglas-reid, line 731:
```suggestion
				t.Errorf("unexpected checksums: (+want, -got)\n%v", diff)
```

pkg/wasm/cache.go
```diff
@@ -375,17 +396,19 @@ func (c *LocalFileCache) purge() {
...
+				}
+				// The module has not be touched for expiry duration, delete it from the map as well as the local dir.
+				if err := os.Remove(m.modulePath); err != nil {
+					wasmLog.Errorf("failed to purge Wasm module %v: %v", m.modulePath, err)
+				} else {
+					for downloadURL := range m.referencingURLs {
+						delete(c.checksums, downloadURL)
 					}
+					delete(c.modules, k)
+					log.Infof("modules :%v", c.modules)
```
douglas-reid, line 410: is this extra log line helpful?  should it be `wasmLog.Infof`? or `wasmLog.Debugf`?

douglas-reid (commented):
looks reasonable. just one question, really.

pkg/wasm/cache.go
ingwonsong: Oh.. my bad. will remove it. Thanks!

douglas-reid (approved):

Issue #38912, opened 2022-05-13
Support additional containers in gateway Helm chart


gjacquet:
**Describe the feature request**

There are some use cases where it would be useful to run a sidecar container along with the gateway main container.

Gateways chart supports it (https://github.com/istio/istio/blob/master/manifests/charts/gateways/istio-ingress/values.yaml#L72) however the gateway chart to deploy a single gateway does not allow defining any additional containers.

**Describe alternatives you've considered**

While there are plenty of other way to get this done (forking the chart, using an admission controller etc), adding the option in this chart would be the most convenient.

**Affected product area (please put an X in all that apply)**

[ ] Docs
[x] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

howardjohn:
See https://docs.solo.io/gloo-edge/latest/installation/gateway/kubernetes/helm_advanced/; I don't think we can reasonable expose every single possible API field in the gateway.

PR #38913, opened 2022-05-13
Gateway additional containers

gjacquet:

Allow defining additional containers for a gateway. Similar to what was done in the gateways chart.
See #38912 for reference.

### DIFF
--- a/manifests/charts/gateway/templates/deployment.yaml
+++ b/manifests/charts/gateway/templates/deployment.yaml
@@ -86,6 +86,9 @@ spec:
             name: http-envoy-prom
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+        {{- with .Values.additionalContainers }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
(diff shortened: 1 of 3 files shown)


howardjohn (changes requested):
I appreciate the use case but I think this is venturing in the category of something better suited for things like https://docs.solo.io/gloo-edge/latest/installation/gateway/kubernetes/helm_advanced/.

For context, while this is simple, there are probably 10 PRs this month for similar simple changes to the helm charts with no overlap between any of them.

Helm and bespoke configuration are not a good match.

gjacquet:
I fully understand that no helm chart will ever support every single use case out there.

However in this case, I was under the impression this chart was coming to replace istio's `manifests/charts/gateways/istio-ingress` / `manifests/charts/gateways/istio-egress` charts and they both already accept `additionalContainers` as a value.

Even if this chart is not really meant to be a replacement,  I believe it would make sense to have all the gateway related charts offer similar capabilities and accepts similar values.

howardjohn:
It is intended to be a replacement, but correcting the mistakes of the initial attempts where we had basically every field exposed which caused complexity and maintenance burdens. This new chart was explicitly designed to be a simple starting point for gateways.

PR #38924, opened 2022-05-15
EnvoyFilter is also applied to HTTP Proxy Listener

### LINKED ISSUE #38673: EnvoyFilter is not applied on listener created for HTTP Proxy

Currently, we rely on HTTP Proxy configuration on Istio by setting the parameter _proxyHttpPort_ since we are creating a non-transparent mesh. This configuration adds another listener, but this listener cannot be altered by the EnvoyFilter, so it is not possible to tweak the listener configuration.
This functionality is referred as a [TODO](https://github.com/istio/istio/blob/ebf3bb6ae3a559e6fc7d7fe613cb49aa8e1080f7/pilot/pkg/networking/core/v1alpha3/listener.go#L837) in the code.

[ ] Docs
[ ] Installation
[X] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

alandiegosantos:
The function _patchOneListener,_ that applies the EnvoyFilter patches in the listener, is called in [buildHTTPProxyListener()](https://github.com/istio/istio/blob/bb93b2a69c73c0b3cbdb2f262e36a3e67bb369f5/pilot/pkg/networking/core/v1alpha3/listener_builder.go#L91). However, the _envoyFilterWrapper_ is only populated in the [patchListener()](https://github.com/istio/istio/blob/bb93b2a69c73c0b3cbdb2f262e36a3e67bb369f5/pilot/pkg/networking/core/v1alpha3/listener_builder.go#L145), so the call to _patchOneListener_ in _buildHTTPProxyListener()_ do not apply any changes to the listener created for the HTTP Proxy.

So, to fix this bug, the _patchOneListener_ for the HTTP Proxy listener moved to inside the _patchListener()_, where the other listener are patched. A new test were added to ensure that the listener are indeed removed.

Closes #38673

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/listener_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener_builder.go
@@ -94,7 +94,6 @@ func (lb *ListenerBuilder) buildHTTPProxyListener() *ListenerBuilder {
 		return lb
 	}
 	removeListenerFilterTimeout([]*listener.Listener{httpProxy})
-	lb.patchOneListener(httpProxy, networking.EnvoyFilter_SIDECAR_OUTBOUND)
 	lb.httpProxyListener = httpProxy
 	return lb
 }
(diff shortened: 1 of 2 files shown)


alandiegosantos:
@zirain Could you add the _release-notes-none_ label in the pull request? I do not think I have permissions to do it.

howardjohn (approved):

Issue #36131, opened 2021-11-18
Propose disable PILOT_ENABLE_SERVICEENTRY_SELECT_PODS


hzxuzhonghu:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**

So far have never heard users who depend on this feature. 
The motivation is: pod is the most frequently changed resource in k8s, and mostly it is also with the biggest number.  It will waste istio cpu/mem when no use.  Making it disabled by default can achieve the best.

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

dddddai:
Agree, if user don't need this feature, it may unnecessarily cost significant memory cos there could be a large amount of pods

sschepens:
We're actually using this feature, maybe it's not the correct approach but we're using it for two reasons:
- Migration of VM workloads to pods, having a service consisting of both VMs and Pods. This case is actually explicitly specified in the documentation. https://istio.io/latest/docs/reference/config/networking/service-entry/
- Services spanning multiple k8s clusters. Is there a better way of handling this case? Also Services require allocation of IP addresses which we don't much like and limits scalability.

sschepens:
@hzxuzhonghu i'm not sure I understand, i was under the impression that k8s Services only select Pods.

case 2: what i mean is that k8s Services allocate a ClusterIP which is not really necessary in a Service Mesh and just limits scalability, we don't really need or want to allocate ClusterIPs. I don't know what's the status of Headless Services in Istio and if they're recommended or not.

hzxuzhonghu:
Service can select both pod and workloadentry

In istio we use the clusterIP as well, just for DNS resolve. in theory the IP ranges is limited, but in reality I think it is big enough to support large scalability.

For headless service, I guess there is a gap now, headless service has  ORIGINAL_DST cluster, but the  DNS resolution can not cover workloadentry. I am not thinking about this thoroughly  though

hzxuzhonghu:
@howardjohn

howardjohn:
[...]

PR #38925, opened 2022-05-16
Only build workloadInstance when se selects pod enabled

hzxuzhonghu:

Fix https://github.com/istio/istio/issues/36131

Generally speaking, Pods event is most frequent. Converting pods into WorkloadInstance is expensive. 
 And also I haven't seen this feature is relied on.  So it is better to turn it off.

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -403,7 +403,7 @@ var (
 		"If enabled, pilot will authorize XDS clients, to ensure they are acting only as namespaces they have permissions for.",
 	).Get()
 
-	EnableServiceEntrySelectPods = env.RegisterBoolVar("PILOT_ENABLE_SERVICEENTRY_SELECT_PODS", false,
+	EnableServiceEntrySelectPods = env.RegisterBoolVar("PILOT_ENABLE_SERVICEENTRY_SELECT_PODS", true,
 		"If enabled, service entries with selectors will select pods from the cluster. "+
 			"It is safe to disable it if you are quite sure you don't need this feature").Get()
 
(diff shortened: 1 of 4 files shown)


ramaraochavali:
We use this feature. We can set it to true but I am little concerned about changing the default value.

ramaraochavali:
Should we probably have an annotation on pod that tells whether it should be used for SE selection instead of doing this heavy work for every pod?

hzxuzhonghu:
Curious: why you donot use k8s service select pods + workloadentry if container and vm app are simultaneously deployed?

ramaraochavali:
> why you donot use k8s service select pods + workloadentry if container and vm app are simultaneously deployed?

This was a legacy usecase where we create service entries for each of the stateful set pods for http services and we rely on Service Entry selecting pods to arrive at endpoints - IIRC, this was used to reduce the DNS lookups I guess. Do not remember exactly

hzxuzhonghu:
Not sure why k8s service for this case

ramaraochavali:
> Not sure why k8s service for this case

As I said, exists because of some legacy reason outside of our control. 

But you can change the value (we can set it to true in our deployment), as long as every one is fine with changing the default behaviour.

WDYT about driving this by annotation also on pod - so even if this flag is enabled, we do not convert every pod unnecessarily?

hzxuzhonghu:
I am ok with that, just some concern on UX

howardjohn:
I do agree this is a questionable feature, but it seems like we are breaking compatibility to just switch the flag after its been around for many releases

hzxuzhonghu:
Any other suggestions

howardjohn:
What are we trying to solve?

hzxuzhonghu:
To improve performance issue found from product

hzxuzhonghu:
And i thought it is beneficial to 99% users, so is a release note ok?

hzxuzhonghu:
Since we can not come to an agreement, i reverted the default value. Will change it in our product.

ramaraochavali (approved):

PR #38928, opened 2022-05-16
Add custom ErrorLog handler to webhook server

kfaseela:

Reduce log level for the webhook server's TLS handshake error to debug to avoid flooding the log in case multiple probes are hitting :8080/ready concurrently.

Background: Pilot-discovery invokes its own webhook readiness server at :15017/httpsReady when receiving a readiness request on :8080/ready. This is done using the webhook readiness client based on http.Client.

The default behavior of http.Client is to reuse an already established TCP connection to the host if possible or otherwise to initiate one. When multiple readiness requests coincide, such as when both readiness and liveness probes are hitting the same endpoint, the webhook readiness client can encounter situations where the initial TCP connection is busy serving the first probe. In such cases http.Client will begin to open a second connection. If the first connection finishes to serve its request and subsequently becomes ready to handle another before the second connection finishes handshaking, http.Client will abort the connection setup by sending a TCP reset and reuse connection 1 also to serve probe 2.

When http.Server/the webhook server receives TCP RST during the TLS handshake, it produces the following error:

http: TLS handshake error from 127.0.0.1:<RANDOM PORT>: EOF

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/bootstrap/webhook.go
+++ b/pilot/pkg/bootstrap/webhook.go
@@ -16,16 +16,34 @@ package bootstrap
 
 import (
 	"crypto/tls"
+	"log"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
-	"istio.io/pkg/log"
+	istiolog "istio.io/pkg/log"
 )
 
 const (
 	HTTPSHandlerReadyPath = "/httpsReady"
 )
+
+type httpServerErrorLogWriter struct{}
+
+// Webhook http.Server.ErrorLog handler specifically to filter
+// http: TLS handshake error from 127.0.0.1:<PORT>: EOF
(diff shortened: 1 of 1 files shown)


ramaraochavali (approved):

pilot/pkg/bootstrap/webhook.go
```diff
@@ -16,34 +16,53 @@ package bootstrap
 
 import (
 	"crypto/tls"
+	"log"
```
hzxuzhonghu, line 19: why use this package

hzxuzhonghu: ic

PR #38937, opened 2022-05-16
Revert "[tf] Consistent namespace creation"

howardjohn:
Reverts istio/istio#38875

This broke namespace cleanup. This is only exposed in postsubmit since cleanup errors are silently dropped (see logs:
```
2022-05-16T09:45:46.285934Z	error	tf	Error during close: 1 error occurred:
	* resource name may not be empty
```

And only postsubmit has enough tests run that we hit max pods per node


nmittler:
@howardjohn have you confirmed that this resolved issues you were seeing in post-submit?

howardjohn:
yes

Issue #36503, opened 2021-12-14
TestEndpointsDeduping flake (regerssion)


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/unit-tests_istio_postsubmit/1470570742142734336

Can reproduce locally

howardjohn:
Regression from 1.12, marking as release blocker

howardjohn:
Root cause is fe69479b63ac4e27c8a55955e2f7943309117320, found by git bisect:

```
good=6aa1f8a274
bad=55d72c192c

cat <<'EOF' > /tmp/bisect.sh
go test -c -race ./pilot/pkg/serviceregistry -o flaker.test || exit 1
p=`pwd`
cd pilot/pkg/serviceregistry
stress -max-runs 1000 -p 64 $p/flaker.test -test.v -test.run ^TestEndpointsDeduping$
EOF

chmod +x /tmp/bisect.sh
git bisect start
git bisect good $good
git bisect bad $bad
git bisect run /tmp/bisect.sh
```

hzxuzhonghu:
Sorry i missed it. It does not occur recently https://testgrid.k8s.io/istio_istio_postsubmit#unit-tests_istio_postsubmit

howardjohn:
Likely just because of low load during holidays? I can reproduce easily with https://github.com/istio/istio/wiki/Test-Flakes#reproducing-test-flakes

hzxuzhonghu:
Good, will investigate tomorrow

hzxuzhonghu:
serviceregistry_test.go:1003: retry.UntilSuccessOrFail: timeout while waiting after 144 attempts, 0/2 successes (last error: wanted [1.2.3.4:80 2.3.4.5:80] got [1.2.3.5:80 2.3.4.5:80]. All endpoints: map[outbound|80||service.namespace.svc.cluster.local:[1.2.3.5:80 2.3.4.5:80] outbound|90||service.namespace.svc.cluster.local:[]])
 
really weird, i got `1.2.3.5:80`

hzxuzhonghu:
Figured out the cause, ConfigGenTest runs the registries, but some registry, here it is kube-controller has been run before, so it runs twice.

```
func (f *ConfigGenTest) Run() {
	go f.Registry.Run(f.stop)
```

howardjohn:
I think Steven has a PR for that

howardjohn:
[...]

PR #38949, opened 2022-05-16
Fix double run of controller

howardjohn:
Fixes https://github.com/istio/istio/issues/36503#issuecomment-1127865955

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/fake.go
+++ b/pilot/pkg/serviceregistry/kube/controller/fake.go
@@ -168,6 +168,7 @@ type FakeControllerOptions struct {
 	XDSUpdater                model.XDSUpdater
 	DiscoveryNamespacesFilter filter.DiscoveryNamespacesFilter
 	Stop                      chan struct{}
+	SkipRun                   bool
 }
 
 type FakeController struct {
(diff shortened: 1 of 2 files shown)


GregHanson (approved):

pilot/pkg/serviceregistry/kube/controller/fake.go
```diff
@@ -220,8 +221,11 @@ func NewFakeControllerWithOptions(t test.Failer, opts FakeControllerOptions) (*F
 	if x, ok := xdsUpdater.(*FakeXdsUpdater); ok {
 		fx = x
 	}
-	go c.Run(c.stop)
-	kubelib.WaitForCacheSync(c.stop, c.HasSynced)
+
+	if !opts.SkipRun {
+		go c.Run(c.stop)
```
hzxuzhonghu, line 226: I feel the right fix is to split out `run` from `new`.

howardjohn: The issue is that then each test needs to remember to call it, and needs to know which stop channel to call it with, close the stop channel, etc. It's not a huge amount of work but adds up.

We could have 2 methods (one that runs and one doesn't).but figured using Options was easier since we already have it and SkipRun is used in other places

hzxuzhonghu: Yes, this is quick and i am concerned in future use, one could easily forget to set the option. Whatever I am good with it right now.

hzxuzhonghu (approved):

PR #38980, opened 2022-05-17
Fix cert rotation for proxyless, don't start SDS

costinm:
The change also moves the injection template - istio-proxy will be first ( for readability and to avoid confusing people - it is reordered by injector anyways ), and handles the case where GKE cert provider is enabled by user via annotation.

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
@@ -44,22 +44,6 @@ metadata:
   }
 spec:
   containers:
-  {{- range $index, $container := .Spec.Containers  }}
-  {{ if not (eq $container.Name "istio-proxy") }}
-  - name: {{ $container.Name }}
-    env:
-    - name: "GRPC_XDS_EXPERIMENTAL_SECURITY_SUPPORT"
-      value: "true"
-    - name: "GRPC_XDS_BOOTSTRAP"
-      value: "/etc/istio/proxy/grpc-bootstrap.json"
-    volumeMounts:
-    - mountPath: /var/lib/istio/data
-      name: istio-data
-    # UDS channel between istioagent and gRPC client for XDS/SDS
-    - mountPath: /etc/istio/proxy
-      name: istio-xds
-  {{- end }}
-  {{- end }}
   - name: istio-proxy
   {{- if contains "/" (annotation .ObjectMeta `sidecar.istio.io/proxyImage` .Values.global.proxy.image) }}
     image: "{{ annotation .ObjectMeta `sidecar.istio.io/proxyImage` .Values.global.proxy.image }}"
@@ -67,9 +51,9 @@ spec:
     image: "{{ .ProxyImage }}"
   {{- end }}
     ports:
-    - containerPort: 15090
+    - containerPort: 15020
       protocol: TCP
-      name: http-envoy-prom
+      name: metrics
     args:
     - proxy
     - sidecar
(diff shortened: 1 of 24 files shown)


pkg/istio-agent/agent.go
```diff
@@ -508,13 +509,58 @@ func (a *Agent) initSdsServer() error {
...
+	b.MaxElapsedTime = 0
+
+	for {
+		sk, err = st.GenerateSecret(security.WorkloadKeyCertResourceName)
+		if err == nil {
+			break
+		}
+		log.Warnf("failed to get certificate: %v", err)
+		time.Sleep(b.NextBackOff())
+	}
```
howardjohn, line 552: seems accidentally duplicated

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
```diff
@@ -44,32 +44,16 @@ metadata:
...
     image: "{{ annotation .ObjectMeta `sidecar.istio.io/proxyImage` .Values.global.proxy.image }}"
   {{- else }}
     image: "{{ .ProxyImage }}"
   {{- end }}
     ports:
-    - containerPort: 15090
+    - containerPort: 15020
       protocol: TCP
-      name: http-envoy-prom
+      name: metrics
```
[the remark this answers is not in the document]
costinm: Some implementations ( google prometheus integration ) default to port name metrics.

But you have a good point, if the app has a port named metrics we may have problems. Will change it to be safe.

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
```diff
@@ -197,7 +181,7 @@ spec:
     volumeMounts:
     - name: workload-socket
       mountPath: /var/run/secrets/workload-spiffe-uds
-    {{- if eq .Values.global.caName "GkeWorkloadCertificate" }}
+    {{- if or (isset .ObjectMeta.Annotations `security.cloud.google.com/use-workload-certificates`) (eq .Values.global.caName "GkeWorkloadCertificate") }}
```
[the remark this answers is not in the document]
costinm: I don't know if we want it for the regular case. 

With gRPC, the intent is to eventually be 'agent-less' and 'injection-less' too - so users can just deploy a pod with a proxyless app and it would work. In this case they would use the annotation ( or whatever equivalent volumes are used for other cert providers ). 

Let's see how it works with grpc - and we can revisit if we should also add it to Istio. Since this is a documented feature 
in GKE users may set it anyways, so probably we need to deal with it - I suspect the naming of the volume to match
what they generate is handling the case.

pkg/istio-agent/agent.go
costinm: Which part ? That's workload and root cert - we do the same in SDS service init.

howardjohn: Oh I read wrong

pkg/istio-agent/agent.go
```diff
@@ -508,13 +509,58 @@ func (a *Agent) initSdsServer() error {
...
+		//
+		// This is based on the code from newSDSService, but customized to have explicit rotation.
+		go func() {
+			st := a.secretCache
+			st.RegisterSecretHandler(func(resourceName string) {
+				// The secret handler is called when a secret should be renewed, after invalidating the cache.
+				// The handler does not call GenerateSecret - it is a side-effect of the SDS generate() method, which
+				// is called by sdsServer.OnSecretUpdate, which triggers a push and eventually calls sdsservice.Generate
+				// TODO: extract the logic to detect expiration time, and use a simpler code to rotate to files.
+				_, _ = a.getWorkloadCerts(st)
```
hzxuzhonghu, line 524: Isn't this func calls `GenerateSecret` trigger another handler running, so will cause infinite loops

hzxuzhonghu (commented):
It is lack of context, how grpc client certificate provision is different?

manifests/charts/istio-control/istio-discovery/files/grpc-agent.yaml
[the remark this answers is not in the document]
costinm: Removed it from the workload and volume - if the annotation is set, their injector will create the same thing.

For proxy - we still have an ordering issue, if their injector goes first istio-proxy will end up without the volume mount.

I'll keep playing with this on the grpc - right now we are not supporting the annotation, but will be useful to not break ( but not related to this PR )

howardjohn (approved):

PR #38994, opened 2022-05-18
Simplify ALPN protocols selection for inbound listener

jewertow:
Signed-off-by: Jacek Ewertowski <jewertow@redhat.com>

While investigating how the selection of ALPN protocols for inbound listeners is implemented, I noticed that the conditions could be simplified, because we have duplicated variables with the protocol list.

### DIFF
--- a/pilot/pkg/security/authn/utils/utils.go
+++ b/pilot/pkg/security/authn/utils/utils.go
@@ -47,15 +47,11 @@ func BuildInboundTLS(mTLSMode model.MutualTLSMode, node *model.Proxy,
 		CommonTlsContext:         &tls.CommonTlsContext{},
 		RequireClientCertificate: protovalue.BoolTrue,
 	}
-	if protocol == networking.ListenerProtocolTCP {
+	if protocol == networking.ListenerProtocolTCP && features.MetadataExchange {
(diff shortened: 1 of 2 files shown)


pilot/pkg/security/authn/utils/utils.go
```diff
@@ -47,15 +47,11 @@ func BuildInboundTLS(mTLSMode model.MutualTLSMode, node *model.Proxy,
...
 	}
-	if protocol == networking.ListenerProtocolTCP {
+	if protocol == networking.ListenerProtocolTCP && features.MetadataExchange {
 		// For TCP with mTLS, we advertise "istio-peer-exchange" from client and
 		// expect the same from server. This  is so that secure metadata exchange
 		// transfer can take place between sidecars for TCP with mTLS.
-		if features.MetadataExchange {
-			ctx.CommonTlsContext.AlpnProtocols = util.ALPNDownstreamWithMxc
-		} else {
-			ctx.CommonTlsContext.AlpnProtocols = util.ALPNDownstream
```
jewertow, line 57: In the `else` branch below, the same protocols are set.

ramaraochavali (approved):
/ok-to-test

hzxuzhonghu (approved):

PR #38998, opened 2022-05-18
remove warnings from istioctl envoy filter analyzer during first install

### LINKED ISSUE #38676: IST0151 on default install
with a fresh install:
```
Warning [IST0151] (EnvoyFilter istio-system/stats-filter-1.11) This envoy filter does not have a priority and has a relative patch operation set which can cause the envoyFilter not to be applied. Using the INSERT_FIRST option or setting the priority may help in ensuring the envoyFilter is applied correctly
```

A freshly installed Istio with no customizations shouldn't have warnings

AndreaM12345:
fixes #38676 

Sets the priority to -1 for envoy filters that are created by default by the initial install of istio

By setting the priority to -1 then won't trigger warning messages from the istioctl envoy filter analyzer.  

EnvoyFilters should be applied in  order of priority, creation time, fully qualified resource name. By setting the priority to -1 it should ensure the envoyFilters created by the initial install get applied first.  This should make sure no change in behavior because of the creation time should have also been first.

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
@@ -1782,6 +1782,7 @@ metadata:
   labels:
     istio.io/rev: default
 spec:
+  priority: -1
   configPatches:
     - applyTo: HTTP_FILTER
       match:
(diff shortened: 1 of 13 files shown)


lei-tang (approved):

stevenctl (approved):

lei-tang:
@AndreaM12345 The gencheck_istio test failed. Can you take a look? You may need to run "make gen".

Issue #37364, opened 2022-02-15
_anonymous_match_nothing_


jonassteinberg1:
### Bug Description

I have two authorization policies.  One is:

```
# wide open
spec:
  rules:
  - {}
```

and the other is actually useful.  When I `istioctl x authz check $POD` _before_ removing the (erroneous) wide open policy istioctl returns:

```
wide-open-policy
policy-that-is-supposed-to-be-here
```

When I remove the wide open policy `istioctl x authz check $POD` istioctl returns:

```
ACTION   AuthorizationPolicy                    RULES
ALLOW    _anonymous_match_nothing_              1
ALLOW    policy-that-is-supposed-to-be-here   1
```

What is the significance of `anonymous_match_nothing`?  Is it saying something equivalent to DENY ALL or is it saying "I'm just here, but I'm not mitigating anything" effectively doing nothing?

And why isn't this documented anywhere?  I see it in source code, but it is quite confusing.

Thanks!

### Version

```prose
client version: 1.11.3
control plane version: 1.7.2
data plane version: 1.7.2 (59 proxies)
```
```


### Additional Information

_No response_

howardjohn:
I think its basically a bug in `istioctl`. If there is empty rule, we end up sending envoy something like `{}`. If there is a rule we have `{name: foo, rule: blah}`. So istioctl no longer has access to the original name

howardjohn:
cc @yangminzhu

yangminzhu:
The `_anonymous_match_nothing_` is referring to the rbac config without any rules that effectively reject all requests ("allow nothing"). This is a result of having authorization HTTP policies on TCP filter chains because the TCP filter chain cannot support HTTP attributes, hence we fallback to generate an empty rule to deny all traffic.

This could be improved to generate a more explicit deny-all rule in this case so that it can have a proper name.

jonassteinberg1:
@yangminzhu the RBAC config I posted above has _an empty rule_ and this signifies ALLOW ALL according to my understanding?

yangminzhu:
> And here's the istioctl output after I remove the first which shouldn't be there:
ACTION   AuthorizationPolicy                    RULES
ALLOW    _anonymous_match_nothing_              1
ALLOW    app.app-services    1

[...]

PR #39025, opened 2022-05-19
fix: using denyAll instead of _anonymousName_

zaunist:

fix: https://github.com/istio/istio/issues/37364

### DIFF
--- a/istioctl/pkg/authz/listener.go
+++ b/istioctl/pkg/authz/listener.go
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	anonymousName = "_anonymous_match_nothing_"
+	denyAll = "denyAll"
 )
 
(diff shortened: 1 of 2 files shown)


istioctl/pkg/authz/listener.go
```diff
@@ -156,7 +156,7 @@ func Print(writer io.Writer, listeners []*listener.Listener) {
 					addPolicy(action, nameOfPolicy, indexOfRule)
 				}
 				if len(rbacHTTP.GetRules().GetPolicies()) == 0 {
-					addPolicy(action, anonymousName, "0")
+					addPolicy(rbacpb.RBAC_DENY, denyAll, "0")
```
howardjohn, line 159: Doesn't it depend on the action? This could be `RBAC_ALLOW` action in which case its really "allow all"?

zaunist:
Sorry, I need to get a little more familiar with istio, so close this pr for now.

PR #39026, opened 2022-05-19
skip workload instance processing if service entries are not there

ramaraochavali:
When `PILOT_ENABLE_SERVICEENTRY_SELECT_PODS` is enabled, we unnecessarily process all service entries to check if SE has matching selector for every pod. This PR optimizes it by doing the logic only if the namespace has SE with workload selectors

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali:
@howardjohn @hzxuzhonghu PTAL if this makes sense

hzxuzhonghu (commented):
How about serviceentry created later?

pilot/pkg/serviceregistry/serviceentry/controller.go
```diff
@@ -510,10 +520,12 @@ func (s *Controller) WorkloadInstanceHandler(wi *model.WorkloadInstance, event m
 		s.serviceInstances.deleteInstances(key, instancesDeleted)
 	}
 
-	if event == model.EventDelete {
-		s.serviceInstances.deleteInstances(key, instances)
-	} else {
-		s.serviceInstances.updateInstances(key, instances)
+	if len(instances) > 0 {
```
howardjohn, line 523: nit: IMO this is not needed, both calls will do nothing if there are no instances already.

Since there is no diff its just a style thing, but IMO its still nice to call them. Should be same cost and it would allow the functions to evolve to do something with just the key, for example

howardjohn (commented):
Just had 1 nit, didn't approve since giving @hzxuzhonghu chance to review not because the nit is blocking.

hzxuzhonghu:
I will take a look

pilot/pkg/serviceregistry/serviceentry/controller.go
```diff
@@ -439,6 +443,14 @@ func (s *Controller) serviceEntryHandler(_, curr config.Config, event model.Even
 func (s *Controller) WorkloadInstanceHandler(wi *model.WorkloadInstance, event model.Event) {
 	log.Debugf("Handle event %s for workload instance (%s/%s) in namespace %s", event,
 		wi.Kind, wi.Endpoint.Address, wi.Namespace)
+
+	s.mutex.RLock()
+	skip := wi.Kind == model.PodKind && len(s.services.seWithWorkloadSelectors[wi.Namespace]) == 0
+	s.mutex.RUnlock()
```
hzxuzhonghu, line 449: Would suggest abstract a method of Controller

pilot/pkg/serviceregistry/serviceentry/controller.go
hzxuzhonghu: I donot think we need to change here, unless you want to do something else

hzxuzhonghu (commented):
My question still exist:  a serviceentry with selector can be created later than pods, in this case the  

WorkloadInstanceHandler does nothing, and serviceentry handler only handle service.

How could the endpointShards refreshed?

ramaraochavali:
> a serviceentry with selector can be created later than pods, in this case the

You mean just after pod creation right? Even there is a race today with that - store.List() may not return the SE. Is it not the case?

hzxuzhonghu:
Not just after, can be any time after se with selector. I am reading it will miss the endpoints, maybe I missed reading something

ramaraochavali:
> Not just after, can be any time after se with selector. I am reading it will miss the endpoints, maybe I missed reading something

This is only for POD events though. If SE comes later, we update normally

hzxuzhonghu:
In this case no serviceInstances are cached, where it is refreshed?

ramaraochavali:
> In this case no serviceInstances are cached, where it is refreshed?

Is n't that the case even before this change also? If pod event comes and SE does not exist, we do not do any thing now.

https://github.com/istio/istio/blob/32afc27ea91ef68e4e5f0ee3dd47d0aedfba35f1/pilot/pkg/serviceregistry/serviceentry/controller.go#L481

I am not understanding may be what you are asking for

hzxuzhonghu:
Oh, yes.  this is a bug existing from beginnning.

hzxuzhonghu:
You can look into how to fix. Then file an issue and add a TODO here. So we can track it.

ramaraochavali:
I have added a TODO. Let me look at it later. Can u please approve/remove hold  if there are no other concerns?

hzxuzhonghu:
>https://github.com/istio/istio/pull/39026#issuecomment-1132735863
 
I misread it. The s.workloadInstances is maintained https://github.com/istio/istio/blob/32afc27ea91ef68e4e5f0ee3dd47d0aedfba35f1/pilot/pkg/serviceregistry/serviceentry/controller.go#L457-L471

And when a see with selector created later, the instances from pods can be correctly correlated,  https://github.com/istio/istio/blob/32afc27ea91ef68e4e5f0ee3dd47d0aedfba35f1/pilot/pkg/serviceregistry/serviceentry/controller.go#L914-L928

pilot/pkg/serviceregistry/serviceentry/controller.go
```diff
@@ -476,6 +476,11 @@ func (s *Controller) WorkloadInstanceHandler(wi *model.WorkloadInstance, event m
 		return
 	}
 
+	if s.shouldSkip(wi) {
```
hzxuzhonghu, line 479: This is a minor optimization to save store  List

hzxuzhonghu (approved):

ramaraochavali:
@howardjohn @hzxuzhonghu I think if this optimization has to be useful, we have to assume that SE is created before pods that use it come up - which in my opinion is a very reasonable assumption (No one will create SEs later operationally). If you think that is a problem, I think we should not move forward with this. Let me know WDYT

hzxuzhonghu:
I think this is a regression

ramaraochavali:
> I think this is a regression

OK. Then I will close this. We can not optimize this for that flow.

PR #39030, opened 2022-05-19
fix log line in aggregate controller

ramaraochavali:
Without this log does not include providerID
2022-05-19T00:59:28.748040Z	info	Registry for the cluster processing-cluster has been deleted.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn (approved):
Should we define a String() method instead?

PR #39040, opened 2022-05-19
[tf] Fix DNS for echo VM

### LINKED ISSUE #39013: [tf] Echo vm incorrectly overrides k8s /etc/hosts
### Bug Description

As described [here](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/#why-does-kubelet-manage-the-hosts-file) kubelet manages the `/etc/hosts` file.  This is causing issues (at least on GKE), where the VM pods can get into a state where they can no longer connect to istiod:

```
2022-05-18T20:34:11.815583Z	warn	sds	failed to warm certificate: failed to generate workload certificate: create certificate: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp: lookup istiod.istio-system.svc on 8.8.8.8:53: no such host"
```

It's not exactly clear when kubelet will overwrite `/etc/hosts`, but bouncing the pod seems to reproduce the problem reliably for me on GKE.

The current echo VM deployment overwrites the `/etc/hosts` [here](https://github.com/istio/istio/blob/master/pkg/test/framework/components/echo/kube/deployment.go#L383). It appears that we need an alternative way of configuring this that will work in all environments.

### Version

```prose
master
```

### Additional Information

_No response_

nmittler:
When running on GKE, tests are flaky for VMs due to the fact that kubelet can overwrite the custom hosts in `/etc/hosts`. See [k8s docs](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/#why-does-kubelet-manage-the-hosts-file).

This PR avoids the use of a custom `/etc/hosts` in favor of specifying the custom istiod host via `hostAliases`.

Tests seem to pass consistently on GKE with this change.

Fixes #39013


pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -327,20 +327,13 @@ spec:
         istio.io/test-vm: {{ $.Service }}
         istio.io/test-vm-version: {{ $subset.Version }}
     spec:
-      # Disable kube-dns, to mirror VM
```
stevenctl, line 330: Getting rid of this will let the VM resolve k8s hostnames which we don't want (without using Istio's DNS). Could lead us to miss bugs in DNS for VMs.

stevenctl, line 330: Getting rid of this will let the VM resolve k8s hostnames which we don't want (without using Istio's DNS). Could lead us to miss bugs in DNS for VMs.

nmittler:
@stevenctl @howardjohn PTAL

pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -380,12 +389,17 @@ spec:
           sudo cp /var/run/secrets/istio/bootstrap/root-cert.pem /var/run/secrets/istio/root-cert.pem
           sudo cp /var/run/secrets/istio/bootstrap/*.env /var/lib/istio/envoy/
           sudo cp /var/run/secrets/istio/bootstrap/mesh.yaml /etc/istio/config/mesh
-          sudo sh -c 'cat /var/run/secrets/istio/bootstrap/hosts >> /etc/hosts'
+
+          # don't overwrite /etc/hosts since it's managed by kubeproxy
+          #sudo sh -c 'cat /var/run/secrets/istio/bootstrap/hosts >> /etc/hosts'
```
howardjohn, line 394: nit: This technically drops our coverage since we don't know that the command generates the proper `IstioHosts`.

Maybe instead of writing it, we can just assert its the same as what we added in `hostAliases`? Like

```
for IstioHosts {
  grep /etc/hosts IstioHosts
}
```

or similar is probably enough..

pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -642,31 +668,35 @@ var RevVMImages = func() map[string]echo.VMDistro {
...
-			vmImage = cfg.VMDistro
-		} else {
-			vmImage = VMImages[echo.DefaultVMDistro]
-		}
-		log.Debugf("no image for distro %s, defaulting to %s", cfg.VMDistro, echo.DefaultVMDistro)
+	systemNS := ist.Settings().SystemNamespace
+	istioHosts = []string{
+		fmt.Sprintf("istiod.%s.svc.cluster.local", systemNS),
+		fmt.Sprintf("istiod.%s", systemNS),
+		fmt.Sprintf("istiod.%s.svc", systemNS),
```
howardjohn, line 688: nit: kind of ties in with my other comment, but I think we _only_ want istiod.%s.svc, or whatever istioctl generates.

An example bug we would miss now is if the agent was making a call to istiod.%s; this would fail on a  real VM but pass here.

nmittler:
@howardjohn PTAL

howardjohn (approved):

nmittler:
@esnible mind approving for UX?

esnible (approved):
LGTM.  Thanks for cleaning up the typos as well.

PR #39041, opened 2022-05-19
[tf] Redirect VM echo logs to a file

nmittler:


nmittler:
I wasn't actually seeing them in stdout for some reason. Although our test framework overrides this and effectively has a copy-and-paste of this logic inside the deployment yaml.

nmittler:
I don't think we need this. Closing.

Issue #38725, opened 2022-05-04
use destination port when inferring destinations from vs for sidecar


yingzhuivy:
### Bug Description

Suppose there is a client, a vs and a service.
vs has port 8080 but routes to service on port 8081.

When the client defines a sidecar with egress listener depending on vs:8080, the service with port 8081 is not included in the sidecar scope, resulting in client proxy can not get the service configs. If client depends on vs:8081, it works. But I don't think it makes sense since vs does not have port 8081.

I think if client specify vs:8080, istio should be smart enough to pull in service on port 8081. 
I believe the code to change is [here](https://github.com/istio/istio/blob/1.12.6/pilot/pkg/model/sidecar.go#L363), if it makes sense, happy to submit a fix.

### Version

```prose
istio 1.12
```

ryan-smick:
Adding on to this:

Our current setup is as Ying described above, but with the minor difference that the sidecar does not specify a port on the egress listener; it only specifies the vs host. What we're seeing is that istio will create a cluster on the client proxy for every port on the service, even if the virtual service only specifies a single port in the destinations.

It seems to me that if the virtual service will only route to one port (or a subset of ports) on the service, we can skip the cluster creation on the client for the remainder of the service ports.

hzxuzhonghu:
Kind of understanding, can you provide an example

yingzhuivy:
in our case, we have a ServiceEntry defined external service `a` with port 3306, and an egress gateway 'b'. That egress gateway has a bunch of ports, and our service `a` is using port `80`. We use the following virtual service to route A traffic to B:

```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: vs-a
  namespace: ns-a
spec:
  hosts:
  - a.ns-a....
  tcp:
  - match:
    - port: 3306
    route:
    - destination:
        host: b.ns-b
        port:
          number: 80
      weight: 100
```

For our client `c`, we have the following sidecar:
```
apiVersion: networking.istio.io/v1beta1
kind: Sidecar
metadata:
  name: sidecar-c
  namespace: ns-c
spec:
  egress:
    - hosts:
      - ns-a/a.ns-a....
  workloadSelector:
    labels:
      app: app-c
```

The sidecar depends on `ns-a/a.ns-a....` which pulls in `b.ns-b` through the virtual service. However, currently it pulls in all the ports on b, even though the virtual service is only routing to port 80.
[...]

PR #39067, opened 2022-05-20
sidecar: filter service ports to VS ports

ryan-smick:

Fixes issue: https://github.com/istio/istio/issues/38725

When constructing the SidecarScope, if a sidecar listed a dependency
on a virtual service, istio will currently add a dependency for the
sidecar on each transitive dependency of that virtual service,
including all ports that each transitive dependency exposes. This
change now filters the ports on those transitive dependencies down
to only those to which the virtual service will route. This allows users
to have services that expose hundreds of ports, but set a specific
hostname for a specific subset of those ports using a virtual service,
and istio will only propagate the configuration needed to route to
that subset of ports.

@hzxuzhonghu @howardjohn 
@yingzhuivy 

Contins behavior changes for the propagation of envoy clusters to sidecars

### DIFF
--- a/pilot/pkg/model/sidecar.go
+++ b/pilot/pkg/model/sidecar.go
@@ -27,6 +27,7 @@ import (
 	"istio.io/istio/pkg/config/labels"
 	"istio.io/istio/pkg/config/protocol"
 	"istio.io/istio/pkg/config/schema/gvk"
+	"istio.io/istio/pkg/util/sets"
 )
 
 const (
@@ -351,7 +352,7 @@ func ConvertToSidecarScope(ps *PushContext, sidecarConfig *config.Config, config
 				Namespace: vs.Namespace,
 			})
 
-			for _, h := range virtualServiceDestinationHosts(v) {
+			for h, ports := range virtualServiceDestinations(v) {
 				// Default to this hostname in our config namespace
 				if s, ok := ps.ServiceIndex.HostnameAndNamespace[host.Name(h)][configNamespace]; ok {
 					// This won't overwrite hostnames that have already been found eg because they were requested in hosts
@@ -359,7 +360,7 @@ func ConvertToSidecarScope(ps *PushContext, sidecarConfig *config.Config, config
 					if matchPort {
 						vss = serviceMatchingListenerPort(s, listener)
 					} else {
-						vss = s
+						vss = serviceMatchingVirtualServicePorts(s, ports)
 					}
 					if vss != nil {
 						addService(vss)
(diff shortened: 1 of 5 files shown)


pilot/pkg/model/sidecar.go
```diff
@@ -663,6 +664,36 @@ func serviceMatchingListenerPort(service *Service, ilw *IstioEgressListenerWrapp
 	return nil
 }
 
+func serviceMatchingVirtualServicePorts(service *Service, vsDestPorts sets.IntSet) *Service {
+	// A value of 0 in vsDestPorts is used as a sentinel to indicate a dependency
+	// on every port of the service.
+	if len(service.Ports) == 1 || len(vsDestPorts) == 0 || vsDestPorts.Contains(0) {
+		return service.DeepCopy()
```
howardjohn, line 671:
```suggestion
		return service
```

Can we either not copy or comment why we need to copy? Its not clear to me its needed/why

pilot/pkg/model/sidecar.go
```diff
@@ -663,6 +664,36 @@ func serviceMatchingListenerPort(service *Service, ilw *IstioEgressListenerWrapp
 	return nil
 }
 
+func serviceMatchingVirtualServicePorts(service *Service, vsDestPorts sets.IntSet) *Service {
+	// A value of 0 in vsDestPorts is used as a sentinel to indicate a dependency
+	// on every port of the service.
+	if len(service.Ports) == 1 || len(vsDestPorts) == 0 || vsDestPorts.Contains(0) {
```
howardjohn, line 670: I don't think it matters much in practice, but is `len(service.Ports) == 1` actually correct?

Imagine I have a Svc with port=1234 and a VS pointing to svc:5678. I think this would select the service?

pilot/pkg/model/sidecar.go
```diff
@@ -663,6 +664,36 @@ func serviceMatchingListenerPort(service *Service, ilw *IstioEgressListenerWrapp
...
+
+	foundPorts := make([]*Port, 0)
+	for _, port := range service.Ports {
+		if vsDestPorts.Contains(port.Port) {
+			foundPorts = append(foundPorts, port)
+		}
+	}
+
+	if len(foundPorts) > 0 {
+		sc := service.DeepCopy()
```
howardjohn, line 682: nit: we could probably reduce a copy when we end up with `len(foundPorts) == len(sc.Ports)`

howardjohn (commented):
Just skimmed and looks good overall, will leave to @hzxuzhonghu

pilot/pkg/model/sidecar.go
ryan-smick: Yes, this would select the service in that case. You're right, I think this is probably incorrect, and should be fine to remove. Will update

hzxuzhonghu (approved):

PR #39073, opened 2022-05-20
enable external CA tests in multi-cluster

aryan16:
**Enabled tests to use external CA in multi-cluster environment**

### DIFF
--- a/pkg/test/csrctrl/controllers/start_csrctrl.go
+++ b/pkg/test/csrctrl/controllers/start_csrctrl.go
@@ -108,7 +108,7 @@ func runManager(mgr manager.Manager, arrSigners []string, signersMap map[string]
 		Signers:        signersMap,
 		appendRootCert: appendRootCert,
 	}).SetupWithManager(mgr); err != nil {
-		log.Infof("Unable to create Controller fro controller CSRSigningReconciler, error: %v", err)
+		log.Infof("Unable to create Controller for controller CSRSigningReconciler, error: %v", err)
 		os.Exit(-1)
 	}
 	ctx, cancel := context.WithCancel(context.Background())
--- a/pkg/test/csrctrl/controllers/start_csrctrl.go
+++ b/pkg/test/csrctrl/controllers/start_csrctrl.go
@@ -23,11 +23,12 @@ import (
 	capi "k8s.io/api/certificates/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	// +kubebuilder:scaffold:imports
 	"istio.io/istio/pkg/test/csrctrl/signer"
+	"istio.io/istio/pkg/test/framework/components/cluster"
 	"istio.io/pkg/log"
 )
 
@@ -51,27 +52,17 @@ type SignerRootCert struct {
 	Rootcert string
 }
 
-func RunCSRController(signerNames string, appendRootCert bool, config *rest.Config, c <-chan struct{},
-	certChan chan *SignerRootCert,
-) {
+func RunCSRController(signerNames string, appendRootCert bool, c <-chan struct{},
+	clusters cluster.Clusters) []*SignerRootCert {
 	// Config Istio log
 	if err := log.Configure(loggingOptions); err != nil {
 		log.Infof("Unable to configure Istio log error: %v", err)
 		os.Exit(-1)
 	}
-	mgr, err := ctrl.NewManager(config, ctrl.Options{
-		Scheme: scheme,
-		// disable the metric server to avoid the port conflicting
-		MetricsBindAddress: "0",
-	})
-	if err != nil {
-		log.Infof("Unable to start manager error: %v", err)
-		os.Exit(-1)
-	}
-
-	arrSingers := strings.Split(signerNames, ",")
-	signersMap := make(map[string]*signer.Signer, len(arrSingers))
-	for _, signerName := range arrSingers {
+	arrSigners := strings.Split(signerNames, ",")
+	signersMap := make(map[string]*signer.Signer, len(arrSigners))
+	var rootCertSignerArr []*SignerRootCert
+	for _, signerName := range arrSigners {
 		signer, sErr := signer.NewSigner(signerRoot, signerName, certificateDuration)
 		if sErr != nil {
 			log.Infof("Unable to start signer for [%s], error: %v", signerName, sErr)
@@ -87,19 +78,37 @@ func RunCSRController(signerNames string, appendRootCert bool, config *rest.Conf
 			Signer:   signerName,
 			Rootcert: string(rootCert),
 		}
-		certChan <- rootCertsForSigner
+		// certChan <- rootCertsForSigner
+		rootCertSignerArr = append(rootCertSignerArr, rootCertsForSigner)
 	}
 
+	for _, cluster := range clusters {
+		mgr, err := ctrl.NewManager(cluster.RESTConfig(), ctrl.Options{
+			Scheme: scheme,
+			// disabel the metric server to avoid the port conflicting
+			MetricsBindAddress: "0",
+		})
+		if err != nil {
+			log.Infof("Unable to start manager error: %v", err)
+			os.Exit(-1)
+		}
+		go runManager(mgr, arrSigners, signersMap, appendRootCert, c)
+	}
+
+	return rootCertSignerArr
+}
+
+func runManager(mgr manager.Manager, arrSigners []string, signersMap map[string]*signer.Signer, appendRootCert bool, c <-chan struct{}) {
 	if err := (&CertificateSigningRequestSigningReconciler{
 		Client:         mgr.GetClient(),
 		SignerRoot:     signerRoot,
 		CtrlCertTTL:    certificateDuration,
 		Scheme:         mgr.GetScheme(),
-		SignerNames:    arrSingers,
+		SignerNames:    arrSigners,
 		Signers:        signersMap,
 		appendRootCert: appendRootCert,
 	}).SetupWithManager(mgr); err != nil {
-		log.Infof("Unable to create Controller for controller CSRSigningReconciler, error: %v", err)
+		log.Infof("Unable to create Controller fro controller CSRSigningReconciler, error: %v", err)
 		os.Exit(-1)
 	}
 	ctx, cancel := context.WithCancel(context.Background())
--- a/pkg/test/csrctrl/controllers/start_csrctrl.go
+++ b/pkg/test/csrctrl/controllers/start_csrctrl.go
@@ -53,7 +53,8 @@ type SignerRootCert struct {
 }
 
 func RunCSRController(signerNames string, appendRootCert bool, c <-chan struct{},
-	clusters cluster.Clusters) []SignerRootCert {
+	clusters cluster.Clusters,
+) []SignerRootCert {
 	// Config Istio log
 	if err := log.Configure(loggingOptions); err != nil {
 		log.Infof("Unable to configure Istio log error: %v", err)
--- a/pkg/test/csrctrl/controllers/start_csrctrl.go
+++ b/pkg/test/csrctrl/controllers/start_csrctrl.go
@@ -53,7 +53,7 @@ type SignerRootCert struct {
 }
 
 func RunCSRController(signerNames string, appendRootCert bool, c <-chan struct{},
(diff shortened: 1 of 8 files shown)


tests/integration/security/external_ca/main_test.go
```diff
@@ -94,15 +92,58 @@ func TestMain(m *testing.M) {
...
 
 func setupConfig(ctx resource.Context, cfg *istio.Config) {
-	certsChan := make(chan *csrctrl.SignerRootCert, 2)
-	go csrctrl.RunCSRController("clusterissuers.istio.io/signer1,clusterissuers.istio.io/signer2", false,
-		ctx.Clusters()[0].RESTConfig(), stopChan, certsChan)
-	cert1 := <-certsChan
-	cert2 := <-certsChan
-
+	certsChan := csrctrl.RunCSRController("clusterissuers.istio.io/signer1,clusterissuers.istio.io/signer2", false, stopChan, ctx.AllClusters())
+	cert1 := certsChan[0]
```
howardjohn, line 96: nit: `certsChan` naming seems confusing since its not a channel?

pkg/test/csrctrl/controllers/start_csrctrl.go
```diff
@@ -51,27 +52,17 @@ type SignerRootCert struct {
...
-		log.Infof("Unable to start manager error: %v", err)
-		os.Exit(-1)
-	}
-
-	arrSingers := strings.Split(signerNames, ",")
-	signersMap := make(map[string]*signer.Signer, len(arrSingers))
-	for _, signerName := range arrSingers {
+	arrSigners := strings.Split(signerNames, ",")
+	signersMap := make(map[string]*signer.Signer, len(arrSigners))
+	var rootCertSignerArr []*SignerRootCert
```
howardjohn, line 64: nit: 
```suggestion
	var rootCertSignerArr []SignerRootCert
```

No need to use pointer here since its not mutable

pkg/test/csrctrl/controllers/start_csrctrl.go
```diff
@@ -87,19 +78,37 @@ func RunCSRController(signerNames string, appendRootCert bool, config *rest.Conf
 			Signer:   signerName,
 			Rootcert: string(rootCert),
 		}
-		certChan <- rootCertsForSigner
+		// certChan <- rootCertsForSigner
```
howardjohn, line 81: nit: cleanup dead code

tests/integration/security/external_ca/main_test.go
```diff
@@ -81,9 +81,7 @@ func TestMain(m *testing.M) {
 	// nolint: staticcheck
 	framework.NewSuite(m).
 		Label(label.CustomSetup).
-		RequireMinVersion(19).
-		RequireSingleCluster().
-		RequireMultiPrimary().
+		RequireMinVersion(20).
```
howardjohn, line 84: just curious why this changed?

aryan16: Changed it for testing, will revert it to the original version.

aryan16:
> looks like this is ~good to go?

kinda, onto the final testing.

pkg/test/csrctrl/controllers/start_csrctrl.go
```diff
@@ -83,23 +75,40 @@ func RunCSRController(signerNames string, appendRootCert bool, config *rest.Conf
...
 		SignerRoot:     signerRoot,
 		CtrlCertTTL:    certificateDuration,
 		Scheme:         mgr.GetScheme(),
-		SignerNames:    arrSingers,
+		SignerNames:    arrSigners,
 		Signers:        signersMap,
 		appendRootCert: appendRootCert,
 	}).SetupWithManager(mgr); err != nil {
-		log.Infof("Unable to create Controller for controller CSRSigningReconciler, error: %v", err)
+		log.Infof("Unable to create Controller fro controller CSRSigningReconciler, error: %v", err)
```
frankbu, line 111:
```suggestion
		log.Infof("Unable to create Controller for controller CSRSigningReconciler, error: %v", err)
```

tests/integration/security/external_ca/main_test.go
```diff
@@ -147,6 +188,7 @@ components:
                 verbs:
                 - approve
 `, map[string]string{"rootcert1": cert1.Rootcert, "signer1": cert1.Signer, "rootcert2": cert2.Rootcert, "signer2": cert2.Signer})
+
 	cfg.ControlPlaneValues = cfgYaml
```
frankbu, line 192: When running an external control plane, the base config file for the primary cluster is different than when running normally (with default profile), so IIUC this is using the same overlay over two different bases. I suppose if the tests pass, it's OK, but it seems this would be testing with a (somewhat?) different istiod config from what users would have in an external control plane cluster?

aryan16: It is still using a separate config file for external primary cluster (https://github.com/istio/istio/blob/master/tests/integration/iop-externalistiod-primary-integration-test-defaults.yaml), we are just adding the external cert info on top of that. 
https://github.com/istio/istio/blob/master/pkg/test/framework/components/istio/operator.go#L296
https://github.com/istio/istio/blob/master/pkg/test/framework/components/istio/operator.go#L475

frankbu (approved):

pkg/test/csrctrl/controllers/start_csrctrl.go
```diff
@@ -83,19 +75,36 @@ func RunCSRController(signerNames string, appendRootCert bool, config *rest.Conf
...
+			Scheme: scheme,
+			// disabel the metric server to avoid the port conflicting
+			MetricsBindAddress: "0",
+		})
+		if err != nil {
+			log.Infof("Unable to start manager error: %v", err)
+			os.Exit(-1)
+		}
+		go runManager(mgr, arrSigners, signersMap, appendRootCert, c)
+	}
```
stevenctl, line 96: cc @zhlsunshine – not sure the reason for the original design with channel, PTAL

pkg/test/framework/components/istio/operator.go
```diff
@@ -400,11 +400,22 @@ func deploy(ctx resource.Context, env *kube.Environment, cfg Config) (Instance,
 			// before the external control plane cluster. Since remote clusters use gateway injection, we can't install the gateways
 			// until after the control plane is running, so we install them here. This is not really necessary for pure (non-config)
 			// remote clusters, but it's cleaner to just install gateways as a separate step for all remote clusters.
-			if err = installRemoteClusterGateways(s, i, c); err != nil {
-				return i, err
+
+			// Config cluster may have a separate config from pure remote clusters.
+			configFiles := []string{
+				filepath.Join(testenv.IstioSrc, IntegrationTestRemoteGatewaysIOP),
+			}
```
stevenctl, line 407: nit: move the config file selection int `installRemoteClusterGateways` to avoid future usages passing the wrong file

pkg/test/framework/components/istio/operator.go
```diff
@@ -400,11 +400,22 @@ func deploy(ctx resource.Context, env *kube.Environment, cfg Config) (Instance,
...
+			if c.IsConfig() {
+				configFiles = append(configFiles, istioctlConfigFiles.remoteIopFile)
```
frankbu, line 409: This seems strange. Why is it using the `remoteIopFile` (not `configIopFile`) for config clusters, and doesn't need anything for pure remote clusters?

aryan16: can't use `configIopFile` because we are explicitly disabling the gateways there and it is a requirement to setup config cluster. We don't need anything for pure remote clusters cuz we already have a dedicated yaml for that (https://github.com/istio/istio/blob/master/tests/integration/iop-remote-integration-test-gateways.yaml) 
@stevenctl  This piece of code is only running for remote clusters so I don't think there's a need for `IsRemote`.

aryan16: `remoteIopFile` is needed because it has the info about `externalCA` with enabled gateways. IIUC Here we are not setting pure config cluster, we are setting remote config cluster (i.e. not the config part of the config cluster)

stevenctl: If I can attempt to summarize why we're changing things here:

1. We install on all remotes (config + pure remote)
2. We later install gateway on remote clusters (this must be done in a separate step)
3. That gateway install part gets rid of the external CA settings

This PR:
4. Overlays the external CA stuff when installing gateways. 

Instead we might:
1. Expose some way to set remote cluster gateway overlays the same way we expos ControlPlaneValues, RemoteClusterValues etc.

aryan16: I agree!! we should have a more generic way to do this.

frankbu: Yeah, this is kind of subtle ... I missed the part about `remoteIopFile` using `controlPlaneValues`. I agree that a good solution would be to enable setting gateway overlays in the test config.

Otherwise, I still wonder why you need to use `remoteIopFile` instead of `configIopFile` given that you are also overlaying the external CA stuff there too: `cfg.ConfigClusterValues = cfgConfigYaml`. Seems like using `remoteIopFile` when installing the gateways could be removing something that should be on the config cluster??

aryan16: We need to remove the gateway installation initially when it tries to setup the config cluster, we are using this config in `configIopFile`. And when it comes to gateway installation in later stages, it still has `enabled: false` for that.
https://github.com/istio/istio/pull/39073/files#diff-dad3a0fd21c90716b69a568d6e74151665c33ec7492869d8c2b46edc57eb2e45R129

frankbu: > in later stages, it still has enabled: false 

Yes, but the gateway yaml changes it to true: https://github.com/istio/istio/blob/dde2015562d8bf0216237322a15dd4625b3530c2/tests/integration/iop-remote-integration-test-gateways.yaml#L11

Is the problem that the override goes in the wrong direction ... `configIopFile` overrides the gateway one, instead the other way around?

I also just noticed that your override for remote clusters is enabling pilot: https://github.com/istio/istio/blob/98336b09e94298a4a9c3f8fa9126023c1b865023/tests/integration/security/external_ca/main_test.go#L136

So that will install istiod now on non-config remote clusters. Normally the remote clusters would also be istiodless.

aryan16: yeah, I guess you're right. Let me try to use  `iop-remote-integration-test-gateways.yaml` on top of `configIopFile`

tests/integration/security/external_ca/main_test.go
```diff
@@ -146,7 +159,9 @@ components:
                 - signers
                 verbs:
                 - approve
-`, map[string]string{"rootcert1": cert1.Rootcert, "signer1": cert1.Signer, "rootcert2": cert2.Rootcert, "signer2": cert2.Signer})
-	cfg.ControlPlaneValues = cfgYaml
-	cfg.DeployEastWestGW = false
+`, map[string]interface{}{
+		"rootcert1": cert1.Rootcert, "signer1": cert1.Signer, "rootcert2": cert2.Rootcert,
```
stevenctl, line 163: super nit: either do one k/v per line or group them logically (e.g. rootcerts on one line signercert1 on a line, 2 on another etc)

pkg/test/framework/suite.go
```diff
@@ -264,6 +266,20 @@ func (s *suiteImpl) RequireMultiPrimary() Suite {
 	return s
 }
 
+func (s *suiteImpl) IgnoreConfigClusterTopology() Suite {
```
stevenctl, line 269:
```suggestion
func (s *suiteImpl) RequireInClusterConfig() Suite {
```

maybe? I was gonna say it should begin with `Require` but the wording is tricky. I think `Skip` is better than `Ignore` though.

pkg/test/framework/suite.go
```diff
@@ -264,6 +266,20 @@ func (s *suiteImpl) RequireMultiPrimary() Suite {
 	return s
 }
 
+func (s *suiteImpl) SkipConfigClusterTopology() Suite {
```
frankbu, line 269: One small naming suggestion, but otherwise LGTM.

I think the following would be easier to understand:
```
func (s *suiteImpl) SkipExternalControlPlaneTopology() Suite {
	fn := func(ctx resource.Context) error {
		for _, c := range ctx.AllClusters() {
			if c.IsExternalControlPlane() {
				s.Skip(fmt.Sprintf("Cluster %s is an external control plane, we can't run external control plane topology",
					c.Name()))
			}
		}
		return nil
	}
	s.requireFns = append(s.requireFns, fn)
	return s
}
```

aryan16: I agree, updated the name.

howardjohn (approved):

Issue #38525, opened 2022-04-22
cluster dropdown in Grafana Dashboard for multicluster


venkatk-25:
**Describe the feature request**
We have istio installed across multiple clusters in a multi-priomary setup and have configured prometheus according to [monitoring-multicluster-prometheus](https://istio.io/latest/docs/ops/configuration/telemetry/monitoring-multicluster-prometheus/). to aggregate in a central prometheus.

The default istio dashboards do not have the dropdown at cluster level and they seem to be built for a single cluster, which makes it very difficult to drill down to a single cluster for visualizing/debugging.

Istio already has cluster label attached to each metrics. It would be nice if we can add source and destination cluster labels in all relevant dashboards since we already have those labels for metrics, so it is easier to monitor/drill down to specific cluster.

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[X] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[X] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

venkatk-25:
@zirain Are you planning to implement this ?

zirain:
feel free to take over, I may be very busy in May.

howardjohn:
https://github.com/istio/istio/pull/39081

PR #39081, opened 2022-05-23
service dashboard add cluster dropdown

zirain:
part of https://github.com/istio/istio/issues/38525

open to hear more voice

image

image

image


douglas-reid:
Can you drop a screenshot (or, even better a working sample dash link) that shows the impact of this change?

zirain:
> Can you drop a screenshot (or, even better a working sample dash link) that shows the impact of this change?

update some screenshots

douglas-reid:
Thanks!  Looks good. WDYT about `Workload Cluster` vs. `Service Cluster` ?  My concern is that services can be multi-cluster, and that the `source_cluster` and `destination_cluster` labels are supposed to be pulled from the workload instances.

zirain:
> Thanks! Looks good. WDYT about `Workload Cluster` vs. `Service Cluster` ? My concern is that services can be multi-cluster, and that the `source_cluster` and `destination_cluster` labels are supposed to be pulled from the workload instances.

My thought is very simple, `Cluster` is one of filter to get workload, just like `Namespace` do. for me, it's enough, glad to here others' voice, can you share more about you concern?

douglas-reid:
> My thought is very simple, `Cluster` is one of filter to get workload, just like `Namespace` do. for me, it's enough, glad to here others' voice, can you share more about you concern?

My concern is that `source_cluster` and `destination_cluster` are taken from workload instance attributes, and not from service attributes. If you had a service that was provided by workloads in two different clusters, it _feels_ like it could be confusing to just call the dropdown `Service Cluster`. A display name like `Service Workload Cluster` seems a bit more precise, but is certainly more unwieldy.

zirain:
Metrics are scraped from workload/pods by prometheus. 
In my understanding, `service` (in this dashboard) is endpoint which handle the request, not a k8s service.

douglas-reid:
> Metrics are scraped from workload/pods by prometheus. In my understanding, `service` (in this dashboard) is endpoint which handle the request, not a k8s service.

The `service` is not an endpoint; it is a collection of endpoints (the implementing workloads). That's why there is a dropdown for `Service Workload`.

The other dropdown is `Service Workload Namespace`.  We should either name them both `Service Workload (Namespace|Cluster)` or both `Service (Namespace|Cluster)` to keep it consistent.

zirain:
>  name them both Service Workload (Namespace|Cluster)

this's a good idea

zirain:
ping @douglas-reid

douglas-reid (approved):

zirain:
ping @istio/wg-environments-maintainers

irisdingbj:
Can you drop an updated screenshot here? Thanks!

irisdingbj (commented):
Can you drop an updated screenshot here? Thanks!

zirain:
@irisdingbj updated with new screenshots

irisdingbj (approved):
LGTM, Thanks!

PR #39089, opened 2022-05-23
Manual common-files update with lint fix.

ericvn:

Manual common-files update with lint fix.

Replaces: https://github.com/istio/istio/pull/39090


ericvn:
#39090 merged so this isn't needed.

Issue #39079, opened 2022-05-21
Allow to configure LoadBalancerClass for service in gateway helm chart


Legion2:
**Describe the feature request**
I want to set the `spec.loadBalancerClass` of the Service in the gateway helm chart, see [Specifying class of load balancer implementation](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class).

**Describe alternatives you've considered**
Use annotations on the Service to select load balancer implementation.

**Affected product area (please put an X in all that apply)**

[ ] Docs
[x] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**
The AWS Load Balancer Controller supports this field to select the load balancer controller instead of the in-tree controller. The `spec.loadBalancerClass` field was added in kubernetes v1.22.

howardjohn:
We cannot add an option for every single field in Kubernetes to the charts. If you want to configure advanced fields, i recommend following steps like https://docs.solo.io/gloo-edge/latest/installation/gateway/kubernetes/helm_advanced/

PR #39102, opened 2022-05-24
[gateways/istio-ingressgateway] expose externalIPs for addition to th…

### LINKED ISSUE #38735: make the gateway chart behave like the gateways/istio-ingress chart
### Bug Description

im finding beetween the gateway chart and the gateways/istio-ingress chart the resulting deployed objects differ considerably. 

using the gateway chart the deployments dont start, complaining of missing sysctl values. even swapping out the securityContext and containerSecurityContext values in the gateway chart with those from a working deployment under the gateways/istio-ingress chart doesn't result in a deployment that can bind to ports 443 and 80 on the scheduled node.

really what im after is the ability to set `externalIPs` for the ingressgateway service. the Gateway chart exposes this but the gateways/istio-ingress chart does not. unless you can point me in the direction of a way to have this set in the gateways/istio-ingress chart?

### Version

```prose
istio-release 1.13.2
kubernetes 1.23
```

### Additional Information

_No response_

daveoy:
…e gateway's service in this chart

this PR fixes #38735 by exposing externalIPs in the service template of the gateways/istio-ingressgateway chart

### DIFF
--- a/manifests/charts/gateways/istio-ingress/templates/service.yaml
+++ b/manifests/charts/gateways/istio-ingress/templates/service.yaml
@@ -27,6 +27,11 @@ spec:
   externalTrafficPolicy: {{$gateway.externalTrafficPolicy }}
 {{- end }}
   type: {{ $gateway.type }}
+{{- if $gateway.externalIPs }}
+  externalIPs: {{- range $gateway.externalIPs }}
+    - {{.}}
+  {{- end }}
+{{- end }}
   selector:
 {{ $gateway.labels | toYaml | indent 4 }}
   ports:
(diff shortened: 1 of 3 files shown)


daveoy:
in case a human actually reads this, this patch is similar to [the gateway chart](https://github.com/istio/istio/blob/63f7699d16945491faded98b7c7d42fb70fb010c/manifests/charts/gateway/templates/service.yaml#L43-L47) where a working template actually uses the behaviour i desired from the gateways/istio-ingressgateway chart (referenced in my issue #38735.

howardjohn (commented):
My 2c is this is like https://github.com/istio/istio/issues/39079#issuecomment-1135133605.

I am going to bring this up to the working groups so we can get an official policy on helm chart changes moving forward

daveoy:
So in fact I am accomplishing this in production right now via the method mentioned in that comment. I just thought it might be easier for others to use basic helm functionality to get this job done without having to dive any deeper. 

Thanks for checking.

daveoy:
unittest and gencheck, both related checks will never pass as this field is not part of the v1alpha1.ingressgateway spec, but rather a mod to the service object. in order for this to be merged are we looking at also a mod to the object API?

PR #39103, opened 2022-05-24
Refactor: renaming localCluster to configCluster

hzxuzhonghu:

The `localCluster` is confusing in external istiod model.  #38654 is trying to add local secret watch, which makes the` localCluster `especially confusing.

Actually, local cluster means primary cluster. And primary is also widely used in official docs.

### DIFF
--- a/pilot/pkg/credentials/kube/multicluster.go
+++ b/pilot/pkg/credentials/kube/multicluster.go
@@ -30,16 +30,16 @@ type secretHandler func(name string, namespace string)
 type Multicluster struct {
 	remoteKubeControllers map[cluster.ID]*CredentialsController
 	m                     sync.Mutex // protects remoteKubeControllers
-	localCluster          cluster.ID
+	primaryCluster        cluster.ID
 	secretHandlers        []secretHandler
 }
 
 var _ credentials.MulticlusterController = &Multicluster{}
 
-func NewMulticluster(localCluster cluster.ID) *Multicluster {
+func NewMulticluster(primaryCluster cluster.ID) *Multicluster {
 	m := &Multicluster{
 		remoteKubeControllers: map[cluster.ID]*CredentialsController{},
-		localCluster:          localCluster,
+		primaryCluster:        primaryCluster,
 	}
 
 	return m
@@ -81,13 +81,13 @@ func (m *Multicluster) ForCluster(clusterID cluster.ID) (credentials.Controller,
 	agg := &AggregateController{}
 	agg.controllers = []*CredentialsController{}
 	agg.authController = m.remoteKubeControllers[clusterID]
-	if clusterID != m.localCluster {
-		// If the request cluster is not the local cluster, we will append it and use it for auth
-		// This means we will prioritize the proxy cluster, then the local cluster for credential lookup
+	if clusterID != m.primaryCluster {
+		// If the request cluster is not the primary cluster, we will append it and use it for auth
+		// This means we will prioritize the proxy cluster, then the primary cluster for credential lookup
 		// Authorization will always use the proxy cluster.
 		agg.controllers = append(agg.controllers, m.remoteKubeControllers[clusterID])
 	}
-	agg.controllers = append(agg.controllers, m.remoteKubeControllers[m.localCluster])
+	agg.controllers = append(agg.controllers, m.remoteKubeControllers[m.primaryCluster])
 	return agg, nil
 }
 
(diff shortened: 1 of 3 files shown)


frankbu:
I'm not sure this isn't more confusing. In Istio there are 3 kinds of clusters:

1. remote cluster - a cluster that has no istiod running on it
2. config cluster - a remote cluster that contains the mesh config resources
3. primary cluster - a cluster running istiod. This is normally also used for mesh config, but in external istiod it is just running an istiod that is watching for config on a separate config cluster.

So, in the istiod code, it seems that we are calling the source of config the "local cluster", which is confusing. But calling it instead, the "primary cluster" is wrong in the external istiod case, since it's the config cluster in that case. 

Maybe instead of changing the field from `localCluster` -> `primaryCluster` it would be more clear to rename `localCluster` -> `configSource`. That would seem to be a clear name for what it is, IIUC.

nmittler:
IIUC `localCluster` effectively means `this cluster`, or the cluster observed by `this` controller. We don't want to introduce new terminology if we can avoid it, and I agree with Frank that Primary cluster isn't the right term here. I think `localCluster` is probably more clear than any alternative listed so far (i.e. no change).

litong01:
@frankbu @hzxuzhonghu it is all extremely confusing, I think that the root cause is that these terms were not defined using same aspects. They are in some cases overlapping. For example, when we say `external`, we did not mean it is external to `istiod`, we meant `external` to workload or to configurations or `config` cluster.  When we start defining `config` we use other aspects of istio. The words `external`, `remote` for normal people would mean something based on location/position/distance of an object. Word `config` has nothing to do with location, it is very much something about an object's function or capability. We really messed up these things together, that is why it is so confusing. Should we simply use the following ones:

1. IstiodCluster
2. configCluster
3. workloadCluster

they are all can be just one cluster. but all based on its function, has nothing to do with locations. my 2 cents.

stevenctl:
1. ControlPlaneCluster: Runs istiod
2. Config Cluster: istiod reads all of it's basic config from this (the cluster it's KUBECONFIG points to)
3. Primary Cluster: Essentially `Control Plane Cluster` and `Config Cluster` combined. 
4. Remote cluster: All this means is there isn't an istiod in-cluster. 

I think 2 "Config Cluster" is the most appropriate. Also maps well with the fact that this is the cluster the kube**config** points to.

nmittler:
@litong01 I'm not against changing terms, but a lot of time was spent on this and there was a lot of back and forth. If we want to change terminology now, this PR is not the place to do it. We need to back up and get general community agreement on the terminology changes first.

stevenctl:
to be clear, I'm not saying we change any terminology; just pointing out what we already agreed on, and agreeing that this name in the code is confusing. 



https://github.com/istio/istio/blob/bdfabd351326f5a36931623d9fdebf152de9fae6/pkg/test/framework/components/cluster/topology.go#L128-L127

https://istio.io/latest/docs/ops/deployment/deployment-models/

frankbu:
+1 that we don't introduce new terminology if we can avoid it, as @nmittler suggested.

> IIUC localCluster effectively means this cluster, or the cluster observed by this controller.

This isn't totally clear because "this controller" is also watching (for endpoints) other remote clusters too. It's only the config cluster that makes this one special.

So, I think I agree with @stevenctl, we should change the variable from `localCluster` -> `configCluster`. Then, when we talk about istiod's "local cluster", we mean the cluster istiod is running in, as one would expect.

hzxuzhonghu:
OK, i will change localCluster to configCluster

hzxuzhonghu:
> Config Cluster: istiod reads all of it's basic config from this (the cluster it's KUBECONFIG points to) 

Also, it is the first (must) serviceregistry

frankbu (approved):
LGTM

hzxuzhonghu:
@stevenctl ptal

stevenctl (approved):

stevenctl (commented):
We also want this rename in https://github.com/istio/istio/blob/master/pilot/pkg/serviceregistry/kube/controller/multicluster.go

hzxuzhonghu:
Sorry for missing this part.

hzxuzhonghu:
@ruigulala 

Why do you cherry pick it? This is not a bug fix

ruigulala:
@hzxuzhonghu Yeah, you're right. It was created by mistake.

Issue #37354, opened 2022-02-15
Umbrella Issue: [unit test] Improve code coverage for complete code base


shivanshuraj1333:
### This umbrella issue can be linked to pull requests improving existing unit test coverage.

 _What to do?_

- If you're working on unit tests and want to send a PR, you can link this issue to your PR.
- If you see some scope of improving/adding unit test, but you're not working on it. You can open an issue describing your observations and link this issue, OR you can add a comment in this umbrella issue; it will help interested people to work on it.

_Suggestions to begginers_
- If you're a beginner and new to the istio codebase and looking to improve your understanding, one good way to start is by setting up a Kubernetes cluster and debugging the istio codebase while navigating through the codebase [see preparing for development guide](https://github.com/istio/istio/wiki/Preparing-for-Development); if you see missing or incomplete unit tests. Feel free to open a pull request.

shivanshu1333:
lifecycle/staleproof

MarkArranz:
I'm looking for a good place to get started contributing to Istio. Are there more tests needed?

PR #39116, opened 2022-05-24
Filter coverage

ramanujadasu:
Added filter_test.go test cases
(Ref: https://github.com/istio/istio/issues/37354)


ramaraochavali:
IMO these kind of tests do not add much value other than increasing the build times. Coverage is not the only criteria for us -  we want to have more meaningful tests.

howardjohn (approved):

hzxuzhonghu:
I agree with @ramaraochavali   Tests for a simple construction function can be included in other complex test

Issue #28402, opened 2020-10-29
ads bug: CDS update without EDS results in 503 UH


howardjohn:
If we push CDS, and update an existing cluster without an EDS update, we will get 503 UH 15s later

Pilot logs:
```
2020-10-29T14:54:52.034503Z     info    ads     Push debounce stable[6] 1: 100.181026ms since last change, 100.180817ms since last push, full=true
2020-10-29T14:54:52.035037Z     debug   ads     InitContext 2020-10-29T14:54:52Z/3 for push took 362.623µs
2020-10-29T14:54:52.035103Z     info    ads     XDS: Pushing:2020-10-29T14:54:52Z/3 Services:7 ConnectedEndpoints:1
2020-10-29T14:54:52.035600Z     info    ads     CDS: PUSH for node:istio-ingressgateway-6bdf7c75d4-6jm4q.istio-system resources:2 size:800B
2020-10-29T14:54:52.035989Z     info    ads     LDS: PUSH for node:istio-ingressgateway-6bdf7c75d4-6jm4q.istio-system resources:1 size:2.7kB
2020-10-29T14:54:52.036183Z     info    ads     RDS: PUSH for node:istio-ingressgateway-6bdf7c75d4-6jm4q.istio-system resources:1 size:368B
2020-10-29T14:54:52.044794Z     debug   ads     ADS:EDS: ACK router~10.36.0.26~istio-ingressgateway-6bdf7c75d4-6jm4q.istio-system~istio-system.svc.cluster.local-1 2020-10-29T14:53:37Z/2 tOy1HgpdiXI=44bd8ce3-e837-48e5-9345-8109f39457ec
2020-10-29T14:54:52.045413Z     debug   ads     ADS:LDS: ACK router~10.36.0.26~istio-ingressgateway-6bdf7c75d4-6jm4q.istio-system~istio-system.svc.cluster.local-1 2020-10-29T14:54:52Z/3 kSd3h94ohvU=6690f8a8-f5c9-46aa-8c4a-3b575f0502a5
2020-10-29T14:54:52.045672Z     debug   ads     ADS:RDS: ACK router~10.36.0.26~istio-ingressgateway-6bdf7c75d4-6jm4q.istio-system~istio-system.svc.cluster.local-1 2020-10-29T14:54:52Z/3 kSd3h94ohvU=de6967bf-57ee-4e26-b38d-523b4d19c176
2020-10-29T14:54:52.531025Z     info    ads     Push Status: {}
[...]

Issue #38709, opened 2022-05-03
Proxy startup stalled by warming EDS clusters


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-124_istio_postsubmit/1521489723254640640

What we see is version 6 is pushed, but version 9 is stuck warming. For ingress gw, it startsup since it completed with v6, but for egress v9 must have blocked the init.

```
2022-05-03T14:27:08.433846Z     info    ads     ADS: new connection for node:istio-egressgateway-56cf788c86-x2x58.istio-system-2
2022-05-03T14:27:08.434802Z     info    ads     CDS: PUSH request for node:istio-egressgateway-56cf788c86-x2x58.istio-system resources:18 size:18.1kB cached:0/17
2022-05-03T14:27:08.441481Z     info    ads     Push debounce stable[8] 10 for config EnvoyFilter/istio-system/stats-filter-1.15 and 9 more configs: 100.201218ms since last change, 114.869484ms since last push, full=true
2022-05-03T14:27:08.446618Z     info    ads     XDS: Pushing:2022-05-03T14:27:08Z/7 Services:6 ConnectedEndpoints:2 Version:2022-05-03T14:27:08Z/7
2022-05-03T14:27:08.447737Z     info    ads     CDS: PUSH for node:istio-egressgateway-56cf788c86-x2x58.istio-system resources:18 size:18.1kB cached:0/17
2022-05-03T14:27:08.448289Z     info    ads     EDS: PUSH request for node:istio-egressgateway-56cf788c86-x2x58.istio-system resources:17 size:2.0kB empty:1 cached:15/17
2022-05-03T14:27:09.036214Z     info    ads     Push debounce stable[9] 10 for config EnvoyFilter/istio-system/tcp-stats-filter-1.12 and 9 more configs: 100.339244ms since last change, 109.692538ms since last push, full=true
2022-05-03T14:27:09.041912Z     info    ads     XDS: Pushing:2022-05-03T14:27:09Z/8 Services:6 ConnectedEndpoints:2 Version:2022-05-03T14:27:09Z/8
2022-05-03T14:27:09.042957Z     info    ads     CDS: PUSH for node:istio-egressgateway-56cf788c86-x2x58.istio-system resources:18 size:18.1kB cached:10/17
[...]

PR #39120, opened 2022-05-24
xds: ensure CDS and EDS filters are the same

howardjohn:
Fixes https://github.com/istio/istio/issues/28402

Related to https://github.com/istio/istio/issues/38709 but probably not
fixing it

### DIFF
--- a/pilot/pkg/xds/cds.go
+++ b/pilot/pkg/xds/cds.go
@@ -16,8 +16,6 @@ package xds
 
 import (
 	"istio.io/istio/pilot/pkg/model"
-	"istio.io/istio/pkg/config"
-	"istio.io/istio/pkg/config/schema/gvk"
 )
 
 type CdsGenerator struct {
@@ -25,25 +23,6 @@ type CdsGenerator struct {
 }
 
 var _ model.XdsDeltaResourceGenerator = &CdsGenerator{}
-
-// Map of all configs that do not impact CDS
-var skippedCdsConfigs = map[config.GroupVersionKind]struct{}{
-	gvk.Gateway:               {},
-	gvk.WorkloadEntry:         {},
-	gvk.WorkloadGroup:         {},
-	gvk.AuthorizationPolicy:   {},
-	gvk.RequestAuthentication: {},
-	gvk.Secret:                {},
-	gvk.Telemetry:             {},
-	gvk.WasmPlugin:            {},
-	gvk.ProxyConfig:           {},
-}
-
-// Map all configs that impacts CDS for gateways.
-var pushCdsGatewayConfig = map[config.GroupVersionKind]struct{}{
-	gvk.VirtualService: {},
-	gvk.Gateway:        {},
-}
 
 func cdsNeedsPush(req *model.PushRequest, proxy *model.Proxy) bool {
 	if req == nil {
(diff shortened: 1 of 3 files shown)


hzxuzhonghu:
I am not sure if this is the right approach, I am still not sure why no eds request comes after the CDS updated.

howardjohn:
That is just how Envoy works for some reason :shrug: .

Even if they did send a request, it would be indistinguishable from an ACK with our current code.

hzxuzhonghu:
That's bad, envoy / xds should mark ACK explicitly rather than depending on the version_info

hzxuzhonghu (approved):

hzxuzhonghu:
Should backport, it is a bug fix

hzxuzhonghu:
@howardjohn I tested with debug logs open, seems this handle is enopugh.
Here the nonce not equal when a request come. 
https://github.com/istio/istio/blob/dde2015562d8bf0216237322a15dd4625b3530c2/pilot/pkg/xds/ads.go#L383-L390

hzxuzhonghu:
If you think there is still other uncovered cases, feel free to remove the hold

howardjohn:
Sorry not sure I 100% follow https://github.com/istio/istio/pull/39120#issuecomment-1138079606. What is the issue/what should we do instead?

This is based on the XDS doc "The management server should not send a [DiscoveryResponse](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/discovery/v3/discovery.proto#envoy-v3-api-msg-service-discovery-v3-discoveryresponse) for any [DiscoveryRequest](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/discovery/v3/discovery.proto#envoy-v3-api-msg-service-discovery-v3-discoveryrequest) that has a stale nonce. A nonce becomes stale following a newer nonce being presented to Envoy in a [DiscoveryResponse](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/discovery/v3/discovery.proto#envoy-v3-api-msg-service-discovery-v3-discoveryresponse)." (https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol)

hzxuzhonghu:
I misread the code, and thought it return true.


But our judgment seems not right,  a request can have different nonce with previous sent nonce.  should take a look at how envoy

howardjohn:
Right in that case we are supposed to ignore it? I think I am not following
you since I am not sure the connection for this PR and nonce?

hzxuzhonghu:
If we can distinguish a request from ACK, then we no need the update in this pr.

hzxuzhonghu:
@howardjohn After investigating a bit more, i could not find how vs update will lead to a existing cluster update, it can lead to add/remove a cluster.  
In both case, eds push is redundant, including stale cluster endpoints for remove.   


For both, eds requests will come later, and istiod will send the up-to-date eds at once.

howardjohn:
To distinguish things, I think we would need to become stateful. See https://github.com/istio/istio/issues/38709#issuecomment-1136137121, we get 2 requests identical requests (one ACK one request):

```
ClusterLoadAssignment, res=[my-cluster], ver=v2, nonce=nonce-eds-1
ClusterLoadAssignment, res=[my-cluster], ver=v2, nonce=nonce-eds-1
```

However from the Envoy issue (https://github.com/envoyproxy/envoy/issues/13009) it seemed like there was no request at all.. so I am a bit confused.

Either way, good point about VS. Perhaps we don't need this at all. This was mostly something I noticed when debugging https://github.com/istio/istio/issues/38709, but it does *not* fix that issue, and I don't have any concrete examples this fixes

PR #39124, opened 2022-05-25
move all pilot-agent proxy args to pk option

vince-0202:
finish TODO:Move most of this to pkg options.

- [ ] Docs
- [ ] Installation
- [X ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/cmd/pilot-agent/app/cmd.go
+++ b/pilot/cmd/pilot-agent/app/cmd.go
@@ -26,6 +26,7 @@ import (
 	"istio.io/istio/pilot/cmd/pilot-agent/config"
 	"istio.io/istio/pilot/cmd/pilot-agent/options"
 	"istio.io/istio/pilot/cmd/pilot-agent/status"
+
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pilot/pkg/util/network"
 	"istio.io/istio/pkg/cmd"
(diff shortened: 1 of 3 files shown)


pilot/cmd/pilot-agent/app/cmd.go
```diff
@@ -26,6 +26,7 @@ import (
 	"istio.io/istio/pilot/cmd/pilot-agent/config"
 	"istio.io/istio/pilot/cmd/pilot-agent/options"
 	"istio.io/istio/pilot/cmd/pilot-agent/status"
+
```
GregHanson, line 29: minor nit: there should not be a space between the import statements since they are all part of the same `istio.io` org

pilot/cmd/pilot-agent/app/cmd.go
```diff
@@ -49,22 +50,9 @@ const (
...
-	// proxy config flags (named identically)
-	serviceCluster         string
-	proxyLogLevel          string
-	proxyComponentLogLevel string
-	concurrency            int
-	templateFile           string
-	loggingOptions         = log.DefaultOptions()
-	outlierLogPath         string
+	loggingOptions = log.DefaultOptions()
+	proxyArgs      *options.ProxyArgs
```
howardjohn, line 55: why a pointer?

vince-0202: > why a pointer?

Cause PilotArgs in the pilot-discover is also a pointer,I'm not sure why, so I do it here.
Would it be better to not use pointer here?

GregHanson (approved):

PR #39135, opened 2022-05-26
Add well known credential uds socket to allow plugin external UDS SDS server

irisdingbj:
- add a well know credential uds socket path to to allow plug in external UDS SDS server
- create a static cluster if the credential uds socket exists
- if credential is named as sds://xxxx  and the credential uds socket exists it will use the external SDS server. otherwise, follow original flow

### DIFF
--- a/manifests/charts/gateways/istio-egress/templates/deployment.yaml
+++ b/manifests/charts/gateways/istio-egress/templates/deployment.yaml
@@ -227,6 +227,8 @@ spec:
           volumeMounts:
           - name: workload-socket
             mountPath: /var/run/secrets/workload-spiffe-uds
+          - name: credential-socket
+            mountPath: /var/run/secrets/credential-uds
           - name: workload-certs
             mountPath: /var/run/secrets/workload-spiffe-credentials
           - name: istio-envoy
@@ -270,6 +272,8 @@ spec:
       volumes:
       - emptyDir: {}
         name: workload-socket
+      - emptyDir: {}
+        name: credential-socket
       - emptyDir: {}
         name: workload-certs
 {{- if eq .Values.global.pilotCertProvider "istiod" }}
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -42,6 +42,7 @@ import (
 	"istio.io/istio/pkg/config/host"
 	"istio.io/istio/pkg/config/protocol"
 	"istio.io/istio/pkg/config/schema/kind"
+	"istio.io/istio/pkg/security"
 	"istio.io/istio/pkg/util/sets"
 )
 
@@ -179,7 +180,10 @@ func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *mod
 		}
 		clusters = append(clusters, patcher.insertedClusters()...)
 	}
-
+	// if credential socket exists, create a cluster for it
+	if proxy.Metadata != nil && proxy.Metadata.Raw[security.CredentialMetaDataName] == "true" {
+		clusters = append(clusters, cb.buildExternalSDSCluster(security.CredentialNameSocketPath))
+	}
 	for _, c := range clusters {
 		resources = append(resources, &discovery.Resource{Name: c.Name, Resource: util.MessageToAny(c)})
 	}
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -103,8 +103,9 @@ type ClusterBuilder struct {
 	proxyIPAddresses  []string                 // IP addresses on which proxy is listening on.
 	configNamespace   string                   // Proxy config namespace.
 	// PushRequest to look for updates.
-	req   *model.PushRequest
-	cache model.XdsCache
+	req                   *model.PushRequest
+	cache                 model.XdsCache
+	credentialSocketExist bool
 }
 
 // NewClusterBuilder builds an instance of ClusterBuilder.
@@ -135,6 +136,9 @@ func NewClusterBuilder(proxy *model.Proxy, req *model.PushRequest, cache model.X
 			}
 		}
 		cb.clusterID = string(proxy.Metadata.ClusterID)
+		if proxy.Metadata.Raw[security.CredentialMetaDataName] == "true" {
+			cb.credentialSocketExist = true
+		}
 	}
 	return cb
 }
@@ -995,7 +999,7 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 		if tls.CredentialName != "" {
 			// If  credential name is specified at Destination Rule config and originating node is egress gateway, create
 			// SDS config for egress gateway to fetch key/cert at gateway agent.
-			authn_model.ApplyCustomSDSToClientCommonTLSContext(tlsContext.CommonTlsContext, tls)
+			authn_model.ApplyCustomSDSToClientCommonTLSContext(tlsContext.CommonTlsContext, tls, cb.credentialSocketExist)
 		} else {
 			// If CredentialName is not set fallback to files specified in DR.
 			res := security.SdsCertificateConfig{
@@ -1034,7 +1038,7 @@ func (cb *ClusterBuilder) buildUpstreamClusterTLSContext(opts *buildClusterOpts,
 		if tls.CredentialName != "" {
 			// If  credential name is specified at Destination Rule config and originating node is egress gateway, create
 			// SDS config for egress gateway to fetch key/cert at gateway agent.
-			authn_model.ApplyCustomSDSToClientCommonTLSContext(tlsContext.CommonTlsContext, tls)
+			authn_model.ApplyCustomSDSToClientCommonTLSContext(tlsContext.CommonTlsContext, tls, cb.credentialSocketExist)
 		} else {
 			// If CredentialName is not set fallback to file based approach
 			if tls.ClientCertificate == "" || tls.PrivateKey == "" {
@@ -1272,4 +1276,35 @@ func defaultTransportSocketMatch() *cluster.Cluster_TransportSocketMatch {
 		Match:           &structpb.Struct{},
 		TransportSocket: xdsfilters.RawBufferTransportSocket,
 	}
+}
+
+// buildExternalSDSCluster generates a cluster that acts as external SDS server
+func (cb *ClusterBuilder) buildExternalSDSCluster(addr string) *cluster.Cluster {
+	ep := &endpoint.LbEndpoint{
+		HostIdentifier: &endpoint.LbEndpoint_Endpoint{
+			Endpoint: &endpoint.Endpoint{
+				Address: &core.Address{
+					Address: &core.Address_Pipe{
+						Pipe: &core.Pipe{
+							Path: addr,
+						},
+					},
+				},
+			},
+		},
+	}
+	c := &cluster.Cluster{
+		Name:                 security.SDSExternalClusterName,
+		ClusterDiscoveryType: &cluster.Cluster_Type{Type: cluster.Cluster_STATIC},
+		ConnectTimeout:       cb.req.Push.Mesh.ConnectTimeout,
+		LoadAssignment: &endpoint.ClusterLoadAssignment{
+			ClusterName: security.SDSExternalClusterName,
+			Endpoints: []*endpoint.LocalityLbEndpoints{
+				{
+					LbEndpoints: []*endpoint.LbEndpoint{ep},
+				},
+			},
+		},
+	}
+	return c
 }
--- a/pilot/pkg/security/model/authentication.go
+++ b/pilot/pkg/security/model/authentication.go
@@ -15,6 +15,7 @@
 package model
 
 import (
+	"strings"
 	"time"
 
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
@@ -76,7 +77,7 @@ var SDSAdsConfig = &core.ConfigSource{
 // from certificates referenced by credentialName in DestinationRule or Gateway.
 // Currently this is served by a local SDS server, but in the future replaced by
 // Istiod SDS server.
-func ConstructSdsSecretConfigForCredential(name string) *tls.SdsSecretConfig {
+func ConstructSdsSecretConfigForCredential(name string, credentialSocketExist bool) *tls.SdsSecretConfig {
 	if name == "" {
 		return nil
 	}
@@ -86,10 +87,40 @@ func ConstructSdsSecretConfigForCredential(name string) *tls.SdsSecretConfig {
 	if name == credentials.BuiltinGatewaySecretTypeURI+SdsCaSuffix {
 		return ConstructSdsSecretConfig(SDSRootResourceName)
 	}
+	// if credentialSocketExist exists and credentialName is using SDSExternalCredentialPrefix
+	// SDS will be served via SDSExternalClusterName
+	if credentialSocketExist && strings.HasPrefix(name, security.SDSExternalCredentialPrefix) {
+		return ConstructSdsSecretConfigForCredentialSocket(name)
+	}
 
 	return &tls.SdsSecretConfig{
 		Name:      credentials.ToResourceName(name),
 		SdsConfig: SDSAdsConfig,
+	}
+}
+
+// ConstructSdsSecretConfigForSDSEndpoint constructs SDS Secret Configuration based on CredentialNameSocketPath
+// if CredentialNameSocketPath exists, use a static cluster 'sds-external'
+func ConstructSdsSecretConfigForCredentialSocket(name string) *tls.SdsSecretConfig {
+	return &tls.SdsSecretConfig{
+		Name: name,
+		SdsConfig: &core.ConfigSource{
+			ConfigSourceSpecifier: &core.ConfigSource_ApiConfigSource{
+				ApiConfigSource: &core.ApiConfigSource{
+					ApiType:                   core.ApiConfigSource_GRPC,
+					SetNodeOnFirstMessageOnly: true,
+					TransportApiVersion:       core.ApiVersion_V3,
+					GrpcServices: []*core.GrpcService{
+						{
+							TargetSpecifier: &core.GrpcService_EnvoyGrpc_{
+								EnvoyGrpc: &core.GrpcService_EnvoyGrpc{ClusterName: security.SDSExternalClusterName},
+							},
+						},
+					},
+				},
+			},
+			ResourceApiVersion: core.ApiVersion_V3,
+		},
 	}
 }
 
@@ -220,11 +251,13 @@ func ApplyToCommonTLSContext(tlsContext *tls.CommonTlsContext, proxy *model.Prox
 
 // ApplyCustomSDSToClientCommonTLSContext applies the customized sds to CommonTlsContext
 // Used for building upstream TLS context for egress gateway's TLS/mTLS origination
-func ApplyCustomSDSToClientCommonTLSContext(tlsContext *tls.CommonTlsContext, tlsOpts *networking.ClientTLSSettings) {
+func ApplyCustomSDSToClientCommonTLSContext(tlsContext *tls.CommonTlsContext,
+	tlsOpts *networking.ClientTLSSettings, credentialSocketExist bool,
+) {
 	if tlsOpts.Mode == networking.ClientTLSSettings_MUTUAL {
 		// create SDS config for gateway to fetch key/cert from agent.
 		tlsContext.TlsCertificateSdsSecretConfigs = []*tls.SdsSecretConfig{
-			ConstructSdsSecretConfigForCredential(tlsOpts.CredentialName),
+			ConstructSdsSecretConfigForCredential(tlsOpts.CredentialName, credentialSocketExist),
 		}
 	}
 	// create SDS config for gateway to fetch certificate validation context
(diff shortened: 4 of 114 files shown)


irisdingbj:
@howardjohn @costinm  Would you please to see if this is the right direction following our discussion? Thanks！

pilot/pkg/features/pilot.go
```diff
@@ -622,6 +622,7 @@ var (
 	CanonicalServiceForMeshExternalServiceEntry = env.RegisterBoolVar("LABEL_CANONICAL_SERVICES_FOR_MESH_EXTERNAL_SERVICE_ENTRIES", false,
 		"If enabled, metadata representing canonical services for ServiceEntry resources with a location of mesh_external will be populated"+
 			"in the cluster metadata for those endpoints.").Get()
+	SDSEndpoint = env.RegisterStringVar("SDS_ADDR", "", "Address of the SDS server. Defaults to discoveryAddress").Get()
```
costinm, line 625: Nit: can you do a search for CA_ADDR - and add it next to it ?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -178,7 +179,12 @@ func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *mod
 		}
 		clusters = append(clusters, patcher.insertedClusters()...)
 	}
-
+	// if SDS_ADDR is defined and is using UDS, create a cluster for it
+	// if SDS_ADDR is defined and is using service name, there should already a cluster added via service discovery
+	externalSDSAddr := features.SDSEndpoint
+	if externalSDSAddr != "" && strings.HasPrefix(externalSDSAddr, authn_model.UnixAddressPrefix) {
```
costinm, line 185: Confused - what is the unix address doing ? 

There are 2 choices, either istiod configs using a cluster for the real SDS server, from envoy - which is a bit more complicated but probably a good idea, in which case there is no unix address involved.

Or Istiod doesn't even know about the SDS_ADDR - except injection template. And pilot-agent uses the SDS_ADDR internally to create a tunnel, like it does for Istiod. 

It seems you went with the first option, so no unix address ?

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -1286,3 +1287,35 @@ func defaultTransportSocketMatch() *cluster.Cluster_TransportSocketMatch {
 		TransportSocket: xdsfilters.RawBufferTransportSocket,
 	}
 }
+
+// generates a cluster that acts as external SDS server
+func (cb *ClusterBuilder) buildExternalSDSCluster(addr string) *cluster.Cluster {
+	addr = strings.TrimPrefix(addr, authn_model.UnixAddressPrefix)
```
costinm, line 1293: Similar comment - if you use unix socket and agent proxy, this is already done.
Otherwise - it should be a cluster with full destination rules applied, etc - i.e. normal code path.

pilot/pkg/networking/core/v1alpha3/cluster.go
costinm: Or 3rd choice - don't touch the agent, don't create a new UDS socket - instead agent talks with Istiod like before, and Istiod is federating.

costinm (commented):
I'm a bit confused about the goals/options. 

I thought you want different workloads to use different SDS server for secret - i.e. at inject time ( via annotations or other mechanism ) we add an SDS_ADDR, and agent just connects to it and keeps everything else the same. No change in Istiod or secret names - just agent.

If you want secret names to include sds: - which I believe doesn't pass validation for K8S Gateway, so probably sds- - it implies envoy will use 2 SDS servers, agent still need to create the proxy ( to have the UDS socket created ). That's significantly more complicated, and I don't think it's worth it. If you want a workload to be able to talk with multiple SDS servers at the same time - I suspect leaving agent as is and federating in Istiod is far better ( like we do for CAs ).

irisdingbj:
@costinm Currently Istiod will watch the gateway/DS configs, generate related `tls_certificate_sds_secret_config`  (using istiod itself  
as SDS server by default).   -- This is the place we want to replace it  other SDS server.   

Below is current config isitod generated :
```
            "tls_certificate_sds_secret_configs": [
             {
              "name": "kubernetes://httpbin-credential",
              "sds_config": {
               **"ads": {},**
               "resource_api_version": "V3"
              }
             }
            ]
```

So  we can not make this change only in agent side -- we need to modify isitod's default behaviour to use other SDS server instead.

There are two cases for the external SDS server: 
 1.  the external SDS server is a normal kubernetes service, in this case istiod will make it as a cluster via service discovery
2. the external SDS server is running as per-node daemon set and is on UDS, in this case we need create a related cluster config for this UDS based SDS server --which requies users to mount the UDS to the pod before hand like the current `./var/run/secrets/workload-spiffe-uds/socket` 


We can not make this change only in agent side, as the gateway config will be detected by istiod now and istiod will use itself as SDS server by default for credentialName.

irisdingbj:
> I'm a bit confused about the goals/options.
> 
> I thought you want different workloads to use different SDS server for secret - i.e. at inject time ( via annotations or other mechanism ) we add an SDS_ADDR, and agent just connects to it and keeps everything else the same. No change in Istiod or secret names - just agent.
> 
> If you want secret names to include sds: - which I believe doesn't pass validation for K8S Gateway, so probably sds- - it implies envoy will use 2 SDS servers, agent still need to create the proxy ( to have the UDS socket created ). That's significantly more complicated, and I don't think it's worth it. If you want a workload to be able to talk with multiple SDS servers at the same time - I suspect leaving agent as is and federating in Istiod is far better ( like we do for CAs ).

for k8s gateway, we can do things like:
```
  tls:
      mode: Terminate
      certificateRefs:
        kind: sds
        name: httpbin-credential
```
for istio gateway, we use:
```
    tls:
      mode: SIMPLE
      credentialName: sds://helloworld-credential
```

costinm:
Why not use same for both ?

costinm:
> @costinm Currently Istiod will watch the gateway/DS configs, generate related `tls_certificate_sds_secret_config`  (using istiod itself  
> as SDS server by default).   -- This is the place we want to replace it  with other SDS server.   
> 
> Below is current config isitod generated :
> ```
>             "tls_certificate_sds_secret_configs": [
>              {
>               "name": "kubernetes://httpbin-credential",
>               "sds_config": {
>                **"ads": {},**
>                "resource_api_version": "V3"
>               }
>              }
>             ]
> ```
> 
> So  we can not make this change only in agent side -- we need to modify isitod's default behaviour to use other SDS server instead.
> 
> There are two cases for the external SDS server: 
>  1.  the external SDS server is a normal kubernetes service, in this case istiod will make it as a cluster via service discovery

Why ? SDS is hanlded by agent, not as a cluster. 

How do you deal with auth if SDS server needs tokens, etc ? And why do things differently ?


> 2. the external SDS server is running as per-node daemon set and is on UDS, in this case we need create a related cluster config for this UDS based SDS server --which requies users to mount the UDS to the pod before hand like the current `./var/run/secrets/workload-spiffe-uds/socket` 

And in this case again we don't need to change istiod.


> 




> 
> We can not make this change only in agent side, as the gateway config will be detected by istiod now and istiod will use itself as SDS server by default for credentialName.
> 


That's a choice we can make. Do we select the SDS provider via pod config, like CA provider ? Or via gateway ?

Since we only support one sds provider - it is clear a pod can only have one setting. So gateway settings can't result in different SDS servers.

irisdingbj:
> > We can not make this change only in agent side, as the gateway config will be detected by istiod now and istiod will use itself as SDS server by default for credentialName.
> 
> That's a choice we can make. Do we select the SDS provider via pod config, like CA provider ? Or via gateway ?
> 
> Since we only support one sds provider - it is clear a pod can only have one setting. So gateway settings can't result in different SDS servers.

Agree we can leverage the exsting SDS server  for workload  in `./var/run/secrets/workload-spiffe-uds/socket`  -- which has been a static cluster in our bootstrap template.

**The issue here is how we change the current istiod's default behaviour that it will use istioid itself as SDS server for `credentialName`?**    

"tls_certificate_sds_secret_configs": [
             {
              "name": "kubernetes://httpbin-credential",
              "sds_config": {
               **"ads": {},**
               "resource_api_version": "V3"
              }
             }
            ]

We need a flag to notify istiod to change this default behaviour to use ``./var/run/secrets/workload-spiffe-uds/socke` instead, right?

howardjohn:
My big question is how does the auth work for non-UDS?

irisdingbj:
@howardjohn  @costinm  

> My big question is how does the auth work for non-UDS?

I am limiting  that only UDS is supported.

irisdingbj:
@howardjohn @costinm  Change in this PR that only UDS type SDS server is allowed. PTAL, Thanks!

pilot/pkg/security/model/authentication.go
```diff
@@ -139,6 +154,41 @@ var (
...
+						{
+							TargetSpecifier: &core.GrpcService_EnvoyGrpc_{
+								EnvoyGrpc: &core.GrpcService_EnvoyGrpc{ClusterName: clusterName},
+							},
+						},
+					},
+				},
+			},
+			ResourceApiVersion:  core.ApiVersion_V3,
+			InitialFetchTimeout: durationpb.New(time.Second * 0),
```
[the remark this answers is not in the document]
irisdingbj: Thanks! Good catch! To address this, I added a new env_var to let user customize this according to their needs and default value is 15s (which is also the default value for envoy)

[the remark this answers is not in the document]
howardjohn: IMO we should not customize; use the same as the current gateway certs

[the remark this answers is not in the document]
irisdingbj: I am just worrying some external SDS server might take longer than the current gateway cert setting (15s).   In this case, they have no way to work around it.  Any suggestions if this happens?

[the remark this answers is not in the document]
irisdingbj: @howardjohn removed the timeout var and keep it same with current gateway certs. PTAL. Thanks

irisdingbj:
@howardjohn @costinm PTAL, Thanks!

howardjohn (approved):

irisdingbj:
@hzxuzhonghu  In case you can leave your comments here.

hzxuzhonghu:
I am not sure why cannot you reuse the `WorkloadIdentitySocketPath` there, instead define another env to indicate it. 
`SDS_ADDR` is confusing, we already have `CA_ADDR`.   IIUC the case here, we can use sds agent to connect to any server, no matter node local or central .

irisdingbj:
`WorkloadIdentitySocketPath ` is a constant and I have used `WorkloadIdentitySocketPath ` in this PR  in case user want to use the same Spire SDS server if this is what you mean.    

The reason we add `SDS_ADDR` is  because that we at least need a way to tell istiod -  when it generate SDS config for gateways(CredentialName), it need to use anohter SDS server instead of itself.

irisdingbj:
@hzxuzhonghu  
By the way, this PR has been discussed in last week's WG meeting and current implementation is what suggested in the meeting.

irisdingbj:
@hzxuzhonghu Please let us know if you still have other comments/suggestions for this Or if it  is good if we remove the 'hold` label from it?  Thanks!

hzxuzhonghu:
1. we should not introduce a confusing env.

 >OR IIUC the case here, we can use sds agent to connect to any server, no matter node local or central .

3. The requirement can be satisfied by connecting to a global existing SDS server.  Also not clear to us why UDS is a must.

irisdingbj:
@hzxuzhonghu  Thanks for the comments.   

> 1. we should not introduce a confusing env.

What is your suggestion here?  Do you suggest other value to replace `SDS_ADDR`? 

> > OR IIUC the case here, we can use sds agent to connect to any server, no matter node local or central .
> 3. The requirement can be satisfied by connecting to a global existing SDS server.  Also not clear to us why UDS is a must.


The reaon why  only UDS is supported is due to Auth.  Please see above comments from @howardjohn

howardjohn:
> OR IIUC the case here, we can use sds agent to connect to any server, no matter node local or central .

AFAIK they want the workload cert and gateway cert decoupled

hzxuzhonghu:
>>OR IIUC the case here, we can use sds agent to connect to any server, no matter node local or central .

>AFAIK they want the workload cert and gateway cert decoupled

Not sure I understand correctly, we never couple them. And we have authorization for each proxy.

howardjohn:
SDS agent only handles workload cert today

irisdingbj:
@hzxuzhonghu Any concerns remaining regarding this?

hzxuzhonghu:
The concern is expressed again and again,  not to say I do not see any doc saying spire can delegate certificate fetching(I know it can be used to sign a cert, but I remember you meant to fetch a cert already there in a store).

and I have to supplement the API is clear.

```
type ServerTLSSettings struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// If set to true, the load balancer will send a 301 redirect for
	// all http connections, asking the clients to use HTTPS.
	HttpsRedirect bool `protobuf:"varint,1,opt,name=https_redirect,json=httpsRedirect,proto3" json:"https_redirect,omitempty"`
	// Optional: Indicates whether connections to this port should be
	// secured using TLS. The value of this field determines how TLS is
	// enforced.
	Mode ServerTLSSettings_TLSmode `protobuf:"varint,2,opt,name=mode,proto3,enum=istio.networking.v1alpha3.ServerTLSSettings_TLSmode" json:"mode,omitempty"`
	// REQUIRED if mode is `SIMPLE` or `MUTUAL`. The path to the file
	// holding the server-side TLS certificate to use.
	ServerCertificate string `protobuf:"bytes,3,opt,name=server_certificate,json=serverCertificate,proto3" json:"server_certificate,omitempty"`
	// REQUIRED if mode is `SIMPLE` or `MUTUAL`. The path to the file
	// holding the server's private key.
	PrivateKey string `protobuf:"bytes,4,opt,name=private_key,json=privateKey,proto3" json:"private_key,omitempty"`
	// REQUIRED if mode is `MUTUAL`. The path to a file containing
	// certificate authority certificates to use in verifying a presented
	// client side certificate.
	CaCertificates string `protobuf:"bytes,5,opt,name=ca_certificates,json=caCertificates,proto3" json:"ca_certificates,omitempty"`
	// For gateways running on Kubernetes, the name of the secret that
	// holds the TLS certs including the CA certificates. Applicable
	// only on Kubernetes. The secret (of type `generic`) should
	// contain the following keys and values: `key:
	// <privateKey>` and `cert: <serverCert>`. For mutual TLS,
	// `cacert: <CACertificate>` can be provided in the same secret or
	// a separate secret named `<secret>-cacert`.
	// Secret of type tls for server certificates along with
	// ca.crt key for CA certificates is also supported.
	// Only one of server certificates and CA certificate
	// or credentialName can be specified.
	CredentialName string `protobuf:"bytes,10,opt,name=credential_name,json=credentialName,proto3" json:"credential_name,omitempty"`
	// A list of alternate names to verify the subject identity in the
	// certificate presented by the client.
	SubjectAltNames []string `protobuf:"bytes,6,rep,name=subject_alt_names,json=subjectAltNames,proto3" json:"subject_alt_names,omitempty"`
	// An optional list of base64-encoded SHA-256 hashes of the SKPIs of
	// authorized client certificates.
	// Note: When both verify_certificate_hash and verify_certificate_spki
	// are specified, a hash matching either value will result in the
	// certificate being accepted.
	VerifyCertificateSpki []string `protobuf:"bytes,11,rep,name=verify_certificate_spki,json=verifyCertificateSpki,proto3" json:"verify_certificate_spki,omitempty"`
	// An optional list of hex-encoded SHA-256 hashes of the
	// authorized client certificates. Both simple and colon separated
	// formats are acceptable.
	// Note: When both verify_certificate_hash and verify_certificate_spki
	// are specified, a hash matching either value will result in the
	// certificate being accepted.
	VerifyCertificateHash []string `protobuf:"bytes,12,rep,name=verify_certificate_hash,json=verifyCertificateHash,proto3" json:"verify_certificate_hash,omitempty"`
	// Optional: Minimum TLS protocol version.
	MinProtocolVersion ServerTLSSettings_TLSProtocol `protobuf:"varint,7,opt,name=min_protocol_version,json=minProtocolVersion,proto3,enum=istio.networking.v1alpha3.ServerTLSSettings_TLSProtocol" json:"min_protocol_version,omitempty"`
	// Optional: Maximum TLS protocol version.
	MaxProtocolVersion ServerTLSSettings_TLSProtocol `protobuf:"varint,8,opt,name=max_protocol_version,json=maxProtocolVersion,proto3,enum=istio.networking.v1alpha3.ServerTLSSettings_TLSProtocol" json:"max_protocol_version,omitempty"`
	// Optional: If specified, only support the specified cipher list.
	// Otherwise default to the default cipher list supported by Envoy.
	CipherSuites []string `protobuf:"bytes,9,rep,name=cipher_suites,json=cipherSuites,proto3" json:"cipher_suites,omitempty"`
}

**Comment by irisdingbj** [comment]
hmm..  .Can you be more precise for the concerns you have?  

Through all your comments here: 

- I am not sure why cannot you reuse the `WorkloadIdentitySocketPath` there, instead define another env to indicate it. 
- `SDS_ADDR` is confusing, we already have `CA_ADDR`.   OR IIUC the case here, we can use sds agent to connect to any server, no matter node local or central . 
- we should not introduce a confusing env.
- OR IIUC the case here, we can use sds agent to connect to any server, no matter node local or central.

You can see response for  all your listed concerns. 

For your newly posted concerns regarding spire, pls see here: https://istio.io/latest/docs/ops/integrations/spire/

**Comment by zhlsunshine** [comment]
Hi @hzxuzhonghu , as we have some discussion via slack, I have explained the solution on how to protect workload mTLS and gateway credential data via SDS server in our solution and `SDS_ADDR` can help user's workloads to locate the remote SDS server where should they can fetch the credential data, such as private key pairs. So, do you have any other comment about this PR? If yes, could you please help to leave your comment here? Thank you so much for your time!

**Comment by costinm** [comment]
I don't think we need to implement another SDS server as part of Istio. We have one - this code is to integrate with other projects.

I thought about this more: I think having the UDS as first step is not bad, but I would drop SDS_ADDR setting and instead define a well-known location, like we have for the certs and the uds socket used by spire. If the file doesn't exist - fallback. This eliminates some risks and conplexity. The agent will still tell Istiod it found the socket, via node meta. 

Regarding who mounts the socket volume - we can also leave it out of Istio. Like spire, certmanager - it's their responsibility to inject and mount. An SDS provider should be similar with a cert provider - and also not specific to Istio.

**Comment by costinm** [comment]
I would also try to talk with spire and certmanager - maybe file a bug for each - asking for support in their implementations, similar or based on what they do for workloads. Ideally Istio wouldn't be in any cert business - except as a minimal 'fallback'.

**Comment by irisdingbj** [comment]
> I can't agree more on we should integrate with a popular ecosystem. unfortunately, now spire seems not support it. But if we let this in, I am sure we can not remove `SDS_ADDR` easily later to adapt to others.

 The purpose of this PR is providing ways to integrate with other projects. 

 We should not make the integration door closed. Only if you open doors, ecostem will be inspired to implement it.  I believe extensibility is a key point to Istio. 

 We can drop SDS_ADDR and define a well-known location as @costinm suggested. The agent will check the socket and tell Istiod it found the socket, via node meta.

**Comment by irisdingbj** [comment]
> I would also try to talk with spire and certmanager - maybe file a bug for each - asking for support in their implementations, similar or based on what they do for workloads. Ideally Istio wouldn't be in any cert business - except as a minimal 'fallback'.

+1 for this

**Comment by costinm** [comment]
CA_ADDR and SDS_ADDR ( and all other env variables ) are not meant to be
part of the API surface - we have ProxyConfig and the protos as API,
annotations, labels,
but not the env variables. I do agree with you we should not add it - but
for different reasons.

You are right - we may need to support it for backward compat for a long
time - but eventually we can remove it. My proposed alternative - a 'well
known' path
also has the same problem, whatever name we decide may not be the one that
is adopted. For workload certs we currently use the naming convention
defined by Spire for UDS and the one used by GKE for platform-generated
sockets - neither is adopted by K8S or a real standard. If a standard
emerges
we will support it - but also check the old paths for a long time.

**Comment by irisdingbj** [comment]
Thanks @costinm 
envoy supported abstract UDS already.  However abstract uds has no access control according to https://man7.org/linux/man-pages/man7/unix.7.html. 

So It is better we continue using the filepath way.  ` ./var/run/secrets/sds/ ` is a good choise.  WDYT?

**Comment by costinm** [comment]
Not sure what you mean by no permission -one of the benefit of UDS is the ancillary permission that allows both ends to be pretty strongly authenticated. Pretty sure this works on both path and abstract.

If you mean file-based permission - that's pretty weak.

**Comment by costinm** [comment]
I don't mind continuing to use path based UDS - just to provide an option in future, if we choose to optimize the agent footprint and take advantage of istio CNI.

**Comment by irisdingbj** [comment]
@costinm @howardjohn ^^, Thanks!

**Comment by irisdingbj** [comment]
@costinm A ping on this, Thanks!

File: pkg/istio-agent/agent.go [hunk 5]
```diff
@@ -267,8 +267,11 @@ func (a *Agent) generateNodeMetadata() (*model.Node, error) {
 	})
 }
 
-func (a *Agent) initializeEnvoyAgent(ctx context.Context) error {
+func (a *Agent) initializeEnvoyAgent(ctx context.Context, credentialSocketExists bool) error {
```
costinm, line 270: I think the param should be instead a field in agent. It's strange for all other params to be in Agent and this to be passed around as param.

pkg/istio-agent/agent.go
```diff
@@ -443,19 +452,18 @@ func (a *Agent) Run(ctx context.Context) (func(), error) {
 	}
 
 	if a.cfg.GRPCBootstrapPath != "" {
-		if err := a.generateGRPCBootstrap(); err != nil {
+		if err := a.generateGRPCBootstrap(credentialSocketExists); err != nil {
```
costinm, line 455: GRPC does not support SDS. So false in all cases.

pkg/istio-agent/agent.go
```diff
@@ -592,9 +600,13 @@ func (a *Agent) initLocalDNSServer() (err error) {
 	return nil
 }
 
-func (a *Agent) generateGRPCBootstrap() error {
+func (a *Agent) generateGRPCBootstrap(credentialSocketExists bool) error {
 	// generate metadata
 	node, err := a.generateNodeMetadata()
+	if credentialSocketExists {
```
costinm, line 606: See above. GRPC doesn't support SDS.

costinm (approved):
Sorry for the delay, vacation and other things. Looks good - left few cosmetic comments that can be addressed later.

manifests/charts/gateways/istio-egress/templates/deployment.yaml
```diff
@@ -227,6 +227,8 @@ spec:
           volumeMounts:
           - name: workload-socket
             mountPath: /var/run/secrets/workload-spiffe-uds
+          - name: credential-socket
+            mountPath: /var/run/secrets/credential-uds
```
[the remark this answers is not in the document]
irisdingbj: Thanks @costinm I will modify this later.

PR #39137, opened 2022-05-26
separate webhook patching from namespace controller

ramaraochavali:
Namespace controller is an optional component only started when kubernetes CA is used but webhook patching is needed in all cases.
https://github.com/istio/istio/pull/38283 combined both of them. 

This PR fixes it.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/leaderelection/leaderelection.go
+++ b/pilot/pkg/leaderelection/leaderelection.go
@@ -37,6 +37,7 @@ import (
 // Various locks used throughout the code
 const (
 	NamespaceController     = "istio-namespace-controller-election"
+	WebhookPatcher          = "istio-webhook-patcher-election"
 	ServiceExportController = "istio-serviceexport-controller-election"
 	// This holds the legacy name to not conflict with older control plane deployments which are just
 	// doing the ingress syncing.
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (approved):
LGTM

howardjohn:
Confused - why did we add leader at all? Validating patcher has no leader election...

cc @frankbu

ramaraochavali:
We discussed about adding it some time ago.  I thought the multi cluster
change that Frank did was intending to do it. I just fixed the bug by
separating it from nscontroller.  If you think there is an issue with
leader election I can remove it but AFAIK it should be fine

howardjohn:
I am not sure I get it. Each revision has its own webhook and only patches their own webhook, right? So how can we have a single leader?

hzxuzhonghu:
Make sense, missed that

ramaraochavali:
You are right. I thought we discussed here
https://github.com/istio/istio/pull/35243 that we want to have to leader
election.  And frank intentionally changed it to use leader election in the
PR that Frank changed
 I can remove leader election though but my intention of this PR was to
separate patching from NS controller
I will remove it on Tuesday

PR #39151, opened 2022-05-26
Fix error msg noise in istiod log

### LINKED ISSUE #38872: Multicluster config generates error mesage noise in istiod log
### Bug Description

With the latest changes to support istiodless remotes, istiod is generating undesirable noise in the error logs:
```
2022-05-11T21:35:14.298487Z	error	klog	error retrieving resource lock istio-system/istio-namespace-controller-election: configmaps "istio-namespace-controller-election" is forbidden: User "system:serviceaccount:istio-system:istio-reader-service-account" cannot get resource "configmaps" in API group "" in the namespace "istio-system"
2022-05-11T21:35:21.976070Z	error	klog	error retrieving resource lock istio-system/istio-namespace-controller-election: configmaps "istio-namespace-controller-election" is forbidden: User "system:serviceaccount:istio-system:istio-reader-service-account" cannot get resource "configmaps" in API group "" in the namespace "istio-system"
```

The errors are because of [this change](https://github.com/istio/istio/pull/38738#discussion_r866283191).

We should figure out a more graceful way to handle this, without generating error messages in the log.

### Version

```prose
master
```

### Additional Information

_No response_

frankbu:

Fixes #38872


pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -280,6 +283,29 @@ func (m *Multicluster) ClusterAdded(cluster *multicluster.Cluster, clusterStopCh
...
+				Namespace: m.opts.SystemNamespace,
+				Verb:      "create",
+				Group:     "",
+				Resource:  "configmaps",
+			},
+		},
+	}
+	response, err := client.Kube().AuthorizationV1().SelfSubjectAccessReviews().Create(context.TODO(), s, metav1.CreateOptions{})
+	if err != nil {
+		log.Errorf("could not determine leadership permission: %v", err)
```
howardjohn, line 303: This means a transient error will permanently change behavior. IMO this is pretty dangerous

frankbu: Could return true in this case, which would then result in the same permission errors in the log that this PR is trying to suppress, but at least we've suppressed them most of the time.

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -200,7 +202,10 @@ func (m *Multicluster) ClusterAdded(cluster *multicluster.Cluster, clusterStopCh
 	// run after WorkloadHandler is added
 	m.opts.MeshServiceController.AddRegistryAndRun(kubeRegistry, clusterStopCh)
 
-	if m.startNsController && (features.ExternalIstiod || localCluster) {
+	canLead := localCluster || m.checkCanLead(client)
```
hzxuzhonghu, line 205: IMO, this code is as bad as shouldStartNsController, it is hard to maintain.

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -314,6 +319,31 @@ func (m *Multicluster) initializeCluster(cluster *multicluster.Cluster, kubeRegi
...
+		},
+	}
+	log.Infof("check resource:\n%v", s)
+	response, err := client.Kube().AuthorizationV1().SelfSubjectAccessReviews().Create(context.TODO(), s, metav1.CreateOptions{})
+	log.Infof("check response:\n%v", response)
+	if err != nil {
+		log.Errorf("could not determine leadership permission: %v", err)
+		return true
+	}
+	return response.Status.Allowed
```
frankbu, line 344: This always retruns false???

@howardjohn do you see anything wrong with the above SelfSubjectAccessReview call?

howardjohn: Not really, seems correct to me. I am, however, fundamentally opposed with doing this type of check rather than having Istiod be aware of what role it has.

frankbu: Yeah, OK, I'm not crazy about this approach either.

Issue #39142, opened 2022-05-26
gateway-api gateway can not get secret in other namespace caused timeout?


13567436138:
### Bug Description

gateway-api gateway can not get secret in other namespace caused timeout?
I created ReferencePolicy??? 

openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048  -keyout cert.key -out cert.crt  -config  openssl.conf

secret

kubectl create -n istio secret tls bookinfo-gateway-secret  --key ./cert.key --cert=./cert.crt

referencepolicies/gw-listeners-tls-certificateRefs-namespace.yaml

kubectl apply -f gw-listeners-tls-certificateRefs-namespace.yaml -n istio-system

```
kind: Gateway
apiVersion: gateway.networking.k8s.io/v1alpha2
metadata:
  name: bookinfo
spec:
  addresses:
  - value: istio-ingressgateway.istio-system.svc.cluster.local
    type: Hostname
  gatewayClassName: istio
  listeners:  
  - hostname: "bookinfo.com"
    port: 443
    name: bookinfo-gw
    protocol: HTTPS
    tls:
      options: {}
      certificateRefs:
      - name: bookinfo-gateway-secret
        namespace: istio
        kind: Secret
    allowedRoutes:
      namespaces:
        from: All
```

rp-secret.yaml

kubectl apply -f rp-secret.yaml -n istio

```
kind: ReferencePolicy
apiVersion: gateway.networking.k8s.io/v1alpha2
metadata:
  name: rp-secret
  namespace: istio
spec:
  from:
  - group: gateway.networking.k8s.io
    kind: Gateway
    namespace: istio-system
  to:
  - group: ""
    kind: Secret
    name: bookinfo-gateway-secret
```

gatewayclass/istio-gc.yaml

kubectl apply -f istio-gc.yaml -n istio-system

```
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: GatewayClass
metadata:
  name: istio
spec:
  controllerName: istio.io/gateway-controller
```

referencepolicies/bookinfo-httproute-system.yaml

kubectl apply -f bookinfo-httproute-system.yaml -n istio

```
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: HTTPRoute
metadata:
  name: bookinfo
spec:
  parentRefs:
  - name: bookinfo
    namespace: istio-system
    kind: Gateway
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
[...]

PR #39157, opened 2022-05-27
gateway-api: recompute verified certs on ReferencePolicy change

dddddai:
Recompute verified certs/push sds on ReferencePolicy change
Fixes https://github.com/istio/istio/issues/39142

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pilot/pkg/xds/sds.go
```diff
@@ -64,7 +64,7 @@ func sdsNeedsPush(updates model.XdsUpdates) bool {
 	if len(updates) == 0 {
 		return true
 	}
-	if len(model.ConfigNamesOfKind(updates, gvk.Secret)) > 0 {
+	if model.ConfigsHaveKind(updates, gvk.Secret) || model.ConfigsHaveKind(updates, gvk.ReferencePolicy) {
```
hzxuzhonghu, line 67: Wait, IIUC ReferencePolicy can reference any resource, how could you say it will influence sds

pilot/pkg/xds/ads.go
```diff
@@ -628,7 +628,7 @@ func (s *DiscoveryServer) computeProxyState(proxy *model.Proxy, request *model.P
 			switch conf.Kind {
 			case gvk.ServiceEntry, gvk.DestinationRule, gvk.VirtualService, gvk.Sidecar, gvk.HTTPRoute, gvk.TCPRoute:
 				sidecar = true
-			case gvk.Gateway, gvk.KubernetesGateway, gvk.GatewayClass:
+			case gvk.Gateway, gvk.KubernetesGateway, gvk.GatewayClass, gvk.ReferencePolicy:
```
hzxuzhonghu, line 631: same confusion here

pilot/pkg/xds/sds.go
dddddai: I did so because now we only supports reference policies for gateway->secret, but I think we will implement more policies

Fixed

pilot/pkg/xds/ads.go
```diff
@@ -636,6 +636,10 @@ func (s *DiscoveryServer) computeProxyState(proxy *model.Proxy, request *model.P
 			case gvk.Ingress:
 				sidecar = true
 				gateway = true
+			case gvk.ReferencePolicy:
+				if rp := s.Env.ConfigStore.Get(gvk.ReferencePolicy, conf.Name, conf.Namespace); isSecretReferencePolicy(rp) {
+					gateway = true
+				}
```
howardjohn, line 642:
```suggestion
					gateway = true
```

I would just keep it simple here like the rest. This computation is not *that* expensive that we need to be this precise

pilot/pkg/xds/sds.go
```diff
@@ -60,13 +60,18 @@ func (sr SecretResource) Cacheable() bool {
...
+func sdsNeedsPush(updates model.XdsUpdates, configStore model.ConfigStore) bool {
 	if len(updates) == 0 {
 		return true
 	}
-	if len(model.ConfigNamesOfKind(updates, gvk.Secret)) > 0 {
+	if model.ConfigsHaveKind(updates, gvk.Secret) {
 		return true
 	}
+	for conf := range model.ConfigsOfKind(updates, gvk.ReferencePolicy) {
+		if rp := configStore.Get(gvk.ReferencePolicy, conf.Name, conf.Namespace); isSecretReferencePolicy(rp) {
```
howardjohn, line 71: `configStore.Get` is not safe here; everything in the push should only be computed from PushContext which is a point in time snapshot. Using the configstore directly leads to some race conditions that make config generation non-determinstic and break our cache

howardjohn: Basically we can just check the type, which is what we do for all other `xxxNeedsPush`

dddddai: That's what I did in https://github.com/istio/istio/commit/a1e6330ec02134345c21554119dfb30248d5edba, would let @hzxuzhonghu confirm

dddddai:
Hello @hzxuzhonghu, could we move this foward if you agree with John's comments?

pilot/pkg/xds/sds.go
hzxuzhonghu: Yep, this is a superset,  not a big deal since this is not used widely

PR #39164, opened 2022-05-27
Cleanup bootstrap ca/certs init

hzxuzhonghu:

the target is to make the code more readable, this includes:

1. args cleanup
2. function abstraction

### DIFF
--- a/pilot/pkg/bootstrap/istio_ca.go
+++ b/pilot/pkg/bootstrap/istio_ca.go
@@ -232,10 +232,8 @@ func (s *Server) loadCACerts(caOpts *caOptions, dir string) error {
 	signingKeyFile := path.Join(dir, ca.CAPrivateKeyFile)
 	if _, err := os.Stat(signingKeyFile); err == nil {
 		return nil
-	} else {
-		if !os.IsNotExist(err) {
-			return fmt.Errorf("signing key file %s already exists", signingKeyFile)
-		}
+	} else if !os.IsNotExist(err) {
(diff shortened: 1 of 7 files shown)


hzxuzhonghu:
@lei-tang @shankgan

pilot/pkg/bootstrap/istio_ca.go
```diff
@@ -229,18 +220,22 @@ func detectAuthEnv(jwt string) (*authenticate.JwtPayload, error) {
...
-func (s *Server) loadRemoteCACerts(caOpts *caOptions, dir string) error {
+func (s *Server) loadCACerts(caOpts *caOptions, dir string) error {
 	if s.kubeClient == nil {
 		return nil
 	}
 
 	signingKeyFile := path.Join(dir, ca.CAPrivateKeyFile)
-	if _, err := os.Stat(signingKeyFile); !os.IsNotExist(err) {
-		return fmt.Errorf("signing key file %s already exists", signingKeyFile)
+	if _, err := os.Stat(signingKeyFile); err == nil {
```
costinm, line 233: Looking at this logic: long ago we were mounting the secrets in the pod, and Istiod was not watching secrets. We now watch secrets - the whole file path and delployment.yaml could be cleaned up. However - who knows what will break...

I think at some point we should start deprecating various code paths like this - and add some metric or status with all the obsolete/deprecated features/code paths.

hzxuzhonghu: Thanks for the input. I can not agree more.

PR #39165, opened 2022-05-27
remove isCADisabled

hzxuzhonghu:

### DIFF
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -1009,11 +1009,6 @@ func getDNSNames(args *PilotArgs, host string) []string {
 
 // createPeerCertVerifier creates a SPIFFE certificate verifier with the current istiod configuration.
 func (s *Server) createPeerCertVerifier(tlsOptions TLSOptions) (*spiffe.PeerCertVerifier, error) {
-	if tlsOptions.CaCertFile == "" && s.CA == nil && features.SpiffeBundleEndpoints == "" {
-		// Running locally without configured certs - no TLS mode
-		return nil, nil
-	}
-	peerCertVerifier := spiffe.NewPeerCertVerifier()
 	var rootCertBytes []byte
 	var err error
 	if tlsOptions.CaCertFile != "" {
@@ -1035,7 +1030,9 @@ func (s *Server) createPeerCertVerifier(tlsOptions TLSOptions) (*spiffe.PeerCert
 		}
 	}
 
+	var peerCertVerifier *spiffe.PeerCertVerifier
 	if len(rootCertBytes) != 0 {
+		peerCertVerifier = spiffe.NewPeerCertVerifier()
 		err := peerCertVerifier.AddMappingFromPEM(spiffe.GetTrustDomain(), rootCertBytes)
 		if err != nil {
 			return nil, fmt.Errorf("add root CAs into peerCertVerifier failed: %v", err)
@@ -1047,6 +1044,9 @@ func (s *Server) createPeerCertVerifier(tlsOptions TLSOptions) (*spiffe.PeerCert
 			features.SpiffeBundleEndpoints, []*x509.Certificate{})
 		if err != nil {
 			return nil, err
+		}
+		if peerCertVerifier == nil {
+			peerCertVerifier = spiffe.NewPeerCertVerifier()
 		}
 		peerCertVerifier.AddMappings(certMap)
 	}
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -1256,24 +1256,6 @@ func (s *Server) initWorkloadTrustBundle(args *PilotArgs) error {
 	return nil
 }
 
-// isCADisabled returns whether CA functionality is disabled in istiod.
-// It returns true only if istiod certs is signed by Kubernetes or
-// workload certs are signed by external CA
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
Can you give some more context? its not clear how/why you would have namespace controller (writing root cert) enabled when there is no CA (which creates the root cert). Where does it come from otherwise?

hzxuzhonghu:
@howardjohn  We check the `features.PilotCertProvider` 

Yesterday, talked with @irisdingbj . I am going to remove the `ENABLE_CA_SERVER`,  `EXTERNAL_ISTIOD`.  These are both legacy for external istiod, now they are meaningless.

costinm (approved):
I would assume the roots will be from MeshConfig ? 

This is actually a very good change - we should replicate the roots even if Citadel is disabled and a different (external) CA is used for pilot or mesh certificates. 

There are few problems with the current scheme of sending ProxyConfig to agent with the mesh roots:
- agent still needs the roots of Istiod to connect - which doesn't have to be a public certificate but may be signed with a private CA (just not citadel)
- proxyless gRPC and 'raw envoy' do not know about ProxyConfig. Using the agent proxy has
benefits, but it should not be required.
- complexity. 
- as we integrate with platform/external CA providers we can't completely rely on the external provider to provision the roots, since MeshConfig API defines additional roots. 

I think we need to go one step further - and also replicate the trust domain aliases and few other settings that define the mesh behavior to the same config map, to reduce the injection complexity.

costinm (changes requested):
Actually - changing this to a hold, can you confirm the master election for the ns controller has proper checks for 'default', i.e. the election can't be won by a random revision ?

We may also need to revisit the merged mesh config - which is off by default, if the NS controller is used to replicate CAs defined in MeshConfig ( plus whatever non-citadel CAs are provisioned on Istiod ) - we want to make sure they are from a mesh-wide config map, not revisioned one ( mesh config is per revision, so depending on which revision wins the election we can have different roots)

hzxuzhonghu:
@costinm  Yes, leaderelection has considered revision here https://github.com/istio/istio/blob/master/pilot/pkg/serviceregistry/kube/controller/multicluster.go#L203-L209

costinm:
I'm not entirely sure what that code does - but it needs to have the same conditions, i.e. istiod should not win election if it didn't start the controller - and should NOT be revisioned - if 2 revs have different roots bad things happen.  
Default istiod should win the election and sync. 

What is the 'm.revision' parameter do ?

costinm:
As an example:
- istiod default revision using citadel 
- istiod external using mounted certificates
- istiod rev=foo using another mounted certificate

Who will win the election ? If all 3 participate, depending on timing/restarts/etc any can become leader and break the others.

Ideally only 'default' ( either default install or a revision that is tagged as default ) will win the election, and somehow sync all the roots. I think we need to start writing not only the Istiod root ( Citadel or the mounted ), but also all roots in MeshConfig.ProxyConfig - and 'external' and 'foo' revision roots can be added to the default mesh config.

It's pretty tricky and risky, I wish we could just add a separate configmap or CRD with the roots of trusts and have a more deterministic behavior.

hzxuzhonghu:
Currently it is the default revision one will win the election

```
	PrioritizedLeaderElection = env.RegisterBoolVar("PRIORITIZED_LEADER_ELECTION", true,
		"If enabled, the default revision will steal leader locks from non-default revisions").Get()
```


>I think we need to start writing not only the Istiod root ( Citadel or the mounted ), but also all roots in MeshConfig.ProxyConfig - and 'external' and 'foo' revision roots can be added to the default mesh config.

This is a good suggestion, I can file a issue to init discussion. I think this should be deeply thought.  Should not couple with the minor change here

costinm:
Agreed, this is a good change and we should make it. Leader election is
very tricky - and the code got even more complicated...

Is there any scenario where a user would disable prioritized leader
election ? Can we add an extra check, and disable the NS controller if
prioritized leader election is off ?

Also would be good to add some comments and docs somewhere, to make it a
bit more clear what is expected ( i.e. 'default' revision will replicate
it's root - which
can be mounted or citadel ). If the default has mounted certs - citadel
root will NOT be replicated if I understand the code correctly - so all
other revisions will stop
working unless they have the same mount ?

Would be great if John takes another look at this.

costinm:
The implication - if a user changes the default tag, will that result in a
different ns controller replicating a different CA ( if the noew revision
that is default
doesn't have mounted certificates) ?

If the master election doesn't take into account revision with the default
tag - we may have other problems...

Citadel had the special property of being a singleton ( at least per
istio-system namespace ). Having the ns controller replicate something that
is not
singleton is the problem I think. With singleton - it doesn't matter too
much who replicates.

I understand the use case for this PR is to support other CAs providing the
Istiod DNS certificate - and we want the ns controller to
replicate them. Could we require that in the case of an external CA, the
root ( or roots ) are saved into the citadel Secret, and the ns
controller replicate it ? In other words, even if we don't use citadel as a
CA, we still use the Secret as singleton source of truth for
all the roots ? That will reduce some of the risk and confusion.

hzxuzhonghu:
@costinm It is a minor change now

hzxuzhonghu:
ping @costinm a cleanup of confusing `isCADisabled`

costinm:
Also - is the leader election now considering default tag ( i.e. default
revision - or a revision having the default tag ) will win ?

PR #39168, opened 2022-05-27
Don't push to gateway proxies when services are not visible from them

### LINKED ISSUE #39110: Prevent the pilot from pushing to gateway proxies when private services are created

We would like to prevent the pilot from pushing to gateway proxies when services are created with annotation `networking.istio.io/exportTo: ~`.

Our cluster a huge churn rate of k8s services, and that services don’t need istio. We used `Sidecar` CRD and `networking.istio.io/exportTo: ~` annotation to not affect any other services using istio. However, the above options do not prevent pushes to gateway proxies. As a result, we experienced high CPU usage on gateway proxies due to a lot of pushes (even if there is no need to push)

I tested two methods to solve this issue. 
Both methods seem to work fine, but I’m not sure if there is anything that I’ve not considered.

1. Filter service with `networking.istio.io/exportTo: ~` to not be discovered by pilot (https://github.com/henrysecond1/istio/commit/c158cd161a1191b1c59892d7c7123c0611420e49)

2) Check dependency of gateway proxies (https://github.com/henrysecond1/istio/commit/81ab458c194a5764bc20d38d9c3091c18057f3ef)

Any help or advice would be greatly appreciated.

- `Sidecar` CRD - only works for sidecar proxies, not gateway proxies
- `discoverySelector` - we can’t entirely exclude namespaces from discovery, so it was not our option. If there is a service-level discovery selector, it would be great.
- `PILOT_FILTER_GATEWAY_CLUSTER_CONFIG` - AFAIK,  this option can’t prevent push operations. Only the cluster size of gateway proxies decreases.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

henrysecond1:

Fixes #39110

### DIFF
--- a/pilot/pkg/xds/proxy_dependencies.go
+++ b/pilot/pkg/xds/proxy_dependencies.go
@@ -17,6 +17,7 @@ package xds
 import (
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pkg/config"
+	"istio.io/istio/pkg/config/host"
 	"istio.io/istio/pkg/config/schema/gvk"
 )
 
@@ -48,7 +49,7 @@ func ConfigAffectsProxy(req *model.PushRequest, proxy *model.Proxy) bool {
 			}
 		}
 
-		if affected && checkProxyDependencies(proxy, config) {
+		if affected && checkProxyDependencies(proxy, config, req.Push) {
 			return true
 		}
 	}
@@ -56,7 +57,7 @@ func ConfigAffectsProxy(req *model.PushRequest, proxy *model.Proxy) bool {
 	return false
 }
 
-func checkProxyDependencies(proxy *model.Proxy, config model.ConfigKey) bool {
+func checkProxyDependencies(proxy *model.Proxy, config model.ConfigKey, push *model.PushContext) bool {
 	// Detailed config dependencies check.
 	switch proxy.Type {
 	case model.SidecarProxy:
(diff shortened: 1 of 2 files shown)


GregHanson:
@henrysecond1 could you add a release not to `releasenotes/notes` directory?  Steps [here](https://github.com/istio/istio/tree/master/releasenotes)

pilot/pkg/xds/proxy_dependencies.go
```diff
@@ -65,6 +66,17 @@ func checkProxyDependencies(proxy *model.Proxy, config model.ConfigKey) bool {
 		} else if proxy.PrevSidecarScope != nil && proxy.PrevSidecarScope.DependsOnConfig(config) {
 			return true
 		}
+	case model.Router:
+		if config.Kind == gvk.ServiceEntry {
+			// If config is ServiceEntry, name of the config is service's FQDN
+			svc, exist := push.ServiceIndex.HostnameAndNamespace[host.Name(config.Name)][config.Namespace]
+			if exist {
+				if !push.IsServiceVisible(svc, proxy.Metadata.Namespace) {
```
howardjohn, line 74: A bit worried linking `IsServiceVisible` to this when its not actually the same function that is used to determine Services available to a gtw. It very likely functions correctly, but I am not super comfortable with these two decouple functions trying to compute the same thing with different logic.

Let me think on it... @hzxuzhonghu cc if you have ideas

hzxuzhonghu: The check here is a superset of `GatewayServices`,  it's safe I think, maybe not that graceful.

@howardjohn As for gateway, we have generated the SidecarScope for it, I think we can use the same logic as SidecarProxy here, though it is still a superset of `GatewayServices`, but not that much

henrysecond1:
@GregHanson Thank you for letting me know. I added it

GregHanson (approved):

Issue #39063, opened 2022-05-20
Various WASM test flakes


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/unit-tests_istio_postsubmit/1527648505185701888

https://prow.istio.io/view/gs/istio-prow/logs/integ-telemetry-mc_istio_postsubmit/1527635431762432000

https://prow.istio.io/view/gs/istio-prow/logs/integ-telemetry-mc_istio_postsubmit/1527648505345085440

Seem to be from recent WASM PRs

ingwonsong:
First issue should be cleared by #39071

ingwonsong:
The flakes in Second/Third are gone by https://github.com/istio/istio/issues/39063

ingwonsong:
After merging the integration test back, there is no flakiness yet.

PR #39172, opened 2022-05-27
[BugFix] WasmPlugin: Call forwardToEnvoy in the same thread

ingwonsong:
This PR prevent using `forwardToEnvoy` function in multiple threads.

Potentially, calling from multiple threads may cause blocking the loop in `handleUpstreamResponse`.

In details, as commented in https://github.com/grpc/grpc-go/blob/e41f868588c99493b35169e258a11b2ce128e139/stream.go#L1416-L1418, we should not call `SendMsg`, which is called by `forwardToEnvoy` at last, in multiple threads.

This fixes the second/third issues in https://github.com/istio/istio/issues/39063 .

### DIFF
--- a/pkg/istio-agent/xds_proxy.go
+++ b/pkg/istio-agent/xds_proxy.go
@@ -533,7 +533,7 @@ func (p *XdsProxy) handleUpstreamResponse(con *ProxyConnection) {
 						// to prevent concurrent access to forwardToEnvoy
 						select {
 						case forwardEnvoyCh <- resp:
-							proxyLog.Infof("wasm send resp to forwardEnvoCh: %v", resp.TypeUrl)
+							proxyLog.Debugf("wasm send response: %v", resp.TypeUrl)
 						case <-con.stopChan:
 						}
 					})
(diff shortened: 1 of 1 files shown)


pkg/istio-agent/xds_proxy.go
```diff
@@ -527,7 +528,15 @@ func (p *XdsProxy) handleUpstreamResponse(con *ProxyConnection) {
 			case v3.ExtensionConfigurationType:
 				if features.WasmRemoteLoadConversion {
 					// If Wasm remote load conversion feature is enabled, rewrite and send.
-					go p.rewriteAndForward(con, resp)
+					go p.rewriteAndForward(con, resp, func(resp *discovery.DiscoveryResponse) {
+						// Forward the response using the thread of `handleUpstreamResponse`
+						// to prevent concurrent access to forwardToEnvoy
+						select {
+						case forwardEnvoyCh <- resp:
+							proxyLog.Infof("wasm send resp to forwardEnvoCh: %v", resp.TypeUrl)
```
howardjohn, line 536:
```suggestion
							proxyLog.Debugf("wasm send response: %v", resp.TypeUrl)
```

howardjohn (approved):
LGTM. Even better if we have a test covering it :-)

PR #39182, opened 2022-05-28
reduce loop count in selectVirtualServices

fatedier:
Reduce pilot CPU usage. Based on the assumption that the number of wildcarded service hosts usually much less than others.

In my environment: 200+ services and 100+ virtualservices, it will reduce CPU usage by 20% when `pushXDS`.

Before:
image

After:
image

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/httproute.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute.go
@@ -220,8 +220,16 @@ func (configgen *ConfigGeneratorImpl) buildSidecarOutboundHTTPRouteConfig(
 // selectVirtualServices selects the virtual services by matching given services' host names.
 func selectVirtualServices(virtualServices []config.Config, servicesByName map[host.Name]*model.Service) []config.Config {
 	out := make([]config.Config, 0)
-	for _, c := range virtualServices {
-		rule := c.Spec.(*networking.VirtualService)
+	// Pre-compute wildcarded service hosts to reduce loop count.
+	wildCardedSvcHosts := []host.Name{}
+	for svcHost := range servicesByName {
+		if svcHost.IsWildCarded() {
+			wildCardedSvcHosts = append(wildCardedSvcHosts, svcHost)
+		}
+	}
+
+	for i := range virtualServices {
+		rule := virtualServices[i].Spec.(*networking.VirtualService)
 		var match bool
 
 		// Selection algorithm:
@@ -237,10 +245,19 @@ func selectVirtualServices(virtualServices []config.Config, servicesByName map[h
 				break
 			}
 
-			for svcHost := range servicesByName {
-				if host.Name(h).Matches(svcHost) {
-					match = true
-					break
+			if host.Name(h).IsWildCarded() {
+				for svcHost := range servicesByName {
+					if host.Name(h).Matches(svcHost) {
+						match = true
+						break
+					}
+				}
+			} else {
+				for _, svcHost := range wildCardedSvcHosts {
+					if host.Name(h).Matches(svcHost) {
+						match = true
+						break
+					}
 				}
 			}
 
@@ -250,7 +267,7 @@ func selectVirtualServices(virtualServices []config.Config, servicesByName map[h
 		}
 
 		if match {
-			out = append(out, c)
+			out = append(out, virtualServices[i])
 		}
 	}
 


pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -237,10 +245,19 @@ func selectVirtualServices(virtualServices []config.Config, servicesByName map[h
...
+					if host.Name(h).Matches(svcHost) {
+						match = true
+						break
+					}
+				}
+			} else {
+				for _, svcHost := range wildCardedSvcHosts {
+					if host.Name(h).Matches(svcHost) {
+						match = true
+						break
```
hzxuzhonghu, line 259: Not understand why vs wildecar host match all service host, while fqdn matches only wildcard svc host?

Don't you miss wildcard match wildcard?

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -220,8 +220,16 @@ func (configgen *ConfigGeneratorImpl) buildSidecarOutboundHTTPRouteConfig(
 // selectVirtualServices selects the virtual services by matching given services' host names.
 func selectVirtualServices(virtualServices []config.Config, servicesByName map[host.Name]*model.Service) []config.Config {
 	out := make([]config.Config, 0)
-	for _, c := range virtualServices {
-		rule := c.Spec.(*networking.VirtualService)
+	// Pre-compute wildcarded service hosts to reduce loop count.
+	wildCardedSvcHosts := []host.Name{}
+	for svcHost := range servicesByName {
+		if svcHost.IsWildCarded() {
+			wildCardedSvcHosts = append(wildCardedSvcHosts, svcHost)
```
hzxuzhonghu, line 227: I think this applies to very small case, for most k8s services, they are not wildcarded host name.

pilot/pkg/networking/core/v1alpha3/httproute.go
fatedier: Cases:
1. `wildcard vs host` and `wildcard svc host`
2. `wildcard vs host` and `fqdn svc host`
3. `fqdn vs host` and `wildcard svc host`
4. `fqdn vs host` and `fqdn svc host`

I think `vs wildecar host match all service host` covers case 1 and case 2.

`fqdn matches only wildcard svc host` covers case 3.

And map search covers case 4

```go
if servicesByName[host.Name(h)] != nil {
    match = true
    break
}
```

The aim of this PR is to remove case 4 in the loop because map search already handle it and it's the most usage case.

pilot/pkg/networking/core/v1alpha3/httproute.go
fatedier: Yes. I want to split it from fqdn hosts.

In my environment, loop between fqdn vs hosts and fqdn svc hosts will consume too much cpu.

fatedier: On the contrary, the optimization effect of this change will be very obvious when there are few wildcard hosts.

For example

```
vs hosts:
a.com, nomatch1.com, nomatch2.com, nomatch3.com

svc hosts:
a.com, b.com, c.com, d.com
```

Loop count before this change:
```
a.com 1
nomatch1.com 4
nomatch2.com 4
nomatch3.com 4
```

Each nomatch host will iterate over all svc hosts(actually only wildcard hosts are needed because fqdn hosts are excluded by map searching).

Loop count after this change:
```
a.com 1
nomatch1.com 1
nomatch2.com 1
nomatch3.com 1
```

hzxuzhonghu: I am also confused, not understand what you do

fatedier: @hzxuzhonghu This case:

```
vs hosts:
a.com, nomatch1.example.com, nomatch2.com, nomatch3.com

svc hosts:
a.com, b.com, c.com, d.com, *.example.com
```

Focus on the `nomatch1.com` loop

```go
for _, h := range rule.Hosts {
	// h: nomatch1.example.com
	// not in svc hosts map [a.com, b.com, c.com, d.com, *.example.com]
	if servicesByName[host.Name(h)] != nil {
		match = true
		break
	}

        // Continue iterating over all the svc hosts [a.com, b.com, c.com, d.com, *.example.com]
        // But [a.com, b.com, c.com, d.com] are excluded by map searching above.
        // Only need to iterate over [*.example.com] and will match.
	for svcHost := range servicesByName {
		if host.Name(h).Matches(svcHost) {
			match = true
			break
		}
	}

	if match {
		break
	}
}
```

I'm not sure if I miss something.

hzxuzhonghu: kind of understand, please add some comments to the code

hzxuzhonghu: this can remove {both not wildcard host - host} match  again

howardjohn: Oh its for optimizing the "no match" case. Got it. Thanks

ramaraochavali: Yeah. We already do similar optimization else where https://github.com/istio/istio/blob/36f988f68a40fb9aecbf84e0004805770fa9afaa/pilot/pkg/networking/core/v1alpha3/route/route.go#L171 also

fatedier: Comments added.

hzxuzhonghu (approved):

pilot/pkg/networking/core/v1alpha3/httproute.go
fatedier: @hzxuzhonghu My mistake. I don't check in the file changes.

Create a new PR https://github.com/istio/istio/pull/39294

PR #39198, opened 2022-05-31
delta xds: build resources based on cache key dependencies

adiprerepa:
this solution is a lot more elegant than a per-type and per-resource generation, which would probably be more buggy and brute-force.

this PR is a continuation of a PR by @howardjohn: https://github.com/istio/istio/pull/37926. Detailed context exists in the description of that PR.

cc @howardjohn @hzxuzhonghu @ramaraochavali


pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -217,7 +173,30 @@ func (configgen *ConfigGeneratorImpl) buildOutboundClusters(cb *ClusterBuilder,
...
+			// the same applies for all subsets
+			for _, ss := range subsetKeys {
+				skipSubset := skipForDelta(&ss, updated) && skipForDelta(oldCacheKeys[ss.clusterName], updated)
+				deleted.Delete(ss.clusterName)
+				// We can only skip if we skip ALL subsets. This could probably be optimized in the future
+				canSkip = canSkip && skipSubset
+			}
+			// replace old cache key with our generated key
+			oldCacheKeys[clusterKey.clusterName] = clusterKey
+			for _, ss := range subsetKeys {
```
GregHanson, line 192: I believe you can make the linter happy with the modification to the for loop as follows:
```
for i, ss := range subsetKeys {
     oldCacheKeys[ss.clusterName] = &subsetKeys[i]
 }
```

pilot/pkg/xds/delta_utils.go
```diff
@@ -112,21 +116,32 @@ func (s *DiscoveryServer) compareDiff(
...
-			log.Errorf("%s: TEST for node:%s unexpected deletions: %v %v", v3.GetShortType(w.TypeUrl), con.proxy.ID, details, extraDeletes)
+			report("%s: TEST for node:%s unexpected deletions: %v %v", v3.GetShortType(w.TypeUrl), con.proxy.ID, details, extraDeletes)
 		}
 		if len(missedDeletes) > 0 {
-			log.Errorf("%s: TEST for node:%s missed deletions: %v %v", v3.GetShortType(w.TypeUrl), con.proxy.ID, details, missedDeletes)
+			report("%s: TEST for node:%s missed deletions: %v %v", v3.GetShortType(w.TypeUrl), con.proxy.ID, details, missedDeletes)
 		}
 		if len(missedChanges) > 0 {
-			log.Errorf("%s: TEST for node:%s missed changes: %v %v", v3.GetShortType(w.TypeUrl), con.proxy.ID, details, missedChanges)
+			report("%s: TEST for node:%s missed changes: %v %v", v3.GetShortType(w.TypeUrl), con.proxy.ID, details, missedChanges)
```
howardjohn, line 133: report being `fatal` makes the check below not actually work

howardjohn: Also I found that out by it happening in my cluster... so probably means there is a bug. im looking into it a bit.

pilot/pkg/xds/delta.go
```diff
@@ -350,32 +351,30 @@ func (s *DiscoveryServer) shouldRespondDelta(con *Connection, request *discovery
...
-	con.proxy.WatchedResources[request.TypeUrl].ResourceNames = deltaResources
-	con.proxy.Unlock()
 
-	oldAck := listEqualUnordered(previousResources, deltaResources)
+	if !isWildcardTypeURL(request.TypeUrl) {
+		deltaResources := deltaWatchedResources(previousResources, request)
+		con.proxy.WatchedResources[request.TypeUrl].ResourceNames = deltaResources
+	}
+	if isWildcardTypeURL(request.TypeUrl) && (len(request.ResourceNamesSubscribe) > 0 || len(request.ResourceNamesUnsubscribe) > 0) {
+		panic(fmt.Sprintf("invalid assumption on %+v", request))
```
howardjohn, line 362: should either guard with the `UNSAFE_PILOT_ENABLE_RUNTIME_ASSERTIONS` or remove. We don't want to panic in prod code.

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -217,7 +173,30 @@ func (configgen *ConfigGeneratorImpl) buildOutboundClusters(cb *ClusterBuilder,
...
+			// clusterKey can't be nil as per buildClusterKey
+			subsetKeys := cb.getSubsetKeys(*clusterKey)
+			// it exists, we shouldn't mark as deleted
+			deleted.Delete(clusterKey.clusterName)
+			// we can only skip generation of the clusters associated with this service iff none of the updated
+			// resources depend on that cluster
+			canSkip := skipForDelta(clusterKey, updated) && skipForDelta(oldCacheKeys[clusterKey.clusterName], updated)
+			// the same applies for all subsets
+			for _, ss := range subsetKeys {
+				skipSubset := skipForDelta(&ss, updated) && skipForDelta(oldCacheKeys[ss.clusterName], updated)
```
howardjohn, line 185: So the thing that is broken (and fails on `TestTraffic/autopassthrough`) is that we now store a cache key. We then look at the cache key and see if it dependsOn the changed types. However, dependsOn is not the only way cache keys are invalidated - Clear/ClearAll can be called on the cache. When this happens, we would still have all of these cache keys in the `oldCacheKeys` map, and we would see the update from some type and get `skipForDelta`. But that is wrong -- we cannot skip it because the cache key is invalidated (we just didn't notice).

adiprerepa: added invalidation logic

hzxuzhonghu:
@adiprerepa [adiprerep](https://github.com/adiprerepa) Can you make the CI happy?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -185,30 +145,46 @@ func (configgen *ConfigGeneratorImpl) buildClusters(proxy *model.Proxy, req *mod
...
+	}
+	// nonzero length -> delete some
+	if len(configgen.InvalidatedResources) != 0 {
+		// remove from oldCacheKeys
+		for _, k := range configgen.InvalidatedResources {
+			delete(oldCacheKeys, k)
+		}
+	}
+	// set to length 0 for next iteration (not nil)
+	configgen.InvalidatedResources = []string{}
```
howardjohn, line 187: This is a global resource, but we are clearing it on a single call for a single proxy.

I think the write has the same issue

pilot/pkg/xds/delta.go
adiprerepa: done

pilot/pkg/model/context.go
```diff
@@ -341,6 +341,10 @@ type WatchedResource struct {
 	// For Delta Xds, all resources of the TypeUrl that a client has subscribed to.
 	ResourceNames []string
 
+	// CacheKeys maps resource names to actual cache entries, which contain the dependencies
+	// of that resource. We use this for delta.
+	CacheKeys map[string]XdsCacheEntry
```
hzxuzhonghu, line 346: I think this is very mem costing, can you have a bench mark. If it is the same scalability as @ramaraochavali's recent scenario. I can not imagine the mem usage.

adiprerepa: what was @ramaraochavali's recent scenario?

and what would we be benchmarking here? as number of clusters increases, the size of the map will increase right? do we benchmark for the size of `XdsCacheEntry`?

ramaraochavali:
@adiprerepa https://github.com/istio/istio/issues/39368 is the issue that @howardjohn is talking about

hzxuzhonghu:
I think it does not make sense from memory view.  We should rethink about https://docs.google.com/document/d/10l6wsJP2swM_IWGb8mrKA-Vargd5xToeELVmLdfbBJY/edit#heading=h.xw1gqgyqs5b

PR #39202, opened 2022-05-31
remove leader election for webhook patching

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/leaderelection/leaderelection.go
+++ b/pilot/pkg/leaderelection/leaderelection.go
@@ -37,7 +37,6 @@ import (
 // Various locks used throughout the code
 const (
 	NamespaceController     = "istio-namespace-controller-election"
-	WebhookPatcher          = "istio-webhook-patcher-election"
 	ServiceExportController = "istio-serviceexport-controller-election"
 	// This holds the legacy name to not conflict with older control plane deployments which are just
 	// doing the ingress syncing.
--- a/pilot/pkg/serviceregistry/kube/controller/multicluster.go
+++ b/pilot/pkg/serviceregistry/kube/controller/multicluster.go
@@ -36,6 +36,7 @@ import (
 	kubelib "istio.io/istio/pkg/kube"
 	"istio.io/istio/pkg/kube/multicluster"
 	"istio.io/istio/pkg/webhooks"
+	"istio.io/istio/pkg/webhooks/validation/controller"
 )
 
 const (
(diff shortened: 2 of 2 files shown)


hzxuzhonghu:
I think with PRIORITIZED_LEADER_ELECTION=true, we should have leader election to prevent back-forth when data inconsistency

ramaraochavali:
Does leader election support with revision? If yes, what is the issue with current implementation?

costinm (commented):
I think we still need leader election - just per revision. A revision may have multiple instances. This probably needs to match the root cert syncer ( namespace controller ) - webhook patcher and ns config use same cert.

However the root problem here - and I think we need to discuss it in a doc/meeting instead of a set of PR comments - is a more deterministic way to identify and use the CA roots of Istiod. 

We have 'mesh wide'/singleton resources - and per-revision resources. I believe the root of trusts
in the mesh - including the root of trusts used by Istiod itself - should be treated as global resources instead of per-revision, and handled by the 'default' revision. 

Alternative, for 'external' istiod or istiod using custom CAs - as this is an advanced mode, we should just add an option to turn off the webhook patcher and remove the RBAC permission, and 
instead rely on the install process/tools. User does need to create the Secret or provision the DNS certs to Istiod, presumably using some scripts/tools - also to handle rotation of the certs, etc.
This would also decrease the risks - webhook patching is one of the most dangerous permissions to have at runtime. The tool could also patch the Citadel secret to add the external root CA to the root CA file, or set up citadel in 'user-supplied intermediary CA' mode.

ramaraochavali:
I am confused now - I thought leader election does not fully support revision based leader election and that is the reason we can not have leader election for webhook patching same was discussed https://github.com/istio/istio/pull/35243 and https://github.com/istio/istio/pull/39137.

It looks like now I am hearing a different feedback ?

costinm:
Sorry, didn't know about this - what problems would revision-based leader
election have ? Pretty sure
we do it at least for the gateway implementation - we had a similar loop
there. It's just a regular leader
election, but with a per-revision key.

hzxuzhonghu:
Revision leader election should be well considered with canary upgrade.  I do believe there is still a gap for graceful upgrade.

ramaraochavali:
> 
Revision leader election should be well considered with canary upgrade. I do believe there is still a gap for graceful upgrade.


So are you suggesting that we remove leader election for webhook patching till we fix that part - essentially move forward with this PR?

@howardjohn WDYT?

howardjohn (approved):
LGTM since this is reverting to previous logic

frankbu:
> LGTM since this is reverting to previous logic

That's true, but this previous logic was really only used before we change the default value of `features.ExternalIstiod` to true. Even though the Multi-Primary MC tests aren't failing, it would seem that now there's nothing stopping the two Primary clusters from patching each other, other than lack of permission?

PR #39212, opened 2022-05-31
Extends the pull policy to http/https as well

ingwonsong:
This PR proposes extending the scope of ImagePullPolicy to deal with not only OCI Image but also the direct fetching of Wasm module via http[s].

This PR was started from the discussion on https://github.com/istio/istio/pull/38475#issuecomment-1111534522, and the changes in API side (not explicit API change, just change of documentation) is https://github.com/istio/api/pull/2336.

### DIFF
--- a/pkg/wasm/cache.go
+++ b/pkg/wasm/cache.go
@@ -175,18 +175,15 @@ func urlAsResourceName(fullURLStr string) string {
 }
 
 func pullIfNotPresent(pullPolicy extensions.PullPolicy, u *url.URL) bool {
-	if u.Scheme == "oci" {
-		switch pullPolicy {
-		case extensions.PullPolicy_Always:
-			return false
-		case extensions.PullPolicy_IfNotPresent:
-			return true
-		default:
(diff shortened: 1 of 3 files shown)


pkg/wasm/cache.go
```diff
@@ -175,18 +175,15 @@ func urlAsResourceName(fullURLStr string) string {
...
-	if u.Scheme == "oci" {
-		switch pullPolicy {
-		case extensions.PullPolicy_Always:
-			return false
-		case extensions.PullPolicy_IfNotPresent:
-			return true
-		default:
-			return !strings.HasSuffix(u.Path, ":latest")
-		}
+	switch pullPolicy {
```
zirain, line 178: rename this function to `ignoreResourceVersion` to match the real meaning, WDYT?

ingwonsong: Hm, it makes sense. Will change it.

ingwonsong: Done

zirain (approved):

GregHanson (approved):

PR #39214, opened 2022-05-31
debug: add EDS to config_dump output

howardjohn:
This allows dumping EDS config, similar to Envoy's config.

As part of this, the code is changed to consistently use findGenerator,
which also makes sure the dumping works for custom generators (gRPC,
etc).

This should have no behavioral changes for standard case, as the EDS is
opt-in

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/debug.go
+++ b/pilot/pkg/xds/debug.go
@@ -662,7 +662,8 @@ func (s *DiscoveryServer) ConfigDump(w http.ResponseWriter, req *http.Request) {
 		s.errorHandler(w, proxyID, con)
 		return
 	}
-	dump, err := s.configDump(con)
+	includeEds := req.URL.Query().Get("include_eds") == "true"
+	dump, err := s.configDump(con, includeEds)
 	if err != nil {
 		handleHTTPError(w, err)
 		return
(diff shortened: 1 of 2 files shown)


pilot/pkg/xds/debug.go
```diff
@@ -672,91 +673,157 @@ func (s *DiscoveryServer) ConfigDump(w http.ResponseWriter, req *http.Request) {
...
+		endpointsAny, err = util.MessageToAnyWithError(&adminapi.EndpointsConfigDump{
+			DynamicEndpointConfigs: endpointConfig,
+		})
+		if err != nil {
+			return nil, err
 		}
 	}
 
 	bootstrapAny := util.MessageToAny(&adminapi.BootstrapConfigDump{})
 	scopedRoutesAny := util.MessageToAny(&adminapi.ScopedRoutesConfigDump{})
```
hzxuzhonghu, line 809: why have these empty structs

howardjohn: The idea is this endpoint behaves exactly like Envoy's config dump, which includes these. Since we never provide these types we just always generate empty

BTW this part of the code is unchanged.

hzxuzhonghu (approved):

Issue #38877, opened 2022-05-12
Improve build+deploy of in cluster e2e test utilities/samples


howardjohn:
Currently, we have a number of utilities we build or use for testing

* samples/extauthz/ext-authz.yaml - we build as part of our standard image building
* pkg/test/framework/components/containerregistry/container_registry_server.yaml - adhoc building in istio/istio
* pkg/test/framework/components/gcemetadata/gce_metadata_server.yaml - adhoc building in istio/istio
* samples/jwt-server - adhoc building in istio/istio
* pkg/test/framework/components/stackdriver/stackdriver.yaml - adhoc building in istio/**proxy**
* Echo app (arguably not in this category) - we build as part of our standard image building
* RLS, OTEL, Prometheus - we use upstream images

This is not ideal. Adhoc builds are hard to test and require super high privilege to update images. Having so many different images could slow down CI as well

Below lists a few options

### Unified toolbox

We can unify the first 5 images into one `istio/test-toolbox` (or similar), and integrate it into the standard image building.

Pros:
* Each new component we need is ~free
* Integrate with standard tooling
* Can change code + test it in 1 PR
Cons:
* Single monolith may make development more complex. Especially 1 time cost to merge

### Move everything to standard build process

We can just add everything to docker.yaml, like we did for ext_authz.

Pros:
* Quick change
* No more priv needed to publish new images
* Can change code + test it in 1 PR
Cons:
* Possibly high cost in CI to build 5 extra images

### Registry mirror

This is like "standard build process", but attempting to avoid the high CI cost.

We very rarely change these image. Rather than building for every commit, we could just keep giving them fixed versions like we do today. In CI, we could set up the registry config to pull from GCR first and fallback to localhost if not found. In postsubmit, we can have a job to build+push the images if the version is updated and they don't yet exist in the remote registry.

To handle CI for when we do actually make a change to the images, we would need some way to detect that the version had changed (can be same logic as postsubmit?); if it has, we need to build and push to the _local_ registry. The CI will try to pull from GCR and fail, then fallback to localhost and succeed. The reason we do the localhost thing is presubmit doesn't have permissions to push to GCR.

Pros:
* Can use the standard tooling, but needs some non-trivial modifications
* No more priv needed to publish new images
* Can change code + test it in 1 PR
* No runtime cost when test apps are not changed
Cons:
* Complex

howardjohn:
cc @ericvn @nmittler @ingwonsong @stewartbutler soliciting opinions

[...]

PR #39225, opened 2022-06-01
WasmPlugin: Add httpimageproxy integration test component

ingwonsong:
This PR proposes to introduce a new component "httpimageproxy" for testing Wasm Remote Loading via HTTP.

httpimageproxy works as a proxy returning a file by extracting it from a docker image pulled from a docker registry upstream.

In details,
- If a user makes a request "get" with a URL path "/gcr.io/awesome-wasm/tag/v1.0/plugin.wasm"
- `httpimageproxy` pulls the image "gcr.io/awesome-wasm:v1.0"
- `/plugin.wasm` is extracted from the image
- `plugin.wasm` is returned to the user in octet-stream.

We have https://github.com/istio/istio/issues/38877 issue, but at this time there is no direction decided, so add this as a separate image/binary with ad-hoc build script.

### DIFF
--- a/pkg/test/env/istio.go
+++ b/pkg/test/env/istio.go
@@ -80,6 +80,9 @@ var (
 
 	// RegistryRedirectorServerInstallFilePath is the registry redirector installation file.
 	RegistryRedirectorServerInstallFilePath = path.Join(IstioSrc, getInstallationFile("registryredirector/registry_redirector_server.yaml"))
+
+	// HTTPImageProxyInstallFilePath is the httpimageproxy installation file.
+	HTTPImageProxyInstallFilePath = path.Join(IstioSrc, getInstallationFile("httpimageproxy/httpimageproxy.yaml"))
 )
 
 var (
--- a/pkg/test/fakes/httpimageproxy/Dockerfile
+++ b/pkg/test/fakes/httpimageproxy/Dockerfile
@@ -1,0 +1,3 @@
+FROM scratch
+COPY ./main /proxy
+ENTRYPOINT ["/proxy"]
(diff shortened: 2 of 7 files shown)


howardjohn:
Not sure what the use of this is for? Why are we pulling OCI images to bytes using a server instead of just directly from the client/test/etc?

ingwonsong:
> Not sure what the use of this is for? Why are we pulling OCI images to bytes using a server instead of just directly from the client/test/etc?

There was not enough explanation. Sorry for that.

As you know, WasmPlugin API can use OCI Image as well as directly downloading via HTTP for getting a remote Wasm binary.
Currently, we have the integration test for OCI Image pulling, but there is no HTTP case.
To test HTTP, we need to have a http server serving the wasm binaries.

To avoid copying the wasm binaries to OCI images on Registry as well as the HTTP server, I designed this http image proxy.

howardjohn:
Do we need to deploy a full server to proxy this though? Seems like the client should just fetch it directly. This can be done directly (`crane blob ubuntu@sha256:4c1d20cdee96111c8acf1858b62655a37ce81ae48648993542b7ac363ac5c0e5 > blob.tar.gz` for example), or even utilizing the fact that GCR is just a redirector for GCS potentially. OCI is already HTTP

howardjohn:
> GCS also seems to keep blobs, which are layers consisting an image.

Aren't the blobs the WASM image we need? WASM is just a single layer?

Or is it also compressed and we don't spport that?

ingwonsong:
understood what you meant.

Yes, Wasm uses single blob and the blob is the binary itself. 

Only thing need to check is the compression as you said. 

Will check it and get back to you.

PR #39235, opened 2022-06-01
tf: make Ingress component more flexible

howardjohn:
Right now it can only be used in istio-system with the `istio=...`
label. This allows arbitrary namespaces and labels to be used. Also
cleans up some dead code.

### DIFF
--- a/pkg/test/framework/components/istio/ingress.go
+++ b/pkg/test/framework/components/istio/ingress.go
@@ -16,12 +16,13 @@ package istio
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"io"
 	"net"
 	"strconv"
 	"time"
+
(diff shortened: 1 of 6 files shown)


pkg/test/framework/components/istio/ingress.go
```diff
@@ -249,17 +242,8 @@ func (c *ingressImpl) schemeFor(opts echo.CallOptions) (scheme.Instance, error)
 	return opts.Port.Scheme()
 }
 
-func (c *ingressImpl) ProxyStats() (map[string]int, error) {
```
nmittler, line 252: Is this not used?

nmittler (approved):

pkg/test/framework/components/istio/ingress.go
howardjohn: yep

PR #39252, opened 2022-06-02
Add priorityClassName to gateway helm chart

robinelfrink:

Add `priorityClassName` to the gateway Helm Chart in `manifests/charts/gateway`.

### DIFF
--- a/manifests/charts/gateway/templates/deployment.yaml
+++ b/manifests/charts/gateway/templates/deployment.yaml
@@ -102,3 +102,6 @@ spec:
       topologySpreadConstraints:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.priorityClassName }}
+      priorityClassName: "{{ .Values.priorityClassName }}"
+      {{- end }}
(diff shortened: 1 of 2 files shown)


howardjohn (changes requested):
Sorry you are caught in the between stages where we have made a decision but not documented it -- but the Environments WG has made a decision to not allow modifications to helm charts for these types of changes and instead prefer last mile customiation: https://docs.google.com/document/d/1wsa06GGiq1LEGwhkiPP0FKIZJqdAiue-VeBonWAzAyk/edit#bookmark=id.4ru6nai2al4s

robinelfrink:
Okay, thanks for the info.

We're using [Flux](https://fluxcd.io/) so we can easily add a `postRenderer` to our `HelmRelease` objects:

```
spec:
  postRenderers:
  - kustomize:
      patchesStrategicMerge:
      - kind: Deployment
        apiVersion: apps/v1
        spec:
          template:
            spec:
              priorityClassName: super-high-priority
```

PR #39259, opened 2022-06-02
added custom operator for eastwest gateway

aryan16:
**Added a way to update the east-west gateway deployment**

### DIFF
--- a/pkg/test/framework/components/istio/eastwest.go
+++ b/pkg/test/framework/components/istio/eastwest.go
@@ -48,7 +48,7 @@ var (
 )
 
 // deployEastWestGateway will create a separate gateway deployment for cross-cluster discovery or cross-network services.
-func (i *operatorComponent) deployEastWestGateway(cluster cluster.Cluster, revision string) error {
+func (i *operatorComponent) deployEastWestGateway(cluster cluster.Cluster, revision string, customSettings string) error {
 	// generate istio operator yaml
 	args := []string{
 		"--cluster", cluster.Name(),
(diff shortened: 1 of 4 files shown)


pkg/test/framework/components/istio/eastwest.go
```diff
@@ -48,7 +48,7 @@ var (
 )
 
 // deployEastWestGateway will create a separate gateway deployment for cross-cluster discovery or cross-network services.
-func (i *operatorComponent) deployEastWestGateway(cluster cluster.Cluster, revision string) error {
+func (i *operatorComponent) deployEastWestGateway(cluster cluster.Cluster, revision string, customSettings string) error {
```
vikaschoudhary16, line 51: I think it would be better if we handle this through existing `genGatewayScript` itself?
Either we can always add the volume mounts for pilot certs through the shell `gen-eastwest-gateway.sh` and if we dont want to mount those volumes always, we can add another flag like `--mountPilotCerts` or something.
We dont need to add new customisation. This is adding complexity unnecesssary

vikaschoudhary16 (changes requested):

pkg/test/framework/components/istio/eastwest.go
aryan16: The advantage to doing to it in this way is that we can handle any kind of update to eastwest gateway settings, we are anyway doing same for ingress and egress.

vikaschoudhary16: If we want to make it more customizable than the just pilot certs, cant we pass arbitrary values to exisitng shell script?

stevenctl: The shell script can't do smart yaml merging which could make things tricky. A human user might use this script to generate the "base" for an easy West gateway (which is really a specialized ingress) and then modify it.

This is the simplest/most flexible thing we can do for automated tests imo.

PR #39260, opened 2022-06-02
Decompress or untar the wasm binary if it is needed when pulling it via HTTP/HTTPS

ingwonsong:
Decompress or/and untar the wasm binary when it is pulled via HTTP/HTTPS.
By this, we can support the compressed stream as well (even with the wrong `content-type`)

In details, when pulling bytes from HTTP/HTTPS server, if the bytes do not have proper magic number for Wasm Binary, 
tries to ungzip or untar based on the magic numbers of each fileformat.

In case of `untar`, this PR assumes that there is only one file in the same sense that OCI Image specification for Wasm binary.

Basically, it does the best effort based on the magic number of bytes.
If it fails to decompress or untar, just uses the original bytes.

Basically, the original idea came from https://github.com/istio/istio/pull/39225#issuecomment-1144207752.

### DIFF
--- a/pkg/wasm/httpfetcher.go
+++ b/pkg/wasm/httpfetcher.go
@@ -15,15 +15,22 @@
 package wasm
 
 import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
 	"context"
 	"crypto/tls"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
 )
+
+var tarMagicNumber = []byte{0x75, 0x73, 0x74, 0x61, 0x72}
+var gzMagicNumber = []byte{0x1f, 0x8b}
 
 // HTTPFetcher fetches remote wasm module with HTTP get.
 type HTTPFetcher struct {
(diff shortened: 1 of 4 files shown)


pkg/wasm/httpfetcher.go
```diff
@@ -15,16 +15,23 @@
...
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
 )
 
+var tarMagicNumber = []byte{0x75, 0x73, 0x74, 0x61, 0x72}
```
vikaschoudhary16, line 32: just for my learning, from where you got these magic numbers?

ingwonsong: You can check them various documents. (e.g., https://en.wikipedia.org/wiki/List_of_file_signatures)

GregHanson (approved):

ingwonsong:
@howardjohn Do you have any thought?

howardjohn (approved):

zirain (approved):

pkg/wasm/httpfetcher.go
```diff
@@ -15,16 +15,25 @@
...
 	"io"
+	"io/ioutil"
 	"net/http"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
 )
 
+var (
+	tarMagicNumber = []byte{0x75, 0x73, 0x74, 0x61, 0x72}
```
zirain, line 33: address where these came from?

zirain:
hold for address comments, feel free to remove it

pkg/wasm/httpfetcher.go
ingwonsong: Done!

Issue #37877, opened 2022-03-11
docker builder --architectures doesn't work


howardjohn:
```
$ ./tools/docker --push --hub=localhost:5000 --targets=pilot --architecures=linux/arm64,linux/amd64 |& hh total= runtime= step= image= arm64 amd64 certs
go: downloading google.golang.org/grpc v1.45.0

real    0m2.691s
user    0m7.107s
sys     0m0.960s
2022-03-11T20:03:09.399413Z     info    Args: {Push:true Save:false BuildxEnabled:true NoClobber:false NoCache:false Targets:[pilot] Variants:[default] Architectures:[linux/arm64 linux/amd64] BaseVersion:master-2022-03-05T19-01-04 ProxyVersion:75393c3d79795e26ed9fc9626992720d13ac195b IstioVersion:1.14-dev Tags:[latest] Hubs:[localhost:5000]}
2022-03-11T20:03:09.399574Z     info    Running command: make build.docker.pilot
GOOS=linux GOARCH=amd64 LDFLAGS='-extldflags -static -s -w' common/scripts/gobuild.sh /home/howardjohn/go/src/istio.io/istio/out/linux_amd64/ -tags="" ./pilot/cmd/pilot-discovery

real    0m13.078s
user    0m57.980s
sys     0m4.504s
./tools/docker-copy.sh pilot/docker/Dockerfile.pilot /home/howardjohn/go/src/istio.io/istio/tools/packaging/common//envoy_bootstrap.json /home/howardjohn/go/src/istio.io/istio/tools/packaging/common//gcp_envoy_bootstrap.json /home/howardjohn/go/src/istio.io/istio/out/linux_amd64/pilot-discovery /home/howardjohn/go/src/istio.io/istio/out/linux_amd64/dockerx_build/build.docker.pilot
2022-03-11T20:03:22.685235Z     info    Running command: docker buildx bake -f /home/howardjohn/go/src/istio.io/istio/out/linux_amd64/dockerx_build/docker-bake.json all
#1 [internal] booting buildkit
#1 sha256:6fa7803fc6b98daf13dce2dbe41ee07f56aae046f280a560e1a98990d3138e25
#1 starting container buildx_buildkit_container-builder0
#1 starting container buildx_buildkit_container-builder0 1.3s done
#1 DONE 1.3s

#2 [internal] load build definition from Dockerfile.pilot
[...]

Issue #26652, opened 2020-08-19
Multi-arch builds on docker.io


scottmalkie:
**Describe the feature request**

With support for linux/arm64 binaries as of 1.6.x, would it be possible to convert the amd64-only images on dockerhub to multi-arch images, so they can be used on mixed-architecture clusters?

Much of the "getting started" and other documentation fails on mixed-architecture or arm64-only clusters since the dockerhub images are amd64-only.

**Describe alternatives you've considered**
Rebuilding and separately hosting multi-arch builds in other repositories.

[ ] Docs
[X] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[X] Test and Release
[X] User Experience
[X] Developer Infrastructure

**Additional context**

It's even possible to use docker's new buildx functionality to easily create and push multi-arch images/manifests.

Thanks!

howardjohn:
Related: https://github.com/istio/istio/issues/21094

morlay:
@scottmalkie 

could you try https://github.com/querycap/istio

Need your feedbacks in your mixed-architecture cluster.

ecliptik:
I got arm64 images building and setup for mutli-arch on Dockerhub (see: https://github.com/istio/istio/issues/21094#issuecomment-702429990), but used a EC2 `a2.xlarge` instance to build Istio on ARM natively. I imagine it's much more work to get Istio builds to play nicely with `docker buildx` and a cross-compile.

Having multi-arch officially baked into Istio images would be great though, since our current plan is to build/host our own Istio images with multi-arch support, but that's unsustainable over time.

morlay:
@howardjohn is any plan to support this? 

I'm glad to merge my work https://github.com/querycap/istio to istio

Here is something may block the process 

### istio envoy binaries

wasms is not care the cpu arch, so could be compiled only once. here only need do for istio/envoy.

* [ ] istio-envoy-build-env for multi-arch

`envoyproxy/envoy-build-ubuntu` upgrade to llvm-10, which not support to build for istio/envoy. 

bazel < 3.5 in istio/envoy. have to need a unofficial bazel to support arm https://github.com/querycap/istio/blob/master/build/istio-envoy-build-env/Dockerfile.istio-envoy-build-env#L33

it could be used like: (amd64 & arm64 host supported both)
```shell
docker run \
   -e=TEST_TMPDIR=/tmp/bazel \
   -v=/tmp/bazel:/tmp/bazel \
[...]

PR #39278, opened 2022-06-03
docker: support multi/cross-architecture builds

howardjohn:
Fixes https://github.com/istio/istio/issues/37877

This introduces support in the docker builder to orchestrate multi-arch
docker builds. Previously, it was required to call the builder twice
with each arch; this doesn't allow aggregating into a single multi-arch
image though.

For docker builder, this is fairly straightforward. We mostly need
changes to allow plan file to be arch-specific, and to call make
multiple times per-arch.

For crane builder, a little more complex since we need to allow passing
multiple plans to it and then deciding whether to make a Index or direct
Image.

For https://github.com/istio/istio/issues/26652#issuecomment-1155682106

### DIFF
--- a/tools/docker-builder/builder/crane.go
+++ b/tools/docker-builder/builder/crane.go
@@ -36,9 +36,19 @@ import (
 	"istio.io/pkg/log"
 )
 
+type BuildSpec struct {
+	// Name is optional, for logging
+	Name  string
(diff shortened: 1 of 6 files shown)


ericvn (approved):
Missed the approve button?

Issue #39255, opened 2022-06-02
TestAuthz_EgressGateway test flake


howardjohn:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/39238/integ-security-multicluster_istio/1532176973567102976

howardjohn:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/39263/integ-security-multicluster_istio/1532507198838542336

this one seems different

howardjohn:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/39277/integ-security-multicluster_istio/1532761517953388544
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/39261/integ-security-multicluster_istio/1532778634010431488
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/36214/integ-security-multicluster_istio/1532751916935680000

howardjohn:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/39277/integ-security-multicluster_istio/1532811593526546432

howardjohn:
fixed

PR #39280, opened 2022-06-03
Skip authz tests while flakes are investigated

howardjohn:
See https://github.com/istio/istio/issues/39255

Currently 25-40% of jobs are failing on this:
https://prow.istio.io/?job=integ-security-multicluster_istio which is
slowing down development

### DIFF
--- a/tests/integration/security/authz/main_test.go
+++ b/tests/integration/security/authz/main_test.go
@@ -37,6 +37,7 @@ var (
 func TestMain(m *testing.M) {
 	framework.
 		NewSuite(m).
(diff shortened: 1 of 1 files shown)


lei-tang (approved):

nmittler:
@howardjohn it seems that the majority of the flakes are related to the egress test. It's concerning to disable all authz tests ... can you carve out a more specific set of tests that are skipped?

howardjohn:
AFAIK its not just the egress test, seems many of them are failing. 

My understanding is these new tests were in _addition_ to the old authz tests? If not, this is definitely concerning to disable all of them...

PR #39283, opened 2022-06-04
add support for redis cluster protocol

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali:
@howardjohn @hzxuzhonghu can you PTAL?

hzxuzhonghu (commented):
I have not used redis for more than 7 years. So not an expert on redis. 

Given previous memory, how is this different from redis sharding previously supported?

ramaraochavali:
> Given previous memory, how is this different from redis sharding previously supported?

This adds support for [redis clusters](https://redis.io/docs/manual/scaling/#:~:text=Redis%20Cluster%20provides%20a%20way,are%20not%20able%20to%20communicate.) compared to standalone redis protocol - improves availability. See https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/clusters/redis/v3/redis_cluster.proto.html?highlight=redis%20cluster

howardjohn (commented):
Is this something we are ready to have first class support for? This seems like it could be done with an EnvoyFilter.

All of our protocol support seems half-baked. It is permanently behind feature flag, has no testing, is incomplete, etc.

ramaraochavali:
Some additional data points, we have been running Redis in large clusters at scale via Mesh for past three years. Also majority of our clusters that run mesh for various customers make use of Redis support in Istio either with in house hosted Redis or with cloud provider managed Redis(AWS Elastic cache for example). So I think we should be OK to add first class support for it. 
We can definitely add additional integration tests .

howardjohn (commented):
I really feel like we are piecemeal adding protocol support and continue to implement a small portion of what is needed. I worry a lot about proliferation of `redis`, `redis-cluster`, etc `redis-foobar` without a unified approach for how we should extend protocols. @nrjpoddar was previously working on this but I don't think its still active.

> Yes. But having support for Redis and not support redis cluster (and asking users to create envoy filter for it) seems odd.

My concern is we already barely support redis. We have a partial implementation, but custom protocol support is slapped onto Istio without cohesive design

PR #39284, opened 2022-06-04
Time: Compound literals should use key-value fields

Chen-Xintong:

Time: Compound literals should use key-value fields


adiprerepa (approved):
thanks!

PR #39288, opened 2022-06-05
unify network handlers

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/network.go
+++ b/pilot/pkg/model/network.go
@@ -80,17 +80,17 @@ func NewNetworkManager(env *Environment, xdsUpdater XDSUpdater) (*NetworkManager
 }
 
 func (mgr *NetworkManager) reloadAndPush() {
-	changed := mgr.reloadNetworkEndpoints()
+	endpointsChanged := mgr.reloadNetworkEndpoints()
 
 	mgr.mu.Lock()
 	oldGateways := make(NetworkGatewaySet)
 	for _, gateway := range mgr.allGateways() {
 		oldGateways.Add(gateway)
 	}
-	changed = changed || !mgr.reload().Equals(oldGateways)
+	gatewaysChanged := !mgr.reload().Equals(oldGateways)
 	mgr.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


ramaraochavali:
@hzxuzhonghu PTAL as discussed in another PR

pilot/pkg/model/network.go
```diff
@@ -79,20 +80,55 @@ func NewNetworkManager(env *Environment, xdsUpdater XDSUpdater) (*NetworkManager
 }
 
 func (mgr *NetworkManager) reloadAndPush() {
+	changed := mgr.reloadNetworkEndpoints()
```
hzxuzhonghu, line 83: I think you should keep `reloadNetworkEndpoints` as another network handler and only register to `NetworksWatcher`

ramaraochavali: SGTM. Done

pilot/pkg/model/network.go
```diff
@@ -71,28 +72,67 @@ func NewNetworkManager(env *Environment, xdsUpdater XDSUpdater) (*NetworkManager
...
+	mgr.mu.Unlock()
 
 	if changed && mgr.xdsUpdater != nil {
 		log.Infof("gateways changed, triggering push")
 		mgr.xdsUpdater.ConfigUpdate(&PushRequest{Full: true, Reason: []TriggerReason{NetworksTrigger}})
 	}
 }
 
+// reloadNetworkEndpoints reloads NetworkEndpoints and triggers a push if they change.
+func (mgr *NetworkManager) reloadNetworkEndpoints() {
```
hzxuzhonghu, line 101: sorry one more comment:  Missing from the original pr, seems this should be protected with EnableHCMInternalNetworks flag

ramaraochavali: Ah..Good catch

hzxuzhonghu (approved):

PR #26059, opened 2020-08-03
add gRPC status in fault abort

ckcd:
Implement of `grpc status` in `fault abort`.

e.g.

```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
...
spec:
  hosts:
  - ratings
  http:
  - fault:
      abort:
        grpcStatus: DEADLINE_EXCEEDED
        percentage:
          value: 100
```

The related api is in [istio/api:virtual_service.proto#L1816](https://github.com/istio/api/blob/master/networking/v1alpha3/virtual_service.proto#L1816) and for now it's `hide_from_docs`.

And we discuss about whether the type of `GrpcStatus` should be `string` or `uint32` in https://github.com/istio/istio/issues/24698. cc @howardjohn 

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #39295, opened 2022-06-06
Support gRPC fault injection (allow specifying gRPC error codes for fault injection)

su225:
Original PR - https://github.com/istio/istio/pull/26059
This PR resurrects it and only adds integration test to complete it.

**Brief description**: This PR adds support to injecting gRPC faults with given status codes. It builds on the original PR which seemed to be abandoned. The actual changes like validation and config-generation were already there in the original PR (full credits to the person who raised it) and was working fine. The only thing I added is the integration test as suggested by @howardjohn in https://github.com/istio/istio/pull/26059#issuecomment-1142690342 .

Once this goes through we have to remove `$hide_from_docs` for the `grpcStatus` field and update the documentation to indicate that users have to specify upper-cased string like `UNAVAILABLE` as mentioned in https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

Open question
- [ ] Should we allow users to inject faults with `OK`? This is because it is technically not a fault. We can allow it to be consistent with HTTP, where specifying 200 OK is allowed.


howardjohn (approved):
This is great!

tests/integration/pilot/common/routing.go
```diff
@@ -657,6 +658,36 @@ spec:
 		workloadAgnostic: true,
 	})
 
+	t.RunTraffic(TrafficTestCase{
+		name: "fault abort gRPC",
```
howardjohn, line 662: we should try setting sourcematchers and targetMatchers to `match.NotNaked, match.NotHeadless`.

Currently we have proxyless grpc opted out; this seems like a REALLY useful feature for proxyless grpc.

I don't know if itworks though, so not certain that will pass

su225: Did it. 
```go
		sourceMatchers:   []match.Matcher{match.NotNaked, match.NotHeadless},
		targetMatchers:   []match.Matcher{match.NotNaked, match.NotHeadless},
```
And it is 🟢
image

Let's enable it?

PR #33111, opened 2021-05-26
Removed legacy chrion webhook controller

hzxuzhonghu:
Cleanup, istio is not using the secret chrion created any more.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[x] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #39298, opened 2022-06-06
remove chiron controller

hzxuzhonghu:

Chiron webhook controller will be started when `MeshConfig.Certificates` is provided, but `MeshConfig.Certificates` is never used later.  It seems a legacy output before istiod.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [x] Security
- [ ] Test and Release
- [ ] User Experience


hzxuzhonghu:
cc @costinm @lei-tang @howardjohn

lei-tang (changes requested):
This PR is a duplicate of https://github.com/istio/istio/pull/33111, in which the reviewer's comment is "We need to continue supporting https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#Certificate API".

hzxuzhonghu:
We need to revisit, every time i come across these codes, we need to rethink how does it work. Very confusing to user, if user wants a tool just generating certificates, there are couples.

costinm (commented):
I can't think of any use case left for Istiod managing DNS certs for other applications as Secrets - and holding associated RBAC. 

LGTM - but given that we don't know if a user saw this and used it, we need to be extra careful. Can you add something to istioctl to warn the user if meshconfig.Certificates is set ? And relnotes. I doubt it is used -  CertManager provides the same feature with better UX.

costinm:
Original use case was pre-istiod we had to create secrets for Galley and pilot.

If I read the PR correctly, we still have code to init istiod own certs using k8s signer, only the secret creation feature is removed.

hzxuzhonghu:
>LGTM - but given that we don't know if a user saw this and used it, we need to be extra careful. Can you add something to istioctl to warn the user if meshconfig.Certificates is set ? And relnotes. I doubt it is used - CertManager provides the same feature with better UX.

Added release note and also analyzer.

>If I read the PR correctly, we still have code to init istiod own certs using k8s signer, only the secret creation feature is removed.

Right, this  utility function is relied on so I left it.

howardjohn:
IMO we should remove this long term but should be through a deprecation
phase. I already hid the docs to quietly discourage it, but if we are going
to remove it we will need to be a bit louder about it. There *are* users,
since we have tried to remove it a few times and people have complained.

The feature is broken on Kubernetes 1.22+ so it makes a much more
compelling case to deprecate now.

hzxuzhonghu:
If we do nothing now like a year ago, the problem will exist until istio die.  Legacy stuff should be cleaned ASAP, we can refer the k8s API deprecation, I remember the service topology (core/v1) hint was still deprecated .  

Come back to certificates API, I do not think deprecating it will break any user. The certificates if generated by this way, and are used during bootstrap, however bootstrap failing does not break the mesh, will it?

howardjohn:
> Come back to certificates API, I do not think deprecating it will break any user. The certificates if generated by this way, and are used during bootstrap, however bootstrap failing does not break the mesh, will it?

People are using it for ingress gateway certs (like replacing cert-manager). I agree they *shouldn't* but they are and we need to give  a deprecation period (As Kubernetes did)

hzxuzhonghu:
I remember we supported signing gateway cert like workload cert?

shankgan:
Agree that we need to cleanup legacy code.

additional point: while the use of the chiron controller to sign "meshConfig.Certificates" is no longer supported, we do have support to use the kubernetes csr api to sign Istiod DNS certificate. In fact, we even have a test here for this - refer https://github.com/istio/istio/blob/master/tests/integration/security/external_ca/main_test.go#L146 for external K8s signers. Some of this controller logic will need to be re-used to refresh the Istiod DNS certificate.

hzxuzhonghu:
Using k8s to sign the Istiod DNS certificate is still there.

hzxuzhonghu:
>Using k8s signer for istiod own cert is also something we should either remove or make it default on the long term.


Why do you say that, multi choices  looks good to me.

PR #39299, opened 2022-06-06
Make ENABLE_AUTO_SNI affects Istio Mutal TLS outbound cluster

l8huang:
Before this change, ENABLE_AUTO_SNI only affects simple TLS and mutual TLS
outbound cluster; it should be applied to Istio mutual TLS as well, to avoid
setting SNI for routing via egress gateway.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -729,7 +729,7 @@ func (cb *ClusterBuilder) applyTrafficPolicy(opts buildClusterOpts) {
 		applyLoadBalancer(opts.mutable.cluster, loadBalancer, opts.port, cb.locality, cb.proxyLabels, opts.mesh)
 		if opts.clusterMode != SniDnatClusterMode {
 			autoMTLSEnabled := opts.mesh.GetEnableAutoMtls().Value
-			tls, mtlsCtxType := cb.buildAutoMtlsSettings(tls, opts.serviceAccounts, opts.istioMtlsSni,
+			tls, mtlsCtxType := cb.buildAutoMtlsSettings(tls, opts.serviceAccounts,
 				autoMTLSEnabled, opts.meshExternal, opts.serviceMTLSMode)
 			cb.applyUpstreamTLSSettings(&opts, tls, mtlsCtxType)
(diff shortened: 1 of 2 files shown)


howardjohn (changes requested):
can you cover the use cases why we want this for istio_mutual?

Istio mutual has special SNI expectations that do NOT match the Host name

l8huang:
Our use case is E2E TLS through egress gateway:

app -- HTTP --> sidecar -- ISTIO MTLS --> egress gateway -- HTTPS --> mesh external service

We created DestinationRule like step 3 in [Egress Gateways with TLS Origination](https://istio.io/latest/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/#perform-tls-origination-with-an-egress-gateway), which has `sni` specified. Unfortunately we run into some scalability issues, so we want to avoid creating subset for every external service(there could be more than thousands external services). To do that, auto SNI need to be enabled in sidecar's outbound cluster.

howardjohn:
I see. IMO its a bit dangerous since its really only useful in select scenarios but its used everywhere. This would break multinetwork.

Use of Istio_mutual without the "istio sni" is really a bit a of hack due to use having some inflexible protocol decisions. Some of the future work on H2 connect as the transport protocol will fix this.

hzxuzhonghu (changes requested):
Auto SNI should not apply to Istio_MUTUAL

l8huang:
I see using feature ENABLE_AUTO_SNI on Istio mutual TLS is improper now. Does it make sense to add a option to ProxyConfig to turn on auto SNI for Istio mutual TLS outbound cluster(intentionally giving up Istio MTLS auto passthrough)?

In our use case, we only need Istio mutual TLS from sidecar to egress gateway, egress gateway will originate a new simple TLS to external service; we don't use multinetwork(AUTO_PASSTHROUGH) at all.

l8huang:
Close because this conflicts with AUTO PASSTHROUGH.
A better solution for this use case could be using EnvoyFilter to enabled auto sni for sidecar's outbound cluster which points to egress gateway.

Issue #39209, opened 2022-05-31
Listeners stuck warming in Delta XDS CI


howardjohn:
https://storage.googleapis.com/istio-prow/logs/integ-security-multicluster_istio_postsubmit/1531624978833739776/artifacts/security-authz-f93e4215f5684783/_suite_context/echo1-3-78501-state-2018355442/primary/delta-v1-64cdc6cdd-wdx28_proxy-config.json

https://prow.istio.io/view/gs/istio-prow/logs/integ-security-multicluster_istio_postsubmit/1531624978833739776

Pod never becomes ready due to stuck _listeners_. Seems like https://github.com/istio/istio/issues/38709 but listeners instead of clusters.

hzxuzhonghu:
Seems all the failures are from multicluster cases?

howardjohn:
it's only on the new authz test which deploys more workloads than any existing suite. So I'm guessing that with multicluster just makes it larger scale and triggers it, not multicluster specific. Just a guess though

howardjohn:
I believe its due to https://github.com/grpc/grpc-go/issues/5393 but not confirmed.

howardjohn:
good idea let me try it tomorrow. Easy to reproduce by running the integration test setup but with 8 namespaces instead of one. Ran on my desktop in kind

howardjohn:
One example that could trigger it (there are more):

1. xdsproxy Sends() a few messages which fill up Istiod's reqCh
2. xdsproxy upstream.Send() is slow for various reasons
3. Envoy sends 10 XDS requests. filling up the reqCh (blocked by Send())
4. xdsproxy stops calling upstream.Recv since its blocked on pushing to reqCh.
5. Because of (4), Istiod Send() blocks
6. Because of (5) and (1), Istiod no longer calls Recv()
7. Because of (6), (2) never completes

Result: complete deadlock.

This is only one possibly trigger, we can get into the state in plenty of different ways

howardjohn:
![PXL_20220606_202534211](https://user-images.githubusercontent.com/623453/172242832-cbe8ed32-2cad-4164-bf3d-947265e951fe.jpg)

hzxuzhonghu:
>Envoy sends 10 XDS requests. filling up the reqCh (blocked by Send())


How could this happen?  In theory, Envoy sends at most CDS/EDS?LDS/RDS/ECDS(not simultaneously), and xds_proxy can send health check, nds, pcds at beginning

[...]

PR #39309, opened 2022-06-06
ads: prefer requests over pushes

howardjohn:
This does not fully fix https://github.com/istio/istio/issues/39209, but was inspired by analyzing the problem there.

### DIFF
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -305,6 +305,27 @@ func (s *DiscoveryServer) Stream(stream DiscoveryStream) error {
 	<-con.initialized
 
 	for {
+		// Go select{} statements are not ordered; the same channel can be chosen many times.
+		// For requests, these are higher priority (client may be blocked on startup until these are done)
+		// and often very cheap to handle (simple ACK), so we check it first.
+		select {
+		case req, ok := <-con.reqChan:
+			if ok {
+				if err := s.processRequest(req, con); err != nil {
+					return err
+				}
+			} else {
+				// Remote side closed connection or error processing the request.
+				return <-con.errorChan
+			}
+		case <-con.stop:
+			return nil
+		default:
+		}
+		// If there wasn't already a request, poll for requests and pushes. Note: if we have a huge
+		// amount of incoming requests, we may still send some pushes, as we do not `continue` above;
+		// however, requests will be handled ~2x as much as pushes. This ensures a wave of requests
+		// cannot completely starve pushes. However, this scenario is unlikely.
 		select {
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (commented):
LGTM, i suggest improving the recvChannel size to 10 same as xds_proxy

pilot/pkg/xds/ads.go
```diff
@@ -305,6 +305,27 @@ func (s *DiscoveryServer) Stream(stream DiscoveryStream) error {
...
+		select {
+		case req, ok := <-con.reqChan:
+			if ok {
+				if err := s.processRequest(req, con); err != nil {
+					return err
+				}
+			} else {
+				// Remote side closed connection or error processing the request.
+				return <-con.errorChan
+			}
```
hzxuzhonghu, line 320: need to break in this case, and do another recvChan read

howardjohn: for which part? the !ok?

hzxuzhonghu: if ok{
goto L311 again
}

hzxuzhonghu: `goto L311 again` can be put at end of this case (case req, ok := <-con.reqChan:)

howardjohn: This is not done intentionally, see comment below:
```
		// If there wasn't already a request, poll for requests and pushes. Note: if we have a huge
		// amount of incoming requests, we may still send some pushes, as we do not `continue` above;
		// however, requests will be handled ~2x as much as pushes. This ensures a wave of requests
		// cannot completely starve pushes. However, this scenario is unlikely.
```

hzxuzhonghu: I understand your meaning, this is a tradeoff, hard to say which is better.

Say if request is not handled, we donot have a chance to update the watched resources, then the push handling could be meaningless

howardjohn: Yeah definitely it is a tradeoff. On the other hand, we may have requests for new resource (applying some new config, doesn't need to be super fast) blocking an EDS push (critical).

I am not sure there is a perfect solution. I went back and forth on this but went with the current approach because its closer to the old way (fully random) so seemed less risky

hzxuzhonghu: The overall experiment is excellent, great work.

I kind of remember priority channel somewhere, could not find it.

howardjohn:
I confirmed this alone does NOT fix the deadlock. I still think its useful though.

hzxuzhonghu (approved):

PR #39311, opened 2022-06-07
added a custom gateway operator

aryan16:
**Added a custom operator for ingress and egress gateway installation to fix the external control plane tests **


howardjohn (approved):

tests/integration/security/filebased_tls_origination/main_test.go
```diff
@@ -35,16 +36,24 @@ func TestMain(m *testing.M) {
...
 		return
 	}
-	cfg.ControlPlaneValues = `
+
+	var isExternalControlPlane bool
+	for _, cluster := range ctx.AllClusters() {
+		if cluster.IsExternalControlPlane() {
+			isExternalControlPlane = true
+		}
+	}
```
frankbu, line 53: Maybe it would be good to add an `IsExternalControlPlane()` method to `resource.Environment` and then this test, and the one from your previous PR, could both just call it.

```suggestion
	isExternalControlPlane := ctx.Environment().IsExternalControlPlane()
```

pkg/test/framework/components/istio/config.go
```diff
@@ -153,6 +153,9 @@ type Config struct {
 
 	// EnableCNI indicates the test should have CNI enabled.
 	EnableCNI bool
+
+	// custom deployment for ingress and egress gateway
```
frankbu, line 157: Might be good to be clear that this only applies to gateways installed on remote clusters?
```suggestion
	// Custom deployment for ingress and egress gateways on remote clusters.
```

Issue #39302, opened 2022-06-06
debug config-dump race


howardjohn:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/39235/unit-tests_istio/1532392629101137920

Basically our main push flow and our debug flow can co-occur; we expect Generate to not be called concurrently. need some way to serialize it.

PR #39315, opened 2022-06-07
Fix debug race with ads 

hzxuzhonghu:

Fix https://github.com/istio/istio/issues/39302

### DIFF
--- a/pilot/pkg/xds/debug.go
+++ b/pilot/pkg/xds/debug.go
@@ -995,12 +995,14 @@ func cloneProxy(proxy *model.Proxy) *model.Proxy {
 
 	proxy.Lock()
 	defer proxy.Unlock()
-	tmp := *proxy
-	out := &tmp
+	// nolint: govet
+	copied := *proxy
+	out := &copied
 	out.RWMutex = sync.RWMutex{}
 	// clone WatchedResources which can be mutated when processing request
 	out.WatchedResources = make(map[string]*model.WatchedResource, len(proxy.WatchedResources))
 	for k, v := range proxy.WatchedResources {
+		// nolint: govet
 		v := *v
 		out.WatchedResources[k] = &v
 	}
--- a/pilot/pkg/xds/debug.go
+++ b/pilot/pkg/xds/debug.go
@@ -992,8 +992,8 @@ func cloneProxy(proxy *model.Proxy) *model.Proxy {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


pilot/pkg/xds/debug.go
```diff
@@ -987,10 +987,30 @@ func (s *DiscoveryServer) forceDisconnect(w http.ResponseWriter, req *http.Reque
...
 
+func cloneProxy(proxy *model.Proxy) *model.Proxy {
+	if proxy == nil {
+		return nil
+	}
+
+	proxy.Lock()
+	defer proxy.Unlock()
+
+	out := &(*proxy)
```
dddddai, line 998: IIRC this doesn't actually perform a shallow copy, I guess it was optimized by the compiler

hzxuzhonghu: Where did you see that

dddddai: You can run a simple test:
```go
type a struct {
	num int
}

func main() {
	x := &a{1}
	y := &(*x)
	println(x.num)	// 1
	y.num = 2
	println(x.num)	// 2
}
```
but if you change it to
```go
func main() {
	x := &a{1}
	y := *x
	z := &y
	println(x.num) // 1
	z.num = 2
	println(x.num) // 1
}
```
the output will be different

hzxuzhonghu: Thanks for this important finding

pilot/pkg/xds/debug.go
```diff
@@ -987,10 +988,31 @@ func (s *DiscoveryServer) forceDisconnect(w http.ResponseWriter, req *http.Reque
 	_, _ = w.Write([]byte("OK"))
 }
 
+func cloneProxy(proxy *model.Proxy) *model.Proxy {
```
howardjohn, line 991: Interesting! I didn't think of this..

The shallow copy may have other races, although I cannot think of any.

One alternative is to have another `debugChannel` along the `select {push, request}`. That way we serialize it; we can send on it a `[]string` of TypeUrls and then have a `debugResponseChannel` to send the responses back or something. Not sure if its better than this though, WDYT

howardjohn (approved):
LGTM, but we should probably add a test calling PushAll and all(or most) of the debug interfaces in parallel. We have had a lot of issues like this in the past as well

PR #39316, opened 2022-06-07
refactor multi cluster add and update events

ramaraochavali:
We have two problems with multi cluster add/update events because we are using Add + Delete for Update
- For Update we are triggering push (because we reuse Delete)
- The locks are inconsistent between ClusterDeleted and ClusterAdded calls in ClusterUpdated as those functions take locks independently - there is no unified locking in update.

This PR fixes both

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/multicluster.go
+++ b/pilot/pkg/serviceregistry/kube/controller/multicluster.go
@@ -174,8 +174,8 @@ func (m *Multicluster) ClusterUpdated(cluster *multicluster.Cluster, stop <-chan
 // are removed.
 func (m *Multicluster) ClusterDeleted(clusterID cluster.ID) error {
 	m.m.Lock()
-	defer m.m.Unlock()
 	m.deleteCluster(clusterID)
+	m.m.Unlock()
 	if m.XDSUpdater != nil {
 		m.XDSUpdater.ConfigUpdate(&model.PushRequest{Full: true, Reason: []model.TriggerReason{model.ClusterUpdate}})
 	}
@@ -211,7 +211,7 @@ func (m *Multicluster) addCluster(cluster *multicluster.Cluster) (*Controller, *
 func (m *Multicluster) initializeCluster(cluster *multicluster.Cluster, kubeRegistry *Controller, options Options,
 	configCluster bool, clusterStopCh <-chan struct{},
 ) error {
-	client := kubeRegistry.client
+	client := cluster.Client
 
 	if m.serviceEntryController != nil && features.EnableServiceEntrySelectPods {
 		// Add an instance handler in the kubernetes registry to notify service entry store about pod events
--- a/pilot/pkg/serviceregistry/kube/controller/multicluster.go
+++ b/pilot/pkg/serviceregistry/kube/controller/multicluster.go
@@ -144,16 +144,54 @@ func (m *Multicluster) close() (err error) {
 // to watch for resources being added, deleted or changed on remote clusters.
 func (m *Multicluster) ClusterAdded(cluster *multicluster.Cluster, clusterStopCh <-chan struct{}) error {
 	m.m.Lock()
-
-	if m.closing {
+	kubeRegistry, options, configCluster, err := m.addCluster(cluster)
+	if err != nil {
 		m.m.Unlock()
-		return fmt.Errorf("failed adding member cluster %s: server shutting down", cluster.ID)
+		return err
+	}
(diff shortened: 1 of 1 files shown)


hzxuzhonghu:
>The locks are inconsistent between ClusterDeleted and ClusterAdded calls in ClusterUpdated as those functions take locks independently - there is no unified locking in update.


What's the problem?


>For Update we are triggering push (because we reuse Delete)


For update,`ClusterAdded` will trigger push, so we can reduce the trigger calling ClusterDeleted

ramaraochavali:
> What's the problem?

The problem is the multicluster state is mutated with different locks and after delete since we release lock, parts of code might see the cluster not existant while we are updating it (not deleting it)

> For update,ClusterAdded will trigger push, so we can reduce the trigger calling ClusterDeleted

That is what this PR achieves

hzxuzhonghu:
>The problem is the multicluster state is mutated with different locks and after delete since we release lock, parts of code might see the cluster not existant while we are updating it (not deleting it)

It is the same lock, not sure I understand this. Can you clarify by example

stevenctl:
I don't think this fixes the inconsistent state. Seems like you're slightly improving the ordering of pushes though. Even though we avoid the push immediately after "delete", we'll still get some pushes while the new k8s registry syncs, and it likely won't have everything that the old one had.

stevenctl (approved):
I think this is slightly better than what we currently have even though it doesn't fully fix https://github.com/istio/istio/issues/36365. Added a hold to allow @hzxuzhonghu to weigh-in.

hzxuzhonghu:
Yes, this is not inconsistent state.   I am worried if it will introduce other regressions since no tests there

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -144,16 +144,54 @@ func (m *Multicluster) close() (err error) {
...
+
+// ClusterDeleted is passed to the secret controller as a callback to be called
+// when a remote cluster is deleted.  Also must clear the cache so remote resources
+// are removed.
+func (m *Multicluster) ClusterDeleted(clusterID cluster.ID) error {
+	m.m.Lock()
+	defer m.m.Unlock()
+	m.deleteCluster(clusterID)
+	if m.XDSUpdater != nil {
+		m.XDSUpdater.ConfigUpdate(&model.PushRequest{Full: true, Reason: []model.TriggerReason{model.ClusterUpdate}})
```
hzxuzhonghu, line 180: This line is not lock needed

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -166,8 +204,14 @@ func (m *Multicluster) ClusterAdded(cluster *multicluster.Cluster, clusterStopCh
 	m.remoteKubeControllers[cluster.ID] = &kubeController{
 		Controller: kubeRegistry,
 	}
+	return kubeRegistry, &options, configCluster, nil
+}
 
-	m.m.Unlock()
+// initializeCluster initializes the cluster by setting various handlers.
+func (m *Multicluster) initializeCluster(cluster *multicluster.Cluster, kubeRegistry *Controller, options Options,
```
hzxuzhonghu, line 211: nit: options is in `m`

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -166,8 +204,14 @@ func (m *Multicluster) ClusterAdded(cluster *multicluster.Cluster, clusterStopCh
...
+	configCluster bool, clusterStopCh <-chan struct{},
+) error {
+	client := kubeRegistry.client
```
hzxuzhonghu, line 214: kubeRegistry.client is private, you can use cluster.Client here

hzxuzhonghu (commented):
LGTM

ramaraochavali:
> I think this is slightly better than what we currently have even though it doesn't fully fix https://github.com/istio/istio/issues/36365.

Agree. I was not targeting to fix that. This is just an improvement over the current state.

pilot/pkg/serviceregistry/kube/controller/multicluster.go
ramaraochavali: These are modified options

PR #39325, opened 2022-06-07
pilot: fix a bug when merge >2 Services together

howardjohn:
I checked back to release-1.12 and the bug existed there, so I don't think its new (or at least not that new).


GregHanson (approved):

stevenctl (approved):

hzxuzhonghu:
I think it is not found before is that the service merge is inconsistent, here it combines port, but in aggregator it does not.

hzxuzhonghu (approved):

PR #39338, opened 2022-06-07
[tf] flag to only deploy specific apps istio.test.onlyWorkloads

stevenctl:
For workload agnostic tests, this allows targeting specific types of deployments. Example usage:

```
go test tests/integration/pilot/... --istio.test.onlyWorkloads proxyless
```

Not very useful on it's own without ways to skip tests that aren't workload agnostic


howardjohn:
Does `istio.test.skipWorkloads` serve the same purpose? I know its slightly different but seems close enough in practice?

stevenctl:
No longer needed for the original purpose. I could still see it being useful for speeding up local test runs against particular workload types.

Issue #39196, opened 2022-05-30
Invalid Header Field Istio Proxy


jimid27:
### Bug Description

In pursuit of deploying an Opentelemetry Collector to pull Istio metrics and traces, I ran into an issue whereby the Istio-proxy sidecar throws the following error:
```
[2022-05-30T14:51:42.631Z] "- - HTTP/2" 0 DPE http2.invalid.header.field - "-" 0 0 0 - "-" "-" "-" "-" "-" - - 10.128.0.122:4317 10.128.0.122:50082 outbound_.4317_._.otelcol-deployment-collector.istio-system.svc.cluster.local -
```

For testing we have the Open Telemetry collector deployed in the Istio System namespace with the manifest coming from the Istio samples (note did change the service name to match how Otel gets deployed via helm) https://raw.githubusercontent.com/istio/istio/release-1.13/samples/open-telemetry/otel.yaml

My istio configmap looks as follows:
```
apiVersion: v1
data:
  mesh: |-
    defaultConfig:
      discoveryAddress: istiod.istio-system.svc:15012
      tracing:
        zipkin:
          address: zipkin.istio-system.svc.cluster.local:9411
    enablePrometheusMerge: true
    extensionProviders:
    - envoyOtelAls:
        port: 4317
        service: otelcol-deployment-collector.istio-system.svc.cluster.local
      name: otel
    - envoyExtAuthzHttp:
        headersToUpstreamOnAllow:
        - authorization
        - path
        - x-prefect-actor-id
        - x-prefect-actor-permissions
        - Origin
        - Access-Control-Request-Method
        includeRequestHeadersInCheck:
        - authorization
        - Origin
        - Access-Control-Request-Method
        port: "4300"
        service: auth.cloud2.svc.cluster.local
      name: sample-ext-authz-http
    rootNamespace: istio-system
    trustDomain: cluster.local
  meshNetworks: 'networks: {}'
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: istiod
    meta.helm.sh/release-namespace: istio-system
  creationTimestamp: "2022-04-06T18:49:12Z"
  labels:
    app.kubernetes.io/managed-by: Helm
    helm.toolkit.fluxcd.io/name: istiod
    helm.toolkit.fluxcd.io/namespace: istio-system
    install.operator.istio.io/owning-resource: unknown
    istio.io/rev: default
    operator.istio.io/component: Pilot
    release: istiod
  name: istio
  namespace: istio-system
[...]

PR #39365, opened 2022-06-09
do not inject OTel-collector

zirain:

fixed: https://github.com/istio/istio/issues/39196

### DIFF
--- a/samples/open-telemetry/otel.yaml
+++ b/samples/open-telemetry/otel.yaml
@@ -60,6 +60,9 @@ spec:
     type: RollingUpdate
   template:
     metadata:
+      annotations:
+        # do not inject
+        sidecar.istio.io/inject: "false"
       labels:
         app: otel-collector
     spec:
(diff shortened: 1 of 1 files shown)


samples/open-telemetry/otel.yaml
```diff
@@ -60,6 +60,9 @@ spec:
     type: RollingUpdate
   template:
     metadata:
+      annotations:
```
howardjohn, line 63: should make it a label - label si preferred to annotation since it skips the webhook entirely (vs applying the webhook, but a NOP)

howardjohn (commented):
I am OK with this but it _should_ work if we do inject it. The fact that we send an invalid `:authority` header is a real bug that we should fix.

zirain:
> I am OK with this but it _should_ work if we do inject it. The fact that we send an invalid `:authority` header is a real bug that we should fix.

the `:authority` header will be fixed in https://github.com/istio/istio/pull/39476

samples/open-telemetry/otel.yaml
zirain: updated to use label

howardjohn (approved):

craigbox (approved):

PR #39370, opened 2022-06-09
only check for verified identify if xds auth is enabled

ramaraochavali:
In ECDS we verify proxy's VerifiedIdentity for sending config for pulling secrets. But if Istiod runs behind gateway/proxy that terminates TLS, people disable XdsAuth but the proxy is still trusted. So we should only verify identity if XdsAuth is enabled.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/ecds.go
+++ b/pilot/pkg/xds/ecds.go
@@ -21,6 +21,7 @@ import (
 	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 
 	credscontroller "istio.io/istio/pilot/pkg/credentials"
+	"istio.io/istio/pilot/pkg/features"
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pilot/pkg/model/credentials"
 	"istio.io/istio/pilot/pkg/networking/util"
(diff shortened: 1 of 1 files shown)


GregHanson (approved):

howardjohn:
IMO this is dangerous and inconsistent. On sds we have the exact same check. You must be authorized to get secrets - as simple as that. We can't just turn on a mode that says "anyone in the world can access secrets.

Adding an authentication method other than the ones we have today (jwt and mTLS) is reasonable but not this approach IMO

ramaraochavali:
> IMO this is dangerous and inconsistent. On sds we have the exact same check. You must be authorized to get secrets - as simple as that. We can't just turn on a mode that says "anyone in the world can access secrets.

How will this deployments that run Istiod behind Envoy expected to make ECDS (specially WASM) work now?

Is n't it the choice that deployment is making by turning off those flags that some other mechanism is used for auth? `XDS_AUTH_PLAINTEXT` is doing the same for authentication if Istiod is behind a proxy.

> Adding an authentication method other than the ones we have today (jwt and mTLS) is reasonable but not this approach IMO

What would that be in this case? Any suggestions?

howardjohn:
Istiod is basically a wrapper over the api-server. Putting a proxy in front of the api-server is common, but it doesn't mean the api-server gives Secret read access to `unauthenticated`.

If you want your gateway to be a super-admin We could add some concept of this that explicitly allows some identity to impersonate others. I don't think just opening everything up is viable though.

XDS-AUTH was not intended as a config knob, only a feature flag for backwards compatibility (backwards compatible to before we serve highly privileged data over xds)

ramaraochavali:
> If you want your gateway to be a super-admin We could add some concept of this that explicitly allows some identity to impersonate others

We treat Istiod as any other mesh service and deploy Envoy in front of it to take care of mTLS and auth. We expose the plain text port to only localhost so that only local envoy can connect to it. So it is not that whole world can connect to it without authentication. 
So like any other mesh service, we rely on proxy to auth etc.

In this deployment model, I was thinking of using those flags so that it is 100% known that we do not rely on Istiod authentication.

However, if you think, there is some form of authentication for Istiod, can we think about XFCC based auth?

howardjohn:
Ah over localhost, I thought you had a gateway. Localhost is a bit better. XFCC sounds great.

Issue #39303, opened 2022-06-06
Start with development help: trying to just make fails with "make[1]: dirname: Operation not permitted"


SpectralHiss:
### Bug Description

I am trying to start development on a feature and run tests, but so far this is not off to a good start:

steps attempted:
- followed guide [here](https://github.com/istio/istio/wiki/Preparing-for-Development)
- added tag and user
- tried cloning repo fresh in case of permissions problem
- running make invariably runs into this issue:

```code
# make
make
make[1]: dirname: Operation not permitted
make[1]: /bin/bash: Operation not permitted
make[1]: /bin/bash: Operation not permitted
make[1]: /bin/bash: Operation not permitted
make[1]: /bin/bash: Operation not permitted
make[1]: /bin/bash: Operation not permitted
make[1]: /bin/bash: Operation not permitted
make[1]: /bin/bash: Operation not permitted
make[1]: /bin/bash: Operation not permitted
make[1]: /bin/bash: Operation not permitted
make[1]: /bin/bash: Operation not permitted
make[1]: /bin/bash: Operation not permitted
make[1]: /bin/bash: Operation not permitted
Makefile.core.mk:299: warning: overriding recipe for target '/work/out/linux_amd64/'
Makefile.core.mk:299: warning: ignoring old recipe for target '/work/out/linux_amd64/'
make[1]: /bin/bash: Operation not permitted
```

### Version

```prose
#  make -v
GNU Make 4.2.1
Built for x86_64-pc-linux-gnu

# docker version
Client: Docker Engine - Community
 Version:           20.10.8
 API version:       1.41
 Go version:        go1.16.6
 Git commit:        3967b7d
 Built:             Fri Jul 30 19:54:27 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.8
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.6
  Git commit:       75249d8
  Built:            Fri Jul 30 19:52:33 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.9
  GitCommit:        e25210fe30a0a703442421b0f60afac609f950a3
 runc:
  Version:          1.0.1
  GitCommit:        v1.0.1-0-g4144b63
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

Please point me towards what i am obviously doing wrong here :sweat:

[...]

Issue #37315, opened 2022-02-13
Pilot root CA file change does not update sidecar mutatingwebhook caBundle


SpectralHiss:
### Bug Description

When using istio-csr as the identity solution for istio, we disable citadel and load the TLS cert for pilot as well as the root of trust CA from a [configmap](https://github.com/cert-manager/istio-csr/blob/43131ddc7a9d13c86ff21a98b6c55e45c160fec5/hack/istio-config-1.10.0.yaml#L52) .

The current logic for Webhookpatcher and the backing Watcher seems to be notified of filesystem changes through an [istiodCertWatcher](https://github.com/istio/istio/blob/2a5e10406d790acba2e054bf3f0d83dfd7f0e97f/pilot/pkg/bootstrap/sidecarinjector.go#L96) keycertbundle.watcher.
This watcher seems to not have implemented looking at CA file but only at key and cert, forcing a restart of pilot for the caBundle to be correct and causing a lot of confusion with "unknown certificate authority" error.
This can manifest to users, if upgrading to use istio-csr from self-signed or external "plug-in" CA or if doing a break-glass root rotation (which requires a brief period of having both roots eg: https://blog.christianposta.com/diving-into-istio-1-6-certificate-rotation/).

There is a TODO note which demonstrates that support for the ca syncing was always lacking in the webhook patcher  (since that block initialises the watcher the patcher subscribes to):
https://github.com/istio/istio/blob/2a5e10406d790acba2e054bf3f0d83dfd7f0e97f/pilot/pkg/bootstrap/certcontroller.go#L228

Bear in mind our requirement are a little unusual,  we can easily just add logic under that block to watch the CAFile on disk, but one has to remember that in our scenario we have citadel off with ENABLE_CA_SERVER=false. 
So we need to ensure that all other CA methods are not inadvertently affected by this change as well.
I will submit a tentative PR to address above issue.

The wider issue that would be great to solve on top of this one is automatic re-issuance of certificate for sidecars themselves when CA changes either when detected through this path or through a change in the plugin-in ("cacerts" method) thus making the CA mechanisms truly extensible and great to operate but that can be the subject of another PR :)

### Version

```prose
tested on latest istio 1.12.4
and kind environment 1.21
```

bianpengyuan:
[...]

PR #39375, opened 2022-06-09
Add new well known pilot dns TLS cert loading - making it zero config for istio-csr installation

SpectralHiss:
This approach takes an alternative approach to loading pilot DNS certs from secrets for pilot on initialisation proposed by @costinm here.
We had previously opted for trying to load these through the existing tls arguments mechanism in Pilot, demonstrated in PRs (#38845 and #37264 ), 
the previous approach required some extra volume mounts in the helm manifest, it made the UX more difficult for installing istio-csr.
The current approach is to load the TLS certs from known locations if args are not set, and mount those volumes in anyways as `optional: true` volumes.

Well known path locations:
```
	PilotWellKnownDNSCertPath   = "/etc/pilot/tls/"
	PilotWellKnownDNSCaCertPath = "/etc/pilot/ca/"

	DefaultPilotTLSCert   = PilotWellKnownDNSCertPath + "tls.crt"
	DefaultPilotTLSKey    = PilotWellKnownDNSCertPath + "tls.key"
	DefaultPilotTLSCaCert = PilotWellKnownDNSCaCertPath + "root-cert.pem"
```
These match how istio-csr for example (custom plugin CA solution) mounts volumes to those paths (it loads the tls secret from an `istiod-tls` secret and trust anchor from `istio-ca-root-configmap`)

### DIFF
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -951,8 +951,9 @@ func (s *Server) initIstiodCerts(args *PilotArgs, host string) error {
 		// Use the DNS certificate provided via args or in well known location.
 		err = s.initCertificateWatches(TLSOptions{
 			CaCertFile: caCertPath,
-			KeyFile: tlsKeyPath,
-			CertFile: tlsCertPath})
+			KeyFile:    tlsKeyPath,
+			CertFile:   tlsCertPath,
+		})
 
 		if err != nil {
 			// Not crashing istiod - This typically happens if certs are missing and in tests.
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -947,7 +947,7 @@ func (s *Server) initIstiodCerts(args *PilotArgs, host string) error {
 	var err error
 
 	s.dnsNames = getDNSNames(args, host)
-	if hasCustomCertArgsOrWellKnown, tlsCertPath, tlsKeyPath, caCertPath := hasCustomTLSCerts(args.ServerOptions.TLSOptions) ; hasCustomCertArgsOrWellKnown {
+	if hasCustomCertArgsOrWellKnown, tlsCertPath, tlsKeyPath, caCertPath := hasCustomTLSCerts(args.ServerOptions.TLSOptions); hasCustomCertArgsOrWellKnown {
 		// Use the DNS certificate provided via args or in well known location.
 		err = s.initCertificateWatches(tlsCertPath, tlsKeyPath, caCertPath)
 
@@ -1060,27 +1060,27 @@ func (s *Server) createPeerCertVerifier(tlsOptions TLSOptions) (*spiffe.PeerCert
 	return peerCertVerifier, nil
 }
 
-func checkPathsExist(paths ...string) (bool) {
-		for _ , path := range paths {
-			fInfo, err = os.Stat(path)
+func checkPathsExist(paths ...string) bool {
+	for _, path := range paths {
+		fInfo, err = os.Stat(path)
 
-			if err != nil && !fInfo.IsDir() {
-				return true	
-			}		
+		if err != nil && !fInfo.IsDir() {
+			return true
 		}
-		return false
+	}
+	return false
 }
 
 // hasCustomTLSCerts returns true if custom TLS certificates are configured via args.
 func hasCustomTLSCerts(tlsOptions TLSOptions) (ok bool, tlsCertPath, tlsKeyPath, caCertPath string) {
 	// load from tls args as priority
 	if tlsOptions.CaCertFile != "" && tlsOptions.CertFile != "" && tlsOptions.KeyFile != "" {
-		return true,  tlsOptions.CertFile, tlsOptions.KeyFile, tlsOptions.CaCertFile
+		return true, tlsOptions.CertFile, tlsOptions.KeyFile, tlsOptions.CaCertFile
 	} else {
 		if ok = checkPathsExist(constants.DefaultPilotTLSCert, constants.DefaultPilotTLSKey, constants.DefaultPilotTLSCaCert); ok {
 			tlsCertPath = constants.DefaultPilotTLSCert
-			tlsKeyPath =  constants.DefaultPilotTLSKey
-			caCertPath =  constants.DefaultPilotTLSCaCert
+			tlsKeyPath = constants.DefaultPilotTLSKey
+			caCertPath = constants.DefaultPilotTLSCaCert
 			return
 		}
 	}
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -1062,7 +1062,7 @@ func (s *Server) createPeerCertVerifier(tlsOptions TLSOptions) (*spiffe.PeerCert
 
 func checkPathsExist(paths ...string) bool {
 	for _, path := range paths {
-		fInfo, err = os.Stat(path)
+		fInfo, err := os.Stat(path)
 
 		if err != nil && !fInfo.IsDir() {
 			return true
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -1064,7 +1064,7 @@ func checkPathsExist(paths ...string) bool {
 	for _, path := range paths {
 		fInfo, err := os.Stat(path)
 
-		if err != nil || !fInfo.IsDir() {
+		if err != nil || fInfo.IsDir() {
 			return false
 		}
 	}
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -1077,14 +1077,15 @@ func hasCustomTLSCerts(tlsOptions TLSOptions) (ok bool, tlsCertPath, tlsKeyPath,
 	// load from tls args as priority
 	if hasCustomTLSCertArgs(tlsOptions) {
 		return true, tlsOptions.CertFile, tlsOptions.KeyFile, tlsOptions.CaCertFile
-	} else {
-		if ok = checkPathsExist(constants.DefaultPilotTLSCert, constants.DefaultPilotTLSKey, constants.DefaultPilotTLSCaCert); ok {
-			tlsCertPath = constants.DefaultPilotTLSCert
-			tlsKeyPath = constants.DefaultPilotTLSKey
-			caCertPath = constants.DefaultPilotTLSCaCert
-			return
-		}
 	}
+
+	if ok = checkPathsExist(constants.DefaultPilotTLSCert, constants.DefaultPilotTLSKey, constants.DefaultPilotTLSCaCert); ok {
+		tlsCertPath = constants.DefaultPilotTLSCert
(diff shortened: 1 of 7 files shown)


GregHanson:
If you are using VS code as an editor, you can add `gofumpt` as the format tool and that will help with most lint errors:
```
    "go.formatFlags": [
        "[\"-ed\", \"-local\", \"istio.io\"]"
    ],
    "go.formatTool": "gofumpt",
```

costinm:
On the 'well known paths' - I think the current practice is to use /var/run/secrets/istiod instead of /etc/... - one good reason is that some backup apps skip this directory ( or so I heard ).  I would also avoid using the old name 'pilot' - we now use Istiod. Sorry I missed this in the original review.

pilot/pkg/bootstrap/server_test.go
```diff
@@ -203,7 +287,7 @@ func TestReloadIstiodCert(t *testing.T) {
 	}
 
 	// setup cert watches.
-	if err := s.initCertificateWatches(tlsOptions); err != nil {
```
costinm, line 206: I think passing tlsOptions is not a bad idea ( instead of N args ). More stable API.

pilot/pkg/bootstrap/server.go
```diff
@@ -1053,8 +1054,37 @@ func (s *Server) createPeerCertVerifier(tlsOptions TLSOptions) (*spiffe.PeerCert
...
+
+// hasCustomTLSCerts returns the tls cert paths, used both if custom TLS certificates are configured via args or by mounting in well known.
+// while tls args should still take precedence the aim is to encourage loading the DNS tls cert in the well known path locations.
+func hasCustomTLSCerts(tlsOptions TLSOptions) (ok bool, tlsCertPath, tlsKeyPath, caCertPath string) {
+	// load from tls args as priority
+	if hasCustomTLSCertArgs(tlsOptions) {
+		return true, tlsOptions.CertFile, tlsOptions.KeyFile, tlsOptions.CaCertFile
+	}
+
+	if ok = checkPathsExist(constants.DefaultPilotTLSCert, constants.DefaultPilotTLSKey, constants.DefaultPilotTLSCaCert); ok {
```
costinm, line 1076: I am not sure the 'root CA' for custom DNS certs is required or commonly used. AFAIK just the key and cert are needed. 

What is the root used for ?

releasenotes/notes/pilot-load-dns-cert-known-location-deprecate-flags.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+# releaseNotes is a markdown listing of any user facing changes. This will appear in the
+# release notes.
+releaseNotes:
+- |
+  **Changed** Pilot will now load its DNS serving certificate from well known locations:
+```
+/etc/pilot/tls/tls.crt
+/etc/pilot/tls/tls.key
+/etc/pilot/ca/root-cert.pem
```
hzxuzhonghu, line 38: What's the motivation for doing this?

pilot/pkg/bootstrap/server.go
SpectralHiss: Although it is not clear and I had to retrace the code path again, I think the root will be loaded / needed, when the certificatewatch is initiated it will load the CA into the trust bundle [here](https://github.com/istio/istio/blob/84d9f86021c64c65bc5cdc0b236da85d2f525907/pilot/pkg/bootstrap/certcontroller.go#L220) although a filewatcher is missing  (there is a seperate issue about that #37315).
At initialisation the above cacert will be retrieved when creating the (webhook patcher https://github.com/istio/istio/blob/b1bb7c1f2f413c8a1812b6c02187b03a4e729d03/pilot/pkg/bootstrap/sidecarinjector.go#L96) , which will be used to patch the webhook ca field
This does raise however the fact that the discovery server is still initiatised with args, so we need to ensure that's okay also: https://github.com/istio/istio/blob/84d9f86021c64c65bc5cdc0b236da85d2f525907/pilot/pkg/bootstrap/server.go#L735

To fully confirm this i will attempt to not set that flag as we were before and i am pretty certain there will manifest "unknown certificate authority errors" at least for any webhook calls to istiod.

The confusing thing is that when using istio-csr (citadel disabled), the root of trust management is handled by istio-csr which replicated the process in citadel that writes the `istio-root-ca-configmap` in every namespace.

releasenotes/notes/pilot-load-dns-cert-known-location-deprecate-flags.yaml
SpectralHiss: I think it might be helpful to step back in history here :)
Initially we wanted to expose the tls pilot dns cert flags on the helm side, this was to allow helm users to install istio-csr with helm istio, which is currently done with overlays in istio operator:
[example](https://github.com/cert-manager/istio-csr/blob/main/hack/istio-config-1.13.4.yaml)

This was not only crucial for helm istio installation users, but for Maistra Openshift service mesh, where we're trying to have better integration, but under the hood their operator mirrors the istio helm charts.

After doing the initial work of exposing the tls args, Costin thought we could simplify the configuration of istio-csr and make it 0 conf, like it is for gateway ssl certs etc.. how this would work is to load the volumes that istio-csr makes for pilot tls cert but with optional:true , and add some detection logic , just like you can see already implemented in PR, this would mean that if using istio-csr you just need to install istio-csr, say Citadel disabled, point CA_ADDRESS to istio-csr and everything would work fine.
 
https://github.com/istio/istio/pull/38845#issuecomment-1141234213

As to the reason why we have to load the external tls cert in the first place that is because our external CA signers need to talk to a pilot that is signed by same CA, so there is a chicken and egg for trust, so we need to load the pilot cert at installation time for istiod.

Hope that makes sense!

pilot/pkg/bootstrap/server_test.go
SpectralHiss: Understood , will amend this, as well as the path location as commented above

SpectralHiss:
> /test integ-security-multicluster_istio

Thanks for kicking this @GregHanson , I think i fixed all them pesky CLA violation, it was different emails setup on different machines.. once all tests pass again i will change WIP status on PR and await a final review. I will update this [issue](https://github.com/istio/istio/issues/36916) and perhaps close [this PR](https://github.com/istio/istio/pull/38845) if this PR is accepted.

SpectralHiss:
> > /test integ-security-multicluster_istio
> 
> Thanks for kicking this @GregHanson , I think i fixed all them pesky CLA violation, it was different emails setup on different machines.. once all tests pass again i will change WIP status on PR and await a final review. I will update this [issue](https://github.com/istio/istio/issues/36916) and perhaps close [this PR](https://github.com/istio/istio/pull/38845) if this PR is accepted.

Oh and one more thing left, need to change the istio manifest to actually load the known volumes :)

costinm (approved):
Looks good to me.

pilot/pkg/bootstrap/server_test.go
```diff
@@ -40,28 +40,69 @@ import (
...
+	if err := os.WriteFile(t.testCaCertFilePath, testcerts.CACert, 0o644); err != nil { // nolint: vetshadow
+		return fmt.Errorf("WriteFile(%v) failed: %v", t.testCaCertFilePath, err)
 	}
-	if err := os.WriteFile(caCertFile, testcerts.CACert, 0o644); err != nil { // nolint: vetshadow
-		t.Fatalf("WriteFile(%v) failed: %v", caCertFile, err)
+
+	return nil
+}
+
+func cleanupCertFSFles(t TLSFSLoadPaths) error {
```
litong01, line 71: not really sure what FSFles mean? can you name it a bit more verbose?

SpectralHiss: Hey, good point, i've just done that!

SpectralHiss: this is done, PTAL when you have a moment

pilot/pkg/bootstrap/server_test.go
```diff
@@ -40,28 +40,69 @@ import (
...
+
+// This struct will indicate for each test case
+// where tls assets will be loaded on disk
+type TLSFSLoadPaths struct {
+	testTLSCertFilePath string
+	testTLSKeyFilePath  string
+	testCaCertFilePath  string
+}
+
+func TestNewServerCertInit(t *testing.T) {
```
howardjohn, line 94: this requires root to run

pkg/config/constants/constants.go
```diff
@@ -21,6 +21,15 @@ const (
 	// AuthCertsPath is the path location for mTLS certificates
 	AuthCertsPath = "/etc/certs/"
 
+	// PilotWellKnownDNSCertPath is the path location for Pilot dns serving cert, often used with custom CA integrations
+	PilotWellKnownDNSCertPath   = "/var/run/secrets/istiod/tls/"
```
howardjohn, line 25:
```suggestion
	PilotWellKnownDNSCertPath   = "./var/run/secrets/istiod/tls/"
```

To avoid needing root

howardjohn (changes requested):

pkg/config/constants/constants.go
SpectralHiss: ah nice, so the workdir of the pod is always the filesystem root? is that why this would work in both contexts?

pilot/pkg/bootstrap/server_test.go
SpectralHiss: ah yes i wasn't running this within my laptop but on CI instead so forgot about this, this is due to[ this other issue](https://github.com/istio/istio/issues/39303) i have to run on my laptop but that's besides the point. thanks!

pkg/config/constants/constants.go
howardjohn: Yeah we make this assumption in a variety of places; for non-k8s it allows running in other directories if desired. Nice for testing as well

pilot/pkg/bootstrap/server_test.go
```diff
@@ -119,6 +192,17 @@ func TestNewServerCertInit(t *testing.T) {
 		t.Run(c.name, func(t *testing.T) {
 			test.SetForTest(t, &features.PilotCertProvider, c.certProvider)
 			test.SetForTest(t, &features.EnableCAServer, c.enableCA)
+
+			// check if we have some tls assets to write for test
+			if c.FSCertsPaths != (TLSFSLoadPaths{}) {
+				err := loadCertFilesAtPaths(c.FSCertsPaths)
```
howardjohn, line 198: nit: we might be able to do `dir := t.TempDir; os.Chdir(dir)` then everything will be relative to our auto-cleaned up tmp dir.

howardjohn (commented):
(recent comment is not blocking, I just didn't have a chance to review the real code yet)

SpectralHiss:
> still seems to require root to run

Sorry, i just missed a dot for the `PilotWellKnownDNSCaCertPath` , can you try again because i still have the other problem that i can't run locally without hanging and maxing out my disk: https://github.com/istio/istio/issues/39303

howardjohn (approved):

howardjohn (approved):
lgtm, thanks! And I apologize for the delay in reviews, appreciate the tenacity here to get this merged

PR #39384, opened 2022-06-09
tf: filter extra workloads after test filters run

howardjohn:
We currently have some logic to filter out "not useful" tests (a->b and
b->c are the same). However, we have some tests that need many
source/destinations. If we filter out these too early, we may lose the
ability to run the test at all.

Instead, we run this filter last and ensure we leave enough
destinations.

Change-Id: I12496b05c20b5b3a1333076e164ea76dba3779a2

### DIFF
--- a/pkg/test/framework/components/echo/echotest/filters.go
+++ b/pkg/test/framework/components/echo/echotest/filters.go
@@ -17,7 +17,7 @@ package echotest
 import (
 	"istio.io/istio/pkg/test/framework/components/echo"
 	"istio.io/istio/pkg/test/framework/components/echo/match"
-	"istio.io/istio/pkg/util/sets"
+	"istio.io/pkg/log"
 )
 
 type (
@@ -32,7 +32,9 @@ type (
 //       Run()
 func (t *T) From(filters ...Filter) *T {
 	for _, filter := range filters {
+		log.Errorf("howardjohn: filter before %v", len(t.sources))
 		t.sources = filter(t.sources)
+		log.Errorf("howardjohn: filter after %v", len(t.sources))
 	}
 	return t
 }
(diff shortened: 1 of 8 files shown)


pkg/test/framework/components/echo/echotest/filters.go
```diff
@@ -97,13 +101,19 @@ func (t *T) applyCombinationFilters(from echo.Instance, to echo.Instances) echo.
 //     The plain-pods are a, b and c.
 //     This filter would result in a, headless, naked and vm.
 // TODO this name is not good
-func SingleSimplePodServiceAndAllSpecial(exclude ...echo.Instance) Filter {
+func SingleSimplePodServiceAndAllSpecial(min int, exclude ...echo.Instance) Filter {
```
stevenctl, line 104: Does the name "single" still make sense?

stevenctl:
LGTM

pkg/test/framework/components/echo/echotest/filters.go
```diff
@@ -28,11 +28,13 @@ type (
 // From applies each of the filter functions in order to allow removing workloads from the set of clients.
 // Example:
 //     echotest.New(t, apps).
-//       From(echotest.SingleSimplePodServiceAndAllSpecial, echotest.NoExternalServices).
+//       From(echotest.SimplePodServiceAndAllSpecial, echotest.NoExternalServices).
 //       Run()
 func (t *T) From(filters ...Filter) *T {
 	for _, filter := range filters {
+		log.Errorf("howardjohn: filter before %v", len(t.sources))
```
ericvn, line 35: Do you want to leave the howardjohn: in here? Also in 2 lines?

GregHanson (approved):

stevenctl (approved):

shankgan (approved):

PR #39405, opened 2022-06-10
add xfcc authenticator

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [X] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali:
@howardjohn PTAL

ramaraochavali:
I am getting the following error while running make gen on mac M1. Any suggestions on how to get it working?

WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested

pilot/pkg/xds/auth.go
```diff
@@ -32,7 +32,7 @@ import (
 )
 
 var AuthPlaintext = env.RegisterBoolVar("XDS_AUTH_PLAINTEXT", false,
-	"Authenticate plain text requests - used if Istiod is behind a gateway handling TLS").Get()
+	"Authenticate plain text requests - used if Istiod is behind a gateway or proxy handling TLS").Get()
```
GregHanson, line 36: just to double check - this is unrelated to your xfcc changes or did you forget to check it?  I don't see the `AuthPlaintext` var called in your other changes

ramaraochavali: This is unrelated. I just updated the description of the flag.

GregHanson: so the incoming xfcc authentication logic doesn't need a feature flag or proxy version check?

pilot/pkg/bootstrap/server.go
```diff
@@ -308,6 +308,7 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
 	// is used as the authentication result.
 	authenticators := []security.Authenticator{
 		&authenticate.ClientCertAuthenticator{},
+		&authenticate.XfccAuthenticator{},
```
howardjohn, line 311: we can't enable this by default or you could just set the header

howardjohn: I think we either need
- check identity with one of the other methods, assert it's an identity we configured as "can delegate" and (ie gateway) and then allow the xfcc check
- flag to enable

cc @costinm has been looking into much of the same

costinm: Right -  XFCC requires double authentication. See my docs on gateway improvements as well.

The request should be authenticated with ClientCert or JWT - and if the gateway doesn't support this and we are running on a 'secure' network, we can use CIDR/IP matching to validate the request is only from a gateway. It is also possible to use NetworkPolicy - but only if all access to Istiod is blocked by policy - except gateway.

In the last case - NetworkPolicy enforcing only possible client to Istiod is the GW - we can have the XfccAuthenticator as in this PR, but we should have some env to indicate 'network policy' mode.

For all other cases - we will need some extra checks, again see my proposal, it is not an istiod-specific problem but all workloads behind a gateway ( including istio gateways behind other LBs) will need this mechanism.

pilot/pkg/xds/auth.go
costinm: If you are updating this - please change it to 'if Istiod is running on a secure/trusted network'. 

"or proxy handling TLS' is not relevant here - plaintext is about not doing mTLS because the low-level network provide equivalent guarantees, having a proxy in front handling mTLS is an issue of authorizing the proxy.

costinm (commented):
Thanks a lot for starting this, I think having Istiod behind a istio or non-istio gateway terminating is 
the best deployment mode for Istio and it forces us to add the elements that are needed to secure
auth integration with istio/non-istio infra.

But as all critical security changes I think we'll need to be very careful.

costinm:
Another comment - I looked at the library you are using and it's not bad, but it may be better to add a dependency to a generic library for parsing 'http structured' headers - like https://pkg.go.dev/github.com/gobwas/httphead ( I didn't do a lot of research into what's the most popular or if golang core libraries support parsing structured headers  - but I believe XFCC is following the model )

security/pkg/server/ca/authenticate/xfcc_authenticator_test.go
```diff
@@ -0,0 +1,81 @@
...
+					"hello.east.example.com",
+					"spiffe://mesh.example.com/ns/otherns/sa/othersa",
+				},
+			},
+		},
+	}
+
+	auth := &XfccAuthenticator{}
+
+	for id, tc := range testCases {
```
howardjohn, line 55: nit: use `t.Run` for subtest, then you do not need `Case: ...`

pilot/pkg/bootstrap/server.go
ramaraochavali: The use case here is envoy front ending Istiod and it takes care of TLS termination. So the connection is from localhost. I think if the peer connection is from localhost and Xfcc Authenticator is enabled, we can rely on xfcc authenticator - Do you agree or any concerns?

ramaraochavali:
@costinm @howardjohn I have addressed all comments except for library comment - as I could not find any better http header parser. We can continue with the current library and if we find a better one, we can swap it out.

pilot/pkg/xds/auth.go
```diff
@@ -57,15 +58,18 @@ func (s *DiscoveryServer) authenticate(ctx context.Context) ([]string, error) {
 		return nil, nil
 	}
 	authFailMsgs := []string{}
+	authContext := security.NewAuthContext(ctx)
 	for _, authn := range s.Authenticators {
-		u, err := authn.Authenticate(ctx)
-		// If one authenticator passes, return
+		u, err := authn.Authenticate(authContext)
```
ramaraochavali, line 63: This is one change - we are now running all the authenticators.

costinm:
Agree, we can continue with that parser until we find a better one.

pilot/pkg/xds/auth.go
costinm: I expect xfcc to be last, and to swap the identity set by the previous ones if it is trusted.

This is the kind of behavior that should have a proper doc and be reviewed for security, but at least add extensive comments.

pkg/security/security.go
```diff
@@ -351,12 +351,37 @@ type Caller struct {
...
 
+func NewAuthContext(ctx context.Context) AuthContext {
+	return AuthContext{RequestContext: ctx, Authenticators: make(map[string]*Caller)}
+}
+
+func (ac *AuthContext) AddAuthenticator(authenticator string, caller *Caller) {
+	ac.Authenticators[authenticator] = caller
+}
+
+func (ac *AuthContext) Identities() []string {
```
costinm, line 375: Not sure this is correct - for xfcc we need to only return the one from xfcc. 

Another option is to return separate identities:
- remote endpoint identity -  from xfcc, jwt or peer cert (in this order)
- connection peer identity - from socket cert or jwt or IP (if cidr range is used)
- user identity - for the case where both mtls and jwt are used, the mtls cert identifies the backend calling and jwt the actual user.

security/pkg/server/ca/authenticate/xfcc_authenticator.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (xff *XfccAuthenticator) AuthenticatorType() string {
+	return XfccAuthenticatorType
+}
+
+// Authenticate extracts identities from Xfcc Header.
+func (xff *XfccAuthenticator) Authenticate(ctx security.AuthContext) (*security.Caller, error) {
+	// TODO: Extend this for other "trusted" IPs based on network policies.
+	if !isLocalHost(ctx.RequestContext) {
+		return nil, fmt.Errorf("call is not from trusted network, xfcc can not be used as authenticator")
```
costinm, line 48: I really don't like the fact we treat normal occurance as error. Now that we call all auth ticators we can remove the error - there is nothing wrong about not having xfcc header.

security/pkg/server/ca/authenticate/xfcc_authenticator.go
```diff
@@ -0,0 +1,110 @@
...
+
+func buildSecurityCaller(xfccHeader string) (*security.Caller, error) {
+	clientCerts, err := xfccparser.ParseXFCCHeader(xfccHeader)
+	if err != nil {
+		return nil, fmt.Errorf("error in parsing xfcc header: %v", err)
+	}
+	ids := []string{}
+	for _, cc := range clientCerts {
+		ids = append(ids, cc.DNS...)
+		if len(cc.URI) > 0 {
```
costinm, line 101: This may create problems. It is good to return all identities/claims in the auth context, but we need to clearly identify the spifee identity of the remote - that's what istiod really uses.

costinm:
Sorry for the nitpicking - this is pretty important and risky.

costinm (commented):
Main concern is returning the right identity and making sure we can support remote gateways authenticating themselves with mtls.

costinm:
For more context: the idea would be to have the east/west or ingress
gateway to act as entry point and terminate
most XDS connections. It would also apply standard istio configs -
including RBAC, traffic split, etc and use
ACME or standard certificates.

I did explore the idea of having the east-west gateway co-located with
Istiod - it simplifies deployment but
doesn't sound right. I don't mind having it as an option.

ramaraochavali:
> Looking at go.sum.diff - did we now pick a dep on kong and thrift ?

That is not expected. Will take a look.

pilot/pkg/xds/auth.go
ramaraochavali: Mtls is first and xfcc is last.  It appends to the list of identities that previous ones have already added. 

> Not sure how this works - first wins?

Any one wins. The identities are merged together.

security/pkg/server/ca/authenticate/xfcc_authenticator.go
```diff
@@ -0,0 +1,110 @@
...
+	xfccHeader := meta.Get(xfccparser.ForwardedClientCertHeader)[0]
+	return buildSecurityCaller(xfccHeader)
+}
+
+func isLocalHost(ctx context.Context) bool {
+	peerInfo, _ := peer.FromContext(ctx)
+	if net.ParseIP(peerInfo.Addr.String()).IsLoopback() {
+		return true
+	}
+	ifaces, _ := net.Interfaces()
```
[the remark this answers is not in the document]
ramaraochavali: I like the idea of changing bool to allowed trusted CIDR list.

ramaraochavali:
@costinm Thank you for your thorough review. I have now introduced concept of DelegatingAuthenticators and added CidrAuthenticator that validates whether the request is from trusted Cidrs, if it is it delegates authentication to XfccAuthenticator. This should help the gateway use case also. PTAL.

regarding kong and go thrift, kong is a [parser](https://github.com/alecthomas/kong) that [participle](https://github.com/alecthomas/participle) uses in examples same with go thrift. I think we do not directly depend on it that is why it is not in go.mod

costinm:
On Fri, Jun 17, 2022 at 9:41 PM Rama Chavali ***@***.***>
wrote:

>
> In pilot/pkg/xds/auth.go
> <https://github.com/istio/istio/pull/39405#discussion_r900696816>:
>
> >  	for _, authn := range s.Authenticators {
> -		u, err := authn.Authenticate(ctx)
> -		// If one authenticator passes, return
> +		u, err := authn.Authenticate(authContext)
>
> Mtls is first and xfcc is last. It appends to the list of identities that
> previous ones have already added.
>
> Not sure how this works - first wins?
>
> Any one wins. The identities are merged together.
>
That's not correct behavior AFAIK - we don't want the identity of the
(trusted) gateway in the list.
Or maybe we want it ( to know the path ) - but the result of auth should be
the identity of the real client.

The authenticators are also used for generating certs ( we do support cert
signing based on a client cert) - and
the cert has one NS and KSA, which should be of the original client.

costinm:
Great, thanks. I suspect we'll iterate on this - no need to implement
everything in a single PR or solve all use cases.

Will take another look - hopefully other people will take a look as well,
security is very hard.

See my comment on end result of authentication reflecting only the original
client ( it is fine if the auth context
keeps track of connection peer and JWT - but if XFCC is present, that
should be used in XDS and for cert generation ).

pkg/security/security.go
```diff
@@ -351,12 +352,36 @@ type Caller struct {
...
+type AuthContext struct {
+	// RequestContext is the context from request.
+	RequestContext context.Context
+	// Authenticators is the list of authenticators that were executed before in the auth chain.
+	Authenticators []string
+	// DelegatedAuthenticators are the list of delegated authenticators that main authenticators can
+	// choose to get identities from. For example, CidrAuthenticator can delegate the authentication to
+	// XfccAuthenticator once it validates the connection is from trusted cidr range. XfccAuthenticator
+	// can then extract identities from peer certificate.
+	DelegatedAuthenticators []Authenticator
```
howardjohn, line 364: Do we need this? It seems to make the calling complex. I was thinking we would just pass in a `Identities []string` as part of the inputs, which is the "currently known identities".  Or I guess if we are doing CIDR auth instead of based on the callers real identity, then just some bool that says "trustednetwork" or whatever and XFCC reads that variable.

hzxuzhonghu: I am confused even with the comments

security/pkg/server/ca/authenticate/cidr_authenticator.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Authenticate extracts identities from trusted cidr ranges.
+func (c *CidrAuthenticator) Authenticate(ctx security.AuthContext) (*security.Caller, error) {
+	peerInfo, _ := peer.FromContext(ctx.RequestContext)
+	if !isAuthenticated(peerInfo.Addr.String()) {
+		return nil, fmt.Errorf("")
+	}
+	ctx.AddDelegatedAuthenticator(XfccAuthenticator{})
+	return &security.Caller{AuthSource: security.AuthSourceDelegate, Identities: []string{peerInfo.Addr.String()}}, nil
```
hzxuzhonghu, line 49: I donot think identities should be equal to peer address

pilot/pkg/bootstrap/server.go
```diff
@@ -308,6 +308,7 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
 	// is used as the authentication result.
 	authenticators := []security.Authenticator{
 		&authenticate.ClientCertAuthenticator{},
+		&authenticate.CidrAuthenticator{},
```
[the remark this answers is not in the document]
ramaraochavali: It can be any where now. If it is configured to use the Cidrs it will be used otherwise it is not. Having said that, I can move it to end if you prefer that

security/pkg/server/ca/authenticate/cert_authenticator.go
```diff
@@ -43,8 +42,8 @@ func (cca *ClientCertAuthenticator) AuthenticatorType() string {
 // method assumes that certificate chain has been properly validated before
 // this method is called. In other words, this method does not do certificate
 // chain validation itself.
-func (cca *ClientCertAuthenticator) Authenticate(ctx context.Context) (*security.Caller, error) {
-	peer, ok := peer.FromContext(ctx)
+func (cca *ClientCertAuthenticator) Authenticate(ctx security.AuthContext) (*security.Caller, error) {
```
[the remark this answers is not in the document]
ramaraochavali: I was n't thinking about it.  Are you thinking that if a gateway terminates TLS and initiate a new TLS connection, we should honour XFCC? But how do we know if it is gateway or the actual proxy?

pkg/security/security.go
ramaraochavali: This is not just for XFCC case. We could add other delegated authenticators later if needed. For XFCC though, having the bool may be sufficient.

security/pkg/server/ca/authenticate/cidr_authenticator.go
ramaraochavali: Please see the comment - It will be swapped with XFCC identity because it added Delegated Authenticator.

costinm:
I would also do the delegation in code - too early to abstract, we only
have this special use case.

ramaraochavali:
> I would also do the delegation in code - too early to abstract, we only
have this special use case.

Are you saying cidr authenticator internally runs xfcc header authenticator and populate identities?

costinm:
No, XFCC authenticator does checks on the remote IP against the CIDR range,
it has nothing to do with a future CIDR authenticator.

I think we are talking about 2 different CIDR authenticators. What you have
in this PR is not really an authenticator - it checks CIDR
ranges, but doesn't determine the identity of the peer. What I am proposing
- for a different PR - is to add a proper CIDR authenticator
that first check the range is 'trusted', and then find the proper namespace
and service account by looking up based on IP, since Istiod
has all the pods.

costinm:
On Wed, Jun 22, 2022 at 6:48 PM Zhonghu Xu ***@***.***> wrote:

> Is there a similar design we can borrow? I think a general solution can be
> somewhere
>
For XFCC - there are quite a few 'delegating auth' designs, including K8S
proxy.

For using the IP to derive the namespace/sa - I don't think I've seen this
anywhere. But that's
a separate proposal, specifically for users who have already secure network
layer and don't
need mTLS - or to add another layer of 'defense in depth'.

ramaraochavali:
Let us solve Xfcc now and we can iterate on this.

ramaraochavali:
Looking at kubernetes docs,  K8s already supports authenticating a proxy https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy which is mostly similar to xfcc header.

pkg/security/security.go
```diff
@@ -351,12 +352,36 @@ type Caller struct {
...
+}
+
 type Authenticator interface {
-	Authenticate(ctx context.Context) (*Caller, error)
+	Authenticate(ctx AuthContext) (*Caller, error)
 	AuthenticatorType() string
 	AuthenticateRequest(req *http.Request) (*Caller, error)
 }
 
+func NewAuthContext(ctx context.Context) AuthContext {
```
kdorosh, line 373: this should return a pointer to an `AuthContext` or else we will just copies around

kdorosh (changes requested):
we should be using pointers to `AuthContext` to avoid bugs (the CIDR auth will add the delegated auth to its _copy_ of the auth context, but this will not persist in the original copy)

also we need to add delegated auth to the istio ca server, not just xds. a gist with a diff for fixes that worked for me is here: https://gist.github.com/kdorosh/648f386285e996e7462d58fcab2741ef

security/pkg/server/ca/authenticate/cidr_authenticator.go
kdorosh: why return an identity at all? it seems like it will be entirely ignored. either the delegate authenticates and we return the swapped identity, or we return authn failure.. returning anything here is confusing imo

kdorosh (changes requested):
we also need to add delegated auth support to `AuthenticateRequest` not just `Authenticate`

ramaraochavali:
@kdorosh yes. I am aware of changes needed in istio CA path and authenticate request path. I want to iterate on the main authenticate and once we agree on the solution - I will make changes to those places.

security/pkg/server/ca/authenticate/cidr_authenticator.go
```diff
@@ -0,0 +1,87 @@
...
+	if len(features.TrustedCIDRRanges) > 0 {
+		cidrs := strings.Split(features.TrustedCIDRRanges, ",")
+		for _, cidr := range cidrs {
+			if isInRange(addr, cidr) {
+				return true
+			}
+		}
+	}
+	// Always trust local host addresses.
+	if net.ParseIP(addr).IsLoopback() {
```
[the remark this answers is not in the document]
ramaraochavali: yes. Correct.

security/pkg/server/ca/authenticate/cidr_authenticator_test.go
```diff
@@ -0,0 +1,83 @@
...
+			caller: nil,
+		},
+	}
+
+	auth := &CidrAuthenticator{}
+
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			if len(tt.cidr) > 0 {
+				test.SetStringForTest(t, &features.TrustedCIDRRanges, tt.cidr)
```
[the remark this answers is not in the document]
ramaraochavali: added one more case to cover this

ramaraochavali:
@costinm PTAL. I have changed based on discussions. Currently I have added feature flag to specify trusted CIDRs, we can easily extend it to networks.

pilot/pkg/features/pilot.go
```diff
@@ -403,6 +403,13 @@ var (
 		"If enabled, pilot will authorize XDS clients, to ensure they are acting only as namespaces they have permissions for.",
 	).Get()
 
+	TrustedCIDRRanges = env.RegisterStringVar(
```
costinm, line 406: Let's add a comment here to avoid future confusion - this is temporary and experimental, will be replaced with Networks ( or something similar ).

Ok to get this in initially, and all env variables are considered internal/experimental - but this one in particular already has an existing API, and there are many reason to define it there.

Also, for name I would use TRUSTED_GATEWAY_CIDR - I don't think support for XFCC can be restricted to only PILOT, and RANGES seems redundant, but if you have to keep PILOT_ prefix I'm ok.
However I would rather qualify it - TRUSTED_CIDR can also mean networks where we don't need mTLS,
for pod to pod. 

"This can only be used when the network where Istiod and the authenticating gateways are running is a trusted/secure network"

pilot/pkg/xds/auth.go
costinm: Not sure the method signature is right - no need to return 'u' - authenticators should modify the context.

Also err is not clear how to use - maybe have a field in the context with error conditions. I believe the current model is pretty bad - it is not an error to not have a JWT or mTLS client cert - but it is an error to have an expired JWT or mTLS cert, or invalid XFCC format. A real error in one of the authenticators should abort, for example if mTLS is good and XFCC is bad we shouldn't let the mTLS identity stand.

pilot/pkg/xds/auth.go
```diff
@@ -57,15 +58,18 @@ func (s *DiscoveryServer) authenticate(ctx context.Context) ([]string, error) {
...
 	authFailMsgs := []string{}
+	authContext := security.NewAuthContext(ctx)
+
 	for _, authn := range s.Authenticators {
-		u, err := authn.Authenticate(ctx)
+		authContext.AddAuthenticator(authn.AuthenticatorType())
+		u, err := authn.Authenticate(authContext)
 		// If one authenticator passes, return
 		if u != nil && u.Identities != nil && err == nil {
+			authContext.Caller = u
```
costinm, line 68: The problem with 'identities' is that there is no guarantee there is only one identity and it is spiffee first. 
The request may have multiple identities - but we care only about the spiffee one ( and really about namespace + KSA - Istiod is not concerned with spiffee, just namespace ).

pkg/security/security.go
```diff
@@ -351,12 +351,37 @@ type Caller struct {
 	Identities []string
 }
 
+type AuthContext struct {
+	// RequestContext is the context from request.
+	RequestContext context.Context
```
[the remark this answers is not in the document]
costinm: We can do this in a future iteration. I meant: have http.Request and grpc Context in AuthContext, so the interface doesn't need 2 methods.

pkg/security/security.go
```diff
@@ -351,12 +352,42 @@ type Caller struct {
...
 	AuthenticateRequest(req *http.Request) (*Caller, error)
 }
 
+// DelegatedAuthenticator determines whether the main authenticator can trust the
+// client and proceed with authentication. This is used in cases where the client
+// authentication can not be directly determined from the request(no TLS certs, JWT etc)
+// but request has additional information to authenticate.
+// For example XfccAuthenticators should validate whether client is from trusted Cidr range
+// before it can trust the Xfcc Header.
+type DelegatedAuthenticator interface {
```
costinm, line 378: I would not define an interface for this. Just code the logic in the authenticator. There are all kind of logical issues here - you may trust the caller but not as an authenticating/delegating proxy, etc.

Too early to abstract this - we have a single case so far.

pkg/security/security.go
```diff
@@ -351,12 +352,42 @@ type Caller struct {
...
+	// CanTrustCaller indicates whether caller is a trusted caller.
+	CanTrustCaller(ctx *AuthContext) bool
+}
+
+func NewAuthContext(ctx context.Context) *AuthContext {
+	return &AuthContext{RequestContext: ctx}
+}
+
+func (ac *AuthContext) AddAuthenticator(authenticator string) {
```
costinm, line 387: Why ? Each authenticator can add itself to the list in context ( and they can do that if they have anything to say - for example if there is no mTLS context, the mtls auth wouldn't add itself ). 

Also I think if we want to keep information about what authenticators where used - a map may be better, so each authenticator can also add its results. It's useful for debug and may be useful for further logic in Istiod.

security/pkg/server/ca/authenticate/cidr_authenticator.go
```diff
@@ -0,0 +1,68 @@
...
+
+// CanTrustedCaller determines if the caller is in trusted cidr range.
+func (c *CidrAuthenticator) CanTrustCaller(ctx *security.AuthContext) bool {
+	peerInfo, _ := peer.FromContext(ctx.RequestContext)
+	return isTrustedAddress(peerInfo.Addr.String())
+}
+
+func isTrustedAddress(addr string) bool {
+	if len(features.TrustedCIDRRanges) > 0 {
+		cidrs := strings.Split(features.TrustedCIDRRanges, ",")
```
costinm, line 42: So we split, create objects, etc on every single request ? May be ok for an experimental feature, but add at least comments that this can't be the final form. 

Also see my other comments - it would be much easier to review and understand if you kept this code local to XFCC authenticator, without attempting to abstract it too early.

security/pkg/server/ca/authenticate/xfcc_authenticator.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (xff XfccAuthenticator) AuthenticatorType() string {
+	return XfccAuthenticatorType
+}
+
+// Authenticate extracts identities from Xfcc Header.
+func (xff XfccAuthenticator) Authenticate(ctx *security.AuthContext) (*security.Caller, error) {
+	// First check if client is trusted client so that we can "trust" the Xfcc Header.
+	delegated := &CidrAuthenticator{}
+	if !delegated.CanTrustCaller(ctx) {
```
costinm, line 45: I see you already used CidrAuthenticator internally - so just move it here and remove the 'authenticator' abstraction. Will just make it difficult for future work on the 'real' IP authenticator for trusted networks.

security/pkg/server/ca/authenticate/xfcc_authenticator.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, fmt.Errorf("caller is not in the trusted network. XfccAuthenticator can not be used")
```
costinm, line 46: I would add the caller IP - easier to debug and actionable ( user can figure what range to add )

security/pkg/server/ca/authenticate/xfcc_authenticator.go
```diff
@@ -0,0 +1,82 @@
...
+
+func buildSecurityCaller(xfccHeader string) (*security.Caller, error) {
+	clientCerts, err := xfccparser.ParseXFCCHeader(xfccHeader)
+	if err != nil {
+		return nil, fmt.Errorf("error in parsing xfcc header: %v", err)
+	}
+	ids := []string{}
+	for _, cc := range clientCerts {
+		if len(cc.URI) > 0 {
+			ids = append(ids, cc.URI)
```
costinm, line 74: IMO  the final result - the spiffee URI - should be set in the auth context by the xfcc authenticator. The split logic is very hard to evaluate.

Since we return a list of identities - we should probably include the other identities  DNS, email, CN from the client certificate. We don't use them here - but the interface is generic.

security/pkg/server/ca/authenticate/xfcc_authenticator.go
```diff
@@ -0,0 +1,82 @@
...
+	return buildSecurityCaller(xfccHeader)
+}
+
+func buildSecurityCaller(xfccHeader string) (*security.Caller, error) {
+	clientCerts, err := xfccparser.ParseXFCCHeader(xfccHeader)
+	if err != nil {
+		return nil, fmt.Errorf("error in parsing xfcc header: %v", err)
+	}
+	ids := []string{}
+	for _, cc := range clientCerts {
```
costinm, line 72: Is this correct ? Does it return the LEAF cert only, or the chain ? 

If client presents multiple identities ( which I don't think Istio will ever do, and is not very easy ) we should select the URI starting with spiffee and having one of our trust domains. 
Or just reject the request - a client with multiple workload identities is not well defined and too early to support. Let's keep it simple...

costinm (commented):
Many of my recent comments can be addressed in a future iteration, but some probably should be fixed in the PR. In particular the separate CIDR authenticator, which I think complicates things unnecessarily and makes the security logic hard to follow and trust.

security/pkg/server/ca/authenticate/cidr_authenticator.go
ramaraochavali: Auth only happens on initial connection

security/pkg/server/ca/authenticate/xfcc_authenticator.go
ramaraochavali: This is all the certs in the call chain - for which ever proxies it has passed through. Typically it will be one. I have added validations to reject if it zero or more than 1.

ramaraochavali:
@costinm Agree with most of the comments and left TODOs for others. One thing I want to do as follow-up is to remove the return values from `authenticate`  and rely on AuthContext. But let us get this in and I will make follow-ups

security/pkg/server/ca/authenticate/xfcc_authenticator.go
costinm: The remote identity is in the first cert I believe - it is fine to have the identity of the other proxies, but mostly irrelevant, it's whatever the proxy in front of us happens to claim. We can't verify anything - just trust the one in front of us. May be useful for debugging/tracing/logging - but for auth it's not relevant I think, we make no use of the proxy path.

What we need to be sure is that identity of a proxy in front doesn't accidentally get treated as identity of the peer - proxies tend to have high SDS permissions. AFAIK Istiod makes all decisions based on the final namespace/KSA extracted from the initial mTLS ( first trusted proxy in the chain ) - I never fully understand why we pass all those identities around in the interface, but they should certainly not include any of the proxies in the chain.

costinm:
Agree, we can do the signature change later and most of the other things - but I think we need to make sure we extract the right identity from the chain ( client - and not any of the trusted gateways/proxies ), and that we get the spiffe one.

pkg/security/security.go
```diff
@@ -351,12 +352,41 @@ type Caller struct {
...
+	Identities []string
+	Messages   []string
+}
+
+// AuthContext carries the contextual information and is passed across authenticators.
+type AuthContext struct {
+	// RequestContext is the context from request.
+	RequestContext context.Context
+	// Authenticators is the list of authenticators that were executed before in the auth chain.
+	Authenticators map[string]AuthResult
```
hzxuzhonghu, line 366: I donot think this is a good struct, making the AuthContext aware of who may use it.

pilot/pkg/features/pilot.go
```diff
@@ -403,6 +403,14 @@ var (
 		"If enabled, pilot will authorize XDS clients, to ensure they are acting only as namespaces they have permissions for.",
 	).Get()
 
+	// TODO: Move this to proper API.
+	TrustedGatewayCIDR = env.RegisterStringVar(
+		"TRUSTED_GATEWAY_CIDR",
+		"",
+		"If set, any connections from gateway to Istiod with this CIDR range are treated as trusted for using authenication mechanisms like XFCC."+
+			" This can only be used when the network where Istiod and the authenticating gateways are running is a trusted/secure network",
+	).Get()
```
hzxuzhonghu, line 412: cidr does not reflect the xfcc

hzxuzhonghu (dismissed):

pilot/pkg/features/pilot.go
ramaraochavali: Can you elobrate more? We are saying here that XFCC is used only when the connection is from trusted CIDR.

pkg/security/security.go
ramaraochavali: Can you please exlain more on why it is not a good struct? AuthContext is a shared struct - to propagate the context of authenticators.

pilot/pkg/features/pilot.go
hzxuzhonghu: cidr has no direct relationship with xfcc headers

pkg/security/security.go
hzxuzhonghu: The context should not contain `Authenticators ` 

context is passed to each authenticator, and it is a arg, not should be aware of which authenticator's decision. 

And I do not see the need to record who(authenticator) auth who (client) with what result, because a successful auth will make istiod break the loop.

pilot/pkg/features/pilot.go
ramaraochavali: It has no direct relationship - but please comments if xfcc authenticator has to be used, the connection address should be in trusted CIDR. Otherwise we can not trust XFCC header.

pkg/security/security.go
ramaraochavali: > context is passed to each authenticator, and it is a arg, not should be aware of which authenticator's decision.

That is for other authenticators to check which all authenticators have been passed unsuccessfully and what is their errors are. Successful authentication currently breaks the loop but before successful authenticator there can be failed authenticators that each authenticator can look at.

hzxuzhonghu: >That is for other authenticators to check which all authenticators have been passed unsuccessfully and what is their errors are.

Why should be that?

pilot/pkg/features/pilot.go
ramaraochavali: That was the whole discussion we had in this PR. Even if user trusts it, how do you guarantee that xfcc is not spoofed?

pkg/security/security.go
ramaraochavali: @costinm ^^

pilot/pkg/features/pilot.go
hzxuzhonghu: If you config it, you trust this way, just like k8s request header authenticator. It leaves the decision to the user. 
In another word, this is a  weaker authentication compared to others

ramaraochavali: > If you config it, you trust this way, just like k8s request header authenticator. It leaves the decision to the user.

request header authenticator does not leave it to user https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy

_In order to prevent header spoofing, the authenticating proxy is required to present a valid client certificate to the API server for validation against the specified CA before the request headers are checked. WARNING: do not reuse a CA that is used in a different context unless you understand the risks and the mechanisms to protect the CA's usage._

It is almost similar here

costinm:
On Mon, Jun 27, 2022 at 1:28 AM Zhonghu Xu ***@***.***> wrote:

>
> In pilot/pkg/features/pilot.go
> <https://github.com/istio/istio/pull/39405#discussion_r907114879>:
>
> > @@ -403,6 +403,14 @@ var (
>  		"If enabled, pilot will authorize XDS clients, to ensure they are acting only as namespaces they have permissions for.",
>  	).Get()
>
> +	// TODO: Move this to proper API.
> +	TrustedGatewayCIDR = env.RegisterStringVar(
> +		"TRUSTED_GATEWAY_CIDR",
> +		"",
> +		"If set, any connections from gateway to Istiod with this CIDR range are treated as trusted for using authenication mechanisms like XFCC."+
> +			" This can only be used when the network where Istiod and the authenticating gateways are running is a trusted/secure network",
> +	).Get()
>
> cidr has no direct relationship with xfcc headers
>

The relation is that XFCC headers from the CIDR will be trusted/processed.
Not sure how direct the relation is - it is not exclusive, i.e. we may
trust XFCC from
workloads that authenticate with a mTLS and match a list of identities.
We'll likely have multiple ways to establish trust with an XFCC gateway -
CIDR is one of them.

ramaraochavali:
@costinm @hzxuzhonghu addressed comments. I still think we need to capture messages in AuthContext because we can remove the return value of each authenticator and can "truly" use context to capture the messages + final identity. Otherwise I do not see use of context at all.

pkg/security/security.go
```diff
@@ -351,12 +352,41 @@ type Caller struct {
...
+}
+
+// AuthContext carries the contextual information and is passed across authenticators.
+type AuthContext struct {
+	// RequestContext is the context from request.
+	RequestContext context.Context
+	// Messages contains list of messages that authenticator wants to record.
+	Messages map[string][]string
+	// Caller indicates the caller that made authentication request.
+	Caller *Caller
```
hzxuzhonghu, line 368: These are redundant as Authenticate returns the caller or error

pilot/pkg/features/pilot.go
hzxuzhonghu: I donot think cidr make much sense as in istio or k8s scenario, istiod within a cluster can not be accessed from outer clients if not exposed Loadbalancer type service or Ingress.

And also many cni support networkPolicy, which can limit the access from underlying network level.

pkg/security/security.go
ramaraochavali: We plan to remove this next PR

ramaraochavali: Only drive by authentication context. Return values will be removed.

ramaraochavali:
@costinm can you please review this if this makes sense - I want to conclude this PR as https://github.com/istio/istio/pull/39690 is trying to refactor it so that the merge can be smooth.

ramaraochavali:
@costinm Can you PTAL when you get chance?

ramaraochavali:
@costinm PTAL when you get chance

ramaraochavali:
@costinm gentle ping. Can you PTAL?

costinm (approved):
I think it's a good first step - it'll need additional iterations but it solves at least one use case
and doesn't hurt existing modes. 

Eventually we need to stop returning if an authenticator worked and let the others do their thing - so the gateway in front of Istiod can use mTLS to authenticate itself instead of CIDRs.

ramaraochavali:
@hzxuzhonghu Can you please see if your concerns are addressed?

pilot/pkg/xds/auth.go
howardjohn: > Any one wins. The identities are merged together.

This doesn't seem correct. Example: we have `admin` identity with `XFFC=low-priv`. Now the low-priv client has admin access if they want?

ramaraochavali: This is not true any more. XFCC is added at the end and will only authenticate based xfcc header if the client connection is part of trusted networks defined in env var or local host. There is no merging of identities. Those were old comments.

security/pkg/server/ca/server.go
```diff
@@ -165,7 +165,7 @@ func Authenticate(ctx context.Context, auth []security.Authenticator) *security.
 	// TODO: apply different authenticators in specific order / according to configuration.
 	var errMsg string
 	for id, authn := range auth {
-		u, err := authn.Authenticate(ctx)
+		u, err := authn.Authenticate(security.NewAuthContext(ctx))
```
howardjohn, line 168: this differs from the XDS logic in ways that seem like they would break things. In the XDS one we share an AuthContext, here we create one per authenticator.

Can/should we refactor them to use a shared impl?

pilot/pkg/xds/auth.go
```diff
@@ -56,17 +56,20 @@ func (s *DiscoveryServer) authenticate(ctx context.Context) ([]string, error) {
...
+
+	authContext := security.NewAuthContext(ctx)
+
 	for _, authn := range s.Authenticators {
-		u, err := authn.Authenticate(ctx)
+		// TODO: Refactor to remove the return value of Authencicate. Just rely on AuthContext.
+		u, err := authn.Authenticate(authContext)
 		// If one authenticator passes, return
 		if u != nil && u.Identities != nil && err == nil {
+			authContext.Caller = u
```
howardjohn, line 67: Caller is also not used. If we also make the `Messages` change I suggested, we no longer need `AuthContext` entirely anymore

pkg/security/security.go
```diff
@@ -351,12 +352,41 @@ type Caller struct {
 	Identities []string
 }
 
+// AuthResult is the result of authentication action.
+type AuthResult struct {
```
hzxuzhonghu, line 356: confusing, after first look i thought it is the return value of `Authenticate()`

pkg/security/security.go
```diff
@@ -351,12 +352,41 @@ type Caller struct {
...
+	Identities []string
+	Messages   []string
+}
+
+// AuthContext carries the contextual information and is passed across authenticators.
+type AuthContext struct {
+	// RequestContext is the context from request.
+	RequestContext context.Context
+	// Messages contains list of messages that authenticator wants to record.
+	Messages map[string][]string
```
hzxuzhonghu, line 366: What is this message used for?

pkg/security/security.go
hzxuzhonghu: Prefer distinguish input and output

security/pkg/server/ca/authenticate/xfcc_authenticator_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}{
+		{
+			name:       "No xfcc header",
+			xfccHeader: "",
+			caller:     nil,
+		},
+		{
+			name: "Xfcc Header",
+			// nolint lll
+			xfccHeader: `Hash=meshclient;Subject="";URI=spiffe://mesh.example.com/ns/otherns/sa/othersa`,
```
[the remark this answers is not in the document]
ramaraochavali: Added few more. I will add more later

security/pkg/server/ca/server.go
ramaraochavali: Added AuthenticationManager take orchestrates Authenticators. PTAL.

ramaraochavali:
@howardjohn @hzxuzhonghu Fixed all comments. PTAL

pkg/security/security.go
```diff
@@ -352,39 +355,44 @@ type Caller struct {
...
+func (am *AuthenticationManager) Authenticate(ctx context.Context) *Caller {
+	var errMsg string
+	for _, authn := range am.Authenticators {
+		u, err := authn.Authenticate(ctx)
+		if u != nil && u.Identities != nil && err == nil {
+			securityLog.Debugf("Authentication successful through auth source %v", u.AuthSource)
+			return u
+		}
+		am.authFailMsgs = append(am.authFailMsgs, fmt.Sprintf("Authenticator %s: %v", authn.AuthenticatorType(), err))
+		securityLog.Warnf("Authentication failed for %v: %s", getConnectionAddress(ctx), errMsg)
```
hzxuzhonghu, line 381: Why does it logging while record the `authFailMsgs`

security/pkg/server/ca/authenticate/xfcc_authenticator.go
```diff
@@ -43,22 +44,20 @@ func (xff XfccAuthenticator) AuthenticatorType() string {
...
 // Authenticate extracts identities from Xfcc Header.
-func (xff XfccAuthenticator) Authenticate(ctx *security.AuthContext) (*security.Caller, error) {
-	authResult := security.AuthResult{}
-	peerInfo, _ := peer.FromContext(ctx.RequestContext)
+func (xff XfccAuthenticator) Authenticate(ctx context.Context) (*security.Caller, error) {
+	peerInfo, _ := peer.FromContext(ctx)
 	// First check if client is trusted client so that we can "trust" the Xfcc Header.
-	if !isTrustedAddress(peerInfo.Addr.String()) {
+	if !isTrustedAddress(peerInfo.Addr.String(), features.TrustedGatewayCIDR) {
 		message := fmt.Sprintf("caller from %s is not in the trusted network. XfccAuthenticator can not be used", peerInfo.Addr.String())
```
hzxuzhonghu, line 51: nit: return fmt.errorf directly

pkg/security/security.go
ramaraochavali: will remove not needed

ramaraochavali:
@hzxuzhonghu addressed comments. PTAL

hzxuzhonghu (commented):
I have a last question: 

in xds authorize, we do check the client identities, which requires spiffe format, so how do you pass this check?

security/pkg/server/ca/authenticate/xfcc_authenticator.go
```diff
@@ -0,0 +1,118 @@
+// Copyright 2017 Istio Authors
```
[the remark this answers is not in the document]
ramaraochavali: :-). I do not know where I copied this from otehr file. Some files still seem to have it. I will clean them later

ramaraochavali:
> in xds authorize, we do check the client identities, which requires spiffe format, so how do you pass this check?

I do not think it is different here. If the identities have spiffeID - one of the identities should be from proxyNamespace.

ramaraochavali:
@hzxuzhonghu fixed comments. For other files that have 2017 in copyright - will fix in a separate PR

hzxuzhonghu:
>>in xds authorize, we do check the client identities, which requires spiffe format, so how do you pass this check?

>I do not think it is different here. If the identities have spiffeID - one of the identities should be from proxyNamespace.

It is a must that at least one identity is spiffeID, so curious how you signed the certs to proxy, then it can make xfcc authenticator work here.

ramaraochavali:
> It is a must that at least one identity is spiffeID, so curious how you signed the certs to proxy, then it can make xfcc authenticator work here.

Yes. But if atleast one identity is not spiffe identity, the identity check will fail. Our certs will always have one spiffeID. Is it what you are asking?

ramaraochavali:
@hzxuzhonghu @howardjohn any more concerns with this?

howardjohn (approved):

costinm:
While I think we will need the CIDR authenticator in the future - for now
it may be best to have the checks in the XFCC authenticator, not as a
separate one.

For 'CIDR' authentication - will be important for the 'trusted network'
case ( where the network has ipsec / etc ) - but with additional code to
lookup the
caller by IP ( based on the discovery info ) and populate the namespace/KSA
from the calling Pod.

In the case of XFCC - the caller IP can be an external load balancer (not a
pod) - we can reuse some code but it's a different kind of CIDR checking,
i.e. against a list of 'trusted auth proxies' CIDR.

costinm:
And we should support ( via additional PRs) the stronger model used by K8S,
i.e. the auth proxy having an mTLS connection to
Istiod and authenticating strongly.

But I think for 'localhost' or if the network is already secure  ( IPSec,
Wireguard, etc can guarantee encryption and authenticate
the IP address ) - CIDR mode can be acceptable too, in particular for
interop/compatibility purposes.

costinm:
On Wed, Jun 22, 2022 at 8:17 AM Rama Chavali ***@***.***>
wrote:

> Ok. Looks like the proposal is to
>
>    - merge the current cidr_authenticator logic in to xfcc authenticator
>    - Drive the trusted ranges from network? Do we need special flag or we
>    have a network with name as "gateway" or we can configure that in the pilot
>    flag like TRUSTED_ NETWORK instead of configuring CIDR ranges
>    directly. Trying to see if we can avoid API change
>
> I think we need to add a flag or label or some way to indicate the
(sub)network is trusted/secure at IP level, and another one to indicate
it is dedicated for gateways, i.e. all requests from that network will be
from a gateway that can use XFCC.

Seems a better option than defining a bunch of env variables or another
network definition config. It is currently a separate config map,
reloaded on change (AFAIK) - at some point we could promote it to a CRD,
but better wait for it to 'settle'.  I think the only broadly used
CRD defining CIDRs is NetworkPolicy - it is unfortunate the syntax is
slightly different, if we later promote networks to a CRD we may
converge them a bit.

costinm:
On Thu, Jun 16, 2022 at 11:09 PM Rama Chavali ***@***.***>
wrote:

>
> In pilot/pkg/bootstrap/server.go
> <https://github.com/istio/istio/pull/39405#discussion_r899792457>:
>
> > @@ -308,6 +308,7 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
>  	// is used as the authentication result.
>  	authenticators := []security.Authenticator{
>  		&authenticate.ClientCertAuthenticator{},
> +		&authenticate.XfccAuthenticator{},
>
> The use case here is envoy front ending Istiod and it takes care of TLS
> termination. So the connection is from localhost. I think if the peer
> connection is from localhost and Xfcc Authenticator is enabled, we can rely
> on xfcc authenticator - Do you agree or any concerns?
>
That limits us to envoy frontend co-located as sidecar with Istiod. It's
useful - but I think the more common/interesting use case
is an envoy acting as gateway ( possibly the main ingress gateway, or the
east/west gateway ).

costinm:
We should return one identity ( swapped identity - if peer identity is a
trusted gateway + XFCC, or the peer identity from cert or JWT ). The IP is
not an identity.

However I think the context should preserve information about peer IP and
identity - as a good auth API. Maybe this will also help having better
messages instead
of the multi-error we have when authenticating, but also if we want to log
for audit purposes ( in particular on the cert signing endpoint )

PR #39156, opened 2022-05-26
enabled security tests in multicluster env

aryan16:
**Enabled Security tests in mutlicluster env**
TestIncludes - 
1) mtlsk8sca
2) https_jwt
3) sds_ingress_k8sca
4) file_based_tls_origination
5) sds_ingress/quic

PR #39406, opened 2022-06-10
recognize istiodless remote clusters

### LINKED ISSUE #39257: multi-cluster: istiod repeatedly fails to patch configmap
```
2022-06-01T06:17:40.048065Z	error	klog	error retrieving resource lock istio-system/istio-namespace-controller-election: configmaps "istio-namespace-controller-election" is forbidden: User "system:serviceaccount:istio-system:istio-reader-service-account" cannot get resource "configmaps" in API group "" in the namespace "istio-system"
```

We enabled "external istiod" flag by default, so we'll try to win leader locks and patch webhooks/etc:

[#39189 ](https://github.com/istio/istio/pull/38283)

But the permissions are still conditional:

https://github.com/istio/istio/blob/12f5ad29c4b83898077902f3e63826d2dd8d7d6f/manifests/charts/istiod-remote/templates/reader-clusterrole.yaml#L45-L55

We should either:
* Always give permission, rely on leader elect to prevent the wrong istiod doing the patching
* Conditionally try to participate in this patching based on some metadata in the remote-secret or a label on the istio-system namespace (some per-cluster metadata that indicates primary, remote, istiodless-remote, etc.)
* Auto-detect the cluster-type somehow? 
* Fail closed on forbidden

frankbu:

Fixes #39257


manifests/charts/istio-control/istio-discovery/templates/mutatingwebhook.yaml
```diff
@@ -44,6 +44,9 @@ metadata:
     operator.istio.io/component: "Pilot"
     app: sidecar-injector
     release: {{ .Release.Name }}
+{{- if .Values.global.externalIstiod }}
+    externalIstiod: "true"
```
hzxuzhonghu, line 48: what's the root cause, why need an additional label?

frankbu: The problem is that istiod doesn't have permission to acquire the leader lock on a non-istiodless remote cluster, so we need a way that istiod can check if it is an istiodless cluster before trying. Without checking the label, istiod will fail to acquire the lock when it doesn't have permission, which would be fine except that it generates error messages in the log. This PR is just needed to prevent the error messages from appearing in the log when istiod is connected to a non-istiodless remote cluster.

frankbu:
@stevenctl @howardjohn PTAL

manifests/charts/istio-control/istio-discovery/templates/mutatingwebhook.yaml
hzxuzhonghu: I am confused since externalIstiod is always true now. why does istiod lack the permission?

https://github.com/istio/istio/blob/12f5ad29c4b83898077902f3e63826d2dd8d7d6f/manifests/charts/istiod-remote/templates/reader-clusterrole.yaml#L45-L55

frankbu: Ah, there are two different externalIstiod configuration parameters.

1. `values.pilot.env.EXTERNAL_ISTIOD` this is used to configure istiod running on an external or primary cluster. It controls whether or not istiod should attempt to manage other remote clusters. This is now true by default.
2. `values.global.externalIstiod` this is used to configure a remote cluster. True for istiodless remotes, and false by default.

It would have been clearer if the two weren't using the same name, but that was done a long time ago, so we're struck with it.

stevenctl: Maybe the label can be the name(s) of primary cluster that should attempt to get the lock. The topology matters, and any EXTERNAL_ISTIOD cluster can't just acquire the lock for any externalIstiod cluster.

tests/integration/security/filebased_tls_origination/main_test.go
```diff
@@ -35,7 +35,7 @@ func TestMain(m *testing.M) {
 		NewSuite(m).
 		Label(label.CustomSetup).
 		Label("CustomSetup").
-		SkipExternalControlPlaneTopology().
+		RequireMultiPrimary(). // SkipExternalControlPlaneTopology(). TODO this test fails when running with istiodless remotes
```
frankbu, line 38: This is reversing a recent change in #39156. Seems it doesn't work after switching to istiodless remotes?
^^^ @aryan16

manifests/charts/istio-control/istio-discovery/templates/mutatingwebhook.yaml
frankbu: Good idea. I added an annotation containing a regex of istiod clusters that should try.

howardjohn: I don't feel comfortable matching cluster names with regex.. that feels like a recipe for headaches. Do we need multiple clusters matching?

frankbu: one cluster is enough but for HA I think we will want to allow multiple. If using regex for this is what concerns you, we could make it a list of cluster names instead. Would that be better?

tests/integration/security/filebased_tls_origination/main_test.go
aryan16: That's strange, let's merge this PR and I will try to take a closer look on this after that.

manifests/charts/istio-control/istio-discovery/templates/mutatingwebhook.yaml
frankbu: @howardjohn switched to one or more comma-separated cluster names

manifests/charts/istiod-remote/templates/mutatingwebhook.yaml
```diff
@@ -44,6 +44,11 @@ metadata:
     operator.istio.io/component: "Pilot"
     app: sidecar-injector
     release: {{ .Release.Name }}
+{{- if .Values.global.externalIstiod }}
+    externalIstiod: "true"
+  annotations:
+    istiodClusterIDs: {{ .Values.global.istiodClusterIDs | default "'*'" }}
```
frankbu, line 50: @stevenctl Just to explain the design here, although I'm open to tweaking this (but preferably in a followup PR)

`istiodClusterIDs` default value:
1. for non-istiodless remotes it defaults to `""`, no istiod should try.
1. for istiodless remotes it defaults to `*`, meaning any istiod should try (permissions and other location preference will still apply in leader election). However users should typically override to specify the expected control plane cluster(s). We do override it in the test framework, and I will update the docs as well.

manifests/charts/istio-control/istio-discovery/templates/mutatingwebhook.yaml
howardjohn: Yeah was specifically regex. Thanks!

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -314,6 +319,40 @@ func (m *Multicluster) initializeCluster(cluster *multicluster.Cluster, kubeRegi
 	return nil
 }
 
+const (
+	externalIstiodLabel     = "externalIstiod"
+	istiodClusterAnnotation = "istiodClusterIDs"
```
stevenctl, line 324: 1. Do we need both annotations? Can we do a label selector with no `=true` to select things that have the label present? 
2. We should iterate on the name to something like `topology.istio.io/primaryCluster`
3. I'm not sure the mutating webhook is the place we want to store this important topology info in the long term, but it's fine for now.

stevenctl (approved):
LGTM but lets bikeshed that label a bit

manifests/charts/istio-control/istio-discovery/values.yaml
```diff
@@ -394,9 +394,13 @@ global:
   # discovery charts are enabled.
   externalIstiod: false
 
-  #  Configure a remote cluster as the config cluster for an external istiod.
+  # Configure a remote cluster as the config cluster for an external istiod.
   configCluster: false
 
+  # Comma-separated list of clusters (or * for any) with an istiod that should
+  # attempt leader election for a remote cluster.
+  istiodClusterIDs: ""
```
hzxuzhonghu, line 402: This is hard to know in advance, any cluster can join the mesh in the future

hzxuzhonghu: NVM, IIRC, it is to specify the cluster which should control remote cluster

pilot/pkg/serviceregistry/kube/controller/multicluster.go
frankbu: 1. I've been wondering if it's really needed. It's really only used to optimize the selector for the query. It is, however, also an indicator that a cluster is istiodless, so maybe good to have in the future.
2. I actually started with `primaryCluster` but changed it because I thought it was confusing for external control planes, where the config cluster is the cluster ID that istiod uses, not a primary cluster.
3. The mutating webhook is certainly a convenient place to add the label, since it's a template in the remote cluster install. Putting on the namesapce, for example, is a little harder, since it would have to be done programmatically, or manually?

frankbu: Or ... maybe we don't need to identify an istiodless remote cluster. All we need is one label that identifies the clusters that should attempt remote leader election on any cluster (primary or remote, istiodless or not)?

pkg/test/framework/components/istio/operator.go
```diff
@@ -865,17 +864,26 @@ func deployCACerts(workDir string, env *kube.Environment, cfg Config) error {
 		if env.IsMultinetwork() {
 			nsLabels = map[string]string{label.TopologyNetwork.Name: c.NetworkName()}
 		}
+		var nsAnnotations map[string]string
+		if c.IsRemote() {
+			const istiodClusterAnnotation = "topology.istio.io/istiodClusters" // TODO proper API annotation.TopologyIstiodClusters.Name
```
stevenctl, line 869:
```suggestion
			const istiodClusterAnnotation = "topology.istio.io/primaryCluster" // TODO proper API annotation.TopologyIstiodClusters.Name
```

stevenctl (approved):
LGTM, although I think the annotation should be the word "primaryCluster" or "controlPlaneCluster" to more closely match our [deployment models](https://istio.io/latest/docs/ops/deployment/deployment-models/) doc. 

Also, in the future we might refactor this to handle changes to this value. (We do this for multi-network already) 

https://github.com/istio/istio/blob/d5a828ebb79c69570cc80c1fead889fbec008d22/pilot/pkg/serviceregistry/kube/controller/controller.go#L1202

frankbu:
> I think the annotation should be the word "primaryCluster" or "controlPlaneCluster" to more closely match our [deployment models](https://istio.io/latest/docs/ops/deployment/deployment-models/) doc.

I think "primaryCluster" would be confusing since there is no primary cluster in the mesh when we have an external control plane. I know that the test framework calls it a primary, which is a little confusing too, but in deployment-models doc an external control plane is not a primary cluster. In fact, the cluster ID of the the external control plane is the config cluster ID, so that makes it even more confusing if we name the annoation "primaryCluster".

That said, I can see "controlPlaneCluster" is probably a better name than "istiodCluster", since it's higher level, not tied to our implementation name, and as you say more closely matches or deploynment-model doc.

pilot/pkg/serviceregistry/kube/controller/multicluster.go
```diff
@@ -314,6 +319,31 @@ func (m *Multicluster) initializeCluster(cluster *multicluster.Cluster, kubeRegi
...
 }
 
+// Comma-separated list of clusters (or * for any) with an istiod that should
+// attempt leader election for a remote cluster.
+const istiodClusterAnnotation = "topology.istio.io/controlPlaneClusters" // TODO make proper API annotation.TopologyControlPlaneClusters.Name
+
+// checkShouldLead returns true if the caller should attempt leader election for a remote cluster.
+func (m *Multicluster) checkShouldLead(client kubelib.Client, systemNamespace string) bool {
+	if features.ExternalIstiod {
+		namespace, err := client.Kube().CoreV1().Namespaces().Get(context.TODO(), systemNamespace, metav1.GetOptions{})
```
howardjohn, line 329: this means that a transient error will permanently break things, and if a user adds the label 1s late it will also never recover. Seems like we should be using a namespace informer?

hzxuzhonghu: It will make things a little bit complicated

frankbu: @howardjohn Added a TODO and made it fail open in the error case for now, so this PR is  NOOP in that case. PTAL

howardjohn: This doesn't feel like a NOP to me. It seems like very unsafe behavior. Failing open is not safe either, since now we will lead when we are not supposed to

howardjohn (approved):
Approving security change on the basis that @stevenctl approved the main changes

jacob-delgado (approved):

pilot/pkg/serviceregistry/kube/controller/multicluster.go
dwj300: @frankbu, any updates on the followup to use a namespace informer?

Issue #39368, opened 2022-06-09
Istio uses significantly high memory 


ramaraochavali:
### Bug Description

We upgraded to the recent Istio version and immediately we saw a jump of Istiod memory usage. We have increased the memory to 14G and still it is not sufficient. Ultimately it gets OOM Killed.

### Version

```prose
Master
```

ramaraochavali:
Initial Analysis turns out that it is because of RDS Cache  enabled by`PILOT_ENABLE_RDS_CACHE`.

Here are some pprof outputs 

<img width="848" alt="Screen Shot 2022-06-08 at 1 13 04 PM" src="https://user-images.githubusercontent.com/17204904/172804864-60f04299-f1ef-441f-b7d4-9adb6c9932e8.png">
<img width="986" alt="Screen Shot 2022-06-08 at 1 13 47 PM" src="https://user-images.githubusercontent.com/17204904/172804887-cfea952c-5235-441f-b0e3-f99a25d76cb1.png">
<img width="1172" alt="Screen Shot 2022-06-08 at 1 16 43 PM" src="https://user-images.githubusercontent.com/17204904/172804892-b065138a-622a-4893-84be-acf17d69c651.png">

hzxuzhonghu:
Config.Spec has a reference to underlying vs

hzxuzhonghu:
And actually lru route cache key is only md5 checksum, not sure how you profile. And from your profile, I could not see how much memory totally istiod costs.

ramaraochavali:
> Config.Spec has a reference to underlying vs

Yes

> And actually lru route cache key is only md5 checksum, not sure how you profile. And from your profile, I could not see how much memory totally istiod costs.

@deveshkandpal1224 can you please share complete profile before/after RDS cache and the memory usage?

howardjohn:
can you upload the actual profile targz as well when you get the new one

deveshkandpal1224:
@howardjohn @hzxuzhonghu - sorry for the delay, attached profiles as requested. PTAL - let me know if there is anything else that you need. Thanks!

low-memory -> 

[pprof.pilot-low-mem-discovery.alloc_objects.alloc_space.inuse_objects.inuse_space.005.pb.gz](https://github.com/istio/istio/files/8874457/pprof.pilot-low-mem-discovery.alloc_objects.alloc_space.inuse_objects.inuse_space.005.pb.gz)


high memory -> 

[...]

PR #39412, opened 2022-06-10
Optimize cache key generation

sschepens:
Optimize various cache key generations by directly appending to hash and not allocating a slice.

Related to https://github.com/istio/istio/issues/39368

### DIFF
--- a/pilot/pkg/xds/endpoint_builder.go
+++ b/pilot/pkg/xds/endpoint_builder.go
@@ -39,6 +39,11 @@ import (
 	"istio.io/istio/pkg/network"
 )
 
+var (
+	Separator = []byte{'~'}
+	Slash     = []byte{'/'}
+)
+
 // Return the tunnel type for this endpoint builder. If the endpoint builder builds h2tunnel, the final endpoint
 // collection includes only the endpoints which support H2 tunnel and the non-tunnel endpoints. The latter case is to
 // support multi-cluster service.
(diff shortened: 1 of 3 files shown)


pilot/pkg/xds/endpoint_builder.go
```diff
@@ -124,30 +124,31 @@ func (b EndpointBuilder) DestinationRule() *networkingapi.DestinationRule {
...
+	hash.Write([]byte(b.tunnelType.ToString()))
+
 	if b.push != nil && b.push.AuthnPolicies != nil {
-		params = append(params, b.push.AuthnPolicies.GetVersion())
+		hash.Write([]byte(b.push.AuthnPolicies.GetVersion()))
 	}
 	if b.destinationRule != nil {
-		params = append(params, b.destinationRule.Name+"/"+b.destinationRule.Namespace)
+		hash.Write([]byte(b.destinationRule.Name))
+		hash.Write([]byte{'/'})
```
howardjohn, line 140: nit: not sure we really need the `/`. I guess we didn't really before either.

In the original form long ago we sort of did, before we did a hash.

sschepens: it could theoretically reduce chances of collision, when two different names and namespaces have the same result when appending, though I don't know how often this could happen.
for example: `name + ns = nam + ens`

let me know if I should remove them or keep them

sschepens:
@howardjohn as I commented on the Issue:

There is still some more optimizations that could be done, every hash Write is copying the string to convert it to a []byte, this copying could be avoided by using some unsafe code which is done in many libraries. Should we do this as well? is this considered a performance-critical path?

pilot/pkg/xds/endpoint_builder.go
howardjohn: ah good point.

I wonder if we need divider for everything then. In the original one we did `strings.Join(params, "~")` to avoid. If write(ab) is the same as `write(a); write(b)` then I think its actually unsafe the way we do things now (with and without this change)?

howardjohn:
It is fairly perf critical but I would like to see benchmarks before + after before justifying use of `unsafe`. I am not sure how much cost the conversion is

pilot/pkg/xds/endpoint_builder.go
sschepens: Yup, `write(ab)` is the same as `write(a); write(b)`, i tested it, if we previously joined with `~` then we should write `~` before every other write

sschepens:
> It is fairly perf critical but I would like to see benchmarks before + after before justifying use of unsafe. I am not sure how much cost the conversion is

The cost is basically only the copying involved, which adds to memory and gc pressure probably, I don't think it's easily measurable. It also probably depends on the amount and size of strings we're copying.

I made a simple benchmark on my laptop:
```
BenchmarkCopy-8     	20283340	        59.20 ns/op	      32 B/op	       1 allocs/op
BenchmarkNoCopy-8   	24001119	        49.01 ns/op	       0 B/op	       0 allocs/op
```

There is effectively one copy being made and not doing it is apparently improving close to 20% the time.

In our code, we're definitely doing a lot of writes, probably not really big strings, but if we sum the cost of copying all those strings, it could add up to a lot. Maybe thats the culprit of the amount of memory that was seen on the profiles attached in the issue.

howardjohn:
I would opt for doing this without unsafe, and see how this+the other PR I think we will have for the LRU fix go. If we see it would bring real world benefits (and its actually safe, despite using `unsafe`) we can consider it?

pilot/pkg/xds/endpoint_builder.go
sschepens: I was thinking a little bit more about this, the chances of collision of all the entities we use in the cache key is probably very low, right? so maybe the dividers are not really that necessary?

On another hand, adding them is very easy, and could prevent some headaches in case we do get collisions.

ramaraochavali: If there is no cost/ very low cost of adding them - I would just go ahead and add

hzxuzhonghu:
Make sense, and i also wonder if we could only call hash.Write once

sschepens:
> Make sense, and i also wonder if we could only call hash.Write once

@hzxuzhonghu i don't think we can `hash.Write` only once, because it takes a `[]byte` and every string we have is a `[]byte` so we would have to append all of them together for a full write.

ramaraochavali:
Thank you. LGTM. Will wait for @howardjohn review

sschepens:
@howardjohn should I add a release note?

PR #39433, opened 2022-06-13
Optimize: make use of string pointer map to reduce mem cost of xds keys

hzxuzhonghu:

This reduces memory for case when n configs influence one xds key.

Inner struct is 👍 

C(0)->key0(copied)
C(1)->key1(copied)
C(n-1)->key n-1(copied)

A string struct holds 16 bytes, and here the total mem is n * 16 bytes

With the patch, the total mem will be n * 8 

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/xds_cache.go
+++ b/pilot/pkg/model/xds_cache.go
@@ -81,19 +81,19 @@ func size(cs int) {
 	}
 }
 
-func indexConfig(configIndex map[ConfigKey]sets.Set, k string, entry XdsCacheEntry) {
+func indexConfig(configIndex map[ConfigKey]sets.StringPointerSet, k *string, entry XdsCacheEntry) {
 	for _, cfg := range entry.DependentConfigs() {
 		if configIndex[cfg] == nil {
-			configIndex[cfg] = sets.New()
+			configIndex[cfg] = sets.NewStringPointerSet()
 		}
 		configIndex[cfg].Insert(k)
 	}
 }
 
-func indexType(typeIndex map[config.GroupVersionKind]sets.Set, k string, entry XdsCacheEntry) {
+func indexType(typeIndex map[config.GroupVersionKind]sets.StringPointerSet, k *string, entry XdsCacheEntry) {
 	for _, t := range entry.DependentTypes() {
 		if typeIndex[t] == nil {
-			typeIndex[t] = sets.New()
+			typeIndex[t] = sets.NewStringPointerSet()
 		}
 		typeIndex[t].Insert(k)
 	}
@@ -144,8 +144,8 @@ func NewXdsCache() XdsCache {
 	return &lruCache{
 		enableAssertions: features.EnableUnsafeAssertions,
 		store:            newLru(),
-		configIndex:      map[ConfigKey]sets.Set{},
-		typesIndex:       map[config.GroupVersionKind]sets.Set{},
+		configIndex:      map[ConfigKey]sets.StringPointerSet{},
+		typesIndex:       map[config.GroupVersionKind]sets.StringPointerSet{},
 	}
 }
 
@@ -154,8 +154,8 @@ func NewLenientXdsCache() XdsCache {
 	return &lruCache{
 		enableAssertions: false,
 		store:            newLru(),
-		configIndex:      map[ConfigKey]sets.Set{},
-		typesIndex:       map[config.GroupVersionKind]sets.Set{},
+		configIndex:      map[ConfigKey]sets.StringPointerSet{},
+		typesIndex:       map[config.GroupVersionKind]sets.StringPointerSet{},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


ramaraochavali:
LGTM. Can you please add tests for the new struct?

Also sets.IntSet, sets.StringPointerSet seems to repeat `Set` - But I do not have better naming suggestion though

sschepens:
is this really the underlying issue? do you have any real-world tests showing the improvements?

Maybe I was wrong, but i was under the impression that we're leaking these Sets, and that's why we see such huge memory allocated.

hzxuzhonghu:
>Maybe I was wrong, but i was under the impression that we're leaking these Sets, and that's why we see such huge memory allocated.

Nope, we do delete in Clear and Clearall. Golang map capacity is unknown and could be out of control

hzxuzhonghu:
>Also sets.IntSet, sets.StringPointerSet seems to repeat Set - But I do not have better naming suggestion though

I think we can rename all later

sschepens:
> Evict is a different abnormal case, we do not cover it. I could hardly imagine there are more than 60000 xds keys in a mesh.

Evict doesn't seem to be abnormal case, that probably why an LRU was chosen, because we know the limit can be exceeded.

If more than 60000 keys is the limit and exceeding it will cause issues then that should be at least documented.

On another note, I can totally see some people having more than 60000 xds keys on big companies.

hzxuzhonghu:
However, this is a totally different fix, could mitigate the memory by reducing 50 %, if you read above desc

ramaraochavali:
>  I could hardly imagine there are more than 60000 xds keys in a mesh.

I agree with @sschepens for bigger clusters this is not enough as I shared in the original issue - there will be evictions (not abnormal)

ramaraochavali (approved):

sschepens:
It would be nice to have some metrics to compare this PR. We're basically trading memory for CPU, because we will have to dereference each string pointer and i'm still not sure how much memory we will actually be saving.

howardjohn (changes requested):
Lets not merge without comparing with https://github.com/istio/istio/pull/39443

sschepens:
@howardjohn it would be nice to have benchmarks on `Clear` method which is most likely to be impacted by this change, since it's there that we would have to dereference the pointers

pilot/pkg/model/xds_cache.go
```diff
@@ -164,10 +164,14 @@ type lruCache struct {
...
 	// It is refreshed when Clear or ClearAll are called
-	token       CacheToken
-	mu          sync.RWMutex
-	configIndex map[ConfigKey]sets.Set
-	typesIndex  map[config.GroupVersionKind]sets.Set
+	token CacheToken
+	mu    sync.RWMutex
+	// configIndex and typesIndex stores the xds key pointer,
+	// because different configs or different kinds of resources can be part of a same xds key.
+	// Say route `80` can be influenced by hundreds of services with port 80.
```
howardjohn, line 171: Since we take md5 hash, its always constant size?

howardjohn (commented):
BenchmarkCache doesn't show much difference

sschepens:
> benchmarking Clear sounds tricky since once we Clear it we would need to fully rebuild the cache... then its hard to measure just Clear since building the cache will overwhelm things?

Yes, it can be tricky, maybe you can build a fully populated cache up to max size outside of the benchmark and also keep a list of all de `DependentConfigs`, then inside the benchmark call `Clear()` for every DependentConfig. 🤔

pilot/pkg/model/xds_cache.go
sschepens: the string should be constant size, yes, because we use take the md5 hash.

what @hzxuzhonghu was trying to do was save a half the memory by using a pointer, a string variable uses 16bytes (pointer to byte array and int for the length), and a pointer to a string is only 8 bytes.

the set stores the keys, so using a pointer as key would mean that those keys use half the amount of memory.

What we don't know is how much is 8bytes per string saving compared to the amount of bytes the rest of the cache is using.
Also the string objects still remain in memory because we reference them, so in some cases we may actually be wasting more memory? (pointer+string in memory). Though we may add the same string to multiple sets, so in that case, string pointers could save some memory.

howardjohn: We may save a bit but I am not really convinced its worth this, and also agree same concern that we still have to store the string. I think the underlying issue in the main issue is pretty clearly a leak on the configindex - micro-optimizations like this would not solve this case and don't seem to help much in general based on the benchmarks we do have. If there is some data that shows otherwise that would help

hzxuzhonghu: It is worth, you can add hundreds of services or other dependencies in your benchmark test to see the diff.

howardjohn: the test does that afaik. Feel free to tweak the benchmark though

hzxuzhonghu: Nope， i can add that

pilot/pkg/xds/bench_test.go
```diff
@@ -560,22 +560,30 @@ func BenchmarkPushRequest(b *testing.B) {
 
 func makeCacheKey(n int) model.XdsCacheEntry {
 	ns := strconv.Itoa(n)
+
+	// 100 services
+	services := make([]*model.Service, 0, 100)
+	// 100 destinationrules
+	drs := make([]*config.Config, 0, 100)
+	for i := 0; i < 100; i++ {
```
hzxuzhonghu, line 568: @howardjohn If i have 100 services here,  we need to store
 svc1->key
 svc2->key 
...

howardjohn (approved):
Thanks for the benchmark improvement!

PR #39435, opened 2022-06-13
Clear index on cache evict

sschepens:
Allow XdsCache to clean indexes of evicted elements.

Related to https://github.com/istio/istio/issues/39368


pilot/pkg/model/xds_cache.go
```diff
@@ -245,8 +286,10 @@ func (l *lruCache) Add(entry XdsCacheEntry, pushReq *PushRequest, value *discove
 }
 
 type cacheValue struct {
-	value *discovery.Resource
-	token CacheToken
+	value            *discovery.Resource
+	token            CacheToken
+	dependentConfigs []ConfigKey
```
howardjohn, line 291: I think we don't set these when we write

pilot/pkg/model/xds_cache.go
```diff
@@ -141,22 +165,26 @@ type XdsCache interface {
...
 // NewXdsCache returns an instance of a cache.
 func NewXdsCache() XdsCache {
-	return &lruCache{
+	cache := &lruCache{
 		enableAssertions: features.EnableUnsafeAssertions,
-		store:            newLru(),
 		configIndex:      map[ConfigKey]sets.Set{},
 		typesIndex:       map[config.GroupVersionKind]sets.Set{},
 	}
+	cache.store = newLru(cache)
```
hzxuzhonghu, line 173: This is not good, you can only pass the evict handler

pilot/pkg/model/xds_cache.go
```diff
@@ -236,7 +277,7 @@ func (l *lruCache) Add(entry XdsCacheEntry, pushReq *PushRequest, value *discove
 		return
 	}
 
-	toWrite := cacheValue{value: value, token: token}
+	toWrite := cacheValue{value: value, token: token, dependentConfigs: entry.DependentConfigs(), dependentTypes: entry.DependentTypes()}
```
hzxuzhonghu, line 280: This will cause more mem cost

hzxuzhonghu (changes requested):

pilot/pkg/model/xds_cache.go
sschepens: it will occupy more memory than previously, yes, but i will still be much less than today, since currently the cache appends indefinitely to the indexes and never deletes from them, causing a memory leak.

Another thing is, this items are already used as keys for the `configIndex` and `typesIndex` so they are actually already kept in memory by those maps, so will we really add that much memory?

hzxuzhonghu: >but i will still be much less than today, since currently the cache appends indefinitely to the indexes and never deletes from them, causing a memory leak.

I do not think so in case there are less than 60000 xds keys

hzxuzhonghu: >Another thing is, this items are already used as keys for the configIndex and typesIndex so they are actually already kept in memory by those maps, so will we really add that much memory?


The struct is map[configkey]map{xds key}struct{}, it is totally different from `entry.DependentConfigs()`, so I think only little part of mem is reused. entry.DependentConfigs() returns a slice, each time it is reconstructed.

sschepens:
@howardjohn can you ran BenchmarkCache with this PR see the performance/memory hit we get?

howardjohn (approved):
This is great, thanks!!

howardjohn:
@hzxuzhonghu ptal

pkg/util/sets/string.go
```diff
@@ -44,8 +44,14 @@ func (s Set) InsertAll(items ...string) Set {
...
 
+// Delete removes an item from the set.
+func (s Set) Delete(item string) Set {
+	delete(s, item)
+	return s
+}
+
 // Delete removes items from the set.
-func (s Set) Delete(items ...string) Set {
+func (s Set) DeleteAll(items ...string) Set {
```
hzxuzhonghu, line 54: why change this

hzxuzhonghu (commented):
Let's wait after https://github.com/istio/istio/pull/39433/files/a194bc4b88f13c8bbc1c8cf3ce77d1fbd21f81f1..e59698694f8959f04adf08101d664a5ac811230c#diff-56363aa8983bcc9977c69541c3992f981e7cdbca5b1de64caa7f8ebf7dc52375R561 is in to see how much additional memory cost added

pkg/util/sets/string.go
sschepens: Most operations delete a single key, it, so it's better to provide a specialized version. It I
is also consistent with the other methods Insert/InsertAll.

sschepens:
@howardjohn @hzxuzhonghu the pointer set PR https://github.com/istio/istio/pull/39433 breaks this one, since we are now storing pointers, we cannot delete strings from the set because we don't keep the original string anywhere, the key stored in the LRU is another struct, and taking a pointer from that gives a different pointer that is not in the set.

I would have to change the LRU to take the pointer as key, but that would break Gets, since they are all done with different copies of the strings.

I'm still not entirely sure why we merged that PR, I didn't see any proof of the memory savings we would get.

howardjohn:
If the two PRs are mutually exclusive I would strongly prefer this one. This PR solves a legitimate memory leak that can cause ~infinite memory usage, and in real world clusters does do essentially this. The str pointer did have some improvement in some cases, but its an incremental benefit in some specific use cases.

If we can do both, great.

sschepens:
@howardjohn I just found another possible issue that could cause memory leaks. If we were to Add a key that was already present (without evictions happening) then we again leak strings (or string pointers) in the sets, because we never remove the previous index keys associated with the element we re-inserted.

I'm thinking of a way to fix this, the most easy one is to check if the key already exists when writing, if it does, clean it before writing it again.

maybe @hzxuzhonghu can think of a way to fix the string pointer stuff, I only see three solutions:
- keep string pointers, save a `map[string]*string` to be able to retreive original string pointers, probably beats the memory savings we got with string pointers in the first place.
- revert string pointers.
- keep string pointers but fix the leak some other way.

pkg/util/sets/string.go
hzxuzhonghu: Prefer no couple unrelated refactor

sschepens:
@hzxuzhonghu there is no use in benchmarking my fix is broken with the chances you introduced un the other PR and no easy way yo fix it.

Please tras previous comments.

sschepens:
I will be adding a test to validate that evictions are not leaking, but it will fail until we make a decision.

hzxuzhonghu:
What do you mean by broken?

hzxuzhonghu:
you can unreference.the key

sschepens:
Just te be clear, what I mean is:
```go
a := "a"
b := "a"
set := StringPointerSet{}
set.Insert(&a)

set.Delete(&b)
// set.IsEmpty() == false
set.Delete(&a)
// set.IsEmpty() == true
```

I cannot clean the original string pointer (`&a`) because I no longer have that original string.

PR #39452, opened 2022-06-14
[tf] Misc cleanup

nmittler:
Also includes some new features that will be used by a follow-up refactoring of some security tests.

### DIFF
--- a/pkg/test/framework/analyzer.go
+++ b/pkg/test/framework/analyzer.go
@@ -63,7 +63,7 @@ func newSuiteAnalyzer(testID string, fn mRunFn, osExit func(int)) Suite {
 	}
 }
 
-func (s *suiteAnalyzer) EnvironmentFactory(fn resource.EnvironmentFactory) Suite {
+func (s *suiteAnalyzer) EnvironmentFactory(_ resource.EnvironmentFactory) Suite {
 	if s.envFactoryCalls > 0 {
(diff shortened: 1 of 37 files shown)


tests/integration/security/sds_ingress/quic/ingress_test.go
```diff
@@ -70,6 +71,7 @@ func TestTlsGatewaysWithQUIC(t *testing.T) {
 // private key, server certificate and CA certificate for each mTLS gateway. Verifies that client can communicate
 // by using both QUIC and TCP/mTLS
 func TestMtlsGatewaysWithQUIC(t *testing.T) {
+	// nolint: staticcheck
```
stevenctl, line 74: Why do these need to be added?

pkg/test/framework/analyzer.go
```diff
@@ -63,7 +63,7 @@ func newSuiteAnalyzer(testID string, fn mRunFn, osExit func(int)) Suite {
 	}
 }
 
-func (s *suiteAnalyzer) EnvironmentFactory(fn resource.EnvironmentFactory) Suite {
+func (s *suiteAnalyzer) EnvironmentFactory(_ resource.EnvironmentFactory) Suite {
```
stevenctl, line 66: Why _ instead of removing the param or leaving it alone?

PR #39453, opened 2022-06-15
[tf] Refactoring top-level security tests

nmittler:
This refactors all of the top-level security tests to use the
common echo deployment and the new echotest framework. Also moves
the authz tests back to the top-level, since it can now share
the same TestMain.


GregHanson (approved):

howardjohn (approved):

pkg/test/framework/components/jwt/server.go
```diff
@@ -0,0 +1,63 @@
...
+
+package jwt
+
+import (
+	"istio.io/istio/pkg/test/framework"
+	"istio.io/istio/pkg/test/framework/components/namespace"
+	"istio.io/istio/pkg/test/framework/resource"
+)
+
+// Server for custom authz.
```
aryan16, line 23: nit : It should be authn right ?

pkg/test/framework/components/jwt/server.go
```diff
@@ -0,0 +1,63 @@
...
+type Server interface {
+	Namespace() namespace.Instance
+	FQDN() string
+	HTTPPort() int
+	HTTPSPort() int
+	JwksURI() string
+}
+
+// New creates a new authz Server.
```
aryan16, line 32: nit : same as above

aryan16 (approved):

PR #39457, opened 2022-06-15
[e2e test]add telemetry e2e test with default provider

zirain:

add telemetry e2e test with default provider

### DIFF
--- a/tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
+++ b/tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
@@ -111,8 +111,8 @@ metadata:
 spec:
   accessLogging:
   - disabled: %v
-`, !enableLogs)
-	t.ConfigIstio().YAML(common.GetAppNamespace().Name(), config).ApplyOrFail(t)
+`, enableLogs)
+	t.ConfigIstio().YAML(common.GetAppNamespace().Name(), config).DeleteOrFail(t)
 }
 
 func runAccessLogsTests(t framework.TestContext, expectLogs bool) {
--- a/tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
+++ b/tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
@@ -42,7 +42,6 @@ import (
 	"istio.io/istio/pkg/test"
 	"istio.io/istio/pkg/test/framework"
 	"istio.io/istio/pkg/test/framework/components/echo"
-	"istio.io/istio/pkg/test/framework/components/istio"
 	"istio.io/istio/pkg/test/util/retry"
 	common "istio.io/istio/tests/integration/telemetry/stats/prometheus"
 )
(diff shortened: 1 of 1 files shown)


douglas-reid (commented):
Can we move the access logging tests out of the `stats/prometheus` hierarchy?

i feel like this should be something like: `tests/integration/telemetry/accesslogs`.

@howardjohn how did this end up inside the `stats` hierarchy again?  were we trying to save on setup costs?  do we need to consider a refactor that exercises all telemetry API features instead (`tests/integration/telemetry/api`) ?

zirain:
> 

+1 move all accesslogs e2e tests to an new folder.

zirain:
open a [ticket](https://github.com/istio/istio/issues/39623) to track this, cc @douglas-reid

tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
```diff
@@ -51,15 +52,55 @@ func TestAccessLogs(t *testing.T) {
...
 }
 
-func runAccessLogsTests(t framework.TestContext, expectLogs bool) {
+func TestAccessLogsDefaultProvider(t *testing.T) {
+	framework.NewTest(t).
+		Features("observability.telemetry.logging.defaultprovider").
+		Run(func(t framework.TestContext) {
+			t.NewSubTest("disabled").Run(func(t framework.TestContext) {
+				cfg := `
+accessLogFile: "/dev/null"
```
howardjohn, line 73: a bit confused about /dev/null.. should it be `""` if we are testing disabled?

tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
```diff
@@ -51,15 +52,55 @@ func TestAccessLogs(t *testing.T) {
...
+`
+				ist := *(common.GetIstioInstance())
+				istio.PatchMeshConfigOrFail(t, ist.Settings().SystemNamespace, t.Clusters(), cfg)
+				runAccessLogsTests(t, true)
+				cfg = `
+accessLogFile: "/dev/stdout"
+defaultProviders:
+  accessLogging: []
+`
+				istio.PatchMeshConfigOrFail(t, ist.Settings().SystemNamespace, t.Clusters(), cfg)
```
howardjohn, line 98: what is this doing? If its trying to revert the change,PatchMeshConfigOrFail automatically does that?

zirain: you're right, the code is move from a really branch, in prev context there's no `PatchMeshConfigOrFail`, we do not need it now.

tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
zirain: fixed

tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
zirain: fixed

tests/integration/telemetry/stats/prometheus/nullvm/accesslogs_test.go
```diff
@@ -69,13 +99,30 @@ spec:
   - providers:
     - name: envoy
     disabled: %v
-`, !expectLogs)
+`, !enableLogs)
+	t.ConfigIstio().YAML(common.GetAppNamespace().Name(), config).ApplyOrFail(t)
+}
+
+func deleteTelemetryResource(t framework.TestContext, enableLogs bool) {
```
douglas-reid, line 106: I'm confused. This looks exactly the same as `applyTelemetryResource`. Shouldn't this end with a `DeleteOrFail` ?

douglas-reid (approved):

PR #39482, opened 2022-06-16
refactoring external CA tests

aryan16:
**This PR adds a separate istio operator config for external control plane, and enabling the external CA tests to run in external Control Plane topology**


tests/integration/security/external_ca/main_test.go
```diff
@@ -98,11 +97,23 @@ func setupConfig(ctx resource.Context, cfg *istio.Config) {
...
+components:
+  ingressGateways:
+  - name: istio-ingressgateway
+    enabled: false
+  egressGateways:
+  - name: istio-egressgateway
+    enabled: false
+  pilot:
+    enabled: false
+`
```
frankbu, line 112: Are you sure lines 102-112 are needed? They don't seem to be changing anything from what's already defined in the remote defaults file: https://github.com/istio/istio/blob/master/tests/integration/iop-remote-integration-test-defaults.yaml

tests/integration/security/external_ca/main_test.go
```diff
@@ -135,7 +146,6 @@ components:
 {{- else }}
```
frankbu, line 146: Same question as above, why are lines 139-144 needed? They don't seem to override anything. Can't you remove them and change line 146 to:
```
{{- if not .isConfigCluster}}
```

pkg/test/framework/components/istio/config.go
```diff
@@ -153,6 +153,9 @@ type Config struct {
 
 	// EnableCNI indicates the test should have CNI enabled.
 	EnableCNI bool
+
+	// Custom External Control Plane cluster values
+	ExternalControlPlaneClusterValues string
```
frankbu, line 158: nit: maybe move this up to line 117, so it's close to `ControlPlaneValues` and rename to `ExternalControlPlaneValues` for better name consistency.

tests/integration/security/external_ca/main_test.go
aryan16: we are adding a separate yaml on top of base remote yaml. In most of the cases it's control plane values (https://github.com/istio/istio/blob/master/pkg/test/framework/components/istio/operator.go#L972)

frankbu: Right, I think you mentioned that to me before already. Looking at it, I'm not sure that code is quite right. Let me think about it for a bit.

frankbu: After looking at this more carefully, I think the problem is because the `createIstioctlConfigFile()` function is doing some clever things that only work with an ordinary Istio installation. Try changing it to the following:
```
func createIstioctlConfigFile(s *resource.Settings, workDir string, cfg Config, isExternalCP bool) (istioctlConfigFiles, error) {
	var err error
	configFiles := istioctlConfigFiles{
		iopFile:       "",
		configIopFile: "",
		remoteIopFile: "",
	}

	// Generate the istioctl config file for control plane(primary) cluster
	configFiles.iopFile = filepath.Join(workDir, "iop.yaml")
	if configFiles.operatorSpec, err = initIOPFile(s, cfg, configFiles.iopFile, cfg.ControlPlaneValues); err != nil {
		return configFiles, err
	}

	// Generate the istioctl config file for remote cluster
	if !isExternalCP && !cfg.IstiodlessRemotes {
		if cfg.RemoteClusterValues == "" {
			cfg.RemoteClusterValues = cfg.ControlPlaneValues
		}
	}
	configFiles.remoteIopFile = filepath.Join(workDir, "remote.yaml")
	if configFiles.remoteOperatorSpec, err = initIOPFile(s, cfg, configFiles.remoteIopFile, cfg.RemoteClusterValues); err != nil {
		return configFiles, err
	}

	// Generate the istioctl config file for config cluster
	if isExternalCP {
		if cfg.ConfigClusterValues == "" {
			cfg.ConfigClusterValues = cfg.RemoteClusterValues
		}
		configFiles.configIopFile = filepath.Join(workDir, "config.yaml")
		if configFiles.configOperatorSpec, err = initIOPFile(s, cfg, configFiles.configIopFile, cfg.ConfigClusterValues); err != nil {
			return configFiles, err
		}
	} else {
		configFiles.configIopFile = configFiles.iopFile
		configFiles.configOperatorSpec = configFiles.operatorSpec
	}

	return configFiles, nil
}
```
After that, I think you should should be able to remove lines 102-112 and also line 165-187, and then you should also be able to remove your `externalControlPlaneFile` parameter from operator.go.

aryan16: So `externalControlPlaneFile` is for external Primary cluster, in the above-mentioned code, settings for external primary cluster is not defined anywhere so I think we require an additional variable to store these values. And, this piece of logic is not looking right because we can't use the remote cluster configuration directly for config cluster as it may have the mesh config details as well.```// Generate the istioctl config file for config cluster
	if isExternalCP {
		if cfg.ConfigClusterValues == "" {
			cfg.ConfigClusterValues = cfg.RemoteClusterValues
		}
		configFiles.configIopFile = filepath.Join(workDir, "config.yaml")
		if configFiles.configOperatorSpec, err = initIOPFile(s, cfg, configFiles.configIopFile, cfg.ConfigClusterValues); err != nil {
			return configFiles, err
		}
	} else {
		configFiles.configIopFile = configFiles.iopFile
		configFiles.configOperatorSpec = configFiles.operatorSpec
	}```

frankbu: 1. The primary cluster overrides should work, since after my suggested changes, the base file in the external control plane case will be correct. It will already have the k8s Deployment patches that you are adding on line 165-187. It should just work without you adding them explicitly in your overrides.
2. Defaulting `ConfigClusterValues to the `RemoteClusterValues` is also fine since they are just common overrides. The base file in the config case (vs remote case) will already include the extra config values. Note, this (like the `RemoteClusterValues` defaulting to `ControlPlaneValues` in the non-istiodless case, is just a convenience. You can still add config specific override instead if you need to (which you are doing).

Please give my suggestion a try. If it doesn't work, I can try it out myself in a couple of days, and see if I'm missing something.

aryan16: Sure, I will give it a shot. One more clarification - the main reason I added `externalControlPlaneFile` because in our current logic for merging multiple istio operator yamls, we don't append values for `overlays` instead we just update it with the values present in the latter yamls (https://github.com/istio/istio/blob/master/operator/pkg/manifest/shared.go#L240). In the test code, we have `overlays` in our istio operator (https://github.com/istio/istio/blob/master/tests/integration/security/external_ca/main_test.go#L147) and we remove the existing overlay from external control plane operator (https://github.com/istio/istio/blob/master/tests/integration/iop-externalistiod-primary-integration-test-defaults.yaml#L19).

frankbu: > in our current logic for merging multiple istio operator yamls, we don't append values for overlays instead we just update it with the values present in the latter yamls

Oh, that's unfortunate. Is there no way to patch in the whole overlay, something like the way volumes are patched in the istiod Deployment at the end of the list? If not, then you will need to keep lines 165-187 in the test, but you still shouldn't need your changes in operator.go (i.e., the separate values field for external control plane). Just use the `ControlPlaneValues` for the control plane overrides. With the conditional include in the template, it should work for both external control plane and ordinary primary cluster control planes. We don't currently have any test configurations that have a mix of both external cp and primary cp clusters, and if we did I think there are a few other things that would also need to change to support that.

aryan16: Yeah I agree we don't have any topology which includes both `externalControlPlane` and `primary cp` both. But to make it clear for the person who is reading the code, it could be feasible to add a separate variable for `externalControlPlane` values. While working on the tests I realized it's pretty confusing how we just play around with the values. That's my suggestion, we can implement it in later PRs if it is required.

frankbu (approved):
LGTM.

Shame about the operator replacing overlays instead of patch/merging them. I wonder if we could fix that in the future, since now it requires careful copy of base config into user config (tests), which could be wrong if the base profile ever changes.

frankbu:
> I believe that could be happening for a reason, probably for some tests we would want to replace the overlays with the new overlays instead of just appending the new value.

Maybe. Then I suppose we would add a field to specify merge or overwrite. I guess we'd need a field (default overwrite) for backward compatibility anyway.

howardjohn (approved):

aryan16:
cc : @zirain

zirain (approved):

PR #39483, opened 2022-06-16
Support pulling multi-arch envoy binaries

howardjohn:
For https://github.com/istio/istio/issues/26652#issuecomment-1155682106

This PR nearly finalizes arm64 support in Istio. With this, release-builder will publish amd64 and arm64 builds. We also have e2e tests using arm64 which works as well.

Changes required to get this to work:
* Refactor env vars (split into https://github.com/istio/istio/pull/40252)
* Update kind image to multi-arch one
* Update jwt-server to be multi-arch
* Update docker build to pull the correct base image per-arch for crane builder, as well as tag the final image with the correct architecture
* Update docker builder to allow building for multiple platforms without multi-arch builds. For --load and --save, multi-arch is not supported. Instead, if these cases are detect we will make single-arch images with an `-arch` suffix. Note: this is fairly internal facing -- once we publish them we will merge them back into a single multi-arch image.
* Update integration tests to dynamically pick architecture based on current arch. Integration tests are always using local arch, not cross compile
* Update app_sidecar test images to support multi-arch. Most of this was in docker-copy.sh, which was updated to also support detecting .deb architecture, which came with some refactoring.
* app_sidecar has RUN commands, requiring qemu for cross compile. This technically works now, but IMO we don't want/need this. For release, we can just skip arm64 app_sidecar builds since they are just test images, for simplicity. We can revisit this in the future if needed. To support this, add a new `--qemu` flag. If its not set, we will skip cross-compiling images that are labeled as requiring emulation.

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -146,6 +146,7 @@ default: init build test
 .PHONY: init
 # Downloads envoy, based on the SHA defined in the base pilot Dockerfile
 init: $(TARGET_OUT)/istio_is_init
+	echo ISTIO_ENVOY_LINUX_RELEASE_PATH=$(ISTIO_ENVOY_LINUX_RELEASE_PATH) TARGET_OUT_LINUX=$(TARGET_OUT_LINUX)
 	@mkdir -p ${TARGET_OUT}/logs
 	@mkdir -p ${TARGET_OUT}/release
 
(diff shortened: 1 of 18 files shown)


ericvn (approved):
LGTM

zirain:
@howardjohn can we move the `hold`?

ericvn (approved):
Approve the Doc file updates.

PR #39488, opened 2022-06-16
pilot: drop tunnel-ability implementation

howardjohn:
This was initially added to support HBONE, but hasn't been used since implemented. I think the real implementation direction is going for a different approach, so we can clean this up.

Essentially reverts https://github.com/istio/istio/commit/86b87f2f923c62344a1887b47adbc69bd6c2ddaa


hzxuzhonghu (commented):
LGTM @lambdai ptal

ramaraochavali (approved):

PR #39492, opened 2022-06-16
Echo Test: Use 17171 as the port for probes for custom grpc app

yashykt:
Use 17171 as the port for probes for the custom grpc app, since 3333 would be listened on by the Go server

### DIFF
--- a/common/scripts/kind_provisioner.sh
+++ b/common/scripts/kind_provisioner.sh
@@ -416,7 +416,7 @@ function cidr_to_ips() {
     python3 - <<EOF
 from ipaddress import ip_network;
 from itertools import islice;
-[print(str(ip) + "/" + str(ip.max_prefixlen)) for ip in islice(ip_network('$CIDR').hosts(), 1000, 1100)]
+[print(str(ip) + "/" + str(ip.max_prefixlen)) for ip in islice(ip_network('$CIDR').hosts(), 200, 1100)]
 EOF
 }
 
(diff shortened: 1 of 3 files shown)


common/scripts/kind_provisioner.sh
```diff
@@ -416,7 +416,7 @@ function cidr_to_ips() {
     python3 - <<EOF
 from ipaddress import ip_network;
 from itertools import islice;
-[print(str(ip) + "/" + str(ip.max_prefixlen)) for ip in islice(ip_network('$CIDR').hosts(), 1000, 1100)]
+[print(str(ip) + "/" + str(ip.max_prefixlen)) for ip in islice(ip_network('$CIDR').hosts(), 200, 1100)]
```
stevenctl, line 419: when merging common-files changes, you have to go through the github.com/istio/common-files repo

pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -263,14 +263,14 @@ spec:
           failureThreshold: 10
         livenessProbe:
           tcpSocket:
-            port: tcp-health-port
+            port: 3333
```
stevenctl, line 266: I think we should continue to use port naming for this, and change the portion of the template near line 234 where we say "if port == 3333 { this is the health port}

common/scripts/kind_provisioner.sh
yashykt: reverted

pkg/test/framework/components/echo/kube/deployment.go
yashykt: done

stevenctl (approved):

costinm (approved):

Issue #39180, opened 2022-05-27
istio ingress helm chart needs more variables/extensibility


ldemailly:
**Describe the feature request**

Need to be able to specify more settings for the ingress deployement istio chart

Specifically we need to change:

```json
{ "spec": { "template": { "spec": { "terminationGracePeriodSeconds" ...
```
and
```json
{ "spec": { "template": { "spec": {"containers":
            [ { "name": "istio-proxy", "lifecycle": { "preStop": ...
```

But having generic placeholder in each section, like there is already for instance
```
{{ toYaml $gateway.additionalContainers | indent 8 }}
```
https://github.com/istio/istio/blob/master/manifests/charts/gateways/istio-ingress/templates/deployment.yaml#L90-L269

would allow customization/extensibility without having to change/patch/fork the chart to support everything possible to change in a k8s object

**Describe alternatives you've considered**

We're using  `kubectl patch deployment` as a helmfile post sync but that cause 2 RS and churn

note: I may be missing an easier option

**Affected product area (please put an X in all that apply)**

[ ] Docs
[X] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[X] User Experience
[ ] Developer Infrastructure

ps: I wonder if additionalContainers can be used to do what we need by cheating on the annotation

zirain:
https://github.com/istio/istio/issues/39079#issuecomment-1135133605

ldemailly:
Ah interesting / thanks @zirain ! so https://docs.solo.io/gloo-edge/latest/installation/gateway/kubernetes/helm_advanced/ suggests to use `helm --post-renderer` with a kustomize script (I wonder if helm can maybe just use `kubectl apply -k` come to think of it)

helmfile doesn't support this yet out of the box (https://github.com/roboll/helmfile/issues/1530) but might be able to use `--args` (tbd)

All this being said; while I agree you can't have istio support every single possible change

- lifecycle hooks are key to errorless upgrades
- there are already a million variables so we could have 2 or 3 more
- but then maybe _instead_ we can have just a block in each main section for extensions ? (that won't help for some cases like overrides of hardcoded items but ...)
-

howardjohn:
[...]

PR #39493, opened 2022-06-17
helm: add custom volume mounts

nolotz:
Hi there,

Similar to how we had it in the old ingress charts, adds this pr custom volume mounts to the new gateway.

Without this change, secrets are difficult to add in Envoy Filters.

### DIFF
--- a/releasenotes/notes/39493.yaml
+++ b/releasenotes/notes/39493.yaml
@@ -1,0 +1,12 @@
+apiVersion: release-notes/v2
+
+kind: feature
+
+area: installation
+
+issue:
+  - 39493
+
+releaseNotes:
+  - |
+    **Added** values to the Istio Gateway Helm chart for configuring [volumes/volumeMounts](https://kubernetes.io/docs/concepts/storage/volumes/#secret) on the Deployment. Can be used to mount secrets and configs for envoy filter configurations.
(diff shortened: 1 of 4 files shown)


morvencao (commented):
/ok-to-test

releasenotes/notes/39493.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: release-notes/v2
+
+kind: feature
+
+area: installation
+
+issue:
+  - 39493
```
GregHanson, line 8: this should be a number reference to an issue on https://github.com/istio/istio/issues, not the PR#.  If there is not an associated github issue, can be left blank

nolotz: Hi @GregHanson,
https://github.com/istio/istio/issues/39493 should work though. 

For GitHub, issues and pull requests are the same and having the pull request linked would add contextual value.

zirain:
maybe we should follow this https://github.com/istio/istio/issues/39180#issuecomment-1142441351

nolotz:
> maybe we should follow this [#39180 (comment)](https://github.com/istio/istio/issues/39180#issuecomment-1142441351)

Hi @zirain,

I previously saw comments from @howardjohn and was already praying.
I'm fine with that even though such things as extraVolumes are already kinda standards in helm files.

A polished & clean helm will protect you better in battle than a dirty one.

Issue #27990, opened 2020-10-15
ServiceEntry without protocol for https port removes PassthroughCluster routing


Shohou:
**Bug description**
I created a ServiceEntry for a host like this:
```
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
  name: a-svc-entry
spec:
  hosts:
    - 'a.test.com'
  exportTo:
    - "."
  ports:
    - number: 443
      name: https
  location: MESH_EXTERNAL
  resolution: DNS
```
And suddenly all outgoing https traffic to any host started showing errors about mismatching certificate. I found the reason, it's a missing protocol for port and of course reference documentation says it's a mandatory field, but why does it work at all or why it works that way or why I can't find this described in docs or why it doesn't figure out protocol from name

[X] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Expected behavior**
Failing validation maybe, I'm not sure

**Steps to reproduce the bug**
Create ServiceEntry for https port without protocol and try to connect any other https host

**Version (include the output of `istioctl version --remote` and `kubectl version --short` and `helm version` if you used Helm)**
$ istioctl version --remote
client version: 1.7.3
control plane version: 1.7.3
data plane version: 1.7.1 (2 proxies), 1.7.3 (35 proxies)

$ kubectl version --short
Client Version: v1.18.8
Server Version: v1.19.0-34+1a52fbf0753680

$istioctl proxy-config listeners mypod
ADDRESS        PORT  MATCH                                      DESTINATION
...
0.0.0.0        443   App: HTTP                                  Route: 443
0.0.0.0        443   ALL                                        Cluster: outbound|443||a.test.com
...

howardjohn:
Basically what is happening is protocol unset means auto detect. Normally this means we listen onthe Service VIP. However, here we do not have one configured (you can set it in the address field, but it rarely makes sense for service entry). Therefor, the only thing we can do is match all traffic on the port.

I think we cannot make this invalid since there are use cases and likely users setting this, but we should add a validation *warning* when protocol is unset and address is also unset.

zufardhiyaulhaq:
@howardjohn can we make analyzer for this one?

howardjohn:
we actually have one sort of, PortNameAnalyzer, we can just extend it to ServiceEntry I think?

[...]

PR #28664, opened 2020-11-06
Add analyzer for empty protocol and addresses in Service Entry

zufardhiyaulhaq:
Issue related: https://github.com/istio/istio/issues/27990

Add warning analyzer for empty Protocol and Addresses in ServiceEntry object. Since if both empty, probably will cause a mesh wide error.

result:
```
istio $ ./out/linux_amd64/istioctl analyze --use-kube=false galley/pkg/config/analysis/analyzers/testdata/serviceentry-empty-protocol-addresses.yaml -n default    master
Warning [IST0133] (ServiceEntry service-entry-test-03.default galley/pkg/config/analysis/analyzers/testdata/serviceentry-empty-protocol-addresses.yaml:49) one or more Protocol empty in ServiceEntry default/service-entry-test-03 with Addresses Empty.
Warning [IST0133] (ServiceEntry service-entry-test-04.default galley/pkg/config/analysis/analyzers/testdata/serviceentry-empty-protocol-addresses.yaml:65) one or more Protocol empty in ServiceEntry default/service-entry-test-04 with Addresses Empty.
```

[x] Configuration Infrastructure
[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes
Please check any characteristics that apply to this pull request. 
[x] Does not have any changes that may affect Istio users.

PR #39495, opened 2022-06-17
Add protocol check for empty address in Service Entry

thehackercat:
Signed-off-by: LexusLee <lexuscyborg103@gmail.com>

Considering empty protocol might remove PassthroughCluster routing, add protocol validation when address is unset in ServiceEntry

image

Issue related: 
- https://github.com/istio/istio/issues/27990

we already have an analyzer for this case, I think we should also add a validation.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes
[x] Does not have any changes that may affect Istio users.

### DIFF
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -3208,6 +3208,11 @@ var ValidateServiceEntry = registerValidateFunc("ValidateServiceEntry",
 			if port.TargetPort != 0 {
 				errs = appendValidation(errs, ValidatePort(int(port.TargetPort)))
 			}
+			if len(serviceEntry.Addresses) == 0 {
+				if port.Protocol == "" || port.Protocol == "TCP" {
+					errs = appendValidation(errs, WrapWarning(fmt.Errorf("one or more protocol invalid in service entry with addresses empty.")))
+				}
+			}
 			errs = appendValidation(errs,
 				ValidatePortName(port.Name),
 				ValidateProtocol(port.Protocol),
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -3210,7 +3210,7 @@ var ValidateServiceEntry = registerValidateFunc("ValidateServiceEntry",
 			}
 			if len(serviceEntry.Addresses) == 0 {
 				if port.Protocol == "" || port.Protocol == "TCP" {
-					errs = appendValidation(errs, WrapWarning(fmt.Errorf("one or more protocol empty in service entry with addresses empty")))
+					errs = appendValidation(errs, WrapWarning(fmt.Errorf("addresses are required for ports serving TCP (or unset) protocol")))
 				}
 			}
 			errs = appendValidation(errs,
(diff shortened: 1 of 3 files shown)


howardjohn (changes requested):
Making something that is valid invalid is a backwards compatible API change. We can add a warning potentially - although if we do it's probably more than just unset protocol (TCP would be the same?)

thehackercat:
> Making something that is valid invalid is a backwards compatible API change. We can add a warning potentially - although if we do it's probably more than just unset protocol (TCP would be the same?)

Thanks for the quick response, I've changed to warning level and added TCP protocol validation.

thehackercat:
> Can you add a release note: [https://github.com/istio/istio/tree/master/releasenotes](https://github.com/istio/istio/tree/master/releasenotes?rgh-link-date=2022-06-17T15%3A00%3A58Z)

done, PTAL

pkg/config/validation/validation.go
```diff
@@ -3208,6 +3208,11 @@ var ValidateServiceEntry = registerValidateFunc("ValidateServiceEntry",
 			if port.TargetPort != 0 {
 				errs = appendValidation(errs, ValidatePort(int(port.TargetPort)))
 			}
+			if len(serviceEntry.Addresses) == 0 {
+				if port.Protocol == "" || port.Protocol == "TCP" {
```
hzxuzhonghu, line 3212: Why TCP invalid?

thehackercat: As explain in this comment https://github.com/istio/istio/pull/28664#issuecomment-725488225
When setting protocol to TCP with empty addresses, it makes all outgoing http traffic via port `xxxx` match `0.0.0.0_xxxx` tcp listener, which is unexpected.

hzxuzhonghu: But 0.0.0.0_xxx would not happen when auto allocate ip for serviceentry.

hzxuzhonghu: I agree with this solution since dns auto allocate is per proxy config

pkg/config/validation/validation.go
```diff
@@ -3208,6 +3208,11 @@ var ValidateServiceEntry = registerValidateFunc("ValidateServiceEntry",
...
+					errs = appendValidation(errs, WrapWarning(fmt.Errorf("one or more protocol invalid in service entry with addresses empty")))
```
howardjohn, line 3213: "invalid" is incorrect here. It is valid - it is just probably a bad idea. 

Maybe we can copy the similar text from the analyzer?

thehackercat: it's done.

pkg/config/validation/validation.go
```diff
@@ -3208,6 +3208,11 @@ var ValidateServiceEntry = registerValidateFunc("ValidateServiceEntry",
 			if port.TargetPort != 0 {
 				errs = appendValidation(errs, ValidatePort(int(port.TargetPort)))
 			}
+			if len(serviceEntry.Addresses) == 0 {
+				if port.Protocol == "" || port.Protocol == "TCP" {
+					errs = appendValidation(errs, WrapWarning(fmt.Errorf("one or more protocol empty in service entry with addresses empty")))
```
howardjohn, line 3213:
```suggestion
					errs = appendValidation(errs, WrapWarning(fmt.Errorf("addresses are required for ports serving TCP (or unset) protocol")))
```

GregHanson (approved):

PR #39497, opened 2022-06-17
Make istio-proxy's sub-server and prober can be uniform and compatible both for IPv4 and IPv6

zhlsunshine:
**Please provide a description of this PR:**

1. Make istio-proxy's sub-server can be compatible both for IPv4 and IPv6

>  status server
>  sts server
>  prober for health checking, readiness and liveness

2. Remove unnecessary wrapped IPv6 localhost

> Change wrapped IPv6 localhost from `[::1]` to `::1`,  this wrapped IPv6 localhost is specified as node.IPaddresses in some place , and then this IP address will be parsed by `net.ParseIP`, this will caused a IPv6 address can not be detected.

 

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #39503, opened 2022-06-17
[Dual Stack] Make istio-proxy's sub-server and prober can be uniform and compatible both for IPv4 and IPv6

zhlsunshine:
1. Make istio-proxy's sub-server can be compatible both for IPv4 and IPv6

>  status server
>  sts server
>  prober for health checking, readiness and liveness

2. Remove unnecessary wrapped IPv6 localhost

> Change wrapped IPv6 localhost from `[::1]` to `::1`,  this wrapped IPv6 localhost is specified as node.IPaddresses in some place , and then this IP address will be parsed by `net.ParseIP`, this will caused a IPv6 address can not be detected.

### DIFF
--- a/pilot/cmd/pilot-agent/main.go
+++ b/pilot/cmd/pilot-agent/main.go
@@ -47,7 +47,7 @@ import (
 
 const (
 	localHostIPv4 = "127.0.0.1"
-	localHostIPv6 = "[::1]"
+	localHostIPv6 = "::1"
 )
 
 // TODO: Move most of this to pkg options.
--- a/pilot/cmd/pilot-agent/status/server_test.go
+++ b/pilot/cmd/pilot-agent/status/server_test.go
@@ -663,14 +663,6 @@ func TestAppProbe(t *testing.T) {
 			ipv6:       true,
 		},
 		{
-			name:       "tcp-livez-wrapped-ipv6",
-			probePath:  "app-health/hello-world/livez",
-			config:     simpleTCPConfig,
-			statusCode: http.StatusOK,
-			podIP:      "[::1]",
-			ipv6:       true,
-		},
-		{
 			name:       "tcp-livez-localhost",
 			probePath:  "app-health/hello-world/livez",
 			config:     simpleTCPConfig,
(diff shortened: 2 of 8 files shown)


pilot/cmd/pilot-agent/main.go
```diff
@@ -47,7 +47,7 @@ import (
 
 const (
 	localHostIPv4 = "127.0.0.1"
-	localHostIPv6 = "[::1]"
```
howardjohn, line 50: If `[::1]` is not parsed properly somewhere, isn't that a bug? Its a valid IP, no?

pilot/cmd/pilot-agent/status/server_test.go
```diff
@@ -662,14 +662,6 @@ func TestAppProbe(t *testing.T) {
 			podIP:      "::1",
 			ipv6:       true,
 		},
-		{
-			name:       "tcp-livez-wrapped-ipv6",
-			probePath:  "app-health/hello-world/livez",
-			config:     simpleTCPConfig,
-			statusCode: http.StatusOK,
-			podIP:      "[::1]",
```
howardjohn, line 670: its a valid ip, still useful to test IMO

pilot/cmd/pilot-agent/status/server_test.go
```diff
@@ -737,6 +729,8 @@ func TestAppProbe(t *testing.T) {
 		defer cancel()
 		go server.Run(ctx)
 
+		// make sure the upstreamLocalAddress would not be nil
+		server.upstreamLocalAddress = &net.TCPAddr{IP: net.ParseIP("127.0.0.1")}
```
howardjohn, line 733: how come this is needed now?

zhlsunshine:
Hi @howardjohn,  how about to move these comments to PR https://github.com/istio/istio/pull/39497? Thanks for your comments

pilot/cmd/pilot-agent/status/server_test.go
zhlsunshine: it's okay to leave it and I will change the expected test result

pilot/cmd/pilot-agent/status/server_test.go
zhlsunshine: I used `upstreamLocalAddress` in my first commit, and forgot to remove it. I will remove it.

PR #39517, opened 2022-06-20
Allow specifying externalIPs on ingress gateways

georgyo:

Currently the gateway chart supports specifying externalIPs:

https://github.com/istio/istio/blob/aba44509973dc73e7fac4babbd6f3e78117f364e/manifests/charts/gateway/values.yaml#L57

But the same not same not true for [istio-ingress](https://github.com/istio/istio/tree/master/manifests/charts/gateways/istio-ingress) chart, which is what istio operator and istioctl use.

This adds that functionality which allows using the ingress with an additional load balancer in front of it.  

Regenerating values_types.pb.go seems to have a very large delta. I was unable to figure out how to trim that down or generate it with the same parameters of the previous generation. It does not seem to affect the building or running of istioctl or the operator. 

- [X] Configuration Infrastructure
- [ ] Docs
- [X] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn:
Seems the same as https://github.com/istio/istio/pull/39102 possibly?

georgyo:
Sorry for the double comment, I was mobile before.

Just read all the context on that change and other related changes. All the context in the following issues https://github.com/istio/istio/issues/39079, https://github.com/istio/istio/issues/39180, as well as the May 25th work group notes on the subject.

The problem is, all of those cover _only_ helm. Helm installation of Istio is marked as Alpha, Operator installation is marked as deprecated, leaving the only GA method of install/upgrade being istioctl. 

Istioctl uses the same helm charts, but currently feeds validates the input leaving no wiggle room for modifying the possible values being passed to a helm chart. While istioctl does allow specifying a `--manifests` option to choose were the charts come from, it is extremely strict on the possible values it will pass to it. This means in order to do something like externalIPs, you would need to hard code the values you want in the actual template. Doable, but very ugly. 

Since istioctl does not run continuously, it is also possible to patch the resources after install or upgrade. However that can cause a moment of downtime when the IP is no longer route-able. 

I agree that supporting all values for all kube objects in the helm chart is unreasonable, so I understand the decision making that might prevent this from being merged. However my 2c is that the barrier for changes needs to be somewhat achievable.

howardjohn:
istioctl also has an arbitrary patching functionality in the IstioOperator
spec

Issue #39468, opened 2022-06-15
access logging: multiple providers doesn't work


howardjohn:
```
  accessLogging:
  - providers:
      - name: envoy
  - providers:
      - name: envoy-json
```

only last one is used due to https://github.com/istio/istio/blob/6a1c1c887bfd7c802f564b15ba10ae9cfc0e0889/pilot/pkg/model/telemetry.go#L467.

cc @zirain

zirain:
the root case is not L467, we may need fix this in `applicableTelemetries`

douglas-reid:
@howardjohn iirc, you can however have multiple providers in the same clause:

```
```

howardjohn:
Hmm.. does that mean it is working as expected? It seems like it shouldn't be a list if it is..?

douglas-reid:
> Hmm.. does that mean it is working as expected? It seems like it shouldn't be a list if it is..?

i don't know if it is `working as expected`. The list allows for different configuration per group of providers, in theory. If the list doesn't support that, it is broken.

j2gg0s:
one Telemetry has many AccessLogging, one HCM has many access_log, So we should map Telemetry's AccessLogging to HCM's access_log, if workload mode match.

one AccessLogging has many Providers, but HCM's access_log dont have corresponding concept, for this oneToMany，what we expected? and what we actually code?

Current code is hard to understand, but before we changed, we should define what we expected first?
Especially second oneToMany, one AccessLogging has many Providers.

PR #39521, opened 2022-06-20
fix multi accesslogging not working

zirain:

fix: https://github.com/istio/istio/issues/39468

### DIFF
--- a/pilot/pkg/model/telemetry.go
+++ b/pilot/pkg/model/telemetry.go
@@ -75,7 +75,7 @@ type Telemetries struct {
 	// creation, we will preserve the Telemetries (and thus the cache) if not Telemetries are modified.
 	// As result, this cache will live until any Telemetry is modified.
 	computedMetricsFilters map[metricsKey]any
-	computedLoggingConfig  map[loggingKey]LoggingConfig
+	computedLoggingConfig  map[loggingKey][]LoggingConfig
 	mu                     sync.Mutex
 }
 
@@ -110,7 +110,7 @@ func getTelemetries(env *Environment) (*Telemetries, error) {
 		RootNamespace:          env.Mesh().GetRootNamespace(),
 		meshConfig:             env.Mesh(),
 		computedMetricsFilters: map[metricsKey]any{},
-		computedLoggingConfig:  map[loggingKey]LoggingConfig{},
+		computedLoggingConfig:  map[loggingKey][]LoggingConfig{},
 	}
 
 	fromEnv, err := env.List(collections.IstioTelemetryV1Alpha1Telemetries.Resource().GroupVersionKind(), NamespaceAll)
@@ -173,8 +173,16 @@ type tagOverride struct {
 type computedTelemetries struct {
 	telemetryKey
 	Metrics []*tpb.Metrics
-	Logging []*tpb.AccessLogging
+	Logging []*computedAccessLogging
 	Tracing []*tpb.Tracing
+}
+
+// computedAccessLogging contains the various AccessLogging configurations in scope for a given proxy,
+// include combined configurations for one of the following levels: 1. the root namespace level
+// 2. namespace level 3. workload level combined.
+type computedAccessLogging struct {
+	telemetryKey
+	Logging []*tpb.AccessLogging
 }
 
 type TracingConfig struct {
@@ -191,8 +199,8 @@ type TracingSpec struct {
 }
 
 type LoggingConfig struct {
-	Logs      []*accesslog.AccessLog
-	Providers []*meshconfig.MeshConfig_ExtensionProvider
+	AccessLog *accesslog.AccessLog
+	Provider  *meshconfig.MeshConfig_ExtensionProvider
 	Filter    *tpb.AccessLogging_Filter
 }
 
@@ -205,7 +213,7 @@ func workloadMode(class networking.ListenerClass) tpb.WorkloadMode {
 	case networking.ListenerClassSidecarOutbound:
 		return tpb.WorkloadMode_CLIENT
 	case networking.ListenerClassUndefined:
-		// this should not happened, just in case
+		// this should not happen, just in case
 		return tpb.WorkloadMode_CLIENT
 	}
 
@@ -215,7 +223,7 @@ func workloadMode(class networking.ListenerClass) tpb.WorkloadMode {
 // AccessLogging returns the logging configuration for a given proxy and listener class.
 // If nil is returned, access logs are not configured via Telemetry and should use fallback mechanisms.
 // If a non-nil but empty configuration is passed, access logging is explicitly disabled.
-func (t *Telemetries) AccessLogging(push *PushContext, proxy *Proxy, class networking.ListenerClass) *LoggingConfig {
+func (t *Telemetries) AccessLogging(push *PushContext, proxy *Proxy, class networking.ListenerClass) []LoggingConfig {
 	ct := t.applicableTelemetries(proxy)
 	if len(ct.Logging) == 0 && len(t.meshConfig.GetDefaultProviders().GetAccessLogging()) == 0 {
 		return nil
@@ -229,28 +237,31 @@ func (t *Telemetries) AccessLogging(push *PushContext, proxy *Proxy, class netwo
 	defer t.mu.Unlock()
 	precomputed, ok := t.computedLoggingConfig[key]
 	if ok {
-		return &precomputed
+		return precomputed
 	}
 
-	cfg := LoggingConfig{}
-	providers, f := mergeLogs(ct.Logging, t.meshConfig, workloadMode(class))
-	cfg.Filter = f
-	for _, p := range providers.SortedList() {
+	providers := mergeLogs(ct.Logging, t.meshConfig, workloadMode(class))
+	cfgs := make([]LoggingConfig, 0, len(providers))
+	for p, f := range providers {
 		fp := t.fetchProvider(p)
 		if fp == nil {
 			continue
 		}
+		cfg := LoggingConfig{
+			Provider: fp,
+			Filter:   f,
+		}
 
-		cfg.Providers = append(cfg.Providers, fp)
 		al := telemetryAccessLog(push, fp)
 		if al == nil {
 			continue
 		}
-		cfg.Logs = append(cfg.Logs, al)
+		cfg.AccessLog = al
+		cfgs = append(cfgs, cfg)
 	}
 
-	t.computedLoggingConfig[key] = cfg
-	return &cfg
+	t.computedLoggingConfig[key] = cfgs
+	return cfgs
 }
 
 // Tracing returns the logging tracing for a given proxy. If nil is returned, tracing
@@ -366,7 +377,7 @@ func (t *Telemetries) applicableTelemetries(proxy *Proxy) computedTelemetries {
 	namespace := proxy.ConfigNamespace
 	// Order here matters. The latter elements will override the first elements
 	ms := []*tpb.Metrics{}
-	ls := []*tpb.AccessLogging{}
+	ls := []*computedAccessLogging{}
 	ts := []*tpb.Tracing{}
 	key := telemetryKey{}
 	if t.RootNamespace != "" {
@@ -374,7 +385,12 @@ func (t *Telemetries) applicableTelemetries(proxy *Proxy) computedTelemetries {
 		if telemetry != (Telemetry{}) {
 			key.Root = NamespacedName{Name: telemetry.Name, Namespace: telemetry.Namespace}
 			ms = append(ms, telemetry.Spec.GetMetrics()...)
-			ls = append(ls, telemetry.Spec.GetAccessLogging()...)
+			ls = append(ls, &computedAccessLogging{
+				telemetryKey: telemetryKey{
+					Root: key.Root,
+				},
+				Logging: telemetry.Spec.GetAccessLogging(),
+			})
 			ts = append(ts, telemetry.Spec.GetTracing()...)
 		}
 	}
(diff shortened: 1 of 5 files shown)


pilot/pkg/model/telemetry.go
```diff
@@ -461,42 +482,54 @@ func mergeLogs(logs []*tpb.AccessLogging, mesh *meshconfig.MeshConfig, mode tpb.
...
+
 		if len(names) > 0 {
-			providerNames = names
+			providerNames = names.SortedList()
 		}
 
-		if m.Filter != nil {
-			loggingFilter = m.Filter
+		if inScopeKey != m.telemetryKey {
+			inScopeKey = m.telemetryKey
```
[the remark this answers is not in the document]
zirain: removed, this is unneeded now.

pilot/pkg/model/telemetry.go
```diff
@@ -162,10 +162,15 @@ type tagOverride struct {
 type computedTelemetries struct {
 	telemetryKey
 	Metrics []*tpb.Metrics
-	Logging []*tpb.AccessLogging
+	Logging []*computedAccessLogging
 	Tracing []*tpb.Tracing
 }
 
+type computedAccessLogging struct {
```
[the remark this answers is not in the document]
zirain: update desc

GregHanson (approved):

pilot/pkg/model/telemetry.go
```diff
@@ -462,41 +485,48 @@ func mergeLogs(logs []*tpb.AccessLogging, mesh *meshconfig.MeshConfig, mode tpb.
...
-		names := getProviderNames(m.Providers)
-		if len(names) > 0 {
-			providerNames = names
+		names := sets.New()
+		for _, p := range m.Logging {
+			subProviders := getProviderNames(p.Providers)
+			names.InsertAll(subProviders...)
+
+			if p.Filter != nil {
+				loggingFilter = p.Filter
```
howardjohn, line 494: this seems like we just pick the last one, shouldn't it be per-block?

pilot/pkg/model/telemetry.go
```diff
@@ -462,41 +485,48 @@ func mergeLogs(logs []*tpb.AccessLogging, mesh *meshconfig.MeshConfig, mode tpb.
...
+			}
 		}
 
-		if m.Filter != nil {
-			loggingFilter = m.Filter
+		if len(names) > 0 {
+			providerNames = names.SortedList()
```
howardjohn, line 499: nit: if we are just adding this to the inScopeProviders do we even need to sort it? Or convert it out of a set?

pilot/pkg/model/telemetry.go
zirain: `Filter` is hard to merged, I have no good idea on it from the begin, so I just keep as before.

howardjohn: Shouldn't it not be merged at all?  Right now its not very safe -- we may try to enabl a provider for only some rare events, but then the filter is ignored and we send all logs?

zirain:
@howardjohn I updated the implement, now `LogFilter` is provider-based, PTAL.

releasenotes/notes/39521.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: telemetry
+issue:
+  - https://github.com/istio/istio/issues/39468
+releaseNotes:
+  - |
+    **Fixed** a bug when specifying multi `accessLogging` in Telemetry, only last one is used.
```
douglas-reid, line 8: this needs to be much clearer, as within a single stanza, you could support multiple access logs and they would all be respected.

pilot/pkg/model/telemetry.go
```diff
@@ -126,10 +126,10 @@ type metricsConfig struct {
...
 type telemetryFilterConfig struct {
 	metricsConfig
-	Provider      *meshconfig.MeshConfig_ExtensionProvider
-	Metrics       bool
-	AccessLogging bool
-	LogsFilter    *tpb.AccessLogging_Filter
+	Provider              *meshconfig.MeshConfig_ExtensionProvider
+	Metrics               bool
+	AccessLogging         bool
+	StackdriverLogsFilter *tpb.AccessLogging_Filter
```
douglas-reid, line 132: why is this renamed to `StackdriverLogsFilter`?  The logging filter should be applicable for all logs. Does this PR disable support for basic envoy logs filtering?

pilot/pkg/model/telemetry.go
```diff
@@ -424,12 +456,19 @@ func (t *Telemetries) telemetryFilters(proxy *Proxy, class networking.ListenerCl
 		}
 		_, logging := tml[k]
 		_, metrics := tmm[k]
+
+		var logsFilter *tpb.AccessLogging_Filter
+		switch p.Provider.(type) {
+		case *meshconfig.MeshConfig_ExtensionProvider_Stackdriver:
```
douglas-reid, line 462: this is not right. the filter should not be only taken for Stackdriver.  why is this change being made?

pilot/pkg/model/telemetry_test.go
```diff
@@ -276,6 +277,96 @@ func TestAccessLogging(t *testing.T) {
 			},
 		},
 	}
+	multiAccessLogging := &tpb.Telemetry{
+		AccessLogging: []*tpb.AccessLogging{
```
douglas-reid, line 281: we need tests cases for a single providers stanza with multiple provider names.

pilot/pkg/model/telemetry_test.go
```diff
@@ -380,6 +471,22 @@ func TestAccessLogging(t *testing.T) {
 			nil,
 			[]string{},
 		},
+		{
+			"server - mutli filters",
```
douglas-reid, line 475:
```suggestion
			"server - multi filters",
```

douglas-reid (changes requested):

pilot/pkg/model/telemetry.go
zirain: in function `buildHTTPTelemetryFilter`, there's a comment `Only prometheus and SD supported currently`

pilot/pkg/model/telemetry.go
zirain: no, basic envoy logs are using `AccessLogging`, right now only SD need `telemetryFilters`.

pilot/pkg/model/telemetry.go
zirain: you're right, updated the implements.

pilot/pkg/model/telemetry.go
```diff
@@ -449,58 +482,66 @@ func (t *Telemetries) telemetryFilters(proxy *Proxy, class networking.ListenerCl
...
 	providerNames := mesh.GetDefaultProviders().GetAccessLogging()
+	filters := map[string]*tpb.AccessLogging_Filter{}
 	for _, m := range logs {
-		names := getProviderNames(m.Providers)
-		if len(names) > 0 {
-			providerNames = names
+		names := sets.New()
+		for _, p := range m.Logging {
+			subProviders := getProviderNames(p.Providers)
+			names.InsertAll(subProviders...)
```
j2gg0s, line 501: If we insert any AccessLogging's providers to `inScopeProviders`, do we still need check `isScopeProviders.Containers` in L524?

releasenotes/notes/39521.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: telemetry
+issue:
+  - https://github.com/istio/istio/issues/39468
+releaseNotes:
+  - |
+    **Fixed** a bug when specifying multi `accessLogging` in Telemetry within a single stanza, now they would all be respected.
```
douglas-reid, line 8:
```suggestion
    **Fixed** a bug that would ignore some configuration when specifying multiple `accessLogging` in Telemetry resources within a single stanza. With this fix, all provided access logging configuration within a single stanza of Telemetry resource are respected.
```

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -167,28 +167,33 @@ func (b *AccessLogBuilder) setTCPAccessLog(push *model.PushContext, proxy *model
...
+			continue
+		}
 
-	for _, p := range spec.Providers {
 		var al *accesslog.AccessLog
-		switch prov := p.Provider.(type) {
+		switch prov := spec.Provider.Provider.(type) {
 		case *meshconfig.MeshConfig_ExtensionProvider_EnvoyFileAccessLog:
 			// For built-in provider, fallback to Mesh Config for formatting options.
-			if p.Name == defaultEnvoyAccessLogProvider {
```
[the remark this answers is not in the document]
zirain: we have discuss this in https://github.com/istio/istio/pull/39606

pilot/pkg/model/telemetry.go
```diff
@@ -449,58 +482,66 @@ func (t *Telemetries) telemetryFilters(proxy *Proxy, class networking.ListenerCl
...
+
+			for _, prov := range subProviders {
+				// TODO: consider pick the non-nil?
```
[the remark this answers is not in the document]
zirain: thinking about the following configuration:
```yaml
kind: Telemetry
metadata:
  name: mutli-accessLogging
  namespace: istio-system
spec:
  accessLogging:
  - providers:
    - name: envoy
    - name: envoy-json
    filter: 
      expression: response.code >= 400
  - providers:
    - name: envoy-json
```

what's  `envoy-json`'s filter should be? ` response.code >= 400` or `nil`?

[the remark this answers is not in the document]
zirain: the root cause is how to merge `filter`.

pilot/pkg/model/telemetry.go
```diff
@@ -449,58 +482,66 @@ func (t *Telemetries) telemetryFilters(proxy *Proxy, class networking.ListenerCl
 }
 
 // mergeLogs returns the set of providers for the given logging configuration.
```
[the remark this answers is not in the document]
zirain: sounds good

zirain:
@douglas-reid @howardjohn refactor base on https://github.com/istio/istio/pull/39751, please review again.

do we need backport this to v1.13/v.14, cause there's a warn validating.
https://github.com/istio/istio/blob/d9762744d63553d6929a2b548ae8300cd3d78eb4/pkg/config/validation/validation.go#L3665

GregHanson (approved):

pilot/pkg/networking/core/v1alpha3/accesslog.go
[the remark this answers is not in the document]
douglas-reid: I don't understand.  Can you please clarify?  That PR was about falling back to MeshConfig. This logic is replacing the provider completely. That will break user expectations completely.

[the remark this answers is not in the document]
zirain: I got what your said, do you means, we should only fallback to MeshConfig when `p.Name == "envoy"` and `prov.EnvoyFileAccessLog.Path==""`?

howardjohn:
> @douglas-reid @howardjohn refactor base on #39751, please review again.
> 
> do we need backport this to v1.13/v.14, cause there's a warn validating.
> 
> https://github.com/istio/istio/blob/d9762744d63553d6929a2b548ae8300cd3d78eb4/pkg/config/validation/validation.go#L3665

I don't think we need to, fine with just on master.

pilot/pkg/networking/core/v1alpha3/accesslog.go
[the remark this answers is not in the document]
douglas-reid: let's leave this for now. i'm worried that people will create their own provider named `envoy` and this will be a mess, but we can address that later.

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -127,9 +127,9 @@ func newAccessLogBuilder() *AccessLogBuilder {
 
 func (b *AccessLogBuilder) setTCPAccessLog(push *model.PushContext, proxy *model.Proxy, tcp *tcp.TcpProxy, class networking.ListenerClass) {
 	mesh := push.Mesh
-	cfg := push.Telemetry.AccessLogging(push, proxy, class)
+	cfgs := push.Telemetry.AccessLogging(push, proxy, class)
 
-	if cfg == nil {
+	if cfgs == nil {
```
douglas-reid, line 132: curious: should we also check `len(cfgs)`? naively, it seems like `len(cfgs) == 0` might be appropriate (here and below) and cover the `nil` case too.

douglas-reid (approved):
one question remaining. otherwise looks reasonable. thanks for tracking this down and your patience in reviews.

pilot/pkg/model/telemetry.go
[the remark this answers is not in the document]
douglas-reid: Do we want to document how such a config might work in the API docs?  That would be helpful.

pilot/pkg/networking/core/v1alpha3/accesslog.go
zirain: we can improve this later, but for now keep the behavior as before, more details can be found in following test case:
```golang
                {
			"empty",
			nil,
			networking.ListenerClassSidecarOutbound,
			sidecar,
			nil,
			nil, // No Telemetry API configured, fall back to legacy mesh config setting
		},
```

pilot/pkg/model/telemetry.go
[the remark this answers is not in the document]
zirain: plan to update all the new things(`mode`, `filter` etc)

zirain:
kindly ping @istio/wg-networking-maintainers-pilot

howardjohn (approved):

PR #39523, opened 2022-06-20
Fix injection check in x injector list

hanxiaop:
Checking annotation is not enough, should also check label. If label is present, prefer to use label


irisdingbj (approved):
LGTM

PR #39525, opened 2022-06-20
Fix x injector ns pod info not accurate when use default

hanxiaop:
The pod revision was compared wrong, which lead to inaccurate info. Actually those pods are injected and the control plane is the default:
```
istioctl x injector list
NAMESPACE                     ISTIO-REVISION POD-REVISIONS
default                       default        <non-Istio>: 8 NEEDS RESTART: 8
istio-io-tcp-traffic-shifting default        <non-Istio>: 3 NEEDS RESTART: 3
istio-operator                default        <no pods>
sample-namespace              MISSING/1-14-0 <no pods>
test                          default        <non-Istio>: 3 NEEDS RESTART: 3
```
After the revision it will be like:
```
istioctl x injector list
NAMESPACE                     ISTIO-REVISION POD-REVISIONS
default                       default        default: 8
istio-io-tcp-traffic-shifting default        default: 3
istio-operator                default        <no pods>
sample-namespace              MISSING/1-14-0 <no pods>
test                          default        default: 3
```

### DIFF
--- a/istioctl/cmd/injector-list.go
+++ b/istioctl/cmd/injector-list.go
@@ -16,6 +16,7 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"sort"
@@ -36,6 +37,7 @@ import (
 	analyzer_util "istio.io/istio/pkg/config/analysis/analyzers/util"
 	"istio.io/istio/pkg/config/resource"
 	"istio.io/istio/pkg/kube"
+	"istio.io/istio/pkg/kube/inject"
 )
 
 type revisionCount struct {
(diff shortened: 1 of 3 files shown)


istioctl/cmd/injector-list.go
```diff
@@ -286,8 +286,15 @@ func podCountByRevision(pods []v1.Pod, expectedRevision string) map[string]revis
 	for _, pod := range pods {
 		revision := pod.ObjectMeta.GetLabels()[label.IoIstioRev.Name]
 		revisionLabel := revision
+		// if expectedRevision is "default", which is the injector revision default value,
```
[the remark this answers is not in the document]
hanxiaop: @howardjohn Thanks for reminding. I think the bug was casused due to the wrong revision extraction. I've improved the logic.

istioctl/cmd/injector-list.go
```diff
@@ -301,6 +303,22 @@ func podCountByRevision(pods []v1.Pod, expectedRevision string) map[string]revis
 	return retval
 }
 
+func extractRevisionFromPod(pod *v1.Pod) string {
+	rev := pod.GetLabels()[label.IoIstioRev.Name]
+	if rev != "" {
+		return rev
+	}
+	statusAnno, ok := pod.GetAnnotations()[annotation.SidecarStatus.Name]
```
howardjohn, line 311: why not just always use the status annotation? It is the real source of truth, not the label

hanxiaop: @howardjohn Fixed. Thanks

zhlsunshine:
LGTM

PR #39533, opened 2022-06-21
xds proxy minor changes

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [X] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/istio-agent/xds_proxy.go
+++ b/pkg/istio-agent/xds_proxy.go
@@ -100,10 +100,10 @@ type XdsProxy struct {
 	tapResponseChannel chan *discovery.DiscoveryResponse
 
 	// connected stores the active gRPC stream. The proxy will only have 1 connection at a time
-	connected           *ProxyConnection
-	initialRequest      *discovery.DiscoveryRequest
-	initialDeltaRequest *discovery.DeltaDiscoveryRequest
-	connectedMutex      sync.RWMutex
+	connected                 *ProxyConnection
+	initialHealthRequest      *discovery.DiscoveryRequest
+	initialDeltaHealthRequest *discovery.DeltaDiscoveryRequest
+	connectedMutex            sync.RWMutex
 
 	// Wasm cache and ecds channel are used to replace wasm remote load with local file.
 	wasmCache wasm.Cache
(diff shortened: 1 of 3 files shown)


pkg/istio-agent/xds_proxy.go
```diff
@@ -202,47 +202,37 @@ func initXdsProxy(ia *Agent) (*XdsProxy, error) {
...
 // to the upstream XDS request we will resend this request.
-func (p *XdsProxy) PersistRequest(req *discovery.DiscoveryRequest) {
+func (p *XdsProxy) sendHealthCheckRequest(req *discovery.DiscoveryRequest) {
 	p.connectedMutex.Lock()
-	// Immediately send if we are currently connect
+	// Immediately send if we are currently connected.
 	if p.connected != nil && p.connected.requestsChan != nil {
 		p.connected.requestsChan.Put(req)
 	}
 	// Otherwise place it as our initial request for new connections
```
costinm, line 234: Can you also remove the 'otherwise' - it is not.

costinm:
On Tue, Jun 21, 2022 at 7:31 AM John Howard ***@***.***>
wrote:

>
> In pkg/istio-agent/xds_proxy.go
> <https://github.com/istio/istio/pull/39533#discussion_r902698084>:
>
> > @@ -100,10 +100,10 @@ type XdsProxy struct {
>  	tapResponseChannel chan *discovery.DiscoveryResponse
>
>  	// connected stores the active gRPC stream. The proxy will only have 1 connection at a time
> -	connected           *ProxyConnection
> -	initialRequest      *discovery.DiscoveryRequest
> -	initialDeltaRequest *discovery.DeltaDiscoveryRequest
> -	connectedMutex      sync.RWMutex
> +	connected                 *ProxyConnection
> +	initialHealthRequest      *discovery.DiscoveryRequest
>
> this is generic code - prefer the old name.
>

It is not very generic - if we want to send some initial requests we would
have a list, or a more complicated structure since some of
the requests of this type may change.

And I'm not sure we have use case for generic code for sending initial
requests - if we find the use case we can refactor again.
I would use an array of callbacks to be called on connection - it would be
more flexible.

Issue #36069, opened 2021-11-13
make the sum of weights across destinations configurable.


ryan4yin:
**Describe the feature request**

>The proportion of traffic to be forwarded to the service version. (0-100). Sum of weights across destinations SHOULD BE == 100. If there is only one destination in a rule, the weight value is assumed to be 100.

configurable `weight_total` is useful, we need a more fine-grained control over the canary release, e.g. `weight_total=1000`.

the limit `weight_total=100` is not enough for us.

related issue: https://github.com/istio/istio/issues/18213
related docs: [VirtualService - HTTPRouteDestination](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRouteDestination)

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

howardjohn:
It would probably be ideal to just not require summing to 100? Then you can put something like `weight=2; weight=1`.

ryan4yin:
@howardjohn yeah, I think it's reasonable and meets our needs.

hzxuzhonghu:
I agree requiring total weight = 100 is a little bad. But not sure i understand a more fine-grained weight control. I am seeing no much diff between 1% and 0.1% in real case.

prabhu-mannu:
1% and 0.1% makes huge difference in high traffic applications/services, we ended up using haproxy for few of our applications to active 0.01%, it has proven very useful for us.
Removing constraint that total needs to be 100 would help, or rising the maximum value to 1000 would be also help.

ryan4yin:
> 1% and 0.1% makes huge difference in high traffic applications/services, we ended up using haproxy for few of our applications to active 0.01%, it has proven very useful for us. Removing constraint that total needs to be 100 would help, or rising the maximum value to 1000 would be also help.

agree with this

hzxuzhonghu:
Make sense, got it

hzxuzhonghu:
Great weaken the validation is permitted according to https://github.com/istio/api/blob/master/GUIDELINES.md

fatedier:
[...]

PR #39544, opened 2022-06-22
Loose route destination total weight validation

hzxuzhonghu:

fix https://github.com/istio/istio/issues/36069

with this, we do not require all the weight to sum up to 100 anymore. 

Previously, we have loosened that in API docs.


pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -842,42 +840,6 @@ func buildDestination(to k8s.BackendRef, ns, domain string) (*istio.Destination,
...
-		if remaining <= 0 {
-			break
-		}
-		remaining--
-		results[idx]++
-	}
-	return results
-}
-
 type argSlice struct {
```
howardjohn, line 843: nit - can this struct be removed too?

pkg/config/validation/validation.go
```diff
@@ -2784,11 +2786,7 @@ func validateRouteDestinations(weights []*networking.RouteDestination) (errs err
 			errs = multierror.Append(errs, errors.New("destination is required"))
 		}
 		errs = appendErrors(errs, validateDestination(weight.Destination))
-		errs = appendErrors(errs, ValidatePercent(weight.Weight))
-		totalWeight += weight.Weight
-	}
-	if len(weights) > 1 && totalWeight != 100 {
-		errs = appendErrors(errs, fmt.Errorf("total destination weight %v != 100", totalWeight))
+		errs = appendErrors(errs, validateWeight(weight.Weight))
```
howardjohn, line 2789: should we make sure total weight is non-zero?

howardjohn (commented):
do we not need any changes to the xds implementation?

hzxuzhonghu:
>do we not need any changes to the xds implementation?
Seems we do not need to change there, I have looked at the xds API, the weight requirement is very wide

pilot/pkg/config/kube/gateway/conversion.go
hzxuzhonghu: good

pkg/config/validation/validation.go
hzxuzhonghu: Let me test that behavior of envoy

hzxuzhonghu:
```	
for _, route := range routes {
		service := push.ServiceForHostname(node, host.Name(route.Destination.Host))
		if route.Weight > 0 {
			clusterName := istioroute.GetDestinationCluster(route.Destination, service, port.Port)
			clusterSpecifier.WeightedClusters.Clusters = append(clusterSpecifier.WeightedClusters.Clusters, &tcp.TcpProxy_WeightedCluster_ClusterWeight{
				Name:   clusterName,
				Weight: uint32(route.Weight),
			})
		}
	}
```
agree with you, since we have this

howardjohn:
```
do we not need any changes to the xds implementation?
Seems we do not need to change there, I have looked at the xds API, the weight requirement is very wide
```

```
	// An integer between 0 and :ref:`total_weight
	// <envoy_v3_api_field_config.route.v3.WeightedCluster.total_weight>`. When a request matches the route,
	// the choice of an upstream cluster is determined by its weight. The sum of weights across all
	// entries in the clusters array must add up to the total_weight, which defaults to 100.
	Weight *wrappers.UInt32Value `protobuf:"bytes,2,opt,name=weight,proto3" json:"weight,omitempty"`
```

Doesn't this mean we need to set total_weight?

hzxuzhonghu:
I thought this is redundant, but from this desc it is necessary

pilot/pkg/networking/core/v1alpha3/route/route_test.go
```diff
@@ -887,14 +887,18 @@ func TestBuildHTTPRoutes(t *testing.T) {
...
 			cluster := weightedCluster.GetClusters()[i]
 			g.Expect(cluster.RequestHeadersToAdd).To(gomega.Equal(expectResult.reqAdd))
 			g.Expect(cluster.RequestHeadersToRemove).To(gomega.Equal(expectResult.reqRemove))
 			g.Expect(cluster.ResponseHeadersToAdd).To(gomega.Equal(expectResult.respAdd))
 			g.Expect(cluster.RequestHeadersToRemove).To(gomega.Equal(expectResult.reqRemove))
 			g.Expect(cluster.GetHostRewriteLiteral()).To(gomega.Equal(expectResult.authority))
+			totalWeight += cluster.Weight.GetValue()
 		}
+		// total weight must be set
+		g.Expect(weightedCluster.GetTotalWeight().GetValue()).To(gomega.Equal(totalWeight))
```
hzxuzhonghu, line 901: added this ut to check total weight

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -579,7 +581,8 @@ func applyHTTPRouteDestination(
 	} else {
 		action.ClusterSpecifier = &route.RouteAction_WeightedClusters{
 			WeightedClusters: &route.WeightedCluster{
```
howardjohn, line 583: we have more usage in buildOutboundNetworkFiltersWithWeightedClusters I think?

hzxuzhonghu: TcpProxy_WeightedCluster has no total_weight field

hzxuzhonghu: That's very weird, and i filed a issue to make this optional https://github.com/envoyproxy/envoy/issues/21845

howardjohn (approved):

hzxuzhonghu:
@istio/wg-user-experience-maintainers ptal

hanxiaop (approved):

ryan4yin:
@hzxuzhonghu thanks for this PR, We really need it.

So is this feature already available in 1.15? I did not found this feature in [announcing-1.15](https://istio.io/latest/news/releases/1.15.x/announcing-1.15/change-notes/)

zirain:
It landed in v1.15, but miss a release-notes.

ryan4yin:
> It landed in v1.15, but miss a release-notes.

ok, got it~

PR #39554, opened 2022-06-22
operator: fix 1.25 install

howardjohn:
There were a few issues here:
* We always used the old versions for prune, and only sometimes added
  the new ones. This means on 1.25 we get a bunch of errors
* We never based the version to the helm renderer, so its ignored and we
  always use the old API versions.

This PR fixes both of these. A lot of plumbing to get the version
through

### DIFF
--- a/operator/cmd/mesh/manifest_shared_test.go
+++ b/operator/cmd/mesh/manifest_shared_test.go
@@ -24,6 +24,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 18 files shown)


GregHanson (approved):

ericvn (approved):

litong01 (approved):
Thanks so much for taking care of these improvements

hanxiaop (approved):

operator/cmd/mesh/manifest_shared_test.go
```diff
@@ -71,7 +72,7 @@ var (
 	testClient            client.Client
 	testReconcileOperator *istiocontrolplane.ReconcileIstioOperator
 
-	allNamespacedGVKs = append(helmreconciler.NamespacedResources,
+	allNamespacedGVKs = append(helmreconciler.NamespacedResources(&version.Info{Major: "1", Minor: "25"}),
```
zhlsunshine, line 75: LGTM, except here, if minor version `25` can be replaced with other one, could we defined a constant for it, including the change at L#209 in this file ?

PR #39568, opened 2022-06-23
pretty print json in debug endpoints

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/debug.go
+++ b/pilot/pkg/xds/debug.go
@@ -81,7 +81,7 @@ var indexTmpl = template.Must(template.New("index").Parse(`<html>
 <body>
 <br/>
 <p style = "font-family:Arial,Helvetica,sans-serif;">
-Note: Use <b>format</b> in query string (like <b>debug/configz?format</b>) to format the output.
+Note: Use <b>pretty</b> in query string (like <b>debug/configz?pretty</b>) to format the output.
 </p>
 <table id="endpoints">
 <tr><th>Endpoint</th><th>Description</th></tr>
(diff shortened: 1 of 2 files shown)


hzxuzhonghu:
not only endpoints, all debug interface json output has no indent

zirain:
can we make this configurable? sometime json output with indent will render too much lines, and will be difficulty to copy.

ramaraochavali:
>not only endpoints, all debug interface json output has no indent

I meant debug endpoints not istio endpoints

ramaraochavali:
> can we make this configurable? sometime json output with indent will render too much lines, and will be difficulty to copy.

Made it configurable via query param

pilot/pkg/xds/debug.go
```diff
@@ -1112,9 +1114,15 @@ func (p jsonMarshalProto) MarshalJSON() ([]byte, error) {
 }
 
 // writeJSON writes a json payload, handling content type, marshaling, and errors
-func writeJSON(w http.ResponseWriter, obj interface{}) {
+func writeJSON(w http.ResponseWriter, obj interface{}, req *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
-	b, err := config.ToPrettyJSON(obj)
+	var b []byte
+	var err error
+	if req.URL.Query().Has("format") {
```
zirain, line 1121: rename to `pretty`?

ramaraochavali: I have no strong preference. Will change

howardjohn:
IMO this is something that is not responsibility of the server - use `jq` on client side.

Fine with it if everyone else wants this though

ramaraochavali:
one use case I found this useful is to view smaller configs like mesh config with formatted output in the browser it self without using any other tools

hzxuzhonghu (approved):
LGTM

PR #39570, opened 2022-06-23
Update SE instances on WE change

dddddai:
Now we don't update service instances per SE in `workloadEntryHandler`, causing we are not able to [clean up old instances](https://github.com/istio/istio/blob/84d9f86021c64c65bc5cdc0b236da85d2f525907/pilot/pkg/serviceregistry/serviceentry/controller.go#L356-L359) on SE update, which potentially leaks memory

### DIFF
--- a/pilot/pkg/serviceregistry/serviceentry/controller.go
+++ b/pilot/pkg/serviceregistry/serviceentry/controller.go
@@ -264,6 +264,11 @@ func (s *Controller) workloadEntryHandler(old, curr config.Config, event model.E
 		}
 		instance := s.convertWorkloadEntryToServiceInstances(wle, services, se, &key, s.Cluster())
 		instancesUpdated = append(instancesUpdated, instance...)
+		if event == model.EventDelete {
+			s.serviceInstances.deleteServiceEntryInstances(namespacedName, key)
+		} else {
+			s.serviceInstances.updateServiceEntryInstancesPerConfig(namespacedName, key, instance)
+		}
 		addConfigs(se, services)
 	}
 
@@ -278,6 +283,7 @@ func (s *Controller) workloadEntryHandler(old, curr config.Config, event model.E
 		}
 		instance := s.convertWorkloadEntryToServiceInstances(wle, services, se, &key, s.Cluster())
 		instancesDeleted = append(instancesDeleted, instance...)
+		s.serviceInstances.deleteServiceEntryInstances(namespacedName, key)
 		addConfigs(se, services)
 	}
 
--- a/pilot/pkg/serviceregistry/serviceentry/controller_test.go
+++ b/pilot/pkg/serviceregistry/serviceentry/controller_test.go
@@ -628,6 +628,69 @@ func TestServiceDiscoveryWorkloadUpdate(t *testing.T) {
 		)
 	})
 
+	t.Run("update service entry host", func(t *testing.T) {
+		updated := func() *config.Config {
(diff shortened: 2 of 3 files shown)


pilot/pkg/serviceregistry/serviceentry/controller.go
```diff
@@ -264,6 +264,11 @@ func (s *Controller) workloadEntryHandler(old, curr config.Config, event model.E
 		}
 		instance := s.convertWorkloadEntryToServiceInstances(wle, services, se, &key, s.Cluster())
 		instancesUpdated = append(instancesUpdated, instance...)
+		if event == model.EventDelete {
+			s.serviceInstances.deleteServiceEntryInstances(namespacedName, key)
```
hzxuzhonghu, line 268: isn't L285-L291 doing this clean?

dddddai: It's updating based on the current SE, however if the SE is changed later, we will not be able to clean up instances associated with the old SE

Let's say we have a SE(host: a), then a WE is added, it will update service instances based on the SE(host: a). 
After that, the SE hostname is changed to b, but the instances added above(host: a) can't get freed

The added unit test is an example, it will fail without this change

hzxuzhonghu: >After that, the SE hostname is changed to b, but the instances added above(host: a) can't get freed

Not fully understand, when this happen, we so such clean job in `serviceEntryHandler`

dddddai: > when this happen, we so such clean job in `serviceEntryHandler`

Yes we [clean old instances by `instancesBySE`](https://github.com/istio/istio/blob/84d9f86021c64c65bc5cdc0b236da85d2f525907/pilot/pkg/serviceregistry/serviceentry/controller.go#L356-L359), but in `workloadEntryHandler` we didn't update `instancesBySE`, so the stale instances can't be correctly deleted in `serviceEntryHandler`

hzxuzhonghu: kind of understand

hzxuzhonghu: @dddddai Can you remove serviceInstancesStore.instancesBySE, i looked a little bit more and looks redundant

dddddai: I'm afraid not.. why do you think it's redundant?
We store instances per SE, so that we can clean up old instances on SE update, if we don't have `instancesBySE` then we will have to look up old instances with the old SE. It's basically a space-time tradeoff I think

hzxuzhonghu: NVM, this is my coarse idea.

hzxuzhonghu (approved):

GregHanson (approved):

ericvn:
Does this also need to be back ported to 1.15?

ericvn:
Also, is there a reason we din't try the automated cherry-pick?

dddddai:
> Also, is there a reason we din't try the automated cherry-pick?

No, thanks for reminding, it would be good to backport this to 1.13 and 1.14

kdorosh:
fyi https://github.com/istio/istio/pull/40792 and https://github.com/istio/istio/pull/40793 are already open

PR #39571, opened 2022-06-23
fix proxy_init.resources missed in sidecar injection

spwangxp:
Signed-off-by: spwangxp <wangshengpeng@cestc.cn>

although [proxy_init.resources defined](https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/values.yaml#L371), but sidecar injection template didn't use it. that's why no matter how to change it, the initContainer.resources are same as istio-proxy.resources.


howardjohn (changes requested):
This is working as intended. It would be great to have a PR to add a comment on the old unused field though

PR #39583, opened 2022-06-23
dump gateway conformance suite

stevenctl:


howardjohn (approved):
LGTM

I was moments away from working on this was well and was thinking of using namespace.Claim(). Not sure if that works though. This is pretty simple and should work :+1:

stevenctl:
I thought about claim as well, but I don't want to accidentally trigger cleanup logic.

PR #39584, opened 2022-06-23
[tf] Fix reached cluster logic for headless

nmittler:
Requests to headless services will only reach outside of the source cluster when DNS capture is enabled. This is due to the fact that we use DNS resolution for headless endpoints.

This change updates the `ReachedTargetClusters` checker to take into consideration whether the control plane is configured for DNS capture. This is done by reading MeshConfig from the cluster.

While doing this, I also did some general cleanup to the istio component and added a number of helpful methods for getting operator values and MeshConfig.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience


pkg/test/framework/components/echo/check/checkers.go
```diff
@@ -363,10 +365,17 @@ func URL(expected string) echo.Checker {
 
 // ReachedTargetClusters is similar to ReachedClusters, except that the set of expected clusters is
 // retrieved from the Target of the request.
-func ReachedTargetClusters(allClusters cluster.Clusters) echo.Checker {
+func ReachedTargetClusters(t framework.TestContext) echo.Checker {
+	dnsCaptureEnabled := istio.GetOrFail(t, t).MeshConfigOrFail(t).DefaultConfig.ProxyMetadata["ISTIO_META_DNS_CAPTURE"] == "true"
```
howardjohn, line 369: this function is probably not super fast... is it something we can call in a Checker efficiently?

howardjohn: Oh we are caching it. Hopefully accurate, given other things can change it...

We could have a function that always fetches fresh, and use a cached variant here, if its an issue.

pkg/test/framework/components/cluster/clusters.go
```diff
@@ -128,6 +128,18 @@ func (c Clusters) MeshClusters(excluded ...Cluster) Clusters {
 	}, exclude(excluded...))
 }
 
+// IsExternalControlPlane indicates whether the clusters are set up in an enternal
+// control plane configuration. An external control plane is a primary cluster that
+// gets its Istio configuration from a different cluster.
+func (c Clusters) IsExternalControlPlane() bool {
```
aryan16, line 134: Why do we require this method as it is already present in our test framework.

pkg/test/framework/components/echo/check/checkers.go
```diff
@@ -363,10 +365,17 @@ func URL(expected string) echo.Checker {
...
 	return func(result echo.CallResult, err error) error {
+		if !dnsCaptureEnabled && result.Opts.To.Config().IsHeadless() {
+			// Special case: requests to headless services will only go cross-cluster when
+			// DNS capture is enabled.
+			return checkReachedSourceClusterOnly(result, t.AllClusters())
+		}
+
 		expectedByNetwork := result.Opts.To.Clusters().ByNetwork()
-		return checkReachedClusters(result, allClusters, expectedByNetwork)
+		return checkReachedClusters(result, t.AllClusters(), expectedByNetwork)
```
aryan16, line 378: I don't think we deploy any workload in external control plane cluster, `t.Clusters()` would be good.

pkg/test/framework/components/cluster/clusters.go
aryan16: nvm got the reason.

pkg/test/framework/components/echo/check/checkers.go
nmittler: Just to be clear, are you suggesting that we have separate checkers for cached/fresh values? I'm not sure we want the tests to care.

WRT caching ... The way the caching works is that it gets updated each time the test environment changes meshconfig. I think that should be sufficient WRT correctness, no?

howardjohn: No, checker should always cache certainly or performance will be terrible. But there are other usages to get MeshConfig, and they may not play well with cached result.

This is all hypothetical.. if all tests pass we can just keep it as is and if needed in the future add a way to get an uncached version.

> WRT caching ... The way the caching works is that it gets updated each time the test environment changes meshconfig. I think that should be sufficient WRT correctness, no?

Possibly, but there is nothing enforcing that it isn't changed by other things (Istio operator, for example)

nmittler:
@howardjohn @aryan16 this is ready to go. Mind giving it another pass?

PR #39587, opened 2022-06-23
operator: fix remaining issue for 1.25 in cluster controller

howardjohn:
Note regarding why we don't see test changes/failures - they are in postsubmit only


litong01 (approved):
Thanks very much John

PR #39597, opened 2022-06-24
prepare selector compute

zouyee:
Signed-off-by: zounengren <zouyee1989@gmail.com>

 ServiceSelectorCache  is a cache of service selectors to avoid high CPU consumption caused by frequent calls to AsSelectorPreValidated


howardjohn (commented):
instead of catching the selector would it be simpler to just re-implement the selector using the map directly? It should be fairly simple and avoid the cost.

A bit worried about complexity here.

zouyee:
> Details

This method has been well practiced in the endpont controller of k8s.

zouyee:
> instead of catching the selector would it be simpler to just re-implement the selector using the map directly? It should be fairly simple and avoid the cost.
> 
> A bit worried about complexity here.

This method has been well practiced in the endpont controller of k8s.

howardjohn:
> This method has been well practiced in the endpont controller of k8s.

I don't think that is a convincing argument to do this on its own. Is there a technical reason this complexity is justified compared to just using the map directly?

hzxuzhonghu:
Not sure have you tested the performance? Which function call is CPU costing?

zouyee:
original `getPodServices`:
```
BenchmarkGetPodServices-8   	    2607	    385935 ns/op	  161913 B/op	    4023 allocs/op
```
current GetPodServiceMemberships which use ServiceSelectorCache
```
BenchmarkGetPodServiceMemberships-8   	    4287	    268554 ns/op	   49981 B/op	    1023 allocs/op
```
 more than 50% performance improvement

hzxuzhonghu:
You have to weigh off cpu and memory. It is not justice to say only CPU

howardjohn:
Please also compare it to just using the map directly.

PR #39602, opened 2022-06-24
Fix merge trafficpolicy

### LINKED ISSUE #39566: mesh h2UpgradePolicy: UPGRADE causes http2 requests to be made to external http endpoints
### Bug Description

Global mesh config is set to use h2UpgradePolicy: UPGRADE in order to gain connection pooling benefits across the istio mesh. We would like to have some requests through the ingress be directed to an S3 website endpoint, which only supports HTTP (no HTTPS) and does not support HTTP/2.

A ServiceEntry and DestinationRule, with `destinationrule.trafficPolicy.connectionPool.http.h2UpgradePolicy: DO_NOT_UPGRADE` have been created for this s3 website endpoint, but requests are failing immediately as the endpoint rejects the connection:
`[2022-06-23T02:51:31.891Z] "GET /abc/testing.txt HTTP/2" 502 UPE upstream_reset_before_response_started{protocol_error} - "-" 0 87 1 - ...`

Unfortunately the `%UPSTREAM_PROTOCOL%` log directive does not appear to be supported yet, so I cannot say it is making HTTP2 requests with 100% certainty.

Unsetting the global h2UpgradePolicy causes requests to succeed, even if `destinationrule.trafficPolicy.connectionPool.http.h2UpgradePolicy: UPGRADE` is set (which I would expect to break connections to the upstream).

### Version

```prose
Istio 1.14.1
```

### Additional Information

```
apiVersion: v1
data:
  mesh: |-
    defaultConfig:
      discoveryAddress: istiod.istio-system.svc:15012
      meshId: cluster.local
      proxyMetadata:
        ISTIO_META_DNS_AUTO_ALLOCATE: "true"
        ISTIO_META_DNS_CAPTURE: "true"
      tracing:
        datadog:
          address: $(HOST_IP):8126
    enablePrometheusMerge: true
    h2UpgradePolicy: UPGRADE
    rootNamespace: null
    trustDomain: cluster.local
  meshNetworks: 'networks: {}'
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: istiod
    meta.helm.sh/release-namespace: istio-system
  creationTimestamp: "2022-05-24T21:38:29Z"
  labels:
    app.kubernetes.io/managed-by: Helm
    install.operator.istio.io/owning-resource: unknown
    istio.io/rev: default
    operator.istio.io/component: Pilot
    release: istiod
  name: istio
  namespace: istio-system
  resourceVersion: "391273686"
  selfLink: /api/v1/namespaces/istio-system/configmaps/istio
  uid: 36efb7cc-ba8b-411e-8735-e6efa74889ec
```

```
kind: VirtualService
metadata:
  creationTimestamp: "2021-04-07T21:30:06Z"
  generation: 13
  name: static-assets-redirect
  namespace: default
  resourceVersion: "391161487"
  selfLink: /apis/networking.istio.io/v1beta1/namespaces/default/virtualservices/lsslite-assets-staging-internal-assets-redirect
  uid: eee17d06-c42a-422e-a98d-8dbd5406ce83
spec:
  gateways:
  - webapp-gateway
  hosts:
  - www.remotasks.dev
  http:
  - match:
    - uri:
        prefix: /abc
    rewrite:
      authority: some-bucket.s3-website-us-west-2.amazonaws.com
      uri: /xyz
    route:
    - destination:
        host: some-bucket.s3-website-us-west-2.amazonaws.com
        port:
          number: 80
```

```
kind: ServiceEntry
metadata:
  creationTimestamp: "2022-06-22T22:24:46Z"
  generation: 3
  name: external-svc-website-test
  namespace: default
  resourceVersion: "391181054"
  selfLink: /apis/networking.istio.io/v1beta1/namespaces/default/serviceentries/external-svc-website-test
  uid: af318651-9cd3-49de-b716-91acf92e2f61
spec:
  hosts:
  - <redacted>.s3-website-us-west-2.amazonaws.com
  location: MESH_EXTERNAL
  ports:
  - name: http
    number: 80
    protocol: HTTP
  resolution: DNS
```

```
kind: DestinationRule
metadata:
  creationTimestamp: "2022-06-22T22:25:31Z"
  generation: 6
  name: external-svc-website-test
  namespace: default
  resourceVersion: "391280428"
  selfLink: /apis/networking.istio.io/v1beta1/namespaces/default/destinationrules/external-svc-website-test
  uid: 1017bd6c-56f2-4030-aad1-b0dac4f13aa7
spec:
  host: <redacted>.s3-website-us-west-2.amazonaws.com
  trafficPolicy:
    connectionPool:
      http:
        h2UpgradePolicy: DO_NOT_UPGRADE
    loadBalancer:
      simple: ROUND_ROBIN
    portLevelSettings:
    - port:
        number: 80
      tls:
        mode: DISABLE
```

istioctl bug-report: 
```
Target cluster context: staging
Running with the following config:

context: staging
istio-namespace: istio-system
full-secrets: false
timeout (mins): 30
include: {  }
exclude: { Namespaces: kube-node-lease,kube-public,kube-system,local-path-storage }
end-time: 2022-06-22 19:42:24.404462 -0700 PDT



Cluster endpoint: https://<redacted>.sk1.us-west-2.eks.amazonaws.com
CLI version:
version.BuildInfo{Version:"1.13.2", GitRevision:"91533d04e894ff86b80acd6d7a4517b144f9e19a-dirty", GolangVersion:"go1.17.8", BuildStatus:"Clean", GitTag:"1.13.2"}

The following Istio control plane revisions/versions were found in the cluster:
Revision default:
&version.MeshInfo{
    {
        Component: "pilot",
        Info:      version.BuildInfo{Version:"1.14.1", GitRevision:"f59ce19ec6b63bbb70a65c43ac423845f1129464", GolangVersion:"", BuildStatus:"Clean", GitTag:"1.14.1"},
    },
    {
        Component: "pilot",
        Info:      version.BuildInfo{Version:"1.14.1", GitRevision:"f59ce19ec6b63bbb70a65c43ac423845f1129464", GolangVersion:"", BuildStatus:"Clean", GitTag:"1.14.1"},
    },
}
```

hzxuzhonghu:

fix #39566

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go
@@ -320,11 +320,20 @@ func MergeTrafficPolicy(original, subsetPolicy *networking.TrafficPolicy, port *
 	if port != nil {
 		for _, p := range subsetPolicy.PortLevelSettings {
 			if p.Port != nil && uint32(port.Port) == p.Port.Number {
-				// per the docs, port level policies do not inherit and instead to defaults if not provided
-				mergedPolicy.ConnectionPool = p.ConnectionPool
-				mergedPolicy.OutlierDetection = p.OutlierDetection
-				mergedPolicy.LoadBalancer = p.LoadBalancer
-				mergedPolicy.Tls = p.Tls
+				// per the docs, default values will be applied
+				// to fields omitted in port-level traffic policies.
+				if p.ConnectionPool != nil {
+					mergedPolicy.ConnectionPool = p.ConnectionPool
+				}
+				if p.OutlierDetection != nil {
+					mergedPolicy.OutlierDetection = p.OutlierDetection
+				}
+				if p.LoadBalancer != nil {
+					mergedPolicy.LoadBalancer = p.LoadBalancer
+				}
+				if p.Tls != nil {
+					mergedPolicy.Tls = p.Tls
+				}
 				break
 			}
 		}
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
@@ -703,6 +703,11 @@ func TestMergeTrafficPolicy(t *testing.T) {
 						Simple: networking.LoadBalancerSettings_LEAST_REQUEST,
 					},
 				},
+				ConnectionPool: &networking.ConnectionPoolSettings{
+					Http: &networking.ConnectionPoolSettings_HTTPSettings{
+						MaxRetries: 10,
+					},
+				},
 			},
 		},
 		{
@@ -751,6 +756,16 @@ func TestMergeTrafficPolicy(t *testing.T) {
 			},
 			port: &model.Port{Port: 8080},
 			expected: &networking.TrafficPolicy{
+				LoadBalancer: &networking.LoadBalancerSettings{
+					LbPolicy: &networking.LoadBalancerSettings_Simple{
+						Simple: networking.LoadBalancerSettings_ROUND_ROBIN,
+					},
+				},
+				ConnectionPool: &networking.ConnectionPoolSettings{
+					Http: &networking.ConnectionPoolSettings_HTTPSettings{
+						MaxRetries: 10,
+					},
+				},
 				OutlierDetection: &networking.OutlierDetection{
 					ConsecutiveErrors: 13,
 				},
--- a/pilot/pkg/networking/core/v1alpha3/cluster_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_test.go
@@ -531,10 +531,10 @@ func TestBuildClustersWithMutualTlsAndNodeMetadataCertfileOverrides(t *testing.T
 							Port: &networking.PortSelector{
 								Number: 8080,
 							},
+							Tls: &networking.ClientTLSSettings{
+								Mode: networking.ClientTLSSettings_DISABLE,
+							},
 						},
-					},
-					Tls: &networking.ClientTLSSettings{
-						Mode: networking.ClientTLSSettings_DISABLE,
 					},
 				},
 			},
--- a/pilot/pkg/networking/core/v1alpha3/cluster_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_test.go
@@ -533,6 +533,9 @@ func TestBuildClustersWithMutualTlsAndNodeMetadataCertfileOverrides(t *testing.T
 							},
 						},
 					},
+					Tls: &networking.ClientTLSSettings{
+						Mode: networking.ClientTLSSettings_DISABLE,
+					},
 				},
 			},
 		},
--- a/releasenotes/notes/39566.yaml
+++ b/releasenotes/notes/39566.yaml
@@ -1,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+issue:
+  - 39566
+releaseNotes:
+  - |
+    **Fixed** merging TrafficPolicy: port level unset traffic policy overrides default one.


hzxuzhonghu:
```
type TrafficPolicy struct {
	...
	Tls *ClientTLSSettings `protobuf:"bytes,4,opt,name=tls,proto3" json:"tls,omitempty"`
	// Traffic policies specific to individual ports. Note that port level
	// settings will override the destination-level settings. Traffic
	// settings specified at the destination-level will not be inherited when
	// overridden by port-level settings, i.e. default values will be applied
	// to fields omitted in port-level traffic policies.
	PortLevelSettings []*TrafficPolicy_PortTrafficPolicy `protobuf:"bytes,5,rep,name=port_level_settings,json=portLevelSettings,proto3" json:"port_level_settings,omitempty"`
	
```

The api says we should do what this patch does

ramaraochavali:
Another thing is current documented behaviour is different from what subset level traffic policy does? I do not know if there are any specific reasons for port level policies to not inherit parent -One  reason I could thin is, if it is a different port in a service, it is considered different set of APIs exposed by same service and hence should not inherit from main policy unlike subsets.

I think we should avoid breaking users - if we have to change may be some flag to control the new behaviour may be better.

GregHanson:
>One reason I could thin is, if it is a different port in a service, it is considered different set of APIs exposed by same service and hence should not inherit from main policy unlike subsets.

I do believe this was the original design reasoning, as well as indecision on how main, subset, and port policies should interact with one another.  Based on the API, the port level policies should be getting the default values specified in MeshConfig: `h2UpgradePolicy`, `tcpKeepalive`, `localityLbSetting`, timeouts, etc but NOT any of the main or subset trafficPolicy settings.  The DestinationRule API was written before MeshConfig, and so I assume as more settings got added to MeshConfig that implementation detail was overlooked.

hzxuzhonghu:
>I am a bit more inclined to just change the API docs to match the real behavior. WDYT?

TBH, I donot think this is a good way. It will make users feel istio community would respect a mistake they made yesterday rather than fixing it.

ramaraochavali:
> TBH, I donot think this is a good way. It will make users feel istio community would respect a mistake they made yesterday rather than fixing it.

I agree. If the original design was to have different port level settings to cater to different set of APIs that @GregHanson seems to also agree with - I think we can introduce a flag to change to new behaviour (documented behaviour) and add a release note to flip it back if needed and we can remove the flag few releases down the line (Similar to how Envoy handles breaking changes with runtime configuration). May be we should need some thing similar in Istio rather than adding too many env var flags. But we can do that later

hzxuzhonghu:
I agree with this suggestion, @howardjohn WDYT?

howardjohn:
> TBH, I donot think this is a good way. It will make users feel istio community would respect a mistake they made yesterday rather than fixing it.

Yes, this is exactly what I want. Our users have given us feedback over and over again that the most painful part of Istio is making backwards incompatible changes. Very few things are more important than preserving compatibility IMO. This doesn't seem like one of them.

> Similar to how Envoy handles breaking changes with runtime configuration

Envoy's API is largely an implementation detail of products, so I wouldn't use it as a rubrik for a good way to make breaking changes, as these are often hidden by higher level APIs like Istio

hzxuzhonghu:
@howardjohn What do you mean? Should i continue fixing in this PR or not?

howardjohn:
I don't think we should change the behavior of our existing APIs.

hzxuzhonghu:
// Traffic policies specific to individual ports. Note that port level
	// settings will override the destination-level settings. Traffic
	// settings specified at the destination-level will not be inherited when
	// overridden by port-level settings, i.e. default values will be applied
	// to fields omitted in port-level traffic policies.
	PortLevelSettings []*TrafficPolicy_PortTrafficPolicy `protobuf:"bytes,5,rep,name=port_level_settings,json=portLevelSettings,proto3" json:"port_level_settings,omitempty"`


This is not changing the API, on the opposite, I think the current implementation is not right.

howardjohn:
It _is_ bad. But making breaking API changes is the absolute worst thing we can do based on user feedback research

PR #39606, opened 2022-06-24
use mesh config defaults when log formatter is not specified

ramaraochavali:
As mentioned https://github.com/istio/istio/pull/36604#issuecomment-1165182725 Istio behaviour is changed when telemetry specifies default accesslog provider("envoy"). We loose the benefit of using mesh config defaults when logFormat is not specified.

This PR brings back that behaviour.

Also changed some function names to avoid confusion.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/accesslog.go
+++ b/pilot/pkg/networking/core/v1alpha3/accesslog.go
@@ -187,7 +187,7 @@ func buildAccessLogFromTelemetry(push *model.PushContext, spec *model.LoggingCon
 		var al *accesslog.AccessLog
 		switch prov := p.Provider.(type) {
 		case *meshconfig.MeshConfig_ExtensionProvider_EnvoyFileAccessLog:
-			// For default provider, fallback to Mesh Config for formatting options.
+			// For built-in provider, fallback to Mesh Config for formatting options.
 			if p.Name == defaultEnvoyAccessLogProvider {
 				al = fileAccessLogFromMeshConfig(prov.EnvoyFileAccessLog.Path, push.Mesh)
 			} else {
--- a/pilot/pkg/networking/core/v1alpha3/accesslog.go
+++ b/pilot/pkg/networking/core/v1alpha3/accesslog.go
@@ -73,6 +73,8 @@ const (
 	devStdout = "/dev/stdout"
 
 	celFilter = "envoy.access_loggers.extension_filters.cel"
+
+	defaultEnvoyAccessLogProvider = "envoy"
 )
 
 var (
@@ -185,7 +187,12 @@ func buildAccessLogFromTelemetry(push *model.PushContext, spec *model.LoggingCon
 		var al *accesslog.AccessLog
 		switch prov := p.Provider.(type) {
 		case *meshconfig.MeshConfig_ExtensionProvider_EnvoyFileAccessLog:
-			al = fileAccessLogFromTelemetry(prov.EnvoyFileAccessLog, push.Mesh)
+			// For default provider, fallback to Mesh Config for formatting options.
+			if p.Name == defaultEnvoyAccessLogProvider {
+				al = fileAccessLogFromMeshConfig(prov.EnvoyFileAccessLog.Path, push.Mesh)
+			} else {
+				al = fileAccessLogFromTelemetry(prov.EnvoyFileAccessLog)
+			}
 		case *meshconfig.MeshConfig_ExtensionProvider_EnvoyHttpAls:
 			al = httpGrpcAccessLogFromTelemetry(push, prov.EnvoyHttpAls)
 		case *meshconfig.MeshConfig_ExtensionProvider_EnvoyTcpAls:
@@ -311,17 +318,15 @@ func tcpGrpcAccessLogFromTelemetry(push *model.PushContext, prov *meshconfig.Mes
 	}
 }
 
-func fileAccessLogFromTelemetry(prov *meshconfig.MeshConfig_ExtensionProvider_EnvoyFileAccessLogProvider,
-	mesh *meshconfig.MeshConfig,
-) *accesslog.AccessLog {
+func fileAccessLogFromTelemetry(prov *meshconfig.MeshConfig_ExtensionProvider_EnvoyFileAccessLogProvider) *accesslog.AccessLog {
 	p := prov.Path
 	if p == "" {
 		p = devStdout
 	}
-
 	fl := &fileaccesslog.FileAccessLog{
 		Path: p,
 	}
+
 	needsFormatter := false
 	if prov.LogFormat != nil {
 		switch logFormat := prov.LogFormat.LogFormat.(type) {
@@ -331,9 +336,7 @@ func fileAccessLogFromTelemetry(prov *meshconfig.MeshConfig_ExtensionProvider_En
 			fl.AccessLogFormat, needsFormatter = buildFileAccessJSONLogFormat(logFormat)
 		}
 	} else {
-		// Provider did not specify formatter. Fallback to Mesh Config. This is useful
-		// in using the default "envoy" provider which does not have formatter specified.
-		return fileAccessLogFromMeshConfig(p, mesh)
+		fl.AccessLogFormat, needsFormatter = buildFileAccessTextLogFormat("")
 	}
 	if needsFormatter {
 		fl.GetLogFormat().Formatters = accessLogFormatters
--- a/pilot/pkg/networking/core/v1alpha3/accesslog.go
+++ b/pilot/pkg/networking/core/v1alpha3/accesslog.go
@@ -142,9 +142,9 @@ type AccessLogBuilder struct {
 
 func newAccessLogBuilder() *AccessLogBuilder {
 	return &AccessLogBuilder{
-		tcpGrpcAccessLog:         buildTCPGrpcAccessLog(false),
-		httpGrpcAccessLog:        buildHTTPGrpcAccessLog(),
-		tcpGrpcListenerAccessLog: buildTCPGrpcAccessLog(true),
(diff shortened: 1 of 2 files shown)


ramaraochavali:
@zirain FYI

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -329,7 +331,9 @@ func buildEnvoyFileAccessLogHelper(prov *meshconfig.MeshConfig_ExtensionProvider
 			fl.AccessLogFormat, needsFormatter = buildFileAccessJSONLogFormat(logFormat)
 		}
 	} else {
-		fl.AccessLogFormat, needsFormatter = buildFileAccessTextLogFormat("")
+		// Provider did not specify formatter. Fallback to Mesh Config. This is useful
+		// in using the default "envoy" provider which does not have formatter specified.
+		return fileAccessLogFromMeshConfig(p, mesh)
```
zirain, line 336: should this only work for `envoy` provider?

zirain: otherwise  we should update the desc
https://github.com/istio/api/blob/8a4564651730043c0600c112740952f65360a933/mesh/v1alpha1/config.proto#L842

pilot/pkg/networking/core/v1alpha3/accesslog_test.go
```diff
@@ -807,9 +822,40 @@ func TestBuildAccessLogFromTelemetry(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "default-envoy-provider",
+			ctx:  ctx,
+			meshConfig: &meshconfig.MeshConfig{
+				AccessLogEncoding: meshconfig.MeshConfig_JSON,
+				AccessLogFormat:   defaultFormatJSON,
+				ExtensionProviders: []*meshconfig.MeshConfig_ExtensionProvider{
```
howardjohn, line 831: can we test that the values are overriden if we have them in extension provider still?

I guess mostly just Path? But you could also have JSON defined in AccessLogEncoding then use non-json in ExtensionProviders

ramaraochavali: Just path I think. If extension provider has log format it s considered - there are already tests for it.

douglas-reid:
> lgtm, cc @douglas-reid to confirm if this is behavior we desire

This seems a bit wonky to me. In the past, we've talked about having a separate "provider" for MeshConfig to allow for transition. It seems somewhat cleaner to go that route than it does to have a provider that pulls its configuration from multiple places.

In other words, wouldn't we prefer, in some sense:
```
providers:
  - name: meshconfig
```
to:
```
providers:
  - name: envoy
```
(where the actual configuration for `envoy` is taken from meshconfig and NOT the `envoy` provider config?)

The other alternative is perhaps to always configure the `envoy` provider to match the MeshConfig configuration. That would lead to some duplication in the short term, but allow for an easy transition.

Thoughts?

costinm:
I also don't like mixing MeshConfig with the new API - and certainly not
adding anything new telemetry-related to MeshConfig - we now have a
dedicated API.

In the absence of a Telemetry config - we should be backward compatible and
respect what is in MeshConfig ( which at some point will need to be
deprecated and we should encourage people to stop using it).

If Telemetry CRD needs a way to specify defaults - I think we have some
patterns in use for this, there are also some mechanisms defined
in the K8S Gateway specs.

douglas-reid:
> If Telemetry CRD needs a way to specify defaults - I think we have some patterns in use for this, there are also some mechanisms defined in the K8S Gateway specs.

This issue, in particular, is about provider config (which is distinct from Telemetry resources). There is a defined, but perhaps underspecified, default behavior when that configuration is not provided (use "default" Istio logging format).

I think it is fine for the "built-in" `envoy` provider to not pull from MeshConfig. There is work to be done to define the transition path. I think creating a synthetic `meshconfig` provider is one mechanism. Having the install values transfer to the `envoy` provider config would be another.

costinm:
You're right, Telemetry doesn't cover the format ( and it's a good thing it
doesn't...).

As both provider and the old config are in MeshConfig - I have no
objections to this PR.

zirain:
> I think it is fine for the "built-in" envoy provider to not pull from MeshConfig.

cause `envoy` is a built-in provider, I prefer this.

hzxuzhonghu:
From the api desc, it is right not to inherite mesh config access log format.

ramaraochavali:
> cause envoy is a built-in provider, I prefer this.

So how do we avoid breaking users who are already using "envoy" provider with the change in behaviour.

My preference is to "Having the install values transfer to the envoy provider config would be another." if we do not want to inherit from meshconfig.

> I think creating a synthetic meshconfig provider is one mechanism

Why another synthetic provider - why not  just make "envoy" behave like that to avoid breaking users?


I think I get the intent of provider inheriting Istio default log if LogFormatter is not specified. That makes sense.  My only point is about "envoy" provider - we need to fix it to avoid breaking users.

ramaraochavali:
Changed to use meshconfig only for "envoy" provider. See if this makes sense.

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -185,13 +187,18 @@ func buildAccessLogFromTelemetry(push *model.PushContext, spec *model.LoggingCon
 		var al *accesslog.AccessLog
 		switch prov := p.Provider.(type) {
 		case *meshconfig.MeshConfig_ExtensionProvider_EnvoyFileAccessLog:
-			al = buildEnvoyFileAccessLogHelper(prov.EnvoyFileAccessLog)
+			// For default provider, fallback to Mesh Config for formatting options.
+			if p.Name == defaultEnvoyAccessLogProvider {
```
zirain, line 191: we need update tests for this?

ramaraochavali: The new test case that I have added, covers this

zirain (commented):
lgtm

ramaraochavali:
> I am not necessarily saying we shouldn't fallback, but its not really breaking?

It is breaking and we were hit by this. We have users using Telemetry API with "envoy" as the provider and the format is controlled at mesh config (because there are some fields we do not to be in access logs). With upgrade, it picked up default format (both type and text) and caused access logs to have fields that we do not want.

howardjohn:
Ah I thought you meant when switching to Telemetry API for the first time. Forgot we had telemetry API before format

douglas-reid:
Since Telemetry is in an Alpha state, I have some mild concerns about trying to preserve behavior from its experimental days as a permanent contract moving forward. We need the flexibility to change behavior, etc., as the API evolves towards a more stable form.

For this particular issue, I worry a bit about future migration away from prior config regimes. At some point, we should want the configuration to live entirely in provider configuration (at least, I assume). I'm not sure if maintaining compat in this fashion will make that easier or harder. We've been burned before by "deprecated" configuration still being active in code paths and leading to unintended, hard to debug, consequences.

It seems the work to transfer the install configuration to the envoy provider configuration will still have to be done in the future.

That being said, it seems like everyone else is OK with this remedy. My slight misgivings shouldn't block this PR.

ramaraochavali:
@douglas-reid thank you. I agree with some of the concerns. I would be in favour of not breaking users(surprises) when they upgrade.

I will let @howardjohn @hzxuzhonghu @zirain decide - I am fine either way.

ramaraochavali:
@howardjohn @hzxuzhonghu WDYT about this? @zirain is [fine](https://github.com/istio/istio/pull/39606#pullrequestreview-1019292647) with this

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -185,13 +187,18 @@ func buildAccessLogFromTelemetry(push *model.PushContext, spec *model.LoggingCon
 		var al *accesslog.AccessLog
 		switch prov := p.Provider.(type) {
 		case *meshconfig.MeshConfig_ExtensionProvider_EnvoyFileAccessLog:
-			al = buildEnvoyFileAccessLogHelper(prov.EnvoyFileAccessLog)
+			// For default provider, fallback to Mesh Config for formatting options.
```
zirain, line 190: this is not `default` provider, `built-in` provider is more precise.

ramaraochavali: yes. makes sense. Updated

ramaraochavali:
@howardjohn @hzxuzhonghu WDYT about this?

hzxuzhonghu (approved):
LGTM, a simple fix

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -185,13 +187,18 @@ func buildAccessLogFromTelemetry(push *model.PushContext, spec *model.LoggingCon
 		var al *accesslog.AccessLog
 		switch prov := p.Provider.(type) {
 		case *meshconfig.MeshConfig_ExtensionProvider_EnvoyFileAccessLog:
-			al = buildEnvoyFileAccessLogHelper(prov.EnvoyFileAccessLog)
+			// For built-in provider, fallback to Mesh Config for formatting options.
+			if p.Name == defaultEnvoyAccessLogProvider {
```
zirain, line 191: should only fallback to `MeshConfig` when `p.Name == "envoy" and prov.EnvoyFileAccessLog.LogFormat==nil` ?
cc @ramaraochavali @douglas-reid

ramaraochavali: Is it the case of user creating access log provider with the same name?

Issue #39619, opened 2022-06-24
Order dependent test:TestListenerAccessLog


howardjohn:
TestListenerAccessLog fails if order is shuffled:

```
go test -vet=off -x ./pilot/pkg/networking/core/v1alpha3 -run TestListenerAccessLog -shuffle=on
```

PR #39624, opened 2022-06-25
fix listener access log test

ramaraochavali:
Fixes https://github.com/istio/istio/issues/39619

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/accesslog_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/accesslog_test.go
@@ -38,7 +38,6 @@ import (
 	"istio.io/istio/pilot/pkg/networking/util"
 	"istio.io/istio/pilot/pkg/serviceregistry/provider"
 	"istio.io/istio/pilot/test/xdstest"
-	"istio.io/istio/pkg/config/mesh"
 	"istio.io/istio/pkg/config/protocol"
 	"istio.io/istio/pkg/test/util/assert"
 	"istio.io/istio/pkg/util/protomarshal"
@@ -104,7 +103,7 @@ func TestListenerAccessLog(t *testing.T) {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			// Update MeshConfig
-			m := mesh.DefaultMeshConfig()
+			m := &meshconfig.MeshConfig{}
 			m.AccessLogFile = "foo"
 			m.AccessLogEncoding = tc.encoding
 			m.AccessLogFormat = tc.format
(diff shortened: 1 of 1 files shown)


pilot/pkg/networking/core/v1alpha3/accesslog_test.go
```diff
@@ -104,7 +103,7 @@ func TestListenerAccessLog(t *testing.T) {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			// Update MeshConfig
-			m := mesh.DefaultMeshConfig()
+			m := &meshconfig.MeshConfig{}
```
howardjohn, line 106: I am not sure I understand this. Can you explain why this fixes it?

It seems like the issue is that we have a shared cache, so maybe we just need `accessLogBuilder.reset()` before the test (or maybe in `buildListeners`, etc)?

ramaraochavali:
@howardjohn @hzxuzhonghu PTAL

pilot/pkg/networking/core/v1alpha3/accesslog_test.go
hzxuzhonghu: IIUC, this is not to mutate the default config. I do not look into the original issue, but this way it makes no harm.

@howardjohn to check whether it fixes original issue?

ramaraochavali: yes. The issue is we are mutating default config in the test which other tests are dependent on.

ramaraochavali: I have run multiple times with shuffle on - it did not fail.

howardjohn: How are we mutating it? DefaultMeshConfig is a new copy.

I know this fixes it but want tounderstand why

ramaraochavali: > DefaultMeshConfig is a new copy.

Ah .. Let me debug a bit a more on how this is working :-)

ramaraochavali: Ok. this is actually not working. When we pass `&meshconfig.MeshConfig{}`, no listeners are constructed as `proxyListenPort` is zero. So it appears to pass but is incorrect. I fixed it now. Basically the accesslog cache  should be reset before buildListener call not after.

howardjohn (approved):

PR #39641, opened 2022-06-27
add delegate virtual services to cache key

ramaraochavali:
We missed to add this here. Made few function renaming and moved them part of struct.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn (approved):
We should introduce a test that mutates each field in the struct and ensures it modifies the cache key.. if possible. may be tough

PR #39645, opened 2022-06-27
hbone: initial echo server/client implementation

howardjohn:
Per https://docs.google.com/document/d/1Ofqtxqzk-c_wn0EgAXjaJXDHB9KhDuLe-W3YGG67Y8g/edit


costinm:
I would really appreciate it if we could have this in istio/pkg repo
instead of istio/istio.

Preferably with a separate go.mod

pkg/config/protocol/instance.go
```diff
@@ -54,6 +54,8 @@ const (
 	Redis Instance = "Redis"
 	// MySQL declares that the port carries MySQL traffic.
 	MySQL Instance = "MySQL"
+	// HBONE declares that the port carries HBONE traffic.
+	HBONE Instance = "HBONE"
```
costinm, line 58: Don't we have a dedicated port ?

costinm: Or 2 ports actually - one for HBone and one for 'plain text/trusted network' hbone, i.e. h2 or h2c

pkg/hbone/README.md
```diff
@@ -0,0 +1,71 @@
+# HTTP Based Overlay Network (HBONE)
+
+HTTP Based Overlay Network (HBONE) is the protocol used by Istio for communication between workloads in the mesh.
+At a high level, the protocol consists of tunneling TCP connections over HTTP/2 CONNECT, over mTLS.
```
costinm, line 4: I would skip 'over mTLS', and add 'over HTTP/2 and H3 CONNECT' ( not clear if QUIC is mTLS, or if in a trusted networks we must have outer mTLS or H2C is allowed )

pkg/hbone/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Specification
+
+TODO
```
costinm, line 8: My doc on gateway improvements has several links, you can add them here ( RFCs, BeyondCorp, browser specs)

pkg/config/protocol/instance.go
howardjohn: This is not something a user can configure by portname on Service -- but used for internal. Not by convention, this is enforced.

pkg/hbone/dialer.go
```diff
@@ -0,0 +1,173 @@
...
+	// handle upstream (mtp server) <-- downstream (app)
+	copyBuffered(pw, conn, log.WithLabels("name", "conn to pipe"))
+
+	wg.Wait()
+	log.Info("stream closed in ", time.Since(t0))
+	return nil
+}
+
+// TLSDialWithDialer is an implementation of tls.DialWithDialer that accepts a generic Dialer
+func TLSDialWithDialer(dialer Dialer, network, addr string, config *tls.Config) (*tls.Conn, error) {
```
costinm, line 139: Why do we need this ? Normal pattern is create Dialer with config and call DialContext/Dial

pkg/hbone/dialer.go
```diff
@@ -0,0 +1,173 @@
...
+	if network != "tcp" {
+		return net.Dial(network, address)
+	}
+	// TODO: use context
+	c, s := net.Pipe()
+	go func() {
+		err := proxyTo(s, d.cfg, address)
+		log.Infof("Tunnel complete: %v", err)
+	}()
+	return c, nil
```
costinm, line 66: I think we need to work on this - see my previous implementation, the HTTP RT should probably be done part of Dial, with proxy remaining in background. So we can return errors properly ( similar with how normal Dial works - i.e. a real connection/stream will be established by dial)

pkg/hbone/dialer.go
howardjohn: I am not sure another way to add TLS to a Dialer?

pkg/hbone/dialer.go
howardjohn: Agreed

costinm:
On Mon, Jun 27, 2022 at 1:55 PM Nathan Mittler ***@***.***>
wrote:

>
> In pkg/config/protocol/instance.go
> <https://github.com/istio/istio/pull/39645#discussion_r907720073>:
>
> > @@ -54,6 +54,8 @@ const (
>  	Redis Instance = "Redis"
>  	// MySQL declares that the port carries MySQL traffic.
>  	MySQL Instance = "MySQL"
> +	// HBONE declares that the port carries HBONE traffic.
> +	HBONE Instance = "HBONE"
>
> Will HBONE always be over h2? If we ever supported h3, would it be
> considered a different protocol?
>
AFAIK we treat H2 and H3 as the same protocol - it really requires
'bidirectional stream support' which is missing in H1.

pkg/config/protocol/instance.go
hzxuzhonghu, line 58: Can we not add it here

howardjohn: We would have to update all of our tests and echo codebase. Any concerns ? It cannot be used by real Service (which I plan to add a comment and test asserting this)

hzxuzhonghu: My  concern is test code should not couple with main. Could be confused at a new protocol?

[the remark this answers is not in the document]
howardjohn: In the future we may make it support h3. TBD if we consider that another protocol here.

This is just internal details here, not API, so we can change

pkg/hbone/dialer.go
```diff
@@ -0,0 +1,173 @@
...
+	"golang.org/x/net/proxy"
+
+	"istio.io/istio/security/pkg/pki/util"
+	istiolog "istio.io/pkg/log"
+)
+
+var log = istiolog.RegisterScope("hbone", "", 0)
+
+type Config struct {
+	ProxyAddress string
```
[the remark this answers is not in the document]
howardjohn: putting it in comments, but ProxyAddress is the HBONE proxy, address is the destination address

pkg/hbone/server.go
```diff
@@ -0,0 +1,77 @@
...
+	// Need to set this to allow timeout on the read header
+	h1 := &http.Transport{
+		ExpectContinueTimeout: 3 * time.Second,
+	}
+	h2, _ := http2.ConfigureTransports(h1)
+	h2.ReadIdleTimeout = 10 * time.Minute // TODO: much larger to support long-lived connections
+	h2.AllowHTTP = true
+	h2.StrictMaxConcurrentStreams = false
+	h2Server := &http2.Server{}
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
howardjohn: `r` and `w` are the standard names for this defined by go stdlib

pkg/hbone/util.go
```diff
@@ -0,0 +1,112 @@
...
+	"net/http"
+	"sync"
+	"time"
+
+	istiolog "istio.io/pkg/log"
+)
+
+// createBuffer to get a buffer. io.Copy uses 32k.
+// experimental use shows ~20k max read with Firefox.
+var bufferPoolCopy = sync.Pool{New: func() interface{} {
```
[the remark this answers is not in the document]
howardjohn: Where are we locking? AFAIK sync.Pool uses no locking internally and I don't see any locking added here?

pkg/config/protocol/instance.go
nmittler: I think "experimental" is more accurate than "test". This only exists in this file. Seems that an adequate "proceed at your own risk" banner should be sufficient, no?

pkg/hbone/README.md
nmittler: For now, isn't it mTLS? I'd rather keep the text representative of what has been built rather than what might be built in the future.

pkg/hbone/README.md
nmittler: +1 ... could you fill out this section?

pkg/hbone/server.go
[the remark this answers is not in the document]
nmittler: doesn't make it clearer :). It's fine ... marking resolved.

pkg/hbone/util.go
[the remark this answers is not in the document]
nmittler: A couple years old, but [hare are some benchmarks](https://medium.com/swlh/go-the-idea-behind-sync-pool-32da5089df72) ... not sure if an extra couple of nanos matters much here :). Go doesn't implement thread-local storage anyway, so you can probably safely ignore my comment.

pkg/test/echo/server/forwarder/config.go
```diff
@@ -55,13 +55,24 @@ type Config struct {
 	urlPath                 string
 	method                  string
 	secure                  bool
+
+	// completed whether the Config is complete or still in progress
+	completed bool
```
nmittler, line 60: why did you need this?

howardjohn: tbh I forget.. some issue about a race condition calling the setup twice. Does seem suspicious, let me look into it

nmittler (approved):

howardjohn:
Figured it out, to newDialer we now pass config which is later mutated causing issues. IMO https://github.com/istio/istio/pull/39735 is a better fix (assuming it doesn't break anything, i split it out to make it more clear).

costinm:
Can we separate the hbone client from the service port name ? I think it's
a very different discussion, better handled in API/doc.

howardjohn:
This doesn't impact the service port - "HBONE" is still not a valid service
port, the protocol enum we are adding to is not implicitly adding it as a
valid option for services

PR #39647, opened 2022-06-27
[tf] Storing deployment templates in files

akshayjnambiar:
The yaml templates for service, echo deployment and vm deployment are copied to files in the template directory. This will allow adding new templates and using it by specifying the DEPLOYMENT_TEMPLATE environment variable.


pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -58,469 +60,37 @@ const (
...
+		return yamlTemplate
 	}
+}
 
-	deploymentTemplate = template.New("echo_deployment")
-	if _, err := deploymentTemplate.Funcs(sprig.TxtFuncMap()).Parse(deploymentYAML); err != nil {
-		panic(fmt.Sprintf("unable to parse echo deployment template: %v", err))
-	}
+func init() {
+	templateDir := path.Join(getCallerDir(), "templates")
```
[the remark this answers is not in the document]
akshayjnambiar: That works when we are running the test from the root directory, but if I try to run a random test in the IDE the cwd is different and I get a file not found error.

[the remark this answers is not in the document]
akshayjnambiar: Used env.IstioSrc to get the repo root dir.

pkg/test/framework/components/echo/kube/templates/echo_deployment.template
```diff
@@ -0,0 +1,213 @@
+{{- $revVerMap := .Revisions }}
```
howardjohn, line 1: nit: name the file with .yaml suffix instead of .template, most IDEs recognize helm templates suffixed with .yaml

pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -669,17 +240,23 @@ func GenerateDeployment(ctx resource.Context, cfg echo.Config, settings *resourc
 		return "", err
 	}
 
-	deploy := deploymentTemplate
+	templateName := "echo_deployment.template"
 	if cfg.DeployAsVM {
-		deploy = vmDeploymentTemplate
+		templateName = "vm_deployment.template"
+	} else if os.Getenv("DEPLOYMENT_TEMPLATE") != "" {
```
nmittler, line 246: Let's make this a flag instead of an environment variable. For completeness, let's add the following:

- `istio.test.echo.kube.template.service`: name of the file to be used for the echo service. If not specified, use the current default service file.
- `istio.test.echo.kube.template.deployment`: name of the file to use by default for non-vms. If not specified, use the current default file name.
- `istio.test.echo.kube.template.deployment.vm`: name of the file to use by default for VMs. If not specified, use the current default file name.

For each of these flags, allow absolute paths to be specified. If the string is not an absolute file path, assume it's relative to the `templates` directory. When we startup, we only have to read the files from the flags, not all files in the directory.

pkg/test/framework/components/echo/kube/templates/echo_deployment.template
nmittler: +1. I've also seen templates use the `.yaml.tmpl` extension, but `yaml` will be more easily understood by IDEs by default.

nmittler: Also, let's just call this file `deployment.yaml`. The `echo` part is already understood.

pkg/test/framework/components/echo/kube/templates/service.template
```diff
@@ -0,0 +1,37 @@
+{{- if .ServiceAccount }}
```
nmittler, line 1: Same comment ... call this `service.yaml`

pkg/test/framework/components/echo/flags.go
```diff
@@ -42,6 +45,12 @@ func init() {
 	flag.IntVar(&callsPerWorkload, "istio.test.echo.callsPerWorkload", callsPerWorkload,
 		"Specifies the number of calls that will be made for each target workload. "+
 			"Only applies if the call count is zero (default) and a target was specified for the call")
+	flag.StringVar(&ServiceTemplateFile, "istio.test.echo.kube.template.service", ServiceTemplateFile,
+		"absolute path or path relative to templates directory of template file to be used for echo services")
```
nmittler, line 49: Maybe something like:

```
Specifies the default template file to be used when generating the
Kubernetes Service for an instance of the echo application. Can be
either an absolute path or relative to the templates directory under 
the echo test component.
```

Similar changes to the other flags.

pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -64,34 +64,53 @@ const (
...
+		serviceYAMLPath = echo.ServiceTemplateFile
+	}
+
+	if serviceYAML, err := os.ReadFile(serviceYAMLPath); err != nil {
+		panic(fmt.Sprintf("unable to read template file %s : %v", serviceYAMLPath, err))
 	} else {
-		yamlTemplate := template.New(file.Name())
-		if _, err1 := yamlTemplate.Funcs(sprig.TxtFuncMap()).Parse(string(yaml)); err1 != nil {
-			panic(fmt.Sprintf("unable to parse template %s : %v", file.Name(), err1))
+		serviceTemplate = template.New("echo_service")
```
nmittler, line 81: Rather than do all of this, let's add a method to https://github.com/istio/istio/blob/master/pkg/test/util/tmpl/parse.go:

```golang
func MustParse(tpl string) *template.Template {
	t.Helper()
	tpl2, err := Parse(tpl)
	if err != nil {
		panic(fmt.Sprintf("tmpl.ParseOrFail: %v", err))
	}
	return tpl2
}
```

Then all of this could be reduced to:

```golang
serviceTemplate := tmpl.MustParse(file.MustAsString("..."))
deploymentTemplate := tmpl.MustParse(file.MustAsString("..."))
vmDeploymentTemplate := tmpl.MustParse(file.MustAsString("..."))
```

pkg/test/framework/components/echo/kube/deployment.go
```diff
@@ -507,20 +70,23 @@ var (
 )
 
 func init() {
-	serviceTemplate = template.New("echo_service")
-	if _, err := serviceTemplate.Funcs(sprig.TxtFuncMap()).Parse(serviceYAML); err != nil {
-		panic(fmt.Sprintf("unable to parse echo service template: %v", err))
+	serviceYAMLPath := path.Join(env.IstioSrc, "pkg/test/framework/components/echo/kube/templates", serviceTemplateFile)
```
nmittler, line 73: make a constant for the path?

pkg/test/framework/components/echo/kube/flags.go
```diff
@@ -0,0 +1,35 @@
...
+import "flag"
+
+var (
+	serviceTemplateFile      = "service.yaml"
+	deploymentTemplateFile   = "deployment.yaml"
+	vmDeploymentTemplateFile = "vm_deployment.yaml"
+)
+
+func init() {
+	flag.StringVar(&serviceTemplateFile, "istio.test.echo.kube.template.service", serviceTemplateFile,
```
nmittler, line 26: nit: indicate that a default will be selected if not specified. Same comment applies to all.

pkg/test/util/tmpl/parse.go
```diff
@@ -37,3 +38,12 @@ func ParseOrFail(t test.Failer, tpl string) *template.Template {
 	}
 	return tpl2
 }
+
+// MustParse calls Parse and panics if it returns error.
+func MustParse(tpl string) *template.Template {
+	tpl2, err := Parse(tpl)
+	if err != nil {
+		panic(fmt.Sprintf("tmpl.ParseOrFail: %v", err))
```
nmittler, line 46: nit: "tmpl.MustParse"

PR #39648, opened 2022-06-27
Refactoring Security Tests

aryan16:
**This PR refactor security tests to support new test framework**


aryan16:
No reviews required as of now.

tests/integration/security/external_ca/main_test.go
```diff
@@ -22,70 +22,27 @@ import (
...
 	// nolint: staticcheck
 	framework.NewSuite(m).
 		Label(label.CustomSetup).
 		RequireMinVersion(19).
-		Setup(istio.Setup(&inst, setupConfig)).
-		Setup(func(ctx resource.Context) error {
-			return SetupApps(ctx, apps)
-		}).
+		Setup(istio.Setup(nil, setupConfig)).
+		Setup(deployment.SetupTwoNamespaces(&apps, deployment.Config{})).
```
howardjohn, line 45: why not one namespace?

howardjohn (approved):

tests/integration/security/external_ca/main_test.go
aryan16: yeah I will update that. I am just testing few different scenarios.

tests/integration/security/ca_custom_root/main_test.go
```diff
@@ -60,27 +51,21 @@ type EchoDeployments struct {
...
 var (
-	inst istio.Instance
-	apps = &EchoDeployments{}
+	inst    istio.Instance
+	apps    deployment.SingleNamespaceView
+	echoApp = &EchoDeployments{}
 )
 
-func SetupApps(ctx resource.Context, apps *EchoDeployments) error {
+func SetupApps(ctx resource.Context, apps deployment.SingleNamespaceView) error {
```
nmittler, line 59: There should no longer be a custom apps in use. The idea is to use the common echo deployment. If the common deployment is missing things that we need here then we should consider adding them. That's what I did for the authz tests.

pkg/test/framework/components/customsetup/setup.go
```diff
@@ -0,0 +1,277 @@
...
+	"istio.io/istio/pkg/test/env"
+	"istio.io/istio/pkg/test/framework/components/echo"
+	"istio.io/istio/pkg/test/framework/components/echo/common/deployment"
+	depl "istio.io/istio/pkg/test/framework/components/echo/deployment"
+	"istio.io/istio/pkg/test/framework/components/echo/match"
+	"istio.io/istio/pkg/test/framework/components/util/cert"
+	"istio.io/istio/pkg/test/framework/resource"
+)
+
+type EchoDeployments struct {
```
nmittler, line 34: Why do we need a custom echo deployment? What is different about this and can we add it to the common one instead?

aryan16: There are some test specific instances with particular naming conventions, it's good if we add that here instead of adding it in main echo deployment? Will use it for other tests as well whenever it will be required.

nmittler: Can we instead change the tests to somehow just use the common echo deployment?

aryan16: okay, will work on this.

aryan16:
@nmittler PTAL

pkg/test/framework/components/echo/common/deployment/echos.go
```diff
@@ -63,7 +63,7 @@ type Config struct {
 	//
 	// Custom echo instances will be accessible from the `All` field in the namespace(s) under which they
 	// were created.
-	Custom []echo.Config
+	Custom echo.CustomGetter
```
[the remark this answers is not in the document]
aryan16: So everytime when we call this from `deployment.SetupSingleNamespace` and `deployment.SetupTwoNamespaces`, we would require to call it this way only otherwise it won't accept it. So IMO it's good to have it like this. We are already doing it for namespace in the same file. Let me know what do you think ?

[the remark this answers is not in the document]
nmittler: I'm not against, TBH ... just curious. I guess it allows you to just pass a common method.

pkg/test/framework/components/echo/common/deployment/echos.go
```diff
@@ -75,18 +75,20 @@ func (c *Config) fillDefaults(ctx resource.Context) error {
...
-	for _, custom := range c.Custom {
-		if custom.Namespace != nil {
-			found := false
-			for _, ns := range c.Namespaces {
-				if custom.Namespace.Name() == ns.Get().Name() {
-					found = true
-					break
+	if c.Custom != nil {
+		for _, custom := range c.Custom.Get() {
+			if custom.Namespace != nil {
```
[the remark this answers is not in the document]
aryan16: We already have this functionality if you don't add any ns then it will automatically deploy it in all the echo ns, if we restrict it to not use any namespace, some tests will fail. Example this test uses a namespace (https://github.com/istio/istio/blob/8cb16ba82cbc044a11632d983ac9ebfb34d88112/tests/integration/security/ca_custom_root/main_test.go#L83) to add a SAN (https://github.com/istio/istio/blob/master/samples/certs/generate-workload.sh#L24). So we would require a namespace while creating the custom deployment sometimes.

nmittler (approved):

PR #39653, opened 2022-06-27
Kind enum

sschepens:

Make ConfigKey.Kind an autogenerated enum with helper methods for coverting to string and from GroupVersionKind.


pilot/pkg/bootstrap/server.go
```diff
@@ -884,7 +885,7 @@ func (s *Server) initRegistryEventHandlers() {
 			pushReq := &model.PushRequest{
 				Full: true,
 				ConfigsUpdated: map[model.ConfigKey]struct{}{{
-					Kind:      curr.GroupVersionKind,
+					Kind:      kind.FromGvk(curr.GroupVersionKind),
```
howardjohn, line 888: This is the risk area of "unknown type". I don't think this _should_ happen and the `panic` should capture any case it does -- hopefully in tests and not in users code.So seems pretty safe

howardjohn (commented):
/test benchmark_istio

pilot/pkg/bootstrap/server.go
sschepens: if panic is not nice I could add some `Unknown` constant and return that

pkg/config/schema/kind/resources.gen.go
```diff
@@ -0,0 +1,222 @@
...
+		return "WorkloadEntry"
+	case WorkloadGroup:
+		return "WorkloadGroup"
+	default:
+		return "Unknown"
+	}
+}
+
+func FromGvk(gvk config.GroupVersionKind) config.Kind {
+	if gvk.Kind == "AuthorizationPolicy" && gvk.Group == "security.istio.io" && gvk.Version == "v1beta1" {
```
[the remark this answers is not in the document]
sschepens: I didn't want to make this package depend on the other generated one, but I could do it if you think it's fine.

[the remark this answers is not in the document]
sschepens: this is only used in `Server.initRegistryEventHandlers`, which is called when the server is created, so I'm guessing it's not a hot path.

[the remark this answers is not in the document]
howardjohn: it is defining a callback function, so the callback is called somewhat often. But "often" is quite small compared to the cost of this function. how it is should be fine.

We can optimize if profiles show issues here (I doubt it)

hzxuzhonghu (commented):
I think it is helpful, will take a closer look after the ci happy

sschepens:
@howardjohn @hzxuzhonghu this is ready for review, let me know if I should add a release note

pilot/pkg/model/config.go
```diff
@@ -50,33 +51,26 @@ func (key NamespacedName) String() string {
...
-		key.Namespace,
-		key.Kind.Kind,
-		key.Kind.Group,
-		key.Kind.Version,
-	} {
-		hash.Write([]byte(v))
-	}
-	var tmp [md5.Size]byte
-	sum := hash.Sum(tmp[:0])
+	hash.Write([]byte(kind.String(key.Kind)))
```
howardjohn, line 61: nit: `byte` and config.Kind are both `uint8`, just cast it?

pilot/pkg/model/config.go
```diff
@@ -50,33 +51,26 @@ func (key NamespacedName) String() string {
...
+	hash.Write([]byte(key.Name))
+	hash.Write([]byte(key.Namespace))
+	sum := hash.Sum(nil)
 	return binary.BigEndian.Uint64(sum)
 }
 
 func (key ConfigKey) String() string {
-	return key.Kind.Kind + "/" + key.Namespace + "/" + key.Name
+	return kind.String(key.Kind) + "/" + key.Namespace + "/" + key.Name
```
howardjohn, line 69: I would make a .String method on `Kind` instead of having a standalone function

howardjohn (commented):
lgtm! just 2 small comments and want to see benchmark result

/test benchmark_istio

pilot/pkg/model/config.go
sschepens: yup, i could add another file in the same package as the generated file with only the declaration, does that seem ok?

I was aiming at leaving the declaration in `config` just to be consistent with `GroupVersionKind`.

howardjohn: yeah I think it would be useful to have the .String(); it makes it so when you print it the string will automatically be used as well

PR #39661, opened 2022-06-28
Make DependentConfigs pointer slice

hzxuzhonghu:


howardjohn (commented):
seems very likely this is a good change but do we have any benchmarks, etc showing how much improvement?

hzxuzhonghu:
The tricky point is there are still many map[ConfigKey]xxx, use the Value as key, which we can not change to a pointer type.

PR #39665, opened 2022-06-28
use config hash for dependent configs

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali:
@hzxuzhonghu @howardjohn @sschepens WDYT about this?

hzxuzhonghu (commented):
I think it is good,  if the collision probability is very low

hzxuzhonghu:
https://www.avira.com/en/blog/md5-the-broken-algorithm#:~:text=MD5%3A%20The%20fastest%20and%20shortest,%3A%201.47*10%2D29.

MD5: The fastest and shortest generated hash (16 bytes). The probability of just two hashes accidentally colliding is approximately: 1.47*10-29.

/lgtm

sschepens:
MD5 is 16bytes but we're storing it in a uint64 which is only 8bytes so we're actually increasing the possibilities of collision.

ramaraochavali:
>  MD5 is 16bytes but we're storing it in a uint64 which is only 8bytes so we're actually increasing the possibilities of collision

I do not think that is big issue given how rare the collisions can be but we change easily change the type if we agree on the direction.

howardjohn:
Why not use the pointer option another PR had? That is simpler, cheaper (no need to hash), and no collision, and same size?

I would really really like to see concrete numbers on this PRs so we don't just have to guess about performance

ramaraochavali:
We all agree that current model is not scalable. I do not have strong opinion on this vs. other PR. My motivation for choosing this approach is sidecar config dependencies is similar to this and it worked there (there are more computations in sidecar code compared to xds cache `Add`). To have similar pattern for same problem.

hzxuzhonghu (approved):
LGTM, in the worst case if two dependent configs hash collides, we will remove a wrong key when clear. 

That does not matter much

PR #39676, opened 2022-06-28
echo: add Java package name and outer class name to be used in grpc java testing

sanjaypujare:
We want to use the echo.proto in grpc java interop tests with Istio. The generated code should go into designated package and outer classname.


howardjohn (approved):

sanjaypujare:
CC @costinm @stevenctl - this is for grpc-java's interop testing with Istio. Note that we import protos like this from their original source repo instead of making copies and modifying them locally. So to the extent possible we would like the original to reflect our requirements

sanjaypujare:
Thanks @howardjohn for the approval. What else needs to happen to finish tests and then merge the PR?

howardjohn:
/ok-to-test

forgot this

PR #39683, opened 2022-06-29
proxyconfig cr merge fix

yokoyang:
Signed-off-by: 奇方 <qifang.ly@alibaba-inc.com>

fix  Issue "Istio ProxyConfig CR still effective after being deleted"  https://github.com/istio/istio/issues/39684

Create the ProxyConfig CR as follows, then restart the pod under the demo namespace, and see from the istio-proxy log that the concurrency is set to 5. Then delete the demo-namespace-proxyconfig or set the spec field to empty. After the demo namespace has no ProxyConfig CR, the expected behavior is that concurrency should revert to the default value of 2. But still 5.

```
kind: ProxyConfig
metadata:
  name: demo-namespace-proxyconfig
  namespace: demo
spec:
  concurrency: 5
```

Troubleshooting found that the following code of EffectiveProxyConfig will overwrite mc.DefaultConfig after execution, because mergeWithPrecedence uses pointer copying when merging. After the merge ends, it will affect the incoming mc.DefaultConfig. After deleting the ProxyConfig CR or setting the field to empty, the value in mc.DefaultConfig is set to the last value, which is 5 in the example. 

```
// Merge the proxy config from default config.
effectiveProxyConfig = mergeWithPrecedence(mc.GetDefaultConfig(), effectiveProxyConfig)
```
pilot/pkg/model/proxy_config.go

```
		if pcs[i].GetConcurrency() != nil {
			merged.Concurrency = pcs[i].GetConcurrency()
		}
		if pcs[i].GetImage() != nil {
			merged.Image = pcs[i].GetImage()
		}
```


yokoyang:
> can we add some tests?

Thanks, I added some ut. If the code in proxy_config.go is not modified, the ut will fails.

zirain:
please run `make gen` locally to fix CI.

yokoyang:
why integ-security-multicluster_istio ut not pass?

yokoyang:
> can we add some tests?

please review again, thx

pilot/pkg/model/proxy_config.go
```diff
@@ -134,10 +135,12 @@ func mergeWithPrecedence(pcs ...*meshconfig.ProxyConfig) *meshconfig.ProxyConfig
...
 		proto.Merge(merged, pcs[i])
 		if pcs[i].GetConcurrency() != nil {
-			merged.Concurrency = pcs[i].GetConcurrency()
+			pcConcurrency := pcs[i].GetConcurrency()
+			merged.Concurrency = wrapperspb.Int32(pcConcurrency.Value)
 		}
 		if pcs[i].GetImage() != nil {
-			merged.Image = pcs[i].GetImage()
+			pcImage := pcs[i].GetImage()
+			merged.Image = &v1beta1.ProxyImage{ImageType: pcImage.ImageType}
```
howardjohn, line 143:
```suggestion
			merged.Image = proto.Clone(pcs[i].GetImage()).(*v1beta1.ProxyImage)
```

I am worried about us adding fields in the future an missing thing. We can do the same on L139 as well

howardjohn:
sorry I totally forgot about this and merged my own in https://github.com/istio/istio/pull/40494.. My apologies

Issue #39623, opened 2022-06-25
reorg Telemetry e2e tests


zirain:
logging e2e mixed with other tests in `stats/prometheus`, it's odd, let's do something to improve it.

howardjohn:
One thing to note is that each new test suite adds 2-3minutes. So it's recommended to only split them if they have different install values. 

We can rename it of course

zirain:
there's not existing suite for logging and WasmPlugin, those tests are put in `stats/prometheus/nullvm/accesslogs_test.go` and `stats/prometheus/wasm/imagepullpolicy_test.go`

we may need two suites for those, but that will cost more time for e2e tests.

howardjohn:
The only reason to make more suites is if they need different Istio
installations. Unlikely normal packages it's not just organization - there
is a huge cost. If there are distinct features they can be split into
files, functions, etc.

Ideally there would be one suite

zirain:
https://github.com/istio/istio/pull/39687#issuecomment-1197526728

reorg telemetry e2e tests:
```
telemetry/
|-- v2/
|-- api/
|-- policy/
```

PR #39687, opened 2022-06-29
[e2e]reorg telemetry tests

zirain:

part of https://github.com/istio/istio/issues/39623

- move most of tests to `common` folder
- move utilities  to package `tests/integration/telemetry/util`


zirain:
cc @douglas-reid PTAL

tests/integration/telemetry/common/wasm/maint_test.go
```diff
@@ -23,7 +23,7 @@ import (
 
 	"istio.io/istio/pkg/test/framework/components/registryredirector"
 	"istio.io/istio/pkg/test/framework/resource"
-	common "istio.io/istio/tests/integration/telemetry/stats/prometheus"
+	"istio.io/istio/tests/integration/telemetry/common"
```
douglas-reid, line 26: is the file name starting with `maint` intentional?  should it be `main` or does the `t` signify anything?

tests/integration/telemetry/util/util.go
```diff
@@ -15,19 +15,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package telemetry
+package util
```
douglas-reid, line 18: i'm curious: why refactor this package (or part of the package)?  `telemetry` is a more descriptive name.  can this be combined with the `common` stuff to make a single `telemetry` package?

tests/integration/telemetry/util/util.go
```diff
@@ -99,6 +103,43 @@ func PromDump(cluster cluster.Cluster, prometheus prometheus.Instance, query pro
...
+	got, err := prometheus.Sum(val)
+	if err != nil {
+		t.Logf("value: %s", val.String())
+		return "", fmt.Errorf("could not find metric value: %v", err)
+	}
+	t.Logf("get value %v", got)
+	return val.String(), nil
+}
+
+func ValidateMetric(t framework.TestContext, cluster cluster.Cluster, prometheus prometheus.Instance, query prometheus.Query, want float64) {
```
douglas-reid, line 124: with this no longer in `prometheus`, I suggest we update the name to be `ValidatePrometheusMetric`. Thoughts?

douglas-reid (commented):
Would you mind providing some description of the reasoning and choices? I'm trying to understand the end goal better.

My initial reaction is that we should either have some structure based around telemetry backend like:

```
telemetry/
|-- prometheus/
|-- stackdriver/
|-- grafana/
|-- envoy/
```

with `outboundtrafficpolicy` and `wasm` pulled into separate hierarchies.

OR maybe allowing room to separate the other tests:

```
telemetry/
|-- accesslogs/
|-- metrics/
   |-- prometheus/
   |-- stackdriver/
|-- dashboard/
```

And within the tests, we could collapse the `api` and `envoyfilter` tests into a single suite that can run both (may be complicated by adding/removing the filters)?

Or it might be easier to collapse stackdriver and prometheus tests into a single suite, and have the production methodology change across packages, like:

```
telemetry/
|-- accesslogs/
   |-- api/
   |-- envoyfilter/
|-- metrics/
   |-- api/
   |-- envoyfilter/
|-- dashboard/
```
Thoughts?

douglas-reid:
I just recalled the original objection.  WDYT about an org structure like:

```
telemetry/
|-- api/
|-- envoyfilter/
|-- utils/
     |-- prometheus/
     |-- stackdriver/
     ...
```

That would give us just two suites (`api` and `envoyfilter`) and then a set of reusable endpoint-specific libraries (and maybe a `common`) ?

zirain:
the original thought is that acesslogging and wasm e2e tests located in folder named `stats/prometheus/*`

IMO, prefer to reorg to something like:
```
telemetry/
|-- accesslogs/
|-- tracing/
|-- metrics/
   |-- prometheus/
   |-- stackdriver/
|-- dashboard/
```

and we merge test suites to reduce time cost.

zirain:
> I just recalled the original objection. WDYT about an org structure like:
> 
> ```
> telemetry/
> |-- api/
> |-- envoyfilter/
> |-- utils/
>      |-- prometheus/
>      |-- stackdriver/
>      ...
> ```
> 
> That would give us just two suites (`api` and `envoyfilter`) and then a set of reusable endpoint-specific libraries (and maybe a `common`) ?

what's `envoyfilter` stand for? policy(like ratelimit)?

douglas-reid:
> what's `envoyfilter` stand for? policy(like ratelimit)?

`envoyfilter` is perhaps the wrong term. i was trying to think of something that captures the non-Telemetry API way of configuring telemetry behavior.

It seems like we can have one test suite that is Telemetry API controlled (no envoyfilters for telemetry, etc.) and one that is the v2-way of configuration (envoyfilters + meshconfig) and then a bunch of sub tests in each suite that exercise all of the telemetry features (logging, monitoring, tracing).  The separation of suites is necessary only because the envoyfilter configured telemetry conflicts with the Telemetry API way of configuration.

Policy and Wasm can live in one or both suites, I suppose. However, it feels like policy, at least, really should live in another suite (like maybe one of the security suites).

zirain:
> > what's `envoyfilter` stand for? policy(like ratelimit)?
> 
> `envoyfilter` is perhaps the wrong term. i was trying to think of something that captures the non-Telemetry API way of configuring telemetry behavior.
> 
> It seems like we can have one test suite that is Telemetry API controlled (no envoyfilters for telemetry, etc.) and one that is the v2-way of configuration (envoyfilters + meshconfig) and then a bunch of sub tests in each suite that exercise all of the telemetry features (logging, monitoring, tracing). The separation of suites is necessary only because the envoyfilter configured telemetry conflicts with the Telemetry API way of configuration.
> 
> Policy and Wasm can live in one or both suites, I suppose. However, it feels like policy, at least, really should live in another suite (like maybe one of the security suites).

so it would looks like 

```
telemetry/
|-- v2/
|-- api/
|-- policy/
```

does statckdrive need a stand-alone suite?

douglas-reid:
> does statckdrive need a stand-alone suite?

🤷‍♂️ i don't _think_ so. It should be able to run along side prometheus with no loss of functionality.

zirain:
replaced by https://github.com/istio/istio/pull/40799

PR #39689, opened 2022-06-29
make benchmark to align with real case

hzxuzhonghu:

Make the 100 svc + 100 dr compose a rds, which is to simulate more real case


hzxuzhonghu:
@ramaraochavali

ramaraochavali:
Is it possible to add delegates as well?

howardjohn (approved):

hzxuzhonghu:
Delegates can be added, it is just adding more dependencies. The effect is same as service/dr

PR #39690, opened 2022-06-29
Refact authenticator

hzxuzhonghu:

It is to merge the two different methods of `Authenticator` and wrapped a AuthenticateRequest which can contain HTTP request or grpc context.

### DIFF
--- a/pkg/security/security.go
+++ b/pkg/security/security.go
@@ -348,7 +348,7 @@ const (
 	authorizationMeta = "authorization"
 )
 
-type AuthenticateContext struct {
+type AuthContext struct {
 	// grpc context
 	GrpcContext context.Context
 	// http request
@@ -363,7 +363,7 @@ type Caller struct {
 
 // Authenticator determines the caller identity based on request context.
 type Authenticator interface {
-	Authenticate(ctx AuthenticateContext) (*Caller, error)
+	Authenticate(ctx AuthContext) (*Caller, error)
 	AuthenticatorType() string
 }
 
@@ -376,7 +376,7 @@ type AuthenticationManager struct {
 
 // Authenticate loops through all the configured Authenticators and returns if one of the authenticator succeeds.
 func (am *AuthenticationManager) Authenticate(ctx context.Context) *Caller {
-	req := AuthenticateContext{GrpcContext: ctx}
+	req := AuthContext{GrpcContext: ctx}
 	for _, authn := range am.Authenticators {
 		u, err := authn.Authenticate(req)
 		if u != nil && len(u.Identities) > 0 && err == nil {
--- a/pkg/security/security.go
+++ b/pkg/security/security.go
@@ -348,6 +348,13 @@ const (
 	authorizationMeta = "authorization"
 )
 
+type AuthenticateContext struct {
+	// grpc context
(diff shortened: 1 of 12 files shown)


ramaraochavali:
Let us first get #39405 to a stage then this can be made on top of it as the direction there is to  use `AuthContext` as parameter and that can have both request and ctx - merging would be easy?

costinm (commented):
If we are refactoring auth - I would use AuthContext instead of AuthRequest for the struct, maybe add a method to return the certs and JWT - to avoid the switch.

And add few more fields to capture the context - like in the first XFCC PR, I think that was
a very good idea and we'll need it to validate XFCC-trusted gateways that authenticate with mTLS or JWTs.

Since I don't see XFCC code - it may be better to wait for that to merge first.

hzxuzhonghu:
> maybe add a method to return the certs and JWT - to avoid the switch.

I would suggest leaving the custom method to specific authneticator. Others sgtm

security/pkg/server/ca/authenticate/cert_authenticator.go
```diff
@@ -43,7 +43,17 @@ func (cca *ClientCertAuthenticator) AuthenticatorType() string {
 // method assumes that certificate chain has been properly validated before
 // this method is called. In other words, this method does not do certificate
 // chain validation itself.
-func (cca *ClientCertAuthenticator) Authenticate(ctx context.Context) (*security.Caller, error) {
+func (cca *ClientCertAuthenticator) Authenticate(authReq security.AuthenticateContext) (*security.Caller, error) {
```
ramaraochavali, line 46: nit authReq -> authCtx. Same in other places

security/pkg/server/ca/authenticate/cert_authenticator.go
```diff
@@ -43,7 +43,17 @@ func (cca *ClientCertAuthenticator) AuthenticatorType() string {
...
+	if authReq.GrpcContext != nil {
```
ramaraochavali, line 47: It would be nice if there is a way to avoid if else in every authenticator - but I guess, there is no easy way.

ramaraochavali (commented):
LGTM,. some minor comments

security/pkg/server/ca/authenticate/cert_authenticator.go
hzxuzhonghu: Yes

pkg/security/security.go
```diff
@@ -348,6 +348,13 @@ const (
 	authorizationMeta = "authorization"
 )
 
+type AuthenticateContext struct {
```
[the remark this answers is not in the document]
hzxuzhonghu: SGTM

security/pkg/server/ca/authenticate/cert_authenticator.go
hzxuzhonghu: SG

hzxuzhonghu:
Fixed

ramaraochavali (approved):

hzxuzhonghu:
@costinm ptal

costinm (approved):

PR #39696, opened 2022-06-29
bugfix(accesslog): allow different provider between modes

j2gg0s:

Take workload's mode into consideration, when compute inScopeProviders.
So, we can allow user config different access log providers for different mode.

This maybe what we expected.

### DIFF
--- a/pilot/pkg/model/telemetry.go
+++ b/pilot/pkg/model/telemetry.go
@@ -463,7 +463,7 @@ func mergeLogs(logs []*tpb.AccessLogging, mesh *meshconfig.MeshConfig, mode tpb.
 	providerNames := mesh.GetDefaultProviders().GetAccessLogging()
 	for _, m := range logs {
 		names := getProviderNames(m.Providers)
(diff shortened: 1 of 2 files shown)


zirain:
this seems related to https://github.com/istio/istio/pull/39696

j2gg0s:
> this seems related to #39521 

Yes, it is related but not same. This PR don't change current logic and just fix obvious unexpected behavior.

zirain:
> > this seems related to #39521
> 
> Yes, it is related but not same. This PR don't change current logic and just fix obvious unexpected behavior.

I think the root case is `multi accesslogging not working`, let's hold this first, WDYK?

j2gg0s:
> 

Yes, two PR is conflict

But if we want to change Telemetry's AccessLogging behavior, Maybe it's better to define what we expected before coding.

one Telemetry has many AccessLoging，one HCM also has many access_log.
one AccessLogging has many providers, how we map this to envoy's HCM?

PR #39699, opened 2022-06-29
Make istio-cni compatible with Talos Linux (go netns implementation + added ca cert path)

### LINKED ISSUE #38794: After Istio CNI Plugin installation, Cilium CNI is broken and pod's can't be started
### Bug Description

Hi, is Istio CNI plugin compatible with Cilium kube-proxy free setup?

I ve enabled [DNS proxying ](https://istio.io/latest/docs/ops/configuration/traffic-management/dns-proxy/) and after that all my pods are starting to fail with error    `Failed to execute: iptables-restore --noflush errors.` ; in istio-init containers.

Then I tried to enable CNI plugin via --set components.cni.enabled=true . I could see that CNI daemonset is running on all the nodes. However, whenever I create a pod with injection enabled/disabled, pods are failing with 

```
Warning  FailedCreatePodSandBox  1s    kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "1d364156d98e1eb7f6ab03765b8fff02474ab17dd8c139e67b015ad08ee06f2a": plugin type="istio-cni" name="istio-cni" failed (add): netplugin failed but error parsing its diagnostic message "2022-05-08T19:56:33.418196Z\twarn\tOS CA Cert could not be found for agent\n": invalid character '-' after top-level value

or 

Warning  FailedCreatePodSandBox  2s    kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "7ec243813ed979c2799a28160a4d9c4176c5ad8183ce5f7a79eb4bcd6ba4816c": plugin type="istio-cni" name="istio-cni" failed (add): decoding version from network config: invalid character '-' after top-level value

```

Did anyone have the same issue before? 

Setup:

Vanilla k8s with cilium installed via `helm install cilium cilium/cilium     --version 1.11.4     --namespace kube-system     --set ipam.mode=kubernetes     --set kubeProxyReplacement=strict     --set k8sServiceHost="my.api.com"     --set k8sServicePort="6443"     --set hostServices.hostNamespaceOnly=true`

Thank you

### Version

```prose
istioctl version
client version: 1.13.3
control plane version: 1.13.3
data plane version: 1.13.3 (1 proxies)

kubectl version --short
Client Version: v1.22.1
Server Version: v1.23.6
```

### Additional Information

Istio CNI pod logs - https://paste.opendev.org/show/bXCRYC98CGoUIiLXKu01/

nberlee:
Talos Linux is a purpose build K8s distro is shipped without nsenter, and their ca cert path is different.

This PR adds the correct ca-certificates path for Talos Linux and implements netns in go. This removes the dependency for nsenter entirely.

closes #38794

### DIFF
--- a/tools/istio-iptables/pkg/dependencies/implementation.go
+++ b/tools/istio-iptables/pkg/dependencies/implementation.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
+	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/spf13/viper"
 
 	"istio.io/istio/pkg/util/sets"
@@ -67,16 +68,18 @@ var XTablesCmds = sets.New(
 // RealDependencies implementation of interface Dependencies, which is used in production
 type RealDependencies struct {
 	NetworkNamespace string
+	HostNSEnterExec  bool
 	CNIMode          bool
 }
 
 func (r *RealDependencies) execute(cmd string, ignoreErrors bool, args ...string) error {
-	if r.CNIMode {
+	if r.CNIMode && r.HostNSEnterExec {
 		originalCmd := cmd
 		cmd = constants.NSENTER
 		args = append([]string{fmt.Sprintf("--net=%v", r.NetworkNamespace), "--", originalCmd}, args...)
 	}
 	log.Infof("Running command: %s %s", cmd, strings.Join(args, " "))
+
 	externalCommand := exec.Command(cmd, args...)
 	stdout := &bytes.Buffer{}
 	stderr := &bytes.Buffer{}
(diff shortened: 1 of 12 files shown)


howardjohn (commented):
/ok-to-test

tools/istio-iptables/pkg/dependencies/implementation.go
```diff
@@ -71,11 +73,28 @@ type RealDependencies struct {
...
+	}
+	defer ns.Close()
+
+	origns, err := netns.Get()
+	if err != nil {
+		return err
 	}
+	defer origns.Close()
+
+	err = netns.Set(ns)
```
howardjohn, line 92: should we use `Do` to avoid the complex sequence here: https://github.com/containernetworking/plugins/blob/main/pkg/ns/README.md#do-the-recommended-thing? It seems recommended

tools/istio-iptables/pkg/dependencies/implementation.go
```diff
@@ -71,11 +73,28 @@ type RealDependencies struct {
 }
 
 func (r *RealDependencies) execute(cmd string, ignoreErrors bool, args ...string) error {
-	if r.CNIMode {
```
howardjohn, line 74: Given the risk here (see linked doc, its not always safe to do this - although I suspect here it is), we may want to add a feature flag to allow disabling this if it causes issues on some platforms, etc

tools/istio-iptables/pkg/dependencies/implementation.go
nberlee: Yes, makes perfect sense. will do

tools/istio-iptables/pkg/dependencies/implementation.go
nberlee: ok

nberlee:
@howardjohn
please review again. 

the err2 was my only solution to pass the linter, otherwise it would fail on line 99 with `ineffectual assignment to err (ineffassign)`

howardjohn (commented):
LGTM but we had mentioned a feature flag would be added to preserve the old behavior. can we add that? It can be off by default (default to using the go netns)

nberlee:
Ok, i have added a `HostNSEnterExec` flag you can set in the CNI configmap.

howardjohn (approved):

PR #39709, opened 2022-06-30
minor renaming of methods to avoid confusion in cluster patcher

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -244,14 +244,14 @@ func (configgen *ConfigGeneratorImpl) buildOutboundClusters(cb *ClusterBuilder, 
 			subsetClusters := cb.applyDestinationRule(defaultCluster, DefaultClusterMode, service, port,
 				clusterKey.proxyView, clusterKey.destinationRule, clusterKey.serviceAccounts)
 
-			if patched := cp.applyResource(nil, defaultCluster.build()); patched != nil {
+			if patched := cp.patch(nil, defaultCluster.build()); patched != nil {
 				resources = append(resources, patched)
 				if features.EnableCDSCaching {
 					cb.cache.Add(clusterKey, cb.req, patched)
 				}
 			}
 			for _, ss := range subsetClusters {
-				if patched := cp.applyResource(nil, ss); patched != nil {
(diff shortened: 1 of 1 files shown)


pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -271,15 +271,15 @@ type clusterPatcher struct {
...
+	cluster := p.tryPatching(hosts, c)
 	if cluster == nil {
 		return nil
 	}
 	return &discovery.Resource{Name: cluster.Name, Resource: util.MessageToAny(cluster)}
 }
 
-func (p clusterPatcher) apply(hosts []host.Name, c *cluster.Cluster) *cluster.Cluster {
+func (p clusterPatcher) tryPatching(hosts []host.Name, c *cluster.Cluster) *cluster.Cluster {
 	if !envoyfilter.ShouldKeepCluster(p.pctx, p.efw, c, hosts) {
```
zirain, line 283: IMO, a `try` function should a `bool` return value?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -271,15 +271,15 @@ type clusterPatcher struct {
...
+func (p clusterPatcher) patch(hosts []host.Name, c *cluster.Cluster) *discovery.Resource {
+	cluster := p.tryPatching(hosts, c)
 	if cluster == nil {
 		return nil
 	}
 	return &discovery.Resource{Name: cluster.Name, Resource: util.MessageToAny(cluster)}
 }
 
-func (p clusterPatcher) apply(hosts []host.Name, c *cluster.Cluster) *cluster.Cluster {
+func (p clusterPatcher) tryPatching(hosts []host.Name, c *cluster.Cluster) *cluster.Cluster {
```
zirain, line 282: IMO, a `try` function should a `bool` return value?

ramaraochavali: I am not aware about that - here the usage it tries to patch  if it patches it returns the patched cluster otherwise nil. Open to other naming if you have suggestions

PR #39713, opened 2022-06-30
XDS Cache Tests and minor improvements

sschepens:
Adding tests for XdsCache.

Also add some minor improvements:
- use ReadLock on Get, we don't do any writing so it should be ok, can improve Get throughput
- avoid using evict function to clear indexes when adding an already existant item, to avoid generating more eviction stats

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/xds_cache_test.go
+++ b/pilot/pkg/model/xds_cache_test.go
@@ -1,0 +1,424 @@
+package model
+
+import (
+	"testing"
+	"time"
+
+	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	"istio.io/istio/pilot/pkg/features"
+	"istio.io/istio/pkg/config"
+	"istio.io/istio/pkg/config/schema/gvk"
+	"istio.io/istio/pkg/test"
+	"istio.io/istio/pkg/test/util/assert"
+	"istio.io/istio/pkg/util/sets"
+)
+
+type entry struct {
+	key              string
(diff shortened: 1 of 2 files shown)


pilot/pkg/model/xds_cache_test.go
```diff
@@ -0,0 +1,424 @@
+package model
```
ramaraochavali, line 15: Can you please move these tests here https://github.com/istio/istio/blob/master/pilot/pkg/xds/xds_cache_test.go? or did you put them here for any specific reason?

sschepens: I put them here because I need to be in the same package and access the indexes which are internal fields

sschepens:
well it appears LRU is not thread safe, so we have to acquire a write lock on gets :(

adiprerepa (commented):
good but the direction of the xds cache seems to be https://github.com/istio/istio/pull/39731 -- might want to take that into consideration

howardjohn (approved):

sschepens:
@ramaraochavali @howardjohn do you think we can merge this now? these tests correctly detected the leak that @ramaraochavali and will help us stop regressing.

ramaraochavali:
@sschepens Yes. Thank you. @howardjohn has already approved. It should be merged now.

PR #39133, opened 2022-05-25
Pilot/use consistent sa view

howardjohn:
**Please provide a description of this PR:**

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #39721, opened 2022-06-30
Revert changes to always notify XdsUpdater in mem discovery

howardjohn:
This impacts tests only.
In https://github.com/istio/istio/pull/39133 we changed to notify on
writes; this ended up triggering a ton of test flake. Instead, just add
explicit functions to notify (new test needs them) and revert the other changes

### DIFF
--- a/pilot/pkg/serviceregistry/memory/discovery.go
+++ b/pilot/pkg/serviceregistry/memory/discovery.go
@@ -129,32 +129,31 @@ func (sd *ServiceDiscovery) AddService(svc *model.Service) {
 	svc.Attributes.ServiceRegistry = provider.Mock
 	sd.services[svc.Hostname] = svc
 	sd.mutex.Unlock()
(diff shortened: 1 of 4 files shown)


nmittler:
@howardjohn do we understand the cause of the flakes? Is this something we'll want to fix and roll-forward later?

howardjohn:
basically tests do "apply some changes, trigger push, wait for push". But
with my change to always trigger push when there are changes , depending on
debounce they may get a different number of pushes.

It's possible to make the tests more robust but there are a lot like this
and they are somewhat complex to fix so I at least don't have bandwidth to
fix in the short term

howardjohn:
depending on the test it's also reasonable to want to explicitly control
the push to test certain things

hzxuzhonghu (approved):

adiprerepa (approved):

PR #39725, opened 2022-06-30
added imagePullSecret in extauthz and jwt server deployment

aryan16:
**Updated the deployment of external authz and jwt-server by adding ImagePullSecrets **

### DIFF
--- a/pkg/test/framework/components/authz/kube.go
+++ b/pkg/test/framework/components/authz/kube.go
@@ -34,6 +34,7 @@ import (
 	"istio.io/istio/pkg/test/kube"
 	"istio.io/istio/pkg/test/scopes"
 	"istio.io/istio/pkg/test/util/tmpl"
+	"istio.io/istio/pkg/test/util/yml"
 	"istio.io/istio/pkg/util/protomarshal"
 )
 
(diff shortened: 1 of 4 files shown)


stevenctl (approved):

pkg/test/framework/components/authz/server.go
```diff
@@ -95,3 +123,37 @@ func SetupLocal(server *Server, ns namespace.Getter) resource.SetupFn {
 		return err
 	}
 }
+
+func AddPullSecret(deploymentYaml string, pullSecret string) (string, error) {
```
nmittler, line 127: Could we just move this to the `yml` package? Seems like it might be generally useful? It's conceptually similar to `yml.ApplyNamespace`.

pkg/test/framework/components/authz/server.go
```diff
@@ -28,6 +35,27 @@ type Server interface {
 	Providers() []Provider
 }
 
+type deploymentMergeSpec struct {
```
nmittler, line 38: why not just use an actual k8s spec? You can just parse the yaml into an actual k8s `Deployment`, make the changes, and then re-marshal to yaml. No merge-patch necessary.

ericvn (approved):
samples update lgtm

howardjohn (approved):

nmittler (approved):

PR #39731, opened 2022-07-01
Reduce `[]ConfigKey` slice allocate numbers during xds caching

hzxuzhonghu:

This is making use  of sync.Pool to relieve pressure on the garbage collector.


hzxuzhonghu:
Only need to look at the latest commit https://github.com/istio/istio/pull/39731/commits/0a6f271b2cce392994f4a8dd5b6b22c254a7bd1b

pilot/pkg/model/xds_cache.go
```diff
@@ -164,20 +188,67 @@ func newLru(evictCallback simplelru.EvictCallback) simplelru.LRUCache {
...
+}
+
+// clearEvicted is to clear configIndex and typesIndex based on evictedKeys.gi
+func (l *LruCache) clearEvicted() {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	// In order to save cpu, if the evicted keys is very small, we can just ignore
+	// TODO: make it configurable
+	if len(l.evictedKeys) < 100 {
```
adiprerepa, line 225: is the optimization worth the inaccuracy? I think things like delta will depend on whether or not entries in the cache are still valid. We could just check the cache AND evictedKeys though -- @howardjohn

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -177,8 +179,13 @@ func (b EndpointBuilder) Cacheable() bool {
 	return b.service != nil
 }
 
-func (b EndpointBuilder) DependentConfigs() []model.ConfigKey {
-	configs := []model.ConfigKey{}
+func (b EndpointBuilder) DependentConfigs(allocator *model.Allocator) []model.ConfigKey {
+	var configs []model.ConfigKey
+	if allocator != nil {
+		configs = allocator.Allocate(2)
```
adiprerepa, line 185: why 2 here?

pilot/pkg/model/xds_cache.go
```diff
@@ -164,20 +188,67 @@ func newLru(evictCallback simplelru.EvictCallback) simplelru.LRUCache {
...
-func (l *lruCache) evict(k interface{}, v interface{}) {
+// This is the callback passed to LRU, it will be called whenever a key is removed.
+func (l *LruCache) onEvict(k interface{}, _ interface{}) {
 	if features.EnableXDSCacheMetrics {
 		xdsCacheEvictions.Increment()
 	}
 
-	key := k.(string)
-	value := v.(cacheValue)
+	if l.trackEvicted {
```
adiprerepa, line 197: in what scenario do we not want to track evicted? do we just not evict if this is not set?

adiprerepa (commented):
optimization should help a lot

adiprerepa:
@hzxuzhonghu should https://github.com/istio/istio/pull/39688 be closed in favor of this?

hzxuzhonghu:
I am ok unless some others not want to couple them.

But the commit is largely dependent on previous pr.

pilot/pkg/xds/endpoint_builder.go
hzxuzhonghu: At most 2 is needed for EndpointBuilder

pilot/pkg/model/xds_cache.go
hzxuzhonghu: Actively `Remove`

pilot/pkg/model/xds_cache.go
hzxuzhonghu: Not sure i understand

hzxuzhonghu: This can be seen as a debouncer, hope you can better understand

howardjohn (commented):
As mentioned before, I don't feel comfortable merging these types of performance changes, which may or may not bring improvement, without numbers actually showing differences.

Currently our benchmarks only show a regression from this change, so if the are not giving accurate coverage then please update them.

hzxuzhonghu:
i have to say can we be fair?  addind dependentconfigs to cache value does not improve any perf, but on opposite mem costing.. i just reverted the bad stuff in #39688

PR #39737, opened 2022-07-01
tf: enable core dumps even if the pod has restarted

howardjohn:
Currently our logic is too restrictive, making it so we will only grab
core dumps if the container hasn't restarted yet.

Example miss:
https://prow.istio.io/view/gs/istio-prow/logs/integ-security-multicluster_istio_postsubmit/1542617360051998720


ericvn (approved):
Thanks for looking into the security-multi cluster flakiness

Issue #39736, opened 2022-07-01
CDS ACK ERR


howardjohn:
`ADS:CDS: ACK ERROR b-v1-6b468b6c78-tkl5k.echo2-53 Internal:Error adding/updating cluster(s) outbound|8888||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|8882||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|80||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|7070||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|85||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|9090||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|443||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|9091||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|9092||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|9093||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|81||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|7071||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|9443||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|82||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|84||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured, outbound|86||fake.external.com: ORIGINAL_DST clusters must have no load assignment configured`

reproducing by running TestReachability

PR #39745, opened 2022-07-01
cluster: fix NACK when using STATIC Service with PASSTRHROUGH

howardjohn:
Fixes https://github.com/istio/istio/issues/39736

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -714,6 +714,8 @@ func applyLoadBalancer(c *cluster.Cluster, lb *networking.LoadBalancerSettings, 
 	case networking.LoadBalancerSettings_PASSTHROUGH:
 		c.LbPolicy = cluster.Cluster_CLUSTER_PROVIDED
 		c.ClusterDiscoveryType = &cluster.Cluster_Type{Type: cluster.Cluster_ORIGINAL_DST}
+		// Wipe out any LoadAssignment, if set. This can occur when we have a STATIC Service but PASSTHROUGH traffic policy
+		c.LoadAssignment = nil
 	default:
 		applySimpleDefaultLoadBalancer(c, lb)
 	}
--- a/pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster_builder_test.go
@@ -151,6 +151,22 @@ func TestApplyDestinationRule(t *testing.T) {
 			},
 		},
 		{
+			name:        "destination rule static with pass",
+			cluster:     &cluster.Cluster{Name: "foo", ClusterDiscoveryType: &cluster.Cluster_Type{Type: cluster.Cluster_STATIC}, LoadAssignment: &endpoint.ClusterLoadAssignment{}},
+			clusterMode: DefaultClusterMode,
+			service:     service,
+			port:        servicePort[0],
+			proxyView:   model.ProxyViewAll,
+			destRule: &networking.DestinationRule{
+				Host: "foo.default.svc.cluster.local",
+				TrafficPolicy: &networking.TrafficPolicy{
+					LoadBalancer: &networking.LoadBalancerSettings{
+						LbPolicy: &networking.LoadBalancerSettings_Simple{Simple: networking.LoadBalancerSettings_PASSTHROUGH},
+					},
+				},
+			},
+		},
+		{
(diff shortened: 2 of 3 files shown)


nmittler (approved):

howardjohn:
https://github.com/istio/istio/pull/39757 seems similar issue but very different solutions... do we need both @ramaraochavali ?

ericvn (approved):
Approve the release note.

ramaraochavali:
@howardjohn I thought about setting clusterLoadAssignments to nil while fixing it in other PR. But I think this seems a config issue - for service entries with Resolution as STATIC/DNS - some one is setting `PASSTHROUGH` Load balancer. By removing the cluster load assignments, we are allowing this cluster to route traffic to any IP that client chooses which I felt incorrect. So I went with ignoring DR and setting default LB model. If some one wants PASSTHROUGH why not set resolution `None`.  This happened in this case probably because of wildcard host in DR - So I am bit concerned about allowing any IP traffic with STATIC/STRICT_DNS. WDYT?

ramaraochavali:
I added more comments to other PR. Please see if that makes sense. I have put hold on both the PRs so that we can discuss.

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -714,6 +714,8 @@ func applyLoadBalancer(c *cluster.Cluster, lb *networking.LoadBalancerSettings,
 	case networking.LoadBalancerSettings_PASSTHROUGH:
 		c.LbPolicy = cluster.Cluster_CLUSTER_PROVIDED
 		c.ClusterDiscoveryType = &cluster.Cluster_Type{Type: cluster.Cluster_ORIGINAL_DST}
+		// Wipe out any LoadAssignment, if set. This can occur when we have a STATIC Service but PASSTHROUGH traffic policy
+		c.LoadAssignment = nil
```
hzxuzhonghu, line 718: This should be not right if  as analyzed in #39757

howardjohn:
> @howardjohn I thought about setting clusterLoadAssignments to nil while fixing it in other PR. But I think this seems a config issue - for service entries with Resolution as STATIC/DNS - some one is setting `PASSTHROUGH` Load balancer. By removing the cluster load assignments, we are allowing this cluster to route traffic to any IP that client chooses which I felt incorrect. So I went with ignoring DR and setting default LB model. If some one wants PASSTHROUGH why not set resolution `None`. This happened in this case probably because of wildcard host in DR - So I am bit concerned about allowing any IP traffic with STATIC/STRICT_DNS. WDYT?

STATIC is essentially what a standard ClusterIP service is. So by that logic we should not have PASSTHROUGH for these services either it would seem, which I think would negate all use cases?

I do agree it's a strange usage - I think the docs for PASSTHROUGH have quite a few caveats listed

ramaraochavali:
> STATIC is essentially what a standard ClusterIP service is. So by that logic we should not have PASSTHROUGH for these services either it would seem, which I think would negate all use cases?

The only difference I can think is ServiceEntry has "Resolution" type None which users can use. However for ClusterIP service there is no such option - unless we ask customers to create headless for usecases like pod->pod with subsets as I mentioned [here]([3 minutes ago](https://github.com/istio/istio/pull/39757#discussion_r912606845))

I do agree it is grey area. I think the reason I think we should not support for STATIC/DNS is right now the cluster is Nack - so no one would have used. But for K8s, people would have already used it as I mentioned in the comment. So I think we should continue to support it.

howardjohn:
> > STATIC is essentially what a standard ClusterIP service is. So by that logic we should not have PASSTHROUGH for these services either it would seem, which I think would negate all use cases?
> 
> The only difference I can think is ServiceEntry has "Resolution" type None which users can use. However for ClusterIP service there is no such option - unless we ask customers to create headless for usecases like pod->pod with subsets as I mentioned [here]([3 minutes ago](https://github.com/istio/istio/pull/39757#discussion_r912606845))
> 
> I do agree it is grey area. I think the reason I think we should not support for STATIC/DNS is right now the cluster is Nack - so no one would have used. But for K8s, people would have already used it as I mentioned in the comment. So I think we should continue to support it.

Headless vs ClusterIP is the same as NONE vs STATIC/DNS, IMO. And we do have conversion logic between those (used in MCP, for example). I am pretty against having different logic for the two cases

ramaraochavali:
I am fine with this fix. My only reservation is about user unknowingly allowing a PASSTHROUGH access for DNS/STATIC type of SE. But if you think that is user's problem - Let us move forward with this.

ramaraochavali (approved):

Issue #39472, opened 2022-06-15
Telemetry access logging: add cache


howardjohn:
In the old style access logs, we had caching of the generated config. In Telemetry, we lost this.

In real prod clusters, we saw 50% of Istiod CPU spent just on this part. It would be good to add a cache here.

I did some initial work here but got distracted. I might be able to complete but if anyone else wants to pick it up, even better.

zirain:
I will take a look this.

zirain:
> In real prod clusters, we saw 50% of Istiod CPU spent just on this part. It would be good to add a cache here.

can you share more details about this? what the cluster's size? how many workloads? how many telemetry resources in the clusters?

`buildAccessLogFromTelemetry` take 50% if istio beyond my unstanding.

zirain:
if you change provider from `envoy-json` to `envoy`, still have the same cpu cost?

zirain:
A long-term goal is [ecds: add support for access log extensions](https://github.com/envoyproxy/envoy/issues/12893)

howardjohn:
Cache done in https://github.com/istio/istio/pull/39751, followup in https://github.com/envoyproxy/envoy/issues/12893 for long term

PR #39751, opened 2022-07-02
add cache for accesslog

zirain:

with this PR, it will reduce times cpu cost in accessloging building, this will improve https://github.com/istio/istio/issues/39472

cc @howardjohn


pilot/pkg/bootstrap/server.go
```diff
@@ -924,6 +930,43 @@ func (s *Server) initRegistryEventHandlers() {
 	}
 }
 
+// telemetryHandler trigger to clean AccessLogging cache when telemetry logging configurations changed
+func (s *Server) telemetryHandler(prev config.Config, curr config.Config, event model.Event) {
+	if accessLoggingChanged(prev, curr, event) {
+		// built-in envoy provider related on MeshConfig, so we reused MeshConfigChanged callback
```
ramaraochavali, line 936: nit: built-in envoy provider related on MeshConfig -> built-in envoy provider relies on MeshConfig

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -138,18 +143,66 @@ type AccessLogBuilder struct {
...
+	val, ok := b.store.Get(key)
+	if !ok {
+		return nil, false
+	}
+
+	logs, ok := val.([]*accesslog.AccessLog)
+	return logs, ok
+}
+
+func (b *AccessLogBuilder) addAccessLog(key interface{}, accessLoggings []*accesslog.AccessLog) {
```
ramaraochavali, line 179: nit: addAccessLog -> cacheAccessLog may be better?

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -138,18 +143,66 @@ type AccessLogBuilder struct {
...
+	defer b.storeMutex.Unlock()
+
+	b.store.Add(key, accessLoggings)
+}
+
+func (b *AccessLogBuilder) purge() {
+	b.storeMutex.Lock()
+	defer b.storeMutex.Unlock()
+
+	b.store.Purge()
```
ramaraochavali, line 190: May be better to reinitialize the store - that is what we do in xds cache

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -237,12 +300,15 @@ func (b *AccessLogBuilder) setHTTPAccessLog(push *model.PushContext, proxy *mode
 		if mesh.EnableEnvoyAccessLogService {
 			connectionManager.AccessLog = append(connectionManager.AccessLog, b.httpGrpcAccessLog)
 		}
+
+		b.store.Add(k, connectionManager.AccessLog)
```
ramaraochavali, line 304: should we use the `addAccessLog` function instead of adding directly here so that we get locks?

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -237,12 +300,15 @@ func (b *AccessLogBuilder) setHTTPAccessLog(push *model.PushContext, proxy *mode
...
 		return
 	}
 
-	if al := buildAccessLogFromTelemetry(push, cfg, false); len(al) != 0 {
+	if al := buildAccessLogFromTelemetry(push, cfg, forListener); len(al) != 0 {
 		connectionManager.AccessLog = append(connectionManager.AccessLog, al...)
 	}
+	b.store.Add(k, connectionManager.AccessLog)
```
ramaraochavali, line 311: same here

pilot/pkg/networking/core/v1alpha3/accesslog.go
```diff
@@ -138,18 +143,66 @@ type AccessLogBuilder struct {
 	mutex                 sync.RWMutex
 	fileAccesslog         *accesslog.AccessLog
 	listenerFileAccessLog *accesslog.AccessLog
+
+	storeMutex sync.RWMutex
```
ramaraochavali, line 147: do we need separate mutex? Can we use the mutex that builder already has?

pilot/pkg/networking/core/v1alpha3/accesslog.go
zirain: yes, it's my mistake

pilot/pkg/networking/core/v1alpha3/accesslog.go
zirain: that will a little confuse with L165 `cachedAccessLog`

pilot/pkg/networking/core/v1alpha3/accesslog.go
zirain: done

pilot/pkg/networking/core/v1alpha3/accesslog.go
ramaraochavali: Ok. This is not clear where it is adding

pilot/pkg/bootstrap/server.go
```diff
@@ -923,6 +929,43 @@ func (s *Server) initRegistryEventHandlers() {
 	}
 }
 
+// telemetryHandler trigger to clean AccessLogging cache when telemetry logging configurations changed
+func (s *Server) telemetryHandler(prev config.Config, curr config.Config, event model.Event) {
```
howardjohn, line 933: PushContext already has a caching layer (or really  it *is* the caching layer). However, in it we are just caching the logging config, not the fully generated envoy config.

Can we just instead build the envoy config as part of push context? That would avoid a lot of this complexity

ramaraochavali: If we plan to build in push context -  we need both the actual config + generated config. I guess there may be cases where we need raw config also (in debug?) ?

zirain: @howardjohn hope this's what's you wanted, PTAL.

zirain:
kindly ping @howardjohn

pilot/pkg/model/push_context.go
```diff
@@ -45,6 +47,9 @@ import (
 	"istio.io/pkg/monitoring"
 )
 
+// TODO: make this configurable
+const cacheSize = 200
```
howardjohn, line 51: I don't think we need a full blown LRU cache within PushContext. PushContext _is_ already a cache. The problem is we are storing an intermediate state (LoggingConfig), and then for every listener (often times called 1000s of times), we convert LoggingConfig -> Envoy config.

I think simply storing the computed config in LoggingConfig would be sufficient optimization

howardjohn: Or something like `computedMetricsFilters` would provide cross-Proxy caching, although that may not really be super important. The above change would already give us orders of magnitude improvement (O(listeners))

pilot/pkg/model/push_context.go
```diff
@@ -221,6 +226,8 @@ type PushContext struct {
 	// Telemetry stores the existing Telemetry resources for the cluster.
 	Telemetry *Telemetries `json:"-"`
 
+	AccessLogCache *AccessLogCache
```
howardjohn, line 229: Whatever we do end up doing should be within Telemetries I think

pilot/pkg/model/push_context.go
howardjohn: Storing in LoggingConfig would require the caller to only fetch a single LoggingConfig though, access log is currently not structured like that. So maybe not the right path.

I would say to just follow the same pattern as `computedMetricsFilters`. Its pretty similar to what you have here except using a simple map vs LRU, and storing in Telemetries

zirain: > Just 1 single telemetry. 5000 proxies. Access logs config are EXTREMELY expensive to generate since its so verbose and so many fields (https://github.com/envoyproxy/envoy/issues/21718)

In this scenario, I believe most of costs are in `setTCPAccessLog\setHTTPAccessLog`(I do not have a real pprof)
If we just cache LoggingConfig, may not help.

> Its pretty similar to what you have here except using a simple map vs LRU, and storing in Telemetries
> On a fairly small cluster I end up with 400 access log configs. The LDS is 700kb. Disabling access logs drops it down to 200kb.

There's a huge memory cost if we use a simple map? 500kb * mum of proxy?

howardjohn: We already do it for http and tcp filters for stats. Almost all proxies have the same config so it will end up being pretty small is my guess?

howardjohn: if it is an issue we can cap the map size, but lru is never needed IMO since the scope is to a single PushContext which is already short lived

pilot/pkg/util/protoconv/protoconv.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"google.golang.org/protobuf/encoding/prototext"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
+
+	"istio.io/pkg/log"
+)
+
+// MessageToAnyWithError converts from proto message to proto Any
+func MessageToAnyWithError(msg proto.Message) (*anypb.Any, error) {
```
zirain, line 28: copy from pkg/netwroking to avoid import cycle

zirain:
```console
BenchmarkInitPushContext/telemetry-api
BenchmarkInitPushContext/telemetry-api-8    	    1346	    910086 ns/op	  152867 B/op	    4627 allocs/op
BenchmarkInitPushContext/telemetry-api-8    	    1357	    883796 ns/op	  152837 B/op	    4627 allocs/op
BenchmarkInitPushContext/telemetry-api-8    	    1393	    882667 ns/op	  153738 B/op	    4627 allocs/op
BenchmarkInitPushContext/telemetry-api-8    	    1345	    874319 ns/op	  152867 B/op	    4627 allocs/op
BenchmarkInitPushContext/telemetry-api-8    	    1306	    902060 ns/op	  152983 B/op	    4627 allocs/op
```

pilot/pkg/model/push_context.go
zirain: updated the implements.

pilot/pkg/util/protoconv/protoconv.go
GregHanson: this function and MessageToAny are in `pilot/pkg/networking` and `pilot/pkg/networking/util`.  Is there a reason these can't all use the same function?

GregHanson: can also be addressed in a followup PR since the refactor is outside the scope of this PR

pilot/pkg/model/telemetry.go
```diff
@@ -100,6 +109,7 @@ func getTelemetries(env *Environment) (*Telemetries, error) {
 		RootNamespace:          env.Mesh().GetRootNamespace(),
 		meshConfig:             env.Mesh(),
 		computedMetricsFilters: map[metricsKey]interface{}{},
+		computedLoggingConfig:  map[loggingKey]*LoggingConfig{},
```
howardjohn, line 112: nit:

```suggestion
		computedLoggingConfig:  map[loggingKey]LoggingConfig{},
```

the LoggingConfig type is small, making it a pointer adds some complexity (can the map hold nil, can we mutate it, etc)

howardjohn (approved):

howardjohn:
> ```
> BenchmarkListenerGeneration/telemetry-api-8                     	     379	   3080852 ns/op	        54.67 kb/msg	         6.000 resources/msg	 1304774 B/op	   19522 allocs/op
> BenchmarkListenerGeneration/telemetry-api-8                     	     376	   3107888 ns/op	        54.67 kb/msg	         6.000 resources/msg	 1304771 B/op	   19521 allocs/op
> BenchmarkListenerGeneration/telemetry-api-8                     	     379	   3108754 ns/op	        54.67 kb/msg	         6.000 resources/msg	 1304754 B/op	   19521 allocs/op
> BenchmarkListenerGeneration/telemetry-api-8                     	     378	   3124344 ns/op	        54.67 kb/msg	         6.000 resources/msg	 1304753 B/op	   19521 allocs/op
> BenchmarkListenerGeneration/telemetry-api-8                     	     385	   3089206 ns/op	        54.67 kb/msg	         6.000 resources/msg	 1304761 B/op	   19521 allocs/op
> ```
> 
> prev: [#39899 (comment)](https://github.com/istio/istio/pull/39899#issuecomment-1182806278)

btw, it is useful to use `benchstat` to compare them. Looks great though

pilot/pkg/util/protoconv/protoconv.go
zirain: I will refactor all of those to same place in a separate PR.

pilot/pkg/model/telemetry.go
zirain: fixed, can we remove the `DNM`?cc @howardjohn

PR #39757, opened 2022-07-02
minor refactor in cluster builder

ramaraochavali:
Minor refactor and additional tests
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -712,8 +712,14 @@ func applyLoadBalancer(c *cluster.Cluster, lb *networking.LoadBalancerSettings,
 	case networking.LoadBalancerSettings_ROUND_ROBIN:
 		applyRoundRobinLoadBalancer(c, lb)
 	case networking.LoadBalancerSettings_PASSTHROUGH:
-		c.LbPolicy = cluster.Cluster_CLUSTER_PROVIDED
-		c.ClusterDiscoveryType = &cluster.Cluster_Type{Type: cluster.Cluster_ORIGINAL_DST}
+		if c.GetType() != cluster.Cluster_ORIGINAL_DST {
```
ramaraochavali, line 715: This is the actual fix

pilot/pkg/networking/core/v1alpha3/cluster_builder.go
```diff
@@ -368,6 +368,13 @@ func (cb *ClusterBuilder) buildDefaultCluster(name string, discoveryType cluster
 		}
 	}
 
+	if discoveryType == cluster.Cluster_ORIGINAL_DST {
```
ramaraochavali, line 371: Moved here so that this is set for both inbound and outbound

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -505,15 +506,14 @@ func convertResolution(proxyType model.NodeType, service *model.Service) cluster
 	case model.DNSRoundRobinLB:
 		return cluster.Cluster_LOGICAL_DNS
 	case model.Passthrough:
+		if service.Attributes.ServiceRegistry == provider.Kubernetes && features.EnableEDSForHeadless {
+			return cluster.Cluster_EDS
+		}
 		// Gateways cannot use passthrough clusters. So fallback to EDS
-		if proxyType == model.SidecarProxy {
```
hzxuzhonghu, line 509: why changing this semantic?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -712,8 +713,24 @@ func applyLoadBalancer(c *cluster.Cluster, lb *networking.LoadBalancerSettings,
 	case networking.LoadBalancerSettings_ROUND_ROBIN:
 		applyRoundRobinLoadBalancer(c, lb)
 	case networking.LoadBalancerSettings_PASSTHROUGH:
-		c.LbPolicy = cluster.Cluster_CLUSTER_PROVIDED
-		c.ClusterDiscoveryType = &cluster.Cluster_Type{Type: cluster.Cluster_ORIGINAL_DST}
+		// This can happen two cases.
+		// 1. A ServiceEntry is created with Resolution as DNS/STATIC and the DestinationRule
+		//    is configured with PASSTHROUGH load balancer.
+		// 2. A regular Kubernetes Service (of type EDS), configures DestinationRule with PASSTHROUGH load balancer.
```
hzxuzhonghu, line 719: Not sure i fully understand, then a regular k8s service plus a pass through DR setting will be no redirect at all? That seems a misconfig too.

pilot/pkg/networking/core/v1alpha3/cluster.go
ramaraochavali: It is for easy readability. Personal preference but I can revert

pilot/pkg/networking/core/v1alpha3/cluster.go
ramaraochavali: As I mentioned there are use cases for regular K8S service to configure PASSTHROUGH if you want direct pod -> pod connectivity.  There are people that use this (we are one among them :-)) https://discuss.istio.io/t/503-between-pod-to-pod-communication-1-5-1/6121/15 is another example

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: Nope, you changed, not refactor

ramaraochavali: Can you check now - basically want to return early if it is gateway

hzxuzhonghu: right now

pilot/pkg/networking/core/v1alpha3/cluster.go
hzxuzhonghu: pod to pod communicate in k8s usually make use of headless service

hzxuzhonghu: FYI https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields

ramaraochavali: I know. This is to avoid that headless service and as I said people already use it because PASSTHROUGH is supported in DR.

hzxuzhonghu: I donot think so, this is influcing normal k8s service. for k8s service, the cluster type is original_dstm why not check it here?

ramaraochavali: This is currently supported for regular k8s service. As I said people use it already. We can not ask, them to move all to headless now - if people have both use cases, they have to create two services one with headless and and one with cluster ip. This is a breaking change. This change is to support that case.

For service entries, it is not supported today also - as cluster will be NACKed.

hzxuzhonghu: Can you tell how people connect to one specific pod?

ramaraochavali: They get IP of pod by some means (it is application dependent) and they will call using IP:PORT. See the second appraoch documented here https://discuss.istio.io/t/503-between-pod-to-pod-communication-1-5-1/6121/15 - create a subset and with PASSTHROUGH and call with ip:port

hzxuzhonghu: Confused if directly access by ip, how does it select a cluster?

ramaraochavali:
@hzxuzhonghu PTAL

hzxuzhonghu (approved):
lgtm for simple refact

PR #39761, opened 2022-07-04
Fix min k8s version supported

hanxiaop:
Based on https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases.
Officially supported k8s version from 1.21-1.24

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/istioctl/pkg/install/k8sversion/version.go
+++ b/istioctl/pkg/install/k8sversion/version.go
@@ -28,7 +28,7 @@ import (
 const (
 	// MinK8SVersion is the minimum k8s version required to run this version of Istio
 	// https://istio.io/docs/setup/platform-setup/
-	MinK8SVersion               = 19
+	MinK8SVersion               = 21
 	UnSupportedK8SVersionLogMsg = "\nThe Kubernetes version %s is not supported by Istio %s. The minimum supported Kubernetes version is 1.%d.\n" +
 		"Proceeding with the installation, but you might experience problems. " +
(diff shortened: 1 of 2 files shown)


istioctl/pkg/install/k8sversion/version.go
```diff
@@ -28,7 +28,7 @@ import (
 const (
 	// MinK8SVersion is the minimum k8s version required to run this version of Istio
 	// https://istio.io/docs/setup/platform-setup/
-	MinK8SVersion               = 19
+	MinK8SVersion               = 21
```
hzxuzhonghu, line 31: We donot need to upgrade this unless it does not work in 1.19 or 1.20 anymore

hanxiaop: @hzxuzhonghu Shouldn't this be the value matches https://preliminary.istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases, and https://github.com/istio/istio.io/blob/4827604b9c90e272d0d25bb173f79e19b2b57b3d/data/args.yml#L31-L32? I've checked the old PRs, and the min k8s version was always updated along with the Istio version.

hzxuzhonghu: This is not to endusers, so i think not a must,  it is ok to be a superset of our announced supported versions.

PR #39763, opened 2022-07-04
Reduce []ConfigKey slice allocate numbers during xds caching 

hzxuzhonghu:

This is making use of sync.Pool to relieve pressure on the garbage collector.

```
BenchmarkDependentConfigs
BenchmarkDependentConfigs/rds_cache_using_mem_pool
BenchmarkDependentConfigs/rds_cache_using_mem_pool-8         	  390069	      2903 ns/op	      96 B/op	       3 allocs/op
BenchmarkDependentConfigs/rds_cache_not_use_mem_pool
BenchmarkDependentConfigs/rds_cache_not_use_mem_pool-8       	  125289	      9862 ns/op	    8288 B/op	       4 allocs/op
```

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/xds_cache.go
+++ b/pilot/pkg/model/xds_cache.go
@@ -173,10 +173,6 @@ type LruCache struct {
 
 	// dependantsPool simply stores dependentConfigs of XdsCacheEntry to avoid additional memory allocations
 	// by caching allocated but unused items for later reuse, relieving pressure on the garbage collector.
-	//
-	// Usage:
-	//  memoryAllocator := runtime.AllocatorPool.Get().(*runtime.Allocator)
-	//  defer runtime.AllocatorPool.Put(memoryAllocator)
 	dependantsPool *sync.Pool
 }
 
@@ -226,7 +222,7 @@ func (l *LruCache) handleEvicted(stopCh <-chan struct{}) {
 		case <-l.evictCh:
 			l.clearEvicted()
 		case <-stopCh:
-			log.Infof("LruCache has been stopped")
+			log.Debug("LruCache has been stopped")
 			return
 		}
 	}
--- a/pilot/pkg/model/xds_cache.go
+++ b/pilot/pkg/model/xds_cache.go
@@ -92,7 +92,7 @@ type XdsCacheEntry interface {
 	DependentTypes() []kind.Kind
 	// DependentConfigs is config items that this cache key is dependent on.
 	// Whenever these configs change, we should invalidate this cache entry.
-	DependentConfigs() []ConfigKey
+	DependentConfigs(allocator *Allocator) []ConfigKey
 	// Cacheable indicates whether this entry is valid for cache. For example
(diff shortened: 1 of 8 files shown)


hzxuzhonghu:
Replace #39731, which is a mix

pilot/pkg/model/xds_cache_allocator.go
```diff
@@ -0,0 +1,40 @@
...
+
+package model
+
+// Note: referenced kubernetes optimization: staging/src/k8s.io/apimachinery/pkg/runtime/allocator.go
+
+// Allocator knows how to allocate memory
+// It exists to make the cost of xds cache Add cheaper.
+// In some cases, it allows for allocating memory only once and then reusing it.
+// This approach puts less load on GC and leads to less fragmented memory in general.
+type Allocator struct {
```
stevenctl, line 23: super nit: ConfigKeyAllocator.

pilot/pkg/model/xds_cache_allocator.go
```diff
@@ -0,0 +1,40 @@
...
+	configs []ConfigKey
+}
+
+// Allocate reserves memory for n configs only if the underlying array doesn't have enough capacity
+// otherwise it returns previously allocated block of memory.
+//
+// Note that the returned array is not zeroed, it is the caller's
+// responsibility to clean the memory if needed.
+func (a *Allocator) Allocate(n int) []ConfigKey {
```
stevenctl, line 32: We could do the nil check here since its a pointer method.. seems weird having to copy around the logic for utilizing the allocator.

pilot/pkg/model/xds_cache.go
```diff
@@ -92,7 +92,7 @@ type XdsCacheEntry interface {
 	DependentTypes() []kind.Kind
 	// DependentConfigs is config items that this cache key is dependent on.
 	// Whenever these configs change, we should invalidate this cache entry.
-	DependentConfigs() []ConfigKey
+	DependentConfigs(allocator *Allocator) []ConfigKey
```
stevenctl, line 95: on second thought.. could we just pass the `preAllocKeys []ConfigKey` itself here; the implementations just `append` so nil slice is okay. I think this simplifies the usage a lot.

stevenctl:
Nice improvement although I feel the usage could be simplified

hzxuzhonghu:
@stevenctl As we are storing the dependentConfigs now, this is not that meaningful.  I am thinking maybe this can be used for xds resource marshaling. Currently, the marshaling is very mem costing and cause great pressure to Garbage collector

howardjohn:
over a year ago I tried using sync pool for XDS marshal and didn't see much
improvement. However, I suspect I did it wrong or measure wrong, since it
should have at least some impact...

Issue #39586, opened 2022-06-23
TestTunnelingOutboundTraffic test flake


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/integ-pilot-multicluster_istio_postsubmit/1540064995352514560

howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/integ-pilot_istio_postsubmit/1540104022755643392 another example

jewertow:
I took a look at config dumps and everything seems to be configured properly. At the moment I have no idea what's wrong, because I never observed similar problem running those tests locally. So first I will configure access logging in the external-forward-proxy to be able to understand what happens in proxy listeners when tests fail.

howardjohn:
Seems to be stabilized

Issue #38982, opened 2022-05-17
IPTables Redirect rule causes tcp connection timeouts with multiple ports


klarose:
### Symptoms

A virtual service for tcp traffic occasionally fails to connect to a kubernetes pod with multiple exposed ports. Occasionally tcp connections from our istio ingress to a pod with an istio sidecar fail with a UF, URX flag. 
Adding retries to our affected tcp proxies removed the connectivity problem from the perspective of the client. However, establishing the connection still took > 1.5s because the ingress would need to wait for the connection timeout to detect the failure, which isn't an acceptable perf hit when it is otherwise avoidable.

Digging in, we found that the failures were caused by the pod dropping the SYN packet and replying to a *different* connection with an [rfc5961](https://datatracker.ietf.org/doc/html/rfc5961) Challenge Ack. In particular, we saw that the istio ingress would send a SYN to port 5002. The destination
pod did not respond to that SYN. Instead, it would send a Challenge ACK to the same ip and port, but from port 5000. After some thought, it became apparent that the iptables redirect rules mapping all connections to the same destination port was
leading to collisions in the pod's connection table, and thus timeouts on the colliding connections.

We could see TcpExtTCPChallengeACK in the linux nstat tool incrementing when this occurs, which aligns with our theory.

### Setup
- One service with port 5000 and port 5002. 
- Two virtual services: One forwarding http from host X to port 5000. The other sni forwarding from host Y to port 5002.

### Analysis

The istio init container configures some iptables rules which ensure that the istio sidecar receives all inbound traffic. It does this by redirecting all connections to a specific destination port (say 15006). You can see this by viewing
the configured listeners:

```
0.0.0.0       15006 Trans: raw_buffer; Addr: *:5002                                                                 Cluster: inbound|5002||
0.0.0.0       15006 Trans: tls; Addr: *:5002                                                                        Cluster: inbound|5002||
0.0.0.0       15006 Trans: tls; App: istio,istio-peer-exchange,istio-http/1.0,istio-http/1.1,istio-h2; Addr: *:5002 Cluster: inbound|5002||
0.0.0.0       15006 Trans: raw_buffer; Addr: *:5000                                                                 Cluster: inbound|5000||
0.0.0.0       15006 Trans: tls; App: istio,istio-peer-exchange,istio-http/1.0,istio-http/1.1,istio-h2; Addr: *:5000 Cluster: inbound|5000||
```

[...]

PR #39774, opened 2022-07-04
Enable TestTunnelingOutboundTraffic and configure access log in forward proxy

jewertow:
Signed-off-by: Jacek Ewertowski <jewertow@redhat.com>

Related issue: #39586

I investigated config dumps from few failed tests and I can't find any misconfiguration. I didn't encounter similar problem on my computer - the tests always pass. So the first step to understand the problem better is adding access log to the forward proxy. I would like to know what status code and flag are returned.
After merging this PR I will look for failing post-submit tests and will try to solve to cause of the problem (I have some ideas, but I am not sure at the moment) or disable test again.

### DIFF
--- a/pkg/test/echo/server/endpoint/http.go
+++ b/pkg/test/echo/server/endpoint/http.go
@@ -211,7 +211,11 @@ type codeAndSlices struct {
 
 func (h *httpHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	id := uuid.New()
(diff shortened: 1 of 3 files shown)


jewertow:
JFYI: I found the real cause of fails on multi-cluster setup. I will push fixes soon.

howardjohn:
https://github.com/istio/istio/issues/38982 is the root cause of many issues we see in other tests. But it seems odd to trigger here since this is mostly testing gateways.

I was waiting to approve since you said you had found a fix. Do we need to merge to get more diagnostics?

howardjohn (approved):

jewertow:
TestReachability still fails

Issue #29264, opened 2020-11-30
Support direct response in VirtualService


illidan80:
For example, I would like to have something like this:

```
apiVersion: networking.istio.io/v1alpha3 
kind: VirtualService 
metadata:
  name: myservice-vs 
  namespace: mynamespace
spec:
  hosts:
  - "my.host"
  gateways:
  - mygateway
  http:
  - match:
    - uri:
        prefix: /mypath
    route:
     - response:
         code: 200
         body: "this is automatic response"
```

hzxuzhonghu:
Looks like a necessary feature

esnible:
Duplicate of auto-closed https://github.com/istio/istio/issues/12096 and discussed at https://discuss.istio.io/t/support-for-envoys-direct-response/2349

I believe this could be done with an EnvoyFilter today https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/direct_response/v2/config.proto

p53:
this would be useful, can be done several ways but with workarounds would consume more resources than optimal

Zeka13:
I am still confused how I can implement this with istio (I am migrating from nginx ingress controller)

esnible:
You can do this with WebAssembly code and `sendLocalResponse()`.  I have tried it and it works.

A better way is to use Envoy's [DirectResponseAction](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-directresponseaction).  I believe Istio can provide this configuration to Envoy using an EnvoyFilter [RouteMatch.Action](https://istio.io/latest/docs/reference/config/networking/envoy-filter/#EnvoyFilter-RouteConfigurationMatch-RouteMatch-Action).  I have not tried it myself nor was I able to locate an example to point you at.

Zeka13:
@esnible Hello! Actually I am also didn't find any proper example and noone answered me back there https://discuss.istio.io/t/please-assist-me-with-using-envoyfilter-routeconfigurationmatch-action-direct-response/10388

This is really not easy to undestand this EnvoyFilter entity without simple examples :(

esnible:
I have tagged `extensions and telemetry`, you really have what the Istio developers would call an "extension" use case.

I will ask if anyone has examples as the next Telemetry and Extensions meeting.

Zeka13:
@esnible thanks a lot

esnible:
[...]

PR #39776, opened 2022-07-04
Virtual service direct response

sschepens:
Implement Virtual Service Direct Responses.

~~Depends on: https://github.com/istio/api/pull/2407~~

Fixes: https://github.com/istio/istio/issues/29264


sschepens:
@howardjohn can you take a look at this and https://github.com/istio/api/pull/2407 please?

pkg/config/validation/virtualservice.go
```diff
@@ -229,8 +230,28 @@ func validateHTTPRouteConflict(http *networking.HTTPRoute, routeType HTTPRouteTy
...
+		if http.Rewrite != nil {
+			errs = appendErrors(errs, errors.New("HTTP route rule cannot contain both rewrite and direct_response"))
+		}
+
+		if http.Redirect != nil {
+			errs = appendErrors(errs, errors.New("HTTP route rule cannot contain both redirect and direct_response"))
+		}
 	} else if len(http.Route) == 0 {
-		errs = appendErrors(errs, errors.New("HTTP route or redirect is required"))
+		errs = appendErrors(errs, errors.New("HTTP route, redirect or direct_response is required"))
```
hzxuzhonghu, line 254: The above need a refactor maybe later.

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -636,6 +638,33 @@ func applyRedirect(out *route.Route, redirect *networking.HTTPRedirect, port int
...
+func applyDirectResponse(out *route.Route, directResponse *networking.HTTPDirectResponse, port int) {
+	action := &route.Route_DirectResponse{
+		DirectResponse: &route.DirectResponseAction{
+			Status: directResponse.Status,
+		},
+	}
+
+	if directResponse.Body != nil {
+		switch op := directResponse.Body.Specifier.(type) {
+		case *networking.HTTPBody_String_:
```
hzxuzhonghu, line 650: can we change this `HTTPBody_String_`, it does not align with our naming convention

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -636,6 +638,33 @@ func applyRedirect(out *route.Route, redirect *networking.HTTPRedirect, port int
...
+			action.DirectResponse.Body = &core.DataSource{
+				Specifier: &core.DataSource_InlineString{
+					InlineString: op.String_,
```
hzxuzhonghu, line 653: same for `op.String_`

pilot/pkg/networking/core/v1alpha3/route/route.go
sschepens: I'm not sure what you mean, this is a protobuf generated struct, this comes from https://github.com/istio/api/blob/master/networking/v1alpha3/virtual_service.proto#L1615

I guess since the field is named `string` protobuf appends a `_` at the end which does not happen with `bytes` field

pilot/pkg/networking/core/v1alpha3/route/route.go
sschepens: same as above.

ericvn:
The release notes test fails since there isn't a release note, but the one in https://github.com/istio/api/pull/2407 should cover this.

pkg/config/validation/validation.go
```diff
@@ -3024,6 +3024,14 @@ func validateHTTPRedirect(redirect *networking.HTTPRedirect) error {
 	return nil
 }
 
+func validateHTTPDirectResponse(directResponse *networking.HTTPDirectResponse) error {
+	if directResponse == nil {
```
howardjohn, line 3028: can we add validation for the size? 

I would give a warning if its >100kb with something like "large responses may impact control plane performance" and then hard failure >1mb

sschepens: done!

howardjohn (approved):

sschepens:
hmm it's weird that unit tests are failing since i haven't really touched those, gonna check.

rverma-nsl:
@sschepens In which release this feature would be available?

sschepens:
I'm guessing 1.15 since it hasn't been backported

GregHanson:
@rverma-nsl new features are not usually backported - they will appear in the next 1.X release

Issue #38077, opened 2022-03-22
Istio CNI Install not using current ServiceAccount tokens


tommyp1ckles:
### Bug Description

On k8s clusters running 1.21+ mounted service account tokens can be created with a expiry. I've started noticing nodes getting into states with errors like `failed to set up sandbox container ... network: Unauthorized`.

This appears to be due to expired or invalid tokens being used in `/etc/cni/net.d/ZZZ-istio-cni-kubeconfig`. Initially we thought this was an aws-vpc-cni issue since restarting that pod would fix the issue on that node.

However this appears to only fix the problem since this triggers a filesystem notify event here: https://github.com/istio/istio/blob/1.11.4/cni/pkg/install/install.go#L78 which causes a full configuration reload, including the kubeconfig file.

Should this loop also not be watching for updates to the mounted service account token as well?

### Version

```prose
Istio Version: 1.11.1
Kubernetes Version: v1.21.5-eks-bc4871b
```

(edit: that was k8s 1.21, not 1.12).

jbilliau-rcd:
Im having this exact same issue on multiple clusters, probably a dozen times over the past 3 months. Running Istio 1.11.4 and EKS 1.21. Thought it was aws-node related, but there are no errors or anything, and then seeing that issue now makes me think it's Istio related?

jbilliau-rcd:
It looks like istio-cni needs to be updated to support `BoundServiceAccountTokens`. See email we just got from AWS:

> Kubernetes version 1.21 graduated BoundServiceAccountTokenVolume feature [1] to beta and enabled it by default. This feature improves security of service account tokens by requiring a one hour expiry time, over the previous default of no expiration. This means that applications that do not refetch service account tokens periodically will receive an HTTP 401 unauthorized error response on requests to Kubernetes API server with expired tokens. You can learn more about the BoundServiceAccountToken feature in EKS Kubernetes 1.21 release notes [2].
> 
> To enable a smooth migration of applications to the newer time-bound service account tokens, EKS v1.21+ extends the lifetime of service account tokens to 90 days. Applications on EKS v1.21+ clusters that make API server requests with tokens that are older than 90 days will receive an HTTP 401 unauthorized  error response.
> 
> How can you resolve the issue?
> 
> To make the transition to time bound service account tokens easier, Kubernetes has updated the below official versions of client SDKs to automatically refetch tokens before the one hour expiration:
> 
> * Go v0.15.7 and later
> * Python v12.0.0 and later
[...]

PR #39801, opened 2022-07-06
Fix bug(#38077) that causes a stale SA token is being used in istio-cni

srggavrilov:
The BoundServiceAccountTokenVolume Kubernetes feature (graduated to stable in 1.22) improves security of service account tokens by requiring a one hour expiry time, over the previous default of no expiration. This means that install-cni should refetch service account tokens periodically.

This PR integrates periodic SA token checks into already existing sleepCheckInstall function and compatible with legacy token secret mounts.


cni/pkg/install/install.go
```diff
@@ -273,3 +278,19 @@ func checkInstall(cfg *config.InstallConfig, cniConfigFilepath string) error {
 	}
 	return nil
 }
+
+// watchSAToken periodically reads SA token file and compares its content with the token stored in the Installer.
+// Sends true into fileModified in case of mismatch.
+// Allows to detect changes in the Bound Service Account Token Volume.
+func (in *Installer) watchSAToken(fileModified chan bool, errChan chan error) {
```
howardjohn, line 285: Seems like this goroutine leaks and is never cleaned up?

GregHanson (approved):
Release note looks good

lei-tang:
The following code in https://github.com/istio/istio/pull/39801 seems to be rather tricky. If the ctx.Done() is not triggered, the goroutine created by each call of watchSAToken() will never return and cause memory leak.

func (in *Installer) watchSAToken(ctx context.Context, fileModified chan bool, errChan chan error) {
        curToken := in.saToken
        go func() {
                for {
                        select {
                        case <-ctx.Done():
                                return
                        default:
                                token, err := readServiceAccountToken(in.saTokenFilepath)
                                if err != nil {
                                        errChan <- err
                                }
                                if curToken != token {
                                        fileModified <- true
                                }
                                time.Sleep(1 * time.Minute)
                        }
                }
        }()
}

PR #39829, opened 2022-07-07
[wip] dont apply client TLS filtering on gateways

stevenctl:


pilot/pkg/xds/endpoint_builder.go
```diff
@@ -384,9 +386,12 @@ type mtlsChecker struct {
 	subsetPolicyMode map[string]*networkingapi.ClientTLSSettings_TLSmode
 	// the tlsMode of the root traffic policy if it's set
 	rootPolicyMode *networkingapi.ClientTLSSettings_TLSmode
+
+	// Indicates the cluster we're checking settings for doesn't care about mTLS settings on the client (DestinationRule).
+	passthroughMode bool
```
hzxuzhonghu, line 391: This is confusing, i think you mean for east west gateway, the tls mode of from DR is ignorred

stevenctl: `cluster` as in xDS cluster. The passthrough clusters don't actually apply client TLS settings since they don't terminate/originate any TLS.

stevenctl:
So I changed this a bit to allow the "workaround destinationrule" to work, even though it shouldn't be needed. 
Without ClientTLSSettings forcing ISTIO_MUTUAL tls mode, we allow either PeerAuthentication or the `tlsMode` label to cause the endpoint filtering. 

In the egress case, it's the tlsMode causing the issue. We only have that label on sidecar pods. We could either:
* Add the `security.istio.io/tlsMode: istio` label to the egress (and probably other gateways?)
* Inside of pilot code (probably [here](https://github.com/istio/istio/blob/dd8baaa8d2db49360eb2d2ef41f6b76bb36a37b1/pilot/pkg/model/service.go#L796)) implicitly make the label `istio: egressgateway` (and probably others) ~= tlsMode: istio.

manifests/charts/gateways/istio-egress/values.yaml
```diff
@@ -16,6 +16,7 @@ gateways:
     labels:
       app: istio-egressgateway
       istio: egressgateway
+      security.istio.io/tlsMode: istio
```
howardjohn, line 19: This doesn't seem safe? Won't we accidentally start doing autoMTLS to the gateway?

pilot/pkg/model/service.go
```diff
@@ -831,6 +832,12 @@ func (ep *IstioEndpoint) DeepCopy() *IstioEndpoint {
 	return copyInternal(ep).(*IstioEndpoint)
 }
 
+var gatewayNames = sets.New("ingressgateway", "egressgateway")
```
stevenctl, line 835: Do we have these as constants anywhere? Is this even truly a "standard"?

hzxuzhonghu: Seems not a standard naming

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -419,16 +425,21 @@ func (c *mtlsChecker) computeForEndpoint(ep *model.IstioEndpoint) {
 	if drMode := c.mtlsModeForDestinationRule(ep); drMode != nil {
 		switch *drMode {
 		case networkingapi.ClientTLSSettings_DISABLE:
-			c.mtlsDisabledHosts[lbEpKey(ep.EnvoyEndpoint)] = struct{}{}
-			return
+			if !c.sniDNAT {
```
hzxuzhonghu, line 428: I am confused here, shouldn't check sniDNAT first and set it to c.mtlsDisabledHosts[lbEpKey(ep.EnvoyEndpoint)] in any case?

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -419,16 +425,21 @@ func (c *mtlsChecker) computeForEndpoint(ep *model.IstioEndpoint) {
...
+			// This setting won't actually apply on the SNI-DNAT clusters, but we still allow the setting as
+			// an escape hatch to avoid the filtering here.
 			return
 		}
 	}
 
-	// if endpoint has no sidecar or explicitly tls disabled by "security.istio.io/tlsMode" label.
-	if ep.TLSMode != model.IstioMutualTLSModeLabel {
+	// the endpoint must either be a part of istio or have the securty.istio.io/tlsMode: istio label
+	if !ep.IsIstioGateway() && ep.TLSMode != model.IstioMutualTLSModeLabel {
```
hzxuzhonghu, line 442: Why checking IsGateway here? IIRC， tls mode is always set on gateway

stevenctl: I don't think that's true from my testing. I don' think checking "is gateway" is a good idea, more of an experiment so far.

stevenctl:
Included #40243 + the "fix" for eastwest gateway being considered TLS-enabled and things look stable.

Issue #39825, opened 2022-07-07
Unready Kubernetes pods reachable via loadbalancer on Istio v1.14.1


schmigware:
### Bug Description

For a simple configuration:

- Some pod with configured readiness probe
- Aforementioned pod is load-balanced by a ClusterIP service
- Istio installed

I expect that when a load-balanced pod is not ready, it should not be reachable via the load-balancing service.

With Istio 1.14.1, if I configure:
- Some pod(s) with ready status FALSE
- ClusterIP service (load-balancer)
-  A pair of pods from which to issue curl requests, one with an envoy sidecar and one without

Then:

- The pod with an envoy sidecar can successfully e.g. curl an unready endpoint via the load balancer. 
- The pod without any envoy sidecar cannot.

For earlier Istio versions - specifically 1.13.2 for instance - I observe:
- The pod with an envoy sidecar cannot successfully e.g. curl an endpoint, returning status **no healthy upstream**
- The pod without any envoy sidecar cannot reach any loadbalanced pod.

The behaviour exhibited by version 1.13.2 is the behaviour I expect. 

Please clarify, many thanks.
 

### Version

```prose
client version: 1.14.1
control plane version: 1.14.1
data plane version: 1.13.2 (2 proxies), 1.14.1 (2 proxies)

Client Version: v1.24.2
Kustomize Version: v4.5.4
Server Version: v1.23.8+k3s1

v3.9.0+g7ceeda6
```

zirain:
may related to https://github.com/istio/istio/pull/37122, cc @ramaraochavali

ramaraochavali:
Can you please share /config_dump?include_eds output on the proxy and share the service name?

schmigware:
[config_dump.log](https://github.com/istio/istio/files/9063180/config_dump.log)

ramaraochavali:
It shows both the endpoints as `UNHEALTHY` - which is correct. What is the curl output you are getting?

```
  "endpoints": [
       {
        "locality": {},
        "lb_endpoints": [
         {
          "endpoint": {
           "address": {
            "socket_address": {
             "address": "10.42.0.250",
             "port_value": 8080
            }
           },
           "health_check_config": {}
          },
          "health_status": "UNHEALTHY",
          "metadata": {
           "filter_metadata": {
            "istio": {
             "workload": "aggregator;dcm;aggregator;latest;Kubernetes"
            },
            "envoy.transport_socket_match": {
             "tlsMode": "istio"
            }
           }
[...]

PR #39834, opened 2022-07-08
disable sending unhealthy endpoints by default

ramaraochavali:
Fixes https://github.com/istio/istio/issues/39825

- [ ] Docs
- [ ] Installation
- [ X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -108,7 +108,7 @@ var (
 		"PILOT_SEND_UNHEALTHY_ENDPOINTS",
 		false,
 		"If enabled, Pilot will include unhealthy endpoints in EDS pushes and even if they are sent Envoy does not use them for load balancing."+
-			"  To avoid, sending traffic to non ready endpoints, enabling this flag, disables pani threshold in Envoy i.e. Envoy does not load balance requests"+
+			"  To avoid, sending traffic to non ready endpoints, enabling this flag, disables panic threshold in Envoy i.e. Envoy does not load balance requests"+
 			" to unhealthy/non-ready hosts even if the number of healthy hosts come below minimum health percentage(panic threshold).",
 	).Get()
 
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -109,7 +109,7 @@ var (
 		false,
 		"If enabled, Pilot will include unhealthy endpoints in EDS pushes and even if they are sent Envoy does not use them for load balancing."+
 			"  To avoid, sending traffic to non ready endpoints, enabling this flag, disables panic threshold in Envoy i.e. Envoy does not load balance requests"+
-			" to unhealthy/non-ready hosts even if the number of healthy hosts come below minimum health percentage(panic threshold).",
+			" to unhealthy/non-ready hosts even if the percentage of healthy hosts fall below minimum health percentage(panic threshold).",
 	).Get()
 
 	// HTTP10 will add "accept_http_10" to http outbound listeners. Can also be set only for specific sidecars via meta.
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -106,8 +106,10 @@ var (
 
 	SendUnhealthyEndpoints = env.RegisterBoolVar(
 		"PILOT_SEND_UNHEALTHY_ENDPOINTS",
-		true,
-		"If enabled, Pilot will include unhealthy endpoints in EDS pushes and even if they are sent Envoy does not use them for load balancing.",
+		false,
+		"If enabled, Pilot will include unhealthy endpoints in EDS pushes and even if they are sent Envoy does not use them for load balancing."+
+			"  To avoid, sending traffic to non ready endpoints, enabling this flag, disables pani threshold in Envoy i.e. Envoy does not load balance requests"+
+			" to unhealthy/non-ready hosts even if the number of healthy hosts come below minimum health percentage(panic threshold).",
 	).Get()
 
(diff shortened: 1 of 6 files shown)


pilot/pkg/features/pilot.go
```diff
@@ -106,8 +106,10 @@ var (
 
 	SendUnhealthyEndpoints = env.RegisterBoolVar(
 		"PILOT_SEND_UNHEALTHY_ENDPOINTS",
-		true,
-		"If enabled, Pilot will include unhealthy endpoints in EDS pushes and even if they are sent Envoy does not use them for load balancing.",
+		false,
+		"If enabled, Pilot will include unhealthy endpoints in EDS pushes and even if they are sent Envoy does not use them for load balancing."+
+			"  To avoid, sending traffic to non ready endpoints, enabling this flag, disables pani threshold in Envoy i.e. Envoy does not load balance requests"+
```
hzxuzhonghu, line 111: pani/panic

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -658,13 +658,19 @@ func applyOutlierDetection(c *cluster.Cluster, outlier *networking.OutlierDetect
...
 	// FIXME: we can't distinguish between it being unset or being explicitly set to 0
-	if outlier.MinHealthPercent >= 0 {
+	minHealthPercent := outlier.MinHealthPercent
+	if minHealthPercent >= 0 {
 		if c.CommonLbConfig == nil {
 			c.CommonLbConfig = &cluster.Cluster_CommonLbConfig{}
 		}
-		c.CommonLbConfig.HealthyPanicThreshold = &xdstype.Percent{Value: float64(outlier.MinHealthPercent)} // defaults to 50
+		// When we are sending unhealthy endpoints, we should disble Panic Threshold. Otherwise
+		// Envoy will send traffic to "Unready" pods when all the pods are "Unready".
```
hzxuzhonghu, line 669: > all the pods are "Unready" 

This  is incorrect,

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -678,6 +684,14 @@ func defaultLBAlgorithm() cluster.Cluster_LbPolicy {
 func applyLoadBalancer(c *cluster.Cluster, lb *networking.LoadBalancerSettings, port *model.Port,
 	locality *core.Locality, proxyLabels map[string]string, meshConfig *meshconfig.MeshConfig,
 ) {
+	// Disable panic threshold when SendUnhealthyEndpoints is enabled as enabling it "may" send traffic to unready
+	// end points when load balancer is in panic mode.
+	if features.SendUnhealthyEndpoints {
+		if c.CommonLbConfig == nil {
+			c.CommonLbConfig = &cluster.Cluster_CommonLbConfig{}
```
hzxuzhonghu, line 691: Can we always set it, currently we are check it nil many places

ramaraochavali: yes. But in another PR :-)

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -658,13 +658,20 @@ func applyOutlierDetection(c *cluster.Cluster, outlier *networking.OutlierDetect
...
+	minHealthPercent := outlier.MinHealthPercent
+	if minHealthPercent >= 0 {
 		if c.CommonLbConfig == nil {
 			c.CommonLbConfig = &cluster.Cluster_CommonLbConfig{}
 		}
-		c.CommonLbConfig.HealthyPanicThreshold = &xdstype.Percent{Value: float64(outlier.MinHealthPercent)} // defaults to 50
+		// When we are sending unhealthy endpoints, we should disble Panic Threshold. Otherwise
+		// Envoy will send traffic to "Unready" pods when the percentage of healthy hosts fall
+		// below minimum health percentage.
+		if features.SendUnhealthyEndpoints {
```
sschepens, line 671: Should we always disable healthy panic threshold if SendUnhealthyEndpoints? Even if the user is explicitly enabling it in outlier detection?

I would rather have it default to disabled when SendUnhealthyEndpoints is on and no users' outlier detection is set, since it may be desired to do both things. Also would clarify users of the implications of turning SendUnhealthyEndpoints with healthy panic threshold.

ramaraochavali: > I would rather have it default to disabled when SendUnhealthyEndpoints is on and no users' outlier detection is set, since it may be desired to do both things. Also would clarify users of the implications of turning SendUnhealthyEndpoints with healthy panic threshold.

But that would send traffic to unready pods which is against K8s contract as @howardjohn mentioned.  I would rather ignore this and clarify it in the docs of outlier detection. There is no good solution that works for both.

howardjohn: I don't suppose envoy has support for 3 modes? Healthy, unhealthy (but can send in panic), and unhealthy (never send)?

that is the ideal

ramaraochavali: No it is not there. We will have to enhance it

howardjohn: Can we open an issue in envoy?

howardjohn (approved):

ramaraochavali:
https://github.com/envoyproxy/envoy/issues/22120 Envoy issue for tracking

PR #39836, opened 2022-07-08
Refactor resolution of network gateway names for more efficiency

sangwa:

This is a follow-up to https://github.com/istio/istio/pull/38781

This PR refactors the function that performs the resolution of network gateway hostnames:
* removes redundant nil checks and slice allocations introduced in the previous PR, to improve code readability
* replaces the type switch with type assertions, as we now always know what type of the DNS record we receive in a reply
* reintroduces parallel DNS resolution for A and AAAA records

As for the parallel resolution that was a point of concern in the previous PR: on each [DNS request](https://github.com/istio/istio/blob/1.14.1/pilot/pkg/model/network.go#L550) to an upstream DNS server, [when connecting](https://github.com/miekg/dns/blob/v1.1.50/client.go#L164) to that upstream, the DNS client instance [creates a new](https://github.com/miekg/dns/blob/v1.1.50/client.go#L114) net.Dialer (unless there is a [fixed predefined dialer](https://github.com/miekg/dns/blob/v1.1.50/client.go#L116) that can be passed when creating the DNS client instance, which [is not the case](https://github.com/istio/istio/blob/1.14.1/pilot/pkg/model/network.go#L536)). Since the new net.Dialer does not have a LocalAddr specified on creation, it binds to a random local port ([see LocalAddr](https://pkg.go.dev/net#Dialer)), and therefore all parallel DNS requests use different local ports and cannot interfere with each other, so they are safe to use.

The tests are not changed because there is no effective change in the implemented behaviour, only refactoring.

**To help us figure out who should review this PR, please put an  in all the areas that this PR affects.**

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/network.go
+++ b/pilot/pkg/model/network.go
@@ -540,7 +540,12 @@ func (n *networkGatewayNameCache) resolve(name string) ([]string, time.Duration)
 		mu.Lock()
 		defer mu.Unlock()
 		for _, rr := range res.Answer {
-			out = append(out, rr.String())
+			switch dnsType {
+			case dns.TypeA:
+				out = append(out, rr.(*dns.A).A.String())
+			case dns.TypeAAAA:
+				out = append(out, rr.(*dns.AAAA).AAAA.String())
+			}
 			if nextTTL := rr.Header().Ttl; nextTTL < ttl {
 				ttl = nextTTL
 			}
--- a/pilot/pkg/model/network.go
+++ b/pilot/pkg/model/network.go
@@ -529,7 +529,7 @@ func (n *networkGatewayNameCache) resolve(name string) ([]string, time.Duration)
 
 	var mu sync.Mutex
 	var wg sync.WaitGroup
-	var doResolve = func(dnsType uint16) {
+	doResolve := func(dnsType uint16) {
 		defer wg.Done()
 
 		res := n.client.Query(new(dns.Msg).SetQuestion(dns.Fqdn(name), dnsType))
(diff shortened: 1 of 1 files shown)


stevenctl:
This LGTM, but I want to double check that I understand what the concern was with parallel resolution in the previous PR.

sangwa:
> This LGTM, but I want to double check that I understand what the concern was with parallel resolution in the previous PR.

@howardjohn [mentioned](https://github.com/istio/istio/pull/38781#discussion_r868168790) that parallel DNS resolution is well known as a potential source of obscure bugs. As I understand it, when UDP datagrams with different DNS queries are sent in parallel from the same source socket, there is no guarantee that replies will be received in the same order neither an effective way to discriminate which reply corresponds to which query, so a race condition arises and replies can be mixed up.

In the current implementation this should not happen, as I mentioned above, since the DNS client library set up the way it is in the code will use new sockets with random ports for each query.

pilot/pkg/model/network.go
```diff
@@ -524,45 +524,39 @@ func stringSliceEqual(a, b []string) bool {
...
+			case dns.TypeAAAA:
+				out = append(out, rr.(*dns.AAAA).AAAA.String())
+			}
+			if nextTTL := rr.Header().Ttl; nextTTL < ttl {
+				ttl = nextTTL
+			}
 		}
 	}
+
+	wg.Add(2)
```
howardjohn, line 555: nit: errgroup may be simpler?

stevenctl (approved):

PR #39839, opened 2022-07-08
custom hpa spec for ingress

ristox1:
I have a use case where I would like to be able to add/change the behaviour for the hpa for the istio ingress gateway

### DIFF
--- a/manifests/charts/gateways/istio-ingress/templates/autoscale.yaml
+++ b/manifests/charts/gateways/istio-ingress/templates/autoscale.yaml
@@ -19,11 +19,15 @@ spec:
     apiVersion: apps/v1
     kind: Deployment
     name: {{ $gateway.name }}
+{{- if not .Values.customAustoscaleSpec }}
   metrics:
   - type: Resource
     resource:
       name: cpu
       targetAverageUtilization: {{ $gateway.cpu.targetAverageUtilization }}
+{{- else }}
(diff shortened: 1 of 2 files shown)


zirain:
https://github.com/istio/istio/issues/39079#issuecomment-1135133605

BTW, doc for helm user https://github.com/istio/istio.io/pull/11480/files

ristox1:
> [#39079 (comment)](https://github.com/istio/istio/issues/39079#issuecomment-1135133605)
> 
> BTW, doc for helm user https://github.com/istio/istio.io/pull/11480/files

Yeah I'm aware that we can add


> [#39079 (comment)](https://github.com/istio/istio/issues/39079#issuecomment-1135133605)
> 
> BTW, doc for helm user https://github.com/istio/istio.io/pull/11480/files

Hey, thanks for your comment! I understand that it's impossible to add every possible configuration, that's why adding an option for users to apply their own configs like in the PR is maybe plausible?

As an end-user, I chose to go with the official istio chart instead of a custom chart in order to reduce maintenance and complexity when upgrading versions as well as ease of deployment. I'm way over my head here thinking maybe creating/editing existing templates with such options like in this PR would be considered although I know it adds a lot of complexity to fix previous versions.

If the official charts don't offer this kind of flexibility it kind of ruins the idea of using an official chart for me/my use case :) Looking forward to your thoughts.

PR #39840, opened 2022-07-08
[tf] load templates outside of init func

akshayjnambiar:
This makes sure that the flags are parsed and the new values mentioned in the flags are used.


nmittler:
Out of curiosity, do you know why init wasn't being called?

stevenctl:
The issue is that the `init` that loads the template is called before `flag.Parse`.

stevenctl (approved):
LGTM

nmittler (approved):

PR #39854, opened 2022-07-11
possible dependent config leak

ramaraochavali:
Built on top of https://github.com/istio/istio/pull/39853.

One possible leak I think is when a configUpdate happens, only one of dependent config change, might trigger a cache key to evicted but the key might exist in other config reference. This PR cleans up key on evict from dependent config list. 

Main changes are in `onEvict` - always clean index in `onEvict`.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pilot/pkg/model/xds_cache.go
```diff
@@ -185,19 +188,22 @@ func (l *lruCache) recordDependentConfigSize() {
 }
 
 // This is the callback passed to LRU, it will be called whenever a key is removed.
-func (l *lruCache) evict(k interface{}, v interface{}) {
+func (l *lruCache) onEvict(k interface{}, v interface{}) {
 	if features.EnableXDSCacheMetrics {
-		xdsCacheEvictions.Increment()
-	}
-
-	if !l.evicted {
```
sschepens, line 193: yes, this seems like a clear leak, the tests I had on another PR should've faild with this change

sschepens: i see that https://github.com/istio/istio/pull/39688 has been reverted, which changed the behaviour of the cache and made my tests invalid, I could now probably merge them.

howardjohn (approved):
lgtm, can you fix the conflict though?

adiprerepa (approved):

pilot/pkg/model/xds_cache.go
```diff
@@ -341,9 +347,9 @@ func (l *lruCache) Clear(configs map[ConfigKey]struct{}) {
 	defer l.mu.Unlock()
 	l.token = CacheToken(time.Now().UnixNano())
 	// not to call evict on keys that are removed actively
-	l.evicted = false
+	l.evictedOnClear = true
```
hzxuzhonghu, line 350: remove the comment

Issue #39598, opened 2022-06-24
DNSProxying support parallel queries to all upstream nameservers


fatedier:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**
I enabled DNSProxying in istio and it will forward external domain queries to upstream nameservers in sequence and fall to the next after one times out.

Can we support parallel queries optional? It will improve performance and reliability of DNS lookups for us.

I see these is already a TODO [comment](https://github.com/istio/istio/blob/master/pkg/dns/client/dns.go#L388). I can contribute to this feature if you think it's no problem.

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[X] Networking
[X] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

ramaraochavali:
Sure

PR #39857, opened 2022-07-11
dns: added DNSForwardParallel to support sending parallel queries to all name servers

fatedier:
Fixes: #39598

### DIFF
--- a/pilot/cmd/pilot-agent/options/options.go
+++ b/pilot/cmd/pilot-agent/options/options.go
@@ -98,6 +98,9 @@ var (
 	DNSCaptureAddr = env.RegisterStringVar("DNS_PROXY_ADDR", "localhost:15053",
 		"Custom address for the DNS proxy. If it ends with :53 and running as root allows running without iptable DNS capture")
 
+	DNSForwardParallel = env.RegisterBoolVar("DNS_FORWARD_PARALLEL", false,
+		"If set to true, agent will send parallel DNS queries to all upstream nameservers")
+
 	// Ability of istio-agent to retrieve proxyConfig via XDS for dynamic configuration updates
 	enableProxyConfigXdsEnv = env.RegisterBoolVar("PROXY_CONFIG_XDS_AGENT", false,
 		"If set to true, agent retrieves dynamic proxy-config updates via xds channel").Get()
--- a/pkg/dns/client/dns.go
+++ b/pkg/dns/client/dns.go
@@ -15,6 +15,7 @@
 package client
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"os"
@@ -53,7 +54,8 @@ type LocalDNSServer struct {
 	proxyDomain      string
 	proxyDomainParts []string
 
-	respondBeforeSync bool
+	respondBeforeSync         bool
+	forwardToUpstreamParallel bool
 }
 
 // LookupTable is borrowed from https://github.com/coredns/coredns/blob/master/plugin/hosts/hostsfile.go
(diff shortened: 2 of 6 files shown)


pkg/dns/client/dns.go
```diff
@@ -396,15 +402,59 @@ func (h *LocalDNSServer) queryUpstream(upstreamClient *dns.Client, req *dns.Msg,
...
 	return response
 }
 
+// queryUpstreamParallel will send parallel queries to all nameservers and return first response immediately.
+func (h *LocalDNSServer) queryUpstreamParallel(upstreamClient *dns.Client, req *dns.Msg, scope *istiolog.Scope) *dns.Msg {
+	responseCh := make(chan *dns.Msg, 1)
+	errCh := make(chan error, len(h.resolvConfServers))
+
+	queryOne := func(upstream string) {
+		cResponse, _, err := upstreamClient.Exchange(req, upstream)
```
howardjohn, line 418: should we use ExchangeContext so we can cancel other requests once one completes?

howardjohn (commented):
At a high level seems reasonable. I would like to see some examples of other go implementations (coredns, go builtin, tailscale, etc all have DNS implementations) to make sure we are following any standards, best practices, edge cases, etc here.

In particular, using alpine as an example of why this is good behavior doesn't seem super compelling given how many times i have heard of alpine causing DNS issues. It may still be, just a bit suspicious of using only alpine as a source of motivation here.

pkg/dns/client/dns.go
fatedier: I tested it but failed. `ExchangeContext` can't be canceled by context. It just extract deadline from the context.

fatedier:
> At a high level seems reasonable. I would like to see some examples of other go implementations (coredns, go builtin, tailscale, etc all have DNS implementations) to make sure we are following any standards, best practices, edge cases, etc here.
> 
> In particular, using alpine as an example of why this is good behavior doesn't seem super compelling given how many times i have heard of alpine causing DNS issues. It may still be, just a bit suspicious of using only alpine as a source of motivation here.

@howardjohn I agree. Parallel DNS queries are not suitable for all scenarios, so I make it optional and not the default behavior. In my scenario, I configure nameservers the ip of node-local-dns and coredns. When the node-local-dns is down, requests to external service will delay up to 8 seconds and failed.  This is almost unacceptable in microservices and lead to multiple nameservers meaningless.

I searched some documents:

* [dnsmasq](https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html#:~:text=etc/resolv.conf-,%2D%2Dall%2Dservers,-By%20default%2C%20when): `--all-servers` Setting this flag forces dnsmasq to send all queries to all available servers. The reply from the server which answers first will be returned to the original requester.
* [tailscale](https://tailscale.com/kb/1194/dns-resolver-order/?q=dns): Tailscale will either proxy all DNS requests—in which case we query all nameservers in parallel and use the quickest response—or defer to the operating system, which we have no control over.
* [RFC 6731](https://datatracker.ietf.org/doc/html/rfc6731.txt#appendix-A.1): It's about RDNSS. I'm not very sure if it's related.
* [fanout](https://coredns.io/explugins/fanout/): A coredns external plugin, parallel proxying DNS messages to upstream resolvers.

By the way, coredns's forward plugin also support [health check](https://coredns.io/plugins/forward/#:~:text=on%20sequential%20ordering.-,health_check,-configure%20the%20behaviour) of the upstream servers. Maybe it's an alternative but don't resolve the problem completely.

pilot/cmd/pilot-agent/options/options.go
```diff
@@ -98,6 +98,9 @@ var (
 	DNSCaptureAddr = env.RegisterStringVar("DNS_PROXY_ADDR", "localhost:15053",
 		"Custom address for the DNS proxy. If it ends with :53 and running as root allows running without iptable DNS capture")
 
+	DNSForwardParallel = env.RegisterBoolVar("ISTIO_META_DNS_FORWARD_PARALLEL", false,
```
howardjohn, line 101:
```suggestion
	DNSForwardParallel = env.RegisterBoolVar("DNS_FORWARD_PARALLEL", false,
```

and clear the model/context.go change

The istio_meta_ and NodeMetadata should only be needed when we are sending this data to Istiod. But this option only impacts the agent locally so no need

howardjohn (approved):

pilot/cmd/pilot-agent/options/options.go
fatedier: Got it, thanks.

PR #39875, opened 2022-07-12
fix bug for the istio-agent wait command to make sure that it is really timeout

zhlsunshine:

Fix bug for the istio-agent wait command to make sure that it is really timeout.
In old code, both `time.Now().Before(timeoutAt)` and `time.Sleep(time.Duration(periodMillis) * time.Millisecond)` determine the istio-agent wait command is time out or not. Whatever the `periodMillis` user sets, there may be a situation that wait command is not really timeout, but timeout because current time may be `After` the `timeoutAt` due to `time.Sleep`

In new code, introduce a channel for time counter and `select` for this timeout check, then remove `periodMillis` option.

### DIFF
--- a/pilot/cmd/pilot-agent/app/wait.go
+++ b/pilot/cmd/pilot-agent/app/wait.go
@@ -28,9 +28,8 @@ import (
 var (
 	timeoutSeconds       int
 	requestTimeoutMillis int
+	periodMillis         int
 	url                  string
-
-	timeoutCheck = make(chan struct{})
 
 	waitCmd = &cobra.Command{
 		Use:   "wait",
@@ -42,15 +41,13 @@ var (
 			log.Infof("Waiting for Envoy proxy to be ready (timeout: %d seconds)...", timeoutSeconds)
 
 			var err error
-			go func() {
-				time.Sleep(time.Duration(timeoutSeconds) * time.Second)
-				timeoutCheck <- struct{}{}
-			}()
+			timeout := time.After(time.Duration(timeoutSeconds) * time.Second)
+
 			for {
 				select {
-				case <-timeoutCheck:
+				case <- timeout:
 					return fmt.Errorf("timeout waiting for Envoy proxy to become ready. Last error: %v", err)
-				default:
+				case <- time.After(time.Duration(periodMillis) * time.Second):
 					err = checkIfReady(client, url)
 					if err == nil {
 						log.Infof("Envoy is ready!")
@@ -86,5 +83,6 @@ func checkIfReady(client *http.Client, url string) error {
 func init() {
 	waitCmd.PersistentFlags().IntVar(&timeoutSeconds, "timeoutSeconds", 60, "maximum number of seconds to wait for Envoy to be ready")
 	waitCmd.PersistentFlags().IntVar(&requestTimeoutMillis, "requestTimeoutMillis", 500, "number of milliseconds to wait for response")
+	waitCmd.PersistentFlags().IntVar(&periodMillis, "periodMillis", 500, "number of milliseconds to wait between attempts")
 	waitCmd.PersistentFlags().StringVar(&url, "url", "http://localhost:15021/healthz/ready", "URL to use in requests")
 }
--- a/pilot/cmd/pilot-agent/app/wait.go
+++ b/pilot/cmd/pilot-agent/app/wait.go
@@ -28,8 +28,9 @@ import (
 var (
 	timeoutSeconds       int
 	requestTimeoutMillis int
-	periodMillis         int
 	url                  string
+
+	timeoutCheck = make(chan struct{})
 
 	waitCmd = &cobra.Command{
 		Use:   "wait",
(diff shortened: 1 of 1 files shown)


pilot/cmd/pilot-agent/app/wait.go
```diff
@@ -79,6 +86,5 @@ func checkIfReady(client *http.Client, url string) error {
 func init() {
 	waitCmd.PersistentFlags().IntVar(&timeoutSeconds, "timeoutSeconds", 60, "maximum number of seconds to wait for Envoy to be ready")
 	waitCmd.PersistentFlags().IntVar(&requestTimeoutMillis, "requestTimeoutMillis", 500, "number of milliseconds to wait for response")
-	waitCmd.PersistentFlags().IntVar(&periodMillis, "periodMillis", 500, "number of milliseconds to wait between attempts")
```
GregHanson, line 82: do we need to add a deprecation warning here rather than removing?  I'm not sure how many people are using this but we don't want to pull the rug out from anyone.

Also since a CLI flag is getting removed, I think a release note should be added so it gets called out in the release/upgrade notes

howardjohn: Similar comment, but I don't get why we are removing or deprecating at all. We can have a poll period and still exit early during sleep. Just do

```go
timeout := time.After(timeoutSeconds)
for {
 select{
  case <-timeout: ...
  case <- time.After(periodMillis): ...
}
```

That being said, I would support a PR to deprecate these and replace them with `period` and `requestTimeout` (without Millis... we can use DurationVar)

howardjohn (changes requested):

pilot/cmd/pilot-agent/app/wait.go
zhlsunshine: > do we need to add a deprecation warning here rather than removing? I'm not sure how many people are using this but we don't want to pull the rug out from anyone.
> 
> Also since a CLI flag is getting removed, I think a release note should be added so it gets called out in the release/upgrade notes

It's a  good idea, I will add a deprecation warning and release note, thanks @GregHanson !

zhlsunshine: > Similar comment, but I don't get why we are removing or deprecating at all. We can have a poll period and still exit early during sleep. Just do
> 
> ```go
> timeout := time.After(timeoutSeconds)
> for {
>  select{
>   case <-timeout: ...
>   case <- time.After(periodMillis): ...
> }
> ```
> 
> That being said, I would support a PR to deprecate these and replace them with `period` and `requestTimeout` (without Millis... we can use DurationVar)

Hi @howardjohn, if the checking logic is filled in the second case(`case <- time.After(periodMillis)`), there may be a timeout notification earlier than time.After notification, in this situation it would not be the really timeout, and it's the same as before, make sense?

howardjohn: no I don't get it.

hzxuzhonghu: I cannot get what bug it is fixing or refactor?

pilot/cmd/pilot-agent/app/wait.go
```diff
@@ -41,17 +42,23 @@ var (
...
 
 			var err error
-			timeoutAt := time.Now().Add(time.Duration(timeoutSeconds) * time.Second)
-			for time.Now().Before(timeoutAt) {
-				err = checkIfReady(client, url)
-				if err == nil {
-					log.Infof("Envoy is ready!")
-					return nil
+			go func() {
+				time.Sleep(time.Duration(timeoutSeconds) * time.Second)
```
hzxuzhonghu, line 46: No need start a thread.

ch := time.After()

pilot/cmd/pilot-agent/app/wait.go
zhlsunshine: Although the reproduce was not stable and very rare, I encountered this rare situation when doing verifications, the checking client showed timeout, but the gRPC server created the UDS file. 
So I analyze that this may be caused by the `sleep`. Let's assume that there may be 450 milliseconds left to be timeout, but the checking UDS file is still not ready, so the checking process will go to sleep 500 milliseconds. However, the timeout will be triggered once the sleep is over.

pilot/cmd/pilot-agent/app/wait.go
zhlsunshine: okay, I will remove the unnecessary thread and use `time.After`, thanks @hzxuzhonghu!

pilot/cmd/pilot-agent/app/wait.go
zhlsunshine: Hi @howardjohn & @hzxuzhonghu, described as above

pilot/cmd/pilot-agent/app/wait.go
```diff
@@ -41,17 +42,23 @@ var (
...
+				timeoutCheck <- struct{}{}
+			}()
+			for {
+				select {
+				case <-timeoutCheck:
+					return fmt.Errorf("timeout waiting for Envoy proxy to become ready. Last error: %v", err)
+				default:
+					err = checkIfReady(client, url)
```
hzxuzhonghu, line 54: Then it is a dead loop calling `checkIfReady`?

zhlsunshine: Yeah, but both timeout and non error check path can stop this loop.

hzxuzhonghu: I am concerned with the high cpu usage maybe

zhlsunshine: yeah, however, as you know, `pilot-agent wait` is running via k8s `postStart` mechanism, so application container is suspended during this wait command is not over. So the CPU usage may not be the key point.

pilot/cmd/pilot-agent/app/wait.go
howardjohn: That's why I recommend using a channel then you can cancel during the poll sleep

pilot/cmd/pilot-agent/app/wait.go
howardjohn: This still wastes node cpu

pilot/cmd/pilot-agent/app/wait.go
zhlsunshine: Yeah, although `select` will randomly pick the case switch, `timeout` and `sleep` channel, I agree that use a channel would be a better method.

pilot/cmd/pilot-agent/app/wait.go
zhlsunshine: so @hzxuzhonghu & @howardjohn,  as mentioned above, using `time.After` channel in select would be a better choice. I will submit again!

pilot/cmd/pilot-agent/app/wait.go
zhlsunshine: > do we need to add a deprecation warning here rather than removing? I'm not sure how many people are using this but we don't want to pull the rug out from anyone.
> 
> Also since a CLI flag is getting removed, I think a release note should be added so it gets called out in the release/upgrade notes

Hi @GregHanson, there are some discussion with john and zhonghu, `periodMillis` flag may need to be changed instead of removing. Hope you know this, thanks for your good suggestion before!

Issue #39720, opened 2022-06-30
TestReachability flake; cluster warming


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/integ-security-multicluster_istio_postsubmit/1542555639413215232

Still debugging, but looks like all clusters are warming

howardjohn:
Investigated this one a bit.

warm: 2022-06-30T17:32:43Z/226
219, all eds
ready: 2022-06-30T17:32:42Z/224 and 226 - it is a mix; 114 on 226 and 219 on 224
28 eds ready, 219 unready

pushes look same for good and bad

bad disconnects at 17:32:55.308978Z, reconects 2s later
test fails around 33:22. starts at 33:02. So its all after reconnect

reconnect has EDS first ????.

Pushed resources seem the same on reconnect

all the ready ones are vm, fake stateful, headless

VM is eds but empty?

---

So a few conclusions

* The broken pods are reconnecting to XDS and getting stuck warming. The get the same number of resources, but somehow warming. The super weird thing is the XDS request order is out of order: EDS,LDS,RDS,CDS.
* VMs are somehow 0 endpoints? also seeing this:
```
$ rg 'transport: authentication handshake failed'
_test_context/echo2-3-10889-state3295745563/cross-network-primary/vm-v1-749fc9946-97pcf_istio-proxy.envoy.err.log
20422:2022-06-30T17:33:31.211172Z       warning envoy config    StreamAggregatedResources gRPC config stream closed since 30s ago: 14, connection error: desc = "transport: authentication handshake failed: read tcp 10.30.0.31:47650->172.18.3.255:15012: read: connection reset by peer"
20423:2022-06-30T17:33:37.756010Z       warning envoy config    StreamAggregatedResources gRPC config stream closed since 36s ago: 14, connection error: desc = "transport: authentication handshake failed: read tcp 10.30.0.31:49634->172.18.3.255:15012: read: connection reset by peer"
20424:2022-06-30T17:33:40.458735Z       warning envoy config    StreamAggregatedResources gRPC config stream closed since 39s ago: 14, connection error: desc = "transport: authentication handshake failed: read tcp 10.30.0.31:50206->172.18.3.255:15012: read: connection reset by peer"

_test_context/echo2-3-10889-state3295745563/remote/delta-v1-c44cc99b6-tl5zq_istio-proxy.log
[...]

PR #39889, opened 2022-07-12
xds: improve incremental logging

howardjohn:
Currently, we do not log incremental pushes at Info level. The intent
behind this is to avoid spam when we have large endpoint churn. However,
because we also do incremental pushes for Full pushes now, we are also
hiding these logs. These logs are both critical to debugging (things
like https://github.com/istio/istio/issues/39720, etc) and not spammy --
while the `Full=false` pushes may add thousands of messages, this change
only adds at most 1 log per push/proxy. For these types of pushes I
don't see a benefit to excluding only EDS.

Additionally, fix SDS to correctly assert it is incremental (when it
is).

### DIFF
--- a/pilot/pkg/xds/sds.go
+++ b/pilot/pkg/xds/sds.go
@@ -148,7 +148,10 @@ func (s *SecretGen) Generate(proxy *model.Proxy, w *model.WatchedResource, req *
 			results = append(results, res)
 		}
 	}
-	return results, model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", cached, cached+regenerated)}, nil
+	return results, model.XdsLogDetails{
+		Incremental:    updatedSecrets != nil,
+		AdditionalInfo: fmt.Sprintf("cached:%v/%v", cached, cached+regenerated),
+	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


pilot/pkg/xds/sds.go
```diff
@@ -148,7 +148,10 @@ func (s *SecretGen) Generate(proxy *model.Proxy, w *model.WatchedResource, req *
 			results = append(results, res)
 		}
 	}
-	return results, model.XdsLogDetails{AdditionalInfo: fmt.Sprintf("cached:%v/%v", cached, cached+regenerated)}, nil
+	return results, model.XdsLogDetails{
+		Incremental:    updatedSecrets != nil,
```
nmittler, line 152: Is this fixing a bug?

howardjohn: Yes but just one impacting logging so not critical at all

nmittler: Oof missed that, sorry for the noise :)

nmittler (approved):

adiprerepa (approved):

ramaraochavali:
LGTM

PR #39901, opened 2022-07-13
Cleanup non standard  token prefix

hzxuzhonghu:


howardjohn (commented):
I believe we are still using this code, is there any reason it needs to be deleted?

hzxuzhonghu:
@costinm Do you know when is this prefix used?

PR #39904, opened 2022-07-13
remove relative operation error for envoy filter MERGE operation

ramaraochavali:
A MERGE operation is not relative. A simple Envoy filter like this 
```
  configPatches:
  - applyTo: NETWORK_FILTER
    match:
      context: SIDECAR_OUTBOUND
      listener:
        filterChain:
          filter:
            name: envoy.filters.network.redis_proxy
        portNumber: 12000
    patch:
      operation: MERGE
      value:
        name: envoy.filters.network.redis_proxy
        typed_config:
          '@type': type.googleapis.com/envoy.extensions.filters.network.redis_proxy.v3.RedisProxy
          settings:
            op_timeout: 1s
```
produces analyzer error which is incorrect. We are asking to specifically merge a listener - no relativity here.

```
  validationMessages:
  - documentationUrl: https://istio.io/v1.15/docs/reference/config/analysis/ist0151/
    level: WARNING
    type:
      code: IST0151
 ```

Same problem exists for `REPLACE` and `REMOVE` cases as well.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


GregHanson (approved):

GregHanson:
cc @AndreaM12345 FYI since I think you added this test

GregHanson:
but MERGE/REPLACE/REMOVE do not edit the existing array - they modify an element existing in the array already.  The order of elements in the array does not change based on these operations correct? Per the wording for IST0151:
>"This EnvoyFilter does not have a priority and has a relative patch operation set which can cause the EnvoyFilter not to be applied. Using the INSERT_FIRST or ADD option or setting the priority may help in ensuring the EnvoyFilter is applied correctly."

Or does setting the priority change the element position after the operation is performed?

howardjohn:
But if you MERGe then replace vs the inverted order it does matter, right?

hzxuzhonghu:
I think we should improve patch logging, rather than using analysis to warn any potential issue

ramaraochavali:
> I think we should improve patch logging, rather than using analysis to warn any potential issue

I agree completely. Either we improve analyzer to detect cases where MERGE needs priority or improve patching. Asking users to set priority for simple merge as shown in the PR description is not correct and that warning is actually giving a wrong impression.

ramaraochavali:
> i think the issue comes up with an upgrade because filters get replaced but you still want something to happen in a certain order. The docs mention this with a reference to the istio.stats filter https://istio.io/latest/docs/ops/common-problems/network-issues/#unchanged-envoy-filter-configuration-suddenly-stops-working

That was for INSERT_FIRST though which is relative.

PR #39915, opened 2022-07-13
tf: improve dumping

howardjohn:
This was initially a bug fix but then I realized I was misguided.
However, along the way I made some improvements to the dumping logic.

* We were dumping istio-system 2x; fix that
* We were using PodExec which is slow. Switch to port forward which is
  10x+ faster. Also use a more robust mechanism to resrve ports
* Dump in parallel
* Add some better logging around dumping
* Clear managedFields from events for size/readability

### DIFF
--- a/pkg/kube/mock_client.go
+++ b/pkg/kube/mock_client.go
@@ -80,6 +80,9 @@ type MockClient struct {
 	IstiodVersion     string
 }
 
+func (c MockClient) SetPortManager(manager PortManager) {
+}
+
 func (c MockClient) WaitForCacheSync(stop <-chan struct{}, cacheSyncs ...cache.InformerSynced) bool {
 	return WaitForCacheSync(stop, cacheSyncs...)
 }
--- a/pkg/test/framework/components/namespace/namespace.go
+++ b/pkg/test/framework/components/namespace/namespace.go
@@ -32,6 +32,9 @@ type Config struct {
 	Revision string
 	// Labels to be applied to namespace
 	Labels map[string]string
+	// SkipDump, if enabled, will disable dumping the namespace. This is useful to avoid duplicate
+	// dumping of istio-system.
+	SkipDump bool
 }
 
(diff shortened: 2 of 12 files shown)


pkg/kube/mock_client.go
```diff
@@ -80,6 +80,9 @@ type MockClient struct {
 	IstiodVersion     string
 }
 
+func (c MockClient) SetPortManager(manager PortManager) {
```
nmittler, line 83: should this panic, so it's obvious from the tests if we call it? Or do we expect it to be called?

pkg/test/framework/components/namespace/namespace.go
```diff
@@ -32,6 +32,9 @@ type Config struct {
 	Revision string
 	// Labels to be applied to namespace
 	Labels map[string]string
+	// SkipDump, if enabled, will disable dumping the namespace. This is useful to avoid duplicate
```
nmittler, line 35: Is istio-system the main reason for introducing this? Can you expand on why we're dumping istio-system twice? Seems like we should fix that without needing this flag, no?

pkg/test/kube/dump.go
```diff
@@ -271,6 +275,12 @@ func DumpPodEvents(_ resource.Context, c cluster.Cluster, workDir, namespace str
 			return
 		}
 
+		for i := range list.Items {
```
nmittler, line 278: Can you clarifying what you're doing here an way in a comment? I'm assuming the managed fields bloat the dump?

pkg/test/framework/components/namespace/namespace.go
howardjohn: `Istio` component has its own dumping logic which is a superset of the namespace dumping. But since we call ClaimNamespace now we end up dumping the istio-system namespaces.

In theory we could make the Istio component not duplicate the ns dumping parts, but that would mean its in a different folder (harder to debug) and relies on ClaimNamespace being called (doesn't always happen)

nmittler (approved):

PR #39916, opened 2022-07-13
xds: fix issue when skipping first request

howardjohn:
The bug (example failure:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/39896/integ-cni_istio/1546984256612339712)
* ingressgateway has 1 Gateway
* Gateway is removed
* envoy disconnects
* Envoy reconnects, requests RDS. This hits the INIT/RECONNECT flow.
* RDS hits 'Gateway missing for route' path and gives no response at all
* Next RDS request, we get a "stale nonce" since we have no previously
  sent nonce
* Envoy stuck forever

The fix:
* Remove code path to return empty route instead of no route (matching
  other paths)
* Add assertions to ensure that we don't send empty response to requests
  and that we never count a "stale nonce" if we somehow have no
previously sent nonce, to ensure there aren't any other issues

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -472,8 +472,12 @@ func (pr *PushRequest) CopyMerge(other *PushRequest) *PushRequest {
 	return merged
 }
 
+func (pr *PushRequest) IsRequest() bool {
+	return len(pr.Reason) == 1 && pr.Reason[0] == ProxyRequest
+}
+
 func (pr *PushRequest) PushReason() string {
-	if len(pr.Reason) == 1 && pr.Reason[0] == ProxyRequest {
+	if pr.IsRequest() {
 		return " request"
 	}
 	return ""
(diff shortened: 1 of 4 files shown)


hzxuzhonghu:
When met `Gateway missing for route`, isn't the listener will be removed too?

howardjohn:
No, the issue happens because listener is still present for 45s (draining) after removal of the Gateway

/test integ-assertion_istio

(assertion failed due to unrelated issue, fixed in another pr)

hzxuzhonghu:
IC, should we fix envoy which should not send rds request for a listener in draining state?

howardjohn:
It seems reasonable to me for envoy to do that otherwise it cannot serve traffic on the draining listener appropriately. We only set drain to 45s but some may do it for hours even where it makes more sense

ramaraochavali:
LGTM. Would be good if we can add a test case that covers this. But if you want to add later, I am OK to approve this. Let me know.

howardjohn:
will follow up with more tests, right now I want to make sure we can
unblock all the failing e2e tests

ramaraochavali (approved):

SpecialYang:
@howardjohn When we met the case `Gateway missing for route`, our gateway respond 404 for a while. We think this pr is the reason. 

Is it right that we push empty RDS? 
https://github.com/istio/istio/blob/ecb615dd104f65fc17d8cf78168a9eefafa0a66f/pilot/pkg/networking/core/v1alpha3/gateway.go#L354-L360

PR #39922, opened 2022-07-13
Add test for unfixed SNI overlapping bug

howardjohn:
I was working on a fix but don't have time to complete for now. At least
we can check in the test case to codify the bug and make it simpler when
we fix it eventually.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/gateway_simulation_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/gateway_simulation_test.go
@@ -1896,5 +1896,72 @@ spec:
 					},
 				},
 			},
-		})
+		},
+		simulationTest{
+			name:           "overlapping SNI match",
+			skipValidation: true, // TODO: https://github.com/istio/istio/issues/39921
+			config: `apiVersion: networking.istio.io/v1beta1
+kind: Gateway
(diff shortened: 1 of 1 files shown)


ramaraochavali (approved):

hzxuzhonghu:
This is not complete

howardjohn:
it's not supposed to be complete it's just adding a test. Is there any
concern with adding a unit test?

pilot/pkg/networking/core/v1alpha3/gateway_simulation_test.go
```diff
@@ -1896,5 +1896,72 @@ spec:
...
+  ports:
+  - number: 443
+    targetPort: 443
+    name: https
+    protocol: HTTPS
+  resolution: STATIC
+  location: MESH_INTERNAL
+  endpoints:
+  - address: 1.1.1.1`,
+			calls: []simulation.Expect{},
```
hzxuzhonghu, line 1964: Am i reading wrong? There is not expected result here?

howardjohn: There are none right now since its broken. Once we fix it we can declare what is expected

PR #39896, opened 2022-07-12
Tf/debug logs

howardjohn:
**Please provide a description of this PR:**

PR #39937, opened 2022-07-14
xds: always respond to EDS request after CDS requests

howardjohn:
This fixes an edge case that happens to be triggering many test flakes. https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/39896/integ-security-multicluster_istio/1547605105740615680 for example.

What happens is:

```
1. Envoy connect
2. Envoy REQ EDS, nonce=e1
3. Istiod pushes EDS, nonce=e2
4. Envoy REQ CDS
5. Istio pushes CDS
6. Envoy REQ EDS, nonce=e2
7. Istiod treats this as an ACK (from push (3))
```

Then envoy cluster is struck warming.

This solution seems super hacky, but it seems to work (in https://github.com/istio/istio/pull/39896 tested 6x without flakes - prior to this it fails like 50%+ of the time). Basically, if we get a CDS request, we always respond to the next EDS request. This ONLY occurs when EDS is the first request and at worst causes 1 extra EDS push (which I don't even think its "extra" - AFAIK its required in all cases).

I will be retesting https://github.com/istio/istio/pull/39896  overnight so should give us about 10 more trials before tomorrow morning if we want more validation this fixes this issue.

BTW - the reason this is suddenly an issue is because we started testing a lower max connection age. These bugs were always present, but they only happened on xds reconnects which never happened in our test environments.

### DIFF
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -394,6 +394,14 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 		log.Debugf("ADS:%s: INIT/RECONNECT %s %s %s", stype, con.conID, request.VersionInfo, request.ResponseNonce)
 		con.proxy.Lock()
 		con.proxy.WatchedResources[request.TypeUrl] = &model.WatchedResource{TypeUrl: request.TypeUrl, ResourceNames: request.ResourceNames}
+		// Due to https://github.com/envoyproxy/envoy/issues/13009, we must always send an EDS response when CDS is pushed.
+		// However, during reconnect scenarios, Envoy may fail to send EDS requests that we can reasonably identify as non-ACKs.
+		// Instead, we force the next EDS request to always respond, rather than an ACK, to ensure we send EDS.
+		if request.TypeUrl == v3.ClusterType {
+			if eds, f := con.proxy.WatchedResources[v3.EndpointType]; f {
+				eds.AlwaysRespond = true
+			}
+		}
 		con.proxy.Unlock()
 		return true, emptyResourceDelta
 	}
@@ -423,6 +431,8 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 	con.proxy.WatchedResources[request.TypeUrl].NonceAcked = request.ResponseNonce
 	con.proxy.WatchedResources[request.TypeUrl].NonceNacked = ""
 	con.proxy.WatchedResources[request.TypeUrl].ResourceNames = request.ResourceNames
+	alwaysRespond := previousInfo.AlwaysRespond
+	previousInfo.AlwaysRespond = false
 	con.proxy.Unlock()
 
 	// Envoy can send two DiscoveryRequests with same version and nonce
@@ -440,6 +450,10 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 		// times, we treat subsequent requests as something we should respond to.
 		if features.PushOnRepeatNonce && request.ResponseNonce == previousNonceAcked {
 			log.Debugf("ADS:%s: REQ %s Repeated nonce received %s", stype, con.conID, request.ResponseNonce)
+			return true, emptyResourceDelta
+		}
+		if alwaysRespond {
+			log.Infof("ADS:%s: REQ %s forced response", stype, con.conID)
 			return true, emptyResourceDelta
 		}
 		log.Debugf("ADS:%s: ACK %s %s %s", stype, con.conID, request.VersionInfo, request.ResponseNonce)
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -355,6 +355,11 @@ type WatchedResource struct {
 	// NonceNacked is the last nacked message. This is reset following a successful ACK
 	NonceNacked string
 
+	// AlwaysRespond, if true, will ensure that even when a request would otherwise be treated as an
+	// ACK, it will be responded to. Typically, this should be set to 'false' after response; keeping it
+	// true would likely result in an endless loop.
+	AlwaysRespond bool
+
 	// LastSent tracks the time of the generated push, to determine the time it takes the client to ack.
 	LastSent time.Time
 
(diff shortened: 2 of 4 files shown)


adiprerepa (approved):

hzxuzhonghu:
Doesn't you have fixed many similar bugs? All seems a little hacky, which should be well resolved by envoy.

pilot/pkg/xds/ads.go
```diff
@@ -394,6 +394,14 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 		log.Debugf("ADS:%s: INIT/RECONNECT %s %s %s", stype, con.conID, request.VersionInfo, request.ResponseNonce)
 		con.proxy.Lock()
 		con.proxy.WatchedResources[request.TypeUrl] = &model.WatchedResource{TypeUrl: request.TypeUrl, ResourceNames: request.ResourceNames}
+		// Due to https://github.com/envoyproxy/envoy/issues/13009, we must always send an EDS response when CDS is pushed.
+		// However, during reconnect scenarios, Envoy may fail to send EDS requests that we can reasonably identify as non-ACKs.
```
hzxuzhonghu, line 398: I understand when reconnect, the connection  between envoy->xds proxy  will be closed too.  Then envoy will start the xds requests as initially. How could this happen?

howardjohn:
Envoy makes no guarantees the first request will be CDS. it surprised me
too...

If someone fixes this in Envoy I would be happy, but it seems like an
extremely large effort. The bug I previously fixed only covered a subset of
the issue - I was planning to modify this PR to revert the original fix but
didn't have a chance before it merged. I'll send a followup

hzxuzhonghu:
Me too, feel so sad about that

ramaraochavali:
> Envoy makes no guarantees the first request will be CDS. it surprised me
too...

It guarantees on initial stream but not on reconnected stream (disconnect and reconnect case)? If that is the case that makes sense to me - otherwise it is a problem

> BTW - the reason this is suddenly an issue is because we started testing a lower max connection age. These bugs were always present, but they only happened on xds reconnects which never happened in our test environments.

This is very useful information. What is the reason for this change? BTW, I am not saying it is bad - it is actually good that we are catching these corner cases, but curious what it is.

My only worry with these fixes is now we are too many if else conditions to take care of - hopefully all/most of it is in INIT/RECONNECT branch?

Another thing I wonder, if it is in INIT/RECONNECT mode should we always respond for all initial requests for each type without checking for any conditions?

hzxuzhonghu:
On suspicious point is why the failure occur recently, while not last several years

ramaraochavali:
> On suspicious point is why the failure occur recently, while not last several years

This is the reason as mentioned in PR desc 

BTW - the reason this is suddenly an issue is because we started testing a lower max connection age. These bugs were always present, but they only happened on xds reconnects which never happened in our test environments.

hzxuzhonghu:
IC

ramaraochavali:
1. Envoy connect
2. Envoy REQ EDS, nonce=e1

For this - Do we have any clusters in bootstrap that use EDS?

hzxuzhonghu:
```
1. Envoy connect
2. Envoy REQ EDS, nonce=e1
3. Istiod pushes EDS, nonce=e2
4. Envoy REQ CDS
5. Istio pushes CDS
6. Envoy REQ EDS, nonce=e2
7. Istiod treats this as an ACK (from push (3))
```

Isn't there a real ACK after push(3)?

howardjohn:
> 1. Envoy connect
> 2. Envoy REQ EDS, nonce=e1
> 
> For this - Do we have any clusters in bootstrap that use EDS or is this Reconnect (seems reconnect but just confirming)?

Reconnect

> Isn't there a real ACK after push(3)?

Its either

1. Not a real ACK
2. A real ACK, but envoy requires a EDS push and never requested it

I believe its (1), otherwise we would see issues if the EDS ACK was faster than CDS req

> Another thing I wonder, if it is in INIT/RECONNECT mode should we always respond for all initial requests for each type without checking for any conditions?

We already do, the issue is we need to respond to the _second_ EDS request here. We do respond to the first one.

> This is very useful information. What is the reason for this change? BTW, I am not saying it is bad - it is actually good that we are catching these corner cases, but curious what it is.

We found a few other bugs like this from a test that happened to start taking >30min (thus triggering reconnect), so we intentionally set it lower to catch these bugs

hzxuzhonghu:
> I believe its (1), otherwise we would see issues if the EDS ACK was faster than CDS req

In ads, I believe envoy does send them one by one. ACK for previous push->request for another

costinm:
Can we escalate and put some pressure on Envoy to add an 'ack' or 'type'
field to DiscoveryRequest ? It seems very easy and safe, backward
compatible too.

It is insane that after 5 years we still have bugs on the most basic
protocol function - identifying the message type.

ramaraochavali:
@howardjohn I am confused . Are n't these two conditions same

```
		// We got a request which looks like an ACK, but we already got the same ACK. Envoy won't ACK the
		// same resource twice. However, in some cases a request for new resources is indistinguishable
		// from an ACK (in particular, when requesting EDS after a CDS push):
		// https://github.com/envoyproxy/envoy/issues/13009. As a result, if we see the same ACK multiple
		// times, we treat subsequent requests as something we should respond to.
		if features.PushOnRepeatNonce && request.ResponseNonce == previousNonceAcked {
			log.Debugf("ADS:%s: REQ %s Repeated nonce received %s", stype, con.conID, request.ResponseNonce)
			return true, emptyResourceDelta
		}
		if alwaysRespond {
			log.Infof("ADS:%s: REQ %s forced response", stype, con.conID)
			return true, emptyResourceDelta
		}
```
Should n't the first if returns true?

howardjohn:
features.PushOnRepeatNonce


is redundant now, I'm planning to remove it. The reason is the new fix
applies for all cases (I hope!) while repeat nonce is only a subset of
cases where the previous nonce has completed handling

costinm:
While I don't doubt the solution is an improvement - I am still very
uncomfortable with the complexity and risks.

If we can convince Envoy to add a bit to indicate the type of the message (
ACK, REQUEST, RECONNECT) - we could clean up
a lot of code. The risk seems pretty big - envoy may be stuck and this can
cause an outage.

SpecialYang:
I'm curious that why envoy doesn't initialize xds starting from CDS when envoy reconnects to istiod.

CC @howardjohn

PR #39942, opened 2022-07-15
fix the bug for rds sync failure due to duplicate entry of domain in some routes

zhanghw0354:
fix the bug for rds sync failure due to duplicate entry of domain in some routes

Signed-off-by: zhanghw0354 <zhanghaiwencmss@139.com>

Sometimes RDS may fail，It shows that RDS status is "STALE(Never Acknowledged)" 
image

There are warn messages in istiod logs like "ADS:RDS: ACK ERROR provider-v0-0-2-5b577b8f-jdph2.sm-11 Internal:Only unique values for domains are permitted. Duplicate entry of domain prometheus-node-exporter.kube-system.svc.cluster.local in route 9100.ipv6"
image

It shows that there is duplicate entry of domain prometheus-node-exporter.kube-system.svc.cluster.local in route 9100.ipv6 in the logs due to two similar virtualhosts with duplicate domain.

This PR is to fix this bug.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/route/route.go
+++ b/pilot/pkg/networking/core/v1alpha3/route/route.go
@@ -152,7 +152,7 @@ func BuildSidecarVirtualHostWrapper(routeCache *Cache, node *model.Proxy, push *
 	}
 
 	// append default hosts for the service missing virtual Services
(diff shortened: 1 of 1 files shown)


zhlsunshine:
LGTM

PR #39951, opened 2022-07-16
added a fix to resolve cross cluster traffic

aryan16:
**Cross cluster communication in headless services won't work without DNS Proxying, adding a check while running the tests**


hzxuzhonghu:
@aryan16 Can you elaborate what this pr is fixing?  Seems skip some tests, curious why without this, there is no failure

aryan16:
> @aryan16 Can you elaborate what this pr is fixing? Seems skip some tests, curious why without this, there is no failure

This PR doesn't include any fix for the existing tests, we require DNS proxying for headless services in the case of cross cluster traffic. We have an existing check here (https://github.com/istio/istio/blob/master/pkg/test/framework/components/echo/check/checkers.go#L413), I am using the same one in other methods to keep everything in sync.

aryan16:
closing this one for this - https://github.com/istio/istio/pull/40023

Issue #39950, opened 2022-07-15
Sidecar is slow to startup when GCP is not accessible


rinormaloku:
The below method is querying metadata from GCP. However, when GCP is not accessible (e.g. firewall rules), this will delay the agent to start up for about 90 seconds. 

https://github.com/istio/istio/blob/6450eccc71f01448d2ff740d1011c2358dbe0a28/pkg/bootstrap/platform/gcp.go#L137

We discovered this when `holdApplicationUntilProxyStarts` is set to  `true`. It caused the pod to be in a infinite crash loop, as the pilot-wait health checks were constantly failed (the health check timeouts in 60 seconds).

How to reproduce:

1. In GKE (with Kubernetes Network Policies enbaled) install istio
2. Create test ns and network policy to block traffic to GCP:
```
kubectl create ns test                                      
kubectl label namespace test istio-injection=enabled 

kubectl -n test apply -f- <<EOF
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: test-network-policy
  namespace: test
spec:
  podSelector:
    matchLabels: 
      app: demo
  policyTypes:
    - Egress
  egress:
    - to:
        - ipBlock:
            cidr: 0.0.0.0/0
            except: 
            - 169.254.169.254/32
        - ipBlock:
            cidr: 10.0.0.0/8
EOF
```

3. Create the following workload
```
kubectl -n test apply -f- <<EOF
apiVersion: apps/v1
kind: Deployment
metadata:
  name:  netshoot
  labels:
    app: demo
spec:
  selector:
    matchLabels:
      app: demo
  replicas: 1
  template:
    metadata:
      annotations:
        sidecar.istio.io/agentLogLevel: "all:debug"
      labels:
        app: demo
    spec:
      containers:
      - name:  netshoot
        image:  nicolaka/netshoot
        imagePullPolicy: IfNotPresent
        args: ['tail', '-f', '/dev/null']
        resources:
          requests:
            cpu: 100m
            memory: 100Mi
          limits:
            cpu: 100m
            memory: 100Mi
EOF
```

The sidecar will take about 2 minutes to start up. If you would annotate it to hold until app starts it would fail. 
To do so add `proxy.istio.io/config: '{ "holdApplicationUntilProxyStarts": true }'` to the above annotations.

[...]

PR #39953, opened 2022-07-16
add timeout for IsGCP

zirain:

fix: https://github.com/istio/istio/issues/39950

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/bootstrap/platform/gcp.go
+++ b/pkg/bootstrap/platform/gcp.go
@@ -126,21 +126,20 @@ func IsGCP() bool {
 
 	timeC := time.After(time.Millisecond * 100)
 	ch := make(chan bool)
-	defer close(ch)
 	go func() {
 		ch <- metadata.OnGCE()
 	}()
 
-	for {
-		select {
-		case res, ok := <-ch:
-			if ok {
-				return res
-			}
-		case <-timeC:
-			return false
+	select {
+	case res, ok := <-ch:
+		if ok {
+			return res
 		}
+	case <-timeC:
+		return false
 	}
+
+	return false
 }
 
 // NewGCP returns a platform environment customized for Google Cloud Platform.
--- a/pkg/bootstrap/platform/gcp.go
+++ b/pkg/bootstrap/platform/gcp.go
@@ -22,6 +22,7 @@ import (
 	"regexp"
 	"strings"
 	"sync"
+	"time"
 
 	"cloud.google.com/go/compute/metadata"
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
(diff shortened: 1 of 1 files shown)


pkg/bootstrap/platform/gcp.go
```diff
@@ -122,7 +123,24 @@ func IsGCP() bool {
 		// Assume this is running on GCP if GCP project env variable is set.
 		return true
 	}
-	return metadata.OnGCE()
+
+	timeC := time.After(time.Millisecond * 100)
+	ch := make(chan bool)
+	defer close(ch)
+	go func() {
+		ch <- metadata.OnGCE()
```
hzxuzhonghu, line 131: this is very dangerous, sending to a closed channel can cause panic?

hzxuzhonghu: The best fix should be in google's lib.

Here you can not close the `ch`

pkg/bootstrap/platform/gcp.go
```diff
@@ -122,7 +123,24 @@ func IsGCP() bool {
...
+	}()
+
+	for {
```
hzxuzhonghu, line 134: rm

howardjohn (commented):
https://github.com/istio/istio/blob/e2ce47e2cb85306bcf57160ae3c4a9026cb28018/pkg/bootstrap/platform/discovery.go#L64 is already handling timeouts. Maybe we are in-appropriately calling this outside of the function?

pkg/bootstrap/platform/gcp.go
```diff
@@ -122,7 +123,23 @@ func IsGCP() bool {
 		// Assume this is running on GCP if GCP project env variable is set.
 		return true
 	}
-	return metadata.OnGCE()
+
+	timeC := time.After(time.Millisecond * 100)
+	ch := make(chan bool)
+	go func() {
+		ch <- metadata.OnGCE()
```
hzxuzhonghu, line 130: This can be blocked when real timeout happen

Random-Liu: Just FYI. In k8s, it just check the local file `/sys/class/dmi/id/product_name` to decide whether it is on GCE or not
https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/gcp/metadata.go#L137

zirain: ok, I will revisit that, thanks.

howardjohn:
that may not be accessible from a pod. not sure

PR #39959, opened 2022-07-18
Fix resiliience port-forward

hzxuzhonghu:

Without this, when err happens it will be stuck sending to errCh, so cause the goroutine hanging

When ` fw.ForwardPorts()` returns without err, we should actually retry

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/kube/portforwarder.go
+++ b/pkg/kube/portforwarder.go
@@ -70,11 +70,15 @@ func (f *forwarder) Start() error {
 			fw, err := f.buildK8sPortForwarder(readyCh)
 			if err != nil {
 				errCh <- err
-				break
+				return
 			}
 			if err = fw.ForwardPorts(); err != nil {
-				errCh <- err
-				break
+				select {
+				case errCh <- err:
+					return
+				default:
+					break
+				}
 			}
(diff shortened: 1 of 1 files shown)


GregHanson (approved):

pkg/kube/portforwarder.go
```diff
@@ -70,11 +70,11 @@ func (f *forwarder) Start() error {
 			fw, err := f.buildK8sPortForwarder(readyCh)
 			if err != nil {
 				errCh <- err
-				break
+				return
```
howardjohn, line 73: I don't think I understand. It seems the difference between break and return here is to skip `readyCh = nil`.

Which seems fine, since the select should handle the errCh anyways. But wasn't sure what you meant by the goroutine hanging?

hzxuzhonghu: Think about when `buildK8sPortForwarder` always fail

howardjohn: then `break` will exit the for loop and the go routine will terminate? Just like return? Maybe I am missing something

PR #39960, opened 2022-07-18
handle non wildcard resources during reconnect

ramaraochavali:
Attempt for a generic fix for https://github.com/istio/istio/pull/39937
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -313,9 +313,6 @@ type Proxy struct {
 	// WatchedResources contains the list of watched resources for the proxy, keyed by the DiscoveryRequest TypeUrl.
 	WatchedResources map[string]*WatchedResource
 
-	// nolint
-	LastSentTypeUrl string
-
 	// XdsNode is the xDS node identifier
 	XdsNode *core.Node
 
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -313,6 +313,9 @@ type Proxy struct {
 	// WatchedResources contains the list of watched resources for the proxy, keyed by the DiscoveryRequest TypeUrl.
 	WatchedResources map[string]*WatchedResource
 
+	// nolint
+	LastSentTypeUrl string
+
 	// XdsNode is the xDS node identifier
 	XdsNode *core.Node
 
@@ -356,8 +359,10 @@ type WatchedResource struct {
 	NonceNacked string
 
 	// AlwaysRespond, if true, will ensure that even when a request would otherwise be treated as an
-	// ACK, it will be responded to. Typically, this should be set to 'false' after response; keeping it
-	// true would likely result in an endless loop.
+	// ACK, it will be responded to. This typically happens when a proxy reconnects to another instance of
+	// Istiod. In that case, Envoy expects us to respond to EDS/RDS/SDS requests to finish warming of
+	// clusters/listeners.
+	// Typically, this should be set to 'false' after response; keeping it true would likely result in an endless loop.
 	AlwaysRespond bool
 
 	// LastSent tracks the time of the generated push, to determine the time it takes the client to ack.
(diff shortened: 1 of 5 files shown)


ramaraochavali:
@howardjohn PTAL if this makes sense.

pilot/pkg/model/context.go
```diff
@@ -356,8 +359,10 @@ type WatchedResource struct {
 	NonceNacked string
 
 	// AlwaysRespond, if true, will ensure that even when a request would otherwise be treated as an
-	// ACK, it will be responded to. Typically, this should be set to 'false' after response; keeping it
-	// true would likely result in an endless loop.
+	// ACK, it will be responded to. This typically happens when a proxy reconnects to another instance of
+	// Istiod. In that case, Envoy expects us to respond to EDS/RDS/SDS requests to finish warming of
```
howardjohn, line 363: This exclusive applies to CDS+EDS as far as I know, not all dependent types

ramaraochavali: > Can you give an example Istiod ads:debug log of this happening?

I have n't seen this happen in Istiod. But based on the above it should apply to any named resource. Let me try to setup a test case for it.

ramaraochavali:
That would be violation of https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#resource-warming in spec.

ramaraochavali:
Ok. Let me take a look at EDS issue. In any case, I think we sending RDS also is safe - it just costs one extra response. But I will try to look at the code and try to create a test case tomorrow.

hzxuzhonghu:
Is the issue is figured out clearly? I am still not sure

ramaraochavali:
> Is the issue is figured out clearly? I am still not sure

I am 100% sure that when cluster is warming up during reconnect, we need to send named response for EDS to finish warming. That is what I fixed in Envoy https://github.com/envoyproxy/envoy/pull/6151 long back. It was clearing endpoints without that.  At that time I remember I validated the same thing happens for LDS/RDS - that is why this section https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#resource-warming is added in Xds docs. I am going to verify if that is the case still for LDS/RDS

hzxuzhonghu:
If this is the cause, then with the recent fixes there should be no ci failure. however, it is still flaky. So that's why I doubt it

howardjohn:
The ci failures are for 3 reasons:
1. delta XDS isn't fixed. I need to see debug logs of one of those failures
to see exact what we need to change. Whould have a fix tomorrow
2. simple CPU exhaustion leads to timeouts. Need investigation on what to
do here.
3. known iptables port conflict bug

this is after analyzing 20 failures across 2000 test executions over the
weekend. There were zero issues caused by reconnect warming

ramaraochavali:
I have validated this behaviour for LDS/RDS with tests in Envoy.
- If Listener is updated with the other details keeping the route config name same, it does not get in to warming for RDS. It reuses the route definition with that name. https://github.com/envoyproxy/envoy/pull/9209 fixed that part.
- If Listener is updated with changing route config name, it gets in to warming and excepts RDS to be sent for new route - which out code handles.
- For secrets also it does not get in to warming.

hzxuzhonghu:
Any reason why they behave different from cluster?

pilot/pkg/xds/ads.go
```diff
@@ -435,27 +441,21 @@ func (s *DiscoveryServer) shouldRespond(con *Connection, request *discovery.Disc
 	previousInfo.AlwaysRespond = false
 	con.proxy.Unlock()
 
-	// Envoy can send two DiscoveryRequests with same version and nonce
+	// We should always respond "alwayRespond" marked requests to let Envoy finish warming
+	// even though Nonce match and it looks like an ACK.
```
howardjohn, line 445: > and it looks like an ACK.

Is not quite right -- it could be legitimate resource change? Which may give confusing logs?

Not likely to happen much/ever in practice though. But it may be more accurate if we only log this inside `len(removed) == 0 && len(added) == 0`

howardjohn (approved):

PR #39961, opened 2022-07-18
Fix data races in sidecar scope processing

johnma14:
We hit an issue in our test clusters where we have defined a SidecarScope for the root (istio-system) namespace and we expect all the user namespaces created to inherit this sidecarscope and that was not happening. Instead of getting the 'default' sidecarscope definition of the istio-system ns, it was getting the 'default-sidecar' sidecarscope. In big clusters, this leads to OOM.


howardjohn (commented):
Its not clear to me what this is fixing or how. Can you elaborate or add a test?

hzxuzhonghu:
Better provide the panic stack

johnma14:
I will go ahead and add a test case for this. I will see if I can add a stack trace as well. Thank you.

johnma14:
Fixed by https://github.com/istio/istio/pull/40575. Locks used in the sidecarIndex processing was refactored in: https://github.com/istio/istio/pull/40688. Unit tests for #40575 added here: https://github.com/istio/istio/pull/41038

PR #39979, opened 2022-07-18
Simpler injectionPath format for cluster env value with / char

### LINKED ISSUE #40116: Injection webhook validation failure when using :ENV path format
### Bug Description

Using the `:ENV` format for configuring an `injectionPath` for the injection webhook fails to validate.

For example, installing this IOP:

```
kind: IstioOperator
metadata:
  namespace: external-istiod
spec:
  profile: external
  values:
    global:
      istioNamespace: external-istiod
      remotePilotAddress: 169.55.170.85
    istiodRemote:
      injectionPath: /inject/:ENV:cluster=cluster1/c1o73jbw0d1g0adp0k90:ENV:net=network
```

produces the following error message:

```
The MutatingWebhookConfiguration "istio-sidecar-injector-external-istiod" is invalid: 
* webhooks[0].clientConfig.service.path: Invalid value: "/inject/:ENV:cluster=cluster1/c1o73jbw0d1g0adp0k90:ENV:net=network1": segment[1]: a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

Note that configuring the same path using an `injectionURL` (instead of `injectionPath`) does not fail validation.

```
kind: IstioOperator
metadata:
  namespace: external-istiod
spec:
  profile: external
  values:
    global:
      istioNamespace: external-istiod
    istiodRemote:
      injectionURL: https://169.55.170.85:15017/inject/:ENV:cluster=cluster1/c1o73jbw0d1g0adp0k90:ENV:net=network1
```

### Version

```prose
latest
```

### Additional Information

_No response_

frankbu:

Provide alternative and deprecate the `:ENV:` path syntax since it is not recommended and doesn't work when setting the path (vs URL) in the webhook.

Fixes #40116

### DIFF
--- a/pkg/kube/inject/webhook.go
+++ b/pkg/kube/inject/webhook.go
@@ -1000,9 +1000,9 @@ func parseInjectEnvs(path string) map[string]string {
 		var newRes []string
 		if len(parts) == 3 {
 			if strings.HasPrefix(parts[2], ":ENV:") {
-				// TODO Deprecate this, because it is not recommended. It also fails validation when used to set
-				//      injectionPath (i.e., service.path in mwh). It doesn't fail validation when used to set
-				//      injectionURL, however. K8s bug maybe?
+				// Deprecated, not recommended.
+				//    Note that this systax fails validation when used to set injectionPath (i.e., service.path in mwh).
+				//    It doesn't fail validation when used to set injectionURL, however. K8s bug maybe?
 				pairs := strings.Split(parts[2], ":ENV:")
 				for i := 1; i < len(pairs); i++ { // skip the first part, it is a nil
 					pair := strings.SplitN(pairs[i], "=", 2)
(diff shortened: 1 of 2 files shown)


frankbu:
@howardjohn 

> why do we need yet another form?

I want to deprecate/remove the other form. It's klunky, not recommended, and doesn't work in injectionPath (see comment in the changed code).

> If we do want to escape / we should probably use the standard %2F

That would be preferable, but unfortunately the path is url unencoded by the time we get to process it.

howardjohn (approved):

Issue #39800, opened 2022-07-06
ServiceEntry-created clusters are not matched in IPv6


LithiumSR:
### Bug Description

Hello everyone,
I think I found an issue with Istio 1.14.1 that is affecting connectivity to external services.
When SE are used in IPv6 to connect to external services, it seems that the requests are going into the BlackHoleCluster as they are not able to match the expected cluster.

The steps to reproduce are as follow:

0. Setup IPv6 KinD cluster or use another IPv6 cluster
1. Install Istio 1.14.1 with `istioctl install --set profile=demo --set meshConfig.outboundTrafficPolicy.mode=REGISTRY_ONLY`
2. Deploy secrets for nginx via  `kubectl create secret tls nginx-server-certs --cert=nginx.example.com.crt --key=nginx.example.com.key -n istio-system`
3. Deploy nginx configmap via `kubectl create configmap nginx-configmap --from-file=nginx.conf=nginx_http_https.conf -n istio-system`
4. Label istio-system with istio-injection=enabled using `kubectl label ns istio-system istio-injection=enabled`
5. Deploy Sleep with sidecar `kubectl apply -f https://raw.githubusercontent.com/istio/istio/master/samples/sleep/sleep.yaml -n istio-system`
6. Deploy nginx without sidecar `kubectl apply -f nginx_http_https_no_sidecar.yaml -n istio-system`
7. Get the podIP from `kubectl get pod $NGINX_POD -n istio-system --output=yaml`
8. Replace "NGINX_IP" in se.yaml with the pod ip you retrieved earlier and apply it `kubectl apply -f se.yaml -n istio-system`.
9. From the sleep container execute `curl https://nginx.example.com --max-time 5 --resolve nginx.example.com:443:[YOUR_NGINX_IP] -k`
It will fail with SSL_ERROR_SYSCALL. 
With the help of the sidecar lgos and you can verify that it did not match the nginx cluster and therefore it went to the blackhole.
```[2022-07-06T09:00:57.041Z] "- - -" 0 UH - - "-" 0 0 13 - "-" "-" "-" "-" "-" BlackHoleCluster - [fd00:10:244::8]:443 [fd00:10:244::f]:51048 - -```
This steps does not fail on IPv4.
10. Try the same curl request from a sleep container without sidecar and everything will work as expected.

I'm attaching the resources needed to execute the aforementioned steps [resources.zip](https://github.com/istio/istio/files/9053374/resources.zip)

### Version

```prose
Istioctl:
[...]

PR #39986, opened 2022-07-19
Fixed bugs for ServiceEntry generated clusters are not matched in IPv6 only cluster

zhlsunshine:
This PR is for [issue#39800](https://github.com/istio/istio/issues/39800). After some investigation, I found the root cause is incorrect listener configuration for service entry. 
The wrong configuration has been highlighted in below snapshot:
image

The address should not be `0.0.0.0` for the service entry in IPv6 only cluster.

Below is the new snapshot after fixing:

And the curl is normal and get the response from nginx:


pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -688,6 +688,13 @@ func buildSidecarOutboundTCPListenerOptsForPortOrUDS(listenerMapKey *string,
 	var destinationCIDR string
 	if len(listenerOpts.bind) == 0 {
 		svcListenAddress := listenerOpts.service.GetAddressForProxy(listenerOpts.proxy)
+		// if listenerOpts.service is not the common k8s svc and headless svc, it may be a svc converted from service entry, etc.
+		// It's hard to detect IP family for this kind of service, such se with only host name, so follows the current proxy's
+		if listenerOpts.service.ClusterVIPs.IsEmpty() && svcListenAddress == constants.UnspecifiedIP {
```
ramaraochavali, line 693: why not detect from ServiceRegistry?

howardjohn (changes requested):
There are no tests changed. This means we are missing coverage. Can we add them?

zhlsunshine:
> There are no tests changed. This means we are missing coverage. Can we add them?

sure, I will

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: Hi @ramaraochavali ,  user may define ServiceRegistry with only hostname and without address, so sometimes we can not detect it from ServiceRegistry side.

zhlsunshine: Thanks @ramaraochavali ! changed it

zhlsunshine:
Hi @howardjohn ,  I simulated the services of SE without addresses and IPv6 proxy to generate the outbound listeners configuration for unit test.

howardjohn (commented):
Shouldn't we have some e2e tests on the ipv6 test that fails without this ?

zhlsunshine:
> Shouldn't we have some e2e tests on the ipv6 test that fails without this ?

Hi @howardjohn, according to the [issue#39800](https://github.com/istio/istio/issues/39800), it's unavailable when using the endpoint address of nginx (pod IP address), however it's okay when using nginx service cluster IP address. In fact this is not entirely a IPv6 issue, it seems like service entry usage. So I think the unit test should be enough. Is that okay?

zhlsunshine:
> Shouldn't we have some e2e tests on the ipv6 test that fails without this ?

Hi @howardjohn, after some investigation for integration test in IPv6, I found out that it's hard to fully support the e2e test due to many manual operations need to be taken to reproduce this scenario based on [issue#39800](https://github.com/istio/istio/issues/39800). So could it be verified via the current unit tests I added and have already covered both IPv4 and IPv6 sidecar for this kind of service entry?

pilot/pkg/networking/core/v1alpha3/listener_test.go
```diff
@@ -564,6 +572,71 @@ func TestOutboundListenerForHeadlessServices(t *testing.T) {
...
+			cg := NewConfigGenTest(t, TestOptions{
+				Services:  tt.services,
+				Instances: tt.instances,
+			})
+
+			proxy := cg.SetupProxy(getIPv6Proxy())
+
+			listeners := NewListenerBuilder(proxy, cg.env.PushContext).buildSidecarOutboundListeners(proxy, cg.env.PushContext)
+			for _, l := range listeners {
+				if l.Address.GetSocketAddress().GetPortValue() == 9999 {
```
jacob-delgado, line 630: Is there a way to also check `GetSocketValue` or similar?

jacob-delgado:
👍🏽 LGTM. Good catch. The change in Envoy was done to better support dual-stack listeners and matching them to incoming requests.

I believe setting it to "::/0" is appropriate as well. However, I am concerned that having `net.ipv6.bindv6only = 1` could cause problems. I don't know how many people set that in IPv6 only clustesr.

pilot/pkg/networking/core/v1alpha3/listener_test.go
zhlsunshine: Okay, will check it

zhlsunshine:
Hi @howardjohn and @jacob-delgado, as the [issue#11184](https://github.com/envoyproxy/envoy/issues/11184) of envoy has been fixed, so we can enable the multiple addresses per listener for Istio to fix this bug, is that okay?

zhlsunshine:
test integ-basic-arm64_istio 
test unit-tests-arm64_istio

zhlsunshine:
test /unit-tests-arm64_istio

jacob-delgado (approved):

jacob-delgado:
Once we agree on a way to fix this, we should try and get this cherry-picked into 1.14 as well.

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -688,6 +688,13 @@ func buildSidecarOutboundTCPListenerOptsForPortOrUDS(listenerMapKey *string,
 	var destinationCIDR string
 	if len(listenerOpts.bind) == 0 {
 		svcListenAddress := listenerOpts.service.GetAddressForProxy(listenerOpts.proxy)
+		// if listenerOpts.service is not the common k8s svc and headless svc, it may be a svc converted from service entry, etc.
+		// It's hard to detect IP family for this kind of service, such se with only host name, so follows the current proxy's
```
ramaraochavali, line 692: // It's hard to detect IP family for this kind of service, such se with only host name, so follows the current proxy's

This is unclear to me

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -688,6 +688,13 @@ func buildSidecarOutboundTCPListenerOptsForPortOrUDS(listenerMapKey *string,
...
+		if listenerOpts.service.Attributes.ServiceRegistry == provider.External {
+			if listenerOpts.proxy.IsIPv6() {
```
ramaraochavali, line 694: nit: combine this with above if

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
```diff
@@ -271,11 +271,16 @@ func (lb *ListenerBuilder) buildInboundChainConfigs() []inboundChainConfig {
...
+			bindAddr := WildcardAddress
+			if lb.node.IsIPv6() {
+				bindAddr = WildcardIPv6Address
+			}
+
 			cc := inboundChainConfig{
 				telemetryMetadata: telemetry.FilterChainMetadata{InstanceHostname: i.Service.Hostname},
 				port:              port,
 				clusterName:       model.BuildInboundSubsetKey(int(port.TargetPort)),
-				bind:              "0.0.0.0", // TODO ipv6
```
ramaraochavali, line 278: So this is being changed for every service not just service entry based clusters?

ramaraochavali (commented):
LGTM. With minor comments. Will let @howardjohn comment on integration test and approve.

howardjohn:
For testing: we have ServiceEntry testing already via https://github.com/istio/istio/blob/b524d9836880a88c7aaa91956dfca24c82c29039/pkg/test/framework/components/echo/common/deployment/external.go#L33. We can/should improve this for IPv6 as needed.

We already run an IPv6 cluster in presubmit, and have logic to skip based on whether IPv4 vs IPv6: https://github.com/istio/istio/blob/b524d9836880a88c7aaa91956dfca24c82c29039/tests/integration/pilot/common/routing.go#L2834

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: Hi @ramaraochavali,  it's okay for normal k8s service in the service mesh, but if there is a service entry out of service mesh, even out of k8s cluster, then we can not really know its service ip family is IPv4, IPv6 or both. So that's what I want to explain.

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
zhlsunshine: Yes, @ramaraochavali, it's being changed for every service. However, `lb.node.IsIPv6()` is different with `node.SupportIPv6`, the former checks that the node is IPv6 only, and the latter checks that the node can be both IPv6 only or Dual. 
So this change only impacts the IPv6 only cluster.

zhlsunshine:
Yeah so @howardjohn & @ramaraochavali, could you please help to approve this PR if possible? Thanks!

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
ramaraochavali: got it

pilot/pkg/networking/core/v1alpha3/listener.go
ramaraochavali: Suggest to change
"It is detect IP family for service entries, so use the current proxy's IP Mode."

zhlsunshine: Okay, thanks @ramaraochavali

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: sure, got it thanks! @ramaraochavali

zhlsunshine: done

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: done

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -693,6 +693,11 @@ func buildSidecarOutboundTCPListenerOptsForPortOrUDS(listenerMapKey *string,
 	var destinationCIDR string
 	if len(listenerOpts.bind) == 0 {
 		svcListenAddress := listenerOpts.service.GetAddressForProxy(listenerOpts.proxy)
+		// if listenerOpts.service is not the common k8s svc and headless svc, it may be a svc converted from service entry, etc.
+		// It is hard to detect IP family for service entries, so use the current proxy's IP Mode.
+		if listenerOpts.service.Attributes.ServiceRegistry == provider.External && listenerOpts.proxy.IsIPv6() {
```
howardjohn, line 698: shouldn't this only be when it is originally `constants.Unspecified` we switch it to `constants.UnspecifiedIPv6`? Seems like this would wipe out any VIPs

howardjohn (commented):
should we expand the integ tests to cover the improvements

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: Hi @howardjohn, `listenerOpts.service.Attributes.ServiceRegistry == provider.External` means that it's not the normal Istio service in service mesh, it should be service entry. So there is no any ClusterVIP(I understand you mentioned `VIP` means `ClusterVIPs`) would be impacted. And we should switch `constants.Unspecified` into `constants.UnspecifiedIPv6` if the listenersOpts's proxy only support IPv6 and there is no any other address bind.  Hope I explains it clearly.

zhlsunshine:
> should we expand the integ tests to cover the improvements

Yes, about the integ tests to cover the improvements, and you mentioned the [external test](https://github.com/istio/istio/pull/39986#issuecomment-1218236681), I think it's okay and I think @LithiumSR would be clear to help this integ tests, and I will discuss with him later.

pilot/pkg/networking/core/v1alpha3/listener.go
howardjohn: ServiceEntry can have a VIP from the `addresses` field

zhlsunshine: Yeah, so you worry on how about if there are IPv4 and IPv6 in service entry's addresses? 
Based on current implementation, `constants.Unspecified` is the only choice, no matter it is correct or not. However, it's okay if we follow the `listenerOpts.proxy` of the service, the available IP family may be elected at least.

zhlsunshine: Yeah @howardjohn, so you worry on how about if there are IPv4 and IPv6 in service entry's addresses? 
Based on current implementation, `constants.Unspecified` is the only choice, no matter it is correct or not. However, it's okay if we follow the `listenerOpts.proxy` of the service, the available IP family may be elected at least.
Moreover, `constants.UnspecifiedIPv6` can be available for both IPv4 and IPv6 by default.

kfaseela:
> > should we expand the integ tests to cover the improvements
> 
> Yes, about the integ tests to cover the improvements, and you mentioned the [external test](https://github.com/istio/istio/pull/39986#issuecomment-1218236681), I think it's okay and I think @LithiumSR would be clear to help this integ tests, and I will discuss with him later.

@howardjohn @zhlsunshine I can take a look at the integration test, would need some more time though :)

pilot/pkg/networking/core/v1alpha3/listener.go
howardjohn: I don't understand how `constants.Unspecified` can be the only choice..? I don't think that is accurate. A service entry is a superset of the `Service` API (more or less), it can have VIPs

Issue #39983, opened 2022-07-18
CNI logging issues


howardjohn:
CNI logging works by sending logs over a UDS socket from the istio-cni binary up to the user-space install-cni container. There are a few issues here around logging levels and scopes.

* https://github.com/istio/istio/blob/764e6688921e326bdaa530d5ef3a9ed5d87372f7/cni/pkg/plugin/plugin.go#L177 sets the scope as configured by `logLevel` in the configmap, which is `Values.cni.logLevel`. This only handles default scope. if we have other scopes, we cannot configure them. Either we should apply this to all scopes, or allow the standard parsing of scopes that log_output_level does
* Daemonset doesn't allow setting the level at all: https://github.com/istio/istio/blob/ff3ccd19943796e58e9ac16db114a99dcaa8b4d6/manifests/charts/istio-cni/templates/daemonset.yaml#L76
* Maybe not a bug, but something we should consider: https://github.com/istio/istio/blob/bf2adb0be5c057924a79d1a78a39ffb87aeb1175/cni/pkg/log/uds.go#L121 loses the original `scope` of the log. If we send scoped logs, we lose that info. Either we can lookup the original scope, or we should add a key to do the equivalent

zirain:
(1) and (3) should be consider together.
For now, all logs from istio-cni binary' scope is `default`, there's why we see `log.FindScope("default").SetOutputLevel(getLogLevel(conf.LogLevel)) `; uds-server will output them in `cni` scope. 
```console
2022-07-19T06:40:16.980423Z	info	cni	istio-cni cmdAdd with k8s args: {CommonArgs:{IgnoreUnknown:true} IP:<nil> K8S_POD_NAME:sleep-5887ccbb67-kjds5 K8S_POD_NAMESPACE:default K8S_POD_INFRA_CONTAINER_ID:0903460cebefc2d22d2d04b152031b4021450b4be4988f08d04d4fd2791198c0}
2022-07-19T06:40:16.980444Z	info	cni	============= Start iptables configuration for sleep-5887ccbb67-kjds5 =============
2022-07-19T06:40:16.980448Z	info	cni	Istio iptables environment:
```
so I prefer to keep the status quo unchanged.

PR #39988, opened 2022-07-19
improve cni logging

zirain:

improve cni logging

part of https://github.com/istio/istio/issues/39983

### DIFF
--- a/manifests/charts/istio-cni/values.yaml
+++ b/manifests/charts/istio-cni/values.yaml
@@ -4,7 +4,9 @@ cni:
   image: install-cni
   pullPolicy: ""
 
-  logLevel: info
+  # Configuration log level of istio-cni binary
+  # by default istio-cni send all logs to UDS server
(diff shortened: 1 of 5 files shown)


manifests/charts/istio-cni/values.yaml
```diff
@@ -4,7 +4,9 @@ cni:
   image: install-cni
   pullPolicy: ""
 
-  logLevel: info
+  # Configuration log level of istio-cni binary
+  # by default istio-cni send all logs to UDS server
+  logLevel: debug
```
howardjohn, line 9: so we send debug logs, but then to see them you need `logging.level=cni:debug`, right?

Can we explain that in the comments maybe?

zirain: updated.

howardjohn (approved):

jacob-delgado (approved):

PR #40007, opened 2022-07-20
gateway-api: support invalid BackendRefs

### LINKED ISSUE #39919: Gateway conformance: support HTTPRouteInvalidNonExistentBackendRef
Currently HTTPRouteInvalidNonExistentBackendRef is skipped. This is because we need to implement support for detecting it _and_ we need https://github.com/envoyproxy/envoy/issues/21930 merged

dddddai:
Fixes #39919


pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -262,27 +262,14 @@ func convertVirtualService(r ConfigContext) []config.Config {
...
-	for _, r := range route.Rules {
-		if len(r.Matches) > 1 {
-			// split the rule to make sure each rule has up to one match
-			matches := r.Matches
-			for _, m := range matches {
-				r.Matches = []k8s.HTTPRouteMatch{m}
-				splitRules = append(splitRules, r)
-			}
-		}
-	}
```
dddddai, line 283: This change is for https://github.com/istio/istio/pull/39085#discussion_r906141636

pilot/pkg/config/kube/gateway/testdata/invalid.yaml
```diff
@@ -123,8 +123,15 @@ spec:
   hostnames: ["second.domain.example"]
   rules:
   - backendRefs:
+    - name: nonexistent
```
howardjohn, line 126: can we still have a test where ALL backends are invalid?

pilot/pkg/config/kube/gateway/testdata/reference-policy-service.status.yaml.golden
```diff
@@ -64,9 +64,10 @@ status:
       status: "True"
       type: Accepted
     - lastTransitionTime: fake
-      message: All references resolved
-      reason: ResolvedRefs
-      status: "True"
+      message: backendRef httpbin/default not accessible to a route in namespace "istio-system"
```
howardjohn, line 67: can we put positive an negative cases into different routes so we can ensure we have coverage of both? this may mask issues by merging them

pilot/pkg/model/push_context.go
```diff
@@ -2061,6 +2061,10 @@ func NewGatewayContext(ps *PushContext) GatewayContext {
 	return GatewayContext{ps}
 }
 
+func (gc GatewayContext) GetService(hostname string) *Service {
+	return gc.ps.ServiceForHostname(nil, host.Name(hostname))
```
howardjohn, line 2065: can we pass hostname + namespace to this method? As is it may give inaccurate results in some cases.

tests/integration/pilot/gateway_conformance_test.go
```diff
@@ -127,9 +120,6 @@ func TestGatewayConformance(t *testing.T) {
 
 			for _, ct := range tests.ConformanceTests {
 				t.Run(ct.ShortName, func(t *testing.T) {
-					if reason, f := skippedTests[ct.ShortName]; f {
```
howardjohn, line 130: nit: keep this in case we need it for the future

pilot/pkg/networking/core/v1alpha3/sidecar_simulation_test.go
```diff
@@ -2445,7 +2445,47 @@ spec:
 					for _, tc := range tt.cfg {
 						cfg = cfg + "\n---\n" + tc.Config(variant)
 					}
-					s := xds.NewFakeDiscoveryServer(t, xds.FakeOptions{ConfigString: cfg})
+					s := xds.NewFakeDiscoveryServer(t, xds.FakeOptions{
+						ConfigString: cfg,
+						Services: []*model.Service{
+							{
+								Hostname: "alt-known-default.example.com",
```
howardjohn, line 2452: This is pretty annoying to have to do everywhere.

I think in general there are two different cases here -- reference _not allowed_ and reference unknown. I wonder if we can have unknown reference generate the same VirtualService but just report an issue in status? 

For not allowed we should explicitly deny it probably though

pilot/pkg/config/kube/gateway/testdata/invalid.yaml
```diff
@@ -123,8 +123,15 @@ spec:
   hostnames: ["second.domain.example"]
```
howardjohn, line 123: can we have a test that shows the status showing unknown backend?

pilot/pkg/config/kube/gateway/conversion.go
```diff
@@ -852,19 +845,38 @@ func buildHTTPDestination(
...
+				invalidBackendErr = err
+				invalid = true
+			} else {
+				return nil, err
+			}
+		}
+		if invalid {
+			// this backend is invalid, set a nonexistent host so that envoy will return 500 when traffic goes to this cluster
+			dst = &istio.Destination{
+				Host: "zzz-invalid-backend.istio.io",
```
howardjohn, line 877: I don't feel very comfortable with this string and random port. This will get exposed indirectly to user through configuration.

Maybe we can do this:
* Set `dst = istio.Destination{}`
* Make validation allow empty destination IF `internal.istio.io/route-semantics` is set
* In route generation, if we have empty destination replace it with a cluster `UnknownService` (which doesn't exist)

howardjohn (commented):
Awesome! Looks really good, just a few comments

dddddai:
Thank you, fixed

pkg/config/validation/validation.go
```diff
@@ -2786,7 +2787,9 @@ func validateHTTPRouteDestinations(weights []*networking.HTTPRouteDestination) (
 			errs = appendErrors(errs, ValidateHTTPHeaderOperationName(name))
 		}
 
-		errs = appendErrors(errs, validateDestination(weight.Destination))
+		if !gatewaySemantics {
+			errs = appendErrors(errs, validateDestination(weight.Destination))
```
howardjohn, line 2791: nit: we could just allow `""` for gatewaySemantics, since validateDestination has some other checks that are still useful

howardjohn (approved):
Excellent!

PR #40009, opened 2022-07-20
remove max concurrent streams default

ramaraochavali:
We added max concurrent streams default when Envoy's default value was 100. Now it has been increased https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-msg-config-core-v3-http2protocoloptions - so it is better to default it to Envoy rather than our own number.

This PR fixes it.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn:
did it change just recently or also in older versions?

hzxuzhonghu (approved):

PR #40015, opened 2022-07-20
xds: stop import v2 APIs which are no longer supported

howardjohn:
Envoy has dropped support for v2 APIs some time ago. This stops
importing all of them, which drops 7mb off the binary size and improves
compile time, etc.


GregHanson:
do we need a release note for this?  Assuming no since removing the imports doesn't break anything

howardjohn:
I believe its already broken to use the v2 APIs so this shouldn't. Ill double check

howardjohn:
Hmm, somehow we are still converting these to v3 automaticaly. Yuck.

ramaraochavali:
> somehow we are still converting these to v3 automaticaly

Where?

GregHanson:
at least one test:
```
=== RUN   TestValidateEnvoyFilter/deprecated_config
    validation_test.go:4371: got err=Envoy filter: could not resolve Any message type: type.googleapis.com/envoy.config.filter.network.ext_authz.v2.ExtAuthz but wanted err=
```

howardjohn:
it's not just the test, If you apply one to real cluster with v2 types it
works... except with this PR it starts breaking.

howardjohn:
I vaguely recall we had some logic to do this, but tbh I don't remember how
it works. It might just be with types that merge not adding new v2 filters

ramaraochavali:
> I vaguely recall we had some logic to do this, but tbh I don't remember how
it works. It might just be with types that merge not adding new v2 filters

I was thinking about that but I thought we only did for ADS (upgrade and downgrade) and we have deleted later. It would be good to figure out where and clean it up.

PR #40023, opened 2022-07-20
check DNS Proxying for headless svc

aryan16:
**Check the DNS Proxying requirement before making calls to headless svc in cross cluster environment**

### DIFF
--- a/pkg/test/framework/components/echo/check/checkers.go
+++ b/pkg/test/framework/components/echo/check/checkers.go
@@ -431,9 +431,19 @@ func ReachedTargetClusters(t framework.TestContext) echo.Checker {
 // For multi-network configurations, verifies the current (limited) Istio load balancing behavior when going through
 // a gateway. Ensures that all expected networks were reached, and that all clusters on the same network as the
 // client were reached.
-func ReachedClusters(allClusters cluster.Clusters, expectedClusters cluster.Clusters) echo.Checker {
+func ReachedClusters(t framework.TestContext, allClusters cluster.Clusters, expectedClusters cluster.Clusters) echo.Checker {
 	expectedByNetwork := expectedClusters.ByNetwork()
 	return func(result echo.CallResult, err error) error {
+		dnsCaptureEnabled := isDNSCaptureEnabled(t)
+		to := result.Opts.To
+		if !dnsCaptureEnabled && to.Config().IsHeadless() {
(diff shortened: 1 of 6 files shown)


pkg/test/framework/components/echo/check/checkers.go
```diff
@@ -431,9 +431,19 @@ func ReachedTargetClusters(t framework.TestContext) echo.Checker {
 // For multi-network configurations, verifies the current (limited) Istio load balancing behavior when going through
 // a gateway. Ensures that all expected networks were reached, and that all clusters on the same network as the
 // client were reached.
-func ReachedClusters(allClusters cluster.Clusters, expectedClusters cluster.Clusters) echo.Checker {
+func ReachedClusters(allClusters cluster.Clusters, expectedClusters cluster.Clusters, t framework.TestContext) echo.Checker {
```
howardjohn, line 434:
```suggestion
func ReachedClusters(t framework.TestContext, allClusters cluster.Clusters, expectedClusters cluster.Clusters) echo.Checker {
```

Standard in go is to have `t` as first arg

howardjohn (approved):

pkg/test/framework/components/echo/check/checkers.go
```diff
@@ -431,9 +431,19 @@ func ReachedTargetClusters(t framework.TestContext) echo.Checker {
...
 	expectedByNetwork := expectedClusters.ByNetwork()
 	return func(result echo.CallResult, err error) error {
+		dnsCaptureEnabled := isDNSCaptureEnabled(t)
+		to := result.Opts.To
+		if !dnsCaptureEnabled && to.Config().IsHeadless() {
+			// Headless services rely on DNS resolution. If DNS capture is
+			// enabled, DNS will return all endpoints in the mesh, which will
+			// allow requests to go cross-cluster. Otherwise, k8s DNS will
+			// only return the endpoints within the same cluster as the source
```
hzxuzhonghu, line 443: This is not correct, it depends on feature flag `MulticlusterHeadlessEnabled` and network of a proxy and headless svc instance

aryan16: I am not sure about the flag, but if you disable the DNS proxying (https://github.com/istio/istio/blob/master/tests/integration/base.yaml#L13) then all the tests based on cross cluster headless svc will be broken.

PR #40028, opened 2022-07-20
Move go.mod to 1.18

howardjohn:
This enables
https://github.com/istio/istio/issues/36308#issuecomment-1190847116
cc @istio/wg-test-and-release-maintainers

PR #40029, opened 2022-07-20
move more fuzzers over to native fuzzers

howardjohn:
This PR moves all of the InternalFuzz ers over to native fuzzing. We can/should move _all_ fuzzers eventually, but its useful to move these since they are excluded from the build which makes them brittle.

This depends on https://github.com/istio/istio/pull/40028


jacob-delgado (approved):

hzxuzhonghu (commented):
From your desc, i am thinking you are using go 1.18 native fuzz. But from the implement  it seems, it just make use of `github.com/AdaLogics/go-fuzz-headers `. 

how does this related to native fuzzer?

stevenctl (approved):

lei-tang (approved):

ericvn (approved):

howardjohn:
It's still using the native fuzzing, but the native fuzzing doesn't
natively support fuzzing arbitrary structs. Instead we fuzz bytes and
convert to struct.

The important part of the native fuzzing isn't the type that's fuzzed - it
will produce similar results this way - but the integration into standard
go tooling and compiler

PR #40030, opened 2022-07-20
Refactor to remove InputParams struct

howardjohn:
This struct is no longer used beyond tests, after we stopped using
'Plugins' concept. This fully cleans up the struct from tests that were
still using it.


pilot/pkg/networking/plugin/plugin.go
```diff
@@ -13,44 +13,3 @@
 // limitations under the License.
 
 package plugin
```
hzxuzhonghu, line 15: why keep a blank plugin

howardjohn: good point, will remove it

ericvn: LGTM after this change.

ramaraochavali (approved):

ericvn (approved):

lei-tang (approved):

Issue #40034, opened 2022-07-21
argo-rollouts pod can't match virtualservice sourceLabels


labaq:
### Bug Description

my virtualservice spec is
```
spec:
  hosts:
    - example-demo02-stable
  http:
    - match:
        - sourceLabels:
            role: canary
      route:
        - destination:
            host: example-demo02-canary
            port:
              number: 8080
    - route:
        - destination:
            host: example-demo02-stable
            port:
              number: 8080
```
and my rollout have set labels to stable pod and canary pod like this
```
    canary:
      canaryMetadata:
        labels:
          role: canary
      stableMetadata:
        labels:
          role: stable
```
on first canary the pod have the wright labels so the rule is match
but when canary done, the canary pod will become stable pod and the label is change to stable, but it still match the canary rule, i think istio have already record this pod have canary label but not update to stable label.
can i have update the record for the right labels?

### Version

```prose
client version: 1.13.5
control plane version: 1.15
data plane version: 1.13.5 (8 proxies)

Client Version: v1.24.1
Server Version: v1.22.8+k3s1
```

hzxuzhonghu:
I think i know what's the cause, you update the pod label at some time after it is stable, right? Unfortunately, istio does not update  proxy labels after startup.

labaq:
sound so bad, i just found if i restart the pod it will match my virtualservice rule
so there is no way can i update the proxy labels? or any workaround can  do?

hzxuzhonghu:
Take a look at my fix

labaq:
i'm not familiar with istio code, how can i test it😅

hzxuzhonghu:
Just build a new pilot discovery image and replace

m-wynn:
We're still seeing this issue.  If I understand correctly, the fixes did not make it into the Istio 1.15 release.  Are there any updates we need to do to do to get this working in our environment, or can we expect to see it in an upcoming release?

hzxuzhonghu:
right

Issue #39792, opened 2022-07-05
Locality based routing not working for istio-proxy to istio-egressgateway traffic...


diranged:
### Bug Description

When our traffic routes out from an `istio-proxy` sidecar through an `istio-egressgateway` pod, we want the traffic to stay within-zone whenever possible so that we can avoid paying cross-AZ network costs. For setting up the EgressGateways, we followed the docs at https://istio.io/latest/docs/tasks/traffic-management/egress/egress-gateway/ (with a minor change, I'll describe it in the "Additional Information" section below).

We have other high volume services within the mesh that use locality-aware routing to great success, so we do understand that it works fundamentally within the mesh. I can confirm that it works when we route traffic _from_ an `istio-ingressgateway` _to_ a pod with an `istio-proxy` sidecar. It just doesn't seem to work in the reverse flow out of an egressgateway.

I've posted on Slack (https://istio.slack.com/archives/C37A4KAAD/p1657034565248319, https://istio.slack.com/archives/C37A4KAAD/p1656698370327109) and also dug through the `discuss.istio.io` page for help on this but I haven't found the answer yet.

My concern here is that this may simply not be a supported thing to do - or if it is, there must be something in the documentation we're missing or not understanding..

### Version

```prose
$ istioctl version
client version: 1.13.3
control plane version: 1.14.0
data plane version: 1.14.0 (19 proxies)
```

```
$ kubectl version --short
Client Version: v1.23.5
Server Version: v1.22.9-eks-a64ea69
```
```


### Additional Information

Other than a fairly standard Istio installation, here are the resources we've configured to try to make the egress routing work. We followed the docs at https://istio.io/latest/docs/tasks/traffic-management/egress/egress-gateway/ pretty closely, however we made one change to reconfigure plaintext HTTP traffic (`http://httpbin.org`) to go out to the target destination via HTTPS:

## `Gateway: istio-egressgateway`
```yaml
apiVersion: networking.istio.io/v1beta1
kind: Gateway
metadata:
  generation: 3
  name: istio-egressgateway
  namespace: istio-system
spec:
  selector:
    istio: egressgateway
  servers:
  - hosts:
    - '*'
    port:
      name: http
      number: 80
      protocol: HTTP
  - hosts:
    - '*'
    port:
      name: tls
      number: 443
      protocol: TLS
    tls:
      mode: PASSTHROUGH
```

## `DestinationRule: istio-egressgateway`

[...]

PR #40036, opened 2022-07-21
Proxy labels should be updated when pod/wle labels updated

hzxuzhonghu:

Fix: #40034 and #39792

1. dynamically update proxy labels when pod/wle labels update

2. Added localityLabel to pod label so we can get proxy locality from pod labels rather than service instances, which may not exist


pilot/pkg/model/context.go
```diff
@@ -859,14 +859,20 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
 	node.ServiceInstances = instances
 }
 
-// SetWorkloadLabels will set the node.Metadata.Labels only when it is nil.
+// SetWorkloadLabels will set the node.Metadata.Labels with priority:
+// 1. Proxy workload labels.
+// 2. node meta labels.
 func (node *Proxy) SetWorkloadLabels(env *Environment) {
```
ramaraochavali, line 865: So is this merging pod labels and metadata labels giving preference to pod labels - which makes sense but the comment is confusing. May be we should say it merges both and gives preference to proxy label.

pilot/pkg/model/context.go
```diff
@@ -859,14 +859,20 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
...
+		node.Metadata.Labels = make(map[string]string)
+	}
+	// First get labels from proxy workload
+	for k, v := range labels {
+		node.Metadata.Labels[k] = v
 	}
-	// Fallback to calling GetProxyWorkloadLabels
-	node.Metadata.Labels = env.GetProxyWorkloadLabels(node)
+
+	// Fallback to get the workload labels from node meta
```
ramaraochavali, line 875: This is redundant and we can remove?

ramaraochavali (commented):
LGTM. Minor nits

pilot/pkg/model/context.go
```diff
@@ -859,14 +859,20 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
...
 
-// SetWorkloadLabels will set the node.Metadata.Labels only when it is nil.
+// SetWorkloadLabels will set the node.Metadata.Labels with priority:
+// 1. Proxy workload labels.
+// 2. node meta labels.
 func (node *Proxy) SetWorkloadLabels(env *Environment) {
-	// First get the workload labels from node meta
-	if len(node.Metadata.Labels) > 0 {
-		return
+	labels := env.GetProxyWorkloadLabels(node)
```
howardjohn, line 866: what if they remove a label? Won't we still keep it?

howardjohn (commented):
should we add an integration test? something like apply workloadSelector rule, assert it doesn't do anything, add label to pod assert it does something, remove label and assert it doesn't do anything?

pilot/pkg/model/context.go
adiprerepa: we should keep the previous proxy labels in the Proxy object and diff with the new

pilot/pkg/model/context.go
adiprerepa, line 875:
```suggestion
```

adiprerepa (commented):
does a pod label update directly trigger `computeProxyState`?

pilot/pkg/model/context.go
hzxuzhonghu: good catch, i will think about

hzxuzhonghu:
>does a pod label update directly trigger computeProxyState?

Not now, because we do not judge if labels change now. Currently the behavior is when we need to push xds to a proxy, the label change can be reflected.

If we want to trigger by labels update, we need to do some further refactor. I am willing to do it later.

diranged:
When reading through this, it isn't clear to me (and maybe that's fine!) how this fixes the egressgateway's not doing locality routing. Perhaps you can just explain that bug a bit more in the PR description so we can understand how this fixes that?

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -1365,6 +1365,12 @@ func (c *Controller) getProxyServiceInstancesByPod(pod *v1.Pod,
 func (c *Controller) GetProxyWorkloadLabels(proxy *model.Proxy) labels.Instance {
 	pod := c.pods.getPodByProxy(proxy)
 	if pod != nil {
+		locality := c.getPodLocality(pod)
+		if pod.Labels == nil {
+			pod.Labels = make(map[string]string)
+		}
+		// Add locality labels
+		pod.Labels[model.LocalityLabel] = locality
```
adiprerepa, line 1373: @diranged it's not entirely obvious, but previously, gateway locality wasn't being respected because we derived locality from just ServiceInstances. Now, we have added a locality label to Pods (including gateways), and we use that locality with (ads.go L622):

```go
if len(proxy.ServiceInstances) > 0 {
	localityStr = proxy.ServiceInstances[0].Endpoint.Locality.Label
} else {
	// If no service instances(this maybe common for a pure client), respect LocalityLabel
	localityStr = proxy.Metadata.Labels[model.LocalityLabel]
}
``` 
The gateway case false into the "else" here.

adiprerepa (commented):
once label removes are handled, lgtm

i'll push something for that.

pilot/pkg/xds/ads.go
```diff
@@ -655,24 +664,12 @@ func (s *DiscoveryServer) initializeProxy(node *core.Node, con *Connection) erro
...
-	}
-}
-
-func (s *DiscoveryServer) computeProxyState(proxy *model.Proxy, request *model.PushRequest) {
-	proxy.SetWorkloadLabels(s.Env)
+func (s *DiscoveryServer) computeProxyState(proxy *model.Proxy, request *model.PushRequest, skipLabels bool) {
 	proxy.SetServiceInstances(s.Env.ServiceDiscovery)
+	if !skipLabels {
+		proxy.SetWorkloadLabels(s.Env)
+		setTopologyLabels(proxy)
```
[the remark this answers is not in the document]
hzxuzhonghu: I think it is not matter much. `setTopologyLabels` are the labels generated by control plane.

And I tried to put it in SetWorkloadLabels, but there is a cycle import

imports istio.io/istio/pilot/pkg/model
	imports istio.io/istio/pilot/pkg/networking/util
	imports istio.io/istio/pilot/pkg/model

adiprerepa (approved):
lgtm

howardjohn:
> should we add an integration test? something like apply workloadSelector rule, assert it doesn't do anything, add label to pod assert it does something, remove label and assert it doesn't do anything?

can we add this test?

hzxuzhonghu:
Yes, @howardjohn I was asking for help on TF from @nmittler in slack at weekend

pilot/pkg/model/context.go
```diff
@@ -859,14 +864,23 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
...
-// SetWorkloadLabels will set the node.Metadata.Labels only when it is nil.
+// SetWorkloadLabels will set the node.Metadata.Labels.
+// It merges both node meta labels and workload labels and give preference to workload labels.
+// Note: must be called after `SetServiceInstances`.
 func (node *Proxy) SetWorkloadLabels(env *Environment) {
-	// First get the workload labels from node meta
-	if len(node.Metadata.Labels) > 0 {
-		return
+	labels := env.GetProxyWorkloadLabels(node)
+	if len(labels) > 0 {
```
howardjohn, line 872: seems we have a rare edge case when a pod has 0 labels we will not respect the GetProxyWorkloadLabels?

It is possible to have no labels

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -1365,6 +1365,12 @@ func (c *Controller) getProxyServiceInstancesByPod(pod *v1.Pod,
 func (c *Controller) GetProxyWorkloadLabels(proxy *model.Proxy) labels.Instance {
 	pod := c.pods.getPodByProxy(proxy)
 	if pod != nil {
+		locality := c.getPodLocality(pod)
+		if pod.Labels == nil {
+			pod.Labels = make(map[string]string)
+		}
+		// Add locality labels
```
howardjohn, line 1372: I thought we cannot mutate objects from informer?

pilot/pkg/serviceregistry/kube/controller/pod.go
```diff
@@ -111,6 +112,19 @@ func GetPodConditionFromList(conditions []v1.PodCondition, conditionType v1.PodC
 	return -1, nil
 }
 
+func (pc *PodCache) labelFilter(old, cur interface{}) bool {
```
howardjohn, line 115: Why is this a filter? It seems like it should be added to the queue and processed in onEvent?

pilot/pkg/xds/ads.go
```diff
@@ -745,8 +742,9 @@ func (s *DiscoveryServer) pushConnection(con *Connection, pushEv *Event) error {
 	pushRequest := pushEv.pushRequest
 
 	if pushRequest.Full {
+		skipLabel := !pushRequest.IsProxyUpdate()
```
howardjohn, line 745: can you add a comment _why_ we skip the labels in this case? is it expensive?

pkg/bootstrap/config.go
```diff
@@ -484,6 +484,7 @@ func extractAttributesMetadata(envVars []string, plat platform.Environment, meta
 		case "ISTIO_METAJSON_LABELS":
 			m := jsonStringToMap(val)
 			if len(m) > 0 {
+				meta.IstioMetaLabels = m
```
howardjohn, line 487: Its not clear to me why the client needs to _send_ both labels. I get why the control plane needs both, but the NodeMetadata here seems inappropriate. I would expect one label field in NodeMetadata that the client sends, and then one Label field not set by client -- probably in model.proxy top level -- with the computed state.

howardjohn: This may also be a breaking change since older clients do not send this.

pilot/pkg/serviceregistry/kube/controller/controller.go
howardjohn: https://github.com/istio/istio/issues/40097

pilot/pkg/model/context.go
hzxuzhonghu: Will respect all the comments here when rework

howardjohn:
Thanks! Will be great to have pod label updates supported, thanks for
working on this

PR #40047, opened 2022-07-21
Re-enable `gci` formatter as bugs are fixed

howardjohn:
Depends on https://github.com/istio/tools/pull/2068

Replicates https://github.com/istio/common-files/pull/633 and https://github.com/istio/common-files/pull/632 - we should merge those first then drop the common/ changes

### DIFF
--- a/common/config/.golangci-format.yml
+++ b/common/config/.golangci-format.yml
@@ -36,6 +36,7 @@ linters:
   enable:
   - goimports
   - gofumpt
+  - gci
   fast: false
 
 linters-settings:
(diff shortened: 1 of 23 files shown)


common/config/.golangci-format.yml
```diff
@@ -36,13 +36,19 @@ linters:
   enable:
   - goimports
   - gofumpt
+  - gci
```
zirain, line 39: this will broken `make gen`

liamawhite (approved):

Issue #39604, opened 2022-06-24
enableNamespacesByDefault doesn't work when revision tag is there and previous configuration has enableNamespacesByDefault disabled


sachin-ks:
### Bug Description

enableNamespacesByDefault doesn't work when revision tag is there and the previous configuration had enableNamespacesByDefault disabled.

### Version

```prose
$ istioctl version
client version: 1.14.1
control plane version: 1.14.1
data plane version: none

$ kubectl version --short
Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
Client Version: v1.24.1
Kustomize Version: v4.5.4
Server Version: v1.22.9-eks-a64ea69
```

### Steps to reproduce
save this as disabledByDefault.yaml :
```
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  namespace: istio-system
  name: istiocontrolplane
spec:
  profile: minimal
  revision: 1-14-1
```

save this as enabledByDefault.yaml :

```apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  namespace: istio-system
  name: istiocontrolplane
spec:
  profile: minimal
  revision: 1-14-1
  values:
    sidecarInjectorWebhook:
      enableNamespacesByDefault: true
```
      
Do:
`$ istioctl install -f disabledByDefault.yaml`
Then do:
`$ istioctl install -f enabledByDefault.yaml`

The configmap istio-sidecar-injector-1-14-1 shows `"enableNamespacesByDefault": true,`
But sidecar injection does not work when a pod is created.

litong01:
@sachin-ks followed your steps, and added the step to create a pod like the following:

```
kubectl create namespace sample

cat << EOF | kubectl apply -n sample -f -
apiVersion: apps/v1
kind: Deployment
metadata:
  name: helloworld-v1
  labels:
    app: helloworld
    istio.io/rev: 1-14-1
    version: v1
spec:
  replicas: 1
  selector:
    matchLabels:
      app: helloworld
      version: v1
  template:
    metadata:
      labels:
        app: helloworld
        version: v1
        sidecar.istio.io/inject: "true"
    spec:
      containers:
      - name: helloworld
        image: docker.io/istio/examples-helloworld-v1
        resources:
          requests:
            cpu: "100m"
        env:
        - name: FLASK_RUN_PORT
          value: "5000"
        - name: FLASK_RUN_HOST
          value: "::"
        - name: FLASK_ENV
[...]

PR #40050, opened 2022-07-21
Revert "Fix: allow enableNamespacesByDefault when revision tag is set"

Monkeyanator:
This commit made it so that we fail installation if a previous install exists without a default tag, which we don't want to do (having an installation without a default tag shouldn't be a critical error for a new install). In addition it makes the current revision the default even if a previous default tag exists, which we also want to avoid (it should be moved with the tag command).


therealmitchconnors:
It looks like the original PR was meant to close https://github.com/istio/istio/issues/39604.  Will this commit address that problem?

ostromart (approved):

lei-tang (approved):

ericvn (approved):

Monkeyanator:
@therealmitchconnors no but I think that's WAI. Right now the assumption is that if a default tag exists, user must run `istioctl tag` to update it (either to enable auto injection or switch the revision it points to)

PR #40053, opened 2022-07-21
Add a flag for app container name and retry verifyTrafficMirror

akshayjnambiar:
For some workloads the container name is different and the logs are written slower. So the verification of mirroring test cases is retried for 20 seconds.

### DIFF
--- a/tests/integration/pilot/mirror_test.go
+++ b/tests/integration/pilot/mirror_test.go
@@ -22,6 +22,7 @@ import (
 	"math"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/go-multierror"
 	"k8s.io/apimachinery/pkg/util/rand"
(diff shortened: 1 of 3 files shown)


tests/integration/pilot/mirror_test.go
```diff
@@ -155,9 +156,10 @@ func runMirrorTest(t *testing.T, options mirrorTestOptions) {
 										if expected == nil {
 											expected = apps.C
 										}
-
-										return verifyTrafficMirror(apps.B, expected, c, testID)
-									}, echo.DefaultCallRetryOptions()...)
+										return retry.UntilSuccess(func() error {
+											return verifyTrafficMirror(apps.B, expected, c, testID)
+										}, []retry.Option{retry.Timeout(20 * time.Second), retry.BackoffDelay(10 * time.Millisecond), retry.Converge(2)}...)
```
howardjohn, line 161: I would consider it broken if the logs are slow. In what environment are they slow?

howardjohn (approved):

PR #40056, opened 2022-07-22
fix lint

zirain:

pre-submit for https://github.com/istio/common-files/pull/632

### DIFF
--- a/tools/bug-report/pkg/kubeclient/kubeclient.go
+++ b/tools/bug-report/pkg/kubeclient/kubeclient.go
@@ -16,7 +16,6 @@ package kubeclient
 
 import (
 	"k8s.io/client-go/kubernetes"
-
 	//  Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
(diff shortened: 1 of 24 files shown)


tools/bug-report/pkg/kubeclient/kubeclient.go
```diff
@@ -16,7 +16,6 @@ package kubeclient
 
 import (
 	"k8s.io/client-go/kubernetes"
-
 	//  Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
```
howardjohn, line 19: missed this one?

howardjohn (approved):

ericvn (approved):
You might want to verify the changes against some other repos as well like `proxy`, `tools`, and `release builder`. Occasionally the go files there will be flag when the `common-files` changes get pushed to them.

tools/bug-report/pkg/kubeclient/kubeclient.go
howardjohn: do we need this one?

liamawhite (approved):

PR #40068, opened 2022-07-22
conditionally set multi-arch selector in the gateway templates

ZhengzheYang:
Delete the multi-arch selector because it's causing issue in the downstream tests.

### DIFF
--- a/manifests/charts/gateways/istio-egress/templates/_affinity.tpl
+++ b/manifests/charts/gateways/istio-egress/templates/_affinity.tpl
@@ -9,23 +9,25 @@ nodeAffinity:
 {{- end }}
 
 {{- define "nodeAffinityRequiredDuringScheduling" }}
+  {{- $nodeSelector := default .global.defaultNodeSelector .nodeSelector -}}
+  {{- if or .global.arch $nodeSelector }}
       nodeSelectorTerms:
       - matchExpressions:
+        {{- range $key, $val := .global.arch }}
         - key: kubernetes.io/arch
           operator: In
           values:
-        {{- range $key, $val := .global.arch }}
           {{- if gt ($val | int) 0 }}
           - {{ $key | quote }}
           {{- end }}
         {{- end }}
-        {{- $nodeSelector := default .global.defaultNodeSelector .nodeSelector -}}
         {{- range $key, $val := $nodeSelector }}
         - key: {{ $key }}
           operator: In
           values:
           - {{ $val | quote }}
         {{- end }}
+  {{- end }}
 {{- end }}
 
 {{- define "nodeAffinityPreferredDuringScheduling" }}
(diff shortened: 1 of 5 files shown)


manifests/charts/gateways/istio-egress/templates/_affinity.tpl
```diff
@@ -2,32 +2,10 @@
...
   {{- include "nodeAffinityPreferredDuringScheduling" . }}
 {{- end }}
 
-{{- define "nodeAffinityRequiredDuringScheduling" }}
-      nodeSelectorTerms:
-      - matchExpressions:
-        - key: kubernetes.io/arch
-          operator: In
-          values:
-        {{- range $key, $val := .global.arch }}
```
howardjohn, line 17: I think we still need ability to set it, just not default?

ZhengzheYang: I added a value to decide whether we want to set this. PTAL.

manifests/charts/gateways/istio-egress/templates/_affinity.tpl
```diff
@@ -11,6 +11,7 @@ nodeAffinity:
 {{- define "nodeAffinityRequiredDuringScheduling" }}
       nodeSelectorTerms:
       - matchExpressions:
+{{- if .global.multiArchs }}
```
howardjohn, line 14: maybe we can make it `if .global.arch`? Then its just one existing flag. If its empty, we do nothing. Otherwise we add it?

ZhengzheYang: I thought about it but does that mean we are deleting the archs from the `.global.arch`? I am not sure if we still need them for the preferred archs in lines https://github.com/istio/istio/blob/master/manifests/charts/gateways/istio-egress/templates/_affinity.tpl#L31-L43? If we are okay with deleting the archs, I can do it that way.

howardjohn: I think we should. The current list doesn'r even make sense, Istio only works on amd64 currently.

ZhengzheYang: then do you want me to delete them altogether or leave only amd64 in there? that would solve the multh-arch issue while keeping the one it actually works.

manifests/charts/gateways/istio-egress/templates/_affinity.tpl
```diff
@@ -9,23 +9,25 @@ nodeAffinity:
 {{- end }}
 
 {{- define "nodeAffinityRequiredDuringScheduling" }}
+  {{- $nodeSelector := default .global.defaultNodeSelector .nodeSelector -}}
+  {{- if or .global.arch .nodeSelector }}
```
howardjohn, line 13:
```suggestion
  {{- if or .global.arch $nodeSelector }}
```
I think we need that?

ZhengzheYang: Thanks! I thought I tested it. Sorry. Updated and tested again with dummy values.

howardjohn (approved):

Monkeyanator (approved):

PR #40072, opened 2022-07-22
pilot: fix issue with TLS and TCP order dependency

howardjohn:
Currently, tls_inspector is only added if its the first Service. If its
after TCP, it is missed.

example config:

```yaml
kind: ServiceEntry
metadata:
  name: tcp
spec:
  addresses:
  - 10.10.10.10/24
  exportTo:
  - .
  hosts:
  - '*.tcp'
  ports:
  - name: tcp-443
    number: 443
    protocol: TCP
---
kind: ServiceEntry
metadata:
  name: tls
spec:
  exportTo:
  - .
  hosts:
  - tls.example.com
  location: MESH_EXTERNAL
  ports:
  - name: https-443
    number: 443
    protocol: HTTPS
  resolution: DNS
```

depending on the order of creation it will work/not work.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/listener.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener.go
@@ -984,6 +984,7 @@ func (lb *ListenerBuilder) buildSidecarOutboundListenerForPortOrUDS(listenerOpts
 		if currentListenerEntry != nil {
 			currentListenerEntry.listener.FilterChains = mergeTCPFilterChains(mutable.Listener.FilterChains,
 				listenerOpts, listenerMapKey, listenerMap)
+			currentListenerEntry.listener.ListenerFilters = mergeListenerFilters(currentListenerEntry.listener.ListenerFilters, mutable.Listener.ListenerFilters)
 		} else {
 			listenerMap[listenerMapKey] = &outboundListenerEntry{
 				services:    []*model.Service{listenerOpts.service},
@@ -1009,10 +1010,12 @@ func (lb *ListenerBuilder) buildSidecarOutboundListenerForPortOrUDS(listenerOpts
 		// Merge two TCP filter chains. HTTP filter chain will not conflict with TCP filter chain because HTTP filter chain match for
 		// HTTP filter chain is different from TCP filter chain's.
 		currentListenerEntry.listener.FilterChains = mergeTCPFilterChains(mutable.Listener.FilterChains, listenerOpts, listenerMapKey, listenerMap)
+		currentListenerEntry.listener.ListenerFilters = mergeListenerFilters(currentListenerEntry.listener.ListenerFilters, mutable.Listener.ListenerFilters)
 	case TCPOverAuto:
 		// Merge two TCP filter chains. HTTP filter chain will not conflict with TCP filter chain because HTTP filter chain match for
 		// HTTP filter chain is different from TCP filter chain's.
 		currentListenerEntry.listener.FilterChains = mergeTCPFilterChains(mutable.Listener.FilterChains, listenerOpts, listenerMapKey, listenerMap)
+		currentListenerEntry.listener.ListenerFilters = mergeListenerFilters(currentListenerEntry.listener.ListenerFilters, mutable.Listener.ListenerFilters)
 
 	case AutoOverHTTP:
 		listenerMap[listenerMapKey] = &outboundListenerEntry{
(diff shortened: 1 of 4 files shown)


hzxuzhonghu:
When did this bug occur?

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -1009,10 +1010,12 @@ func (lb *ListenerBuilder) buildSidecarOutboundListenerForPortOrUDS(listenerOpts
 		// Merge two TCP filter chains. HTTP filter chain will not conflict with TCP filter chain because HTTP filter chain match for
 		// HTTP filter chain is different from TCP filter chain's.
 		currentListenerEntry.listener.FilterChains = mergeTCPFilterChains(mutable.Listener.FilterChains, listenerOpts, listenerMapKey, listenerMap)
+		currentListenerEntry.listener.ListenerFilters = mergeListenerFilters(currentListenerEntry.listener.ListenerFilters, mutable.Listener.ListenerFilters)
```
hzxuzhonghu, line 1013: The example is tcp over tcp, could not understand why merge listener filter for NoConflict and `TCPOverAuto`

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -1539,6 +1542,29 @@ func isConflictWithWellKnownPort(incoming, existing protocol.Instance, conflict
 	return true
 }
 
+func mergeListenerFilters(a, b []*listener.ListenerFilter) []*listener.ListenerFilter {
```
hzxuzhonghu, line 1545: similar to appendListenerFiltersm what's the diff?

hzxuzhonghu, line 1545: I think the target of this function is  to return a union b, if so the implement is not correct

howardjohn: Why is it not correct?

howardjohn:
> When did this bug occur?

There is an example of 2 SE in the PR description

hzxuzhonghu:
I am asking to know if this is a regression

ericvn (approved):
Release note looks good.

howardjohn:
@hzxuzhonghu AFAIK its been broken forever

pilot/pkg/networking/core/v1alpha3/listener.go
howardjohn: Can you elaborate?

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -1009,10 +1010,12 @@ func (lb *ListenerBuilder) buildSidecarOutboundListenerForPortOrUDS(listenerOpts
 		// Merge two TCP filter chains. HTTP filter chain will not conflict with TCP filter chain because HTTP filter chain match for
 		// HTTP filter chain is different from TCP filter chain's.
 		currentListenerEntry.listener.FilterChains = mergeTCPFilterChains(mutable.Listener.FilterChains, listenerOpts, listenerMapKey, listenerMap)
+		currentListenerEntry.listener.ListenerFilters = mergeListenerFilters(currentListenerEntry.listener.ListenerFilters, mutable.Listener.ListenerFilters)
 	case TCPOverAuto:
 		// Merge two TCP filter chains. HTTP filter chain will not conflict with TCP filter chain because HTTP filter chain match for
 		// HTTP filter chain is different from TCP filter chain's.
 		currentListenerEntry.listener.FilterChains = mergeTCPFilterChains(mutable.Listener.FilterChains, listenerOpts, listenerMapKey, listenerMap)
+		currentListenerEntry.listener.ListenerFilters = mergeListenerFilters(currentListenerEntry.listener.ListenerFilters, mutable.Listener.ListenerFilters)
```
hzxuzhonghu, line 1018: Can you elaborate on why do it here? I think  for auto, we have already added inspector filter for it.

pilot/pkg/networking/core/v1alpha3/listener.go
hzxuzhonghu: I mean you just return the combination of a and b  for `TLSInspector` and `HTTPInspector`.

So how about b contains another listener filter that a does not contain?

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -984,6 +984,7 @@ func (lb *ListenerBuilder) buildSidecarOutboundListenerForPortOrUDS(listenerOpts
 		if currentListenerEntry != nil {
 			currentListenerEntry.listener.FilterChains = mergeTCPFilterChains(mutable.Listener.FilterChains,
 				listenerOpts, listenerMapKey, listenerMap)
+			currentListenerEntry.listener.ListenerFilters = mergeListenerFilters(currentListenerEntry.listener.ListenerFilters, mutable.Listener.ListenerFilters)
```
hzxuzhonghu, line 987: Can you elaborate here why need mergeListenerFilters for NoConflict

howardjohn:
@hzxuzhonghu I modified the PR and test to be generic regardless of conflict type - we always take the union of the filters when we merge. The test also tests all possible protocol combinations. PTAL.

Note: the behavior is the same as original PR, just simpler implementation

hzxuzhonghu (approved):
This is better

PR #40082, opened 2022-07-23
minior improve ecds generate

zirain:

- refactor secretConfigsUpdatedOnly
- skip when only secret updated but no referenced secret

### DIFF
--- a/pilot/pkg/xds/ecds.go
+++ b/pilot/pkg/xds/ecds.go
@@ -59,49 +59,26 @@ func ecdsNeedsPush(req *model.PushRequest) bool {
 	return false
 }
 
+// Check if the secret updates is relevant to Wasm image pull. If not relevant, skip pushing ECDS.
+// this occurs when user only update the Secret referenced by WasmPlugin/EnvoyFilter.
+func secretConfigsUpdatedOnly(req *model.PushRequest) bool {
+	return !model.ConfigsHaveKind(req.ConfigsUpdated, kind.WasmPlugin) &&
+		!model.ConfigsHaveKind(req.ConfigsUpdated, kind.EnvoyFilter) &&
+		model.ConfigsHaveKind(req.ConfigsUpdated, kind.Secret)
+}
+
 // Generate returns ECDS resources for a given proxy.
 func (e *EcdsGenerator) Generate(proxy *model.Proxy, w *model.WatchedResource, req *model.PushRequest) (model.Resources, model.XdsLogDetails, error) {
 	if !ecdsNeedsPush(req) {
 		return nil, model.DefaultXdsLogDetails, nil
 	}
 
-	secretResources := referencedSecrets(proxy, req.Push, w.ResourceNames)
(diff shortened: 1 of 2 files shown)


pilot/pkg/xds/ecds.go
```diff
@@ -59,6 +59,14 @@ func ecdsNeedsPush(req *model.PushRequest) bool {
 	return false
 }
 
+// Check if the secret updates is relevant to Wasm image pull. If not relevant, skip pushing ECDS.
+// this occurs when user only update the Secret referenced by WasmPlugin/EnvoyFilter.
+func secretConfigsUpdatedOnly(req *model.PushRequest) bool {
```
hzxuzhonghu, line 64: I would prefer also to abstract `updatedSecrets` cal into this function and rename it, as it is very confusing

zirain: updated, PTAL.

pilot/pkg/xds/ecds.go
```diff
@@ -59,65 +59,92 @@ func ecdsNeedsPush(req *model.PushRequest) bool {
...
+
+	return resources, model.DefaultXdsLogDetails, nil
+}
+
+func (e *EcdsGenerator) pullSecrets(proxy *model.Proxy, w *model.WatchedResource, req *model.PushRequest) (secrets map[string][]byte, needsPush bool) {
 	secretResources := referencedSecrets(proxy, req.Push, w.ResourceNames)
-	// Check if the secret updates is relevant to Wasm image pull. If not relevant, skip pushing ECDS.
-	if !model.ConfigsHaveKind(req.ConfigsUpdated, kind.WasmPlugin) && !model.ConfigsHaveKind(req.ConfigsUpdated, kind.EnvoyFilter) &&
-		model.ConfigsHaveKind(req.ConfigsUpdated, kind.Secret) {
+	if secretConfigsUpdatedOnly(req) {
```
hzxuzhonghu, line 99: IMO, this is an optimization to skip push ecds only when secret update, but the secret is not relevant. Add more comments

pilot/pkg/xds/ecds.go
```diff
@@ -59,65 +59,92 @@ func ecdsNeedsPush(req *model.PushRequest) bool {
...
+		resources = append(resources, &discovery.Resource{
+			Name:     c.Name,
+			Resource: protoconv.MessageToAny(c),
+		})
+	}
+
+	return resources, model.DefaultXdsLogDetails, nil
+}
+
+func (e *EcdsGenerator) pullSecrets(proxy *model.Proxy, w *model.WatchedResource, req *model.PushRequest) (secrets map[string][]byte, needsPush bool) {
```
hzxuzhonghu, line 97: The function name is confusing

pilot/pkg/xds/ecds.go
hzxuzhonghu: Maybe better to move the if clause out

PR #39058, opened 2022-05-20
Fix stale svc clusterVIP fo multicluster

hzxuzhonghu:
**Please provide a description of this PR:**

Fix https://github.com/istio/istio/issues/39039

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #40083, opened 2022-07-23
fix concurrent map access in debug handlers

ramaraochavali:
Fixes https://github.com/istio/istio/issues/36448

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/debug.go
+++ b/pilot/pkg/xds/debug.go
@@ -306,7 +306,7 @@ func (s *DiscoveryServer) registryz(w http.ResponseWriter, req *http.Request) {
 	for _, svc := range all {
 		copied = append(copied, svc.DeepCopy())
 	}
-	writeJSON(w, all, req)
+	writeJSON(w, copied, req)
 }
 
 // Dumps info about the endpoint shards, tracked using the new direct interface.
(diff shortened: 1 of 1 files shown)


pilot/pkg/xds/debug.go
```diff
@@ -302,7 +302,11 @@ func (s *DiscoveryServer) Syncz(w http.ResponseWriter, req *http.Request) {
 // Can be combined with the push debug interface to reproduce changes.
 func (s *DiscoveryServer) registryz(w http.ResponseWriter, req *http.Request) {
 	all := s.Env.ServiceDiscovery.Services()
-	writeJSON(w, all, req)
+	copied := make([]*model.Service, len(all))
+	for _, svc := range all {
+		copied = append(copied, svc.DeepCopy())
```
hzxuzhonghu, line 307: maybe it has been fixed by #39058, whatever we need a test

ramaraochavali: I don't this it is fixed by that. The concurrent map write might still come if JSON encoder is encoding AddressMap while it is modified. There is already a test but it is difficult to create a test for this I guess. Any ideas?

ramaraochavali: @hzxuzhonghu any thoughts?

hzxuzhonghu: I believe there is no mutating of `AddressMap` before deepcopy now. So this can not happen now.

ramaraochavali: Ok. Cool I will close this

PR #40105, opened 2022-07-25
Gcp concurrent requests

rinormaloku:
This PR resolves issue #39950. By implementing these two changes:
- Execute queries to the GCP Metadata API as a batch whenever possible.
- Checking if the API is accessible in the first place.

- [x] Installation
- [x] Sidecar

### DIFF
--- a/pkg/bootstrap/platform/gcp.go
+++ b/pkg/bootstrap/platform/gcp.go
@@ -55,8 +55,10 @@ var (
 )
 
 var (
+	// shouldFillMetadata returns whether the workload is running on GCP and the metadata endpoint is accessible
+	// In contrast, DiscoverWithTimeout only checks if the workload is running on GCP
 	shouldFillMetadata = func() bool {
-		return metadata.OnGCE() && IsMetadataEndpointAccessible()
+		return metadata.OnGCE() && isMetadataEndpointAccessible()
 	}
 	projectIDFn        = metadata.ProjectID
 	numericProjectIDFn = metadata.NumericProjectID
@@ -110,8 +112,6 @@ var (
 	}
 )
 
-var isMetadataEndpointAccessible bool
-
 type (
 	shouldFillFn     func() bool
 	metadataFn       func() (string, error)
@@ -123,8 +123,11 @@ type (
 
 type gcpEnv struct {
 	sync.Mutex
-	metadata map[string]string
+	metadata           map[string]string
+	shouldFillMetadata bool
 }
+
+var gcpEnvOnce sync.Once
 
 // IsGCP returns whether or not the platform for bootstrapping is Google Cloud Platform.
 func IsGCP() bool {
@@ -145,11 +148,15 @@ func NewGCP() Environment {
 // Metadata returns GCP environmental data, including project, cluster name, and
 // location information.
 func (e *gcpEnv) Metadata() map[string]string {
+	gcpEnvOnce.Do(func() {
+		e.shouldFillMetadata = shouldFillMetadata()
+	})
+
 	md := map[string]string{}
 	if e == nil {
 		return md
 	}
-	if GCPMetadata == "" && !shouldFillMetadata() {
+	if GCPMetadata == "" && !e.shouldFillMetadata {
 		return md
 	}
 
@@ -158,50 +165,49 @@ func (e *gcpEnv) Metadata() map[string]string {
 	if e.metadata != nil {
 		return e.metadata
 	}
-	envPid, envNPid, envCN, envLoc, err := parseGCPMetadata()
+	envPid, envNPid, envCN, envLoc := parseGCPMetadata()
 
-	// If no error is returned, then we have a valid GCP environment.
-	// Else we query the API to discover the needed information
-	if err == nil {
+	if envPid != "" {
 		md[GCPProject] = envPid
+	}
+	if envNPid != "" {
 		md[GCPProjectNumber] = envNPid
+	}
+	if envLoc != "" {
 		md[GCPLocation] = envLoc
+	}
+	if envCN != "" {
 		md[GCPCluster] = envCN
-	} else if IsMetadataEndpointAccessible() {
-		wg := sync.WaitGroup{}
+	}
 
-		suppliers := []metadataSupplier{
-			createMetadataSupplier(GCPProject, projectIDFn),
-			createMetadataSupplier(GCPProjectNumber, numericProjectIDFn),
-			createMetadataSupplier(GCPCluster, clusterNameFn),
-			createMetadataSupplier(GCPLocation, clusterLocationFn),
+	if e.shouldFillMetadata {
+		// suppliers is an array of functions that supply the metadata for missing properties
+		var suppliers []metadataSupplier
+		if _, found := md[GCPProject]; !found {
+			suppliers = append(suppliers, createMetadataSupplier(GCPProject, projectIDFn))
+		}
+		if _, found := md[GCPProjectNumber]; !found {
+			suppliers = append(suppliers, createMetadataSupplier(GCPProjectNumber, numericProjectIDFn))
+		}
+		if _, found := md[GCPLocation]; !found {
+			suppliers = append(suppliers, createMetadataSupplier(GCPLocation, clusterLocationFn))
+		}
+		if _, found := md[GCPCluster]; !found {
+			suppliers = append(suppliers, createMetadataSupplier(GCPCluster, clusterNameFn))
 		}
 
-		for _, mdSupplier := range suppliers {
-			wg.Add(1)
-			property, supplierFunction := mdSupplier.Property, mdSupplier.Fn
-			go func() {
-				defer wg.Done()
-				if result, err := supplierFunction(); err == nil {
-					md[property] = result
-				} else {
-					log.Warnf("Error fetching GCP Metadata property %s: %v", property, err)
-				}
-			}()
-		}
-
+		wg := waitForMetadataSuppliers(suppliers, md)
 		wg.Wait()
 	}
+
 	if GCPQuotaProjectVar != "" {
 		md[GCPQuotaProject] = GCPQuotaProjectVar
 	}
 	// Exit early now if not on GCE. This allows setting env var when not on GCE.
-	if !shouldFillMetadata() {
+	if !e.shouldFillMetadata {
 		e.metadata = md
 		return md
 	}
-
-	wg := sync.WaitGroup{}
 
 	suppliers := []metadataSupplier{
 		createMetadataSupplier(GCEInstance, instanceNameFn),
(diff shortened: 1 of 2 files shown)


pkg/bootstrap/platform/gcp.go
```diff
@@ -148,26 +158,39 @@ func (e *gcpEnv) Metadata() map[string]string {
...
-		md[GCPLocation] = l
-	}
-	if envCN != "" {
 		md[GCPCluster] = envCN
-	} else if cn, err := clusterNameFn(); err == nil {
-		md[GCPCluster] = cn
+	} else if IsMetadataEndpointAccessible() {
+		wg := sync.WaitGroup{}
+
+		suppliers := []metadataSupplier{
```
rinormaloku, line 173: @zirain if you have any suggestion/pattern that I can implement here to make it more readable, let me know and I will update the code accordingly

zirain: I prefer to change the implements of `IsGCP`, just detected metadata endpoint?
@howardjohn WDYT?

howardjohn:
In https://github.com/istio/istio/blob/e2ce47e2cb85306bcf57160ae3c4a9026cb28018/pkg/bootstrap/platform/discovery.go#L64 we already set a timeout on the discovery. Is the issue that we are not calling these GCP meta through that function (and thus ignoring timeout)? Or that we are, but the timeout is too slow anyways?

rinormaloku:
@howardjohn IsGCP is returning true as it uses DNS resolution as a means of identifying if the instance is in GCP. However, that doesn't mean that we can access the Metadata endpoint (i.e. port 80 could be blocked by the firewall). 

Thus all requests to the Metadata API are taking 10+ seconds and those were not parallelized so it adds up to about 2 minutes (that's what the PR changes as well as validating if a TCP connection to the Metadata endpoint is possible)

(More details in the original issue #39950)

howardjohn:
Thanks for the extra details, makes more sense. One last thing that isn't 100% clear is why we need isMetadataEndpointAccessibleFn _and_ parallelism. I would have thought the former is sufficient to avoid this issue. Is the parallelism just an additional optimization?

rinormaloku:
> Thanks for the extra details, makes more sense. One last thing that isn't 100% clear is why we need isMetadataEndpointAccessibleFn _and_ parallelism. I would have thought the former is sufficient to avoid this issue. Is the parallelism just an additional optimization?

Yes, the parallelism is just further optimization.

pkg/bootstrap/platform/gcp.go
```diff
@@ -52,8 +54,12 @@ var (
 	GCPQuotaProjectVar = env.RegisterStringVar("GCP_QUOTA_PROJECT", "", "Allows specification of a quota project to be used in requests to GCP APIs.").Get()
 )
 
+var isMetadataEndpointAccessible bool
+
 var (
-	shouldFillMetadata = metadata.OnGCE
+	shouldFillMetadata = func() bool {
+		return metadata.OnGCE() && isMetadataEndpointAccessibleFn()
```
howardjohn, line 61: There seems to be some disconnect between this and DiscoverWithTimeout. DiscoverWithTimeout is only checking OnGCE - if we cannot actually access the metadata server should we still consider it "On GCP" or just use the generic (none) platform?

if we do want to consider it "On GCP" but then not "shouldFillMetadata", I feel like we need some comments linking these two and explaining the interaction, etc.

We may also want to align the timeouts. DiscoverWithTimeout has a 5s timeout on OnGCE, but here we have no timeout on OnGCE but a 10s timeout on isMetadataEndpointAccessibleFn

pkg/bootstrap/platform/gcp.go
```diff
@@ -235,7 +294,7 @@ func zoneToRegion(z string) (string, error) {
 // Locality returns the GCP-specific region and zone.
 func (e *gcpEnv) Locality() *core.Locality {
 	var l core.Locality
-	if metadata.OnGCE() {
+	if metadata.OnGCE() && isMetadataEndpointAccessibleFn() {
```
howardjohn, line 297: should we just use `shouldFillMetadata`?

pkg/bootstrap/platform/gcp.go
```diff
@@ -52,8 +54,12 @@ var (
 	GCPQuotaProjectVar = env.RegisterStringVar("GCP_QUOTA_PROJECT", "", "Allows specification of a quota project to be used in requests to GCP APIs.").Get()
 )
 
+var isMetadataEndpointAccessible bool
+
 var (
-	shouldFillMetadata = metadata.OnGCE
+	shouldFillMetadata = func() bool {
```
howardjohn, line 60: Do we want just a function defined here? it seems we end up calling this quite a bit.. does it make sense to store the result in the `gcpEnv` struct?

pkg/bootstrap/platform/gcp.go
```diff
@@ -104,11 +110,29 @@ var (
 		return fmt.Sprintf("https://container.googleapis.com/v1/projects/%s/locations/%s/clusters/%s",
 			projectID, clusterLocation, clusterName), nil
 	}
+
+	isMetadataEndpointAccessibleFn = func() bool {
+		envOnce.Do(func() {
```
howardjohn, line 115: See comment above, we may be able to avoid a `Once` if we store it in gcpEnv? Or we can have it still in `Once` to lazy load it, but scoped to gcpEnv instead of global?

pkg/bootstrap/platform/gcp.go
```diff
@@ -177,18 +216,33 @@ func (e *gcpEnv) Metadata() map[string]string {
...
+
+	suppliers := []metadataSupplier{
+		createMetadataSupplier(GCEInstance, instanceNameFn),
+		createMetadataSupplier(GCEInstanceID, instanceIDFn),
+		createMetadataSupplier(GCEInstanceTemplate, instanceTemplateFn),
+		createMetadataSupplier(GCEInstanceCreatedBy, createdByFn),
 	}
-	if cb, err := createdByFn(); err == nil {
-		md[GCEInstanceCreatedBy] = cb
+
```
howardjohn, line 226: Seems identical to L192, should we abstract it out?

pkg/bootstrap/platform/gcp.go
```diff
@@ -218,7 +272,12 @@ func parseGCPMetadata() (pid, npid, cluster, location string) {
 			}
 		}
 	})
-	return envPid, envNpid, envCluster, envLocation
+
+	if envPid == "" || envNpid == "" || envCluster == "" || envLocation == "" {
```
howardjohn, line 276: It seems like it should be valid to only  set a subset of metadata?

douglas-reid, line 276: i don't think this is quite right, and will cause issues (at least somewhere). It is entirely possible for someone to use `GCP_METADATA=my-project|||`, etc.

This appears to add a requirement that all evaluate to something that isn't the empty string, if I understand correctly. And the fallback is that if any are empty, then all must be retrieved from the MD server?  I think we still need the support for partial parsing.

pkg/bootstrap/platform/gcp.go
```diff
@@ -148,26 +172,41 @@ func (e *gcpEnv) Metadata() map[string]string {
 	if e.metadata != nil {
 		return e.metadata
 	}
-	envPid, envNPid, envCN, envLoc := parseGCPMetadata()
-	if envPid != "" {
+	envPid, envNPid, envCN, envLoc, err := parseGCPMetadata()
+
+	// If no error is returned, then we have a valid GCP environment.
+	// Else we query the API to discover the needed information
+	if err == nil {
```
douglas-reid, line 179: i think we want per-value checking, with a single available check.

```
md[GCPProject] = envPid
...
if isMetadataEndpointAccessibleFn() {
  // fill in empty values if possible
  if md[GCPProject] == "" {...}
  ...
}

Thoughts?

**Review by douglas-reid — CHANGES_REQUESTED** [review]

File: pkg/bootstrap/platform/gcp.go (hunk 2 above)
Comment (rinormaloku) [reply pkg/bootstrap/platform/gcp.go]: >  if we cannot actually access the metadata server should we still consider it "On GCP" or just use the generic (none) platform?

I opted to keep GcpEnv (instead of none) just in case that at some point the environments are used for more then just metadata.

I added a comment in the code and aligned the timeouts

File: pkg/bootstrap/platform/gcp.go (hunk 3 above)
Comment (rinormaloku) [reply pkg/bootstrap/platform/gcp.go]: I followed your suggestion of storing the result in the gcpEnv object.

File: pkg/bootstrap/platform/gcp.go (hunk 5 above)
Comment (rinormaloku) [reply pkg/bootstrap/platform/gcp.go]: That's a great idea. I updated the code according to your suggestion

File: pkg/bootstrap/platform/gcp.go (hunk 6 above)
Comment (rinormaloku) [reply pkg/bootstrap/platform/gcp.go:212]: done

File: pkg/bootstrap/platform/gcp.go [hunk 9]
```diff
@@ -104,11 +110,29 @@ var (
 		return fmt.Sprintf("https://container.googleapis.com/v1/projects/%s/locations/%s/clusters/%s",
 			projectID, clusterLocation, clusterName), nil
 	}
+
+	isMetadataEndpointAccessibleFn = func() bool {
```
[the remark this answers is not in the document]
rinormaloku: > so the issue is that resolution is OK, but actual connection is blocked?

Yes.

> And the issue is that using something like metadata.Zone() as a proxy won't work because the default retry is too long?

The issue is that we have a series of requests that each takes about 10 seconds because the connection is not possible,

With the change we discover that the connection is not possible early on, and we skip all those attempts.

pkg/bootstrap/platform/gcp.go
rinormaloku: It makes sense, I reverted the changes to allow for partial parsing. Though this section of the code is very difficult to read.

pkg/bootstrap/platform/gcp.go
rinormaloku: I updated this a little to follow your suggestion, but assigning empty values to the map causes some unit tests to fail ( because of the method `constructGKEClusterURL` that expects entries to either be valid or not present), we can update the logic in that instance, but maybe there are other areas of the code that expect the same. (We can clean up all those areas if you think it's worthwhile)

howardjohn (approved):
Thanks!

therealmitchconnors (approved):

howardjohn:
If there were plans to backport, please hold off for a bit. Got some reports on our internal CI that this caused some issues for us. Will report back once we root cause - could just be an issue on our side

zirain:
@howardjohn this seems should be backport to v1.14 and v1.15?

howardjohn:
This PR is broken without the fix: https://github.com/istio/istio/pull/40700. need to backpotr both

lei-tang:
https://github.com/istio/istio/pull/40791 backports https://github.com/istio/istio/pull/40700 to release-1.13.

BRONSOLO:
@howardjohn @zirain is this fix intended to land on v1.14 + v1.15 / any ETA on this if so? Thanks!

PR #40109, opened 2022-07-26
optimize handleStats temporary big byte slice

### LINKED ISSUE #40092: Every stats scraping will create big temporary bytes slices

When the size of envoy metrics is big,  every stats scraping will create a big temporary bytes slice. This puts pressure on the golang gc.
see `allocate_space` profiler graph:
image
file: [profile (2).pb.gz](https://github.com/istio/istio/files/9179697/profile.2.pb.gz)

One alternative solution:
Using a global stats bytes buffer. And reseting buffer on beginning of the pilot-agent scrap requests.  If the size of metrics becomes smaller, we truncate the buffer on the end of scrap requests.   Protect global stats bytes buffer access by a mutex.

Pros:
    1.  Avoids frequent object creation.
Cons:
    1. No concurrent data access.
    2. Not release buffer if never be scraped.
    
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

Patrick0308:
fix #40092


Patrick0308:
@hzxuzhonghu PTAL

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -508,29 +522,44 @@ func (s *Server) handleStats(w http.ResponseWriter, r *http.Request) {
...
 		metrics.AgentScrapeErrors.Increment()
 	}
+
 	if envoy != nil {
-		if _, err := w.Write(envoy); err != nil {
-			log.Errorf("failed to write envoy metrics: %v", err)
-			metrics.EnvoyScrapeErrors.Increment()
+		if format == expfmt.FmtOpenMetrics {
+			envoyStatsBuffer := &bytes.Buffer{}
+			_, err = envoyStatsBuffer.ReadFrom(envoy)
```
sschepens, line 536: Isn't there a way to process metrics incrementally instead of reading the whole response in memory?

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -508,29 +522,44 @@ func (s *Server) handleStats(w http.ResponseWriter, r *http.Request) {
...
+			if err != nil {
+				log.Errorf("failed scraping envoy metrics: %v", err)
+				metrics.EnvoyScrapeErrors.Increment()
+			}
+			standardBytes := processMetrics(envoyStatsBuffer.Bytes())
```
sschepens, line 541: this method seems to only be replacing `\n\n` for `\n` this could most definitely be done incrementally, reading chunks and replacing on those chunks, maybe an edge case if the last byte of the chunk was `\n` and the first one from the next is also `\n`, but can be handled.

howardjohn (changes requested):
Any performance optimizations would be best to include benchmark results before and after

Patrick0308:
Before optimize:
```
BenchmarkStats/stats-fmttext
BenchmarkStats/stats-fmttext-8         	     493	   2416748 ns/op	 2664353 B/op	    1310 allocs/op
BenchmarkStats/stats-fmttext-8         	     519	   2475154 ns/op	 2664140 B/op	    1310 allocs/op
BenchmarkStats/stats-fmttext-8         	     492	   2492018 ns/op	 2665170 B/op	    1310 allocs/op
BenchmarkStats/stats-fmttext-8         	     475	   2535687 ns/op	 2662752 B/op	    1309 allocs/op
BenchmarkStats/stats-fmttext-8         	     469	   2491757 ns/op	 2665791 B/op	    1310 allocs/op
BenchmarkStats/stats-fmtopenmetrics
BenchmarkStats/stats-fmtopenmetrics-8  	     480	   2475891 ns/op	 2664162 B/op	    1309 allocs/op
BenchmarkStats/stats-fmtopenmetrics-8  	     482	   2523144 ns/op	 2662922 B/op	    1309 allocs/op
BenchmarkStats/stats-fmtopenmetrics-8  	     480	   2493956 ns/op	 2665636 B/op	    1310 allocs/op
BenchmarkStats/stats-fmtopenmetrics-8  	     481	   2499650 ns/op	 2663401 B/op	    1310 allocs/op
BenchmarkStats/stats-fmtopenmetrics-8  	     482	   2554097 ns/op	 2664494 B/op	    1309 allocs/op
```

Patrick0308:
After optimize:
```
goos: linux
goarch: amd64
pkg: istio.io/istio/pilot/cmd/pilot-agent/status
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
BenchmarkStats
BenchmarkStats/stats-fmttext
BenchmarkStats/stats-fmttext-8         	     802	   1422399 ns/op	 1044493 B/op	    1256 allocs/op
BenchmarkStats/stats-fmttext-8         	     862	   1398682 ns/op	 1043458 B/op	    1254 allocs/op
BenchmarkStats/stats-fmttext-8         	     865	   1384312 ns/op	 1042661 B/op	    1255 allocs/op
BenchmarkStats/stats-fmttext-8         	     885	   1428625 ns/op	 1041741 B/op	    1255 allocs/op
BenchmarkStats/stats-fmttext-8         	     816	   1400443 ns/op	 1039389 B/op	    1255 allocs/op
BenchmarkStats/stats-fmtopenmetrics
BenchmarkStats/stats-fmtopenmetrics-8  	     642	   1770525 ns/op	 1338681 B/op	    1268 allocs/op
BenchmarkStats/stats-fmtopenmetrics-8  	     682	   1687488 ns/op	 1339349 B/op	    1270 allocs/op
BenchmarkStats/stats-fmtopenmetrics-8  	     729	   1682919 ns/op	 1338682 B/op	    1270 allocs/op
BenchmarkStats/stats-fmtopenmetrics-8  	     711	   1708028 ns/op	 1340631 B/op	    1272 allocs/op
BenchmarkStats/stats-fmtopenmetrics-8  	     688	   1690986 ns/op	 1337924 B/op	    1271 allocs/op
```

Patrick0308:
@howardjohn @sschepens  PTAL

sschepens:
Just pasting the comparison of the two benchmarks using `benchstat`
```
name                          old time/op    new time/op    delta
Stats/stats-fmttext-8           2.48ms ± 3%    1.41ms ± 2%  -43.32%  (p=0.008 n=5+5)
Stats/stats-fmtopenmetrics-8    2.51ms ± 2%    1.71ms ± 4%  -31.93%  (p=0.008 n=5+5)

name                          old alloc/op   new alloc/op   delta
Stats/stats-fmttext-8           2.66MB ± 0%    1.04MB ± 0%  -60.88%  (p=0.008 n=5+5)
Stats/stats-fmtopenmetrics-8    2.66MB ± 0%    1.34MB ± 0%  -49.74%  (p=0.008 n=5+5)

name                          old allocs/op  new allocs/op  delta
Stats/stats-fmttext-8            1.31k ± 0%     1.25k ± 0%   -4.20%  (p=0.016 n=4+5)
Stats/stats-fmtopenmetrics-8     1.31k ± 0%     1.27k ± 0%   -2.99%  (p=0.008 n=5+5)
```

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -481,24 +487,38 @@ type PrometheusScrapeConfiguration struct {
...
+	}()
 
 	// Gather all the metrics we will merge
 	if !s.config.NoEnvoy {
-		if envoy, _, err = s.scrape(fmt.Sprintf("http://localhost:%d/stats/prometheus", s.envoyStatsPort), r.Header); err != nil {
+		if envoy, envoyCancel, _, err = s.scrape(fmt.Sprintf("http://localhost:%d/stats/prometheus", s.envoyStatsPort), r.Header); err != nil {
 			log.Errorf("failed scraping envoy metrics: %v", err)
 			metrics.EnvoyScrapeErrors.Increment()
 		}
-		// Process envoy's metrics to make them compatible with FmtOpenMetrics
```
howardjohn, line 493: nit: we lost this commentary. Can we explain *why* we are doing processing. Even better if there is a link to some the  prom/openmetrics spec describing this, etc. Understand if that isn't possible though

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -584,22 +651,21 @@ func getHeaderTimeout(timeout string) (time.Duration, error) {
...
+	var cancel context.CancelFunc
 	ctx := context.Background()
 	if timeoutString := header.Get("X-Prometheus-Scrape-Timeout-Seconds"); timeoutString != "" {
 		timeout, err := getHeaderTimeout(timeoutString)
 		if err != nil {
 			log.Warnf("Failed to parse timeout header %v: %v", timeoutString, err)
 		} else {
-			c, cancel := context.WithTimeout(ctx, timeout)
-			ctx = c
-			defer cancel()
```
howardjohn, line 597: one thing that isn't clear to me - why do we cancel outside of scrape now instead of inside?

If we do need to, since its not obvious can we explain why (in code comment)?

howardjohn: does `cancel()` mean the `res.Body` will be also be closed (before we read it)?

howardjohn: The unit tests seem to pass if we keep the `defer cancel()`, so maybe its possible or we don't have enough coverage.

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -542,24 +571,62 @@ func negotiateMetricsFormat(contentType string) expfmt.Format {
...
 }
 
-func processMetrics(metrics []byte) []byte {
-	return bytes.ReplaceAll(metrics, []byte("\n\n"), []byte("\n"))
+func copyAndProcessMetrics(dst io.Writer, src io.Reader, buf []byte) (written int64, err error) {
+	var sideBreak bool
+	for {
+		nr, er := src.Read(buf)
+		if nr > 0 {
+			rbuf := bytes.ReplaceAll(buf[0:nr], []byte("\n\n"), []byte("\n"))
```
howardjohn, line 579: So essentially we are trying to copy everything from src to dst, utilizing `buf` along the way, and replacing `\n\n` with `\n`, yes?

A couple concerns here: 
* Its pretty complex. https://pkg.go.dev/strings#NewReplacer is something *really* close to this, but operates on strings. I wonder if we can somehow reuse some of that for our puroses?
* Its error prone. For this type of low level parsing stuff I think we absolutely must have fuzzing. See https://github.com/istio/istio/blob/master/pkg/fuzz/README.md for more info.


On the fuzzing in particular, I wrote up a first draft since I wanted to check some of my assumptions:

```go

func FuzzWrite(f *testing.F) {
	f.Fuzz(func(t *testing.T, d []byte, size uint8) {
		if size == 0 {
			return
		}
		if bytes.Contains(d, []byte("\n\n\n")) {
			// Doesn't work yet
			return
		}
		if bytes.HasSuffix(d, []byte("\n")) {
			// Doesn't work yet
			return
		}
		t.Logf("input: %v", d)
		buf := make([]byte, size)
		t.Logf("buffer size: %v/%v", len(buf), size)
		writer := &bytes.Buffer{}
		reader := bytes.NewReader(d)
		w, err := copyAndProcessMetrics(writer, reader, buf)
		if err != nil {
			t.Fatalf("got err: %v", err)
		}
		t.Logf("output: %v", writer.Bytes())

		if int(w) != (len(d) - bytes.Count(d, []byte("\n\n"))) {
			t.Fatalf("didn't write all: %v vs %v", w, len(d))
		}
		if bytes.Contains(writer.Bytes(), []byte("\n\n")) {
			t.Fatalf("contains double newline: %v", writer.String())
		}
	})
}
```

This exposes a few different issues, in particular a crash in some cases, and issues where the `\n\n` is split across the `buf` boundary (ex: buf size is 100, we have `\n\n` at index 100 and 101).

howardjohn (changes requested):
Benchmark results look great. I think largely this makes sense but needs some improvements. Thanks for the work here

pilot/cmd/pilot-agent/status/server.go
Patrick0308: Blank lines are not mentioned on  [OpenMetrics spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md). So they're not permitted .  See https://github.com/OpenObservability/OpenMetrics/issues/250. 
[Prom spec](https://github.com/prometheus/docs/blob/512f28293efc7dba62c13dc6e9ea5247c39f27d4/content/docs/instrumenting/exposition_formats.md?plain=1#L45) are mentioned that they are be ignored. 

Actually I working on removing empty line on envoy through https://github.com/envoyproxy/envoy/pull/22403 .  But I think we should replace the `\n\n` due to compatible with old envoy version.

howardjohn: istio-agent is tightly coupled with envoy versions (bundled in same.docker image) so I would actually be fine with just fixing in envoy and making this code simple

pilot/cmd/pilot-agent/status/server.go
Patrick0308: Reading `res.Body()` after `cancel()`,  it flakes. Maybe success when body's size is small.
See discussion on https://groups.google.com/g/golang-nuts/c/2FKwG6oEvos

Patrick0308: We should add a unit test to cover big response body.

Patrick0308:
https://github.com/envoyproxy/envoy/pull/22403 is be merged. We're waiting for it be released.

howardjohn:
It will automatically get pulled into istio/proxy and then istio/istio in a
day or 2. We don't wait for tagged releases

hzxuzhonghu:
great work

hzxuzhonghu:
We can then use io.Copy

Patrick0308:
The latest comparison of the after and before benchmarks:
```
name                               old time/op    new time/op    delta
Stats/stats-fmttext-1kb-8             935µs ± 2%     964µs ± 3%     ~     (p=0.056 n=5+5)
Stats/stats-fmtopenmetrics-1kb-8      984µs ± 6%     947µs ± 3%     ~     (p=0.056 n=5+5)
Stats/stats-fmttext-1mb-8            5.75ms ± 4%    2.99ms ± 3%  -48.06%  (p=0.008 n=5+5)
Stats/stats-fmtopenmetrics-1mb-8     5.68ms ± 1%    3.01ms ± 6%  -46.97%  (p=0.008 n=5+5)
Stats/stats-fmttext-10mb-8           52.8ms ± 3%    17.1ms ± 2%  -67.65%  (p=0.008 n=5+5)
Stats/stats-fmtopenmetrics-10mb-8    52.8ms ± 2%    17.1ms ± 4%  -67.57%  (p=0.008 n=5+5)

name                               old alloc/op   new alloc/op   delta
Stats/stats-fmttext-1kb-8             206kB ± 0%     178kB ± 0%  -13.43%  (p=0.008 n=5+5)
Stats/stats-fmtopenmetrics-1kb-8      206kB ± 0%     179kB ± 0%  -13.45%  (p=0.008 n=5+5)
Stats/stats-fmttext-1mb-8            8.65MB ± 0%    3.89MB ± 0%  -54.97%  (p=0.008 n=5+5)
Stats/stats-fmtopenmetrics-1mb-8     8.65MB ± 0%    3.89MB ± 0%  -54.96%  (p=0.008 n=5+5)
Stats/stats-fmttext-10mb-8           84.0MB ± 0%    33.1MB ± 0%  -60.54%  (p=0.008 n=5+5)
Stats/stats-fmtopenmetrics-10mb-8    84.0MB ± 0%    33.1MB ± 0%  -60.54%  (p=0.008 n=5+5)

name                               old allocs/op  new allocs/op  delta
Stats/stats-fmttext-1kb-8             1.26k ± 0%     1.25k ± 0%   -0.97%  (p=0.008 n=5+5)
Stats/stats-fmtopenmetrics-1kb-8      1.26k ± 0%     1.25k ± 0%   -0.98%  (p=0.008 n=5+5)
Stats/stats-fmttext-1mb-8             1.33k ± 0%     1.29k ± 0%   -3.16%  (p=0.008 n=5+5)
Stats/stats-fmtopenmetrics-1mb-8      1.33k ± 0%     1.29k ± 0%   -3.14%  (p=0.008 n=5+5)
Stats/stats-fmttext-10mb-8            1.35k ± 0%     1.31k ± 0%   -3.56%  (p=0.008 n=5+5)
Stats/stats-fmtopenmetrics-10mb-8     1.35k ± 0%     1.31k ± 0%   -3.50%  (p=0.008 n=5+5)
```

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -508,29 +527,33 @@ func (s *Server) handleStats(w http.ResponseWriter, r *http.Request) {
...
 		metrics.AgentScrapeErrors.Increment()
 	}
+
+	buf := bufPool.Get().(*[]byte)
+	defer bufPool.Put(buf)
+
 	if envoy != nil {
-		if _, err := w.Write(envoy); err != nil {
-			log.Errorf("failed to write envoy metrics: %v", err)
+		_, err = io.CopyBuffer(w, envoy, *buf)
```
howardjohn, line 542: I believe this is not needed - we can do without buffer entirely.

`w` is a `*http.response` which implements `io.ReaderFrom`. In CopyBuffer, there is an optimatization for ReaderFrom which ignores the `buf` entirely (because it uses its own buffer pool under the hood).

As a result, I think this probably does nothing except in unit tests/benchmarks (where w does *not* implement ReaderFrom).

I think in real world we can drop the bufpool and simplify this a bit by using `io.Copy`

Patrick0308: > w is a *http.response which implements io.ReaderFrom.

Thanks. I just found this out.

pilot/cmd/pilot-agent/status/server.go
```diff
@@ -584,22 +593,21 @@ func getHeaderTimeout(timeout string) (time.Duration, error) {
 // scrape will send a request to the provided url to scrape metrics from
 // This will attempt to mimic some of Prometheus functionality by passing some of the headers through
 // such as accept, timeout, and user agent
-// Returns the scraped metrics as well as the response's "Content-Type" header to determine the metrics format
-func (s *Server) scrape(url string, header http.Header) ([]byte, string, error) {
+// Returns the scraped metrics reader as well as the response's "Content-Type" header to determine the metrics format
+func (s *Server) scrape(url string, header http.Header) (io.ReadCloser, context.CancelFunc, string, error) {
+	var cancel context.CancelFunc
```
hzxuzhonghu, line 598: can you call cancel in this function

Patrick0308: Can't. See https://github.com/istio/istio/pull/40109#discussion_r931741061

hzxuzhonghu: Thanks

hzxuzhonghu (approved):

howardjohn (approved):
Thanks!!

Issue #35398, opened 2021-09-28
Global rate limit Failed and filter_chain_not_found ERROR after enable istio-injection and MTLS


shingshing222:
### Bug Description

Steps to reproduce the error:

1. Follow the steps of installation of istio and sample application (Bookinfo) https://istio.io/latest/docs/setup/getting-started/

2. Follow the steps of enabling Rate Limits https://istio.io/latest/docs/tasks/policy-enforcement/rate-limit/

3. Access the bookinfo application (/productpage), it returns 200 OK response and check sidecar log of Rate Limits:
istio-proxy [2021-09-28T14:55:10.730Z] "POST /envoy.service.ratelimit.v3.RateLimitService/ShouldRateLimit HTTP/2" 200 - via_upstream - "-" 52 23 2 1 "172.17.0.9" "-" "c67ff961-6386-9e13-86a0-583996663589"

4. Enable MTLS by applying https://istio.io/latest/docs/reference/config/security/peer_authentication/:
apiVersion: security.istio.io/v1beta1
kind: PeerAuthentication
metadata:
name: default
namespace: istio-system
spec:
mtls:
mode: STRICT

5. Access the bookinfo application (/productpage) again and it returns 500 error and also filter_chain_not_found error is found in the sidecar log of Rate Limits:
istio-proxy [2021-09-28T14:55:12.134Z] "- - -" 0 NR filter_chain_not_found - "-" 0 0 0 - "-" "-" "-" "-" "-" - - 172.17.0.8:8081 172.17.0.1:38748 - -

### Version

```prose
istioctl version
Client version: 1.10.2
control plane version: 1.10.2
data plane version: 1.10.2 (10 proxies)

kubectl version --short
Client Version: v1.21.2
Server Version: v1.20.7
```

bianpengyuan:
How do you send request to bookinfo service?

shingshing222:
I sent the request by using the browser visiting http://127.0.0.1/productpage
I think it is nothing to do with bookinfo service. It is just a sample application to allow the request received from istio ingress. 
I guess there should be an issue that envoyfilter sending traffic to another sidecar or receiving traffic from another sidecar when MTLS is enabled at mesh level

bianpengyuan:
Does the request go through an ingress? I think you get `filter_chain_not_found` because all filter chains expects istio mTLS, but request coming in is not encrypted (i.e. not going through an ingressgateway in your case)

hschanhkjc:
[...]

PR #40114, opened 2022-07-26
remove redundant todo

zirain:

cluster will send to Gateway when enable `PILOT_FILTER_GATEWAY_CLUSTER_CONFIG`

```console
root@ecs-zirain:~/go/src/istio.io/istio# istioctl pc c istio-ingressgateway-5768d85bb8-tgqfm -nistio-system  | grep rate_limit_cluster
rate_limit_cluster     -        -          -             STRICT_DNS   

root@ecs-zirain:~/go/src/istio.io/istio# k get deploy istiod -nistio-system -oyaml | grep PILOT_FILTER_GATEWAY_CLUSTER_CONFIG -A5
        - name: PILOT_FILTER_GATEWAY_CLUSTER_CONFIG
          value: "true"
        - name: REVISION
          value: default
        - name: JWT_POLICY
          value: third-party-jwt
```

```yaml
kind: EnvoyFilter
metadata:
  name: filter-ratelimit
  namespace: istio-system
spec:
  workloadSelector:
    # select by label in the same namespace
    labels:
      istio: ingressgateway
  configPatches:
    - applyTo: CLUSTER
      patch:
        operation: ADD
        # Adds the rate limit service cluster for rate limit service defined in step 1.
        value:
          name: rate_limit_cluster
          type: STRICT_DNS
          connect_timeout: 10s
          lb_policy: ROUND_ROBIN
          http2_protocol_options: {}
          load_assignment:
            cluster_name: rate_limit_cluster
            endpoints:
              - lb_endpoints:
                  - endpoint:
                      address:
                        socket_address:
                          address: ratelimit.default.svc.cluster.local
                          port_value: 8081

```

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -785,7 +785,6 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 }
 
 // add services from MeshConfig.ExtensionProviders
-// TODO: include cluster from EnvoyFilter such as global ratelimit [demo](https://istio.io/latest/docs/tasks/policy-enforcement/rate-limit/#global-rate-limit)
 func getHostsFromMeshConfig(ps *PushContext) sets.Set {
 	hostsFromMeshConfig := sets.New()
 


pilot/pkg/model/push_context.go
```diff
@@ -785,7 +785,6 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 }
 
 // add services from MeshConfig.ExtensionProviders
-// TODO: include cluster from EnvoyFilter such as global ratelimit [demo](https://istio.io/latest/docs/tasks/policy-enforcement/rate-limit/#global-rate-limit)
```
howardjohn, line 788: It does when _directly_ adding the cluster via EnvoyFilter ,but not if you just point to a cluster created by a Service I think.

zirain: do you have an example for that?

howardjohn: Sure, in the linked rate limit  example just replace `cluster_name: rate_limit_cluster` with something like `cluster_name: outbound|80||ratelimit.ns.svc.cluster.local`. IMO this is actually a better pattern because it allows using standard Istio functionality like mtls, etc easily

zirain: > Sure, in the linked rate limit example just replace `cluster_name: rate_limit_cluster` with something like `cluster_name: outbound|80||ratelimit.ns.svc.cluster.local`. IMO this is actually a better pattern because it allows using standard Istio functionality like mtls, etc easily

this remind me about something I forgot in https://github.com/istio/istio/issues/35398#issuecomment-1184664572

I will send a PR to fix the doc, let's keep this TODO.

PR #40115, opened 2022-07-26
Set istiod seccompProfile to RuntimeDefault

### LINKED ISSUE #39791: Set the Seccomp profile

It is good practice to [Restrict a Container's Syscalls with seccomp](https://kubernetes.io/docs/tutorials/security/seccomp/). To do this one sets the `seccompProfile` field on the Pod's `securityContext`. Most container runtimes provide a sane set of default syscalls that are allowed or not, and these can be called using `RuntimeDefault`. 

I'd propose that Istio should set the `RuntimeDefault` for all pods it creates (`istiod` being the one I'm especially interested in).
It should also allow over-riding the `securityContext` of every pod and container it creates.

You could just make the `securityContext`s configurable.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Extensions and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

hottea773:

Fixes https://github.com/istio/istio/issues/39791. This is a WIP PR because it takes the naivest approach. I suspect that this may not get merged, but I want to see.


howardjohn (commented):
2 main points

1. Should we do it for everything, not just istiod?
2. I think we need to make this configurable and possibly off by default. I have a major fear some runtime will have bad defaults that break us.

hottea773:
@howardjohn thanks:

1. That probably makes sense, I'll do that once I know what the approach should look like.  
2. Probably, yes. A couple question before I make any changes here:
- Do you think it would make sense to do this by making the whole `securityContext` configurable? Some other projects do this; for example https://github.com/prometheus-community/helm-charts/blob/main/charts/alertmanager/values.yaml#L49-L56. It has the benefit of meaning that users can do such things in future without needing more PRs, but has the usual set of problems associated with extra config (more maintenance, more scope for mis-configuration etc). 
- Alternatively, could we add a `merge` field, which does the same thing, but with slightly different semantics? That is what we do currently with `istio-proxy` using https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/values.yaml#L97-L108
- Otherwise, I'd just add a new field `Values.pilot.seccompProfile` which won't be set by default but will have a commented-out value of `type: RuntimeDefault` as done elsewhere.

hottea773:
For now I've done the simple change to make the seccomp profile configurable, would this be sensible? (If copied across every pod type?)

hottea773:
@howardjohn any opinions here?

hottea773:
@howardjohn I've got the existing UTs passing, A couple more questions before I can fix up the release note and remove WIP etc?

- Are you generally happy with this change? (If you are I will go ahead and make equivalent changes to the other containers)
- Does this change need new tests written? (If it does, would you be able to give me a hint as to where would be sensible to write them?)

howardjohn (approved):

hottea773:
@costinm @jacob-delgado @Monkeyanator would you be able review/approve please?

(Sorry if this isn't the right way to try and get approval.)

ericvn:
I notice there are some additional changes post @howardjohn 's approval and if I simply approve the release notes this will merge. Waiting to see if John has thoughts on new changes.

ericvn (approved):
Release Notes are good.

jacob-delgado (approved):

PR #40136, opened 2022-07-27
Set ISTIO_META_GENERATOR only fro grpc

costinm:
Small adjustment to proxyless grpc: we should be able to run applications that use both grpc and TCP/HTTP,
with envoy used for the TCP/HTTP traffic and grpc proxyless. 

We want the grpc bootstrap to have ISTIO_META_GENERATOR, so grpc gets expected config - without having
to specify it in the env variables explicitly. 

This is the first step - there are few other adjustments to the injection template, mixed mode will need to use 
the regular injection.


pkg/istio-agent/agent.go
```diff
@@ -613,6 +613,10 @@ func (a *Agent) generateGRPCBootstrap(credentialSocketExists bool) error {
 		return fmt.Errorf("failed generating node metadata: %v", err)
 	}
 
+	// GRPC bootstrap requires this. Original implementation injected this via env variable, but
```
howardjohn, line 616: do we need to remove the injection part?

pkg/istio-agent/agent.go
```diff
@@ -613,6 +613,10 @@ func (a *Agent) generateGRPCBootstrap(credentialSocketExists bool) error {
...
+	// this interfere with envoy, we should be able to use both envoy for TCP/HTTP and proxyless.
+	node.Metadata.Generator = "grpc"
```
howardjohn, line 618: do we want to only set it if unset? I don't know why, but hypothetically someone might want to override it?

howardjohn (commented):
Makes sense to me, cc @stevenctl

stevenctl:
SGTM, but I guess it does make sense to guard to only `!= ""`

costinm:
I don't think it's possible to override it - setting the env variable is
possible and will impact envoy behavior.

The problem is that in a mixed app ( TCP + proxyless gRPC ) we need to have
both envoy and proxyless
bootstrap files, and only one setting.

Also don't think we'll have another generator for grpc - if we do, we'll
add a new setting or change it.

Regarding injection template - I'm thinking it would be better to just
remove the entire template, there are few
other divergences ( like no lifecycle override ). Still evaluating - main
issue is how to cleanly bypass
interception for outbound proxyless gRPC, and have a nice UX not involving
iptables annotations.

stevenctl (approved):

hzxuzhonghu (approved):

technicianted:
@costinm is there an issue/project tracking mixed mode?

anuragagarwal561994:
@howardjohn when this fix can be released, this is crucial to our istio setup as we are migrating all the services to GRPC proxyless.

Also can it back ported to 1.13 and 1.14 or are there any workarounds for the these versions?

Issue #2437, opened 2018-01-05
Broken link in pilot/doc/design.md referring to configuration-flow.md


al3xandru:
`pilot/doc/design/md` includes a link to `configuration-flow.md` which looks to have been removed in the commit 238eec2ffc674bef08c02acc43c5989f09c60c4f#diff-7ededb7c32142221a5c20ed5f5f64eb4

christian-posta:
Yah seems related to this https://github.com/istio/istio/commit/238eec2ffc674bef08c02acc43c5989f09c60c4f#diff-0bf1e3e19574963fdf511247da316b61

I've sent a PR to fix

rshriram:
fixed

PR #40137, opened 2022-07-27
Use absolute path when adding file watcher

nirvanagit:
Fixes: https://github.com/istio/istio/issues/39985

Given there exists a DestinationRule with mode: MUTUAL, and paths for the private key, certificate and ca certificate of the form:
```
kind: DestinationRule
metadata:
  name: bad-dr
  namespace: foobar
spec:
  exportTo:
  - .
  host: service-mesh-debug
  trafficPolicy:
    tls:
      caCertificates: /etc/certs/root-cert.pem
      clientCertificate: /etc/certs/cert-chain.pem
      mode: MUTUAL
      privateKey: /etc/certs/key.pem
```

When the listeners are configured to listen on `etc/certs/cert-chain.pem` (notice there is no slash `/` before etc), then there are two sets of file watch add events:
1. One for files with path starting from `etc/certs`
2. and another for files with path starting from `/etc/certs`

When these files are removed, and re-added as part of certificate rotation, the operating system will only send one notification event, and that will be responsible for doing an SDS push. But given that the paths are different, and there is only one inotify event, only one set of SDS pushes will happen. Which means that this condition - https://github.com/istio/istio/blob/9a896c99db0ee9300651a08fb5cf410da7026bc9/security/pkg/nodeagent/cache/secretcache.go#L684 will only be met once. Resulting in just one of the sets (either one starting with `/` or one starting without `/`) of certificates being pushed to envoy via SDS

### DIFF
--- a/security/pkg/nodeagent/cache/secretcache.go
+++ b/security/pkg/nodeagent/cache/secretcache.go
@@ -343,7 +343,11 @@ func (sc *SecretManagerClient) tryAddFileWatcher(file string, resourceName strin
 	// avoid processing duplicate events for the same file.
 	sc.certMutex.Lock()
 	defer sc.certMutex.Unlock()
-	file = filepath.Clean(file)
+	file, err := filepath.Abs(file)
+	if err != nil {
+		cacheLog.Errorf("%v: error finding absolute path of %s, retrying watches [%s] %v", resourceName, file, err)
+		return err
+	}
 	key := FileCert{
 		ResourceName: resourceName,
 		Filename:     file,
--- a/security/pkg/nodeagent/cache/secretcache_test.go
+++ b/security/pkg/nodeagent/cache/secretcache_test.go
@@ -692,3 +692,54 @@ func TestOSCACertGenerateSecretEmpty(t *testing.T) {
 		t.Fatal("Certs did match")
 	}
 }
+
+func TestTryAddFileWatcher(t *testing.T) {
+	var (
+		dummyResourceName = "default"
+		relativeFilePath  = "./testdata/file-to-watch.txt"
+	)
+	absFilePathOfRelativeFilePath, err := filepath.Abs(relativeFilePath)
+	if err != nil {
+		t.Fatalf("unable to get absolute path to file %s, err: %v", relativeFilePath, err)
+	}
+	fakeCACli, err := mock.NewMockCAClient(time.Hour, true)
+	if err != nil {
+		t.Fatalf("unable to create fake mock ca client: %v", err)
+	}
+	sc := createCache(t, fakeCACli, func(resourceName string) {}, security.Options{WorkloadRSAKeySize: 2048})
+	cases := []struct {
+		name               string
+		filePath           string
+		expFilePathToWatch string
+		expErr             error
+	}{
+		{
+			name: "Given a file is expected to be watched, " +
+				"When tryAddFileWatcher is invoked, with file path which does not start with /" +
+				"Then tryAddFileWatcher should watch on the absolute path",
+			filePath:           relativeFilePath,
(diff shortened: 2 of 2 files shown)


ramaraochavali:
why is clean not happening for DR file paths?

nirvanagit:
> why is clean not happening for DR file paths? Is it because the default paths used for ROOTCA etc have `./etc'
> 
> https://github.com/istio/istio/blob/6f43dc2c0f9595a3c1e6d49d27fe709f508129e1/pkg/security/security.go#L40
> 
> ? If yes, can you try in your giving same path without this fix and see if the problem goes away?

Clean will remove `./` from the path, which will cause it to be different than what is specified in the DR, and cause this issue. Reference: https://go.dev/play/p/fVx6ahaT80d

nirvanagit:
@howardjohn @ramaraochavali - should we do the solutions I have implemented here or should we remove `.` from `./` in definitions listed here: https://github.com/istio/istio/blob/f15e4a14798b987d3afe0787d14de09f860c58c3/pkg/security/security.go#L40-L52

ramaraochavali:
I think ./ was added to support local dir testing. This fix seems good to me.

ramaraochavali:
LGTM. Will leave it for @howardjohn to review and approve.

howardjohn (commented):
unit test failure seems related

nirvanagit:
@howardjohn - it is failing on
```
HEAD is now at bb854084 Automator: update common-files@master in istio/api@master (#2433)
buf generate --config tools/proto/buf.yaml --path operator/pkg/ --output operator  --template tools/proto/buf.golang.yaml
buf generate --config tools/proto/buf.yaml --path pkg/dns/ --output pkg  --template tools/proto/buf.golang.yaml
Failure: plugin go-grpc: exit status 127
make: *** [tools/proto/proto.mk:23: dns-proto] Error 1
```
any recommendations on how to get it working? The PR does not have changes related to the failure.

nirvanagit:
@howardjohn - thanks for re-triggering the tests.

All checks passed, except one which is Pending due to no approvals.

howardjohn (approved):

nirvanagit:
> @nirvanagit: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> gencheck_istio	[99aae10](https://github.com/istio/istio/commit/99aae10f4c7c90b49358ee882c219ac77cc23b56)	[link](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/40137/gencheck_istio/1554227778646183936)	unknown	`/test gencheck_istio`

I think gencheck_istio ran only after the PR was approved. And it failed with the same reason:
```If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at e44fdbd4 Automator: update common-files@master in istio/api@master (#2437)
buf generate --config tools/proto/buf.yaml --path operator/pkg/ --output operator  --template tools/proto/buf.golang.yaml
buf generate --config tools/proto/buf.yaml --path pkg/dns/ --output pkg  --template tools/proto/buf.golang.yaml
buf generate --config tools/proto/buf.yaml --path pkg/test/echo --output pkg  --template tools/proto/buf.golang.yaml
Failure: plugin go-grpc: exit status 127
make: *** [tools/proto/proto.mk:26: echo-proto] Error 1```

SantiMunoz:
Hello, we are facing the same problem, in which Istio version should we expect this fix? in the current 1.15.2 is not included as far as I can see.
Thanks!
cc @nirvanagit @howardjohn

Issue #39925, opened 2022-07-14
TestEDSUnhealthyEndpoints flake


zirain:
### Bug Description

https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/39457/unit-tests_istio/1547382153052950528

https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/40114/unit-tests_istio/1551866083395244032

PR #40140, opened 2022-07-28
move send unhealthy endpoint flag to atomic bool

ramaraochavali:
Fixes https://github.com/istio/istio/issues/39925

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"time"
 
+	"go.uber.org/atomic"
 	"google.golang.org/protobuf/types/known/durationpb"
 
 	"istio.io/istio/pkg/config/constants"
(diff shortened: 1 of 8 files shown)


ramaraochavali:
@howardjohn PTAL if this makes sense as a pattern. If it does, I will move rest of the feature flag in separate PR.

pilot/pkg/features/pilot.go
```diff
@@ -104,13 +105,13 @@ var (
 			" EDS pushes may be delayed, but there will be fewer pushes. By default this is enabled",
 	).Get()
 
-	SendUnhealthyEndpoints = env.RegisterBoolVar(
+	SendUnhealthyEndpoints = atomic.NewBool(env.RegisterBoolVar(
```
howardjohn, line 108: I do wonder if we should change the `env` package -- possibly at the same time as making it use generics -- to make it so everything is atomic?

This problem is universal, just happens that one test triggers it most often

howardjohn (approved):

pilot/pkg/features/pilot.go
ramaraochavali: > I do wonder if we should change the env package -- possibly at the same time as making it use generics -- to make it so everything is atomic?

Yes.
Should I change others to also this model for now and update env package later when we have generics?

howardjohn:
Yeah lets wait

PR #40142, opened 2022-07-28
add support for maglev for consistent hash load balancer

ramaraochavali:
Implements https://github.com/istio/api/pull/2434
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali:
Benchmark results of Envoy
```
benchmarkRingHashLoadBalancerBuildRing/100/65536                  54.8 ms         54.8 ms            1 memory=0 memory_per_host=0
benchmarkRingHashLoadBalancerBuildRing/200/65536                  54.7 ms         54.7 ms            1 memory=0 memory_per_host=0
benchmarkRingHashLoadBalancerBuildRing/500/65536                  55.3 ms         55.3 ms            1 memory=0 memory_per_host=0
benchmarkRingHashLoadBalancerBuildRing/100/256000                  216 ms          216 ms            1 memory=0 memory_per_host=0
benchmarkRingHashLoadBalancerBuildRing/200/256000                  215 ms          215 ms            1 memory=0 memory_per_host=0
benchmarkRingHashLoadBalancerBuildRing/500/256000                  215 ms          215 ms            1 memory=0 memory_per_host=0
benchmarkMaglevLoadBalancerBuildTable/100                         10.5 ms         10.5 ms            1 memory=0 memory_per_host=0
benchmarkMaglevLoadBalancerBuildTable/200                         10.4 ms         10.4 ms            1 memory=0 memory_per_host=0
benchmarkMaglevLoadBalancerBuildTable/500                         10.9 ms         10.9 ms            1 memory=0 memory_per_host=0
benchmarkRingHashLoadBalancerChooseHost/100/65536/100000          97.1 ms         97.1 ms            1 mean_hits=1000 relative_stddev_hits=0.0449153 stddev_hits=44.9153
benchmarkRingHashLoadBalancerChooseHost/200/65536/100000          99.8 ms         99.8 ms            1 mean_hits=500 relative_stddev_hits=0.0739416 stddev_hits=36.9708
benchmarkRingHashLoadBalancerChooseHost/500/65536/100000           101 ms          101 ms            1 mean_hits=200 relative_stddev_hits=0.113084 stddev_hits=22.6168
benchmarkRingHashLoadBalancerChooseHost/100/256000/100000          105 ms          105 ms            1 mean_hits=1000 relative_stddev_hits=0.0368953 stddev_hits=36.8953
benchmarkRingHashLoadBalancerChooseHost/200/256000/100000          107 ms          107 ms            1 mean_hits=500 relative_stddev_hits=0.0506154 stddev_hits=25.3077
benchmarkRingHashLoadBalancerChooseHost/500/256000/100000          107 ms          107 ms            1 mean_hits=200 relative_stddev_hits=0.0827714 stddev_hits=16.5543
benchmarkMaglevLoadBalancerChooseHost/100/100000                  76.1 ms         76.1 ms            1 mean_hits=1000 relative_stddev_hits=0.029227 stddev_hits=29.227
benchmarkMaglevLoadBalancerChooseHost/200/100000                  79.2 ms         79.2 ms            1 mean_hits=500 relative_stddev_hits=0.0415841 stddev_hits=20.7921
benchmarkMaglevLoadBalancerChooseHost/500/100000                  81.1 ms         81.1 ms            1 mean_hits=200 relative_stddev_hits=0.0687299 stddev_hits=13.746
benchmarkRingHashLoadBalancerHostLoss/500/65536/1/10000            138 ms          138 ms            1 host_loss_over_N_optimal=0.2 percent_different=0.15
benchmarkRingHashLoadBalancerHostLoss/500/65536/2/10000            138 ms          138 ms            1 host_loss_over_N_optimal=0.4 percent_different=0.34
benchmarkRingHashLoadBalancerHostLoss/500/65536/3/10000            137 ms          137 ms            1 host_loss_over_N_optimal=0.6 percent_different=0.46
benchmarkMaglevLoadBalancerHostLoss/500/1/10000                   44.4 ms         44.4 ms            1 host_loss_over_N_optimal=0.2 percent_different=1
benchmarkMaglevLoadBalancerHostLoss/500/2/10000                   44.3 ms         44.3 ms            1 host_loss_over_N_optimal=0.4 percent_different=1.65
benchmarkMaglevLoadBalancerHostLoss/500/3/10000                   44.0 ms         44.0 ms            1 host_loss_over_N_optimal=0.6 percent_different=1.9
benchmarkMaglevLoadBalancerWeighted/500/5/1/1/10000               44.2 ms         44.2 ms            1 optimal_percent_different=0 percent_different=0
benchmarkMaglevLoadBalancerWeighted/500/5/1/127/1000              41.2 ms         41.2 ms            1 optimal_percent_different=81.9863 percent_different=80.8
benchmarkMaglevLoadBalancerWeighted/500/5/127/1/10000             46.5 ms         46.5 ms            1 optimal_percent_different=81.9863 percent_different=81.72
benchmarkMaglevLoadBalancerWeighted/500/50/1/127/1000             38.7 ms         38.7 ms            1 optimal_percent_different=49.2188 percent_different=50.8
benchmarkMaglevLoadBalancerWeighted/500/50/127/1/10000            44.1 ms         44.1 ms            1 optimal_percent_different=49.2188 percent_different=50.16
benchmarkMaglevLoadBalancerWeighted/500/95/1/127/1000             38.1 ms         38.1 ms            1 optimal_percent_different=4.95857 percent_different=6.5
benchmarkMaglevLoadBalancerWeighted/500/95/127/1/10000            45.2 ms         45.2 ms            1 optimal_percent_different=4.95857 percent_different=7.25
benchmarkMaglevLoadBalancerWeighted/500/95/25/75/1000             39.1 ms         39.1 ms            1 optimal_percent_different=0.139958 percent_different=0.8
benchmarkMaglevLoadBalancerWeighted/500/95/75/25/10000            44.0 ms         44.0 ms            1 optimal_percent_different=0.139958 percent_different=0.74
```
Also see minimal disruption explained  https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-maglevlbconfig and https://docs.google.com/document/d/1IU4b76AgOXijNa4sew1gfBfSiOMbZNiEt5Dhis8QpYg/edit#

releasenotes/notes/40142.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+
+releaseNotes:
+- |
+  **Updated** the load balancing algorithm used for consistent hash to MAGLEV inplance of RING_HASH.
+  This is expected to perform better with low host lookup times and also handles changes to backend
```
howardjohn, line 8: 'The downside of Maglev is that it is not as stable as ring hash.' seems to contradict that it handles changes better?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -783,19 +783,23 @@ func ApplyRingHashLoadBalancer(c *cluster.Cluster, lb *networking.LoadBalancerSe
...
-		},
+	if features.UseMaglevForConsistentHash {
+		// Maglev performs consistently better when consistent hashing is needed.
+		c.LbPolicy = cluster.Cluster_MAGLEV
+	} else {
+		// TODO MinimumRingSize is an int, and zero could potentially be a valid value
+		// unable to distinguish between set and unset case currently.
+		// 1024 is the default value for envoy
+		minRingSize := &wrappers.UInt64Value{Value: 1024}
+		if consistentHash.MinimumRingSize != 0 {
```
howardjohn, line 794: Now this is ignored. We may want to warn if its set? Or possibly even switch to ring_hash if its set?

howardjohn (commented):
Seems reasonable to me but I don't know much about these features to know the possible impact

cc @hzxuzhonghu

releasenotes/notes/40142.yaml
ramaraochavali: It seems a doc issue. 
"With that said, for many applications including Redis, Maglev is very likely a superior drop in replacement for ring hash."
and the bench marks shows the diff and this https://docs.google.com/document/d/1IU4b76AgOXijNa4sew1gfBfSiOMbZNiEt5Dhis8QpYg/edit# explains better.

pilot/pkg/networking/core/v1alpha3/cluster.go
ramaraochavali: warning seems better and if user wants RING_HASH he can flip the flag. Should we deprecate this if we moved forward with this?

nmittler:
Is the issue that we don't allow the user to specify maglev today?

ramaraochavali:
> Is the issue that we don't allow the user to specify maglev today?

Correct. But consistent hash is what we allow customer to specify and RING_HASH or MAGLEV is an implementation detail (choice of algo) for consistent hash to work.

nmittler:
@ramaraochavali ah so it's a bit murky then. Maybe we should actually expand the enum to include both ring hash and maglev in order to allow the user to be more explicit. That would make the switching of the default for consistent hashing safer and would allow a user a way to opt out without changing istiod flags.

ramaraochavali:
@nmittler I was actually thinking that we should switch to MAGLEV by default and deprecate the ring size and add https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-maglevlbconfig. 

Its not in the existing enum actually if we want to expose in the API. We have two choices https://github.com/istio/api/blob/master/networking/v1alpha3/destination_rule.proto#L607 and if user choose ConsistentHash may be we can add algorithm as one of the options and default it to RING_HASH - Is that what you are suggesting as well?

If we are good with that approach - I can add that option in ConsistentHash to choose the Algo

nmittler:
> @nmittler I was actually thinking that we should switch to MAGLEV by default and deprecate the ring size and add https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-maglevlbconfig.

So I assume you're suggesting to deprecate ring size because the ring hash algorithm would also be deprecated? Is it deprecated in Envoy? If not, I'm not sure we should deprecated it either ... if a user wants to use it, we should let them. There may be valid reasons apart from performance (i.e. stability). I'd rather just expand the enumeration and add any other config that we think is needed.

@howardjohn WDYT?

howardjohn:
If we expand the enum we cannot really set a 'default' for consistent hash
I think? We would just make maglev a new option folks could opt into? That
feels like it might be a bit to low level for Istio APIs -  I don't think
most higher level APIs allow distinguishing between these.

One thing to note if we do change an API is that Envoy now has support for
cookie based sticky sessions, so if we do change the API we should probably
make it compatible with that. Maybe a "type" field under
https://istio.io/latest/docs/reference/config/networking/destination-rule/#LoadBalancerSettings-ConsistentHashLB
or something...

ramaraochavali:
If we change the API I think it is better to add new field in ConsistentHashLB that tells what to pick MAGLEV, RING_HASH - But I think it is an implementation detail of consistent hash. So not 100% sure about exposing that in API

Regarding stateful session, it may be better to add it as a separate field as it is slightly from consistent hash because 
- it only supports cookie
- And it is actually differs from consistent hash - there is no hashing involved here.

releasenotes/notes/40142.yaml
ramaraochavali: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-maglevlbconfig clearly states it is intended for minimal disruption. So doc inconsistency.

ramaraochavali:
@howardjohn @nmittler Thinking about this more the following makes more sense to me

-  Add a new config that controls maglev table size
```
    // The minimum number of virtual nodes to use for the hash
    // ring. Defaults to 1024. Larger ring sizes result in more granular
    // load distributions. If the number of hosts in the load balancing
    // pool is larger than the ring size, each host will be assigned a
    // single virtual node.
    uint64 minimum_ring_size = 4;

    // The table size for Maglev hashing. This helps in controlling the 
    // disruption when the backend hosts change.
    // Increasing the table size reduces the amount of disruption. 
    uint64 maglev_table_size = 5;
```
- Choose the algorithm based on what user sets. If he sets minimum_ring_size that means he might have tuned the ring hash and that is a hint for us to apply ring hash. If he sets maglev_table_size that means he has chosen maglev.
- By default if none of these parameters are set, we default to Maglev and we let users switch back to consistent hash for couple of releases and remove that flag.

WDYT?

hzxuzhonghu:
>More keys will move position when hosts are removed (simulations show approximately double the keys will move).

I would agree with @nmittler no one is the best in any case as the doc said.

ramaraochavali:
> Honestly, I think instead of an enum we should have used a oneof, which would allow us to select an algorithm and define its properties all in a single structure. Maybe we should switch to that and deprecate the old field?

I am fine with that. Should I change API like this?

hzxuzhonghu:
I do agree add this option in API and give clear docs about pros and cons.
Then user have the right to select which algorithm they want to use

howardjohn:
I am not sure about switching to a oneoff, but an explicit API sounds right

pilot/pkg/features/pilot.go
```diff
@@ -151,6 +151,12 @@ var (
 		"UseRemoteAddress sets useRemoteAddress to true for side car outbound listeners.",
 	).Get()
 
+	UseMaglevForConsistentHash = env.RegisterBoolVar(
```
hzxuzhonghu, line 154: can be removed

ramaraochavali: API changes are not merged yet to main repo. I will fix it after api is reflected here.

hzxuzhonghu: why is it still here since the API is merged

ramaraochavali: API changes have not been reflected to Istio/Istio. I am waiting for it to happen so that I can change this.  https://github.com/istio/istio/pull/40479/files is merged today. So I will rebase and fix it.

hzxuzhonghu: What change do u mean? Not https://github.com/istio/api/pull/2434?

ramaraochavali: That change only was not updated to main repo because it was failing lint validation for deprecated field in that PR. Automatic API did not happen. https://github.com/istio/istio/pull/40479/files manually updated the API.

hzxuzhonghu: So let's wait for it. I believe eventually we meant to implement it rather than this env flag.

ramaraochavali:
@hzxuzhonghu @howardjohn implemented the new API. PTAL.

pkg/config/validation/validation.go
```diff
@@ -1427,27 +1427,31 @@ func validateConnectionPool(settings *networking.ConnectionPoolSettings) (errs e
...
 			if httpCookie.Name == "" {
-				errs = appendErrors(errs, fmt.Errorf("name required for HttpCookie"))
+				errs = appendValidation(errs, fmt.Errorf("name required for HttpCookie"))
 			}
 			if httpCookie.Ttl == nil {
-				errs = appendErrors(errs, fmt.Errorf("ttl required for HttpCookie"))
+				errs = appendValidation(errs, fmt.Errorf("ttl required for HttpCookie"))
 			}
 		}
+		if consistentHash.MinimumRingSize != 0 {
```
howardjohn: can we error if both are set? Since the deprecated one would be ignored in this case

howardjohn (approved):

pkg/config/validation/validation.go
ramaraochavali: Sure I can add

PR #40151, opened 2022-07-28
start admin interface earlier

stevenctl:
In some scenarios, it would be very useful to call debug functions before the server is ready to debug startup issues.

### DIFF
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -231,7 +231,7 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
 	if err := s.initIstiodAdminServer(args, whc); err != nil {
 		return nil, fmt.Errorf("error initializing debug server: %v", err)
 	}
-	if err := s.serveHttp(); err != nil {
+	if err := s.serveHTTP(); err != nil {
 		return nil, fmt.Errorf("error serving http: %v", err)
 	}
 
@@ -1279,7 +1279,7 @@ func (s *Server) initStatusManager(_ *PilotArgs) {
 	})
 }
 
-func (s *Server) serveHttp() error {
+func (s *Server) serveHTTP() error {
 	// At this point we are ready - start Http Listener so that it can respond to readiness events.
 	httpListener, err := net.Listen("tcp", s.httpServer.Addr)
 	if err != nil {
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -199,6 +199,21 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
 		internalStop:            make(chan struct{}),
 		istiodCertBundleWatcher: keycertbundle.NewWatcher(),
 	}
+
+	// Used for readiness, monitoring and debug handlers.
+	var (
+		whMu sync.RWMutex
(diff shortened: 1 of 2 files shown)


pilot/pkg/bootstrap/server.go
```diff
@@ -199,6 +199,24 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
...
+			return wh.Config.RawTemplates
+		}
+		return map[string]string{}
+	}
+	if err := s.initIstiodAdminServer(args, whc); err != nil {
+		return nil, fmt.Errorf("error initializing debug server: %v", err)
+	}
+
 	// Apply custom initialization functions.
 	for _, fn := range initFuncs {
```
howardjohn, line 221: seems like we should at least apply these first -they could change the args?

pilot/pkg/bootstrap/server.go
```diff
@@ -199,6 +199,24 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
...
+	)
+	whc := func() map[string]string {
+		whMu.RLock()
+		defer whMu.RUnlock()
+		if wh != nil {
+			return wh.Config.RawTemplates
+		}
+		return map[string]string{}
+	}
+	if err := s.initIstiodAdminServer(args, whc); err != nil {
```
howardjohn, line 216: do we know that we won't get crashes, etc from all the endpoints? For example kubeClient being nil may be problematic ?

Might be worth testing that things work or fail gracefully if we were to sleep(infinity) right after this

howardjohn (commented):
overall lgtm

hzxuzhonghu:
could be useful, make sense to me

stevenctl:
had to actually move where we start serving on the http port.. this means we have to say no to readiness before that. also, tested every endpoint and guarded a few nil-pointer accesses. Even when I did hit them, they didn't bubble up and crash the app.

stevenctl:
cc @howardjohn  @shankgan

shankgan (approved):

pilot/pkg/xds/debug.go
```diff
@@ -1039,6 +1049,9 @@ func (s *DiscoveryServer) instancesz(w http.ResponseWriter, req *http.Request) {
 }
 
 func (s *DiscoveryServer) networkz(w http.ResponseWriter, req *http.Request) {
+	if s.Env == nil || s.Env.NetworkManager == nil {
```
hzxuzhonghu, line 1052: Would this introduce race for test

pilot/pkg/bootstrap/server.go
```diff
@@ -209,6 +224,17 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
...
 	prometheus.EnableHandlingTimeHistogram()
 
+	// make sure we have a readiness probe before serving HTTP to avoid marking ready too soon
+	s.addReadinessProbe("discovery", func() (bool, error) {
+		return s.XDSServer.IsServerReady(), nil
+	})
+	if err := s.initIstiodAdminServer(args, whc); err != nil {
+		return nil, fmt.Errorf("error initializing debug server: %v", err)
+	}
+	if err := s.serveHTTP(); err != nil {
```
hzxuzhonghu, line 234: Check http server is well shutdown in ut

howardjohn (approved):

PR #40156, opened 2022-07-28
refactor response checker

aryan16:
**Exposed DNSProxying method and created a separate method to check the response in the same cluster only**


howardjohn (approved):

ericvn (approved):
Thanks.

PR #40158, opened 2022-07-28
[release-1.13] Run update_deps.sh

ericvn:

Run update_deps.sh for 1.13. The proxy build isn't done so the test may fail until that proxy artifact is available.


lei-tang (approved):

docker/Dockerfile.distroless
```diff
@@ -1,5 +1,5 @@
 # prepare a distroless source context to copy files from
-FROM gcr.io/distroless/static-debian11@sha256:21d3f84a4f37c36199fd07ad5544dcafecc17776e3f3628baf9a57c8c0181b3f as distroless_source
+FROM gcr.io/distroless/static-debian11@sha256: as distroless_source
```
stewartbutler, line 2: This doesn't look right.

ericvn: Not sure why these are wrong, but running commands themselves so no changes required.

PR #40162, opened 2022-07-29
fix generate manifest duplicate services and endpoints

### LINKED ISSUE #40022: When using external profile install or generate manifest duplicate services and endpoints created.
### Bug Description

When use the following command to generate manifest file, duplicate istiod service and endpoint will be produced.

```
cat <<EOF | istioctl manifest generate -f -
kind: IstioOperator
spec:
  profile: external
  values:
    global:
      remotePilotAddress: 1.2.3.4
    istiodRemote:
      injectionPath: /inject/cluster/cluster2/net/network1
  components:
    base:
      enabled: true
EOF
```

The following services will be duplicated in the output.

```
apiVersion: v1
kind: Endpoints
metadata:
  name: istiod
  namespace: istio-system
subsets:
- addresses:
  - ip: 1.2.3.4
  ports:
  - port: 15012
    name: tcp-istiod
    protocol: TCP
---
apiVersion: v1
kind: Service
metadata:
  name: istiod
  namespace: istio-system
spec:
  ports:
  - port: 15012
    name: tcp-istiod
    protocol: TCP
  # if the remotePilotAddress is IP addr, we use clusterIP: None.
  # else, we use externalName
  clusterIP: None
```

using k8s overlay can only modify one instance.

### Version

```prose
This error exists in the following releases tested.
1.13.x
1.14.x
1.15 (master local build)
```

### Additional Information

_No response_

xichengliudui:

fix: https://github.com/istio/istio/issues/40022
Generate the pilot remote service and endpoint using the Base component


litong01:
@xichengliudui I am not really sure that the fix will be right. With your fixes, when only external profile is used and base is not enabled, then we will have no service or endpoint, that will actually cause errors. I think that the right fixes is check if base is enabled, if it does, then remote should not have that, if it does not, then remote should have service and endpoints. That is the easy part, but this also causes issue when one wants to overlay his or her changes, it gets a bit complex in this case. I am not sure it can be fixed easily.

xichengliudui:
> @xichengliudui I am not really sure that the fix will do what we want. With your fixes, when external profile is used and base is not enabled, then we will have no istiod service or endpoint, that will actually cause errors. I think that the right fixes is check if base is enabled, if it is, then remote should not have the service and endpoint since base will have that, if it is not, then remote should have service and endpoints. That is the easy part, but this also causes issue when one wants to overlay his or her changes, it gets a bit complex in this case. I am not sure it can be fixed easily.

That's what I thought, i am updating and testing locally
Because I am not familiar with this part of the source code so the progress is a little slow

litong01:
@xichengliudui any plan to enhance this PR? I think you are going the right direction, just need to have some logic in the template based on if base or remote configuration enabled. Also there are a lot of errors in the tests which also need to be addressed.

frankbu:
You can't delete these files, they are copied from base during the build. Also, this is not bug anyway, it is working as designed. See comment here: https://github.com/istio/istio/issues/40022#issuecomment-1215647342

PR #40163, opened 2022-07-29
fix configuration to match ipv6 address

caiqi1111:

### DIFF
--- a/samples/addons/prometheus.yaml
+++ b/samples/addons/prometheus.yaml
@@ -121,14 +121,7 @@ data:
         - __meta_kubernetes_service_annotation_prometheus_io_path
         target_label: __metrics_path__
       - action: replace
-        regex: ([^:]+)(?::\d+)?;(\d+)
-        replacement: $1:$2
-        source_labels:
-        - __address__
-        - __meta_kubernetes_service_annotation_prometheus_io_port
-        target_label: __address__
-      - action: replace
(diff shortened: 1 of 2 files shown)


zirain:
cc @zhlsunshine would you please help to confirm this?

samples/addons/prometheus.yaml
```diff
@@ -127,6 +127,13 @@ data:
         - __address__
         - __meta_kubernetes_service_annotation_prometheus_io_port
         target_label: __address__
+      - action: replace
```
howardjohn, line 130: This file is generated from Prometheus helm chart (manifests/addons/values-prometheus.yaml). Can we fix it upstream? Or if it is, just update the version we use

caiqi1111: Prometheus Community has update the configuration to match ipv6 (https://github.com/prometheus-community/helm-charts/pull/2127/files). Maybe this file should have similar update?

howardjohn: The only accepted changes are in manifests/addons/gen.sh or manifests/addons/values-prometheus.yaml. The file being modified is generated from those

caiqi1111: Thanks. I understand how to change this file now. Prometheus helm-charts version should be 15.9.3 or higher to support IPV6.

PR #40164, opened 2022-07-29
Fix ist0103 msg incorrect displaying when injection label is set to false

hanxiaop:
```
Warning [IST0103] (Pod default/reviews-v2-5b5dc658c5-z4c9s) The pod default/reviews-v2-5b5dc658c5-z4c9s is missing the Istio proxy. This can often be resolved by restarting or redeploying the workload.
Error: Analyzers found issues when analyzing namespace: default.
```
```
kubectl get deploy reviews-v2 -oyaml | grep sidecar. -B2
      labels:
        app: reviews
        sidecar.istio.io/inject: "false"
```

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/pkg/config/analysis/analyzers/testdata/injection.yaml
+++ b/pkg/config/analysis/analyzers/testdata/injection.yaml
@@ -117,7 +117,7 @@ spec:
 apiVersion: v1
 kind: Pod
 metadata:
-  name: podinjectiondisabled-label
+  name: podinjectiondisabled-label-and-anno
   namespace: default
   annotations:
     sidecar.istio.io/inject: "true"
--- a/pkg/config/analysis/analyzers/testdata/injection.yaml
+++ b/pkg/config/analysis/analyzers/testdata/injection.yaml
@@ -99,3 +99,31 @@ spec:
   containers:
   - image: gcr.io/google-samples/microservices-demo/adservice:v0.1.1
     name: server
+---
+# Pod that explicitly disables injection with label. Should not generate a warning.
+apiVersion: v1
+kind: Pod
+metadata:
+  name: podinjectiondisabled-label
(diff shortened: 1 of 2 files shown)


pkg/config/analysis/analyzers/testdata/injection.yaml
```diff
@@ -99,3 +99,17 @@ spec:
...
     name: server
+---
+# Pod that explicitly disables injection with label. Should not generate a warning.
+apiVersion: v1
+kind: Pod
+metadata:
+  name: podinjectiondisabled-label
+  namespace: default
+  labels:
+    sidecar.istio.io/inject: "false"
```
GregHanson, line 110: do we support this as a label?  I thought `sidecar.istio.io/inject` was supported as an annotation ([doc](https://istio.io/latest/docs/reference/config/annotations/)).  In which case there is a test already covering the annotation on line 76

GregHanson (changes requested):

pkg/config/analysis/analyzers/testdata/injection.yaml
hanxiaop: @GregHanson See https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy. If both are set, we prefer the label.

GregHanson: @hanxiaop thanks for clarifying, I see the annotation and label logic in `/pkg/kube/inject/inject.go`.

GregHanson (approved):

PR #40165, opened 2022-07-29
When ServiceEntry are with the identical host name, try to `autoAllocateIPs` same IP

yannuil:
Signed-off-by: Yann Liu <yannliu@redhat.com>**Please provide a description of this PR:**

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


tests/integration/pilot/dnsautoallocation/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+var (
+	i istio.Instance
+
+	// Below are various preconfigured echo deployments. Whenever possible, tests should utilize these
+	// to avoid excessive creation/tear down of deployments. In general, a test should only deploy echo if
+	// its doing something unique to that specific test.
+	apps = deployment.SingleNamespaceView{}
+)
+
+func TestMain(m *testing.M) {
```
howardjohn, line 39: I appreciate the effort put into a test here. However, I think there may be some high cost vs value for this that may be better suited for a different type of test.

Every TestMain adds 1-2mins to each CI job. Which isn't that long, but its a death by 1000 cuts situation; 2 minutes for a single test doesn't provide a strong cost:value tradeoff in most cases.

It does seem like most of this could be handled by unit tests. The fix here is that "with autoallocation, two SEs without `address` should behave like two SE with `address`". We do already have integration tests for SE (although they could probably be improved -- I certainly wouldn't mind additional test cases their to existing suites, if you are interested), so we could just test the two are equivilent which could be done in unit tests.

In particular, tests in https://github.com/istio/istio/blob/0fb12bfcfe582458b70b5cb911fb0307ee927d3c/pilot/pkg/networking/core/v1alpha3/sidecar_simulation_test.go#L76 follow a pattern that provides what looks *a lot* like an integration tests, but is done through static xds analysis rather than full deployments; this may provide much of the same coverage and run ~instantly

howardjohn (commented):
cc @ramaraochavali @hzxuzhonghu

hzxuzhonghu:
I could not understand what is the caveat currently? In other words what's the problem now?

yannuil:
@howardjohn Thank you for your time to review this PR. I understand the cost of adding a new test suite. I tried to use the suite to create a reproducer. I will remove the test suite.

@hzxuzhonghu I will explain my findings.

In an istio setup with both `DNS_CAPTURE` and `DNS_AUTO_ALLOCATION` set,  two SEs with the same host name but different port are applied. [This function](https://github.com/istio/istio/blob/745c500b50b76429dabf56207d39552215e7d99e/pilot/pkg/serviceregistry/serviceentry/controller.go#L871) will try to give each SE a virtual IP. In [my test](https://github.com/istio/istio/blob/cea741c90743d81481c8db17e650af804aa4f37d/tests/integration/pilot/dnsautoallocation/mesh_external_service_entry_test.go#L68-L86), `240.240.0.1` and `240.240.0.2` are used. 

In a pod with the mesh, when a request is made to `istio.io`, which defined in the SEs, a DNS look up will be performed first to resolve the domain name to one of these two addresses. Depends on the order in which this "internal" DNS records get added in the `istio-agent`, the internal address resolved by the DNS is consistent. 

In [my test cases](https://github.com/istio/istio/blob/cea741c90743d81481c8db17e650af804aa4f37d/tests/integration/pilot/dnsautoallocation/mesh_external_service_entry_test.go#L39-L50), there are a 80 normal TCP port and 443 TLS port. Based on my test case, `240.240.0.1` is related to the mesh external cluster with a TCP port 80. It seems to be the "default" DNS record for domain `istio.io` in the mesh, a HTTPs request made to `https://istio.io` will fail to get routed by istio and try to "escape" the mesh. This will result a 500 upstream error from envoy.  My understanding is when envoy tries to find `240.240.0.1` with TLS port 443, it could not find a matching route. In the `ALLOW_ANY` case, the request will be let go into the wild.

The [sidecar_simulation_test](https://github.com/istio/istio/blob/0fb12bfcfe582458b70b5cb911fb0307ee927d3c/pilot/pkg/networking/core/v1alpha3/sidecar_simulation_test.go#L76) seems to be testing if the rules in envoy is doing correctly. In this case, the rules are doing what is supposed to.

What is the best way to test this?

hzxuzhonghu:
>two SEs with the same host name but different ports

Currently one address is allocated to this. Then what's the issue?

hzxuzhonghu (changes requested):
It is not clear, what it is fixing

yannuil:
@hzxuzhonghu Thank you for your time.

> two SEs with the same host name but different ports

Currently, there will be two virtual IPs assigned to these two SEs. It will cause an request to one of the SEs to be blocked. I try to show the problem with [this test set](https://github.com/istio/istio/pull/40165/commits/61c49a5f682edffbcef20d9681babdd724794c11). In the test set, one of the sub test will fail.

This PR will give a single virtual IP for these two SEs, then traffic will go through as expected.

rubrodrig:
Hi everyone,

I am also facing the problem that this PR is trying to resolve.

> Currently one address is allocated to this. Then what's the issue?

When I create two distinct `ServiceEntry` k8s objects :
1. First referencing `www.google.com:443` -> Istio VIP 240.240.0.1
2. Second referencing `www.google.com:80` -> Istio VIP 240.240.0.2

When I try to curl `https://www.google.com:443`, the VIP resolved is the one associated with the second ServiceEntry object (240.240.0.2).

Thanks you @yannuil for implementing a resolution.
Thank you @hzxuzhonghu for your time on this PR.

hzxuzhonghu:
I did not understand this issue well, will try to take a deep look

yannuil:
@hzxuzhonghu thank you for your time.

The existing code, with `DNS_AUTO_ALLOCATE` configured, allocates multiple virtual IPs for any given hostname, when the same hostname is referenced in multiple ServiceEntry resources.  When using `DNS_CAPTURE`, only one IP will be returned for the DNS request and this will cause issues with routing. This patch ensures that the same virtual IP is used for any given hostname, regardless of how many ServiceEntry resources reference the hostname.  In other words it ensures there is only ever one virtual IP defined for a hostname.

@howardjohn I have studied [the unit test example]([istio/pilot/pkg/networking/core/v1alpha3/sidecar_simulation_test.go](https://github.com/istio/istio/blob/0fb12bfcfe582458b70b5cb911fb0307ee927d3c/pilot/pkg/networking/core/v1alpha3/sidecar_simulation_test.go#L76)) you provided. If I understand it correctly, it tests the envoy's cluster definition with the istio behavior. I think the cause of the issue is mainly due to DNS. 

Do I need to add any release notes?

pilot/pkg/serviceregistry/serviceentry/controller.go
```diff
@@ -886,23 +888,42 @@ func autoAllocateIPs(services []*model.Service) []*model.Service {
...
+				thirdOctet := v[0]
+				fourthOctet := v[1]
+
+				svc.AutoAllocatedIPv4Address = fmt.Sprintf("240.240.%d.%d", thirdOctet, fourthOctet)
+
+				if thirdOctet == 0 {
+					svc.AutoAllocatedIPv6Address = fmt.Sprintf("2001:2::f0f0:%x", fourthOctet)
+				} else {
+					svc.AutoAllocatedIPv6Address = fmt.Sprintf("2001:2::f0f0:%x%x", thirdOctet, fourthOctet)
+				}
```
hzxuzhonghu, line 904: Can resue these logic

hzxuzhonghu (commented):
Generally lg

pilot/pkg/serviceregistry/serviceentry/controller.go
yannuil: Done

hzxuzhonghu (approved):
Thanks, awesome catch

hzxuzhonghu:
BTW, why remove the integration test?

yannuil:
I removed the integration test because John suggested me to explore a different type of test in [his review](https://github.com/istio/istio/pull/40165#discussion_r933345677). 

I could not figure out how to construct an unit test for this problem. I am happy to add the integration back in. 

@howardjohn could you give me some suggestions?

yannuil:
And for release note like [this one](https://github.com/istio/istio/blob/master/releasenotes/notes/xds-push-deadlock.yaml)?

hzxuzhonghu:
Yes, please add one

GregHanson (approved):
release note looks good

hzxuzhonghu:
@istio/release-managers

jwendell (approved):

GregHanson:
@ericvn any idea why the cherry-pick PR's weren't created for this one?

ericvn:
@GregHanson Sometimes there is an issue with cherry-picks. We had an issue for that at one point (would have to see if it's still open). I've also noticed that things sometime run much later, ie. I see multiple cherry-picks attempted. I wonder if some backend job gets stalled and then things start up later.

Tried kicking a CP using the command to see if that does something.

ericvn:
I did go back and look at the last page of commits on the main istio/istio branch and any PRs with the cherry-pick labels on them, did attempt to do the cherry-picks.

PR #40176, opened 2022-07-29
test: use `T.Setenv` to set env vars in tests

Juneezee:
This PR replaces `os.Setenv` with `t.Setenv`. Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

This saves us at least 2 lines (error check, and unsetting the env var) on every instance.


howardjohn (approved):

ericvn (approved):
Thanks. It looks like at least one case of not resetting that will be fixed.

PR #40186, opened 2022-07-30
Rework proxy labels update

hzxuzhonghu:

Fix: https://github.com/istio/istio/issues/40034 and https://github.com/istio/istio/issues/39792

1. dynamically update proxy labels when pod/wle labels update

2. Added localityLabel to pod label so we can get proxy locality from pod labels rather than service instances, which may not exist


pilot/pkg/model/context.go
```diff
@@ -849,14 +854,23 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
...
-// SetWorkloadLabels will set the node.Metadata.Labels only when it is nil.
+// SetWorkloadLabels will set the node.Metadata.Labels.
+// It merges both node meta labels and workload labels and give preference to workload labels.
+// Note: must be called after `SetServiceInstances`.
 func (node *Proxy) SetWorkloadLabels(env *Environment) {
-	// First get the workload labels from node meta
-	if len(node.Metadata.Labels) > 0 {
-		return
+	labels := env.GetProxyWorkloadLabels(node)
+	if len(labels) > 0 {
```
hzxuzhonghu, line 862: Not possible a pod has 0 labels in istio, at least some labels are injected by istio

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -1365,7 +1365,20 @@ func (c *Controller) getProxyServiceInstancesByPod(pod *v1.Pod,
...
+		locality := c.getPodLocality(pod)
+		if locality == "" {
+			return pod.Labels
+		}
+		out := make(map[string]string, len(pod.Labels)+1)
+		for k, v := range pod.Labels {
+			out[k] = v
+		}
+		// Add locality labels
+		out[model.LocalityLabel] = locality
```
hzxuzhonghu, line 1380: make deepcopy of pod labels

pilot/pkg/serviceregistry/kube/controller/pod.go
```diff
@@ -111,6 +112,19 @@ func GetPodConditionFromList(conditions []v1.PodCondition, conditionType v1.PodC
 	return -1, nil
 }
 
+func (pc *PodCache) labelFilter(old, cur interface{}) bool {
+	oldPod := old.(*v1.Pod)
+	curPod := cur.(*v1.Pod)
+
+	// If labels updated, trigger proxy push
+	if curPod.Status.PodIP != "" && !reflect.DeepEqual(oldPod.Labels, curPod.Labels) {
+		pc.proxyUpdates(curPod.Status.PodIP)
```
hzxuzhonghu, line 121: >Why is this a filter? It seems like it should be added to the queue and processed in onEvent?

This is a good question, i meant to do that. But when implementing, found in the queue, we do not pass the original object, so we can not compare old and new object labels

hzxuzhonghu: I think we can refactor the queue interface if more such cases occur

pkg/bootstrap/config.go
```diff
@@ -484,6 +484,7 @@ func extractAttributesMetadata(envVars []string, plat platform.Environment, meta
 		case "ISTIO_METAJSON_LABELS":
 			m := jsonStringToMap(val)
 			if len(m) > 0 {
+				meta.IstioMetaLabels = m
```
hzxuzhonghu, line 487: We need to distinguish which labels are from `ISTIO_METAJSON_LABELS` and which are from pod labels. 

This is easy to handle pod label removal. FYI: https://github.com/istio/istio/pull/40036#discussion_r926894456

hzxuzhonghu: I reworked this by inferring it.   `IstioMetaLabels `= meta.Labels minus  `GetProxyWorkloadLabels`

So do not need client to send duplicate labels

pilot/pkg/model/context.go
howardjohn: Not all pods are injected

hzxuzhonghu: The pod is originating the xds req, isn't it injected with a sidecar?

howardjohn: Not necessarily, could be a proxy less grpc. Or gateway, or istioctl kube-inject, or maybe custom injection template.

Tbh it's probably very rare edge case

hzxuzhonghu: For safety, we donot need to consider this labels = 0, in reality, no CICD will have no labels.

pilot/pkg/model/context.go
```diff
@@ -533,10 +537,11 @@ type NodeMetadata struct {
 	// Mostly used when istiod requests the upstream.
 	IstioRevision string `json:"ISTIO_REVISION,omitempty"`
 
-	// Labels specifies the set of workload instance (ex: k8s pod) labels associated with this node.
+	// Labels specifies the set of workload instance (ex: k8s pod) labels associated with this node. Labels is a
+	// superset of IstioMetaLabels.
 	Labels map[string]string `json:"LABELS,omitempty"`
```
howardjohn, line 542: IMO this feels a bit backwards. We have `Proxy{MetaLabels: map, NodeMeta: {Labels: map}`.

Shouldn't it be:

`Proxy{Labels: map, NodeMeta: {MetaLabels: map}`? Its confusing that compute labels are in "node metadata" but the labels from metadata are not?

pilot/pkg/model/context.go
howardjohn: should we just have a bool returned with the map to indicate if it was set or not? That makes things a bit more explicit, and should be a simple addition. Even though the edge case is rare it seems like it should be handled

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -1365,7 +1365,20 @@ func (c *Controller) getProxyServiceInstancesByPod(pod *v1.Pod,
...
+		}
+		locality := c.getPodLocality(pod)
+		if locality == "" {
+			return pod.Labels
+		}
+		out := make(map[string]string, len(pod.Labels)+1)
+		for k, v := range pod.Labels {
+			out[k] = v
+		}
+		// Add locality labels
```
howardjohn, line 1379: Can you comment *why* we do it, rather than what? Its not clear why we need to do it here

pilot/pkg/serviceregistry/kube/controller/pod.go
howardjohn: hmm.. I thought most core k8s controllers are written in this model (just get the name of the resource in the queue). How do they handle these types of things?

IIRC the reason we remove the old object is because it makes ordering dependency which breaks SyncAll?

howardjohn (changes requested):
Thanks for some of the improvements. I think this is a great change but we will need some integration test like described in the first PR to have confidence here. Since this is impacting the core AuthorizationPolicy (due to label selector) I think we have a fairly high bar here as incorrect labels will be a CVE

pilot/pkg/model/context.go
hzxuzhonghu: The reason is we largely depend on NodeMeta.Labels across the repo, should change in a separate pr

hzxuzhonghu:
@howardjohn ready for review, ptal

pilot/pkg/model/context.go
howardjohn: Ok, maybe we can merge that PR first and then do this. It seems very confusing to have it inverted

hzxuzhonghu: sure, i will work on refactor first

hzxuzhonghu:
cc @labaq @diranged

howardjohn (approved):

dddddai:
IIUC this only triggers a xDS push to the changed proxy. Shouldn't we also update endpoints labels and push EDS to all proxies?

hzxuzhonghu:
We intended to do xds update to the changed proxy, in case it uses locality routing. 

>Shouldn't we also update endpoints labels and push EDS to all proxies?

 it will be a great cost, and remeber we intended not to support it

PR #40194, opened 2022-08-01
minor fix to consume review comments for 39135

irisdingbj:


irisdingbj:
@costinm This is the follow-up PR to land your comments in #39135 , Please take a review, Thanks.

GregHanson (approved):

manifests/charts/istiod-remote/files/injection-template.yaml
```diff
@@ -371,8 +371,8 @@ spec:
     volumeMounts:
     - name: workload-socket
       mountPath: /var/run/secrets/workload-spiffe-uds
-    - name: credential-socket
-      mountPath: /var/run/secrets/credential-uds
+    - name: gateway-cert-socket
```
howardjohn, line 374: Why is it called "gateway"? This is a sidecar

irisdingbj: make sense, will `external-cert-socket` be more approriate?

PR #40209, opened 2022-08-01
security tests refactor

aryan16:
**Refactored few security tests to support new test framework, added support for customizable east-west gateway configuration**

### DIFF
--- a/pkg/test/framework/components/echo/common/deployment/echos.go
+++ b/pkg/test/framework/components/echo/common/deployment/echos.go
@@ -17,6 +17,7 @@ package deployment
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
@@ -63,7 +64,7 @@ type Config struct {
 	//
 	// Custom echo instances will be accessible from the `All` field in the namespace(s) under which they
 	// were created.
-	Custom echo.CustomGetter
+	Configs echo.ConfigGetter
 }
 
 func (c *Config) fillDefaults(ctx resource.Context) error {
@@ -74,21 +75,24 @@ func (c *Config) fillDefaults(ctx resource.Context) error {
 		c.Echos = &Echos{}
 	}
 
+	if c.Configs == nil {
+		defaultConfigs := c.DefaultEchoConfigs(ctx)
+		c.Configs = echo.ConfigFuture(&defaultConfigs)
+	}
+
 	// Verify the namespace for any custom deployments.
-	if c.Custom != nil {
-		for _, custom := range c.Custom.Get() {
-			if custom.Namespace != nil {
-				found := false
-				for _, ns := range c.Namespaces {
-					if custom.Namespace.Name() == ns.Get().Name() {
-						found = true
-						break
-					}
+	for _, config := range c.Configs.Get() {
+		if config.Namespace != nil {
+			found := false
+			for _, ns := range c.Namespaces {
+				if config.Namespace.Name() == ns.Get().Name() {
+					found = true
+					break
 				}
-				if !found {
-					return fmt.Errorf("custom echo deployment %s uses unconfigured namespace %s",
-						custom.NamespacedName().String(), custom.NamespaceName())
-				}
+			}
+			if !found {
+				return fmt.Errorf("custom echo deployment %s uses unconfigured namespace %s",
+					config.NamespacedName().String(), config.NamespaceName())
 			}
 		}
 	}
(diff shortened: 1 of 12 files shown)


pkg/test/framework/components/echo/common/deployment/namespace.go
```diff
@@ -70,122 +70,48 @@ type EchoNamespace struct {
...
-			Service:         VMSvc,
-			Namespace:       n.Namespace,
-			ServiceAccount:  true,
-			Ports:           ports.All(),
-			DeployAsVM:      true,
-			AutoRegisterVM:  true,
-			Subsets:         []echo.SubsetConfig{{}},
-			IncludeExtAuthz: cfg.IncludeExtAuthz,
-		})
+	if cfg.Custom == nil {
```
nmittler, line 73: nit: maybe rename the variable `cfg.Configs` or something? Get rid of the idea of this as being "custom".

pkg/test/framework/components/echo/common/deployment/namespace.go
```diff
@@ -70,122 +70,48 @@ type EchoNamespace struct {
...
+		defaultConfigs := CreateDefaultConfig(n.Namespace, cfg)
```
nmittler, line 74: A few things here:

1. Maybe just call it `DefaultEchoConfigs`
2. Instead of having this logic here, just add a method to the config `cfg.getConfigs` or something. It just returns `cfg.Configs` if not null, otherwise returns `DefaultConfigs()`.
3. Just have all the default configs have an empty namespace.  The logic should interpret an empty namespace to mean: deploy to all namespaces.

pkg/test/framework/components/echo/common/deployment/namespace.go
```diff
@@ -70,122 +70,48 @@ type EchoNamespace struct {
...
-				Service:        ProxylessGRPCSvc,
-				Namespace:      n.Namespace,
-				ServiceAccount: true,
-				Ports:          ports.All(),
-				Subsets: []echo.SubsetConfig{
-					{
-						Annotations: map[echo.Annotation]*echo.AnnotationValue{
-							echo.SidecarInjectTemplates: {
-								Value: "grpc-agent",
+		if !skipDeltaXDS(t) {
```
nmittler, line 80: All of this has to be in `DefaultConfigs`

pkg/test/framework/components/echo/common/deployment/echos.go
```diff
@@ -75,8 +75,8 @@ func (c *Config) fillDefaults(ctx resource.Context) error {
 	}
 
 	// Verify the namespace for any custom deployments.
-	if c.Custom != nil {
-		for _, custom := range c.Custom.Get() {
+	if c.Configs != nil {
```
nmittler, line 78: Actually, can we set c.Configs to the default value here if it's nil? Seems we'd kill two birds with one stone. Then you wouldn't need to check nil here.

pkg/test/framework/components/echo/common/deployment/echos.go
```diff
@@ -152,6 +152,14 @@ func (c *Config) fillDefaults(ctx resource.Context) error {
 	return g.Wait()
 }
 
+func (c *Config) getConfigs(t resource.Context) []echo.Config {
```
nmittler, line 155: And if you follow the suggestion above, we wouldn't need this method.

pkg/test/framework/components/echo/common/deployment/echos.go
aryan16: added one change, hopefully that resolves this query.

pkg/test/framework/components/echo/common/deployment/namespace.go
```diff
@@ -70,130 +70,16 @@ type EchoNamespace struct {
...
-					{
-						Annotations: map[echo.Annotation]*echo.AnnotationValue{
-							echo.SidecarInjectTemplates: {
-								Value: "grpc-agent",
-							},
-						},
-					},
-				},
-			})
+	if cfg.Configs == nil {
```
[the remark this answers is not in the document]
aryan16: oh yes, sorry I got confused with the last comment.

pkg/test/framework/components/echo/common/deployment/echos.go
nmittler, line 78: Shouldn't this just be something like:

```golang
if c.Configs == nil {
  defaultConfigs := CreateDefaultEchoConfigs(*c, ctx)
  c.Configs = echo.ConfigFuture(&defaultConfigs)
}

for _, custom := range c.Configs.Get() {
  ...
}
```

aryan16: I don't this its needed -
1) Default echo configs won't be having any namespaces
2) We just need this check for real custom configs

The current implementation signifies that we verifying the namespaces for custom deployment, good for someone who is reading the code IMO

nmittler: I'm just proposing typical logic for handling defaults. Apply defaults first, then you don't have to ever consider it being nil from that point forward. It's actually simpler logic overall.

aryan16: I agree, done !!

pkg/test/framework/components/echo/common/deployment/namespace.go
```diff
@@ -278,3 +160,122 @@ spec:
 
 	return cfg.Apply(apply.NoCleanup)
 }
+
+func CreateDefaultEchoConfigs(cfg Config, t resource.Context) []echo.Config {
```
nmittler, line 164: Let's just make this a member function of `Config`. Also, you can probably just drop the `Create` in the name. So the call would be `cfg.DefaultEchoConfigs(t)`.

nmittler (approved):

shankgan (approved):

howardjohn (approved):

Issue #40198, opened 2022-08-01
Strange behaviour about LocalityLoadBalancerSetting.failoverPriority config


ktalg:
### Bug Description

There are 3 simple workloads: curl,httpbin-v1(hv1),httpbin-v2(hv2), assuming that the httpbin is the server workload and curl is client. 

```
apiVersion: v1
kind: Service
metadata:
  name: httpbin
  labels:
    app: httpbin
    service: httpbin
spec:
  ports:
    - name: http
      port: 8000
      targetPort: 80
  selector:
    app: httpbin
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: httpbin-v1
spec:
  replicas: 2
  selector:
    matchLabels:
      app: httpbin
      version: v1
  template:
    metadata:
      labels:
        app: httpbin
        version: v1
        a: a
        b: b
    spec:
      containers:
        - image: docker.io/kennethreitz/httpbin
          imagePullPolicy: IfNotPresent
          name: httpbin
          ports:
            - containerPort: 80
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: httpbin-v2
spec:
  replicas: 1
  selector:
    matchLabels:
      app: httpbin
      version: v2
  template:
    metadata:
      labels:
        app: httpbin
        version: v2
        a: a
    spec:
      containers:
        - image: docker.io/kennethreitz/httpbin
          imagePullPolicy: IfNotPresent
          name: httpbin
          ports:
            - containerPort: 80
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: curl
spec:
  replicas: 1
  selector:
    matchLabels:
      app: curl
      version: v1
  template:
    metadata:
      labels:
        app: curl
        version: v1
        a: a
        b: b
    spec:
      containers:
        - image: docker.io/curlimages/curl
          imagePullPolicy: IfNotPresent
          name: curl
          command:
            - sleep
            - 365d
---
apiVersion: networking.istio.io/v1alpha3
kind: DestinationRule
metadata:
  name: dr
spec:
  host: httpbin.test.svc.cluster.local
  trafficPolicy:
    outlierDetection:
      consecutive5xxErrors: 20
      interval: 5m
      baseEjectionTime: 15m
    loadBalancer:
      localityLbSetting:
        failoverPriority:
          - "b"
        enabled: true
---

apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: vs
spec:
  hosts:
[...]

PR #40221, opened 2022-08-02
Add failover priorities to eds cache key

dddddai:
Fixes https://github.com/istio/istio/issues/40198
The root cause is that istiod caches the same endpoints regardless of failover priority labels of each proxy, this pr adds them to eds cache key

### DIFF
--- a/pilot/pkg/xds/endpoint_builder.go
+++ b/pilot/pkg/xds/endpoint_builder.go
@@ -27,6 +27,7 @@ import (
 	networkingapi "istio.io/api/networking/v1alpha3"
 	"istio.io/istio/pilot/pkg/features"
 	"istio.io/istio/pilot/pkg/model"
+	"istio.io/istio/pilot/pkg/networking/core/v1alpha3/loadbalancer"
 	"istio.io/istio/pilot/pkg/networking/util"
 	"istio.io/istio/pilot/pkg/security/authn/factory"
 	"istio.io/istio/pkg/cluster"
@@ -44,14 +45,15 @@ var (
 
 type EndpointBuilder struct {
 	// These fields define the primary key for an endpoint, and can be used as a cache key
-	clusterName     string
-	network         network.ID
-	proxyView       model.ProxyView
-	clusterID       cluster.ID
-	locality        *core.Locality
-	destinationRule *model.ConsolidatedDestRule
-	service         *model.Service
-	clusterLocal    bool
+	clusterName            string
+	network                network.ID
+	proxyView              model.ProxyView
+	clusterID              cluster.ID
+	locality               *core.Locality
+	destinationRule        *model.ConsolidatedDestRule
+	service                *model.Service
+	clusterLocal           bool
+	failoverPriorityLabels []byte
 
 	// These fields are provided for convenience only
 	subsetName string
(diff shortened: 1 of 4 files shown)


pilot/pkg/xds/endpoint_builder.go
```diff
@@ -94,15 +95,16 @@ func NewEndpointBuilder(clusterName string, proxy *model.Proxy, push *model.Push
...
+		clusterName:            clusterName,
+		network:                proxy.Metadata.Network,
+		proxyView:              proxy.GetView(),
+		clusterID:              proxy.Metadata.ClusterID,
+		locality:               proxy.Locality,
+		service:                svc,
+		clusterLocal:           push.IsClusterLocal(svc),
+		destinationRule:        dr,
+		tunnelType:             GetTunnelBuilderType(clusterName, proxy, push),
+		failoverPriorityLabels: make(map[string]string),
```
howardjohn, line 107: can we keep it nil by default? no need to alloc

pilot/pkg/networking/core/v1alpha3/loadbalancer/loadbalancer.go
```diff
@@ -240,6 +242,13 @@ func applyPriorityFailover(
 			}
 		}
 	}
+
+	if failoverPriorityLabels != nil {
+		for _, key := range failoverPriorities {
+			failoverPriorityLabels[key] = proxyLabels[key]
```
howardjohn, line 248: we should be doing this in NewEndpointbuilder, we cannot mutate here

howardjohn (changes requested):

hzxuzhonghu (commented):
Where do you set proxy priority labels?

dddddai:
Sorry it was not completed yesterday, now it's ready for review

pilot/pkg/xds/endpoint_builder.go
```diff
@@ -88,6 +90,8 @@ func NewEndpointBuilder(clusterName string, proxy *model.Proxy, push *model.Push
 		port:       port,
 	}
 
+	b.populateFailoverPriorityLabels()
```
howardjohn, line 93: I am not sure why we need to convert this to bytes and then later also convert it to hash. Why not just directly go to hash?

dddddai: If we convert this to hash, what would we do in `Key()`? Do you mean hashing it again in `Key()`?

hzxuzhonghu: This is good, I am just concerned if the failover priority labels are very customized, the scale of the cache could be per proxy in the worst case.

Do we need to only cache some with standard labels, like Netwok. Cluster, Region/zone/subzone

dddddai: > Do we need to only cache some with standard labels, like Netwok. Cluster, Region/zone/subzone

I'm not sure, @howardjohn WDYT

howardjohn: I think we need all of the ones referenced? otherwise if they select more it won't work?

hzxuzhonghu: Since we have cache size limit, so looks not a big matter. Will trigger evict if the keys number is  too large

hzxuzhonghu (approved):

pilot/pkg/xds/endpoint_builder.go
howardjohn, line 93: nit: I would just inline this. Making it a method makes it seem like it can be called at arbitrary times - it must only be called in NewEndpointBuilder

howardjohn (approved):

howardjohn:
/retest

known flake, I have a fix in another PR

PR #40243, opened 2022-08-03
fix empty endpoints weight for AUTO_PASSTHROUGH gateway

bincherry:

For AUTO_PASSTHROUGH gateway, typically istio-eastwestgateway, the endpoints of SNI-dnat cluster will be filtered by mtls checker. Yet the filtered `LocalityLbEndpoints` results have empty `LoadBalancingWeight`, that would cause LB error in sidecar.

### DIFF
--- a/pilot/pkg/xds/ep_filters.go
+++ b/pilot/pkg/xds/ep_filters.go
@@ -224,6 +224,7 @@ func (b *EndpointBuilder) EndpointsWithMTLSFilter(endpoints []*LocalityEndpoints
 			lbEndpoints.append(ep.istioEndpoints[i], lbEp)
 		}
 
+		lbEndpoints.refreshWeight()
 		filtered = append(filtered, lbEndpoints)
 	}
 
--- a/pilot/pkg/xds/ep_filters_test.go
+++ b/pilot/pkg/xds/ep_filters_test.go
@@ -175,6 +175,269 @@ var networkFiltered = []networkFilterCase{
 	},
 }
 
+var mtlsCases = map[string]map[string]struct {
+	Config         config.Config
+	Configs        []config.Config
+	IsMtlsDisabled bool
+}{
+	gvk.PeerAuthentication.String(): {
+		"mtls-off-ineffective": {
+			Config: config.Config{
+				Meta: config.Meta{
+					GroupVersionKind: gvk.PeerAuthentication,
+					Name:             "mtls-partial",
+					Namespace:        "istio-system",
+				},
+				Spec: &security.PeerAuthentication{
+					Selector: &v1beta1.WorkloadSelector{
+						// shouldn't affect our test workload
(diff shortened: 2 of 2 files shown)


howardjohn (changes requested):
Thanks for the PR! Can you give more details? How to replicate? What error? And we need tests covering this

tedli:
The case is that, service call cross clusters through eastwest gateway with lb policy of consistency hash, result no healthy upstream. Because the locality weight in cluster load assignment is 0.

Ring hash logic in envoy code, read the locality weight:

https://github.com/envoyproxy/envoy/blob/652a89b6a76e6a59521a46d764b2bd52317421a0/source/common/upstream/thread_aware_lb_impl.cc#L54

https://github.com/envoyproxy/envoy/blob/652a89b6a76e6a59521a46d764b2bd52317421a0/source/common/upstream/ring_hash_lb.cc#L54

bincherry:
> Thanks for the PR! Can you give more details? How to replicate? What error? And we need tests covering this

Errors would happen in sidecar because of the empty load_balancing_weight of endpoints, which should be as least 1. For exmaple as described by tedli.

For now the empty load_balancing_weight will happen on every eastwestgateway, and we could get config_dump and check endpoints of SNI DNAT clusters to find it.

```

    {
     "endpoint_config": {
      "@type": "type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment",
      "cluster_name": "outbound_.8081_._.web.sky.svc.cluster.local",
      "endpoints": [
       {
        "locality": {},
        "lb_endpoints": [
         {
          "endpoint": {
           "address": {
            "socket_address": {
             "address": "172.30.40.144",
             "port_value": 8081
            }
           },
           "health_check_config": {}
          },
          "health_status": "HEALTHY",
          "metadata": {
           "filter_metadata": {
            "envoy.transport_socket_match": {
             "tlsMode": "istio"
            },
            "istio": {
             "workload": "web;sky;web;latest;cid-5a3b17793356"
            }
           }
          },
          "load_balancing_weight": 1
         }
        ],
        "load_balancing_weight": 0             ### HERE
       }
      ],
      "policy": {
       "overprovisioning_factor": 140
      }
     }
    },
```

hzxuzhonghu:
cc @stevenctl 

/ok-to-test

stevenctl (approved):

stevenctl:
Good catch, probably part of the fix for https://github.com/istio/istio/issues/38704

stevenctl:
cc @howardjohn if you want to drop the change request

hzxuzhonghu (approved):

tedli:
Hi all,

During we debug with this pr, we also found another issue https://github.com/istio/istio.io/pull/11680, which was not documented. Then we give a pr to the doc repo.

Could someone help to assign reviewers to take a look?

PR #40252, opened 2022-08-03
Cleanup makefile env vars

howardjohn:
This moves a bunch of env vars from Makefile to init.sh. The reason is these are scoped to init.sh. By defining them in Makefile, we run into issues when we call nested Make.

For example:

```
make docker:
  TARGET_ARCH=amd64 make init
  TARGET_ARCH=amd64 make init
```

This doesn't work, because each var allows explicitly overriding it. On the first make, we set them once lead to them not being updated on subsequent makes.

Instead, we just set them locally so they don't impact anything beyond `init.sh`.

End result is the same for all cases except future multi-arch builds.


ericvn (approved):
Thanks. This should solve some of the issues I chased in the past with nested make calls.

PR #40263, opened 2022-08-04
Bump master to 1.16

Monkeyanator:
For https://github.com/istio/istio/issues/40127

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
@@ -2465,6 +2465,232 @@ spec:
                     local:
                       inline_string: "envoy.wasm.stats"
 ---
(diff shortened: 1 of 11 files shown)


manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
```diff
@@ -2465,6 +2465,232 @@ spec:
                     local:
                       inline_string: "envoy.wasm.stats"
 ---
+# Source: istiod/templates/telemetryv2_1.16.yaml
```
ericvn, line 2468: I'm not sure, but do we remove the .13?

ericvn, line 2468: Do we remove the .13 files as they should be EOL by the time this branch is published as 1.16? Or do we wait until 1.13 is EOL in approx. 10 weeks?

howardjohn (approved):

PR #40295, opened 2022-08-05
No need to set total weight anymore

hzxuzhonghu:

TotalWeight is now an optional field, which defaults to the behavior we added before.


ramaraochavali:
Do we need version check for this?

zirain:
link to envoy PR https://github.com/envoyproxy/envoy/pull/21892

hzxuzhonghu:
No need to introduce a version check to clean TotalWeight set, which is a small stuff.

PR #40304, opened 2022-08-05
Don't gen default webhooks if operatorManageWebhooks is true

### LINKED ISSUE #37674: External istiod installation overriding local injector webhook
### Bug Description

The [external control plane installation instructions](https://istio.io/latest/docs/setup/install/external-controlplane/) start by installing Istio on the external control plane cluster using the default profile. This istiod will be used to control an ingress gateway that will provide access to the (yet to be installed) external control plane. After [this installation step](https://istio.io/latest/docs/setup/install/external-controlplane/#set-up-a-gateway-in-the-external-cluster) there are two mutatingwebhookconfigurations on the external cluster:

1.  `istio-revision-tag-default`
2. `istio-sidecar-injector` (this one seems to be disabled with `istio.io/deactivated: never-match` ???)

Injection is working fine using the `istio-revision-tag-default` webhook at this point.

Later in the instructions we [install istiod again](https://istio.io/latest/docs/setup/install/external-controlplane/#set-up-the-control-plane-in-the-external-cluster) on the same cluster, but in a different namespace, `external-istiod`. This istiod is the external control plane that is configured with remote keys to manage a remote config cluster. The IOP for this install includes `operatorManageWebhooks: true` because it is not supposed to manage local webhooks.

However, as of Istio 1.12, the second install seems to be creating a new webhook on the local configuration, `istio-revision-tag-default-external-istiod` and also seems to be deleting the correct local webhook, `istio-revision-tag-default`.
At this point injection is broken on the local cluster, because it's pointing to the wrong istiod.

It seems that there is code that is doing all this management that should not be if `operatorManageWebhooks: true`?
https://github.com/istio/istio/blob/393f88cd255f1c0caea6fac5d21aaf0d3e412aad/operator/cmd/mesh/install.go#L195

### Version

```prose
1.12+
```

### Additional Information

_No response_

frankbu:

Fixes #37674

### DIFF
--- a/operator/cmd/mesh/install.go
+++ b/operator/cmd/mesh/install.go
@@ -194,17 +194,20 @@ func Install(rootArgs *RootArgs, iArgs *InstallArgs, logOpts *log.Options, stdOu
 
 	// Detect whether previous installation exists prior to performing the installation.
 	exists := revtag.PreviousInstallExists(context.Background(), kubeClient.Kube())
-	pilotEnabled := iop.Spec.Components.Pilot != nil && iop.Spec.Components.Pilot.Enabled.Value
 	rev := iop.Spec.Revision
-	if rev == "" && pilotEnabled {
+	isDefaultInstallation := rev == "" && iop.Spec.Components.Pilot != nil && iop.Spec.Components.Pilot.Enabled.Value
+	operatorManageWebhooks := operatorManageWebhooks(iop)
+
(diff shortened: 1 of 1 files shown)


litong01:
@frankbu Can this be done in template files instead of in the go code?

frankbu:
@litong01 no, this has to be done manually with helm.

Monkeyanator (approved):

Issue #40299, opened 2022-08-05
Retry policy without perTryTimeout in VirtualService makes grpc request timeout unexpectlly


suninuni:
### Bug Description

Retry policy without `perTryTimeout` in VirtualService will cause some UT request for grpc.

![image](https://user-images.githubusercontent.com/7345529/183043283-4f3cb073-2182-4355-9f53-3648c6297efa.png)

The default retry policy in RDS is 

```json
"route": {
    "cluster": "outbound|1600||grpc-service.default.svc.cluster.local",
    "timeout": "0s",
    "retryPolicy": {
        "retryOn": "connect-failure,refused-stream,unavailable,cancelled,retriable-status-codes",
        "numRetries": 2,
        "retryHostPredicate": [
            {
                "name": "envoy.retry_host_predicates.previous_hosts"
            }
        ],
        "hostSelectionRetryMaxAttempts": "5",
        "retriableStatusCodes": [
            503
        ]
    },
    "maxStreamDuration": {
        "maxStreamDuration": "0s",
        "grpcTimeoutHeaderMax": "0s"
    }
}
```

After I apply the VirtualService

```yaml
spec:
  hosts:
  - grpc-service
  http:
  - retries:
      attempts: 1
      retryOn: unavailable,cancelled
    route:
    - destination:
        host: grpc-service
        port:
          number: 1600
```

The RDS config changes to 

```json
"route": {
    "cluster": "outbound|1600||grpc-service.default.svc.cluster.local",
    "timeout": "0s",
    "retryPolicy": {
        "retryOn": "unavailable,cancelled",
        "numRetries": 1,
        "retryHostPredicate": [
            {
                "name": "envoy.retry_host_predicates.previous_hosts"
            }
        ],
        "hostSelectionRetryMaxAttempts": "5"
    },
    "maxGrpcTimeout": "0s"
}
```

It deleted `maxStreamDuration` and added `maxGrpcTimeout`, I can't confirm which configuration is causing those UT requests.

But I try to modify the `defaultHttpRetryPolicy` in mesh config, it will not cause these promblem.

```yaml
    defaultHttpRetryPolicy:
      retryOn: unavailable
      attempts: 1
```

```json
"route": {
    "cluster": "outbound|1600||grpc-service.default.svc.cluster.local",
    "timeout": "0s",
    "retryPolicy": {
        "retryOn": "unavailable",
        "numRetries": 1,
        "retryHostPredicate": [
            {
[...]

PR #40320, opened 2022-08-07
unify timeout logic across service and virtual service

ramaraochavali:
There is discrepancy in how we set timeout when a VS is defined (without timeout) and VS is not defined at all. We should be consistently setting the timeouts to `notimeout` in both cases.

This PR fixes it.
Fixes https://github.com/istio/istio/issues/40299

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/route/route.go
+++ b/pilot/pkg/networking/core/v1alpha3/route/route.go
@@ -28,6 +28,7 @@ import (
 	matcher "github.com/envoyproxy/go-control-plane/envoy/type/matcher/v3"
 	xdstype "github.com/envoyproxy/go-control-plane/envoy/type/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
+	"github.com/golang/protobuf/ptypes/duration"
 	anypb "google.golang.org/protobuf/types/known/anypb"
 	"google.golang.org/protobuf/types/known/durationpb"
 	wrappers "google.golang.org/protobuf/types/known/wrapperspb"
@@ -64,6 +65,8 @@ const DefaultRouteName = "default"
 // prefixMatchRegex optionally matches "/..." at the end of a path.
 // regex taken from https://github.com/projectcontour/contour/blob/2b3376449bedfea7b8cea5fbade99fb64009c0f6/internal/envoy/v3/route.go#L59
 const prefixMatchRegex = `((\/).*)?`
+
+var notimeout = durationpb.New(0)
 
 // VirtualHostWrapper is a context-dependent virtual host entry with guarded routes.
 // Note: Currently we are not fully utilizing this structure. We could invoke this logic
@@ -489,19 +492,11 @@ func applyHTTPRouteDestination(
 		RetryPolicy: retry.ConvertPolicy(policy),
 	}
 
-	// Configure timeouts specified by Virtual Service if they are provided, otherwise set it to defaults.
-	action.Timeout = features.DefaultRequestTimeout
-	if in.Timeout != nil {
-		action.Timeout = in.Timeout
+	if in.Retries != nil {
+		action.RetryPolicy = retry.ConvertPolicy(in.Retries)
 	}
-	if node.IsProxylessGrpc() {
-		// TODO(stevenctl) merge these paths; grpc's xDS impl will not read the deprecated value
-		action.MaxStreamDuration = &route.RouteAction_MaxStreamDuration{MaxStreamDuration: action.Timeout}
-	} else {
-		// Use deprecated value for now as the replacement MaxStreamDuration has some regressions.
-		// nolint: staticcheck
-		action.MaxGrpcTimeout = action.Timeout
-	}
+
+	setTimeouts(action, features.DefaultRequestTimeout, in.Timeout, node)
 
 	if model.UseGatewaySemantics(vs) && util.IsIstioVersionGE115(node.IstioVersion) {
 		// return 500 for invalid backends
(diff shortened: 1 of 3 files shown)


pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -489,20 +492,12 @@ func applyHTTPRouteDestination(
...
-	}
-	if node.IsProxylessGrpc() {
-		// TODO(stevenctl) merge these paths; grpc's xDS impl will not read the deprecated value
-		action.MaxStreamDuration = &route.RouteAction_MaxStreamDuration{MaxStreamDuration: action.Timeout}
-	} else {
-		// Use deprecated value for now as the replacement MaxStreamDuration has some regressions.
-		// nolint: staticcheck
-		action.MaxGrpcTimeout = action.Timeout
+	if in.Retries != nil {
+		action.RetryPolicy = retry.ConvertPolicy(in.Retries)
```
hzxuzhonghu, line 496: duplicate with L492

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -1091,13 +1085,44 @@ func BuildDefaultHTTPInboundRoute(clusterName string, operation string) *route.R
 	return val
 }
 
+func setTimeouts(action *route.RouteAction, defaultTimeout *duration.Duration,
+	vsTimeout *duration.Duration, node *model.Proxy,
```
hzxuzhonghu, line 1089: add comments about the args

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -1091,13 +1085,44 @@ func BuildDefaultHTTPInboundRoute(clusterName string, operation string) *route.R
...
+) {
+	// If default request timeout or virtual service timeout is not set, default to no timeout.
+	if defaultTimeout.AsDuration().Nanoseconds() == 0 && vsTimeout == nil {
```
hzxuzhonghu, line 1092: why need this separate check fo 0?

hzxuzhonghu: cannot L1102-L1115 cover this case?

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -1091,13 +1085,44 @@ func BuildDefaultHTTPInboundRoute(clusterName string, operation string) *route.R
 	return val
 }
 
+func setTimeouts(action *route.RouteAction, defaultTimeout *duration.Duration,
```
hzxuzhonghu, line 1088: s/setTimeouts/setTimeout

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -1091,13 +1085,44 @@ func BuildDefaultHTTPInboundRoute(clusterName string, operation string) *route.R
...
+
 // BuildDefaultHTTPOutboundRoute builds a default outbound route, including a retry policy.
 func BuildDefaultHTTPOutboundRoute(clusterName string, operation string, mesh *meshconfig.MeshConfig) *route.Route {
-	// Start with the same configuration as for inbound.
-	out := BuildDefaultHTTPInboundRoute(clusterName, operation)
+	out := buildDefaultHTTPRoute(clusterName, operation)
 
 	// Add a default retry policy for outbound routes.
 	out.GetRoute().RetryPolicy = retry.ConvertPolicy(mesh.GetDefaultHttpRetryPolicy())
+	setTimeouts(out.GetRoute(), features.DefaultRequestTimeout, nil, nil)
```
hzxuzhonghu, line 1125: Is this missed before?

ramaraochavali: yes

pilot/pkg/networking/core/v1alpha3/route/route.go
ramaraochavali: Because we have to disable timeouts completely. and we can not set non zero maxstream in L1102 because of bug in Envoy. That is why we are using that deprecated field. This is much safer to do.

		action.MaxStreamDuration = &route.RouteAction_MaxStreamDuration{
			MaxStreamDuration: notimeout,
			// If not configured at all, the grpc-timeout header is not used and
			// gRPC requests time out like any other requests using timeout or its default.
			GrpcTimeoutHeaderMax: notimeout,
		}

hzxuzhonghu: IC, could not treate`defaultTimeout` = `features.DefaultRequestTimeout` here.

Thet confuses me.

Why not separate inbound, which always has no timeout

ramaraochavali: If it is confusing, I can separate inbound in to a separate function.

hzxuzhonghu (commented):
LGTM

howardjohn (approved):

ericvn:
It looks like 1.15 wasn't part of the cherry-picks while earlier versions where which would lead to a partial regression in 1.15.

ramaraochavali:
@ericvn I thought 1.15 branch was not yet cut. Thanks for that.

ericvn:
Branch was cut 1 1/2 weeks ago. I'm not sure the best way for RMs to get the word out.

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -1091,13 +1087,38 @@ func BuildDefaultHTTPInboundRoute(clusterName string, operation string) *route.R
...
+		// Set MaxStreamDuration only for notimeout cases otherwise it wont be honored.
+		if action.Timeout.AsDuration().Nanoseconds() == 0 {
+			action.MaxStreamDuration.MaxStreamDuration = action.Timeout
+			action.MaxStreamDuration.GrpcTimeoutHeaderMax = action.Timeout
+		} else {
+			// If not configured at all, the grpc-timeout header is not used and
+			// gRPC requests time out like any other requests using timeout or its default.
+			// Use deprecated value for now as the replacement MaxStreamDuration has some regressions.
+			// nolint: staticcheck
+			action.MaxGrpcTimeout = action.Timeout
```
halfrost, line 1111: I have a small quick question about this line. Are MaxStreamDuration and MaxGrpcTimeout exactly the same? There is also a MaxStreamDuration field in my project, but without MaxGrpcTimeout.

ramaraochavali: They are not exactly same. Please see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html?highlight=max_grpc_timeout#config-route-v3-routeaction-maxstreamduration

There are issues with max_stream_duration. Refer to https://github.com/envoyproxy/envoy/issues/19073 and https://github.com/envoyproxy/envoy/issues/19073

PR #40322, opened 2022-08-08
Refactor proxy labels: add labels fields to Proxy

hzxuzhonghu:

This is for https://github.com/istio/istio/pull/40186#discussion_r935661479
 

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pilot/pkg/model/context.go
```diff
@@ -849,14 +858,35 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
...
 func (node *Proxy) SetWorkloadLabels(env *Environment) {
-	// First get the workload labels from node meta
-	if len(node.Metadata.Labels) > 0 {
+	if len(node.Labels) > 0 {
 		return
 	}
-	// Fallback to calling GetProxyWorkloadLabels
-	node.Metadata.Labels = env.GetProxyWorkloadLabels(node)
+	labels := env.GetProxyWorkloadLabels(node)
+	// when IstioMetaLabels unset, calculate the IstioMetaLabels. It is only set once.
```
ramaraochavali, line 868: Do we want to do this as part of this PR?

hzxuzhonghu: Maybe, to show usage

pilot/pkg/model/context.go
```diff
@@ -273,6 +273,15 @@ type Proxy struct {
 	// NOTE: DO NOT USE THIS FIELD TO CONSTRUCT DNS NAMES
 	ConfigNamespace string
 
+	// Labels specifies the set of workload instance (ex: k8s pod) labels associated with this node.
+	// Labels can be different from that in Metadata because of pod labels update after startup,
+	// while NodeMetadata.Labels are set during bootstrap.
+	Labels map[string]string `json:"LABELS,omitempty"`
```
howardjohn, line 279:
```suggestion
	Labels map[string]string
```

Do we need `json` stuff? It seems confusing since this isn't actually marshaled to/from json

pilot/pkg/model/context.go
```diff
@@ -273,6 +273,15 @@ type Proxy struct {
...
+
+	// IstioMetaLabels contains the labels specified by ISTIO_METAJSON_LABELS and platform instance,
+	// so we can tell the difference between user specified labels and istio labels.
+	IstioMetaLabels map[string]string `json:"ISTIO_META_LABELS,omitempty"`
```
howardjohn, line 283: Do we need to copy this? or just keep it in Nodemetadata.Labels?

pilot/pkg/model/context.go
```diff
@@ -536,7 +545,7 @@ type NodeMetadata struct {
 	// Labels specifies the set of workload instance (ex: k8s pod) labels associated with this node.
 	Labels map[string]string `json:"LABELS,omitempty"`
```
howardjohn, line 546: So IIUC, this should ~never be used except for in the initialization. Everything else should use `proxy.Labels` not `proxy.Metadata.Labels`? If so, it seems like we missed a few - EffectiveProxyConfig for example

pilot/pkg/model/context.go
howardjohn, line 868: I am a bit confused. Why are we setting IstioMetaLabels to something other than the LABELS set in metadata?

It seems unexpect that IstioMetaLabels would fetch labels from GetProxyWorkloadLabels

pilot/pkg/model/context.go
```diff
@@ -849,14 +858,35 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
...
+	}
+
+	node.Labels = make(map[string]string, len(labels)+len(node.IstioMetaLabels))
+	// we can't just equate proxy workload labels to node meta labels as it may be customized by user
+	// with `ISTIO_METAJSON_LABELS` env (pkg/bootstrap/config.go extractAttributesMetadata).
+	// so, we fill the `ISTIO_METAJSON_LABELS` as well.
+	for k, v := range node.IstioMetaLabels {
+		node.Labels[k] = v
+	}
+	for k, v := range labels {
```
howardjohn, line 887: are we missing label removal here? Seems node metadata labels cannot be deleted?

pilot/pkg/model/push_context_test.go
```diff
@@ -565,6 +570,9 @@ func TestWasmPlugins(t *testing.T) {
 			name: "ingress",
 			node: &Proxy{
 				ConfigNamespace: "other",
+				Labels: map[string]string{
```
howardjohn, line 573: can we make https://github.com/istio/istio/blob/db637b3e4d3c09da0f8fc8f3222231e1834668ed/pilot/pkg/networking/core/v1alpha3/fake.go#L209 handle this so we don't need to change every test to repeat the labels multiple times?

pilot/pkg/model/context.go
hzxuzhonghu: `IstioMetaLabels` exist here to help processing pod label delete, because it caches the `ISTIO_METAJSON_LABELS`

pilot/pkg/model/context.go
hzxuzhonghu: hope answered below

pilot/pkg/model/push_context_test.go
hzxuzhonghu: have to take a look

pilot/pkg/model/context.go
hzxuzhonghu: seems no

pilot/pkg/model/context.go
hzxuzhonghu: `EffectiveProxyConfig` is called in injector, which is using pod labels, so no influence on it

```
	if wh.env.PushContext != nil && wh.env.PushContext.ProxyConfigs != nil {
		if generatedProxyConfig := wh.env.PushContext.ProxyConfigs.EffectiveProxyConfig(
			&model.NodeMetadata{
				Namespace:   pod.Namespace,
				Labels:      pod.Labels,
				Annotations: pod.Annotations,
			}, wh.meshConfig); generatedProxyConfig != nil {
			proxyConfig = generatedProxyConfig
		}
	}
```

pilot/pkg/model/push_context_test.go
hzxuzhonghu: We could, but ConfigGenTest.SetupProxy require more params

hzxuzhonghu:
@howardjohn @ramaraochavali in case you miss it

pilot/pkg/model/push_context_test.go
howardjohn: I am not sure what you mean? I just mean that we should have users of SetupProxy only set labels once and then we compute them (via SetWorkloadLabels?) for them automatically so they do not need to remember to set 2 sets of the same label

hzxuzhonghu: I mean many ut are not using `ConfigGenTest`

SetWorkloadLabels is triggered by xds request, xds push, right?

Here it is the ut, which will not call `SetWorkloadLabels`. Do you mean to explicitly call it in `SetupProxy`?

And we transit all tests to use `ConfigGenTest`. We should file a issue to track it, WDYT

howardjohn: Ah I see, misunderstood. thanks. I think `model` cannot use ConfigGenTest

pilot/pkg/security/authn/v1beta1/policy_applier_test.go
```diff
@@ -1536,6 +1536,9 @@ func TestInboundMTLSSettings(t *testing.T) {
 	}
 
 	testNode := &model.Proxy{
+		Labels: map[string]string{
```
howardjohn, line 1539: nit: should we replace Metadata.Labels with Labels instead of adding? Only the latter should be used (?)

pilot/pkg/model/context.go
```diff
@@ -273,6 +273,15 @@ type Proxy struct {
 	// NOTE: DO NOT USE THIS FIELD TO CONSTRUCT DNS NAMES
 	ConfigNamespace string
 
+	// Labels specifies the set of workload instance (ex: k8s pod) labels associated with this node.
+	// Labels can be different from that in Metadata because of pod labels update after startup,
+	// while NodeMetadata.Labels are set during bootstrap.
+	Labels map[string]string
+
+	// IstioMetaLabels contains the labels specified by ISTIO_METAJSON_LABELS and platform instance,
```
howardjohn, line 281: Is it "ISTIO_METAJSON_LABELS and platform instance" or "ISTIO_METAJSON_LABELS **minus** platform instance"?

pilot/pkg/model/context.go
```diff
@@ -849,14 +858,35 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
...
-	// First get the workload labels from node meta
-	if len(node.Metadata.Labels) > 0 {
+	if len(node.Labels) > 0 {
 		return
 	}
-	// Fallback to calling GetProxyWorkloadLabels
-	node.Metadata.Labels = env.GetProxyWorkloadLabels(node)
+	labels := env.GetProxyWorkloadLabels(node)
+	// when IstioMetaLabels unset, calculate the IstioMetaLabels. It is only set once.
+	if node.IstioMetaLabels == nil {
```
howardjohn, line 869: This is a bit tricky since we are computing it _once per connection_, not once per proxy lifecycle.

For example. Before state:

```json
  "echo-9bfbfd967-ldf8r.default-7": {
    "Labels": {
      "app": "echo",
      "howardjohn-test": "foo",
      "later": "added",
      "pod-template-hash": "9bfbfd967",
      "security.istio.io/tlsMode": "istio",
      "service.istio.io/canonical-name": "echo",
      "service.istio.io/canonical-revision": "latest",
      "topology.istio.io/cluster": "Kubernetes"
    },
    "IstioMetaLabels": {
      "howardjohn-test": "foo"
    },
    "Metadata.Labels": {
      "app": "echo",
      "howardjohn-test": "foo",
      "security.istio.io/tlsMode": "istio",
      "service.istio.io/canonical-name": "echo",
      "service.istio.io/canonical-revision": "latest"
    }
  },
```

then I remove 'service.istio.io/canonical-name' label, and reconnect to istiod:

```json
  "echo-9bfbfd967-ldf8r.default-1": {
    "Labels": {
      "app": "echo",
      "howardjohn-test": "foo",
      "later": "added",
      "pod-template-hash": "9bfbfd967",
      "security.istio.io/tlsMode": "istio",
      "service.istio.io/canonical-name": "echo",
      "service.istio.io/canonical-revision": "latest",
      "topology.istio.io/cluster": "Kubernetes"
    },
    "MetaLabels": {
      "howardjohn-test": "foo",
      "service.istio.io/canonical-name": "echo"
    },
    "MetadataLabels": {
      "app": "echo",
      "howardjohn-test": "foo",
      "security.istio.io/tlsMode": "istio",
      "service.istio.io/canonical-name": "echo",
      "service.istio.io/canonical-revision": "latest"
    }
  },
```

My Labels hasn't actually changed to account for this; in effect we cannot remove any labels that our pod started with.

I am not quite sure a reliable way to know this though...

hzxuzhonghu: Right, this is the pre step, which only does refactor part. In #40186, it will update labels.

ramaraochavali (approved):

pilot/pkg/model/context.go
hzxuzhonghu: platform instance labels only from GCP vm.

For other envs, it is only that in ISTIO_METAJSON_LABELS

pilot/pkg/model/context.go
howardjohn: This bug is still present.

Start: a pod has a custom metadata.labels
```
 "shell-858c44699f-knnkq.default-23": {
    "Labels": {
      "app": "shell",
      "cluster": "foo",
      "pod-template-hash": "858c44699f",
      "security.istio.io/tlsMode": "istio",
      "service.istio.io/canonical-name": "shell",
      "service.istio.io/canonical-revision": "latest",
      "topology.istio.io/cluster": "Kubernetes"
    },
    "MetaLabels": {
      "cluster": "foo"
    },
    "MetadataLabels": {
      "app": "shell",
      "cluster": "foo",
      "security.istio.io/tlsMode": "istio",
      "service.istio.io/canonical-name": "shell",
      "service.istio.io/canonical-revision": "latest"
    }
  },

```

Then: `k label pod shell-858c44699f-knnkq service.istio.io/canonical-revision-`
```
    "shell-858c44699f-knnkq.default-23": {
      "Labels": {
        "app": "shell",
        "cluster": "foo",
        "pod-template-hash": "858c44699f",
        "security.istio.io/tlsMode": "istio",
        "service.istio.io/canonical-name": "shell",
        "topology.istio.io/cluster": "Kubernetes"
      },
      "MetaLabels": {
        "cluster": "foo"
      },
      "MetadataLabels": {
        "app": "shell",
        "cluster": "foo",
        "security.istio.io/tlsMode": "istio",
        "service.istio.io/canonical-name": "shell",
        "service.istio.io/canonical-revision": "latest"
      }
    },
```

This is correct so far...

Now, xds disconnects and reconnects:
```
      "shell-858c44699f-knnkq.default-2": {
        "Labels": {
          "app": "shell",
          "cluster": "foo",
          "pod-template-hash": "858c44699f",
          "security.istio.io/tlsMode": "istio",
          "service.istio.io/canonical-name": "shell",
          "service.istio.io/canonical-revision": "latest",
          "topology.istio.io/cluster": "Kubernetes"
        },
        "MetaLabels": {
          "cluster": "foo",
          "service.istio.io/canonical-revision": "latest"
        },
        "MetadataLabels": {
          "app": "shell",
          "cluster": "foo",
          "security.istio.io/tlsMode": "istio",
          "service.istio.io/canonical-name": "shell",
          "service.istio.io/canonical-revision": "latest"
        }
      },
```

Now the label is back unexpectedly

howardjohn (approved):
This PR looks good. I have one comment but it really applies to the other PR which already merged accidentally. Lets see if we can forward fix it

howardjohn:
/retest

infra flake

pkg/istio-agent/xds_proxy.go
```diff
@@ -444,6 +449,18 @@ func (p *XdsProxy) handleUpstreamRequest(con *ProxyConnection) {
...
 			}
 
+			// override the first xds request node metadata labels
+			// TODO: if we want to regenerate all metadata, use `GetNodeMetaData`
+			if !initialRequestsSent.Load() {
+				if req.Node != nil {
+					meta, _ := model.ParseBootstrapNodeMetadata(req.Node.Metadata)
+					if meta != nil {
+						meta.Labels, _ = bootstrap.GetNodeMetaDataLabels(p.platform)
+						req.Node.Metadata = meta.ToStruct()
```
hzxuzhonghu, line 459: @howardjohn ptal here, this is to fix the reconnect

pkg/istio-agent/xds_proxy.go
```diff
@@ -444,6 +449,18 @@ func (p *XdsProxy) handleUpstreamRequest(con *ProxyConnection) {
 				return
 			}
 
+			// override the first xds request node metadata labels
+			// TODO: if we want to regenerate all metadata, use `GetNodeMetaData`
+			if !initialRequestsSent.Load() {
+				if req.Node != nil {
+					meta, _ := model.ParseBootstrapNodeMetadata(req.Node.Metadata)
```
howardjohn, line 456: not sure we should ignore error here?

pkg/istio-agent/xds_proxy.go
```diff
@@ -444,6 +449,18 @@ func (p *XdsProxy) handleUpstreamRequest(con *ProxyConnection) {
...
+					if meta != nil {
+						meta.Labels, _ = bootstrap.GetNodeMetaDataLabels(p.platform)
```
howardjohn, line 458: same here, on error this may silently wipe out labels

pkg/istio-agent/xds_proxy.go
```diff
@@ -444,6 +449,18 @@ func (p *XdsProxy) handleUpstreamRequest(con *ProxyConnection) {
 				return
 			}
 
+			// override the first xds request node metadata labels
+			// TODO: if we want to regenerate all metadata, use `GetNodeMetaData`
+			if !initialRequestsSent.Load() {
```
howardjohn, line 454: `initialRequestsSent` is not the same has "this is the first request"; it means we sent the custom Istio types, which occur after LDS. This just happens to work because `req.Node` is nil for all subsequent requests due to `set_node_on_first_message_only`. If `set_node_on_first_message_only=false` was ever used, I think we would set the labels for the first 2 requests, but not subsequent ones which would be incorrect.

I think the more correct and simpler approach is just to always do this if req.Node is non-nil

pkg/bootstrap/config.go
```diff
@@ -619,6 +630,39 @@ func GetNodeMetaData(options MetadataOptions) (*model.Node, error) {
 	}, nil
 }
 
+func GetNodeMetaDataLabels(platform platform.Environment) (map[string]string, error) {
```
howardjohn, line 633: can we reuse this within `GetNodeMetaData` to avoid code drift?

pkg/istio-agent/xds_proxy.go
```diff
@@ -444,6 +449,18 @@ func (p *XdsProxy) handleUpstreamRequest(con *ProxyConnection) {
 				return
 			}
 
+			// override the first xds request node metadata labels
```
howardjohn, line 452: Overall I am pretty concerned about mutating the incoming XDS requests... a few possible concerns:

1. istiod 1.16 will need to support proxy 1.15, which do not have this code, so the new label stuff won't work which may lead to incorrect config (including authz)
2. Possible performance impact of mutating request each time. Not likely a major concern though.
3. Only works in xds proxy; this means other clients (proxyless grpc, etc) will be broken still which can then lead to RBAC bypass.

hzxuzhonghu: I think 1 and 2 are not big problem, since this is a new feature.
For 3, I think grpc should so something in itself to reload metadata?

pkg/bootstrap/config.go
hzxuzhonghu: If so, i would rather make use of `func (a *Agent) generateNodeMetadata() `

pkg/istio-agent/xds_proxy.go
howardjohn: 1 is a problem. Labels are an existing feature, and we are changing how they work. if we now incorrectly determine the set of labels, that could be a CVE. Same with 2 -- this impacts everyone, not just users mutating labels.

on 3 - If we are changing requirements on clients maybe we should have wider discussion outside this PR?

hzxuzhonghu: As you said labels update by design has no problem(k8s long support labels update), incorrectly determine labels could cause some bug or CVE. 

For 2, this is why I only regen labels only instead if all meta. Not a big cost though, it is run in proxy rather than istiod.

For 3, we should have a document about labels behavior.

PR #40327, opened 2022-08-08
add identifier to crd client logs

ramaraochavali:
To clearly distinguish controller logs
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/config/kube/crdclient/client.go
+++ b/pilot/pkg/config/kube/crdclient/client.go
@@ -224,15 +224,15 @@ func (cl *Client) SetWatchErrorHandler(handler func(r *cache.Reflector, err erro
 // Run the queue and all informers. Callers should  wait for HasSynced() before depending on results.
 func (cl *Client) Run(stop <-chan struct{}) {
 	t0 := time.Now()
-	scope.Info("%s:Starting Pilot K8S CRD controller", cl.identifier)
+	scope.Infof("%s:Starting Pilot K8S CRD controller", cl.identifier)
 
 	if !kube.WaitForCacheSync(stop, cl.informerSynced) {
-		scope.Error("%s:Failed to sync Pilot K8S CRD controller cache", cl.identifier)
+		scope.Errorf("%s:Failed to sync Pilot K8S CRD controller cache", cl.identifier)
 		return
 	}
 	cl.SyncAll()
 	cl.initialSync.Store(true)
-	scope.Info("%s:Pilot K8S CRD controller synced ", time.Since(t0), cl.identifier)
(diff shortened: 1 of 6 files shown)


hzxuzhonghu (commented):
Maybe not very correlated, I doubt can we reuse the crd client

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -183,7 +187,7 @@ func NewForSchemas(client kube.Client, revision, domainSuffix string, schemas co
 			if _, f := known[name]; f {
 				handleCRDAdd(out, name, nil)
 			} else {
-				scope.Warnf("Skipping CRD %v as it is not present", s.Resource().GroupVersionKind())
+				scope.Warnf("%s:Skipping CRD %v as it is not present", identifier, s.Resource().GroupVersionKind())
```
howardjohn, line 190: we can make `client.Scope = scope.WithLabel("controller", identifier)` and avoid needing to add this everywhere. Then just do `client.Scope.Warnf` etc

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -163,6 +164,7 @@ func NewForSchemas(client kube.Client, revision, domainSuffix string, schemas co
 			GroupVersionResource()).Informer(),
 		beginSync:   atomic.NewBool(false),
 		initialSync: atomic.NewBool(false),
+		scope:       log.WithLabels("controller", identifier),
```
howardjohn, line 167:
```suggestion
		scope:       scope.WithLabels("controller", identifier),
```

otherwise we are in default scope instead of "kube"

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -245,13 +247,13 @@ func (cl *Client) Run(stop <-chan struct{}) {
 	})
 
 	cl.queue.Run(stop)
-	scope.Info("controller terminated")
+	cl.scope.Infof("%s:controller terminated")
```
howardjohn, line 250:
```suggestion
	cl.scope.Infof("controller terminated")
```

howardjohn (approved):
2 small nits, remove hold when ready. thanks!

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -95,16 +95,17 @@ type Client struct {
 	schemasByCRDName    map[string]collection.Schema
 	client              kube.Client
 	crdMetadataInformer cache.SharedIndexInformer
+	scope               *log.Scope
```
hzxuzhonghu, line 98: suggest s/scope/logger

Issue #40087, opened 2022-07-24
build failing on Mac


Pawan-Bishnoi:
### Bug Description

```
➜  istio git:(31fce3d756) ✗ make build
GOOS=darwin GOARCH=amd64 LDFLAGS='-extldflags -static -s -w' common/scripts/gobuild.sh /work/out/darwin_amd64/ ./istioctl/cmd/istioctl ./pilot/cmd/pilot-discovery ./pkg/test/echo/cmd/client ./pkg/test/echo/cmd/server ./samples/extauthz/cmd/extauthz ./operator/cmd/operator ./cni/cmd/istio-cni ./cni/cmd/istio-cni-taint ./cni/cmd/install-cni ./tools/istio-iptables ./tools/bug-report
go build github.com/containernetworking/plugins/pkg/ns: build constraints exclude all Go files in /go/pkg/mod/github.com/containernetworking/plugins@v1.1.1/pkg/ns

real	0m31.389s
user	0m38.665s
sys	0m13.751s
make[1]: *** [Makefile.core.mk:286: build] Error 1
make: *** [build] Error 2
```

It works when I go back to this commit:
```
➜  istio git:(31fce3d756) ✗ git checkout 5c25f8fe2d
Previous HEAD position was 31fce3d756 Make istio-cni compatible with Talos Linux (go netns implementation + added ca cert path) (#39699)
HEAD is now at 5c25f8fe2d dns: added DNSForwardParallel to support sending parallel queries to all nameservers (#39857)
➜  istio git:(5c25f8fe2d) ✗ make build
GOOS=darwin GOARCH=amd64 LDFLAGS='-extldflags -static -s -w' common/scripts/gobuild.sh /work/out/darwin_amd64/ ./istioctl/cmd/istioctl ./pilot/cmd/pilot-discovery ./pkg/test/echo/cmd/client ./pkg/test/echo/cmd/server ./samples/extauthz/cmd/extauthz ./operator/cmd/operator ./cni/cmd/istio-cni ./cni/cmd/istio-cni-taint ./cni/cmd/install-cni ./tools/istio-iptables ./tools/bug-report

real	0m31.818s
user	0m39.003s
sys	0m13.257s
GOOS=darwin GOARCH=amd64 LDFLAGS='-extldflags -static -s -w' common/scripts/gobuild.sh /work/out/darwin_amd64/ -tags=agent ./pilot/cmd/pilot-agent

real	0m3.303s
user	0m1.663s
sys	0m1.628s
```

[...]

PR #40343, opened 2022-08-08
Fix Mac build, moving linux specific code to its own files

ericvn:

Fixes: https://github.com/istio/istio/issues/40087
Fix the issue with the build failing on Macs


tools/istio-iptables/pkg/capture/run_linux.go
```diff
@@ -84,3 +85,21 @@ func configureTProxyRoutes(cfg *config.Config) error {
 	}
 	return nil
 }
+
+func nsContainerCode(cfg *config.Config) error {
```
ericvn, line 89: Might be able to come up with a better function name.

ericvn: Just moved whole function.

ericvn:
I do note that `make recommit` still fails on a Mac, but not in the pipeline, with:
```
...
INFO [runner] linters took 1.947750552s with stages: goanalysis_metalinter: 816.240192ms, structcheck: 72.705µs
cni/pkg/cmd/root.go:88:32: SA4023: this comparison is always true (staticcheck)
		if err = installer.Run(ctx); err != nil {
		                             ^
cni/pkg/install/install.go:80:22: SA4023(related information): (*istio.io/istio/cni/pkg/install.Installer).Run never returns a nil interface value (staticcheck)
                     ^
INFO File cache stats: 2 entries of total size 21.5KiB
...
```

howardjohn (approved):

GregHanson:
@ericvn does this need to be cherrypicked to 1.15?

ericvn:
Thanks for the catch @GregHanson

Issue #38440, opened 2022-04-19
[Multi-Cluster Discovery] Pilot Starts Regardless of Failure to Load Multi-Cluster Context


Minh-Ng:
### Bug Description

Background:
- Rolling out a canary version update from 1.10.3 (with XDS_PROXY disabled) to 1.12.4 (with XDS_PROXY enabled)
- Encountered a scenario where 1.12.4 control planes eventually get into a state where they push empty endpoints to proxies. Restarting Istiod-1-12-4 mitigates the issue but it is not clear how to discern a cluster that truly has no healthy endpoints from one that has healthy endpoints but is encountering a bug.
- Running a multi-cluster setup with one local cluster and ~20-40 remote clusters being managed and setup via istio-remote-secret-. It previously appeared that this but only occurs when network connectivity is lost between the control plane and remote api server. We then encountered this bug when remote cluster count via istiod_managed_clusters{cluster_type="remote"} remains steady over the past 7d window.
- 1.12.4 pilots running in the same cluster with the same remote secrets setup can have different state. Since this bug is non-deterministic we have observed one affected pilot with the same remote cluster count missing all endpoints for a given remote cluster in localhost:8080/debug/endpointz while another pilot with the same setup will return the expected endpoints from the remote cluster when checking the debug endpoint.
- We have a test workload that is restarted every 10 minutes and should generate an EDS PUSH as such; however, it seems that the remote cluster is no longer watched at all.

```
EDS: PUSH request for node:<POD_NAME>.<NAMESPACE_NAME> resources:200 size:<SIZE>kB empty:4 cached:<N-4>/<N>
```

Update:
It turns out this issue actually occurs on pilot start and is related to multi-cluster secret loading. Affected pilots are timing out while establishing communication to a remote API server and yet are still marked as healthy and never try to load the cluster context.
```
failed creating config configStore for cluster <CLUSTER_NAME> context deadline exceeded
```

This causes incorrect/asymmetric routes to be pushed to be pushed depending on which pilot a proxy connects to. It seems that a pilot should not be able to start without completely loading any multi-cluster configurations.

### Version

```prose
$ istioctl-1.12.4 version
client version: 1.12.4
istiod version: 1.12.4
$ kubectl version --short
Client Version: v1.21.0
Server Version: v1.20.15-gke.1000
```

Minh-Ng:
Found in 1.9.4 change notes:
```
[...]

PR #40347, opened 2022-08-09
retry remote clusters that fail to init

stevenctl:
https://github.com/istio/istio/issues/38440

This should help these clusters recover.

### DIFF
--- a/pkg/kube/multicluster/secretcontroller.go
+++ b/pkg/kube/multicluster/secretcontroller.go
@@ -366,6 +366,9 @@ func (c *Controller) addSecret(name types.NamespacedName, s *corev1.Secret) {
 		if err := callback(remoteCluster, remoteCluster.stop); err != nil {
 			remoteCluster.Stop()
 			log.Errorf("%s cluster_id from secret=%v: %s %v", action, clusterID, secretKey, err)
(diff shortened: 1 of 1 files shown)


stevenctl:
One worry I had is will this treat all the retries as "update" instead of "add". But after inspecting the code it seems like we will always treat things as add. @hzxuzhonghu 

We delete here

https://github.com/istio/istio/blob/53ae9db2d3595aa31d170d134701f6a05319fc3a/pkg/kube/multicluster/secretcontroller.go#L328-L334

then the check here never passes:

https://github.com/istio/istio/blob/8da4506000e267f80ddd926efa771c4bd7b0a8a0/pkg/kube/multicluster/secretcontroller.go#L343

hzxuzhonghu:
Should be very careful to be idempotent

hzxuzhonghu:
@stevenctl 

The delete only happens on those that exist in secret before, but after secret updated, the items are not there

Minh-Ng:
@stevenctl @hzxuzhonghu is this still relevant? seem like https://github.com/istio/istio/pull/40542 helped with retries but for extended periods of time Istiod will still never recover to the desired state without being restarted

Issue #39963, opened 2022-07-18
Istio init container  Init:CrashLoopBackOff for IPv6 only k8 clusters 1.21.x


abasitt:
### Bug Description

We are using Istio in Ipv6 only clusters and it's been working well till version 1.11.x on kubernetes version 1.21.5
I was planning to upgrade it 1.12.x but I am hitting an iptable error. It's currently happening on all Istio releases starting 1.12.x,1.13.x and 1.14.x.

- The controlplane has no issues and errors
- Ingress gateway installed successfully

The below error is for sidecar proxies on the init container

:~/istio-1.12.8# k get pods
NAME                     READY   STATUS                  RESTARTS   AGE
sleep-78cd8c94c9-mgv92   0/2     Init:0/1                0          2s
sleep-7ffd9f7c85-pxskj   0/2     Init:CrashLoopBackOff   1          19s
root@kube64m1:~/istio-1.12.8# k logs sleep-7ffd9f7c85-pxskj -c istio-init
2022-07-18T14:12:48.641911Z     info    Istio iptables environment:
ENVOY_PORT=
INBOUND_CAPTURE_PORT=
ISTIO_INBOUND_INTERCEPTION_MODE=
ISTIO_INBOUND_TPROXY_ROUTE_TABLE=
ISTIO_INBOUND_PORTS=
ISTIO_OUTBOUND_PORTS=
ISTIO_LOCAL_EXCLUDE_PORTS=
ISTIO_EXCLUDE_INTERFACES=
ISTIO_SERVICE_CIDR=
ISTIO_SERVICE_EXCLUDE_CIDR=
ISTIO_META_DNS_CAPTURE=true
INVALID_DROP=

2022-07-18T14:12:48.641940Z     info    Istio iptables variables:
PROXY_PORT=15001
PROXY_INBOUND_CAPTURE_PORT=15006
PROXY_TUNNEL_PORT=15008
PROXY_UID=1337
PROXY_GID=1337
INBOUND_INTERCEPTION_MODE=REDIRECT
INBOUND_TPROXY_MARK=1337
INBOUND_TPROXY_ROUTE_TABLE=133
INBOUND_PORTS_INCLUDE=*
INBOUND_PORTS_EXCLUDE=15090,15021,15020
OUTBOUND_IP_RANGES_INCLUDE=*
OUTBOUND_IP_RANGES_EXCLUDE=
OUTBOUND_PORTS_INCLUDE=
[...]

PR #40358, opened 2022-08-09
Fix bug(#39963) that causes istio-init container failure

ssuvasanth:
While configuring IPV6 address for loopback device, netlink.LinkByName("lo")
fails with "protocol not available" for Linux Kernel < 4.x.x.
(https://github.com/vishvananda/netlink/issues/791)
Running "ip -6" command to add lo addr if Kernel release is <4.

Signed-off-by: Vasanth Sundaravelu <vasanth.sundaravelu@rakuten.com>

Fix bug(#39963) that causes istio-init container failure

- [ ] Docs
-  [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/tools/istio-iptables/pkg/capture/run.go
+++ b/tools/istio-iptables/pkg/capture/run.go
@@ -269,10 +269,23 @@ func SplitV4V6(ips []string) (ipv4 []string, ipv6 []string) {
 // for IPv6 but not IPv4, as IPv4 defaults to `netmask 255.0.0.0`, which allows binding to addresses
 // in the 127.x.y.z range, while IPv6 defaults to `prefixlen 128` which allows binding only to ::1.
 // Equivalent to `ip -6 addr add "::6/128" dev lo`
-func configureIPv6Addresses(cfg *config.Config) error {
+func configureIPv6Addresses(cfg *config.Config, ext dep.Dependencies) error {
 	if !cfg.EnableInboundIPv6 {
 		return nil
 	}
+	// Issue: https://github.com/istio/istio/issues/39963
+	// netlink.LinkByName("lo") fails with "protocol not available" for Linux Kernel < 4.x.x
+	// (https://github.com/vishvananda/netlink/issues/791)
+	// Running "ip -6" command directly to add lo addr if Kernel release is <4.
+	KernelRelease := getKVersion()
+	if len(KernelRelease) > 0 && int(KernelRelease[0]) < 52 { // byte 52 is string 4 (ascii)
+		err := ext.Run(constants.IP, "-6", "addr", "add", "::6/128", "dev", "lo")
+		if err != nil {
+			return fmt.Errorf("failed to add 'lo' address: %s", err.Error())
+		}
+		return nil
+	}
+
 	link, err := netlink.LinkByName("lo")
 	if err != nil {
 		return fmt.Errorf("failed to find 'lo' link: %v", err)
--- a/tools/istio-iptables/pkg/capture/run_linux.go
+++ b/tools/istio-iptables/pkg/capture/run_linux.go
@@ -18,6 +18,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"syscall"
 
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/vishvananda/netlink"
@@ -107,7 +108,7 @@ func ConfigureRoutes(cfg *config.Config, ext dep.Dependencies) error {
 		defer nsContainer.Close()
 
 		return nsContainer.Do(func(ns.NetNS) error {
-			if err := configureIPv6Addresses(cfg); err != nil {
+			if err := configureIPv6Addresses(cfg, ext); err != nil {
 				return err
 			}
 			if err := configureTProxyRoutes(cfg); err != nil {
@@ -117,11 +118,26 @@ func ConfigureRoutes(cfg *config.Config, ext dep.Dependencies) error {
 		})
 	}
 	// called through 'nsenter -- istio-cni configure-routes'
-	if err := configureIPv6Addresses(cfg); err != nil {
+	if err := configureIPv6Addresses(cfg, ext); err != nil {
 		return err
 	}
 	if err := configureTProxyRoutes(cfg); err != nil {
 		return err
 	}
 	return nil
+}
+
+// Gets linux kernel version
+func getKVersion() []byte {
+	var uts syscall.Utsname
+	var release []byte
+	err := syscall.Uname(&uts)
+	if err != nil {
+		log.Errorf("Unable to get Linux Kernel version: %s", err.Error())
+		return []byte{}
+	}
+	for _, v := range uts.Release {
+		release = append(release, byte(v))
+	}
+	return release
 }
--- a/tools/istio-iptables/pkg/capture/run_unspecified.go
+++ b/tools/istio-iptables/pkg/capture/run_unspecified.go
@@ -35,3 +35,8 @@ func configureTProxyRoutes(cfg *config.Config) error {
 func ConfigureRoutes(cfg *config.Config, ext dep.Dependencies) error {
 	return ErrNotImplemented
 }
+
+// Gets kernel version - not implemented for unspecified OS
+func getKVersion() []byte {
+	return nil
+}
--- a/tools/istio-iptables/pkg/constants/constants.go
+++ b/tools/istio-iptables/pkg/constants/constants.go
@@ -172,4 +172,5 @@ const (
 
 const (
 	CommandConfigureRoutes = "configure-routes"
+	IP                     = "ip"
 )


hzxuzhonghu:
Can this be fixed in https://github.com/vishvananda/netlink?

ssuvasanth:
> Can this be fixed in https://github.com/vishvananda/netlink?

Hi @hzxuzhonghu,
Haven't got a chance to look into netlink implementation and no one else got a chance yet to look into the issue reported at [netlink repo](https://github.com/vishvananda/netlink/issues/791).
IMHO, IP linux util would be better tested for ipv6 with multiple Kernel versions, especially with Kernel API changes. And, network utils are generally updated as part of system update along with the Kernel by users and any incompatibility would be noticed at the earliest.
Please let me know your thoughts.

tools/istio-iptables/pkg/capture/run.go
```diff
@@ -307,10 +307,23 @@ func ConfigureRoutes(cfg *config.Config, ext dep.Dependencies) error {
 // for IPv6 but not IPv4, as IPv4 defaults to `netmask 255.0.0.0`, which allows binding to addresses
 // in the 127.x.y.z range, while IPv6 defaults to `prefixlen 128` which allows binding only to ::1.
 // Equivalent to `ip -6 addr add "::6/128" dev lo`
-func configureIPv6Addresses(cfg *config.Config) error {
+func configureIPv6Addresses(cfg *config.Config, ext dep.Dependencies) error {
 	if !cfg.EnableInboundIPv6 {
 		return nil
 	}
+	// Issue: https://github.com/istio/istio/issues/39963
+	// netlink.LinkByName("lo") fails with "protocol not available" for Linux Kernel < 4.x.x
```
howardjohn, line 315: Ah its only old kernel versions? TBH I don't feel great about adding support for 10 year old OS's with Istio alpha features (IPv6)...

If you are going to use cutting edge features on one product, expecting them to pair with ancient OS's is a bit tricky.

howardjohn: We are introducing some complexity, runtime cost, and risk for all users to appease a small set of users using extremely old software

ssuvasanth: @howardjohn RHEL 7 / Centos 7 / etc are not end of life yet and in fact would be running the majority of the production  Linux workloads. RHEL 7 only ships 3.10.x Kernels and started using 4.4.x only from RHEL8.

This issue and fix was tested on the following Kernel which was less than 2-3 months old.

`[vas@vas-istio-dev istio]$ uname -a
Linux vas-istio-dev 3.10.0-1160.66.1.el7.x86_64 #1 SMP Wed May 18 16:02:34 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
[vas@vas-istio-dev istio]$`

ssuvasanth: @howardjohn , Additionally, the issue may not have been seen by a large userbase because this issue is seen only on IPv6 deployments.

ssuvasanth:
Hi @hzxuzhonghu @howardjohn , please let me know if there are any further thoughts.

tools/istio-iptables/pkg/capture/run.go
hzxuzhonghu: There are many vms running even 2.6 kernels as I know

ssuvasanth: I agree. IMHO, we should at minimum support the current Kernels in the RHEL7/Centos7 world. (3.10.x)
I have rebased the commit and pushed again.
Can I close this review as resolved? Can you please approve if its okay? @hzxuzhonghu @howardjohn 
Thanks in advance.

howardjohn: IMO we should fix upstream. This is dangerous.

How do I reproduce? I tried running centos 7, works fine...

Can you try to modify https://github.com/vishvananda/netlink/blob/afa2eb2a66aac1f8f370287f236ba93d4c078dd6/link_linux.go#L1662 to be unix.AF_INET6 and see if it works?

howardjohn (changes requested):

tools/istio-iptables/pkg/capture/run.go
ssuvasanth: @howardjohn ,
Humbly requesting to explain the dangerous part in the fix. Is there any disadvantage in using the native 'ip' command? Please help me understand if I am missing something.

IMO - native 'ip' command is robust, reliable and without compatibility issues.
Like how having IPv6 enabled in the Kernel, IP6tables modules loaded in the Kernel, etc are external dependencies...similarly, having non-trivial action which requires admin privilege like assigning IP address is an external dependency, IMO, and should be handled using native methods/commands. Please correct me and help me with an explanation if I am wrong.
Implementing a part of the capabilities of native 'ip' command internally would convert the external dependency into an internal dependency on the Kernel API/release.

For reproducing the issue, please try to inject Istio sidecar (version > 1.11 where netlink was introduced) on K8s cluster with only IPv6 enabled. Issue is seen on both istio-init and istio-cni modes.

Thanks in advance.

howardjohn: `ip` command does not exist on all machines. This PR also introduces risk for newer kernels, etc if the detection logic is incorrect or slow, etc.

ssuvasanth: 1. AFAIK, `ip` command is installed even as part of the minimal installations in case of adding sidecar to VMs/baremetal/etc. If considering container images, then we require `ip` command only inside `proxyv2` images where it is already present.
2. IMO, using native `ip` command is less riskier compared to netlink in case of newer Kernels, Kernel API changes, compatability, etc.. Userbase for Linux+ip would be of substantial amount. 
3. Please let me know your concerns about the detection logic, I will try to address them. So far, I have used the `syscall` package which is used by many built-in packages such as `net` and I am calling `uname` to get the Kernel info and uname has been around for ages in Linux and do not expect this to change (AFAIK). Even if it does, Linux usually handles backward compatibilities very well given `uname` exists like forever. 
4. For latency concerns, we may try to measure as the logic is straight forward (even string operations are avoided for latency reasons).  I would like to discuss specific concerns if any so that we can address them. 

Looking forward to collaborating in taking this forward.

howardjohn:
```
$ docker run --entrypoint ip gcr.io/istio-testing/proxyv2:latest-distroless
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "ip": executable file not found in $PATH: unknown.
```

Distroless images (security best practice) do not have it

ssuvasanth: ah.. okay... distroless image doesn't have them. The other proxyv2 images has it.
How was the distroless image handled in 1.11.x and before where `ip` command was used instead of netlink?

howardjohn:
It used to have it, we removed it since it is a security risk (and bloat to
image)

ssuvasanth:
https://github.com/istio/istio/pull/40358#issuecomment-1212442022 :

okay, understood. But we still have iptables commands in disrtoless and do you have info on the security aspects/risks of having `ip` command?

The concern I still have is that we are trying to implement a part of `ip` utility within the product which sounds less than ideal as `ip` is already well done, reliable and no compatibility issue with future Kernels. IMO, this sounds similar to implementing a new ssh client or tcp stack or other well established Linux primitives because we want to remove the native implementations.

Let me know if you have any thoughts on ways forward.

howardjohn:
We are not replacing primitives - we are using the same kernel functions
with a native implementation that is orders of magnitude simpler (binary
size and security). I would do the same with iptables binary as well if I
could but it's a bigger effort.

Did you have a chance to try the fix I suggested?

ssuvasanth:
I will try them tomorrow and let you know.

ssuvasanth:
@howardjohn , noticed the following behaviour in different istio : netlink versions:

```
master - issue not seen 				github.com/vishvananda/netlink v1.2.1-beta.2
1.13.8 - issue not seen			github.com/vishvananda/netlink v1.1.1-0.20210330154013-f5de75959ad5
1.14.4 - issue reproducible     	github.com/vishvananda/netlink v1.2.0-beta
```
Trying to find 'lo' link using netlink in a standalone binary also shows that the issue in netlink is not seen in netlink versions v.1.1 or v.1.2.1-beta.2.
Issue is seen in netlink 1.2.0-beta which is used in 1.14.4

May I know what is the dependency version upgrade strategy / vendoring strategy?
Will the netlink module version in go.mod bumped automatically by some pipeline for 1.14.4 too?
Why are we using a module's beta version in istio? Shouldn't we stick to prod releases and stick to tested versions?

Let me know your thoughts.

tools/istio-iptables/pkg/capture/run.go
howardjohn: We used a fork of distoless with `ip` added

ssuvasanth: https://github.com/istio/istio/pull/40358#issuecomment-1244211009

howardjohn:
Using this module's non-prod release is standard in the ecosystem as the last release was 2 years ago: https://github.com/containernetworking/plugins/blob/main/go.mod#L22.

1.14 will not be updated.

PR #40360, opened 2022-08-09
chore: do not collapsed default dashboard panels

Xunzhuo:
Signed-off-by: Xunzhuo <mixdeers@gmail.com>

 do not collapsed default dashboard panels

image


howardjohn:
I believe it was to lazy load expensive pages. Not sure though

Xunzhuo:
> I believe it was to lazy load expensive pages. Not sure though

In my view, it is not set for this, and it makes users hard to look through the service/workload dashboards. Every time user tries to check these two dashboards, they need to open the folders.

PR #40370, opened 2022-08-09
This fixes `make precommit` failing on a Mac

ericvn:

This fixes a failure in `make precommit` on a Mac.

Currently this yields an error:
```
...
INFO [runner] linters took 4.100221257s with stages: goanalysis_metalinter: 2.83420288s, structcheck: 113.212µs
cni/pkg/cmd/root.go:88:32: SA4023: this comparison is always true (staticcheck)
		if err = installer.Run(ctx); err != nil {
		                             ^
cni/pkg/install/install.go:80:22: SA4023(related information): (*istio.io/istio/cni/pkg/install.Installer).Run never returns a nil interface value (staticcheck)
                     ^
...
```

I'm not sure why the pipeline doesn't see the same issue since the same build-tools image should be used in both places.

### DIFF
--- a/cni/pkg/cmd/root.go
+++ b/cni/pkg/cmd/root.go
@@ -85,14 +85,13 @@ var rootCmd = &cobra.Command{
 
 		repair.StartRepair(ctx, &cfg.RepairConfig)
 
-		if err = installer.Run(ctx); err != nil {
-			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-				log.Infof("Installer exits with %v", err)
-				// Error was caused by interrupt/termination signal
-				err = nil
-			} else {
-				log.Errorf("Installer exits with %v", err)
-			}
+		err = installer.Run(ctx)
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			log.Infof("Installer exits with %v", err)
+			// Error was caused by interrupt/termination signal
+			err = nil
+		} else {
+			log.Errorf("Installer exits with %v", err)
 		}
 
 		if cleanErr := installer.Cleanup(); cleanErr != nil {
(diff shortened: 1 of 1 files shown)


ericvn:
DNM will I verify the initial problem exists on another Mac.

cni/pkg/cmd/root.go
```diff
@@ -85,14 +85,13 @@ var rootCmd = &cobra.Command{
 
 		repair.StartRepair(ctx, &cfg.RepairConfig)
 
-		if err = installer.Run(ctx); err != nil {
```
howardjohn, line 88: `if err := installer.Run(ctx); err != nil {`

does that fix instead?

ericvn: That's the original code which fails. I believe the linter was looking at Run and sees that it would never return an err that was nil, so decided there is no reason for the nil check of the returned value.

howardjohn: the `:=` vs `=` was the difference here. Shadowing here seems concerning, I feel like I am not understanding something

ericvn: Ahh, missed the `:`. If I try that I get:
```
cni/pkg/cmd/root.go:92:5: ineffectual assignment to err (ineffassign)
				err = nil
```
I believe the main issue is Run uses a passed in `err`:
```
```

ericvn: I did restructure the Run to just have error and then used := for the err assignments in the function nd changed the return's to be return err, and that also passes the lint check.

howardjohn (approved):

howardjohn:
do we know why linux doesn't have this issue?

ericvn:
I figured out if I clear the Docker volumes I can get `make lint` to pass locally. I expect something was in the GoCache that was causing the failure, but not sure what. Noting it takes 6 times longer to run with a clean cache. Still worked a second time with the new cache.

Maybe add something in the wiki mentioning this. I have a todo to do some dev updates to the wiki.

PR #40373, opened 2022-08-10
loose telemetry accesslogging validation 

zirain:

after https://github.com/istio/istio/pull/39521  we should loose telemetry accesslogging validation


zirain:
> /cherrypick release-1.14

this's a new feature, and should not backport to v1.14

GregHanson:
@zirain you are right, sorry.  I thought I remembered it going in but verified that is not the case

hzxuzhonghu (approved):
Does it need to be in 1.15

zirain:
> Does it need to be in 1.15

yes, v1.15 is still alpha.

PR #40385, opened 2022-08-10
Fix API group in distribution report configmap ownerReference

luksa:
Prior to this commit, the status reporter tried to create an ownerReference with an invalid API group, causing the following error:

Error writing Distribution Report: unable to create ConfigMap: configmaps "istiod-78c5bd88d-72m8k-distribution" is forbidden: cannot set blockOwnerDeletion in this case because cannot find RESTMapping for APIVersion meta.k8s.io/v1 Kind Pod: no matches for kind "Pod" in version "meta.k8s.io/v1"

This commit fixes the API group/version to "v1"


luksa:
@jwendell what's the label for cherrypicking this to the previous versions?

howardjohn (approved):

howardjohn:
/ok-to-test

Added hold to allow the feature own @therealmitchconnors to look as well

PR #40386, opened 2022-08-10
kube client: use disk cache for CLI tools

howardjohn:
When using istioctl on slow api-servers, things are super slow. The
issue is *every* call does API server discovery. This process is super
slow.

To workaround this slowness, k8s has caching. Today, we use an in-memory
cache. This is useless for CLI tools, which only run it once. Another
cache exists, which is used by kubectl, to cache to disk.

This PR introduces a disk based cache. This is enabled only for the
`NewExtendedClient`. To make it more explicit this should only be used
for CLI tools (and similar), the function is renamed to `NewCLIClient`.
Note all current cases of this function are correctly used in CLIs, so
this is just a renaming.

Additionally, add a new `pkg/lazy` package. This facilitates lazily
computing values, similar to `sync.Once` but more ergonomic and safer
(ability to retry, etc).

```
$ hyperfine -i '~/istio/istio-1.15.0-beta.0/bin/istioctl ps svc/fake' 'istioctl ps svc/fake'
Benchmark 1: ~/istio/istio-1.15.0-beta.0/bin/istioctl ps svc/fake
  Time (mean ± σ):      1.222 s ±  0.361 s    [User: 0.264 s, System: 0.066 s]
  Range (min … max):    0.801 s …  1.940 s    10 runs

  Warning: Ignoring non-zero exit code.

Benchmark 2: istioctl ps svc/fake
  Time (mean ± σ):     620.5 ms ±  78.8 ms    [User: 256.9 ms, System: 77.0 ms]
  Range (min … max):   514.7 ms … 782.9 ms    10 runs

  Warning: Ignoring non-zero exit code.

Summary
  'istioctl ps svc/fake' ran
    1.97 ± 0.63 times faster than '~/istio/istio-1.15.0-beta.0/bin/istioctl ps svc/fake'
```


litong01:
@howardjohn john, is there a location where the cache will go? is it configurable to use it or not to use it? If this improvement is added, then consecutive runs of istioctl command would benefit or even one istioctl call (many cases we make multiple k8s API calls in one istioctl command) also benefit? Thanks.

howardjohn:
> john, is there a location where the cache will go? is it configurable to use it or not to use it? If this improvement is added, then consecutive runs of istioctl command would benefit or even one istioctl call (many cases we make multiple k8s API calls in one istioctl command) also benefit? Thanks.

~/.kube/cache like kubectl.

Even a single istioctl call will benefit because users may have previously run `kubectl`

pkg/kube/client_factory.go
```diff
@@ -70,22 +71,30 @@ func newClientFactory(clientConfig clientcmd.ClientConfig, diskCache bool) util.
 			discoveryCacheDir := computeDiscoverCacheDir(filepath.Join(cacheDir, "discovery"), restConfig.Host)
 
 			return diskcached.NewCachedDiscoveryClientForConfig(restConfig, discoveryCacheDir, httpCacheDir, 6*time.Hour)
```
litong01, line 73: @howardjohn how do we know here that the cached has expired? is 6 hours hard coded expiration time? Thanks.

howardjohn: 6hrs is the value used by kubectl. There is some automatic logic to evict the cache on miss, its not just a static 6 hours.

The logic here makes istioctl identical to kubectl. It also makes it identical to the recommended way to write CLI tools (using https://pkg.go.dev/k8s.io/cli-runtime/pkg/genericclioptions)

ericvn (approved):
Looks good. I was curious what lead to the initial test failures.

pkg/lazy/lazy.go
```diff
@@ -0,0 +1,78 @@
...
+		// Outlined slow-path to allow inlining of the fast-path.
+		return l.doSlow()
+	}
+	return l.res, l.err
+}
+
+func (l *lazyImpl[T]) doSlow() (T, error) {
+	l.m.Lock()
+	defer l.m.Unlock()
+	if l.done == 0 {
```
nmittler, line 63: I guess you're relying on the lock forcing a sync across all threads, so you don't have to do an atomic get?

pkg/lazy/lazy.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Lazy represents a value whose computation is deferred until the first access
+type Lazy[T any] interface {
+	// Get returns the value, computing it if necessary.
+	Get() (T, error)
+}
+
+type lazyImpl[T any] struct {
+	getter func() (T, error)
+	retry  bool
```
nmittler, line 30: retry indefinitely?

pkg/lazy/lazy.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// NewWithRetry returns a new lazily computed value. The value will be computed on each call until a
+// non-nil error is returned.
+func NewWithRetry[T any](f func() (T, error)) Lazy[T] {
+	return &lazyImpl[T]{getter: f, retry: true}
+}
+
+func (l *lazyImpl[T]) Get() (T, error) {
+	if atomic.LoadUint32(&l.done) == 0 {
```
nmittler, line 53: nit: maybe make a helper method `l.isDone`?

pkg/kube/client_factory.go
```diff
@@ -38,21 +45,57 @@ type clientFactory struct {
...
+			return nil, err
+		}
+		// Setup cached discovery. CLIs uses disk cache, controllers use memory cache.
+		if diskCache {
+			cacheDir := filepath.Join(homedir.HomeDir(), ".kube", "cache")
+
+			httpCacheDir := filepath.Join(cacheDir, "http")
+			discoveryCacheDir := computeDiscoverCacheDir(filepath.Join(cacheDir, "discovery"), restConfig.Host)
+
+			return diskcached.NewCachedDiscoveryClientForConfig(restConfig, discoveryCacheDir, httpCacheDir, 6*time.Hour)
```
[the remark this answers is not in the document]
howardjohn: No constant in k8s. The code is copy+pasted. I added a link to the source

pkg/kube/client_factory.go
```diff
@@ -38,21 +45,57 @@ type clientFactory struct {
...
 
 // newClientFactory creates a new util.Factory from the given clientcmd.ClientConfig.
-func newClientFactory(clientConfig clientcmd.ClientConfig) util.Factory {
+func newClientFactory(clientConfig clientcmd.ClientConfig, diskCache bool) *clientFactory {
 	out := &clientFactory{
 		clientConfig: clientConfig,
 	}
 
 	out.factory = util.NewFactory(out)
+	out.discoveryClient = lazy.NewWithRetry(func() (discovery.CachedDiscoveryInterface, error) {
```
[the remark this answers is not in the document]
howardjohn: The code in K8s is pretty buried so it seems hard to extract. Most is copy pasted; I added a link to the source

pkg/lazy/lazy.go
howardjohn: until non-nil error; I added more comments

pkg/lazy/lazy.go
howardjohn: Explained here: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/sync/once.go;l=49-53 . This code is mostly copy+pasted with minor tweaks to support our needs. I thought I had commented that but it seems not -- I updated the comments a bit

pkg/lazy/lazy.go
howardjohn: Its intentionally this way to make sure its inlined so that when we have already computed the value its fast

nmittler (approved):

costinm (approved):

ingwonsong:
@howardjohn Would you mind if I cherrypick this PR to release-1.15?

howardjohn:
sure

PR #40398, opened 2022-08-11
add queue metrics

ramaraochavali:
Adds metrics for Queue to easily diagnose queue issues.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pkg/queue/instance.go
```diff
@@ -138,23 +185,26 @@ func (q *queueImpl) Run(stop <-chan struct{}) {
...
 			return
 		}
 
 		backoffTask := q.tasks[0]
+		active.With(idTag.Value(q.id)).Increment()
 		// Slicing will not free the underlying elements of the array, so explicitly clear them out here
 		q.tasks[0] = nil
 		q.tasks = q.tasks[1:]
-
+		size.With(idTag.Value(q.id)).RecordInt(int64(len(q.tasks)))
```
howardjohn, line 204: can we precompute all the `size.With(idTag.Value(q.id))` etc parts?

Without it performance is terrible - its really expensive to make new tags

```
name             old time/op    new time/op    delta
Queue/prerun-8     80.5µs ± 2%  5683.7µs ±18%   +6958.66%  (p=0.002 n=8+5)
Queue/postrun-8     104µs ±20%   11186µs ±48%  +10681.94%  (p=0.001 n=10+5)

name             old alloc/op   new alloc/op   delta
Queue/prerun-8     58.3kB ± 0%  5402.9kB ± 0%   +9161.42%  (p=0.003 n=7+5)
Queue/postrun-8    58.4kB ± 0%  5403.6kB ± 0%   +9154.25%  (p=0.002 n=10+4)

name             old allocs/op  new allocs/op  delta
Queue/prerun-8      2.03k ± 0%   100.05k ± 0%   +4833.43%  (p=0.001 n=9+5)
Queue/postrun-8     2.03k ± 0%   100.05k ± 0%   +4833.35%  (p=0.002 n=10+4)
```

hzxuzhonghu:
As i commented here https://github.com/istio/istio/pull/40406#discussion_r944047676

pkg/queue/instance.go
ramaraochavali: I think we can let Queue creator pass these metrics predefined and we can define an interface for this. I will make it that way.

ramaraochavali: Even if we precompute, there is still overhead of metric updates

With and Without metrics
```
goos: darwin
goarch: arm64
pkg: istio.io/istio/pkg/queue
BenchmarkQueue
BenchmarkQueue-10    	     532	   2346046 ns/op	   55582 B/op	    2078 allocs/op
PASS
ok  	istio.io/istio/pkg/queue	1.733s

goos: darwin
goarch: arm64
pkg: istio.io/istio/pkg/queue
BenchmarkQueue
BenchmarkQueue-10    	     171	   7031024 ns/op	  603219 B/op	   10082 allocs/op
PASS
ok  	istio.io/istio/pkg/queue	2.179s
```

So I controlled it with a feature flag so that it can only enabled for debugging.

ramaraochavali:
Closing this as without additional types this may not be much useful. Just looked at client-go it provides metrics though. We can enable it when we needed.

Issue #39828, opened 2022-07-07
Test target uses binaries path from host pwd despite mounting to /work


emike922:
### Bug Description

Running `make build test`, expecting to build and test in docker, fails.
The following warning implies to me that LOCAL_OUT is wrongly set to a path on the docker host file system:

```
2022-07-07T14:34:51.414247Z	warn	unable to resolve LOCAL_OUT. Dir /var/jenkins/workspace/istio-pre-cr/out/linux_amd64 does not exist
```

To confirm:
```
$ make shell
build-tools:/work$ printenv LOCAL_OUT
/var/jenkins/workspace/istio-pre-cr/out/linux_amd64
build-tools:/work$ ls -lh out/linux_amd64/
total 794M
-rwxr-xr-x 1 user docker  76M Jul  7 16:43 bug-report
-rw-r--r-- 1 user docker 3.1K Jul  7 16:35 cdsv2_sidecar.json
-rwxr-xr-x 1 user docker  27M Jul  7 16:43 client
-rwxr-xr-x 1 user docker  90M Jul  7 16:34 envoy
...
```

The same instructions work fine when overriding `TARGET_OUT_LINUX` using `TARGET_OUT_LINUX=/work/out/linux_amd64 make build test`

### Version

```prose
Istio 1.14.1
```

### Additional Information

Machine is running Ubuntu 18.04.6 LTS and has both docker and make

ericvn:
I'm trying to understand your problem and have been unable to recreate. 
Running `make build test` does not throw any immediate errors for me on the current main, nor the release-1.14 branch (even using the 1.14.1 tag).

LOCAL_OUT is an environment variable that points to a location on the host, not within the container, hence the prefix LOCAL. The CONTAINER prefix would indicate a location within the container. 

There was some work done in this space in https://github.com/istio/istio/pull/37992 so maybe the commit you are on might be earlier. I do note that before this PR, LOCAL_OUT is not set inside the SHELL, which does match what you are posting. 

Would you confirm the commit you are seeing the issue on.

emike922:
I'm afraid it's still occurring on the latest commit on main branch. This time I used a local VM running Ubuntu 20.04.4..

```
$ git clone https://github.com/istio/istio.git
$ cd istio
$ git log -1
commit 9c5f437581c53b336ee22838f2273a0d3cffd98a (HEAD -> master, origin/master, origin/HEAD)
Author: Istio Automation <istio-testing-bot@google.com>
[...]

PR #40413, opened 2022-08-11
Update envoy and pilot tests to use TARGET_OUT instead of LOCAL_OUT

ericvn:

Looking to fix https://github.com/istio/istio/issues/39828

A particular test istio.io/istio/pkg/dns/server TestNewWithDuplicateOptionsShouldFail is failing as the binary.go code is using LOCAL_OUT to try and find the binaries (https://github.com/istio/istio/blob/master/pkg/test/envoy/binary.go#L61). This fails when run inside a container as LOCAL_OUT is pointing to the host directory paths.

We don't see a failure in the pipeline as the pipeline does not use a container for testing (BUILD_WITH_CONTAINER=0).

This updates the test to point at TARGET_OUT (which should be the same as LOCAL_OUT when run without a container), and set to the container directories inside a container.

### DIFF
--- a/common/scripts/setup_env.sh
+++ b/common/scripts/setup_env.sh
@@ -176,16 +176,13 @@ fi
 # This is used when we need to run a build artifact during tests or later as part of another
 # target.
 if [[ "${FOR_BUILD_CONTAINER:-0}" -eq "1" ]]; then
-  LOCAL_OUT="${TARGET_OUT_LINUX}"
-else
-  LOCAL_OUT="${TARGET_OUT}"
-fi
-
-if [[ "${FOR_BUILD_CONTAINER:-0}" -eq "1" ]]; then
   # Override variables with container specific
   TARGET_OUT=${CONTAINER_TARGET_OUT}
   TARGET_OUT_LINUX=${CONTAINER_TARGET_OUT_LINUX}
+  LOCAL_OUT="${TARGET_OUT_LINUX}"
   REPO_ROOT=/work
+else
+  LOCAL_OUT="${TARGET_OUT}"
 fi
 
 go_os_arch=${LOCAL_OUT##*/}
--- a/common/scripts/setup_env.sh
+++ b/common/scripts/setup_env.sh
@@ -176,13 +176,16 @@ fi
 # This is used when we need to run a build artifact during tests or later as part of another
 # target.
 if [[ "${FOR_BUILD_CONTAINER:-0}" -eq "1" ]]; then
+  LOCAL_OUT="${TARGET_OUT_LINUX}"
+else
+  LOCAL_OUT="${TARGET_OUT}"
+fi
+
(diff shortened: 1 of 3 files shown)


common/scripts/setup_env.sh
```diff
@@ -175,17 +175,14 @@ fi
...
-  LOCAL_OUT="${TARGET_OUT_LINUX}"
-else
-  LOCAL_OUT="${TARGET_OUT}"
-fi
-
 if [[ "${FOR_BUILD_CONTAINER:-0}" -eq "1" ]]; then
   # Override variables with container specific
   TARGET_OUT=${CONTAINER_TARGET_OUT}
   TARGET_OUT_LINUX=${CONTAINER_TARGET_OUT_LINUX}
+  LOCAL_OUT="${TARGET_OUT_LINUX}"
```
ericvn, line 182: This changes LOCAL_OUT to point to CONTAINER_TARGET_OUT_LINUX which is a location within the container instead of a location on the host. In this case we are running in a Linux container so that is correct. It will also point to a directory within the contianer.

ericvn:
It may be that we simply need to update the binary.go to use something other than LOCAL_OUT, maybe TARGET_OUT? Need to look the binary.go code closer.

common/scripts/setup_env.sh
ericvn: As I think about this more, LOCAL_OUT has and still needs to point the the LOCAL host structure, else it will have an adverse affect on the `docker` script.

ericvn:
The reason unit tests work in the pipeline is they aren't run within the container (BUILD_WITH_CONTAINER=0). in that case, LOCAL_OUT is set to TARGET_OUT_LINUX which points to something that is accessible.

ericvn:
With commit 1, all tests passed (didn't wait on integ-security-multicluster_istio).
Now try with the binary.go change.

pkg/test/envoy/binary.go
```diff
@@ -58,7 +58,7 @@ func FindBinaryOrFail(t test.Failer) string {
 
 func findBinaries() ([]string, error) {
 	binPaths := make([]string, 0)
-	err := filepath.Walk(env.LocalOut, func(path string, f os.FileInfo, err error) error {
+	err := filepath.Walk(env.IstioOut, func(path string, f os.FileInfo, err error) error {
```
ericvn, line 61: Note: IstioOut is set to the TARGET_OUT env variable.

ericvn:
Have verified that the failure is not happening with this fix using a build container on Linux and Mac.

ericvn:
DNM as looking at some other test failures as well.

ericvn:
One issue with the current change is that TARGET_OUT on a Mac points to the Mac Envoy, which will fail to run inside the container (wrong format binary).

howardjohn (approved):

ericvn:
Closing this PR in favor of https://github.com/istio/common-files/pull/655 which includes the first commit of this PR.

PR #40441, opened 2022-08-13
refactor get and process functionality of queue

ramaraochavali:
To better understand and segregate get and process functions of queue

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/queue/instance.go
+++ b/pkg/queue/instance.go
@@ -117,6 +117,48 @@ func (q *queueImpl) Closed() <-chan struct{} {
 	return q.closed
 }
 
+// Get blocks until it can return a task to be processed. If shutdown = true,
+// the processing go routine should stop.
+func (q *queueImpl) get() (task *BackoffTask, shutdown bool) {
+	q.cond.L.Lock()
+	defer q.cond.L.Unlock()
+	// wait for closing to be set, or a task to be pushed
+	for !q.closing && len(q.tasks) == 0 {
+		q.cond.Wait()
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/queue/instance.go
```diff
@@ -117,6 +117,48 @@ func (q *queueImpl) Closed() <-chan struct{} {
...
+// the processing go routine should stop.
+func (q *queueImpl) get() (task *BackoffTask, shutdown bool) {
+	q.cond.L.Lock()
+	defer q.cond.L.Unlock()
+	// wait for closing to be set, or a task to be pushed
+	for !q.closing && len(q.tasks) == 0 {
+		q.cond.Wait()
+	}
+
+	if len(q.tasks) == 0 {
```
hzxuzhonghu, line 130: This is not equal to previous behavior

ramaraochavali: I can change to previous condition but it is same behaviour. Wait only comes out if len(q.tasks) is zero

hzxuzhonghu: Previously, if queue is stopped, no tasks will be run if there is any left unprocessed.

ramaraochavali: ok. any way I changed to previous condition.

hzxuzhonghu (commented):
LGTM, but need more eyes on it.

howardjohn (approved):
LGTM, seems just a refactor with no behavior change

PR #40449, opened 2022-08-15
move maxstream duration creation inside condition

ramaraochavali:
#40320 has moved the maxstream duration condition outside of condition which if it is not used assumes some defaults at HCM. this PR moves it inside.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/route/route.go
+++ b/pilot/pkg/networking/core/v1alpha3/route/route.go
@@ -1094,15 +1094,18 @@ func setTimeout(action *route.RouteAction, vsTimeout *duration.Duration, node *m
 	if vsTimeout != nil {
 		action.Timeout = vsTimeout
 	}
-	action.MaxStreamDuration = &route.RouteAction_MaxStreamDuration{}
 	if node != nil && node.IsProxylessGrpc() {
 		// TODO(stevenctl) merge these paths; grpc's xDS impl will not read the deprecated value
-		action.MaxStreamDuration.MaxStreamDuration = action.Timeout
+		action.MaxStreamDuration = &route.RouteAction_MaxStreamDuration{
+			MaxStreamDuration: action.Timeout,
+		}
 	} else {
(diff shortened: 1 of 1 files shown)


pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -1094,15 +1094,18 @@ func setTimeout(action *route.RouteAction, vsTimeout *duration.Duration, node *m
...
+			MaxStreamDuration: action.Timeout,
+		}
 	} else {
 		// Set MaxStreamDuration only for notimeout cases otherwise it wont be honored.
 		if action.Timeout.AsDuration().Nanoseconds() == 0 {
-			action.MaxStreamDuration.MaxStreamDuration = action.Timeout
-			action.MaxStreamDuration.GrpcTimeoutHeaderMax = action.Timeout
+			action.MaxStreamDuration = &route.RouteAction_MaxStreamDuration{
+				MaxStreamDuration:    notimeout,
+				GrpcTimeoutHeaderMax: notimeout,
```
hzxuzhonghu, line 1107: not sure i understand

ramaraochavali: earlier this `action.MaxStreamDuration &route.RouteAction_MaxStreamDuration{}` was set for all cases even if max_stream_duration is not set, for example if VS has timeout. So that assumes some defaults from Envoy like HCM max stream duration because the top level struct is set,.

hzxuzhonghu: // Use deprecated value for now as the replacement MaxStreamDuration has some regressions.

It does not apply to notimeout?

ramaraochavali: it does not apply. Only when we have value at route timeout, it messes up.

hzxuzhonghu (approved):

PR #40451, opened 2022-08-15
move xds cache metrics to interface

ramaraochavali:
To simplify if else conditions and only register metrics if they are enabled.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/xds_cache.go
+++ b/pilot/pkg/model/xds_cache.go
@@ -30,57 +30,96 @@ import (
 	"istio.io/pkg/monitoring"
 )
 
-func init() {
-	monitoring.MustRegister(xdsCacheReads)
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
This is fine but I wonder if we should build it directly into `monitoring.Metric` to have a `disabled` mode. Since this is a fairly common pattern.

Something like `monitoring.RegistryIf(func() bool)` possibly

WDYT?

ramaraochavali:
> Something like monitoring.RegistryIf(func() bool) possibly

Makes sense. Let me change.

ramaraochavali:
I will close this and open a separate PR in that repo first.

PR #40452, opened 2022-08-15
CA client tls config ServerName first set to CAEndpoint host

hzxuzhonghu:

Fix when CAEndpoint uses a different address, the ServerName will be empty

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/istio-agent/xds_proxy.go
+++ b/pkg/istio-agent/xds_proxy.go
@@ -722,13 +722,13 @@ func (p *XdsProxy) getTLSDialOption(agent *Agent) (grpc.DialOption, error) {
 			}
 			return &certificate, nil
 		},
-		RootCAs: rootCert,
+		RootCAs:    rootCert,
+		MinVersion: tls.VersionTLS12,
 	}
 
-	// strip the port from the address
-	parts := strings.Split(agent.proxyConfig.DiscoveryAddress, ":")
-	config.ServerName = parts[0]
-
+	if host, _, err := net.SplitHostPort(agent.proxyConfig.DiscoveryAddress); err != nil {
+		config.ServerName = host
+	}
 	// For debugging on localhost (with port forward)
 	// This matches the logic for the CA; this code should eventually be shared
 	if strings.Contains(config.ServerName, "localhost") {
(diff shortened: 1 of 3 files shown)


pkg/istio-agent/xds_proxy.go
```diff
@@ -728,19 +728,11 @@ func (p *XdsProxy) getTLSDialOption(agent *Agent) (grpc.DialOption, error) {
...
-	if strings.Contains(config.ServerName, "localhost") {
-		config.ServerName = "istiod.istio-system.svc"
-	}
-
 	if p.istiodSAN != "" {
 		config.ServerName = p.istiodSAN
 	}
 	// TODO: if istiodSAN starts with spiffe://, use custom validation.
 
-	config.MinVersion = tls.VersionTLS12
```
howardjohn, line 743: why remove this? because it is default?

howardjohn (changes requested):
This breaks https://github.com/howardjohn/local-istio-development and doesn't allow running proxy locally.

pkg/istio-agent/xds_proxy.go
lei-tang: I second that this line should not be removed. It is more straightforward and less error-prone to explicitly configure the min version. Thus, in case the default in Envoy change, the Istio behavior remains stable and unchanged.

lei-tang (changes requested):

pkg/istio-agent/xds_proxy.go
hzxuzhonghu: Yes

hzxuzhonghu:
>This breaks https://github.com/howardjohn/local-istio-development and doesn't allow running proxy locally 

IMO, for localhost debug, it is better not to leave an empty tlsconfig

hzxuzhonghu:
@howardjohn @lei-tang Reverted all change. And only try to fix missing to set ServerName when CAEndpoint is specified.

security/pkg/nodeagent/caclient/providers/citadel/client.go
```diff
@@ -150,6 +150,9 @@ func (c *CitadelClient) getTLSDialOption() (grpc.DialOption, error) {
 		RootCAs: certPool,
 	}
 
+	// strip the port from the address
+	parts := strings.Split(c.opts.CAEndpoint, ":")
```
howardjohn, line 154: net.SplitHostPort may be more appropriate - this could fail on an ipv6 address for example.
Setting this when it's an IP may fail in general actually? not certain

hzxuzhonghu: Right, i will fix the xds client as well

howardjohn (approved):

lei-tang (approved):

howardjohn:
fyi: https://github.com/istio/istio/pull/40832

PR #40453, opened 2022-08-15
Unify cert validity check for xds client and caClient

hzxuzhonghu:

Not load certs that expires

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


security/pkg/pki/util/verify_cert.go
```diff
@@ -197,3 +198,23 @@ func FindRootCertFromCertificateChainBytes(certBytes []byte) ([]byte, error) {
...
+	var certDERBlock *pem.Block
+	certDERBlock, _ = pem.Decode(certPEMBlock)
+	if certDERBlock == nil {
+		return true, fmt.Errorf("failed to decode certificate")
+	}
+	x509Cert, err := x509.ParseCertificate(certDERBlock.Bytes)
+	if err != nil {
+		return true, fmt.Errorf("failed to parse the cert, err is %v", err)
+	}
+	return x509Cert.NotBefore.After(time.Now()) || x509Cert.NotAfter.Before(time.Now()), nil
```
howardjohn, line 219: nit: this is not meeting the `IsCertExpired` name. Since we are checking NotBefore as well now.

I am not sure its safe to check NotBefore here. For example, consider I have clock skew between CA and proxy:

```
* CA signs with "NotBefore 00:05"
* XDS proxy at 00:00 checks. IsCertExpired=true, so we do not use cert
* We should have used the cert -- we just got it
```

howardjohn (commented):
Nice improvement, one small comment

security/pkg/pki/util/verify_cert.go
hzxuzhonghu: Maybe, let me revert

howardjohn (approved):

PR #40455, opened 2022-08-15
Add lifecycle + terminationgracepriod to GW Chart

iainbowman:

The intention of this PR is to provide support for lifecycle (preStop hooks) and terminationGracePeriodSeconds to the Gateway deployment Helm chart.

### DIFF
--- a/manifests/charts/gateway/templates/deployment.yaml
+++ b/manifests/charts/gateway/templates/deployment.yaml
@@ -41,6 +41,9 @@ spec:
         - name: net.ipv4.ip_unprivileged_port_start
           value: "0"
       {{- end }}
+{{- if .Values.terminationGracePeriodSeconds }}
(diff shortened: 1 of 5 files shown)


zirain:
see https://github.com/istio/istio/issues/38912#issuecomment-1126512011

iainbowman:
Thanks, I wasn't aware of https://github.com/istio/istio/issues/38912#issuecomment-1126512011 and I understand your position on not adding lots of new fields. Would love to give this a try but unfortunately as we are using ArgoCD to deploy and it currently does not support helm post-render this solution will not work for my use case.

howardjohn:
Seems it could be done via https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/#kustomizing-helm-charts

iainbowman:
> Seems it could be done via https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/#kustomizing-helm-charts

Thanks I'll look into this 👍

Issue #40474, opened 2022-08-15
A few port matching Host header bugs


howardjohn:
This is primarily documentation for the bugs fixed in a PR.

There are two sets of issues:

1. Wildcard domains: we cannot match `*.com:*`, so we make the port explicit (`*.com:80`). If a user users an unexpected port for the Host, we will not match. We only put one port

2. More complex -- we have two route configs. One has wildcard and one doesn't:

```yaml
- domains: [*.example.com:443] # match wildcard
- domains: [foo.example.com, foo.example.com:*] # match concrete
```

If a user sends `foo.example.com:443` it would match the first one; we want the latter.

howardjohn:
https://github.com/envoyproxy/envoy/issues/22664 will be the fix

PR #40475, opened 2022-08-15
Ignore port number in domain matching entirely

howardjohn:
Fixes https://github.com/istio/istio/issues/40474

This is finally available in Envoy so we can drop our hacks of adding
the port, adding `:*`, etc.

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -641,7 +641,7 @@ var (
 	LocalClusterSecretWatcher = env.RegisterBoolVar("LOCAL_CLUSTER_SECRET_WATCHER", false,
 		"If enabled, the cluster secret watcher will watch the namespace of the external cluster instead of config cluster").Get()
 
-	SidecarIgnorePort = env.RegisterBoolVar("SIDECAR_IGNORE_PORT", true, "If enabled, port will not be used in vhost domain matches.").Get()
+	SidecarIgnorePort = env.RegisterBoolVar("SIDECAR_IGNORE_PORT_IN_HOST_MATCH", true, "If enabled, port will not be used in vhost domain matches.").Get()
 )
 
 // EnableEndpointSliceController returns the value of the feature flag and whether it was actually specified.
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -640,6 +640,8 @@ var (
 
 	LocalClusterSecretWatcher = env.RegisterBoolVar("LOCAL_CLUSTER_SECRET_WATCHER", false,
 		"If enabled, the cluster secret watcher will watch the namespace of the external cluster instead of config cluster").Get()
+
+	SidecarIgnorePort = env.RegisterBoolVar("SIDECAR_IGNORE_PORT", true, "If enabled, port will not be used in vhost domain matches.").Get()
 )
 
 // EnableEndpointSliceController returns the value of the feature flag and whether it was actually specified.
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -531,7 +531,7 @@ var (
 
 	StripHostPort = env.RegisterBoolVar("ISTIO_GATEWAY_STRIP_HOST_PORT", false,
 		"If enabled, Gateway will remove any port from host/authority header "+
-			"before any processing of request by HTTP filters or routing.").Get()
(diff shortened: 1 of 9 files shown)


hzxuzhonghu:
IIUC, s/ignore port name/port number

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -520,11 +529,20 @@ func getVirtualHostsForSniffedServicePort(vhosts []*route.VirtualHost, routeName
 	return virtualHosts
 }
 
+func SupportsIgnorePort(node *model.Proxy) bool {
+	return util.IsIstioVersionGE115(node.IstioVersion) && !node.IsProxylessGrpc()
```
hzxuzhonghu, line 533: Does grpc not support it?

pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -397,9 +401,13 @@ func BuildSidecarOutboundVirtualHosts(node *model.Proxy, push *model.PushContext
 		var domains []string
 		var altHosts []string
 		if svc == nil {
-			domains = []string{util.IPv6Compliant(hostname), name}
+			if SupportsIgnorePort(node) {
```
hzxuzhonghu, line 404: why need to ignore port in sidecar?  Shouldn't we only concentrate on the gateway?

howardjohn: I agree gateway is more important but it seems like we shouldn't ever be looking at the port?

pilot/pkg/networking/core/v1alpha3/httproute.go
howardjohn: Yes, not supported

pilot/pkg/networking/core/v1alpha3/httproute.go
hzxuzhonghu: I remember the original issue is on gateway, the access from remote users can use a port totally different from the gateway service because of lb in front.

But for sidecar, we do not meet such problem. We should not change the behavior, in case we miss some risks

ericvn (approved):
Tests and release notes look good.

pilot/pkg/networking/core/v1alpha3/httproute.go
howardjohn: Ok, I added a feature flag for sidecar

pilot/pkg/features/pilot.go
```diff
@@ -640,6 +640,8 @@ var (
 
 	LocalClusterSecretWatcher = env.RegisterBoolVar("LOCAL_CLUSTER_SECRET_WATCHER", false,
 		"If enabled, the cluster secret watcher will watch the namespace of the external cluster instead of config cluster").Get()
+
+	SidecarIgnorePort = env.RegisterBoolVar("SIDECAR_IGNORE_PORT", true, "If enabled, port will not be used in vhost domain matches.").Get()
```
hzxuzhonghu, line 644: SIDECAR_IGNORE_PORT is not clear, 
how about SIDECAR_IGNORE_PORT_IN_HOST_MATCH?

and set to false if we want to back port to release 1.15

howardjohn: will rename. I'll set to false on the backport, can we keep true on master? It took me hours to update all the test cases

hzxuzhonghu: Yes, i think it can be true on master

hzxuzhonghu:
@howardjohn feel free to remove the hold once renamed

releasenotes/notes/ignore-port.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+issues:
+- 40474
+releaseNotes:
+- |
+  **Fixed** an issue causing traffic to not match (and return a `404`) when using wildcard domain names and including an unexpected port in the `Host` header.
```
ramaraochavali, line 8: Add the new feature flag here for sidecars

ramaraochavali (approved):

ericvn:
Does this also need to be cherry-picked to 1.16?

howardjohn:
Its in 1.16, this was merge a long time ago.

PR #40476, opened 2022-08-15
bump gateway-api to support arm64

howardjohn:


ericvn (approved):

hzxuzhonghu (commented):
Two questions:

1. how is it related to arm64?
2. Do we need to follow gateway official release or commit?

howardjohn:
They fixed arm64 in the gateway -api commit. We are tracking the head
because there are conformance test fixes not back ported yet. Once they
ship those we can go to a release version.

The API itself doesn't change it's just the tests, etc so it's pretty safe
to stay on latest

hzxuzhonghu (approved):

PR #40488, opened 2022-08-16
Add file watcher for pod labels, only override xds request node when …

hzxuzhonghu:
…labels update

This is to reduce calling `Agent.generateNodeMetadata` by watching the files.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn (commented):
I wonder if it may be simpler to include a specific field for LABELS_FROM_KUBERNETES (but a better name). Then the control plane can know to not use those if it finds the pod, but use the common LABELS that are derived?

howardjohn:
Actually I think istio/proxy reads LABELS, so both this and the comment ^ may lead to unexpected results since those labels do not update

pilot/pkg/model/context.go
```diff
@@ -862,28 +866,26 @@ func (node *Proxy) SetServiceInstances(serviceDiscovery ServiceDiscovery) {
 // SetWorkloadLabels will set the node.Labels.
 // It merges both node meta labels and workload labels and give preference to workload labels.
 func (node *Proxy) SetWorkloadLabels(env *Environment) {
+	// If this is VM proxy, do not override labels at all, because in istio test we use pod to simulate VM.
+	if node.IsVM() {
```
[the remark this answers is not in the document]
howardjohn: oh, nvm. its really more like "node.IsTestVM". Maybe we can rename in a followup

pkg/istio-agent/agent.go
```diff
@@ -216,11 +216,11 @@ type AgentOptions struct {
...
-		proxyConfig:   proxyConfig,
-		cfg:           agentOpts,
-		secOpts:       sopts,
-		envoyOpts:     eopts,
-		caFileWatcher: filewatcher.NewWatcher(),
+		proxyConfig: proxyConfig,
+		cfg:         agentOpts,
+		secOpts:     sopts,
+		envoyOpts:   eopts,
+		fileWatcher: filewatcher.NewWatcher(),
```
howardjohn, line 223: do we need the filewatcher changes now?

howardjohn: Looks like its just generic now so not bad

howardjohn (approved):

hzxuzhonghu:
Thanks for rename the title

pkg/istio-agent/agent.go
hzxuzhonghu: yes

Issue #40445, opened 2022-08-14
Proxy configuration getting replicated to Gateway


anuragagarwal561994:
### Bug Description

We came across an unexpected bug last week in our production cluster. We are still trying to replicate and figuring out the root cause of the issue.

I would be happy to list down the possible configuration as directed, but I would quote the series of events that happened.

Our istiod setup is placed in istio-system namespace in kubernetes using `istioctl`.

We have application specific namespace, let us call this as `demo`.

We have our `ingress-gateway` setup up in `demo` namespace where we are running a micro-service with `istio-proxy` injected.

The `ingress-gateway` was set with requestCpu as 10 and limit as 12 on a 16 core cluster.

`istio-proxy` is setup using pod annotations with request cpu as 2 and limit as 4 using istio annotations. The concurrency was set using `proxy.istio.io/config: "concurrency: 0"`.

Now we wanted to test if increasing the concurrency of istio proxy would give us some advantage. So for setting this up, we took a dump of one of the pods and brought up another pod (not using deployment) and changed the configuration to use

`proxy.istio.io/config: "concurrency: 4"`

This was supposed to be the control bucket of our experiment. But to our surprise, another pod of `ingress-gateway` that came up because of auto-scaling after bringing this pod also copied this configuration and came up with concurrency: 4 which affected our traffic and auto-scaling logic a lot.

This is because when all the `ingress-gateway` pods restarted they all came up with `concurrency: 4` on a 16 core machine. The HPA was set to use 80% of the CPU but on a 16 core machine the concurrency 4 means usage of only 25% of the CPU and hence the ingress-gateway pods stopped auto-scaling thus impacting our production traffic.

As per our investigation, we also asked ourselves few questions which I am listing down here and the answer too.

Q. Did we change the concurrency in some previous configuration which led to this?
A. No we didn't change any configuration in `ingress-gateway` setup.

<img width="1626" alt="Screenshot 2022-08-14 at 3 29 51 PM" src="https://user-images.githubusercontent.com/6075379/184531903-7e5bc17e-15a3-4cf9-a346-39af8566ebae.png">

As we can see from the above image, a pod came as a result of auto-scaling at 2:35 (starting of the graph) and it started using all the 10 cores as expected.

The changes we did (of bringing a standalone pod), was done at around 2:51-55 AM (not shown in the graph) but as per logs.

Thereafter we can see that the pod that came at around 3:30 AM was using 4 cores only.

[...]

PR #40494, opened 2022-08-16
Fix ProxyConfig merge impacting shared state

howardjohn:
The issue was we were overriding things by setting them to pointers.
These pointers are then shared, and when `proto.merge` is used they are
all mutated.

Instead, take deepcopies and add a test
Fixes https://github.com/istio/istio/issues/40445


ericvn (approved):
Release note and tests look good.

ramaraochavali (approved):

Issue #39680, opened 2022-06-29
Support inline header with multiple values semantics in AuthorizationPolicy.condition


edieYoung:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**
Does istio Authz supports matching among multiple values separated by commas of an inline header? (No according to my experiments on 1.12.6)

There is support for this inline multiple header values semantics from Envoy: https://github.com/envoyproxy/envoy/issues/13751

In our use case, we have a header key with multiple values separated by comma, and we want to use Istio AuthorizationPolicy.conditions to match if the wanted header value appears in the inline multi-values header.

For example the header is:
x-example: "v1,v2,v3"

We want it to be allowed by the rule:
```
rules:
    - when:
        - key: request.headers[x-example]
          values: ["v1"]
```

Is it possible? Is this the expected behavior? 

**Describe alternatives you've considered**
This can also be supported if we can have string match in the authorization policy values, similar issue as https://github.com/istio/api/pull/2173

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[x] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

edieYoung:
Hi @howardjohn, could you take a look? Since you were in the same discussion thread, do you have any updates on this issue? https://github.com/kubernetes-sigs/gateway-api/issues/213#issuecomment-888754154 I think we do need a way to support inline multi-value headers authz.

edieYoung:
Currently, this is already causing an semantics inconsistency in Istio Authz,
See following examples:
```
Policy: ALLOW x-foo: bar
Request: curl -H "x-foo: bar" -H "x-foo: not-bar"
```
This request will be denied.
```
Policy: ALLOW x-foo: bar*
Request: curl -H "x-foo: bar" -H "x-foo: not-bar"
``` 
But this request will be allowed, because the header gets merged as `x-foo:bar,not-bar`

edieYoung:
reopening issue.

edieYoung:
reopen

Issue #16585, opened 2019-08-27
Support regex for ServiceRole spec.rules.paths


wereeder:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**
Support regex paths for ServiceRole spec.rules.paths, similar to how the Policy supports regex for spec.peers.jwt.trigger_rules.excluded_paths

**Describe alternatives you've considered**
Paths currently support wildcard patching for suffix, prefix or exact match, but not a mix of these. I tried using wildcard at beginning and end but this did not work. Without regex support, I have had to define several variations for allowing access to the swagger-ui.html. For example, I have added the following:

```yaml
apiVersion: rbac.istio.io/v1alpha1
kind: ServiceRole
metadata:
  name: view-swagger
  namespace: dp
spec:
  rules:
  - services: ["*"]
    paths:
    - "*/swagger-ui.html"
    - "*/api-docs"
    - "*/swagger-resources"
    - "*/swagger-resources/configuration/ui"
    - "*/swagger-resources/configuration/security"
    - "*/springfox-swagger-ui/fonts/open-sans-v15-latin-regular.woff2"
    - "*/springfox-swagger-ui/fonts/open-sans-v15-latin-700.woff2"
    - "*/springfox-swagger-ui/fonts/source-code-pro-v7-latin-300.woff2"
    - "*/springfox-swagger-ui/fonts/titillium-web-v6-latin-700.woff2"
    - "*/springfox-swagger-ui/springfox.js?v=2.9.2"
    - "*/springfox-swagger-ui/swagger-ui-bundle.js?v=2.9.2"
    - "*/springfox-swagger-ui/swagger-ui-standalone-preset.js?v=2.9.2"
    - "*/springfox-swagger-ui/springfox.css?v=2.9.2"
    - "*/springfox-swagger-ui/swagger-ui.css?v=2.9.2"
    - "*/springfox-swagger-ui/favicon-32x32.png?v=2.9.2"
    - "*/springfox-swagger-ui/favicon-16x16.png?v=2.9.2"
```

I would like to simplify this with a regex, such as:
```yaml
apiVersion: rbac.istio.io/v1alpha1
kind: ServiceRole
metadata:
  name: view-swagger
  namespace: dp
spec:
  rules:
  - services: ["*"]
    regex_paths:
    - \/.*swagger.*
    - \/.*api-docs
```

**Affected product area (please put an X in all that apply)**

[ ] Configuration Infrastructure
[...]

PR #40499, opened 2022-08-16
Authorization policy: support inline multi-values header semantics

edieYoung:

Fix issue: https://github.com/istio/istio/issues/39680
This PR adds support in Authorization Policy to recognize inline multi-values header format by using safe regex match.

For example:
given the allowed header { key: x-foo, value: [a, b]} the request with header {key: x-foo value: "a,b,c"} will be allowed after this fix. Before this fix, this is not allowed.

### DIFF
--- a/pilot/pkg/security/authz/matcher/header.go
+++ b/pilot/pkg/security/authz/matcher/header.go
@@ -15,6 +15,7 @@
 package matcher
 
 import (
+	"fmt"
 	"regexp"
 	"strings"
 
@@ -52,6 +53,40 @@ func HeaderMatcher(k, v string) *routepb.HeaderMatcher {
 		Name: k,
 		HeaderMatchSpecifier: &routepb.HeaderMatcher_ExactMatch{
 			ExactMatch: v,
+		},
+	}
+}
+
+// HeaderMatcherWithRegex converts a key, value string pair to a corresponding
+// HeaderMatcher using regex to support the inline multi-values header that is
+// concatenated by commas, as per RFC7230.
+func HeaderMatcherWithRegex(k, v string) *routepb.HeaderMatcher {
+	// We must check "*" first to make sure we'll generate a non empty value in the prefix/suffix case.
+	// Empty prefix/suffix value is invalid in HeaderMatcher.
+	var regex string
+	if v == "*" {
+		return &routepb.HeaderMatcher{
+			Name: k,
+			HeaderMatchSpecifier: &routepb.HeaderMatcher_PresentMatch{
+				PresentMatch: true,
+			},
+		}
+	} else if strings.HasPrefix(v, "*") {
+		regex = `.*` + regexp.QuoteMeta(v[1:])
+	} else if strings.HasSuffix(v, "*") {
+		regex = regexp.QuoteMeta(v[:len(v)-1]) + `.*`
+	} else {
+		regex = regexp.QuoteMeta(v)
+	}
+	return &routepb.HeaderMatcher{
+		Name: k,
+		HeaderMatchSpecifier: &routepb.HeaderMatcher_SafeRegexMatch{
+			SafeRegexMatch: &matcherpb.RegexMatcher{
+				EngineType: &matcherpb.RegexMatcher_GoogleRe2{
+					GoogleRe2: &matcherpb.RegexMatcher_GoogleRE2{},
+				},
+				Regex: fmt.Sprintf(`^%s$|^%s,.*|.*,%s,.*|.*,%s$`, regex, regex, regex, regex),
+			},
 		},
 	}
 }
--- a/pilot/pkg/security/authz/matcher/header_test.go
+++ b/pilot/pkg/security/authz/matcher/header_test.go
@@ -63,6 +63,55 @@ func TestHeaderMatcher(t *testing.T) {
 	}
 }
 
+func TestHeaderMatcherWithRegex(t *testing.T) {
+	testCases := []struct {
+		Name   string
+		K      string
+		V      string
+		Expect *routepb.HeaderMatcher
+	}{
+		{
+			Name: "exact match",
+			K:    ":path",
+			V:    "/productpage",
+			Expect: &routepb.HeaderMatcher{
+				Name: ":path",
+				HeaderMatchSpecifier: &routepb.HeaderMatcher_SafeRegexMatch{
+					SafeRegexMatch: &matcherpb.RegexMatcher{
+						EngineType: &matcherpb.RegexMatcher_GoogleRe2{
+							GoogleRe2: &matcherpb.RegexMatcher_GoogleRE2{},
+						},
+						Regex: "^/productpage$|^/productpage,.*|.*,/productpage,.*|.*,/productpage$",
+					},
+				},
+			},
+		},
+		{
+			Name: "suffix match",
+			K:    ":path",
+			V:    "*/productpage*",
+			Expect: &routepb.HeaderMatcher{
+				Name: ":path",
+				HeaderMatchSpecifier: &routepb.HeaderMatcher_SafeRegexMatch{
+					SafeRegexMatch: &matcherpb.RegexMatcher{
+						EngineType: &matcherpb.RegexMatcher_GoogleRe2{
+							GoogleRe2: &matcherpb.RegexMatcher_GoogleRE2{},
+						},
+						Regex: `^.*/productpage\*$|^.*/productpage\*,.*|.*,.*/productpage\*,.*|.*,.*/productpage\*$`,
+					},
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		actual := HeaderMatcherWithRegex(tc.K, tc.V)
+		if !cmp.Equal(tc.Expect, actual, protocmp.Transform()) {
+			t.Errorf("expecting %v, but got %v", tc.Expect, actual)
+		}
+	}
+}
+
 func TestHostMatcherWithRegex(t *testing.T) {
 	testCases := []struct {
 		Name   string
--- a/pilot/pkg/security/authz/model/generator.go
+++ b/pilot/pkg/security/authz/model/generator.go
@@ -194,7 +194,7 @@ func (requestHeaderGenerator) principal(key, value string, forTCP bool) (*rbacpb
 	if err != nil {
 		return nil, err
 	}
-	m := matcher.HeaderMatcher(header, value)
+	m := matcher.HeaderMatcherWithRegex(header, value)
 	return principalHeader(m), nil
 }
 
--- a/pilot/pkg/security/authz/model/generator_test.go
+++ b/pilot/pkg/security/authz/model/generator_test.go
@@ -204,9 +204,11 @@ func TestGenerator(t *testing.T) {
 			key:   "request.headers[x-foo]",
 			value: "foo",
 			want: yamlPrincipal(t, `
-         header:
-          exactMatch: foo
-          name: x-foo`),
+        header:
+          name: x-foo
+          safeRegexMatch:
+            googleRe2: {}
+            regex: ^foo$|^foo,.*|.*,foo,.*|.*,foo$`),
 		},
 		{
 			name:  "requestClaimGenerator",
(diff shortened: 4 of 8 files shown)


pilot/pkg/security/authz/model/generator.go
```diff
@@ -194,7 +194,7 @@ func (requestHeaderGenerator) principal(key, value string, forTCP bool) (*rbacpb
 	if err != nil {
 		return nil, err
 	}
-	m := matcher.HeaderMatcher(header, value)
+	m := matcher.HeaderMatcherWithRegex(header, value)
```
lei-tang, line 197: Thanks for the PR! Can you add a release notes to the PR describing how the PR changes the current header matching behavior, and potential impacts to the users that use the current header matching? Meanwhile, the docs of current header matching behavior need to be updated and a user guide for this feature needs to be added in follow-up PRs in istio.io/.

pilot/pkg/security/authz/matcher/header.go
```diff
@@ -56,6 +57,40 @@ func HeaderMatcher(k, v string) *routepb.HeaderMatcher {
...
+		regex = regexp.QuoteMeta(v)
+	}
+	return &routepb.HeaderMatcher{
+		Name: k,
+		HeaderMatchSpecifier: &routepb.HeaderMatcher_SafeRegexMatch{
+			SafeRegexMatch: &matcherpb.RegexMatcher{
+				EngineType: &matcherpb.RegexMatcher_GoogleRe2{
+					GoogleRe2: &matcherpb.RegexMatcher_GoogleRE2{},
+				},
+				Regex: fmt.Sprintf("^%s$|^%s,|,%s,|,%s$", regex, regex, regex, regex),
```
howardjohn, line 88: This feels like we may be introducing a high runtime cost to all users in all cases to support a niche feature - and risk of bugs

pilot/pkg/security/authz/matcher/header_test.go
```diff
@@ -63,6 +63,55 @@ func TestHeaderMatcher(t *testing.T) {
...
+			K:    ":path",
+			V:    "/productpage",
+			Expect: &routepb.HeaderMatcher{
+				Name: ":path",
+				HeaderMatchSpecifier: &routepb.HeaderMatcher_SafeRegexMatch{
+					SafeRegexMatch: &matcherpb.RegexMatcher{
+						EngineType: &matcherpb.RegexMatcher_GoogleRe2{
+							GoogleRe2: &matcherpb.RegexMatcher_GoogleRE2{},
+						},
+						Regex: "^/productpage$|^/productpage,|,/productpage,|,/productpage$",
```
howardjohn, line 84: I think this introduces an expectation that commas will not have any spaces around them - not sure this is true in real world?

pilot/pkg/security/authz/matcher/header_test.go
```diff
@@ -63,6 +63,55 @@ func TestHeaderMatcher(t *testing.T) {
...
+			K:    ":path",
+			V:    "*/productpage*",
+			Expect: &routepb.HeaderMatcher{
+				Name: ":path",
+				HeaderMatchSpecifier: &routepb.HeaderMatcher_SafeRegexMatch{
+					SafeRegexMatch: &matcherpb.RegexMatcher{
+						EngineType: &matcherpb.RegexMatcher_GoogleRe2{
+							GoogleRe2: &matcherpb.RegexMatcher_GoogleRE2{},
+						},
+						Regex: "^.*/productpage\\*$|^.*/productpage\\*,|,.*/productpage\\*,|,.*/productpage\\*$",
```
howardjohn, line 100: Very worried about this complexity -- if a simple rule becomes this complex regex, its hard to grok the security posture and settings, and is a big risk

howardjohn (changes requested):
This seems like we are being overly lenient and will lead to an RBAC bypass.

So before:

policy: `ALLOW x-foo: bar`

Now I send `curl -H "x-foo: bar" -H "x-foo: not-bar-attack"`. The backend reads the last value only, we bypassed their policy.

howardjohn:
IMO _if_ we  want to make this change we should add native matching support to Envoy instead of hacking around with complex regex. However, i am not sure it is safe even then (per comment above)

edieYoung:
> This seems like we are being overly lenient and will lead to an RBAC bypass.
> So before:
> policy: ALLOW x-foo: bar
> Now I send curl -H "x-foo: bar" -H "x-foo: not-bar-attack". The backend reads the last value only, we bypassed their policy.

This header will be merged into `x-foo: bar,not-bar-attack` as per this Envoy [change](https://github.com/envoyproxy/envoy/commit/2c60632d41555ec8b3d9ef5246242be637a2db0f), and if the policy is to allow `x-foo: bar`, I think [by definition](https://istio.io/latest/docs/reference/config/security/conditions/#supported-conditions) it should be allowed. 
If user wants to match all of the header values, they should specify multiple conditions under the same rule.

edieYoung:
> IMO _if_ we want to make this change we should add native matching support to Envoy instead of hacking around with complex regex. However, i am not sure it is safe even then (per comment above)

Agreed a native matching support to Envoy would be a better choice. Here is the [issue](https://github.com/envoyproxy/envoy/issues/13751) but it is kind of stale.

edieYoung:
@howardjohn Thanks for your comments, I agree with your concern, that's why I am considering maybe we can make this fix an opt-in for users, so that users can opt-in/out this support by specifying a mesh config flag. They will need to make the judgement on whether or not to turn on in order to fit for their own system. wdyt?

For example, in our system, this semantics is important to our authz framework and the requests using this support is not performance sensitive(we conducted perf test and the results are acceptable given our scale).

pilot/pkg/security/authz/matcher/header_test.go
edieYoung: it should be true, with this change in Envoy https://github.com/envoyproxy/envoy/commit/2c60632d41555ec8b3d9ef5246242be637a2db0f

pilot/pkg/security/authz/matcher/header.go
edieYoung: maybe we can consider to make it an option for users, replied in https://github.com/istio/istio/pull/40499#issuecomment-1217215266

pilot/pkg/security/authz/model/generator.go
edieYoung: Thanks @lei-tang, I will update it when we reach a consensus on the solution :)

lei-tang: Replacing HeaderMatcher() with HeaderMatcherWithRegex() changes the current behavior. Whether HeaderMatcherWithRegex() is used or not should be protected by a feature-enablement flag.

edieYoung: a feature-enablement flag sgtm.

pilot/pkg/security/authz/model/generator_test.go
```diff
@@ -204,9 +204,11 @@ func TestGenerator(t *testing.T) {
 			key:   "request.headers[x-foo]",
 			value: "foo",
 			want: yamlPrincipal(t, `
-         header:
-          exactMatch: foo
-          name: x-foo`),
+        header:
```
lei-tang, line 207: The test for the original behavior should remain. The test for the new feature should be a separate test.

lei-tang (changes requested):

edieYoung:
@lei-tang @howardjohn another solution for this issue, do you think it would be more acceptable if we just expose a regex match field in the header Authz API? For example:
```
when:
      - key: request.headers[x-foo]
        regex: ""
```

lei-tang:
> @lei-tang @howardjohn another solution for this issue, do you think it would be more acceptable if we just expose a regex match field in the header Authz API? For example:
> 
> ```
> when:
>       - key: request.headers[x-foo]
>         regex: ""
> ```

Can you add more details to this alternative solution? For example, a detailed example of how users use this alternative solution.

edieYoung:
@lei-tang this alternative solution is to allow users to specify the regex in the authz condition, so that we can specify any regex as need, here is an example API, 

```
rules:
- when:
    - key: request.headers[x-foo]
       regex: `'^bar,.*`
```

lei-tang:
> @lei-tang this alternative solution is to allow users to specify the regex in the authz condition, so that we can specify any regex as need, here is an example API,
> 
> ```
> rules:
> - when:
>     - key: request.headers[x-foo]
>        regex: `'^bar,.*`
> ```

There are concerns on supporting regex in authz policy, e.g., https://github.com/istio/istio/issues/16585. 
@edieYoung If you like, maybe you can write a proposal for supporting regex in authz policy and present it on Istio WG meeting to get feedback?

rrobb-aspen:
> #39680

We also have the same concern.  We have put a PR into Istio API for this issue.

edieYoung:
Hey @lei-tang @howardjohn, circling back on this, **being able to define a policy rule to match among multiple values in the inline header as per this Envoy's [change](https://github.com/envoyproxy/envoy/issues/13751)** is still a demanding feature for many of us (Note that the ask is not to support arbitrary regex or contain match, etc, they are just some of the solutions), and we have been using this patch in our fork for a while and lived happy with it . 

To continue on this, consider the concern of breaking the existing behavior, what about introducing a separate condition attribute called `request.inline-headers` to distinguish from the original header attribute? By using this attribute, we can clearly describe how it behaves differently from `request.header` that it can match the value among the multiple values in inline header format while `request.header` does not. So that only the users who explicitly know they are expecting inline headers will consciously use this feature.

The new usage would be like:
The following policy will allow the request with header in the inline format `x-foo: bar,baz`:
```
ALLOW
rules:
- when:
    - key: request.ineline-headers[x-foo]
       value: bar
```

I will send out a PR once it's ready, but just want to get some quick feedbacks, also please advise any process I need to go through to introduce this new feature, thanks in advance!

howardjohn:
@edieYoung have we looked into adding native support in Envoy? Multi header sematnics is very very complex - I don't feel comfortable implementing it with a regex regardless of the user facing API

edieYoung:
@howardjohn totally agree this semantics is complex, but I think it's worthy to do something to catch up given inline header format is the standard as well.
Unfortunately, Envoy doesn't add support for this specifically, it provides multiple matching types to leave the choice of semantics to user-side implementation. (reading from https://github.com/envoyproxy/envoy/issues/13751, the proposed solution is only about providing generic matching API, which provides contains/regex match in string match by treating header value merely as a string).

In my opinion, the use of Regex in this case is not a major concern since we are only using a fixed and strict regex as an internal implementation detail and restricting its blast radius through an opt-in API. To be extra cautious, we could label it as experimental, such as `request.experimental-inline-header`. What do you think?

This is the draft PR https://github.com/istio/istio/pull/46183.

howardjohn:
I really don't believe regex will ever be acceptable here. Aside from performance concerns, I don't think its secure or correct. See https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.1.2 - its not so simple. `foo , ,bar,charlie` is a valid header.

Can you parse a lot of headers with regex this way - sure... but this is a security policy. We need to be 100% certain that we can parse **all** headers correctly, or we have a CVE. I am doubtful I will ever have that level of confidence with a regex.

edieYoung:
I see, I guess the most ideal way would be having Envoy to support something like ListMatcher which can parse the header value as a list properly.
However, in current situation, multiple value headers will always be concatenated with comma by Envoy(https://github.com/envoyproxy/envoy/commit/2c60632d41555ec8b3d9ef5246242be637a2db0f) before matching, so it's 100% that we are always expecting a comma separated string(no blank space, no empty value).

howardjohn:
> so it's 100% that we are always expecting a comma separated string(no blank space, no empty value).

I don't think this is true... `curl echo -H "foo: bar, , baz"` will very likely have spaces, for example. And that is just one case. I have no confidence at all that Istio (especially with regex) is going to properly handle all the crazy edge cases of HTTP. Even envoy did not without regex, hence the CVE you linked.

karloramone:
Is there any plans to continue with this PR?

edieYoung:
we'd love to continue this if we can get approvals from the community, since this is stale, I just revived in another PR https://github.com/istio/istio/pull/56085 and made it a separate attribute so the original header attribute doesn't change.

PR #40503, opened 2022-08-16
add stat_prefix support for virtual service api

varks:
adds support for https://github.com/istio/api/pull/2405

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/route/route.go
+++ b/pilot/pkg/networking/core/v1alpha3/route/route.go
@@ -434,6 +434,11 @@ func translateRoute(
 		Match:    translateRouteMatch(node, virtualService, match),
 		Metadata: util.BuildConfigInfoMetadata(virtualService.Meta),
 	}
+
+	if match != nil && match.StatPrefix != "" {
+		out.StatPrefix = match.StatPrefix
+	}
+
 	authority := ""
 	if in.Headers != nil {
 		operations := translateHeadersOperations(in.Headers)
--- a/pilot/pkg/networking/core/v1alpha3/route/route_test.go
+++ b/pilot/pkg/networking/core/v1alpha3/route/route_test.go
@@ -334,22 +334,11 @@ func TestBuildHTTPRoutes(t *testing.T) {
 			serviceRegistry, nil, 8080, gatewayNames, false, nil)
 		xdstest.ValidateRoutes(t, routes)
 		g.Expect(err).NotTo(gomega.HaveOccurred())
-		g.Expect(len(routes)).To(gomega.Equal(1))
+		g.Expect(len(routes)).To(gomega.Equal(2))
 		g.Expect(routes[0].GetMatch().GetPrefix()).To(gomega.Equal("/foo"))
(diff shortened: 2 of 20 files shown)


pilot/pkg/networking/core/v1alpha3/route/route_test.go
```diff
@@ -1856,6 +1869,43 @@ var virtualServiceMatchingOnSourceNamespace = config.Config{
...
+		GroupVersionKind: gvk.VirtualService,
+		Name:             "acme",
+	},
+	Spec: &networking.VirtualService{
+		Hosts: []string{},
+		Http: []*networking.HTTPRoute{
+			{
+				Name: "foo",
+				Match: []*networking.HTTPMatchRequest{
+					{
```
ramaraochavali, line 1883: Can you add another MatchRequest that does not have stat prefix configured and validate that route does not have it configured for completeness?

ramaraochavali:
LGTM, with minor test request.

pilot/pkg/networking/core/v1alpha3/route/route.go
```diff
@@ -434,6 +434,11 @@ func translateRoute(
 		Match:    translateRouteMatch(node, virtualService, match),
 		Metadata: util.BuildConfigInfoMetadata(virtualService.Meta),
 	}
+
+	if match != nil && match.StatPrefix != "" {
+		out.StatPrefix = match.StatPrefix
```
hzxuzhonghu, line 439: Why made StatPrefix in  `HTTPMatchRequest`? This API is to specify matching conditions.

ramaraochavali: Because we want metrics per match (path) condition. If service exposes multiple paths, we want stats per route. This is mainly useful for Ingress Gateway.

pilot/pkg/networking/core/v1alpha3/route/route_test.go
varks: @ramaraochavali done.

ramaraochavali (approved):
Nice. Thank you

pkg/bootstrap/testdata/all_golden.json
```diff
@@ -232,6 +232,9 @@
           "suffix": "shadow_denied"
           },
           {
+          "safe_regex": {"google_re2":{}, "regex":"vhost.*.route.*"}
```
howardjohn, line 235: `vhost.*\.route.*`

Do we need to escape the `.`?

howardjohn: This is our first regex, is there perf cost associated here? @douglas-reid @kyessenov

ramaraochavali: Good point about escaping. I forgot about that. Yes it needs to be escaped. @varks can you please fix it?

varks: @howardjohn thanks for catching this. Here's the fix: https://github.com/istio/istio/pull/40594

PR #40505, opened 2022-08-17
Make use of Proxy.Labels instead of Meatdata.Labels

hzxuzhonghu:

This is to fix misuse of labels because of rebase issue

### DIFF
--- a/pilot/pkg/model/context.go
+++ b/pilot/pkg/model/context.go
@@ -278,10 +278,6 @@ type Proxy struct {
 	// while NodeMetadata.Labels are set during bootstrap.
 	Labels map[string]string
 
-	// IstioMetaLabels contains the labels specified by ISTIO_METAJSON_LABELS and platform instance,
-	// so we can support pod labels update by plus the fundamental labels.
-	IstioMetaLabels map[string]string
-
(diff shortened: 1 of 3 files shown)


pilot/pkg/model/context.go
```diff
@@ -542,8 +542,8 @@ type NodeMetadata struct {
 	// Mostly used when istiod requests the upstream.
 	IstioRevision string `json:"ISTIO_REVISION,omitempty"`
 
-	// Labels specifies the set of workload instance (ex: k8s pod) labels associated with this node. Labels is a
-	// superset of IstioMetaLabels.
+	// Labels specifies the set of workload instance (ex: k8s pod) labels associated with this node.
+	// Note: should not be used during xds generation because pod labels update support.
 	Labels map[string]string `json:"LABELS,omitempty"`
```
howardjohn, line 547: Maybe we should rename it (the go type, not the `LABELS`), to something like StaticLabels or similar. So its harder to misuse?

howardjohn (approved):

pilot/pkg/model/context.go
hzxuzhonghu: Will think about it

PR #37122, opened 2022-02-02
send unhealthy endpoints to Envoy (take 2)

ramaraochavali:
- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

Issue #18367, opened 2019-10-28
Consider sending "Not Ready" endpoints over EDS as unhealthy


howardjohn:
Currently, we are only sending "Ready" endpoints over EDS. This makes some fields like minHealthPercent (panic threshold) a bit hard to reason about. If we keep removing endpoints from the endpoint set, then we keep moving the target. If envoy health checking perfectly aligned with k8s (generally it won't, since one is passive and one is active), you could never actually hit this since envoy would almost never have "unhealthy" endpoints -- Pilot would just remove them right when they became unhealthy.

This same applies to Locality load balancing. Generally, after some % of endpoints becomes unhealthy we start sending more and more traffic to other localities, but this % moving as endpoints are marked unhealthy by k8s readiness.

We should investigate other benefits of sending unhealthy endpoints, as well as costs/problems. 

This will of course be slightly less performant, but generally there are few "not ready" endpoints. We may actually see better performance in Envoy side for pods that are flipping between healthy/unhealthy.

@rshriram @hzxuzhonghu

hzxuzhonghu:
I am concerned about too many not ready endpoints will lead to 503. As the default max retry is limited.

howardjohn:
I mean we send them with "health_status=unhealthy" so traffic is not routed
to them

ramaraochavali:
IMO, while it is good to not send an initial NOT ready endpoint to Envoy i.e Pod becomes initially ready, I think later flips back to Not ready/Live check failures should be better handled by Envoy. The outlier detection, retries etc are meant to handle such scenarios assuming that POD readiness/liveness flip is a transient one.  Otherwise control plane is doing what Envoy is supposed to do and if there is a gap , we should fix it in Envoy.

BTW, How will this work for VMs and external services created with service entries?

howardjohn:
External services have no notion of 'not ready' in the service entry API, so there would be no changes there, this would only apply to k8s.

Just wondering, why do you think we should not send the initial not ready endpoint?

ramaraochavali:
> External services have no notion of 'not ready' in the service entry API, so there would be no changes there, this would only apply to k8s.

Agree. I am more wondering about from the behavior perspective, not from the actual change. Those services will depend on Envoy's capabilities because control plane has no knowledge of their health/readiness.

> Just wondering, why do you think we should not send the initial not ready endpoint?

[...]

PR #40506, opened 2022-08-17
Enable locality failover without outlier detection if unready endpoints are sent to Envoy

sangwa:

At the moment locality failover on load balancing is enabled by default but is only applied if there is outlier detection configured for the destination service, as originally there was no other way for the sidecar Envoy to know when upstream endpoints are unhealthy and perform the actual failover (since unready endpoints were just excluded from the EDS updates).

That behavior led to some unwanted effects like the locality failover not actually working as expected, sending all the traffic to a single locality unless that locality failed completely (see e.g. #18367) and was one of the reasons to implement support for sending unready endpoints to sidecars, besides support for gradual service warmup (#36274, #37122).

As now we have the ability to configure the control plane to send unready endpoints, there is no requirement for outlier detection to enable locality failover in that case.

In my situation we would like to have locality failover enabled to have traffic pinned within the same AZ to prevent extra costs of unchecked cross-zone data transfer in a public cloud, but don't want to have outlier detection necessarily configured for our services (some workloads for example can be unhealthy without returning 500 errors, which could be caught by custom readiness probes).

This PR implements necessary changes with related tests covering both cases of enabled and disabled unready EDS (since the feature flag is currently disabled by default) and some refactoring to make the code more clear on the implementation.

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/cluster.go
+++ b/pilot/pkg/networking/core/v1alpha3/cluster.go
@@ -850,11 +850,10 @@ func ApplyRingHashLoadBalancer(c *cluster.Cluster, lb *networking.LoadBalancerSe
 func applyLocalityLBSetting(locality *core.Locality, proxyLabels map[string]string, cluster *cluster.Cluster,
 	localityLB *networking.LocalityLoadBalancerSetting,
 ) {
-	// Failover should only be applied with outlier detection, or traffic will never failover.
-	enabledFailover := cluster.OutlierDetection != nil
+	// Failover should only be applied with outlier detection or SendUnhealthyEndpoints, or traffic will never failover.
+	odEnabled := cluster.OutlierDetection != nil
 	if cluster.LoadAssignment != nil {
-		// TODO: enable failoverPriority for `STRICT_DNS` cluster type
-		loadbalancer.ApplyLocalityLBSetting(cluster.LoadAssignment, nil, locality, proxyLabels, localityLB, enabledFailover)
+		loadbalancer.ApplyLocalityLBSetting(cluster.LoadAssignment, nil, locality, proxyLabels, localityLB, odEnabled)
 	}
 }
 
--- a/pilot/pkg/networking/core/v1alpha3/loadbalancer/loadbalancer.go
+++ b/pilot/pkg/networking/core/v1alpha3/loadbalancer/loadbalancer.go
@@ -24,6 +24,7 @@ import (
 	wrappers "google.golang.org/protobuf/types/known/wrapperspb"
 
 	"istio.io/api/networking/v1alpha3"
+	"istio.io/istio/pilot/pkg/features"
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pilot/pkg/networking/util"
 )
@@ -63,11 +64,13 @@ func ApplyLocalityLBSetting(
 	locality *core.Locality,
 	proxyLabels map[string]string,
 	localityLB *v1alpha3.LocalityLoadBalancerSetting,
-	enableFailover bool,
+	outlierDetectionEnabled bool,
 ) {
 	if localityLB == nil || loadAssignment == nil {
 		return
 	}
+
+	enableFailover := outlierDetectionEnabled || features.SendUnhealthyEndpoints.Load()
 
 	// one of Distribute or Failover settings can be applied.
 	if localityLB.GetDistribute() != nil {
--- a/pilot/pkg/xds/eds.go
+++ b/pilot/pkg/xds/eds.go
@@ -325,9 +325,9 @@ func (s *DiscoveryServer) generateEndpoints(b EndpointBuilder) *endpoint.Cluster
 	l := b.createClusterLoadAssignment(localityLbEndpoints)
 
 	// If locality aware routing is enabled, prioritize endpoints or set their lb weight.
-	// Failover should only be enabled when there is an outlier detection, otherwise Envoy
-	// will never detect the hosts are unhealthy and redirect traffic.
-	enableFailover, lb := getOutlierDetectionAndLoadBalancerSettings(b.DestinationRule(), b.port, b.subsetName)
+	// Failover should only be enabled when there is an outlier detection, or when SendUnhealthyEndpoints
+	// feature is enabled, otherwise Envoy will never detect the hosts are unhealthy and redirect traffic.
+	odEnabled, lb := getOutlierDetectionAndLoadBalancerSettings(b.DestinationRule(), b.port, b.subsetName)
 	lbSetting := loadbalancer.GetLocalityLbSetting(b.push.Mesh.GetLocalityLbSetting(), lb.GetLocalityLbSetting())
 	if lbSetting != nil {
 		// Make a shallow copy of the cla as we are mutating the endpoints with priorities/weights relative to the calling proxy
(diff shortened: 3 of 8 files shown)


pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -850,11 +850,10 @@ func ApplyRingHashLoadBalancer(c *cluster.Cluster, lb *networking.LoadBalancerSe
 func applyLocalityLBSetting(locality *core.Locality, proxyLabels map[string]string, cluster *cluster.Cluster,
 	localityLB *networking.LocalityLoadBalancerSetting,
 ) {
-	// Failover should only be applied with outlier detection, or traffic will never failover.
-	enabledFailover := cluster.OutlierDetection != nil
+	// Failover should only be applied with outlier detection or SendUnhealthyEndpoints, or traffic will never failover.
+	odEnabled := cluster.OutlierDetection != nil
 	if cluster.LoadAssignment != nil {
-		// TODO: enable failoverPriority for `STRICT_DNS` cluster type
```
ramaraochavali, line 856: Why is this TODO removed?

pilot/pkg/networking/core/v1alpha3/cluster.go
```diff
@@ -850,11 +850,10 @@ func ApplyRingHashLoadBalancer(c *cluster.Cluster, lb *networking.LoadBalancerSe
 func applyLocalityLBSetting(locality *core.Locality, proxyLabels map[string]string, cluster *cluster.Cluster,
 	localityLB *networking.LocalityLoadBalancerSetting,
 ) {
-	// Failover should only be applied with outlier detection, or traffic will never failover.
-	enabledFailover := cluster.OutlierDetection != nil
+	// Failover should only be applied with outlier detection or SendUnhealthyEndpoints, or traffic will never failover.
+	odEnabled := cluster.OutlierDetection != nil
```
ramaraochavali, line 854: nit: can we expand this to outlierDetectionEnabled?

pilot/pkg/xds/eds.go
```diff
@@ -325,9 +325,9 @@ func (s *DiscoveryServer) generateEndpoints(b EndpointBuilder) *endpoint.Cluster
 	l := b.createClusterLoadAssignment(localityLbEndpoints)
 
 	// If locality aware routing is enabled, prioritize endpoints or set their lb weight.
-	// Failover should only be enabled when there is an outlier detection, otherwise Envoy
-	// will never detect the hosts are unhealthy and redirect traffic.
-	enableFailover, lb := getOutlierDetectionAndLoadBalancerSettings(b.DestinationRule(), b.port, b.subsetName)
+	// Failover should only be enabled when there is an outlier detection, or when SendUnhealthyEndpoints
+	// feature is enabled, otherwise Envoy will never detect the hosts are unhealthy and redirect traffic.
+	odEnabled, lb := getOutlierDetectionAndLoadBalancerSettings(b.DestinationRule(), b.port, b.subsetName)
```
ramaraochavali, line 330: same comment about name

releasenotes/notes/locality-lb-failover-with-unready-eds.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+releaseNotes:
+- |
+  **Added** Locality load balancing failover is now automatically enabled if the control plane
+  is configured to send unready endpoints (PILOT_SEND_UNHEALTHY_ENDPOINTS feature flag)
```
ramaraochavali, line 7: unless explictly disabled, even if no outlier detection is configured for target services. is not clear to me

pilot/pkg/networking/core/v1alpha3/loadbalancer/loadbalancer.go
```diff
@@ -63,12 +64,14 @@ func ApplyLocalityLBSetting(
...
 	proxyLabels map[string]string,
 	localityLB *v1alpha3.LocalityLoadBalancerSetting,
-	enableFailover bool,
+	outlierDetectionEnabled bool,
 ) {
 	if localityLB == nil || loadAssignment == nil {
 		return
 	}
 
+	enableFailover := outlierDetectionEnabled || features.SendUnhealthyEndpoints.Load()
```
howardjohn, line 73: I think because of some questionable decisions we made years ago, this is fairly unsafe.

Currently, locality lb is "on" by default, if you have OD. Failover mode is probably not a reasonable default, but at least now it is scoped to only when you create OD, so if you just install Istio your traffic doesn't suddenly change anywhere.

In the future, if/when we enable SendUnhealthyEndpoints by default, that would change here.

I wonder if we should try to change the default to not enable locality lb... its hard to change behavior in Istio (backwards compat), but this one is pretty bad IMO. In our managed product we disabled it from the start.

pilot/pkg/networking/core/v1alpha3/cluster.go
sangwa: My fault probably, this comment was a bit misleading to me, as the following code that applies locality failover settings is always called for both STRICT_DNS and EDS clusters, so it's not immediately clear what exactly is TODO.

On second thought, STRICT_DNS clusters are only generated from ServiceEntries and in that case we need API support in the CRD to explicitly set localities for specified endpoints (or we would have no way to know about them), but I'm not sure if there are any plans to implement that. So probably should comment about that instead, for clarity.

pilot/pkg/networking/core/v1alpha3/loadbalancer/loadbalancer.go
sangwa: These changes will indeed break backward compatibility for users who have SendUnhealthyEndpoints (explicitly) enabled. Locality failover will start to be implicitly applied to all services, even without outlier detection, and besides being a potentially unexpected change in the default behavior caused by upgrading the control plane, it can lead to service degradation if user workloads are unevenly distributed across localities and some locality won't be able to handle the increase in the prioritized traffic from local clients.

I also agree that having locality balancing implicitly enabled by default and popping up as a side effect of configuring outlier detection is unintuitive and the documentation does not seem very clear on that matter, so it's reasonable to have it disabled by default instead.

But this change will as well break backward compatibility for users who may already rely on that implicit behavior. It will also affect locality distribution balancing that is currently applied regardless of outlier detection but is controlled by the same switch. (Also it seems that locality distribution balancing should consider if an endpoint is unready when calculating locality weights with SendUnhealthyEndpoints enabled but that doesn't seem the case at the moment and likely should be fixed.)

I will consider changes that can maintain reasonable backward compatibility and get back to discuss shortly.

dm3ch: @howardjohn What about if we wouldn't change the default contract but would introduce a separate feature flag to enable failover for services without explicitly configured Outlier Detection?

howardjohn: @dm3ch what is the goal? To globally enable locality aware LB for every service? Or just an easy way to do it?

since this PR, we made it so if you set Service.spec.trafficDistribution=PreferClose (this is a core k8s flag) it does this (failover without outlier detection)

dm3ch: Ok, nice to hear. But am I right that using traffocDistribution=PreferClose again leads to enablement of traffic affinity per service. Is there a way to enable traffic affinity with zonal failover globally?

Issue #40325, opened 2022-08-08
add Secret CRD to support passing secrets over MCP


doujiang24:
**Describe the feature request**
Add Secret CRD to describe secrets, so that we could pass secrets over MCP.

**Describe alternatives you've considered**

```
apiVersion: networking.istio.io/v1alpha3
kind: Secret
metadata:
  name: mygateway
spec:
  workloadSelector:
    matchLabels:
      app: ratings
  serverCertificate: # string value of server certificate
  privateKey: # string value of private key
  caCertificates: # string value of ca cert
```

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[X] Security
[ ] Test and Release
[ ] User Experience
[X] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[X] Virtual Machine
[ ] Multi Control Plane

**Additional context**

Now the secrets only come from k8s, with fixed namespace take effects for a single secret resource.
After having the secret CRD, we could pass secrets over MCP from an alternative MCP server.
Also, having a flexible scope to take effect, with workloadSelector in the CRD.

So that, we could control the secrets in an upper-level control plane/dashboard, even without k8s.
(with MCP & the existing CRD, we could have almost control in the upper-level control plane, except secrets now.)

Any feedbacks are welcome, thanks!

howardjohn:
We can just allow k8s `core.v1.Secret`. no need to make a new type.

doujiang24:
Thanks, good idea, `core.v1.Secret` should be good enough, even without `workloadSelector`.

For the implementation details, these are my thoughts:

1. set `credentialName: mcp://bookinfo-secret` in the Gateway CR, with the `mcp://` prefix, since it prefix with `kubernetes://` by default when there is no prefix.
```
    tls:
      mode: SIMPLE # enables HTTPS on this port
      credentialName: mcp://bookinfo-secret
```

2. introduce the new `ISTIOD_ENABLE_SECRET_OVER_MCP` environment variable to enable the new secrets over MCP feature.

3. if enabled, we add a new resource type `mcp` here:
https://github.com/istio/istio/blob/f98481879ae4f869478ad900af6c96db9b8593bd/pilot/pkg/xds/sds.go#L415

4. to generate secret resources from the xDS client, eg `mcp://bookinfo-secret`.

Does this make sense? if yes, we will have a try. Thanks.

PR #40509, opened 2022-08-17
draft: support passing Secret source over MCP

doujiang24:
It's a draft for #40325 , and using k8s `core.v1.Secret` according to @howardjohn 's suggestion.

TODO:
1. need to authorize secret resources, simply skip for now, since the current authorize way depends on k8s, but the original intention is to bypass k8s; I haven't figured out a way for it yet.
2. tests and doc.

@howardjohn @hzxuzhonghu  Could you please take a look? Just want to make sure it's in the right direction.
Any feedbacks are welcome, thanks!

### DIFF
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -512,9 +512,8 @@ func (s *Server) initSDSServer() {
 		log.Warnf("skipping Kubernetes credential reader; PILOT_ENABLE_XDS_IDENTITY_CHECK must be set to true for this feature.")
 		return
 	}
-	var creds *kubecredentials.Multicluster
 	if s.kubeClient != nil {
-		creds = kubecredentials.NewMulticluster(s.clusterID)
+		creds := kubecredentials.NewMulticluster(s.clusterID)
 		creds.AddSecretHandler(func(name string, namespace string) {
 			s.XDSServer.ConfigUpdate(&model.PushRequest{
 				Full: false,
@@ -528,12 +527,14 @@ func (s *Server) initSDSServer() {
 				Reason: []model.TriggerReason{model.SecretTrigger},
 			})
 		})
+		s.XDSServer.Generators[v3.SecretType] = xds.NewSecretGen(creds, s.XDSServer.Cache, s.clusterID, s.environment.Mesh())
 		s.multiclusterController.AddHandler(creds)
 		if ecdsGen, found := s.XDSServer.Generators[v3.ExtensionConfigurationType]; found {
 			ecdsGen.(*xds.EcdsGenerator).SetCredController(creds)
 		}
+	} else {
+		s.XDSServer.Generators[v3.SecretType] = xds.NewSecretGen(nil, s.XDSServer.Cache, s.clusterID, s.environment.Mesh())
 	}
-	s.XDSServer.Generators[v3.SecretType] = xds.NewSecretGen(creds, s.XDSServer.Cache, s.clusterID, s.environment.Mesh())
 }
 
 // initKubeClient creates the k8s client if running in an k8s environment.
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -504,14 +504,17 @@ func (s *Server) WaitUntilCompletion() {
 
 // initSDSServer starts the SDS server
 func (s *Server) initSDSServer() {
-	if s.kubeClient == nil {
+	if s.kubeClient == nil && !features.EnableSecretOverMCP {
 		return
 	}
-	if !features.EnableXDSIdentityCheck {
+	if !features.EnableXDSIdentityCheck && !features.EnableSecretOverMCP {
 		// Make sure we have security
 		log.Warnf("skipping Kubernetes credential reader; PILOT_ENABLE_XDS_IDENTITY_CHECK must be set to true for this feature.")
-	} else {
-		creds := kubecredentials.NewMulticluster(s.clusterID)
+		return
+	}
+	var creds *kubecredentials.Multicluster
+	if s.kubeClient != nil {
+		creds = kubecredentials.NewMulticluster(s.clusterID)
 		creds.AddSecretHandler(func(name string, namespace string) {
 			s.XDSServer.ConfigUpdate(&model.PushRequest{
 				Full: false,
@@ -525,12 +528,12 @@ func (s *Server) initSDSServer() {
 				Reason: []model.TriggerReason{model.SecretTrigger},
 			})
 		})
-		s.XDSServer.Generators[v3.SecretType] = xds.NewSecretGen(creds, s.XDSServer.Cache, s.clusterID, s.environment.Mesh())
 		s.multiclusterController.AddHandler(creds)
 		if ecdsGen, found := s.XDSServer.Generators[v3.ExtensionConfigurationType]; found {
 			ecdsGen.(*xds.EcdsGenerator).SetCredController(creds)
 		}
 	}
+	s.XDSServer.Generators[v3.SecretType] = xds.NewSecretGen(creds, s.XDSServer.Cache, s.clusterID, s.environment.Mesh())
 }
 
 // initKubeClient creates the k8s client if running in an k8s environment.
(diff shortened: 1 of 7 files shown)


doujiang24:
Hi team, gentle ping.
Do you think it's the right way to pass `v1.Secret` through MCP?

cc @howardjohn @hzxuzhonghu

pilot/pkg/xds/sds.go
```diff
@@ -18,6 +18,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"istio.io/istio/pilot/pkg/credentials/kube"
```
zirain, line 21: please run `make gen & make lint`

hzxuzhonghu (commented):
Will it influence configcontroller with mcp

pilot/pkg/bootstrap/server.go
```diff
@@ -504,14 +504,17 @@ func (s *Server) WaitUntilCompletion() {
 
 // initSDSServer starts the SDS server
 func (s *Server) initSDSServer() {
-	if s.kubeClient == nil {
+	if s.kubeClient == nil && !features.EnableSecretOverMCP {
 		return
 	}
-	if !features.EnableXDSIdentityCheck {
+	if !features.EnableXDSIdentityCheck && !features.EnableSecretOverMCP {
```
howardjohn, line 510: we should still check this for MCP I think? I think you mentioned that in PR description though

pilot/pkg/model/push_context.go
```diff
@@ -197,6 +198,9 @@ type PushContext struct {
 	// gatewayIndex is the index of gateways.
 	gatewayIndex gatewayIndex
 
+	// secrets for each namespace
+	SecretsByNameSpace map[string]map[string]config.Config
```
howardjohn, line 202: what are the keys? namespace -> name -> config?

pilot/pkg/model/push_context.go
```diff
@@ -1265,6 +1276,12 @@ func (ps *PushContext) updateContext(
 		ps.ServiceAccounts = oldPushContext.ServiceAccounts
 	}
 
+	if secretChanged {
+		if err := ps.initSecrets(env); err != nil {
+			return err
+		}
+	}
```
howardjohn, line 1283: do we need `else { .. copy old ones ... }`?

pilot/pkg/model/push_context.go
```diff
@@ -1416,6 +1433,22 @@ func (ps *PushContext) initServiceRegistry(env *Environment) error {
 	return nil
 }
 
+// pre computes secrets per namespace
+func (ps *PushContext) initSecrets(env *Environment) error {
+	secretConfigs, err := env.List(gvk.Secret, NamespaceAll)
```
howardjohn, line 1438: Won't this get all secrets from all components? Including k8s?

pilot/pkg/model/push_context.go
howardjohn: should we just store a struct of NamespacedName?

howardjohn: and `*config.Config` possibly to reduce copies?

pilot/pkg/model/push_context.go
doujiang24: I think, no. It get configs from `model.ConfigStore`. but, seems the secrets from k8s, is on another path, `credentials.Controller`, which is not a kind of `model.ConfigStore`.

doujiang24:
> Will it influence configcontroller with mcp

@hzxuzhonghu Thanks.
I think it depends. when EnableSecretOverMCP = true,
1. if means only use secrets from MCP, I think they are two independent ways, they won't influence each other,
2. if means using secrets from k8s and MCP together, then we have to decide the priority, k8s first? or MCP first.

In the current implementation, we don't use secrets from k8s and MCP together, only use one.

pilot/pkg/xds/sds.go
doujiang24: Thanks, addressed the style issue. but I got such an error to run `make gen`.

```
bin/update_crds.sh
Cloning into '/tmp/tmp.FlyhketY9O'...
The authenticity of host 'github.com (140.82.114.3)' can't be established.
ED25519 key fingerprint is SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU.
This key is not known by any other names
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Could not create directory '/home/user/.ssh' (No such file or directory).
Failed to add the host to the list of known hosts (/home/user/.ssh/known_hosts).
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

Seems it's something unexpected in the base docker image, will take a deep look later.

hzxuzhonghu:
If we intend to use secret over MCP to provide gateway certs, it is a little bit complex. Even worse in multi cluster, we have to respect both secret namespace and cluster attribute.

From the original issue, I guess you want to use it to provide gateway cert. So it is just a first step for transport,  not for usage

doujiang24:
Thanks. All addressed.

I have two questions, looking forward to your help, thanks.

1. for the secrets, provide a new option way to bypass k8s totally, is the right way? or do you guys think it is worth maintaining it?

background: we are migrating a large & old north-south ingress, from Nginx to Istio + Envoy. And the old control plane is not in the k8s system. So, we hope the new Istio way does not rely on k8s, at least for the first step.
Also, we are processing well in our internal fork, in the bypass k8s way.

2. is it worth introducing a new secret controller?

As said in this TODO comment, so that we may register customer secret authorize method in the feature.
https://github.com/istio/istio/pull/40509/commits/8a21eaa253502dadc260db714a19cf7c92eb3b82#diff-91e0a18a58ea36d2c656d7b2706c52917a74103820c07fc843f48a110c847678R536

We just passing a `nil` secret controller now, which makes it difficult to add `authorize` for secrets.

doujiang24:
> If we intend to use secret over MCP to provide gateway certs, it is a little bit complex. Even worse in multi cluster, we have to respect both secret namespace and cluster attribute.
> 
> From the original issue, I guess you want to use it to provide gateway cert. So it is just a first step for transport, not for usage

Yeah, we want to pass secrets through MCP, so that the data plane could get the gateway cert from istio, and the related name of the gateway cert is defined in `credentialName` in `Gateway` CRD.
https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings

For the usage:
the current implementation (in this PR) does only honor the namespace and secret name, for now.
Missing the `authorize` method, that the biggest left issue for this PR, I think.

costinm:
I am concerned with this - we need a VERY strong security review/discussion first, messing with secrets is dangerous. 

Also: Istiod already has a mechanism to distribute secrets using SDS ( which is a first class XDS object ), with a relatively reasonable security model we discussed at length. I don't know if k8s Secret and envoy Secret are a good match - I don't 
mind exposing k8s Secret, but I think it must have the same security model, the caller must have some explicit permission for the secret.

costinm:
On a related note - we currently use the special generator for MCP, originally it was intended to keep things clean and it was an experiment, but I think it is very reasonable to allow 'MCP' generators to be used with the default (envoy) and envoy resources to be returned on the MCP generator. 

Only 'special' generator is proxyless - mainly because it requires a different format for LDS/RDS, but it also reuses EDS/CDS.

doujiang24:
Sorry, I think I made a mistake in the title, "passing" should be "accepting".

For istiod, we still deliver secret resources over SDS, just like it does now.
One thing that changed is istiod also accepts k8s secret from MCP server,
just like it accept k8s secret from k8s server.

Backgroud: we want to get rid of k8s server, in our north-south gateway.

Yeah, I totally agree with you about security.
I think we may introduce a new secret controller to implement the `authurize`.
But I'm not sure about it, as said in:
https://github.com/istio/istio/pull/40509#issuecomment-1225130704

Need your help, thanks.

doujiang24:
Thanks for your reply.

Sorry, I think I made a mistake in the title, "passing" should be "accepting".

For istiod, we still deliver secret resources over SDS, from istiod to envoy, just like it does now.
eg. using `credentialName` in the `Gateway` CRD, (https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings)

One thing that changed is istiod also accepts k8s secret from MCP server,
just like it accept k8s secret from k8s server.

Backgroud: we want to get rid of k8s server, in our north-south gateway.

Yeah, I totally agree with you about security.
I think we may introduce a new secret controller to implement the `authurize`.
But I'm not sure about it, as said in:
https://github.com/istio/istio/pull/40509#issuecomment-1225130704

Need your help, thanks.

PR #40523, opened 2022-08-17
istio: add metric for debouncing

yingzhuivy:
This CL adds metric for the delay between a first config change
enters deboucing until the final merged push request is pushed into
the push queue. This time plus the proxy convergence time give us
an upper bound on the total delay between a config change and the
change is pushed to proxies.

Also increased the buckets since logging shows that the debounce time
is pretty long (more than 1 minute).
```
{"level":"info","time":"2022-08-05T21:55:46.153204Z","scope":"ads","msg":"Push debounce stable[2555] 1193 for config ServiceEntry/balboa-production/balboa-production.balboa-production.svc.ea1.us.a and 182 more configs: 92.851119ms since last change, 1m9.085819569s since last push, full=true"}
```

### DIFF
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -870,10 +870,6 @@ func (s *DiscoveryServer) startPush(req *model.PushRequest) {
 			log.Debugf("Starting new push while %v were still pending", currentlyPending)
 		}
 	}
-
-	t := time.Since(req.DebounceStart)
-	debounceTime.Record(t.Seconds())
-
 	req.Start = time.Now()
 	for _, p := range s.AllClients() {
 		s.pushQueue.Enqueue(p, req)
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -871,8 +871,8 @@ func (s *DiscoveryServer) startPush(req *model.PushRequest) {
 		}
 	}
 
-	debounceTime := time.Since(req.DebounceStart)
-	debounceTime.Record(debounceTime.Seconds())
+	t := time.Since(req.DebounceStart)
+	debounceTime.Record(t.Seconds())
 
 	req.Start = time.Now()
 	for _, p := range s.AllClients() {
--- a/pilot/pkg/xds/ads.go
+++ b/pilot/pkg/xds/ads.go
@@ -870,6 +870,10 @@ func (s *DiscoveryServer) startPush(req *model.PushRequest) {
 			log.Debugf("Starting new push while %v were still pending", currentlyPending)
 		}
 	}
+
+	debounceTime := time.Since(req.DebounceStart)
+	debounceTime.Record(debounceTime.Seconds())
+
 	req.Start = time.Now()
 	for _, p := range s.AllClients() {
 		s.pushQueue.Enqueue(p, req)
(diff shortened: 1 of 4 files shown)


pilot/pkg/xds/monitoring.go
```diff
@@ -118,10 +118,16 @@ var (
...
+		"pilot_debounce_time",
+		"Delay in seconds between the first config enters debouncing and the merged push request is pushed into the push queue.",
+		[]float64{.01, .1, 1, 3, 5, 10, 20, 30, 40, 50, 60, 80, 100, 120},
+	)
+
 	pushContextInitTime = monitoring.NewDistribution(
 		"pilot_pushcontext_init_seconds",
 		"Total time in seconds Pilot takes to init pushContext.",
-		[]float64{.01, .1, 0.5, 1, 3, 5},
+		[]float64{.01, .1, 1, 3, 5, 10, 20, 30, 40, 50, 60, 80, 100, 120},
```
howardjohn, line 130: This seems like too broad of a range... if it takes over 5s something is terribly wrong -- we don't need to distinguish 30s (terrible) vs 40s (also terrible)?

pilot/pkg/xds/ads.go
```diff
@@ -870,6 +870,10 @@ func (s *DiscoveryServer) startPush(req *model.PushRequest) {
 			log.Debugf("Starting new push while %v were still pending", currentlyPending)
 		}
 	}
+
+	t := time.Since(req.DebounceStart)
+	debounceTime.Record(t.Seconds())
```
howardjohn, line 875: Is it feasible to just handle this entirely in the debounce() function? it seems leaky to put this in PushRequest

pilot/pkg/xds/monitoring.go
yingzhuivy: our logging shows that debounce time is taking more than 1minute (we use default max delay 10s) and this suggests to me that init push context is taking very long blocking the next push. unfortunately we don't have the push context init data right now in production (because it's not [registered](https://github.com/istio/istio/pull/40049/files)).

this seems very wrong to you? how about i change the buckets to
`.01, .1, 1, 3, 5, 10, 20, 30` similar to all the others

howardjohn: IMO metrics buckets do not need super high granularity at outlier levels. They just need to indicate "something is wrong, it is super high!" to alert to look into it with more detailed data (logs, profiles, etc).

1min seems excessively long

yingzhuivy: That makes sense, for the debounce metric, I'm gonna use `.01, .1, 1, 3, 5, 10, 20, 30` which is the same as the other push time metrics (since 10s is pretty normal given the default max delay), for this init context one, do you suggest me to keep original?

howardjohn: Yeah I think original one for init and what you mentioned for debounce is ok. thanks!

yingzhuivy: updated

pilot/pkg/xds/ads.go
yingzhuivy: updated

yingzhuivy:
@howardjohn please take another look, thanks!

howardjohn (approved):

PR #40524, opened 2022-08-17
Creating image pull secrets for GatewayConformance test namespaces.

stewartbutler:
Precreates the namespaces used by the GatewayConformance tests (which are sourced from upstream Kubernetes). Then creates the ImagePullSecrets for each of the GC namespaces, and modifies the default ServiceAccount to use them.

This allows the test to correctly work when the images under test are stored in a private docker registry.

### DIFF
--- a/tests/integration/pilot/gateway_conformance_test.go
+++ b/tests/integration/pilot/gateway_conformance_test.go
@@ -79,6 +79,16 @@ func TestGatewayConformance(t *testing.T) {
 		Run(func(ctx framework.TestContext) {
 			DeployGatewayAPICRD(ctx)
 
+			// Precreate the GatewayConformance namespaces, and apply the Image Pull Secret to them.
+			if ctx.Settings().Image.PullSecret != "" {
+				for _, ns := range conformanceNamespaces {
+					namespace.Claim(ctx, namespace.Config{
+						Prefix: ns,
+						Inject: false,
+					})
+				}
+			}
+
 			mapper, _ := gatewayConformanceInputs.Client.UtilFactory().ToRESTMapper()
 			c, err := client.New(gatewayConformanceInputs.Client.RESTConfig(), client.Options{
 				Scheme: kube.IstioScheme,
--- a/pkg/test/framework/components/namespace/kube.go
+++ b/pkg/test/framework/components/namespace/kube.go
@@ -239,6 +239,14 @@ func (n *kubeNamespace) createInCluster(c cluster.Cluster, cfg Config) error {
 		if err := c.ApplyYAMLFiles(n.name, s.Image.PullSecret); err != nil {
 			return err
 		}
+		_, err := c.Kube().CoreV1().ServiceAccounts(n.name).Patch(context.TODO(),
+			"default",
+			types.JSONPatchType,
+			[]byte(`[{"op": "add", "path": "/imagePullSecrets", "value": [{"name": "test-gcr-secret"}]}]`),
+			metav1.PatchOptions{})
+		if err != nil {
+			return err
+		}
 	}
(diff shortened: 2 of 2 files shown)


tests/integration/pilot/gateway_conformance_test.go
```diff
@@ -79,6 +86,48 @@ func TestGatewayConformance(t *testing.T) {
...
 			DeployGatewayAPICRD(ctx)
 
+			// Precreate the GatewayConformance namespaces, and apply the Image Pull Secret to them.
+			if ctx.Settings().Image.PullSecret != "" {
+				for _, ns := range conformanceNamespaces {
+					namespace.Claim(ctx, namespace.Config{
+						Prefix: ns,
+						Inject: false,
+					})
+					// Cannot apply the existing secret directly; need to sanitize the
```
howardjohn, line 96: Why not? We do it in `(n *kubeNamespace) createInCluster`

howardjohn: This seems like the input is just wrong...?

tests/integration/pilot/gateway_conformance_test.go
```diff
@@ -79,6 +86,48 @@ func TestGatewayConformance(t *testing.T) {
...
+						t.Fatalf("Error unmarshalling yaml into secret: %s", err)
+					}
+					n := secret.ObjectMeta.Name
+					secret.ObjectMeta.Reset()
+					secret.ObjectMeta.SetName(n)
+					yml, err := yaml.Marshal(secret)
+					if err != nil {
+						t.Fatalf("Error marshalling secret: %s", err)
+					}
+					ctx.ConfigKube().YAML(ns, string(yml)).ApplyOrFail(t, apply.NoCleanup)
```
howardjohn, line 116: can we do this in `claimKube`?

howardjohn: Maybe with a bool option if we need it

tests/integration/pilot/gateway_conformance_test.go
stewartbutler: The specific error I encountered was something like:
```
operation cannot be fulfilled on secrets: the object has been modified; please apply your changes to the latest version and try again
```

From what I understand, this happens when you apply a resource with a resourceVersion in its metadata that differs from the original. In this case, the resource that the PullSecret points to appears to already have that data in it, so to bypass the issue I import it as an object, sanitize everything but the name from the metadata, and then reapply it in the right namespace.

tests/integration/pilot/gateway_conformance_test.go
stewartbutler: Yeah, let me rework it.

tests/integration/pilot/gateway_conformance_test.go
howardjohn: Why don't we just point to a PullSecret that is valid already?

stewartbutler: Actually, it may be that createInCluster creates the same secret already, so the apply I was making was failing because it had a different resourceVersion specified. Let me check if the rewrite fixes it.

tests/integration/pilot/gateway_conformance_test.go
stewartbutler: That wound up being a much cleaner fix, good call.

tests/integration/pilot/gateway_conformance_test.go
stewartbutler: Fixed in rewrite.

howardjohn (approved):

PR #40531, opened 2022-08-17
Changing default value of ENABLE_LEGACY_FS_GROUP_INJECTION in charts

kfaseela:

This was missed when https://github.com/istio/istio/pull/38724 was merged.
Release notes was previously added indicating the default value is changed. But without the current
change it was actually not in effect.

- [X] Configuration Infrastructure
- [ ] Docs
- [X] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/kube/inject/testdata/inject/explicit-security-context.yaml.injected
+++ b/pkg/kube/inject/testdata/inject/explicit-security-context.yaml.injected
@@ -177,7 +177,7 @@ spec:
           runAsNonRoot: false
           runAsUser: 0
       securityContext:
-        fsGroup: 1337
+        fsGroup: 1234
       volumes:
       - name: workload-socket
       - name: credential-socket
--- a/pkg/kube/inject/testdata/inject/user-volume.yaml.injected
+++ b/pkg/kube/inject/testdata/inject/user-volume.yaml.injected
@@ -192,8 +192,6 @@ spec:
           runAsGroup: 0
           runAsNonRoot: false
           runAsUser: 0
-      securityContext:
-        fsGroup: 1337
       volumes:
       - name: workload-socket
       - name: credential-socket
--- a/manifests/charts/istio-control/istio-discovery/files/gateway-injection-template.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/gateway-injection-template.yaml
@@ -223,7 +223,7 @@ spec:
     - name: {{ . }}
     {{- end }}
   {{- end }}
-  {{- if eq (env "ENABLE_LEGACY_FSGROUP_INJECTION" "true") "true" }}
+  {{- if eq (env "ENABLE_LEGACY_FSGROUP_INJECTION" "false") "true" }}
   securityContext:
     fsGroup: 1337
   {{- end }}
--- a/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
+++ b/manifests/charts/istio-control/istio-discovery/files/gen-istio.yaml
@@ -698,7 +698,7 @@ data:
             - name: {{ . }}
             {{- end }}
           {{- end }}
-          {{- if eq (env "ENABLE_LEGACY_FSGROUP_INJECTION" "true") "true" }}
+          {{- if eq (env "ENABLE_LEGACY_FSGROUP_INJECTION" "false") "true" }}
           securityContext:
             fsGroup: 1337
           {{- end }}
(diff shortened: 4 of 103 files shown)


howardjohn (approved):

pkg/kube/inject/testdata/inject/explicit-security-context.yaml.injected
```diff
@@ -177,7 +177,7 @@ spec:
           runAsNonRoot: false
           runAsUser: 0
       securityContext:
-        fsGroup: 1337
+        fsGroup: 1234
```
AndreaM12345, line 180: curious why the change to 1234 here but not seeing a corresponding verification change?

pkg/kube/inject/testdata/inject/user-volume.yaml.injected
```diff
@@ -192,8 +192,6 @@ spec:
           runAsGroup: 0
           runAsNonRoot: false
           runAsUser: 0
-      securityContext:
-        fsGroup: 1337
```
AndreaM12345, line 196: noticed that the scripts removed a lot of the securityContext sections in the testdata  I think the behavior wanted is that if fsGroup is set then it would stay what the user wanted instead of being over-written with 1337.  However was unclear if there is no fsGroup set do we want it to default to 1337 or not have a value?

kfaseela: @AndreaM12345 : testdata was auto-generated when I ran makegen.. Even before this PR, I think the expectation was that we set fsGroup to 1337 only when the feature flag is enabled. I understand your point that by default it was having fsGroup set even in the absence of flag, but that is because the default was to enable the fsGroupInjection?

pkg/kube/inject/testdata/inject/explicit-security-context.yaml.injected
howardjohn:
```
pkg/kube/inject/testdata/inject/explicit-security-context.yaml
19:        fsGroup: 1234
```

I think this is WAI. Before we had to override users fsGroup, now we let it remain

howardjohn:
@kfaseela this is broken because https://github.com/istio/istio/blob/e0110ff89739f8dc15b69c4a9a3c53854bb57ca1/operator/pkg/manifest/shared.go#L447 will only set `true` -- it will never set it to `false`. As a result, older k8s versions will be broken (see test failures https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-116_istio_postsubmit/1560006373633167360, etc). I think we need to revert or enhance that function

ericvn:
@howardjohn Do we need to revert the original PR as well that says we are setting the default to false as I don't think that is true either.

kfaseela:
> @howardjohn Do we need to revert the original PR as well that says we are setting the default to false as I don't think that is true either.

https://github.com/istio/istio/pull/40549

kfaseela:
> @kfaseela this is broken because
> 
> https://github.com/istio/istio/blob/e0110ff89739f8dc15b69c4a9a3c53854bb57ca1/operator/pkg/manifest/shared.go#L447
> 
> will only set `true` -- it will never set it to `false`. As a result, older k8s versions will be broken (see test failures https://prow.istio.io/view/gs/istio-prow/logs/integ-k8s-116_istio_postsubmit/1560006373633167360, etc). I think we need to revert or enhance that function

As discussed offline, the logic is setting the value to true only for k8s version older than 1.19, which is correct. But I understand from you that this logic applies only for IstioOperator, and helm does not have it. The below PR hopefully solves the problem for helm.

AndreaM12345:
@howardjohn just for clarification it sounded like  the value of ENABLE_LEGACY_FSGROUP_INJECTION should be `true` in the case of a k8s clusters with a version of 1.18 or older unless the  JWT_POLICY is "first-party-jwt"  So while I understand that the `getFSGroupOverlay` function might need to be enhanced to also return `false` in the case of newer k8s clusters...I'm not clear why the older k8s versions are broken because the value of `true` was returned?   Thinking that the older k8s versions of the cluster are broken because maybe the `getFSGroupOverlay` function is not called at the correct time and so its not setting the environment variable to true?  Previous to the changes the default would have been true.

kfaseela:
> @howardjohn just for clarification it sounded like the value of ENABLE_LEGACY_FSGROUP_INJECTION should be `true` in the case of a k8s clusters with a version of 1.18 or older unless the JWT_POLICY is "first-party-jwt" So while I understand that the `getFSGroupOverlay` function might need to be enhanced to also return `false` in the case of newer k8s clusters...I'm not clear why the older k8s versions are broken because the value of `true` was returned? Thinking that the older k8s versions of the cluster are broken because maybe the `getFSGroupOverlay` function is not called at the correct time and so its not setting the environment variable to true? Previous to the changes the default would have been true.

You are right. See my comment: https://github.com/istio/istio/pull/40531#issuecomment-1219728704

Issue #40394, opened 2022-08-11
Support multiple addresses per listener to generate LDS configuration in Istio


zhlsunshine:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**
As feature [issue#11184](https://github.com/envoyproxy/envoy/issues/11184) in envoy has been completed. According to current implementation, the field of `AdditionalAddresses` has been added into the `Listener` struct. It means that one listener can be used for multiple addresses in envoy. 
This feature can be used by Istio which is deployed in dual-stack k8s cluster. The services in the same cluster can have 2 cluster IP addresses in this scenario, the one is IPv4 and the other is IPv6.
In general, Isito can implement dual stack based on this envoy feature, moreover, it also can help to eliminate duplicated configuration for current dual stack solution which is implemented in extra branch `experimental-dual-stack` of Istio.

**Describe alternatives you've considered**
There is an implementation for dual stack support in extra branch `experimental-dual-stack` of Istio. But it imports duplicated configuration and inflexible.

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

hzxuzhonghu:
One case useful i can think of for now is for tcp headless service, we build one listener per pod instance. With the new version,  we can build only one listener.

zhlsunshine:
> One case useful i can think of for now is for tcp headless service, we build one listener per pod instance. With the new version, we can build only one listener.

Hi @hzxuzhonghu, yeah, that's a good use case. However, we have to consider how to implement traffic route if the network session is stateful.

kyessenov:
@hzxuzhonghu Are you sure the listener update is in-place? Otherwise, you're draining connections any time another endpoint stands up for a headless service.

hzxuzhonghu:
I am not sure, i donot have a chance to look at that yet

kyessenov:
It'd be strictly worse without support for delta updates for additional addresses. I haven't checked either.

costinm:
For dual stack - clearly good idea. 

I think for stateful sets it is also strictly better than what we have today.

[...]

PR #40409, opened 2022-08-11
wip: experimental with single listener for headless

howardjohn:
**Please provide a description of this PR:**

PR #40539, opened 2022-08-18
Initialized PR for support multiple addresses per listener to generate LDS configuration in Istio

zhlsunshine:
Initialized PR for support multiple addresses per listener to generate LDS configuration in Istio for [issue#40394](https://github.com/istio/istio/issues/40394)


pilot/pkg/model/service.go
```diff
@@ -774,6 +778,21 @@ func (s *Service) GetAddressForProxy(node *Proxy) string {
...
 }
 
+// GetExtraSvcAddresses returns a Service's extra addresses to the cluster where the node resides.
+// Especially for dual stack k8s service to get other IP family addresses.
+func (s *Service) GetExtraSvcAddresses(node *Proxy) []string {
+	if len(s.ExtraSvcAddresses) == 0 && node.Metadata != nil {
+		if node.Metadata.ClusterID != "" {
+			addresses := s.ClusterVIPs.GetAddressesFor(node.Metadata.ClusterID)
+			if len(addresses) > 1 {
+				s.ExtraSvcAddresses = addresses[1:]
```
howardjohn, line 788: Mutating Service here isn't safe

pilot/pkg/model/service.go
```diff
@@ -78,6 +78,10 @@ type Service struct {
 	// Do not access directly. Use GetAddressForProxy
 	DefaultAddress string `json:"defaultAddress,omitempty"`
 
+	// ExtraDefaultAddresses specifies the extra service addresses of the load balancer.
+	// This field can be used by dual stack service to store other IP family addresses.
+	// Do not access directly. Use GetExtraSvcAddresses
+	ExtraSvcAddresses []string `json:"extraSvcAddresses,omitempty"`
```
howardjohn, line 84: `ClusterVIPs` already holds a map of `Cluster->[]Address`. What does it mean to have a singular `ExtraSvcAddresses` when we can have multiple clusters?

It seems like `ClusterVIPs` _already_ holds the correct data. The issue is more that `GetAddressForProxy` ignores all of them (`return addresses[0]`). It seems like we could have an alternative method `GetAddressesForProxy` that returns all of them...? WDYT?

Long term we can remove GetAddressForProxy, but its probably hard to update every single call site at once to support multiple

pilot/pkg/model/service.go
zhlsunshine: Yeah, based on below comment, there is no need to mutate the service here.

pilot/pkg/model/service.go
zhlsunshine: Hi @howardjohn, yeah, it's a great suggestion,  `ClusterVIPs` have all necessary addresses. BTW, there is an existing function named `getAllAddresses()` can returns all addresses of the service, so there are 2 options:

- [ ] 1. I can change the name as `GetAllAddresses()`
- [ ] 2. Create a new function named `GetAddressesForProxy` and pass the proxy node as its parameter

Which is the better?

zhlsunshine: Hi @howardjohn, I think the `ExtraSvcAddresses` is needed for the Istio service converted from service entry, especially in dual stack cluster. Because the `ClusterVIPs` of Istio service is empty in this situation.

howardjohn:
Hm, I thought ServiceEntry already supports multiple addresses? I'll need
to see how that works

zhlsunshine:
> Hm, I thought ServiceEntry already supports multiple addresses? I'll need to see how that works

Hi @howardjohn, I think it's another multiple addresses support for service entry, dual stack ip addresses support. I think [here](https://github.com/istio/istio/blob/0a5b8c148b1bd90601702a0683bb008ccf7e4436/pilot/pkg/serviceregistry/serviceentry/conversion.go#L200) is the root cause for the problem.

zhlsunshine:
Hi @howardjohn,  new submit based on your comment, discussed as above, this PR just focus on k8s service converting into Istio service. And there should be other changes for service entry.

hzxuzhonghu:
Have we made sure about  https://github.com/istio/istio/issues/40394#issuecomment-1212705007 before implement?

zhlsunshine:
> Have we made sure about [#40394 (comment)](https://github.com/istio/istio/issues/40394#issuecomment-1212705007) before implement?

Hi @hzxuzhonghu, I think this comment should relate to another PR from @howardjohn : https://github.com/istio/istio/pull/40409. And yes, we need to consider more for both headless and service entry.

howardjohn (approved):

PR #39062, opened 2022-05-20
Retry for adding remote cluster

miaoyq:
**Please provide a description of this PR:**

Fix #38988
/cc @hzxuzhonghu @howardjohn

PR #40542, opened 2022-08-18
Retry for adding remote cluster

hzxuzhonghu:

Take over #39062

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/config/kube/crdclient/client.go
+++ b/pilot/pkg/config/kube/crdclient/client.go
@@ -29,7 +29,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/cenkalti/backoff/v4"
 	jsonmerge "github.com/evanphx/json-patch/v5"
 	"github.com/hashicorp/go-multierror"
 	"go.uber.org/atomic"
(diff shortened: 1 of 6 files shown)


pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -151,7 +151,7 @@ func DetectEndpointMode(kubeClient kubelib.Client) EndpointMode {
 	useEndpointslice, ok := features.EnableEndpointSliceController()
 
 	// we have a client, and flag wasn't set explicitly, auto-detect
-	if kubeClient != nil && !ok && kubelib.IsAtLeastVersion(kubeClient, 21) {
+	if kubeClient != nil && !ok && !kubelib.IsLessThanVersion(kubeClient, 21) {
```
ramaraochavali, line 154: Why this change? Also can we reuse the same function/check at both places?

pkg/kube/multicluster/secretcontroller.go
```diff
@@ -46,6 +48,7 @@ import (
 
 const (
 	MultiClusterSecretLabel = "istio/multiCluster"
+	maxRetries              = 15
```
ramaraochavali, line 51: Is this some arbitrary number?

hzxuzhonghu:
Something need to be fixed, this is all from https://github.com/istio/istio/pull/39062

pilot/pkg/serviceregistry/kube/controller/controller.go
hzxuzhonghu: This is because   when err happens , it may detect the k8s versions error, in this case if the k8s cluster is 1.20, then we could wrongly use endpointslice.
```
func IsAtLeastVersion(client Client, minorVersion uint) bool {
	clusterVersion, err := client.GetKubernetesVersion()
	if err != nil {
		return true
	}
	return IsKubeAtLeastOrLessThanVersion(clusterVersion, minorVersion, true)
}
```

hzxuzhonghu:
@stevenctl Ptal

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -423,24 +422,14 @@ func (cl *Client) kind(r config.GroupVersionKind) (*cacheHandler, bool) {
 // knownCRDs returns all CRDs present in the cluster, with timeout and retries.
 func knownCRDs(crdClient apiextensionsclient.Interface) (map[string]struct{}, error) {
 	var res *crd.CustomResourceDefinitionList
-	b := backoff.NewExponentialBackOff()
```
stevenctl, line 426: What's the reasoning behind this change?

pkg/kube/controllers/queue.go
```diff
@@ -157,8 +157,9 @@ func (q Queue) processNextItem() bool {
 
 	err := q.workFn(key)
 	if err != nil {
-		if q.queue.NumRequeues(key) < q.maxAttempts {
-			q.log.Errorf("error handling %v, retrying: %v", key, err)
+		retryCount := q.queue.NumRequeues(key)
```
stevenctl, line 160: Does numRequeues get reset after we successfully process it? We might have > maxAttempts updates to the same key.

stevenctl: Looks like it will behave how we want.

stevenctl:
LGTM, but why did we remove the knownCRDs retry? I don't think we go through the multi-cluster retries for all usages of that?

pilot/pkg/config/kube/crdclient/client.go
hzxuzhonghu: We will  retry at the upper level

hzxuzhonghu:
@stevenctl I confirmed， there are 3 kinds of usage:

1. during pilot bootstrap
2.  multicluster
3. CLI analysis
So I think removing the retry in function does not matter

stevenctl:
Hmm. I think the pilot bootstrap case probably gets value but I don't have enough context about why we added the retry to begin with.

stevenctl (approved):

PR #40547, opened 2022-08-18
install: autodetect cniBinDir on GKE

howardjohn:
Makes users not need
https://istio.io/latest/docs/setup/additional-setup/cni/#hosted-kubernetes-settings
when installing via `helm install` or `istioctl install`.

Note: `helm template` and `istioctl manifest generate` do not use
version information, so this is still needed for those cases.

This adds GKE detection. OpenShift detection is possible as well, I just
don't have a test cluster or know how to detect OpenShift

### DIFF
--- a/operator/pkg/manifest/shared.go
+++ b/operator/pkg/manifest/shared.go
@@ -437,8 +437,38 @@ func getClusterSpecificValues(client kube.Client, force bool, l clog.Logger) (st
 	} else {
 		overlays = append(overlays, jwtStr)
 	}
-
+	cni := getCNISettings(client)
+	if cni != "" {
+		overlays = append(overlays, cni)
+	}
 	return makeTreeFromSetList(overlays)
+}
+
+// getCNISettings gets auto-detected values based on the Kubernetes environment.
+// Note: there are other settings as well; however, these are detect inline in the helm chart.
+// This ensures helm users also get them.
+func getCNISettings(client kube.Client) string {
+	ver, err := client.GetKubernetesVersion()
+	if err != nil {
+		return ""
+	}
+	// https://istio.io/latest/docs/setup/additional-setup/cni/#hosted-kubernetes-settings
+	// GKE requires deployment in kube-system namespace.
+	if strings.Contains(ver.GitVersion, "-gke") {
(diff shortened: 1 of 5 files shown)


jacob-delgado (approved):

jacob-delgado:
Should we cherry-pick this into release 1.15 and release 1.14? Possibly 1.13 as well? We'd love to get this and OCP into 1.14 ourselves.

howardjohn:
> We'd love to get this and OCP into 1.14 ourselves.

Would be great to have the OCP equivalent added if you/someone has time!

I don't mind backport too much. Not sure how risky this is -- it feels pretty safe to me. Existing users should already have these fields explicitly set, so this change is NOP... hopefully

costinm (approved):
Great change !

operator/pkg/manifest/shared.go
```diff
@@ -441,9 +441,30 @@ func getClusterSpecificValues(client kube.Client, force bool, l clog.Logger) (st
...
 	}
+	cni := getCNISettings(client)
+	if cni != "" {
+		overlays = append(overlays, cni)
+	}
 	return makeTreeFromSetList(overlays)
 }
 
+// getCNISettings gets auto-detected values based on the Kubernetes environment.
+// Note: there are other settings as well; however, these are detect inline in the helm chart.
```
kfaseela, line 452: nit: these are detected

operator/pkg/manifest/shared.go
```diff
@@ -441,9 +441,30 @@ func getClusterSpecificValues(client kube.Client, force bool, l clog.Logger) (st
...
+// This ensures helm users also get them.
```
kfaseela, line 453: From the previous issue with fsGroup injection, this call hierarchy was basically for operator and istioctl install ? So I got a little confused with this comment.

howardjohn: Helm users MUST explicitly set namespace; in istioctl they do not need to. So this just detects namespace.

kfaseela: Okay, thanks for the clarification.

PR #40549, opened 2022-08-18
Add automatic setting logic for ENABLE_LEGACY_FSGROUP_INJECTION

kfaseela:
The previous logic was available only for IstioOperator,
hence helm based integration tests on k8s 1.16 were failing.

- [ ] Docs
- [X] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
+++ b/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
@@ -149,6 +149,10 @@ spec:
             value: "{{ $val }}"
           {{- end }}
           {{- end }}
+{{- if semverCompare "<1.19" .Capabilities.KubeVersion.GitVersion }}
+          - name: ENABLE_LEGACY_FSGROUP_INJECTION
+            value: "true"
+{{- end }}
 {{- if .Values.pilot.traceSampling }}
           - name: PILOT_TRACE_SAMPLING
             value: "{{ .Values.pilot.traceSampling }}"


howardjohn (approved):

manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
```diff
@@ -149,6 +149,10 @@ spec:
             value: "{{ $val }}"
           {{- end }}
           {{- end }}
+{{- if semverCompare "<1.19" .Capabilities.KubeVersion.GitVersion }}
+          - name: ENABLE_LEGACY_FSGROUP_INJECTION
```
AndreaM12345, line 153: does a check for JWT Policy values need to be added too?

howardjohn: It doesn't _need_ to be false -- it just _can_ be. IMO its not needed; this is strictly an improvement for Helm. While we could add the jwt check in helm as well, the jwt change is from Kubernetes 1.12 which is ~forever ago, so I am not sure its worth the effort

AndreaM12345: okay thanks for the clarification

ericvn:
Do we need this in release-1.15?

kfaseela:
@ericvn : I would wait till all the post submit tests pass before putting to 1.15

Issue #38716, opened 2022-05-04
Istiod memory leak with 1.13.3


JeffreyWa:
### Bug Description

Istiod memory and cpu consumption are growing continuously.
% k top po -n istio-system
NAME                                     CPU(cores)   MEMORY(bytes)
authservice-0                            1m           6Mi
cluster-local-gateway-64f58f66cb-vz2pd   27m          84Mi
istio-egressgateway-5c597cdb77-jdhv6     25m          70Mi
istio-ingressgateway-8d7d49b55-fjvvx     28m          76Mi
istiod-54c54679d7-9559c                  2826m        6149Mi

top - 00:46:48 up 14:45,  0 users,  load average: 3.43, 2.59, 2.91
Tasks:   5 total,   1 running,   4 sleeping,   0 stopped,   0 zombie
%Cpu(s): 12.9 us,  0.6 sy,  0.0 ni, 85.7 id,  0.0 wa,  0.0 hi,  0.4 si,  0.4 st
MiB Mem : 128845.1 total, 110897.2 free,   8045.7 used,   9902.2 buff/cache
MiB Swap:      0.0 total,      0.0 free,      0.0 used. 119708.5 avail Mem 

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                                                                                               
    1 istio-p+  20   0   10.1g   5.2g  66572 S 165.8   4.1   1413:06 pilot-discovery                                                                                                                       
   39 istio-p+  20   0    6104   3288   2776 R   0.3   0.0   0:00.01 top                                                                                                                                   
   29 istio-p+  20   0    2616    604    528 S   0.0   0.0   0:00.01 sh                                                                                                                                    
   35 istio-p+  20   0    2616    104      0 S   0.0   0.0   0:00.00 sh                                                                                                                                    
   36 istio-p+  20   0    4248   3552   2952 S   0.0   0.0   0:00.00 bash                                                                                                                                  

### Version

```prose
k8s version: 1.21.1
istio version: 1.13.3
kubeflow version: 1.5.0
```

JeffreyWa:
BTW, I am testing in lab env. So there isn't traffic from outside.

JeffreyWa:
k top po -n istio-system
NAME                                     CPU(cores)   MEMORY(bytes)
authservice-0                            1m           6Mi
cluster-local-gateway-64f58f66cb-vz2pd   21m          83Mi
[...]

Issue #24521, opened 2020-06-08
high cpu usage for workload entry registration


anson627:
**Bug description**
Istio CPU usage is high when there is a large amount of workload entry registration.

For example, istiod pod was throttled by cpu limit (8 cpu cores), when workload entry was created at QPS of 1k/s, uniformly distributed among 1k namespaces.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[x] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Expected behavior**
Pilot can handle registration of appropriate number of WorkloadEntry instances, and gracefully handles high churn of WorkloadEntry instances.

**Steps to reproduce the bug**
wrote a simple program to concurrently create workload entry at 1k/s QPS, uniformly under 1k namespaces.

have we done any perf testing as the test plan mentioned in the [design doc](https://docs.google.com/document/d/10TyPy-6bOs0_7JljGpc2RneDdsaR9GktjwwaaVbz6fs/edit#).

I also followed the instruction [here](https://github.com/istio/istio/wiki/Analyzing-Istio-Performance) and did some cpu profiling. 
[pprof.pilot-discovery.samples.cpu.003.pb.gz](https://github.com/istio/istio/files/4748775/pprof.pilot-discovery.samples.cpu.003.pb.gz)

**Version (include the output of `istioctl version --remote` and `kubectl version` and `helm version` if you used Helm)**
```
istioctl version --remote
client version: 1.6.0
control plane version: 1.6.0
data plane version: none
```

```
kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T23:35:15Z", GoVersion:"go1.14.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.12-airbnb0", GitCommit:"da578f6bba72b15d934dd1a9d53acbc2c61863e7", GitTreeState:"clean", BuildDate:"2020-05-27T13:58:10Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
```

**How was Istio installed?**
istioctl to generate manifest
kubectl to apply generated yamls
[...]

PR #40561, opened 2022-08-19
Disable CONFIG_DISTRIBUTION_TRACKING

hzxuzhonghu:

This is to disable config distribution tracking, which is not mature, especially on CPU andMem cost.

https://github.com/istio/istio/issues/24521
https://github.com/istio/istio/issues/38716#issuecomment-1119187983

It is used for debug interface, so I think not much risk

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -220,7 +220,7 @@ var (
 
 	EnableDistributionTracking = env.RegisterBoolVar(
 		"PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING",
-		true,
+		false,
 		"If enabled, Pilot will assign meaningful nonces to each Envoy configuration message, and allow "+
 			"users to interrogate which envoy has which config from the debug interface.",
 	).Get()


ramaraochavali:
Should this have a release notes?

howardjohn:
I think it's also used by `istioctl wait ` so it's not just debug interface.

I am not necessarily opposed to it though as `wait` also has some unaddressed issues making it unreliable

hzxuzhonghu:
Yeah, since we all agreed to disable it, will add a release note

ramaraochavali (approved):

releasenotes/notes/40561.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: traffic-management
+
+releaseNotes:
+  - |
+    **Fixed** disable `PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING` temporarily to fix some memory/CPU cost issues.
```
howardjohn, line 7: we should not this breaks `istioctl x wait`. I would make this `**Removed** default `PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING` due to performance issues; this also makes `istioctl x wait` not functional`.

howardjohn (approved):

therealmitchconnors:
As pointed out, this change breaks istioctl functionality, which is relied on widely in docs tests.  We should either revert this commit or update our docs and go through the process of deprecating the command.

hzxuzhonghu:
We can definitely disable doc tests or set env for the test

ericvn:
This did indeed cause a test failure: https://github.com/istio/istio.io/issues/11780

That is a release blocker PR as https://preliminary.istio.io/latest/docs/ops/configuration/mesh/config-resource-ready/ is no longer accurate for the main branch. I'm not sure if it's adequate so have the user have to add another set or if that would work.

PR #40568, opened 2022-08-20
Consider the health of remote endpoints when applying the Split Horizon EDS

sangwa:

I have discovered a bug when `PILOT_SEND_UNHEALTHY_ENDPOINTS` is set to `true` and in a multi-cluster, multi-network configuration for a multi-cluster service the cluster endpoints of remote network gateways are neither removed nor marked UNHEALTHY if all endpoints of that service in the respective remote network are UNHEALTHY.

This leads to Envoy sending traffic to unready remote pods if e.g. the locality failover is disabled or set to distribute or local pods of that service are also all unready. It also obviously breaks the contract of Kubernetes of not sending traffic to unready  pods under those conditions.

The current implementation of the split horizon EDS filter is not aware of the SendUnhealthyEndpoints feature and does not handle the case when all endpoints for a service in a remote network are unhealthy. I suppose the proper behavior would be marking the remote gateway endpoints unhealthy if all endpoints in that network are unhealthy.

This patch implements the corresponding fix.

### DIFF
--- a/pilot/pkg/xds/ep_filters.go
+++ b/pilot/pkg/xds/ep_filters.go
@@ -17,6 +17,7 @@ package xds
 import (
 	"math"
 
+	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	endpoint "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
 	"google.golang.org/protobuf/proto"
 	wrappers "google.golang.org/protobuf/types/known/wrapperspb"
@@ -64,6 +65,9 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocalityEndpoint
 		// Create a map to keep track of the gateways used and their aggregate weights.
 		gatewayWeights := make(map[model.NetworkGateway]uint32)
 
+		// Create a map to keep track of aggregate health of the gateways.
+		gatewayHealth := make(map[model.NetworkGateway]bool)
+
 		// Process all of the endpoints.
 		for i, lbEp := range ep.llbEndpoints.LbEndpoints {
 			istioEndpoint := ep.istioEndpoints[i]
@@ -103,6 +107,13 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocalityEndpoint
 				continue
 			}
 
+			// A gateway is considered healthy only if it has any healthy endpoints reachable.
+			if istioEndpoint.HealthStatus == model.Healthy {
+				for _, gw := range gateways {
+					gatewayHealth[gw] = true
+				}
+			}
+
 			// Apply the weight for this endpoint to the network gateways.
 			splitWeightAmongGateways(weight, gateways, gatewayWeights)
 		}
@@ -123,9 +134,19 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocalityEndpoint
 			}
 			epAddr := util.BuildAddress(gw.Addr, gw.Port)
 
+			// Set the gateway endpoint's health status from the aggregate health of that gateway.
+			healthStatus := core.HealthStatus_UNHEALTHY
+			epHealthStatus := model.UnHealthy
+
+			if gatewayHealth[gw] {
+				healthStatus = core.HealthStatus_HEALTHY
+				epHealthStatus = model.Healthy
+			}
+
 			// Generate a fake IstioEndpoint to carry network and cluster information.
 			gwIstioEp := &model.IstioEndpoint{
-				Network: gw.Network,
+				HealthStatus: epHealthStatus,
+				Network:      gw.Network,
 				Locality: model.Locality{
 					ClusterID: gw.Cluster,
 				},
@@ -134,6 +155,7 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocalityEndpoint
 
 			// Generate the EDS endpoint for this gateway.
 			gwEp := &endpoint.LbEndpoint{
+				HealthStatus: healthStatus,
 				HostIdentifier: &endpoint.LbEndpoint_Endpoint{
 					Endpoint: &endpoint.Endpoint{
 						Address: epAddr,
--- a/pilot/pkg/xds/ep_filters.go
+++ b/pilot/pkg/xds/ep_filters.go
@@ -66,7 +66,7 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocalityEndpoint
 		gatewayWeights := make(map[model.NetworkGateway]uint32)
 
 		// Create a map to keep track of aggregate health of the gateways.
-		gatewayHealth := make(map[model.NetworkGateway]bool)
+		gatewayHealth := make(map[model.NetworkGateway]struct{})
 
 		// Process all of the endpoints.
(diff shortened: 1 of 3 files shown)


pilot/pkg/xds/ep_filters.go
```diff
@@ -64,6 +65,9 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocalityEndpoint
 		// Create a map to keep track of the gateways used and their aggregate weights.
 		gatewayWeights := make(map[model.NetworkGateway]uint32)
 
+		// Create a map to keep track of aggregate health of the gateways.
+		gatewayHealth := make(map[model.NetworkGateway]bool)
```
ramaraochavali, line 69: nit use map[model.NetworkGateway]struct{} instead of map[model.NetworkGateway]bool

ramaraochavali (commented):
Thanks for fixing this. Minor comment. Still need to look at tests

pilot/pkg/xds/ep_filters.go
sangwa: Thanks, forgot about that option. Fixed.

pilot/pkg/xds/ep_filters.go
```diff
@@ -64,6 +65,9 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocalityEndpoint
 		// Create a map to keep track of the gateways used and their aggregate weights.
 		gatewayWeights := make(map[model.NetworkGateway]uint32)
 
+		// Create a map to keep track of aggregate health of the gateways.
+		gatewayHealth := make(map[model.NetworkGateway]struct{})
```
zirain, line 69: this seems can be improved if we enable go generic?

hzxuzhonghu:
> and in a multi-cluster, multi-network configuration for a multi-cluster service the cluster endpoints of remote network gateways are neither removed nor marked UNHEALTHY if all endpoints of that service in the respective remote network are UNHEALTHY.

What's wrong with this? It looks reasonable to remove the gateway when all the endpoints are unready.

ramaraochavali:
>  It also obviously breaks the contract of Kubernetes of not sending traffic to unready pods under those conditions.

FWIW, this is being tracked as Envoy issue https://github.com/envoyproxy/envoy/issues/22120

pilot/pkg/xds/ep_filters.go
```diff
@@ -103,6 +107,13 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocalityEndpoint
 				continue
 			}
 
+			// A gateway is considered healthy only if it has any healthy endpoints reachable.
+			if istioEndpoint.HealthStatus == model.Healthy {
+				for _, gw := range gateways {
```
hzxuzhonghu, line 112: I am concerned about the performance

pilot/pkg/xds/ep_filters.go
```diff
@@ -134,6 +155,7 @@ func (b *EndpointBuilder) EndpointsByNetworkFilter(endpoints []*LocalityEndpoint
 
 			// Generate the EDS endpoint for this gateway.
 			gwEp := &endpoint.LbEndpoint{
+				HealthStatus: healthStatus,
```
hzxuzhonghu, line 158: We should respect the env

sangwa:
> What's wrong with this? It looks reasonable to remove the gateway when all the endpoints are unready.

The gateway is not removed with SendUnhealthyEndponts enabled, which leads to the traffic being sent to a network with no healthy endpoints.

The patch fixes that by marking the gateway endpoints unhealthy as well. Another option is to remove them, but I'm not sure that won't unexpectedly affect recalculated weights if a locality distribute/failover is enabled. (Also SendUnhealthyEndpoints enabled assumes the unhealthy endpoints are sent, not removed.)

pilot/pkg/xds/ep_filters.go
sangwa: I can add a check, but it seems redundant: if SendUnhealthyEndpoints is off, unhealthy endpoints are not on the list and never considered, so the status will be always set to Healthy. Otherwise the check is always required to prevent the healthy gateway for an unhealthy endpoint set.

I can probably optimize by moving the calculation of aggregate health under an env check.

pilot/pkg/xds/ep_filters.go
sangwa: I think in most cases a remote network's gateways are a handful of endpoints, not sure if iterating over a few items here would have a significant impact.

hzxuzhonghu: Nope, remote network can be common as local network

pilot/pkg/xds/ep_filters.go
hzxuzhonghu: I mean we should only mark the HealthStatus only when `SendUnhealthyEndpoints` is on

hzxuzhonghu (changes requested):
What if some endpoints are ready while some are not in remote network?
How would you set the gateway weight? And what will happen for failover?

pilot/pkg/xds/ep_filters.go
sangwa: > Nope, remote network can be common as local network

To be honest I don't really understand what you mean by that. I would greatly appreciate if you could elaborate on it and why you think there is a performance concern.

The very next line of code, a call to `splitWeightAmongGateways()`, is doing exactly the same in that function - iterating over the list of gateways to update their aggregate weights in a map. If you're specifically concerned that I duplicate that iteration, it can be refactored and will likely be gone when I fix the patch for the rest of the comments.

pilot/pkg/xds/ep_filters.go
sangwa: Yes, I understand that. Like I said, it seems redundant, because the default (unset) status is UNKNOWN which is translated to HEALTHY by Envoy, so setting it to HEALTHY explicitly without checking for the SendUnhealthyEndpoints flag effectively changes nothing. On the other hand setting it to UNHEALTHY assumes the knowledge of unhealthy endpoints which is only available when the flag is enabled.

sangwa:
> What if some endpoints are ready while some are not in remote network?
> How would you set the gateway weight? And what will happen for failover?

The current patch indeed does not adjust weights if some (or all) endpoints in a remote network are unhealthy.

When locality failover is disabled, this is easily fixable by skipping the unhealthy endpoints when calculating the aggregate weight for the gateways.

However you're right that with locality failover this is more complicated. If we want Envoy to perform priority routing properly it should be aware of the percentage of unhealthy endpoints in a network, and generated gateway endpoints should be adjusted accordingly.

This seems to require some non-trivial computation and passing the knowledge of whether the locality failover is specifically enabled for the network/cluster topology to the endpoints filter. I'll try to see if it can be feasibly implemented.

Any fix still seems better to me than no fix at all, even if we don't adjust properly for the very particular case of locality failover on the network topology. Without any fix enabling SendUnhealthyEndpoints at the moment actually breaks multi-cluster by allowing traffic to unready remote endpoints.

pilot/pkg/xds/ep_filters.go
hzxuzhonghu: There are two loops

hzxuzhonghu:
I would recommend drafting a design to be clear of all the cases, this is far from perfect now.

PR #40575, opened 2022-08-21
Fix PushContext.sidecarIndex.rootConfig flipping issue

l8huang:
When a PushRequest doesn't have Service/VirtualService/DestinationRule/Sidecar change,
in PushContext.updateContext(), initSidecarScopes() will not be invoked, new PushContext
only copies sidecarIndex.sidecarsByNamespace from old one, sidecarIndex.rootConfig becomes
nil.

When workload namespace doesn't have Sidecar resource, this causes new ADS connection's
SidecarScope being created by DefaultSidecarScopeForNamespace(), the global default
sidecar in istio root namespace is not respected.

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -1352,7 +1352,11 @@ func (ps *PushContext) updateContext(
 			return err
 		}
 	} else {
-		ps.sidecarIndex.sidecarsByNamespace = oldPushContext.sidecarIndex.sidecarsByNamespace
+		// new ADS connection may insert new entry to computedSidecarsByNamespace/gatewayDefaultSidecarsByNamespace
+		oldPushContext.sidecarIndex.defaultSidecarMu.Lock()
+		defer oldPushContext.sidecarIndex.defaultSidecarMu.Unlock()
+
+		ps.sidecarIndex = oldPushContext.sidecarIndex
 	}
 
 	return nil
--- a/pilot/pkg/model/push_context_test.go
+++ b/pilot/pkg/model/push_context_test.go
@@ -904,6 +904,19 @@ func TestInitPushContext(t *testing.T) {
 			ExportTo: []string{".", "ns1"},
 		},
 	})
+	_, _ = configStore.Create(config.Config{
+		Meta: config.Meta{
+			Name:             "default",
+			Namespace:        "istio-system",
+			GroupVersionKind: gvk.Sidecar,
+		},
+		Spec: &networking.Sidecar{
+			Egress: []*networking.IstioEgressListener{
+				{Hosts: []string{"test1/*"}},
+			},
+		},
+	})
+
 	store := istioConfigStore{ConfigStore: configStore}
 
 	env.ConfigStore = &store


pilot/pkg/model/push_context.go
```diff
@@ -1353,6 +1353,7 @@ func (ps *PushContext) updateContext(
 		}
 	} else {
 		ps.sidecarIndex.sidecarsByNamespace = oldPushContext.sidecarIndex.sidecarsByNamespace
+		ps.sidecarIndex.rootConfig = oldPushContext.sidecarIndex.rootConfig
```
l8huang, line 1356: remove this line, test case `TestInitPushContext()` will fail

hzxuzhonghu: Why not copy whole `sidecarIndex`?

l8huang: - `defaultSidecarMu` is Mutex -- A Mutex must not be copied after first use.
- `computedSidecarsByNamespace` and `gatewayDefaultSidecarsByNamespace` are lazy-ly computed, copying them may improve performance in some cases, e.g. an incoming ADS connection already has a computed entry; but for a fresh new ADS connection which doesn't have Sidecar(namespace/workload level), its SidecarScope still need to be lazy-ly computed.

not copying `rootConfig`  could cause issues, this PR is mainly for bug fix, performance is not concerned -- performance is just like before.

howardjohn: defaultSidecarMu is a `*Mutex`, not a `Mutex`. It can be copied.

Not sure why we cannot copy the computed stuff either...? The data within them should still be valid, right?

l8huang: Per my understanding, the computed stuff are still valid, but I didn't copying them because they were not copied before, and not 100% sure copying them is side-effect free.

howardjohn: IMO we should copy, it will help performance and make this simpler

howardjohn: and if we are not confident its safe we can improve the tests :-)

l8huang: PR updated, PTAL.

pilot/pkg/model/push_context.go
```diff
@@ -1352,7 +1352,11 @@ func (ps *PushContext) updateContext(
 			return err
 		}
 	} else {
-		ps.sidecarIndex.sidecarsByNamespace = oldPushContext.sidecarIndex.sidecarsByNamespace
+		// new ADS connection may insert new entry to computedSidecarsByNamespace/gatewayDefaultSidecarsByNamespace
+		oldPushContext.sidecarIndex.defaultSidecarMu.Lock()
+		defer oldPushContext.sidecarIndex.defaultSidecarMu.Unlock()
```
howardjohn, line 1357: nit: this works now but if we refactored this it may hold the lock to long. Should we not use defer maybe?

howardjohn (approved):

PR #40581, opened 2022-08-22
export SetDestinationRules only for tests

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -1061,7 +1061,7 @@ func (ps *PushContext) destinationRule(proxyNameSpace string, service *Service) 
 		return out
 	}
 
-	// 5. service DestinationRules were merged in SetDestinationRules, return mesh/namespace rules if present
+	// 5. service DestinationRules were merged in updateDestinationRules, return mesh/namespace rules if present
 	if features.EnableDestinationRuleInheritance {
 		// return namespace rule if present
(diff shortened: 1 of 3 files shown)


pilot/pkg/model/push_context.go
```diff
@@ -1683,11 +1683,10 @@ func newConsolidatedDestRules() *consolidatedDestRules {
...
 }
 
-// SetDestinationRules is updates internal structures using a set of configs.
+// updateDestinationRules updates internal structures using a set of configs.
 // Split out of DestinationRule expensive conversions, computed once per push.
-// This also allows tests to inject a config without having the mock.
-// This will not work properly for Sidecars, which will precompute their destination rules on init
-func (ps *PushContext) SetDestinationRules(configs []config.Config) {
+// This will not work properly for Sidecars, which precomputes destination rules on init.
+func (ps *PushContext) updateDestinationRules(configs []config.Config) {
```
hzxuzhonghu, line 1689: I donot think frequently renaming is friendly to developers, a good practice is only do this once the function name could not express what it is doing,

ramaraochavali: Yeah. I agree. Just exploring this path.

PR #40594, opened 2022-08-22
fix regex expression for stat_prefix

varks:
fixes regex expression for stat_prefix added in https://github.com/istio/istio/pull/40503

### DIFF
--- a/pkg/bootstrap/config.go
+++ b/pkg/bootstrap/config.go
@@ -61,7 +61,7 @@ const (
 	requiredEnvoyStatsMatcherInclusionSuffixes = rbacEnvoyStatsMatcherInclusionSuffix + ",downstream_cx_active" // Needed for draining.
 
 	// required for metrics based on stat_prefix in virtual service.
-	requiredEnvoyStatsMatcherInclusionRegexes = "vhost\\.*\\.route\\.*"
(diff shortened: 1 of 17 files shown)


pkg/bootstrap/config.go
```diff
@@ -61,7 +61,7 @@ const (
 	requiredEnvoyStatsMatcherInclusionSuffixes = rbacEnvoyStatsMatcherInclusionSuffix + ",downstream_cx_active" // Needed for draining.
 
 	// required for metrics based on stat_prefix in virtual service.
-	requiredEnvoyStatsMatcherInclusionRegexes = "vhost.*.route.*"
+	requiredEnvoyStatsMatcherInclusionRegexes = "vhost\\.*\\.route\\.*"
```
howardjohn, line 64: nit: use backtick instead of " to avoid needing double escapes

howardjohn (approved):

pkg/bootstrap/config.go
varks: ack, fixed it.

varks:
@howardjohn when you get a chance, could you add the `ok to test` label. thank you.

zirain:
can you add a e2e tests for this?

Issue #40580, opened 2022-08-22
istiod ha high cpu


jduepmeier:
### Bug Description

After some time some of the istiod pods (not all) will consume all cpu of a node. HPA is enabled so new pods will spawn until the limit is reached but the old pods will never recover.

Deployment is via default helm chart.

### Version

```prose
$ istioctl version
client version: 1.14.3
control plane version: 1.14.3
data plane version: 1.14.3 (26 proxies)

$ kubectl version --short
Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
Client Version: v1.24.2
Kustomize Version: v4.5.4
Server Version: v1.21.6+vmware.1
WARNING: version difference between client (1.24) and server (1.21) exceeds the supported minor version skew of +/-1

$ helm version --short
v3.9.3+g414ff28

$ helm list
istio-base          	istio-system  	1       	2022-07-22 11:53:26.284845268 +0200 CEST	deployed	base-1.14.1                   	1.14.1 (no changes between 1.14.1 and 1.14.3, crds are applied from crd-all.gen.yaml and crd-operator.yaml files for the version)
istio-cni           	kube-system   	2       	2022-08-05 10:21:37.924369286 +0200 CEST	deployed	cni-1.14.3                    	1.14.3
istio-egressgateway 	istio-gateways	3       	2022-08-05 10:36:19.886377351 +0200 CEST	deployed	gateway-1.14.3                	1.14.3
istio-ingressgateway	istio-gateways	3       	2022-08-05 10:36:19.889675063 +0200 CEST	deployed	gateway-1.14.3                	1.14.3
istiod              	istio-system  	3       	2022-08-05 10:36:09.368492077 +0200 CEST	deployed	istiod-1.14.3
```

### Additional Information

I created the pprof reports (https://github.com/istio/istio/wiki/Analyzing-Istio-Performance) from one of the affected pods.

[goroutines.txt](https://github.com/istio/istio/files/9392001/goroutines.txt)
[pprof.pilot-discovery.alloc_objects.alloc_space.inuse_objects.inuse_space.001.pb.gz](https://github.com/istio/istio/files/9392002/pprof.pilot-discovery.alloc_objects.alloc_space.inuse_objects.inuse_space.001.pb.gz)
[pprof.pilot-discovery.samples.cpu.001.pb.gz](https://github.com/istio/istio/files/9392003/pprof.pilot-discovery.samples.cpu.001.pb.gz)

### Affected product area

- [ ] Docs
[...]

PR #40595, opened 2022-08-22
analysis: optimize performance and fix dead loop

howardjohn:
Two fixes:

1. We are infinite loop on stop when it triggers. `break` is for the
   switch, not the `for`.
2. No need to use regex
Fixes https://github.com/istio/istio/issues/40580

### DIFF
--- a/pkg/config/analysis/incluster/controller.go
+++ b/pkg/config/analysis/incluster/controller.go
@@ -113,7 +113,7 @@ func (c *Controller) Run(stop <-chan struct{}) {
 			log.Debugf("finished enqueueing all statuses")
 		case <-stop:
 			t.Stop()
-			break
+			return
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/config/analysis/incluster/controller.go
```diff
@@ -113,7 +113,7 @@ func (c *Controller) Run(stop <-chan struct{}) {
 			log.Debugf("finished enqueueing all statuses")
 		case <-stop:
 			t.Stop()
-			break
+			return
```
hzxuzhonghu, line 116: This is a bug, but how is it related to CPU high?

hzxuzhonghu (approved):

pkg/config/analysis/incluster/controller.go
howardjohn: it's an infinite looping, end up with 100% cpu on selectgo. Once stop is closed

hzxuzhonghu: I mean stop can be closed only when the process exit?

howardjohn:
I think it's from losing leader election which is why the original issue
was only with multiple istiod replicas

PR #40599, opened 2022-08-22
added new TCP based test for deny policies

aryan16:
**Behavior of deny policies on TCP Ports in the following scenarios**
1) When we have any L7 rule (only http attributes)
2) When we don't have a pure L7 rule i.e. mixed rule (L7 + L4)

### DIFF
--- a/tests/integration/security/authz_test.go
+++ b/tests/integration/security/authz_test.go
@@ -1649,6 +1649,105 @@ func TestAuthz_CustomServer(t *testing.T) {
 		})
 }
 
+func TestAuthz_TCP_Pure_L7_Rule_Policy(t *testing.T) {
+	framework.NewTest(t).
+		Features("security.authorization.mtls-local",
+			"security.authorization.grpc-protocol",
+			"security.authorization.tcp",
+			"security.authorization.negative-match").
+		Run(func(t framework.TestContext) {
+			allowed := apps.Ns1.A
+			denied := apps.Ns2.A
+
+			from := allowed.Append(denied)
+			fromMatch := match.AnyServiceName(from.NamespacedNames())
+			toMatch := match.Not(fromMatch)
+			to := toMatch.GetServiceMatches(apps.Ns1AndNs2)
+			fromAndTo := to.Instances().Append(from)
+
+			config.New(t).
+				Source(config.File("testdata/authz/mtls.yaml.tmpl")).
+				Source(config.File("testdata/authz/tcp-L7-rule.yaml.tmpl")).
+				BuildAll(nil, to).
+				Apply()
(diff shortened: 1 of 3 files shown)


lei-tang (approved):

tests/integration/security/authz_test.go
```diff
@@ -1649,6 +1649,105 @@ func TestAuthz_CustomServer(t *testing.T) {
 		})
 }
 
+func TestAuthz_TCP_L7(t *testing.T) {
```
lei-tang, line 1652: Nit: consider naming the test TestAuthz_TCP_L7_policy_only.

tests/integration/security/authz_test.go
```diff
@@ -1649,6 +1649,105 @@ func TestAuthz_CustomServer(t *testing.T) {
...
+							To(to).
+							Allow(c.allow).
+							Path(c.path).
+							BuildAndRunForPorts(t, c.ports...)
+					}
+				})
+		})
+}
+
+func TestAuthz_TCP_Mixed(t *testing.T) {
```
lei-tang, line 1702: Nit: consider naming the test TestAuthz_TCP_L4_L7_policy_mixed.

tests/integration/security/testdata/authz/tcp-L7.yaml.tmpl
```diff
@@ -0,0 +1,17 @@
...
+      "app": "{{ .To.ServiceName }}"
+  action: DENY
+  rules:
+    - to:
+        - operation: # HTTP
+            methods: [ "GET" ]
+    - to:
+        - operation: # HTTP and TCP
+            methods: [ "GET" ]
+            ports: [ "{{ (.To.PortForName `tcp`).WorkloadPort }}" ]
```
lei-tang, line 17: This policy contains "ports". Is it a L7 only policy?

aryan16: So we check the policy rule by rule. In case of deny, any one rule matches then the request will be denied. In this case one rule has pure L7 attributes and other one has mixed (L7 and L4). So any request to any port will always be matched with rule1 (pure L7), hence it denies the request. Let's call this kind of policy as L7 policy.

lei-tang: So this test case TestAuthz_TCP_L7 is similar to the mixed test case TestAuthz_TCP_Mixed as both of them have L4 and L7 attributes in the authz policy. Maybe the test names can be TestAuthz_TCP_Mixed_1 TestAuthz_TCP_Mixed_2.

PR #40601, opened 2022-08-22
go get cloud.google.com/go/compute

therealmitchconnors:
needed for bugfix in v1.9.0.  master updated in #40597


lei-tang (approved):

ericvn:
Not sure if this needs the updated 1.13 build tools image, but I notice the RMs haven't allowed the new image to merge. I do think it related to a new gcloud version.

Random-Liu:
In k8s, it just check the local file /sys/class/dmi/id/product_name to decide whether it is on GCE or not
https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/gcp/metadata.go#L137

Could istio reuses the same logic?

PR #40602, opened 2022-08-22
istioctl precheck: disable slow check when not needed

howardjohn:
Rationale is in code comment


therealmitchconnors (approved):
This might merit a minor release note...

PR #40606, opened 2022-08-23
tests: consolidate opentelemetry-collector

### LINKED ISSUE #40495: Consolidate opentelemetry-collector
```
$ rg opentelemetry-collector:
samples/open-telemetry/otel.yaml
82:          image: otel/opentelemetry-collector:0.54.0

pkg/test/framework/components/opentelemetry/opentelemetry-collector.yaml
92:          image: "otel/opentelemetry-collector:0.54.0"
```

Currently we have test deployment and sample. We should probably have the tests use the sample/ deployment. Then we have less duplication and test what our users run

cc @zirain

zirain:

fix: https://github.com/istio/istio/issues/40495

cc @howardjohn


Xunzhuo (approved):

samples/open-telemetry/otel.yaml
```diff
@@ -1,58 +1,71 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: otel-collector-conf
-  namespace: istio-system
+  name: opentelemetry-collector-conf
```
howardjohn, line 4: should we keep the namespace? all the other samples have it baked in?

zirain: that would forbidden user install the other namespace?
```console
kubectl apply -f samples/open-telemetry/otel.yaml  --namespace=default
the namespace from the provided object "istio-system" does not match the namespace "default". You must pass '--namespace=istio-system' to perform this operation.
```

howardjohn: It is just a sample though, they can change it. Also the meshconfig in docs, etc will use the hardcoded namespace

zirain: upadted

zirain: ping @howardjohn

howardjohn (approved):

PR #40607, opened 2022-08-23
Fix bootstrap discovery service 

hzxuzhonghu:

The logic of bootstrap config discovery can lead to goroutine leak if it times out.

And also separate the bootstrap discovery stream.

### DIFF
--- a/pkg/istio-agent/agent.go
+++ b/pkg/istio-agent/agent.go
@@ -882,7 +882,7 @@ func (a *Agent) newSecretManager() (*cache.SecretManagerClient, error) {
 
 		if tlsOpts.RootCert == "" {
 			log.Infof("Using CA %s cert with system certs", a.secOpts.CAEndpoint)
-		} else if _, err := os.Stat(tlsOpts.RootCert); os.IsNotExist(err) {
+		} else if !fileExists(tlsOpts.RootCert) {
 			log.Fatalf("invalid config - %s missing a root certificate %s", a.secOpts.CAEndpoint, tlsOpts.RootCert)
 		} else {
 			log.Infof("Using CA %s cert with certs: %s", a.secOpts.CAEndpoint, tlsOpts.RootCert)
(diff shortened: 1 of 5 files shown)


pkg/istio-agent/agent.go
```diff
@@ -308,100 +302,41 @@ func (a *Agent) initializeEnvoyAgent(ctx context.Context) error {
...
-	sent        bool
-	received    bool
-}
-
-// Send refers to a request from the xDS proxy.
-func (b *bootstrapDiscoveryRequest) Send(resp *discovery.DiscoveryResponse) error {
-	if resp.TypeUrl == v3.BootstrapType && !b.received {
-		b.received = true
-		if len(resp.Resources) != 1 {
-			b.envoyWaitCh <- fmt.Errorf("unexpected number of bootstraps: %d", len(resp.Resources))
```
hzxuzhonghu, line 368: This can be blocking forever

ramaraochavali (commented):
Do we have tests to validate this flow?

zirain:
> Do we have tests to validate this flow?

we have tests for `BOOTSTRAP` in `tests/integration/telemetry/stats/prometheus/customizemetrics/customize_metrics_test.go`

ramaraochavali (approved):

pkg/istio-agent/agent.go
```diff
@@ -308,100 +302,40 @@ func (a *Agent) initializeEnvoyAgent(ctx context.Context) error {
...
-					envoyUpdate: envoyProxy.UpdateConfig,
-				}
-				_ = a.xdsProxy.handleStream(request)
-				select {
-				case <-a.envoyWaitCh:
-					break retries
-				default:
+		// wait indefinitely and keep retrying with jittered exponential backoff
+		b := backoff.NewExponentialBackOff()
+		b.MaxElapsedTime = 30 * time.Second
```
howardjohn, line 310: I think with non-zero MaxElapsedTime, after we hit max time the library starts returning -1 so we will have infinite tight loop.

IMO this library is too easy to use poorly, maybe we should create a wrapper in Istio pkg

hzxuzhonghu: Let me take a close look at it. If so, it is necessary to wrap it

PR #40612, opened 2022-08-23
use registerIf to disable metrics in xds cache

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/xds_cache.go
+++ b/pilot/pkg/model/xds_cache.go
@@ -30,11 +30,15 @@ import (
 	"istio.io/pkg/monitoring"
 )
 
+var enableStats = func() bool {
+	return features.EnableXDSCacheMetrics
+}
+
 func init() {
-	monitoring.MustRegister(xdsCacheReads)
-	monitoring.MustRegister(xdsCacheEvictions)
-	monitoring.MustRegister(xdsCacheSize)
-	monitoring.MustRegister(dependentConfigSize)
+	xdsCacheReads = monitoring.RegisterIf(xdsCacheReads, enableStats)
+	xdsCacheEvictions = monitoring.RegisterIf(xdsCacheEvictions, enableStats)
+	xdsCacheSize = monitoring.RegisterIf(xdsCacheSize, enableStats)
+	dependentConfigSize = monitoring.RegisterIf(dependentConfigSize, enableStats)
 }
 
 var (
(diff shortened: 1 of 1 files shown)


pilot/pkg/model/xds_cache.go
```diff
@@ -30,11 +30,15 @@ import (
...
+var enableStats = func() bool {
+	return features.EnableXDSCacheMetrics
+}
+
 func init() {
-	monitoring.MustRegister(xdsCacheReads)
-	monitoring.MustRegister(xdsCacheEvictions)
-	monitoring.MustRegister(xdsCacheSize)
-	monitoring.MustRegister(dependentConfigSize)
+	xdsCacheReads = monitoring.RegisterIf(xdsCacheReads, enableStats)
```
howardjohn, line 38: should we just wrap L45 instead of using init?

pilot/pkg/model/xds_cache.go
```diff
@@ -30,35 +30,28 @@ import (
 	"istio.io/pkg/monitoring"
 )
 
-func init() {
-	monitoring.MustRegister(xdsCacheReads)
-	monitoring.MustRegister(xdsCacheEvictions)
-	monitoring.MustRegister(xdsCacheSize)
-	monitoring.MustRegister(dependentConfigSize)
+var enableStats = func() bool {
+	return features.EnableXDSCacheMetrics
```
hzxuzhonghu, line 34: no need a func wrapper

ramaraochavali: Sorry. I did not understand. Can you please eloborate?

hzxuzhonghu: I mean

var enableStats = features.EnableXDSCacheMetrics

ramaraochavali: But RegisterIf takes a func that returns bool?

hzxuzhonghu: OK, but i doubt it is not necessary in this case.

ramaraochavali: Are you suggesting to rework RegisterIf function?

hzxuzhonghu: I am not sure, just feel it makes things a little complex with cb

hzxuzhonghu (approved):

PR #40623, opened 2022-08-24
`istioctl x proxy-status` via istio agents.

ingwonsong:
This PR proposes an experimental option `--xds-via-agents` only for `istioctl x proxy-status` command, which sends a query for `istioctl x proxy-status` via istio agents, instead of direct query to `istiod`.

`istioctl x proxy-status` try to find all the istiod instances by searching pods in `istio-system` namespace.
But, if we use the external istiods, this way may not work.

For example, `istiod` can be deployed in the other cluster or out of K8S. Moreover, in front of the istiod instances, some sort of load balancer can be set to distribute the connections to multiple istiods. This kind of configuration may happen when the customer want to isolate their control plane from the workloads, or when they are using some managed control plane.

In such case, `istioctl x proxy-status` cannot enumerate and access to each istiod instance directly. So, this PR proposes to access `istiod` via "istio agent".

Please note that, by https://github.com/istio/istio/pull/40566, we can make a query to istiod via istio agents efficiently.

### DIFF
--- a/istioctl/cmd/proxystatus.go
+++ b/istioctl/cmd/proxystatus.go
@@ -227,7 +227,6 @@ Retrieves last sent and last acknowledged xDS sync from Istiod to each Envoy in 
 				}
 				return c.Diff()
 			}
-
 			xdsRequest := xdsapi.DiscoveryRequest{
 				TypeUrl: pilotxds.TypeDebugSyncronization,
 			}
@@ -244,6 +243,8 @@ Retrieves last sent and last acknowledged xDS sync from Istiod to each Envoy in 
 	centralOpts.AttachControlPlaneFlags(statusCmd)
 	statusCmd.PersistentFlags().StringVarP(&configDumpFile, "file", "f", "",
 		"Envoy config dump JSON file")
+	statusCmd.PersistentFlags().BoolVar(&centralOpts.XdsViaAgents, "xds-via-agents", false,
+		"Access Istiod via the tap service of each agent.")
 
 	return statusCmd
 }
--- a/istioctl/cmd/proxystatus.go
+++ b/istioctl/cmd/proxystatus.go
@@ -244,7 +244,10 @@ Retrieves last sent and last acknowledged xDS sync from Istiod to each Envoy in 
 	statusCmd.PersistentFlags().StringVarP(&configDumpFile, "file", "f", "",
 		"Envoy config dump JSON file")
 	statusCmd.PersistentFlags().BoolVar(&centralOpts.XdsViaAgents, "xds-via-agents", false,
-		"Access Istiod via the tap service of each agent.")
+		"Access Istiod via the tap service of each agent")
+	statusCmd.PersistentFlags().IntVar(&centralOpts.XdsViaAgentsLimit, "xds-via-agents-limit", 100,
+		"Maximum number of pods being visited by istioctl when `xds-via-agent` flag is true."+
+			"To iterate all the agent pods without limit, set to 0")
 
 	return statusCmd
 }
--- a/istioctl/cmd/proxystatus.go
+++ b/istioctl/cmd/proxystatus.go
@@ -157,6 +157,7 @@ func newKubeClient(kubeconfig, configContext string) (kube.CLIClient, error) {
 func xdsStatusCommand() *cobra.Command {
 	var opts clioptions.ControlPlaneOptions
 	var centralOpts clioptions.CentralControlPlaneOptions
+	var multiXdsOpts multixds.Options
 
 	statusCmd := &cobra.Command{
 		Use:   "proxy-status [<type>/]<name>[.<namespace>]",
@@ -194,6 +195,7 @@ Retrieves last sent and last acknowledged xDS sync from Istiod to each Envoy in 
 			if err != nil {
 				return err
 			}
+			multiXdsOpts.MessageWriter = c.OutOrStdout()
 
 			if len(args) > 0 {
 				podName, ns, err := handlers.InferPodInfoFromTypedResource(args[0],
@@ -217,7 +219,7 @@ Retrieves last sent and last acknowledged xDS sync from Istiod to each Envoy in 
 					ResourceNames: []string{fmt.Sprintf("%s.%s", podName, ns)},
 					TypeUrl:       pilotxds.TypeDebugConfigDump,
 				}
-				xdsResponses, err := multixds.FirstRequestAndProcessXds(&xdsRequest, centralOpts, istioNamespace, "", "", kubeClient)
+				xdsResponses, err := multixds.FirstRequestAndProcessXds(&xdsRequest, centralOpts, istioNamespace, "", "", kubeClient, multiXdsOpts)
 				if err != nil {
 					return err
 				}
@@ -230,7 +232,7 @@ Retrieves last sent and last acknowledged xDS sync from Istiod to each Envoy in 
 			xdsRequest := xdsapi.DiscoveryRequest{
 				TypeUrl: pilotxds.TypeDebugSyncronization,
 			}
-			xdsResponses, err := multixds.AllRequestAndProcessXds(&xdsRequest, centralOpts, istioNamespace, "", "", kubeClient)
+			xdsResponses, err := multixds.AllRequestAndProcessXds(&xdsRequest, centralOpts, istioNamespace, "", "", kubeClient, multiXdsOpts)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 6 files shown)


istioctl/pkg/multixds/gather.go
```diff
@@ -210,8 +310,15 @@ func MultiRequestAndProcessXds(all bool, dr *xdsapi.DiscoveryRequest, centralOpt
 		}, nil
 	}
 
-	// Self-administered case.  Find all Istiods in revision using K8s, port-forward and call each in turn
-	responses, err := queryEachShard(all, dr, istioNamespace, kubeClient, centralOpts)
+	var responses []*xdsapi.DiscoveryResponse
```
zirain, line 313:
```suggestion
	var (
	responses []*xdsapi.DiscoveryResponse
	err error
	)
```

istioctl/cmd/proxystatus.go
```diff
@@ -244,6 +243,8 @@ Retrieves last sent and last acknowledged xDS sync from Istiod to each Envoy in
 	centralOpts.AttachControlPlaneFlags(statusCmd)
 	statusCmd.PersistentFlags().StringVarP(&configDumpFile, "file", "f", "",
 		"Envoy config dump JSON file")
+	statusCmd.PersistentFlags().BoolVar(&centralOpts.XdsViaAgents, "xds-via-agents", false,
```
zirain, line 246: it's safe to set default value to `true` on an experimental cmd, which will give us more feed back?

ingwonsong: This is an option for the special case: istiods are not in the local k8s cluster, so cannot be reached directly. If not in such case, the original approach is efficient.

If this PR is approved and submitted, after some period of tests, I would like to make a subsequent PR to apply this way when there is no "istiod" in the k8s cluster, which means "external istiod".

pkg/kube/client.go
```diff
@@ -884,6 +888,39 @@ func (c *client) GetIstioVersions(ctx context.Context, namespace string) (*versi
...
+		req.Param(k, v)
+	}
+
+	res := req.Do(ctx)
+	if res.Error() != nil {
+		return nil, fmt.Errorf("unable to retrieve Pods: %v", res.Error())
+	}
+
+	list := &v1.PodList{}
+	if err := res.Into(list); err != nil {
```
hzxuzhonghu, line 917: why not use k8s client-go warpped api? kubernetes.Interface

ingwonsong: Done!

pkg/kube/client.go
```diff
@@ -884,6 +887,25 @@ func (c *client) GetIstioVersions(ctx context.Context, namespace string) (*versi
...
+		LabelSelector: label.ServiceCanonicalName.Name,
+		FieldSelector: "status.phase=Running",
+		Limit:         limit,
+		Continue:      token,
+	}
+	if c.revision != "" {
+		opts.LabelSelector = fmt.Sprintf("%s,%s=%s", label.ServiceCanonicalName.Name, label.IoIstioRev.Name, c.revision)
+	}
+	// get pods from all the namespaces.
+	list, err := c.kube.CoreV1().Pods("").List(ctx, opts)
```
howardjohn, line 901:
```suggestion
	list, err := c.kube.CoreV1().Pods(metav1.NamespaceAll).List(ctx, opts)
```

nit

istioctl/cmd/proxystatus.go
zirain: make sense

zirain (approved):
lgtm

pkg/kube/client.go
```diff
@@ -884,6 +887,25 @@ func (c *client) GetIstioVersions(ctx context.Context, namespace string) (*versi
...
 
+func (c *client) GetProxyPods(ctx context.Context, limit int64, token string) (*v1.PodList, error) {
+	opts := metav1.ListOptions{
+		LabelSelector: label.ServiceCanonicalName.Name,
+		FieldSelector: "status.phase=Running",
+		Limit:         limit,
+		Continue:      token,
+	}
+	if c.revision != "" {
+		opts.LabelSelector = fmt.Sprintf("%s,%s=%s", label.ServiceCanonicalName.Name, label.IoIstioRev.Name, c.revision)
```
hzxuzhonghu, line 898: %s,%s=%s is this right?

ingwonsong: Yes. 

Please look at https://github.com/istio/istio/blob/abfd63329f7fbdaaea2b0a9e5d66ec86bb87947f/pkg/kube/client.go#L819.
It is for finding the Istiod pod, but the label is the same.

pkg/kube/client.go
```diff
@@ -884,6 +887,25 @@ func (c *client) GetIstioVersions(ctx context.Context, namespace string) (*versi
 	return &res, errs
 }
 
+func (c *client) GetProxyPods(ctx context.Context, limit int64, token string) (*v1.PodList, error) {
+	opts := metav1.ListOptions{
+		LabelSelector: label.ServiceCanonicalName.Name,
+		FieldSelector: "status.phase=Running",
```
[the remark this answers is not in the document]
ingwonsong: Yes. But, there are pretty many codes which use this way. 

```
❯ ack "status.phase="
istioctl/pkg/multixds/gather.go
74:             "fieldSelector": "status.phase=Running",

istioctl/cmd/precheck.go
286:            FieldSelector: "status.phase=Running",

istioctl/cmd/describe.go
888:            FieldSelector: "status.phase=Running",

pkg/kube/client.go
751:            "fieldSelector": "status.phase=Running",
846:            "fieldSelector": "status.phase=Running",
```

Do we need to change all the things at this PR?

istioctl/pkg/multixds/gather.go
```diff
@@ -113,6 +118,104 @@ func queryEachShard(all bool, dr *xdsapi.DiscoveryRequest, istioNamespace string
...
+				resp, err := queryToOnePod(&pod)
+				if err != nil {
+					return nil, err
+				}
+				responses = append(responses, resp)
+			}
+			token = list.ListMeta.GetContinue()
+			if token == "" {
+				break
+			}
```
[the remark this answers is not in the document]
ingwonsong: In my experiments, which were conducted in GKE, there was 1 second per 1000 workload(sidecar proxies) of overhead when comparing with normal `istioctl x proxy-status`. 
So, when I did the experiment with 10,000 proxies and 20 istiod, it tooks around 23 seconds with the flag `--xds-via-agent` while 13 seconds was taken without the flag.

pkg/kube/client.go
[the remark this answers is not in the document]
ingwonsong: I changed to use "constant" for the hard coded field selector for `gather.go` and `client.go` which are changed by this PR. For the other files, will make another PR. Thanks!

pkg/kube/client.go
howardjohn: Hopefully that isn't used anywhere important since its not correct. A sidecar pod doesn't necessarily have the istio.io/rev label

ingwonsong: @howardjohn I have been confused. Thanks for pointing out.

I changed the logic to use the annotated injection status to identify "revision".
(It uses pretty dumb loop, though)

If we have a label for distinguishing the revision at the injected pod, it seems to be handy.

hzxuzhonghu (approved):

ingwonsong:
@therealmitchconnors Wam reminder

ingwonsong:
To wait for https://github.com/istio/istio/pull/40721, I set this to `do-not-merge`.

ingwonsong:
As https://github.com/istio/istio/pull/40721 supports the 15004 port as well, don't need to wait for the PR.

istioctl/pkg/multixds/gather.go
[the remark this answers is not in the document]
therealmitchconnors: How does this code handle multiple instances of istiod?  The sidecar agent of a given pod will only be able to contact the istiod instance it is connected to, right?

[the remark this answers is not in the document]
ingwonsong: Yes. So, this logic iterates all the proxies (sidecar agents + gateways).
But, if we got information about a proxy already from the previous iteration, the proxy is skipped.
So, we can effectively visited all the istiods. 
In addition, each istiod is only visited just once if there are no changes, like reconnections, during iterating the proxies.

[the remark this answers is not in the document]
therealmitchconnors: In that case, we should add flags that can limit the number of proxies we scan for large clusters.  Does the namespace flag work with this command?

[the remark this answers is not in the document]
ingwonsong: `namespace` flag does not work with `proxy-status`.
For `limit` flag, it sounds good to me.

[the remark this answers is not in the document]
ingwonsong: @therealmitchconnors  I just have added `--xds-via-agents-limit` flag.

istioctl/pkg/multixds/gather.go
```diff
@@ -113,6 +120,110 @@ func queryEachShard(all bool, dr *xdsapi.DiscoveryRequest, istioNamespace string
...
+		for {
+			list, err := kubeClient.GetProxyPods(context.TODO(), int64(kubeClientGetPodLimit), token)
+			if err != nil {
+				return nil, err
+			}
+			// Iterate all the pod.
+			for _, pod := range list.Items {
+				touchedPods++
+				if centralOpts.XdsViaAgentsLimit != 0 && touchedPods > centralOpts.XdsViaAgentsLimit {
+					break GetProxyLoop
```
therealmitchconnors, line 184: In this case we should have output alerting the user that some data may be missing.  It's good to have an upper limit here, we just need to make sure the user knows they've hit the limit, and might not have complete data.

ingwonsong: Totally agree with it. What could be proper way to display the message? stderr?

ingwonsong: Hm, I refactored it a little bit, and displayed the message. Please take a look.

therealmitchconnors (approved):

Issue #40626, opened 2022-08-24
revist all ExponentialBackOff.NextBackOff usage


hzxuzhonghu:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**

As @howardjohn pointed, and I confirmed. By default ExponentialBackOff.NextBackOff  will return `-1`,  then we use it as delay for time.Sleep(), the function will return immediately, this is very bad the process can fall in dead loop.

**Describe alternatives you've considered**

Istio may need to wrap ExponentialBackOff to prevent such behavior. 

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

hzxuzhonghu:
this is done

PR #40629, opened 2022-08-24
Backoff wrapper

hzxuzhonghu:

For https://github.com/istio/istio/issues/40626

### DIFF
--- a/pkg/backoff/exponential.go
+++ b/pkg/backoff/exponential.go
@@ -45,7 +45,6 @@ func NewExponentialBackOff(initFuncs ...func(off *ExponentialBackOff)) BackOff {
 		fn(b)
 	}
 	b.Reset()
-	b.NextBackOff()
 	return b
 }
 
--- a/pkg/backoff/exponential.go
+++ b/pkg/backoff/exponential.go
@@ -1,0 +1,60 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


hzxuzhonghu:
cc @howardjohn @ramaraochavali

hzxuzhonghu:
I will replace all github.com/cenkalti/backoff/v4 usage if it makes sense

ramaraochavali:
LGTM

pkg/backoff/exponential.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// ExponentialBackOff is a wrapper of backoff.ExponentialBackOff to override its NextBackOff().
+type ExponentialBackOff struct {
+	*backoff.ExponentialBackOff
+}
+
+// NewExponentialBackOff creates an istio wrapped ExponentialBackOff.
+func NewExponentialBackOff(initFuncs ...func(off *ExponentialBackOff)) BackOff {
+	b := &ExponentialBackOff{}
```
howardjohn, line 42: Do we need a pointer? The inner struct is already a pointer

howardjohn (approved):

pkg/backoff/exponential.go
hzxuzhonghu: Not a must

howardjohn: IMO wrapper should not be a pointer, can we swap it?

hzxuzhonghu: Done in the following

Issue #35263, opened 2021-09-17
Provide better back pressure for XDS


howardjohn:
![2021-17-09_15-41-01](https://user-images.githubusercontent.com/623453/133860847-4b6522e7-45de-44e0-b0b1-e2f88d419bd0.png)

Above shows connecting 5k proxies and some configs.

What happens is we all start generating CDS for all of them. This means we are holding 5k XDS requests (the big one! because the first has the node and is giant as a result) and in-flight CDS configs.

Because we are trying to generate 5k CDS at once, our CPU grinds to a halt due to throttling. Some responses take *3 minutes* to complete.

Clearly we are better off not allowing so many concurrently...

howardjohn:
Same test with `set env deployment/istiod PILOT_PUSH_THROTTLE=25 PILOT_MAX_REQUESTS_PER_SECOND=25`
![2021-17-09_15-57-04](https://user-images.githubusercontent.com/623453/133862258-edf7f2fb-bbb1-4bd0-aa34-3895c0699abe.png)

And with only `PILOT_MAX_REQUESTS_PER_SECOND=25` (throttle at default of 100):
![2021-17-09_16-04-04](https://user-images.githubusercontent.com/623453/133862251-01ba9b69-989b-49f5-adf2-a61e149dda95.png)

IMO we should drop the default PILOT_MAX_REQUESTS_PER_SECOND. Currently its 100.

hzxuzhonghu:
>IMO we should drop the default PILOT_MAX_REQUESTS_PER_SECOND. Currently its 100.

You mean `PILOT_PUSH_THROTTLE`?

howardjohn:
No there are 2 vars. See https://github.com/istio/istio/issues/35263#issuecomment-922125529. The rate limit alone is enough to prevent this in my tests

hzxuzhonghu:
I remember `PILOT_MAX_REQUESTS_PER_SECOND` was recently added because when a huge number of connections will break down pilot

hzxuzhonghu:
So this is a new finding?

howardjohn:
Yes we recently add the env var but set it to 100rps. IMO this is too high
and we should drop it to 25, since from the tests above 100 wasn't
sufficient to prevent overload.

In our product we are already using 25, it's working great for us.

hzxuzhonghu:
This is interesting, 25rps works best.

ramaraochavali:
+1 to changing the default to 25 rps

[...]

PR #40631, opened 2022-08-24
improve xds proxy logs

ramaraochavali:
As discovered https://github.com/istio/istio/issues/35263

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/istio-agent/xds_proxy.go
+++ b/pkg/istio-agent/xds_proxy.go
@@ -495,7 +495,7 @@ func (p *XdsProxy) handleUpstreamRequest(con *ProxyConnection) {
 				p.ecdsLastNonce.Store(req.ResponseNonce)
 			}
 			if err := sendUpstream(con.upstream, req); err != nil {
-				proxyLog.Errorf("upstream [%d] send error for type url %s: %v", con.conID, req.TypeUrl, err)
+				proxyLog.Debugf("upstream [%d] send error for type url %s: %v", con.conID, req.TypeUrl, err)
 				con.upstreamError <- err
 				return
 			}
--- a/pkg/istio-agent/xds_proxy.go
+++ b/pkg/istio-agent/xds_proxy.go
@@ -41,6 +41,7 @@ import (
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/reflection"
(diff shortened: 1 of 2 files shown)


pkg/istio-agent/xds_proxy.go
```diff
@@ -495,7 +495,7 @@ func (p *XdsProxy) handleUpstreamRequest(con *ProxyConnection) {
 				p.ecdsLastNonce.Store(req.ResponseNonce)
 			}
 			if err := sendUpstream(con.upstream, req); err != nil {
-				proxyLog.Errorf("upstream [%d] send error for type url %s: %v", con.conID, req.TypeUrl, err)
+				proxyLog.Debugf("upstream [%d] send error for type url %s: %v", con.conID, req.TypeUrl, err)
```
ramaraochavali, line 498: This log is redundant as we already handle upstreamError channel other place

pkg/istio-agent/xds_proxy_delta.go
```diff
@@ -162,7 +162,7 @@ func (p *XdsProxy) handleDeltaUpstream(ctx context.Context, con *ProxyConnection
 		select {
 		case err := <-con.upstreamError:
 			// error from upstream Istiod.
-			if istiogrpc.IsExpectedGRPCError(err) {
+			if istiogrpc.IsExpectedGRPCError(err) || isRateLimited(err) {
```
ramaraochavali, line 165: considers rate limiting as expected error

howardjohn (approved):

pkg/istio-agent/xds_proxy_delta.go
hzxuzhonghu: I am not sure if this is right. In this case, we should let users know what happen, instead of hide it

ramaraochavali: But Envoy reconnects like any other gRPC problem like Istiod disconnects every 30 mins. I do not know how is different from that. Istiod disconnects every 30 mins to shed the load.

hzxuzhonghu: See https://github.com/istio/istio/pull/40644#issuecomment-1226887934

PR #40634, opened 2022-08-24
Add extra addresses for listeners configuration to support dual stack in Istio

zhlsunshine:
Add extra addresses for listeners configuration to support dual stack in Istio, this PR is for [issue#40394](https://github.com/istio/istio/issues/40394)

This PR just focus on normal k8s service in service mesh, and add the additional addresses for specified listener of service. There fore, it does not include below:
1.  No code change for service entry 
2.  No code change for statefulsets/headless services with TCP ports, and empty service address field (There is another PR for this: https://github.com/istio/istio/pull/40409)
3.  No code change for gateway listener configuration

Below snapshots are test results and the attach contains the whole configdump file exclude eds via `curl localhost:15000/config_dump`

**Inbound listener additional addresses:**
inbound

**Outbound listener additional addresses:**
outbound

**Application/Service listener additional addresses:**
application_service

Config dump:
[extra_addresses_configdump.log](https://github.com/istio/istio/files/9416703/extra_addresses_configdump.log)

Note: I did not add any test case and I will add necessary test case after all the business code changes are available for reviewers.


jacob-delgado:
cc @adams-shaun can you help us review this?

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -292,7 +292,12 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
 ) []*listener.Listener {
 	noneMode := node.GetInterceptionMode() == model.InterceptionNone
 
-	actualWildcard, actualLocalHostAddress := getActualWildcardAndLocalHost(node)
+	var actualWildcard, actualExtraWildcard, actualLocalHostAddress, actualExtraLocalHostAddress string
```
howardjohn, line 295: this declaration isn't needed

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -121,7 +121,11 @@ func (lb *ListenerBuilder) buildVirtualOutboundListener() *ListenerBuilder {
 
 	filterChains := buildOutboundCatchAllNetworkFilterChains(lb.node, lb.push)
 
-	actualWildcard, _ := getActualWildcardAndLocalHost(lb.node)
+	var actualWildcard, extrActualWildcard string
```
howardjohn, line 124: this seems to complex and error prone and duplicated. Can we refactor this a bit to make it easier to use?

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
```diff
@@ -192,14 +195,26 @@ func (lb *ListenerBuilder) buildInboundListeners() []*listener.Listener {
...
+		actualWildcard, _ = getActualWildcardAndLocalHost(lb.node)
+	}
 
 	// Build the "virtual" inbound listener. This will capture all inbound redirected traffic and contains:
 	// * Passthrough filter chains, matching all unmatched traffic. There are a few of these to handle all cases
 	// * Service filter chains. These will either be for each Port exposed by a Service OR Sidecar.Ingress configuration.
 	allChains := buildInboundPassthroughChains(lb)
 	allChains = append(allChains, chains...)
-	return lb.buildInboundListener(model.VirtualInboundListenerName, util.BuildAddress(actualWildcard, ProxyInboundListenPort), false, allChains)
+	l := lb.buildInboundListener(model.VirtualInboundListenerName, util.BuildAddress(actualWildcard, ProxyInboundListenPort), false, allChains)
```
howardjohn, line 209: should we do this in buildInboundListener?

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -693,17 +743,24 @@ func buildSidecarOutboundTCPListenerOptsForPortOrUDS(listenerMapKey *string,
...
 	if len(listenerOpts.bind) == 0 {
 		svcListenAddress := listenerOpts.service.GetAddressForProxy(listenerOpts.proxy)
+		svcExtraListenAddress := listenerOpts.service.GetExtraAddressesForProxy(listenerOpts.proxy)
 		// We should never get an empty address.
 		// This is a safety guard, in case some platform adapter isn't doing things
 		// properly
 		if len(svcListenAddress) > 0 {
 			if !strings.Contains(svcListenAddress, "/") {
 				listenerOpts.bind = svcListenAddress
+				if len(svcExtraListenAddress) > 0 {
```
howardjohn, line 753: you can just set it always, setting to empty is NOP, right?

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
```diff
@@ -321,6 +348,15 @@ func (lb *ListenerBuilder) buildInboundChainConfigs() []inboundChainConfig {
 			if cc.bind == "" {
 				// If user didn't provide, pick one based on IP
 				cc.bind = getSidecarInboundBindIP(lb.node)
+				// if no global unicast address
+				if cc.bind == WildcardAddress || cc.bind == WildcardIPv6Address {
+					bindAddress, exBindAddresses := getDualStackActualWildcard(lb.node)
+					// if it's dual stack environment
+					if bindAddress != "" && exBindAddresses != "" {
+						cc.bind = bindAddress
+						cc.exBind = []string{exBindAddresses}
```
howardjohn, line 357: IIRC before in dual stack we made 2 listeners. Now are we creating 2 listeners with 2 addresses each? Or did e remove the 2 listeners somewhere I missed?

howardjohn (commented):
Conceptually looks good. I think the way the code is structured right now makes it very error prone for use, I think we need to work on cleaning it up a bit.

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: yes, should remove

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: yes, set `exBind` directly

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
zhlsunshine: Hi @howardjohn, I think it's better to separate the default service and extra service address, because 1. no change for `buildInboundListener` ； 2. no impact for single stack logic; 3.  Extra service address should be added separately, because it's not just for dual stack, it may be used for other in future.   Any idea?

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
```diff
@@ -60,6 +60,9 @@ type inboundChainConfig struct {
 	// and telemetry.
 	bind string
 
+	// exBind is string slice and each element is similar with bind address and support multiple addresses for 'virtual' listener
+	exBind []string
```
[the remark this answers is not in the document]
zhlsunshine: okay, I will change all `exBind` into `extraBind `

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -1229,7 +1287,15 @@ func buildListener(opts buildListenerOpts, trafficDirection core.TrafficDirectio
...
 			ConnectionBalanceConfig: connectionBalance,
 		}
-
+		// add extra addresses for the listener
+		if len(opts.exBind) > 0 {
+			for _, exbd := range opts.exBind {
+				extraAddress := &listener.AdditionalAddress{
+					Address: util.BuildAddress(exbd, uint32(opts.port.Port)),
+				}
+				res.AdditionalAddresses = append(res.AdditionalAddresses, extraAddress)
```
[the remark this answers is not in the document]
zhlsunshine: Hi @howardjohn, yes, `AdditionalAddresses` is a new field. And yes, but how about to check the Istio version?
BTW, when I execute cmd: `make build` it's okay with the new field of `AdditionalAddresses`, but when I execute cmd `make test`, there are errors message like below:

```
{Failed      listener_builder_test.go:79: listener 127.0.0.1_15007 is invalid: invalid Listener.AdditionalAddresses[0]: embedded message failed validation | caused by: invalid AdditionalAddress.Address: embedded message failed validation | caused by: invalid Address.SocketAddress: embedded message failed validation | caused by: invalid SocketAddress.Address: value length must be at least 1 runes}
```
So I think `make build` or `make test` have different dependency libs version for `github.com/envoyproxy/go-control-plane/envoy/config/listener/v3`  Any idea?

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
zhlsunshine: Hi @howardjohn, according to original solution, there are 2 listeners for a service with both IPv4 and IPv6 addresses. For current solution, there is only 1 listener and there are address and additional addresses (new field) for it. The former is for the first address of the service, and the latter is for the rest addresses of the service.

pilot/pkg/networking/core/v1alpha3/listener_builder.go
zhlsunshine: Sure, will refactor them

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: done

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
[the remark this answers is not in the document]
zhlsunshine: done

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: done

pilot/pkg/networking/core/v1alpha3/listener.go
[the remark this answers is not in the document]
zhlsunshine: Added a new function named `IsIstioVersionGE116` to make sure that the field of `AdditionalAddresses` can be added if istioctl version is greater or equal 1.16.

pilot/pkg/networking/core/v1alpha3/listener_builder.go
zhlsunshine: done, mark them with sufix `IPv4` and `IPv6`

hzxuzhonghu:
cc @hzxuzhonghu @ramaraochavali

pilot/pkg/networking/core/v1alpha3/listener.go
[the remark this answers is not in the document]
jacob-delgado: You need go.mod updated to use https://github.com/envoyproxy/go-control-plane/commit/2cb7986b23ccfde4a77c26ebf3bfbfa71126fafd

Try running `go get github.com/envoyproxy/go@2cb7986b23ccfde4a77c26ebf3bfbfa71126fafd for your PR.

I have a PR to update the to a version of go.mod, but it's failing due to other reasons.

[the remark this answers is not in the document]
jacob-delgado: It appears that updating the api is also blocked on https://github.com/envoyproxy/envoy/pull/22804 as definitions were removed that Istio relies on for other features. Maybe you can get Alex to help you Steve to get it pushed through Envoy. This is blocking Istio as well.

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -374,12 +378,20 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
 			// If captureMode is not NONE, i.e., bindToPort is false, then
 			// we will bind to user specified IP (if any) or to the VIPs of services in
 			// this egress listener.
-			bind := egressListener.IstioListener.Bind
+			var bind string
+			var extraBind []string
+			bind = egressListener.IstioListener.Bind
```
howardjohn, line 383:
```suggestion
			bind := egressListener.IstioListener.Bind
```

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -451,6 +467,16 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
 
 					// bind might have been modified by below code, so reset it for every Service.
 					listenerOpts.bind = bind
+					// bind and extraBind should be specified at the same time for dual stack env
+					// and make sure that it's necessary to add the extra addresses, such as it's a dual stack env
+					if actualLocalHostAddrIPv6 != "" {
+						if bind == actualLocalHostAddrIPv4 {
+							listenerOpts.extraBind = []string{actualLocalHostAddrIPv6}
```
howardjohn, line 474: do we want to make `bind` and `extraBind`? Should we just have `bind` be a list and we automatically make them "extra"?

In envoy, is there any difference between the primary address and the "extra" addresses?

zhlsunshine: > do we want to make `bind` and `extraBind`? Should we just have `bind` be a list and we automatically make them "extra"?
> 
> In envoy, is there any difference between the primary address and the "extra" addresses?

Hi @howardjohn, as discussing in previous PR, `ClusterVIPs` can holds all IP instances which is available both for `bind` and `extraBind`.

Yes, according to the issue https://github.com/envoyproxy/envoy/issues/11184 in envoy, `additinal_addresses` can hold all other IP instances for one specified listener, except the primary one. And [here ](https://github.com/envoyproxy/envoy/pull/21391/files)is the API definition for it.

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: okay, done!

pilot/pkg/networking/core/v1alpha3/listener.go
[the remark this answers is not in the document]
zhlsunshine: Hi @jacob-delgado, thanks, this error has been fixed!

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: @howardjohn  For example, if there are 2 IP instances ["127.0.0.1", "::1"],  then the primary address should be `127.0.0.1`, and the extra addresses should be [`::1`]

zhlsunshine:
Hi @howardjohn, could you please help to approve this PR if it's okay?  Thanks! ^_^

pilot/pkg/networking/core/v1alpha3/listener.go
howardjohn: I get that's how it's implemented but my question is why? Do we really need to carry around a "address" and "extra addresses" or just a list of addresses and stick a random one as the main on in envoy config?

My concern with this PR, as noted before, is not the generates XDS but the large complexity we are adding to the code for something relatively simple, and lack of integration tests.

If we need all of these complex if statements and juggling around 4 address (localhost + wildcard x v4 v6) it's complex and error prone. Maybe this logic can be factored out into a struct, etc

zhlsunshine: Hi @howardjohn, how about to factor the extra addresses adding logic into the function of structs `buildListenerOpts` and `inboundChainConfig`, then the similar if statements, such as `if len(cc.extraBind) > 0 && util.IsIstioVersionGE116(lb.node.IstioVersion)` will be eliminated.
For integration test, the dual stack integration test will be added in later and the use case would be based on the really user scenarios. @kfaseela, @zhanghw0354 and I would co-work on it.

howardjohn: We don't need to make all of our code much more complex just because envoy has a weird API, imo. We should pick the right API for our usage and then translate it to envoy as needed

howardjohn: Integration tests are an expectation for all features in Istio; please have them alongside the PR instead of afterwards. We can wait. If you want to merge without integration tests it can be done in the experimental branch

zhlsunshine: > We don't need to make all of our code much more complex just because envoy has a weird API, imo. We should pick the right API for our usage and then translate it to envoy as needed

Sure,  agree with this idea, I will make the code as simple as possible, then re-submit it.

zhlsunshine: > Integration tests are an expectation for all features in Istio; please have them alongside the PR instead of afterwards. We can wait. If you want to merge without integration tests it can be done in the experimental branch

@howardjohn  Sure, I will add integration tests for this feature, and make sure the `additional_addresses` can be correctly added. Thanks for the comment here.

hzxuzhonghu (commented):
Do we have integration test which has dual nics?

zhlsunshine:
> Do we have integration test which has dual nics?

Hi @hzxuzhonghu , that's a good question,  in fact @jacob-delgado is focusing on building the integration test environment for dual stack.

zhlsunshine:
Hi @howardjohn & @hzxuzhonghu, I refactored the code, so the duplicated code for adding extra addresses to a listener have been eliminated, then merged `getDualStackWildcard` and `getDualStackLocalHost` into `getDualStackWildcardAndLocalHost`, so many if statements have also been eliminated. I think the code is more brief than before. and I also added dual-stack proxy for unit tests. And @jacob-delgado also is working on making the integration test ready.
Hope you know this, and could help to review and approve this PR. Thanks!

zhlsunshine:
BTW,  here is the newest configdump file for this change.
[newest_extra_addresses_configdump.txt](https://github.com/istio/istio/files/9493649/newest_extra_addresses_configdump.txt)

jacob-delgado:
Put up a PR to do similar testing for dual-stack clusters similar to how it's done for ipv6 (https://github.com/istio/test-infra/pull/4278). But we still need to create a specific test for this based on this set of changes.

zhlsunshine:
Hi @howardjohn, besides the unit test, the integration test for dual stack `integ-ds_istio` has been added as well, could you please help to review and approve this PR? Thanks! ^_^

howardjohn:
There is a integ tests with a cluster running dual stack, but everything running inthe cluster is single stack so I don't think we actually have any tests coverage?

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -291,7 +291,11 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
 ) []*listener.Listener {
 	noneMode := node.GetInterceptionMode() == model.InterceptionNone
 
-	actualWildcard, actualLocalHostAddress := getActualWildcardAndLocalHost(node)
+	actualWildcardIPv4, actualWildcardIPv6, actualLocalHostAddrIPv4, actualLocalHostAddrIPv6, isDualStack := getDualStackWildcardAndLocalHost(node)
```
howardjohn, line 294: My concerns around the complexity of this function are still present. Please abstract this into a struct or other structured form; a 5 tuple respones is almost impossible to use correctly and adds tons of complexity to all usages (as can be seen at each callsite in this PR)

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -132,6 +135,11 @@ func (lb *ListenerBuilder) buildVirtualOutboundListener() *ListenerBuilder {
 		FilterChains:     filterChains,
 		TrafficDirection: core.TrafficDirection_OUTBOUND,
 	}
+	// add extra addresses for the listener
+	err := util.BuildExtraAddresses([]string{extrActualWildcard}, uint32(lb.push.Mesh.ProxyListenPort), ipTablesListener, lb.node)
```
howardjohn, line 139: BuildExtraAddresses shouldn't return an error IMO

zhlsunshine:
> There is a integ tests with a cluster running dual stack, but everything running inthe cluster is single stack so I don't think we actually have any tests coverage?

Yeah, @howardjohn,  correct, there is still no dual stack test coverage, however, we can make sure that all dual stack code change would not impact ipv4/ipv6 only use case with `inte-ds_istio` job. There would be dual stack test cases be added with new flag `istio.test.enableDualStack` in later

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: okay, sure, I will abstract it into a struct, thanks a lot for your comments @howardjohn!

pilot/pkg/networking/core/v1alpha3/listener_builder.go
zhlsunshine: okay, if so, I will make `BuildExtraAddresses` without any result.

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: And I may add this into Proxy struct.

zhlsunshine:
Hi @howardjohn, according to above comments, I did some changes as below:
1.  Added a struct named `WildcardAndLocalHost` to hold the wildcard and localhost addresses. And there is no model proxy for this struct, because it may be re-used by other external services, such as service entry.
2.  `GetWildcardAddresses()` and `GetLocalHostAddresses()` are used to separated when using in different places, and no matter which ip family is, the first address of service can bind to listener, the rest can be added in additional addresses.
3.  Fixed a bug for bench_test when creating a node proxy based on the current implementation.
Hope you know this, and thank you so much for your review and comments for this PR! ^_^

zhlsunshine:
Hi @howardjohn, according to above comments, I did some changes as below:
1.  Added a struct named `WildcardAndLocalHost` to hold the wildcard and localhost addresses. And there is no model proxy for this struct, because it may be re-used by other external services, such as service entry.
2.  `GetWildcardAddresses()` and `GetLocalHostAddresses()` are used to separated when using in different places, and no matter which ip family is, the first address of service can bind to listener, the rest can be added in additional addresses.
3.  Fixed a bug for bench_test when creating a node proxy based on the current implementation.

Here is the newest configdump file: 
[the_newest_extra_addresses_configdump.txt](https://github.com/istio/istio/files/9615201/the_newest_extra_addresses_configdump.txt)
Hope you know this, and thank you so much for your reviewing and commenting for this PR! ^_^

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -121,17 +121,31 @@ func (lb *ListenerBuilder) buildVirtualOutboundListener() *ListenerBuilder {
...
 	filterChains := buildOutboundCatchAllNetworkFilterChains(lb.node, lb.push)
 
-	actualWildcard, _ := getActualWildcardAndLocalHost(lb.node)
+	oWildcardAndLocalHost := NewWildcardAndLocalHost(lb.node.GetIPMode())
+	if oWildcardAndLocalHost == nil {
+		log.Warnf("inboundVirtualListener: Can not fetch wildcar and localhost from proxy [%s]", lb.node.ID)
+		return nil
+	}
+	actualWildcards := oWildcardAndLocalHost.GetWildcardAddresses()
+	if len(actualWildcards) == 0 {
```
howardjohn, line 130: this is impossible, we can drop this

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
```diff
@@ -280,11 +300,23 @@ func (lb *ListenerBuilder) buildInboundChainConfigs() []inboundChainConfig {
 				Protocol:   i.ServicePort.Protocol,
 			}
 
+			var bindAddr, extraBindAddrs string
```
howardjohn, line 303: should we be using the NewWildcardAndLocalHost here?

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
```diff
@@ -329,6 +361,20 @@ func (lb *ListenerBuilder) buildInboundChainConfigs() []inboundChainConfig {
 			if cc.bind == "" {
 				// If user didn't provide, pick one based on IP
 				cc.bind = getSidecarInboundBindIP(lb.node)
+				// if no global unicast address
+				if cc.bind == WildcardAddress || cc.bind == WildcardIPv6Address {
+					oWildcardAndLocalHost := NewWildcardAndLocalHost(lb.node.GetIPMode())
+					if oWildcardAndLocalHost == nil {
+						log.Warnf("inboundVirtualListener: Can not fetch wildcar and localhost from proxy [%s]", lb.node.ID)
```
howardjohn, line 368: No need for this as mentioned previously

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
```diff
@@ -329,6 +361,20 @@ func (lb *ListenerBuilder) buildInboundChainConfigs() []inboundChainConfig {
 			if cc.bind == "" {
 				// If user didn't provide, pick one based on IP
 				cc.bind = getSidecarInboundBindIP(lb.node)
+				// if no global unicast address
+				if cc.bind == WildcardAddress || cc.bind == WildcardIPv6Address {
+					oWildcardAndLocalHost := NewWildcardAndLocalHost(lb.node.GetIPMode())
```
howardjohn, line 366: This logic seems a bit weird. We call getSidecarInboundBindIP which computes the wildcard address. Then we check if it is a wildcard, we compute the wildcard _again_ (with slightly different logic than getSidecarInboundBindIP/getActualWildcardAndLocalHost) and overwrite it.

Maybe we should just make getSidecarInboundBindIP use NewWildcardAndLocalHost to begin with?

howardjohn:
Benchmarks show some troubling results: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/40634/benchmark_istio/1572678114784317440

maybe due to adding ipDiscovery call? Might be worth breaking that out so we can analyze the main PR in isolation.

pilot/pkg/networking/util/util.go
```diff
@@ -173,6 +173,23 @@ func BuildAddress(bind string, port uint32) *core.Address {
...
+				continue
+			}
+			extraAddress := &listener.AdditionalAddress{
+				Address: BuildAddress(exbd, listenPort),
+			}
+			ll.AdditionalAddresses = append(ll.AdditionalAddresses, extraAddress)
+		}
+		return
+	}
+	log.Errorf("error: No extra addresses need to be added to listener [%s]", ll.Name)
```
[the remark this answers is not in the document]
zhlsunshine: okay, change it into a warning.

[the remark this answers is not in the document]
howardjohn: There should be no log at all - it's not an issue or relevant to the user?

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -121,17 +121,31 @@ func (lb *ListenerBuilder) buildVirtualOutboundListener() *ListenerBuilder {
...
+		log.Warnf("inboundVirtualListener: actualWildcard addresses can not be fetched in [%s]", lb.node.ID)
+		return nil
+	}
 
 	// add an extra listener that binds to the port that is the recipient of the iptables redirect
 	ipTablesListener := &listener.Listener{
 		Name:             model.VirtualOutboundListenerName,
```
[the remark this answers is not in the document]
zhlsunshine: okay, mark this one as a remind.

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
```diff
@@ -175,7 +178,10 @@ func (lb *ListenerBuilder) buildInboundListeners() []*listener.Listener {
 
 		if cc.bindToPort {
 			// If this config is for bindToPort, we want to actually create a real Listener.
-			listeners = append(listeners, lb.inboundCustomListener(cc, chains))
+			l := lb.inboundCustomListener(cc, chains)
+			// add extra addresses for the listener
+			util.BuildExtraAddresses(cc.extraBind, cc.port.TargetPort, l, lb.node)
```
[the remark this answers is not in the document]
zhlsunshine: done, I moved `BuildExtraAddresses` into `buildInboundListener`

pilot/pkg/networking/core/v1alpha3/listener_address.go
```diff
@@ -36,6 +36,60 @@ const (
...
+	case model.IPv6:
+		wcLh.wildCardIPv6 = WildcardIPv6Address
+		wcLh.localHostIPv6 = LocalhostIPv6Address
+	case model.Dual:
+		wcLh.wildCardIPv4 = WildcardAddress
+		wcLh.wildCardIPv6 = WildcardIPv6Address
+		wcLh.localHostIPv4 = LocalhostAddress
+		wcLh.localHostIPv6 = LocalhostIPv6Address
+	default:
+		return nil
```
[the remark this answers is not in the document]
zhlsunshine: okay, panic is here

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
```diff
@@ -192,14 +198,28 @@ func (lb *ListenerBuilder) buildInboundListeners() []*listener.Listener {
 
 // inboundVirtualListener builds the virtual inbound listener.
 func (lb *ListenerBuilder) inboundVirtualListener(chains []*listener.FilterChain) *listener.Listener {
-	actualWildcard, _ := getActualWildcardAndLocalHost(lb.node)
+	oWildcardAndLocalHost := NewWildcardAndLocalHost(lb.node.GetIPMode())
+	if oWildcardAndLocalHost == nil {
+		log.Warnf("inboundVirtualListener: Can not fetch wildcar and localhost from proxy [%s]", lb.node.ID)
```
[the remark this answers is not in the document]
zhlsunshine: okay, `oWildcardAndLocalHost` should be there once `NewWildcardAndLocalHost` is called

pilot/pkg/networking/core/v1alpha3/listener_address.go
```diff
@@ -36,6 +36,60 @@ const (
...
+	}
+	if wl.localHostIPv6 != "" {
+		addresses = append(addresses, wl.localHostIPv6)
+	}
+	return addresses
+}
+
 // getActualWildcardAndLocalHost will return corresponding Wildcard and LocalHost
 // depending on value of proxy's IPAddresses.
 func getActualWildcardAndLocalHost(node *model.Proxy) (string, string) {
```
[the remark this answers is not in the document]
zhlsunshine: Yeah, correct, added TODO for removing it in future

pilot/pkg/networking/core/v1alpha3/listener_address.go
```diff
@@ -36,6 +36,60 @@ const (
 	InboundPassthroughBindIpv6 = "::6"
 )
 
+type WildcardAndLocalHost struct {
```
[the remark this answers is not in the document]
zhlsunshine: sure, changed as comments

pilot/pkg/networking/core/v1alpha3/listener_builder.go
zhlsunshine: dropped for all

pilot/pkg/networking/core/v1alpha3/listener_address.go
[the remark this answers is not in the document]
ramaraochavali: or move Wildcards() and Localhosts() on proxy as it has ipMode already and we can compute these when ipMode is detected?

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -291,7 +291,17 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
 ) []*listener.Listener {
 	noneMode := node.GetInterceptionMode() == model.InterceptionNone
 
-	actualWildcard, actualLocalHostAddress := getActualWildcardAndLocalHost(node)
+	oWildcardAndLocalHost := NewWildcardAndLocalHost(node.GetIPMode())
```
ramaraochavali, line 294: Have methods like Wildcards() and Localhosts() on node it self. That simplifies lot of code I guess.

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -291,7 +291,17 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
...
+	if oWildcardAndLocalHost == nil {
```
ramaraochavali, line 295: This will never be nil?

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -291,7 +291,17 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
...
+		log.Warnf("buildSidecarOutboundListeners: Can not fetch wildcar and localhost from proxy [%s]", node.ID)
+		return nil
+	}
+	actualWildcards := oWildcardAndLocalHost.GetWildcardAddresses()
+	actualLocalHosts := oWildcardAndLocalHost.GetLocalHostAddresses()
+	if len(actualWildcards) == 0 || len(actualLocalHosts) == 0 {
```
ramaraochavali, line 301: Same here. When will these be empty? once we detect proxy IPMode, these will have some values.

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
zhlsunshine: Sure, u r right!

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
zhlsunshine: done

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
zhlsunshine: I prefer the latter, changed `getSidecarInboundBindIP` into `getSidecarInboundBindIPs`, and use `NewHostAddresses` at the beginning.

zhlsunshine:
> Benchmarks show some troubling results: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/40634/benchmark_istio/1572678114784317440
> 
> maybe due to adding ipDiscovery call? Might be worth breaking that out so we can analyze the main PR in isolation.

It seems that the job is passed,  I think `ipDiscovery` should be called once the proxy node is created and it should be required in later.

ramaraochavali (commented):
Few more comments

zhlsunshine:
Hi @ramaraochavali, I think your above comments have mentioned that `Wildcards()` and `Localhosts()` may be moved into `proxy` struct. However, to import a new struct now named `HostAddresses` is not only for normal service whose pods hold the injected proxy, but also some other service, such as service entry which has no any injected proxy.

howardjohn:
> It seems that the job is passed, I think ipDiscovery should be called once the proxy node is created and it should be required in later.

Its a benchmark, its not pass/fail. It shows 2x cost on almost all tests

howardjohn:
No, look at the logs. It shows a diff for each benchmark

howardjohn (commented):
I broke out part into https://github.com/istio/istio/pull/41120 so we can see just the diff from the real changes here

zhlsunshine:
Hi @howardjohn, I compared the 2 bench tests (1572678114784317440 for this PR and 1572987049097564160 for PR#41120 ) log, I noticed that there are some downloading job before the bench test begin in former. 
Besides that, the big differences are focusing on `BenchmarkStats` in `istio.io/istio/pilot/cmd/pilot-agent/status`
It took 2x costs at the beginning of the bench test but less at the end in former, it's opposite for the latter.

zhlsunshine:
This is 1572678114784317440 for this PR:
image

This is 1572987049097564160 for PR#41120:
image

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -373,12 +375,19 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
 			// If captureMode is not NONE, i.e., bindToPort is false, then
 			// we will bind to user specified IP (if any) or to the VIPs of services in
 			// this egress listener.
+			var extraBind []string
 			bind := egressListener.IstioListener.Bind
 			if bind == "" {
 				if bindToPort {
-					bind = actualLocalHostAddress
+					bind = actualLocalHosts[0]
```
hzxuzhonghu, line 382: It is hard to understand what 0 indicate

zhlsunshine: there should be 2 addresses for actualLocalhosts, one is for IPv4 127.0.0.1, another is for IPv6 ::1.  And actualLocalhosts[0] is the first localhost address defined in application's service yaml file with IPFamilies.

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -121,17 +121,23 @@ func (lb *ListenerBuilder) buildVirtualOutboundListener() *ListenerBuilder {
 
 	filterChains := buildOutboundCatchAllNetworkFilterChains(lb.node, lb.push)
 
-	actualWildcard, _ := getActualWildcardAndLocalHost(lb.node)
+	oWildcardAndLocalHost := NewHostAddresses(lb.node.GetIPMode())
```
hzxuzhonghu, line 124: I think instead of constructing each time, we can define several vars mapping to IP mode

hzxuzhonghu (commented):
Generally, looks good. But i am concerned with the hacking into so many places, can we reduce

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -291,7 +291,9 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
 ) []*listener.Listener {
 	noneMode := node.GetInterceptionMode() == model.InterceptionNone
 
-	actualWildcard, actualLocalHostAddress := getActualWildcardAndLocalHost(node)
+	oWildcardAndLocalHost := NewHostAddresses(node.GetIPMode())
```
ramaraochavali, line 294: nit: oWildcardAndLocalHost -> hostAddresses

ramaraochavali: same every where

pilot/pkg/networking/core/v1alpha3/listener_builder.go
ramaraochavali: This was one of my suggestions of adding it to node structure. So that majority of cases we are covered, we compute only when we detect IP mode

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: sure, I will change all places.

zhlsunshine: done @ramaraochavali

zhlsunshine:
> Generally, looks good. But i am concerned with the hacking into so many places, can we reduce

Hi @hzxuzhonghu, I understand your worries, however, all these places are required to be changed for generating the necessary configuration.

howardjohn:
The `kb/msg` metric is concerning:

```
name                                              old kb/msg         new kb/msg         delta
pkg:istio.io/istio/pilot/pkg/xds goos:linux goarch:amd64
RouteGeneration/gateways-8                               1.16k ± 0%         1.16k ± 0%      ~     (all equal)
RouteGeneration/gateways-shared-8                        5.37k ± 0%         5.37k ± 0%      ~     (all equal)
RouteGeneration/empty-8                                   38.9 ± 0%          77.8 ± 0%   +99.77%  (p=0.008 n=5+5)
RouteGeneration/tls-8                                     38.9 ± 0%          77.8 ± 0%   +99.77%  (p=0.008 n=5+5)
RouteGeneration/telemetry-8                               38.9 ± 0%          77.8 ± 0%   +99.77%  (p=0.008 n=5+5)
RouteGeneration/telemetry-api-8                           38.9 ± 0%          77.8 ± 0%   +99.77%  (p=0.008 n=5+5)
RouteGeneration/virtualservice-8                          96.6 ± 0%          96.6 ± 0%      ~     (all equal)
RouteGeneration/authorizationpolicy-8                     0.41 ± 0%          0.41 ± 0%      ~     (all equal)
RouteGeneration/peerauthentication-8                      0.41 ± 0%          0.41 ± 0%      ~     (all equal)
RouteGeneration/knative-gateway-8                          258 ± 0%           258 ± 0%      ~     (all equal)
RouteGeneration/serviceentry-workloadentry-8              38.9 ± 0%          77.8 ± 0%   +99.77%  (p=0.008 n=5+5)
ClusterGeneration/gateways-8                               429 ± 0%           429 ± 0%      ~     (all equal)
ClusterGeneration/gateways-shared-8                        429 ± 0%           429 ± 0%      ~     (all equal)
ClusterGeneration/empty-8                                  375 ± 0%           376 ± 0%    +0.08%  (p=0.008 n=5+5)
ClusterGeneration/tls-8                                    337 ± 0%           338 ± 0%    +0.09%  (p=0.008 n=5+5)
ClusterGeneration/telemetry-8                              188 ± 0%           189 ± 0%    +0.11%  (p=0.008 n=5+5)
ClusterGeneration/telemetry-api-8                          188 ± 0%           189 ± 0%    +0.11%  (p=0.008 n=5+5)
ClusterGeneration/virtualservice-8                        43.4 ± 0%          43.6 ± 0%    +0.58%  (p=0.008 n=5+5)
ClusterGeneration/authorizationpolicy-8                   1.49 ± 0%          1.75 ± 0%   +16.96%  (p=0.008 n=5+5)
ClusterGeneration/peerauthentication-8                    1.49 ± 0%          1.75 ± 0%   +16.96%  (p=0.008 n=5+5)
ClusterGeneration/knative-gateway-8                       1.13 ± 0%          1.13 ± 0%      ~     (all equal)
ClusterGeneration/serviceentry-workloadentry-8             395 ± 0%           395 ± 0%    +0.08%  (p=0.008 n=5+5)
ListenerGeneration/gateways-8                            2.09k ± 0%         2.10k ± 0%    +0.29%  (p=0.008 n=5+5)
ListenerGeneration/gateways-shared-8                      4.00 ± 0%          4.03 ± 0%    +0.75%  (p=0.008 n=5+5)
ListenerGeneration/empty-8                                18.5 ± 0%          46.6 ± 0%  +152.41%  (p=0.008 n=5+5)
ListenerGeneration/tls-8                                  12.6 ± 0%          38.2 ± 0%  +202.54%  (p=0.008 n=5+5)
ListenerGeneration/telemetry-8                            29.1 ± 0%         108.1 ± 0%  +271.48%  (p=0.008 n=5+5)
ListenerGeneration/telemetry-api-8                        54.7 ± 0%         232.4 ± 0%  +325.10%  (p=0.008 n=5+5)
ListenerGeneration/virtualservice-8                       7.32 ± 0%         13.16 ± 0%   +79.90%  (p=0.008 n=5+5)
ListenerGeneration/authorizationpolicy-8                  56.4 ± 0%         109.9 ± 0%   +95.00%  (p=0.008 n=5+5)
ListenerGeneration/peerauthentication-8                   6.41 ± 0%          9.83 ± 0%   +53.45%  (p=0.008 n=5+5)
ListenerGeneration/knative-gateway-8                      3.56 ± 0%          3.59 ± 0%    +0.84%  (p=0.008 n=5+5)
ListenerGeneration/serviceentry-workloadentry-8           18.5 ± 0%          46.6 ± 0%  +152.41%  (p=0.008 n=5+5)
```

Why are we seeing such a huge increase in config size?

zhlsunshine:
Hi @howardjohn, thanks for your comparison above, then I see no. and I also checked the comparison between`1557747988594429952 by ramaraochavali at Aug 11 23:17:51` and `1572987049097564160 by howardjohn at Sep 23 00:37:35`.
As you commented here: https://github.com/istio/istio/pull/40634#issuecomment-1254226823, I think you are correct that `ipDiscovery` called in `pilot/pkg/xds/bench_test.go` causes such huge increase in config size.
Moreover, I also checked the comparison between `1572987049097564160 by howardjohn at Sep 23 00:37:35` and `1573129384842235904	by zhlsunshine at Sep 23 09:58:00`, there is no huge increase for them.

According to current analysis, this PR does not cause the increasing in configuration size if it drops the bech_test's `DiscoverIPMode` calling. However, in fact, it confused me that `DiscoverIPMode` calling causes such huge increase in configuration size. You know it just sets the IP mode and GlobalUnicastIP of the proxy.

zhlsunshine:
Hi @howardjohn, thanks for your comparison above, then I see now. and I also checked the comparison between`1557747988594429952 by ramaraochavali at Aug 11 23:17:51` and `1572987049097564160 by howardjohn at Sep 23 00:37:35`.
As you commented here: https://github.com/istio/istio/pull/40634#issuecomment-1254226823, I think you are correct that `DiscoverIPMode` called in `pilot/pkg/xds/bench_test.go` causes such huge increase in config size.
Moreover, I also checked the comparison between `1572987049097564160 by howardjohn at Sep 23 00:37:35` and `1573129384842235904	by zhlsunshine at Sep 23 09:58:00`, there is no huge increase for them.

According to current analysis, this PR does not cause the increasing in configuration size if it drops the bech_test's `DiscoverIPMode` calling. However, in fact, it confused me that `DiscoverIPMode` calling causes such huge increase in configuration size. You know it just sets the IP mode and GlobalUnicastIP of the proxy.

pilot/pkg/networking/core/v1alpha3/listener_address.go
```diff
@@ -36,6 +36,61 @@ const (
...
+	case model.IPv6:
+		wcLh.wildCardIPv6 = WildcardIPv6Address
+		wcLh.localHostIPv6 = LocalhostIPv6Address
+	case model.Dual:
+		wcLh.wildCardIPv4 = WildcardAddress
+		wcLh.wildCardIPv6 = WildcardIPv6Address
+		wcLh.localHostIPv4 = LocalhostAddress
+		wcLh.localHostIPv6 = LocalhostIPv6Address
+	default:
+		panic("Unknow IP mode")
```
howardjohn, line 62:
```suggestion
		panic("Unknown IP mode")
```

pilot/pkg/networking/core/v1alpha3/listener.go
howardjohn: Why do we need `extraBind` at all? Instead of `bind` being a list?

I am very worried that somewhere else we will assume there is one bind and mutate it.

For example:

```
bind := 127.0.0.1
extraBind := ::1
... lot of code later ..
bind = SomeOtherIp // Oops! forgot to change extrabind!
```

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -390,11 +399,14 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
 				port:       listenPort,
 				bindToPort: bindToPort,
 			}
+			if len(extraBind) > 0 {
+				listenerOpts.extraBind = append(listenerOpts.extraBind, extraBind...)
+			}
```
howardjohn, line 404:
```suggestion
				bindToPort: bindToPort,
				extraBind: extraBind,
			}
```

pilot/pkg/networking/util/util.go
```diff
@@ -173,6 +173,23 @@ func BuildAddress(bind string, port uint32) *core.Address {
...
+				continue
+			}
+			extraAddress := &listener.AdditionalAddress{
+				Address: BuildAddress(exbd, listenPort),
+			}
+			ll.AdditionalAddresses = append(ll.AdditionalAddresses, extraAddress)
+		}
+		return
+	}
+	log.Warnf("warning: No extra addresses need to be added to listener [%s]", ll.Name)
```
howardjohn, line 190: Remove this

pilot/pkg/networking/core/v1alpha3/listener_address.go
zhlsunshine: done!

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: Hi @howardjohn, `bind == ""` is a condition to do such specification for the `bind`, if bind is not empty, it should not be some other IP. This is reasonable.

pilot/pkg/networking/util/util.go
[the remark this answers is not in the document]
zhlsunshine: okay, remove it

pilot/pkg/networking/core/v1alpha3/listener_builder.go
zhlsunshine: Hi @hzxuzhonghu & @ramaraochavali,  besides the service with proxy, we also need to think about the service without proxy, so we may need to balance. Any idea?

pilot/pkg/networking/util/util.go
zhlsunshine: done

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: done!

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: Hi @howardjohn,  please ignore my previous answer. I understand your concerns, in fact listenerOpt's `bind` would not be specified in other place once it's available in upper function, such as `buildSidecarOutboundListeners`. But there is an except in `buildSidecarOutboundTCPListenerOptsForPortOrUDS` when `bind` is empty like below:
```
......
if len(listenerOpts.bind) == 0 {
                svcListenAddress := listenerOpts.service.GetAddressForProxy(listenerOpts.proxy)
                svcExtraListenAddress := listenerOpts.service.GetExtraAddressesForProxy(listenerOpts.proxy)
                // We should never get an empty address.
                // This is a safety guard, in case some platform adapter isn't doing things
                // properly
                if len(svcListenAddress) > 0 {
                        if !strings.Contains(svcListenAddress, "/") {
                                listenerOpts.bind = svcListenAddress
                        } else {
                                // Address is a CIDR. Fall back to 0.0.0.0 and
                                // filter chain match
                                destinationCIDR = svcListenAddress
                                listenerOpts.bind = actualWildcard
                        }
                        listenerOpts.extraBind = svcExtraListenAddress
                }
        }
......
```
so I think the situation is under the control.

pilot/pkg/networking/core/v1alpha3/listener_builder.go
hzxuzhonghu: Not sure what you mean, a service without proxy can not be run with this function

ramaraochavali: I get that part. But I think having it node will help majority cases as I explained.  For others we can build on the fly. Also buildVirtualOutboundListener is for regular proxies.

zhlsunshine: Hi @ramaraochavali & @hzxuzhonghu,  as you required, I move the `HostAddresses` logic into proxy node now. And it's okay to build the abnormal situation on the fly. Thanks for your comments! ^_^

pilot/pkg/networking/core/v1alpha3/listener_address.go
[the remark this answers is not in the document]
zhlsunshine: > or move Wildcards() and Localhosts() on proxy as it has ipMode already and we can compute these when ipMode is detected?

done

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: Hi @ramaraochavali, changed them into proxy node.

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: done

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: refactor

zhlsunshine: refactored

pilot/pkg/model/context.go
```diff
@@ -306,6 +306,12 @@ type Proxy struct {
 	// IPMode of proxy.
 	ipMode IPMode
 
+	// wildCardIPs contains the wildcard host IPs, including IPv4 and IPv6
+	wildCardIPs []string
+
+	// localHostIPs contains the wildcard host IPs, including IPv4 and IPv6
```
ramaraochavali, line 312: nit: localHostIPs contains the local host IPs,

pilot/pkg/networking/core/v1alpha3/listener.go
ramaraochavali: can you please add a comment on why we are using bind = actualLocalHosts[0]?

ramaraochavali (commented):
Generally LGTM. Couple of nits. Still not clear what is the issue with bench marks though

pilot/pkg/model/context.go
zhlsunshine: done!

zhlsunshine: done!

pilot/pkg/networking/core/v1alpha3/listener.go
zhlsunshine: sure, I added comment for both `bind = actualLocalHosts[0]` and `bind = actualWildcards[0]` here.

hzxuzhonghu (commented):
One question, when user specify Bind explicitly, will we add additional Bind now?

zhlsunshine:
> One question, when user specify Bind explicitly, will we add additional Bind now?

Hi @hzxuzhonghu, what do you mean `specify Bind explicitly`?  the additional bind is added when istio service has more than one addresses.

pilot/pkg/model/context.go
```diff
@@ -884,14 +890,39 @@ func (node *Proxy) SetWorkloadLabels(env *Environment) {
...
+		wildCardIPs = append(wildCardIPs, constants.WildcardIPv6Address)
+		localHostIPs = append(localHostIPs, constants.LocalhostIPv6Address)
 	} else {
 		node.ipMode = Dual
+		wildCardIPs = append(wildCardIPs, []string{constants.WildcardAddress, constants.WildcardIPv6Address}...)
+		localHostIPs = append(localHostIPs, []string{constants.LocalhostAddress, constants.LocalhostIPv6Address}...)
 	}
 	node.GlobalUnicastIP = networkutil.GlobalUnicastIP(node.IPAddresses)
+	node.wildCardIPs = wildCardIPs
+	node.localHostIPs = localHostIPs
```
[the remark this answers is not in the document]
zhlsunshine: Hi @hzxuzhonghu, in fact they are not caching same,  the content for `wildCardIPs`  and `localHostIPs` depends on the service deployment, it can be, for example to `localHostIPs`, ["127.0.0.1"], ["::1"] or ["127.0.0.1", "::1"].

BTW, 
1. if not store them in proxy, then Istio must calculate them,
2. if not calculate them, then must store them
One of them should be choose.

pilot/pkg/model/context.go
```diff
@@ -884,14 +890,39 @@ func (node *Proxy) SetWorkloadLabels(env *Environment) {
...
+}
+
+// Wildcards returns the wildCard IP addresses for current proxy
+func (node *Proxy) Wildcards() []string {
+	if len(node.wildCardIPs) == 0 {
+		node.DiscoverIPMode()
```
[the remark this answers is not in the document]
zhlsunshine: Hi @hzxuzhonghu, yeah, `DiscoverIPMode` would be called one time if `len(node.wildCardIPs) == 0` or `len(node.localHosts) == 0`. 
In fact, I'm afraid that `DiscoverIPMode` is not called, so just in case. Do you suggest that remove the calling in both `Wildcards()` and `Localhosts()`?

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -373,12 +374,23 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
...
+					}
 				} else {
-					bind = actualWildcard
+					// the first wildcard address would be the binding address and
+					// the rest would be the additional addresses
+					bind = actualWildcards[0]
+					if len(actualWildcards) > 1 {
+						extraBind = actualWildcards[1:]
+					}
 				}
```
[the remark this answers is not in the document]
zhlsunshine: Hi @hzxuzhonghu, I think this may be the same as above answer of mine.

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -389,12 +401,13 @@ func (lb *ListenerBuilder) buildSidecarOutboundListeners(node *model.Proxy,
 				bind:       bind,
 				port:       listenPort,
 				bindToPort: bindToPort,
+				extraBind:  extraBind,
```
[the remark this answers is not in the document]
zhlsunshine: the `extraBind` will be translated into additional addresses for listener in later.

pilot/pkg/networking/util/util.go
```diff
@@ -173,6 +173,22 @@ func BuildAddress(bind string, port uint32) *core.Address {
 	}
 }
 
+// BuildExtraAddresses can add extra addresses to additional addresses for a listener
+func BuildExtraAddresses(extrAddresses []string, listenPort uint32, ll *listener.Listener, node *model.Proxy) {
```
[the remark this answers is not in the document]
zhlsunshine: Hi @hzxuzhonghu, the API definition of listener do not expose the field of `AdditionalAddresses`, moreover, additional addresses are wrapped in listener's socket address struct.

hzxuzhonghu:
Not sure this is right in this case

zhlsunshine:
> Not sure this is right in this case

Hi @hzxuzhonghu, as discussed here: https://github.com/istio/istio/issues/40394, a listener supports multiple addresses is a good idea for dual stack. Moreover, it also can be used for other scenario, such as `tcp headless service` you mentioned. I think it's still valid based on current code change. Any idea?

pilot/pkg/networking/util/util.go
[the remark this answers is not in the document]
zhlsunshine: okay, I understand,  generate the additional addresses in `BuildExtraAddresses` and remove the listener parameter, then specify as `listener.AdditionalAddresses = BuildExtraAddresses()`

pilot/pkg/model/context.go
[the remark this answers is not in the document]
zhlsunshine: removed!

pilot/pkg/networking/util/util.go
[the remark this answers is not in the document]
zhlsunshine: done!

pilot/pkg/model/context.go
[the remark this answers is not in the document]
zhlsunshine: use 2 map to maintain `wildCardIPs` and `localHostIPs`

zhlsunshine:
Hi @hzxuzhonghu, re-submitted, maintain 2 maps to fetch host addresses according to proxy’s IP mode. Thanks for your comments! ^_^

pilot/pkg/networking/util/util.go
```diff
@@ -173,6 +173,23 @@ func BuildAddress(bind string, port uint32) *core.Address {
 	}
 }
 
+// BuildExtraAddresses can add extra addresses to additional addresses for a listener
+func BuildExtraAddresses(extrAddresses []string, listenPort uint32, node *model.Proxy) []*listener.AdditionalAddress {
```
hzxuzhonghu, line 177: replace Extra with Additional

hzxuzhonghu (approved):

pilot/pkg/networking/util/util.go
zhlsunshine: okay, no problem!

zhlsunshine: done!

PR #40642, opened 2022-08-25
validate workloadSelector's empty value

MouceL:

This pr is to solve this problem

apply a enovyfilter with workloadSelector in test ns

```
  workloadSelector:
    labels:
      app: ""
```
all labels of pods in test ns are not included "app" labels,

according to the logic of the current code, this enovyfilter will apply to all pods in ns test.

```
// SubsetOf is true if the label has identical values for the keys
func (i Instance) SubsetOf(that Instance) bool {
	if len(i) == 0 {
		return true
	}

	if len(that) == 0 || len(that) < len(i) {
		return false
	}

	for k, v := range i {
		if that[k] != v {
			return false
		}
	}
	return true
}
```

so we should ensure that the value is not empty

### DIFF
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -775,6 +775,10 @@ func validateAlphaWorkloadSelector(selector *networking.WorkloadSelector) error 
 				errs = appendErrors(errs,
 					fmt.Errorf("empty key is not supported in selector: %q", fmt.Sprintf("%s=%s", k, v)))
 			}
+			if v == "" {
+				errs = appendErrors(errs,
+					fmt.Errorf("empty value is not supported in selector: %q", fmt.Sprintf("%s=%s", k, v)))
+			}
 			if strings.Contains(k, "*") || strings.Contains(v, "*") {
 				errs = appendErrors(errs,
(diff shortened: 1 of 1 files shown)


hzxuzhonghu (commented):
more strictly we should validate similar as k8s

ramaraochavali:
currently it also allows specifying workloadSelector without labels - which also is error prone. We should warn in that case.

howardjohn (approved):
Is empyt value legal in k8s? if so, must be here as well

MouceL:
> Is empyt value legal in k8s? if so, must be here as well

'a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character'

howardjohn:
If it can be an empty string shouldn't we allow that here?

MouceL:
> SubsetOf

To be consistent with k8s, we should not change the value of labels

Maybe we can modify the logic of SubsetOf to solve this problem?

MouceL:
> If it can be an empty string shouldn't we allow that here?

To be consistent with k8s, we should not change the value of labels.

Maybe we can modify the logic of SubsetOf to solve this problem?

PR #40643, opened 2022-08-25
make use of istio backoff wrapper

hzxuzhonghu:

Follow #40629


pkg/backoff/exponential.go
```diff
@@ -57,3 +59,7 @@ func (b *ExponentialBackOff) NextBackOff() time.Duration {
 	}
 	return duration
 }
+
+func (b *ExponentialBackOff) Reset() {
```
howardjohn, line 63: On L58 (can't comment there) shouldn't we do `return b.MaxDuration` not infinite?

I think our expectation is if we have max=30s then backoffs are:

0s, 1s, 2s, 4s, 8s, 16s, 30s, 30s, 30s, 30s

not

0s, 1s, 2s, 4s, 8s, 16s, 30s, infinity

?

pkg/backoff/retry.go
```diff
@@ -0,0 +1,39 @@
...
+		if err = operation(); err == nil {
+			return nil
+		}
+
+		next = b.NextBackOff()
+		if next == MaxDuration {
+			return err
+		}
+
+		time.Sleep(next)
```
howardjohn, line 37: should we pass a context to cancel early?

pkg/backoff/retry.go
```diff
@@ -0,0 +1,39 @@
...
+	var next time.Duration
+
+	for {
+		if err = operation(); err == nil {
+			return nil
+		}
+
+		next = b.NextBackOff()
+		if next == MaxDuration {
+			return err
```
howardjohn, line 34: This behavior was unexpected to me, we should either document it or change? Or maybe make the function named something closer to RetryUntilMaxDurationHit to make it clear

pkg/istio-agent/agent.go
```diff
@@ -450,26 +451,32 @@ func (a *Agent) initSdsServer() error {
...
 // TODO: evaluate replacing the STS server with a file data source, to simplify Envoy config
+// TODO: Fix this method with unused return value
+// nolint: unparam
 func (a *Agent) getWorkloadCerts(st *cache.SecretManagerClient) (sk *security.SecretItem, err error) {
-	b := backoff.NewExponentialBackOff()
-	b.MaxElapsedTime = 0
-
-	for {
+	b := backoff.NewExponentialBackOff(func(off *backoff.ExponentialBackOff) {
+		off.MaxElapsedTime = 0
```
howardjohn, line 458: does this mean we may eventually backoff for days?

howardjohn (commented):
there seems to be some unexpected behavioral changes where we previously retry forever but now exit early.

pkg/istio-agent/agent.go
hzxuzhonghu: Yes, i found it is the original meaning. I am going to fix it.

pkg/backoff/exponential.go
hzxuzhonghu: There are two fields MaxInterval vs MaxElapsedTime

If MaxElapsedTime = 0, we will return 0s, 1s, 2s, 4s, 8s, 16s, 30s, 30s, 30s, 30s,30.........

pkg/backoff/retry.go
hzxuzhonghu: I like the way, will add a context

istioctl/pkg/multicluster/remote_secret.go
```diff
@@ -237,6 +237,7 @@ func waitForTokenData(client kube.CLIClient, secret *v1.Secret) (ca, token []byt
 	}
 
 	log.Infof("Waiting for data to be populated in %s", secret.Name)
+
```
howardjohn, line 240: nit: could remove if you send another change

pkg/backoff/exponential.go
```diff
@@ -38,22 +40,28 @@ type ExponentialBackOff struct {
...
+	b := ExponentialBackOff{}
 	b.ExponentialBackOff = backoff.NewExponentialBackOff()
+	b.ExponentialBackOff.MaxElapsedTime = 0
 	for _, fn := range initFuncs {
-		fn(b)
+		fn(&b)
 	}
 	b.Reset()
-	return b
+	return &b
```
howardjohn, line 52: I thought we had decided to do

```suggestion
	return b
```

or are there issues there?

pkg/backoff/exponential.go
howardjohn: I see. thanks

pkg/istio-agent/agent.go
```diff
@@ -450,26 +451,32 @@ func (a *Agent) initSdsServer() error {
...
 //
 // TODO: evaluate replacing the STS server with a file data source, to simplify Envoy config
+// TODO: Fix this method with unused return value
+// nolint: unparam
 func (a *Agent) getWorkloadCerts(st *cache.SecretManagerClient) (sk *security.SecretItem, err error) {
 	b := backoff.NewExponentialBackOff()
-	b.MaxElapsedTime = 0
-
-	for {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
howardjohn, line 458: Not sure we can feasibly make this change, but I was thinking about this a bit.. its kind of weird that we have a context w/ timeout but also we have MaxElapsedTime - they seem similar purpose. 

Context is still useful if we are canceling for other reason (ctrl+c, etc), but I am not sure if its best to use with timeout?

WDYT? I don't mind either way really but we should probably be consistent/add comments.

pkg/istio-agent/agent.go
```diff
@@ -450,26 +451,32 @@ func (a *Agent) initSdsServer() error {
...
 		log.Warnf("failed to get certificate: %v", err)
-		time.Sleep(b.NextBackOff())
+		return err
+	}, b)
+	if err != nil {
+		return nil, err
 	}
-	for {
+	b.Reset()
+	err = backoff.RetryWithContext(context.TODO(), func() error {
```
howardjohn, line 472: nit: if we use timeout about we probably should here?

pkg/istio-agent/agent.go
```diff
@@ -450,26 +451,32 @@ func (a *Agent) initSdsServer() error {
 // It will not return until both workload cert and root cert are generated.
```
howardjohn, line 451: This comment is no longer accurate I think. Should we preserve old model of retry forever?

pkg/backoff/retry.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"time"
+)
+
+// RetryWithContext tries the operation until it does not return error, BackOff stops,
+// or when the context expires, whichever happens first.
+// o is guaranteed to be run at least once.
+// RetryWithContext sleeps the goroutine for the duration returned by BackOff after a
+// failed operation returns.
+func RetryWithContext(ctx context.Context, operation func() error, b BackOff) error {
```
howardjohn, line 28: nit: should we make it a method on `BackOff` instead of a standalone function?

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -414,16 +412,15 @@ func (sc *SecretManagerClient) generateRootCertFromExistingFile(rootCertPath, re
...
 	// To handle that case, check if cert and key are valid if they are valid then only send to proxy.
-	b := backoff.NewExponentialBackOff()
-	b.InitialInterval = sc.configOptions.FileDebounceDuration
+	b := backoff.NewExponentialBackOff(func(off *backoff.ExponentialBackOff) {
+		off.InitialInterval = sc.configOptions.FileDebounceDuration
+		off.MaxElapsedTime = totalTimeout
+	})
 	secretValid := func() error {
 		_, err := tls.LoadX509KeyPair(certChainPath, keyPath)
-		if errors.Is(err, os.ErrNotExist) {
```
howardjohn, line 421: do we need/want this behavior?

security/pkg/nodeagent/sds/sdsservice.go
```diff
@@ -112,31 +112,30 @@ func newSDSService(st security.SecretManager, options *security.Options, pkpConf
...
 	go func() {
 		b := backoff.NewExponentialBackOff()
-		b.MaxElapsedTime = 0
-		for {
-			_, err := st.GenerateSecret(security.WorkloadKeyCertResourceName)
-			if err == nil {
-				break
-			}
-			sdsServiceLog.Warnf("failed to warm certificate: %v", err)
+		// context for both timeout and channel, whichever stops first, the context will be done
```
howardjohn, line 115: I am not sure we ever want to timeout

pkg/backoff/retry.go
```diff
@@ -0,0 +1,46 @@
...
+
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("%v with last error: %v", context.DeadlineExceeded, err)
+		default:
+			next := b.NextBackOff()
+			if next == MaxDuration {
+				return err
+			}
+			time.Sleep(next)
```
howardjohn, line 43: I think we want `<-time.After(next)` in a select with ctx.Done so it can pre-empt it?

howardjohn (commented):
sorry lots of nitpicks... Looks good except the behavioral changes and Sleep vs After

cc @ramaraochavali would be good to have a second pair of eyes

hzxuzhonghu:
NP, we need more eys to make the review more effective

pkg/istio-agent/agent.go
hzxuzhonghu: changed to use MaxElapsedTime to be consistent with other places

pkg/backoff/retry.go
hzxuzhonghu: Looks a good suggestion

security/pkg/nodeagent/cache/secretcache.go
hzxuzhonghu: We donot need it, seems very hacky

security/pkg/nodeagent/sds/sdsservice.go
hzxuzhonghu: Forever retry is meaningless, this is actually pre generate to cache it. These logic can be refactor, maybe I can leave it for future

hzxuzhonghu:
This is a good place to find many arbitrary retry logic, we need to revisit.  In my view, some are abusing.

pkg/istio-agent/agent.go
```diff
@@ -306,8 +306,9 @@ func (a *Agent) initializeEnvoyAgent(ctx context.Context) error {
 		a.dynamicBootstrapWaitCh = make(chan error, 1)
 		// Simulate an xDS request for a bootstrap
 		// wait indefinitely and keep retrying with jittered exponential backoff
-		b := backoff.NewExponentialBackOff()
-		b.MaxElapsedTime = 30 * time.Second
+		b := backoff.NewExponentialBackOff(func(off *backoff.ExponentialBackOff) {
```
ramaraochavali, line 309: Do we have use cases for this init funcs rather than just setting the fields directly. Looks like mostly used fields are initInterval and MaxElapsedTime . Any benefit of this init funcs?

pkg/queue/instance.go
```diff
@@ -29,7 +29,7 @@ type Task func() error
 
 type BackoffTask struct {
```
ramaraochavali, line 30: I think this is no longer needed?

pkg/queue/instance_test.go
```diff
@@ -108,33 +107,6 @@ func TestRetry(t *testing.T) {
 	wg.Wait()
 }
 
-func TestRetryWithBackoff(t *testing.T) {
```
ramaraochavali, line 111: Can we move this test to exponential_test and add a test case for RetryWithContext as that is the replacement for this?

pkg/queue/instance.go
hzxuzhonghu: right

pkg/queue/instance_test.go
hzxuzhonghu: seems ok

pkg/istio-agent/agent.go
hzxuzhonghu: some others are for unit test

douglas-reid (approved):
p&t owned files LGTM.

pkg/backoff/exponential.go
```diff
@@ -38,22 +45,53 @@ type ExponentialBackOff struct {
...
 	return b
 }
 
 const MaxDuration = 1<<63 - 1
 
-func (b *ExponentialBackOff) NextBackOff() time.Duration {
+func (b ExponentialBackOff) NextBackOff() time.Duration {
 	duration := b.ExponentialBackOff.NextBackOff()
 	if duration == b.Stop {
 		return MaxDuration
```
howardjohn, line 65: This seems very unsafe.. callers will `sleep(forever)` when this happens.

IMO all backoffs either need to explicitly fail when reaching max time or just be capped at max interval

hzxuzhonghu:
@howardjohn Look at the latest commit

hzxuzhonghu:
@howardjohn

pkg/backoff/exponential.go
hzxuzhonghu: fixed

pkg/istio-agent/agent.go
```diff
@@ -450,26 +451,31 @@ func (a *Agent) initSdsServer() error {
...
+// nolint: unparam
 func (a *Agent) getWorkloadCerts(st *cache.SecretManagerClient) (sk *security.SecretItem, err error) {
-	b := backoff.NewExponentialBackOff()
-	b.MaxElapsedTime = 0
-
-	for {
+	b := backoff.NewExponentialBackOff(func(off *backoff.ExponentialBackOff) {
+		off.MaxElapsedTime = 10 * time.Second
+	})
+	err = b.RetryWithContext(context.TODO(), func() error {
```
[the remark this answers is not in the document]
hzxuzhonghu: done

pkg/backoff/exponential.go
```diff
@@ -38,22 +45,51 @@ type ExponentialBackOff struct {
...
+// failed operation returns.
+func (b ExponentialBackOff) RetryWithContext(ctx context.Context, operation func() error) error {
+	b.Reset()
+	for {
+		err := operation()
+		if err == nil {
+			return nil
+		}
+		next := b.ExponentialBackOff.NextBackOff()
+		if next == b.Stop {
```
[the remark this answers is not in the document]
howardjohn: Wow it can. `b.ExponentialBackOff.NextBackOff()` vs `b.NextBackOff()` are different. That seems confusing..

I wonder if we should stop exposing the internal library at all. We also have the issue that MaxElapsedTime is now the same as MaxInterval with the recent changes. It may make more sense to expose a minimal public interface allowing only MaxInterval,Multiplier, and InitialInterval?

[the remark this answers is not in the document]
howardjohn: Or if not, we should at least comment we are explicitly using the internal NextBackOff not the external one so that we can get b.Stop?

pkg/backoff/exponential.go
```diff
@@ -38,22 +45,51 @@ type ExponentialBackOff struct {
...
+		return b.ExponentialBackOff.MaxInterval
 	}
 	return duration
 }
+
+func (b ExponentialBackOff) Reset() {
+	b.ExponentialBackOff.Reset()
+}
+
+// RetryWithContext tries the operation until it does not return error, BackOff stops,
```
howardjohn, line 72:
```suggestion
// RetryWithContext tries the operation until it does not return error, BackOff hits MaxElapsedTime,
```

To meet current implementation.

however, IMO we should not expose MaxElapsedTime at all to users, and instead have them use only the context. 2 benefits:
* only 1 way to cancel the retry instead of 2
* only 1 way to set max internal instead of 2

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -323,18 +323,16 @@ func (sc *SecretManagerClient) addFileWatcher(file string, resourceName string)
 	if err := sc.tryAddFileWatcher(file, resourceName); err == nil {
 		return
 	}
-	// Retry file watcher as some times it might fail to add and we will miss change
+	// RetryWithContext file watcher as some times it might fail to add and we will miss change
```
howardjohn, line 326: nit: this shouldn't change

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -323,18 +323,16 @@ func (sc *SecretManagerClient) addFileWatcher(file string, resourceName string)
...
 	// reasonable threshold (when the problem is not transient) and restart the pod.
 	go func() {
 		b := backoff.NewExponentialBackOff()
-		for {
-			if err := sc.tryAddFileWatcher(file, resourceName); err == nil {
-				break
-			}
-			time.Sleep(b.NextBackOff())
-		}
+		_ = b.RetryWithContext(context.TODO(), func() error {
```
howardjohn, line 332: There are a lot of usages of `RetryWithContext` without a context.

I would recommend we make Retry and RetryContext (no With, to match stdlib usage)

Then this whole thing could be

```go
backoff.NewExponentialBackOff().Retry(func() error { return sc.tryAddFileWatcher(file, resourceName) })
```

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -414,16 +412,15 @@ func (sc *SecretManagerClient) generateRootCertFromExistingFile(rootCertPath, re
 func (sc *SecretManagerClient) generateKeyCertFromExistingFiles(certChainPath, keyPath, resourceName string) (*security.SecretItem, error) {
 	// There is a remote possibility that key is written and cert is not written yet.
 	// To handle that case, check if cert and key are valid if they are valid then only send to proxy.
-	b := backoff.NewExponentialBackOff()
-	b.InitialInterval = sc.configOptions.FileDebounceDuration
+	b := backoff.NewExponentialBackOff(func(off *backoff.ExponentialBackOff) {
+		off.InitialInterval = sc.configOptions.FileDebounceDuration
+		off.MaxElapsedTime = totalTimeout
```
howardjohn, line 417: I think with recent changes MaxElapsedTime is no longer a timeout. We also didn't have a timeout before... do we want to be adding one here?

pkg/backoff/exponential.go
hzxuzhonghu: This is because we are using backoff differently, NextBackOff in some places, retry in some other places

security/pkg/nodeagent/cache/secretcache.go
hzxuzhonghu: Seems not needed, as this is a comment for what the below does

security/pkg/nodeagent/cache/secretcache.go
hzxuzhonghu: IMO, this will make things more complex. If you look at the k8s, previously the client-go does not have `Context` at all.  Just at a release, community introduced it.

security/pkg/nodeagent/cache/secretcache.go
hzxuzhonghu: yes

security/pkg/nodeagent/cache/secretcache.go
howardjohn: We should have one way to timeout though, not 2. If that's always context then that's fine with me but we should get rid of MaxElapsedTime

hzxuzhonghu: yes, i removed the maxelapsedtime

security/pkg/nodeagent/cache/secretcache.go
hzxuzhonghu: TestWorkloadAgentGenerateSecretFromFileOverSdsWithBogusFiles failed with no timeout

pkg/backoff/exponential.go
```diff
@@ -15,45 +15,94 @@
...
 	}
 	return duration
 }
+
+func (b ExponentialBackOff) Reset() {
+	b.exponentialBackOff.Reset()
+}
+
+// RetryWithContext tries the operation until it does not return error,
+// or when the context expires, whichever happens first.
```
howardjohn, line 86: currently it also stops when hitting maxInterval I think?

pkg/istio-agent/bootstrap_stream.go
```diff
@@ -58,11 +58,7 @@ func (b *bootstrapDiscoveryStream) Send(resp *discovery.DiscoveryResponse) error
 			sendToChannelWithoutBlock(b.errCh, fmt.Errorf("failed to update bootstrap from discovery: %v", err))
 			return nil
 		}
-		select {
-		case <-b.errCh:
-		default:
-			close(b.errCh)
-		}
+		close(b.errCh)
```
howardjohn, line 61: I think this was trying to avoid double close -> panic. Not sure if that is possible though... as a followup this code may need more tests since it currently (not from this PR) has some issues in error handling

howardjohn (approved):
Thanks for all the improvements here!

pkg/backoff/exponential.go
hzxuzhonghu: Nope, actually  no way to set maxElapsedTime now. I think this can be removed

PR #40644, opened 2022-08-25
Part of revert #40631, not to hide rate limited info, it is an import…

hzxuzhonghu:
…ant signal to users

As described, rate limited info is an important sign to tell operator to tune istiod paramaters

### DIFF
--- a/pkg/istio-agent/xds_proxy.go
+++ b/pkg/istio-agent/xds_proxy.go
@@ -41,7 +41,6 @@ import (
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/reflection"
-	"google.golang.org/grpc/status"
 	anypb "google.golang.org/protobuf/types/known/anypb"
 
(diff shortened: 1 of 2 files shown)


hzxuzhonghu:
assign @ramaraochavali

ramaraochavali:
I think it is difficult to look at every xds proxy logs. Logs from Istiod side would be more useful for operators to look at. Even better stats.

hzxuzhonghu:
From xds proxy, we need to figure out why it cannot be served. This capability is super useful for distributed system

hzxuzhonghu:
You could never let a user guess why his app cannot work well

ramaraochavali (approved):
I do not know in production systems with so many envoys running, when **few envoys**  requests fail for rate limited, it would be easy to figure out the proxies failed using proxy logs. It would be much easier to look Istiod logs and see which proxies it has rate limited. 

Having said, that I am OK to approve this, since it is restoring the previous state.

hzxuzhonghu:
yes, as adopters who provide paas with istio, we would want as many as logs if it can help figure out issues.

PR #40650, opened 2022-08-25
add additional test cases for service visibility

ramaraochavali:
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/push_context_test.go
+++ b/pilot/pkg/model/push_context_test.go
@@ -859,6 +859,47 @@ func TestIsServiceVisible(t *testing.T) {
 			},
 			expect: false,
 		},
+		{
+			name:        "service visible to none",
+			pushContext: &PushContext{},
+			service: &Service{
+				Attributes: ServiceAttributes{
+					Namespace: "bar",
+					ExportTo: map[visibility.Instance]bool{
+						visibility.None: true,
+					},
+				},
+			},
+			expect: false,
+		},
+		{
+			name:        "service has both public visibility and none visibility",
+			pushContext: &PushContext{},
+			service: &Service{
+				Attributes: ServiceAttributes{
+					Namespace: "bar",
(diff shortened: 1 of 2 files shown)


pilot/pkg/model/push_context_test.go
```diff
@@ -859,6 +859,47 @@ func TestIsServiceVisible(t *testing.T) {
...
+		},
+		{
+			name:        "service has both public visibility and none visibility",
+			pushContext: &PushContext{},
+			service: &Service{
+				Attributes: ServiceAttributes{
+					Namespace: "bar",
+					ExportTo: map[visibility.Instance]bool{
+						visibility.Public: true,
+						visibility.None:   true,
```
hzxuzhonghu, line 883: I think this is not a right config, maybe we should prevent it ealier

pilot/pkg/model/push_context_test.go
```diff
@@ -859,6 +859,47 @@ func TestIsServiceVisible(t *testing.T) {
...
+		},
+		{
+			name:        "service has both none visibility and private visibility",
+			pushContext: &PushContext{},
+			service: &Service{
+				Attributes: ServiceAttributes{
+					Namespace: "bar",
+					ExportTo: map[visibility.Instance]bool{
+						visibility.Private: true,
+						visibility.None:    true,
```
hzxuzhonghu, line 897: same here not invalid case

pilot/pkg/model/push_context_test.go
ramaraochavali: Current code supports it and the comments are very clear on what to do when this is configured. That is why I added these tests to ensure the order is not messed up in future. I do not know if we can prevent this/change this now for BC reasons.

hzxuzhonghu (approved):
keep the previous behavior

PR #40657, opened 2022-08-25
Fix race-condition with istio-ca-cert creation

kfaseela:
When you deploy multiple controller replicas, all of them will try
to create the istio-ca-secret secret if it is not found on the cluster.
A race condition on "if istio-ca-secret not exists -> create" can
lead to all (or a subset) of the replicas to try to create the secret
but only one of them will succeed. The other replicas will fail as the
secret was already written by the one lucky replica and will fail the
readiness probe as a consequence.

Once the readiness fail, the affected replicas are restarted and then they
will boot successfully as the race-condition for the secret creation is no longer there.
This does not seem to be a very severe issue as all the replicas will be
ready after one restart. On the other end it would be nice if Istio did not
restart if it fails in writing a istio-ca-secret which has been created
in the meantime by someone else.

This is the type of message you will find in the discovery containers before the restart has took place:
```
2022-08-24T14:52:05.495464Z info    pkica   Failed to get secret (error: secrets "istio-ca-secret" not found), will create one
2022-08-24T14:52:10.722863Z error   pkica   Failed to write secret to CA (error: secrets "istio-ca-secret" already exists). Abort.Error: failed to create discovery service: failed to create CA: failed to create a self-signed istiod CA: failed to create CA due to secret write error
```
The issue can be reproduced by the below steps:

LOOP UNTIL RESTART OF ONE REPLICA:
    $helm install istiod istio/istiod -n istio-system --set pilot.replicaCount=2 --set pilot.autoscaleEnabled=false --create-namespace 
    $helm uninstall istiod -n istio-system
    $kubectl delete ns istio-system

### DIFF
--- a/security/pkg/pki/ca/ca.go
+++ b/security/pkg/pki/ca/ca.go
@@ -183,11 +183,15 @@ func NewSelfSignedIstioCAOptions(ctx context.Context,
 		// Write the key/cert back to secret so they will be persistent when CA restarts.
 		secret := k8ssecret.BuildSecret("", CASecret, namespace, nil, nil, nil, pemCert, pemKey, istioCASecretType)
 		if _, err = client.Secrets(namespace).Create(context.TODO(), secret, metav1.CreateOptions{}); err != nil {
-			pkiCaLog.Errorf("Failed to write secret to CA (error: %s). Abort.", err)
-			return nil, fmt.Errorf("failed to create CA due to secret write error")
+			// Ignore error if secret got created by another replica
+			if caSecret, scrtErr = client.Secrets(namespace).Get(context.TODO(), CASecret, metav1.GetOptions{}); scrtErr != nil {
+				pkiCaLog.Errorf("Failed to write secret to CA (error: %s). Abort.", err)
+				return nil, fmt.Errorf("failed to create CA due to secret write error")
+			}
 		}
 		pkiCaLog.Infof("Using self-generated public key: %v", string(rootCerts))
-	} else {
+	}
+	if scrtErr == nil {
 		pkiCaLog.Infof("Load signing key and cert from existing secret %s:%s", caSecret.Namespace, caSecret.Name)
 		rootCerts, err := util.AppendRootCerts(caSecret.Data[CACertFile], rootCertFile)
 		if err != nil {
--- a/security/pkg/pki/ca/ca.go
+++ b/security/pkg/pki/ca/ca.go
@@ -187,9 +187,8 @@ func NewSelfSignedIstioCAOptions(ctx context.Context,
 			if caSecret, scrtErr = client.Secrets(namespace).Get(context.TODO(), CASecret, metav1.GetOptions{}); scrtErr != nil {
 				pkiCaLog.Errorf("Failed to write secret to CA (error: %s). Abort.", err)
 				return nil, fmt.Errorf("failed to create CA due to secret write error")
-			} else {
-				pkiCaLog.Infof("Secret created by another replica, reusing the same.")
 			}
+			pkiCaLog.Infof("Secret created by another replica, reusing the same.")
 		} else {
 			pkiCaLog.Infof("Using self-generated public key: %v", string(rootCerts))
 		}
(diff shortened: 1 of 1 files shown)


security/pkg/pki/ca/ca.go
```diff
@@ -183,11 +183,15 @@ func NewSelfSignedIstioCAOptions(ctx context.Context,
 		// Write the key/cert back to secret so they will be persistent when CA restarts.
 		secret := k8ssecret.BuildSecret("", CASecret, namespace, nil, nil, nil, pemCert, pemKey, istioCASecretType)
 		if _, err = client.Secrets(namespace).Create(context.TODO(), secret, metav1.CreateOptions{}); err != nil {
-			pkiCaLog.Errorf("Failed to write secret to CA (error: %s). Abort.", err)
-			return nil, fmt.Errorf("failed to create CA due to secret write error")
+			// Ignore error if secret got created by another replica
```
howardjohn, line 186: won't we have the wrong data though? we are storing or data at L200 but its not the right key

lei-tang: The PR may cause inconsistence between the CA key and CA bundle. The following is an example sequence of how the inconsistence occurs.
- Controller A and B both try to create a CA secret.
- A succeeds and B fails at creating a CA secret.
- B reads the secret created by A successfully and proceed to line 200 to write B's CA bundle.
- B writes CA bundle successfully, which is inconsistent with the CA secret created by A.

lei-tang (changes requested):

security/pkg/pki/ca/ca.go
kfaseela: @lei-tang : step number 4 in your sequence - line 187 is overwriting the caSecret with the data read from A in case of conflict, and that will be used at line 200, won't that help?

lei-tang: Another inconsistency under such scenario is the rootCerts on line 192 is B's rootCerts whereas the CA secret is A's CA secret.
line 192  pkiCaLog.Infof("Using self-generated public key: %v", string(rootCerts))

lei-tang: The line 192 uses rootCerts from B at line 174, whereas at line 196, rootCerts is read from CA secret generated by A.

security/pkg/pki/ca/ca.go
```diff
@@ -183,11 +183,15 @@ func NewSelfSignedIstioCAOptions(ctx context.Context,
...
+			if caSecret, scrtErr = client.Secrets(namespace).Get(context.TODO(), CASecret, metav1.GetOptions{}); scrtErr != nil {
```
lei-tang, line 187: One replica fails at writing a CA secret does not necessarily mean the Get(CASecret) will succeed (e.g., it may take a while for the CA secret writing from another replica to complete). Thus, the Get(CASecret) here may not always work.

security/pkg/pki/ca/ca.go
kfaseela: ack about the logging inconsistency

security/pkg/pki/ca/ca.go
kfaseela: Would we get write failures for "already exists", when the other write was in progress? Or only once completed? If so, checking for specific error "secret already exists" and doing a read only in that case helps?

lei-tang: It is hard to verify the correctness of code behavior under concurrent writes/reads. 

The existing implementation is simple and easy to analyze its correctness, which returns and retries if writing CA secret fails.I think the behavior of existing implementation is fine: the error log is harmless and may be improved by changing the log type from "error" to "info".

The PR has a more complicated implementation than the existing implementation and its behavior under concurrent writes/reads is more difficult to verify. Moreover, the existing implementation is stable and battle-tested.

kfaseela: @lei-tang Understand your concern. I have anyways updated the PR adding additional logs, in case that makes the new implementation easier to debug in case of failures. Also fixed the logging inconsistency. Ofcourse retries are already there, which makes the problem less severe, but just that we can avoid an unnecessary retry.

PR #40658, opened 2022-08-25
improve release/downloadIstioCandidate

my-git9:
improve release/downloadIstioCandidate


ericvn:
I would also update downloadIstioCtl.sh

ericvn:
Local testing with commands like `ISTIO_VERSION=1.11.0 TARGET_ARCH=armv8 ./testing.sh` (I had copied the file to testing.sh) work as expected. 

We need to be careful of these files as they back the istio.io URLs.

ericvn (approved):
Thanks for the contribution.

PR #40660, opened 2022-08-25
session affinity support for proxyless gRPC, EDS draining

costinm:
Adding session affinity for gRPC - currently using an annotation on Service.

We may define an API - or include this as a Gateway/GAMMA policy. 

As prior art, Service has a sessionAffinity  field already - but only support ClientIP, not cookie.
Ingress is using an annotation (for Nginx). CloudRun has a Service annotation as well. 
GCP Gateway is using BackendService CRD. 

The equivalent change in gRPC is still in progress - and they may define the annotation or
API ( GRPCRoute ?). For now this PR allows generating the filter that they expect.


ramaraochavali:
@costinm Many people (including us) are looking for this feature. Let us just add proper API. I have created this PR https://github.com/istio/api/pull/2469. Please review it. I am happy to implement this in Istio once the API is merged.

costinm:
@sanjaypujare 

This is WIP to unblock development on gRPC side - when a proper API is defined we can switch it.

Note that Nginx and few other are using annotations for their 'sticky session' policies - and likely this is better implemented
as a new CRD implementing Gateway API with HttpRoute attachment. I mentioned in the other PR that having it in DestinationRule is probably wrong - the 'path' is very important.

sanjaypujare:
> @costinm Many people (including us) are looking for this feature. Let us just add proper API. I have created this PR [istio/api#2469](https://github.com/istio/api/pull/2469). Please review it. I am happy to implement this in Istio once the API is merged.

@ramaraochavali Doing this PR doesn't impact anything else as far as I can see. [istio/api#2469](https://github.com/istio/api/pull/2469) could take longer but we need this one soon so it will be good to get this merged. Also I haven't checked if the other PR does exactly the same thing as this one.

pilot/pkg/networking/grpcgen/lds.go
```diff
@@ -281,6 +285,28 @@ func buildOutboundListeners(node *model.Proxy, push *model.PushContext, filter l
...
+					filters = append(filters, &hcm.HttpFilter{
+						Name: "envoy.filters.http.stateful_session", // TODO: wellknown.
+						ConfigType: &hcm.HttpFilter_TypedConfig{
+							TypedConfig: protoconv.MessageToAny(&stateful_sessionv3.StatefulSession{
+								SessionState: &core.TypedExtensionConfig{
+									Name: "envoy.http.stateful_session.cookie",
+									TypedConfig: protoconv.MessageToAny(&cookiev3.CookieBasedSessionState{
+										Cookie: &httpv3.Cookie{
+											Path: "/",
+											Ttl:  &durationpb.Duration{Seconds: 120},
```
sanjaypujare, line 300: So these (`Path` and `Ttl`) are fixed and not configurable from the control plane API? For a temporary solution this might be okay but wondering if it's easy to get the value from attrs similar to the cookie name (that we get from `sv.Attributes.Labels["x-session-cookie"]`)

pilot/pkg/networking/grpcgen/lds.go
```diff
@@ -281,6 +285,28 @@ func buildOutboundListeners(node *model.Proxy, push *model.PushContext, filter l
...
+				filters := supportedFilters
+				sessionCookie := sv.Attributes.Labels["x-session-cookie"]
+				if sessionCookie != "" {
+					filters = append(filters, &hcm.HttpFilter{
+						Name: "envoy.filters.http.stateful_session", // TODO: wellknown.
+						ConfigType: &hcm.HttpFilter_TypedConfig{
+							TypedConfig: protoconv.MessageToAny(&stateful_sessionv3.StatefulSession{
+								SessionState: &core.TypedExtensionConfig{
+									Name: "envoy.http.stateful_session.cookie",
+									TypedConfig: protoconv.MessageToAny(&cookiev3.CookieBasedSessionState{
```
sanjaypujare, line 297: Regarding `stateful_sessionv3.StatefulSession` and `cookiev3.CookieBasedSessionState` : I don't know how they translate to `@type`  `type.googleapis.com/envoy.extensions.filters.http.stateful_session.v3.StatefulSession` and `type.googleapis.com/envoy.extensions.http.stateful_session.cookie.v3.CookieBasedSessionState` strings respectively. May be it's all right but just pointing it out

sanjaypujare:
> @sanjaypujare
> 
> This is WIP to unblock development on gRPC side - when a proper API is defined we can switch it.
> 
> Note that Nginx and few other are using annotations for their 'sticky session' policies - and likely this is better implemented as a new CRD implementing Gateway API with HttpRoute attachment. I mentioned in the other PR that having it in DestinationRule is probably wrong - the 'path' is very important.

Looks good to me. It will be good to get this merged regardless of the other PR (which uses CRD etc)

pilot/pkg/networking/grpcgen/lds.go
costinm: TTL - we can make it configurable. 

The Path is VERY complicated - sticky sessions are also very common with end users ( browsers and gateway ), and even in the mesh one destination (service) may be the backend for multiple routes. 

If we were attaching this to a HTTPRoute or GRPCRoute - the path would be automatically populated from the route matching rule, we wouldn't want to allow user to specify path "/foo" if the service is routed as "/bar". 

I would defer the Path for a future change.

ramaraochavali:
> Doing this PR doesn't impact anything else as far as I can see. https://github.com/istio/api/pull/2469 could take longer but we need this one soon so it will be good to get this merged. Also I haven't checked if the other PR does exactly the same thing as this one.

Sure. The other PR is aimed at doing the same thing. But if you are in hurry, we can move forward with this and once it is merged, we can change this implementation.

pilot/pkg/networking/grpcgen/lds.go
costinm: AFAIK MessageToAny will set the type automatically, at least they do in other cases.

sanjaypujare:
> > Doing this PR doesn't impact anything else as far as I can see. [istio/api#2469](https://github.com/istio/api/pull/2469) could take longer but we need this one soon so it will be good to get this merged. Also I haven't checked if the other PR does exactly the same thing as this one.
> 
> Sure. The other PR is aimed at doing the same thing. But if you are in hurry, we can move forward with this and once it is merged, we can change this implementation.

Thanks @ramaraochavali . Would you mind approving the PR (after reviewing), so that we can get unblocked for the specific scenario we have?

pilot/pkg/networking/grpcgen/lds.go
costinm: What would be a reasonable default ? 1 h ? And should I add an annotation as well, or leave it to the final API ? 

120s is a bit too short.

session-cookie-max-age ( mirroring nginx annotation ) ? I see traefik doesn't have a setting for this, tomcat session timeout default was 30 min.

sanjaypujare: > What would be a reasonable default ? 1 h ?

Sounds good to me

>  And should I add an annotation as well, or leave it to the final API ?

Annotation will be good.

> 
> 120s is a bit too short.
> 
> session-cookie-max-age ( mirroring nginx annotation ) ?

If that's the name/annotation you are suggesting, sounds good to me.

costinm:
Updated with draining support and customizable label name.

Noticed that we don't track service attributes - it seems Gateway impl also uses a service label for an experimental feature, so sticking with labels until a proper API is defined. 

Adding the filter to envoy is not part of this PR, should be added in a separate PR - currently users add it using EnvoyFilter, which is not supported in proxyless gRPC. However envoy users will also get draining endpoints for persistent sessions.
( adding requires passing the service trough - so few method signature refactoring, better in separate change)

pilot/pkg/features/pilot.go
```diff
@@ -113,6 +113,18 @@ var (
 			" to unhealthy/non-ready hosts even if the percentage of healthy hosts fall below minimum health percentage(panic threshold).",
 	).Get())
 
+	PersistentSessionLabel = atomic.NewString(env.Register(
+		"PILOT_PERSISTENT_SESSION_LABEL",
+		"persistent-session",
```
howardjohn, line 118: As discussed offline:

* label needs some dns name. Probably istio.io is easiest
* We need to verify we don't push on DRAINING (so it is zero cost). I think that is the case but we should verify. I will pont to the existing test we can expand tomorrow morning

costinm: Fixed the names, and added an extra safety - if persistent session label is disabled we no longer track draining ( so no change in behavior if we also don't track unhealthy ).

howardjohn (approved):

howardjohn:
would be nice to add this test in eds_test.go (mostly copied from the unhealthy one):

```go

func TestEDSDrainingEndpoints(t *testing.T) {
	s := xds.NewFakeDiscoveryServer(t, xds.FakeOptions{})
	s.Discovery.MemRegistry.AddService(&model.Service{
		Hostname: "unhealthy.svc.cluster.local",
		Ports: model.PortList{
			{
				Name:     "tcp-dns",
				Port:     53,
				Protocol: protocol.TCP,
			},
		},
	})
	fullPush(s)
	adscon := s.Connect(nil, nil, watchEds)
	_, err := adscon.Wait(5 * time.Second)
	if err != nil {
		t.Fatalf("Error in push %v", err)
	}

	validateEndpoints := func(expectPush bool, healthy []string, unhealthy []string) {
		t.Helper()
		// Normalize lists to make comparison easier
		if healthy == nil {
			healthy = []string{}
		}
		if unhealthy == nil {
			unhealthy = []string{}
		}
		sort.Strings(healthy)
		sort.Strings(unhealthy)
		if expectPush {
			upd, _ := adscon.Wait(5*time.Second, v3.EndpointType)

			if len(upd) > 0 && !contains(upd, v3.EndpointType) {
				t.Fatalf("Expecting EDS push as endpoint health is changed. But received %v", upd)
			}
		} else {
			upd, _ := adscon.Wait(50*time.Millisecond, v3.EndpointType)
			if contains(upd, v3.EndpointType) {
				t.Fatalf("Expected no EDS push, got %v", upd)
			}
		}

		// Validate that endpoints are pushed.
		lbe := adscon.GetEndpoints()["outbound|53||unhealthy.svc.cluster.local"]
		eh, euh := xdstest.ExtractHealthEndpoints(lbe)
		gotHealthy := sets.New(eh...).SortedList()
		gotUnhealthy := sets.New(euh...).SortedList()
		if !reflect.DeepEqual(gotHealthy, healthy) {
			t.Fatalf("did not get expected endpoints: got %v, want %v", gotHealthy, healthy)
		}
		if !reflect.DeepEqual(gotUnhealthy, unhealthy) {
			t.Fatalf("did not get expected unhealthy endpoints: got %v, want %v", gotUnhealthy, unhealthy)
		}
	}

	// Validate that we do not send initial unhealthy endpoints.
	validateEndpoints(false, nil, nil)
	adscon.WaitClear()

	// Change the status of endpoint to Healthy and validate Eds is pushed.
	s.Discovery.MemRegistry.SetEndpoints("unhealthy.svc.cluster.local", "",
		[]*model.IstioEndpoint{
			{
				Address:         "10.0.0.53",
				EndpointPort:    53,
				ServicePortName: "tcp-dns",
				HealthStatus:    model.Healthy,
			},
		})

	// Validate that endpoints are pushed.
	validateEndpoints(true, []string{"10.0.0.53:53"}, nil)

	t.Log("drain!")

	// Now change the status of endpoint to draining and validate Eds is not pushed.
	s.Discovery.MemRegistry.SetEndpoints("unhealthy.svc.cluster.local", "",
		[]*model.IstioEndpoint{
			{
				Address:         "10.0.0.53",
				EndpointPort:    53,
				ServicePortName: "tcp-dns",
				HealthStatus:    model.Draining,
			},
		})
	validateEndpoints(true, nil, nil)
	s.Discovery.MemRegistry.SetEndpoints("unhealthy.svc.cluster.local", "", []*model.IstioEndpoint{})
	validateEndpoints(false, nil, nil)
}
```

or similar, to ensure the pushing works as expected (no extra push

howardjohn:
Actually that test fails - might be a real bug?

I had tested in a real cluster and didn't see extra push though, so maybe its more of an issue with the test expectations...

pilot/pkg/serviceregistry/kube/controller/endpointslice.go
```diff
@@ -231,9 +231,17 @@ func (esc *endpointSliceController) updateEndpointCacheForSlice(hostName host.Na
 	discoverabilityPolicy := esc.c.exports.EndpointDiscoverabilityPolicy(esc.c.GetService(hostName))
 
 	for _, e := range slice.Endpoints() {
+		// Draining tracking is only enabled if persistent sessions is enabled.
+		// If we start using them for other features, this can be adjusted.
+		draining := features.PersistentSessionLabel.String() != "" &&
+				e.Conditions.Ready != nil &&
+				e.Conditions.Serving != nil &&
+				*e.Conditions.Serving &&
```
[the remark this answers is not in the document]
costinm: Agreed. To be fair - they kept backward compatibility, I am also confused...

costinm:
The test fails because it calls SetEndpoints which trigger a push. I think what prevents the push is that EndpointSlice controller is skipping it if there is no change. 

I added an extra check in the ES controller, will only track draining if the service label is also set.

Not sure what is detecting if the slice is identical to previously pushed one - but AFAIK for all other services the output
of the slice method will be identical - the 'draining' bit is only true if the service has the label and the special status.

PR #40661, opened 2022-08-25
manifests: add variant option to select distroless

howardjohn:
We currently have this in ProxyConfig but that only applies for
injection. Add a new field to control everything. This also sets the
ProxyConfig field for convenience.

Commit 1 is the real changes, commit 2 is `make gen`


costinm (approved):
Sounds reasonable and consistent with what we have been doing. 

I'm not very happy that we're adding one more values.yaml field that is install-only, at least MeshConfig is reloaded and can be changed without having to reinstall. 

However not sure if moving the variant to mesh config or proxy config is best either, and we can do it later for all image-related things.

howardjohn:
This is inherently install time IMO since its configuring the actual image
of Istiod - that isn't something we can control at runtime. For the proxy
it is still runtime config though

ANGkeith:
hi, I am planning to use distroless image like that
```yaml
global:
  tag: '1.14.3-distroless'
```

may i know what is the diff with using new `variant` values ?

howardjohn:
You don't need to hardcode the tag, which would change every version. Can
just set variant=distroless

PR #40667, opened 2022-08-26
fix service export 'None' in config namespace with vs destination

ramaraochavali:
This https://github.com/istio/istio/pull/31808 fixed it for other namespaces. But if we have exportTo 'None' in configNamespace, the service is still used.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/sidecar.go
+++ b/pilot/pkg/model/sidecar.go
@@ -357,7 +357,7 @@ func ConvertToSidecarScope(ps *PushContext, sidecarConfig *config.Config, config
 
 			for h, ports := range virtualServiceDestinations(v) {
 				// Default to this hostname in our config namespace
-				if s, ok := ps.ServiceIndex.HostnameAndNamespace[host.Name(h)][configNamespace]; ok {
+				if s, ok := ps.ServiceIndex.HostnameAndNamespace[host.Name(h)][configNamespace]; ok && ps.IsServiceVisible(s, configNamespace) {
 					// This won't overwrite hostnames that have already been found eg because they were requested in hosts
 					var vss *Service
 					if matchPort {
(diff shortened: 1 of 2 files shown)


pilot/pkg/model/sidecar.go
```diff
@@ -357,7 +357,7 @@ func ConvertToSidecarScope(ps *PushContext, sidecarConfig *config.Config, config
 
 			for h, ports := range virtualServiceDestinations(v) {
 				// Default to this hostname in our config namespace
-				if s, ok := ps.ServiceIndex.HostnameAndNamespace[host.Name(h)][configNamespace]; ok {
+				if s, ok := ps.ServiceIndex.HostnameAndNamespace[host.Name(h)][configNamespace]; ok && ps.IsServiceVisible(s, configNamespace) {
```
hzxuzhonghu, line 360: I think this may violate original intention, it meant to infer svc from virtual service destination, actually these services are not exported to the sidecar

ramaraochavali: But we do the same filtering in else block

ramaraochavali: @hzxuzhonghu can you PTAL when you get chance?

hzxuzhonghu: This is unneccessary, service export to `NONE` will not be in SidecarIndex at all.

And here the service is in same namespace as proxy configNamespace, so it must be visible to proxy even it is private.

ramaraochavali: These are services determined by virtual service destination not services in sidecarIndex.  See https://github.com/istio/istio/blob/master/pilot/pkg/model/sidecar.go#L385 - It does similar check. I think we need same check in if block. BTW, we ran in to this issue recently. Service with export to `~` has its clusters and routes created.

PR #40668, opened 2022-08-26
do not complain about conflict if service instance ports are same

ramaraochavali:
Service Instance creation does not consider visibility rules of service.

We have a two services in the same namespace, one is the actual service and other is created for [external ips](https://kubernetes.io/docs/concepts/services-networking/service/#external-ips) using same selector for providing access outside of cluster by some other tools. In this case, Istio complains of inbound listener conflict because it sees two service instances.

This PR reuses the same service instance port (existing behaviour) instead of complaining (new behaviour - skips reporting conflict in this case).

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/listener_inbound.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener_inbound.go
@@ -143,6 +143,22 @@ type ServiceInstancePort struct {
 	Protocol protocol.Instance `json:"protocol,omitempty"`
 }
 
+func (sip *ServiceInstancePort) equals(other *ServiceInstancePort) bool {
+	if sip == nil && other == nil {
+		return true
+	}
+	if sip.Port != other.Port {
+		return false
+	}
+	if sip.TargetPort != other.TargetPort {
+		return false
+	}
+	if sip.Protocol != other.Protocol {
+		return false
+	}
+	return true
+}
+
 // buildInboundListeners creates inbound listeners.
 // Typically, this a single listener with many filter chains for each applicable Service; traffic is redirect with iptables.
 // However, explicit listeners can be used in NONE mode or with Sidecar.Ingress configuration.
@@ -287,7 +303,11 @@ func (lb *ListenerBuilder) buildInboundChainConfigs() []inboundChainConfig {
 			// The Service is *almost* not relevant, but some Telemetry is per-service.
 			// If there is a conflict, we will use the oldest Service. This impacts the protocol used as well.
 			if old, f := chainsByPort[port.TargetPort]; f {
-				reportInboundConflict(lb, old, cc)
+				if !port.equals(&old.port) {
+					// Report conflict if service Instance port is not equal - otherwise use the
+					// previous port.
+					reportInboundConflict(lb, old, cc)
+				}
 				continue
 			}
 			chainsByPort[port.TargetPort] = cc
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -1317,6 +1317,11 @@ func (c *Controller) getProxyServiceInstancesByPod(pod *v1.Pod,
 	var out []*model.ServiceInstance
 
 	for _, svc := range c.servicesForNamespacedName(kube.NamespacedNameForK8sObject(service)) {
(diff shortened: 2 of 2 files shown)


ramaraochavali:
@howardjohn @hzxuzhonghu Does this make sense?

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -1317,6 +1317,11 @@ func (c *Controller) getProxyServiceInstancesByPod(pod *v1.Pod,
 	var out []*model.ServiceInstance
 
 	for _, svc := range c.servicesForNamespacedName(kube.NamespacedNameForK8sObject(service)) {
+		// Check if service is visible. Service can be invisible if exportTo is set to '~' for example.
+		// We should consider service instances that are only visible to pod.
+		if proxy.LastPushContext != nil && !proxy.LastPushContext.IsServiceVisible(svc, pod.Namespace) {
```
hzxuzhonghu, line 1322: unnecessary, the pod itself is the service instance

ramaraochavali: see my description of the PR. without this, we get inbound listener conflict if an there are multiple service select same pods. Any way to prevent that?

hzxuzhonghu: I do not think it is a reasonable case

hzxuzhonghu: Say you define two services for a proxy, and you do not want to use one.

ramaraochavali: But we want to explicitly say we do not want to use it by specifying "networking.istio.io": "~" on the service that is not meant to be used for this purpose.

Thinking aloud, may be we we should also merge service instances by port to prevent this inbound conflicts?

ramaraochavali: Changed it to not complain if service instance ports are equal

ramaraochavali:
@hzxuzhonghu @howardjohn when you get chance PTAL. Not urgent

pilot/pkg/networking/core/v1alpha3/listener_inbound.go
```diff
@@ -287,7 +303,11 @@ func (lb *ListenerBuilder) buildInboundChainConfigs() []inboundChainConfig {
 			// The Service is *almost* not relevant, but some Telemetry is per-service.
 			// If there is a conflict, we will use the oldest Service. This impacts the protocol used as well.
 			if old, f := chainsByPort[port.TargetPort]; f {
-				reportInboundConflict(lb, old, cc)
+				if !port.equals(&old.port) {
+					// Report conflict if service Instance port is not equal - otherwise use the
+					// previous port.
+					reportInboundConflict(lb, old, cc)
```
hzxuzhonghu, line 309: Even the ports are equal, the telemetry data conflicts for two services.

hzxuzhonghu (commented):
cc @istio/wg-policies-and-telemetry-maintainers

ramaraochavali:
I do not understand why telemetry would have any impact with this change. All we are doing here is not to complain (increment conflict listeners) ,rest of the code is same.

hzxuzhonghu:
This is changing the complain condition, like service name, which is used by telemetry.

hzxuzhonghu:
FYI https://github.com/istio/istio/pull/40668/files#diff-4afa570feba990275e28fa8ca0b5be122e9ff5d6f0beee9d2e8f54cb57d46613R292

ramaraochavali:
>. This is changing the complain condition, like service name, which is used by telemetry.

Do you mean conflicting listener metric or some thing else?

hzxuzhonghu:
> It is just that we do not increment conflicting listener metric now if ports are same.

Yes, it's this one

ramaraochavali:
@istio/wg-policies-and-telemetry-maintainers any concerns with this?

ramaraochavali:
@istio/wg-policies-and-telemetry-maintainers any concerns with this impacting telemetry?

zirain:
I would like to keep previous behavior, that warn user of duplicate service created.

ramaraochavali:
Sure. fair enough

PR #40687, opened 2022-08-26
Move Fake test helpers out of core packages

howardjohn:
Currently, we have our test fakes in our real core packages. This
introduces unnecesary bloat into production binaries - a non trivial
amount in some cases; with some series of changes including this one we
could drop 5mb of the agent binary size.

However, we cannot simply move this to another package. We have tests in
`v1alpha3` that depend on the Fake, and the Fake depends on `v1alpha3` -
we get a cycle. The simple fix would be to move all tests to
`v1alpha3_test` package, but that isn't viable.

Instead, we implement
https://www.cockroachlabs.com/blog/outsmarting-go-dependencies-testing-code/


ingwonsong:
I like this approach. I expect that we can remove many unnecessary dependencies, too.

I know this is in WIP, but I felt that we need to make some directory structures for organizing fakes.

PR #40688, opened 2022-08-27
optimize default sidecar mutex

ramaraochavali:
This PR does the following
- Optimize default sidecar mutex. We need mostly read locks except for initialization. Changes it to RWMutex and uses read locks most of the time
- Change SideCarIndex and have rootConfigSidecarByNamespace and defaultSidecarByNamespace so that we do not have gateway specific fields in the struct.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -131,25 +131,26 @@ type sidecarIndex struct {
 	// user configured sidecars for each namespace if available.
 	sidecarsByNamespace map[string][]*SidecarScope
 	// the Sidecar for the root namespace (if present). This applies to any namespace without its own Sidecar.
-	rootConfig *config.Config
-	// rootConfigSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar.
-	// These are converted from rootConfig if it exists.
+	meshRootSidecarConfig *config.Config
+	// meshRootSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar.
+	// These are converted from root namespace sidecar if it exists.
 	// These are lazy-loaded. Access protected by defaultSidecarMu
-	rootConfigSidecarsByNamespace map[string]*SidecarScope
-	// defaultSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar,
-	// These are *always* computed from DefaultSidecarScopeForNamespace.
+	meshRootSidecarsByNamespace map[string]*SidecarScope
+	// catchAllSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar,
+	// These are *always* computed from DefaultSidecarScopeForNamespace i.e. a sidecar that has listeners
+	// for all services in the mesh. This will be used if there is no sidecar specified in root namespace.
 	// These are lazy-loaded. Access protected by defaultSidecarMu
-	defaultSidecarsByNamespace map[string]*SidecarScope
-	// mutex to protect default sidecars.
-	defaultSidecarMu *sync.RWMutex
+	catchAllSidecarsByNamespace map[string]*SidecarScope
+	// mutex to protect derived sidecars i.e. not specified by user.
+	derivedSidecarMu *sync.RWMutex
 }
 
 func newSidecarIndex() sidecarIndex {
 	return sidecarIndex{
-		sidecarsByNamespace:           map[string][]*SidecarScope{},
-		rootConfigSidecarsByNamespace: map[string]*SidecarScope{},
-		defaultSidecarsByNamespace:    map[string]*SidecarScope{},
-		defaultSidecarMu:              &sync.RWMutex{},
+		sidecarsByNamespace:         map[string][]*SidecarScope{},
+		meshRootSidecarsByNamespace: map[string]*SidecarScope{},
+		catchAllSidecarsByNamespace: map[string]*SidecarScope{},
+		derivedSidecarMu:            &sync.RWMutex{},
 	}
 }
 
@@ -970,10 +971,10 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Insta
 	var computed *SidecarScope
 	switch proxy.Type {
 	case Router:
-		ps.sidecarIndex.defaultSidecarMu.RLock()
+		ps.sidecarIndex.derivedSidecarMu.RLock()
 		// Gateways always use default sidecar scope.
-		sc, f := ps.sidecarIndex.defaultSidecarsByNamespace[proxy.ConfigNamespace]
-		ps.sidecarIndex.defaultSidecarMu.RUnlock()
+		sc, f := ps.sidecarIndex.catchAllSidecarsByNamespace[proxy.ConfigNamespace]
+		ps.sidecarIndex.derivedSidecarMu.RUnlock()
 		if f {
 			// We have already computed the scope for this namespace, just return it.
 			return sc
@@ -981,15 +982,15 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Insta
 		// We need to compute this namespace
 		computed = ps.computeSidecarScope(proxy)
 	case SidecarProxy:
-		ps.sidecarIndex.defaultSidecarMu.RLock()
+		ps.sidecarIndex.derivedSidecarMu.RLock()
 		var sc *SidecarScope
 		var exists bool
-		if ps.sidecarIndex.rootConfig != nil {
-			sc, exists = ps.sidecarIndex.rootConfigSidecarsByNamespace[proxy.ConfigNamespace]
+		if ps.sidecarIndex.meshRootSidecarConfig != nil {
+			sc, exists = ps.sidecarIndex.meshRootSidecarsByNamespace[proxy.ConfigNamespace]
 		} else {
-			sc, exists = ps.sidecarIndex.defaultSidecarsByNamespace[proxy.ConfigNamespace]
+			sc, exists = ps.sidecarIndex.catchAllSidecarsByNamespace[proxy.ConfigNamespace]
 		}
-		ps.sidecarIndex.defaultSidecarMu.RUnlock()
+		ps.sidecarIndex.derivedSidecarMu.RUnlock()
 
 		if exists {
 			// We have already computed the scope for this namespace, just return it.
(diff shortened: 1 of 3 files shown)


pilot/pkg/model/push_context.go
```diff
@@ -128,28 +128,28 @@ func newDestinationRuleIndex() destinationRuleIndex {
...
-	// computedSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar.
-	// These may be DefaultSidecarScopeForNamespace if rootConfig is empty or ConvertToSidecarScope if not.
+	// rootConfigSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar.
+	// These are converted from rootConfig if it exists.
 	// These are lazy-loaded. Access protected by defaultSidecarMu
-	computedSidecarsByNamespace map[string]*SidecarScope
-	// gatewayDefaultSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar,
-	// for gateways.
-	// Unlike computedSidecarsByNamespace, this is *always* the output of DefaultSidecarScopeForNamespace.
+	rootConfigSidecarsByNamespace map[string]*SidecarScope
```
hzxuzhonghu, line 138: root is also confusing to make me think this is from `istio-system`

pilot/pkg/model/push_context.go
```diff
@@ -128,28 +128,28 @@ func newDestinationRuleIndex() destinationRuleIndex {
...
+	// defaultSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar,
+	// These are *always* computed from DefaultSidecarScopeForNamespace.
 	// These are lazy-loaded. Access protected by defaultSidecarMu
-	gatewayDefaultSidecarsByNamespace map[string]*SidecarScope
-	defaultSidecarMu                  *sync.Mutex
+	defaultSidecarsByNamespace map[string]*SidecarScope
```
hzxuzhonghu, line 142: also confusing `default`

pilot/pkg/model/sidecar.go
```diff
@@ -552,7 +552,14 @@ func (sc *SidecarScope) DependsOnConfig(config ConfigKey) bool {
...
 // SidecarScope being used to avoid concurrent read/write.
-func (sc *SidecarScope) AddConfigDependencies(dependencies ...ConfigHash) {
+// Used only in tests.
+func (sc *SidecarScope) AddConfigDependenciesForTesting(dependencies ...ConfigHash) {
+	sc.addConfigDependencies(dependencies...)
+}
+
+// addConfigDependencies add extra config dependencies to this scope. This action should be done before the
+// SidecarScope being used to avoid concurrent read/write.
+func (sc *SidecarScope) addConfigDependencies(dependencies ...ConfigHash) {
```
hzxuzhonghu, line 562: I doubt the necessity of doing this, if we are changing this, I think we have many other methods/function that can be rewrite.

ramaraochavali: In general I think if it is just used in tests, it is better to follow this pattern so that we do not unnecessarily expose these. Wherever I am seeing this, I am changing if I am changing code near this. But I am not so strong about this - If you think, it is not good, I can revert this.

pilot/pkg/model/push_context.go
ramaraochavali: it is not istio-system. It can be any thing that is configured as rootNamespace in mesh-config

pilot/pkg/model/push_context.go
ramaraochavali: default is the 'Default' sidecar - I can not think of any better name. But If you have better suggestions, let me know

ramaraochavali:
Ok. I can revert this. Do you have any suggestions for naming?

hzxuzhonghu:
TBH, did not think about it. we need to think closely to prevent back and forth

ramaraochavali:
renamed see if this is better. But I am open to other suggestions

pilot/pkg/model/push_context.go
```diff
@@ -128,28 +128,29 @@ func newDestinationRuleIndex() destinationRuleIndex {
...
-	// for gateways.
-	// Unlike computedSidecarsByNamespace, this is *always* the output of DefaultSidecarScopeForNamespace.
+	meshRootSidecarsByNamespace map[string]*SidecarScope
+	// catchAllSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar,
+	// These are *always* computed from DefaultSidecarScopeForNamespace i.e. a sidecar that has listeners
+	// for all services in the mesh. This will be used if there is no sidecar specified in root namespace.
 	// These are lazy-loaded. Access protected by defaultSidecarMu
-	gatewayDefaultSidecarsByNamespace map[string]*SidecarScope
-	defaultSidecarMu                  *sync.Mutex
+	catchAllSidecarsByNamespace map[string]*SidecarScope
```
hzxuzhonghu, line 143: Maybe this can be called  defaultSidecarsByNamespace

pilot/pkg/model/push_context.go
```diff
@@ -128,28 +128,29 @@ func newDestinationRuleIndex() destinationRuleIndex {
...
-	// sidecars for each namespace
+	// user configured sidecars for each namespace if available.
 	sidecarsByNamespace map[string][]*SidecarScope
 	// the Sidecar for the root namespace (if present). This applies to any namespace without its own Sidecar.
-	rootConfig *config.Config
-	// computedSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar.
-	// These may be DefaultSidecarScopeForNamespace if rootConfig is empty or ConvertToSidecarScope if not.
+	meshRootSidecarConfig *config.Config
+	// meshRootSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar.
+	// These are converted from root namespace sidecar if it exists.
```
hzxuzhonghu, line 136: Have you changed this?

pilot/pkg/model/push_context.go
```diff
@@ -128,28 +128,29 @@ func newDestinationRuleIndex() destinationRuleIndex {
...
+	meshRootSidecarsByNamespace map[string]*SidecarScope
+	// catchAllSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar,
+	// These are *always* computed from DefaultSidecarScopeForNamespace i.e. a sidecar that has listeners
+	// for all services in the mesh. This will be used if there is no sidecar specified in root namespace.
 	// These are lazy-loaded. Access protected by defaultSidecarMu
-	gatewayDefaultSidecarsByNamespace map[string]*SidecarScope
-	defaultSidecarMu                  *sync.Mutex
+	catchAllSidecarsByNamespace map[string]*SidecarScope
+	// mutex to protect derived sidecars i.e. not specified by user.
+	derivedSidecarMu *sync.RWMutex
```
hzxuzhonghu, line 145: mutex

pilot/pkg/model/push_context.go
```diff
@@ -128,28 +128,29 @@ func newDestinationRuleIndex() destinationRuleIndex {
 
 // sidecarIndex is the index of sidecar rules
 type sidecarIndex struct {
-	// sidecars for each namespace
+	// user configured sidecars for each namespace if available.
 	sidecarsByNamespace map[string][]*SidecarScope
 	// the Sidecar for the root namespace (if present). This applies to any namespace without its own Sidecar.
-	rootConfig *config.Config
```
hzxuzhonghu, line 134: this is clear

pilot/pkg/model/push_context.go
```diff
@@ -128,28 +128,29 @@ func newDestinationRuleIndex() destinationRuleIndex {
...
-	// These may be DefaultSidecarScopeForNamespace if rootConfig is empty or ConvertToSidecarScope if not.
+	meshRootSidecarConfig *config.Config
+	// meshRootSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar.
+	// These are converted from root namespace sidecar if it exists.
 	// These are lazy-loaded. Access protected by defaultSidecarMu
-	computedSidecarsByNamespace map[string]*SidecarScope
-	// gatewayDefaultSidecarsByNamespace contains the default sidecar for namespaces that do not have a sidecar,
-	// for gateways.
-	// Unlike computedSidecarsByNamespace, this is *always* the output of DefaultSidecarScopeForNamespace.
+	meshRootSidecarsByNamespace map[string]*SidecarScope
```
hzxuzhonghu, line 138: I can see its meaning, this is computed from mesh global sidecar for namespaces that have no sidecar specified.

pilot/pkg/model/push_context.go
```diff
@@ -967,33 +968,56 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Insta
 
 	// We didn't have a Sidecar in the namespace. This means we should use the default - either an implicit
 	// default selecting everything, or pulling from the root namespace.
-	ps.sidecarIndex.defaultSidecarMu.Lock()
-	defer ps.sidecarIndex.defaultSidecarMu.Unlock()
-	if proxy.Type == Router {
-		sc, f := ps.sidecarIndex.gatewayDefaultSidecarsByNamespace[proxy.ConfigNamespace]
+	var computed *SidecarScope
+	switch proxy.Type {
```
hzxuzhonghu, line 972: can this select case be merged with abover one

pilot/pkg/model/push_context.go
```diff
@@ -967,33 +968,56 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Insta
...
+}
+
+func (ps *PushContext) computeSidecarScope(proxy *Proxy) *SidecarScope {
+	ps.sidecarIndex.derivedSidecarMu.Lock()
+	defer ps.sidecarIndex.derivedSidecarMu.Unlock()
 	var computed *SidecarScope
-	if ps.sidecarIndex.rootConfig != nil {
-		computed = ConvertToSidecarScope(ps, ps.sidecarIndex.rootConfig, proxy.ConfigNamespace)
-	} else {
+	switch proxy.Type {
```
hzxuzhonghu, line 1009: same here. so no need to abstract another function, since this makes this lock spread across functions

pilot/pkg/model/push_context.go
ramaraochavali: Just comment and rename

ramaraochavali:
@hzxuzhonghu any further comments?

pilot/pkg/model/push_context.go
```diff
@@ -925,22 +926,41 @@ func (ps *PushContext) DelegateVirtualServices(vses []config.Config) []ConfigHas
...
-	if sidecars, ok := ps.sidecarIndex.sidecarsByNamespace[proxy.ConfigNamespace]; ok {
-		// TODO: logic to merge multiple sidecar resources
-		// Currently we assume that there will be only one sidecar config for a namespace.
-		if proxy.Type == Router {
+	// TODO: logic to merge multiple sidecar resources
+	// Currently we assume that there will be only one sidecar config for a namespace.
+	sidecars, hasSidecar := ps.sidecarIndex.sidecarsByNamespace[proxy.ConfigNamespace]
+	switch proxy.Type {
+	case Router:
+		if hasSidecar {
```
hzxuzhonghu, line 934: For me it conflicts with the doc we say `sidecarsByNamespace` only contains sidecar specified by user, which should not be applied to gateway.

Though I know this is the previous logic (not sure https://github.com/istio/istio/pull/40688/files#diff-a8e332e06b003dfbf58c29f37de893fbb0e45f642200ea3ea2faa9847fc7e1e0R941 can be satisfied), I think we can safely remove it

pilot/pkg/model/push_context.go
```diff
@@ -925,22 +926,41 @@ func (ps *PushContext) DelegateVirtualServices(vses []config.Config) []ConfigHas
...
 				// Gateways should just have a default scope with egress: */*
 				if wrapper.Sidecar == nil {
 					return wrapper
 				}
 			}
 		}
-		if proxy.Type == SidecarProxy {
+		// We didn't have a Sidecar in the namespace. This means we should use the default - either an implicit
+		// default selecting everything, or pulling from the root namespace.
+		ps.sidecarIndex.derivedSidecarMutex.RLock()
```
hzxuzhonghu, line 948: lock and defer unlock here

pilot/pkg/model/push_context.go
```diff
@@ -963,38 +983,32 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Insta
 				return wrapper
 			}
 		}
-	}
+		ps.sidecarIndex.derivedSidecarMutex.RLock()
```
hzxuzhonghu, line 986: lock and defer unlock here

pilot/pkg/model/push_context.go
```diff
@@ -963,38 +983,32 @@ func (ps *PushContext) getSidecarScope(proxy *Proxy, workloadLabels labels.Insta
...
+			// We have already computed the scope for this namespace, just return it.
 			return sc
 		}
-		computed := DefaultSidecarScopeForNamespace(ps, proxy.ConfigNamespace)
-		ps.sidecarIndex.gatewayDefaultSidecarsByNamespace[proxy.ConfigNamespace] = computed
+		// We need to compute this namespace
+		computed := ConvertToSidecarScope(ps, ps.sidecarIndex.meshRootSidecarConfig, proxy.ConfigNamespace)
+		ps.sidecarIndex.derivedSidecarMutex.Lock()
+		if ps.sidecarIndex.meshRootSidecarConfig != nil {
+			ps.sidecarIndex.meshRootSidecarsByNamespace[proxy.ConfigNamespace] = computed
```
hzxuzhonghu, line 1004: can this be moved in L989 clause

pilot/pkg/model/push_context.go
ramaraochavali: I did this because, we can do a read only lock here as that is the most common use case once pod connects to istiod for further pushes. Write lock is only needed on init connects

ramaraochavali:
@hzxuzhonghu PTAL when you get chance

Issue #13861, opened 2019-05-07
Unable to remove server header


aaronmell:
**Bug description**
I should be able to remove/disable the server: istio-envoy when using istio  

**Expected behavior**

using the following configuration on a virtual service, I should be able to 

Headers:
      Response:
        Remove:
          server

**Steps to reproduce the bug**
Add a gateway that routes to a virtual service

**Version (include the output of `istioctl version --remote` and `kubectl version`)**
1.1.4
1.13

**How was Istio installed?**
Helm

**Environment where bug was observed (cloud vendor, OS, etc)**
Local kubernetes cluster on macos docker

**Affected product area (please put an X in all that apply)**

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[X ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience

nisan270390:
I'm facing with the same issue, I've upgraded istio from 1.0.6 to 1.1.5 and trying to manipulate the response server header.

> headers:
     response:
       remove:
         server

OR 
> headers:
     response:
       set:
         server: xxxxxx

but the server header haven't changed

prestonvanloon:
I tried to remove this in a rule as well, but the server field persists. 

```yaml
apiVersion: config.istio.io/v1alpha2
kind: rule
metadata:
  name: remove-server
  namespace: istio-system
spec:
  match: source.workload.name == "istio-ingressgateway"
  responseHeaderOperations:
    - name: server
      operation: REMOVE
```

kyessenov:
Envoy treats some headers as special, including this server header and content-length, and it overrides any plugin (e.g. istio) actions when HTTP specification suggests it. For example, empty bodies are going to have content length 0 because that's what clients expect.

Do you have an issue with the server header itself, or the value that the header is set to?

nisan270390:
@kyessenov I prefer to cover the proxy backend we're using. I'm wondering if there is an option to overwrite the server header value

aaronmell:
@kyessenov Our security/compliance team has an issue with it.

xring:
got same problem with 1.1.6, for security reasons we should be able to override this header.

elvizlai:
same issue.

rtmie:
similar experience on 1.1.4. Tried both set and remove on server header. Set would be sufficient for my use case.

slabber:
Is it possible that this PR will help:
[...]

PR #40689, opened 2022-08-27
add support for skipping server header

ramaraochavali:
Fixes https://github.com/istio/istio/issues/13861. As discussed in the issue, we want to default the ServerHeaderTransformation to 'PASSTHROUGH'. However for BC, added env var so that people can specify empty. We can change the default in future releases.

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -344,11 +344,15 @@ var (
 		" Pilot will use to keep configuration status up to date.  Smaller numbers will result in higher status latency, "+
 		"but larger numbers may impact CPU in high scale environments.").Get()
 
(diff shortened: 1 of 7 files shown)


howardjohn (commented):
should we do https://github.com/istio/api/pull/2240 instead? I know it seems abandoned but I was hoping to revive it this week..

ramaraochavali:
Sure I am fine with that as well. I can hold off on this till it is done

howardjohn:
Ok - feel free to review that one as well :-). Notably missing is anything
other than on/off... do you need that?

PR #40692, opened 2022-08-29
gateway-api: add mesh routes to RDS dependent configs

dddddai:
For http/tcp/tls mesh routes, we should be adding them(rather than the generated virtual services) to RDS dependent configs, otherwise the cache cannot get cleared correctly

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pkg/config/constants/constants.go
```diff
@@ -120,8 +120,9 @@ const (
 	// Label to skip config comparison.
 	AlwaysPushLabel = "internal.istio.io/always-push"
 
-	// InternalParentName declares the original resource of an internally-generate config. This is used by ingress and the gateway-api.
-	InternalParentName     = "internal.istio.io/parent"
+	// InternalParentNames declares the original resources of an internally-generate config. This is used by k8s gateway-api.
+	// It is a comma separated list. For example, "HTTPRoute/foo.default,HTTPRoute/bar.default"
+	InternalParentNames    = "internal.istio.io/parents"
```
hzxuzhonghu, line 125: I think this is unrelated

pilot/pkg/networking/core/v1alpha3/route/route_cache.go
```diff
@@ -96,7 +97,12 @@ func (r *Cache) DependentConfigs() []model.ConfigHash {
 		}.HashCode())
 	}
 	for _, vs := range r.VirtualServices {
-		configs = append(configs, model.ConfigKey{Kind: kind.VirtualService, Name: vs.Name, Namespace: vs.Namespace}.HashCode())
+		if model.UseGatewaySemantics(vs) {
```
hzxuzhonghu, line 100: Do we miss this for Ingress?

pilot/pkg/networking/core/v1alpha3/route/route_cache.go
```diff
@@ -96,7 +97,12 @@ func (r *Cache) DependentConfigs() []model.ConfigHash {
...
+			// add http/tcp/tls routes to dependent configs
+			configs = addInternalParents(configs, vs)
```
hzxuzhonghu, line 102: This is not the right way for xds cache to be aware of gateway API. Except for gateway controller, other istio modules should be unaware of K8s Gateway/Ingress

pilot/pkg/networking/core/v1alpha3/route/route_cache.go
dddddai: AFAIK rds cache only applies to sidecars, so we don't need to care about ingress
But for gateway-api we are supporting **mesh** routes applying to sidecars

pilot/pkg/networking/core/v1alpha3/route/route_cache.go
dddddai: > Except for gateway controller, other istio modules should be unaware of K8s Gateway/Ingress

Ideally we should do this, but so far I don't see another way to clear cache on route updates

hzxuzhonghu: I donot look into it, but this looks not good to me

howardjohn: @hzxuzhonghu I am not sure any other possibility. A Gateway-API VirtualService is synthetic and has dependencies outside itself. We could make it not say "GatewaySemantics" in the function name, but I am not sure how we can avoid this check.

Thinking about it more -- doesn't the same issue apply to `Ingress`?

dddddai: > Thinking about it more -- doesn't the same issue apply to `Ingress`?

No, please see https://github.com/istio/istio/pull/40692#discussion_r957258617

howardjohn (approved):

PR #40694, opened 2022-08-29
fix AddRunningKubeSourceWithRevision return error when IstioOperator updated

### LINKED ISSUE #39599: Istio Operator Error Loops
### Bug Description

We're upgrading our Istio cluster from v1.12.7 to v1.14.1 using Istio Operator. Istio cluster itself is fine, ingressgateway and Istio proxy are updated, but Istio Operator keep looping error and usage memory keep increased.

### Version

```prose
$ istioctl version
client version: 1.14.1
control plane version: 1.14.1
$ kubectl version --short
Client Version: v1.23.3
Server Version: v1.21.11-gke.900
```

### Additional Information

```
error analysis error setting up error handling for kube crdclient: 2 errors occurred:
	* informer has already started
	* informer has already started
```

zirain:
fixes: #39599

### DIFF
--- a/pkg/config/analysis/local/istiod_analyze.go
+++ b/pkg/config/analysis/local/istiod_analyze.go
@@ -262,11 +262,11 @@ func (sa *IstiodAnalyzer) AddReaderKubeSource(readers []ReaderSource) error {
 
 // AddRunningKubeSource adds a source based on a running k8s cluster to the current IstiodAnalyzer
 // Also tries to get mesh config from the running cluster, if it can
-func (sa *IstiodAnalyzer) AddRunningKubeSource(c kubelib.Client) {
-	sa.AddRunningKubeSourceWithRevision(c, "default")
+func (sa *IstiodAnalyzer) AddRunningKubeSource(c kubelib.Client, ignoreSetWatchError bool) {
+	sa.AddRunningKubeSourceWithRevision(c, "default", ignoreSetWatchError)
 }
 
-func (sa *IstiodAnalyzer) AddRunningKubeSourceWithRevision(c kubelib.Client, revision string) {
+func (sa *IstiodAnalyzer) AddRunningKubeSourceWithRevision(c kubelib.Client, revision string, ignoreSetWatchError bool) {
 	// TODO: are either of these string constants intended to vary?
 	// This gets us only istio/ ones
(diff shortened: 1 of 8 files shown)


pkg/config/analysis/local/istiod_analyze.go
```diff
@@ -281,7 +281,7 @@ func (sa *IstiodAnalyzer) AddRunningKubeSourceWithRevision(c kubelib.Client, rev
 		// better to fail fast, and get a good idea for the failure.
 		scope.Analysis.Errorf("Failed to watch crd resource for analysis: %s", err)
 	})
-	if err != nil {
+	if err != nil && !ignoreSetWatchError {
```
howardjohn, line 284: This seems like either a code smell OR something that is under-documented. Ignoring error is generally a bad idea, so if we should be doing it we need some explanation in the comments.

However, it does seem odd to do this. It seems like it would be better to have some smarter logic about setting it only once rather than ignoring errors. check before we attempt rather than ignoring error

litong01:
@zirain was there an issue open on this? I like to understand a bit better what problems we are trying to fix.

pkg/config/analysis/local/istiod_analyze.go
hzxuzhonghu: Why could we met SetWatchErrorHandler return error?

zirain:
> @zirain was there an issue open on this? I like to understand a bit better what problems we are trying to fix.

https://github.com/istio/istio/issues/39599

pkg/config/analysis/local/istiod_analyze.go
zirain: In Istio-Operator pod, `Reconcile` is called more than once with a share client.

hzxuzhonghu: IC when the informer has been started, the SetWatchErrorHandler will return error. 

So can we check and SetWatchErrorHandler when the informer has not been started. I think this way is better than passing a bool

zirain: there's no proper function on `cache.SharedInformer`, `HasStarted` can not be called,  `HasSynced` seems not suitable for this scenario.

zirain: after talking with @hzxuzhonghu offline, will find a better way to resolve this.

zirain:
replaced by https://github.com/istio/istio/pull/40727

PR #40696, opened 2022-08-29
Fix incorrect 1123label validation

hanxiaop:

### DIFF
--- a/pkg/config/labels/instance.go
+++ b/pkg/config/labels/instance.go
@@ -25,7 +25,7 @@ import (
 
 const (
 	DNS1123LabelMaxLength = 63 // Public for testing only.
-	dns1123LabelFmt       = "[a-zA-Z0-9](?:[-a-zA-Z0-9]*[a-zA-Z0-9])?"
+	dns1123LabelFmt       = "[a-z0-9](?:[-a-z0-9]*[a-z0-9])?"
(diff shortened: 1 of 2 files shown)


pkg/config/labels/instance.go
```diff
@@ -25,7 +25,7 @@ import (
 
 const (
 	DNS1123LabelMaxLength = 63 // Public for testing only.
-	dns1123LabelFmt       = "[a-zA-Z0-9](?:[-a-zA-Z0-9]*[a-zA-Z0-9])?"
+	dns1123LabelFmt       = "[a-z0-9](?:[-a-z0-9]*[a-z0-9])?"
```
hzxuzhonghu, line 28: Unlike the k8s, we are using this to validate Kind, DNS 1123 can contain upper cases

https://github.com/kubernetes/kubernetes/pull/39675/files

howardjohn (changes requested):
This seems like a breaking change. Is there any more context on why we should do this?

pkg/config/labels/instance.go
hanxiaop: @hzxuzhonghu Thanks

hanxiaop:
> This seems like a breaking change. Is there any more context on why we should do this?

@howardjohn I was trying to add some dns1123label validation to iop fields in validate, and I think the uppercase should be checked as well. I'll close this one and do it in another way.

PR #40697, opened 2022-08-29
Gateway api route kind status fixes

### LINKED ISSUE #40675: K8s Gateway should set ResolvedRefs condition to false for incompatible route type
### Bug Description

A Gateway should fail to become ready if it explicitly supports any route type incompatible with the protocol type and the gateway listener should have a false ResolvedRefs condition with reason InvalidRouteKinds.

https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.ListenerStatus

### Version

```prose
latest
```

### Additional Information

_No response_

### Affected product area

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Upgrade
- [ ] Multi Cluster
- [ ] Virtual Machine
- [ ] Control Plane Revisions

frankbu:

Fixes #40675

### DIFF
--- a/pilot/pkg/config/kube/gateway/conditions.go
+++ b/pilot/pkg/config/kube/gateway/conditions.go
@@ -257,17 +257,25 @@ func reportListenerCondition(index int, l k8s.Listener, obj config.Config, condi
 			gs.Listeners = append(gs.Listeners, k8s.ListenerStatus{})
 		}
 		cond := gs.Listeners[index].Conditions
+		supported, valid := generateSupportedKinds(l)
+		if !valid {
+			conditions[string(k8s.ListenerConditionResolvedRefs)] = &condition{
+				reason:  string(k8s.ListenerReasonInvalidRouteKinds),
+				status:  metav1.ConditionFalse,
+				message: "Invalid route kinds",
+			}
+		}
 		gs.Listeners[index] = k8s.ListenerStatus{
 			Name:           l.Name,
 			AttachedRoutes: 0, // this will be reported later
-			SupportedKinds: generateSupportedKinds(l),
+			SupportedKinds: supported,
 			Conditions:     setConditions(obj.Generation, cond, conditions),
 		}
 		return gs
 	})
 }
 
-func generateSupportedKinds(l k8s.Listener) []k8s.RouteGroupKind {
+func generateSupportedKinds(l k8s.Listener) ([]k8s.RouteGroupKind, bool) {
 	supported := []k8s.RouteGroupKind{}
 	switch l.Protocol {
 	case k8s.HTTPProtocolType, k8s.HTTPSProtocolType:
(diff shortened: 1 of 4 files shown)


pilot/pkg/config/kube/gateway/testdata/route-binding.yaml
```diff
@@ -30,6 +30,8 @@ spec:
     allowedRoutes:
       namespaces:
         from: All
+      kinds:
+      - kind: HTTPRoute
```
howardjohn, line 34: I don't get why this is rejected. It is HTTPRoute for HTTP... isn't that valid?

howardjohn (commented):
should we also bump the gateway-api dependency so we have test coverage for this?

pilot/pkg/config/kube/gateway/testdata/route-binding.yaml
frankbu: Yes, this one is valid. I added this to test the case where the kind is explicitly set to a valid kind, because we had no such test, and therefore didn't catch a bug where supportedKinds wasn't being set correctly  (`s == kind` check in `generateSupportedKinds` didn't work).

> should we also bump the gateway-api dependency so we have test coverage for this?

I think there's still another test that will fail. I can either try to fix that in this PR, before we merge. Otherwise, I was planning to look at that one separately, and then bump the dependency when everything is working.

pilot/pkg/config/kube/gateway/conditions.go
```diff
@@ -288,13 +296,25 @@ func generateSupportedKinds(l k8s.Listener) []k8s.RouteGroupKind {
...
+	}
+	return supported, true
+}
+
+// This and the following function really belongs in some gateway-api lib
+func routeGroupKindEqual(rgk1, rgk2 k8s.RouteGroupKind) bool {
+	return rgk1.Kind == rgk2.Kind && getGroup(rgk1) == getGroup(rgk2)
+}
+
+func getGroup(rgk k8s.RouteGroupKind) k8s.Group {
```
howardjohn, line 315: nit: elsewhere we are using `defaultIfNil((*string)(p.Group), gvk.KubernetesGateway.Group)`. I don't mind having a specific getGroup but we may want to use those primitives (defaultIfNil and the group constant)

pilot/pkg/config/kube/gateway/conditions.go
```diff
@@ -257,17 +257,25 @@ func reportListenerCondition(index int, l k8s.Listener, obj config.Config, condi
 			gs.Listeners = append(gs.Listeners, k8s.ListenerStatus{})
 		}
 		cond := gs.Listeners[index].Conditions
+		supported, valid := generateSupportedKinds(l)
+		if !valid {
+			conditions[string(k8s.ListenerConditionResolvedRefs)] = &condition{
+				reason:  string(k8s.ListenerReasonInvalidRouteKinds),
+				status:  metav1.ConditionFalse,
+				message: "Invalid route kinds",
```
howardjohn, line 265: nit: could be helpful to make this more clear, like `Invalid route kinds: HTTPRoute cannot be used with protocol TLS` or something

howardjohn (approved):
thanks!

pilot/pkg/config/kube/gateway/conditions.go
frankbu: This could get a little wordy since in the general case because there could be several incompatible kinds that don't intersect. The spec says that the `supportedKinds` field must include the subset that are compatible, so comparing that to the requested kinds indicates which ones are invalid.

PR #40700, opened 2022-08-29
Fix sync issue in GCP platform

Aakash2017:
Fix issue with SyncOnlyOnce being global.

- [ X] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ X] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/bootstrap/platform/gcp.go
+++ b/pkg/bootstrap/platform/gcp.go
@@ -123,11 +123,13 @@ type (
 
 type gcpEnv struct {
 	sync.Mutex
-	metadata           map[string]string
-	shouldFillMetadata bool
+	metadata map[string]string
 }
 
-var gcpEnvOnce sync.Once
+var (
+	fillMetadata bool
+	gcpEnvOnce   sync.Once
+)
 
 // IsGCP returns whether or not the platform for bootstrapping is Google Cloud Platform.
 func IsGCP() bool {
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
While this seems reasonable, one question I have is why does it being global break things? Shouldn't the metadata be persistently the same throughout the lifecycle?

Aakash2017:
> While this seems reasonable, one question I have is why does it being global break things? Shouldn't the metadata be persistently the same throughout the lifecycle?

IIUC gcpEnvOnce, a global var, is used to determine whether metadata, a member function of gcpEnv, should be populated with various data such as cluster_url, gce instance id, etc. On the first call, metadata is constructed correctly. On subsequent calls, shouldFillMetadata gets set to false and the metadata is constructed incorrectly.

How the gcpEnvOnce determines metadata: https://github.com/istio/istio/blob/master/pkg/bootstrap/platform/gcp.go#L151

Example of usage: https://github.com/istio/istio/blob/master/security/pkg/nodeagent/plugin/providers/google/stsclient/stsclient.go#L167

ingwonsong:
I agree with @howardjohn. 
I think that [gcpEnv](https://github.com/istio/istio/blob/bbebdff870098cd5a0ca002632eb47e60b2f34a3/pkg/bootstrap/platform/gcp.go#L143) can be a singleton. Metadata can be changed on running?

pkg/bootstrap/platform/gcp.go
```diff
@@ -286,7 +288,7 @@ func zoneToRegion(z string) (string, error) {
 // Locality returns the GCP-specific region and zone.
 func (e *gcpEnv) Locality() *core.Locality {
 	var l core.Locality
-	if e.shouldFillMetadata {
+	if fillMetadata {
```
howardjohn, line 291: I think this may be broken (before and after this change). If Metadata isn't called first then `fillMetadata` is always empty.

It may be worth just switching to `pkg/lazy` which makes these errors ~impossible

howardjohn (approved):
strictly better so Ill approve now, but we may want to followup/send another commit

pkg/bootstrap/platform/gcp.go
Aakash2017: sounds good, lets fix it in a follow up commit

PR #40715, opened 2022-08-30
gateway-api: skip GatewaySecretMissingReferencedSecret for now

frankbu:


howardjohn (approved):
should we bump go.mod aswell?

frankbu:
> should we bump go.mod aswell?

Sure, we can try. I expect it to pass, but we'll see

Is there a specific process for what exactly to bump it to?

howardjohn:
Just go straight to latest main/master

PR #40719, opened 2022-08-30
fix PILOT_ENABLE_RDS_CACHE flag not working

yingzhuivy:

When building RDS, cache get is always conducted not respecting the
PILOT_ENABLE_RDS_CACHE flag. This PR fixes that.

Cache add already respects the flag:
https://github.com/istio/istio/blob/1.14.3/pilot/pkg/networking/core/v1alpha3/httproute.go#L212

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/httproute.go
+++ b/pilot/pkg/networking/core/v1alpha3/httproute.go
@@ -351,19 +351,21 @@ func BuildSidecarOutboundVirtualHosts(node *model.Proxy, push *model.PushContext
 			return services[i].Hostname <= services[j].Hostname
 		})
 
-		routeCache = &istio_route.Cache{
-			RouteName:               routeName,
-			ProxyVersion:            node.Metadata.IstioVersion,
-			ClusterID:               string(node.Metadata.ClusterID),
-			DNSDomain:               node.DNSDomain,
-			DNSCapture:              bool(node.Metadata.DNSCapture),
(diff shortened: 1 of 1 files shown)


pilot/pkg/networking/core/v1alpha3/httproute.go
```diff
@@ -375,9 +377,11 @@ func BuildSidecarOutboundVirtualHosts(node *model.Proxy, push *model.PushContext
 	// Get list of virtual services bound to the mesh gateway
 	virtualHostWrappers := istio_route.BuildSidecarVirtualHostWrapper(routeCache, node, push, servicesByName, virtualServices, listenerPort)
 
-	resource, exist := xdsCache.Get(routeCache)
-	if exist && !features.EnableUnsafeAssertions {
-		return nil, resource, routeCache
+	if features.EnableRDSCaching {
```
hzxuzhonghu, line 380: for get we donot need this

yingzhuivy:
yes, disabling it helps with propagation delay. I posted my testing here: https://github.com/istio/istio/issues/40744

pilot/pkg/networking/core/v1alpha3/httproute.go
yingzhuivy: why is that? for cds we do respect it for get as well
https://github.com/istio/istio/blob/1.14.3/pilot/pkg/networking/core/v1alpha3/cluster_builder.go#L1271

hzxuzhonghu (approved):
improve on not construct

Issue #35833, opened 2021-11-02
xds piggyback: merge with StatusPort (15020)


howardjohn:
Right now we have 2 debug interfaces serving HTTP traffic. I am not sure we need 2; I think we can consolidate to just port 15020.

cc @esnible

ingwonsong:
@howardjohn Are there two HTTP interfaces? 
If I read the code correctly, `pilot/pkg/xds/statusgen.go` (xds) and `pilot/pkg/xds/debug.go` (http) /`pilot/pkg/xds/debuggen.go` (xds) seems to be such debugging interfaces.

howardjohn:
there is already port 15020 used for metrics, readiness, etc

ingwonsong:
Yes. 
IIUC, `pilot/pkg/xds/debug.go` is served in port 15020 with the other stuffs. 

Is there another HTTP interface for debugging?

ingwonsong:
Oh.. I understood. 
You mean "pilot agent" not "pilot. So, you suggested to merge 15004 into 15020. Right?

howardjohn:
yeah I don't think we need 2?wdyt?

ingwonsong:
yes. I agree with you a ton.

BTW, is there no need to be compatible with previous versions of istioctl?
(just question)

howardjohn:
is the piggyback even used yet? I didn't think it was.

If it is we can always add it on both ports for 1 release then move it. but
since it's experimental probably nbd

ingwonsong:
I am sorry for my confusion. 
You are right. it is not used in istioctl as far as I know.

PR #40721, opened 2022-08-30
Merge the Tap Service to the Status server

ingwonsong:
Fixes: https://github.com/istio/istio/issues/35833 (to avoid breaking change, 15004 port is still supported in this PR)

The changes are as follows:
- `tap service` (previously only using 15004 port) is served on 15004 as well as the status port (15020).
- `tap_proxy.go` is moved to `pilot/cmd/pilot-agent/status/debugtap/proxy.go`.
- Due to the import cycle, `pilot/cmd/pilot-agent/status/server.go` is moved to `pilot/cmd/pilot-agent/status/server/server.go`, and only the parts which are shared with the other packages are exposed in `pilot/cmd/pilot-agent/status/status.go`.
- `tap service` can be disabled by `PROXY_XDS_DEBUG_VIA_AGENT` on both of 15004 and 15020
- If `PROXY_XDS_DEBUG_VIA_AGENT_PORT` is set to `0`, the status port is only used for serving `tap service`.

cc: @howardjohn


hzxuzhonghu:
This is a breaking change

ingwonsong:
The port number 15004 is not used currently. But, if we need to keep it, I can support 15004 as well for keeping compatibility.

@howardjohn what do you think?

ingwonsong:
To give some transition periods for "just-in-case", `15004` port also serves the debugging interface.
15004 port can be disabled by setting `PROXY_XDS_DEBUG_VIA_AGENT_PORT=0`.

ingwonsong:
https://github.com/istio/istio/pull/40687 would be another way to resolve the dependency cycle issue instead of refactoring "server.go".

@howardjohn Can I get your opinion?

pilot/cmd/pilot-agent/status/server/server.go
```diff
@@ -355,6 +329,24 @@ func (s *Server) Run(ctx context.Context) {
...
 
+	var handler http.Handler = mux
+	if s.config.DebugTapClient != nil {
+		debugTapProxy := debugtap.NewProxy(s.config.DebugTapClient)
+
+		// Registers HTTP Handlers for Debug Tap
+		debugTapProxy.RegisterHTTPHandler(mux)
+		grpcs := grpc.NewServer(istiogrpc.ServerOptions(istiokeepalive.DefaultOption())...)
+		debugTapProxy.RegisterGRPCHandler(grpcs)
+		mixedHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
howardjohn, line 340: do we need the `isRequestFromLocalhost` stuff?

pilot/cmd/pilot-agent/app/cmd.go
```diff
@@ -172,6 +165,14 @@ func newProxyCommand() *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			// If a status port was provided, start handling status probes.
+			if proxyConfig.StatusPort > 0 {
+				if err := initStatusServer(ctx, proxy, proxyConfig, agentOptions.EnvoyPrometheusPort, agent); err != nil {
```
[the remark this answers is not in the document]
ingwonsong: I see.

I revert the order of initialization as you commented, and instead, changed for the tap server to get the xds proxy in lazy way.

Please take a look into that once again. Thanks!

pilot/cmd/pilot-agent/status/server/server.go
ingwonsong: Good point. Done.

PR #40727, opened 2022-08-31
operator: should not call `SetWatchErrorHandler` if store has started

zirain:
Signed-off-by: hejianpeng <hejianpeng2@huawei.com>

1. add `HasStarted() bool` to model.ConfigStoreController
2. do not can SetWatchErrorHandler if it's started

### DIFF
--- a/pilot/pkg/config/kube/ingress/controller.go
+++ b/pilot/pkg/config/kube/ingress/controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"sort"
 	"sync"
+	"sync/atomic"
 
 	"github.com/hashicorp/go-multierror"
 	ingress "k8s.io/api/networking/v1beta1"
@@ -95,6 +96,8 @@ type controller struct {
 	serviceLister   listerv1.ServiceLister
 	// May be nil if ingress class is not supported in the cluster
 	classes v1beta1.IngressClassInformer
+
+	started atomic.Bool
 }
 
 // TODO: move to features ( and remove in 1.2 )
@@ -180,7 +183,12 @@ func NewController(client kube.Client, meshWatcher mesh.Holder,
 	return c
 }
 
+func (c *controller) HasStarted() bool {
+	return c.started.Load()
+}
+
 func (c *controller) Run(stop <-chan struct{}) {
+	c.started.Store(true)
 	c.queue.Run(stop)
 }
 
--- a/pilot/pkg/model/config.go
+++ b/pilot/pkg/model/config.go
@@ -194,7 +194,12 @@ type ConfigStoreController interface {
 
 	// Run until a signal is received
 	Run(stop <-chan struct{})
+
+	// SetWatchErrorHandler should be call if store has started
 	SetWatchErrorHandler(func(r *cache.Reflector, err error)) error
+
+	// HasStarted return treu after store started.
+	HasStarted() bool
 
 	// HasSynced returns true after initial cache synchronization is complete
 	HasSynced() bool
--- a/pilot/pkg/model/config.go
+++ b/pilot/pkg/model/config.go
@@ -198,7 +198,7 @@ type ConfigStoreController interface {
 	// SetWatchErrorHandler should be call if store has started
 	SetWatchErrorHandler(func(r *cache.Reflector, err error)) error
 
-	// HasStarted return treu after store started.
+	// HasStarted return ture after store started.
 	HasStarted() bool
 
 	// HasSynced returns true after initial cache synchronization is complete
--- a/pkg/config/analysis/local/istiod_analyze.go
+++ b/pkg/config/analysis/local/istiod_analyze.go
@@ -213,6 +213,10 @@ func (d dfCache) RegisterEventHandler(kind config.GroupVersionKind, handler mode
 	panic("implement me")
 }
 
+func (d dfCache) HasStarted() bool {
+	return true
+}
+
 // Run intentionally left empty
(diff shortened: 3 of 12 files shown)


pkg/kube/client.go
```diff
@@ -548,6 +552,18 @@ func (c *client) RunAndWait(stop <-chan struct{}) {
...
+func (c *client) startInformer(stop <-chan struct{}) {
+	if !c.HasStarted() {
+		c.started.Store(true)
+		c.kubeInformer.Start(stop)
+		c.dynamicInformer.Start(stop)
+		c.metadataInformer.Start(stop)
+		c.istioInformer.Start(stop)
+		c.gatewayapiInformer.Start(stop)
+		c.extInformer.Start(stop)
+	}
```
hzxuzhonghu, line 564: This is not right, say kubeInformer, it is a informer factory, we can use it to init pod/node/service, etc informers, if at one time only pod informer is initiated, and this function is called. But then node informer is initiated, and need to start, it will skip the start here.

cainelli:
Thank you for looking into this @zirain . Would it make sense to backport this to 1.14.x? If yes, is `atomic.Bool` available? I see 1.14.3 is still using Go 1.17.

pilot/pkg/config/kube/ingress/controller.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"sort"
 	"sync"
+	"sync/atomic"
```
cainelli, line 24: shouldn't it be "go.uber.org/atomic"?

zirain:
linked to https://github.com/istio/istio/issues/39599

pilot/pkg/config/kube/ingress/controller.go
zirain: I think we can use `sync/atomic` in go1.18

releasenotes/notes/40727.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: installation
+
```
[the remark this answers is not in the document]
zirain: done

GregHanson (approved):

pilot/pkg/model/config.go
```diff
@@ -194,8 +194,13 @@ type ConfigStoreController interface {
 
 	// Run until a signal is received
 	Run(stop <-chan struct{})
+
+	// SetWatchErrorHandler should be call if store has started
 	SetWatchErrorHandler(func(r *cache.Reflector, err error)) error
 
+	// HasStarted return treu after store started.
```
ericvn, line 201: c/treu/true

releasenotes/notes/40727.yaml
```diff
@@ -0,0 +1,10 @@
...
+issue:
+  - 39599
+
+releaseNotes:
+  - |
+    **Fixed** an issue `AddRunningKubeSourceWithRevision` return error cause Istio Operator going to error loops.
```
ericvn, line 10:
```suggestion
    **Fixed** an issue where `AddRunningKubeSourceWithRevision` returns an error causing the Istio Operator to go into an error loop.
```

ericvn:
> Thank you for looking into this @zirain . Would it make sense to backport this to 1.14.x? If yes, is `atomic.Bool` available? I see 1.14.3 is still using Go 1.17.

The 1.14 branch is built with Go 1.18, but the go.mod:go value is still reflecting 1.17.

zirain:
> > Thank you for looking into this @zirain . Would it make sense to backport this to 1.14.x? If yes, is `atomic.Bool` available? I see 1.14.3 is still using Go 1.17.
> 
> The 1.14 branch is built with Go 1.18, but the go.mod:go value is still reflecting 1.17. sync/atomic Bool was introduced in 1.19.

I'd keep use sync/atomic in master, and will be replace to `go.uber.org/atomic` during cherry-pick

ericvn (approved):

pkg/config/analysis/local/istiod_analyze.go
```diff
@@ -276,15 +280,20 @@ func (sa *IstiodAnalyzer) AddRunningKubeSourceWithRevision(c kubelib.Client, rev
...
-	err = store.SetWatchErrorHandler(func(r *cache.Reflector, err error) {
-		// failed resources will never be synced, which causes the process to hang indefinitely.
-		// better to fail fast, and get a good idea for the failure.
-		scope.Analysis.Errorf("Failed to watch crd resource for analysis: %s", err)
-	})
-	if err != nil {
-		scope.Analysis.Errorf("error setting up error handling for kube crdclient: %v", err)
-		return
+	if !store.HasStarted() {
+		err = store.SetWatchErrorHandler(func(r *cache.Reflector, err error) {
```
hzxuzhonghu, line 284: Looks this will make some informer having no watch error handler, not a big deal.

Just wondering if we could check HasStarted() for each informer

zirain: In an real world, this will not happen in istioctl, istioctl always run with new client.
For istio-operator, all informers started in loop-0, when users modify the iop configuration, re-set watch error handler will cause error.

hzxuzhonghu (approved):

pilot/pkg/config/kube/ingress/controller.go
hzxuzhonghu: Isn't atomic.Bool introduced in 1.19, while istio still use 1.18?

PR #40730, opened 2022-08-31
Replace anypb.New with MessageToAny

Chen-Xintong:

Use the `MessageToAny `function defined by Istio to convert other proto type messages into Any types, thereby simplifying the code and unifying the interface

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/tracing.go
+++ b/pilot/pkg/networking/core/v1alpha3/tracing.go
@@ -146,7 +146,7 @@ func configureFromProviderConfig(pushCtx *model.PushContext, meta *model.NodeMet
 					CollectorCluster: clusterName,
 					AccessTokenFile:  provider.Lightstep.AccessToken,
 				}
-				return protoconv.MessageToAny(lc), nil
+				return protoconv.MessageToAnyWithError(lc)
 			})
 
 	case *meshconfig.MeshConfig_ExtensionProvider_Opencensus:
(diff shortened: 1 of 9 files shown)


GregHanson (approved):

howardjohn (approved):

pilot/pkg/networking/core/v1alpha3/tracing.go
```diff
@@ -269,14 +270,14 @@ func zipkinConfigGen(hostname, cluster string) (*anypb.Any, error) {
 		TraceId_128Bit:           true,
 		SharedSpanContext:        wrapperspb.Bool(false),
 	}
-	return anypb.New(zc)
+	return protoconv.MessageToAny(zc), nil
```
douglas-reid, line 273: doesn't have to be in this PR, but the repeated `protoconv.MessageToAny(), nil` in these `*Gen` fns makes me wonder if `typedConfigGenFromClusterFn` should be refactored to drop the `error` return parameter (as nothing seems to use it).

douglas-reid (approved):

pkg/config/model.go
```diff
@@ -128,7 +129,7 @@ func ToProto(s Spec) (*anypb.Any, error) {
 	// golang/protobuf 1.4+ will have this interface. Older golang/protobuf are gogo compatible
 	// but also not used by Istio at all.
 	if pb, ok := s.(protoreflect.ProtoMessage); ok {
-		return anypb.New(pb)
+		return protoconv.MessageToAny(pb), nil
```
hzxuzhonghu, line 132: why not `MessageToAnyWithError`, so we would not miss the err value

Chen-Xintong: `MessageToAny` will call `MessageToAnyWithError`, and it will return error if failed. But what you said makes sense, here we can directly call the `MessageToAnyWithError `function, thx.

Chen-Xintong: Fixed.

hzxuzhonghu (approved):

hanxiaop (approved):

PR #40739, opened 2022-08-31
Remove obsolete kubernetes-blog sample

frankbu:

This sample was referenced by a very old Kubernetes blog post which is now completely outdated, so we can remove the sample now.


ericvn (approved):

ericvn:
I don't see a reference to the files in the current istio.io repository.

frankbu:
> I don't see a reference to the files in the current istio.io repository.

Yeah, it was never referenced from istio.io. It was used in a 5 year old k8s blog post that configured routes using RouteRule instead of VirtualService :) 

https://kubernetes.io/blog/2017/10/request-routing-and-policy-management/

Issue #40744, opened 2022-08-31
XDS cache impacts propagation delay


yingzhuivy:
### Bug Description

This is more of a performance issue rather than a bug.

We notice that turning on xds cache helps with cpu, but impacts propagation delay a lot due to cache locking. We conducted the following tests:

# setup
Control plane:
- 1 istiod pod 
- memory 26 G, cpu 10 cores

Data plane:
- 50 services and 500 pods
- and we mimic changes by pod rotation (roughly 5 pods rotated every 10 seconds) and istio config change (one virtual service routing change randomly per 5 seconds)

# results
I got the following results:
- When we have **all cache enabled** (default), the proxy convergence time (pilot_proxy_convergence_time) was on average 2.9s for P90, and 4.26s for P99.
- **Disable CDS cache alone**, the proxy convergence time was on average 2.1s for P90, and 3s for P99.
- **Disable both CDS and RDS cache** (with this [change](https://github.com/istio/istio/pull/40719) patched), the proxy convergence time was on average 0.94s for P90, and 2.55s for P99. Some CPU throttling occurs so I think if more CPU is given, propagation delay (especially P99) should further improve.
- **Disable all cache**, the proxy convergence time was on average 0.85s for P90, and 2.39s for P99. CPU throttling was heavier than previously as expected.

I also enabled the mutex profile in a custom istiod build and saw 70% of time (209s out of 300s) was spent on waiting for lock for lruCache in the default case (all cache enabled):
<img width="777" alt="Screen Shot 2022-08-29 at 10 49 20 AM" src="https://user-images.githubusercontent.com/55459889/187754608-392fd152-93ca-4a87-a648-10330f914efd.png">

# proposal
I think for the short term, we should fix this bug to allow users to opt out RDS caching properly: https://github.com/istio/istio/pull/40719
For the longer term, is there any work planned for improving the xds caching? Some ideas:
- Cache per xds type. Right now seems like all xds calculation uses the same cache, separating out might help with lock contention
- Explore high concurrency lru caches (for example https://github.com/karlseguin/ccache), or other cache options

### Version

```prose
Istio 1.13.5
```

hzxuzhonghu:
I could not understand `Disable ` will cause pilot_proxy_convergence_time down. Could you elaborate a little bit more

hzxuzhonghu:
@yingzhuivy can you retest with the above patch if possible

hzxuzhonghu:
[...]

PR #40750, opened 2022-09-01
Reduce cache.Get lock conflict

hzxuzhonghu:

For #40744

### DIFF
--- a/pilot/pkg/model/xds_cache.go
+++ b/pilot/pkg/model/xds_cache.go
@@ -319,8 +319,8 @@ func (l *lruCache) Get(entry XdsCacheEntry) (*discovery.Resource, bool) {
 		return nil, false
 	}
 	k := entry.Key()
-	l.mu.Lock()
-	defer l.mu.Unlock()
+	l.mu.RLock()
+	defer l.mu.RUnlock()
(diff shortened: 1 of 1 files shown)


hzxuzhonghu:
Yes, i see the ut raced.

hzxuzhonghu:
It is not evict, but `c.evictList.MoveToFront(ent)`

hzxuzhonghu:
looks like this should be fixed in upstream

stevenctl:
That repo hasn't seen a commit in > 1.5 year. If we want the change we probably have to contribute it ourselves. 
I wonder if we're hitting the ceiling (XDSCacheMaxSize) of the LRU in practice.

jefferai:
It's really not clear from the issue reported in `golang-lru` what the problem is, or from this PR. Happy to have this fixed in `golang-lru`, especially if you know what's wrong and can PR it.

howardjohn:
I don't think its a bug in golang-lru at all; by nature of the implementations Get() in the LRU mutates data (to evict data/record use), so using a read lock only (rather than write lock) is not sufficient on Get().

jefferai:
Ah, OK, a misunderstanding. I will close the upstream bug then.

PR #40759, opened 2022-09-01
add tcp-echo makefile

my-git9:
Add tcp-echo makefile


craigbox:
LGTM, thoughts @frankbu ?

Issue #38562, opened 2022-04-25
meshNetworks weird error: Invalid network number input


apolovov:
### Bug Description

If there is meshNetworks configured in istiod CM:
```
  meshNetworks: |-
    networks:
      network-ice-local-3250915932:
        endpoints:
        - fromRegistry: icebottle
        gateways:
        - address: 1.2.3.4
          port: 30429
```
 There are weird errors in istiod logs:
```
2022-04-25T15:28:30.510018Z	error	kube	Invalid network number input
2022-04-25T15:28:30.510567Z	error	kube	Invalid network number input
2022-04-25T15:28:30.510590Z	error	kube	Invalid network number input
2022-04-25T15:28:30.510595Z	error	kube	Invalid network number input
2022-04-25T15:28:30.510618Z	error	kube	Invalid network number input
2022-04-25T15:28:30.510622Z	error	kube	Invalid network number input
2022-04-25T15:28:30.510626Z	error	kube	Invalid network number input
2022-04-25T15:28:30.510641Z	error	kube	Invalid network number input
2022-04-25T15:28:30.510644Z	error	kube	Invalid network number input
2022-04-25T15:28:30.510649Z	error	kube	Invalid network number input
2022-04-25T15:28:30.514130Z	error	kube	Invalid network number input
```
And everything works fine.
With empty `networks` section there aren't errors.

### Version

```prose
./istioctl -i d8-istio version
client version: 1.12.6
istiod version: 1.10.1
istiod version: 1.13.3
data plane version: 1.10.1 (6 proxies), 1.13.3 (1 proxies)

kubectl version --short
Client Version: v1.21.11
Server Version: v1.21.11
```

PR #40775, opened 2022-09-02
fix weird error message when networks is used

ramaraochavali:
Fixes https://github.com/istio/istio/issues/38562

- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -450,9 +450,13 @@ func (c *Controller) networkFromMeshNetworks(endpointIP string) network.ID {
 	}
 
 	if c.ranger != nil {
-		entries, err := c.ranger.ContainingNetworks(net.ParseIP(endpointIP))
+		ip := net.ParseIP(endpointIP)
+		if ip == nil {
+			return ""
+		}
+		entries, err := c.ranger.ContainingNetworks(ip)
 		if err != nil {
-			log.Error(err)
+			log.Errorf("error getting cidr ranger entry from endpoint ip %s", endpointIP)
 			return ""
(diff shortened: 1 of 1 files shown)


pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -450,9 +450,13 @@ func (c *Controller) networkFromMeshNetworks(endpointIP string) network.ID {
 	}
 
 	if c.ranger != nil {
-		entries, err := c.ranger.ContainingNetworks(net.ParseIP(endpointIP))
+		ip := net.ParseIP(endpointIP)
+		if ip == nil {
+			return ""
```
hzxuzhonghu, line 455: what does this mean?

hzxuzhonghu: when can net.ParseIP(endpointIP) == nil  happen? I mean why it can fix the original issue

ramaraochavali: https://github.com/istio/istio/blob/276fb292f30e3f20cc036fabba6328b76291c781/pilot/pkg/serviceregistry/serviceentry/workloadentry.go#L34 - I guess it can be empty.

ramaraochavali: We can check in that function. Checking it here would handle any such calls in future

hzxuzhonghu: IC, make sense to me

hzxuzhonghu (approved):

PR #40778, opened 2022-09-02
Fix proxyconfig annotation lead to kube-inject panic

hanxiaop:
MeshConfig is nil if injector exists. Set it to default proxyconfig to avoid of nil pointer panic.

### DIFF
--- a/pkg/config/mesh/mesh.go
+++ b/pkg/config/mesh/mesh.go
@@ -141,11 +141,6 @@ func DefaultMeshConfig() *meshconfig.MeshConfig {
 // will not be modified.
 func ApplyProxyConfig(yaml string, meshConfig *meshconfig.MeshConfig) (*meshconfig.MeshConfig, error) {
 	mc := proto.Clone(meshConfig).(*meshconfig.MeshConfig)
-	if mc == nil {
-		mc = &meshconfig.MeshConfig{
-			DefaultConfig: DefaultProxyConfig(),
-		}
-	}
 	pc, err := applyProxyConfig(yaml, mc.DefaultConfig)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


hzxuzhonghu (commented):
can you elaborate more when this can happen

pkg/config/mesh/mesh.go
```diff
@@ -141,6 +141,11 @@ func DefaultMeshConfig() *meshconfig.MeshConfig {
 // will not be modified.
 func ApplyProxyConfig(yaml string, meshConfig *meshconfig.MeshConfig) (*meshconfig.MeshConfig, error) {
 	mc := proto.Clone(meshConfig).(*meshconfig.MeshConfig)
+	if mc == nil {
```
howardjohn, line 144: how does clone return `nil`? is `meshConfig` already nil? If so, seems like the check should be before clone?

Also it seems odd that we will make a empty mesh config. in 99% of places we use DefaultMeshConfig.

This seems like we should be changing the place that _calls_ ApplyProxyConfig, instead of ApplyProxyConfig itself?

hanxiaop:
> can you elaborate more when this can happen

@hzxuzhonghu @howardjohn When using `kube-inject` cmd without providing the injector file, it will use the webhook injector and return nil meshconfig. In this case, when a pod has the annotation `proxy.istio.io/config`, the `ApplyProxyConfig` func will panic with the nil meshconfig. I think John is right, and I just changed the code place instead of revising the `ApplyProxyConfig` itself.

howardjohn (approved):
should we add a test?

howardjohn (approved):

intel-dlanders:
What's the process here? This bug breaks your blog post on [proxyless grpc](https://istio.io/latest/blog/2021/proxyless-grpc/#deploy-the-application). It looks like we're waiting for one reviewer approval. Can we get it?! Please.

irisdingbj (approved):
Thanks for the fix!

GregHanson:
@hanxiaop does this need to be cherrypicked to 1.14 or 1.15 release branches?

Issue #40720, opened 2022-08-30
Istiod starts accepting requests before it enables HTTP2


qfel:
### Bug Description

It appears istiod starts accepting request before it enables HTTP2 support. This causes a (usually short) time window during which HTTP2 requests fail due to protocol mismatch.

 - We start listening in [serveHTTP in NewServer](https://github.com/istio/istio/blob/ddb839267d020c528ace957a33bd7fbf0cd093ac/pilot/pkg/bootstrap/server.go#L1285)
 - Start updates the handler [to support H2](https://github.com/istio/istio/blob/ddb839267d020c528ace957a33bd7fbf0cd093ac/pilot/pkg/bootstrap/server.go#L478)
 - Start is called after NewServer
 - Ergo between NewServer and Start, requests will be accepted but not failed

Note that I've seen this (possibly) happen via HTTP2 webhook requests, even though the code seems to associate this with gRPC. Also the issue may not be relevant in K8s (due to readiness checks) but on platforms where the only signal accepted is the port listener, this can cause an issue.

### Version

```prose
The issue exists at least in master.
```

hzxuzhonghu:
How do you access istiod?  We have a readiness  probe, before it is ready, sidecar can not connect to istiod by k8s service.

qfel:
Cloud Run service (the only readiness it supports is "the port is open")

hzxuzhonghu:
I think there can be case a port is listened without serving any http

PR #40783, opened 2022-09-02
Fix h2 multiplexing race

howardjohn:
Fixes https://github.com/istio/istio/issues/40720

Instead of later swapping the h1 server with a multiplexed server, build
the correct one the first time. This avoids the logical race in the
issue, and a legitimate data race (swap unprotected handler).


ramaraochavali (approved):
LGTM. Lint error seems legit

Issue #40784, opened 2022-09-02
Support Authorization Policy Contains Matching


rrobb-aspen:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**
Authorization Policy currently supports prefix matching and suffix matching on headers in conditionals.  We would like to add support for matching on strings contained within the header when it may not be either a prefix or suffix.

**Describe alternatives you've considered**
Regex and CEL.  Regex has potential security and performance issues, and CEL would require changes to the Envoy API.  See PR https://github.com/istio/api/pull/2470.

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**
This is similar to Issue https://github.com/istio/istio/issues/39680 but it's not quite the same requirements.

rrobb-aspen:
We've taken @costinm's thoughts and the feedback from the community into consideration.  We are hoping to have a solution presented to the community in the on-coming weeks.

PR #40786, opened 2022-09-02
Add new contains match for Authorization Policy

rrobb-aspen:
Adds a contains match to Authorization Policy header matching.

If a user wants to match on part of the header when the string to match is not either the prefix or suffix of the string, user can provide the match as `*<header-string-to-match>*` and these changes will check if the header contains the provided string which has been bounded with `*`. 

This change will replace the previous behavior of headers being bounded with `*` only matching on the suffix.  So instead of `*/productpage*` returning matches for `/productpage*`, it will instead return matches for `/productpage` and suffix matches are more strictly defined as `*/productpage`.

Fix Issue: https://github.com/istio/istio/issues/40784

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [x] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/releasenotes/notes/40784.yaml
+++ b/releasenotes/notes/40784.yaml
@@ -1,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: feature
+area: security
+issue:
+  - 40784
+releaseNotes:
+  - |
+    **Added** contains matching to Authorization Policy HeaderMatcher
--- a/releasenotes/notes/40784.yaml
+++ b/releasenotes/notes/40784.yaml
@@ -5,4 +5,9 @@ issue:
   - 40784
 releaseNotes:
   - |
-    **Added** contains matching to Authorization Policy HeaderMatcher
+    **Added** contains matching to Authorization Policy HeaderMatcher.
+  - |
+    **Improved** behavior where headers that were bracketed by two wildcards 
+    would match as suffix matches with the trailing wildcard being matched 
+    with the string in the header. New behavior will match the string between 
+    two wildcards as a contains match.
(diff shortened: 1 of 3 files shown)


lei-tang (commented):
Quoted from the PR description: "So instead of */productpage* returning matches for /productpage*, it will instead return matches for /productpage and suffix matches are more strictly defined as */productpage." 

The proposed matching method (e.g., "*/productpage*" matches /productpage) has a non-standard interpretation of the regular expression wildcard "*". I suggest following standard regular expression rules for "*", suffix matching, and prefix matching.

releasenotes/notes/40784.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: feature
+area: security
+issue:
+  - 40784
+releaseNotes:
+  - |
+    **Added** contains matching to Authorization Policy HeaderMatcher
```
lei-tang, line 8: The PR describes that this PR changes existing behavior: "This change will replace the previous behavior of headers being bounded with \* only matching on the suffix."

The release notes should warn the changing behavior.

rrobb-aspen: Updated the release notes.

liminw:
As discussed in https://github.com/istio/api/pull/2470, we need a design for this feature before committing the code.

lei-tang:
I agree that this feature needs a design document and a review for the design.

lei-tang:
A do-not-merge label is added to the PR so the PR will not be merged until the design is approved.

jacob-delgado:
> As discussed in [istio/api#2470](https://github.com/istio/api/pull/2470), we need a design for this feature before committing the code.

We are evaluating our options and making sure it passes the CI/CD. We will have a formal doc in the coming weeks detailing our findings. We understand that it can take time to get community feedback before anything is approved, but we would like to understand the pros/cons better before we document this. I'll put up another PR up soon that has a CEL approach (hopefully) to aid in this discussion.

rrobb-aspen:
We've taken the feedback from the community into consideration.  We are hoping to have a solution presented to the community in the on-coming weeks.

PR #40684, opened 2022-08-26
[WIP] Implement "TrafficSelector" of WasmPlugin

ingwonsong:
This PR implements https://github.com/istio/api/pull/2412

PR #38775, opened 2022-05-06
Cherrypick 38702

GregHanson:
**Please provide a description of this PR:**

and a few other dependent PRs:
https://github.com/istio/istio/pull/38499
https://github.com/istio/istio/pull/38595
https://github.com/istio/istio/pull/38652
https://github.com/istio/istio/pull/38702
https://github.com/istio/istio/pull/38524
https://github.com/istio/istio/pull/38690

PR #40787, opened 2022-09-03
Apply WasmPlugin to Outbound traffic by the feature toggle

ingwonsong:
After https://github.com/istio/istio/pull/38595, WasmPlugin filter is injected for both inbound and outbound of sidecar traffic. 

This PR provides a feature toggle `APPLY_WASM_PLUGINS_TO_OUTBOUND` to select if apply WasmPlugin to outbound traffic or just inbound only. 

In case of Gateway, WasmPlugin is applied regardless of this toggle.

### DIFF
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -647,8 +647,9 @@ var (
 
 	SidecarIgnorePort = env.Register("SIDECAR_IGNORE_PORT_IN_HOST_MATCH", true, "If enabled, port will not be used in vhost domain matches.").Get()
 
-	ApplyWasmPluginsToOutbound = env.Register("APPLY_WASM_PLUGINS_TO_OUTBOUND", true, "If enabled, Wasm Plugins are applied to the outbound traffic of sidecars, too."+
-		"In other words, Wasm Plugins are applied to all the traffic directions.").Get()
+	ApplyWasmPluginsToOutbound = env.Register("APPLY_WASM_PLUGINS_TO_OUTBOUND", true,
+		"If enabled, Wasm Plugins are applied to the outbound traffic of sidecars, too."+
+			"In other words, Wasm Plugins are applied to all the traffic directions.").Get()
 )
 
 // EnableEndpointSliceController returns the value of the feature flag and whether it was actually specified.
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -646,6 +646,9 @@ var (
 		"If enabled, the cluster secret watcher will watch the namespace of the external cluster instead of config cluster").Get()
 
 	SidecarIgnorePort = env.Register("SIDECAR_IGNORE_PORT_IN_HOST_MATCH", true, "If enabled, port will not be used in vhost domain matches.").Get()
+
+	ApplyWasmPluginsToOutbound = env.Register("APPLY_WASM_PLUGINS_TO_OUTBOUND", true, "If enabled, Wasm Plugins are applied to the outbound traffic of sidecars, too."+
+		"In other words, Wasm Plugins are applied to all the traffic directions.").Get()
 )
 
 // EnableEndpointSliceController returns the value of the feature flag and whether it was actually specified.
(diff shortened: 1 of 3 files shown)


ingwonsong:
@howardjohn I put the default value of the toggle "true" here to keep the behavior of 1.14.

What do you think?

howardjohn:
> @howardjohn I put the default value of the toggle "true" here to keep the behavior of 1.14.
> 
> What do you think?
I think we want to have the default whatever we want long term as the default (including when we have the match API, etc)... but then we end up with a weird flip-flopping between versions if we decide we only want inbound by default. I'm not sure... certainly in the patch branches we likely want to preserve existing behavior since changing in a patch isn't great? 

Any idea why we haven't heard issues about this? Are WASM users not on impacted versions or they don't notice/mind? 

Did we ship it in 1.14.x or 1.14.0?

ingwonsong:
Interesting question. That was the reason why I thought that there is an intended change in early time.

I have just tested it on 1.14.0 and 1.13.7, and wasm plugin is inserted in both of inbound and outbound from 1.14.0.

According to git logs, it was cherry-picked before 1.14.0-beta.0. (cherrypick PR: https://github.com/istio/istio/pull/38775)

ingwonsong:
If the users do not have serious issues by this, I think it would be better to turn it back to the original behavior. (by setting the flag "false" in default).

howardjohn:
IMO in the 1.14 branch and 1.15 we can't change anything at this point,
since we want to. minimize change in patch release. For 1.16 - which will
come. with the API as well - it's more reasonable to change the default
back to inbound only. if we want?

zirain:
@howardjohn why istio only allow WasmPlugin on outbound?

ingwonsong:
> @howardjohn why istio only allow WasmPlugin on outbound?

TL;DR - WasmPlugin was applied to Gateway and inbound of Sidecar, but it was changed to apply to all traffics from 1.14.
So, this is a patch to give an option to choose the previous behavior in <= 1.13.

ingwonsong:
@howardjohn OK. I agree with you.
Then, at this time, let's keep the behaviors as 1.14 and 1.15 in default. (current PR)

Hopefully, in 1.16, https://github.com/istio/api/pull/2412 and https://github.com/istio/istio/pull/40684 can be there, and we can set the default behavior to use inbound.

ramaraochavali:
Why do we need this feature flag if 1.16 comes with an API? Is it for users of 1.13 and 1.14 to only have for inbound if they want? May be we should do when a need arises, instead of proactively doing this? Just want to see if we can avoid at another feature flag for which api is already being discussed?

ingwonsong:
@ramaraochavali Yes. This is for 1.14 and 1.15 users, who want the original behavior.
and we can keep this until an explicit request is raised.

However, IMHO, in this case, proactive reactions seems to be good because it is not originally intended. 
In addition, I think we need to consider the case that the API extension in discussion cannot be accepted and implemented before 1.16.

ramaraochavali:
> However, IMHO, in this case, proactive reactions seems to be good because it is not originally intended.

Fair point. But we have already changed the behaviour and the default is also not inline with old behaviour.  

In addition, I think we need to consider the case that the API extension in discussion cannot be accepted and implemented before 1.16.

Agree, we need to remove this when API is there

ingwonsong:
>  But we have already changed the behaviour and the default is also not inline with old behaviour.
 Yes true. From 1.14/1.15, the behavior is already changed. But, there seems

@costinm Do you have any thoughts?

ramaraochavali:
> Basically, this patch is for the existing users who used 1.13 or older version.

Ideally we would retain the old behaviour by flipping the default value of flag. If there is a concern in doing this, I would suggest we need to add this to upgrade notes.

ingwonsong:
"hold" because https://github.com/istio/istio/pull/40862 may override this PR.

ingwonsong:
Close this PR by https://github.com/istio/istio/pull/40862

howardjohn:
IMO even with API the flag is still useful. in 1.16 so users can set the
flag in 1.15 and upgrade and have it respected

PR #40804, opened 2022-09-05
Make top level pkg/ default to T&R

howardjohn:
This avoids all of the smaller helper libraries we have been adding
lately requiring top level owner approval. Most of these are generic
utils that seem reasonable to be under T&R. For anything not, they
should be explicily catagorized anyways - not top level.


ericvn (approved):

hzxuzhonghu:
Seems not suitable to be T&R, we should give them finer granularity 
pkg is a general directory, which contains many subdirectories, like adsc/ envoy/dns and also security related.

PR #40805, opened 2022-09-05
Add a new `pkg/sleep` package to sleep safely with cancelation

howardjohn:


pkg/sleep/sleep.go
```diff
@@ -0,0 +1,41 @@
...
+func Until(ch <-chan struct{}, d time.Duration) bool {
+	timer := time.NewTimer(d)
+	select {
+	case <-ch:
+		if !timer.Stop() {
+			<-timer.C
+		}
+		return false
+	case <-timer.C:
+		return true
```
hzxuzhonghu, line 39: You need to stop the timer before return

pkg/sleep/sleep.go
```diff
@@ -0,0 +1,41 @@
...
+package sleep
+
+import (
+	"context"
+	"time"
+)
+
+// Context sleeps for the given duration, or until the context is complete.
+// Returns true if the sleep completes the full duration
+func Context(ctx context.Context, d time.Duration) bool {
```
hzxuzhonghu, line 24: This name is confusing, indicates it returns a context.

cni/pkg/install/install.go
```diff
@@ -297,7 +298,7 @@ func (in *Installer) watchSAToken(ctx context.Context, fileModified chan bool, e
 				if curToken != token {
 					fileModified <- true
 				}
-				time.Sleep(1 * time.Minute)
+				sleep.Context(ctx, 1 * time.Minute)
```
hzxuzhonghu, line 301: Do we still need L291?

pkg/sleep/sleep.go
howardjohn: Any suggestions? When you call it's `sleep.Context()` so it was pretty clear to me.

cni/pkg/install/install.go
howardjohn: it protects the first call I guess

pkg/sleep/sleep.go
hzxuzhonghu: How about removing this function. ctx.Done() is simple to call

ramaraochavali: Either remove or name `UntilContext` ?

ramaraochavali: UntilContext may be better?

hzxuzhonghu (approved):

ericvn (approved):

Issue #40767, opened 2022-09-01
Support `regex` on the `Sidecar::IstioEgressListener::hosts`


qudongfang:
**Describe the feature request**

Support `regex` on the [`Sidecar::IstioEgressListener::hosts`](https://istio.io/latest/docs/reference/config/networking/sidecar/#IstioEgressListener).

Current state:
> One or more service hosts exposed by the listener in namespace/dnsName format.

> The dnsName should be specified using FQDN format, optionally including a wildcard character in the left-most component (e.g., prod/*.example.com). Set the dnsName to * to select all services from the specified namespace (e.g., prod/*).

Use case: Our ServiceEntry host format is something like `${serviceName}-${protocol}.${namespace}.${region}.${provider}.kube`(Multi provider & Multi Region), and we want to setup `Sidecar` per `Service/Deployment` to let `istio-proxy` setup the listeners of required `Service`s only.

It would be great to have the following `Sidecar` config working.

```
apiVersion: networking.istio.io/v1beta1
kind: Sidecar
metadata:
  name: echo-sidecar
  namespace: echo-dev
spec:
  egress:
  - hosts:
    - 'hello-namespace/hello-*.gcp.kube'
  workloadSelector:
    labels:
      app.kubernetes.io/instance: echo
      app.kubernetes.io/name: echo
```

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

howardjohn:
I don't think we should do this. I understand your specific use case but I don't think it's a reasonable API.

howardjohn:
Prefix is already supported. Suffix matching hostnames is not a part of most Istio APIs and is a bit backwards from how domains work

PR #40814, opened 2022-09-06
fix:github.com/istio/istio/issues/40767

cobb-tx:
Signed-off-by: cobb-tx <464693850@qq.com>

fix：https://github.com/istio/istio/issues/40767

### DIFF
--- a/pkg/config/host/name.go
+++ b/pkg/config/host/name.go
@@ -15,6 +15,7 @@
 package host
 
 import (
+	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/config/host/name.go
```diff
@@ -82,8 +83,10 @@ func (n Name) SubsetOf(o Name) bool {
 		return strings.HasSuffix(string(n), string(o[1:]))
 	}
 
-	// both are non-wildcards, so do normal string comparison
-	return n == o
+	expr := strings.Replace(o.String(), "*", "(.*)", -1)
+	reg, _ := regexp.Compile(expr)
+
+	return reg.MatchString(n.String()) || n == o
```
hzxuzhonghu, line 89: There is no good way to judge regex subset. This is not correct

cobb-tx:
Does not support this

PR #40831, opened 2022-09-06
sets: add `InsertContains` to handle common dedupe cases

howardjohn:
it is a common pattern to build a set as we iterate, to check for
duplicates. This introduces a new function to ease this pattern.
No behavior change expected, just simpler code.

### DIFF
--- a/pkg/util/sets/string.go
+++ b/pkg/util/sets/string.go
@@ -150,6 +150,20 @@ func (s Set) SortedList() []string {
 	return res
 }
 
+// InsertContains inserts the item into the set and returns if it was already present.
+// Example:
+//
+//		if !set.InsertContains(item) {
+//			fmt.Println("Added item for the first time", item)
+//	  }
+func (s Set) InsertContains(item string) bool {
+	if s.Contains(item) {
+		return true
+	}
+	s[item] = struct{}{}
+	return false
+}
+
 // Contains returns whether the given item is in the set.
 func (s Set) Contains(item string) bool {
 	_, ok := s[item]
--- a/tests/integration/telemetry/stackdriver/common.go
+++ b/tests/integration/telemetry/stackdriver/common.go
@@ -359,10 +359,9 @@ func logDiff(t test.Failer, tp string, query map[string]string, entries []map[st
 	for _, s := range entries {
 		b, _ := json.Marshal(s)
 		ss := string(b)
-		if seen.Contains(ss) {
+		if seen.InsertContains(ss) {
 			continue
 		}
-		seen.Insert(ss)
 		misMatched := map[string]string{}
 		for k, want := range query {
 			got := s[k]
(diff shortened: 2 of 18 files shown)


pkg/util/sets/string.go
```diff
@@ -150,6 +150,20 @@ func (s Set) SortedList() []string {
 	return res
 }
 
+// InsertContains inserts the item into the set and returns if it was already present.
+// Example:
+//
+//		if !set.InsertContains(item) {
+//			fmt.Println("Added item for the first time", item)
+//	  }
+func (s Set) InsertContains(item string) bool {
```
douglas-reid, line 159: any reason to not just have this be `Insert` with the same contract ? 

```
added := set.Insert(a)
if added {
...
}
```

douglas-reid: oh, i see. we used `Insert` already with a fluent-style return.

howardjohn: yeah in theory we could switch it still and change current usages. However, I do worry slightly about if the contains() check is too slow for some use cases - some of the set usages are in hot paths

douglas-reid: i'd leave the existing `Insert` alone. No sense in confusing those that already know and use the package. I wish there was a pithier way of expressing the concept, but `CondInsert`, `TryInsert`, or even `MaybeInsert` don't seem any better to me. And I'm blanking on better ideas at the moment.

tests/integration/telemetry/stackdriver/common.go
```diff
@@ -359,10 +359,9 @@ func logDiff(t test.Failer, tp string, query map[string]string, entries []map[st
 	for _, s := range entries {
 		b, _ := json.Marshal(s)
 		ss := string(b)
-		if seen.Contains(ss) {
+		if seen.InsertContains(ss) {
```
douglas-reid, line 362: nit: this is why i was trying to brainstorm naming -- here the code is clearer (imho) as written previously; even tho the collection is not modified in the new example, that takes a second to grok with the new name (in this particular setting).

douglas-reid (approved):
approving telemetry owned files.

ericvn (approved):
Approve the test related files

pkg/util/sets/string.go
ramaraochavali: nit: General name for this type of operation in other languages like Java is `AddIfAbsent` or `InsertIfAbsent`

howardjohn: Its not quite the same IMO. Since it is a set there is no difference between just adding vs checking. The important part is the return.

ramaraochavali: correct. InsertIfAbsent typically returns bool to say if it is added or not that indicates whether it is contained or not. https://www.geeksforgeeks.org/copyonwritearraylist-addifabsent-method-in-java/ but on list. 

But NBD, the current one is fine

ramaraochavali (approved):

lei-tang (approved):

PR #40836, opened 2022-09-06
Validate proxy more in fuzz tests

howardjohn:


hzxuzhonghu (commented):
@howardjohn tests/fuzz/testdata/FuzzValidateClusters/6380814220263424 added by mistake?

howardjohn:
> @howardjohn tests/fuzz/testdata/FuzzValidateClusters/6380814220263424 added by mistake?

no, it's a regression test

hzxuzhonghu:
I remember it is auto generated to cache some state

howardjohn:
the fuzzer auto generates test cases theb when we hit failures we check in
the failed case as a regression test and fix the failure

hzxuzhonghu:
I am wondering this can not be called fuzz, isn't it fixed test cases then

howardjohn:
We run a few fixed test cases in unit test for known regression tests, then
continuously we are generating new cases. The checked in ones are just a
small check to verify things compile and run properly and we don't regress
- most of the work is on the real fuzzing (which is how we found the bug)

hzxuzhonghu:
That make sense

hzxuzhonghu (approved):

ericvn (approved):

PR #40840, opened 2022-09-06
Add validations for workload selector

varks:
Istio CRDs like EF, SE, Sidecar allows specifying workloadSelector without labels - which could be unintentional. The present behavior is to apply CRDs to all proxies in the namespace. This PR does not change that behavior, but adds a warning if a workload selector is specified without the labels.

### DIFF
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -798,6 +798,12 @@ var ValidateEnvoyFilter = registerValidateFunc("ValidateEnvoyFilter",
 			return nil, err
 		}
 
+		// If workloadSelector is defined and labels are not set, it is most likely
+		// an user error. Marking it as a warning to keep it backwards compatible.
+		if rule.WorkloadSelector != nil && len(rule.WorkloadSelector.Labels) == 0 {
+			errs = appendValidation(errs, WrapWarning(fmt.Errorf("Envoy filter: labels not defined in workload selector"))) // nolint: stylecheck
+		}
+
 		for _, cp := range rule.ConfigPatches {
 			if cp == nil {
 				errs = appendValidation(errs, fmt.Errorf("Envoy filter: null config patch")) // nolint: stylecheck
--- a/pkg/config/validation/validation.go
+++ b/pkg/config/validation/validation.go
@@ -767,8 +767,10 @@ func validateExportTo(namespace string, exportTo []string, isServiceEntry bool, 
 	return
 }
 
-func validateAlphaWorkloadSelector(selector *networking.WorkloadSelector) error {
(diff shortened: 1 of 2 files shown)


pkg/config/validation/validation.go
```diff
@@ -798,6 +798,12 @@ var ValidateEnvoyFilter = registerValidateFunc("ValidateEnvoyFilter",
 			return nil, err
 		}
 
+		// If workloadSelector is defined and labels are not set, it is most likely
+		// an user error. Marking it as a warning to keep it backwards compatible.
+		if rule.WorkloadSelector != nil && len(rule.WorkloadSelector.Labels) == 0 {
+			errs = appendValidation(errs, WrapWarning(fmt.Errorf("Envoy filter: labels not defined in workload selector"))) // nolint: stylecheck
```
ramaraochavali, line 804: nit: change the message as 
"Envoy filter:  workload selector specified with out labels. This will filter will be applied to services in the namespace"

ramaraochavali (commented):
LGTM. @hzxuzhonghu @howardjohn any concerns with this?

pkg/config/validation/validation.go
```diff
@@ -798,6 +798,12 @@ var ValidateEnvoyFilter = registerValidateFunc("ValidateEnvoyFilter",
 			return nil, err
 		}
 
+		// If workloadSelector is defined and labels are not set, it is most likely
+		// an user error. Marking it as a warning to keep it backwards compatible.
+		if rule.WorkloadSelector != nil && len(rule.WorkloadSelector.Labels) == 0 {
```
hzxuzhonghu, line 803: I think this can be checked in `validateAlphaWorkloadSelector` to apply to all resources

ramaraochavali: good point

varks: @hzxuzhonghu I did consider that initially. But the above function does not pass back warning, but would need additional re-factoring.

varks: @hzxuzhonghu @ramaraochavali was able to refactor it. Ptal when you get a chance.

pkg/config/validation/validation.go
```diff
@@ -767,8 +767,10 @@ func validateExportTo(namespace string, exportTo []string, isServiceEntry bool,
 	return
 }
 
-func validateAlphaWorkloadSelector(selector *networking.WorkloadSelector) error {
+func validateAlphaWorkloadSelector(selector *networking.WorkloadSelector) (Warning, error) {
```
hzxuzhonghu, line 770: nit： return Validation

hzxuzhonghu (approved):

GregHanson:
since this adds a warning validation message versus error message, should be non-breaking and safe to back port

PR #40841, opened 2022-09-07
docker: fix check for buildx instance

howardjohn:
This was incorrectly parsing before; simplify with new command


ericvn (approved):

keithmattix:
The regex here is fragile; `docker buildx inspect` prints a tab [character](https://github.com/docker/buildx/blob/master/commands/inspect.go#L52), but the regex pattern only matches a single space

Issue #39445, opened 2022-06-14
Support failover cluster in DestinationRule


shonecyx:
This is related with https://github.com/istio/istio/issues/38105 for the failover feature.
Currently we only support locality based failover but we would like to extend this feature to support failover to a specified cluster for a given route. This is very helpful if the destination upstream cluster is down and we can failover to a default service, for example the homepage service.

Since VirtualService is not the right place, I would re-propose DestinationRule.

**Option 1:Add defaultService in the failover field.**
```
apiVersion: networking.istio.io/v1beta1
kind: DestinationRule
metadata:
  name: helloworld
spec:
  host: helloworld.sample.svc.cluster.local
  trafficPolicy:
    loadBalancer:
      simple: ROUND_ROBIN
      localityLbSetting:
        enabled: true
        failover:
          - defaultService: homepage.sample.svc.cluster.local
 ```

**Option 2: Using Envoy Aggregate Cluster with a large enough overprovisioning_factor value can archive the fallback effect.**

Looks like defining the failover destination in DestinationRule's [LoadBalancerSettings](https://istio.io/latest/docs/reference/config/networking/destination-rule/#LoadBalancerSettings) could be a better choice:
```
apiVersion: networking.istio.io/v1beta1
kind: DestinationRule
metadata:
  name: helloworld
spec:
  host: helloworld.sample.svc.cluster.local
  trafficPolicy:
    loadBalancer:
      simple: ROUND_ROBIN
      failover:
        failoverFactor: 5 # when healthy endpoint lower than this percentage, start to failover;
                          # this is used to calculate Aggregate Cluster's primary cluster's overprovisioning_factor
        destination:
          host: homepage.sample.svc.cluster.local
          subset: v2
          port:
            number: 8080
```
When failover is defined, the outbound cluster can be generated as an Envoy Aggregate Cluster, with native destination as primary cluster, and the failover destination as Secondary cluster.

Actually both options leverage the envoy priority level but implemented through different spec.

[ ] Docs
[ ] Installation
[ x] Networking
[ ] Performance and Scalability
[ ] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

shonecyx:
@howardjohn  any comment?

shonecyx:
[...]

PR #40847, opened 2022-09-07
support failover service in DestinationRule

shonecyx:
Try to implement https://github.com/istio/istio/issues/39445

Introduced 3 annotations in DestinationRule and Leverage envoy[Priority levels](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/priority#priority-levels) to achieve the failover.

With failover service specified, eds will merge the upstream service endpoints with the failover service endpoints and do traffic failover based on the overprovisioning factor.

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/endpoint_builder.go
+++ b/pilot/pkg/xds/endpoint_builder.go
@@ -17,6 +17,7 @@ package xds
 import (
 	"crypto/md5"
 	"encoding/hex"
+	"math"
 	"sort"
 	"strconv"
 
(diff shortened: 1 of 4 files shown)


pilot/pkg/xds/endpoint_builder.go
```diff
@@ -230,9 +231,68 @@ func (b *EndpointBuilder) populateFailoverPriorityLabels() {
...
 func (b *EndpointBuilder) buildLocalityLbEndpointsFromShards(
 	shards *model.EndpointShards,
+	failoverShards *model.EndpointShards,
 	svcPort *model.Port,
+	failOverPort uint32,
 ) []*LocalityEndpoints {
 	localityEpMap := make(map[string]*LocalityEndpoints)
+	locEps := make([]*LocalityEndpoints, 0, len(localityEpMap))
+
+	if shards != nil {
```
howardjohn, line 241: We can't just swap out endpoints for a cluster. Endpoints are tightly bound to Clusters. If we failover we would now have random endpoints from some other service, but still the metadata, SAN matching, etc from the original Service. Additionally ,they may be entirely different types (DNS vs STATIC, etc)

howardjohn (changes requested):
Thanks for the initiative. I think this needs a thorough design and API review

Issue #40854, opened 2022-09-07
Istio creates an invalid inbound listener when InterceptionMode=NONE


alandiegosantos:
### Bug Description

When Istio is configured with **InterceptionMode** to **NONE**, a listener is created on Envoy configuration with an empty cluster. In addition, the port used by listener conflicts with the port used by the application, causing istio-proxy to reject the configuration.

## How to reproduce:
1. Apply the following manifests:
```
kubectl apply -f - << EOF
apiVersion: v1
kind: Namespace
metadata:
  name: echoserver
  labels:
    istio-injection: enabled
---
apiVersion: v1
kind: Service
metadata:
  name: echoserver
  namespace: echoserver
spec:
  selector:
    app: echoserver
  ports:
  - protocol: TCP
    port: 8080
    targetPort: 8080
    name: http
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: echoserver
  namespace: echoserver
spec:
  selector:
    matchLabels:
      app: echoserver
  replicas: 1 # tells deployment to run 2 pods matching the template
  template:
    metadata:
      labels:
        app: echoserver
      annotations:
        sidecar.istio.io/interceptionMode: NONE
    spec:
      containers:
      - name: echoserver
        image: k8s.gcr.io/echoserver:1.4
        ports:
        - containerPort: 8080
EOF
```
2. istio-proxy rejects the configuration:
```
$ kubectl logs -n echoserver $(kubectl get pods -l app=echoserver -o custom-columns=:metadata.name --no-headers -n echoserver | head -n 1) -c istio-proxy
2022-09-07T15:20:40.328609Z	info	cache	generated new workload certificate	latency=109.405105ms ttl=23h59m59.671397866s
2022-09-07T15:20:40.328747Z	info	cache	Root cert has changed, start rotating root cert
2022-09-07T15:20:40.328801Z	info	ads	XDS: Incremental Pushing:0 ConnectedEndpoints:0 Version:
2022-09-07T15:20:40.328844Z	info	cache	returned workload trust anchor from cache	ttl=23h59m59.671157905s
2022-09-07T15:20:40.336381Z	info	xdsproxy	connected to upstream XDS server: istiod.istio-system.svc:15012
2022-09-07T15:20:40.350754Z	info	ads	ADS: new connection for node:echoserver-7f696dc84c-trpng.echoserver-1
[...]

PR #40855, opened 2022-09-07
Do not add the inbound listener if the InterceptionMode is NONE 

alandiegosantos:
Fix the bug described in #40854

This PR changes [buildInboundChainConfigs()](pilot/pkg/networking/core/v1alpha3/listener_inbound.go) to not add any inbound listener if the interception mode is NONE and there is no ingress listener defined in the sidecar.

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/listener_inbound.go
+++ b/pilot/pkg/networking/core/v1alpha3/listener_inbound.go
@@ -262,6 +262,14 @@ func (lb *ListenerBuilder) buildInboundChainConfigs() []inboundChainConfig {
 	chainsByPort := make(map[uint32]inboundChainConfig)
 	// No user supplied sidecar scope or the user supplied one has no ingress listeners.
 	if !lb.node.SidecarScope.HasIngressListener() {
+
+		// We should not create inbound listeners in NONE mode based on the service instances
+		// Doing so will prevent the workloads from starting as they would be listening on the same port
+		// Users are required to provide the sidecar config to define the inbound listeners
+		if lb.node.GetInterceptionMode() == model.InterceptionNone {
+			return nil
+		}
+
 		// We will look at all Services that apply to this proxy and build chains for each distinct port.
 		// Note: this does mean that we may have multiple Services applying to the same port, which introduces a conflict
 		for _, i := range lb.node.ServiceInstances {


pilot/pkg/networking/core/v1alpha3/listener_inbound.go
```diff
@@ -262,6 +262,14 @@ func (lb *ListenerBuilder) buildInboundChainConfigs() []inboundChainConfig {
 	chainsByPort := make(map[uint32]inboundChainConfig)
 	// No user supplied sidecar scope or the user supplied one has no ingress listeners.
 	if !lb.node.SidecarScope.HasIngressListener() {
+
+		// We should not create inbound listeners in NONE mode based on the service instances
+		// Doing so will prevent the workloads from starting as they would be listening on the same port
+		// Users are required to provide the sidecar config to define the inbound listeners
```
ramaraochavali, line 268: What about outbound listeners? We create bind to port with [true](https://github.com/istio/istio/blob/dab79d8e213b3e0c187ffe22b7608643912f2588/pilot/pkg/networking/core/v1alpha3/listener.go#L320), would it also conflict?

alandiegosantos: Outbound listeners bind to localhost interface (127.0.0.1). 

When the interception mode is NONE, an outbound listener is created matching the port(s) of each registered service. So, it is possible that the outbound listener tries to bind to the same port used by the application. 

IMO, the disabled interception mode requires other controls and a heavily use of the Sidecar and Envoy Filter resource.

ramaraochavali: To me interception mode is 'None' not well tested. Should we deprecate this? @hzxuzhonghu @howardjohn  - not sure if any one uses it with these bugs?

alandiegosantos: @ramaraochavali I totally agree that the InterceptionMode NONE is not fully tested but I think deprecating it is not a good strategy. In my company, we are designing a mesh to handle > 3Mrps using the InterceptionMode NONE.

In this case, we allow administrators to define what works best for them and support multiple use-cases.

ramaraochavali: The reason why I was thinking of deprecating is , we have not announced it in the doc https://github.com/istio/api/blob/master/annotation/annotations.yaml#L158.
If we want to keep it, we should add more tests around that update the docs. Otherwise, we might hit surprising behaviours.

howardjohn: It would be best to make it more tested and documented rather than deprecated probably.. it is fairly used.

I guess my general concern with this change is we are making a big behavioral change. So what is the impact on existing users? Is EVERYONE using this mode using explicit Sidecar.ingress config, so this doesn't really matter (the only change is to make non-Sidecar case go from NACK -> empty config?)?

ramaraochavali: > explicit Sidecar.ingress config

Sidecar.ingress already has capture Mode and that is what we use. Not sure, if some one uses sidecarIngress config and additionally use this annotation to set to None. What is the benefit of both ?

alandiegosantos: By adding the annotation `sidecar.istio.io/interceptionMode: NONE`, the injection webhook does not add the istio-init container that creates the interception rules. In our setup, it is not an issue because we inject the sidecar using other tools.

Setting the captureMode in the Sidecar resource, among other things, sets the bindToPort of Envoy listeners. When looking only on the perspective of traffic routing, it is possible to achieve the same setup by setting the annotation  `sidecar.istio.io/interceptionMode: NONE` or setting the captureMode in the Sidecar resource.

Given that, if we can disable the injection of the istio-init container by the webhook, I do not see any issue in setting the captureMode in the Sidecar resource and deprecating the annotation `sidecar.istio.io/interceptionMode: NONE`.

hzxuzhonghu (approved):
LGTM, not complete because for outbound, we also need to do some work

alandiegosantos:
@hzxuzhonghu I agree that this PR is not complete because the outbound. I created #40945 to track the progress related to the outbound listeners.

howardjohn:
I am not sure my concern is https://github.com/istio/istio/pull/40855#discussion_r967269807 was responded to. Are we 100% certain this will not break users?

hzxuzhonghu:
@howardjohn I think this patch does not violate your concern, for inbound without ingress listener specified now, if we generate a listener with bindToPort set, there will be port conflict, like this `0.0.0.0_8080: cannot bind '0.0.0.0:8080': Address already in use`

howardjohn:
I think that's probably true I am just not super familiar with all the use
cases for NONE so wanted to make sure we are certain

howardjohn:
for example maybe the app binds to a different IP like localhost (although
I don't think that works now for other reasons) possibly...?

hzxuzhonghu:
I think not possible, because the listener for `None` we generated bind address is 0.0.0.0

Issue #40824, opened 2022-09-06
TestXdsProxyStatus test flake


howardjohn:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/40808/integ-pilot-istiodremote_istio/1567164950667857920

Conerning log: https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/40808/integ-pilot-istiodremote_istio/1567164950667857920/artifacts/pilot-83229990ba474abda24d3c607/TestXdsProxyStatus/_test_context/istio-state2624545497/externalistiod-1/istiod-5cf7f9bd6b-tdwtr_discovery.log seems we have some endless loop with no backoff maybe?

ingwonsong:
seems to be resolved. close.

PR #40856, opened 2022-09-07
adsc: Skip to ack for debug type responses 

ingwonsong:
Regarding https://github.com/istio/istio/issues/40824, to prevent the infinite "ack" loop, 
this PR proposes to skip to "ack" for the debug type response.

Please note that, currently, the debug type is handled with the special way in `pilot`, so `pilot` generates responses  regardless of whether it is a request or ack.

Alternative way is that `pilot` does not send any responses if "nonce" is prefixed by `pilot` own way, because ADSC generates a nonce for a request (not ack) with a form of "date/time".


ingwonsong:
cc: @howardjohn

howardjohn (approved):

Issue #39345, opened 2022-06-08
WasmPlugin to scope to specific routes


ramaraochavali:
It seems WASMPlugin currently is not scoped to inbound/outbound or even a port. When we define a WASMPlugin with a workload selector, the WASM config is applied to all listeners both inbound and outbound.

Is there a way to control this and scope to specific routes/listeners?

ramaraochavali:
@bianpengyuan any thoughts?

zirain:
prefer to reuse match in `EnvoyFilter` cc @ingwonsong @douglas-reid

ramaraochavali:
Yes. That is what I was thinking - we should reuse applicable matches in Envoy filter and do this.

mathetake:
Also routes can be obtained via :path header so this could be implemented inside the Wasm plugin

zirain:
`EnvoyFilter` selector route by `route_name`, but it's optional, we may need a better way?

ingwonsong:
I took a look and re-thought.

Using `match` of `EnvoryFilter` itself seems to be conflict with the current design of WasmPlugin.

In `WasmPlugin`, there is a notion of `phase`, which determines where the wasm plugin should be inserted into.
I think that `phase` is an abstraction as an effort to simplify (and restrict) the configuration.

If we use `match` of `EnvoyFilter` to insert the wasm plugin as an http filter (without `phase`), we need to include the match expression below:
```
          filter:
            name: "envoy.filters.network.http_connection_manager"
            subFilter:
              name: "envoy.filters.http.router"
```

As we can see, `phase` is much more handy than `match` of `EnvoyFilter` because `phase` does not require to know `"envoy.filters.network.http_connection_manager"` or the other http filters to define the location where the injection happen.

Therefore, I would like to suggest to create new match expressions which is similar with EnvoyFilter, but well-fitted to `WasmPlugin` instead of re-use `EnvoyFilter`.

WDYT?

zirain:
> Therefore, I would like to suggest to create new match expressions which is similar with EnvoyFilter, but well-fitted to WasmPlugin instead of re-use EnvoyFilter.

+1 `match` in `EnvoyFilter` right now can not handler route selector without name, the new design should cover this scenario.

zirain:
linked same requirements in `Telemetry`, https://github.com/istio/istio/issues/38271

ingwonsong:
@zirain AFAIK, there is no `per-route` Wasm filter in Envoy. 

[...]

Issue #36238, opened 2021-11-24
wasmplugin doesn't support for SIDECAR_OUTBOUND


Adol1111:
### Bug Description

Test WasmPlugin in 1.12.0
```yaml
apiVersion: extensions.istio.io/v1alpha1
kind: WasmPlugin
metadata:
  name: custom-filter
  namespace: default
spec:
  url: oci://harbor-host/test/custom-filter:0.1
```

```bash
$ istioctl pc listener demo-78cfb7f5cf-zj4dz -o yaml
```

All custom-filter found in inbound filters,  no outbound filters
```
httpFilters:
- configDiscovery:
     configSource:
       ads: {}
       initialFetchTimeout: 0s
       resourceApiVersion: V3
    typeUrls:
    - type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm
  name: default.custom-filter
....
routeConfig:
  name: inbound|2222||
```

---
Then I use envoyfilter to append a wasm filter

```yaml
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: custom-filter
spec:
  configPatches:
  - applyTo: HTTP_FILTER
    match:
      context: ANY
      listener:
        portNumber: 2222
    patch:
      operation: INSERT_BEFORE
      value:
        name: custom-filter
        config_discovery:
          config_source:
            ads: {}
            initial_fetch_timeout: 0s # wait indefinitely to prevent filter chain being disabled
          type_urls: [ "type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm"]
---
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: custom-filter-config
spec:
  configPatches:
  - applyTo: EXTENSION_CONFIG
    match:
      context: ANY
    patch:
      operation: ADD
      value:
        name: custom-filter
        typed_config:
          "@type": type.googleapis.com/udpa.type.v1.TypedStruct
          type_url: type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm
          value:
            config:
              vm_config:
                vm_id: custom-filter_vm
                runtime: envoy.wasm.runtime.v8
                code:
                  remote:
                    http_uri:
                      uri: http://test-host/filter.wasm
                      timeout: 10s
              configuration:
                '@type': type.googleapis.com/google.protobuf.StringValue
                value: |
                  {}
```

outbound
``` yaml
....
- accessLog:
      ....
[...]

PR #40862, opened 2022-09-07
wasm: Implement "TrafficSelector" of WasmPlugin

ingwonsong:
This PR implements https://github.com/istio/api/pull/2412

With the PR for API, this PR fixes https://github.com/istio/istio/issues/39345 and https://github.com/istio/istio/issues/36238.

For the details and discussion, please refer to the [draft design doc](https://docs.google.com/document/d/1qknaZqXItogEmROVnQD61144pEW1X47bEQraI5D52uY/edit?usp=sharing&resourcekey=0-wHAgrakvVaMcnBm40bKuSg), too.

### DIFF
--- a/pilot/pkg/model/extensions.go
+++ b/pilot/pkg/model/extensions.go
@@ -98,7 +98,7 @@ func (p *WasmPluginWrapper) MatchListener(proxyLabels map[string]string, li *Was
 }
 
 type WasmPluginListenerInfo struct {
-	Port  Port
+	Port  int
 	Class istionetworking.ListenerClass
 }
 
@@ -122,13 +122,13 @@ func matchMode(workloadMode typeapi.WorkloadMode, class istionetworking.Listener
 	return false
 }
 
-func matchPorts(portSelectors []*typeapi.PortSelector, port Port) bool {
+func matchPorts(portSelectors []*typeapi.PortSelector, port int) bool {
 	if len(portSelectors) == 0 {
 		// If there is no specified port, match with all the ports.
 		return true
 	}
 	for _, ps := range portSelectors {
-		if ps.GetNumber() != 0 && ps.GetNumber() == uint32(port.Port) {
+		if ps.GetNumber() != 0 && ps.GetNumber() == uint32(port) {
 			return true
 		}
 	}
--- a/pilot/pkg/model/extensions.go
+++ b/pilot/pkg/model/extensions.go
@@ -92,7 +92,7 @@ type WasmPluginWrapper struct {
 	WasmExtensionConfig *envoyWasmFilterV3.Wasm
 }
 
-func (p *WasmPluginWrapper) ShouldApplyTo(proxyLabels map[string]string, li *WasmPluginListenerInfo) bool {
+func (p *WasmPluginWrapper) MatchListener(proxyLabels map[string]string, li *WasmPluginListenerInfo) bool {
 	workloadMatch := (p.Selector == nil || labels.Instance(p.Selector.MatchLabels).SubsetOf(proxyLabels))
 	return workloadMatch && matchTrafficSelectors(p.Match, li)
 }
--- a/pilot/pkg/model/extensions.go
+++ b/pilot/pkg/model/extensions.go
@@ -27,9 +27,12 @@ import (
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
 	extensions "istio.io/api/extensions/v1alpha1"
+	typeapi "istio.io/api/type/v1beta1"
 	"istio.io/istio/pilot/pkg/model/credentials"
+	istionetworking "istio.io/istio/pilot/pkg/networking"
 	"istio.io/istio/pilot/pkg/util/protoconv"
 	"istio.io/istio/pkg/config"
+	"istio.io/istio/pkg/config/labels"
 	"istio.io/istio/pkg/util/protomarshal"
 )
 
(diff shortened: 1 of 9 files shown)


pilot/pkg/model/extensions.go
```diff
@@ -56,6 +92,49 @@ type WasmPluginWrapper struct {
 	WasmExtensionConfig *envoyWasmFilterV3.Wasm
 }
 
+func (p *WasmPluginWrapper) ShouldApplyTo(proxyLabels map[string]string, li *WasmPluginListenerInfo) bool {
```
zirain, line 95: I do think the `ShouldApplyTo` is good name for this, prefer to something like `Match` or `MatchListener`?

ingwonsong: MatchListener seems to be a better name.
Changed.

ingwonsong: Let's go with `MatchListener`. It seems to be more consistent with the names of the sub-functions.

Done.

zirain (approved):
lgtm

pilot/pkg/networking/core/v1alpha3/listener.go
```diff
@@ -1348,6 +1349,7 @@ func (ml *MutableListener) build(builder *ListenerBuilder, opts buildListenerOpt
 			if len(opt.httpOpts.statPrefix) == 0 {
 				opt.httpOpts.statPrefix = strings.ToLower(ml.Listener.TrafficDirection.String()) + "_" + ml.Listener.Name
 			}
+			opt.httpOpts.port = *opts.port
```
hzxuzhonghu, line 1352: This only works for outbound listener, you need to set port for in bound in function `buildSidecarInboundHTTPOpts`

pilot/pkg/model/extensions.go
```diff
@@ -56,6 +92,49 @@ type WasmPluginWrapper struct {
...
+	return false
+}
+
+func matchPorts(portSelectors []*typeapi.PortSelector, port Port) bool {
+	if len(portSelectors) == 0 {
+		// If there is no specified port, match with all the ports.
+		return true
+	}
+	for _, ps := range portSelectors {
+		if ps.GetNumber() != 0 && ps.GetNumber() == uint32(port.Port) {
```
hzxuzhonghu, line 131: here we only compare the port number, so we do not all other infos in model.Port

pilot/pkg/model/push_context_test.go
```diff
@@ -629,6 +642,34 @@ func TestWasmPlugins(t *testing.T) {
 				},
 			},
 		},
+		{
+			// Detailed tests regarding TrafficSelector are in extension_test.go
+			// Just test the integrity here.
+			name: "port-match",
```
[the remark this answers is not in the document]
ingwonsong: This testcase deals with both of the "not-matched" and "matched" cases.
To avoid ambiguity, I added the comments.

pilot/pkg/networking/core/v1alpha3/listener.go
ingwonsong: For the inbound case, please look at https://github.com/istio/istio/pull/40862/files#diff-4afa570feba990275e28fa8ca0b5be122e9ff5d6f0beee9d2e8f54cb57d46613R650

pilot/pkg/model/extensions.go
ingwonsong: I just wanted to leave some space for the future extension, but I agree with your comment.
Let's keep it simple.

All the ports in this PR are now using a `int` type instead of `Port struct`.

hzxuzhonghu (approved):

gooooogle5:
Will the fix be backported to 1.12?

zirain:
> Will the fix be backported to 1.12?

1.12 is EOL

LugiaChang:
> > Will the fix be backported to 1.12?
> 
> 1.12 is EOL

Ok, we need to upgrade to 1.13 soon. I expect the fix will be backported to 1.13, now we can only use wasm by envoyfilter in outbound.

howardjohn:
It's not a "bug fix",it's a new API. It won't be backported even to 1.15.

releasenotes/notes/wasm-traffic-selector.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: feature
+area: extensibility
+issue: [39345]
+releaseNotes:
+  - |
+    **Added** Add the `match` field in the WasmPlugin API. With this `match` clause, a WasmPlugin can be applied to more specific traffic (e.g., traffic to a specific port).
```
GregHanson, line 7:
```suggestion
    **Added** the `match` field in the WasmPlugin API. With this `match` clause, a WasmPlugin can be applied to more specific traffic (e.g., traffic to a specific port).
```

pilot/pkg/networking/core/v1alpha3/listener_builder.go
```diff
@@ -354,7 +354,10 @@ func (lb *ListenerBuilder) buildHTTPConnectionManager(httpOpts *httpListenerOpts
 	routerFilterCtx, reqIDExtensionCtx := configureTracing(lb.push, lb.node, connectionManager, httpOpts.class)
 
 	filters := []*hcm.HttpFilter{}
-	wasm := lb.push.WasmPlugins(lb.node)
+	wasm := lb.push.WasmPluginsByListenerInfo(lb.node, &model.WasmPluginListenerInfo{
```
howardjohn, line 357:
```suggestion
	wasm := lb.push.WasmPluginsByListenerInfo(lb.node, model.WasmPluginListenerInfo{
```

No need for pointer, just introduces more possibility for errors

pilot/pkg/model/extensions.go
```diff
@@ -27,9 +27,12 @@ import (
 	"google.golang.org/protobuf/types/known/wrapperspb"
```
howardjohn, line 27: we should be adding validation to pkg/config/validation/validation.go for the new fields

howardjohn (commented):
lgtm, just minor comment

pilot/pkg/model/extensions.go
```diff
@@ -56,6 +74,58 @@ type WasmPluginWrapper struct {
...
+	}
+}
+
+func matchPorts(portSelectors []*typeapi.PortSelector, port int) bool {
+	if len(portSelectors) == 0 {
+		// If there is no specified port, match with all the ports.
+		return true
+	}
+	for _, ps := range portSelectors {
+		if ps.GetNumber() != 0 && ps.GetNumber() == uint32(port) {
```
howardjohn, line 122: very nit: slightly better to check `ps` against `nil` than GetNumber 0 since it could be explicitly set to `0`. In practice this is impossible due to validation, though

howardjohn (approved):

Issue #40801, opened 2022-09-05
Egress workload credential usecase not working since 1.15.0


LithiumSR:
### Bug Description

Hello everyone.
I have noticed from my testing that an egress scenario that uses the workload credential is not working anymore since 1.15.0 release (1.14.3 was tested to work fine).
The issue itself lead to the curl request targeting the out-of-mesh svc to return 
```
"upstream connect error or disconnect/reset before headers. retried and the latest reset reason: connection failure, transport failure reason: TLS error: 268435581:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED".
```
The scenario is also described in the istio doc: https://istio.io/latest/docs/tasks/traffic-management/egress/egress-tls-origination/#mutual-tls-origination-for-egress-traffic
I have verified that the issue has started showing up since #39133 (@howardjohn).

In order to reproduce the issue I have prepared the following steps (resources: [repro.tar.gz](https://github.com/istio/istio/files/9488746/repro.tar.gz)):
- Deploy a KinD cluster or have a cluster already ready
- Install a Istio 1.15.0 instance
- Execute the script that creates: certificates, sleep, sleep-role, sleep-rolebinding, nginx, nginx conf.
`sh script.sh`
- Run a curl request `kubectl -n test exec "$(kubectl -n test get pod -l app=sleep -o jsonpath={.items..metadata.name})" -c sleep -- curl -s http://my-nginx.mesh-external.svc.cluster.local:31443`
- Check that the curl request failed as mentioned.

Is the change in intentional consequence of the PR or is that a bug?

### Version

```prose
client version: 1.15.0
control plane version: 1.15.0
data plane version: 1.15.0 (3 proxies)
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:41:02Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.7", GitCommit:"1f86634ff08f37e54e8bfcd86bc90b61c98f84d4", GitTreeState:"clean", BuildDate:"2021-11-17T14:35:38Z", GoVersion:"go1.16.10", Compiler:"gc", Platform:"linux/amd64"}
```

### Additional Information

[...]

PR #40863, opened 2022-09-07
SAN improvements

howardjohn:
Fixes https://github.com/istio/istio/issues/40801

First, need to understand the bugs:

1. One bug is we mutated DestinationRule. While in push context we deepcopy the whole rule, we are still mutating shared copies within push context which leads to broken behavior when a DR applies to multiple Services
2. Due to a regression in https://github.com/istio/istio/pull/39133, we are not actually respecting `svc.ServiceAccount`. This is a fairly uncommon field which is service-level SAs; typically we get from Pod/WorkloadEntry
3. pushContext.ServiceAccounts does not take into account service namespace. So if there are multiple hostnames in different namespace (due to SE) we get pollution between them

This PR fixes all 3 of these issues and adds some tests. The behavior here is identical to 1.14 now, with the exception that bug (1) is present in 1.14 as well.

Release plan is to merge to master and 1.15 as is, and then merge the DeepCopy change and tests to 1.14


pilot/pkg/model/push_context.go
```diff
@@ -1432,21 +1438,30 @@ func SortServicesByCreationTime(services []*Service) []*Service {
...
+			var accounts sets.Set
+			func() {
+				// First get endpoint level service accounts
+				shard, f := env.EndpointIndex.ShardsForService(string(svc.Hostname), svc.Attributes.Namespace)
+				if f {
+					shard.RLock()
+					defer shard.RUnlock()
+					accounts = shard.ServiceAccounts
+				}
+				if len(svc.ServiceAccounts) > 0 {
```
ramaraochavali, line 1454: lock on service?

howardjohn: I am not sure what you mean?

ramaraochavali: Earlier we used to have a lock on service before reading service accounts like 
```

	s.RLock()
	sa := spiffe.ExpandWithTrustDomains(s.ServiceAccounts, ps.Mesh.TrustDomainAliases).SortedList()
	s.RUnlock()
```

howardjohn: thats from the `shard` (renamed from `s` to avoid confusion with service -- you are commenting on the model.Service. The shard is still locked

ramaraochavali: I see. Makes sense

stevenctl (approved):

kfaseela:
@howardjohn can we please have this cherry-picked to 1.15 as well?

Issue #39652, opened 2022-06-27
increased pilot pushes after istiod upgrade from 1.11.8 to 1.12.8


turbotankist:
### Bug Description

Istio version 1.11.8 is End of Support so we decided to upgrade our clusters to 1.12.8
We use istio-operator for deploy it.
The upgrade was successful but in one time we are figure out that some jobs fails because istio-proxy was not ready. We checked istiod resource usage and it was huge increased.
I had tried to downgrade only the istiod deployment to 1.11.8 and all restored to normal usage. In 20 minutes I changed the version to 1.12.8 again and usage was increased.
<img width="1147" alt="image" src="https://user-images.githubusercontent.com/9462236/176014926-a200698b-2843-4cb5-aa1c-74dd1788fcd3.png">
istiod has no other logs just they much more than early version
```
...
{"level":"info","time":"2022-06-27T18:36:36.676194Z","scope":"ads","msg":"CDS: PUSH for node:istio-ingressgateway-dps-7cc486466b-lnpr4.istio-system resources:6977 size:6.3MB cached:3497/3499"}
{"level":"info","time":"2022-06-27T18:36:36.677275Z","scope":"ads","msg":"EDS: PUSH for node:panel-7c6659d478-vcjs5.panel-dev resources:3061 size:772.2kB empty:0 cached:3061/3061"}
{"level":"info","time":"2022-06-27T18:36:36.715449Z","scope":"ads","msg":"LDS: PUSH for node:panel-7c6659d478-vcjs5.panel-dev resources:551 size:2.9MB"}
{"level":"info","time":"2022-06-27T18:36:37.100065Z","scope":"ads","msg":"NDS: PUSH for node:delay-queue-handler-stable-374-546894796f-8qppx.data-ceh-stable resources:1 size:297.4kB"}
{"level":"info","time":"2022-06-27T18:36:37.136090Z","scope":"ads","msg":"RDS: PUSH for node:clientpool-6c6c6bc77c-2vx8w.core-clientpool-dev resources:263 size:2.0MB cached:0/263"}
{"level":"info","time":"2022-06-27T18:36:37.141757Z","scope":"ads","msg":"NDS: PUSH for node:clientpool-6c6c6bc77c-2vx8w.core-clientpool-dev resources:1 size:297.4kB"}
[...]

PR #40866, opened 2022-09-08
Disable full push on scale from 1->0->1

howardjohn:
Fixes https://github.com/istio/istio/issues/39652

This reverts https://github.com/istio/istio/pull/36882. At the time,
that PR was needed because EDS ServiceAccounts and CDS ServiceAccounts
were decoupled; Since https://github.com/istio/istio/pull/39133, this is
no longer true, and the fix in 36882 is not needed any longer.

This PR *removes* the test added in 36882 (since it tests low level
details that are not relevant anymore). It improves the existing
TestEndpointFlipFlops test -- while that test *would* have caught the
regression, it didn't actually set any service accounts so it was
missed. The update changes it to correctly detect the behavior (it now
fails without this PR, passes with it).

### DIFF
--- a/pilot/pkg/model/endpointshards.go
+++ b/pilot/pkg/model/endpointshards.go
@@ -199,14 +199,6 @@ func (e *EndpointIndex) deleteServiceInner(shard ShardKey, serviceName, namespac
 	epShards := e.shardsBySvc[serviceName][namespace]
 	epShards.Lock()
 	delete(epShards.Shards, shard)
-	epShards.ServiceAccounts = sets.Set{}
-	for _, shard := range epShards.Shards {
-		for _, ep := range shard {
-			if ep.ServiceAccount != "" {
-				epShards.ServiceAccounts.Insert(ep.ServiceAccount)
-			}
-		}
-	}
 	// Clear the cache here to avoid race in cache writes.
 	e.clearCacheForService(serviceName, namespace)
 	if !preserveKeys {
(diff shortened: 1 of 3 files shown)


johnma14 (approved):

hzxuzhonghu:
cc

hzxuzhonghu (commented):
Could not see how you diable full push.

howardjohn:
before on delete we clean up the SAs, so it ends up empty. Then when the
pod comes back up, new SA is added, which is a change so it triggers a full
push.

Now we avoid changing the SA when it's removed so it doesn't change when
the new one comes back up

howardjohn:
it is fairly subtle... but we do have a test for it at least (and now it
works... before it didn't unfortunately)

hzxuzhonghu:
In case all pods of a service is down, and at this time another event triggers a full push, we will result in having incorrect cluster.

howardjohn:
Why? the cluster is unchanged during 1 pod, 0 pod, and then back to 1 pod -
it always has the last service account when there is 0 pods to avoid a full
push the remove the SA when it doesn't matter (if there are no endpoints SA
is never used anyways)

hzxuzhonghu:
But the cluster upstream tls setting can have the sas of deleted pods, we could know whether they will come back

howardjohn:
I believe that when we scale back yo we will 'reset' the service accounts.
So typically we come back up with the same account and no update needed. If
it's a different account, however, it should clear the old one out and
trigger a full push.

Let me add a test for that case

howardjohn:
@hzxuzhonghu I added the new case. So this shows that we full push if the SA changes, but do a partial push otherwise, which is the expected behavior.

ramaraochavali (approved):

howardjohn:
This should not be cherrypicked.

PR #40871, opened 2022-09-08
[WIP] Adding support for L4 metrics on Ambient's ztunnel

ymesika:

This PR adds the TCP stats filter to the ztunnel inbound for supporting L4 metrics with labels about source and destination workloads.

For the stats filter to set the correct labels on inbound metrics the following metadata needed to be available:
* Source workload metadata - This is available by propogating the baggage header from the client ztunnel to the server ztunnel either directly on the CONNECT request or via PEP.
* Destination workload metadata - Currently the Istio stats filter assumes a cluster metadata workload to be set and parses it to set the labels. This PR is setting this cluster metadata based on the matched cluster.

In a nutshell, there was a need to add an internal listener the inbound listener sends the traffic to the destination pod through it. The main (technical) reasons for this extra virtual hop are:
* To support the baggage handler HTTP filter to first extract the header on the CONNECT request and only then pick it up by an listener filter on the internal listener that prepares it for the stats filter to consume.
* Be able to set cluster metadata with the destination workload details


kyessenov (changes requested):
L4 proxy should avoid using internal listeners (or at most one) in the majority of cases.

kyessenov:
Ideally, no internal listeners are truly necessary for L4.

kris-nova:
So where is the ztunnel source code? 👀

howardjohn:
> So where is the ztunnel source code? 👀

This is a PR to the ztunnel (control plane side). The data plane part is just envoy

kyessenov:
@kris-nova ztunnel is Envoy that encaps and decaps h2 CONNECT for different service accounts. There's no code in istio for that, it's all in upstream envoy.

ymesika:
@kyessenov I can simplify the code in this PR by having a single internal listener and a set of metadata/stats filters.
The main reason for having an internal listener is that the beggage header needs to be mapped into metadata object by an http filter and only then metadata-to-peer-node and stats filter can use this object. But as you know hcm is the last one.

Can you think of a way to overcome this requirement without an internal listener?

kyessenov:
The main problem is that you're trying to leverage Istio's Wasm telemetry filters as a black box. That seems misguided. You could simplify by running an HTTP telemetry filter that emits TCP telemetry.

ymesika:
But then I won't be able to emit the source metadata (baggage) as metrics labels. Will I?

kyessenov:
@ymesika are you saying that encapping tcp_proxy doesn't process CONNECT response headers? That does seem like the case, but it is mostly because that's unusual. I think we can fix the upstream to get access to those headers.

kyessenov:
Alternatively, you can install stats extension as an L7 upstream filter instead for encapping proxy. This is a fairly new capability so it needs to be tested first.

kyessenov:
I filed https://github.com/envoyproxy/envoy/issues/23116. The least risky option is to just copy some response headers verbatim to tcp_proxy stream state. The other options seem rather complex.

ymesika:
I'm reading your comments above and the Issue/PR that you've already opened (Thanks!) and it seems like we have some mis-understanding. Just want to make sure we are aligned.
I think that you are focusing on a scenario where the source ztunnel is emitting the tcp metrics and I guess that's why you mentioned the baggage in the response headers?
I was focusing on the tcp metrics emitted by the destination ztunnel and therefore the baggage header is arriving on the CONNECT request and holds the metadata about the source workload. The metadata about the destination is based on the ECDS metadata that matches the original dst address (local workload).

kyessenov:
Yeah, I think it's better to adjust (fork?) stats filter so it behaves as follows:

- *Source*: network filter preceding encapping `tcp_proxy` that reads destination metadata from baggage response header in CONNECT (the linked upstream capability;
- *Destination*: HTTP filter preceding terminating `router` that reads source metadata directly from the header.

In both cases, local metadata comes from another per-node source (details TBD I assume).

I don't think it's important to split "peer info" extraction split into a separate filter like MX used to. The functionality is tiny and we're not going to run them as Wasm.

ymesika:
@kyessenov I've been experimenting with your suggestion of
> running an HTTP telemetry filter that emits TCP telemetry  

and I noticed that the TCP hooks (e.g. `onDownstreamData`) of the stats filter are not being called when running it as an HTTP filter.
So I wonder how to set the TCP metrics (such as `tcp_received_bytes_total`).
I can get the received/sent sizes from onRequestBody/onResponseBody callbacks but as for the number of connection opened/closed I'm not sure how to get them.

Just checking if that's what you had in mind when you mentioned the above and where can I get those TCP counters from.

kyessenov:
@ymesika The callbacks for HTTP and TCP filters are distinct. I think we'd need to add a special mode for the stats filter so that it'd act for HTTP callbacks as if they are TCP callbacks (e.g. implement `onRequestBody` for HTTP). Logically, that'd mean:

1. onRequestHeaders should register open-a-connection;
2. onRequestBody,onResponseBody should count bytes for the underlying stream;
3. onDone,onLog, etc should register close-a-connection.

Interval reporting is already implemented for HTTP for gRPC stats, so we just need to verify it works and enabled in this special mode.

ymesika:
Replaced by https://github.com/istio/istio/pull/41106

costinm:
AFAIK telemetry generation currently happens in Istio-specific code -
pretty sure Envoy upstream doesn't emit 'istio' metrics. I think we can
modify that code
so it doesn't ever use a tcp-level listener. The http telemetry generation
should be able to detect CONNECT and generate TCP telemetry - it is much
simpler
and cleaner.

Alternatively - and maybe better - we just use http telemetry as is - AFAIK
it is a super-set of TCP metrics. Since ambient protocol is to tunnel all
TCP
as H2 streams - with metadata, etc - I don't see why telemetry would not
reflect this, and cut 1/2 of the complexity and confusion.

kyessenov:
Re: second point. It's handled by tcp_proxy downstream and HCM upstream. So I think they should produce the same telemetry (TCP or HTTP) but not mix them up as that would be even more confusing.

kyessenov:
@ymesika I don't think it's safe to rely on Wasm/istio_stats filter for shared proxies. There's a memory leak caused by permanent metric retention, so it's very easy for one tenant to exhaust all metrics. I started https://github.com/istio/proxy/pull/4079 to rewrite it from scratch, and I think we'll need to invest into proper budgeting of metrics to avoid resource exhaustion.

costinm:
Not sure what 'budgeting' means - but the current model of metrics, with
all the per-client labels/attributes is pretty expensive, and may need a
fresh look at the design,
requirements and assumptions we made when the metric were originally
implemented.

Shared proxies are even more susceptible to  this - but it's a problem with
sidecars too.

kyessenov:
@costinm By budgeting I just mean that metrics need to be distributed fairly, so we have to account for resource consumption by each client group. Yes, this is already causing memory leaks in gateways, for example (or any hub-type proxy), because the existing implementation is naive.

kyessenov:
Like Ambient itself, the filter rewrite is an experimental feature. That means it should be revertible at any point. Because it'll take time to get to the same level, you're probably better off using older stats filter for now, and we'll duplicate code in two places later.

costinm:
Kuat - filter rewrite would have a great start with a doc/proposal - it is
expected that ambient telemetry will be a bit different, now is a good time
to revisit
cardinality and labels we enable by default, as well as possible alignment
with standards schemas (like open telemetry), or replacing the use of
deprecated open census. Many things have changed since the first filter and
design...

kyessenov:
@costinm No, my intent is not to redesign the external telemetry. I think we can optimize performance and scale of sidecar-to-sidecar performance (no ambient) using the building blocks of ambient without any visible changes. What I think we should not do is re-usesidecar telemetry building blocks for ambient, which you seem to agree with. It'd be great to have a new design for ambient telemetry.

kyessenov:
I completed the rewrite for (bar some minor features) (https://github.com/istio/proxy/pull/4079). Maybe we should switch to using that for Ambient instead?

PR #40887, opened 2022-09-08
`sets`: move to generic set

howardjohn:

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ramaraochavali (approved):

pkg/util/sets/string.go
```diff
@@ -14,44 +14,47 @@
...
-func NewWithLength(l int) Set {
-	return make(Set, l)
+func NewWithLength[T constraints.Ordered](l int) Set[T] {
+	return make(Set[T], l)
 }
 
 // New creates a new Set with the given items.
-func New(items ...string) Set {
-	s := NewWithLength(len(items))
+func New[T constraints.Ordered](items ...T) Set[T] {
```
hzxuzhonghu, line 31: This is making the usage more complex, as a fundamental lib , better ux is more important than dedup.

hzxuzhonghu (commented):
https://planetscale.com/blog/generics-can-make-your-go-code-slower 

Do we need a benchmark?

howardjohn:
I'll add one

pkg/util/sets/string.go
ramaraochavali: I agree to an extent with this. The usage certainly is more complex compared to earlier.

ramaraochavali: I agree to an extent with this. The usage certainly is more complex compared to earlier.

ramaraochavali: I agree to an extent with this. The usage certainly is more complex compared to earlier.

howardjohn: Its _slightly_ more complex for strings - but there are 100s of places where we _should_ be using Sets throughout the code but do not because it isn't generic.

If the concern is `sets.Set` -> `sets.Set[string]` is verbose I can add `type String = Set[string]` and then you can use it like `sets.String`?

ramaraochavali: That seems like a good idea for me for readability.

howardjohn:
```
name            old time/op    new time/op    delta
Set/insert-8      11.3µs ±84%    10.9µs ± 8%     ~     (p=0.156 n=10+9)
Set/contains-8    23.4ns ±25%    10.6ns ± 6%  -54.62%  (p=0.000 n=10+9)
Set/sorted-8       204µs ±54%      96µs ±66%  -53.07%  (p=0.000 n=10+9)

name            old alloc/op   new alloc/op   delta
Set/insert-8        192B ± 0%      192B ± 0%     ~     (all equal)
Set/contains-8     0.00B          0.00B          ~     (all equal)
Set/sorted-8      10.3kB ± 0%    10.2kB ± 0%   -0.24%  (p=0.000 n=10+9)

name            old allocs/op  new allocs/op  delta
Set/insert-8        2.00 ± 0%      2.00 ± 0%     ~     (all equal)
Set/contains-8      0.00           0.00          ~     (all equal)
Set/sorted-8        2.00 ± 0%      1.00 ± 0%  -50.00%  (p=0.000 n=10+10)
```

I think the `contains` being faster is a fluke, but sorted is actually a bit faster (less alloc). Main operations are identical

howardjohn:
Ok I made 3 tweaks:

* Added benchmarks, showing same performance before and after
* Added `sets.String` type to avoid `sets.Set[string]`
* Elided the type in `New[string](...)` where possible to be just `New()` like before

ericvn (approved):

pkg/util/sets/set.go
```diff
@@ -14,44 +14,49 @@
...
-import "sort"
+import (
+	"golang.org/x/exp/constraints"
+	"golang.org/x/exp/slices"
+)
 
-type Set map[string]struct{}
+type Set[T constraints.Ordered] map[T]struct{}
+
+type String = Set[string]
```
ramaraochavali, line 24: Can we please do the same for Int and also there are still usages like sets.New[string]?

howardjohn: New[string] cannot be avoided in all cases, only when we have given a hint already - either by predeclaring the return type or by passing on the variadic args 

For int - I didn't think we even have any for int anymore, the one we did use I switch to uint32

howardjohn: I think it's strictly worse and only feels better since generics aren't as common *yet*. Ots more wordy than just `New[string]` and less clear

howardjohn: Do you have any other concerns?

hzxuzhonghu (approved):

hanxiaop (approved):

zirain (approved):

Monkeyanator (approved):

PR #40892, opened 2022-09-09
Adding min istio version for new features to avoid running on older revisions

hemendrateli:

We are adding min istio version for tests related to below PRs as this functionalities were not there in previous revisions :-
	a. gRPC fault
           injection(https://github.com/istio/istio/pull/39295)
	b. Ignore port number in domain
           matching(https://github.com/istio/istio/pull/40475)
	c. Tunneling outbound traffic :- new tunnel field got
           introduced(https://github.com/istio/istio/pull/37968)
	d. Fix consistent hash based on source IP for TCP
           proxy(https://github.com/istio/istio/pull/38438)
	e. Traffic policy load balancer API
           changes(https://github.com/istio/istio/pull/39742)

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [X ] Test and Release
- [ ] User Experience

### DIFF
--- a/tests/integration/security/reachability_test.go
+++ b/tests/integration/security/reachability_test.go
@@ -33,6 +33,7 @@ import (
 	"istio.io/istio/pkg/test/framework/components/echo/echotest"
 	"istio.io/istio/pkg/test/framework/components/echo/match"
 	"istio.io/istio/pkg/test/framework/components/istio"
+	"istio.io/istio/pkg/test/framework/resource"
 )
 
 const (
@@ -113,6 +114,8 @@ func TestReachability(t *testing.T) {
 				expectCrossCluster condition
 				expectCrossNetwork condition
 				expectSuccess      condition
+				// minIstioVersion allows conditionally skipping based on required version
+				minIstioVersion string
 			}{
 				{
 					name: "global mtls strict",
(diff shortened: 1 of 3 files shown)


hzxuzhonghu:
Not sure i understand why you are testing older istio version with test cases in master

hemendrateli:
@hzxuzhonghu This is to test migration use case where multiple versions are installed and workloads communicate across revisions.

howardjohn (approved):

tests/integration/security/reachability_test.go
```diff
@@ -147,6 +151,7 @@ func TestReachability(t *testing.T) {
 					expectCrossCluster: notFromNaked,
 					expectCrossNetwork: notNaked,
 					expectSuccess:      notToNaked,
+					minIstioVersion:    "1.15.0",
```
aryan16, line 154: does that mean it won't work for versions lesser than 1.15 ?

hemendrateli: 1. If we try to test this master branch reachability tests against versions lesser than 1.15 then it won't work (so we are trying to skip) because of new API changes done(PR#39742) which are not there in version 1.13.
2. 1.13 branch already has its reachability tests which will work on 1.13 version.

PR #40897, opened 2022-09-09
Add image variable for gateway and keep auto as default

tiwarisanjay:
We are using istio for enterprise and we keep asm image in our private gcr. 
if we dont mention image explicitly it shows as a auto and our validation fails to verify the deployment for gateway. 
Just want to add the image variable in values.yaml. Still kept auto as default so nothing breaks for any existing setup. 

- [ ] Ambient
- [ ] Docs
- [ X] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ X] User Experience


howardjohn (commented):
You can use https://istio.io/latest/docs/setup/additional-setup/customize-installation-helm/

Issue #40890, opened 2022-09-09
Can't build istioctl for MacOS anymore


ajhall:
### Bug Description

When I run `BUILD_WITH_CONTAINER=0 make istioctl` on MacOS, the build quickly fails, saying:

```
Downloading envoy: https://github.com/istio/proxy/releases/download/1.0.2/istio-proxy-1.0.2-macos.tar.gz to /Users/andy.hall/Scratch/istio/out/darwin_arm64/release/envoy-1.0.2
curl: (22) The requested URL returned error: 404
```

I tried building https://github.com/istio/istio/tree/1.15.0 and the [current head of the `master` branch](https://github.com/istio/istio/commit/76ff6b45bcf8ada3dacd2a4fa8fa824a75afd694) and got the same result. This worked fine just a few days ago on [September 6th](https://github.com/macports/macports-ports/pull/15961/checks) when I submitted [an update to the MacPorts port](https://github.com/macports/macports-ports/pull/15961) of istioctl.

The problem is that https://github.com/istio/proxy/releases/download/1.0.2/istio-proxy-1.0.2-macos.tar.gz seems to be missing. I don't see any releases at all on [istio/proxy/releases](https://github.com/istio/proxy/releases), but I do see ([in the Wayback Machine](https://web.archive.org/web/20200916121147/https://github.com/istio/proxy/releases/tag/1.0.2)) that there used to be a 1.0.2 release that had that file. The [live version of that same link](https://github.com/istio/proxy/releases/tag/1.0.2) is just a tagged commit with no file attached.

## Build logs
```
$ BUILD_WITH_CONTAINER=0 make istioctl
TARGET_OUT=/Users/andy.hall/Scratch/istio/out/darwin_arm64 ISTIO_BIN=/Users/andy.hall/go/bin GOOS_LOCAL=darwin bin/retry.sh SSL_ERROR_SYSCALL bin/init.sh
Skipping envoy debug. Set DEBUG_IMAGE to download.
~/Scratch/istio/out/linux_arm64/release ~/Scratch/istio
Downloading envoy: https://storage.googleapis.com/istio-build/proxy/envoy-alpha-d81804c50e60d7ae17f3802006d76c24c98f2c5f-arm64.tar.gz to /Users/andy.hall/Scratch/istio/out/linux_arm64/release/envoy-d81804c50e60d7ae17f3802006d76c24c98f2c5f

real    0m1.577s
user    0m0.566s
sys     0m0.124s
[...]

PR #40912, opened 2022-09-09
Quick fix to get builds working on a Mac

ericvn:

https://github.com/istio/istio/issues/40890 was opened for an issue on Macs where the `make init` target fails. It fails since the old (1.0.2) Mac binary is no longer available. I do not necessarily know why the releases got removed from the repo.

The change is to simply copy the Linux binary as a place holder (at least to get things working). I don't know if anyone is actually trying to run the local Envoy binary on the Mac as it was very old, so just using the Linux binary as a placeholder should work. If this fails, as someone is using the Envoy binary, we can revise this later.

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience

### DIFF
--- a/bin/init.sh
+++ b/bin/init.sh
@@ -175,10 +175,8 @@ download_envoy_if_necessary "${ISTIO_ENVOY_LINUX_RELEASE_URL}" "$ISTIO_ENVOY_LIN
 download_envoy_if_necessary "${ISTIO_ENVOY_CENTOS_RELEASE_URL}" "$ISTIO_ENVOY_CENTOS_LINUX_RELEASE_PATH" "${SIDECAR}-centos"
 
 if [[ "$GOOS_LOCAL" == "darwin" ]]; then
-  # Download and extract the Envoy macOS release binary
-  download_envoy_if_necessary "${ISTIO_ENVOY_MACOS_RELEASE_URL}" "$ISTIO_ENVOY_MACOS_RELEASE_PATH" "${SIDECAR}"
-  ISTIO_ENVOY_NATIVE_PATH=${ISTIO_ENVOY_MACOS_RELEASE_PATH}
-else
+  # There is no longer an Istio built Envoy binary available. Copy the Linux binary as the Mac binary was
+  # very old and likely no one was really using it (at least temporarily).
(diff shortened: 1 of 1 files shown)


bin/init.sh
```diff
@@ -175,10 +175,8 @@ download_envoy_if_necessary "${ISTIO_ENVOY_LINUX_RELEASE_URL}" "$ISTIO_ENVOY_LIN
 download_envoy_if_necessary "${ISTIO_ENVOY_CENTOS_RELEASE_URL}" "$ISTIO_ENVOY_CENTOS_LINUX_RELEASE_PATH" "${SIDECAR}-centos"
 
 if [[ "$GOOS_LOCAL" == "darwin" ]]; then
-  # Download and extract the Envoy macOS release binary
-  download_envoy_if_necessary "${ISTIO_ENVOY_MACOS_RELEASE_URL}" "$ISTIO_ENVOY_MACOS_RELEASE_PATH" "${SIDECAR}"
-  ISTIO_ENVOY_NATIVE_PATH=${ISTIO_ENVOY_MACOS_RELEASE_PATH}
-else
+  # There is no longer an Istio built Envoy binary available. Copy the Linux binary as the Mac binary was
+  # very old and likely no one was really using it (at least temporarily).
   ISTIO_ENVOY_NATIVE_PATH=${ISTIO_ENVOY_LINUX_RELEASE_PATH}
```
howardjohn, line 180: before this was "else" - don't we need this to run still on linux? So just no "if"?

ericvn:
Yes, I should have just removed the `if` since we will always use the Linux binary (in the linux and Darwin out directories).

howardjohn (approved):

ericvn:
It looks like the directory is gitignored and may have been a remnant from something. I removed and lint now passes.

Not sure if the directory might want to be cleaned as well (being a out directory).

ericvn:
Create labels to push back to the currently supported releases as well.

howardjohn:
lint-yaml probably shouldn't lint the out/ folder...

PR #40797, opened 2022-09-05
operator: check pilot exist before get proxy info

zirain:
fixes: #40796 

**Please provide a description of this PR:**

check pilot exists before get proxy info

PR #40926, opened 2022-09-12
Address istiooperator removal loop potential

### LINKED ISSUE #40702: Istio operator will never delete finalizer if it fails to on the initial Reconciler attempt
### Bug Description

If something causes the Istio Operator deployment to fail on its initial attempts to delete the finalizer then it will never succeed to delete it because it then cannot successfully run the Helm reconciler even though its execution is unnecessary because it already succeeded. 

In the code below there is one attempt to retry but if that retry fails and we arrive at line 327 then it should retry this again during the next reconciliation. 
image
However, it also tries to unnecessarily run the  helm reconciler again resulting in this infinite loop which will never result in the finalizer being deleted.
image

This is an issue for uninstalls and canary upgrades.
 

### Version

```prose
❯ istioctl version
client version: 1.13.4
control plane version: 1.13.4
data plane version: none
❯ kubectl version --short
Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
Client Version: v1.24.0
Kustomize Version: v4.5.4
Server Version: v1.23.4
```

### Additional Information

_No response_

dhawton:

Closes #40702 

If you hit the Kube API server with multiple DELETE requests, it can cause the max retires of 1 to still fail due to the IstioOperator resource changing. On the next reconcile, HelmReconciler will fail because it cannot find the istiod deployment and this leaves the IstioOperator undeleted even though the resources have been removed.

The PR changes max retires to increase the likelihood that even if the Kube API server gets multiple DELETE requests it can succeed, and also looks for the Helm Reconciler erroring due to istiod not being in the cluster any longer to remove the IstioOperator.

### DIFF
--- a/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
+++ b/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
@@ -55,8 +55,10 @@ import (
 	"istio.io/istio/operator/pkg/util"
 	"istio.io/istio/operator/pkg/util/clog"
 	"istio.io/istio/operator/pkg/util/progress"
+	"istio.io/istio/pkg/config/constants"
 	"istio.io/istio/pkg/errdict"
 	"istio.io/istio/pkg/kube"
+	"istio.io/istio/pkg/proxy"
 	"istio.io/istio/pkg/url"
 	"istio.io/pkg/log"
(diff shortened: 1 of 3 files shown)


zirain:
is this related to https://github.com/istio/istio/pull/40797?

PR #40948, opened 2022-09-13
socks5 proxy for h2c

akshayjnambiar:

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/test/echo/server/forwarder/http.go
+++ b/pkg/test/echo/server/forwarder/http.go
@@ -30,6 +30,7 @@ import (
 	"github.com/lucas-clemente/quic-go"
 	"github.com/lucas-clemente/quic-go/http3"
 	"golang.org/x/net/http2"
+	"golang.org/x/net/proxy"
 
 	"istio.io/istio/pkg/hbone"
 	"istio.io/istio/pkg/test/echo"
@@ -119,7 +120,12 @@ func newHTTP2TransportGetter(cfg *Config) (httpTransportGetter, func()) {
 			// So http2.Transport doesn't complain the URL scheme isn't 'https'
 			AllowHTTP: true,
 			// Pretend we are dialing a TLS endpoint. (Note, we ignore the passed tls.Config)
-			DialTLS: func(network, addr string, _ *tls.Config) (net.Conn, error) {
+			DialTLSContext: func(ctx context.Context, network, addr string, _ *tls.Config) (net.Conn, error) {
+				if strings.Contains(cfg.Proxy, "socks5") {
+					splitProxyURL := strings.Split(cfg.Proxy, "://")
+					dialer, _ := proxy.SOCKS5("tcp", splitProxyURL[1], nil, proxy.Direct)
+					return dialer.Dial(network, addr)
+				}
 				return newDialer(cfg).Dial(network, addr)
 			},
 		}


hzxuzhonghu (commented):
What is it for?

pkg/test/echo/server/forwarder/http.go
```diff
@@ -119,7 +120,12 @@ func newHTTP2TransportGetter(cfg *Config) (httpTransportGetter, func()) {
 			// So http2.Transport doesn't complain the URL scheme isn't 'https'
 			AllowHTTP: true,
 			// Pretend we are dialing a TLS endpoint. (Note, we ignore the passed tls.Config)
-			DialTLS: func(network, addr string, _ *tls.Config) (net.Conn, error) {
+			DialTLSContext: func(ctx context.Context, network, addr string, _ *tls.Config) (net.Conn, error) {
+				if strings.Contains(cfg.Proxy, "socks5") {
+					splitProxyURL := strings.Split(cfg.Proxy, "://")
```
howardjohn, line 125: should probbly use url.Parse

pkg/test/echo/server/forwarder/http.go
```diff
@@ -119,7 +120,12 @@ func newHTTP2TransportGetter(cfg *Config) (httpTransportGetter, func()) {
...
+					dialer, _ := proxy.SOCKS5("tcp", splitProxyURL[1], nil, proxy.Direct)
+					return dialer.Dial(network, addr)
```
howardjohn, line 127: I am not sure this is the appropriate place. Why are we using socks5 only for http2?

It seems like it should be in the `newDialer` function

pkg/test/echo/server/forwarder/http.go
akshayjnambiar: done

pkg/test/echo/server/forwarder/http.go
howardjohn: hbone.Dialer is just proxy.Dialer + proxy.DialerContext. socks5 actually meets this, so I am pretty sure you could just type cast it.

Its still not clear to me why we use socks5 only for h2c though?

stewartbutler: @howardjohn So, to make sure I understand your proposal, your argument is that instead of having a separate function newProxyDialer which is called in only the h2 context we should instead put that code into newDialer, so it would look something like this:

```
func newDialer(cfg *Config) hbone.Dialer {

       // This is new
	if proxyURL.Scheme == "socks5" {
		out, _ := proxy.SOCKS5("tcp", proxyURL.Host, nil, proxy.Direct)
		return out
	}
	// End new code
	
	if cfg.Request.Hbone.GetAddress() != "" {
		out := hbone.NewDialer(hbone.Config{
			ProxyAddress: cfg.Request.Hbone.GetAddress(),
			Headers:      cfg.hboneHeaders,
			TLS:          cfg.hboneTLSConfig,
		})
		return out
	}
	out := &net.Dialer{
		Timeout: common.ConnectionTimeout,
	}
	if cfg.forceDNSLookup {
		out.Resolver = newResolver(common.ConnectionTimeout, "", "")
	}
	return out
}
```

Which should allow socks5 to be used as dialer for anything that calls newDialer (http1, http2, but not http3). And this works because proxy.Dialer can be cast to hbone.Dialer without issue.

Is that correct?

howardjohn: Yeah

Issue #40950, opened 2022-09-13
Istio gateway mode, CPU performance problems when the path magnitude is large


12345XXX:
In istio gateway mode, there are more than 5000 routes in our scenario, and each route has its own current limiting filter. The nested loop in patchHTTPRoute will repeatedly call proto.Clone, thus occupying the CPU

<img width="1406" alt="pilot " src="https://user-images.githubusercontent.com/43161619/189881767-fe15aa80-8070-4eee-9004-4b90e97e8925.png">

PR #40951, opened 2022-09-13
Let the double-layer loop deep copy object of large array change to once

12345XXX:
fix [#40950](https://github.com/istio/istio/issues/40950)
In our scenario, there are more than 5000 routers. When the router is in the merge filter, it will use proto.clone in a loop nested in the large array of routers. This modification moves the deep copy method to the outside of a loop,I think it is unnecessary to deep copy every time in the inner loop, a full deep copy can be made before the loop, so that the original A large number of calculations of n*n become n times

now:
pilot
fixed:
fix


12345XXX:
> I think we cannot share since they will be mutated by MERGE

It looks like this deep copy is a solution to different virtualHosts may share same routes pointer [ #18273]，(https://github.com/istio/istio/pull/18273) I think my modification different virtualHosts still has his own routes pointer

howardjohn:
Ah maybe so, I didn't look closely since the PR was closed.. did you mean to close it?

PR #40953, opened 2022-09-13
Add leader election for ambient waypoint controller

hzxuzhonghu:

### DIFF
--- a/pilot/pkg/ambient/controller/aggregate.go
+++ b/pilot/pkg/ambient/controller/aggregate.go
@@ -19,6 +19,7 @@ import (
 
 	"istio.io/istio/pilot/pkg/ambient"
 	"istio.io/istio/pilot/pkg/config/kube/crdclient"
+	"istio.io/istio/pilot/pkg/leaderelection"
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pkg/cluster"
 	"istio.io/istio/pkg/config/schema/gvk"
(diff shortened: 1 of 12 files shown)


pilot/pkg/ambient/controller/aggregate.go
```diff
@@ -118,21 +119,26 @@ func (a *Aggregate) clusterAdded(cluster *multicluster.Cluster, stop <-chan stru
...
-		// TODO handle istiodless remote clusters
-		initAutolabel(opts)
-		go func() {
-			if crdclient.WaitForCRD(gvk.KubernetesGateway, opts.Stop) {
-				waypointController := NewWaypointProxyController(opts.Client, opts.ClusterID, opts.WebhookConfig)
-				waypointController.Run(opts.Stop)
-			}
-		}()
+		election := leaderelection.
+			NewLeaderElection(opts.SystemNamespace, "pod name ", leaderelection.NamespaceController, "", opts.Client).
```
howardjohn, line 129: `"pod name "` shouldn't be a constant but derived from `args.PodName`

howardjohn: I think we want `revision` as well

hzxuzhonghu: Yes, i haven't done that

hzxuzhonghu:
The test definitely has goroutine leak, i will take a look

PR #40964, opened 2022-09-13
authz: add logging back

howardjohn:
We were building the log but never reporting it. This was lost in a refactoring that moved authz building from per-proxy to being more cached and reused. This required a bit more refactoring since we build HTTP/TCP once and parse it.


hzxuzhonghu (approved):
Looks not worth the complexity of logging once.

PR #40966, opened 2022-09-13
istio: improve deep copy for ServiceAttribute

yingzhuivy:

our istiod CPU profile shows that 10% of time is spent on DeepCopy of ServiceAttribute, so removing the use of copystructure. We already have fuzz test and benchmark for Service DeepCopy, which covers this part since ServiceAttribute is part of Service.

Benchmark results:
(before)
```goos: darwin
goarch: amd64
pkg: istio.io/istio/pilot/pkg/model
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkServiceDeepCopy
BenchmarkServiceDeepCopy-16    	  132760	      8190 ns/op
PASS
```
(after)
```goos: darwin
goarch: amd64
pkg: istio.io/istio/pilot/pkg/model
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkServiceDeepCopy
BenchmarkServiceDeepCopy-16    	 1213035	      1019 ns/op
PASS
```

### DIFF
--- a/pilot/pkg/model/service.go
+++ b/pilot/pkg/model/service.go
@@ -566,7 +566,47 @@ func (s *ServiceAttributes) DeepCopy() ServiceAttributes {
 	// Nested mutexes are configured to be ignored by copystructure.Copy.
 	// Disabling `go vet` warning since this is actually safe in this case.
 	// nolint: vet
-	return copyInternal(*s).(ServiceAttributes)
+	out := *s
+
+	if s.Labels != nil {
+		out.Labels = make(map[string]string, len(s.Labels))
+		for k, v := range s.Labels {
+			out.Labels[k] = v
+		}
+	}
+
+	if s.ExportTo != nil {
+		out.ExportTo = make(map[visibility.Instance]bool, len(s.ExportTo))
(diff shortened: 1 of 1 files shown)


pilot/pkg/model/service.go
```diff
@@ -566,7 +566,49 @@ func (s *ServiceAttributes) DeepCopy() ServiceAttributes {
 	// Nested mutexes are configured to be ignored by copystructure.Copy.
 	// Disabling `go vet` warning since this is actually safe in this case.
 	// nolint: vet
-	return copyInternal(*s).(ServiceAttributes)
+	out := *s
```
hzxuzhonghu, line 569: remove the above comment

howardjohn (approved):
https://github.com/istio/istio/pull/40987 tested it more, lgtm

pilot/pkg/model/service.go
yingzhuivy: followed up in https://github.com/istio/istio/pull/40990

PR #40970, opened 2022-09-14
ambient: minor refactor existsProxies/wantProxies

zirain:

- minor refactor existsProxies/wantProxies

### DIFF
--- a/pilot/pkg/ambient/controller/waypoint_controller.go
+++ b/pilot/pkg/ambient/controller/waypoint_controller.go
@@ -21,7 +21,6 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	klabels "k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	listerappsv1 "k8s.io/client-go/listers/apps/v1"
 	listerv1 "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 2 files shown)


pilot/pkg/ambient/controller/waypoint_controller.go
```diff
@@ -80,21 +83,11 @@ func NewWaypointProxyController(client kubelib.Client, clusterID cluster.ID, con
...
-			}, subresources...)
-			return err
-		},
 	}
 
 	rc.queue = controllers.NewQueue("waypoint proxy",
 		controllers.WithReconciler(rc.Reconcile),
-		controllers.WithMaxAttempts(5))
-
+		controllers.WithMaxAttempts(maxReconcileAttempts))
```
howardjohn, line 90: Nit: IMO (but probably not others opinion) this refactor of moving the constant just masks readability, not improves it

pilot/pkg/ambient/controller/waypoint_controller.go
```diff
@@ -127,7 +120,7 @@ func (rc *WaypointProxyController) Reconcile(name types.NamespacedName) error {
 		// Mostly used to avoid issues with local runs
 		return fmt.Errorf("injection config invalid, skipping reconile")
 	}
-	log := waypointLog.WithLabels("gateway", name.String())
+	scopedLog := waypointLog.WithLabels("gateway", name.String())
```
howardjohn, line 123: replacing `log` pattern is intentionally used here (and elsewhere) to ensure folks use the correctly scoped log and don't accidentally fall back to the unscoped one

howardjohn (commented):
Can you explain why we want to do this? https://github.com/istio/istio/blob/47af21a4088d20b3530df8872f1678f8fbb43618/pilot/pkg/config/kube/gateway/deploymentcontroller.go#L60 explains the rationale for the current approach

zirain:
> Can you explain why we want to do this?
> 
> https://github.com/istio/istio/blob/47af21a4088d20b3530df8872f1678f8fbb43618/pilot/pkg/config/kube/gateway/deploymentcontroller.go#L60
> 
> explains the rationale for the current approach

my original just like `no one likes YAML templating...`, but I do not notice the comments (that convince me).
I will revert this part.

PR #40986, opened 2022-09-14
Let the double-layer loop deep copy object of large array change to once

12345XXX:
fix https://github.com/istio/istio/issues/40950
In our scenario, there are more than 5000 routers. When the router is in the merge filter, it will use proto.clone in a loop nested in the large array of routers. This modification moves the deep copy method to the outside of a loop,I think it is unnecessary to deep copy every time in the inner loop, a full deep copy can be made before the loop, so that the original A large number of calculations of n*n become n times

### DIFF
--- a/pilot/pkg/networking/core/v1alpha3/envoyfilter/rc_patch.go
+++ b/pilot/pkg/networking/core/v1alpha3/envoyfilter/rc_patch.go
@@ -151,6 +151,8 @@ func patchHTTPRoutes(patchContext networking.EnvoyFilter_PatchContext,
 	routeConfiguration *route.RouteConfiguration, virtualHost *route.VirtualHost, portMap model.GatewayPortMap,
 ) {
 	routesRemoved := false
+	// different virtualHosts may share same routes pointer
+	virtualHost.Routes = cloneVhostRoutes(virtualHost.Routes)
 	// Apply the route level removes/merges if any.
 	for index := range virtualHost.Routes {
 		patchHTTPRoute(patchContext, patches, routeConfiguration, virtualHost, index, &routesRemoved, portMap)
@@ -249,9 +251,6 @@ func patchHTTPRoute(patchContext networking.EnvoyFilter_PatchContext,
 			routeConfigurationMatch(patchContext, routeConfiguration, rp, portMap) &&
 			virtualHostMatch(virtualHost, rp) &&
 			routeMatch(virtualHost.Routes[routeIndex], rp) {
-
-			// different virtualHosts may share same routes pointer
-			virtualHost.Routes = cloneVhostRoutes(virtualHost.Routes)
 			if rp.Operation == networking.EnvoyFilter_Patch_REMOVE {
 				virtualHost.Routes[routeIndex] = nil
 				*routesRemoved = true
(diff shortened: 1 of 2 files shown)


pilot/pkg/networking/core/v1alpha3/envoyfilter/rc_patch.go
```diff
@@ -151,6 +151,8 @@ func patchHTTPRoutes(patchContext networking.EnvoyFilter_PatchContext,
 	routeConfiguration *route.RouteConfiguration, virtualHost *route.VirtualHost, portMap model.GatewayPortMap,
 ) {
 	routesRemoved := false
+	// different virtualHosts may share same routes pointer
+	virtualHost.Routes = cloneVhostRoutes(virtualHost.Routes)
```
howardjohn, line 155: This is actually more expensive for the cases where we do not patch anything (common case), maybe we can make it a bit more conditional?

pilot/pkg/networking/core/v1alpha3/envoyfilter/rc_patch.go
```diff
@@ -249,9 +251,6 @@ func patchHTTPRoute(patchContext networking.EnvoyFilter_PatchContext,
 			routeConfigurationMatch(patchContext, routeConfiguration, rp, portMap) &&
 			virtualHostMatch(virtualHost, rp) &&
 			routeMatch(virtualHost.Routes[routeIndex], rp) {
-
-			// different virtualHosts may share same routes pointer
-			virtualHost.Routes = cloneVhostRoutes(virtualHost.Routes)
```
howardjohn, line 254: maybe instead of full clone we just shallow copy the list, and then when we merge (L260) we just copy+merge that single route?

Even with your patch we still clone more than we need in many cases?

pilot/pkg/networking/core/v1alpha3/envoyfilter/rc_patch.go
12345XXX: > This is actually more expensive for the cases where we do not patch anything (common case), maybe we can make it a bit more conditional?

It is true, it will be a big improvement in the scenario where our large number of routers need merge every time, but it will be more expensive when merge is not needed, I will try to add some conditions to him

pilot/pkg/networking/core/v1alpha3/envoyfilter/rc_patch.go
12345XXX: > maybe instead of full clone we just shallow copy the list, and then when we merge (L260) we just copy+merge that single route?
> 
> Even with your patch we still clone more than we need in many cases?

You mean we need to identify which router pointers are shared by different virtualHosts, we just do a deep copy of the shared routers so that we won't have clones we don't need, but this doesn't seem easy to implement

howardjohn: Not even identify them, just on L260 instead of `proto.Clone(original)` we do `proto.Clone(copy(original))`. Then we copy a single one instead of the whole list.

Imagine we have 1000 routes and just modify one

12345XXX: > Not even identify them, just on L260 instead of `proto.Clone(original)` we do `proto.Clone(copy(original))`. Then we copy a single one instead of the whole list.
> 
> Imagine we have 1000 routes and just modify one

I have submitted new code, you can review the new code

12345XXX:
@howardjohn I have a new submission, please help me to review it if you have time, thank you

hzxuzhonghu (commented):
LGTM, can you also add a  ut

12345XXX:
> LGTM, can you also add a ut

okay i have submitted some unit tests

howardjohn (approved):

zirain:
this seems a major issue about performance, we need backport to 1.15

/cherry-pick release-1.15

PR #40995, opened 2022-09-15
`PILOT_FILTER_GATEWAY_CLUSTER_CONFIG`: don't push xds when unrelated services are updated

dddddai:
For https://github.com/istio/istio/issues/39110#issuecomment-1139661877
If `PILOT_FILTER_GATEWAY_CLUSTER_CONFIG` is enabled, just skip the push when unrelated services change

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -99,7 +99,7 @@ type virtualServiceIndex struct {
 
 	// This contains destination hosts of virtual services, keyed by gateway's namespace/name,
 	// only used when PILOT_FILTER_GATEWAY_CLUSTER_CONFIG is enabled
-	destinationsByGateway map[string]sets.Set[string]
+	destinationsByGateway map[string]sets.String
 }
 
 func newVirtualServiceIndex() virtualServiceIndex {
@@ -110,7 +110,7 @@ func newVirtualServiceIndex() virtualServiceIndex {
 		delegates:                    map[ConfigKey][]ConfigKey{},
 	}
 	if features.FilterGatewayClusterConfig {
-		out.destinationsByGateway = make(map[string]sets.Set[string])
+		out.destinationsByGateway = make(map[string]sets.String)
 	}
 	return out
 }
@@ -772,21 +772,19 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 		return nil
 	}
 
-	if log.DebugEnabled() {
-		// host set.
-		hostsFromGateways := sets.New[string]()
-		for _, gw := range proxy.MergedGateway.GatewayNameForServer {
-			hostsFromGateways.Merge(ps.virtualServiceIndex.destinationsByGateway[gw])
-		}
-		log.Debugf("GatewayServices: gateway %v is exposing these hosts:%v", proxy.ID, hostsFromGateways)
+	// host set.
+	hostsFromGateways := sets.String{}
+	for _, gw := range proxy.MergedGateway.GatewayNameForServer {
+		hostsFromGateways.Merge(ps.virtualServiceIndex.destinationsByGateway[gw])
 	}
+	log.Debugf("GatewayServices: gateway %v is exposing these hosts:%v", proxy.ID, hostsFromGateways)
 
 	gwSvcs := make([]*Service, 0, len(svcs))
 
 	for _, s := range svcs {
 		svcHost := string(s.Hostname)
 
-		if ps.ServiceAttachedToGateway(svcHost, proxy) {
+		if _, ok := hostsFromGateways[svcHost]; ok {
 			gwSvcs = append(gwSvcs, s)
 		}
 	}
@@ -809,7 +807,7 @@ func (ps *PushContext) ServiceAttachedToGateway(hostname string, proxy *Proxy) b
 
 // add services from MeshConfig.ExtensionProviders
 // TODO: include cluster from EnvoyFilter such as global ratelimit [demo](https://istio.io/latest/docs/tasks/policy-enforcement/rate-limit/#global-rate-limit)
-func addHostsFromMeshConfig(ps *PushContext, hosts sets.Set[string]) {
+func addHostsFromMeshConfig(ps *PushContext, hosts sets.String) {
 	for _, prov := range ps.Mesh.ExtensionProviders {
 		switch p := prov.Provider.(type) {
 		case *meshconfig.MeshConfig_ExtensionProvider_EnvoyExtAuthzHttp:
@@ -1494,7 +1492,7 @@ func (ps *PushContext) initVirtualServices(env *Environment) error {
 	ps.virtualServiceIndex.publicByGateway = map[string][]config.Config{}
 
 	if features.FilterGatewayClusterConfig {
-		ps.virtualServiceIndex.destinationsByGateway = make(map[string]sets.Set[string])
+		ps.virtualServiceIndex.destinationsByGateway = make(map[string]sets.String)
 	}
 
 	virtualServices, err := env.List(gvk.VirtualService, NamespaceAll)
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -96,15 +96,23 @@ type virtualServiceIndex struct {
 	publicByGateway map[string][]config.Config
 	// root vs namespace/name ->delegate vs virtualservice gvk/namespace/name
 	delegates map[ConfigKey][]ConfigKey
+
+	// This contains destination hosts of virtual services, keyed by gateway's namespace/name,
+	// only used when PILOT_FILTER_GATEWAY_CLUSTER_CONFIG is enabled
+	destinationsByGateway map[string]sets.Set[string]
 }
 
 func newVirtualServiceIndex() virtualServiceIndex {
-	return virtualServiceIndex{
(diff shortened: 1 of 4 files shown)


pilot/pkg/model/push_context.go
```diff
@@ -793,6 +783,24 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 	return gwSvcs
 }
 
+func (ps *PushContext) GetHostsFromGateways(proxy *Proxy) sets.Set {
+	hostsFromGateways := sets.New()
+	for _, gw := range proxy.MergedGateway.GatewayNameForServer {
+		for _, vsConfig := range ps.VirtualServicesForGateway(proxy.ConfigNamespace, gw) {
+			vs, ok := vsConfig.Spec.(*networking.VirtualService)
+			if !ok { // should never happen
+				log.Errorf("Failed in getting a virtual service: %v", vsConfig.Labels)
```
howardjohn, line 792: I would not handle this and let it panic if it does - it's a programming bug if so and would fail immediately and be detected

pilot/pkg/xds/proxy_dependencies.go
```diff
@@ -67,6 +68,13 @@ func checkProxyDependencies(proxy *model.Proxy, config model.ConfigKey, push *mo
 		}
 	case model.Router:
 		if config.Kind == kind.ServiceEntry {
+			if features.FilterGatewayClusterConfig {
+				hostsFromGateways := push.GetHostsFromGateways(proxy)
```
howardjohn, line 72: this may be called many times, any concern about cost to compute? I suspect it's fine

hzxuzhonghu (commented):
good to see this as a start

pilot/pkg/xds/proxy_dependencies.go
dddddai: Yeah I have the same concern, I've add a cache for it

pilot/pkg/model/push_context.go
```diff
@@ -793,36 +802,59 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
...
+		hosts := sets.New()
+		for _, vsConfig := range ps.VirtualServicesForGateway(proxy.ConfigNamespace, gw) {
+			vs := vsConfig.Spec.(*networking.VirtualService)
+			for host := range virtualServiceDestinations(vs) {
+				hosts.Insert(host)
+			}
+		}
+		addHostsFromMeshConfig(ps, hosts)
+		// save the result
+		ps.virtualServiceIndex.destinationsByGateway[gw] = hosts
```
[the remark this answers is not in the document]
howardjohn: No, this is called concurrently

pilot/pkg/model/push_context.go
```diff
@@ -1569,6 +1577,42 @@ func (ps *PushContext) initVirtualServices(env *Environment) error {
 				}
 			}
 		}
+
+		if features.FilterGatewayClusterConfig {
```
howardjohn, line 1581: is it feasible to re-use virtualServiceDestinations instead of copy?

howardjohn (commented):
I was working on an unrelated PR and I think it may overlap with this one - possibly make it easier. Let me try some things and report back. If I cannot figure it out we can go with this though, I think its a good improvement

howardjohn:
Ok my original idea was  https://github.com/howardjohn/istio/pull/new/gateway/cluster-scope - I wanted to make the GatewayServices take into account the _previous_ services to workaround https://github.com/envoyproxy/envoy/issues/18958.

I was thinking I could utilize PrevSidecarScope but its not quite right...

One pattern I did end up on was just storing it in model.Proxy. That might simplify things here a bit possibly...

Note: it turns out my PR doesn't actually solve the problem I wanted, btw

pilot/pkg/model/push_context.go
dddddai: I don't think so, because then we can't do O(1) lookup by hostname, and `util/set` does not support types other than int, float and string

dddddai:
> Ok my original idea was https://github.com/howardjohn/istio/pull/new/gateway/cluster-scope - I wanted to make the GatewayServices take into account the _previous_ services to workaround [envoyproxy/envoy#18958](https://github.com/envoyproxy/envoy/issues/18958).

That's interesting, but what if the destination does not exist?(i.e. the service has not been created yet) 
"previous services" would not include it, so when the service is created afterwards, the proxy can't get xds push 

Also it would store duplicated services in per proxy if there are multiple gateway instances, while in this pr they share the same info. Maybe not a big deal though..

pilot/pkg/model/push_context.go
dddddai: Fixed

pilot/pkg/model/push_context.go
```diff
@@ -800,36 +798,44 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
...
 // TODO: include cluster from EnvoyFilter such as global ratelimit [demo](https://istio.io/latest/docs/tasks/policy-enforcement/rate-limit/#global-rate-limit)
-func getHostsFromMeshConfig(ps *PushContext) sets.String {
-	hostsFromMeshConfig := sets.New[string]()
-
+func addHostsFromMeshConfig(ps *PushContext, hosts map[string]sets.Set[int]) {
 	for _, prov := range ps.Mesh.ExtensionProviders {
 		switch p := prov.Provider.(type) {
 		case *meshconfig.MeshConfig_ExtensionProvider_EnvoyExtAuthzHttp:
-			hostsFromMeshConfig.Insert(p.EnvoyExtAuthzHttp.Service)
+			hosts[p.EnvoyExtAuthzHttp.Service] = nil
```
howardjohn, line 818: I don't get it, why are we setting to nil? doesn't that erase all the ports we have in the inner set?

pilot/pkg/model/push_context.go
```diff
@@ -1480,6 +1486,10 @@ func (ps *PushContext) initVirtualServices(env *Environment) error {
 	ps.virtualServiceIndex.privateByNamespaceAndGateway = map[string]map[string][]config.Config{}
 	ps.virtualServiceIndex.publicByGateway = map[string][]config.Config{}
 
+	if features.FilterGatewayClusterConfig {
+		ps.virtualServiceIndex.destinationsByGateway = make(map[string]map[string]sets.Set[int])
```
howardjohn, line 1490: why `sets.Set[int]`? I don't see us using the port value anywhere?

dddddai: > is it feasible to re-use virtualServiceDestinations instead of copy?

Because `virtualServiceDestinations` returns `map[string]sets.Set[int]`, if we want to keep hosts only, we will have to copy the keys out

So we have 3 options:
1. extract destinations directly from vs
2. reuse `virtualServiceDestinations`
3. reuse `virtualServiceDestinations` and copy hosts out

Which one would you prefer?

pilot/pkg/model/push_context.go
```diff
@@ -800,36 +798,44 @@ func (ps *PushContext) GatewayServices(proxy *Proxy) []*Service {
 	return gwSvcs
 }
 
+func (ps *PushContext) ServiceAttachedToGateway(hostname string, proxy *Proxy) bool {
```
hzxuzhonghu, line 801: I am thinking this function call can be not very efficient

Actually, there are two loops

dddddai: IMO it's relatively trivial compared to a whole push

pilot/pkg/model/push_context.go
dddddai: Because we don't care about the ports, we only need to include these hosts in `destinationsByGateway`

pilot/pkg/model/push_context.go
hzxuzhonghu: What i am thinking is can we use a map

dddddai: What's the key/value of the map?

dddddai:
@howardjohn Any comments?

howardjohn (commented):
Mostly lgtm, just a bit concerned with the possible double loop @hzxuzhonghu mentioned. IIUC that is going to be pretty large size on both iterations making it N^2?

pilot/pkg/model/push_context.go
dddddai: I've removed the nested loop in `GatewayServices`, PTAL

pilot/pkg/model/push_context.go
```diff
@@ -1572,6 +1578,21 @@ func (ps *PushContext) initVirtualServices(env *Environment) error {
...
+				if gw == constants.IstioMeshGateway {
+					continue
+				}
+				if _, f := ps.virtualServiceIndex.destinationsByGateway[gw]; !f {
+					ps.virtualServiceIndex.destinationsByGateway[gw] = sets.New[string]()
+				}
+				for host := range virtualServiceDestinations(rule) {
+					ps.virtualServiceIndex.destinationsByGateway[gw].Insert(host)
+				}
+				addHostsFromMeshConfig(ps, ps.virtualServiceIndex.destinationsByGateway[gw])
```
hzxuzhonghu, line 1593: nit: This can also be improved,

hzxuzhonghu (approved):

PR #40997, opened 2022-09-15
pilot: fix remote JWKS endpoint host assignment

icereval:

Fixes a bug where remote JWKS URI's specified without a host port fail to parse into their host and port components.

Example: https://go.dev/play/p/vtXEUwF8s7S
- jwtRule.JwksUri = "http://istio.io/jwks"
  - net.SplitHostPort would error, and host would be empty.

### DIFF
--- a/pilot/pkg/security/authn/v1beta1/policy_applier.go
+++ b/pilot/pkg/security/authn/v1beta1/policy_applier.go
@@ -16,10 +16,7 @@ package v1beta1
 
 import (
 	"fmt"
-	"net"
-	"net/url"
 	"sort"
-	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
should we add some tests?

icereval:
Sounds good, I'll work on the unit tests tonight.

icereval:
@howardjohn, I see now there is an existing function for parsing JWKS URI's that was already [tested](https://github.com/istio/istio/blob/master/pkg/config/security/security_test.go#L24), so it seemed appropriate to reuse `ParseJwksURI`.

howardjohn (approved):
Nice!

Issue #40984, opened 2022-09-14
EKS nodes don't come up after scaling down and then scaling up if enableNamespacesByDefault is true


sachin-ks:
### Bug Description

EKS nodes don't come up after scaling down and scaling up the EKS nodes if enableNamespacesByDefault is true.

Warning from istiod-1-15-0 pod:

 

>  Warning  FailedScheduling  3m43s                 default-scheduler  0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.
> Warning  FailedScheduling  41s (x3 over 2m41s)   default-scheduler  0/2 nodes are available: 2 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.

Warning from istiod-1-15-0 service:

> Warning | FailedToUpdateEndpointSlices | 3 minutes | endpoint-slice-controller | Error updating Endpoint Slices for Service istio-system/istiod-1-15-0: node "ip-192-168-5-17.us-west-2.compute.internal" not found

Screenshot of the above warning:
<img width="1127" alt="Screenshot 2022-09-14 at 4 19 24 PM" src="https://user-images.githubusercontent.com/20788238/190135374-f42c422a-9c2a-4d69-829f-b4e4a67786da.png">

Warning from kube-proxy DaemonSet:

> Warning | FailedCreate | 13 minutes | daemonset-controller | Error creating: Internal error occurred: failed calling webhook "auto.sidecar-injector.istio.io": failed to call webhook: Post "https://istiod-1-15-0.istio-system.svc:443/inject?timeout=10s": no endpoints available for service "istiod-1-15-0"

Warning from aws-node DaemonSet:

> Warning | FailedCreate | 14 minutes | daemonset-controller | Error creating: Internal error occurred: failed calling webhook "auto.sidecar-injector.istio.io": failed to call webhook: Post "https://istiod-1-15-0.istio-system.svc:443/inject?timeout=10s": no endpoints available for service "istiod-1-15-0"

Screenshot of above warning from EKS UI:
<img width="1159" alt="Screenshot 2022-09-14 at 4 08 23 PM" src="https://user-images.githubusercontent.com/20788238/190132649-04931894-a2dd-4e5c-8bde-60af18dd0dca.png">

### Version

```prose
$ ./istioctl version
client version: 1.15.0
control plane version: 1.15.0
data plane version: none
$ kubectl version --short
[...]

PR #41020, opened 2022-09-16
Exclude system namespaces from auto.sidecar-injector.istio.io namespaceSele…

cnvergence:
Since Kubernetes v1.21, it is possible to exclude namespaces by label "kubernetes.io/metadata.name".
Fixes:
- https://github.com/istio/istio/issues/40984

### DIFF
--- a/manifests/charts/default/templates/mutatingwebhook.yaml
+++ b/manifests/charts/default/templates/mutatingwebhook.yaml
@@ -113,6 +113,9 @@ webhooks:
       operator: DoesNotExist
     - key: istio.io/rev
       operator: DoesNotExist
+    - key: "kubernetes.io/metadata.name"
+      operator: "NotIn"
+      values: [ "kube-system" ]  
   objectSelector:
     matchExpressions:
     - key: sidecar.istio.io/inject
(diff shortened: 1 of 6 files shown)


manifests/charts/default/templates/mutatingwebhook.yaml
```diff
@@ -113,6 +113,9 @@ webhooks:
       operator: DoesNotExist
     - key: istio.io/rev
       operator: DoesNotExist
+    - key: "kubernetes.io/metadata.name"
+      operator: "NotIn"
+      values: [ "kube-system" ]
```
sachin-ks, line 118: Should we add all the namespaces that are ignored by Istio for injection?
https://github.com/istio/istio/blob/master/pkg/kube/inject/initializer.go#L38

howardjohn: +1 - anytime these actually _are_ triggered they will be ignored immediately, so no point ever triggering them

cnvergence: 👍🏻

howardjohn (approved):

PR #41052, opened 2022-09-19
update golang to 1.19 for tcp-echo

my-git9:
update golang to 1.19 for tcp-echo

### DIFF
--- a/samples/tcp-echo/src/Dockerfile
+++ b/samples/tcp-echo/src/Dockerfile
@@ -16,7 +16,7 @@
 ARG BASE_DISTRIBUTION=default
 
 # build a tcp-echo binary using the golang container
-FROM golang:1.14.2 as builder
+FROM golang:1.19 as builder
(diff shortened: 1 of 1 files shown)


frankbu:
This PR, and some of the other recent changes to sample Dockerfiles, are changing the build images without actually building, testing, and using the the new image. Although mostly trivial changes, we are not verifying that they work, and more importantly, the images that people are using are not exactly what they think they are from looking at the source. I wonder if we should not be changing any sample source code without also updating the images in dockerhub?

ericvn:
Maybe we need to move some of the README info either out to the main samples directory or create/update the various READMEs as needed. Similar to Bookinfo we would want to discuss how to build the temp images and test, discuss working with someone (not sure who actually has authority to push images at present besides Frank and John) to get the images updated, and then add the appropriate yaml changes to use those new images.

ericvn:
I do think we want the ability to update src and images, but agree that the process there is not the best.

